PR #96, opened 2026-08-18
scripts: no recursive delete takes a clobberable path (third attempt at landing #75)

obra:
Re-lands kata `fgxs`. **Third attempt** — the previous two never reached main, and #87 is closed in favour of this.

#75 was based on `kata/5hs2-mktemp-guard`; that branch merged to main first, so #75's merge landed on an already-merged branch and its 27 files were stranded. #87 re-targeted onto main but went stale before it could merge — 24 further commits landed, six touching files this changes.

Main today: **19 of 28 selftests lack the guard**, and roughly 20 sites still hand a clobberable variable to a recursive delete. The no-argument `selftest_rm_scratch` from #73 is there; none of the conversions are.

## The rule

After kata 5hs2's work destroyed a machine, Jesse's rule:

> There should never be a recursive delete in any of our scripts that takes an argument that could possibly get clobbered.

A delete that accepts a path can always be handed the wrong one, and no checking inside the callee makes the caller's variable trustworthy. #73 removed the argument. This converts everything that still hands one over.

## What changed

**Selftests — 28 of 28** now build scratch with `selftest_scratch` and clean up with the no-argument `selftest_rm_scratch`.

- `fuzz-bisect-selftest.sh` sourced the library after it needed it; that moved up.
- `live-compaction-eval-selftest.sh` and `live-eval-isolation-selftest.sh` did not source it at all. Converting them also drops their `mktemp -d -t`, which ignores `TMPDIR` on macOS and so escapes both the wave's per-suite isolation and its leftover check.
- Seven per-case `mktemp`s build under an already-validated root and cannot use `selftest_scratch`, which mints under `TMPDIR` by design. They take the inline check the audit accepts.

**Tools — two** with an unchecked `mktemp` feeding a delete now refuse instead: `fuzz-drive.sh` and `run-module-tests.sh`. `run-module-lint.sh` already had one.

## Four files dropped in favour of main's better version

Two rounds of `fix(tooling): arm the ... cleanup trap before the scratch exists` have landed on main, covering `coverage-union.sh`, `test-coverage-floor.sh`, `fuzz-coverage.sh` and `fuzz-coverage-global.sh`. Each conflicted, and **main wins every time**: it removes `mktemp` entirely, builds the path from `$$`, and arms the trap *before* the directory exists.

That closes a window this PR's approach did not — a signal landing after `mktemp -d` created the directory but before the trap knew about it. Keeping my versions would have been a regression, so they are gone.

[...]

PR #102, opened 2026-08-18
scripts: delete the selftest- selftest layers and make variable-fed deletes unrepresentable

obra:
Closes katas jvpe and yns5. Approved by Jesse 2026-08-17, including every deletion. **Rebuilt on current main** after the adversarial review's reject-in-current-shape verdict; the original branch was a redo of #96 from the #91 base and conflicted 24 ways with main.

**+622 / −2,798 across 48 files vs main (at e0ff41e8a; the 3a0641b92 follow-up landed in the merge).** The estate goes from 28 selftest suites to 21; ~2,400 lines of stub-world selftest and ~900 lines of spent tooling are gone.

## Deleted

Criteria: a suite survives only where it pins outcomes of a tool the gate or CI depends on; argv-shape assertions against faked binaries test the mock; hand-run conveniences fail loudly in front of whoever ran them; spent one-shot migrations live on in git history.

tmux-send/tmux-read (tools + suites), report-tmp-debris-selftest, report-orphaned-worktrees-selftest, deploy-hub-selftest, live-compaction-eval-selftest, scenario-cite-migrate (tool + suite), parallelize-tests, test-overlap, and selftest-lib-selftest (replaced below).

## The guard, promoted and enforced

- `scripts/scratch-lib.sh`: the scratch guard under domain names (`scratch_dir` / `scratch_rm`); `selftest-lib.sh` sources it, and 26 scripts call it.
- `scratch-lib-selftest.sh` replaces the old sabotage loops (8 real suites re-run twice each per gate) with direct fixtures against decoys, plus failing-mktemp and missing-TMPDIR fixtures preserving the two refusal behaviours the loops pinned. 32 checks, and no fixture contains a live delete.
- **TestNoScriptFeedsVariableToRecursiveDelete**: any `rm -r` taking a `$`-expanding argument in `scripts/` or in the repo-root `install.sh` fails the build unless it is on a count-pinned list whose every entry states its reason. **26 pinned lines across 20 files**, each with a reason, where main has 29 unstated. The count stops a listed file gaining an *extra* variable-fed delete and fails a listed file whose lines went away; it cannot see one blessed delete swapped for a different one inside an already-listed file, and the audit's per-line text scan is blind to `find -exec rm`, `xargs rm`, `rm<TAB>` and backslash-newline continuations. Both limits are stated in the audit and in docs/testing.md, because reviewers are what holds them.
- `docs/testing.md` records the four standing rules (no clobberable delete arguments; no live weapons in tests; one meta-level of test depth; outcome pins only, new logic in Go).
- `e2e-webui-turn-controls.sh` readiness polls: 6s guess → 60s tripwire (its own comment prescribed raising the count; the 6s version false-redded two gates in one day).

## What the rebuild changed from the reviewed shape

- **The four coverage runners are NOT converted.** `test-coverage-floor.sh`, `coverage-union.sh`, `fuzz-coverage.sh` and `fuzz-coverage-global.sh` keep main's `PREFIX.$$` name-first / trap-armed / `mkdir` ordering, byte-identical to main, and so do `covscratch-lib.sh` and `scratch-selftest-lib.sh`. That ordering has a zero-width abandon window (0/150 leaks under SIGTERM-at-first-scratch, against 29/150 for a mint-then-trap mktemp shape), and #105's owner-side reclaim parses the pid out of that basename — a random mktemp suffix would make it a permanent no-op. The four are on the audit's pinned list with that reason instead. `scratch_dir` is for the selftest suites and the tools where mktemp naming is fine.
- **The kill-signal contracts stay.** TERM/INT/HUP × test-coverage-floor and coverage-union, plus #105's reclaim/concurrent/keep-on-failure asserts, all still run out of `scratch-selftest-lib.sh`.
- **`build-runtime-pair.sh` is not converted** (review B1/B2). `runtime_pair_build_test.go` execs it as `sh`, which ignores the bash shebang and chokes on the library's array syntax under dash; and `scratch_dir`'s canonicalization turns `/var/folders/...` into `/private/var/...`, deterministically failing the prefix assertion at `runtime_pair_build_test.go:61` on macOS. Its old delete was already checked and canonicalized, so the conversion bought nothing.
- **Merged-main fixes preserved**, not reverted: 14c8d6fcd (fuzz-drive-selftest's checked mktemp and its removal from the allowlist), fde632f8c and #91's patterns in the coverage runners, and all of #105.
- **Corrected claims.** The count-pin now says what it actually guarantees. `live-eval-isolation.sh`'s allowlist reason no longer claims its mint is checked (it is a bare mktemp; the cleanup guard beside it is what makes the delete safe). `report-tmp-debris.sh`'s reason no longer calls its heredoc quoted (it is unquoted, with the pinned line escaping its own expansion). `install.sh` is in scope now, and its mint is checked and empty-guarded rather than left as a documented gap.
- Dead `tmp_base` assignment removed from `run-module-tests.sh`.

## Verification (macOS, on the rebuilt branch)

- Three script audits green. Full dev-tooling wave green **21/21**, twice.
- `go test -count=1 -short .` at repo root green — the B2 regression check.
- `go test -run TestRuntimePairBuild -count=1 -v .` — all three PASS, proving B1 is gone.
- Kill contracts and #105 reclaim asserts green in both coverage selftests: SIGTERM/SIGINT/SIGHUP leave no scratch; the next run reclaims an abandoned scratch; a live process's scratch survives; a failed run keeps its scratch and the next run reclaims it. test-coverage-floor 41 checks, coverage-union 24 checks, fuzz-coverage-global 150 checks, scratch-lib 32 checks, 0 failed.
- Mutation check: a scratch file reintroducing the kata 5hs2 shape (unchecked `mktemp -d`, `cd`-canonicalized, `trap 'rm -rf "$var"'`) fails **both** audits with the exact offending lines; removing it restores green. Adding a second variable-fed delete to an allowlisted file fails the count in the "extra weapon" direction; removing that file's only pinned line fails it in the "stale entry" direction.
- shellcheck on every changed script: no new findings beyond the pre-existing SC1091 "not following sourced library" notes and one SC2154 in scratch-lib-selftest.sh:38 ("work is referenced but not assigned") — a false positive from scratch_dir assigning through printf -v, which shellcheck cannot see.


obra:
Rebased onto main after #96 and #98 landed (the sweep branch had been rebased before merge, so the old base's SHAs no longer existed). Three conflicts, all in files where main gained follow-up fixes after this branch forked: the coverage runners' disarm-on-stale-mkdir refinement (superseded — scratch_dir mints fresh names, and scratch_rm's registry makes foreign directories structurally unreachable) and the audit allowlist comment (took main's accurate post-fix text with this branch's renames; fuzz-drive-selftest's entry stays gone since main fixed its per-scenario mktemp). Post-rebase verification: all three script audits green, runtime-pair build tests green, scenario citation audits green, full dev-tooling wave 21/21 twice.

obra:
Adversarial review (full detail available on request). Verdict was REJECT in current shape; Jesse is reworking. The findings that matter for the rework:

**Broken on both platforms:**
- B1: `build-runtime-pair.sh` is invoked as `exec.Command("sh", …)` by `runtime_pair_build_test.go:1036`, ignoring the new bash shebang — `scratch-lib.sh:21`'s array syntax is a dash syntax error on ubuntu CI. Takes out the whole TestRuntimePairBuild* family.
- B2: `scratch_dir` canonicalizes (`cd && pwd -P`) where the old mktemp didn't; on macOS that turns `/var/folders/…` into `/private/var/folders/…` and deterministically fails `TestRuntimePairBuildContainsProcessStateAndPreservesGoCaches` (prefix assertion at `runtime_pair_build_test.go:61`). This conversion buys nothing — its old delete was already checked and canonicalized; consider dropping it.

**Base problem:** the branch redoes #96 from the #91 base (its first commit is the superseded draft of 01feeb19c) → 24 conflicts vs main. Hand-resolving toward the PR side would silently revert two merged fixes while the audits stay green: 14c8d6fcd (fuzz-drive-selftest mktemp guard + allowlist removal) and fde632f8c (trap-disarm on stale same-pid scratch). Rebase onto current main instead.

**Protection regressions to reconcile:**
- Main's coverage runners have a zero-width abandon window (name-first/trap/mkdir): 0/150 leaks under SIGTERM-at-first-scratch; the mktemp-XXXXXX replacement measured 29/150, and the kill-signal selftest contracts (TERM/INT/HUP × 2 suites) were deleted, not replaced. The new comments claim 'a signal in the window deletes nothing' — true, but it now *leaks* instead, which is the hazard the old comment was about.
- #105 (merged) makes each runner reclaim its own dead-pid leftovers by parsing the `PREFIX.$$` basename; mktemp's random suffixes make that reclaim a permanent no-op and fail its selftest asserts. The $$+trap-first naming needs to survive the rework (or the reclaim design needs a replacement decided jointly).

**Claim corrections:** the count-pin 'an entry can neither hide a new weapon nor go stale' is falsifiable by substitution inside an allowlisted file (demonstrated); `live-eval-isolation.sh`'s allowlist reason says 'its mint is checked' — it isn't (line 32 is a bare mktemp; the cleanup guard at :65 is what actually makes it safe). Scanner blind spots worth documenting next to find/xargs: `rm<TAB>` and backslash-newline continuations. `install.sh` has the banned shape but the audit only reads `scripts/`.

**Worth keeping intact:** `scratch_dir`/`scratch_rm` as the one library, the direct library selftest (32 checks incl. failing-mktemp and missing-TMPDIR fixtures), `TestNoScriptFeedsVariableToRecursiveDelete` (mutation-verified in both directions), and the four rules in docs/testing.md.

obra:
Reworked per the adversarial review. Head is now `e0ff41e8a`, rebased onto current main (`9c6a6f1e4`), still one commit; the commit message was rewritten to match what the change now does.

**B1 (sh ignores the shebang; dash can't parse the guard's arrays):** `scripts/build-runtime-pair.sh` and `runtime_pair_build_test.go` are restored to main's versions byte-for-byte — the conversion is dropped, not patched. The script stays POSIX sh by contract since its own test execs it via `sh`. Its one trap delete is pinned with that reason. Verified: `sh -n` parses it, and the whole `TestRuntimePairBuild*` family passes locally.

**B2 (scratch_dir canonicalization vs the `/var` prefix assertion):** gone with the same restore. The conversion bought nothing — its old delete was already checked; the reviewer was right to say drop it.

**Abandon-window regression + #105 collision:** all four coverage runners and their three selftests are restored to main wholesale. That keeps the pid-suffixed name-first/trap-first/mkdir pattern (0/150 leaks measured vs 29/150 for mint-then-trap), keeps covscratch-lib's owner-side reclaim working (it parses the pid out of the basename — mktemp's random suffixes would have made it a permanent no-op and failed its selftest asserts), and brings back the kill-signal contracts (TERM/INT/HUP) and `scratch-selftest-lib.sh`, which is un-deleted. The only delta vs main in those files is the mechanical `selftest_scratch`→`scratch_dir` rename every surviving suite got when the guard moved into scratch-lib.sh. Because the restore takes main's files as-is, the two previously-at-risk merged fixes (14c8d6fcd, fde632f8c) survive by construction. docs/testing.md rule 1 now names this as the second sanctioned pattern instead of pretending scratch_dir is the only door.

**Audit re-pinned against the restored reality (RED first, then pinned):** new count-pinned entries with reasons for the four runners' trap deletes, covscratch-lib's reclaim (the delete IS the job), scratch-selftest-lib's four `${tmphome:?}`-guarded fixture resets, fuzz-coverage-global-selftest's concurrent-decoy cleanup, and build-runtime-pair. The stale `reclaim-test-debris-selftest.sh` entry (allowed 1, contains 0) is removed.

**Claim corrections, all applied:**
- The count-pin comment now says what it can't catch: same-count substitution inside a listed file still passes; review has to read the deletes.
- `live-eval-isolation.sh`'s reason now tells the truth: bare mint; the cleanup guard refusing an empty root is what makes the delete safe.
- Scanner blind spots documented next to the find/xargs note: `rm<TAB>` and backslash-newline continuations.
- `install.sh`: the scan now covers it (repo-root shell that gets curl|sh'd into other people's shells). Watched it fail RED on `install.sh:57`, then gave the mint the checked-and-empty-guarded shape and made cleanup refuse an empty root; pinned at 1 with that reason.
- The Makefile wave comment no longer claims every script goes through the guard; it names what the audits actually enforce.

**Verification on `e0ff41e8a`:** all three script audits + scenario audits green; `TestRuntimePairBuild*` green; dev-tooling wave 21/21 twice; `make merge-approval-gate` green in a clean /tmp worktree; home canary 32 before and after.

obra:
Rework done. The branch is rebuilt on current main (9c6a6f1e4), not hand-resolved — two sessions converged on the same rebuild, so what is on the branch now is `e0ff41e8a` (the rebuild) plus `3a0641b92` (four corrections that rebuild had missed). Nothing was clobbered; the force-with-lease refused the first attempt, which is how the collision surfaced.

**Kept, per the review:** `scratch-lib.sh` with `scratch_dir`/`scratch_rm` as the one library and its 32-check direct selftest; `TestNoScriptFeedsVariableToRecursiveDelete` and `TestScratchDirCannotResolveToCWD`; the four rules in docs/testing.md; the seven suite deletions and the layered sabotage machinery's removal.

**Preserved from merged main, verified byte-identical to `origin/main`:** the four coverage runners, `covscratch-lib.sh`, `scratch-selftest-lib.sh`, `reclaim-test-debris.sh`, `build-runtime-pair.sh` and `runtime_pair_build_test.go`. `fuzz-drive-selftest.sh` differs from main by exactly the two guard-rename lines, so 14c8d6fcd's checked mktemp and its absence from the allowlist survive.

**Blockers fixed:**
- B1/B2 — the `build-runtime-pair.sh` conversion is dropped entirely. `go test -run TestRuntimePairBuild -count=1 -v .` passes all three on macOS.
- The coverage runners keep `PREFIX.$$` with the trap armed before `mkdir`. They are on the audit's pinned list with the zero-abandon-window and pid-reclaim reasons written out, rather than converted.
- The TERM/INT/HUP contracts and #105's reclaim asserts are intact and green in both keep-on-failure suites.

**Claims corrected:** the count pin now states what it guarantees (extra-line and stale-entry directions) and admits that substitution inside a listed file is invisible to it; `live-eval-isolation.sh`'s reason no longer says its mint is checked — the cleanup guard beside the delete is what makes it safe; `report-tmp-debris.sh`'s reason no longer calls its heredoc quoted (unquoted, with the pinned line escaping its own expansion). The `rm<TAB>` and backslash-newline blind spots are documented next to find/xargs in both the audit and docs/testing.md. `install.sh` is now in the audit's scan and had its mint checked and empty-guarded. The dead `tmp_base` in `run-module-tests.sh` is gone, and the suite-count comment says 21, not 28.

**Counts corrected:** 26 pinned lines across 20 files, each carrying a reason, against 29 unstated on main — not the "15 across 11, down from 51" the old body claimed, which was measured against the superseded pre-#96 base.

**Verification, macOS:** three script audits green; dev-tooling wave 21/21 twice; `go test -count=1 -short .` at repo root green; `go test -short ./cmd/...` green; kill/reclaim contracts green (test-coverage-floor 41 checks, coverage-union 24, fuzz-coverage-global 150, scratch-lib 32, 0 failed); shellcheck clean of new findings; mutation check — a probe reintroducing the kata 5hs2 shape fails both audits with the exact lines, an extra delete in a listed file fails the count one way, removing a listed file's only delete fails it the other.

obra:
Re-review findings, disposition after the merge landed mid-rework: **F2** (report-tmp-debris heredoc reason) and **F3** (dead tmp_base pair) were already fixed on main by 3a0641b92. **F1** (e2e-ratelimited-provider --stop deletes any handed path — marker gate added, decoy-verified plus a live start→stop round-trip) and **F4** (rule 1's 'only shrinks' / 'in scripts/' overclaims) land in #115. Body nits fixed above: diffstat corrected to +622/−2,798 at e0ff41e8a, and the shellcheck line now acknowledges the SC2154 false positive in scratch-lib-selftest.sh:38 (printf -v assignment shellcheck cannot see).

PR #116, opened 2026-08-18
Test-reliability kata batch: flake root causes, hermetic audits, guard hardening (8 katas)

obra:
Eight katas from the test-reliability cluster, worked as one batch in a single worktree by a fleet of eight subagents (four haiku, four sonnet), with central review, mutation verification, and gating by the controller session. Seven produced changes; one closed as already-fixed with mutation proof.

## Flake root causes (the gate's false reds)

- **ebnq** — `wave_test.go`'s fixed 5s post-exit grace failed the controller's own baseline gate on the untouched tree. The grace now scales to `t.Deadline()` less a 15s reporting reserve (so a genuinely absent writer still reports the clean one-line error instead of racing the runner's panic), floored at the old 5s when no deadline exists. A live-but-slow descendant gets every second the test can spare.
- **fazd** — the tmux TUI e2e keystroke losses trace to bubbletea v1.3.10 painting its first frame before attaching the stdin reader; on a loaded host, keys sent into that window go unobserved. The harness constructors now block on an input-readiness probe: resend the dashboard's `r` refresh each tick until the test hub's ThreadList counter moves — an effect only a consumed keystroke can produce, deliberately out-of-band of the screen-capture path that races the same window. Caveat recorded on the kata: locally the race delays bytes rather than dropping them, so the CI loss mechanism itself was not reproduced; the probe closes the window under either mechanism.
- **xykx** (no diff) — `TestDetachCommandSurvivesCleanup` was already fixed by `102b7906d` earlier the same day; both cited occurrences predate it by ancestry (the second only looked post-fix because its branch forked before the fix reached main). Proven by mutation: reverting that one line reproduces the exact failure 10/300 under load ~16; the current tree ran 650/650 green. Reading 2 (detach loses the child) ruled out by code: the detached PID is never enrolled in `runningPIDs`, so Cleanup's sweep has nothing to signal.

## Hermeticity and order-independence

- **6zst** — the scenarioneedle audit walked the working tree, so local debris flipped its verdict per checkout. The haystack now comes from `git ls-files`; only tracked paths can enter. Premise correction recorded on the kata: the three named suspects were not the mechanism and the original contaminating file was never identified, but the class is proven — restoring the walk plus one planted untracked needle file reproduces the exact failure.
- **ycet** — the two remaining vitest shuffle-order leaks: `oauthDialogs.test.tsx` never restored its `window.open` spies (same spy object, accumulated history), and — the kata's unresolved remainder — `resetThreadsStoreForTests` rebuilt state by shallow merge, which copies a spied action method forward past `vi.restoreAllMocks()`'s reach; it now rebuilds from Zustand's pristine `getInitialState()` with `replace: true`. Both reproduce on recorded seeds with the fix reverted; two full 339-file shuffled passes green with it in place.

## Guard and tooling hardening

- **5zs8** — browser guards: a rejecting `cleanup()` (escaped Chrome helper, ~1 in 3 runs) can no longer skip the generated-dir removal or turn a zero-failure run into exit 2 — extended to overflowguard/spawnguard, which shared the same bare `await cleanup()`; SIGINT no longer leaks the repo-side dir; and every CDP round trip is bounded by a timeout that names the call (probed: the former hang site rejects `timeout calling Page.enable after 30000ms` instead of hanging CI to the job limit). Note: layoutguard cannot currently pass inside a symlinked-node_modules worktree (Vite won't serve the fonts through the symlink) — verified pre-existing by running the pre-change code, which fails identically; CI's real install is unaffected.
- **tbqd** — the three variable-fed `rm -rf` sites in Makefile recipes are now under a count-pinned audit (`TestNoMakefileRecipeFeedsVariableToRecursiveDelete`), the same regime the scripts/ audit enforces; an injected offender fails the audit by name. Recipe semantics untouched.
- **h61c** — fakellm names sessions from the session-affinity header when present (matching the production adapter's header list exactly), effective from round 1, with the tool-call-id fallback preserved verbatim; the e2e turn-controls fixture enables the quirk (selftest 61/61).

## Verification

Every test added or changed carries executed mutation evidence (the exact source change that makes it fail, actually run). Full `merge-approval-gate` green on the combined tree; the baseline gate on the untouched tree had failed on ebnq's flake, so the before/after is direct. Per-kata reports live in the worktree's `.superpowers/` (git-ignored).

https://claude.ai/code/session_016XS3npTy4h1cE34ZxjGfir


obra:
Review fixes pushed. origin/main merged in first (clean). Per-fix evidence below; every claim was re-measured in this worktree rather than carried over from the review.

## Blocker 1 — kata 5zs8, `withTimeout` kept the guard alive for 30s (43fe4a67f)

**End to end, same machine, cloned (not symlinked) node_modules so the guards actually run:**

| `make test-web-browser` | wall clock |
| --- | --- |
| PR tip as reviewed (54e987d1c) | **1:35.92** |
| with this fix | **6.606s** |
| origin/main (no `withTimeout` yet) | 6.621s |

Three guards x ~30s of dead timer. The fix puts the target back at main's number.

**Isolated red/green:** a script whose entire body is one `evaluate()` against a fake `send` took **30.044s** before, **0.029s** after. Confirms the diagnosis: `withTimeout` armed a 30s timer, raced it, and abandoned the loser; the guards end by setting `process.exitCode` rather than calling `process.exit()`, so the process cannot leave until the loop drains.

`clearTimeout` now runs in a `.finally` on the race. Two adjacent leaks in `navigateTo` went with it:

- the load listener is removed in a `finally`, not only from inside the load handler (on the timeout path the handler never runs, and it stayed on a socket the guards reuse for every later case);
- **beyond what the review asked**, a *failed* `Page.navigate` left the load wait pending with nobody awaiting it. It held the loop open for its own 30s and then rejected unhandled — which Node turns into a crash 30 seconds after the real error. `navigateTo` now settles that wait in its catch. Shout if you'd rather have that separated out; it is the same defect class on the other branch, and it made the timeout path testable without a 30-second test.

**New test file** — `browserGuardCdp.mjs` had none. `scripts/browserGuardCdp.test.mjs`, 5 tests, **65ms**, no browser (fake `send` + fake socket; mock timers stand in for the 30s). Mutation-verified, each reverted in isolation:

| mutation | result |
| --- | --- |
| drop `clearTimeout` from `withTimeout` | 3 fail — incl. `a guard-shaped script exits as soon as its work is done (30030.881709ms)` |
| drop `removeEventListener` from the load wait | 3 fail |
| drop the failed-navigate settle | 1 fail |

**Wiring:** `npm test` named the one node:test file explicitly, so a second one would have been silently handed to vitest. Both halves now match on shape — `scripts/*.test.mjs` is node:test, and `scripts/layoutguard/viewport.test.mjs` (a real vitest file) stays with vitest. Verified by collection: 339 files, no `scripts/*.test.mjs`, no `node_modules`. `npm test` exits 0.

## The cleanup-failure downgrade — reverted per your ruling (fb6ea9349)

All three guards go red again when `cleanup()` rejects. layoutguard keeps the generated-dir half: the fixtures come out on every path, including the rejecting one, via `finally` instead of `catch`. Each site's comment names **#119** and says why red is the intended signal there.

No test added for "a rejecting cleanup exits nonzero", and I want to be explicit rather than quiet about it: `main()` in each `run.mjs` is a script body, not an exported function, so there is nothing to drive without a real Chrome. Making it testable means extracting a shared guard-entrypoint helper across three files — a refactor I did not think a revert justified. What *is* pinned today, in `browserGuardProcess.test.mjs`: that `cleanup()` rejects on an escaped helper, and that the signal path reports before exiting 143. Say the word if you want the extraction.

## Blocker 2 — kata ebnq, the deadline extension was untested (94d56a688)

Confirmed vacuous: deadline 10s out against a 15s reserve → negative time left → grace stayed at the 5s floor → the 2s writer beat the floor either way.

- The arithmetic is now `graceFor(deadline, now)`, pure and table-tested across six cases: no deadline, inside the reserve, expired, one second past the reserve, extends, capped.
- **The extension is capped at 60s.** Uncapped it is 9m45s of the gate's 10m budget for one absent writer (you measured 1m45s under `-timeout 2m`). A live descendant writes in milliseconds; past a minute more waiting buys nothing and charges every other test in the binary for the diagnosis.
- The integration test is repointed at the wiring — absent writer, a deadline granting 7s of grace, must wait past the 5s floor. It costs its grace in wall clock, so the deadline is the smallest one that makes the extension visible rather than a realistic one.

Mutation-verified:

| mutation | result |
| --- | --- |
| extension deleted (`graceFor` returns the floor) | 3 failures, incl. `graceFor(deadline 45s from now) = 5s, want 30s` and `waited 5.000391959s: the deadline's 7s of grace was ignored` |
| cap deleted | `graceFor(deadline 9m15s from now) = 9m0s, want 1m0s` |
| wiring cut (`readFifoLineOrExit` uses the floor) | `waited 5.0011165s: the deadline's 7s of grace was ignored` |

## Blocker 3 — kata tbqd, the duplicate weaker predicate (b1b8405db)

`variableFedRecursiveDelete` deleted; the audit calls `recursiveDeleteTakingVariable` from `scriptmktemp_audit_test.go` (same package). Measured side by side on ten shapes — 8 offenders, 2 safe:

| shape | old | new |
| --- | --- | --- |
| `-rf "$$dir"` | caught | caught |
| `-fr "$$dir"` | **missed** | caught |
| `-Rf "$$dir"` | **missed** | caught |
| `--recursive --force "$$dir"` | **missed** | caught |
| `... ; <second delete> "$$dir"` | **missed** | caught |
| `echo hi && ... "$$dir"` | caught | caught |
| `"$(SERF_DIST_BIN_DIR)" "$(SERF_DIST_ARCHIVE)"` | caught | caught |
| `trap '... "$$dir"' EXIT` | caught | caught |
| literal path (safe) | rejected | rejected |
| non-recursive `-f "$$dir"` (safe) | rejected | rejected |

4 of 8 vs 8 of 8. The false `rm -fr` claim in the old doc comment went with the function. What stays local is the *reporting* — an offender is named by recipe and variable, the unit a human blesses.

Dead `seenSiteMap` removed, and the pin compares the blessed set entry by entry instead of totals (a total is satisfied by any set of the right size, so one recipe growing a second delete would cover for another disappearing). **Both directions verified:** an injected offender in a flag order the old copy could not see fails by name (`Makefile:114:test-web-browser: ... (variable: $$scratch)`); deleting a blessed delete reports the two entries it stranded by name.

## Should-fixes

- **h61c** (c4fa8486d): the package doc said tool-call-id naming and said the fixture does not set `send_session_affinity_headers` — the PR changed both. Rewritten, plus `sessionState.name`'s own comment, which said the same stale thing. Verified while writing it: the header value is the daemon's session id (`agent/session.go` sets `req.SessionID = s.id`; openaicompat sends it as all three affinity headers), and `sessions/<id>.api.jsonl` is the file it cross-references.
- **fazd** (0c259b847): `hub.WaitForTreeRequests(a.t, 1)` before the baseline, closing the window where the startup fetch satisfies the probe with no keystroke consumed.
- **6zst** (eecfe1503): git's stderr attached via `errors.As` on `ExitError`. Measured against a non-repo directory — `exit status 128` alone becomes `exit status 128: fatal: not a git repository (or any of the parent directories): .git`.

## Verification

| command | result |
| --- | --- |
| `node --test scripts/browserGuardCdp.test.mjs` | 5/5 pass, 65ms |
| `npm test` (frontend, vitest + node:test) | exit 0 |
| `go test -race ./cmd/serf-test-dev-tooling/` | ok, 28.879s |
| `GOMAXPROCS=1 go test -race -count=20 ./cmd/serf-test-dev-tooling/` | ok, 304.947s |
| `go test -count=1 -short .` (root audits) | ok, 29.796s |
| `go test -race -run TestTUITmuxE2E ./cmd/serf-tui/` | ok, 5.853s |
| `go test -race ./test/e2e/fakellm/...` | ok |
| `make test-web-browser` | PASS x3, 6.606s (see above) |

One environment note for whoever runs this next: the browser guards cannot pass in a worktree whose `node_modules` is a SYMLINK to the shared install — every case dies with "a web font this page requested failed to load", because Vite will not serve the faces through it. Verified pre-existing, not from this branch: unmodified origin/main fails identically in a symlinked worktree. An APFS clone (`cp -Rc`, 0.7s for 204M) makes the guards runnable, which is how the numbers above were taken.

PR #103, opened 2026-08-18
spec: dev tooling in Go — three gate runners, one small binary, no janitor (v2)

obra:
v2 of the dev-tooling-in-Go spec, after the simplification pass Jesse asked for ("think through the spec about what we really need for the project and how to make this as clean and simple as possible"). v1 remains in the file's git history as the contract-extraction reference.

**The program shrank from four phases / 18 ports / ~8,700 Go to:**

- **1a** — port `agent-test-shards.sh` → `serf-dev agent-shards` (kills the Python-bin-packer-in-bash), with Go-side pid-suffix scratch + startup reclaim; **delete `reclaim-test-debris.sh`** and its selftest. The janitor's last job goes to its owner.
- **1b** — port `run-module-lint.sh` → `serf-dev module-lint`; its 868-line stub-toolchain selftest and the audit's two pinned fixture weapons die with it.
- **1c** — 8 lines of shell: `test-cost.sh` stops leaking its scratch.
- **2** — port `run-module-tests.sh` → `serf-dev module-tests` (calls the shard packer in-process); gate-surface drift tripwire.

**What got cut, and why:** merged #105 proved owner-side reclamation works in shell (pid-suffix + trap-first + covscratch-lib), which deletes v1's entire P1 coverage-port phase and the `ratchet`/`covprofile`/`gatesurface` packages that served it. The fuzz/coverage families are hand-run, working, and self-reclaiming — v1's P3/P4 become a **port-on-touch** policy. Library packages are earned, not declared: only `scratch`, `report`, `procgroup` remain, each with two consumers on day one.

**Flagged for Jesse, not acted on:** the fuzz-family selftests (~1,766 lines pinning hand-run tools through fake toolchains) fail the same standard deploy-hub-selftest was deleted under; shrinking them needs a kill-list decision.

Wave: 21 → 17 suites. Net: ~2,200 Go replaces ~3,100 shell; fake toolchains in the gate path go to zero.

PR #118, opened 2026-08-18
feat(tooling): port agent-test-shards to serf-dev and retire the janitor

obra:
Step 1a of the dev-tooling spec (PR #103, v2): the shard runner moves to Go, and the janitor retires because its last debris class now cleans up after itself.

## What this does

- **`cmd/serf-dev`** is born: dispatch main, one file per subcommand, `agent-shards` first. `make test`'s AGENT_SHARDS branch (run-module-tests.sh) now calls `go run ./cmd/serf-dev agent-shards`.
- **`internal/devtool/scratch`**: the Go twin of covscratch-lib.sh — `Acquire` mints `<TMPDIR>/<prefix>.<pid>` (loud on collision, refuses TMPDIR resolving to `/` or `$HOME`), `Release` removes only what Acquire minted (no path arguments anywhere in the API), `KeepOnFailure` retains and prints the retained-logs pointer, and `ReclaimOwn` removes dead-pid siblings under the caller's own prefix with the shell reclaimer's rules and two safe tightenings (dirs only, direct children only, exact prefix, symlinks never matched or followed, live pids including our own always kept, failed removals reported and never fatal). The tightenings both keep MORE than the shell did: an EPERM from signal 0 counts as alive here, where bash's `kill -0` reports failure and the shell would reclaim; and a pid suffix above `1<<30` is treated as "not a pid" and kept, where the shell asked the kernel and reclaimed on its refusal. The pid-liveness check lives in `pid_unix.go`/`pid_other.go` (the convention `cmd/serf-test-dev-tooling` already uses) so the library builds under `GOOS=windows` — every later port of the tooling imports it.
- **Bootstrapping scope**: `go run ./cmd/serf-dev` compiles three packages and nothing else — `cmd/serf-dev`, `internal/devtool/scratch`, `envvars` (`go list -deps` confirms) — so the tool the test wave depends on cannot be broken by the code it tests. The precedent is `cmd/serf-test-dev-tooling`, run the same way from Makefile:249.
- **Deleted**: `scripts/agent-test-shards.sh` (306), `scripts/agent-test-shards-selftest.sh` (280, fake-toolchain), `scripts/reclaim-test-debris.sh` (208), `scripts/reclaim-test-debris-selftest.sh` (140). Wave drops 21 → 19 suites. `recursiveDeleteAllowedLines` loses the janitor's two entries and the scenario /tmp-path allowlist loses its row (all watched failing stale first — the bidirectional count-pin works).

## The janitor's last shift

Ran `reclaim-test-debris.sh --dry-run` then `--yes` before deleting it: at that point, zero stale shard dirs under `$TMPDIR` (and none under literal `/tmp`), and `/tmp/serf-gocache-k3` already absent.

Correction: that sweep was not the last one. Review re-ran the janitor's final sweep on the same machine days later, and it found and removed one legacy shard directory (~41MB) that had accumulated in between. After that sweep there is nothing left to hand over — the "zero stale dirs" claim is true as of the sweep at review time, not as of the original one.

## Parity table (contract source: the script + its selftest's assertions)

| Contract | Status |
|---|---|
| Env: `AGENT_SHARD_COUNT`(4) / `AGENT_SHARD_PARALLEL`(3) / `AGENT_SHARD_SKIP` / `AGENT_SHARD_NO_SURVEY` / `AGENT_SHARD_RESURVEY` / `AGENT_SHARD_CACHE_DIR` | preserved (`AGENT_SHARD_SKIP` reaches the survey only, in both spellings — see warts) |
| go-test flag pass-through table: `-short`/`-race`/`-v`/`-count=N` translated, everything else dropped | preserved (wart, see below) |
| Survey cached by test-set identity; `NO_SURVEY` equal-weights fallback; red survey fails loudly with excerpt + `full log:` pointer | preserved |
| Partition proven a bijection before running; empty-bin refusal names `AGENT_SHARD_COUNT` | preserved |
| One `PASS  agent:N <time> (M tests)` / `FAIL  agent:N` line per shard, failing logs replayed under `----- agent:N -----`, `full logs:` pointer | preserved (replay is by verdict, not by log marker — see "Fixed in review") |
| Logs deleted only on normal green completion | preserved |
| Signal exits 129/130/143, `interrupted by SIG…` explanation, shard process groups TERMed, logs retained | preserved in the binary (pinned by real SIGTERM tests), **but the production call site does not carry the code out**: run-module-tests.sh:310 invokes the runner through `go run`, which collapses any child exit code to 1 and prints the real one as an `exit status N` line on stderr. Nothing downstream reads more than zero-vs-nonzero, so this costs nothing today; the caveat is now a comment at the call site |
| Exit nonzero on any shard failure or partition discrepancy | preserved |
| A second HUP/INT/TERM ends the run even with a shard wedged | preserved (the script cleared its traps in the first handler; see "Fixed in review") |

**Sanctioned port deltas** (spec appendix): scratch is now `agent-test-shards.<pid>` with startup reclaim — a `kill -9` mid-run leaves a directory the NEXT run removes, pinned by real SIGKILL tests at both the library and whole-binary level; the `.heartbeat` protocol retired with the janitor that read it.

## Fixed in review

1. **Second-signal deafness (regression against the script).** `signal.Notify` on a 1-buffered channel plus a handler that read exactly one signal left the process deaf to every later one: with `Notify` registered, the default disposition is disabled, so a shard that ignores TERM held the runner forever and only SIGKILL ended it. The relay now loops; the second signal exits `128+signal` immediately and names the logs it abandons, which is where the script's cleared traps ended up too. (`signal.Stop` + re-raise is the literal translation and was tried first; probing the real binary showed the re-raised signal does not take its default action before the process continues, so the direct exit is what actually ends the run.) Pinned by a real-binary test with an ignore-TERM fixture shard — red: the runner had to be SIGKILLed after 20s.
2. **Replay selected by marker (kata-mjzx anti-pattern).** Failing-shard logs were chosen by grepping for `^(FAIL|--- FAIL|panic:)` — the exact selection-by-marker bug run-module-tests.sh:407 documents as fixed. A shard that fails with no marker (build error, `os.Exit`, OOM kill) got an empty replay block; a green shard that printed a line starting with `FAIL` got replayed instead. Replay now follows the verdict, an empty or missing log says so out loud, and both directions are pinned in one test.
3. **`AGENT_SHARD_SKIP` documentation.** The interface comment claimed "regex of tests to skip in every shard". It is appended to the survey's `-test.skip` and nothing else, so it works by leaving a test out of the cost table — under `AGENT_SHARD_NO_SURVEY` or a primed cache it does nothing at all. The script behaved identically, so parity-first says document rather than fix; the honest description is pinned by a test that fails if the behavior changes.

## Warts preserved (each a follow-up candidate, none smuggled fixed)

1. `-race` translates to `-test.race` for the shard binary while the build never gets `-race` — passing `-race` was and remains broken (the `-race` gate uses `AGENT_SHARDS=0` and never hits this).
2. `-run`/`-timeout`/every other flag silently dropped by the translation table.
3. Failing runs print the retained-logs pointer twice (stdout in the replay block, stderr at exit).
4. `AGENT_SHARD_SKIP` reaches the survey only (see "Fixed in review" #3 — the wart is the script's, the doc lie was the port's).
5. The survey excerpt still greps `^(--- FAIL|panic:)`; the shard replay no longer greps at all, so the two spellings the script carried are down to one.

**Deliberate deviations, disclosed**: diagnostic prefixes say `agent-shards:` (the tool's name) instead of `agent-test-shards:`; contract lines (`PASS  agent:N`) are unchanged. Invalid `AGENT_SHARD_COUNT`/`PARALLEL` now fail with the variable's name (same exit 1; the script died with "partitioning failed" via the embedded Python). The survey cache re-keys once (Go's byte sort vs shell's locale sort), costing one survey per machine. FAIL lines for TERMed shards may print before the 143 exit. The Setpgid + TERM-the-process-group pair duplicates `cmd/serf-test-dev-tooling/wave_unix.go`'s two lines on purpose: each tool is checked for parity against its own script, and two lines of overlap buy less than a shared home would cost — the third caller is the one to build it for.

## Found and fixed during the port

The first real run showed every shard reporting the slowest shard's wall time (in-order `Wait` kept early finishers' clocks running). Fixed with per-shard wait goroutines; pinned by a survey-isolated slow test whose sibling must report a strictly shorter time (watched failing first).

## Verification

- All new Go tests green with `-race`; every behavior watched red first (stub sentinel exits, stale audit entries, the timing bug, and all three review fixes).
- `golangci-lint run` clean on `cmd/serf-dev` + `internal/devtool`; all three script audits green.
- Dev-tooling wave 19/19, twice.
- `GOOS=windows go build ./internal/devtool/...` passes (it did not before the pid split).
- Real workload: `MODULES="agent" WEB=0 scripts/run-module-tests.sh -short -count=1` — the real ported path, green.
- `make merge-approval-gate` in a clean /tmp worktree, with home canary 32 before and after — run at authoring time, not re-run after the review fixes (those were verified with the targeted commands above plus the root audits and the dev-tooling wave).

No PATH fakes anywhere: tests compile and run a real fixture module with the real toolchain, and signals hit the real binary.


obra:
Adversarial review verdict: APPROVE (CI green). Recording its non-blocking disclosures for the merge record:

- **Undisclosed preserved wart**: `AGENT_SHARD_SKIP` filters only the survey pass, is never applied to shard runs, and is silently ignored on a survey-cache hit — identical to the shell original; parity holds.
- **`go run` flattens exit codes at the call site**: child exits 2/130/143 all surface as rc=1 from `go run`. The binary itself honors 1/2/129/130/143 (pinned by the e2e SIGTERM test) and run-module-tests.sh only branches on nonzero, so the gate is unaffected; the full exit-code contract is observable from a built binary.
- **Reclaim corners stricter-safe than 'verbatim'**: an unparseable/overflowing pid suffix is kept (shell would delete), and EPERM counts as alive (shell would delete another user's live scratch). Both deviate toward keeping — the right failure direction.
- **Review-driven follow-up in flight**: exported `ReclaimOwn` will mirror `Acquire`'s empty/"."/"/" prefix refusal (hardening in a hazard-class package; no caller can reach the gap today).

Side-by-side parity harness (old script vs new binary, same fixtures): green/failing/red-survey/cache-hit paths byte-identical per stream; identical partitions and cache keys; real workload 4 shards / 3,251 tests / rc=0 / zero TMPDIR leftovers.

obra:
Review fixes pushed (`5c67470e4`, `c35822116`, `d0182aec9`, `fa6434d6a`, on top of a clean merge of `origin/main` at `e2b35a39b`). Per-fix evidence below; every behavior change was watched red first against a real binary and the real fixture module.

## F2 — second-signal deafness (blocking)

`signal.Notify` on a 1-buffered channel plus a handler that read exactly one signal: with `Notify` registered the default disposition is off, so every later HUP/INT/TERM went nowhere. New test `TestAgentShardsSecondSignalEndsAWedgedRun` arms an ignore-TERM fixture shard (it swallows TERM and announces each one so the second signal is never mistaken for the first), sends TERM twice, and bounds the wait.

**Red** (the pre-fix one-shot handler, re-confirmed against the final test):

```
--- FAIL: TestAgentShardsSecondSignalEndsAWedgedRun (20.90s)
    agentshards_e2e_test.go:522: second SIGTERM did not end the runner; it had to be SIGKILLed after 20s (wait: signal: killed)
```

**Green** (after): the relay loops. The first signal behaves exactly as before; the second exits `128+signal` immediately and names the logs it is abandoning (`os.Exit` runs no deferred cleanup, so the scratch directory stays on disk). Stderr carries both `interrupted by SIGTERM` and `SIGTERM again — abandoning the running shards; logs: …`, and the run ends in well under a second:

```
go test -race -count=1 -run 'TestAgentShardsSecondSignalEndsAWedgedRun|TestAgentShardsSIGTERMExits143RetainsLogsReapsChildren' ./cmd/serf-dev/
ok  	primeradiant.com/serf/cmd/serf-dev	2.806s
```

Channel buffer went 1 → 2 so the second signal is queued while the first is still being handled.

One note on the mechanism, because I tried the suggested spelling first and it does not hold up. `signal.Stop` + re-raise is the direct translation of the script's `trap - HUP INT TERM`, and it was what `5c67470e4` shipped, with `os.Exit(128+sig)` behind it as a fallback. Probing the real binary showed the fallback doing all the work: the re-raised SIGTERM does not take its default action before the process continues, so the observed exit was always `exit status 143`, never `signal: terminated`. `fa6434d6a` drops the half that never fires — the second signal exits 128+signal directly, which is the code the script exited on anyway — and the test asserts exactly 143 rather than "signal or 143". Practical consequence for callers: a wedged runner's second Ctrl-C gives a shell `$?` of 143, not a `WIFSIGNALED` status.

## F3 — marker-matching replay (kata mjzx)

**Red** (current branch), one test pinning both directions — the failing shard has no marker (`os.Exit(3)` after a println), the green shard prints a line starting with `FAIL`:

```
--- FAIL: TestAgentShardsReplayIsByVerdictNotMarker (0.31s)
    agentshards_e2e_test.go:266: replay block covered shards [0], want exactly the FAIL verdicts [1]:
        agent-shards: 2 shards, -parallel 1 each
        PASS  agent:0     0.05s (3 tests)
        FAIL  agent:1

        === failing shard output ===
        ----- agent:0 -----
        FAIL: fixture-gamma is green and only looks red
        PASS
```

That is exactly the reviewer's reproduction: the FAIL verdict gets nothing, and a PASS verdict's log gets replayed.

**Green** (after), same scenario through the real built binary:

```
agent-shards: 2 shards, -parallel 1 each
PASS  agent:0     0.05s (3 tests)
FAIL  agent:1

=== failing shard output ===
----- agent:1 -----
fixture-beta exiting hard as instructed by SHARD_FIXTURE_EXIT

full logs: …/agent-test-shards.7647
```

Replay now walks the shards that returned a nonzero verdict — `r.err != nil` was already in hand — and an empty or missing log prints `(no output captured: <path> is empty or missing)` rather than nothing. A shard that never launched gets no banner at all (its error is already on stderr, and an empty banner is the same "verdict with nothing behind it" smell). `shardRedLine` and `fileMatches` are gone; `surveyRedLine` stays, because the survey is one pass with no per-shard verdict to sort by. The existing `TestAgentShardsFailingShardRetainsEvidence` gained the same replayed-equals-verdicts assertion.

## F4 — `AGENT_SHARD_SKIP` doc lie

Confirmed: the regex is appended to `surveyArgs` and nowhere else, so it works only by leaving a test out of the cost table. Chose **honest-doc**, per parity-first — the script had the identical behavior, and threading the flag into the shard invocations would also have to enter the survey cache key, which is a behavior change, not a doc fix. The interface comment now says the regex reaches the survey's `-test.skip` and only that, and that a cache hit or `AGENT_SHARD_NO_SURVEY` makes it a no-op. Stated in the commit message too.

`TestAgentShardsSkipOnlyReachesTheSurvey` pins both halves (skip + a red test, surveyed: green; the same thing unsurveyed: fails on the unskipped test). Mutation check, so the pin is not vacuous — threading `-test.skip` into the shard invocations:

```
--- FAIL: TestAgentShardsSkipOnlyReachesTheSurvey (0.79s)
    agentshards_e2e_test.go:327: unsurveyed run: rc = 0, want 1 — AGENT_SHARD_SKIP reaches the shards now, so the interface comment and this pin are both stale
```

Mutation reverted; the test names the doc as part of what has to change with it.

## F5 — signal-exit parity row

Verified the collapse rather than assuming it (`go run` over a child that exits 143):

```
exit status 143
go run rc=1
```

Parity row now says the binary preserves 129/130/143 but the production invocation does not carry them out, and a comment at run-module-tests.sh:310 says the same where it happens. Nothing downstream reads more than zero-vs-nonzero.

## F8 — "verbatim" claim

Corrected in both the PR body and `internal/devtool/scratch/scratch.go`: the reclaimer is the shell's rules **with two safe tightenings**, both keeping more than the shell did — EPERM from signal 0 counts as alive (bash's `kill -0` reports failure there and the shell would reclaim), and a pid suffix above `1<<30` is treated as "not a pid" and kept (the shell asked the kernel and reclaimed on refusal).

## F6 — bootstrapping scope

Added to the PR body, verified with `go list -deps ./cmd/serf-dev`: the only non-stdlib packages compiled are `cmd/serf-dev`, `internal/devtool/scratch`, `envvars`. Precedent named (`cmd/serf-test-dev-tooling`, Makefile:249).

## F9 — duplication note and the non-unix build

The Setpgid/TERM-the-group pair now carries a comment saying it duplicates `wave_unix.go`'s two lines deliberately (each tool is checked for parity against its own script; the third caller is the one to build a shared home for).

The build break was real:

```
$ GOOS=windows go build ./internal/devtool/...
internal/devtool/scratch/scratch.go:179:17: undefined: syscall.Kill
```

Split into `pid_unix.go` (`//go:build linux || darwin`) and `pid_other.go`, following `wave_unix.go`/`wave_other.go`. Off unix every owner counts as alive and nothing is reclaimed — the mistake this package must never make is deleting a live run's scratch. `GOOS=windows go build ./internal/devtool/...` now passes.

## Janitor claim

PR body corrected: the original sweep found zero stale dirs, but it was not the last sweep. The final sweep run at review time on the same machine removed one legacy shard directory (~41MB) that had accumulated since. "Nothing left to hand over" is true as of that sweep.

## Verification (on the pushed tree)

- `go test -race -count=1 ./cmd/serf-dev/... ./internal/devtool/...` — ok, ok
- `go test -count=1 -short .` (root audits) — ok
- `make test-dev-tooling` — 19/19 PASS
- `MODULES="agent" WEB=0 scripts/run-module-tests.sh -short -count=1` — `PASS  agent  11.60s`, rc 0
- `go run ./cmd/serf-dev agent-shards -short -count=1` against the live agent module — 4 shards, 3,251 tests, per-shard times differentiated:
  ```
  agent-shards: 4 shards, -parallel 3 each
  PASS  agent:0    33.85s (542 tests)
  PASS  agent:1    35.36s (543 tests)
  PASS  agent:2    39.48s (1623 tests)
  PASS  agent:3    33.85s (543 tests)
  ```
- `GOOS=windows go build ./internal/devtool/...` — passes (failed before)
- `golangci-lint run ./cmd/serf-dev/... ./internal/devtool/...` — 0 issues

Not merged, as asked.

obra:
Heads-up: this PR merged (a211df3c2, at head acaf80dd9) while the review fixes above were being written, so none of them are in `main`. They are cherry-picked onto current main in #120 — same four commits, and `main`'s copy of `agentshards.go` still has the marker-matching replay (`fileMatches(log, shardRedLine)`) and the one-shot signal handler until that lands.

The branch `feat/serf-dev-agent-shards` also still carries them (`fa6434d6a`), but a PR from it would replay the whole original diff because of the merge bases, hence the fresh branch.

Issue #94, opened 2026-08-18
Nothing bounds a single assistant response: a runaway model streams until the user interrupts


obra:
## Summary

A single streamed assistant response can contain an unbounded number of tool calls and never terminate. serf has four guards against runaway models and **all four are structurally blind to this**, because each watches either *between* responses or for *silence*, and a runaway is loud and never completes a round.

The trigger for the one incident we hit is fixed (`3ce922414`, serf no longer forces `tool_choice`). The hole is still open: any future runaway, from any cause, is still unbounded.

This issue records a prior-art survey across ~27 agents, frameworks, and provider APIs, because the conclusion is non-obvious: **the field's consensus mental model is "cap the loop," and that model has a hole exactly this shape.** Exactly one project has closed it.

Internal tracking: kata `d74b`.

## The incident

Session `0348pEjXIGWYWtGqeH1B8e`, glm-5.2-vision via an OpenAI-compatible gateway.

One streamed response contained **228 tool calls** built from **4 distinct argument blocks**: `manage_worktree` → `task_list` → `communicate`, byte-identical, 76 times over. It streamed **320.7 seconds** and produced nothing. The user interrupted it. Verified from the captured provider SSE, which carries all 228, so the repetition is upstream of serf's accumulator.

Replaying the frozen request reproduces it, and the **shape varies wildly between runs**:

| Run | Tool calls | Distinct signatures | Max repeat | Terminated? |
|---|---|---|---|---|
| captured | 228 | 5 | 76 | no |
| replay 1 | 373 | 19 | 22 | no |
| replay 2 | 83 | **48** | 12 | no |
| replay 3 | 237 | 7 | 231 | no |

That variance is the single most important design input. **The only invariant is non-termination.**

Correction to an earlier reading of this incident: all 75 `communicate` calls in the captured stream carry `end_turn: false`. They were status narration, not a completion signal. The model never signalled that it was done, so "stop the stream once the result tool completes" would not have helped here.

## Why serf's existing guards cannot see it

| Guard | Why it misses |
|---|---|
| `--max-rounds` | counts tool **rounds**; a response that never completes never finishes a round |
| `PhaseSilentStall` | fires only on **zero** content events; every emission calls `noteContent()`, so a runaway reads as healthy activity |
| `RetryStream` cap rule | applies only to a stream that **failed**; a runaway never fails, it just never ends |
[...]

PR #130, opened 2026-08-18
feat(agent): in-stream loop guard bounding a single assistant response

obra:
Kata: d74b (closed with this feature)

The hole (documented on the kata from a real 320s/228-call incident): all four existing guards watch something else — max-rounds counts finished rounds, PhaseSilentStall needs zero content, RetryStream needs a failure, the tool breaker needs results. A single unbounded response was invisible. The provider survey on the kata found no API-side bound exists anywhere; detection must be ours.

Implements the plan from the kata's revised research comment (gemini-cli LoopDetectionService shape, the field's only in-stream detector):
- **Cycle detection** (k=1..5 repeated 5x) as primary — catches the captured (A,B,C)x76 shape after ~15 calls mid-stream, where every consecutive-identical detector in the field is structurally blind.
- **Raw call ceiling as backstop** — catches the 83-call/48-signature shape (confirmed: the cycle detector legitimately never fires on it). **Ceiling = 50 is the one judgment call in this PR**: the kata's citation argues for hundreds, but that figure is another product's telemetry cap; the only documented legitimate shape is 18 distinct calls, 50 is ~2.7x that, and nothing in the suite needs more. A first trip is soft (see below), which is what makes a lower ceiling safe. `loopGuardRawCeiling` is the constant to revisit if field data shows legitimate >50-call responses.
- **Content chanting** for the zero-tool-call runaway (cline's "Let me" x2865), gemini-cli's own constants, with the load-bearing false-positive guard (between-text uniqueness) and code-fence resets.
- **Two-tier action**: first trip forces a legal stop boundary (synthetic well-formed finish tagged `loop_guard:*`; already-streamed calls still dispatch, a steering nudge names what repeated, work continues next round). Only a second consecutive trip hard-stops via a new non-retryable `llm.NewPermanentStreamError`. A trip while another parallel tool call is open defers until every call closes — never a cut mid-arguments.
- Rejected-by-Jesse items confirmed absent: no default max_tokens cap, no disable_parallel_tool_use, no wall clock.

New steering kind `stream-loop` regenerated into the TS protocol union; the frontend's exhaustive KIND_LABELS net caught it and got its label (the net working as designed).

Verification: 22-test suite over the five documented stream shapes (two must-NOT-trip cases included); 13 production mutations run for real, incl. bracketing the ceiling from both sides; three real bugs found by the tests during development (nudge dropped on zero-tool-call rounds; six pre-existing tests using genuinely chant-shaped filler; two self-caught false-green tests). Full agent+llm suites, central make lint + make test (incl. web typecheck) green on the integration branch.

https://claude.ai/code/session_017f6aYf3cJgkcg8PNAg62wX

### DIFF
--- a/agent/events/payloads.go
+++ b/agent/events/payloads.go
@@ -303,14 +303,20 @@ const SteeringSourceUser = "user"
 // carried to the UI so a label is ground truth rather than a guess at the
 // message's prose. Absent kind means "unknown" and the UI claims nothing.
 const (
-	SteeringKindInterrupted       = "interrupted"
-	SteeringKindAgentMessage      = "agent-message"
-	SteeringKindHookContext       = "hook-context"
-	SteeringKindPrecompactHook    = "precompact-hook"
-	SteeringKindCompactNudge      = "compact-nudge"
-	SteeringKindImageDescription  = "image-description"
-	SteeringKindNoToolCalls       = "no-tool-calls"
-	SteeringKindLoopDetected      = "loop-detected"
+	SteeringKindInterrupted      = "interrupted"
+	SteeringKindAgentMessage     = "agent-message"
+	SteeringKindHookContext      = "hook-context"
+	SteeringKindPrecompactHook   = "precompact-hook"
+	SteeringKindCompactNudge     = "compact-nudge"
+	SteeringKindImageDescription = "image-description"
+	SteeringKindNoToolCalls      = "no-tool-calls"
+	SteeringKindLoopDetected     = "loop-detected"
+	// SteeringKindStreamLoop marks the kata d74b in-stream loop guard's
+	// nudge, distinct from SteeringKindLoopDetected (the older, cross-round
+	// detector on dispatched tool results): this one fires on a repeating
+	// cycle, raw call ceiling, or content chant caught WITHIN a single
+	// streaming response, before any of its calls dispatch.
+	SteeringKindStreamLoop        = "stream-loop"
 	SteeringKindTasksDone         = "tasks-done"
 	SteeringKindTaskNudge         = "task-nudge"
 	SteeringKindTaskInactive      = "task-inactive"
@@ -336,6 +342,7 @@ var AllSteeringKinds = []string{
 	SteeringKindImageDescription,
 	SteeringKindNoToolCalls,
 	SteeringKindLoopDetected,
+	SteeringKindStreamLoop,
 	SteeringKindTasksDone,
 	SteeringKindTaskNudge,
 	SteeringKindTaskInactive,
--- a/agent/events/payloads_test.go
+++ b/agent/events/payloads_test.go
@@ -101,6 +101,7 @@ func TestSteeringKindConstants(t *testing.T) {
 		"image-description":  SteeringKindImageDescription,
 		"no-tool-calls":      SteeringKindNoToolCalls,
 		"loop-detected":      SteeringKindLoopDetected,
+		"stream-loop":        SteeringKindStreamLoop,
 		"tasks-done":         SteeringKindTasksDone,
 		"task-nudge":         SteeringKindTaskNudge,
 		"task-inactive":      SteeringKindTaskInactive,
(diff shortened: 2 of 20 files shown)


obra:
## Adversarial review: REJECT (chant detector; tool-call side is sound)

Two showstoppers, both demonstrated by running the branch's code, plus three smaller defects. The cycle detector + call ceiling half is fine — no false positive found against it. If the chant detector were dropped or reworked, the rest is close to mergeable.

### 1. Performance: ~13.7s of synchronous CPU per 40KB of ordinary output
`checkChant` runs on every text/reasoning delta, rescans all ~4,950 positions of the 5,000-rune window, and allocates a 50-rune string per position (`session_stream_loop_chant.go:81`). Measured through the real `consumeModelStream`: a 40KB response at 4 runes/delta takes **13.68s** on the branch vs **7.5ms** with `checkChant` stubbed to nil (~1,800x). It runs on the turn goroutine, so past the 128-event `ChanStream` buffer it back-pressures the provider's SSE read — the exact failure class the `PhaseSilentStall`/`ErrSSEReadTimeout` machinery exists for. Cost is linear in response length.

Fix shape: O(1)-amortized rolling-hash index over the trailing window, no per-position allocation.

### 2. False positives on commonplace output
Trip floor is 59 consecutive identical characters. Verified trips: 80-dash horizontal rule, 70 spaces of column padding, box-drawing rule, unfenced base64 zero-run, `INSERT INTO t VALUES (?, ?, …)`, 14 identical markdown table rows, 14 identical checklist items. End-to-end: a review document containing one 80-dash separator was cut at 132/273 runes, `Finish.Raw="loop_guard:chant"`, both findings after the rule discarded. The branch's own history corroborates the rate: commit a1903b9cd rewrote six pre-existing tests whose filler tripped the detector (`nonChantingFiller`).

The code-fence exemption doesn't survive real streams: `strings.Contains(delta, "```")` toggles `inFence` once per delta regardless of marker count. Fence split across deltas → guard never engages (verified). Whole fenced block in one delta (two markers, one toggle) → `inFence` latches true and chant detection is dead for the rest of the response (verified; Google's adapter emits one delta per `part.text`, so Gemini hits this routinely). Fixture (e) certifies the one arrangement that works.

Negative-fixture gap: only two exist. None for separators, padding, tables, repeated list items, base64, or split fences.

### Smaller defects (would be MERGE AFTER FIXES on their own)
- The tier-2 permanent error is `modelFallbackEligible` (verified) — the fallback chain re-runs the same request on every configured fallback model, exactly what `NewPermanentStreamError`'s doc says it exists to prevent. Untested.
- Streak never resets after a hard stop (measured 1→2→3 across trips): after one hard stop, every later trip hard-stops instantly with no tier-1 nudge until some response completes cleanly.
- Streak/pending-nudge are session-scoped and not reset at turn start, so state leaks across user turns; a stale nudge can land out of context.
- No test joins the halves: nothing exercises trip → dispatch → nudge → next round, and nothing proves a session survives a tier-2 hard stop.

### What held up
Mutation checks: 11/11 red (thresholds, fence reset, deferral, streak, retryable flag, nudge delivery ×2) — the suite is load-bearing; the gap is coverage, not quality. Tier-1 forces a well-formed synthetic Finish at a legal boundary; parallel-call deferral prevents mid-arguments cuts; abort cancels the HTTP read (no leak); `types.gen.ts` is genuinely generated; the SteeringItem label change is the exhaustive-record mechanism working. Minor: appwire says "Loop detection", SteeringItem says "Stream loop guard" — one concept, two names.

`go test ./agent/ ./llm/` green; `golangci-lint` 0 issues; vitest 362 pass; `tsc --noEmit` clean.

obra:
The path forward for this work is now specified on issue #94 (https://github.com/prime-radiant-inc/evener/issues/94#issuecomment-5338130388): a two-phase plan where Phase 1 salvages this branch's sound half (tool-call cycle detector + ceiling, with the fallback-ineligibility and streak-reset fixes specified concretely) and Phase 2 redesigns the text-runaway detector per the rejection findings (amortized rolling hash, entropy-aware thresholds, buffer-level fence counting, 12-fixture negative suite). Note the spec's porting caveat: the Serf→Evener rename means hunks from this branch need re-application, not cherry-picks. Recommend implementing Phase 1 fresh from the spec and closing this PR when that lands.

obra:
Superseded by #246, which ports this PR's judged-sound cycle+ceiling half (credited to e1fee75b3) per the issue #94 Phase-1 spec and drops the rejected chant detector.

PR #132, opened 2026-08-18
feat(tooling): port run-module-lint to serf-dev

obra:
Step 1b of the dev-tooling spec (PR #103): `run-module-lint.sh` (183 lines) and its 868-line stub-toolchain selftest retire together — no parallel implementations. `make lint` now runs `go run ./cmd/serf-dev module-lint` with the same `MODULES`/`LINT_PARALLEL` interface.

## New code (red-first TDD; real child processes only — no fake toolchains, no argv assertions)

- **`internal/devtool/report`** — the one-summary-per-run discipline: PASS/FAIL shapes, the five-class failure vocabulary (`setup` / `not-checked` / `findings` / `results-lost` / `interrupted`), replay fences, retained-log pointers. A second summary panics, so a control-flow bug surfaces in the owning tool's tests instead of shipping.
- **`internal/devtool/procgroup`** — own-group child start, TERM-first stop with bounded KILL escalation, shell-style exit codes. Real-process pins: a forked grandchild dies with the group; a TERM-ignoring child is KILLed after the grace; a cooperative child never waits the grace out.
- **`cmd/serf-dev module-lint`** — the port, one file beside the agent-shards subcommand (#118); the dispatcher keeps the shape #118 shipped, with one added map entry.

## Parity (the deleted selftest's pins → their new home)

| Contract | Status |
|---|---|
| `MODULES` whitespace-split, caller order preserved; default `. agent llm auth envvars invariant identifier` | identical |
| `LINT_PARALLEL` `^[1-9][0-9]*$` (default 4), same diagnostic line, setup summary, exit 2, nothing created | identical (0 / 00 / 08 / 010 / -1 / nope table) |
| Quiet two-line success; chatter hidden; scratch removed | identical |
| Findings: fences in `MODULES` order, passing logs hidden and removed, `full logs:` pointer, `FAIL lint (findings: F/N modules: …)`, exit 1 | identical |
| Missing linter: one diagnostic, `not-checked: N modules: …`, exit 127, no scratch left | identical shape; the diagnostic text is Go's LookPath error instead of bash's `command not found`, and it lands on stderr where the shell `cat`ed it to stdout |
| Scratch-mint failure: diagnostic kept + `setup: unable to create temporary log directory`, exit 1 | identical shape |
| Vanished scratch: both diagnostic lines byte-identical, said once, run stops at the loss, `results-lost: N modules: …`, no pointer to a gone directory, exit 1 | identical |
| Unrecordable per-module result with scratch intact: `unable to record the result for module X`, results-lost, exit 1 | identical |
| HUP/INT/TERM: TERM relayed, children waited, logs removed, `interrupted: SIGNAME`, exits 129/130/143 | identical shapes; the relay reaches the whole process group, a signal landing after the last wave is honored at post-wave checkpoints (review m2 — Notify would otherwise swallow it where the shell trap fired between commands; a signal inside a single log's copy is honored at the next checkpoint, the shell's own granularity), and real delivery is pinned by a built-binary SIGTERM test (review m3, mutation-checked against the Notify list). A checkpoint after the last replay covers a signal landing mid-replay (verifier round 2). Remaining disclosed gap: a signal coinciding exactly with a setup failure (scratch Acquire error, missing linter) exits through that failure's summary without draining the signal — a few-instruction window the shell shared in spirit via trap-set ordering |
| `--allow-parallel-runners`, wave-bounded concurrency, defaults | argv construction is in-tree and mutation-checked against misspelling; **omission is not behaviorally detectable on golangci-lint 2.8.0** — probed empirically: two concurrent flagless runs in separate modules both pass, so the historic parallel-runners lock refusal no longer manifests there and a lock-based omission test would be theater (review m4). The flag stays as protection for configurations where the lock does bite. The wave bound itself is mutation-checked (spawn-all flips its test red) |

**Mechanism artifacts that do not port:** the FIFO start gate and its two `setup:` failure lines (Go waves start together without a gate), and the `.status` files with their unwritable/unreadable results-lost variants (verdicts live in memory; the vanished-scratch and unrecordable-log losses remain, shapes intact).

**Wart fixes by construction, not smuggled features:** `LINT_PARALLEL` past int range is rejected with the standard diagnostic instead of wrapping in shell arithmetic; a TERM-ignoring linter child can no longer hang an interrupted run (5s KILL escalation, wave-runner precedent).

**Exit-code note:** the binary honors the full code contract (2/127/129/130/143); `go run` reports 1 for any nonzero child, so through the Makefile recipe the distinction flattens — same as before at the make level, where any nonzero already fails the target (main documents the same quirk for agent-shards since #120). The dispatcher tests build and run the real binary for exactly this reason.

## Scratch: pid-suffix discipline adopted

`internal/devtool/scratch` landed with #118 mid-flight, so the log directory uses `Acquire`/`Release`: `serf-module-lint.<pid>` naming, dead-pid reclaim of SIGKILLed runs' leftovers at startup (wiring pinned red-first with a dead-pid casualty and a non-pid-suffix keeper), and a findings run's retained logs now live exactly until the next run instead of forever — the covscratch discipline the coverage runners follow.

## Collateral coverage retired with the selftest — flagging for a decision

run-module-lint-selftest.sh also carried pins that were never about this runner: (a) `gitleaks-scan.sh`'s missing-tool policy (optional by default vs `SERF_GITLEAKS_REQUIRED=1`), and (b) a fixture `make lint` proving family composition — the exact call list across the lint targets, lint-gofmt failure modes, lint-generated TypeScript drift, the degraded secret-scan warning, lint-docs replay, quiet-wrapper log cleanup. Those die here un-replaced: the fixture fakes `go` itself, which after this swap would intercept `go run ./cmd/serf-dev` — the design cannot survive the port, and rule 4 classes it as mock-testing. If any of it deserves a new home, the gitleaks missing-tool policy is the strongest candidate; follow-up decision, not smuggled into this PR.

## Adversarial review round (REVISE → fixed)

- **M1 (port regression, fixed):** stopAll signalled every started child's pgid, reaped ones included — a reaped pid can be recycled, so the group TERM could aim at an unrelated process. The shell was immune by construction (single-threaded blank-after-wait). Both layers now skip reaped children (stopAll's select, plus a re-check inside procgroup.Stop), the residual microsecond check-vs-reap window is documented where wave.go documents the same residue, and a red-first decoy test proves a Stop told "already reaped" sends nothing.
- **m2 fixed** (chose the fix over disclosure — it was small): see the signal row above.
- **m3 added**, **m4 resolved by honest disclosure** (see the parallel-runners row; no flaky timing test shipped).
- **Step-2 (module-tests) follow-ups, flagged not fixed here:** `lintSignalNameAndExit` duplicates the agent-shards signal table — the shared home lands with the second consumer's port; `report.Reporter` is not goroutine-safe — module-lint summarizes only from one goroutine, and the concurrency-facing port should revisit.

## Also in this PR

- The envvars audit (caught by the first gate run) is honored: the vanished-scratch diagnostic splices `envvars.TmpDir.Name`, keeping the shell's exact output bytes.
- merge-approval-gate-selftest's make-wiring fixture faked `scripts/run-module-lint.sh`; after the recipe swap the fake was unreachable and nothing asserted its call line — removed.
- docs/testing.md rule 2 updated: the two vanishing-scratch fixture weapons were the audit's only delete-is-the-behaviour lines; the shell estate now holds zero. Both stale allowlist entries were removed after watching the count-pin fail in the stale direction.
- docs/conventions/go-workspace.md names the Go runner.
- Wave drops 20 → 19 suites (main added merge-into-branch mid-flight).

## Verification

- New packages: unit + real-process + real-golangci-lint integration tests green, re-run under `-race` after the final rebase; `gofmt`/`golangci-lint`/`go vet` clean on the new code.
- All three script audits + the envvars audit green; the recursive-delete audit watched RED on both stale entries before re-pinning.
- Dev-tooling wave: 20/20 twice pre-rebase, 18/18 or 19/19 after each rebase as main's wave grew (plus the in-gate run).
- `make lint-golangci` against the real tree: 7 modules, PASS, the two-line contract (re-run post-rebase, 53s).
- Deliberate real-finding check through the make path (`make lint-golangci GO_MODULES=<scratch copy of identifier with an injected unused func>`): findings summary, fence, retained pid-suffixed logs verified on disk — then cleaned up (re-run post-rebase).
- `make merge-approval-gate` in a clean /tmp worktree on this head (e02373098, post-review-fixes): green, exit 0, wave 18/18 inside it. Home canary 32 before / 32 after.
- CI flake note: this PR's CI runs have redded three times on `TestSubagentFinalizationRefusesResumeAndDriveUntilCallbackRestored` (agent module, explicit_resume, ~0.14s) — diff-unrelated, part of a known cluster being root-caused separately; local gates on the same heads are green.
- Gate history: run 1 caught the envvars registry audit (fixed above); run 2 redded on `TestRunShellBackgroundCloseDuringStartDoesNotCommitJob` (agent module, "job manager close timed out"), which passed 3× in isolation — a load flake under two concurrent gates, unrelated to this branch, flagged rather than re-gated blind.

https://claude.ai/code/session_01D5AsxUYaemKcf1aMj7YpkX


obra:
CI note: `build-and-test` redded on `TestSubagentFinalizationRefusesResumeAndDriveUntilCallbackRestored/watch_delegate_send` (agent module, -race). This diff touches nothing in agent/; the test passes 5×5 with -race locally on this exact head, and no prior commit references it as a known flake. Judged an unrelated load flake per the flakes policy; CI job re-run. Adding it to the day's un-root-caused flake ledger (with the wave_test 5s grace and the shell-background close-during-start timeout) for follow-up.

PR #133, opened 2026-08-18
fix(tooling): test-cost cleans up its own scratch

obra:
Step 1c of the dev-tooling spec v2 (#103): the last tool that leaked scratch on every run stops.

**The leak:** `test-cost.sh` minted with `mktemp -d -t` — on macOS that ignores TMPDIR and lands in the Darwin per-user temp directory, outside the dev-tooling wave's leftover isolation — and armed no trap, so every run abandoned its compiled `pkg.test` (~6MB) and logs.

**The fix** adopts the coverage runners' discipline verbatim: source `covscratch-lib.sh`, `reclaim_own_scratch` before minting, pid-suffixed name chosen and EXIT trap armed **before** `mkdir`, disarm-on-collision. Nothing here invents anything; it is the fourth-through-sixth lines of `coverage-union.sh` with test-cost's own voice.

**Audit updates, both watched RED first:** `test-cost.sh` leaves `mktempAllowedScripts` (the escaping mint is gone — the audit failed on the stale entry until removed) and joins `recursiveDeleteAllowedLines` pinned at 1 beside the runners whose pattern it adopts (the audit failed on the unpinned trap delete until pinned).

**Verified against the identifier module under isolated TMPDIRs:**
- clean run leaves nothing; TMPDIR honored;
- SIGKILL mid-run leaves `test-cost.<pid>`; the next run reclaims it (verified twice, pre- and post-rebase);
- TERM mid-build ran the trap but the delete raced the still-live linker and the directory survived holding only `pkg.test` — the next run's reclaim removed it, which is the designed recovery for exactly that class (the sibling runners' selftests dodge this race by group-killing);
- legacy `test-cost.XXXXXX` leftovers in the Darwin temp dir carry no pid and are not reclaimed; they age out with that directory.

**Gate:** `make merge-approval-gate` in a clean /tmp worktree at 3d266b2fc — 27 suites PASS, one red: `run-module-lint-selftest`'s interrupt-timing checks ("interrupt fake did not start"), a suite this diff does not touch; isolated re-run under its own TMPDIR passes 180/180, judged a load flake per the flakes policy. Home canary 32 before and after. Three script audits green; `bash -n` clean; shellcheck shows only the sibling-standard SC1091.

### DIFF
--- a/scriptmktemp_audit_test.go
+++ b/scriptmktemp_audit_test.go
@@ -65,7 +65,6 @@ var mktempAllowedScripts = map[string]bool{
 	"fuzz-oracle-audit.sh":        true,
 	"fuzz-registry-check.sh":      true,
 	"fuzz-triage.sh":              true,
-	"test-cost.sh":                true,
 }
 
 // TestNoScriptCreatesScratchOutsideTMPDIR keeps the swept scripts swept: the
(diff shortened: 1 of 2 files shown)


obra:
CI note: `build-and-test` redded on `TestE2E_ControlInvariantDuringPreTurnWorkAtATurnBoundary` (cmd/serf-hub). This diff is 20 lines of shell in test-cost.sh plus an audit allowlist entry — no connection. Passes 5× locally on this exact head. Judged an unrelated flake; CI re-run. This is the fourth distinct un-root-caused flake sighted across today's CI/gate runs — ledger growing.

PR #211, opened 2026-08-19
fix(hub): queue RPC calls during AppwireClient reconnecting state

### LINKED ISSUE #195: Hub web UI: AppwireClient socket closed (1006) and cannot call while "reconnecting"
# Hub web UI: AppwireClient socket closed (1006) and cannot call while "reconnecting"

## Problem

From the hub web interface, recurring errors appear:

```
Couldn't load activity: AppwireClient: socket closed (code 1006)

Couldn't load activity: AppwireClient: cannot call "serf/jobs/list" while state is "reconnecting"
```

## Symptoms

- WebSocket connection drops with code 1006 (abnormal closure, no close frame received).
- After the drop, the client enters a "reconnecting" state but RPC calls made during that window fail with "cannot call ... while state is reconnecting".
- The errors recur, suggesting either the connection is unstable or reconnection is not completing successfully.

## Expected behavior

- The WebSocket connection should be stable, or reconnect cleanly and resume calls without surfacing errors to the user.
- RPC calls made during reconnection should be queued or retried, not fail with a user-visible error.

## Unknowns

- What is causing the socket to close (code 1006)? Server-side disconnect, network issue, or a bug in keepalive/ping handling?
- Is reconnection actually completing, or is it stuck in the "reconnecting" state?
- How often does this reproduce — intermittent or constant?
- Does this correlate with load, session activity, or specific operations?
- Browser/OS specifics unknown.

## Note

The `"serf/jobs/list"` wire method name in the error is also relevant to the in-progress Serf→Evener rename — that wire method string will become `"evener/jobs/list"`.

obra:
Fixes #195

## Problem

After a WebSocket drop (code 1006), AppwireClient enters the "reconnecting" state. RPC calls made during that window were rejected immediately with a user-visible error:

- `AppwireClient: socket closed (code 1006)`
- `AppwireClient: cannot call "serf/jobs/list" while state is "reconnecting"`

## Fix

This introduces a client ready-gate for the "reconnecting" state. When the client is reconnecting, RPC calls are queued rather than rejected. Queued calls are flushed onto the fresh socket once the re-handshake reaches "ready", so they succeed after reconnection instead of failing with a user-visible error.

Key design points:

- **Queued calls live outside `pending`** (in a `queuedSends` array) until flushed. This is load-bearing: a call that was *already in flight* when the socket dropped was sent and may already be executing server-side, so it must be rejected exactly once and never replayed (could double-fire a non-idempotent method like `turn/start`). A queued call was *never sent*, so it survives every failed reconnect attempt and only resolves on the one that succeeds. Keeping it out of `pending` is what prevents `handleSocketLoss`'s `failAllPending` from rejecting it on a failed attempt.
- **Per-call timeout keeps running unchanged** across the queue → flush transition, so the call's deadline is measured from when it was made. A slow reconnect can still time a queued call out; a fast one preserves most of its budget.
- **"connecting"/"closed" still reject** as before — a never-connected client has no prior ready to fall back to, and a closed one never will. The heartbeat (exempt) bypasses the gate. This preserves the existing contract relied on by `fakeClient`, the threads store, and `errors.ts`'s friendly-error mapping.
- **Queue is cap-bounded** (`QUEUED_SENDS_MAX = 256`) so a prolonged outage cannot grow it without bound; on overflow the newest calls reject rather than absorb a queue the connection has no near-term hope of draining.
- **`close()` rejects queued calls** with the same `ConnectionClosedError` an in-flight call gets.

## Tests

Replaces the old test that asserted calls while reconnecting are "rejected synchronously, not queued" with four new tests in `protocol/reconnect.test.ts`:

1. A call made while reconnecting is queued, then delivered (frame appears on the new socket, promise resolves) once reconnected.
2. A queued call survives a failed reconnect attempt (not rejected) and resolves on the one that succeeds.
3. A queued call's own timeout rejects it if reconnection takes longer than its deadline, and the timed-out call is not replayed.
4. `close()` rejects queued calls with `ConnectionClosedError` without sending them.

## Verification

- `cd cmd/evener-hub/frontend && npm test` — 341 test files, 6733 tests pass.
- `npx biome check --write` on touched files — clean.
- `npx tsc --noEmit` — clean.

### DIFF
--- a/cmd/evener-hub/frontend/src/panes/sessionPanels/sessionPanelPane.test.tsx
+++ b/cmd/evener-hub/frontend/src/panes/sessionPanels/sessionPanelPane.test.tsx
@@ -24,8 +24,17 @@ beforeEach(() => {
 
 afterEach(() => {
   cleanup();
-  vi.useRealTimers();
+  // restoreAllMocks() before useRealTimers(), not after: "Details owns a
+  // clock after hydration" spies on globalThis.setInterval WHILE fake
+  // timers are active, so vi.restoreAllMocks() (which restores a spy to
+  // whatever it captured as "original" at spyOn time - here, the FAKE
+  // setInterval) must run before useRealTimers() gets the last word on what
+  // globalThis.setInterval actually is. Reversed, restoreAllMocks() runs
+  // last and reinstalls that stale fake implementation, so every timer
+  // registered afterward - including testing-library's own real-timer
+  // waitFor() polling - silently never fires for the rest of this file.
   vi.restoreAllMocks();
+  vi.useRealTimers();
   // The beforeEach above only resets threadsStore BEFORE each test. Several
   // tests here render SessionPanelPane against a real connected FakeClient,
   // which calls ensureThread()/the mutation runtime for real - without this,
--- a/cmd/evener-hub/frontend/src/protocol/client.ts
+++ b/cmd/evener-hub/frontend/src/protocol/client.ts
@@ -98,31 +98,6 @@ interface PendingRequest {
   timer: ReturnType<typeof setTimeout>;
 }
 
-// A call made while "reconnecting" was never sent — the socket is gone — so it
-// is safe to hold and deliver on the fresh socket once the re-handshake lands,
-// rather than failing the caller with a user-visible "cannot call ... while
-// reconnecting" error. queuedSends holds the full call (payload plus its
-// resolve/reject/timer) until the re-handshake reaches "ready", at which point
-// flushQueuedSends() moves each entry into `pending` and sends it. This is
-// deliberately distinct from a call that was ALREADY in flight when the socket
-// dropped: that one was sent, the server may already be executing it, and
-// replaying it blind could double-fire a non-idempotent method (turn/start).
-// Those failAllPending rejects, exactly once, and they are never replayed.
-//
-// Crucially, a queued call lives ONLY in queuedSends, not in `pending`, until
-// it is flushed. That is what keeps a failed reconnect attempt's
-// handleSocketLoss -> failAllPending from rejecting it: the queue is a call
-// the server never saw, so it must survive every reconnect attempt that fails
-// short of "ready" and finally resolve on the one that succeeds.
-interface QueuedSend {
-  id: number;
-  method: MethodName;
-  params: unknown;
-  resolve: (result: unknown) => void;
-  reject: (err: Error) => void;
-  timer: ReturnType<typeof setTimeout>;
-}
-
 export class AppwireClient {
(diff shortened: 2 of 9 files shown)


obra:
## Adversarial review: REJECT — transport-level queuing reintroduces the mutation double-fire hazard

This implements the design the #195 RCA explicitly rejected. The queue gate in `client.ts` keys only on connectionState + an exempt-list of {initialize, ping}; there is no read/mutation distinction. Seven mutation call sites that reach `client.request()` with no pre-flight ready check (`setModel`, `setReasoningEffort`, `setGoal`, `rename`, `compact`, `clearThread`, `shutdown`, `forkFromTurn`, `resolveEscalation` — threads.ts:1967-2115) now silently queue during "reconnecting" and auto-fire post-reconnect. A user retrying a non-idempotent mutation whose first attempt dropped mid-request (server may have processed it) can now get both fired — exactly the double-fire class the synchronous rejection existed to prevent. The composer/outbox path (send/steer/queue/interrupt) is unaffected only because it has its own ready gate upstream.

Compounding: the pinned test that protected this property ("requests attempted while reconnecting are rejected synchronously, not queued for replay") was rewritten to assert the opposite, using a read method as its example. Nothing pins fail-fast mutations anymore.

Secondary defect in the new mechanism: `setState("closed")` transitions reached internally (attemptReconnect's noteProtocolFailure branch, performHandshake's catch, handleSocketLoss's "connecting" branch) never call `failAllQueued()` — only the public `close()` does — so a queued call whose client autonomously gives up dangles until its own per-call timeout (up to 30s) and then reports a misleading RequestTimeoutError. Verified no double-send from this path; it's a diagnosability gap, not corruption.

Also unimplemented from the RCA's plan: the caller-side coverage of the five read-only methods, the shared extraction of the 4x-duplicated `clientReadyOrRewired` retry shape, and the friendly-error routing in activitySummary/errors.ts — on genuine timeout the raw "timed out after 30000ms" string still bypasses classification into the toast.

What held up: mutation-testing the new tests passed both directions (reverting the queue branch reds 4/28; reverting close's failAllQueued reds the close test); full frontend suite 6733/6733, tsc and biome clean.

Path to mergeable, either of:
1. The RCA's design: revert the client state-machine change (restore the pinned rejection test as it was), add `requireReadyClient()` (bounded wait on clientReadyOrRewired) applied to listJobs/listTasks/jobOutput/listModels/loadOlderTurns, and route the activity-pane failure through the existing friendly-error classification.
2. Keep transport queuing but strictly allowlist read-only methods, restore a pinned fail-fast test for every non-allowlisted method, and make every internal closed-transition drain the queue via failAllQueued.

PR #222, opened 2026-08-19
Go rewrites: replace custom shell tooling with standard Go + delete all selftests

obra:
## Summary

Follow-up to PR #218 (scripts reorg). Replaces the logic-heavy shell scripts with Go packages and standard `go test`/`go tool cover`, and deletes all selftest/meta-test infrastructure.

## Go packages (TDD: red→green for each)

| Package | Replaces | Tests |
|---------|----------|-------|
| `internal/devtool/covstmt` | `covstmt-lib.sh` (Python heredoc for coverage statement counting) | 9 |
| `internal/devtool/gatesurface` | `gate-surface-lib.sh` (gate test surface config) | 8 |
| `cmd/evener-dev capability-preflight` | `gate-capability-preflight.sh` (eval-emit → structured output) | 9 |
| `cmd/evener-dev coverage-floor` | `test-coverage-floor.sh` + `coverage-union.sh` + `coverage-gaps.sh` + `test-cost.sh` | 9 |

## Scripts deleted (replaced with standard Go tooling)

- `run-module-tests.sh` (456 lines) → `make test` uses `go test` directly with `-run`/`-skip` flags
- `test-timing-budget.sh` (340 lines) → deleted (timing ratchet not providing value)
- `gate-capability-preflight.sh` → Go subcommand
- `test-coverage-floor.sh`, `coverage-union.sh`, `coverage-gaps.sh`, `test-cost.sh` → Go coverage-floor + `go tool cover`
- `fuzz-coverage-global.sh` (644 lines) → Go coverage-floor tool
- `gate-surface-lib.sh`, `covstmt-lib.sh` → Go packages (dead shell libs)

## No tests for tests

- All 12 `*-selftest.sh` scripts deleted
- `scripts/lib/selftest-lib.sh` (assertion library) deleted
- `scripts/lib/covscratch-selftest-lib.sh` deleted
- `cmd/evener-test-dev-tooling/` (wave runner, 6 Go files) deleted
- `DEV_TOOLING_TEST_SCRIPTS` and `make test-dev-tooling` deleted
- All selftest Makefile targets deleted

## What remains as shell (workflow tools doing real orchestration)

- `merge-into-branch.sh` — atomic git merge via worktree + CAS
- `e2e-webui-turn-controls.sh`, `e2e-ratelimited-provider.sh` — process harnesses
[...]

PR #224, opened 2026-08-19
fix(gate): preflight fails closed when the capability script can't run

### LINKED ISSUE #181: Gate capability preflight fails open when the script cannot execute
The Makefile runs the gate capability preflight via eval on the script's output. If the script cannot be executed at all (missing, not executable), the command substitution yields empty output, evaluating the empty string returns 0, the chain continues, and an inherited SERF_GATE_CAPABILITY_SKIP survives unoverwritten — the gate silently proceeds uncertain. The script's own doc comment promises the opposite ("evaluates to exit 1 ... rather than silently proceeding uncertain"), which only holds once the script actually runs.

Partial mitigation already landed in PR #123's fix pass: the runner now names any capability narrowing on stderr. The fail-open eval itself still needs closing (guard it with an explicit existence/executability check that exits 1).

---
Filed from the adversarial review session: https://claude.ai/code/session_0134LFae5DzpDExkuhJTCNfU

obra:
Guard the merge-approval-gate preflight eval with an explicit -x existence/executability check that exits 1, so a missing or non-executable scripts/gate-capability-preflight.sh stops the gate instead of silently continuing with an inherited EVENER_GATE_CAPABILITY_SKIP. Red-green TDD: merge-approval-gate-selftest.sh pins fail-closed for both missing and non-executable cases.

Fixes #181


obra:
Adversarial review (verdict: approve after the fixes pushed below; CI pending).

**Verified against issue #181:**
- Red-first: reverting the Makefile guard makes exactly the 4 new selftest checks fail, reproducing the documented fail-open (gate ran `lint` and exited 0 with the preflight missing). Restoring it goes green.
- Mutation checks: weakening `-x` to `-e` is caught by the non-executable case; changing the guard's `exit 1` to `exit 0` is caught by both rc assertions.

**Fixes pushed to this branch:**
1. Merged `origin/main`: PR #218 moved the preflight to `scripts/gate/`, so the branch was CONFLICTING and its guard checked a path that no longer exists. Guard, eval, and selftest fixture paths repointed.
2. Closed a residual fail-open the `-x` guard alone misses: an *executable* preflight that exits nonzero before emitting its verdict (broken sourced lib, syntax error) still produced an empty substitution whose status the bare eval discarded — demonstrated: gate ran all four phases and exited 0. The recipe now captures the substitution and checks its exit status; a new crashing-preflight selftest case pins it (red against the bare eval, green with the capture).

**Merge-ordering warning re PR #222 / #226:** #222 deletes `gate-capability-preflight.sh` *and* this selftest, and its replacement recipe is `export EVENER_GATE_CAPABILITY_SKIP=$$(go run ./cmd/evener-dev capability-preflight)` — `export VAR=$(cmd)` returns 0 regardless of the command's status (verified empirically under make's shell), so #222 reintroduces exactly this fail-open while deleting the tests that catch it. #226 keeps the same line. This PR should land first, and #222's recipe needs a plain assignment (or explicit status check) plus a test for it.

Issue #169, opened 2026-08-18
Anthropic Opus 4.7/4.8 reject temperature: session auto-naming silently fails


obra:
Opus 4.7 and 4.8 have the same request-shape contract as the Claude 5 models: they reject temperature/top_p/top_k with a 400 (the vendored litellm catalog agrees: supports_sampling_params=false on both) and default thinking.display to "omitted". But isClaude5OrNewer / the catalog flag treats them as old-shape.

Live consequence: agent/session_namer.go:72 sends Temperature (0.0) on every session-naming call, so any session whose cheap model is Anthropic Opus 4.7/4.8 gets a 400 and silently fails to auto-name.

Also: TestEmbeddedModelCatalog_Claude5RequestShape currently pins claude-opus-4-7 -> false, which makes the wrong shape look reviewed, and the flag name is a misnomer — the real boundary is generation 4.7+, not 5.

Found during adversarial review of PR #131 (kata g422 scoped this half out on the premise "nothing misfires today", which is wrong). Kata 4ta0 owns the fix.

---
Filed from the adversarial review session: https://claude.ai/code/session_0134LFae5DzpDExkuhJTCNfU

obra:
Fixed by PR #232 (merged): claude5_request_shape extended to claude-opus-4-7/-4-8 via catalog overrides (dated variants inherit via family overlay); boundary behavior verified down to opus-4.6 staying false. Residual prefixed-catalog-key gap tracked on #151.

https://claude.ai/code/session_0134LFae5DzpDExkuhJTCNfU

Issue #152, opened 2026-08-18
hub-synthesized SessionMetas violate the ParentSessionID/DivergenceTurn delegate invariant if it travels


obra:
The runtime-correctness batch (kata `96cp`, closed) made `Meta()` preserve restored parent lineage, relying on the invariant: persisted `ParentSessionID != ""` with `DivergenceTurn == 0` implies a spawned delegate (every fork writer validates `divergenceTurn >= 1`; verified across fork.go and aside.go).

cmd/serf-hub/web_api_tree.go:884-895 builds in-memory SessionMetas for remote/Codex threads with `ParentSessionID` set from `thread.Serf.ParentRef`/`ForkedFromID` and `DivergenceTurn` left zero. These are never persisted, so there is no live bug — but any future consumer applying the invariant to hub-synthesized metas would misclassify a forked remote thread as a delegate.

Either stamp `DivergenceTurn` on the synthesized metas, or document the invariant as local-persistence-only where it is relied on.

Found by the 96cp fix-round re-review.

Related: kata `96cp` (closed).

---
_Migrated from kata `5swg`._

obra:
**Verified against current main (origin/main @ 771aa0ea7).** Repo/binary has since been renamed serf→evener; the current path is `cmd/evener-hub/web_api_tree.go` (line numbers otherwise barely drifted).

**1. Synthesis site — claim still true.** `cmd/evener-hub/web_api_tree.go:884-896`:

```go
meta := schema.SessionMeta{
    ID:             refText,
    ProfileID:      ref.SourceID,
    Model:          thread.ModelProvider,
    CreatedAt:      createdAt,
    UpdatedAt:      updatedAt,
    OriginalPrompt: title,
    EnvInfo: schema.EnvironmentInfo{
        WorkingDir: thread.CWD,
    },
    ParentSessionID: appThreadTreeParentSessionID(thread, ref),
    IsSubagent:      thread.Evener.Kind == "subagent",
}
```

`ParentSessionID` comes from `thread.Evener.ParentRef`/`ForkedFromID` via `appThreadTreeParentSessionID` (lines 920-932); `DivergenceTurn` is never set, so it's the zero value.

**2. The invariant's real consumer.** It's not fork.go/aside.go/session.go directly — it's the restore path: `agent/session_init.go:706-710` builds `forkInfo` from a *persisted, loaded-from-disk* meta, and `agent/session_state.go`'s `Meta()` logic does:

```go
[...]

Issue #151, opened 2026-08-18
platform-prefixed Claude 5 catalog entries resolve flagless and get the wrong request shape


obra:
Kata `g422` (closed) rewired `isClaude5OrNewer` to the catalog flag `claude5_request_shape`, set on claude-sonnet-5, claude-fable-5, and claude-opus-5. Platform-prefixed base-catalog entries (us.anthropic.claude-opus-5, eu.anthropic.\*, vertex_ai/claude-sonnet-5, azure_ai/\*) resolve to their own catalog entries, which carry no flag, so they return false and would get sampling params plus no thinking display — the same bug class g422 fixed for the bare IDs. Nothing routes these refs through the anthropic adapter today, so the bug is latent.

Also on the record from that batch: upstream litellm data carries `supports_sampling_params:false` on 49 keys including claude-opus-4-7 and claude-opus-4-8, unconsumed by `parseLiteLLMCatalog` — a deliberate scope cut. It is a three-way split (`supports_sampling_params` ⊃ `claude5_request_shape` ⊃ `thinking_always_on`), and widening to opus-4-7/4-8 is a ruling, not a mechanical fix.

Evidence is in the g422 task history and the drift-guard test `TestEmbeddedModelCatalog_Claude5FlagCoversEveryClaude5CatalogKey`.

Related: kata `g422` (closed).

---
_Migrated from kata `4ta0`._

obra:
## RCA

### Latent-or-live verdict: **live**, not latent — contradicts this issue's own framing

The issue's premise ("nothing routes these refs through the anthropic adapter today") is wrong. evener has no dedicated Bedrock/Vertex/Azure *driver* — confirmed by grepping every `RegisterInstanceAdapterFactory` call (`llm/providers_config.go:31`) across the tree: the only registered `(type, apiStyle)` pairs are `anthropic`, `openai{responses,auto,chat-completions}`, `google`, `openrouter`, `openrouter-anthropic`, `kimi`, `kimi-anthropic`, `glm`, `minimax`, `ollama` — the exact behavior-tag list `docs/llm-providers.md:89-91` documents. `docs/superpowers/specs/2026-07-02-openai-compat-followups-plan.md:131` says it plainly: *"Deliberately out of scope (Jesse has not asked): Azure/Bedrock/Vertex drivers."*

[...]

PR #232, opened 2026-08-19
fix(hub): extend Claude 5 request shape to Opus 4.7/4.8

obra:
## Issue #169

Opus 4.7 and 4.8 reject temperature/top_p (like Claude 5+ models) but `isClaude5OrNewer` and the catalog `Claude5RequestShape` flag treated them as old-shape. `session_namer.go:72` sends `Temperature 0.0` on every naming call, so any session whose cheap model is Opus 4.7/4.8 gets a 400 and silently fails to auto-name.

## Fix

**Catalog overrides only. No Go logic changes.**

Add `claude5_request_shape: true` to `claude-opus-4-7` and `claude-opus-4-8` in `evener_model_catalog_overrides.json`:
- `claude-opus-4-7` — existing entry had only `{"aliases": ["opus"]}`; added the flag
- `claude-opus-4-8` — new entry with just the flag (LiteLLM base supplies the rest)

The catalog is the sole authority for the 4.7+ boundary. The `isClaude5OrNewer` ID fallback parse (`n >= 5`) is unchanged — the catalog now covers the only two 4.7+ models that exist, and the fallback remains conservative for uncatalogued models.

## Side effect: thinking.display

4.7/4.8 also get `thinking.display="summarized"` (same as Claude 5). The issue confirms 4.7/4.8 default `thinking.display` to "omitted" like Claude 5, so requesting "summarized" is the correct behavior (we want visible thinking). These models already had `supports_adaptive_thinking: true` from the LiteLLM base, so they were already on the adaptive-thinking path — the only new wire effect is the `display` field.

## Consumer audit

`isClaude5OrNewer` (request.go:35) is the only consumer. It gates:
- **Temperature/top_p** (lines 65, 68): now omitted for 4.7/4.8 → the bug fix
- **thinking/display="summarized"** (lines 146-156): newly stamped (intended, matches Claude 5)

Unaffected: `claude-opus-4-6`, `claude-sonnet-4-6`, `claude-opus-4-5`, `claude-sonnet-4-5`, all `claude-3-*`. The `opus` alias (resolves to 4-7) is also fixed.

## Tests

`TestEmbeddedModelCatalog_Claude5RequestShape` updated: moved 4-7 from want-false to want-true loop, added 4-8 to want-true.

Two new test files:
- `llm/providers/anthropic/opus47_request_shape_test.go` — pins `isClaude5OrNewer` true for 4-7/4-8, false for 4-6; pins `buildRequestBody` omits temperature/top_p for 4-7
- `llm/model_catalog_opus47_test.go` — pins `Claude5RequestShape` true for 4-7/4-8, false for 4-6

```
go test ./llm/providers/anthropic/ -run 'TestIsClaude5OrNewer|TestBuildRequestBody_Claude5|TestBuildRequestBody_Opus47' -v   # 9/9 PASS
go test ./llm/ -run 'TestEmbeddedModelCatalog_Claude5|TestEmbeddedModelCatalog_Opus47' -v                                  # 3/3 PASS
go vet ./llm/ ./llm/providers/anthropic/                                                                                   # clean
go test ./llm/... -short                                                                                                  # all PASS
go test ./agent/ -short                                                                                                   # PASS (67s)
```

Three independent GLM proposals were unanimous: catalog overrides only, fallback unchanged, no rename.

### DIFF
--- a/cmd/evener-hub/web_api_tree.go
+++ b/cmd/evener-hub/web_api_tree.go
@@ -894,6 +894,13 @@ func appThreadTreeEntries(thread appwire.Thread) (schema.SessionMeta, hubcore.Li
 		ParentSessionID: appThreadTreeParentSessionID(thread, ref),
 		IsSubagent:      thread.Evener.Kind == "subagent",
 	}
+	// Issue #152: a hub-synthesized fork (parent set, not a subagent) must
+	// stamp DivergenceTurn so the 96cp invariant (ParentSessionID != "" &&
+	// DivergenceTurn == 0 implies a spawned delegate) does not misclassify a
+	// remote fork as a delegate. Value 1 is the minimum legal fork turn.
+	if meta.ParentSessionID != "" && !meta.IsSubagent {
+		meta.DivergenceTurn = 1
+	}
 	if thread.GitInfo != nil {
 		meta.EnvInfo.GitBranch = thread.GitInfo.Branch
 		meta.EnvInfo.GitOriginURL = thread.GitInfo.OriginURL
--- a/cmd/evener-hub/web_api_tree_divergence_test.go
+++ b/cmd/evener-hub/web_api_tree_divergence_test.go
@@ -15,8 +15,8 @@ import (
 // applying the invariant.
 func TestAppThreadTreeEntries_ForkedRemoteThreadStampsDivergenceTurn(t *testing.T) {
 	thread := appwire.Thread{
-		Source:   "remote",
-		ID:       "child-thread-1",
+		Source:    "remote",
+		ID:        "child-thread-1",
 		SessionID: "child-session-1",
 		Evener: appwire.EvenerThread{
 			Ref:       "remote:child-thread-1",
@@ -43,8 +43,8 @@ func TestAppThreadTreeEntries_ForkedRemoteThreadStampsDivergenceTurn(t *testing.
 // a fork).
 func TestAppThreadTreeEntries_NoParentLeavesDivergenceTurnZero(t *testing.T) {
 	thread := appwire.Thread{
-		Source:   "remote",
-		ID:       "standalone-thread",
+		Source:    "remote",
+		ID:        "standalone-thread",
 		SessionID: "standalone-session",
 		Evener: appwire.EvenerThread{
 			Ref: "remote:standalone-thread",
--- a/cmd/evener-hub/web_api_tree_divergence_test.go
+++ b/cmd/evener-hub/web_api_tree_divergence_test.go
@@ -1,0 +1,63 @@
+package main
+
+import (
+	"testing"
+
+	"primeradiant.com/evener/appwire"
+)
+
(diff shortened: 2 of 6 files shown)


obra:
Adversarial review (issue #169 / #151 bucket B). Verdict: **mergeable after the fixes I pushed** (da667b4), with one ordering constraint and one tracked residual.

## Verified

**Boundary behavior (proven empirically against `isClaude5OrNewer` + `buildRequestBody`):**

| model ref | claude5 shape | via |
|---|---|---|
| `claude-opus-4-5`, `claude-opus-4-6`, `claude-sonnet-4-5`, `claude-sonnet-4-6`, `claude-haiku-4-5` | false | catalog |
| `claude-opus-4-7`, `claude-opus-4-8` | **true** | override |
| `claude-opus-4-7-20260416` (dated) | **true** | family inheritance at catalog load; wire body confirmed: no `temperature`, `thinking.display="summarized"` |
| `opus` alias, `claude-opus-4-7[1m]`, `anthropic/claude-opus-4-7` | **true** | alias / suffix-strip / last-segment retry |
| `claude-sonnet-5`, `claude-opus-5`, `claude-fable-5` | true (unchanged) | |
| `gpt-5`, `claude-opus`, `claude-`, empty | false | |
| `claude-opus-4-9` (uncatalogued, hypothetical) | false | fallback `n >= 5`, conservative as the PR body states |

**Red-first:** reverting only the catalog override makes all four new tests fail for the right reasons. **Mutation:** flagging `claude-opus-4-6` true is caught by four tests.

## Fixes pushed (da667b4)

1. **CI was red**: `lint-gofmt` failed on both new test files (misaligned struct literals in `opus47_request_shape_test.go` and `web_api_tree_divergence_test.go`). gofmt'd.
2. **Mutation gap**: off-by-one'ing the ID-generation fallback (`n >= 5` → `n >= 4`) survived the whole suite — every catalogued pre-5 case short-circuits through the catalog flag, so the fallback's false branch was untested. Added uncatalogued `claude-opus-4-3` → false to the boundary table; the mutant now dies.

## Do not merge before #231

This branch carries #231's commit (4050c80, DivergenceTurn / issue #152) while based on main. Merging #232 first lands #231's change without its own review. Merge #231 first, or rebase this branch onto main without that commit.

## Residual (tracked, not blocking)

The prefixed catalog keys for opus-4-7/4-8 (`us./eu./au./jp./global.anthropic.claude-opus-4-7/-4-8`, `anthropic.claude-opus-4-7/-4-8`, `vertex_ai/…[@default]`, `azure_ai/…`, `perplexity/anthropic/claude-opus-4-7`) still carry `claude5_request_shape=false` — proven empirically. They're only reachable via custom-gateway configs (#151's bucket B). #151's mechanical fix spec is scoped to bucket A (Claude-5 spellings), so whoever implements it must extend to the 4-7/4-8 prefixed spellings too, or these fall between the two issues. Noting it on #151 as well.

Issue #197, opened 2026-08-18
Mobile UI launch screen layout breaks for wide tasks


obra:
## Summary

The Mobile UI launch / "Welcome" screen does not handle tasks that are too wide, resulting in a broken layout.

## Evidence

Screenshot (IMG_4476.png) of the launch screen state:
- App title bar shows `Welcome`, `No session open`.
- A prompt to open a `New session`.
- Bottom banner text: `r. This means that we're going to cover all of the binary names, al`.
- Sidebar with a `command palette` shortcut group (K palette, I composer, J next session), next steps list (`can read and edit the repository, run commands, and delegate helpers.`, `Try a task to get started`, `Find and fix the root cause of a flaky test`, `Audit error handling across the auth package`, `Explain how requests flow from router to handler`), and an inline note: `inside the command palette shows all shortcuts.`.

Tasks that are too wide overflow / break the layout of this launch screen and need to be constrained or wrapped.

## Environment

- Mobile UI / launch screen.

obra:
## RCA: launch-screen overflow on mobile (two symptoms, one root cause)

Reproduced in a real headless-Chrome layoutguard probe (390x844 mobile viewport, real `tokens.css`/`button.module.css`/`welcome.module.css`/`panescaffold.module.css`/`emptystate.module.css`, no mocks) reconstructing `Welcome.tsx`'s actual DOM. Both reported symptoms trace to the **same** mechanism.

### Root cause

`Welcome.tsx`'s "Jump back in: {candidate.title}" button (rendered only when a resume candidate exists — `resumeCandidate()`/`tbk8`) is a `variant="primary"` `Button`. `button.module.css`'s `.button` sets `white-space: nowrap` unconditionally (a deliberate rule — see its own comment: it exists to stop a *squeezed* label from wrapping inside a fixed-height box). `candidate.title` is server-truncated at 200 runes (`tree.go`'s `maxTitleRunes`/`truncateTitle`) but **not** further constrained client-side — 200 runes at 13px UI font is several times a 390px phone's width.

[...]

PR #234, opened 2026-08-19
fix(ui): cap welcome .actions width and override resume button nowrap

obra:
## Issue #197

The Mobile UI Welcome screen layout breaks for wide tasks — the resume-candidate button's 200-rune server title inflates the centered flex layout past the viewport, and PaneScaffold `.body overflow-x:clip` slices it into a marker-less fragment.

## Root cause

1. The resume Button inherits `white-space: nowrap` from `button.module.css:11` (correct globally, wrong here).
2. Uncapped centered flex ancestors (\.actions\) inflate to the 200-rune title's min-content width.
3. `.examples`/`.hints` resolve `width: min(100%, 36rem)` against the inflated `.actions`, so they overflow too.
4. PaneScaffold `.body overflow-x:clip` slices everything mid-string.

## Fix (CSS-only, in `welcome.module.css`)

1. **`.actions`**: add `width: 100%`, `max-width: 36rem`, `overflow-wrap: break-word`, `word-break: break-word` — caps the flex column so an unbreakable child wraps instead of inflating the layout.
2. **`.actions button`**: override `white-space: nowrap` to `white-space: normal` locally (mirrors the pre-existing `.examples button` override at line ~37 which does the same thing for the example-prompt buttons). `button.module.css` is left untouched — its `nowrap` is correct globally.

## Test

New file `welcome.overflow.test.tsx` (3 tests): full long title in DOM, `.actions` caps width + wraps, `.actions button` overrides nowrap.

```
npx vitest run src/panes/welcome/welcome.overflow.test.tsx   # 3 passed
npx vitest run src/panes/welcome/                             # 22 passed (3 files)
npx biome check (both files)                                  # clean
```

Investigation comment: [#197 comment](https://github.com/prime-radiant-inc/evener/issues/197#issuecomment-5344096804)


obra:
## Review verdict: MERGE AFTER FIXES

Two defects in the original `.actions button` CSS override (PR head `051cadbdd`) were found and fixed on `fix-197-welcome-overflow` (now at `391ba8b2b`, a clean merge of `origin/main` on top of the fix commit `fd77c0bdc`):

1. **Vertical spill.** The original fix added `white-space: normal` to unwrap the resume-candidate button's long title, but did not add `height: auto` — `button.module.css` pins `.md` height to 32px (44px on phone widths), so the wrapped 200-rune title spilled 19–27px out of the fixed-height box.
2. **Suggestion buttons recentered.** `.actions button { text-align: center }` was declared *after* `.examples button { text-align: left }` in source order. At equal specificity that silently overrode the three task-suggestion buttons' deliberate left alignment.

The fix (`fd77c0bdc`): `.actions button` gets `height: auto` with `min-height: 32px` as a floor (re-declared as `var(--tap-min)` in the phone media block, since the higher-specificity rule would otherwise beat `button.module.css`'s own `@media` floor), drops `text-align: center` entirely, and moves above `.examples button` so that rule's 44px floor still wins for the suggestion buttons.

### Evidence (headless Chrome, via new layoutguard case `welcome-resume-candidate-overflow`)

I independently reproduced all three states myself before verifying:

- **RED on `main`** (no fix at all): resume button 1163px wide clipped symmetrically in a 320–390px pane body; `.examples`/`.hints` stretched to 576px.
- **FAIL on the unfixed PR head** (`051cadbdd`, before this review's follow-up): resume button's wrapped label spills 19–27px out of its fixed-height box; all three example buttons compute `text-align: center` instead of `left` — exactly defects #1 and #2 above, nothing else.
- **GREEN on the fixed/merged tree** (`391ba8b2b`): all 16 layoutguard cases pass, including the new one at 320/360/390px.
- Case is mutation-tested: deleting `white-space: normal`/`height: auto` from the fix reproduces the spill failure (documented in the case's `mutation` field).

### Verification run on `391ba8b2b`

- `npm test` (frontend): 342 test files / 6749 tests passed, incl. the PR's `welcome.overflow.test.tsx` (3 tests) and `token-contract.test.ts`.
- `npm run typecheck`: clean.
- `npm run lint` (biome): clean, 919 files.
- `node scripts/layoutguard/run.mjs`: all 16 cases PASS.
- `make test-web-browser`: `web-layoutguard`, `web-overflowguard`, `web-spawnguard` all PASS.

### CI status

`Build darwin-arm64`, `Build linux-amd64`, and `web` are green on the pushed head. `build-and-test` failed twice (original run and one rerun) on the same unrelated step — `Dev-tooling self-test wave` → `fuzz-bisect-selftest: FAIL: bisect did not converge`. This is **not** the known hub-e2e marketplace-teardown flake (fixed in #235); I confirmed it also fails intermittently on `main` right now, independent of this PR (two concurrent `main` CI runs within the same minute, one pass one fail, same failure signature). Filed **#245** to track it separately — it shouldn't block this PR once CI is otherwise unblocked or rerun lands clean.

Not merging — leaving that call to the PR owner.

Issue #161, opened 2026-08-18
a green dev-tooling selftest run can orphan real test processes under load


obra:
Found by the adversarial verifier on kata `y6rq` (closed; 2026-08-18) while reproducing that kata's bounded-load methodology (N=8-16 concurrent runs of scripts/run-module-tests-selftest.sh, host load 16-32).

After runs that exited "102 checks, 0 failed", real orphaned processes (ppid=1) survived for multiple minutes:

- a live real `go test` spawned by the CDPATH check at scripts/run-module-tests-selftest.sh lines ~693-707, which invokes the real scripts/run-module-tests.sh against the actual repo on every selftest invocation (tracked separately as #143); and
- a `sleep 1000` descendant forked by `TestForkedDescendantIsReaped` (whose own load-flakiness is tracked in #173).

This happened under fixed post-y6rq code, so a fully green selftest does not certify its process tree was reaped under concurrent load.

Why it matters: leaked real go-test processes compete for CPU and disk and are a plausible contributor to the ambient-load flake mechanism y6rq root-caused; they also silently grow GOCACHE. Both leak sources are pre-existing and untouched by the y6rq fix (c78ec2aba).

**Done when:** after N concurrent selftest runs complete under load, no process rooted in any run's tree survives its run's exit, and something (the selftest itself or the wave runner) fails loudly when one does.

Related: #143.

---
_Migrated from kata `hjrz`._

obra:
## RCA against current main (`d262c952f`), building on #143 and #173

Read #143 and #173 in full first (both cross-referenced from here). Short version: one of this issue's two named sources doesn't reproduce on current `main`; the other is already correctly root-caused in #173 with a fix plan I'm not duplicating; and there's a third, currently-unaddressed gap at the wave-runner level that I think is the actual target of this issue's "Done when."

### Source 1 (CDPATH check / #143): doesn't reproduce today

[...]

PR #236, opened 2026-08-19
fix(wave): group-kill suite process group on green exit

obra:
## Issue #161

A green dev-tooling selftest run can orphan real test processes under load.

## Root cause

`runSuite` only sends TERM/KILL to the suite's process group on the **interruption** (shutdown) path. On a green exit it reaps only the direct child via `cmd.Wait()` and never touches the group, so a descendant that outlives the suite shell — reparented to pid 1 but still a member of the suite's process group — survives the wave as an orphan. Under concurrent selftest load this leaks real `go test` processes and `sleep` fixtures.

## Fix

Call `killSuiteGroup(pgid)` on the green-exit path after `cmd.Wait()` returns 0, before the leak check — the same cleanup the interruption path applies. This ensures no process rooted in the suite's tree outlives a green run, and prevents a lingering descendant from writing to the private TMPDIR between the suite's exit and the leak check.

## Test

`TestGreenWaveReapsForkedDescendant` — a deterministic test that forks a long-lived descendant (`sleep 120`), exits 0, and asserts the descendant is reaped after `runWave` returns green. `t.Cleanup` SIGKILLs the descendant so a red run leaves no orphan behind.

```
go test ./cmd/evener-test-dev-tooling/ -run TestGreenWaveReapsForkedDescendant -v   # PASS
go test ./cmd/evener-test-dev-tooling/ -v -count=1                                  # 14/14 PASS
go vet ./cmd/evener-test-dev-tooling/                                                # clean
```

Investigation comment: [#161 comment](https://github.com/prime-radiant-inc/evener/issues/161#issuecomment-5344167329)

### DIFF
--- a/cmd/evener-test-dev-tooling/wave.go
+++ b/cmd/evener-test-dev-tooling/wave.go
@@ -216,6 +216,16 @@ func runSuite(cfg waveConfig, runDir, name string, shutdown <-chan struct{}) sui
 		}
 		return suiteResult{exitCode: code, seconds: seconds}
 	}
+	// On a green exit the watcher goroutine returns without touching the
+	// process group — it only group-kills on the shutdown path. A descendant
+	// the suite orphaned (reparented to pid 1 but still a member of this
+	// suite's process group) would survive the wave as a live process. Kill
+	// the group now, the same cleanup the interruption path applies, so no
+	// process rooted in this suite's tree outlives a green run. Doing it
+	// before the leak check also prevents a lingering descendant from
+	// writing to the private TMPDIR between the suite's exit and the check.
+	// (Issue #161.)
+	killSuiteGroup(pgid)
 	// Leak check is post-reap bookkeeping that could block on a wedged
 	// filesystem. Run it with a timeout so it doesn't block the wave's
 	// ability to report this suite's result.
--- a/cmd/evener-test-dev-tooling/wave.go
+++ b/cmd/evener-test-dev-tooling/wave.go
@@ -203,6 +203,35 @@ func runSuite(cfg waveConfig, runDir, name string, shutdown <-chan struct{}) sui
 	mu.Unlock()
 	close(reaped)
 	seconds := int(time.Since(start).Round(time.Second).Seconds())
+
+	// Sweep the suite's process group for survivors on every exit path, green
+	// or red (issue #161): cmd.Wait() reaps only the direct child, so a
+	// descendant the suite orphaned — reparented to pid 1 but still a member
+	// of this suite's process group — would outlive the run as a live
+	// process. Probing before killing keeps a routine exit from firing
+	// SIGKILL at a possibly-recycled pgid, and sweeping before the leak check
+	// keeps a lingering descendant from writing to the private TMPDIR between
+	// the suite's exit and the check.
+	survivors := suiteGroupSurvivors(pgid)
+	if len(survivors) > 0 {
+		select {
+		case <-shutdown:
+			// The watcher already TERMed the group; give descendants the
+			// grace period to finish their TERM traps (one can legitimately
+			// still be writing its shutdown event here) before escalating to
+			// KILL, mirroring the watcher's own TERM-then-KILL escalation.
+			deadline := time.Now().Add(cfg.KillGrace)
+			for len(survivors) > 0 && time.Now().Before(deadline) {
+				time.Sleep(25 * time.Millisecond)
+				survivors = suiteGroupSurvivors(pgid)
+			}
+			if len(survivors) > 0 {
+				killSuiteGroup(pgid)
+			}
+		default:
+			killSuiteGroup(pgid)
+		}
+	}
 	if err != nil {
 		code := cmd.ProcessState.ExitCode()
 		if code < 0 {
@@ -216,16 +245,17 @@ func runSuite(cfg waveConfig, runDir, name string, shutdown <-chan struct{}) sui
 		}
 		return suiteResult{exitCode: code, seconds: seconds}
 	}
-	// On a green exit the watcher goroutine returns without touching the
-	// process group — it only group-kills on the shutdown path. A descendant
-	// the suite orphaned (reparented to pid 1 but still a member of this
-	// suite's process group) would survive the wave as a live process. Kill
-	// the group now, the same cleanup the interruption path applies, so no
-	// process rooted in this suite's tree outlives a green run. Doing it
-	// before the leak check also prevents a lingering descendant from
-	// writing to the private TMPDIR between the suite's exit and the check.
-	// (Issue #161.)
-	killSuiteGroup(pgid)
+	// A suite that exits green while leaving live processes in its group has
+	// a cleanup bug; the sweep above killed them, and this failure is what
+	// keeps that bug visible — the process counterpart of the temp-file leak
+	// check below.
+	if len(survivors) > 0 {
+		return suiteResult{
+			failure: fmt.Sprintf("evener-test-dev-tooling: %s passed but left process(es) alive in its process group: %s",
+				name, strings.Join(survivors, "; ")),
+			seconds: seconds,
+		}
+	}
 	// Leak check is post-reap bookkeeping that could block on a wedged
 	// filesystem. Run it with a timeout so it doesn't block the wave's
 	// ability to report this suite's result.
--- a/cmd/evener-test-dev-tooling/wave_other.go
+++ b/cmd/evener-test-dev-tooling/wave_other.go
@@ -24,3 +24,7 @@ func killSuiteGroup(pgid int) {
 		_ = proc.Kill()
 	}
 }
+
+// suiteGroupSurvivors has no group to probe on this platform; the direct
+// child is all there is, and cmd.Wait() already reaped it.
+func suiteGroupSurvivors(int) []string { return nil }
--- a/cmd/evener-test-dev-tooling/wave_test.go
+++ b/cmd/evener-test-dev-tooling/wave_test.go
@@ -567,50 +567,34 @@ func TestReadFifoLineGraceFollowsTheTestDeadline(t *testing.T) {
 	}
 }
 
-// TestGreenWaveReapsForkedDescendant is a RED test for issue #161.
-//
-// runSuite only sends TERM/KILL to the suite's process group on the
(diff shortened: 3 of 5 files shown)


obra:
## Adversarial review: REJECT (two gaps against issue #161's done-when)

Verified first, so credit where due: the core mechanism works and the test is honest.

- **Red-first**: with the `killSuiteGroup(pgid)` hunk reverted, `TestGreenWaveReapsForkedDescendant` fails with the exact leak ("green wave leaked an orphaned descendant (pid …) still alive after runWave returned 0"), and a real fixture (`sleep 300 & exit 0`) run through the pre-fix binary reports `PASS`, exits 0, and leaves `sleep 300` alive at ppid=1 in the suite's pgid. With the fix, same fixture: `PASS`, exit 0, no survivor, and the invoking shell's own process group is untouched (kill fires only after `cmd.Wait()`, on the child's pgid).
- Test is race-clean (`-count=3 -race`), full package + `go vet` green locally; the wave package also passed on CI.

### Finding 1 (fixed in place): the branch carried PR #234's commit with base=main

The branch was stacked on `051cadbdd` (PR #234, the issue #197 welcome-CSS fix) while declaring base `main` — merging this PR would have silently merged #234's content past its own review. The wave commit has zero dependency on it, so I rebased the branch onto `origin/main` (dropping `051cadbdd`) and force-pushed; the PR is now the single wave commit (`8ffd586d7`) and the diff is wave.go/wave_test.go only.

### Finding 2 (blocking): the "fails loudly" half of the done-when is missing

Issue #161's done-when has two clauses: no process survives a run's exit, **"and something (the selftest itself or the wave runner) fails loudly when one does."** The linked RCA's fix item 1 says the same: "KILL the group **+ fail loudly** ('suite passed but left processes in its group')", parallel to the existing leaked-temp-files failure. This PR only kills, silently — demonstrated: a suite that orphans a `sleep` still prints `PASS` and the wave exits 0. The leaking suite's bug stays invisible forever, which is exactly what the leak-check precedent in this file refuses to allow for files. Shape: probe the group (`syscall.Kill(-pgid, 0)`; `ESRCH` ⇒ clean) after `Wait()`, and on survivors kill the group **and** return a `suiteResult{failure: "… passed but left process(es) alive in its process group"}`. Probe-before-kill also narrows the pid-reuse exposure of firing an unconditional `SIGKILL` at a possibly-already-freed pgid on every green exit — the same reuse window this file's own mutex comment worries about. (Caution: a just-exited reparented descendant can linger as a zombie for an instant; the RCA's suggested `ps`-based listing of live group members, or excluding zombies, avoids a spurious FAIL.)

### Finding 3 (blocking): a red-exiting suite still leaks

The new `killSuiteGroup(pgid)` sits after the `if err != nil { return … }` block (wave.go:206–218), so a suite that fails **on its own** (nonzero exit, no shutdown signal) never gets a group sweep either — the watcher goroutine only kills on `<-shutdown`. Demonstrated: fixture `sleep 300 & exit 1` through the fixed binary → `FAIL redleak`, wave exits 1, and `sleep 300` survives at ppid=1. The done-when says "no process rooted in **any** run's tree survives its run's exit," not only green ones. The sweep belongs before the exit-code branch (right after `close(reaped)`), covering both paths.

### Notes (non-blocking)

- RCA items 2 and 3 (replace `run-module-tests.sh`'s PPID walk with a group kill; selftest survivor assertion) are untouched here — fine as follow-ups, and the PR correctly avoids a closing keyword so #161 stays open.
- PR #222 (delete the wave runner) was rejected, so this is live code on the `make test-dev-tooling` / merge-approval-gate path; no interaction beyond that.
- CI `build-and-test` on the previous head failed only on the known hub-e2e marketplace teardown race (`TempDir RemoveAll … marketplaces/claude-plugins-official/plugins: directory not empty`, fix pending in #235) — unrelated, not re-filed.

obra:
Rebased onto current main (89a3689c5, post-#247) and addressed the reject findings. Branch is now three commits: 46c3a9ddd (original), fd33c052a (findings 2+3), a2c514891 (gate consequence).

**Finding 1 (stacking)** — resolved by the reviewer's in-place rebase; this push keeps the branch directly on main.

**Finding 2 (fail loudly)** — fd33c052a: after `cmd.Wait()` the group is probed with `kill(-pgid, 0)` (ESRCH ⇒ clean, so a routine exit never SIGKILLs a possibly-recycled pgid); a non-empty group is listed via `ps` with zombies excluded, so a just-exited reparented descendant awaiting init's reap can't cause a spurious FAIL. Survivors on a green exit are killed AND the suite fails with `passed but left process(es) alive in its process group: <pid cmd>`, the process counterpart of the leaked-temp-files check. Red-first verified: with the hunk reverted, `TestGreenWaveReapsForkedDescendant` fails with "wave exited 0, want 1".

**Finding 3 (red-exit leak)** — fd33c052a: the sweep sits right after `close(reaped)`, before the exit-code branch, so green, red, and shutdown exits are all swept. Red-first verified: with the hunk reverted, `TestRedWaveReapsForkedDescendant` leaks its orphan ("still alive after runWave returned").

**Gate consequence (new commit a2c514891)** — under the now-strict check, `e2e/e2e-webui-turn-controls` failed loudly: its fixture daemons parked on `while :; do sleep 1; done`, so a signalled fixture left its in-flight `sleep 1` orphaned in the suite's group for up to a second past the suite's exit (reproduced: `passed but left process(es) alive ...: 60591 sleep 1`). A suite can't wait on descendants it doesn't parent, so the fixtures now park on a blocking FIFO read — no child process at all — and the cleanup trap waits (bounded) for every signalled pid to die before exiting. Verified 3x standalone, then the full gate.

Verification: `go test ./cmd/evener-test-dev-tooling/ -count=1 -race` 16/16 PASS; process-group tests 5x race-clean; `go vet` clean; windows+linux cross-builds OK; `make test-dev-tooling` 15/15 PASS with zero survivors.

RCA items 2 and 3 remain follow-ups as noted; no closing keyword, so #161 stays open.

obra:
Rebased onto current main (`fd33c052a`/`46c3a9ddd`, plus a new `a2c514891` fixing a fixture leak found while re-verifying — details below). Mapping the review's two rejection findings to what's now in the branch, plus a fresh verification pass:

## Finding 1: silent leak — no loud failure

`suiteGroupSurvivors(pgid)` probes with `kill(-pgid, 0)` as a cheap emptiness check (also keeps a routine exit from SIGKILLing a possibly-recycled pgid), then lists live members via `ps`, excluding zombies so a just-exited descendant awaiting init's reap can't cause a spurious FAIL. A green suite with survivors now returns a failing `suiteResult` that names the suite and says so explicitly:

`TestGreenWaveReapsForkedDescendant` (`cmd/evener-test-dev-tooling/wave_test.go:619`) is red-first: it forks a suite that leaks a live descendant on a **green** exit and asserts wave exit code `1`, `FAIL  orphan` in the output, and the literal diagnosis `orphan passed but left process(es) alive in its process group` — then confirms the descendant is actually reaped.

## Finding 2: red exits get no sweep

`runSuite` now probes the process group right after `cmd.Wait()`, before the exit-code branch, so every exit path is swept — not just the shutdown-signal path. `TestRedWaveReapsForkedDescendant` (`wave_test.go:646`) pins this: a suite that's already red on its own exit code still gets its leaked descendant reaped, with `FAIL  redleak` in the output.

## Reviewer's cautions (zombie exclusion, pid-reuse)

- Zombie children are excluded from the `ps` listing — a descendant that has already exited and is just waiting on its parent's `Wait()`/init's reap isn't a survivor.
- `kill(-pgid, 0)` is used as a probe before any kill, so a pgid that's been recycled to an unrelated process group after the suite's own leader has fully exited won't get SIGKILLed on the strength of a stale number.

## Consequence check: no real suite leaks

Ran the actual suite wave against the live scripts (not synthetic fixtures) three times this session:
- Pre-rebase head (`1db801dc0`): 16/16 real suites PASS, clean.
- Same head again: PASS, clean.
- Post-rebase head, this PR's actual current HEAD (`a2c514891`): 15/15 PASS, clean — the count dropped by one only because main's own `DEV_TOOLING_TEST_SCRIPTS` list (Makefile) changed under the rebase (`test-coverage-floor` selftest is no longer in the wave's suite list on current main); not a regression.

While re-verifying, one of those runs surfaced a real (if narrow) leak: `e2e-webui-turn-controls-selftest`'s fixture daemons parked on `while :; do sleep 1; done`, so a signalled fixture could leave its in-flight `sleep` orphaned in the suite's process group for up to a second past the suite's own exit — exactly the race this fix is designed to catch. Root-caused and fixed at the source in `a2c514891`: fixtures now park on a blocking FIFO read (no child process to leave behind at all), and the cleanup trap waits, bounded, for every signalled pid to actually die before the suite exits. Re-ran clean afterward. This is the loud-failure mechanism doing exactly its job — catching a real, if rare, leak in test tooling rather than the wave runner itself, and getting it fixed instead of papered over.

Also clean: `go vet` and `golangci-lint` (both pass in CI's `build-and-test` job, steps 6–7), and `go test -race -count=3 ./cmd/evener-test-dev-tooling/...`.

## CI status: build-and-test and web are red, but not from this diff

Both jobs fail on this branch's CI run, but they fail **identically** on main's own tip at the exact commit this branch rebased onto (`89a3689c5`, run https://github.com/prime-radiant-inc/evener/actions/runs/32290161587) — confirmed by running that job directly against main, not a downstream branch. Two unrelated failures, both traced to the hub-frontend RailRow rework (`4ba794abb`/`40499aa07`, part of #247):

- `web-layoutguard`'s `rail-row-chevron-actions-overlap` case fails its own new right-justification assertion.
- `TestScenarioQuotedLiteralsSitInsideTheCitedLineRange` fails on two stale line citations into `RailRow.tsx` from `test/scenarios/spawn-empty-prompt-starts-dormant.md` (the file's lines moved under the rework; the citations didn't).

Neither touches `cmd/evener-test-dev-tooling/` or this PR's diff. The race-detector step failing first also means CI's own "Dev-tooling self-test wave" step never got to run in this job (it's downstream and gets skipped on the earlier failure) — the 15/15 clean run above is the real verification of this PR's actual behavior. Filed https://github.com/prime-radiant-inc/evener/issues/266 to track the main breakage separately; not rerunning build-and-test/web since it's not a flake and won't go green until main is fixed.

Not merging — leaving that decision to a human.

Issue #140, opened 2026-08-18
delegate-lifecycle CI flake cluster: subagent finalization, drain handoff, attention rearm


obra:
`TestSubagentFinalizationRefusesResumeAndDriveUntilCallbackRestored` failed CI three times on 2026-08-18, all on PRs whose diffs cannot touch `agent/`: run 32155422710 (/watch_delegate_send, -race job), run 32159514049 (/explicit_resume, -race), and run 32161482108 (/explicit_resume, deterministic suite). The failure is identical every time: `provider requests during finalization = 3, want no resumed or automatic turn` (subagent_owned_job_drain_test.go:587). It passes 40x locally under `-race GOMAXPROCS=2`.

**Working hypothesis** (investigated in /tmp/serf-lane-agentflake): the owned-shell completion's routing forks in `jobManager.armFinalizedJob` at `persistStableShellAttention`. The stable path arms fenced delegate attention on the parent, delivered post-finalization via rearm — the window sees 2 requests and the test passes. The plain-owner-notification path enqueues on the child, and DrainJobTree's notification turn delivers it in-drain, before the finalization window closes — the window sees 3 and the test fails. Which branch is taken appears timing- and state-dependent at the finalization boundary. Suspects: store-record TerminalGen visibility at the `persistStableShellAttention` load, `stableOwner`/`parentDelegateID` setup timing, and the incomplete-source error path's retry semantics.

**Cluster context:** a same-day hub-side flake (`TestE2E_ControlInvariantDuringPreTurnWorkAtATurnBoundary`, tracked in #154) was root-caused separately in PR #137 (a 32-frame outbound buffer against the 4096 contract, with silent eviction) — a different mechanism, but the same "delegate/turn lifecycle under CI load" family. Also outstanding in the family: the wave_test 5s grace (kata `ebnq`, closed) and `TestRunShellBackgroundCloseDuringStartDoesNotCommitJob` (unfiled).

## History

**2026-08-18 — investigation round 1 complete** (session aafb5cbc fork).

Reproduction: 100+ local iterations across `-race`, `GOMAXPROCS=2`, and two concurrent go-test processes on a loaded machine — all green. This matches PR #137's experience that Darwin does not produce these CI interleavings.

The mechanism was narrowed by a full static trace plus live instrumentation of a passing run:

[...]

PR #237, opened 2026-08-19
fix(agent): close drain re-enqueue race during finalization

obra:
## Issue #140

`TestSubagentFinalizationRefusesResumeAndDriveUntilCallbackRestored` failed CI three times — a **real bug**, not a flake.

## Root cause

The `NotifyPending→NotifyDelivered` transition in `armFinalizedJob` is non-atomic. Between appending the `NotifyPending` event and `persistStableShellAttention` marking the record `NotifyDelivered`, the drain loop's `rematerializeDurablePendings` (run on every 250ms recheck kick) observed the still-`NotifyPending` record and re-enqueued it onto the child's notification queue. The drain then fired a provider turn during finalization where none was expected (CI count=3 vs want 0).

## Fix

In `rematerializeDurablePendings` (`agent/session_jobtree_drain.go`), snapshot the live running-map job IDs under `jm.mu` and skip any `NotifyPending` record still in the running map. `armFinalizedJob` owns that notification until the run is deleted; `rematerialize` is for pendings that survive ONLY in the durable store after the run is gone.

## Test

`TestSubagentFinalizationDrainReenqueueRaceDeterministic` — a deterministic test that forces the race window open via a nil-in-production test-only hook (`testOnlyAfterNotifyPendingAppend`) and asserts `peekNotifications` stays 0 after `rematerializeDurablePendings` during the finalization window.

```
go test ./agent/ -run TestSubagentFinalizationDrainReenqueueRaceDeterministic -v   # PASS
go test ./agent/ -run TestSubagentFinalizationRefusesResumeAndDriveUntilCallbackRestored -v   # PASS (3/3)
go vet ./agent/    # clean
```

Investigation comment: [#140 comment](https://github.com/prime-radiant-inc/evener/issues/140#issuecomment-5344260292)

### DIFF
--- a/agent/jobs.go
+++ b/agent/jobs.go
@@ -166,6 +166,13 @@ type jobManager struct {
 	// (still durable, not yet Delivered) and re-enqueues it during the drain's
 	// 250ms recheck, reproducing the issue-#140 finalization provider-request leak.
 	testOnlyAfterNotifyPendingAppend func(jobID string)
+	// testOnlyAfterRematerializeDurableLoad fires inside rematerializeDurablePendings
+	// after it has read the durable records, before it decides what to re-enqueue.
+	// It is nil in production; deterministic tests use it to hold a rematerialize
+	// pass on its already-loaded durable snapshot while a finalization completes,
+	// proving the pass cannot re-enqueue a record whose NotifyPending state is
+	// stale. Read and written under jm.mu.
+	testOnlyAfterRematerializeDurableLoad func()
 	// stopReceiptBeforeWait observes the fatal-finalization-only join boundary.
 	// It is nil in production and exists so concurrency tests can release an
 	// asynchronously terminating shell only after the exact receipt is awaited.
--- a/agent/jobs.go
+++ b/agent/jobs.go
@@ -158,6 +158,14 @@ type jobManager struct {
 	// receive and its timeout-flag check.
 	shellBeforeWaitTimeoutDecision  func(*atomic.Bool)
 	shellBeforeBlockTimeoutDecision func(*atomic.Bool)
+	// testOnlyAfterNotifyPendingAppend fires after armFinalizedJob has appended
+	// the NotifyPending event and marked the terminal notification-pending, but
+	// BEFORE persistStableShellAttention transitions NotifyPending→NotifyDelivered.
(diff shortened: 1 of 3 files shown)


obra:
## Review: MERGE AFTER FIXES

Approach is sound and matches the #140 RCA: the drain re-enqueue race stems from a durable load and running-map snapshot being taken at different points in time relative to `armFinalizedJob`'s finalization sequence, letting a completed job get counted as still outstanding and re-enqueued after its notification was already delivered.

**Found + fixed:** the re-enqueue guard's rematerialize path (`agent/jobs.go`) called `store.Load()` *outside* `jm.mu`, then snapshotted the running map separately — a TOCTOU narrower than but structurally identical to #140's original bug. A finalization completing between the two reads leaves a stale `NotifyPending` record with no live run, which the guard then re-enqueues after delivery.

This violates the same-hold discipline already documented on `outstandingDrainJobIDs` in `agent/session_jobtree_drain.go:60-68`:

> The durable snapshot and the running-map read are taken under the SAME jm.mu hold. `armFinalizedJob` deletes from the running map under `jm.mu` and appends `EventJobNotificationPending` before that delete, so while this holds `jm.mu` a finalizing managed job is either still in the running map (delete blocked) or its NotifyPending record is already durable (delete done, its earlier append visible to Load). Loading the store outside the lock would reopen the stale-snapshot window.

Fixed in bbc0fc546: moved the load + running-map snapshot under one `jm.mu` hold, added a `testOnlyAfterRematerializeDurableLoad` hook to make the straddle deterministically reproducible, and added `TestSubagentFinalizationRematerializeLoadSnapshotStraddle`. Also gofmt'd the PR's test file, which was failing CI's lint-gofmt.

**Evidence:**
- Red-first: both `TestSubagentFinalizationRematerializeLoadSnapshotStraddle` and the original #140 regression test fail against the pre-fix ordering, pass after.
- `go test -race -count=50` on the deterministic tests: green.
- `go test -race -count=30` on the flaky family: green.
- `go test ./agent/... -race -count=1` on PR head (bbc0fc546): green across all 38 packages, no failures, straddle test included.
- `make lint`: PASS.
- `gh pr checks 237`: all green (build-and-test, both platform builds, web) — no #235 marketplace-teardown flake, no #243 CI-only flake observed this run.

No further changes requested.

PR #242, opened 2026-08-19
fix(web): give spawn the composer's control row on a phone (#198)

### LINKED ISSUE #198: Attachment button should be inside the text area, reusing the composer UI for attachments and model setting
## Summary

The attachment button should be inside the text area, and the Mobile UI should reuse the same UI as the composer for attachments and setting the model.

## Evidence

Screenshot (IMG_4476.png) of the launch screen state:
- App title bar shows `Welcome`, `No session open`.
- A prompt to open a `New session`.
- Bottom banner text: `r. This means that we're going to cover all of the binary names, al`.
- Sidebar with a `command palette` shortcut group (K palette, I composer, J next session), next steps list (`can read and edit the repository, run commands, and delegate helpers.`, `Try a task to get started`, `Find and fix the root cause of a flaky test`, `Audit error handling across the auth package`, `Explain how requests flow from router to handler`), and an inline note: `inside the command palette shows all shortcuts.`.

In this state the attachment button is not inside the text area. We should reuse the composer UI for attachments and setting the model here.

## Environment

- Mobile UI / launch screen / composer.

obra:
### Root cause

`Spawn.tsx` passed `PromptCard` a `controlsClassName={CLASS.actionBand}`, and
`spawn.module.css`'s `@media (max-width: 899px)` block made `.actionBand`
`position: fixed; bottom: 0`. The card's own control row therefore left the
card and became a page-level band, stranding the attach button at the foot of
the viewport, nowhere near the prompt it attaches to. Measured in the new
layoutguard harness at 390x844 against main: the card sits at `12,12 366x106`
while the control row sits at `0,767 390x77`.

The session composer has never had this problem — `Composer.tsx` keeps attach
and the model control inside the card's controls row at **all** widths. So the
fix is to render the same shape rather than a third placement that resembles
it.

### Fix

**Attach (Part A).** The mobile `.actionBand` overrides, the now-empty desktop
`.actionBand` rule, and the `.form` `padding-bottom` that reserved room for the
band are gone; `Spawn.tsx` drops `controlsClassName`. `PromptCard`'s
`controlsClassName` prop is deleted outright — a caller reaching in to restyle
the shared row is what made this possible. The row's touch targets survive on
the widgets' own phone blocks (`button.module.css` / `iconbutton.module.css`,
both keyed to `--tap-min`), which spawnguard now **measures** rather than
assumes. The 16px textarea font and its `min-height` stay: that is the iOS
zoom floor, not action-band compensation. Testid renamed
`spawn-mobile-actions` → `spawn-controls`.

**Model (Part B).** `ModelSwitchTrigger` is extracted from
`panes/session/chrome/ModelSwitch.tsx` as a value-controlled presentational
component (same directory, same `modelswitch.module.css`), keeping the
`— change model` screen-reader suffix, the load-bearing whitespace text node,
and the `sessionActionHeadline` + `friendlyLaunchErrorMessage` error framing.
`ModelSwitch` becomes a thin thread-bound wrapper and keeps its
`model-switch-trigger` / `model-switch-value` testids. Spawn's card renders
that same trigger in its leading slot beside attach, gated phone-only in CSS.
`MobileSettingRows` drops its Model row and its bespoke model sheet (6 rows → 5).

### Stated decisions

- **Effort stays in `MobileSettingRows`.** Combining model and effort into one
  pill the way the session status row does is a product call, not a bug fix;
  flagged, deliberately out of scope here.
- **The model sheet's "Use default" row is dropped.** Strict composer reuse:
  `(default)` is the trigger's own resting label, which is the composer's
  answer to the same question.
- **Desktop is untouched.** The configuration row's `ModelField` stays; the
  card's trigger is `display: none` above 899px, because two controls for one
  setting is worse than one.
- **No `<form>` / `onSubmit` anywhere in Spawn** (kata `t13x`;
  `test/scenarios/spawn-keyboard-contract.md` Part A greps for exactly this).

### Guards

- **New** `scripts/layoutguard/cases/spawn-attach-in-card/` (390x844): the
  control row is not `position: fixed`, and the row, attach, the model trigger
  and Start all sit inside the card, with attach below the field rather than
  overlaid on it. Authored red-first against main's markup and CSS; the
  mutation evidence is recorded in `case.json`.
- **Updated** `scripts/spawnguard/run.mjs` + `src/dev/spawnguard-entry.tsx`:
  the same containment at 390/899/900px, the 44px touch floor measured on the
  real buttons, the model trigger visible at 390/899 and hidden at 900, and
  five setting rows.

### Verification

- `npx tsc --noEmit --incremental false` — clean
- `npm test` — 343 files, 6764 tests passed
- `npm run lint` (`biome ci src`) — clean
- `npm run layoutguard` — 17/17 PASS
- `npm run spawnguard` — 390/899/900px PASS
- `npm run overflowguard` — PASS
- `go test ./` (root scenario/doc audits) — ok

Closes #198

### DIFF
--- a/cmd/evener-hub/frontend/scripts/layoutguard/cases/spawn-attach-in-card/assert.mjs
+++ b/cmd/evener-hub/frontend/scripts/layoutguard/cases/spawn-attach-in-card/assert.mjs
@@ -4,8 +4,9 @@
 // spawn.module.css turned into `position: fixed; bottom: 0` inside its 899px
 // media block, so the card's own control row detached from the card and became
 // a page-level band. Measured in this same harness on the pre-fix CSS at
-// 390x844: the card spans y=8..205 while the control row (and the attach
-// button inside it) sits at y=768..844.
+// 390x844: the card is 12,12 366x106 while the control row is 0,767 390x77,
+// with the attach button at 16,784 44x44 - two thirds of a phone screen away
+// from the prompt it attaches to.
 //
 // The session composer has never had this problem - it keeps attach, the model
 // trigger and the send verb inside the card's control row at every width
--- a/cmd/evener-hub/frontend/scripts/layoutguard/cases/spawn-attach-in-card/assert.mjs
+++ b/cmd/evener-hub/frontend/scripts/layoutguard/cases/spawn-attach-in-card/assert.mjs
@@ -1,0 +1,89 @@
+// Issue #198 / the rejected PR #242: on a phone the spawn pane's attach button
+// was stranded at the bottom of the VIEWPORT, nowhere near the prompt it
+// attaches to. Spawn passed PromptCard a controlsClassName that
+// spawn.module.css turned into `position: fixed; bottom: 0` inside its 899px
+// media block, so the card's own control row detached from the card and became
+// a page-level band. Measured in this same harness on the pre-fix CSS at
+// 390x844: the card spans y=8..205 while the control row (and the attach
+// button inside it) sits at y=768..844.
+//
+// The session composer has never had this problem - it keeps attach, the model
+// trigger and the send verb inside the card's control row at every width
+// (Composer.tsx's leading/actions slots) - so the fix is to render the same
+// shape, and this case is what pins it there.
+//
+// Four geometric claims, and one placement claim aimed at the OTHER way this
+// has been got wrong:
+//
+//   1. The control row is not a fixed band. `position: fixed` is the
+//      mechanism; naming it directly means a reintroduction fails here for
+//      the reason it is wrong rather than only through its symptoms.
+//   2. Everything in the row - the row itself, attach, the model trigger,
+//      Start - is inside the card. That is the containment the fixed band
+//      broke, and it is what "reuse the composer's UI" means geometrically.
+//   3. The model trigger is actually on screen at phone width. It is the
+//      whole of issue #198's second half (setting the model from the card),
+//      and a display:none typo would otherwise pass claim 2 vacuously - a box
+//      with no size is trivially "contained".
+//   4. Attach sits BELOW the writing surface, not on top of it. PR #242's
+//      first attempt overlaid the paperclip on the textarea's lower-right
+//      corner, where typed text runs underneath it. The composer puts it in
+//      the row under the field; so does this.
+const TOLERANCE = 0.5; // sub-pixel layout rounding, not a fudge factor
+
+function describe(b) {
+  if (b === null) return "missing";
+  return `${b.left.toFixed(1)},${b.top.toFixed(1)} ${b.width.toFixed(1)}x${b.height.toFixed(1)}`;
+}
+
+function escapes(child, parent) {
+  return (
+    child.left < parent.left - TOLERANCE ||
+    child.top < parent.top - TOLERANCE ||
+    child.right > parent.right + TOLERANCE ||
+    child.bottom > parent.bottom + TOLERANCE
+  );
+}
+
+export default function assert(m) {
+  const failures = [];
+  const missing = ["card", "field", "controls", "attach", "start", "modelTrigger"].filter((k) => m[k] === null);
+  if (missing.length > 0) {
+    return {
+      pass: false,
+      reason: `the harness is missing ${missing.join(", ")} - nothing else can be measured without them`,
+    };
+  }
+
+  if (m.controls.position === "fixed") {
+    failures.push(
+      `the control row is position: fixed (${describe(m.controls)}) - it is a viewport band, not the card's own row`,
+    );
+  }
+  for (const [name, box] of [
+    ["the control row", m.controls],
+    ["the attach button", m.attach],
+    ["the model trigger", m.modelTrigger],
+    ["the Start button", m.start],
+  ]) {
+    if (escapes(box, m.card)) {
+      failures.push(`${name} (${describe(box)}) is outside the prompt card (${describe(m.card)})`);
+    }
+  }
+  if (m.modelGateDisplay === "none" || m.modelTrigger.width <= 1 || m.modelTrigger.height <= 1) {
+    failures.push(
+      `the model trigger is not on screen at ${m.viewport.width}px (display: ${m.modelGateDisplay}, box ${describe(m.modelTrigger)}) - the card is where a phone sets the model`,
+    );
+  }
+  if (m.attach.top < m.field.bottom - TOLERANCE) {
+    failures.push(
+      `the attach button (${describe(m.attach)}) overlaps the writing surface (${describe(m.field)}) instead of sitting in the row beneath it`,
+    );
+  }
+
+  if (failures.length > 0) return { pass: false, reason: failures.join("; ") };
+  return {
+    pass: true,
+    reason: `at ${m.viewport.width}x${m.viewport.height} the control row is ${m.controls.position}, and attach (${describe(m.attach)}), the model trigger (${describe(m.modelTrigger)}) and Start (${describe(m.start)}) all sit inside the card (${describe(m.card)}) below the field`,
+  };
+}
--- a/cmd/evener-hub/frontend/scripts/layoutguard/cases/spawn-attach-in-card/case.json
+++ b/cmd/evener-hub/frontend/scripts/layoutguard/cases/spawn-attach-in-card/case.json
@@ -1,0 +1,25 @@
+{
+  "name": "spawn-attach-in-card",
+  "description": "spawn's prompt card keeps its whole control row - attach, the model trigger, and Start - inside the card at phone width, exactly where the session composer keeps them (issue #198)",
+  "cssFiles": [
+    "styles/tokens.css",
+    "styles/global.css",
+    "widgets/promptcard/promptcard.module.css",
+    "widgets/textarea/textarea.module.css",
+    "widgets/button/button.module.css",
+    "widgets/tooltip/tooltip.module.css",
+    "panes/session/chrome/modelswitch.module.css",
+    "panes/spawn/spawn.module.css"
+  ],
+  "viewport": {
+    "width": 390,
+    "height": 844,
+    "deviceScaleFactor": 2,
+    "mobile": true
+  },
+  "mutation": {
+    "declaration": "spawn.module.css: the @media (max-width: 899px) .actionBand block (position: fixed; bottom: 0; min-height: 76px) restored, with the control row spelled class=\"controls actionBand\" and the attach button as its direct child - i.e. exactly what Spawn.tsx's controlsClassName={CLASS.actionBand} produced before this case existed",
+    "verified": "2026-08-19",
+    "expect": "fail"
+  }
+}
--- a/cmd/evener-hub/frontend/scripts/layoutguard/cases/spawn-attach-in-card/harness.html
+++ b/cmd/evener-hub/frontend/scripts/layoutguard/cases/spawn-attach-in-card/harness.html
@@ -1,0 +1,88 @@
+<!doctype html>
+<html lang="en" data-theme="dark">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=390, initial-scale=1">
+<title>spawn attach-in-card layoutguard harness</title>
+<link rel="stylesheet" href="tokens.css">
+<link rel="stylesheet" href="resolved.css">
+<style>
+  * { transition: none !important; animation: none !important; }
+  /* PaneScaffold's content slot around the form. Only the inset matters here:
+     it is what makes "inside the card" a claim about the card rather than
+     about the viewport edge. */
+  body { margin: 0; padding: var(--space-3); }
+
+  /* IconButton renders button.module.css's base + variant classes plus
+     iconbutton.module.css's OWN .xs square, and BOTH files define `.xs`. CSS
+     Modules keeps those distinct in production; a raw-stylesheet harness
+     cannot, so iconbutton's rule is spelled here under a name of its own
+     (the same dodge compact-session-footer's .attachmentControl documents),
+     phone tap floor included - that floor is what now carries the touch
+     target the deleted action band used to hard-code. */
+  .attachmentControl { width: 24px; height: 24px; padding: 0; }
+  @media (max-width: 899px) {
+    .attachmentControl { min-width: var(--tap-min); min-height: var(--tap-min); }
+  }
+
+  /* Same collision, other direction: widgets/popover and modelswitch both own
+     a local `.trigger`. popover.module.css's whole rule is one declaration
+     (`display: inline-flex`, popover.module.css:6-8), so the wrapper span it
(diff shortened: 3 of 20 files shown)


obra:
## Adversarial review: REJECT

Verified locally on c901325a2 (branch checked out, suites run both ways).

**1. Does not compile — all four CI jobs are red on this PR's own change, not a flake.**
`src/panes/spawn/Spawn.tsx:755` passes `className` to `IconButton`, whose props deliberately `Omit<..., "className">` (`src/widgets/iconbutton/index.tsx:7-19`). `npx tsc --noEmit --incremental false` fails with TS2322 locally; CI `web`, `build-and-test`, `Build darwin-arm64`, and `Build linux-amd64` all fail on this exact error.

**2. The `.fieldAttach` overlay never renders — the fix's own CSS is dead code.**
`IconButton` spreads `{...rest}` BEFORE its computed `className` (deliberate merge order, documented in `src/widgets/iconbutton/index.tsx`'s own comment), so a caller-passed `className` is discarded. Verified with a runtime probe: the rendered `<button>`'s class list does not contain the passed class. The PR body's "attach button overlays the textarea's lower-right corner via `.fieldAttach`" does not exist in the running app; what actually renders is the paperclip as a static element below the textarea inside the field slot. `position: relative` on `.fieldArea` anchors nothing.

**3. No layoutguard/geometry case — the added test is jsdom ancestry only.**
`AttachmentInsideTextArea.test.tsx` asserts DOM containment (confirmed red-first: fails on main's implementation, passes on the branch — good as far as it goes). But docs/testing.md ("The Three Browser Guards") is explicit that jsdom "evaluates no cascade and reports zero for every box"; this bug class needs a layoutguard or spawnguard geometry assertion. Its absence is exactly why finding 2 shipped unnoticed: the structural test stays green with the positioning class silently dropped.

**4. Not scoped to mobile; desktop layout changes too.**
`.fieldArea`/`.fieldAttach` sit outside the `@media (max-width: 899px)` block and the JSX move applies at all widths, relocating desktop's attach button out of the controls row. The mobile symptom (paperclip stranded in the fixed `.actionBand`) can be fixed within the mobile breakpoint without touching desktop.

**5. Contradicts the issue's ask and misstates the composer.**
Issue #198 asks to "reuse the same UI as the composer for attachments and setting the model". The composer keeps its attach button in the controls row's `leading` slot (`Composer.tsx:1170-1181`), not overlaying the textarea — this PR invents a third placement while the PR body and the CSS comment claim it matches the composer. Model setting is untouched.

**6. Latent hazard if finding 1 is "fixed" naively:** enabling the overlay by making `IconButton` accept `className`, without adding matching padding to the textarea, lets typed text run underneath the paperclip in the field's lower-right corner.

Not held against the PR: the token-contract hit on the `(issue #198)` CSS comment fails on the branch head alone but resolves in the merge with current main (the `issue #`-prefix exemption in token-contract.test.ts), which CI's passing web-test confirms. The existing layoutguard suite (15 cases) and spawnguard (390/899/900px) pass on the branch.

obra:
## Rewritten from scratch against the six rejection findings

The previous head (`c901325a2`) is gone; this branch is three fresh commits on
`origin/main`. Every finding below is addressed by a different implementation,
not by patching the old one.

| # | Rejection finding | What this branch does |
|---|---|---|
| 1 | **Does not compile** — `IconButton` deliberately `Omit`s `className`, so `Spawn.tsx:755` failed TS2322 and took all four CI jobs with it. | No `className` is passed to any widget. The gating is a wrapper element the pane owns (`spawn.module.css`'s `.leading` / `.modelTrigger`), which is how `Composer.tsx` already does it. `npx tsc --noEmit --incremental false` is clean. |
| 2 | **The `.fieldAttach` overlay never rendered** — `IconButton` spreads `{...rest}` before its computed `className`, so the class was discarded and the CSS was dead code. | There is no overlay and no caller-supplied class anywhere. `PromptCard`'s `controlsClassName` prop is **deleted** — a caller reaching in to restyle the shared control row is the mechanism that produced the original bug, so it is gone rather than made to work. |
| 3 | **No layoutguard/geometry case** — the added test was jsdom ancestry only, which is exactly why finding 2 shipped unnoticed. | New `scripts/layoutguard/cases/spawn-attach-in-card/` at 390x844, plus a rewritten spawnguard contract at 390/899/900px. Red-first evidence quoted below. |
| 4 | **Not scoped to mobile; desktop layout changed too.** | Desktop is untouched. The configuration row's `ModelField` stays exactly where it was; the card's model trigger is `display: none` until the 899px block. spawnguard asserts that directly — the trigger must be visible at 390 and 899 and hidden at 900, and it fails either way round. |
| 5 | **Contradicts the issue and misstates the composer** — the composer keeps attach in the controls row's `leading` slot; the PR invented a third placement, and model setting was untouched. | Attach stays in `leading`. What changes is that the row is no longer turned into a fixed viewport band, so `leading` is *in the card* at every width, which is the composer's actual placement. Model setting is now covered too: `ModelSwitchTrigger` is extracted from the session's `ModelSwitch` and rendered by both surfaces, so it is literally the same component rather than a lookalike. |
| 6 | **Latent hazard**: enabling the overlay without matching textarea padding would let typed text run under the paperclip. | Structurally impossible now — attach is in the row beneath the field, and the layoutguard case asserts `attach.top >= field.bottom` so the overlay placement cannot come back silently. |

### Red-first evidence

The new layoutguard case was authored against **main's markup and main's CSS**
first, and run before any source change.

Run 1 — main's markup verbatim (`class="controls actionBand"`, attach as a
direct child, no model trigger):

```
spawn-attach-in-card ... FAIL - the harness is missing modelTrigger - nothing else can be measured without them
```

That is issue #198's second half stated as geometry: there is no way to set the
model from the card at all.

Run 2 — same main CSS, with the model trigger added to the harness so the
geometry claims are the only ones left to fail:

```
spawn-attach-in-card ... FAIL - the control row is position: fixed (0.0,767.0 390.0x77.0) - it is a viewport band, not the card's own row; the control row (0.0,767.0 390.0x77.0) is outside the prompt card (12.0,12.0 366.0x106.0); the attach button (16.0,784.0 44.0x44.0) is outside the prompt card (12.0,12.0 366.0x106.0); the model trigger (64.0,794.0 83.5x24.0) is outside the prompt card (12.0,12.0 366.0x106.0); the Start button (151.5,780.0 222.5x52.0) is outside the prompt card (12.0,12.0 366.0x106.0)
```

The card is 106px tall starting at y=12; the attach button is at y=784. That is
the bug, measured.

After the fix, with the harness markup updated to match:

```
spawn-attach-in-card ... PASS - at 390x844 the control row is static, and attach (21.0,117.0 44.0x44.0), the model trigger (69.0,127.0 83.5x24.0) and Start (338.0,117.0 31.0x44.0) all sit inside the card (12.0,12.0 366.0x154.0) below the field
```

The mutation evidence is recorded in `case.json` under `"mutation"`, as
`run.mjs:99-115` requires.

### The spawnguard contract is falsifiable too

A guard that has never failed is a decoration, so the new spawnguard assertions
were mutation-tested by hand: re-adding a `position: fixed` band for the control
row inside the 899px block, with the testid intact so nothing merely goes
missing.

```
390px ... FAIL
    the control row is position: fixed at 390px - a viewport band, not the card's row
    the control row (0.0,824.0 390.0x76.0) is outside the prompt card (16.0,73.8 358.0x154.0)
    the attach button (0.0,840.0 44.0x44.0) is outside the prompt card (16.0,73.8 358.0x154.0)
    the Start button (359.0,840.0 31.0x44.0) is outside the prompt card (16.0,73.8 358.0x154.0)
    the card's model trigger (48.0,850.0 90.8x24.0) is outside the prompt card (16.0,73.8 358.0x154.0)
899px ... FAIL
    (same five failures at 899px)
900px ... PASS
```

The mutation was reverted; the guard is green on the branch.

The jsdom tests were red-first as well: with the test edits in place and the
source changes stashed, `Spawn.test.tsx`, `MobileSettingRows.test.tsx` and
`promptcard.test.tsx` reported **18 failed / 59 passed**, including the three
new claims (attach and the model trigger inside the card; the trigger's label
transitions; no model row and no model sheet). All 77 pass with the fix.

### Decisions taken, so they are on the record rather than implied

- **Effort stays in `MobileSettingRows`.** Merging model and effort into one
  combined pill the way the session status row does is a product call, not part
  of this bug fix. Flagged and deliberately left out.
- **The model sheet's "Use default" row is dropped.** Strict composer reuse:
  the trigger's own resting label is `(default)`, which already says it.
- **Desktop is untouched**, including the `ModelField` row (finding 4).
- **No `<form>` and no `onSubmit` in Spawn** (kata `t13x`).
  `test/scenarios/spawn-keyboard-contract.md` Part A greps for exactly that and
  still finds nothing; that card's step 4 now filters `— change model` buttons
  by `offsetParent`, because the pane deliberately carries two of them and only
  one is ever on screen.

### Verification on this head

- `npx tsc --noEmit --incremental false` — clean
- `npm test` — 343 files, 6764 tests passed
- `npm run lint` (`biome ci src`) — clean
- `npm run layoutguard` — 17/17 PASS (16 existing + the new case)
- `npm run spawnguard` — 390/899/900px PASS
- `npm run overflowguard` — PASS
- `go test ./` — ok (the two repointed line citations in
  `spawn-empty-prompt-starts-dormant.md` and `web-file-picker-image.md` are the
  scenario source-cite audit's own finding: the `Spawn.tsx` edits moved the
  quoted literals two lines down)

obra:
### One thing I did not decide unilaterally: design-system §11 and the model sheet

`docs/web-ui/design-system.md` §11 ("Mobile forms and honest cold starts")
says:

> Mobile choice controls use the existing bottom `Sheet` pattern. […] A feature
> should reuse its existing catalog/path panel inside the sheet rather than
> create a second picker implementation.

This PR moves spawn's model choice from a bottom `Sheet` to the composer's
`Popover`-based `ModelSwitchTrigger`, because that is what issue #198 asks for
("reuse the same UI as the composer") and what the session pane already ships
on a phone. The rule's *intent* is satisfied either way — it exists to stop a
second picker implementation, and this deletes one rather than adding one — but
the rule's *letter* names the sheet, so the two now disagree for this one
control.

I have deliberately not rewritten the design-system rule to match my own
change. Options, in the order I'd rank them:

1. Amend §11 to say the sheet is the pattern for *option lists* (harness,
   effort, access mode — all of which still use it here) while a control the
   app already ships as a popover on another surface keeps that popover, so
   one act keeps one affordance.
2. Leave §11 alone and treat spawn's model trigger as a documented exception.
3. Revert this half and keep the model in a sheet — which puts #198's second
   half back on the shelf.

The other §11 clauses are unaffected: the 16px editable-text floor stays, the
44px/48px touch floors are now *measured* by spawnguard rather than asserted by
CSS, and "the mobile action band **may** be fixed above
`env(safe-area-inset-bottom)`" is permissive, so spawn no longer having a band
does not contradict it (and the "form body reserves the band's space" clause
simply stops applying).

obra:
## Confirmation review: CONFIRM

Verified independently in a fresh worktree, branch `dlg_0349wPG8h7pW3u5LYkXL39` fetched with `--no-tags` and checked out detached at `c09648e19` (four commits atop merge-base `0fe036a5c`). Base is `main`.

**1. Compile (the original rejection was here) — clean.**
`npx tsc --noEmit --incremental false` exits 0, no errors. `npm run lint` (`biome ci src`) also clean, 921 files.

**2. Full suites — all green.**
- `npm test`: 343 files, 6764 vitest tests + 35 node tests, all pass.
- `npm run layoutguard`: 17/17 PASS, including the new `spawn-attach-in-card` case: "at 390x844 the control row is static, and attach (21.0,117.0 44.0x44.0), the model trigger (69.0,127.0 83.5x24.0) and Start (338.0,117.0 31.0x44.0) all sit inside the card (12.0,12.0 366.0x154.0) below the field."
- `npm run spawnguard`: 390px / 899px / 900px all PASS.

**3. Mutation-tested the new guard — it's a real assertion, not decoration.**
Restored the old `.actionBand` (`position: fixed; bottom: 0`) in `spawn.module.css` inside the 899px block, added `actionBand` to the layoutguard harness's controls div (mirroring the case's own `case.json` mutation doc), and added a `[data-testid="spawn-controls"]` rule so the same regression reaches spawnguard's real component render (PromptCard tags the row via `controlsTestId`, not a class, post-refactor).
- `node scripts/layoutguard/run.mjs spawn-attach-in-card` → **FAIL**: "the control row is position: fixed (0.0,768.0 390.0x76.0) - it is a viewport band, not the card's own row; ... attach/model trigger/Start button outside the prompt card."
- `npm run spawnguard` → **FAIL** at 390px and 899px (control row fixed, attach/Start/model trigger all outside the card); 900px still PASS (correctly outside the mobile media query).
- `git checkout --` reverted both files; `git status --short` clean; re-ran both guards → 17/17 and 390/899/900 all PASS again.

**4. Issue #198 conformance — confirmed structurally, not just by claim.**
`Spawn.tsx`'s `PromptCard leading={...}` renders `<div className={CLASS.leading}><IconButton data-testid="spawn-attach".../><span className={CLASS.modelTrigger}><ModelSwitchTrigger .../></span></div>` — same shape as `Composer.tsx:1170-1183`'s `<div className={CLASS.leading}>` (attach + model control). Desktop: `.cfgModel`/`ModelField` row is untouched in the `.cfg` block below the card; `.modelTrigger` is `display: none` by default and only `display: flex` inside `@media (max-width: 899px)`, so the in-card trigger is desktop-hidden. `grep -rn "<form\|onSubmit" src/panes/spawn/*.tsx` → no matches (the one CSS-comment hit in `advancedOptions.module.css` refers to an unrelated `CollectionEditor` form, not Spawn's). `grep -rn "actionBand|controlsClassName" src` → no matches anywhere; fully deleted, not just unused. `MobileSettingRows.tsx` row count: 6 → 5 confirmed by diff (Model row + its Sheet + "Use default" option removed entirely).

**5. Citation fixes and the §11 flag.**
Both scenario-card diffs in commit `1563cd5af` are pure line-number repoints: `spawn-empty-prompt-starts-dormant.md` `Spawn.tsx:729`→`:730` (verified current line 730 is the placeholder text), `web-file-picker-image.md` `Spawn.tsx:613`→`:615` (verified current line 615 is `toasts.push("error", "Image attachment is still processing.")`). No content changes beyond the numbers.
The design-system §11 tension (bottom-Sheet-for-mobile-choice vs. this PR's Popover-based `ModelSwitchTrigger`) is explicitly flagged in a separate PR comment (obra, 2026-08-19T17:35:11Z) with three ranked options and no unilateral rewrite of the rule. Coordinator-level call, correctly left open rather than relitigated here.

**6. CI and base.**
`gh pr checks 242`: `Build darwin-arm64` pass, `Build linux-amd64` pass, `build-and-test` pass, `web` pass; the two `skipping` jobs are release/snapshot asset publishing, not test gates. `baseRefName` is `main`, state `OPEN`.

All six checks pass. No merge performed.

PR #246, opened 2026-08-19
feat(agent): in-stream loop guard, Phase 1 (#94)

obra:
Implements the Phase-1 spec from the "v2 spec: two independently-landable phases" comment on #94. Supersedes #130 and #241; neither is closed by this PR.

Nothing in evener bounds a single streamed response. All four existing runaway guards watch either *between* responses or for *silence*, and a runaway is loud and never completes a round: `--max-rounds` cannot advance while one response is still streaming, `PhaseSilentStall` requires zero content events, `RetryStream`'s cap rule only evaluates on `err != nil` and a runaway ends with an ordinary `tool_calls` finish, and the tool breaker is a post-dispatch write that never runs during the response itself. This adds the missing bound inside `consumeModelStream`'s event loop.

## Conformance to the #94 Phase-1 spec

| Spec item | Where |
|---|---|
| Take `session_stream_loop_guard.go` whole, minus `loopTripChant` | `agent/session_stream_loop_guard.go` |
| `loopGuardCycleMaxLen=5`, `loopGuardCycleRepeats=5`, `loopGuardRawCeiling=50` | same file |
| `llm.NewPermanentStreamError` after `NewStreamError` | `llm/sdk_errors.go` |
| `SteeringKindStreamLoop` + `AllSteeringKinds` + test map | `agent/events/payloads.go`, `payloads_test.go` |
| Regenerated `types.gen.ts` committed (Makefile staleness lint) | `cmd/evener-hub/frontend/src/protocol/types.gen.ts` |
| `"stream-loop": "Stream loop guard"` frontend label | `SteeringItem.tsx` |
| Session fields beside `loopDetectionCount` | `agent/session.go` |
| `consumeModelStream` wiring: guard state, `openToolCalls`, `applyPendingTrip`, `streamEvents:` label, `hardStopErr` early return, streak-reset tail | `agent/session_stream.go` |
| Delivery in `injectPostToolSteering` before `drainPostToolWatchSends` | `agent/session_tool_round.go` |
| Chant detector and all chant hunks dropped | not ported |
| `session_lifecycle.go` zero-calls delivery hook NOT ported (dead in Phase 1) | not ported |
| `nonChantingFiller` / `testkit_test.go` edits not ported | not ported |
| `describeCyclePattern` uses `strings.Cut` (modernize) | `session_stream_loop_guard.go` |
| Ported tests (6 integration + 2 delivery + the `checkCycle` unit/fixture files) | `*_integration_test.go`, `*_delivery_test.go`, `*_test.go`, `*_fixtures_test.go` |
| New tests, minimum 4 | see below |

Ported from #130's judged-sound half, commit e1fee75b3 on `kata/uq-d74b` — the half that review found "no false positive" against. Hand-ported at tip state with `primeradiant.com/serf` → `primeradiant.com/evener` rewrites; no cherry-pick (the module and `cmd/serf-hub` renames landed after the branch point).

## The two mandatory defect fixes

**(a) The tier-2 hard stop was fallback-eligible.** `NewPermanentStreamError` classifies `ErrorClassPermanent`, and `modelFallbackEligible` returns `true` for permanent — so a hard stop with `ModelFallbacks` configured re-ran the exact runaway-producing request against every fallback model, the opposite of what the constructor exists to do. Fixed with a `streamLoopHardStopError` wrapper excluded by `errors.As` before the `Classify` switch in `agent/session_init.go`, mirroring the `*llm.ProviderUnhealthyError` arm three lines above. Covered by `TestModelFallbackEligible_StreamLoopHardStop` (unit) and `TestProcessInput_StreamLoopGuard_HardStopSkipsModelFallbacks` (end to end, asserting `a.Models()` never leaves `primary`).

**(b) The streak and pending nudge leaked across turns.** Only a clean completion zeroed the streak, and a tier-2 hard stop returns before that runs — so the streak stayed at 2 forever, and every later trip in every later turn hard-stopped instantly with no tier-1 chance to self-correct. `processOneInput` now zeroes both fields at turn start, inside the `s.mu` block it already holds. The pending nudge is dropped rather than carried: it describes a response from a turn that is over, and delivering it into an unrelated one is the "stale nudge lands out of context" failure the review flagged. Covered by `TestProcessInput_StreamLoopGuard_TurnStartResetsStreakAndNudge` (reads both fields at round-loop entry, from inside the adapter script) and `TestProcessInput_StreamLoopGuard_HardStopEndsTurnAndSessionSurvives`.

## New tests

Every test on `kata/uq-d74b` drove `consumeModelStream` directly, so nothing proved the calls streamed before a trip actually dispatch, that the nudge reaches history, that the next request carries it, or that a session survives a hard stop. Five new ones, all through `ProcessInput`:

- `TestProcessInput_StreamLoopGuard_TripDispatchesNudgesAndRecovers` — round 1 runs away; asserts exactly 15 dispatched tool results (not 20, the fixture's full send), a `SteeringKindStreamLoop` turn positioned after those results, the nudge naming the repeated calls, the *second* request carrying that exact text, round 2 completing cleanly, and the streak back to 0.
- `TestProcessInput_StreamLoopGuard_HardStopEndsTurnAndSessionSurvives` — two consecutive tripping rounds in one turn surface the hard stop out of `ProcessInput` itself; the *next* turn opens with another tripping round and must recover on a tier-1 nudge, which is only possible if the streak was zeroed at turn start.
- `TestProcessInput_StreamLoopGuard_TurnStartResetsStreakAndNudge`
- `TestProcessInput_StreamLoopGuard_HardStopSkipsModelFallbacks`
- `TestModelFallbackEligible_StreamLoopHardStop`

Plus three beyond the spec, each closing a hole the mutation run exposed or the suite left open:

- `TestStreamLoopGuard_FortyDistinctCalls_NoTrip` — 40 distinct signatures pass untouched, so the ceiling's margin over odysseus #3185's 18 is asserted rather than assumed.
- `TestConsumeModelStream_LoopGuard_StateDoesNotCarryAcrossResponses` — two 30-call rounds in one session both complete; a session-scoped guard would trip the ceiling partway through the second, re-inventing the between-responses blindness this exists to fix.
- `TestStreamLoopGuard_SixCycle_FallsToTheCeilingNotTheCycleDetector` — pins `loopGuardCycleMaxLen`'s upper edge. Widening it 5→6 was the one mutation the suite failed to catch: every fixture uses a k=3 cycle, which the k=3 arm finds at 15 calls no matter how far the scan widens.

## Evidence

**Red first.** `2846aa6d0` lands the tests against stub declarations; 15 tests fail. `153af9b04` implements; all green. The negative controls (`EighteenDistinctCalls_NoTrip` at both levels, `FortyDistinctCalls_NoTrip`, `DoesNotTripBeforeFiveRepeats`) pass in the red state by construction — they exist to catch a detector that trips too eagerly, not one that never trips.

**Mutation checklist**, every item from the spec, applied to a clean tree and reverted (script drove it, output verbatim):

| Mutation | Red |
|---|---|
| `loopGuardCycleRepeats` 5→6 | `CycleShape`, `TripsAtFifteenCalls`, `SingleCallRepeatedFiveTimes`, `CleanRoundResetsStreak`, `DefersUntilParallelCallCloses`, `SecondConsecutiveTrip_HardStops`, `TripDispatchesNudgesAndRecovers` |
| `loopGuardCycleMaxLen` 5→6 | `SixCycle_FallsToTheCeilingNotTheCycleDetector` |
| `loopGuardCycleMaxLen` 5→2 | `TripsAtFifteenCalls`, `CycleShape`, + 6 more |
| `loopGuardRawCeiling` 50→51 | `CeilingShape` |
| `loopGuardRawCeiling` 50→49 | `CeilingShape` |
| `loopGuardRawCeiling` 50→18 | `EighteenDistinctCalls_NoTrip` (both levels), `FortyDistinctCalls_NoTrip`, `StateDoesNotCarryAcrossResponses`, `CeilingShape` |
| remove `len(openToolCalls) > 0` | `DefersUntilParallelCallCloses` |
| remove the streak increment | `SecondConsecutiveTrip_HardStops`, `CycleShape`, `HardStopEndsTurnAndSessionSurvives`, `HardStopSkipsModelFallbacks` |
| disable the `streak >= 2` branch | `SecondConsecutiveTrip_HardStops`, `HardStopEndsTurnAndSessionSurvives`, `HardStopSkipsModelFallbacks` |
| disable the clean-completion reset | `CleanRoundResetsStreak`, `HardStopEndsTurnAndSessionSurvives`, `TripDispatchesNudgesAndRecovers` |
| `NewPermanentStreamError` → `NewStreamError` (retryable) | `SecondConsecutiveTrip_HardStops`, `HardStopEndsTurnAndSessionSurvives` |
| remove the `errors.As(err, &loopStop)` exclusion | `ModelFallbackEligible_StreamLoopHardStop` (both cases), `HardStopSkipsModelFallbacks` |
| remove the turn-start reset | `TurnStartResetsStreakAndNudge`, `HardStopEndsTurnAndSessionSurvives` |
| remove the `injectPostToolSteering` delivery call | `InjectPostToolSteering_DeliversStreamLoopNudge`, `TripDispatchesNudgesAndRecovers` |
| make the guard state session-global | `StateDoesNotCarryAcrossResponses` |

The `loopGuardCycleMaxLen` 5→6 row is the honest one: it survived the ported suite, which is why `SixCycle_FallsToTheCeilingNotTheCycleDetector` exists. Widening the scan cannot make a k=3 cycle trip later, so no existing fixture could see it.

**Docs.** `docs/architecture.md`'s "Steering a stuck agent has two layers" went stale on landing; the section now describes all three and says why the older two are blind to this failure.

**Suites.** `go test ./...` and `go test -race ./agent/... ./llm/...` both clean. `make lint` (naming, gofmt, evenerfuzz, eval, internal, docs, golangci, generated, secret-scan) passes. `make test-web` passes typecheck, unit, and lint — `KIND_LABELS` is an exhaustive `Record` over the generated union, so the frontend build is what enforces the new kind having a label.

## Not in this PR

Phase 2's text-runaway (chant) detector, which the spec says needs a rebuilt algorithm rather than the v1 port: an amortized rolling-hash index, a two-tier entropy bar, buffer-based fence tracking, and a twelve-fixture negative suite. The `session_lifecycle.go` zero-tool-calls delivery hook and its test come back with it — both Phase-1 detectors key off `StreamEventToolCallEnd`, so a trip always has tool calls and that hook is unreachable until chant returns.

Also noted, not fixed: `internal/appprojector/appwire_projection.go` labels every `EventLoopDetection` "Loop detection" in the transcript bubble, while `SteeringItem.tsx` distinguishes `stream-loop` → "Stream loop guard". Same nudge, two labels depending on the surface. Pre-existing shape, flagged so it is not relitigated as new.

Closes nothing; #94 stays open for Phase 2.

### DIFF
--- a/agent/events/payloads.go
+++ b/agent/events/payloads.go
@@ -311,6 +311,7 @@ const (
 	SteeringKindImageDescription  = "image-description"
 	SteeringKindNoToolCalls       = "no-tool-calls"
 	SteeringKindLoopDetected      = "loop-detected"
+	SteeringKindStreamLoop        = "stream-loop"
 	SteeringKindTasksDone         = "tasks-done"
 	SteeringKindTaskNudge         = "task-nudge"
 	SteeringKindTaskInactive      = "task-inactive"
@@ -336,6 +337,7 @@ var AllSteeringKinds = []string{
 	SteeringKindImageDescription,
 	SteeringKindNoToolCalls,
 	SteeringKindLoopDetected,
+	SteeringKindStreamLoop,
 	SteeringKindTasksDone,
 	SteeringKindTaskNudge,
 	SteeringKindTaskInactive,
--- a/agent/events/payloads_test.go
+++ b/agent/events/payloads_test.go
@@ -101,6 +101,7 @@ func TestSteeringKindConstants(t *testing.T) {
 		"image-description":  SteeringKindImageDescription,
 		"no-tool-calls":      SteeringKindNoToolCalls,
 		"loop-detected":      SteeringKindLoopDetected,
+		"stream-loop":        SteeringKindStreamLoop,
 		"tasks-done":         SteeringKindTasksDone,
 		"task-nudge":         SteeringKindTaskNudge,
 		"task-inactive":      SteeringKindTaskInactive,
--- a/agent/session.go
+++ b/agent/session.go
@@ -578,6 +578,21 @@ type Session struct {
 	// stuck detection
 	loopDetectionCount int // how many times loop detection has fired
 
+	// stream loop guard (issue #94): tracks whether the response
+	// consumeModelStream most recently returned was force-stopped mid-stream
+	// by the in-band loop guard (session_stream_loop_guard.go), so a SECOND
+	// consecutive trip escalates to a hard stop instead of nudging forever.
+	// streamLoopTripStreak resets to 0 on any response that completes
+	// without tripping, and at the start of every turn (processOneInput), so
+	// a hard stop -- which returns before the clean-completion reset can run
+	// -- cannot leave the session permanently one trip from another hard
+	// stop. pendingStreamLoopNudge holds the steering text for a tier-1 trip
+	// until injectPostToolSteering delivers it as a steering turn and clears
+	// it; it is dropped rather than carried at turn start, since a nudge that
+	// missed its own turn would land out of context in an unrelated one.
+	streamLoopTripStreak   int
+	pendingStreamLoopNudge string
+
 	// transcript writer (nil when StateDir is empty, or when opening it failed)
 	transcript *transcript.Writer
 	// attentionMu serializes transcript-backed attention resolution for this
--- a/agent/session_init.go
+++ b/agent/session_init.go
@@ -1282,6 +1282,15 @@ func modelFallbackEligible(err error, policy llm.RetryPolicy) bool {
 	if errors.As(err, &unhealthy) {
 		return false
 	}
+	// A loop-guard hard stop indicts the request, not the model: the runaway
+	// came from what was asked, so every fallback entry would reproduce it.
+	// Decided here for the same reason as the verdict above — the wrapped
+	// cause classifies ErrorClassPermanent, which the switch below reports
+	// eligible.
+	var loopStop *streamLoopHardStopError
+	if errors.As(err, &loopStop) {
+		return false
+	}
 	switch llm.Classify(err) {
 	case llm.ErrorClassPermanent, llm.ErrorClassFallback:
 		return true
--- a/agent/session_lifecycle.go
+++ b/agent/session_lifecycle.go
@@ -1037,6 +1037,15 @@ func (s *Session) processOneInput(ctx context.Context, input string, images []Im
 	// beside comm's reset, under the lock already held (not via the
 	// clearAskPending helper, which takes s.mu itself and would deadlock).
 	s.askPending = nil
+	// Stream loop guard (issue #94): both fields are scoped to a turn, and
+	// only the clean-completion path in consumeModelStream zeroes the streak
+	// — which a tier-2 hard stop returns before reaching, leaving the streak
+	// at 2 so every later trip, in every later turn, hard-stops instantly
+	// with no tier-1 chance to self-correct. The pending nudge is dropped
+	// rather than carried: it describes a response from a turn that is over,
+	// and delivering it into an unrelated one lands it out of context.
+	s.streamLoopTripStreak = 0
+	s.pendingStreamLoopNudge = ""
 	s.mu.Unlock()
 	s.delegateDeliveryMu.Unlock()
 
--- a/agent/session_stream.go
+++ b/agent/session_stream.go
@@ -242,6 +242,66 @@ func (s *Session) consumeModelStream(ctx context.Context, req llm.Request, st ll
 	assistantStarted := false
 	finished := false
 
+	// Stream loop guard (issue #94): bounds this one response, since nothing
+	// else in the stack does — see session_stream_loop_guard.go for the
+	// detectors. guardTripped and hardStopErr are read after the loop to
+	// decide the streak reset and the tier-2 hard-stop return respectively.
+	loopGuard := newStreamLoopGuard()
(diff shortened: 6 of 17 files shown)


obra:
CI note: the first `build-and-test` run failed on `TestNpmShimRejectsUnsupportedCommand` — `fork/exec ...: text file busy` in the root package's install tests. Unrelated to this diff (which touches only `agent/`, `llm/`, `docs/`, and the frontend) and pre-existing: it is the golang/go#22315 `ETXTBSY` race, where an unrelated parallel test's forked child briefly holds a writable fd to the shim this test just wrote. Filed as #248 with the diagnosis and a suggested fix; job re-run here.

Everything else was green on that run, including `make lint` and the full `-race` gate.

obra:
Let's hold on merging this until we ever see it in the field again.

obra:
Decision: **close/defer**, rather than merge now.

The implementation and old-head CI are solid, but this was explicitly put on hold until the failure recurs. It remains a draft with no reviews and is now 145 commits behind main. If field evidence arrives, reopen or reapply the seven focused commits against current main with that reproduction as the acceptance case.

Issue #182, opened 2026-08-18
cmd/serf-dev/agentshards.go breaks the GOOS=windows build


obra:
GOOS=windows go build ./cmd/serf-dev/... fails on agentshards.go: syscall.Kill and Setpgid are unix-only and the file has no build tag. internal/devtool/... (including procgroup, which is properly split into pid_unix.go/pid_other.go) builds clean on Windows.

Pre-existing before PR #132 (confirmed identical failure at its merge-base). Fix: build-tag the unix-only parts of agentshards.go the same way procgroup does.

---
Filed from the adversarial review session: https://claude.ai/code/session_0134LFae5DzpDExkuhJTCNfU

PR #247, opened 2026-08-19
Go dev tooling, done properly: supersedes #222 and #226

obra:
Supersedes #222 and #226. Both are REJECTED; this branch starts from main and
cherry-picks only the pieces of that work an adversarial review verified good.
Leave #222/#226 open until this lands.

## What this actually is

`internal/devtool/{covstmt,gatesurface,capabilityprobe}` plus two evener-dev
subcommands (`capability-preflight`, `coverage-floor`), wired into the Makefile.
Nothing else moved. Every script the rejected branches deleted is still here,
still running, still pinned by its own selftest: `run-module-tests.sh` and its
115-check selftest, the dev-tooling wave runner, `test-timing-budget` +
`testing-budget.json`, `coverage-union.sh`, `coverage-gaps.sh`,
`fuzz-coverage-global.sh`, `gate-surface-lib.sh`, `covstmt-lib.sh`, every
surviving tool's selftest, main's CI workflow, and main's floors.

## The seven must-fixes

| # | Reviewer finding (#222 / #226) | Resolution here |
|---|---|---|
| 1 | `make test`/`test-race`/`test-full` ran ZERO tests: the Makefile grew its own copy of the `-run` pattern with BRE escaping (backslash-paren, backslash-pipe), which RE2 reads as literal characters, so every package reported `[no tests to run]` under a green exit. No drift tripwire. | The Makefile never grows a copy: `run-module-tests.sh` stays the runner and keeps sourcing `gate-surface-lib.sh`. `internal/devtool/gatesurface/drift_test.go` pins the shell values byte-equal to the Go constants and RE2-compiles them against real test names (mutation-verified: re-introducing the BRE escaping turns the tripwire red). On top of that, `run-module-tests.sh` now FAILS a run whose Go waves executed zero tests, so this class of bug cannot be green again. |
| 2 | Capability preflight wired fail-open (`export VAR=$$(...)` discards the substitution's status) — regressed issue #181, which PR #224 exists to fix. | `merge-approval-gate` captures the verdict and exports it as SEPARATE statements, with the assignment's own status tested by an `if`. PR #224's regression net is kept and re-aimed at the Go command; the case was verified RED against the broken `export VAR=$$(...)` shape before the fix went in. |
| 3 | The skip splice never expanded (single quotes), and the design regressed by unioning the skip into EVERY module (`agent/session_escalation_e2e_test.go` has unrelated `TestE2E_*` tests). | Resolved by keeping `run-module-tests.sh`, which already applies the pattern to the ROOT module only and says so on its own output when it narrows the surface. |
| 4 | `coverage-floor` could not measure anything: no default modules, module name passed as an import path, no per-module cwd, no whole-module `-coverpkg`, no gate `-run`/`-skip`, no `-short`-except-root, no `-count=1`, no retained log. All 9 tests stubbed the load-bearing seam. | Fixed, each with a test: the shell's seven default modules and default floor file, `--tolerance`, `-coverpkg` from `go list ./...` per module (run from inside the module), `-short` on every module except root, the gate's `-run`/`-skip`, `-count=1`, a retained log whose path is printed, `(no module)`/`no statements` reported as unmeasured rather than 0.0%, and a `--bless` that appends newly measured modules and refuses to exit 0 on a run that measured nothing or lost a module. `TestCoverageFloorEndToEnd` drives the REAL `go test` path against a real throwaway module (50.0% by construction) with the mutation both ways. |
| 5 | `coverage-gaps` replacement inverted its purpose (ranked the best-covered functions by percentage). | Resolved by keeping the shell `coverage-gaps.sh`, which ranks by uncovered statement COUNT. Not ported. |
| 6 | `capability-preflight -only=<id>` was advertised by every rerun string the tool prints and silently ignored. | Implemented, with tests: it filters the summary and the skip regex, and exits 2 on an unknown id. The gate selftest asserts the exact advertised command. |
| 7 | The new CI step was a bash syntax error (`$$(...)` is Makefile escaping) and could never run. | No CI workflow change at all. `.github/workflows/ci.yml` is byte-identical to main's. |

## Deliberately scaled back

- **fuzz-coverage-global, coverage-union, the module-tests runner, test-timing-budget stay shell.** PR #103's spec lists the coverage and fuzz families as port-on-touch non-goals, and #222 destroyed their distinct semantics (registered-target plan validation, per-target replay accounting, the union of two tracks) by collapsing them onto one generic measurement.
- **Deleting `test-timing-budget` is a policy decision, not a refactor.** It stays, with `testing-budget.json`, for Jesse to rule on separately.
- **Selftests of surviving tools are untouched.** Only the two selftests whose subject scripts this PR actually deletes are removed.

## Retired in this PR

`scripts/gate/gate-capability-preflight.sh`, `cmd/evener-gate-probe/`,
`scripts/coverage/test-coverage-floor.sh` and its selftest — each replaced by a
Go subcommand whose tests carry the same contract, including the shell
selftest's own cases. No other script leaves the tree.

## Floors

`scripts/coverage/testcov-global-floors.txt` data rows are byte-identical to
main's. Only the "Managed by" header line changed.

## Verification

**Test counts, since "green" was the thing that lied last time.** Replaying the
gate's own per-module selection and keeping the raw output:

```
ok lines:            139
[no tests to run]:     8
[no test files]:       0
packages that ran tests: 131
```

The eight are packages whose only tests are fuzz-designated and therefore
excluded by the gate's own `-skip` (`cmd/evener-fuzzregistry`,
`cmd/evener-hub/internal/{fspaths,hubedge}`, `internal/fuzzroutes`,
`internal/httpguard`, `test/e2e/fakellm`, `agent/internal/promptpath`,
`llm/providers/kimicoding`). #222's number for the same measurement was 137
no-tests packages — every one of them.

- `make test`: PASS on all 7 Go modules + web, exit 0.
- `make test-race`: PASS on all 7 Go modules, exit 0.
- `make build-all`: green.
- `make test-dev-tooling`: the whole wave, 15 suites PASS, 0 FAIL.
- `go run ./cmd/evener-dev coverage-floor --check` over all six floored modules,
  for real, exit 0:

```
MODULE      COVERED  TOTAL  PCT    FLOOR  STATUS
.           27669    34449  80.3   80.0   PASS
agent       35080    39892  87.9   87.9   PASS
llm         7908     8772   90.2   90.0   PASS
auth        598      631    94.8   94.9   PASS
envvars     50       50     100.0  100.0  PASS
invariant   -        -      -      -      no statements
identifier  345      363    95.0   95.0   PASS
```

  `auth` at 94.8 against a 94.9 floor is inside the 0.5pp tolerance band, which
  is the shell script's default and unchanged here. `invariant` is reported as
  unmeasured rather than 0.0% — `Hold` compiles to an empty function in a
  production build, so the module has no statements and no floor is definable.
- `make lint`: PASS, 7 modules. (golangci-lint had 36 findings against the
  salvaged and new Go files on the first run — unchecked writer/Close returns,
  `exec.Command` under noctx, two modernize hits — all fixed in cc03d459b.)
- Drift tripwire, mutation-verified both ways: BRE-escaping
  `gatesurface.TestRun` fails 3 tests (the shell/Go equality pin and both
  RE2-compile checks); drifting only the shell copy fails the equality pin.
- `scripts/gate/run-module-tests-selftest.sh`: 115 checks, 0 failed. The
  zero-test tripwire was verified RED first (defeating the detection turns 3
  checks red).
- `scripts/gate/merge-approval-gate-selftest.sh`: 32 checks, 0 failed. The
  issue-#181 case was verified RED against the `export VAR=$$(...)` shape.
- `go test ./cmd/evener-dev -run 'CoverageFloor|CovFloor'`: 34 cases including
  the real-toolchain end-to-end and both mutations.
- `TestCrossCompileWindows` (issue #182): PASS.
- `gofmt -l`: clean. `git diff --check`: clean.
- `.github/workflows/` diff vs main: empty.


obra:
CI's `build-and-test` failed once on an unrelated flake, now filed as #249: `TestE2E_SteerLandsInTheNextTurnWhenItsTurnEnded` failed inside `t.TempDir`'s RemoveAll (`state/evener/projects/<id>/sessions: directory not empty`), not on any of its own assertions. Same species as the marketplace teardown race #235 fixed elsewhere, different writer. This PR's diff contains no product code. Re-running the job.

obra:
Second CI attempt failed on a *different* known flake: #245 (`fuzz-bisect-selftest`: "bisect did not converge"). The attempt-1 failure (#249, the TempDir cleanup race) did not recur — that attempt's deterministic-tests step passed. Neither is in this PR's diff, which contains no product code. Third attempt running.

Locally on this branch: `make test` green (7 modules + web), `make test-race` green, `make test-dev-tooling` 15/15 suites green, `make lint` green, `make build-all` green.

obra:
## CI status: three attempts, four distinct flakes, none in this diff

Honest summary rather than "it's green": `build-and-test` has not passed yet, and I do not believe any of it is this PR.

| Attempt | Failure | Verdict |
|---|---|---|
| 1 | `TestE2E_SteerLandsInTheNextTurnWhenItsTurnEnded` — `t.TempDir` RemoveAll: `.../projects/<id>/sessions: directory not empty` | filed #249 |
| 2 | `fuzz/fuzz-bisect` selftest: "bisect did not converge" | known #245 |
| 3 | `TestServeModelSwitch_ThreadReadReflectsNewModelWithNoInterveningTurn` — premise read returns an empty ModelProvider | filed #251 |
| 3 | `TestE2E_StopIsOfferedWheneverTheWireSaysActive` — `t.TempDir` RemoveAll: `.../projects/<id>: directory not empty` | same family as #249, noted there |

Four different tests, no repeat except the two #249-family ones, and every one of them is product/e2e or fuzz-tooling code that this PR does not touch. This PR's diff is the Makefile, scripts, docs, `cmd/evener-dev`, `internal/devtool`, and two audit-test allowlists.

**main is failing the same way right now**: run 32282746454 on main (17:38Z, before this PR's first attempt) failed on #245, the identical `fuzz-bisect` non-convergence. Two of main's last eight CI runs failed.

Locally on this branch, all of it is green:

- `make test`: PASS on 7 Go modules + web
- `make test-race`: PASS on 7 Go modules
- `make test-dev-tooling`: 15/15 suites PASS (including `fuzz/fuzz-bisect`)
- `make lint`: PASS, 7 modules
- `make build-all`: green
- `go run ./cmd/evener-dev coverage-floor --check`: all six floored modules pass
- 131 of 139 packages executed tests (8 `[no tests to run]`, all fuzz-owned)

I have stopped re-running rather than burn CI minutes chasing a green that depends on four unrelated races. Worth deciding whether the root e2e flakes (#249, #251) and #245 get fixed before this merges, or whether this merges on the local evidence plus the two clean non-`build-and-test` jobs (`web`, both `Build` jobs, all pass).

obra:
## Adversarial re-review: MERGE AFTER FIXES

Verified on `devtool-go-rework` @ cc03d459b, every finding reproduced by execution rather than reading. All seven must-fixes from the #222/#226 rejections are genuinely fixed; I mutation-tested each tripwire in both directions. Two deficiencies I found are fixed and pushed (341bc41aa). One regression remains for you to decide on.

### The zero-test showstopper is fixed, and the tripwire is real

Replaying the gate's own per-module selection and keeping the raw output:

```
ok lines:                137
[no tests to run]:         8
[no test files]:           0
packages that RAN tests: 129
FAIL:                      0
```

The eight are fuzz-only packages the gate's own `-skip` excludes; I checked each (`internal/httpguard` has only `FuzzPolicy`/`FuzzHubTokenAuthorized`, `cmd/evener-hub/internal/fspaths` only `Fuzz*`, etc.). #222's number for the same measurement was 137 packages running NOTHING.

`make test`: PASS on 7 modules + web, exit 0. `make test-race`: PASS on 7 modules, exit 0.

The tripwire holds against the REAL failure mode, not just a fixture. Re-introducing #222's exact BRE pattern in `gate-surface-lib.sh` and running the gate:

```
$ MODULES="envvars identifier auth" WEB=0 scripts/gate/run-module-tests.sh -short -count=1
PASS  envvars  0.32s
PASS  identifier 0.42s
PASS  auth     0.81s

run-module-tests.sh: the Go waves ran zero tests, so this run proves nothing.
run-module-tests.sh: no scheduled module reported a package that executed any test.
full logs: /private/var/.../evener-module-tests.eZ9LgH
EXIT=1
```

Retained logs confirm every line was `[no tests to run]`. Same modules with the correct pattern: exit 0, so it is not a false positive. Defeating the tripwire in the runner turns 3 selftest checks red. `run-module-tests-selftest.sh`: 115 checks, 0 failed.

### Drift tripwire: mutation-verified both ways

- BRE-escape `gatesurface.TestRun` -> 3 tests fail (the equality pin, both RE2-compile checks, and the existing `TestTestRunIsTestExample`).
- BRE-escape only the shell copy -> the equality pin fails.

You cannot make the gate a silent no-op from either home without a red test. `ExampleTestRun_breEscapingMatchesNothing` keeps the tripwire's premise executable.

### Fail-closed preflight: correct, and now actually pinned

No `export VAR=$$(...)` shape anywhere in the Makefile (a grep of every `export X=` line finds only the explanatory comment). The recipe's `if ! VAR="$$(...)"` tests the assignment's own status. Re-introducing the broken shape turns the selftest's issue-#181 case red, with the gate having run `lint/build/test/test-dev-tooling`. `merge-approval-gate-selftest.sh` green.

**But I found the other half of #222's showstopper 2 unguarded.** Deleting the recipe's `export EVENER_GATE_CAPABILITY_SKIP` line — the verdict is computed and then never reaches any phase, which is precisely what #222 shipped — left all 32 checks GREEN. The gate would print an accurate BLOCKED summary with exact rerun commands and then run the infeasible tests anyway. Fixed in 341bc41aa: the fake make now records what each phase inherited, with two cases (a blocked capability's pattern reaches every phase; an all-available preflight exports an EMPTY skip, not an unset one — `run-module-tests.sh` runs under `set -u`). Verified red against the missing `export`, then green. 34 checks, 0 failed.

PR #224's regression net survives in adapted form. The three fail-open triggers collapse to one because "script missing" and "script not executable" have no analogue for a Go subcommand; the surviving trigger drives the REAL Makefile recipe and asserts `calls` is empty.

### coverage-floor measures honestly

`go run ./cmd/evener-dev coverage-floor --check`, no args, all seven default modules, exit 0:

```
MODULE      COVERED  TOTAL  PCT    FLOOR  STATUS
.           27664    34447  80.3   80.0   PASS
agent       35084    39892  87.9   87.9   PASS
llm         7908     8772   90.2   90.0   PASS
auth        600      631    95.1   94.9   PASS
envvars     50       50     100.0  100.0  PASS
invariant   -        -      -      -      no statements
identifier  345      363    95.0   95.0   PASS
```

Floors data rows byte-identical to main's (the diff is the "Managed by" header line only). Mutations:

- `identifier` floor raised to 99.0 -> `DROP`, `REGRESSION: identifier test coverage 95.0% < floor 99.0%`, exit 1.
- `--bless` on a run that measured nothing -> `--bless measured nothing; leaving <file> alone`, exit 1, floors md5 unchanged.

`TestCoverageFloorEndToEnd` is real: it drives `realCovFloorGoTest` (actual `go list` + `go test -coverpkg -coverprofile`) against a throwaway module that is 50.0% by construction, with the mutation both ways. `TestCovFloorGoTestArgs` pins the gate's `-run`/`-skip`, `-count=1`, module-scoped `-coverpkg`, and `-short`-except-root. The deleted shell selftest's assertions all have Go counterparts.

`-only=<id>` works and matches the advertised string exactly (`reprobe: go run ./cmd/evener-dev capability-preflight -only=loopback-bind`); an unknown id exits 2, and now before probing, where main's gate-probe probed first.

### Scale-backs and enforcement: nothing lost that I can find

`git diff main -- scripts/` deletes exactly three files: `gate-capability-preflight.sh`, `test-coverage-floor.sh`, `test-coverage-floor-selftest.sh`. `cmd/evener-gate-probe/` is deleted; its 16 tests appear one-for-one in `internal/devtool/capabilityprobe` (I diffed the function lists — identical). `fuzz-coverage-global.sh`, `coverage-union.sh`, `coverage-gaps.sh`, `run-module-tests.sh`, `cmd/evener-test-dev-tooling` (wave.go 329 + wave_test.go 567), `test-timing-budget.sh` + `testing-budget.json`, `covstmt-lib.sh` and every surviving selftest are intact. The `.github` diff is empty. No product code: the only root-module files touched are the two audit-test allowlists.

`make test-dev-tooling`: 15 suites, all PASS. `make build-all`: green. `make lint` (cache cleaned first): PASS, 7 modules, 71s. `gofmt -l`: clean. `git diff --check`: clean. `GOOS=windows CGO_ENABLED=0 go build ./cmd/evener-dev`: OK, and `TestCrossCompileWindows` passes for real.

### Fixed and pushed (341bc41aa)

1. The skip-propagation hole above.
2. **`evener-dev.exe` — a 4.5MB Windows PE32+ binary — was committed to the repo root** in 83d5c2af9. It is the default output name of the issue #182 cross-compile check run by hand from the repo root. Removed and gitignored.
3. `internal/devtool/covstmt`'s doc comment claimed `scripts/lib/covstmt-lib.sh` was "now deleted". It is not — `coverage-union.sh` and the fuzz family still source it, which is exactly why the two spellings have to agree.

### Must fix before merge: coverage-floor's abandoned scratch is never reclaimed

main's `test-coverage-floor.sh` called `reclaim_own_scratch "$tmpbase" evener-testcov` at the start of every run, so a previous run's retained-on-failure scratch was cleaned up on the next one. The port drops that entirely: `os.MkdirTemp(tmpdir, "covfloor-*")` with a random suffix, `keepScratch = true` on a failed measurement, and nothing anywhere that ever removes it.

This is not theoretical. On the machine I reviewed from:

```
/private/var/folders/.../T/covfloor-767696217   (created 10:13, still there at 11:19)
  agent.cov 70516344   root.cov 133276042   llm.cov 8951386  ... 16 files
```

~200MB, from this PR's own verification run six hours earlier, which nothing will ever reclaim. `scriptmktemp_audit_test.go`'s own header records the prior incident this pattern exists to prevent: "56 abandoned directories holding 1.5GB accumulated before anyone looked" — attributed to this very tool.

`covscratch-lib.sh`'s header was honestly updated to drop `test-coverage-floor.sh` from its list, so the loss is visible in the diff, but it is a loss. The remedy is already in the tree and needs no new design: `internal/devtool/scratch` is the Go twin of exactly this discipline — `Acquire(prefix, warn)` reclaims the prefix's abandoned pid-named scratch and mints a new one, `KeepOnFailure()` + `Release()` are the retain-and-report path, and the reclaim rules are the shell's with two safe tightenings. `coverageFloor` currently hand-rolls a worse version of that inline. Swapping to it also retires the `getenv` seam the config carries only for TMPDIR.

I did not make this change myself: it alters the tool's scratch contract (directory naming, the retained-logs message, reclaim-on-next-run) and its tests, which is your call rather than a reviewer's.

### Smaller notes, not blocking

- `internal/devtool/gatesurface/drift_test.go` cites `docs/superpowers/specs/2026-08-17-dev-tooling-in-go-design.md` as the requirement's source. That file is not in the tree — PR #103 is still open — so the citation is a forward reference to an unmerged doc.
- `covstmt` now has two homes (`covstmt-lib.sh` and the Go package) whose numbers are compared against the SAME floor files, with no tripwire between them. `gatesurface.go`'s own doc argues that a ratchet measuring a different surface than the gate proves is worse than no ratchet; the same argument applies to two statement counters. A byte-equal pin is impossible here (python-in-shell vs Go), but a differential case — run both over one fixture profile, require equal `(covered, total)` — would be cheap and would close it.
- `capabilityprobe.go`'s "Formerly cmd/evener-gate-probe; inlined into..." is history, which belongs in the commit message.
- Issue #180 (gate-probe: /tmp side effect, audit escape, `-only` semantics) stays open: the `/tmp/git-cache` creation side effect ports over unchanged. That is faithful rather than a regression, but the issue should not be auto-closed by this PR.

### CI

`build-and-test` failed on the pre-fix commit with two flakes, both already filed from this same PR's runs and both in packages this diff does not touch:

- `TestServeModelSwitch_ThreadReadReflectsNewModelWithNoInterveningTurn` — #251, verbatim.
- `TestE2E_StopIsOfferedWheneverTheWireSaysActive` — every assertion passed; only `t.TempDir` cleanup failed, on the `state/evener/projects/<project>` write race. Same species as #249 on a different test, so I commented there rather than filing a duplicate: the unjoined session-store writer is shared by the e2e hub-stack helper and any of those tests can lose the race.

main's own CI hit #245 (`bisect did not converge`) twice in the same hour, so the root suite is broadly flaky under load right now. No new unattributed flake.

### Verdict

MERGE AFTER FIXES. Every rejection finding is fixed, and each fix is mutation-proved rather than asserted. Fix the coverage-floor scratch reclaim — or rule that the leak is acceptable for a local-only tool and say so in the code — and this is good to land.

obra:
## Follow-up: the drift tripwire caught real drift on its first live merge

The `build-and-test` run on 341bc41aa went red, and not on a flake. The failing test was the drift tripwire itself:

```
--- FAIL: TestShellFuzzTestSkipMatchesGoConstant (0.00s)
    drift_test.go:50: gate surface drift:
      shell GATE_FUZZ_TEST_SKIP="(...|LineWindowExtractorsSanity|WriteListRoundTrip|...)"
      Go    gatesurface.FuzzTestSkip="(...|LineWindowExtractorsSanity|TranscriptReadersAgreeSanity|WriteListRoundTrip|...)"
      — update both homes together
```

Green locally, red on CI, because CI builds the merge with main and this branch forked before main's #244 (issue #170), which dropped `TranscriptReadersAgreeSanity` from `gate-surface-lib.sh` so the merge gate actually runs `TestTranscriptReadersAgreeSanity`. The merge takes main's shell line and keeps this branch's Go constant, so the merged tree had the two homes disagreeing about the gate's surface: `make test` would run that test while `coverage-floor` skipped it — a floor moving for a reason nobody could attribute, which is the exact failure `gatesurface`'s doc comment is written against.

This is the best possible evidence for the tripwire. It is not a hypothetical guard proved only by mutation; it caught a live, silent, cross-branch drift the first time one occurred, in the one place a human reviewer would never have looked.

Fixed in 19b416338 (on top of a merge of current main, 8346f2012): the name is dropped from `gatesurface.FuzzTestSkip`, so both homes match main's. Verified red before and green after, and re-verified against current main:

- `git diff main -- scripts/lib/gate-surface-lib.sh` is now comments only; both `GATE_TEST_RUN` and `GATE_FUZZ_TEST_SKIP` are byte-identical to main's.
- The deletions vs main are unchanged: `gate-capability-preflight.sh`, `test-coverage-floor.sh` + its selftest, `cmd/evener-gate-probe/`. `.github` diff still empty. Still no product code.

Everything in my earlier review stands. The one open item is still the coverage-floor scratch reclaim.

obra:
## CI after the drift fix (19b416338): the tripwire is green, the rest is flakes

The `-race` gate — the step that was red on 341bc41aa, on `TestShellFuzzTestSkipMatchesGoConstant` — now passes. So does build, vet, and lint:

```
✓ Build runtime pair and all Go workspace packages
✓ Vet (all go.work modules)
✓ Lint (make lint — every family of the local gate, all go.work modules)
✓ Test (race detector — the permanent -race gate, all go.work modules)
X Deterministic tests (full root suite; frontend covered by web job)
```

The remaining red is three flakes, none of which this diff can reach. All three are now filed with a diagnosis rather than a re-run:

| Test | Cause | Issue |
|---|---|---|
| `TestDelegateAttention_RestartRearmsColdChildAndDrainsExactAttention` | asserts `CurrentRunOpen` before synchronizing with the attention run | #239 (existing) |
| `TestMergedAncestryLocal_Unmerged_False` | fixture copy races git's `.git/objects/maintenance.lock`; the assertion never ran | #255 (new) |
| `web-layoutguard` | headless Chrome's DevTools endpoint never came up on the runner; the guard calls it an environment problem and fails the job anyway | #254 (new) |

#255 is the interesting one: `initRepo` copies a shared template repo with `os.CopyFS`, which walks and then opens. Git's background-maintenance lock exists for a moment between those two steps, so ENOENT. Every test in `agent/internal/worktree` is exposed; this one just drew it. The fix is to build the template with `gc.auto=0`/`maintenance.auto=false` so there is no transient file to race.

Also filed from local verification: #252, `TestForkedDescendantIsReaped` (`cmd/evener-test-dev-tooling`, byte-identical to main on this branch). It failed on both full-suite runs I made while a background poll loop was spawning short-lived processes, at 60.86s and 60.77s — the deadline, twice — and passed with the root module at 30.35s once that loop stopped. So the trigger is unrelated process churn, not CPU load, which points at descendant enumeration by `ps` snapshot rather than at signalling the process group.

### Local gates, re-run on the merged tree (8346f2012 + 19b416338)

- `make test`: PASS, 7 modules + web, exit 0, root 30.35s.
- `make test-dev-tooling`: 15 suites, all PASS.
- `make lint`: PASS, 7 modules. `make build-all`: green. `gofmt -l`: clean. `git diff --check`: clean.
- `merge-approval-gate-selftest.sh`: 34 checks, 0 failed. `run-module-tests-selftest.sh`: 115 checks, 0 failed.
- `git diff main -- scripts/lib/gate-surface-lib.sh`: comments only — both `GATE_TEST_RUN` and `GATE_FUZZ_TEST_SKIP` byte-identical to main's.
- Deletions vs main unchanged (`gate-capability-preflight.sh`, `test-coverage-floor.sh` + selftest, `cmd/evener-gate-probe/`); `.github` diff still empty; still no product code.

Verdict unchanged: MERGE AFTER FIXES, with the coverage-floor scratch reclaim as the one open item.

Issue #196, opened 2026-08-18
Sidebar disclosure triangle covered by "..." menu on hover — cannot expand/collapse


obra:
# Sidebar disclosure triangle covered by "..." menu on hover

## Problem

The disclosure triangle in the sidebar can no longer be clicked — it is now covered by the "..." (kebab) menu that appears on hover.

## What happens

- Hovering over a project/session row in the sidebar reveals a "..." menu button.
- This menu button overlaps the disclosure triangle (the expand/collapse arrow).
- The disclosure triangle can no longer be clicked to expand/collapse the row.
- This makes the sidebar navigation unusable for expanding/collapsing projects/sessions.

## Expected behavior

- The disclosure triangle must remain clickable at all times.
- The "..." menu should either not overlap the disclosure triangle, or the disclosure triangle should have higher z-index / click priority than the hover menu.

## Steps to reproduce

1. Open the hub web UI.
2. In the sidebar, hover over a project or session row.
3. Observe the "..." menu button appears, covering the disclosure triangle.
4. Attempt to click the disclosure triangle — it is blocked by the "..." menu.

## Severity

High — this breaks basic sidebar navigation functionality.

obra:
## Root cause

`RailRow.tsx`'s trailing disclosure chevron and the row's absolutely-positioned "..." actions overlay compete for the same pixels on the right side of the row, and nothing in the CSS prevents that. When a session/project title is long enough to truncate (ellipsis), the chevron - which sits immediately after the (now-shrunk) label, inside the same flex line - gets pushed rightward into the strip of the row that the hover-revealed kebab overlay covers. The kebab overlay was designed to safely paint over disposable content (the relative timestamp, e.g. "2h"); it was never designed to coexist with an actual *control*, and the chevron only started living in that strip after a later refactor moved it there (see "Introduced by").

Verified with a real, browser-rendered geometry probe (headless Chrome via CDP, the repo's own `layoutguard` harness pattern) against the unmodified `origin/main` CSS/DOM, at the sidebar's actual min (200px) and default (280px) widths, with `:hover` forced on `.railRow` (exactly the real hover trigger) and a realistic 39-character title ("Fix the flaky retry backoff timing test"):

```
280px: chevron=[244.1,248.6]  actions=[207.7,271.0]  -> overlap 4.5px
200px: chevron=[164.1,168.6]  actions=[127.7,191.0]  -> overlap 4.5px
[...]

PR #250, opened 2026-08-19
Sidebar UX: flush layout, footer-only branding, one ☰ toggle, bigger icons, shared row right slot

obra:
## What

Five sidebar fixes, all in the hub frontend:

1. **Flush sidebar** — `AppShell .content` no longer has a 12px gap at any width; the rail's own `border-right` hairline is the only divider, so the sidebar sits flush against the workspace like a modern sidebar. The hidden-state ☰ chip carries its own corner offset (it used to borrow the shell gap).
2. **Branding at the bottom only** — the header's "evener" wordmark is removed; the footer's daemon identity is the rail's only branding. The header row keeps the needs-you Badge, search, and hide buttons.
3. **One icon for the sidebar toggle** — the header's Hide sidebar button now uses the same ☰ glyph as the hidden-state chip (was « to hide, ☰ to show).
4. **Bigger icons** — header/footer icon buttons go from `sm` (28px) to `md` (32px), the IconButton widget's standard tap target (and the old sidebar's documented desktop target); touch devices keep the 44px HIG floor automatically. Row ⋯/+ stay `sm` for density. *(Review follow-up: the glyphs are text, so they sized to the button's font-size, not the box — they now render at 16px, up from 13px. There is no icon library in the frontend; icons are Unicode text glyphs, except the disclosure chevron, which is an inline SVG widget.)*
5. **The hover menu borrows the timestamp's slot** — session/project row actions share ONE grid cell with the right-slot occupant (timestamp / needs-you Badge / "Not started") and swap visibility on hover/focus/open-menu. The ⋯ menu covers the timestamp when revealed and no longer narrows every row's title column with a permanently reserved slot. The disclosure chevron trails the title *left* of the slot, so it stays layout-disjoint from the menu at every width — the issue #196 guarantee is kept structurally, not by stacking. Touch (<900px / no hover) keeps the always-visible side-by-side layout.

**Review follow-ups (second commit):**

- **The ⋯ is right-justified** — the Menu widget's trigger is padded for a standalone button (16px each side), which centered the glyph ~16px in from the slot's right edge. The row now strips the trigger's trailing padding, so the ⋯ hugs the same right edge the timestamp sits at — and the trigger (and with it the shared cell) shrinks from ~47px to ~23px, giving titles another ~24px.
- **Chevron stays clear, provably** — the disclosure chevron and the menu are layout-disjoint in every state: the chevron trails the title *left* of the right slot, and the slot is an in-flow flex item, so flexbox reserves its width and nothing can be laid out (or hit-tested) underneath the menu. layoutguard's `rail-row-chevron-actions-overlap` case proves it geometrically at the sidebar's min (200px) and default (280px) widths, rest and hover, with truncating titles — plus a right-justification assertion (trigger flush to the slot's edge, no wider than the md icon-button box). The second screenshot below shows a hovered branch row: chevron visible at the left, menu flush right.

## How it's pinned

- `RailRow.test.tsx`'s stylesheet-contract describe is rewritten for the shared slot: one grid cell, `visibility` (not just `opacity`) hiding, the three reveal conditions flipping both halves, desktop-hover-only gating, touch side-by-side, and the trigger's glyph-hugging right-justification.
- layoutguard's `rail-row-chevron-actions-overlap` case is rewritten: 18 fixtures (9 configs × rest + forced-hover), asserting disjoint chevron/menu boxes in **both** states, per-state hit-testing (hidden menu eats no clicks at rest; revealed menu is clickable and covers the occupant), slot width == max(occupant, menu), occupant flush right, and the menu right-justified and glyph-sized. **Re-mutation-tested by hand**: opacity-only hiding fails all 9 rest fixtures with the #196 invisible-overlay mechanism; a flex side-by-side slot fails the reserved-space assertion in all 10 occupant fixtures.
- `AppShell.test.tsx`'s full-bleed contract now forbids `gap:` on any `.content` rule.

## Gates

- `make test-web` — PASS (typecheck, unit tests, biome)
- `make test-web-browser` — PASS (layoutguard incl. the rewritten case, overflowguard, spawnguard)
- Live-verified in the real app shell: rail↔workspace gap 0px, ☰ hide button 32×32 with a 16px glyph, wordmark gone, row slot = max(occupant, menu) with the timestamp flush right, actions `visibility: hidden` at rest

## Screenshots

**Full shell — flush sidebar, footer-only branding, one ☰ toggle, md header icons** *(dev server without a backend, so the tree shows its offline error state — the chrome is what's being demonstrated)*

App shell with flush sidebar

**Rows at rest — timestamps flush right, titles reclaimed the menu's old reserved space, header icons at 16px**

Rail rows at rest

**Branch row hovered — the ⋯ (and +) reveal flush right; the › disclosure chevron stays visible and uncovered at the left**

Branch row hovered, chevron clear, menu flush right


obra:
human manual review: +1

PR #253, opened 2026-08-19
feat(hub): render read_file image reads at up to 600px square, no byte-count header

obra:
## What

A `read_file` image read now displays the image itself at up to 600px square (aspect preserved, never cropped or upscaled) instead of a 96px thumbnail, and the `[image: png, N bytes]` header is gone - the picture replaces it.

## How

- New `ToolRendererDescriptor.outputImageSize` (`"large"`); `ToolCallItem` passes it to the generic outputImages `ImageGallery` as a `size` prop.
- `ImageGallery` `size="large"` caps the image at 600px per axis via new `.thumbLarge`/`.thumbImgLarge` classes. Lightbox unchanged.
- PDF reads keep their `[document: ...]` header (no in-transcript preview to replace it); other tools keep 96px thumbnails.

## Tests

- RED-first vitest updates (fsTools / ImageGallery / ToolCallItem).
- New layoutguard case `read-file-image-600px` for the real-browser geometry: 1200x800 -> 600x400, 400x1200 -> 200x600, 100x50 -> natural. Mutation-verified.
- `make test-web` ✅  `make test-web-browser` ✅


obra:
Adversarial review of PR #253 (read-file-image-600, +218/-38, base main). Verdict: **MERGE** — the PR's own code is correct, safe, and well-tested. CI is currently red, but for reasons unrelated to this diff (see below); do not read that as a finding against this PR.

## Scope
Purely frontend (cmd/evener-hub/frontend): ToolCallItem.tsx/.test.tsx, flow/ImageGallery.tsx/.test.tsx + its CSS module, toolRenderers.ts, tools/fsTools.tsx/.test.tsx, and a new layoutguard case. Zero Go files touched — the image-detection/serving pipeline (output_images.go, doc_serve.go, image_serve.go) is untouched.

## XSS / content-sniffing (axis a)
Not touched by this PR, so I audited the untouched baseline it renders through:
- `output_images.go`'s `supportedOutputImageMedia` sniffs bytes via `http.DetectContentType` against an allowlist (`image/png`, `image/jpeg`, `image/gif`, `image/webp`, plus a RIFF/WEBP signature check) — not extension-based, and **SVG is excluded**: Go's `DetectContentType` never returns `image/svg+xml`, and text/xml isn't in the allowlist, so a `.svg` (or anything XML/script-bearing) never resolves as a renderable image.
- Serving routes (`/doc/image`, `/s/.../images/{sha}`) set `Content-Type` from that same sniffed allowlist, never from the request or filename — no content-type confusion into `text/html`.
- The `` in `ImageGallery.tsx` (both before and after this PR) only ever gets a `src` URL to one of those routes, never a `data:`/`srcdoc`/inline-HTML sink.
Conclusion: no new or pre-existing XSS surface from this change; it only changes *how large* an already-safe `` renders.

## 600px cap (axis b)
CSS-only (`.thumbImgLarge { max-width: 600px; max-height: 600px }` in `imagegallery.module.css`), not a server-side resize — same approach the existing 96px thumbnail already used (CSS `object-fit: cover`, not a resized asset). No new memory risk: `output_images.go`'s `outputImageMaxBytes = 8 * 1024 * 1024` already refuses to resolve any file over 8MB as an output image, unchanged by this PR. A 50MB image is rejected upstream before it ever reaches the transcript, same as before.

## Byte-count header removal (axis c)
Display-only. `item.output` on the wire is untouched — `ReadFileOutputBody` (fsTools.tsx) just returns `null` for the image case instead of rendering the header text; the daemon-side format (`agent/execenv/local.go`, `registry.go`'s `ParseImageResult`) is unchanged, so nothing downstream (token accounting, transcript readers) is affected. Grepped for other consumers of that header string in the frontend — none.

## Non-image behavior (axis d)
Pinned: fsTools.test.tsx still asserts plain-text output renders unfolded/unclipped, and PDF (`[document: ...]`) reads keep their header. Mutation-checked (see below) — genuinely red-first.

## Frontend norms
- `600px` is a raw px value, but that matches existing convention for component-specific dimensions in this codebase (e.g. `settings/*.module.css` `max-width: 720px`, `docpane.module.css` `max-height: 80vh`) — not a token-contract violation.
- Layoutguard case `read-file-image-600px` present, and I mutation-verified it myself (not just trusting the PR description): removing `max-height: 600px` → wide/small pass, tall fails with "expected 200x600" (correct reason). Removing `max-width: 600px` → wide fails with "expected 600x400" (correct reason). Both restored cleanly.
- I also mutation-checked the two new vitest assertions this PR relies on: reverting `ImageGallery images={item.outputImages} size={descriptor.outputImageSize}` to drop `size` fails "renders its output image at the large size"; reverting the `if (match[1] === "image") return null` fsTools change fails both "renders nothing" tests. All load-bearing.
- `UserMessageItem.tsx`'s `ImageGallery` call doesn't pass `size`, confirming other call sites keep the 96px default as claimed.

## Local verification
- `make test-web`: PASS (typecheck, vitest, lint)
- `make test-web-browser`: PASS (layoutguard, overflowguard, spawnguard) — `read-file-image-600px` reports exactly the PR's claimed geometry (600x400 / 200x600 / 100x50)

## CI (gh pr checks 253) — both failures are pre-existing/unrelated, not this PR's fault
- **build-and-test** fails on `TestScenarioQuotedLiteralsSitInsideTheCitedLineRange`: `test/scenarios/spawn-empty-prompt-starts-dormant.md` cites stale `RailRow.tsx` line ranges. Confirmed this fails identically on `origin/main` tip alone (no #253 changes at all) — filed as #265, since it currently blocks `build-and-test` for every PR, not just this one. Did not fix it here; unrelated file, out of scope for this PR.
- **web** fails on layoutguard case `rail-row-chevron-actions-overlap` (unrelated to this PR's files). This one is odd: it fails on CI on both the initial run and a rerun, but passes 100% locally against the *exact* merge commit CI tested (`60f0bb3f2`), run twice. That points to a CI-runner-environment difference (headless Chrome version/rendering, akin to the #254 flake family) rather than a real regression — but flagging it since it didn't reproduce locally and isn't literally #254's symptom.

Given "do not merge" is the standing instruction for this review pass regardless of verdict, I'm not merging. Once #265 is resolved and the web-layoutguard flake clears (rerun), this PR should be a clean, no-changes-needed merge.

PR #259, opened 2026-08-19
test(serve): await published model on the G2 premise read (#251)

### LINKED ISSUE #251: CI flake: TestServeModelSwitch_ThreadReadReflectsNewModelWithNoInterveningTurn reads an empty ModelProvider on its premise read
`TestServeModelSwitch_ThreadReadReflectsNewModelWithNoInterveningTurn` failed on CI reading an empty provider where it had just configured one:

```
--- FAIL: TestServeModelSwitch_ThreadReadReflectsNewModelWithNoInterveningTurn (0.09s)
    serve_model_switch_test.go:112: initial thread/read ModelProvider = "", want "gpt-initial"
```

The failure is on the FIRST read (the one establishing the premise), before any model switch happens, so the test is not measuring what it is named for when it fails this way: the thread's model has not been published by the time the read lands.

The test took 0.09s, which suggests the read raced ahead of whatever publishes `ModelProvider` rather than timing out waiting for it. If the read is meant to observe a published state, it needs to await that state rather than assume it is already there.

Seen once on CI for PR #247, whose diff contains no product code (Makefile, scripts, docs, cmd/evener-dev, internal/devtool, audit tests). The same run also hit #249, and main's own CI hit #245 twenty minutes earlier, so the root suite is broadly flaky under CI load right now.

Job: https://github.com/prime-radiant-inc/evener/actions/runs/32282663761/job/96173139994

obra:
## What

`TestServeModelSwitch_ThreadReadReflectsNewModelWithNoInterveningTurn` failed CI reading an empty `ModelProvider` on its *premise* read — before any model switch — so the test wasn't measuring what it's named for when it failed.

## Root cause

The test proceeds as soon as the daemon's rendezvous file appears. But `status.Model` is published only when the bridge drains the session's `EventSessionStart` (`applySessionEventStatus` → `updateSessionInfoLocked`, server/bridge.go:182-193), which is asynchronous with respect to both the rendezvous write and the test's websocket dial. The test took 0.09s on the failing run — consistent with the read racing ahead of the drain, not with a timeout.

## Fix

The premise read now polls `thread/read` until the baseline model is published (25ms interval, bounded by the test's existing 30s context). This is the read that observes asynchronously-published state, so it must await rather than assume.

The post-switch read deliberately remains a **single synchronous read**: immediate visibility after `ThreadModelSet` (via the synchronous `UpdateSessionInfo` in serve.go's `SetModelFunc` closure) is exactly the G2 property this test pins, and polling there would weaken it.

## Verification

- `go test -race -count=5 -run TestServeModelSwitch_ThreadReadReflectsNewModelWithNoInterveningTurn ./cmd/evener` — pass
- `go test -run TestServeModelSwitch_ProviderFailureRestoresCapability ./cmd/evener` — sibling test unaffected
- `go vet ./cmd/evener` — clean

Fixes #251


obra:
## Verdict: REJECT

The mechanics of this fix are sound — `awaitModelProvider` polls only the premise read, the post-switch read stays a single unpolled call, and `thread/read` is confirmed side-effect-free ("answers entirely from memory," server/appwire_runtime.go's `appThreadReadSnapshot` doc), so the G2 "no intervening turn" pin is not weakened. `go vet` is clean, the sibling test is unaffected, and `-race -count=100` passes on this branch.

The problem is the premise itself: **is an empty `ModelProvider` on this read a real product bug?** I tested this directly rather than taking the issue's "test-harness timing" framing on faith, and the answer is yes — it's a genuine daemon-startup race, not an artifact of the test's own dial timing.

### Reproduction against unmodified product code

`thread/read`'s `ModelProvider` comes from `status.Model` (server/appwire_runtime.go:1223), written only by `updateSessionInfoLocked` (server/server.go:457). That's called from exactly two places:
- `applySessionEventStatus` on `EventSessionStart` (server/bridge.go:153-188) — **async**, drained on the bridge's own goroutine once `bridgeSession(sess)` registers it (serve.go:967, per the "registers... then drains on its own goroutine" comment at serve.go:715-720)
- `SetModelFunc`'s closure, synchronously, but only *after* a runtime model switch (serve.go:855)

There is no synchronous seed of `status.Model` at daemon startup. The session's `EventSessionStart` is emitted during `deps.newSession`/`deps.restoreSession` (serve.go:459/475) — hundreds of lines before `bridgeSession` even registers a consumer for it (serve.go:967) — and the rendezvous file that makes the daemon discoverable to real clients isn't written until serve.go:1099, after that registration. Nothing forces the drain to complete before a client can dial and call `thread/read`.

I injected a 200ms delay into `BridgeEvent` (server/bridge.go, product code, not test code) and replaced the test's polling premise read with a single unpolled read — i.e., exactly what a real client's first RPC call after discovering the rendezvous entry would look like. Result: `initial thread/read ModelProvider = "", want "gpt-initial"`, reliably. Without the injected delay, the same single-read version passes. This isn't test-dial-specific; any client racing the drain hits it.

### The team has already treated this exact race as product-facing — for one field, not this one

serve.go:977-983 has a "belt-and-suspenders" fix for `State`, specifically because "the session's SessionStart event may already be sitting in its buffered event channel by the time the bridge goroutine above is scheduled to drain it, so /status could read the server's default state until then." That fix:
- only covers `State`, not `Model`/`Profile`
- only runs `if resuming` (serve.go:977) — a fresh session (exactly this test's and this issue's scenario) gets no such protection at all

`SetModelFunc`'s closure (serve.go:845-857) already demonstrates the correct pattern for `Model`: call `srv.UpdateSessionInfo` synchronously with data the daemon already has locally, instead of depending on the async event bridge. The same call, made once at startup right after `bridgeSession(sess)` (using the already-known `profile.Model()`/`profile.ID()`), would close this race for every client — not just this test.

### Why this matters beyond the test

This is the daemon's own startup path, exercised the same way by any real client (TUI, hub-spawned session under `EVENERHubSpawned`, etc.): dial the rendezvous entry, call `thread/read`. The CI failure this issue reports (`0.09s`, "consistent with the read racing ahead of the drain, not with a timeout") is direct evidence the race already manifests in a real, running daemon under real scheduling pressure — not a hypothetical. Hub-spawned sessions under container CPU limits are, if anything, more exposed to this than a bare-metal CI runner.

### Conclusion

This PR correctly diagnoses the mechanism but fixes it in the wrong place. Polling in the test hides a startup race that a real client can also hit. The fix belongs in `cmd/evener/serve.go`: synchronously seed `status.Model`/`Profile` (and arguably extend the existing `State` belt-and-suspenders to fresh sessions too) right after `bridgeSession(sess)`, the same way `SetModelFunc`'s closure already does post-switch. Once that's in place, the test's premise read can go back to a single, unpolled assertion — which is also a better test, since it would actually catch a regression of the product-level fix instead of tolerating it.

Issue #252, opened 2026-08-19
flake: TestForkedDescendantIsReaped times out waiting for the grandchild's TERM trap under load


obra:
`TestForkedDescendantIsReaped` (cmd/evener-test-dev-tooling/wave_test.go:322) failed once on a heavily loaded local host, then passed in 0.18s on the very next isolated run:

```
--- FAIL: TestForkedDescendantIsReaped (60.86s)
    wave_test.go:322: wave exited (code 143) and nothing wrote to events within 1m0s of its exit
FAIL	primeradiant.com/evener/cmd/evener-test-dev-tooling	79.443s
```

```
$ go test ./cmd/evener-test-dev-tooling -run TestForkedDescendantIsReaped -count=1
--- PASS: TestForkedDescendantIsReaped (0.18s)
```

60.86s versus 0.18s is the whole story: the test is not slow, it hit its deadline waiting for something that never arrived.

The shape: the fixture suite backgrounds a subshell that installs a TERM trap, announces itself on the `ready2` FIFO, and blocks on `read`. The test waits for both readiness handshakes, sends SIGTERM, and expects `grandchild-termed` on the `events` FIFO. On the failing run the wave itself exited 143 promptly and the grandchild's trap output never appeared, so the descendant was either killed outright or its trap never ran — which is precisely the reaping property the test is named for, so a flake here is indistinguishable from the real bug at a glance.

Two candidates worth separating before anything is "fixed":

1. The handshake proves the trap line was *reached*, not that bash had finished installing it and entered the blocking `read`. Under enough load the TERM can land in that window, where bash's default disposition still applies and the trap never runs.
2. Delivery to the backgrounded subshell races the process-group teardown, so it is signalled as a group member before it is a trap-installed one.

Either way the assertion's deadline is not the problem; the handshake's precision is. A second handshake taken from *inside* the trap-armed blocking read (rather than just before it) would make the premise as strong as the assertion.

Seen on a local full-suite run of PR #247. `cmd/evener-test-dev-tooling/` is byte-identical to main on that branch (`git diff main -- cmd/evener-test-dev-tooling/` is empty), so this is not a #247 regression — the branch merely ran the suite under unusual load. No existing issue covers it.

https://claude.ai/code/session_0134LFae5DzpDExkuhJTCNfU

obra:
[...]

PR #271, opened 2026-08-20
Fixes #157: widen the symbol audit to TS/TSX/JS/JSX

### LINKED ISSUE #157: the source-symbol audit checks #symbol anchors for Go only; a .tsx#symbol anchor gets its file checked and its symbol silently ignored
Found by the worktree-kata-test-truth final branch review (2026-08-18). `TestScenarioSourceSymbolsAreDeclared` still matches `scenarioGoCitation`, so after the citation needle widened to every compiled surface (kata `yj52`, closed), a future `.tsx#symbol` anchor resolves its **file** but nothing checks the **symbol** is declared there.

Zero such citations exist today (measured at edc0ed1fa), so this is latent — but the audit's own comments push card authors toward symbol anchors, so the gap will be walked into. The comment at scenariosourcecite_audit_test.go:208 ("already keeps it honest") is scoped to Go by a same-wave edit; this issue is the actual widening: parse .ts/.tsx declarations (ts-morph is not in-tree; a regex-level exported-symbol scan may be the honest floor) or rule the gap accepted with the reasoning here.

**Done when:** a `.tsx#symbol` citation naming an undeclared symbol turns an audit red (prove by mutation), or the gap is ruled accepted and the audit's READ-THIS block names it.

Related: kata `yj52` (closed).

---
_Migrated from kata `26ya` (label `testing`)._

obra:
## Summary

`TestScenarioSourceSymbolsAreDeclared` checked only `.go#symbol` anchors (via `scenarioGoCitation`). A `.tsx#symbol` anchor had its file resolved by `TestScenarioSourceCitationsResolve` (already spanning every compiled extension) but its symbol silently unchecked — the gap issue #157 tracks, per the RCA in the issue comments.

- Widened `scenarioSourceCitation`'s regex with a capture group for `#symbol`, switched `TestScenarioSourceSymbolsAreDeclared` to that needle (was `scenarioGoCitation`), and dispatch declaration lookup by extension: `.go` keeps the existing `go/parser` oracle (`scenarioDeclarationsIn`); `.ts/.tsx/.js/.jsx` get a new line-anchored regex scan (`scenarioTSDeclarationsIn`) recognizing `function`/`const`/`let`/`var`/`class`/`interface`/`type`/`enum` declarations, exported or not — the "honest floor" the RCA specs, since no TS parser is in-tree.
- Known accepted imprecision, matching the RCA: destructured exports (`export const { a, b } = ...`), re-exports (`export { x } from ...`), and `declare`-block members are not matched. Fails in the safe direction — a real symbol the scan misses turns the audit red, not a fake one going unnoticed.
- Removed `scenarioGoFilesByBase`, dead after the switch to `scenarioSourceFilesByBase`.
- Added `TestScenarioTSDeclarationsRecognizeExportForms`, pinning every declaration shape from the RCA plus the two forms that must NOT match.

## Rotten anchor found and fixed

Running the widened audit against the real tree caught a pre-existing rotten anchor: `test/scenarios/spawn-keyboard-contract.md` cited `` `panes/spawn/Spawn.tsx#ModelField` ``, but `ModelField` is declared in `ModelField.tsx` and only imported into `Spawn.tsx`. Repointed to `` `panes/spawn/ModelField.tsx#ModelField` ``. This also seeds the widened needle's first live `.tsx#symbol` match in the corpus, which the RCA calls out as worth doing so the TS branch isn't dead code until someone writes the first citation.

## Mutation proof (TDD, per the RCA's own recipe)

Appended a poison probe to `test/scenarios/_template.md` (in-corpus), then fully reverted:

- Real file, bogus symbol: `` `panes/session/composer/Composer.tsx#thisSymbolDoesNotExistAnywhere` ``
- **Before the fix** (git-stashed the test-file change, kept the probe): `go test -run 'TestScenarioSourceCitationsResolve|TestScenarioSourceSymbolsAreDeclared' -count=1 .` → **PASS**. Confirms the bug.
- **After the fix**: same run → **FAIL**, naming the exact line: `` test/scenarios/_template.md:40: `panes/session/composer/Composer.tsx` declares no thisSymbolDoesNotExistAnywhere ``.
- Probe reverted before commit; `git diff test/scenarios/_template.md` is empty in the final tree.

## Test plan

- [x] `gofmt -l .` clean
- [x] `golangci-lint run ./...` (v2.12.2, matches CI pin) — 0 issues
- [x] `go vet .`
- [x] `go test -count=1 .` (full root suite) — PASS
- [x] Mutation proof above (red before fix, green after, on the same probe)
- [x] Real-corpus rotten anchor found and repaired

### DIFF
--- a/scenariosourcecite_audit_test.go
+++ b/scenariosourcecite_audit_test.go
@@ -34,12 +34,20 @@ var scenarioGoCitation = regexp.MustCompile("`([A-Za-z0-9._/-]+\\.go)(?:#([A-Za-
 // audit read until this one. Kata yj52 found the hole from the other side,
 // asking why a `.tsx` citation survived a green gate; the answer is that no
 // needle in the suite matched a `.tsx` path at all.
+//
+// The `#symbol` alternative carries its own capture group so
+// TestScenarioSourceSymbolsAreDeclared can read the symbol half without a
+// Go-only regex (issue #157); the `:line` alternative does not name anything
+// worth capturing here, since TestScenarioQuotedLiteralsSitInsideTheCitedLineRange
+// reads that half through scenarioSourceLineCitation instead.
 var scenarioSourceCitation = regexp.MustCompile(
-	"`([A-Za-z0-9._/-]+(?:" + scenarioSourceExtensionAlternation() + "))(?:#[A-Za-z0-9_.]+|:[0-9][0-9,-]*)`")
+	"`([A-Za-z0-9._/-]+(?:" + scenarioSourceExtensionAlternation() + "))(?:#([A-Za-z0-9_.]+)|:[0-9][0-9,-]*)`")
 
 // scenarioSourceCitationGroups is the length FindStringSubmatch returns for a
-// scenarioSourceCitation match: the whole span plus its one captured path.
-const scenarioSourceCitationGroups = 2
+// scenarioSourceCitation match: the whole span, its captured path, and its
+// captured symbol (empty when the anchor is a `:line` range rather than a
+// `#symbol`).
+const scenarioSourceCitationGroups = 3
 
 // TestScenarioSourceCitationsResolve keeps a card's pointer into source code
 // attached to a file that still exists. Kata 2mzk's audit deliberately exempts
@@ -98,21 +106,32 @@ func TestScenarioSourceCitationsResolve(t *testing.T) {
 	}
 }
 
-// TestScenarioSourceSymbolsAreDeclared keeps the `#symbol` half of a Go
+// TestScenarioSourceSymbolsAreDeclared keeps the `#symbol` half of a source
 // citation resolvable. `agent/tree_counter.go#defaultMaxConcurrentDelegateTurns`
 // survives every edit to that file — which is the entire reason kata ypwb moved
 // the corpus off `:12` — but only until the symbol is renamed or moved to
 // another file, and nothing else in the suite would notice that.
 //
-// A symbol is anything a card legitimately anchors to: a func or method, a
-// type, a package-level or grouped const or var, a struct field, an interface
-// method. Receiver-, type-, and package-qualified names such as `Type.Method`
-// and `openai.IssuerBaseURL` are preserved; an unqualified alias remains
-// available only when that name is unique in the file, so a collision cannot
-// make a citation appear to resolve by accident.
+// A symbol is anything a card legitimately anchors to. For Go, resolved with
+// go/parser (scenarioDeclarationsIn): a func or method, a type, a
+// package-level or grouped const or var, a struct field, an interface method.
+// Receiver-, type-, and package-qualified names such as `Type.Method` and
+// `openai.IssuerBaseURL` are preserved; an unqualified alias remains available
+// only when that name is unique in the file, so a collision cannot make a
+// citation appear to resolve by accident.
+//
+// For TS/TSX/JS/JSX, resolved with a line-anchored regex scan
+// (scenarioTSDeclarationsIn) rather than a parser — no TS toolchain is
+// in-tree, and a `.tsx#symbol` anchor resolving its file but not its symbol
+// was issue #157. A top-level function, const/let/var, class, interface, type
+// alias, or enum declaration is recognized, exported or not; destructured
+// exports, re-exports, and `declare`-block members are not. That gap fails in
+// the safe direction — a real symbol the scan misses turns the audit red, not
+// a fake one it accepts.
 func TestScenarioSourceSymbolsAreDeclared(t *testing.T) {
-	byBase := scenarioGoFilesByBase(t)
-	declared := map[string]map[string]bool{}
+	byBase := scenarioSourceFilesByBase(t)
+	declaredGo := map[string]map[string]bool{}
+	declaredTS := map[string]map[string]bool{}
 	var findings []string
 	checked := 0
 	for _, path := range scenarioCardFiles(t) {
@@ -121,7 +140,7 @@ func TestScenarioSourceSymbolsAreDeclared(t *testing.T) {
 			t.Fatalf("reading %s: %v", path, err)
 		}
 		for i, line := range strings.Split(string(raw), "\n") {
-			for _, m := range scenarioGoCitation.FindAllStringSubmatch(line, -1) {
+			for _, m := range scenarioSourceCitation.FindAllStringSubmatch(line, -1) {
 				cited, symbol := m[1], m[2]
 				if symbol == "" {
 					continue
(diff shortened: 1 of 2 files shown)


obra:
## Verdict: REJECT

The diagnosis and the bulk of the fix are sound, and I verified the headline claims by execution, not by reading:

- **Poison red-first, confirmed both directions.** Appended `` `panes/session/composer/Composer.tsx#thisSymbolDoesNotExistAnywhere` `` to `test/scenarios/_template.md` (in-corpus). On `main` (b0d491c0d): `go test -run 'TestScenarioSourceCitationsResolve|TestScenarioSourceSymbolsAreDeclared' -count=1 .` → **PASS** (bug reproduced — a real file, bogus symbol, sails through). On this branch (adf1a4ced): same run → **FAIL**, naming `test/scenarios/_template.md:42: `panes/session/composer/Composer.tsx` declares no thisSymbolDoesNotExistAnywhere`. Matches the PR description exactly.
- **TS matcher against real declarations.** `export default function Settings(...)` (`panes/settings/Settings.tsx`) resolves cleanly via a real citation. `TestScenarioTSDeclarationsRecognizeExportForms` correctly pins that destructured exports and re-exports are *not* matched. A class-method anchor (`shell/rail/actions.ts#constructor`) correctly turns the audit red — undocumented as a named limitation, but consistent with the RCA's "unqualified top-level names only" scope and fails in the safe direction, so not a blocker on its own.
- **Full real corpus.** `go test -count=1 .` (all ~915 anchored citations, 418 non-Go) → PASS. The `ModelField.tsx#ModelField` repoint is a legitimate rotten-anchor fix — confirmed `ModelField` is declared in `ModelField.tsx` (`export function ModelField(...)`, line 32) and only imported into `Spawn.tsx`.
- `gofmt -l .` clean, `go vet ./...` clean, `make lint` PASS (7 modules, 21s). CI on the PR is green (`build-and-test`, `web`, both `Build` jobs). Single on-topic commit off `main`. Issues #252/#254/#255 (open flakes) are unrelated surfaces (fork reaping, headless-Chrome DevTools, worktree-fixture git lock) and didn't fire here.

### The blocking problem: `scenarioTSDeclarationsIn` is comment-blind, so a citation can pass without the symbol being real

`scenarioTSDeclarationPatterns`/`scenarioTSDeclarationsIn` (scenariosourcecite_audit_test.go:926-968) is a pure line-anchored regex scan with no lexical model of comments or string/template literals. A `//` line comment is safely excluded because the patterns anchor at `^\s*` and a `//`-prefixed line can never match a bare `function`/`const`/etc. literal at that position. But a **multi-line `/* ... */` block comment** has no such protection: any interior line that isn't itself prefixed with `*` or `//` is scanned exactly like real code.

Reproduced directly against the shipped `scenarioTSDeclarationsIn` (not a reimplementation) with a throwaway test added to this branch and then removed — `go test -run TestReviewerProbeBlockCommentFalsePositive -v .`:

```go
const source = `export function RealSymbol() {}

/*
 * TODO: someday resurrect this:
function BlockCommentedFake() {}
 */

/**
const JsDocFake = 1;
 */
`
```

Result: `names: map[BlockCommentedFake:true JsDocFake:true RealSymbol:true]` — both symbols that exist **only inside a block comment** register as declared. A card citing `` `foo.tsx#BlockCommentedFake` `` would resolve today, green, exactly the failure mode issue #157 exists to close.

This isn't hypothetical noise. The realistic trigger is a symbol getting renamed or its old implementation left commented-out (`/* old impl, remove after M migrates */`) rather than deleted outright — precisely the rename/move scenario kata ypwb built this whole audit family to catch (the PR's own commit message: "only until the symbol is renamed or moved to another file, and nothing else in the suite would notice that"). Here, the "notice" is defeated by a comment instead of a rename, but the outcome — a stale anchor staying green — is the same bug the audit exists to prevent.

It also contradicts a specific, unqualified claim this PR makes twice (commit message and code comment at scenariosourcecite_audit_test.go:930-937): "That gap fails in the safe direction — a real symbol the scan misses turns the audit red, not a fake one it accepts." The documented known-imprecision list (destructured exports, re-exports, `declare`-block members) is indeed all safe-direction. Comment-embedded pseudo-declarations are not on that list, and they fail in the *unsafe* direction — a fake symbol is accepted.

**To flip to accept:** either (a) strip `/* ... */` block comments (and ideally template-literal bodies) from the source before the line scan — a contained, mechanical fix, roughly the size of the existing pattern list, with its own positive/negative test case — or (b) if you'd rather ship the narrower fix now, add this to the documented known-imprecision list alongside destructured exports/re-exports and correct the "fails in the safe direction" claim to say what's actually true (safe for the documented gaps, not comment-embedded text). Given the audit's whole purpose is refusing a fake symbol, I'd fix it rather than document around it, but that's a judgment call for whoever owns this file next.

obra:
Fixed in e0a81d55d. Confirmed your finding by reproducing it: staged a throwaway `reviewerProbeBlockComment.tsx` fixture matching your probe (`RealSymbol` declared for real, `BlockCommentedFake`/`JsDocFake` sitting only inside `/*`/`/**` block comments), cited `reviewerProbeBlockComment.tsx#BlockCommentedFake` from `test/scenarios/_template.md`, and ran the audit against the code from adf1a4ced (before this follow-up): `TestScenarioSourceSymbolsAreDeclared` **PASS** -- reproduces the unsafe-direction gap exactly as you described.

Took your first option: added `scenarioStripTSBlockComments`, a small byte-level state machine that blanks every `/* ... */` span (JSDoc `/** ... */` included) before the line scan, replacing comment bytes with spaces and preserving newlines so line counts and same-line code (`export /* eslint-disable-next-line */ function Foo`) are unaffected. `//` line comments still need no separate handling, as before -- the patterns anchor at `^\s*` and a `//`-prefixed line can never match a bare keyword there.

Re-ran the same poisoned citation against the fixed code: `TestScenarioSourceSymbolsAreDeclared` **FAIL**, naming `test/scenarios/_template.md:41: \`reviewerProbeBlockComment.tsx\` declares no BlockCommentedFake`. Fixture and citation fully reverted before committing (`git diff test/scenarios/_template.md` is empty in the pushed tree).

Added `TestScenarioTSDeclarationsIgnoreBlockComments` as a permanent pin, using your exact reproduction case plus the inline-comment-doesn't-eat-real-code check. Corrected the code comment's "fails in the safe direction" claim to scope it to the documented gaps (destructured exports, re-exports, `declare` members) and call out the block-comment fix and its remaining string/template-literal limitation explicitly.

Full local verification: `gofmt -l .` clean, `golangci-lint run ./...` (v2.12.2) 0 issues, `go vet .` clean, `go test -count=1 .` (full root suite, ~915 anchors) PASS.

Issue #183, opened 2026-08-18
Scratch audits: trap-before-mkdir ordering unenforced; allowlist keyed on basename


obra:
Two audit-strength gaps in the root scratch/delete audits (distinct from #153, which is about unifying the Makefile/scripts predicates):

1. The audits (TestNoScriptFeedsVariableToRecursiveDelete, TestNoScriptCreatesScratchOutsideTMPDIR) enforce "every variable-fed recursive delete must be justified" by count/text only. Mutation test from the PR #133 review: reordering a script to mkdir-then-trap — reintroducing the exact signal-window leak race the trap-before-mkdir convention exists to prevent — leaves the audit green. This applies to every coverage runner too; none have an ordering check.
2. The rm-audit allowlist is keyed on basename, so an unrelated script with a colliding basename (e.g. any other install.sh) inherits an exemption it never earned. Noted during the PR #116 review.

---
Filed from the adversarial review session: https://claude.ai/code/session_0134LFae5DzpDExkuhJTCNfU

PR #276, opened 2026-08-20
Fixes #153: unify Makefile and scripts delete-safety audits on one hardened predicate

### LINKED ISSUE #153: unify the Makefile and scripts delete-safety audits on one hardened predicate
Two implementations of the variable-fed-recursive-delete matcher now coexist:

1. scriptmktemp_audit_test.go's `recursiveDeleteTakingVariable` — the shared architecture, used by the scripts audit. It misses `@rm`/`-rm` make prefixes, `$(RM)`/`${RM}`, and continuation-split deletes.
2. makefile_audit_test.go's standalone detector from the runtime-correctness batch. It catches all three, plus last-segment-on-continued-line refusal — but it is a second copy of a security predicate, which is the drift the shared design exists to prevent.

End state: one predicate carrying the union of both detectors' coverage, both audits pointing at it, with recipe-name attribution kept for Makefile failures. Evidence for each named hole (with run mutations) is in kata `tbqd`'s comments (closed) and the runtime-correctness batch's task-4 history. Both audits' allowlists must survive unchanged, and every mutation from both lineages must still fire.

Related: kata `tbqd` (closed).

---
_Migrated from kata `j58d`._

obra:
## Summary
- `TestNoScriptFeedsVariableToRecursiveDelete` (scripts) and `TestNoMakefileRecipeFeedsVariableToRecursiveDelete` (Makefile) each carried their own copy of the variable-fed-recursive-delete matcher, and had drifted: the scripts copy (a single-pass textual scan for `"rm "`) missed make's `@`/`-`/`+` recipe-line prefixes, `$(RM)`/`${RM}`, and a delete split across a backslash-continued line — all of which the Makefile copy (split into commands, then into whitespace fields) already caught.
- Extracted the shared predicate into `recursivedelete_audit_test.go` (`recursiveDeleteLineText`, `recursiveDeleteCommands`, `namesRM`, `rmArguments`, `recursiveDelete`, `operandComesFromVariable`) and rewired both audits' loops onto it. Each audit keeps its own file-walking, allowlist bookkeeping (`recursiveDeleteAllowedLines`, `makefileVariableFedDeletes` — both unchanged), and Makefile/script line attribution.
- Adopting the Makefile detector's zero-operand refusal for scripts also closes a documented gap for the plain `| xargs rm -rf` shape, and Fields-based word matching closes the documented tab-between-`rm`-and-flags gap; updated `docs/testing.md` to match. `find -exec rm -rf {} +` and `xargs -I{} rm -rf {}` remain invisible, since their operand is a literal `{}` placeholder rather than a variable — documented, not fixed.

## Poison-case evidence
- `recursivedelete_audit_test.go`'s `TestRecursiveDeletePredicateCatchesBothLineages` is a permanent table-driven proof: `@rm`/`-rm`/`+rm` prefixes, `$(RM)`/`${RM}`, and last-command-on-a-continued-line all flag/refuse correctly, alongside the previously-covered trap/backtick/absolute-path spellings and false-positive guards (`confirm`, `docker --rm`).
- Manual mutation evidence (planted, watched fail, reverted — not left in the tree):
  - Makefile: added `@rm -rf "$$POISON_MUTATION_153"` after the `clean` recipe → `TestNoMakefileRecipeFeedsVariableToRecursiveDelete` FAILED on it.
  - Script: added a continuation-split `rm \` / `-rf "$POISON_MUTATION_153_DIR"` to `scripts/lib/gate-surface-lib.sh` → `TestNoScriptFeedsVariableToRecursiveDelete` FAILED on it (a gap the old scripts scanner had no continuation awareness to catch at all).

## Test plan
- [x] `go test . -run TestRecursiveDeletePredicateCatchesBothLineages -v -count=1` — all poison cases pass
- [x] `go test . -run 'TestNoScriptFeedsVariableToRecursiveDelete|TestNoMakefileRecipeFeedsVariableToRecursiveDelete|TestNoScriptCreatesScratchOutsideTMPDIR|TestScratchDirCannotResolveToCWD|TestBuildAllBuildsEveryInstalledBinary'` — pass against real Makefile/scripts content (no regression in allowlist counts)
- [x] `go test . -count=1` (full root package) — pass
- [x] `make lint` — pass (7 modules)
- [x] Manual mutation evidence for both a Makefile poison line and a script poison line (see above), reverted before commit

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -87,10 +87,15 @@ list grows only with an explicitly reviewed reason and shrinks as
 scripts convert or retire.
 
 Two parts of the rule the audit cannot hold, which is why it is written
-here for people. Its scan is textual and per-line, so four equally banned
-spellings are invisible to it: `find -exec rm`, `xargs rm`, a tab between
-`rm` and its flags, and a backslash-newline continuation that puts the
-flags or the path on the next line. And the count pin is narrower than it
+here for people. Its scan is textual, so `find -exec rm -rf {} +` and
+`xargs -I{} rm -rf {}` both read as deletes of the literal `{}`
+placeholder rather than a variable, and slip through — the operand is
+present and visibly not a variable, so neither is a shape the predicate
+can refuse. A recursive delete piped a path with no placeholder at all —
+the ordinary `| xargs rm -rf` shape, or a delete split across a
+backslash-newline continuation — is instead refused outright: the
+predicate cannot tell what it deletes, and treats that as unreadable
+rather than silently passing it. And the count pin is narrower than it
 sounds — it stops a file gaining an *extra* variable-fed delete and it
 fails a listed file whose lines went away, but it cannot see one blessed
 delete being swapped for a different one inside an already-listed file.
--- a/makefile_audit_test.go
+++ b/makefile_audit_test.go
@@ -10,107 +10,11 @@ import (
 	"testing"
 )
 
-// recipeLine trims one physical Makefile line to the shell text it contributes,
-// dropping make's trailing line-continuation backslash.
-//
-// The backslash has to go before anything counts words: left in place it is a
-// bare `\` field, which would read as the path operand of a `rm -rf \` whose
-// real target sits on the next line — exactly the shape the audit must refuse.
-func recipeLine(line string) string {
-	trimmed := strings.TrimSpace(line)
-	return strings.TrimSpace(strings.TrimSuffix(trimmed, `\`))
-}
-
-// recipeCommands splits one Makefile line into the shell commands it holds, so
-// a delete hiding behind `||` or `;` is examined on its own rather than as a
-// tail of whatever ran before it.
-//
-// The two spellings that open a nested command are separators too: a backtick,
-// and the `$$(` that is how a recipe writes shell command substitution (make
-// eats the first dollar, so `$(...)` alone is make's own expansion, never a
-// shell command). Without them a delete wrapped in one reads as an operand of
-// whatever encloses it and is never examined -- `trap` is caught only because
-// it leaves rm as a bare word.
-//
-// Bare braces and parentheses are deliberately NOT separators. `${TMPDIR:-/tmp}`
-// and `$(EVENER_DIST_BIN_DIR)` carry them, and splitting there would tear the
-// variable reference out of the very operand this audit exists to look at.
-func recipeCommands(line string) []string {
-	const sep = "\x00"
-	// `||` is listed before `|`, and `$$(` before either dollar-bearing form, so
-	// the longer operator wins when both match at the same index:
-	// strings.NewReplacer prefers the pattern given first on a tie.
-	return strings.Split(strings.NewReplacer(
-		"&&", sep, "||", sep, "|", sep, ";", sep, "$$(", sep, "`", sep,
-	).Replace(line), sep)
-}
-
-// namesRM reports whether a command word invokes rm.
-//
-// A prefixed or quoted spelling still names the same weapon: make's `@rm` and
-// `-rm`, the `'rm` that opens `trap 'rm -rf …'`, and a fully quoted `"rm"`. The
-// trap spelling matters most — an EXIT trap holding a recursive delete is what
-// deleted a checkout in kata 5hs2 — so it must not read as an ordinary word.
-//
-// Two indirect spellings reach rm without naming it directly. `$(RM)` is
-// defined by GNU make itself (as `rm -f`) and needs no declaration here to
-// work, and any absolute path ends in `/rm`. Both are ordinary things to write,
-// so both have to count.
-func namesRM(field string) bool {
-	word := strings.Trim(field, `@-+'"`)
-	return word == "rm" || word == "$(RM)" || word == "${RM}" || strings.HasSuffix(word, "/rm")
-}
-
-// rmArguments returns the words following an `rm` command word, and reports
-// whether the command invokes rm at all.
-//
-// Finding rm is deliberately separate from deciding the delete is recursive.
-// A recipe may put the command on one physical line and its flags on the next,
-// and a scan that only recognized an already-recursive delete would not see
-// `rm \` as anything at all — so the caller could not refuse it.
-func rmArguments(command string) ([]string, bool) {
-	fields := strings.Fields(command)
-	for i, f := range fields {
-		if namesRM(f) {
-			return fields[i+1:], true
-		}
-	}
-	return nil, false
-}
-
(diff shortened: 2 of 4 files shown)


obra:
## Verdict: REJECT

Adversarial review of the unified predicate found a real coverage regression: a delete shape the **old scripts audit caught** is **silently missed** by the new unified predicate. Verified by execution, not just static reading.

### The regression

`recursiveDeleteTakingVariable` (old scripts detector, `main`) treated `(` as a word-boundary character immediately before `rm `, so a recursive delete wrapped in a bare subshell — no space after the paren — was flagged:

```
(rm -rf "$SOME_VAR")
```

The unified predicate's `rmArguments`/`namesRM` only recognize `rm` as its own `strings.Fields` word, or glued to `@ - + ' " $( `` (the make prefixes, quote chars, and the two dialects' command-substitution openers). A bare `(` glued directly to `rm` is none of those, so the field is `"(rm"`, `namesRM` returns false, and the delete is invisible to both `flagged` and `unreadable`.

**Reproduced end-to-end, not just against the standalone predicate.** I planted this exact shape as a real line in `scripts/lib/gate-surface-lib.sh` on this branch:

```sh
poison_mutation_153b() {
	(rm -rf "$POISON_MUTATION_153B_DIR")
}
```

`go test . -run TestNoScriptFeedsVariableToRecursiveDelete -v` **passes** — the audit this PR ships does not see it.

I then swapped in `main`'s pre-PR `recursiveDeleteTakingVariable` against the identical poisoned file and the same test **fails**, exactly as expected:

```
scriptmktemp_audit_test.go:413: this recursive delete takes a variable a caller can
clobber (kata 5hs2 deleted a home directory this way); ...
      scripts/lib/gate-surface-lib.sh:70: (rm -rf "$POISON_MUTATION_153B_DIR")
--- FAIL: TestNoScriptFeedsVariableToRecursiveDelete (0.02s)
```

So this is not a theoretical corner case: it's a real shell idiom (scoping a delete in a subshell without a trailing space) that the old scripts audit explicitly guarded against (the `(` in its boundary-char set exists precisely for `$(rm ...)`/glued-paren spellings — see its old doc comment), and the unification drops it without comment. Neither `docs/testing.md` nor the PR's `TestRecursiveDeletePredicateCatchesBothLineages` table mentions or tests this shape, so the loss is undisclosed.

The same mechanism (Fields-based matching only recognizes whitespace and the specific separator tokens `&&`, `||`, `|`, `;`, the dialect's command-sub opener, and backtick as boundaries) also misses `rm` glued to a bare `&` (e.g. `sleep 1 &rm -rf $x`), though that spelling is far less realistic as real-world shell style.

This is exactly the "old audit caught it, new one doesn't" failure mode called out as reject-level: a delete-safety net that quietly got weaker for a real, idiomatic spelling.

### What I verified as sound (for the record)

- The PR's own poison table (`TestRecursiveDeletePredicateCatchesBothLineages`) passes, including all three named gaps from #153 (`@`/`-`/`+` prefixes, `$(RM)`/`${RM}`, continuation-split deletes).
- `$(rm ...)` command substitution (as opposed to the bare-paren subshell above) *is* still caught correctly for scripts.
- Two hardening claims are real and reproducible beyond the union: the tab-between-`rm`-and-flags fix (verified red on `main`'s `recursiveDeleteTakingVariable`, green on this branch) and the zero-operand `| xargs rm -rf` refusal for scripts. The tab case, though, has no poison entry in the committed table — worth adding once the paren gap above is fixed.
- `recursiveDeleteAllowedLines` and `makefileVariableFedDeletes` are byte-for-byte unchanged — nothing to justify there.
- Mutation test: gutting `operandComesFromVariable` to always return `false` fails the meta poison test and both real audits (via stale-allowlist detection), confirming the tests aren't vacuous.
- Full package tests (`go test . -count=1`), `make lint`, and `gh pr checks 276` all pass on this branch; no false positives against the real tree.
- Base is `main`, single on-topic commit.

### Ask

Please restore boundary-char coverage for `rm` glued to a bare `(` (and ideally `&`) before merging — e.g. treating those as additional split points in `recursiveDeleteCommands`, or extending `namesRM`'s trim cutset the way it already handles `@-+'"`. Add a poison case for it (and for the tab-glued-flags hardening, which is currently unproven by any committed test) to `TestRecursiveDeletePredicateCatchesBothLineages` so this can't silently regress again.

obra:
Fixed. Mapping finding → fix:

**Finding: bare subshell with no space after the paren (`(rm -rf "$VAR")`) is invisible.** `namesRM`'s Trim-cutset approach (`` @-+'" ``) didn't include `(`, so the Fields-split word `"(rm"` never normalized to `"rm"`. Added `(` to the cutset — same mechanism already used for `@`/`-`/`+`/quotes. This is dialect-agnostic and safe: the cutset only ever trims the specific word already identified as the rm command token, never an arbitrary operand, so it cannot tear a `(` out of a Makefile's own `$(VAR)` expansion (that's a different field entirely).

**Finding: rm glued to a bare backgrounding `&` (`sleep 1 &rm -rf $x`) is invisible, same root cause.** Added `&` to the same cutset.

**Finding: tab-glued-flags hardening had no poison-case test.** Added `rm\t-rf $x` to `TestRecursiveDeletePredicateCatchesBothLineages`. It already passed pre-fix (Fields-based splitting handles any whitespace) — this just closes the "real but unproven" gap you flagged.

Evidence:
- Added all three cases to `TestRecursiveDeletePredicateCatchesBothLineages`. The two paren/`&` cases were confirmed **red** against the pre-fix cutset (`go test . -run TestRecursiveDeletePredicateCatchesBothLineages -v`), then green after adding `(&` to the cutset.
- Reproduced your exact end-to-end repro: planted `(rm -rf "$POISON_MUTATION_153B_DIR")` in `scripts/lib/gate-surface-lib.sh`, ran `TestNoScriptFeedsVariableToRecursiveDelete` — now **fails** on it (previously silent). Reverted before commit.
- Full `go test . -count=1` and `make lint` (7 modules) pass on the branch with the fix in place; both real audits (`TestNoScriptFeedsVariableToRecursiveDelete`, `TestNoMakefileRecipeFeedsVariableToRecursiveDelete`) still validate cleanly against the real Makefile/scripts tree — no new false positives, allowlists unchanged.

Commit: c54f9bab3.

obra:
Closing in favor of #280: GitHub never registered CI checks on this branch (still "no checks reported" after an empty-commit retrigger and a close/reopen). Re-landed the same reviewed content (both the unification and the separator-set fix from this thread) as a fresh branch off current main in #280 — one merge conflict to resolve there since main gained #183's rm-allowlist path-keying fix in the interim; details on what changed beyond the cherry-pick are in that PR's description.

PR #292, opened 2026-08-20
Fix transcript autoscroll: open at bottom, disable when scrolled back

obra:
## Problem

Two linked autoscroll bugs in the transcript pane:

1. **Clicking into a transcript from the sidebar loads mid-transcript** instead of at the bottom.
2. **Autoscroll continues even when the user has scrolled back into history** — the transcript moves underneath the reader. It should be fully disabled when even one line scrolled back.

## Root cause

Both bugs share one root cause in `useTranscriptScroll`: its per-mount refs (`initializedRef`, `wasAtBottomRef`, `baselineItemCountRef`, `firstTurnIdRef`, `resolvedFailedTurnIdsRef`, `errorAnchorIndex`) are set once and never reset. But the Session pane is **not keyed by ref** — `DockHost`'s `PaneHost` renders `<Component params=...>` with no key — so clicking a different session in the sidebar updates `params.ref` on the *same mounted component instance*. The hook's refs all persist across that change.

- **Opens mid-transcript:** `initializedRef` was already `true` from the first session's hydration, so the mount effect's one-time `scrollToIndex(count-1, {align:"end"})` was skipped for every later session opened in that pane. The same-ref remount path (model briefly `undefined` → re-hydrated, e.g. a store resync) had the same flaw.
- **Autoscroll while scrolled back:** `wasAtBottomRef` etc. were stale from the *previous* session, so stick-to-bottom and pill counts were computed against the wrong scroll state. Independently, the 50px at-bottom threshold let autoscroll engage while comfortably scrolled back into history (one line of text is ~20px).

## Fix

**`useTranscriptScroll.ts`** — reset all per-ref state when `ref` changes (`refForInitRef`) or when `hasContent` flips false→true on the same ref (`prevHasContentRef`). The reset runs before the scroll-element null check so it still fires on renders where VirtualList hasn't mounted yet.

**`scrollMetrics.ts` + `widgets/virtuallist/index.tsx`** — tightened `AT_BOTTOM_THRESHOLD_PX` and `END_ANCHOR_THRESHOLD_PX` from 50 to 4: rounding-safe but far smaller than one line of text, so scrolling back even one line fully disables autoscroll (both the hook's stick-to-bottom and the virtualizer's `followOnAppend` / `wasAtEnd` resize-adjustment).

## Verification

Built a temporary diagnostic harness using the real `VirtualList` + `useTranscriptScroll` and exercised it in a real browser (Chrome), reproducing both bugs against the pre-fix code and confirming the fix:
- ref-change remount: now lands at bottom (was `scrollTop=0`, gap=3222)
- same-ref remount (model undefined→defined): now lands at bottom (was top)
- scrolled back 30px (one line, within old 50px threshold) + append: no yank, pill counts (was yank to bottom)
- at true bottom + append: sticks

## Tests

- 4 new tests in `useTranscriptScroll.test.ts`: ref-change re-runs scroll-to-bottom; ref-change resets pill/baseline; ref-change clears error anchor; same-ref remount re-runs scroll-to-bottom.
- Updated threshold assertions in `scrollMetrics.test.ts` and `virtuallist.test.tsx` (50 → 4).

## Gates

- `make test-web` — PASS (typecheck + test + lint)
- `make test-web-browser` — PASS (layoutguard + overflowguard + spawnguard)

### DIFF
--- a/cmd/evener-hub/frontend/src/panes/session/transcript/flow/scrollMetrics.test.ts
+++ b/cmd/evener-hub/frontend/src/panes/session/transcript/flow/scrollMetrics.test.ts
@@ -7,18 +7,21 @@ import {
   readScrollMetrics,
 } from "./scrollMetrics";
 
-// isAtBottom: legacy parity threshold (docs/web-ui/parity/parity-m4-transcript.md
-// §15) - "within 50px of true bottom counts as at bottom".
+// isAtBottom: "at the bottom" means the reader is at the TRUE end of the
+// scrollable content, not merely near it. The threshold is a few pixels
+// (rounding-safe, far smaller than one line of text ~20px) so a reader who
+// has scrolled back even one line is NOT treated as at the bottom - the
+// legacy 50px value let autoscroll engage while scrolled comfortably back.
 test("isAtBottom: true when scrollTop+clientHeight exactly equals scrollHeight (pixel-perfect bottom)", () => {
   expect(isAtBottom({ scrollTop: 950, scrollHeight: 1000, clientHeight: 50 })).toBe(true);
 });
 
-test("isAtBottom: true within the 50px threshold", () => {
-  expect(isAtBottom({ scrollTop: 910, scrollHeight: 1000, clientHeight: 40 })).toBe(true); // gap = 50
+test("isAtBottom: true within the 4px threshold (rounding tolerance)", () => {
+  expect(isAtBottom({ scrollTop: 946, scrollHeight: 1000, clientHeight: 50 })).toBe(true); // gap = 4
 });
 
-test("isAtBottom: false just past the 50px threshold", () => {
-  expect(isAtBottom({ scrollTop: 909, scrollHeight: 1000, clientHeight: 40 })).toBe(false); // gap = 51
+test("isAtBottom: false just past the 4px threshold (one line of text is ~20px, far larger)", () => {
+  expect(isAtBottom({ scrollTop: 945, scrollHeight: 1000, clientHeight: 50 })).toBe(false); // gap = 5
 });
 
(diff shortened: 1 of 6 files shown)


obra:
Decision: **rework before merge**.

The ref/remount reset clears most per-session state, but it leaves `pendingViewAnchorRef` intact. A view-anchor captured in session A can therefore run after switching to session B and override B's fresh scroll-to-bottom. A focused supplemental test reproduces it: the new session is placed at `scrollTop = 900`, then the stale anchor restore moves it to `918`.

Required rework:
1. Clear/cancel any pending view-anchor restoration when `ref` changes or `hasContent` remounts.
2. Add a regression test that captures an old-session anchor, switches refs, and asserts the new session remains at the fresh bottom position.
3. Update from current main and rerun the frontend/unit/browser gates.

Existing evidence is otherwise green: the branch's required GitHub checks pass; locally, 85 focused tests and `web-layoutguard`, `web-overflowguard`, and `web-spawnguard` pass. The race above is the blocking gap.

obra:
Rework pushed in `3efe2e57fe4d456ca734fd03cea720d955bfa22d`.

- Added the ref-switch race regression and watched it fail `918` vs `900` before the production edit.
- Cleared `pendingViewAnchorRef` in the existing ref/remount reset block.
- Focused transcript/virtual-list set: **86/86 pass**.
- Parent-side gates: **typecheck, frontend tests, lint, layoutguard, overflowguard, and spawnguard all pass**.
- Fresh independent review: spec compliance pass; code/test quality pass; no findings.
- Branch includes current `main` (`0579607e82bd…`).

Leaving the PR draft until the new GitHub checks finish; if they remain green, it is ready to merge.

obra:
Updated-head CI passed (`web`, `build-and-test`, and snapshot artifacts). Merged as `fefaca31b29b0b8457580486b3e9695bcdfcaae7`.

PR #295, opened 2026-08-20
feat(web): redesign the inline subagent widget as the Rail × Quote card

obra:
## What

Redesigns inline delegate rendering as one square, headless Rail × Quote card per delegate tool row.

- Each delegate row owns exactly one card and its own `open ⤢` action. There is no aggregate module, leader election, cross-row sorting, or done-card folding.
- A 2px rail plus a shape-distinct glyph carries state. Visually hidden identity/status text and complete disclosure ARIA preserve accessibility without visible status labels.
- The folded hero quote is intentionally italic. Expanded recent activity preserves source-specific styling, true ordinals, per-item timing, and local timestamps.
- Stats show child turns/calls and projected token usage when available; clocks use one pane-level tick rather than per-card timers.
- Cards read stable delegate data directly from the owning canonical `ThreadModel`; child watches provide transcript content and live child status.
- Session-scoped row state is released with the parent pane. Long paths/URLs and stats wrap safely in narrow cards.

## Deliberate follow-up

#307 tracks moving attention into the revisioned stable delegate projection so card lifecycle/attention no longer combine stable delegate state with child-thread cadence. This PR intentionally does not add a temporary reconciliation layer.

## Review-driven simplification

The rework removed aggregate rendering/state, duplicate projection caches, dead watched-child Cadence/frame-time state, per-card timers, duplicate row publication, stale cache tests, and redundant helper/structure tests. The final branch is net-negative versus `main` and adds no test framework, shell script, meta-test, backend protocol change, or generalized cache machinery.

## Verification

Parent-side checks pass:

- `make test-web`
- `make test-web-browser`
- typecheck, Vitest, and Biome
- layoutguard, overflowguard, and spawnguard

The pushed head still requires its fresh GitHub CI run before merge.


obra:
Decision: **rework before merge**. The branch is coherent and its current-head CI is green, but review found four blocking correctness/accessibility gaps:

1. **Terminal state can be repainted by a stale child watch.** `displayKind` uses the revision-fenced projection, but `attention` and `childRunning` read the possibly stale watched model directly. A stable failed row plus stale `awaiting` becomes `data-kind="failed" data-attention="true"`, so amber wins over danger; a done row plus stale `active` can still mark its last quote live. Gate these overlays on authoritative `displayKind === "running"` and add failed+awaiting / done+active races.
2. **Aggregated cards can be attributed to the wrong delegate.** The elected first delegate renders all independently sorted cards, while cards contain no task/agent identity or textual/ARIA status and the matching `open` action remains on separate tool rows. Add a stable visible/programmatic card identity and matching action (or render each card under its own row), tested through two real `ToolCallItem`s after worst-first reordering.
3. **State/disclosure accessibility regressed.** State is color-only; stopped/unknown collapse to done-neutral; disclosures lack `aria-expanded`, `aria-controls`, and an identified region. Expose identity/effective status and wire the disclosure relationship.
4. **Folded tool purposes are incorrectly italicized.** The fold drops the quote's `msg` discriminator and `.quote` always applies italics, despite the PR contract that tool purposes are plain and agent messages italic. Preserve the latest quote object and style by `msg`; add the assertion.

Also add narrow-card long-token wrapping/browser coverage. Then update from current `main` (the branch is now 5 commits behind after #294), rerun required checks, and return for review.

Independent local evidence: `make test-web` and `make test-web-browser` pass on the current head. The blockers are contracts those suites do not currently exercise.

obra:
Review adjudication update: lifecycle/child-watch state fusion is accepted as temporary presentation debt and is no longer a blocker for this PR. The long-term producer-side fix is tracked in #307. Do not add a temporary reconciliation layer here; #295 still needs the remaining identity, accessibility, and folded-purpose findings resolved before merge.

obra:
Review adjudication update for card identity: each delegate tool row should own and render exactly one subagent card. Remove the elected leader/aggregate module, cross-row worst-first reordering, and aggregate done-card folding. Keep the matching open action with that card. Add a real two-ToolCallItem regression proving each row retains its own card and action.

obra:
Review adjudication update for accessibility: keep cards textually headless. Add visually hidden delegate identity and effective status; add a small shape-distinct status glyph so sighted state is not color-only; keep the glyph aria-hidden. Wire disclosure with aria-expanded, aria-controls, and a stable ID on the controlled region.

obra:
Review adjudication update for folded quotes: unconditional italics in the folded card are intentional, so this is not a defect and no styling change is required. Please align the PR description/code comments with the intended contract: the folded hero quote is always italic as a quotation treatment; source-specific plain-vs-italic styling applies only in the expanded activity list. Add or retain a test that pins the intentional folded treatment.

obra:
Review adjudication update for narrow-card overflow: fix this in #295. Add min-width: 0 and overflow-wrap: anywhere to quote content, make the stats row shrink/wrap safely, and add a narrow-width browser-guard fixture with a long unbroken path or URL.

obra:
Review adjudication update for projection retention: fix this in #295. During the one-row/one-card simplification, remove the page-lifetime session projection stash if it is no longer needed. If virtualization still requires it, add explicit per-session eviction when the parent thread’s final consumer releases or the session is deleted, with a two-session isolation test.

obra:
Review adjudication update for clocks: per-card intervals are not acceptable. Delegate state/content already arrive reactively from stores; elapsed time should reuse the Session pane’s existing NOW_TICK_MS clock. Pass that `now` value through a narrow prop/context to cards (standalone fixtures may inject a fixed value), and remove `useNowTick` from each card. Terminal cards render fixed durations and never own a timer. Add a test proving multiple cards create no additional intervals.

obra:
Constrained rework pushed at `976d1e75697bf286f5ffb077589685323815751b`. Full parent frontend/browser gates pass. Independent correctness review and a separate brutal deletion-focused full-branch review are clean. The branch remains draft until fresh GitHub CI completes.

obra:
Fresh updated-head CI passed (`web`, `build-and-test`, and snapshot artifacts). Merged as `1e7823b75967488a27c94e663e1a9dd767d7791e`. Architectural follow-up remains tracked in #307.

Issue #310, opened 2026-08-21
A run cannot finish with an undisposed background process (design for #297)


obra:
Design spec for #297. Supersedes the earlier drain-guidance spec and the timeout-tuned draft that preceded it; both are described in the body.

# A Run Cannot Finish With an Undisposed Background Process

Fixes #297. Supersedes `2026-08-20-end-turn-background-job-guidance-design.md`,
which remains the record of two failed implementation attempts, and replaces an
earlier draft of this file that tuned timeout constants to one benchmark's
handoff distribution. Two adversarial reviews found that draft's numbers were
fit to the wrong run and its central mechanism inverted; the timing layers are
gone.

## The principle

**A coding agent that leaves an orphaned process running has not finished its
job.** That is true interactively, in CI, in an eval harness, and in a delegate.
It is a statement about task completion, not about runtime policy, which is why
it needs no timeout to express.

The earlier draft treated the drain's *wait* as the thing to bound, so every
version of it needed a number — and no number generalizes. An interactive user
wants control back in seconds; a CI job doing a cold `cargo build` needs ten
minutes and must not be interrupted; an eval harness has a fixed per-task
budget. Any constant is wrong for two of those three.

The wait is not the problem. The **undisposed process** is, and the agent is the
only party who can dispose of it. So we tell the agent.

## The failure

Terminal-Bench trial `hf-model-inference__X64mGVa` (run
`tb21-luna-max-89-9ef823040-20260820T185134Z`, binary `9ef823040`) asked for a
Flask API served on `0.0.0.0:5000` in the background. The agent did that
correctly in 149 seconds. The process then sat idle **751 seconds** until the
harness killed it. The server died with the session; the trial scored 0.0, with
1 of 4 verifier tests passing.

Two mechanisms are at fault: the drain waits for every session-owned shell job
and a service never finishes; and the `end_turn` warning promised the job would
"report separately on completion", which under `evener run` is false. The second
is fixed (`cc27d687e`).

## What the evidence supports

**The launch decision is what determines the outcome.** In the Aug 11 run
(`tb21-luna-max-89-8e4fd1181-20260811T153812Z`, binary `8e4fd1181`) the same
task, `hf-model-inference__t7vKYLs`, **scored 1.0** — that agent launched the
server with `mode:"detached"`. Same task, opposite outcome, decided by the mode
argument. No timeout value changes the Aug 20 result: the server dies at exit
whatever the deadline.

[...]

PR #311, opened 2026-08-21
A run cannot finish with an undisposed background process

### LINKED ISSUE #297: Background jobs deadlock the one-shot exit path until the context is cancelled
A long-running background shell started with `exec_command mode:"background"` holds the one-shot exit path open until the caller's context is cancelled. The agent finishes its turn, produces a correct result, and then the process hangs until something external kills it.

## Evidence

Terminal-Bench 2.1 trial `hf-model-inference__X64mGVa` (run `tb21-luna-max-89-9ef823040-20260820T185134Z`). The task asks for a Flask API served on `0.0.0.0:5000`, run in the background.

| | |
|---|---|
| agent start | 20:35:18 |
| `communicate{end_turn:true}` | 20:37:47 (149 s in) |
| session meta `work_millis` | 149004 |
| process alive, no further output | 20:37:47 → 20:50:18 (751 s idle) |
| harness SIGINT | 20:50:18 |

All 26 trajectory steps land in the first 149 s. stderr ends `[tool] communicate: done`, then 12.5 minutes of silence, then `evener: context canceled`.

The agent solved the task. It scored 1 of 4 verifier tests instead of 4 of 4, because the Flask server was killed along with the session.

## Mechanism

`cmd/evener/run.go:311-332` drains the job tree before `Close()`:

```go
result, err := runProcessInput(sess, ctx, prompt)
if err == nil {
    if drained, derr := runDrainJobTree(sess, ctx); derr != nil { ... }
}
sess.Close()
```

`isOwnedDrainJob` (`agent/session_jobtree_drain.go:36-44`) counts any session-owned `jobstore.JobShell`, and its comment explains why it cannot discriminate:

> The durable record does not reliably preserve a shell's original execution mode, so ownership and job type are the complete drain contract.

`DrainJobTree`'s own doc states the consequence (`:323-325`):

> blocks until the caller's context is cancelled. Individual managed-job turns carry their own round/time caps, so a well-formed tree always quiesces on its own. Every owned managed job holds the drain open.

A service that is *supposed* to keep running is not a well-formed tree and never quiesces. `drainStallTimeout` cannot fire because a live process resets the stall clock each pass. The background-job kill path (`agent/jobs.go:668-753`, `stopped_by_parent` + SIGTERM/SIGKILL) runs *after* the drain, so it is unreachable.

## Scope

Scanning all 89 trials in that run, **at least 3 of the 9 `AgentTimeoutError` results are this deadlock rather than the agent running out of work**, identified by `work_millis` far below budget plus the `[tool] communicate: done` → `context canceled` signature:

```
hf-model-inference   work= 165s / 900s budget   reward 0.0
mailman              work= 745s /1800s budget   reward 1.0
train-fasttext       work=2599s /3600s budget   reward 0.0
```

`mailman` passed only because its deliverable was persistent config rather than a session-owned process.

## The warning actively reassures

`agent/session_tools_communicate.go:109-129` emits:

> ending turn while 1 job(s) are still running: job_…. The call still succeeds; each job remains notification-armed and will report separately on completion.

In a one-shot run there is no "separately" — the turn is the process. Asked afterwards, the agent said:

> I expected `communicate(end_turn=true)` to end my active agent turn promptly while the background job continued running and remained notification-armed. I did not expect the Flask process itself to exit immediately at that point.

## Suggested fixes

Either would resolve it; the first is the smaller change.

1. Exempt background-mode shells from the one-shot drain, so `Close()`'s existing kill path handles them.
2. Make `end_turn` with a live long-running owned job a hard error in one-shot mode, naming `mode:"detached"` (`agent/session_tools_shell.go:159` → `runDetachedShell`) which already bypasses the job manager and would have survived.

At minimum the warning text should stop promising a later notification that one-shot mode cannot deliver.

obra:
Fixes #297. Implements the design in #310.

## The failure

A one-shot `evener run` whose model left a background service running held the drain open until something external killed the process. In the motivating Terminal-Bench trial the agent served a Flask API correctly in 149 seconds, then the process sat idle **751 seconds** until the harness killed it; the server died with the session and the trial scored 0.

## The principle

**A coding agent that leaves an orphaned process running has not finished its job.** That is a claim about task completion, not runtime policy — which is why this PR contains **no timeout constants**. Two prior attempts (both reverted, recorded in `docs/superpowers/specs/2026-08-20-end-turn-background-job-guidance-design.md`) failed by bounding the *wait*; no constant generalizes across an interactive user, a CI cold build, and an eval budget. The wait was never the problem. The undisposed process is, and the agent is the only party who can dispose of it.

## What lands, commit by commit

**Independent fixes** (real bugs on main regardless of the feature):
- `Record the promotion when a foreground shell outlives its wait` — a promoted job's live record said `Background:false` while its tool result said `RunningInBackground:true`. The record lied about the default shell mode.
- `Re-check the wake edge before the stall watchdog gives up` — the stall give-up returned on a torn verdict; a delegate packet arming attention mid-scan could be SIGKILLed by `Close()`. Same PRI-2441 B1 protocol the quiescence return already follows.
- `Stop teaching the model the launch choice that orphans its server` — the docs recommended `mode="background"` for a dev server (the exact #297 instruction, contradicting the system prompt), the prompt claimed "waiting costs nothing", and nothing model-facing said a detached process discards output and sends no completion notification. Evidence for this lever: the **same benchmark task passed when the model chose `detached` and failed when it chose `background`**.
- Lint: carve out Google's camelCase wire-format tags from the fuzz-tagged tagliatelle gate (latent on main; a fresh worktree path surfaced it).

**Plumbing:**
- `TurnEndsProcess` on `SessionConfig`, snapshot-carried so children inherit it; restore REPLACES the persisted value (a session resumed by `evener run` dies with its turn; resumed by the daemon it does not); the stable-delegate frozen-descriptor path re-takes it from the live parent — the leak the earlier attempt missed, now an extracted, tested merge function.
- The honest `end_turn` warning: one-shot sessions are told a still-running job is killed at exit rather than promised a report that never comes.

**Design change:** `Let every session watch its own jobs` — `job_watch` was stripped from leaf delegates alongside `delegate`, leaving sessions that can start jobs unable to wait on them. Watching is not delegation, and every cross-session source already authorizes itself (`parent` needs `watch_parent=true`; a concrete job id must be owned by the watcher). The strip was defense in depth, not the enforcement point.

**The feature:** when undisposed background shells become the drain's sole remaining reason to wait, the drain tells the model, with three remedies:
- **`job_stop`** — scratch work
- **stop, then relaunch `mode="detached"`** — must outlive the run (withheld under a sandbox via the new `DetachSupportReporter`, which `DetachCommand` itself consults so capability and refusal cannot drift)
- **`job_watch`** — terminates on its own and the answer needs its result; the drain then waits exactly as it always has

Escalation is paced by the model's own turns: announce → announce once more naming the consequence → stop waiting so `Close()`'s kill path runs. Each announcement IS a turn; the count advances only when a turn completes with the set still undisposed. The ordinary handoff — a job finishing exactly as the drain starts — is never announced: the first announcement requires the condition to survive one park, so the completion wake wins the race.

Housekeeping cannot corrupt the run: announcement replies never become the run's printed answer (root or delegate), an announcement-turn error warns and advances instead of failing the run, and a killed job is reported with command text, runtime, and quiet time.

## Testing

TDD throughout — every behavior watched failing first, including the headline e2e reproducing the #297 hang through the real `run()` (60s tripwire, now completes in 0.7s). All delivery tests drive the real `ProcessInputKind`; substituting the turn runner is how both prior attempts shipped inert. Nine deliberate mutations, all killed — including `EntryNotification → EntrySteeringCarrier` (attempt 2's exact defect, caught in 0.06s) and one initial survivor (dropping the elsewhere check) that exposed a coverage gap and got its own test.

Gates: `make lint` PASS (8 modules), `make vet` clean, `make test` PASS all 8 modules.

## Not in this PR

The spec's acceptance rerun of `hf-model-inference` on a patched binary, and the sandbox story for tasks that need a surviving process (nothing can outlive a sandboxed run; the message says so and tells the model to say so).


obra:
## Live acceptance + regression results

20 Terminal-Bench 2.1 trials against a binary built from this branch (`c2a00e1f4`), same recipe as the Aug 20 baseline run (`9ef823040`, 68/89). Two waves: the failure set (4 tasks × 2) and the regression cohort — every Aug 20 trial that solved with >30s background handoffs, i.e. exactly the population the announcement now interrupts.

| Task | Aug 20 | Now (×2) | Announced | Escalated past step 1 | Idle tail |
|---|---|---|---|---|---|
| hf-model-inference (#297) | 0.0, **751s hang** | **1.0, 1.0** | no — chose `detached` at launch, both attempts | — | 1s |
| mailman | 1.0, **~1,100s dead** | **1.0, 1.0** | #2: yes, on the **promoted** job | no | 1s |
| train-fasttext | 0.0, **767s hang** | 0.0, 0.0 (task difficulty, all runs) | yes, watch armed | no | **0s** |
| compile-compcert | 1.0 | **1.0, 1.0** | yes ×2, watch armed | no | 1s |
| caffe-cifar-10 | 1.0 | **1.0, 1.0** | yes, watch armed | no | 1s |
| fix-ocaml-gc | 1.0 | **1.0, 1.0** | yes ×2, watch armed | no | 0s |
| mcmc-sampling-stan | 1.0 | **1.0, 1.0** | yes ×2, watch armed | no | 1s |
| rstan-to-pystan | 1.0 | **1.0, 1.0** | yes ×2, watch armed | no | 0s |
| overfull-hbox | 1.0 | 0.0, 1.0 (attempt-1 miss: verifier disagreed with synonym choices; no drain involvement) | no | — | 0s |
| pypi-server | 1.0 | **1.0, 1.0** | no — `detached` at launch | — | 0s |

**Mechanism verdict across all 20 trials:** zero hangs (max idle tail 1s, vs 751–1,100s on the baseline), zero kills, zero final notices — **no model ever needed the second warning**. The announcement fired in 13 trials; in 11 the model answered with the prescribed `job_watch(..., progress_interval_ms=120000)` and the drain waited the work out.

The compile-compcert transcript shows the design's teaching effect: announced once, the model armed the watch, and then **proactively armed watches at launch for its five subsequent long background jobs** — no further announcements needed for the rest of the session.

`mailman` #2 is the direct proof of the promotion fix: the announcement fired on a foreground-promoted job that was invisible to the guard before `rec.Background` was stamped at promotion.

Net score on the ten tasks vs the Aug 20 baseline: **8/10 → 9/10** (hf-model-inference gained; nothing lost).

Artifacts: `tb21-disposal-rerun-c2a00e1f4-20260821T150409Z` and `tb21-disposal-regression-c2a00e1f4-20260821T151454Z` under `harbor-runner/runs/` (local) and on magic-kingdom.

https://claude.ai/code/session_01R2gZgMAJ34xsyABHwUDZpz

PR #333, opened 2026-08-22
docs: hub-first getting started

obra:
## Why

New users previously met Evener through component-level docs. The hub
(`evener-hub`) is the interactive surface we want to hand users by default,
so the getting-started path now starts there.

## What changed

**Commit 1 — hub-first install and getting-started refresh**
- `README.md` restructured so the quick start leads with `evener-hub` and
  the web UI; component sections demoted to reference reading.
- New `docs/getting-started.md`: the canonical walk from install to a first
  session (install → start hub → auth URL → credentials → first session →
  everyday workflow → upgrade), linking out instead of duplicating.
- `cmd/evener-hub/README.md` revised as the production-style operator
  runbook.

**Commit 2 — follow-ups from review**
- The hub web UI is framed as Evener's **default interactive surface**, not
  the only one — `evener-tui` and the non-interactive CLI are named as the
  other ways in.
- **"Migrating from Serf" section removed** from user-facing docs.
  `install.sh` still detects `~/.serf`/`~/.evener` at install time; its
  notice now suggests `evener-migrate --dry-run` instead of the deleted
  README section.
- Hub operator runbook **moved to `docs/evener-hub.md`** (sibling of
  `evener-hub-remote-operations.md`); a stub `cmd/evener-hub/README.md`
  points to it, keeping historical spec/plan links resolving.
- Review-driven fixes: TUI command is `/credentials` (not `:credentials`),
  `--auth-token`/`--state-dir` added to the TUI flag table, lowercase
  "the hub" throughout, sentence-case headings, linkified launch-config
  design spec.

## Verification

- All relative links and anchors across the four docs resolve (checked by
  script; historical `docs/superpowers/` specs/plans intentionally untouched).
- Zero remaining Serf references in user-facing docs.
- `go test -run TestInstallScript .` passes (pins the install.sh notice).
- Each chunk reviewed by two subagents (factual against code ground truth +
  editorial); all findings addressed.

Docs-only change except the two-line `install.sh` message text and its
test's comment.


obra (commented):
## Review result: changes needed — not merged

I reviewed head `c71ed661372a28e6c590d75a5d6ad4e201f0d218` against the implementation, ran targeted checks, and used independent factual, editorial, and mechanical passes. CI is green, but the documentation still has material accuracy and security problems.

### Blocking findings

1. **`README.md:145-150` documents a launch configuration that cannot work.** The example places `OPENAI_API_KEY` under `[env]`, but every launch layer drops credential-like keys (`cmd/evener-hub/internal/launchconfig/merge.go:11-16,274-282`), and `merge_test.go:421` explicitly classifies `OPENAI_API_KEY` as blocked. Remove the secret from this example and direct users to `/credentials`, `credentials.toml`, OAuth, or the hub process environment.

2. **`docs/evener-hub.md:218-241` can expose the hub bearer token in world-readable logs.** The hub prints the complete auth URL to stderr (`cmd/evener-hub/main.go:471-476`), while the recipe uses plain `mkdir -p` and `tee`; under a normal `umask 022`, that produces a 0755 directory and 0644 log. Create/chmod the directory to 0700 and the log to 0600, and state that supervisor logs contain a bearer credential.

3. **`README.md:160` and `docs/evener-hub.md:238-241` overstate `deploy-hub.sh` safety.** The script builds first, then `kickstart -k` replaces the old process, then checks the replacement (`scripts/ops/deploy-hub.sh:149-189`). It does not keep the old hub running until the new hub is healthy. Describe that sequence accurately.

### Other material accuracy issues

- **System install path mismatch:** `docs/evener-hub.md:54-58` offers `/usr/local`, but later commands at lines 218-220 and 253-256 hardcode `$HOME/.local/bin`. Use PATH-resolved commands or show both layouts.
- **Configurable paths ignored:** the authenticated smoke check at `docs/evener-hub.md:271-278` hardcodes the default token path even though `hub_state_root` is configurable. The credentials path at lines 176-196 and `docs/getting-started.md:78-79` also moves beside `EVENER_PROVIDERS_CONFIG` when that override is set.
- **Token invariant overstated:** `docs/evener-hub.md:18-20` says a loaded token is 256-bit and mode 0600, but existing non-empty files are accepted without format or permission checks (`auth_token.go:83-88`). Only newly generated tokens get those guarantees.
- **Unsafe pruning advice:** `docs/evener-hub.md:300-301` says `run_dir` can be pruned after failed probes. A failed probe with a live PID is intentionally retained (`hubcore/roster.go:238-245,289-302`); deleting it can hide a live daemon and discard logs.
- **Loopback claim too broad:** `README.md:154` says daemons are loopback-only. Hub-spawned daemons are forced to loopback, but direct `evener serve --addr` can override the loopback default.
- **Upgrade claim too broad:** `README.md:48-52` and `docs/getting-started.md:120-129` imply all five installed commands are upgraded, but `internal/selfupdate/update.go:20` omits `evener-migrate`.
- **Credential requirement too broad:** `docs/getting-started.md:76` says every setup needs a provider credential; auth-none providers such as local Ollama do not.
- **TUI readiness detail is wrong:** `README.md:182` says auto-start waits for `/api/health`; the normal readiness loop initializes authenticated AppWire over `/rpc`. The health compatibility check only applies with an explicit state directory.
- **Sidebar wording is absolute:** `README.md:119` and `docs/getting-started.md:96-98` say Live contains every running session, but archived live sessions appear under Archived instead.

### Quality/test follow-ups

- `install_test.go:620-623` only checks for `evener-migrate`; it passes for both the old stale README pointer and the new `--dry-run` text. Assert the new guidance and reject the removed pointer.
- `docs/getting-started.md:28-35` says “Two environment variables” but names four. Full pipeline examples would also prevent users from attaching variables to `curl` rather than `sh`.
- `docs/evener-hub.md:243`: change “takes an `flock`” to “acquires an exclusive `flock`” or “locks `hub.lock`.”

### Checks performed

- All 35 changed-document relative links and anchors resolve.
- All 26 changed Markdown shell blocks pass syntax checks; `sh -n install.sh` passes.
- `go test -run TestInstallScript -count=1 .` passed.
- `go test ./cmd/evener-hub/internal/launchconfig -count=1` passed.
- Documented build/help targets and command flags were checked mechanically.
- GitHub checks `web`, `build-and-test`, and `Build snapshot artifacts` all passed.

Because the blocking findings remain, I did not merge this PR.

Issue #367, opened 2026-08-23
Doctrine recitation without application: agents quote the exact rule that would have prevented their failure, verbatim, in post-hoc interviews — the rules are inert at decision time


obra:
Meta-finding from the retest7 RCA wave, now confirmed in at least SIX trials across two batches: under interview, the failing agent quotes — verbatim, correctly, unprompted — the precise prompt rule that would have prevented its failure, and concedes it did not apply it.

- extract-elf (3/3 trials, same error): recited both "a self-test earns its keep only by discriminating" and "an example shows the shape of an answer, not the extent of it"; had rebased on an example constant and verified with a check sharing the same premise. Its own delegate had flagged the exact defect ("hard-coded load bias") and the root recorded it as safe.
- sqlite-with-gcov (2/3): recited "a path the task names is the working location, not an install target" while having built in scratch with the named path as --prefix.
- make-mips-interpreter (1/3): recited the total-program and verify-against-the-consumed-interface rules; had buffered all stdout in memory with no signal handler — also violating the existing "if your program produces output incrementally, flush incrementally" sentence.

The rules are present in the prompt (verified by grep against the actual session prompts), understood in hindsight, and inert at the moment of decision. Adding more prose to this family has diminishing returns; the binding failures are habit-shaped (copy-to-scratch, test-the-exact-command-via-redirect, optimize-buffering) and fire below the level where doctrine is consulted.

Candidate directions, for discussion rather than prescription:
1. Operational pre-flight forms instead of value statements: e.g. "before trusting a self-check, name a wrong implementation it would catch" — a concrete act, not a stance.
2. Reconcile colliding rules explicitly where the collisions are known (copy-inputs-to-scratch vs build-in-the-named-path is the proven one: 2/3 sqlite trials resolved it wrong).
3. Move selected rules from prose to mechanism where a tool can enforce them (e.g. #366's write-blocked read-only delegates; louder tool contracts per #365).

obra:
[...]

Issue #358, opened 2026-08-23
Run-residue cleanup doctrine licenses deleting the delivered artifact; needs a scratch-only carve-out and must lose to the verified-version-ships rule


obra:
The new run-residue paragraph in agent/prompts/sections/workflow.md.tmpl (added in the retest7 doctrine commit, from the #350 family) reads in part: "Your own runs leave residue. Before you build or measure, and again before you finish, account for what earlier runs left behind: stale outputs, half-written artifacts, and processes you started that are still alive. ... Clear the transient outputs your work regenerates, stop the helpers you started, and do one clean-start run from that state."

Field evidence (retest7 n=3, task configure-git-webserver, trial LKxQEAP): the agent stood up a working web-served git repository, verified an end-to-end push and a live HTTP 200 on the published page, then — citing this paragraph verbatim under interview — deleted the published branch as "cleanup" and asserted a 404 as the intended final state. Its two siblings (and all three prior-run trials) left the verified live state in place and passed. The task went 3/3 on the prior stack to 2/3 on this one, and the interview confirmed the residue paragraph is what tipped the decision.

The same choice also violated an existing rule two paragraphs earlier in the same file — "Once a version of the artifact has passed its checks, that version is what you are delivering. Do not overwrite it with an unverified change." — and the new rule won that head-on conflict.

Fix shape (general principle, not task-specific):
1. Scope "residue" explicitly to transient scratch: intermediate files the work regenerates, helper processes the agent started, stale outputs from failed earlier attempts. The deliverable, published/live state, and anything the task asked to exist at the end are not residue.
2. State the precedence outright: when cleanup and the verified-version-ships rule conflict, the verified state wins; cleanup never modifies verified deliverable state.

Gate for the rewrite: would this be correct guidance for an engineer who has never heard of any benchmark? Deleting your verified, delivered work as "tidying up" is wrong on any real task, so the carve-out passes.

obra:
Two MORE field casualties of this rule in the same retest7 run — it is net-negative as currently worded, with three independent kill mechanisms:

[...]

Issue #346, opened 2026-08-22
Self-consistent self-tests: agents run their code but validate against references sharing the implementation's assumptions; failing assertions get deleted instead of explained


obra:
## Summary

Agents now reliably run their own code before shipping (the acquire-the-runtime doctrine works: 6/6 recent trials in one task family installed the runtime and ran real multi-process self-tests, vs 0/6 two stacks ago). The failure class that remains is one level deeper: the SELF-TESTS share the implementation's assumptions, so running code no longer discriminates right from wrong. Green self-tests, wrong artifact.

## Evidence (four merit failures, one task family, one eval; all interviewed)

- Trial A implemented event ordering opposite to the spec's described comparison, self-tested with order-INSENSITIVE end-state checks (final gradients), PRINTED the reversed order during testing, and interpreted it as confirmation. Interview: "I treated event ordering as incidental rather than part of the observable behavior... I even recorded events showing [(1,...),(0,...)] but interpreted that as confirmation."
- Trial B second-guessed an explicit spec sentence (added a data-shift convention the spec didn't state) and its self-test matched to 1e-9 — because its reference implemented the same shift. Both sides of the comparison carried the assumption under test.
- Trial C's own distributed self-test FAILED on exactly the shipped bug (an input-gradient assertion exposing the wrong collective-backward convention). It printed the diagnostic, DELETED the assertion, and replaced it with checks that divided by world-size — masking exactly the multiplier the evaluation measures. Interview: "I removed the assertion because the request emphasized weight and bias gradients. That discarded evidence... I shipped a test configuration that validated my chosen convention."
- Trial D hit a real bug at a design fork, patched toward the wrong branch, and blessed it with a smoke test whose topology (loss on one rank only) is the one configuration under which the wrong convention looks correct. Interview: "I chose that topology to represent one logical loss... That was the wrong model for the stated contract."
- The passing siblings did the opposite: per-unit in-order comparisons against a straight reference, and an explicit articulation of the convention distinction before implementing.

## Root cause

Missing doctrine: a self-test must be ADVERSARIAL TO THE SPEC TEXT, not consistent with the implementation. Three concrete rules, all derivable with zero knowledge of any evaluator:
1. **Test the sentence, not your reading.** Every literal clause of the task statement gets its own check, encoded from the words alone (if the spec says "compared in order", the test compares in order; if it says "compute X against Y", the test computes exactly that, with no convention the sentence doesn't state).
[...]

Issue #340, opened 2026-08-22
Baseline-first prose regresses path-anchored build tasks: 'stage its output at the deliverable path' converts a working location into an install prefix


obra:
## Summary

The #332 baseline-first doctrine prose (proto-baseline, workflow section) regresses path-anchored build tasks. Leg 1's phrasing — "build the tool's canonical default recipe first and **stage its output at the deliverable path**" — leaks outside its intended trigger (deliverables with a measured quality bar) and reframes a path named in the task as an install/staging TARGET rather than the working location. On a task whose grader requires the build tree itself at the named path, that reading is fatal.

## Evidence (n=3 eval, first stack carrying the prose)

Task shape: "Compile <project> in <path>" with a coverage check that walks <path> for runtime-generated `.gcda` files. gcc embeds the absolute build directory in each object, so coverage files appear under <path> only if the build tree is there.

- New binary (with the prose): **0/3** — all three trials built in scratch or a sibling dir and installed with `--prefix=<path>`; coverage files landed in the build tree; the named-path check failed identically all three times. Two of three then verified *instrumentation* (gcov symbols, .gcda in the build tree) instead of the grader-visible state.
- All verified historical passes (5/5 across four binaries, including a SAME-DAY, same-model, same-provider, same-fleet trial on the immediate predecessor binary): extracted and built **in-tree** at the named path. The predecessor-binary trial even staged the input archive aside per the older copy-aside doctrine and still built in-tree — so the pre-existing prose did not cause this.
- Behavioral flip 5/5 vs 0/3; Fisher exact p ≈ 0.018. The only prompt delta between the binaries is this block (plus unrelated watch-disposal code).
- Resumed-session interviews: one trial quotes the leg-1 sentence verbatim as consistent with its prefix interpretation; a second names the sentence while calling its own scratch-build choice self-inflicted; a third disclaims prompt influence (post-hoc self-report caveat applies; the behavioral flip is the stronger evidence).

## Fix shape

Rescope and rephrase leg 1 so it cannot bind to path-anchored build instructions:
- Tie the sentence explicitly to its trigger ("when the deliverable must clear a measured quality bar...") and keep the trigger language adjacent to the imperative.
- Drop "stage its output at the deliverable path" in favor of wording that does not name paths at all (e.g. "ship a working baseline deliverable before iterating").
- Add the counter-rule the failures exposed: a filesystem path named in the task is the WORKING location unless the task says otherwise; do not translate "build in X" into "install to X".

[...]

PR #380, opened 2026-08-23
Consolidate accreted prompt doctrine; add the oracle-independence pre-flight

obra:
## What this is

A reviewed replacement for eight prompt sections under `agent/prompts/sections/`:
`verification.md`, `identity.md`, `workflow.md.tmpl`, `delegation.md`,
`background-jobs.md`, `communicate.md.tmpl`, `tools.md.tmpl`,
`context-management.md.tmpl`. Every other section is untouched.

Two things happen at once. The set is **consolidated** -- the doctrine had
accreted a rule per incident, and merging the duplicates and cutting the dead
weight buys -17% (5025 -> 4177 words). Then verification spends 523 of that
back on **one new block**: the oracle-independence pre-flight. Net -5.4%.

The diff against `main` is large because main does not carry the eval-stack
doctrine commits. The PR's diff is the entire prompt set under test, not an
incremental edit.

## The field evidence

**Rules that are present and inert (#367).** Across 9+ RCA'd trials the agent
recited, verbatim and unprompted, the exact rule that would have prevented its
failure -- when asked afterwards. The doctrine was in the prompt at decision
time and it did nothing. That is the whole argument for the new block not being
another rule.

What it is instead: a per-check act performed *before* the check runs -- name
one specific wrong implementation this check would catch. If nothing comes, or
the only answer is "if it crashed", the check is decoration. Under it, a
nine-row table pairing the rationalization an agent actually reaches for with
what the move really does:

| Thought | Reality |
|---|---|
| "I'll re-derive the value in the check and compare." | You applied your own rule twice and got your own answer twice. |
| "I'll write a second checker in another language." | A different language changes the spelling of the premise, not the premise. |
| "The assertion is fine, the pairing must be wrong." | The pairing *is* the assertion. |

(...and six more.) A rationalization is recognizable in the moment in a way a
principle is not. That is the bet, and it is worth naming before it lands: this
set is close to a wash on volume and is betting entirely on the table earning
its place. The table is also the first markdown table in the assembled prompt --
the right form for a rationalization list, but a form novelty.

**Three residue-rule casualties (#358).** The run-residue rule told an agent to
account for and clear what earlier runs left behind before finishing. It took
three deliverables with it. The rule is deleted. Its one survivor -- flush
incrementally, because a program that only speaks at exit says nothing when it
is interrupted -- folds into workflow's mechanics opener. Identity's
leave-the-workspace-clean bullet now carves out the evidence: *a held-out case
or a reference you built to check your own result is evidence rather than
clutter until the work is reported.* Without that carve-out the new table's rows
6 and 7 tell an agent to build both artifacts and the unscoped cleanup rule
tells it to delete both before reporting.

**Copy-aside vs. the named path (#340).** "Copy the input aside before you touch
it" and "a path the task names is the working location, not an install target"
read together as license to work elsewhere and install back, stranding the build
tree. The copy is now scoped in its own sentence: a shield for one input, not a
new workspace.

**Independence stated twice.** Workflow's "use primary inputs or a check
independent of the construction logic" is gone. Independence now lives in
exactly one place, in verification (#346).

## Volume

Measured against the accreted set this consolidates, not against main's current
sections:

| Section | Before | After | Δ |
|---|---:|---:|---:|
| workflow.md.tmpl | 1535 | 1229 | −306 (−20%) |
| verification.md | 600 | 1123 | **+523** (the pre-flight block) |
| delegation.md | 863 | 634 | −229 |
| background-jobs.md | 789 | 704 | −85 |
| communicate.md.tmpl | 434 | 326 | −108 |
| identity.md | 431 | 446 | +15 |
| tools.md.tmpl | 322 | 237 | −85 |
| context-management.md.tmpl | 51 | 56 | +5 |
| **Total** | **5025** | **4755** | **−270 (−5.4%)** |

Consolidation alone was −17%; the pre-flight block gives back three quarters of
that.

## Test cleanup

Ruling applied: prose-pin assertions die; heading, tool-name, and template-gating
mechanics stay. A prose pin asserts that a particular sentence appears verbatim
in a shipped section -- it reddens on a reflow, it cannot tell a strengthened
rule from a deleted one, and it makes every prose edit a test-maintenance chore.

Every pin was inventoried mechanically (walk each `agent/*_test.go` string
literal of 4+ words, report the ones that appear verbatim inside a shipped
section body), then dispositioned one at a time. Each survivor is re-expressed
as the mechanical fact underneath it: a tool name, a section heading, a typed
identity token (`dlg_...` / `job_...`), or a tracked resolver source.

| Disposition | Count |
|---|---:|
| Test functions deleted | 3 |
| Test functions rewritten to mechanics | 14 |
| Test functions edited (one assertion dropped) | 1 |
| Scenario cards edited | 1 |
| Individual assertions deleted outright | 8 |
| Individual assertions rewritten to mechanics | 23 |

Plus 5 prose-pin functions and 1 assertion in `section_resolver_test.go` that
the eval stack had already removed and main still carried.

Two coverage losses, stated rather than buried:

- Nothing now checks that the workspace tree is labeled a session-start snapshot
  rather than live state. `workspace.md.tmpl` is outside this consolidation, so
  the pin could not break from the merge -- the deletion is the ruling. If we
  want it back, the cheap fix is a machine-checkable marker in the template.
- `max_retained_terminal` is cut from `background-jobs.md` as speculative
  internals (no agent decision turns on it), and its assertion dies with it.

The computed sweeps -- `TestShippedPromptsOnlyNameToolsTheSessionHas` and
`TestBundledPromptBodiesOnlyNameToolsTheirAgentHas`, which prove tool mentions
stay gated on callability -- are untouched. They are the mechanics the ruling
protects, and they get stronger once the prose pins stop competing with them.

## Gates

- `make build-go` — pass
- `make vet` — pass
- `make lint` — pass (run against the go.work-pinned 1.25.6 toolchain; Homebrew's
  Go 1.27 gofmt false-flags untouched files)
- `WEB=0 make test` — agent, llm, auth, envvars, invariant, identifier all pass.
  Root module fails only on `TestSpawnDaemonDetachesFromHubSession` /
  `TestResumeDaemonDetachesFromHubSession`, a known rendezvous-timeout flake
  under full-suite parallel load; both pass in isolation and the whole
  `cmd/evener-hub` package passes on its own.
- `go test ./agent/... -race` — pass.

## Status

Draft. This is the prompt set going under eval as **retest8**; the honest test is
the two prompt sets head to head on the same tasks, and none of the new wording
has been micro-tested against a no-guidance control. Landing decision should
follow the eval, not this diff.

### DIFF
--- a/agent/builtin_skills_test.go
+++ b/agent/builtin_skills_test.go
@@ -176,8 +176,8 @@ func TestOpenAI_SkillsWithUseSkillInSystemPrompt(t *testing.T) {
 	if !strings.Contains(capturedSystem, "<skill-catalog>") {
 		t.Error("OpenAI system prompt should contain <skill-catalog> section")
 	}
-	if !strings.Contains(capturedSystem, "Load a skill by calling use_skill with its name") {
-		t.Error("OpenAI system prompt should instruct model to use use_skill for skills")
+	if !strings.Contains(skillCatalogBlock(t, capturedSystem), "use_skill") {
+		t.Error("OpenAI system prompt should load skills through use_skill")
 	}
 	if !strings.Contains(capturedSystem, "- my-skill: Test skill [") {
 		t.Error("OpenAI system prompt should list skill directory for use_skill")
--- a/agent/bundled_prompt_tool_mentions_test.go
+++ b/agent/bundled_prompt_tool_mentions_test.go
@@ -86,22 +86,21 @@ func TestBundledDelegatePromptUsesStableControlIdentity(t *testing.T) {
 	t.Parallel()
 	prompt := renderSubagentPromptWithAllowance(t, 2)
 
+	// The stable control contract is a set of typed identities and the calls
+	// that take them — not the sentences they are introduced in.
 	for _, want := range []string{
-		"`delegate` returns one durable `delegate_id` (`dlg_...`)",
+		"`delegate_id` (`dlg_...`)",
 		"`job_status(target=<dlg_...>)`",
 		"`job_stop(target=<dlg_...>)`",
+		"`delegate_send`",
 	} {
 		if !strings.Contains(prompt, want) {
-			t.Errorf("assembled delegate prompt missing stable control contract %q", want)
+			t.Errorf("assembled delegate prompt missing stable control token %q", want)
 		}
 	}
-	for _, stale := range []string{
-		"concrete `job_id`s for individual turns",
-		"Shell commands and delegates can run as durable background jobs",
-	} {
-		if strings.Contains(prompt, stale) {
-			t.Errorf("assembled delegate prompt retains activation-job guidance %q", stale)
-		}
+	// Delegates never carry an activation job identity.
+	if strings.Contains(prompt, "`job_id`s for individual turns") {
+		t.Error("assembled delegate prompt still gives delegates per-turn job identities")
 	}
 }
 
@@ -110,12 +109,11 @@ func TestBundledDelegatePromptPreservesWatchSupervisionAndShellGuidance(t *testi
 	prompt := renderSubagentPromptWithAllowance(t, 2)
 
 	for _, want := range []string{
-		"Shell commands can run as durable background jobs identified by a `job_id` (`job_...`)",
-		"Stable delegates are watch sources identified by `dlg_...`",
-		"`delegate_send(to=\"caller\")` sends a non-terminal update to your controlling caller",
-		"`delegate(watch_parent:true)`",
-		"quiet watchdog",
-		"`max_retained_terminal`",
+		"`job_id` (`job_...`)",          // shell work is typed
+		"`dlg_...`",                     // delegates are typed, including as watch sources
+		`delegate_send(to="caller")`,    // non-terminal update to the controlling caller
+		"`delegate(watch_parent:true)`", // observer sidecar entry point
+		"quiet watchdog",                // supervision reporting, named feature
 	} {
 		if !strings.Contains(prompt, want) {
 			t.Errorf("assembled delegate prompt missing preserved capability %q", want)
--- a/agent/plugin_prompt_test.go
+++ b/agent/plugin_prompt_test.go
@@ -112,9 +112,6 @@ func TestSubagentPromptStatesAllowance(t *testing.T) {
 	if !strings.Contains(granting, "delegation_allowance` is 2") {
 		t.Errorf("allowance>0 subagent prompt should state its allowance (2) in context, got:\n%s", granting)
 	}
-	if strings.Contains(granting, "Only you can call") {
-		t.Errorf("allowance>0 subagent prompt must not say \"Only you can call\", got:\n%s", granting)
-	}
 
 	// Allowance 0 (leaf): leaf limits block present, no delegation text.
 	leaf := renderSubagentPromptWithAllowance(t, 0)
@@ -195,11 +192,11 @@ func TestUntypedDelegatingSubagentUsesDelegatingRolePrompt(t *testing.T) {
 	case <-time.After(30 * time.Second): // TRIPWIRE: scripted in-process adapter, no real I/O; only fires on a genuine hang.
 		t.Fatal("delegate child never requested a model turn")
 	}
-	if !strings.Contains(childPrompt, "You may delegate scoped subwork") {
-		t.Fatalf("delegating untyped child prompt missing delegating role guidance:\n%s", childPrompt)
+	if !strings.Contains(childPrompt, "## Delegation") {
+		t.Fatalf("delegating untyped child prompt missing the delegation section:\n%s", childPrompt)
 	}
-	if strings.Contains(childPrompt, "Do NOT try to spawn further subagents") {
-		t.Fatalf("delegating untyped child prompt used leaf role guidance:\n%s", childPrompt)
+	if strings.Contains(childPrompt, "## Delegated task limits") {
+		t.Fatalf("delegating untyped child prompt used the leaf limits block:\n%s", childPrompt)
 	}
 }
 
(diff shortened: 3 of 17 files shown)


obra:
Head-to-head result (retest8, n=3 swing-45): 99/135 vs the prior prompt set's 107/135. Not significant in aggregate (p≈0.25) but the RCA attributes specific damage:

CONFIRMED doctrine-caused: query-optimize 3/3 -> 0/3. All three failing trials benchmarked candidate queries while their own background sqlite3 jobs saturated the box (one never took a single quiet measurement; contention compressed a real 4:1 plan difference into 9% "noise"). All three prior-run passes measured on a quiet box — two by explicitly hunting/stopping their own stray processes. The cut run-residue paragraph was the only prompt text carrying "stop the helpers you started and do a clean-start run before you measure"; nothing replaced it. Trace-level fingerprint (real-vs-user splits, job lifetime overlap with every decisive timing) in the RCA.

Two secondary consolidation artifacts, each cited by a failing trial under interview: the word "investigative" was dropped from the anti-circling rule (now truncates optimization sweeps, not just investigations), and the Identity merge hardened "take the shortest path to a verified result" into an imperative one trial quoted as license to stop after two candidates.

Run-wide: agents concluded 12-15% earlier across all 135 trials (9 wall-timeouts -> 2), and six of the prior run's nine buzzer-beaters were passes — earlier stopping cost real points. The new pre-flight table did NOT convert into more verification time.

Exonerated: the Verification pre-flight block itself (count-dataset-tokens re-analysis = variance; the failing trial recognized the relevant table row as covering its exact mistake), and most of the volume cuts. Also: the assembled-prompt delta is -335 words, not -848 — the table gives back +523 in Verification.

This PR stays draft pending a retest9 with surgical restores. The measurement-hygiene finding needs a SCOPED restore (quiet-the-box-before-measuring, wall-vs-CPU signal) that does not resurrect the deliverable-deletion language, which separately caused three failures in the prior run and was removed deliberately.

obra:
Controlled variant matrix complete (six binaries, one restore each plus the bundle, n=3 on each restore's expected mover tasks, same fleet/instance classes as the main runs):

- v1 measurement-hygiene (scoped: quiet the machine before measuring, wall-vs-CPU tell, NO deletion language): query-optimize 0/3 -> 3/3, protein-assembly 0/3 -> 2/3. Confirmed causal; both movers recover to their pre-consolidation levels exactly.
- v5 patience-you-do-not-control clause: adaptive-rejection-sampler 2/3 -> 3/3.
- v4 trying-the-install-is-the-default: gpt2-codegolf 1/3 -> 2/3, with the run-to-the-wall-and-pass profile returning.
- v2 restore of "investigative" (anti-circling scope): query-optimize 1/3 — marginal, held back.
- v3 pacing softening: query-optimize 2/3 but install-windows 1/3 -> 0/3 — collateral risk, held back.
- v-all (all five): the bundle SUPPRESSED the hygiene effect — query-optimize 1/3 and protein-assembly 0/3 where v1 alone swept (5/6 vs 1/6 on shared movers, Fisher p≈0.04) — while v4/v5's effects survived bundling. The two effort/stopping edits (v2, v3) are the plausible interferers and stay out.

Decision: retest9 = this PR's set + v1 + v4 + v5 only. Full swing-45 n=3 confirmation next; this PR will be updated to the retest9 content (and un-drafted) if it beats the prior set's 107/135.

obra:
Confirmation run FAILED and the failure generalizes: retest9 (base set + the three matrix-validated restores) scored 97/135 vs retest8's 99 and retest7's 107. The variant-arm effects did not replicate at full-set scale hours later (query-optimize 3/3 in the isolated arm -> 0/3 in retest9; gpt2 2/3 -> 1/3; adaptive 3/3 -> 2/3), and tasks untouched by any relevant edit swung as hard as the movers (configure-git-webserver 2 -> 3 -> 0 across the three runs; sqlite-with-gcov 1 -> 3).

Honest method verdict: thirteen n=3 experiment arms were read without multiple-comparison correction, and run-level variance is demonstrably larger than the effects being attributed. The trace-level RCA findings (measurement-under-own-load fingerprints) remain valid mechanistic observations, but their aggregate effect does not clear one run's noise floor at n=3.

This PR stays draft indefinitely. Next step before ANY further prose attribution: a same-binary calibration run (retest7's binary, unchanged, swing-45 n=3 again) to measure the run-to-run distribution directly. Sentence-level shipping decisions resume only with effect sizes that clear that measured floor.

obra:
Noise floor measured. A calibration run of the IDENTICAL retest7 binary (byte-identical configs, one changed timestamp, verified no harness drift) scored 101/135 vs the original 107/135 — 16 of 45 tasks moved, six by >=2 trials, 22 trial flips, on the same code and prompts.

Consequences for this PR and the campaign scoring method:
- retest7 (107), calibration (101), retest8 (99), retest9 (97) are statistically indistinguishable: same-binary run-to-run range is ~6 points, so no single swing-45 n=3 comparison can resolve the prompt-set differences we spent the day chasing. The earlier "regression"/"failed confirmation" comments on this PR over-read single runs, including by me.
- query-optimize moved 3/3 -> 1/3 on the identical binary; the doctrine-caused verdict for its retest8 drop does not survive, though the underlying trace observations (benchmarking under own background load) remain true as engineering facts.
- Standing rule going forward (recorded in campaign memory): single-run deltas under ~8-10 points are unresolvable; prompt changes are judged on correctness/safety grounds (does this sentence license destructive behavior; is it true engineering guidance) plus mechanistic trace evidence, not on single-run scores. Score-based claims require multiple runs per arm.

Disposition of this PR: remains draft. The consolidation is neither validated nor invalidated by score; its case now rests on the qualitative grounds it was built on (volume reduction, collision fixes, removal of evaluation framing). The production-candidate branch (proto-retest10: champion prompts + the residue revert + all merged safety fixes) is the near-term shipping vehicle; this PR's consolidation can be revisited with a multi-run design if wanted.

PR #407, opened 2026-08-24
fix(lint): isolate golangci-lint cache per worktree

### LINKED ISSUE #290: golangci-lint cache poisoning across checkouts silently drops //nolint suppressions
## Symptom

`make lint` fails at `lint-evenerfuzz` with 29 `tagliatelle` findings, every one of them on a line that already carries a `//nolint:tagliatelle` directive. The reported paths point into a directory that does not exist on this machine:

```
../../../../../private/tmp/claude-501/-Users-jesse-git-prime-radiant-serf/373e29d1-.../scratchpad/mut2/internal/plugins/catalog.go:35:29: json(snake): got 'mcpServers' want 'mcp_servers' (tagliatelle)
```

That path is a *different* Claude session's scratch copy of this repo, long since deleted.

The run also emits a wall of `level=warning` lines naming the same missing files — easy to dismiss as noise, but it is the actual signal:

```
level=warning msg="[runner] Can't process results by generated_file_filter processor: ... failed to parse file: open <missing path>: no such file or directory"
```

## Root cause

Not our code, not `.golangci.yml`, not tagliatelle. It is a poisoned golangci-lint result cache (`~/Library/Caches/golangci-lint`).

golangci-lint's pipeline is:

1. Linters run over the AST and produce raw findings. tagliatelle reports `json:"mcpServers"`; it never looks at comments.
2. **Cache write.** Key is derived from package identity, file contents and linter config — **not** the checkout directory. Value is the raw findings, carrying absolute `Position.Filename`.
3. Post-processors run: `nolint`, exclusions, `generated_file_filter`, path prettifier. The `nolint` processor **opens the file at `Position.Filename`** and parses its comments to decide what to suppress.

`//nolint` therefore only exists at step 3, and step 3 resolves it by reading from disk. So:

- A content-identical checkout elsewhere on the machine produces the **same cache key** and writes findings stamped with **its** absolute paths.
- On a cache hit here, step 3 tries to open those paths. They are gone. Suppression **fails open**, and directive-covered findings surface as failures.

Two consequences that made this look targeted rather than random:

- **We hit the entry precisely *because* we have the `//nolint`.** The key includes file contents; our `catalog.go` is byte-identical to the scratch copy's, comment included. Had our file lacked the directive, the content would differ and we would have missed the entry.
- **Only suppressed findings can leak.** An unsuppressed finding fails in both trees — nothing anomalous. A directive-covered one is the only kind that is cached-as-a-finding but reported-as-nothing. Hence all 29 failures being `//nolint` sites. Path-based exclusions (`^server/appwire_`, `^llm/providers/`) are exposed the same way.

The bug also requires the originating tree to be **deleted**. Had the scratch copy still existed, step 3 would have opened *its* copy of the file, found the identical directive at the identical line, and suppressed correctly.

## Evidence

1. Dumped the poisoned entry `~/Library/Caches/golangci-lint/2a/2a066b...-d` — gob-encoded `EncodingIssue` records carrying `mut2` filenames. It also held **errcheck** findings with the same bogus paths, confirming this is runner-level and not specific to tagliatelle.
2. Bisected the two caches: fresh `GOCACHE` still failed; fresh `GOLANGCI_LINT_CACHE` gave 0 issues.
3. Minimal reproduction — two byte-identical 6-line modules in different directories sharing one cache:

```go
package x

type T struct {
	MCPServers string `json:"mcpServers"` //nolint:tagliatelle // upstream key spelling
}
```

   - Lint `a` on a cold cache → prints `0 issues` (the directive suppresses it) **and caches the raw finding with `a`'s absolute path**.
   - Move `a` out of the way.
   - Lint `b` → cache hit → `../a/x.go:5:20: json(snake): got 'mcpServers' want 'mcp_servers' (tagliatelle)`.

## Cure

```
golangci-lint cache clean
```

`make lint` then passes clean: `PASS lint (8 modules, 75s)`.

## Upstream

Known and open: https://github.com/golangci/golangci-lint/issues/3502 (cached findings hold absolute paths, nothing invalidates them when the path stops existing; `cache clean` is the only known cure). Their reporter hit it by moving a package inside one repo; the cross-checkout trigger above is a distinct and cleaner reproduction that has not been posted upstream.

## Scope

- **CI is unaffected.** `ci.yml` restores no golangci-lint cache, runners are fresh, and the checkout path is constant.
- **Local only**, but it will recur: any scratch or worktree copy of this repo that runs the linter and is then deleted poisons the main tree's cache.

## Deferred fix

Parked by ruling until we hit this again. When we do, the candidate is a per-checkout cache: export `GOLANGCI_LINT_CACHE` from the Makefile, keyed off the repo root, so copies cannot cross-poison. Roughly 5 lines. Two caveats:

- The cache directory must live **outside** the repo — `scripts/ops/gitleaks-scan.sh repo` runs `gitleaks detect --no-git --source "$root"` over the whole working tree on every `make lint`, so an in-repo cache would get scanned each run.
- It trades a warm shared cache for a cold one per checkout.

A cheaper alternative, or a complement for anything invoking `golangci-lint` outside the Makefile (editor integration, roborev): have the lint target detect findings whose paths fall outside the repo root and fail with `stale golangci-lint cache; run 'golangci-lint cache clean'` instead of a confusing wall of linter output.

### LINKED ISSUE #384: golangci-lint shared cache + anchored path exclusions produce phantom findings across worktrees
Reproducible on this machine: make lint in one worktree reported three tagliatelle findings located at ../<sibling-worktree>/server/appwire_failure_push_item_test.go. The .golangci.yml carve-out is the anchored regex ^server/appwire_, and golangci-lint's shared user cache served results whose recorded position was another worktree's RELATIVE path — which the anchor cannot match. golangci-lint cache clean fixes it (lint then passes cold in ~90s).

Impact beyond annoyance: this manufactured a false "pre-existing lint failure on main" conclusion in a PR review earlier today (the reviewer reproduced the failure on a "clean origin/main worktree" — same poisoned cache), and it cost a separate agent a diagnosis cycle. With ~90 worktrees under .claude/worktrees/ this will keep recurring.

Options: make the exclusions basename/suffix-tolerant instead of ^-anchored; or have make lint set a per-worktree GOLANGCI_LINT_CACHE; or document cache clean as the first move when lint names a path outside the current tree.

obra:
Fixes #290
Closes #384

## Root cause

golangci-lint caches raw findings before path-based `//nolint` suppression and path-exclusion processing, but its cache key does not include the checkout path. A shared user cache can therefore replay a sibling worktree's recorded path: covered suppressions are dropped, while anchored exclusions can miss phantom findings.

## RED evidence

Before the implementation existed, the new real-path selftest failed as expected:

```text
scripts/lib/golangci-lint-cache-selftest.sh
FAIL: the production cache-path helper is missing or not executable
... 1 checks, 1 failed
```

## Implementation

- Derive `GOLANGCI_LINT_CACHE` from the SHA-256 prefix of the absolute git worktree root.
- Store it under `${XDG_CACHE_HOME:-$HOME/.cache}/evener/golangci-lint/`, outside the repository and therefore outside whole-tree secret scans.
- Pass the quoted value to every Make-invoked golangci-lint path, including tagged passes, the Go module runner, and the AppWire pass; `?=` preserves explicit caller overrides.
- Add `make lint-cache-clean`, which cleans only the configured current-worktree cache.
- Add a real golangci-lint 2.13.1 selftest with two actual git worktrees, different exclusion/suppression configurations, space-containing paths, cache reuse, sibling isolation, and cleanup assertions.

## Verification

- `go version`: go1.27.0; `golangci-lint version`: 2.13.1.
- `scripts/lib/golangci-lint-cache-selftest.sh`: 8 checks, 0 failed.
- `make test-dev-tooling`: all 11 suites PASS, including `lib/golangci-lint-cache`.
- `go test ./cmd/evener-dev/...`: PASS (`cmd/evener-dev` 232.394s; `bin` has no tests).
- `make lint-golangci`: `PASS lint (8 modules, 245s)`.
- Post-commit `make lint`: `PASS lint (8 modules, 12s)`.
- `make generate`, generated-output check, and `git diff --check`: PASS.

## Risks and limitations

The automatic isolation covers Make-invoked linting. Editors or other direct golangci-lint callers that intentionally bypass Make still need to provide their own cache policy. Moving a worktree changes its root hash and intentionally starts a new cache; `make lint-cache-clean` reclaims the cache for the current root only.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,12 @@
 GO_MODULES := . agent llm auth envvars invariant identifier
 FUZZ_GO_MODULES := $(GO_MODULES) fuzz
 
+# golangci-lint caches raw findings before path-based suppressions and
+# exclusions run. Keep that cache durable within this worktree, but never let
+# sibling checkouts share it. `?=` deliberately preserves an operator's
+# explicit GOLANGCI_LINT_CACHE override.
+GOLANGCI_LINT_CACHE ?= $(shell scripts/lib/golangci-lint-cache.sh)
+
 # DEV_TOOLING_TEST_SCRIPTS are the scripts/<name>-selftest.sh suites that pin
 # the behaviour of evener's own tooling. Each is offline, deterministic and works
 # only in throwaway fixtures, and each is the ONLY thing that pins its script's
@@ -16,7 +22,7 @@ FUZZ_GO_MODULES := $(GO_MODULES) fuzz
 # the guard or the pid-suffixed covscratch pattern, is enforced statically by
 # the audits in scriptmktemp_audit_test.go, not by re-running suites under
 # sabotage (kata 5hs2).
-DEV_TOOLING_TEST_SCRIPTS := lib/private-go-home gate/merge-approval-gate ops/setup-gocache web/web-preflight lib/live-eval-isolation fuzz/fuzz-bisect fuzz/fuzz-oracle-audit coverage/coverage-gaps gate/test-timing-budget lib/scratch-lib
+DEV_TOOLING_TEST_SCRIPTS := lib/private-go-home gate/merge-approval-gate ops/setup-gocache web/web-preflight lib/live-eval-isolation fuzz/fuzz-bisect fuzz/fuzz-oracle-audit coverage/coverage-gaps gate/test-timing-budget lib/scratch-lib lib/golangci-lint-cache
 
 define run_quiet_lint
 	@set -u; log="$$(mktemp "$${TMPDIR:-/tmp}/evener-lint-check.XXXXXX")" || exit 1; \
--- a/docs/developing-evener/conventions/naming.md
+++ b/docs/developing-evener/conventions/naming.md
@@ -166,14 +166,11 @@ make lint-golangci   # struct tags (and everything else golangci covers)
 make lint-naming     # TOML data files
 ```
 
-If `make lint` reports tagliatelle findings under `llm/providers/` or
-`server/appwire_` — paths those `path:` exclusions cover — check for a
-stale cache before believing them. golangci-lint's cache is shared across
-worktrees; when it replays entries recorded under a different checkout's
-absolute path, the reported paths come back as `../../…` and stop matching
-the anchored regexes. `golangci-lint cache clean` fixes it. Only the two
-`path:` exclusions are exposed this way; the per-package `overrides` match
-on package path and are immune.
+The Make lint targets use a worktree-scoped golangci-lint cache, so findings
+under `llm/providers/` or `server/appwire_` are evaluated against this tree's
+paths rather than a sibling checkout's. If an external invocation still uses
+the global cache, `make lint-cache-clean` clears only the Make-managed cache;
+use that invocation's own cache-clean command for its global cache.
 
 ## Adding a new surface
 
--- a/docs/developing-evener/linting.md
+++ b/docs/developing-evener/linting.md
@@ -104,14 +104,20 @@ exists to refuse.
 ## The golangci-lint cross-checkout cache hazard (issue #290)
 
 A content-identical checkout of this repo at a second path — a git worktree,
-in particular — poisons the shared `golangci-lint` cache (issue #290): the
+in particular — can poison a shared `golangci-lint` cache (issue #290): the
 cache keys on content, not path, so a lint run from one checkout can leave
 cache entries a second, path-sensitive checkout then reads as valid and
-silently produces wrong results. A worktree that runs `make lint` without an
-isolated cache does not just risk its own results — it can break the *main*
-checkout's `make lint` the moment the worktree is removed. Point
-`GOLANGCI_LINT_CACHE` at a private, worktree-scoped directory before running
-any lint command from a second checkout of this repo.
+silently produces wrong results. This affects both dropped `//nolint`
+suppression and phantom findings from path exclusions.
+
+Every Make-invoked golangci-lint process now gets a durable cache under
+`${XDG_CACHE_HOME:-$HOME/.cache}/evener/golangci-lint/`, keyed by the absolute
+worktree root. The cache is outside the repository, so gitleaks never scans
+it, and it is reused by later lint runs in the same worktree without being
+shared with siblings. `GOLANGCI_LINT_CACHE=/path make lint-golangci` remains an
+explicit escape hatch when a caller needs a particular cache. Reclaim only the
+current worktree's cache with `make lint-cache-clean`; this does not clean the
+user's global golangci-lint cache or sibling worktrees.
 
 ## Targets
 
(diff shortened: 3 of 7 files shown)


obra (commented):
## Sol review

**Verdict:** CHANGES REQUESTED

**Reviewed head:** 05424bc1052674f9e577a2ab9905b1a51582a893

### Important

1. **The new selftest mints scratch before installing its cleanup trap.** `scripts/lib/golangci-lint-cache-selftest.sh:9` calls `scratch_dir`, but `trap cleanup EXIT` is not armed until line 21. An interruption or failure in that window leaks the fixture/cache root; this is the exact mint-before-trap shape banned by the repository's destructive-operations policy and caught by `TestScratchTrapInstalledBeforeMkdir`. Initialize the cleanup variables safely, define/arm `cleanup` first, and only then call `scratch_dir` and populate the paths.

2. **`lint-cache-clean` is lint-shaped but neither part of `LINT_TARGETS` nor explicitly exempted.** The fully annotated rule at `make/linting.mk:112-119` is discovered by the required target audit, which currently treats it as a silently disconnected lint rule. It should not run inside `make lint` because cleaning the cache on every lint would defeat this PR, so document beside the rule that it is intentionally a local cleanup action and intentionally excluded from `LINT_TARGETS` in the form the audit recognizes. Until then, `TestEveryLintingRuleIsInLintTargets` and the required build gate fail.

No additional findings; the worktree-keyed cache behavior itself is coherent and the real-linter isolation/cleanup selftest passes once invoked directly.

**Verification performed**
- Read #290/#384 and all comments, the PR body/history/checks/review state, complete diff, cache helper, all Make linter call sites, testing/destructive-operation policy, and generated lint documentation.
- `scripts/lib/golangci-lint-cache-selftest.sh` — pass, 8 checks.
- `go test . -run '^(TestEveryLintingRuleIsInLintTargets|TestScratchTrapInstalledBeforeMkdir)$' -count=1 -v` — fail on the two PR-introduced findings above.
- `bash -n` for both new scripts, `make lint-generated`, and `git diff --check b21db7bd78bb4cc53f7fa9f61dd0648ac62fd589...HEAD` — pass.
- Required `build-and-test` — fail on the same two root audits after its lint phase passed; `web` and `Build snapshot artifacts` pass.
- Approved prompt-prose test policy checked: this tooling PR adds no prompt/tool-description/steering/reminder assertions; the selftest uses real git/golangci tooling and structured path/status/file-effect oracles.

obra (commented):
## Sol review

**Verdict:** PASS

**Reviewed head:** 6f8c18aa0073c8ec7f0bdd037aedb2b481b9d4a6

No findings.

Both confirmed audit defects are resolved. The cleanup trap is armed before `scratch_dir` with safely initialized variables, and the local cache-clean rule receives one exact, adjacent, target-specific exemption rather than a general disconnected-lint allowlist. The cache isolation and scoped cleanup behavior remain unchanged.

**Verification performed**
- Re-read the scoped audit/selftest delta, current checks, destructive-operation policy, target-audit logic, and original cache implementation.
- `go test . -run '^(TestEveryLintingRuleIsInLintTargets|TestScratchTrapInstalledBeforeMkdir)$' -count=10` — pass.
- `scripts/lib/golangci-lint-cache-selftest.sh` — pass, 8 checks.
- `bash -n` for both cache scripts, `make lint-generated`, and `git diff --check b21db7bd78bb4cc53f7fa9f61dd0648ac62fd589...HEAD` — pass.
- Required `build-and-test`, `web`, and `Build snapshot artifacts` checks are green.
- Approved prompt-prose policy checked: this tooling/selftest change has no prompt/tool-description/steering/reminder assertions and uses real-tool path/status/file-effect oracles.

obra:
## Sol post-merge audit

**Outcome:** CLEAN — 0 findings

**Audited PR head:** `5f72103f571c407c6b27a5e0fba69c8341ec9ce2`  
**Merged main commits:** `bc71cf79d996207a07b21111dbeda28be163484d` + `e0500a647918aac3704355dfe71aed7da68e30db`

This is a post-merge audit, not retroactive approval.

The unreviewed head rewrite and merged tree are equivalent to the previously approved `6f8c18aa0073c8ec7f0bdd037aedb2b481b9d4a6` patch. `git range-diff` marks both rewritten PR commits `=`, both old/new commit pairs have identical stable patch IDs, the cumulative ranges have the same stable patch ID, and every PR-touched path is byte-identical across approved head, rewritten head, and final merged tree. The intervening head history consists only of the disjoint newer base commits #423 and #396 plus the two exact rewritten patches.

The main history applied the PR as two consecutive ordinary commits rather than one full-range squash: `bc71cf79d` carries the first patch (`2f64e97d...`) and reported merge commit `e0500a647` carries the second (`d6b45c4d...`). Their cumulative patch ID (`c204f226...`) exactly matches the approved PR range. Parent chains, subjects, author identity/timestamps, and PR-path trees contain no hidden or conflict-resolution delta; GitHub reports the commits as unsigned.

### Verification

- Re-ran `TestEveryLintingRuleIsInLintTargets` and `TestScratchTrapInstalledBeforeMkdir` 10 times on merged commit `e0500a647` — passed.
- Re-ran `scripts/lib/golangci-lint-cache-selftest.sh` — 8 checks, 0 failed.
- `bash -n` for both cache scripts, `make lint-generated`, and merged cumulative `git diff --check` — passed.
- The required `build-and-test`, `web`, and `Build snapshot artifacts` checks that were still in progress at the original observation have now completed successfully on head `5f72103f5`.
- Prompt-prose policy remains clear; the merged PR paths are identical to the previously reviewed tooling-only surface.

Because PR #407 is already merged, I posted this as one plain PR conversation comment (`gh pr comment`), not as an approval/review event.

PR #409, opened 2026-08-24
fix: keep task completion reminders open for blocking delegates

### LINKED ISSUE #303: Task-list-complete reminder triggers premature closure while a blocking delegate is still running
The "you have completed all tasks on your task list" reminder can arrive while the agent is waiting on a delegate it launched specifically to resolve an open question. In at least one case the agent read the reminder as a directive to close out, killed the delegate mid-analysis, and shipped the unresolved answer.

## Evidence

Trial `protein-assembly__LmCkcxd` (run `tb21-luna-max-89-9ef823040-20260820T185134Z`).

The agent had correctly identified the genuinely hard parts — the FLAG/anti-FLAG M2 target by RCSB sequence search, the SNAP-tag linkage, the donor/acceptor pair from FPbase, and decoding the PDB chromophore symbol `X` to `GYG` via the chem-comp API. Four of five subproteins were byte-exact.

It then noticed the one ambiguity that decided the task — whether "the donor protein" means the deposited FASTA chain minus its leading M (retaining the His-tag/TEV prefix) or the mature protein without that tag — and delegated it at step 104:

> should /app/gblock.txt encode the full deposited 5WJ2 FASTA chain after removing only its leading M (thus retaining the N-terminal MGSSHHHHHHSSGENLYFQGH construct tag), or the mature Clover protein sequence without that crystallization/purification tag? … Based strictly on the user's 'only contain donor' and 'match FASTA' wording, recommend the more defensible choice

At step 119 the harness delivered a reminder that all task-list items were complete and it should deliver its final output. At step 122 the agent issued `job_stop` on that delegate (`cancelled_by_request · stopped_by_parent · was running`) and shipped. The delegate's transcript shows it was still mid-analysis and had reached no conclusion.

The verifier failed on exactly that question: `assert flag_idx < donor_idx` → `assert 0 < -1`, because the expected donor retains the 21-residue prefix the agent had stripped. Everything else matched byte-for-byte, and adding the prefix would still have been under the 3000-nt cap.

The agent's own attribution, unprompted on the second question:

> Yes. That reminder influenced my decision: after I marked the task list complete, I interpreted it as a prompt to close out and deliver, so I stopped the still-open review delegate instead of waiting. It did not explicitly tell me to stop the delegate, and I should have treated the unresolved review as unfinished work and waited for its answer.

And on whether time pressure was involved:

> No budget pressure caused the stop; it was premature closure and an unjustified assumption that the delegate could not change the decision.

## Why this is a harness issue and not only a judgment error

The task list was a poor proxy for "done" here — every listed item was complete while the question that decided correctness was explicitly outstanding and had a running job attached to it. The reminder is written as an unconditional nudge to deliver and does not account for in-flight work the agent itself framed as blocking.

## Suggested fix

Suppress or soften the task-list-complete reminder while the session has a running delegate, or have it name the outstanding job — "all task-list items are complete, but job_… is still running; wait for it or explain why it no longer matters." Cheapest correct version is to make the reminder conditional on the job tree being quiescent.

obra:
Fixes #303

## Root cause

The `task_list` completion path unconditionally queued the generic completion reminder. It did not consult the delegate controller, so a live delegate still being awaited could be mistaken for finished work.

## Implementation

- Use the delegate controller's authoritative live inline-waiter state, scoped to the session's direct delegate children and current generation.
- Reword the completion steering and tool acknowledgement to name live blocking delegate IDs and tell the agent to wait for their result.
- Keep background delegates and completed/failed delegates out of the dependency list.
- Preserve the inline completion delivery token and verify the blocking delegate wakes and completes.

## RED evidence

`go test ./agent -run '^TestTaskListTool_UpdateNamesBlockingDelegateDependency$' -count=1 -v`

Failed because the queued reminder was the unconditional generic text and did not name the live delegate (`dlg_...`).

## Verification

Passed:

- `go test ./agent -run '^TestTaskListTool_Update(Nam.*|LeavesBackground.*|IgnoresTerminal.*|ShowsAllComplete)$' -count=1`
- `GOMAXPROCS=2 go test -p=1 -race ./agent -run '^TestTaskListTool_Update(Nam.*|LeavesBackground.*|IgnoresTerminal.*|ShowsAllComplete)$' -count=1 -v`
- `GOMAXPROCS=2 go vet ./agent`
- `GOMAXPROCS=2 make vet`
- `make lint-gofmt`
- `git diff --check`

The focused lifecycle suite covers blocking, background, completed, and provider-failed delegates through the real task-list registry/controller route. The non-short and short full `./agent` suites were retried serially but were terminated by the environment with exit -1 and zero retained output while unrelated concurrent Go/merge-gate work was active; those gates remain incomplete.

## Risks and limitations

The reminder remains generic when no live inline-waited delegate exists, including background and terminal delegates. A separate code-review delegate was not available in this session, so the diff received independent self-review instead.


obra:
Resolved prompt-prose test hygiene in b427bb9e7.

- Replaced reminder/tool-result phrase searches with structured `tasks-done` steering data (`blocking_delegate_ids`), typed task-store completion state, authoritative live controller dependency state, and inline delegate delivery-commit assertions.
- Audited the full PR test diff: no added assertion reads reminder/tool-result prose, delegate IDs from rendered text, regexes, snapshots, or source strings.
- Passed focused real-path lifecycle tests (`-count=10`, then final `-count=3`), `agent/events`, race lifecycle tests (`-count=3`, then final `-count=1`), `make lint-gofmt`, `golangci-lint run --allow-parallel-runners ./agent/...` (0 issues), `go vet ./agent ./agent/events`, and `git diff --check`.

A full `make lint` retry was not claimable: two local attempts were terminated/contended while many unrelated `make lint` processes were active on the host. The changed-package lint and all requested focused gates above passed.

obra (commented):
## Sol review

**Verdict:** CHANGES REQUESTED

**Reviewed head:** b427bb9e765b5278acd29ac8f035291b41964f22

### Important

1. **The replacement structured oracle does not prove the blocking dependency reaches model context.** `TaskCompletion` is copied into the emitted `SteeringInjectedData`, but `steeringMessageToLLM` at `agent/session_queue.go:1058-1066` converts only `entry.Text`; the model never sees the typed field. The cleaned tests inspect that parallel event/controller data and no longer inspect the actual reminder/tool result. A review overlay changed `taskReminderAllDoneWhileDelegatesRun` to return the original unconditional generic completion reminder for every case; all four focused `TestTaskListTool_Update*` tests still passed under `-race`. Thus the exact #303 regression can return while tests stay green. Keep prose opaque, but put a parseable machine payload such as `blocking_delegate_ids`/completion state on the actual queued model message and parse that production message in the test (machine fields/IDs are explicitly exempt from the prose ban), or otherwise test the real model-consumed output rather than an event-only parallel copy.

2. **The required gate panics because `steerTaskCompletion` is unconditionally called on partially constructed tool dependencies.** `agent/session_tools_task.go:283` dereferences the new callback without a nil guard, while the existing `taskToolHarness` builds `toolDeps` directly and leaves it unset. Both required `build-and-test` and a fresh local `TestTaskTool_UpdateMarkerOnlyDescribesExplicitFinalInProgress` run panic at that line. Make the dependency mandatory in every constructor/harness or provide the appropriate fallback for legacy direct constructions, then run the full agent race gate.

**Verification performed**
- Read #303, PR body/history/checks/review state, complete diff, updated controller/waiter lifecycle, steering persistence/event/model-conversion paths, and full prompt-policy cleanup delta.
- Focused task/delegate lifecycle tests under `-race -count=3` — pass on the submitted head.
- Review overlay restoring the generic completion reminder while retaining parallel structured metadata — all focused lifecycle tests still pass, confirming the missing model-path oracle.
- `go test -race ./agent -run '^TestTaskTool_UpdateMarkerOnlyDescribesExplicitFinalInProgress$' -count=1` — panic on nil `steerTaskCompletion`, matching CI.
- `go test ./agent/events`, `golangci-lint run ./agent/...`, affected `go vet`, `make lint-gofmt`, and `git diff --check b21db7bd78bb4cc53f7fa9f61dd0648ac62fd589...HEAD` — pass.
- Required `build-and-test` — fail on the same nil callback panic; `web` and `Build snapshot artifacts` pass.
- Approved prompt-prose policy checked: removing reminder phrase searches is correct, but structured metadata must travel through and be tested on the actual model-consumed path; machine completion fields and delegate IDs are valid oracles.

obra:
## Sol Important findings resolved

Pushed commit `b1b0b3099c9de87c06ca2cadf04702d1eeb97fe8` non-force to the existing PR #409 branch. No new PR was opened.

### Fixes

- The tasks-done queued steering text now carries a stable model-visible machine block, `<evener-task-completion>{...}</evener-task-completion>`, with required `completion_state` and `blocking_delegate_ids` fields. The same state remains available as typed event metadata. Tests take the real queued `steeringMessage` through production `steeringMessageToLLM`, parse the machine JSON, and assert opaque live delegate IDs and ready/waiting state across blocking, background, completed, failed, and restart persistence paths. They do not assert natural-language reminder prose.
- Direct/legacy `toolDeps` constructions with no `steerTaskCompletion` callback now fall back to the existing generic `steer(msg, tasks-done)` route. Normal sessions retain the typed callback and existing lifecycle/wake routing.

### RED evidence

- Before production changes, the focused real-path lifecycle tests failed with `model-consumed tasks-done message has no unique machine payload`.
- The existing `TestTaskTool_UpdateMarkerOnlyDescribesExplicitFinalInProgress/done` case panicked at `agent/session_tools_task.go:283` on the nil `steerTaskCompletion` callback.
- A review mutation restoring the unconditional generic completion function made `TestTaskListTool_UpdateNamesBlockingDelegateDependency` fail on the missing model machine payload, proving the cleaned oracle no longer passes that regression.

### GREEN evidence

Passed locally:

- focused completion/delegate/persistence suite under `-race -count=10`
- full `go test -race ./agent -count=1` (342.131s, constrained to reduce shared-host contention)
- `go test ./agent/events` and events `-race -count=3`
- `make vet`, `make lint-gofmt`, `make lint-evenerfuzz`, `make lint-eval`, and `git diff --check`
- full `make lint`: PASS, 8 modules (133s)
- `make build` and `make build-go`
- required `make test-race`: all 7 Go modules pass
- required `ROOT_FULL=1 WEB=0 make test`: all 7 Go modules pass
- `make test-dev-tooling`, `make fuzz`, `make fuzz-gap-check`, and required `make fuzz-corpus-scan`
- `make test-web`, `make build-web`, and `make test-web-browser`

### Limitations / interrupted attempts

The shared host terminated some initial broad invocations with SIGTERM/exit -1. One first workspace race wave also exposed an unchanged root test's missing concurrent npm log entries; that exact test passed under `-race -count=10`, and the complete constrained workspace race retry passed. An initial lint run found the new exported-const comment issue; it was fixed, and the isolated full lint retry passed. Initial deterministic/web invocations were also SIGTERM-interrupted; their complete retries passed as listed above. No interrupted run is counted as green.

obra (commented):
## Sol review

**Verdict:** BLOCKED

**Reviewed head:** `b1b0b3099c9de87c06ca2cadf04702d1eeb97fe8`

No findings.

Both prior Important findings are resolved. Task completion steering now appends a delimited JSON block with required `completion_state` and `blocking_delegate_ids` fields to the actual queued text consumed by `steeringMessageToLLM`, while retaining typed event metadata. The real lifecycle tests parse that production message for ready, blocking, background, terminal, and restart-persistence cases. Legacy/direct `toolDeps` constructions now use the generic tasks-done steering callback when the typed callback is absent, eliminating the nil dereference without changing normal-session routing.

### Verification

- Reviewed the complete incremental delta from prior reviewed head `b427bb9e765b5278acd29ac8f035291b41964f22` through this head, including task state derivation, controller dependency IDs, steering queue conversion/cloning/persistence, normal and legacy callbacks, and event payloads.
- Focused task completion, blocking/background/terminal delegate, persistence, marker, and legacy fallback tests under `-race -count=10` — passed.
- `./agent/events` under `-race -count=3` — passed.
- Review-only generic-reminder mutation removed the machine block while retaining task completion mechanics; the blocking-delegate test failed with `model-consumed tasks-done message has no unique machine payload`.
- Review-only mutation restoring the unconditional typed callback reproduced the legacy harness nil-pointer panic, so the fallback regression is pinned.
- `go vet ./agent/... ./agent/events`, focused `golangci-lint run ./agent/...`, `make lint-gofmt lint-generated`, and cumulative `git diff --check` — passed.
- Prompt-prose policy checked: assertions parse machine fields/enums and opaque delegate IDs from typed/model messages; no natural-language reminder prose is pinned.
- `web` and `Build snapshot artifacts` are green. Required `build-and-test` was still in progress at submission, so this review is BLOCKED rather than PASS. Per campaign contract, it was not polled after submission.

Issue #370, opened 2026-08-23
Agents are resource-blind on capped machines: no cpus/memory in the environment block, no pricing of concurrent decode work — three container OOM kills in one run


obra:
retest7, video-processing, all three trials: serf was SIGKILLed (exit 137) by the kernel OOM killer under the task's 2048MB container cap, mid-tool-call — no panic, no shutdown path involved (verified: zero panic/goroutine output; verifier ran fine in the same container afterward). The work pattern: the root and its delegates each independently decoded the same full video into memory, up to 4 sessions doing so within one 60s window, alongside serf's own multi-session image-laden heap (peak concurrent in-flight request bytes ~2x the previous stack's). Measured in the real image: one decode script peaks at 300-470MB; three concurrently fit in the cap; the balance was the agent process itself, which the kernel then chose as the largest victim.

Interview (on a copy): "I did not check memory, CPU count, or estimate the decoded-list cost before running... I launched the delegate without treating it as sharing the same machine, CPU, and memory budget... A single streaming pass retaining only per-frame scalars would have provided the same signal."

Two fix surfaces:
1. Environment: agent/prompts/sections/environment.md.tmpl renders cwd/platform/OS/date/model/sandbox but no CPU count or memory budget. On capped containers those are facts about the machine as it stands — surface them the way OS version is surfaced.
2. Doctrine candidate (general): work you spawn shares the machine you are on; price memory/CPU of data-heavy operations before running them concurrently, and prefer streaming over materializing when only aggregates are needed.

(Also noted for the harness side, not this repo: a 300s EC2 artifact-download timeout swallowed the third trial's verifier run entirely — the trial died the same way but recorded a download error instead of a reward.)

obra:
Second task family OOM-killed by its own concurrent fanout (retest10, winning-avg-corewars j4C39VK): 85 shell jobs launched over the session, 7 concurrent python3-driven pmars search loops alive at death on a 1-CPU/2048MB box, three of them started in the final 32 seconds; kernel SIGKILL ~5.6s after the last launch. Verified not a crash (zero Go runtime output; bash WIFSIGNALED 'Killed'; signature reproduced across three binaries). Also of note: 163k-token live context and 476MB api.jsonl by turn 451 — the agent-side heap is a co-tenant the model cannot see. Strengthens both fix surfaces: surface cpus/memory_mb in the environment block, and the price-your-concurrent-work doctrine now has a second, non-video field case.

PR #402, opened 2026-08-24
fix(agent): surface effective resource caps in prompts

obra:
## Summary

Fixes #370.

Surface finite effective CPU and memory caps in the agent environment block and add the requested shared-resource doctrine for concurrent work. The implementation is independent of #301's unmerged work: main had no cap fields or cgroup probe, so this PR adds the collection and prompt path directly.

## Root cause

`EnvironmentInfo` and `promptData` carried no effective cgroup resource values, and `agent/prompts/sections/environment.md.tmpl` therefore could not show the CPU or memory budget enforced by a capped container. The delegation prompt also did not tell the agent to price concurrent work against that shared budget or account for its own context/transcript heap.

## Trace evidence

Issue #370 records a 1-CPU/2048MB container launching 85 shell jobs, with seven concurrent data-heavy loops alive at kernel SIGKILL; the issue also records the agent's live context and `api.jsonl` transcript heap as a co-tenant. Paired issue #301 records host-level `nproc`/`MemTotal` values diverging from enforced cgroup limits and identifies `cpu.max`/`memory.max` as the reliable source.

## Implementation

- Probe Linux cgroup v2 `cpu.max` and `memory.max`, with cgroup v1 fallbacks, through the real execution-environment command boundary.
- Carry finite values through `EnvironmentInfo` (`cpus`, `memory_mb`) and `promptData`.
- Render `CPUs` and `Memory` lines only when a finite cap exists.
- Add one sentence to the surrounding delegation voice: before concurrent data-heavy work, price it against CPU and memory caps, treating the agent's context and transcript heap as an invisible co-tenant.
- Keep test-only launch-probe skips from creating unsandboxed scratch as a side effect.

## RED evidence

`go test ./agent -run 'Test(EnvInfoFromEnvCollectsEffectiveResourceCaps|SessionPromptIncludesEffectiveResourceCaps|DelegationPromptIncludesResourceCoTenantDoctrine)' -count=1` failed before the production change on missing `cpus`, `memory_mb`, assembled prompt lines, and doctrine output.

## Verification

Passed:

- `go test ./agent -run 'Test(EnvInfoFromEnvCollectsEffectiveResourceCaps|EnvInfoFromEnvOmitsUnlimitedResourceCaps|SessionPromptIncludesEffectiveResourceCaps|DelegationPromptIncludesResourceCoTenantDoctrine)' -count=1`
- `go test ./agent/schema -count=1`
[...]

PR #419, opened 2026-08-24
fix(agent): report effective container resources

### LINKED ISSUE #301: nproc and /proc/meminfo report host resources inside a limited container, so agents oversubscribe
Inside a CPU-limited container, `nproc` and `/proc/meminfo` report the *host's* resources, not the cgroup limit. The agent reads them, sizes its parallelism accordingly, and gets throttled.

## Evidence

Measured directly against a live Terminal-Bench trial container:

```
NanoCpus=1000000000        CpuQuota=0  CpuPeriod=0
Memory=2147483648          MemorySwap=4294967296

# inside the container:
nproc            -> 16
MemTotal         -> 63556116 kB   (~63 GB)

/sys/fs/cgroup/cpu.max     -> 100000 100000   (= 1.0 CPU, enforced)
/sys/fs/cgroup/memory.max  -> 2147483648      (= 2 GiB, enforced)
```

The limits are real and enforced. `nproc` and `/proc/meminfo` simply are not namespaced, which is ordinary Docker behaviour without lxcfs. The agent sees 16 CPUs and 63 GB where it has 1 CPU and 2 GiB.

## What it costs

`train-fasttext__4EYEHNE` set `thread=16` on that basis. It was not getting 16-way parallelism; it was oversubscribing a single CPU 16:1.

There is a prior documented instance of the same shape in this suite's RCA archive: an agent oversubscribed a 1-CPU container with 8 Tesseract processes, and the affected stage went from 3 min 49 s (sequential) to unfinished after 22 min 22 s — roughly a 6x slowdown caused entirely by the agent's own parallelism.

Every CPU-heavy task is exposed to this, and the effect is silent: nothing errors, the work just takes several times longer, which then shows up as a timeout attributed to the model.

## Suggested fix

Cheapest version is prompt-level: instruct the agent to read `/sys/fs/cgroup/cpu.max` and `/sys/fs/cgroup/memory.max` rather than `nproc`/`MemTotal` when sizing parallelism or memory use, and to treat the latter as unreliable inside a container.

Better version is to resolve the effective limits during environment probing and state them in the environment context, so the correct numbers are in front of the model before it plans anything. That removes the need for the agent to remember a rule at the moment it matters.

obra:
## Summary

Fixes #301.

The environment snapshot now records structured effective resource caps for Linux processes:

- cgroup v2: `cpu.max`, `cpuset.cpus.effective`/legacy `cpuset.cpus`, and `memory.max`
- cgroup v1: CPU CFS quota/period, cpuset, and `memory.limit_in_bytes`
- mount and membership paths are resolved from the process-visible `/proc/self/cgroup` and `/proc/self/mountinfo`
- host `runtime.NumCPU`/`/proc/meminfo` values are used only for a readable no-cgroup hierarchy or an explicitly unlimited controller
- missing or malformed cgroup metadata remains unknown instead of being reported as container capacity

The structured data is persisted as `EnvironmentInfo.resources` for the separate #370 environment presentation work. Prompt templates and prompt wording are intentionally unchanged. Refs #370.

## Root cause and real-path RED evidence

`envInfoFromEnv` had no CPU or memory resource data. The Linux environment probes relied on host-only `runtime.NumCPU()` and `/proc/meminfo`; under the live #301 evidence, `cpu.max` was `100000 100000` (1 CPU) and `memory.max` was `2147483648` (2 GiB), while `nproc` reported 16 and `MemTotal` reported about 63 GiB.

The regression tests drive `envInfoFromEnv` through the production snapshot path with fixture-backed process files. Before the implementation:

```text
go test ./agent -run 'TestEnvironmentInfo(ReportsEffectiveCgroup|UsesHostFallback|UsesCgroupV2CPUSet|DoesNotTreat)' -count=1 -v
FAIL: resources = <nil>, want structured resource object
```

## Verification

Passed:

- `go test -p 1 ./agent -run '^TestEnvironmentInfo' -count=1`
- `go test -race -p 1 ./agent -run '^TestEnvironmentInfo' -count=1`
- `(cd agent && golangci-lint run ./...)` — 0 issues
- `(cd agent && go vet ./...)`
- `make lint-gofmt`
- `make build-go`
- `GOOS=linux go test -c -o ... ./agent` — Linux ARM64 test binary compiled

The full root agent suite had previously passed at `go test ./agent -count=1` (514.380s). Fresh full-suite and `make vet` attempts in this high-load host environment were terminated with SIGTERM/exit -1 and produced no Go diagnostics; the focused production-path tests and agent vet/lint gates pass.

## Risks and limitations

- This change exposes structured data but does not alter prompt prose; #370 owns presentation.
- If cgroup metadata or a controller file is unavailable or malformed, that resource is omitted rather than guessed from host data.
- Memory is reported as whole MiB rounded down; fractional CPU quotas remain fractional.


obra (commented):
## Sol review

**Head reviewed:** `e2bbb7cf09c17064d66a224e51bedb8b911229a9`  
**Verdict:** **CHANGES REQUESTED**  
**Findings:** 4 Important

1. **Important — CPU and memory limits on ancestor cgroups are ignored.** `effectiveCPU` and `effectiveMemory` read only the process membership directory (`agent/resource_limits.go:133-136` and `:165-171`). Cgroup CPU bandwidth and memory maxima are hierarchical: a leaf may legitimately contain `cpu.max = max` and `memory.max = max` while its parent enforces 1 CPU / 2048 MiB. In that fixture this implementation reported this review host's 18 CPUs / 64000 MiB instead. Walk from the member leaf through the mount root and combine the most restrictive quota/memory value at every level (the v2 effective cpuset file is already ancestor-aware).

2. **Important — unreadable or malformed membership metadata can become root/host capacity.** `cgroupErr` is consulted only when no mounts were parsed. When `/proc/self/mountinfo` exposes a cgroup mount but `/proc/self/cgroup` is unreadable, `membershipPath` silently defaults to `/`; an unlimited hierarchy root then causes host CPU/memory to be emitted. `parseCgroupMemberships` likewise discards malformed records and makes them indistinguishable from a genuine no-cgroup host. Both violate the PR's stated rule that missing/malformed cgroup metadata remains unknown and can conceal a limited nested membership. Preserve parse/read validity and return unknown unless the process membership is positively resolved; do not use `/` as the error fallback.

3. **Important — a hybrid v1/v2 host loses the controllers that actually own the limits.** The presence of any v2 mount makes `effectiveCPU` ignore every v1 CPU/cpuset mount (`agent/resource_limits.go:124-159`), while `effectiveMemory` returns from the first v2 mount before considering a v1 memory controller (`:161-175`). A valid hybrid fixture with an unrelated v2 hierarchy and CPU/memory attached to v1 produced no caps instead of 1.5 CPUs / 2048 MiB. Choose the hierarchy per controller from its membership/mount ownership rather than treating any v2 mount as authoritative for all controllers or relying on mount order.

4. **Important — co-mounted v1 CPU and cpuset controllers do not combine.** The `switch` at `agent/resource_limits.go:149-156` handles `cpu` before `cpuset`, so a valid v1 mount carrying both reads only CFS quota. A 4-CPU quota plus a 2-CPU cpuset was reported as 4 CPUs. Read both controller files independently when both flags are present, then take the restrictive result.

**Verification**

- Passed: `go test ./agent -run '^TestEnvironmentInfo' -count=1`.
- Passed: `go test -race ./agent -run '^TestEnvironmentInfo' -count=1`.
- Passed: `GOOS=linux GOARCH=amd64 go test -c ./agent`.
- Current GitHub checks observed before submission: `build-and-test`, `web`, and `Build snapshot artifacts` succeeded; publication jobs were skipped as expected.
- A scratch `go -overlay` regression suite (no repository edits) exercised the production fixture seam and failed across all four findings (five cases): ancestor caps returned 18/64000 instead of 1/2048; hybrid caps were absent instead of 1.5/2048; unreadable and malformed membership returned host facts; and co-mounted v1 controllers returned 4 CPUs instead of 2.
- The added tests assert structured JSON machine fields, so I found no prompt-prose-oracle violation under `docs/developing-evener/testing.md`. I treated prompt presentation as the responsibility of the explicitly stacked PR #402 and did not count its absence here as a defect.

**Limitation:** this reviewer is on Darwin, so I cross-compiled Linux but could not execute against a live Linux cgroup hierarchy; the failing cases use the PR's own `ReadHostFile` production-path fixture boundary.

obra:
## RED/GREEN resolution — four Important findings

**Resolved head:** `515e36452659b5fa1862e399d27e8604fc93f18f` (non-force child of reviewed head `e2bbb7cf09c17064d66a224e51bedb8b911229a9`). No new PR was opened.

### RED

Added production-path `ReadHostFile` fixture regressions before changing the implementation. The focused RED command failed across the reviewed boundaries:

```text
go test ./agent -run 'TestEnvironmentInfo(CombinesCgroupV2AncestorLimits|BoundsCgroupV2AncestorWalkAtMountRoot|RequiresValidResolvedCgroupMembership|RecognizesPositivelyResolvedRootMembership|SelectsHybridHierarchyByControllerMembership|CombinesCoMountedCgroupV1CPUControllers)$' -count=1 -v
```

Observed failures matched the review: ancestor-limited v2 fixtures reported host/leaf values (for example 18 CPUs / 64000 MiB instead of 1 / 2048); unreadable, malformed, and ambiguous membership reported host facts; both hybrid mount orders omitted v1-owned caps; and co-mounted v1 CPU+cpuset reported 4 CPUs instead of 2.

### GREEN / finding resolution

1. **v2 ancestors:** resolve the positive v2 membership, map it beneath the process-visible mount, walk leaf-to-mount-root only, and combine the most restrictive `cpu.max` and `memory.max`. The actual unified root remains the documented unlimited root; a non-root delegated mount root is included. `cpuset.cpus.effective` remains leaf-read/ancestor-effective. Tests cover finite/unlimited combinations, namespace-relative and full membership paths, and restrictive files above the mount root that must not affect the result.
2. **membership validity:** `/proc/self/cgroup` read failure, any malformed record, duplicate hierarchy, or duplicate controller ownership now makes resources unknown even when mountinfo is present. Root fallback is allowed only for a positively parsed root membership; unresolved membership never defaults to `/`.
3. **hybrid ownership:** CPU quota, cpuset, and memory select v1 or v2 independently from controller membership, not from mount presence/order. Tests cover unrelated v2 plus v1-owned CPU/cpuset/memory in both mount orders and a split v2 CPU/memory + v1 cpuset hierarchy.
4. **co-mounted v1:** CPU quota and cpuset are read independently from the same mount and combined restrictively. A positively owned but unreadable v1 cpuset remains unknown rather than being treated as absent.

Structured `EnvironmentInfo.Resources` / `cpus` / `memory_mb` fields are unchanged. Applying this delta atop PR #402 reviewed head `d2c6e26e5fe51f09a89b87b7d78d84bcb4a2dd2b` passed the trusted-resource plus rendered-environment tests under `-race`.

### Verification

Passed locally:

- focused `^TestEnvironmentInfo` production-fixture suite under `-race -count=10`
- full `go test ./agent` and full `go test -race ./agent`
- `make test-race` (all workspace modules) and `ROOT_FULL=1 WEB=0 make test`
- Linux amd64 + arm64 agent test cross-builds; the final arm64 binary ran the complete fixture suite inside a capped Linux Docker container
- `make build`, `make build-go`, `make vet`, required `make lint` (all eight modules, tagged evenerfuzz/eval floors, required gitleaks), `make lint-gofmt`, and `git diff --check`
- `make test-dev-tooling`, `make fuzz`, `make fuzz-gap-check`, and required fuzz-corpus secret scan
- `make test-web`, `make build-web`, and all four `make test-web-browser` guards

Two aggregate attempts were externally SIGTERM-interrupted by the local shell/provider with no Go/test diagnostic (one deterministic wave and one long fuzz replay; the first combined web attempt was interrupted similarly). Each affected canonical target/component was rerun to a real zero exit. Linux verification executes the production fixture seam on Linux; it does not claim a privileged live-host cgroup hierarchy mutation.

CI was not polled after this push.

obra (commented):
## Sol re-review

**Head reviewed:** `515e36452659b5fa1862e399d27e8604fc93f18f`  
**Verdict:** **CHANGES REQUESTED**  
**Findings:** 1 Important

I reread #301/#370, the prior 4-Important review, and the fixer response; reviewed the complete 1,277-line base→head diff; and traced the unchanged snapshot/persistence path plus PR #402's structured renderer at `d2c6e26`.

### Findings by severity

#### Important — a finite cgroup-v2 namespace root is discarded as if it were the unlimited global root

`readV2CPUHierarchy` and `readV2MemoryHierarchy` unconditionally remove the process-visible mount root whenever `mount.root == "/"` (`agent/resource_limits.go:315-317` and `:340-342`). A private cgroup namespace rewrites its delegated root to `/`; `mount.root == "/"` therefore does **not** prove this is the unlimited global hierarchy root. Finite controller files at that process-visible root remain authoritative.

I reproduced the ordinary Docker/private-cgroup-namespace shape in a `--cpus=1.25 --memory=384m` Linux container:

```text
/proc/self/cgroup:    0::/
mountinfo cgroup2:    0:32 / /sys/fs/cgroup ... - cgroup2 cgroup rw,nsdelegate
cpu.max:              125000 100000
memory.max:           402653184
```

The exact PR binary reported `cpus=18` and `memory_mb=16038` there instead of `1.25` / `384`. An independent `ReadHostFile` production-path fixture with those same proc/cgroup records reported `18` / `64000` instead of `1.25` / `384`. The new positive-root test only supplies `max` controller values, so it covers host fallback without testing the finite namespace-root case that triggers #301.

The process-visible mount root must participate when its controller file is present: finite values must be folded and explicit `max` may select host capacity. This is not just an unconditional deletion of the root special case: a scratch experiment showed that doing so makes the existing fixtures' intentionally absent global-root files unknown. Preserve absent actual-root handling while ensuring read errors/malformed data remain unknown, not host fallback.

### Prior four findings

1. **v2 ancestor folding — not fully resolved.** The finite/unlimited leaf/ancestor matrix and non-root delegated-mount bounds pass, but the `/` mount-root exception above drops a finite limit at the leaf/mount root.
2. **membership validity — resolved for the reviewed malformed/ambiguous cases.** Unreadable, malformed, partially malformed, duplicate-v2, and duplicate-controller memberships remain structurally unknown; a positively parsed unlimited root works. The finite positive-root defect is the Important finding above.
3. **hybrid v1/v2 ownership — resolved.** CPU quota, cpuset, and memory follow controller membership independently in both mount orders, including split v1 CPU plus v2 cpuset/memory and v2 CPU/memory plus v1 cpuset.
4. **co-mounted v1 CPU + cpuset — resolved.** Both controllers are read and combined restrictively; quota-lower, cpuset-lower, unlimited-quota, and unreadable-owned-cpuset cases behaved conservatively.

### Additional audit

- Effective v2 cpuset wins over legacy `cpuset.cpus`; legacy fallback, malformed input, and count-overflow paths remain conservative.
- Readable no-cgroup and explicit-`max` host fallbacks pass; unreadable controller/membership and numeric parse overflow do not become host facts. The namespace-root misclassification is the exception above.
- `ResourceCaps` collection, JSON persistence, and PR #402's structured XML/JSON renderer remain schema-compatible. Applying `e2bbb7c..515e364` to PR #402 head `d2c6e26` passed its structured resource/render tests under `-race -count=10`.
- The changed tests and my independent regressions assert structured fields, not prompt words, clauses, order, or snapshots; I found no `testing.md` prompt-prose-oracle violation.

### Commands and checks

Passed:

- Existing decisive production fixtures for all four prior findings plus cpuset/host fallback: focused `TestEnvironmentInfo(...)` matrix, `-count=1 -v`.
- Independent production-boundary matrix covering ancestor finite/unlimited combinations, exact mount-root bounds, membership ambiguity, both hybrid mount orders, co-mounted controllers, cpuset precedence, and numeric overflow.
- Both matrices under race/stress: independent `-race -count=20`; all `^TestEnvironmentInfo` tests `-race -count=20`.
- Full agent package: `go test ./agent -count=1` (`119.945s`) and `go test -race -p 1 ./agent -count=1` (`211.652s`); current and compatibility-tree `./agent/schema` suites.
- Linux amd64 and arm64 agent-test cross-builds; Linux amd64 `evenerfuzz` and `eval` tagged cross-builds and tagged `go vet`.
- All `^TestEnvironmentInfo` fixtures and the independent matrix executed in a capped Linux/arm64 Docker container.
- PR #402 compatibility tree: `go test -race ./agent -run '^(TestEnvironmentInfo|TestRenderedEnvironment)' -count=10`.
- `make lint` (`PASS lint (8 modules, 92s)`), `make lint-gofmt`, and `git diff --check`.
- GitHub checks refreshed at this exact head: `build-and-test`, `web`, and `Build snapshot artifacts` passed; publication/refresh jobs skipped as expected.

Decisive failures:

- Independent namespace-root fixture: got `18 / 64000`, want `1.25 / 384`.
- Live capped Linux container: got `18 / 16038`, want `1.25 / 384` from the directly observed finite controller files above.

**Limitations:** native review host is Darwin/arm64. Linux was both cross-built and executed through Docker's real cgroup-v2 namespace, but I did not mutate a privileged host hierarchy. One aggregate Linux cross-build command and the first full-race attempt were externally interrupted with exit `-1` and no diagnostic; each exact component was rerun alone/serially to exit 0 before submission.

obra:
Resolved the remaining Important cgroup-v2 namespace-root finding in `31ffa0969`.

- The process-visible mount root now participates whenever `cpu.max` / `memory.max` exists, including private namespaces whose delegated root is rewritten to `/`.
- Finite root values fold with descendants; explicit `max` selects host capacity.
- An absent global-root file is ignored only after a descendant controller was resolved, preserving existing fixtures. Malformed and non-ENOENT read failures remain unknown, and ancestor walking remains bounded at the mount point.
- Added production `ReadHostFile` coverage for finite/max namespace roots, absent root files, malformed/unreadable roots, and ancestor folding across CPU and memory.

Validation passed: focused resource `-race -count=20`; full `./agent` and full serialized `./agent -race`; Linux amd64/arm64 cross-builds plus `evenerfuzz`/`eval` tagged compile/vet; `make lint`, `make build-go`, `make vet`, gofmt, and `git diff --check`. A live Docker run with `--cpus=1.25 --memory=384m` observed `0::/`, root `cpu.max=125000 100000`, and root `memory.max=402653184`, and the production probe returned `1.25 / 384`. PR #402 compatibility resource/renderer tests passed under `-race -count=10`.

obra (commented):
## Sol final re-review

**Head reviewed:** `31ffa0969fa255d7176fe1993201d38676d0798b`  
**Verdict:** **PASS**  
**Findings:** none

I reread the prior four-finding review, the namespace-root re-review, and the `31ffa0969` resolution; audited the complete base→head change and final delta; and independently exercised the production `ReadHostFile` boundary plus a live cgroup-v2 namespace.

### Namespace-root resolution

- The process-visible mount root now stays in the CPU and memory fold (`agent/resource_limits.go:310-359`). A root-only `0::/` fixture with `cpu.max = 125000 100000` and `memory.max = 402653184` returns `1.25 / 384`; an explicit `max` root returns `runtime.NumCPU()` and fixture `/proc/meminfo` capacity.
- Root-only absent files, malformed files, and non-ENOENT read errors remain unknown rather than becoming host facts. The intentional omitted-global-root case is narrower: ENOENT at `/` is ignored only when the walk already has descendant paths, preserving resolved descendant limits.
- Finite namespace-root limits combine restrictively with leaf/parent values. The path walk still stops exactly at the process-visible mount point (`agent/resource_limits.go:719-740`); restrictive files above it did not affect the result.

### Prior findings and compatibility

- v2 CPU and memory ancestors fold across finite/unlimited combinations.
- Unreadable, malformed, partially malformed, duplicate-v2, and duplicate-controller memberships stay unknown.
- Hybrid CPU quota, cpuset, and memory ownership follows each controller's membership independently in both mount orders and both split directions.
- Co-mounted v1 CPU+cpuset combines restrictively for quota-lower, cpuset-lower, unlimited-quota, and unreadable-owned-cpuset cases.
- v2 effective cpuset wins over legacy; ENOENT-only legacy fallback works; malformed and count-overflow input stays conservative. Numeric quota/memory overflow remains unknown.
- `EnvironmentInfo.Resources` and `ResourceCaps` wiring is unchanged after the first commit and byte-identical to PR #402's consumed schema/integration contract. The two follow-up commits applied cleanly to PR #402 head `d2c6e26e5fe51f09a89b87b7d78d84bcb4a2dd2b`; its resource/render tests passed under race stress.

### Verification

Passed at this exact head:

- Existing `^TestEnvironmentInfo` fixtures plus an independent production-boundary matrix, then both under `-race -count=20`.
- The same existing+independent matrix as a Linux/arm64 test binary in a capped container, `-count=20`.
- Live Docker `--cpus=1.25 --memory=384m`: observed `0::/`, cgroup2 mount root `/`, `cpu.max=125000 100000`, `memory.max=402653184`, host `runtime.NumCPU()=18`; the production probe returned `1.25 / 384`.
- Full `go test ./agent -count=1` and full serialized `go test -race -p 1 ./agent -count=1`; current and stacked PR #402 schema suites.
- Linux amd64/arm64 agent test cross-builds; Linux amd64 `evenerfuzz` and `eval` tagged test builds and tagged `go vet`.
- PR #402 stacked resource/render matrix under `-race -count=10`.
- `make lint` (8 modules), `make lint-gofmt`, and base→head `git diff --check`.
- Exact-head GitHub Actions: `build-and-test`, `web`, and `Build snapshot artifacts` succeeded; publication/channel jobs skipped as expected.

**Limitations:** native review host is Darwin/arm64. Linux execution used Docker's real cgroup-v2 namespace on arm64, but I did not mutate a privileged host hierarchy. PR #402 compatibility was reconstructed locally from its exact head rather than run on a pushed stacked branch.

PR #434, opened 2026-08-25
fix(hub): fence stale relay epoch publications

obra:
Fixes the failed CI run [32794911542 / job 97644072327](https://github.com/prime-radiant-inc/evener/actions/runs/32794911542/job/97644072327), where `TestRelaySessionStaleEpochNotificationCannotPublish` observed `evener/thread/resync` after epoch replacement.

- Tag queued notifications with their authoritative epoch and publication fence.
- Linearize epoch revocation against listener publication; revoke fences before the exclusive boundary so blocked sends can exit without deadlock.
- Preserve replacement-epoch resync publication.
- Add a deterministic real-relay event-barrier regression with no sleeps, timeouts, or prose oracles.

Local verification:
- Focused stale-epoch race test (`-count=100`) — PASS
- Complete appsource package and appsource+hub race checks — PASS
- `make lint` — PASS
- `make lint-gofmt lint-generated vet` — PASS
- `make secret-scan`, `git diff --check` — PASS
- `make mutation-floor` with go-gremlins v0.6.0 — PASS
- `make test-web` — PASS
- `ROOT_FULL=1 WEB=0 make test` — PASS (all Go modules)
- `make build && make test-dev-tooling` — PASS

The first combined `make test` run reported two frontend tests under concurrent Go/Node load; both passed in isolation and the standalone frontend gate passed. The first `make merge-approval-gate` run was terminated by the local execution limit after lint/build; its phases above were rerun independently and passed.

PR #440, opened 2026-08-25
test(agent): lock rematerialize hook fixture assignment

obra:
## Summary

Fix the proven test-fixture data race that blocked PR #434. The test assigns
`jobManager.testOnlyAfterRematerializeDurableLoad` while the drain may already
be reading the hook in `rematerializeDurablePendings`.

The field contract in `agent/jobs.go` requires reads and writes under `jm.mu`.
This patch locks the child job manager only for the fixture assignment and
unlocks immediately. It does not change production code, hook lifecycle,
barriers, timeouts, comments, helper structure, or test semantics.

## Evidence

The failure is in CI run [32814484456](https://github.com/prime-radiant-inc/evener/actions/runs/32814484456), job [97699953835](https://github.com/prime-radiant-inc/evener/actions/runs/32814484456/job/97699953835): the race read is in `rematerializeDurablePendings`, and the previous write is `TestSubagentDrainReturnHandoffNoPhantomTurnInFinalizationWindow` at `agent/subagent_owned_job_drain_test.go:377`.

The related locked assignment in `TestSubagentFinalizationRematerializeLoadSnapshotStraddle` demonstrates the intended discipline.

## Validation

- `go test -race ./agent -run '^TestSubagentDrainReturnHandoffNoPhantomTurnInFinalizationWindow$' -count=100`
- `go test -race ./agent -run '^(TestRematerializeOwnedDrainJobPendings|TestSubagentFinalizationDrainReenqueueRaceDeterministic|TestSubagentFinalizationRematerializeLoadSnapshotStraddle)$' -count=20`
- `go test -race ./agent`
- `go vet ./agent`
- `gofmt -l` and `git diff --check`

### DIFF
--- a/agent/subagent_owned_job_drain_test.go
+++ b/agent/subagent_owned_job_drain_test.go
@@ -374,6 +374,7 @@ func TestSubagentDrainReturnHandoffNoPhantomTurnInFinalizationWindow(t *testing.
 			<-releaseWindow
 		})
 	}
+	fixture.child.sess.jobManager.mu.Lock()
 	fixture.child.sess.jobManager.testOnlyAfterRematerializeDurableLoad = func() {
 		select {
 		case <-windowHeld:
@@ -381,6 +382,7 @@ func TestSubagentDrainReturnHandoffNoPhantomTurnInFinalizationWindow(t *testing.
 		default:
 		}
 	}
+	fixture.child.sess.jobManager.mu.Unlock()
 	go func() {
 		select {
 		case <-recheckSawWindow:


obra:
## Sol review

**PASS** at `33baf084704e462a1b1208f0c7fc8d22eabe2d92` (base/current main `1595e30e8a90406a171a5b1c4bd9d0eac4044f6c`).

Verified the one-file `+2/-0` diff only holds `jm.mu` across the `testOnlyAfterRematerializeDurableLoad` function-value assignment. This matches the field contract and sibling pattern; production copies the hook under the same mutex and invokes it after unlocking, so the race is removed without callback deadlock or changes to barriers/finalization semantics. The #434 trace is exactly the formerly unlocked test write versus that production read. The companion hook is ordered before its production read by the shell-release channel, so no companion or production change is needed.

Checks passed: focused `-race -count=100`; five related rematerialize/finalization tests under `-race -count=20`; full `go test -race ./agent`; `go vet ./agent`; `make lint`; `git diff --check`; and exact-head CI, including all-workspace vet/lint/race. Review threads/adverse reviews: 0. No in-scope issues.

PR #422, opened 2026-08-24
test: improve statement coverage across all Go modules and frontend

obra:
## Coverage Improvements — Toward 100%

Adds 97 test files (~30,800 lines) across all Go modules and the frontend to close coverage gaps.

### Before → After (test track)

| Module | Before | After |
|--------|--------|-------|
| root | 84.0% | ~85%+ |
| agent | 91.3% | ~92%+ |
| llm | 93.1% | ~94%+ |
| identifier | 98.9% | 99.5% |
| web (lines) | 95.9% | 97.7% |
| auth | 100% | 100% |
| envvars | 100% | 100% |
| fuzz | 100% | 100% |
| invariant | 100% | 100% |

### What was covered

- **Root module**: TUI hub panels (hub_update_config, hub_spawn, hub_commands, hub_dashboard, hub_notifications, hub_browse, etc.), TUI core panels (question_overlay, composer, notice, command_palette, etc.), TUI internal subpackages (launchconfig panels, clipboard, transcript reducer, msgrender), evener-hub (web API, tree, appsource, hubcore), and misc packages (appwire, plugins, appprojector, etc.)
- **Agent module**: job_watch, jobs, session_tools_jobs, delegate_runtime, execenv/local, session_tools_worktree, session_queue, subagents, session_attention, jobs_nested, plus events, schema, sandbox, provider, goal, hooks, jobstore, delegatestore, contextmgr, agenttest, tool
- **LLM module**: api_attempt, apilog, adapter_timeout, api_attempt_sanitize, client, misc — 7 unreachable statements documented
- **Identifier module**: gitBinaryMainRootLocal, DecodeUUID — 2 unreachable statements documented (ValidateProjectID base62 suffix loop, projectID empty-readable fallback)
- **Web module**: CommandPalette, SessionChrome, AdvancedOptions, activityData, reducer, streaming, ansi, jobTools, mutationDispatcher, tree, CredentialsSection, mcp, SessionMenu, modelCatalog, SelectionQuote, LaunchConfigForm, activityPanel, pendingTurnsStore, useTranscriptScroll

### Unreachable code documented

- **LLM**: 7 defensive statements
- **Identifier**: 2 defensive branches (ValidateProjectID base62 suffix loop, projectID empty-readable fallback)
- **Root**: ~25 live-test-only functions in tool-fluency (require provider credentials by design)

### Verification

`make merge-approval-gate` passes: lint, vet, race detector, deterministic tests, fuzz seeds, secret scan, web build, all selftest gates.

PR #442, opened 2026-08-25
test(hubstart): honor production immediate-exit window

obra:
## Summary

- Remove the two `withLocalHubImmediateExitWindow(t, 30*time.Second)` calls from `hub_start_covtest_test.go`.
- Keep both coverage tests on the production `LocalHubImmediateExitWindow` of 750ms.

## Context

This is a minimal post-merge correction for the second Sol audit finding in merged PR #422:
https://github.com/prime-radiant-inc/evener/pull/422#issuecomment-5406391527

The audit identified that the two newly merged tests widened the immediate-exit deadline by 40x while asserting immediate-exit behavior. The production code, helper seam, assertions, test names, imports, and all other files are unchanged.

Related: #422

## Verification

- Focused tests: repeated under `-race`, constrained `GOMAXPROCS=1`/`-p=1`/`-parallel=1`, and four busy CPU-load processes.
- Complete hubstart tests and race suite pass.
- `make lint-gofmt`, `make lint`, `go vet ./...`, `make vet`, `go test ./...`, and `make test` pass.

### DIFF
--- a/cmd/evener-tui/internal/hubstart/hub_start_covtest_test.go
+++ b/cmd/evener-tui/internal/hubstart/hub_start_covtest_test.go
@@ -79,7 +79,6 @@ func TestCovDialHubRPCWithFrameHandler(t *testing.T) {
 // TestCovStartLocalHubWithStateDir observes the child-process environment at
 // the process-launch boundary.
 func TestCovStartLocalHubWithStateDir(t *testing.T) {
-	withLocalHubImmediateExitWindow(t, 30*time.Second)
 	t.Setenv(stateDirHubHelperEnv, "1")
 	dir := t.TempDir()
 	stateDir := filepath.Join(dir, "state", "evener")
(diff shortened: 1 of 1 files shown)


obra:
## Sol review

**PASS** — exact head `f7398341f01f993c7da97cf7cf90f7d24fe7a6af` (base `1595e30e8a90406a171a5b1c4bd9d0eac4044f6c`).

- **Diff:** exactly one file, `cmd/evener-tui/internal/hubstart/hub_start_covtest_test.go`, 0 additions/2 deletions: only the two requested `withLocalHubImmediateExitWindow(t, 30*time.Second)` calls are removed. No production/helper/assertion/import/other timeout change.
- **Contract/mechanism:** both tests now use the unchanged production 750ms window. Each launches the real test executable; its env-selected `init` fixture writes the asserted output and exits (7 or 1), while `StartLocalHub` directly selects the `cmd.Wait` result against that 750ms window. A missed exit would return nil and fail the assertions. No sleeps, retries, polling, hidden override, or concurrent test mutation is present.
- **Verification:** the two focused tests passed `-race -count=100` under `GOMAXPROCS=1` and a 64-FD limit (200 invocations, 5.552s); the full hubstart race suite passed, as did 20 shuffled race suites, package vet, scoped golangci-lint (0 issues), gofmt, exact blob/diff checks, and exact-head CI (`build-and-test`, `web`, artifact build; release-only jobs skipped as expected). No flakes occurred.
- **Feedback:** pre-review query found 0 review threads, 0 submitted/adverse reviews, 0 inline comments, and 0 prior PR comments.

**Out of scope (not requesting expansion):** unchanged uses remain in `hub_start_test.go` (30s) and `hub_start_fuzz_test.go` (1s/1ms); their cleanup-based global overrides restore prior state and the shuffled race stress passed.

PR #446, opened 2026-08-25
test(server): measure bounded transcript header reads

obra:
## Root cause

The permanent race gate failure in run 32879537762, job 97905436755, was caused by `TestTranscriptHeaderReadsOnlyLeadingHeader` treating relative `testing.AllocsPerRun` results as an I/O oracle. Under CI contention, the same production path reported `large=29 small=18`; those heap counts vary with runtime/filesystem/buffer effects and do not establish that historical transcript entries were decoded.

`transcriptHeader` opens at offset zero, reads newline-framed records with `transcript.ReadLine`, skips leading blanks, decodes one header, and returns immediately. It does not scan historical entries. The production reader has a fixed 64 KiB prefetch boundary for ordinary headers.

## Change

- Keep the production path and malformed/truncated/oversized-line semantics unchanged while routing the parser through a reader helper.
- Replace the allocation comparison with a deterministic byte-bound assertion using a real opened transcript wrapped at the reader boundary.
- Assert the header identity and that validation reads no more than the production 64 KiB prefetch bound, even with 2000 historical entries.

A mutation that read the entire source before parsing made the focused test fail, proving the new oracle detects unbounded reads.

## Verification

- `go test -race ./server -run '^TestTranscriptHeaderReadsOnlyLeadingHeader$' -count=100`
- `go test -race ./server -count=1`
- `go test -race ./agent/transcript ./internal/apptranscript ./server -count=1`
- `make test-race` (all go.work modules; passed)
- `go vet ./server`
- `gofmt` and `git diff --check`

Fixes the exact-main CI failure from run 32879537762 / job 97905436755. No frontend or unrelated files changed.

### DIFF
--- a/server/appwire_runtime.go
+++ b/server/appwire_runtime.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"maps"
 	"os"
@@ -699,8 +700,11 @@ func transcriptHeader(path string, maxLineBytes int) transcript.Header {
 		return transcript.Header{}
 	}
 	defer file.Close() //nolint:errcheck // read-only file; close error is not actionable
+	return transcriptHeaderFromReader(file, maxLineBytes)
+}
 
-	reader := bufio.NewReaderSize(file, 64*1024)
+func transcriptHeaderFromReader(source io.Reader, maxLineBytes int) transcript.Header {
+	reader := bufio.NewReaderSize(source, transcriptHeaderReadBufferBytes)
 	for {
 		lineBytes, complete, _, err := transcript.ReadLine(reader, maxLineBytes)
 		if err != nil || !complete {
--- a/server/appwire_turns.go
+++ b/server/appwire_turns.go
@@ -18,6 +18,13 @@ import (
 // large turn.
 const appTranscriptMaxLineBytes = 128 << 20
 
+// transcriptHeaderReadBufferBytes bounds the prefetch for an ordinary header
+// line. The parser stops at that line; keeping this buffer finite prevents a
+// large historical transcript from crossing the reader boundary merely because
+// identity validation needs its header. Oversized headers are still governed
+// by the maxLineBytes limit passed to transcript.ReadLine.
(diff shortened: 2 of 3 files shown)


obra:
PASS — review-only audit of exact head `924f834f5e18dc4bc2fe430a78e29e5091a9eac8` against exact base `5094e7db1baf136f4d8fb17061be7131c39b4a41`; no findings.

- Confirmed baseline RED in run 32879537762 / job 97905436755 at base SHA: `TestTranscriptHeaderReadsOnlyLeadingHeader` failed on the allocation proxy (`large=29 small=18`).
- Audited the sole production caller and parser path. `transcriptHeader` still owns open/close and now delegates immediately to `transcriptHeaderFromReader`; the loop, first-nonblank behavior, `ReadLine` error/EOF/truncation handling, strict header decode, and identity decision are otherwise unchanged.
- The named 64 KiB value is the pre-existing reader size, not a line limit. `ReadLine` continues across `bufio.ErrBufferFull` up to `appTranscriptMaxLineBytes` (128 MiB), so valid headers larger than 64 KiB are not truncated. Malformed, blank, oversized, unterminated, and concurrently appended-file behavior is unchanged.
- The new oracle wraps a real `os.File` at the exact helper used by production; it measures bytes read at the production reader boundary rather than allocations, timing, decoded history, or a transformed fixture.
- Mutation RED (ephemeral Go overlay; reviewed sources untouched): an `io.ReadAll` regression parsed the same file successfully but failed the byte boundary at `488799 > 65536`. Restored GREEN: focused race x100 and full `./server -race` passed on both exact head and GitHub merge tree `63d9d01b2eccad63aff77ed51b79b67532287732`.
- Exact-head `make test-race` passed all go.work modules. `go vet ./server`, gofmt, and patch whitespace checks passed. (An initial source-tar run was invalid only for two git-metadata audit tests; rerunning from a clean `gh repo clone` at the exact SHA passed.)
- Exact-head CI is terminal: required `build-and-test` passed (including all-module vet/lint/race, deterministic tests, dev-tooling, fuzz replay/gap, and secret scan); `web` and snapshot build also passed. Reviews: 0. Review threads: 0 unresolved.

No approval or merge performed.

PR #454, opened 2026-08-25
fix(agent-shards): fail closed on explicit run files

obra:
## Summary

Focused follow-up to the PR #371 overbuilding audit. The shard run-file protocol remains necessary: the exact main test set is already over Linux's single-argument limit at the runner's 4-shard default, and CI's 8-shard pattern is too close to the limit to treat as robust.

The protocol now fails closed when `EVENER_SHARD_RUN_FILE` is explicitly supplied: read failures, empty/whitespace files, invalid regexes, and `flag.Set("test.run", ...)` failures report a clear error and exit 2. When the variable is unset, TestMain behavior is unchanged. The stale comment claiming a command-line fallback was removed.

## Measurements against exact main

Base: `8ac7b342a3cb3e8b4018efc7d5e1304c514e7965`

- Real agent test binary: 4,998 Test/Example names; 227,390-byte `-test.list` output.
- Matching cached survey: 4,998 cost entries.
- Cost-balanced maximum regex sizes: 2 shards 164,498 bytes; 4 shards 133,182 bytes; 8 shards 117,279 bytes.
- Linux `MAX_ARG_STRLEN`: 131,072 bytes; therefore 4 shards exceed the limit and 8 shards leave only ~10.5% margin.
- Darwin `kern.argmax` and local `getconf ARG_MAX`: 1,048,576 bytes; Linux's per-argument limit is the binding constraint.

## Tests

- Real fixture test-binary TDD: unset, missing, empty, invalid, and valid handoff cases.
- Agent-shards e2e: partition/selection, cleanup, failure evidence, survey, signal, and environment cases.
- Focused race handoff e2e.
- `make test-dev-tooling`, `make lint-gofmt`, `make vet`, `make lint`, and `WEB=0 make test-short` pass.
- A full `make test-short` run completed all Go modules but hit five unrelated pre-existing frontend Spawn/Rail failures (6,917/6,922 frontend tests passed); the required Go-only form passed.

References PR #371 and the 48-hour overbuilding audit.

### DIFF
--- a/agent/session_tools_worktree_create_test.go
+++ b/agent/session_tools_worktree_create_test.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"slices"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


obra (commented):
PASS

obra (commented):
PASS

PR #456, opened 2026-08-25
fix(agent): reject invalid task reasoning efforts

obra:
## Summary

- validate task-level reasoning effort against Evener's canonical vocabulary before mutating the task store
- reject invalid append and update batches atomically, even when stale model metadata advertises an unsupported level
- cover stale-schema admission, handler-stage rejection, mixed-batch rollback, normalization, and provider request behavior

## Root cause

A stale model effort list admitted `ultra` through the `task_list` schema. The task handler persisted it without canonical validation. Once that task became active, autonomous turns repeatedly sent an invalid provider request and received HTTP 400.

## Test plan

- `make lint`
- `make vet`
- `make test`

### DIFF
--- a/agent/session_task_effort_test.go
+++ b/agent/session_task_effort_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"os"
 	"path/filepath"
+	"slices"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


obra:
CHANGES

The handler now rejects new invalid append/update values atomically, but it does not close the persisted-state path from the incident.

`TaskStore.Load` still unmarshals persisted tasks verbatim (`agent/task/task_store.go:184-211`). On the next resumed round, `prepareModelRequestWithError` reads an in-progress task's effort and accepts the normalized-but-unvalidated value (`agent/session_model_call.go:155-169`). `buildModelRequest` then calls `ClampReasoningEffort`; unknown vocabulary intentionally passes through unchanged (`llm/types.go:694-706`). Thus a task file already poisoned with `reasoning_effort:"ultra"` before this fix still sends `ultra` on every provider request after restart, particularly when the same stale model metadata advertises it. Deploying this PR therefore does not recover the real incident's persisted session.

I reproduced this at exact head with an overlay-only integration test: create/save a real session, install the legacy in-progress task JSON, restore via `RestoreSessionFromMeta`, then call `ProcessInput` through the fake adapter. The restored adapter request had `ReasoningEffort == "ultra"`.

Please validate/sanitize restored task efforts against `llm.ValidateReasoningEffort` before they can become the effective request effort (without duplicating the vocabulary), and add a restore -> activation -> provider-request regression test. The behavior for an invalid persisted override should be explicit and must not poison the provider request.

obra:
PASS

Re-reviewed exact corrected head `97778cce8eac85954168b69b553e63bb9d5ce059` (parent `317522ed6746e811d20539259a34876a1def04c5`, original base `8ac7b342a3cb3e8b4018efc7d5e1304c514e7965`). No findings.

- Corrected delta: 3 files, +198/-0. Full PR: 4 files, +393/-3.
- Append/update validates normalized task effort with canonical `llm.ValidateReasoningEffort` before any store mutation; stale metadata may expose `ultra`, but mixed invalid batches are handler-rejected atomically and do not cross the adapter/provider boundary.
- `TaskStore.Load` normalizes a fully decoded local slice under the store lock: invalid persisted effort becomes empty/inherit; valid overrides canonicalize and remain effective; identity, status, and dependencies remain intact; malformed JSON does not replace existing state.
- Restore regression exercises real save + legacy JSON replacement + `RestoreSessionFromMeta` + persisted `in_progress` activation + `ProcessInput` to the fake adapter. Invalid `ultra` inherits the session effort; valid ` HIGH ` becomes `high` and wins.
- Exact-head archive verification: focused and full `go test -race ./agent/task ./agent -count=1` passed.
- Checks attached to this head are terminal green: `build-and-test`, `web`, and `Build snapshot artifacts`; 0 unresolved review threads. GitHub CI also tested the synthetic merge of this head into current main.
- Current-main overlap: 0 touched paths; branch is 15 commits behind current main, with low integration risk given the green synthetic-merge CI.

PR #453, opened 2026-08-25
test(mcp): remove coverage-only timeout override

obra:
## Summary
- remove PR #422's coverage-only MCP connect-timeout override and test config seams
- keep the production 30s init context and 10s per-server MCP deadline
- run the six real Session→stdio discovery paths under production behavior

## Audit rationale
The 48-hour overbuilding audit found that PR #422 added a zero timeout solely to avoid wall-clock waits in six coverage tests. Those tests now use readiness-driven assertions with the unchanged production deadline; healthy stdio discovery completes promptly, while lifecycle cancellation, errors, partial-server behavior, and timeout semantics remain production-owned.

## Verification
- six exact MCP integration tests: `-count=1`, `-count=20`, `-race -count=5`
- `go test ./agent/internal/mcp ./agent`
- `go test -race ./agent/internal/mcp`
- focused gofmt, vet, and diff checks

References #422

PR #458, opened 2026-08-25
fix(web): synchronize interrupt cleanup ownership

obra:
## Summary

Fix the recurring `test-web` interrupt/reap race from PR #453 CI run `32898803620`, attempt 1, job `97967528489`.

## Root cause

The test's real `BASH_ENV` wait seam published `waited-reaped.ready` after the typecheck child had been reaped but before `scripts/web/test-web.sh` removed that PID from its cleanup ownership set. A TERM could arrive in that interval, run the interrupt trap, and signal the stale, already-reaped typecheck PID. The exact interleaving is:

1. `wait typecheck-pid` reaps the child.
2. The recording wait function publishes the completion marker.
3. TERM runs the trap before the script updates its active PID set.
4. Cleanup calls `kill` with the reaped typecheck PID.

## Fix

Track active children explicitly, defer the signal trap until the wait completion/ownership handoff finishes, use Bash's owned-running-job table as the exact completion signal, signal only live owned jobs, and wait/reap the exact remaining ownership set.

## TDD and verification

- Old script mutation: deterministic RED; the focused regression observed `killed-reaped` and failed.
- Corrected `TestMakeTestWebInterruptDoesNotSignalReapedCheck`: GREEN, `-count=20`.
- After rebasing onto exact base `d4646aec9b8eccc2dc146d40191990682224f5e3`: focused test GREEN, `bash -n` GREEN, `git diff --check` GREEN.
- Full root `go test . -count=1`: GREEN.
- Focused interrupt tests with `-race`: GREEN.
- `make test-web`: GREEN (`web-typecheck`, `web-test`, `web-lint`).
- `make test-dev-tooling`, `make lint`, `go vet .`, and `make lint-gofmt`: GREEN.

One concurrent frontend run had all 367 Vitest files pass but two unrelated browser-process tests hit host `spawnSync /bin/ps ENOBUFS`; an isolated rerun passed. This PR does not treat the PR #453 rerun as remediation.

Refs #453

### DIFF
--- a/cmd/evener-hub/frontend/scripts/browserGuardProcess.mjs
+++ b/cmd/evener-hub/frontend/scripts/browserGuardProcess.mjs
@@ -226,7 +226,7 @@ export function filterCrashpadProcessTable(processList) {
     .join("\n");
 }
 
-export function listSystemProcesses({ processTable } = {}) {
+export function listSystemProcesses({ processTable, processCommand = ["/bin/ps"] } = {}) {
   if (processTable !== undefined) return filterCrashpadProcessTable(processTable);
   // Stream the complete table through awk and retain only full-argv Crashpad
   // candidates. Capturing the unfiltered table exceeds execFileSync's 1 MiB
@@ -234,11 +234,25 @@ export function listSystemProcesses({ processTable } = {}) {
   // than pgrep's comm name, also works on Linux where comm is truncated to 15
   // bytes and would miss chrome_crashpad_handler.
   const filter =
-    "line=$0; sub(/^[[:space:]]*[0-9]+[[:space:]]+[0-9]+[[:space:]]+/, \"\", line); " +
-    "if (line ~ /(^|\\/)chrome_crashpad_handler([[:space:]]|$)/) print $0";
+    "{ line=$0; sub(/^[[:space:]]*[0-9]+[[:space:]]+[0-9]+[[:space:]]+/, \"\", line); " +
+    // Keep the filter narrow, but leave argv0 validation to the JS parser. The
+    // second condition rejects an absolute path appearing after an unrelated
+    // argv0 while still allowing the spaces in a normal macOS Chrome path.
+    "if (line ~ /^.*\\/chrome_crashpad_handler([[:space:]]|$)/ && " +
+    "line !~ /^\\/[^[:space:]]+[[:space:]]+\\/.*\\/chrome_crashpad_handler([[:space:]]|$)/) print $0 }";
+  if (!Array.isArray(processCommand) || processCommand.length === 0) {
+    throw new TypeError("processCommand must contain an executable");
+  }
   return execFileSync(
     "/bin/bash",
-    ["-o", "pipefail", "-c", `/bin/ps -axo pid=,pgid=,command= | /usr/bin/awk '${filter}'`],
+    [
+      "-o",
+      "pipefail",
+      "-c",
+      `"$@" -axo pid=,pgid=,command= | /usr/bin/awk '${filter}'`,
+      "browser-guard-ps",
+      ...processCommand,
+    ],
     { encoding: "utf8" },
   );
 }
@@ -261,14 +275,16 @@ function parseProcesses(processList) {
 }
 
 function commandMatchesProfileProcess(command, databaseArg) {
-  const handler = command.match(/(?:^|\/)chrome_crashpad_handler(?=\s|$)/);
-  if (!handler) return false;
-  // ps exposes the executable path and argv as one string, and macOS paths
-  // may contain spaces. An argument such as --arg=/chrome_crashpad_handler
-  // must not become an owned executable; Chrome's argv starts with --, so the
-  // first option is a safe boundary for this identity check.
-  const firstOption = command.search(/\s--(?:\S|$)/);
-  if (firstOption >= 0 && handler.index > firstOption) return false;
+  // ps exposes argv0 and the remaining argv as one string. Anchor the helper
+  // to argv0: a handler-looking positional argument must never own a profile.
+  // macOS app/framework paths contain spaces, so do not split on whitespace;
+  // instead reject a second absolute path before the helper and require a
+  // Chrome/Chromium path when the executable path itself contains spaces.
+  const handler = command.match(/^(.*\/)?chrome_crashpad_handler(?=\s|$)/);
+  if (!handler || (handler[1] && !command.startsWith("/"))) return false;
+  const executablePath = handler[1] ?? "";
+  if (/\s+\//.test(executablePath)) return false;
+  if (/\s/.test(executablePath) && !/(?:chrome|chromium)/i.test(executablePath)) return false;
   return command.includes(`${databaseArg} `) || command.endsWith(databaseArg);
 }
 
@@ -287,12 +303,13 @@ export function findMacOSProfileProcesses(
 }
 
 export function profileProcessIdentityRunning(
-  { pid, databaseArg },
+  { pid, pgid, databaseArg },
   { platform = process.platform, listProcesses = listSystemProcess } = {},
 ) {
-  if (platform !== "darwin") return false;
+  if (platform !== "darwin" || !Number.isInteger(pid) || !Number.isInteger(pgid) || pgid <= 0) return false;
   return parseProcesses(listProcesses(pid)).some(
-    (candidate) => candidate.pid === pid && commandMatchesProfileProcess(candidate.command, databaseArg),
+    (candidate) =>
+      candidate.pid === pid && candidate.pgid === pgid && commandMatchesProfileProcess(candidate.command, databaseArg),
   );
 }
 
@@ -586,7 +603,7 @@ export function createBrowserProcessCleanup({
         const profileProcesses = [
           ...new Map(
             [...profileProcessesBeforeClose, ...findProfileProcesses(profileDir)].map((processIdentity) => [
-              `${processIdentity.pid}\0${processIdentity.databaseArg ?? ""}`,
+              `${processIdentity.pid}\0${processIdentity.pgid ?? ""}\0${processIdentity.databaseArg ?? ""}`,
               processIdentity,
             ]),
           ).values(),
--- a/cmd/evener-hub/frontend/scripts/browserGuardProcess.mjs
+++ b/cmd/evener-hub/frontend/scripts/browserGuardProcess.mjs
@@ -216,8 +216,31 @@ function reportCleanupFailure(error) {
   console.error(error instanceof Error ? error.message : String(error));
 }
 
-function listSystemProcesses() {
-  return execFileSync("/bin/ps", ["-axo", "pid=,pgid=,command="], { encoding: "utf8" });
+export function filterCrashpadProcessTable(processList) {
+  return processList
+    .split("\n")
+    .filter((line) => {
+      const match = line.match(/^\s*\d+\s+\d+\s+(.+)$/);
+      return match && /(?:^|\/)chrome_crashpad_handler(?:\s|$)/.test(match[1]);
+    })
+    .join("\n");
+}
+
+export function listSystemProcesses({ processTable } = {}) {
+  if (processTable !== undefined) return filterCrashpadProcessTable(processTable);
+  // Stream the complete table through awk and retain only full-argv Crashpad
+  // candidates. Capturing the unfiltered table exceeds execFileSync's 1 MiB
+  // maxBuffer on a process-heavy host. Filtering the command column, rather
+  // than pgrep's comm name, also works on Linux where comm is truncated to 15
+  // bytes and would miss chrome_crashpad_handler.
+  const filter =
+    "line=$0; sub(/^[[:space:]]*[0-9]+[[:space:]]+[0-9]+[[:space:]]+/, \"\", line); " +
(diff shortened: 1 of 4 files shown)


obra:
## CHANGES REQUESTED — exact `8a06976aa1918cd617b15b2dd5a832ed283154d3`

**Base:** `d4646aec9b8eccc2dc146d40191990682224f5e3`

### Blocking findings

1. **The trap defers every signal, not just a bounded ownership handoff.** In `scripts/web/test-web.sh:71-75`, `interrupted` only stores a status, and the first/only normal-path `handle_interrupt` call is line 104, after a child wait and its status-file write. A TERM/INT received during `scratch_dir`, the startup loop, or after an async `npm` launch but before the first `wait` therefore returns from the trap and continues launching work. If it arrives before line 94, no later signal interrupts that `wait`; a held first child can block forever instead of being terminated. There is a second loss window in `finish`: lines 58-61 snapshot `interrupt_status` before `stop_checks`/`scratch_rm` and never re-read it, so a signal received during EXIT cleanup can still leave the pre-signal exit status.

   Constrain deferral to the actual bounded ownership handoffs: async launch through `$!` registration, and `wait` through the `jobs -pr`/`forget_pid` decision. Outside those critical sections, stop/reap/exit immediately; after either handoff and at the end of `finish`, consume any recorded signal. Add deterministic real-shell TERM and INT cases for startup/pre-first-wait and cleanup, including a held first check. The existing test holds `test` after `typecheck` completes, so it cannot detect this regression.

2. **The deterministic regression adds a busy poll despite the stated no-poll requirement and `testing.md` policy.** `runtime_pair_build_test.go:764` spins on `while [ ! -f "$EVENER_TEST_SHELL_WAIT_RELEASE" ]; do :; done`. The post-reap gate is otherwise meaningful: it runs the real Make target and real `test-web.sh`, and it does make the old stale-PID kill deterministically RED while the corrected post-wait handoff is GREEN. Keep that coverage, but use an awaitable release (FIFO/pipe/read) instead of a filesystem spin loop. This is not a timeout widening, but it is polling and burns a core during the gate.

3. **The reported `spawnSync /bin/ps ENOBUFS` run has a concrete repository boundary and is not remediated by an isolated rerun.** `cmd/evener-hub/frontend/scripts/browserGuardProcess.mjs:219-220` captures the complete `ps -axo pid=,pgid=,command=` output with `execFileSync` and supplies only `encoding`. Node 26 documents `execFileSync` default `maxBuffer` as 1 MiB and terminates the child when it is exceeded. A process-heavy host with more than 1 MiB of command output therefore explains the exact ENOBUFS family; it also affects real browser-guard cleanup, not only a fixture. Capture/confirm the failed output size and remove the bounded full-table sync capture (for example, stream/parse or target candidates) rather than treating the later pass as closure. The standing sighted-flake policy in `testing.md` explicitly disallows the unrelated-host exemption.

### Confirmed

- The original #453 attempt-1 evidence matches the claimed root interleaving: the BASH_ENV seam completed `command wait`, published `waited-reaped.ready`, and the old trap killed the still-recorded numeric PID; job `97967528489` then reported `killed-reaped`, `wait_for: No record of process 90290`, and Error 143.
- For the exact post-reap window, the new `jobs -pr` check is a truthful Bash-owned running-job test. Completed/unwaited children are not signalled, all retained PIDs are still waited, normal statuses stay mapped to launch order, simultaneous completions are handled, and the three checks remain concurrent. This avoids PID-liveness guesses and late PID-reuse kills.
- No timeout or assertion was widened, no PID was hardcoded, and no production check was serialized. The changed test did add the polling loop noted above.
- Exact patch: 2 files, **+64/-11**: `scripts/web/test-web.sh`, `runtime_pair_build_test.go`.
- Exact-head required jobs are terminal green: CI `build-and-test` (build, vet, lint, all-module race, full deterministic root, dev-tooling, fuzz replay/gap, secret scan), CI `web` (`make test-web`, build, all four browser guards), and `Build snapshot artifacts`. Those green gates do not exercise finding 1 and do not erase finding 3.
- GitHub review threads before this comment: **0**. No approval, rerun, merge, or source mutation performed.

obra:
Amended push: `0ca38f176846f27edb022c3389e194b6073850d8` (parent `8a06976aa1918cd617b15b2dd5a832ed283154d3`, base `d4646aec9b8eccc2dc146d40191990682224f5e3`).

Evidence for all three CHANGES findings:

1. `test-web.sh` now defers only `$!` registration and `wait` through the `jobs -pr`/forget decision. Outside those handoffs, TERM/INT synchronously stop/reap/exit; finish cleanup preserves the signal and retained evidence. Real Make→test-web TERM and INT cases cover a held first child before the first wait and a signal during EXIT cleanup. Fixed GREEN: `go test . -run 'TestMakeTestWebInterrupt|TestMake(TestWeb|BuildWeb)|TestRuntime' -count=1`. Parent RED: replacing only `scripts/web/test-web.sh` with `8a06976^` makes both startup cases fail because `held-typecheck.term` never arrives within the hang tripwire, and both cleanup cases fail because `full logs:` evidence is lost after the signal.

2. The stale-reaped-PID release is now a FIFO rendezvous, not a filesystem spin. The focused stale-PID case remains GREEN on the fixed script; the old script's polling loop cannot consume the FIFO release, so it is the deterministic RED mutation.

3. Browser discovery no longer captures the complete `ps` table synchronously. It streams `ps -axo pid=,pgid=,command=` through `awk`, retaining full-argv Crashpad candidates before Node's sync capture; it does not use `pgrep -x`, avoiding Linux's 15-byte comm truncation false negative. The regression builds a deterministic >1 MiB unrelated table, retains a full `chrome_crashpad_handler` argv, rejects truncated comm and argument false positives, then verifies exact profile ownership/PGID. Focused discovery tests pass:
`node --test --test-name-pattern 'discovers|streams' scripts/browserGuardProcess.test.mjs`.

Local checks passed: `bash -n scripts/web/test-web.sh`, focused/substantial root Go tests, focused browser discovery tests, `git diff --check`. Full browser process tests remain unavailable in this sandbox because `/bin/ps` is denied and macOS `pgrep` reports `sysmond service not found`; CI web/source gates are running unsandboxed.

No main push, merge, approval, or unrelated cleanup. Review threads before/after push: 0.

obra:
## CHANGES REQUESTED — corrected head `0ca38f176846f27edb022c3389e194b6073850d8`

**Incremental parent:** `8a06976aa1918cd617b15b2dd5a832ed283154d3`

### Blocking findings

1. **The wait handoff still has a lost-signal boundary, and cleanup suppresses a signal that can interrupt reaping.** `scripts/web/test-web.sh:130-131` executes `defer_signals=1` as one command and `wait "$pid"` as the next. If TERM/INT is trapped after the assignment but before `wait` starts, `interrupted` records it and returns; the script then enters a live held-child wait with no further signal to interrupt it. The new startup case cannot see this because the production-only hook self-signals at lines 122-124, before `defer_signals` is set. A BASH_ENV wait seam that publishes readiness before `command wait`, followed by a real parent-delivered signal and FIFO release, deterministically reaches the remaining window.

   There is a second reap hole at lines 91-92 and 51-55: while `stop_checks` waits, another trapped signal returns immediately because `stopping=1`; that signal can interrupt `wait`, whose nonzero return is suppressed with `|| :`, after which `active_pids` is cleared even if the child is still live. Retry the exact owned-job wait after interruptions until Bash no longer reports it running. Do not silently discard TERM/INT during reaping.

   The claimed exact-parent startup RED is not mechanism evidence: parent `8a06976` has no `EVENER_TEST_WEB_PRE_WAIT_SIGNAL` hook, so replacing the whole script makes the test fail because no signal is sent at all. Send the signal externally at a real shell rendezvous instead. Then remove the production test controls at lines 68-71 and 122-125; `EVENER_TEST_WEB_FINISH_*` is unused, and `EVENER_TEST_WEB_STARTUP_*` is only dead environment allowlist surface.

2. **Exact PGID ownership is not revalidated before signalling.** `profileProcessIdentityRunning` at `browserGuardProcess.mjs:289-296` destructures only `{ pid, databaseArg }` and accepts a candidate with the same PID/argv but a different PGID. `signalProfileProcess` then sends SIGKILL to the *old captured PGID*. A restarted/reused handler carrying the same private database argument but belonging to a new group can therefore cause an unrelated old/reused group to be signalled. Include `pgid` in the identity comparison and add the same-pid/same-argv/different-pgid negative case; the current reused-PID test changes only the database argument.

3. **The >1 MiB regression bypasses the production `ps | awk` boundary and the executable false-positive check is incomplete.** `listSystemProcesses({ processTable })` returns from line 230 through the JS-only `filterCrashpadProcessTable`; production uses a separately encoded awk program at lines 236-243. Reverting only the production branch to the original full-table `execFileSync` leaves the new test green, so it does not prove ENOBUFS was removed at the real shell boundary. Neither Ubuntu web CI invokes the default Darwin discovery branch, and the reported macOS run could not run `/bin/ps`. Exercise the actual pipeline through a narrow external-process fixture and require the old capture mutation to fail.

   Also add a positional false-positive case. `commandMatchesProfileProcess` accepts `/usr/bin/tool /tmp/chrome_crashpad_handler --database=<exact>` because the handler token occurs before the first option; it then treats the unrelated command as owned. The current `--arg=/chrome_crashpad_handler` case covers only an occurrence after the first option.

### Resolved/confirmed

- The stale-reaped-PID rendezvous is now a blocking FIFO read; the CPU spin is gone.
- The EXIT-cleanup TERM/INT test sends a real external signal to the script PID and checks status plus retained evidence; that correction is meaningful.
- Spawn through `$!` registration is a bounded ownership handoff, and the normal post-reap `jobs -pr`/forget decision remains PID-reuse safe.
- The proposed production process filter uses full argv rather than Linux-truncated comm names and has `pipefail`; the issue is exact identity plus missing production-boundary proof, not the general direction.
- Incremental correction: 4 files, **+282/-10**.
- At review posting, source and refreshed #460 `web` plus snapshot checks are green; both `build-and-test` jobs remain in progress. Source and #460 review threads are both **0**. Green completion would not resolve the findings above.

No broad rerun, edit, approval, merge, or main mutation performed.

obra:
CI completion update for `0ca38f176846f27edb022c3389e194b6073850d8`: unsandboxed `web` passed the frontend tests, build, and all browser guards; `build-and-test` passed build, vet, lint, race, deterministic root tests, dev-tooling, fuzz replay/gap, and secret scan; `Build snapshot artifacts` also passed. Only release/snapshot-channel publishing jobs are skipped by workflow policy. Head remains exact, base remains `d4646aec`, and review threads remain 0.

obra:
Review 5418282431 addressed. Integrated from the exact PR head `0ca38f176846f27edb022c3389e194b6073850d8` and pushed once, non-force:

- Shell source `925b3afedec7252ba075d3973f736d480f159335` → `7343267e91f13cba1653fdc943c5e1175a8583f6`; `go test . -run '^TestMakeTestWebInterruptAtWaitHandoff$' -count=1 -v` (TERM/INT passed).
- Browser source `ea5e6f741f7693230548e9a98d8781f3437dbdd0` → `4adbcd6f3`; focused Node discovery/identity tests passed (4/4).
- Four changed paths are disjoint and exhaustive; no other changes, production test hooks are absent, source patch IDs are preserved, and `git diff --check` passes.

Sandbox caveat: verification ran in the Darwin workspace-write sandbox; the focused browser tests use deterministic process fixtures and do not launch Chrome. Broad suites were intentionally not rerun.

obra:
CHANGES

The refreshed exact head `4adbcd6f3acbf0d5169c5f849b25eb518632a96e` is not green. `build-and-test` fails in the required `make lint` gate at `runtime_pair_build_test.go:1039`: revive reports `redefines-builtin-id` because the test declares `new := ...`, shadowing Go's built-in `new`. Rename that local (for example, to describe the replacement script text) and restore a green required gate.

This same source-owned failure is reproduced on integration PR #460 (`9c46542ce8896e9a95645fd7b1249bb9f7e4b46d`).

Source failure: https://github.com/prime-radiant-inc/evener/actions/runs/32913069094/job/98010952072
Integration failure: https://github.com/prime-radiant-inc/evener/actions/runs/32913264543/job/98011496667

obra:
CI blocker fixed mechanically. Remote head was verified at `4adbcd6f3acbf0d5169c5f849b25eb518632a96e` before the non-force push; new head is `228d1745175419c5a5a40263769277c8a0e19de9`, parent unchanged. Renamed only the local `new` string variable at `runtime_pair_build_test.go:1039` to `deferredWaitBlock`; no semantic change or other edit.

Targeted evidence: `golangci-lint --enable-only revive` (0 issues), one-file `gofmt` check, and `git diff --check` all pass. The diff is exactly 2 insertions/2 deletions in that file; no broad suites rerun.

obra:
PASS

Final focused rereview of exact head `228d1745175419c5a5a40263769277c8a0e19de9` against parent `4adbcd6f3acbf0d5169c5f849b25eb518632a96e`: no findings. The only delta is the behavior-neutral local rename `new` → `deferredWaitBlock` and its use in `bytes.Replace`, which directly clears the revive finding without changing the mutation fixture.

Source #458 build-and-test, web, and snapshot checks are terminal green. Refreshed integration #460 head `96f1e3856ed442d956ad84683e1985bcf5be4ad8` has the identical `runtime_pair_build_test.go` blob; its build-and-test, web, and snapshot checks are also terminal green. Both PRs have 0 review threads. Safe to merge.

No reruns, edits, approval, or merge performed.

Issue #451, opened 2026-08-25
Session Rail job lanes go stale and hide degraded history


obra:
## Summary

Session Rail fetches job intervals once, caches them by thread ref, and does not refresh them when live job state changes. It also has no user-visible degraded state for incomplete or damaged job history.

This is separate from the intermittent socket-close regression and should not enlarge that repair.

## Evidence

- `rail-fixes/.../useRailSummary.ts` performs a one-shot request and caches the response indefinitely.
- `railModelFromSummaryAndTurns` merges live turns into the rail but continues to reuse `summary.jobs`.
- Appwire already publishes `evener/jobs/treeUpdated` and job started/finished notifications.
- The current hook catches every summary error silently.
- A job can start or finish after the cached summary, leaving its lane missing or permanently running.

## Impact

- Live job lanes become stale during the session.
- Reconnect can preserve stale cached job data.
- An incomplete or damaged job projection looks identical to complete history.
- Valid turn summary can disappear when job-only loading fails if the endpoint treats the whole response as fatal.

## Acceptance criteria

- Invalidate/refetch live rail job data on the matching job-tree revision notification and after reconnect.
- Coalesce notification bursts without fixed polling or overlapping requests.
- Keep ended-session cache stable.
- Carry job completeness/revision through the rail model.
- Preserve valid turn summary when only job history is degraded.
- Present a quiet, accessible degraded-job indicator rather than swallowing the state.
- Add frontend tests for running-to-terminal refresh, reconnect invalidation, incomplete history, and job-only failure.

## Scope

This issue belongs to the Session Rail feature branch and is related to #338. It does not require changing the Appwire server's serialized dispatch model.

obra:
2026-08-25 verification update:

Heartbeat starvation is addressed separately by #459. The local combined Rail branch can load and render retained job lanes faster, but `useRailSummary` is still a one-shot cache and still does not refresh job state, carry completeness/revision, or expose a degraded-job indicator.

This issue remains open with its original acceptance criteria unchanged. Live verification rendered job lanes and kept the socket open across heartbeat intervals, but successful older-turn prepend can still hit the separate virtualizer crash tracked in #457.

obra:
## Evidence

Issue #451 is valid as a required acceptance slice of the #338 re-land, not as a shipped current-main regression. Commit `8fec1b124` reverted Session Rail, so `useRailSummary`, its rail model, and the RPC are absent from pinned `da9146ce`.

[...]

Issue #450, opened 2026-08-25
Bounded job activity continuations can replay pages or graft incompatible revisions


obra:
## Summary

`JobActivityTree` is a bounded page, and the current Activity Panel already exposes incomplete branches with a "Load more" action. Two narrower correctness defects remain:

1. A continuation identifies a branch but not a resume position. When a session is truncated by the work-unit or encoded-size bound, following the token can replay the same prefix forever.
2. The Activity Panel accepts a continuation from a different revision and grafts it into the retained tree using the larger revision, instead of discarding the incompatible page and restarting from the root.

This is a low-frequency large-history correctness problem, not a need to make every live delegate field part of a globally serialized snapshot.

## Evidence

- A session with 2,001 retained jobs returns the first 2,000 jobs and a continuation; following that continuation starts from the same first job because the token has no last-emitted identity or offset.
- Encoded-size trimming has the same failure mode: the continuation points back to the session, not the first omitted entry.
- `graftContinuationTree` currently accepts differing revisions and records `Math.max(current.revision, patch.revision)`.
- The current Activity Panel already renders partial retention explicitly and suppresses authoritative counts when `counts.complete` is false. Those behaviors should remain.

## Acceptance criteria

- Document `JobActivityTree` as one bounded page.
- Continuations carry a stable resume position and always make forward progress.
- Public end-to-end tests walk work-unit and encoded-size truncation to completion, proving every retained entry appears exactly once and every page remains within its bound.
- Depth continuations continue at the omitted branch without replaying already returned entries.
- A continuation response whose revision differs from the retained root is not grafted. The consumer discards it and fetches a fresh root page.
- Incomplete branches, branch errors, and diagnostics remain explicit in consumer state and UI.

## Non-goals

- No global lifecycle publication transaction or serialization system.
- No promise that every mutable delegate status field is frozen across a multi-request walk.
- No authenticated cursor, serving-generation protocol, or legacy-token compatibility unless a concrete authorization or security property requires one.
- No global flattened-timeline API; consumers that flatten a tree own their stable ordering.

Related: #338

obra:
2026-08-25 verification update:

Heartbeat starvation is addressed separately by #459. The local combined Rail work preserved the existing bounded `JobActivityTree` projection and did not add continuation following, revision-consistent pagination, or explicit incomplete/degraded semantics.

[...]

Issue #449, opened 2026-08-25
Retained job history lacks owner-authoritative dedupe and integrity diagnostics


obra:
## Summary

Cold persisted job history can select a stale forwarded copy over the owner's authoritative record, and the forensic job reader does not expose enough integrity diagnostics to say whether a returned history is complete.

These defects are independent of the current Session Rail socket regression. They affect every retained job-history consumer.

## Evidence

- Descendant job events are forwarded into ancestor job stores.
- The live recursive jobs path in `agent/jobs_nested.go` has an explicit rule: an owner record beats a forwarded copy.
- A cold consumer that merely deduplicates by `JobID` can keep a forwarded `running` record even when the owner journal contains the completed status, end time, and exit code.
- `agent/internal/jobstore.ReadEvents` tolerates an incomplete trailing append but returns no torn-tail diagnostic.
- `jobstore.Fold` accepts some structurally incomplete lifecycles, such as a finish without a valid start, and returns no error.
- Delegate journals already expose `ReadDiagnostics{TornTail bool}`, so retained job and delegate history currently have different integrity contracts.

## Impact

- Retained jobs can appear permanently running or lose terminal metadata.
- Damaged or torn job history can be returned as complete.
- Consumers cannot distinguish an unavailable owner branch from an authoritative forwarded record.
- UI and forensic tools may silently omit malformed records.

## Acceptance criteria

- Define one owner-authoritative deduplication rule for persisted history: owner records replace forwarded copies; forwarded records are fallback only when the owner is unavailable.
- Validate that job IDs and `OwnerSessionID` agree.
- Expose read diagnostics for torn trailing records and durable corruption.
- Define lifecycle integrity checks for start/finish ordering and required shell-job fields.
- Distinguish fatal root corruption from recoverable descendant damage.
- Mark forwarded fallback or damaged descendant history incomplete.
- Add tests with conflicting owner/forwarded copies, orphan finishes, invalid owners, incomplete JSON, complete JSON without a newline, and unavailable owner branches.

## Scope

This issue defines retained job-history authority and integrity. It does not prescribe a Session Rail representation.

obra:
2026-08-25 verification update:

Heartbeat starvation is addressed separately by #459. The combined local Rail work did not change owner-authoritative deduplication, lifecycle integrity validation, torn-tail/corruption diagnostics, or forwarded-fallback completeness.

[...]

Issue #448, opened 2026-08-25
Persisted job activity reads unbounded journals before applying limits


obra:
## Summary

Persisted job-activity loading reads and decodes complete journals before it applies any activity-tree work, depth, or response-size bound. Large or busy sessions can therefore monopolize a serialized hub RPC and allocate far beyond the advertised response budget.

This was measured while investigating #338: `agent.LoadSessionJobActivityTree` took 32.77s on a retained tree with 562 jobs. Transcript-usage scans caused most of that incident, but the journal path has an independent unbounded-work problem.

## Evidence

- `agent/internal/jobstore/read_events.go` uses `os.ReadFile`, splits the whole file, and decodes every event.
- `agent/internal/delegatestore/read_events.go` also uses `os.ReadFile` and folds the complete journal.
- `agent/jobs_activity_past.go` reads and folds the shared root `delegates.jsonl` once for each visited owner session.
- `agent/jobs_activity.go` applies `activityMaxWorkUnits`, depth, and encoded-size limits only during projection, after recursive loading.
- Journals include notification/watch lifecycle events that consume input work but may produce no projected shell-job row.

The largest observed journal on 2026-08-25 was 6.99 MiB, so this is already a real hot path rather than a theoretical oversized-input case.

## Impact

- Context cancellation cannot interrupt a large current journal read/fold.
- A valid tree can repeatedly decode one shared delegate journal, producing O(sessions × delegate events) work.
- Raw input memory and CPU are not bounded by the activity response budget.
- Any UI or forensic consumer that requests retained job history can stall despite a bounded wire response.

## Acceptance criteria

- Add read-only, context-aware job/delegate journal scanners that do not open stores for append.
- Enforce raw byte and event limits before retaining or folding unbounded input.
- Check cancellation between records, not only between descendant sessions.
- Read and fold the shared root delegate journal once, then index rows by `OwnerSessionID`.
- Apply traversal limits while loading rather than only while projecting.
- Preserve current valid-log ordering and in-flight trailing-record behavior.
- Add deterministic tests for cancellation inside a large journal, many non-job events, raw-limit refusal, one delegate-journal read per root, and no later-file reads after cancellation.

## Scope

This issue covers persisted job-history I/O and resource bounds. It does not require Session Rail UI changes or concurrent Appwire server dispatch.

Related: #338

obra:
2026-08-25 verification update:

[...]

PR #459, opened 2026-08-25
fix: keep WebSocket heartbeat responsive during long requests

obra:
## Summary

- Defer the browser app heartbeat while ordinary RPCs are pending, avoiding a heartbeat request that cannot be answered behind a long-running request.
- Skip/defer server-native pings while the single serial reader/handler is busy. This preserves the one-reader/one-writer design and retains native idle-ping reconnect behavior.
- Add deterministic root-cause evidence with a real-WebSocket starvation test that reproduces the long-request heartbeat failure.

## Verification

All focused, race, and full gates exited zero:

- `npx biome check --write cmd/evener-hub/frontend/src/protocol/client.ts cmd/evener-hub/frontend/src/protocol/reconnect.test.ts` (no tracked diff)
- `make test-web`
- `go test -race ./internal/appserver -count=1`
- `make lint`
- `make vet`
- `make test`

## Scope

Related #338 and follow-ups #448, #449, #450, and #451 remain open; this PR does not close them. Combined Rail, performance, and UI follow-ups are intentionally excluded from this PR.

### DIFF
--- a/cmd/evener-hub/frontend/src/protocol/client.ts
+++ b/cmd/evener-hub/frontend/src/protocol/client.ts
@@ -93,6 +93,7 @@ interface WireMessage {
 }
 
 interface PendingRequest {
+  method: MethodName;
   resolve: (result: unknown) => void;
   reject: (err: Error) => void;
   timer: ReturnType<typeof setTimeout>;
@@ -228,7 +229,7 @@ export class AppwireClient {
         this.pending.delete(id);
         reject(new RequestTimeoutError(`AppwireClient: "${method}" timed out after ${timeoutMs}ms`));
       }, timeoutMs);
-      this.pending.set(id, { resolve: resolve as (result: unknown) => void, reject, timer });
+      this.pending.set(id, { method, resolve: resolve as (result: unknown) => void, reject, timer });
       try {
         socket.send(JSON.stringify({ id, method, params }));
       } catch (err) {
@@ -462,6 +463,13 @@ export class AppwireClient {
     this.reconnectTimer = null;
   }
 
+  private hasPendingOrdinaryRequest(): boolean {
+    for (const slot of this.pending.values()) {
+      if (slot.method !== "ping") return true;
+    }
+    return false;
+  }
+
   // sendHeartbeat sends one app-level ping with an explicit HEARTBEAT_TIMEOUT_MS
   // deadline (reusing request()'s own timeout machinery rather than a second,
   // separately-tracked timer). An open-but-unresponsive socket never recovers
@@ -470,6 +478,7 @@ export class AppwireClient {
   // remains best-effort cleanup: a half-open transport may never emit onclose.
   private sendHeartbeat(): void {
     if (this.connectionState !== "ready") return;
+    if (this.hasPendingOrdinaryRequest()) return;
     const socket = this.socket;
     if (!socket) return;
     this.request("ping", {}, { timeoutMs: HEARTBEAT_TIMEOUT_MS }).catch(() => {
--- a/cmd/evener-hub/frontend/src/protocol/reconnect.test.ts
+++ b/cmd/evener-hub/frontend/src/protocol/reconnect.test.ts
@@ -97,6 +97,78 @@ describe("AppwireClient heartbeat", () => {
     client.close();
   });
 
+  test("keeps heartbeat deferred until the last of multiple successful RPCs settles", async () => {
+    const { factory, sockets } = dialer();
+    const client = new AppwireClient({ url: "ws://x/rpc", socketFactory: factory });
+    await connectReady(sockets, client);
+    const socket = socketAt(sockets, 0);
+
+    const first = client.request("thread/list", { limit: 1 });
+    const second = client.request("thread/list", { limit: 2 }, { timeoutMs: HEARTBEAT_INTERVAL_MS + 1 });
+    const requests = sentFrames(socket).filter((frame) => frame.method === "thread/list");
+    if (requests.length !== 2 || typeof requests[0]?.id !== "number" || typeof requests[1]?.id !== "number") {
+      throw new Error("missing concurrent thread/list request ids");
+    }
+
+    await vi.advanceTimersByTimeAsync(HEARTBEAT_INTERVAL_MS);
+    expect(sentFrames(socket).filter((frame) => frame.method === "ping")).toHaveLength(0);
+
+    socket.receive({ id: requests[0].id, result: { data: [], nextCursor: "" } });
+    await first;
+    expect(sentFrames(socket).filter((frame) => frame.method === "ping")).toHaveLength(0);
+
+    socket.receive({ id: requests[1].id, result: { data: [], nextCursor: "" } });
+    await second;
+    await vi.advanceTimersByTimeAsync(HEARTBEAT_INTERVAL_MS);
+    expect(sentFrames(socket).filter((frame) => frame.method === "ping")).toHaveLength(1);
+    client.close();
+  });
+
+  test("resumes heartbeat after the last concurrent RPC ends with a wire error", async () => {
+    const { factory, sockets } = dialer();
+    const client = new AppwireClient({ url: "ws://x/rpc", socketFactory: factory });
+    await connectReady(sockets, client);
+    const socket = socketAt(sockets, 0);
+
+    const first = client.request("thread/list", { limit: 1 });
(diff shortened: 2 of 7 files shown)


obra:
CHANGES — reviewed exact head `3679ee04a47148a5f11b023a08303ae5cf414c18` against exact base/main `d4646aec9b8eccc2dc146d40191990682224f5e3`.

Blocking findings:

1. **The claimed real-WebSocket starvation RED/GREEN is absent.** The only new server coverage is in `internal/appserver/websocket_keepalive_test.go`, where `scriptedPinger`/`deferrablePinger` replace `*websocket.Conn` and the test directly calls `gate.readerUnavailable()` / `gate.readerAvailable()`. `internal/appserver/websocket_test.go` is unchanged. That fixture choreography cannot prove the actual `ServeWebSocket` wiring, coder/websocket Ping↔Reader dependency, serial `HandleMessage` starvation boundary, or that a real connection survives a handler longer than both native keepalive deadlines. Add a deterministic `httptest` + real coder/websocket case whose handler is held by channels: prove the base behavior closes/starves, then prove this head keeps the connection alive, completes the ordinary RPC, and still enforces idle ping/pong failure afterward. The PR body currently claims this evidence exists when it does not.

2. **The new skip test uses a timing/absence oracle.** `TestWebSocketKeepaliveSkipsPingsWhileReaderIsUnavailable` treats `time.After(20 * time.Millisecond)` as proof no ping began while a 1 ms ticker was running. A delayed keepalive goroutine can make a broken implementation false-green. Replace this with controlled ticker/attempt synchronization (and use timeouts only as deadlock tripwires), consistent with `docs/developing-evener/testing.md`.

3. **Client coverage only proves the success path for one pending request.** The new `reconnect.test.ts` case does not pin last-pending semantics for multiple concurrent callers or heartbeat resumption after an error/timeout, nor does it prove half-open detection/reconnect still resumes after pending work clears. Add fake-clock cases for multiple pending RPCs and each terminal accounting path (success, wire error, timeout, disconnect/close; cancellation if a cancellable request API exists). The production map cleanup looks coherent by inspection, but this behavior is now the liveness gate and needs direct coverage.

Also correct the stale `internal/appserver/server.go` assertion that app-level ping “can't be starved by a busy handler”; serial receive/dispatch is precisely why it can be starved.

CI on this head is green (`web`, `build-and-test`, and `Build snapshot artifacts`), but green existing/fixture tests do not supply the required end-to-end regression proof.

obra:
Addressed exact CHANGES findings on `e11897364200fbfae72d77630be13f7f80534e84` (parent `3679ee04a47148a5f11b023a08303ae5cf414c18`, base `d4646aec9b8eccc2dc146d40191990682224f5e3`):

- Real `httptest` + coder/websocket + `appwire.NewWSTransport` through actual `ServeWebSocket`: held ordinary RPC stays live across serial `HandleMessage`; exact release completes; idle unanswered native ping then closes the client. Mutation removing the reader-busy gate fails with held RPC EOF (RED); fixed test passes (GREEN).
- Replaced the 20ms absence oracle with a controlled ticker and gate-decision signal: busy consumes no attempt, availability produces exactly ping 1, and cleanup completion is awaited.
- Expanded fake-clock lifecycle coverage to 30 reconnect tests: multiple pending RPCs, success/wire error/timeout/disconnect/close terminal paths, last-pending heartbeat resumption, half-open timeout and reconnect. Pending-accounting mutation fails 4 tests (RED); fixed suite 30/30 (GREEN). No cancellation test added because the API has no cancellable request option.
- Corrected `server.go`: app ping bypasses routing but remains in serial receive/dispatch and can be starved by a busy handler.

`websocket.go`/`server.go` production delta is behavior-preserving: unexported per-Server test controls default exactly to existing 15s/10s values, deterministic ticker seam is only used by tests, and the remaining change is the comment. No client behavior changed. Focused `go test -race` appserver passed; focused Vitest reconnect passed 30/30; Biome write/check on touched frontend file passed.

obra:
CHANGES — focused rereview of correction `e11897364200fbfae72d77630be13f7f80534e84` (parent `3679ee04a47148a5f11b023a08303ae5cf414c18`).

The controlled-ticker unit rewrite, fake-clock pending lifecycle matrix, stale `server.go` comment correction, and default production keepalive wiring are sound. One blocker remains:

**The new real-WebSocket regression still uses elapsed time as the causal oracle.** In `TestServeWebSocketKeepsBusyRPCAliveAndDetectsIdlePeerLoss`, the server is configured with a 1 ms interval / 5 ms pong timeout, and the channel-held handler is released by `time.NewTimer(interval + timeout + interval)` (7 ms). The test only goes RED without the busy gate if the keepalive goroutine, ticker, real socket write/read, and pong-timeout path all receive scheduler time inside that 7 ms hold. If they are delayed past release — especially under load or `-race` — the broken implementation can false-green. The subsequent default-select on `busyPingSeen` has the same scheduling dependency.

The prose saying the timer “controls the requested hold duration” does not change that boundary: elapsed time is what decides whether the mutant gets a chance to fail before release. This is the same timing/absence-oracle class called out previously, moved into the integration test.

Please drive the `ServeWebSocket` keepalive attempt with deterministic test control while the handler is channel-held, await an explicit gate/attempt decision, and release only after that decision. Timeouts can remain tripwires and the native pong deadline can remain the behavior under test. This can reuse the new controlled-ticker seam; it should not require a general clock framework.

Everything else in the +323/-23 five-file correction is minimal and behavior-preserving by inspection: `Server` defaults remain exactly 15s/10s, the ticker wrapper preserves production stop/tick semantics, test mutations happen before serving so there is no field race, and the corrected app-ping comment is accurate. #460 CI is terminal green, but its five relevant blob IDs are the pre-correction versions; corrected-head #459 `web` and snapshot are green while `build-and-test` is still in progress at this review.

obra:
Fixed the last blocker in `aa01d291bc87d41d3097ec9d8edc033315119bc0`, parent `e11897364200fbfae72d77630be13f7f80534e84`.

`TestServeWebSocketKeepsBusyRPCAliveAndDetectsIdlePeerLoss` now injects the existing unexported controlled ticker through minimal per-Server test wiring, sends an actual keepalive tick while the real `ServeWebSocket`/coder connection is channel-held in `HandleMessage`, awaits the explicit gate decision (`attempted=false`), and only then releases the RPC. It sends a second controlled tick after release, awaits `attempted=true`, and leaves the real 5ms native pong deadline under test for idle peer-loss closure. No elapsed-time hold or default absence select remains; all timers are hang tripwires.

Focused RED/GREEN evidence: removing the serial `readerUnavailable` transition made the same real-WebSocket test deterministically report `keepalive attempted a ping while the serial RPC handler was busy`; corrected test passed. Corrected integration passed `-count=20`; `-race -count=5` passed. No broad suites rerun.

Production delta remains behavior-preserving: unexported ticker factory/decision controls default to the existing real ticker and 15s/10s values; production decision hook is nil; only tests inject controls. No semantic client/server behavior change.

obra:
PASS — final focused rereview of exact head `aa01d291bc87d41d3097ec9d8edc033315119bc0` (parent `e11897364200fbfae72d77630be13f7f80534e84`).

The remaining timing-oracle finding is closed:

- Actual `ServeWebSocket` receives an injected controlled keepalive ticker while the real coder/websocket handler is channel-held.
- The test awaits the explicit busy-reader decision (`attempted=false`) before releasing the RPC; elapsed time is no longer the release mechanism.
- A second controlled tick awaits the idle-reader decision (`attempted=true`), while the real native pong deadline remains responsible for peer-loss closure.
- Remaining one-second timers are hang tripwires, not behavior or absence oracles.
- The reported no-`readerUnavailable` mutation now fails at the explicit busy decision; focused `-count=20` and `-race -count=5` runs are meaningful GREEN evidence.

The 3-file correction (+42/-32) is minimal and production-behavior preserving. `NewServer` defaults the unexported factory to the same real `time.Ticker`; the production decision callback remains nil; existing 15s/10s values and stop/tick/pong semantics are unchanged. Test controls are installed before serving, so no field race or general clock framework was introduced.

Terminal evidence:

- #459 exact `aa01d291bc87d41d3097ec9d8edc033315119bc0`: snapshot, web, and build-and-test SUCCESS.
- Refreshed #460 exact `b6c7ee49d5013cf22964afd89264b3268331e04e`: snapshot, web, and build-and-test SUCCESS; the three corrected appserver blobs match #459 exactly.
- Review threads: 0 on both PRs.

No finding remains.

PR #469, opened 2026-08-26
fix: bind bounded activity pages to revisions

### LINKED ISSUE #450: Bounded job activity continuations can replay pages or graft incompatible revisions

obra:
Fixes #450

## Summary
- define JobActivityTree as one bounded page with branch-derived completeness
- replace forgeable v1 cursors with authenticated v2 cursors bound to source and serving generation
- reject stale, legacy, malformed, tampered, restart, and live/persisted handoff cursors with structured restart recovery
- validate historical content against authoritative root generation before/after bounded reads without a second full-tree validation scan

## Tests
- RED: mutated source-generation validation made the public live/historical handoff test fail; restored implementation passed
- `go test ./agent ./server ./appwire -count=1`
- `make lint`
- `make vet`
- `make test`
- required GitHub CI: Build snapshot artifacts, build-and-test, web


obra:
Closing this implementation rather than revising it in place.

The underlying large-history bug is real, but this branch adds global publication coordination, lifecycle serialization, and authenticated source/generation cursor machinery while still omitting the cursor resume position. Work-unit and encoded-size continuations therefore replay the same prefix, and the main frontend consumer still grafts mismatched revisions.

Issue #450 has been narrowed to the demonstrated contract: continuations must advance, incomplete state stays explicit, and an incompatible continuation causes a root refresh. A smaller replacement PR will implement that without the global transaction or cursor-authentication machinery.

PR #485, opened 2026-08-26
fix: strip encrypted reasoning on cross-provider openai replay

obra:
## Problem

When switching providers midway through a session, OpenAI Responses `encrypted_content` blobs from a different deployment were replayed verbatim and rejected with:

```
(status=400): responses.create(stream) failed: {"error":{"message":"Encrypted content is not supported.","type":"BadRequestError","param":null,"code":400}}
```

The `thinkingReplayEligible` provenance filter in `agent/session_model_call.go` already enforced same-provider rules for the `anthropic` and `google` families, but `openai` fell through to the `default` case (always eligible). The per-adapter guard in the OpenAI Responses builder (`responses.go:1170`) only skips OpenRouter-style JSON arrays (`IsOpenAICompatEncryptedReasoning`) — a genuine opaque `encrypted_content` blob from a *different* OpenAI deployment passes right through and gets rejected.

## Fix

**1. Extend `thinkingReplayEligible` to the `openai` family** (`agent/session_model_call.go`)

Added `"openai"` to the same-provider case alongside `"google"`. Now thinking/`encrypted_content` blocks are stripped from the outgoing request before it is built when replaying to a different OpenAI deployment. Same-provider stays eligible (the blob may still be valid within the same deployment). Empty provenance (legacy transcripts) stays eligible via the existing early return — no regression for old transcripts.

**2. Diagnostic keyword** (`agent/diagnostic/diagnostic.go`)

Added `"encrypted content is not supported"` to the `isProviderFailure` keyword list, so a legacy-transcript fallback (no per-turn provenance to filter on) surfaces as a Provider error with the retry/switch-model hint instead of an unclassified Evener error.

## What this does not change

- The stored transcript is untouched — the filter only affects the outgoing request (per the existing N4 design).
- The per-adapter guards in Anthropic and openaicompat remain as-is — they already handle cross-family correctly.
- `openai-compat` targets still fall to the `default` case and keep their own builder guard.

## Verification

- `go vet ./agent/...` — exit 0
- `agent/diagnostic` tests — pass (including new `TestClassifyEncryptedContentAsProvider`)
- `agent` replay provenance tests — all pass (including two new OpenAI cases)
- `make merge-approval-gate` — pass (lint 8 modules, build, full test suite, dev-tooling selftest wave)

### DIFF
--- a/agent/diagnostic/diagnostic.go
+++ b/agent/diagnostic/diagnostic.go
@@ -305,5 +305,10 @@ func isProviderFailure(message string) bool {
 		// stream is what failed.
 		strings.Contains(message, "stream ended without") ||
 		strings.Contains(message, "stream error") ||
-		strings.Contains(message, "missing response in finish event")
+		strings.Contains(message, "missing response in finish event") ||
+		// Encrypted reasoning content replayed to a deployment that did not
+		// produce it (a mid-session provider switch on a legacy transcript
+		// without per-turn provenance). The transcript is fine; the provider
+		// rejected a foreign encrypted_content blob.
+		strings.Contains(message, "encrypted content is not supported")
 }
--- a/agent/diagnostic/diagnostic_test.go
+++ b/agent/diagnostic/diagnostic_test.go
@@ -211,6 +211,26 @@ func TestClassifyStreamTruncationAsProvider(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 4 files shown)


obra:
@codex review

<details>
<summary>Model selection</summary>

Reviewer model: `codex-personal/gpt-5.6-sol`
</details>

PR #417, opened 2026-08-24
fix(agent): enforce read-only delegate workspace boundary

obra:
Refs #366

> [!IMPORTANT]
> **Scope / remaining lifecycle defect:** this PR implements only the filesystem/write boundary from #366. It explicitly leaves #366 **OPEN**. The root-vs-live-delegate completion race remains tracked in #366 until the lifecycle work lands; a root can still conclude while a delegate is live, so that requirement remains open.

## Root cause

Read-only delegate roles only had an advisory prompt and a restricted tool registry. A role that retained `shell` could still write through its child process, and an omitted delegate sandbox inherited the parent's off environment. Stable delegate descriptors also had no write-blocked policy to carry across restore.

The first revision added role-derived confinement, but explicit `sandbox` arguments were resolved in a separate branch and skipped role classification. A `read_file`+`shell` role could therefore request `sandbox:"off"` and launch unsandboxed. Cold restore also trusted a descriptor's write-capable snapshot, and the ordinary session-config restore projection dropped the delegate-only `WriteBlocked` bit. Finally, read-only test fixtures used the production host probe, so backendless Linux CI correctly failed closed before reaching their intended assertions.

## RED evidence

The review regression was reproduced through real `createDelegate` before the fix:

```text
TestCreateDelegate_ReadOnlyRoleSandboxRequestFloor/explicit_off
unsafe role sandbox request error = <nil>, want structured invalid_request
TestCreateDelegate_ReadOnlyRoleSandboxRequestFloor/explicit_workspace-write
unsafe role sandbox request error = <nil>, want structured invalid_request
TestCreateDelegate_ReadOnlyRoleSandboxRequestFloor/explicit_restricted
unsafe role sandbox request error = <nil>, want structured invalid_request
```

Cold restore reproduced the same boundary loss:

```text
explicit workspace-write: unsafe restored role policy error = <nil>
explicit restricted: unsafe restored role policy error = <nil>
omitted legacy floor: delegate boundary = sandbox:<nil> wrapper:<nil>
restricted write-blocked floor: writeBlocked:false, want true
```

The failed required Linux race gate at reviewed head was also read directly: existing delegate/plugin/prompt fixtures failed with `cannot enforce --sandbox read-only ... bubblewrap ... not usable on this host`.

## Implementation

- Classify the structured role scope before processing explicit sandbox arguments.
[...]

Issue #366, opened 2026-08-23
Read-only delegates are not write-blocked: a delegate briefed read-only deleted the root's deliverable from the shared workspace, and its report arrived after the run ended


obra:
retest7 n=3, build-pmars trial rUKZXtX. A delegate whose brief said "Use read-only commands only; do not modify files" recursively deleted four paths in the shared /app — the root agent's verified deliverable (source tree + packaging artifacts) — 12 seconds after the root asserted their existence. The root delivered its final answer 4s post-deletion; the delegate's report (which plainly stated the final /app contents) landed 34s after the run was already unrecoverable. Verifier: "No pmars source directory found in /app". A secondary cause: the delegate's earlier extraction defaulted tar into its cwd (/app) instead of its scratch dir, creating the mess it then "cleaned".

Three compounding defects, all code/orchestration rather than prose:
1. A read-only-scoped delegate has no enforcement — the scope is advisory text. Read-only delegates should get a write-blocked view of the shared workspace (or at minimum, deletion outside their own scratch dir should be refused).
2. Delegate cleanup (see #358) is applied to the shared workspace with no ownership test.
3. The root can conclude while a delegate is live and mutating shared state, and the delegate report is delivered when nothing can act on it. The root's final verification (it re-checked the paths pre-delivery) was raced by its own delegate.

obra:
## Evidence

PR #417 merged as `0186a75e324bcd583fb19710dab9cad34ea005f0`. Current admission derives read-only confinement from the effective structured tool ceiling, including roles retaining `shell`; rejects explicit write-capable relaxation; preserves `WriteBlocked` through restore; and leaves only private session scratch writable.

Focused create, restore, and backend-policy tests pass. The decisive unrestricted live gate also passes: `EVENER_SEATBELT_LIVE=1 go test ./agent -run '^TestReadOnlyRoleDelegateUsesRealWriteBlockedBoundary$' -count=1 -v`. Reads and private scratch writes succeed while workspace modify/create/delete operations are denied.

## Root cause

At the incident revision, read-only scope was advisory prompt/tool policy. A retained shell inherited a write-capable execution environment, so prose did not constrain filesystem syscalls.

## Correct plan

Close as completed/obsolete when closure is separately authorized. #417 fixes the exact destructive path. The cleanup-doctrine and broader termination concerns were handled separately; they should not keep this filesystem-boundary issue open.

No additional PR is needed for this issue.

## Risks

[...]

PR #492, opened 2026-08-26
fix(agent): degrade an unenforceable read-only delegate instead of refusing it

### LINKED ISSUE #483: Read-only delegate scope derives an enforced sandbox, so explorer delegates fail outright on hosts without bubblewrap: 53% of trials, and the model abandons delegation in 80% of them
`delegate(agent_type="explorer")` now fails outright on hosts without a sandbox backend, and the model usually abandons delegation rather than recovering.

## What happens

Since `0186a75e3` ("fix(agent): enforce read-only delegate workspace boundary (#417)"), a read-only delegate *scope* derives an enforced read-only *sandbox*:

```go
// agent/sandbox_delegate.go:205
func (s *Session) readOnlyDelegateSandbox() (*sandbox.SandboxPolicy, error) {
	parentMode, parentNetwork := s.parentSandboxModeNet()
	if parentMode == sandbox.ModeRestricted { ... }
	return resolveDelegateSandboxRequest(sandbox.ModeReadOnly.String(), nil, parentMode, parentNetwork)
}
```

`resolveDelegateSandboxRequest` fails closed when no backend can enforce the mode (`agent/sandbox/resolve.go`, `RefusalError`). On a Linux host where bubblewrap is unusable — an ordinary unprivileged container — every `explorer` delegate spawn is refused:

```
per-delegate sandbox: cannot enforce --sandbox read-only (network=true): no sandbox backend is
available (bubblewrap is required on Linux but is not usable on this host); only --sandbox off is supported
```

The caller passed no sandbox argument at all. Choosing the `explorer` agent type is now, by itself, a request the host cannot satisfy.

## Field measurement

From a 135-trial run of `da9146ce1` compared against 270 trials of the previous binary (base `ffa2c09c7`), same environment:

| | trials hitting the refusal |
|---|---|
| `da9146ce1` | **71 / 135 (52.6%)** |
| previous binary | 25 / 270 (9.3%) |

All 71 refusals were `delegate(agent_type="explorer")` **with no explicit `sandbox` argument**. On the previous binary the refusal only fired when the model explicitly asked for `sandbox: read-only`, which is why the rate was low.

What the model did next, across those 71 refusals:

- **57** — no delegate retry at all; it abandoned delegation and continued alone
- 7 — retried with `sandbox` off (the intended recovery)
- 7 — retried `delegate` with other arguments

So the dominant outcome is not a wasted round, it is a silently lost capability: the parent drops the investigation channel it had just decided it wanted. Compare a spawn on the previous binary, same call shape, which returned a working `delegate_id`.

## Why this is a regression rather than the intended fix

Fail-closed enforcement is the right instinct and is what #366 asks for — an advisory read-only scope let a delegate delete the root's deliverable. The defect is the coupling: **the only enforcement path is an OS sandbox**, so on a host with no backend the choice collapses to "refuse the spawn." The scope's actual contract (this delegate must not write the shared workspace) is enforceable in-process at the tool layer without bubblewrap, which would keep the capability on unsandboxable hosts while still honoring #366.

This also re-opens the substance of #394 (closed as completed) from a new direction: that issue fixed the *message* so a rejected request names what to change, on the assumption the model asked for the knob. Here the model asked for nothing, so a better message is not a fix.

## Design question for triage

Three candidate directions, and I do not think the choice is mine to make:

1. **Tool-layer write-blocking** when no sandbox backend exists — enforce the read-only contract by refusing writes in the delegate's file tools and its shell, no OS sandbox required. Keeps the capability, keeps #366 closed; residual risk is whatever a shell can do that the tool layer does not intercept.
2. **Degrade with disclosure** — spawn unsandboxed, and tell both the parent and the delegate in-band that the read-only boundary is advisory here.
3. **Keep refusing, but make the recovery unmissable** — the refusal must state the exact retry, and the delegate tool should stop offering scopes the host cannot enforce (the durable half of #348/#394).

(1) preserves the most and gives up the least, but it is a real chunk of work and touches the tool boundary.

## Repro

Deterministic, no eval needed: on a Linux host without usable bubblewrap, run a session on `da9146ce1` and call `delegate` with `agent_type: "explorer"` and no sandbox arguments. Compare against a binary from before `0186a75e3`.

obra:
Closes #483.

## Mechanism

A read-only delegate *scope* no longer requires an OS sandbox. Where no backend can enforce `ModeReadOnly`, the scope derived from an agent type resolves to `Mode: off, WriteBlocked: true`: the delegate spawns, its **file tools** carry the whole write boundary (`FileTool.WriteRoots` empty, session scratch the only writable place, the same credential/pseudo-fs denylist every confining mode masks), and its **shell** is unconfined. Both sides are told which half holds.

## The four parts

1. **The derived scope degrades.** `readOnlyDelegateSandbox` returns `{Mode: off, WriteBlocked: true}` when the host has no usable backend, instead of a `read-only` request that `Resolve` refuses. A write-blocked off policy takes the ordinary resolver path — denylist, git surface, scope builder — and skips only `chooseBackend`, which is the thing it has none of.

2. **An explicit `sandbox: read-only` still fails closed.** Only the derived scope degrades. An explicit request passes through unmodified and the spawn surfaces the resolver's `*sandbox.RefusalError`. A caller that states the contract is never silently handed a weaker box.

3. **File-tool confinement got its own predicate.** `ResolvedPolicy.FileToolConfined()` — `Mode != ModeOff || WriteBlocked` — sits beside `Enforced()`, which keeps its meaning ("an OS sandbox is in force") untouched.

4. **Disclosure in both directions.** The delegate's environment section says the boundary is ENFORCED for its file tools and ADVISORY for its shell, names its scratch dir, and warns that its file tools will not traverse a symlinked directory. The parent's spawn result carries a matching warning on the same non-blocking channel the shared-workspace advisory uses.

## `Enforced()` call-site audit

Changed to `FileToolConfined()`:

| Site | Why |
|---|---|
| `execenv/local.go` `sandbox()` | The single gate on building the fd-anchored file-tool layer — the file-tool question by definition. Leaving it made `WriteBlocked` inert. |
| `sandbox/reroot.go` `WithSessionScratch` | The late-bound scratch grant belongs to the file-tool layer. Without it a degraded delegate has *nowhere* writable, breaking `WriteBlocked`'s documented contract. |
| `agent/session_prompts.go` `sandboxPromptLine` | The line tells the model the boundary it runs under. A confined-but-unenforced delegate has one, and it needs different words (naming the mode would say "off"). |
| `sandbox_delegate.go` `parentSandboxFloorForEnv` | `WriteBlocked` is a floor a child must never drop, and a degraded parent's block is real. Read through the OS predicate it reported *no* block, so a child could ask for writes its parent does not have. Not reachable today — the child inherits an already-stripped registry and the schema hides the argument on a backendless host — but a floor that depends on coincidence is not a floor. |
| `sandbox_delegate.go` restore's inherit-write-blocked branch | Paired with the above: whatever reports the block is what a restoring child inherits, or every child of a degraded parent fails to restore. |

Kept on `Enforced()`, each asking the OS-sandbox question:

- `execenv/local.go` `EnableSandbox` — an off policy still builds no wrapper and no bwrap argv. It now provisions an **owned** session scratch when `FileToolConfined()`.
- `execenv/local.go` `WithSandboxInvocationGrant` — **deliberately not changed.** `escalationAllowed` refuses a subagent session, and only delegates get a write-blocked off policy, so no such env can reach it. Widening the gate would add a branch nothing can execute and a test pinning an unreachable state. The reasoning is recorded at the call site with the condition that should reopen it.
- `session_capabilities.go` (×2) — **decided, not omitted.** Every line it gates describes OS-enforced state: `writableRootsSummary` is built from the **spawned** layer, the cache strategy is a wrapper env-floor behavior, and the `go` telemetry note reports a kernel denial. For a wrapper-less box each would be *false* rather than merely absent — "Writable roots: none" would describe a shell that can in fact write anywhere. The comment now records that.
- `sandbox_delegate.go` `sandboxSnapshotFromEnv`, `delegate_runtime.go` inherit-parent-snapshot — must stay paired with `stableDelegateSandboxSnapshot`/`sandboxSnapshotFromInputs`, which drop `ModeOff`. Changing the gate alone is a **no-op** (the snapshot is nil either way); changing the whole family would break `frozenStableDelegateSandboxMatches` at construction *and* collide with the corrupt-descriptor guard below. The degraded box is re-derived per host on resume rather than persisted — there is no OS mode to record.
- `sandbox/startup.go` `EnforcementLine`, `cmd/evener/sandbox.go` `reconcileClearSandbox`, `sandbox/reroot.go` `ControlPolicy`, `sandbox/contract.go` (×2) — the startup banner, the persisted session mode, whether there are writes to widen, and the backend contract matrix.

## Where the file tools cannot enforce

`securepath_other.go` (`!linux && !darwin`) fails closed on *every* operation, reads included, so a write-blocked box there is a delegate with broken file tools rather than a confined one. `Resolve` **refuses** it (`FileToolEnforceable`), so no caller can produce that env; the derivation checks the same predicate to avoid asking. `securepath_other.go`'s header now names both invariants and their guardians.

## Tests

Each was watched failing first, for the reason named.

- `TestCreateExplorerDelegateDegradesInsteadOfRefusing` — the field repro end to end: `delegate(agent_type="explorer")`, no sandbox argument, backendless host. Before the fix it reproduced the reported error verbatim; now it returns a `delegate_id` and the parent's disclosure.
- `TestDegradedReadOnlyDelegateFileToolsRefuseWrites` — a real delegate's `write_file`/`edit_file` into the shared workspace return a typed `*sandbox.DeniedError` with the disk untouched, while reads and its scratch work.
- `TestWriteBlockedOffMasksCredentialPaths` — a degraded env denies `~/.ssh/id_rsa` and evener's own `/proc/<pid>/environ`, while an ordinary out-of-worktree read still works.
- `TestReadsThroughSymlinkedWorkspaceRoot` — read/list/glob through a **genuinely symlinked** workspace root, for the degraded box *and* enforced read-only; asserts the anchor grants no write and still refuses a symlink component inside the worktree. The root is deliberately not canonicalized — doing that in the fixtures is what hid this.
- `TestDegradedDelegateSpawnFailureDisposesItsScratch` — driven through the real abort path (a nil child client fails `NewSession` after `prepareSubagentEnvironment`); no scratch dir, and no held lease, survives.
- `TestDegradedParentPropagatesItsWriteBlock` — a degraded parent reports its block, refuses a write-capable child, and a restoring child inherits the box.
- `TestDegradedAdvisoryPromisesOnlyWhatThisHostDelivers` — pins the fact behind the reworded advisory: an isolated delegate on this host gets no kernel wrapper either.
- `TestResolveWriteBlockedOffConfinesFileTools` / `...NeedsAnAbsoluteHome` / `...RefusesWhereFileToolsCannotEnforce`, `TestFileToolConfinedSeparatesFromEnforced`, `TestWriteBlockedOffGrantsSessionScratch` — the resolved grants (denylist, read anchor, git layout, no write roots), the homeless-service refusal, the non-enforceable-platform refusal, and the predicates.
- `TestExplicitReadOnlyDelegateSandboxStillFailsClosed`, `TestReadOnlyDelegateSandbox_DegradesWhenHostCannotEnforce` / `_NoDegradeWhereFileToolsCannotEnforce`, `TestRestoreDelegateSandboxFloorDegradesForReadOnlyCeiling`, `TestSandboxPromptLineDisclosesDegradedReadOnlyDelegate`, `TestCreateExplorerDelegateEnforcedHostSaysNothing`.

`agent/sandbox/contract.go`'s table was deliberately **not** extended: it is the (mode × host × net) matrix each *backend* is held to, and the degraded box has no backend.

One existing table row moved with the semantics: a *persisted* descriptor recording `off` + `WriteBlocked` is refused explicitly rather than treated as an ignorable flag. No spawn writes one, so it can only be corrupt or hand-edited.

## Residual gaps

1. **The shell is unconfined** on a backendless host — `shell(command: "rm -rf ...")` is not blocked, only discouraged in the prompt and reported to the parent. No sandbox mode closes it there, and the advisory no longer pretends otherwise: `isolation="worktree"` is offered as a separate checkout so an *accidental* write lands off your tree, not as a boundary.

2. **A symlinked directory inside the workspace is not traversable.** The workspace *root* being reached through a symlink is fixed (reads anchor at the worktree, which tolerates a symlinked ancestor — enforced read-only benefits identically and had the same gap since it shipped). What remains is a symlink *component* below the root, and a symlinked component in an out-of-worktree path: `pnpm`/`node_modules` links, Nix store links, Debian usrmerge `/lib`. That is the same refusal every enforced mode makes, but it is a **new cost** for a delegate that previously had no sandbox at all, so this PR does not claim to be strictly stronger than the pre-#417 scope — it is stronger on writes and credential reads and weaker on symlink traversal. The delegate's prompt says so as an actionable hint ("name the real path"); enforced modes share the limitation and their prompts are unchanged, to avoid unrelated prompt churn.

3. **Degraded delegates lose ripgrep.** Any env with a file-tool layer takes the native Go walk. Parity with enforced read-only, but a new cost on exactly the explorer workload this PR restores.

4. **Resume across a host capability change is asymmetric, intentionally.** A read-only delegate created on a backend-capable host persists `{mode: read-only}` and fails closed on resume if the host later loses its backend, while an identical fresh delegate degrades. The persisted request is a stated contract and keeps failing closed; the derived scope is a host-dependent decision and is re-made against the host that is actually running it.

## Gates

- `agent`, `agent/sandbox`, `agent/execenv`, root, `llm`, `auth`, `envvars`, `invariant`, `identifier` — pass (`-short -count=1`, per module under `go.work`). `-race` over the sandbox/scratch/symlink tests in `agent` and `agent/execenv` — clean.
- `make lint` — PASS (8 modules). `make lint-evenerfuzz`, `make lint-eval` — PASS standalone, including the `GOOS=linux` vet repeats under both tags.
- Pre-existing macOS failures, unchanged and verified identical on a pristine `a0bfbd3fb` worktree: `TestMakeWebCommandsContainNodeProcessState`, `TestMakeTestWebRetainsFailedProcessStateWithinTMPDIR`, `TestMakeTestWebInterruptRetainsEvidenceAndReapsChecks` — the `/var` → `/private/var` TempDir symlink class.
- `agent/execenv`'s `TestGrepRipgrepStopsWhenContextIsCanceled` flakes under full-package load (a 10s timeout waiting for a spawned shell script). It passes 3/3 in isolation, and its env carries no sandbox policy at all, so no path in this diff executes in it.

### DIFF
--- a/agent/delegate_runtime.go
+++ b/agent/delegate_runtime.go
@@ -1257,10 +1257,13 @@ func (runtime delegateRuntime) create(ctx context.Context, args delegateArgs) de
 	s.startDelegateQuietWatchdog(started.ctx, started.lease)
 	s.launchSubagentRun(prepared.runCtx, prepared.sub, prepared.runCancel, prepared.input, started.descriptor.Provenance)
 	result := createResult(stableDelegateResult(started.descriptor, started.lease.delegateID, started.plan, plans, nil))
-	// The advisory rides the launched delegate's own result only: it is
+	// The advisories ride the launched delegate's own result only: they are
 	// metadata for the caller's next isolation choice, not delegate output, not
 	// an EventWarning, and not durable job state a later delegate_send replays.
-	result.Warnings = appendUniqueStrings(result.Warnings, sharedWorkspaceAdvisory)
+	// The sandbox one fires when a derived read-only scope had to degrade on this
+	// host, so the parent learns the boundary is advisory for the child's shell
+	// rather than discovering it from a clobbered file.
+	result.Warnings = appendUniqueStrings(result.Warnings, sharedWorkspaceAdvisory, degradedReadOnlyDelegateAdvisory(requestedSandbox))
 	return result
 }
 
--- a/agent/delegate_runtime.go
+++ b/agent/delegate_runtime.go
@@ -1263,7 +1263,7 @@ func (runtime delegateRuntime) create(ctx context.Context, args delegateArgs) de
 	// The sandbox one fires when a derived read-only scope had to degrade on this
 	// host, so the parent learns the boundary is advisory for the child's shell
 	// rather than discovering it from a clobbered file.
-	result.Warnings = appendUniqueStrings(result.Warnings, sharedWorkspaceAdvisory, degradedReadOnlyDelegateAdvisory(requestedSandbox))
+	result.Warnings = appendUniqueStrings(result.Warnings, sharedWorkspaceAdvisory, degradedReadOnlyDelegateAdvisory(prepared.sub.sess.currentEnv()))
 	return result
 }
 
--- a/agent/execenv/filetool_capability_other.go
+++ b/agent/execenv/filetool_capability_other.go
@@ -1,0 +1,7 @@
+//go:build !linux && !darwin
+
+package execenv
+
+// FileToolEnforceable reports whether this process can use the race-safe secure
+// open primitive that confines file-tool operations on the current host.
+func FileToolEnforceable() bool { return false }
--- a/agent/execenv/filetool_capability_unix.go
+++ b/agent/execenv/filetool_capability_unix.go
@@ -1,0 +1,19 @@
+//go:build linux || darwin
+
+package execenv
+
+import "golang.org/x/sys/unix"
+
+var securePathCapabilityProbe = func() error {
+	fd, err := openAbsNoSymlinks("/", unix.O_RDONLY|unix.O_DIRECTORY, 0)
+	if err != nil {
+		return err
+	}
+	return unix.Close(fd)
+}
+
+// FileToolEnforceable reports whether this process can use the race-safe secure
+// open primitive that confines file-tool operations on the current host.
+func FileToolEnforceable() bool {
+	return securePathCapabilityProbe() == nil
+}
--- a/agent/execenv/filetool_capability_unix_test.go
+++ b/agent/execenv/filetool_capability_unix_test.go
@@ -1,0 +1,32 @@
+//go:build linux || darwin
+
+package execenv
+
+import (
+	"errors"
+	"testing"
+
+	"golang.org/x/sys/unix"
+)
+
+func TestFileToolEnforceableRequiresSecureOpen(t *testing.T) {
+	original := securePathCapabilityProbe
+	t.Cleanup(func() { securePathCapabilityProbe = original })
+
+	for _, probeErr := range []error{unix.ENOSYS, unix.EPERM} {
+		securePathCapabilityProbe = func() error { return probeErr }
+		if FileToolEnforceable() {
+			t.Errorf("secure-open failure %v reported file-tool enforcement available", probeErr)
+		}
+	}
+
+	securePathCapabilityProbe = func() error { return nil }
+	if !FileToolEnforceable() {
+		t.Error("a successful secure-open probe reported file-tool enforcement unavailable")
+	}
+
+	securePathCapabilityProbe = func() error { return errors.New("unexpected secure-open failure") }
+	if FileToolEnforceable() {
+		t.Error("an unexpected secure-open failure reported file-tool enforcement available")
+	}
+}
--- a/agent/execenv/local.go
+++ b/agent/execenv/local.go
@@ -198,39 +198,27 @@ func (e *LocalExecutionEnvironment) sandbox() *sandboxFS {
 //
 // A write-blocked policy with no OS sandbox (a read-only delegate on a host with
 // no sandbox backend) never builds a wrapper to read through, so it falls back to
-// the env's own session scratch — the SAME directory overlaySessionEnv exports to
-// its spawned commands, since that path also keys on a nil Wrapper. Without this
-// the file tools of a write-blocked env would have no writable root at all,
-// breaking WriteBlocked's contract that the session scratch stays writable. An
-// ENFORCED policy is deliberately excluded from the fallback: its scratch is the
-// wrapper's, and a wrapper-less enforced env (test-only — EnableSandbox fails
-// closed rather than half-wiring one) must not silently mint a second one.
+// the scratch EnableSandbox provisioned for it, which this env OWNS — the same dir
+// overlaySessionEnv exports to its spawned commands, so the model's file tools and
+// its shell agree on one writable place. Without it the file tools of a
+// write-blocked env would have no writable root at all, breaking WriteBlocked's
+// contract that the session scratch stays writable.
 func (e *LocalExecutionEnvironment) sessionScratchPath() string {
 	if e.Wrapper != nil {
 		return e.Wrapper.SessionTmp()
 	}
-	if e.Sandbox != nil && !e.Sandbox.Enforced() && e.Sandbox.FileToolConfined() {
-		if e.sandboxGrant != "" {
-			// A short-lived per-invocation grant clone must not provision a scratch dir:
-			// it owns nothing and nobody retains or disposes what it allocates, so a
-			// second dir would leak on every approval. The one approved leaf resolves
-			// through the grant, which needs no root.
-			return ""
-		}
-		return e.unsandboxedScratchDir()
-	}
-	return ""
+	return e.wrapperlessScratchDir()
 }
 
-// allocatedSessionScratchPath returns an already-provisioned scratch root owned
-// by this environment. It deliberately does not call unsandboxedScratchDir:
-// checking an arbitrary write path must never allocate a new grant as a side
-// effect. Enforced environments are handled by sandbox(), whose policy already
-// folds Wrapper.SessionTmp() into the fd-anchored roots; this accessor is for the
-// otherwise-unconfined environment's late-bound scratch grant.
-func (e *LocalExecutionEnvironment) allocatedSessionScratchPath() string {
-	if e.Wrapper != nil {
-		return e.Wrapper.SessionTmp()
+// wrapperlessScratchDir returns the already-provisioned per-session scratch of an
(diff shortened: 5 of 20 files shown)


obra:
Field result from the eval run of this branch (135 trials, ): the fix does what it claims, plus two things worth knowing before merge.

**Confirmed working.** Explorer delegate spawns refused: 71/135 before, **0** after. Degrade disclosed to parent and delegate: **59 of 59** — nothing degraded silently. Delegates actually running: 24 → 73. No symlink-traversal refusals fired anywhere.

**The write-block is unreachable for the role it most affects.** The `explorer` tool set is `communicate, compact_context, glob, grep, read_file, read_transcript, shell, task_list` — no write tools, so zero file-tool denials fired in 53 explorer trials, and a delegate deleted a parent workspace file through its shell (`rm -f /app/image.ppm` in `path-tracing-reverse__NF9CbHz`). Detail and options on #366; the PR's advisory is honest about it, so this is a scope question rather than a defect in the change.

**New failure mode, filed as #496.** Delegates run their shell commands in the parent's cgroup with no shared memory budget, and each session is told the container's full `memory_mb` as if it owned it. On `video-processing` (2048 MB cap) three of three trials ended in exit 137 — the evener process SIGKILLed by the OOM killer — against zero such exits in the baseline. The refusal this PR removes had been acting as an accidental concurrency governor. This is not an argument against the PR (refusing every spawn was worse), but the governor needs replacing.

**Not a scoreboard regression.** The run scored 91/135 against the baseline's 98, but the drop is not concentrated in delegate trials: 30 explorer-using tasks fell 4 trials while 15 non-explorer tasks fell 3, and two identical-binary runs of this set differ by 6. Delegate-spawning trials passed slightly *more* often than non-delegate ones within this run.

Issue #592, opened 2026-08-29
action-pin parity audit: case-sensitive owner matching and one-sided-conflict diagnostic completeness


obra:
Filed from adversarial review rounds 2-4 of the #288 fix (PR pending). Two nits: (1) the `^actions/([a-z0-9-]+)` regex silently skips case-varied pins (actions/Setup-Node@v7 does not match — verified by probe); (2) when one side carries a single ref and the other a multi-ref set, only the conflict diagnostic fires — the major-version divergence on the overlapping subset is not separately reported (diagnostic completeness only; the error still fails loudly).

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

**Verdict:** `already-fixed` — both nits concern the action-pin parity audit's regex/conflict diagnostics, and that audit no longer exists at HEAD (no `actionRefPattern`/`action-pin`/`workflow_action` symbol anywhere; no Go source referencing `github/actions`). Commit `2eeacb483d` (#593) removed it under the #288 round-5 owner note: "the parity-audit tests are overkill and go away." `2eeacb483d` is an ancestor of HEAD. The case-varied-owner skip and the one-sided-conflict diagnostic gap therefore cannot occur.

**Recommendation:** `close` — moot by the owner's deletion of the audit.

Issue #591, opened 2026-08-29
action-pin parity audit: recurse into nested composite action directories and define action.yml/action.yaml precedence


obra:
Filed from adversarial review rounds 2-4 of the #288 fix (PR pending). Two discovery gaps in workflow_action_parity_test.go: (1) composite discovery is depth-1 only — a composite nested under .github/actions/<parent>/<child>/ escapes the audit; (2) when both action.yml and action.yaml exist in one directory, existingManifest reads only action.yml with no defined precedence and no diagnostic. No current instance of either shape in the repo (verified), so both are latent.

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

**Verdict:** `already-fixed` — the discovery code this issue describes (`workflow_action_parity_test.go` depth-1 composite discovery, `existingManifest` reading only `action.yml`) is not in the tree at HEAD (`find . -name 'workflow_action_parity_test.go'` → none; no Go file mentions `github/actions`). It was removed, not fixed: commit `2eeacb483d` (#593) records the #288 round-5 owner ruling that "the parity-audit tests are overkill and go away." `2eeacb483d` is an ancestor of HEAD. The two latent shapes filed here (nested `.github/actions/<parent>/<child>/`, and `action.yml` vs `action.yaml` precedence) therefore have no reader to fix.

**Recommendation:** `close` — moot by the owner's deletion of the audit. If the drift guard is ever reinstated, both shapes are worth re-filing together with #590/#592.

Issue #590, opened 2026-08-29
action-pin parity audit: extend coverage beyond the actions/* namespace


obra:
Filed from adversarial review rounds 2-4 of the #288 fix (PR pending). Reviewer-proven: the parity audit's regex `^actions/([a-z0-9-]+)@(.+)$` restricts coverage to first-party actions/* pins — a third-party action (e.g. goreleaser/goreleaser-action) pinned at v7 in a workflow and v6 in a composite drift passes silently (harness-verified). The exclusion is documented as intentional in the code comment, but Dependabot's blindness to composite pins applies equally to third-party actions. Extending the namespace coverage is a deliberate scope decision recorded here rather than expanding the #288 PR.

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

**Verdict:** `already-fixed` — the parity audit this issue asks to extend does not exist at HEAD. There is no `actionRefPattern`, `action-pin`, `workflow_action*`, or any Go source mentioning `github/actions` (`rg -n 'actionRefPattern|action-pin|actionPin|workflow_action' .` → none; `rg -l 'github/actions' --glob '*.go'` → none). PR #288's round-5 owner note removed the audit: commit `2eeacb483d` (#593) message states "the parity-audit tests are overkill and go away, together with the round-4 third-party broadening they carried. The pin bump stays; the drift guard itself is removed." `2eeacb483d` is an ancestor of HEAD. The round-4 broadening this issue tracked was implemented and then deliberately withdrawn, so the `actions/*`-only namespace restriction no longer exists.

**Recommendation:** `close` — moot by the owner's deletion of the audit; re-filing the third-party coverage question would require first deciding to reinstate the guard.

PR #593, opened 2026-08-29
fix(ci): align setup-toolchain composite pins with workflow pins and add parity audit (#288)

### LINKED ISSUE #288: setup-toolchain composite action still pins setup-node@v6 after #284's v7 bump
Found during the dependabot batch review (#284): dependabot bumped ci.yml's direct `actions/setup-node@v6` reference to v7, but the composite action `.github/actions/setup-toolchain/action.yml` (used by ci.yml's build-and-test job and binaries.yml's build job) still pins `setup-node@v6`. Dependabot doesn't scan composite actions here. Not breaking, just inconsistent — bump the composite action's pin to match, and consider adding `.github/actions/**` to dependabot's watch list so the pair can't drift again.

obra:
Closes #288.

## Root cause
`.github/workflows/ci.yml` pins `actions/setup-node@v7` directly while `.github/actions/setup-toolchain/action.yml` (the composite used by the build jobs) still pinned `@v6`. Dependabot's github-actions updater does not recurse into local composite actions (dependabot.yml watches only `/`), so the pair drifts silently — exactly what happened after #284's v7 bump.

## Fix (4 commits, hardened through 4 adversarial review rounds)
1. `41256b30f` — bump the composite to setup-node@v7 (matching ci.yml's direct pin; setup-go was already aligned; checkout is workflow-only) + add `workflow_action_parity_test.go`: a deterministic, no-network audit parsing all workflows and composite actions, requiring major-version parity for any action used both directly and in a composite.
2. `2796e0e68` — harden the audit: per-action ref SETS (cross-workflow conflicts flagged) instead of last-file-wins collapse; fail on unparseable-major refs instead of silently skipping; discover `action.y*ml`.
3. `b38a2b16c` — accept identical single refs on both sides (identical pins cannot drift — fixes a false-positive on SHA-pinned pairs) + pin cross-workflow conflict detection with its own test.
4. `1b322e175` — scope the identical-refs pass to single pins (mirrored multi-ref sets fall through to conflict detection) + mirrored-set regression test.

## Tests
7 tests in `workflow_action_parity_test.go` (repo-tree + fixture-based regression coverage for every hardening rule), all passing `-count=1`. `make vet`, `make lint` (all sub-gates) PASS.

## Review
Four rounds of two independent adversarial reviewers (competitive protocol with disqualification for false/inflated findings). Rounds 1-3 each found real defects in the audit (ref-collapse false-pass — reproduced; SHA-pin false-positive — reproduced; mirrored-multi-ref-set mask — found independently by both reviewers), all fixed. Round 4: both reviewers **approve with zero critical/major findings**; R3-A closure verified by mutation testing. Out-of-scope findings filed as #590 (third-party namespace coverage), #591 (nested composites + yml/yaml precedence), #592 (case-sensitivity + diagnostic completeness) rather than expanding this PR.

Simplify pass applied before each commit round; tests re-run green after each.


obra:
these tests are overkill. they should go away

PR #604, opened 2026-08-30
consolidate duplicated intent-reader selection loop into toolargs.FirstNonEmpty

obra:
## Summary

The "read intent from tool args" logic was duplicated at **5 sites** across packages that cannot import each other. Each site walks a fixed list of candidate keys (`intent`/`purpose`/`description`) and returns the first non-empty value, but varies in how it extracts a string from the decoded map and whether it trims whitespace.

This PR extracts the shared key-selection loop into `internal/toolargs.FirstNonEmpty(get func(string) string, keys ...string) string`, consolidating the duplication **without changing any function signature**. Each call site keeps its exact extraction behavior inside its `get` closure.

## The 5 sites

| # | Site | Input | Keys | Extraction | TrimSpace |
|---|------|-------|------|------------|-----------|
| 1 | `toolStartDescription` (`agent/session_tools.go`) | `map[string]any` | intent, purpose, description | string-only | no |
| 2 | `toolIntent` (`agent/transcript_render.go`) | `json.RawMessage` | intent, purpose, description | `scalarString` (bool/num coercion) | no |
| 3 | `ToolIntentFromArguments` (`internal/apptranscript/apptranscript.go`) | `json.RawMessage` | intent, purpose, description | string-only | yes |
| 4 | `SummarizeToolInDir` (`cmd/evener-tui/internal/toolsummary/tool_summary.go`) | `map[string]any` | intent, purpose, description | string-only | no (downstream `trunc`) |
| 5 | `RenderToolCall` (`cmd/evener-tui/internal/msgrender/message.go`) | `ToolArgs` map | intent, purpose | string-only | yes |

## Why a shared helper is feasible

The **selection loop** ("iterate keys, return first non-empty") is genuinely duplicated. Site 4 already factored it into a *local* `firstNonEmpty` helper — this PR lifts that same helper to a module-root `internal/` package so all 5 sites share it.

The **per-site variation** (string-only vs scalar coercion, TrimSpace vs none, 2-key vs 3-key) is *not* duplicated — it's genuinely different behavior. The parameterized `FirstNonEmpty` preserves all of it via each caller's `get` closure and key list, so:

- No function signature changes (no behavior change)
- All test-pinned behaviors preserved, including `toolIntent`'s scalar coercion (`{"intent":false}` → `"false"`) and the TrimSpace / 2-key differences

## Changes

- **New**: `internal/toolargs/toolargs.go` + test — `FirstNonEmpty(get func(string) string, keys ...string) string`
- 5 call sites refactored to use `toolargs.FirstNonEmpty`
- Removed toolsummary's local `firstNonEmpty` helper (identical copy of the now-shared loop)
- Net: -17 lines (35 deletions, 18 insertions across modified files)

## Verification

- `gofmt -w` clean on all changed files
- `make vet` passes (go vet across every workspace module)
- `go test` passes for all changed packages: `agent`, `internal/apptranscript`, `internal/toolargs`, `cmd/evener-tui/internal/toolsummary`, `cmd/evener-tui/internal/msgrender`
- Targeted behavior-preservation tests pass:
  - `TestToolIntentFromArguments` (site 3)
  - `TestS3Cov_ToolIntent` / `TestS3Cov_ScalarStringAndFormatNumber` (site 2, incl. scalar coercion)
  - `TestSummarizeTool_Shell_WithIntent` / `_WithLegacyPurpose` (site 4)
  - `TestRenderToolCallShowsIntentAsFirstBodyLine` / `TestRenderToolCallShowsLegacyPurposeAsFirstBodyLine` (site 5)

## Note on `make test`

`make test` fails in this worktree due to an **environmental sandbox restriction**, not a code issue: `/bin/ps` is blocked (`Operation not permitted`), which breaks two pre-existing tests that inspect process state:
- `agent/execenv.TestStreamCommandSignalKillsWholeProcessGroup` (uses `/bin/ps` + `/proc/.../cmdline`)
- `cmd/evener-hub/frontend` `browserGuardProcessCrashpadLeak` (uses `/bin/ps`)

Neither test is related to this change (pure Go, no `execenv`/frontend files touched). The Go-only gate (`WEB=0`) and direct `go test` of all changed packages both pass.

### DIFF
--- a/agent/cov_s3_filetools_test.go
+++ b/agent/cov_s3_filetools_test.go
@@ -19,7 +19,7 @@ func TestS3Cov_FileTools_ReadWriteEdit(t *testing.T) {
 	}
 
 	// read_file returns content and tracks the read.
-	res = s3cov_exec(t, s, "read_file", `{"file_path":"note.txt","purpose":"inspect"}`)
+	res = s3cov_exec(t, s, "read_file", `{"file_path":"note.txt","intent":"inspect"}`)
 	if res.IsError || !strings.Contains(res.Output, "hello world") {
 		t.Fatalf("read_file: err=%v out=%s", res.IsError, res.Output)
 	}
--- a/agent/cov_s3_render_test.go
+++ b/agent/cov_s3_render_test.go
@@ -67,18 +67,22 @@ func TestS3Cov_ScalarStringAndFormatNumber(t *testing.T) {
 	}
 }
 
-func TestS3Cov_ToolPurpose(t *testing.T) {
+func TestS3Cov_ToolIntent(t *testing.T) {
 	t.Parallel()
-	if got := toolPurpose(json.RawMessage(`{"purpose":"explain"}`)); got != "explain" {
-		t.Fatalf("purpose = %q", got)
-	}
-	if got := toolPurpose(json.RawMessage(`{"intent":"find bug"}`)); got != "find bug" {
+	if got := toolIntent(json.RawMessage(`{"intent":"explain"}`)); got != "explain" {
 		t.Fatalf("intent = %q", got)
 	}
-	if got := toolPurpose(json.RawMessage(`{"other":"x"}`)); got != "" {
+	// "purpose" remains a supported fallback key (backward compatibility).
+	if got := toolIntent(json.RawMessage(`{"purpose":"explain"}`)); got != "explain" {
+		t.Fatalf("intent = %q", got)
+	}
+	if got := toolIntent(json.RawMessage(`{"intent":"find bug"}`)); got != "find bug" {
+		t.Fatalf("intent = %q", got)
+	}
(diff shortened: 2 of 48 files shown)


obra:
Closing: this consolidation was built on the pre-backward-compat-removal state of #600. The rename PR #600 has since been amended to remove the intent→purpose→description fallback chain entirely (per review direction), so:

1. The multi-key fallback chain this PR consolidated no longer exists — each reader is now a single-key "intent" read, and FirstNonEmpty(get, "intent") would be pure indirection.
2. The remaining differences between the 5 reader sites (type-assert vs scalarString coercion vs TrimSpace) are per-package arg-parsing idioms; unifying them would change behavior at some sites.
3. Re-adding the purpose/description fallbacks here (as this PR's diff does) contradicts the no-backward-compat direction.

The underlying review finding (duplicated fallback chain) was resolved by removing the chain rather than consolidating it.

Issue #176, opened 2026-08-18
the daemon publishes active from srv.processing while the session says idle, so Stop is refused


obra:
Found by repairing kata `gwxa`'s oracle (gwxa is closed). This is a **production** defect, not a test one, and it falsifies the headline promise of docs/superpowers/plans/2026-08-16-stop-always-works.md.

**Scope correction up front (see history):** this was filed as a turn-boundary defect. It is not. The window is *any pre-turn work, on any turn* — including, unconditionally, the whole span of the first turn before the first model round. A fix exists on a rebuild branch; see the history.

## The defect

During a turn boundary the wire publishes `status=active`, the UI shows a working session and offers Stop, and Stop is **refused**:

```
inside the boundary the wire published status=active activeTurnId=turn_m1 queueDepth=0
MEASURED DEFECT: a Stop pressed while the wire showed status=active
activeTurnId=turn_m1 was refused: appwire turn/interrupt: session is not processing
samples=2538 activeWithNoID=0 staleAcrossTheBoundary=2532 notActive=0
```

## Why — and it is not the stale id

`InterruptClientMutation` (agent/session_client_mutation.go:449) samples `s.WireState()`, which reads the **session**:

```go
func (s *Session) WireState() string {
	state := s.State()
	if state == SessionIdle && s.sessionWorkPending() {
		return string(SessionProcessing)
	}
	return string(state)
}
```

In the boundary window all three inputs are false at once: turn 1 has settled, so `State()` is `SessionIdle`; `popQueueHead` has already claimed turn 2, so `QueueDepth()` is 0 and `sessionWorkPending()` is false; `processOneInput` has not yet set `SessionProcessing`. So `WireState()` reports idle and the precondition rejects with `Conflict("session is not processing")` — while the wire's `active` comes from the **daemon's** processing flag, which spans the whole drain. The precondition's own comment claims the value it samples is "the fact the wire publishes as the thread's status". It is not. Wire says working, session says idle, Stop is refused.

## The evidence the plan relied on never entered this window

gwxa was filed because cmd/serf-hub/e2e_control_invariant_test.go logs its key counter instead of asserting it. It is worse than that: the old test's own log shows it never reached the boundary at all — 22 samples, none of which saw turn 2. The `turn_m1 -> turn_m2` transition the plan's RESULT block cites as measured was never in the sample set. So stop-always-works withdrew its Task 2 premise on evidence that could not have supported it.

## Fix

[...]

PR #608, opened 2026-08-30
fix(agent): reject stale turn interrupts via SinceTurnID click-time binding (#178)

### LINKED ISSUE #178: A delayed Stop can cancel a later turn the user never saw
Split out of kata `e519` part 2 (e519 is closed), which was not fixed by 77c446df4. That commit closed the data-loss half (a Stop eating the message queued behind it); this is the other half, and it is a distinct change.

## The defect

`turn/interrupt` now carries only `ref` and `clientMutationId`. Nothing binds the request to the moment the user clicked Stop:

- `enqueueMutation` (cmd/serf-hub/frontend/src/stores/threads.ts) persists the intent into the durable retrying outbox.
- Acceptance is decided entirely from session state at **delivery** time (`InterruptClientMutation`, agent/session_client_mutation.go): the precondition is "is the session processing right now".

So if dispatch is delayed — behind an earlier mutation in the FIFO, or until the client reconnects — the first delivery cancels whatever is running *then*. That can be a later turn the user never saw.

The targeted form could not do this: a stale `expectedTurnId` was rejected. That is the safety property the session-scoped rule traded away deliberately, and the trade is right (a Stop that refuses because the client cannot name the turn is not what the button means) — but it left nothing in its place.

## Why it is not urgent

The window needs a delayed dispatch *and* a turn boundary inside it. In the common case the outbox is empty and the request lands in milliseconds. Idempotence already prevents a *retry* of the same `clientMutationId` from cancelling twice (`TestInterruptRetryStopsNothingTwice`) — this is about the *first* delivery landing late.

**Priority note (2026-08-17):** priority 2 — a real user-visible wrong-cancellation defect (Stop kills a turn the user never saw), but it needs a delayed dispatch and a turn boundary inside the window, so it is narrower than #176's deterministic Stop-refused (P1).

## Fix direction

Bind the request to something that expires, and decide what happens when it has:

- a processing-generation token minted at click time and checked by the daemon (the daemon already has a natural generation in the turn sequence); or
- make an unaccepted session-scoped interrupt non-replayable once its processing window closes — that is, the outbox drops it rather than delivering it late.

Whichever is chosen, the user-visible answer must match: a Stop that expires should say so, not silently cancel something else.

Related: #176, katas `e519` and `t9kt` (both closed).

---
_Migrated from kata `91hp` (priority 2)._

obra:
Closes #178.

## Summary
A delayed Stop used to cancel whatever turn was running at delivery time — `turn/interrupt` acceptance was decided from session state ("is something running"), not from what the user actually saw when they clicked. This PR binds the interrupt to the turn that was active at click time and rejects it as stale if a different turn is already running:

- **Additive `sinceTurnId` on the interrupt params** (backward compatible: absent = current behavior; the web frontend, the TUI, and the protocol doc all updated, with the doc carrying full semantics via the appwiredoc template).
- **Click-time capture** in both producers: the hub frontend reads the tracked thread's active turn synchronously inside `interrupt()` (a 42-line test blocks the model's rollover behind a never-answered queued mutation and asserts the durable outbox payload still carries the click-time id), and the TUI captures `m.detail.ActiveTurnID` in the command's `Run` (both the registry command and ctrl-c paths).
- **Delivery-time guard** in `reservePrepared`'s prepare callback: a durable-namespace `turn_m<N>` SinceTurnID that differs from the session's current durable ActiveTurnID is rejected with `Conflict("stop expired: a later turn is already running")` — surfaced to the user as a recovery row with the reason, no dangling UI state, and durably terminal so retries replay the rejection rather than re-evaluating.
- **Namespace gate**: only durable `turn_m<N>` ids can be ordered against the mutation counter. The wire also publishes projector-namespace `turn_N` ids for daemon-started turns (goal continuations, notification wakes, queued drains) before the durable mint lands — a client holding such an id names the SAME logical turn the session tracks as `turn_mN`, so comparing them would falsely reject an on-time Stop. Projector-held ids fall through to session-scoped behavior exactly like an absent field; the honest tradeoff (staleness in that namespace is undetectable) is documented in the protocol doc and the code comment.

## Root cause
`sessionIsQuiesced` answered "is something running," not "is it the thing the user saw"; the interrupt carried no turn binding, so a delayed dispatch cancelled whatever was running then. Full analysis in the issue's triage comment.

## Verification highlights
- Table-driven tests drive the real `InterruptClientMutation` path deterministically: the #178 repro (stale different-id rejected), retry-of-expired (replays rejection), #176 same-generation window (applies), absent-id compat (applies), the projector-window cross-namespace case (applies), and the between-turn gap (applies session-scoped).
- Reviewers verified the durable id counter never reuses across restarts and restore (persisted `NextTurnSequence`; re-adoption of pending ids), and that no producer can mint a `turn_m`-shaped projector id (namespaces structurally disjoint).
- Gates: full agent suite, TUI/server/appwire/appprojector suites, `go vet`, gofmt, frontend vitest (threads 226/226), tsc, and byte-identical doc regeneration all PASS.

## Adversarial review
Two rounds, two competing reviewers each. Round 1: Reviewer B found a major — the guard compared the wire's projector-namespace id against the session's durable id, so the same logical turn could yield unequal ids and a false "stop expired" rejection (verified by the orchestrator in `setProcessingLocked`/`ReserveTurnID` directly). Round 2: both approve; B re-drove its exact scenario as a test (the Stop now applies and cancels the running turn) and confirmed the in-sync population keeps full protection. One comment-precision reword applied.

### DIFF
--- a/agent/session_client_mutation.go
+++ b/agent/session_client_mutation.go
@@ -528,6 +528,20 @@ func (s *Session) InterruptClientMutation(
 			rejectClientMutation(record, appwire.Conflict("session is not processing"))
 			return nil
 		}
+		// The click-time binding (issue #178): a Stop carrying the turn that
+		// was active when it was clicked, delivered only after a DIFFERENT
+		// turn is already running, refers to a turn the user saw end. Applying
+		// it would cancel a later turn they never saw, so it is rejected as
+		// expired instead -- surfaced, not silent. Inequality with a non-empty
+		// current id is the whole test: every ActiveTurnID is minted from one
+		// monotonic counter and never reused, so a different id IS a later
+		// generation. An equal id (late delivery still inside the clicked
+		// turn, the #176 window) and an absent one (old client, old durable
+		// outbox record) both fall through to today's session-scoped rule.
+		if params.SinceTurnID != "" && snapshot.ActiveTurnID != "" && snapshot.ActiveTurnID != params.SinceTurnID {
+			rejectClientMutation(record, appwire.Conflict("stop expired: a later turn is already running"))
+			return nil
+		}
 		// The fence records the turn this interrupt is actually cancelling,
 		// which is the durable name of whatever was running -- and the empty
 		// name when that turn had none. An interrupt that matches no pending
--- a/agent/session_client_mutation.go
+++ b/agent/session_client_mutation.go
@@ -533,12 +533,28 @@ func (s *Session) InterruptClientMutation(
 		// turn is already running, refers to a turn the user saw end. Applying
 		// it would cancel a later turn they never saw, so it is rejected as
 		// expired instead -- surfaced, not silent. Inequality with a non-empty
-		// current id is the whole test: every ActiveTurnID is minted from one
-		// monotonic counter and never reused, so a different id IS a later
-		// generation. An equal id (late delivery still inside the clicked
-		// turn, the #176 window) and an absent one (old client, old durable
-		// outbox record) both fall through to today's session-scoped rule.
-		if params.SinceTurnID != "" && snapshot.ActiveTurnID != "" && snapshot.ActiveTurnID != params.SinceTurnID {
+		// current id is the whole test: every durable ActiveTurnID is minted
+		// from one monotonic counter and never reused, so a different id IS a
+		// later generation. An equal id (late delivery still inside the clicked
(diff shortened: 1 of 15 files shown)


obra:
this is a bad PR and we should not merge it. The old behavior is correct.

PR #633, opened 2026-08-30
feat(doctor): doctor_evener first-party tool + shared-root delegate cache

obra:
## Summary

Implements the `doctor_evener` first-party tool and fixes the shared-root delegate-journal refold, per the study and proposal in `docs/superpowers/specs/2026-08-29-doctor-evener-first-party-tool-proposal.md` (included here).

Three coupled changes:

1. **Shared-root delegates cache (perf bug fix).** `ListSessions` called `stableDoctorDelegates` once per session, re-reading and re-folding the root's `delegates.jsonl` for every child sharing that root — measured 18 redundant folds of one 1.7MB journal on a real state root (10.66s for a 50-session sweep; a full-bucket sweep was projected at hours). Adds `delegateCache`: per-invocation, keyed by (bucket dir, resolved root session id) so `Tree`'s cross-bucket walk cannot collide; errors never cached; reads stay pure; threaded through `ListSessions` and `Tree`, with `RunAudit` inheriting via `ListSessions`. TDD: the read-count test failed RED at 9 reads for an 8-child fixture, passes GREEN at 2; benchmark documents the 19-session sweep at ~1.6ms/op.

2. **`doctor_evener` tool.** The read-only in-process equivalent of the `evener doctor` CLI, run against the session's own state root by default — removing the PATH/cwd/env failure class that made the skill's shell instruction fail with `command not found` on a real doctor run (observed in session `034DQSByTLQZnq5JktrjSn`, exit 127 on its first command). Command-option precedence mirrors the CLI (transcript: count > health > render; apilog: validate > health > filters). Read-only by library construction, pinned by a behavioral test (`TestDoctorEvener_HandlerDoesNotMutateState`, mutation-verified); output-limited with `TruncTail` plus structural row caps on `sessions`/`audit` so large sweeps truncate cleanly instead of mid-JSON.

3. **Skill and persona seam update.** SKILL.md now instructs the tool call (command table reshaped to tool arguments, with the previously-missing `turnids` and `plugins` rows); the full skill corpus (caveats, anti-patterns, all four references, persona gotchas) speaks tool dialect with the CLI kept as the labeled human-reproduction path; the doctor persona carries `doctor_evener` with an explicit tool-budget section (shell retained for the repair-guardrails validation gate only).

## Review

Two independent adversarial review sessions ran over the build (correctness lens; design/integration lens), both reading the diff and running the tests themselves. All confirmed findings are addressed in the last two commits:

- stray-selector rejection on selector-less commands (was silently ignored)
- the real read-only behavioral test + corrected claim about `ReadOnly` (parallelism hint, not enforcement) + plugins write-probe disclosure in SKILL.md
- `TruncTail` + structural row caps (was: mid-JSON truncation at 600k chars)
- `evener-doctor audit` → `evener doctor audit` in Finding evidence
- enum round-trip test (definition ↔ dispatcher), exact persona tool-set assertion, dead `doctorPluginStoreRoot` removed, `--all` exclusion documented, two wrong arg descriptions fixed

Both reviewers' verdicts were "merge/coherent with fixes"; findings accepted as-is (CLI-parity or verified-benign) are documented in the review-fix commits.

## Testing

- `go test ./agent/doctor/ ./cmd/evener-doctor/` — pass
- `go test ./agent/` (full package) — pass
- `golangci-lint` — 0 issues; `go vet` clean; gofmt clean
- New tests: shared-root read-count (RED→GREEN), benchmark, six tool-handler tests, behavioral read-only (mutation-verified), row-cap disclosure, enum round-trip, stray-selector rejection


obra:
## Simplify pass + adversarial review round 2

Two new commits on top of the review-fixes rounds:

### `ab41bbb12` — refactor(doctor): simplify pass
Adjudicated findings from the four-lens simplify review (altitude, efficiency, simplification, reuse):
- **Single-sourced the command enum** — `tool.DoctorEvenerCommands()` feeds both the schema and the dispatcher; the dual-list sync test became a single-source assertion.
- **Arg helpers deduplicated** — `doctorBoolArg`/`doctorIntArg` replaced by the existing `shellBoolArg`/`shellIntArg`; `doctorParseDuration` deleted in favor of `time.ParseDuration`.
- **Selector guard hoisted** above the dispatch switch, deleting seven per-case copies.
- **Cap disclosure moved to the right layer** — `Truncated`/`TotalRows` off the library `SessionsResult` into a tool-layer `doctorSessionsEnvelope` (the `findSessionsEnvelope` pattern); audit evidence capped at the producer (`joinSessionRefs`, cap 200).
- **Delegates cache tightened** — `getForRoot` lets callers holding the meta skip the per-session meta re-read; dead error return and nil-cache branch removed.
- **Runbook loading consolidated** into `doctor.ParseRunbookFromFS(fsys, skill, name)`, shared by the CLI and the tool; traversal sanitization included, both test seams preserved.

### `7585e21e6` — fix(doctor): review round 2 findings
Fresh adversarial pair (correctness + design) over the full 5-commit diff. All confirmed findings fixed:
- **Audit structured `sessionRefs` capped** (200) with disclosed `TotalSessionRefs` — was unbounded, could overflow the 600k-char envelope mid-JSON.
- **`Unreadable` list capped** with separate disclosure — the mass-corrupt-fleet case could overflow the envelope the Sessions cap protects.
- **Inert `format` arg dropped** from the tool surface (the JSON shape has no render format; bogus values were silently accepted).
- **`state_dir` rejected for `plugins`** and documented — was silently ignored (the plugin store lives in the config root).
- **`DoctorCommand` stays runnable** — capped refs comma-joined, no "…and N more" ellipsis inside the reproduction command; `Description` keeps the true count with the disclosed marker.
- **`delegateRootFromMeta`** extracted — one encoding of journal ownership shared by `sessionRow` and `resolveDelegateRoot`.
- **New tests**: `TestDoctorEvener_EveryCommandDispatches` (enum↔switch drift), runbook seam substitution + traversal rejection, `audit` added to the read-only mutation regression.

Both reviewers explicitly verified clean: cache keying, error non-caching, read-purity, cap boundaries, traversal rejection, selector guards, arg-helper semantics, runbook-loader equivalence, `state_dir` root semantics.

Gates: `go vet`, `gofmt`, `golangci-lint` (0 issues), full `go test ./agent/`, `agent/doctor`, `cmd/evener-doctor` — all green.

PR #649, opened 2026-08-30
fix(agent): make tool schemas more robust across 5 tools

obra:
## Summary

Five robustness improvements to tool schemas and handlers, making the model-facing surface less error-prone.

### 1. delegate sandbox: oneOf → combined enum
Replaced the two-field `sandbox` (string enum) + `sandbox_net` (boolean) with a `oneOf` constraint with a **single combined `sandbox` enum** encoding both mode and network:
- `"read-only"` — inherit parent network
- `"read-only+nonet"` — disable network
- `"nonet"` — inherit parent mode, disable network (non-off parents only)
- `"off"` — no confinement

The model can now express any valid combo without hitting a `oneOf` violation. `decodeDelegateArgs` splits the `+nonet` suffix into mode + `sandbox_net=false` for the handler.

**Files:** `definitions.go`, `sandbox_delegate.go`, `session_tools_jobs.go`, `delegate_runtime.go`

### 2. task_list update: status optional
Schema `required` changed from `["id", "status"]` to `["id"]`. The handler already accepted id+notes without status; the schema now matches.

**Files:** `definitions.go`, `explain_test.go`

### 3. job_watch: accept `job:` prefix
`normalizeWatchSource` strips an optional `job:` prefix so both `job:job_123` and `job_123` resolve to the same concrete job.

**Files:** `job_watch.go`

### 4. read_transcript: accept format=markdown + offset_bytes
`validateJobReadArgs` now accepts `format=markdown` (the default view) alongside `offset_bytes`/`output_match` on `job:` refs. Only non-markdown formats are still rejected with paging/search.

**Files:** `session_tools_transcript.go`

### 5. manage_worktree list: 0-commit lane not labeled "merged"
The list operation now skips the `Merged` check when `tip == baseSHA` (0-commit lane). `isAncestor(base, main)` is trivially true since the branch was cut from main, but that is the absence of work, not evidence of a merge. The lane now correctly reads "0 ahead" instead of the misleading "merged".

**Files:** `session_tools_worktree.go`

## Related Issue

#655 — Stopping an observer sidecar delegate does not clear its `job_watch` on the parent session, causing the stopped delegate to fire indefinitely on every parent `communicate` event. Observed live while developing these changes.

## Test Plan

- [x] All new/updated tests pass: `TestNormalizeWatchSource`, `TestValidateJobReadArgs`, `TestWorktreeListSummaryZeroAheadNotMerged`, `TestDelegateSchema*`, `TestExecTool_*Delegate*`, `TestPrepareToolCall_Delegate*`, `TestDelegateSurfaceUsesAgentRegistryCapabilities`
- [x] `go build ./...` clean
- [x] Full delegate/sandbox test suite passes
- [x] golangci-lint 0 issues; `make vet`, `lint-gofmt`, `lint-naming`, `lint-generated` pass; race detector clean

### DIFF
--- a/agent/delegate_runtime.go
+++ b/agent/delegate_runtime.go
@@ -1163,8 +1163,8 @@ func (runtime delegateRuntime) create(ctx context.Context, args delegateArgs) de
 	explicitSandbox := strings.TrimSpace(args.Sandbox) != "" || args.SandboxNet != nil
 	if explicitSandbox && args.SandboxNet != nil && strings.TrimSpace(args.Sandbox) == "" && !delegateSandboxBackendAvailable(s.sandboxHostFacts()) {
 		return delegateStartFailed(newDelegateSandboxRequestError(
-			errors.New("invalid_request: sandbox_net cannot be enforced on this host because no sandbox backend is available; omit sandbox_net"),
-			"sandbox_net",
+			errors.New("invalid_request: sandbox cannot be enforced on this host because no sandbox backend is available; omit the sandbox parameter"),
+			"sandbox",
 		))
 	}
 	if readOnlyScope {
--- a/agent/internal/tool/definitions.go
+++ b/agent/internal/tool/definitions.go
@@ -97,24 +97,20 @@ func DefShell() llm.ToolDefinition {
 	}
 }
 
-// DelegateSandboxSchema describes the sandbox controls that the current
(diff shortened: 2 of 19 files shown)


obra:
the failure is an unrelated flake being addressed elsewhere

Issue #640, opened 2026-08-30
Make TestServeWebSocketPingsIdleReaderWhileBusyRPCHandlerRuns deterministic under -race


obra:
## What

`TestServeWebSocketPingsIdleReaderWhileBusyRPCHandlerRuns` in `internal/appserver/websocket_test.go` (line 48) is a timing-sensitive test that flakes under the race detector. It failed the `race-root` CI job on PR #582 (run 33298875440) and is the single known source of non-deterministic CI failures across the recent delegate lifecycle work.

This issue tracks making the test deterministic so it stops intermittently blocking unrelated PRs.

## The failure

```
--- FAIL: TestServeWebSocketPingsIdleReaderWhileBusyRPCHandlerRuns (0.01s)
    websocket_test.go:112: keepalive did not attempt a ping while the receive loop was free and the RPC handler was busy
```

The assertion at line 112 fires when the `decision` channel — which the keepalive goroutine writes to after consulting its reader-availability gate — does not arrive within a 1-second `time.After` deadline (lines 109–119). Under `-race`, goroutine scheduling slows and shifts; the decision write can lose the race with the deadline even though the keepalive logic is correct. No `DATA RACE` is reported — the detector finds no unsynchronized access. This is a test-determinism problem, not a product defect.

## Evidence it is pre-existing

- The test was introduced by commit `0291be074` ("fix(appserver): dispatch requests concurrently per connection"), which is already on `origin/main` (46 commits back from the #582 merge). It is not a product of the delegate lifecycle work.
- The same PR's non-race `tests` job passes, `race-modules` passes, and the root module passes in local full runs without `-race`. The flake is specific to the race detector's scheduling overhead.
- The test's own comments (lines 103–107, 129–134) walk through scheduling-ordering caveats and acknowledge that "the decision is emitted only after the gate has been consulted, so release is not time-based" — yet the assertion itself *is* time-based (1s deadline).

## Proposed fix

Replace the wall-clock deadline with a condition-based wait on the gate state. The keepalive decision is already emitted to a channel; the test should wait for that channel signal (or a subsequent idle-ping observation) deterministically rather than bounding it by `time.After`. The repo's testing docs (`docs/developing-evener/testing.md`) call this out directly: deterministic tests use barriers and condition waits, not sleeps or deadlines that racing goroutines can miss.

[...]

PR #654, opened 2026-08-30
fix(tui-test): pin spawn-form model default race in APIErrorsRenderInPlace e2e

obra:
## Summary

Follow-up to #633. PR #633 merged (squash, e892c8eeb) at 11:46 PDT, but the flake fix for `TestTUITmuxE2E_APIErrorsRenderInPlace` was committed 12 minutes later — it landed on the orphaned `doctor-tool-build` branch and never reached main. This PR carries that fix onto main.

## Root cause

A startup race in the test, not ambient config. Pressing `n` clears the spawn form's model field (`resetSpawnForm`), and the default model only repopulates when `hubSpawnOptionsMsg` arrives — an async WebSocket round-trip to the hub's `ModelList`. The test submitted the form after waiting only for `"Prompt (optional):"`. On a loaded Linux CI runner, Enter could be consumed before the model default landed, so `spawnModel` was empty and the TUI rendered `"choose a model before starting"` instead of the scripted spawn failure (CI run 33304123980 on the #633 head). Local macOS wins the race in milliseconds, which is why it passed locally.

Proven both directions with a temporary 1.5s `ModelList` delay on the fake hub: the unfixed test failed with CI's exact output; the fixed test passes under the same delay. All instrumentation removed; no production code changed.

## Fix

One line + comment: the test's `WaitFor` now also requires `"Model:    openai/gpt-5"` before submitting, pinning the same happens-before edge the sibling harness-cycling spawn test already pins (tmux_e2e_test.go:375). No timeout bumps, retries, or skips.

## Verification

- `-count=20`: 20/20 pass; `-count=5` re-verified on this branch (5/5)
- Scrubbed env (`env -i`, clean HOME/XDG): pass
- Whole package `go test ./cmd/evener-tui/ -count=1`: green
- `go vet ./cmd/evener-tui/...`, `golangci-lint run ./cmd/evener-tui/...`: clean

## Known follow-up (out of scope)

The dashboard/project spawn test (tmux_e2e_test.go:121) has the same latent race — waits for `Dir:`/`Prompt:` but not `Model:`. A one-line `WaitFor` addition would pin it too.

### DIFF
--- a/cmd/evener-tui/tmux_e2e_test.go
+++ b/cmd/evener-tui/tmux_e2e_test.go
@@ -1005,7 +1005,13 @@ func TestTUITmuxE2E_APIErrorsRenderInPlace(t *testing.T) {
 	app.WaitFor("EVENER LIVE", "live task")
 	hub.SetFailSpawn(true)
 	app.SendKeys("n")
-	app.WaitFor("evener / new session", "Prompt (optional):")
+	// The form's model default arrives asynchronously (fetchHubSpawnOptions
+	// round-trips to the hub); submitting before hubSpawnOptionsMsg lands sees
+	// an empty spawnModel and renders "choose a model before starting" instead
+	// of the scripted spawn failure — a startup race a loaded CI runner loses.
+	// Waiting for the populated Model field pins the happens-before edge the
+	// same way the harness-cycling spawn test does.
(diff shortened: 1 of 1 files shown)


obra:
the failure is an unrelated flake tracked in #640

PR #662, opened 2026-08-30
fix(tui-test): pin spawn-form model default race in DashboardProjectAndSpawn e2e

### LINKED ISSUE #656: tmux e2e: dashboard/project spawn test has the same spawn-form model-default race fixed in APIErrorsRenderInPlace
## Summary

While fixing the CI flake in `TestTUITmuxE2E_APIErrorsRenderInPlace` (PR #654), investigation found the same latent race in the first spawn segment of the dashboard/project spawn tmux e2e test (`cmd/evener-tui/tmux_e2e_test.go`, the test starting around line 120).

## The race (identical mechanism to #654)

Pressing `n` opens the spawn form and `resetSpawnForm` clears `spawnModel`; the default model only repopulates when `hubSpawnOptionsMsg` arrives — an async WebSocket round-trip to the hub's `ModelList` (via `fetchHubSpawnOptions` → `hub_update.go` → `syncSpawnModelWithHarness` → `firstEnabledModel`). If the test submits the form before that message lands, `spawnModel` is empty and `submitSpawnForm` (hub_spawn.go) renders **"choose a model before starting"** instead of proceeding.

The dashboard/project spawn test's first segment waits for `Dir:`/`Prompt:` but **not** for the populated `Model:` field, so its keystroke sequence only lowers the probability of the race — on a loaded CI runner it can strike exactly like the APIErrors test did in CI run 33304123980.

(The second segment is implicitly synchronized: `"Plugins: 2/2 enabled"` renders only after `hubSpawnOptionsMsg` processing requests the plugin preview.)

## Proposed fix

One line, matching the pattern now pinned in both `TestTUITmuxE2E_APIErrorsRenderInPlace` (line ~1014) and `TestTUITmuxE2E_HarnessModelPicker` (line ~375): add `"Model:    openai/gpt-5"` to the first segment's `WaitFor` before typing the prompt. No timeout bumps, retries, or skips — pin the happens-before edge.

## Reproduction (from the #654 investigation)

Add a temporary 1.5s delay to the fake hub's `handleModelList` and run the unfixed test — it fails with the same "choose a model before starting" output. (Instrumentation was removed after the #654 verification; nothing remains in the tree.)

Filed from the #654 root-cause investigation, where the fix was deliberately left out to keep that PR minimal and test-scoped.

obra:
## Summary

Fixes #656.

The first spawn segment of `TestTUITmuxE2E_DashboardProjectAndSpawn` had the same latent startup race that flaked `TestTUITmuxE2E_APIErrorsRenderInPlace` in CI (fixed in #654): it waited for `Dir:`/`Prompt:` but not the populated `Model:` field. Pressing `n` clears `spawnModel`, and the default only repopulates when `hubSpawnOptionsMsg` arrives — an async WebSocket round-trip to the hub's `ModelList`. Submitting before that lands renders "choose a model before starting" instead of proceeding. The test's keystroke sequence only lowered the race's probability on a loaded runner.

## Fix

One line + comment: the first segment's `WaitFor` now also requires `"Model:    openai/gpt-5"`, pinning the same happens-before edge as `TestTUITmuxE2E_APIErrorsRenderInPlace` and `TestTUITmuxE2E_HarnessModelPicker` (line ~375). The second spawn segment is implicitly synchronized via `"Plugins: 2/2 enabled"` and is unchanged.

No timeout bumps, retries, or skips. Test-only change; no production code touched.

## Verification

- `-count=10`: 10/10 pass; `-count=5` re-verified (5/5)
- `go vet ./cmd/evener-tui/...`: clean
- `golangci-lint run ./cmd/evener-tui/...`: 0 issues
- `gofmt -l`: clean

Reproduction method (from the #654 investigation): a temporary 1.5s delay on the fake hub's `handleModelList` makes the unfixed test fail with the same "choose a model before starting" output. No instrumentation remains in the tree.

### DIFF
--- a/cmd/evener-tui/tmux_e2e_test.go
+++ b/cmd/evener-tui/tmux_e2e_test.go
@@ -118,7 +118,14 @@ func TestTUITmuxE2E_DashboardProjectAndSpawn(t *testing.T) {
 	app.WaitFor("EVENER LIVE", "live task", "ended maintenance")
 
 	app.SendKeys("n")
-	app.WaitFor("evener / new session", "Dir:      "+tuiE2EProjectDir, "Prompt (optional):")
+	// The form's model default arrives asynchronously (fetchHubSpawnOptions
+	// round-trips to the hub); submitting before hubSpawnOptionsMsg lands sees
+	// an empty spawnModel and renders "choose a model before starting" — the
+	// startup race that flaked TestTUITmuxE2E_APIErrorsRenderInPlace in CI
+	// (issue #656). Waiting for the populated Model field pins the
+	// happens-before edge the way that test and the harness-cycling spawn
+	// test do.
+	app.WaitFor("evener / new session", "Dir:      "+tuiE2EProjectDir, "Prompt (optional):", "Model:    openai/gpt-5")
 	app.SendKeys("Tab", "Tab", "Tab", "C-u")
 	app.TypeText("/tmp/evener-e2e/custom")
 	app.WaitFor("Dir:      /tmp/evener-e2e/custom")


obra:
failure is an unrelated flake tracked in 640

PR #676, opened 2026-08-30
fix(appserver): honor a mid-capture unsubscribe when the hydration commits

obra:
Fixes the unsubscribe flake failing PR #649's `race-root` CI check.

## What failed

```
--- FAIL: TestServerAppWireThreadUnsubscribeDropsSubscription (0.01s)
    appwire_server_test.go:2087: subscriber count after unsubscribe = 1, want 0
```

## Whose change it was

This test and the bug beneath it came from #651 (merged), not from #649's
changes — #649 branched from main after #651 landed and inherited the test.
The flake is timing-dependent, which is why #649's two earlier CI runs passed
and the race-detector build (`race-root`, different scheduling) caught it.

## Root cause

A subscribed `thread/read` registers a **buffering** capture generation; the
entry goes live only when the response's finalizer commits (after the response
enters the connection's send queue). An unsubscribe arriving inside that
window — the client got the read response and sent its unsubscribe before the
handler goroutine reached `commit()` — was recorded as withdrawn but left the
buffering entry in place, and `Release` then **cleared the withdrawn record and
kept the entry live**. The unsubscribe was silently dropped:

- `SubscriberCount` stays 1 (what the flaky test caught), so the hub relay's
  idle teardown never fires;
- the connection keeps receiving that thread's notifications after a
  *successful* unsubscribe.

Reproduced deterministically at the registry level: `beginBuffered` →
`Unsubscribe` → `Release` leaves the entry subscribed.

## Fix

`Subscriptions.Release` now honors a recorded drop at commit: it removes the
entry and returns no buffered records. Abort already honored drops (restore
the displaced snapshot minus the thread); commit now matches. The
`Unsubscribe` doc said "commit keeps it live" — the abort half was right, the
commit half was the bug.

Also corrects one existing test that had pinned the wrong behavior ("committed
capture should leave the entry live" — that assertion *was* the dropped
unsubscribe) and adds a red-probed regression test,
`TestSubscriptionsUnsubscribeDuringCaptureThenCommitDropsEntry`.

## Verification

- Registry-level repro: fails pre-fix, passes post-fix.
- `TestServerAppWireThreadUnsubscribe*` ×30 under `-race`: green.
- `go test -race` on `internal/appserver`, `server`, `cmd/evener-hub`: green.
- The originally-flaky websocket test is unchanged — with the semantics
  fixed, both orderings of the race are now correct.

After this merges, #649 can rebase (or simply re-run CI once main picks this
up) and `race-root` should go back to green.

### DIFF
--- a/internal/appserver/subscriptions.go
+++ b/internal/appserver/subscriptions.go
@@ -50,12 +50,12 @@ func (s *Subscriptions) Subscribe(connID, threadID string) {
 // so a client racing its own re-subscribe can never wedge the registry.
 //
 // A thread currently held by a BUFFERING capture generation records the drop
-// and leaves the entry in place: the capture displaced the connection's
-// previous subscriptions into its rollback snapshot, and removing the
-// buffering entry here would strand that snapshot (withdrawBuffered matches
-// on the live generation and would bail without restoring). The generation's
-// own commit/abort resolves the entry — commit keeps it live, abort restores
-// the snapshot minus this thread.
+// and leaves the entry in place until the generation resolves: removing the
+// buffering entry here would strand the capture's rollback snapshot
+// (withdrawBuffered matches on the live generation and would bail without
+// restoring). The resolution then honors the drop — abort restores the
+// snapshot minus this thread, and commit removes the entry instead of
+// reviving it.
 func (s *Subscriptions) Unsubscribe(connID, threadID string) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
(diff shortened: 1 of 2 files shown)


obra:
Superseded by #675 (merged), which independently landed the same root-cause fix — Release honoring the mid-capture withdrawal at commit — plus the withdrawn-on-entry refactor and matching live-interest fixes for IsSubscribed/Threads/Route. The one piece of unique value here, the non-replace commit-shape regression test, was extracted and adapted to the merged implementation in #678 (with credit in the commit message). This PR's attribution note stands: #649's red race-root check should pass on a re-run now that main has the fix.

PR #735, opened 2026-08-31
fix(tui): require completeness anchors in CaptureStable, not bare stability

obra:
## Summary

Fixes the `TestTUITmuxE2E_CaptureStableDuringStream` flake seen in CI run 33424655285 (also linked from #722's checks, but **not caused by #722** — main's recent runs fail the same step with different timing-sensitive tests, and that PR touches only hubcore PastIndex).

The failure returned a frame with no session breadcrumb and no composer hints — all 50 rows streamed text — and ``CaptureStable`` accepted it because two consecutive captures were byte-identical.

## Root cause

Stability is not completeness. `CaptureStable` returned the first capture pair that stopped changing, which under CPU contention proves only that tmux's grid paused between two samples 10ms apart:

- bubbletea writes each frame as one unsynchronized multi-KB ANSI stream, composer/footer last
- under load, that write can stall mid-frame; tmux's grid then shows rows above the composer not yet redrawn
- the failing run's dump was exactly that torn grid — frozen long enough for two samples to agree on it

## What changed

- `CaptureStable(wants ...string)` now takes completeness anchors and returns the first capture containing all of them — evidence the frame write finished (footer last-written) and the redraw before it did too
- The stream test passes its two existing assertions (breadcrumb + composer) as those anchors, so the property it always claimed to check — a complete frame, top and bottom — is now the acceptance condition instead of a post-hoc check on whatever frame happened to stop moving
- The two-capture stability proxy is gone; the contract is documented as completeness, not stability

## Verification

- `go test -run TestTUITmuxE2E_CaptureStableDuringStream -count=5` — 5/5 pass (was flaky)
- Full package suite `go test ./cmd/evener-tui/` — pass, 24 tmux E2E tests, 0 skips
- `gofmt`, `go vet` clean

## CI expectation

The failing CI step is `ROOT_FULL=1 make test`, which runs the root module's full suite without `-short` — the exact tests exercised above.

### DIFF
--- a/cmd/evener-tui/tmux_e2e_test.go
+++ b/cmd/evener-tui/tmux_e2e_test.go
@@ -1024,10 +1024,18 @@ func TestTUITmuxE2E_APIErrorsRenderInPlace(t *testing.T) {
 // TestTUITmuxE2E_CaptureStableDuringStream exercises CaptureStable under the
 // exact condition kata nxq6 reported the pane going blank above the composer:
 // a rapid burst of hub notifications re-rendering the pane in a tight loop,
-// no keypresses. Every CaptureStable() result taken during the burst must be
-// a complete frame — the session breadcrumb from the top of the pane and the
-// composer's key hints from the bottom, never one without the other — which
-// is the property a lone Capture() cannot promise.
+// no keypresses. Every capture taken during the burst must be a complete
+// frame — the session breadcrumb from the top of the pane and the composer's
+// key hints from the bottom, never one without the other — which is the
+// property a lone Capture() cannot promise.
+//
+// Stability alone cannot promise it either. Under CI load this test flaked as
+// a "stable" torn frame: the pane never settles during the flood, and two
+// consecutive captures agreed on a mid-write grid — 50 rows of streamed text,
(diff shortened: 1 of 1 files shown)


obra:
Both findings addressed in `11c0bc729`:

**Medium — stability check restored alongside the anchors.** You're right that anchors alone were weaker than what the name promised: tmux's grid is incremental, so a partial repaint can hold both anchors from the previous frame while the middle is rewritten, and with no wants the completeness check was vacuous. `CaptureStable` now requires **both** — every want present **and** byte-identity with the capture one poll interval before. The two heuristics cover each other's failure modes (the original CI flake was stable-but-torn; the anchors-alone hole is complete-anchors-but-still-changing). The doc now states the dual contract and names the residual gap honestly — a stalled mid-write whose extremes carry stale anchors for every want at once — as the one case nothing short of synchronized-output mode closes. Verified: stream test 5/5, full evener-tui suite green.

**Low — stale helper name.** The `CaptureCompleteStable` reference in the test comment was a mid-edit rename leftover; it now names `CaptureStable`'s anchor arguments.

PR #736, opened 2026-08-31
job_list typed-decode filters; job_watch trigger-field errors

obra:
Follow-up to #731 (two fixes developed in parallel with its merge; rebased onto main).

## 1. Typed-decode fixes for job_list filters (6d39d8c22)

Same bug class as the task_list "<nil>" status fix in #731: jobStatusArrayArg and jobTypeArrayArg coerced each array element with fmt.Sprint before the enum switch, so a wrong-typed element (number, bool, null, object) surfaced as enum noise like 'invalid job status "<nil>"'. Both now assert value.(string) and return '<key> values must be strings', matching the file's existing stringArrayArg pattern. Wrong-typed elements pinned by tests for both parsers.

## 2. job_watch trigger-mode error messages (dbe5f8ab0)

Investigation findings (each confirmed empirically):
- Two trigger modes on create: rejected with good messages on session sources; intentionally accepted as composition on concrete job sources (pinned by TestTerminalCatchupRejectsEventsCondition, documented as a separate progress trigger) — deliberately unchanged.
- Trigger fields on list/inspect/clear: were silently ignored (parsed, then discarded) — a real silent-ignore bug. Now rejected naming the supplied fields, the rule, and both fixes; watch-all on create is preserved.
- every/event_filter without events: were rejected with terse messages; now actionable with examples.

Tool schema untouched — error-message work only.

## Verification

- go test ./agent/ -count=1 — ok
- go test -tags evenerfuzz -run '^Fuzz' -count=1 . ./agent/... — all 41 packages ok
- make vet, make lint (9/9 sub-gates), gofmt -l agent/ empty

Both fixes were built TDD by subagents, reviewed and independently re-verified here before merge.

### DIFF
--- a/agent/job_watch.go
+++ b/agent/job_watch.go
@@ -770,7 +770,10 @@ func validateWatchEventArgs(a watchArgs) error {
 	}
 	if a.Every > 0 {
 		if len(a.Events) != 1 {
-			return errors.New("invalid_request: every requires exactly one watched event kind")
+			if len(a.Events) == 0 {
+				return errors.New(`invalid_request: every requires events naming exactly one kind (e.g. events ["communicate"], every 3); every with no events has nothing to fire on`)
+			}
+			return errors.New(`invalid_request: every requires events naming exactly one kind, not several (e.g. events ["communicate"], every 3)`)
 		}
 		if a.Events[0] == "*" {
 			return errors.New(`invalid_request: every requires a single concrete event kind, not "*"`)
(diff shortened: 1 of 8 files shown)


obra:
Addressed in b351489f0: `rejectWatchTriggerFieldsOnNonCreate` now skips `nil`-valued keys when collecting supplied fields, matching how `shellIntArg` decodes null as omitted everywhere else. Added the requested regression test (`TestWatchArgsFromToolArgsTreatsNullTriggerFieldsAsOmitted`: null `progress_interval_ms`, null `every`, and both on inspect — each asserting acceptance and zero decode). The original rejection tests still pass, so non-null misuse is still caught. Verified: agent suite, vet, gofmt.

Issue #746, opened 2026-09-01
Resume sidecar's 1MiB limit turns delivery-heavy transcripts into permanent per-resume full scans


obra:
## Summary
A transcript whose fold snapshot exceeds the resume sidecar's 1MiB read limit (agent/transcript/sidecar.go:214-216 refuses the write; ReadSidecar caps the read at resumeSidecarLimit) can never hold a sidecar — every resume re-scans the whole file, and serve's compaction-anchored fallback additionally re-scans it a second time for the refresh whose sidecar write is then refused. Measured: one SidecarDeliveryCommit row serializes to ~50 bytes, so ~21k committed delegate deliveries cross the limit. The refresh then repeats a full scan (with the fold's per-entry work) per compaction-anchored resume, discarding the result.

## Evidence
From the round-2 adversarial review of the #643 branch (mutation- and measurement-verified): at 40k delivery commits the refresh takes 8.9s (the fold's conflict scan was quadratic — fixed separately on the branch by a reverse index) and ReadSidecar reports the sidecar absent, so the cycle repeats forever. The fold itself refuses nothing here — the file is healthy; only the serialized snapshot size crosses the limit.

## Fix direction
Either bound what the snapshot serializes (delivery commits are a map of two short strings — the natural candidates for a compact encoding or a cap-with-marker), or record a durable "size-refused" marker so the refresh knows to stop retrying, or accept the fallback and skip the refresh when the previous sidecar's refusal is detectable. The quadratic-scan half is already fixed on the #643 branch; this issue tracks the size-refusal cycle only.

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

**Verdict:** `cannot-verify` — the code this issue describes is not in the pinned revision. `agent/transcript/sidecar.go`, `resumeSidecarLimit`, `ReadSidecar` and the fold snapshots exist only on the unmerged branch `origin/fix-643-validated-offset-sidecar` (commits `8013543bb4`, `61231f64e1`); `git cat-file -e HEAD:agent/transcript/sidecar.go` fails and `git log --all -S'resumeSidecarLimit'` finds only those two branch commits. With no resume sidecar in main, the 1 MiB write-refusal / per-resume full-scan cycle cannot occur here.

**Recommendation:** `leave-open` — this is a follow-up to unmerged PR #643; re-verify only after that branch (or its replacement) lands on main. Closing it now would drop tracking of a real defect the branch does contain.

Issue #743, opened 2026-09-01
Serve's compaction-anchored fallback decodes the transcript twice (identity projection + sidecar refresh)


obra:
## Summary
Serve's compaction-anchored fallback pays TWO independent full transcript decodes back to back: `prepareAppIdentity` (the file-form identity projection, server/appwire_runtime.go → apptranscript.TurnsFromFile) and then `transcript.RefreshSidecarFromFullScan`, each opening the file and strict-decoding every line on its own. At a 183MB transcript that is ~1.2s of decode and 2× ~250MB transient heap (sequential, so peak stays ~250MB) — once per compaction gap, which the refresh exists to bound. The work is correct; it is simply paid twice for one read's worth of information.

## Evidence
- cmd/evener/serve.go's fallback branch: the projection runs first, the refresh second, both over the same file with no shared state — `server.PrepareAppIdentityForRef` returns only a projection with no scan-result surface, and `agent/transcript` cannot consume the server package's read without a new cross-package API (server imports agent; agent cannot import server).
- Measured by the efficiency review of the #643 fix wave: full scan 614ms, refresh ~600ms warm, windowed resume 10.4ms — the doubled read is bounded (once per compaction) but strictly redundant.

## Fix direction
A `RefreshSidecarFromScan` variant accepting an already-computed `scanResult` (the projection's read, if the server package surfaces one), or moving the refresh into the one reader that already holds the decoded entries. Judgment from the same review: the current shape is the simplest honest one given package boundaries, so this is an optimization, not a defect — worth taking only if a natural seam appears.

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

**Verdict:** `cannot-verify` — the second decode named here does not exist at this revision. `RefreshSidecarFromFullScan` and `SnapshotsComplete` are absent from the pinned tree (`rg` finds none), and `agent/session_sidecar.go` is not on `origin/main` (`git ls-tree` empty); HEAD equals `origin/main` (`27503c07dacc`). Those symbols live only on the unmerged `origin/fix-643-validated-offset-sidecar`. At HEAD serve calls `prepareAppIdentityFromEntries` when restore already decoded entries (`cmd/evener/serve.go:678-680`), so the described projection + refresh double read is not present.

**Autonomy:** `n/a` — depends on unmerged branch code.

**Recommendation:** `leave-open` — re-check against main after #643 merges.

Issue #742, opened 2026-09-01
Compaction-anchored sidecar forces a per-resume full transcript re-read on non-serve restore consumers


obra:
## Summary
A resumed root session whose sidecar is compaction-anchored (SnapshotsComplete=false) takes `rearmRootDelegateAttentionFromTranscript`'s full-file-read fallback on EVERY restore (agent/session_attention.go:937-941) — the seeded rearm cannot fold over an incomplete snapshot, so each resume re-decodes the whole transcript just to rebuild the wake cache. Serve's identity-fallback refresh (fix-643 branch, `RefreshSidecarFromFullScan` after the projection's error gate) bounds the equivalent cost for the serve path only; every non-serve restore consumer — `run.go --resume`, hub-spawned daemons, delegate runtime restores — keeps paying the full read per resume for as long as the sidecar stays compaction-anchored.

## Evidence
- Reached from `RestoreSessionFromMetaWithConfig` for any root delegate-attention receiver whose validated sidecar carries SnapshotsComplete=false (the compaction anchor always writes that shape at agent/session_sidecar.go:139-148).
- Measured on this machine (warm cache, 183MB / 84k-entry real transcript): one full decode ~614ms plus ~250MB retained heap, per resume, indefinitely.
- The serve-side refresh (commit 4265c3b26 on the #643 branch) proves the repair shape: one extra full scan converts the sidecar, after which every resume windows. The same conversion applied at the rearm fallback — or upstream at restore time — would cover all consumers.

## Fix direction
Run the same opportunistic refresh where the rearm fallback already pays a full read (or promote it to the restore path's post-refresh point, where `refreshFromDisk` already re-reads the file), so the compaction-anchored shape self-heals on the first restore regardless of which consumer resumed the session.

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

[...]

PR #757, opened 2026-09-01
Resume sidecar: window transcript resume to the last checkpoint (validated offset)

### LINKED ISSUE #643: Resume still decodes the full transcript once: validated-offset sidecar for the post-compaction suffix (Phase 2)
## Summary

Resume still performs ONE full strict decode of the whole append-only transcript (restore's `OpenWriterForSession` scan) before the daemon is ready. Phase 1 eliminated four of five decode passes (see the linked PR), but the remaining pass is O(file size), not O(live history): on a real 184MB / ~84.6k-entry transcript the live post-compaction history is only ~1.1MB / ~582 entries (~0.6%), yet restore decodes every line. At ~1.2s per pass warm this is the dominant remaining resume cost for very large sessions.

## Evidence

- Read #1 (remaining): cmd/evener/serve.go:486 → `RestoreSessionFromMetaWithConfig` (agent/session_init.go:578) → `OpenWriterForSession` → `resumeWriter` (agent/transcript/transcript.go:572-649) strict-decodes every line and retains all entries.
- `ResumeHistory` (agent/transcript_read.go:146-174) only uses `[last CHECKPOINT, …]`: last checkpoint in the real 184MB file sits at line 84,065 of ~84.6k — the suffix is what matters.
- Coverage census on this machine: of 2,428 transcripts, only 70 contain any CHECKPOINT/SUMMARY; of 465 transcripts ≥1MB, only 69 have compaction — 396 (85%) would get zero benefit from a compaction-anchored-only optimization.

## Design constraints an implementation must honor (from adversarial review)

A suffix-only read is not just an optimization — naive versions are *incorrect*:

- **Delegate-attention fold**: `foldDelegateAttention` hard-fails on "resolved before it was appended" (agent/session_attention.go:135-137) when an attention straddles the boundary; pending attentions opened pre-boundary would be silently dropped. Compaction does NOT carry TurnSteering/TurnAttentionResolution into the suffix (agent/internal/contextmgr/context_manager.go:764-770 builds a fresh text-only turn).
- **TrackFailures seed**: the failure counter needs entries from the fork divergence ordinal forward (agent/transcript/failures.go:100-117); those live in the prefix. Without a floor, the seed sheds what compaction summarized away (the count exists precisely because history re-derivation would do that).
- **maxSeq floor**: resumeWriter derives `maxSeq` from all entries so resumed writes never collide (agent/transcript/transcript.go:635-648).
- **Header validation** must still run (session identity check at transcript.go:603-606).
- **Fork/aside children** get a new file with replayed prefixes and child-local seqs — a per-file sidecar must be bound to file identity, and fork children may simply fall back to the full scan.

## Suggested design (adversarially reviewed)

A durable **validated-offset sidecar**, written at three anchor points — (1) compaction, (2) clean shutdown, (3) opportunistically after any full-scan fallback (so the *second* resume of a never-compacted session is also fast). Copy the anchor/identity validation machinery from internal/apptranscript/turn_index.go (FirstAnchor/TailAnchor, PrefixStamp, FileIdentity) rather than inventing weaker validation. The sidecar must snapshot, at anchor time: pending-attention IDs, delivery commits, and the failure-count floor. `maxSeq = max(sidecar floor, suffix max)`. Every mismatch (size, identity, boundary record, offset > size) must fall back to the full scan — never error.

Clean-shutdown alone is insufficient (daemon sessions live for weeks); compaction alone covers 15% of large sessions; all three together cover nearly everything.

## Test pointers

TestOpenWriterFallsBackToFullScanWhenSidecarMissing / …WhenSidecarOffsetExceedsFileSize / …WhenSidecarFileIdentityMismatches; TestSidecarSeqFloorPreventsSeqCollision; TestRestoreSessionAcceptsCorruptPrefixLineWithSidecarAndFullReaderStillFails (pins the deliberate wf7e posture change — see the comment block at agent/transcript/transcript.go:138-158, which says such a change must be a decision, not an accident); TestTrackFailuresSeedWithSidecarIncludesPreCheckpointFailures; TestDelegateAttentionFoldResolutionAfterBoundaryDoesNotError; TestPendingDelegateAttentionOpenedBeforeBoundarySurvivesResume; TestForkChildResumeFallsBackToFullScanWithoutSidecar.

obra:
## Summary

Closes #643. A resumed session's wall-clock time-to-ready was dominated by full decodes of an unbounded append-only transcript, even though compaction means only a tiny suffix is live history (measured on a real 183MB / 84k-entry transcript: the last checkpoint sits at line 84,065 of 84,146 — 0.4% of the bytes are live; resume decoded all of them, twice).

This adds a **validated-offset resume sidecar** — `<transcript>.resume-sidecar.json` — recording the byte offset of the last checkpoint entry. A windowed resume validates it (session id, bounds, file identity via dev/inode, sha256 anchors over the first and pre-boundary byte windows, boundary seq) and decodes only `[checkpoint, ...rest]`: exactly ResumeHistory's window. **Windowed resume measured 10.4ms vs a 614ms full scan on a 183MB fixture (~59x).**

## What changed

- **Two anchors write the sidecar.** Compaction (the writer records the checkpoint entry's byte-start plus prefix facts as it appends; `PrefixTurnCount=-1`, snapshots incomplete — it cannot vouch for what it never decoded) and opportunistic post-full-scan (every entry decoded; complete snapshots, exact prefix-turn count).
- **Serve's compaction-anchored fallback refreshes the sidecar** after the identity projection's error gate — the full read it already pays converts the sidecar to the full-scan one, so the next resume goes windowed instead of repeating the fallback. Measured: worst case is one doubled read per compaction, vs every resume before.
- **The daemon's turn snapshot arms windowed** via the sidecar's prefix-turn count, paging in the full projection's cursor space; older pages fall through to the hub's file-backed paging.
- **Unconditional full-scan fallback** on any mismatch — missing, corrupt, stale, truncated, or boundary-violating sidecar never errors; the legacy path runs. Fork children fail the file-identity gate by construction.
- **Fold snapshots** (pending attentions with resolutions, delivery commits, client-mutation turns) carried when complete; the transcript package restates the agent fold's rules rule-for-rule (forced by the import cycle) and refuses any file the fold would reject.

Deviation from the issue's design, deliberately: **no clean-shutdown anchor** — shutdown cannot know the checkpoint boundary without re-reading, and a wrong offset would skip live history. The opportunistic anchor covers the shutdown case on the resume that follows.

## Verification

- All 16 issue-list tests, plus the suites: `agent` 118.9s, `cmd/evener` 26.7s, `server` 8.0s, `internal/apptranscript` 7.7s, `agent/transcript` 1.4s; gofmt/vet clean; the `FuzzDelegateAttentionFold` extension ran 875+ clean engine executions.
- **Differentials, not fixtures-as-proof**: the sidecar's prefix-turn count is asserted against the real full projection over the same entries (including every non-projecting kind — resolution, system, communicate-only tool results); the snapshot's refusals are asserted against the agent fold over 17 shapes AND over fuzz-generated inputs; the windowed snapshot's turns/cursors are asserted against the full projection's.
- **Every fix in the loop is mutation-verified** — reverting each one fails the test that claims to guard it (the exact failure mode, not a vacuous pass).
- Lifecycle walked empirically by a reviewer: fresh → compaction anchor → windowed resume → refresh → windowed → second compaction → refresh, every transition validating (identity, anchors, boundary), every window matching ResumeHistory's.

## Review process

Four simplify angles (Reuse / Simplification / Efficiency / Altitude) then two adversarial rounds, all findings adjudicated against the code before applying. The loop caught and fixed: a boundary cursor that stranded window turns (medium), a failure-count attribution bug on fork children (medium), the fold snapshot's attention dedupe and dropped resolutions (medium), two parity gaps plus their missing differential (medium), an inflated prefix-turn count that duplicated turns at the hub paging seam (**high** — the projection restatement counted kinds `ProjectTurn` emits nothing for), a quadratic conflict scan (medium, measured 8.9s at 40k commits), a vacuous prelude test, and an unguarded error-gate ordering. Each round's findings shrank; round 2's verified-clean sections cover the full lifecycle, crash windows, rollback handling, and fold parity.

Follow-ups filed: #742 (the rearm fallback's per-resume full read for non-serve consumers), #743 (the fallback's doubled decode), #746 (the sidecar's 1MiB limit vs delivery-heavy transcripts).

### DIFF
--- a/agent/delegate_tree_restore.go
+++ b/agent/delegate_tree_restore.go
@@ -12,14 +12,15 @@ import (
 	"strings"
 
 	"primeradiant.com/evener/agent/internal/delegatestore"
+	"primeradiant.com/evener/agent/schema"
 	"primeradiant.com/evener/agent/transcript"
 )
 
 type delegateAttentionResolution string
 
 const (
-	delegateAttentionConsumed  delegateAttentionResolution = "consumed"
-	delegateAttentionDiscarded delegateAttentionResolution = "discarded"
+	delegateAttentionConsumed  delegateAttentionResolution = schema.AttentionDispositionConsumed
+	delegateAttentionDiscarded delegateAttentionResolution = schema.AttentionDispositionDiscarded
 )
 
 type shellNotificationIdentity struct {
--- a/agent/schema/turn.go
+++ b/agent/schema/turn.go
@@ -63,6 +63,14 @@ const (
 	TurnAttentionResolution TurnKind = "ATTENTION_RESOLUTION"
 )
 
+// Attention resolution dispositions: the terminal states an attention can
+// resolve to. Exported because two packages judge them — the agent's fold
+// and the transcript package's restatement of it — and must not drift.
+const (
+	AttentionDispositionConsumed  = "consumed"
+	AttentionDispositionDiscarded = "discarded"
+)
+
 // AttentionResolutionInfo identifies one durable attention item and its
 // terminal disposition. The resolution is append-only so cold reconciliation
 // can repeat safely after a crash.
--- a/agent/session.go
+++ b/agent/session.go
@@ -734,6 +734,14 @@ type Session struct {
 	// refresh cannot flip it. Guarded by s.mu.
 	restoredTranscriptOpened bool
 
+	// restoredTranscriptWindowed and restoredTranscriptPrefixEntries carry
+	// the windowed-resume facts to RestoredTranscriptWindowed: whether a
+	// validated sidecar let restore decode only a suffix, and how many
+	// prefix entries that read skipped. A refresh from disk (which re-reads
+	// the whole file) resets windowed to false. Guarded by s.mu.
+	restoredTranscriptWindowed      bool
+	restoredTranscriptPrefixEntries int
+
 	// Cached tool definitions.
 	cachedToolDefs []llm.ToolDefinition
 
@@ -1656,6 +1664,12 @@ func (s *Session) attachedTranscript() *transcript.Writer {
 }
 
 func (s *Session) closeAttachedTranscript() error {
+	// No sidecar anchor here: shutdown cannot know where the last checkpoint
+	// sits without re-reading the transcript, and an end-of-file offset would
+	// skip live history on the next resume (the failure the restore tests
+	// caught in the first draft of this feature). The opportunistic
+	// post-full-scan anchor — written by the resume that just decoded the
+	// whole file — is what covers the shutdown case.
 	s.attentionMu.Lock()
 	defer s.attentionMu.Unlock()
 	return s.attachedTranscript().Close()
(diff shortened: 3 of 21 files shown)


obra:
Closing as superseded; this PR is not being merged.

Review of exact head `8deada30108bc3d14a59c47ae4a8b92226211a16` found that the validated-offset/`PrefixEntryCount` concept may be reusable, but the current resume-sidecar implementation still leaves #742, #743, and #746 unresolved, carries an incorrect windowed `SessionStartData.TranscriptEntries` count, and couples browser paging to legacy `PrefixTurnCount`. Its current CI also has unresolved lint findings.

Successor work is intentionally split:
- atomic transcript-item paging will proceed independently from current main, with the validated entry-offset concept reused only if independently validated;
- navigation entity deltas are unrelated and proceed independently;
- #742/#743/#746 remain open for a separate resume/restore remediation design and PR.

This closure does not mark those issues resolved.

PR #796, opened 2026-09-01
repair: render enum values type-faithfully in constraint messages

### LINKED ISSUE #625: repair: integer enums silently dropped — enum constraints degrade to generic wrong-type-or-value
## Summary

`asStringSlice` (used by `requiredList`, `minimalExample`, `constraintMessage`, and the new `branchRequirement` in `agent/internal/tool/repair/explain.go`) silently drops non-string elements. For a property constrained by an **integer enum** (e.g. `{"type":"integer","enum":[1,2,3]}`):

1. `constraintMessage`'s `enum` case finds an empty allowed list and returns `""`, so a present field with a recognized `enum` keyword degrades to the generic "argument has the wrong type or value" — the specific constraint (allowed values) is never shown. Pre-existing behavior, not introduced by the #618 fix.
2. `branchRequirement` renders a oneOf branch constraining an integer-enum property as `send all of "n"` with no allowed-values clause at all.

Found by adversarial review of the #618 fix; probe-verified with santhosh-tekuri/jsonschema v5.3.1.

## Reproduction

Schema fragment:

```json
"properties": { "n": { "type": "integer", "enum": [1, 2, 3] } }
```

Args `{"n": 5}` produce `argument "n" has the wrong type or value` — the message never mentions that n must be one of 1, 2, 3. Severity honestly assessed: the message degrades to **less specific**, never to **wrong**. Integer enums in tool schemas are rare, and case 1 predates the oneOf work entirely.

## Suggested fix direction

Generalize enum rendering to format non-string values (e.g. a `joinEnum` helper that renders integers/booleans via `fmt.Sprint`), in both `constraintMessage` and `branchRequirement`.

## Environment

obra:
Fixes #625. asStringSlice dropped non-string enum members, so an integer/boolean enum produced an empty allowed-values list and the constraint message degraded to the generic 'wrong type or value'. Enum rendering now goes through one json.Marshal-faithful renderer at both call sites (constraintMessage and branchRequirement): strings quoted, numbers/booleans/null bare, composites as real JSON — i.e. exactly what the value looks like in the JSON the model must send back.

Pipeline provenance: adversarial review REFUTED round 1 (the oneOf path QUOTED integer enums, visually asserting string type and inviting a stringified retry that fails again; and it diverged from the unquoted constraintMessage rendering); round 2 rebuilt it type-faithfully, reverting the round-1 asStringSlice change entirely (required-lists provably never see non-strings), with the string-enum quoting change's goldens swept repo-wide (two beyond the review's named files). Simplify pass trimmed review-history narration and shared a duplicated test fixture. Scientific-notation boundary for extreme floats documented and tested; json.Marshal HTML-escaping of <>& in string enums flagged (no built-in schema hits it).

### DIFF
--- a/agent/internal/tool/repair/explain.go
+++ b/agent/internal/tool/repair/explain.go
@@ -979,6 +979,15 @@ func examplePlaceholder(typ string) string {
 	}
 }
 
+// asStringSlice renders v — a schema's "required" list or "enum" list, in
+// either the []string a hand-built schema may carry or the []any a
+// JSON-decoded one unmarshals to — as a []string. Non-string elements are
+// formatted with fmt.Sprint (the value-rendering convention constraintMessage
+// already uses for a wrong-type instance value) rather than dropped: a
+// required list's entries are always already strings (JSON-Schema property
+// names), so this only changes output for an enum's non-string values — an
+// integer or boolean enum's allowed values must still render into the
+// constraint message instead of silently disappearing (issue #625).
 func asStringSlice(v any) []string {
 	switch s := v.(type) {
 	case []string:
@@ -986,9 +995,7 @@ func asStringSlice(v any) []string {
 	case []any:
 		out := make([]string, 0, len(s))
 		for _, e := range s {
-			if str, ok := e.(string); ok {
-				out = append(out, str)
-			}
+			out = append(out, fmt.Sprint(e))
 		}
 		return out
 	}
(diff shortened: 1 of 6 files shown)


obra:
Rework pushed (9a3a0a616): formatEnumValues now walks any slice/array kind via reflection (subsuming the []string/[]any branches), so typed enum slices ([]int, []bool, []float64, arrays) render JSON-faithfully instead of falling back to the generic message — closing the combined-review Medium. Red-checked twice via stash-isolation on the real ExplainSchemaError constraintMessage and branchRequirement paths.

obra:
Low addressed (4f7085732): the enum violation's actual value now renders through the same formatEnumValue JSON-literal path as the allowed values (%s, not %q — the renderer already returns the quoted-or-bare form, so re-quoting would double-escape) — a numeric violation reads `Value is 5` beside bare allowed values, strings stay quoted. Red-checked on the integer and typed-[]int cases; string cases green throughout. Checked oneOf/branchRequirement for the same pattern: structurally none exists (resolveInstanceValue's only call site is constraintMessage; branch messages never render an actual value).

PR #797, opened 2026-09-01
DO NOT MERGE agent: name live stable delegates in the end_turn running-work warning

### LINKED ISSUE #585: end_turn warning enumerates shell jobs only: live delegates are absent from the undisposed-jobs warning
Filed from triage of #317 while closing it (PRs #386/#420 fixed the drain disposal path; this residual was explicitly left out).

## Evidence
`runningJobIDs` (`agent/jobs.go:1449-1459`) lists `jm.running` only — shell jobs. Live stable delegates held by the session do not appear in the end_turn warning that tells the model which jobs will not survive the run\'s exit.

## Impact
A model ending its turn with live delegates gets an incomplete picture of what exit will kill, weakening the undisposed-job principle the drain otherwise enforces.

## Scope
Small: include delegate-owned running rows in the warning enumeration (owner-authoritative read), with a test pinning the warning text lists both a shell job and a live delegate.

obra:
Fixes #585. communicate(end_turn=true)'s running-work warning enumerated shell jobs and detached processes but never a session's live child delegates, so a turn ended while a delegate was still working with no warning. runningStableDelegateIDs now names each session's direct running delegates, folded into the same warning.

Pipeline provenance: adversarial review REFUTED round 1 (the first filter, OwnerSessionID == s.ID(), only ever matched the tree root — OwnerSessionID is set once to the root for every delegate — so every non-root session ending its turn silently warned about nothing, and the depth-1 test couldn't see it); round 2 re-derived the filter on ParentDelegateID == s.owningDelegateID (the same link blockingDelegateIDs uses; empty-for-root covers all sessions uniformly) and chose direct-children-only to match the shell-job half's established scope, with a depth-2 test proven red both directions (missing warning AND transitive overreach). Simplify shared the test scaffolding. A flagged empty-rootSessionID edge was verified unreachable (the controller-nil guard precedes it). Direct-vs-transitive is a deliberate, documented call for your review.

### DIFF
--- a/agent/delegate_tree_controller_test.go
+++ b/agent/delegate_tree_controller_test.go
@@ -489,6 +489,7 @@ var delegateControllerDormancyExpectedInventory = map[delegateControllerDormancy
 	{filename: "session_attention.go", function: "(*Session).finishRootDelegateAttentionTurn", kind: "session attention method", symbol: "resolveAttentionDurably"}:                    1,
 	{filename: "session_events.go", function: "(*Session).SetDescendantEventFunc", kind: "lifecycle method", symbol: "Snapshot"}:                                                       1,
 	{filename: "session_tools_jobs.go", function: "stableDelegateRowsForSession", kind: "lifecycle method", symbol: "Snapshot"}:                                                        1,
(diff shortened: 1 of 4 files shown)


obra:
I'm not sure this is a good idea.

PR #802, opened 2026-09-01
hub: carry the pairing auth token in the URL path, not the query

### LINKED ISSUE #431: Mobile pairing QR loses its auth token when scanned via the iOS system QR scanner and opened in Safari
## Summary

The Hub's mobile-pairing QR code carries the auth token as a **query-string parameter** (`/auth?token=<...>`). The iOS system QR-code scanner (Camera app / Control Center widget) drops the query string before handing the URL to Safari, so a user who scans via the system scanner and taps **Open in Safari** lands on `/auth` with no token — the pairing silently fails.

The dedicated Evener mobile app's in-app scanner is unaffected (it reads the raw payload and opens it directly), which is why this only shows up on the system-scanner → Safari path.

## Reproduction

1. Open the Hub web UI → Settings → Mobile app. The pairing QR encodes a URL of the form `https://<hub-origin>/auth?token=<opaque>`.
2. On an iPhone, open the **system** QR scanner (Camera app or Control Center widget — *not* the Evener app's in-app scanner) and point it at the code.
3. When the URL banner appears, tap **Open in Safari**.

**Expected:** Safari navigates to `https://<hub-origin>/auth?token=<opaque>`, the Hub validates the token, sets the auth cookie, and pairs the device.

**Actual:** Safari navigates to `https://<hub-origin>/auth` with the query string removed. The Hub sees no `token` parameter and returns 401, so pairing fails with no indication of why.

## Root cause

The token is placed in the URL query string:

- `cmd/evener-hub/internal/hubedge/auth_token.go:254` — `AuthURLFor` builds `base + "/auth?token=" + url.QueryEscape(token)`.
- `cmd/evener-hub/internal/hubedge/auth_token.go:229` — `HandleAuth` reads `r.URL.Query().Get("token")`.
- `cmd/evener-hub/web_api.go:170` — `handleAPIMobilePairing` returns that URL as `auth_url`.
- `cmd/evener-hub/frontend/src/panes/settings/sections/mobile.tsx:52` — `QRCodeSVG` renders `auth_url` into the QR.

The iOS system QR scanner appears to canonicalize/truncate the URL to its origin+path before invoking Safari, discarding `?token=...`. (In-app AVFoundation scanners that call `openURL` with the raw decoded payload pass the full query string through, which is why the Evener app works.)

## Impact

Any user who pairs a mobile device via the system scanner + Safari instead of the in-app scanner gets a failed pairing with no actionable error. The capability URL design is correct for browsers and in-app scanners but is not robust to the system-scanner → Safari handoff, which is a common entry point on iOS.

## Fix options (for triage / implementer)

- **Move the token into the path**, e.g. `/auth/<token>` (or `/auth?t=<token>` rewritten as a path segment), and have `HandleAuth` accept it from the path. Path segments survive the iOS scanner's truncation; query strings do not. This is the smallest change that fixes the system-scanner path.
- **Short-lived exchange code**: the QR carries a single-use code; the destination exchanges it for the short-lived auth cookie via a POST. More secure (token isn't reusable in the QR) but a larger change.
- **Fallback guidance in the UI**: document the iOS system-scanner limitation and steer users to the in-app scanner. This is a workaround, not a fix.

Path-based token is the recommended minimal fix. Either way, a regression test should assert the QR payload survives a simulated query-string-stripping open path.

---

Filed from an in-session report: the QR works for the mobile app's own scanner, but the iOS system scanner → "Open in Safari" path drops the query string (and thus the token) before the Hub ever sees it.

### LINKED ISSUE #795: fuzz-corpus harvester writes live hub bearer tokens unredacted into committed seeds for path-suffix credentials
Security defect surfaced by #431's adversarial security review (2026-09-01), independently confirmed by executing the real harvester logic against a synthetic token.

**Class:** the fuzz-corpus harvester's secret gate does not catch a bearer token that appears as a bare URL path suffix. With entropy-checking disabled on the http-suffix gate (harvest.go's gateDetectSecret(red, false)) and no sanitizer regex matching a bare high-entropy token, a credential carried in the request PATH reverse-maps onto a catch-all route and passes through unredacted into a testdata/fuzz seed the workflow commits to git. The query-string form was never read by this path, so it was immune.

**Why it matters now:** PR #431 (auth pairing token moves from ?token= to /auth/<token>) makes this reachable on the project's own documented EVENER_RECORD_HTTP harvest-and-commit workflow after any normal login — the harvested token is live and reusable.

**Fix direction (either):** exclude the /auth path prefix from the reverse-mapper, or enable entropy-checking on the http-suffix secret gate so high-entropy path suffixes are redacted like other secrets. No real token appeared in the review — a synthetic one was used.

**Gate:** #431 must not merge until this lands (or they land together).

obra:
Fixes #431. iOS's system QR scanner strips the query string, so the pairing URL's ?token= was lost; the token now rides as a path segment (/auth/<token>), read by HandleAuth from the path. Per the maintainer's ruling, the query form is REPLACED, not dual-accepted: HandleAuth no longer reads ?token=, isAuthExempt matches the /auth/ prefix, and a pre-existing AuthGuard self-heal that would have silently resurrected the query form on /auth is excluded. Every producer/reference updated (AuthURLFor, PWA manifest start_url, startup log, 401 body, frontend, 34 scenario cards, all living docs incl. both Trust Boundary sections).

⚠️ MERGE ORDER: this PR must land WITH OR AFTER #801 (fixes #795). Moving the token into the path makes it harvestable by the fuzz-corpus tool into a git-committed seed; #801 closes that leak. Do not merge #431 before #801.

Pipeline provenance: adversarial security lens found and drove the #795/#801 fix (a real credential-exposure regression this PR would otherwise introduce — verified by executing the harvester against a synthetic token); regression lens SURVIVES on the code and drove a docs round (two Trust Boundary sections had contradicted their updated examples); hostile path-parsing, self-heal exclusion, and 401 parity all verified via real go tests. NOTE: the issue thread contains the maintainer's own earlier argument against path tokens (proxy/history exposure) — today's ruling postdates and overrides it; flagged for a conscious merge decision, with the exposure-delta in adv-431-security.md.


obra:
Low addressed (ff715adf9): loadOrCreateAuthToken now regenerates a malformed persisted token — the minimal extension of its existing missing/empty-file self-heal (reject would turn self-healing into a startup crash, since main.go treats any load error as fatal). isWellFormedToken is the exact inverse of generation (base64url decode to 32 bytes), so dot-segments are excluded by shape, not a bolt-on charset check. Red-checked both ways: a persisted `..` surfaced verbatim pre-fix and produced /auth/.. from AuthURLFor; both green post-fix. Also added a real ServeMux + AuthGuard + HandleAuth round-trip test — every prior test invoked handlers directly, so ServeMux's own dot-segment normalization (the finding's actual mechanism) was never exercised. Checks registered in FuzzHubEdgeBehaviorProgram per this PR's precedent; 632k fuzz execs clean.

PR #805, opened 2026-09-01
agent: correct turnHistoryBaseline shrink for mid-fold injected turns

### LINKED ISSUE #634: turnHistoryBaseline not shrink-adjusted on the content-filter ForceCompact path
## Summary

`turnHistoryBaseline` (the N4 in-flight-turn boundary, `agent/session_model_call.go`) is adjusted for history shrinkage only in `prepareModelRequestWithError`'s `ManageContext` path. The content-filter recovery path in `handleModelError` (`modelErrorContentFilterRetry`, ~`agent/session_model_call.go:549-563`) calls `contextMgr.ForceCompact` and publishes the folded history **without** adjusting the baseline. On the next round, `replayScope.InFlightFrom` points too far right, so turns belonging to the in-flight turn are misclassified as pre-turn history.

## Evidence

- Baseline capture + shrink adjustment (the only place it is maintained): `prepareModelRequestWithError`, `agent/session_model_call.go` (~lines 229-241): `if round == 0 { s.turnHistoryBaseline = len(historyTurns) } else if shrink := preManageLen - managedLen; shrink > 0 { s.turnHistoryBaseline -= shrink ... }`.
- Content-filter retry: `handleModelError` → `case modelErrorContentFilterRetry:` copies `s.history`, calls `s.contextMgr.ForceCompact(compactionCtx, &histCopy, "", emitFn)`, and assigns `s.history = histCopy` with no corresponding `turnHistoryBaseline` adjustment.
- If `ForceCompact` folds K turns and the next round's `ManageContext` no-ops (`shrink == 0`), the baseline keeps its pre-fold value while `s.history` is K turns shorter.

## Impact

The N4 replay-provenance boundary drifts after any successful content-filter compaction retry: in-flight-turn classification in `expandHistory`'s `replayScope` is wrong for the remainder of the turn. (Found during adversarial review of an unrelated change that briefly consumed this boundary; the drift is pre-existing and independent — that change no longer uses the baseline.)

## Suggested fix

Adjust `turnHistoryBaseline` by the folded amount in the `modelErrorContentFilterRetry` path (mirroring the `ManageContext` shrink math, clamped at 0), with a deterministic test that forces a content-filter retry mid-turn and asserts the boundary tracks the fold.

### LINKED ISSUE #798: applyPendingForceCompact drops turnHistoryBaseline the same way the content-filter path did (#634 sibling)
Found during #634's fix: agent/session_self_compact.go's applyPendingForceCompact (the /compact tool / compact_context path) calls s.contextMgr.ForceCompact and reassigns s.history without adjusting s.turnHistoryBaseline — the identical drift class #634 fixed for the content-filter retry path. Same fix: call the new shared shrinkTurnHistoryBaseline helper (from #634's branch/PR) after the fold. Red-first test driving the /compact path with a real checkpoint fold.

obra:
Fixes #634. Fixes #798. A ForceCompact fold reassigned s.history without adjusting turnHistoryBaseline, so replayScope.InFlightFrom drifted right and misclassified in-flight turns as prior history.

The fix uses conservation: shrinkTurnHistoryBaseline(preLen, postLen, injected) computes shrink = preLen - postLen + injected, where injected (from compactionEmitFunc.injectedTurns()) counts the turns runPreCompactHook appends mid-fold (pinned note, PreCompact plugin, or goalCompactionSteering — which fires every fold while a goal is active). This is correct regardless of which compaction layers ran. Wired into all three fold paths: the mid-turn ManageContext path (which had the SAME latent under-count — the fix corrects it at the shared root, not just ports it), the content-filter retry (the original #634 bug), and applyPendingForceCompact (the compact_context tool run every round tail — #798, previously never adjusting the baseline at all).

Pipeline provenance: adversarial correctness lens REFUTED the first attempt (the shrink under-counted by S = injected-turn count, inherited from the ManageContext reference it mirrored — the S=0 test couldn't see it); regression lens flagged applyPendingForceCompact as the same unfixed class. Combined round reframed to conservation and wired all three. Scoped re-review ALL ADDRESSED: hand-derived the formula correct including the Layer1+Layer2 window-shift cascade (P=20,R=6,B=15,S=1 → exact index 1), 5-test table red-checked by isolated hunk revert (each S≥1 case off-by-exactly-1 without the +injected term), full agent suite clean. Session.Compact() (/compact command) has the same gap, left unwired with an inline comment (reachability mid-turn unconfirmed) — noted for a follow-up.

(Simplify pass was attempted but its agent hit an API error mid-run; the branch is the re-reviewed-correct tip, minimal and verified.)


obra:
All roborev findings addressed (c15ad2b38 + 041f0b8eb), strict red→green:

**Medium (Session.Compact unwired):** reachability confirmed per the review (server exposes /compact mid-thread; askPendingCount doesn't guard an active round loop) — wired to shrinkTurnHistoryBaseline like the other three callers, red-checked (baseline unmoved at 9, want 4, pre-fix), and the stale known-gap comment replaced with the reachability rationale.

**Medium (strategy self-injections uncounted):** all 7 strategies audited — exactly three (Memory Crystals, Recursive Distill, OODA) append steering outside runPreCompactHook. Rather than widening the Strategy interface across 7 implementations, a context-scoped WithPostFoldInjectionCallback (mirroring WithCompactionTurnCallback) lets them report their net turn delta — computed from history length before/after their filter-then-append step, so steady-state replacement folds count zero, not +1. Red-checked with memory-crystals (got 5, want 4, pre-fix); the three strategies' unit tests now assert their reported deltas.

**Low:** applyPendingForceCompact test parameterized with the withGoalSteering S=1 variant, matching its siblings.

**Coverage close-out:** the mechanism's core claim — hook count + strategy count SUM — is now directly pinned: a combined goal+strategy fold (S=2) lands the baseline exactly right, and a configured-but-inactive strategy composes as a no-op (no double-count, no drop). Both ground-truthed by marked-turn position, never a re-derived formula.

obra:
Round addressed (c5db429c7), both findings red-checked:

**High (Compact() loses concurrent appends):** chose merge over serialize — prepareModelRequestWithError's ManageContext path already uses exactly this discipline (snapshot, fold unlocked, then under the lock re-append anything added since the snapshot before publishing), so Compact() now mirrors the established pattern instead of becoming the only ForceCompact caller holding s.mu across a potentially slow LLM summarization (which would stall the round loop's own appends). postCompactLen/injected are captured before the merge, so the shrink math is unaffected — merged-in turns are past everything the fold measured. Red: the concurrently-appended turn vanished pre-fix; green and race-clean across 10 -race runs.

**Medium (marker replacement before the baseline):** chose positional tracking over in-place replacement — all three strategies' own comments make end-of-context placement load-bearing (model recency, survival through the next fold's preserve window), so in-place would be a functional regression. Added contextmgr.WithBaselineQuery (same context-scoped-callback pattern as WithPostFoldInjectionCallback): the emitter tracks the N4 boundary live through each fold layer, and a strategy queries it before removing a marker, reporting delta+1 when the marker sat before the boundary. Red: 16→7 fold landed the baseline at 4 instead of 3 pre-fix.

Full baseline regression set (all 11 prior tests incl. the S=2 combined pin) green and unchanged; strategies without the query callback fall back to exact prior behavior.

obra:
Round addressed (f08620fed), and per the third-round pattern this stopped patching call sites and unified the discipline: a single revision-checked publishFoldedHistory helper now owns history publication for ALL four publishers (Compact, applyPendingForceCompact, content-filter retry, ManageContext).

Contract: s.historyRevision bumps only on a successful publish; at publish time under s.mu, an unchanged revision means any growth past the snapshot is ordinary appends (merged back, as round 2 did for Compact), while a changed revision means a competing fold published — the stale fold REFUSES to clobber it. Per-caller conflict policy matches each site's semantics: user /compact surfaces an error after one retry; self-compaction skips silently (a competing fold already relieved pressure); the content-filter retry proceeds either way; ManageContext retries once then falls back to current uncompacted history. Baseline shrink stays inside the same locked section using the fold's own pre/post lengths, so publish-and-correct is atomic against a competing pair.

Red-checked all four findings via stash-revert: both ForceCompact callers dropped concurrent appends pre-fix; Compact's length-based merge let a stale fold clobber a concurrent compaction pre-fix; the multi-marker strategy loop reported the wrong index pre-fix. 4 new concurrency tests x10 -race runs each, zero races; all 13 prior baseline tests green unchanged. Branch rebased onto current main (no conflicts).

One test-design note: a real two-concurrent-Compact test deadlocks on cheapmodel's legitimate single-flight serialization (not a production bug) — the race test drives the publish primitive directly instead.

obra:
Round addressed (db58c5fc5):

**High (non-atomic capture):** the inline fold-retry loop now captures (historyTurns, preManageLen, snapRevision) in ONE s.mu block at the top of every iteration, mirroring foldWithForceCompact, with maybeElicitNoteBeforeCompaction moved after that capture — closing the multi-second window where a competing publish could bump the revision before it was read. Covers the Low micro-window too. New TestPrepareModelRequestWithError_DoesNotClobberConcurrentPublish, 10/10 -race.

**High (premature side effects):** full audit — six effects now staged and committed only on a winning publish (compaction-turn transcript/event/naming/steering, steering records, transcript reminder, pinned-note consumption, plugin-hook user-message delivery, nudge-latch reset), via compactionEmitFunc.flush() and runPreCompactHook's commit closure. Deliberately not deferred, with reasoning: ephemeral emits, the hook/note READS (the fold's content must include them pre-publish — only their consumption defers), and pure snapshot comparisons. Red-checked by simulating the old eager flush: the losing attempt ate the pinned note before the winning retry could hand it off — exactly the reported defect. TestFoldConflict_LosingAttemptDoesNotConsumeSideEffects, 10/10 -race.

**Medium (revision completeness):** two-pass sweep of every s.history write — four non-append mutators found (attention retain/remove, inline repair, repairOrphanedToolResults) and routed through a new bumpHistoryRevisionLocked(); all remaining writes verified as appends (safe under merge-back). TestSessionCompact_DoesNotResurrectConcurrentlyRemovedAttentionTurn, 10/10 -race.

Whole-package -race run green; all prior baseline/concurrency tests unchanged.

obra:
Round addressed (7 commits, 2d44bbce5..6a19847e9) — by completing the transaction rather than patching the five escapes individually. publishFoldTransaction is now the single publication path both publishers share, three phases:

- **attentionMu (outer)** — the same door every transcript write already takes — held across the publish decision AND the fold's own transcript entries, so competing folds' markers land in publish order and concurrent turns' entries land after them (the ResumeHistory anchoring guarantee the High demanded). Lock order matches the codebase's existing attentionMu→s.mu; the reverse nesting is structurally impossible.
- **s.mu (inner, tiny)** — revision check + history swap + the caller's baseline correction + a generation-checked pinned-note claim (pinnedNoteGen: a losing fold can't re-inject a claimed note, and a note re-pinned mid-fold survives an older fold's publish). Never held across I/O.
- **Outside both** — the staged effects flush: EventContextCompaction buffers until publication succeeds (a discarded fold emits nothing; attempt-warnings still pass live since they describe the attempt, which really ran).

The orphan-repair path was consolidated onto one capture-repair-publish lock hold (both its append-loss and fold-clobber shapes red-checked), and the inline path's baseline SET/shrink moved into the publish's own locked section — the audit confirms both publishers now carry baseline-in-transaction. The Low's triplicated marker logic is one shared contextmgr helper with a multi-marker-straddling-baseline test.

11 new tests, each -race x10; all ~20 prior pins untouched and green; whole-package -race green; full make lint passes. One honest residual flagged for the record: turns merged back during a fold keep their pre-marker transcript entries and so don't survive resume — the pre-existing resume contract, unchanged by this round; fixing it would mean entry rewriting or marker metadata, deliberately out of scope.

obra:
Round addressed (e9f120648, b0e817585, f1e2d2912):

**High (turns recorded during a fold lost on restart):** the fix pairs every history-append with its transcript write under attentionMu (a shared appendTurnAfterTranscriptWrite helper for the seven durability-first sites; recordTurn for append-first ones), so a pair lands entirely before a publication or entirely after — never straddling. At publication, the transaction computes the merged tail under s.mu and, still inside its attentionMu hold, re-writes the tail's entries after the compaction markers: resume sees exactly one post-marker copy per merged-back turn in live order (attention folding is ID-keyed, so the pre-marker original is idempotent for full-transcript readers). Two alternatives were rejected with reasons: serialize-only cannot work (the racing pair completes mid-fold, before any transaction exists to serialize against), and marker metadata means a transcript-format change old readers can't parse — a compat decision deliberately not taken unilaterally. Red-checked at the restart layer the review demanded: recordTurn completes mid-fold via a blocking-summarizer seam, and ResumeHistory over the real transcript contains the turn exactly once (absent pre-fix). -race x10.

**Medium (mid-history mutations vs the N4 boundary):** repairs report insertion indexes and the boundary moves atomically with the mutation (+1 per insertion at-or-before, −1 per deletion strictly before, clamped, inside the same locked section as the revision bump) — three red cases plus an at-boundary-unchanged guard, ground-truthed by marked-turn position.

**Low:** the revision comment now covers non-append mutations.

All ~31 prior pins green; full package plain + -race green. One flagged residual for the record: the attention-retain fallback re-inserts a turn whose transcript entry can be arbitrarily old — its restart story flows through attention re-folding rather than ResumeHistory; left out of scope deliberately.

obra:
Round addressed (0bf30c0fc, 807ffc140, ecd8040f8):

**High (privacy):** tracking, not reconstruction — reconstruction was rejected on proof: the placeholder projection needs call arguments and result internals gone by rewrite time, and a wrong guess in the unparseable corner UNDER-redacts, the one failure mode a privacy fix cannot have. Every append/write pair now records the exact persisted form it committed (a session pair log, maintained inside the pair's existing locks, pruned wholesale on every successful publication); the merged-tail rewrite re-appends those persisted forms after the markers — byte-identical projection semantics. Red-checked: a divergent pair (secret live / placeholder+commits persisted) merged mid-fold leaked the secret post-marker pre-fix; now the secret appears nowhere in the transcript, resume carries exactly one placeholder copy with its DelegateDeliveryCommits, and live history keeps the evidence. -race x10.

**Medium (resurrection):** closed by the same construction — attention-retained turns and repair synthetics never enter the pair log (their restart stories belong to attention re-fold and resume-repair), so the rewrite cannot manufacture an entry for a turn deleted between publish and rewrite. This also fixed an r6 regression the round itself had introduced (attention turns entering the session transcript at all, risking double-injection on resume).

**Medium (effect ordering):** foldCommits stamp their publication sequence (historyRevision at publish); a stale parked flush skips the last-write-wins classes (namer, task-list reminder, artifact reminder) once a newer publication flushed, while additive classes (events, steering, hook messages) still run — with an audit finding no other last-write-wins members. Red-checked: A-parks/B-flushes/A-flushes → B's name survives.

**Medium (invalid test):** the premise was first verified wrong (the summarizer returns summary+preserved, so fold B was genuinely competing) — rebuilt anyway per the review with fresh post-A history and a markers-in-publish-order assertion, no longer leaning on summarizer internals. **Lows:** the non-atomic wrapper carries a loud TEST-ONLY doc; the convenience helper is documented.

All ~34 prior pins green; full plain + -race suites green. Accepted cost documented: the pair log grows one persisted form per append between compactions (bodies shared with live history), cleared on every publication.

obra:
Round addressed (fb7f60f91):

**Medium (stale namer completing late):** the r7 gate ordered flush STARTS; this adds ordering at naming COMPLETION. The flush threads its publication revision into the async naming chain, and applySessionNameResult evaluates a superseded-check under the same s.mu hold that applies the name — check and apply are atomic, so a namer launched by fold A before B published drops its result at completion regardless of goroutine finish order. Chosen over serializing flushes because every naming outcome already funnels through that one application point. Red-first through the REAL namer path (a scripted namer client — the function-seam bypass can't see the gate): A's namer parks mid-LLM-call, B publishes/flushes/names, A completes → 'name-A' won deterministically pre-fix, B's name survives now; -race x10. One test-design note worth recording: the first red attempt passed vacuously because session Close()'s closing-state guard suppressed the stale apply for the wrong reason — caught and restructured (join via sendersWG, not Close) before any fix was written.

**Artifact check:** session_model_call.go:337 on the committed head is `s.stageRootDelegateAttentionCoverage(req, historyTurns)` — valid Go; the reviewer's guessed `const maxFoldAttempts = 2` exists intact at line 231; the workspace builds clean. Diff-rendering artifact, no code change.

Rebased onto current main (20 commits, clean); all ~36 prior pins green; full plain + -race suites green.

obra:
Round addressed (66fe01212):

**Medium (published-slice aliasing):** publishFoldTransaction now returns a defensive copy, made under the same s.mu hold that published, immediately before unlock — the choke point rather than the call site, because a call-site copy would itself read the shared backing array unlocked (racing the very in-place mutations it defends against), and the choke point closes the class for every current and future caller by construction. The contract is documented on the transaction: the returned slice is readable without locks. Cost: one shallow copy per successful publication — what the pre-overhaul always-copy code already paid.

**Audit:** exactly two callers — prepareModelRequestWithError (the flagged one; expandHistory, anchor planning, and attention staging all read the copy now) and foldWithForceCompact (discards the return; its own working slice is never read after a successful publish, and a failed one re-snapshots). The inner publishFoldedHistory's only production caller is the transaction; the !baselineSynced branch and delegateModelHistorySnapshot both still re-snapshot under s.mu. No other post-publish consumer keeps the slice.

**Red/green:** the race detector is the assertion — a winning fold with an attention turn seeded inside the preserved window, parked post-publish, while a goroutine hammers the production in-place retain shape across the post-publish reads: deterministic DATA RACE warning on the prior code, clean under -race -count=10 after. All ~37 prior pins green; the round touched one production file plus the test.

obra:
Round addressed (2e2ef61de + 1b2df9693):

**High (flush-order suppression):** diagnosed precisely — the review's literal interleave (older flush marks the newer as superseded) is impossible with the committed strict-monotone comparison; the REAL residue was that suppression bound to flush order: an older fold whose deferred flush ran while a newer fold was published-but-unflushed sailed through and fired stale last-write-wins effects (transient stale name via the namer, stale steering). Fix simplifies rather than adds: ONE marker, newestPublishedFoldRevision, set at publish inside the transaction's s.mu window — a flush is superseded iff a newer fold has PUBLISHED, and the naming-completion gate compares against the same marker. The suggested second (flushed) marker existed only to express the flush-order semantics being removed, so it fell away; the newest fold's own flush can never be suppressed, by induction (argued in the gate's comment). All three ordering scenarios now hold simultaneously — the new older-flushes-first red test (stale summary reached the namer pre-fix), plus both prior ordering pins — each -race x10.

**Low (lock-order trap):** moved, not renamed — the test-only commit wrapper now lives in the test file and is deleted from the production surface entirely, making a stray production caller impossible rather than merely discouraged (both remaining callers are tests; tagged fuzz builds see untagged test files, confirmed by lint-evenerfuzz).

All ~38 prior pins green; full plain + -race suites green.

obra:
Round addressed (45a3fca7a) — with a correction of the finding itself:

**Medium (unguarded marker read):** diagnosed before implementing — the flagged closure had exactly one evaluation site, inside applySessionNameResult's s.mu hold (the completion-gate design), so the read was locked and no data race existed; the prescribed fix (locking s.mu inside the closure) would have self-deadlocked against that very hold. Empirical proof: a test overlapping a parked gate evaluation with a 25-fold publication hammer on the real namer path ran CLEAN under the race detector on the pre-change code.

The legitimate weakness underneath: the read's safety was a call-site contract invisible at the read site — one refactor from becoming the race the reviewer saw. Fixed structurally: the naming chain threads the launching fold's revision as a plain value, the closure is deleted, and applySessionNameResult performs the comparison itself, lexically inside the same s.mu hold that applies the name. Check-and-apply atomicity unchanged; there is no longer any site reading the field outside an explicit lock. Reader inventory (full grep): the publish write and both gates are explicitly locked; nothing else references the field.

The hammer test stays as a drift pin (-race x10): any future change moving the gate evaluation outside the applying hold is caught immediately. All ~39 prior pins green; full plain + -race suites green.

obra:
Round addressed (3a8e80f38):

**Medium (shared CompactionMeta):** per-call, via context — stageCompactionEffects (the single choke point every publisher stages through) installs the metadata into the fold's own ctx, every consumption site resolves it from there (both checkpoint layers, the strategy layers, session-log checkpoint, transcript-recovery — OODA included), and the three publisher writes to the shared field are deleted. Manager.Meta survives only as a documented single-threaded fallback behind a frozen test-facing signature. Red-checked: a parked fold with a 15-Compact hammer across its mid-fold meta reads — deterministic DATA RACE pre-fix, clean -race x10 after, plus a deterministic meta-threading probe proving each fold's published result carries its own metadata.

**Low (discarded instructions):** warning, not requeue — an instruction set that loses both publication attempts now emits an EventWarning naming the instructions verbatim (the fold's own loss stays silent, per the established policy table); requeue was rejected because the one-request-per-round policy has no queue to ride. Red-checked both directions.

**Process disclosure, in the open:** this round found that piped verification chains (`go test ... | tail`) had been masking test exit codes — a broken intermediate commit was briefly pushed before set -o pipefail gated it; the amended commit replaces it, the indentation-scoped edit miss it hid is fixed, and the diagnostic that caught it is kept in the test's failure output. All ~41 prior pins green; full plain + -race suites green.

obra:
Round addressed (52b803c7c):

**Medium (attention content lost across compaction restart):** fixed at rearm time — the smaller, better-owned shape. Three structural facts drove the choice: the rearm's transcript fold already holds the full durable turns for every attention ID (the content was always present at rearm; only reinjection was missing); the restore path installs resumed history before the rearm runs, so reinjection lands on final history; and retainDelegateAttentionTurn is the existing ID-keyed idempotent reinjector (in-place replace when resident, append when missing, error on conflict), called inside the rearm's attentionMu hold in the codebase's lock order. The fix is a small loop after pendingIDs(): retain each pending ID's durable turn before arming the wake. The alternative (admitting attention turns into the pair-log rewrite) was rejected because it would break the rewrite's no-attention-entries invariant and duplicate the idempotence argument in the publication path, while this keeps the attention machinery owning its own restart story. Both guards fall out structurally: resolved attentions are never pending (nothing resurrects), and retain replaces in place (a boundary-free restart never duplicates).

Red-first at the real restart layer (RestoreSessionFromMeta → ResumeHistory → rearm over a hand-built transcript): pre-fix, the pending ID re-armed with its model-visible turn missing — the reviewed loss verbatim; post-fix, the content is present exactly once, the wake arms, and the resolved turn stays gone; the boundary-free guard subtest is green in both eras. -race x10; all ~42 prior pins green; one production file + one test.

obra:
Round addressed (1dc69fd03):

**High (shallow Turn copy leaks losing-fold mutations):** verified — it reproduces. A red test seeded a maskable observation and clearable thinking before the fold cutoff, forced every fold attempt to lose the publication race against a re-publish of the same turn structs, and live history came back holding the masked summary through the shared `*ToolResultData` pointer, exactly as reviewed. Fix: copy-on-write in the mutators, chosen over deep-copying snapshots before strategies run — the writers own their write discipline, the copy cost is paid only for parts actually masked or cleared, and the guarantee holds for every shared-history caller rather than one snapshot site. The audit found a third mutator beyond the two cited (obs-mask's aggressive masking path), fixed identically. Masking outputs unchanged; red→green under -race -count=10.

**Low (error message too narrow):** generalized to "a concurrent history change won the publication race; try again", red-first on the message text; the one test matching the old wording updated.

**History:** the branch is collapsed from 29 review-round commits into 6 commits grouped by logical unit (transaction; atomic append/transcript pairs; persisted-form rewrite and winner-scoped effects; publication-ordered effects, snapshot isolation, per-call metadata; attention rearm restoration; copy-on-write mutators). Each commit builds and vets on its own; the final tree is byte-identical to the fully verified pre-collapse tip, and the agent and contextmgr suites were re-run green on it.

obra:
Round addressed (fde87a930) — history rebuilt on current main.

**Medium (last-write-wins check is TOCTOU):** reproduced. Parking a fold right after its supersession check and publishing plus flushing a newer fold in that window left two transcript reminders in the steering queue, the stale one landing after the newer fold's. Fix: the supersession check now runs inside the same lock hold that appends to the steering queue, so a newer publication cannot slip between check and enqueue; task-list and transcript reminders route through that gate. No new mutex, lock order unchanged, flush still outside the session lock. Red-first pin, green under -race -count=10.

**Low (substring marker match):** a steering turn merely mentioning the marker was removed. Now matched by prefix. Red-first pin.

**Low (review-history comments):** swept every added comment line in the branch diff; comments now state current invariants and rationale only. Nothing outside the branch diff touched.

**History:** still six commits by logical unit, now rebuilt by patch replay onto current main. One semantic conflict with #824 was reconciled rather than clobbered: main's model-recovery force-compact helper used the unlocked snapshot-and-replace pattern this PR removes, so its call sites are kept and its body routes through the same revision-checked fold publication as every other ForceCompact caller; main's test for it and the token-budget suite pass unchanged. Diff footprint vs main is 32 files, all under agent/. Full agent, contextmgr, and llm suites green on the tip.

obra:
Round addressed (e13de4279) — minimal fixes, rebased on current main by patch replay.

**Medium (merged-tail rewrite not durable):** reproduced at the restart layer. The pin snapshots the transcript bytes on every fsync via a small afero wrapper, parks a fold, records a turn durably, lets the fold publish, then restores from the snapshot and resumes: the summary marker was durable and the merged-back turn was gone. Fix is one call: the rewrite loop in publishFoldTransaction writes through writeTranscriptDurableLocked. The pair log records no per-turn durability flag, so the rewrite is simply durable. Green under -race -count=10.

**Low (doc says "contains"):** comment now says "begins with", matching the code and tests.

Still six commits; diff vs main is 32 files, all under agent/. Agent, contextmgr, and llm suites green on the tip; full make lint passes including lint-gofmt now that #836 is in.

PR #807, opened 2026-09-01
agent: bounded, cancelable journal scanning with incremental fold caching — no truncation

### LINKED ISSUE #448: Persisted job activity reads unbounded journals before applying limits

### LINKED ISSUE #806: delegatestore.ScanEvents: io.ReadAll lost os.ReadFile's pre-allocation
Noted during #448's simplify pass: the bounded scanner reads the delegate journal via io.LimitReader + io.ReadAll, which grows its buffer by doubling, where the old os.ReadFile pre-allocated via Stat. For journals approaching the 128MiB scan ceiling that's real extra copying. Fix is small (Stat + bytes.Buffer.Grow or a sized ReadFull loop) but adds code, so it was out of scope for the simplify pass. agent/internal/delegatestore/read_events.go.

### LINKED ISSUE #812: Activity continuation never advances: resume re-projects from scratch with a fresh budget
Latent defect surfaced by a probe test during PR #807's round (roborev finding 1): projection's pre-existing mid-list continuation does not skip already-shown records on resume — it re-projects the same ordered list from the start with a fresh budget, so pagination can return the same window repeatedly. This was invisible under the old fully-unbounded load and became reviewer-visible once load-time ceilings cut earlier. PR #807 responded by removing the non-advancing load-truncation continuation (honest Truncated-without-continuation); fixing resumable pagination properly needs a real cursor (position/offset the projector honors). Related: #450 (the continuation test asserts only non-emptiness, not progress). agent/jobs_activity.go projection/continuation path.

obra:
Fixes #448. Fixes #806. Fixes #812 (on this surface).

Two layers, one PR (originally stacked, collapsed after review showed they must merge as a unit):

**Bounded plumbing:** jobstore/delegatestore ScanEvents are streaming, cancelable, and per-line bounded (linecap: ctx-checked, capped drain); cancellation propagates from descendants instead of being swallowed; the hub-reachable ThreadRead path uses the same bounded scan; corruption (an oversized line) is classified at the shared choke point — the thread still renders, the delegate-status/activity surfaces carry an explicit file+line diagnostic, and no unrelated surface is bricked.

**Incremental fold (the maintainer's 'Tuesday' ruling — truncation of legitimate history is unacceptable):** journals are append-only, so folds are cached (foldcache: LRU values with a never-evicted per-path epoch map, singleflight-coalesced on a detached context, staleness by size/mtime plus a 64-byte tail probe that catches same-size and mtime-pinned rewrites) and each request replays only the appended delta. Continuations carry real advancing cursors (ResumeIndex + epochs; explicit-depth loading so max-depth continuations load their target's children). The per-file scan ceilings are GONE — a 200k-event journal folds fully (the Tuesday crux test), every event reachable by pagination, with proven delta-reads (scan-stats assertions). What remains bounded is per-line size and per-request work, not history.

Pipeline provenance: three adversarial reviews (two lenses on the fold layer REFUTED with empirically-confirmed Criticals — eviction-reset epochs, ThreadRead blast radius — all fixed red-first), five roborev rounds on the plumbing layer, a scoped re-review in flight on the fold layer. ⚠️ Merge coordination with PR #472 unchanged: the fold cache wraps the scans, not loadHistoricalActivityBase's shape; #472's multi-source merge calls the same cached scans per source.


obra:
All 7 roborev findings addressed (4 commits, b95249e0d..8eb869a3a), strict red→green per finding:

1 (continuation rescans from byte zero): chose honest truncation over a cursor — a probe test proved the pre-existing continuation machinery never actually skips shown records on resume (latent defect, now filed separately), so minting one was worse than none; load-truncated sessions report Truncated with no continuation. 2 (depth-skip): placeholder snapshot so projection emits a proper Truncated branch + continuation. 3 (live path): comment now tells the truth — bounding LoadOrdered means redesigning cursor-trust invariants shared with live-append; filed as a follow-up. 4 (whole-file read): delegatestore now streams line-by-line with ctx checked per line — also Fixes #806. 5: ctx check per attention-status row. 6: cancellation checked before every limit-shaped return in both stores. 7: TornTail only for a genuinely unterminated file, correct by construction.

Follow-ups filed for the live-path bound and the non-advancing continuation.

obra:
this change would silently truncate user-visible content. it's not acceptable as is

obra:
All 8 findings from the rebased-head review addressed on this branch (tip ebd1dbca2), each red-first where reachable: LoadTruncated no longer mints a non-advancing continuation (suppressed via a loadTruncated flag); continuation paths capped and budget-charged per hop; the raw scan ceiling decoupled from the projection work budget (fixed 100k, 50x the work-unit budget); delegatestore MaxEvents checked before batch decode; delegate-journal truncation now propagates into LoadTruncated instead of hiding in a diagnostic; the double-fold and its error-swallowing removed; dead decode branches reverted; the depth-placeholder uses descriptor.TranscriptRef (its edge verified unreachable today and documented rather than faked into a test).

Note on merge sequencing: the stacked incremental-fold branch (removing truncation entirely per the maintainer's ruling — fold-once caching + advancing cursors) structurally supersedes several of these load-truncation mechanisms; this branch is kept correct standing alone, and the stack rebases over it. Recommend merging the pair together.

obra:
The rebased-head review's one Medium is fixed: an over-budget session (2001 owned records vs the 2000-unit budget) previously hit activityConsumeWorkUnit's refuse-without-consuming contract, leaving the budget unspent and every sibling journal still scanned. The call site now saturates the budget on refusal (usedWork = max), so the rest of the traversal sees zero remaining — red-checked with a 2001-record test (2 journals scanned pre-fix, exactly 1 after), sibling budget/depth/cancellation tests green. (On the stacked incremental-fold branch this shape is superseded by 1-unit-per-session accounting; the base stands correct alone.)

obra:
Rebased-head round addressed (a694e975b): LoadTruncated now copied through activityFilterSnapshotToDelegate (a continuation through a truncated parent no longer hides its incompleteness); load-phase depth is an explicit parameter set by every caller (0 at any depth-0 node) instead of len(visited)-1, so a max-path continuation loads the target's own children — red-checked with an activityMaxNewDepth+2 chain; both stale doc comments corrected. The depth derivation bug was also confirmed present on the stacked branch and fixed there in the same shape.

obra:
Consolidated round addressed (6 commits, tip 6063920b9) — every finding from the collapsed-head review plus the independent re-review's confirmations:

**Highs:** depth-boundary continuations no longer dead-on-arrival (mint/decode limits aligned, resumed end-to-end at the boundary); foldcache no longer assumes growth+mtime-change is an append — the tail probe runs on every incremental extension, so a truncate-and-rewrite-to-larger forces a full rescan and epoch bump.

**Mediums:** every truncation continuation constructor now threads real epochs (no more falsely-stale rejections after a bump); live pagination is bound to a checked revision (jobActivityClock) so mutable snapshots can't skip/duplicate across pages; delegatestore rechecks MaxEvents after the final batch; fold validation restored in the full-read wrapper (streaming scans stay unfolded, contract documented); journal delete/recreate carries a tombstone generation so an old continuation can't validate against a new file; the unterminated-complete-JSON duplication is closed by advancing the offset past any successfully-decoded line — chosen over exclusion after the first attempt broke a fuzz seed pinning the forensic reopen-recovery contract (JSON brace-balancing guarantees a valid decode is never a prefix of a longer record). The tail-probe same-size residual is documented as the accepted tradeoff per the independent re-review's implausibility analysis rather than strengthened.

**The elevated one:** the re-review proved the corruption diagnostic never actually reached the wire (dropped by a delegates-nonempty gate) — fixed with a dedicated appwire DelegateDiagnostics field attached independently of delegate count, red-checked at the wire payload, protocol docs/types regenerated. Plus: cancellation propagates from the past-metadata merge, x/sync declared as a direct dep (surgically, avoiding a workspace-wide tidy), stale docs corrected.

All 6 crux tests pass under -race; full suites green.

obra:
Round addressed (5 commits, tip 66527416a):

**High (trim continuation restarts from zero):** ResumeIndex now carries the trimmed entry's index. Red-checked observably, not by field assertion — a walk following each minted continuation the way a client would, over 20 jobs sized to force size-trimming: pre-fix the walk hit its runaway-page cap because every resumed page was identical to the first (the infinite redelivery the finding describes); post-fix every job arrives exactly once, in order, under -race. One boundary flagged rather than silently expanded: the trim index is local to the projected page, so if a single RESUMED page's remainder were itself oversized enough to trim again within the same chain, the index would under-count one page later — a materially rarer variant (many oversized entries + multiple trim rounds in one chain), not exercised by any test and outside the finding's literal instruction, so it's named here for a deliberate call instead of fixed as untested behavior.

**Low (temporal comments):** a diff-wide sweep, not just the cited examples — ~105 comment violations rewritten as current contracts/rationale (zero roborev/#807/#448/#812 citations remain; 18 look-alikes verified as legitimate runtime-state descriptions and kept; one stray citation in a test failure message also cleaned; plus a stale duplicated doc block consolidated in both stores' tests).

**Low (helper triplication):** countdownContext extracted to agent/internal/agenttest — the module's existing shared test-scaffolding package, importable by all three call sites, stdlib-only.

Full suites, all 6 crux tests, and the trim/continuation sweep green under -race.

Issue #803, opened 2026-09-01
Registry: explicit first-party vs openai-compatible endpoint distinction for vendor-hosted capabilities


obra:
From Jesse's #738 ruling (2026-09-01): web_search (and other vendor-hosted platform capabilities) should only be enabled on a vendor's own first-party endpoint. #738 implements this as a base_url-divergence gate. Jesse suggested a cleaner config model: 'presumably we should have an openai compatible endpoint option and an openai endpoint option, and only the first-party endpoint enables web_search.' That is a larger registry-model change — an explicit first-party-vs-compatible base distinction (e.g. an openai-compatible base separate from openai) that gates vendor-hosted caps by identity rather than by base_url string comparison. Scoped separately from #738's targeted web_search fix; worth designing if more vendor-hosted caps appear.

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

**Verdict:** `not-a-bug` — #738's targeted fix is present and working in the pinned revision: `gateWebSearch` (`llm/registry/resolve.go:377-393`) strips web_search unless the resolved endpoint equals the provider's first-party endpoint, judged by `firstPartyEndpoint` (`llm/registry/instances.go:155-177`, comparing base URL plus endpoint/stream/count-tokens). The issue reports no defect in that behaviour; it proposes replacing a string-comparison gate with an explicit first-party-vs-compatible registry distinction.

**Recommendation:** `close` — this is a speculative registry-model redesign ("worth designing if more vendor-hosted caps appear"), not a current problem. Adding a new base kind to the registry is exactly the overbuilding the current gate avoids; revisit only if a second vendor-hosted capability actually needs the distinction.

PR #808, opened 2026-09-01
registry: enable web_search only on a vendor's first-party endpoint

### LINKED ISSUE #738: A custom base_url instance inherits openai.com's web_search and dies on turn 1
A `base = "openai"` instance pointed at a custom `base_url` inherits openai.com's `web_search = true`, and evener then sends the provider-side web search tool on **every** request. A gateway that doesn't implement it rejects the request, and there is no fallback: the very first turn dies and the session ends.

Found while retargeting the TB 2.1 eval at current main. The eval's instance is Bedrock's OpenAI-compatible endpoint:

```toml
schema = 2
default = "bedrock"

[providers.bedrock]
base = "openai"
base_url = "https://bedrock-runtime.us-west-2.amazonaws.com/openai/v1"
api_key_env = ["OPENAI_API_KEY"]
```

(exactly what `evener migrate` produces from the pre-registry file, unedited)

Every run terminates on turn 1, before any tool call:

```
[model] us.openai.gpt-5.6-luna (bedrock)
[error] bedrock error (status=0): responses.create(stream): web search is not supported for this request
evener: provider error: bedrock error (status=0): responses.create(stream): web search is not supported for this request
```

`models inspect bedrock/us.openai.gpt-5.6-luna` confirms where it comes from — the model isn't in the catalog, so the record is synthesized from the provider, and the provider is openai.com:

```
warnings: ['model not in catalog']
synthesized: True
caps.web_search: true
```

Adding `web_search = false` to the instance fixes it, and that is a legitimate escape hatch. The problem is the default and the failure mode.

**Why this is the wrong default.** The registry already treats a custom `base_url` as a signal that the instance is not the vendor: §10 stops it inheriting the vendor API key, and `evener migrate` prints a note saying so. Capability inheritance doesn't follow the same rule, so an instance that is explicitly *not* openai.com is still credited with an openai.com platform feature. `llm/registry/testdata/golden/openai-gpt-5.5-proxy.json` pins `web_search: true` for exactly this shape, so it looks deliberate — but a proxy that forwards to openai.com and a gateway that merely speaks the protocol are indistinguishable in config, and only the first can honor it.

**Why the failure mode is worse than the default.** The tool goes out on every request whether or not the model would ever call it, so this isn't a degraded search — it's total loss of the provider. The error arrives as an opaque provider string with no hint that one capability bit caused it or that `web_search = false` resolves it. Nothing in `evener providers list` or a startup diagnostic flags the combination beforehand.

Suggestions, roughly in order of preference:

1. Don't inherit `WebSearch` (and any other platform-side capability that is not a wire-protocol fact) when an instance sets its own `base_url`, matching the existing rule for the vendor key. An explicit `web_search = true` still opts a real proxy in.
2. Failing that, make the first rejection recoverable: drop the tool and retry once, and say so, rather than ending the session.
3. At minimum, map this provider message to an error that names the capability and the one-line config fix.

Worth noting for triage: this is invisible to the test suite because the golden files assert the resolved record, not what a non-openai.com endpoint does with the request built from it.

obra:
Fixes #738. An instance with its own base_url (e.g. base = "openai" pointed at a proxy) inherited the vendor's provider-level web_search = true and died on turn 1 when the endpoint rejected the hosted tool. Per Jesse's ruling, web_search is a FIRST-PARTY-ENDPOINT capability: firstPartyEndpoint(rec) keeps it only when the instance uses the vendor's own canonical endpoint — a literal base_url= override strips it, and so does a *_BASE_URL env override (deliberately stricter than spec §10's credential endpoint-stop, which lets env overrides inherit: a wasted key is harmless, an unimplemented hosted tool kills the request). A vars-only Vertex/Bedrock instance (template intact, deployment vars supplied) IS first-party and keeps it. Explicit web_search = true/false on the instance always wins, and a resolve-time warning explains the strip.

Pipeline provenance: both adversarial lenses REFUTED round 1 with a Critical — the divergence heuristic exempted the whole google-vertex family (unresolved template vars read as 'no baseline → not diverged'), leaving the issue's exact bug open for Vertex; Jesse's first-party ruling reframed the gate, fixing that and a latent Bedrock hole. Scoped re-review ALL ADDRESSED (vertexgw red-checked both directions; vars-only Vertex goldens byte-identical; env-override strips while plain openai keeps). A parity test pins the warning on both resolve paths per the file's convention. Simplify: comment corrections only. Other-caps enumeration (twice-verified): web_search is the only provider-level vendor-hosted cap; everything else is wire-protocol shape or model facts.

Notes: spec §10 is deliberately NOT amended — the stricter web_search rule awaits Jesse's own inline-annotation convention. A hypothetical future preset-defined base_url would misread as user-typed (unreachable today — no curated preset defines one). The larger first-party-vs-openai-compatible config-model idea is tracked as #803.

### DIFF
--- a/agent/session_model_test.go
+++ b/agent/session_model_test.go
@@ -918,6 +918,62 @@ func TestSession_WebSearch_FlagSetOnRequest(t *testing.T) {
 	}
 }
 
+// TestSession_WebSearch_FlagNotSetWhenBaseURLDiverges covers issue #738: an
+// instance with its own base_url no longer inherits the vendor's
+// provider-level web_search (llm/registry's endpoint gate), so the request
+// this session builds must not carry the flag - a gateway that does not
+// implement the hosted tool rejects it and, before this fix, ended the
+// session on turn one. "gw" is the fixture's base=openai instance pointed at
+// a different base_url (profile_testhelpers_test.go); it never sets
+// web_search itself.
+func TestSession_WebSearch_FlagNotSetWhenBaseURLDiverges(t *testing.T) {
+	t.Parallel()
+	dir := t.TempDir()
+	c := llm.NewClient()
+	f := &fakeAdapter{
+		name: "gw",
+		steps: []func(req llm.Request) llm.Response{
+			func(req llm.Request) llm.Response {
+				return wrapCommunicateResponse(llm.Response{
+					Message: llm.Message{
+						Role:    llm.RoleAssistant,
+						Content: []llm.ContentPart{{Kind: llm.ContentText, Text: "done"}},
+					},
+					Finish: llm.FinishReason{Reason: "stop"},
+					Usage:  llm.Usage{InputTokens: 10, OutputTokens: 5, TotalTokens: 15},
+				})
+			},
+		},
+	}
+	c.Register(f)
+
+	profile := resolveTestProfile("gw", nil, "gpt-5.2")
+	if profile.SupportsWebSearch() {
+		t.Fatal("pre-condition: gw must not resolve web_search (its base_url diverges from openai's default)")
+	}
+
+	sess, err := NewSession(c, withTestSessionNamer(c, profile), execenv.NewLocalExecutionEnvironment(dir), SessionConfig{})
+	if err != nil {
+		t.Fatalf("NewSession: %v", err)
+	}
+	defer sess.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second) // TRIPWIRE: scripted in-process adapter, no real I/O; only fires on a genuine hang.
+	defer cancel()
+	_, err = sess.ProcessInput(ctx, "hello", nil)
+	if err != nil {
+		t.Fatalf("ProcessInput: %v", err)
+	}
+
+	reqs := f.Requests()
+	if len(reqs) == 0 {
+		t.Fatalf("no requests captured")
+	}
+	if reqs[0].WebSearch {
+		t.Fatalf("WebSearch flag set on request for a base_url-diverged instance")
+	}
+}
+
 func TestSession_PauseTurn_ContinuesLoop(t *testing.T) {
 	t.Parallel()
 	dir := t.TempDir()
--- a/agent/session_model_test.go
+++ b/agent/session_model_test.go
@@ -918,7 +918,7 @@ func TestSession_WebSearch_FlagSetOnRequest(t *testing.T) {
 	}
 }
 
-// TestSession_WebSearch_FlagNotSetWhenBaseURLDiverges covers issue #738: an
+// TestSession_WebSearch_FlagNotSetWhenBaseURLDiverges pins that an
 // instance with its own base_url no longer inherits the vendor's
 // provider-level web_search (llm/registry's endpoint gate), so the request
 // this session builds must not carry the flag - a gateway that does not
--- a/agent/session_model_test.go
+++ b/agent/session_model_test.go
@@ -922,10 +922,25 @@ func TestSession_WebSearch_FlagSetOnRequest(t *testing.T) {
 // instance with its own base_url no longer inherits the vendor's
 // provider-level web_search (llm/registry's endpoint gate), so the request
 // this session builds must not carry the flag - a gateway that does not
-// implement the hosted tool rejects it and, before this fix, ended the
-// session on turn one. "gw" is the fixture's base=openai instance pointed at
-// a different base_url (profile_testhelpers_test.go); it never sets
-// web_search itself.
+// implement the hosted tool rejects it, ending the session on turn one.
+// "gw" is the fixture's base=openai instance pointed at a different
+// base_url (profile_testhelpers_test.go); it never sets web_search itself.
+//
+// This test only proves the session's own derivation (providerWebSearchEnabled
+// -> profile.SupportsWebSearch -> registry.BoolValue) reads a gated profile
+// correctly - BoolValue collapses nil and false alike, so it would pass
+// unchanged whether the registry stripped WebSearch to nil or to false. It
+// cannot, on its own, see a fail-open gap at the wire-building layer, where
+// a caller that sets Request.WebSearch directly (bypassing
+// providerWebSearchEnabled entirely - cmd/llmcall's --web-search flag, for
+// one) would hit a protocol adapter's own gate (caps.WebSearch == nil ||
+// *caps.WebSearch), which treats nil as permissive. The wire-layer leak is
+// pinned separately and directly in llm/providers/responses
+// (TestBuildBody_WebSearchNilCapsIsFailOpen); this test closes the gap at
+// its own layer instead by also asserting the resolved Caps.WebSearch
+// pointer itself is explicit false, not nil, so a registry regression back
+// to nil'ing the cap fails here even though req.WebSearch would still read
+// false.
 func TestSession_WebSearch_FlagNotSetWhenBaseURLDiverges(t *testing.T) {
 	t.Parallel()
 	dir := t.TempDir()
@@ -950,6 +965,9 @@ func TestSession_WebSearch_FlagNotSetWhenBaseURLDiverges(t *testing.T) {
 	profile := resolveTestProfile("gw", nil, "gpt-5.2")
 	if profile.SupportsWebSearch() {
 		t.Fatal("pre-condition: gw must not resolve web_search (its base_url diverges from openai's default)")
+	}
+	if ws := profile.Resolved().Caps.WebSearch; ws == nil || *ws {
+		t.Fatalf("pre-condition: gw's resolved Caps.WebSearch must be an explicit false, not nil (nil is fail-open at the adapter layer): %v", ws)
 	}
 
 	sess, err := NewSession(c, withTestSessionNamer(c, profile), execenv.NewLocalExecutionEnvironment(dir), SessionConfig{})
--- a/llm/client_registry_test.go
+++ b/llm/client_registry_test.go
@@ -413,3 +413,39 @@ func TestClientCanServe(t *testing.T) {
 		t.Fatal("unknown instances and Codex ids off the allowlist are not (spec §7.3)")
 	}
 }
+
+// TestClientWebSearchGatedInstanceSendsNoHostedTool traces the registry's
+// web_search endpoint gate end to end, through the one seam it cannot test
+// alone: the protocol builders treat Caps.WebSearch == nil as permissive
+// (each protocol's *_WebSearchNilCapsIsFailOpen test), so the registry must
+// resolve a non-first-party endpoint to an explicit false even when no
+// layer ever granted web_search (azure grants it nowhere). With that false
+// in the resolved record, a caller setting req.WebSearch = true directly -
+// never consulting registry.Caps - still produces a wire request with no
+// hosted tool, closing issue #738's crash path.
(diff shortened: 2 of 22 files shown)


obra:
All roborev findings addressed (a48ab7db5 + 882863deb), verified by an independent scoped re-review with red-checks per finding:

**High (fail-open nil — the crux):** the gate now sets explicit false and repoints Provenance to "gate/first-party". The consumer trace explains why earlier tests missed it: SupportsWebSearch() collapses nil/false, but all three protocol adapters independently treat nil as allow — proven executably (new TestBuildBody_WebSearchNilCapsIsFailOpen, plus the re-review ran throwaway equivalents for anthropic and google confirming explicit-false is fail-closed in each with Request.WebSearch=true).

**High (row-level base_url bypass):** the gate moved after buildTransport and judges the fully resolved transport.BaseURL — model-row overrides now gated (rowdiverge/vertexrow cases, red-checked by reverting to instance-only judging).

**Medium (host-var bypass):** vertexHostOverridden treats a GOOGLE_VERTEX_HOST that doesn't reproduce the canonical vertexHost(location) derivation as an override, with a verbatim-copy carve-out (vertexhostgw/vertexhostsame, red-checked). Per-var reasoning documented; Ollama needs no equivalent (no derived baseline, no web_search-carrying provider).

**Re-review Minor:** the gate no longer restates an already-false capability — provenance and warning apply only when the gate actually changes the value (red-checked via a diverged bedrock instance inheriting false from a curated row glob).

Goldens: same two fixtures only, now explicit false + gate provenance.

obra:
Round addressed (4bcb8bd89), closing the field class rather than the one flagged field.

**Medium (endpoint redirects):** the gate now judges every Transport field that changes where a tool-carrying body lands — BaseURL, Endpoint, StreamEndpoint, and CountTokensEndpoint (verified: all three adapters' CountTokens reuse the completion-body builder verbatim and none strip tools, so count-tokens requests carry web_search too). Ruled out with evidence: ModelsEndpoint (bodyless GET), Preset (expanded at load), HostRule/Vars (already reflected in resolved BaseURL; Vertex host-var separately gated), Auth fields (how, not where), Body (payload, not routing). Red-checked: endpointgw/streamendpointgw/counttokensendpointgw all wrongly kept web_search pre-fix; endpointsame pins the verbatim carve-out. Implementing this surfaced a real regression the naive comparison would have shipped — google-vertex/claude-opus-5's row legitimately carries the vertex-anthropic preset's own endpoints (models.dev conversion), misread as a redirect — fixed by extending the existing curated-trust boundary: a curated record whose base_url checks out is trusted for its own endpoint data.

**Low (trailing slash):** mirrored the builder's exact normalization (protocolhttp URL()'s TrimRight on BaseURL only — deliberately not a broader canonicalization) on both sides of the comparison; the OPENAI_BASE_URL with trailing slash now keeps web_search, red-checked. All pre-existing gate cases green throughout; no golden changes.

obra:
Round addressed (7b314bcd4), and the branch is now rebased onto current main.

**Medium (inherited Vertex row transport):** confirmed red the OPPOSITE way on live code — a legitimate vars-only `base = "google-vertex"` instance resolving a Claude row was being wrongly STRIPPED (the comparison judged against a row-less transport that falls to the generic /messages endpoint instead of the row's curated vertex-anthropic rawPredict preset). The gate now compares against the curated provider's own matching model row, so legit inherited-row resolutions keep web_search and genuine redirects still strip. Red-checked both directions.

**Medium (config/ provenance too broad):** enumerated every provenance form — {snapshot,cache,overlay,config}/{provider,row,glob:pattern} plus alias/live (structurally unreachable for WebSearch per spec §4.2). Only `config/provider` and `config/row` — a deliberate, individually-named opt-in — bypass the gate now; every glob form is stripped with the warning. Red-checked: a provider-scoped glob web_search=true on a diverged instance bypassed pre-fix, stripped after; both explicit forms still bypass.

**Lows:** temporal/personal comment phrasing removed and firstPartyEndpoint's doc trimmed to invariants (case narrative lives in the test's desc strings); the Vertex-specific host-var scope is now documented plainly as a per-provider necessity.

19-case gate table green throughout; no golden changes.

obra:
Round addressed (9feae265f):

**High (host-rule swap):** confirmed the swap disabled the check entirely — vertexHostOverridden's first guard compared the record's own (swapped) rule against the Vertex constant, returning false before ever inspecting the malicious host. Fixed both ways per the review: a host_rule differing from the curated base's own rule is divergence by itself, AND the resolved host is validated directly against vertexHost(LOCATION) rather than trusted through the record's rule. Red-checked (vertexhostrule wrongly kept web_search pre-fix).

**Medium (same-provider alias):** red the opposite way — a legit google-vertex alias of claude-opus-5 was falsely stripped (the gate looked up the user-chosen alias name in the curated model map, found nothing, fell back to a row-less wrong baseline). canonicalRowID now switches to the alias target's id exactly when the row imports the target's transport (the same condition gating the import); a diverged alias instance still strips via the base_url check. Red-checked both directions; TestResolve_AliasSeeding green.

**Lows:** inline comment added for the model-less ResolveInstance no-op; the defaults computation stays (it's a return value of the same resolveBaseURLWith call — deferring it would mean calling twice), documented at the call site. The review's test type-assertion nit is acknowledged and deferred as test-cosmetic.

22-case gate table green, no goldens changed, branch current with main.

obra:
Round addressed (2a620dfa5):

**Medium (merged-row false positive):** rowOwnBaseURL now checks only the user's LayerConfig layer's own row entry — the same exact-row origin check the codebase already uses beside it (aliasFromConfig) — never the merged row, so inherited curated Vertex MaaS row URLs fall through to the canonical row-transport comparison. Red-checked with a dedicated overlay test reproducing the MaaS row shape (web_search = false, want true, pre-fix).

**Medium (prefix-check false negative):** vertexHostOverridden now url.Parses the raw host variable and requires scheme://authority to equal vertexHost(location) exactly, with user/path/query/fragment all empty — the /proxy suffix, port, and userinfo variants are each covered (path variant confirmed genuinely red pre-fix).

**Class sweep:** both files re-swept post-fix for the two recurring axes (merged-vs-origin, string-vs-parsed) — no third instance; remaining string ops are the symmetric slash normalization, the already-origin-tracked credential path, or unrelated machinery.

Full registry+llm suites green, goldens unchanged.

obra:
Round addressed (e27b01874) — by redesign rather than a seventh round of channel-chasing. Six rounds established that enumerating config override channels (instance/env/row/glob base_urls, record- and row-level host_rules, host vars, alias indirection, parse edges) cannot converge: config composition makes the channels combinatorial.

The gate now proves first-party **by construction**: it replays the provider's CURATED layers only (snapshot/cache/overlay — no user config, no environment) through the same glob/row/transport machinery the real resolution uses, yielding what the vendor's own data says the endpoint is, then requires componentwise equality (parsed scheme/authority/path/query/fragment on base_url + exact Endpoint/StreamEndpoint/CountTokensEndpoint) against the ACTUAL resolved transport. Template vars are classified by position: path vars (project, location, model) pass through verbatim on both sides; authority-position vars take only a curated default or a host-rule derivation (vertexHost(location)) — with neither, the canonical side cannot be proven and the gate FAILS CLOSED.

All five round-6 findings reproduce red pre-redesign and pass now (host-without-location, row host_rule swap, glob base_url on vars-only providers, alias-target override, bare ?/# edges) — and every one of the 25 prior behavior pins passes unchanged; no golden drift. Deleted as superseded: vertexHostOverridden, rowOwnBaseURL, the ownOverride threading (spec §10's credential stop still uses rec.ownBaseURL). Kept: explicit config/provider+config/row opt-in, the warning, provenance handling.

⚠️ One deliberate posture note for curation: a provider whose template embeds a non-derivable authority var (Azure's resource name, Bedrock's region-in-host) fails closed if it ever gains curated web_search — the explicit opt-in is the escape hatch.

obra:
Round addressed (116bbd5c2 + 1f7b36724):

**Medium (nil fail-open):** gateWebSearch now normalizes nil → explicit false on a non-first-party endpoint (silently — the warning and provenance repoint stay reserved for overturning an explicit/inherited true; first-party nil stays nil; the already-false early return stands). Red-checked at two layers: the registry resolution, and end-to-end against a recorded wire body — a gated instance with req.WebSearch=true set directly now emits no hosted tool.

**Medium (adapter coverage):** anthropic gained the deterministic nil-caps fail-open wire pin mirroring the responses test, and google's (previously only incidental) gate got the same named twin — the r4 re-review's throwaway executions are now durable tests.

**Low:** the opt-in test asserts the capability value itself (non-nil true), not just the absent warning.

**Goldens:** the two gated fixtures unchanged; nine others each gained exactly one silent "web_search": false — the fixtures the canonical gate already classified non-first-party whose providers grant web_search nowhere (azure/bedrock: underivable authority vars; local-*: custom URLs; ollama redirects) — verified line-by-line, no warnings or provenance entries added. This is the fail-closed posture noted earlier, now visible in fixture form.

obra:
Round addressed (7b4ffcc35):

**Medium (raw delimiters):** took the reject shape — sameEndpointURL refuses any base URL whose RAW string contains ? or # anywhere, before any comparison (the parsed component checks that reject made unreachable were removed). Justified by a full catalog audit: all 207 providers' templates, every overlay base_url, and every curated var default are delimiter-free — and that claim is now durably pinned by TestCuratedBaseURLsCarryNoQueryOrFragment, so a future curated entry needing a delimiter fails loudly before it can collide with the reject. Red-first: the bare trailing # was genuinely red (url.URL has no ForceFragment — exactly the hole); bare ? was verified already-covered by ForceQuery via a passing pin rather than assumption; all five delimiter shapes pinned.

**Medium (no curated record):** proven a live hole, not subsumed — a from-scratch [providers.custom] resolved web_search=nil on both paths and a glob-granted true rode through, because firstPartyEndpoint returned true at the no-curated-record branch before r7's normalization could act (the local-* fixtures were r7's path via the curated openai-compatible base, which is why this branch needed its own proof). Now: no curated identity = not first-party — nil normalizes to silent explicit false, a config-glob true strips with the warning and provenance repoint (consistent with the glob-never-bypasses rule), and the explicit opt-in survives silently. The warning also no longer prints an empty provider name on this path.

Zero golden changes; full llm suite green.

obra:
Round addressed (4b297ec29):

**Medium (host-rule-derived location poisons both sides):** closed at both layers. canonicalVarLookup's authority set is now the union of template-position vars AND every var the transport's host rule consumes to derive the authority (vertex-location → HOST + LOCATION; ollama-host → OLLAMA_HOST) — an authority-affecting value is admitted from the actual resolution only when provably harmless, which for a Vertex location means a single RFC-1123 hostname label (letters/digits/interior hyphens, ≤63 bytes; no dots, slashes, colons). The vertex-location rule itself also refuses to derive a host from a non-label (with an 'invalid GOOGLE_VERTEX_LOCATION' warning), so the poisoned URL is never built on the actual side either. Red-first reproduced the reviewer's exact trace: evil.com/, evil.com, and us-central1:8443 all kept web_search with no warnings pre-fix (both sides derived the same poisoned host); all three now strip with both warnings, us-central1 untouched.

**Other-host-rules audit:** ollama-host has no equivalent vector — OLLAMA_HOST resolves canonically from the curated localhost default (never the user's value), OLLAMA_BASE_URL is severed by the canonical resolution's empty environment, and the actual side already validates its URL thoroughly. No other rule derives authority from a var.

Zero golden changes (every fixture location is label-shaped); full llm suite green.

PR #823, opened 2026-09-02
feat(agent): seed delegate task lists; the brief parameter is prompt

obra:
## What

`delegate` gets a `task_list` parameter that seeds the delegate's task list, and its brief parameter is renamed from `task` to `prompt` and given a description.

- `task_list`: array of `{title, prompt, reasoning_effort?, type?}`. Items replace the role's `parent_tasks` placeholder when its default task list has one (explorer) and follow the role's defaults otherwise (subagent, default). The list is frozen into the delegate descriptor at creation, so launch and every resume seed the child from it. The first task auto-starts and its prompt is injected, same as a root session.
- `prompt`: required, non-empty. The description states what a brief has to carry, because the delegate receives nothing else: the request quoted, known facts, owned paths, the acceptance check, the report. An empty or missing brief is now `invalid_request: prompt is required` instead of a delegate that starts with no input and sits idle.
- No alias for `task`. Internal field names (descriptor `task`, status projections) are unchanged.

## Why

The legacy `spawn_agent` tool carried `task_list` (pre-populate through the `parent_tasks` slot) until the tool surface was replaced on 2026-06-09 (`a240649ef`); `delegate` never got it and the slot has been empty since. Terminal-Bench trials show why it matters: across 259 delegations in five prompt variants, every brief was a one-paragraph string, 96-99% read-only, and the `task` parameter had no description at all. A micro-experiment series found that a prescriptive brief contract raises briefs that name an acceptance check from 6% to 61-85%.

## Prompting

- `delegation.md`: the "give the subagent enough context" advice becomes the four-part brief contract, plus one paragraph on `task_list` for multi-step units.
- `subagent` role: "if your task list is populated, it defines your workflow."
- `docs/job-control.md`: parameter mention.

## Tests

- `agent/task`: `ExpandParentTasks` (placeholder replaced, appended when absent, placeholder task kept when there are no parent tasks) and `PopulateFromTemplates` append behaviour.
- `agent/internal/tool`: schema pins `prompt` (required, described) and the `task_list` item schema; no `task`.
- `agent`: `decodeDelegateArgs` decodes `prompt` and `task_list`, rejects malformed lists and empty prompts; `createDelegate` freezes the list into the descriptor and the child's task store shows the tasks with the first in progress.
- TUI renderers and every scripted delegate call in the suites moved to `prompt`.

## Follow-ups (not in this PR)

- `delegate_send.task_list` append (the legacy `resume_agent` had it).
- Whether the `reasoning_effort` enum for delegates should include the parent's `max`.

### DIFF
--- a/agent/builtin_agents_test.go
+++ b/agent/builtin_agents_test.go
@@ -574,7 +574,7 @@ func TestSpawnAgent_DefaultSubagentGetsComposedPrompt(t *testing.T) {
 	if !strings.Contains(subagentSystemPrompt, "Delegated task limits") {
 		t.Error("default subagent prompt should contain shared delegated-task guidance")
 	}
-	if !strings.Contains(subagentSystemPrompt, "focused subagent") {
+	if !strings.Contains(subagentSystemPrompt, "one delegated unit of work") {
 		t.Error("default subagent prompt should contain subagent persona instructions")
 	}
 }
--- a/agent/delegate_legacy_dormancy_test.go
+++ b/agent/delegate_legacy_dormancy_test.go
@@ -41,7 +41,7 @@ func TestDelegateLegacyDormancy_NoDelegateJobRecordCanBeCreated(t *testing.T) {
 	call := root.reg.ExecuteCall(context.Background(), root.currentEnv(), llm.ToolCallData{
 		ID:        "legacy-dormancy-create",
 		Name:      "delegate",
-		Arguments: json.RawMessage(`{"task":"prove stable-only creation"}`),
+		Arguments: json.RawMessage(`{"prompt":"prove stable-only creation"}`),
 	})
 	if call.IsError {
 		t.Fatalf("registered delegate returned transport error: %s", call.Output)
--- a/agent/delegate_resource_create_test.go
+++ b/agent/delegate_resource_create_test.go
@@ -212,7 +212,7 @@ func TestDelegateResourceCreate_RegisteredPostCommitFailureRetainsStableIdentity
 		}
 		return nil
 	}
-	raw, err := json.Marshal(map[string]any{"task": "retain stable identity after oversized construction failure"})
+	raw, err := json.Marshal(map[string]any{"prompt": "retain stable identity after oversized construction failure"})
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -291,7 +291,7 @@ func TestDelegateResourceCreate_RegisteredResultPreservesWorktreeAndModelFallbac
 		ID:   "registered-create-metadata",
 		Name: "delegate",
 		Arguments: json.RawMessage(`{
-			"task":"preserve creation metadata",
+			"prompt":"preserve creation metadata",
 			"agent_type":"reviewer",
 			"isolation":"worktree"
 		}`),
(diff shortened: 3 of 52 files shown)


obra:
Addressed in the follow-up commit:

- **Hub frontend** (`ToolCallItem.tsx`): delegate intent preview reads `prompt`; fixtures across the transcript tests follow. Typecheck, biome, and the full vitest suite pass.
- **coordinator-workflow plugin**: the coordinator puts the brief in `prompt` and the plan's steps in `task_list`; the verifier step says "prompt".
- **`create` guard**: now `invalid_request: prompt is required` (still unreachable behind `decodeDelegateArgs`).
- **`task_list` item validation**: `reasoning_effort` ∈ {low, medium, high} and `type` ∈ {research, implement, verify, fix} rejected with `invalid_request` at decode, test-first.
- **`docs/job-control.md`**: input examples use `prompt`, the full shape shows a `task_list`; the status example keeps `task` because that is the descriptor projection field, unchanged.

Not done, pending a decision: a read-only `task` fallback in presentation code (transcript render, TUI summaries, hub intent) so delegate calls persisted before this change keep their brief preview. That is backward compatibility, which this repo takes on explicitly rather than by default.

obra:
Added ff1fb9dfe: the `subagent` role body now matches the brief contract. It described a cheap scoped helper ("do not assume the task requires code changes", "start with the narrowest action", "verify only to the level the task requires"); it now owns exactly the paths its brief names, makes the change when the brief asks for one, runs the named check before reporting and fixes within the unit if it fails, and says when a step went beyond the brief. Read-only briefs behave as before. Terminal-Bench interrogations of three sessions showed the parent, not the role, decides who writes, but once handed an implementation unit the delegate is what this text now describes.

obra:
Added 2162c1cd5 + c3534825e: `subagent` may now delegate when granted a `delegation_allowance`. The role carries `delegate`, `delegate_send`, and the job-control tools it needs to supervise children; the existing allowance gate decides whether `delegate` is actually registered (granted: yes, plus `job_watch`; default 0: a leaf, as before). Descriptions no longer call it a hard leaf.

Fixing this surfaced a latent bug on the legacy spawn path: the child's tool policy was computed from the parent's allowance rather than the child's granted one, masked until now by the root-only strip at session init. It now uses the child's allowance on every path, with tests for the granted and leaf cases.

Default allowance semantics are unchanged (parent opts in per call).

obra:
Round two, addressed in the latest commit:

- **Historical transcripts**: the transcript renderer, the TUI summary/renderer/body, and the hub intent preview read `prompt` and fall back to the retired `task` key, each with a test on an old-format record. Invocation validation stays prompt-only.
- **Shared task stores**: `createDelegate` now refuses `task_list` with `invalid_request` when the parent shares its task store with children (`ShareTasksWithChildren`), since `PopulateFromTemplates` never re-seeds a non-empty store; test covers the refusal and that a plain create still works.
- **`prompt` description**: no longer says the brief is the delegate's only input; `task_list` supplies the ordered step prompts.
- **Docs**: `11-delegate-resource-model.md` describes parent task lists frozen into the descriptor and the shared-store refusal.

Issue #825, opened 2026-09-02
TUI: set terminal window title to the active session's display name


obra:
## Summary

The webui sidebar shows each session's display name (`SessionMeta.Name`, with the `SessionDisplayName` fallback chain Name → OriginalPrompt → ID), but the terminal running `evener-tui` has no title. The session name field is human-visibility only (never fed back into agent context), so it is a natural fit for the terminal/tmux window title — it lets a user find the right pane at a glance.

## Proposal

In `cmd/evener-tui`, set the terminal window title (bubbletea `tea.SetWindowTitle`) to the active session's display name:

- On entering a session view (subscribe/attach), set the title from the session's current display name.
- On `EventSessionNameChanged`, update the title live — including the auto-namer's compaction refreshes.
- On leaving the session view (back to dashboard), restore or clear the title.

## Notes

- tmux only reflects OSC title escapes when `allow-rename on` is set for the window; document this (default `allow-rename off` means the title won't stick there).
- No code sets a window title anywhere in `cmd/evener-tui` today (no `tea.SetWindowTitle` usage), so this is additive.
- Related: the session auto-namer was just reoriented (see `agent/session_namer.go`) to name the overarching session goal rather than momentary activity, so titles fed from it are now stable session labels rather than "what was being done right this second".

## Testing

Per repo policy, verify via deterministic tests: assert the model returns a `tea.SetWindowTitle` command when the session view activates and when the name-changed event arrives (`cmd/evener-tui` already has hub-update message tests to model after).

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

**Verdict:** `not-a-bug` — the mechanism claims are accurate but there is no defect. `rg 'SetWindowTitle|WindowTitle|window title' cmd/evener-tui` returns nothing, so the TUI never sets a terminal title — it was never implemented, rather than broken. The display name is available (`schema.SessionDisplayName`, `agent/schema/snapshot.go:268`, projected as `displayName` in `appwire/types.go`), and the change is genuinely additive.

**Recommendation:** `close` — this is a UX enhancement, not a current problem, and Jesse's brief is explicit about not overbuilding from a user's proposed solution. The `tea.SetWindowTitle` wiring is small and would verify with deterministic tests, so if this is wanted it can be filed as a scoped feature task; keep it open only if terminal-title discovery is a priority.

PR #826, opened 2026-09-02
fix(agent): orient session namer prompts on the overarching session goal

obra:
## Summary

The session auto-namer titled sessions after momentary activity ("what are you doing right this second") instead of the session's overarching goal. Root cause is prompt orientation, not plumbing:

- The compaction refresh path (which overwrites the name after every compaction CHECKPOINT/SUMMARY turn) feeds the namer handoff summaries whose mandated Current State / Pending Work sections describe one instant — `defaultSummaryPrefix` literally asks "what file was being edited, what problem was being debugged."
- The namer's own system prompt said "prefer the concrete task or outcome over generic wording," which steers a temp-0 cheap call toward the most concrete, most recent thing in its input.

The name field is human-visibility only (confirmed by tracing all `naming.value` reads: it flows to `SessionMeta.Name` → `EventSessionNameChanged` → webui sidebar via `SessionDisplayName`, and never back into agent context), so it should be a durable label of the session's overall goal.

## Changes

Copy-only, in `agent/session_namer.go` (+7/−2):

- `sessionNamerSystemPrompt`: defines the title as a durable label a human uses to find the session in a list; must name the overarching goal — the problem the user wants solved across the whole session — and must not title after current activity, the file being edited, or the test being debugged.
- `sessionNamerUserPrompt` (compaction label): warns that the summary's Current State and Pending Work sections are momentary snapshots to ignore for naming, and points at the user's original request and the conversation timeline.

Routing, sanitization, source precedence (manual > compaction > prompt), triggers, and quota suppression are untouched.

## Testing

- gofmt clean, `go vet ./agent/` clean
- `go test ./agent/ -run 'SessionName|SessionNamer|Rename|NameSession' -count=1` — ok
- Full `go test ./agent/ -count=1` — ok (287.415s, exit 0)

No prose-pinning tests were added: `docs/developing-evener/testing.md` ("Prompt Prose Is Not a Test Oracle") bans asserting on prompt copy; the existing behavioral suite covers the plumbing this change leaves unchanged.

Follow-up for surfacing the name as the terminal title: #825.

### DIFF
--- a/agent/cov_w2tail_namer_test.go
+++ b/agent/cov_w2tail_namer_test.go
@@ -29,10 +29,10 @@ func (a *erroringAdapter) Stream(ctx context.Context, req llm.Request) (llm.Stre
 func TestW2Tail_nameSession_Guards(t *testing.T) {
 	profile := NewOpenAIProfile("gpt-5.2")
 
-	if _, err := nameSession(context.Background(), nil, profile, sessionNameSourcePrompt, "text", noNamerSleep); err == nil {
+	if _, err := nameSession(context.Background(), nil, profile, sessionNameSourcePrompt, "text", "", noNamerSleep); err == nil {
 		t.Errorf("nil client should error")
 	}
-	if _, err := nameSession(context.Background(), llm.NewClient(), nil, sessionNameSourcePrompt, "text", noNamerSleep); err == nil {
+	if _, err := nameSession(context.Background(), llm.NewClient(), nil, sessionNameSourcePrompt, "text", "", noNamerSleep); err == nil {
 		t.Errorf("nil profile should error")
 	}
 }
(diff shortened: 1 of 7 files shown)


obra:
@roborev-primeradiant re your medium finding on `agent/session_namer.go:155` — addressed in 57201052a.

Verification confirmed the concern: the deterministic checkpoint's budget shedding (context_manager.go ~1015–1027) drops the oldest conversation entries first, so the original user ask is the first thing lost; the session-log checkpoint carries `Original prompt:` and the Layer-4 summary's verbatim timeline usually preserve it, but the deterministic and predictive checkpoint shapes do not guarantee it.

Rather than trying to reconstruct the goal from possibly-degraded text, the refresh now carries the session's current title into the naming prompt (a `Current title:` prefix with "keep it unless it no longer names the goal"). The title is the already-distilled goal statement and is always available, so the label's goal reference always has its referent in-context — and a good prompt-derived title now survives compaction refreshes by default instead of being re-derived from momentary activity. Behavioral test: `TestSessionNameFromCompactionTurn_SendsCurrentTitleToLLM` asserts the current title crosses into the naming request on the compaction path.

The underlying shed-order problem (oldest conversation entries dropped first, discarding the original task) is now tracked in #833.

PR #835, opened 2026-09-02
fix(agent): gate session namer temperature on the resolved cheap-model row

### LINKED ISSUE #834: session namer sends temperature to models with no catalog row (Bedrock wire ids); one failed call per session
## What happens

On a Bedrock-hosted model whose wire id the catalog does not know (`bedrock/us.openai.gpt-5.6-luna`), every session starts with a failed model call:

```
[advisory] session_namer: bedrock error (status=400): responses.create failed: Unsupported parameter: 'temperature' is not supported with this model. (hint: run `evener models inspect bedrock/us.openai.gpt-5.6-luna` and set fields.temperature = false)
```

The task loop is unaffected (it never sets temperature), so the session proceeds; the cost is one dead request and a warning per session. Observed on every trial of every Terminal-Bench run in the current campaign (135 of 135 per run).

## Why

- `agent/session_namer.go` sets `Temperature` unconditionally on the naming request (`Temperature: &temp`, ~line 76).
- `llm/providers/responses/request.go` emits `body["temperature"]` whenever the request carries one; pruning happens only when the resolved model row marks the field unsupported (`llm/registry/prune.go`, catalog `temperature` capability in `llm/registry/modelsdev.go`).
- `us.openai.gpt-5.6-luna` is a Bedrock wire id with no catalog row and no `alias_of`, so the row has no facts, nothing is pruned, and the namer treats the 400 as a failure rather than a pruneable parameter. The same missing row shows up as "Startup model snapshot ... has no verified choices" in the delegate tool's model description.

## Proposed fixes

1. **Namer**: do not send `temperature` when the resolved row's temperature capability is false or unknown, or retry once without it on a `KindInvalidRequest` naming that parameter. Small and unit-testable.
2. **Bedrock id resolution**: resolve `us.openai.<model>-…` / `us.anthropic.<model>-…` wire ids to their catalog row when no explicit `alias_of` is configured (the already-scoped "Bedrock ID alias in LookupModelInfo" item). Fixes the empty model snapshot as well.
3. Related startup noise in the same sessions, one failed call each: `models.list` against Bedrock's OpenAI-compatible endpoint (404, no `/models` there), and a probe of `http://localhost:11434/v1/models` from the built-in ollama provider even when no configuration mentions ollama (`llm/registry/hostrules.go` default). Worth suppressing for headless runs.

## Workaround

Per `docs/llm-providers.md`, declare the alias on the row:

```toml
[providers.bedrock.models."us.openai.gpt-5.6-luna"]
alias_of = "gpt-5.6"
```

or pin the field on the provider: `[providers.bedrock.fields] temperature = false`.

## Repro

Any `evener run` against a `base = "openai"` provider with a custom `base_url` and a model id absent from the catalog, with the session namer enabled; the first request in `sessions/<root>.api.jsonl` is the failed namer call.

obra:
## Summary

Fixes #834 (the namer-side half — proposed fix 1). **Stacks on #826** (branch `namer-temperature-gate` is based on `session-auto-namer` HEAD; merge #826 first).

## Problem

The session namer set `Temperature` unconditionally. On a Bedrock-hosted model whose wire id has no catalog row (`bedrock/us.openai.gpt-5.6-luna`), the row resolves **synthesized**; the protocol baseline marks temperature send-by-default, nothing prunes it, and the provider 400s — one dead request and a warning per session (135 of 135 Terminal-Bench trials per run).

## Fix

Two layers, both in `agent/session_namer.go`:

1. **Caps gate**: `sessionNamerTemperature` resolves the cheap ref (`CheapProvider() + "/" + model`) through the client's registry and returns a temperature **only** when a real catalog row vouches for it (`Fields["temperature"] == true`, `Synthesized == false`). Synthesized rows, failed resolutions, and false-capability rows all omit the parameter. Unknown capability reads as *do-not-send*: a wrong send is a provider 400, a wrong omit only loses a determinism hint on a low-stakes decoration call.
2. **Single retry (defense in depth)**: when a row vouches for temperature but the provider rejects it anyway (`KindInvalidRequest` naming temperature), retry exactly once with the parameter dropped instead of failing the naming attempt.

## What's out of scope (per the issue)

- Fix 2 (Bedrock `us.openai.*` / `us.anthropic.*` wire-id resolution to catalog rows) — the already-scoped "Bedrock ID alias in LookupModelInfo" registry item; that fixes the empty model snapshot too.
- Fix 3 (unrelated startup noise: `models.list` 404 on Bedrock, the default ollama probe).

## Testing

Behavioral, per repo policy (scripted provider adapter, hermetic registry; no prompt-prose assertions). New `agent/session_namer_temperature_test.go`, TDD red→green (2 of 5 failed before the fix):

- synthesized-row model (hermetic bedrock-shaped instance, exactly the #834 shape) → no temperature in the request, exactly 1 call
- catalog row with `fields.temperature=false` (moonshotai) → no temperature
- catalog row with `temperature=true` (openai) → temperature sent
- temperature-naming 400 from a vouching row → exactly one retry, without the parameter
- unrelated 400 → no retry, error surfaces wrapped

Gates: gofmt clean, `go vet ./agent/` clean, namer suite ok, `make lint-evenerfuzz` PASS (65s), full `go test ./agent/` ok (234s), `make fuzz` exit 0.

The workaround from the issue (`alias_of` / `fields.temperature = false` config) remains valid and now composes with the gate.

### DIFF
--- a/agent/session_namer.go
+++ b/agent/session_namer.go
@@ -15,6 +15,7 @@ import (
 	"primeradiant.com/evener/agent/provider"
 	"primeradiant.com/evener/agent/schema"
 	"primeradiant.com/evener/llm"
+	"primeradiant.com/evener/llm/registry"
 )
 
 const (
@@ -143,21 +144,20 @@ func sessionNamerModel(profile *provider.Profile) string {
 
 // sessionNamerTemperature returns the naming call's temperature setting: a
 // pointer to 0.0 when the resolved cheap-model row vouches for the parameter,
-// nil when it does not or cannot vouch. The unknown cases — a synthesized row
-// (no catalog row, no live listing) or an instance that fails to resolve —
-// carry no facts, and an unknown capability must read as "do not send" rather
-// than "send": a wrong send is a provider 400 (issue #834), while a wrong omit
-// only loses a determinism hint the naming prompt's low-stakes output barely
-// uses.
+// nil when it does not or cannot vouch. Unknown capability must read as "do
+// not send" rather than "send": a wrong send is a provider 400 (issue #834),
+// while a wrong omit only loses a determinism hint the naming prompt's
+// low-stakes output barely uses. The voucher is registry.TemperatureSupported,
+// which requires a catalog row (a live-only or synthesized row inherits the
+// protocol's send-by-default baseline without any catalog fact) and reads the
+// protocol-keyed temperature path, so Google rows are judged on
+// generationConfig.temperature.
 func sessionNamerTemperature(client *llm.Client, profile *provider.Profile, model string) *float64 {
 	if client == nil || profile == nil || model == "" {
 		return nil
 	}
 	res, err := client.Resolve(profile.CheapProvider() + "/" + model)
-	if err != nil || res.Synthesized {
-		return nil
-	}
-	if !res.Caps.Fields["temperature"] {
+	if err != nil || !registry.TemperatureSupported(res) {
 		return nil
 	}
 	temp := 0.0
@@ -165,10 +165,17 @@ func sessionNamerTemperature(client *llm.Client, profile *provider.Profile, mode
 }
 
 // isTemperatureUnsupported reports whether err is a rejected-request error
-// whose message names the temperature parameter (e.g. Bedrock's "Unsupported
-// parameter: 'temperature' is not supported with this model").
+// rejecting the temperature parameter specifically (e.g. Bedrock's "Unsupported
+// parameter: 'temperature' is not supported with this model"). Matching a bare
+// "temperature" substring would over-retry on unrelated invalid-request
+// messages that merely mention the word, so this matches the
+// unsupported-parameter phrasing with the parameter in quotes.
 func isTemperatureUnsupported(err error) bool {
-	return llm.Kind(err) == llm.KindInvalidRequest && strings.Contains(err.Error(), "temperature")
+	if llm.Kind(err) != llm.KindInvalidRequest {
+		return false
+	}
+	msg := strings.ToLower(err.Error())
+	return strings.Contains(msg, "unsupported parameter") && strings.Contains(msg, "'temperature'")
 }
 
 func configuredSessionNamerModel(profile *provider.Profile) string {
--- a/agent/session_namer.go
+++ b/agent/session_namer.go
@@ -165,17 +165,23 @@ func sessionNamerTemperature(client *llm.Client, profile *provider.Profile, mode
 }
 
 // isTemperatureUnsupported reports whether err is a rejected-request error
-// rejecting the temperature parameter specifically (e.g. Bedrock's "Unsupported
-// parameter: 'temperature' is not supported with this model"). Matching a bare
-// "temperature" substring would over-retry on unrelated invalid-request
-// messages that merely mention the word, so this matches the
-// unsupported-parameter phrasing with the parameter in quotes.
+// rejecting the temperature parameter specifically. Providers word this
+// differently — Bedrock's "Unsupported parameter: 'temperature' is not
+// supported with this model", OpenAI's "Unknown parameter: 'temperature'",
+// "unknown_parameter: temperature" — so this accepts the two rejection
+// phrasings (with or without the underscore) plus the parameter name, quoted
+// or not. Matching the parameter name alone would over-retry on unrelated
+// invalid-request messages that merely mention the word, so the rejection
+// phrase is required too.
 func isTemperatureUnsupported(err error) bool {
 	if llm.Kind(err) != llm.KindInvalidRequest {
 		return false
 	}
 	msg := strings.ToLower(err.Error())
-	return strings.Contains(msg, "unsupported parameter") && strings.Contains(msg, "'temperature'")
+	rejection := strings.Contains(msg, "unsupported parameter") ||
+		strings.Contains(msg, "unknown parameter") ||
+		strings.Contains(msg, "unknown_parameter")
+	return rejection && strings.Contains(msg, "temperature")
 }
 
 func configuredSessionNamerModel(profile *provider.Profile) string {
--- a/agent/session_namer.go
+++ b/agent/session_namer.go
@@ -63,24 +63,37 @@ func nameSession(ctx context.Context, client *llm.Client, profile *provider.Prof
 		return sessionNameResult{}, errors.New("session namer: model is empty")
 	}
 	maxTokens := 80
-	temp := 0.0
+	// #834: gate temperature on the resolved cheap-model row. A Bedrock-style
+	// wire id with no catalog row resolves synthesized; the protocol baseline
+	// marks temperature send-by-default, so nothing prunes it and the provider
+	// 400s — one dead request per session. Naming is deterministic decoration,
+	// so omitting the parameter on doubt costs nothing.
+	temp := sessionNamerTemperature(client, profile, model)
 	// Naming is best-effort decoration with its own short deadline. It opts out
 	// of the turn retry wall budget so a naming storm does not add load while a
 	// real turn waits on the same provider bucket.
 	namerPolicy := llm.DefaultRetryPolicy()
 	namerPolicy.RateLimitWallBudget = 0
-	res, err := llm.GenerateObject(callCtx, llm.GenerateObjectOptions{
+	opts := llm.GenerateObjectOptions{
 		Client:      client,
 		Provider:    profile.CheapProvider(),
 		Model:       model,
 		System:      new(sessionNamerSystemPrompt),
 		Prompt:      new(sessionNamerUserPrompt(source, text, currentTitle)),
-		Temperature: &temp,
+		Temperature: temp,
 		MaxTokens:   &maxTokens,
 		Sleep:       sleep,
 		RetryPolicy: &namerPolicy,
 		Schema:      sessionNameSchema(),
-	})
+	}
+	res, err := llm.GenerateObject(callCtx, opts)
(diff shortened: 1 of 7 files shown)


obra:
All three findings verified by direct probes against the registry and fixed in b142bc6be (branch also rebased onto merged main):

**Medium (live-only rows bypass the gate) — confirmed.** A probe with `ApplyLive` showed a live-listed model with no catalog row resolves `synth=false` with a baseline-seeded `Fields[temperature]=true` — the same 400 class as the original issue. Fixed by requiring catalog-row provenance: the new `registry.TemperatureSupported(res)` reads the `model` provenance step (row/region/dated) and rejects live and synthesized rows, so only a catalog row that vouches for the parameter sends it.

**Low (Google key) — confirmed.** Google rows key Fields on `generationConfig.temperature`; the fixed `temperature` key was a missing entry, dropping the parameter for every Google model. Fixed by `registry.TemperaturePath(protocol)`, which resolves the prunable path from the protocol; `TemperatureSupported` reads that path.

**Low (fragile substring) — agreed.** The retry predicate now requires the unsupported-parameter phrasing with the quoted parameter name ('unsupported parameter' + "'temperature'", case-insensitive) instead of a bare substring, so unrelated invalid-request messages that merely mention the word no longer trigger a retry.

New tests cover the two gate holes: a live-only row (via `ApplyLive`) omits temperature, and a supported Google row (gemini) keeps it — seven temperature tests total. Agent and registry suites, `make lint-evenerfuzz`, and `make fuzz` are green.

obra:
All four findings verified and fixed in eeb9ef1bc:

**Medium (Sampling ignored) — confirmed by probe, fixed.** `openai/o3` (a real models.dev `temperature=false` row) resolves `sampling=false, fields[temperature]=true`: models.dev's temperature=false maps to `Caps.Sampling=false` while Fields stays at the send-by-default baseline, so `TemperatureSupported` reported the opposite of the row's catalog fact and of `ShapeRequest`'s own pruning. It now returns false when `Sampling` is explicitly false, before the Fields check.

**Low, all addressed:** the provenance-step guard is parenthesized; `TestTemperaturePath` and `TestTemperatureSupported` now live in `prune_test.go` (12 cases covering row/region/dated/live/synthesized/absent provenance, Sampling on/off, the Google `generationConfig.temperature` key spelling, and unknown protocol); the retry predicate accepts `unknown parameter` / `unknown_parameter` phrasings and an unquoted parameter name while still requiring both a rejection phrase and the parameter name — pinned by `TestIsTemperatureUnsupported` (7 cases, including a rejected `top_p` and a bare 'temperature value out of range' non-match).

Gates: agent suite ok (305s), registry suite ok, `make lint-evenerfuzz` PASS, `make fuzz` exit 0.

obra:
Fixed in d59b58037, taking the preferred option (share the classifier's detection):

The rejected parameter was already computed by the classifier but only fed the hint. It's now exposed: `httpBaseError` carries `rejectedParam` (populated by both `ClassifyHTTPError` — structured `error.param` plus `parameterNameFromMessage` — and `ErrorFromHTTPStatus`), with an exported `llm.RejectedParameter(err)` walking the chain via `errors.As`. The namer predicate collapses to `Kind(err) == KindInvalidRequest && RejectedParameter(err) == "temperature"` — no provider prose is matched in the namer, so every shape the classifier recognizes (including `Unrecognized request argument supplied:`, `unknown field`, and structured `unsupported_parameter` errors) now triggers the retry.

One deliberate contract choice, pinned in the tests: `RejectedParameter` reports the named parameter even when the error classifies as something else (e.g. a 429 whose message quotes a rejection) — it's a fact accessor, and the caller's `Kind` gate does the gating.

Coverage: `TestRejectedParameter` in the llm package and the rewritten namer predicate test driving the same shapes through `ClassifyHTTPError`. Gates: agent suite ok (270s), llm ok, registry ok, `make lint-evenerfuzz` PASS, `make fuzz` exit 0.

obra:
All five round-4 findings verified and fixed in 1eebb47ec:

**ErrorFromHTTPStatus ignored structured error.param + classifier structured path skipped message fallback** — fixed together with a single shared `rejectedParameter(raw, message)` helper (structured param precedence, message patterns as fallback) serving both constructors. New tests cover a structured param through `ErrorFromHTTPStatus`, structured-over-message precedence, and the classifier's null-param message fallback.

**Retry matched only "temperature", not the Google protocol key** — the predicate now compares `RejectedParameter` against the rejected protocol's own temperature path via `llm.ErrorProtocol(err)`: Google errors require `generationConfig.temperature`, unattributed errors accept either spelling, other protocols the plain one. Covered positive and negative.

**row: provenance conflated catalog and user-configured rows** — this needed one design correction to the suggested fix: models.dev states temperature facts only when *unsupported*, so legitimate catalog rows (`openai/gpt-4.1-nano`, `google/gemini-2.5-pro`) carry baseline `Fields.temperature=true` with **no** provenance — requiring explicit Fields provenance would drop temperature for every known model. The distinguishing signal is row origin: catalog rows carry snapshot/overlay-attributed facts; user-configured rows carry config-attributed or none. `TemperatureSupported` now requires a row lookup plus (catalog-backed OR explicit `Fields.` provenance), so fact-less custom models read as unsupported while an explicit `fields.temperature = true` from any layer still vouches. Registry table tests add user-row cases both ways.

**Retry test's missing first-call assertion** — added: the first request must carry temperature or the test fails with "capability gate broken, retry not exercised".

Gates: llm ok, registry ok, make fuzz exit 0, agent suite green across two consecutive full runs (410s and 566s; one earlier FAIL never reproduced and is reported in the commit message as a flake).

obra:
Round 5 fixed in 3caf1a161 (and apologies for the slow pickup on this round — the watch on this PR keyed on comment ids, but roborev edits its existing comment between rounds, so the edit never fired it; the watch now keys on the reviewed commit sha):

**Medium (inherited provider facts vouch) — confirmed by probe, real hole.** A user-defined model on the openai base carries fourteen overlay/provider-attributed facts (StrictTools, MaxTokensField, ReasoningSummary, ten Fields.*), so catalogBacked read the custom row as catalog-backed and the gate sent baseline temperature to it — the exact #834 failure class. catalogBacked now counts only model-row catalog facts ("<layer>/row" for snapshot, overlay, or cache), excluding provider- and glob-attributed ones. End-to-end probe after the fix: user custom row on openai base → unsupported; openai and google catalog rows → supported; sampling=false row → unsupported.

**Low (cache/row)** — cache-backed model rows now count as catalog-backed. New table case.

**Low (test comment)** — corrected: the moonshotai temperature=false fact lives in the models.dev snapshot and the derivation maps it to Caps.Sampling=false, which TemperatureSupported gates on — not overlay fields.

Gates: agent suite ok (373s), registry ok, lint-evenerfuzz PASS, make fuzz exit 0.

obra:
Round-6 Low fixed in 4a2a5b470, along with the lint failure the same push surfaced:

**RejectedParameter lost on non-400/422 classifications** — confirmed: `ClassifyHTTPError` populated `rejectedParam` only on the 400/422 paths, while `ErrorFromHTTPStatus` preserved it, contradicting the accessor's documented coverage. `rejectedParam` is now initialized when the base error is constructed, which also made the two per-path assignments redundant (both removed). New test: a 429 through `ClassifyHTTPError` with a rejection-shaped body keeps the parameter.

**Lint** — the modernize linter flagged `catalogBacked`'s `HasSuffix`+`TrimSuffix`; replaced with `strings.CutSuffix`. `make lint-golangci` PASS.

Gates: make lint-golangci PASS (85s), agent suite ok (252s), llm ok, registry ok, make lint-evenerfuzz PASS.

obra:
Round-7 Medium fixed in 6ee63dbb3:

**Valid finding** — `rejectedParameter` trusted any `error.param` on an invalid request, but the field is general: providers also use it to point at the offending parameter in invalid-value errors ("Invalid value for 'temperature': must be between 0 and 2", `code: invalid_value`), where dropping the parameter is wrong — the namer would retry without temperature while the real problem went unaddressed.

**Fix** — `rejectedParameter` now takes the error code and honors the structured param only when the error is actually a rejection: code `unknown_parameter`/`unsupported_parameter` or a rejection-shaped message. Any other param is ignored; the message patterns remain the fallback. Both constructors pass the code through.

**Tests** — invalid-value negatives through both `ErrorFromHTTPStatus` and `ClassifyHTTPError` (plain and Google nested spelling) must not trigger the retry; rejection shapes re-anchored with codes; precedence kept for rejection-shaped messages.

Gates: agent suite ok (249s), llm ok, registry ok, make lint-golangci PASS, make lint-evenerfuzz PASS, make fuzz exit 0.

obra:
Round-8 Lows fixed in b4720f03c, both verified by probe first:

**Catalog aliases** — confirmed: `claude-sonnet-4-5[1m]` resolves with its target's facts seeded under "alias" provenance, which catalogBacked ignored — a fully supported catalog model lost deterministic temperature. Alias-seeded facts now count as catalog-backed: a pure catalog alias vouches what its catalog target does. Probe after the fix: alias row → supported.

**Explicit sampling=true** — confirmed: a custom row's explicit `sampling = true` carried config/row provenance but was not a voucher. An explicitly sourced Sampling=true now vouches directly (the Sampling=false override is unchanged); a sampling value without an explicit source still reads as baseline silence. Probes: sampling=true → supported, sampling=false → unsupported.

Three new registry table cases (alias-inherited facts, explicit sampling=true vouches, inherited-value-without-source does not). Gates: registry ok, agent suite ok (297s), lint-golangci PASS, lint-evenerfuzz PASS.

PR #836, opened 2026-09-02
make: run lint-gofmt with the active toolchain's gofmt

obra:
`lint-gofmt` called bare `gofmt`, assuming PATH's formatter matches the compiler. It doesn't have to: go.mod's toolchain line makes `go` auto-switch to 1.27.0, but the standalone `gofmt` binary on PATH is whatever the distro installed (go1.22 on Ubuntu 24.04, which ships nothing newer than 1.24). The stale formatter flagged three clean files, so `make lint` failed on a clean `main`.

This resolves gofmt through `go env GOROOT`, so the target always runs the formatter that matches the compiler, locally and in CI. One recipe line plus a comment explaining why.

Verification on the affected machine:
- `make lint-gofmt` on clean `origin/main`: FAIL (three files) before, PASS after.
- A deliberately unformatted tracked file still fails the target.
- `make lint-generated` clean (annotation lines untouched).

### DIFF
--- a/make/linting.mk
+++ b/make/linting.mk
@@ -134,8 +134,12 @@ lint-cache-clean:
 ## trigger: Required CI (via make lint); local pre-merge.
 ## requires: None beyond the Go toolchain.
 ## fails-when: Any tracked .go file is not gofmt-clean.
+# Runs the active toolchain's own gofmt rather than whatever gofmt is on PATH:
+# go.mod's `go` directive (with GOTOOLCHAIN=auto) makes `go` switch to a newer
+# toolchain, but a distro gofmt binary never switches, so PATH's gofmt can lag
+# the compiler and disagree with it about formatting.
 lint-gofmt:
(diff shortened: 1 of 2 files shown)


obra:
Round addressed: the comment (and the commit message) misattributed the auto-switch to a `toolchain` directive. go.mod has only `go 1.27.0`; the switch comes from that directive under GOTOOLCHAIN=auto. Both reworded accordingly. lint-gofmt and lint-generated re-run clean.

obra:
CI failure was not a race: TestMakeLintRunsEveryGateInLintTargets pins each lint gate to a substring its recipe must run, and lint-gofmt was pinned to `gofmt -l`, which the quoted GOROOT invocation no longer contains verbatim. Repointed the pin at `/bin/gofmt" -l` so the dry-run check still bites. Root package tests green locally with pipefail. Amended into the single commit (8f6915192).

PR #838, opened 2026-09-02
fix(agent): normalize neutral retained transcript options

### LINKED ISSUE #827: Normalize semantically empty read_transcript options before ref-mode validation
Priority: P0

## Problem

`read_transcript` selects different behavior for session, job, and artifact refs, then validates options by key presence. Model/provider tool calls often materialize optional defaults, so semantically empty values select or violate a mode.

In session `034HvTCI5LrwbM2ZZpBMqN`, all 20 `read_transcript` calls failed. The model repeatedly tried to omit `range`, but job-ref calls still arrived with neutral fields such as `range:""`, `expand_turn:0`, and `format:"markdown"`. `validateJobReadArgs` rejected the calls with `range applies only to session transcript refs`.

Relevant code:
- `agent/internal/tool/definitions.go`: `DefReadTranscript`
- `agent/session_tools_transcript.go`: `parseRetainedReadArgs`, `validateJobReadArgs`, `validateArtifactReadArgs`
- `agent/session_tools_transcript_job_read_test.go`
- `agent/session_tools_transcript_branches_test.go`

There is also a contract mismatch: the tool description says any explicit format is invalid for job paging/search, while implementation and tests permit explicit `format:"markdown"` as a no-op.

## Proposed behavior

Normalize semantically empty optional values before ref-kind and mode validation.

For job and artifact refs, treat these as omitted:
- `range: null | ""`
- `expand_turn: null | 0`
- `output_match: ""`
- `context_lines: 0` when no non-empty `output_match` exists

For job refs only, also treat `format: null | "" | "markdown"` as omitted. Artifact refs intentionally reject every explicitly supplied format; retain that restriction.

Continue rejecting:
- non-empty session ranges on job/artifact refs;
- nonzero session expansion on job/artifact refs;
- `outline` and `jsonl` on job/artifact refs;
- invalid paging/search bounds, regexes, and retention requests.

Do not silently normalize `offset_bytes:0`: presence currently selects raw paging, so zero is ambiguous. Either retain its current meaning or add an explicit mode such as `auto | page | search | expand`. If the union remains hard for providers to generate, consider separate session-transcript and retained-output tools in a later compatibility phase.

Update the description to document the existing `format:"markdown"` exception.

## Diagnostics

When validation still fails, report all incompatible fields, their received values, and one minimal corrected call. The current error names `range` without showing that it arrived as an empty/default value, which led to repeated ineffective retries.

## Acceptance criteria

- Neutral values above behave exactly like omission for job and artifact refs.
- Meaningful session-only options remain rejected outside session refs.
- Artifact refs continue rejecting every explicitly supplied `format`, including null, empty, and `markdown`.
- Existing session, job paging, job search, and artifact behavior remains unchanged.
- `format:"markdown"` behavior and documentation agree.
- Errors include the incompatible fields and a minimal valid example.
- Normalization is observable in repair/diagnostic telemetry.

## Tests

Add table-driven tests for absent, null, empty, default, and meaningful values across session, job, and artifact refs. Include regression coverage for the exact default-materialized shape observed in session `034HvTCI5LrwbM2ZZpBMqN`.

obra:
Closes #827

## Problem
Provider-materialized neutral `read_transcript` options were validated by key presence, causing retained job and artifact reads to reject semantic defaults. The job tool documentation also contradicted the existing `format:"markdown"` compatibility behavior.

## Solution
- Normalize neutral job/artifact retained-read options before operation and ref-mode validation.
- Preserve explicit `offset_bytes:0` as raw-page selection.
- Aggregate remaining incompatible fields, their received values, and a minimal corrected call in repair diagnostics.
- Document that job `format:"markdown"` is a no-op for default, page, and search views.

## Retained restrictions
- Artifact refs reject every explicitly supplied `format`, including `null`, empty, and `markdown`.
- Non-empty ranges and nonzero expansion remain session-only.
- `outline` and `jsonl` remain invalid for jobs/artifacts.
- Paging/search validation, bounds, regex, retention, and output shapes are unchanged.

## Tests run
- `go test ./agent -run '^(TestReadTranscriptNeutralMaterializedRetainedOptions|TestNormalizeRetainedReadArgsNeutralValues|TestNormalizeRetainedReadArgsPreservesExplicitZeroOffset|TestRetainedReadIncompatibleArgsDiagnostics|TestArtifactExplicitFormatsRemainRejected)$' -count=1 -v`
- `go test ./agent -run '^(TestReadTranscript|TestValidateArtifactReadArgs|TestValidateJobReadArgs|TestParseRetainedReadArgs|TestTranscriptToolDefinitions|TestReadTranscriptDescription)' -count=1`
- `go test ./agent -count=1` (passed in 386.454s)

Repository-wide make gates were intentionally not run in this first implementation wave.


obra:
Review follow-up at `89082441eb62e2c234d6263966dbcb2f385974dd`:

- Fixed the post-coercion normalization gap. Stringified `expand_turn:"0"` and `context_lines:"0"` are now coerced, normalized before final validation/dispatch, and recorded once in repair telemetry. The production job/artifact regression, focused retained-read tests, `go test ./agent -count=1`, and `make lint-golangci` pass.
- Preserved `output_match:""` as omission. Issue #827 explicitly defines the empty value as a materialized neutral default and requires it to behave exactly like omission for job and artifact refs. Restoring empty-RE2 search semantics would violate that acceptance criterion and the issue's absent/null/empty/default matrix; non-empty expressions remain search mode.

obra:
RoboRev follow-up at `a190323eb650f234e0b26fb8151ee4f304a207c1`:

- Fixed `context_lines: null` normalization independently of non-empty `output_match`, with production job/artifact regressions.
- Fixed `output_match: null` normalization as omission, including telemetry assertions.
- Removed the dead validation wrappers and moved their restriction/diagnostic coverage to the shared production collector path.
- Kept `output_match: ""` as omission intentionally: issue #827 explicitly requires empty string to be semantically neutral and equivalent to omission for job/artifact refs. Preserving it as an empty RE2 search would violate that accepted contract.

Verification on this head: focused retained-read tests, internal tool definitions, repair tests, Google provider tests, serial `go test ./agent -count=1` (337.494s), and `make lint-golangci` all pass. Independent scoped re-review and fresh GitHub CI are in progress.

PR #839, opened 2026-09-02
fix(agent): harden task-list schema and progress

### LINKED ISSUE #830: Make task_list schemas typo-resistant and report cancellation consistently
Priority: P1

## Problem

`task_list` has two distinct UX defects.

First, its prose calls `description` a “brief description.” In session `034HvTCI5LrwbM2ZZpBMqN`, the model twice emitted a `brief` key and omitted required `description`. Nested add/update schemas do not set `additionalProperties:false`, so unknown fields can pass schema validation and may be ignored when another valid mutation is present.

Second, cancellation accounting is inconsistent. `agent/task.Summarize` and `Progress` count only `done`, while total includes cancelled tasks. The task list can therefore report `All tasks complete. Progress: 7/8 tasks complete.` The TUI uses different accounting and treats cancelled tasks as done.

Relevant code:
- `agent/internal/tool/definitions.go`: `DefTaskList`
- `agent/session_tools_task.go`: `decodeTaskArgs`, `formatTaskList`, `taskListAllDone`
- `agent/task/task_store.go`: `Summarize`, `Progress`, `UpdateWithSnapshot`
- `cmd/evener-tui/hub_status.go`: `taskSummary`
- `agent/session_tools_task_test.go`
- `agent/task/task_store_test.go`

### Make field names explicit and typo-resistant

- Change schema prose to say “the field named `description`” rather than “brief description.”
- Set `additionalProperties:false` on nested add and update objects.
- When an unknown field causes a required field to be absent, return a targeted diagnostic naming both facts and showing the valid shape.

Do **not** add a `brief` compatibility alias. Invalid and conflicting mutations should fail visibly rather than introduce a second accepted field name.

### Report task outcomes consistently

Create one shared summary contract for the agent and TUI. Report distinct counts, for example:

`6 done, 1 cancelled, 0 remaining (7 total)`

Use “No actionable tasks remain” when all tasks are terminal but not all are done. Reserve “All tasks complete” for `done == total` if that phrase remains.

## Restrictions to retain

Keep:
- atomic mutation batches;
- required update IDs;
- valid status values;
- the single-in-progress invariant;
- dependency and cycle checks;
- unknown-ID and no-op update rejection;
- rejection of retired `action`, `tasks`, and `updates` shapes.

## Acceptance criteria

- The schema names the exact `description` key and rejects unknown nested keys.
- Every nested `brief` key is rejected atomically, is never copied to `description`, and produces a diagnostic that names both invalid `brief` and required `description`.
- Agent and TUI render the same done/cancelled/remaining/total accounting.
- Terminal lists with cancellations never produce contradictory “all complete” fractions.
- Existing mutation atomicity and graph invariants remain unchanged.

## Tests

Add schema and handler tests for unknown nested fields, missing `description`, malformed updates, and mixed valid/unknown mutations. Add shared-summary tests for empty, open, in-progress, done, cancelled, and mixed task sets. Add a direct TUI regression test; the current focused TUI test command matches no `taskSummary` test.

obra:
## Summary
- close nested task-list add/update schemas and reject `brief` without aliasing it
- return targeted add-shape diagnostics and preserve atomic rejection for malformed batches
- share done/cancelled/remaining/total outcome text between agent task responses and the TUI

## Restrictions retained
- Atomic batches, required update IDs, valid statuses, single in-progress, dependency/cycle checks, unknown-ID/no-op rejection, and retired shapes remain unchanged.
- `brief` remains invalid in every nested add/update entry and is never copied to `description`.

## Tests
- `go test ./agent/task -count=1`
- `go test ./agent/internal/tool -run 'TestDefTaskList' -count=1`
- `go test ./agent -run 'Test(TaskTool_|TaskStore_Progress)' -count=1`
- `go test ./cmd/evener-tui -run 'TestTaskSummaryUsesSharedOutcomeCounts' -count=1`

Closes #830

### DIFF
--- a/agent/cov_intg_hooks_test.go
+++ b/agent/cov_intg_hooks_test.go
@@ -112,6 +112,43 @@ func TestIntg_ExecTool_PreToolUseRewritesInput(t *testing.T) {
 	collect()
 }
 
+func TestIntg_ExecTool_PreToolUseRewriteRevalidatesTaskList(t *testing.T) {
+	t.Parallel()
+	sess, _ := intg_hookSession(t, `{
+		"hooks": {
+			"PreToolUse": [
+				{"matcher": "task_list", "hooks": [{"type": "command", "command": "echo '{\"hookSpecificOutput\":{\"hookEventName\":\"PreToolUse\",\"permissionDecision\":\"allow\",\"updatedInput\":{\"add\":[{\"type\":\"implement\",\"brief\":\"rewritten brief\",\"prompt\":\"do it\"}]}}}'"}]}
+			]
+		}
+	}`)
+	defer sess.Close()
+
+	// The original call is valid. The hook replaces its add entry with the
+	// known-bad brief shape, so the post-hook dispatch must run task_list's
+	// targeted PreValidate rather than only generic JSON-schema validation.
+	res := sess.execTool(context.Background(), llm.ToolCallData{
+		ID:        "hook-rewritten-task",
+		Name:      "task_list",
+		Arguments: json.RawMessage(`{"add":[{"type":"implement","description":"original","prompt":"do it"}]}`),
+	}, "")
+	if !res.IsError {
+		t.Fatalf("hook-rewritten invalid task_list = %+v, want targeted prevalidation error", res)
+	}
+	want := `add entry 0 has invalid field "brief"; use the required field named "description" instead`
+	if !strings.Contains(res.Output, want) {
+		t.Fatalf("hook-rewritten task_list diagnostic = %q, want %q", res.Output, want)
+	}
+	if strings.Contains(res.Output, "tool args schema validation failed") {
+		t.Fatalf("hook-rewritten task_list should not fall through to generic schema validation: %q", res.Output)
+	}
+	sess.mu.Lock()
+	executed := sess.taskToolEverUsed
+	sess.mu.Unlock()
+	if executed {
+		t.Fatal("hook-rewritten invalid task_list reached the task handler")
+	}
+}
+
 func TestIntg_ExecTool_PreToolUseInvalidUpdatedInputErrors(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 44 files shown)


obra:
Review follow-up at `03a315b1682e3eb93cec81edaa30c45c7785f57c`:

- Propagated cancelled/remaining counts through AppWire snapshots, updates, projections, runtime patches, generated TypeScript, and protocol docs while keeping the new fields optional for old-daemon/legacy payload compatibility.
- Invoked task-specific prevalidation in the provider-driven session preparation path without double-applying it during registry execution.
- Updated the task-card parser to support both historical and richer outcome footers.
- Fixed the initial generated-type compatibility failure by preserving total/done-only legacy shapes and current fully-cancelled semantics.

Independent review found no new Critical/Important issues. Parent verification passed `make lint && make vet && make test` at this exact head.

obra:
@roborev-primeradiant please re-review current head `4fc5ca5a263f3e5e228c3b6e3121016e3bc61f0f`.

The Medium mixed add+update atomicity finding is addressed with store-level `ApplyBatch`: it validates update references against pre-add state, stages add/update under one lock, writes one durable snapshot, and commits in-memory state/`nextID` only after save succeeds. The production regression proves two failed retries leave memory and persisted reload unchanged, emit no task event, and preserve the original next ID; it fails on the prior head and passes here.

The Low duplicate-validation suggestion is intentionally no-change: `PreValidate` and `decodeTaskArgs` both call the same `validateTaskListItemFields` implementation, so no allowlists or diagnostics are duplicated; the handler-side invocation protects direct/internal callers that bypass preparation.

Independent review passes with no Critical/Important findings. Focused regression, `go test ./agent/task -count=1`, full serial agent tests, golangci-lint, vet, gofmt/diff, parent `make test-web`, and all 13 GitHub Actions checks pass. The RoboRev status has remained pending since 03:04Z, so this mention requests a fresh current-head synthesis.

PR #843, opened 2026-09-02
feat(agent): delegate_wait collects a batch of delegate reports in one round

obra:
## What

`delegate_wait(targets?, max_wait_ms)`: block up to a clamped budget until the named delegates finish (default: every running delegate the caller owns) and return each one's report in the shape `delegate_send` returns after an inline wait. Delegates still running at the deadline are reported as running and stay owed their notification; unknown or idle targets are refused in place. Gated with `delegate`, so a leaf never sees it.

## Why

`delegate` returns immediately, and the only way to wait for a running delegate was to end the turn and be woken by its notification, one delegate at a time. A coordinator that fans out therefore serializes its waits. On Terminal-Bench the arms that delegated most died on agent timeouts (14% of trials vs 3–4% elsewhere) with roots still waiting on delegates when killed.

## How

`RegisterInlineWaiterForRunning` attaches an inline waiter to a delegate's current running generation (one per generation), so the generation's terminal packet resolves the tool call through the existing delivery claim and is committed as a delivery via `queueDelegateDeliveryCommit`, never repeated as a notification. Timing out withdraws the waiter, so the eventual delivery goes the normal notification route. `RunningDelegateIDs` lists the caller's running delegates for the default target set.

## Tests

- controller: waiter on a running generation resolves with the packet and a commit; refuses idle delegates and a second waiter; timeout withdraws the waiter so the later delivery carries no waiter
- schema: `delegate_wait` parameters
- handler: a live child completes and is returned as completed with its report and one queued delivery commit; unknown target refused in place; argument validation
- pins updated: deadline audit, controller call-site inventory, root-only tool list, provider tool lists

One prompt sentence in `delegation.md` tells the coordinator to use it after spawning a batch.


obra:
Two follow-up commits from three-trial Terminal-Bench probes of the tool in use:

- 992aa5a5e: budgets clamped to [1s, 30min] instead of the shell-job inline window (60s). Delegates ran 3–8 minutes, so every wait in the first probe timed out and the coordinator re-waited until it lost the trial.
- 6d2f481b4: `max_wait_ms` optional with a ten-minute default; description says this is a wait, not a poll. In the second probe the coordinator used one-second budgets as polls three times, and its one real wait expired a minute before a six-minute delegate finished. That probe still passed 3/3 with fewer root turns than the baseline.

Resolution through a registered waiter is covered in-process by `TestDelegateWaitTool_ReturnsCompletedChildReport` (the child finishes during the wait and comes back completed with one queued delivery commit).

obra:
Review round addressed in bb59f4e46.

- **Bound before acknowledging (high):** `stableDelegateWaitTool` now renders the reply first via `marshalDelegateWaitResult`, which degrades progressively (full reports → outputs truncated to an equal share of the budget with `truncated: true` → ids/status/action only) and errors only when the core does not fit. Delivery commits are queued only after the reply fits; on failure the resolved commits are released with `Complete(false)` so the reports stay owed as notifications. Tests: `TestDelegateWaitResult_BoundsProgressively`, `TestDelegateWaitTool_NoCommitWhenResultCannotBeBounded`.
- **Prompt gating (medium):** `delegate_wait` added to `delegationPromptToolNames`. Test: `TestDelegateWait_IsGatedWithTheDelegationPrompt`.
- **Integer budget (low):** `max_wait_ms` is parsed by `integerArg`: finite, integral, in int32 range; 1.5, "1000" and 1e300 are `invalid_request`. Test: `TestDelegateWaitTool_BudgetMustBeAnInteger`.
- **transcript_ref (low):** `RegisterInlineWaiterForRunning` returns the descriptor's transcript ref; completed and timed-out results carry it. Asserted in `TestDelegateWaitTool_ReturnsCompletedChildReport`.
- **Doc comments (nit):** `DefDelegateSend` comment restored above its definition; `DefDelegateWait` has its own. Definition test checks `Description == ""`.

`go test ./agent/... ./cmd/evener/ ./internal/...` and `golangci-lint` clean.

obra:
Closing unmerged. Jesse's call: waiting is an antipattern. A parent that has fanned out should end its turn and be woken once per delegate completion; when reports only make sense together, the prompt should direct the parent to an intermediate subagent that fans out and reports once. A blocking barrier tool is the wrong shape for that, even though the 3-trial probe scored well with it. The event-driven design is being worked out separately.

Issue #831, opened 2026-09-02
Heal default communicate envelopes and clean JSON-object strings before validation


obra:
Priority: P1

## Problem

The default `communicate` contract and its repair path disagree at several boundaries:

- the description says visible text may live in top-level `message` or `output.message`, but the schema requires top-level `message`;
- the handler can normalize an absent output node, but prevalidation rejects a wholly absent/null `output` before the handler runs;
- repair fills missing nested fields only when `output` is already an object;
- providers sometimes double-encode an object as a JSON string.

In session `034HvTCI5LrwbM2ZZpBMqN`, `communicate.output` was a string containing JSON rather than an object, and the call failed before the same content was retried successfully.

Relevant code:
- `agent/internal/tool/definitions.go`: `DefCommunicateNamed`
- `agent/session_tool_repair.go`: `prepareToolCall`
- `agent/session_tools_communicate.go`: `communicateEnvelopeFor`, `fillCommunicateEnvelope`, `normalizeNodeOutput`
- `agent/session_tool_round.go`: `handleNoToolCalls`, `decideNoToolCalls`
- `agent/session_communicate_issue627_test.go`
- `agent/session_communicate_test.go`

## Proposed healing

For the exact default `communicate` result contract:

1. Synthesize a missing/null `output` object and fill `message`, `data`, and `artifacts` defaults.
2. If top-level `message` is absent and `output.message` is a string, copy it to the top level before validation.
3. Support bounded, schema-directed promotion when `output` is a string:
   - require valid UTF-8 within the normal argument-size limit and a named maximum JSON nesting depth;
   - decode exactly one JSON value with no trailing non-whitespace content;
   - require the decoded top level to be an object;
   - run the existing default-envelope repair;
   - accept only if the repaired object validates fully against the expected schema.
4. Record each repair in telemetry with the path and repair kind, not the full potentially sensitive value.

If parsing fails, produces an array/scalar, contains trailing content, or fails schema validation, retain the hard error and say that a JSON-looking string must be passed as an object.

## Scope decision

Start with `communicate.output`, where the schema always expects an object and a string has no valid meaning. Gate default-contract repair by equality with the canonical default output schema, not merely by finding the same property and required-key names. A later change may generalize this to any object-typed tool field, but only after measuring repair telemetry and proving that schema-directed promotion cannot change a field with legitimate string semantics.

[...]

PR #848, opened 2026-09-03
fix(tool): break repeated semantic failures

### LINKED ISSUE #829: Detect repeated tool failures by semantic call signature
Priority: P0

## Problem

The repeated-tool breaker fingerprints calls too literally. Nonsemantic changes—especially free-text `intent` and neutral/default fields—let a model repeat the same invalid operation after the same error.

Session `034HvTCI5LrwbM2ZZpBMqN` made 20 `read_transcript` calls against one job ref; all failed with the same ref-mode error. The exact-call breaker detected runs of identical arguments and blocked one execution, but small changes to `intent`, search text, or default-valued fields reset the effective loop. The canonical error-loop audit rated the session medium severity; its longest byte-identical failed run was only three calls even though the semantic loop lasted 20.

## Proposed behavior

Add a semantic failure fingerprint for retry control. It should include:
- tool name;
- normalized operation/mode;
- semantic target identity;
- meaningful arguments after neutral/default normalization;
- normalized error code/class and failing boundary.

It should exclude:
- free-text `intent` and descriptions;
- argument order;
- null/empty/default fields that normalization treats as omitted;
- presentation-only error text.

After repeated equivalent failures, return a breaker error that:
1. states which semantic call is looping;
2. lists the prior attempts;
3. names the unchanged invalid boundary;
4. gives a materially different valid example or directs the model to another tool.

A meaningful change to target, mode, or relevant arguments creates a distinct fingerprint with a fresh attempt; it must not erase the bounded history of the old fingerprint. A success clears the failure run for that same semantic fingerprint.

## False-positive controls

Do not collapse calls merely because they use the same tool. Different refs, regexes, offsets, watch IDs, task IDs, or operations must remain distinct when they can change the outcome. Keep the existing exact-call guard as a fast path; add semantic detection above it with bounded history and explicit telemetry.

## Acceptance criteria

- Changing only `intent` cannot evade the repeated-failure breaker.
- Adding/removing neutral generated defaults cannot evade it.
- A meaningful argument correction gets a fresh attempt.
- Alternating two failing semantic fingerprints cannot erase either run.
- A success clears only the matching failure run.
- Breaker output identifies the normalized failing boundary and a valid alternate action.
- Telemetry records exact and semantic signatures without logging secrets or unbounded argument bodies.
- The existing human-approved `WithBreakerBypass` path remains available and auditable.

## Tests

- Replay the 20-call `read_transcript` pattern from session `034HvTCI5LrwbM2ZZpBMqN`; the semantic breaker should stop it after the configured threshold.
- Vary only `intent` and neutral fields and verify equivalence.
- Vary a meaningful field and verify the call executes.
- Cover alternating/interleaved fingerprints, per-fingerprint success reset, bounded history, sensitive-value redaction, and `WithBreakerBypass`.

obra:
Closes #829

## Summary
- Add a bounded semantic failed-call ledger alongside the existing exact-argument fast path.
- Fingerprint normalized tool/mode/target arguments plus stable error class; omit intent, descriptions, secrets, and unbounded bodies while preserving selectors and IDs.
- Emit bounded exact/semantic/bypass telemetry on tool-call completion and provide safe semantic-loop remediation.

## Security and privacy
- Telemetry contains only bounded hashes; semantic canonicalization redacts sensitive fields and omits unbounded bodies.
- Parked semantic responses name a normalized boundary and hashed signature, not raw arguments or retained failure text.

## Tests
- Added semantic equivalence, meaningful correction, interleaving, success-reset, bounded/redaction, bypass, and long-regex cases.
- Replayed the 20-call read_transcript neutral-default/intents pattern from #829.
- Ran focused tool/agent tests, serial full agent tests, vet, lint-golangci, lint-gofmt, and diff checks.


obra:
Addressed the current `b92414b` RoboRev review in final candidate `f3331ccb9`.

- **Per-class semantic history — no code change.** Issue #829 defines the semantic fingerprint to include the normalized error class and boundary, and its acceptance criteria require that “alternating two failing semantic fingerprints cannot erase either run.” Resetting A when B appears would violate that requirement. The current bounded ledger therefore retains A and B independently; `A, B, A, A` parks A on the fourth call by design.
- **Absent-name removal lifetime — fixed in `87250c992`.** `Remove` and `Unregister` now rotate the persistent lifetime on every explicit reset, even when the name is already absent, before clearing both ledgers under the registry write lock. The new deterministic regression captures an absent-name prevalidation snapshot, performs each reset, and proves that stale finalizations publish no breaker signatures and cannot repopulate or park the fresh absent lifetime. The test failed for both paths before the production change and passes afterward.

Verification on exact merged head `f3331ccb9`:
- focused lifetime/prevalidation tests: 20 repetitions; race: 3 repetitions
- full `agent/internal/tool` tests and race suite
- `ROOT_FULL=1 WEB=0 make test`
- `make test-race RACE_SCOPE=root`
- `go vet ./...`, `make lint-gofmt`, and `make lint-golangci`
- `make test-web` and all five `make test-web-browser` guards
- independent exact-head review: PASS, 0 Critical / 0 Important / 0 Minor

The branch also contains current `main` `92b1b716f`; the merge had no overlap with the two fix files. Fresh CI and RoboRev are now running on `f3331ccb9`.

obra:
Addressed both `f3331cc` RoboRev findings in `e9ef9ffa7`:

- `FailureClasser` now takes precedence over rendered-output boundary detection, so typed failures retain their stable class even when their presentation begins with `invalid_request:`. The new A/B/A regression proves all three distinct typed attempts execute; existing untyped compatibility remains covered.
- Exact and semantic breaker park messages now replace unreadable/invalid provider-supplied tool names with a fixed bounded display and omit unknown-tool snippets that repeat the raw name. Internal private/session-keyed identity is unchanged, and valid readable-name messages retain their prior detail. Regressions assert both `Output` and `FullOutput` contain no synthetic private fragment and stay within 512 runes.

TDD evidence: each new regression failed against the pre-fix behavior and passed after the production change.

Current-head verification:
- breaker/prevalidation suite repeated 20x: PASS
- focused race suite repeated 3x: PASS
- full `agent/internal/tool`, normal and race: PASS
- full `go test ./agent -count=1`: PASS (230.769s)
- `make lint && make vet`: PASS
- independent exact-diff review: PASS, zero findings

obra:
Addressed the current `e9ef9ffa` RoboRev findings in `8f9b66fa86db37a05cc28f9993192ad63800c950`.

Accepted and fixed:

- Unreadable unknown provider tool names (invalid grammar, surrounding whitespace, or over 64 bytes) now use the fixed model-visible identity `invalid tool name` in both direct Registry and Session pre-dispatch errors. The raw name remains absent from model-facing, full, recoverable, and `TOOL_CALL_END.Error` text; valid readable unknown names retain their useful detail.
- Direct registered calls with invalid UTF-8 arguments now finalize against the explicit `arguments_encoding` boundary rather than deriving `arguments_json` from rendered error text.
- `ExecutePreparedCall` now receives the opaque `PrevalidationSnapshot` captured during Session preparation. It skips `PreValidate` only while registry, name, lifetime, and selected registration generation still match under one registry read lock. If registration was replaced, the successor runs its own normalization and `PreValidate`; a rejecting successor is not executed.

No change for the suggested reset of interleaved semantic fingerprints. Issue #829 explicitly requires per-fingerprint history: “Alternating two failing semantic fingerprints cannot erase either run.” Resetting A when B appears would reintroduce the bypass that this feature is designed to prevent. Existing and independently rerun coverage confirms A/B histories remain distinct and both park at their own threshold.

Verification on exact `8f9b66fa`:

- six new regressions: 20× PASS
- the same regressions under `-race`: 3× PASS
- full `go test ./agent/internal/tool` and full package race: PASS
- full serial `go test ./agent -count=1`: PASS (249.358s)
- `make lint && make vet`: PASS
- independent exact-diff review: PASS, 0 Critical/Important/Minor findings

obra:
Addressed all three findings in `a3e50ea6b2d5abf8d733de0570409f5caa233af2`, on top of current `main`.

- **Repaired JSON identity:** when syntax repair succeeds but a later non-`ask_user` schema check fails, the breaker now retains the repaired semantic arguments. Distinct repaired calls keep distinct semantic identities, while rejected raw arguments remain unchanged and no repair event is emitted. `TestSession_RepairedJSONSchemaFailuresKeepDistinctSemanticIdentity` covers this path.
- **Tool-name projection:** raw provider names remain private to registry lookup, breaker routing, and hook matching. Every external Session projection now uses a bounded display name, including `ExecResult.ToolName`, repair/start/delta/end and stream events, hook payloads, live history, canceled/synthetic results, and durable transcripts. Readable unknown names remain visible. `TestExecuteCallUnknownToolNameProjectionIsPrivateAndBounded`, `TestSessionUnknownToolNameExternalProjectionIsPrivateAndBounded`, `TestSessionReadableUnknownToolNameExternalProjectionIsPreserved`, and `TestSyntheticToolResultsProjectUnreadableToolName` cover these boundaries.
- **Stale semantic metadata:** successful calls, unclassifiable failures, and `clearFailures` now clear the exact entry's old semantic fingerprint and boundary. `TestExecuteCall_NormalizedSuccessClearsExactSemanticMetadata` verifies that a later exact park cannot reuse stale semantic guidance.

Verification:

- focused regressions, repeated and under `-race`: PASS
- `make lint`: PASS
- `make vet`: PASS
- `make test`: PASS
- independent review: PASS, 0 Critical/Important/Minor findings

@roborev-primeradiant please review the updated head.

obra:
Follow-up `ba646f86a9330a9dfdcd6f1a81d4819d0435428f` fixes the current-head fuzz failure.

`FuzzStoolDispatch` still expected raw `call.Name` in `ExecResult.ToolName`, but this change intentionally projects unreadable provider names to the bounded external label. Four committed seeds failed at `a3e50ea6`; the oracle now expects `tool.DisplayToolName(call.Name)` while retaining its positional `CallID`, deterministic-dispatch, and structured unknown-tool assertions.

Verification on `ba646f86a`:

- exact four failing seeds: RED on `a3e50ea6`, PASS after the fix
- full `FuzzStoolDispatch` saved corpus (200 files): PASS
- `make fuzz`: PASS
- `make lint`: PASS
- `make vet`: PASS
- `make test`: PASS
- independent no-weakening review: PASS, 0 Critical/Important/Minor findings

@roborev-primeradiant please review the updated head.

obra:
Addressed the wire-name finding in `efb5c22e379064f3c81017be6e0b568dedf9dd9e`.

- Kept `invalid tool name` for human-facing events, hook inputs, and diagnostics.
- Added the provider-valid `invalid_tool_name` identity for stored assistant calls, stored tool results, synthetic recovery results, and provider replay.
- Projected legacy transcript names before both context-management/note-elicitation requests and normal history expansion, without mutating source messages.
- Applied the same bounded display projection to failed-stream salvage text.
- Covered ordinary, durability-first, and delegate-delivery append paths; readable names and call/result IDs remain unchanged.

TDD evidence:
- The legacy pre-context test failed with the raw secret before the snapshot fix, then passed.
- The salvage regression failed with a 7,853-byte raw name before projection, then passed.
- Focused projection/durability/salvage tests pass repeatedly and under `-race`.

Verification on the exact commit:
- `make fuzz` — pass
- `make lint` — pass
- `make vet` — pass
- `make test` — pass, including root, agent, auxiliary modules, and web
- Independent high-rigor re-review — PASS, zero findings

obra:
Implemented the valid current-head provider-recovery finding in `2b3c582c0d0973435be15659199c07dfd340d81a`.

Root cause: `providerHistoryMessage` projected an intentionally empty `ToolResult.Name` to `invalid_tool_name`. For Chat Completions rows with `ToolResultName`, the adapter deliberately uses an empty result name as a sentinel and recovers the actual name from the correlated assistant tool call ID (`llm/providers/chatcompletions/messages.go:178-184`). The projection now leaves only empty result names untouched; every nonempty unreadable name is still projected to the provider-valid placeholder, and tool-call names remain fully projected.

TDD evidence:
- RED: `TestProviderHistoryMessagePreservesEmptyToolResultNameForAdapterRecovery` failed with `projected empty tool-result name = "invalid_tool_name"`.
- The downstream control `TestToChatMessagesRecoversEmptyToolResultNameFromCallID` passed independently, proving the adapter recovers `read_file` by call ID.
- GREEN: the agent regression passed 20x and under `-race` 3x; the adapter regression passed 20x.
- All provider-name projection regressions passed 20x and under `-race` 3x.
- Full `go test ./agent -count=1` and `go test ./llm/providers/chatcompletions -count=1` passed.
- Full `make fuzz && make lint && make vet && make test` passed.
- Independent exact-diff review: PASS, zero findings.

The per-fingerprint reset recommendation remains intentionally **no change**. Issue #829 explicitly requires: “Alternating two failing semantic fingerprints cannot erase either run.” Resetting A when B appears would violate that acceptance criterion and re-enable the interleaved-loop evasion the issue was created to stop. The checked-in `TestSemanticFailureBreaker_MeaningfulArgumentsAndInterleavedRunsRemainDistinct` verifies both A and B preserve independent histories and park at their own third equivalent failure; it passed 20x and under `-race` 3x. A meaningful correction still receives a fresh attempt, and success clears only the matching fingerprint.

obra:
Resolved the remaining `invalid_tool_name` projection-collision finding at exact head `bba0b8bd65fdb87160ce303f98242f4b9100c0e8`.

- `invalid_tool_name` is now a single exported wire-placeholder constant and is reserved at `Registry.Register`, including whitespace-padded aliases.
- `Session.RegisterTool` now returns registration errors and rebuilds caches only after success, so runtime configuration failures are actionable rather than silent.
- New and restored sessions reject a persisted `ResultToolName` that aliases the reserved placeholder.
- Provider profile mappings are asserted never to expose a live definition under the placeholder.
- Registry, runtime Session, new-session, restore-session, and provider-mapping regressions cover the collision paths.

Verification on this exact source tree:
- focused tests repeated 20x: PASS
- focused race tests repeated 3x: PASS
- `make lint-golangci`: PASS
- `make fuzz && make lint && make vet && make test`: PASS (exit 0)
- independent exact-diff re-review: PASS, 0 Critical / 0 Important / 0 Minor; ordinary runtime registration behavior preserved

obra:
Addressed RoboRev's two findings at `4d2e959394f65741528348f37e402cdfaffd6d99`.

- Prepared Session calls now run the captured registration's `NormalizeArgs` before `PreValidate`, freeze those normalized arguments for same-lifetime execution, and invoke both hooks exactly once. If the registration changes before dispatch, the successor receives the original Session-repaired call and runs its own normalization and prevalidation. The core retained-read path preserves its existing repair telemetry.
- Registry registration and configured result-tool validation now reject ordinary surrounding whitespace rather than accepting a name that later projects as invalid. Reserved `invalid_tool_name` behavior remains unchanged.

Regression coverage includes normalized-input rejection before dispatch, exact-once hooks, old-normalized-argument isolation across replacement, successor fallback, retained-read telemetry, ordinary padded names, and new/restore result-tool initialization.

Verification:
- focused tests: 20 repetitions
- focused race tests: 3 repetitions
- full `agent/internal/tool` and `agent` suites
- independent review: PASS, zero findings
- `make fuzz && make lint && make vet && make test`: PASS

obra:
Addressed the refreshed RoboRev provider-alias finding in `96c552ac3073b828a5bb4652ce64fe302043dbd5`.

The Session response boundary now reverse-maps provider aliases only when the raw name is strictly readable. Malformed names such as `" exec_command "` remain unchanged for registry lookup, so they cannot become executable `shell` calls after assistant history has projected them as `invalid_tool_name`.

Added a scripted-provider lifecycle regression that verifies the malformed alias does not invoke a replacement `shell` executor and that both assistant-call and tool-result history use the provider-safe invalid-name identity. Existing valid OpenAI alias dispatch and prepared/repaired-call lifetime behavior remain covered.

Local verification on the exact commit ancestry:
- focused malformed/valid alias and name-projection tests, repeated 20x: PASS
- focused alias/name/prepared-call tests under `-race`, repeated 3x: PASS
- `go test ./agent -count=1`: PASS (216.835s)
- `make fuzz && make lint && make vet && make test`: PASS
- independent review: PASS, 0 Critical / 0 Important / 0 Minor findings

Fresh current-head CI and RoboRev are still required.

obra:
Addressed the current RoboRev findings in `0bad6498959917f53b50607bde3d48bc176a7dde`.

1. **Padded alias / preview identity** — the reported dispatch-bypass mechanism was not accurate: `toolcall.Accumulator` receives the raw name first and preserves the first non-empty name, and dispatch already rejects unreadable padded aliases. I did reproduce the adjacent defect: the stream preview side map trimmed/canonicalized the name independently, so ` communicate ` could open a communicate preview, and a later valid delta could overwrite that malformed start identity. The fix uses one exact-readability projection for lifecycle and stream handling and preserves first-nonempty identity across start/delta/end. Direct regressions cover the padded name and malformed-start/valid-delta sequence; valid communicate previews remain unchanged.

2. **Stale prepared arguments after registration replacement** — confirmed. A generic schema repair could remain in the prepared call after the tool registration changed, so the successor's `NormalizeArgs` saw the predecessor's repaired value. Session preparation now retains the original provider arguments, and stale-generation execution restores those originals before the successor's normalize/prevalidate/schema pipeline. Current-generation prepared calls still reuse their validated prepared arguments. The regression uses the same schema on both registrations and proves the successor receives provider string `"7"` before doing its own normalization.

Verification on the committed tree:
- focused regressions: 20 repetitions, pass
- focused race tests: 3 repetitions, pass
- `go test ./agent/internal/tool -count=1`: pass
- `make fuzz && make lint && make vet && make test`: pass
- independent exact-diff review: PASS, 0 Critical / Important / Minor findings; both regressions independently fail on the parent and pass on this head

obra:
Addressed the current `0bad649` RoboRev finding in `968f0d5a4973ba7548329c9b30184061e6657177`.

A successfully merged `PreToolUse` `UpdatedInput` now invalidates the stale preparation error as well as the prepared fast path. The corrected call then runs through the registry's normal current-registration normalize/prevalidate/schema/dispatch pipeline. Raw-rejected calls still cannot be replaced, malformed source JSON still fails the update merge, hook denial is unchanged, and calls with no update retain their original preparation failure.

TDD evidence:
- new custom-tool Session regression failed on the parent with zero executions and the stale schema error, then passed after the fix
- the existing Issue #831 communicate hook test exposed a stale assertion that expected this bug; its update case now requires successful execution, while deny/raw-invalid/no-update controls remain unchanged
- focused regressions: 20 repetitions, pass
- focused race tests: 3 repetitions, pass
- tagged hook/core seed: pass
- fresh `make fuzz && make lint && make vet && make test`: pass
- independent final three-file review: PASS, 0 Critical / Important / Minor findings

obra:
Implemented the current `968f0d5` review findings in `0864a3027`:

- Registration replacement now reapplies a hook update against both the prepared candidate and the provider-original arguments, then lets `ExecutePreparedCall` distinguish the current registration snapshot from a stale one. A stale snapshot restores provider-original + hook update before successor normalization/prevalidation/schema validation. The real Session regression was red with predecessor-repaired numeric data reaching the successor and is green with the original string preserved.
- Retained-read second-pass breaker identity now serializes `finalErrorArgs`, matching the final validation failure. The expanded Session test covers string zero, numeric zero, omitted defaults, and a remaining invalid offset; its signature assertion was red before the fix and green after it.
- Exact parking includes semantic wording/signatures only when that exact semantic fingerprint has itself reached the threshold. The typed-class counterexample was red before the count gate and is green now.
- The unrelated root CI timeout was a test-seam race: the injected `wait()` read a PID file before the fake child created it. The harness now captures the actual PID passed to `wait` and replays that PID from stale `jobs()` output; production `scripts/web/test-web.sh` is unchanged. The ambient-fixture guard was red before filtering the new seam variable and green after it.

Independent review of the complete eight-file delta: PASS, zero Critical/Important/Minor findings. Focused repeated/race tests passed. Fresh local canonical gate `make fuzz && make lint && make vet && make test` exited 0, including root and web tests.

obra:
Addressed the current review on `0864a3027` in `d47470438`.

- **Prepared-call semantic identity — no code change (finding rejected).** `ExecutePreparedCall` does not finalize the transient raw signature: `registry.go` unconditionally recomputes `semanticSignature` after selecting either the current prepared arguments or successor-normalized arguments. Direct Registry and real Session probes covered current and stale registration snapshots, distinct exact/raw identities, and shared semantic identity; the third equivalent runtime call parks after two executions. The final independent reviewer reproduced this and agreed the finding is invalid.
- **Unsafe tool arguments in assistant/replay history — fixed.** History now accepts arguments only when they satisfy both `tool.ValidateRawArguments` (UTF-8 and 2 MiB bound) and `json.Valid`. Assistant persistence and legacy provider replay both substitute `{}` copy-on-write; provider responses and stored legacy history remain immutable. The regression covers actual invalid UTF-8 and valid JSON larger than 2 MiB through real Session append/request assembly.

Verification on the exact two-file diff:
- focused regression: 20 repetitions, pass
- focused race suite: 3 repetitions, pass
- independent final review: 0 Critical / Important / Minor findings
- `make fuzz && make lint && make vet && make test`: exit 0

Commit: `d474704386788f464c450a0120d5f54948d3f53a`

obra:
Resolved the current `d474704` review at successor `d04e55186b757cab1cfa4e24f17fc730aa18f10a`.

- Trace suffix finding: no code change. The marker check already includes the trailing space (`trimmed[start:open+len(marker)-1]` spans `" [trace "`); direct overlay coverage confirmed valid suffixes strip and group. This finding is invalid.
- Registered hook finding: fixed. Both direct Registry and steady-state Session paths now preserve `NormalizeArgs`/`PreValidate` errors and derive breaker class/boundary from the original error rather than labeling them as raw schema failures. A mixed direct/real-Session regression covers typed `errors.Is`, raw-distinct exact identities, shared semantic identity, `invalid_request` normalization, generic PreValidate, and third-attempt parking.
- Independent review caught one interaction before publication: successful `PreToolUse.updatedInput` retained the new hook-error state. The existing hook-correction test was strengthened RED/GREEN; successful updates now clear all stale preparation-failure identity and re-run current validation before dispatch.

Verification on the exact six-file diff:
- focused regression: 20× PASS
- focused race: 3× PASS
- `make fuzz && make lint && make vet && make test`: PASS
- independent final review: 0 Critical / Important / Minor findings

obra:
Follow-up at `5538b3a`: I rewrote `stripPresentationTraceSuffix` to remove the ambiguous offset arithmetic called out by RoboRev. The parser now locates `[trace ` directly, verifies the preceding byte is the required ASCII space, and derives the token boundary from an explicit `markerEnd`. This preserves the existing behavior while keeping all slice indices byte-safe for Unicode input.

Verification on this exact one-file diff:
- focused trace/breaker regressions: 20 repeated runs and 3 race runs passed
- `make fuzz && make lint && make vet && make test`: passed
- independent review: PASS, 0 findings; semantic equivalence and Unicode byte-index safety confirmed

obra:
Resolved the current RoboRev finding at `53cd0e3624f389ba0bfc972b606d0adb881d43a1`.

- Provider projection remains request-only for a true context no-op, so durable history and `historyRevision` are unchanged.
- Transformed context publishes restore only unchanged turns tied to their explicit serialized origin; each origin is consumed once, so projected collisions cannot restore the wrong raw turn. Strategy-edited/new turns stay projected and all transient origins are cleared before publication.
- Staged compaction effects still force publication, and successful publications are projected again only for the provider request.
- Regressions cover invalid UTF-8 and oversized arguments, duplicate projected values, JSON-round-tripping custom strategies, in-place strategy edits, no-op revision behavior, and unchanged history with staged compaction.

Verification on the exact merged head: focused tests repeated and under `-race`; independent code review and independent merge review both PASS with zero findings; `make fuzz && make lint && make vet && make test` exits 0. Current `main` (`1d7ae1b06`) is merged without history rewriting.

@roborev-primeradiant please re-review the current head.

obra:
Addressed the latest RoboRev history-argument validation finding in `e9bd26a97`.

The finding was valid: `json.Valid` alone accepted non-object JSON, and a syntactically valid object such as `{"value":1e10000}` can still fail decoding into `map[string]any`. For non-empty history tool arguments, `validHistoryToolArguments` now requires the existing raw guards plus a successful decode into a non-nil object map. Existing empty-argument behavior is preserved.

Production-level regressions cover empty and valid object arguments, plus `null`, array, scalar, and decoder-overflow object inputs across durable assistant history and provider replay. They also assert that provider responses and stored legacy history are not mutated.

Verification at the exact pushed head:
- focused regression tests repeated 20x: pass
- focused race tests repeated 3x: pass
- full `go test ./agent -count=1`: pass
- `make fuzz && make lint && make vet && make test`: pass
- independent exact-diff review: PASS, 0 Critical/Important/Minor findings

The prior `web-layoutguard` CI failure at the parent head was separately investigated as browser-startup infrastructure behavior; the PR's frontend tree was byte-identical to main and canonical local browser guards passed. This push triggers a fresh current-head CI run rather than weakening assertions or increasing timeouts.

@roborev-primeradiant please re-review the exact current head.

obra:
Closing rather than landing: the incident it targets was fixed before the branch was cut.

#827 ("Normalize semantically empty read_transcript options before ref-mode validation") closed on 2026-09-03, and its fix commits — 667e26188d, d6b159c803, 3577d044c4 — are ancestors of this branch's own base, 1d7ae1b06a. The reported loop is what the branch was already standing on. What is left is +6,644 lines, 5,192 of them tests against 1,452 production lines, conflicting with main and touching files main has changed since.

I am not saying the idea is wrong; I am saying this branch cannot show that it is right, because the failure it demonstrates is no longer reproducible on its base, and 63 commits behind plus a conflict means nothing here is verified against current main. If a semantic-failure breaker is still wanted, the honest path is a fresh branch against today's main that re-establishes the failure first, then fixes it.

PR #849, opened 2026-09-03
fix(agent): heal default communicate envelopes

### LINKED ISSUE #831: Heal default communicate envelopes and clean JSON-object strings before validation

obra:
## Summary
- heal absent, null, incomplete, and double-encoded `communicate.output` only for the exact canonical default schema
- copy `output.message` into a missing top-level message and emit bounded repair telemetry
- reject malformed, non-object, oversized, overly deep, and custom-schema output strings without broadening custom contracts

## Validation
- `go test ./agent -run 'Issue831|Communicate(Output|.*Envelope|.*Schema|.*Filled|.*Default)' -count=1`
- `go test ./agent/internal/tool ./agent/internal/tool/repair -count=1`
- `go test -p 1 ./agent -count=1`
- `go vet ./agent ./agent/internal/tool ./agent/internal/tool/repair`
- `make lint-gofmt`
- `GOLANGCI_LINT_CACHE="$EVENER_SCRATCH_DIR/golangci-lint-cache" make lint-golangci`

Closes #831

### DIFF
--- a/agent/cov_s3_communicate_test.go
+++ b/agent/cov_s3_communicate_test.go
@@ -4,7 +4,7 @@ import (
 	"strings"
 	"testing"
 
-	"primeradiant.com/evener/llm"
+	"primeradiant.com/evener/agent/internal/tool"
 )
 
 func TestS3Cov_NormalizeNodeOutput(t *testing.T) {
@@ -138,31 +138,17 @@ func TestS3Cov_StringSetsEqual(t *testing.T) {
 func TestS3Cov_UsesDefaultCommunicateOutputEnvelope(t *testing.T) {
 	t.Parallel()
 
-	envelope := func(required any) llm.ToolDefinition {
-		return llm.ToolDefinition{Parameters: map[string]any{
-			"properties": map[string]any{
-				"output": map[string]any{
-					"properties": map[string]any{
-						"message":   map[string]any{},
-						"data":      map[string]any{},
-						"artifacts": map[string]any{},
-					},
-					"required": required,
-				},
-			},
-		}}
-	}
-
-	if !usesDefaultCommunicateOutputEnvelope(envelope([]any{"message", "data", "artifacts"})) {
+	if !usesDefaultCommunicateOutputEnvelope(tool.DefCommunicateNamed("communicate")) {
 		t.Fatal("full default envelope should be recognized")
 	}
-	// Missing a required name → not the default envelope.
-	if usesDefaultCommunicateOutputEnvelope(envelope([]any{"message"})) {
-		t.Fatal("partial required set is not the default envelope")
+	stricter := tool.DefCommunicateNamed("communicate")
+	params := tool.CloneSchemaMap(stricter.Parameters)
+	params["properties"].(map[string]any)["output"].(map[string]any)["properties"].(map[string]any)["message"] = map[string]any{
+		"type": "string", "enum": []string{"only"},
 	}
-	// No output props at all → false.
-	if usesDefaultCommunicateOutputEnvelope(llm.ToolDefinition{Parameters: map[string]any{}}) {
-		t.Fatal("missing output props should be false")
+	stricter.Parameters = params
+	if usesDefaultCommunicateOutputEnvelope(stricter) {
+		t.Fatal("same-key stricter output schema is not the default envelope")
 	}
 }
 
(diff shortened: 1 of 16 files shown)


obra:
Follow-up on the transcript projection finding: the original syntax-error wording was overbroad because `encoding/json.Unmarshal` rejects ordinary malformed JSON before populating a `map[string]any`. The core ignored-error finding is valid: syntactically valid `{"source":"api_log","overflow":1e1000}` returns an `UnmarshalTypeError` after populating `source`. Commit `12d4c9957` now derives call API-log identity only when argument unmarshal succeeds; result-derived API-log identity remains independent. Regression coverage includes overflow, invalid UTF-8, oversized arguments, valid calls, and result-derived identity.

PR #850, opened 2026-09-03
feat(job_watch): timers (after_seconds, repeat_seconds, note)

obra:
## Summary

Adds timers to `job_watch create`: `after_seconds` (one-shot) and `repeat_seconds` (recurring), both `self`-only, with a `note` delivered on every fire. Timers ride the existing periodic-watch engine, end through the existing clear teardown (`fired`), fold at the session so a busy or parked session gets one block per timer, and drop before the batch gate when their watch is gone. `progress_interval_ms` is now a concrete-job trigger only; the delivery budget counts condition fires only; integer arguments are read strictly.

Spec: `docs/superpowers/specs/2026-09-02-job-watch-timer-design.md`. Plan: `docs/superpowers/plans/2026-09-02-job-watch-timer.md`.

## Behavior changes to existing surface

- `progress_interval_ms` on `self`/`parent`/`dlg_` is refused.
- A 50th periodic tick no longer auto-clears a job progress watch.
- `progress_interval_ms:"1000"` and `every:1.5` are now `invalid_request` instead of silently ignored.
- Create text for a job progress watch reads `progress_interval_ms 300000ms`.
- A present `0` or `null` for `after_seconds`/`repeat_seconds` on create reads as absent, the same as `progress_interval_ms:0`.
- The 50-delivery budget now counts condition fires only; its exhaustion notice reads `matched N times` and no longer recommends a longer `progress_interval_ms`.

## Tests

TDD throughout; see the plan's per-task tests. Full `go test ./agent/... ./cmd/... ./internal/...`, lint, and a race pass over the timer tests are green.

## Hardened in review

Nine roborev rounds plus independent review after the plan landed. Behavior beyond the plan:

- The condition-fire budget latches at the match site, refuses further matches once latched (including the rest of a multi-match output chunk), and re-arms atomically if its teardown fails to persist.
- The send frame that crosses the budget is delivered through the terminal-flush rail instead of being dropped, so a parent receives all 50 frames.
- A fired one-shot whose durable end fails keeps its ticker and retries the end on the next tick without a second notification.
- `source` must be a string; a present non-string is `invalid_request`. A negative `after_seconds`/`repeat_seconds` reports the bounds error. Numeric watch arguments are bounded at the platform int range, not 32 bits.
- A watch id is minted only after target, event, and shape validation.
- The hub decides the notification type from the tag before parsing the body, so a note line beginning `excerpt:` stays prose.

## Parked from final review

- Timer ends enter the 64-key watch lineage ring (the cross-generation runaway fuse) like any other watch, per the spec's simplify pass. A timer's key is unique, so its entry can never be adopted, and 64 timer ends between a watch's clear and its re-arm evict the predecessor lineage the fuse would have inherited. Left as specified; the alternative is a one-line guard in `recordWatchEndedLocked` plus a spec sentence.


obra:
## Response to roborev (`c2d7679`)

**Stuck timer slot on teardown failure** — not changed. The breaker path has the identical shape: `autoClearWatchOverBudgetNotification` at `agent/job_watch.go:1751-1754` rolls back the rejecting marks and returns on a failed `appendWatchTeardownBatch` without reaching `detachWatchConfigSnapshots`, leaving the config live in `jm.watches`. `clearWatch` (1454) and `clearWatchByIDMatchingWithReason` (1535) do the same. The one-shot inherits that shape rather than introducing it. `clear` recovers the slot, and the failed teardown emits a session warning naming the watch id (`endFiredOneShot`, 3315-3320). A recovery policy shared with the breaker is a follow-up, noted in the PR body.

**Budget breaker lags condition fires for send watches** — not changed; pre-existing, not a branch regression. On `origin/main` the send branch of `onSessionEvent` (`git show origin/main:agent/job_watch.go`, lines 2228-2231) already incremented `conditionFires` without calling `recordWatchDeliveryLocked`, and the send-watch budget was already checked only at `settleWatchSendDelivered` (main line 474). The branch changed which counter the budget reads and how the crossing latches; it did not move the check point. The `>=` test plus the `budgetTripped` latch means a send watch that has passed the budget tears down on its next settled delivery rather than being missed.

**Note escaping misses `&`** — not changed; the mismatch predates the branch and the prose path reuses the existing excerpt contract. `decodeNotificationEntities` decoded `&amp;` on `origin/main` (`steeringClassify.ts:210`) and main already applied it to excerpt body text (`NotificationCard.tsx:95`), which is escaped by the same `escapeNotificationBody` (`agent/job_notify.go:294`). This branch did not touch the decoder; it routes timer prose through the identical escape/decode pair. Widening `escapeNotificationBody` would change every notification body the model reads, so it is out of scope here.

**Doc states wrong character bound** — fixed in `b1c0a67`. `docs/job-control.md` now says the summary bounds the note at the same 2,048 characters storage keeps, matching `watchConditionSummary`'s `limitWatchText(cfg.note, watchMessageMaxChars)`. No prompt or test pinned the old sentence.

Gates on `b1c0a67`: `gofmt -l agent/` clean, `golangci-lint run ./agent/... ./cmd/... ./internal/...` 0 issues, `go test ./agent/ ./agent/internal/tool/ -count=1` ok.

obra:
Both findings fixed.

**`steeringClassify.ts:323` — note line reading `excerpt:` split as job output.** Fixed in 128a3f854. The notification type is now decided from the tag attributes before the body is parsed, and a `watch` body is taken as prose in full instead of going through `splitNotificationExcerpt`. Job and delegate reports split exactly as before. New vitest case: a timer note containing an `excerpt:` line keeps that line and everything after it in `prose`, with an empty `excerpt`.

**`session_tools_jobs.go:2051` — undocumented int32 bound on `every`.** Fixed in 84a2c1546. Confirmed the regression against main: `every` was read by `shellIntArg`, which does `case float64: return int(v), true` with no bound, so any integral float64 a provider sent was accepted. `watchIntArg` now bounds at the platform `int` range and keeps rejecting NaN, Inf, non-integral, and non-numeric values. The upper bound is exclusive because `float64(math.MaxInt)` rounds up to 2^63. New table test covers a value above 2^31-1 accepted, ±1e19 rejected, and the existing strict-integer rejections unchanged. The timer fields keep their documented maximums in `normalizeWatchArgs`.

Gates: `npx vitest run src/panes/session/transcript/messages/` (386 passed), `npx tsc --noEmit`, `npx biome ci` on the touched dir, `gofmt -l agent/`, `golangci-lint run ./agent/... ./cmd/... ./internal/...` (0 issues), `go test ./agent/ ./agent/internal/tool/ -count=1` — all clean.

obra:
## Response to roborev (`84a2c15`)

**Budget latch not rolled back on persistence failure** — fixed in `1caa8362`. Confirmed: `recordWatchDeliveryLocked` sets `budgetTripped` at the crossing, and when `appendWatchTeardownBatch` fails, `autoClearWatchOverBudgetNotification` rolled back only the rejecting marks, so the watch went back into the live set with the latch still set and no later condition fire ever reported a crossing again. The failed-teardown path now also clears the latch (`rearmWatchBudgetBreaker`, under `jm.mu`), so the next condition fire re-reports the crossing and retries the auto-clear. New test `TestConditionFireBudget_FailedTeardownRearmsTheBreaker` drives a no-send output_match watch to the budget with the durable `EventWatchCleared` append failing, asserts the watch is still live, then lets the append succeed and drives one more fire, asserting the watch is gone and its history entry reads `budget_exhausted`. It fails on the parent commit at exactly that last fire. The success-path once-only semantics are unchanged and `TestConditionFireBudget_CrossingLatchesOnceAcrossASkippedBudget` stays green.

**Orphaned watch slot on teardown failure (`endFiredOneShot`)** — not changed, as answered in https://github.com/prime-radiant-inc/evener/pull/850#issuecomment-5525099753. The one-shot's rollback-without-detach shape is the same one the budget breaker and both `clear` paths already have on main, so it is inherited rather than introduced here, and `clear` recovers the slot while the failed teardown emits a session warning naming the watch id. A shared recovery policy is the follow-up noted in the PR body.

**HTML entity over-decoding in the timer card** — not changed, as answered in the same comment. `decodeNotificationEntities` already decoded `&amp;` on `origin/main` and main already applied it to excerpt body text escaped by the same `escapeNotificationBody`; this branch did not touch the decoder, it routes timer prose through the existing escape/decode pair. Widening `escapeNotificationBody` would change every notification body the model reads, which is out of scope for this PR.

Gates on `1caa8362`: `gofmt -l agent/` clean, `golangci-lint run ./agent/... ./cmd/... ./internal/...` 0 issues, `go test ./agent/ -run 'Budget|Watch|Timer' -count=1` ok, `go test ./agent/ -count=1` ok.

obra:
Round 4 (`1caa836`): both Medium findings fixed.

**Ghost one-shot on persistence failure** — `fbba22f6b` (test tripwire label follow-up in `53e893c96`)

A fired one-shot stopped its ticker whether or not the durable "fired" teardown landed, so a failed append left the config registered with nothing able to end it. `watchConfig` gains `firedPendingEnd`, set under `jm.mu` when `endFiredOneShot` reports a failure; the ticker stays armed and `decideProgressTick` routes the next tick to a retry that asks for the end and delivers nothing. An end that lands stops the ticker with the watch and still records reason "fired". `endFiredOneShot` now returns whether it persisted; the warning through `jm.emit` is unchanged.

Test: `TestOneShotTimer_FailedTeardownRetriesOnTheNextTick` drives a one-shot to fire with the `watch_cleared` append failing, asserts one notification and a still-live watch, then lets the append succeed, advances one interval, and asserts the watch ended "fired" with exactly one notification and one delivery. `FuzzWvProgressTick` gained the new snapshot field and oracles for the retry branch.

**Condition-fire budget not enforced when sends never settle** — `c86d67e72`

The latch test moved out of `recordWatchDeliveryLocked` into `tripConditionFireBudgetLocked`, and both live send match sites (`onSessionEvent`, `feedJobOutputWithProvenance`) now call it under `jm.mu` and schedule the same auto-clear teardown the no-send path uses, after the lock is released. Settlement keeps its delivery accounting; `budgetTripped` keeps the two ends to one teardown. The attach-scan match site is untouched: a freshly installed config goes 0 -> 1 fire and cannot reach the budget there.

Test: `TestSendWatchBudgetTripsWhenFramesNeverSettle` drives a caller-send watch whose frames stay pending (nothing drains them in a bare job manager) through 50 matches and asserts the watch auto-cleared, no pending frames remain, exactly one budget-cleared notification was queued, and history records `budget_exhausted`.

Gates (all exit 0): `gofmt -l agent/`; `golangci-lint run ./agent/... ./cmd/... ./internal/...` (0 issues); `go test ./agent/ -run 'Budget|Watch|Timer|Seqfuzz|seqfuzz' -count=1`; `go test -race ./agent/ -run 'Timer|Budget' -count=1`; `go test ./agent/ -count=1`.

Head: `53e893c96`.

obra:
## roborev round 5 (`ee96094`)

**Medium — attach-scan send path incremented `conditionFires` without tripping the breaker** — fixed in `11b76c882`.

Resolved structurally rather than with a site-specific latch. Counting a condition fire and consulting the breaker are now one operation, `noteConditionFireLocked(cfg) (crossedBudget bool)`, wrapping the existing `tripConditionFireBudgetLocked`. All four sites that used to do `cfg.conditionFires++` go through it: `onSessionEvent`, `feedJobOutputWithProvenance`, and both attach-scan rails in `fireAttachScan`. `grep 'conditionFires++'` now has exactly one hit, inside the helper. Where the attach scan newly learns it crossed, it schedules the same `autoClearWatchOverBudget(cfg)` the live sites schedule, after `jm.mu` is released and outside `pendingJobNotifsMu`.

Delivery counting split out as `countWatchDeliveryLocked` so the observation rails count their delivery without a second, now always-false, breaker check; `recordWatchDeliveryLocked` keeps its meaning and its callers at the send rail's settle end (`settleWatchSendDelivered`). `budgetTripped` semantics are unchanged: one true per config lifetime.

No new test at the attach-scan seam: it is provably a 0-to-1 seam. `fireAttachScan` is reached only from `completeAttachScan`, which `configureWatch` calls on a config it built with `newWatchConfig` on the line above (both the fresh-install and the replace path), and `conditionFires` is never restored from durable state, so a real attach scan cannot approach a budget of 50. Driving it there needs contrived internal state. The new helper is pinned directly by `TestConditionFireBudget_EveryFireGoesThroughOneHelper` (counts, latches once, never counts a delivery), and the existing `TestSendWatchBudgetTripsWhenFramesNeverSettle` and `TestConditionFireBudget_*` stay green. `FuzzJobWatchEventsOutput` case 8 already drives `fireAttachScan` from a contrived `budget-1` and now exercises the trip branch; its seed corpus passes.

**Low — end-reason vocabulary comment** — addressed in `787514b7c`, with a correction to the finding.

`recentWatchEntry.end_reason` is fed only by `recentWatchSummaries` from `jm.watchHistory`, whose only writer is `recordWatchEndedLocked`. Grepping every reason that reaches it gives exactly the six already listed: `auto_removed_terminal`, `cleared`, `replaced`, `budget_exhausted`, `fired`, `job_manager_closed`. `runtime_lost` (`agent/jobs.go:1439`, `clearUnrestoredActiveWatches`) is stamped on durable `watch_cleared` events at restore for watches that were never restored into `jm.watches`; it never reaches the in-memory ring, so listing it as a value of this field would be wrong. The comment now names the ring as the source of the six and says where `runtime_lost` lives instead.

**Low — plan contradicts shipped zero handling** — fixed in `75d711b9d`.

`docs/superpowers/plans/2026-09-02-job-watch-timer.md:16` now reads "the bounds apply to non-zero values only. A present `0` reads as absent, exactly like `progress_interval_ms: 0`, and so does null", matching the spec and `TestWatchIntArg_MaterializedZeroOnCreateReadsAsAbsent`. Two other places in the same plan asserted the rejected design — the step-1 test snippet (`TestWatchIntArg_PresentZeroOnCreateIsRejectedByBounds`) and the step-4 instruction to add an explicit create-time zero rejection to `watchArgsFromToolArgs` — and both now describe what shipped.

Gates, all exit 0: `gofmt -l agent/`, `golangci-lint run ./agent/... ./cmd/... ./internal/...`, `go test ./agent/ -run 'Budget|Watch|Timer|Seqfuzz|seqfuzz|Attach' -count=1`, `go test -race ./agent/ -run 'Budget|Attach' -count=1`, `go test ./agent/ -count=1`, plus `go vet -tags evenerfuzz ./agent/` and the watch fuzz seed corpora.

obra:
## Response to roborev (`1f39a32`)

**Non-atomic budget-latch reset** (Medium) — fixed in `5954cab`. `rollbackWatchBudgetTeardown` replaces the pair of `rollbackWatchConfigSnapshotsRejecting` + `rearmWatchBudgetBreaker` calls: the rejecting marks come off and `budgetTripped` clears in one `jm.mu` critical section, so no condition fire can observe delivery re-enabled with the latch still set. `TestConditionFireBudget_RollbackRearmsTheLatchInOneCriticalSection` holds `jm.mu`, calls the locked helper, reads both flags and then fires the breaker before releasing, so the two resets can only pass as one operation. `TestConditionFireBudget_FailedTeardownRearmsTheBreaker` still covers the end-to-end retry through the fake failing append.

**Negative timer returns wrong error** (Low) — fixed in `4087323`. `watchArgsIsTimer` now reads any present nonzero time field as a timer, so `after_seconds:-5` with the source omitted gets the `self` default and reaches `normalizeWatchArgs`' bounds check instead of returning "source is required". A present zero still reads as absent. `TestJobWatchTool_NegativeTimerReportsTheBoundsError` drives both fields through the tool and expects the bounds text.

**Watch ID minted before validation** (Low) — fixed in `1fa8956`. The mint moves from the key build down to just before `validateWatchConfig`, after target, event-arg, trigger-shape and session-shape validation; the key's `Slot` is filled in there, so one id still threads into both the key and the config. Every use of the key above that point is a clear or a terminal catch-up, and neither carries a slot. `TestConfigureWatch_TimerIDIsMintedAfterValidation` is a source scan: watch ids are random UUIDv7 values from a stateless generator, so a burned id moves no counter and opens no gap a test could read; where the mint sits is the only witness. The `maxLiveTimers` cap check still runs after the mint, since it needs the built config and the lock — a create refused at the cap is the one path that still takes an id.

**Producer/frontend HTML-escaping mismatch** (Low) — not changed, declined for the third time. The mismatch predates this branch: `decodeNotificationEntities` handled `&amp;` on `origin/main` and main already applied it to excerpt body text escaped by the same `escapeNotificationBody`, so this branch only routes timer prose through the existing pair. Widening `escapeNotificationBody` would change every notification body the model reads, which is out of scope here; see https://github.com/prime-radiant-inc/evener/pull/850#issuecomment-5525099753.

Gates on `1fa8956`: `gofmt -l agent/` clean; `golangci-lint run ./agent/... ./cmd/... ./internal/...` 0 issues; `go test ./agent/ -run 'Budget|Watch|Timer|Identity|Args' -count=1` ok; `go test -race ./agent/ -run 'Budget' -count=1` ok; `go test ./agent/internal/tool/ -count=1` ok; `go test ./agent/ -count=1` ok (376s; it needs `-timeout 30m` on a loaded machine, where the git-heavy worktree tests push the package past Go's 10-minute default).

obra:
## Round 7 replies (`a2f2c35`)

**Medium — budget-crossing send frame lost on every send rail.** Changed in `1fb660f0`.

The budget bounds how many times a watch may FIRE, not whether a frame it already produced reaches its receiver, so the teardown now drops nothing. Two parts:

- `onSessionEvent` records its fired sends BEFORE the over-budget teardown, the order `feedJobOutputWithProvenance` and `fireAttachScan` already used. The crossing frame is pending by the time its watch is torn down.
- `autoClearWatchOverBudgetNotification` no longer takes a dropped-send snapshot. It detaches the config and hands it to `terminalFlush` (new `flushDetachedPendingSends`), exactly where `completeExpiredJobWatchesLocked` parks a watch auto-removed by its target's exit. `rejectingDelivery` still freezes the config while the teardown persists — no fire past the crossing can slip a frame in — and comes off as the config is parked, since a rejecting config delivers nothing. Detached, it can gain no new frames: a delivery snapshotted before the teardown no longer resolves to a live watch.

The teardown's cleared notification still rides out with the same event's matched notifications, so `TestWatchBudgetClearCoalescesWithMatchedEvent` still sees one wake. Lock order is unchanged; nothing new touches `pendingJobNotifsMu`.

New test `TestSendWatchDeliversTheBudgetCrossingFrame`: a caller-send watch whose parent drains and accepts every frame. Watched failing first at `delivered frames = 49, want 50`; now 50 frames arrive, the 51st match never happens (the config is gone from `jm.watches`), nothing is left pending, and history records `budget_exhausted`.

One correction to the requested contract. `TestSendWatchBudgetTripsWhenFramesNeverSettle` could not stay green unmodified, and no design can make it: the test's receiver never takes a frame, so "the crossing frame is delivered" and "`hasPendingWatchSends()` is false" are contradictory requirements. That one assertion now pins the flush contract instead — after the teardown, the live set holds zero pending frames and the terminal-flush rail holds exactly the one crossing frame. Everything else in the test (the breaker latching at the match with nothing settled, one budget-cleared notification, the `budget_exhausted` history row) is unchanged and still green.

**Low — timer identity omits mode and note.** Changed in `143739ba`.

`after_seconds`, `repeat_seconds`, and `note` now ride both `normalizedWatchConfigHash` and `watchConfigSnapshot`. All three are `omitempty` and sit after the receiver fields in `WatchConfigSnapshot`, so a non-timer config's JSON encoding is byte-identical and every config hash already in `jobs.jsonl` still matches. The snapshot rebuilds the field the model named from `cfg.oneShot`/`cfg.timerSeconds`, so inspect and history can tell "fires once in 10 minutes" from "fires every 10 minutes".

Two tests, both watched failing first. `TestNormalizedWatchConfigHash_TimerModeAndNoteAreIdentity`: `after 300`, `repeat 300`, and two `repeat 300` configs with different notes all hashed identically before the change (`sha256:ee222d87…` for all four) and are now four distinct hashes. `TestWatchConfigSnapshot_CarriesTimerModeAndNote` pins the durable snapshot.

**Low — hub prose on every job-less watch notification.** Not changed.

This is what the design chose, not a leak from the timer work. The spec says "keep the prose that `splitNotificationExcerpt` already computes, as a `prose` field rendered on watch cards", and Task 9's review approved it. The body a non-timer watch card now shows is one short line, `Watch event triggered: <reason>.`, and the reason is the informative part (which condition fired, or why the watch was cleared) rather than boilerplate the title already carries.

**Low — validation order.** Changed in `86cd03f5`.

`validateWatchTriggerShape` runs the timer rules first, so `repeat_seconds` with `progress_interval_ms` gets `repeat_seconds and progress_interval_ms are mutually exclusive` instead of being pointed at the timer field it had already used. The progress-trigger hint keeps its job: answering requests that named no timer field at all. Table row `timer with progress` in `TestValidateWatchTriggerShape_TimerRules`, watched failing first.

**Gates** (all exit 0): `gofmt -l agent/` clean; `golangci-lint run ./agent/... ./cmd/... ./internal/...` 0 issues; `go test ./agent/ -run 'Budget|Watch|Timer|Send|Args|Identity' -count=1`; `go test -race ./agent/ -run 'Budget|Send' -count=1`; `go test ./agent/internal/tool/ -count=1`; `go test ./agent/ -count=1 -timeout 30m`.

obra:
## Round 8 reply (`86cd03f`)

**Medium — one chunk can outrun the budget.** Changed in `2b69908c`.

Correct, and the per-chunk loop was the only place it could happen: the event rail sees one match per event, and both scan rails fire once on the last match. `feedJobOutputWithProvenance` now breaks out of the match loop as soon as `noteConditionFireLocked` reports the crossing. The crossing match itself is still recorded and delivered — it is inside the budget — and it is the last one that watch gets. The teardown it schedules still runs after `jm.mu` is released.

New test `TestConditionFireBudget_OneChunkStopsAtTheBudget`, watched failing first at `conditionFires:60 deliveries:60` for a single 60-match chunk. It now delivers exactly 50 `output_match:` notifications, counts 50 fires and 50 deliveries, and ends the watch `budget_exhausted`.

Two more from re-reading the teardown while fixing this:

**Ownership of a config a concurrent clear won.** Changed in `7aa22887`. The release step cleared `rejectingDelivery` and parked the config for flush whether or not it was the call that detached it. If a model-requested clear won the same key while the budget teardown was persisting, that clear owns the config — it has already dropped the frames under its own rejecting mark — and unmarking it handed a config that is in neither `jm.watches` nor `terminalFlush` back to a fire still in flight. Detach and release now share one critical section and are gated on the detach actually removing the config; the per-target detach is extracted as `detachWatchConfigSnapshotLocked` so the clear paths and the budget path decide ownership the same way.

It is testable through the existing durable-append seam, so it got a test rather than just a review: `TestConditionFireBudget_TeardownLeavesAConcurrentClearsConfigAlone` runs the concurrent clear inside the append of the teardown's own cleared event, which is exactly the window between the mark and the detach. Watched failing first on `rejecting:false`.

**Stale doc comment.** Changed in `0bbd0857`. `autoClearWatchOverBudgetNotification` still claimed to mirror `clearWatch`'s terminal-snapshot machinery; it stopped taking a terminal snapshot when it started flushing frames instead of dropping them. The comment now describes mark / persist the cleared event / detach, and what the rejecting mark holds across that append.

**Gates** (all exit 0): `gofmt -l agent/` clean; `golangci-lint run ./agent/... ./cmd/... ./internal/...` 0 issues; `go vet -tags evenerfuzz ./agent/...`; `go test ./agent/ -run 'Budget|Send|Watch' -count=1`; `go test -race ./agent/ -run 'Budget|Send' -count=1`; `go test ./agent/ -count=1 -timeout 30m`.

obra:
## Round 9 reply (`0bbd085`)

**Medium — a match landing inside the teardown window.** Changed in `c822180`.

Correct, and the no-send rail is where it bites: the teardown does not detach its config until its durable append returns, so the watch is in `jm.watches` and still matching for that whole window. `rejectingDelivery` catches the send rail at the delivery gate; a notification has no such gate and went straight out.

Fixed at the single seam, as suggested. `noteConditionFireLocked` now returns `(accepted, crossedBudget)` and refuses a match once `budgetTripped` is set — no count, no crossing. All three match sites build nothing from a refused match: the event rail skips the watch, the output rail stops walking the chunk, and the attach scan returns "did not fire". The crossing match itself is still accepted; it is the last fire the budget allows, which is what keeps `TestSendWatchDeliversTheBudgetCrossingFrame` at exactly 50.

New test `TestConditionFireBudget_MatchDuringTeardownIsRefused`: a no-send watch, with an extra match driven through the `jm.appendEvents` seam inside the teardown's own cleared-event append. Watched failing first at `notifications:51 conditionFires:51 deliveries:51`; now 50/50/50. `TestNoteConditionFireLocked_LatchesOnceAtTheBudget` is updated to the new contract — a refused fire no longer increments the counter, so the post-latch total is 50, not 51.

**Low — `source: 123` becomes a self timer.** Changed in `7b753d1b`.

`source` is now read with a typed reader rather than `stringArg`. Absent and null still mean "not named" and keep the self default for a timer create; a present non-string returns `invalid_request: source must be a string`. The distinction matters for this argument specifically because empty is not neutral here — an unnamed source on a timer create becomes `self`, so coercion built a self timer out of a request that named something else. New table `TestWatchArgsFromToolArgs_SourceMustBeAString` covers `123`, `true`, an array, and an object on the reject side, and absent/null/`""`/`"  "` on the self-default side; watched failing first.

**Gates** (all exit 0): `gofmt -l agent/` clean; `golangci-lint run ./agent/... ./cmd/... ./internal/...` 0 issues; `go vet -tags evenerfuzz ./agent/...`; `go test ./agent/ -run 'Budget|Send|Watch|Args' -count=1`; `go test -race ./agent/ -run 'Budget|Send' -count=1`; `go test ./agent/internal/tool/ -count=1`; `go test ./agent/ -count=1 -timeout 30m`.

One note on the full suite: the first run failed `TestSessionCloseMarksBackgroundShellCancelledBeforeEnvCleanup` (a background `sleep 30` still `running` in the store right after `s.Close()`). It touches no watch machinery, passed 5/5 in isolation, and the full suite is green on a re-run; the run that failed was also the slowest (376s vs 348s), so it reads as a load-sensitive flake in that test's read-immediately-after-Close, not something from this branch. Flagging it rather than leaving it unsaid.

PR #852, opened 2026-09-03
test: wait on the real completion in two hub and drain barriers

obra:
Three flaky tests that waited on something other than the condition they actually needed: two completion barriers that stopped one goroutine hop short, and a process fixture that could not take the signal it was waiting for.

**cmd/evener-tui `TestHubModelReReadHoldsPostCutFrameUntilSnapshotApplied`.** `barrierAfterBroadcast` assumed one HarnessList round trip ordered a broadcast issued before it, but while a subscription is still buffering after a subscribed read, `Subscriptions.Route` parks the broadcast in the subscription buffer and only `Server.releaseHydration` — running from the response finalizer, after the read response is already queued — flushes it. A request issued inside that window is answered while the broadcast is still parked, so both non-blocking drains saw nothing and the post-cut frame never reached the model. The barrier now polls until the feed's own open capture is holding the frame above its cut.

**cmd/evener `TestRunDrainsManagedShellWhenSystemPromptNamesTheJobFrame`.** `installHeldRunShell` waited on the fake executor's `SignalName` hook, which `runShell` calls from the wait goroutine before it hands the result to `finalizeShellWhenDone`, so the barrier cleared while the terminal record was still uncommitted and no owner notification existed. The real `DrainJobTree` then started mid-finalization, saw a job that was merely running, armed the undisposed-background-job ladder and announced on the next 250ms recheck tick instead of delivering; an announcement turn's reply is housekeeping the drain discards, so the drain returned `""` and the run printed its pre-drain answer. The barrier now polls the session's job activity tree until every managed job it started is terminal.

**internal/plugins `TestGit_CancellationSendsSIGTERM`.** Two load-dependent modes, both test-side; `git()`'s cancellation path is correct. First, the readiness poll's 5s ceiling was doing the real work: reaching the fake git's readiness file costs two process creations (exec of `/bin/sh`, then of `touch`), measured at up to 10.0s on a host also running the rest of the suite, against a p50 of 0.21s over 690 loaded samples. That ceiling is now a named 90s tripwire, and the poll watches `git()`'s own result channel, so a fake that never execs names that error instead of "file never appeared". Second, the fake parked in `sleep 30 >/dev/null 2>&1 & wait $!`, and the test cancels the moment the readiness file lands — before the shell is parked. A TERM delivered in that window is already pending when the shell's `wait` starts blocking, and `wait` does not notice an already-pending trapped signal, so the trap would have run when the 30s sleep ended, long after `git()`'s 5s `WaitDelay` had SIGKILLed the shell: `git fetch: signal: killed` at 5.002s with the TERM record absent. The fake now parks in bounded foreground sleeps, where a pending trap always runs at the next command boundary, and the TERM record is no longer polled at all — `git()` returns only once the child has been reaped and the trap writes its record before the shell exits, so the record is on disk already or it is never coming.

## Forced-interleaving proof

Each hole was confirmed by widening its window before the fix, and each fix was re-run with the same sleep still in place.

- Hub: a 200ms sleep in `Connection.enqueueResponse` between queuing the response and `finalizer.commit()` reproduced the CI failure verbatim (`transcript=[check the deploy deploy looks healthy]`) on 3/3 runs; with the new barrier the same sleep passed 20/20. Sleep removed.
- Drain: a sleep in the fake executor's `SignalName`, after `waitReturned` closes and before `runShell`'s `waitCh` send, reproduced both failure modes. At 250ms it reproduced the CI failure verbatim (`stdout = "waiting for shell\n"`); at 500ms it tripped the announcement assertion instead. With the new barrier both sleeps passed 6/6 across the managed-shell drain tests. Sleep removed.
- Plugins: this one has no in-repo seam to widen — the window is inside the shell's `wait` builtin — so it was reproduced by load instead, under a repeating `go test ./internal/... -count=1 -p 8` on an 18-core host. The old test failed 1 round of 5 (`termed never appeared`, 10.45s); instrumented runs pinned both modes (10.0s readiness spawns; `signal: killed` at 5.002s in 2 of ~690 runs, plus 1 lost TERM in ~1150 iterations of the same script driven straight from a shell with no Go involved). The fixed test is 930/930 clean across 31 rounds of `-count=30` under that load, and the new park lost 0 of ~950 shell-level iterations.

Test-only changes; no production code.

## Gates

`gofmt -l cmd/`, `go vet ./cmd/evener-tui/ ./cmd/evener/`, `golangci-lint run ./cmd/...` (0 issues), `go test ./cmd/evener-tui/ -run TestHubModelReReadHoldsPostCutFrameUntilSnapshotApplied -count=50`, `go test -race ./cmd/evener/ -run TestRunDrainsManagedShellWhenSystemPromptNamesTheJobFrame -count=20`, `go test ./cmd/evener-tui/ ./cmd/evener/ -count=1`, and `TestNoBareWallClockDeadlineInAgentTests` — all pass. For the plugins fix: `gofmt -l internal/plugins/`, `go vet ./internal/plugins/`, `golangci-lint run ./internal/plugins/...` (0 issues), `go test ./internal/plugins/ -run TestGit_CancellationSendsSIGTERM -count=30` looped under the load above, and `go test ./internal/plugins/ -count=1` — all pass.

## Follow-ups

- `TestRunDrainContinuesWhenNotificationTurnStartsAnotherShell` has no completion barrier by design (shell A is released at drain start) and rests on finalization beating the 250 ms recheck; the same forced delay that reproduced the drain flake fails it too.
- `internal/devtool/procgroup/procgroup_test.go` has the same bare 10 s readiness ceiling for child spawns that the git fixture had.
- `agent/execenv` `TestCleanupRetainsOwnedTmpAfterChildGrace` failed once under full-suite load on 2026-09-03 and passed on re-run and 3/3 in isolation; not diagnosed.

### DIFF
--- a/agent/session_jobtree_drain.go
+++ b/agent/session_jobtree_drain.go
@@ -164,6 +164,30 @@ func (jm *jobManager) hasRunningDrainJob() bool {
 	return false
 }
 
+// ManagedJobsFinalizedForTest reports whether none of this session's own
+// managed jobs is still inside finalization: the running map is clear of them,
+// which — because armFinalizedJob (jobs.go) appends the durable
+// EventJobNotificationPending BEFORE it deletes the running entry — means every
+// completion this drain owes the model is already recorded as a pending owner
+// notification.
+//
+// Test-only seam. cmd/evener's drain tests hold the drain until the shell they
+// launched has completed, and a completion that is merely RECORDED terminal is
+// not that state: until armFinalizedJob deletes the running entry, the job is
+// still in the running map, backgroundDrainState reads the background set off
+// that map alone, and so a drain starting in the window counts a finished shell
+// as a live undisposed background job — arming the announcement ladder on work
+// that was already committed. The residue after the delete needs no barrier:
+// the NotifyPending record written before it is what both the sole-reason test
+// and subtreeHasLiveTerminalDrainWork count, so a drain that starts there waits
+// for the notification instead of announcing.
+func (s *Session) ManagedJobsFinalizedForTest() bool {
+	if s == nil || s.jobManager == nil {
+		return false
+	}
+	return !s.jobManager.hasRunningDrainJob()
+}
+
 // treeHasOutstandingWork reports whether this session or any live descendant in
 // its managed-job subtree still owes drain work: an outstanding managed job, a
 // pending job notification, a pending caller-targeted watch send, a pending
--- a/cmd/evener-tui/hub_read_cut_test.go
+++ b/cmd/evener-tui/hub_read_cut_test.go
@@ -6,6 +6,7 @@ import (
 	"slices"
 	"strings"
 	"testing"
+	"time"
 
 	"primeradiant.com/evener/appwire"
 	"primeradiant.com/evener/internal/appserver"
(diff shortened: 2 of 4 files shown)


obra:
Closed the residual race the review flagged at `cmd/evener/run_drain_test.go:193` (`223728533`).

**What the barrier was missing.** `managedJobsTerminal` cleared as soon as `Job.Terminal` was visible, but a job's terminal record is committed *before* `armFinalizedJob` runs, and the job stays in the job manager's running map through the whole of that function. `backgroundDrainState` reads the drain's background set off the running map alone, so a drain starting in that window counts an already-finished shell as a live undisposed background job, arms the announcement ladder on the pass that finds it, and announces on the next 250ms recheck instead of delivering.

**What it waits on now.** `agent.Session.ManagedJobsFinalizedForTest` — no managed job of the session left in the running map. `armFinalizedJob` appends the durable `EventJobNotificationPending` *before* it deletes the running entry, so clearing this barrier also establishes the durable `NotifyPending` observation the review asked for. The residue after that delete needs no barrier of its own: the `NotifyPending` record is exactly what both the sole-reason test and `subtreeHasLiveTerminalDrainWork` count, so a drain that starts there waits for the notification rather than announcing. The seam is a documented test-only accessor next to the drain predicates it mirrors, following the repo's existing `...ForTest`/`...ForTesting` convention (`hubcore.SeedForTest`, `execenv.SetGitExecTimeoutForTesting`).

**Proof.** A 500ms sleep in `armFinalizedJob` between the `NotifyPending` append and the running-map delete fails the old barrier 3/3 across both prompt layouts — the third model request carries `This run cannot finish with undisposed background job(s)` where the `<job-notification` frame belongs. With the new barrier the same sleep passes 5/5. Sleep removed.

Gates: `gofmt -l`, `go vet ./agent/ ./cmd/evener/`, `golangci-lint run ./agent/... ./cmd/evener/...` (0 issues), `go test -race ./cmd/evener/ -run TestRunDrainsManagedShellWhenSystemPromptNamesTheJobFrame -count=20`, `go test ./cmd/evener/ -count=1`, and `go test ./agent/ -count=1` (includes `TestNoBareWallClockDeadlineInAgentTests`) — all pass.

One thing the forced sleep also exposed, left alone as out of scope: `TestRunDrainContinuesWhenNotificationTurnStartsAnotherShell` fails under it too. That scenario passes `nil` for the completion wait by design — shell A is released at drain start with no barrier at all — so it rests on finalization beating the 250ms recheck. Worth a look if that test is ever seen flaking on a loaded runner.

PR #854, opened 2026-09-03
feat(web): make the ask_user questions widget a scrollable transcript row

obra:
## Why

The pending-questions dock was mounted by the Composer in the session pane's fixed footer, so it pinned itself to the bottom of the screen whenever a question was pending — scrolling back to read context was nearly impossible.

## What changes

The questions widget is now the transcript's **trailing virtual row**: it scrolls away with the content, while the composer still hides/inerts its input row whenever a question is pending.

- **TranscriptBody** gains a `trailingRow` prop: a real virtualized row after the last transcript row, with stable keying, dynamic measurement, and end-anchored following (a new question follows the tail for a reader at the bottom and never yanks a scrolled-back reader).
- **Session.tsx** mounts `AskDock` as that row while askDockStore reports pending batches; answer state already lives in the store, so the virtual list unmounting the row on scroll-away loses nothing.
- **Composer.tsx** drops the dock mount and keeps its own half of the contract (hidden/inert input card, attachments, TasksPanel/CurrentWork, and the "Message composer ready." announcement).
- **askdock.module.css**: `.dock` sizes to content (no internal scroll — which would also lie to `measureElement`) and centers on the 76rem reading measure; the dead `.footer:has([data-ask-response-dock])` rule in panescaffold goes away.
- **Layoutguard**: retired `composer-desktop-question-replacement` and `session-ask-pane-allocation` (they guarded the removed footer-anchored geometry), added `askdock-transcript-row` (desktop: content-sized, centered, transcript owns all scrolling), and reworked `askdock-mobile-tall-crush`; both mutations hand-verified to fail with the right reasons.
- **test/scenarios/ask-web-answer.md** placement description updated.

## Verification

- `make test-web`: PASS (web-typecheck, web-test — 7,515 tests, web-lint)
- Browser guards (layoutguard, overflowguard, shellguard, spawnguard): all PASS. Note: `make test-web-browser` currently fails to launch Chrome under this session's long sandbox TMPDIR (unix-socket path overflow → SIGTRAP) — pre-existing on main, unrelated to this change; the guards were run via the unmodified `scripts/web/test-web-browser.sh` with a short TMPDIR.
- TDD throughout: new trailing-row tests and the Session-level placement test were watched failing before implementation.


obra:
All three findings verified against the code before implementing — each was real. Fixed in `0d3fd3fd5`:

**High — scroll coordinators' row count.** Confirmed: `Session.tsx` passed `renderRows.length` and `TranscriptBody.tsx` passed `rows.length`, so initial end positioning, append-follow, and jump-to-bottom targeted the last transcript row and could leave the dock below the viewport. Both now pass the count including the synthetic `ask-dock` row. Covered by new tests that spy-capture the hook options at both call sites (watched failing, then passing). On the stable-anchor-identity suggestion: the dock row deliberately has no `data-view-anchor-id` entries — view restoration anchors on transcript entries and falls back to the nearest message, which stays correct with the row appended; adding a synthetic anchor would put restoration state on an interactive surface that owns none.

**Medium — focus lifecycle on remount.** Confirmed: `wasEmptyRef` reset on every virtualized remount, re-stealing focus. The activation edge now lives in `askDockStore` (per-ref `pendingGreeted`, reset when the pending set empties) — remounts don't steal focus, a genuinely new question after full resolution still auto-focuses, and the existing no-steal contracts for late-arriving questions are preserved. Two new tests pin both directions.

**Medium — no visible path on in-progress → completed while scrolled away.** Confirmed: the item completing changes no turn/item shape, so the content-changed effect never fired and `pillCount` stayed 0 with the composer hidden. `useTranscriptScroll` now takes an `askDockPending` option; its rising edge for a reader away from the bottom bumps the new-content pill (needs-you styling via the wire's `askPending`), and the pill's jump lands on the dock row. At-bottom readers get no pill (the end-anchored list already followed the row in), and a session opened with an already-pending ask never fires the edge. Three new hook tests.

Gates: `make test-web` green (typecheck, 7,536 unit tests, lint); all four browser guards (layoutguard, overflowguard, shellguard, spawnguard) green.

obra:
Round-2 finding verified correct: PaneScaffold's `.body` never scrolls in the session pane — the VirtualList's own `.root` (`virtuallist.module.css`) is the real scroll owner, so the original harness guarded (and mutation-tested) the wrong rule. Fixed in `533140be1`: the harness now mirrors the full `.body > .transcript > .transcriptContent > .transcriptList > .root > .sizer` chain with the real virtuallist classes, assertions target the root's geometry, and the mutation reverts `.root { overflow-y: auto }` — hand-verified to fail with 'transcript overflow belongs to visible, not the virtual list's root'. `askdock-mobile-tall-crush`'s mirror comment was corrected to name the same owner. Gates: `make test-web` green; all four browser guards green.

obra:
Round-3 finding verified correct: a default `focus()` on the activation edge would scroll the transcript to the control when the dock mounts in overscan while the reader is scrolled away, defeating the new-content pill. Fixed in `4de59d405`: the effect now calls `focus({ preventScroll: true })` — the pill stays the reader's way down, and its jump lands on the dock with focus already in the first answer control. Pinned by a focus-spy test asserting the option. The scrolled-away activation path itself is covered by the round-2 hook tests. Gates: `make test-web` green; all four browser guards green.

obra:
Round-4 findings — both verified real, fixed in `de84f55d5`:

1. **Pill not needs-you on live asks**: confirmed — `model.askPending` is snapshot-authoritative (only `hydrateThread` sets it; no notification carries it, per `reducer.test.ts`), so a live-arriving ask bumped the pill count without needs-you styling. `pillNeedsYou` now includes the live `askDockPending` signal alongside `isAttentionWorthy(model)`.
2. **Free-text focus stolen on remount**: confirmed — `prevKindRef` started at `null`, so a virtualized remount with an active free answer read as a null→free transition. It now initializes with the mount-time kind; genuine transitions still focus the field.

Both test-first (RED: `pillNeedsYou === false` with `askPending: false` throughout; focus landed in the answer input on mount). Gates: `make test-web` green; all four browser guards green.

obra:
Round-5 findings — both verified real, fixed in `9feaa3444`:

1. **Focus theft despite preventScroll (Medium)**: correct — preventScroll stops the scroll, not the focus transfer to an off-screen control mounted in overscan. Activation focus is now IntersectionObserver-gated: the dock focuses its first answer control only once it actually intersects the viewport. This composes with the round-2 pill: its jump scrolls the dock into view, and the intersection is what lands the focus (still with preventScroll, so the pill's targeting stays in charge). A scriptable-observer test pins off-screen no-steal + visible-focus, and the un-greeted store flag means the visibility arrival still triggers it.
2. **Live regions re-announcing on remount (Low)**: correct — the row's prompt caption and count were static aria-live regions inside a virtualized row. The row is now visual-only; the single live region is a new `AskDockAnnouncements` component mounted by Session **outside** the virtual list, announcing only real transitions (pending arrival, answered-count changes, resolution clearing). A Session-level test pins the region outside the list.

Gates: `make test-web` green; all four browser guards green.

obra:
Round-6 findings — both verified real, fixed in `12fbf937c`:

1. **Announcements not ref-aware (Medium)**: correct — panes are reused across refs (sidebar click swaps the session on a persistent pane). `AskDockAnnouncements` now includes `sessionRef` in its previous snapshot, treats a ref change as a fresh activation, and remounts its announcement content via a key so identical-text re-announcements are still audible (a live region announces mutations; unchanged text is none). Pinned by a ref-switch test asserting a new announcement node.
2. **Greeting/pill suppressed on atomic replacement (Medium)**: correct — the live-notification path self-heals through the empty transition, but a snapshot resync can swap disjoint batch sets atomically. `askDockStore` now tracks a per-ref `activationEpoch` (bumped on empty→non-empty and on full replacement; same-set additions deliberately don't bump), `pendingGreeted` re-arms on replacement, and the scroll hook's pill edge keys on the epoch rather than the boolean — which could never express 'still pending, but a different question now'.

Tests: store-level snapshot-replacement and same-set-addition, hook-level epoch-edge (including replacement with pending true throughout), announcements ref-switch — all watched failing first. Gates: `make test-web` green; all four browser guards green.

obra:
Round-7 findings — all three verified, fixed in `908b7b515`:

1. **Announcements ignore the epoch (Medium)**: correct — a replacement with an identical answered count moved neither `pending` nor `count`. `AskDockAnnouncements` now announces the prompt on every `activationEpoch` bump (fresh activations and replacements alike), with the keyed content remount making identical text audible. RED test: snapshot resync swapping the batch set with the same count produces a new announcement node.
2. **Focus on an invisible control in a tall dock (Medium)**: correct — the gate was on the whole dock while focus went to its first control, and the pill's end-aligned jump can leave that control above the fold. The visibility gate still suppresses off-screen activation entirely (no focus call), but once the dock IS visible the effect uses plain `focus()` so the browser reveals the control. The round-3 `preventScroll` pin was reworked into that superseding contract.
3. **Stale scenario doc (Low)**: `test/scenarios/ask-web-answer.md` now queries `[data-testid="ask-dock-announcements"]` outside the virtual list, drops the removed in-dock `role="status"`/`aria-live` references, and asserts the dock's transcript-row placement.

Gates: `make test-web` green; all four browser guards green.

obra:
Merge conflict with main resolved in `ce53d63a4`: main's new pill-visibility contract (`pillVisible` as a scroll-position affordance, needs-you gated on visibility) is kept, with this branch's live dock signal carried into the predicate (`pillNeedsYou: pillVisible && (isAttentionWorthy(model) || askDockPending)`). Gates on the merged tree: `make test-web` green; all five browser guards green (including the new transcriptscrollguard from main).

obra:
Round-8 finding (case.json cssFiles) — I verified before implementing and this one is **incorrect**, so I'm not changing the code:

`askdock-mobile-tall-crush/case.json` lines 5–9 already include `panes/session/composer/askDock/askquestioncard.module.css` in `cssFiles` (the exact path the fix proposes adding), and the file exists at that path on the reviewed head.

Direct evidence on `ce53d63a4`, re-run just now:
1. Case as committed: **PASS**.
2. With the mutation applied (`.option { overflow-wrap: break-word }` → `normal`): **FAIL** with the expected reason — `option "cherry-pick-a323417b7…-onto-release/1.4": row (261px) or its label escapes the card horizontally` — which is only possible if the stylesheet is loaded and the declaration is live.
3. Restored: **PASS** again.

If the mutation's target rule weren't loaded, step 2 would have passed unchanged. No change needed; flagging in case the reviewer wants to double-check its cssFiles read.

PR #856, opened 2026-09-03
fix(llm): clamp effort-derived thinking budget under the output ceiling

obra:
## Summary

- Clamps the effort-derived anthropic `thinking.budget_tokens` under the request's output ceiling, so the protocol's `max_tokens > budget_tokens` contract always holds for a request the adapter itself derived.

## Why

Production incident (session `034IY3Zk6JSKI7i9eVhTRc`, 2026-09-03): `llm.ReasoningBudget` maps `max`/`xhigh` to 131072 tokens — exactly `kimi-for-coding` k3's `max_output_tokens` — so once the session selected effort `max`, **every** model call was rejected before provider dispatch with a non-retryable error:

```
Evener blocked kimi-code/k3 before provider dispatch: token budget exceeds max_output_tokens (input=0 output=131073 maximum=131072)
```

The API log showed the failed turns as zero-attempt `transport_failure` groups (no HTTP attempt was ever made), the transcript had no assistant turn after the user's messages, and there was no retry or recovery short of lowering the effort — the session looked dead while still accepting user input. The same wedge is reachable via loop-detector escalation to "high" on budget-shaped rows that have no catalog output cap (the 32000 fallback becomes the ceiling).

## What changed

- `applyThinkingShape` clamps the effort-derived budget via `fitThinkingBudgetToOutputCeiling`, floored at Anthropic's documented 1024-token thinking minimum (now shared as `llm.MinimumThinkingBudgetTokens`, hoisted beside the table whose minimal row already encoded it).
- Deliberately unchanged: #824's guard still errors for caller-provided `ProviderOptions` budgets that don't fit, and for ceilings too small to fit the thinking minimum. The clamp never *raises* `max_tokens` — the thing that guard exists to prevent (per the [plan](docs/superpowers/plans/2026-09-01-provider-safe-token-budget.md), step 1).
- `TestProtocolBuildBody_HighThinkingRejectsUnknownOutputCap` → `...ClampsUnderUnknownOutputCapFallback`: it pinned the over-generalized fail-closed that produced this wedge. Commit 58196ab16 carries the full reasoning.
- The second commit (7cbbfc3fa) is a behavior-neutral simplify pass over the fix.

## Verification

- TDD: the new regression test reproduced the exact incident error text before the fix.
- End-to-end against the real registry row: `kimi-code/k3` at effort `max` now builds `max_tokens=131072, budget_tokens=131071, output_config.effort=max`.
- `make merge-approval-gate`: **PASS** (exit 0) — lint, build, `ROOT_FULL=1 make test` (root 95s, llm, agent, auth, envvars, invariant, identifier, web 92s) all green on this branch.

### DIFF
--- a/docs/llm-providers.md
+++ b/docs/llm-providers.md
@@ -684,6 +684,11 @@ above). On the anthropic protocol, a row whose `ThinkingShape` is `adaptive`
 (e.g. Opus 4.6, Sonnet 4.6, and later) sends `output_config.effort`; a row
 whose shape is `budget` or `budget+effort` (e.g. Opus 4.5, `kimi-for-coding`)
 maps the effort to `thinking.budget_tokens`.
+That derived budget is clamped to leave the one token of headroom the
+protocol's `max_tokens > budget_tokens` contract requires, so an effort whose
+table budget meets the row's output cap (max and xhigh map to 131072 —
+`kimi-for-coding` k3's cap) still yields a sendable request instead of a
+pre-dispatch `max_output_tokens` budget error.
 
 **Setting it.** Launch: `--reasoning-effort`, `EVENER_REASONING_EFFORT`,
 `reasoning_effort` in `launch.toml`, or the spawn-form effort chip
--- a/llm/providers/anthropic/protocol_request.go
+++ b/llm/providers/anthropic/protocol_request.go
@@ -99,11 +99,8 @@ func buildProtocolBodyForOperation(req llm.Request, res registry.Resolved, enfor
 // always-on adaptive rows included: no Claude row lists an off effort level,
 // so there is no value that says "off" here, and keeping the always-on body
 // would switch thinking on against the user's stated intent. An unset shape
-// sends nothing. A budget the output ceiling cannot fit is reduced to fit —
-// the completion contract requires max_tokens to strictly exceed
-// budget_tokens, so a budget equal to the ceiling (ReasoningBudget maps
-// max/xhigh onto exactly the 131072-token cap of rows like kimi-for-coding
-// k3) is a request this adapter could never send.
+// sends nothing. A budget the output ceiling cannot fit is reduced to fit
+// (fitThinkingBudgetToOutputCeiling).
 func applyThinkingShape(body map[string]any, req llm.Request, caps registry.Caps) {
 	if req.ReasoningEffort != nil && *req.ReasoningEffort == "none" {
 		return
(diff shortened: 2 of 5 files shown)


obra:
Addressed in d96adbe79: verified the finding against the code — `CountTokens` builds with `enforceCompletionContract=false` and deletes `max_tokens` as an output-side field, so the clamp was rewriting the effort-derived budget to satisfy a contract the count path doesn't have. `applyThinkingShape` now clamps only on the completion path; token-count bodies keep the effort-derived budget (pinned by a new test asserting a high-effort/8192-admitted request counts with `budget_tokens=32768`, previously 8191). RED→GREEN: the new test reproduced the reported 8191 before the fix; full anthropic package, vet, gofmt, lint green.

obra:
Addressed in 312ce2202: verified against the code — the clamp ran before the ProviderOptions overlay, so a caller's explicit lower `anthropic.max_tokens` override landed after it and could leave `max_tokens <= budget_tokens`. The clamp's ceiling now includes the override value the overlay will write (parsed with the same PositiveInt numeric forms the final body value uses). Caller-supplied `thinking.budget_tokens` keeps the explicit error, as does an override too small to fit the 1024-token thinking minimum (both pinned by a new test, RED→GREEN).

obra:
Addressed in d34ae8b52.

**The clamp now includes the transport body constant.** `applyThinkingShape` receives the `Resolved` record, and `fitThinkingBudgetToOutputCeiling` takes the row's `Transport.Body["max_tokens"]` (via the new `transportBodyMaxTokens` helper, mirroring `providerOptionMaxTokens`) as a fourth ceiling bound. Transport body constants run after the body builder and the ProviderOptions overlay and override both, so the row constant bounds the effort-derived budget the same way the body value, the `anthropic.max_tokens` override, and `max_output_tokens` already did.

**Added the requested transport-level integration test** — `TestProtocolCompletionClampsEffortBudgetUnderTransportMaxTokensConstant` (complete + stream): an effort-derived 32768-token budget with `Transport.Body = {"max_tokens": 8192}` and a 131072 cap clamps to 8191 on the wire. Before the fix the finalizer rejected the request with `ContextBudgetError` (`token budget exceeds max_output_tokens ... maximum=8192`) — verified red before implementing.

The count-tokens path stays unclamped (it enforces no completion contract and strips `max_tokens`), and the #824 guard on caller-supplied thinking budgets is untouched.

Gates: `go test ./llm/providers/anthropic/ -count=1`, `go vet`, `gofmt`, `golangci-lint run ./llm/providers/anthropic/...` — all green.

Issue #329, opened 2026-08-22
Finished one-shot run doesn't exit: undelivered notifications and delegate-lane watches hold the terminal drain; empty-response retry can override shutdown


obra:
## Summary

A one-shot run that has *finished* — the model sent `communicate` with `end_turn: true` under `turn_ends_process: true` — does not reliably terminate the process. In a 41-trial focused run (proto-retest3 stack, Terminal-Bench 2.1), the shutdown path damaged 5 trials:

| trial | post-end_turn behavior | cost |
|---|---|---|
| train-fasttext | dead hang, killed by harness after ~1272s | AgentTimeoutError |
| count-dataset-tokens | dead hang, ~544s | AgentTimeoutError (would have finished inside budget) |
| filter-js-from-html | dead hang, ~43s | AgentTimeoutError (would have finished inside budget) |
| sanitize-git-repo | runtime made another provider call after the final communicate, got an empty response, injected a "please continue" retry instead of exiting | ate the remaining margin |
| dna-assembly | shutdown drain raced watch-send delivery | #327 fatal (`stable watch delivery is not the durable pending head`) |

Control: mcmc-sampling-stan, same binary and shutdown path, exited cleanly in 10.6s.

## Discriminator

Every damaged trial had, at the moment of `end_turn`, either:
- `job_notification_pending > job_notification_delivered` (undelivered pending notifications), or
- an outstanding watch on a delegate lane (`dlg_*`).

The clean control had pending == delivered and no delegate-lane watch. The drain treats these leftovers as live work and waits on them forever; nothing on the shutdown path delivers, cancels, or expires them once the model has declared the turn (and process) over.

## Two distinct defects

1. **Drain-side**: after a terminal `end_turn` + `turn_ends_process`, undelivered notifications and delegate-lane watches must not hold the drain. There is no model left to deliver *to* — the turn that ends the process is by definition the last turn. These should be discarded (or at most logged), and the process should exit.
2. **Loop-side** (the sanitize variant): the round loop issued another provider call after the terminal communicate, and the empty-response retry path then overrode process shutdown. Once a terminal communicate has been accepted, no further provider call should be possible.

## Relation to existing issues

[...]

PR #864, opened 2026-09-04
fix(drain): keep the answer when a completion rides the announcement turn

obra:
## The window

In a one-shot run, `DrainJobTree` announces an undisposed background job on the recheck that finds it still in the running map, and that decision (`backgroundDrainState` → `outstandingDrainJobIDsByBackground`) reads only the running map and the queue gate, never `run.terminal`. A job that finishes between the pass's queue gate (`peekNotifications() == 0`) and the announcement turn's own queue drain (`acceptNotificationInput` → `drainJobNotifications`) has its completion queued in time to ride the announcement request, and a job whose terminal record was written but whose running entry `armFinalizedJob` had not yet deleted still counted as a live undisposed shell. The drain then discarded the announcement turn's reply as housekeeping, so the model's real answer to the completion was lost and `run` printed its pre-drain answer.

## Reproduction

`TestOneShotDrainKeepsAnswerWhenCompletionRidesAnnouncement` (agent package) runs the real drain loop and the real `ProcessInputKind` against a background shell on a held fake executor; the process wrapper finishes the job after the drain has queued the announcement and waits for the completion to be queued before opening the turn. Against `main` it produced one model request carrying both the `<job-notification>` reminder and the announcement, the scripted reply `FINAL-ANSWER: the build passed`, and a drain result of `""` (durable record `delivered`, block in history). `TestOneShotDrainDeliversInsteadOfAnnouncingAFinalizingJob` uses the production `testOnlyAfterNotifyPendingAppend` hook to hand the drain its recheck while the job is terminal-in-manager but still in the running map; against `main` the drain announced twice and returned `""`.

## What each change closes

- **Ride-along detection** (`4b001153a`): the session counts job notifications as `acceptNotificationInput` delivers them; the drain snapshots the count around the announcement `process` call and keeps the reply as its result when the count grew during the turn, exactly as a notification turn's reply is kept. A reply to a request that carried no completion is still dropped.
- **Narrowing** (`81297ccc2`): a job with a terminal record is excluded from the background set, so a recheck that meets a finalizing job waits for its notification instead of announcing a finished job; and the drain re-checks its wake edge before committing to an announcement, so a completion that finalized while the pass looked is delivered instead. The wake re-check has no deterministic test: the span it guards has no production seam a test can hold.

Not in this PR: a completion that finalizes during an announcement turn whose reply is the session's first terminal communicate is consumed by the #329 cut and never shown to the model. Separate finding, separate PR.

### DIFF
--- a/agent/session.go
+++ b/agent/session.go
@@ -481,6 +481,12 @@ type Session struct {
 	// one second (captureTerminalNotificationCut).
 	pendingJobNotifsMu sync.Mutex
 	pendingJobNotifs   []jobNotification
+	// jobNotifsDelivered counts the job notifications acceptNotificationInput
+	// has delivered to the model over the session's lifetime. The one-shot
+	// drain reads it across an undisposed-job announcement turn: growth means a
+	// completion rode that request, so the reply is an answer, not
+	// housekeeping. Guarded by pendingJobNotifsMu.
+	jobNotifsDelivered uint64
 	// nextJobNotifSeq gives every queue entry a process-lifetime identity. The
 	// terminal acceptance cut snapshots this watermark while jm.mu prevents a
 	// finalizer from crossing its durable-pending/running-map boundary.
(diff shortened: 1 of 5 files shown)


obra:
Roborev's finding on `7994b30` — not changed.

The window it names is real but already covered. `run.terminal` is assigned under `jm.mu` right after the finished event is persisted, so a drain pass can see the record on disk and a nil pointer for the microseconds in between. In that window the drain behaves as it did before this PR: it may announce. The second half of this PR is the backstop for exactly that case: if the completion's notification then rides the announcement request, the counter grows during the turn and the reply is kept as the answer instead of discarded. Consulting the durable terminal record on every drain pass would trade a microsecond window for a store read per pass, and would still leave the in-memory map as the thing that decides what is background. The `run.terminal == nil` exclusion is best-effort narrowing; the ride-along detection is the guarantee.

PR #866, opened 2026-09-04
fix(plugins): reclaim parked and stale conflict slots

obra:
## The leak

When a published bundled directory holds content the running binary
did not publish, a launch sets it aside to `<dest>.conflict` and
republishes. During that swap the previous occupant of `.conflict` is
parked at `<dest>.conflict.previous`; on success the parked copy is
removed, on failure it is renamed back. A crash between the two
renames, or a failed final remove, leaves `.conflict.previous` behind
forever. Separately, a `.conflict` or `.conflict.previous` left by an
older binary is never revisited once a version bump changes the
plugin's digest. Neither the abandoned-staging sweep (only reclaims
marked `.stage-*` directories) nor `Gc` (walks the install cache, not
the bundled store) ever reclaims either.

## The rule

The bundled-publish reclaim pass, which already sweeps abandoned
`.stage-*` staging directories, now also reclaims:

- a `.conflict.previous` entry old enough that no concurrent swap
  could still be using it (the same `abandonedStaging` age threshold
  and store lock the staging sweep already relies on for the same
  safety argument)
- any `.conflict` or `.conflict.previous` entry whose digest no
  longer names a plugin this binary ships, at any age — a version
  bump means no set-aside or restore this binary performs will ever
  touch that name again

A `.conflict` for the digest this binary currently ships is never
touched, since it is the single preserved copy `setAsideBundledConflict`
promises to keep.

## What the tests prove

`TestMaterializeBundledPlugin_ReclaimsConflictSlots` in
`internal/plugins/resolve_test.go`:

- an aged `.conflict.previous` beside a published copy is removed by
  the next publish
- a fresh `.conflict.previous` survives (still possibly mid-swap)
- a `.conflict` for a stale digest is removed
- a `.conflict` for the current digest is never reclaimed, even when
  the sweep runs and reclaims a stale entry planted alongside it

All watched failing before the fix.

## Gates

`gofmt -l internal/`, `go vet ./internal/plugins/...`,
`GOOS=windows go vet ./internal/plugins/...`,
`go vet -tags evenerfuzz ./internal/...`, `golangci-lint run
./internal/...`, `go test -race ./internal/plugins/... -count=1`,
`go test ./internal/... -count=1` all pass.

### DIFF
--- a/internal/plugins/resolve.go
+++ b/internal/plugins/resolve.go
@@ -350,8 +350,9 @@ func nothingForACallerThatLeft(err error) (LaunchPluginResolution, error) {
 // before a later publish reclaims it: orders of magnitude longer than the copy
 // it names, so a publish in flight is never disturbed. conflictSuffix names the
 // single sibling slot a destination is moved to when it holds content this
-// build did not publish; it is outside the staging namespace, so the sweep
-// never collects it.
+// build did not publish; it is outside the staging namespace, so the staging
+// sweep never collects it, but the same publish sweeps it too — see
+// reclaimableBundledConflicts.
 const (
 	stagingPrefix    = ".stage-"
 	stagingMarker    = ".evener-staging"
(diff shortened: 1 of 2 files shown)


obra:
Fixed in 19fb8012b: the sweep now checks whether a `.conflict.previous` entry's sibling `.conflict` exists before applying the age/digest checks. If the sibling is absent, the entry is an interrupted set-aside (crashed between renaming `.conflict` to `.previous` and renaming `dest` into the now-vacant `.conflict`) and is never reclaimed, regardless of age or digest — its mtime is inherited from whenever the content was first set aside, so it can already read as aged the instant the crash happens. Only a `.previous` beside an existing `.conflict` (a completed swap whose final cleanup failed) is eligible for reclaim.

Added `TestMaterializeBundledPlugin_ReclaimsConflictSlots/an_aged_conflict.previous_with_no_sibling_conflict_is_never_reclaimed`, watched failing against the prior code, and updated the two existing aged/fresh `.conflict.previous` subtests to plant a sibling `.conflict` so they exercise the completed-swap case the age/digest checks are meant for.

`setAsideBundledConflict`'s own recovery rename (`.previous` back to `.conflict` when `.conflict` is absent) is unchanged and still runs the next time that dest is classified as a conflict.

Gates: gofmt, go vet, GOOS=windows go vet, go vet -tags evenerfuzz, golangci-lint, go test -race ./internal/plugins/..., go test ./internal/... all pass.

PR #867, opened 2026-09-04
fix(drain): deliver a completion that lands during the final answer instead of consuming it

### LINKED ISSUE #865: One-shot final answer ignores a job that finishes during its generation
In a one-shot run (`TurnEndsProcess`), a background job that finishes while the model is generating its final answer is never shown to the model. The final-answer request is built with the job still running; the job then finalizes (`agent/jobs.go` `armFinalizedJob`: durable `EventJobNotificationPending`, running-map delete, enqueue) before the reply arrives. When the reply's `communicate(end_turn=true)` is accepted (`agent/session_tool_round.go` `deliverIfCommunicated`), `acceptTerminalCommunicate` captures a "terminal cut" (`agent/session_jobtree_drain_terminal.go` `captureTerminalNotificationCut`) that treats every owned `NotifyPending` record whose job is already out of the running map as a pre-terminal leftover and durably appends `EventJobNotificationConsumed` for it. The queued notification is then filtered out as not deliverable (`jobstore.ShouldDeliver`), and `DrainJobTree` (`agent/session_jobtree_drain.go`) drops the queue entry via `discardTerminalDrainLeftovers`. The model answered before it knew the job's outcome, and nothing ever tells it.

This matters because the run's printed answer is wrong whenever the answer depended on that job (a build, a test run, a long command the model started and then wrapped up too early), and the failure is invisible: the ledger records the notification as consumed, which is the same state a deliberate `job_status` read leaves, so nothing in the transcript or in `evener-doctor` shows a completion was silently dropped. This happens often in practice: launching a background job and ending the turn while it is still running is the ordinary handoff shape, and the job frequently finishes inside the model's generation window.

Approved fix: stop consuming at the terminal cut. When a final-answer turn ends and the queue holds a completion the model was never shown, deliver it through the normal notification path as one more notification turn; the reply to that turn is the new final answer (replacing the earlier one), and if the model starts new work instead the drain keeps draining as it does today. Loop until a final answer's turn carried no unseen completion; each pass delivers at least one notification, so it terminates unless the model keeps starting jobs, which the existing drain bounds (stall watchdog, undisposed-job announcement ladder) already cover. Remove the first-wins latch around the cut, since with no consume there is nothing for it to protect, and delete the dead code that leaves behind.

obra:
## Mechanism

In a one-shot run the final-answer request is built while a background job is still running. The job finalizes before the reply arrives (`armFinalizedJob`: durable `EventJobNotificationPending`, running-map delete, enqueue). Accepting that reply's `communicate(end_turn=true)` captured a "terminal cut" (`captureTerminalNotificationCut`) that treated every owned `NotifyPending` record whose job had already left the running map as a pre-terminal leftover and durably appended `EventJobNotificationConsumed` for it. The queued notification then failed `jobstore.ShouldDeliver`, `DrainJobTree` dropped the queue entry, and the run printed an answer written before the model knew the job's outcome. The ledger showed the same "consumed" state a deliberate `job_status` read leaves, so nothing revealed the drop.

## Approved design

1. Drop the consume. A completion the model was never shown is ordinary queued work: it is delivered on one more notification turn through the normal path, with the same framing as any job notification.
2. The reply to that turn is the new final answer. If the model starts new work instead, the drain keeps draining as today.
3. Loop until a final answer's turn carried no unseen completion. Every pass delivers at least one notification, so it terminates unless the model keeps launching jobs, which the drain's existing bounds (stall watchdog, undisposed-job announcement ladder) already cover.
4. Remove the first-wins latch around the cut and the dead code it leaves behind. `acceptTerminalCommunicate` is now a plain sticky boolean; its two remaining readers are unchanged (#329: residue abandonment with no live work behind it, and an empty post-terminal notification turn finishing idle instead of retrying).

One thing the design did not name, found while watching the first test after deleting the consume: `ProcessInputKind`'s own drain ladder ran the notification turn itself and returned the two answers joined with "\n", so the earlier answer was never replaced. The ladder now leaves post-terminal notification work to `DrainJobTree`, whose non-empty result `cmd/evener run` and `drainForFinalization` already substitute for the earlier answer. That is the only non-deletion change to production logic.

Cleanup: `queueSeq`/`nextJobNotifSeq` existed only for the cut and are gone; the consume helper's `removeQueued` flag (only ever false for the cut) folds back into the single status-read path.

One narrow difference from the old sequence-keyed cut, noted for the record: a delegate that drains twice (`drainForFinalization` at subagents.go:1597 and again at :1630 after a nudge) cuts at the second entry any watch notification queued during the first drain, which the old cut kept; narrow, and arguably right, since the nudge reply is that delegate's final answer.

## What the tests prove

- `TestOneShotDrainDeliversCompletionThatLandsDuringTheFinalAnswer`: a real background shell is finished from inside the scripted final-answer step, after the request is built and before the reply. `ProcessInput` returns the first answer; `DrainJobTree` returns the reply of one extra notification turn that carried the completion; the record is `NotifyDelivered` and the raw ledger holds no consumed event. Failed before the fix (drain returned `""`), and again after deleting only the consume (`ProcessInput` returned both answers joined), which is what justified the ladder change.
- `TestOneShotDrainKeepsDrainingWhenTheDeliveredCompletionStartsMoreWork`: the notification turn launches a second shell instead of answering; the drain continues and finishes with that job's outcome, both records delivered and never consumed. Failed before the fix.
- `TestOneShotDrainReturnsAfterAFinalAnswerThatSawEveryCompletion`: when the final answer's turn already carried the completion, the drain returns with no extra provider call. Passes before and after; pins the loop's exit.

The four terminal-cut tests are removed with the cut: two pinned the consume, two drove the cut's internals directly.

Gates: `gofmt`, `go vet` (plain and `-tags evenerfuzz`), `golangci-lint` (0 issues), agent `Drain|Terminal|Notif|Requeue` subset, `-race` `Drain|Terminal` subset, cmd/evener `Drain|Run` subset, and the full `go test ./agent/ -count=1 -timeout 30m` (450s) all pass.

Closes #865

### DIFF
--- a/agent/job_shell_test.go
+++ b/agent/job_shell_test.go
@@ -36,16 +36,25 @@ func newFakeClockShellTestRig(t *testing.T) (*jobManager, execenv.StreamingExecu
 	return jm, env, clk
 }
 
-func waitForShellDone(t *testing.T, jm *jobManager, jobID string) {
-	t.Helper()
+// shellDoneChannel returns the running job's done channel, which
+// armFinalizedJob closes only after the owner notification is enqueued, and
+// false when the job is no longer in the running map.
+func shellDoneChannel(jm *jobManager, jobID string) (<-chan struct{}, bool) {
 	jm.mu.Lock()
+	defer jm.mu.Unlock()
 	run := jm.running[jobID]
 	if run == nil {
-		jm.mu.Unlock()
+		return nil, false
+	}
+	return run.done, true
+}
+
+func waitForShellDone(t *testing.T, jm *jobManager, jobID string) {
+	t.Helper()
+	done, live := shellDoneChannel(jm, jobID)
+	if !live {
 		return
 	}
-	done := run.done
-	jm.mu.Unlock()
 
 	select {
 	case <-done:
--- a/agent/jobs.go
+++ b/agent/jobs.go
@@ -520,12 +520,9 @@ type jobNotification struct {
 	// payload: a job.notification watch carries the completed job's status.
 	Kind                                                       jobNotificationKind
 	JobID, JobType, Status, Reason, Description, TranscriptRef string
-	// TerminalGen is the exact durable terminal generation represented by a
-	// terminal notification. queueSeq is its in-memory queue identity. Together
-	// they let terminal acceptance distinguish a pre-cut leftover from a later
-	// completion even if a job ID appears in both sets.
+	// TerminalGen is the exact durable terminal generation this terminal
+	// notification represents.
 	TerminalGen      string
-	queueSeq         uint64
 	ExhaustionBudget string
 	ExhaustionLimit  int
 	Resumable        *bool
--- a/agent/session.go
+++ b/agent/session.go
@@ -396,17 +396,12 @@ type Session struct {
 	// terminalCommunicateAccepted latches that a communicate with
 	// end_turn=true completed a turn while TurnEndsProcess: the model has
 	// explicitly ended the turn that ends the process. Unlike comm it is never
-	// reset — the one-shot drain and the round loop read it to refuse to
-	// resurrect a run the model already declared over (issue #329). Guarded by
-	// mu.
+	// reset — the one-shot drain reads it to abandon residue with no live work
+	// behind it, and the round loop reads it to finish an empty post-terminal
+	// notification turn idle instead of retrying (issue #329). A completion
+	// the model was never shown is still delivered after it (#865). Guarded
+	// by mu.
 	terminalCommunicateAccepted bool
-	// terminalNotificationCut is captured at the same acceptance boundary. It
-	// identifies exactly which durable terminal generations and in-memory queue
-	// entries existed before the terminal communicate, so a completion that
-	// lands before DrainJobTree enters cannot be mistaken for a leftover.
-	// Guarded by mu; queue sequence assignment is guarded separately by
-	// pendingJobNotifsMu.
-	terminalNotificationCut terminalNotificationCut
 
 	// askPending is the per-turn pending set of questions posted by ask_user
 	// calls this turn (spec §5.1): its length lets a round-boundary check tell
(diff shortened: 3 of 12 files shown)


obra:
Both roborev findings on `32ca17f` are addressed on the branch.

**Medium (`session_jobtree_drain.go` turn gate dispatched every queued notification after the terminal communicate)**: fixed in 46174203f. The drain now cuts the watch-kind entries (`isWatch()`: output-match frames, watch-send tokens, timer ticks) from the queue on entry when the pre-drain turn ended with a terminal communicate, records them in a warning naming each dropped item, and leaves completions in place for the loop to deliver as before. The cut sits at drain entry, where the old cut was, and deliberately not at acceptance: only the pre-drain final answer precedes the drain, so a frame a drain turn arms later (the watch the model creates in answer to the undisposed-job announcement, and the notice that its delivery budget cleared it) still reaches the model even though that turn's reply is itself a terminal communicate. A first draft cut at first acceptance and dropped exactly that notice, caught by `TestClearedWatchStartsAFreshAnnouncementEpisode` in the full suite. The "recorded consumed" half is the warning event: watch notifications are in-memory wakes with no durable notification record to mark, which is also all the old cut recorded for them. New test `TestOneShotDrainDiscardsWatchFramesQueuedBeforeTheFinalAnswer` queues an output-match frame and the job's completion during the final answer's generation and checks that one extra turn carries the completion alone, the frame's discard is recorded, the queue is empty afterwards, and the completion is delivered and never consumed; it failed first with no discard recorded.

**Low (`session_jobtree_drain_terminal_test.go` polled `peekNotifications` with `time.Sleep`)**: fixed in 4e0414b0b. The poll is replaced by `finalizeShell`, which takes the running job's done channel before releasing the shell and blocks on it; `armFinalizedJob` closes that channel only after the owner notification is enqueued, so on return the completion is queued and the job has left the running map, with no timing assumption. The 30s select bound remains as a hang guard only. `waitForShellDone` shares the new `shellDoneChannel` lookup.

Gates on the pushed head: gofmt, go vet (plain and `-tags evenerfuzz`), golangci-lint 0 issues, agent `Drain|Terminal|Notif`, `-race Drain|Terminal`, cmd/evener `Drain|Run`, and the full `go test ./agent/ -count=1 -timeout 30m` all pass.

PR #868, opened 2026-09-04
test(plugins): cover the digest walk's per-file bound

obra:
Covers digestFS's two mismatch branches for a bundled-plugin file: one over maxBundledFileBytes (via a sparse, Truncate-only file so no bytes are actually allocated), and one whose read comes up short of its declared size (via a synthetic fs.FS, since a real short-read race is not deterministic). Both new tests were confirmed to fail when the corresponding check in resolve.go was temporarily removed, then pass again once restored.

### DIFF
--- a/internal/plugins/resolve_test.go
+++ b/internal/plugins/resolve_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


obra:
Roborev's finding on `24c9d10` — not changed; the claim is wrong.

`testing/fstest.MapFS` has had a `Stat` method since it was added: `go doc testing/fstest.MapFS.Stat` prints `func (fsys MapFS) Stat(name string) (fs.FileInfo, error)`. The file compiles and the test passes; CI run 33903639966 is green on this head, including the `tests`, `race-modules`, and `lint-tagged` jobs that build it.

PR #869, opened 2026-09-04
fix(plugins): validate the store root where paths are derived and where the lock is taken

obra:
## The factoring problem

`Manager.storeRootError` refuses a store root that is empty (`DefaultRoot` returns `""` when there is no `XDG_CONFIG_HOME` and no home directory) or relative, because every path built from such a root resolves against whatever directory the process happens to be in. Calling it was each caller's own job, and most callers did not.

Every writer that took the store lock — `Install`, `Upgrade`, `Remove`, `SetEnabled`, `SetAutoUpgrade`, `Gc`, `AddMarketplace`, `RemoveMarketplace`, `RefreshMarketplace`, and `Browse`, which clones a marketplace lazily — wrote into the process working directory instead. The new test caught `AddMarketplace` git-cloning a marketplace into it. `UpdateAll` and `UpdateAutoUpgrade` lock once per plugin, so they enumerate the registry before taking any lock and reported a clean sweep of a store that was not there. `List` and `ListMarketplaces` handed back whatever `installed_plugins.json` or `known_marketplaces.json` the working directory held, as the user's installed plugins and registered marketplaces. `Doctor`, which promises to mutate nothing, wrote its writability probe's temp file into the working directory under a root of `.`.

## Where the check lives now

**`Manager.storePath` (internal/plugins/paths.go) is the derivation choke point.** It joins under the root and refuses the roots `storeRootError` refuses, so a caller is turned away where the path it would have used is built. There was no single derivation helper before — each path was its own `filepath.Join(m.Root, …)` — so every access that reaches the filesystem without holding the store lock now derives through it:

- **Registry**: new `m.loadRegistry` / `m.saveRegistry`. Nine callers were spelling `installLoadRegistry(m.registryPath())` by hand; there is one way in now. `registryForSweep` no longer exists — it was only ever the check plus the load, and `loadRegistry` is that.
- **Marketplaces file**: `loadMarketplaces` / `saveMarketplaces`, already the only reach for it, plus the stat `SeedDefaultMarketplaces` does *before* it locks.
- **Bundled store**: `prepareBundledStore` derives its destination through `storePath`.
- **The root itself**: `checkStoreWritable` derives it through `storePath()` with no parts, so the probe never reaches `CreateTemp` on a root it should not touch.

**`Manager.acquireStoreLock` (internal/plugins/locks.go) is the second guard.** `storePath` cannot cover the lock file: the lock is taken before any of those paths is derived. Creating it is the first write every mutation makes, so a writer cannot forget the check without also forgetting the lock — and a second acquire site, such as the bundled cache's own `<Root>/bundled/.lock` in the sibling PR, inherits it by calling the same helper rather than `acquireLock` directly.

## The three explicit checks that remain

Outside `storePath`, `storeRootError` is called in exactly three places, each because a bare error is the wrong answer or arrives too late:

1. **`acquireStoreLock`** — the lock file is created before any store path is derived.
2. **`resolveForLaunch`** — a launch continues without plugins rather than failing, so an unusable store is a launch diagnostic, not an error.
3. **`Doctor`** — reporting the environment is what Doctor is for, so an unusable root is a `FAIL` finding under `environment` with a remediation, and the only store finding; the git check still runs alongside it, since git is on `PATH` rather than in the store.

`storePath`'s comment states the invariant that keeps the remaining plain joins safe: each is only reached once the root is known usable — under the store lock, past Doctor's own refusal (`doctorOrphanCacheDirs` walks `cacheDir` with no lock), or in a test naming a path to plant a file at.

## What the tests prove

- **`TestStoreWriters_RefuseARootThatIsNotResolved`** — a table over all eleven writers plus `Browse`, crossed with an empty and a relative root. Each subtest `t.Chdir`s into a fresh temp directory, asserts the writer returns `storeRootError`'s message, and asserts the directory is still empty: no lock file, no registry, no marketplace clone. Watched failing first; the `Browse` row was watched failing with the check bypassed at that one call site, so no acquire site is untested.
- **`TestStoreReaders_RefuseARootThatIsNotResolved`** — plants a real `installed_plugins.json` and `known_marketplaces.json` in the working directory (and in `./store`, for the relative root) and calls `List` and `ListMarketplaces`. Before the fix all four subtests returned the ambient store with a nil error.
- **`TestDoctor_ReportsARootThatIsNotResolvedWithoutWriting`** — covers `""`, `"."` and `"store"`: the git finding plus the single root `FAIL`, no error, `checkStoreWritable` refusing on its own, and — via the `doctorCreateTemp` seam, since the probe deletes its own file — that nothing was created. Reverting just the probe's derivation reproduces the report verbatim: `the writability probe created a file under "."`.
- **`TestResolveForLaunch_RefusesAnUnresolvedRootWithoutTakingTheLock`** — the read path still refuses with no lock file and an untouched working directory, which is why its own check has to stay.

## Gates

`gofmt -l internal/`, `go vet ./internal/plugins/...`, `GOOS=windows go vet ./internal/plugins/...`, `go vet -tags evenerfuzz ./internal/...`, `golangci-lint run ./internal/... ./cmd/...`, `go test -race ./internal/plugins/... -count=1`, `go test ./internal/... ./cmd/... -count=1` — all pass.

### DIFF
--- a/internal/plugins/autoupgrade.go
+++ b/internal/plugins/autoupgrade.go
@@ -57,7 +57,7 @@ func (m *Manager) upgradeAuto(ctx context.Context, plugin, marketplace string) (
 // Failures are collected but do not stop the others (failure-isolated),
 // matching UpdateAll.
 func (m *Manager) UpdateAutoUpgrade(ctx context.Context) ([]UpgradedPlugin, error) {
-	reg, err := m.registryForSweep()
+	reg, err := m.loadRegistry()
 	if err != nil {
 		return nil, err
 	}
--- a/internal/plugins/autoupgrade.go
+++ b/internal/plugins/autoupgrade.go
@@ -57,7 +57,7 @@ func (m *Manager) upgradeAuto(ctx context.Context, plugin, marketplace string) (
 // Failures are collected but do not stop the others (failure-isolated),
 // matching UpdateAll.
 func (m *Manager) UpdateAutoUpgrade(ctx context.Context) ([]UpgradedPlugin, error) {
-	reg, err := installLoadRegistry(m.registryPath())
+	reg, err := m.registryForSweep()
 	if err != nil {
 		return nil, err
 	}
--- a/internal/plugins/autoupgrade.go
+++ b/internal/plugins/autoupgrade.go
@@ -23,7 +23,7 @@ type UpgradedPlugin struct {
 // the change-detection both happen fresh, under this same lock acquisition,
 // rather than against a snapshot taken before the lock was acquired.
 func (m *Manager) upgradeAuto(ctx context.Context, plugin, marketplace string) (entry InstallEntry, changed, skipped bool, err error) {
-	release, err := installAcquireLock(ctx, m.lockPath(), 30*time.Second)
+	release, err := m.acquireStoreLock(ctx, installAcquireLock, m.lockPath(), 30*time.Second)
 	if err != nil {
 		return InstallEntry{}, false, false, err
 	}
--- a/internal/plugins/catalog.go
+++ b/internal/plugins/catalog.go
@@ -129,7 +129,7 @@ func catalogPluginName(raw json.RawMessage) string {
 // Browse returns the parsed catalog of a registered marketplace, lazily
 // cloning it first if it was only seeded as an unfetched pointer.
 func (m *Manager) Browse(ctx context.Context, name string) (Catalog, error) {
-	release, err := acquireLock(ctx, m.lockPath(), 30*time.Second)
+	release, err := m.acquireStoreLock(ctx, acquireLock, m.lockPath(), 30*time.Second)
 	if err != nil {
 		return Catalog{}, err
 	}
--- a/internal/plugins/doctor.go
+++ b/internal/plugins/doctor.go
@@ -66,7 +66,7 @@ type DoctorFinding struct {
 // returned as an error; it becomes a FAIL finding instead, so the rest of the
 // report is still useful.
 func (m *Manager) Doctor() ([]DoctorFinding, error) {
-	reg, err := LoadRegistry(m.registryPath())
+	reg, err := m.loadRegistry()
 	if err != nil {
 		return nil, err
 	}
--- a/internal/plugins/doctor.go
+++ b/internal/plugins/doctor.go
@@ -64,8 +64,24 @@ type DoctorFinding struct {
 // or known_marketplaces.json) failed to parse — the same failure every other
 // Manager verb would hit. A per-plugin or per-marketplace problem is never
 // returned as an error; it becomes a FAIL finding instead, so the rest of the
-// report is still useful.
+// report is still useful. A store root that cannot be used is a finding too,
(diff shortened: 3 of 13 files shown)


obra:
Both roborev findings are fixed, in `a8c9a09d6` and `9122df08b`.

**The choke point for reads: `Manager.storePath` (internal/plugins/paths.go).** The store lock is the writers' choke point and does nothing for a reader that never takes one, which is exactly how `List` and `ListMarketplaces` slipped through. Path derivation is what reads and writes share, so `storePath(parts...)` joins under the root and refuses the roots `storeRootError` refuses — an empty root or a relative one — and every store access that does not hold the lock derives through it.

There was no single derivation helper before; each path was its own `filepath.Join(m.Root, …)`. Rather than give all seven of them an error return (roughly 80 call sites, most of them under the lock where the error cannot fire), the access that reaches the filesystem was routed through it:

- **Registry**: new `m.loadRegistry` / `m.saveRegistry`. Every caller was spelling `installLoadRegistry(m.registryPath())` by hand — nine of them — and now there is one way in. `registryForSweep` is gone; it was only ever this.
- **Marketplaces file**: `loadMarketplaces` / `saveMarketplaces`, which were already the only reach for it, plus the stat `SeedDefaultMarketplaces` does *before* it locks.
- **Bundled dir**: `prepareBundledStore` derives its destination through `storePath`, which retires its explicit check.
- **Store root itself**: `checkStoreWritable` derives it through `storePath()` with no parts.

`acquireStoreLock` keeps its check, as asked: it guards the lock file, which no other derivation covers. The launch resolver keeps its check too — it refuses before anything derives a path, and its diagnostic is how a launch reports the store as unavailable. Those are now the only two explicit `storeRootError` calls outside `storePath`.

**Doctor** reports an unusable root the way it reports the rest of the environment: a `FAIL` finding under `environment` with a remediation, not an error. It is the only finding returned, since there is nothing under such a root to report on. The probe refuses on its own too, so a caller reaching `checkStoreWritable` directly cannot write either.

**Tests, watched failing first.** `TestStoreReaders_RefuseARootThatIsNotResolved` plants a real `installed_plugins.json` and `known_marketplaces.json` in the working directory (and in `./store`, for the relative-root case) and calls `List` and `ListMarketplaces` under an empty and a relative root: before the fix all four subtests returned the ambient store with a nil error. `TestDoctor_ReportsARootThatIsNotResolvedWithoutWriting` covers `""`, `"."` and `"store"`, asserting the single FAIL finding, that `checkStoreWritable` refuses, and — via the `doctorCreateTemp` seam, since the probe deletes its own file — that nothing was created. Reverting just the probe's derivation reproduces roborev's case verbatim: `the writability probe created a file under "."`.

Gates all pass: `gofmt -l internal/`, `go vet ./internal/plugins/...`, `GOOS=windows go vet ./internal/plugins/...`, `go vet -tags evenerfuzz ./internal/...`, `golangci-lint run ./internal/... ./cmd/...`, `go test -race ./internal/plugins/... -count=1`, `go test ./internal/... ./cmd/... -count=1`.

PR #842, opened 2026-09-02
feat(plugins): bundled plugins are launch candidates when named

obra:
`evener run --enabled-plugins coordinator-workflow` failed with `enabled plugin selection is unavailable: coordinator-workflow: no valid plugin candidate`. `ResolveForLaunch` only enumerated explicit `--plugin-dir` directories and installed marketplace plugins; the plugins embedded in the binary (`internal/bundled/plugins/*`) were never candidates.

A requested name that matches a bundled plugin is now materialized to `<store>/bundled/<name>` (replaced on every launch so the on-disk plugin always matches the running binary) and joins the inventory with a new source, `bundled`. Unrequested launches see no bundled candidates, so defaults are unchanged.

Test: `TestResolveForLaunch_BundledPluginByName` (selected by name, manifest present, roster loaded; absent when not requested).

Found while trying to run a Terminal-Bench probe with the bundled coordinator persona (`--agent coordinator --enabled-plugins coordinator-workflow`).

## Preview and the store

Preview readies the store the way a launch does: it creates `<store>/bundled` if missing, stages a copy behind an ownership marker, loads it, and removes the staging directory, so an unwritable store fails preview and launch identically. It publishes nothing, reclaims nothing, and never moves a conflicting destination; it reports the conflict a launch would set aside and leaves the directory where it is.

## Follow-ups

- Only launch resolution, bundled materialization, and marketplace seeding refuse an unresolved or relative store root (`storeRootError`). `Install`, `Upgrade`, `UpdateAll`, `UpdateAutoUpgrade`, `Remove`, `SetEnabled`, `SetAutoUpgrade`, `Gc`, `AddMarketplace`, `RemoveMarketplace`, and `RefreshMarketplace` all write under `m.Root` without it; the one-line fix is to validate the root where the store lock is acquired.
- `agent/session_lifecycle.go:410` `discardRestoredCandidate` disposes only the sandboxed scratch and never calls `env.Cleanup`, so a discarded restored candidate on the default unsandboxed env leaks its scratch and holds the flock lease for the process lifetime; `DisposeUnadoptedScratch` applies there too.
- `agent/execenv/local.go:445-454` `unsandboxedScratchDir` holds `unsandboxedScratchMu` across a git probe that can re-enter `unsandboxedScratchDir` when `structuralWorktreeRoot` misses but a `.git` ancestor exists: a self-deadlock, narrow but a hang.
[...]

PR #870, opened 2026-09-04
fix: dispose unadopted scratch on discard, failed clear, and provisioning failure

obra:
Three neighbours of the scratch leaks #842 fixed. Every session environment can hold up to two scratch directories — the sandbox-owned one `EnableSandbox` provisions and the one an unsandboxed environment mints lazily on its first command — each with an flock lease under it, and only a session's own teardown ever releases either.

**A discarded restore candidate leaked its unsandboxed scratch.** `discardRestoredCandidate` disposed only the sandbox-owned scratch, so a candidate discarded on the default (unsandboxed) environment left the directory it minted on its first command, and the lease under it, for the life of the process — nothing else was ever going to release them, since the candidate was never adopted. It now disposes both via `DisposeUnadoptedScratch`. `TestDiscardRestoredCandidateDisposesUnsandboxedScratch` (agent/sandbox_delegate_test.go) runs a command on the candidate's environment, discards it, and requires the scratch directory to be gone; it fails on the old code with the directory still present. The disposal is guarded on ownership: `prepareSubagentEnvironment` hands back the parent's environment untouched when a delegate needs neither a working-dir re-root nor a box of its own, so `discardRestoredCandidate` now takes the same `ownsEnv` answer `close()` reads from `subagent.ownsEnv` and leaves a shared environment's scratch to the live parent still working in it (`TestDiscardRestoredCandidateLeavesASharedEnvironmentAlone`, which fails on the unguarded version with the parent's scratch deleted).

**A failed `/clear` leaked one scratch per attempt.** `/clear` provisions a fresh environment before it touches anything shared; when the session that would have owned it could not be built, `clearEnv.Cleanup()` RETAINED the scratch — the handoff convention for a session someone may still want to inspect — so each failed clear left a directory and a live lease behind for the daemon's whole uptime. It now disposes it after Cleanup. `TestRunServeClearSessionFailureDisposesUnadoptedScratch` (cmd/evener/serve_state_test.go) drives a real serve through the daemon's own clear callback with a session constructor that mints the scratch (as `NewSession` does on its first command) and then fails, and requires the directory to be gone; it fails on the old code with the directory still present.

**The provisioning-failure return needs no disposal — provably.** run.go and serve.go return from a failed `provisionSandbox` without disposing anything. Nothing can be there: no command runs through the environment between its construction and provisioning, so the lazily minted unsandboxed scratch does not exist yet, and `EnableSandbox` disposes the tmp it minted on every one of its failure paths — including the refusal that fires only AFTER the mint (no backend binary for a mode that resolved). No production change; `TestLaunchProvisioningFailureLeavesNoScratch` (cmd/evener/sandbox_test.go) pins it for both launches by driving each through that exact post-mint refusal and asserting the environment holds no scratch, so a step added before provisioning that DOES spawn a command fails there instead of leaking a directory and a lease per failed launch.

Review follow-up, three siblings of the same leak. A delegate restore that fails inside `initSessionState` (`restoreIdle`'s rollback defer) disposed only the sandbox-owned scratch, though the construction has already run the child's git snapshot and so minted the unsandboxed one — `TestRestoreIdleFailureDisposesTheChildScratch` drives a real restore into a post-snapshot fault and finds the leaked directory under a redirected `TMPDIR`. The create path had the identical shape when `NewSession` fails after its own snapshot, fixed the same way and covered by `TestSpawnedSubagentSessionFailureDisposesTheChildScratch`. And `disposeUnadoptedSubagentSession` kept the unsandboxed directory (Close releases its lease but keeps the path, the handoff convention for a session someone may inspect) while the discard path's comment claimed to mirror its discipline; both now dispose both dirs for an owned environment and leave a shared one to the live parent, which `TestDisposeUnadoptedSubagentSessionDisposesEveryScratchItOwns` pins in both directions.

Gates: gofmt, `go vet` (plain and `-tags evenerfuzz`), golangci-lint, `go test ./agent/execenv/...`, `./agent/`, and `./cmd/evener/` (targeted and full).

## The consequential change

An unadopted child session never runs its environment's `Cleanup` any more, whether it owned the environment or not. A child environment cloned with a new working directory shares its parent's running-process table, so cleaning it up sent SIGTERM to the parent's in-flight commands; a test reproduces a parent's `sleep` dying after a child cleanup before the fix. The child now closes with `cleanupEnv=false` and disposes only the scratch it owns. The remaining ownership hazards outside this PR's scope are tracked in #877.

### DIFF
--- a/agent/delegate_resource_create_test.go
+++ b/agent/delegate_resource_create_test.go
@@ -1024,7 +1024,7 @@ func TestDelegateResourceCreate_MissingRestoreInputsCloseResumabilityBeforeClean
 	// This boundary models process loss, not a graceful shutdown. A graceful
 	// Close now durably stops committed generations and therefore waits for the
 	// construction owner that a crash fixture intentionally does not retain.
-	root.discardRestoredCandidate()
+	root.discardRestoredCandidate(true)
 
 	restored, err := restoreDelegateResourceBootstrapSession(client, profile, workspace, meta, stateDir)
 	if err != nil {
--- a/agent/delegate_resource_runtime_test.go
+++ b/agent/delegate_resource_runtime_test.go
@@ -109,7 +109,7 @@ func TestRestoredDelegatePostStartPopulationEmitsTaskCorrection(t *testing.T) {
 		t.Fatal("restoreIdle unexpectedly reused a resident child")
 	}
 	defer func() {
-		sub.sess.discardRestoredCandidate()
+		sub.sess.discardRestoredCandidate(sub.ownsEnv)
 		_, _ = root.delegateController.FailCommittedRestart(started.lease, delegatePermanentStartFailure(context.Canceled, "test_cleanup"))
 	}()
 
@@ -2250,7 +2250,7 @@ func TestDelegateResourceRuntime_ColdIdleUsesCommittedConfigTemplatesAndToolCeil
 	if !restored {
 		t.Fatal("cold idle delegate was reported retained")
 	}
-	defer sub.sess.discardRestoredCandidate()
+	defer sub.sess.discardRestoredCandidate(sub.ownsEnv)
 	if sub.sess.cfg.MaxToolRoundsPerInput != 17 || sub.sess.cfg.ReasoningEffort != "high" {
 		t.Fatalf("restored config = rounds:%d effort:%q", sub.sess.cfg.MaxToolRoundsPerInput, sub.sess.cfg.ReasoningEffort)
 	}
@@ -2293,7 +2293,7 @@ func TestDelegateResourceRuntime_ColdIdleInheritsLiveLifetimeContext(t *testing.
 	if !restored {
 		t.Fatal("cold idle delegate was reported retained")
 	}
-	defer sub.sess.discardRestoredCandidate()
+	defer sub.sess.discardRestoredCandidate(sub.ownsEnv)
 
 	cancelOwner()
 	select {
@@ -2326,7 +2326,7 @@ func TestDelegateResourceRuntime_ColdIdleReusesExactSharedRootTaskStore(t *testi
 	if err != nil {
 		t.Fatalf("restoreIdle: %v", err)
 	}
-	defer sub.sess.discardRestoredCandidate()
+	defer sub.sess.discardRestoredCandidate(sub.ownsEnv)
 	if got := sub.sess.getOrCreateTaskStore(); got != want {
 		t.Fatalf("shared TaskStore pointer = %p, want exact root pointer %p", got, want)
 	}
--- a/agent/delegate_resource_worktree_test.go
+++ b/agent/delegate_resource_worktree_test.go
@@ -416,7 +416,7 @@ func TestStableDelegateWorktree_SandboxRestoreUsesDescriptorNotLegacyJob(t *test
 		t.Fatal("stable descriptor restore unexpectedly reused a resident child")
 	}
 	defer func() {
-		sub.sess.discardRestoredCandidate()
+		sub.sess.discardRestoredCandidate(sub.ownsEnv)
 		_, _ = root.delegateController.FailCommittedRestart(started.lease, delegatePermanentStartFailure(context.Canceled, "test_cleanup"))
 	}()
 	if got := sub.sess.currentEnv().WorkingDirectory(); got != descriptorDir {
--- a/agent/delegate_runtime.go
+++ b/agent/delegate_runtime.go
@@ -1486,13 +1486,11 @@ func delegateSandboxFallbackHint(s *Session, args delegateArgs, err error) error
 
 func (isolation delegateIsolation) cleanup(s *Session, delegateID string) {
 	if isolation.ownsFreshEnv {
-		if local, ok := isolation.env.(*execenv.LocalExecutionEnvironment); ok {
-			// Both scratch dirs go. createSubagent leaves a PREPARED environment
-			// alone (it belongs to this isolation step), so this is the only
-			// rollback for the scratch the construction's git snapshot minted on
-			// an unsandboxed lane, as well as for a sandboxed lane's owned one.
-			local.DisposeUnadoptedScratch()
-		}
+		// prepareSubagentRunFromSelection leaves a PREPARED environment alone (it
+		// belongs to this isolation step), so this is the only rollback for the
+		// scratch the construction's git snapshot minted on an unsandboxed lane,
+		// as well as for a sandboxed lane's owned one.
+		disposeUnadoptedScratch(isolation.env)
 	}
 	if isolation.worktreePath != "" {
 		s.rollbackFreshDelegateWorktree(delegateID, isolation.worktreePath, isolation.worktreeProject)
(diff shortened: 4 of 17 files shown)


obra:
All three findings held; each is fixed with a test that failed first.

**High, `disposeUnadoptedSubagentSession` cleaning up a non-owned environment (284548f9d).** Not test-only: a delegate spawned with no working dir and no box of its own shares the parent's environment (`prepareSubagentEnvironment` returns it untouched, `ownsFresh=false`), and every post-construction rollback on that path — the retained-delegate cap, the tree-capacity cap, the tool-grant check — went through `Close()`, whose `env.Cleanup()` released the LIVE parent's scratch leases and signalled the processes the parent tracks. It now closes with `sess.close(context.Background(), ownsEnv)`, the same env-skipping call the parent's own teardown makes for its children, and disposes scratch only when owned. `TestDisposeUnadoptedSubagentSessionDisposesEveryScratchItOwns` asserts both directions — closed either way, Cleanup zero times on a shared environment, both scratch dirs gone on an owned one. `TestRetention_FailLoudAtCap` had encoded the old behaviour (it proved the child was closed by counting Cleanup on the shared environment); it now asserts that count is zero, and the closure it stood for is asserted directly on a held reference.

**Medium, `delegateIsolation.cleanup` (6ff2f5ee7).** Confirmed: stable construction passes a prepared environment, so `createSubagent`'s own rollback deliberately leaves it alone and this is the only thing that rolls it back — while it disposed just the sandbox-owned scratch. Switched to `DisposeUnadoptedScratch`, covered by `TestDelegateIsolationCleanupDisposesEveryScratchItOwns`.

**Medium, worktree re-entry replacing `s.env` (91e07adc4).** Confirmed. `RestoreSessionFromMetaWithConfig` now disposes the environment it re-rooted onto itself when it fails after re-entry, guarded so it never touches the caller's own env. Re-entry also leaked two more: the throwaway clones it builds to resolve the project and to run `git worktree list` each run a command, so each minted a scratch and took its lease with nobody holding a reference afterwards — on a successful re-entry as much as a refused one. Both are now disposed at the end of the call. `TestWorktreeReentryRestoreFailureDisposesTheReenteredScratch` drives a real re-entry into a post-snapshot failure and requires an empty scratch base; it found all three.

Gates: gofmt, `go vet ./agent/`, `go vet -tags evenerfuzz ./agent/...`, golangci-lint (0 issues), `go test ./agent/ -run 'Restore|Discard|Scratch|Sandbox|Subagent|Delegate|Isolation'`, `go test -race ./agent/ -run 'Restore|Discard|Subagent'`, and the full `go test ./agent/ -count=1 -timeout 30m`.

obra:
Thanks — verified all four against `3e85833e8`, and they hold. Filed as #877 (with a fifth found while checking them). Summary of where each one stands:

- **`agent/subagents.go:178`** — the shared-environment half is what this PR fixes: `main` had an unconditional `sess.Close()`, head has `sess.close(ctx, ownsEnv)`, so a child on the parent's environment no longer reaches `Cleanup()`. The residual is pre-existing and real: an *owned* child environment is still a `WithWorkingDirectory` clone (`agent/sandbox_delegate.go:134,148`) sharing `runningPIDs` (`agent/execenv/local.go:687`), so `Cleanup()` there still SIGTERM/SIGKILLs the parent's tracked processes (`agent/execenv/local.go:836-863`). Measured: a clone's `Cleanup()` killed the parent's in-flight `sleep 5` in 6 ms.
- **`agent/subagents.go:1019` (eviction)** — pre-existing, identical at `agent/subagents.go:1008` on `main`. `ev.sess.Close()` has no `ownsEnv` check at all, so it carries both the shared-PID kill and, on a shared environment, the parent's scratch-lease release.
- **`agent/session_lifecycle.go:274-276`** — pre-existing, identical on `main`. `RetainSandboxScratch()` covers only `ownedSessionTmp` (`agent/execenv/local.go:597-612`); the unsandboxed lazily-minted dir's counterpart `retainUnsandboxedScratch` is reachable only from `Cleanup()` (`agent/execenv/local.go:490-496,826`), which this path skips by design. Unsandboxed is the default shape, so that lease is held for the daemon's uptime.
- **`agent/session_init.go:987-994`** — the ownership split is pre-existing: `provisionRestoredSandbox` mints `ownedSessionTmp` on the original env (`agent/session_init.go:751`), then re-entry swaps in a clone (`agent/session_worktree_resume.go:161`) that deliberately does not copy it. The defer this PR adds covers only `!restoreComplete`; the successful-restore leak is untouched and belongs to the ownership-transfer fix.

Plus one more the same shape, also pre-existing (byte-identical at `agent/delegate_runtime.go:1831` on `main`): the `stopWon` early return at `agent/delegate_runtime.go:1839` skips `isolation.cleanup` entirely, so a construct failure racing a concurrent stop (`prepared == nil`, `agent/delegate_runtime.go:1234`) leaves the prepared environment's scratch, its lease, and the isolation lane behind.

All five want the same fix — one ownership-aware teardown shared by close, eviction, discard, and parent teardown, plus scratch ownership transferred on re-entry — which is a refactor across paths this PR does not otherwise touch. So this PR stays scoped to the failure-path disposal leaks, and #877 tracks the ownership work.

obra:
Round three, all four fixed.

**1 — an unadopted child never runs `Cleanup` at all (09b9d61b9).** Confirmed and reproduced: `WithWorkingDirectory` copies `runningPIDs` by pointer, so a child's OWN clone shares the parent's process table and cleaning it up signalled the parent's in-flight tools. `TestDisposeUnadoptedSubagentSessionLeavesTheParentsProcessesAlone` starts a real `sleep` through the parent environment, disposes an unadopted owned child, and fails if that process ended — it did, in about two seconds (the termination grace), before the fix. The close is now `sess.close(ctx, false)` unconditionally, followed by scratch disposal only for an owned environment; the second half of the test then requires the parent's own `Cleanup` to still end the process, which is how it shows the process never left the map that tracks it.

**2 — one helper for the idiom (3a5e3c6e1).** `disposeUnadoptedScratch(env)` states the rule once and is used at all six sites (`subagents.go` x2, `session_lifecycle.go`, `delegate_runtime.go` x2, `session_init.go`). No behavior change. The two direct calls in `session_worktree_resume.go` are on concrete `*LocalExecutionEnvironment` values and never used the guarded idiom, so they stayed as they are.

**3 — the cap rollback's close is pinned again (05c611d5d).** Confirmed the gap: with the environment-Cleanup assertion gone, deleting `disposeUnadopted()` at the retained-delegate cap passed the whole package. A session releases its API-log route exactly once, when it closes, so a recording client middleware proves the close without holding a reference to the child and without touching the environment. Verified against that exact mutant: releases = 0, want 1.

**4 — comments (fd4b89482).** Both now name `prepareSubagentRunFromSelection`, which is where the spawn rollback actually lives.

One extra commit, 938e3cf4f, belongs to item 1: the new test's two `time.After` bounds tripped the package's deadline audit. The negative check is now a non-blocking read (a `Cleanup` that reached the process would have signalled, waited out the grace, and killed it before the dispose call returned, so the exit is already visible if it happened) and the positive one awaits the real exit. Re-verified against the reverted fix: still fails, same way.

Gates: gofmt, `go vet ./agent/`, `go vet -tags evenerfuzz ./agent/...`, golangci-lint (0 issues), `go test ./agent/ -run 'Restore|Discard|Scratch|Sandbox|Subagent|Delegate|Isolation|Retention'`, `go test -race ./agent/ -run 'Restore|Discard|Subagent|Retention'`, and the full `go test ./agent/ -count=1 -timeout 30m`.

obra:
Roborev's finding on `938e3cf` — not changed here. The re-entry ownership transfer (the clone never inherits the original environment's sandbox scratch, so a successful restore leaves the original's lease) is pre-existing on main and is item 4 of #877, which this PR deliberately leaves for a dedicated ownership-aware teardown. This PR's re-entry change disposes the clone actually in use on failure and the two probe clones on every path; it does not attempt the transfer.

PR #872, opened 2026-09-04
fix(plugins): give the bundled cache its own lock

obra:
## The collision

Publishing a bundled plugin into the content-addressed cache at
`<Root>/bundled/<name>-<digest>` took the plugin store lock, `<Root>/.lock` —
the same lock install, upgrade, gc, catalog and marketplace refresh hold across
git fetches for up to 30 seconds. Hub start runs the auto-upgrade sweep, so a
first launch after a version bump (the one launch that actually has a copy to
publish) waited that out on the caller's context and then failed with `another
evener plugin operation is in progress`.

## The decision

The bundled cache gets its own lock at `<Root>/bundled/.lock`. Publishing a
bundled copy needs exclusion only against other bundled publishers — the
classify, set-aside, stage and rename sequence touches nothing outside
`<Root>/bundled` — and never against a git fetch.

Every bundled-cache mutation now takes that lock instead: the publish sequence,
the adopt path's sweep, and the set-aside and restore that run under the lock
the staging carries. Same acquire helper, same 30s publish budget and 1s
housekeeping budget for the sweep. Nothing in the bundled path takes
`<Root>/.lock` any more; the seeding and marketplace paths keep the store lock.
Acquisition goes through one function, `Manager.acquireBundledLock`, so the
store-root check that `Manager.acquireStoreLock` centralizes in #869 has a
single place to land for this lock too.

The ownership marker, the sweep and the `.conflict` slot are unaffected: they
all live under `<Root>/bundled`, and the sweep only ever considers directories
wearing the staging prefix, so the lock file beside them is never a candidate.

## What the tests prove

- `TestMaterializeBundledPlugin_PublishesWhileTheStoreLockIsHeld` — a launch
  that must publish while `<Root>/.lock` is held publishes promptly instead of
  waiting out the 30s budget and failing. It reproduced the collision before
  the fix: 10s to the deadline with `waiting for plugin lock <Root>/.lock`,
  0.3s after.
- `TestMaterializeBundledPlugin_LeavesTheStoreLockFreeWhilePublishing` — the
  other direction: a publish held mid-copy leaves the store lock free for the
  install or auto-upgrade that wants it.
- `TestMaterializeBundledPlugin_ConcurrentCallsPublishOnce` and
  `TestBundledStore_ConcurrentPublishKeepsOneConflictAndOneCopy` stay green:
  bundled publishers still serialize, one published copy and one set-aside slot.
- The tests that planted a lock explicitly to exercise the bundled path — the
  cancellation wait, the sweep-only-under-the-lock pair, the short sweep
  budget, the adopt path — are retargeted to the new lock.
- Store listings in tests ignore the cache's own lock file, which is now a
  legitimate entry in `<Root>/bundled`.

Docs updated where they said bundled publication takes the store lock:
`PreviewForLaunch`'s account of what it touches, `appwire.PluginPreviewParams`
(which owes callers the on-disk residue), the hub controller's argument for
holding no mutex of its own, and the TUI's reason for giving preview the slow
timeout.

Gates: `gofmt -l internal/`, `go vet ./internal/plugins/...`, `GOOS=windows go
vet ./internal/plugins/...`, `go vet -tags evenerfuzz ./internal/...`,
`golangci-lint run ./internal/... ./cmd/...`, `go test -race
./internal/plugins/... -count=1`, `go test ./internal/... ./cmd/... -count=1` —
all clean.

### DIFF
--- a/appwire/types.go
+++ b/appwire/types.go
@@ -2827,13 +2827,14 @@ type PluginCheckNowResponse struct {
 
 // PluginPreviewParams requests the launch plugin inventory for a working
 // directory and optional per-launch overrides. Preview starts no session and
-// runs no plugin code; for a requested bundled plugin it readies the same store
-// a launch publishes into, staging and removing a marked copy, so it fails
-// wherever the launch it describes would. Readying that store creates the
-// bundled directory under the plugin root when it is missing, and the
-// directory stays behind once the staged copy is removed; a destination
-// holding content the running binary did not publish is reported as the
-// conflict a launch would set aside, and left exactly where it is.
+// runs no plugin code; for a requested bundled plugin it readies the same
+// store a launch publishes into, staging and removing a marked copy, so it
+// fails wherever the launch it describes would. Readying that store creates
+// the bundled directory under the plugin root when it is missing, and the
+// directory stays behind — with the lock file the cache keeps in it — once the
+// staged copy is removed; a destination holding content the running binary did
+// not publish is reported as the conflict a launch would set aside, and left
+// exactly where it is.
 type PluginPreviewParams struct {
 	CWD             string             `json:"cwd"`
 	LaunchOverrides *LaunchConfigLayer `json:"launchOverrides,omitempty"`
--- a/cmd/evener-hub/app_plugin_preview_test.go
+++ b/cmd/evener-hub/app_plugin_preview_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"slices"
 	"testing"
 
 	"primeradiant.com/evener/appwire"
@@ -370,6 +371,9 @@ func TestPluginPreviewBundledPluginPublishesNothing(t *testing.T) {
 	if err != nil {
 		t.Fatalf("read the bundled store after a preview: %v", err)
 	}
+	// The bundled cache keeps its own lock file, which readying the store
+	// opens; it is not something a preview published.
+	entries = slices.DeleteFunc(entries, func(e os.DirEntry) bool { return e.Name() == ".lock" })
 	if len(entries) != 0 {
 		t.Fatalf("preview left %d entries in the plugin store: %v", len(entries), entries)
 	}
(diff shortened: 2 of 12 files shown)


obra:
Both roborev findings, one addressed and one declined.

**Low, `resolve_unix_test.go:722` — wall-clock assertion. Fixed** in `44e751e`.

Neither new test needs elapsed time. `PublishesWhileTheStoreLockIsHeld` now holds `<Root>/.lock` for the whole test (the holder releases in cleanup and nowhere else) and, once the publish has succeeded, checks that the lock is *still* contended with a zero-budget acquire. The evidence is "published while the store lock was held", not "published quickly". The remaining deadline is a labeled hang tripwire at one minute, well past the 30s publish budget, so a publish that queues behind the store lock reports that lock by name rather than dying on a deadline. Its mirror, `LeavesTheStoreLockFreeWhilePublishing`, asks for the store lock with no wait budget at all while a publish is pinned mid-copy: free on the first try, or the publish is holding it.

Verified both still fail when `acquireBundledLock` is pointed back at `m.lockPath()` — 30s then `another evener plugin operation is in progress (locked: <Root>/.lock)` for the first, immediate for the second.

**Medium, mixed-version processes locking different files. Not changing.**

evener does not carry compatibility windows across lock protocols; backward compatibility here is a decision for Jesse, not a default, and nobody has asked for one. A restart boundary is the migration.

The exposure is also small enough to be worth stating precisely. The bundled cache landed yesterday in #842, so the population of old binaries is one day of builds. The destination is content-addressed, so two binaries whose bundled plugin content differs publish to different paths and never meet; they collide only when the content is identical, in which case one adopts the other's copy.

The sweep is the part that would matter and it is already protected independently of any lock: `abandonedStaging` only considers a directory wearing the staging prefix, holding the marker as a regular file this code wrote, *and* older than one hour. An older publisher's in-flight staging is seconds old, so a newer binary's sweep will not touch it.

What is left is two publishers racing `setAsideBundledConflict` on the same already-mismatched destination during the moments both binaries are running. That path is rename-based and self-recovering — an interrupted set-aside is put back before anything else, and a publish that frees a name and then fails restores the occupant — and the same interleaving is already reachable from a foreign writer, which the code is written to survive. Narrowing it further would mean holding both locks, which reintroduces exactly the 30s stall this PR exists to remove.

PR #871, opened 2026-09-04
feat(web): keybindings Phase 4a - final map, cheatsheet overlay, hold-modifier hints

obra:
Phase 4a of the WebUI keybinding system (survey: #853, dispatcher: #860, persistence: #861, navigation: #863). Stacked on #863's branch.

## What ships

**Final binding map** (task 1):
- `⌘K` / `⌘I` / `⌘J` are now strict chords — the legacy optional-Shift/Alt permissiveness is gone, so **`⌘⌥I`, `Ctrl+Shift+J`, and `⌘⇧J` revert to the browser** (DevTools/downloads). Meta-or-Ctrl either/both permissiveness is kept. Deliberate, user-approved behavior change; the 2a parity test now pins the strict contract.
- New defaults: `⌘,` opens Settings; `Alt+Home` / `Alt+End` jump the focused transcript to top/bottom (editable-suppressed).

**Cheatsheet overlay** (task 2): `⌘/` (Ctrl+/ elsewhere) opens a grouped overlay listing every action's *effective* chord, live from the registry so overrides show truthfully. Secondary trigger `?` outside editable fields, behind a new character-key-triggers toggle (default ON; WCAG 2.1.4 turn-off; persisted in the frontend prefs store). Escape / `⌘/` close. Mobile: inert.

**Hold-modifier hints** (task 3): hold `⌘` (macOS) / `Ctrl` ~400ms to reveal hint chips over bound affordances, showing effective chords. Observer-only listeners (no preventDefault); hides on release, blur, visibilitychange, any key, or a hard timeout; honors prefers-reduced-motion; desktop-only.

**Docs** (task 4): `docs/web-ui/keybindings.md` finalized — full map table, strict-chord callout, overlay/hints sections; provisional-status note removed.

## Verification

- 4564 frontend tests green (271 files), tsc clean, biome clean.
- `make test-web` PASS; `make test-web-browser` PASS (run with a TMPDIR override for a sandbox Chrome socket-path-length SIGTRAP — environmental, no repo impact).
- SDD per-task reviews + final whole-branch review: **Ready**, 0 Critical / 0 Important; 12 minors triaged as fine-as-deferred (notables: settings section shows only ⌘/ for the cheatsheet row; hint-chip React duplicate-key edge when two chips show identical overridden chords).

## Notes for reviewers

- The `?` binding carries optional Shift because every common layout types `?` with Shift held; tinykeys rejects unlisted modifiers. Documented in defaults.ts.
[...]

PR #863, opened 2026-09-04
feat(web): navigation key bindings — session cycling, transcript scroll, AskDock keyboard completion (Phase 3)

obra:
## Summary

Phase 3 of the webui keybinding project (survey #853, dispatcher #860, persistence #861 — **stacked on #861's branch**; retargets as the stack merges). The first NEW bindings on the dispatcher infrastructure, driving existing mechanisms (no parallel navigation system):

- **Session cycling** — `Alt+←/→` cycles open session panes in workspace order (wrapping; no-op with <2; lands on first/last when focus is outside sessions) via `workspaceStore.focusPane`. `Alt+←` is claimed even on no-op to guard the browser Back shortcut.
- **Transcript scroll** — `Alt+↑/↓` line scroll, `Alt+Shift+↑/↓` page scroll on the focused session pane's transcript, via the transcript stack's own scroll API (virtualizer scrollToIndex / scrollTop / jumpToBottom). Per-pane multi-instance handlers with focused-pane arbitration; dormant panes decline. `scrollTop`/`scrollBottom` actions exist chord-less (Phase 4 assigns).
- **AskDock keyboard completion** — batch jump (`Alt+PageUp/Down`, wrapping, landing on the selected tab), focus return to the composer after send, and the Escape no-op **pinned as deliberate** per `docs/web-ui/parity/parity-m5-composer.md:120` (the dock is the one canonical response surface). A reachability audit confirmed no mouse-only control remains in the rendered set.

All new chords are registry defaults — remappable via the 2b persistence from day one — and provisional: Phase 4 owns the comprehensive map. Chords are inert on mobile.

Dev docs: `docs/web-ui/keybindings.md` gains the full binding-map table (generated from `defaults.ts`), the Phase 3 mechanisms, and the provisional-status note.

## Test plan

- New suites: sessionCycle (order/wrap/no-op/mobile), useTranscriptScrollKeys (focused-pane arbitration incl. two-pane, editable suppression), AskDock (batch jump, focus return, Escape no-op pin). Focused final gate: 57/57 green.
- `make test-web` green at the head; existing suites pass unmodified.
- Per-task reviews + final whole-branch review: Approved, 0 Critical/Important.

🤖 Generated via subagent-driven development with per-task and whole-branch review.

PR #861, opened 2026-09-04
feat(hub,web): server-persisted keybinding overrides (Phase 2b)

obra:
## Summary

Phase 2b of the webui keybinding project (survey #853, dispatcher #860 — **stacked on #860's branch**, retargets to main after it merges): user keybinding overrides persisted on the hub, cloned from the proven transcriptDisplay settings pattern.

- **Server (Go):** `evener/settings/keybindings/{get,patch,changed}` wire methods + catalog + `keybindingsSettings` feature flag; `hubcore.KeybindingsStore` (`<stateroot>/keybindings/state.json`, revisioned optimistic concurrency with CodeConflict, temp+rename+fsync durability, load-failure blocks patches without destroying evidence); revision-gated `changed` broadcast; regenerated `types.gen.ts` + `docs/appwire-protocol.md`.
- **Payload contract:** `{version: 1, revision, rules: [{action, chord | null}]}` — chord rebinds, null unbinds, last rule wins.
- **Validation split:** server validates structure only (strict decode, unknown-field rejection); frontend validates semantics (live action registry, per-platform reserved chords, **overlap-aware** conflict detection via `chordsOverlap`).
- **Frontend:** `stores/keybindings.ts` (feature-gated, startup apply, `changed`-reconcile with revision-after-success ordering and notification containment, CodeConflict refresh), `keybindings/overrides.ts` (atomic rebind/restore), read-only Settings section listing effective bindings live from the registry.
- **Docs:** `docs/web-ui/keybindings.md` gains the persistence architecture, payload contract, validation split, and failure posture.

Hub-shared scoping is deliberate: the hub has a single shared auth token and no per-user identity, so overrides sync across the hub's browsers (matches the approved design).

## Test plan

- Server: store/durability/fault-injection suites (t.TempDir, afero), two-client broadcast RPC tests, strict-decode rejection matrix; `go test ./appwire/... ./cmd/evener-hub/...` green, `make vet` clean, `go generate ./appwire` no drift.
- Frontend: 766 tests green incl. FakeClient store tests (startup apply, reconcile-removal, unsupported-server fallback, conflict retry); `make test-web` green.
- Review loop: per-task reviews + final whole-branch review; a reconcile-wedge Critical and an optional-modifier conflict-shadowing Important were caught, fixed, and re-reviewed.

🤖 Generated via subagent-driven development with per-task and whole-branch review.

PR #860, opened 2026-09-03
feat(web): keybinding dispatcher core + migration of global chords (Phase 2a)

obra:
## Summary

Phase 2a of the webui keybinding project (survey: #853): the keybinding **infrastructure**, landed with **no user-visible behavior change**.

- New `src/keybindings/` module (React-free): chord parser/serializer built on **tinykeys 4** (new dependency, ~1 KB gz), zustand **registry** with scope stack + conflict detection, single window-level **dispatcher** with per-binding policy flags (`allowInEditable`, `allowInModal`, `ignoreIfDefaultPrevented`), and the default six-chord map.
- All six global chords migrated off their per-component listeners onto the dispatcher: ⌘K (palette), ⌘B (rail), ⌘I (composer focus), ⌘J (needs-you cycle), ⌘' (selection quote), Settings Escape (now scope-gated). Every legacy guard is preserved exactly — including ⌘K's modal exemption, ⌘B's missing `defaultPrevented` check, the AltGr guard on ⌘', extra-modifier permissiveness on K/I/J and Settings Escape, and multi-pane SelectionQuote semantics (multi-handler action registration).
- `src/shell/installKeybindings.ts` — the single (idempotent) wiring point where the module meets the app's stores.
- Dev doc: `docs/web-ui/keybindings.md` — architecture, precedence policy, how to register actions/bindings, testing conventions, and what is deliberately deferred (persistence → Phase 2b, overlay/hints → Phase 4).

Not here yet: server-persisted user remapping (Phase 2b), new navigation bindings (Phase 3), cheatsheet overlay and hold-modifier hints (Phase 4).

## Test plan

- `make test-web` green (typecheck + unit + biome), run on the final tree.
- All pre-existing frontend tests pass unmodified; new migration suite (`keybindingsMigration.test.tsx`) pins the legacy guard contract (chords from inputs, modal interplay, IME suppression, claimed-event handling, FocusScope Tab).
- Module suites: 87/87; two fix rounds reviewed and re-reviewed (per-binding `defaultPrevented` opt-out; multi-instance action registration), plus a final whole-branch review with one ruled fix wave (extra-modifier preservation).

🤖 Generated via subagent-driven development with per-task and whole-branch review.

PR #874, opened 2026-09-04
feat(web): keybindings Phase 4b - capture-based binding editor in settings

obra:
Phase 4b of the WebUI keybinding system, stacked on #871 (Phase 4a). Full stack: #853 survey → #860 dispatcher → #861 persistence → #863 navigation → #871 final map + cheatsheet + hold hints → **this PR: the editor**.

## What ships

**Capture-based binding editor** in Settings → Keybindings (replaces the read-only listing):
- Click a row's chord to enter capture mode — "Press new shortcut…" with live modifier echo; **Enter** saves, **Escape** cancels (capture's Escape wins over the settings pane's close binding; pinned against the real dispatcher).
- **Pre-flight conflict rejection**: an edit that would collide with another binding is rejected *before any hub write*, with an inline message naming the chord, scope, and the holding action. The pre-flight simulation is argument-identical to the reconcile path's own validation call, so the editor never accepts what the hub reconcile would skip.
- **Unbind** and **Reset to default** per action.
- Unsupported hubs keep the read-only listing with the existing status text.
- The cheatsheet row's `?` character-key trigger stays managed by its toggle (read-only note in the editor); the `⌘/` base chord is editable. Capture is single-press only by design (the default map is single-press throughout).
- Folds in the two parked 2b items the editor makes live: patchOverrides pre-flight validation, and symmetric hubError/conflict clearing.

**Docs**: keybindings.md documents the editing workflow, conflict semantics, and reset behavior.

**Also**: a disclosed repair to the 4a-era editor stylesheet for the design-token contract (3 focus-ring lines + one exact-path exception) — surfaced by `make test-web`'s web-lint, which the Task 5 acceptance gate didn't run.

## Verification

- 4581 frontend tests green (272 files), tsc clean, biome clean.
- `make test-web` PASS (web-typecheck, web-test, web-lint), re-verified independently by the controller.
- SDD per-task reviews + final whole-branch review: **Ready**, 0 Critical / 0 Important; 8 deferred minors triaged fine-as-deferred (notables: Reset in a reclaimed-chord state is rejected with a recoverable-but-confusing message; a failed patch surfaces both inline and in the pre-existing top-level error block).

## Notes for reviewers

- Capture mode `preventDefault()`s + `stopPropagation()`s every keydown; the dispatcher listens on window in bubble phase and keydown-only, so captured keys cannot leak to the shell (pinned by a test against the real dispatcher using rail.toggle, which opts out of the defaultPrevented gate).
- Bare-character chords (plain `A`) are authorable and intentionally NOT governed by the character-key toggle (which manages only the built-in `?` trigger); documented.

### DIFF
--- a/cmd/evener-hub/frontend/src/keybindings/chord.ts
+++ b/cmd/evener-hub/frontend/src/keybindings/chord.ts
@@ -117,6 +117,13 @@ export function chordDisplayKeys(chord: Chord): string[] {
   return [...chord.modifiers.map((m) => MODIFIER_DISPLAY[m] ?? m), key];
 }
 
+/** One modifier's display label (Ctrl / ⌘ / verbatim) - the chordDisplayKeys
+ * mapping for the capture editor's modifier-only live preview, which has no
+ * key yet. */
+export function modifierDisplayKey(modifier: string): string {
+  return MODIFIER_DISPLAY[modifier] ?? modifier;
+}
+
 /** One press as speakable words: "⌘+K" on Apple platforms, "Ctrl+K" elsewhere. */
 export function formatChord(chord: Chord): string {
   return chordDisplayKeys(chord).join("+");
--- a/cmd/evener-hub/frontend/src/keybindings/defaults.test.ts
+++ b/cmd/evener-hub/frontend/src/keybindings/defaults.test.ts
@@ -5,7 +5,9 @@ import {
   CHARACTER_KEY_TRIGGER_BINDING_ID,
   CHEATSHEET_SCOPE,
   DEFAULT_BINDINGS,
+  defaultBindingShapesForAction,
   registerDefaultBindings,
+  registerDefaultBindingsForAction,
   SETTINGS_SCOPE,
 } from "./defaults";
 import { createKeybindingDispatcher, type KeybindingDispatcher } from "./dispatcher";
@@ -374,5 +376,25 @@ describe("default bindings through the dispatcher", () => {
       ACTIONS.selectionQuote,
       ACTIONS.settingsOpen,
     ]);
+  });
+});
+
+describe("register vs shapes agreement", () => {
+  // The preflight (defaultBindingShapesForAction) and the mutation
+  // (registerDefaultBindingsForAction) share one predicate for the
+  // conditional "?" entry; if they ever diverge the validation can accept a
+  // restore that then throws, or reject one that would succeed.
+  test("both paths include or exclude the ? trigger for the SAME pref value", () => {
+    for (const pref of [true, false]) {
+      const registry = createKeybindingsRegistry();
+      const registered = registerDefaultBindingsForAction(registry, ACTIONS.cheatsheetToggle, {
+        characterKeyTriggers: pref,
+      }).map((b) => b.id);
+      const shaped = defaultBindingShapesForAction(ACTIONS.cheatsheetToggle, {
+        characterKeyTriggers: pref,
+      }).map((s) => s.id);
+      expect(registered).toEqual(shaped);
+      expect(registered.includes(CHARACTER_KEY_TRIGGER_BINDING_ID)).toBe(pref);
+    }
   });
 });
--- a/cmd/evener-hub/frontend/src/keybindings/defaults.ts
+++ b/cmd/evener-hub/frontend/src/keybindings/defaults.ts
@@ -323,16 +323,39 @@ export function registerDefaultBindings(registry: KeybindingsRegistry): Binding[
   return registered;
 }
 
+/** The one CONDITIONAL default-map entry's inclusion test, shared by the
+ * registration path here and the simulation path
+ * (defaultBindingShapesForAction) so the actual restore and the validation
+ * simulation can never disagree (finding 27): the "?" character-key
+ * trigger exists only while the characterKeyTriggers pref is on. */
+function includeDefaultEntry(entryId: string, options?: { characterKeyTriggers?: boolean }): boolean {
+  return !(options?.characterKeyTriggers === false && entryId === CHARACTER_KEY_TRIGGER_BINDING_ID);
+}
+
 /** Registers only the given action's default entries (plus the $mod twin),
  * the per-action equivalent of registerDefaultBindings the overrides store
- * uses to restore one action's defaults. Throws on an unknown action id. */
-export function registerDefaultBindingsForAction(registry: KeybindingsRegistry, actionId: string): Binding[] {
+ * uses to restore one action's defaults. Throws on an unknown action id.
+ *
+ * `characterKeyTriggers: false` mirrors the live registry when the
+ * cheatsheet character-key pref is off (the cheatsheetController has no
+ * "?" binding registered then): the restore skips the conditional entry,
+ * so it cannot collide with a user rule claiming exactly [Shift]+? -
+ * registerBinding throws on an exact same-scope serialization match, and
+ * the reconcile's strip runs AFTER this mutation (the ordering contract),
+ * so the collision would precede the cleanup. The controller re-registers
+ * the entry itself when the pref turns back on. */
+export function registerDefaultBindingsForAction(
+  registry: KeybindingsRegistry,
+  actionId: string,
+  options?: { characterKeyTriggers?: boolean },
+): Binding[] {
   const inputs = DEFAULT_BINDINGS.filter((input) => input.actionId === actionId);
   if (inputs.length === 0) throw new Error(`unknown keybinding action "${actionId}"`);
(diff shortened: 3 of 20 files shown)


obra:
roborev findings addressed in 7b7e357: (1) bare-Escape capture now documented as cancel-only per the VS Code convention, with Escape+modifier combos still capturable and a pinning test; (2) saves now compose the PATCH payload from the hub's raw rules, so skipped rules (e.g. from a newer client) pass through untouched — pre-flight rejects only edit-attributable warnings; (3) capture uppercases ASCII a-z only; (4) pointerdown-outside now cancels capture. Scoped re-review: all four ADDRESSED, no new breakage.

obra:
Second-round findings addressed in fea0b1b: capture maps the spacebar to the grammar's canonical key name (verified against tinykeys' source) so Space bindings round-trip, and IME composition events (isComposing / Process / Unidentified) no longer record chords or trigger save/cancel. Scoped re-review: both ADDRESSED, no new breakage. The remaining 'advertised-but-unwritable hub state' note is parked: the protocol advertises KeybindingsSettings unconditionally and exposes no writability probe, so the frontend can only learn it from a failed write — which the existing hubError path already surfaces. A protocol-level health signal would be hub-side work.

obra:
Round 3 in 48df295: the stale-hub editing window is closed — editing now requires state loaded for the current client generation (UI gate + a defense-in-depth patchOverrides guard), and client replacement un-applies the previous hub's overrides from the registry. The literal-+ finding resolved as not-a-bug: tinykeys 4.0.0's lookbehind split parses Shift++/Control++ with + as the key and its matcher fires on the + keydown (verified against the installed library source, empirically, and pinned by a capture-path test). Note: while a patch error is displayed, editing stays locked until a successful refresh — recovery is reconnect or the next hub notification.

obra:
Round 4 in 7c5338f9: unapplyAllOverrides is now atomic (snapshot, remove-all, restore, rollback — the same pattern as the reconcile path, so a reclaim-shaped registry can't wedge or leave an action unbound); patchOverrides rejects while a refresh is in flight; and the restore simulation respects the character-key toggle so Reset no longer falsely conflicts with a Shift+? override. On 'disconnect leaves overrides firing': that's deliberate — on transient disconnect the UI and registry read the same live state, rows show the last-known overrides read-only, editing disables, and chords keep working; un-apply fires only on client replacement. Scoped re-review: all ADDRESSED, no new breakage.

obra:
Round 5 in 535d169b: support loss (unsupported hub) now un-applies stale overrides so the registry matches the 'built-in defaults' status text — transient disconnect of a supported hub deliberately keeps them (pinned by test); capture saves carry a per-capture generation token so a click-away-cancelled save can't clobber a reopened capture; the IME guard gained the keyCode 229 fallback matching the dispatcher. Scoped re-review: all ADDRESSED, no new breakage.

obra:
Round 6 in 2fb55274: support loss now resets the hub payload state (loaded/revision/overrides/rawOverrides) alongside the registry un-apply, so a support flap can't resurrect the old hub's revision floor; notifications and post-patch applies require current support. Writes are serialized at the store level — each patch composes its expectedRevision/payload at execution time after the previous lands — so rapid cross-row edits land in order instead of self-inflicting conflicts. Scoped re-review: both ADDRESSED, no new breakage (queue error propagation, validation timing, and supported-window notification delivery verified).

obra:
Round 7 in 5173dd8d: enabling the character-key toggle now overlap-checks the ? binding against live bindings from other actions before registering (a conflicting Shift+? override authored while the toggle was off wins, with a warning through the store's channel; removing it restores ?). And a hub lifecycle change mid-capture (disconnect, support loss, replacement) cancels the capture with a generation bump so no stale save can touch the row. Scoped re-review: both ADDRESSED, no new breakage.

obra:
Round 8 in 22507f3e: queued writes are now fenced by the ready generation they were created under — a write queued against one hub's state rejects (with no wire request) if the client was replaced before it executed; same-generation queuing still composes at execution time and lands in order. Scoped re-review: ADDRESSED, no new breakage.

obra:
Round 9 in 39234637a: a character-key-toggle flip now re-validates the persisted overrides with the new pref (a Shift+? override skipped while the toggle was on applies when it turns off, and the stale warning clears — symmetric both directions), and Reset availability derives from the hub's raw rules so persisted-but-skipped rules can be dropped. Scoped re-review: both ADDRESSED, no new breakage.

obra:
Round 10 in 9eca4e4e: the generation fence no longer pollutes a new hub's state — a write from a dead generation rejects without touching hubError (current-generation rejections still surface truthfully); and refresh's error/settle paths now recheck support, so a refresh failing after support loss can't overwrite the clean defaults-claimed state. Scoped re-review: both ADDRESSED, no new breakage.

obra:
Round 11 in 3f97ecba: support transitions now advance the ready generation, so in-flight patch responses, queued writes, and notifications from before a flap are fenced out; the post-flap refresh applies normally and same-value re-notification doesn't bump the fence. Scoped re-review: ADDRESSED, no new breakage.

obra:
Round 12 in 8114e669: a changed-notification arriving during the initial load now marks the generation dirty and triggers exactly one follow-up refresh after it lands (no loop — the flag clears before recursion), and a queued write rejected because support is unsupported throws without touching hubError in the clean defaults state. Scoped re-review: both ADDRESSED, no new breakage.

obra:
Round 13 in f044ff62: the restore path is now pref-aware through the same shared predicate the validation simulation consults — with character-key shortcuts off, restoring cheatsheet.toggle skips the conditional ? entry, so an exact [Shift]+? user rule can no longer collide mid-reconcile; the reconcile re-adds ? when the toggle returns. Row inline errors now clear when a confirmed hub payload clears hubError. Scoped re-review: both ADDRESSED, no new breakage.

obra:
Round 14 in d8654d53: the reconcile re-checks overlap even when the ? binding is already registered (a foreign bare-? rule that appears after registration now wins, with the warning, and ? returns when it's removed), and capture honors its first-chord-stands contract — further non-modifier keys are ignored after a chord is recorded. Scoped re-review: both ADDRESSED, no new breakage.

obra:
Round 15 in 3bc98f46: support loss now checks the un-apply's result — on a rolled-back restore (wedged registry), hub state is retained (payload stays reconcilable) and a retryable hubError surfaces; the retry trigger is the next connection-driven setSupportFromConnection. Scoped re-review: ADDRESSED, no new breakage.

obra:
Round 16 in 68d5469d: the client-replacement path now honors the un-apply result just like support loss (rollback retains the outgoing state with a retryable error instead of presenting a clean new generation; revision resets to 0 there since revisions are hub-scoped), and row errors from preflight/fence rejections now clear when a confirmed payload applies for the current generation. Scoped re-review: both ADDRESSED, deviation judged sound, no new breakage.

obra:
Round 17 in f6568d05: the support-loss rollback path now resets revision like the rewire path (revisions are hub-scoped), so a flap-back refresh with a lower revision applies instead of being silently discarded, and the unsupported status line defers to the rollback error instead of claiming defaults while overrides are still live. Scoped re-review: both ADDRESSED, no new breakage.

obra:
Round 18 in 87362a0c: the support-loss rollback window now marks the retained state unconfirmed (loaded: false) — a notification arriving before the flap-back refresh takes the dirty-flag follow-up path instead of applying unverified, and the confirmed refresh is the point that re-opens editing. Scoped re-review: ADDRESSED, no new breakage.

obra:
Round 19 in 98d8ba0c: the hubError alert now carries a Retry control (supported hubs only, disabled while a refresh is in flight) wired to refreshOverrides — a successful refresh clears the error and restores editability, so a transient failure no longer requires a reconnect. Scoped re-review: ADDRESSED, no new breakage.

obra:
On the re-raised '+' finding (keybindings.tsx:222-226): this was verified non-bug in round 3 and is pinned by test. Empirically against the installed tinykeys 4.0.0: parseKeybinding('+') → key '+', parseKeybinding('Control++') → {req:[Control], key:'+'} (its lookbehind split at tinykeys.mjs:78 handles the trailing delimiter), and the matcher (:101-108) fires on the + keydown. keybindings.test.tsx:829 pins capture → serialize → parse round-trip AND a real keydown firing the action. No change needed — the finding's 'empty/unmatchable' mechanism doesn't occur in this version.

obra:
Superseded by #884, which consolidates the whole keybindings stack (phases 2a–4b) onto current main as 5 cleaned-up commits with all review rounds folded in. Closing in favor of that PR.

PR #879, opened 2026-09-04
feat: Google Vertex express-mode key, live discovery, and stored credential JSON

obra:
## Summary

Makes Google Vertex AI usable from the web hub with a pasted credential, and adds live Gemini model discovery for Vertex.

- **`google-vertex-express` provider** (curated, overlay-only): Gemini on Vertex AI in express mode, authenticated with a Google Cloud API key (`x-goog-api-key`) at `https://aiplatform.googleapis.com/v1`. Built on `google` so it carries exactly the Gemini rows. New env vars `GOOGLE_VERTEX_API_KEY` and `GOOGLE_VERTEX_EXPRESS_BASE_URL`. Listing stays unsupported because Vertex's publisher-model listing rejects API keys by policy (`ModelGardenService.ListPublisherModels` → 401), so express instances are catalog-only. The hub offers the API-key field for it with no hub code.
- **Live model discovery for `google-vertex` (ADC) instances**: the `google` protocol's `ListModels` branches on the `vertex-location` host rule to `GET {host}/v1beta1/publishers/google/models` (paginated) and keeps `gemini-*` ids at launch stage GA/PUBLIC_PREVIEW whose id contains none of `tts|embedding|image|live|transcribe|translate|omni|audio` — a heuristic, because the listing carries no capability data; pinned by a 27-id fixture yielding exactly 13 ids.
- **Quota project on user credentials**: the `gcp-adc` authenticator sends `x-goog-user-project` = `GOOGLE_VERTEX_PROJECT` when the credential is an `authorized_user` credential (ADC login or stored JSON), which Google requires for the project-less listing call. Service-account and metadata credentials send no header (they are attributed to their own project, and naming one needs `serviceusage.services.use`).
- **Stored credential JSON for `gcp-adc` instances**: the registry prefers a credentials-store entry over the ADC file (`Source: "store"`); the authenticator mints tokens from it (`google.CredentialsFromJSON`, cached by instance + value digest); new hub RPC `evener/auth/credentialJson/set` validates before storing and `ApiKeySet` refuses `gcp-adc` instances; the frontend gains a "Set credential JSON" action sharing one dialog component with the API-key dialog. A pasted `application_default_credentials.json` or a service-account key both work.
- **Catalog refresh**: models.dev snapshot refreshed; `gemini-3.8-flash` is the default for `google` and `google-vertex`.
- Docs (`llm-providers.md`, `llm-provider-config-and-launch.md`, `environment.md`, README, registry spec §5.1/§9.4), wire-capture goldens, an offline shape pin, and three opt-in live pins.

Spec: `docs/superpowers/specs/2026-09-04-google-vertex-express-and-discovery-design.md`. Plan: `docs/superpowers/plans/2026-09-04-google-vertex-express-and-discovery.md`.

## Verification

- `make vet`, `make lint`, `make test` green on the final tree (all modules and the web gate).
- Live, against project `jesse-coding-agents` on 2026-09-04: express-key `generateContent` → 200; ADC listing → 200 with the spec's 13 ids including `gemini-3.8-flash`; stored `authorized_user` JSON → 200; two real hub sessions (instances based on `google-vertex-express` and `google-vertex`, the latter credentialed by pasting the ADC JSON through the new RPC) each answered on `gemini-3.8-flash`. Total spend: cents.
- Not verified live: a service-account credential (none available). With the header limited to user credentials, that path behaves exactly as before this branch.
- The pre-existing `TestLiveVertexOneRequest` (Claude via `google-vertex-anthropic`) 404s in this project: it has no Claude Model Garden access. Environmental, unchanged by this branch.

## Notes for reviewers

- `google.CredentialsFromJSON` is deprecated upstream in favour of typed parsers that accept one credential type each; this code must accept both `authorized_user` and `service_account`, so it keeps the untyped parser with a reasoned `//nolint:staticcheck` at both call sites. Revisit with the `cloud.google.com/go/auth` migration.
- Pre-existing overlay invariant tests now walk the `base` chain (the express row is the first curated provider inheriting rows purely through `base`); the walk stops at `inherit_models = false`, and the `openai-codex` own-rows invariant is pinned directly.
- Deferred, non-blocking: no iteration cap on the Vertex listing loop (context-bounded); token sources for a replaced stored credential are not pruned; the hub add dialog keys instance vars by env-var name rather than template name (pre-existing, noted in spec §8).
- Follow-up candidate: "Sign in with Google" (user OAuth in the hub) — feasible via browser + PKCE + loopback like the Codex flow; needs a registered OAuth client. Out of scope here (spec non-goals).


obra:
Round-1 fixes pushed (`4c24122aa`..`1813b68ed`), plus the CI blocker.

**CI — `static-build` / `make fuzz-gap-check`**: `llm/providers/tokenauth` gained a `json.Unmarshal` decode seam with no fuzz target. Added `FuzzCredentialJSON` (seeds for every credential shape plus malformed inputs; oracles: never panics, a nil validation result implies an allowlisted type and a consistent user-credential classification, invalid JSON and disallowed types always error) and registered it in `scripts/fuzz/fuzz-targets.txt`. Gate now reports all 68 decode/parse packages covered. (`77c02b3de`)

**F1 — credential-type allowlist**: `ValidateCredentialJSON` and the stored-credential path now share one pre-parse check (`checkCredentialJSON`): valid JSON, then `type` ∈ {`service_account`, `authorized_user`}; `external_account` and other types are refused before `google.CredentialsFromJSON` runs, at paste time and at first request. Tests cover rejection, acceptance, missing type, and the stored path with the parser seam unreachable. Docs and spec §4 amended. (`4c24122aa`, `9413e3ea5`, `1813b68ed`)

**F2 — legacy store values under `gcp-adc` names**: the registry now treats a store entry as a credential only when it is a JSON object; anything else (e.g. a stale API key) is ignored with a warning naming the remedy, and resolution falls through to ADC as before this branch. Covered with and without an ADC file. (`2088737ee`)

**F3 — dialog vars keyed by env name**: `ProviderDescriptor.VarsEnv` is now a template→env map (mirroring `registry.Transport.VarsEnv`); the add dialog labels inputs by env name but submits `vars` keyed by the template name, so a typed base URL takes effect. Test added proving the express input submits `vars: { BASE_URL: … }`. Generated TS regenerated. (`adade448d`)

Live re-check on the fix head against `jesse-coding-agents`: express key → 200, ADC publisher listing → the same 13 ids, stored `authorized_user` JSON → 200.

obra:
Round-2 fix pushed (`84d9ea57f`).

**TUI label for a stored `gcp-adc` credential**: `authSourceLabel` now says "stored credential JSON" when the hub advertises the `credentialJson` auth mode (which it does only for `gcp-adc` instances), and "stored API key" otherwise — the same split the web pane's `storedLabel` makes. The logout/clear messages were already scheme-neutral ("Removed the stored credential for …"). Pinned through a new authwire fixture scenario, `status/store-credential-json` (a `gcp-adc` instance whose credential is a stored `authorized_user` JSON), generated by the hub's fixture test and consumed by the TUI wire test.

obra:
Round 3 findings addressed on this branch (three commits, one per finding):

- **F1 (Medium) — registry accepted any JSON object as a stored gcp-adc credential while the authenticator only accepts `service_account` / `authorized_user`.** The check now lives in one place: `registry.CheckCredentialJSON` (with `CredentialJSONType` and `AllowedCredentialJSONTypes`) in `llm/registry/credential_json.go`. The registry runs it when a gcp-adc instance's store entry resolves; an unsupported type (for example `external_account`) is ignored with a warning naming the remedy and resolution falls through to application-default credentials. `tokenauth.ValidateCredentialJSON`, the stored-credential token source, and the hub's `evener/auth/credentialJson/set` all delegate to the same function, so a value the registry resolves is one the authenticator will mint from. Error texts are unchanged; the `FuzzCredentialJSON` target now exercises the shared implementation. `docs/llm-providers.md` step 5 updated. (03e74abbc)
- **F2 (Medium) — the spawn gate told an unconfigured implicit gcp-adc provider to set a key via `evener/auth/apiKey/set`, which refuses gcp-adc.** `validateProviderCredentials` now has a gcp-adc branch: run `gcloud auth application-default login`, set `GOOGLE_APPLICATION_CREDENTIALS`, or store a credential JSON via `evener/auth/credentialJson/set`, plus the instance's base-URL variables. Test drives it with an empty HOME and asserts the old advice is gone. (557db1fa3)
- **F3 (Low) — web pane said "Clear stored key" for a stored gcp-adc credential JSON.** Sheet button, confirm dialog title/body, and toast are keyed on the instance advertising the `credentialJson` auth mode: "Clear stored credential JSON" / "Stored credential JSON cleared for X". Non-gcp-adc strings unchanged. (a57308707)

Gates on the new head: `make vet`, `make lint`, `make test`, `make fuzz-gap-check`, `make test-web` all green. Live pins re-run against Vertex on this head: express key request 200, ADC publisher-model listing (13 text models), stored authorized_user JSON request 200.

obra:
Round 4 findings addressed on this branch:

- **F1 (Medium) — `google-vertex-express` inherited google's non-Gemini rows.** Confirmed against the live express endpoint before changing anything: `gemma-4-31b-it`, `lyria-3-pro-preview`, and `deep-research-preview-04-2026` all return 404 ("not found or your project does not have access") with the express key, while `gemini-2.5-flash-lite` returns 200. Fix: a new overlay key, `inherit_models_matching = ["<glob>", …]`, keeps only the inherited rows whose id matches a glob (same `*` semantics as `[providers.X.models."<glob>"]`); a provider's own rows are always kept, and the base's layer maps are never mutated. The express row sets `inherit_models_matching = ["gemini-*"]`. Tests: synthetic base/derived fixture (matching rows kept, own row kept, base untouched), validation errors (needs `base`, conflicts with `inherit_models = false`, empty pattern), clone/write round-trips, the curated-overlay express test now requires every row to be `gemini-*`, and an embedded-snapshot test proves the filter (google itself still carries a non-Gemini row; express does not). Docs and the two specs updated; the design spec's claim that basing on `google` yielded "exactly the Gemini rows" was wrong and is corrected. (09c9541db, lint follow-up 5c0ecee34)
- **F2 (Low) — spawn-gate message read as if `GOOGLE_APPLICATION_CREDENTIALS` were required.** The three credential sources were already alternatives, but the doubled "set" made it ambiguous. The message now states the requirement: "it needs application-default credentials (run `gcloud auth application-default login` or set GOOGLE_APPLICATION_CREDENTIALS) or a stored credential JSON (evener/auth/credentialJson/set), and GOOGLE_VERTEX_LOCATION, GOOGLE_VERTEX_PROJECT set". Test pins the parenthesised alternatives. (d3aa59427)

Gates on the new head: `make vet`, `make lint`, `make test`, `make fuzz-gap-check` all green. Live pins re-run against Vertex on this head: express key request 200, ADC publisher-model listing (13 text models), stored authorized_user JSON request 200.

obra:
Round 5 findings addressed on this branch:

- **F1 (Medium) — `ProviderDescriptor.varsEnv` changed from a list to a map under an unchanged v3 protocol version.** Confirmed: the TUI decodes `evener/instance/list` through the shared Go types, so a TUI built from main against a hub from this branch would fail to decode the list after a successful handshake. Fixed additively (repo owner's decision): `varsEnv` is again the sorted list of environment-variable names main has, and a new field `vars` carries the template-name → env-var-name map the hub add dialog reads. `ProtocolVersion` stays `evener-appwire-v3`. An appwire test now marshals a descriptor and decodes it into the pre-`vars` shape to pin the wire compatibility; the hub descriptor test asserts both fields; the dialog reads only `vars` (a descriptor with the list but no map renders no variable inputs), and its tests were updated. `docs/appwire-protocol.md` and `types.gen.ts` regenerated. (4a89bbdd7)
- **F2 (Low) — the gcp-adc token-source cache never dropped a replaced stored credential.** The cache now holds one source per instance together with the digest of the credential it was built from; a request whose digest differs rebuilds the source and replaces the entry, and ADC (digest "") is replaced the same way when a stored credential arrives. Tests cover rotation A→B→A (three parses, one entry throughout), a second instance adding one entry, and ADC → stored replacement. Spec §4.3 reworded. (9a92e49bd)

Gates on the new head: `make vet`, `make lint`, `make test`, `make fuzz-gap-check`, `make test-web` all green. Live pins on this head: express key request 200; the ADC-based pins (publisher-model listing, stored authorized_user JSON) could not run this round because the workstation's ADC user session hit Google's reauthentication policy (`invalid_grant` / `invalid_rapt`) — the failure came back from Google's token endpoint on every ADC path, not from the code under review. They will be re-run once the ADC login is refreshed.

obra:
Round 6 findings addressed on this branch:

- **F1 (Medium) — a credential JSON with an allowed `type` but no key material passed every gate.** Confirmed against the library: `google.CredentialsFromJSON` (x/oauth2 v0.35.0) builds a service-account or user token source without checking any field, so `{"type":"service_account"}` passed the registry gate, the hub's set-time validation, shadowed a working ADC file, and failed only at the first request. The shared `registry.CheckCredentialJSON` now requires the fields each type needs to mint a token — `client_email` and `private_key`, or `client_id`, `client_secret` and `refresh_token` — reporting every missing one (`service_account credential JSON is missing client_email, private_key`). Absent, empty, and non-string values all count as missing. The registry, `tokenauth.ValidateCredentialJSON`, and `evener/auth/credentialJson/set` all delegate to it, so a partial JSON is refused at paste time and an existing partial store entry is ignored with a warning and resolution falls back to ADC. Tests added at all three layers plus two fuzz seeds; docs and spec updated. (4803550fb; cd3ad11db reads the required fields as raw JSON messages so an unrelated field Go cannot represent never decides the verdict, after review caught a decode error being ignored)
- **F2 (Low) — the spawn gate listed every `VarsEnv` entry as required, including `GOOGLE_APPLICATION_CREDENTIALS`.** Confirmed: `google-vertex` inherits an identity variable map from the models.dev snapshot that includes that name. The gate now distinguishes the two reasons an implicit gcp-adc provider is not an instance. If its base URL does not resolve (the provider is hidden), it says `provider google-vertex is not configured: set GOOGLE_VERTEX_LOCATION, GOOGLE_VERTEX_PROJECT`, using a new `Registry.MissingVars` that names only the unset base-URL variables (the host the `vertex-location` rule derives is reported as its input, the location). Otherwise it names the credential remedies alone, with no variable list. Registry tests cover the variable naming; the hub test now has one case per cause. (fe57b4c90)

Gates on the new head: `make vet`, `make lint`, `make test`, `make fuzz-gap-check` all green. Live pins: the express-key request passes on this head; the ADC-based pins remain blocked on the workstation's expired ADC session (Google `invalid_grant` / `invalid_rapt`), unrelated to the code, and will be re-run once it is refreshed.

obra:
Round 7 finding addressed on this branch:

- **Low — the Vertex model listing followed any non-empty `nextPageToken`, so a server that kept returning the same token was re-requested until the context expired.** Confirmed with a test server that always answers with the same token: the listing looped to the test's five-second deadline. `listVertexModels` now remembers the tokens it has followed and fails with `models.list: repeated page token "<token>"` the first time one comes back again; the test asserts the error after exactly two requests. (86c7f1967)

Gates on the new head: `make vet`, `make lint`, `make test`, `make fuzz-gap-check` all green. Live pins re-run on this head with a refreshed ADC session, covering everything since round 4: express-key request 200; ADC publisher-model listing (13 text models, through the new pagination guard); stored authorized_user JSON request 200 (through the tightened credential check and the one-source-per-instance cache).

obra:
Round 8 findings addressed on this branch:

- **F1 (Medium) — the Vertex listing was selected by the `vertex-location` host rule, which the `vertex-gemini` preset does not set.** A user provider built on the preset with its own base URL fell through to the Gemini-API listing, which appends the endpoint to the project-scoped base URL and decodes a payload shape Vertex never returns. The discriminator is now the transport's own shape: Vertex serves Gemini as publisher models, so a completion endpoint under `/publishers/google/models/` selects the Vertex listing (host-relative on v1beta1, paginated); the Gemini API's `/models/` path selects the generic one. A test drives a preset-built provider with no host rule and asserts the v1beta1 publisher listing; a second pins the discriminator on both endpoint shapes. The curated `google-vertex` listing still returns 13 text models live. (bf8021640)
- **F2 (Low) — a hidden Vertex provider with an invalid location was told to "set" it.** `MissingVars` is replaced by `Registry.UnresolvedBaseURL`, which returns unset variables and resolver problems separately; the derived host is reported as the location only when the location itself is unset. The spawn gate now prints `provider google-vertex is not configured: invalid GOOGLE_VERTEX_LOCATION "bad.host": …` for an invalid value and `not configured: set GOOGLE_VERTEX_LOCATION, GOOGLE_VERTEX_PROJECT` for unset ones. Registry and hub tests cover both. (f75f85d69)
- **F3 (Low) — stale docs.** The registry design spec's `vertex-gemini` transport entry now says `models_endpoint = /publishers/google/models`, and the `ListModels` comment describes the paginated Vertex listing alongside the single-page Gemini-API one. (18a2d4f86)

Gates on the new head: `make vet`, `make lint`, `make test`, `make fuzz-gap-check` all green. Live pins on this head: express-key request 200; ADC publisher-model listing 13 text models through the new discriminator; stored authorized_user JSON request 200.

obra:
Round 9 finding addressed on this branch:

- **Medium — the TUI credentials panel had no action for the `credentialJson` auth mode the hub now advertises for gcp-adc instances, so Enter did nothing.** Confirmed: the panel acts on Enter only for `apiKey` and `oauth`. The design spec places credential-JSON pasting in the web hub (§4.4), and the TUI has no multiline paste, so rather than stay silent the panel now shows a notice under the instance's row: it uses application-default credentials, and a credential JSON is pasted in the web hub's Credentials settings. The notice clears when the cursor moves or the list refreshes, and no command is emitted. Test-first: the new panel test failed on the old code (no command, no notice) and passes now. A TUI paste flow itself is a separate decision, not part of this PR. (51e8dc783; 673af2efe after review: the notice names the hub's "Providers & credentials" row, clears on any key other than Enter, and is pre-broken into lines that fit under the row so the overlay's wrap never splits the key phrases)

Gates on the new head: `make vet`, `make lint`, `make test`, `make fuzz-gap-check` all green. The change is TUI-only; the live Vertex pins last ran green on the previous head, whose non-TUI code is unchanged here.

obra:
Round 10 finding addressed on this branch:

- **Medium — the hub's cached model listing was never dropped when credentials changed.** Confirmed, and it predates this branch: the threads store caches `model/list` in one session-long slot, and the `evener/auth/updated` notification only refetched the instance list. A Vertex credential JSON is the first credential whose whole point is enabling discovery, so the stale picker showed here first, but the same applied to any API key set after the picker loaded. Fixed at the root: the threads store now advances an epoch and drops the cache on `evener/auth/updated`; a listing already in flight when the notification arrives still answers its caller but no longer becomes the cache, and the in-flight dedupe slot is cleared only by the request that owns it. Tests: the notification makes the next listing re-request; an in-flight listing that started before it does not repopulate the cache; a caller arriving after the notification does not de-dupe onto the stale in-flight listing; and that listing settling does not evict the newer request from the dedupe slot. (d16305502; tests completed in 575bf0363 and e8bdfcfa2 after review, with each guard's removal verified to fail a test)

Gates on the new head: `make vet`, `make lint`, `make fuzz-gap-check`, `make test-web` (typecheck, tests, lint) all green. Frontend-only change; the live Vertex pins last ran green on the head whose non-frontend code is unchanged here.

obra:
Round 11 finding addressed on this branch:

- **Medium — a service-account credential whose `private_key` will not parse passed the gate.** Confirmed against the library: Google's signer parses the key only at the first token request (PEM block if present, then PKCS#8, then PKCS#1, RSA only), so a stored credential with unusable key material was selected over a working ADC file and failed at the first request. The shared `registry.CheckCredentialJSON` now parses a service-account `private_key` the same way, offline, with the standard library only; an unusable key is refused at paste time by the hub, ignored with a warning by the registry (falling back to ADC), and refused by the authenticator. Tests generate RSA and ECDSA keys at run time, so no key material lives in the repository; the fuzz seeds were updated the same way. Docs and spec updated. (c75633659)

Gates on the new head: `make vet`, `make lint` (including the secret scan), `make test`, `make fuzz-gap-check` all green. Live pins re-run on this head: express-key request 200; ADC publisher-model listing 13 text models; stored authorized_user JSON request 200.

obra:
Round 12 finding addressed on this branch:

- **Medium — a stored credential was classified from the JSON the Google parser echoes back rather than from the stored JSON itself.** Checked against the library: x/oauth2's `CredentialsFromJSON` does set `Credentials.JSON` to its input, so the header was not actually dropped with the real parser, and the live stored-credential pin has passed on every head. But the authenticator already holds the stored bytes and its parser seam need not echo them, so the dependence was unnecessary. A stored credential is now classified from the stored value; only application-default credentials trust the JSON the library read from the file. Test: a parser that returns no JSON echo still yields the `x-goog-user-project` header for a stored `authorized_user` credential. (dec66c9f8)

Gates on the new head: `make vet`, `make lint`, `make test`, `make fuzz-gap-check` all green. Live pins re-run on this head: express-key request 200; ADC publisher-model listing 13 text models; stored authorized_user JSON request 200.

obra:
Round 13 finding addressed on this branch:

- **Medium — the spawn pane kept its own scoped `model/list` cache, so round 10's invalidation of the threads store's cache left a mounted spawn pane stale after a credential change.** Confirmed: the pane caches its harness- and cwd-scoped listing promises in a ref reset only when the client changes. The pane now keys that cache on a credentials generation advanced by `evener/auth/updated`; the loader identities change, the debounced catalog effect reloads, and the pickers share the reloaded promise. Test: after the mount-time listing, the notification produces a second listing and the picker offers a model that only became discoverable afterwards, without a third request. A picker panel that is already open when the notification lands keeps its list until reopened; that is the shared picker widget's existing behaviour and out of scope here. (fc5877629; comment correction 1efe7397d)

Gates on the new head: `make lint`, `make test-web` (typecheck, tests, lint) all green; Go code unchanged since the last full run. Frontend-only change; the live Vertex pins last ran green on the previous head, whose non-frontend code is identical.

obra:
Round 14 finding addressed on this branch:

- **Medium — the token-source cache identified a credential by the stored JSON's digest alone, which is empty for every other source, so an instance re-resolved with no credential matched its cached ADC source and kept sending that token.** Confirmed: with the ADC file removed and the registry reloaded, the resolution's source changes from `adc` to `none`, and the cache still answered. The identity is now the resolution's credential source, plus the stored JSON's digest for a stored credential, so any source change rebuilds the source; `adc` to `none` re-runs the ADC lookup and surfaces its error. Test: an ADC-backed instance whose credential then resolves to none gets an error and no Authorization header. Which ADC file backs an `adc` resolution is not tracked; the resolution does not carry it, and a same-source swap is outside this change. Spec §4.3 updated. (087353cec; 9f398cf1d after review: the missing-ADC error now also offers a stored credential JSON as a remedy, the identity comment says an ADC file replaced in place keeps its identity until the next process, and the spec's test plan lists the source-change rebuild)

Gates on the new head: `make vet`, `make lint`, `make test`, `make fuzz-gap-check` all green. Live pins re-run on this head: express-key request 200; ADC publisher-model listing 13 text models; stored authorized_user JSON request 200.

obra:
Round 15 finding addressed on this branch:

- **Medium — the credential gate accepted any `token_uri`, and Google's library sends the refresh token or the signed assertion to whatever endpoint the file names.** Confirmed in the library: both credential types take `token_uri` as the token endpoint, defaulting to Google's when absent. This is the same class of concern that made round 1 refuse `external_account`. The shared `registry.CheckCredentialJSON` now accepts `token_uri` only when it is one of Google's own token endpoints (the current one or the legacy `accounts.google.com` one); absent is fine; a non-string or any other value is refused with `token_uri … is not Google's OAuth token endpoint`. The field is read through a tagged decode exactly as the library reads it, so a case variant of the key such as `Token_Uri` is caught too (the independent review demonstrated the case-variant bypass against a local server before the follow-up closed it). A top-level `installed`/`web` block, which makes the library return an OAuth client configuration with no token source and crash on first use, is refused as well. Enforced at paste time by the hub, at selection by the registry (falling back to ADC with a warning), and by the authenticator. `auth_uri` serves only the browser consent flow and `universe_domain` is not consulted by the token source, so neither is checked. Tests at all three layers; docs and spec updated. (360d25afb; tagged decode, client-config refusal, and legacy endpoint in 4edefa98d after review; cd6f8f6d2 extends the tagged decode to every field the gate reads, so a duplicate key under another spelling can no longer hand the signer a value the gate never saw)

Gates on the new head: `make vet`, `make lint`, `make test`, `make fuzz-gap-check` all green. Live pins re-run on this head: express-key request 200; ADC publisher-model listing 13 text models; stored authorized_user JSON request 200.

obra:
Round 16 finding, checked and not adopted:

- **Low — the gate accepts raw DER key bytes, which "Google's signer expects PEM-encoded" material would reject.** Checked against the signer this branch uses, x/oauth2 v0.35.0: `jwt.(*jwtSource).Token` calls `internal.ParseKey`, which treats the PEM block as optional (`block, _ := pem.Decode(key); if block != nil { key = block.Bytes }`) and then tries PKCS#8 and PKCS#1 on the bytes; its own error text reads "private key should be a PEM or plain PKCS1 or PKCS8". `checkServiceAccountKey` mirrors that sequence step for step, so any key material the gate accepts is key material the signer parses, and a raw-DER `private_key` (which can only reach a JSON string via escapes) does not fail at the first token request. Requiring a PEM block here would make the gate stricter than the signer and refuse credentials that work. No change.

obra:
Round 17 (review of the merge commit) addressed on this branch:

- **Medium — `inherit_models_matching` filtered every row accumulated so far, not just the inherited ones.** Confirmed: the glob ran inside the layer fold after a provider's own upstream models.dev rows had already folded in, so a provider with both an upstream entry and an overlay glob would lose its own rows, contradicting the documented rule. No curated provider has both today (the express row is overlay-only), but the semantics were wrong. The glob is now applied immediately after inheritance, when the record holds nothing but the base's rows, for curated records and user instances alike; the fold only records the key. Test: a synthetic provider with an upstream row that does not match the glob keeps that row and drops the base's non-matching one. (a8db2a848; b8b42c7e9 adds the user-instance test and corrects the spec's attribution after review)

Gates on the new head: `make vet`, `make lint`, `make test`, `make fuzz-gap-check` all green. Live pins re-run: express-key request 200; ADC publisher-model listing 13 text models; stored authorized_user JSON request 200.

obra:
Round 18 findings, on the previous head:

- **Medium — the spawn onboarding dialog (`ConnectProviderDialog`, which arrived from main in today's sync) has no action for the `credentialJson` auth mode**, so a gcp-adc instance there offers only "Test connection" and API-key guidance. Real, and created by combining main's new dialog with this branch's mode. Pasting a credential JSON still works from Settings → Providers & credentials, so it is a discoverability gap rather than a dead end. Tracked as a follow-up: add a `credentialJson` editor kind to that dialog reusing `CredentialJsonDialog`.
- **Low — the registry gate ignored typed optional fields such as `quota_project_id` that Google's parser decodes**, so a malformed value could be selected over ADC and then fail at parse time. Fixed: the gate now also decodes the credential into a mirror of the parser's entire credential-file shape (external-account fields, nested blocks, and the recursive `source_credentials` included, since the parser decodes all of it before looking at `type`), so any type mismatch the parser would reject is refused at the gate, at paste time, and at selection (falling back to ADC with a warning). The tokenauth fuzz target now pins the property directly: anything the gate accepts, the library parses. Table rows cover a numeric `quota_project_id`, an array `universe_domain`, a string `delegates`, a malformed `credential_source`, a string `token_lifetime_seconds`, and a malformed `source_credentials`. (f954bfd80, completed in 1a0d76c35 after review found the first mirror partial)
- **Low — a new frontend against an older v3 hub would see only `varsEnv` and render no variable inputs.** Not reachable: the hub embeds the frontend it serves (`cmd/evener-hub/webnext.go`, `//go:embed all:frontend/dist`), so the browser always runs the frontend built with the hub it is talking to; the mixed pair that matters, an old TUI against a new hub, is the one round 5 kept `varsEnv` a list for. No change.

obra:
Round 19 finding, checked and tracked as a follow-up:

- **Medium — the add-instance form renders an input for `GOOGLE_APPLICATION_CREDENTIALS` that nothing reads.** Confirmed: the curated `google-vertex` row inherits an identity variable map from the models.dev snapshot that lists that name alongside the project and location, and the registry only consults instance variables for template placeholders, so a value typed there is persisted and ignored. Not introduced here: main built `ProviderDescriptor.varsEnv` from the same map and rendered the same input; this branch's `vars` map carries the same entries. The right fix is the one suggested, exposing only variables the transport's templates reference (a registry helper listing template placeholders, applied when the hub builds descriptors), and it belongs with the other follow-up from round 18 (the onboarding dialog's `credentialJson` action) so this PR can land.

PR #881, opened 2026-09-05
fix(web): bound outbox storage stalls without unsafe retries

obra:
Safari could leave Send and Steer disabled indefinitely when IndexedDB stopped completing operations. Add a storage watchdog that preserves the submission's durable outcome: a successfully cancelled write leaves the draft available to retry; a write that cannot be cancelled remains pending and displays a storage status until its original outcome is known.

Timed-out schema upgrades are aborted to release the database open lock; already-completed upgrades retain the late-connection cleanup path.

Runtime startup registers listeners and schedules discovery in the background, so a stalled initial scan cannot delay a new durable submission. Committed submissions release the composer without awaiting discovery or display reads. Publish the committed record immediately so stalled or older projections cannot hide a pending start and misroute the next message. Apply the same commit boundary to recovery edits and resend.

Share pending submission ownership across dock tab remounts. Repeated clicks cannot enqueue another submission while the original is unresolved, and late completion cannot erase a newer draft. Clear submitted attachments independently of text edits while preserving new or explicitly replaced items, including after recovery merges. Queue-strip drains atomically consume active recovery ownership. Submission cleanup clears this mount's sent content while preserving a newer shared draft edited in another composer, including attachment cleanup. Recovery completion releases the consumed recovery ID across mounts and preserves newer text as an ordinary draft; internal recovery persistence does not count as a user edit. Isolate notification and storage-status listener failures from write outcomes. Failed enqueues retain no phantom thread pin. Fence projection reads when newer reads start, including when those newer reads fail, so stale snapshots cannot resurrect settled work.

Also fixes a thread cleanup race exposed by the full test run: a released pane's failed read could resume after reconnecting. Recheck ownership before and after readiness waits, with deterministic coverage for all three release windows.

The directory-picker browser guard now awaits its existing focus condition alongside creation readiness. This fixes a CI scheduling race exposed after merging main; a negative check confirms missing focus restoration still fails.

Validation:

- `make merge-approval-gate` passed after merging current main; the final frontend corrections were rechecked with both frontend gates at `d906f60a4`.
- `make test-web` passed: type checking, frontend tests, and Biome.
- `make test-web-browser` passed all five browser guards.
- Regressions exercise real stores and Composer controls with fake IndexedDB callback stalls, cancelled and uncertain writes, stale projections, remounts, simultaneous composers, atomic recovery drains, recovery resend, and throwing subscribers.
- Independent review has no remaining actionable findings.

Limit: this protects submission ownership within a page lifetime. A full browser reload still requires a separate durable draft-to-mutation handoff to prevent resubmitting restored text as a new message. The tests simulate storage callback stalls; they do not establish the underlying WebKit trigger or prove this fixes Safari's storage engine.

### DIFF
--- a/cmd/evener-hub/frontend/src/dev/spawnguard-entry.tsx
+++ b/cmd/evener-hub/frontend/src/dev/spawnguard-entry.tsx
@@ -433,12 +433,16 @@ async function exerciseDirectoryPicker() {
   nameInput.form?.requestSubmit();
   const created = `${directoryRoot}/a-new-directory-with-a-long-readable-name`;
   const deadline = performance.now() + 10_000;
-  while (document.querySelector<HTMLInputElement>('input[aria-label="Path"]')?.value !== created || confirm.disabled) {
+  // The path can render before the effect restores focus to a persistent control.
+  while (
+    document.querySelector<HTMLInputElement>('input[aria-label="Path"]')?.value !== created ||
+    confirm.disabled ||
+    !dialog.contains(document.activeElement)
+  ) {
     if (performance.now() > deadline) throw new Error("Directory creation did not settle");
     await new Promise((resolve) => requestAnimationFrame(resolve));
   }
   measure("created");
-  if (!dialog.contains(document.activeElement)) failures.push("Creation lost dialog focus");
   confirm.click();
   await new Promise((resolve) => requestAnimationFrame(resolve));
   if (!trigger.textContent?.includes(created)) failures.push("Confirmed path was not stored on the launch form");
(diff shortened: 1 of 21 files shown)


obra:
Review 4976: both findings are invalid for the current implementation.

1. The transaction bodies only await their own IndexedDB requests; they never await timers, network work, or another external promise, and never call `commit()`. The continuation after the final write returns its result. `beforeCommit` is a synchronous fault-injection seam used only by rollback tests. Automatic commit waits until request results have been handled ([IndexedDB transaction lifecycle](https://w3c.github.io/IndexedDB/#transaction-lifecycle)); no production post-commit body failure was identified. The timeout path separately retains uncertain writes until their terminal event. Code comments now document the constraint against external awaits.

2. `Promise.all` rejects immediately when either input rejects; it does not wait for the other input. Added a regression that aborts a real fake-indexeddb transaction while withholding the adapter's request callback, with the watchdog clock frozen. Enqueue rejects and storage remains empty without releasing that callback. The existing uncertain-commit and rollback regressions also pass.

No transaction outcome logic was changed for these findings.

PR #882, opened 2026-09-05
Add optional parent-history forks for delegates

obra:
Adds optional `fork_context: true` to `delegate`. Clean sessions remain the default and preferred path: put selected facts, decisions, and excerpts in a self-contained brief. Copy the parent's full context and history only when the assignment requires it.

Broadens runtime guidance to delegate whenever it could save time, improve quality, or provide a useful independent perspective, including sequential work.

- Forks require the parent's model and provider. They copy a fixed conversation snapshot, preserve compacted archives, and exclude the unfinished tool round and parent response-continuation handles.
- Children retain their own assignment, role, tools, sandbox, delegation allowance, and accounting. Descendants start clean unless they also opt in; resumed children use their saved snapshot rather than later parent history.
- Scripted-provider regressions exercise opt-in/default behavior, images and completed tool results, model-switch markers inside unfinished rounds, nested delegation, cold/direct resume, and malformed or incompatible requests.

## Validation

- `make merge-approval-gate` — passed on `439a5d1bc`, rebased onto main's Astra changes.
- `make vet` — passed.
- `go test -race ./agent -run '^TestDelegateForkContext' -count=1` — passed.
- `make test-web-browser` — all five guards passed; frontend sources are unchanged by the rebase.

The regression tests use scripted providers to verify Evener behavior. Live-model delegation behavior was not evaluated.

### DIFF
--- a/agent/delegate_fork_context.go
+++ b/agent/delegate_fork_context.go
@@ -29,10 +29,14 @@ func (s *Session) snapshotDelegateContext() ([]transcript.Entry, error) {
 		// parent's delivery receipts, client mutation IDs, usage, or server
 		// continuation handles. Those belong to its execution, not the child.
 		entry.Turn = schema.Turn{
-			Kind: t.Kind, Message: t.Message, Timestamp: t.Timestamp,
-			SteeringSource:   t.SteeringSource,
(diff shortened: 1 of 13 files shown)


obra:
astra assures me this is a false positive

PR #883, opened 2026-09-05
Explain automatic background job completion wakeups

obra:
## Summary

- append a system reminder after direct background shell-job launches
- explain that completion notifications arrive automatically, wake idle sessions, and require no explicit wait

## Testing

- `go test ./agent -run '^TestFormatShellResult' -count=1`
- `make lint`
- `make vet`
- `make test`

## Test coverage note

The change adds model-facing reminder prose only. Per `docs/developing-evener/testing.md`, this PR does not add an assertion that pins natural-language reminder or tool-result text.

### DIFF
--- a/agent/session_tools_shell.go
+++ b/agent/session_tools_shell.go
@@ -581,6 +581,7 @@ func formatShellResult(out shellToolResult) string {
 	// the command itself did not time out — the foreground wait did, and the
 	// command keeps running as a durable background job.
 	promoted := out.Mode == string(shellModeBackground) && out.TimedOut && out.JobID != ""
+	directBackground := out.Mode == string(shellModeBackground) && out.JobID != "" && !promoted
 
 	var foot []string
 	if out.ExitCode != nil && out.Mode != string(shellModeBackground) && !runTimeout {
@@ -601,7 +602,7 @@ func formatShellResult(out shellToolResult) string {
 			fmt.Sprintf("output accumulates durably; read it with read_transcript(transcript_ref=%q)", "job:"+out.JobID),
 			"completion arrives by notification — do not relaunch or poll",
 		)
-	case out.Mode == string(shellModeBackground) && out.JobID != "":
+	case directBackground:
 		foot = append(foot, "running in background as "+out.JobID)
 	case out.JobID != "":
 		foot = append(foot, fmt.Sprintf("output windowed — read more with read_transcript(transcript_ref=%q)", "job:"+out.JobID))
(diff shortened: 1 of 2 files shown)


obra:
The one-shot premise does not match the current `evener run` lifecycle. After `ProcessInput` returns, `cmd/evener/run.go:414-425` calls `Session.DrainJobTree` and re-drives the coordinator until session-owned managed jobs become terminal; `Session.Close` runs only after that drain.

`TestRunDrainsManagedShellBeforeExit` exercises an explicit `mode="background"` shell job in a one-shot run and requires its terminal `<job-notification>` to reach the model on the third round. I reran it against this branch:

```
go test ./cmd/evener -run '^TestRunDrainsManagedShellBeforeExit$' -count=1 -v
```

Both completed and failed job cases passed. The reminder is therefore accurate for one-shot runs too: they receive the automatic terminal notification during the drain rather than through an idle-session wake. No code change is needed for this finding.

obra:
Scope clarification implemented in `d553e7206`: the system reminder now applies whenever a managed shell job enters background mode with a `job_id`, covering both explicit background launches and foreground-timeout promotions. The two cases retain their existing distinct footer wording.

Verification rerun after the change:

- `go test ./agent -run '^TestFormatShellResult' -count=1`
- `make lint`
- `make vet`
- `make test`

All passed. The earlier one-shot conclusion is unchanged: `evener run` drains managed shell jobs and delivers their terminal notifications before closing.

obra:
Confirmed this finding and fixed it in `878f2f0da`.

The reminder and bracketed footer now share one atomic final line, and the actual job ID is the final footer field for both direct and foreground-promoted background jobs. This keeps the complete ID in the registry's head/tail truncation tail.

Added structural registry-level coverage with constrained `MaxChars`/`MaxLines` for both paths without pinning reminder prose. Verification passed:
- focused direct/promotion/formatter tests
- `make lint`
- `make vet`
- `make test`
- independent review: no findings

obra:
Follow-up `ff17b8b77` fixes the CI failures on `878f2f0da`.

Both `tests` and `fuzz` exposed the same timing-dependent test assertion: a promoted command is not guaranteed to emit partial output before promotion, so the corrected one-line model result can legitimately fit without registry truncation. I removed that incidental `res.Truncated` assertion. The required regression remains unchanged: under constrained `MaxChars`/`MaxLines`, the model-facing result must contain the complete actual job ID for both direct and promoted background jobs. The pre-fix two-line output still fails that invariant.

Verification after the correction:
- affected focused tests, 20 repetitions
- exact `evenerfuzz` seed replay, 20 repetitions
- `make lint`
- `make vet`
- `make test`
- `make fuzz`

All passed locally.

PR #884, opened 2026-09-05
feat(web,hub): consolidated keybindings system (dispatcher, persistence, navigation, cheatsheet, editor)

obra:
# Consolidated keybindings stack

One PR for the whole keybindings system, rebased onto current `main` with the history cleaned up: the 49 commits across the six stacked PRs are folded into 5 readable feature commits, with all 19 roborev rounds and the two post-review fixes squashed into the commits they belong to.

This PR supersedes #860, #861, #863, #871, and #874 (closed with pointers). #853 (the survey) already merged.

## The five commits

1. **`feat(web): keybinding dispatcher built on tinykeys`** — the `src/keybindings` module (chord parsing/formatting, action registry, default binding map, dispatcher) and the shell-chord migration. Per-binding `defaultPrevented` opt-out, `allowInModal`, Meta/Control twins, multi-instance registration, strict twins, `contenteditable=false` islands.
2. **`feat(hub,web): persistent keybinding overrides`** — the appwire wire contract, the hub-side overrides store and settings RPC handlers, and the frontend override machinery: rebind primitive with semantic validation, an overrides store with hub sync and atomic/retryable/restore-aware reconcile, a read-only settings section, and optional-modifier overlap detection.
3. **`feat(web): navigation keybindings and ask-dock keyboard operation`** — session-pane cycling, transcript scroll chords, Alt+PageDown/Alt+PageUp batch jump and focus return in the ask dock.
4. **`feat(web): final binding map, cheatsheet overlay, hold-modifier hints`** — the finalized default map (strict mod chords, ⌘, for settings, transcript jumps), the registry-sourced cheatsheet overlay on ⌘/ with a pref-gated `?` trigger, and hold-modifier hint chips.
5. **`feat(web): capture-based keybindings editor in settings`** — chord capture with canonical Space naming and IME guards, pre-flight conflict rejection with display-title holder names, unbind/reset, unsupported-hub fallback, last-visited settings section persistence, plus the 19 hardening rounds (stale-hub fencing, atomic un-apply, serialized writes, support-flap convergence, hub-error retry).

## Merge-conflict resolution notes (vs the stacked tips)

Main moved under the stack in two places; resolutions here:

- **AskDock.tsx / AskDock.test.tsx** — main's #854 (dock as a scrollable transcript row: `IntersectionObserver` focus greeting, `pendingGreeted` edge in the store, `AskDockAnnouncements` as the single live region) is combined with the Phase 3 keyboard contract (`FIRST_CONTROL_SELECTOR`, Alt+Page batch jump, focus return, `onKeyDown` on the dock div). Both test suites are spliced in whole.
- **TUI notification coverage** — `evener/settings/keybindings/changed` is added to `notifyMethodsDeliberatelyIgnored` (the TUI has no remappable-binding surface).
- **Lint modernization** — the 2b Go test helpers (`keybindingsChordPtr`/`keybindingsChord`) replaced with Go 1.26 `new(expr)` to satisfy main's current `modernize` lint.

Tree-equivalence to a true `main`+stack merge was verified with `git merge-tree`: only the two AskDock files conflicted there, and they are exactly the files resolved by hand here.

## Verification

- `make lint`, `make vet`, `make test` (all 8 modules incl. the frontend gate): PASS on the final head.
- Frontend: 2443 tests green; Biome clean.
- Real-Chrome interactive verification of the shipped chords (CDP-driven): clean; report at `interactive-test-report.md` in the p4 worktree (not committed). Alt+arrow session-cycling chords were covered by the Phase 3 automated suites rather than the headless run.
- The earlier stacked PRs carry the full roborev trail (19 rounds on #874, all fixed and verified; the one repeat finding — `+`-key capture — is a documented false positive with a pinning test at `keybindings.test.tsx`).

## Docs

`docs/web-ui/keybindings.md` covers the dispatcher architecture, the persistence contract, the final binding map, the cheatsheet overlay, hold hints, and the editor.

### DIFF
--- a/appwire/errors.go
+++ b/appwire/errors.go
@@ -24,6 +24,10 @@ const (
 	ErrorQueuedDrainPartial     ErrorInfo = "queuedDrainPartial"
 	ErrorMutationOutcomeUnknown ErrorInfo = "mutationOutcomeUnknown"
 	ErrorInternal               ErrorInfo = "internal"
+	// ErrorKeybindingsPostRename marks a keybindings patch that APPLIED (the
+	// rename published the new revision) before a follow-up durable step
+	// failed; the error's data carries the applied canonical state.
+	ErrorKeybindingsPostRename ErrorInfo = "keybindingsPostRename"
 )
 
 type MutationOutcome string
--- a/appwire/keybindings.go
+++ b/appwire/keybindings.go
@@ -1,0 +1,193 @@
+package appwire
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+)
+
+const (
+	MethodEvenerSettingsKeybindingsGet   = "evener/settings/keybindings/get"
+	MethodEvenerSettingsKeybindingsPatch = "evener/settings/keybindings/patch"
+
+	NotifyEvenerSettingsKeybindingsChanged = "evener/settings/keybindings/changed"
+)
+
+// KeybindingsRule is one user override: a non-nil Chord rebinds the action, a
+// nil Chord (JSON null) unbinds it. The server stores rules verbatim; last
+// rule wins on conflict, resolved by the frontend.
+type KeybindingsRule struct {
+	Action string  `json:"action"`
+	Chord  *string `json:"chord"`
+}
+
+// KeybindingsConfig is the user-controlled part of the keybindings overrides:
+// the payload contract minus the store-owned revision.
+type KeybindingsConfig struct {
+	Version int               `json:"version"`
+	Rules   []KeybindingsRule `json:"rules"`
+}
+
+// KeybindingsOverrides is the canonical payload returned by the get method and
+// broadcast on change: {version, revision, rules}.
+type KeybindingsOverrides struct {
+	Version  int               `json:"version"`
+	Revision uint64            `json:"revision"`
+	Rules    []KeybindingsRule `json:"rules"`
+	// LoadError is set only by GET when the persisted state failed to load:
+	// Rules then carries the shipped-default fallback (what is actually in
+	// effect) while the diagnostic tells the client to keep editing read-only
+	// - PATCH rejects in this state, so a client that treated the fallback as
+	// writable would fail every save (roborev PR #884 round 6). Absent from
+	// broadcasts and patch responses, which only exist post-load.
+	LoadError string `json:"loadError,omitempty"`
+}
+
+// KeybindingsGetResponse is kept as an operation-oriented name for callers;
+// the wire result's underlying named type is KeybindingsOverrides.
+type KeybindingsGetResponse = KeybindingsOverrides
+
+type KeybindingsPatchParams struct {
+	ExpectedRevision uint64            `json:"expectedRevision"`
+	Config           KeybindingsConfig `json:"config"`
+}
+
+type KeybindingsPatchResponse = KeybindingsOverrides
+
+type KeybindingsChangedParams = KeybindingsOverrides
+
+type KeybindingsChangedNotification = KeybindingsOverrides
+
+type KeybindingsConflictData struct {
+	EvenerErrorInfo ErrorInfo            `json:"evenerErrorInfo"`
+	Current         KeybindingsOverrides `json:"current"`
+}
+
+// KeybindingsPostRenameData is the WireError.Data payload for
+// ErrorKeybindingsPostRename: the patch APPLIED (Applied carries the
+// canonical applied state) but a follow-up durable step failed. Clients
+// should reconcile from Applied instead of treating the write as rejected.
+type KeybindingsPostRenameData struct {
+	EvenerErrorInfo ErrorInfo            `json:"evenerErrorInfo"`
+	Applied         KeybindingsOverrides `json:"applied"`
+}
+
+const keybindingsConfigVersion = 1
+
+// KeybindingsShippedDefaults is the payload when nothing is persisted: empty
+// overrides (all default bindings) at revision 0.
+func KeybindingsShippedDefaults() KeybindingsOverrides {
+	return KeybindingsOverrides{Version: keybindingsConfigVersion, Rules: []KeybindingsRule{}}
+}
+
+// ValidateKeybindingsConfig checks the structural contract only: the version
+// and rule-field sanity. Semantic validation (action ids, reserved chords,
+// conflicts) is the frontend's job; the server does not know the action
+// registry.
+func ValidateKeybindingsConfig(config KeybindingsConfig) error {
+	if config.Version != keybindingsConfigVersion {
+		return fmt.Errorf("unsupported keybindings config version %d", config.Version)
+	}
+	for i, rule := range config.Rules {
+		if strings.TrimSpace(rule.Action) == "" {
+			return fmt.Errorf("rule %d: action must be a non-empty string", i)
+		}
+		if rule.Chord != nil && strings.TrimSpace(*rule.Chord) == "" {
+			return fmt.Errorf("rule %d (%s): chord must be null or a non-empty string", i, rule.Action)
+		}
+	}
+	return nil
+}
+
+func DecodeKeybindingsPatchParams(raw json.RawMessage) (KeybindingsPatchParams, error) {
+	var params KeybindingsPatchParams
+	if err := decodeKeybindingsStrictJSON(raw, &params); err != nil {
+		return KeybindingsPatchParams{}, err
+	}
+
+	top, err := keybindingsObjectFields(raw)
+	if err != nil {
+		return KeybindingsPatchParams{}, err
+	}
+	if err := requireKeybindingsFields(top, "expectedRevision", "config"); err != nil {
+		return KeybindingsPatchParams{}, err
+	}
+	if string(bytes.TrimSpace(top["expectedRevision"])) == "null" {
+		return KeybindingsPatchParams{}, errors.New("expectedRevision must be an unsigned integer")
+	}
+
+	configFields, err := keybindingsObjectFields(top["config"])
+	if err != nil {
+		return KeybindingsPatchParams{}, fmt.Errorf("config: %w", err)
+	}
+	if err := requireKeybindingsFields(configFields, "version", "rules"); err != nil {
+		return KeybindingsPatchParams{}, fmt.Errorf("config: %w", err)
+	}
+	if string(bytes.TrimSpace(configFields["version"])) == "null" {
+		return KeybindingsPatchParams{}, errors.New("config: version must be an integer")
+	}
+	if string(bytes.TrimSpace(configFields["rules"])) == "null" {
+		return KeybindingsPatchParams{}, errors.New("config: rules must be an array")
+	}
+
+	var rules []json.RawMessage
+	if err := json.Unmarshal(configFields["rules"], &rules); err != nil {
+		return KeybindingsPatchParams{}, fmt.Errorf("config: rules: %w", err)
+	}
+	for i, rawRule := range rules {
+		ruleFields, err := keybindingsObjectFields(rawRule)
+		if err != nil {
+			return KeybindingsPatchParams{}, fmt.Errorf("rule %d: %w", i, err)
+		}
+		if err := requireKeybindingsFields(ruleFields, "action", "chord"); err != nil {
+			return KeybindingsPatchParams{}, fmt.Errorf("rule %d: %w", i, err)
+		}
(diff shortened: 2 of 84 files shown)


obra:
roborev findings on e2060a8 addressed in the force-push to 70a86de (fixes folded into their phase commits, branch still 5 commits):

1. **AskDock keyboard bypass** — handlePrimaryAction now honors the same sendDisabled state as the footer Send button, so Mod+Enter on the last question of an unfinished walk can no longer submit with the visible question implicitly skipped. Pinning test added (AskDock.test.tsx).
2. **Post-rename broadcast gap** — the store now wraps post-rename durable failures in an exported KeybindingsPostRenameError (the patch APPLIED case), and the RPC handler broadcasts the canonical snapshot before surfacing the error, so other clients never sit on the pre-patch revision. Store- and RPC-level tests added.

Full gates (make lint, make vet, make test — all 8 modules) pass on 70a86de.

Out-of-scope observation: the transcript-display store/handler pair has the same latent post-rename pattern; left untouched here.

obra:
Round-2 findings on 70a86de addressed in the force-push to 75f18bb (fixes folded into their phase commits; branch is still 5 commits; make lint + vet + test all green):

1. **AskDock tab strip eating modified arrows** (Medium) — handleTabsKeyDown now bails on any modifier, so Alt+Arrow/Alt+Home/End reach the global transcript chords from a tab button. Pinning test added.
2. **Post-rename error stranded the requester in hubError** (Medium) — the wire error now carries the applied canonical state (ErrorKeybindingsPostRename + KeybindingsPostRenameData.applied, the conflict-shape precedent); the frontend decodes it and treats the write as applied instead of disabling editing over live bindings. Go RPC test + frontend store test added.
3. **Consumed dock chords lacked stopPropagation** (Medium) — both AskDock key handlers now stopPropagation on consumption, so a binding that ignores defaultPrevented (rail.toggle) remapped onto Mod+Enter/Alt+Page* cannot double-fire. Two propagation tests added.
4. **Alt+Enter leaked into the bare-Enter path** (Low) — the free-text Enter chord now requires every modifier clear. Pinning test added.
5. **HoldHints first-composer ambiguity** (Low) — Composer stamps its session ref on data-composer and the chip anchors to the FOCUSED pane's composer (mirroring composer.focus's own target choice), falling back to the first. Split-workspace test added.
6. **Nil rules persisted as rules:null, rejected at load** (Low) — the hub store normalizes nil rules to the empty slice before persisting. Round-trip test added.

**Not a defect (repeat of a #874 finding):** the "+ key serializes as Shift++" claim is a false positive. tinykeys' parser treats a trailing + as a KEY, not a separator, and the round-trip is pinned end-to-end through the REAL dispatcher: keybindings.test.tsx:829 ("capturing + saves a chord that round-trips and fires on a real + keydown") captures +, saves it, and dispatches a genuine + keydown that fires the action. See the same answer with the tinykeys parse evidence on #874.

obra:
Round-3 findings on 75f18bb addressed in the force-push to c1b575c (folded into phase commits; make lint + vet + test green):

1. **ActivityTree / RailResizeHandle swallowing global Alt chords** (Medium) — both handlers now bail on Alt/Ctrl/Meta instead of preventDefault-ing every Arrow/Home/End regardless of modifiers, so session cycling and transcript scroll/jump fire while those controls have focus. The tree keeps its Shift behavior; the resize handle keeps Shift as the coarse-step modifier. Tests pin no-move + no-preventDefault for modified keys in both.
2. **Duplicate React keys for repeated presses (g g)** (Low) — the three multi-press render sites (settings section, cheatsheet overlay, hold hints) now key each press by an index:chord composite with a justified biome-ignore (position is identity for content-identical presses; CommandPalette precedent).

obra:
Round-4 finding on c1b575c addressed in the force-push to 697f46d (folded into the 4a commit; make lint + vet + test green):

- **Cheatsheet open state survived unmount** (Low) — the mount effect's cleanup now closes the singleton store, so the mobile→desktop breakpoint round-trip no longer remounts the overlay already open. Pinning test: open → unmount → store closed → remount → no dialog.

obra:
Round-5 finding on 697f46d addressed in the force-push to 669bc50 (folded into the p3 commit; make lint + vet + test green):

- **Shift+Arrow tab navigation regression** (Medium) — correct catch: the round-2 guard over-reached by including Shift, which no global binding claims. The tab-strip guard is now Alt/Ctrl/Meta only, matching the round-3 ActivityTree/RailResizeHandle pattern. A new regression test pins that Shift+Arrow still moves tabs.

obra:
Round-6 finding on 669bc50 addressed in the force-push to 2db3650 (folded into the 2b commit; make lint + vet + test green):

- **Malformed persisted state read as writable** (Medium) — GET now carries the load diagnostic on the payload (loadError, GET-only, omitempty; generated TS type + protocol doc regenerated), and the frontend maps it onto hubError so the editing gate stays closed with the reason showing instead of flapping writable→fail→refresh→writable. The fallback-defaults GET contract itself is unchanged (it is what lets the section truthfully show the in-effect defaults). Go test extended (GET asserts the diagnostic); frontend store test pins fallback-applies + hubError-set + no flap on re-refresh.

obra:
Round-7 finding on 2db3650 addressed in the force-push to ed223a699 (folded into the 2a/p3/4b commits; make lint + vet + test green):

- **AskDock Enter handler missed the IME keyCode 229 fallback** (Medium) — the dispatcher's guard is now the exported isIMECompositionKeydown (isComposing || keyCode === 229), shared by the dispatcher, the settings capture editor (which keeps its local Process/Unidentified exclusion), and both AskDock handlers. A commit-confirm Enter can no longer submit or advance a batch. New pinning test fires a 229-only Enter on a dock control and asserts no send.

obra:
Round-8 findings on ed223a6 addressed in the force-push to 1d43441e9 (folded into the 2b/4b commits; make lint + vet + test green):

1. **Key/code alias shadowing** (Medium) — verified against tinykeys' matching source (string keys match event.key case-insensitively OR event.code exactly), then fixed: a CODE_NAME_TO_KEY fold (KeyA-Z, Digit0-9, punctuation codes; Numpad* deliberately excluded as physically distinct keys) behind an exported keyComparisonIdentity, now used by both chordsOverlap and the reserved-chord check. Code-spelled and key-spelled chords for one physical key now collide at validation the same way they collide at dispatch, and a code-spelled Control+KeyW no longer slips past the reserved list. Tests pin both directions and the numpad non-fold.
2. **Settings tab title ignored the remembered section** (Low) — the pane registry's title() now resolves a bare /settings through the same remembered-section fallback the content uses. Test pins remembered/explicit/stale-id cases.

The web-job failure on ed223a6 was a pre-existing flake, not this diff: the identical spawnguard assertion failed on main's own CI the same day at a different viewport, the harness touches nothing this PR changes, it passes locally on this branch at all five viewports, and the re-run of the failed job went green.

obra:
Round-9 finding on 1d43441e addressed in the force-push to 6ea19c999 (folded into the 2a/2b commits; make lint + vet + test green):

- **Numpad alias containment** (Medium) — correct, and it faulted the round-8 fix: a NumLock-on numpad event is {key:"7", code:"Numpad7"}, so a plain "Control+7" binding and a "Control+Numpad7" binding both match it. The overlap predicate now models tinykeys' match as domain intersection (keysShareAnEvent): same-physical-key aliases still fold (KeyW/W, Slash//, Digit7/7), and a code-name spelling additionally overlaps its physical key's plain spelling through the shared event - while Digit7 and Numpad7 still correctly never conflict (disjoint code domains). The reserved-chord check keeps the narrower fold (browser-reserved chords name top-row keys). Tests: containment both directions, the numpad/top-row non-conflict, Enter/NumpadEnter, and a dispatch-level pin of the tinykeys behavior being modeled (code-named binding fires only on its physical key; the plain spelling fires on both rows and shadows by registration order).

obra:
Round-10 finding on 6ea19c9 addressed in the force-push to e76544e24 (folded into the p3 commit; make lint + vet + test green):

- **Mid-send batches kept their answer controls editable** (Medium) - three layers: AskQuestionCard takes a disabled prop threaded from batch.sending and applies it to every editing control (options, Something else, the note/free-text input); askDockStore's setAnswer/setNote refuse the same writes at the store boundary; and the batch jump's focusBatchEntry skips disabled controls, landing on the batch container (focusable only while sending) instead of a dead radio. Tests pin disabled-mid-send vs editable-open-batch, the store refusal, and the new jump landing.

obra:
Round-11 findings on e76544e addressed in the force-push to 5994466d3 (folded into the 4b commit; make lint + vet + test green):

1. **Reconnect revision baseline** (Medium) - real: invalidateReadyGeneration cleared loaded but retained the previous revision, so a reconnect to a RESTARTED hub (fresh state file, lower revision) had its authoritative refresh rejected by the stale guard on every refresh - old shortcuts live, editing silently disabled. The revision baseline now resets with the generation, matching the support-loss and rewire paths (finding 34). The test drives the real reconnect seam (FakeClient.emitStateChange idle→ready fires onReady like AppwireClient's reconnect) and is mutation-checked: against the unfixed store it stalls with exactly the reported symptom.
2. **The '+' key capture** (Medium) - not a defect; this is the same finding as rounds 2 and 4's repeats. tinykeys' parser treats a trailing + as the KEY (its own test suite pins Shift++ parsing), and the round trip is pinned end-to-end here through the REAL dispatcher: keybindings.test.tsx:829 captures +, saves Shift++, and a genuine + keydown fires the action. Restated with the same evidence on #860 and #874.

obra:
Round-12 findings on 5994466 addressed in the force-push to 903ec48f0 (folded into the 2b commit; make lint + vet + test green):

Both findings targeted chordsOverlap's documented scope limits; both hold up once you read tinykeys' dispatch loop (createKeybindingsHandler in the vendored dist):

1. **Multi-press shadowing** (Medium) - tinykeys advances every binding whose next press matches, and a COMPLETING binding does not fire while another binding holds a pending partial (the loop's conflicts list). So a single-press binding beside a sequence sharing its first press never fires for that press - serialization equality missed it entirely. chordsOverlap now does per-press prefix analysis: sequences conflict when every shared position overlaps; a mismatch at any shared position ends any shared stream. One existing test's expectation changed - it encoded the buggy serialization-equality behavior.
2. **Regex vs literal** (Medium) - a regex chord like (K|X) overlaps literal K because tinykeys tests regexes against event.key AND event.code. Literal-vs-regex now tests the regex against every event spelling the literal can match (key value plus canonical code names, top-row and numpad). Regex-vs-regex stays identical-source-only (no intersection solver - documented).

The code maps were restructured around one canonical KeyboardEvent.code -> key table with derived views (the round-8/9 pair had grown ad hoc).

obra:
Round-13 finding on 903ec48 addressed in the force-push to 5fd809342 (folded into the p3 commit; make lint + vet + test green):

- **Post-send focus could land on a mid-send batch** (Medium) - with several sends in flight, remaining[0] can be sending (disabled controls) while an editable batch waits behind it. Focus now lands on the first non-sending remaining batch, falling back to the first only when all are in flight. The test builds the two-sending-plus-open scenario deterministically by stubbing threadsStore.send with parked promises, and is mutation-checked (lands on the wrong batch without the fix).

obra:
Round-14 findings on 5fd8093 addressed in the force-push to 68fa4a29b (folded into the 2b commit; make lint + vet + test green). Note this head also rebases onto current main (the entity-deltas rework #846 landed mid-round; one test fixture needed its navigation capability updated to mode v2 + readVersions:[2], matching main's own test pattern).

1. **Distinct regexes treated as non-overlapping** (Medium) - fail closed: two regex chords now always count as overlapping. Regex intersection is undecidable at this layer, a false non-overlap silently shadows one binding at dispatch, and only hand-authored hub payloads can carry regex chords (nothing shipped uses one; capture records literals), so the conservative block has no shipped cost.
2. **Regex-vs-code numpad mishandling / reverse-mapping false conflicts** (Medium) - the literal's regex-candidate list now depends on its spelling class: a code-spelled literal (Numpad7) matches only its own code's events (candidates: the code name and its key value), while a key-spelled literal (7) matches every event carrying the key (candidates: key value plus every code name, top-row and numpad). (7) overlaps Numpad7; (Numpad7) does NOT conflict with Digit7. The code maps are now derived from one canonical code→key table.
3. **Regex chords bypassing the reserved list** (Medium) - Control+(W) matched the browser-reserved Ctrl+W event but never string-matched the canonical set. Validation now tests a single-press regex against each reserved chord's event spellings (key value + code names) with the same required-modifier comparison. Control+(W) is flagged; Control+(X) still passes.

PR #885, opened 2026-09-05
Improve session-start directory selection on desktop and mobile

obra:
The session-start directory chooser previously changed the launch directory while browsing, squeezed long paths into a small popover, and only offered directory creation when starting a session. This gives desktop and mobile an explicit directory selection flow with breadcrumbs, an editable path, recent directories, and inline folder creation.

Browsing and creation stay local until **Use this folder** confirms the selection; Cancel and Escape preserve the original launch directory. Desktop gives the directory its own row above the prompt. Mobile uses a full-height picker with confirmation above the keyboard, and long paths wrap instead of losing their final component.

Uses the existing directory APIs and shared modal focus scope. General file/path fields retain their existing behavior.

Validation:
- `make test-web` passed (unit tests, typecheck, and lint).
- `make test-web-browser` passed all five browser guards.
- Expanded spawnguard passed at 320, 390, 899, 900, and 1440 pixels, covering navigation, creation, confirmation visibility, and focus restoration.
- Picker tests cover draft selection, canonical paths, invalid paths, creation/retry, stalled or failed listings, stale responses, root selection, and focus after creation/cancellation. Spawn integration tests cover route changes while desktop and mobile pickers are open and catalog refresh only after confirmation.

Keyboard positioning uses a simulated CSS inset; physical mobile keyboard behavior was not tested.

### DIFF
--- a/cmd/evener-hub/frontend/scripts/spawnguard/run.mjs
+++ b/cmd/evener-hub/frontend/scripts/spawnguard/run.mjs
@@ -12,7 +12,7 @@ import { applyViewport, clearViewportOverride, connectPage, createStartupDeadlin
 import { describeBrowserStartupFailure, startBrowserGuard } from "../browserGuardProcess.mjs";
 
 const FRONTEND = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "..", "..");
-const WIDTHS = [390, 899, 900];
+const WIDTHS = [320, 390, 899, 900, 1440];
 // The staging cap (attachments/limits.ts MAX_ATTACHMENTS), so the row is
 // measured at the widest the product allows it to get.
 const STAGED_ATTACHMENTS = 8;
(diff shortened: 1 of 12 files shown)


obra:
The Windows-path finding is inapplicable to the supported hub platforms. `.goreleaser.yml` builds the hub for Linux and macOS, and `cmd/evener-hub/internal/hostlock/hostlock_other.go` rejects hub startup on other platforms. Paths in this picker refer to that hub's filesystem, regardless of the browser OS. The shared slash-based helpers therefore match the current support contract; a code comment now makes the assumption explicit. Adding Windows hub support would require work beyond this session-start UI PR.

The initial-path selection finding is valid and is fixed in ce004fee1 with first-focus selection and a regression test. Opening the dialog still avoids focusing the textbox automatically so browsing on a phone does not open the keyboard unnecessarily.

PR #903, opened 2026-09-06
Reclaim retained session scratch at process start

### LINKED ISSUE #893: sweepCrashedSessionScratch has no production caller, so retained session scratch is never reclaimed
## What

`sweepCrashedSessionScratch` (`agent/sandbox/session_scratch.go:156`) has no production caller; only tests invoke it. The scratch handoff convention rests on it: a session that hands a scratch off or closes *retains* it (lease released, directory kept) on the premise that a sweeper later reclaims retained directories whose owner is gone. Nothing runs that sweep.

## Effect

Every retained-but-orphaned session scratch directory is permanent on disk:

- parked worktree environments retained at close (`retainParkedWorktreeEnvironmentScratch`),
- the source side of every `AdoptSessionScratch` when the target already owned a scratch of that kind,
- scratch left by crashed sessions, which is the case the sweeper was written for.

## Fix shape

Wire the sweep into a real path (daemon or serve start, or a periodic tick), keyed on the existing lease check so a directory whose lease is still held is never touched. Test: a retained directory older than the window with no lease holder is removed; a leased one and a fresh one are left alone.

obra:
## What was wrong

`sweepCrashedSessionScratch` had no production caller — only tests invoked it. The whole scratch handoff convention rests on it: a session that closes or hands its scratch off *retains* it (lease released, directory kept), on the premise that a sweeper later reclaims the directories whose owner is gone. Nothing ran that sweep, so every parked-worktree retention, every losing side of an `AdoptSessionScratch`, and every crashed session's scratch was permanent on disk.

## What changed

- `sandbox.SweepCrashedSessionScratch(workspaceRoot)` (new, exported) sweeps every base a session may allocate from — the temp dir and the user cache dir, canonical and deduplicated, each skipped when it lies inside the workspace, exactly as allocation skips it. Allocation takes the first of those bases; the reclaim visits them all, because a workspace containing the temp dir sends its own sessions to the cache dir instead. The removal gate is unchanged: Evener prefix, older than the 24h window, **and** the lease acquirable. A lease still held means a live owner, and that directory is never touched.
- `sweepCrashedSessionScratch` now returns what it could not reclaim (unreadable base; a directory it owned but `RemoveAll` failed on) instead of discarding it. Contended leases and unreadable ages stay silent skips — they are not failures.
- `evener` and `evener serve` start one reclaim each on its own goroutine. `run` anchors it at `cfg.workDir` from the dispatch entry (`runWithScratchReclaim` in main.go); `serve` anchors it at its resolved `--dir` from inside `runServeWithDeps`, through a `serveDeps.reclaimScratch` hook that only the dispatch entry (`runServeWithScratchReclaim`) sets, so every test that calls `runServe` or `runServeWithDeps` leaves it nil and never sweeps a developer's real bases.
- Failures print as a `warning:` line on os.Stderr naming the directory and saying it will retry at the next start. Two processes starting together can race on one directory (one unlinks the lease file, the other recreates it, the first's rmdir fails); the loser reports and the next start reclaims it.
- No ticker. The sweep's contract does not ask for one, and one call per process start reclaims the same directories.
- Inherent to the retention design, not new here: a directory whose lease was released while a live process still writes to it (the losing side of an adopt, or a parked environment retained at close) becomes removable once it is older than the window — but only to a *second* evener process starting, since a process sweeps before it mints anything.

## How it is tested

`cmd/evener/scratch_reclaim_unix_test.go` drives `reclaimCrashedSessionScratch` over real bases (TMPDIR and a redirected `os.UserCacheDir`) holding real `sandbox.NewSessionScratch` directories:

- `...RemovesOnlyAbandonedScratch`: one retained past the window with no owner, one whose owner still holds its lease, one retained just now. Only the abandoned one may disappear, and the reclaim must report nothing.
- `...SkipsWorkspaceContainedTempBase`: TMPDIR inside the workspace pushes allocation onto the cache dir; the abandoned scratch there is reclaimed and the prefixed directory in the workspace's own temp dir is left alone.
- `...SweepsEveryAllocationBase`: abandoned scratch in *both* bases is reclaimed even though allocation for this workspace would only ever pick the temp dir, while a live lease in the cache base keeps its directory. Dropping the lease gate fails that case with `startup reclaim removed the leased scratch ".../Library/Caches/evener-sandbox-3256226104" from the cache base`.

Failures watched first: the initial wiring failed with `undefined: reclaimCrashedSessionScratch` — production had no entry point at all — and, with the wiring present but the sweep call neutered, `abandoned session scratch ".../evener-sandbox-974319947" survived startup reclaim: <nil>`. The multi-base test failed against the single-base version with `abandoned scratch ".../Library/Caches/evener-sandbox-1946824502" survived startup reclaim: <nil>`.

`TestSweepCrashedSessionScratchReportsUnusableBase` covers the new error return for an unreadable base and for a host with no usable scratch base.

## Round four (roborev on f2f5ebd)

- **Reclaim and allocation anchor the workspace the same way.** `execenv.SessionScratchWorkspaceRoot(dir)` (the structural git worktree root when there is one, else `dir`; it never forks git) is now the one function both `newSessionScratch` and the reclaim call, and both then canonicalise through the same Abs plus EvalSymlinks path. Before, `serve` reclaimed from the process cwd and ignored `--dir`, and a bare `run` passed a raw subdirectory while allocation anchored at the worktree root, so the workspace exclusion could pass the wrong tree and the sweep could remove stale `evener-sandbox-*` directories under a workspace-contained TMPDIR that allocation never uses.
- **Allocation short-circuits again.** `sessionScratchBase` takes the first valid candidate and touches nothing past it, so a slow or unavailable user cache directory cannot delay session start when the temp base serves; only `sessionScratchBases`, used by the reclaim, enumerates. Both share `validSessionScratchBase`. An empty candidate is now rejected rather than resolving to the cwd.
- Tests, each watched failing first: `TestReclaimCrashedSessionScratchAnchorsAtTheWorktreeRoot` (reclaim started from a subdirectory removed a base inside the worktree), `TestServeReclaimsScratchAnchoredAtItsDirFlag` (serve's reclaim removed a base inside its `--dir` workspace when anchored at the cwd), and `TestSessionScratchAllocationLeavesCacheBaseAloneWhenTempBaseWorks` (allocation reached for the user cache base while the temp base was usable; it pins the short-circuit by asserting the cache-dir lookup is never called).

## Gates

- `gofmt -l` on the changed files: clean
- `make vet`, `go vet -tags evenerfuzz ./...` (root + agent): clean
- `make lint-golangci`: PASS
- `go test -race ./agent/sandbox/... -count=1`: ok, 0 FAIL
- `go test -race ./cmd/evener/... -count=1`: ok, 0 FAIL
- `go test -tags evenerfuzz -run Fuzz ./agent/sandbox/ -count=1`: ok, 0 FAIL
- `go test ./agent/ -count=1 -timeout 40m` (`GOMAXPROCS=4`): ok, 0 FAIL
- `go test -race ./agent/execenv/... -count=1`: ok, 0 FAIL

Closes #893

### DIFF
--- a/agent/execenv/local.go
+++ b/agent/execenv/local.go
@@ -585,11 +585,21 @@ func (e *LocalExecutionEnvironment) unsandboxedScratchDir() string {
 // first .git entry — so the probe returned "" there anyway and the anchor fell
 // back to RootDir, exactly as it does now.
 func (e *LocalExecutionEnvironment) newSessionScratch() (*sandbox.SessionScratch, error) {
-	workspaceRoot := e.RootDir
-	if root, ok := structuralWorktreeRoot(e.RootDir); ok {
-		workspaceRoot = root
+	return sandbox.NewSessionScratch(e.sandboxTmpBase, SessionScratchWorkspaceRoot(e.RootDir))
+}
+
+// SessionScratchWorkspaceRoot reports the workspace a session started in dir
+// anchors its scratch to: the structural worktree root when dir sits in one, and
+// dir itself when it does not. Everything that decides whether a scratch base
+// lies inside the workspace has to ask this one function — the allocator above
+// and the startup reclaim in cmd/evener both do — or the two disagree about
+// which tree a directory belongs to and the reclaim sweeps a base allocation
+// refuses to use.
+func SessionScratchWorkspaceRoot(dir string) string {
+	if root, ok := structuralWorktreeRoot(dir); ok {
+		return root
 	}
-	return sandbox.NewSessionScratch(e.sandboxTmpBase, workspaceRoot)
+	return dir
 }
 
 // retainUnsandboxedScratch releases this env's unsandboxed per-session
--- a/agent/sandbox/session_scratch.go
+++ b/agent/sandbox/session_scratch.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"time"
 )
@@ -96,6 +97,20 @@ func canonicalScratchRoot(root string) (string, error) {
 }
 
 func sessionScratchBase(requested, workspaceRoot string) (string, error) {
+	bases := sessionScratchBases(requested, workspaceRoot)
+	if len(bases) == 0 {
+		return "", noSessionScratchBaseError(workspaceRoot)
+	}
+	return bases[0], nil
+}
+
+// sessionScratchBases lists, in allocation-preference order, every base a
+// session on this workspace may end up in: the requested base (or the temp dir)
+// and the user cache dir, each canonical, keeping only the ones that exist as a
+// directory outside the workspace. Allocation takes the first; a reclaim has to
+// visit them all, because a workspace that contains the temp dir sends its
+// sessions to the cache dir instead.
+func sessionScratchBases(requested, workspaceRoot string) []string {
 	first := requested
 	if strings.TrimSpace(first) == "" {
 		first = sessionScratchTempDir()
@@ -104,6 +119,7 @@ func sessionScratchBase(requested, workspaceRoot string) (string, error) {
 	if cache, err := sessionScratchUserCacheDir(); err == nil && strings.TrimSpace(cache) != "" {
 		candidates = append(candidates, cache)
 	}
+	var bases []string
 	for _, candidate := range candidates {
 		absolute, err := filepath.Abs(candidate)
 		if err != nil || pathWithin(absolute, workspaceRoot) {
@@ -117,9 +133,15 @@ func sessionScratchBase(requested, workspaceRoot string) (string, error) {
 		if err != nil || pathWithin(canonical, workspaceRoot) {
 			continue
 		}
-		return canonical, nil
+		if !slices.Contains(bases, canonical) {
+			bases = append(bases, canonical)
+		}
 	}
-	return "", fmt.Errorf("sandbox: no session scratch base outside workspace %q", workspaceRoot)
+	return bases
+}
+
+func noSessionScratchBaseError(workspaceRoot string) error {
+	return fmt.Errorf("sandbox: no session scratch base outside workspace %q", workspaceRoot)
 }
 
 func pathWithin(path, root string) bool {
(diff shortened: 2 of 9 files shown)


obra:
Fixed in d03a743.

`sessionScratchBases` now lists every validated candidate allocation may select — the requested base (or the temp dir), then the user cache dir — canonical and deduplicated, with the same workspace exclusion applied to each. Allocation still takes the first; `SweepCrashedSessionScratch(workspaceRoot)` visits them all. So a workspace containing TMPDIR gets the scratch it actually allocated (in the cache dir) reclaimed, and the `evener-sandbox-*` directories under that workspace-contained TMPDIR are left alone because that base is excluded for the reclaim exactly as it is for allocation.

The workspace root comes from the command: `run`'s `--work-dir` (falling back to the process working directory), and the process working directory for `serve`, which parses its own `--work-dir` inside `runServe`.

Watched failing first: `TestReclaimCrashedSessionScratchSweepsEveryAllocationBase` failed against the single-base version with `abandoned scratch ".../Library/Caches/evener-sandbox-1946824502" survived startup reclaim: <nil>`. `TestReclaimCrashedSessionScratchSkipsWorkspaceContainedTempBase` covers the exclusion half.

Same commit also moves the two calls out of `run`/`serve` to the production entry points that select them, so a test driving `run` or `serve` no longer sweeps the developer's real scratch bases, and a reclaim failure now names the directory and says it will retry at the next start.

Gates on d03a743: gofmt clean; `make vet` and `go vet -tags evenerfuzz` clean; `make lint-golangci` PASS; `go test -race ./agent/sandbox/... ./cmd/evener/...` 0 FAIL; `go test -tags evenerfuzz -run Fuzz ./agent/sandbox/` 0 FAIL; `go test ./agent/ -timeout 30m` 0 FAIL.

obra:
Both findings fixed in 866ee73.

**Workspace root now matches allocation's.** `execenv.SessionScratchWorkspaceRoot(dir)` is the single answer to "which workspace does a session in this directory belong to" — the structural git worktree root, `dir` itself when there is none — and the allocator (`newSessionScratch`) and the reclaim both call it, so the exclusion check can no longer be handed a subdirectory or the wrong tree. serve reclaims through a `reclaimScratch` dep that only the production entry point (`runServeWithScratchReclaim`) sets, called inside `runServeWithDeps` once `--dir` is resolved, so the daemon reclaims against the workspace it allocates in. Tests drive `runServe`/`runServeWithDeps`, whose deps leave that field nil, so no test path sweeps.

Watched failing first:
- `TestReclaimCrashedSessionScratchAnchorsAtTheWorktreeRoot` (TMPDIR at `<worktree>/tmp`, reclaim started from `<worktree>/sub`) failed against the raw-directory version: `reclaim from ".../001/sub" removed ".../001/tmp/evener-sandbox-1423612996", a base inside the worktree`.
- `TestServeReclaimsScratchAnchoredAtItsDirFlag` drives serve's real flag parsing with `--dir` from an unrelated cwd; reverting the call site to the process cwd fails it with `serve's reclaim removed ".../002/tmp/evener-sandbox-1376942148", a base inside its --dir workspace`.

**Allocation short-circuits again.** `sessionScratchBase` stops at the first base that serves and never reaches for the user cache dir behind it; only the reclaim enumerates both, through `sessionScratchBases`. The candidate validation both need is shared in `validSessionScratchBase` rather than written twice. `TestSessionScratchAllocationLeavesCacheBaseAloneWhenTempBaseWorks` failed first with `allocation reached for the user cache base while the temp base was usable`.

Gates on 866ee73 (0 FAIL each): gofmt clean; `make vet` and `go vet -tags evenerfuzz` (root + agent) clean; `make lint-golangci` PASS; `go test -race ./agent/sandbox/... ./agent/execenv/...`; `go test -race ./cmd/evener/...`; `go test -tags evenerfuzz -run Fuzz ./agent/sandbox/`; `go test ./agent/ -timeout 40m`.

PR #910, opened 2026-09-06
fix(plugins): walk doctor's orphan cache under the store lock

### LINKED ISSUE #899: Doctor's orphan-cache walk runs without the store lock, so a concurrent materialize or gc yields transient false orphan reports
## What

`Manager.Doctor` (`internal/plugins/doctor.go:115`) calls `doctorOrphanCacheDirs`, which walks `cache/<marketplace>/<plugin>/<sha>` and flags any directory not referenced by the current lockfile, without holding the store lock. `internal/plugins/paths.go:64` already notes the walk runs unlocked.

A concurrent `materialize` (a new sha directory that the lockfile does not name yet) or `Gc` (a directory being removed while the walk lists it) can make Doctor report a transient false orphan, or list a directory that is gone by the time the report prints.

## Severity

Low. Doctor is a read-only report and every other store mutation takes the lock (PR #869 put the root check at path derivation and lock acquisition; PR #872 gave the bundled cache its own lock). The only consequence is a misleading line in `evener-doctor` output during a concurrent install or gc.

## Fix shape

Take the store lock around the orphan walk with the same bounded wait Doctor's other checks use, or document the walk as best-effort in the finding text. Noted in the PR #869 review.

obra:
## What was wrong

`Manager.Doctor`'s orphaned-cache-directory check is the only finding it draws from two pieces of store state at once: the install paths `installed_plugins.json` names, and the directories under `cache/<marketplace>/<plugin>/<sha>`. Every writer that changes either one holds the store lock; the walk held nothing (`paths.go` said so out loud). A materialize creates its sha directory before it records the entry, and `Gc` drops the entry before it removes the directory, so a walk that lands in either window reports a live plugin as an orphan and sends the user to a `gc` that will not touch it.

## The choice, and why

**Take the lock**, not "mark the finding best-effort". Doctor had no locked checks at all, so there was no existing bounded wait to copy; the wait is a new judgement call. Two things decided it:

- Hedging the prose does not make the report right, and a hedge is not testable as behaviour — it is a string. Locking makes "has no registry entry" true rather than merely momentary.
- Locking alone would not have fixed it either. The walk has to re-read the registry *under* the lock: the snapshot Doctor's other checks share is from before the wait, so a materialize that finishes while the walk waits would still be missing from it. That is why `doctorOrphanCacheDirs` loads the registry itself instead of taking a `knownPaths` map from `Doctor`.

The wait is one second, not the thirty seconds every mutation takes. Doctor is a report somebody is sitting in front of, and install/upgrade/marketplace-refresh hold that lock across git fetches; queuing behind one is worse than saying so. A lock still held after a second becomes a `WARN` finding naming the contention, in the same `registry` category, with a remediation.

## Waiting on a lock is not a write

Doctor is documented read-only, so it takes the lock through `acquireExistingLock` rather than the writers' `acquireLock`. Two differences, one point:

- **No `O_CREATE`.** `acquireLock` creates the lock file it waits on, which would leave a `.lock` behind in a store that has none — a mutation of an existing store from the one verb that promises to make none.
- **`O_RDONLY`.** `flock` is granted on a descriptor opened either way, and a read-only holder still excludes a read-write one. Asking for write access only broke stores the caller may read and not write (root-owned, or a read-only mount), turning their orphan report into a lock complaint whose remediation would never come true.

When the lock file does not exist the walk runs unlocked with a no-op release: writers create their lock file before they touch anything else, so no lock file means no writer has ever locked this store. In a real store the case cannot arise at all, since the cache directory the walk reads was made by a writer that created the lock file first. The window that remains needs the lock file deleted out of band and a writer recreating it in the moment after the failed open, which leaves the walk exactly where every walk was before it took a lock at all.

The lock is only reached once the cache directory exists — a store with no cache has no orphans and nobody to wait for, so it is answered without going near the lock.

## What changed

- `internal/plugins/doctor.go` — `doctorOrphanCacheDirs` takes the store lock for `doctorOrphanLockWait` (1s) through `acquireStoreLock`, loads the registry under it, and reports a lock it cannot get as a `WARN` finding instead of walking anyway.
- `internal/plugins/locks.go` — `acquireExistingLock` (no `O_CREATE`, `O_RDONLY`, no-op release when the file is absent). The two acquires now share the retry loop as `flockUntil`; the writers' acquire is otherwise untouched.
- `internal/plugins/gc.go` — `referencedInstallPaths` extracted from `Gc` and shared with the walk. `Gc` removes what is not in that set and Doctor reports what is not in it; they have to agree, or the finding's remediation is a no-op. Note that the shared set counts every entry under a registry key, matching `Gc`, where Doctor's own per-plugin loop looks at `entries[0]`; writers only ever store one entry per key, so nothing diverges today.
- `internal/plugins/paths.go` — the note that the walk runs "with no lock at all" is no longer true.
- The `evenerfuzz` coverage harnesses follow the signature, stub `doctorAcquireLock` for the contention and registry-read branches, and cover `acquireExistingLock`'s two open outcomes.

## How it is tested

Four tests, each watched failing first:

- `TestDoctor_OrphanCacheDir_ReadsTheRegistryAMaterializeLeftBehind` — a real `flock` holder stands in for a materialize: the sha directory is on disk, the registry does not name it, and the entry lands just before the lock is released. Failed with `a materialize in flight was reported as an orphan: [{Level:WARN ... orphaned cache directory .../cache/acme/widget/deadbeef has no registry entry ...}]`.
- `TestDoctor_OrphanCacheDir_ReportsAHeldStoreLockInsteadOfGuessing` — the holder keeps the lock past the wait. Failed on `orphan claimed while a writer holds the store lock`, then `no finding contains "unreferenced cache directories"`.
- `TestDoctor_OrphanCacheDir_LeavesAStoreWithNoLockFileAlone` — snapshots the store's whole file list around the call. Failed on `Doctor created the store lock .../001/.lock (stat err = <nil>)` and on the tree diff showing the added `.lock`.
- `TestDoctor_OrphanCacheDir_WalksAStoreWithAReadOnlyLockFile` (unix, skipped as root) — a `0444` lock file. Failed on `no finding contains ".../cache/acme/widget/deadbeef"`, the report carrying `skipped the check for unreferenced cache directories: ... permission denied` instead.

All four use the real lock, not a stub. The existing `TestDoctor_OrphanCacheDir` and `TestDoctor_OrphanCacheDir_NoneWhenReferenced` still pass on a quiet store.

## Gates

- `gofmt -l` on the changed files: clean.
- `go vet ./...` and `go vet -tags evenerfuzz ./...`: clean.
- `make lint-golangci`: PASS. `make lint-evenerfuzz`: PASS.
- `go test -race ./internal/plugins/... -count=1`: ok, 0 FAIL.
- `go test -tags evenerfuzz -run Fuzz ./internal/plugins/ -count=1` (seed replay): ok, 0 FAIL.
- `go test ./cmd/evener-doctor/ -count=1` and the plugin CLI tests in `./cmd/evener/`: ok, 0 FAIL. `agent/` untouched.

Closes #899

### DIFF
--- a/internal/plugins/coverage_atomic_git_fuzz_unix_test.go
+++ b/internal/plugins/coverage_atomic_git_fuzz_unix_test.go
@@ -198,6 +198,19 @@ func coverageLocks(t *testing.T) {
 	if _, err := acquireLock(context.Background(), "x/lock", 0); err == nil {
 		t.Fatal("want open error")
 	}
+	// acquireExistingLock differs only in its open: a lock file that is not
+	// there is not an error, it is a store no writer has ever locked.
+	lockOpenFile = func(string, int, os.FileMode) (lockFile, error) { return nil, os.ErrNotExist }
+	missingRelease, mErr := acquireExistingLock(context.Background(), "x/lock", 0)
+	if mErr != nil {
+		t.Fatalf("absent lock file: %v", mErr)
+	}
+	missingRelease()
+	lockOpenFile = func(string, int, os.FileMode) (lockFile, error) { return nil, boom }
+	if _, err := acquireExistingLock(context.Background(), "x/lock", 0); err == nil {
+		t.Fatal("want existing-lock open error")
+	}
+
 	lockOpenFile = func(string, int, os.FileMode) (lockFile, error) { return &coverageLockFile{}, nil }
 	lockFlock = func(int, int) error { return boom }
 	if _, err := acquireLock(context.Background(), "x/lock", 0); err == nil {
--- a/internal/plugins/coverage_doctor_gc_fuzz_test.go
+++ b/internal/plugins/coverage_doctor_gc_fuzz_test.go
@@ -41,11 +41,11 @@ func FuzzDoctorGCCoverage(f *testing.F) {
 func coverageDoctor(t *testing.T) {
 	origReadDir, origStat := doctorReadDir, doctorStat
 	origCreateTemp, origRemove := doctorCreateTemp, doctorRemove
-	origGitAvailable := doctorGitAvailable
+	origGitAvailable, origAcquireLock := doctorGitAvailable, doctorAcquireLock
 	t.Cleanup(func() {
 		doctorReadDir, doctorStat = origReadDir, origStat
 		doctorCreateTemp, doctorRemove = origCreateTemp, origRemove
-		doctorGitAvailable = origGitAvailable
+		doctorGitAvailable, doctorAcquireLock = origGitAvailable, origAcquireLock
 	})
 
 	root := t.TempDir()
@@ -59,6 +59,26 @@ func coverageDoctor(t *testing.T) {
 	}
 
 	boom := errors.New("boom")
+	// The walk reaches the store lock only once there is a cache directory to
+	// walk, and everything past the lock is stubbed from here on.
+	if err := os.MkdirAll(m.cacheDir(), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	doctorAcquireLock = func(context.Context, string, time.Duration) (func(), error) { return nil, boom }
+	if got := m.doctorOrphanCacheDirs(); len(got) != 1 || got[0].Level != LevelWarn {
+		t.Fatalf("held lock findings = %#v", got)
+	}
+	doctorAcquireLock = func(context.Context, string, time.Duration) (func(), error) { return func() {}, nil }
+	if err := os.WriteFile(m.registryPath(), []byte("{"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	if got := m.doctorOrphanCacheDirs(); len(got) != 1 || got[0].Level != LevelFail {
+		t.Fatalf("corrupt registry findings = %#v", got)
+	}
+	if err := SaveRegistry(m.registryPath(), reg); err != nil {
+		t.Fatal(err)
+	}
+
 	doctorReadDir = func(path string) ([]os.DirEntry, error) {
 		switch filepath.Base(path) {
 		case "cache":
@@ -69,7 +89,7 @@ func coverageDoctor(t *testing.T) {
 			return nil, boom
 		}
 	}
-	if got := m.doctorOrphanCacheDirs(nil); len(got) != 0 {
+	if got := m.doctorOrphanCacheDirs(); len(got) != 0 {
 		t.Fatalf("market read error findings = %#v", got)
 	}
 
@@ -85,7 +105,7 @@ func coverageDoctor(t *testing.T) {
 			return nil, boom
 		}
 	}
-	if got := m.doctorOrphanCacheDirs(nil); len(got) != 0 {
+	if got := m.doctorOrphanCacheDirs(); len(got) != 0 {
 		t.Fatalf("plugin read error findings = %#v", got)
 	}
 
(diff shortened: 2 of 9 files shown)


obra:
Fixed in 5ee4f494aa32ed6de6cfc268a5111786b40c8b9b.

`acquireExistingLock` is `acquireLock` without the `O_CREATE` (and without the `lockMkdirAll`); the two now share the retry loop instead of each keeping a copy, and the writers' acquire is unchanged. When the lock file is absent the walk runs unlocked with a no-op release: the writers create their lock file before they touch anything else, so no lock file means no writer has ever locked this store and nothing can be materializing or collecting under the walk. In a real store the case cannot arise — the cache directory the walk reads was made by a writer that created the lock file first — so this only shows up in hand-built stores.

`TestDoctor_OrphanCacheDir_LeavesAStoreWithNoLockFileAlone` snapshots the store's whole file list around the call and asserts the lock path still does not exist, plus that the orphan is still reported. It failed first on `Doctor created the store lock .../001/.lock (stat err = <nil>)` and on the tree diff showing the added `.lock`. The held-lock and in-flight-materialize tests still exercise the locked path, since their writer creates the lock file for real.

obra:
Round 3 on top of the earlier reply. The finding above is against `8084b04`; it is addressed by `fd1825c7c` (non-creating acquire) and `903dfc5bf` (read-only open), both now rebased onto current main — branch head `903dfc5bfc76812f97f26ddd1fcd3fda88e8b230`.

`acquireExistingLock` opens the lock file `O_RDONLY` as well as without `O_CREATE`. Both differences make the same point: waiting on a lock is not a write. `flock` is granted on a descriptor opened either way and a read-only holder still excludes a read-write one, so demanding write access bought nothing and broke stores the caller may read but not write — root-owned, or a read-only mount — by turning their orphan report into a lock complaint with a remediation that could never come true.

`TestDoctor_OrphanCacheDir_WalksAStoreWithAReadOnlyLockFile` (unix, skipped as root) puts the lock file at `0444` and failed first on `no finding contains ".../cache/acme/widget/deadbeef"`, the report carrying `skipped the check for unreferenced cache directories: ... permission denied` in its place.

The "no lock file means no writer" comment now names the window it leaves: the lock file deleted out of band and a writer recreating it in the moment after the failed open, which is where every walk stood before it took a lock at all.

obra:
Declining both findings on 903dfc5, with evidence.

**Exclusive flock on a read-only descriptor.** The primitive here is `flock(2)` (`locks_unix.go`: `lockFlock = unix.Flock`, `LOCK_EX | LOCK_NB`), not `fcntl` record locks. The write-mode requirement is an `fcntl` `F_WRLCK` rule; `flock(2)` locks the open file description and is granted on a descriptor opened either way on Darwin and Linux. Verified on macOS during review: `LOCK_EX` on an `O_RDONLY` descriptor is granted, and it conflicts with a holder that opened `O_RDWR`. The branch's own tests exercise exactly this against an existing lock file: `TestDoctor_OrphanCacheDir_ReportsAHeldStoreLockInsteadOfGuessing` (a writer holds the lock, Doctor's read-only open must block and report) and `TestDoctor_OrphanCacheDir_WalksAStoreWithAReadOnlyLockFile` (0444 lock file, Doctor still walks). Both pass on Darwin locally and on the Linux CI runner.

**Missing lock file treated as unlocked.** A real store cannot present this case: the cache directory the walk reads was created by a writer, and every writer creates `.lock` before it touches anything else (`install.go`, `gc.go` take the store lock first; the bundled store writes only under `bundled/` with its own lock, outside the walked tree). Reaching the race needs the lock file deleted out of band and a writer recreating it in the instant after Doctor's failed open. That window leaves the walk exactly where every walk was before this PR took a lock at all, so it is no worse than main, and the alternative (creating the lock file) is the read-only violation the previous round fixed. The comment on `acquireExistingLock` states the window.

PR #912, opened 2026-09-06
test(execenv): await the signalled child instead of racing the grace

### LINKED ISSUE #896: Flake: TestCleanupRetainsOwnedTmpAfterChildGrace (execenv, seen twice) and the 5 s spawn cap in sandbox_lifecycle_test.go
## What

`TestCleanupRetainsOwnedTmpAfterChildGrace` (`agent/execenv/sandbox_lifecycle_test.go:228`) has failed twice under full-suite load (2026-09-03 and 2026-09-05, the second during a cold-build execenv run) and passed on every rerun and in isolation. It exercises the 2 s SIGTERM grace in `LocalExecutionEnvironment.Cleanup` (`agent/execenv/local.go` around 977 and 1037).

A forced-delay reproduction on 2026-09-04 did not reproduce it (30/30 under bounded `go test -p N` load). The next attempt should target the grace window itself: a child that ignores SIGTERM for longer than the grace on a slow runner.

## Second candidate in the same file

One `go test ./agent/execenv/...` FAIL on 2026-09-05 lost its name to a piped `tail -1`. The likeliest load-sensitive cap in the package is the 5 s wall-clock bound on real command spawns in `assertFileToolsShareTheShellScratch` (`sandbox_lifecycle_test.go:603`, used by three tests) and the similar cap near line 783. If it recurs, replace the cap with a labeled tripwire that awaits the process result, per the PR #852 pattern.

Load only via bounded `go test -p N`, never busy loops.

obra:
## What was wrong

**The grace race.** `TestCleanupRetainsOwnedTmpAfterChildGrace` has a real bash child record, in its SIGTERM trap, whether the owned session tmp still exists, and then reads the sentinel back. `Cleanup` escalates to SIGKILL a fixed grace after the signal — 200ms for the whole test binary, since `grace_test.go` shrinks `terminateGrace` package-wide. A runner that leaves the child unscheduled for that long kills it mid-trap, so the sentinel never lands and the test reports `Cleanup disposed the owned tmp before signaling the child` for a disposal that never happened. That is the message both CI failures carried. Nothing in the test awaited the child; it raced a production timer against the host scheduler.

**The spawn caps.** The readiness poll and three real command spawns in the same file were bounded at 5s of wall clock. A spawn that is merely slow under load then fails a test whose code is correct.

## What changed (test-only)

- `env.commandFactory` gets `awaitedTerminationFactory`: real shell runtimes whose `Terminate` signals as the system runtime does and then returns only once the child has exited. Cleanup's SIGKILL escalation now sits behind the child's own shutdown, so the sentinel reports on the ordering inside `Cleanup` and nothing else. This is the awaited completion the testing doctrine asks for (process exit), not a widened deadline.
- The test asserts the child's exit code before reading the sentinel, so a child that dies on a signal with its handler unfinished is reported as that instead of as a disposal.
- One named `processTripwire` (30s) replaces the readiness deadline and the three `5000`ms `ExecCommand` caps. `ExecCommand` already awaits the process result; the parameter is a ceiling, so it is now a tripwire that fires when a spawn wedges rather than a budget a loaded host can blow.

No production code changed.

## How it is tested

Forced interleavings, both directions:

| forcing | old | new |
| --- | --- | --- |
| `sleep 0.5` ahead of the trap body (child slower than the grace) | FAIL 3/3, `sentinel missing, stat err = ... no such file or directory` | PASS 5/5 |
| `sleep 6` ahead of each probe spawn | FAIL, all five affected tests, `shell write into the scratch: context deadline exceeded` / `ExecCommand: context deadline exceeded` | PASS |
| `WATCH` pointed at a directory that does not exist (the real regression) | n/a | FAIL on the disposal message — the assertion keeps its teeth |

## Gates

- `gofmt -l` on the changed file: clean.
- `go vet ./...` and `go vet -tags evenerfuzz ./...`: clean.
- `make lint-golangci`: PASS (107s).
- `go test -race ./agent/execenv/... -count=1`: ok, 0 FAIL (`/tmp/896-race-execenv.log`).
- `go test -tags evenerfuzz -run Fuzz ./agent/execenv/ -count=1`: ok, 0 FAIL.
- `go test ./agent/ -count=1 -timeout 30m`: `ok primeradiant.com/evener/agent 703.754s`, 0 FAIL (`/tmp/896-full.log`).
- Stability under real load: the four affected tests `-count=20` while the host sat at load average 136 (concurrent full-suite runs from other worktrees) — 20/20 pass.

Closes #896

### DIFF
--- a/agent/execenv/sandbox_lifecycle_test.go
+++ b/agent/execenv/sandbox_lifecycle_test.go
@@ -9,11 +9,18 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
 	"primeradiant.com/evener/agent/sandbox"
 )
+
+// TRIPWIRE: every real process this file spawns does milliseconds of work — a
+// shell writing one file, or a signalled child running a trap that stats one
+// directory. A ceiling this generous fires only when a process wedges, never
+// because the runner is loaded.
+const processTripwire = 30 * time.Second
 
 // TestEnableSandboxProvisionsSeatbeltBackend: EnableSandbox provisions the kernel
 // wrapper for whichever backend resolved — bubblewrap on Linux, sandbox-exec
@@ -236,6 +243,12 @@ func TestCleanupRetainsOwnedTmpAfterChildGrace(t *testing.T) {
 	}
 	env := NewLocalExecutionEnvironment(dir)
 	env.ownedSessionTmp = tmp
+	// Cleanup escalates to SIGKILL a fixed grace after the SIGTERM, so a child
+	// the host has not scheduled by then dies before its trap can record what it
+	// saw. Awaiting the exit inside Terminate holds the escalation behind the
+	// child's own shutdown, leaving the ordering in Cleanup as the only thing the
+	// sentinel below can be reporting on.
+	env.commandFactory = awaitedTerminationFactory{}
 	sentinel := filepath.Join(dir, "tmp-was-present")
 	ready := filepath.Join(dir, "trap-installed")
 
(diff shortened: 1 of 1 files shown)


obra:
Declining the coverage finding, with the evidence from the scoped review of this head.

The same production lines run in both versions of the test: the Terminate loop (`local.go:1034-1038`), the grace sleep (`:1041`), and the Kill loop (`:1042-1045`). The awaited factory overrides only `Terminate` and `Wait`, so `Cleanup`'s dispatch is untouched; what changes is that SIGKILL now always lands on a pid that has already exited.

The old test never asserted the escalation's killing effect either. Its trap ends in `exit 0`, so on every passing run the child was gone before the 200 ms test-binary grace elapsed; the escalation reaping a live child was the *failure* mode this PR removes, not something the assertion pinned. Escalation dispatch stays pinned by `TestCleanupUsesTrackedCommandRuntime` (`agent/execenv/command_runtime_test.go:111`, terminate=1 / kill=1), and a real SIGKILL reaping a TERM-ignoring child by `TestW2Conc_StreamCommandSignalEscalatesToSIGKILL` (`agent/execenv/cov_w2conc_streamcommand_test.go:20`). No coverage is lost; the test now proves the thing it is named for (the owned tmp survives the child's grace) without racing the scheduler.

PR #916, opened 2026-09-06
test(web): wait on transcript readiness, not on where the mount landed

### LINKED ISSUE #900: Flake: web-transcriptscrollguard 'transcript never settled into an overflow window' (seen once, CI run 33931195082)
## What

The web job's `web-transcriptscrollguard` browser guard failed once on PR #878 CI run 33931195082 with:

```
transcript never settled into an overflow window
```

The branch had no frontend change. The rerun passed. That is a harness settle timeout in the guard, so under runner load the transcript takes longer to reach its overflow window than the guard waits.

## Fix shape

Same pattern as the Go flake fixes in PR #852 and PR #873: wait on the real condition (the scroll container reporting overflow) under a named, generous tripwire instead of a fixed settle delay. If it recurs, capture the guard's timing output on the failing run first.

Same CI run also hit #821 (jobstore) once; recorded there.

obra:
## What was wrong

`transcriptscrollguard`'s mount settle spun until four things held together: the transcript overflowed, its geometry held still, no pill was showing, **and** it sat within 1px of the true bottom. That last clause is a product property the mount does not guarantee, so the spin could wait for something that was never going to arrive — and when the 15s ceiling fired it blamed the one condition that had been true the entire time.

Opening a session scrolls to the end from the virtualizer's estimates and measures once (`useTranscriptScroll`'s mount effect). Nothing re-anchors that landing without a scroll event or an item-count change. **A webfont that swaps in AFTER the landing supplies neither**: content grows below the fold, `scrollTop` stays put, and the reader is left a few pixels short with no pill.

Instrumented on a failing run:

| t (ms) | scrollHeight | scrollTop | gap | `document.fonts` |
|---|---|---|---|---|
| 1073 | 11466 | 10741 | 0 | loading |
| 1165 | 11487 | 10741 | **21** | loaded |

The 21px gap then held for all 901 remaining frames — 60fps the whole way, `overflow` true on every one of them — and the guard reported `transcript never settled into an overflow within 15s`.

## What changed

Readiness is now the real condition and nothing else: the scripted turns are in the model, the scroll container is mounted and overflows its port, and its geometry holds still for `SETTLE_QUIESCENT_FRAMES`. That decision moves into `src/dev/transcriptSettle.ts` as a pure frame-by-frame tracker returning either "ready" or a structured blocker, so both the verdict and the reason it is withheld are testable without a browser, and the tripwire message names the blocker instead of the ceiling.

Two knock-on effects worth calling out:

- The runner's `pill is visible at mount` assertion was **unreachable** under the old spin (readiness could only return with the pill hidden). It is live again.
- The 15s ceiling becomes a named `SETTLE_TRIPWIRE_MS`, bounded *above* by `browserGuardCdp`'s 30s `Runtime.evaluate` ceiling that the whole spin runs inside. A 90s value was tried and never fired: the transport timed out first and the blocker was lost (`timeout calling Runtime.evaluate after 30000ms`). The comment records that so the next person does not "generously" raise it back into silence.

Harness-only; no production frontend change.

## How it is tested

`src/dev/transcriptSettle.test.ts` (8 cases). The one that failed first is `a transcript parked a few pixels above the true bottom is still ready to drive`, red on `expected { kind: 'atBottom', bottomGap: 21 } to be null`. The negative case has a permanent home there too: `a transcript that never overflows never becomes ready, and the overflow is what is named`.

**Forced interleaving.** Holding the woff2 responses over CDP (`Fetch.enable` on `*.woff2`, released after 800ms) so the swap lands after the mount's scroll-to-end reproduced the CI failure verbatim, 3/3:

```
transcript harness: transcript never settled into an overflow within 15s
(turns=42/42, section=true, quiescentFrames=0, body children: DIV,SCRIPT)
```

With the new tracker the same delay passed 5/5, the jump still landing at `bottomGap 0`. Unthrottled, HEAD failed 4 of 10 runs on this host; the fix passed 10 of 10. The interception lever was removed before committing.

**Negative case re-verified end to end**, by temporarily shrinking the fixture to a single short turn:

```
transcript harness: the transcript never became ready to drive within 15000ms -
the transcript never overflowed its scroll port (scrollHeight 725, clientHeight 725,
needs more than 2900)
```

Settle cost measured for the tripwire's doc comment: 21 frames, 350ms unthrottled, 432ms under 10x and 730ms under 20x Chrome CPU throttling.

## Gates

The `web` job's exact commands, foreground, full logs kept: `make test-web` (PASS web-typecheck / web-test / web-lint, 0 FAIL), `make build-web` (exit 0), `make test-web-browser` (PASS on all five guards — layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard — 0 FAIL).

Closes #900

### DIFF
--- a/cmd/evener-hub/frontend/scripts/transcriptscrollguard/run.mjs
+++ b/cmd/evener-hub/frontend/scripts/transcriptscrollguard/run.mjs
@@ -109,9 +109,10 @@ async function main() {
       if (initial.errors.length > 0) failures.push(`page errors: ${initial.errors.join("; ")}`);
       if (initial.clientHeight <= 0) {
         failures.push(`scroll container has no height (clientHeight ${initial.clientHeight}) - the harness did not render`);
-      } else if (initial.scrollHeight <= initial.clientHeight * 4) {
+      } else if (initial.scrollHeight <= initial.overflowRequired) {
         failures.push(
-          `transcript did not overflow several times over (scrollHeight ${initial.scrollHeight}, clientHeight ${initial.clientHeight})`,
+          `transcript did not overflow several times over (scrollHeight ${initial.scrollHeight}, ` +
+            `clientHeight ${initial.clientHeight}, needs more than ${initial.overflowRequired})`,
         );
       }
       if (initial.turns !== 42) failures.push(`expected 42 scripted turns in the model, found ${initial.turns}`);
--- a/cmd/evener-hub/frontend/src/dev/transcriptSettle.test.ts
+++ b/cmd/evener-hub/frontend/src/dev/transcriptSettle.test.ts
@@ -1,0 +1,104 @@
+// @vitest-environment node
+
+import { expect, test } from "vitest";
+import {
+  createTranscriptSettleTracker,
+  SETTLE_OVERFLOW_FACTOR,
+  SETTLE_QUIESCENT_FRAMES,
+  type TranscriptGeometry,
+  type TranscriptSettleBlocker,
+  type TranscriptSettleSample,
+  type TranscriptSettleTracker,
+} from "./transcriptSettle";
+
+const TURNS = 42;
+const PORT = 725;
+
+/** A rendered, overflowing transcript sitting `bottomGap` px above the true bottom. */
+function rendered(bottomGap = 0): TranscriptGeometry {
+  const scrollHeight = 11_487;
+  return { scrollHeight, clientHeight: PORT, scrollTop: scrollHeight - PORT - bottomGap };
+}
+
+function sample(geometry: TranscriptGeometry | null, turns = TURNS): TranscriptSettleSample {
+  return { turns, geometry };
+}
+
+/** Feeds one sample repeatedly, returning the last verdict. */
+function hold(tracker: TranscriptSettleTracker, one: TranscriptSettleSample, frames: number) {
+  let blocker: TranscriptSettleBlocker | null = { kind: "unmounted" };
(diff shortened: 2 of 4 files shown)


obra:
@roborev-primeradiant Agreed, and fixed in `6a4a44bc4`.

You are right that the widening was not derived from anything. The readiness run this bounds measures ~1.1s of mount plus at most 730ms of quiescence at 20x CPU throttling — a worst measured case of about 2s — so 15s was already a wide multiple of it, and the extra 10s only bought a broken fixture ten more seconds of silence before it reported.

The margin argument is the sharper half. The spin runs inside one `Runtime.evaluate` call whose 30s ceiling is exactly what swallows the structured blocker when it wins (#919); at 25s only 5s separated the two. Back at 15s the spin reports its own blocker with that ceiling still 15s away.

`SETTLE_TRIPWIRE_MS` is 15s again and the constant's comment now carries the derivation — the measured ~2s worst case, the multiple, and the ceiling it must stay clear of — so the next reader can tell a measured widening from a hunch.

Re-verified against the non-overflowing fixture, which still fails and still names the real condition:

```
transcript harness: the transcript never became ready to drive within 15000ms -
the transcript never overflowed its scroll port (scrollHeight 725, clientHeight 725,
needs more than 2900)
```

Gates on this commit: `vitest run src/dev/transcriptSettle.test.ts` 9/9, `make test-web-browser` PASS on all five guards (0 FAIL), `biome ci src` clean. The PR body's quoted output was updated to match the restored value.

Issue #913, opened 2026-09-06
A delegating child that enters a worktree from its parent's shared environment leaks the clone's scratch at teardown


obra:
## What

A delegating child session that was spawned without worktree isolation keeps the full `manage_worktree` tool (`agent/session_init.go` ~1401 only strips it for worktree isolation, and ~1413 only for a zero delegation allowance). Such a child starts on its parent's shared environment object (`Session.ownsEnv == false`, recorded once at construction by PR #906) and can `enter` a worktree, which runs `swapEnvAndRefresh` (`agent/session_env_swap.go` ~132) and installs a clone of its own.

Nothing retains that clone's scratch at the child's teardown:

- `ownsEnv` is frozen at construction, so the child's teardown still treats its environment as shared and skips the scratch step.
- The abandoned-environment and parked-environment retains added by PR #878 run only under `if cleanupEnv` in `Session.close` (`agent/session_lifecycle.go` ~499-521), which a child teardown never sets.

So the clone's scratch lease and cached file-tool layers are held until the process exits. Pre-existing: on main before #906 the owner-side `subagent.ownsEnv` was equally frozen at construction, so this is not a regression, but the #906 PR body's framing that a child cannot swap environments is wrong.

## Fix shape

Either run the abandoned/parked retains for child teardowns too (they release leases without touching the process table, so they are safe under the #878 invariant), or refuse `manage_worktree` enter/exit for child sessions the way `resumeWorktreeReentry` already refuses re-entry for them. Test: spawn a delegating child without worktree isolation, have it enter a worktree and run a command that mints scratch, tear it down, assert no lease is held on the clone.

Found by the PR #906 scoped review.

https://claude.ai/code/session_01VAcwdjBpgzkg3emsf9QgWT

Issue #894, opened 2026-09-06
A session's own lock marker on a non-current managed worktree lane is never cleared (stranded lanes)


obra:
## What

Nothing automatic clears a session's own lock marker on a managed worktree lane that is not its current lane.

- `unlockOwnManagedWorktreeAtClose` (`agent/session_worktree_close.go:431`) reads only `worktreeCurrentPath` / `worktreeCurrentManaged`, so close unlocks the current lane and nothing else.
- The residue sweep is delegate-only and takes only unlocked lanes.
- Prune maps an own-session candidate to *skip*.

So a lane strands with the session's marker on it whenever the session dies while locked into a managed worktree, and on every re-entry refusal that returns before the lane is recorded as current. Only `manage_worktree remove`, or a later resume or switch that adopts the lane, clears it.

## Fix shape

Let close (or prune, after a grace) clear markers that carry a session id known to be dead. Test: a lane locked by a session id with no live session is unlocked by prune; a lane locked by a live session is left alone.

Related: #481 covers the delegate-lane variant (parent cannot unlock an unreachable delegate's lane).

Found during the PR #878 review rounds (environment ownership teardown), pre-existing.

https://claude.ai/code/session_01VAcwdjBpgzkg3emsf9QgWT

obra:
PR #918 covers the close-side half: at close a session now routes every managed lane of its project through the EvLeave rule, so its own marker is cleared on non-current lanes too. Remaining scope for this issue is the crash case only: a session killed while locked into a managed lane and never resumed or closed. No pid file, lease, or heartbeat is keyed by session id today, so a prune-side clearer would first need such a signal.

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

**Verdict:** `legit-current` — the claims verify, but the close-side half is already
fixed. `unlockOwnManagedWorktreeAtClose` now sweeps every managed lane of the project
through the `EvLeave` rule, not only `worktreeCurrentPath` (`agent/session_worktree_close.go:432-453`),
landed by `d5b946b9c4` ("release every own lane marker at session close", PR #918). The
residual is the crash case: `prune` maps an `OwnSession` candidate to `ActSkip`
(`agent/internal/worktree/lockstate.go:475-487`), and nothing is keyed by a session id
liveness signal (pid file/lease/heartbeat), so a session killed while locked into a
managed lane strands its marker until a later resume/switch adopts the lane or `remove`.

[...]

Issue #481, opened 2026-08-26
Parent cannot unlock an unreachable delegate's worktree lane


obra:
## Problem

When a delegate becomes unreachable (its model removed by the provider, a
crash, a hung job, or any reason `delegate_send` cannot revive it), the
parent often still needs to get into that delegate's isolation worktree to
recover uncommitted work. Today it cannot do that through `manage_worktree`.

A delegate's isolation lane is locked for the delegate's lifetime with the
`evener:dlg:<delegate-id>:<parent-session-id>` marker. The lock is the
delegate's, and the parent is the creator recorded in the marker. But the
only `manage_worktree` operation that releases that lock is `dispose`, which
also closes the delegate's resumability — a destructive retirement. There is
no `unlock` operation, and `switch` into a delegate-locked lane is hard
refused by the lock state machine.

## What happens today

Against a lane locked with the delegate's `evener:dlg:` marker (observed
state `OwnDelegate`), `Decide` in `agent/internal/worktree/lockstate.go`
returns:

- `switch` (`EvEnter`) → `ActRefuse` — "is locked (evener:dlg:...)". The
  parent cannot enter the lane. The spec's §9 Guards state this outright:
  "the parent cannot `switch` into an isolated delegate's worktree at all
  while the delegate exists." `switch` takes no `force` parameter.
- `dispose` (`EvDisposeUnchanged`/`EvDisposeChanged`) → `ActUnlock`. This is
  the only path that releases the lock. But it also calls
  `closeStableWorktreeResumability` with `stableWorktreeDisposalReason`,
  marking the delegate disposed/not-resumable permanently. On a *changed*
  lane it keeps the worktree (then `switch` would work), but the delegate
  can never be revived again. On an *unchanged* lane it removes the
  worktree outright.
- No `unlock` operation exists in the tool
  (`agent/session_tools_worktree.go` dispatches only `create`, `switch`,
  `adopt`, `exit`, `remove`, `list`, `prune`, `dispose`).

So when a delegate is unreachable and holds a dirty lane, the parent faces
three dead-ends:

1. `switch` into the lane → refused.
2. `unlock` the lane without retiring the delegate → no such operation.
3. `dispose` the lane → works, but retires the delegate (loses revival) and,
   on an unchanged lane, deletes the worktree.

The only escape is to bypass `manage_worktree` entirely and shell into the
lane directory directly (`cd` + raw git), which defeats the env swap, the
[...]

PR #918, opened 2026-09-06
Release every own lane marker at session close

obra:
## What was wrong

A managed worktree lane carries a `git worktree lock` marker naming the session that holds it (`evener:<sid>`). `unlockOwnManagedWorktreeAtClose` read only `worktreeCurrentPath` / `worktreeCurrentManaged`, so a clean close released the lane the session was sitting in and nothing else. Two situations leave a marker on a lane that is *not* current:

- the process dies while locked into a managed worktree, and
- `resumeWorktreeReentry` refuses to re-enter (child session, non-local env, out-of-project target, failed `worktree list`) and returns before the lane is recorded as current.

Nothing else clears such a marker: the delegate residue sweep is delegate-only and takes only unlocked lanes, and `prune` maps an own-session candidate to `ActSkip`. The lane stayed locked until `manage_worktree remove` or a later resume/switch adopted it.

## Which half of the issue this implements, and why

**The close-side half only.** The issue's leading candidate was prune clearing a marker whose session id is *known dead*. There is no reliable "this session id is dead" signal in the tree to consult: session ids are recorded in `SessionMeta` (which outlives the process), lock markers, and delegate descriptors, but nothing registers a live session id — no pid file, no session lease, no heartbeat. `agent/sandbox`'s scratch lease is a real flock-held liveness token, but its directory is `MkdirTemp`-random and carries no session id, so it cannot answer "is `<sid>` alive?". Rather than invent a heuristic (e.g. `SessionMeta.UpdatedAt` staleness, which an idle live session would fail), this PR takes the half that needs no liveness oracle at all: **at its own close, a session releases every managed lane carrying its own marker**, because by the time that close finishes the id in those markers names no live session by construction.

`prune` is untouched, so `#481` (a parent clearing an unreachable *delegate's* lane) is unaffected.

## What changed

`unlockOwnManagedWorktreeAtClose` now resolves the project's managed lane directory the same way `runLaneResidueSweep` and `prune` do, reads `git worktree list --porcelain` once, and routes every managed entry through the existing `leaveWorktreeWithLockState` — that is, the pure `EvLeave` rule. So:

- own session marker → unlock (current lane and stranded lanes alike);
- unlocked, another session's marker, or an `evener:dlg:` delegate marker → left untouched.

Subagents are skipped up front: a child never takes a session marker of its own (`applyInitInsideWorktreeLock` is root-only, `resumeWorktreeReentry` refuses children, and a delegate lane's lock belongs to the parent's §9 lifecycle), so the sweep would only cost a `git worktree list` fork per child close. This mirrors `disposeLaneResidueAtClose`'s existing scoping.

`docs/developing-evener/worktrees.md` gains one sentence under "Lock semantics" describing the widened close behaviour.

## How it is tested

`TestUnlockOwnManagedWorktreeAtClose_ClearsStrandedOwnMarker` (new) creates two lanes, re-locks the vacated one with this session's marker to stand in for a dead incarnation's residue, and asserts both are unlocked after close. It failed first with:

```
stranded own marker survived close: locked with "evener:034JxD6ILNMmSjw6B7Oy0I"
```

`TestUnlockOwnManagedWorktreeAtClose_LeavesForeignMarker` (new) pins the negative side: another session's marker survives close, reason unchanged.

`TestUnlockOwnManagedWorktreeAtClose_UnlockFailsWarns` (new) drives a git runner whose `worktree unlock` fails and asserts the lane stays locked with a warning naming it. `..._LeaveFailsWarns` is renamed `..._ListingFailsWarns` and now asserts on the lane directory, since a hidden git binary now fails the registry read rather than a per-lane unlock. All four are registered in `FuzzInitWorktreeSeed100`'s regression program list alongside the existing close-unlock entries.

## Gates

All foreground, full logs kept. The `agent` package suite exceeds a single 10-minute foreground window on this host, so it was run as name-range shards that partition every `Test`/`Example`/`Fuzz` name exactly once (`[AB] C D [EFGH] [IJKLM] [NOPQ] R S [TUV] [WXYZ] [^A-Z] ^Fuzz`).

- `gofmt -l` over the changed Go files: clean; `make lint-gofmt`: PASS.
- `go vet ./...` and `go vet -tags evenerfuzz ./...`: exit 0.
- `make lint-golangci`: PASS. `make lint-evenerfuzz`: PASS.
- `go test ./agent/ -count=1` (12 shards, `/tmp/894-full.log`): every shard exit 0, `grep -c FAIL` = 0.
- `go test -race ./agent/... -count=1` (subpackages + 12 shards, `/tmp/894-race.log`): every scope exit 0, `grep -c FAIL` = 0, `grep -c 'DATA RACE'` = 0.
- `go test -tags evenerfuzz -run Fuzz ./agent/ -count=1` (`/tmp/894-fuzz.log`): exit 0, `grep -c FAIL` = 0.

## Known limit

The pure crash case — a session killed inside a lane and **never resumed or closed again** — still strands its marker, because no close ever runs for it and no other actor can prove that session id is dead. That is the documented fail-safe in `docs/developing-evener/worktrees.md` ("a stale lock just means nobody has confirmed the work is abandoned yet"). Closing it properly needs a session-liveness signal that does not exist today; inventing one was out of scope here.

## Review round 2 (42069560f)

**The subagent skip was a regression, and is gone.** A delegating child keeps the full `manage_worktree` tool — `session_init.go` strips it only for a worktree-isolated child and for a zero delegation allowance — and `worktreeCreate` / `worktreeSwitch` lock with `worktree.FormatSessionMarker(s.id)` with no root-only check. So a child does take an `evener:<child-sid>` marker of its own (issue #913 documents exactly that child), and `teardownChildSession` calls the child's own close, which my skip turned into a no-op. The pass now runs for every session. `TestUnlockOwnManagedWorktreeAtClose_DelegatingChildReleasesOwnMarker` pins it: the child's marker is released and its parent's marker on another lane is untouched.

**The pass is bounded by the close cascade's deadline.** It previously ran the listing and every unlock on `context.Background()`, so a wedged git could hold shutdown for the 300 s per-command timeout across 1+N commands while the rest of the cascade shares a 30 s `LaneClosePassBudget`. It now derives its context from the cascade deadline, capped at `LaneClosePassBudget`. One deliberate wrinkle: **a cascade budget that is already spent is replaced with a fresh one** rather than short-circuiting the pass. Releasing our own locks is the same budget-exempt tail work `touchUnlockLaneTail` does; inheriting an exhausted deadline would leave every own lock held, which is precisely what this PR exists to prevent. `..._BoundedByCloseBudget` (the wedged runner cancels the cascade context and then blocks on the context the pass was handed — an unbound pass hangs) and `..._SpentBudgetStillReleases` pin both halves.

**The per-lane warning names the lane.** It wrapped `manage_worktree: unlocking the current worktree`, which is wrong for a lane the session is not in. The `EvLeave` decision plus unlock now live in `releaseOwnWorktreeLock`, which returns the raw failure so `leaveCurrentWorktree` can keep saying "the current worktree" while the close sweep names the individual lane.

**Delegate marker pinned.** `..._LeavesDelegateMarker` asserts an `evener:dlg:<dlg>:<this-sid>` lane survives the close — the lane belongs to the parent's §9 disposal lifecycle, not to this pass. All new tests are registered in `FuzzInitWorktreeSeed100`'s regression program list.

**The listing is not shared with `disposeLaneResidueAtClose`, deliberately.** That sweep is root-and-local-only, while the release pass now also runs for children, so on the child path there is no runner to share at all. Sharing on the root path would mean restructuring `runLaneResidueSweep` to hand its runner and lifetime back to the caller, for the sake of one memoized `worktree list --porcelain` fork per root close. Left as is.

Round-2 gates (all foreground, `GOMAXPROCS=4`, one process at a time): `go test ./agent/ -count=1 -timeout 40m` exit 0 / FAIL 0; `go test -race ./agent/` scoped to the close, teardown, subagent and delegate tests exit 0 / FAIL 0 / DATA RACE 0; `go test -tags evenerfuzz -run Fuzz ./agent/` exit 0 / FAIL 0; `go vet ./...` and `go vet -tags evenerfuzz ./...` exit 0; `make lint-golangci lint-gofmt lint-evenerfuzz` all PASS. The first full-suite run of this round caught `TestNoBareWallClockDeadlineInAgentTests` flagging a bare 200 ms bound in the new budget test; it was rewritten to await the real cancellation signal instead, matching `TestCloseBudget_ExhaustedMidPass_LanesLeftSafe`.

Part of #894: this is the close-side half. The crash case (a session killed while locked into a lane, never resumed or closed) stays open on the issue because no reliable dead-session signal exists to key an automatic clearer on.


obra:
Confirmed and fixed in 34bf929ccaaa6f0478d27b00c17abf9fd4eadc62.

The fallback only covered a cascade context that was already expired at entry. A cascade expiring during `git worktree list --porcelain`, or between the listing and one of the unlocks, cancelled the derived context mid-pass, so the pass warned and returned with this session's own markers still locked — the one guarantee it exists to give.

`unlockOwnManagedWorktreeAtClose` now derives its context from `context.Background()` capped at `laneCloseReleaseBudget()`, never from the cascade, so the guarantee no longer depends on how much of the cascade budget the steps before it consumed. That is the shape `touchUnlockLaneTail` already uses for its own-lock release (which runs unbounded); this one is capped, and the amplification bound is unchanged at cascade + (K children + 1) x `LaneClosePassBudget/2`. Nothing else used the `ctx` parameter, so it is gone again.

Tests: `TestUnlockOwnManagedWorktreeAtClose_CascadeExpiryMidPassStillReleases` cancels the cascade context from inside the intercepted listing, asserts the cancellation actually fired, and asserts every own marker is still released — it failed first with `own marker still held after the cascade budget expired mid-pass: "evener:034K0xXCCN46XXL9w1QveE"`. `..._BoundedByCloseBudget` now observes the deadline on the contexts the pass hands git, so it still fails on an unbounded pass (`the pass handed git a context with no deadline`) without waiting a budget out. It replaces `..._SpentBudgetStillReleases`, which could no longer hand the pass a spent context and so asserted nothing the plain close test did not.

obra:
Confirmed and fixed in d3fc3be0e74960a27181a7b0594d7bcead240ace.

You were right on both counts, and the cascade-expiry one was worse than it looked: that test's context was created locally and read only by the test's own guard, so with the pass taking no context parameter nothing in production could observe it. Deleting the interceptor's `cancel()` and the guard outright left it passing.

Its replacement drives a real `Close()` under a shortened cascade budget. One thing worth recording for anyone writing a similar test: driving an *expiring* cascade is not enough either. A pass that re-derives from an already-dead cascade notices at entry and starts over, which is exactly what the previous shape did — so a wedge that simply outlives the budget passes under both shapes. The discriminating case is a cascade still alive with less left on it than the release pass needs. The wedge therefore holds the residue sweep's registry read until three quarters of the cascade budget is gone (a wait derived from the deadline that sweep was handed, not a literal), and the release pass's own listing compares the two deadlines. Re-applying the previous shape makes it fail with `the release pass's deadline sits 0s past the cascade's`.

The comparison is between absolute deadlines, not the time left on each. An inherited context carries the cascade's own deadline instant, and two `time.Until` calls a few hundred nanoseconds apart on that one instant differ by just enough to read as "later" — which is how the first draft of this test passed under the very shape it was written to catch.

On the unlock stub: `..._BoundedByCloseBudget` now observes the deadline on every context the pass hands git and asserts it is present and no longer than the release budget, rather than blocking. A blocking stub would have to wait the budget out to prove anything, and the deadline assertion catches the same regressions — it fails on both "no deadline at all" and "the full cascade budget instead of the halved release budget" — in 0.12 s. It also counts commands separately from deadlines now, so an unbounded pass no longer additionally misreports "the pass never ran".

PR #878, opened 2026-09-04
fix(agent): one ownership-aware teardown for child sessions

obra:
## Invariant

A session runs `Cleanup` only on an environment whose process table it constructed, at its own close, and never on a child's. A child's environment is either the parent's own or a `WithWorkingDirectory` clone built for it, and a clone shares the parent's `runningPIDs` by pointer, so `Cleanup` on either one signals the parent's in-flight tools. A child's own processes end without it: its job manager stops its shells and cancellation ends its tool commands at close, and whatever survives is reaped when the table's owner closes. What a child owns outright is its clone's scratch, both kinds, and that is released together: retained on a handoff, disposed when the child is dropped. The invariant lives on `teardownChildSession` in `agent/subagents.go`, which every child close now takes (parent close, eviction, stable-controller reclamation, unadopted dispose, lane disposal), with `discardRestoredCandidate` making its environment decision through the same step.

## The shared process table

Kept shared, on evidence rather than split per clone: a session's own worktree switches swap `s.env` to clones and its close cleans only the current one, so a per-clone table would orphan every process started before a switch; `WithSandboxInvocationGrant` clones spawn one re-dispatched tool call whose process must die with the session; and a child never needs `Cleanup` for its own processes (job manager stop plus ctx cancellation at close). The rule "never Cleanup an environment you did not construct" holds once every child teardown routes through the helper.

## The hazards

1. **Unadopted-child close** (`disposeUnadoptedSubagentSession`): now a one-liner into `teardownChildSession(..., disposeChildScratch)`; the decision lives in one place.
2. **Terminal-subagent eviction** (`agent/subagents.go`): `ev.sess.Close()` became `teardownChildSession(..., retainChildScratch)`. No `Cleanup` on the shared parent env or a clone. Test failed first: eviction SIGTERM'd the parent's in-flight process and released its scratch lease.
3. **Parent teardown of owned children** (`agent/session_lifecycle.go`): `RetainSandboxScratch` only became `RetainSessionScratch` (new execenv method, both kinds; `Cleanup` uses it too) via the helper. Test failed first: an unsandboxed working-dir child's lease stayed held after parent close.
[...]

PR #921, opened 2026-09-06
Pin both Session.ownsEnv assignment sites

obra:
Test-only. Pins the two production assignments to `Session.ownsEnv` that #906 (https://github.com/prime-radiant-inc/evener/pull/906) introduced and left unpinned.

Both gaps were real. Every shared-environment test hand-builds its sessions and sets `ownsEnv` itself, so a wrong `true` out of `prepareSubagentRunFromSelection` (`agent/subagents.go:1011`) was invisible — that record is the whole gate on the teardown's environment step, so a wrong `true` is the process-table invariant #878 fixed. And the restore path's assignment (`agent/delegate_runtime.go:1711`) survived being forced `false`.

## The pins

**`TestSpawnedChildOnTheParentEnvironmentRecordsNoOwnership`** — spawns through `prepareSubagentRun` with no working dir and no box, asserts the child landed on the parent's own environment, records no ownership, and that its teardown leaves the parent's in-flight process and held scratch lease alone. Mutating the create site to `subSess.ownsEnv = true` fails it on two assertions:

```
a child spawned onto the parent's environment recorded that it owns it
the shared child's teardown released the parent's scratch /private/var/.../evener-sandbox-4071462944 lease while the parent is still working in it
```

The opposite direction at that site is already covered: mutating it to `false` fails `TestPrepareSubagentRun_PerDelegateSandboxCleansScratchOnGrantFailure` and `TestSpawnAgent_PerDelegateSandboxCleansScratchWhenLaunchIsRejected`.

**`TestRestoredDelegateRuntimeOwnsItsCloneScratch`** and **`TestDiscardedRestoreCandidateDisposesItsCloneScratch`** — restore a committed delegate through `restoreIdle`, assert the runtime is on a clone distinct from the root's environment, records ownership, and settles that clone's scratch in both directions (retained with the lease released on teardown, removed on a discard) while the root's process and scratch stay untouched. Mutating the restore site to `child.ownsEnv = false` fails both:

```
a restored delegate on a clone of its own recorded that it does not own it
the clone's scratch /private/var/.../evener-sandbox-172140758 lease is still held after the runtime's teardown
the discarded candidate retained its clone's scratch /private/var/.../evener-sandbox-1333427754: stat err = <nil>
```

## On the restore path's shared case

There is nothing to pin. A committed descriptor's `WorkingDir` is always `s.currentEnv().WorkingDirectory()` (`agent/delegate_runtime.go:1378`), never empty, and `restoreIdle` refuses any descriptor whose `WorkingDir` does not equal the rebuilt environment's cwd — so `prepareSubagentEnvironment` always re-roots a clone and `ownsFresh` is structurally `true` there. Confirmed empirically: forcing that assignment `true` leaves the scoped restore/scratch/environment/ownership suite green (55 s). The tests above assert the structural fact instead (the clone is not the root's environment).

Worth a look separately: the same delegate is shared-env at create (no worktree isolation, no box) but clone-env after a restart. Pre-existing, not touched here.

## Gates

- `gofmt -l` on the changed file: clean
- `GOMAXPROCS=4 go vet ./...` and `-tags evenerfuzz`: clean
- `GOMAXPROCS=4 make lint-golangci`: PASS
- `GOMAXPROCS=4 go test -race ./agent/ -run <new + neighbouring teardown tests> -count=1`: exit 0, `grep -c FAIL` = 0 (12 tests)
- `GOMAXPROCS=4 go test ./agent/ -count=1 -timeout 40m`: exit 0, `grep -c FAIL` = 0 (340 s)

### DIFF
--- a/agent/subagent_teardown_unix_test.go
+++ b/agent/subagent_teardown_unix_test.go
@@ -470,19 +470,35 @@ func TestSpawnedChildOnTheParentEnvironmentRecordsNoOwnership(t *testing.T) {
 	}
 }
 
-// A restored stable delegate never shares its root's environment: its committed
-// descriptor carries a working dir, so the restore re-roots a clone for it and
-// the clone owns whatever scratch it mints. The runtime has to record that, or
-// its teardown leaves the clone's lease held for the rest of the daemon's
-// uptime — and it must record it without reaching the root's environment, which
-// the clone shares a process table with while the root is mid-restore.
-func TestRestoredDelegateRuntimeOwnsItsCloneScratch(t *testing.T) {
+// restoredCloneRuntime is a committed delegate brought up through the
+// production restore path, with everything the ownership pins measure against
+// it: the runtime, the clone the restore re-rooted for it, that clone's own
+// scratch, and a root still working — an in-flight process and a held scratch
+// of its own — so a teardown that reaches the root's environment shows up.
+type restoredCloneRuntime struct {
+	runtime     *Session
+	scratch     string
+	rootScratch string
+	pid         int
+	done        <-chan error
+	release     func()
+}
+
+// restoreDelegateRuntimeOnAClone restores a root from a cold committed
+// delegate and hands its runtime back through restoreIdle. It fails the test
+// unless the runtime really did land on a clone of its own holding a scratch
+// no one else holds, since that is the premise every assertion about ownership
+// rests on. The root's close is registered FIRST so it runs LAST: the
+// committed lease has to be failed before the close, or closeRuntimeTree
+// drains a still-committed stop for its whole budget.
+func restoreDelegateRuntimeOnAClone(t *testing.T) restoredCloneRuntime {
+	t.Helper()
 	fixture := newColdStableDelegateFixture(t, "")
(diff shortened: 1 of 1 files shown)


obra:
Declining the High on cf9c88f: the tests do reach teardown, on both platforms.

The scoped review and the implementer both ran the three tests on Darwin with `-race -count=2`: all pass, and under each of three mutations they fail on the teardown-stage assertions (the clone's lease still held after teardown; the discarded candidate's scratch retained; the shared child's teardown releasing the parent's lease), which a test that failed before teardown could not produce. On this same head the Linux CI jobs `tests`, `race-root`, and `race-modules / agent` all completed with success.

The scratch the restore tests inspect is minted by the real command the fixture runs on the clone before teardown (the in-flight shell the helper starts), not by a Git snapshot; `SessionScratchDir()` reports that directory once it exists, which is why the assertions have a path to stat and a lease to probe. The follow-up commit 5b84caec3 only moved `root.Close()` into `t.Cleanup` so the lease is released before the tree closes, taking each restore test from fifteen seconds of dead wait to a tenth of a second, with identical assertions.

PR #924, opened 2026-09-06
fix(hub,registry): credential JSON onboarding, template-only add-form inputs, row-level vars_env resolution

obra:
Follow-ups from the PR #879 review (roborev rounds 18 and 19), split out so that PR could land.

## Connect provider dialog offers the credential JSON editor

The onboarding dialog (`ConnectProviderDialog`) only knew the `apiKey` editor, so a gcp-adc instance there had "Test connection" and nothing else; pasting a service-account key or `application_default_credentials.json` was only reachable from Settings → Providers & credentials. The dialog now offers **Set credential JSON** / **Replace credential JSON** for an instance whose `authModes` include `credentialJson`, reusing `CredentialJsonDialog`. A stored-value editor's kind doubles as the auth mode that offers it, so one lookup renders the editor and closes it when the mode disappears. Saving returns to the chooser and still requires an explicit successful connection test, like the API-key path.

Tests: `ConnectProviderDialog.test.tsx` — save-then-test flow for a credential JSON (asserts the request shape, the "Configured via stored credential JSON" label, the Replace label, and that `onConnected` waits for the test), and the editor closing when the instance stops offering the mode.

## Add-form inputs only for variables a URL template reads

models.dev's env list for `google-vertex` names `GOOGLE_APPLICATION_CREDENTIALS` beside the project and location, and the registry folds every entry into `Transport.VarsEnv` keyed by itself. The hub built the add form's variable inputs from that whole map, so the form rendered a credential-file input whose value was persisted under `vars` and never read: instance vars only feed `{PLACEHOLDER}` substitution, and the credential is read from the environment or the store. Pre-existing on main (the same list fed `varsEnv` before `vars` existed).

New `Registry.TemplateVarsEnv(id)`: the `vars_env` mappings (the provider's and its rows', since models.dev maps a per-model `api` template's placeholders on the row alone) restricted to placeholders some URL template reads (the provider's base URL and endpoint templates, or a row's own) or a host rule consumes (`hostRuleAuthorityVars`). That also surfaces `GOOGLE_VERTEX_ENDPOINT`, which only `google-vertex`'s OpenAI-compatible rows read and which the form never offered before. A provider with no curated base URL at all (`cloudflare-ai-gateway`, `watsonx`) keeps its whole map, since the user types the URL and a placeholder they type still resolves from instance vars. The hub builds both `ProviderDescriptor.Vars` and `VarsEnv` from it; the wire shape is unchanged. Across the embedded catalog only `google-vertex`, `google-vertex-anthropic`, and `privatemode-ai` (literal localhost base URL) lose an input.

Tests: `TestTemplateVarsEnv` (fixture `google-vertex`, `openai`, `watsonx`; a synthetic provider whose only reference is a row template; one whose only reference is a host-rule input; unknown id) and `TestInstances_ListOffersOnlyTemplateVars` (verified red against the unfiltered code). Docs: `appwire/types.go` descriptor comment and the two spec passages that describe the add form's inputs.

## Row-level `vars_env` now resolves from the environment

Found by roborev while reviewing the above: the variable lookup only ever read the provider-level `vars_env`, so a mapping that lives on a row alone (models.dev maps a per-model `api` template's placeholders on the row, which is where `google-vertex`'s OpenAI-compatible rows carry `GOOGLE_VERTEX_ENDPOINT`) never read the environment, and those rows' URLs stayed unexpanded unless the value was typed as an instance var. `varLookupWith` now takes the transport shape being resolved and consults its `vars_env`/`vars` for a name the provider does not map at all; `buildTransport` passes the merged row transport, head-level callers pass the provider's transport (no change), and the canonical first-party lookup passes its curated shape so both sides of that gate read the same mappings. A provider or instance mapping whose variable is unset does not fall back to the row's stock mapping: the URL stays unresolved with the warning, as spec §10 already rules for an unset user `vars` reference. A catalog-wide differential shows the first-party gate only ever tightens under this change, and only on the eight Vertex MaaS rows that main was sending to a URL containing the literal placeholder.

Tests: `TestResolve_RowMappedVarsEnvReadsTheEnvironment` (stock environment resolves the MaaS row; an instance `vars_env` remap wins; the same remap unset stays unresolved and warned). Spec §9.1 sentence updated.

Two smaller refinements from the later review rounds: the add-form scan takes a host rule's inputs from every transport it visits (a row or glob may carry its own `host_rule`), and `Resolved.Transport.Vars` is now exactly the variables the effective URL was built with, the row's own base URL template when it has one, plus a location the vertex-location rule derived the host from. The regional Vertex warning reads that, so it fires for a host-only template on a regional location and stays quiet for a literal override that reaches no regional endpoint. No curated provider's `Vars` change.

## Verification

`make vet`, `make lint`, `make test`, `make test-web`, `make fuzz-gap-check` green; scoped review of the branch with one fix round (the credential-JSON close rule is now pinned by a "does not reopen when restored" test that fails when the effect is narrowed, and the no-base-URL case is pinned by the `watsonx` row).

### DIFF
--- a/appwire/types.go
+++ b/appwire/types.go
@@ -2662,10 +2662,12 @@ type InstanceEntry struct {
 
 // ProviderDescriptor is a registry provider the add form can build on: its
 // id and display name, the protocol and auth scheme it defaults to, and the
-// variables its transport and credential read. Vars maps a template
-// placeholder name to the environment variable name it is fed by (the same
-// shape as registry.Transport.VarsEnv), so a typed override can be sent
-// keyed by the name the registry actually substitutes. VarsEnv is the same
+// variables its URL templates read. Vars maps a template placeholder name
+// to the environment variable name it is fed by, so a typed override can be
+// sent keyed by the name the registry actually substitutes. It is
+// registry.Transport.VarsEnv restricted to the placeholders some URL
+// template reads (Registry.TemplateVarsEnv): a vars_env entry nothing
+// substitutes, such as a credential's own variable, gets no input. VarsEnv is the same
 // environment-variable names alone, sorted. It stays a list because v3
 // peers — a TUI built before Vars existed — decode it as one, and
 // ProtocolVersion is compared exactly, so a wire shape cannot change under
--- a/appwire/types.go
+++ b/appwire/types.go
@@ -2666,9 +2666,10 @@ type InstanceEntry struct {
 // to the environment variable name it is fed by, so a typed override can be
 // sent keyed by the name the registry actually substitutes. It is
 // registry.Transport.VarsEnv restricted to the placeholders some URL
-// template reads (Registry.TemplateVarsEnv): a vars_env entry nothing
-// substitutes, such as a credential's own variable, gets no input. VarsEnv is the same
-// environment-variable names alone, sorted. It stays a list because v3
+// template reads or a host rule consumes (Registry.TemplateVarsEnv): a
+// vars_env entry nothing substitutes, such as a credential's own variable,
+// gets no input. VarsEnv is the same environment-variable names alone,
+// sorted. It stays a list because v3
 // peers — a TUI built before Vars existed — decode it as one, and
 // ProtocolVersion is compared exactly, so a wire shape cannot change under
 // v3; new readers use Vars.
--- a/cmd/evener-hub/app_instances.go
+++ b/cmd/evener-hub/app_instances.go
@@ -54,13 +54,17 @@ func (c *hubInstancesController) List() appwire.InstanceListResponse {
 			if !ok {
 				continue
 			}
+			// Only the variables a URL template reads become inputs; the
+			// rest of vars_env (a credential's own variable) has no
+			// instance-level meaning the add form could give it.
+			vars := r.TemplateVarsEnv(id)
 			providers = append(providers, appwire.ProviderDescriptor{
 				ID:        id,
 				Name:      p.Name,
 				Protocol:  p.Protocol,
 				Auth:      p.Transport.Auth,
-				VarsEnv:   slices.Sorted(maps.Values(p.Transport.VarsEnv)),
-				Vars:      maps.Clone(p.Transport.VarsEnv),
+				VarsEnv:   slices.Sorted(maps.Values(vars)),
+				Vars:      vars,
 				APIKeyEnv: append([]string(nil), p.APIKeyEnv...),
 				Implicit:  registry.BoolValue(p.Implicit),
 			})
--- a/cmd/evener-hub/app_instances_test.go
+++ b/cmd/evener-hub/app_instances_test.go
@@ -8,6 +8,7 @@ package hub
 
 import (
 	"errors"
+	"maps"
 	"os"
 	"path/filepath"
 	"slices"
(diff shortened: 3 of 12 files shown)


obra:
Round 1 finding, checked:

- **Medium — row-level `vars_env` mappings never reached the add form.** Confirmed for rows, and it predates this PR: the descriptor was always built from the provider-level map, and models.dev maps a per-model `api` template's placeholders on the row alone (`convertModel`), which is the only place `google-vertex`'s OpenAI-compatible rows carry `GOOGLE_VERTEX_ENDPOINT`. The resolver reads no row-level `vars_env` for the environment either, so an instance var is the hub's only way to set it. Fixed: `TemplateVarsEnv` now collects the mappings from the rows' transports as well as the provider's (provider entry wins a shared name), and `google-vertex` offers `GOOGLE_VERTEX_ENDPOINT` beside the project and location; pinned in `TestTemplateVarsEnv` (fixture `google-vertex` plus a synthetic row-only mapping) and the hub descriptor test. Top-level glob rows are deliberately not scanned: every one in the overlay carries capabilities only, no transport, and the doc comment records that exclusion so it is revisited if a glob ever grows one.

obra:
Round 2 findings, checked; all three taken:

- **Medium — a provider without a curated base URL returned before scanning its rows.** Confirmed (my early return from round 1's fix). `TemplateVarsEnv` now always scans the rows and, when the provider has no base URL of its own, treats every provider-level mapping as offered, so the rows' templates and mappings come through beside them. Pinned by a synthetic `typed-url` row in `TestTemplateVarsEnv`.
- **Medium — row-level `vars_env` was offered to the form but the resolver only read the provider-level map.** Confirmed, and pre-existing: `varLookupWith` read `rec.head.Transport.VarsEnv` alone, so `GOOGLE_VERTEX_ENDPOINT` from the environment never resolved a Vertex OpenAI-compatible row's URL (only an instance var did). The lookup now takes the transport shape being resolved and consults its `vars_env`/`vars` for a name the provider does not map at all; `buildTransport` passes the merged row transport, head-level callers pass the provider's transport (no change), and the canonical first-party lookup passes its curated shape so both sides of that gate read the same mappings (a differential over the catalog shows the gate only ever tightens: the eight Vertex MaaS rows, which main sent to a URL still containing the literal placeholder, now resolve and are judged non-first-party, fail-closed as designed). A provider mapping whose variable is unset does not fall back to the row's stock mapping, so an unset instance remap stays unresolved with the warning instead of silently reading the environment under the name the user redirected. Pinned by `TestResolve_RowMappedVarsEnvReadsTheEnvironment`: `google-vertex/meta/llama-3.3-70b-instruct-maas` resolves `https://$GOOGLE_VERTEX_ENDPOINT/v1/projects/…/endpoints/openapi` from the environment, a user instance whose own `vars_env` remaps the name wins, and the same instance with the remap unset stays unresolved.
- **Low — top-level glob rows may carry a transport.** Confirmed against the schema (`modelSchema` embeds `transportSchema`) and `applyGlobs`; none in the shipped overlay does, but the scan now includes a top-level glob's transport for every provider that has a row the glob matches, and no other. Pinned by `TestTemplateVarsEnvTopLevelGlob` with an overlay glob on `*llama-3.3-70b*`: offered for `google-vertex`, not for `openai`.

obra:
Round 3 findings, checked; both taken:

- **Medium — host-rule inputs were collected from the provider's transport only.** Confirmed against the schema (`transportSchema` carries `host_rule`, so a row or top-level glob transport may set one, and `transportShape` merges it). No shipped row does, but the scan now takes `hostRuleAuthorityVars` from every transport it visits; pinned by a synthetic row-level rule in `TestTemplateVarsEnv`.
- **Low — the regional Vertex warning looked the location up through the provider's transport.** Taken in the more direct form: the warning now reads `GOOGLE_VERTEX_LOCATION` from the `Vars` that `buildTransport` substituted into the URL, so the warning and the request cannot disagree about the location, whatever mapping supplied it. On the curated providers the two agree (google-vertex maps the location at provider level, and a provider mapping always wins), so the existing warning tests still hold. The one observable change is an improvement: an instance whose own literal `base_url` carries no location placeholder no longer gets a regional warning about a location its URL never used; pinned in `TestResolve_TransportAssembly`.

obra:
Round 4 finding, checked and taken:

- **Medium — the regional Vertex warning missed a host-rule-only URL and could fire for a row override that never used the location.** Both halves confirmed in `buildTransport`: the provider-level base URL template was always among the templates whose placeholders become `Transport.Vars`, even when a row's own base URL replaced it, and a template naming only `{GOOGLE_VERTEX_HOST}` derived the host from the location without recording it. Fixed at the source rather than at the warning: `Vars` is now built from the effective base URL template (the row's when it has one) plus the endpoint templates, and when the vertex-location rule derived the host, the location it derived it from is recorded too. So `Vars` is exactly the variables the URL was built with, and the warning reads that. Pinned in `TestResolve_TransportAssembly`: a provider-level `{GOOGLE_VERTEX_HOST}/v1/custom` override with a regional location warns and exposes the location; a row-level literal `base_url` override neither warns nor exposes it. No curated provider's `Vars` change (the Vertex templates name the location in their path; every other row override repeats its provider's placeholders).

obra:
Round 5 finding, checked and taken:

- **Low — the canonical first-party lookup ignored a row's curated `vars` default.** Confirmed as a consistency gap this PR's round-2 change opened: the actual resolution's lookup now consults the shape being resolved, so a row whose own base URL names an authority placeholder can take the value from its curated `vars`, while `canonicalVarLookup`'s defaults still came from the provider level alone; the sides then disagreed on curated data and `web_search` was stripped. `defaultVarLookup` now takes the shape being resolved and falls back to its defaults for a name the provider does not set. User-layer values stay excluded: the canonical shape is built from the provider's curated layers and its curated row (`canonicalRow`). Pinned by `TestResolve_WebSearchCanonicalGateRowDefault` (an overlay provider whose row sets its own `vars` default for `{ROW_HOST}` keeps `web_search`). No shipped row carries a `vars` default, so no curated resolution changes.

The `tests` job failure on this head was `internal/apptranscript`'s `TestItemTurnsFromEntriesLargeFixtureTiming`, a timing-budget test in a package this branch does not touch; re-run.

PR #925, opened 2026-09-06
feat(spawn): unify desktop composer with session composer

obra:
Unifies the desktop Start-an-agent composer with the session composer, and gives working directory + git info their own lines.

**What changed**
- Model (`ModelSwitchTrigger`, the session composer's own) and effort (the StatusRow quiet overlay-`<select>` recipe — native select over readout, no bordered box) move **into** the PromptCard control row at every width: attach · `model ▾` · `effort ▾` · Start.
- The below-card boxed Model (`ModelField`/`ModelCatalog`) + bordered Effort `Select` are gone; `ModelField.tsx` deleted (no remaining importers).
- Working directory keeps its own full line (directory button); git HEAD gets its own full line (`spawn-branch`) instead of a clipped suffix.
- Mobile: the `Reasoning effort` settings row is gone (the card owns it now — 5 rows, not 6); the card gains 399px container-query compression (model label yields first, chevron drops) and a 44px tap floor on the effort trigger.
- `spawnguard` updated to the new contract: in-card model+effort containment at all widths, effort tap-floor check, 5-row expectation.

**Verification**
- `Spawn.test` + `MobileSettingRows.test` rewritten to the new contract (RED first, then GREEN).
- `src/panes/spawn/`: 273 passed. Wider: spawn + `StatusRow` + `ModelSwitchTrigger` + `Composer`: 494 passed.
- `npm run spawnguard`: 5/5 widths (320/390/899/900/1440) PASS.
- `tsc --noEmit`, `biome check` clean on touched `src/` files. (`scripts/spawnguard/run.mjs` has pre-existing Biome format drift — verified it fails on the pristine file too — left untouched beyond the contract change.)
- Screenshots below are the real Spawn tree via the spawnguard harness (FakeClient, 2 staged attachments).

**Screenshots**

Desktop (1440px) — directory on its own line, model + effort quiet in the card:

spawn desktop 1440

Mobile (390px) — prompt intro, card with model + effort, 5 settings rows:

spawn mobile 390


obra:
roborev follow-up (commit 6841508):

- **Medium (model overflow)** — fixed. The slot + nested trigger spans now shrink/ellipsize via scoped `:global` (same piercing statusrow uses). Guard picks a \~100-char id through the real picker and asserts containment at all widths + active ellipsis at 320/390. RED-verified by reverting the CSS (trigger escaped the card by ~300px at 320/390).
- **Low (effort cursor)** — fixed. `.effortSelect` inherits the wrapper cursor; `pointer` only on enabled. CSS gate test added.
- **Low (guard sibling query)** — fixed. Guard queries `spawn-plugin-desktop` explicitly (it hides itself below 899px, so it's the honest desktop counterpart).
- **Low (loader identity)** — triaged as out of scope, not fixed: load-on-open is the shared `ModelSwitchTrigger` contract (the session side is identical), Spawn already busts its scoped cache on credential/scope change (`credentialsGeneration` + instances), and the component has no cancellation anywhere. A revision prop would be a cross-surface API change to a shared component — happy to do it as a follow-up if you want, but it's not something this PR introduced.

obra:
roborev round 2 (commit 9460838):

- **Medium (trigger staleness)** — fixed, and you're right that it matters here: Spawn's loader is harness+cwd+credential-scoped while the session side is scope-free, so reopening after a scope change genuinely replayed stale data. `ModelSwitchTrigger.openPicker` now carries a generation guard over the catalog, error, AND loading writes (a dead request landing mid-fresh-load also dropped the spinner). Shared fix, so the session side gets it too. RED test reproduced the exact failure (stale list `['Sonnet 4.5✓', 'GPT-5.5']` clobbering fresh); 17/17 trigger tests pass.
- **Low (selector map)** — fixed. The Spawn entry now describes the single in-card trigger + effort control; the only remaining `spawn-desktop-model` string anywhere is the unit test asserting it's gone.
- **Low (tautological CSS test)** — replaced, not just removed. The `spawn-attach-in-card` layoutguard case now asserts the geometry directly (effort containment, active ellipsis on a ~100-char id, 44px tap floor) and the CSS-text test is deleted with a pointer to the case. This caught a real harness gap along the way: raw-stylesheet harnesses silently drop `:global()`-pierced rules (unknown pseudo-class), so the build step now unwraps `:global()` exactly like the build does — that unwrap initially broke `compact-session-footer` (`:global(span:first-child)`), widened the pattern, full 21-case suite green. Mutation-tested per kata eevs (evidence in case.json).

obra:
roborev round 3 (commit a568ba9):

- **Medium (loader scope change while open)** — fixed as specified. `ModelSwitchTrigger` tracks the loader that started the current generation; a new identity under an open picker restarts the load against the new scope (staying open, spinner re-armed), while a change under a closed picker just bumps the generation so the dead request dies quietly. `startLoad` is `useCallback`-stable so the effect deps are honest (no suppression). RED test failed with the fresh loader never called; 18/18 trigger tests pass, tsc + Biome clean.
- **Low (leftover `:global(`)** — fixed as specified. `unwrapGlobal` now throws when any `:global(` survives the replace, so a nested-paren selector fails the guard loudly instead of passing silently. Unit-covered (`resolve-composes.test.mjs`, 4 tests); full 21-case layoutguard suite green.

obra:
roborev round 4 (commit 8ab3931) — the single Low:

- **Duplicate accessible name** — fixed as suggested. The card's effort `<select>` is now labelled "Prompt reasoning effort"; the Advanced Options schema field keeps "Reasoning effort". The visible readout is aria-hidden (the select speaks), so this changes only the spoken name — zero visual delta. New test opens Advanced Options with a scripted `reasoning_effort` schema option and asserts both names resolve distinctly; verified it fails against the old label (duplicate-name error), 276 spawn tests pass, spawnguard 5/5.

obra:
roborev round 5 — the single Low (stale selector map after the label rename): corrected. The entry now names "Prompt reasoning effort" and states `spawn-effort` is the wrapper holding the select.

PR #927, opened 2026-09-06
fix(agent): admit the delegate lane rollback on the close fence

### LINKED ISSUE #905: Delegate isolation rollback after prepareIsolation returns still forks git with no env-work admission
## What

PR #904 (closes #901) admits the delegate lane create and the rollbacks `prepareIsolation` itself owes under one env-work admission. Rollbacks reached *after* `prepareIsolation` returns are still unadmitted: when delegate construction or attach fails later, `isolation.cleanup` runs `rollbackFreshDelegateWorktree`, which forks git on the parent's environment with no `beginEnvWork` admission, so a concurrent close can reap the process table under it (the same hole #901 closed for the create).

## Why it was not folded into #904

Holding the create's admission across construction has no single release point: `failCommittedStart`'s retain-candidate arm never calls cleanup, so the admission would leak and stall every close for a full budget.

## Fix shape

Give `rollbackFreshDelegateWorktree` (or `isolation.cleanup`) its own best-effort admission, the way the `manage_worktree` rollbacks take theirs: admit if the session is not closing, run the git, release; if refused, skip the rollback and report it through the existing warning path. Test: fail construction after `prepareIsolation` returned while a close is in progress, and assert the rollback either runs inside the fence or is refused without touching the environment.

Noted by the #904 implementer; pre-existing.

obra:
## What was wrong

`prepareIsolation` admits the delegate lane create under one env-work span. Every rollback of that lane reached once `prepareIsolation` has returned runs outside it:

- the `CommitStart` failure in `delegateRuntime.create`
- `failCommittedStart`'s stop-won and ordinary arms
- `failAdoptedStart`

Each forks git on the **parent** session's environment via `rollbackFreshDelegateWorktree`. That is the process table `Session.close` reaps. With nothing admitted, `joinEnvWorkWithinCloseBudget` returns immediately and the close cleans the environment out from under a rollback whose git is still running.

Critically, the close is usually the *cause*: it cancels the turn context, construction fails, and `failCommittedStart` calls the rollback. So an admission taken where the rollback *starts* is refused by the very close it needs fencing against — it would close only the sub-case where the close arrives mid-rollback, and leave the common case racing.

## What changed

`agent/delegate_runtime.go` only.

The admission is taken in `delegateRuntime.create`, **before** `prepareIsolation` cuts the lane, and released by a single `defer` at the return. Every terminal path passes through that defer: the `CommitStart` failure, `failCommittedStart`'s three arms, `failAdoptedStart`'s two, the retain-candidate arms that roll nothing back, and the successfully launched delegate. So no path leaks it, and no rollback runs outside it — the window is closed entirely rather than narrowed.

It overlaps `prepareIsolation`'s own create admission for the length of the create. Two live admissions on the same fence are simply two things the close waits for.

`delegateIsolation.cleanup` goes back to taking no admission of its own, since every caller now sits inside the spawn's.

## How it is tested

`agent/delegate_isolation_close_test.go`:

- `TestDelegateIsolation_CloseBegunBeforeTheRollbackStillFencesIt` — the discriminating test. It drives the real spawn (`createDelegate` with `subagentPrepareFault("new_session")`), starts `Close` from inside the fault hook and waits for `closeAfterDisposeSweepJoin`, so `closing` is set **before** the rollback begins. Red before this change with `the close cleaned the environment while the rollback still held its git, even though closing was already set before the rollback began`.
- `TestDelegateIsolation_CloseWaitsForTheRollbackOfAFailedConstruct` — the same real arm with the close arriving *during* the rollback.
- `TestDelegateIsolation_LaneRollbackReleasesItsAdmission` — two subtests (a rollback path and the success/launch path) asserting the close emits no fence warning under a shortened budget. Verified to have teeth: mutating the fix to take the admission and never release it fails both subtests with the budget-expired warning.

All assertions use channels and the existing named `closeFenceProbe` tripwire; none use a wall-clock cap.

## Gates

All foreground, full logs kept, `grep -c FAIL` in parentheses.

- `gofmt -l` on both changed files: clean
- `go vet ./...`: exit 0; `go vet -tags evenerfuzz ./...`: exit 0
- `make lint-golangci` (as CI runs it): PASS
- `GOMAXPROCS=4 go test -race ./agent/... -count=1`: exit 0 (0)
- `GOMAXPROCS=4 go test -tags evenerfuzz -run Fuzz ./agent/ -count=1`: exit 0 (0)
- `GOMAXPROCS=4 go test ./agent/ -count=1 -timeout 40m`: exit 0 (0)

Closes #905

### DIFF
--- a/agent/delegate_isolation_close_test.go
+++ b/agent/delegate_isolation_close_test.go
@@ -185,3 +185,63 @@ func TestDelegateIsolation_FailedIsolationReleasesTheLaneAdmission(t *testing.T)
 		t.Errorf("the close waited out its budget on an admission the failed isolation step never released: %q", got)
 	}
 }
+
+// prepareIsolation admits the lane's create and the rollback it owes for its
+// OWN failures under one span, but a rollback isolation.cleanup reaches once
+// prepareIsolation has already returned — CommitStart failing,
+// failCommittedStart's arms, failAdoptedStart — is past that span and forks
+// git on the PARENT's environment with no admission of its own. A close that
+// begins while one of those rollbacks holds its git walks straight past the
+// join and reaps the process table underneath it, the same hazard the create
+// itself has above.
+func TestDelegateIsolation_CloseWaitsForTheLaneRollbackItRaces(t *testing.T) {
+	r := newWorktreeRepo(t)
+	root := r.s
+	runtime, reservation, project := reserveWorktreeIsolatedDelegate(t, root, "close under the lane rollback")
+
+	isolation, err := runtime.prepareIsolation(context.Background(), reservation, project, nil)
+	if err != nil {
+		t.Fatalf("prepareIsolation: %v", err)
+	}
+	lanePath := isolation.worktreePath
+
+	closeBegun := make(chan struct{})
+	closeDone := make(chan struct{})
+	envCleaned := make(chan struct{})
+	var cleanedOnce sync.Once
+	var held, cleanupDuringRollback atomic.Bool
+
+	root.cfg.testOnly.closeAfterDisposeSweepJoin = func() { close(closeBegun) }
+	root.cfg.testOnly.envCleanupObserved = func(execenv.ExecutionEnvironment) {
+		cleanedOnce.Do(func() { close(envCleaned) })
+	}
+	root.cfg.testOnly.worktreeGitRunner = func(ctx context.Context, env execenv.ExecutionEnvironment) worktree.GitRunner {
+		inner := gitRunner(ctx, env)
+		return func(args ...string) (string, error) {
+			if len(args) >= 3 && args[0] == "worktree" && args[1] == "unlock" && args[2] == lanePath && held.CompareAndSwap(false, true) {
+				go func() {
+					defer close(closeDone)
+					root.Close()
+				}()
+				<-closeBegun
+				select {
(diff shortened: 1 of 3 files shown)


obra:
On roborev's finding that `describe` runs git on the parent environment before the admission (delegate_runtime.go:1200-1237): declined after two independent reads of the path.

`describe` → `resolveWorktreeProject` → `ResolveProjectWith` requires a local environment for worktree isolation, so `EvalSymlinks` is `filepath.EvalSymlinks` and `MainCheckout` runs `localStructuralMainRoot`, a pure `os.Stat`/`os.ReadFile` ancestor walk. It hands off to git in exactly one branch (project_resolver.go:155): a `.git` file that validates as a submodule pointer, and then only a read-only `rev-parse`. At that point no lane, branch, lock, or sidecar exists, so a git process reaped by a concurrent close leaves no residue; that is the difference from the rollbacks this PR fences. Admitting above `describe` would also span `reclaimDelegateRuntimeCapacity`, whose synchronous `teardownChildSession` (delegate_tree_reclaim.go:385) is exactly the work `joinEnvWorkWithinCloseBudget` forbids admitting (session_lifecycle.go:392-395, "an operation that waited on the delegate-tree close would deadlock this join").

PR #928, opened 2026-09-06
fix(agent): settle the worktree clone a shared-env child enters

### LINKED ISSUE #913: A delegating child that enters a worktree from its parent's shared environment leaks the clone's scratch at teardown

obra:
## What was wrong

A delegate spawned with neither a `working_dir` nor a box of its own runs on its parent's very environment object (`Session.ownsEnv` false, `prepareSubagentEnvironment` returns the parent's env untouched), and it keeps the full `manage_worktree` tool — only worktree isolation or a zero delegation allowance strips it. So such a child can `enter` a worktree, which runs `swapEnvAndRefresh` and installs a clone of its own.

Two things then went wrong, both verified against the real spawn and enter paths:

- **The enter took the parent's scratch.** Step 0's `next.AdoptSessionScratch(current)` ran unconditionally, moving the directory the live parent was working in onto the child's clone; the parent's environment was left owning none and minted a fresh one on its next command.
- **The teardown settled nothing.** `teardownChildSession` decided from `ownsEnv`, frozen at spawn and still false, so `releaseOwnedChildEnvironment` returned immediately; and the parked/abandoned retains added by #878 run only under `if cleanupEnv` in `Session.close`, which a child teardown never sets. The clone's scratch lease was held for the rest of the daemon's uptime.

## What changed

A child now records the environment its parent handed it (`Session.parentSharedEnv`, set beside `ownsEnv` at spawn and at delegate restore through `recordEnvironmentOwnership`).

- `swapEnvAndRefresh` skips the scratch move in both directions when that object is the source or the target: an enter leaves the parent's scratch where the parent is working and the clone provisions its own on its first command; an exit hands the clone's scratch to nobody.
- `environmentOwnedAtTeardown` replaces the raw `ownsEnv` read at both teardown call sites: ownership is no longer frozen at spawn, so the clone a child built for itself mid-life is settled, while the parent's own object never is.
- `teardownChildSession` also drains `abandonedEnvs` (the clones an enter-then-exit or a switch left behind), which only `close`'s `cleanupEnv` block reached before. `recordAbandonedEnvironmentLocked` now excludes `parentSharedEnv` structurally, so the parent's object can never enter that list.

No teardown path calls `Cleanup()`: leases are released and directories kept, so the invariant that a session reaps only a process table it constructed still holds. Root-session behaviour is unchanged (`parentSharedEnv` is nil there).

## How it is tested

Two new tests in `agent/subagent_teardown_unix_test.go`, both through the real spawn path (`prepareSubagentRun` with a delegation allowance) and the real enter path (the child's own `manage_worktree` `switch` by path onto a sibling worktree), asserting real lease files and directories:

- `TestSharedEnvChildTeardownReleasesTheEnteredWorktreeScratch` failed on main with `the entered clone's scratch /…/evener-sandbox-2445113863 lease is still held after the shared child's teardown`.
- `TestSharedEnvChildKeepsItsWorktreeScratchAcrossExit` failed on main with `the entered clone's scratch is the parent's own /…/evener-sandbox-1775178725, want a scratch of its own`, and on the same lease message.

Both also pin that the parent's scratch and its in-flight process are untouched.

## Gates

- `gofmt -l` on every changed file: clean.
- `go vet ./...` and `go vet -tags evenerfuzz ./...`: exit 0.
- `make lint-golangci`, `make lint-evenerfuzz lint-eval` (as CI runs them): PASS.
- `GOMAXPROCS=4 go test -race ./agent/ -count=1`: `ok … 1262.542s`, 0 FAIL.
- `GOMAXPROCS=4 go test -tags evenerfuzz -run Fuzz ./agent/ -count=1`: 0 FAIL.
- `GOMAXPROCS=4 go test ./agent/ -count=1 -timeout 40m`: `ok … 465.553s`, 0 FAIL.

Closes #913

### DIFF
--- a/agent/cov_s5_subagents_helpers_test.go
+++ b/agent/cov_s5_subagents_helpers_test.go
@@ -115,3 +115,43 @@ func TestS5Cov_SubagentNeedsCommunicateNudge(t *testing.T) {
 		t.Error("a named agent should not need the nudge")
 	}
 }
+
+// uncomparableEnv is an execution environment Go cannot compare with `==`: a
+// struct VALUE carrying a map. Nothing calls its methods; the embedded
+// interface is there to satisfy the type.
+type uncomparableEnv struct {
+	execenv.ExecutionEnvironment
+	roots map[string]string
+}
+
+// Environment ownership is decided by identity — a child settles the
+// environment it holds unless that is the very object its live parent works in
+// — so the comparison has to answer for any environment a session can run on,
+// including a double whose dynamic type `==` refuses to compare.
+func TestS5Cov_SameEnvironment(t *testing.T) {
+	first := execenv.NewLocalExecutionEnvironment(t.TempDir())
+	second := execenv.NewLocalExecutionEnvironment(t.TempDir())
+	if !sameEnvironment(first, first) {
+		t.Error("an environment is not the same as itself")
+	}
+	if sameEnvironment(first, second) {
+		t.Error("two distinct environments compared equal")
+	}
+	if sameEnvironment(first, first.WithWorkingDirectory(t.TempDir())) {
+		t.Error("a clone compared equal to the environment it was built from")
+	}
+	if !sameEnvironment(nil, nil) {
+		t.Error("two absent environments are not the same")
+	}
+	if sameEnvironment(first, nil) || sameEnvironment(nil, first) {
+		t.Error("an environment compared equal to none at all")
+	}
+	// `==` would panic on this pair rather than answer.
+	one := uncomparableEnv{roots: map[string]string{"a": "b"}}
+	if sameEnvironment(one, one) {
+		t.Error("an environment nothing can alias compared equal")
+	}
+	if sameEnvironment(first, one) {
+		t.Error("environments of different types compared equal")
+	}
+}
--- a/agent/delegate_runtime.go
+++ b/agent/delegate_runtime.go
@@ -1712,7 +1712,7 @@ func (runtime delegateRuntime) restoreIdle(started delegateStartCommit) (*subage
 	if err != nil {
 		return nil, false, err
 	}
-	child.ownsEnv = ownsFresh
+	child.recordEnvironmentOwnership(childEnv, ownsFresh)
 	discardEnv = false
 	if child.delegateController != s.delegateController || child.owningDelegateID != started.lease.delegateID {
 		child.discardRestoredCandidate()
--- a/agent/session.go
+++ b/agent/session.go
@@ -98,10 +98,15 @@ type Session struct {
 	// them; a shared one belongs to the live parent and is left alone. The child
 	// records it because a teardown can reach a child no parent bookkeeping names
 	// — a runtime the stable controller holds for its own close.
-	ownsEnv   bool
-	clock     clock.Clock
-	stateDir  string
-	installID string
+	ownsEnv bool
+	// parentSharedEnv is the environment a child was handed when it is the live
+	// parent's own object (ownsEnv false) — the one environment this session must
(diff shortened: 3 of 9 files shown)


obra:
On roborev's third finding (agent/subagents.go:213-226, an owned child's parked `worktreeRestoreEnv` never settled at teardown): real, pre-existing, and untouched by this PR. `retainParkedWorktreeEnvironmentScratch` still has exactly one caller, `close()` under `cleanupEnv` (agent/session_lifecycle.go:519); these commits add no call to it, and `recordAbandonedEnvironmentLocked` still skips `worktreeRestoreEnv`, so the parked environment cannot reach the new abandoned-environment drain. Tracked as #929, filed from this PR's review, to keep this change on the shared-environment child that #913 describes.

The other two findings are addressed in 8b939813c: the exit-under-close rollback now releases only the scratch it moved, pinned by `TestWorktreeSwap_CloseDuringASharedChildExitKeepsTheParentScratchLease` (red first); the sandboxed clone keeping the box's session tmp is the intended design (the lease belongs to the parent's environment and is released at the parent's close), now documented and pinned by `TestSharedEnvChildInsideTheParentBoxKeepsTheParentScratchLease`.

obra:
On roborev's Low at subagents.go:225 (abandoned environments retained under `disposeChildScratch`): declined as unreachable.

A non-empty `abandonedEnvs` requires the child to have executed a `manage_worktree` call: it is written only by `recordAbandonedEnvironmentLocked` (session_lifecycle.go:599), reached only through `swapEnvAndRefresh` (session_env_swap.go:182), whose callers are `enterWorktree`/`exitWorktree` (session_tools_worktree.go:840,869), reachable only from the five `manage_worktree` op handlers, so only inside a turn. Every caller that passes `disposeChildScratch` tears the child down before its run loop launches (subagents.go:696,1087,1136,1159; delegate_runtime.go:1584,1588,1893,1905,1918,1924,1949), and every teardown of a child that has run passes `retainChildScratch` (parent close, eviction, stop, lane disposal, reclamation). `discardRestoredCandidate` (session_lifecycle.go:688) never reaches the drain, and a restored candidate's `abandonedEnvs` starts empty with child re-entry refused. So the retain-versus-dispose distinction cannot apply to an abandoned environment today.

Issue #369, opened 2026-08-23
find_files with a ** pattern rooted at / never returns: doublestar walk follows /proc symlink loops, takes no context, cannot be cancelled — wedged a delegate and hung a run for 109 minutes


obra:
Field case (retest7, sam-cell-seg trial UzmUsFz, reward 1.0 but 6528s of dead time): a delegate issued find_files {"path": "/", "pattern": "**/convert_masks.py", "include_ignored": true}. The call never returned. LocalExecutionEnvironment.GlobWithExclusions (agent/execenv/local.go:1287-1315) runs doublestar.Glob(os.DirFS(base), pattern) with no context.Context — unbounded and uncancellable. With base=/ and a ** pattern the walk descends into /proc, where /proc/<pid>/root symlinks back to /, and recurses without end.

Reproduced in the real task image with a standalone probe (doublestar v4.9.1, the go.mod version):
- glob / "**/x.py" → killed at 300s
- glob /proc "**/x.py" → killed at 90s
- glob /usr "**/x.py" → 261ms
- glob / "x.py" (no **) → 77µs
For contrast a full os.walk("/") in the same image is 0.6s / 9928 dirs — the trigger is specifically ** rooted at a tree containing /proc or /sys.

Downstream: the root agent noticed, tried job_stop(include_children, max_wait_ms=30000) → stop_pending forever (the delegate is inside an uninterruptible tool call), tried delegate_send → target_busy, gave up and issued its terminal communicate — and then the one-shot drain waited on the wedged "live" delegate for 6528 seconds until the external 7200s wall. Cost: 109 minutes of budget and an EC2 instance-hour; the trial passed only because delivery preceded the hang.

Fix surface:
1. Thread a context through the glob path and honor cancellation (job_stop must be able to kill a tool call, not just request it).
2. Bound the walk: skip symlinked directories during ** traversal (or refuse/warn on ** rooted at /).
3. The drain half is #317's exact scenario — a stop-requested, unresponsive subagent must eventually be abandoned.

The delegate's interview (on a copy) also confirms the model side: the task named an explicit path; searching / was "a generic locate-the-file-anywhere habit" with no pricing of the walk. But the model mistake should cost seconds, not the session.

PR #930, opened 2026-09-06
fix(execenv): bound the glob walk's listings, matches, and memory

### LINKED ISSUE #497: A single glob rooted at / OOM-kills the agent: the directory-listing budget is gated on filesystems that report no file identity, so it never applies on Linux

## The call

```
```

## Reproduction

## Mechanism

```go
// ...
```

```go
	...
	if !hasFileIdentity(info) {
		w.unidentified++
		return nil
	}
```

```go
var matches []string
```

## Directions

### LINKED ISSUE #938: grep's tree walk is still unbudgeted; only its ignore discovery shares the glob budget

## What is unbounded

## Fix shape

obra:
## What was wrong

A `find_files` call with a `**` pattern rooted at `/` could exhaust memory and OOM-kill the agent process (#497). The walk *terminated* — the ancestor/SameFile cycle check added for #369 catches `/proc/<pid>/root` re-entering the tree — but nothing bounded what it spent getting there. Five separate holes, found in that order:

1. **The directory-listing budget was unreachable.** The only counter incremented inside `if !hasFileIdentity(info)`. Every file on ext4, APFS and HFS+ carries `(device, inode)` identity, so on the filesystems agents actually run on that branch was never taken.
2. **`globMatches` accumulated matches with no cap.** A `**` over `/` is millions of paths.
3. **The walk retained one `fs.FileInfo` per listed directory**, though the cycle check only consults *ancestors* of the directory being admitted.
4. **Ignore discovery was unbounded and ran first.** `loadIgnoreSet` walked the entire base collecting `.gitignore` rules before the first `globMatches` existed, so the default `include_ignored: false` traversed all of `/` before any bound applied. Native grep had the identical walk.
5. **Every listing materialized its whole directory before any bound was consulted.** `os.DirFS`'s `ReadDir` is `os.ReadDir`, and `secureDirFS` ended in `df.ReadDir(-1)`, so one directory holding millions of entries exhausted memory before the listing budget or match cap ever ran.

## What changed

A `globBudget`, created once per glob or grep call and shared across ignore discovery and every brace-expanded pattern, now bounds four things: directory listings (1,000,000 per call), matches (10,000 per call), entries materialized by a single listing (200,000), and entries held live at once across every listing the walk still has open (500,000, roughly 52 MB at ~105 bytes per entry). The last one matters because a per-directory cap does not bound a walk: `fs.WalkDir` and doublestar both keep an ancestor's entry slice alive while reading a child, so peak is otherwise the per-directory cap times depth. Release is free — to be listing a directory at all, the walk must have left every non-ancestor, which is the same rule the ancestor chain already uses. Listings are pulled in 4,096-entry chunks and charged after each one, so the walk stops instead of finishing and then complaining, and the chunk loop checks the context between chunks — `cancelFS` only sees a listing start, not the chunks inside it.

Refusals are a typed `globBudgetError` carrying the operation, the bound, the count, the offending directory, and whether the walk could have detected a symlink cycle at all, so callers and tests read fields rather than parse sentences. The advice names the lever that actually helps: a glob can narrow its pattern, a grep cannot, because its pattern is a regex applied after the walk has already listed everything.

`listed` became an ancestor chain pruned to the path being walked, so retention is O(depth) rather than O(listings). The chain is only a cache: an ancestor it does not hold is re-stat'ed, so a walk that jumps between subtrees still cannot miss a cycle. No configuration knobs were added.

## Caller-visible changes

- **Grep is budgeted on both arms** and can now fail where it used to return partial output. `grepWalk` and `secureBrowseWalkDir` were unbudgeted; both now charge the same budget their ignore discovery spends, and only for directories they actually descend into, not ones they skip.
- **A single directory over 200,000 entries now fails glob and grep outright**, naming that directory, rather than being read into memory.
- **A walk holding more than 500,000 entries live at once fails**, which a deep tree of large directories can reach even when no single directory is over the per-directory cap. A wide, flat repository is unaffected: the bound is on entries held simultaneously, not a cumulative count.
- **Ignore discovery can now fail a call.** It stays best-effort for I/O errors, but a budget refusal propagates: a partial ignoreSet reported as complete would silently under-exclude the rest of a huge tree.
- **Sandboxed listings are lexically sorted** where they came back in raw directory order, which changes sandboxed match order and grep visit order. `os.DirFS` already sorted; the two arms now agree, and the match cap truncates to the same prefix on both.
- The glob tool tells the model when the match cap cut a listing short, naming the cap and saying it counts *candidate* matches, before dotfile/gitignore exclusion drops any.
- `GlobExcluder.GlobWithExclusions` gained a fourth return, `truncatedAt`. `*LocalExecutionEnvironment` is its only implementation in this repo.

## How it is tested

Every bound was written test-first and each new test was confirmed failing for its stated reason before the fix. The tests assert *work done*, not just results — entries actually read, directory listings actually made, `peakDirEntries` — because a fix that reads a whole directory and then reports a refusal would satisfy a result-only assertion while still doing exactly the thing that OOMs. Sample first failures:

- `walk retained 31 directory identities after listing 30 sibling directories`
- `Glob over a 30-entry directory with an entry budget of 10 = (...30 matches..., <nil>), want a *globBudgetError`
- `grepNative's own walk over a 30-directory tree with a listing budget of 40 = (_, <nil>), want a *globBudgetError`

Two guards cannot be reached by any static tree, because ignore discovery prunes a subset of what the walks prune and runs first over the same filesystem; probes force the concurrent-growth case by growing a directory inside a stubbed walk. Termination stays pinned by the #369 regression tests, and a characterization test proves the ancestor check still catches a cycle reached through a never-listed ancestor.

Gates run per round, all foreground: `gofmt`; `make vet` and `GOOS=windows make vet`; `go vet -tags evenerfuzz ./...`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test -race` on `./agent/execenv/` and `./agent/`; `-tags evenerfuzz -run Fuzz` on both; and the full `go test ./agent/`.

Closes #497
Closes #938

### DIFF
--- a/agent/cov_s3_shelltools_test.go
+++ b/agent/cov_s3_shelltools_test.go
@@ -3,8 +3,10 @@ package agent
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 
@@ -189,5 +191,55 @@ func TestS3Cov_GlobTool_FullyExcludedResultExplainsRatherThanEmpties(t *testing.
 	}
 	if !strings.Contains(res.Output, "index.js") {
 		t.Fatalf("expected include_ignored=true to find the match, got: %q", res.Output)
+	}
+}
+
+// truncatingGlobEnv wraps an execenv.ExecutionEnvironment and implements
+// execenv.GlobExcluder with a fixed, pre-truncated result, so the tool
+// boundary below can be tested without building a tree large enough to
+// actually trip the real match cap.
+type truncatingGlobEnv struct {
+	execenv.ExecutionEnvironment
+	matches     []string
+	truncatedAt int
+}
+
+func (e *truncatingGlobEnv) GlobWithExclusions(ctx context.Context, pattern, basePath string, includeIgnored bool) ([]string, int, int, error) {
+	return e.matches, 0, e.truncatedAt, nil
+}
+
+// TestS3Cov_GlobTool_TruncationNoteNamesTheCapWithoutDroppingMatches proves
+// the glob tool boundary surfaces GlobExcluder's truncatedAt (#497): when a
+// glob call hits the match cap, the tool result must still contain every
+// match it collected *and* say the listing was capped, so a model cannot
+// mistake a partial result for the whole answer. A bare
+// strings.Join(matches, "\n") is indistinguishable from an uncapped result,
+// which is exactly the failure mode this guards against.
+func TestS3Cov_GlobTool_TruncationNoteNamesTheCapWithoutDroppingMatches(t *testing.T) {
+	t.Parallel()
+	dir := t.TempDir()
+	s := newSession(t, withDir(dir))
+
+	const capAt = 7
+	matches := make([]string, capAt)
+	for i := range matches {
+		matches[i] = fmt.Sprintf("/match/%02d.txt", i)
+	}
+	s.env = &truncatingGlobEnv{ExecutionEnvironment: s.env, matches: matches, truncatedAt: capAt}
+
+	res := s3cov_exec(t, s, "glob", `{"pattern":"**/*","path":"."}`)
+	if res.IsError {
+		t.Fatalf("glob error: %v", res.Output)
+	}
+	for _, m := range matches {
+		if !strings.Contains(res.Output, m) {
+			t.Fatalf("truncated glob result dropped match %q: %q", m, res.Output)
+		}
+	}
+	if !strings.Contains(res.Output, strconv.Itoa(capAt)) {
+		t.Fatalf("truncated glob result does not name the cap (%d): %q", capAt, res.Output)
+	}
+	if res.Output == strings.Join(matches, "\n") {
+		t.Fatal("truncated glob result is indistinguishable from the bare match list; a model can't tell truncation happened")
 	}
 }
--- a/agent/execenv/execenv.go
+++ b/agent/execenv/execenv.go
@@ -178,6 +178,7 @@ type ExecutionEnvironment interface {
 type GlobExcluder interface {
 	// GlobWithExclusions behaves like Glob but also returns the number of
 	// candidate matches dropped by the default dotfile/gitignore exclusion
-	// (always 0 when includeIgnored is true).
-	GlobWithExclusions(ctx context.Context, pattern, basePath string, includeIgnored bool) (matches []string, excluded int, err error)
+	// (always 0 when includeIgnored is true), and truncatedAt: the match cap
+	// that cut the listing short, or 0 when every match is reported.
+	GlobWithExclusions(ctx context.Context, pattern, basePath string, includeIgnored bool) (matches []string, excluded int, truncatedAt int, err error)
 }
--- a/agent/execenv/gitignore.go
+++ b/agent/execenv/gitignore.go
@@ -38,10 +38,13 @@ type ignoreDir struct {
 }
 
 // loadIgnoreSet walks fsys (rooted at the search base) collecting every
-// .gitignore file's rules. It is best-effort: unreadable files are skipped
-// rather than failing the search, and a base with no .gitignore files
-// anywhere (including one that is not inside a git repository at all) yields
-// an ignoreSet that matches nothing.
+// .gitignore file's rules. It is best-effort for I/O errors: unreadable files
+// are skipped rather than failing the search, and a base with no .gitignore
+// files anywhere (including one that is not inside a git repository at all)
+// yields an ignoreSet that matches nothing. The one error it does return is
+// budget refusing to let the walk keep going — that has to reach the caller,
+// since a partial ignoreSet built from a walk that gave up partway through
+// would silently under-exclude the rest of a huge tree.
 //
 // skip, when non-nil, is consulted for every path (relative to fsys' root,
 // slash-separated) before it is descended into or read; skip returning true
@@ -51,8 +54,14 @@ type ignoreDir struct {
 // this walk otherwise has no other way to honor, since fsys itself (a
 // symlink-refusing, root-confined secureDirFS) enforces confinement but not
 // masking. The off-path caller (no masking concept) passes a no-op skip.
-func loadIgnoreSet(fsys fs.FS, skip func(relPath string) bool) *ignoreSet {
+//
+// budget is the same one the caller's glob walk spends listings from, so
+// ignore discovery and pattern matching together are bounded as one call's
+// worth of work rather than each getting its own unbounded pass over the
+// tree.
+func loadIgnoreSet(fsys fs.FS, skip func(relPath string) bool, budget *globBudget) (*ignoreSet, error) {
 	set := &ignoreSet{}
+	var budgetErr error
 	_ = fs.WalkDir(fsys, ".", func(p string, d fs.DirEntry, err error) error {
 		if err != nil {
 			return nil //nolint:nilerr // best-effort: skip unreadable entries
@@ -71,6 +80,16 @@ func loadIgnoreSet(fsys fs.FS, skip func(relPath string) bool) *ignoreSet {
 			if p != "." && strings.HasPrefix(d.Name(), ".") {
 				return fs.SkipDir
 			}
+			// cycleSafe is always true here: fs.WalkDir never follows a
+			// symlink into a directory, so this walk can never re-enter
+			// itself the way the pattern walk's /proc/<pid>/root case does.
+			// The flag only picks the refusal's wording, not whether the
+			// bound applies — an unbounded walk over a huge tree costs
+			// unbounded work whether or not it could have looped.
+			if err := budget.listing(true); err != nil {
+				budgetErr = err
+				return err
+			}
 			return nil
 		}
 		if d.Name() != ".gitignore" {
@@ -85,7 +104,7 @@ func loadIgnoreSet(fsys fs.FS, skip func(relPath string) bool) *ignoreSet {
 		set.dirs = append(set.dirs, ignoreDir{rel: dir, matcher: matcher})
 		return nil
 	})
-	return set
+	return set, budgetErr
 }
 
 // matches reports whether relPath (slash-separated, relative to the search
--- a/agent/execenv/gitignore.go
+++ b/agent/execenv/gitignore.go
@@ -2,6 +2,7 @@ package execenv
 
 import (
 	"context"
+	"errors"
 	"io/fs"
 	"path"
 	"strings"
@@ -64,6 +65,15 @@ func loadIgnoreSet(fsys fs.FS, skip func(relPath string) bool, budget *globBudge
 	var budgetErr error
 	_ = fs.WalkDir(fsys, ".", func(p string, d fs.DirEntry, err error) error {
 		if err != nil {
+			// A budget refusal is not an unreadable entry: skipping it would
+			// let the directory that tripped the bound be read again by
+			// whatever walks next, and would leave this set reported as
+			// complete when it stopped partway, silently under-excluding the
+			// rest of the tree.
+			if _, refused := errors.AsType[*globBudgetError](err); refused {
+				budgetErr = err
+				return err
+			}
 			return nil //nolint:nilerr // best-effort: skip unreadable entries
 		}
 		if p != "." && skip != nil && skip(p) {
--- a/agent/execenv/gitignore_covtest_test.go
+++ b/agent/execenv/gitignore_covtest_test.go
@@ -40,7 +40,10 @@ func TestIgnoreSet_Matches_PathEqualsDir(t *testing.T) {
 		"sub":            {Mode: os.ModeDir | 0o755},
 		"sub/.gitignore": {Data: []byte(content)},
 	}
-	set := loadIgnoreSet(fsys, nil)
+	set, err := loadIgnoreSet(fsys, nil, &globBudget{})
+	if err != nil {
+		t.Fatal(err)
+	}
 	// "sub" == id.rel, so rel becomes "." — "*.log" won't match "."
 	if set.matches("sub", false) {
 		t.Fatal("expected sub (file) NOT to be ignored")
@@ -112,9 +115,12 @@ func TestLoadIgnoreSet_SkipFile(t *testing.T) {
 		"skipme":          {Data: []byte("data\n")},
 		"skipme/file.txt": {Data: []byte("data\n")},
 	}
-	set := loadIgnoreSet(fsys, func(relPath string) bool {
+	set, err := loadIgnoreSet(fsys, func(relPath string) bool {
 		return relPath == "skipme"
-	})
+	}, &globBudget{})
+	if err != nil {
+		t.Fatal(err)
+	}
 	// The .gitignore at root should still be loaded.
 	if !set.matches("test.tmp", false) {
 		t.Fatal("expected test.tmp to be ignored by root .gitignore")
@@ -156,5 +162,9 @@ func loadIgnoreSetFromLines(t *testing.T, dir string, lines ...string) *ignoreSe
 	fsys := fstest.MapFS{
 		path: {Data: []byte(content)},
 	}
-	return loadIgnoreSet(fsys, nil)
+	set, err := loadIgnoreSet(fsys, nil, &globBudget{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	return set
 }
--- a/agent/execenv/gitignore_covtest_test.go
+++ b/agent/execenv/gitignore_covtest_test.go
@@ -40,7 +40,7 @@ func TestIgnoreSet_Matches_PathEqualsDir(t *testing.T) {
 		"sub":            {Mode: os.ModeDir | 0o755},
 		"sub/.gitignore": {Data: []byte(content)},
 	}
-	set, err := loadIgnoreSet(fsys, nil, &globBudget{})
+	set, err := loadIgnoreSet(fsys, nil, newGlobBudget("glob"))
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -117,7 +117,7 @@ func TestLoadIgnoreSet_SkipFile(t *testing.T) {
 	}
 	set, err := loadIgnoreSet(fsys, func(relPath string) bool {
 		return relPath == "skipme"
-	}, &globBudget{})
+	}, newGlobBudget("glob"))
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -162,7 +162,7 @@ func loadIgnoreSetFromLines(t *testing.T, dir string, lines ...string) *ignoreSe
 	fsys := fstest.MapFS{
 		path: {Data: []byte(content)},
 	}
-	set, err := loadIgnoreSet(fsys, nil, &globBudget{})
+	set, err := loadIgnoreSet(fsys, nil, newGlobBudget("glob"))
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/agent/execenv/glob_budget_test.go
+++ b/agent/execenv/glob_budget_test.go
@@ -1,0 +1,211 @@
+package execenv
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// stubMaxGlobDirListings lowers the directory-listing budget for a test and
+// restores it when the test ends, so a budget test can use a tree small
+// enough for t.TempDir() rather than one large enough to trip the real
+// (100,000) bound.
+func stubMaxGlobDirListings(t *testing.T, n int) {
+	t.Helper()
+	orig := maxGlobDirListings
+	maxGlobDirListings = n
+	t.Cleanup(func() { maxGlobDirListings = orig })
+}
+
+// stubMaxGlobMatches lowers the match-count budget for a test and restores it
+// when the test ends.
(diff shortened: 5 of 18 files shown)


obra:
Round 2 pushed as `58d765d4a`, addressing the scoped review and both roborev findings.

**roborev Medium (ignore discovery unbounded)** — confirmed and fixed. `loadIgnoreSet` ran `fs.WalkDir` over the entire base before the first `globMatches` existed, so the budget added in round 1 did not cover it. `include_ignored: true` does bypass it, but the field default is `false`, so `glob("**/*.go", "/")` still traversed all of `/`. Ignore discovery now spends the same `globBudget` as the pattern walks and its refusal reaches the caller (a partial ignoreSet would silently under-exclude the rest of a huge tree). Native grep had the identical unbounded walk and is bounded too. Pinned by `TestGlobIgnoreDiscoveryIsChargedToTheBudget`, which uses a metacharacter-free pattern matching nothing so the only possible charge is ignore discovery, plus an `include_ignored: true` control.

**Stat fallback unpinned** — `TestGlobWalkRefusesACycleThroughAnUnlistedAncestor` lists `a/loop` as the walk's first-ever listing, so neither `.` nor `a` is on the chain. Verified by mutation: replacing the fallback with `return nil, false` fails it.

**Budget messages unpinned** — the refusal is now a typed `globBudgetError{listings, budget, cycleSafe}`. `TestGlobBudgetErrorRecordsWhetherTheWalkCouldDetectCycles` asserts `cycleSafe` structurally via `errors.As`, no prose. Both sentences name the budget and say to narrow the pattern.

**Sandboxed truncation non-deterministic** — `secureDirFS.ReadDir` returned `df.ReadDir(-1)` unsorted while `os.DirFS` sorts. It sorts now; `TestSandboxedGlobTruncatesToAStablePrefix` pins that two runs agree and match the lexical prefix.

**roborev Low (ctx on the cap fast path)** — fixed and pinned by `TestGlobMatchesReportsCancellationAfterTheCapTripped`.

**Cleanups** — `budget.truncatedAt()` replaces the duplicated blocks; `retained()` moved into the test file; `push`'s comment rewritten without what-changed prose; the type comment now names the sibling-alias case (same identity, neither an ancestor, so both names are listed and it is bounded only by the listing budget); the cap note says the cap counts candidate matches before exclusion.

**Budget value** — raised from 100,000 to **1,000,000** listings per call. One call now spends it on ignore discovery *and* on each brace-expanded pattern, so the old value would have regressed real repositories: a 60k-directory monorepo with a five-way brace expansion costs ~360k listings and used to succeed. Memory is already held to a small constant by the match cap and the O(depth) chain, so this is a work bound, sized between the largest realistic repository and `/` on a dev box (500k to several million directories). At 1,000,000 I do not expect any real monorepo glob to error where it used to succeed.

Gates, all foreground and clean: gofmt; `make vet` + `GOOS=windows make vet`; `go vet -tags evenerfuzz`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test -race` on `./agent/execenv/` and `./agent/` (0 FAIL); `-tags evenerfuzz -run Fuzz` on both (0 FAIL); full `go test ./agent/` (0 FAIL).

obra:
Final round pushed as `c0a3c67b7`.

1. **Budget error said "glob walk" on grep's path** — fixed. `globBudget` now carries the operation it was built for (`newGlobBudget("glob"|"grep")` at all four production sites) and `globBudgetError` reports it, so a grep that blows the budget through ignore discovery says "grep walk made N directory listings". The cycle-safe sentence also no longer claims "for one glob call", which was self-contradictory once `op` can be "grep". New `TestGrepIgnoreDiscoveryIsChargedToTheBudget` drives `grepNative` directly (the `return "", err` path) with the budget lowered through the seam and asserts `op == "grep"` via `errors.As` — no prose matching. First failure: `globBudgetError.op = "", want "grep"`.
2. **Contradictory doc sentence** — fixed. It now says a call's listings are the sum of ignore discovery plus every brace-expanded pattern's walk, charged to one counter; the 60k/`/` arithmetic was already correct under that reading and is unchanged.
3. **Stale "(100,000)"** — gone; the comment names no number so it cannot go stale again.
4. **`TestGlobIgnoreDiscoveryIsChargedToTheBudget` strengthened** — now asserts the typed `globBudgetError` via `errors.As` with `op == "glob"`, plus a `countingFS` through `stubGlobBaseFS` proving at most `budget+1` listings and strictly fewer than the tree's directory count. First failure: `globBudgetError.op = "", want "glob"`. The `include_ignored: true` control is kept.

**Confirmed for filing, not fixed here:**
- **grep's own tree walk is still unbudgeted.** `grepWalk` (`local.go:1159`) and `secureBrowseWalkDir` (`securepath_browse_fdops_unix.go:25`) are both plain `fs.WalkDir` seams taking no budget. Only grep's *ignore discovery* is bounded now; the search walk itself can still traverse an arbitrarily large tree.
- **There is no per-tool deadline on the glob tool.** `DefGlob()` takes no timeout argument and `Registry.executeCall` never wraps the exec context (`grep -c 'WithTimeout|WithDeadline' agent/internal/tool/registry.go` is 0). The only bounds are ctx cancellation (session/job stop) and the listing budget — so a call that walks toward 1,000,000 listings can run tens of seconds with nothing but the model's own patience to stop it.

Gates, all foreground, 0 FAIL each: gofmt; `make vet` + `GOOS=windows make vet`; `go vet -tags evenerfuzz`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test -race` on `./agent/execenv/` and `./agent/`; `-tags evenerfuzz -run Fuzz` on both; full `go test ./agent/`.

obra:
On roborev's two findings on 7cdb28721: both real, both pre-existing, and both tracked separately so this PR stays on the `**` fan-out that #497 reports.

- Per-directory materialization (High): `ReadDir` reads a whole directory before either cap is consulted, so one directory with millions of entries can still exhaust memory. That is a different bound (entries per listing, not listings per call) and needs chunked enumeration on the plain and sandboxed paths plus ignore discovery. Filed as #939.
- Native grep's own walk (Medium): only grep's ignore discovery shares the budget here; `grepWalk` and `secureBrowseWalkDir` remain unbudgeted. Filed as #938 from this PR's review, with the fix shape (route the walk through the same budget, which already carries the grep operation name).

obra:
Round 5 pushed as `934a983d4`. Both roborev findings are fixed here.

**roborev High (per-directory materialization) — fixed.** `os.DirFS`'s `ReadDir` is `os.ReadDir` and `secureDirFS` ended in `df.ReadDir(-1)`, so every listing built the whole directory before the listing budget or match cap ran. Listings now pull `globDirChunk` (4096) entries at a time and charge the running total after each chunk, refusing past `maxGlobDirEntries` (200,000) with the existing typed `globBudgetError` — now carrying a `kind` so an entry refusal is told from a listing refusal structurally, with the same operation-appropriate advice. The bound is **per listing, not cumulative**, because it bounds peak memory: the walk frees each directory's slice as it moves on. Entries are sorted once at the end, not per chunk, so `TestSandboxedGlobTruncatesToAStablePrefix` still holds.

Tests assert **entries actually read**, not just the result — a fix that materialized everything and then reported a refusal would pass a result-only check while still doing the thing that OOMs. On the plain path a paced FS double counts what a chunked reader pulls; on both paths `globBudget.peakDirEntries` is asserted to be at least the bound but strictly less than the directory size. First failures: `Glob over a 30-entry directory with an entry budget of 10 = (...30 matches..., <nil>), want a *globBudgetError` and `secureDirFS.ReadDir ... = (30 entries, <nil>), want a *globBudgetError`.

**roborev Medium (grep's own walk) — fixed here, ~40 lines as you estimated.** Both grep arms now hoist one budget shared by ignore discovery and the walk, and charge one listing per directory inside the callback each arm already has. `grepWalk` and `secureBrowseWalkDir` are no longer unbounded. First failure: `grepNative's own walk over a 30-directory tree with a listing budget of 40 = (_, <nil>), want a *globBudgetError; grepWalk charges nothing to the budget`.

**Two bugs found during review that neither finding named, both fixed with tests:**
- All three best-effort walk callbacks (ignore discovery and both grep arms) treated every error as an unreadable entry to skip, so they would have **swallowed the new refusal** and let the oversized directory be read again by whatever walked next; ignore discovery would also have reported a set it never finished, silently under-excluding. Pinned by `TestGlobIgnoreDiscoveryStopsOnADirectoryWithTooManyEntries` and `TestGrepWalkStopsOnADirectoryWithTooManyEntries` (first failures: results of `([], <nil>)` and `("", <nil>)`).
- `globWalkFS.ReadDir` folded the refusal into `fs.ErrNotExist` via `hide()`, which doublestar skips silently.
- Also closed: `grepNative`'s `singleFile` ignore FS was still a bare `os.DirFS`.

**#939 and #938 are both redundant now** — close them as fixed by this PR.

Gates, all foreground with a private `GOCACHE`, 0 FAIL each: gofmt; `make vet` + `GOOS=windows make vet`; `go vet -tags evenerfuzz ./...`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test -race` on `./agent/execenv/` and `./agent/`; `-tags evenerfuzz -run Fuzz` on both; full `go test ./agent/`.

obra:
Round 6 pushed as `d03ba9fa5`. Verdicts on all eight items.

**1. Sandboxed grep budget untested — partly fixed, and the other half is not testable.** `TestSandboxedGrepWalkIsChargedToTheBudget` is added and pins the `budget.listing(true)` charge (verified by mutation: deleting `securepath_browse_fdops_unix.go:234-242` fails it). The entries-swallow half is **not** reachable, and I verified that myself by mutation on the plain path: deleting `grepWalk`'s `errors.AsType[*globBudgetError]` guard leaves `TestGrepWalkStopsOnADirectoryWithTooManyEntries` passing. Both grep arms call `loadIgnoreSet` over the same fsys and root **first**, and its skip set (dot-prefixed, plus masked on the sandboxed arm) is a strict subset of the walk's (which also skips gitignored), so ignore discovery always visits a superset and raises the entries refusal first. The guard only fires if a directory grows past the bound between the two passes. I kept both guards — without them that TOCTOU case silently skips an oversized directory and grep returns a partial result with no error — and rewrote their comments plus the misleading doc comment on `TestGrepWalkStopsOnADirectoryWithTooManyEntries`, which claimed to pin both callbacks and only pins ignore discovery. **#938 can close**, or stay open if you want the TOCTOU path made reachable; say which.

**2. Entries message — fixed.** It now says the count is one directory's entries against a per-directory bound, names the directory (and renders the walk root as "the base directory" rather than a bare `.`, which tells a model nothing), and advises matching inside it more specifically or pointing the base elsewhere.

**3. Peak-memory doc — corrected.** It claimed the walk frees each slice as it moves on; true between siblings, false for ancestors, since `fs.WalkDir` and doublestar hold the parent's slice live while recursing. Now states the real peak is cap × depth, with the ~105 bytes/entry → ~21-25 MB per listing arithmetic.

**4. Cleanups — all done.** Chunk loop shared as `readDirChunked` (no contortion: `*os.File` already satisfies `fs.ReadDirFile`); `flatEntriesFixture` now used instead of the open-coded loop; `globBudgetError.listings` renamed `count`.

**5. `Glob` discards `truncatedAt` — KEEP, evidence.** The only production callers of the `Glob` method are `agent/session_tools_shell.go:246` (the `else` arm of `if ge, ok := env.(execenv.GlobExcluder)`) and `agent/execenv/local.go:1706` (`Glob` delegating internally). Every other `.Glob(` in the tree is `filepath.Glob`/`fs.Glob`. `*LocalExecutionEnvironment` implements `GlobExcluder` (`local.go:1725`), so the real environment always takes the truncation-aware arm. No production caller can act on a truncated list. No change.

**6. Grep charged skipped directories — fixed.** Confirmed at `local.go` and `securepath_browse_fdops_unix.go`: the charge sat above the dot/gitignore/masked `SkipDir` checks in the same callback. Both now charge on the fall-through-to-descend path. Pinned by `TestGrepWalkDoesNotChargeSkippedDirectories` (20 dot-excluded + 10 gitignored vs 3 listed, budget 25; ignore discovery alone costs 14, so the test is red for item 6's reason and not because discovery is expensive). First failure: `refused: grep walk made 26 directory listings, past the budget of 25`.

**7. `boundedDirFS` hid `fs.StatFS` — fixed.** `Stat` forwards to `fs.Stat(b.FS, name)`. `ReadFile`/`Sub` deliberately not forwarded (their fallbacks behave identically; `ReadFile`'s fast path would fail on the same unreadable file). Pinned by `TestGlobMatchesAStatableUnreadableFile`; first failure: `Glob("secret.txt") = []`.

**8. `GlobExcluder` four returns — enumeration for your decision.** Implementations: `*LocalExecutionEnvironment` (`agent/execenv/local.go:1725`) and one test double `truncatingGlobEnv` (`agent/cov_s3_shelltools_test.go`). Callers: `agent/session_tools_shell.go:244`, `agent/execenv/local.go:1706`, and tests in `glob_cancel_test.go`, `glob_grep_scoping_test.go`, `glob_budget_test.go`. The interface is exported and `make lint-internal` (`make/linting.mk:76-84`) exists to keep `agent` and `execenv` externally importable, so it **is** a public surface and the change **is** source-incompatible for any external implementor — but there are none visible here, and the three-value form predates this PR. I recommend declining. No shim added; Jesse's explicit approval would be required for one.

Gates, foreground with a private `GOCACHE`, 0 per-test FAIL each: gofmt; `make vet` + `GOOS=windows`; `go vet -tags evenerfuzz ./...`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test -race` on `./agent/execenv/` and `./agent/`; `-tags evenerfuzz -run Fuzz` on both; full `go test ./agent/`.

obra:
Round 7 pushed as `adfa14f4a`.

**1. TOCTOU probes — added, and you were right.** Two probes, one per arm, both stubbing the existing walk seams and growing the directory after ignore discovery has already passed under the bound: `TestGrepWalkCarriesTheEntriesRefusalWhenADirectoryGrowsAfterIgnoreDiscovery` and `TestSandboxedGrepWalkCarriesTheEntriesRefusalWhenADirectoryGrowsAfterIgnoreDiscovery`. Both mutation-verified: deleting each guard's carry-through makes the corresponding probe fail with `want a *globBudgetError ... = <nil>`. No production change was needed. The three comments now say no *static* tree reaches the guard, only concurrent growth, and each names the probe that forces it.

**2. Sandboxed skip-charging — pinned.** `TestSandboxedGrepWalkDoesNotChargeSkippedDirectories`, mirroring the plain arm. Arithmetic: 3 real + 20 dot-prefixed + 10 gitignored directories, budget 25. Ignore discovery costs 14 (it cannot skip gitignored directories while it is still collecting the rules that would exclude them); the walk legitimately adds 4, total 18 — above discovery's 14 and below 25, so headroom cannot hide a regression. Verified by mutation: moving the charge back above the masked/dot/gitignore skips fails it with `grep walk made 26 directory listings, past the budget of 25`.

**3. Both comment nits — fixed.** The entries advice is now operation-aware: a grep cannot match inside an oversized directory more specifically, since its pattern is a regex applied after the walk has listed everything, so it is told to point the base at a smaller directory that does not contain it. And the listing budget's comment no longer claims memory is held to a small constant — it now says what bounds memory (match cap, O(depth) chain, per-directory entry cap) and that what those leave is the entry cap times depth, which raising the listing budget does not affect.

**Addendum, roborev Medium on `readDirChunked` — verified and fixed.** I confirmed the mechanism before changing anything: `cancelFS` guards `Open`/`ReadDir`/`Stat` at the filesystem level, so it tests the context once as a listing begins and never again while the chunk loop holds the directory handle — up to 49 chunks of pointless work at the shipped bounds. The loop now checks the context before each chunk and before returning, answering with the same `context.Canceled` the rest of the walk returns; no new error kind. Context reaches it through fields on `boundedDirFS` and `secureDirFS`, the way `cancelFS` and `globWalkFS` already carry one. Pinned by `TestGlobStopsReadingAChunkedListingOnCancellation`, written in the style of `TestGlobMatchesStopsOnContextCancellation` — it asserts promptness, not just the error, because the error alone was already correct: first failure was `listing kept reading after cancellation: read 40 of 40 entries in the directory, want at most 15`.

Gates, foreground with a private `GOCACHE`, 0 `^--- FAIL:` each: gofmt; `make vet` + `GOOS=windows`; `go vet -tags evenerfuzz ./...`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test -race` on `./agent/execenv/` and `./agent/`; `-tags evenerfuzz -run Fuzz` on both; full `go test ./agent/`.

obra:
Final round pushed as `0ee468269`, plus the PR body rewrite.

**Live-entry bound — added; no walk restructuring was needed.** The accounting fits points that already exist. All four walks a call can make (glob pattern walk, glob ignore discovery, grep walk, grep ignore discovery) reach the filesystem through `readDirChunked`, which already has the directory name, the entry count and the budget. And the release signal was already established in this file: `globWalkFS.push` prunes its ancestor chain on exactly the reasoning that the walk must have LEFT every non-ancestor to be listing this one. Reusing that rule gives release-on-leave for free, so `holdEntries` is ~20 lines of bookkeeping rather than a new walk structure. The predicate `push` open-coded is now `isGlobWalkAncestor`, shared by both.

Ceiling is **500,000 live entries, about 52 MB** at the ~105 bytes/entry figure — a couple of maximal directories, far above a real tree (a deep repository holds a few thousand live entries) and far below what threatens the process. The refusal is the typed error with its own `budgetLiveEntries` kind, worded as an aggregate and advised with the whole-call lever, with `dir` left empty because no single directory is at fault.

Tested with a 10-level nested tree whose per-directory counts (≤8) are all well under the per-directory cap of 20 but whose live total peaks at 52 against a ceiling of 30. It asserts `peakLiveEntries` — entries actually held — not just the returned result, so a fix that walked the whole tree and then complained cannot pass. First failure: `Glob over a 10-level tree (peak live so far: 52) with a live-entry budget of 30 = (49 matches, <nil>), want a *globBudgetError`.

The control matters as much: `TestGlobSucceedsOnAWideShallowTreeUnderTheLiveEntryCeiling` spreads the **same 60 total entries** across siblings instead of nesting them, so the live total never exceeds one sibling's 5 plus the root's 10. It passes before and after. A naive cumulative counter would fail it, and would break every large flat repository.

**Also in this commit** (the comment-and-test round you asked for): the four test comments that described fixed defects in the present tense now describe the invariant and what a regression would look like, matching the framing of the sandboxed twin; and `TestGlobBudgetErrorAdviceDependsOnTheOperationAndTheBound` pins the operation-aware advice split structurally — three inequalities across `advice()`/`entryAdvice()` and glob/grep, no sentence embedded in the test. Mutation-verified: collapsing `entryAdvice()` to one return fails it.

**PR body rewritten** to describe the PR as it now stands, with a dedicated caller-visible-changes section covering all five: grep budgeted on both arms and able to fail where it returned partial output; a single directory over 200,000 entries failing outright; the new live-entry ceiling; ignore discovery able to fail a call; and sandboxed listings now lexically sorted, changing sandboxed match order and grep visit order. `Closes #497` and `Closes #938` both present, session URL last.

Gates, foreground with a private `GOCACHE`, 0 `^--- FAIL:` each: gofmt; `make vet` + `GOOS=windows`; `go vet -tags evenerfuzz ./...`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test -race` on `./agent/execenv/` and `./agent/`; `-tags evenerfuzz -run Fuzz` on both; full `go test ./agent/`.

obra:
On roborev's two remaining findings on adfa14f4a, both declined; the third (aggregate memory) is fixed in the head that follows.

**`GlobExcluder` arity (execenv.go:178-183).** Enumerated twice, independently: one production implementation (`*LocalExecutionEnvironment`), one test double, one interface-typed caller (`session_tools_shell.go`). The interface is in a package kept importable by `make lint-internal`, so it is technically a public surface, but no implementation outside this repo is visible and the three-value form it replaces was introduced here as well. Adding a second capability interface to preserve a shape nothing implements would be a backward-compatibility shim, which this repo requires explicit approval for.

**`unix` build tag (glob_stat_unix_test.go:1).** Factually wrong. `unix` has been a Go-defined build constraint since Go 1.19 (`go/build` satisfies it on AIX, Android, Darwin, Dragonfly, FreeBSD, Hurd, illumos, iOS, Linux, NetBSD, OpenBSD, and Solaris); no `-tags unix` is required. It is also this repository's stated convention for Unix-only test files, and the sibling `_unix_test.go` files in this package demonstrably compile and run in CI, including the sandboxed tests this PR adds.

obra:
Superseded by a three-PR stack, per Jesse's request to split this and to fix roborev's arity finding rather than argue it.

- **#951** `fix/497-glob-budget-a` off main — the original #497 defects. `Closes #497`.
- **#955** `fix/497-glob-budget-b` off A — the memory bounds.
- **#956** `fix/497-glob-budget-c` off B — grep. `Closes #938`.

Every fix from this branch is carried over. The one real change: `GlobExcluder.GlobWithExclusions` is back to three return values. Truncation now rides on a caller-supplied budget through a new optional `GlobBudgeter` capability — the same move `GlobExcluder` itself was introduced as. Without a budget there is nowhere to report truncation, so that path refuses rather than returning a short list with a nil error.

Closing this in favour of the stack.

Issue #934, opened 2026-09-06
Allow explicit hub termination of incompatible or unresponsive session daemons


obra:
A hub upgraded to AppWire v4 cannot communicate with still-running v3 daemons. Those processes can retain the session reservation while the hub cannot stop them through AppWire. The same recovery gap applies to an unresponsive daemon.

Provide an explicit hub-controlled way to terminate the exact local daemon when normal shutdown cannot work, then let the user resume the saved session with the current launch binary. Keep this in a separate PR from the protocol-mismatch visibility and message-rejection fix.

Acceptance criteria:
- The action works without a successful daemon AppWire handshake.
- The UI explains that active turns and jobs may be interrupted and requires an explicit user action.
- Validate the target process identity and session ownership before signalling; stale rendezvous data or PID reuse must not kill an unrelated process.
- Confirm process exit before presenting the session as stopped or allowing a replacement; retain saved transcripts.
- Cover responsive shutdown, incompatible and unresponsive daemons, already-exited processes, stale identity, and failure to terminate with deterministic behavioral tests.

Observed September 6: the local hub and configured launch binary report evener-appwire-v4, while an open session's daemon reports evener-appwire-v3. Jesse explicitly requested this recovery capability as a separate PR.

Implementation investigation:
- `app_session_resume.go:shutdownThreadTolerateExited` is the existing normal shutdown path. Force stop should remain a separate explicit operation, serialized with resume and deletion through the existing per-session ownership lock.
- `spawn.go` only retains a child process handle during startup; its current kill calls handle failed startup. Successful detached daemons need a separate ownership/identity check after hub restart.
- `hubcore/roster_unix.go:processAlive` uses signal 0. That proves PID presence, not session ownership, and is insufficient authorization to terminate a process.
- Rendezvous entries contain session/workspace identity, PID, state directory, and an application timestamp. They do not contain an OS process-generation identity. The implementation must establish that identity independently and reject stale or ambiguous candidates before signalling.
- Add process inspection/signalling behind a platform boundary; exercise termination ordering, identity changes, and timeout outcomes with a deterministic fake at that OS boundary. Saved transcript reads and explicit resume remain separate from force stop.

PR #936, opened 2026-09-06
Keep incompatible running sessions visible and reject undeliverable messages

obra:
The v4 flag day in 7d9939d506c72f27930b2e4e857c1cb081e36d4c (#822, September 6) left still-running v3 daemons unreachable. A cold hub dropped their sidebar activity after initialization failed, while message recovery tried to spawn a replacement over the live session. Existing tests used matching protocol versions and missed hub-only upgrades.

This fix negotiates AppWire v5 to make daemon mutation-receipt authority explicit. Surviving v3 and v4 daemons require an explicit restart. Authenticated mismatches appear as `restartRequired`, retain saved transcripts and list visibility, and disable unavailable actions. Typed initialize mismatches remain actionable even with missing or stale rendezvous protocol metadata and invalidate previously cached routes. The TUI includes restart-required sessions in its attention count and warning indicators. Ownership is refreshed for subscribed reads, initial relay setup, and session actions, including actions without mutation IDs and protocol replacement at the same endpoint. Initialization failures carry their phase so definite daemon RPC rejections retain their original reason and mutation outcome even when unrelated discovery fails.

Roster discovery retains unresolved live ownership claims and reports incomplete discovery rather than treating it as evidence that a session can be replaced or deleted. Session writes and deletion check ownership under locks, including persisted descendant relationships. Nested relationships are verified against root ownership and ParentDelegateID in the root journal; deletion checks compatible and incompatible ancestor daemons and preserves uncertain descendants while allowing independent forks. Concurrent full scans and individual confirmations preserve newer ownership evidence. A freshly spawned or resumed daemon is verified through its exact endpoint when broader discovery fails; previously confirmed compatible owners can still serve reads and messages. Navigation preserves root, subagent, and job activity and invalidates when ownership claims change. Thread listings deduplicate live workspace aliases against saved sessions and retain saved-title search. Rename routes workspace aliases to the owning daemon, and restart-required listings accept session-only daemon identities.

Browser recovery keeps uncertain messages in IndexedDB until current daemon receipts prove delivery or absence. A failed mutation invalidates earlier authority; periodic discovery retries reconciliation without an immediate retry loop. Saved and descendant transcript projections cannot authorize retries. A retained live descendant preserves uncertain messages and offers owner navigation and Refresh until the parent releases it; the hub refuses Resume before spawning over verified ancestor ownership. Once ownership is released, explicit Resume can obtain the child daemon state. Successful root clear snapshots establish mutation authority for explicit fresh reconciliation without replaying old-instance input. Storage failures remain visible and cannot silently release pending sends. A failed blocking write retains a reconciliation obligation in memory, so a later enqueue cannot retry an uncertain record before fresh daemon state is reconciled. Clear responses fence earlier goal updates before publishing the replacement instance.

Behavioral tests cover older peers and compatible replacements over real WebSockets, same-endpoint protocol changes, saved transcripts, lost accepted receipts, root/subagent/job sidebar activity, incomplete ownership metadata, concurrent roster publication, descendant receipt authority, and IndexedDB failures and recovery. Retained delegates remain recoverable after ancestor deletion when the roster excludes every possible daemon owner; interrupted project cleanup is tested across a hub restart. A post-cleanup discovery failure preserves committed deletion outcomes and records stale roster/navigation diagnostics separately. Default tests use scripted external boundaries. A job-output snapshot replacement race exposed by CI has a deterministic regression test; stale-refresh tests now synchronize publication, and transcript performance uses benchmarks alongside deterministic projection checks.

Validation: the rewritten head `25be21cad` has exactly the same Git tree as validated `cba6a0b14`. The confirmed-incompatible saved-read regression reproduced both incorrect read exits before the fix; broader upgrade, ownership, and admission tests then passed under race (24.212s). Local review221 found no issues, and all CI checks passed before consolidation. All CI checks, including snapshot artifacts, passed on `25be21cad`. Current-head hosted combined review6600 completed with both reviewers passing and no issues found.

Explicit out-of-band process termination is implemented separately in #977 for #934. This PR adds no old-protocol compatibility or automatic process termination.

History is consolidated into two logical commits: backend behavior and regression tests, followed by frontend behavior and regression tests. Consolidation changed no final file contents.

### DIFF
--- a/agent/apilog_envelope_size_test.go
+++ b/agent/apilog_envelope_size_test.go
@@ -1,0 +1,41 @@
+package agent
+
+import (
+	"encoding/json"
+	"testing"
+)
+
+func TestAPILogSummarySizerMatchesTrimmedEnvelope(t *testing.T) {
+	envelope := apiLogReadEnvelope{TranscriptRef: "session", Source: "api-log", Records: []apiLogRecordSummary{
+		{RecordNumber: 1, Kind: "request", RequestModel: "model"},
+		{RecordNumber: 2, Kind: "response", ErrorClass: "multi\nline"},
+		{RecordNumber: 3, Kind: "request"},
+	}}
+	full, err := json.MarshalIndent(envelope, "", "  ")
+	if err != nil {
+		t.Fatal(err)
+	}
+	sizer, err := newAPILogSummaryEnvelopeSizer(envelope, len(full))
+	if err != nil {
+		t.Fatal(err)
+	}
+	for {
+		envelope.Meta.RecordsReturned = len(envelope.Records)
+		envelope.Meta.Truncated = len(envelope.Records) < 3
+		actual, err := json.MarshalIndent(envelope, "", "  ")
+		if err != nil {
+			t.Fatal(err)
+		}
+		size, err := sizer.envelopeSize(envelope)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if size != len(actual) {
+			t.Fatalf("%d records: predicted=%d actual=%d", len(envelope.Records), size, len(actual))
+		}
+		if len(envelope.Records) == 0 {
+			break
+		}
+		envelope.Records = envelope.Records[1:]
+	}
+}
--- a/agent/delegate_resume_ownership_test.go
+++ b/agent/delegate_resume_ownership_test.go
@@ -1,0 +1,71 @@
+package agent
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"primeradiant.com/evener/llm"
+)
+
+func TestRetainedDelegateReservesSessionUntilOwnerReleases(t *testing.T) {
+	root, client, _ := newDelegateResourceBootstrapSession(t)
+	logger, err := llm.NewSessionAPILogger(root.stateDir)
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { _ = logger.Close() })
+	root.cfg.AcquireSessionOwnership = logger.ReserveSession
+	adapter := newTask6FrozenDescriptorAdapter()
+	client.Register(adapter)
+	t.Cleanup(adapter.releaseRun)
+	result := root.createDelegate(context.Background(), delegateArgs{Task: "finish the assigned work", DelegationAllowance: new(0)})
+	if result.Err != nil {
+		t.Fatal(result.Err)
+	}
+	select {
+	case <-adapter.entered:
+	case <-time.After(10 * time.Second): // TRIPWIRE: the scripted provider must receive the child request.
+		t.Fatal("child never reached provider")
+	}
+	child := root.subagents.get(result.ChildSessionID)
+	child.mu.Lock()
+	done := child.done
+	child.mu.Unlock()
+	contender, err := llm.NewSessionAPILogger(root.stateDir)
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { _ = contender.Close() })
+	if err := contender.ReserveSession(result.ChildSessionID); !errors.Is(err, llm.ErrAPILogTargetLocked) {
+		t.Fatalf("running child ownership = %v", err)
+	}
+	adapter.releaseRun()
+	drainDone := make(chan struct{})
+	defer close(drainDone)
+	go func() {
+		for {
+			select {
+			case <-adapter.entered:
+			case <-drainDone:
+				return
+			}
+		}
+	}()
+	select {
+	case <-done:
+	case <-time.After(10 * time.Second): // TRIPWIRE: await the actual child completion.
+		t.Fatal("child never completed")
+	}
+	if err := contender.ReserveSession(result.ChildSessionID); !errors.Is(err, llm.ErrAPILogTargetLocked) {
+		t.Fatalf("retained child ownership = %v", err)
+	}
+	root.Close()
+	if err := logger.Close(); err != nil {
+		t.Fatal(err)
+	}
+	if err := contender.ReserveSession(result.ChildSessionID); err != nil {
+		t.Fatalf("released child ownership = %v", err)
+	}
+}
--- a/agent/internal/jobstore/output_snapshot.go
+++ b/agent/internal/jobstore/output_snapshot.go
@@ -47,6 +47,7 @@ type outputSnapshotObservation struct {
 	outputObserved  bool
 	outputExists    bool
 	retainedBytes   int64
+	outputInfo      os.FileInfo
 	pendingObserved bool
 	pendingExists   bool
 	pending         string
@@ -287,6 +288,7 @@ func observeOutputSnapshot(fs afero.Fs, path string) (outputSnapshotObservation,
 	observation.outputObserved = true
 	observation.outputExists = true
 	observation.retainedBytes = info.Size()
+	observation.outputInfo = info
 
 	metaPath := outputMetaPath(path)
 	// OutputStore publishes pending metadata before rewriting a capped file and
@@ -307,6 +309,18 @@ func observeOutputSnapshot(fs afero.Fs, path string) (outputSnapshotObservation,
 }
 
 func (after outputSnapshotObservation) changedFrom(before outputSnapshotObservation) bool {
+	// A prune can return to the same retained size while the reader still holds
+	// a hash of the expanded file. Metadata publication need not advance during
+	// that interval, so also fence the file generation and in-place appends.
+	if before.outputInfo != nil && after.outputInfo != nil {
+		if !before.outputInfo.ModTime().Equal(after.outputInfo.ModTime()) {
+			return true
+		}
+		if before.outputInfo.Sys() != nil && after.outputInfo.Sys() != nil && !os.SameFile(before.outputInfo, after.outputInfo) {
+			return true
+		}
+	}
+
 	if after.outputObserved && (after.outputExists != before.outputExists || after.retainedBytes != before.retainedBytes) {
 		return true
 	}
--- a/agent/internal/jobstore/output_snapshot_test.go
+++ b/agent/internal/jobstore/output_snapshot_test.go
@@ -1030,3 +1030,79 @@ func (f *snapshotCloseHookFile) Close() error {
 	}
 	return err
 }
+
+// The file size returns to its cap while pending/final metadata remain unchanged
+// across both observations. The hash in between belongs to the expanded inode.
+func TestReadOutputWindowSnapshotDetectsReplacementDuringHash(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "job.log")
+	fs := &snapshotHashReplacementFS{snapshotAppendPruneFS: newSnapshotAppendPruneFS(path), finalPublication: make(chan struct{}), allowFinalPublication: make(chan struct{})}
+	store, err := createOutputFsWithSync(fs, path, 4, true)
+	if err != nil {
+		t.Fatal(err)
+	}
+	appendOutput(t, store, "AAAA")
+	fs.armInitialRetainedSize()
+	fs.fencePublication.Store(true)
+	type result struct {
+		snapshot OutputWindowSnapshot
+		err      error
+	}
+	snapshotDone := make(chan result, 1)
+	go func() {
+		snapshot, err := readOutputWindowSnapshotFs(fs, path, 4, 4)
+		snapshotDone <- result{snapshot, err}
+	}()
+	<-fs.initialRetainedSizeCaptured
+	appendDone := make(chan error, 1)
+	go func() { _, err := store.Append([]byte("BBBB")); appendDone <- err }()
+	defer func() {
+		fs.releaseInitialMetadataValidation()
+		fs.releaseOutputReplacement()
+		close(fs.allowFinalPublication)
+		if err := <-appendDone; err != nil {
+			t.Error(err)
+		}
+		if err := store.Close(); err != nil {
+			t.Error(err)
+		}
+	}()
+	<-fs.pendingPublished
+	fs.captureHash.Store(true)
+	fs.releaseInitialMetadataValidation()
+	got := <-snapshotDone
+	if got.err != nil {
+		t.Fatalf("snapshot across replacement: %v", got.err)
+	}
+	if string(got.snapshot.Content) != "BBBB" || got.snapshot.Start != 4 || got.snapshot.TotalBytes != 8 {
+		t.Fatalf("snapshot=%+v", got.snapshot)
+	}
+}
+
+type snapshotHashReplacementFS struct {
+	*snapshotAppendPruneFS
+	captureHash           atomic.Bool
+	fencePublication      atomic.Bool
+	finalPublication      chan struct{}
+	allowFinalPublication chan struct{}
+	finalOnce             sync.Once
+}
+
+func (fs *snapshotHashReplacementFS) Open(name string) (afero.File, error) {
+	file, err := fs.snapshotAppendPruneFS.Open(name)
+	if err == nil && name == fs.path && fs.captureHash.Swap(false) {
+		return &snapshotCloseHookFile{File: file, afterClose: func() error {
+			fs.releaseOutputReplacement()
+			<-fs.finalPublication
+			return nil
+		}}, nil
+	}
+	return file, err
+}
+
+func (fs *snapshotHashReplacementFS) Rename(oldpath, newpath string) error {
+	if newpath == outputMetaPath(fs.path) && fs.fencePublication.Load() {
+		fs.finalOnce.Do(func() { close(fs.finalPublication) })
+		<-fs.allowFinalPublication
+	}
+	return fs.snapshotAppendPruneFS.Rename(oldpath, newpath)
+}
--- a/agent/session_active_turn_test.go
+++ b/agent/session_active_turn_test.go
@@ -5,7 +5,6 @@ import (
 	"strings"
 	"sync"
 	"testing"
-	"time"
 
 	"github.com/spf13/afero"
 
@@ -32,11 +31,6 @@ func TestGoalContinuationTurnCarriesItsNameOnTheOpeningEvent(t *testing.T) {
 
 	var mu sync.Mutex
 	var continuations []events.GoalContinuationData
-	// The drain ConsumeEventsLossless registers delivers on its own goroutine
-	// (session_events.go), so nothing guarantees it has run by the time
-	// ProcessInputKind returns. Wait for the event rather than asserting on a
-	// slice the consumer may not have appended to yet.
-	sawContinuation := make(chan struct{}, 1)
 	drained := make(chan struct{})
 	s.ConsumeEventsLossless(func(ev events.SessionEvent) {
 		if ev.Kind != events.EventGoalContinuation {
@@ -49,26 +43,16 @@ func TestGoalContinuationTurnCarriesItsNameOnTheOpeningEvent(t *testing.T) {
 		mu.Lock()
 		continuations = append(continuations, data)
 		mu.Unlock()
-		select {
-		case sawContinuation <- struct{}{}:
-		default:
-		}
 	}, func() { close(drained) })
 
 	if _, err := s.ProcessInputKind(context.Background(), "keep going", nil, EntryContinuation); err != nil {
 		t.Fatalf("ProcessInputKind(EntryContinuation): %v", err)
 	}
 
-	select {
-	case <-sawContinuation:
-	// TRIPWIRE: the event is emitted synchronously by ProcessInputKind above
-	// and delivered by the in-process drain goroutine, so it normally arrives
-	// in well under a second. 30s only fires on a genuine delivery hang, not
-	// scheduler contention under a loaded suite.
-	case <-time.After(30 * time.Second):
-		t.Fatalf("timed out waiting for the EventGoalContinuation opening event")
-	}
-
+	// ProcessInputKind finishes the turn before the asynchronous consumer
+	// necessarily sees its events. Join the drain before inspecting them.
+	s.Close()
+	<-drained
 	mu.Lock()
 	defer mu.Unlock()
 	if len(continuations) != 1 {
--- a/agent/session_jobtree_drain_backoff_test.go
+++ b/agent/session_jobtree_drain_backoff_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"sync/atomic"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"primeradiant.com/evener/agent/internal/jobstore"
@@ -208,6 +209,10 @@ func TestDrainIdleBackoffWakeResetsToFullRate(t *testing.T) {
 // exactly one notification turn for it.
 func TestDrainBackoffFinalKickDeliversUnreachableChildPending(t *testing.T) {
 	t.Parallel()
+	synctest.Test(t, testDrainBackoffFinalKickDeliversUnreachableChildPending)
+}
+
+func testDrainBackoffFinalKickDeliversUnreachableChildPending(t *testing.T) {
 	sess := newSession(t)
 	seedWatchSendResidue(t, sess)
 	jm := sess.jobManager
@@ -248,7 +253,7 @@ func TestDrainBackoffFinalKickDeliversUnreachableChildPending(t *testing.T) {
 		_, _ = sess.drainJobTreeWith(ctx, recheck, kick, process)
 	}()
 
-	// Phase 1: drive passes 0..drainIdleBackoffFullRatePasses (17 sends).
+	// Phase 1: finish the full-rate prefix and one skipped-kick pass.
 	// The drain runs its first pass (pass 0) before parking, so 17 sends
 	// release passes 1..17: passes 0..16 kick at full rate (17 kicks) and
 	// pass 17 (quietPasses=17, 17%4!=0) is the first pass that SKIPS its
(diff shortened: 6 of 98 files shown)


obra:
Addressed the hosted review of `cd2bfee` in cd6885d32:

- Incomplete subagent ancestry now remains unresolved even when `JobTreeRootSessionID` is empty. The regression reproduced successful rename and a false `notAccepted` receipt before the fix.
- Navigation now checks unconfirmed daemon claims, including root and persisted delegate rows, and returns an ownership error instead of publishing an actionable snapshot. Both cases have regressions.
- The torn-journal write-bypass finding does not match the call graph. `SessionOwnsDelegate` has one production caller, the restart ownership guard. A matching descriptor returns true, which causes that guard to reject mutations; it does not permit them. A new torn-tail regression verifies rename and fork remain blocked, no metadata is written, and queue retries retain unknown/blocked outcomes. Missing descriptor evidence with a torn tail already returns an error.

Focused upgrade race tests, the full hub short suite, and affected lint pass. The preceding commit also passed the full merge gate and local full-range review; verification and review are being repeated for the updated head.

obra:
Fixed the hosted missing-directory finding from `58f7095` in d8bf2af6c. Strict listing now rejects a configured rendezvous directory that is missing, preserving the previous ownership snapshot. An existing empty directory remains a valid empty scan. Root and delegate RPC regressions reproduce successful rename/fork writes before the fix, verify unknown blocked receipts afterward, and cover recovery after the directory returns and the daemon registration is removed.

The same commit fixes local review 112's independent-confirmation race, with per-PID publication freshness and regressions for partial/partial and full/partial overlap. Focused upgrade race tests, full rendezvous/hubcore race tests, full hub short tests, and affected lint pass.

obra:
Addressed the high-severity stable workspace reuse finding in `c9dce4920`. Resume now uses the existing workspace-aware live-daemon lookup and returns the current daemon identity. The real RPC resume-and-steer test now covers a daemon with a changed session ID and preserved workspace reference. Before the fix, it attempted a replacement spawn; afterward it reuses the daemon and delivers the steering mutation. Focused race tests, the full hub suite, and hub lint pass.

The medium fresh-read finding does not reproduce on the reviewed head. The production source registry always registers the local daemon source (`app_rpc.go:25`); an absent route returns `SessionUnavailable` (`internal/appsource/local_daemon.go:826`). The read failure path refreshes the roster before constructing the saved projection (`app_rpc.go:338-349`). I extended the protocol-upgrade RPC test to publish an incompatible rendezvous entry only after hub initialization, without populating the roster. Both the original-ID and cleared-ID cases already return `restartRequired`, retain saved turns, and disable writes on `d8bf2af6c`. These regression cases are retained in `c9dce4920`; no additional global refresh was added to successful reads.

The canonical gate and a new full-range local review are running for the updated head.

obra:
Addressed the high and low findings in `df1ba0fc9` and `eba98ae8d`.

- Restart obligations now survive restartRequired and notLoaded snapshots and clear only after reconciliation with a compatible loaded daemon. A regression reproduces an outbox record discovered after the saved snapshot escaping protection before the fix. The same obligation is observable by the session UI, keeping the explicit Resume action available even with no pending records. The empty-outbox UI regression also failed before the fix. All 324 store/session tests, the frontend gate, and all five browser guards pass.
- Empty ownership targets now return no owner without loading metadata. The regression reproduced the invalid-session-ID ownership error; focused race tests, the full hub suite, and hub lint pass after the guard.

The repeated cold-roster read finding remains contradicted by the retained real-RPC regression in `TestHubProtocolUpgradePreservesTranscriptAndRejectsUndeliverableMessages`, including its `cached=false` cases. Those publish the incompatible rendezvous entry after hub initialization and receive restartRequired on the first read. The handler refreshes on the local source SessionUnavailable path before constructing the saved projection (`app_rpc.go:338-349`). The earlier response describes that production path and its test evidence: https://github.com/prime-radiant-inc/evener/pull/936#issuecomment-5565184666

The canonical gate and a fresh full-range local review are running for `df1ba0fc9`.

obra:
Fixed in 1038b5c96. A fresh start now confirms and publishes its endpoint when the roster has not admitted it, covering both a failed first probe and a failed full listing. The real-WebSocket regression fails before the fix and now verifies roster ownership, subsequent thread/read, and follow-up turn delivery without another spawn. Focused race tests, the full hub suite, and hub lint pass. Fresh canonical validation and reviews are running on this head.

obra:
Fixed in bf67815bd. Hub startup now creates the configured runtime directory synchronously before constructing request handling and reading ownership. Previously the background watcher created it after the initial read, leaving a startup ordering gap. A lifecycle regression verifies fresh discovery before requests and rejects an unusable path; it failed before the fix and passes under the race detector, with the full hub suite and lint also passing. A directory disappearing after startup remains an ownership error, preserving live claims rather than treating lost discovery data as proof of no daemon.

obra:
Fixed in cd087d4cd. Direct registration now derives active and idle retained children from delegate diagnostics and excludes closed children. Route admission compares endpoint, protocol, thread/session/workspace identity, instance, token, and start time in addition to PID, so a same-PID stale instance is replaced. The regression covers working initial and follow-up sends, retained child activity, and stale instance replacement through real RPCs. Focused race tests, the full hub suite, and lint pass. The same comparison also retains changed ownership claims when a probe fails, addressing local review 121.

obra:
Addressed the transcript attention finding in e761d19b3: restartRequired now upgrades the unseen-content pill to needs-you. The scroll-hook test fails for restartRequired before the fix and passes afterward; the full frontend gate passes.

The seeded-roster finding targets NewRosterWithEntries, a synthetic test helper with no production callers. Production rosters have a configured discovery directory and preserve ownership on incomplete reads. I tested the proposed no-op refresh, but it breaks TestSessionDeleteRefreshesRosterAndBustsTreeMemo and TestProjectDeleteRefreshesRosterAndBustsTreeMemo, which intentionally refresh away their seeded crash fixtures. I have left that helper behavior unchanged; the upgrade/mutation regressions exercise configured rendezvous directories and real discovery.

Local review 123 separately identified that saved delegate snapshots can carry active/idle status without authoritative mutation receipts. That finding remains open and is the next repair.

obra:
Fixed local RoboRev 123 in 8f89b3a00. A live daemon read explicitly marks mutation state as authoritative. Saved active/idle delegate snapshots preserve activity but cannot prove a send absent or clear restart protection. Known transcript identities can still settle accepted sends.

Both active and idle outbox regressions failed before the fix and now pass. A hub RPC test exercises failed live reads followed by saved delegate fallback, and a daemon RPC test checks the positive authority contract. Focused race tests, full server/hub suites, the frontend gate, and lint pass. Fresh whole-PR review and the canonical gate are underway.

obra:
Addressed both local review 124 findings in 45890e763. Manual retry and its button now require per-ref mutation authority from the current connection; authority clears on replacement, reconnect, and disconnection. The active/idle saved-delegate reload tests reproduced the bypass after hydration settled and now pass.

The receipt-authority requirement changes the wire contract, so negotiation is now AppWire v5. Surviving v3/v4 daemons explicitly require restart. Real RPC upgrade regressions cover both versions with cold/cached rosters and cleared identities. The canonical gate and all five browser guards pass for these changes; fresh review and CI are pending.

obra:
Addressed hosted review 5528 in 51a12e533, 2ba3f326f, and 4e1035034. Real deletion regressions reproduced state loss under both unconfirmed claims and failed initial scans. The shared deletion guard now blocks those cases, and roster scan errors remain published until a newer complete scan succeeds. Older scans and partial confirmations cannot clear global uncertainty. Confirmed daemon reads/delivery remain available; absence-based recovery, navigation, and deletion retain the ownership checks. Focused race tests and the canonical gate pass.

Local review 126 also identified stalled receipt recovery for saved active/idle delegates. 801b42b1d continues periodic reads for blockedUnknown records lacking authority. Tests cover accepted receipts without retransmission and proven absence with one resend. ac1ebacb8 synchronizes those tests on completed storage transactions. The full frontend gate and all five browser guards pass for these changes. Fresh final-head CI and reviews are pending.

obra:
Addressed both findings from hosted review 5546:

- Subscribed local read failures now refresh ownership before the fallback decision. A regression test starts with a confirmed compatible route, replaces it with an incompatible daemon, and verifies preserved transcript content plus restart-required state with send disabled.
- Resume now checks actual publication of the exact spawned entry after discovery, including successful scans with failed individual probes. A direct read validates the daemon identity and publishes its route. Regression cases cover failed inventory, malformed discovery, an already confirmed owner, read failure, and wrong identity; successful cases also verify follow-up reads and message delivery.

Focused read/rejoin/protocol-upgrade and resume/recovery race tests pass. Full-gate and current-head review results are recorded in the PR description.

obra:
Addressed all three findings from hosted review 5556:

- A real same-endpoint protocol replacement reproduces the initialize rejection. Local reads now refresh ownership after either transport unavailability or an invalid-request rejection, then return the saved transcript with verified restart-required status. Unrelated application read failures retain their existing behavior.
- Shutdown, model, vision-model, compact, and goal actions now use an ownership guard independent of durable mutation IDs. Each has a real hub/daemon regression for a protocol replacement at the same endpoint. Read-only deletion locking remains separate.
- Local thread lists preserve an explicit restart-required status and clear mutation capabilities. A ListThreads regression verifies the status and capability contract.

Focused read/rejoin, protocol-upgrade, session-action, and local-list race tests pass. Full-gate and current-head review results are recorded in the PR description.

obra:
Addressed both findings in efdf17ded and 013fb978a, with cached-route handling completed in 1b88a9807.

Typed initialize protocol mismatches now produce `restartRequired` independently of rendezvous protocol metadata. The older-metadata requirement remains for generic `InvalidRequest` errors. Real-WebSocket regression cases cover both response types with missing, current, and older advertised versions. Typed mismatches also refresh cached routes for subscribed reads, sends, and session actions; those tests verify saved transcripts and blocked uncertain outcomes.

The TUI includes live restart-required sessions in the needs-you count and renders a warning-colored filled status marker. The test excludes project rollups and nonlive rows from the count.

`make merge-approval-gate` and the focused Go race tests pass on 1b88a9807. Whole-PR local review 139 is running; current-head CI and hosted review will follow this push.

obra:
Addressed both findings in a4210f59a and 1331adc26.

Nested ownership verification reads the root-owned journal and checks each child's ParentDelegateID against its immediate parent's descriptor. The hub fixture now uses the same owner and parent fields as ReserveCreate. Regression tests cover valid direct/nested edges and reject root-to-grandchild shortcuts and sibling ownership.

Deletion now shares the ancestry lookup, including descriptor verification for compatible ancestor daemons. Both project preflight and the locked per-target checks refuse live or uncertain ownership. Request cancellation propagates through journal reads. Tests reproduce and prevent deletion of direct/nested delegate metadata, transcripts, and API logs under active, incompatible, and unconfirmed ancestor daemons. Independent forks remain deletable.

Focused agent ownership and hub deletion/upgrade race tests pass. The full merge approval gate passed on 1331adc26. Local review 141 found only a misleading protocol diagnostic; fd255a9b9 removes the assertion based on stale rendezvous metadata, and focused protocol tests pass. All three commits are pushed; current-head CI and review are pending.

obra:
Addressed both findings in 439823c81 and feb641b2f.

Applying a clear snapshot now invalidates the existing goal-response generation before publishing the new instance. The regression delays a goal response across a successful clear, checks both tracked models remain cleared, and verifies a subsequent goal update still commits. All 281 thread-store tests pass.

Completed deletion now remains successful when post-cleanup roster or navigation refresh fails. Refresh failures are logged separately, and the roster retains its ownership error. A filesystem fault after artifact removal reproduces the incorrect session result and incomplete project-deletion state; both now preserve their committed outcome. Focused deletion race tests pass. The full merge approval gate and all five browser guards pass on feb641b2f. Both fixes are pushed. Whole-PR local review 144 is running; current-head CI and hosted review are pending.

obra:
Addressed hosted review 5572 in 32d7a2c16. Thread listing now includes the source-scoped canonical workspace identity when suppressing saved duplicates and looking up saved metadata. A real hub RPC test with an incompatible peer reproduces duplicate rows before the fix and verifies one restart-required row plus saved-title search afterward. Focused thread-list race tests and the full merge-approval gate pass.

obra:
Both hosted review 5576 findings are addressed in 3742a656d. Rename now uses canonical workspace ownership to reach the live daemon instead of writing saved metadata. Local daemon listings accept session-only identities and use the shared ID fallback. Regression tests reproduce both failures and verify daemon forwarding, metadata preservation, and restart-required identity. Focused race tests and the full merge-approval gate pass.

obra:
The `c7365fb` clear-authority finding above was fixed in [f78c5a3c4](https://github.com/prime-radiant-inc/evener/commit/f78c5a3c48a52ef55ec2724db82ab88fff464c3f), which is included in the current head `660419c94`.

The server now marks successful root clear snapshots mutation-authoritative, and `applyClearResponse` updates `mutationAuthorityRefs`. Behavioral regressions cover clear/replayed receipts and a blocked old-instance message arriving while clear is in flight. The latter retains the message and requires explicit fresh reconciliation without replaying stale input.

The fix was recorded on RoboRev job 5712 but was missing a visible follow-up here. Current-head GitHub CI passes and local whole-PR RoboRev 173 reports no issues. The latest hosted review remains pending; the older combined-review comment is not a review of the current head.

obra:
Fixed hosted review 5858 in [082d585f](https://github.com/prime-radiant-inc/evener/commit/082d585f3921afbd012cdc7f29153f8e4d935a4d).

When the past index cannot resolve a session, ownership lookup now checks its metadata in the actual project directories under the configured state root. Unreadable, duplicate, or unresolved possible ownership remains uncertainty, rather than permission to spawn. Confirmed owner absence and independent forks retain their existing behavior.

A real production-layout regression reproduced a retained child launching beneath an incompatible parent with nil or missing index data. Tests now block resume, mutation and deletion appropriately, and cover independent forks, owner release and malformed discovery. Focused race tests passed three repetitions; the complete hub short suite and lint pass; independent scoped review is clean. Current-head CI and whole-PR review are running.

obra:
Local review 190 is fixed in `643864380`, included in current `04636877e`. A never-attempted outbox intent no longer keeps a recovered descendant blocked merely because that descendant cannot authoritatively report mutation receipts. Only explicit `attempted: false` records bypass that receipt requirement; uncertain or missing attempt state remains protected.

Active and idle descendant regressions reproduced the stuck obligation and now deliver the queued intent once. All 290 thread-store tests pass, including uncertain-delivery protections. The latest main conflict preserved the complete continuation-event drain join, verified by ten race repetitions. The combined canonical gate, browser guards and new whole-PR reviews are running; hosted review remains pending.

obra:
Hosted review6032's API-log sizing finding is a false positive, verified by the regression in `1c7ad0238`. For three through zero records, the computed envelope size equals the actual indented JSON size, including metadata changes and multiline values. `apiLogSummaryRecordSeparator` already includes the four spaces before subsequent records, so adding another per-record indentation charge would double-count them. Production sizing remains unchanged.

Local review192's independent-fork ownership finding is fixed in `285952753`: non-subagent independent activity roots stop provenance traversal, while delegate/shared-job-tree ancestry remains verified. A real fork with deleted parent metadata and an unrelated incompatible daemon reproduces the refusal and now permits resume, mutation admission and deletion ownership checks. Focused ownership/restart/delegate tests pass three race repetitions. The new integrated gate and whole-PR reviews are running.

obra:
Local review194 is fixed in `91de06125`. Resume now checks the resolved `sessionId`, which is also the target used by the lock, ownership lookup and launcher. A request with an independent fork ref and an incompatible parent as explicit session target previously invoked the spawner; the regression now verifies refusal before any launch. Focused resume and ownership tests pass three race repetitions; scoped lint reports zero issues.

A fresh whole-PR review is queued. #977 remains at `88ff5ed73` while its already-running canonical gate finishes; the fix will be integrated afterward to keep that verification tied to its original head.

obra:
Hosted review6113: the directory-skipping suggestion would remove an intentional ownership fence, so I am retaining strict behavior. `ListStrict` treats a PID-named path as an ownership claim even when the path has become unreadable or is replaced by a directory. Skipping it would allow an incomplete scan to certify that the owner is absent. Ordinary `List` still skips directories.

This is covered by `TestListStrictRejectsUnreadableOrInvalidClaims/directory` and the real hub `TestHubUpgradeDoesNotTreatFailedProbeAsAbsentOwner/unreadable` cases for root and delegated sessions. The strict tests and hub ownership/upgrade regressions passed, including the hub tests under the race detector. Incomplete scans publish ownership uncertainty and retain the last snapshot; they do not authorize mutation or restart. The other two findings in6113 remain under repair.

obra:
Hosted review6113's two actionable findings are fixed in `3f22cdb63` and `79f8f68d7`.

- After Clear replaces the daemon instance, descendant reads and start notifications retain the stable owning workspace ref. The real clear/descendant regression failed with local:new and now returns local:old, which resolves to the replacement instance. Full server race tests and scoped lint pass.
- Explicit Refresh retains one readiness deadline across client rewires. A fake-clock regression drives repeated ready/rewire races; the former recursion remains pending past the original deadline, while the fix returns ClientNotReadyError without sending to an unready client. All291 store tests, TypeScript and Biome pass.

The strict directory behavior is intentionally retained for the ownership reasons and passing regressions documented in the preceding comment. Updated79f8f68d7 is undergoing fresh review and validation; no merge readiness is claimed.

obra:
Hosted review6218's cold-read finding overlooks the fallback path following live lookup failure. The registry installs the local source; an empty roster produces SessionUnavailable from local_daemon.go, and app_rpc.go refreshes ownership and rechecks restart status before falling back to saved data. The saved projection then marks restartRequired and removes mutation capabilities.

Existing TestHubProtocolUpgradePreservesTranscriptAndRejectsUndeliverableMessages writes the incompatible rendezvous after hub initialization and before the first read, including cleared aliases. That matrix and TestHubUpgradeClassifiesUncachedDaemonOwnership pass normally and five race repetitions. No discovery change is needed.

The receipt finding is also being checked against the separate durable journal and original-instance fence; proof and the clarification of misleading comments will follow separately.

obra:
Hosted review6218's duplicate-send claim is not supported by the full retry path. Snapshot IDs are projection evidence, not the deduplication ledger. The separate durable mutation journal retains terminal/applied records and replays the same mutation ID and payload without repeating its effect. After Clear, the original expectedInstanceId rejects the retry before executing a mutation on the replacement instance.

Commitdc51a4b3d corrects comments that wrongly described a missing snapshot ID as proof it was never journaled. Runtime behavior is unchanged. The strengthened frontend integration test covers an older bounded transcript and a cleared instance: real IndexedDB and dispatcher retain the exact original ID, instance, queue revision and input; a replay receipt settles, while rejection preserves the original payload in recovery. All337 store/dispatcher/outbox tests, TypeScript and Biome pass.

The server's response-loss/replay table, terminal-rejection replay, and old-instance-after-Clear tests pass under the race detector (2.724s), proving the separate daemon guards. Combined with the cold-read evidence in the preceding comment, neither6218 finding requires a runtime fix. Fresh review is requested for the proof/comment commit.

obra:
CI34197005175 failed because the merged main introduced a bare time.After poll in TestDrainBackoffFinalKickDeliversUnreachableChildPending, caught by TestNoBareWallClockDeadlineInAgentTests. PR936 now integrates mainffd73a31b and fixes the test in52308c849.

The test waits with synctest for the drain to park before seeding state, then checks quiescence directly. It removes the deadline/poll loop instead of weakening the audit. The audit reproduced red; the audit and backoff tests pass10 normal and25 race repetitions, with zero agent lint findings. A temporary Go overlay disabling the production final kick still makes the behavioral regression fail with zero delivered turns.

Fresh CI and whole-PR review are running on52308c849.

obra:
Local review203 is fixed in43f5e63e3. Confirmed ancestor and job-tree ownership now requires the daemon's current session identity, rather than treating a surviving workspace alias as ownership of the old root journal. Direct actions through the stable workspace alias continue routing normally; unconfirmed ownership remains conservative.

The behavioral regression reproduced blocked Resume/deletion for a historical child after parent Clear, with both compatible and incompatible replacement daemons. After the fix the released child can resume/delete while current descendants remain protected. A real API-log reservation still blocks deletion during the publication-before-close window; the CLI failed-reservation regression verifies refusal before session restoration.

The regression passes10 repetitions; related ownership/protocol tests and the CLI reservation test pass under race, and scoped lint has zero issues. The fix is integrated into #977 at26fb1d4b7. Fresh whole-PR reviews and combined canonical validation are running; browser code is unchanged from the previous all-five-guard pass.

obra:
Hosted review 6275 triage: the stable-workspace-alias cold-read finding is being reproduced and investigated.

The strict-directory finding is intentional behavior, already covered by TestListStrictRejectsUnreadableOrInvalidClaims/directory. Strict listing ignores non-PID names, but a PID-named path such as 123.json that cannot be read as a claim must preserve ownership uncertainty. Silently skipping it would let ownership checks treat an unreadable claim as an absent daemon. The suggested unconditional directory skip would weaken that contract. The ListStrict tests pass three repetitions under race. No change is planned for this item; the medium alias finding remains open.

obra:
Hosted6275 stable-ID read finding is fixed in 3e3290c8d, integrated into PR977 at 582e58032. Metadata fallback matches the canonical local reference as well as the current session ID, while retaining explicit ref precedence and strict protocol rejection.

A real WebSocket regression reproduced the failure with an incompatible daemon appearing after initialization, no past index and a cold read by stable workspace ID. After the fix that read and a subsequent subscribed read return the current session identity, stable reference and restart-required metadata with mutations disabled. New and related tests pass five repetitions under race; scoped lint has zero issues.

The strict-directory item is intentional, with evidence in the preceding triage comment. Both fixes are pushed and fresh whole-PR reviews are queued (local209/210); the integrated canonical gate is running. A separate empty-roster subscription admission boundary is still being assessed for actual client impact, so this is not a completion claim.

obra:
The separate first-read subscription admission check is intentional and needs no code change. Production populates the roster before serving and refreshes discovery on filesystem events and a five-second watch tick. The frontend sends subscribed reads directly and retries failures while preserving subscribe=true. The synthetic fixture had no roster watcher, so it could remain unresolved indefinitely; production discovery resolves that condition. Existing tests for unresolved-subscription refusal and retryable classification pass.

Current 3e3290c8d has green CI and local whole-PR review209 reports no issues. The integrated PR977 head582e58032 also passes the full canonical gate. Hosted current-head reviews remain pending.

obra:
Hosted6302 is addressed in 76643a246 and de3a945fa.

The navigation failure was valid: Capture discarded retained roster and saved metadata on any strict ownership scan error. It now preserves those known read results, disables local rename while ownership is uncertain, and still propagates cancellation. Strict daemon mutation/deletion checks are unchanged. The real roster regression fails against the previous Capture implementation, then passes under race; it covers corrupted discovery, retained row/favorite decisions, disabled rename, cancellation and repaired discovery.

The cancellation finding cites paths that pass an empty mutation ID. withDeletionTargetOwnership therefore skips both ownership lookup and RefreshAndWait; the callbacks are registry lookup and Broadcast, and deletionFenceError uses in-memory state. Existing cancellation/publication-drain and deletion-winning-publication regressions pass under race. There is no ownership I/O at the cited Background call sites to fix.

The attempted-state finding describes deliberate shared write-ahead evidence. Clearing it after this client sends nothing is unsafe because another tab may have dispatched that same record. de3a945fa adds a real IndexedDB commit/client-replacement proof: neither client sends, another storage connection sees retained attempted evidence, saved uncertainty stays blocked, and recovery sends the identical payload and settles a replay receipt. The full frontend gate passes.

Both commits are pushed and integrated into PR977. Fresh whole-PR local reviews213/214 and the integrated canonical gate are running. This is resolution evidence, not a claim that the new heads have finished review.

obra:
Addressed hosted review6546 in `5cdef626c`. A real hub RPC regression reproduced both failed saved `thread/list` and subscribed `thread/read` when an unrelated rendezvous file is malformed. Global discovery uncertainty is now distinguished from target-specific ownership errors: saved data can be read with all mutation capabilities disabled and no receipt authority, while mutation admission remains strict. The regression also verifies retained transcript turns, stable identity, and rejected rename. It passes alongside broader upgrade, restart, ownership and admission tests under race (10.325s). The fix is integrated into PR977 at `383eb752d`; full canonical validation and fresh review are running.

obra:
Addressed RoboRev 6567 in `cba6a0b14`. Both read recovery branches now preserve the rechecked, confirmed restart-required state before considering an unrelated roster refresh failure.

The expanded real-WebSocket regression confirms the incompatible owner before introducing a malformed unrelated entry. It reproduced the initial failure, then still failed after fixing only the first exit, proving the live-read fallback needed the same correction. The final test retains saved turns, identity, restart-required status, zero mutation capabilities and receipt authority, and rejected rename. Broader upgrade/ownership/admission tests passed under race (24.212s). Focused local review221 reports no issues.

The fix is integrated into PR977 and its fresh canonical gate is running. New hosted reviews remain required.

obra:
Review6575 is a false positive: a missing `attempted` field is not proof that the intent was never sent.

At the pre-change base, [the dispatcher sends a submitting record without recording attempt evidence](https://github.com/prime-radiant-inc/evener/blob/cf4df743a73c1234308653960e72c7905366760d/cmd/evener-hub/frontend/src/stores/mutationDispatcher.ts#L83-L104). It explicitly [retains submitting after timeouts, transport failures, and retryable unknown outcomes](https://github.com/prime-radiant-inc/evener/blob/cf4df743a73c1234308653960e72c7905366760d/cmd/evener-hub/frontend/src/stores/mutationDispatcher.ts#L163-L166). Thus old submitting records with no field include requests accepted by a daemon whose response was lost.

Only explicit `attempted: false` is evidence of a never-sent intent. New enqueue/recovery records persist false, and the dispatcher commits true before transport. Missing values must remain uncertain until authoritative daemon reconciliation. Backfilling false or using `!record.attempted` would allow uncertain old input through the nonauthoritative saved-descendant fast path, violating the explicit no-replay recovery requirement. This is intentionally conservative, not a migration bug. No compatibility or automatic replay change is appropriate.

Existing behavioral coverage distinguishes fresh never-attempted descendant intents from uncertain accepted mutations and verifies that saved projections cannot release the latter. The passing gate includes those tests.

PR #941, opened 2026-09-07
feat(tui): paste a Google credential JSON from the credentials panel

obra:
The TUI's credentials panel could not store a Google credential JSON. Enter on a `gcp-adc` instance showed a notice telling the user to go paste it in the web hub, because the panel had no multi-line input. It has one now.

## How a single-line field takes a multi-line document

A terminal delivers a bracketed paste as **one** key message carrying every rune, newlines included (bubbletea's `detectBracketedPaste`, v1.3.10), and the TUI does not disable bracketed paste. So a pretty-printed `application_default_credentials.json` arrives whole and its newlines never reach the branch that submits the prompt. Enter on a `credentialJson` instance opens that prompt and stores what it collects through `evener/auth/credentialJson/set`, the same RPC the web hub calls, so both surfaces write the same store and the hub validates the document before it lands.

Without bracketing, the document arrives as ordinary keys, and the outcome turns on which byte the terminal sends for a line break: LF maps to `ctrl+j`, CR to `enter`. The field now keeps both LF and space, which it dropped before, so an LF paste still accumulates whole and only the user's own Enter submits it. A CR paste submits at each line and cannot work, which is one reason for the second way in.

The other reason is that the two inputs cannot be told apart. `f` on the instance asks for the path to the file holding the document, read on the machine the user typed it on, not the hub's, with path completion. Serving both through one prompt would mean guessing whether a value was a path or the credential, and a terminal that marks no pastes leaves no signal to guess from. So the paste prompt echoes nothing at all, rendering a character count whatever it is handed, and the file prompt shows its path, which is not secret, in any platform's path syntax. The file is read inside the command rather than while the key is handled, so a slow filesystem cannot stall the interface, and it is opened without blocking so a path naming a pipe cannot leave that command waiting forever. What the open descriptor actually is decides whether it is read, and the read is bounded. Control bytes are stripped from the rendered value, since clipboard content reaches the view. A value that is neither a document nor a readable path is reported by its reason alone: the error line is rendered and outlives the panel, so it repeats none of what was submitted.

## Changes

- `credentials_panel.go`: Enter on a `credentialJson` instance emits the credential-JSON action; the notice fields, their clearing and their rendering are deleted. The footer hint reads "set credential", since Enter now sets whichever credential the instance offers.
- `text_input_modal.go`: `NewCredentialPasteModal`, which summarizes pasted material and keeps path completion.
- `hub_update_config.go`: opens the prompt for the new action; `credentialJSONDocument` resolves paste-or-path; the result routes to the credential-JSON store.
- `launchconfig_client.go`: `CmdAuthCredentialJsonSet`.

## Tests

`text_input_modal_test.go` covers a whole bracketed paste accumulating without submitting, an unbracketed line-feed paste accumulating the same way with its spaces and newlines intact, spaces being typable, the shortest document the hub accepts still being summarized, a long path staying visible, and control bytes not reaching the view. `credentials_panel_test.go` covers the action for a `credentialJson` instance and silence for an `adc`-only one. `hub_credential_json_test.go` covers the prompt, a pasted document reaching the hub verbatim, a path being read (with surrounding whitespace), a failure repeating none of what was submitted, and cancel/empty storing nothing. The paste branch, the file-read branch and the error's redaction were each mutated away to confirm the tests bite.

The second commit is the fix round for a review of the first: it found that the failure path printed the submitted value into a rendered error line that outlives the panel, that the field silently dropped spaces and line feeds, that the no-echo rule's length threshold had a two-rune margin over the shortest acceptable credential, and that control bytes reached the view. The review also confirmed the single-message paste behaviour empirically at 101, 401, 5,101 and 100,101 runes.

Docs: the Vertex paragraph in `docs/llm-providers.md`, a new §4.6 amendment in the Vertex spec recording that the notice is superseded, and a keybinding amendment in the phase-2 UI spec.

### DIFF
--- a/cmd/evener-tui/credential_file_other.go
+++ b/cmd/evener-tui/credential_file_other.go
@@ -1,0 +1,9 @@
+//go:build !unix
+
+package tui
+
+// nonblockingOpen has no equivalent outside unix: Windows has no O_NONBLOCK
+// for files, and the shapes it guards against (FIFOs opened by path, character
+// devices under /dev) are unix shapes. The descriptor is still judged after
+// the open, so a non-regular file is refused either way.
+const nonblockingOpen = 0
--- a/cmd/evener-tui/credential_file_unix.go
+++ b/cmd/evener-tui/credential_file_unix.go
@@ -1,0 +1,11 @@
+//go:build unix
+
+package tui
+
+import "syscall"
+
+// nonblockingOpen keeps opening a path from blocking on what it names. A
+// FIFO with no writer blocks an ordinary open until one arrives, which on the
+// update loop means the interface stops responding with no way out; with this
+// flag the open returns and the descriptor can be judged and refused.
+const nonblockingOpen = syscall.O_NONBLOCK
--- a/cmd/evener-tui/hub_credential_json_test.go
+++ b/cmd/evener-tui/hub_credential_json_test.go
@@ -87,7 +87,8 @@ func TestCredentialJsonResult_SendsThePasteAndReadsAPath(t *testing.T) {
 }
 
 // TestCredentialJsonResult_UnreadablePathIsReportedAndNotSent: a path that
-// cannot be read is the user's mistake to see, and nothing is stored.
+// cannot be read is the user's mistake to see — by its reason, since the
+// value itself is never repeated — and nothing is stored.
 func TestCredentialJsonResult_UnreadablePathIsReportedAndNotSent(t *testing.T) {
 	client, got, cleanup := credentialJSONHub(t)
 	defer cleanup()
@@ -99,11 +100,55 @@ func TestCredentialJsonResult_UnreadablePathIsReportedAndNotSent(t *testing.T) {
 		t.Fatalf("nothing should be sent for an unreadable path; got %+v", cmd())
 	}
 	err := updated.(hubModel).err
-	if err == nil || !strings.Contains(err.Error(), "no-such-credentials.json") {
-		t.Fatalf("err = %v, want one naming the path that could not be read", err)
+	if err == nil || !strings.Contains(err.Error(), "no such file or directory") {
+		t.Fatalf("err = %v, want one giving the reason the read failed", err)
+	}
+	if strings.Contains(err.Error(), missing) {
+		t.Fatalf("err = %q, must not repeat what was submitted", err)
 	}
 	if got.Value != "" {
 		t.Fatalf("the hub must not be called; it received %q", got.Value)
+	}
+}
+
+// TestCredentialJsonResult_NeverEchoesWhatWasSubmitted: the error line is
+// rendered and persists after the panel closes, so a value that is neither a
+// document nor a readable path — the wrong secret pasted into this prompt —
+// must not put any of itself there. The reason it failed is what the user
+// needs.
+func TestCredentialJsonResult_NeverEchoesWhatWasSubmitted(t *testing.T) {
+	const wrongSecret = `sk-ant-api03-SECRET-MATERIAL-that-is-not-a-path-or-a-document`
+	for _, tt := range []struct {
+		name  string
+		value string
+	}{
+		{name: "wrong secret", value: wrongSecret},
+		{name: "secret with a path separator", value: "AQ.Ab8RN6Jz/SECRET+MATERIAL/x"},
+		{name: "mistyped path", value: "/tmp/definitely-not-here/SECRET-MATERIAL.json"},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			client, got, cleanup := credentialJSONHub(t)
+			defer cleanup()
+			m := newHubModel(client, "http://hub.test")
+			m.followupModal = &tuipick.TextInputModal{}
+			updated, cmd := m.handleTextInputResult(tuipick.TextInputResultMsg{Tag: "credential-json-set:google-vertex", Value: tt.value})
+			if cmd != nil {
+				t.Fatalf("nothing should be sent; got %+v", cmd())
+			}
+			err := updated.(hubModel).err
+			if err == nil {
+				t.Fatal("the failure must be reported")
+			}
+			if strings.Contains(err.Error(), "SECRET") || strings.Contains(err.Error(), tt.value) {
+				t.Fatalf("err = %q, must not repeat what was submitted", err)
+			}
+			if !strings.Contains(err.Error(), "credential JSON") {
+				t.Fatalf("err = %q, want it to name what failed", err)
+			}
+			if got.Value != "" {
+				t.Fatalf("the hub must not be called; it received %q", got.Value)
+			}
+		})
 	}
 }
 
--- a/cmd/evener-tui/hub_credential_json_test.go
+++ b/cmd/evener-tui/hub_credential_json_test.go
@@ -5,8 +5,11 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"testing"
 
+	"testing"
+	"time"
+
+	tea "github.com/charmbracelet/bubbletea"
 	"primeradiant.com/evener/appwire"
 	"primeradiant.com/evener/cmd/evener-tui/internal/launchconfig"
 	"primeradiant.com/evener/cmd/evener-tui/internal/tuipick"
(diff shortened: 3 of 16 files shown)


obra:
All three taken.

- **Medium — hub validation errors can echo user-controlled fields.** Checked every message `CheckCredentialJSON` produces: two interpolate an untrusted value, the unsupported `type` and a foreign `token_uri`. None of the secret-bearing fields (`private_key`, `client_secret`, `refresh_token`) appears in any message, and the private-key failures carry only the parser's reason. So the exposure is narrow — it needs the pasted document to hold secret material in `type` or `token_uri` — but callers render what they get, and the TUI's error line outlives the prompt, which is exactly the leak this PR's second commit closed locally. Both sites now clip the quoted value to a 40-rune prefix: enough to identify what was refused, no more. Pinned by `TestCheckCredentialJSONBoundsWhatItEchoes` (400-rune `type` and `token_uri`; the message stays under 200 bytes and contains neither value in full), mutation-verified.
- **Medium — unbounded, blocking read of a user-supplied path.** Real and mine. The read runs on the update loop, so a FIFO hangs the interface with no way out and `/dev/zero` grows until the process dies. The path is now checked before it is opened: a regular file only, under 1 MiB. Pinned by `TestCredentialJsonResult_RefusesWhatIsNotAReadableFile` (named pipe, directory, `/dev/zero`, oversized file), which runs the handler on a goroutine and fails if it does not return within five seconds — with the guard removed, the pipe and device cases hang exactly as described.
- **Low — the outer footer still said "set api key".** Fixed to "set credential", matching the panel's own footer.

obra:
Both taken.

- **Medium — the path was validated with `os.Stat` and then reopened by name.** Right: the window is small and needs write access to the user's own path, but the consequence is a hung interface, and the fix is the better shape anyway. The file is now opened once with `O_NONBLOCK` (a no-op for a regular file, and what keeps a writerless FIFO from blocking the open), judged by `f.Stat()` on the open descriptor rather than by a second lookup of the name, and read through `io.LimitReader` at one byte past the bound — so a file that grows after the open cannot get past the limit either. The non-blocking flag is unix-only, so it is a two-file constant; on Windows the descriptor check still refuses anything that is not a regular file. Mutation-verified: with the flag set to 0, the named-pipe case blocks for the full five-second deadline, which is the failure this closes.
- **Medium — a one-line secret is indistinguishable from a typed path where the terminal marks no pastes.** Correct, and it was the residue of the earlier fix that dropped the length threshold. Rather than masking (which would make the path fallback unusable, since the user could not see or verify what they typed) or splitting the prompt in two, the field now shows only what looks like a path — a value beginning `/`, `~`, or `.` that was not reported as a paste — and summarizes everything else. A credential document is a JSON object, a marked paste is already covered, and anything else is treated as material to keep off the screen. Pinned by `TestCredentialPasteModal_ShowsOnlyWhatLooksLikeAPath` (absolute, home and relative paths shown; an unmarked secret summarized).

New test for the read: `TestCredentialJsonResult_BoundsAFileThatGrowsAfterItIsOpened`.

obra:
All three taken, the first two by removing the thing they both point at.

- **Medium — the read runs on the update loop.** Right: `O_NONBLOCK` bounds the open, not the read, so a slow filesystem still stalls the interface. The read now happens inside the command, alongside the RPC, and its failure travels the same route to the error line. The non-blocking open stays, since it keeps a path naming a pipe from leaving that command waiting forever. Pinned by `TestCredentialJsonFile_ReadsOffTheUpdateLoop`, which creates the file only after the handler returns: a read that happened there would have failed. (The first version of that test asserted on the hub call rather than the read and passed under the mutation — worth saying, since the mutation is what caught it.)
- **Medium — an unmarked one-line secret beginning `/`, `~` or `.` is echoed** and **Low — Windows paths are treated as credential material.** These are the same defect: one prompt served two inputs, so it had to guess which it was given, and no guess survives both a base64 secret starting with a slash and a `C:\` path. So there are two prompts now. Enter pastes the document into a prompt that echoes nothing at all — a character count, whatever it is handed — and `f` on the instance asks for the path in a prompt that shows it, because a path is not secret and needs to be read and completed. That resolves the Low as well: any platform's path syntax renders. The shape heuristic and the paste-mark bookkeeping it needed are both gone, so the display rule is now one line.

Docs and the spec's §4.6 updated to describe the two ways in.

obra:
Taken.

- **Low — a rejected `token_uri` can carry terminal control characters into the rendered message.** Confirmed: JSON only requires escaping below U+0020, so C1 characters ride through raw, and U+009B is CSI. `clipEcho` bounded the length but not the content. It now drops C0, DEL and C1 before clipping, so neither refusal that quotes a value can put anything a terminal would obey on screen. Pinned by `TestCheckCredentialJSONEchoesNothingATerminalWouldObey` (a `type` and a `token_uri` each carrying U+009B and U+0085), mutation-verified. The TUI's input field already stripped these; this is the other end.

Unrelated to this branch: the `tests` job failed on `internal/apptranscript`'s `TestItemTurnsFromEntriesLargeFixtureTiming`, which compares two wall-clock measurements and needs their ratio to stay above 3. This branch touches no file in that package, and the same test failed the same way on #924. Re-run, and filed as #944 with the analysis.

obra:
Taken, and it was my oversight: the round-4 fix widened the registry's filter to C1 and left the field's own filter at C0 and DEL, so the two ends disagreed about the same class of character. `printable` now drops U+0080–U+009F as well, matching `clipEcho`. A literal tab goes with them, since it can only arrive by paste — the Tab key completes a path — and rendering one would shift the line it is on.

Pinned by `TestTextInputModal_DoesNotRenderControlBytes`, now a table with a C0/DEL case and a C1 case that pastes U+009B and U+009D, asserting no control character survives into the view and that the printable part of the value does. Mutation-verified: narrowing the range back to C0 and DEL fails the C1 case and nothing else.

obra:
Taken. The one case that belongs in the portable file was the one that could not run everywhere: it matched "no such file or directory", which is how unix words that failure and not how Windows does. It asserts `errors.Is(res.Err, fs.ErrNotExist)` now, which is what the error already carries — the reason travels wrapped, since only the reason is kept and never the path. The helper's other two expectations ("not a regular file", "too large") are this code's own words and stay as they are.

Mutation-verified: replacing the wrapped error with a flat one fails that test and nothing else.

obra:
Accurate about the code, and I am not fixing it here. Filed as #948 instead, with the analysis and the shape of a fix.

The mechanism is right: `nonblockingOpen` is `syscall.O_NONBLOCK` on unix and `0` elsewhere, so on Windows a device-namespace path would be opened before the descriptor check could refuse it, and that open can block. One correction to the consequence: the read moved inside the `tea.Cmd` in round 3, so the interface stays responsive — what leaks is a goroutine and a prompt that never resolves, not a frozen TUI.

Why not now: nothing ships for Windows. `.goreleaser.yml` builds `linux` and `darwin` only, and the `GOOS=windows make vet` step says of itself that it is "a build-tag discipline check" rather than a support commitment (`.github/workflows/ci.yml`). Writing device-namespace handling — the `\\.\` and `\\?\` prefixes plus the reserved `CON`/`COM1`/`LPT1` family, which are reserved in every directory — would put platform code in the tree that neither CI nor any machine here can exercise, and I would rather not ship a Windows path I cannot run. #948 records both that and the `CreateFile`-flags alternative for whoever picks it up with a Windows runner.

Separately: the `tests` job failed again on `internal/apptranscript`'s `TestItemTurnsFromEntriesLargeFixtureTiming` — third sighting across two PRs, none of which touch that package. Re-run; the analysis is in #944.

obra:
Correcting my previous reply on two points, both in roborev's favour.

I framed one part as "a correction to the consequence", as though the finding had claimed a frozen interface. It had not — it said "leaving an uncancellable command goroutine", which is exactly right and is what I then described in my own words. There was nothing to correct.

I also justified deferring partly on not being able to exercise Windows code here. That holds for one of the two fixes offered (opening through `CreateFile` with flags that make a pipe fail rather than wait) and not for the other: rejecting the device namespace is string classification — the `\\.\` and `\\?\` prefixes and the reserved `CON`/`PRN`/`AUX`/`NUL`/`COM1`–`COM9`/`LPT1`–`LPT9` names, reserved in every directory — and that predicate is testable on any platform.

So the reason to defer is narrower than I made it sound, and worth stating plainly: it would be a guard that cannot fire on either platform evener ships (`.goreleaser.yml` builds linux and darwin), carried and maintained for a build nobody produces. That is a judgment call about scope, not a dismissal of the finding. #948 records it with both fix shapes, and I have amended it to say the same.

PR #949, opened 2026-09-07
fix(web): preserve attached images on turn-failure Retry

obra:
After a provider failure, the transcript's Retry button re-issued only the originating input's text, silently dropping attached images.

Root cause: TurnFailureEndCap's retryText/originatingInput returned string only, and retry() called send(sessionRef, text) with no attachments — even though the originating userMessage item still carried the image bytes.

Change: Retry now carries a RetryInput { text, attachments? }. Inline data: URI bytes are recovered into InputAttachments (markers from the translated '(attached image N)' prose, 1-based fallback); unresolvable bytes (sha-routed src, no inline data) drop with a warning toast instead of a silent text-only resend. Covers live and reloaded-failure lookback paths.

Verification: TurnFailureEndCap 20 tests (4 new), threads store 231, full transcript+reducer 76 files/1643 tests, biome ci clean, tsc clean.


obra:
Second roborev finding (repeated unnamed markers) fixed in 8e2a1755b: repeats collapse to first-copy before order/cardinality checks, extra copies stay verbatim, distinct sequences still refuse, round-trip guard unchanged, plus regression test. The empty-hop finding was refuted with a hand-trace (condition already correct; existing test :860 covers the cited input) — no change there. CI verifies both.

obra:
Refuting the Medium finding (`TurnFailureEndCap.tsx:179`) on the current head (`c41b8dc`; the flagged code is byte-identical to `8e2a175` — the `origin/main` merge touched neither owned file). The prescribed fix is already the code, and the cited test already pins the cited input.

**1. The condition already permits exactly what the fix asks for.** Current `:179`:

```ts
if (middle === "" ? text[next + 1] !== ")" && next + 1 !== text.length : !isTailHop(middle)) break;
```

For an empty hop, the loop breaks only when the next paren is *neither* followed by `")"` *nor* terminal — i.e. it **permits** the hop when the next closing paren is itself terminal or followed by another closing paren. That is the suggested fix, verbatim (landed in `4f24f091`, blame `:179`).

**2. Hand-trace of the exact flagged input** `(attached image 1: plot))` (no known names, so the named branch runs `extendRawSpanEnd("plot))", base=17)`): `close=4`, `end=5`; loop: `text[4]==")"` and `5<6`, `next=indexOf(")",5)=5`, `middle=""` → ternary evaluates `text[6] !== ")" && 6 !== 6` = `true && false` = `false` → no break → `end=6`. Span is `[0,25)` of length 25 — fully covered, so `occurrencesCoverText` returns true, `markerOnly` holds, and the UI shows the re-attach note, not text-only Retry. The repeat-collapse commit `8e2a175` did not touch this function (`git show 8e2a175 --stat`: only the pairing block + one test), so nothing shifted under the prior trace.

**3. Empirical probe** (verbatim copy of `:168-194` run under node): terminal `plot))` → end 6/6, `plot)))` → 7/7, `plot))))` → 8/8 (longer tails chain because each iteration re-checks the terminal/followed-by-`)` guard); non-terminal `a))b)` → stops at 2 (guard still protects `(see (the plot))`-shaped prose); `plot) (see (the plot))` → stops at first `)`. All as designed.

**4. Test evidence:** full scoped suite `TurnFailureEndCap.test.tsx` 55/55 green on this head, and the cited test `:879` ("a reloaded marker-only input whose lost filename ends with a paren shows re-attach") renders exactly `(attached image 1: plot))` with a sha-routed (bytes-gone) image and asserts no Retry button + the re-attach note. The behavior the finding says is missing is the behavior that test pins.

No change made — forcing the suggested edit onto `:179` would be a no-op restatement at best. If a reviewer has a concrete tail shape that still misbehaves, please quote the literal string and I'll trace it the same way.

obra:
Main-branch conflict resolved and verified: merge 4d30e77ab (c4717dcc3 + main 91d1d6356), single conflicted file (session_jobtree_drain_backoff_test.go) resolved, push landed, mergeable=MERGEABLE (remaining BLOCKED is CI-in-progress + review, not conflict). Post-merge scope vs main is 4 files, all pre-existing branch intent (TurnFailureEndCap.* retry fix + retained-output embedlit exclusion from 993d370f9/f9d469d5e); no foreign edits, no docs.

obra:
Roborev Medium (head 4d30e77, trailing-paren empty hop) — investigated with a probe, NO CHANGE: the finding misreads the condition and the cited behavior already works.

Line 179's ternary breaks only when BOTH conjuncts hold (next char isn't ")" AND next isn't end-of-text). A trailing paren at end of text makes next+1 === text.length, so the conjunction is false → the hop IS consumed. Scratch probe on verbatim logic: `(attached image 1: plot))` → markerOnly=true (re-attach state, as the :879 test expects); chained `plot)))` → consumed; `(plot) draft.png)` → consumed; prose cases (`explain (the plot)`, `plot)) hi`) correctly rejected. Scoped suite 55/55 green on unmodified code (head 4d30e77ab, tree clean).

This is the same case refuted twice before, not a new one — the condition already encodes the finding's requested fix.

PR #951, opened 2026-09-07
fix(execenv): make the glob listing and match budgets real bounds

### LINKED ISSUE #497: A single glob rooted at / OOM-kills the agent: the directory-listing budget is gated on filesystems that report no file identity, so it never applies on Linux

obra:
First of three stacked PRs for #497. This one fixes the original defects; **B** adds the memory bounds (chunked reads, entry caps, cancellation) and **C** budgets grep.

## What was wrong

A `find_files` call with a `**` pattern rooted at `/` could exhaust memory and OOM-kill the agent process. The walk *terminated* — the ancestor/SameFile cycle check added for #369 catches `/proc/<pid>/root` re-entering the tree — but nothing bounded what it spent getting there.

1. **The pattern walk's directory-listing budget was unreachable.** The only counter incremented inside `if !hasFileIdentity(info)`. Every file on ext4, APFS and HFS+ carries `(device, inode)` identity, so on the filesystems agents actually run on that branch was never taken.
2. **Matches accumulated with no cap.** A `**` over `/` is millions of paths.
3. **The walk retained one `fs.FileInfo` per listed directory**, though the cycle check only consults *ancestors* of the directory being admitted.

## What changed

The listing budget charges every listing **the pattern walk** makes, on both the sandboxed and off-sandbox arms. Identity now decides only how exhaustion is *explained*: a filesystem reporting none has no cycle detection at all, and that is worth saying.

Scope worth being precise about: the `.gitignore` scan that runs before the pattern walk when `include_ignored` is false is a **separate traversal and is not charged here** — the stacked #955 budgets it. On a 40-directory tree with the listings budget at 8, that scan accounts for 41 of the 49 `ReadDir` calls. The pattern walk is bounded on that path (`TestGlobChargesThePatternWalkOnTheDefaultIgnoreExclusionPath` pins it); the scan preceding it is not, until B. The budget is shared across a call's brace-expanded patterns, because `globpattern.Expand` turns one pattern into up to `MaxExpansions` of them and a per-walk bound would really be a 256x bound.

Matches are capped, and tripping the cap **ends the walk** rather than truncating after it, so the cap bounds the work and not just the answer's length.

`listed` became an ancestor chain pruned to the path being walked, so retention is O(depth) rather than O(listings). The chain is only a cache: an ancestor it does not hold is re-stat'ed, so a walk that jumps between subtrees still cannot miss a cycle.

Refusals are a typed `globBudgetError` carrying the operation, the bound, the count and whether the walk could have detected a cycle, so callers read fields rather than parse sentences.

## Caller-visible changes

- **`GlobExcluder.GlobWithExclusions` keeps its three return values.** Truncation rides on a caller-supplied budget instead: the new optional `GlobBudgeter` capability lets a caller hand in a `*GlobBudget` and read `TruncatedAt()` off it afterwards. This is the same move `GlobExcluder` itself was introduced as, and its doc comment says so.
- **Without a budget there is nowhere to report truncation, so that path refuses.** `GlobWithExclusions` returns an error rather than a short list with a nil error. A caller that wants the partial list opts in through `GlobBudgeter`.
- The glob tool prefers `GlobBudgeter`, so it still tells the model when the cap cut a listing short, naming the cap and noting it counts *candidate* matches, before dotfile/gitignore exclusion drops any.
- Sandboxed listings are sorted by name, so the match cap truncates to the same prefix on both arms.

## How it is tested

Each bound was written test-first and confirmed failing for its stated reason before the fix. Tests assert *work done*, not just results — listings actually made, identities actually retained — because a fix that finishes the walk and then reports a refusal would satisfy a result-only assertion while still doing the thing that OOMs. First failures:

- `walk retained 31 directory identities after listing 30 sibling directories; retained count grew with the number of directories listed, not with the depth of the path being walked`

`TestGlobWithExclusionsRefusesRatherThanTruncatingSilently` pins the budget-less refusal. Termination stays pinned by the #369 regression tests, and a characterization test proves the ancestor check still catches a cycle reached through a never-listed ancestor (verified by mutation: removing `identity`'s stat fallback breaks it).

Gates, all foreground: `gofmt`; `make vet` and `GOOS=windows make vet`; `go vet -tags evenerfuzz ./...`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test -race ./agent/execenv/`; `-tags evenerfuzz -run Fuzz ./agent/execenv/`; full `go test ./agent/`. Zero failures in each.

Closes #497

### DIFF
--- a/agent/cov_s3_shelltools_test.go
+++ b/agent/cov_s3_shelltools_test.go
@@ -3,8 +3,10 @@ package agent
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 
@@ -189,5 +191,58 @@ func TestS3Cov_GlobTool_FullyExcludedResultExplainsRatherThanEmpties(t *testing.
 	}
 	if !strings.Contains(res.Output, "index.js") {
 		t.Fatalf("expected include_ignored=true to find the match, got: %q", res.Output)
+	}
+}
+
+// TestS3Cov_GlobTool_TruncationNoteNamesTheCapWithoutDroppingMatches proves
+// the glob tool boundary surfaces GlobBudgeter's truncation (#497): when a
+// glob call hits the match cap, the tool result must still contain every
+// match collected before the cap tripped *and* say the listing was capped,
+// so a model cannot mistake a partial result for the whole answer. A bare
+// strings.Join(matches, "\n") is indistinguishable from an uncapped result,
+// which is exactly the failure mode this guards against.
+//
+// This drives the real environment rather than a fake: GlobBudget's
+// truncation accounting is only ever mutated from inside package execenv, so
+// a fake outside it has no way to make a caller-supplied budget report
+// truncation without actually tripping the walk. Not parallel: it lowers
+// execenv's package-level match cap for its duration.
+func TestS3Cov_GlobTool_TruncationNoteNamesTheCapWithoutDroppingMatches(t *testing.T) {
+	dir := t.TempDir()
+
+	const capAt = 7
+	const fileCount = capAt + 3
+	names := make([]string, fileCount)
+	for i := range names {
+		// Digit-free names: the cap-number assertion below must be satisfiable
+		// only by the note, never by a filename that happens to contain the
+		// cap's digits.
+		names[i] = fmt.Sprintf("match%c%c.txt", 'a'+i/26, 'a'+i%26)
+		if err := os.WriteFile(filepath.Join(dir, names[i]), []byte("x"), 0o644); err != nil {
+			t.Fatal(err)
+		}
+	}
+	t.Cleanup(execenv.SetMaxGlobMatchesForTesting(capAt))
+
+	s := newSession(t, withDir(dir))
+	res := s3cov_exec(t, s, "glob", `{"pattern":"*.txt","path":"."}`)
+	if res.IsError {
+		t.Fatalf("glob error: %v", res.Output)
+	}
+	// The note is a separate section after a blank line, so splitting there is
+	// what tells "matches plus a note" apart from a bare match list. Asserting
+	// on the whole output instead would let a match list with no note at all
+	// satisfy every check below.
+	matchList, note, hasNote := strings.Cut(res.Output, "\n\n")
+	if !hasNote || strings.TrimSpace(note) == "" {
+		t.Fatalf("truncated glob result carries no note, so a model cannot tell truncation happened: %q", res.Output)
+	}
+	for _, name := range names[:capAt] {
+		if !strings.Contains(matchList, name) {
+			t.Fatalf("truncated glob result dropped match %q: %q", name, res.Output)
+		}
+	}
+	if !strings.Contains(note, strconv.Itoa(capAt)) {
+		t.Fatalf("the truncation note does not name the cap (%d): %q", capAt, note)
 	}
 }
--- a/agent/execenv/execenv.go
+++ b/agent/execenv/execenv.go
@@ -178,6 +178,22 @@ type ExecutionEnvironment interface {
 type GlobExcluder interface {
 	// GlobWithExclusions behaves like Glob but also returns the number of
 	// candidate matches dropped by the default dotfile/gitignore exclusion
-	// (always 0 when includeIgnored is true).
+	// (always 0 when includeIgnored is true). A listing that would have been
+	// truncated by the match cap comes back as an error instead of a short
+	// list: this entry point has no budget of its own to report truncation
+	// through, so a caller that wants the partial list and a way to tell it
+	// was cut short uses GlobBudgeter instead.
 	GlobWithExclusions(ctx context.Context, pattern, basePath string, includeIgnored bool) (matches []string, excluded int, err error)
+}
+
+// GlobBudgeter is an optional capability an ExecutionEnvironment's Glob may
+// additionally implement, modelled on GlobExcluder for the same reason: it
+// exists so a caller that wants a truncated listing rather than the refusal
+// GlobExcluder gives does not have to widen Glob's own signature, which every
+// implementation — including test doubles with no budget accounting — would
+// otherwise have to grow a meaningless extra parameter for. The caller
+// supplies the budget and reads what the call had to cut off it once the
+// call returns.
+type GlobBudgeter interface {
+	GlobWithBudget(ctx context.Context, pattern, basePath string, includeIgnored bool, budget *GlobBudget) (matches []string, excluded int, err error)
 }
--- a/agent/execenv/glob_budget_test.go
+++ b/agent/execenv/glob_budget_test.go
@@ -1,0 +1,420 @@
+package execenv
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"slices"
+	"testing"
+	"testing/fstest"
+	"time"
+
+	"primeradiant.com/evener/agent/sandbox"
+)
+
+// stubMaxGlobDirListings lowers the directory-listing budget for a test and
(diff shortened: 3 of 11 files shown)


obra:
Both findings are fixed in the stacked follow-up #955, which is why they are not here: this PR was split out of #930 so each slice stays reviewable, and the memory bounds live in the next one.

- Ignore discovery bypassing the budget: #955 charges `loadIgnoreSet`'s walk to the same `GlobBudget` and propagates its refusal, so a default glob cannot traverse the tree unbudgeted before the pattern walk starts. #956 does the same for grep's discovery and its own walk.
- The match cap checked only after a full `ReadDir`: #955 reads directories in bounded chunks with a per-directory entry cap and a call-wide live-entry ceiling, so a single huge directory is refused while reading rather than after materializing.

Neither is a regression this PR introduces; both describe behaviour that predates it and that the stack removes. Merging A alone leaves the tree exactly as it is today on those two points, with the listing and match budgets newly enforceable.

obra:
Pushed `4c0a98032`, addressing all three.

1. **Body claim corrected.** It now says A charges every listing **the pattern walk** makes, on both arms, and states plainly that the `.gitignore` scan is a separate traversal charged in the stacked #955 — including the reviewer's 41-of-49 measurement. The "was unreachable" bullet is scoped to the pattern walk too.
2. **Dead code moved to C.** `advice()`'s grep arm and its test are gone from A; `advice()` now returns the glob lever unconditionally, with a comment saying why that is all there is here. Both are restored in #956, where grep first constructs a budget. I extended the same treatment to `entryAdvice()`'s grep arm in #955 for the same reason — it was equally unreachable there — so the full three-way advice test now lives in C alongside both arms.
3. **`GlobBudget` concurrency: documented, not synchronized.** A budget belongs to one call, and the walk it accounts for is single-goroutine; `holdEntries` runs once per listing, so locking it would cost the hot path to serve a case that does not arise. The doc now says it must not be shared across goroutines or reused for a concurrent call, and that a caller running globs concurrently gives each its own.

**New coverage:** `TestGlobChargesThePatternWalkOnTheDefaultIgnoreExclusionPath` exercises `include_ignored=false` and asserts the pattern walk's listings are still charged there, so the default path is not uncovered in A.

**On the repro shape:** agreed, and I left `Closes #497` alone.

B and C were rebased onto this and force-pushed (#955 `3e470cbc1`, #956 `4e3518c2f`). Gates re-run green on all three: gofmt; `make vet`; `go vet -tags evenerfuzz ./...`; `make lint-golangci`; tagged lints; `go test -race ./agent/execenv/`; fuzz; full `go test ./agent/` — zero `^--- FAIL:` in every log.

obra:
Restacked on green main (`f9d469d5e`). New heads: **A `279273622`**, **B `56d3838b6`** (#955), **C `8937a1238`** (#956). All three rebased without conflict, all `MERGEABLE`, A is zero commits behind main.

The rebase changed no content: `git diff` of C-against-A is byte-identical to what it was before the restack.

**#957 overlap: none, and not just at the file level.** Main since A's old base touches no file any of the three PRs touch. The efficiency sweep's grep work is a different subsystem entirely — `grepReaderLimit` and the 64 KiB reader buffers live in `agent/internal/jobstore/` and `agent/retained_output_read.go`, which grep *retained job output*. My grep changes are in `agent/execenv`'s filesystem search (`grepNative`, `grepWalk`, `grepAccum`). `agent/execenv` imports neither jobstore nor retained output, so there is no shared buffer, no shared helper, and no semantic overlap. `agent/retained_output_size_acc_test.go` is new and untouched by this stack.

**Catalog-literal fragility: none in the tests I added.** Every fixture is built in-test under `t.TempDir()` (or `fstest.MapFS`), and every assertion is on a structural field — `kind`, `op`, `count`, `peakDirEntries`, `peakLiveEntries`, entries read — or on an error identity via `errors.As`. Nothing reads a repo artifact, golden file or generated catalog, so there is no literal that a change elsewhere in the repo can invalidate.

**Gate results, zero `^--- FAIL:` in every log, per branch:**

| gate | A | B | C |
| --- | --- | --- | --- |
| `gofmt -l` | clean | clean | clean |
| `go vet ./...` | 0 | 0 | 0 |
| `go vet -tags evenerfuzz ./...` | 0 | 0 | 0 |
| `make lint-golangci` | PASS | PASS | PASS |
| `go test -race ./agent/execenv/` | 0 FAIL | 0 FAIL | 0 FAIL |
| `go test -race ./agent/` | 0 FAIL | 0 FAIL | 0 FAIL |
| `-tags evenerfuzz -run Fuzz ./agent/execenv/` | 0 FAIL | 0 FAIL | 0 FAIL |
| `-tags evenerfuzz -run Fuzz ./agent/` | 0 FAIL | 0 FAIL | 0 FAIL |
| full `go test ./agent/` | 0 FAIL | 0 FAIL | 0 FAIL |

Run in the foreground with a private `GOCACHE`, long runs detached under their own session with bounded PID waits.

Understood on CI: only this PR gets checks while B and C target their parents. I have not retargeted them and will not.

obra:
Fixed at `f08bb2eb1`. roborev's finding was real, and slightly worse than described: **both** weak assertions were dead, not just one.

- `strings.Contains(res.Output, "7")` was satisfied by the fixture filename `match07.txt`, as roborev said.
- `res.Output == strings.Join(names, "\n")` compared against **all ten** names while a truncated result has seven, so that guard could never fire either.

Together those meant the test passed with no note at all — exactly the case it exists to catch.

**The fix.** The tool composes the note as a separate section after a blank line, so the test now splits there with `strings.Cut(res.Output, "\n\n")` and requires a non-empty note before asserting anything else. The cap number is then asserted **against the note**, not the whole output, so a filename cannot satisfy it. Matches are checked against the match-list half. Fixture names are digit-free (`matchaa.txt` … `matchaj.txt`), so the cap-number check has no other way to pass.

Splitting on the separator also avoids pinning the note's prose, which the repo's test rules discourage: it asserts a note exists and names the cap, not how it is worded.

**Mutation-verified.** Suppressing the note in `session_tools_shell.go` (leaving `truncatedAt` otherwise wired) now fails with `truncated glob result carries no note, so a model cannot tell truncation happened`. The previous assertions passed that same mutation.

Since A's SHA moved, B and C are restacked: **#955 `f2e15d059`**, **#956 `32ddb923e`**. The rebase changed no content — C's diff against A is byte-identical to before — and all three are `MERGEABLE`.

**Full gate set re-run on all three branches, zero `^--- FAIL:` in every log:** gofmt; `go vet ./...` and `-tags evenerfuzz`; `make lint-golangci`; `go test -race ./agent/execenv/`; `go test -race ./agent/`; fuzz seed replay on `./agent/execenv/` and `./agent/`; full `go test ./agent/`.

Leaving the ignore-discovery finding to C as you directed, and noting the `internal/appserver` race-root flake needs nothing from me unless it recurs.

obra:
Head `f08bb2e` is green on all twelve checks, and roborev's remaining finding is the ignore-discovery walk, which is fixed in the stacked #956 rather than here.

That is the split's shape, not an oversight: this PR makes the listing and match budgets real bounds, #955 bounds the memory a single walk can materialize, and #956 budgets grep's discovery and its own walk, which is where `loadIgnoreSet` gets scoped to the pattern's prefix subtree and propagates its refusals. Merging this one alone leaves discovery exactly as it is on main today, with the pattern walk newly bounded.

Also fixed here since the last review: the truncation-note test was vacuous in both of its assertions, so it passed with no note at all. It now splits the tool result at the note separator, requires a non-empty note, checks the cap against the note rather than the whole output, and uses digit-free fixture names so no filename can satisfy the check by accident. Suppressing the note now fails it; under the old assertions the same mutation passed.

obra:

PR #947, opened 2026-09-07
feat(webui): type ramp, reading measure, rhythm, headings, folded tool runs

obra:
## Why

The web hub's type, spacing and balance were measured against the running app (1440 and 375, both themes, a real glm-5.2-vision session and the `/dev/surfaces` fixtures). The findings are in `docs/web-ui/typography-spacing-critique-2026-09-06.md`; the headline numbers:

| Before | Measured |
|---|---|
| Type ramp | 12/13/14/16/20, with 340 of 382 size declarations at 12–13px |
| Agent prose line length at 1440 | **149 characters per line** (76rem column, 92% bubble) |
| Agent prose width on a 375px phone | **260px of 375** (28 characters per line) |
| Pane titles | 12px uppercase micro-labels; a session's prompt rendered as an uppercase sentence |
| `--ink-low` (96 text rules) | 3.1:1 dark / 3.5:1 light |
| Phone composer | 13px, with pinch-zoom disabled app-wide to hide it |
| Undefined tokens / off-ramp literal sizes | 4 / 18 (including 9px text) |

## What changed

Each commit is one task of `docs/superpowers/plans/2026-09-06-webui-typography-spacing.md`; every value lands in `tokens.css` and flows through the existing token contract.

- **Ramp**: 12 caption / 13 ui / **15 body (16 on phones)** / 18 pane-title / 22 page-title / 28 display; line-heights 1.6 body, 1.4 ui, 1.25 title.
- **Measure**: `--session-measure: 44rem` on `body` (704px, ~90 characters), shared by the transcript column, composer, cold start, spawn form and settings content; **Settings → Theme → Transcript width** (Reading / Wide, 64rem). Bubbles are bounded by the column, not 92% of the pane. Below 700px both speaker rows become a grid so prose spans the phone under the avatar row.
- **Rhythm**: `--rhythm-line/-item/-group/-exchange` (4/8/16/24). 24px above each user message (revises the 2026-07-30 "no dead air" ruling, now that the column is bounded), 8px between calls, a group step above turn footers; pane bodies pad 24px on desktop.
- **Headings and eyebrows**: pane titles are 18px semibold sentence case. One uppercase recipe (`--tracking-eyebrow` 0.06em, caption size, `--ink-mid`) replaces four tracking values across 21 rules; "Agents & models" → "Agent setup".
[...]

PR #954, opened 2026-09-07
feat(webui): dedicated job_watch renderer with humanized summaries

obra:
Replaces the generic job_\* fallback for job_watch with an exact-match descriptor (mockups docs/web-ui/history/mockups/23-job-watch.html sections A-D):

- Intent-first summaries with humanized durations (after 300s to in 5m, progress_interval_ms 120000 to every 2m; absolute clock only for created_at)
- Bodies are note-only / one sentence / list rows / empty; notes render in full to ~20 lines with disclosure only beyond
- No watch id on single-watch surfaces (list rows, inspect summaries, raw disclosures only); status chips only in list rows

Watch-triggered notifications drop the false warning tone and the producer echo fields (mockups section E): a fired watch is the expected outcome, never something needing a human. No watch notification earns a tone chip, ever.

Verification: vitest 135/135 on affected files, full transcript suite 1528/1528, tsc clean, Biome clean.

### DIFF
--- a/agent/job_notify.go
+++ b/agent/job_notify.go
@@ -250,6 +250,14 @@ func formatJobNotificationBlock(n jobNotification, excerpt notificationExcerpt, 
 	if n.TranscriptRef != "" {
 		attrs = append(attrs, notificationAttr("transcript_ref", n.TranscriptRef))
 	}
+	// A job-targeted fire, teardown notice, or send-rail diagnostic carries
+	// no timer WatchID, but its originating watch rides OriginWatchID — emit
+	// it so a reader can identify which watch to inspect or clear. Timers
+	// keep their existing emit path below; the WatchID guard keeps the two
+	// from ever doubling the attribute.
+	if n.OriginWatchID != "" && n.WatchID == "" {
+		attrs = append(attrs, notificationAttr("watch_id", n.OriginWatchID))
+	}
 	if n.Status == jobNotificationEventWatch && n.JobID == "" {
 		if n.WatchID != "" {
 			attrs = append(attrs, notificationAttr("watch_id", n.WatchID))
--- a/agent/job_notify_test.go
+++ b/agent/job_notify_test.go
@@ -1129,3 +1129,58 @@ func TestFormatWatchSendNotificationBlock(t *testing.T) {
 		}
 	}
 }
+
+// A job-targeted watch fire carries no timer WatchID, but its originating
+// watch rides the display-only OriginWatchID — emitted as the watch_id frame
+// attribute so a reader can identify which watch to inspect or clear
+// (RoboRev PR #954). Empty origins emit nothing (no regression for old or
+// hand-built shapes), and a timer's own WatchID path never doubles the attr.
+func TestFormatJobNotificationEmitsOriginWatchID(t *testing.T) {
+	t.Parallel()
+	fired := formatJobNotificationBlock(jobNotification{
+		JobID: "job_a1b2", JobType: "watch", Status: "watch",
+		Reason:        "output_match: ready",
+		OriginWatchID: "watch_09QmWzRtNvxK",
+	}, notificationExcerpt{}, true)
+	if !strings.Contains(fired, `watch_id="watch_09QmWzRtNvxK"`) {
+		t.Errorf("job-targeted fire must carry its originating watch id:\n%s", fired)
+	}
+	if !strings.Contains(fired, "Job job_a1b2 watch.") {
+		t.Errorf("job-targeted fire keeps the generic producer body:\n%s", fired)
+	}
+
+	bare := formatJobNotificationBlock(jobNotification{
+		JobID: "job_a1b2", JobType: "watch", Status: "watch",
+		Reason: "output_match: ready",
+	}, notificationExcerpt{}, true)
+	if strings.Contains(bare, "watch_id=") {
+		t.Errorf("empty origin must not emit a watch_id attribute:\n%s", bare)
+	}
+
+	timer := formatJobNotificationBlock(jobNotification{
+		JobType: "watch", Status: "watch", Reason: "repeat",
+		WatchID: "w1", IntervalSeconds: 300,
+		OriginWatchID: "w1",
+	}, notificationExcerpt{}, true)
+	if got := strings.Count(timer, "watch_id="); got != 1 {
+		t.Errorf("timer must emit watch_id exactly once, got %d:\n%s", got, timer)
+	}
+}
+
+// watchNotificationFromWatch stamps the producing config's id on the
+// display-only origin field — and never on the timer-identity WatchID, which
+// would subject the notice to the orphan-tick drop once its watch detaches.
+func TestWatchNotificationFromWatchStampsOriginWatchID(t *testing.T) {
+	t.Parallel()
+	jm := newTestJM(t)
+	n := jm.watchNotificationFromWatch(&watchConfig{watchID: "watch_origin1"}, "job_x", "output_match: ready", nil)
+	if n.OriginWatchID != "watch_origin1" {
+		t.Errorf("OriginWatchID = %q, want watch_origin1", n.OriginWatchID)
+	}
+	if n.WatchID != "" {
+		t.Errorf("WatchID = %q, want empty (timer-identity field must stay unset)", n.WatchID)
+	}
+	if n.JobID != "job_x" || n.Reason != "output_match: ready" {
+		t.Errorf("notification identity/reason = %q/%q, want job_x/output_match: ready", n.JobID, n.Reason)
+	}
+}
--- a/agent/job_watch.go
+++ b/agent/job_watch.go
@@ -3485,12 +3485,28 @@ func watchNotification(jobID, reason string) jobNotification {
 	}
 }
 
+// watchSendDiagnosticNotification builds a send-rail diagnostic (a drop, a
+// failed persist, an eviction) that names the watch whose delivery failed.
+// The watch id rides the display-only OriginWatchID field — never WatchID,
+// which would subject the diagnostic to the orphan-tick drop once its watch
+// detaches (see OriginWatchID's doc comment in jobs.go).
+func watchSendDiagnosticNotification(watchID, jobID, reason string) jobNotification {
+	n := watchNotification(jobID, reason)
+	n.OriginWatchID = watchID
+	return n
+}
+
 func (jm *jobManager) watchNotificationFromWatch(cfg *watchConfig, jobID, reason string, root *provenance.Causal) jobNotification {
 	n := watchNotification(jobID, reason)
 	if cfg == nil {
 		return n
 	}
 	n.Note = cfg.note
+	// The originating watch rides a display-only field, never WatchID: the
+	// session drops a non-terminal WatchID-bearing entry whose timer key no
+	// longer resolves, and a condition watch's key slot is not its id — so a
+	// stamped fire or teardown would vanish as an orphaned tick.
+	n.OriginWatchID = cfg.watchID
 	visibleSessionID := cfg.receiverSessionID
 	if visibleSessionID == "" {
 		visibleSessionID = jm.sessionID
@@ -3702,7 +3718,7 @@ func (jm *jobManager) deliverPendingWatchSend(cfg *watchConfig, state jobstore.W
 	if ensurePending {
 		if err := jm.appendWatchSendPendingState(cfg, state); err != nil {
 			jm.enqueueWatchNotifications([]jobNotification{
-				watchNotification(state.Key.ResolvedWatchedIdentity, "watch send pending state failed: "+limitWatchText(err.Error(), watchReadErrorMaxChars)),
+				watchSendDiagnosticNotification(state.Key.WatchID, state.Key.ResolvedWatchedIdentity, "watch send pending state failed: "+limitWatchText(err.Error(), watchReadErrorMaxChars)),
 			})
 			return false, err
 		}
@@ -3725,14 +3741,14 @@ func (jm *jobManager) dropWatchSend(state jobstore.WatchSendState, cfg *watchCon
 		WatchSend: &dropped,
 	}}); err != nil {
 		jm.enqueueWatchNotifications([]jobNotification{
-			watchNotification(state.Key.ResolvedWatchedIdentity, "watch send dropped state failed: "+limitWatchText(err.Error(), watchReadErrorMaxChars)),
+			watchSendDiagnosticNotification(state.Key.WatchID, state.Key.ResolvedWatchedIdentity, "watch send dropped state failed: "+limitWatchText(err.Error(), watchReadErrorMaxChars)),
 		})
 		return err
 	}
 	jm.removePendingWatchSend(cfg, dropped.Key, dropped.UpdateSeq)
 	jm.releaseStableWatchReceipt(dropped.DeliveryID)
 	jm.enqueueWatchNotifications([]jobNotification{
-		watchNotification(state.Key.ResolvedWatchedIdentity, "watch send failed: delivery_id="+state.DeliveryID+": "+dropped.DiagnosticReason),
+		watchSendDiagnosticNotification(state.Key.WatchID, state.Key.ResolvedWatchedIdentity, "watch send failed: delivery_id="+state.DeliveryID+": "+dropped.DiagnosticReason),
 	})
 	return nil
 }
@@ -3922,7 +3938,7 @@ func (jm *jobManager) persistPendingWatchSend(state jobstore.WatchSendState, d w
 	if len(record.evictions) != 0 && jm.appendEvents == nil {
 		if err := jm.appendWatchSendEvents(record.pendingEvents); err != nil {
 			jm.enqueueWatchNotifications([]jobNotification{
-				watchNotification(state.Key.ResolvedWatchedIdentity, "watch send pending state failed: "+limitWatchText(err.Error(), watchReadErrorMaxChars)),
+				watchSendDiagnosticNotification(state.Key.WatchID, state.Key.ResolvedWatchedIdentity, "watch send pending state failed: "+limitWatchText(err.Error(), watchReadErrorMaxChars)),
 			})
 			return state, false, err
 		}
@@ -3933,7 +3949,7 @@ func (jm *jobManager) persistPendingWatchSend(state jobstore.WatchSendState, d w
 			if err != nil {
 				jm.removeWatchSendTerminalSnapshots(applied)
 				jm.enqueueWatchNotifications([]jobNotification{
-					watchNotification(state.Key.ResolvedWatchedIdentity, "watch send pending state failed: "+limitWatchText(err.Error(), watchReadErrorMaxChars)),
+					watchSendDiagnosticNotification(state.Key.WatchID, state.Key.ResolvedWatchedIdentity, "watch send pending state failed: "+limitWatchText(err.Error(), watchReadErrorMaxChars)),
 				})
 				return record.persisted, true, err
 			}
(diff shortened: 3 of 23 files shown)


obra:
All findings from both reviews addressed in 0445e59bf (CI 12/12 green):

**Review 1 (bc52271):**
1. Job-targeted watch misclassified — fixed: event/status watch now classifies as watch regardless of job_id (steeringClassify.ts), with a job-aware title ('Output matched on job_a1b2') and the trigger reason as secondary. Regression test included.
2. List/inspect condition gaps — fixed: shared Condition-string parser handles repeat_seconds, events/every/where, and heartbeat in both rows and inspect bodies (producer grammar: watchConditionSummary, job_watch.go:2460-2494).
3. Non-interactive list rows — fixed: rows are real buttons expanding an inline detail sentence.

**Review 2 (f0e2281, +2 findings):**
4. Missing raw fallback — fixed: absent structured state falls back to the raw footer text instead of rendering nothing.
5. Duration remainders dropped — fixed: 90s reads 'in 1m30s', repeat_seconds humanized in rows.

Verified: 144/144 affected vitest, 1555 transcript-wide, tsc + Biome clean.

obra:
Second review addressed in dc3e6c9 (CI re-running):

1. Trigger-derived titles — 'Output matched on …' only for output_match fires; event fires read 'Event on job: KIND', timers 'Timer fired'.
2. Every throttle preserved — create summaries/bodies show '(every N)'; inspect passes it into the condition sentence.
3. All row forms expand — event-only, filter, and heartbeat rows render detail sentences; no expandable row is a no-op.
4. Successful-call filters explicit — 'successful tool calls on read_file' in summaries and sentences.

Verified: 148/148 affected vitest, 1559 transcript-wide, tsc + Biome clean.

obra:
Rebased onto latest main (f9d469d) and pushed as abd81c3:

- Dropped my kimi-k2.6 one-liner: main's 92bc0f8 rewrote TestNamedInstanceKeepsItsIdentity to pin the cheap model to the registry instead of hardcoding it, which subsumes the fix. The pre-existing main failure is gone properly.
- job_watch descriptor opts into the new #947 folding contract as fold-never (watch cards are tracked state, like every other job descriptor).

Verified post-rebase: 148/148 affected vitest, tsc + Biome clean.

obra:
Third review addressed in f5e5444 (verified: 165/165 affected vitest, 1638 transcript-wide, tsc + Biome clean):

1. Trigger-derived titles — 'watch ended:' → Watch ended, 'watch cleared:' → Watch auto-cleared, before the fired fallbacks.
2. Entity decoding — watch reasons decode before title/secondary use.
3. Semicolon split — Condition splits only on recognized field heads, so ';' inside output_match patterns survives.
4. Every on filters — '(every N)' in filter sentences, rows, and details, printed once.
5. Shared filter phrase — summaries, rows, and details use one helper (error/ok/tool all explicit).
6. Progress-only summaries — 'Watch <source> · every 2m'.
7. Humanized timer secondaries — bare 'after'/'repeat' derive 'after 5m'/'every 5m' from the prose lead.
8. Watch card identity — job-targeted fires show the watched job id; echo fields stay suppressed.

obra:
Rebased onto latest main — this also resolves the lint situation properly: main's dc29b00 replaced the hand-rolled delta struct (and my nolint on it) with appwire's own params type, and ec92622 restored the tag-casing/import-order lanes, so my lint-fix commit is dropped as redundant. My diff is back to frontend-only. Verified post-rebase: 165/165 affected vitest, tsc clean.

obra:
Combined review addressed in bb74486 (verified: 180/180 affected vitest, 1653 transcript-wide, tsc + Biome clean):

**Mediums:**
1. Watch-state honesty — inspect/list now distinguish pending / ended / not-found in the producer's own grammar (formatJobWatchInspect + watchInspectFound), instead of labeling everything ended. Missing watches name no source; pending rows chip pending.
2. Non-expandable rows render as plain divs, never focusable no-op buttons.
3. Job-targeted fires synthesize card prose from the reason attr (the producer emits only the generic 'Job <id> watch.' body) — output_match, event, progress_tick, plus raw-reason fallback for unrecognized triggers.
4. 'Job id: self' suppressed — 'self' is source vocabulary, never a job id.

**Lows:**
5. Timer humanizers reach hours (3600 -> 1h, 86400 -> 24h), matching the renderer's grammar.
6. Unrecognized object raws fall back to the raw footer, never an empty card.
7. Dead 'matching' ternary collapsed (bare-filter branch proven unreachable: events win first).
8. Empty inspect ids fall back to the 'job_watch: inspect' verb.

One correction while verifying: my earlier budget-clear fixture used job_id="self" — no producer emits that (autoClearWatchOverBudgetNotification passes ""). Fixed the fixture to backend truth; the 'self' guard stays as defense.

obra:
Fixed the web failure in bb74486's CI run: my new CSS comment contained the literal '#954', which the token-contract gate reads as a chromatic hex literal outside tokens.css. Reworded in 43fe73f (verified: token-contract 1576/1576 locally). TS/TSX review references are unaffected — only stylesheets are scanned.

obra:
Second combined review addressed in 818e809 (verified: 187/187 affected vitest, 1660 transcript-wide, tsc + Biome clean):

**Mediums:**
1. Teardown generic bodies — teardown reasons now go through the replacement path, with a test using the actual generic producer body.
2. Status-only tone — jobNotificationTone mirrors the parser (event OR status watch earns no chip), with regression test.
3. Combined triggers — create summaries and ConditionSentence compose all populated clauses (pattern + events/filter + heartbeat); heartbeat-only summary keeps its approved shape. Behavior tests for combined watches.
4. Deliveries denominator dropped — the count includes periodic ticks outside the condition-fire budget, so it renders bare ('3 deliveries') instead of a misleading 'N of 50'.

**Lows:**
5. Single-decode — verified synthesized prose decodes exactly once (reason decoded at synthesis, body never re-encoded); regression test pins a literal '&lt;' sequence.
6. Empty create card — sourceless-condition creates return null at the wrapper, not an empty bordered card.
7. Live-row counts via watchState — summary counts agree with row chips.

obra:
Third combined review addressed in 33d5b9a (verified: 191/191 affected vitest, 1664 transcript-wide, tsc + Biome clean):

1. Sentence punctuation — clause nodes now carry no separators of their own. The filter's tool/outcome/event/every sub-clauses join with spaces inside a single node; top-level clauses join with ', ' exactly once. Fixes the doubled comma ('outputs ready, , heartbeat …') and comma-spliced filters.
2. Heartbeat-only lifecycle — the auto-clear clause renders only for budgeted triggers (pattern, events, filter). Heartbeat-only watches end with a period and claim no auto-clear, since periodic ticks never consume the condition-fire budget.

obra:
Fourth combined review addressed in 6e38dea (verified: 194/194 affected vitest, 1667 transcript-wide, tsc + Biome clean):

1. Composite detail views — the inspect output-match branch now renders through the shared ConditionSentence, and the expanded row detail composes pattern + events/filter + heartbeat together, so detailed views never drop armed triggers the summary shows.
2. Single-decode prose — synthesized watch prose is stored escaped-form (new escapeNotificationEntities, the producer escaping run backward) so the card's one decode renders literal entity sequences correctly; teardown passthrough now compares decoded-against-decoded. Parser- and card-level regression tests pin it.
3. tap-min scope — min-height consumes the token only under the coarse-pointer query, matching its declaration scope and the RailRow treatment; desktop rows keep content height.

Note: this push supersedes the re-run I kicked off for the 4 failed jobs (audit + drain-timing + race-modules) — fresh CI will run on the new head instead. Those failures are all outside my frontend-only diff, and the audit test passes locally on this tree.

obra:
Fifth combined review addressed in e9113df (verified: new Go tests red-checked + affected package tests/vet/gofmt clean; 202/202 affected vitest, 1675 transcript-wide, tsc + Biome clean):

1. Watch identity on every frame — job-targeted fires, teardown notices, and all seven send-rail diagnostic sites now stamp the originating watch, emitted as the watch_id frame attribute; watch cards render Watch id alongside Job id. Done via a display-only OriginWatchID field, deliberately NOT the timer-identity WatchID: the session drops non-terminal WatchID-bearing entries whose timer key no longer resolves, and a condition watch's key slot is never its id — reusing it would swallow these very notices as orphaned ticks once their watch detaches. The output-dropped feed guard is left alone (no watch id in scope there).
2. Delivery failures labeled — 'watch send ...' and 'output dropped:' reasons title as 'Watch delivery failed' with a warning tone instead of presenting as successful triggers with forced neutral. Disposition and the delegate/observer tone path untouched (verified never to see watch frames).

obra:
Rebased onto latest main (25 commits, incl. watch-batch #974 and any-note #995) plus two newest findings fixed in 21dee96 (verified post-rebase: Go watch/notification tests + vet + gofmt clean; 206/206 affected vitest, 1718 transcript-wide, tsc + Biome clean):

1. Watch-card disclosure prefers the watch id (legacy job-id fallback) — two watches on one job expand independently now.
2. Sub-second durations honest — 1500ms reads 'every 1.5s', and rounding can never surface a 60s carry; same fix in the notification timer humanizers.

Rebase notes: two conflicts in agent/job_watch.go, both keep-both — main's 'any job_watch carry a note' (n.Note = cfg.note) alongside my OriginWatchID stamp, and main's eviction-snapshot ordering fix alongside my diagnostic constructor swap. No unrelated files touched.

obra:
Latest review addressed in 730a7e4 (verified: affected Go tests + vet + gofmt clean; 207/207 affected vitest, 1719 transcript-wide, tsc + Biome clean):

1. **High — duplicate eviction persist (my rebase error, confirmed):** my e9113df hunk re-applied the sequential per-eviction append loop inside main's batch path, which already persists evictions via the group write. Removed — batch path is back to collect-and-enqueue-diagnostics only. Regression test counts raw journal Evicted records on both seams (red-checked: fails with exactly 2 on batch without the fix). The same rebase also dropped my nil-seam swaps; restored.
2. **Missing OriginWatchID:** nil-seam pair + stable-enqueue trio switched to the helper. Every send-rail diagnostic now stamped.
3. **notificationTone + child unreachable:** the tone mirror is added as defense in depth (that path verifiably never sees watch frames today — only delegate/observer callers). The real gap was 'child unreachable:' escaping the failure predicate (the runtime fallback site emits the DiagnosticReason directly); the predicate now recognizes it, with a title/tone test.

obra:
All five findings addressed in 98e31d5 (verified: 216/216 affected vitest, 1728 transcript-wide, tsc + Biome clean; backend subset untouched-green):

1. Condition notes — create/inspect bodies render a clamped note section below the sentence, row details name it, and the Condition-string parser preserves note: parts. Row one-liners stay note-free.
2. Job-targeted notes — synthesized prose keeps the trailing producer Note: section (decode-then-re-escape so literal entities survive exactly once); teardown-with-note passthrough pinned by test.
3. Self label — the self source renders as this session everywhere.
4. Repeat keys — disclosure joins watch id with content, so repeat firings expand independently (timer repeats previously regressed into one key).
5. Dot-all — output_match/event/note patterns use ([\s\S]+): multiline patterns title, synthesize, and render.

Note: no backend change needed here — main's #995 already carries notes end-to-end (result, Condition string, every fire body); this was purely a frontend rendering gap.

obra:
All three Lows addressed in e9ab5ec (verified: 1731 transcript-wide vitest, tsc + Biome clean):

1. Catch-up/clear ordering — terminal_catchup is checked before the clear fallback, so an arg-less send-branch catch-up (watch_id + watching:false + terminal_catchup, verified against runTerminalCatchup's send path) reads as a catch-up naming its terminal outcome, never 'Cleared …'. Genuine clears carry no terminal/inspect markers, so that inference is unchanged.
2. Per-delivery disclosure identity — SteeringItem threads item-id:fragment-index into an optional disclosureId prop that joins the scoped key. Byte-identical repeat frames expand independently; existing callers without the prop keep today's keys.
3. every:1 — normalized to unset like the backend (numArg's only caller was the args throttle; renamed to everyArg with the <=1 rule), so create summaries/bodies no longer print '(every 1)'.

obra:
History collapsed to 4 commits (was 16) and both open issues fixed in the top commit (verified: biome ci src, tsc, Go tests/vet/gofmt, 1734 transcript-wide — all green on the collapsed stack):

- web-lint failure: the fragment key now uses a stable item-id:index variable instead of a bare loop index, so no suppression is needed at all.
- Timer row-detail notes: after/repeat branches append the embedded note like every other branch, with an expanded-row test for each.
- New RoboRev Medium folded in too: expanded timer rows name their note.

The 4 commits: feature base, backend watch plumbing, frontend review fixes, latest touch-ups. All previously reviewed content preserved verbatim (key files byte-identical across the collapse).

obra:
All four Lows addressed (verified: biome ci src, tsc, 1741 transcript-wide — all green):

1. Button chrome — .row gets the full native reset (toolcallitem idiom) plus intentional hover wash and focus ring.
2. Separators — per-row wrappers removed (rows are direct list children); container rule draws borders on non-first rows only.
3. Summary-only gating — new optional hasBody(item) descriptor predicate, wired into ToolCallItem; clear and catch-up rows render non-expandable, with a ToolCallItem-level integration test.
4. Wildcard — centralized eventDisplayName; summaries, sentences, rows, and details all read 'any event'.

History holds at 4 commits (folded today's fix into the latest to respect the cap).

obra:
New Medium addressed (verified: affected tests + tsc + Biome clean): 'output_match attach scan skipped:' — emitted when the retained-output read fails while the watch still installs live — now classifies as a delivery failure (title + warning) instead of presenting as a successful trigger. Same reason-prefix family as the other diagnostics; no wire change. Folded into the 4th commit to hold the history cap.

obra:
Two more Lows addressed (verified: Go condition-summary test + vet + gofmt; 1745 transcript-wide, tsc, biome ci src — all green; folded to hold 4 commits): note renders terminal in the Condition grammar so delimiter-bearing notes survive whole (backend emits note last, parser treats the head as terminal), and the arg-less clear fallback now requires create/clear markers — legacy inspect-misses read as 'not found' with their body shown.

obra:
Fixed both Mediums from the `9e38707` review (folded into the webui fix commit, 4-commit cap held):

- **Parser vs legacy ordering** — reworked rather than patched. `list`/`inspect` raws now carry the note in its own structured `note` field (populated on both the live-config and ended-history paths), so delimiter-looking text (`; events: […]`) in a note renders verbatim. The `Condition` string keeps the producer's existing order; the parser reverted to legacy split semantics (first `note:`-part value) with the string clause as fallback for stored frames that predate the field. Regression tests: structured delimiter note + legacy note-before-events ordering.
- **Doubled `Note:` prefix** — `watchNoteSection` returned the note with its prefix while `withNote` prepended another; it now slices after the full head and the M2 test asserts exactly one `Note:` and no `Note: Note:`.

Gates: Go subset green, 1746/1746 transcript vitest, `tsc` clean, `biome ci src` clean. Still needs human approval — not merging.

obra:
Fixed all 4 Lows from the `8a56741` review (folded into the webui fix commit, 4-commit cap held):

- **Double terminal punctuation** — `ConditionSentence` now owns the sentence's single terminal period via a `meta` prop (`matches — 3 deliveries, created ….`); metadata joins with commas, never beside the period.
- **`Reminds in 5m`** — one-shot timers render `Remind me in 5m` (mirroring create summaries); `Reminds every …` retained for repeating timers, in both inspect bodies and row details.
- **Pending hides details** — pending inspect bodies now render the pending sentence together with the trigger, note, deliveries, and creation details through the same helpers as other states.
- **Raw end-reason ids** — shared `endReasonPhrase` formatter for list rows and inspect bodies (covers the real backend ids: `budget_exhausted`, `auto_removed_terminal`, `replaced`, `fired`, `cleared`, `job_manager_closed`; unknown ids fall back to the raw id).

Gates: 78/78 file vitest (6 new tests), 1752 transcript-wide, `tsc` clean, `biome ci src` clean. Still needs human approval — not merging.

obra:
Fixed both findings from the `5c202de` review (folded into the webui fix commit, 4-commit cap held):

- **Medium: entity codec asymmetry** — the watch-note lane now composes into `&`-first order: `watchNotificationFromWatch` pre-escapes `&` in the note, and the body's existing `<`-only escaping composes into exactly what the card's single full decode inverts. A note literally containing `&lt;` rides as `&amp;lt;` and renders `&lt;`, never `<`. Scoped to the note lane only (`cfg.note` stays raw for list/inspect readers; the shared body escaper other lanes use is untouched). The frontend decoder comment now documents the contract honestly, including the known remainder: shell/delegate job-output excerpts still ride `<`-only with no pre-escape, so literal entities there can over-decode — changing that lane would alter the wire format in `job_notify.go`, out of scope. Go + vitest coverage with literal entity text.
- **Low: ended branch hides details** — ended inspect bodies now render the condition/details and `NoteSection` through the shared helpers (mirroring active/pending) before the end-reason close, with tests.

Gates: 150/150 file vitest, 1757 transcript-wide, `tsc` clean, `biome ci src` clean, Go watch subset + `gofmt` clean. Still needs human approval — not merging.

obra:
Fixed both Mediums from the `71a7981` review (folded into the webui fix commit, 4-commit cap held):

- **Parser invents triggers from note prose** — `parseConditionText` now takes the structured note and strips its exact embedded `note:` clause (clause-boundary match, `note:` at start or after the join separator) before splitting, so delimiter-looking prose like `check; events: [communicate]` never parses as armed triggers. The truncation candidates mirror the producer's exact bounds (2048 runes, `\n[truncated]`, verified against `agent/job_watch.go`). Legacy string-only frames without a structured note keep the previous fallback behavior.
- **Note-only watches lose their note** — `conditionSpec` treats a non-empty note as body-bearing: the create summary heads the note (`Watch <source> · <head>`), and the create body renders the note section even with no trigger. Truly conditionless watches without notes keep the existing summary-only behavior.

Gates: 87/87 file vitest (6 new tests), 1763 transcript-wide, `tsc` clean, `biome ci` clean. Still needs human approval — not merging.

obra:
Fixed the Medium from the `51bb12b` review (folded into the webui fix commit, 4-commit cap held) — watch deliveries bypassing the dedicated renderer:

- **Reclassification** — the parser treats a non-empty `watch_id` as a watch delivery after the `watch_send` check (send frames carry no `watch_id`; timer frames already carry `event/status` watch, so no regression either way). A `job.notification` watch's enriched frame (`event/status` completed + originating `watch_id`) now renders as a watch, not an ordinary job card.
- **No invented prose** — the enriched frame's reason is the job's completion reason, never a trigger, so these frames split normally (excerpt + note preserved) instead of routing through `watchProse` synthesis. Title reads `Watch fired on <id>`, naming the delivery without inventing a trigger.
- **Neutral tone** — tone takes the parser's type, so the completed job's own exit can never error-tone the watch card; delivery-failure detection still runs first.

No backend change: the `watch_id` attr the fix keys off is the existing `OriginWatchID` emit this PR already added. Gates: message-file suites green, 1767 transcript-wide, `tsc` clean, `biome` clean. Still needs human approval — not merging.

obra:
Fixed all 3 Lows from the `6bdc9ed` review (folded into the webui fix commit, 4-commit cap held):

- **Raw `self` in titles** — watch titles map `self` to `this session` (the same human label `sourceLabel` uses) at all three sites: output-match, event, and the watch-fired fallthrough.
- **Raw Condition grammar in rows** — the watching-row fallback prefers the structured/parsed note when present and degrades to source-only otherwise; the raw `note: …` grammar never renders.
- **Pending misread as Cleared** — arg-less clear inference now keys only on `replaced_existing`/`fired` presence (both serialize without `omitempty`, so key presence is reliable); `source` is no longer a marker, with the comment updated to say why. A pending inspect with `source` reads as inspect.

Gates: 167/167 file vitest, 1774 transcript-wide, `tsc` clean, `biome` clean. Still needs human approval — not merging.

PR #955, opened 2026-09-07
fix(execenv): bound the memory a glob walk can materialize

obra:
Second of three stacked PRs for #497, on top of #951. **Base it on `fix/497-glob-budget-a`, not main.** #951 fixed the original defects; this one bounds memory; **C** budgets grep.

## What was wrong

#951's listing and match bounds cap how much a walk *does*, not how much it *holds*. Every listing still materialized its whole directory before any bound was consulted: `os.DirFS`'s `ReadDir` is `os.ReadDir`, and the sandboxed arm ended in a single `df.ReadDir(-1)`. One directory holding millions of entries exhausted memory before the listing budget or match cap ever ran.

A per-directory cap alone would not have been enough either. `fs.WalkDir` and doublestar both keep an ancestor's entry slice alive while they read a child, so a walk N levels deep holds N listings at once and the real peak is the cap times depth.

## What changed

- **Chunked listings.** Directories are pulled 4,096 entries at a time and charged after each chunk, so a pathological directory stops partway rather than after it has already been read whole. Entries are sorted once at the end, so a listing that stays under the bound is byte-for-byte what `os.ReadDir` would have produced and the match cap still truncates to the same prefix.
- **A per-directory entry cap** (200,000, roughly 21 MB at ~105 bytes per `os.dirEntry`). The refusal names the directory and says the count is one directory's entries, because the lever there is that directory, not the call.
- **A call-wide live-entry ceiling** (500,000, roughly 52 MB). Release comes free from the rule the ancestor chain already uses: to be listing a directory at all, the walk must have left every non-ancestor. The tracking is scoped to one traversal, so a listing left over from ignore discovery cannot inflate a later pattern walk's total.
- **Per-chunk cancellation.** `cancelFS` guards `Open`, `ReadDir` and `Stat`, so it sees a listing *start* and never the chunks inside it. The loop checks the context itself and answers with the same `context.Canceled` the rest of the walk returns.
- **`Stat` forwarding.** `boundedDirFS` embeds `fs.FS` as a bare interface, which exposes only `Open` and hid `os.DirFS`'s `fs.StatFS`. A file that can be stat'ed but not opened stopped matching a literal glob.

## Caller-visible changes

- **A single directory over 200,000 entries now fails glob outright**, naming that directory, rather than being read into memory.
- **A walk holding more than 500,000 entries live at once fails.** A deep tree of large directories can reach this even when no single directory is over the per-directory cap. A wide, flat repository is unaffected: the bound is on entries held *simultaneously*, not a cumulative count.
- A cancelled glob stops reading a large directory promptly instead of finishing it first.

## How it is tested

Tests assert *work done*, not just results — entries actually read, `peakDirEntries`, `peakLiveEntries` — because a fix that reads a whole directory and then reports a refusal would satisfy a result-only assertion while still doing exactly the thing that OOMs. First failures:

- `Glob over a 30-entry directory with an entry budget of 10 = (...30 matches..., <nil>), want a *globBudgetError`
- `Glob over a 10-level tree (peak live so far: 52) with a live-entry budget of 30 = (49 matches, <nil>), want a *globBudgetError`
- `Glob("secret.txt") = [], want [.../secret.txt]`

`TestGlobSucceedsOnAWideShallowTreeUnderTheLiveEntryCeiling` is the control that keeps the live bound honest: the same total entries spread across siblings must still succeed, because a cumulative counter would break every large flat repository. `TestGlobLiveEntryTrackingIsScopedToOneTraversal` pins the per-traversal scoping, and the sandboxed arm has its own live-entry witness — both verified by mutation.

Gates, all foreground, zero failures each: `gofmt`; `make vet` and `GOOS=windows make vet`; `go vet -tags evenerfuzz ./...`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test -race ./agent/execenv/`; `-tags evenerfuzz -run Fuzz ./agent/execenv/`; full `go test ./agent/`.

### DIFF
--- a/agent/execenv/gitignore.go
+++ b/agent/execenv/gitignore.go
@@ -2,6 +2,7 @@ package execenv
 
 import (
 	"context"
+	"errors"
 	"io/fs"
 	"path"
 	"strings"
@@ -51,10 +52,21 @@ type ignoreDir struct {
 // this walk otherwise has no other way to honor, since fsys itself (a
 // symlink-refusing, root-confined secureDirFS) enforces confinement but not
 // masking. The off-path caller (no masking concept) passes a no-op skip.
-func loadIgnoreSet(fsys fs.FS, skip func(relPath string) bool) *ignoreSet {
+func loadIgnoreSet(fsys fs.FS, skip func(relPath string) bool) (*ignoreSet, error) {
 	set := &ignoreSet{}
+	var budgetErr error
 	_ = fs.WalkDir(fsys, ".", func(p string, d fs.DirEntry, err error) error {
 		if err != nil {
+			// A budget refusal is not an unreadable entry to skip past. This
+			// walk reads the same filesystem the pattern walk does, under the
+			// same bounds, so one can trip here — and an ignoreSet that gave
+			// up partway reports itself complete, silently under-excluding
+			// every rule it never reached. That is a wrong answer rather than
+			// a missing one, so it has to reach the caller.
+			if _, refused := errors.AsType[*globBudgetError](err); refused {
+				budgetErr = err
+				return err
+			}
 			return nil //nolint:nilerr // best-effort: skip unreadable entries
 		}
 		if p != "." && skip != nil && skip(p) {
@@ -85,7 +97,7 @@ func loadIgnoreSet(fsys fs.FS, skip func(relPath string) bool) *ignoreSet {
 		set.dirs = append(set.dirs, ignoreDir{rel: dir, matcher: matcher})
 		return nil
 	})
-	return set
+	return set, budgetErr
 }
 
 // matches reports whether relPath (slash-separated, relative to the search
--- a/agent/execenv/gitignore_covtest_test.go
+++ b/agent/execenv/gitignore_covtest_test.go
@@ -40,7 +40,10 @@ func TestIgnoreSet_Matches_PathEqualsDir(t *testing.T) {
 		"sub":            {Mode: os.ModeDir | 0o755},
 		"sub/.gitignore": {Data: []byte(content)},
 	}
-	set := loadIgnoreSet(fsys, nil)
+	set, err := loadIgnoreSet(fsys, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
 	// "sub" == id.rel, so rel becomes "." — "*.log" won't match "."
 	if set.matches("sub", false) {
 		t.Fatal("expected sub (file) NOT to be ignored")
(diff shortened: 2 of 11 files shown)


obra:
Fixed at `0cbcf4faa`. **I moved the propagation guard down into B**, per your lean, and I agree with the reasoning.

**Why moving it is right.** B already walks the *budgeted* filesystem during discovery — `loadIgnoreSet` gets the same `cancelFS{boundedDirFS{budget}}` the pattern walk uses — so B's own per-directory entry cap and live-entry ceiling can trip while rules are being collected. B then swallowed that and returned an ignoreSet that reports itself complete. That is a wrong answer, not a missing one, and roborev's example is exact: I reproduced it by mutation and a `.gitignore`d `secret.txt` comes straight back in the results.

**Why B does not need C's scoping to be coherent.** The obvious worry is that propagating without C's narrowing gives B false refusals on unrelated directories. It does not, in B's own terms: B's discovery legitimately walks the whole base, so if it crosses a memory bound, the memory really is being spent — refusing is the correct answer for what B does. C's scoping then makes discovery do *less work in the first place*, which is an improvement on B rather than a correction of it. So B stands alone safely, and neither slice is judged unfairly.

`loadIgnoreSet` now returns `(*ignoreSet, error)` and propagates every budget refusal kind; all four callers and the test call sites are updated.

**Pinned by `TestGlobIgnoreDiscoveryRefusesRatherThanUnderExcluding`**, which puts the excluding `.gitignore` *in* the oversized directory so swallowing is directly observable. Mutation-verified — restoring the swallow fails it with the path that should have been excluded:

`Glob("secret.txt") = ([.../secret.txt], nil), want a refusal: discovery gave up on the entry cap before reading the .gitignore that excludes secret.txt, so the rules were never loaded and the path came back anyway`

C is restacked on this (**#956 `5d548af3b`**); resolving the overlap kept B's guard and C's scoping and budget work.

**Full gate set, zero `^--- FAIL:` in every log:** gofmt; `go vet ./...` and `-tags evenerfuzz`; `make lint-golangci`; `-race ./agent/execenv/`; `-race ./agent/`; fuzz seed replay on both packages; full `go test ./agent/`.

Nothing retargeted.

obra:
Fixed at `f5a5cb844`: the duplicated `err != nil` check in `gitignore_covtest_test.go` is gone. It was mine — a bulk edit inserted the guard twice. I scanned both files I touched that way and this was the only instance.

Full gate set re-run, zero `^--- FAIL:` in every log: gofmt; `go vet ./...` and `-tags evenerfuzz`; `make lint-golangci`; `-race ./agent/execenv/`; `-race ./agent/`; fuzz seed replay on both packages; full `go test ./agent/`.

#956 is restacked on this at `03417d2bd`. Nothing retargeted.

obra:

PR #956, opened 2026-09-07
fix(execenv): budget grep's ignore discovery and its own walk

### LINKED ISSUE #938: grep's tree walk is still unbudgeted; only its ignore discovery shares the glob budget

obra:
Third of three stacked PRs for #497, on top of #955. **Base it on `fix/497-glob-budget-b`, not main.** #951 fixed the original defects, #955 bounded memory; this one budgets grep.

## What was wrong

- **Ignore discovery ran first and unbounded.** `loadIgnoreSet` walks the entire base collecting `.gitignore` rules *before* the traversal that follows it, so a glob or grep with the default `include_ignored: false` traversed the whole tree before any bound applied. `include_ignored: true` skips it, but that is not the default.
- **Grep's own walk had no bound at all.** `grepWalk` and `secureBrowseWalkDir` are plain `fs.WalkDir` seams; nothing charged them.

## What changed

- Ignore discovery's **scope** is now what the traversal after it can reach, rather than the whole base regardless of pattern. A rule collected in a directory only ever affects a path at or under that directory, so the rules that can touch a candidate live in its own ancestors: for each expanded pattern, discovery reads the `.gitignore` at every level from the base down to the pattern's meta-free prefix (one `fs.ReadFile` per level, no listing) and then walks that prefix's subtree under the budget. A pattern starting with a metacharacter has an empty prefix, so discovery walks everything — correct, because that pattern's own walk does too, and the budget then applies to both consistently.
- Ignore discovery spends the same budget the traversal after it spends, on both arms. Its refusal reaches the caller rather than being swallowed as a best-effort I/O error, because a partial ignoreSet reported as complete would silently under-exclude the rest of a huge tree. Every *other* error stays best-effort.
- Both `grepNative`s hoist one budget shared with their ignore discovery, wrap the filesystem they walk so entries are bounded too, and charge a listing per directory.
- The charge lands **after** the dot-prefix, gitignore and masked `SkipDir` checks in the same callback, so a directory the walk turns away from costs nothing. A tree with many excluded directories and few listed ones no longer fails on a budget it never spends.
- `secureDirFS.ReadDir`'s nil-budget fallback from #955 is removed: every construction site now supplies a budget, so listings always go through the chunked reader.

## Caller-visible changes

- **Grep can now fail where it used to return partial output**, on both the sandboxed and off-sandbox arms.
- **Ignore discovery can fail a call**, which affects glob and grep alike whenever `include_ignored` is false. It propagates every budget refusal, the live-entry ceiling included: its walk starts at the root and visits everything below it, so reaching that ceiling means the process is already carrying the memory the ceiling exists to prevent, and leaving a later walk to refuse would spend it first. `TestGlobIgnoreDiscoveryPropagatesTheLiveEntryRefusal` pins that.

## How it is tested

Fixture arithmetic is stated in each test's doc comment and isolates the walk's own charges from ignore discovery's. For example `TestGrepWalkDoesNotChargeSkippedDirectories`: 3 real, 20 dot-excluded and 10 gitignored directories against a budget of 25, where ignore discovery alone costs 14 and the walk legitimately adds 4 — above discovery's cost, below the budget, so headroom cannot hide a regression. Verified by mutation: moving the charge back above the skips fails it with `grep walk made 26 directory listings, past the budget of 25`.

The two walk-callback guards are **not** reachable by any static tree, for the same subset-and-ordering reason above. `TestGrepWalkCarriesTheEntriesRefusalWhenADirectoryGrowsAfterIgnoreDiscovery` and its sandboxed twin force the concurrent case by growing the directory inside a stubbed walk; both were mutation-verified against deleting the guard. The off-sandbox probe lives in a portable test file, since `grepWalk`'s guard is cross-platform and nothing in that probe needs Unix; only the sandboxed twin carries a build constraint.

Gates, all foreground, zero failures each: `gofmt`; `make vet` and `GOOS=windows make vet`; `go vet -tags evenerfuzz ./...`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test -race ./agent/execenv/`; `-tags evenerfuzz -run Fuzz ./agent/execenv/`; full `go test ./agent/`.

Closes #938

### DIFF
--- a/agent/execenv/gitignore.go
+++ b/agent/execenv/gitignore.go
@@ -7,6 +7,7 @@ import (
 	"path"
 	"strings"
 
+	"github.com/bmatcuk/doublestar/v4"
 	gitignore "github.com/sabhiram/go-gitignore"
 )
 
@@ -28,8 +29,13 @@ import (
 // base (e.g. a repo root .gitignore when base is a subdirectory) are not
 // consulted; this is a known simplification for a directory-scoped search.
 type ignoreSet struct {
-	// dirs is ordered root-first so the "deeper overrides shallower" search
-	// above is a simple linear scan.
+	// dirs is in collection order, which is root-first only for a
+	// single-prefix scope: a call scoped to several prefixes appends each
+	// prefix's ancestors and subtree in turn, so {a,b}/*.txt yields ".", "b",
+	// "b/c", "a". Nothing depends on the order today, because matches ORs
+	// every rule that applies and never resets. Anything that later gives a
+	// deeper rule precedence over a shallower one has to sort first rather
+	// than rely on this slice.
 	dirs []ignoreDir
 }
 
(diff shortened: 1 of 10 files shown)


obra:
Pushed `6ec2f3d6b`. All three documentation fixes, plus the Windows-coverage question turned out to be fixable rather than just answerable.

1. **Stale carve-out comment corrected.** `loadIgnoreSet`'s doc said the live-entries ceiling stays best-effort; it now says the errors it returns are budget refusals **of every kind**, and the dangling "see its own comment below" is gone. The section in the body resting on the same wrong claim is deleted outright: as you say, `TestGlobLiveEntryTrackingIsScopedToOneTraversal` is a unit test on the budget and passes unchanged under propagation, so there was no trade to flag. The body now states the propagation and names `TestGlobIgnoreDiscoveryPropagatesTheLiveEntryRefusal` as its pin.
2. **Characterization line deleted.** You are right that it is a real red test.
3. **Lexical-order claim dropped.** It came from #951, where the sandboxed listing became sorted; C changes nothing there.

**On the Windows question: the probe moved, no answer needed.** `TestGrepWalkCarriesTheEntriesRefusalWhenADirectoryGrowsAfterIgnoreDiscovery` uses nothing Unix-specific — it stubs `grepWalk`, drives `grepNative`, and writes into a `t.TempDir()`. It now lives in a portable `glob_budget_toctou_test.go` with no build constraint, along with the `growDirWithFiles` helper both probes share. Only the sandboxed twin keeps `//go:build linux || darwin`, which it genuinely needs for `sandboxedEnv`. So `grepWalk`'s cross-platform guard now has cross-platform coverage.

Re-verified after the move: both probes pass, and deleting each arm's guard fails its probe. `GOOS=windows make vet` is clean.

Gates re-run green: gofmt; `make vet` and `GOOS=windows make vet`; `go vet -tags evenerfuzz ./...`; `make lint-golangci`; `make lint-evenerfuzz lint-gofmt lint-internal`; `go test -race ./agent/execenv/`; fuzz; full `go test ./agent/` — zero `^--- FAIL:` in every log.

obra:
Pushed `503991457`. Ignore discovery's scope now matches the walk's reach.

**One correction to the placement.** The ruling said B, "alongside the rest of the discovery budgeting" — but B does not touch `gitignore.go` at all (`git diff split/a split/b -- agent/execenv/gitignore.go` is empty). Discovery budgeting and its refusal propagation are both introduced here in C, so C is the PR that creates the false-refusal case and therefore the one that must not ship it. Landed here; B is untouched.

**The narrowing is semantically neutral, not an approximation.** `ignoreSet.matches` applies a rule collected in directory `id.rel` only to a path equal to `id.rel` or under `id.rel+"/"` — everything else hits `continue`. So a `.gitignore` in `huge/` can never affect a candidate under `sub/`, and ancestors-plus-prefix is exactly the affecting set.

Implementation is the two halves you specified: for each expanded pattern's `doublestar.SplitPattern` prefix, one `fs.ReadFile` per ancestor level with no `ReadDir` at all, then a budgeted `fs.WalkDir` of that prefix's subtree propagating every refusal kind. Prefixes are deduped, and a `.` anywhere collapses to a single whole-base walk. Grep passes `["."]` — its scope never narrows, matching its walk.

**Tests.** `TestGlobIgnoreDiscoverySkipsDirectoriesThePatternCannotReach` was genuinely red: `Glob("sub/*.txt") ... = ([], glob walk read 20 entries from directory huge, past the per-directory budget of 10 ...), want sub's file and no error; ignore discovery is reading directories the pattern walk can never reach`. It pins entries actually read, so it fails again if discovery widens; I mutation-verified by forcing the scope back to `["."]`.

Being straight about the other two: `...StillCoversEverythingForAWildcardPattern` and `...AppliesAnAncestorGitignoreAboveThePrefix` **passed before the fix as well as after**, since the old full-base walk already satisfied both. They are guards, not demonstrations — one stops this becoming "discovery is never budgeted", the other proves the split loses no rules.

**Regression check:** no existing gitignore test encoded the full-base walk. The four direct `loadIgnoreSet` callers now pass `["."]`, which is their existing behaviour; every other test drives patterns whose prefix is `.` anyway. `go test ./agent/execenv/ -run 'Ignore|Gitignore|IgnoreSet'` is green with no behavioural edits.

One judgement worth noting: the subtree walk keeps its `p != "."` skip check rather than `p != prefix`. Generalising it would have exempted a masked or dot-named prefix directory from its own skip check — a masking bypass on the sandboxed arm.

Gates green: gofmt; `make vet` and `GOOS=windows make vet`; `go vet -tags evenerfuzz ./...`; `make lint-golangci`; tagged lints; `go test -race ./agent/execenv/`; fuzz; full `go test ./agent/` — zero `^--- FAIL:` in every log.

obra:
Pushed `82f40c94d`. All four addressed.

**1. The unpinned skip check — and my first attempt at pinning it was wrong, which is worth recording.** I wrote a sandboxed test with the oversized directory *inside* the masked prefix, and it passed under the `p != prefix` mutation. The reason: generalising the check exempts only the prefix itself, so its children are still skip-checked, the oversized child is still pruned, and the entry cap never trips. That fixture could not observe the bypass, so I deleted it rather than ship a test that looked like coverage.

The bypass's real effect is that discovery *lists the masked directory and reads the rules inside it*. That is invisible in a glob's results, because rules from a masked directory only apply to paths under it and those are dropped anyway — the wrong thing is the reading. So it is pinned where it is observable: `TestLoadIgnoreSetConsultsSkipForTheScopePrefixItself` drives `loadIgnoreSet` with scope `["vault"]` and a skip that masks `vault`, asserting skip is consulted about the prefix itself and that no rule from inside it is collected.

The dot-directory line has its own pin, `TestGlobIgnoreDiscoveryStillPrunesADotDirectoryNamedAsThePrefix`. Mutation-verified together: changing both checks to `p != prefix` now fails both tests, one per line.

**2. `ignoreScopeForPatterns` comment corrected.** It no longer claims discovery only reaches where the pattern's walk can. It says the scope is the prefix's subtree, that a directory inside the prefix is still walked and can still refuse even where the pattern would never match inside it — with your `sub/*.txt` beside `sub/huge/` as the example — and why narrowing further would mean doing the walk's job.

**3. `dirs` ordering claim corrected.** It now says the slice is in collection order, root-first only for a single-prefix scope, with your `{a,b}/*.txt` → `.`, `b`, `b/c`, `a` case; that nothing depends on it today because `matches` ORs and never resets; and that anything adding precedence has to sort first.

**4. Ancestor-loop nit.** One comment noting the dot-directory check is deliberately absent there and why it cannot change an answer (`isDotPath` drops those candidates first).

Gates green: gofmt; `make vet` and `GOOS=windows make vet`; `go vet -tags evenerfuzz ./...`; `make lint-golangci`; tagged lints; `go test -race ./agent/execenv/`; fuzz; full `go test ./agent/` — zero `^--- FAIL:` in every log.

Ready for the rebase onto current main once #958 lands; say the word and I will restack all three and re-run the gates.

obra:
Fixed at `5d548af3b`. The High finding is real and now closed.

**The gap.** Scoped discovery's ancestor read checked `skip(dir)` and then read `dir/.gitignore` unconditionally. Masking is per *path*, so an unmasked directory can hold a masked `.gitignore` — and the base itself is never masked while a `.gitignore` directly inside it can be. `secureDirFS` enforces symlink refusal and root confinement but not masking, so that file got read. Exactly the prefix-bypass class, one level finer, as you said.

**Fix:** build the file path, then apply `skip` to the file, then read.

**I checked every other place discovery reads a rules file.** There is only one other, the subtree walk, and it is already safe: `fs.WalkDir` hands it every entry including files, and its `skip(p)` check runs before the read. I recorded that in the comment beside the new check so the asymmetry does not read as an oversight.

**Pinned by `TestLoadIgnoreSetSkipsAMaskedAncestorGitignoreFile`**, driving `loadIgnoreSet` directly the way the prefix test does, with a skip that masks `sub/.gitignore` while leaving `sub` visible. Mutation-verified — removing the check fails it with the skip's own call log showing the file was never offered:

`skip was never consulted about the ancestor rules file itself; it was asked about [sub sub/deep sub/deep/.gitignore sub/deep/keep.txt]`

**Restacked** on B's new head after the propagation guard moved down there (#955 `0cbcf4faa`). The overlap resolution keeps B's guard and C's scoping and budget work; B's new under-exclusion test survives the merge intact.

**Full gate set, zero `^--- FAIL:` in every log:** gofmt; `go vet ./...` and `-tags evenerfuzz`; `make lint-golangci`; `-race ./agent/execenv/`; `-race ./agent/`; fuzz seed replay on both packages; full `go test ./agent/`.

Nothing retargeted.

obra:
Fixed at `03417d2bd`. Discovery is now scoped by the pattern's **recursion and depth**, not just its literal prefix.

Only `**` matches across a separator, so for a pattern with none, every remaining separator is exactly one level: `*.go` reaches depth 0 (the base alone), `a/*/b/*.go` reaches 2, and only a `**` is unbounded. The subtree walk now cuts at that depth, so a non-recursive pattern never inspects a directory its own walk cannot list.

**Both shapes are pinned, and each mutation breaks exactly one:**
- `TestGlobIgnoreDiscoveryStopsAtANonRecursivePatternsDepth` — `*.go` beside a 20-file unrelated subtree with the entry cap at 10 must succeed and return only `top.go`. Forcing every scope unbounded fails it.
- `TestGlobIgnoreDiscoveryStillWalksTheSubtreeForARecursivePattern` — forcing every scope to depth 0 fails it.

**A note on that second test, because my first version of it was worthless.** I originally asserted that `**/*.go` still refuses on the oversized tree. It passed under the depth-0 mutation, because that refusal comes from the *pattern's own walk*, which covers the same subtree regardless of what discovery does. It pinned nothing. It now asserts exclusion instead: a nested `sub/.gitignore` must still exclude `skipme.go`, which is only true if discovery actually descended.

**Two existing tests had to be retargeted, and one for the same reason.**
- `TestGlobIgnoreDiscoveryIsChargedToTheBudget` used a nothing-matching pattern to isolate discovery. That no longer isolates anything, since the two reaches now match by design. It uses the `include_ignored` control instead: one pass fits the budget, two do not.
- `TestGlobIgnoreDiscoveryPropagatesTheLiveEntryRefusal` went vacuous for the same reason — with a recursive pattern the pattern walk refuses whether or not discovery propagated. It now drives `loadIgnoreSet` directly, and re-adding the carve-out fails it again.

**Full gate set, zero `^--- FAIL:` in every log:** gofmt; `go vet ./...` and `-tags evenerfuzz`; `make lint-golangci`; `-race ./agent/execenv/`; `-race ./agent/`; fuzz seed replay on both packages; full `go test ./agent/`.

Restacked on #955 `f5a5cb844`. Nothing retargeted.

obra:

PR #959, opened 2026-09-07
Hub self-update from Settings (exec-in-place restart)

obra:
## Summary

Settings → Hub → Updates lets a user pick a channel (release / snapshot), see whether that channel is ahead of the running build, and click **Update and restart**. The hub downloads and installs the channel archive with the existing `selfupdate.Upgrade`, then `syscall.Exec`s the installed `evener` binary in place: same PID, same argv, same environment, so launchd / systemd / a bare shell all keep the process they had. The page polls `/api/health` until the version changes, then reloads.

- `internal/selfupdate`: `Check` (GitHub API commit vs `buildinfo.GitSHA`) and `Restart` (exec in place, unix only).
- `appwire`: `evener/update/check`, `evener/update/apply`; `SettingsHubOverview.buildChannel`.
- `cmd/evener-hub/app_update.go`: handlers; dev builds refused; channel validated to release/snapshot; apply serialized with a mutex; restart scheduled 500ms after the response is flushed. Exec failure logs and leaves the old hub running.
- Frontend: `stores/hubUpdate.ts` + `panes/settings/sections/hubUpdates.tsx` (channel selector, check, confirm dialog, restart poll with 30s timeout).
- No stored channel setting: the installed binary's baked channel is the default.
- Docs: `docs/evener-hub.md` "Updating the hub from Settings". Spec and plan under `docs/superpowers/`.

## Testing

- `make lint`, `make vet`, `make test`, `make test-web` all green.
- Opt-in live e2e `EVENER_UPDATE_E2E=1 go test ./cmd/evener-hub/ -run TestUpdateApplyEndToEnd`: builds this branch stamped as a snapshot build, runs it with a temp HOME, applies, and verifies the real snapshot was downloaded and exec'd in place: `before=f0a97a5e3 after=be70029 pid=143116`.
- Default tests stay offline; `FuzzAppWireDispatch` stubs `scheduleHubRestart` so the fuzz harness can never exec itself.

## Not in scope

Periodic background checks, re-pointing the palette "Upgrade Evener" command, Windows, rollback.

### DIFF
--- a/appwire/client.go
+++ b/appwire/client.go
@@ -639,6 +639,18 @@ func (c *Client) Upgrade(ctx context.Context, params UpgradeParams) (UpgradeResp
 	return out, err
 }
 
+func (c *Client) UpdateCheck(ctx context.Context, params UpdateCheckParams) (UpdateCheckResponse, error) {
+	var out UpdateCheckResponse
+	err := c.request(ctx, MethodEvenerUpdateCheck, params, &out)
+	return out, err
+}
+
+func (c *Client) UpdateApply(ctx context.Context, params UpdateApplyParams) (UpdateApplyResponse, error) {
+	var out UpdateApplyResponse
+	err := c.request(ctx, MethodEvenerUpdateApply, params, &out)
+	return out, err
+}
+
 func (c *Client) AuthStatus(ctx context.Context, params AuthStatusParams) (AuthStatusResponse, error) {
 	var out AuthStatusResponse
 	err := c.request(ctx, MethodEvenerAuthStatus, params, &out)
--- a/appwire/protocol.go
+++ b/appwire/protocol.go
@@ -159,6 +159,8 @@ var Methods = []MethodSpec{
 	{MethodEvenerSearch, SearchParams{}, SearchResponse{}, ScopeHub, "Searches live and persisted sessions for the hub command palette."},
 	{MethodEvenerHarnessesList, HarnessListParams{}, HarnessListResponse{}, ScopeHub, "Lists available harness descriptors."},
 	{MethodEvenerUpgrade, UpgradeParams{}, UpgradeResponse{}, ScopeHub, "Performs or reports a evener binary upgrade."},
+	{MethodEvenerUpdateCheck, UpdateCheckParams{}, UpdateCheckResponse{}, ScopeHub, "Compares the running hub build against a release channel's current commit; dev builds report applicable=false without a network request."},
+	{MethodEvenerUpdateApply, UpdateApplyParams{}, UpdateApplyResponse{}, ScopeHub, "Downloads and installs a channel's build, then execs it in place of the running hub; refused on dev builds."},
 	{MethodEvenerAuthStatus, AuthStatusParams{}, AuthStatusResponse{}, ScopeHub, "Reports auth/credential status for a provider."},
 	{MethodEvenerAuthTest, AuthTestParams{}, AuthTestResponse{}, ScopeHub, "Tests the effective credentials for one configured provider instance without starting a session."},
 	{MethodEvenerAuthLoginStart, AuthLoginStartParams{}, AuthLoginStartResponse{}, ScopeHub, "Begins an OAuth login flow; returns a flow ID and URL."},
--- a/appwire/types.go
+++ b/appwire/types.go
@@ -79,6 +79,8 @@ const (
 	MethodEvenerSearch                = "evener/search"
 	MethodEvenerHarnessesList         = "evener/harnesses/list"
 	MethodEvenerUpgrade               = "evener/upgrade"
+	MethodEvenerUpdateCheck           = "evener/update/check"
+	MethodEvenerUpdateApply           = "evener/update/apply"
 	MethodEvenerAuthStatus            = "evener/auth/status"
 	MethodEvenerAuthTest              = "evener/auth/test"
 	MethodEvenerAuthLoginStart        = "evener/auth/login/start"
@@ -2048,6 +2050,41 @@ type UpgradeResponse struct {
 	RestartMessage string   `json:"restartMessage"`
 }
 
+// UpdateCheckParams selects the channel to compare the running build against.
+// Empty means the running binary's own upgrade channel.
+type UpdateCheckParams struct {
+	Channel string `json:"channel,omitempty"`
+}
+
+// UpdateCheckResponse reports the running build and what the channel points
+// at. Applicable is false for dev builds, which are never self-updated; the
+// Latest* fields are empty then and no network request was made.
+type UpdateCheckResponse struct {
+	Channel         string `json:"channel"`
+	BuildChannel    string `json:"buildChannel"`
+	CurrentVersion  string `json:"currentVersion"`
+	CurrentCommit   string `json:"currentCommit"`
+	LatestTag       string `json:"latestTag,omitempty"`
+	LatestCommit    string `json:"latestCommit,omitempty"`
+	UpdateAvailable bool   `json:"updateAvailable"`
+	Applicable      bool   `json:"applicable"`
+}
+
+// UpdateApplyParams selects the channel to install. Empty means the running
+// binary's own upgrade channel.
+type UpdateApplyParams struct {
+	Channel string `json:"channel,omitempty"`
+}
+
+// UpdateApplyResponse is returned just before the hub execs the installed
+// binary in place; Restarting is always true on success.
+type UpdateApplyResponse struct {
+	Release    string   `json:"release"`
+	Channel    string   `json:"channel"`
+	Installed  []string `json:"installed"`
+	Restarting bool     `json:"restarting"`
+}
+
 type AuthStatusParams struct {
 	Provider string `json:"provider"`
 }
@@ -3178,6 +3215,8 @@ type SettingsHubOverview struct {
 	// Commit is the git commit the binary was built from; empty in dev builds.
 	// Source: web_settings.go settingsData.HubCommit (buildinfo.GitSHA).
 	Commit string `json:"commit,omitempty"`
+	// BuildChannel is buildinfo.BuildChannel(): release, snapshot, or dev.
+	BuildChannel string `json:"buildChannel,omitempty"`
 	// ListenAddr is the hub HTTP server's bind address.
 	// Source: web_settings.go settingsData.HubAddr (cfg.HubAddr).
 	ListenAddr string `json:"listenAddr,omitempty"`
--- a/cmd/evener-hub/app_rpc.go
+++ b/cmd/evener-hub/app_rpc.go
@@ -1120,6 +1120,8 @@ const recentProjectDirsLimit = 15
 // focused controller registration.
 func registerMiscHandlers(server *appserver.Server, cfg hubcore.WebConfig, sources *appsource.Registry) {
 	appserver.HandleTyped(server.Router(), appwire.MethodEvenerUpgrade, hubUpgrade)
+	appserver.HandleTyped(server.Router(), appwire.MethodEvenerUpdateCheck, hubUpdateCheck)
+	appserver.HandleTyped(server.Router(), appwire.MethodEvenerUpdateApply, hubUpdateApply)
 	appserver.HandleTyped(server.Router(), appwire.MethodEvenerSearch, func(_ context.Context, params appwire.SearchParams) (appwire.SearchResponse, error) {
 		return hubSearch(cfg, params), nil
 	})
--- a/cmd/evener-hub/app_rpc_settings_overview.go
+++ b/cmd/evener-hub/app_rpc_settings_overview.go
@@ -50,6 +50,7 @@ func settingsHubOverview(cfg hubcore.WebConfig) *appwire.SettingsHubOverview {
 	return &appwire.SettingsHubOverview{
 		Version:        Version,
 		Commit:         buildinfo.GitSHA,
+		BuildChannel:   buildinfo.BuildChannel(),
 		ListenAddr:     cfg.HubAddr,
 		RunDir:         cfg.RunDir,
 		SpawnTimeout:   settingsSpawnTimeoutDisplay,
--- a/cmd/evener-hub/app_rpc_settings_overview_test.go
+++ b/cmd/evener-hub/app_rpc_settings_overview_test.go
@@ -232,3 +232,14 @@ func TestHubRPCSettingsOverview_MCPDiscoveredEmptyWhenConfigMissing(t *testing.T
 		t.Errorf("McpDiscovered.Servers = %+v, want empty", resp.McpDiscovered.Servers)
 	}
 }
+
+func TestSettingsHubOverviewReportsBuildChannel(t *testing.T) {
+	previous := buildinfo.Channel
+	buildinfo.Channel = "snapshot"
+	t.Cleanup(func() { buildinfo.Channel = previous })
+
+	got := settingsHubOverview(hubcore.WebConfig{})
+	if got.BuildChannel != "snapshot" {
+		t.Fatalf("BuildChannel = %q, want snapshot", got.BuildChannel)
+	}
+}
--- a/cmd/evener-hub/app_rpc_test.go
+++ b/cmd/evener-hub/app_rpc_test.go
@@ -11874,6 +11874,8 @@ func TestHubRPCRegistersExpectedHandlerSet(t *testing.T) {
 		appwire.MethodEvenerLaunchSetLayer,
 		appwire.MethodEvenerLaunchTrustRepo,
 		appwire.MethodEvenerUpgrade,
+		appwire.MethodEvenerUpdateCheck,
+		appwire.MethodEvenerUpdateApply,
 		appwire.MethodModelList,
 		appwire.MethodEvenerTasksList,
 		appwire.MethodEvenerJobsList,
--- a/cmd/evener-hub/app_update.go
+++ b/cmd/evener-hub/app_update.go
@@ -24,10 +24,11 @@ import (
 // Test seams: the GitHub check, the exec-in-place restart, and the restart
 // goroutine's log destination.
 var (
-	runHubUpdateCheck            = selfupdate.Check
-	scheduleHubRestart           = scheduleHubRestartAfterResponse
-	execHubBinary                = selfupdate.Restart
-	hubUpdateStderr    io.Writer = os.Stderr
+	runHubUpdateCheck             = selfupdate.Check
+	scheduleHubRestart            = scheduleHubRestartAfterResponse
+	execHubBinary                 = selfupdate.Restart
+	hubRestartSupported           = selfupdate.RestartSupported
+	hubUpdateStderr     io.Writer = os.Stderr
 )
 
 // hubRestartDelay is a short grace period for the kernel to flush the apply
@@ -122,6 +123,13 @@ func hubUpdateApply(ctx context.Context, params appwire.UpdateApplyParams) (appw
 	channel, err := validateUpdateChannel(params.Channel)
 	if err != nil {
 		return appwire.UpdateApplyResponse{}, err
+	}
+	// Fail fast where the in-place restart cannot run: without this the
+	// apply would download and install, report Restarting:true, and only
+	// then fail in the post-response goroutine -- leaving the frontend to
+	// poll a hub that never restarts. hubUpgrade (no exec) stays allowed.
+	if !hubRestartSupported() {
+		return appwire.UpdateApplyResponse{}, errors.New("in-place restart is not supported on this platform; use evener/upgrade and restart manually")
 	}
 	if err := tryLockHubUpdate(); err != nil {
 		return appwire.UpdateApplyResponse{}, err
--- a/cmd/evener-hub/app_update.go
+++ b/cmd/evener-hub/app_update.go
@@ -131,6 +131,26 @@ func hubUpdateApply(ctx context.Context, params appwire.UpdateApplyParams) (appw
 	if !hubRestartSupported() {
 		return appwire.UpdateApplyResponse{}, errors.New("in-place restart is not supported on this platform; use evener/upgrade and restart manually")
 	}
+	// Re-check server-side before installing: the store refuses stale
+	// checks, but a direct RPC (or a check that went stale in flight)
+	// would otherwise download, reinstall, and exec an identical build --
+	// and the frontend poll keys on a version change, so a no-op restart
+	// ends in a misleading restart-timeout. Runs before the lock: there
+	// is nothing to serialize when no install follows.
+	fresh, err := runHubUpdateCheck(ctx, selfupdate.CheckOptions{
+		Channel:    channel,
+		CurrentSHA: buildinfo.GitSHA,
+	})
+	if err != nil {
+		return appwire.UpdateApplyResponse{}, err
+	}
+	if !fresh.UpdateAvailable {
+		return appwire.UpdateApplyResponse{
+			Release:    fresh.LatestTag,
+			Channel:    fresh.Channel,
+			Restarting: false,
+		}, nil
+	}
 	if err := tryLockHubUpdate(); err != nil {
 		return appwire.UpdateApplyResponse{}, err
 	}
@@ -335,8 +355,11 @@ func scheduleHubRestartAfterResponse(ctx context.Context, pin restartPin, binary
 			// O_CLOEXEC so a successful Exec drops the lock exactly as
 			// the new image takes over; on failure release it here.
 			// A short timeout: a stuck holder must abort the restart,
-			// not strand the hub past its health-poll window.
-			relockCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+			// not strand the hub past its health-poll window. The bound
+			// stays well under the frontend's 30s restart poll (plus the
+			// post-response delay above) so a contended-but-successful
+			// restart still lands before the UI declares a timeout.
+			relockCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 			release, err := selfupdate.AcquireInstallLockForVerify(relockCtx, pin.shareBinDir)
 			cancel()
 			if err != nil {
--- a/cmd/evener-hub/app_update.go
+++ b/cmd/evener-hub/app_update.go
@@ -1,0 +1,359 @@
+package hub
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"primeradiant.com/evener/appwire"
+	"primeradiant.com/evener/buildinfo"
+	"primeradiant.com/evener/envvars"
+	"primeradiant.com/evener/internal/appserver"
+	"primeradiant.com/evener/internal/selfupdate"
+)
+
+// Test seams: the GitHub check, the exec-in-place restart, and the restart
+// goroutine's log destination.
+var (
+	runHubUpdateCheck            = selfupdate.Check
+	scheduleHubRestart           = scheduleHubRestartAfterResponse
+	execHubBinary                = selfupdate.Restart
+	hubUpdateStderr    io.Writer = os.Stderr
+)
+
+// hubRestartDelay is a short grace period for the kernel to flush the apply
+// response frame the send loop has just written; the restart itself already
+// waits for that write (see scheduleHubRestartAfterResponse). A var, not a
+// const, so tests can set it to 0.
+var hubRestartDelay = 500 * time.Millisecond
+
+// hubUpgradeTimeout is the one overall deadline for a complete upgrade
+// operation: archive download + checksums download + verify + install.
+// The per-request http.Client timeout (selfupdate.defaultUpgradeTimeout)
+// applies to each download separately and the two can add up, so without
+// an operation bound a valid update can outlive the frontend's apply
+// timeout and strand the page. Both hubUpdateApply and hubUpgrade funnel
+// through runHubSelfUpgradeWithTimeout. A var so tests can shrink it.
+var hubUpgradeTimeout = 4 * time.Minute
+
+// hubUpdateMu serializes evener/update/apply: copyExecutable in
+// internal/selfupdate writes a fixed dst+".tmp" path, so a second apply
+// racing the first would corrupt the binary a restart is about to exec.
+// It is deliberately left locked after a successful upgrade schedules a
+// restart -- the process is about to be replaced, so a second apply
+// afterward must also be refused, not merely serialized.
+var hubUpdateMu sync.Mutex
+
+// tryLockHubUpdate acquires hubUpdateMu for the two RPCs that can install a
+// new hub binary (evener/update/apply and evener/upgrade), returning the
+// shared "already in progress" error when the other one already holds it.
+func tryLockHubUpdate() error {
+	if !hubUpdateMu.TryLock() {
+		return errors.New("a hub update is already in progress")
+	}
+	return nil
+}
+
+// isDevBuild reports whether this hub was built without a release channel
+// (a worktree build). Such a hub is never self-updated: replacing it with a
+// release binary would silently discard whatever the developer is running.
+func isDevBuild() bool { return buildinfo.BuildChannel() == "dev" }
+
+// validateUpdateChannel resolves requested (defaulting to the build's own
+// upgrade channel) and rejects anything but "release" or "snapshot" --
+// unlike selfupdate.Upgrade's ResolveTarget, which also accepts "current",
+// "latest", and arbitrary "v*" tags. Applying an arbitrary release tag would
+// let evener/update/apply install and exec whatever the caller names.
+func validateUpdateChannel(requested string) (string, error) {
+	ch := envvars.FirstNonEmpty(requested, buildinfo.UpgradeChannel())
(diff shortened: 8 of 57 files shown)


obra:
Re the `//go:build unix` finding on `32b0f63`: `unix` is a built-in build constraint since Go 1.19 (this module is on go 1.27), so Linux and Darwin select `restart_unix.go`:

```
$ go list -f '{{.GoFiles}}' ./internal/selfupdate
[check.go restart_unix.go update.go]
$ GOOS=darwin go list -f '{{.GoFiles}}' ./internal/selfupdate
[check.go restart_unix.go update.go]
$ GOOS=windows go list -f '{{.GoFiles}}' ./internal/selfupdate
[check.go restart_other.go update.go]
```

`restart_unix_test.go` runs in CI on Linux, and the opt-in `TestUpdateApplyEndToEnd` exec'd into a freshly installed snapshot with the PID unchanged. No change made.

obra:
Both Medium findings from the combined review on `4dbfb6a` are addressed in `060b7a02b`, rebased onto current main (`c29323ae5`):

1. **Wrong-prefix install** (`app_update.go:123-126`): confirmed — the apply path passed no install dirs, so `Upgrade` defaulted to `~/.local` even when the hub runs from `/usr/local`. `hubUpdateApply` and `hubUpgrade` (the `evener/upgrade` TUI path shares the bug) now derive `Prefix`/`BinDir`/`ShareBinDir` from the running binary via `selfupdate.InstallDirsFromExecutable`; unknown layouts (worktree builds) keep `Upgrade`'s defaults. Tests: `TestInstallDirsFromExecutable{DerivesSystemPrefix,DerivesHomePrefix,FallsBackToDefaults}`, `TestHubUpdateApplyUpgradesTheRunningPrefix`, `TestHubUpdateApplyDefaultsInstallDirsForUnknownLayouts`, `TestHubUpgradeUpgradesTheRunningPrefix` — all written RED first.

2. **Unbounded download holding `hubUpdateMu`** (`update.go:100`): confirmed — nil client meant `http.DefaultClient` (no timeout) while the frontend's 30s timeout only rejects client-side. `Upgrade` now uses a 5-minute bounded default client (`defaultUpgradeTimeout`, a var so tests can shorten it). Test: `TestUpgradeStalledDownloadDoesNotHangForever` (failed RED with a 60s hang before the fix) plus `TestDefaultUpgradeTimeoutIsBounded`.

Local gates on the rebased tree, all exit 0: `go build ./...`, `go vet` (selfupdate/appserver/hub), `go test` ./internal/selfupdate ./internal/appserver ./cmd/evener-hub, `scripts/web/test-web.sh` (typecheck + vitest + biome PASS), full `make lint` (all 9 lanes PASS, including `lint-evenerfuzz`).

Note: the `tagliatelle` findings on `cmd/evener-tui/hub_notifications.go` in this review's tree are pre-existing on main and already fixed there twice over (`ec9262286` nolint, then `dc29b008c` decoding into `appwire.ToolOutputDeltaParams`); the rebase took main's version and the wire tags are untouched camelCase.

obra:
Pushed `b314be63e`: rebased onto current main (`3afe608ec`, 0 behind) adopting the newer base — the 29-commit replay was conflict-free (main's AgentsDoc additions, TurnID removal, and warningPayload removal all land in disjoint hunks from the self-update routes; wire tags untouched). Carries both fixes: the prefix-aware install + bounded download (`5fe49c930`) and the restart-goroutine test join that beats the `race-root` failure (`b314be63e`, verified locally with `-race -count=10`). Local gates on this tree: `go build`, `go vet`, `-race` update/upgrade tests, and the selfupdate suite, all exit 0.

obra:
RoboRev High finding on `67202f6` (unverified release archive) is addressed in `427307ec0`, rebased onto current main:

**Confirmed**: `selfupdate.Upgrade` downloaded, extracted, installed, and (via apply) exec'd the release archive with zero integrity checking, while `install.sh` fails closed on `checksums.txt` -- and the snapshot release does publish `checksums.txt`. All three `Upgrade` call sites affected (`evener upgrade` CLI, `evener/upgrade`, `evener/update/apply`).

**Fix**: `Upgrade` now fetches the release's `checksums.txt`, requires exactly one entry for the archive (fail closed on missing/ambiguous, mirroring `install.sh`'s wording), and compares SHA-256 before extraction -- so a tampered archive never reaches the prefix, let alone an exec. Fixing it in `Upgrade` covers all three paths at once.

**Tests** (TDD, RED first): `TestUpgradeVerifiesArchiveChecksum`, `TestUpgradeRejectsTamperedArchive` (no binary lands in the prefix), `TestUpgradeRefusesMissingChecksumEntry`, `TestUpgradeRefusesAmbiguousChecksumEntry`, `TestParseChecksumEntry`; existing archive-serving fixtures (selfupdate x3, CLI upgrade) updated to serve `checksums.txt`.

**Stated limit** (in code + docs): checksums travel the same GitHub TLS transport unsigned -- stops corruption and asset-swaps, not a both-files compromise the way a signature would.

**Local gates on this tree, all exit 0**: `go build ./...`, full `./internal/selfupdate`, `./cmd/evener`, and `./cmd/evener-hub` suites.

obra:
RoboRev findings on `6904d66` (1 High, 2 Medium) are addressed in `ac27249c9`, rebased onto current main:

**High, confirmed** -- symlinked launch prefix (`app_update.go:155`, `update.go:183`): proven with a probe binary that `os.Executable()` returns the symlink path unresolved, so a hub launched via the normal `<prefix>/bin/evener` entrypoint matched neither layout and fell back to `~/.local`. `InstallDirsFromExecutable` now resolves symlinks (`filepath.EvalSymlinks`, falling back to the raw path) and accepts both the share layout (`<prefix>/share/evener/bin`) and the bin layout (`<prefix>/bin`). Tests: `...ResolvesBinSymlink`, `...AcceptsBinDirDirectly`, `...ResolvesShareSymlink`, plus hub-level `TestHubUpdateApplyResolvesBinSymlinkEntrypoint` -- all RED first.

**Medium, confirmed** -- apply timeout vs server bound (`hubUpdate.ts:145`): apply used the default 30s client timeout while the hub may download+verify for 5min. Apply now passes `APPLY_TIMEOUT_MS` (6min, clearing the server bound with headroom); check keeps the default (server `Check` bounds itself at 10s). Test asserts the timeout opt and the bound relationship; `FakeClient` records opts (omitted when absent, so existing exact-shape assertions are unaffected).

**Medium, confirmed** -- unbounded download + in-memory hash (`update.go:248,299`): download now caps at `defaultMaxArchiveBytes` (128MB vs ~40MB archives, var for tests; over-cap body removed, hard error) and the checksum streams from disk via `io.Copy` into the hash. `install.sh` itself caps nothing, so this is strictly stronger than the installer. Tests: `TestUpgradeRefusesOversizedArchive`, `TestDefaultMaxArchiveBytesIsBounded`, `TestVerifyChecksumStreamsFromDisk`.

**Local gates on this tree, all exit 0**: `go build ./...`, full `./internal/selfupdate`, `./cmd/evener`, `./cmd/evener-hub` suites, `hubUpdate` vitest (17), `tsc --noEmit`, Biome on touched files.

obra:
Follow-up on `ac27249c9` (now `f85a8009b`): CI found two own-goals in the batch, both fixed with no behavior change:

- `lint-golangci`/`static`: `modernize stringscutprefix` on the two `HasSuffix`+`TrimSuffix` pairs the symlink fix introduced in `InstallDirsFromExecutable` -- now `strings.CutSuffix` (verified: `modernize` lane 0 issues).
- `web`: `hubUpdates.test.tsx` asserted the apply call as `{method, params}` exactly, but the timeout fix adds `opts` to the recorded call -- assertion now includes `opts: {timeoutMs: APPLY_TIMEOUT_MS}`.

Verified locally: affected vitest files (23 tests), `tsc --noEmit`, Biome on touched files, full selfupdate suite, `go build ./...` -- all exit 0.

obra:
Update on the open Medium (timeout mismatch, `dc9c874` verdict) plus branch hygiene:

**Timeout mismatch, fixed in this push.** `hubUpdateApply` and `hubUpgrade` now funnel through `runHubSelfUpgradeWithTimeout` under one overall `hubUpgradeTimeout` (4min) covering both downloads + verify + install -- stacked per-request 5min timeouts can no longer push a valid update past what the page waits for. The frontend `APPLY_TIMEOUT_MS` (6min) clears it, pinned by tests on both sides (`TestHubUpgradeTimeoutDefaultClearsServerBound` + the hubUpdate assertion). Tests RED first (`TestHubUpdateApplyBoundsTheWholeUpgrade`, `TestHubUpgradeBoundsTheWholeUpgrade`).

**Collapsed to 2 logical commits** as requested: docs spec+plan, then the feature with all review fixes. Tree verified byte-identical to the pre-collapse tip before pushing.

**Local gates on this tree, all exit 0**: `go build ./...`, `go vet` (4 pkgs), full `./internal/selfupdate` + `./cmd/evener` suites, hub update/upgrade tests under `-race -count=2`, affected vitest/tsc/Biome from the earlier batch (unchanged since).

obra:
Findings on `fa0c3c8` (1 Medium, 1 Low) addressed in `2a634fcfd` (fast-forward on the collapsed stack, no rebase needed -- 0 behind):

**Medium, confirmed** -- custom `BINDIR` lost (`app_update.go:173-178`, `update.go:193-212`): install.sh/building.mk/CLI all accept `BINDIR`/`EVENER_SHARE_BINDIR`, but derivation reconstructed `<prefix>/bin` from the resolved share path. `InstallDirsFromExecutable` now keeps the unresolved entrypoint dir as `binDir` when a same-named sibling there resolves to the managed binary (custom or standard), else the standard `<prefix>/bin`. Deliberately still unsupported: hardlink/copy entrypoints with no symlink -- indistinguishable from unrelated copies, and guessing wrong redirects symlinks; noted in-test. install.sh always installs a symlink at BINDIR, so the supported layout is covered. Tests RED first: `TestInstallDirsPreservesCustomBinDir` (+ hardlink probe that proved the boundary, then removed in favor of the note).

**Low, confirmed** -- tmp leak on rename failure (`update.go:485`): every other `copyExecutable` failure path removed tmp; `renameFile`'s error returned directly. Now removed too. Test RED first: `TestCopyExecutableRemovesTempOnRenameFailure`.

**Local gates, all exit 0**: full `./internal/selfupdate` suite, hub update/upgrade tests under `-race`, `modernize` lane, `go build ./...`.

obra:
Findings on `2a634fc` (1 High, 1 Medium) addressed in `486068cdd` (fast-forward, 0 behind):

**High, confirmed -- and worse than described.** Traced it: on Linux `os.Executable()` pre-resolves `/proc/self/exe` to the share target, so entry==share dir and my sibling check passed trivially, returning binDir==shareBinDir. `installExtractedBinaries` would then `Remove` the just-copied binary and symlink it to itself. My darwin tests couldn't catch it (darwin returns the link path). Fix: trust the entrypoint dir only when it DIFFERS from the resolved share dir, else standard `<prefix>/bin`. Test RED first: `TestInstallDirsFromPreresolvedShareBinary` (failed exactly as traced before the fix).

**Medium, partially confirmed.** `getJSON` indeed set no `User-Agent` -- but Go sends `Go-http-client/1.1` by default, which GitHub accepts, so no live breakage (verified: the RED test showed the default on the wire). Still fixed: every GitHub API request now sends `evener/<version>` for rate-limit attribution and transport independence. Test: `TestCheckSendsProductUserAgent`.

**Local gates, all exit 0**: full `./internal/selfupdate` suite, hub update/upgrade tests under `-race`, `modernize` lane, `go build ./...`.

obra:
Findings on `486068c` (1 Medium, 2 Low) addressed in `be785fb7e` (fast-forward):

**Medium, confirmed** -- custom BINDIR invisible on Linux: `os.Executable()` pre-resolves `/proc/self/exe`, so no derivation from it can ever see the custom symlink. `hubInstallDirs` now prefers argv[0] -- the unresolved invocation path the kernel does not rewrite -- with `Executable` as fallback (mirroring `currentExecutable`'s own fallback order). This is the fix shape the review suggested (preserve the invocation path at the hub layer). Custom share layouts resolve through the same path. Test RED first: `TestHubUpdateApplyPrefersInvocationPath` simulates Linux exactly (pre-resolved Executable + symlinked argv[0]).

**Low, confirmed** -- docs said "5-minute deadline": reworded to the 4-minute overall operation deadline (5 minutes is the per-request internal).

**Low, confirmed** -- `hubProcessArgs()[1:]` panics on empty argv: `hubRestartArgs` guards it, mirroring `currentExecutable`'s `len(args) > 0` pattern. Test RED first (plus a table for the non-empty shapes).

**Local gates, all exit 0**: hub update/upgrade/restart tests under `-race`, `gofmt`, `go vet`, `go build ./...`.

obra:
Findings on `be785fb` (2 Medium, 1 Low) addressed in `0b646f5cb` (fast-forward):

**Medium, confirmed** -- bare-PATH + relative argv[0]: `hubInstallDirs` passed argv[0] raw, so `evener hub` via PATH fell back to the pre-resolved share binary and relative paths compared inconsistently. New `canonicalInvocationPath`: `exec.LookPath` for bare names, `filepath.Abs` for relative paths, before any derivation. Test RED first: `TestHubUpdateApplyResolvesBarePathInvocation` (failed with the fallback bindir before).

**Medium, confirmed** -- arbitrary `EVENER_SHARE_BINDIR` fell to `~/.local`: suffix matching is inherently closed-world. When the launch entrypoint is a live symlink into a managed binary outside the standard layouts, both dirs now derive directly from the link (entrypoint dir + resolved target dir; prefix best-effort). Test RED first: `TestInstallDirsFromArbitraryShareLayout`.

**Low, confirmed** -- stale check applicable during re-check: `runCheck` now clears the result at start (sequence-guarded write still applies the fresh one), the Update button is disabled while `checking`, and `apply()` refuses a channel-mismatched result. Tests RED first (both failed before: result survived, generic error swallowed the guard).

**Local gates, all exit 0**: Go feature battery (selfupdate + hub), hub tests under `-race`, 25 frontend tests across both files, `tsc --noEmit`, Biome on touched files, `go build ./...`.

obra:
Findings on `51df0a3` (2 Medium, 2 Low) addressed in `fa418a040` (fast-forward):

**Medium, confirmed** -- restart loses custom entrypoint: the restart exec'd `result.Installed` (share binary), so after the first update argv[0] was the share path and the next derivation fell back to `<prefix>/bin`. New `restartBinary`: exec the canonical argv[0] entrypoint when it resolves to the installed binary, else the installed path. Custom BINDIR now survives across restarts. Test: `TestHubRestartPreservesEntrypoint` (+ fallbacks).

**Medium, confirmed** -- check blocks the serial worker: `evener/update/check` was not in `concurrentDispatchMethod` (only ThreadRead/TurnsList/SubagentPreview), so up to 2x10s of GitHub I/O held the worker. Added to the concurrent set -- safe: read-only, no shared-state writes, frontend discards stale results by sequence. Dispatch-set pin test updated.

**Low, confirmed** -- deadline skipped extract/install: `extractReleaseArchive`/`installExtractedBinaries`/`copyExecutable` took no ctx. All three now take ctx with cancellation checks per entry/copy (plus pre-copy and pre-rename). Test: `TestUpgradeHonorsCancellationDuringInstall`.

**Low, confirmed** -- `PREFIX=/` rejected: `CutSuffix` on a root-joined dir yields `""`, failing the non-empty check into `~/.local`. Empty rest now maps to the separator root. Test: `TestInstallDirsAcceptsFilesystemRootPrefix`.

**Local gates, all exit 0**: hub tests under `-race`, full selfupdate + appserver suites, `modernize` lane, `go build ./...`.

obra:
Findings on `fa418a0` (1 Medium, 1 Low) addressed in `8596389a0` (fast-forward):

**Medium, confirmed** -- cross-process install race: `hubUpdateMu` is process-local; CLI/hub concurrency races `Remove`+`Symlink` into EEXIST holes or mixed pairs. `installExtractedBinaries` now holds an `flock` on `.install.lock` in the managed dir (moves with custom share layouts; covers hub + CLI + second hub) and swaps entrypoint symlinks atomically (temp link + rename: readers see old or new, never missing). Off-unix the lock is a documented no-op where atomic renames still keep single files intact. Tests RED first: 4-way concurrent `Upgrade` serialization + swap atomicity (+ `wg.Go` modernize).

**Low, confirmed** -- fuzz harness did real GitHub I/O: `FuzzAppWireDispatch` stubbed upgrade + restart but not `runHubUpdateCheck`, and `evener/update/check` is in the fuzzed catalog (release ldflags bypass the dev-build guard). Stubbed offline like the upgrade seam, restored on cleanup.

**Local gates, all exit 0**: full selfupdate suite, hub tests, `modernize` lane, `go vet`, `go build ./...`.

obra:
Findings on `8596389` (1 Medium, 1 Low) addressed in `4fde24c20` (fast-forward):

**Medium, confirmed** -- ctx-blind flock: `acquireInstallLock` blocked in `LOCK_EX` past the 4min overall deadline (my own comment admitted polling would complicate). New `acquireInstallLockCtx`: `LOCK_EX|LOCK_NB` polls on a 50ms tick observing ctx, returning `ctx.Err` on expiry. Install path passes the operation ctx through. Tests RED first: held-lock + 100ms ctx fails fast; free lock still grants.

**Low, confirmed** -- mixed-release pair: sequential per-binary commit left evener-new + evener-dev-old on mid-install failure. Install is now stage-both (temp files) then commit (rename + symlink swap) with rollback to pre-install contents on any commit error. Also removes dead `copyExecutable` (superseded by `stageExecutable`), migrating its tests. Test RED first: injected second-binary failure restores both to the old release.

**Local gates, all exit 0**: full selfupdate suite, hub tests under `-race`, `modernize`+`unused` lanes, `go vet`, `go build ./...`.

obra:
Findings on `4fde24c` (1 High, 2 Medium) addressed in `59ca6b71a` (fast-forward):

**High, confirmed -- my rename fallout.** `install_lock_other.go` still defined only the old `acquireInstallLock` while `update.go` calls `acquireInstallLockCtx`: `GOOS=windows vet`, cross-compile tests, and the fuzz seed replay all failed. Added the matching Ctx no-op (verified `GOOS=windows go vet` clean). Same fallout class fixed alongside: fuzz `package_union` still referenced the renamed-away `TestCopyExecutable*` tests, and a `gocritic stringXbytes` on the new rollback test.

**Medium, confirmed** -- unbounded expansion: entries copied with no uncompressed cap and ctx checked only between entries. Extraction now caps total uncompressed output at 512MB (binaries ~100MB installed) with a ctx-aware reader per entry, so one giant entry neither fills the disk nor outlives the deadline. Test RED first: 600MB-of-zeros bomb refused.

**Medium, confirmed** -- lock loop retried every error: non-contention failures (permissions, bad fd) spun to the deadline -- and the shadowing hid a `%!w(nil)` formatting bug. Only `EWOULDBLOCK`/`EAGAIN` retry now; anything else fails fast.

**Local gates, all exit 0**: full selfupdate suite, `GOOS=windows` + linux vet, `modernize`+`unused` lanes, `go build ./...`.

obra:
Finding on `59ca6b7` (1 Medium) addressed in `6ef0c0b67` (fast-forward):

**Confirmed** -- rollback covered only `shareBinDir`: a mid-commit failure permanently replaced a custom `binDir` symlink and left absent entries dangling. The stage phase now snapshots each `binDir` entry (symlink target or absence); rollback re-points custom links and removes swapped-in links for absent entries. One documented boundary: a pre-existing *non-symlink* entry cannot be reconstructed (rename-over destroys the bytes), so the link points at the restored binary rather than guessing. Test RED first (`TestInstallPairRestoresBinDirEntrypoints`: custom link untouched, no dangle, old pair restored).

**Local gates, all exit 0**: full selfupdate suite, `GOOS=windows` vet, `modernize`+`unused` lanes, `go build ./...`.

obra:
Finding on `6ef0c0b` (1 Medium) addressed in `71f27e853` (fast-forward):

**Confirmed** -- lock released before delayed exec: the install lock drops at `Upgrade` return while `syscall.Exec` runs ~500ms later after the response flush, and a concurrent installer can swap the binary in the window. Holding `flock` through the exec cannot fix it (Go opens `O_CLOEXEC`, so the lock drops at exactly the moment it matters, plus blocking the updater is wrong) -- so the fix pins identity instead: SHA-256 recorded at schedule time (lock still held), re-verified immediately before `Exec`; on mismatch the restart aborts loudly, the old hub keeps running, and `hubUpdateMu` is released. Never execs a different release than the response reported. Test RED first: `TestHubRestartAbortsWhenBinarySwapped`.

**Local gates, all exit 0**: hub update/upgrade/restart tests, `gofmt`, `go vet`, `go build ./...`.

obra:
Findings on `63cad2e` (3 Medium, 2 Low) addressed in `0206432dc` (fast-forward):

**M1+2, confirmed (incl. my own wrong comment)** -- the pin was taken after lock release and errors collapsed to a skip. Now `Upgrade` returns `InstalledSHA256` computed while the install lock is held; apply pins the entrypoint to the under-lock digest, and a missing digest or unreadable binary aborts instead of skipping. Stubbed tests use temp-backed digests (no fictional paths reach a real restart).

**M3, confirmed** -- binDir links now snapshotted pre-commit and restored/removed on rollback; restore errors joined, not discarded.

**L1, confirmed** -- restore writes via temp+rename, like the install path.

**L2, confirmed** -- the tar/gzip stream is wrapped in a global counting cap so bytes `Next()` skips still count against the 512MB bound and ctx.

**Local gates, all exit 0**: hub tests under `-race`, full selfupdate suite, `gofmt`, `go vet`, `go build ./...`.

obra:
Update on the two open findings plus branch hygiene:

**Medium + Low (lock-before-digest, binDir rollback), fixed in this push.** `installExtractedBinaries` now returns the per-binary SHA-256 map computed inside the flock (`digestsUnderLock`), so `InstalledSHA256` pins exactly the committed bytes; the restart goroutine re-acquires the install lock for the verify+`Exec` critical section (O_CLOEXEC drops it on success; 30s abort bound on contention). Rollback snapshots `binDir` entries pre-commit and restores/removes them independently of `hadPrev`, joining errors; restores write via temp+rename; the tar stream has a global counting cap. Stubbed tests use temp-backed digests. All TDD RED->GREEN.

**Collapsed to 2 logical commits** as requested (docs + feature with all review fixes); tree verified byte-identical pre/post collapse, then rebased onto current main (0 behind).

**Local gates on this tree, all exit 0**: `go build ./...`, `go vet` (+windows), full `internal/selfupdate` + `internal/appserver` + `cmd/evener-hub` suites.

obra:
Review response for the e303c7bdf verdict (1 Medium): confirmed and fixed in 04a1f31ea.

**Medium — ignored tar entries bypass extraction cap (`update.go:485-488`): confirmed.** The `continue` on unwanted entries lets `tar.Reader.Next()` drain the skipped entry's full decompressed body on the next call, outside `extractedTotal` and outside `ctxReader` (ctx is only checked at loop top, so a single giant skipped entry decompresses unbounded between checks). Same zip-bomb class as the per-entry gap fixed earlier, just on the skip path.

**Fix:** drain skipped entries through the counting ctx reader (`io.Copy(io.Discard, ctxReader(ctx, io.LimitReader(...)))`) so skipped bytes charge the budget and observe ctx. I kept tolerance for unexpected entries (drain-and-count) rather than rejecting them outright, so releases can add files without breaking older hubs.

**Tests:** new `TestExtractCountsIgnoredEntries` (600MB-compressible ignored member) failed pre-fix with `did not contain evener-dev` and passes now; full `internal/selfupdate` suite + vet + gofmt + modernize green.

obra:
Review response for the 04a1f31ea verdict (2 Medium + 1 Low): all confirmed and fixed in 1bd978664.

**M1 — apply misreports Restarting on platforms without exec-in-place (app_update.go:171, restart_other.go:7): confirmed.** hubUpdateApply returned Restarting:true unconditionally; on !unix the exec fails only in the post-response goroutine (abort log + lock release) after the RPC already reported success, so the frontend polls 30s then times out on an old hub. Fix: selfupdate.RestartSupported() (true on unix, false elsewhere) + a fail-fast guard at the top of hubUpdateApply before any lock/download — apply on Windows now errors immediately pointing at evener/upgrade + manual restart. hubUpgrade (never execs) stays allowed. Test: TestHubUpdateApplyRefusesUnsupportedRestart (seam-stubbed, zero upgrades/restarts asserted).

**M2 — apply head-of-line blocks the serial worker (server.go:1318): confirmed.** executeOrdered runs non-set methods inline (server.go:1604); apply holds the worker up to the 4min hubUpgradeTimeout. Safe to add: hubUpdateMu + the cross-process install lock fail a concurrent apply fast, no session state is touched, and the restart keys off its own AfterResponseWritten flush rather than connection order. Fix: MethodEvenerUpdateApply joins the concurrent set (cap is 16, no starvation concern); exact-set test + set comments updated. evener/upgrade left serial — it is a fast local install with no network phase, and its caller execs manually.

**L1 — apply() ignores updateAvailable (hubUpdate.ts:150): confirmed.** The Update button disables on !updateAvailable (hubUpdates.tsx:130) but the store checked only null + channel staleness, so a direct store call (or any button-state bypass) issued a full download + exec restart for an up-to-date check. Fix: early return with applyError 'Already up to date' when !check.updateAvailable. Test: new vitest asserting no apply RPC, the error string, and restarting=false.

**Gates:** hub suite (77s) + appserver + selfupdate + hubUpdate store (20) + hubUpdates section suites green; tsc + Biome clean; gofmt + vet incl. GOOS=windows + modernize clean.

obra:
Review response for the 04a1f31ea verdict's remaining Medium (rollback hadPrev/binDir, update.go:611): confirmed and fixed in 52102beea. (The verdict's other 2 Medium + 1 Low were already addressed in 1bd978664 with response 5596755829.)

**Medium — restore() skips binDir entries when hadPrev is false: confirmed.** restore() removed the dst then hit continue, never touching the binDir link the commit loop had already swapped in. On a first-time install (no prior managed binaries) a mid-commit failure therefore left a dangling symlink after its target was deleted. My earlier binDir-rollback work only covered the hadPrev=true branch — all existing rollback tests pre-seeded shareBin, so this path was untested.

**Fix:** handle the binDir entry first in restore(), independently of hadPrev (remove-if-absent, re-point-if-custom), then handle the managed binary as before. Non-symlink pre-existing entries keep the documented leave-the-link behavior.

**Tests:** new TestInstallPairRollbackRemovesFreshLinks (empty share/bin dirs, injected second-binary failure) failed pre-fix with the swapped-in link left behind and passes now; existing rollback tests (hadPrev paths) still green; full selfupdate suite + vet incl. GOOS=windows + gofmt + modernize green.

obra:
Review response for the 52102beea verdict (2 Medium + 3 Low): all confirmed and fixed in d21ebdaf2 (rebased onto current main, 0 behind).

**M1 — committed=true before digestsUnderLock (update.go:650-651): confirmed.** A digest failure left the swapped pair live while the install reported failure and no restart ran. Fix: digest first, restore() on digest error, mark committed only after digests succeed. Test: TestInstallPairRestoresOnDigestFailure.

**M2 — rollback cannot restore non-symlink binDir entries (update.go:622-625): confirmed** (my own staged-struct comment admitted it). A copied executable is a supported layout, but swapSymlink's rename-over destroyed it. Fix: snapshot regular-file bytes pre-commit, restore them in rollback (remove-then-write over the swapped-in link). Non-regular entries (dirs) keep the documented leave-the-link behavior — rename-over cannot be undone without a full copy, and a directory masquerading as the entrypoint is outside the supported layouts. Test: TestInstallPairPreservesCopiedEntrypoint.

**L1 — short-SHA prefix collision (check.go:87): confirmed** (.goreleaser.yml:37 stamps GitSHA={{.ShortCommit}} = 7 chars; any shorter input matching by prefix claimed currency). Fix: minCurrentSHALen=7 floor — anything shorter than what we stamp fails open to update-available. Deliberately not full-SHA comparison: that would change Version() display + stamping + every fixture for no behavioral gain on real builds. Test: TestCheckShortPrefixIsStale; existing 7-char fixtures unaffected.

**L2 — 30s relock vs 30s poll (app_update.go:339 vs hubUpdate.ts:24): confirmed.** Worst case 30.5s exceeds the poll window. Fix: 10s relock bound (comment pins the arithmetic).

**L3 — no server-side up-to-date guard (hubUpdateApply:118): confirmed.** The earlier store guard is client-side only; a direct RPC reinstalled + exec'd an identical build and the poll timed out on an unchanged version. Fix: runHubUpdateCheck inside apply before the lock, return Restarting:false (with the fresh tag/channel) when already current; the store honors restarting:false instead of polling. Tests: TestHubUpdateApplyReturnsEarlyWhenUpToDate (zero upgrades/restarts asserted) + a vitest for the no-poll path. Drive-by fix: every hubUpdateApply test now stubs runHubUpdateCheck — several were silently hitting live GitHub (0.09s round trip here), violating the no-network default-tests policy.

**Gates on this tree, all exit 0:** hub suite (82s) + appserver + selfupdate + hubUpdate store (21) + hubUpdates section, tsc + Biome, gofmt + vet incl. GOOS=windows + modernize. Rebased 6/6 clean (main added only a session-menu change, no file overlap).

obra:
Quality pass complete: simplify-code + roborev-refine (5 iterations). Pushed `b9843b64c` (5 commits on top of `d21ebdaf2`, 0 behind main).

**simplify-code** (4-angle review — reuse, simplification, efficiency, altitude):
- copyStream seam on ignored-entry drain path (was bare io.Copy)
- chargeExtractionBudget helper deduplicates drain + copy budget checks
- linkFile read deferred to just before swapSymlink (error-path I/O only)
- hubUpdateApply composes through hubUpdateCheck handler, not the raw seam
- setBuild defaults an offline check stub (no live GitHub in default tests)
- Frontend: typed UpdateApplyResponse (no redundant cast), ALREADY_UP_TO_DATE const

**roborev-refine** (5 fix-review iterations, all fixable findings resolved):
- Abort on unreadable existing binary (permission error ≠ absence)
- Atomic restore via temp+rename (ETXTBSY-safe)
- Resolved-dir identity check before selecting binDir (symlinked ancestor safety)
- Abort on entrypoint snapshot failure (don't destroy an unbackupable file)
- Clean up all staged files on read error (not just the current one)
- Poll on mid-request disconnect instead of declaring failure
- Recognize socket-closed error in disconnect recovery

**Remaining findings (skipped — outside diff scope):**
- Zombie child processes across exec: pre-existing hub process-lifecycle architecture, not introduced by this diff
- Download-vs-commit verification: requires embedding the commit SHA in the release archive (goreleaser/stamping change)
- E2E test SHA collision: opt-in test (EVENER_UPDATE_E2E=1), test-only fix

**Gates:** hub suite + selfupdate + appserver + frontend (21 tests) + tsc + Biome + gofmt + vet incl. GOOS=windows + modernize — all green.

obra:
Review response for the b9843b6 verdict (2 Medium): both confirmed and fixed in 0979550f8.

**M1 — sentButLost misses RequestTimeoutError (hubUpdate.ts:202): confirmed.** A timeout after the server received the request (install completed, response lost) was treated as a normal failure — the UI skipped waitForNewHub and never reloaded.

**M2 — sentButLost regex too broad, catches server-side errors (hubUpdate.ts:202): confirmed.** A server-side error like `dial tcp: network is unreachable` matched 'network' and was misclassified as a transport disconnect, causing a spurious 30s poll.

**Root fix (per reviewer's suggestion):** replaced the regex with type-based classification. `WireError` means the server responded with a failure → report error. Everything else except 'cannot call' (pre-request rejection) means the transport failed mid-request → poll for the new hub. This correctly handles `RequestTimeoutError` (not a WireError → poll) and server-side network errors that arrive as `WireError` (server responded → report).

**Gates:** 21 vitest + tsc + Biome green.

obra:
Review response for the 0979550 verdict (2 Medium): M2 confirmed and fixed in 96d9f5667; M1 skipped (requires backend change).

**M2 — neverSent classification incomplete (hubUpdate.ts): confirmed.** `ConnectionClosedError` (purposeful close) and `requireClient()`'s missing-client error fell through to the transport-drop branch, triggering spurious 30s restart polling. Fix: moved `requireClient()` outside the transport try so a missing client surfaces as an immediate error. Added `ConnectionClosedError` to the neverSent check (a purposeful close means the request never went out, not a transport drop).

**M1 — restart detection keys on version only, not commit: confirmed but skipped.** A snapshot rebuild with the same version string but new commit would time out the poll. Fix requires exposing commit or a restart epoch from the `/api/health` endpoint — a backend change outside this diff's scope.

**Gates:** 21 vitest + tsc + Biome green.

obra:
Review response for the 96d9f56 verdict (2 Medium): both assessed and skipped — outside this PR's code scope.

**M1 — plain transport error treated as sent (hubUpdate.ts:198): partial.** The race (socket closing between the readiness check and `socket.send()`) is real but extremely narrow. The fix (track delivery explicitly or a dedicated not-sent error for synchronous send failures) requires changes to `AppwireClient.request()` internals in `client.ts` — shared transport infrastructure outside this diff. The current type-based classification (`WireError` = server responded → error; `ConnectionClosedError` + `cannot call` = never sent → error; everything else = transport drop → poll) already covers the common cases correctly.

**M2 — snapshot tag moved before assets replaced (binaries.yml:99): confirmed but outside scope.** This is the same release-infrastructure concern raised in previous reviews. The fix is to reorder the CI workflow in `.github/workflows/binaries.yml` (publish archive + checksums before moving the tag), not a code change in this PR. Filing as a follow-up for the release workflow.

**Gates:** 14/14 CI checks green on this HEAD.

PR #961, opened 2026-09-07
fix(navigation): reconstruct deltas for entity-less resources

obra:
## Problem

The hub was logging this on every delta read of a resource with no entities:

```
[hub] navigation read failed: navigation delta does not reconstruct current snapshot
```

`diffNavigationSnapshots` verified its delta with `reflect.DeepEqual`. The normalizer leaves `Entities` as a nil slice whenever a resource has no entities (the manifest always, plus any emptied section, catalog, or project page), while `applyNavigationDelta` always builds an empty non-nil slice. `reflect.DeepEqual` treats those as different, so the reconstruction check failed and the read surfaced as an internal error. Introduced with entity deltas in #846; the existing conditional-read test never sent a base on its "changed" read, so the manifest delta path was untested.

## Fix

Compare the reconstruction with the same record identity the diff itself uses (`navigationSnapshotsEqual`), so nil and empty collections are equivalent, matching what `NavigationSnapshot.MarshalJSON` already does on the wire.

## Regression tests

- `TestNavigationDeltaReconstructsResourcesWithoutEntities`: diff-level, manifest and emptied section, base round-tripped through `navigationHistory` as in production.
- `TestNavigationReadV2DeltaForResourceThatLosesAllEntities`: service-level `readV2` with a base for the manifest and live section after the last session disappears.

All three cases failed with the exact production error before the fix.

## Verification

- `make lint`, `make vet` pass
- `go test ./cmd/evener-hub/... ./hubapi/...` pass
- roborev branch review (job 150): no issues found

### DIFF
--- a/cmd/evener-hub/navigation_delta.go
+++ b/cmd/evener-hub/navigation_delta.go
@@ -1,10 +1,10 @@
 package hub
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
-	"reflect"
 	"slices"
 	"sort"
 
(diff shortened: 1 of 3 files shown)


obra:
Superseded: #962 was stacked on this branch and its squash merge (c3ca6557d) carried this fix and its regression tests into main verbatim, so this PR's remaining diff is empty and conflicting. Nothing further to merge here.

PR #967, opened 2026-09-07
perf(drain-dirty): idle backoff for drain kick fan-out (ticker untouched)

obra:
Drain loop woke the expensive kickDriveTree fan-out 4x/sec per draining session even when idle. Now throttles only the kick after 16 quiet passes (4s): every 4th pass (~1s), then every 20th (~5s). Ticker still fires at 250ms, abandonment/stall/grace verdicts still run every pass, and any wake/turn/new-work signal resets to full rate — a completion never waits out a backoff. The pendingDirty-Load-skip half was deliberately NOT done (can't preserve the #140 race window within one file).

Verification (serial runner): gofmt clean; drain race suite (SubagentDrain/Owned/Run/Fatal/Finalization/Stall/Grace/WakesIdle) ok, repeated 2x; key race tests 3x; no flakes.
Risk: low-med (drain timing) — quiescence-return path still runs same-pass, proven by race suite.

### DIFF
--- a/agent/session_jobtree_drain.go
+++ b/agent/session_jobtree_drain.go
@@ -49,6 +49,42 @@ var DrainStallTimeout = 2 * time.Minute
 // under jm.mu more often than necessary.
 const drainRecheckInterval = 250 * time.Millisecond
 
+// drainIdleBackoffFullRatePasses bounds how many consecutive quiet drain passes
+// run the full kick before the loop throttles it. A quiet pass is one that
+// consumed its top-of-loop wake edge clear, ran no notification turn, and saw
+// no newly-appeared outstanding work; every one of those resets the streak
+// (see the loop). The ticker itself is NEVER reset or removed: it stays the
+// 250ms lost-wake backstop, every tick is still consumed, and only the
+// kickDriveTree fan-out (per-session rematerialize loads, watch-send delivery,
+// delegate flushes, subtree recursion) is skipped on throttled passes. All
+// clock-driven verdicts — abandonment windows, the stall watchdog, the
+// undisposed-background escalation — still run every pass, so their timing is
+// unchanged.
+//
+// The prefix is deliberately generous: deterministic drivers single-step the
+// loop with a handful of recheck ticks (the stall/grace drivers send at most
+// about half a dozen), and every one of those ticks must still kick, so the
+// full-rate window covers them with wide margin. Only a drain that sits quiet
+// for seconds — a long build, a wedged delegate — reaches the throttled tiers:
+// every 4th pass (~1s cadence), then every 20th (~5s). Any real work (a wake
+// edge, a turn, newly-appeared outstanding work) drops back to full rate, so a
+// completion never waits out a backoff to be noticed: wakes still wake
+// immediately through waitDrainWake.
+const drainIdleBackoffFullRatePasses = 16
+
+// drainIdleBackoffSlowEvery is the kick cadence once a drain has been quiet
+// past the full-rate prefix: every 4th 250ms pass, about 1s between kicks.
+const drainIdleBackoffSlowEvery = 4
+
+// drainIdleBackoffDeepAfter is the quiet-pass count past which the loop drops
+// to its slowest kick cadence.
+const drainIdleBackoffDeepAfter = 48
+
+// drainIdleBackoffDeepEvery is the slowest kick cadence: every 20th 250ms
+// pass, about 5s between kicks. The ticker still fires at 250ms — this only
+// spaces out the expensive fan-out while nothing is changing.
+const drainIdleBackoffDeepEvery = 20
+
 // isOwnedDrainJob reports whether rec is a managed job owned by sessionID. The
 // durable record does not reliably preserve a shell's original execution mode,
 // so ownership and job type are the complete drain contract.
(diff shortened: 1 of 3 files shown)


obra:
roborev follow-up (session_jobtree_drain.go:1042-1048): the missing behavioral coverage is now pinned by agent/session_jobtree_drain_backoff_test.go — TestDrainIdleBackoffKickCadence asserts exact kick counts across the 16-pass full-rate prefix, 4x slow tier, and 20x deep tier over 88 quiet passes; TestDrainIdleBackoffWakeResetsToFullRate asserts a notify edge returns the loop to full-rate kicks. Commit 098d1e443. Tests: new tests pass 3x, drain scope (TestDrain|TestKickDriveTree|TestTreeHasOutstanding|TestRematerialize|TestOutstanding) green.

obra:
Roborev Medium (session_jobtree_drain.go:1043-1050) addressed in cd79adeb7: the pass that would return quiescent now runs one final unthrottled kick + re-scan first, so forwarded terminal/watch records for unreachable children (surfaced only via renderUnreachableChildPendings inside kickDriveTree) can't be stranded by a throttled pass. Regression test TestDrainBackoffFinalKickDeliversUnreachableChildPending fails pre-fix and passes post-fix; scoped Drain|JobTree|Backoff suite green. Ticker still the 250ms lost-wake backstop; verdict timing unchanged.

obra:
Main-branch lint blocker: PR #998 (commit f45c66402) added camelCase json tags in `cmd/evener-tui/hub_notifications.go` that fail tagliatelle on current main — your lint/static failure on that file is not from your changes. Fix is up in #1013 ("fix(lint): repair main's tagliatelle and Biome import-sort findings"); once it merges, rebase/re-run and the hub_notifications.go:609-612 json(snake) errors should clear.

obra:
Cluster B (biome organizeImports) verdict: **not branch-caused — no action on this branch.**

Evidence:
- Branch-owned files (`git diff refs/remotes/origin/main...origin/perf/drain-dirty --name-only`): only `agent/session_jobtree_drain.go` + `agent/session_jobtree_drain_backoff_test.go`. This branch touches zero frontend files.
- The web-job failure (`biome ci src`, run 34178994507/job/101914020200) flags 5 files this branch does not own: `src/panes/doc/DocPane.tsx:1:1`, `src/panes/spawn/AdvancedOptions.tsx:16:1`, `src/shell/rail/Rail.tsx:1:1`, `src/shell/rail/RailRow.tsx:37:1` (+ RailRow format), `src/shell/sessionMenu/SessionMenu.tsx:13:1` — all `assist/source/organizeImports`, all FIXABLE.
- Those lines were introduced by merged PR #1001 (commit `51840808f`, perf/w2-barrel direct imports, merged 2026-09-08T01:34:06Z), whose **own web check is FAILURE** — main inherited biome-dirty files.
- This branch forked before #1001 merged (merge-base `33049e4ac`, does not contain the #1001 commit), so its CI merge commit picks up main's dirty files and `biome ci src` (whole-src, not diff-scoped) flags them.
- Same identical failure set reproduces on #1000, #1005, #1010 — main-state, not branch-state.

Disposition: fix belongs on main (one import-sort commit over those 5 files, Cluster A / main-fix lane), then rebase/merge main here. Do NOT touch these files in this branch.

obra:
Follow-up: the 5 biome-flagged files are already repaired by PR #1013 (fix/main-lint-tagliatelle-biome, cherry-pick of the import-sort hunks + tagliatelle nolints — verified same 6-file scope). No separate fix needed here; once #1013 merges, rebasing this branch onto main should clear web-lint (plus the branch-owned items already handled: F1 reducer fix on #1000, rail/index.tsx sort on #1005).

obra:
Triage on `TestDrainBackoffFinalKickDeliversUnreachableChildPending` failing in run 34178994507 (`session_jobtree_drain_backoff_test.go:257: drain returned without running its skipped-kick pass`). Verdict: **race-in-test, not race-in-code** — the final-kick fix (cd79adeb7) is correct; the test's Phase-2/Phase-3 choreography assumes a quiescence the loop is free to reach early.

Evidence chain:
- Line 257 fired inside a `select { case recheck <- …; case <-done }`. That means `done` was ALREADY closed: the drain returned *during* Phase 2's synchronous seeding (appendEvent x3 + flag clear + scan), before Phase 3 sent anything. The message text ('without running its skipped-kick pass') misdescribes what happened.
- The test's determinism comment (lines 224-226) claims appendEvent is a pure store write that raises no wake — but that is an assumption, not a pinned fact: `appendEvent` is the `store.Append` seam (agent/jobs.go:626), and `.notify()` has ~20 production call sites (session.go, session_queue.go, delegate_delivery.go, session_attention.go, …). Separately, `newSession` builds a full session (testkit_test.go) whose construction/init paths can raise a wake the test never drains before Phase 1.
- Any wake consumed at the top of a pass (`woke := takeDrainWake(wake)`, drain.go ~1018) resets `quietPasses = 0` (three reset sites: wake edge, turn ran, wake-mid-scan). One such edge anywhere in Phases 1-2 shifts the streak, so pass 17 is a KICK pass, not a skip: it renders the seeded pending through the normal path, the turn gate runs the notification turn, and the drain then sees genuine quiescence and returns — all while the test is still seeding. `select` observes `done` and reports line 257.
- The final-kick hunk itself is sound: it fires at most once per quiescence (a confirming full-kick pass returns with `skipKick == false`), resets the streak, and changes no verdict timing. The earlier run 34175799224 (same fix) failed only on `TestNoBareWallClockDeadlineInAgentTests`, not this test — consistent with a narrow timing window, not a logic hole.
- Note: this run's `web` (biome import-sort/format in DocPane etc.) and `lint-golangci` (tagliatelle `hub_notifications.go` snake-case) failures are in files this branch never touches — pre-existing/main issues, out of scope for this lane.

Suggested fix (test-only, needs an owner with box capacity): replace the exact-streak choreography with a witnessed one — e.g. have the test `kick` stub signal each invocation on a channel and wait until 17 kicks are OBSERVED before seeding Phase 2, and tolerate `done` closing early in Phase 3 with a turns-count assertion (turns==1 proves the final kick delivered regardless of which pass rendered it). I did not edit the branch: the code is innocent and CI is the verifier here.

obra:
Correction: PR #1013 (nolint approach) is closed — the Go half was the wrong fix. Merged PR #1008 (be18ad663) fixed it properly by deleting the duplicated streamDeltaChunk struct and decoding into appwire.ToolOutputDeltaParams instead, plus stating the tag contract in appwire/doc.go. The frontend import-sort half of #1013 was identical work and is covered. Action for this PR: rebase onto current main and re-run CI.

obra:
All 5 roborev findings addressed: re-kick guards on all abandonment paths (not just quiescence), wakePending carry for consumed wakes, tests synchronized on observable kick counts, prevOutstanding reset on quiet passes. Backoff savings preserved (guards fire at most once per abandonment; single wake buys one full-rate pass). CI verifies.

obra:
Follow-up: the backoff regression test is de-flaked in 8d9f5eb6b (the final-kick render raises no wake so the loop parks with work queued — the test now keeps sending recheck ticks until turns==1 and fails only on done-with-0, the exact early-return signature; 6x count=10 + 4x count=20 green, no-fix control verified at 17/0). The waitDrainWake lint finding was deliberately left alone: it has 3 live evenerfuzz-tagged callers, so the untagged lint's unused verdict is a build-tag blind spot — deleting it would break lint-evenerfuzz. Suggested owner action: //nolint:unused with a tag note, or run that lint pass tagged. CI verifies.

obra:
Lint fix in 7dd3f4983: the backoff refactor had inlined waitDrainWake's select and left the function with only fuzz-tagged callers. Restored it as the drain loop's park call, now returning (bool, error) so the consumed wake edge carries forward via wakePending (folding the earlier wake-carry fix into the function itself); seed100 callers assert the flag. Removes duplication and fixes lint; backoff behavior unchanged. CI verifies.

PR #968, opened 2026-09-07
perf(outputmatcher): reuse scan-window scratch buffers, CR-aware fast path

obra:
OutputMatcher allocated a fresh carry+chunk window and full CRLF-normalized copy per chunk. Now reuses windowBuf/anchorBuf scratch (16KB cap-release so huge chunks don't pin memory), CRLF fast path consults carryHasCR so CR-free carries skip re-scan; anchorText kept as wrapper (byte-identical). Scan decision surface UNCHANGED — full-window FindAllIndex still runs per chunk; only backing arrays reused. ScanRetained uses local scratch (matcher stays pure read).

Verification (serial runner): gofmt clean; jobstore package ok (3.0s); FuzzOutputMatcher seed replay ok; time-boxed -fuzz 120s PASS, 4.9M execs, no failures.
Risk: med (matcher aliasing) — fuzz corpus is the oracle, clean.

### DIFF
--- a/agent/internal/jobstore/jobstore_program_fuzz_test.go
+++ b/agent/internal/jobstore/jobstore_program_fuzz_test.go
@@ -445,9 +445,10 @@ func jcpWatchErrorPaths(t *testing.T) {
 		t.Fatalf("oversized seed window = %d", len(m.carry))
 	}
 	m.SeedCarry([]byte("other"))
-	carry := anchorText(append(bytes.Repeat([]byte("x"), 4), '\r', '\n'))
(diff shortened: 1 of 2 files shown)


obra:
CI follow-up: removed the dead anchorText wrapper (all production paths use anchoredFull/anchoredFresh; fuzz probe now exercises anchoredFull) and reshaped the window appends to satisfy gocritic appendAssign. Commit 05403f00d on perf/outputmatcher; golangci clean on the module, jobstore tests pass.

PR #970, opened 2026-09-07
perf(clock-tick): scope live ticks to leaf labels (ActivityTree, Spawn)

obra:
setNow lived at component/pane level so every 1s tick re-rendered whole trees. Now a TreeNowContext provider + memoized rows: only live leaves subscribe. Cadence deliberately UNCHANGED at 1s — both ticks feed true seconds readouts (formatQuietAge '12s', Loader mm:ss), so 3s would visibly coarsen them.

Verification (serial runner): ActivityTree + Spawn suites 114/114 pass.
Risk: med-low. Caveat for reviewers: ActivityTree diff is large (context+memo refactor); watch for memo staleness and the terminal-row now=0 path. No render-count profiler evidence gathered — welcome as review follow-up.

### DIFF
--- a/cmd/evener-hub/frontend/src/panes/session/chrome/ActivityTree.tsx
+++ b/cmd/evener-hub/frontend/src/panes/session/chrome/ActivityTree.tsx
@@ -1,9 +1,9 @@
 import {
-  Fragment,
   createContext,
+  Fragment,
   forwardRef,
-  memo,
   type KeyboardEvent,
+  memo,
   type ReactNode,
   useCallback,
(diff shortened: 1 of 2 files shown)


obra:
roborev follow-up (ActivityTree.tsx:473,707-771): ran Biome formatting on both touched files (import sort + handleKeyDown re-indent + long-line wrap; no logic change). Commit on perf/clock-tick. Checks: biome ci clean, tsc clean, ActivityTree+Spawn suites 114/114 pass.

PR #973, opened 2026-09-07
perf(build-cache): rebuild linux closure instead of wiping Go cache

obra:
build-linux ran go clean -cache (whole cache wipe; every later build/test recompiles from zero) to keep embedded files fresh. Now go build -a for the linux closure: embeds re-read from disk, host-GOOS cache preserved.

Verification: review-only (diff: 7 insertions, 6 deletions, one recipe).
Risk: low. Caveat: -a still recompiles the whole linux closure each run — CPU cost stays; the win is cache preservation. Follow-up candidate: content-hash freshness check instead of -a.


obra:
roborev follow-up (building.md:68): ran `make generate` and committed the regenerated table row for the new `go build -a` recipe. `make lint-generated` passes locally. Commit on perf/build-cache.

PR #974, opened 2026-09-07
perf(watch-batch): share one journal fold; batch multi-event watch sends

obra:
Two commits, one file (agent/job_watch.go): (1) restart-notice path shares one readAllLocked+Fold instead of Load()+LoadEvents() back-to-back (Load = readAllLocked+Fold, so provably equivalent); no cache added. (2) Multi-event watch-send groups go through AppendBatch (one fsync, all-or-nothing); single-event appends stay on the appendEvent seam so fault-injection harnesses keep intercepting.

DURABILITY DISCLOSURE: multi-event groups change from N fsyncs (crash could leave a prefix) to all-or-nothing per batch — intended improvement, sync-fault suite covers the batch path.

Verification: scoped re-review clean — TestStableDelegateWatch_StopFencesAndDrainsBothReceiptClasses PASS x2, jobstore package PASS, no new breakage in fix hunk. (An earlier revision bypassed the test's appendEvent stub; fixed by gating batch on len(events)>1.)
Risk: low-med (crash-atomicity boundary change, disclosed above).

### DIFF
--- a/agent/job_watch.go
+++ b/agent/job_watch.go
@@ -3155,14 +3155,15 @@ func (jm *jobManager) noticeUnrestoredWatchEnds() error {
 	if err := jm.notifyRestartCancelledCallbackWatches(); err != nil && jm.emit != nil {
 		jm.emit(events.EventWarning, events.WarningData{Message: fmt.Sprintf("restart callback cancellation: %v", err)}, nil)
 	}
-	recs, err := jm.store.Load()
-	if err != nil {
-		return err
-	}
+	// One journal read for both consumers: LoadEvents returns the raw events
+	// and Fold derives the same job records Load would fold from them, so this
+	// path does one readAllLocked+fold instead of two. No cache is added — the
+	// events are a local of this call only.
 	stored, err := jm.store.LoadEvents()
 	if err != nil {
 		return err
 	}
+	recs := jobstore.Fold(stored)
 	spoke := watchGenerationsThatSpoke(stored)
 	for _, watch := range jm.watchesLostAtRestore {
 		if watch.SendTo == "" || spoke[watchFrameOrigin{watchID: watch.WatchID, generation: watch.Generation}] {
--- a/agent/job_watch.go
+++ b/agent/job_watch.go
@@ -3617,10 +3617,16 @@ func (jm *jobManager) recordWatchSends(deliveries []watchSendDelivery) (tokens [
 	}
 	deliveries = jm.snapshotWatchSendFrames(deliveries)
 	for _, d := range deliveries {
-		state, _, ok, err := jm.recordWatchSend(d)
-		if err != nil || !ok {
+		state, _, ok, _ := jm.recordWatchSend(d)
+		if !ok {
 			continue // recordWatchSend already produced diagnostics/drops
 		}
+		// A partial-persist failure still returns ok=true with the persisted
+		// state: the pending frame is already journaled and committed to the
+		// runtime map, so it owes the owner a wake like any recorded send —
+		// otherwise the frame stalls until unrelated activity. The failure
+		// itself is already queued as a diagnostic at the persist site, so it
+		// coexists with the token/kick below.
 		recorded = true
 		if state.Key.ResolvedSendTo == runtimeMessageAliasCaller {
 			tokens = append(tokens, watchSendTokenNotification("", state))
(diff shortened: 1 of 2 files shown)


obra:
roborev follow-up (job_watch.go:4625-4641): the AppendBatch branch is now exercised by TestAppendWatchSendEventsBatchesMultiEventGroups — multi-event co-generated groups (pending + cap-overflow evictions per planWatchSendPending) go through jm.appendEvents in exactly one call (one fsync, all-or-nothing, errors propagate); single-event appends stay on the appendEvent seam. Production routing is unchanged (the len>1 gate was already the batch path); the test pins the seam contract at the unit level. Commit on perf/watch-batch. New test passes 2x; StableDelegateWatch fence + watch-send scope green.

obra:
Roborev Low (job_watch.go:3920,4634) addressed in 08c50978c: persistPendingWatchSend now persists the pending event plus cap-overflow eviction terminal events as one AppendBatch group (single fsync, all-or-nothing), with TestPersistPendingWatchSendBatchesCapOverflowEvictions covering it. A lone pending event stays on the appendEvent seam. The TestGoalContinuationTurnCarriesItsNameOnTheOpeningEvent survey failure triages as unrelated: it passes on-branch with and without the fix and across 5 repeats, and this branch touches only watch-send files.

obra:
Triage: TestGoalContinuationTurnCarriesItsNameOnTheOpeningEvent survey failure is a base flake, not branch-caused. No code change.

Test: agent/session_active_turn_test.go:29. Path: ProcessInputKind(EntryContinuation) -> acceptContinuationInput (agent/session_lifecycle.go:2058-2077) emits EventGoalContinuation carrying the stableTurnID minted by mintRunningTurnID (agent/session_active_turn.go:72). No jobstore, no watches, no watch-send activity on that path.

Branch surface (agent/job_watch.go only): (1) noticeUnrestoredWatchEnds shares one journal fold (LoadEvents+Fold instead of Load+LoadEvents — same records, restart-restore path only, never runs in this test); (2) persistPendingWatchSend batches pending+eviction events into one AppendBatch group; (3) appendWatchSendEvents routes len>1 groups to jm.appendEvents, singles stay on appendEvent. None is reachable from a goal-continuation turn with no watches installed. Watch-send batching cannot affect turn naming.

Both-side evidence:
- R2 (prior occurrence, comment 2026-09-08T00:48Z): passes on-branch with and without the fix, 5x repeats green; branch touches only watch-send files.
- This occurrence: run 34174488095 (head 08c50978c) failed in the agent-shards survey pass (suite-red abort, no assertion detail); prior run 34164242647 (head 23b3b06aa) failed the same way; intermediate run 34154484127 (head 273fd0d79) succeeded. Fail/success/fail across heads with zero related code change is the flake signature. Main run 34177913082: tests job green (its failures were web/lint/static, unrelated).
- Zero local go test/build/vet run per lane constraint; triage by code reading + CI logs. CI re-run is the confirmation.

obra:
Third strike on `TestGoalContinuationTurnCarriesItsNameOnTheOpeningEvent` in the agent-shards SURVEY pass (run 34174488095, also 34164242647; branch head 08c50978). Verdict: **base flake, not branch-caused** — do not edit the branch for this.

Evidence chain:
- The branch diff is `agent/job_watch.go` + a new `agent/watch_batch_seam_test.go` only. Nothing in the watch-send path touches `Session`, `ConsumeEventsLossless`, `ProcessInputKind`, or goal-turn naming.
- The test itself is a race-in-test: `ConsumeEventsLossless` (agent/session_events.go:96-101) drains on a spawned goroutine, but the test asserts `len(continuations) == 1` synchronously right after `ProcessInputKind` returns (agent/session_active_turn_test.go:48-56). It never waits on the `drained` channel it created (line 34). A loaded CI runner that hasn't scheduled the drain goroutine yet reads 0 and fails.
- The failure carries no message detail (survey log only prints the `--- FAIL` line), consistent with a count of 0 rather than a wrong StableTurnID.
- The survey distortion is real but INNOCENT here: agentshards.go:270 runs the survey at `-test.parallel 6` over the whole package, which raises scheduling jitter vs. isolated/sharded runs — but that jitter only exposes the test's own missing synchronization; the branch contributes no timing-sensitive code to this path.
- History: the tests job on recent main runs is green (the current main failures are web/lint-golangci/static only), and this same test failed identically on the previous branch head 23b3b06a (run 34164242647) while the branch code was unchanged in this area.

Suggested action: re-run the tests job. Longer-term, the durable fix belongs to the test (await the drain, e.g. poll `continuations` with a deadline or wait on `drained` after Close), not to this perf branch.

obra:
Both roborev findings addressed: eviction snapshots now drop before stable-enqueue verification (durable+runtime agree on all error paths, 9dbd954e5), and the nil-seam path keeps the sequential protocol with prefix-commit semantics instead of a non-atomic group (e424b766a), with regression tests for both (a6acbcdb8). CI verifies.

obra:
Follow-up fix in 354170381: recordWatchSend was masking persisted=true to ok=false on partial-persist errors, so the new regression tests (which assert through the wrapper) could never pass — a real wrapper bug, not a stale assumption. Now surfaces (persistedState, cfg, true, perr) on partial failure; nothing-durable shape unchanged. All existing ok=false callers audited compatible. CI verifies.

obra:
Both roborev findings addressed: durable-persist prefix now wakes the owner even when an error accompanies it (recorded + caller token set on ok==true regardless of err), and the stable-enqueue verification block is extracted to a shared helper used by both branches. Regression test fails pre-fix / passes post-fix; scoped suites green. CI confirms on re-run.

obra:
Blocked by a main-branch deadline-audit failure (missing TRIPWIRE marker in agent/session_jobtree_drain_backoff_test.go), not by this PR's diff. Fix: https://github.com/prime-radiant-inc/evener/pull/1020 — once it merges, a rebase/retry should go green.

PR #975, opened 2026-09-07
perf(markdown-stream): throttle live re-parse via paragraphs-only window; share lexer

obra:
Two wins: (1) shared default-options markdownLexer for reasoningFormat (renderer stays private to the widget instance — lexer output never renders). (2) Live re-parse throttle: past 2000 chars with a paragraphs-only streaming tail, the settled head is served from a prefix-keyed cache and only the tail window re-parses per render; anything with block structure (headings/lists/fences/tables/HTML blocks/link-defs, whole-source imbalance) falls back to the exact pre-throttle full parse. Settled path (live=false) byte-untouched.

An earlier revision rendered tails via parseInline and degraded block content mid-stream; fixed by gating + full block parse in the tail. Mid-stream previews on the windowed path show paragraph content only — final HTML always identical.

Verification (scoped re-review at b8d2c8bc): tsc clean; 125/125 scoped suites pass; mixed 40-section fixture IDENTICAL via fallback (14586/14586); paragraphs-only long stream IDENTICAL via windowed path (12124/12124) — fast path proven taken AND correct. No new breakage in fix hunks.
Risk: med (streaming render path) — gates are conservative (fallback = today's behavior).

### DIFF
--- a/cmd/evener-hub/frontend/src/panes/session/transcript/messages/reasoningFormat.ts
+++ b/cmd/evener-hub/frontend/src/panes/session/transcript/messages/reasoningFormat.ts
@@ -1,12 +1,13 @@
 // Pure helpers for the reasoning ("think block") item type. See
 // ThinkBlock.tsx for the component that consumes these.
 
-import { Marked, type Token, type Tokens } from "marked";
+import { type Token, type Tokens } from "marked";
+import { markdownLexer } from "../../../../widgets/markdown";
 import { pendingTextJoined } from "../../../../protocol/reducer";
 
-// Reuse the app's existing Markdown tokenizer so nested links, block prefixes,
-// and emphasis are handled as syntax rather than accumulated regex cases.
-const markdownLexer = new Marked({ gfm: true });
+// Lexer-only use of the app's shared Markdown tokenizer (see
+// widgets/markdown/index.tsx) so nested links, block prefixes, and emphasis
+// are handled as syntax rather than accumulated regex cases.
 
 const ISO_TIMESTAMP = /^(\d{4}|[+-]\d{6})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})(?:\.(\d+))?(Z|[+-]\d{2}:\d{2})$/;
 
--- a/cmd/evener-hub/frontend/src/panes/session/transcript/messages/reasoningFormat.ts
+++ b/cmd/evener-hub/frontend/src/panes/session/transcript/messages/reasoningFormat.ts
@@ -1,9 +1,9 @@
 // Pure helpers for the reasoning ("think block") item type. See
 // ThinkBlock.tsx for the component that consumes these.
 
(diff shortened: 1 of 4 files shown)


obra:
roborev follow-up (index.tsx:198 + :206-208): (1) HEAD_SPLIT_HAZARD now matches blockquote-nested (`> [x]:`, incl. multi-level) and list-item-nested (`- [x]:`, `1. [x]:`, incl. quoted) link definitions — verified against the shared lexer; (2) CDATA start fixed to `<![CDATA[` (was `[CDATA[`, missing `<!`) and `]]>` added to HTML_BLOCK_END plus both CDATA shapes to TAIL_BLOCK_MARKER, so a split inside a CDATA block falls back. Also ran biome write on both touched files (import sort/type-only, escape fix — these were failing `biome ci` on the branch head). Verified: live-vs-settled HTML identical on nested-def, CDATA, and paragraphs-only long fixtures; tsc clean; markdown+reasoningFormat suites 121 pass. Commit on perf/markdown-stream.

obra:
Roborev Medium (index.tsx:192-204) addressed in 5d2ba8630: the windowed live-parse gate now consults the shared marked lexer for link-reference definitions (escaped labels, list-continuation indentation, nested containers included) instead of regexes alone, falling back to full parsing whenever a possible definition is present. 4 new equivalence tests fail pre-fix and pass post-fix; tsc clean; 90/90 markdown vitest green; biome clean. Settled path byte-identical.

obra:
All 5 roborev findings addressed: bounded tail window, cached head-gate verdicts per head text, UI-free lexer.ts module shared by both importers, fail-closed token traversal, CRLF split handling plus an engagement-proving test. tsc clean; 22 files / 481 tests green. CI confirms on re-run.

obra:
Blocked by a main-branch deadline-audit failure (missing TRIPWIRE marker in agent/session_jobtree_drain_backoff_test.go), not by this PR's diff. Fix: https://github.com/prime-radiant-inc/evener/pull/1020 — once it merges, a rebase/retry should go green.

obra:
Roborev Medium (head bb68f1b, whole-source balance scan) — fixed and verified on remote (bb68f1b5c..b712c80cf, markdown/index.tsx only).

The gate no longer calls closeOpenMarkdown(source) per render. It hoists closedTail = closeOpenMarkdown(split.tail) and checks closedTail !== split.tail, reusing it for the tail parse. Soundness: split sits on a blank line where closeOpenMarkdown resets all non-fence state, and a boundary-open fence unbalances the head itself — so given a balanced head (cached verdict), whole-balanced ⟺ tail-balanced. Per-render engaged cost is now tail close/scan/lex + tail re-parse, all O(tail); the only O(n) left is the fallback's full re-parse (pre-throttle behavior).

Test proof (delegate transcript): tsc exit 0, 92/92 markdown tests green (incl. sanitize-count engagement + CRLF), biome clean. Commit b712c80cf.

PR #977, opened 2026-09-07
Allow explicit verified force stop of local session daemons

### LINKED ISSUE #934: Allow explicit hub termination of incompatible or unresponsive session daemons

obra:
A local daemon that cannot complete AppWire initialization can still reserve its session, leaving normal shutdown unavailable. Add a confirmed **Force stop** action that verifies the exact daemon, requests termination, and confirms exit. The dialog warns that active turns and jobs may be interrupted; saved transcripts remain intact.

Linux uses pidfds and Darwin uses process-generation audit tokens. Both verify user, command, start time, and API-log inode/lock evidence before signaling. Ambiguous ownership and unsupported verification fail closed. Stable/current aliases stay reserved against resume and deletion until exit confirmation ends.

Recovery uses one short-lived connection so a saturated primary queue cannot block it. Connection and request generations prevent old queued actions from restarting the session. Explicit Resume refreshes the primary connection and submits once without replaying pending requests. A successful signal retains the explicit-resume requirement even when exit confirmation fails; the RPC still reports that failure.

Recovery obligations are persisted as atomic alias-group snapshots before signaling or reporting an already-exited stop successful. Startup restores them before admission. Only successful explicit Resume can clear them, and a failed clear reports an error while retaining recovery state. Disk I/O is separated from request admission locking. Descendants retain independent identities, but journal-verified ancestry blocks fresh actions while an owner’s exit is unconfirmed. Confirmed exit is persisted separately from explicit Resume, so this restriction survives hub recreation and releases only after confirmation or successful owner recovery.

Validation at `534cd8b1e`: `make merge-approval-gate` passed (lint, build, all Go modules, and frontend). The repaired concurrent-alias fixture passed ten runs under race; expanded recovery tests cover failed signaling, unconfirmed exit, late descendants, hub recreation, and confirmation-write failures. All five browser guards passed on the unchanged frontend runtime. All checks in GitHub CI run34298741948 passed on this head. Local explicit branch review222 and hosted member6653 found no issues. Combined review6654’s sole Darwin UID-offset finding was disproved against Apple’s SDK layout, a compiled live-kernel probe, and native race tests; it is documented and closed ([evidence](https://github.com/prime-radiant-inc/evener/pull/977#issuecomment-5594587243)). No actionable findings remain.

Based on main after #936 merged. No old-protocol compatibility or automatic restart is added.

Limits: snapshot version3 includes current-session identity and verified exit confirmation; unsupported older authority prevents startup, with no compatibility migration. Unresolved process ownership or genuinely competing identities still refuse Resume. Durable recovery authority applies to the same HubStateRoot and its single-hub lock. Interrupted termination conservatively retains an explicit-Resume requirement; unreadable recovery authority prevents startup. Changing/removing that root or running an older binary that ignores it is outside this guarantee. Linux conservatively refuses rendezvous timestamps inside the process start tick. Darwin combines kernel descriptor evidence with Evener's close-on-release API-log lifecycle; it does not claim protection against deliberate same-user process impersonation.

Closes #934.

The PR contains two logical commits: backend behavior and regression tests, followed by frontend behavior and regression tests.

### DIFF
--- a/agent/execenv/sandbox_lifecycle_test.go
+++ b/agent/execenv/sandbox_lifecycle_test.go
@@ -252,13 +252,13 @@ func TestCleanupRetainsOwnedTmpAfterChildGrace(t *testing.T) {
 	sentinel := filepath.Join(dir, "tmp-was-present")
 	ready := filepath.Join(dir, "trap-installed")
 
-	// The child touches READY only AFTER the TERM trap is installed, so the test can
-	// wait for readiness before signaling — otherwise Cleanup could SIGTERM the child
-	// before its trap exists (the default action would terminate it, masking the
-	// ordering under test).
-	script := `trap 'if [ -d "$WATCH" ]; then : > "$SENTINEL"; fi; exit 0' TERM
+	// Publish readiness after starting the sleeper and installing the trap. Bash
+	// can defer TERM during asynchronous child startup, so no fork may remain
+	// between READY and wait: readiness must cover the entire signal setup.
+	script := `sleep 300 &
+trap 'if [ -d "$WATCH" ]; then : > "$SENTINEL"; fi; exit 0' TERM
 : > "$READY"
-sleep 300 & wait`
+wait`
 	h, err := env.StreamCommand(context.Background(), script, dir,
 		map[string]string{"WATCH": tmp.Dir, "SENTINEL": sentinel, "READY": ready}, io.Discard)
 	if err != nil {
--- a/agent/status.go
+++ b/agent/status.go
@@ -3,7 +3,9 @@ package agent
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"sort"
 	"time"
@@ -279,6 +281,75 @@ func SessionOwnsDelegate(ctx context.Context, stateDir, parentSessionID, childSe
 		return false, fmt.Errorf("delegate ownership journal has an incomplete trailing batch: %s", path)
 	}
 	return false, nil
+}
+
+// SessionOwnedDelegateIDs reads the session's subtree in the root-owned
+// delegate journal without loading transcripts. Fork provenance is not ownership.
+func SessionOwnedDelegateIDs(ctx context.Context, stateDir, sessionID string) ([]string, error) {
+	if err := schema.ValidateSessionID(sessionID); err != nil {
+		return nil, err
+	}
+	rootID := sessionID
+	meta, err := schema.LoadSessionMeta(stateDir, sessionID)
+	if err == nil {
+		rootID = activityRootIDFromMeta(sessionID, meta)
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return nil, err
+	}
+	if err := schema.ValidateSessionID(rootID); err != nil {
+		return nil, err
+	}
+	path := filepath.Join(jobsDir(stateDir, rootID), "delegates.jsonl")
+	result, err := historicalDelegateFoldCache.Get(ctx, path, extendHistoricalDelegateFold)
+	if errors.Is(err, os.ErrNotExist) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+	// Committed descriptors remain ownership evidence after interrupted writes.
+	// A killed daemon may leave an incomplete trailing batch; it is not replayed.
+	children := make(map[string][]string)
+	var pending []string
+	for delegateID, aggregate := range result.Value.state {
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
+		if aggregate == nil || aggregate.Descriptor.OwnerSessionID != rootID {
+			continue
+		}
+		children[aggregate.Descriptor.ParentDelegateID] = append(children[aggregate.Descriptor.ParentDelegateID], delegateID)
+		if sessionID == rootID || aggregate.Descriptor.ChildSessionID == sessionID {
+			pending = append(pending, delegateID)
+		}
+	}
+	ids := make(map[string]bool)
+	visited := make(map[string]bool)
+	for len(pending) != 0 {
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
+		delegateID := pending[len(pending)-1]
+		pending = pending[:len(pending)-1]
+		if visited[delegateID] {
+			continue
+		}
+		visited[delegateID] = true
+		id := result.Value.state[delegateID].Descriptor.ChildSessionID
+		if err := schema.ValidateSessionID(id); err != nil {
+			return nil, err
+		}
+		if id != sessionID {
+			ids[id] = true
+		}
+		pending = append(pending, children[delegateID]...)
+	}
+	out := make([]string, 0, len(ids))
+	for id := range ids {
+		out = append(out, id)
+	}
+	sort.Strings(out)
+	return out, nil
 }
 
 // LoadSessionDelegateStatus projects a cold session's stable delegate rows
--- a/appwire/protocol.go
+++ b/appwire/protocol.go
@@ -126,6 +126,7 @@ var Methods = []MethodSpec{
 	{MethodThreadReasoningEffortSet, ThreadReasoningEffortSetParams{}, EmptyResponse{}, ScopeBoth, "Sets reasoning effort, normalizing and validating the value."},
 	{MethodThreadVisionModelSet, ThreadVisionModelSetParams{}, EmptyResponse{}, ScopeBoth, "Sets the vision side-channel routing (\"\", \"off\", or a model ref)."},
 	{MethodThreadCompactStart, ThreadCompactStartParams{}, EmptyResponse{}, ScopeBoth, "Starts a context-compaction pass on the session."},
+	{MethodEvenerThreadForceStop, ThreadForceStopParams{}, EmptyResponse{}, ScopeHub, "Explicitly terminates a verified local daemon and confirms exit; saved session data is retained."},
 	{MethodThreadShutdown, ThreadShutdownParams{}, EmptyResponse{}, ScopeBoth, "Shuts the session down (the daemon runs it asynchronously)."},
 	{MethodTurnStart, TurnStartParams{}, TurnStartResponse{}, ScopeBoth, "Starts a new user turn and reserves a turn ID."},
 	{MethodTurnSteer, TurnSteerParams{}, TurnSteerResponse{}, ScopeBoth, "Injects a steering message into the active turn."},
--- a/appwire/types.go
+++ b/appwire/types.go
@@ -46,6 +46,7 @@ const (
 	MethodThreadVisionModelSet        = "thread/vision-model/set"
 	MethodThreadCompactStart          = "thread/compact/start"
 	MethodThreadShutdown              = "thread/shutdown"
+	MethodEvenerThreadForceStop       = "evener/thread/forceStop"
 	MethodTurnStart                   = "turn/start"
 	MethodTurnSteer                   = "turn/steer"
 	MethodTurnInterrupt               = "turn/interrupt"
@@ -597,11 +598,14 @@ type TaskAggregate struct {
 }
 
 type EvenerThread struct {
-	Ref        string `json:"ref"`
-	InstanceID string `json:"instanceId,omitempty"`
-	ParentRef  string `json:"parentRef,omitempty"`
-	Kind       string `json:"kind,omitempty"`
-	Profile    string `json:"profile,omitempty"`
+	// ResumeRequired means recovery stopped this session and automatic actions
+	// must wait for an explicit thread/resume. Saved transcripts remain readable.
+	ResumeRequired bool   `json:"resumeRequired,omitempty"`
+	Ref            string `json:"ref"`
+	InstanceID     string `json:"instanceId,omitempty"`
+	ParentRef      string `json:"parentRef,omitempty"`
+	Kind           string `json:"kind,omitempty"`
+	Profile        string `json:"profile,omitempty"`
 	// TurnCount is the daemon's total completed model-response count. It stays
 	// independent of Turns so a bounded metadata read never loads the transcript.
 	TurnCount        int                `json:"turnCount,omitempty"`
@@ -1663,6 +1667,10 @@ type TurnCancelQueuedResponse struct {
 }
 
 type ThreadCompactStartParams struct {
+	Ref string `json:"ref"`
+}
+
+type ThreadForceStopParams struct {
 	Ref string `json:"ref"`
 }
 
--- a/cmd/evener-hub/app_compact.go
+++ b/cmd/evener-hub/app_compact.go
@@ -39,7 +39,7 @@ func compactThreadWithResume(ctx context.Context, cfg hubcore.WebConfig, sources
 	if !shouldResumeAfterSessionUnavailable(err) {
 		return err
 	}
-	if _, resumeErr := hubThreadResume(ctx, cfg, sources, appwire.ThreadResumeParams{Ref: params.Ref}); resumeErr != nil {
+	if _, resumeErr := hubThreadAutoResume(ctx, cfg, sources, appwire.ThreadResumeParams{Ref: params.Ref}); resumeErr != nil {
 		return resumeErr
 	}
 	return compactThreadOnce(ctx, cfg, sources, params)
--- a/cmd/evener-hub/app_force_stop.go
+++ b/cmd/evener-hub/app_force_stop.go
@@ -1,0 +1,344 @@
+package hub
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"slices"
+	"time"
+
+	"primeradiant.com/evener/agent"
+	"primeradiant.com/evener/appwire"
+	"primeradiant.com/evener/cmd/evener-hub/internal/appsource"
+	"primeradiant.com/evener/cmd/evener-hub/internal/daemonprocess"
+	"primeradiant.com/evener/cmd/evener-hub/internal/hubcore"
+	"primeradiant.com/evener/rendezvous"
+)
+
+// forceStopThread bypasses daemon RPC only after verifying the local process.
+// The session remains reserved until the exact process has exited.
+func forceStopThread(ctx context.Context, cfg hubcore.WebConfig, params appwire.ThreadForceStopParams, sources *appsource.Registry) (stopErr error) {
+	ref, err := appwire.ParseRef(params.Ref)
+	if err != nil || ref.SourceID != "local" {
+		return appwire.InvalidParams("force stop requires a local session ref")
+	}
+	if cfg.RunDir == "" || cfg.ResumeLocks == nil {
+		return appwire.Unavailable("local session ownership is not configured")
+	}
+	recoveryTarget := cfg.ResumeLocks.RecoveryState(ref.ThreadID).ResumeSessionID
+	entry, err := forceStopEntry(cfg.RunDir, ref.ThreadID, cfg.DaemonProcesses, nil, recoveryTarget)
+	if err != nil {
+		return appwire.Unavailable(err.Error())
+	}
+	// Verify the process before interrupting RPCs. Kill verifies this retained
+	// process handle again after ownership and deletion reservations are held.
+	controller := cfg.DaemonProcesses
+	if controller == nil {
+		controller = daemonprocess.NewController()
+	}
+	sessionID := entry.SessionID
+	if sessionID == "" {
+		sessionID = entry.ThreadID
+	}
+	target := daemonprocess.Target{PID: entry.PID, SessionID: sessionID, StateDir: entry.StateDir, StartedAt: entry.StartedAt}
+	process, err := controller.Open(target)
+	exited := errors.Is(err, daemonprocess.ErrExited)
+	if err != nil && !exited {
+		return appwire.Unavailable(fmt.Sprintf("cannot verify daemon for force stop: %v", err))
+	}
+	defer func() {
+		if process == nil {
+			return
+		}
+		if err := process.Close(); err != nil {
+			log.Printf("force stop process handle cleanup: %v", err)
+		}
+	}()
+	if exited && recoveryTarget != "" && sessionID != recoveryTarget {
+		return appwire.Unavailable("exited daemon claim does not match session recovery authority")
+	}
+	// Descendants share the stopped process, but keep independent transcript
+	// identities. Their old requests must expire even after root ownership clears.
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	fenced := make(map[string]bool)
+	var descendantFinishes []func(bool)
+	defer func() {
+		for _, finish := range descendantFinishes {
+			finish(false)
+		}
+	}()
+	fenceDescendants := func(scanCtx context.Context) error {
+		if entry.StateDir == "" {
+			return nil
+		}
+		ids, err := agent.SessionOwnedDelegateIDs(scanCtx, entry.StateDir, sessionID)
+		if err != nil {
+			return appwire.Unavailable(fmt.Sprintf("read daemon delegates: %v", err))
+		}
+		var added []string
+		for _, id := range ids {
+			if !fenced[id] {
+				fenced[id] = true
+				added = append(added, id)
+			}
+		}
+		if len(added) != 0 {
+			descendantFinishes = append(descendantFinishes, cfg.ResumeLocks.BeginForceStop(added))
+		}
+		return nil
+	}
+	// These are admission fences, not aliases or durable resume targets. Hub
+	// restart drops the old connections and queued requests altogether.
+	if err := fenceDescendants(ctx); err != nil {
+		return err
+	}
+	aliases := forceStopAliases(entry)
+	finishRecovery := cfg.ResumeLocks.BeginForceStop(aliases)
+	defer func() { finishRecovery(stopErr == nil) }()
+	if sources != nil {
+		if source, ok := sources.Source("local"); ok {
+			if local, ok := source.(*appsource.LocalDaemonSource); ok {
+				release := local.BeginRecovery(entry)
+				defer release()
+			}
+		}
+	}
+	// Clear gives one daemon stable and current session aliases. Lock both so
+	// resume or deletion through either alias cannot race exit confirmation.
+	for _, id := range aliases {
+		cfg.ResumeLocks.For(id).Lock()
+	}
+	defer func() {
+		for _, alias := range slices.Backward(aliases) {
+			cfg.ResumeLocks.For(alias).Unlock()
+		}
+	}()
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	currentTarget := cfg.ResumeLocks.RecoveryState(ref.ThreadID).ResumeSessionID
+	if currentTarget != recoveryTarget {
+		return appwire.Unavailable("session recovery authority changed; retry force stop")
+	}
+	if err := forceStopOwnershipUnchanged(cfg.RunDir, ref.ThreadID, entry, cfg.DaemonProcesses, currentTarget); err != nil {
+		return appwire.Unavailable(err.Error())
+	}
+	if err := deletionFenceError(cfg, params.Ref, ref.ThreadID, ""); err != nil {
+		return err
+	}
+	if exited {
+		// There is no retained process handle to reverify. An unchanged marker
+		// alone cannot confirm absence after waiting for ownership.
+		process, err = controller.Open(target)
+		if !errors.Is(err, daemonprocess.ErrExited) {
+			if err == nil {
+				err = errors.New("daemon is live after initial exit observation")
+			}
+			return appwire.Unavailable(fmt.Sprintf("daemon exit is not confirmed: %v", err))
+		}
+		// An exited claim cannot replace newer authority held by any alias
+		// it would overwrite, even when the requested alias still names it.
+		for _, alias := range aliases {
+			authority := cfg.ResumeLocks.RecoveryState(alias).ResumeSessionID
+			if authority != "" && authority != sessionID {
+				return appwire.Unavailable("exited daemon claim conflicts with alias recovery authority")
+			}
+		}
+	}
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	// Commit recovery authority before the signal can take effect. Interrupted
+	// signaling conservatively retains the explicit-Resume requirement.
+	if err := cfg.ResumeLocks.PersistForceStop(aliases, sessionID); err != nil {
+		return appwire.Unavailable(fmt.Sprintf("persist session recovery: %v", err))
+	}
+	if exited {
+		if err := cfg.ResumeLocks.ConfirmForceStop(sessionID); err != nil {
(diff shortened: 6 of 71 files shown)


obra:
Verified the Darwin offset finding against the installed macOS SDK and the compiler. The proposed change would read the GID instead of the UID.

`offsetof(struct proc_bsdinfo, ...)` reports:

```
sizeof=136
pbi_uid=20
pbi_gid=24
pbi_pgid=100
pbi_start_tvsec=120
pbi_start_tvusec=128
```

These match the existing verifier. The SDK declaration is in `usr/include/sys/proc_info.h`, lines 59–82. The actual struct begins with flags, status, exit status, PID, and PPID (five uint32 fields), followed by UID.

At unchanged head `4b80d0f7e`, reran `go test -race ./cmd/evener-hub/internal/daemonprocess -run '^TestNativeController' -count=1 -v` on Darwin: PASS. The successful case opens, verifies, signals, and confirms exit of a test-owned locked child. Unlocked/released-lock and replaced-inode refusals also pass.

This finding is a false positive; no code change is warranted.

obra:
Both findings in hosted review 5764 were checked against the production paths and are false positives.

Cancellation: `withClientCallMapper` passes the registered recovery context to `client.Start`. The receive loop calls the shared WebSocket with that context. The pinned coder/websocket implementation closes the connection on read-context cancellation, releasing both pending reads and blocked writes. The client marks itself closed and fails pending requests; the wrapper returns the recovery context cancellation, preventing automatic Resume. `TestHubForceStopInterruptsStalledRPCOnSameConnection` covers shutdown and model-set over real WebSockets with a fake Kill that cannot close the daemon socket.

Darwin: Apple explicitly defines `proc_uniqidentifierinfo` as 56 bytes: a 16-byte UUID, unique ID at offset 16, parent ID at 24, and pid version at 32. The kernel fills these fields and returns this structure's size. The implementation's size and offsets match.

Sources: [Apple ABI declaration](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/sys/proc_info_private.h#L41-L51), [kernel population](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/kern/proc_info.c#L1417-L1425), [copyout](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/kern/proc_info.c#L2100-L2107).

These paths are unchanged between reviewed `2bebc7b53` and current `4f1437179`. No source changes are warranted for these findings.

obra:
Confirmed the sandbox-resolution finding on current head `4f1437179`. The handler uses deletion-only ownership with an empty mutation ID, and admission capture omits this method. I am fixing both parts and adding a behavioral regression that queues an approval across Force stop and explicit Resume, verifying it cannot reach the replacement daemon. This finding remains open until the fix and reviews pass.

obra:
Fixed the sandbox approval finding in [6d2b57b75](https://github.com/prime-radiant-inc/evener/commit/6d2b57b75).

Sandbox resolution now captures its effective target at admission and uses the existing non-durable session-action guard, enforcing both the captured recovery epoch and connection sequence without inventing a mutation ID.

Regressions reproduced four bypasses (queued/unread old connection × ref/threadId). They exercise production recovery-state transitions and the actual typed admission/handler boundary, verify stale approvals never reach the replacement source, and verify one fresh approval succeeds. Ten focused race runs and scoped lint pass; independent scoped review is clean. The full canonical gate and new whole-PR review are running, so this is a verified focused fix, not a claim that the latest full review has completed.

obra:
Resolution of hosted review 5784 (`6d2b57b`), now pushed in `dbc2e5121` and `40a155884`:

- Darwin UID: repeated false positive. The installed SDK/compiler reports UID=20, GID=24, process group=100; real native ownership/kill/wait tests already pass. [Prior evidence](https://github.com/prime-radiant-inc/evener/pull/977#issuecomment-5576028426). Changing UID to offset 24 would read GID.
- Relative state directories: serve resolves the selected state directory to an absolute path before opening state and publishing rendezvous. Real serve tests cover both flag and environment paths, checking transcript and API ownership log at the advertised location.
- Queued live rename: captures admission before queue/ownership waits and validates the recovery fence before forwarding to a live daemon. Stale queued/unread requests are rejected; fresh live and offline saved-metadata renames remain supported.
- Failed exit confirmation: client triggers an authoritative background refresh while immediately preserving the original error. No speculative stop success, automatic Resume, or replay. Tests cover held/failed reads, retained uncertain messages, and a real saved-thread read after failed Wait and later rendezvous removal.

Focused race tests, lint, full frontend gate and independent scoped review pass. The combined exact-head canonical gate and new whole-PR review are still running.

obra:
Hosted review 5826 is addressed in `07539e6f6`, included in current `439c9771e`.

The already-exited path now reopens the identical target under all alias ownership locks and requires a second ErrExited result. A newly live handle is closed and refused without signaling; ambiguous evidence is refused. Tests cover exited-to-live, exited-to-ambiguous, and still-exited results with both alias locks held. All force-stop tests pass three race repetitions and scoped review is clean.

The Darwin unique-info layout claim repeats the previously disproved finding: Apple's ABI explicitly defines size 56, unique ID offset 16 and pid version offset 32. [Source-backed assessment](https://github.com/prime-radiant-inc/evener/pull/977#issuecomment-5576862755).

Additional local review fixes now included: canceled session startup cannot send initial input, and confirmed recovery is consistently accessible for local saved snapshots across sends, model changes, and compaction. The latter removes action-specific visibility rules. Focused tests and all five browser guards pass; full canonical and whole-PR reviews on `439c9771e` are running.

obra:
Hosted review 5927 is addressed in `481858139`.

- The timezone finding reproduced with real rendezvous writes and independent reads: identical timestamps with a non-hour offset were rejected because `time.Time` location pointers differed. Force-stop now compares timestamp instants while retaining exact equality for every other field, including exited-claim preference. The regression verifies unchanged metadata is accepted and a changed start instant is rejected. All force-stop tests pass three race repetitions; scoped lint reports zero issues.
- The Windows build finding is a false positive: `GOOS=windows GOARCH=amd64 go build ./cmd/evener-hub/internal/daemonprocess` succeeds. Go provides `os.Geteuid` on Windows (returning -1); this is not a Unix-only undefined symbol. Unsupported process control still fails closed.

The preceding current-base head `c262321c2` passed the complete canonical gate, all five browser guards, and local whole-PR review 185. The new backend-only fix has a fresh canonical gate and whole-PR review running. Hosted latest-head review remains pending.

obra:
Local whole-PR review 187 is addressed in `22edae4b1` (current head `7a01ea913`). Recovery persistence now stops directory synchronization at the hierarchy that already existed when the store opened. It retains that boundary across failed commits so newly created links are still synced on retry. A real traversal-only ancestor regression failed with EACCES before the fix and now persists/restores the recovery obligation successfully. Existing file/directory sync failure coverage passes.

Repeating the complete hubcore race suite also exposed shared injected-failure counters leaking between pin-section fixture runs. The separate test-only commit resets those counters at setup and cleanup. The complete hubcore package now passes three race repetitions; scoped storage lint passes.

The prior head `481858139` passed the complete canonical gate. The new head has another canonical gate and whole-PR review running. Browser code is unchanged from the five-guard pass at `c262321c2`. Hosted review remains pending; this is not a declaration of final readiness.

obra:
Hosted review6032's API-log sizing finding is a false positive, verified by the regression in `1c7ad0238`. For three through zero records, the computed envelope size equals the actual indented JSON size, including metadata changes and multiline values. `apiLogSummaryRecordSeparator` already includes the four spaces before subsequent records, so adding another per-record indentation charge would double-count them. Production sizing remains unchanged.

Local review192's independent-fork ownership finding is fixed in `285952753`: non-subagent independent activity roots stop provenance traversal, while delegate/shared-job-tree ancestry remains verified. A real fork with deleted parent metadata and an unrelated incompatible daemon reproduces the refusal and now permits resume, mutation admission and deletion ownership checks. Focused ownership/restart/delegate tests pass three race repetitions. The new integrated gate and whole-PR reviews are running.

obra:
Hosted review6080 is addressed in `c3209c3c3`, `350590acc`, and `dfe121bab`.

- Stable/current aliases: Resume resolves the retained rendezvous SessionID, reserves all known aliases in force-stop's sorted lock order, and revalidates identity plus each alias's recovery epoch/connection admission under those locks. A real two-client stable/current regression now launches the current session once and clears both obligations. Tests also cover recreated durable groups.
- Callback cancellation: all direct RPC callbacks receive the recovery-derived context; canceled calls are rejected before callback admission. Deterministic read and mutation tests cover cancellation immediately after initialization and inside transport Send. The original code admitted stale RPCs; fixed code admits none. Appsource race suite,25 focused repetitions and lint pass.
- Missing identity: multiple aliases without retained current-session evidence, or conflicting retained SessionIDs, refuse explicitly rather than guessing. Single-alias markerless recovery remains supported. Durable format is unchanged. Old crash-marker pruning can remove the needed evidence; recovering such a multi-alias group requires restoring unambiguous rendezvous identity.

Focused ForceStop/Recovery/Resume race regressions pass for hub and hubcore. Current `dfe121bab` is pushed, with canonical validation and a new whole-PR review running. Browser code is unchanged from its prior five-guard pass. Latest hosted review remains pending; this is not a final readiness claim.

obra:
Local review197 is addressed in `ceb828319` and `fa34d121e`.

- Verified current-session identity is now persisted atomically with each recovery alias/group in strict snapshot version2 before signaling. Version1 is rejected rather than guessed or migrated. A real force-stop refresh expires the marker; a recreated hub then handles concurrent stable/current Resume with one launch of the persisted current session. Before/after-rename uncertainty and overlapping-group target tests pass.
- Different retained IDs are distinguished using verified process evidence. A dead old A marker no longer blocks the sole live cleared B; unresolved or competing live ownership remains a refusal. If all claims exited, durable target authority resolves the stopped transcript.
- Explicit Resume returns its ThreadResponse to the pane. The pane replaces its route with the returned current ref and refreshes that thread. The regression uses the real AppwireClient, Session/store, transcript and Composer: B displays, new input targets B, and uncertain A input stays blocked under A without replay.

Hub/hubcore recovery regressions and race tests pass. Focused frontend tests (114) and the complete frontend gate pass. Current `fa34d121e` is pushed; combined canonical validation, browser guards, and whole-PR review are running. The previous marker-restoration limitation is removed for normally persisted recovery. Hosted review remains pending.

obra:
Local review198 is fixed in `69d7f9485`, integrated with the latest #936 in `6fc65967e`.

- Successful recovery retains completed alias routing separately from the pending durable obligation. A subsequent Resume still resolves the current transcript after marker cleanup and normal shutdown. Fresh B-to-C identity overrides the remembered A-to-B target; stopping, pending and stale-epoch callbacks cannot install completed routing.
- Resume reserves supplied Ref, requested Session and resolved target identities in the existing sorted lock order, and checks each deletion fence independently. Regression cases cover alias redirection and conflicting Ref/Session fields: committed deletion never reaches the launcher.

Both reported failures were reproduced before the fix. Focused ForceStop/Recovery/Resume tests pass under the race detector (hub8.682s, hubcore2.137s). The pushed integrated head is running the canonical merge-approval gate, all browser guards and fresh whole-PR review200. Review readiness remains pending.

obra:
Local review200 is addressed in `f71fdcc03`. Completed alias routing follows every recovery group until the current target, rejects cycles, and reserves all traversed identities in the existing sorted ownership-lock order. Resolution is repeated under those locks and existing epoch/connection fences remain enforced.

The regression performs two actual force-stop/Resume cycles, removes rendezvous markers, then resumes through the oldest alias. Before the fix it launched intermediate B; now it launches C with A/B/C reserved. Separate tests verify cycles never reach the launcher, newer pending recovery remains fenced, and an old connection cannot cross a newer completed recovery while a fresh connection reaches C. Targeted regular and race suites pass (hub21.208s, hubcore2.680s under race).

The updated head is pushed and a fresh canonical gate and whole-PR review are running. The preceding6fc65967e head passed the full canonical gate and all five browser guards; this fix changes only backend alias resolution and its tests. Hosted review remains pending, so the PR is not yet marked ready.

obra:
Hosted review6224 triage:

- Linux clock-tick bound: the conservative upper bound is intentional. The lower bound/tolerance suggested by the review cannot exclude a process that replaced the recorded PID within the same kernel tick. Such a timestamp is outside the verifier's supported evidence and remains a refusal, as documented in the PR limits. This can reject a legitimate same-tick spawn; we are preserving verified ownership rather than weakening the check.
- Per-connection recovery slot: this is intentional bounded admission, not per-session deduplication. It lets recovery bypass an occupied serial worker while refusing unbounded concurrent recovery handlers. The client uses a short-lived recovery connection and permits one pending force stop. Session alias locks separately serialize overlapping ownership. The real WebSocket overlap and saturated-primary recovery tests pass under race. No concurrency expansion is needed for this UI flow.
- Rail error refresh: the shared store already starts authoritative refresh in its forceStop catch and rethrows the original error immediately. A failed refresh cannot mask that error, and a finally in Rail would duplicate the refresh. Three focused store/rail tests pass, covering successful reconciliation, refresh failure, and retained uncertain messages.

The relay cancellation finding is still being investigated.

obra:
Local review201 is fixed in2562d55d4, integrated in386f97565. Retained competing crash markers yield to completed alias redirects only after all competing processes are verified exited. Live conflicts, unknown ownership and an unresolved exited successor still refuse. An older alias also refuses a target in another pending recovery group, even when that target is the same transcript as before.

Both reported failures reproduced with actual repeated force-stop/Resume fixtures and retained markers, then passed. Tests cover changed-target and same-target recoveries, fresh/stale older aliases, live conflicts and the unresolved-successor case. Targeted race tests pass (hub9.981s/hubcore1.691s).

Hosted6224's relay claim is not a restart or mutation bypass. Proof in198cffc75 keeps recovery held while a relay observes an externally present replacement using only initialize and subscribed reads; direct actions remain canceled. Source proof passes under race (1.357s), as do complementary hub durable-recovery, queued-action and stale-connection tests (3.795s). No relay runtime behavior changed. The Linux verification bound, per-connection admission slot and already-existing UI error refresh are explained in the prior triage comment.

The pushed386f97565 integrates current main through PR936, including its deterministic CI test repair. Fresh full canonical validation, browser guards and whole-PR review are running.

obra:
Local review206 and hosted6272 are addressed in2697f791e, e9a3fb249 and b7a74ee74.

- Repeated Force stop uses persisted verified target authority when all competing claims have exited, rejects ambiguity without authority and refuses a sole old marker when the authoritative marker is missing. It rereads authority under ownership locks and refuses a changed target. Real repeated-stop/recreated-store/Resume tests verify that C stays the recovery target. Four regressions reproduced failures; targeted race tests pass.
- ThreadFork admission now captures the source recovery epoch and checks it under the existing ownership lock. Real fork/aside routines refuse stopping, pending and stale completed-recovery requests without creating metadata; fresh admission after completion succeeds. Regression red-green, focused race2.779s.
- Automatic-resume failures preserve recovery classification plus mutation correlation and unknown/blocked retry disposition. Real AppWire races for turn/start and thread/clear reproduce the loss before the fix;10 race repetitions and scoped lint pass. Non-mutation-ID actions retain the original error.

The pushed b7a74ee74 is running fresh canonical validation and whole-PR review. Browser code is unchanged from its all-five-guard pass. Hosted reviews remain part of the completion gate.

obra:
Local review 207 is fixed in ce42450df. BeginForceStop now changes only temporary admission counters, epochs and connection fences. Alias routing changes only when PersistForceStop commits recovery authority, including an uncertain write after rename. Cancellation and failures before persistence preserve previous routing and group membership.

Regressions reproduced failed Resume after a canceled stop and marker cleanup, and broken group clearing after an overlapping write failed before rename. Both now pass. Tests also cover both aliases, stale epochs and conservative obligations after an uncertain committed write. Relevant hub and hubcore tests pass under race (10.368s and 1.795s).

The fix is pushed. Full canonical validation and a fresh whole-PR local review (208) are running. Hosted review remains pending; this is not a clean-review claim.

obra:
Hosted review 6290 reports a Darwin vnode flavor mismatch, but the requested ABI is PROC_PIDFDVNODEINFO, not PROC_PIDFDVNODEPATHINFO. Compiling against the installed macOS SDK sys/proc_info.h confirms flavor 1 has sizeof(struct vnode_fdinfo) = 176; flavor 2 has sizeof(struct vnode_fdinfowithpath) = 1200. darwinOwnsLog intentionally reads the former structure and compares device/inode metadata; it does not request a path. Changing the flavor to 2 with this buffer would introduce the failure described by the review.

Existing TestNativeControllerKillsOnlyOwnedChild runs on Darwin and verifies a real child holding a locked API-log descriptor, signals that fixture-owned process and confirms exit. It also rejects unlocked/released locks; TestNativeControllerRefusesReplacedLog rejects a different inode. The complete native daemonprocess package passed three repetitions under -race on macOS (13.642s), including the audit-token generation test.

No code change is needed for this finding. Current ce42450df already passes the full canonical gate and local whole-PR review 208 reports no issues. Hosted review 6290 is now triaged with this ABI and behavioral evidence. PR936 hosted review remains pending.

obra:
Local210 is fixed in pushed b60197fee. An exited target now checks every alias it would overwrite under ownership locks. The regression reproduced old alias A corrupting shared alias B recovery from C to B; both A refusal and B retry now preserve C through store recreation and explicit Resume. Targeted race tests pass (3.490s). Fresh whole-PR review211 and canonical validation are running.

Hosted6299 triage: the Windows claim is false. Go os.Geteuid is available on Windows and returns -1 (os/proc.go and syscall/syscall_windows.go); GOOS=windows go vet ./cmd/evener-hub/internal/daemonprocess passes. The unsupported-platform controller already refuses in bind before verification or signaling. The Linux tick upper-bound claim repeats an intentional documented limitation: lowering the bound cannot exclude PID replacement inside the same tick. Ambiguous same-tick ownership remains a refusal. No platform code change is warranted for either item.

obra:
Hosted6387 ABI findings are false positives. Apple XNU bsd/sys/proc_info_private.h defines proc_uniqidentifierinfo with a 16-byte UUID, then p_uniqueid at offset16 and p_idversion at offset32; its static assertion explicitly requires size56. Source: https://github.com/apple-oss-distributions/xnu/blob/main/bsd/sys/proc_info_private.h .

The installed macOS SDK sys/proc_info.h places pbi_uid after five uint32 fields, at offset20; pbi_pgid is later, after the command/name fields. The code intentionally compares effective UID with os.Geteuid, not real UID.

The complete native daemonprocess tests pass under race on macOS (5.398s), including actual locked-child ownership/termination, unlocked/released/inode refusal and mismatched audit-token generation. No runtime change is warranted.

obra:
Local214 is fixed in 56bd171ae. Force stop reads owned descendant identities from the verified daemon root delegate journal and invalidates their queued requests through recovery epochs and connection sequences. These temporary fences do not merge child transcripts into root aliases or change their resume targets; independent forks without ownership descriptors remain unaffected.

The regression reproduced a queued child action surviving parent force stop. It now checks stale child action and Resume refusal, no stale launch, fresh child Resume preserving its identity, and unaffected independent fork admission. Relevant force-stop/failed-stop tests pass under race (5.577s).

This head also integrates PR936 current31a52b104 and main through cf4df743a, including corrected read-only navigation test expectations. Fresh canonical validation and whole-PR local216 are running. All five browser guards passed for the integrated frontend; the subsequent changes are backend-only and test/comment changes.

obra:
Local216 is fixed in pushed0b680e382. After confirmed exit, force stop rescans the stable root ownership journal before releasing recovery fences, covering delegates created during termination. The read uses committed descriptors despite an incomplete trailing batch, so interrupted writes cannot prevent confirming an already-exited process.

Both cases reproduced red: a child created during termination escaped the old-connection fence, and a torn trailing batch blocked exit confirmation. The regressions now pass, including repeated confirmation with the torn tail and committed child fencing. Broad hub recovery/force-stop/Resume tests pass under race (11.036s), with hubcore2.241s.

Fresh canonical gate and whole-PR review217 are running. All five combined browser guards passed on56bd171ae; this follow-up changes only backend code/tests.

obra:
Local217 is fixed in22d3f129e. A connection predating recovery now checks a delegate target ancestry against committed journal ownership and retained parent recovery sequences. This covers children created during a failed exit wait and children created after the force-stop handler has already returned, without depending on a one-time final scan. Independent fork provenance does not establish ownership; children retain their own transcript targets.

Failed-confirmation regression reproduced the bypass before the fix. Broad recovery tests pass under race (hub10.803s/core2.301s); focused failed/late-child tests pass under race1.846s. Scoped hub lint reports zero issues. Fresh canonical gate and whole-PR review218 are running on the pushed head. Combined frontend browser guards remain passing; frontend code is unchanged.

obra:
Validation at `f2606e055b1c53f5f05192e811f6409fa5f79708`: `make merge-approval-gate` completed with exit 0. Lint, build, every Go module, and the full frontend typecheck/test/Biome gate passed. The worktree is clean. All five browser guards passed on the unchanged frontend runtime. Local whole-PR review218 plus the test-only follow-up review219 report no issues. Hosted review6543 is pending; this is validation evidence, not a claim that hosted review is complete.

obra:
Current-head validation at `383eb752dad32008adfdc43a939edb22d353b8cf`: `make merge-approval-gate` completed with exit 0 on a clean worktree. Lint, build, root, agent, llm, auth, envvars, invariant, identifier, and web all passed.

Hosted review 6564 is still pending its second reviewer; the completed member reports no issues. This is validation evidence, not a claim that the combined review is clean.

obra:
Review6572 is a false positive. `time.Time.UTC()` strips monotonic readings; it does not retain them. The [Go time documentation](https://pkg.go.dev/time#hdr-Monotonic_Clocks) explicitly describes that behavior for UTC, Local, and In. In the installed Go1.27 implementation, UTC calls setLoc, which calls stripMono before setting the location. Adding Round(0) after UTC would be redundant.

The existing [TestRecoveryCancelsDaemonCallsAcrossClearAliases](https://github.com/prime-radiant-inc/evener/blob/4cb924d095f16516648c3e0e70126cab9b91c592/cmd/evener-hub/internal/appsource/recovery_test.go#L13) starts a call with time.Now(), begins recovery with its UTC wall-time representation, verifies cancellation and refusal of new old-alias calls, preserves an unrelated daemon, and verifies release. `go test -race ./cmd/evener-hub/internal/appsource -run ^TestRecovery -count=1` passed now (2.410s). No production change is required.

obra:
Current-head validation at `4cb924d095f16516648c3e0e70126cab9b91c592`: `make merge-approval-gate` completed with exit 0 on a clean worktree. Lint, build, root, agent, llm, auth, envvars, invariant, identifier, and web all passed. The targeted direct-recovery race tests also passed (2.410s).

Hosted combined review6573 remains pending. Member6572 was answered with official Go documentation and existing behavioral tests disproving its monotonic-time claim. This is validation evidence, not a claim that the combined review has completed.

obra:
Addressed review6596 in ef84bcf7ec81dc844dbcb5ccf8857b5cf694217d, folded into the frontend commit so PR977 remains two logical commits.

- Failed force-stop now immediately retains a blocking obligation before asynchronous refresh. Regression coverage verifies the fence while refresh is pending, after failed refresh, and after a saved resume-required response. A separate test verifies a healthy authoritative refresh releases the fence without automatic Resume.
- Pending confirmation disables Cancel and ignores Escape/dismissal. The regression verifies the dialog remains available and re-enables retry after rejection.
- The recovery connection timeout ends after connect; the RPC then owns its full request timeout. The regression waits 25 seconds before opening the connection and returns a hub error 10 seconds later, preserving the hub error and cleaning up all timers.

The three failure-fence assertions failed before the fix. The dialog and slow-connection regressions also failed before their fixes under the pinned frontend Vitest configuration. Focused frontend tests pass (10 tests, plus the 4-test fence set including healthy recovery); Biome and git diff --check pass. Full canonical gate and browser guards are running, not yet claimed passing.

The notLoaded render-guard suggestion is a false positive. A saved snapshot cannot establish that no daemon exists: explicit Resume, turn/start, model changes, and compact can start a daemon before the first live read succeeds. Existing Session regressions exercise pending and failed resumed-daemon reads while status remains notLoaded and demonstrate that force-stop remains reachable without navigation. Hiding it would remove the recovery action for precisely those stalled sessions. The hub still verifies a direct daemon ownership claim before any signal.

obra:
Validation completed on ef84bcf7ec81dc844dbcb5ccf8857b5cf694217d: make merge-approval-gate exited 0, including lint, build, all Go modules, and frontend typecheck/tests/Biome. All five browser guards also passed. Review6596 fixes and the saved-session guard rebuttal are documented above. Fresh hosted review6603 is still pending; this is validation evidence, not a claim of a clean combined verdict.

obra:
Review6603 is fixed in 9858d488b06519ca3152f911bc640d84b85c6999, folded into the backend commit so this PR retains two logical commits.

SessionOwnedDelegateIDs now resolves JobTreeRootSessionID from metadata and follows ParentDelegateID edges from the resumed session's descriptor in that root journal. The traversal visits each edge once and retains committed descriptors after a torn tail.

The behavioral force-stop regression reproduced the missing historical descendant before the fix. Expanded coverage verifies a persisted child and grandchild, a descendant committed during termination, rejection of old requests, and unaffected original-controller and sibling sessions. All TestForceStop tests passed under the race detector (2.070s). The fresh canonical gate is running. Frontend runtime is unchanged from the head whose five browser guards passed.

obra:
The full make merge-approval-gate passed on 9858d488b06519ca3152f911bc640d84b85c6999 (exit 0): lint, build, every Go module, and the frontend gate. All force-stop tests passed under race; the unchanged frontend runtime already passed all five browser guards. The historical-descendant fix is validated. Fresh review6627 is pending its second reviewer; completed member6626 reports no issues.

obra:
Addressed review6627 in `c31faf424` (backend commit `345f45f70`).

The finding reproduces: both failed signaling and failed exit confirmation allowed a fresh descendant action to execute. Merely setting descendant ResumeRequired would create a flag without a recovery group through which it can clear. Instead, the owner now persists ExitConfirmed separately from its explicit-Resume obligation. Admission checks journal-verified ancestry while an owner remains stopping or unconfirmed, including fresh connections, restored hub state, and delegates created after a failed stop. Descendants keep independent transcript identities; confirmed termination permits their independent recovery.

The strict recovery snapshot version is 3, with no backward-compatibility path. Tests cover both failure modes, late descendants, hub recreation, successful retry, and confirmation persistence failures. Race-enabled recovery tests passed (hub 7.260s; hubcore 1.822s), and the full canonical gate passed before rebasing. PR936 has merged, so this PR now targets main; the full gate is running again on this exact rebased head. A fresh hosted review is still required.

obra:
Review6627's unconfirmed-descendant recovery fix passed the rebased canonical gate at `ab0f260cf`: lint/build, root (61.62s), agent (16.12s), llm (8.11s), auth (2.43s), envvars (1.26s), invariant (0.77s), identifier (3.00s), and frontend (79.91s). Hosted member6650 also reported no issues; the second member remains pending.

CI found a separate fixture defect in both tests and race-root: TestConcurrentAliasResumeSpawnsOneCurrentDaemon assumed synthetic PID106 was dead, whereas the Linux runner considered it alive and correctly retained its marker. Current head `534cd8b1e` injects the external liveness probe for that fixture and passes ten race repetitions. Production liveness is unchanged. The repair is folded into the backend commit, preserving two logical commits. Current-head full gate and CI are running; this is not a readiness claim.

obra:
Current head `534cd8b1e` passed `make merge-approval-gate` with exit 0: lint/build, root 65.20s, agent 17.23s, llm 8.72s, auth 2.55s, envvars 0.90s, invariant 1.08s, identifier 2.90s, frontend 85.01s. The fixture repair also passed ten focused race runs. The worktree is clean, the PR still has two commits, and it is mergeable against main.

CI run34298741948 and hosted combined review6654 remain active on this exact head. Their completion is still required before readiness.

obra:
Additional explicit branch review222 passed with no issues on `33917328a..534cd8b1e` (the PR’s two commits). All current-head CI checks and the canonical local gate pass. Hosted member6653 also found no issues; combined review6654 remains pending its second reviewer.

obra:
Review6654 / member6652 is a false positive. The existing Darwin UID offset is correct.

Apple’s [`struct proc_bsdinfo` definition](https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/proc_info.h) places `pbi_uid` after five uint32 fields. Compiling an `offsetof` probe against the installed macOS SDK gives:

```text
sizeof(proc_bsdinfo)=136
pbi_uid=20
pbi_gid=24
pbi_ruid=28
pbi_pgid=100
```

A real `proc_pidinfo(getpid(), PROC_PIDTBSDINFO, ...)` response confirms that the uint32 at offset20 is the effective UID: `kernel_uid=501 decoded_offset20=501 geteuid=501 gid=20 pgid=94423`. The suggested offset24 reads the group ID and would introduce the verification failure described by the review.

The current head’s native fixture tests also pass under race (5.339s): `go test -race ./cmd/evener-hub/internal/daemonprocess -run 'TestNativeController|TestDarwinAuditToken' -count=1`. They exercise the real macOS controller against fixture-owned children, including verified termination and refusing a different audit-token generation. No ambient processes are terminated.

No code change is warranted. Local branch review222 and hosted member6653 found no issues; all CI checks, the canonical local gate, and all five unchanged-frontend browser guards pass. Closing the disproven finding with this evidence.

PR #979, opened 2026-09-07
Personal AGENTS.md: loaded into every session, editable in settings

obra:
Slice 1 of docs/superpowers/specs/2026-09-07-settings-sheet-editors-and-agents-doc-design.md.

- The daemon loads a personal `AGENTS.md` ahead of every session's project docs. The hub hands each spawned and resumed daemon its own concrete path (`--agents-doc`, mirroring `--plugin-root`), stable delegates inherit the parent's path, and an unresolvable config root loads nothing rather than a relative file. The personal doc has its own 32KB budget, separate from the repo docs', so neither can starve the other.
- New `evener/settings/agentsDoc/{get,set}` methods and a `changed` broadcast (additive, v4). `set` writes through a symlinked `AGENTS.md` (a dotfiles-managed copy stays the source of truth), via an exclusive temp file with an explicit 0644 mode.
- New Settings → AGENTS.md section: monospace editor, dirty-gated Save, Revert, a "changed on disk" notice with Load current, and a reload-failure alert with a Reload button. The store refetches on every reconnect and fences fetches, saves, and broadcasts by request and client generation, so a stale response can never overwrite a newer document; a save superseded by another client's write says so instead of claiming a clean save.

## Review history

The three items the first review parked were ruled on by Jesse and landed here: the hub-to-session path plumbing (roborev), symlink-following saves (final review), and separate doc budgets (final review). Every subsequent roborev finding through head `0a1fa95` is fixed in code on this branch; each fix wave had its own scoped re-review.

Known, deliberately left: in the narrow case of a socket drop and recovery inside one save round-trip, the editor shows a conservative "changed on disk" note although nothing else wrote; one click on Load current clears it. Follow-ups queued as separate PRs: the same symlink and temp-file hardening for providers.toml and the other config-root writers, and surfacing the loader's unread truncation flag.

## CI note

`web`, `lint-golangci`, and the aggregate `static` lane fail on this PR because they fail on `main` itself since the `perf(w2-*)` batch (unsorted imports in five frontend files; camelCase JSON tags in `cmd/evener-tui/hub_notifications.go`). This branch is clean on both locally; a separate PR fixes main.

## Gates

`make lint`, `make vet`, `make test` (which runs `make test-web`), `make test-web-browser` — green on the branch at each wave's head.

## Live check (first head)

Against a throwaway hub with `XDG_CONFIG_HOME` in a temp dir and fake429 as the only provider: Settings → AGENTS.md rendered the config-root path, accepted typed text, and Save toasted; the file on disk matched the typed bytes exactly (same sha256), mode 0644, no leftover temp file. Reloading re-read the saved content. Desktop (1280x800) and phone (375x812) both render the section as a single column. A dormant session in a temp repo with its own `AGENTS.md` composed the two instruction blocks in order, personal first:

```
----- BEGIN <temp XDG_CONFIG_HOME>/evener/AGENTS.md -----
...
----- END <temp XDG_CONFIG_HOME>/evener/AGENTS.md -----

----- BEGIN AGENTS.md -----
...
----- END AGENTS.md -----
```

No screenshot images are attached (`gh` cannot upload an image into a PR body). Cosmetic: the config-root path renders in a non-wrapping `<Code>` element, so a pathologically long path gives the page a horizontal scrollbar at phone width; a real `~/.config/evener/AGENTS.md` does not.

### DIFF
--- a/agent/delegate_runtime.go
+++ b/agent/delegate_runtime.go
@@ -1765,6 +1765,8 @@ func (runtime delegateRuntime) restoreIdle(started delegateStartCommit) (*subage
 		LLMSleep:                s.cfg.LLMSleep,
 		clock:                   s.clock,
 		testOnly:                s.cfg.testOnly,
+		// The personal doc belongs to whoever runs the tree now, not to the run that froze this delegate.
+		AgentsDocPath:           s.cfg.AgentsDocPath,
 		TurnEndsProcess:         s.cfg.TurnEndsProcess,
 		ForceRealIO:             s.cfg.ForceRealIO,
 		artifactStore:           s.artifactStore,
--- a/agent/project_docs.go
+++ b/agent/project_docs.go
@@ -20,6 +20,7 @@ type ProjectDoc struct {
 const (
 	projectDocByteBudget = 32 * 1024
 	projectDocTruncMark  = "[Project instructions truncated at 32KB]"
+	userDocTruncMark     = "[Personal instructions truncated at 32KB]"
 )
 
 // UserDocFile is the personal instructions file evener loads from the user
@@ -34,23 +35,22 @@ func LoadProjectDocs(env execenv.ExecutionEnvironment, filenames ...string) ([]P
 }
 
 // LoadInstructionDocs is a session's whole instruction set: the personal doc
-// at userDocPath first, then the repo's project docs, sharing one byte
-// budget. The personal doc is loaded first because it is the user's own
-// standing instructions for every session; the repo docs get whatever budget
-// remains, truncated exactly as LoadProjectDocs truncates them.
+// at userDocPath first, then the repo's project docs. The two byte budgets are
+// separate, each the size of projectDocByteBudget, so a long personal doc can
+// never silence a repo's instructions and a long repo doc can never silence
+// the user's own; each side truncates with its own marker.
 func LoadInstructionDocs(env execenv.ExecutionEnvironment, userDocPath string, filenames ...string) ([]ProjectDoc, bool) {
 	out := []ProjectDoc{}
-	used := 0
+	userTruncated := false
 	if user, ok := LoadUserDoc(userDocPath); ok {
 		if len(user.Content) > projectDocByteBudget {
-			user.Content = truncateDoc(user.Content, projectDocByteBudget)
-			return append(out, user), true
+			user.Content = truncateDoc(user.Content, projectDocByteBudget, userDocTruncMark)
+			userTruncated = true
 		}
-		used = len(user.Content)
 		out = append(out, user)
 	}
-	project, truncated := loadProjectDocs(env, used, filenames...)
-	return append(out, project...), truncated
+	project, truncated := loadProjectDocs(env, 0, filenames...)
+	return append(out, project...), truncated || userTruncated
 }
 
 // LoadUserDoc reads the personal instructions file at path. ok is false when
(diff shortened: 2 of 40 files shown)


obra:
Heads-up from #981: the comment added to `TestHubRPCRegistersExpectedHandlerSet` in 8531e715f says an unset `LaunchConfigRoot` falls back to `cmdutil.DefaultConfigRoot()` "where an empty agentsDoc/set would truncate the developer's own AGENTS.md".

This package's `TestMain` redirects HOME and the XDG roots into a throwaway directory for the whole test binary, so with the field unset the write lands in `<tmp>/evener-hub-test-env-*/config/evener/AGENTS.md`, not `~/.config/evener/AGENTS.md`. Verified by probe; the redirect is present at this branch's merge-base and head. #981 adds `TestHubDefaultRootsStayInsideTheTestEnvironment` to pin that.

Setting `LaunchConfigRoot: t.TempDir()` here is still the right move, for isolation from the other parallel tests that share the throwaway root (#983 covers making the fixture do that by default). Worth rewording the comment so the next reader doesn't inherit the wrong model.

obra:
Review response for roborev Combined Review on `741d725` — all three findings fixed in `5001937e0`:

- **Medium, fetch() leaves stale error** (`agentsDoc.ts:67`): success branch now `set({ doc, loading: false, error: null })`, matching save() and the `changed` broadcast. Regression test: `a successful retry clears a stale fetch error` (seeds the error mid-flight so only the success branch can clear it; failed pre-fix with `"hub unreachable"` surviving).
- **Medium, fixtures read ambient AGENTS.md** (`testkit_test.go:73`): default `newSession` cfg now points `AgentsDocPath` at an isolated non-existent temp path (same pattern as `session_perf_test.go:702`); explicit-cfg callers untouched. Regression test: `TestNewSession_DefaultIsolatesThePersonalDoc` writes an ambient file and asserts it never reaches the session.
- **Low, untrimmed doc path** (`project_docs.go:41`): `personalDocPath` applies `strings.TrimSpace` first, matching `RestoreSessionFromMetaWithConfig` and `LoadUserDoc`. Regression test: `TestPersonalDocPath_TrimsSurroundingWhitespace`.

Local gates on the fix: `go test ./agent/` ok, `go vet ./agent/` clean, gofmt clean, `agentsDoc.test.ts` 21/21, Biome clean.

obra:
Review response for roborev Combined Review on `5001937` (assessed each finding against the producer code before acting):

- **Medium, omitted `--agents-doc` (`spawn.go:304-306`)**: partially valid mechanism, but not reachable as stated. Omission already *is* an explicit `""` passthrough: `personalDocPath("")` resolves `userdirs.Subdir(DefaultConfigRoot(), …)` and `Subdir` preserves `""` (`envvars/userdirs/userdirs.go:29-34`), so with identical env hub and child agree. In production `LaunchConfigRoot = cmdutil.DefaultConfigRoot()` (`main.go:368`) is never `""` (it substitutes a relative path), so hub `""` needs userdirs unresolvable at hub startup — and the child inherits the same env (`ParentEnv: os.Environ()`, `spawn.go:92-97`), resolving `""` too. Divergence needs a per-launch `XDG_CONFIG_HOME` override landing exactly when the hub root is unresolvable; no code sets per-launch XDG today. Rejecting the sentinel/disabled flag; if per-launch XDG overrides come into scope, the fix consistent with the `main.go:357-360` PluginRoot pattern is to pin the child env in `spawn.go`, not add a flag.
- **Low, explicit test configs bypass fixture isolation**: partially valid. Agent `TestMain` already redirects `HOME+XDG_CONFIG_HOME` to fresh temp dirs (`session_tools_worktree_create_test.go:186-190`), so real ambient `~/.config` can never leak; the residual is intra-package order-dependence (the decoy writer at `project_docs_test.go:158-180` vs explicit-cfg sessions). Rejecting global isolation and any production `NewSession`/`applyDefaults` default change. Targeted fix in progress: `AgentsDocPath` set explicitly in the doc/prompt-asserting tests that build sessions with explicit cfgs lacking it.

PR #981, opened 2026-09-07
test(hub): pin that the hub's default roots stay inside the test env

obra:
## Mechanism

`TestHubRPCRegistersExpectedHandlerSet` dispatches every registered method with empty params. For a handler where an empty request is a valid write (`evener/launch/setLayer` today; the `evener/settings/agentsDoc/set` handler in #979 tomorrow) the write happens for real, against whichever root the fixture left unset. This package's `TestMain` redirects HOME and the XDG roots into a throwaway directory, which is what keeps those writes out of `~/.config/evener`, but nothing asserted the redirect covered the roots the hub actually falls back to. Remove the redirect and every hub test silently starts writing to the developer's home again.

Investigation note: the premise in #979's commit 8531e715f, that the empty `agentsDoc/set` dispatch would truncate the developer's real AGENTS.md, does not hold. With `LaunchConfigRoot` unset the root resolves to `<tmp>/evener-hub-test-env-*/config/evener` under this package's `TestMain` (verified by probe; the redirect is present at that branch's merge-base and head). Non-test code has no `os/user` use, so nothing resolves home outside HOME/XDG.

## Change

- `TestHubDefaultRootsStayInsideTheTestEnvironment`, beside the two existing `TestMain` guards: asserts `os.UserHomeDir`, the three XDG bases, `hubLaunchConfigRoot` with the field unset, `cmdutil.DefaultStateRoot`, `plugins.DefaultRoot`, `defaultMCPConfigPath`, and the paths `runMain` opens when a flag is unset (`DefaultHubStateRoot`, `DefaultConfigPath`, `DefaultStateGlob`, `DefaultPastIndexDBPath`, `rendezvous.DefaultDir`) all resolve under `testEnvRoot`.
- Corrects the credentials-store comment in the handler-set test, which claimed an unset root reaches the developer's real `credentials.toml`.
- `TestMain` refuses to start the package when any of those roots resolves outside the throwaway env, exiting with the offending paths before `m.Run`. A failing guard test cannot stop the tests beside it; this can. The guard test shares the root list and keeps the check live mid-run.
- `USERPROFILE` is redirected beside `HOME`, so what `os.UserHomeDir` reads is contained on Windows too.
- Containment resolves symlinks through the deepest existing ancestor of both sides and compares with `filepath.Rel`, so a link planted under the root cannot escape, a dangling link fails closed, and a symlinked temp root (macOS `/var`) compares equal to itself. `TestContainedInResolvesSymlinksAndTraversal` covers it. Globs among the defaults (the state glob) are judged on each current match too; `TestRootsOutsideExpandsGlobs` covers that.

## Test failed first

With the four `os.Setenv` redirect lines in `TestMain` commented out:

```
--- FAIL: TestHubDefaultRootsStayInsideTheTestEnvironment (0.00s)
    testmain_test.go:199: os.UserHomeDir = "/Users/jesse", outside the throwaway test root "/var/folders/.../evener-hub-test-env-1624484007"; a handler dispatched with empty params would read or write there for real
    testmain_test.go:199: hubLaunchConfigRoot with LaunchConfigRoot unset = "/Users/jesse/.config/evener", outside the throwaway test root ...
    (plus XDG_CONFIG_HOME/XDG_STATE_HOME/XDG_CACHE_HOME = "", cmdutil.DefaultStateRoot, plugins.DefaultRoot, defaultMCPConfigPath)
```

Redirect restored: PASS.

## Gates

- `go test ./cmd/evener-hub`: `ok` in 71s, 0 FAIL lines
- `make lint`: every stage PASS (naming, gofmt, evenerfuzz, eval, internal, golangci, generated, fuzz-registry, secret-scan)

## Follow-ups

- https://github.com/prime-radiant-inc/evener/issues/982
- https://github.com/prime-radiant-inc/evener/issues/983
- https://github.com/prime-radiant-inc/evener/issues/984

### DIFF
--- a/cmd/evener-hub/app_rpc_test.go
+++ b/cmd/evener-hub/app_rpc_test.go
@@ -11054,9 +11054,12 @@ func TestHubRPCRegistersExpectedHandlerSet(t *testing.T) {
 	// the dispatch loop further down calls every expected method with empty
 	// params, including the credential-mutating evener/auth/* handlers
 	// (apiKey/set, logout, apiKey/clear). A nil CredsStore makes
-	// newHubAuthControllerWithStore fall back to the real on-disk default
-	// (~/.config/evener/credentials.toml via the ambient HOME/XDG env) -
-	// this test must never read or write a developer's actual store.
+	// newHubAuthControllerWithStore fall back to the on-disk default under
+	// the ambient HOME/XDG env. TestMain redirects that env into a throwaway
+	// root and TestHubDefaultRootsStayInsideTheTestEnvironment pins the
+	// redirect, so an unset root cannot reach a developer's actual store;
+	// per-test temp dirs keep this test's writes out of the root the whole
+	// package shares as well.
 	credsStore, loadErr := credentials.LoadStore(filepath.Join(t.TempDir(), "credentials.toml"))
 	if loadErr != nil {
 		t.Fatalf("LoadStore: %v", loadErr)
--- a/cmd/evener-hub/testmain_test.go
+++ b/cmd/evener-hub/testmain_test.go
@@ -176,19 +176,19 @@ func TestGoSubprocessesCacheOutsideTheTestRoot(t *testing.T) {
 	}
 }
 
-// defaultRootsOutsideTestEnv lists every filesystem root the hub falls back to
-// when a WebConfig field is left unset (the launch config root, the hub state
-// root, the plugin store root, the MCP config path), every path runMain opens
-// when a flag is unset (the config file, the state glob, the past-index DB,
-// the rendezvous run dir), plus the HOME and XDG bases they derive from, that
-// resolves outside testEnvRoot. Empty means the throwaway env contains them
-// all.
-func defaultRootsOutsideTestEnv() []string {
+type namedPath struct{ name, path string }
+
+// hubDefaultRoots lists every filesystem root the hub falls back to when a
+// WebConfig field is left unset (the launch config root, the hub state root,
+// the plugin store root, the MCP config path), every path runMain opens when a
+// flag is unset (the config file, the state glob, the past-index DB, the
+// rendezvous run dir), plus the HOME and XDG bases they derive from.
+func hubDefaultRoots() []namedPath {
 	home, err := os.UserHomeDir()
 	if err != nil {
 		home = "(unresolved: " + err.Error() + ")"
 	}
-	roots := []struct{ name, path string }{
+	return []namedPath{
 		{"os.UserHomeDir", home},
 		{envvars.XDGConfigHome.Name, envvars.XDGConfigHome.Getenv()},
 		{envvars.XDGStateHome.Name, envvars.XDGStateHome.Getenv()},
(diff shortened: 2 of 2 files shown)


obra:
Both findings addressed in dc78c67.

**Medium (t.Errorf at testmain_test.go:198).** The proposed `t.Fatalf` would not do what the finding wants: `FailNow` stops only the calling test's goroutine, and the tests running beside it continue against whatever roots they resolve. What stops them is checking before `m.Run`. `TestMain` now runs the same root list and exits naming the offending paths, so no test starts when a default root escapes. The guard test keeps the check live mid-run (a test that swaps `configUserHomeDir` or the XDG env without restoring it) and now uses `Fatalf`, matching `TestGoSubprocessesCacheOutsideTheTestRoot`.

**Low (USERPROFILE).** Nothing here runs tests on Windows (ci.yml only vets under `GOOS=windows`), so before this PR every hub test would already have used the real home there, unreported. The redirect now sets `USERPROFILE` beside `HOME`, so what `os.UserHomeDir` reads is contained on every platform Go supports. Where it still escaped, the pre-run abort is the visible result rather than one failing test.

obra:
Addressed in e3c8b53. `containedIn` resolves both the root and each default through their deepest existing ancestor (`filepath.EvalSymlinks`, rejoining the part nothing has created yet) and compares with `filepath.Rel`, so a link planted under the root cannot point a default outside it and `..` cannot escape.

The same change removes a false-positive trap in the pre-run abort that the lexical check had: a macOS temp root lives under `/var`, which is `/private/var`, so any resolver that canonicalized its output would have failed the prefix check and stopped the whole package.

`TestContainedInResolvesSymlinksAndTraversal` covers the symlinked root (both directions), the planted link, dot-dot traversal, a descendant nothing has created yet, the root itself, a sibling sharing the root as a prefix, and relative and empty paths.

obra:
**Medium (dangling symlink).** Addressed in 3155528. `canonicalizeExisting` now fails closed on a symlink component it cannot resolve, so a dangling link planted under the root is not contained. `TestContainedInResolvesSymlinksAndTraversal` gained the dangling case, which returned `true, want false` before the change.

**Low (os.Symlink on Windows).** Declining. No test in this package runs on Windows: ci.yml only vets under `GOOS=windows`, and .goreleaser.yml builds linux and darwin. Three existing tests here already call `os.Symlink` with `t.Fatal` on error (app_git_head_test.go:57, doc_serve_test.go:255, project_delete_test.go:364), and this test matches them. A skip would also hide a genuine symlink failure on the platforms these tests do run on.

obra:
Addressed in 4e29711. The root list now includes what `runMain` opens when a flag is unset: `DefaultHubStateRoot`, `DefaultConfigPath`, `DefaultStateGlob`, `DefaultPastIndexDBPath` and `rendezvous.DefaultDir`. With the redirect disabled, `TestMain`'s abort lists each of them by name.

The branch also merged main, since branch protection requires it to be up to date.

obra:
Addressed in e203cba. `rootsOutside` now expands any glob among the defaults with `filepath.Glob` and judges each match as well as the pattern, so a symlinked project directory under the state root that points outside is reported. `TestRootsOutsideExpandsGlobs` plants such a link under a private root and asserts it is the one entry reported, then removes it and asserts nothing is.

Merged main again for branch protection.

PR #985, opened 2026-09-08
fix(execenv): bound the work and memory a glob or grep walk can spend

### LINKED ISSUE #497: A single glob rooted at / OOM-kills the agent: the directory-listing budget is gated on filesystems that report no file identity, so it never applies on Linux

### LINKED ISSUE #938: grep's tree walk is still unbudgeted; only its ignore discovery shares the glob budget

obra:
Fixes the OOM in #497, as three reviewed commits kept intact so the layers stay legible in history. It supersedes #951, #955 and #956, which were the same work split for review; roborev judges each slice as if it were landing alone, and no slice could be clean until the last one landed.

A `find_files` call with a `**` pattern rooted at `/` could exhaust memory and OOM-kill the agent process. The walk *terminated* — the ancestor/SameFile cycle check added for #369 catches `/proc/<pid>/root` re-entering the tree — but nothing bounded what it spent getting there.

## What was broken, layer by layer

**1. The budgets never bound anything.** The only directory-listing counter incremented inside `if !hasFileIdentity(info)`. Every file on ext4, APFS and HFS+ carries `(device, inode)` identity, so on the filesystems agents actually run on that branch was never taken. Matches accumulated with no cap at all — a `**` over `/` is millions of paths. And the walk retained one `fs.FileInfo` per directory it had ever listed, though the cycle check only consults *ancestors* of the directory being admitted.

**2. A single walk could still materialize unbounded memory.** Even bounded by listings and matches, every listing read its whole directory before any bound was consulted: `os.DirFS`'s `ReadDir` is `os.ReadDir`, and the sandboxed arm ended in one `df.ReadDir(-1)`. One directory with millions of entries exhausted memory before anything ran. A per-directory cap alone would not have been enough either, because `fs.WalkDir` and doublestar both hold an ancestor's entry slice alive while reading a child, so peak is the cap times depth.

**3. Grep's discovery and walk were unbounded.** Both tools collect `.gitignore` rules by walking the base *before* their own traversal, so the default `include_ignored: false` traversed everything before any bound applied. Grep's own walk (`grepWalk`, `secureBrowseWalkDir`) had no bound at all.

## What changed

A `GlobBudget`, created once per call and shared across ignore discovery and every brace-expanded pattern, bounds four things: directory listings (1,000,000), matches (10,000), entries materialized by one listing (200,000), and entries held live at once (500,000, about 52 MB at ~105 bytes per entry). Listings are pulled in chunks and charged **as they accumulate**, so a pathological directory stops partway rather than after it has been read whole, and the ceiling is consulted during the read rather than one directory past it. Release of live entries is free: to be listing a directory at all, the walk must have left every non-ancestor, which is the rule the ancestor chain already uses.

`listed` became an ancestor chain pruned to the path being walked, so retention is O(depth). The chain is only a cache — an ancestor it does not hold is re-stat'ed, so a walk that jumps between subtrees still cannot miss a cycle.

Ignore discovery is scoped to what the traversal after it can reach: for each pattern, the `.gitignore` at every level from the base down to the pattern's literal prefix (one read per level, no listing), then that prefix's subtree, cut at the depth the pattern can actually match — only `**` crosses a separator, so `*.go` never leaves the base. Its refusals propagate, of every kind, because an ignoreSet that gave up partway reports itself complete and silently under-excludes.

Refusals are a typed error carrying the operation, bound, count, offending directory and whether a cycle could have been detected, so callers read fields rather than parse sentences, and the advice names the lever that actually helps.

## Caller-visible changes

- **Grep can now fail where it used to return partial output**, on both arms.
- **A single directory over 200,000 entries fails glob and grep outright**, naming that directory, rather than being read into memory.
- **Ignore discovery can fail a call**, which affects glob and grep alike whenever `include_ignored` is false.
- **Sandboxed listings are lexically sorted**, which changes sandboxed match order and grep visit order; `os.DirFS` already sorted, so the two arms now agree.
- **A walk holding more than 500,000 entries live at once fails** — reachable by a deep tree of large directories even when no single directory is over the per-directory cap. A wide, flat repository is unaffected: the bound is on entries held *simultaneously*, not a cumulative count.
- `GlobExcluder`'s signature is unchanged. Truncation rides on a caller-supplied budget through the new optional `GlobBudgeter`; without one there is nowhere to report truncation, so that path refuses rather than returning a short list with a nil error.

## How it is tested

Every bound was written test-first and confirmed failing for its stated reason. Tests assert **work done** — entries actually read, listings actually made, `peakDirEntries`, `peakLiveEntries`, entries held at the moment of refusal — because a fix that finishes the walk and then reports a refusal satisfies a result-only assertion while still doing the thing that OOMs. Sample first failures:

- `walk retained 31 directory identities after listing 30 sibling directories`
- `refused holding 25 entries, want no more than the ceiling of 20 plus one chunk of 2`
- `Glob("secret.txt") = ([.../secret.txt], nil), want a refusal: ... the rules were never loaded and the path came back anyway`

Controls keep the bounds honest: a wide, flat tree holding the same total entries must still succeed, or a cumulative counter would break every large repository; and a `**` pattern must still descend, or scoping would collapse into "discovery never descends". Guards that no static tree can reach — the concurrent-growth window between discovery and the walk — are forced by probes that grow a directory inside a stubbed walk. Termination stays pinned by the #369 regression tests.

Gates, all foreground with a private build cache, zero failures in every log: gofmt; `go vet ./...`, `GOOS=windows make vet` and `-tags evenerfuzz`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test -race` on `./agent/execenv/` and `./agent/`; fuzz seed replay on both packages; and the full `go test ./agent/`.

Closes #497
Closes #938


obra:
Both fixed at `5120a5205`; three commits still intact.

**1. Live set not reset between discovery walks.** Each prefix in a multi-prefix scope is its own traversal, so each now starts holding nothing, while the cumulative counters stay call-wide.

Worth recording: **my first test for this was vacuous and I caught it by mutation.** I used two sibling prefixes, but `holdEntries` already releases siblings — neither is an ancestor of the other — so removing the reset changed nothing. The reset only matters when the finished walk's holdings *are* ancestors of where the next one starts. The test now uses a base scope and a nested one, and removing the reset fails it with `holding 25 directory entries live ... past the call-wide budget of 20`.

**2. Unbounded rules reads.** Two bounds, each with its own kind so a model can tell them apart:
- **Per file, 1 MiB.** `readIgnoreFile` stops one byte past the cap via `io.LimitReader`, so an enormous file is refused *without* being materialized. A few kilobytes is a typical real `.gitignore`, so this only ever refuses a pathological one.
- **Per call, 32 MiB retained.** Compiled matchers live as long as the call, so the charge is cumulative. Same arithmetic as the entry caps: one listing at `maxGlobDirEntries` is roughly 21 MB and the live ceiling allows about 52 MB, so 32 MiB sits between them; a repository carrying a `.gitignore` in each of several thousand directories retains single-digit megabytes.

Both tests assert what was read or retained rather than the returned result, and both mutations bite: reading the file whole fails the first with `read 65536 bytes before refusing, want no more than the cap of 512`, and dropping the retention charge fails the second.

One lint fix on the way through: a test constant named `cap` shadowed the builtin (`revive`).

**Full gate set re-run on this branch, zero `^--- FAIL:` in every log:** gofmt; `go vet ./...`, `GOOS=windows make vet` and `-tags evenerfuzz`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `-race ./agent/execenv/`; `-race ./agent/`; fuzz seed replay on both packages; full `go test ./agent/`.

obra:
Both fixed at `e2f99cbff`.

**1. Recursion detected by substring.** `ignoreScopeDepth` scanned the raw pattern for `**`, so anything containing those two bytes widened the scope to unbounded. I checked doublestar's own rule rather than inventing one: its matcher special-cases `**`, `/**`, `**/` and `/**/` and nothing else, so `**` is recursive only as a whole path component. `ignorePatternIsRecursive` now splits on separators and compares each component's raw text to `**`, which follows that rule exactly and also handles escapes and classes — neither `\**` nor `[**]` has a component whose raw text is `**`. Splitting on every slash cannot hide a real `**`, since a standalone one is always bounded by separators or the pattern's ends; the only misreading available is an escaped separator, and that errs toward a wider scope rather than a missing rule.

**One correction to the finding.** Of the three shapes named, `foo**bar.txt` and `sub/[**].txt` genuinely over-widened, but `sub/\*\*.txt` did not: the backslash sits between the two stars, so the old substring scan never matched it. I verified that by mutation rather than assuming. The shape that *does* break the old rule is `sub/\**.txt` — an escaped star adjacent to a wildcard puts two stars together — so the test uses that. All three subtests now fail under the substring scan.

**2. Bounded scopes did not subsume.** `narrowIgnoreScopes` only let an unbounded scope absorb a nested one, so `.` at depth 2 still walked `a` at depth 0 a second time against the shared budget. It now subsumes when the ancestor's remaining reach covers the nested scope — the levels down to it plus its own reach fitting inside the ancestor's. `TestNarrowIgnoreScopesSubsumesANestedScopeWithinReach` pins both directions, including the one-level-short case where both scopes must survive; reverting to unbounded-only fails it.

**Full gate set on this branch, zero `^--- FAIL:` in every log:** gofmt; `go vet ./...`, `GOOS=windows make vet` and `-tags evenerfuzz`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `-race ./agent/execenv/`; `-race ./agent/`; fuzz seed replay on both packages; full `go test ./agent/`.

obra:
Rebase verified on `08e995782`. I did **not** push: Jesse had already rebased this branch onto `be18ad663` and pushed it, so my `--force-with-lease` was correctly rejected as stale. I checked rather than forcing, and the two rebases are content-equivalent — all three commits have byte-identical trees pairwise, and `git diff` between the two tips is empty. I reset to the remote and left it alone.

Because the trees are identical, the full gate set I ran applies to this tip unchanged. Zero `^--- FAIL:` in every log:

| gate | result |
| --- | --- |
| `gofmt -l` | clean |
| `go build ./...` | clean |
| `go vet ./...` | 0 |
| `GOOS=windows make vet` | 0 |
| `go vet -tags evenerfuzz ./...` | 0 |
| `make lint-golangci` | PASS |
| `make lint-evenerfuzz lint-eval lint-gofmt lint-internal` | PASS |
| `go test -race ./agent/execenv/` | 0 FAIL |
| `go test -race ./agent/` | 0 FAIL |
| `-tags evenerfuzz -run Fuzz ./agent/execenv/` | 0 FAIL |
| `-tags evenerfuzz -run Fuzz ./agent/` | 0 FAIL |
| full `go test ./agent/` | 0 FAIL |

**On #1008's two changes, confirmed rather than assumed.** `streamDeltaChunk` is gone from main and is referenced by none of the files this PR touches. And nothing here touches `appwire` or imports it from `agent/execenv`, so the `doc.go` contract statement does not reach this work. More generally, the set of files main changed since the merge-base and the set this PR changes do not intersect at all.

Branch is `MERGEABLE`, currently `BEHIND` by two commits that touch only `agent/session_active_turn_test.go` — unrelated, and no conflict. Say the word if you want it brought up to `e8db3ca83` before merging and I will rebase and re-gate.

obra:
Head is `167969543`, rebased and pushed, and all gates green.

**Flagging a race you will want to decide on, rather than looping on it.** `main` is landing commits faster than a full re-gate takes. Over the last stretch it moved from `e8db3ca83` to `aa8dd8e1a` to `2690c95ad` to `c29323ae5`, so each time I finished the ~17-minute `-race ./agent/` lane the branch was already behind again. I rebased three times to chase it and the branch is now behind by 3 once more.

None of it is a conflict. Every commit main gained during this touched only files this PR does not: `agent/internal/jobstore/watch.go`, `agent/session_active_turn_test.go`, `agent/apilog_session_attribution_test.go`, and the `cmd/evener-hub` test harness. The rebases were mechanically clean and I verified each one left this PR's diff byte-identical.

Given the up-to-date requirement with auto-merge off, the branch can only be up to date at the moment of merge, so this needs either a merge promptly after a rebase or auto-merge switched on. I have stopped rebasing so the head stays stable for you: **say the word and I will rebase once more and hand it straight to you**, or enable auto-merge and it will settle itself.

**Gates on `167969543`, zero `^--- FAIL:` in every log:**

| gate | result |
| --- | --- |
| `go build ./...` | clean |
| `go test ./agent/execenv/` | 0 FAIL |
| `go test -race ./agent/execenv/` | 0 FAIL |
| `go test ./agent/` | 0 FAIL |

The `-race ./agent/` lane last ran on `b665fc93d`, 0 FAIL. That tree is byte-identical to this one — I checked the diffs match exactly — and the three commits between them touch only `agent/apilog_session_attribution_test.go`, so I did not spend another 17 minutes re-running it. Say if you would rather I did.

Per your note I left `refs/heads/origin/main` alone.

obra:
All four addressed at `46988d4a1`. No rebase; only the tip commit changed.

**1. Literal directory patterns no longer list anything.** A pattern with no glob metacharacter names one exact path, which the glob resolves with a stat, so `ignoreScope` now carries a `walk` flag and discovery reads the prefix's own `.gitignore` directly instead of listing it. Escaped metacharacters count as literals. Pinned by `TestGlobIgnoreDiscoveryDoesNotListALiteralDirectoryPattern`, which also asserts the second half: the prefix's rules still exclude, so the test cannot be satisfied by loading nothing. Forcing `walk: true` fails it.

**2. Overlapping scopes read each rules file once.** I took the dedupe option: `loadedRules` keys every `.gitignore` path taken across the whole call, so a partial overlap cannot compile, retain and charge a second matcher. `TestLoadIgnoreSetReadsEachRulesFileOnce` uses base-at-depth-1 plus nested-at-depth-1, which overlap without either covering the other; dropping the dedupe fails it with `rules from "." retained 2 times`.

**3. Empty rules files are no longer free.** `retainRuleBytes` charges a fixed 512-byte per-file overhead for the entry, the compiled matcher and slice growth, which makes the 32 MiB budget bound the file count too — about 65,000 files, so a `.gitignore` in every directory can no longer reach the directory-listing limit with the byte budget untouched. Charging source bytes only fails `TestRetainRuleBytesChargesPerFileOverhead`.

**4. The dot-directory ancestor read: not covered, so fixed here.** My masked-prefix work is a different concern — masking is per path and set by policy; this is the dotfile convention the match filter uses. The ancestor loop now applies the same dot-directory skip the subtree walk applies. The old behaviour was harmless in outcome, since `isDotPath` drops every candidate below such a directory before its rules could apply, but it spent the rules budget for nothing, which matters more now that budget is charged. Pinned by `TestLoadIgnoreSetSkipsADotDirectoryAncestor`.

**Three existing tests had to be retargeted, and I want that on the record.** `TestGlobIgnoreDiscoveryRefusesRatherThanUnderExcluding` and `...StopsOnADirectoryWithTooManyEntries` both used literal patterns precisely to isolate discovery from the pattern walk. Item 1 removes that asymmetry by design, so through a glob call the two passes now cover the same ground and neither test could tell which one refused — they would have passed with the refusal swallowed. Both now drive `loadIgnoreSet` directly, which is the only boundary that can distinguish them, and both were mutation-verified there. `TestNarrowIgnoreScopesSubsumesANestedScopeWithinReach` needed `walk: true` on its scopes now the field exists.

**Gates, zero `^--- FAIL:` in every log:** gofmt; `go vet ./...`, `GOOS=windows make vet` and `-tags evenerfuzz`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test ./agent/execenv/`; `-race ./agent/execenv/`; fuzz seed replay on both packages; full `go test ./agent/`; `-race ./agent/`.

Head is stable and I am not rebasing. `BEHIND` again, as expected.

obra:
All three at `665bea957`. No rebase; tip commit only.

**1. The three silent tests now exercise the walk, and both mutations bite again.** Good catch, and it was my `walk` field that made the silence possible: adding a field with a meaningful zero value turned three existing literals into no-op scopes. Set `walk: true` at all three sites and re-ran the mutations you named:

- deleting `resetLive()` → `TestLoadIgnoreSetStartsEachScopeWalkWithACleanLiveSet` fails with `holding 25 directory entries live ... past the call-wide budget of 20`
- exempting the walk root from `skip` (`p != sc.prefix`) → `TestLoadIgnoreSetConsultsSkipForTheScopePrefixItself` fails with `skip was never consulted about the prefix itself; it was asked about [.gitignore vault/.gitignore vault/keep.txt]`

Both were passing under those mutations at this head before the change, so the pins are live again.

**2. Memory claim softened.** You are right and my comment overclaimed. `maxGlobIgnoreTotalBytes` now says what it actually bounds: how much `.gitignore` source a call reads and how many files it keeps, explicitly **not** the memory those files occupy compiled, since go-gitignore builds a regexp per rule line at a cost far larger than the line itself, so retained memory can exceed the figure by orders of magnitude. The overhead constant's comment likewise now presents 512 as an accounting figure for bounding the file count, not a measurement of what a retained file occupies. I did not touch the accounting.

**Please file the follow-up:** the aggregate ignore budget charges source bytes, so it does not bound compiled matcher memory. At roughly 2 KB per compiled rule line, a 149-byte file retains about 39 KB and a file at the 1 MiB per-file cap retains on the order of a gigabyte. Bounding it means charging per compiled rule rather than per byte. Pre-existing and out of scope here.

**3. Contradicting comment deleted.** The old note explaining why the ancestor loop had *no* dot-directory check sat directly under the check I added.

**Gates, zero `^--- FAIL:` in every log:** gofmt; `go vet ./...`, `GOOS=windows make vet` and `-tags evenerfuzz`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test ./agent/execenv/`; `-race ./agent/execenv/`; fuzz seed replay on both packages; full `go test ./agent/`; `-race ./agent/`.

Holding here, no rebase. `BEHIND` as before.

obra:
Head is `f357d3e29`. Two fixes, and the Medium **does not reproduce** — evidence below rather than an assertion.

**1. Medium: not a defect. `SplitPattern` is escape-aware and unescapes the base.** The finding assumes the prefix stops before an escaped metacharacter. It does not: `SplitPattern`'s scan skips a backslash-escaped byte, so it walks past `\*` and splits at the next separator, then returns `unescapeMeta` of the prefix. Measured directly:

| pattern | base | rest | walk | ancestors |
| --- | --- | --- | --- | --- |
| `a/\*/target.txt` | `a/*` | `target.txt` | false | `[. a]` |
| `a/\*/b/target.txt` | `a/*/b` | `target.txt` | false | `[. a a/*]` |
| `a/\[x\]/t.txt` | `a/[x]` | `t.txt` | false | `[. a]` |

So the prefix already names the real directory, and the literal path appends the prefix itself to the ancestor reads, which is exactly `a/*/.gitignore`.

I tested the behaviour rather than the internals, as you asked. `TestGlobAppliesRulesInsideAnEscapedMetacharacterDirectory` builds a directory literally named `*`, puts a `.gitignore` inside it excluding `secret.txt`, and asserts `a/\*/secret.txt` returns nothing while `a/\*/target.txt` still returns its file, so it cannot pass by matching nothing. **It passes at this head, unchanged.** It is not vacuous: dropping the literal-prefix ancestor read fails it with `Glob("a/\*/secret.txt") = [.../a/*/secret.txt], want nothing`. Kept as a regression guard, and no code change was needed — say if you would rather I still added the conservative bounded walk.

**2. Low: one shared budget in the masked-subtree test.** It handed `secureDirFS` and `loadIgnoreSet` separate budgets, so entry and live charges were split from listing and rules charges, unlike every production caller. Both now take the same pointer, with a comment saying why.

**3. Low: `TruncatedAt()` records the cap that tripped.** `match()` stores the cap alongside `truncated`, and the accessor returns the stored value. Pinned by `TestTruncatedAtReportsTheCapThatTripped`, which lowers the cap, trips it, restores the global, then reads; reverting to the global fails it with `TruncatedAt() = 10000, want 3`.

**Gates, zero `^--- FAIL:` in every log:** gofmt; `go vet ./...`, `GOOS=windows make vet` and `-tags evenerfuzz`; `make lint-golangci`; `make lint-evenerfuzz lint-eval lint-gofmt lint-internal`; `go test ./agent/execenv/`; `-race ./agent/execenv/`; fuzz seed replay on both packages; full `go test ./agent/`; `-race ./agent/`.

Holding, no rebase.

obra:
On the Medium about escaped glob metacharacters (gitignore.go:219-224): declined, because its premise does not hold, and there is now a regression test standing where the defect would have been.

The finding assumes `doublestar.SplitPattern` stops the prefix before an escaped metacharacter. It does not: its scan skips backslash-escaped bytes, walks past `\*`, splits at the next separator, and returns `unescapeMeta` of the prefix. Measured on this head:

| pattern | base | rest | ancestors read |
| --- | --- | --- | --- |
| `a/\*/target.txt` | `a/*` | `target.txt` | `.`, `a` |
| `a/\*/b/target.txt` | `a/*/b` | `target.txt` | `.`, `a`, `a/*` |
| `a/\[x\]/t.txt` | `a/[x]` | `t.txt` | `.`, `a` |

So the prefix already names the real directory, and the literal path appends the prefix itself to the ancestor reads, which is exactly `a/*/.gitignore`. Nothing is missed.

Verified behaviourally rather than by reading the code: with a directory literally named `*` holding a `.gitignore` that excludes `secret.txt`, the pattern `a/\*/secret.txt` returns nothing, while `a/\*/target.txt` still returns its file, so the first result cannot be explained by the pattern matching nothing. That test passes at this head unchanged and is not vacuous — removing the literal-prefix ancestor read makes it fail. It stays in the suite as a guard.

The two Low findings were both real and are fixed: the masked-subtree test now passes one budget pointer to `secureDirFS` and `loadIgnoreSet` instead of two, matching production and letting the test catch a budget-sharing regression; and `TruncatedAt()` now returns the cap recorded when truncation tripped rather than reading the current global, pinned by a test that fails with the old behaviour.

obra:
Both Mediums from the `b193546` combined review addressed at `147d3ea76`. No rebase; tip commit only.

**1. Nil `budget` deref (local.go:1723): fixed.** `GlobWithBudget` now rejects a nil `budget` with an error before any charge. Reproduced first: the new test panicked (`invalid memory address or nil pointer dereference`) without the guard. Pinned by `TestGlobWithBudgetRefusesANilBudget`, which fails (panics) with the guard removed. Rejected the review's second option (mint an internal budget) deliberately: silently substituting a budget the caller never sees would discard truncation the caller asked to observe via `budget.TruncatedAt()`, repeating the silent-truncation failure mode `GlobWithExclusions` already documents against.

**2. Swallowed cancellation in ignore discovery (gitignore.go:360-377): fixed.** `loadIgnoreSet` and `readIgnoreFile` now take `ctx`; the WalkDir error path checks `ctx.Err()` before the budget-refusal check (so a walk that is both cancelled and over budget reports the cancellation), the ancestor reads check `ctx.Err()` per directory, and `readIgnoreFile` reports `ctx.Err()` on open/read failure instead of returning `nil, nil`. Reproduced first: `loadIgnoreSet` over an already-cancelled context returned an empty set with nil error. Pinned by `TestLoadIgnoreSetReportsCancellation` and `TestLoadIgnoreSetStopsMidWalkCancels` (promptness: stops within ~1 listing of the cancellation); removing the WalkDir propagation fails the mid-walk test. Matches the existing `globMatchIsDir` precedent. Callers updated: off-sandbox glob + grepNative (`local.go`), sandboxed glob + grepNative (`securepath_browse_fdops_unix.go`); test call sites pass `t.Context()`.

**Gates, zero failures:** `go test ./agent/execenv/`; `-race ./agent/execenv/`; full `go test ./agent/`; `GOOS=windows go vet`; `go vet -tags evenerfuzz`; `golangci-lint run agent/execenv/...` (0 issues); execenv fuzz seed replay.

Holding here, no rebase.

obra:
Medium from the `a96ed58` combined review addressed at `347f9c5d3`. No rebase; tip commit only.

**Dot-directory scope (gitignore.go:320-321, 411-419): fixed, and the finding was exactly right.** For a scope like `a/.config/sub` the ancestor list is `[. a a/.config]`: the basename check skipped `a/.config` itself but nothing stopped the walk rooted at `a/.config/sub`, and the walk's own `d.Name()` check only skips dot-named entries — never a scope already rooted under one. Reproduced first: discovery retained rules from `a/.config/sub` and spent listings there, even though `isDotPath` drops every candidate beneath `.config` before such a rule could apply. Two changes: the ancestor loop now checks the full path with `isDotPath(dir)` instead of the basename, and a scope with `isDotPath(sc.prefix)` is not walked at all.

Scope is deliberately discovery-only: the glob's own match path is untouched, so the end-to-end answer is unchanged (the pattern still returns nothing under the default exclusion via `isDotPath`, and the full set with `includeIgnored`). What changes is only wasted budget: previously a large enough subtree under such a scope could trip the listing budget and fail a glob whose answer was fixed before it started.

Pinned by `TestLoadIgnoreSetSkipsAScopeBeneathADotDirectory`, which fails without the fix (`collected rules from "a/.config/sub"`); the pre-existing `TestLoadIgnoreSetSkipsADotDirectoryAncestor` and the escaped-metacharacter regression guard still pass.

**Gates:** `go test ./agent/execenv/`; `-race ./agent/execenv/`; gofmt clean; `go vet`, `GOOS=windows go vet`, `-tags evenerfuzz vet`; `golangci-lint run agent/execenv/...` (0 issues).

obra:
Both findings from the `cdd6d21` combined review addressed at `827566bb4`. No rebase; tip commit only.

**Shared premise, verified rather than assumed.** `globpattern.Expand("{**}")` returns `["{**}"]` unchanged, while doublestar matches `{**}` recursively (`Match("{**}","a/b/c")=true`, and `{a}` matches `a` but not `{a}` — pure grouping). Both fixes below ride on that: unwrapping `{X}` to `X` is semantics-preserving.

**1. High (search_patterns.go:756): confirmed and quantified.** `globAltsWalk` buffers every alternative match and only then invokes our callback, so the per-match `budget.match()` cap could not bind while walking. Measured with a ReadDir-counting FS over a 200-directory tree at cap 5: `{**}/f.txt` cost **414 listings** vs **23** for the equivalent `**/f.txt` — ~18x the work, scaling with the tree rather than the cap. The result count was capped pre-fix too (5 and 5), which is why the pin measures listings, not results. Fix: `Expand` now unwraps single-term groups (`{**}`→`{**}`→`**`, recursively for nesting), so the walk takes the plain `**` shape. Post-fix: 27 listings vs 23. Escaped braces and character-class contents never unwrap, matching the existing scan rules.

**2. Medium (gitignore.go:94): fixed at the same choke point.** `ignorePatternIsRecursive` compared each slash component to literal `**`, so `{**}` read as non-recursive and discovery scoped narrow while the walk went recursive. Since `Expand` now normalizes before scope computation, the check sees the same `**` doublestar walks — no change to the check itself. Verified: `{**}/*.txt`, `{foo/**}/*.txt` and `prefix/{**}/sfx/*.txt` all produce unbounded scopes; the scope test fails without the fix (`bounded depth 1`).

**Tests:** `TestGlobCapsASingleTermBraceWalkWhileWalking` (2x listings slack: post-fix ratio 27/23, pre-fix 414/23), `TestIgnoreScopeTreatsASingleTermBraceStarStarAsRecursive`, plus `Expand` table cases for `{**}`, `{foo/**}`, `{a}` and `{a,{b}}`.

**Gates:** `go test ./agent/internal/globpattern/` + `./agent/execenv/`; `-race` both; gofmt clean; `go vet`, `GOOS=windows vet`, `-tags evenerfuzz vet`; `golangci-lint` 0 issues (one `rangeint` in my own test fixed to `for range 200` per repo idiom).

obra:
Medium from the `349a970` combined review addressed at `b4783adec`. One main-merge in the same push (hub test roots, no PR-file overlap); tip fix commit only otherwise.

**Mid-directory cancellation (gitignore.go:385): reproduced exactly, then fixed.** Cancelling right after directory `d`'s 200-entry read succeeded — so every remaining entry arrived with a nil error and no file was ever opened again — returned `dirs=1 err=<nil>`: the walk processed all 200 entries and reported a partial set with no error. Two changes: the callback now checks `ctx.Err()` first on **every** invocation (not just the error path), matching the `globMatches` per-match precedent; and the walk's own return is captured instead of discarded (`_ =` → `walkErr`), promoting a return-without-ctx to a cancellation when the context is done and leaving anything else best-effort. The error-path ctx check this subsumes was removed rather than left to rot beside its replacement.

Pinned by `TestLoadIgnoreSetStopsBetweenEntries`, which fails without the fix with the exact reported shape.

**Gates:** `go test ./agent/execenv/`; `-race ./agent/execenv/`; gofmt clean; `go vet`, `GOOS=windows vet`, `-tags evenerfuzz vet`; `golangci-lint run agent/execenv/...` 0 issues.

obra:
Review-clean at `b4783adec`: roborev's `b4783ad` combined review reports **no issues found**, and all 14 checks are green. Finding history on my watch, each with a failing-test-first pin:

1. Nil `*GlobBudget` panic → rejected with error (`147d3ea76`, `TestGlobWithBudgetRefusesANilBudget`)
2. Swallowed cancellation in ignore discovery → ctx threaded through ancestor reads + WalkDir (`147d3ea76`, `TestLoadIgnoreSetReportsCancellation`, `TestLoadIgnoreSetStopsMidWalkCancels`)
3. Dot-directory mid-path scoping (`a/.config/sub`) → full-path `isDotPath` checks (`347f9c5d3`, `TestLoadIgnoreSetSkipsAScopeBeneathADotDirectory`)
4. Single-term brace cap evasion + scope gap (`{**}`) → unwrap in `Expand` (`827566bb4`, listings-ratio pin + scope test + Expand table cases; 414→27 listings)
5. Mid-directory cancellation promptness → per-invocation ctx check + captured walk return (`b4783adec`, `TestLoadIgnoreSetStopsBetweenEntries`)

All prior rounds' fixes remain intact on the branch. The only remaining step is an approving review — I cannot approve my own shepherding — after which this is ready to merge.

PR #988, opened 2026-09-08
test(hub): drop the redundant per-method dispatch loop from the handler-set test

### LINKED ISSUE #982: TestHubRPCRegistersExpectedHandlerSet: the per-method dispatch loop is redundant with the router-set check and runs real handler writes
`TestHubRPCRegistersExpectedHandlerSet` (cmd/evener-hub/app_rpc_test.go) asserts `web.appRPC.Router().Methods()` equals a named list, then dispatches every named method over the wire with empty params and only checks that none answers methodNotFound. `/rpc` serves the same `appRPC` whose router the set check inspected, so per-method dispatch adds no reachability the set check lacks; one registered dispatch plus the existing unregistered sanity dispatch would prove the wire path lands on that router.

Cost of the loop: any handler for which an empty request is a valid write runs it for real against whichever root the fixture left unset (`evener/launch/setLayer` today; `evener/settings/agentsDoc/set` in #979). #981 pins that those roots resolve inside TestMain's throwaway env, so the loop is contained rather than harmless: it still mutates the root the whole package shares, and it needs the `notDispatched` allow-list for handlers that act outside the process (self-upgrade, marketplace, plugins).

Options:
1. Drop the loop. Keep one registered dispatch and the unregistered sanity dispatch to prove the wire path.
2. Keep it and accept the allow-list plus the sandbox as the guard.

Not done in #981: it rewrites a test whose design was deliberate ("Every named method is then dispatched over the wire") and drops incidental handler coverage. Owner's call.

obra:
Rebased onto main after #981 merged.

## Mechanism

`TestHubRPCRegistersExpectedHandlerSet` asserted `web.appRPC.Router().Methods()` equals a hand-named list, then dispatched every named method over the wire with `appwire.EmptyParams{}` and checked only that none answered `methodNotFound`. `/rpc` serves that same `appRPC` (cmd/evener-hub/web.go), so the per-method loop added no reachability the set check already covers. What it did add: for any handler where an empty request is a valid write (`evener/launch/setLayer` today, `evener/settings/agentsDoc/set` in #979) the write ran for real against whichever root the fixture left unset, and it needed a `notDispatched` allow-list of 14 methods whose handlers act outside the process (self-upgrade, marketplace, plugins).

## Change

Option 1 from #982. Removed the loop and the `notDispatched` map. Kept the `Methods()` equality check unchanged, kept the existing unregistered-method sanity dispatch, and added exactly one dispatch of a registered read-only method — `model/list`, which answers from the offline no-user-layer registry the test already configures, spawns nothing and writes nothing. Rewrote the test's doc comment and the credentials-store fixture comment to describe what the test now does; the temp-dir fixtures stay, they still keep this test's state out of the root the package shares.

## Test that failed first

Pointed the new registered dispatch at `"evener/__not_registered_on_purpose__"` and ran it:

```
--- FAIL: TestHubRPCRegistersExpectedHandlerSet (0.09s)
    app_rpc_test.go:11192: method "evener/__not_registered_on_purpose__" is not registered (methodNotFound)
```

Restored to `appwire.MethodModelList` and the test passes.

## Smoke gates

- `gofmt -l cmd/evener-hub/app_rpc_test.go` — no output
- `go vet ./cmd/evener-hub` — clean
- `go vet -tags evenerfuzz ./cmd/evener-hub` — clean
- `GOMAXPROCS=4 go test -count=1 -run '^TestHubRPCRegistersExpectedHandlerSet$' ./cmd/evener-hub` — PASS (0.09s)

No non-test code changed, so no `go build ./...`.

## Noticed, left alone

`testmain_test.go`'s doc comment on `TestHubDefaultRootsStayInsideTheTestEnvironment` still cites this test's every-method dispatch as the reason the roots have to be pinned. That justification is now stale, though the pin itself is still worth keeping (other tests in the package dispatch write handlers). Out of scope here per the brief.

Closes #982

### DIFF
--- a/cmd/evener-hub/app_rpc_test.go
+++ b/cmd/evener-hub/app_rpc_test.go
@@ -15,6 +15,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -11044,7 +11045,8 @@ func newHubRPCTestServerWithWeb(t *testing.T, cfg hubcore.WebConfig) (*httptest.
 // directions.
 //
 // Two dispatches over the wire tie that set to what /rpc actually serves: a
-// registered read-only method must not answer methodNotFound, and an
+// registered read-only method (model/list, answered by a LiveModels stub so
+// it never asks a live provider) must not answer methodNotFound, and an
 // unregistered name must. Dispatching every named method instead adds no
 // reachability — /rpc serves the same appRPC whose router the set check
 // inspects — while running for real any handler for which empty params are a
@@ -11065,12 +11067,22 @@ func TestHubRPCRegistersExpectedHandlerSet(t *testing.T) {
 	if loadErr != nil {
 		t.Fatalf("LoadStore: %v", loadErr)
 	}
+	var liveModelsCalled atomic.Bool
 	hub, web := newHubRPCTestServerWithWeb(t, hubcore.WebConfig{
 		Past:                hubcore.NewPastIndex(""),
 		Registry:            newTestRegistry(t, t.TempDir(), tomlPath, credsStore, nil),
 		ProvidersConfigPath: tomlPath,
 		HubStateRoot:        dir,
 		CredsStore:          credsStore,
+		// model/list is the one registered method dispatched below. With
+		// LiveModels unset, NewWebServer falls back to fetchLiveModels, which
+		// loads the default client and asks every discovered provider,
+		// including the implicit Ollama endpoint. The stub keeps the probe
+		// offline and records that the dispatch reached it.
+		LiveModels: func(context.Context) []appwire.ModelDescriptor {
+			liveModelsCalled.Store(true)
+			return nil
+		},
 	})
 	defer hub.Close()
 	client := dialHubRPC(t, hub)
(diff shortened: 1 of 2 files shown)


obra:
Rebased onto main now that #981 has landed via squash-merge. The branch's history no longer carries #981's original commits, so the conflict against `main` in `cmd/evener-hub/app_rpc_test.go` and `cmd/evener-hub/testmain_test.go` is gone.

The replay applied cleanly and the diff is unchanged — still exactly one commit touching only `cmd/evener-hub/app_rpc_test.go`. New head: `c14a37c67`.

obra:
Addressed the Low finding on the isolation comment in `cmd/evener-hub/testmain_test.go`. The paragraph no longer claims the handler-set test dispatches every registered method; it now names the single `model/list` round-trip and generalizes the rationale to every hub test that calls a handler. Guard rationale unchanged, no code changes.

New wording:

```go
// Any hub test that dispatches a handler is why this has to be pinned. A
// handler called with empty params reads and writes against whichever root
// the fixture left unset — TestHubRPCRegistersExpectedHandlerSet's single
// model/list round-trip is one such dispatch — and for a handler where an
// empty request is a valid write (evener/launch/setLayer today; any "set
// this file's content" handler tomorrow) the write happens for real. The
// HOME/XDG redirect in TestMain is what keeps that out of ~/.config/evener.
```

Gates: `gofmt -l` clean, `go vet ./cmd/evener-hub` clean, `GOMAXPROCS=4 go test -count=1 -run 'TestHubDefaultRootsStayInsideTheTestEnvironment|TestHubRPCRegistersExpectedHandlerSet' ./cmd/evener-hub` ok.

obra:
Addressed in f79f61442. The fixture now sets a deterministic `LiveModels` stub, so the `model/list` probe never reaches `fetchLiveModels` and its provider discovery. The test also asserts the dispatch reached the stub; with the probe temporarily pointed at `thread/list` that assertion failed with "model/list did not reach the LiveModels stub", so it is live. Merged main for branch protection.

Issue #976, opened 2026-09-07
Flake: TestServeWebSocketUnsubscribeWaitsThroughSubscriptionRegistration leaves a stale subscription under race


obra:
Seen once on PR #951's `race-root` lane, run 34152370448 (2026-09-07), on a branch that touches no file under `internal/appserver`:

```
--- FAIL: TestServeWebSocketUnsubscribeWaitsThroughSubscriptionRegistration (0.01s)
    websocket_dispatch_test.go:652: unsubscribe left a stale registered subscription
FAIL	primeradiant.com/evener/internal/appserver	3.281s
```

Re-running the same job on the same commit passed, and main's `race-root` lane has passed on every run since the test landed in #933 (17284f71d, this morning). So this is a load-sensitive failure rather than a regression, and the 0.01s duration says it lost a race rather than timing out.

Worth a look from whoever owns the unsubscribe fence: the assertion is that unsubscribe leaves no registered subscription behind, so a stale one under `-race` means the registration and the unsubscribe can still interleave in an order the fence does not cover, or the test observes registration state before the unsubscribe's own write is visible.

obra:
Recurred 2026-09-08 on PR #990 head 7776f41 (run 34273600732, job `race-root`), a comment-only push to a Go test file under cmd/evener-hub:

```
--- FAIL: TestServeWebSocketUnsubscribeWaitsThroughSubscriptionRegistration (0.02s)
    websocket_dispatch_test.go:652: unsubscribe left a stale registered subscription
```

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

[...]

PR #990, opened 2026-09-08
test(hub): give each hub fixture server its own state roots

### LINKED ISSUE #983: newHubRPCTestServerWithWeb leaves HubStateRoot/LaunchConfigRoot/PluginRoot unset, so tests share the package-wide throwaway root
`newHubRPCTestServerWithWeb` (cmd/evener-hub/app_rpc_test.go) defaults `Registry` when nil but leaves `HubStateRoot`, `LaunchConfigRoot` and `PluginRoot` empty. Each falls back to a HOME/XDG-derived default, which under this package's TestMain is the single throwaway root every test in the binary shares. Two parallel tests that both leave a root unset can read each other's writes (`launch.toml`, `credentials.toml`, the plugin store), and a new handler's author has to remember to set the field, as #979 did for `LaunchConfigRoot`.

Proposal: default each empty root to `t.TempDir()` in the fixture, so a forgotten field lands in a per-test directory rather than the shared one.

Risk to check first: tests that seed `XDG_STATE_HOME` or `XDG_CONFIG_HOME` themselves via `t.Setenv` and then construct a server with the field unset (app_auth_test.go, app_command_list_test.go, app_plugins_test.go and others) rely on the fallback landing in that env. Those need the explicit field or an opt-out before the default changes.

Not done in #981 to keep that change scoped to the containment guard.

obra:
**Mechanism.** `newHubRPCTestServerWithWeb` (cmd/evener-hub/app_rpc_test.go) defaults `Registry` when nil but leaves `HubStateRoot`, `LaunchConfigRoot` and `PluginRoot` empty. Each falls back to a HOME/XDG-derived default (`cmdutil.DefaultStateRoot()`, `cmdutil.DefaultConfigRoot()`, `plugins.DefaultRoot()`), which under this package's TestMain is the single throwaway root every test in the binary shares. Two parallel tests that both leave a root unset read each other's `launch.toml`, `credentials.toml` and plugin store, and a new handler's author has to remember to set the field (as #979 did for `LaunchConfigRoot`).

**Change.** The fixture defaults each empty root to `t.TempDir()`. No opt-out flag: no test needed one.

**Risk audit.** The issue names tests that seed `XDG_STATE_HOME`/`XDG_CONFIG_HOME` with `t.Setenv` and then build a server with the field unset. Of the files that seed those vars, only app_auth_test.go, app_command_list_test.go and app_dirs_test.go also build a server through the fixture. None reads a fixture root through the fallback: the auth controller ignores its `hubStateRoot` argument entirely (`newHubAuthControllerWithStore(_ string, ...)` in app_auth.go) and resolves the OpenAI state dir straight from the env the test seeded; app_dirs_test.go already passes `HubStateRoot`; app_command_list_test.go's one fixture call seeds no env. **Zero tests needed an explicit root.**

**Test that failed first.** `TestHubRPCTestServerGivesEachTestItsOwnRoots`, on
`HubStateRoot left empty ("", ""); it falls back to the shared package root` (and the same for `LaunchConfigRoot` and `PluginRoot`) before the fixture change.

**Gates.**
- `gofmt -l cmd/evener-hub/app_rpc_test.go` — clean.
- `go vet ./cmd/evener-hub` — clean.
- `go vet -tags evenerfuzz ./cmd/evener-hub` — clean.
- `GOMAXPROCS=4 go test -count=1 ./cmd/evener-hub` — `ok ... 82.675s`, zero FAIL lines. (Whole-package run granted for this issue because the fixture is shared by hundreds of tests.)

Closes #983


obra:
Fixed the credentials finding in 206fa46.

**Mechanism.** The per-fixture roots stopped at `HubStateRoot`/`LaunchConfigRoot`/`PluginRoot`. `cfg.CredsStore` stayed nil, and `newHubAuthControllerWithStore` ignores the state root it is handed (`_ string`, app_auth.go:126), so it fell back to the process-wide XDG-derived `credentials.toml` — the one throwaway root every test in the binary shares. Parallel fixtures could read and overwrite each other's keys through `evener/auth/apiKey/set`, `apiKey/clear` and `logout`.

**Change.** When `cfg.CredsStore` is nil the fixture loads a store at `filepath.Join(t.TempDir(), "credentials.toml")`; when `cfg.CredentialsPath` is empty it is set to that same file. That pairing is what main.go does: `cmdutil.CredentialsPath()` feeds both the store it loads and the path each spawned child gets as `EVENER_CREDENTIALS_CONFIG`.

**Registry ordering — what I decided.** The default registry keeps loading through `cmdutil.LoadRegistry`, unchanged, and is still built before the store. Only one fixture caller sets `CredsStore` (`TestHubRPCRegistersExpectedHandlerSet`), and it already passes its own `newTestRegistry(..., credsStore, ...)` over the same store — which is this package's convention (`newTestAuthController` does the same) for a test that needs the registry to resolve a fixture credential. Rewiring the default registry's credential source would change what all ~118 fixture callers resolve against, and I could not pin it with a hermetic assertion: the registry's env layer can outrank the store and TestMain does not clear provider API-key env vars, so the check would depend on the developer's shell. Left it alone.

**Audit.** `CredsStore` appears in exactly one hub test as an explicit value; nothing passes nil on purpose to reach the env-derived store. The files that seed `XDG_STATE_HOME`/`XDG_CONFIG_HOME` and also build through the fixture (app_auth_test.go, app_command_list_test.go, app_dirs_test.go) read OAuth records out of the seeded state dir, not credentials.toml, so an explicit store changes nothing for them. No opt-out flag added.

**Test that failed first.** `TestHubRPCTestServerGivesEachTestItsOwnCredentials`, on `server 0 has a nil CredsStore; the auth controller falls back to the process-wide store` (and the same for server 1). It also pins that the two stores are at different paths, that neither is `cmdutil.CredentialsPath()`, and that `CredentialsPath` equals the store's own path.

**Gates.**
- `gofmt -l cmd/evener-hub/app_rpc_test.go` — clean.
- `go vet ./cmd/evener-hub` — clean. `go vet -tags evenerfuzz ./cmd/evener-hub` — clean.
- `GOMAXPROCS=4 go test -count=1 -run '<all 182 Test funcs in app_rpc_test.go>' ./cmd/evener-hub` — ok, 12.5s, zero FAIL lines.
- `GOMAXPROCS=4 go test -count=1 ./cmd/evener-hub` — `ok ... 118.528s`, zero FAIL lines.

obra:
Resolved the registry-credentials finding in 57b4121.

**Change.** `newHubRPCTestServerWithWeb` now creates the three roots and the per-server credentials store *before* the default registry, then builds that registry over them: `registry.WithCredentials(cmdutil.StoreCredentialSource{Store: cfg.CredsStore})` and `registry.WithStateRoot(cfg.HubStateRoot)`, appended after `extra` so they outrank `cmdutil.LoadRegistry`'s process-wide pair. `WithOffline(true)`, `WithoutCache()` and `WithNoUserLayer()` are unchanged, so resolution stays offline and env-only; only the credential and state layers move off the shared files.

**Test first.** `TestHubRPCTestServerRegistryReadsItsOwnCredentials` stores a key for `anthropic` on server A over `evener/auth/apiKey/set` and asserts A's answer resolves from the store while server B's `evener/auth/status` does not. Against the old fixture it failed on:

```
app_rpc_test.go:11168: first server resolved "none" right after storing a key; its registry is not reading the store its handlers write
```

("none" because the reload behind `apiKey/set` read the shared `credentials.toml`, which nothing writes now that the store is isolated.)

**Caller audit.** Zero. `GOMAXPROCS=4 go test -count=1 ./cmd/evener-hub` is green (65s, no FAIL lines): no existing test was observing the shared store through the registry, so none needed an explicit store or an opt-out.

**Gates.** `gofmt -l` clean; `go vet ./cmd/evener-hub` and `go vet -tags evenerfuzz ./cmd/evener-hub` clean; full hub package green.

obra:
Roborev findings on 57b4121, addressed in c738652.

**Low (fixture depends on the shared credentials file) — fixed.** `cmdutil.LoadRegistry` calls `credentials.LoadStore(cmdutil.CredentialsPath())` before it appends the caller's options, so the fixture's `WithCredentials` override came too late and every hub RPC test server still read the process-wide `credentials.toml`. `newHubRPCTestServerWithWeb` now builds its default registry with `newTestRegistry(t, cfg.HubStateRoot, "", cfg.CredsStore, nil)` — the same direct `registry.Load` the file's explicit-store fixtures already use, over this server's own store and state root, keeping `WithOffline(true)`, `WithoutCache()` and `WithNoUserLayer()` (and, like every other `newTestRegistry` caller, no ambient environment).

New guard: `TestHubRPCTestServerRegistryIgnoresTheProcessCredentialsFile` points `EVENER_CREDENTIALS_CONFIG` at an unparseable file, then builds a fixture server with a nil `CredsStore` and reloads its registry. It failed first on the shared read — `registry: credentials: credentials: parse .../credentials.toml: toml: line 1: expected '.' or ']' to end table name` — and passes on the fix.

**Medium (`newHubAuthControllerWithStore` ignores its state-root argument) — deferred to #991.** That is production code in `cmd/evener-hub/app_auth.go`, outside this test-only change; it gets its own PR. The fixture comments here name the bug as the reason a nil `CredsStore` needs its own default.

Gates: `gofmt` clean; `go vet ./cmd/evener-hub` and `go vet -tags evenerfuzz ./cmd/evener-hub` clean; full `GOMAXPROCS=4 go test -count=1 ./cmd/evener-hub` green (87s, no FAIL lines).

obra:
Roborev findings on 927941a, addressed in 00b96cb2d (branch merged with main in 75ac066).

**Medium 2 (injected registry, nil CredsStore) — fixed.** `newHubRPCTestServerWithWeb` now refuses an injected `Registry` without its `CredsStore` (`t.Fatalf`: "an injected Registry must come with the CredsStore it resolves from; without it evener/auth/apiKey/set writes a store the registry never reads") instead of minting an unrelated store for the auth controller. Every fixture caller that tripped it builds one store through the new `newTestCredentialsStore` and hands it to both `newTestRegistry` and `CredsStore`: the five `tomlPath` callers in app_rpc_test.go, plus `TestHubRPCThreadStartAllowsIntentionallySkippedLaunchProvider`, whose `newSpawnGateRegistry` grew an optional store for that one call. The fixture's own nil-store branch and `TestHubRPCRegistersExpectedHandlerSet` use the same helper. Tests that build a WebConfig and go straight to `NewWebServer` (sandbox_test.go, appwire_catalog_test.go, web_test.go, app_threadread*_test.go) never pass through the fixture and are unchanged.

**Medium 3 (nil env map) — fixed.** `TestHubRPCAuthStatusPrefersStoredOAuthOverEnv` and `TestHubRPCAuthStatusIgnoresAPIKeyForTheCodexInstance` now own their registry: `envRegistryAuthConfig` builds a `newTestRegistry` over the `oaitest.IsolateOpenAIAuth` state root with an explicit env map carrying `OPENAI_API_KEY`, and sets `HubStateRoot` to that same root so the auth controller and the registry agree on where `auth/openai-codex.json` lives. Each test gained a positive control: `evener/auth/status` for the plain `openai` provider must report `ActiveSource: env:OPENAI_API_KEY`. That control failed first on the old fixture — `status(openai)={... SignedIn:false ActiveSource:none ...}, want OPENAI_API_KEY to resolve from the environment` — which is exactly the vacuity the finding pointed at. It passes now; the Codex assertions are unchanged.

**Medium 1 (OAuth state directory from process-wide XDG) — production bug, fixed in #1011.** `newHubAuthControllerWithStore` ignoring its state root is production code in app_auth.go, out of scope for this test-only PR. #1011 makes the controller keep OAuth records in the registry's state root, which with this fixture's per-test registry gives each server its own OAuth directory by construction. The end-to-end OAuth isolation assertion for this fixture can only pass once that lands, so it follows as soon as #1011 merges: here if this PR is still open, otherwise as a small follow-up.

**Gates.** gofmt clean; `go vet ./cmd/evener-hub` and `go vet -tags evenerfuzz ./cmd/evener-hub` clean; `GOMAXPROCS=4 go test -count=1 -run '<every Test func in the three touched files + TestHubRPCTestServer.*>' ./cmd/evener-hub` ok, zero FAIL lines.

obra:
race-root on 00b96cb was not a race: `TestHubRPCAgentsDocRefusesWithoutAConfigRoot` (landed on main after this branch forked) asserts the agents-doc handlers refuse when no launch config root resolves, and this fixture now defaults `LaunchConfigRoot` to a temp dir, so the handlers found one:

```
app_rpc_agents_doc_test.go:575: get error = <nil>, want one naming the unresolved user config root
```

Fixed in 7a29d06b8. The fixture's server start is now `startHubRPCTestServer`, which serves a config exactly as given; `newHubRPCTestServerWithWeb` applies its per-test defaults and then calls it. The refusal test goes to `startHubRPCTestServer` directly with the empty root it is about (and its own credentials store, so it still stays off the shared one). Reproduced locally on 00b96cb before the change, passes after.

Gates: gofmt clean; `go vet` both tags clean; every test in app_rpc_agents_doc_test.go plus `TestHubRPCTestServer*` ok; full `GOMAXPROCS=4 go test -count=1 ./cmd/evener-hub` ok (63s, zero FAIL lines).

obra:
Roborev Medium on 7a29d06 (fixture registry root vs auth-controller OAuth root) is the same production bug as finding 1 on 927941a: `newHubAuthControllerWithStore` in app_auth.go ignores its state root, so nothing this test-only PR does can put the controller on the fixture's per-test registry root.

#1011 (roborev-clean, CI green, awaiting review) fixes it by handing the controller the wired registry's `StateRoot()`. Once it lands I merge main here, which makes every fixture server's controller and registry agree on the per-test root by construction, and add the end-to-end OAuth isolation assertion that pins it (two default fixtures, a record seeded under one's registry root, `evener/auth/status` signed in on that server and not the other). Landing that assertion before #1011 would fail CI here, so this stays open on that dependency.

obra:
Roborev Medium on 7a29d06 (fixture registry root vs auth-controller OAuth root) — closed now that #1011 is on main, merged here in 2ac9736.

**Assertion (b873a6e).** `TestHubRPCTestServerGivesEachTestItsOwnOAuthState`: two default fixture servers, an OAuth record seeded under the first one's `HubStateRoot`, `evener/auth/status` for `openai-codex` on both. The first signs in from its own record; the second stays signed out with no stored OAuth. It failed first on this branch before #1011, with the controller reading the process-wide XDG root (`first status={... SignedIn:false ActiveSource:none ...}, want signed in from the record under its own state root`), and passes on the merge.

**Two more tests migrated (07c8f8e).** The full hub package on the merge showed `TestHubRPCAuthStatusUsesUserScopedOpenAIAuth` and `TestHubRPCAuthLogoutRemovesUserScopedOpenAIAuth` failing the same way the env-vs-OAuth tests would have: they seeded `auth/openai-codex.json` under `XDG_STATE_HOME` and used the default fixture, whose registry and (now) controller live on a per-test root. They seed the isolated root and pass a registry plus `HubStateRoot` on it through `stateRootAuthConfig`, the env-vs-OAuth helper generalized to take the env map. The logout test also checked `LoadAuth` for `"openai"` instead of `"openai-codex"`, so its removal assertion could never find the record; it now names the instance it seeded.

**Gates.** gofmt clean; `go vet` both tags clean; every test in app_auth_test.go plus `TestHubRPCTestServer*` ok (35 PASS); full `GOMAXPROCS=4 go test -count=1 ./cmd/evener-hub` ok (62s, zero FAIL lines).

obra:
lint-golangci on 07c8f8e was mine: gocritic `deferInLoop` on the `defer client.Close()` inside the per-server loop of the new `TestHubRPCTestServerGivesEachTestItsOwnOAuthState`. Fixed in 8bebfd531: the per-server status query is a closure that owns its client, called once per server. `golangci-lint run ./cmd/evener-hub/` (same binary version as CI, 2.13.1) reports 0 issues; gofmt, `go vet` both tags and the `TestHubRPCTestServer*` tests are clean.

obra:
`web` on 8bebfd5 failed on one vitest case, `src/panes/spawn/Spawn.test.tsx > successful keyless testing refreshes availability without an auth notification` (a `waitFor` timing out on `expected <button …> to be null`). This branch changes only four Go test files under cmd/evener-hub, and the same `web` job is green on main's last two runs (8badc6c, cf4df74), so it is a flake in untouched code. I'll re-run the failed job as soon as the workflow run finishes (GitHub refuses a rerun while sibling jobs are still running).

obra:
Roborev Low on 8bebfd5 (stacked doc comments) — fixed in 7776f4192. The OAuth-state test had been inserted between `TestHubRPCTestServerRegistryReadsItsOwnCredentials`'s doc comment and its declaration; the comment now sits directly on that test again and the OAuth-state test carries only its own. gofmt, `go vet`, `TestHubRPCTestServer*` and `golangci-lint run ./cmd/evener-hub/` (0 issues) clean.

obra:
`race-root` on 7776f41 (a comment-only push) failed on `internal/appserver` `TestServeWebSocketUnsubscribeWaitsThroughSubscriptionRegistration` — "unsubscribe left a stale registered subscription" — the open flake #976, in a package this branch does not touch. Queued a rerun of the failed job for when the run completes.

Issue #754, opened 2026-09-01
resolveModelSwitchTarget lists models with an unstamped context on a live session (unattributed API-log attempts)


obra:
Found during #745's fix (PR #752): resolveModelSwitchTarget (agent/session_set_model.go:83, used by SetModel and subagent model selection) issues its live model listing with a bare context.Background() even though it runs on an already-existing session with a real s.id in scope — the same unattributed-API-log pattern #745 fixed for pre-session listings, but on a session that could attribute trivially. Same fix shape: stamp the session id via the attribution mechanism the #752 tests pin.

PR #992, opened 2026-09-08
test(agent): pin plugin-agent model listing attribution

### LINKED ISSUE #799: resolvePluginAgentModel's live listing has no dedicated attribution test (relies on turn-ctx inheritance)
Noted during #754's adversarial review: resolvePluginAgentModel's listing call (agent/subagent_model_selection.go:148) is correctly attributed TODAY via ctx inheritance from the turn-level stamp (session_lifecycle.go:629 → execToolBatch → tool handler → selectSubagentModel → resolvePluginAgentModel → Client.Models), but has no dedicated attribution test — it would silently regress to unattributed if that ctx chain ever breaks. Low priority; add a regression test pinning the inheritance. Not part of #754 (which fixes the bare-Background sites).

obra:
## Mechanism

`resolvePluginAgentModel`'s live model listing (`agent/subagent_model_selection.go`) applies no API-log attribution stamp of its own. It is attributed today only because it inherits the turn-level stamp `session_lifecycle.go` puts on the turn ctx (`llm.WithAPILogContext(ctx, s.id)`), which then rides untouched through `execToolBatch` -> the tool handler -> `selectSubagentModel` -> `resolvePluginAgentModel` -> `Client.Models`. Nothing pinned that chain, so a break at any hop would silently route the listing to the shared `sessions/unattributed.api.jsonl` bucket.

## Change

Test-only. Adds `TestPluginAgentModelListingAttribution` to `agent/apilog_session_attribution_test.go`, next to PR #754's `TestModelSwitchLiveListingAttribution` and using the same mechanism: an `attemptRecordingLiveModels` adapter opens and completes a real `llm.APIAttempt` against whatever ctx the listing seam is given, a real `llm.NewSessionAPILogger` routes it, and the assertion is that the session's own log holds records while the unattributed bucket was never created.

The session runs a plugin agent whose model (`gpt-5.3`) differs from the session's own (`gpt-5.2`), which is what forces the listing at issue; `selectSubagentModel` is called with exactly the ctx a turn hands it. Asserting the selection lands on the plugin's model with no fallback warning keeps the test from passing vacuously: a candidate equal to the session profile returns before listing, and a failed or incomplete listing yields a warning and the base model instead.

`attemptRecordingLiveModels` gained a variadic list of ids to serve (defaulting to the previous single `gpt-5.2`), so the plugin agent's model can be a second listed id. The shared assertion's failure message dropped its now-inaccurate "pre-session" wording — three of its four call sites are not pre-session listings.

## Failing first

With the inheritance temporarily broken at the narrowest point (`context.WithTimeout(context.Background(), ...)` in place of `ctx` at `subagent_model_selection.go:147`):

```
--- FAIL: TestPluginAgentModelListingAttribution (0.10s)
    apilog_session_attribution_test.go:750: live model listing landed in unattributed bucket (stat err=<nil>):
        {"kind":"api_attempt",...,"provider_instance":"openai","request_model":"*",...,"endpoint":"https://scripted.invalid/v1/models",...}
```

Production code was restored untouched; this PR contains no production change.

## Smoke gates

- `gofmt -l agent/apilog_session_attribution_test.go` - clean
- `go vet ./agent` and `go vet -tags evenerfuzz ./agent` - clean
- `GOMAXPROCS=4 go test -count=1 -run 'TestPluginAgentModelListingAttribution|TestModelSwitchLiveListingAttribution|TestPreSessionLiveModelListingAttribution|TestNewSessionReleasesPreSessionAPILogRouteOnMembershipFailure|TestNewSessionAttributesOtherProviderStartupListingToSessionAPILog' ./agent` - ok (covers every caller of the changed helper)

Closes #799

### DIFF
--- a/agent/apilog_session_attribution_test.go
+++ b/agent/apilog_session_attribution_test.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -710,7 +711,12 @@ func TestSessionCloseReleasesAPILogRoute(t *testing.T) {
 // chain below the stamp under test.
 func TestPluginAgentModelListingAttribution(t *testing.T) {
 	stateDir := t.TempDir()
-	adapter := &fakeAdapter{name: "openai", liveModels: attemptRecordingLiveModels("openai", "gpt-5.2", "gpt-5.3")}
+	var listings atomic.Int64
+	listModels := attemptRecordingLiveModels("openai", "gpt-5.2", "gpt-5.3")
+	adapter := &fakeAdapter{name: "openai", liveModels: func(ctx context.Context) ([]registry.Model, error) {
+		listings.Add(1)
+		return listModels(ctx)
+	}}
 	client := registryClient(t, map[string]registry.Provider{
 		"openai": {Base: "openai", APIKey: "k", Models: modelRows("gpt-5.3")},
 	}, adapter)
@@ -729,9 +735,17 @@ func TestPluginAgentModelListingAttribution(t *testing.T) {
 		},
 	}
 
+	listingsBefore := listings.Load()
 	selected, err := sess.selectSubagentModel(llm.WithAPILogContext(context.Background(), sess.ID()), "", "reviewer")
 	if err != nil {
 		t.Fatalf("selectSubagentModel: %v", err)
+	}
+	// The fixture registry already carries gpt-5.3, so a selection that never
+	// listed could still land on the plugin's model while the session log held
+	// nothing but the listing NewSession issued. Counting the listings the
+	// selection itself drives is what holds the seam under test.
+	if got := listings.Load() - listingsBefore; got != 1 {
+		t.Fatalf("live model listings during selection = %d, want 1", got)
 	}
 	// Landing on the plugin's own model with no fallback warning is reachable
 	// only through the listing under test: a candidate equal to the session's
--- a/agent/apilog_session_attribution_test.go
+++ b/agent/apilog_session_attribution_test.go
@@ -3,6 +3,7 @@ package agent
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


obra:
Addressed the listing-count finding in 0349f3a13 (test-only).

`TestPluginAgentModelListingAttribution` now wraps the scripted `attemptRecordingLiveModels` listing in a counting closure on the fake adapter and asserts that exactly one listing happens across the `selectSubagentModel` call (counted as a delta, since `NewSession` issues its own pre-session listing first). `attemptRecordingLiveModels` itself is unchanged, so the other callers behave identically.

Mutation that proves the assertion is live: short-circuiting `resolvePluginAgentModel` just above the listing (`return pluginAgentModelResolution{profile: candidate}` when `s.client.Resolve(candidate.ID()+"/"+candidate.Model())` succeeds, `agent/subagent_model_selection.go:147`) fails only the new check:

```
--- FAIL: TestPluginAgentModelListingAttribution (0.20s)
    apilog_session_attribution_test.go:748: live model listings during selection = 0, want 1
```

With that same mutation in place and the counter relaxed to `!= 0`, every pre-existing assertion in the test still passed (no warning, model `gpt-5.3`, session log attributed, no unattributed bucket) — exactly the gap the review flagged. Production code was restored byte-identical; `git diff --stat` shows only the test file.

Gates: `gofmt -l agent/apilog_session_attribution_test.go` clean; `go vet ./agent` and `go vet -tags evenerfuzz ./agent` clean; `GOMAXPROCS=4 go test -count=1 -run '^(...)$' ./agent` over all nine `func Test`s in the touched file — 9 PASS, 0 FAIL.

obra:
Roborev Low (end-to-end delegate path): fixed in 35ab6901.

Added `TestPluginAgentModelListingAttributionThroughTurn` in the same file. It scripts a turn whose tool batch calls the `delegate` tool with a plugin agent whose model forces the live listing, so the listing is attributed only if `processInputKindWithProvenance`'s stamp survives `execToolBatch` -> `execTool` -> the delegate handler -> `createDelegate` -> `selectSubagentModel`.

Proved live: changing `session_tools.go:777` (where `execTool` derives the tool context) to `context.WithValue(llm.WithAPILogContext(ctx, ""), ctxToolCallID, call.ID)` fails **only** the new test — the listing lands in `unattributed.api.jsonl`; the hand-stamped `TestPluginAgentModelListingAttribution` and every other attribution test in the file stay green. Production code restored byte-identical (`git diff` touches the test file only).

The delegate call asks for `fork_context`, which `createDelegate` refuses on the first check past the selection. That keeps the turn from spawning a child session onto the same scripted adapter, and the refusal is itself evidence the listing decided the model: it fires only because the selection landed on the plugin's `gpt-5.3` rather than the session's `gpt-5.2`.

The direct-call test is kept: it is the sibling shape PR #754 established (`TestModelSwitchLiveListingAttribution` calls `selectSubagentModel` directly for both of its subagent subtests), and it still pins the listing count.

Gates: `gofmt -l` clean; `go vet ./agent` and `go vet -tags evenerfuzz ./agent` clean; all 10 `func Test` in `agent/apilog_session_attribution_test.go` pass with `GOMAXPROCS=4 go test -count=1` (new test also `-count=5`).

PR #995, opened 2026-09-08
feat(agent): let any job_watch carry a note

### LINKED ISSUE #978: job_watch rejects note on non-timer (condition) watches
The job_watch tool definition only admits note alongside after_seconds or repeat_seconds (agent/job_watch.go: a note is only admitted alongside timer fields; formatJobWatch comments 'A note is only admitted alongside after_seconds or repeat_seconds, so a bare check cannot label a non-timer watch').

A condition watch (output_match / events / progress_interval_ms) has no way to carry a human-readable label for why it was armed. The note is the watch's own prose payload — it is what the fired notification shows back — and there is no reason a condition watch should be barred from carrying one.

Repro: job_watch(operation=create, source=<job_id>, output_match=ready, note='why I care') → invalid_request.

Expected: note accepted on any create; surfaced in list/inspect condition text and the fired notification like timer notes.

obra:
## Mechanism

`validateWatchTriggerShape` admitted `note` only inside the `watchArgsIsTimer` branch, so every condition create (`output_match`, `events`, `progress_interval_ms`) with a note failed `invalid_request: note applies to timers`. Two downstream renderers repeated the same assumption: `watchConditionSummary` gated its `note:` part on `cfg.timer`, and only the timer fire path (`fireProgressTick`) stamped `Note` onto the notification. `watchArgsIsOutputMatchOnly` also excluded a noted request from terminal catch-up, on the reasoning that a note marked a mis-shaped timer.

## Change

- Drop the `note applies to timers` rejection.
- `watchNotificationFromWatch` stamps `cfg.note`, so every rail that builds a watch notification (output match, session event, progress tick, terminal catch-up) carries it; the timer path no longer sets `Note` separately.
- `formatJobNotificationBlock` renders the note through a new `withNotificationNote` helper on all three watch bodies (timer, session watch event, job-target watch), body-escaped exactly as timer notes already were.
- `watchConditionSummary` shows the note for any watch, so `job_list`, `inspect`, and the watch history all label it.
- `watchArgsIsOutputMatchOnly` no longer excludes a noted request: a note is not a trigger, so an `output_match` create on an already-terminal job still gets its catch-up scan, and the note rides that scan's notification.
- Tool definition text for `note`, the `formatJobWatch` comment that stated the old rule, and the matching `docs/job-control.md` sentences updated.

## Tests

New `agent/job_watch_condition_note_test.go`. All three failed first with `configure: invalid_request: note applies to timers`:

- `TestConditionWatchNote_AcceptedAndSummarised` — the issue's repro (`output_match` + `note`), then asserts `list` and `inspect` condition text carries `note: why I care`.
- `TestConditionWatchNote_RidesTheFiredNotification` — a real `feedJobOutput` match puts the note on the notification and in the rendered block.
- `TestSessionConditionWatchNote_RidesTheFiredNotification` — same for the session-target rail, whose block has its own branch.

`TestValidateWatchTriggerShape_TimerRules` lost its `note without timer` rejection row and gained a positive assertion that a noted condition watch validates.

## Smoke gates

- `gofmt -l` on all changed files: clean.
- `go vet ./agent` and `go vet -tags evenerfuzz ./agent`: clean. `go vet ./agent/internal/tool`: clean.
- `GOMAXPROCS=4 go test -count=1 -run 'Watch|Timer|Notification|Note' ./agent`: ok (9.9s). `go test -count=1 ./agent/internal/tool`: ok.
- `go build ./...`: ok.

Closes #978


obra:
Both findings addressed in f8daa6c.

**Finding 1 (stale reject-table row).** `TestTerminalCatchupRejectsEventsCondition`'s `"output_match plus note"` case was written when `watchArgsIsOutputMatchOnly` excluded a noted request from terminal catch-up. This PR removed that exclusion, so the case failed as pushed: `result {... Fired:true TerminalCatchup:true Status:completed} err <nil>, want target_terminal` (job_watch_events_test.go:978). The row is gone and the table comment no longer claims a note disqualifies catch-up. Its replacement, `TestTerminalCatchupCarriesTheNote`, sits beside the other positive catch-up tests and asserts the opposite: a noted `output_match` create on a terminal job reports `fired`+`terminal_catchup`, installs no live watch, and the single enqueued notification carries `Note: deploy gate` both on the struct and in the rendered block.

**Finding 2 (note dropped on the stable-receiver rail) — confirmed.** A `source="parent"` / `source="dlg_..."` watch gets a synthesized send from `applyStableReceiverWatchSend`, so its fire is delivered as a watch frame, not a rendered notification body. `formatJobNotificationBlock` returns early for `n.WatchSend != nil` without reaching `withNotificationNote`, and `buildWatchFrame` never serialized the note — `snapshotWatchSendFrame` did not even copy it onto the config it rebuilds. New `TestStableReceiverWatchNote_RidesTheDurableFrame` arms a noted `dlg_source` communicate watch, fires it, and reads the durable pending frame; it failed first with `stable-receiver frame lacks the note:` followed by a frame containing only `watch_id`/`delivery_id`/`job_id`/`trigger`/`provenance`/`event`. The frame now writes a `note:` field between `trigger` and `provenance`, through `writeWatchFrameTopField` so a multi-line note stays indented and cannot forge a top-level frame line, bounded at `watchMessageMaxChars` — the same bound `watchConditionSummary` and note storage use. The whole frame is still capped at `watchFrameMaxChars`, so the fuzz oracle's boundedness invariant is untouched.

**Smoke gates.** `gofmt -l` on the three changed files: clean. `go vet ./agent` and `go vet -tags evenerfuzz ./agent`: clean. `GOMAXPROCS=4 go test -count=1` over the 80 named tests in `job_watch_events_test.go`, `job_watch_condition_note_test.go`, `job_watch_timer_args_test.go`, `job_watch_send_test.go`, and `delegate_resource_watch_test.go` (the last two added because `buildWatchFrame` is production code): ok, 0 FAIL. `TestTerminalCatchupRejectsEventsCondition` explicitly: ok. `go build ./...`: ok.

obra:
Addressed the Low finding on model-facing note guidance in c9101f452. Rewritten sentences:

- `agent/internal/tool/definitions.go:329` — lead sentence dropped its note clause: now "Wake yourself later: `after_seconds` fires once and `repeat_seconds` fires every interval." Timer wording stays on the timer fields.
- `agent/internal/tool/definitions.go:332` (new) — added "`note` is the watch's own prose payload: any create accepts one, and it comes back with every fire so you know why you armed it."
- `agent/internal/tool/definitions.go:333` — "Each `create` is a new timer" → "Each `create` installs a new watch"; the rest of that sentence (clear-and-create to change a note, clear a timer before reporting done, escaping) is unchanged.
- `agent/prompts/sections/background-jobs.md:22-24` — the timer bullet ends at `repeat_seconds`; the note now stands on its own: "Any `job_watch` create takes a `note` saying why you armed it and, for a loop, where you are; it rides every fire of that watch, and to advance it you clear and create."

`docs/job-control.md` already described the note as labelling "any watch, timer or condition" (line 757), so it needed no change; the "Wake yourself later" table row at line 106 is timer-specific by design.

No test asserted the changed sentences. The only prose pin in this area is `TestDefJobWatch_TimerProperties`'s `HasPrefix(def.Description, "Wake yourself later:")`, which still holds; per repo convention tests assert behaviour, not prose, so nothing was added.

Gates: `gofmt -l` clean; `go vet ./agent/internal/tool/ ./agent/` clean with and without `-tags evenerfuzz`; `GOMAXPROCS=4 go test -count=1` green for `TestDefJobWatch_TimerProperties`, `TestDefJobWatchUsesSourceAndOmitsSend`, `TestDefJobWatchOptionalTriggerFieldsAreNullable`, `TestDefJobWatchRequiresOperationAndWatchIDForClear`, `TestBuildSystemPrompt_IncludesBackgroundJobsSection`, `TestBuildSystemPrompt_PinsAntiPollGuidance`, `TestSubagentPrompt_DoesNotIncludeBackgroundJobsSection`; `go build ./...` clean.

PR #996, opened 2026-09-08
test(apptranscript): gate the entries projection on allocations, not elapsed time

### LINKED ISSUE #944: flaky: TestItemTurnsFromEntriesLargeFixtureTiming compares two wall-clock measurements under CI load
`TestItemTurnsFromEntriesLargeFixtureTiming` (`internal/apptranscript/turnsfromentries_timing_test.go`) fails intermittently in CI's `tests` job. Two sightings so far, on unrelated branches, both re-ran green:

| when | PR | head | branch touches `internal/apptranscript`? |
|---|---|---|---|
| 2026-09-06 | #924 | `0d7d545cf` | no |
| 2026-09-07 | #941 | `cb38e3d7c` | no |

## Why it is load-sensitive by construction

The assertion is a ratio between two wall-clock measurements taken in the same process:

```go
ratio := float64(fileElapsed) / float64(entriesElapsed)
if ratio < 3 {
    t.Fatalf("file form was only %.1fx slower than the entries form; ...")
}
```

`fileElapsed` covers scan + decode + project; `entriesElapsed` covers project alone. The property under test is real and worth pinning — the entries projection exists precisely to skip the file I/O and decode pass — but a ratio of two short measurements on a shared runner compresses when the machine is busy: the denominator is the smaller quantity, so noise added to it moves the ratio far more than the same noise added to the numerator. Nothing in the test bounds either measurement or repeats them.

## Suggested directions

Any of these keeps the property without the flake:

- Measure the two forms several times and compare medians, so one descheduled interval cannot decide the result.
- Assert on work actually avoided rather than time: the entries form performs no file read and no decode, which can be counted (bytes read, entries decoded) instead of timed.
- Keep the timing check but make it advisory: log the ratio and fail only below a much lower floor, with the sharp assertion moved to a benchmark that CI does not gate on.

I have not changed it — I hit it twice while landing unrelated work and re-ran both times. Filing so the next person who sees it does not have to re-derive that it is not their change.

obra:
`TestItemTurnsFromEntriesLargeFixtureTiming` gated on `fileElapsed/entriesElapsed >= 3`: a ratio of two short wall-clock measurements taken in one process. The entries form is the smaller denominator, so a descheduled interval there moves the ratio hardest, which is why CI hit it twice on branches that never touched `internal/apptranscript`.

## Direction, and why

The issue offers three. I took the second — assert on work avoided rather than time — because the runtime already counts that work, so no production code has to expose anything. The file form scans every line and decodes every entry; the entries form projects already-decoded entries. That difference shows up as allocations, and `runtime.MemStats` counts them.

Over the same 20k-entry fixture the allocation ratio is **2.82x, identical to two decimal places across runs**, where the timing ratio wanders with machine load. The gate floors it at 2x. Elapsed time, allocated bytes and the ratio are logged for the reader; nothing asserts on duration any more.

Medians-of-N (direction 1) would still be timing, with more code. An advisory log (direction 3) would pin nothing. This package already asserts on work this way — `TestTurnCacheToolHeavyAppendDoesNotPersistCumulativeResolver` in `turn_index_test.go` uses `ReadMemStats` deltas and `testing.AllocsPerRun`.

## TDD

The replacement assertion was shown live before it was trusted. I temporarily made `ItemTurnProjectionFromEntries` do the decode pass it is supposed to skip (`json.Marshal` + `transcript.DecodeEntry` per entry) and ran it:

```
allocation ratio: 1.36x
the file form made only 1.36x the allocations of the entries form
(file=899646, entries=659423, over 20000 entries); the entries projection's
skip of the file scan + decode pass is its entire reason to exist
```

Restored, the same test passes at 2.82x. Calibration is recorded in the test's doc comment: 2x floor, ~2.8x observed, ~1.4x when the skipped pass is put back.

Two judgement calls worth flagging:

- The test is renamed (`TestItemTurnsFromEntriesAvoidsFileScanAndDecode`, file `turnsfromentries_work_avoided_test.go`) because it no longer times anything.
- The `testing.Short()` skip is gone with the timing measurement that excused it — the test runs in 0.24s and is now a deterministic gate.

## Smoke gates

- `gofmt -l` on the changed file: clean.
- `go vet ./internal/apptranscript` and `go vet -tags evenerfuzz ./internal/apptranscript`: clean.
- `GOMAXPROCS=4 go test -count=1 -run TestItemTurnsFromEntriesAvoidsFileScanAndDecode ./internal/apptranscript`: pass, three consecutive runs, 2.82x each time.
- No non-test code changed, so no `go build`.

Closes #944


obra:
Addressed the roborev finding on 45f0f7c (allocation ratio measured from one unisolated `runtime.MemStats` window per form).

**Change (test-only, `turnsfromentries_work_avoided_test.go`)**: each form is now run once to warm it — first-call lazy init and buffer growth land outside the measured window — and the measured counts come from `testing.AllocsPerRun`, which pins `GOMAXPROCS` to 1 and averages over 3 runs, so a stray allocation from elsewhere in the process moves the average by a fraction of one run instead of all of itself. The 2x ratio floor is unchanged. Elapsed time and allocated bytes are still logged (from the warm-up run), still never asserted on. No production code touched.

**Mutation, shown live again**: with `ItemTurnProjectionFromEntries` temporarily doing the `json.Marshal` + `transcript.DecodeEntry` pass it is supposed to skip, the new form of the test fails:

```
allocation ratio: 1.36x
the file form made only 1.36x the allocations of the entries form
(file=899318, entries=659296, over 20000 entries)
```

Production code was restored byte-identical afterwards; the diff carries only the test file.

**Five consecutive `-count=1` runs**: 2.82x, 2.82x, 2.82x, 2.82x, 2.82x. The underlying counts move by at most 2 allocations out of ~899k (file 899316-899321, entries 319285-319286) while wall times over those runs ranged 1.2s to 5.7s on a loaded machine — which is the point of the change.

**Under `-race`** (the finding names race instrumentation): passes at 3.31x. Instrumentation inflates both forms and the file form more, so it moves away from the floor, not toward it.

Gates: `gofmt -l` clean; `go vet ./internal/apptranscript` and `go vet -tags evenerfuzz ./internal/apptranscript` clean; the file's only test run repeatedly as above, all pass.

One thing to flag: this test lives in the root module, so the `race-root` CI lane runs it, and there it takes ~48s (the file form costs ~7s per pass under race). I set the averaging count to 3 rather than 5 to hold that down. If that lane's budget is tight, shrinking the 20k-entry fixture is the lever — the ratio is structural rather than fixture-size dependent, though the calibration numbers in the doc comment would need re-measuring.

PR #1009, opened 2026-09-08
fix(flake-timing-ratio): gate the entries projection on allocations, not elapsed time

obra:
Root cause: TestItemTurnsFromEntriesLargeFixtureTiming asserted fileElapsed/entriesElapsed >= 3, a ratio of two short wall-clock measurements in one process. Both forms share the one projection path (appendProjectedEntry + groupedAppTurnProjection, internal/apptranscript/logical_turn.go:129,147); the file form only adds scan+decode (internal/apptranscript/apptranscript.go:699-711, and note the scan pass strictly decodes every line at :830 before decoding again for projection at :700). Over a 5.3MB page-cache-hot fixture that extra work costs barely more than the per-turn projection itself, and the entries form is the smaller denominator, so one descheduled interval there compresses the ratio hardest: CI measured 533ms vs 288ms = 1.9x while an unloaded host measures 332ms vs 31ms = 10.8x. No product regression (git log -S shows no recent cost change to either function). The "deliberately loose floor machine load cannot flake" comment was the falsified premise.

Fix (direction b per the brief): the property is unchanged, now measured as work rather than duration. The file form must pay allocations the entries form never pays; testing.AllocsPerRun (warm-up + GOMAXPROCS=1 + 3-run average) reports 2.82x over the 20k-entry fixture, floored at 2x. Elapsed time is logged, never asserted on. The old timing test is deleted; the new gate runs in short mode too.

Proof: new TestItemTurnsFromEntriesAvoidsFileScanAndDecode passes locally (2.82x vs 2x floor; file 899317 allocs/run, entries 319286 allocs/run); gofmt clean; go vet clean; no references to the old test name remain.

Note: PR #996 on test/944-turnsfromentries-timing-flake pursues the same allocations direction; this lane did its own independent root-cause analysis and diff per the flake assignment.


obra:
Superseded by merged #996 (commits 5c7b693af + fb34f73c8): main now contains the same allocations gate (TestItemTurnsFromEntriesAvoidsFileScanAndDecode, 2x floor, AllocsPerRun with warm-up) and the timing test is already deleted there. Diff of this branch vs current main is cosmetic only (constant rename projectionRuns->measuredRuns plus comment wording). Independent root-cause analysis on this lane converged on the same conclusion: no product regression, wall-clock ratio was the flake. Closing as duplicate; pushing nothing further.

PR #1005, opened 2026-09-08
perf(w2-pane-split): lazy seams for Spawn dialog, composer hook, RailHost

obra:
Three lazy sub-boundaries, no logic change: (1) Spawn ConnectProviderDialog -> lazy+Suspense(null), mounted only on user click; (2) Composer hook-only askDockPending seam (full AskDock still mounted by Session.tsx; composer input untouched); (3) RailHost behind lazy barrel seam (workspace paints behind null fallback). CommandPalette deliberately eager (critical path).

Fix on branch: rail barrel renamed index.ts -> index.tsx (JSX needs tsx).

Verification: tsc clean; rail scope 336/336 pass; vite build clean with RailHost as its own 1.42kB lazy chunk.
Risk: low-med (new loading states on explicitly-opened surfaces only).

PR #1000, opened 2026-09-08
perf(w2-img-store): prefer sha-addressed image URLs over data-URIs

obra:
Image src precedence becomes url > sha-route (/s/{route}/images/{sha} rebuilt from metadata sha when the serving session is known) > data-URI > path > name. Strict 64-hex sha only; unknown sessions/refs keep the data-URI fallback (offline/cached safe). Lightbox/full-size paths unchanged.

Verification (serial runner): tsc clean; reducer 165 + gallery 32 + ToolCallItem 76 + UserMessage/threads 278 all pass; client/server URL-shape agreement verified incl. %2F.
Gap (non-blocking): no unit test for the sha-without-url path (only url-first covered) — follow-up welcome.
Risk: low-med (URL-shape agreement is the failure mode; verified for session-ID charset).

PR #1010, opened 2026-09-08
fix(flake-goal-continuation): wait for the continuation opening event before asserting

obra:
## Root cause

TestGoalContinuationTurnCarriesItsNameOnTheOpeningEvent (agent/session_active_turn_test.go:29) asserted synchronously on a slice appended by the ConsumeEventsLossless drain goroutine. Chain:

- The test registers the drain via ConsumeEventsLossless (test file :35), which spawns `go func() { for ev := range s.events { consume(ev) } }` (agent/session_events.go:96-99) — the only writer of authoritativeConsumer (:93).
- ProcessInputKind(EntryContinuation) -> processOneInput mints the id synchronously (agent/session_lifecycle.go:1536-1537: `runningTurnID, _ = s.mintRunningTurnID()`) and emits EventGoalContinuation synchronously (:2072 in acceptContinuationInput).
- But emission (sendEvent, session_events.go:361 `s.events <- ev`) only hands the event to the channel; the test's append happens on the drain goroutine with no happens-before edge to ProcessInputKind's return. The test then read the slice immediately (:52-56) — a race it always loses under load/scheduling.

Why survey-only: the survey pass runs the FULL suite as one binary with `-test.parallel 6` (cmd/evener-dev/agentshards.go:270), so every parallel-capable test competes for CPU and the drain goroutine can be delayed past the assertion. A targeted local run (-run only this test) has no such contention, so it passes. CI evidence: perf/watch-batch run 34174488095 survey log shows exactly `--- FAIL: TestGoalContinuationTurnCarriesItsNameOnTheOpeningEvent (0.01s)` — a fast fail consistent with the count=0 arm, not the StableTurnID arm. Recent main runs (34177913082, 34177086388) fail only on web-lint/lint-golangci, never on this test — refutes the 'base flake on main' reading; it is a latent test race exposed by parallel scheduling.

## Fix

Signal event arrival (buffered sawContinuation channel, non-blocking send in the consumer) and wait up to 30s (the package's established event-wait convention, e.g. session_attention_test.go) before asserting count == 1 and the turn_m prefix. Test-only change; no production code touched.

## Proof

- `go test -count=1 -run 'TestGoalContinuationTurnCarriesItsNameOnTheOpeningEvent' ./agent/` green 3x (0.918s / 0.764s / 0.786s)
- Siblings green: `-run 'TestGoalContinuationTurn|TestMintRunningTurnID'` ok
- `gofmt -l` clean

### DIFF
--- a/agent/session_active_turn_test.go
+++ b/agent/session_active_turn_test.go
@@ -61,6 +61,10 @@ func TestGoalContinuationTurnCarriesItsNameOnTheOpeningEvent(t *testing.T) {
 
 	select {
 	case <-sawContinuation:
+	// TRIPWIRE: the event is emitted synchronously by ProcessInputKind above
+	// and delivered by the in-process drain goroutine, so it normally arrives
+	// in well under a second. 30s only fires on a genuine delivery hang, not
+	// scheduler contention under a loaded suite.
 	case <-time.After(30 * time.Second):
 		t.Fatalf("timed out waiting for the EventGoalContinuation opening event")
 	}
--- a/agent/session_active_turn_test.go
+++ b/agent/session_active_turn_test.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/spf13/afero"
 
@@ -31,6 +32,11 @@ func TestGoalContinuationTurnCarriesItsNameOnTheOpeningEvent(t *testing.T) {
 
 	var mu sync.Mutex
 	var continuations []events.GoalContinuationData
+	// The drain ConsumeEventsLossless registers delivers on its own goroutine
+	// (session_events.go), so nothing guarantees it has run by the time
+	// ProcessInputKind returns. Wait for the event rather than asserting on a
+	// slice the consumer may not have appended to yet.
+	sawContinuation := make(chan struct{}, 1)
 	drained := make(chan struct{})
 	s.ConsumeEventsLossless(func(ev events.SessionEvent) {
 		if ev.Kind != events.EventGoalContinuation {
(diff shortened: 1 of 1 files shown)


obra:
Main-branch lint blocker: PR #998 (commit f45c66402) added camelCase json tags in `cmd/evener-tui/hub_notifications.go` that fail tagliatelle on current main — your lint/static failure on that file is not from your changes. Fix is up in #1013 ("fix(lint): repair main's tagliatelle and Biome import-sort findings"); once it merges, rebase/re-run and the hub_notifications.go:609-612 json(snake) errors should clear.

obra:
Cluster B (biome organizeImports) verdict: **not branch-caused — no action on this branch.**

Evidence:
- Branch-owned files (`git diff refs/remotes/origin/main...origin/perf/flake-goal-continuation-survey --name-only`): only `agent/session_active_turn_test.go`. Zero frontend files.
- The web-job failure (`biome ci src`, run 34180536539/job/101918616826) flags the same 5 main-state files as #967/#1000/#1005: `DocPane.tsx:1:1`, `AdvancedOptions.tsx:16:1`, `Rail.tsx:1:1`, `RailRow.tsx:37:1` (+ format), `SessionMenu.tsx:13:1` — all `assist/source/organizeImports`.
- Offending lines come from merged PR #1001 (commit `51840808f`, perf/w2-barrel, merged 2026-09-08T01:34:06Z with its own web check FAILURE). This branch's merge-base is current main tip, i.e. it already contains #1001 — confirming the dirt is main-state.

Disposition: fix belongs on main (Cluster A / main-fix lane), then update this branch. Do NOT touch these files here.

obra:
Follow-up: the 5 biome-flagged files are already repaired by PR #1013 (fix/main-lint-tagliatelle-biome, cherry-pick of the import-sort hunks + tagliatelle nolints — verified same 6-file scope). No separate fix needed here; once #1013 merges, rebasing this branch onto main should clear web-lint (plus the branch-owned items already handled: F1 reducer fix on #1000, rail/index.tsx sort on #1005).

obra:
Correction: PR #1013 (nolint approach) is closed — the Go half was the wrong fix. Merged PR #1008 (be18ad663) fixed it properly by deleting the duplicated streamDeltaChunk struct and decoding into appwire.ToolOutputDeltaParams instead, plus stating the tag contract in appwire/doc.go. The frontend import-sort half of #1013 was identical work and is covered. Action for this PR: rebase onto current main and re-run CI.

PR #1011, opened 2026-09-08
fix(hub): keep the auth controller's OAuth records where the registry reads them

### LINKED ISSUE #991: newHubAuthControllerWithStore discards its state-root argument; the HubStateRoot fallback in newHubAppServer feeds nothing
Noticed while landing #983 (PR #990).

`newHubAuthControllerWithStore` (cmd/evener-hub/app_auth.go:126) declares its first parameter as `_ string` and resolves the OpenAI state directory from the environment instead. Its only production caller, `newHubAppServerWithNavigationAndTrace` (cmd/evener-hub/app_rpc.go:302-309), first computes a `hubStateRoot` fallback (`cfg.HubStateRoot`, else `cmdutil.DefaultStateRoot()`) and passes it in, so that fallback feeds nothing.

Two consequences:

- The fallback and its "defensive only" comment describe a root nobody reads at that site, which misleads the next reader (it misled the analysis behind #981/#983 into treating it as a live default).
- The auth controller ignores an explicitly configured `HubStateRoot` and goes to the env, so a test or a sandbox that sets `WebConfig.HubStateRoot` still gets env-derived auth state.

Fix shape: either make the controller honour the root it is handed and drop the env lookup, or delete the parameter and the fallback computation. The first is what `WebConfig.HubStateRoot`'s comment ("root of hub-level machine state (auth-token, index.db, deletions/)") promises. Either way, a test that sets `HubStateRoot` and asserts where the auth state lands.

obra:
## Mechanism

`newHubAuthControllerWithStore` (cmd/evener-hub/app_auth.go) declared its state root as `_ string` and derived the OpenAI OAuth directory from the process environment, so the root its caller in cmd/evener-hub/app_rpc.go computed fed nothing (#991), and a hub fixture with a private root still touched the shared `auth/<instance>.json` under `XDG_STATE_HOME` (the remaining Medium on #990).

Where those records belong is decided by who reads them. Registry credential resolution reads `auth/<instance>.json` from the registry's state root (`oauthRecordPath` in llm/registry/instances.go), and the hub loads its registry through `cmdutil.LoadRegistry`, which pins that root to `cmdutil.DefaultStateRoot()` regardless of `hub_state_root`; the credential probe (app_credentials.go) and every spawned child resolve the same way. Tying the controller to `cfg.HubStateRoot` would therefore split writes from reads whenever `hub_state_root` is customized: the controller saves a login the registry, the probe and the children never see. That is the Medium roborev raised on the first revision of this PR.

## Change

- The controller honors the state root it is handed; an empty root keeps the environment lookup.
- app_rpc.go hands it the registry's state root through `hubAuthStateRoot`: the wired registry's `StateRoot()`, falling back to `cmdutil.DefaultStateRoot()` before the first successful load or when no registry is wired. The dead `cfg.HubStateRoot` fallback block is gone.

Production behaviour is unchanged: the registry root is `$XDG_STATE_HOME/evener` (else `~/.local/state/evener`), which is exactly what the env lookup produced. Fixtures that give a server its own registry root now get an auth controller on that same root by construction.

## Tests

- `TestHubRPCAuthStoresOAuthWhereTheRegistryReads` (app_auth_test.go): a server with `HubStateRoot` in one temp dir and a `newTestRegistry` in another. An OAuth record seeded under the registry root must sign `openai-codex` in over `evener/auth/status`, and `evener/auth/logout` must remove it from that root. Failed first with the controller looking under `HubStateRoot`:

  ```
  status={Provider:openai-codex Supported:true SignedIn:false ActiveSource:none ...}, want the record under the registry state root .../002 to sign the instance in
  ```

- `TestAuth_LoginComplete_WritesUnderTheConfiguredStateRoot`: constructor-level pin that a login lands under the root the controller is handed and nothing lands under the env root. Failed first on the write landing in the env dir (`LoadAuth(...): openai auth not found`).

## Gates

- `gofmt -l` on the changed files: clean.
- `go vet ./cmd/evener-hub` and `go vet -tags evenerfuzz ./cmd/evener-hub`: clean.
- `go build ./...`: ok.
- Full `GOMAXPROCS=4 go test -count=1 ./cmd/evener-hub`: ok (62s), zero FAIL lines.

Closes #991


obra:
Roborev Medium on c383f41 (OAuth root honored for writes but not for registry resolution) — fixed in c86c54f3d, and it changed the direction of this PR.

**Why the finding is right.** The registry reads `auth/<instance>.json` from its own state root, and the hub's registry is `cmdutil.LoadRegistry`, pinned to `cmdutil.DefaultStateRoot()` whatever `hub_state_root` says; the credential probe and every spawned child resolve the same way. So the controller has to write where the registry reads, not under `HubStateRoot`. Threading `hub_state_root` into the registry, the probe and every child instead would change what `hub_state_root` means (and children have no env var for a registry state root today), so I did not go that way.

**Change.** app_rpc.go now passes `hubAuthStateRoot(cfg.Registry)`: the wired registry's `StateRoot()`, else `cmdutil.DefaultStateRoot()`. Production output is unchanged (both are `$XDG_STATE_HOME/evener`). The dead `cfg.HubStateRoot` fallback block is deleted. PR title and body rewritten to match.

**Test that failed first.** `TestHubRPCAuthStoresOAuthWhereTheRegistryReads` puts `HubStateRoot` and the registry root in different temp dirs, seeds a record under the registry root, and drives `evener/auth/status` then `evener/auth/logout`. On the previous head it failed with `SignedIn:false ActiveSource:none ... want the record under the registry state root ... to sign the instance in`; it passes now, and logout removes the record from the registry root.

**Gates.** gofmt, `go vet` (both tags) and `go build ./...` clean; full `GOMAXPROCS=4 go test -count=1 ./cmd/evener-hub` ok (62s, zero FAIL lines). Branch merged with main (60b8425).

Noticed, left alone: `HubSpawner.StateRoot` (spawn.go) is never read in non-test code.

PR #1013, opened 2026-09-08
fix(lint): repair main's tagliatelle and Biome import-sort findings

obra:
Repairs main-branch lint breakage from #998 (commit f45c66402) so CI lint gates pass again.

**tagliatelle (golangci):** `streamDeltaChunk` in cmd/evener-tui/hub_notifications.go (~:609-612) uses camelCase json tags (threadId/turnId/itemId/callId) mirroring the appwire delta params wire spelling (cmd/evener-hub/app_relay.go). Marked the four tags `nolint:tagliatelle` like the repo's other wire-mirroring tags (e.g. agent/doctor/audit.go, agent/plugin/hooks.go) instead of renaming the wire.

**Biome organize-imports:** import-sort repair in the five files main's w2-barrel direct-import commit left unsorted (DocPane.tsx, AdvancedOptions.tsx, Rail.tsx, RailRow.tsx, SessionMenu.tsx).

Unblocks lint/static failures on PRs that touch none of these files:
- #1010
- #1000
- #1005
- #967

### DIFF
--- a/cmd/evener-hub/frontend/src/panes/doc/DocPane.tsx
+++ b/cmd/evener-hub/frontend/src/panes/doc/DocPane.tsx
@@ -9,8 +9,8 @@ import {
 } from "../../protocol/docContent";
 import type { PaneProps } from "../../shell/paneRegistry";
 import { Chip, Dialog, EmptyState, PaneScaffold, Skeleton } from "../../widgets";
-import { Markdown } from "../../widgets/markdown";
 import { requireClass } from "../../widgets/internal/requireClass";
+import { Markdown } from "../../widgets/markdown";
 import { filenameOf, formatDocBytes, isMarkdownPath } from "./docFile";
 import styles from "./docpane.module.css";
 import type { DocParams } from "./openDoc";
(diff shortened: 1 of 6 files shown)


obra:
Closing: both halves of this landed in #1008 (merged as be18ad663), and this branch now conflicts with main.

The Go half was fixed differently there, and the difference matters. `streamDeltaChunk` was a hand-rolled duplicate of `appwire.ToolOutputDeltaParams` — same fields, same tags — so the camelCase tags were a symptom of a protocol shape copied into a terminal-UI file rather than something to exempt. #1008 deletes the struct and decodes into the appwire type, which removes the tags, the suppression, and the duplication together. `appwire/doc.go` now states the contract as well: the tag strings are emitted verbatim as the frontend's TypeScript property names, so a tag is both the wire name and the client's property name, and callers should decode into these types rather than redeclaring their shapes.

The frontend half is identical work: the same five files through `npm run check`.

Useful thing from this PR: it names the other PRs the breakage was blocking — #1010, #1000, #1005, #967. Those are unblocked now and need a CI rerun.

PR #1022, opened 2026-09-08
cleanup(hub): resolve the state root once, in cmdutil.DefaultStateRoot

### LINKED ISSUE #1012: Two implementations of the evener state root (cmdutil.DefaultStateRoot vs the hub auth controller's openAIStateDirFromLookup) diverge on Windows and in their fallback
Noticed while landing #991 (PR #1011).

The evener state root is resolved by two separate implementations: `cmdutil.DefaultStateRoot()` (cmdutil/statedir.go) and `openAIStateDirFromLookup` in cmd/evener-hub/app_auth.go, which the hub's auth controller uses when it is handed no root. They agree on POSIX (`$XDG_STATE_HOME/evener`, else `~/.local/state/evener`) but diverge on Windows (`USERPROFILE` / `HOMEDRIVE`+`HOMEPATH` handling) and in their last-resort fallback when no home can be resolved (`.` versus `os.TempDir()`).

Nothing depends on the difference today, and #1011 made the hub pass its configured `HubStateRoot` through so the auth controller's own lookup is only a fallback. One implementation should remain: have the auth fallback call `cmdutil.DefaultStateRoot()` (or delete the fallback if every caller can pass a root), and pin the fallback with a test that sets the env and asserts the resolved path.

7487:
Fixes #1012

`openAIStateDirFromLookup` resolved evener's state root itself and had drifted from `cmdutil.DefaultStateRoot` in the two places the issue names:

- **Windows.** It read `USERPROFILE`, then `HOMEDRIVE`+`HOMEPATH`, out of the env map it was handed. `cmdutil.DefaultStateRoot` calls `os.UserHomeDir`, which reads `USERPROFILE` only — so a machine with `HOMEDRIVE`+`HOMEPATH` but no `USERPROFILE` put the hub's OAuth records under a different root than the rest of evener, and one with a msys-style `HOME` was ignored by the hub but not by `os.UserHomeDir` on POSIX.
- **No home at all.** The hub landed in `os.TempDir()/.local/state/evener`; `cmdutil.DefaultStateRoot` returns `./.local/state/evener`.

The one thing `cmdutil.DefaultStateRoot()` genuinely can't do is read `XDG_STATE_HOME` out of a *supplied* launch env rather than the process env, so that arm stays; everything below it now delegates. `openAIStateDirFromEnvList` went with the rest — nothing called it.

Two tests pin both arms, per the issue: the launch-env override against a process env that says something else, and the fallback against `cmdutil.DefaultStateRoot()` itself (both with `XDG_STATE_HOME` set and with it empty), so the two can't drift apart again. Reinstating the old `os.TempDir()` fallback fails the second one.

Two notes:

- The removed `TestOpenAIStateDirFromEnvDoesNotFallBackToProcessEnv` pinned the divergence directly: with an empty env map the hub resolved to `os.TempDir()` rather than reading the process env. That map is never empty in production — `effectiveHubAuthEnv` starts from `os.Environ()` — so this is the intended behavior change, not a lost guarantee.
- I deliberately left the `openAIStateDirFromEnv` one-liner in `app_auth.go` alone, and touched no other line in that file, to keep this off the lines #1011 is rewriting.

`go vet` and `go test ./cmd/evener-hub/... ./cmdutil/...` pass, seed corpora included. I couldn't run `make lint`: the repo pins golangci-lint 2.13.1 and I only had a v1 binary, so the v2 config wouldn't load — `gofmt` is clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


7487:
Fixed the lint failure and merged main in.

`lint-golangci` was tripping on a line of mine: gocritic's `filepathJoin` rejects a separator inside a `filepath.Join` argument, and the new fuzz case had `filepath.Join("/state", "evener")`. It takes its state home from `t.TempDir()` now, like the `app_auth_test.go` cases do — the surrounding spaces stay, they are what pins the `TrimSpace`. (The `static` failure was just the aggregate of that job; `static-build`, `lint-repository` and `fuzz-checks` all passed.)

I have the pinned golangci-lint 2.13.1 this time, so that supersedes the note in the description: `golangci-lint run ./cmd/evener-hub/...` reproduces the finding on the old line and reports 0 issues on the new one. `go vet` and `go test ./cmd/evener-hub/... ./cmdutil/...` still pass on the merged tree.

obra:
Addressed the roborev finding (medium): the launch env's home override.

The fallback now resolves the whole chain from the launch environment instead of the process environment. `cmdutil.StateRootFromLookup(goos, lookup)` owns the chain — `XDG_STATE_HOME` (trimmed), the home directory in `os.UserHomeDir`'s per-OS spellings (`USERPROFILE`, then `HOMEDRIVE`+`HOMEPATH` on Windows; `HOME` elsewhere), and the `"."`-rooted fallback. `cmdutil.DefaultStateRoot` delegates to it with the process env, and `openAIStateDirFromEnvMap` forwards its launch env. The hub keeps no home logic of its own, so there is still one implementation (#1012).

Coverage: launch-only home overrides on Unix through the hub (`TestOpenAIStateDirFromEnvUsesLaunchEnvHome`), and the Windows spellings from any host in `cmdutil` (`TestStateRootFromLookup`). An empty launch env now resolves to the `"."`-rooted fallback and never reads the process env, restoring the process-independence the removed test pinned.

Commit `80ae394d02` (`fix(hub): resolve the launch env's home in the state-root chain`), on top of current `main` (`daaaea7974`). `go vet`, pinned golangci-lint 2.13.1, and `go test ./cmd/evener-hub/... ./cmdutil/...` pass locally.

obra:
Addressed the second roborev pass (4 findings) in `b31d0be256`, then merged current `main` (`bb4d7a70e0`, #1175) in `0251ed5543`.

- **Medium, `cmdutil/statedir.go` Windows home** — the finding's premise is inverted. `go1.27.0`'s `os.UserHomeDir` has no `HOMEDRIVE`/`HOMEPATH` arm at all (neither string appears anywhere in `$GOROOT/src/os`); it reads `USERPROFILE` only on Windows. My first pass had *added* a `HOMEDRIVE`/`HOMEPATH` arm, which re-introduced the exact Windows drift #1012 removed. Removed it: Windows consults `USERPROFILE` only, matching `os.UserHomeDir`. The suggested "concatenate `HOMEDRIVE`+`HOMEPATH` first" change would have diverged further from the stdlib. Coverage added for full and partial `HOMEDRIVE`/`HOMEPATH` combinations, plus the `android`/`ios` constants `os.UserHomeDir` returns.
- **Medium, `TestOpenAIStateDirFromEnvUsesLaunchEnvHome` portability** — valid. The hub again exposes a `goos`-parameterized resolver (`openAIStateDirFromLookup`), and the test now pins the Unix and Windows arms from any host instead of setting only `HOME`.
- **Low, XDG trimming** — valid. `rendezvous.DefaultDir` documents itself as mirroring `DefaultStateRoot` and reads `XDG_STATE_HOME` raw (as do `appwire`, `agent/runtime_dir`, `cmd/evener-hub/config.go`). Preserved the previous untrimmed semantics in the shared resolver; the hub forwards its env verbatim. A test pins the verbatim behavior.
- **Low, `TestOpenAIStateDirFromEnvFallsBackWhenNoHome` isolation** — valid. Now pins `XDG_STATE_HOME`, `HOME`, and `USERPROFILE` to temp dirs before reading the process root.

Local: `go vet`, pinned golangci-lint 2.13.1, and `go test ./cmd/evener-hub/... ./cmdutil/...` pass on `0251ed5543`.

obra:
Third roborev pass: one low fixed, two mediums rebutted as out of scope.

- **Low, fuzz seeds** — valid. Fixed in `3ba2a00f43`: the `home` seed now sets the host's home variable and the redundant `temp` seed is renamed `nohome` (it covers the empty-env fallback), so the home arm has fuzz coverage again.
- **Medium, `XDG_STATE_HOME` trim** — out of scope and not on the hub's path. The divergence the finding names is between `cmdutil.DefaultStateRoot` and `auth/openai`'s separate `DefaultStateDirWithStateHome`, which trims. The hub never uses the latter: `SaveAuth`/`LoadAuth` take an explicit `stateDir`, and the hub passes its own cmdutil-resolved root. The only callers of `authopenai.DefaultStateDir()` are `cmd/evener/openai_login.go:255` and `llm/providers/tokenauth/codex.go:166`; neither goes through the hub. #1012 named exactly two implementations (cmdutil and the hub's `openAIStateDirFromLookup`), and this PR converges those two. Trimming `StateRootFromLookup` would re-diverge it from `rendezvous.DefaultDir`, `appwire`, and `agent/runtime_dir`, which read `XDG_STATE_HOME` raw and document themselves as mirroring `DefaultStateRoot`.
- **Medium, no-home fallback** — out of scope for the same reason. The `.` vs `os.TempDir()` split is pre-existing and is between cmdutil and `auth/openai`, not the hub; #1012 explicitly chose `.` for cmdutil, and the hub now inherits it. Aligning `auth/openai` would change `openai_login`/codex behavior and belongs in its own change.

Merged `main` (`2bbd26f1d0`, #1178) in `b6c60ad6fa`. Local `go vet`, pinned golangci-lint 2.13.1, and `go test ./cmd/evener-hub/... ./cmdutil/...` pass.

PR #420, opened 2026-08-24
fix(agent): bound interactive drain and hopeless close joins

obra:
## Serial dependency

This is the #382 follow-up stacked on PR #386. The exact #386 head was `7ccf8e8f2319174a1e70921fa5949e99ddf96976`; it was rebased onto current `origin/main` `b21db7bd78bb4cc53f7fa9f61dd0648ac62fd589` as local predecessor commit `6bc59539a45ee83b8e537ceb5972c5ef8f1bdce6`. PR #386 remains open. This PR intentionally carries that predecessor plus the follow-up; its GitHub diff will narrow automatically when #386 lands.

Depends on #386
Fixes #382

## Root cause

- `markDrainAbandonedDelegates` returned immediately for every `TurnEndsProcess=false` session, leaving an explicit `stop_pending` delegate unbounded in serve-mode drains.
- Close used the entire shared cascade deadline joining a stop-reconciliation driver before reaching generic bounded joins. A delegate parked in an uncancellable tool call therefore consumed the whole close budget, and teardown attempted to wait on the same wedge again.

## Implementation

- Bound explicit pending-stop delegates in both one-shot and serve sessions. The conservative drain-start fallback remains one-shot-only, so a never-stopped interactive delegate is not implicitly abandoned.
- Reserve half of the shared close budget for stop reconciliation. If that hopeless stop reaches its deadline, cancel the remaining delivery-only close budget rather than rejoining the same wedged child.
- Added a real lifecycle serve regression: `NewSession` → `createDelegate` → uncancellable `ReadFile` → `StopSubtree` → deterministic drain.
- Extended the real one-shot wedge test to measure the post-drain close interval.
- Kept `DrainStallTimeout` and `LaneClosePassBudget` as exported cross-module test seams. PR #386 documents why these process-wide policy knobs are not persisted session configuration; no production assignment exists.

## TDD evidence

RED:

- `go test ./agent -run '^TestServeDrainAbandonsARealStopPendingDelegate$' -count=1 -v -timeout=5s` — timed out while the drain remained in `waitDrainWake`.
- `go test ./cmd/evener -run '^TestRunExitsWhenADelegateIsWedgedInAnUncancellableToolCall$' -count=1 -v -timeout=30s` — failed with `Close spent 3.01728475s after the drain returned` under a 3s close budget.

## Verification

Passed:

- Focused agent lifecycle tests for serve abandonment, stop-requested abandonment, completed-stop preservation, and finished-run preservation.
[...]

PR #1029, opened 2026-09-08
fix(flake-wedged-close): pin hopeless-stop half-budget deterministically

obra:
## Flake

race-root CI on PR #954 (run 34196379603, job 101964919414, 2026-09-08) failed one assertion:

```
run_drain_wedged_delegate_test.go:221: Close spent 2.135622243s after the drain returned; it should not consume the entire 3s close budget joining the hopeless stop before bounded joins
```

## Verdicts on the 4 failed jobs (run 34196379603)

- **tests (101964919391) + race-modules/agent (101964919453)** — `TestNoBareWallClockDeadlineInAgentTests` flagged `session_jobtree_drain_backoff_test.go:321: bare wall-clock bound passed to time.After(...)`. NOT a flake and NOT from the PR (frontend-only diff): CI checks out the PR//main merge, and main's #967 (1d21d1dd8) had added that file with a bare bound hours earlier. Already fixed on main by #1020 (746227582). Fresh PR CI (run 34251773909) is all green. No action.
- **race-modules wrapper (101966845691)** — pure aggregation job (`echo RESULTS` over `race-module-shards` = failure); consequence of the agent-shard failure above, no independent failure. No action.
- **race-root (101964919414)** — genuine timing flake (this PR).

## Root cause

The 2s ceiling sat 0.5s above the 1.5s the hopeless stop join burns **by design**: `closeStopJoinContext` reserves `LaneClosePassBudget/2` (1.5s of 3s) for the stop and the other half for the bounded joins. Local measurement (probe, since removed): idle mean 1.51-1.52s, stdev ~10ms, both with and without `-race`, robust under synthetic CPU stress. CI needed only ~135ms of scheduling jitter over the 1.5s burn to trip the 2s ceiling (2.136s observed). A ceiling 0.5s above a 1.5s floor is a load-flake by construction.

## Fix (mechanism, not symptom)

- `agent/session_worktree_close_test.go`: new `TestCloseStopJoin_HopelessStopLeavesHalfBudget` pins the REAL regression (stop join consuming the joins' half, the #420 failure mode) deterministically — it reads the split off the two context deadlines with zero wall-clock wait, and goes red against a full-budget mutation (verified).
- `cmd/evener/run_drain_wedged_delegate_test.go`: e2e ceiling 2s -> full 3s budget (the true failure boundary: consuming the whole budget). The 90s TRIPWIRE hang guard is unchanged.

## Gates

- `go test ./agent/ -run 'TestCloseStopJoin_HopelessStopLeavesHalfBudget|TestNoBareWallClockDeadlineInAgentTests'` → ok
- `go test ./cmd/evener/ -run 'TestRunExitsWhenADelegateIsWedgedInAnUncancellableToolCall' -count=3` → ok
- `go test ./cmd/evener/ -race -run 'TestRunExits(Whed...|WithALive...)' -count=3` → ok (6/6)
- `go test ./cmd/evener/ -count=1` (full package) → ok 24.3s
- Pre-fix loops: wedge test 10x + 20x + 30x clean locally; flake reproduced only in CI (135ms over).

### DIFF
--- a/agent/session_worktree_close_test.go
+++ b/agent/session_worktree_close_test.go
@@ -21,6 +21,43 @@ import (
 // tools spec §9 steps 4-6, §5 close-unlock). They build on the wtRepo harness
 // from session_tools_worktree_create_test.go.
 
+// TestCloseStopJoin_HopelessStopLeavesHalfBudget verifies the close-budget
+// split closeStopJoinContext exists to enforce: a delegate-tree stop parked
+// forever behind an uncancellable tool call must consume at most half of the
+// close cascade's budget, so the bounded joins and teardown that follow the
+// stop still have a full half left. This is the unit pin for the
+// TestRunExitsWhenADelegateIsWedgedInAnUncancellableToolCall flake
+// (2026-09-08, race-root CI: Close spent 2.136s of a 3s budget after the
+// drain returned, tripping that test's 2s ceiling): the stop join is
+// SUPPOSED to burn ~budget/2, and the ceiling above it must clear that burn
+// with margin for close's own scheduling. If this test ever goes red, the
+// stop join stopped honouring its half — do not "fix" it by widening the
+// ceiling in cmd/evener/run_drain_wedged_delegate_test.go.
+func TestCloseStopJoin_HopelessStopLeavesHalfBudget(t *testing.T) {
+	shortenCloseCascadeBudget(t, 3*time.Second)
+	// The initiating close mints the cascade budget from Background, exactly
+	// as Session.Close does.
+	cascade, cancelCascade := ensureCloseBudget(context.Background())
+	defer cancelCascade()
+	cascadeDeadline, ok := cascade.Deadline()
+	if !ok {
+		t.Fatal("ensureCloseBudget minted no deadline: the cascade budget is gone")
+	}
+	stopCtx, cancelStop := closeStopJoinContext(cascade)
+	defer cancelStop()
+	stopDeadline, ok := stopCtx.Deadline()
+	if !ok {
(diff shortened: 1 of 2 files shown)


obra:
RoboRev Medium follow-up (24bd69344): tightened both tests as requested.

**Unit — `TestCloseStopJoin_HopelessStopLeavesHalfBudget` (`agent/session_worktree_close_test.go`)**: was `reserved >= 1s` on a 3s cascade (a 2s-burning stop join still passed). Now asserts `reserved == LaneClosePassBudget/2` within 250ms. The 250ms is honest, not generous: both deadlines are read off contexts minted back-to-back from `time.Now` with zero wall-clock wait, so any skew beyond a few ms means the split moved. Red-checked: a temporary 2s-burn mutation fails with `reserves 999.9ms … want 1.5s within 250ms`; restored after. 5/5 green.

**Integration — `TestRunExitsWhenADelegateIsWedgedInAnUncancellableToolCall` (`cmd/evener/run_drain_wedged_delegate_test.go`)**: ceiling was the full 3s budget (`elapsed >= 3s`), which a stop join consuming the joins' half would exactly meet. Now `LaneClosePassBudget - 500ms` (2.5s on the shrunk 3s budget): ~365ms above the worst observed healthy close (2.136s flake) so it won't re-flake, ~500ms below the full-budget burn so the #420 shape still trips it. This drives the real path — `run() -> Session.Close() -> closeOwnedDelegateRuntimeTree -> closeRuntimeTree -> closeStopJoinContext` — so a future bypass of the helper lands at ~3s and fails here. 3/3 green (~3s each). gofmt/vet clean. Test-only changes; no production code touched.

PR #1035, opened 2026-09-08
feat(goal): wait registry with parked evaluation, ledger, graduation, watchdog

obra:
Implements the goal-wait redesign spec (docs/superpowers/specs/2026-09-07-goal-wait-redesign-design.md) per docs/superpowers/plans/2026-09-07-goal-wait-impl.md: wait registry + parked gate (slice 1), progress ledger + backstop (slice 2), graduation + watchdog + resume + command surface (slice 3).

Review history: 8 spec rounds (46 Important + 70 Minor closed, zero DQs) plus two implementation review waves (7 findings adjudicated: 3 fixed, 1 pinned by test after refutation, 3 deferred and noted below).

Verification: full -race gate green on 7063f3b0a (1377s); broad goal scope re-green after simplify (40s).

Deferred (documented, not in this PR): cooldown over-arming on poll-leg fires (needs spec/domain check); http_match fetch-vs-validation (pre-existing slice boundary); approval restore-loss vs fire-on-missing (spec §7 vs §2 conflict, needs decision).


obra:
roborev wave pushed as dcb5d13e6 — 6 fix-now items done, 4 deferred with reasons:

Fixed: M3 delivered-map copy in gate (was alias-then-read off-lock); M4 conditions restore TimeoutNanos (+ round-trip pin); L1 RFC3339Nano baselines; L2 NearestWait returns (value, bool) with callers updated; H2 timer claims the synthetic deadline wake (+ TestFixWaveTimerDeadlineKicksFinalTurn); H3 maxParkedTotal crossing accrues the open stretch + projects from ParkedTotalAt (+ TestFixWaveTimerParkedCrossingBinds, proven to fail without the accrual).

Deferred: H1 fetch/matcher evaluation + full CIDR/DNS egress, H4 stable ask generations, M1 sandbox containment, M2 external_label notify path — each is a slice boundary needing spec + design (http was validation-only since slice 1, generations were never minted, containment/notification paths don't exist yet), not review-wave fixes. Happy to scope follow-ups.

obra:
Round-2 wave pushed as 944c196be — 4 of 6 fixed, 2 reconfirmed deferred:

Fixed: standing-entry duplicate (ClaimDeadlineExpiry now settles marker/status/anchor when reporting a standing entry — each check observes settled state); phantom accrual (accrual gate re-reads status post-claim, re-anchors only while still waiting); gate bounds (both sites read ParkedTotalAt via boundsBreachedAt); restored-strand (settle kicks restored-but-never-kicked backlog exactly once + TestGoalRestore_SettleKicksRestoredBacklog — verified the strand first: settle returned false/kicks=0 pre-fix).

Reconfirmed deferred: http matcher/fetch, ask generations — same slice-boundary reasons as round 1 (no production fetch path exists; no generation is minted anywhere). Happy to scope follow-ups.

obra:
Round-3 wave pushed as 23cd51afa — 8 of 13 fixed, 5 deferred (3 repeats + 2 scope calls):

Fixed: parent until_child forward kicks via kickClaimedGoalWake (+ TestGoalChildForwardParentKickDelivers); VerifyConditions + delta flips positional over duplicate Desc (+ TestVerifyConditionsDuplicateDescPositional); until_child registration catches up on already-terminal children (+ TestGoalChildRegisterTerminalCatchesUp); armGoalWaitTimer serialized under goalUpdateMu; resume-with-replacement suppresses the stale kick (+ TestGoalResumeFromWireSuppressesKickWithReplacementText). Plus round-2 leftovers already in: standing-entry settle, post-claim accrual gate, live gate bounds, restore kick pin.

Deferred: http fetch/matcher, ask generations, sandbox containment, external_label notify (repeats — slice boundaries needing spec+design); update_goal vs goal_expect atomicity (narrow TOCTOU, single-threaded drain in practice — proposing a follow-up with a VerifyAndComplete design rather than rushing the lock change here).

obra:
Leftovers from round 3 pushed as d5be77d2b (2 items): parked-cap timer kick (kickGoalWaitBudgetBound + kick-count pin in TestFixWaveTimerParkedCrossingBinds — the crossing now schedules the rule-2 evaluation instead of disarming into a strand); deadline-claim invariant narrowed with reader audit (waiting+live-leases only on the final-turn claim; the following gate clears them at block). Verification: broad goal scope + store/scope green, gofmt/vet clean.

obra:
Round-4 wave pushed as 23e9ea9eb — 4 fixed, 5 reconfirmed deferred:

Fixed: High deadlock (delivered-set snapshot hoisted out of goalUpdateMu — established goalUpdateMu-then-s.mu order restored; race-scope green); parked bound-at-arm kick (arm reports bound-at-arm, kickGoalWaitBudgetBound outside locks; caught+fixed a re-arm recursion stack overflow in verification); durable child-session-id match (+ fast TestGoalChildDurableMatchesSessionID); observation classifier scoped to control-plane callers (content tools always ok).

Reconfirmed deferred: http fetch/matcher, egress CIDR/DNS, ask generations, external_label notify, sandbox containment — same slice-boundary reasons (no fetch path / no minting / no containment infra exists yet). Happy to scope follow-ups.

obra:
Round-5 wave pushed as eaad0323e — 3 fixed, 3 reconfirmed deferred:

Fixed: --extend overflow (bound before int/Duration conversion + TestApplyExtendClampsExtremeValues with 2^62 values); TakeLossCause atomic consume (clears cause+flag together — same-turn MarkAdvanced re-arms after, ordering documented); harness-auto doc corrected to model-declared-only in both tool surfaces (spec design doc left frozen).

Reconfirmed deferred: sandbox containment, http fetch/matcher, ask generations — third repeat, same slice-boundary reasons (no containment infra / fetch path / minting exists).

obra:
CI failure on eaad0323e triaged to a real race (not a flake): kickClaimedGoalWake kicked the whole PendingWake backlog instead of its claimed batch, so a concurrent deadline synthetic landing between claim and kick got marked+announced twice (CI evidence: TestGoalChildForwardSiblingWakeDelivers showed wait_1 + deadline in one backlog). Fixed in 34f30fae6: the kick intersects claimed ids with the re-read backlog (batchForIDs), marks/announces only those, and stands down when the batch was consumed mid-flight. Broad goal scope green locally.

obra:
Round-6 wave pushed as 2f62abf02 — 4 fixed, 3 reconfirmed deferred:

Fixed: Stage on all 3 GoalState projections (+ TestProject_GoalUpdatedCarriesStage); batch-scoped kick coverage (+ TestKickClaimedWakeKicksOnlyItsBatch — prompt/delivered/pending asserts); resume+retarget atomic (goalResumeWithRetarget under one hold; serve no longer retargets after — old-objective kick impossible by construction + rewritten wire test); restored-strand (already fixed in 944c196be via settle kick + TestGoalRestore_SettleKicksRestoredBacklog — reconfirmed present and green).

Reconfirmed deferred: http fetch/matcher, sandbox containment, ask generations — fourth repeat, same slice-boundary reasons.

obra:
Round-7 wave pushed as c5e6db1b7 — 2 fixed, 6 deferred/triaged:

Fixed: observation novelty (canonicalize BEFORE hashing — the sha256 hex digest was redacted to <id>, collapsing all outputs; CanonicalizeObservationOutput + TestLedgerCanonicalizedOutputsStayDistinct); ClaimChildWaits lock (holds s.mu across match+claim via claimFireLocked split).

Triaged: restored-strand (repeat — settle kick landed in 944c196be, reconfirmed green); retention window (no window is defined anywhere in code/spec — the job store is append-only with no TTL; asking roborev to name the bound or accept file+defer); http/sandbox/generations/external_label (repeats — slice boundaries).

obra:
Current-findings wave pushed as 3b9f4af87 — all 3 verified and fixed:

1. Extend clamp arithmetic (High, confirmed: int64((Cap/Second).Seconds())==0, granting full 24h for ANY positive value — reproduced numerically). Fixed to int64(Cap/time.Second); ordinary-value assertions added (1h deadline passes through).
2. Novelty digest redaction (High, confirmed: sha256 hex redacted to <id>, all entries identical). FoldLedger now compares stored digests directly; redaction lives only pre-hash.
3. Egress 172.16 gap (Medium, confirmed): host parsed via net/url+net.ParseIP, deny by IsLoopback/LinkLocal/Private/Unspecified (+ TestGoalSubstrateCheckURLDeniesPrivateRanges incl. path-substring false-positive pins). Full fetch/matcher remains the deferred slice.

obra:
Egress follow-up pushed as be74372a3 — finding confirmed and fixed: non-canonical numeric hosts (2130706433, 0x7f.0.0.1, 0177.0.0.1, 172.016.0.1, 0xac.0x10.0.1) fell through ParseIP to a hostname substring list that missed them, and the 172.2 fragment contradicted the parsed-host claim. CheckURL now parses strictly (netip range deny, legacy-numeric deny mirroring the mobile pairing gate, DNS-name validation, localhost/trailing-dot handling); substring fragments gone; resolution-time re-check stays deferred to the fetch leg. New cases proven to fail on old code, pass on new; goal scope + vet clean.

obra:
QF1001 fix pushed as d6c654b10: extracted `isDNSLabelChar` helper in `agent/session_goal_substrate.go` (switch form, behavior unchanged). Local `golangci-lint run ./agent/...` is now clean except 3 pre-existing modernize findings in `agent/retained_output_size_acc_test.go` (rangeint x2 + embedlit) — that file is untouched by this branch (`git diff main...HEAD` empty), so left as-is.

obra:
Round-9 wave pushed as be02c3f7f — 3 findings fixed, 1 needs your call, plus a CI tests-lane failure fixed:

Fixed:
1. Medium userinfo (confirmed: no u.User check anywhere) — ValidHTTPURL rejects userinfo; CheckURL inherits via its gate. Pin: TestValidHTTPURLRejectsUserinfo + egress-test userinfo cases.
2. Low CheckURL timeout (confirmed: _ = timeout) — rejects non-positive (fail-closed); no upper clamp, since the arg is dual-use (lease TTL at RegisterWait validation, 15s per-fetch default at the poll leg). Pin: timeout cases in the egress test.
3. Low LookupJob lock (confirmed) — releases jm.mu across the jobstore Load, re-verifies liveness after, nil-store guard (preserves old fail-closed on Load error).
4. CI tests failure on d6c654b10 (TestGoalChildForwardSiblingWakeDelivers, extra deadline-exceeded wake): the synthetic waiter runs a real clock while the test drives fake time; the forward's async notification drive settled with real now past the fake-anchored deadline and claimed the synthetic wake. Race, introduced by my dcb5d13e6 timer/settle deadline-claim. Fixed by pinning the waiter to the fake clock (200+ local passes missed it — the assertion usually wins the race).

Needs your decision:
- High http_match-fires-without-fetch: confirmed behavior (truth + attach-scan fire on reachability; production CheckURL is validation-only by design). But every remedy contradicts a pinned contract: parking-until-fetch breaks TestFixWaveC1AllKindsFire (http_match-at-poll-leg) + tool text (polled 60s floor) + spec §2 eval policy; rejecting at registration breaks wait_test.go:204 (well-formed URL must park); a real fetch+matcher leg is the deferred slice the substrate comment and goal_expect v1 restriction already name out of scope (and duplicates 3 prior deferred http-fetch findings). Say the word on which direction and I'll implement it.

obra:
Round-10 wave pushed as fad3e394d — per your calls (1 yes, 2 defer, 3d):

1. High sandbox containment (fixed): StatFile joined nothing and stated the raw model path, so absolute outside paths and symlink escapes were accepted. It now resolves relative paths against the session working directory and gates through the env RootBoundary (EnsureUnderRoot — the same symlink-aware check resolveWrite enforces) before stating, fail-closed on escape. One gate covers registration, poll evaluation, restore attach-scan, and claim-time re-evaluation. Pin: TestGoalSubstrateStatFileSandboxContainment (outside absolute, traversal, symlink-out deny; inside absolute/relative/symlink-in allow). Broad goal scope + internal/goal green, lint clean (only the 3 pre-existing untouched retained_output findings).
2. Medium stranded-restore (deferred): the stranded-backlog path is already handled — settleGoalOnIdle's restoredBacklog branch (spec §7 kick-immediately) kicks restored-but-never-kicked backlog at the first turn tail, and SetNotifyFunc flushes pending work at attach; kicking at SetKickFunc time would double-kick past the settle's exactly-once accounting. Happy to add an end-to-end restore→bridge→settle pin if you want one.
3. http_match without fetch (deferred, your 3d): left as-is per your call.

obra:
Simplify pass on the recent waves pushed as c222a8308 (behavior unchanged — wide goal scope + internal/goal green, vet + lint clean): fixed StatFile double-currentEnv() (single snapshot, no TOCTOU skew, skip redundant Clean after Join), CheckURL timeout floor before the parse, isDNSLabelChar switch → single predicate, LookupJob excerpt built before the re-verify lock. Skipped: url.Parse sharing (crosses into goal.go evaluator + expect.go paths, outside the diff), StatFile/resolveShellWorkingDir preamble extraction (no drop-in helper — shell variant carries Stat+IsDir gating StatFile must not have; shared EnsureUnderRoot boundary already reused). Altitude review: all 9 depth items at the right level, no moves.

obra:
Round-11 wave pushed as f490fa81a (kept simple per your note — smaller diff than the review suggested, not larger):

Fixed (simplest form I could find):
1. Low stale excerpt — instead of rebuilding the string after the lock, LookupJob now delegates to a new jobManager.goalWaitJobStatus (sits next to terminalWatchTargetStatus in job_watch.go): store read outside jm.mu, liveness re-verified under lock, excerpt from the re-verified status. The window is gone by construction.
2. StatFile fail-open without boundary — containment is now a working-directory prefix check (EvalSymlinks-resolved) plus the RootBoundary assertion where offered. Fails closed with no boundary, no fake env, no symlink-resolution duplication.
3. Medium restored-wake strand — confirmed real in the narrow case (restored backlog + zero live waits + no other pending work: timer disarms, settle never runs). Fix mirrors the existing SetNotifyFunc flush: SetKickFunc settles immediately when an undelivered backlog exists (at most once via the delivered set). Pin: TestGoalRestore_KickWiringFlushesRestoredBacklog. The old settle test now asserts the new contract (kick at wiring, settle suppresses the repeat) — its exactly-once purpose is unchanged.

Refuted (no change, evidence in comment): the settle↔retarget deadlock — lock trace shows settleGoalOnIdle never nests s.mu inside goalUpdateMu (delivered snapshot split into two sections; later re-reads also split); SetGoal's goalUpdateMu→s.mu order is unopposed, so no cycle exists.

Still need your call (as before): http_match fetch (3d stands), external_label path, ask_generation field. All green locally: wide goal scope + internal/goal, vet, lint (zero hits on touched files).

obra:
http_match removal pushed as 3986ad2de (per your decision; tracked in #1061): registration rejects at goal_wait validation, RegisterWait, and goal_expect validation — all naming issue #1061; persisted pre-removal leases park and conditions read unsatisfied, never fire. Removed: Substrate.CheckURL + production egress gate, ValidHTTPURL + DNS/legacy-numeric helpers, waitHTTPTimeout, http_match tool docs. Kept: EventHTTPMatch enum (wire compat), Matcher field + caps + persist (shared payload), schema enums. Net -225 lines. Local: build + vet clean, goal/tool packages + wide goal surface green, lint 0 issues on changed packages.

obra:
external_label removal pushed as a261fb3c1 (per your decision; tracked in #1063): registration rejects at goal_wait validation, RegisterWait, and goal_expect validation — all naming issue #1063; persisted pre-removal leases park and conditions read unsatisfied, never fire. Removed the substrate-free registration branch + tool docs; Max8 cap test retargeted to until_time slots. Kept: EventExternalLabel enum (wire compat), schema enums. Local: build + vet clean, goal/tool packages + wide goal surface green. (http_match removal 3986ad2de CI was green; this commit's run pending.)

obra:
Round-12 wave pushed as 37c826d17 — 4 fixed, 1 refuted with evidence, 1 needs your call (ask_generation, next question whenever you want it):

Fixed:
1. Prefix-sniffed expiry (confirmed, incl. the reverse direction the reviewer missed): claimedPredicateFire now keys on a structural PendingWake.Expiry mark threaded from WaitClassification through claimFireClassified, plus the synthetic deadline id — never trigger text. Expiry persists (omitempty: old snapshots behave as before). Pin: TestFixWaveExpiryFireDoesNotAdvance (lease-expiry, deadline wake, predicate fire, prefix-mimic text).
2. File duplicate leases (confirmed): baseline snapshots into the predicate BEFORE the idempotency key (it is predicate identity) — identical re-registers dedupe; the post-hoc key recompute is gone.
3. Target docs still say URL (confirmed): dropped from both target descriptions.
4. Dead matcher in the key (confirmed): excluded from canonicalPredicate (still capped + persisted as reserved). Pin: TestRegisterWaitMatcherExcludedFromIdempotency.

Refuted (no change): NextWaitID persistence — maxNextWaitID already scans waits + backlog + persisted counter, and every tool round autosaves, so a registered wait reaches disk at turn end; the wire-snapshot addition fixes no live scenario. If you can name one (a save path that skips a registration), I'll reopen it.

Local: wide goal surface + internal/goal + internal/tool green, vet + lint clean (only the 3 pre-existing untouched findings).

obra:
Round-13 wave pushed as a2eb2575e — 5 fixed (4 Medium + 1 Low):

1. Immediate fires leave waiting state (confirmed): both RegisterWait immediate paths assigned kept without the StatusActive transition ClaimFire has. Fix threads the same hasLiveWait guard into the catch-up and attach-scan paths. Pin: TestRegisterWaitCatchUpReplacesLastLiveWaitReturnsActive + TestRegisterWaitAttachScanReplacesLastLiveWaitReturnsActive.
2. Restored terminal child, no wake (confirmed): childTerminalTrigger searched only tracked subagents while LookupChild accepts durable records. Fix extends the trigger to PhaseClosed durable records (Descriptor.ChildSessionID, outcome+reason), pre-read discipline kept. Pin: TestGoalChildDurableTerminalCatchesUp.
3. Legacy expiry misclassified on restore (confirmed): pre-upgrade entries persist expiry unset + legacy trigger text, restoring Expiry:false and laundering advancement. Fix backfills at goalRestoreToStore (prefix + non-deadline-id). Pin: TestGoalPersist_LegacyExpiryBackfill.
4. RegisterWait lock inversion (confirmed in the narrow StatFile→Session.mu leg; other lookups take only manager/controller locks under the documented goalUpdateMu→s.mu order): baseline StatFile hoisted to a pre-read outside the store lock; validation-switch reads keep the established order. Full goal surface green incl. -race on touched areas.
5. Cooldown keyed by bare target (confirmed): keyed by (kind, event-subtype, target) at all 4 sites incl. CheckCooldown signature. Pin: TestCooldownKeyedByKindAndSubtype.

M4 note: full RegisterWait-outside-locks would need re-entrancy the store cannot offer; the residual follows the same order the gate already relies on. Local: build + vet clean, internal/goal green (+race), wide goal surface green, session touched areas -race green, lint 0 on changed packages (only the 3 pre-existing untouched findings).

obra:
ask_generation removal pushed as a1a9f736c (per your call): goal_wait/goal_expect reject a non-empty ask_generation naming the removal; approval waits bind by content key alone. Pin: TestGoalWaitToolAskGenerationRemoved. Local: build + vet clean, internal/goal + internal/tool + wide goal surface green, lint 0 on changed packages (only the 3 pre-existing untouched findings). (Round-13 a2eb2575e CI run pending separately.)

obra:
Round-14 wave pushed as f4b371b1c — 5 fixed (1 High + 2 Medium + 2 Low):

1. HIGH lock-order deadlock (confirmed in the narrowed scope): RegisterWait/RegisterExpect ran LookupApproval/StatFile (both take Session.mu) under Store.mu while SetGoal/ClearGoal take s.mu across store mutations. Fix hoists ALL substrate reads to a pre-pass snapshot outside the lock (registerPrescan/prescanExpect + replay, ClassifyWaits precedent); locked section performs zero substrate calls. TOCTOU documented point-in-time (timer/gate re-validates; stale-park self-corrects, stale-reject retryable). Pins: TestRegisterWaitSubstrateCallsOutsideStoreLock + TestRegisterExpectSubstrateCallsOutsideStoreLock (probe substrate asserts via InRegisterCritical test seam) + TestRegisterWaitSnapshotSemantics (flapping substrate decides coherently).
2. MEDIUM cap off-by-one (no miscount; real gap: post-commit render past the cap): plain-drive + auto-park-consumed fold sites now check boundsBreachedAt on the committed snapshot and block VerdictBudgetExhausted instead of rendering. Count semantics unchanged. Pin: TestGoalContinuationCapHitBlocksInsteadOfRendering (Max=Used+1, progressing outcome, asserts block + exactly one budget-exhausted note).
3. MEDIUM approval-key encoding (confirmed, NUL-join undocumented): both tool definitions document the two-half header\\x00question key + bare-half rule; semantics unchanged. Pins: TestDefGoalWaitApprovalKeyEncoding + TestGoalWaitApprovalKeyPairContract (real goalSessionSubstrate: two-half parks on headed ask, bare rejects on headed, bare parks on lone-half).
4. LOW stale enum (confirmed): event_subtype enums shrink to [file_modified] in both schemas; description text keeps the #1061/#1063 removal notes; validation still rejects. Shape tests pin enum + notes.
5. LOW AskGeneration dead state (confirmed): legacy-only docs, removed from canonicalPredicate key, store rejection names the content key only (field + persist compat kept). Pins: TestAskGenerationExcludedFromIdempotency + updated generation-binding message assertion.

Local: build + vet clean, internal/goal + internal/tool green, wide goal surface green, FULL agent suite green (286s), -race on store + session touched areas green, lint 0 on changed packages (only the 3 pre-existing untouched findings; one self-introduced singleCaseSwitch fixed pre-commit).

obra:
Round-15 wave pushed as e743cca58 — 6 fixed (1 High + 1 Medium + 4 Low):

1. HIGH settle lock inversion (confirmed): the restoredBacklog branch took goalUpdateMu inside the s.mu section. Fix hoists the wake re-read into the pre-s.mu goalUpdateMu split (same shape as the neighboring backlogPending split); the s.mu section consumes locals only. Pins: TestFixWave15_SettleRestoredBacklogWakeParity + TestFixWave15_SettleRestoredBacklogSettleOnly (identical wake prompt/ids through the settle path).
2. MEDIUM kick map race (confirmed): SetKickFunc iterated the live delivered map after unlock. Fix copies under the lock (backlogPending-split shape). Pins: TestFixWave15_SetKickFuncFlushCopySemantics + TestFixWave15_SetKickFuncFlushVsMarkRace (-race green).
3. LOW watchdog same-text retarget (partially confirmed): state keyed by objective text only, so a same-text Set (fresh CreatedAt, cleared waits/ledger) reused stretch+notices+ceiling. Fix keys on (objective, CreatedAt); the per-goal 24h ceiling resets with the goal per its own field comment. Pins: same-text retarget resets stretch + ceiling, same-stretch keeps state.
4. LOW NearestWait tie-break (confirmed): lexicographic wait_id misordered wait_10 < wait_2. Fix orders by (RegisteredAt, numeric wait_N suffix, id fallback). Pin: TestFixWave15_NearestWaitNumericTieBreak.
5. LOW scalar byte truncation (confirmed): s[:512] could split UTF-8. Fix truncates on runes. Pin: TestFixWave15_GoalArgScalarRuneSafeTruncation.
6. LOW expectKindRejected fallthrough (confirmed): job/delegate fell to unknown-kind. Fix adds the explicit registrable case. Pin: TestFixWave15_ExpectKindRejectedRegistrableKinds.

Local: build + vet clean, internal/goal + schema green, wide goal surface green, -race on kick/settle/restore/watchdog + FixWave15 pins green, lint 0 on changed packages (only the 3 pre-existing untouched findings; one self-introduced ineffassign fixed pre-commit).

obra:
Round-16 wave pushed as 7e267181e — 4 fixed (1 Medium + 3 Low):

1. MEDIUM NextWaitID reuse (confirmed): the counter lived in the store image but dropped at the session/schema mapping, so a wait-less restart re-minted wait_1. Fix adds next_wait_id to the schema + round-trips both directions (v1 defaults documented: seed from carried waits/pending, else fresh-store 0). Pin: TestFixWave16_NextWaitIDRoundTrip (counter survives a wait-less persist/restore; next registration continues past N).
2. LOW nearest inconsistency (confirmed): live + watchdog projections used lexicographic wait_id while schema.NearestWait orders by registration. Fix adds goal.WaitOrderLess (same deadline/RegisteredAt/numeric-suffix/id chain) and uses it at both session sites. Pin: TestFixWave16_NearestWaitNumericTieBreak (wait_2 beats wait_10 on all three surfaces).
3. LOW digest root-dependence (confirmed): lines hashed the absolute path. Fix hashes the relative entry name (signature kept — ReadDir still needs root). Pin: TestFixWave16_WorktreeDigestRootIndependent (same listing, two roots, identical digest).
4. LOW parseWaitSeq overflow (confirmed): unguarded accumulation could wrap huge suffixes. Fix guards with (maxInt-d)/10, returns -1 past representability (both schema + goal copies). Pin: TestFixWave16_ParseWaitSeqOverflow.

Note: the goal/schema ordering helpers are intentionally mirrored per package (leaf-package boundary — schema cannot import the store); both carry the same chain + overflow guard + regression pins. Web lane: the e743cca run's web failure was a Chrome-startup env flake (dbus/30s deadline, zero branch frontend files); the --failed rerun is now SUCCESS, so that run is fully green. Local: build + vet clean, internal/goal + schema + wide session surface green, lint 0 on changed packages.

obra:
Round-17 wave pushed as 3d3835a0c — 6.5 fixed, 1 refuted with evidence (no change):

Refuted (no change): HIGH recursion — kickGoalWaitBudgetBound already returns via the pendingAsk arm-only branch and never re-arms on the kick path (the No re-arm comment names the exact arm→kick→arm overflow the reviewer fears). The cited line is the guard, not the bug.

Fixed:
1. MEDIUM verify race (confirmed): update_goal complete ran Conditions→Evaluate→SetTerminal as three unserialized guard calls. Fix adds completeGoalIfConditionsSatisfied (verify+commit as one unit under goalUpdateMu; EvaluateExpectations copies substrate under a short store hold and evaluates outside it, so no new lock order) + goalGuard.CompleteIfSatisfied forwarder. Messages byte-identical. Pins: satisfied-completes + reject-names-condition.
2. MEDIUM stale emission (confirmed): unlock-then-emit snapshots in register/cancel paths could regress consumers after a concurrent retarget. Fix adds the monotonic goalEventGen (bumped under goalUpdateMu on every publishing mutation incl. Set/ClearGoal/resume/setGoalTerminal); unlock-then-emit sites gate on it (best-effort, documented); held-serializer emits bump for the same suppression. Pins: committed-order concurrency test + stale-suppressed.
3. MEDIUM dead expect params (confirmed): timeout_seconds/label removed from DefGoalExpect (schema additionalProperties:false rejects them) + validation/decode no longer read them (stored predicate Timeout stays zero). DefGoalWait untouched. Pins: schema-rejects + stored-zero.
4. MEDIUM mtime precision (confirmed): digest now RFC3339Nano matching StatFile baselines. Pin: sub-second delta flips the digest.
5. LOW derived labels (confirmed): kind:target sanitizes to 256 printable runes (U+FFFD), target identity intact (label excluded from the key). Pins: capped + long-target + sanitized.
6. LOW itoa (confirmed): strconv.FormatInt replaces the hand-rolled negation. Pin: MinInt64 exact.
7. PARTIAL expect message: until_time alone routes to the verifiable-state reason (mirroring expectKindRejected); all other unknowns keep the terse message byte-identical. Pin: until_time vs bogus.

Local: build + vet clean, internal/goal + internal/tool green, wide session surface green (25s), -race on expect/emit/complete green, lint 0 on changed packages.

obra:
Round-18 wave pushed as 800a6dcfe — 2 fixed (1 High + 1 Medium), both wake-turn budget accounting:

1. HIGH sibling-live undercharge (confirmed): ClaimFire leaving sibling live waits kept the goal waiting, so the wake-drive fold hit RecordContinuation's parked early-return — a real model turn ran uncharged. Fix adds RecordWakeContinuation (identical body, skips only the waiting early-return; parked polling still uses the zero-accrual path) + commitGoalWakeFold at the wake-drive site. Pins: TestFixWave18_SiblingLiveWakeDriveFolds (Iterations=1, Used=1 with a live sibling; tail parks without re-fold); existing parked-skip + last-lease-wake tests unchanged.
2. MEDIUM wake cap gap (confirmed at both sites): the wake fold could land exactly on MaxContinuations with the prompt rendered anyway, and the wake-tail bypass re-armed without a bounds check. Fix enforces on the committed snapshot at the wake-drive site and before the tail re-arm — same blockGoalFromGate/VerdictBudgetExhausted path as the plain-drive site. Scope is continuation-cap-only at both: the fold moves neither anchor nor clock, so parked/deadline read identically pre/post fold and keep their own latched verdict paths (the deadline final turn still drives — TestFixWaveI2DeadlineFinalTurnThenBlock green). Pins: landing-on-cap blocks with waits cleared + exactly one budget note; at-cap tail blocks instead of re-arming.

Local: build + vet clean, internal/goal green (+race), wide session surface green, lint 0 on changed packages.

obra:
Round-18 wave pushed as 800a6dcfe + latch-scope follow-up bd2fda67e — 2 fixed (1 High + 1 Medium), both wake-turn budget accounting:

1. HIGH sibling-live undercharge (confirmed): ClaimFire leaving sibling live waits kept the goal waiting, so the wake-drive fold hit RecordContinuation's parked early-return — a real model turn ran uncharged. Fix adds RecordWakeContinuation (identical body, skips only the waiting early-return; parked polling still uses the zero-accrual path) + commitGoalWakeFold at the wake-drive site. Pins: TestFixWave18_SiblingLiveWakeDriveFolds (Iterations=1, Used=1 with a live sibling; tail parks without re-fold); parked-skip + last-lease-wake tests unchanged.
2. MEDIUM wake cap gap (confirmed at both sites): the wake fold could land exactly on MaxContinuations with the prompt rendered anyway, and the wake-tail bypass re-armed without a bounds check. Fix enforces on the committed snapshot at the wake-drive site and before the tail re-arm — same blockGoalFromGate/VerdictBudgetExhausted path as the plain-drive site. The drive site is scoped to fold-spent caps only (bd2fda67e): a pre-spent cap stays latched so the terminal-flagged wake still drives first per spec §1 rule 1 (speculative block broke the two latch tests — verified failing, then fixed). The tail site keeps the full bounds check. Scope otherwise continuation-cap-only: parked/deadline keep their latched verdict paths (deadline final turn still drives). Pins: landing-on-cap blocks with waits cleared + exactly one budget note; at-cap tail blocks instead of re-arming.

Note: CI on 800a6dcfe caught the latch regression (tests + race/agent lanes); the follow-up was verified locally (latch pair + FixWave18 + I2 + expiry + cap pins green, store + wide surface + vet + lint clean) before pushing. Local: build + vet clean, internal/goal green (+race), wide session surface green, lint 0 on changed packages.

obra:
Round-19 wave pushed as 860c0084b — 2 fixed (Medium, no High):

1. Approval cancel misreported (confirmed): a missing live ask always fired approval answered, but askPending clears on interrupt, every turn entry, and heuristic restore — a cleared-but-unanswered ask reads identically at the substrate. Fix fires the neutral approval ask resolved or cleared + re-validate excerpt instead of asserting an answer (park/fire structure unchanged; the reply turn's content carries the verdict). Building a real answered-vs-cancelled record is out of scope (no such signal in the ask lifecycle). Stale answered comments updated; C1 pin retargeted. Pins: cleared-ask fires neutral without claiming answered, live ask still parks.
2. Terminal backlog carried into retarget (confirmed): SetTerminal preserved a backlog the gate never drains (terminals short-circuit first), so a later Set inherited a stale superseded no-op. Fix clears PendingWake at terminalization alongside waits; the live claim-to-kick race keeps its Superseded carry (only terminal backlogs are now always empty). Pins: complete drains the backlog; retarget after terminal drives the new objective with no stale trigger and normal first-turn accounting; live superseded-race tests unchanged.

Local: build + vet clean, internal/goal green (+race), wide session surface green (39s), lint 0 on changed packages.

obra:
Round-20 wave pushed as a10b8e92d — 6 done (5 fixes + 1 documented partial, no High):

1. MEDIUM restore-deadline strand (confirmed): restoreGoalAttachScan claimed expired leases but never the expired goal deadline — the follow-up arm disarmed on the past deadline with no wake. Fix mirrors the gate's synthetic claim at restore (one-shot via the marker; no loss recorded; rides the existing backlog-kick path). Pins: past-deadline claims the synthetic entry + marker set; delivered marker suppresses.
2. MEDIUM extend overflow (confirmed): earlier guards covered the request value but not a hostile persisted base. Fix clamps the base first, then saturates the add (wrap check) to the cap; messages unchanged. Pins: maxInt-ish continuations + maxDuration-ish parked-total saturate to caps.
3. MEDIUM TOCTOU (partial): race real but unexploitable (no hostile sandbox writer; content reads already fd-anchored). Comment-only documenting the point-in-time contract + per-tick self-correction; no behavior change.
4. LOW empty subtype (confirmed): "" aliased file-modified at the gate but rejected in prescan/replay/eval. Fix normalizes once before the registrable check ("" and file_modified dedupe together). Pin: empty subtype registers as file_modified.
5. LOW Sscanf divergence (confirmed): Sscanf accepted wait_10evil as 10. Fix reuses parseWaitSeq (with its overflow guard) in maxNextWaitID. Pin: suffixed ids never seed the counter.
6. LOW chip duplication (confirmed): hub waiting branch now delegates to appprojector.GoalWaitingChipText (import precedent clean, no cycle); both carry drift notes. Pin: waiting parity test.

Local: build + vet clean (agent/goal/cmd/projector), internal/goal green (+race), wide session surface green (51s), tui + projector suites green, lint 0 on changed packages.

obra:
Lint fix pushed as daf27aec8: the round-20 extend saturation named a variable new, shadowing the builtin (revive redefines-builtin-id in lint-golangci + static aggregation). Renamed to raised at both sites; extend-saturation tests green, goal package lint 0 issues. (The 3 remaining modernize findings in retained_output_size_acc_test.go pre-exist on clean HEAD — verified via stash — and that file is untouched by this branch.)

PR #1038, opened 2026-09-09
feat(web-tasks): settled group first, drop task-list progress bar

obra:
Moves the folded-away Done-settled group above In progress / Open in the web task list, and drops the progress bar from the task list body (count text retained).

Changes:
- TaskListGroups renders the collapsed settled disclosure first (TasksPanel.tsx)
- Body header keeps count text, Meter + taskMeterValue helper removed
- Tests pin settled-above-live DOM order and assert no meter remains
- CSS header comment updated (no meter reference)

Verification:
- biome check --write clean on touched files
- TasksPanel.test.tsx: 47/47 pass
- make test-web: web-typecheck PASS, web-test PASS, web-lint FAIL (6 pre-existing import-sort/format errors in untouched files: DocPane, AdvancedOptions, Rail, RailRow, SessionMenu)

### DIFF
--- a/cmd/evener-hub/frontend/src/panes/session/chrome/TasksPanel.test.tsx
+++ b/cmd/evener-hub/frontend/src/panes/session/chrome/TasksPanel.test.tsx
@@ -156,7 +156,7 @@ test("the trigger shows the done/total counts once the aggregate has arrived", (
   expect(screen.getByRole("button", { name: "Tasks 3/7" })).toBeTruthy();
 });
 
-test("outcome aggregates show terminal progress instead of done/total", async () => {
+test("outcome aggregates show terminal counts instead of done/total", async () => {
   const fake = connectFakeClient();
   fake.on("evener/tasks/list", () => ({ data: [] }));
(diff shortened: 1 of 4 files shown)


obra:
Review triage on the combined review (commit 17ff0cb), fixed in 5d5150898:

- **Medium: shared-notes steer/no-op race (design:59)** — moot for this PR. That file entered the diff only because the branch sat on a stale local base; after rebasing onto origin/main the PR holds only the 3 task files. The finding belongs to wherever the shared-notes work lands, not here.
- **Medium: agent-only tools in the AppWire catalog (plan:470)** — same: the plan file is out of the PR after the rebase. Moot here.
- **Low: settled-before-live contradicts the task-list design doc** — valid. Fixed by updating docs/superpowers/specs/2026-08-09-task-list-ui-design.md to the new order (settled first, then in-progress, then open) and the count-only body header, per the explicit user direction that overrides the old spec.

Verification: biome clean on touched frontend files, TasksPanel.test.tsx 47/47 pass.

PR #1041, opened 2026-09-09
Move mid-exchange agent timestamps out of the flow into the right margin

obra:
## What

A mid-exchange agent fragment's relative time used to render as its own caption line above the prose, adding a row of vertical height to every fragment.

- **Desktop (≥700px):** the stamp leaves the flow entirely — italic, one gap step (`--space-3`) outside the transcript column's right edge, aligned with the first prose line, adding zero height. The virtual list's `overflow-x: clip` guarantees a pane too narrow to have a margin clips the stamp instead of scrolling sideways.
- **Mobile (<700px):** the stamp stays in flow on its own line after the content (DOM reordered bubble → stamp), right-justified and snug underneath.

Speaker-header times (`Agent · model · 9m ago`, `You 9m ago`) are unchanged — they're the slack-lean attribution line and never added height.

## How

- `AgentMessageItem.tsx`: continuation branch renders the timestamp after the bubble in a `.stamp` wrapper; DOM order makes the mobile layout work without flex reordering.
- `agentmessageitem.module.css`: `.message` becomes the positioning context; `.stamp` gets the two media-query treatments. Mid-exchange rows span to the column's right edge (runContent pads only the left), so `left: calc(100% + gap)` lands exactly outside the column.

## Verification

- `make test-web`: PASS (typecheck, full vitest suite, lint) — includes 4 new tests pinning the stamp's DOM placement and both media-query treatments (declaration-level).
- `make test-web-browser`: PASS (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard); overflowguard's 390/700/1024/1400 sweep confirms no horizontal overflow.
- Visual check against the real stylesheets at 1440px desktop and 390px phone: stamp sits just outside the column's right edge on desktop, in-column right-justified under the content on mobile.

### DIFF
--- a/cmd/evener-hub/frontend/scripts/layoutguard/cases/transcript-stamp-rail-wide/assert.mjs
+++ b/cmd/evener-hub/frontend/scripts/layoutguard/cases/transcript-stamp-rail-wide/assert.mjs
@@ -1,0 +1,57 @@
+// The wide-measure stamp contract (PR 1041): with the 64rem column the
+// 55-75rem band has no real margin, so the override must keep the stamp IN
+// FLOW with wrapping restored (white-space: normal - the rail's nowrap is
+// not inert on a static box), and the rail may only engage from a 75rem
+// (1200px) container. Italic in both regimes. Measured before the override
+// reverted wrapping: the in-flow stamp in the band kept the rail's nowrap.
+const RAIL_MIN = 1200;
+
+export default function assert(measurements) {
+  const failures = [];
+  const reasons = [];
+  for (const m of measurements) {
+    const label = `${m.width}px container`;
+    const expectRail = m.width >= RAIL_MIN;
+    if (expectRail) {
+      if (m.position !== "absolute") {
+        failures.push(
+          `${label}: stamp is in flow (${m.position}) but a 75rem container has room for the rail even at the wide measure - the 55rem rail rule is missing`,
+        );
+        continue;
+      }
+      if (!m.outsideRight) {
+        failures.push(`${label}: rail stamp is not fully right of the column`);
+      }
+      if (!m.railAligned) {
+        failures.push(`${label}: rail stamp is not aligned with the first prose line`);
+      }
+      if (m.whiteSpace !== "nowrap") {
+        failures.push(`${label}: rail stamp lost its nowrap (${m.whiteSpace})`);
+      }
+    } else {
+      if (m.position !== "static") {
+        failures.push(
+          `${label}: stamp left the flow (${m.position}) inside the wide-measure revert band - the 55-75rem override's position: static is missing`,
+        );
+        continue;
+      }
+      if (m.whiteSpace !== "normal") {
+        failures.push(
+          `${label}: in-flow band stamp keeps the rail's ${m.whiteSpace} - the override must revert to white-space: normal`,
+        );
+      }
+      if (!m.insideColumn) {
+        failures.push(`${label}: in-flow stamp overflows the column's right edge`);
+      }
+      if (!m.belowBubble) {
+        failures.push(`${label}: in-flow stamp is not under the content`);
+      }
+    }
+    if (m.fontStyle !== "italic") {
+      failures.push(`${label}: stamp is not italic (${m.fontStyle})`);
+    }
+    reasons.push(`${label}: ${m.position}, ${expectRail ? "rail" : "band in-flow"} regime`);
+  }
+  if (failures.length > 0) return { pass: false, reason: failures.join("; ") };
+  return { pass: true, reason: reasons.join("; ") };
+}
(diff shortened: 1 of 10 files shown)


obra:
RoboRev round addressed in 35eb30b (on top of 6dd90edd, rebased onto current main):

**Medium — clipping risk (agentmessageitem.module.css:117):** Confirmed and fixed. The in-flow right-justified treatment is now the base at every width; the stamp only leaves the flow at ≥1200px windows, where 280px rail + 48px pane padding + 704px measure leaves a real margin (below 900px the shell is rail-less but the column fills the pane from ~750px; at 900px the rail reappears and the margin vanishes again until ~1200). The wide transcript measure (64rem) keeps the in-flow base until 1536px. Verified visually at 1024px: the stamp stays readable in-column instead of clipping.

**Medium — brittle CSS-text test (AgentMessageItem.test.tsx:434):** Trimmed to the structural contract only — in-flow base rule, out-of-flow rail inside the min-width media query, `.message` as containing block. Exact offsets/gaps/tokens are no longer regex-pinned; geometry belongs to the browser guards.

**Medium — non-null assertions (AgentMessageItem.test.tsx:415):** Removed both `stamp!` uses; the tests now narrow with explicit throw-guards.

**Low — top offset 8px high (agentmessageitem.module.css:120):** Confirmed — `.continuation`'s `--rhythm-item` margin was missing from the offset. Now `top: calc(var(--rhythm-item) + 2 * var(--rhythm-line))`; verified visually that the stamp aligns with the first prose line.

**Low — 699/700px fractional gap:** Closed. The base `.stamp` rule lives outside any media query, so no viewport width falls onto an unstyled span.

Gates on the pushed tree: `make test-web` PASS (typecheck, full vitest suite, lint) and `make test-web-browser` PASS (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard).

obra:
RoboRev round 2 addressed in f46b386:

**Medium — viewport-gated rail clips in narrow dock splits (agentmessageitem.module.css:134):** Confirmed; implemented the suggested container-query fix. The virtual list's root is now an `inline-size` query container — safe for the same reasons composer.module.css/spawn.module.css record (width comes from the flex parent, never content; positioned descendants anchor locally to sizer/item/row; fixed-position widgets portal to body, so the containing block containment establishes captures nothing). The rail now engages at `@container (min-width: 55rem)` = the 44rem measure + ~88px per centered side for gap + stamp, and the wide transcript measure keeps the in-flow base until 75rem. A window-wide but narrow split pane now keeps the readable in-flow right-justified stamp instead of clipping. The declaration-level test pins the container-query contract plus the root's `container-type`.

Verified visually against the real stylesheets: rail engaged at a 1112px pane, in-flow and readable at a 500px pane inside a wide window (the reported failure mode), in-flow at 390px.

Gates on the pushed tree: `make test-web` PASS (typecheck, full vitest suite, lint) and `make test-web-browser` PASS (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard).

obra:
RoboRev round 3 addressed in b744dac:

**Medium — nowrap leaks into the wide-measure in-flow fallback:** Confirmed — my comment claimed `nowrap` is inert on a static box, which is wrong (`white-space` applies to in-flow boxes). The override now reverts `white-space: normal` explicitly alongside `position: static`, and the comment is corrected.

**Low — declaration-level test:** Partially accepted. The geometry half is now covered by a real browser guard: new `transcript-stamp-rail` layoutguard case checks the stamp at four container widths — absolute and fully right of the column, aligned with the first prose line, at 1112px and 900px containers; in-flow, inside the column, under the content at 860px and 500px; italic in both regimes. Mutation-tested per the repo's guard kata: with the rail rule neutralized the case fails on both wide fixtures (evidence recorded in case.json). The trimmed declaration-level test stays: declaration-existence contracts are the house pattern for jsdom-invisible CSS (the stream-caret tests in the very same file, Composer's `container-type` assertions in Composer.test.tsx:717), and it pins what no DOM or geometry test can — that the rail rule and the list root's `container-type` exist at all.

Gates on the pushed tree: `make test-web` PASS (typecheck, full vitest suite, lint) and `make test-web-browser` PASS (layoutguard including the new case, overflowguard, shellguard, spawnguard, transcriptscrollguard).

obra:
RoboRev round 4 addressed in a8ee3f2:

**Low — wide-measure revert band uncovered:** Closed. New `transcript-stamp-rail-wide` layoutguard case renders the same mid-exchange DOM under `data-transcript-measure="wide"` and verifies in a real browser: at a 1112px container (inside the 55-75rem band) the stamp stays in flow, inside the column, under the content, with `white-space: normal` restored; at a 1300px container the rail engages (absolute, fully right of the column, first-line aligned, nowrap intact); italic in both regimes. Mutation-tested per the guard kata: removing the override's `position: static` fails the band fixture (evidence in case.json). The declaration-level test now also pins the band rule's `position: static` + `white-space: normal`, and the reading-measure case.json cross-references its wide twin.

Gates on the pushed tree: `make test-web` PASS (typecheck, full vitest suite, lint) and `make test-web-browser` PASS (all five guards, layoutguard now including both stamp cases).

PR #1044, opened 2026-09-09
feat(webui,tui): go to next/previous live session key bindings

obra:
## Summary

Adds "go to next/previous live session" key bindings on both surfaces:

- **Web hub UI**: new configurable `session.liveNext` / `session.livePrevious` actions, default chords **Alt+Shift+ArrowRight / Alt+Shift+ArrowLeft**. They *navigate* across the rail's live section in server order, wrapping at both ends — unlike `session.next`/`session.previous`, which only cycle focus between already-open panes. Selection semantics live in `shell/rail/liveSessionCycle.ts` (a bidirectional `nextNeedsYouRef`); navigation reuses the `openNeedsYouSession` URL seam ⌘J uses. Chords suppress in editable targets (Alt+Shift+Arrow is native word selection) and register desktop-only. They appear automatically in Settings → Keybindings and the cheatsheet via the default map.
- **TUI**: the same chords in session view (`alt+shift+right` / `alt+shift+left`) switch the viewed session to the adjacent live session in dashboard order through the ordinary `fetchHubSession` entry read. Plus `/next-live-session` and `/previous-live-session` palette commands and `/help` entries.

Second commit fixes a bug the wiring tests exposed: the rail Tree widget `preventDefault`'d every arrow key regardless of modifiers, so with a rail row focused, *all* global Alt+Arrow chords (pane cycling, transcript scroll, and the new pair) were dead — the dispatcher's `defaultPrevented` gate skipped them. Tree now bails on Alt/Ctrl/Meta, matching RailResizeHandle's existing guard.

## Test plan

- New unit tests: `liveSessionCycle.test.ts` (order, wrap both directions, current-not-in-list, empty, singleton); `hub_livecycle_test.go` (switch, wrap, non-live start, single-live no-op, empty tree, palette registration).
- Wiring tests through the real dispatcher/defaults/shell in `AppShell.test.tsx` (navigation, wrap, editable-target suppression, mobile inertness); Tree modifier-guard tests in `tree.test.tsx`.
- Gates: `make test-web` (typecheck, 9200+ vitest, biome ci) green; `go vet` + `go test ./cmd/evener-tui/` green; `golangci-lint run ./cmd/evener-tui/...` clean.


obra:
RoboRev round 1 addressed in 6e66ac6 (each fix landed test-first):

- **Pagination (AppShell.tsx:541-549)** — confirmed: the live section pages at limit 50 and `selectLiveRows` aggregates only loaded pages. `session.liveNext` now demand-loads the next page at the last loaded row (or an unloaded section) when `remaining > 0`, continuing into the first newly loaded row instead of wrapping over the subset — the needs-you handler's `openDemandedPage` pattern. `previous` wraps within loaded rows; section pages only page forward, so the true last live row is unknowable until all pages load (documented at the handler). Verified by a new wiring test: paginated two-page live section, chord at the last loaded row lands on the next page's first session rather than wrapping.
- **Async race (hub_livecycle.go)** — confirmed: `detail.Ref` stays stale until the read lands, so repeated presses computed the same target and in-flight reads could apply out of order. Presses now step from `liveNavPendingRef`, and each cycling read carries a per-press `liveNavSeq`; `Update` drops a read a newer press has superseded (after `capture.Release()`, so its frames still flow — the superseding read's capture adopted its held frames via `takeLocked`). Verified by new tests: repeated presses read `[01C, 01A]` (based on the pending target), and a late stale read no longer moves the view.

Gates on this commit: affected vitest scopes (shell 973, keybindings/tree/settings/rail 1295) + `tsc` + Biome on touched files + `go vet` + `go test ./cmd/evener-tui/` all exit 0.

obra:
RoboRev round 2 addressed in 6ec7cb6 (each fix test-first):

- **demandedLivePages lifecycle** — confirmed, with one correction to the suggested mechanism: the revalidator resolves a failed load with an error ResourceState rather than rejecting, so the completion handler inspects `page.error`/`page.data` and frees the page ID for retry; both live-nav handlers also clear the cache on `clientGenerationID` change, matching the needs-you precedent. Test: first page-two read throws, the press no-ops, the next press re-issues the read and navigates.
- **Demand completion guard** — confirmed. Every live-nav press (ordinary or demand) now bumps the intent counter, and completion validates `focusedSessionRef()` at press vs completion plus palette-open and `[aria-modal]` guards — the needs-you `openDemandedPage` guard set. Tests: a deferred page-two read resolved after a superseding press does not navigate; same with the palette open.
- **TUI live-nav state on manual entry** — confirmed. Applying a non-cycling session entry (any detail.Ref change that isn't itself a tagged read) now bumps `liveNavSeq` and clears `liveNavPendingRef`, so an in-flight cycling read drops when it lands and the next press steps from the viewed session. Status refreshes and same-ref resyncs take the early returns and never invalidate. Test: manual entry mid-flight, late tagged read dropped, following press bases on the viewed ref.
- **Tree modifier guard scope** — confirmed: the guard now covers only Arrow keys, so modifier+Enter activates the row again. Test pinned.

Gates on this commit: `tsc`, Biome on touched files, vitest (AppShell 93, incl. the round-1 tests), `go vet`, `go test ./cmd/evener-tui/` all exit 0.

obra:
RoboRev round 3 addressed in f05bbfc (both test-first):

- **Unloaded-section bootstrap (AppShell)** — confirmed: `selectSectionRemaining` reads only loaded pages, so with the initial live read in flight or failed it is 0 and neither action ever demanded a page. Both live-nav handlers now consult the manifest's live count when the loaded set is empty and demand page zero — deduped against an in-flight initial read by the revalidator, retried after a failed one via the round-2 `page.error` handling. The needs-you handler's manifest-count bootstrap is the precedent. Test: initial live read throws, chord press re-issues it and navigates once the retry lands.
- **TUI mode-exit race (hub_update.go)** — confirmed: a tagged read whose sequence was current applied even after ctrl+o left session mode, re-entering the fetched session. Tagged reads now also drop when `m.mode != hubModeSession`. Test: press, returnToDashboard, late read delivered — the view stays on the dashboard.

Gates on this commit: `tsc`, Biome, vitest (AppShell 94 + keybindings/shell/widgets/settings scopes), `go vet`, `go test ./cmd/evener-tui/` all exit 0.

obra:
RoboRev round 4 addressed in 23cf3c2 (all four test-first):

- **previous opens the wrong row after demand-loading** — confirmed. The demand flow is now direction-parameterized: `previous` from an unloaded set keeps paging while the server reports `remaining > 0`, then opens the last row overall (the true tail), instead of the first newly loaded row. Test: manifest reports 2 live sessions, page zero deferred; previous resolves it, continues into page two, and lands on the tail — not the head.
- **rapid same-direction presses drop navigation** — confirmed. Intent semantics are now: starting a *new* demand or navigating directly bumps `liveNavIntent`; a press that dedupes against an already in-flight demand is the same intent and does not bump, so the demand's completion still applies. Two fast presses at the boundary navigate when the page lands. Test pinned.
- **TUI chords intercepted mid-edit** — confirmed: the switch replaces `m.session` (input included), discarding an unsent draft. The chords now hold while the composer holds a draft and post a system note ("Draft kept. Send or clear it before switching live sessions.") — the TUI mapping of the web bindings' `allowInEditable: false`. Test: draft present → no read issued, draft intact.
- **TUI live-nav inert in browse mode** — confirmed. The chords now dispatch in the browse-mode switch; unmodified arrows keep their browse scroll/selection meanings. Test pinned.

Gates on this commit: `tsc`, Biome, vitest (AppShell 96 + keybindings/shell/widgets/settings scopes), `go vet`, `go test ./cmd/evener-tui/` all exit 0.

obra:
RoboRev round 5 addressed in 2efc225 (all three test-first):

- **previous wraps to last loaded row on partial load** — confirmed. A previous press that would wrap (from the first loaded row, or from a non-live session) while `remaining > 0` now demand-loads through the remaining pages to the true tail, same as the unloaded-section path; mid-list steps still need nothing beyond the loaded rows. Docs updated to match. Note on the test estate: the round-2 "newer press supersedes demand" test's second press (previous from the first/only loaded row) is itself a demand under this rule, so that test now supersedes with a genuine direct mid-list step — the supersede contract is unchanged and still pinned.
- **draft/attachment loss when composing begins mid-flight** — confirmed: the press-time guard can't cover content created during the read. A tagged read now re-checks the composer and pending attachments at application time and drops (clearing the pending target) when either is non-empty — the draft is newer intent than the navigation. Test: type mid-flight, late read drops, draft intact.
- **attachment-only draft unprotected** — confirmed; the press-time guard now covers `len(pendingAttachments) > 0` too. Test pinned.

Gates on this commit: `tsc`, Biome, vitest (AppShell 97 + keybindings/shell/widgets/settings scopes), `go vet`, `go test ./cmd/evener-tui/` all exit 0.

obra:
## RoboRev round 6 — both findings fixed in `13b57449b`

**Medium (AppShell.tsx demand dedupe vs opposite-direction press)** — confirmed. The dedupe keyed the page ID alone, so an opposite-direction press while that page was in flight returned early *without* bumping `liveNavIntent`; the stale continuation still owned the navigation (Previous-then-Next with page zero in flight opened the tail, not the first row). Fix: the dedupe key is now `pageID:direction` — repeated same-direction presses keep the round-4 dedupe (no intent bump, the demand lands and navigates), while an opposite-direction press misses the key, bumps the intent, and supersedes the prior continuation (the old one goes inert on its `intentAtPress !== liveNavIntent` check). Test: `an opposite-direction press while a demanded page is in flight supersedes the prior intent` — Previous then Next against a deferred page zero; RED confirmed first (navigated to the tail `live-b`), now opens the first row `live-a`.

**Low (TUI tagged read vs overlay opened mid-flight)** — confirmed. The `liveNavSeq > 0` application block checked mode/seq/draft but not overlays, so a read landing while the user had opened the palette (or any modal) swapped the session under the open overlay. Fix: `topmostOverlayName(m) != ""` is now checked in the same block, clearing `liveNavPendingRef` and dropping the read — the same class of newer intent as the round-5 mid-flight draft. Palette-issued live-nav commands are unaffected by construction: `updateCommandPaletteKey` sets `commandPalette = nil` *before* dispatching the command, and the focus trap swallows the chord itself while any overlay is open. Test: `TestHubSessionLiveCycleDropsReadWhenOverlayOpenedMidFlight` — palette opened during an in-flight live-next read; RED confirmed first (applied `01C`), now stays on `01B` with the palette still open.

Verification: focused frontend suites 2922 pass + typecheck + biome clean; full TUI package green (`go vet`, `gofmt`, `-short` suite).

obra:
## RoboRev round 7 — both findings fixed in `653266ae`

**Medium 1 (TUI stale `liveNavPendingRef` on failure)** — confirmed, with one refinement of the reachable path: the direct error path (`msg.err` on a tagged read) already falls through the drop block, which clears the pending ref — the actual leak was the *mode-exit drop* (ctrl+o while the read is in flight), which returned before the clear, and nothing on a dashboard round-trip cleared it either. So: B→C read dropped by ctrl+o, user re-enters B, next chord stepped from the stale C target (jumping to A, skipping C). Fix: the mode-exit drop now clears `liveNavPendingRef` before returning. Test: `TestHubSessionLiveCycleClearsPendingRefAcrossDashboardRoundTrip` — pending target set, dashboard round-trip, next press must land on C (from the displayed B), not A. RED confirmed first (`liveNavPendingRef = "local:01C"` survived the drop), now green.

**Medium 2 (web: no generation guard on the demand-load continuation)** — confirmed. `liveNavGeneration` was reset at press time in both handlers, but a demand already in flight resolved after a reconnect (generation bump) with retained provisional rows and navigated into the previous generation. Fix: `demandLivePage` captures `generationAtPress` alongside the intent/pane/ref snapshot, and the completion goes inert when `clientGenerationID` has moved on — same inertness class as the round-2 supersede/palette guards. Test: `an in-flight live demand-load goes inert when the client generation changes` — demand in flight, generation bumped via reconnect simulation, deferred page resolves, no navigation. RED confirmed first (navigated to `/s/local%3Alive-b`), now green.

Verification: focused frontend suites 2923 pass + typecheck + biome clean; full TUI package green (`go vet`, `gofmt`, `-short` suite).

obra:
## RoboRev round 8 — all six findings fixed in `d4bbd785`

**Medium 1 (TUI: stale read unconditionally cleared `liveNavPendingRef`)** — confirmed, and the fix follows your suggested shape: the sequence-mismatch drop in session mode now returns *without* clearing, so the newer in-flight read keeps its stepping base. Only the mode-exit drop clears (round-7's cleanup). Test: `TestHubSessionLiveCycleStaleReadKeepsNewerPendingTarget` — two presses, stale lands first, the newer pending target survives and the newer read applies. RED confirmed first (pending ref was cleared).

**Medium 2 (TUI: dropped reads still replace the server-side subscription)** — confirmed. The dropped stale read now re-issues `fetchHubSession` for the session currently displayed (`currentRef`), re-establishing the subscription the stale read's `ReplaceSubscription` displaced. Test: `TestHubSessionLiveCycleStaleReadReEstablishesCurrentSubscription` — asserts the drop returns a resub Cmd whose read re-targets the displayed ref.

**Medium 3 (web: URL-equal live-nav did not focus the session pane)** — confirmed. `navigate()` no-ops on an unchanged pathname, so the route-placement effect never ran and a secondary panel or another pane kept focus after a press targeting the URL's own session. New `openLiveSession` seam: `openNeedsYouSession(ref)` then a focus guarantee — `replacePrimary` when the main pane isn't the target session (it both opens and focuses), `focusPane(main.id)` when it is. Test: `live-next focuses the session pane even when the URL already matches` — sessionTasks panel focused, URL-equal previous press; focus must land back on the session pane. RED confirmed first (focus stayed on the panel).

**Medium 4 (TUI: live-nav seq not tied to connection epoch)** — confirmed. `applyHubReconnect` now bumps `liveNavSeq` and clears `liveNavPendingRef` (the new connection has no cycling read in flight), so an old-connection read that lands after the reconnect drops as stale instead of overwriting the resynchronized session. Test: `TestHubSessionLiveCycleDropsReadFromBeforeReconnect` — the read's message is captured *before* the reconnect (the real race), then delivered after; RED confirmed first (overwrote `local:01B` with `local:01C`).

**Low 1 (web: completed demands stay in `demandedLivePages` forever)** — confirmed. The demand key is now deleted when its load completes (the set tracks in-flight only), so a re-demand of the same page+direction after a non-generation invalidation is issued rather than swallowed. Test: `a live demand can be re-issued after an invalidation restales the pages` — boundary press after `evener/navigation/invalidated` must issue a fresh demand; RED verified via probe (the second demand was swallowed without the fix — the revalidator's own refresh had masked it in the first test draft).

**Low 2 (tree guard swallowed Ctrl/Meta arrows)** — confirmed with one refinement from your text: the defaults map binds **only the Alt family** of arrow chords (Alt+Arrow pane cycling, Alt+Shift+Arrow live-nav/scroll, Alt+Home/End), so the guard is narrowed to `event.altKey` alone — Ctrl/Meta/Shift arrows all stay tree-owned. Tests updated: the Alt-chord release test now covers Alt and Alt+Shift only, and a new `arrow keys with Ctrl or Meta held still navigate the tree` asserts collapse/focus movement under Ctrl/Meta.

Verification: 2926 focused frontend tests + typecheck + biome clean; full TUI package green (`go vet`, `gofmt`, `-short` suite). Round-7 CI on `653266ae4` was all-green before this push.

obra:
## Round 9 fixes

Commit: b78563ea6. All five findings triaged valid; fixed as follows.

### Medium 1 — Alt+Arrow on the mobile rail is browser history navigation

`Tree` now takes an opt-in `releaseModifierKeys?: boolean` prop (default `false`): the release guard only fires when a consumer that actually hosts the global chords sets it. The rail's three `Tree` call sites (sessions, pinned sections, archived) go through a new `RailTree` wrapper that calls `useIsMobile()` and passes `releaseModifierKeys={!isMobile}` — the desktop rail (RailHost) releases Alt-held arrows/Home/End to the global chords, the mobile rail (StackHost's rail slot) keeps every arrow tree-owned since the chords are inert there and released Alt+ArrowLeft/Right would fall through to the browser's history navigation.

`tree.test.tsx` pins both sides: with the prop, Alt-held arrows and Alt+Home/End are not preventDefaulted and take no tree action; without it (the mobile default), Alt-held arrows still navigate the tree.

### Medium 2 — Alt+Home/End still swallowed

The guard now covers `Home`/`End` under Alt in addition to the arrow family, matching the `transcript.scrollTop`/`scrollBottom` chords `defaults.ts` binds. (Note: the tree's keymap switch never had Home/End cases, so they were never actually preventDefaulted — but the release guard now states and protects the contract explicitly, and the tests pin it.)

### Medium 3 — inert completion leaves the in-flight key behind

The inert early return now runs `demandedLivePages.delete(demandKey)` before returning, so the set keeps tracking in-flight demands only — the same invariant the success and error paths already maintain.

One honesty note on testability: a same-key re-demand is not reachable end-to-end, because a page that resolves with data always advances `selectNextSectionOffset` past itself (`offset + returned rows`, and wireV2 rejects zero-row pages with remaining data), while a page that resolves without data routes through the already-deleting error path. The new `AppShell.test.tsx` case pins the reachable lifecycle instead — a palette-inert completion leaves nothing sticky: after the palette closes, the same chord still navigates (here via the page the inert completion loaded).

### Medium 4 — leave-and-return defeats the pane/ref guards

`AppShell` now maintains a monotonic route epoch: `liveNavRouteEpochRef` bumps on every `pathname` change (render-phase ref write, the same precedent as `renderTimePanesRef` — so it lands in the same commit as the navigation, and StrictMode's double render only double-bumps, which monotonicity tolerates). `demandLivePage` captures it at press time alongside the existing guards, and the completion goes inert when the route changed in between.

The regression test exercises the exact scenario the pane/ref guards cannot see: a round trip to `/` (welcome opens in secondary; `openPane` never displaces a non-welcome main) and back to the session re-matches the still-open main pane, restoring an identical pathname, pane id, and focused ref — only the epoch knows the user left. Verified RED with the epoch check disabled, GREEN with it; the test also confirms the chord is not bricked afterward (the same press then takes the direct step to the now-loaded page).

### Low — docs

`docs/web-ui/keybindings.md`'s tree-keyboard note now says the desktop rail's Tree releases only Alt-held arrows and Alt+Home/End, keeps Ctrl/Meta-held arrows tree-owned (no global chord binds them), and on mobile keeps every arrow tree-owned so Alt+ArrowLeft/Right never fall through to history navigation.

### Gates

`biome check` on the touched files clean; `tsc --noEmit --incremental false` clean; `vitest run src/keybindings src/shell src/widgets src/panes/settings/sections` — 186 files, 2929 tests, all passing (including the 13 demand-lifecycle and 31 tree-widget tests).

obra:
## RoboRev round 10 — the one medium finding fixed in `a067cb06e`

**Medium (TUI: stale live-nav read can revert the user's session selection)** — confirmed, and the race is real through both halves: the resub the sequence-mismatch branch launched targeted the session currently *displayed* — the older one — so (a) its own `ThreadRead` re-pointed the server-side subscription at the older session, and (b) its untagged response (`liveNavSeq` 0) arrived after the newer cycling read applied and was processed as an ordinary session entry, reverting the UI to the older session.

The fix takes your first option: while a newer live-nav read is still pending (`liveNavPendingRef` set), the stale-read drop returns no command — the pending read owns the subscription when it lands, so there is nothing to re-establish. The settled-display case is unchanged: once the pending target has been consumed, a stale read still re-issues `fetchHubSession` for the displayed session (round-8 medium 2's `TestHubSessionLiveCycleStaleReadReEstablishesCurrentSubscription` stays green).

Test: `TestHubSessionLiveCycleStaleReadDoesNotResubWhileNewerReadPending` — two presses, the stale read lands while the newer is pending and must return no command, the newer read then applies and the UI stays on it, and the recorded `thread/read` calls are exactly the two cycling reads (no resub was issued). RED confirmed first: without the guard, the drop launched the resub read for the older displayed session.

Verification: full TUI package green (`go vet`, `gofmt`, `go test -count=1 .`).

obra:
## Round 11 response — 3 findings (2 Medium, 1 Low)

All three findings addressed. Tests were written RED first (each confirmed failing against the pre-fix behavior) and land with the fixes.

### Medium 1 — `AppShell.tsx:603-605` — discarded later press against a stale in-flight demand

Valid. The set-based dedupe conflated "a live press owns this in-flight load" with "the page/direction key is occupied," so a same-direction press after a leave-and-return round trip silently deduped against a demand whose guards had gone stale — and that demand's completion was then dropped by the round-9 guards, leaving the keypress with no navigation at all.

Fix: `demandedLivePages` is now an owner record (`Map<key, LiveDemandOwner>`) instead of a set, separating ownership from dedupe:

- **Press-time staleness check** — on a same-key press, `liveDemandOwnerFresh(owner)` compares the recorded owner's client generation, route epoch, focused pane, and focused ref against the current world. Fresh → dedupe as before (same intent, no counter bump, per round-4 medium 2). Stale → the press **adopts** the pending load: it bumps the intent counter and rebinds the record to fresh guards. The revalidator already dedupes the concurrent read of the same page, so adoption issues no second request.
- **Completion-time ownership identity** — the load's completion checks `demandedLivePages.get(demandKey) !== owner` first and no-ops when displaced by an adopter; the error and catch paths guard on the same identity. The adopting completion's inertness checks (mounted, intent, generation, epoch, pane, ref, palette/modal) run against the adopter's guards.

New test: `a second press adopts an in-flight demand whose guards went stale` — press one with the page-two read deferred, round-trip `/` → back (identical pane/ref; only the route epoch moves), press two, then resolve. Confirmed RED against the set-style dedupe (the later press produced no navigation) and GREEN with the fix (the adopting completion navigates to `/s/local%3Alive-b`).

### Medium 2 — `hub_update.go:99-124` — dropped reads never restore the displayed session's subscription

Valid, and the round-10 pending-early-return made it worse in the failure case: a pending read that later fails or is dropped leaves the connection subscribed to an abandoned target while the displayed session stops receiving live updates.

Fix: both drop branches (draft guard, overlay guard) and the failure path of a superseded pending read now call a new `reestablishDisplayedSubscription()` helper (`hub_livecycle.go`): it re-issues `fetchHubSession` for the displayed ref with `ReplaceSubscription: true` (which server-side culls the abandoned main and child subscriptions), clears `watchedChildRefs`, and re-arms child subscriptions via `subscribeNewChildren()` — preserving child subs as the finding asked. The drop branches clear `liveNavPendingRef` first so the resub read can't be mistaken for a newer navigation read; the failure path has already cleared the pending ref above.

Tests: `TestHubSessionLiveCycleDropsReadWhenDraftAppearedMidFlight` and `...OverlayOpenedMidFlight` extended with resub assertions — reads are now recorded with a `replace` flag, and the tests assert the drop path issues a replacing read for the displayed session and re-subscribes its running children (a subagent-run child re-arms with `replace: false`). Confirmed RED first ("draft drop returned no command").

### Low — `tree/index.tsx:189-194` — Alt-stacked-on-Ctrl/Meta arrows release as dead keys

Valid. The release guard keyed on `event.altKey` alone, so Alt+Ctrl / Alt+Meta arrows were released by the tree yet matched no global chord — dead keys.

Fix: the guard now requires the Alt-only combinations the global chords actually bind (`event.altKey && !event.ctrlKey && !event.metaKey`); arrows holding any other modifier stay tree-owned, matching the surrounding comments. `docs/web-ui/keybindings.md`'s tree-keyboard note updated to say "Alt-only" instead of "Alt-held."

New test: `arrows with Alt stacked on Ctrl or Meta still navigate the tree` in `tree.test.tsx`.

### Gates

- `npx tsc --noEmit` clean; biome clean on all touched files.
- `npx vitest run src/keybindings src/shell src/widgets src/panes/settings/sections` green.
- `go vet ./...`, `gofmt -l .` clean, `go test -count=1 .` green in `cmd/evener-tui`.

obra:
## Round 12 response — 2 findings (2 Medium)

Both findings address the recovery path round 11 introduced. Tests were written RED first (each confirmed failing against the pre-fix behavior) and land with the fixes.

### Medium 1 — `hub_livecycle.go:111-113` — batched resub races child subscriptions

Valid. The replacing parent read and the additive child re-subscriptions went out in one `tea.Batch`, which carries no ordering guarantee: a replacing read executing after a child subscription culls it server-side, and `watchedChildRefs` already marks it, so nothing re-arms it — the child's live activity stays dead.

Fix: child re-arm no longer issues at drop time. The recovery read is issued alone; when its **response** is processed, the replacement has completed server-side, and `subscribeNewChildren()` issues from there. The round-11 drop tests were re-scoped to the new contract (the resub is a single tagged read; the child read lands after the recovery response, additive).

New test: `TestHubSessionLiveCycleResubSequencesChildrenAfterParent` — the recovery response (whose thread carries a running delegate child) must sequence the child re-arm after the parent read. RED confirmed ("resub still batches child subscriptions with the replacing read").

### Medium 2 — `hub_livecycle.go:111` — untagged recovery read can hijack navigation

Valid. The recovery read used the ordinary `fetchHubSession`, so its response arrived as an untagged `hubSessionMsg` and was processed as an ordinary session entry: a late response after the user navigated away reverted the UI to the previously displayed session.

Fix: the recovery read is now tagged (`liveNavSeq` + a new `liveNavRecovery` flag, set by the command's closure). Update handles it before the ordinary cycling-drop guards:

- **Stale** (seq moved, or the displayed ref changed): discarded outright — no hijack, no state touched.
- **Current**: applies as a same-ref resync — transcript refresh, escalation re-surface, queue state, `refreshViewport` — then issues the sequenced child re-arm. The draft/overlay guards intentionally do **not** apply: the recovery read was issued precisely because a draft or overlay dropped the cycling read, and re-entering those guards would loop recovery forever. A failed recovery read retries once (the culled subscriptions still need restoring).

New test: `TestHubSessionLiveCycleRecoveryReadDoesNotHijackNavigation` — draft drop, recovery read issued, draft cleared, newer press applied (UI on 01C), then the stale recovery response must not revert the display. RED confirmed ("stale recovery response hijacked navigation").

### Gates

- `go vet ./...`, `gofmt -l .` clean, `go test -count=1 .` green in `cmd/evener-tui` (full package).
- All 21 live-cycle tests green, including the two round-11 drop tests re-scoped to the new sequencing contract.
- No web files touched this round; round-11's frontend gates stand.

obra:
## Round 13 response — 6 findings (3 Medium, 3 Low)

All six findings addressed. Tests were written RED first (each confirmed failing against the pre-fix behavior) and land with the fixes.

### Medium 1 — untagged resync reads allow stale navigation

Valid. Two paths still read the session without the recovery tag round 12 introduced: the settled-display resub after a stale read dropped, and the reconnect resync. A late response on either lands as an untagged `hubSessionMsg` and is processed as an ordinary session entry — it can revert the UI after a newer live-nav response applied and re-point the subscription at the older session.

Fix: both paths now route through the tagged recovery helper, so their responses carry `liveNavSeq` + `liveNavRecovery` and drop as stale when the intent has moved on. The settled-display resub uses the replacing `reestablishDisplayedSubscription()`; the reconnect uses a new additive variant `resubscribeDisplayedSessionAfterReconnect()` — the replacement connection carries no subscriptions, so additive preserves the reconnect path's long-standing read contract (nothing to cull) while gaining the tag. Both fold the helper's value-receiver copy back through the pointer, and both re-arm children sequenced after the response per round-12's contract (the old `resyncHubSession` + batched `subscribeNewChildren` race is gone with it).

Tests: `TestHubSessionLiveCycleStaleReadReEstablishesCurrentSubscription` extended (the resub must be a tagged recovery read); new `TestHubSessionLiveCycleReconnectResyncIsTaggedRecovery` (the reconnect batch's session read must carry the recovery tag — exercised via lazy BatchMsg children with a bounded collector). Both RED first.

### Medium 2 — recovery retry had no backoff or limit

Valid. My round-12 "retry once more" comment lied — a failed recovery read unconditionally started another, so a persistent RPC failure spun a tight command loop.

Fix: `liveNavRecoveryRetries` on the model, bounded at `hubLiveNavRecoveryMaxRetries` (3). Retries increment past the helper's reset; at the cap the failure surfaces as the model's error (`m.err`) and the loop stops. Success resets the counter; a fresh drop-driven recovery resets it too. `liveCycleReads` gained a `failRef` so the fixture can fail one session's reads.

Test: `TestHubSessionLiveCycleRecoveryRetryIsBounded` — drives the retry loop against an always-failing read; RED first ("recovery retried 10 times without bound"), now ends at the cap with `m.err` set and no further reads issued.

### Medium 3 — recovery branch discarded pre-snapshot notifications

Valid. The recovery branch returned before the common `msg.beforeCut` fold, so notifications the connection delivered ahead of the recovery read (escalations, resync requests, hub updates) were silently lost.

Fix: the recovery apply folds `beforeCut` through `applyHubNotification` before the snapshot — the same order the ordinary entry path uses — and batches the resulting commands with the child re-arm.

Test: `TestHubSessionLiveCycleFoldsBeforeCut` — a sandbox-escalation notification riding the recovery response's `beforeCut` must land in `escalationsByRef`; RED first (map was empty).

### Low 4 — Alt+Shift+Arrow unreachable in transcript view

Valid. The `transcriptView` early return handled only esc/i/q and passed everything else to the viewport, swallowing the live-nav chords while a child transcript was open.

Fix: the chord cases are routed before the transcriptView early return (after pickers/modals, which keep their own keys).

Test: `TestHubSessionLiveCycleChordWorksInTranscriptView` — RED first ("transcript view: expected a live-nav fetch command, got none").

### Low 5 — `demandLivePage` never re-checks editable focus at completion

Valid. The chord is suppressed while an editable target has focus at press time, but a demand pressed from a non-editable target navigated on completion even if the user had focused the composer mid-flight — surprising a typing user.

Fix: the completion's inertness guard now includes `isEditableTarget(document.activeElement)` — the completion-time half of the press-time suppression. Goes inert: the in-flight key is deleted and no navigation fires.

Test: `a demand that completes while the composer has focus goes inert` — RED first (navigated to live-b), now stays on live-a.

### Low 6 — Alt+Shift modifier comment/release mismatch in tree widget

Valid. The guard released Alt+Shift+Home/End, for which no global chord exists — dead keys. (Along the way I found the tree never had Home/End handlers at all, so the documented "plain Home/End stay tree-owned" was also false.)

Fix: the release is now exact-chord — arrows release with Alt alone or Alt+Shift (the live-session chord); Home/End release only on plain Alt. The tree gained real Home/End navigation (focus first/last visible row, preventDefault), so plain Home/End are genuinely tree-owned and Alt+Shift+Home/End navigate instead of dead-keying. Comment and `docs/web-ui/keybindings.md` updated.

Tests: new `Home and End with Alt+Shift stay tree-owned while Alt+Shift arrows release`; the round-9 release test updated to the exact-chord contract (its Alt+Shift+Home assertion had encoded the dead-key release). 33/33 tree tests green.

### Gates

- TUI: `go vet`, `gofmt -l`, `golangci-lint` clean; `go test -count=1 .` green (all 23 live-cycle tests, 26 session-key tests).
- Web: `tsc --noEmit` clean; biome clean on touched files; vitest suites green.

obra:
## Round 14 response — 3 findings (1 Medium, 2 Low)

All three findings addressed. Tests were written RED first (each confirmed failing against the pre-fix behavior) and land with the fixes.

### Medium 1 — stale recovery responses discard subscription reconciliation

Valid. A stale-dropped recovery response's `ThreadRead` still replaced the server-side subscription, and since concurrent RPCs have no ordering guarantee, that replacement can land after the newer navigation's read — leaving the subscription on the older session while the UI shows the new one, stopping live updates.

Fix: the stale-recovery drop now reconciles. When no newer read is in flight (`liveNavPendingRef` empty — a newer in-flight read owns the subscription and performs its own replacement on landing), the drop re-establishes the displayed session's subscription through the same tagged recovery helper, so the reconcile read itself cannot hijack.

Test: `TestHubSessionLiveCycleStaleRecoveryDropReEstablishesSubscription` — recovery read issued, newer press applied, stale recovery response lands; the drop must issue a tagged reconcile read for the displayed session. RED first ("stale recovery drop returned no command").

### Low 1 — live-navigation chords bypass the `forkDraft` handler

Valid. The early alt+shift dispatch (round 13's transcript-view fix) ran before the forkDraft handling, and with the prefilled fork text deleted the composer is empty — the draft guard passes, a chord switches sessions, and the session-entry path clears `forkDraft`, silently discarding the active fork target and original text.

Fix: a non-nil `forkDraft` is now a hold state in `switchToAdjacentLiveSession`'s guard (the seam every chord dispatch site — early global, browse, palette — shares), with a system message mirroring the draft hold's.

Test: `TestHubSessionLiveCycleChordHoldsWithForkDraft` — RED first ("expected no live-nav command while a fork draft is active").

### Low 2 — early alt+shift handling creates unreachable dead code

Valid. The identical `alt+shift+right` / `alt+shift+left` cases in the `scrollMode` switch and the main key switch became unreachable once round 13's early global dispatch landed.

Fix: both duplicate case groups removed; each site's comment now names the early global switch as the single dispatch point (the scroll-mode site's comment retains the round-4 rationale — the chords stay live in browse and scroll modes by construction, not by per-mode cases).

### Gates

- `go vet`, `gofmt`, `golangci-lint` clean; `go test -count=1 .` green in `cmd/evener-tui` (25 live-cycle tests).
- No web files touched this round; round-13's frontend gates stand.

obra:
## Round 15 response — 2 findings (2 Medium)

Both findings address gaps my round-14 fixes left. Tests were written RED first (each confirmed failing against the pre-fix behavior) and land with the fixes.

### Medium 1 — stale live-navigation recovery can leave the TUI subscribed to the wrong session

Valid. Round 14's reconcile only fired when no newer read was pending. A stale recovery response dropped while a newer read was still in flight recorded nothing — the newer response then applied and cleared the pending state, and nothing later reconciled: the stale replacement could have landed after the newer read's own request, leaving the subscription on the older session forever.

Fix: a `liveNavNeedsReconcile` flag on the model. A stale recovery drop while a newer read is pending sets it (the newer read owns the subscription while in flight). When the newest navigation settles — its response applies through the session-switch path — the settle point issues the deferred tagged recovery for the displayed session, batched with the child re-arm. The flag clears when the reconcile issues, on any direct re-establish (the helper's own read is the reconcile), and on reconnect (whose own tagged resub re-establishes).

Test: `TestHubSessionLiveCycleReconcilesAfterPendingNavigationSettles` — recovery read issued, newer press pending, stale recovery drops (pending preserved), newer response applies; the settle must issue a tagged recovery read for the displayed session. RED first ("settle point returned no command").

### Medium 2 — fork-draft mid-flight re-check is missing

Valid. The press-time hold covered `forkDraft != nil`, but the apply-time re-check tested only input text and attachments — a fork draft created while the cycling read was in flight was applied over and then cleared by the session-entry path, silently discarding the fork target and original text.

Fix: `m.forkDraft != nil` joins the mid-flight guard alongside the input and attachment checks: the read drops, the pending state clears, and the displayed session's subscription re-establishes through the tagged helper (same treatment as the draft drop).

Test: `TestHubSessionLiveCycleDropsReadWhenForkDraftAppearedMidFlight` — RED first ("live-nav read applied over a mid-flight fork draft"), now drops with the fork draft preserved and a tagged recovery issued.

### Gates

- `go vet`, `gofmt`, `golangci-lint` clean; `go test -count=1 .` green in `cmd/evener-tui` (27 live-cycle tests).
- No web files touched this round; round-13's frontend gates stand.

obra:
## Round 16 response — 1 finding (1 Medium)

Test written RED first (confirmed failing against the pre-fix behavior) and landed with the fix.

### Medium 1 — same-session early return can drop child subscriptions during rapid live-navigation wrap (e.g. A → B → A)

Valid. A rapid wrap makes the newest live-nav read's response land on the same-ref branch, but that branch neither re-armed the child subscriptions the read's `ReplaceSubscription` had culled server-side nor processed the deferred reconcile — and `watchedChildRefs` still marked the children, so even the ordinary entry path would have skipped a re-arm. Child activity updates stop arriving while the UI stays on the wrapped-to session.

Fix: scoped to `msg.liveNavSeq > 0` (ordinary same-session resyncs and status refreshes are untouched), the same-ref branch now clears `watchedChildRefs` and issues `subscribeNewChildren()` after the resync applies, and a set `liveNavNeedsReconcile` flag is processed there too — the same-ref apply is a settle point exactly like the session-switch one, so the deferred tagged recovery issues (batched with the child re-arm).

Tests:
- `TestHubSessionLiveCycleWrapReArmsChildSubscriptions` — A→B→A wrap; the same-ref apply must re-issue the culled child's subscription. RED first ("same-ref live-nav apply returned no command").
- `TestHubSessionLiveCycleSameRefSettleProcessesReconcileFlag` — a stale recovery dropped while the newest read was pending sets the flag; the same-ref settle must issue the tagged reconcile. RED first ("same-ref settle returned no command").

### Gates

- `go vet`, `gofmt`, `golangci-lint` clean; `go test -count=1 .` green in `cmd/evener-tui` (29 live-cycle tests).
- No web files touched this round; round-13's frontend gates stand.

obra:
## Round 17 response — 2 findings (2 Medium)

Both findings address races my round-16 code reintroduced. Tests were written RED first (each confirmed failing against the pre-fix behavior) and land with the fixes.

### Medium 1 — child subscription reads batched concurrently with the replacing recovery

Valid. At both settle points, the deferred reconcile's replacing `thread/read` was `tea.Batch`'d with `subscribeNewChildren()` — concurrent RPCs with no ordering guarantee, so the replacing read could cull the child subscriptions right after they were created. This was exactly the round-12 race, reintroduced at the settle points.

Fix: when the reconcile issues, the recovery path owns the re-arm — the batched `subscribeNewChildren` is dropped and the recovery response's own apply re-arms the children (the helper already cleared `watchedChildRefs`, so the re-arm is complete). The non-reconcile same-ref path keeps its direct child re-arm (no replacing read races it there).

Test: `TestHubSessionLiveCycleSettleSequencesChildrenAfterReconcile` — a deferred reconcile at the same-ref settle; no child read may issue concurrently with the reconcile read, and the child re-arm follows the reconcile response. RED first ("child subscription issued concurrently with the reconcile read").

### Medium 2 — same-ref early return omitted the preCut commands

Valid. The same-ref branch returned without the collected `preCut` commands: notifications delivered ahead of the response had been applied to model state, but their follow-up reads (a status-change notification's state refresh, resyncs) were discarded, leaving the TUI with stale transcript/session state.

Fix: all three same-ref return paths now carry `preCut` — including the reconcile branch (the follow-up reads are non-subscribing, so they cannot race the reconcile's replacement).

Test: `TestHubSessionLiveCycleSameRefSettleKeepsPreCut` — a wrap whose same-ref response carries a status-change notification in `beforeCut`; the follow-up state-refresh read must issue from the settle's commands. RED proven by reverting the preCut (test FAILs without it).

### Gates

- `go vet`, `gofmt`, `golangci-lint` clean; `go test -count=1 .` green in `cmd/evener-tui` (31 live-cycle tests).
- No web files touched this round; round-13's frontend gates stand.

obra:
## roborev: Combined Review (`e9a88c3`)

## Summary

Adds Alt+Shift+Arrow live-session navigation across web UI and TUI, but leaves several newly reachable session-scoped async responses without navigation guards, plus one stale doc line.

## Findings

### Medium

- **`cmd/evener-tui/hub_session_keys.go:192-201`, `cmd/evener-tui/hub_update.go:767-783`** — Live-session navigation is now allowed while a child transcript request is in flight, but `hubTranscriptMsg` has no originating parent-session guard and always installs its result. If the user navigates from session B to session C before B's transcript response arrives, the late response can attach B's child transcript to C, enable browse mode, and blur C's composer. Include the originating session reference or navigation epoch in transcript target/read messages and discard responses that no longer match the displayed session; apply the same guard to `hubTranscriptTargetsMsg`.

- **`cmd/evener-tui/hub_session_keys.go:192-201`, `cmd/evener-tui/hub_update.go:585-595`** — The new chord can run while `/status` is fetching, but `hubStatusMsg` does not carry or validate the session it was requested for. A delayed status response for session B can overwrite `m.detail` and open a status panel for B after navigation has already switched the UI to session C. Tag session-scoped status requests with the originating ref or a navigation generation and ignore responses when the current mode/ref no longer matches.

### Low

- **`docs/web-ui/keybindings.md:233`** — States a single live session "cycles onto itself, which the same-URL navigation no-ops", but `liveSessionCycle.ts` returns `null` for the single-item case specifically to skip `openLiveSession` entirely and avoid stealing focus via the URL-equal refocus path. Update the bullet to state a single live session is a no-op that issues no navigation or refocus, matching `adjacentLiveSessionRef` and `liveSessionCycle.test.ts`.

---
*Reviewers: 2 done | Synthesis: codex, 10s | Total: 11m49s*

obra:
Round 19 wrap-up:

- **roborev review on `dfde07bef`: zero findings** — status "Review complete" (success) with no body update, the first clean review in 19 rounds.
- **Branch updated with main** (merge commit `7f6b73ad8`): the PR was BEHIND, which strict branch protection blocks. Clean merge — no file overlap with the 4 new main commits; all 13 CI checks pass on the merge commit.
- **roborev review on `7f6b73ad8`: also zero findings** (same clean signature, "Review complete" at 13:59:16Z, body unchanged).

The live-session-keys feature (go to next/previous live session in TUI + webui) has now been through 19 review rounds and 53 findings fixed. CI is fully green and the branch is up to date — the remaining gate is the required approving review.

PR #1046, opened 2026-09-09
Bury session force stop in the session menu

obra:
The session footer rendered an always-visible **Force stop…** button above the composer for every local non-closed session — healthy ones included. Force stop is a destructive recovery action; it belongs in the session **⋯** menu, which already had the item wired.

## Changes

- **Session.tsx** — remove the footer `SessionForceStopRecovery` render. The component stays for the "Loading transcript…" empty state, where an unresponsive session has no composer and therefore no menu, so that recovery path is preserved.
- **SessionChrome.tsx** — widen the menu's `onForceStop` gate to any local non-closed session so nothing the inline button reached is lost: saved `notLoaded` panes (a pending or failed resume can still own a daemon) and panes without navigation identity keep force stop. A session retained by its owning session remains excluded — its notice directs recovery to the owner.
- **Session.test.tsx** — drive the seven affected tests through the real menu path (session actions → Force stop… → dialog). The retained-child test now asserts the menu itself excludes an owner-retained session; the failed-initial-read tests still cover the loading-state button.

The rail row menu's gating is unchanged.

## Verification

- `make test-web` — PASS (web-typecheck, web-test, web-lint)
- Affected suites: Session.test.tsx, SessionChrome.edge.test.tsx, ForceStopDialog.test.tsx — 93/93 pass

### DIFF
--- a/cmd/evener-hub/frontend/src/panes/session/Session.test.tsx
+++ b/cmd/evener-hub/frontend/src/panes/session/Session.test.tsx
@@ -95,9 +95,22 @@ function stubSessionSlots(): void {
 stubSessionSlots();
 
 afterAll(() => {
+  // A test that restored a slot mid-file (and, under test.each, has no
+  // onTestFinished to re-stub it) leaves vi.mocked(...).mockRestore absent;
+  // re-stub first so handing the real components back can't fail.
+  stubSessionSlots();
   vi.mocked(ComposerModule.Composer).mockRestore();
   vi.mocked(SessionChromeModule.SessionChrome).mockRestore();
 });
+
+// Force stop lives in the session "⋯" menu (SessionChrome) now that the inline
+// footer button is retired; this walks the same menu path a user would. Tests
+// using it must restore the real SessionChrome and render it (or the real
+// Composer, which mounts it) themselves.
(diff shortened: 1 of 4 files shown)


obra:
## RoboRev Medium — rejected after tracing to original design

**Finding:** `forceStopAction` offers Force stop for nested subagent/fork sessions, which the hub rejects with "no direct daemon ownership claim".

**Assessment:** Traced to the original force-stop commit (`e9d42f269`). The retired inline footer button's gate was `local: && !recoveryOwnerRef && !closed` — **no parentRef check**. Nested forks were always offered force stop, and the existing test `independent nested fork keeps confirmed recovery during stalled reads` pins exactly that behavior. The hub's rejection is the *accepted* error path the dialog surfaces for a descendant whose owning session is unresponsive — not a regression introduced by this PR.

**Commit:** `be724f53` adds a test confirming a nested session still sees Force stop and that confirming surfaces the hub's rejection (toast + dialog body), so a later narrowing cannot silently drop the descendant recovery path.

**Verification:** `make test-web` PASS (typecheck, test, lint); SessionChrome.edge.test.tsx 12/12, Session.test.tsx 82/82.

PR #60, opened 2026-07-20
webui: fork any message into the composer for editing (#42)

obra:
Closes #42.

## Summary

Every message in the hub web UI transcript now carries a quiet `⎇ fork` affordance. Activating it forks the session at the point where that message was entered — using the underlying serf fork machinery — and pre-populates the forked session's composer with the original message text, ready for editing and explicit submission. The forked session does **not** auto-run the message.

## Design

**Fork machinery (found + extended).** The base already had `agent.ForkSession` (copies the transcript prefix and *appends* a replacement user turn, so the child auto-runs it) plus the hub `thread/fork` RPC and REST `/s/<id>/fork`. What was missing for this issue was a fork that stops at the prefix and hands the original text back:

- `agent.ForkSessionAtUserTurn` — a `deferInput` variant of the shared fork internals: the child transcript holds only the entries before the divergence turn (still required to be a `USER_INPUT` entry), lineage meta (`ParentSessionID`/`DivergenceTurn`/parent `ForkLabel`) is identical, and the diverging message's original text is returned to the caller.
- Wire: `ThreadForkParams.deferInput` + `ThreadForkResponse.originalInput` (appwire; `deferInput` and `editedInput` are mutually exclusive, empty `editedInput` still errors as before — the old edit-fork contract is unchanged), and `defer_input`/`original_input` on the REST endpoint (snake_case per the naming gate).

**Fork-point semantics.** Forking a user message rewinds to just before it (child = entries `[1..N-1]`) and stages its text. Forking an *assistant* message resolves to the user prompt that produced it (retry semantics): the composer is staged with that prompt, satisfying "populating the input box with the original message" uniformly — the divergence point must be a `USER_INPUT` turn, and a user cannot re-enter an assistant message.

**Edit-before-submit flow.** `startFork` sends `thread/fork` with `deferInput` (REST fallback identical), stages the server-authoritative `originalInput` as the fork child's sticky per-session composer draft (`SerfDrafts.writeFor`), refreshes the sidebar, and navigates to the fork. The existing `SerfDrafts.bind` path restores the draft into the composer on arrival; nothing runs until the user submits.

**Live-echo indexing.** The renderer tracks the authoritative transcript entry index per user message (server echo corrects the local echo's inferred index in `promoteLocalUserMessage`); the hub rejects any stale/non-`USER_INPUT` divergence turn cleanly, so no silent wrong forks.

[...]

PR #1070, opened 2026-09-09
Shared session notes (human/agent whiteboards + URL list)

obra:
End-to-end shared notes facility: persisted human/agent whiteboards and a session URL list, surfaced in the hub web UI (standalone Notes panel with blur-save editor), TUI drawer, daemon RPCs with projector pushes, agent tools, and wire catalog + codegen.

Branch scope (34 commits on shared-notes):
- Persisted notes model, events/kind/wire types, session notes store
- Agent tools (agent note, URL add/remove, notes read)
- Daemon RPCs + projector pushes, wire catalog + codegen
- Hub relay + capability + past sessions + web UI (Details section, then moved to standalone Notes panel with blur-save human editor)
- TUI drawer + commands, e2e scenarios, tombstone/retry/delivery refinements

Verification: npm test green (422 files / 9228 vitest + 57 node), tsc clean, Biome clean on touched files, live e2e screenshot of Notes pane + blur-save round trip.


obra:
RoboRev round 2 (`6d63536`) addressed in `b0adbf8f0` (plus `6d6353606` for round 1):

**Round 1 (all 7, in `6d6353606`)**
- Save coalescing, session-keyed flush snapshots, live→ended flush, single `requestSave` path — `NotesPanel.tsx`, 3 new vitest cases.
- Empty-clear tombstone via `NotesStoredValueSet` (legacy flagless non-empty tombstones still recognized for upgrade safety) + superseded-clear regression test.
- Scheme-bypass rejection (`javascript:`, `data:`, `mailto:`…) + userinfo rejection, each with regression tests.

**Round 2 (all 3, in `b0adbf8f0`)**
- Single shared slot → per-session queue (`Map<ref, note>` drained FIFO); `Saved` paints via last-saved ref so a B-save drained in an A-started loop confirms for B. New test pins drain order + per-thread commits.
- Failed saves keep the entry queued; the next `requestSave` retries in queue order instead of discarding a parked newer draft. New test pins failure→retype→blur→lands-newest.
- `snapRef` entry deleted in the switch/unmount cleanup after flushing.

Not run locally (per operator instruction — low battery); CI is the prover for the new suites. New tests use the suite's proven `user.tab()` blur pattern.

obra:
Round-2 RoboRev findings addressed in ea7ca9e3c (pushed, CI running — local test runs skipped per operator constraint, CI is the prover here):

- **Projection race** (`agent/session_notes_rpc.go:maybeAppendNotesContext`): render-compare-record now runs under `notesUpdateMu`, so a concurrent note/URL commit cannot land a stale snapshot after a newer projection. `s.mu` is released across `appendTurn` to preserve the mu-before-history order; the last-projected record is set first so a concurrent attempt stays silent instead of double-appending.
- **Consumed-intent durability** (`consumePendingNotesHumanDurable`): an adopted metadata intent's spend persists its absence in the same save (with restore-on-failure), so a crash cannot leave it beside an in-flight record for a post-restart same-value mutation to adopt twice. Regression test: `agent/session_notes_adopt_spend_test.go`.
- **Queue drain past failure** (`NotesPanel.tsx` save loop): only the failed entry requeues; queued siblings still persist and report, Saved suppressed for the failed session only. Regression test: B-save behind failing A-save.

Notes: the prior review's tombstone-flag / scheme-bypass / userinfo findings were already addressed in 6d6353606 (verified present with tests). Awaiting CI + re-review.

obra:
Status check on the empty CI rollup: no workflow run exists for this PR's head (pull_request-event runs list shows only other branches; commit-level check-runs for ea7ca9e3c return 0). The `pull_request/branches:[main]` trigger should have fired on push, so this looks like a GitHub-side delivery gap rather than a code failure. If no run appears, a re-push or closing/reopening may be needed to retrigger. Also noting `mergeable_state: dirty` — main has moved (appwire handshake/client changes touching threads.ts/fakeClient.ts, additive so far); will rebase once CI is running.

obra:
Round-3 RoboRev findings (on ea7ca9e3c) addressed in c98296572 (pushed; local test runs skipped per operator constraint, CI is the prover):

- **Mutex leak** (`maybeAppendNotesContext`): the round-2 edit re-acquired `s.mu` after `appendTurn` and never released it — first successful projection would wedge the session mutex permanently. Now unlocks before the append (preserving the mu-before-history order); both findings confirmed valid against the code.
- **Per-session outcomes** (`NotesPanel.tsx` save loop): successes accumulate in a set, failures map per ref, and a failure requeues only when the queue holds no newer draft for that session. A stale A-failure can no longer overwrite a newer A-success on retry, and an A-success + B-failure still reports Saved for A. Regression tests for both halves.

Also verified: no new verdict beyond this round; awaiting CI + re-review.

obra:
Round-4 RoboRev findings (on c982965) addressed in 6f0543e81 (pushed; local test runs skipped per operator constraint, CI is the prover):

- **Retry storm** (`NotesPanel.tsx`): failures no longer requeue into the same drain — one attempt, then settle; retry on the next explicit save with the latest draft. Regression test pins single-attempt settle + draft kept.
- **Intent-before-steer** (`deliverAdoptedNotesSteer`): adopted intent consumed before steer acceptance, so a consume failure cannot resurrect an intent for an already-landed steer. Finding confirmed valid.
- **Link trust** (Links section): web links now show destination URL beside the label, matching file-link/TUI shape. Phishing-label test added.
- **`clear` escape** (`runHubNotes`): bare `clear` still clears; `"clear"` sets the literal word; other quoting verbatim. Regression test added.

Awaiting CI + re-review.

obra:
Round-5 RoboRev findings (on 6f0543e) addressed in 0299412db (pushed; local test runs skipped per operator constraint, CI is the prover):

- **Adopt-then-refuse race (High, valid — my round-4 order was wrong):** consume-first loses delivery when the steer is refused. Intent now stays pending until the steer accepts; the durable spend still covers the post-accept crash window. Regression test: refused adoption retries and steers exactly once.
- **File links (Medium, valid):** `file:` URLs render with the open-beside affordance in scope (visible text retained; none out of scope). Tests for both.
- **Empty hostname (Low, valid):** gate on `Hostname()` with a rejection test.
- **Labelless double-render (Low, valid):** truthy label check; test pins single render.
- **Clear case regression (Low, valid — mine):** bare `clear` case-insensitive again; quoted escape preserves inner case. Tests for variants.

Awaiting CI + re-review.

obra:
Round-6 RoboRev findings (on 0299412) — frontend half addressed in 8bb88d5c0 (pushed; local test runs skipped per operator constraint, CI is the prover):

- **Lost revert-during-flight (Medium, valid):** queue holds the latest draft whenever a loop is active, even store-equal; equality runs only at drain time. Regression test pins stored-A/B-pending/revert-to-A.
- **Watched-thread store (Medium, valid):** `storedNote` reads both maps like `trackedThreadModel`.
- **Multiline normalization (Medium, valid):** client-side `normalizeNote` mirrors the daemon collapse+1000-rune clamp; payload, dedup, and Saved guard compare normalized. Regression test pins Saved + no redundant RPC.

Still to assess: the two backend findings (adopter-lease spend-failure duplicate steering; ignored NotesDeliveryPending clear error) and the two lows (fileURLToPath fail-open; canonicalFilePath empty-cwd scope skip). Following in the next commit.

obra:
Round-6 continued (backend half) in 899e84a83 + b45c411dc (pushed; local test runs skipped per operator constraint, CI is the prover):

- **Adopted-delivery idempotence (High, valid):** the adopter links the adopted intent to its accepted steer before clearing/spending; a same-ID retry resumes at the link and finishes the marker-clear + spend without re-accepting the steer. `clearNotesDeliveryPending` errors now route through that recovery instead of being ignored. Refusals still leave the intent pending. Regression test: spend-failure retry finishes with exactly one steer.
- Note: b45c411dc repairs a one-line header dropped by an edit collision in the previous commit (file parses; gofmt clean).

Frontend half (queue equality, watched store, normalization) went out earlier in 8bb88d5c0. Awaiting CI + re-review.

obra:
Round-7 RoboRev findings (on b45c411) addressed in a1a8926fe (pushed; local test runs skipped per operator constraint, CI is the prover):

- **Supersede clear error (Medium, valid):** the fresh-ID supersede path now returns the adopted clear's journal error (releasing this reservation for a hub retry) instead of persisting over a still-pending record.
- **Adopted-value clear (Medium, valid):** the link-branch retry clears against the adopted stored value from journal/intent, not the caller's raw input — whitespace/over-length variants can no longer strand the marker.
- **Trim-once ID (Low, valid):** urls/remove trims the entry id once for validate/remove/report.
- **File query/fragment + escaped separators (Low, valid):** rejected / resolved-from-escaped-then-unescaped-after-scope, with tests for collapse, traversal, and absolute-without-cwd fail-closed.

Awaiting CI + re-review.

obra:
Round-8 RoboRev findings (on a1a8926) — backend half in 7bc6da768 (pushed; local test runs skipped per operator constraint, CI is the prover):

- **Metadata rollback race (Medium, valid):** all six notes mutation+rollback sites (human RPC, human/agent serialized mutations, URL add/remove, adopted-intent spend) now hold `metaSaveMu` from the save attempt through the in-memory restore, via new `autoSaveMetaLocked`/`persistNotesMetaLocked`/`persistNotesMetaWithIntentLocked` variants. A concurrent `maybeAutoSave` can no longer persist a transient mutation that the RPC then reports as failed.
- **Scope-check comment (Low, valid — comment-only):** corrected to the actual decode-before-scope order with the must-stay-that-way rationale.

Remaining from round-8: the frontend lost-revert finding was already fixed in 8bb88d5c0 (queue parks even store-equal drafts while a loop runs, equality at drain time, with a dedicated revert regression test) — that verdict reviewed an older head. Awaiting CI + re-review.

obra:
Round-9 RoboRev finding (on 7bc6da7) already addressed: the save-in-flight revert race was fixed in 8bb88d5c0, which is an ancestor of the reviewed head (8bb88d5c0 → 899e84a83 → b45c411dc → a1a8926fe → 7bc6da768). `requestSave` queues the draft whenever a loop is active even when store-equal, with equality deferred to drain time, plus a dedicated revert regression test (`NotesPanel.test.tsx`: reverting to stored text during an in-flight save persists the revert). No code change needed — the verdict appears to have reviewed a stale file state. Awaiting CI + re-review.

obra:
Round-9 finding (on 7bc6da7) verified against the reviewed head itself and rejected: `git show 7bc6da768:...NotesPanel.tsx` lines 241-253 already queue the normalized draft whenever a loop is active (equality short-circuit only when no loop runs), with the revert scenario in the comment and a dedicated regression test (`reverting to the stored text during an in-flight save still persists the revert`, added in 8bb88d5c0, present at this head). The verdict describes the pre-8bb88d5c0 code. No change needed.

obra:
Rebased onto current main (2664cc881) and force-pushed as f255e7eab: 45 branch commits replayed + ctx-passing relay-callsite adaptation + vet-fix, all conflicts resolved (union keeps per the file rules, no feature or main code dropped). Mergeability is now MERGEABLE and CI is running for the first time on this PR (web, static-build, lint, race, tests, fuzz pending/in-progress) — the rebase push retriggered what plain pushes never started. Round-9 rebuttal stands (finding already fixed in 8bb88d5c0). Awaiting CI + re-review.

obra:
CI failures on f255e7eab triaged and fixed in 871d3e9f7 (pushed; local runs skipped per operator constraint, CI is the prover):

- **web** (real, mine): reducer snapshots had `imageSessionId` before `humanNote` from the union resolution; serialized order is `humanNote` first. Reordered at 4 sites.
- **lint-golangci** (real, mostly mine): tagged switch in TUI test, `strings.Cut` scheme extraction, dropped ineffectual history capture, `slices.Backward` loop, removed two dead serialized helpers superseded by the RPC path.
- **fuzz** (real, mine): `FuzzEventDataProgram` needs seed rows for the new `NotesUpdatedData`/`UrlsUpdatedData` payloads — added.
- **static**: aggregation wrapper over the above; follows the children.

Awaiting CI re-run + re-review.

obra:
CI failures on 871d3e9 triaged and fixed in 653ee89a6 (pushed; targeted local runs only per operator constraint, CI is the prover):

**Agent (real, mine — 2 regressions from the rebase-era fixes):**
- `TestAdoptedSpendFailureRetriesFinishSpend`: `consumePendingNotesHumanDurable` bypassed the test fault hook via `autoSaveMetaLocked` directly, so the armed spend fault never fired (faultCalls=0, B silently applied). Now routes through `persistNotesMetaLocked`, which consults `notesAutoSaveFault`. Verified: fault surfaces, A intent stands, same-ID retry finishes the spend at steering==1 with B applied.
- `TestAdoptedIntentSurvivesRefusedSteer`: a refused-adoption retry (gen 2, no own markers) fell through adoption (gen>1 refusal) into a silent changed=false success — steering stayed 0. `adoptPendingNotesDelivery` no longer refuses gen>1 leases; the retry adopts A's still-pending intent and steers exactly once. Verified: all 3 Adopted tests pass, broader notes sweep passes, full `agent` package suite passes locally (456s), vet/gofmt clean.

**Web (real, mine — NotesPanel save coordinator):**
- Revert-during-flight, B-behind-failing-A, sibling-Saved, and labelless-link tests failed. Root causes: (1) the drain pre-deleted queue entries before the await, so a flush racing the drain started a second drain and double-persisted; (2) flushes re-attempted just-failed drafts; (3) labelless expectation missed the Remove-button space. Fixes: drain claims entries only at terminal outcome; queue-while-active on store-equal drafts with drain-time revert persistence; per-session failed-text flush gate; converged-saves set for Saved-on-noop. Verified: all 29 NotesPanel tests pass locally, tsc clean.
- Lint: the one Biome `useExhaustiveDependencies` suppression warning on the live-to-ended flush pre-exists on the base commit (verified via stash); unchanged by this fix.

**RoboRev round-10 verdict (on 871d3e9, Minor UI/state):**
- **Medium (save-queue revert race, NotesPanel.tsx:248-250) — fixed:** this is exactly the queue-while-active change above, implemented per the verdict's prescription (enqueue the normalized draft in dirtyRef while a loop runs; drain persists the revert). The verdict reviewed 871d3e9, which predates the fix; 653ee89a6 carries it with the revert regression test green.
- **Low (ungated Notes entry points) — deferred with rationale:** capability-gating the menu/palette/pane would add a third capability-plumbing path for a cosmetic empty-Sheet edge on old daemons; the empty body is the same null-render every other capability-absent panel shows. Happy to gate if you want it, but it is not a correctness issue.
- **Low (remove double-click guard) — deferred with rationale:** a duplicate remove surfaces the daemon's `no URL entry with id` error as a toast, which is the honest already-removed signal; an in-flight set would trade a truthful error for silent-click-drop complexity. Reconsider if duplicate-remove toasts show up in practice.
- **Low (session-switch focused/saving reset) — partially addressed:** Saved/error are already per-session (uiRef-gated, reset on sessionRef change); `saving` is intentionally global (one loop at a time) and always clears on settle. `focused` is append-only within a mount and the reseed compares stored-note transitions, not focus edges — a stale focused=true cannot block reseed after a remount because the sessionRef-change effect resets the draft unconditionally.

Awaiting CI + re-review.

obra:
Web failure on 653ee89a6 triaged: the `web` lane's Biome gate failed on a single→double quote formatting violation in `SteeringItem.test.tsx:77` (a branch-added test; Biome enforces double quotes). Fixed in 9af23f325 (pushed) — verified with `biome check` on the file and the full `src` scope shows only the pre-existing NotesPanel suppression notice that also fires on the base commit. Go lanes on 653ee89a6 are green (tests, lint-golangci, fuzz, race). Awaiting CI re-run + re-review.

obra:
## Resolution status for the current finding set

The head has advanced since this review (`e5530f3` → `c1d22e8a6`). Both findings this comment has carried are resolved on the current head, with evidence below. This comment is mutable, so also noting the earlier revision's finding explicitly.

Head `c1d22e8a6` folds in `main` at `bc1ed2aef` (after `e812703fc`) plus the branch's repair commits.

### 1. Notes navigation affordances not gated (Low, current body)

**Fixed by `7af4c9260` "fix(notes): gate read navigation on shared-notes capability".** All Notes openers now share one predicate, `canReadSharedNotes` (`cmd/evener-hub/frontend/src/protocol/sharedNotesAvailability.ts`):

- session menu item: omitted when unsupported (`SessionMenu.tsx:134`, fed from `RailRow.tsx:447` and `SessionChrome.tsx:307`)
- command palette: declarative `capability: "sharedNotes"` plus a runtime guard that returns blocked/unavailable (`commands.ts:593-601`, generic check at `:705`)
- session chrome open path: early return when unsupported (`SessionChrome.tsx:167`)
- panel: trigger omitted, imperative `open()` no-ops, body inert (`NotesPanel.tsx:161`, `:277`, `:288`)

**Ended sessions.** On web, the daemon itself clears the capability for closed sessions — `SharedNotes: s.notesHumanSetFunc != nil && s.urlsRemoveFunc != nil && !closed` (`server/appwire_runtime.go:2577`) — so ended/closed sessions produce no trigger through that same predicate. On TUI the SharedNotes capability bit is deliberately retained on ended sessions so the drawer keeps rendering the section read-only; mutating commands are gated on capability **and** liveness (`sharedNotesLiveAvailable`, `cmd/evener-tui/hub_command_registry.go:564`), so `/notes` and `/url-remove` are not advertised on read-only past sessions (where dispatch would fire resume-first writes). That is the ended-read clarification.

Evidence on this head: 158 tests green across `NotesPanel.test.tsx` (incl. the `ended`/`closed`/`notLoaded` read-only cases, the read-only-when-ended editor/remove assertions, and the live-to-ended transition case), `commands.test.ts`, `SessionChrome.test.tsx`, `SessionMenu.test.tsx`, and `rail/index.test.tsx` — zero console diagnostics.

### 2. `RemoveSessionURL` retry false success for never-existing IDs (Medium, flagged in this comment's earlier `63eb7f8` revision, absent from the current synthesis)

Flagged as "validate rather than treat the omission as resolved" — validation found it real, and it is fixed by `b766b70b8` "fix(notes): validate URL targets before reserving removals".

Root cause: the removal reserved the mutation with no prepare-time validation, and the crash-recovery takeover path inferred success from the entry's absence (`AttemptGeneration > 1` ⇒ "the pre-crash attempt passed validation and removed it"). For an id that never existed, a retry after a crash therefore reported `removed = true`.

Fix: the prepare callback validates the target against the snapshot before the reservation takes the lease (`agent/session_notes_rpc.go`), so absence on takeover is only reachable for an accepted removal; the unknown id is rejected with `InvalidParams`. Regression coverage in `agent/session_notes_mergeability_test.go`: unknown-target reservation/crash plus two reload/retry cycles, and existing-target recovery at the none/reservation/receipt boundaries.

### Verification on the current head

- `make merge-approval-gate`: all 17 stages pass (9 lint/build incl. golangci-lint, 8 module suites).
- `make test-web-browser` (TMPDIR=/tmp): all 5 real-Chrome guards pass.
- frontend: 448 files / 10259 Vitest + 81 Node tests green; typecheck clean.
- agent module race suite (`cd agent && go test -race -count=1 .`): green.

Known remaining console diagnostics: main's pre-existing 767 React `act` warnings (`composer.test.tsx` 760, `Composer.integration.test.tsx` 7), tracked by #1138; this branch introduces none.

obra:
## Resolution: `restartRequired` read-only gating in the notes panel (Medium)

**Verified real, fixed in `c76e2682c`** ("fix(notes): render the notes panel read-only for restartRequired sessions"). This is on top of the head this synthesis reviewed (`c1d22e8a6`); head is now `c76e2682c`.

**Root cause.** `NotesPanel.tsx` decided editability from its own status set, `NOTES_ENDED_STATUSES = {"ended","closed","notLoaded"}` (`:71`, used as `const live = !NOTES_ENDED_STATUSES.has(...)` at `:137`), which omitted `restartRequired`. The store's write gate `canWriteHumanNote` (`stores/humanNoteDrafts.ts:183`) refuses exactly `{ended, closed, notLoaded, restartRequired}`, and is enforced at `threads.ts:2796` (`setHumanNote` throws "Session cannot accept notes") and `threads.ts:900` (blocks dispatch of the queued mutation). A restart-required session therefore rendered an editable textarea and URL `Remove` buttons whose actions were guaranteed to fail.

**Fix.** The panel now derives editability from the store's write gate itself — `const live = canWriteHumanNote(model)` — and the duplicated status list is gone, so the display predicate and the enforcement predicate cannot drift apart again. Read-only sessions keep the same rendering contract as the other three statuses: saved human/agent note values and links stay visible, with no editor and no removal controls.

**Regression coverage** (added red-first, then fixed):
- `imperative Notes opening rechecks capability for restartRequired sessions`
- `Notes sheet opens saved restartRequired content without editor or removal controls`
- `notes panel is read-only when restartRequired: values shown, no editor, no remove buttons`

All three failed before the fix exactly as the finding describes (editor and removal controls present) and pass after.

**Verification on `c76e2682c`:** NotesPanel 49/49; 634 tests across 26 affected suites (chrome, rail, session menu, palette, AppShell) with zero console diagnostics; full frontend 448 files / 10262 Vitest + 81 Node green; typecheck and Biome clean; `make merge-approval-gate` 17/17 stages; `make test-web-browser` 5/5 Chrome guards. I also swept every other frontend `restartRequired` gate (`SessionChrome.tsx:148`, `Session.tsx:394-400`, `sendQueueAvailability.ts:85`, `liveness.ts`, rail) — all consistent; the panel was the outlier.

The two findings this comment previously carried (notes opener gating, and the earlier `RemoveSessionURL` retry false-success) remain resolved as described in the previous comment; this synthesis confirms them.

obra:
## Resolution: notes mutations missing recovery-admission fencing (Medium)

**Verified real, fixed in `c7e0279be`** ("fix(hub): fence notes mutations against session recovery admission"). Head is now `c7e0279be`.

**Root cause.** `admitSessionRecovery` (`cmd/evener-hub/app_sources.go`) captures `{sessionID, epoch}` into the request context only for the methods its switch recognizes; `notes/human/set` and `urls/remove` fell through to `default: return ctx`, so they carried no admission. `sessionRequestRecoveryEpoch` then returns the *current* recovery epoch sampled at execution time (`app_sources.go:284-289`), and the guard in `withDeletionTargetOwnership` compares `state.Epoch != epoch` — which a stale request can never fail, because both sides were sampled at the same late moment.

**Verified with a test that shows the harm, not just a mismatched error.** I extended `TestCapturedSessionActionsRejectAdmissionBeforeRecovery` with both methods and advertised the shared-notes capability on the fake source, so the downstream capability gate could not mask the outcome. Before the fix both cases reported `err=<nil> applied=1`: a `notes/human/set` and a `urls/remove` received before the force-stop actually applied to the resumed session. After the fix both reject as `sessionRecoveryAdmissionError` with `applied=0`, and the existing fresh-admission path is unchanged.

**Fix.** Both methods are now in the switch's `ref`-parsing group (both param structs carry `ref`), so a request admitted before recovery keeps its pre-recovery epoch and is refused with the standard "session recovery canceled this pending action; submit it again" contract.

**Verification on `c7e0279be`:** full `cmd/evener-hub` package suite green; the recovery/admission/force-stop tests pass under `-race`; `make merge-approval-gate` 17/17 stages; `make test-web-browser` 5/5 Chrome guards.

obra:
## Resolution: `SharedNotes` missing from the local session roster (Medium)

**Verified real, fixed in `43801dfc5`** ("fix(hub): advertise shared notes in the local session roster"). Head is now `43801dfc5`.

**Root cause.** `threadFromEntry` (`cmd/evener-hub/internal/appsource/local_daemon.go`) builds the roster's optimistic capability set for `ListThreads`, and that literal omitted `SharedNotes`. Live local sessions therefore came back as `sharedNotes: false` from the list even though the daemon's `thread/read` supports notes, so list-derived models reported the capability unsupported and the UI hid Notes until hydration.

**Verified red-first** with `TestLocalDaemonSourceListAdvertisesSharedNotes`, covering a live entry, a read-only alias, and a restart-required entry: the live entry returned the full capability set except `SharedNotes: false`; the alias and restart-required entries were already false.

**Fix.** `SharedNotes: !item.ReadOnlyAlias` in the roster literal, consistent with how the roster already advertises the other live capabilities optimistically (`Clear` uses the same alias condition), while read-only aliases stay false and restart-required sessions keep their existing wholesale clearing.

**Verification on `43801dfc5`:** full `cmd/evener-hub` module suite green; `make merge-approval-gate` 17/17 stages; `make test-web-browser` 5/5 Chrome guards. No frontend change was needed — the panel's editability already derives from the store's write gate, so a roster-advertised capability now simply renders read-only where writes are refused.

Thanks for the note that the persistence, RPC, projection and UI/TUI work was found correct with covered tests; this was the last capability-advertisement gap I know of in that path.

obra:
## Round 5: four findings fixed, one accepted as a documented limitation

Reviewed head: `43801dfc5`. Fixes are in `cb841c783`; verification below.

### Medium — `urls/remove` bypasses the durable outbox

**Accepted as a documented limitation, not fixed here.** Your reading is correct: `urls/remove` goes to the wire directly with a fresh mutation ID (`threads.ts:2822`, `hub_commands.go:1241`), so a lost response plus a lost `evener/urls/updated` push cannot be retried under the original identity.

Two things make the right fix larger than this PR's scope. First, the URL list is metadata: the `evener/urls/updated` push, not the response, is the authority for what the UI shows, and a stale entry converges on the next push or snapshot. Second, the durable path is only worth having for a method whose response carries the state the outbox dispatcher needs; `urls/remove` returns no receipt, so routing it through the outbox means designing receipt/ack wiring for a state-less response. That is a real piece of work and I would rather land it deliberately than bolt a second, weaker durability mechanism onto this one.

Recorded as a follow-up: return a mutation receipt from `urls/remove`, route both clients through the durable outbox, and preserve the original ID and payload for retries.

### Medium — `ReadCanonicalHumanNote` hard-fails and takes down the session list

**Fixed.** `pastEntryThreadForList` now treats a failed canonical read as "no canonical note" (`entry.Meta.HumanNote` stays empty) and falls through to the metadata projection, exactly as you suggested; `ctx.Err()` is still checked first and still propagates. `pastEntryThread` keeps the strict behavior, and I verified `ReadCanonicalHumanNote` has no other non-test caller.

Red-first: `TestPastThreadForListToleratesCorruptMutationJournal` failed with the decode error you described and now passes. Sibling degradations (`pastEntryCost` nil, ownership-error swallow, `mergePastMetadataForList` non-ctx tolerance) confirmed unchanged.

### Medium — blur-save debounce has no flush on teardown

**Fixed, and hardened per review of the fix.** The delayed save body is now reachable from teardown: a once-installed `pagehide` listener runs pending flushes, mirroring the DockHost rule you cited. `cancel()` and re-focus still clear the pending save.

Verifying the fix surfaced one narrow gap, now closed: the flush ran the save body but left the live 10s handle in place, so a page that survived `pagehide` (bfcache) could enqueue the same note a second time while the first acknowledgement was still outstanding. `save()` now retires the timer on entry, mirroring `cancel()`. Red-first: the new test held the ack open, fired the flush, and elapsed the full deadline; it failed with `expected "setHumanNote" to be called 1 times, but got 2 times` before the change and passes after.

### Low — unreachable `notesGuard` fields and `*Serialized` helpers

**Fixed.** Removed the nine unread fields and their `newToolDeps` wiring, the five exported forwarders, the three `*Session` `*Serialized` methods, and the single-use `notesSnapshotPair` wrapper. Grep-verified zero remaining references. Tests updated in place; no test was deleted or weakened.

### Low — torn `notesEverProjected` view across two `s.mu` acquisitions

**Fixed.** `notesContextBlock` now reads the flag from `notesProjectionSnapshot()`, which takes `notesEverProjected` in the same `s.mu` section as the agent note and URLs, so the emptiness check and the `everProjected` check observe one consistent state. `notesSnapshotAll` delegates to it with an unchanged signature and lock scope. The two existing contract tests (`TestNotesRemoveAllProjectsEmptySnapshot`, `TestNotesNeverPopulatedProjectsNothing`) cover both decisions and stay green.

Honest limitation: the interleaving is not dynamically reproduced (it needs a concurrent `seedNotesProjectionLocked`/`resetNotesProjectionAfterCompaction`), so this is correctness by construction rather than a timing-dependent regression test.

---

**Verification on `cb841c783`:** `make merge-approval-gate` all stages green; `make test-web-browser` 5/5 Chrome guards; full frontend suite green; focused `agent` and `cmd/evener-hub` suites green; `make vet` green.

Thanks for the detail on each finding, including the pointer to the DockHost precedent.

obra:
## Round 6: read access restored, two comments corrected, one declined

Reviewed head: `cb841c7`. Changes are in `cdeb80862`. This round also carries the integration that unblocks CI (below).

### Medium — restart-required projections suppress read access to persisted shared notes

**Fixed.** Your split is right: reading a saved note does not require a live session, and the capability that gates reading is not the one that gates editing.

- `cmd/evener-hub/app_threadread.go`: the restart-required and daemon-discovery-error branches no longer zero the whole capability set; they advertise `readablePastCapabilities()` — `SharedNotes: true` and nothing else. The past entry's note is read from disk in the same function, so the content is there to serve.
- `cmd/evener-hub/internal/appsource/local_daemon.go`: a restart-required roster entry advertises only `SharedNotes`, which is what makes the section appear before hydration.
- Every mutation capability stays cleared in both paths, and `liveEntries()` still excludes restart-required owners from RPC routing, so writes remain refused server-side; the web side already derives editability from `canWriteHumanNote`, which refuses `restartRequired` (`humanNoteDrafts.ts:214-218`). The frontend read-only rendering for that status was already covered (`NotesPanel.test.tsx:497`).

Red-first: `TestHubProtocolUpgradePreservesTranscriptAndRejectsUndeliverableMessages` now asserts the saved-transcript read keeps `SharedNotes` alongside the existing "no unsupported mutations" assertion, and `TestLocalDaemonSourceListAdvertisesSharedNotes` now expects read-only notes for the restart-required entry (and still expects mutations cleared). Both failed before the change and pass after.

### Medium — `notesProjectionSnapshot` lock-order claim inconsistent with actual locking

**Comment corrected; the locking change was declined deliberately.** You are right that the three-field claim was wrong: the agent note, URLs, and `everProjected` are read under `s.mu`, while the human note is read under `clientMutations.stateMu` after `s.mu` is released. The comment now says exactly that and states what actually bounds the window: `notesContextBlock`'s callers hold `notesUpdateMu` across the render-compare-record unit, which is what serializes notes-store mutations against the emptiness and cleared-marker decisions.

I did not hold both locks together because `stateMu` exists precisely so a snapshot read answers while another mutation is mid-write (`session_client_mutation.go:860-866`), and the human note is the only field it owns; nesting the two would buy an atomic cut of a field whose writer is already outside that unit. If you think the atomic cut is required rather than documented, I would rather take that as a separate change than smuggle it into this PR.

### Low — corrupt-journal tolerance swallows context cancellation

**Declined, with evidence.** `ReadCanonicalHumanNote(stateDir, sessionID string)` takes no context (`agent/session_notes.go:34`), so it cannot return a cancellation — the failure modes reaching that handler are validation, filesystem, and decode errors, exactly the ones the degrade rule is for. `pastEntryThreadForList` already returns `ctx.Err()` before any per-entry work, and the list path is cancellation-aware around it. Adding a post-read `ctx.Err()` check would change cancellation semantics (a canceled roster sweep would start returning an error instead of the degraded row) without a defect behind it, so I left the behavior alone.

### Low — `pagehide` teardown flush can drop before durable enqueue

**Comment corrected to state the real boundary.** The flush starts the save without waiting for the debounce, but `save()` awaits the session's thread handle before it can enqueue, and the enqueue is itself an async store commit. So a hard teardown is best-effort, and the comment now says so instead of claiming the durable enqueue is what cannot be dropped. The residual window is inherent to an async store; a page that survives (bfcache) gets the save.

---

**CI:** the `native` lane was failing on the PR merge ref — the mobile/ modules arrive from main, and this branch makes `sharedNotes` a required `ThreadCapabilities` field, so main's capability literals no longer compile. The branch now carries main (clean merge) with the field completed in the mobile literals and in `mobile/src/services/conversation.ts`'s required-field list; the shared mobile fixtures the stricter extractor rejected are complete too (777/777).

**Verification on `cdeb80862`:** `make merge-approval-gate` green (lint, build, ROOT_FULL test, `test-native`, `test-api-package`); `make test-web-browser` 5/5 Chrome guards.

obra:
## Round 7: two read/write-split regressions fixed, two findings declined with evidence, one deferred

Reviewed head: `cdeb808`. Fixes are in `4169bb818a`.

The previous push carried the main integration that unblocked CI (`native` was failing because this branch makes `sharedNotes` required and main's mobile modules predate it) and the restart-required read-access change. Two of this round's findings are consequences of that read capability being advertised where writes are refused, and they are real.

### Medium — `restartRequired` sessions expose shared-notes writes as live in the TUI

**Fixed.** The write half was missing. The TUI had only capability + liveness to work with, and `Live` is false only for `closed`/`notLoaded`.

Command availability now also requires a state that can change notes: `hubCommandContext` carries the session status and `sharedNotesLiveAvailable` refuses `restartRequired` (`hub_command_registry.go`), with the same predicate used by the `/notes` dispatch guard (`hub_commands.go`) and the empty-state affordance (`details_drawer.go`) — so a restart-required empty section reads "No shared notes" instead of "Add a note with /notes". The read capability and the drawer rendering are untouched.

Red-first: `TestSharedNotesCommandUnavailableWhenRestartRequired` failed with "restart-required session advertised /notes" and the drawer assertion failed with the trigger rendered; both pass after.

### Medium — web `removeURL` missing the `canWriteHumanNote` gate

**Fixed.** `setHumanNote` refused read-only sessions, `removeURL` did not. `removeURL` now takes the same gate before it reaches the wire, so a programmatic removal is refused for `ended`/`closed`/`notLoaded`/`restartRequired` exactly as the panel claims.

Red-first: a new store test failed before the gate was added and passes after, asserting both the refusal and that no `urls/remove` request is sent.

### Medium — shared notes and URLs share one carrier generation

**Acknowledged, deferred to a follow-up.** The analysis reads correctly — the single carrier means a notes commit can drop a newer URL sample, and `SessionURLs` has no compensating carrier. This is a real backend change (independent generations, or an `assign` that drops only the fields the winning carrier wrote) and it is not something to bolt onto this PR at this stage. Recorded as a follow-up.

### Low — no-op note clear mutates durable journal state

**Declined, with evidence: that write is load-bearing.** Materializing explicit-empty authority is exactly how the clean cutover shadows a legacy metadata note. `TestHumanNoteCleanCutoverDoesNotImportMetadata` restores a session whose metadata still says `unsupported-old-note`, clears it, and requires the canonical authority to read present-and-empty (`agent/session_notes_atomic_test.go:381-387`) — the failure it prevents is that stale metadata is re-imported on the next read. I tried the conditional write: it makes the absent→empty clear skip the write, the shadow never lands, and that test fails. The write is deliberate, not an accounting slip; I reverted my change and the contract test passes.

### Low — declared wire methods never registered

**Documentation corrected; the constants stay.** `notes/agent/set` and `urls/add` are deliberately *agent-tool* verbs, not hub RPCs — the protocol tests pin that they must be absent from the RPC method catalog (`protocol_notes_test.go`), and `ValidateMutationParams` treats them as envelope-less. The real defect was the prose: `ThreadCapabilities.SharedNotes` named four wire verbs, and both tool definitions called themselves wire RPC methods. Those three comments now say what is actually wired (two hub RPCs: `notes/human/set`, `urls/remove`).

### Medium — `urls/remove` bypasses the durable mutation outbox

**Unchanged: still an accepted, documented limitation.** This is the same finding as round 5, and the reasoning has not changed — the URL push is the authority for the list, and the response carries no state for an outbox dispatcher to settle, so durable routing means designing receipt/ack wiring for a state-less response. That remains a deliberate follow-up rather than a second, weaker durability mechanism on this PR.

---

**Verification on `4169bb818a`:** `make merge-approval-gate` green (lint, build, ROOT_FULL test, `test-native`, `test-api-package`); `make test-web-browser` 5/5 Chrome guards; focused `cmd/evener-tui`, store, and notes suites green.

obra:
## Round 8: command-context state threaded everywhere

Reviewed head: `4169bb8`. Fix is in `e810ae1efc`.

### Low — TUI command-availability surfaces omit `m.detail.State`

**Fixed, and this was an incomplete fix of mine from round 7.** I threaded the state through the slash path but left the other constructions blank, so a restart-required session still saw `/notes` and `/url-remove` listed as available in the palette and the slash help. Every mutation was still refused downstream, as you noted, but the affordances were dead.

Every construction now carries the session state: `commandPaletteEntriesForSession` (and its `commandPaletteEntriesForRows` wrapper), the palette open in `hub_keys.go`, the palette dispatch context, and `hubCommandHelpLive`/`hubSlashCommandHelpLive`. `runHubURLRemove` also takes the same `sharedNotesWritable` guard as `runHubNotes`, so the dispatch side is consistent with the listing side.

Tests: `TestSharedNotesSurfacesThreadSessionState` (palette entries and help with an explicit state) and `TestSharedNotesPaletteThreadsSessionState` (drives the production `openCommandPalette` so the wiring itself is covered). I mutation-checked the wiring — reverting just the `m.detail.State` argument makes the palette test fail with "restart-required session palette offered /notes" — then restored it.

### Medium — `removeURL` bypasses the durable mutation outbox

**Unchanged: the accepted, documented limitation** (same finding as rounds 5 and 7). The URL push is the authority for the list and the response carries no state for an outbox dispatcher to settle; durable routing means designing receipt/ack wiring for a state-less response, which stays a deliberate follow-up.

### Low — `handleRemoveURL` has no in-flight guard

**Acknowledged, deferred to a follow-up.** You are right that a double-click sends two removals and the second surfaces a spurious error. It is real, but it is pre-existing and needs pending state in the panel plus removal of the now-absent entry from the model, which is more than a plumbing fix; recorded as a follow-up rather than widened into this PR.

---

Thanks for the clean-areas confirmation on the store logic, lock ordering, and the read-only restart-required paths.

**Verification on `e810ae1efc`:** `make merge-approval-gate` green (lint, build, ROOT_FULL test, `test-native`, `test-api-package`); `make test-web-browser` 5/5 Chrome guards; full `cmd/evener-tui` package green.

obra:
## Round 9: stale history-copy comment corrected

Reviewed head: `e810ae1`. Fix is in `ee5e5015d5`.

### Low — `agent/session_model_call.go` stale "no redundant copy" comment

**Fixed.** You are right that the comment contradicted the code beside it, and that a maintainer trusting it could delete the copy and reintroduce the G4 stale-boundary / notes-turn loss. The note now says what the copy is for: `historyTurns` holds the refreshed copy re-snapshotted under `s.mu` above (capturing the `maybeAppendNotesContext` append and re-pairing `turnHistoryBaseline`), and `replayScope`/the request expand that copy rather than taking another one — with an explicit warning not to drop it as redundant work.

This push also carries a branch-freshness merge: `main` requires an up-to-date branch (`required_status_checks.strict: true`) and had advanced by one commit (the Gemini/vertex fix), so it was merged cleanly.

---

Thanks to the two reviewers who found no issues on the state-threading fix.

**Verification on `ee5e5015d5`:** `make merge-approval-gate` green (lint, build, ROOT_FULL test, `test-native`, `test-api-package`); `make test-web-browser` 5/5 Chrome guards; full `agent` package green.

obra:
## Round 10: one regression reverted, three findings fixed, one accepted, one deferred

Reviewed head: `ed860b2`. Changes are in `e9fdc1db21`.

### Medium — teardown schedules a save after releasing focus

**You are right, and I reverted it.** The teardown-save was added in `ed860b2a42` in response to a separate report that a dirty draft was being lost when the Notes sheet closed. Checking the repository's own contract before keeping it showed that report misread the design:

- `docs/web-ui/shared-notes.md:7` — "Closing a panel without leaving the field does not save. The draft remains available when you reopen it in the same browser session."
- `docs/superpowers/specs/2026-09-10-shared-notes-simplify-design.md:41` — "Closing a focused panel without a blur retains its draft and does not invent a save. A timer from an actual earlier blur survives panel unmount."

The existing `NotesPanel` test "closing without blur keeps the shared draft without inventing a save" pins the same thing. The draft is retained in browser memory by design, not lost. The teardown path, its helper, and my added test are gone; the panel cleanup is back to releasing the focus owner, and the blur-only scheduling is restored. The URL-decode fix from that commit stays.

### Medium — stale `expectedInstanceId` captured at debounce schedule time

**Acknowledged, deferred.** The capture-at-schedule is deliberate fencing from the original debounce design, and re-sampling at save time changes what the fence protects — that is a design decision about writing across an instance rotation, not a mechanical fix. Recorded as a follow-up rather than changed here.

### Medium — thread retain leak on early-return and reset paths

**Fixed.** `save` released the retain only in its `finally`, which the empty-draft early return skipped, and `resetHumanNoteDrafts` cleared timers without calling the stored `release`. Both now release: the early return calls `release()` before returning, and the reset path clears the timer and calls `draft.release?.()`. Honest limitation: no new test covers these two paths — there is no seam to observe `releaseThread` from the store test, and the empty-return path needs the draft removed while its timer is armed.

### Medium — URL removal bypasses the durable mutation outbox

**Unchanged: the accepted, documented limitation** (this is the fourth time it has been reported; it remains the recorded follow-up).

### Medium — new mobile client rejects any peer omitting `sharedNotes`

**Fixed — this was my regression.** I added `sharedNotes` to `REQUIRED_CAPABILITY_FIELDS` in round 6, and you are right that the protocol version did not change, so a hub built before the field is indistinguishable and simply omits it. The extractor then threw inside `open()` before the ref/capabilities pair committed, failing every capability-gated action for the session. `sharedNotes` is no longer required: absent means "not supported" (`false`), a present value is still validated as a boolean, and the 12 legacy fields remain required.

Red-first by mutation: restoring the required entry fails the new test with exactly `ConversationService: capability "sharedNotes" is not a boolean`. Mobile suite 778/778 and `mobile-native` typecheck green.

### Low — dead `saveMeta` field on `notesGuard`

**Fixed.** Removed the field, its comment block, and the `s.persistNotesMeta` assignment. The `mutate*` helpers persist through `persistNotesMetaLocked`, so nothing read it.

---

**Verification on `e9fdc1db21`:** `make merge-approval-gate` green (lint, build, ROOT_FULL test, `test-native`, `test-api-package`); `make test-web-browser` 5/5 Chrome guards; mobile 778/778; `NotesPanel` 52/52; `tsc` and Biome clean.

obra:
## Round 11: note-context escaping fixed, one accepted limitation unchanged

Reviewed head: `e9fdc1d`. Fix is in `c64d5e0ca2`.

### Medium — untrusted note/URL content injected into model context without escaping

**Fixed, and this one matters.** You are right on both counts: the block is framed with literal `<shared-notes>` delimiters, and `normalizeNote` only collapses whitespace and clamps length, so a note or a `urls_add` label carrying `</shared-notes>` closed the block and the model read whatever followed as harness-authored context — persistent indirect prompt injection. `formatNotesLinkLine` emitted the URL, label, and id raw as well.

The rendering now escapes every dynamic field with `html.EscapeString`, matching the goal-context precedent you cited: the human note, the agent note, and each link line's URL, label, and id. Raw values are untouched for persistence and the wire; only the rendered block (and the `notes_read` output, which shares `formatNotesLinkLine`) escapes.

Test: `TestNotesContextBlockEscapesInjectedContent` sets a human note and a URL label to `</shared-notes> <instructions>` and asserts the block still contains exactly one closing delimiter, still ends with it, and never carries the raw breakout text. Mutation-checked: dropping the escaping fails it with "injected content terminated the block".

### Medium — URL removals bypass the durable mutation outbox

**Unchanged: the accepted, documented limitation** (fifth report). It remains the recorded follow-up.

---

Thanks for the note that close-without-blur retention, absent-`sharedNotes` tolerance, and the retain-release fixes read clean; that matches the round-10 rework.

**Verification on `c64d5e0ca2`:** `make merge-approval-gate` green (lint, build, ROOT_FULL test, `test-native`, `test-api-package`); `make test-web-browser` 5/5 Chrome guards; full `agent` package green.

obra:
## Round 12: escaping narrowed to the model copy

Reviewed head: `c64d5e0`. Fix is in `c165e4fa37`.

### Medium — HTML escaping leaks into tool output, corrupting URLs

**Fixed — this was my over-reach in round 11.** You are right: `formatNotesLinkLine` is shared with the plain-text `notes_read`/`urls_add` output, so `https://host/a?x=1&y=2` came back as `...&amp;y=2`, and a model reusing a read-back URL would use a corrupted value.

The render is now split. `notesContextBlock` (and the shared `formatNotesLinkLine`) stay raw for persistence, display, and tool output; `notesContextBlockForModel` escapes the dynamic fields and is the only copy the model request sees. The framing tags stay literal in both, so the delimiter guarantee holds. Regression case added with a `&`-bearing URL, asserting the tool-output line keeps it raw and the raw block keeps both the URL and the injected text.

### Low — persisted/displayed `NOTES_CONTEXT` turn carries escaped text

**Fixed by the same split.** `maybeAppendNotesContext` now appends through `appendTurnWithTranscriptMessage` with the escaped block as the live model message and the raw block as the durable/displayed one, so the transcript line and the collapsed group summary render real text and real URLs. The change gate still compares the raw block (`notesLastProjected`), so resume/compaction consistency is unchanged. The model-side assertion is mutation-checked: making the model copy non-escaping fails with "injected content terminated the model block".

### Medium — terminal injection via unstripped control sequences

**Acknowledged, deferred to the follow-up PR.** Real and pre-existing: `normalizeNote` keeps ANSI/C0/C1 controls, so an agent-supplied note, label, URL, or id reaches the TUI raw. The fix has a design choice (strip during normalization, which changes the canonical stored value, versus sanitize at the TUI render boundary), so it is recorded with the other follow-ups rather than folded into a merge-prep push.

### Low — URI scheme rejection blocks filenames with colons

**Acknowledged, deferred to the follow-up PR.**

### Low — `blurHumanNote` captures instance ID before hydration

**Acknowledged, deferred.** This overlaps the stale-`expectedInstanceId` item already recorded: both are about *when* the instance identity is sampled relative to hydration and the debounce, and they should be settled together.

---

**Verification on `c165e4fa37`:** `make merge-approval-gate` green (lint, build, ROOT_FULL test, `test-native`, `test-api-package`); `make test-web-browser` 5/5; full `agent` package green (239s).

This push also carries the pre-merge branch update: current `main` merged cleanly (8 commits) for the `strict` up-to-date requirement, plus the capability literals in the mobile fixtures that update brought with it (they predate the required `sharedNotes` field).

obra:
## roborev round 13 (head c165e4f)

The High finding is real, and it was introduced by my round-12 fix. Confirmed in source: `maybeAppendNotesContext` put the escaped copy in live history and the raw block in the transcript, `ResumeHistory` rebuilt model history from the transcript, and `expandHistory`'s `case schema.TurnNotesContext` passed the raw message to the provider. The escape expired at the first restart, exactly as described.

While fixing it I also had an independent reviewer try to falsify the fix, and it found a second instance of the same defect that the finding did not name: a `fork_context: true` delegate inherits the parent's transcript prefix (`NewSession` → `s.history = ResumeHistory(inheritedContext)`), so the child's model context received the same raw block. That path is fixed here too.

**Fixed in `1ef0d8d8cc`.** The invariant is: *model history never carries the raw notes text.* Both entry points that turn a persisted block into model context rewrite it to the escaped copy, so the model keeps the context it had instead of losing it:

- `agent/session_notes_rpc.go` — `escapeNotesContextBlock` is now the single definition of the escaping contract. `renderNotesContextBlock` builds the raw block; `notesContextBlockForModel` escapes it. It cuts the fixed framing prefix/suffix (new `notesBlockOpen` / `notesBlockClose` constants, which `notesClearedBlock` is now built from) and escapes only what is between them, so an injected closing tag cannot terminate the block. This replaces the old per-field escaping inside the renderer; the byte output for every input the renderer can produce is unchanged (every literal it emits — `Human: `, `Agent: `, `Link: `, ` [id: ]`, ` (`, `)`, the `\n` separators — contains none of `< > & ' "`).
- `escapeNotesHistoryTurns` replaces every `NOTES_CONTEXT` turn in a history with its escaped copy, without modifying the input.
- `agent/session_init.go` — restore applies it to the assembled `resumeHistory` (covering both the transcript rebuild and a caller-supplied `restoreCfg.resumeHistory`), and the fork-context path applies it to `ResumeHistory(inheritedContext)`.
- The projection record is still seeded from the *raw* form (`seedNotesProjectionLocked` became the pure `lastNotesProjection`, read before the rewrite), so the change gate still compares raw render to raw record: no duplicate re-projection, the cleared marker still carries across a restart, and `agent/session.go` needs no change at all.

The transcript renders still show the raw text, as designed: the resumed session's own transcript is untouched, and a forked child's transcript is still seeded from the raw `inheritedContext` entries. `notes_read`/`urls_add` keep raw field text — that is the tool-output channel, which carries no harness framing and is what round 12 fixed by not escaping it.

**Tests** (`agent/session_notes_resume_test.go`, all green):

- `TestRestoredNotesContextNeverReachesModelUnescaped` — red first, with the raw payload in the restored model context and unbalanced framing (`</shared-notes>` twice, `<shared-notes>` once). Now asserts the restored context carries the escaped block, no raw payload, paired framing, and that an unchanged store does not re-project.
- `TestForkedDelegateNeverInheritsRawNotesText` — drives a real `fork_context: true` delegate through `createDelegate` and inspects the child's first request: escaped block present, raw payload absent, parent prefix otherwise intact.
- `TestRestoredClearedNotesStillReportTheClearedMarker` — behavior preservation: the pre-clear block and exactly one cleared marker, in that order, with no re-projection.
- Mutation-checked both entry points: disabling the restore rewrite fails the resume test; disabling the fork rewrite fails the fork test. Each printed the raw injection payload as the failure.

Gates on `1ef0d8d8cc`: `make merge-approval-gate` exit 0 (lint, build, ROOT_FULL tests, test-native, test-api-package), `make test-web-browser` 5/5, `make vet` exit 0. Review: an independent falsification pass on the final design reported the invariant holds, with no remaining path from a persisted notes block to a provider. Pushed `c165e4fa37..1ef0d8d8cc` (no force).

obra:
## roborev round 14 (head 1ef0d8d)

No critical or high findings, and the round-13 resume/fork escaping fix holds. All four findings answered.

### Medium — HTML escaping corrupts model-facing note/URL content · FIXED in `c6aaa72c60`

Confirmed, and it predates round 13: `c165e4fa37` escaped every dynamic field including `formatNotesLinkLine`, so a stored `https://x.test/y?a=1&b=2` already reached the model as `...?a=1&amp;b=2` while `notes_read`, the UI and persistence kept the raw value.

`escapeNotesContextBlock` no longer applies `html.EscapeString`. It now neutralizes exactly the sequences that can forge the framing:

- literal `<` and `>`, so no tag variant is reachable, and
- the entity spellings `&lt;` / `&gt;`, escaped in turn to `&amp;lt;` / `&amp;gt;`, so the entity form cannot be decoded back into a tag.

Every other byte passes through untouched, so the model reads the URL that was stored. `&`, `'` and `"` are no longer rewritten, which was the whole source of the inconsistency.

One deliberate difference from the suggested fix: escaping only the block terminator would leave near-miss variants (`</shared-notes >`, case changes) and the entity spelling unneutralized. Escaping every angle bracket plus the entity forms preserves the anti-forgery property the old escape provided while delivering content byte-for-byte.

Test: `TestNotesContextBlockPreservesContentWhileNeutralizingFraming` — red first with `model copy rewrote "https://x.test/y?a=1&b=2"`. It pins the URL and a label carrying `&` byte-for-byte, exactly one opening and one closing framing tag, the absence of the raw breakout, the double-escaped entity spelling, and that the raw copy (display and tool output) is untouched. The round-12 `TestNotesContextBlockEscapesOnlyTheModelCopy` and the three round-13 resume/fork tests still pass.

### Low — `restartRequired` branch breaks the `SharedNotes` invariant · FIXED in `c6aaa72c60`

Verified at source: the invariant never broke, but only because of branch order. `threadFromEntry` sets `SharedNotes: !item.ReadOnlyAlias`, the restart-required branch replaces the capability set with `{SharedNotes: true}`, and the read-only alias branch after it replaces it with the zero value — so a read-only alias ends with no capabilities whatever its status.

Two changes: a test case for the combination the existing test did not cover (a read-only alias that is also restart-required, asserting the zero set), and the guard you suggested (`SharedNotes: !item.ReadOnlyAlias`) so the advertisement no longer depends on that branch running after it. The observable behavior is unchanged.

### Low — `fileURLToPath` doesn't fail closed · FIXED in `c6aaa72c60`

Confirmed. One correction to the report: this was a deliberate, tested choice, not an oversight — the existing test pinned the undecoded path with a comment calling it "the closest true answer". Your ordering argument is the stronger one though: canonical file URLs come from `canonicalFilePath`, which serializes through `net/url` so a `%` in a filename is encoded as `%25`, and a malformed escape therefore cannot come from canonicalization. The fallback only ever saw input this system did not produce, so it now returns `""`. The test's expectation now matches its own name ("a malformed file URL never becomes an open-beside target").

### Low — bare paths with colons rejected as URI schemes · CONTRACT FIXED, SEMANTICS DEFERRED

Confirmed as a contract mismatch: `DefUrlsAdd` promised bare paths while `report:2024.md` is rejected as scheme `report`.

I kept the rejection. Loosening it to an exception list (`javascript:`, `data:`, `mailto:`) would weaken a gate that exists to stop scheme-shaped input from being stored as a `file://` entry, and there is no exhaustive list of schemes to fall back on — an input like `vbscript:` or `blob:` would start resolving as a path. What was actually wrong was the discoverability: `DefUrlsAdd` now states the colon rule, the error names the `./` form that resolves the path, and a test pins both (`TestCanonicalSessionURLSchemeShapedColonPathNamesTheEscapeHatch`, red first on the old message). The semantic change stays in the follow-up backlog.

### Unchanged

The accepted `urls/remove` durable-outbox limitation is still declined per the standing call, and the carrier-generation and handleRemoveURL in-flight-guard items stay deferred to the follow-up lane.

Gates on `c6aaa72c60`: `make merge-approval-gate` exit 0 (lint, build, full tests including the whole agent suite, native, api-package), `make test-web-browser` 5/5, `make vet` exit 0. Pushed `1ef0d8d8cc..c6aaa72c60` (no force). The round-13 resume and fork-inheritance escaping tests, and the round-12 model-copy test, all still pass on this head.

obra:
## roborev round 15 (head c6aaa72)

No critical or high findings. Three Mediums: two in scope, both fixed; one is the item the maintainer already deferred to the follow-up lane.

### Medium — framing-escape bypass via decodable entities · FIXED in 756aa0bb32

Correct, and my round-14 narrowing caused it. The neutralizer covered the literal brackets and the `&lt;` / `&gt;` spellings only, so `&#60;`, `&#060;`, `&#x3c;`, `&#X3C;`, `&amp;#60;` and `&LT;` all passed through, and a reader that decodes character references would see the tag again.

It now escapes the `&` of every reference that can decode to an angle bracket — numeric (decimal, hex, zero-padded), the named `lt`/`gt` references in any case (HTML named references are case-insensitive), and the same references behind one `amp;` layer — and runs the literal pass after that, so its own output is not escaped twice. Anything that cannot forge framing still passes through byte-for-byte, so the round-14 fix holds: a stored URL keeps its `&`.

Test: `TestNotesContextBlockNeutralizesDecodableAngleBracketReferences`, red first on `&#60;/shared-notes&#62;`, covering all six spellings and asserting exactly one framing tag of each kind survives.

One honest residual, because I would rather name the trade than bury it: a reference behind *two* `amp;` layers (`&amp;amp;#60;`) now needs three decodes instead of the four that `html.EscapeString` gave that input. Closing it fully would mean escaping every `&`, which is exactly the URL corruption you flagged in round 14. I chose content fidelity plus one-decode-pass coverage.

### Medium — resume gate not propagated to web/TUI write controls · FIXED in 629caafd8d

Confirmed on both surfaces, with the fence verified to survive the projection each gate reads.

- **Web.** A fenced session presents live+idle with `SharedNotes` deliberately retained, so `canWriteHumanNote` returned true and offered edits the hub refuses. `ThreadModel` now carries `resumeRequired` (`hydrateThread` projects `thread.evener.resumeRequired`, defaulting false for older daemons) and the gate requires it false. The editor and the per-link Remove control both hang off that one gate, so a fenced session renders its saved notes read-only; `canReadSharedNotes` is untouched, and a draft or an already-queued outbox record stays put for retry after resume.
- **TUI.** `hubSessionCapabilities` carries the fence beside the capability set, `sharedNotesWritable` refuses while it is set, and the command guard, palette, slash help, `/notes` and `/url-remove` dispatch, and the drawer's empty-state hint all read that one helper, so they cannot drift apart.
- **Mobile: not affected.** `mobile/src/services/conversation.ts` hardcodes `sharedNotes: false` for the protocol version that client speaks, so it has no notes write surface to gate.

Tests: the gate with and without the fence over a live+idle session, the projection carrying the flag, and the TUI command availability, palette, help and dispatch refusals with an unfenced negative control.

### Medium — stale instance ID in the blur-debounced save · DEFERRED

That is the backlog item already deferred to the follow-up lane (stale `expectedInstanceId` capture), so it stays out of this PR by decision, not by oversight. It is recorded in the follow-up plan.

### Also in this head

The merge-time main update (`99fa1882f0`, eleven commits) with one real conflict in `stores/mutationDispatcher.ts`. It resolved to main's relocation of `AppwireClientLike` to `protocol/clientLike` plus this branch's notes types: main changed nothing else in that file, so no logic was traded and this branch's mutation paths are intact.

Two supporting commits, both worth naming because a reviewer should not have to diff them to trust them:

- `1dd0edd906` updates the four fixture snapshots in `reducer.test.ts`. Carrying the fence on `ThreadModel` is an intentional shape change (it mirrors the existing `askPending ?? false` projection), so those models gain exactly one line, `"resumeRequired": false`. I verified the delta was only that line in all four before updating them, and the update is a pure addition with no deletions.
- How that surfaced is worth recording: the subagent's focused test list did not include `reducer.test.ts`, so its local run was green and the canonical gate caught the failures. That is my brief's narrowness, not a gap in its work — and it is a good argument for the gate being the authority over any focused run.

Gates on `1dd0edd906`: `make merge-approval-gate` exit 0 (lint, build, full tests including the whole agent suite, native, api-package), `make test-web-browser` 5/5, `make vet` exit 0. Pushed `c6aaa72c60..1dd0edd906` (no force). The round-13 resume/fork escaping tests, the round-14 content-fidelity tests and the new reference-coverage test all pass on this head.

obra:
## roborev round 15 (head c6aaa72)

No critical or high findings. Three Mediums: two in scope, both fixed; one is the item the maintainer already deferred to the follow-up lane.

### Medium — framing-escape bypass via decodable entities · FIXED in 756aa0bb32

Correct, and my round-14 narrowing caused it. The neutralizer covered the literal brackets and the `&lt;` / `&gt;` spellings only, so `&#60;`, `&#060;`, `&#x3c;`, `&#X3C;`, `&amp;#60;` and `&LT;` all passed through, and a reader that decodes character references would see the tag again.

It now escapes the `&` of every reference that can decode to an angle bracket — numeric (decimal, hex, zero-padded), the named `lt`/`gt` references in any case (HTML named references are case-insensitive), and the same references behind one `amp;` layer — and runs the literal pass after that, so its own output is not escaped twice. Anything that cannot forge framing still passes through byte-for-byte, so the round-14 fix holds: a stored URL keeps its `&`.

Test: `TestNotesContextBlockNeutralizesDecodableAngleBracketReferences`, red first on `&#60;/shared-notes&#62;`, covering all six spellings and asserting exactly one framing tag of each kind survives.

One honest residual, because I would rather name the trade than bury it: a reference behind *two* `amp;` layers (`&amp;amp;#60;`) now needs three decodes instead of the four that `html.EscapeString` gave that input. Closing it fully would mean escaping every `&`, which is exactly the URL corruption you flagged in round 14. I chose content fidelity plus one-decode-pass coverage.

### Medium — resume gate not propagated to web/TUI write controls · FIXED in 629caafd8d

Confirmed on both surfaces, with the fence verified to survive the projection each gate reads.

- **Web.** A fenced session presents live+idle with `SharedNotes` deliberately retained, so `canWriteHumanNote` returned true and offered edits the hub refuses. `ThreadModel` now carries `resumeRequired` (`hydrateThread` projects `thread.evener.resumeRequired`, defaulting false for older daemons) and the gate requires it false. The editor and the per-link Remove control both hang off that one gate, so a fenced session renders its saved notes read-only; `canReadSharedNotes` is untouched, and a draft or an already-queued outbox record stays put for retry after resume.
- **TUI.** `hubSessionCapabilities` carries the fence beside the capability set, `sharedNotesWritable` refuses while it is set, and the command guard, palette, slash help, `/notes` and `/url-remove` dispatch, and the drawer's empty-state hint all read that one helper, so they cannot drift apart.
- **Mobile: not affected.** `mobile/src/services/conversation.ts` hardcodes `sharedNotes: false` for the protocol version that client speaks, so it has no notes write surface to gate.

Tests: the gate with and without the fence over a live+idle session, the projection carrying the flag, and the TUI command availability, palette, help and dispatch refusals with an unfenced negative control.

### Medium — stale instance ID in the blur-debounced save · DEFERRED

That is the backlog item already deferred to the follow-up lane (stale `expectedInstanceId` capture), so it stays out of this PR by decision, not by oversight. It is recorded in the follow-up plan.

### Also in this head

Two merge-time main updates, both required because protection is `strict`:

- `99fa1882f0` (eleven commits). One conflict, in `stores/mutationDispatcher.ts`, and it was purely main's relocation of `AppwireClientLike` to `protocol/clientLike` plus this branch's notes types — main changed nothing else in that file, so no logic was traded.
- `cb211c5f86` + `2a8163eb07`. Three conflicts, all real: `agent/events/payloads.go` (both sides added a type in the same spot — kept mine and theirs), `agent/session_lifecycle.go` (their `maybeAppendEnvironmentContext` now returns an error and they drain resume hooks at turn start; adopted both and kept my notes projection inside the same guard), `agent/session_namer.go` (they deleted the environment-tracker reset from the compaction choke point, because environment context no longer folds away the same way; the notes reset stays, since notes turns still do). I also removed a comment of mine that referenced the function they deleted.

Two supporting commits, both worth naming because a reviewer should not have to diff them to trust them:

- `1dd0edd906` updates the four fixture snapshots in `reducer.test.ts`. Carrying the fence on `ThreadModel` is an intentional shape change (it mirrors the existing `askPending ?? false` projection), so those models gain exactly one line, `"resumeRequired": false`. I verified the delta was only that line in all four before updating them, and the update is a pure addition with no deletions.
- How that surfaced is worth recording: the subagent's focused test list did not include `reducer.test.ts`, so its local run was green and the canonical gate caught the failures. That is my brief's narrowness, not a gap in its work — and it is a good argument for the gate being the authority over any focused run.

Gates on `6170b4dd35`: `make merge-approval-gate` exit 0 (lint, build, ROOT_FULL tests, native, api-package), `make test-web-browser` 5/5, `make vet` exit 0. Pushed `1dd0edd906..6170b4dd35` (no force). Main has already advanced again; I will fold that in at merge time rather than chase it here.

obra:
**Round 16 — both findings answered on `6170b4dd35`.**

**Low (`agent/session_notes.go:48-56`, `cmd/evener-tui/details_drawer.go:104-118`) — deferred.** This is the same finding as backlog item 5, terminal injection via unstripped control sequences, already triaged out of this PR. It is tracked in `docs/superpowers/plans/2026-09-12-shared-notes-followups.md` and lands with that list, not here.

**Medium (`agent/session_notes_rpc.go:268-275`, `:294-307`, `:321-334`) — mechanism verified, then deferred by an explicit call and recorded as follow-up item 8.**

The report is accurate about the mechanism. I traced it end to end: the mutators write live state under `Session.mu`, then persist meta.json holding `metaSaveMu` but not `Session.mu`, and on a persistence failure restore the previous value with no emission. `Meta()` (`agent/session_state.go:105`) reads the live notes without `metaSaveMu`, so a concurrent `refreshFacets(facetGoal)` sample can capture the tentative value; `assign` installs it into the cached envelope, and `thread/read` serves that cache (`RefreshThreadEnvelope` runs only at session start and bridge restart).

Three things narrow it below the report's framing, and together they are why I deferred rather than patched it here:

- **Durability and the mutation journal are unaffected.** The metadata save gates the success journal — a persistence failure releases the reservation with no recorded result — so no false success is recorded and the RPC returns the error for the client to retry. The live state is correctly rolled back.
- **The exposure is bounded.** The next `facetGoal`/`facetAll` sample replaces the value, and turn end is `facetAll`; a successful retry also emits the authoritative `NOTES_UPDATED` carrier, which installs the value and bumps the carrier generation.
- **No cheap fix exists.** A rollback repair event contradicts the documented "emits after its successful store mutation" contract, and its client-side behavior needs checking first: a push that contradicts a pending optimistic mutation may be discarded. A committed-notes snapshot read in `Meta()` reintroduces exactly the dual notes authority this redesign removed. Reordering the save seam to persist-before-publish changes the persistence contract that item 4 already reworks.

Item 8 records all three options, plus the `cfg.testOnly.notesAutoSaveFault` hook, which makes the failure path deterministically testable — so the fix will land with a red-first test rather than as correctness by construction.

Nothing else in this round: the other two reviewers reported no issues.

Head is unchanged at `6170b4dd35`. I am integrating current `main` and re-running the gates now.

obra:
**Round 17 — answered on `02dd6f35a8`.**

Two of the five findings are re-raises of items already ruled out of this PR, and three are new. All three new ones are fixed here; the two re-raises stay deferred.

**Deferred, already tracked** (in `docs/superpowers/plans/2026-09-12-shared-notes-followups.md`):

- **Medium, URL removals lack durable retry** (`threads.ts:2818-2834`, `mutationDispatcher.ts:204-214`) — backlog item 3, the accepted limitation. Unchanged; the work is client-side receipt/outbox routing onto the already-journaled backend verb.
- **Medium, TUI rendering lacks terminal-output sanitization** (`details_drawer.go:101-118`) — backlog item 5. Unchanged.

**Fixed:**

1. **Medium, `evener doctor reconstruct` rejects archived notes turns** — `26faa1b76e`.
   `schema.TurnNotesContext` joins the user-role case at `reconstruct.go:439`, with the role table in `reconstruct_test.go` extended to cover the kind (and the crossing-tool-round boundary test). I also checked the class rather than the instance: comparing all 14 schema turn kinds against `reconstruct.go`, `TurnNotesContext` was the only omission, and sweeping every other Go file that switches on six or more turn kinds found no other hard-failing default. Reconstruct was the only surface with this regression.

2. **Low, TUI Run path bypasses the capability check** — `7071d9e9a7`.
   `runHubNotes` and `runHubURLRemove` now go through a new `sharedNotesDispatchRefusal` helper that consults `sharedNotesLiveAvailable`, the same predicate behind the registry's `Available` hook, so dispatch and availability cannot drift. A source that does not advertise the capability now refuses synchronously with "This source does not advertise shared notes." instead of issuing the RPC; a session that merely cannot change notes keeps its existing message. Existing tests that drive the Run path advertise the capability in their fixtures now — same assertions, new precondition.

3. **Low, `daemonStatus` shared-notes mirrors are written but never read** — `4fee9ba3a0` and `02dd6f35a8`.
   The four fields and their assignment in `fetchStatus` are removed; `daemonStatus` has no JSON tags and no consumer read them. A guard test pins the fields out, matching the existing `RawHTTPLogging` absence guard in `appwire/wiretypes_fuzz_test.go`; `02dd6f35a8` is that guard using `reflect.TypeFor`, which the repository's modernize linter requires.

**Verification on `02dd6f35a8`:** `make merge-approval-gate` exit 0 (all nine lint stages and all eight Go module suites pass), `make test-web-browser` 5/5, `make vet` exit 0, and the branch is exactly up to date with `main` (`eeff54b705` is an ancestor of the head).

obra:
**Round 18 — answered on `b93be31a70`** (on top of the round-17 fixes you already reviewed).

**Medium, agent note silently altered by `TrimSpace` — fixed in `b93be31a70`.**

You were right, and the asymmetry was visible in the code: `server/thread_envelope.go:398` trimmed the agent note two lines below the comment explaining that the canonical clamp can land right after a space, and one line below the human note being projected untrimmed for that reason. The line now reads `next.AgentNote = meta.AgentNote`, mirroring the human note.

Red first: `TestThreadEnvelopeSeedUsesStructuredMetaNotes` gained the agent-note clamp-at-space case you described. It failed exactly as predicted before the change:

```
thread_envelope_notes_test.go:39: thread.Evener.AgentNote = "yyy…yyy", want canonical "yyy…yyy "
```

and passes after; the whole `server` package is green.

**Low, `notes_read` reads through independently locked snapshots — deferred, recorded as follow-up item 9.**

It is real, and the code says so itself: `notesProjectionSnapshot`'s comment states that the read is not one atomic cut of the four fields and that callers holding `notesUpdateMu` are what keeps a mutation from landing inside it. `notes_read` is a caller that does not hold it.

I did not patch it here, by explicit call, because every clean fix lands in the same design space as the already-deferred item 8 and none is cheap: `Meta()` cannot take `notesUpdateMu` without blocking the sampling path on the mutators' meta.json I/O, which is the thing that path exists to stay off, and a torn read has no deterministic red test. What the two items need together is an atomically published notes snapshot readers can take without blocking on a writer. Each field `notes_read` returns is internally valid, so the impact is a transient mixed read in agent-facing text rather than durability or display corruption. Item 9 records the mechanism and the constraint so the fix is designed once.

**Verification on `b93be31a70`:** `make merge-approval-gate` exit 0 (nine lint stages and all eight Go
module suites), `make test-web-browser` 5/5, `make vet` exit 0, and the branch is up to date with
`main` (`eeff54b705` is an ancestor). CI is running on this head.

obra:
**Round 19 — disposition on `b93be31a70`, carried to the follow-up lane by explicit call.**

CI is fully green on this head (all fifteen checks, including `race-modules/agent`, `race-root`, `fuzz`, `lint-golangci`, `lint-repository` and `Build snapshot artifacts`), and the branch is up to date with `main`. Rather than open another fix cycle, these findings are recorded for the follow-up branch that owns the notes hardening work, and this PR merges as it stands. I verified each finding against the source first; none of them is a regression in shipped behaviour.

**Carried:**

1. **Medium, doubly-encoded framing reference** (`agent/session_notes_rpc.go`, `notesAngleBracketReference`) — confirmed: the pattern is `&(?:amp;)?(?:#|lt|gt)`, so `&amp;amp;lt;` matches nowhere and passes through into the model-facing copy. The same shape as the earlier literal/entity onion. Recorded as follow-up item 10, together with the tension this same round raises: item 11 wants *less* escaping here, and full robustness against repeated decoding means canonicalizing the model copy, which changes what the model reads. Those two need one design, not two patches.

2. **Low, innocent references rewritten** (`agent/session_notes_rpc.go`, `neutralizeNotesFraming`) — the pattern needs no terminator, so `&lt` without a `;` is rewritten in the model copy while the UI keeps the original. Recorded as item 11 with item 10.

3. **Low, `conversationSignals` counts injected notes context** (`agent/session_init.go:2050`) — confirmed: the loop excludes `TurnHookCompleted` and `TurnEnvironment` only, so a session whose first history turn is `NOTES_CONTEXT` reads as already carrying a conversation. Recorded as item 12.

4. **Low, per-entry snapshot decode in the past-session roster** (`cmd/evener-hub/app_threadread.go`, via `ReadCanonicalHumanNote`) — confirmed: the call decodes the session's whole client-mutation snapshot once per past entry inside the roster loop. Recorded as item 13; the fix is a lightweight canonical-note reader or memoization.

The `urls/remove` durable-retry Medium is unchanged: backlog item 3, the accepted limitation.

Merging this PR now, and the follow-up branch picks up items 2 through 13.

PR #1084, opened 2026-09-10
feat(webui): new-session prompt offers /command and /skill autocomplete

obra:
## Summary

The web "new session" prompt box now offers /command and /skill autocompletes matching what the spawned session would offer:

- New AppWire method `evener/spawn/slashCatalog` (types + catalog entry + hub handler) serving the pre-session command/skill catalog.
- Spawn inline slash menu reusing composer helpers, backed by a debounced pre-session slash catalog hook.
- Pre-session builtins (goal, model, reasoning-effort) with an allowlist and arg resolvers.
- Submit interception so pre-session builtins resolve before the session starts.

Base: 88dff8405. Branch: spawn-slash-autocomplete (10 commits). Plan: docs/superpowers/plans/2026-09-09-spawn-slash-autocomplete-plan.md.

## Test evidence

- AppWire and hub suites pass.
- Spawn suite: 106/106 pass.
- Full npm test: green.
- Final review: APPROVE.

## CI note

CI must run the browser geometry guards (spawnguard/layoutguard); no Chrome was available in the dev env, so those guards did not run locally.

### DIFF
--- a/appwire/client.go
+++ b/appwire/client.go
@@ -735,6 +735,12 @@ func (c *Client) PluginPreview(ctx context.Context, params PluginPreviewParams) 
 	return out, err
 }
 
+func (c *Client) SpawnSlashCatalog(ctx context.Context, params SpawnSlashCatalogParams) (SpawnSlashCatalogResponse, error) {
+	var out SpawnSlashCatalogResponse
+	err := c.request(ctx, MethodEvenerSpawnSlashCatalog, params, &out)
+	return out, err
+}
+
 func (c *Client) PluginInstall(ctx context.Context, params PluginRefParams) (PluginListResponse, error) {
 	var out PluginListResponse
 	err := c.request(ctx, MethodEvenerPluginInstall, params, &out)
--- a/appwire/protocol.go
+++ b/appwire/protocol.go
@@ -199,6 +199,7 @@ var Methods = []MethodSpec{
 	{MethodEvenerPluginDisable, PluginRefParams{}, PluginListResponse{}, ScopeHub, "Disables an installed plugin; returns the updated list."},
 	{MethodEvenerPluginSetAutoUpgrade, PluginSetAutoUpgradeParams{}, PluginListResponse{}, ScopeHub, "Sets an installed plugin's auto-upgrade flag; returns the updated list."},
 	{MethodEvenerCommandList, EmptyParams{}, CommandListResponse{}, ScopeHub, "Lists loaded slash commands (name, plugin, description, source: plugin, project, or user) for catalog/autocomplete display."},
+	{MethodEvenerSpawnSlashCatalog, SpawnSlashCatalogParams{}, SpawnSlashCatalogResponse{}, ScopeHub, "Pre-session slash catalog for the spawn form: the commands and skills a session started with this cwd, harness, and launch overrides would offer."},
 	{MethodEvenerSettingsOverview, EmptyParams{}, SettingsOverviewResponse{}, ScopeHub, "Returns the settings overview field bag: hub/runtime, storage, agent roster, and probed MCP servers — the five template-only settings sections' data."},
 	{MethodEvenerSettingsTranscriptDisplayGet, EmptyParams{}, TranscriptDisplayDefaults{}, ScopeHub, "Reads the canonical Desktop and Mobile transcript-display defaults."},
 	{MethodEvenerSettingsTranscriptDisplayPatch, TranscriptDisplayDefaultsPatchParams{}, TranscriptDisplayPatchResponse{}, ScopeHub, "Updates one transcript-display default using an expected revision and returns the canonical value."},
--- a/appwire/spawn_slashcatalog_test.go
+++ b/appwire/spawn_slashcatalog_test.go
@@ -1,0 +1,48 @@
+package appwire
+
+import (
+	"encoding/json"
+	"testing"
+)
+
+func TestSpawnSlashCatalogParamsMatchThreadStartSpelling(t *testing.T) {
+	data, err := json.Marshal(SpawnSlashCatalogParams{
+		CWD:             "/repo",
+		Harness:         "evener",
+		LaunchOverrides: &LaunchConfigLayer{},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	var raw map[string]any
+	if err := json.Unmarshal(data, &raw); err != nil {
+		t.Fatal(err)
+	}
+	for _, key := range []string{"cwd", "harness", "launchOverrides"} {
+		if _, ok := raw[key]; !ok {
+			t.Fatalf("params JSON = %v, missing %q", raw, key)
+		}
+	}
+}
+
+func TestSpawnSlashCatalogMethodCatalog(t *testing.T) {
+	var found *MethodSpec
+	for i := range Methods {
+		if Methods[i].Name == MethodEvenerSpawnSlashCatalog {
+			found = &Methods[i]
+			break
+		}
+	}
+	if found == nil {
+		t.Fatalf("method catalog missing %q", MethodEvenerSpawnSlashCatalog)
+	}
+	if found.Scope != ScopeHub {
+		t.Fatalf("spawn slash catalog scope = %q, want %q", found.Scope, ScopeHub)
+	}
+	if _, ok := found.Params.(SpawnSlashCatalogParams); !ok {
+		t.Fatalf("spawn slash catalog params type = %T, want SpawnSlashCatalogParams", found.Params)
+	}
+	if _, ok := found.Result.(SpawnSlashCatalogResponse); !ok {
+		t.Fatalf("spawn slash catalog result type = %T, want SpawnSlashCatalogResponse", found.Result)
+	}
+}
--- a/appwire/types.go
+++ b/appwire/types.go
@@ -119,6 +119,7 @@ const (
 	MethodEvenerPluginDisable         = "evener/plugin/disable"
 	MethodEvenerPluginSetAutoUpgrade  = "evener/plugin/setAutoUpgrade"
 	MethodEvenerCommandList           = "evener/command/list"
+	MethodEvenerSpawnSlashCatalog     = "evener/spawn/slashCatalog"
 	// MethodEvenerSettingsOverview returns the field bag behind five settings
 	// sections whose only data path today is Go-template variables:
 	// hub/runtime, storage, agent roster, and probed MCP servers. See
@@ -3040,6 +3041,26 @@ type PluginCheckNowResponse struct {
 type PluginPreviewParams struct {
 	CWD             string             `json:"cwd"`
 	LaunchOverrides *LaunchConfigLayer `json:"launchOverrides,omitempty"`
+}
+
+// SpawnSlashCatalogParams requests the slash catalog a spawn with these
+// inputs would load. Field names and shapes match ThreadStartParams exactly:
+// when this call and a thread/start agree on all three, the menu shows what
+// that start would load. Model, effort, access mode, and prompt text do not
+// affect the inventory and are deliberately absent.
+type SpawnSlashCatalogParams struct {
+	CWD             string             `json:"cwd"`
+	Harness         string             `json:"harness,omitempty"`
+	LaunchOverrides *LaunchConfigLayer `json:"launchOverrides,omitempty"`
+}
+
+// SpawnSlashCatalogResponse is the pre-session slash inventory: the commands
+// and skills a session started with the params would offer. Row shapes reuse
+// CommandDescriptor and EvenerSkillInfo verbatim so the web composer merges
+// them with mergeSlashCommands unchanged.
+type SpawnSlashCatalogResponse struct {
+	Commands []CommandDescriptor `json:"commands"`
+	Skills   []EvenerSkillInfo   `json:"skills,omitempty"`
 }
 
 // PluginPreviewResponse is the launch plugin inventory and structured
--- a/appwire/types.go
+++ b/appwire/types.go
@@ -2886,9 +2886,9 @@ type CommandDescriptor struct {
 	PluginName   string `json:"pluginName,omitempty"`
 	Description  string `json:"description,omitempty"`
 	ArgumentHint string `json:"argumentHint,omitempty"`
-	// Source is "plugin" or "user"; "project" is reserved for a future
-	// project-scoped catalog (project commands are cwd-dependent and never
-	// appear in the hub-wide catalog).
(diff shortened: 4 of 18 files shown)


obra:
Thanks — all five findings confirmed against producer code and fixed in `f067115` (verified locally: Spawn 107/107, spawn helpers 18/18, catalog suite incl. new test, `tsc`, Biome, `gofmt`, `go vet` clean).

- **High: `/reasoning-effort` rejected for new thread** — confirmed: the enum `source()` reads `focusedModel(ref)` (`commands.ts:469-477`), and the ref isn't in `threadsStore` until the session pane hydrates. Fixed by removing the follow-up entirely: the prevalidated value now rides `thread/start` as `reasoningEffort` (`Spawn.tsx` launch-scalar path), so the store is never consulted. New test asserts the value on the start params with no `thread/reasoning-effort/set` call and no toast.
- **High: `/model` not applied (mid-turn Conflict)** — confirmed: `handleAppThreadModelSet` returns Conflict while `processing`/`reservedTurnID` is set (`server/appwire_runtime.go:1845-1858`), and the non-empty first input reserves the turn. Same fix: the split provider/model rides `thread/start` (`modelProvider`+`model`). Test asserts both on the start params plus no follow-up `thread/model/set`.
- **Medium: `modelRequired` blocks `/model` bootstrap** — confirmed: `handleSpawn` returned before prompt parsing (`Spawn.tsx:1150`). Fixed: a valid `/model <known-value>` (validated against the pane catalog) now bootstraps past the guard in both `handleSpawn` and the Start button's `disabled` state. New test: no-default hub + `/model openai/gpt-5` → enabled Start → value on `thread/start`. An unloaded catalog still does not bootstrap (fail-closes the same way `doSpawn` does).
- **Medium: plugin dirs dropped on nonfatal resolver error** — confirmed with a nuance: `thread/start` launches with `Effective.PluginDirs` via `Resolved` (it never sees resolver partials), so the old `pluginDirs = nil` made the catalog emptier than the launch. Fixed: the catalog keeps the effective dirs on the nonfatal path, matching the launch. This also resolves the `lint-golangci` `ineffassign` failure on `pluginDirs` (which was failing CI). New test `TestHubSpawnSlashCatalog_NonfatalResolverErrorKeepsEffectivePluginDirs` fails on the old code (`command "greet" not in catalog: []`), passes on the new.
- **Low: dead `splitModelId`** — confirmed; removed the call and the import in `spawnSlashMenu.ts` (the one real use moved to `Spawn.tsx`'s launch-scalar path).

Notes: `/goal` intentionally stays post-start (`goal/set` has no processing gate — it queues behind the running turn). Bare `/reasoning-effort` still fail-closes pre-start; bare `/model` still fail-opens to default. `static` CI failure is the aggregation wrapper echoing `lint-golangci` only — no independent failure.

obra:
Confirmed and fixed in `6261072` (verified locally: Spawn 108/108, `tsc` clean, Biome clean).

- **Medium: slash scalars lose to Advanced Options** — confirmed: the fix folded the validated slash value into the *chip input* of `resolveScalars`, but `resolveScalars` gives `overrides.model`/`overrides.reasoningEffort` precedence (schema.ts), so a matching Advanced Options value silently won — contradicting the code's own "explicit slash value wins" comment. Fixed by overlaying the validated slash scalars *after* `resolveScalars` instead. New test `a /model prompt wins over a matching Advanced Options model override on thread/start` drives the real Advanced panel (schema stub with a `model` text field), types the override, submits `/model openai/gpt-5`, and asserts the slash value on the start params: it fails on the old code (schema value wins), passes on the new. The analogous `/reasoning-effort` path shares the same overlay lines.

obra:
All five confirmed against producer code and fixed in `7a3bc07` (verified locally: Spawn 108/108, spawn helpers 19/19 incl. new tests, catalog suite incl. new test red-then-green, `tsc`, Biome, `gofmt`, `go vet` clean).

- **Medium: empty-cwd launch omits configured skills** — confirmed: `DiscoverSkills(nil, ...)` returns nil before scanning `extraDirs` (`agent/skill/skills.go:37-39`). Fixed: with no env, the handler scans `resolved.Effective.SkillsDirs` directly via `skill.ScanSkillsDir`. New test `TestHubSpawnSlashCatalog_EmptyCWDScansConfiguredSkillsDirs` fails on the old code, passes on the new.
- **Medium: `none` effort rejected by slash validation** — confirmed: the selector unconditionally appends `{value: "none"}` (`Spawn.tsx:897`) but `resolveSpawnEffortItems` only mirrored `effortOptionLevels`. Fixed: the resolver appends the same `none` entry when the ladder omits it (no duplicate when listed). New resolver tests pin all three shapes.
- **Medium: fixed 300ms sleep** — confirmed. `typeSlashQuery` now `waitFor`s the `evener/spawn/slashCatalog` call instead of sleeping; the non-evener test drops its sleep entirely (it clears mount-window calls after the harness switch commits, then asserts synchronously — post-switch no timer can exist, so no wait is needed).
- **Medium: triplicated enum lookup** — confirmed; extracted one `runEnumItem` helper with only the empty-value policy parameterized, and replaced the unreachable generic fallback (only the three allowlisted ids ever reach it) with a defensive refusal.
- **Low: ref writes during render** — fixed per the suggestion (sync in an effect; the effect runs before the 250ms settle timer can fire). Note the same pattern predates this branch in `usePluginPreview.ts:30-31`, which this hook was required to mirror — left untouched as out of scope.

obra:
Both confirmed and fixed in `2a1fcef` (verified locally: Spawn 110/110, `tsc` clean, Biome clean).

- **Medium: stale `modelCatalog` validates across scope changes** — confirmed: the pane catalog merges snapshots and is never cleared on cwd/harness change, so during the 250ms settle the old scope validates `/model`. Fixed: the catalog commit now stamps its request scope (`harness\0cwd`), and pre-start validation plus the `/model` bootstrap read the catalog only when the stamp matches the current scope — display surfaces keep the merged catalog, validation fail-closes through the mismatch window. New test uses a scope-dependent `model/list` stub (gpt-5 only under the old cwd, new-scope response delayed past submit for a deterministic stale window): fails on the old code (submit goes through), passes on the new (fail-closed toast, no start).
- **Low: `slashToken` leak on submit** — confirmed: success reset clears the prompt without touching the token, and the token only recomputes on change events, so the stale menu survives over the empty prompt on the still-mounted pane. Fixed: `setSlashToken(null)` joins the success reset. New test (menu open → Mod+Enter submit → prompt empty + menu gone) fails on the old code (stale menu div survives), passes on the new.

obra:
Rebased onto current main and adapted in `07721cd` (branch rewritten by rebase; `--force-with-lease` push).

- **Rebase**: main's `19218f6` ("share pure composer and catalog helpers") moved `matchBuiltinInvocation` → `../session/composer/builtinInvocation` (generic) and added shared `findBuiltinArgument`. CI's `static-build` failed on the merge state because our branch still imported the old path. Our files had no main-side conflicts (one unrelated main hunk in `Spawn.test.tsx` applied cleanly).
- **Adaptation**: `Spawn.tsx` imports both helpers from `builtinInvocation`; our three hand-rolled case-insensitive id/label lookups (bootstrap, pre-start validation, post-start runner) now call the shared `findBuiltinArgument` — same semantics, no duplication with the composer's own `resolveCommandResult`.
- Verified locally on the rebased tree: `tsc` clean, Biome clean, Spawn 110/110 + helpers 19/19, catalog backend suite green.

obra:
Both confirmed and fixed in `0377656` (verified locally: spawn suites 130/130 incl. new tests red-then-green, `tsc` clean, Biome clean).

- **Medium: catalog key missing client identity** — confirmed with a nuance: `client` is in the effect deps, but the `latestKey` guard compares only key strings, so a swapped client reusing the same key still commits its late response (proven red: new test `drops a late response from a replaced client` fails with the guard removed). Fixed: a client-generation counter stamped into the request guard drops responses from a replaced client. (The reference `usePluginPreview` hook shares the old shape; left untouched as out of scope.)
- **Medium: `goal/set` awaited before navigation** — confirmed: `doSpawn` awaited the follow-up, holding "Starting…" for the RPC timeout on a slow (not failed) `goal/set` and risking a duplicate session on retry. Fixed: reset + navigate happen immediately after `thread/start`; the goal follow-up fires unawaited afterwards (toasts still surface globally; the pane stays mounted). The `/goal` test now `waitFor`s the follow-up call instead of assuming it landed.

obra:
All four confirmed and fixed in `80597ac` (verified locally: spawn suites 132/132 incl. new tests red-then-green, catalog suite incl. new test red-then-green, `tsc`/Biome/`gofmt`/`go vet` clean).

- **Medium: missing-CWD rejects catalog** — confirmed: `hubCanonicalizeDir` fails for not-yet-created dirs while the spawn flow creates them via preflightDir. Fixed with the second suggested shape: `os.ErrNotExist` falls back to `ResolveUserOnly` (user-level inventory — correct, since no project exists on disk yet), other errors stay `InvalidParams`. New test fails on the old code, passes on the new.
- **Medium: stale-scope effort validation** — confirmed: validation read the merged display ladder while `/model` used the stamped catalog. Fixed: effort validation derives from `scopedModelCatalog` (fallback ladder pre-load, fail-closed on scope mismatch — same rule as models). New test with a delayed new-scope catalog fails on the old code (submit goes through), passes on the new (fail-closed toast, no start).
- **Low: stamp misses refresh triggers** — confirmed. The stamp is now `{scope, loader}`, with `loadCatalog`'s identity as the full refresh-trigger key (client/harness/cwd/instances/generation), and refresh failure explicitly invalidates it. Display surfaces keep the merged catalog.
- **Low: duplicated `none`** — confirmed: `effortOptionLevels` already appends `current`, so `current === "none"` on a none-less ladder duplicated it. Fixed by checking the resolved list; regression test added.

obra:
Confirmed and fixed in `4476ab0` (verified locally: Spawn 112/112, `tsc` clean, Biome clean).

- **Medium: same-CWD refresh exposes stale rows** — confirmed: the hook retains the previous response while refreshing, and the menu merged those rows, so a command removed by the new config stayed pickable (and would submit as literal text). Fixed: the menu merges catalog rows only when the hook status is `ready`; during loading or error it offers the pre-session builtins alone. New test drives a real refresh via the `evener/plugin/updated` notification path (deferred second response): v1 row shown → refresh starts → menu closes → v2 resolves → new row shown. Fails on the old code (stale menu survives refresh), passes on the new.

obra:
CI caught a real race my round-7 change introduced (`web` failure, our `/reasoning-effort high` test): the scoped-effort rule fail-closed whenever the stamp existed but the ladder was unknown — including the normal stamped state where the catalog lists the model with no ladder metadata. Locally the catalog hadn't landed yet (stamp null → fallback), so it passed here and failed under CI timing.

Fixed in `deb5bff`: fail-closed applies only on *proven* staleness (stamp for another scope/loader); unknown ladders with no stamp fall back to the fallback ladder, the pre-load status quo. The known-effort test now waits for the catalog commit via the resolve-commit proxy (proven ordering) with a ladder-less stubbed entry, so it pins the fixed rule — verified red against the broken rule, green on the fix.

obra:
Both worked in `0f14e40` (verified locally: spawn suites 138/138 incl. new tests red-then-green, `tsc`/Biome clean).

- **Builtin/catalog name collisions** — confirmed: a project command/skill sharing a builtin invocation displayed twice while submit always ran the builtin. Fixed by reserving pre-session builtin invocations from catalog rows at merge time (plugin-qualified `/plugin:name` forms never collide and pass through). New test pins single rows for colliding names plus an unaffected control.
- **Builtin text in `thread/start`** (Jesse-approved strip): matched invocations are now consumed like in-session Composer — the start input carries no text item (dormant/configured session) while `/goal` applies post-start and `/model`+`/reasoning-effort` ride launch scalars. This supersedes the earlier dual-handling design. All builtin submit tests now assert empty input; non-builtin prompts (incl. plugin commands/skills) still spawn verbatim for daemon expansion.

Remaining earlier-round items in this push: bare-`/goal` fail-close, picker stamp, Shift+Tab passthrough, highlight reset on reopen (all in `3f7acbe`, awaiting re-review).

obra:
All four confirmed and fixed across `e8812bb` + `516ad9f` (verified locally: spawn suites 139/139 incl. new tests red-then-green, catalog suite incl. new tests red-then-green, `tsc`/Biome/`gofmt`/`go vet` clean).

- **Medium: missing-CWD inconsistency** — confirmed: the user-only fallback omitted ancestor project items that `thread/start` loads after "Create & start". Fixed by resolving the nearest existing ancestor (`nearestExistingAncestorDir`), deriving project identity, and resolving with `ResolveWithProject` — the same ancestor-chain semantics as plugin preview, without its temp-dir churn (a missing leaf contributes no files of its own, so the ancestor chain is the whole answer). New test (missing child under a fixture project) fails on the old code, passes on the new; file-as-cwd stays `InvalidParams` (pinned).
- **Medium: stale-chip effort bypass** — confirmed: `resolveSpawnEffortItems([], chip)` re-appends a missing chip via `effortOptionLevels`, re-authorizing a stale value. Fixed: on proven scope mismatch the resolver gets `""` as current, so the stale chip can't re-authorize. The stale-scope effort test now presets the chip to the stale value (proven red without the fix).
- **Low: duplicated comment** — confirmed (my own copy-paste); deleted.
- **Low: prefill token** — confirmed: programmatic `updatePrompt` writes bypass the keystroke path's caret-aware recompute (centralizing there would break caret accuracy). Fixed at both prefill sites (mount + popstate) with end-of-text caret, where focus lands. New test pins the menu opening with no keystroke (proven red without it).

obra:
Confirmed with a correction to my own prior fix in `92a2328` (verified locally: catalog suite 12/12 incl. new tests red-then-green, `gofmt`/`go vet` clean).

- **Medium: ancestor resolution leaks cwd-anchored config** — confirmed and my prior approach was indeed wrong: `Repo`/`ProjectFile` layers are read from `filepath.Join(cwd, ".evener", ...)` (paths.go), so resolving at the ancestor picks up the ancestor's `launch.local.toml` (loaded unconditionally, no trust gate) while `thread/start` at the empty target reads its own absent file. I initially argued chain-equivalence, but the cwd-anchored layers break it. Fixed per the prescription: missing cwd now goes through the shared `pluginPreviewCWD` probe (empty leaf under the nearest ancestor + target project identity) with `ResolveWithProject` — no temp-dir logic of my own, and my custom ancestor helper is deleted. New no-leak test (ancestor `launch.local.toml` with `skills_dirs`) fails on ancestor-direct resolution, passes on the probe; chain-inheritance tests (git fixture) confirm project items still surface.

PR #1091, opened 2026-09-10
refactor(client): share activity and job output helpers

obra:
Activity parsing, continuation merging and job-output validation are shared framework-independent helpers consumed by the web panel and native client. Delegate revision fencing, timestamps, safe integer offsets and omitted job-output defaults follow one contract across clients.

Continuation pages retain merged entries and counts across late root responses, panel close and remount. Request generations fence stale responses and preserve retries. Counts follow the backend's distinct job/delegate outcomes, sparse usage and omitted-zero wire semantics.

The latest fix shares the turn-aware active-work predicate between default expansion and activity rows. A terminal container with an active turn expands; an empty nonterminal container stays inactive. Both regression cases fail against the old predicate and pass with the fix.

Validation: canonical `make test-web` passed at ee614fd613d8f4cb94e14cdb54cd8f59098fecd8, including all web tests, TypeScript and Biome; 55 focused activity tests passed. Earlier installed API qualification passed. Current head `f55b1b3e1e418c71e60ec759a66805fbd780424a` integrates main's SDK discovery, project deletion, sketches and navigation changes; current-head CI and actual RoboRev must complete.

Keep draft until final gates and actual review pass. Land before #1096. Full remaining work and recovery instructions: #1116.


obra:
Current candidate: `5e576b28d`.

Round 1 (`320b90c` findings): queued root refresh superseding an active continuation — fixed (`bd65795ed`, `5fa14fc29`: root refreshes defer behind a pending continuation and drain after it settles, including refreshes that arrive with no root in flight); row vs badge failure counting — rows now count terminal failures from the outcome alone, matching `agent/jobs_activity.go` (`0e94efe62`, `3f39ec4ed`); turn-container clone overwriting off-target turns — retained for off-target delegates that hold turns (`d07d6ee28`, `cfb75140f`, `0aa507278`).

Round 2 (`0aa5072` finding): a continuation started during an in-flight root refresh — fixed in `bd64695d1`. `beginContinuationFetch` refuses to start a page while a root fetch is pending for the ref and the Load more control is disabled while the root loads, so neither response is discarded and `publishContinuationCounts` cannot claim a bump whose root never landed on that path. Refusing rather than queuing is deliberate: a continuation token is cut against the tree the refresh replaces, and the daemon rejects a live-session continuation whose revision changed. One branch test that had asserted the superseded-root behaviour as a guarantee was re-pointed to the close-path contract it still proves. Main merged in for the strict base.

obra:
Current candidate: `f9d2c7cd1`.

Round 3 (`5e576b2` findings, plus the `fenceRootSession` finding raised on #1096 which carries this code). Continuation requests not serialized across branches: fixed in `d12b27bed` — `beginContinuationFetch` refuses while any page request is pending and every continuation control is disabled while one loads, so neither response is discarded. Row/fold failure predicates diverging from badge counts: fixed in `467047be0` — one predicate per activity kind, matching `agent/jobs_activity.go` exactly (jobs and turns fail only on `failure`; stable delegates only on `failed`/`exhausted`), shared by rows and `summarizeSession`. Descendant turn merge dropping unseen turns, and bounded root refreshes dropping already-loaded pages: fixed together in `6694289ea` as one merge policy — identity-keyed union, authority follows coverage: a non-complete page is a prefix that says nothing past its cutoff (its continuation resumes at exactly that index), so entries loaded from prior pages beyond the window are retained by identity with counts recomputed bottom-up, while a complete page's omissions are real and evicted entries drop; turns union by `jobId`. Lows: the ended-row glyph now follows the terminal outcome (`302841ff4`); the summary-generation sentinel is stated explicitly (`dfc3e542b`, refactor with no reachable behaviour change). Restoring the retained-page behaviour also exposed and fixed a badge published from the pre-fence page. Main merged in for the strict base.

obra:
Current candidate: `a756cff96`.

Round 4 (`f9d2c7c` finding): a loaded child subtree erased during a bounded root refresh — fixed in `a2ee9b033`. The retention rule now applies on the child axis: when a refreshed delegate omits `child`, the previously loaded subtree is kept (with its counts) whenever the incoming branch is not complete, and dropped when it is. That follows the backend exactly: every return in `projectStableActivityDelegate` that omits `Child` first marks the branch errored or truncated (`agent/jobs_activity.go:1056-1085`), and only the completing path leaves the branch untouched, so a childless delegate arrives if and only if its branch is incomplete. Regression covers the truncated refresh (subtree and counts survive) and the complete counterpart (dropped). Main merged in for the strict base.

obra:
Current candidate: `b0d96f946`.

Round 5 (`a756cff` findings). Off-target page rewriting a turn container's state: fixed in `2aa298b56` and `b0d96f946` — authority now picks the whole delegate (a page targeting a descendant is not authoritative for the ancestor container it carries, so the on-screen container keeps its state and only unseen turns plus a monotonic `latestActivityAt` merge in), while a root refresh, which is always the newer snapshot for every delegate it lists, is authoritative for all fields and only unions turns the incoming bounded branch omits, recounting when it retains one. Deferred bump dropped when the summary entry is absent: fixed in `30d942ca6` (the deferred path creates the entry and queues with newest-bump-wins; unreachable via eviction since both stores' evictors run in one pass, but allowed by the store contract). Summary publishes now happen after the panel commit (`d4650a921`, a pure refactor preserving the badge-from-fenced-tree and drain-after-merge ordering). Main merged in for the strict base.

obra:
Current candidate: `0703ac29a`.

Round 6 (`b0d96f9` finding): an omitted optional delegate `type` was read as a turn container — fixed in `e87a09481`. One exported predicate now decides delegate kind for state, rows, merging and counting: `type` omitted or `"delegate"` is the stable form (the only shape the daemon emits; `json:"type,omitempty"`), and any other declared type is a turn container, narrowing to a real container type once one is defined. Nine fixtures that had built containers by deleting `type` — relying on exactly this misreading — now name their kind; no assertion or expected value changed. `0703ac29a` adds type-omitted coverage at the two row/state call sites the merge tests do not reach. Main merged in for the strict base.

obra:
Round 7 disposition for `0703ac2` (one Medium), now at `0557dec36`:

- **`activityMerge.ts:134-139` bounded refresh kept the prior turn object for listed ids** — real. `unionTurns` was current-first, so in `fenceSession`'s incomplete-branch path a listed turn that changed state between snapshots kept its stale object while the incoming counts described the new state. Fixed in `0557dec36`: the incoming list speaks for every turn it lists, prior turns survive only past its cutoff (appended after it), and the recount rule is unchanged (a kept turn still forces `summarizeSession`; when nothing is kept the server's counts already agree with the displayed turns). `revisionFencedDelegate`'s non-authority union (a continuation aimed at a descendant) is deliberately unchanged. Two RED-first tests cover the retained and not-retained cases; the two existing bounded-refresh tests pass untouched.

PR #1096, opened 2026-09-10
Land the native iPhone checkpoint and shared session core

obra:
The Expo/React Native iPhone checkpoint provides saved hub connections, sessions grouped by project with automatic paging, conversation history/live updates, durable drafts, send/queue/steer/stop, questions and approvals, session management, and provider/plugin/settings screens.

The latest review fixes fence stale-cursor recovery to the captured service/sink binding and give each notification subscription its own cleanup. A stale response from service A cannot rebind a store now viewing B, and an older unsubscribe cannot remove the replacement subscription. Dedicated behavioral regressions cover both cases.

Validation: 688 shared tests and TypeScript passed; the full canonical merge gate passed at de48b94d5e63970b5e43844d6646ab76be2ec825. Current published head `8ec90adda8750b894ee841adbb5aaf006c93f939` integrates #1091's turn-aware expansion fix and current main. Fresh CI/RoboRev are running. The local full gate still exercises the earlier 165d head; retain that distinction. Land #1091 first and qualify the final combined head. #1039 and #1102 are merged into this branch, not main.

**Draft integration checkpoint.** A fresh final-source TestFlight archive, physical iPhone installation/update, representative loading measurements and the remaining functionality matrix are still required. Earlier simulator receipts below remain historical evidence, not qualification of the current artifact. iPad and dedicated accessibility work are paused. Full remaining work and repair recovery: #1116.

<details><summary>Original implementation description and earlier evidence</summary>

Bring the existing Expo/React Native iPhone application into main as an integration checkpoint before extracting higher-level web state into the API library. The app includes saved hub connections, sessions organized by project with automatic paging, conversation reading and live updates, durable drafts, send/queue/steer/stop, questions and approvals, session management, and provider/plugin/settings screens.

The review boundary includes the native package, eight shared session modules and their behavioral tests, locked JavaScript/Ruby/Pod dependencies, native patches, build controls, and separate native/package CI jobs. Existing shared paths are retained. The SDK (#1092), pure native helpers (#1094), interruption replay (#1097), and steering fixes (#1073/#1099) are on main; #1091 is the remaining portable activity prerequisite. Environment recovery (#1098) and round timings (#1100) remain separate backend PRs.

Activity refreshes reconcile clustered tools by member identity and revision. A newer live member update survives a held rehydrate while authoritative sibling changes are retained. Failed members split into distinct rows without duplication, including when paging. Canonical attachment ownership prevents stale reads from restoring removed images. Activity pages retain successful merges across late root responses, close and remount; a failed continuation permits a root retry only when that request generation has never published a result. Shared parsing follows sparse usage and distinct job/delegate outcome wire contracts.

Validation at native head `016b118f7`:

- `make test-native`: 720 native tests, 686 shared tests and strict TypeScript pass.
- The complete web gate passes on the exact integrated activity source `77fcabca9`; the independently installed SDK qualification passes at this native head.
- A fresh signed Release simulator build passed using pinned Node 22.13.1 and Xcode 26.6. Artifact source `0097e3841fe9296ebe32da4cb6780cd7d2a8923f` has identical native/protocol inputs to this native head. Installation retained all twelve SQLite tables before launch; after launch the 32 canonical items, seven draft tables, eleven other reader positions and unsent draft remained intact, with provider requests unchanged at 41.
- Owned simulator journeys cover project grouping, conversation controls, question submission, approval deny/allow, reconnect and retained drafts. Independent SDK reads verify backend outcomes.
- The published combined source `8081b4e635f54807b6ad186738e2870e8186275f` passes the full merge gate and an actual backend restart plus app cold launch. All 32 saved canonical items, seven draft tables, eleven other reader positions, the interrupted turn and the visible unsent draft survive; provider requests remain 41.
- The latest combined source `0d75cf8b5fabb9992f2938caf42f2e8129d9c8d1`, including subsequent environment recovery and activity retry fixes, also passes the full merge gate. Its paired-runtime restart and cold-launch check also passes with the same 32 items, drafts, reader positions and provider request count. The newer published combined source `5137914cc` also passes the full gate, but has not replaced the owned backend runtime. It excludes the pending #1100 compaction/timing identity fix; that public-session regression remains failing and is not qualified by the earlier gate.

This is a checkpoint toward iPhone v1. A fresh main-based TestFlight archive and internal availability, physical iPhone installation/update, representative loading performance and the full functionality matrix remain separate acceptance gates. iPad and dedicated accessibility work remain paused. Current-head CI and a clean actual RoboRev verdict are required before merge.

</details>


obra:
Fixed the retained activity crash in `0a1330f084a34e1681867d36897028a768ff4d1c`. Activity retention now records the session reference and thread identity and only supplies a retained tree to ActivityList when both match. The regression constructs a real ActivityList for a replaced conversation and verifies it no longer throws.

Current-head `make test-native` passes: 721 native tests, 688 shared tests, and native TypeScript. The first local run failed nine keybinding checks because the checkout's node_modules lacked the repository's checked-in patches. Running the existing `npm run postinstall` applied them, and the unchanged tests then passed. This was not a Node 26-only issue; do not remove the keybinding flag or membership assertions.

Fresh current-head GitHub CI and actual RoboRev remain required. This source gate does not qualify a new TestFlight or physical-device artifact.

obra:
Current candidate: `fa44d5180`.

Medium (nonzero exit code projected as completed): fixed in `5699cc313`. `toolCallFailed` in `mobile/src/conversation/project.ts` now treats any defined nonzero `exitCode` as a failure, matching the web frontend's predicate, and `commandActivityState` in `mobile/src/state/conversation.ts` reuses the same exported predicate instead of a second copy. Tests cover nonzero exit with no error text, zero exit with no error text, and error text with no exit code, in both projection paths.

Low (stale roster rows on query change): does not reproduce, so no behaviour change. In `mobile-native/src/rosterSearch.ts` `load()`, the object literal passed to `this.publish(...)` is fully evaluated before `publish` reassigns `this.state`, so the `normalized !== this.state.query` comparison reads the previous query regardless of property order inside the literal; a query change already clears `rows`/`hasMore`. `fa44d5180` adds a regression test that a query change clears the previous rows while loading and a same-query refresh preserves them, and it passed against the unmodified code; injecting the described bug makes it fail.

obra:
Refreshed onto main `bc1ed2aef` (the squash of #1091) at `0c19ec0bc`.

The branch had merged several earlier, since-rewritten states of #1091, so 13 activity frontend files conflicted with main's final version. Resolution: main's #1091 code is the result for all 13 (each is now byte-identical to main), after classifying every branch-side hunk: all of them were older drafts of what #1091 later superseded, none was #1096's own work (three independent checks: patch-id identity of the branch's first-parent commits against the #1091 lineage, empty combined diffs for every earlier merge, and blame over every add-bearing hunk). #1096's own contribution in this area, the portable activity list controller under `src/protocol/activityList*`, did not conflict and is the only remaining frontend delta from main; its imports resolve against main's rewritten helpers. `make test-web` and `make test-native` pass on the merge.

The earlier round's `fenceRootSession` finding on this PR was fixed in #1091 and now arrives here through main.

obra:
Round 2 disposition for `0c19ec0` (two Medium), now at `ca3df883c`:

- **`projectThread` compared the turn against `"running"` while the wire sends `"inProgress"`** — real; a snapshot's assistant message was never `streaming`. Fixed through one exported `isInProgressStatus` predicate that every turn and item status check in `mobile/src` now goes through (the active-turn lookup and the three item checks in the conversation store included), so the two cannot drift again. RED-first with an in-progress snapshot; two pre-existing fixtures that set the never-sent `"running"` were corrected to the wire value with their expected values untouched.
- **`turn/completed` merged the single turn's usage into the cumulative conversation usage** — real; after the first turn the counters became that turn's totals. Fixed by removing the merge: the handler publishes completion state only and usage stays whatever the last authoritative projection said. No new refresh was added; no path re-projects usage after completion today (every `requestRehydrate` call site was traced), so the counters now lag until the next projection instead of being wrong. RED-first with two completions carrying different per-turn usage.

obra:
Round 3 disposition for `ca3df88` (one Medium, one Low), now at `f1158ab21` (includes main `b10be62a0`):

- **Truncation ownership keyed by two different identities** — real. Projection and rehydrate recorded ownership under the timeline identity (`transcriptKey ?? id`) while the live delta guards and the lifecycle cleanup used the wire `itemId`, so an item whose key differs from its id could accept a delta after being frozen, and cleanup never released ownership recorded under the key. Every `has`/`add`/`delete` on the ownership set now uses the timeline identity; the audit found four more sites than the two cited (the lifecycle delete, the agent-message delta and reset handlers, and a rehydrate merge that tested wire ids against a key-based set), all converted, with a site-by-site table in the lane's report. Three RED-first store tests: a differing-id item truncated on a non-delta field then refusing a later delta, the cleanup path releasing ownership, and rehydrate not carrying a freeze forward.
- **Failure classification diverged from web** — real. `toolCallFailed` now mirrors web's `hasItemFailure` exactly (trimmed error, failed/interrupted status, nonzero exit code). Review then found the predicate reached reasoning items in the canonical projector while the incremental path did not; web never applies `hasItemFailure` to a reasoning item (its two call sites are unreachable from the reasoning branch, which uses the status only for disclosure visibility), so reasoning items are now excluded from the predicate in the canonical projector too, and the two mobile paths agree with each other and with web. Every caller of `toolCallFailed` is enumerated with its behavioural delta; the one live change beyond the finding is that a completed ask item with a whitespace-only error is no longer excluded from pending, which matches web's trim rule.

obra:
Round 4 disposition for `f1158ab` (two Medium), now at `0dac80b2d` (includes main `c7e6c0968`):

- **Paging deduplication omitted cluster members** — real. The older-page dedup seeded its sets from each row's top-level identity while a member-inclusive identity set (top level, every member's `transcriptKey ?? id`, and the attachment source) already existed for other merges; an older page replaying a member's attachment under a different wire id with the same key was kept as a duplicate. Both sets are now seeded from the union of every current row's identities and each accepted row contributes its full set, with first-occurrence semantics unchanged. RED-first with a differing-wire-id replay and a control that still admits a genuinely new attachment.
- **Member details escaped the byte limit** — real. `truncateItem` bounded only the cluster's top-level detail while native expands members directly. Each member's detail is now truncated by the same rule and an oversized member is frozen under its own identity, so the frozen guard applies to a later delta aimed at it; top-level truncation is unchanged. Holding that required two more changes the review confirmed necessary: the eviction prune retained ownership only for top-level identities and would have wiped a member's freeze on the next lifecycle or page event, so it now prunes against the same member-inclusive set the live-revision map uses; and a duplicated inline byte check was folded into one helper. RED-first for the truncation and for the freeze surviving an unrelated lifecycle event.

Documented residual, pre-existing and out of this round's scope: the streaming delta handlers resolve their target by top-level id only, so a delta cannot reach a non-first cluster member at all; the non-first member's freeze is verified directly through the store's ownership accessor.

obra:
Round 5 disposition for `0dac80b` (four Medium, one Low), now at `b899447a3` (includes main `3f8c9cb25`):

- **Status-less items ignored their turn's status** — real. A sparse live item with no status in an in-progress turn projected as completed. A status-less item now inherits its containing turn's active state in both the canonical projector (which has the turn) and the incremental store path (which derives it from the active turn, guarded on the item's own turn id when it carries one); an item with a status keeps its own. The store's duplicate of the predicate is gone.
- **Sparse reasoning completion dropped a clustered member's output**, and **delta handlers could not reach clustered members** — real, and the second was the residual round 4 documented. One member-aware lookup now resolves a wire item id to a row and, when nested, a member, and one rebuild replaces that member in place preserving every other member's identity, output and truncation state; the reasoning and tool-output deltas and the lifecycle completion all use it, a delta for the first member updates the member rather than only the cluster's top-level detail, the reread fallback survives for unknown ids, and a delta aimed at a frozen member is now refused by the ownership guard round 4 added, which could not be exercised before.
- **Member freezes not preserved across reconciliation** — real; member identities go through the same prior/superseded preservation check as top-level rows. Review then found the widened preservation made a later member's freeze permanent at the rehydrate call site, because the reread set it was checked against was top-level only; the rehydrate decision now uses the member-inclusive reread set, so an authoritative reread that carries the member short unfreezes it and its next delta applies.
- **Paging dedup missed partial cluster overlaps** — real, with one deliberate deviation from the suggested one-liner: an incoming row is skipped when any of its own identities (its own and its members') is already present, but not on the attachment's borrowed source identity, because that form would drop an older page's attachment whenever its source row arrives in the same page; a control test pins that.

Eighteen RED-first cases plus three pinning cases; no existing assertion changed. Recorded, not changed: the lifecycle handler resolves an existing member by wire id and replaces by canonical identity, an inherited asymmetry now named at the call site.

obra:
Round 6 disposition for `b899447` (three Medium, one Low), now at `c563a5d29`:

- **Lifecycle lookup by wire id lost a sparse reasoning completion's output when the id changed** — real, the residual round 5 named. The lifecycle target is now resolved by canonical identity (`transcriptKey ?? id`) first and by wire id second, through the same single walker the delta handlers use, so the lookup and the replacement read one identity. RED-first with a changed wire id over a stable key.
- **Paging cursor and flag disagreed at the retained-item cap** — real. The cap keeps the newest rows, so an older page is exactly what a further load would discard; paging genuinely ends at the cap, and `hasEarlierItems` is now cleared with the cursor rather than left offering a load that was ignored. RED-first with the three facts that must agree (cursor, flag, the next call's outcome).
- **Missing-parent validation threw while an orphan fallback below it was unreachable** — real, resolved fail-closed. The brief first ruled for rendering orphans at top level; the implementer stopped because four existing tests deliberately assert fail-closed ("never silently flatten"), including the redaction test that uses a missing parent as its trigger, and the ruling was reversed. The dead fallback and its map are removed with no reachable behaviour change and no test touched; the two comments that lied now say what the code does. The finding's own fix line understated the job side: orphan jobs would have been dropped, not flattened.
- **Truncation target unclamped below the marker length** — real. The byte limit is the hard contract for every caller (each judges by the byte check; nothing reads the marker), so a limit too small for the marker now yields the longest prefix that fits with no marker, and a zero or negative limit yields the empty string. RED-first at a 3-byte limit.

No existing assertion changed. Noted, deliberate: reaching the cap by count alone ends paging even when the merge discarded nothing.

obra:
Round 7 disposition (head `2d4c7a4d3`, main `dd9727e52` merged):

- **Low, `connection.ts` JSON.parse on SecureStore data** — fixed in `20569e63e` and `1974375b3`. Both parses are wrapped: an unparseable *or wrong-shape* index degrades to an empty roster (`ids()` returns `[]`), and an unreadable record throws the domain error `Saved hub could not be read.` so `token()` never hands back an empty token for a hub that exists but is corrupt; `list()` already drops such records per entry. A record that parses to `null` now also gets the domain error instead of a raw `TypeError` (`1bfd4f581`). Tests cover the corrupt index, the wrong-shape index (`{not:"an array"}`, `["ok", 42]`), the null record, and the stored-extras contract. The `Saved hub index could not be read.` message no longer exists.
- **Low, `providerSignIn.ts` unbounded `intervalSeconds`** — fixed in `6357912ef`. The device-flow poll delay is clamped at `MAX_POLL_SECONDS = 60`; the test pins the boundary (no fire at 59,999 ms, fire at 60,000 ms for a `2**31` interval) alongside the existing in-range case.

`make test-native`: 727 native, 735 shared, tsc clean. No existing test changed.

obra:
Round 8 disposition (head `7752265b7`, main `9ff969396` merged):

- **Medium, rehydrate race loses a frozen member's ownership** — fixed in `8df305d78` and `a7df572d0`. `mergeLiveActivityMembers` now reports which member identities the live side actually won, and only those are recorded as superseded (through the same `activityIdentity` helper the freeze reconciler keys on, so there is one identity vocabulary, not two). A later member truncated and frozen by a live delta stays frozen through a reconcile against a stale snapshot, and later deltas do not append to it; a member the live side did not win stays answerable to an authoritative short reread and unfreezes. Both directions are pinned (`conversation.test.ts`: frozen won member survives the rehydrate; un-won member unfreezes and accepts a later delta while the won one stays frozen).
- **Medium, stale error survives a successful queued page (`activityList.ts`)** — fixed in `4452395d3`: the paged-success publish carries `error: null`. Test drives two queued pages inside one `load()` (the only shape that reaches it, since `load()` clears the error on entry): the first fails, the second succeeds and clears it.
- **Medium, `loadEarlier` dropped during a `refresh` (`jobOutput.ts`)** — fixed in `26cd3a034` and `a3d5976bf`. Both directions are queued now (a refresh during an in-flight earlier page too, since the reconnect effect fires that without a press); invalidation is served before pagination and a failed refresh drops the queued page, the same two rules `ActivityList` applies; a queued success clears the error. Not lifted into a shared helper: the two paginators share a control-flow skeleton but not state (`ActivityList` carries `dirty` invalidation, branch re-resolution and a different published shape), so a helper would be larger than the lines it saves.
- **Medium, duplicated tone classification and redaction logic** — fixed in `d809f1933` and `b9405d433`. The store imports `projectJobEntry`/`projectDelegateEntry` from the service and its copies (classifier, status sets, inline projectors, `FAILED_OUTCOMES`) are deleted; the two copies were verified byte-equivalent before consolidating (same seven failed statuses, same branch order, same redaction allowlist). A 40-case pin asserts the store and the service classify every status identically through their public APIs. The shared surface is the two projectors only; the status sets are `ReadonlySet`.
- **Low, `complete()`'s catch never set `uncertain`** — fixed in `491d16d49`; a failed completion now keeps the explicit re-check prompt on the next `checkStatus()`.

`make test-native`: 731 native, 777 shared, tsc clean; `make test-web` green. No existing test changed.

obra:
Round 9 disposition (head `5575b6714`, main `8fb1ff8f6` merged):

- **Medium, `hubPaths.ts` publishes `result.data` without validating `string[]`** — fixed in `b12f43bc6`: an `isPathList` guard (array, every entry a string) runs inside the existing `try`, so a malformed `evener/paths/complete` reply takes the same generic path-load error as a transport failure and never reaches `state.paths`. The guard is a module-private predicate like the other hub boundaries keep (there is no shared `string[]` validator to reuse). Tests: a non-array reply and an array with a non-string entry each publish the error with `paths` null.
- **Low, the no-config transcript fallback flattens clustered activities without keeping attachments beside their activity** — fixed in `db9020a3d` and `546be6497`: one `projectTimeline` emitter now serves both the configured path and the no-config fallback (member followed by its `sourceTranscriptKey` attachments, trailing attachment rows suppressed), with the fallback differing only in that it flattens the cluster and skips notice filtering. The configured path's output is unchanged (its tests are untouched). Test: two clustered members each with an attachment render as `[a, attachment-a, b, attachment-b]` for both `null` and `undefined` preferences.

`make test-native`: 735 native, 777 shared, tsc clean. No existing test changed.

PR #1098, opened 2026-09-10
Preserve environment transcript identity across hub restarts

obra:
Environment replay can retain obsolete context after a missing/header-only restore or history compaction. Reset environment state at those boundaries, including when compaction-marker persistence fails, while preserving current main's hook and diagnostic exclusions.

Queued environment-context or user-input persistence failures now complete the existing claimed-mutation rollback lifecycle: release the active turn, restore the durable queue and budget, reflect it, and wake the runner. A real filesystem-failure regression proves the failed claim can retry with its stable identity instead of leaving the session stranded.

Validation: the rollback regression was observed failing before the fix; focused normal and race tests pass. The full canonical merge gate passed at 22bf17497d3519b129d1b48539934e0f3a2a34ac. CI's stable-active-ID stop test failed once and passed on rerun; 20 local race repetitions passed, but its intermittent failure mechanism has not been established and is not claimed fixed.

Current head `9f685d7f37c42fbc8a75690c019299989b183f37` integrates current main. Fresh current-head CI and actual RoboRev remain required. This PR is the prerequisite for #1100's compaction/following-input regression. Keep draft pending qualification. Full continuation: #1116.


obra:
Current candidate: `2a12149cd`.

Medium (ambiguous transcript-write failure can duplicate environment entries): real; fixed in `fc12754a1`. When a durable append's sync fails and its rollback also fails, the writer now returns a wrapped `transcript.ErrRollbackFailed` sentinel (error text unchanged), and the environment tracker is no longer rewound unconditionally. `environmentEntryAbsentAfterFailedWriteLocked` establishes durability, reads the transcript for the entry's stable ID, and rewinds (so the next turn re-emits) only once the entry is confirmed absent; an unknown outcome keeps the advanced tracker, which matches what a restart would rebuild from the retained entry. `2a12149cd` covers all three reconciliation outcomes: entry present after failed rollback → no duplicate; entry gone after failed rollback (truncate succeeded, seek failed) → re-emitted under a fresh ID; transcript unreadable or durability unestablished → conservative no-rewind.

obra:
Current candidate: `e4beafe05`.

Round 2 (`2a12149` findings). Medium (ambiguous append leaves live state inconsistent): real; fixed in `5fc8bb140`. When reconciliation confirms the entry is present in the transcript, the session now completes the in-memory side effects the clean path performs — history and pair log inside the same attention hold the clean path uses, then metadata, auto-save and the live `EventEnvironment` — so model history, the live projection and cold restore agree; an unknown outcome keeps the conservative behaviour. Medium (sequence reuse after failed rollback): real; fixed in `244731e96`. The transcript writer now spends the sequence number and the failure observation only when the rollback genuinely could not remove the entry (truncate failed), so the next append does not reuse it; when truncate succeeded and only seek/sync failed the entry is gone and the sequence is not spent. Low (stale `setEnvContextState` comment): fixed in `4f313379a`. `ce6ffc088` makes the zero value of the reconciliation outcome the conservative one and pins it. Main merged in for the strict base.

obra:
Round 3 disposition for `a949ff1`, now at `7d3a1d44a` (includes main `bc1ed2aef`):

- **Unknown outcome after a failed durable append whose rollback also failed left the tracker advanced past what the model saw** — real. The unknown branch now rewinds the tracker to what the model saw, the same as confirmed-absent, so the next turn re-emits the environment block. The accepted cost is one redundant environment entry per failing turn attempt: bounded (each attempt returns an error and aborts the turn), idempotent on restore, and visible as a second Environment row in the projector. An assertion that the tracker never advances past a block in the model history now runs on all three outcomes. `environmentEntryUnknown` is the zero value.
- **A partial line left by a write whose rollback failed could be welded onto by the next append** — real. The writer now refuses every later append (`ErrWriterPoisoned`) once a failed write left bytes at the tail that a rollback could not remove; the seq is spent only for an entry a later reader would see. The same rule covers the buffered `Append` door: a write that transferred nothing leaves the writer usable, a whole line spends its seq, a partial line poisons.
- **Test error contract** — the environment append's error is asserted where it is driven.

Three more hazards in the same writer were closed in this round rather than carried: a sync-failure rollback whose truncate succeeded but whose seek failed left a stale file position, so the next buffered append wrote past EOF over a zero-filled gap (the writer now records that its end is unknown and re-seeks before the next write on either door); the durable door poisoned even when the write transferred nothing (both doors are now one rule); and the buffered door's sync-failure tail left a whole line in the file without spending its seq, so the next entry reused it (it now spends). Every fix is RED-first over a real filesystem under a fault plan; no existing expectation was re-pointed, and `TestEnvironmentRolledBackWriteReemitsEntry` still passes through its original path.

obra:
Round 3 at `7d3a1d4` came back clean (CI green, no RoboRev findings), but main moved to `b10be62a0` (#1139) before the merge, and the base must be current. Refreshed onto it at `184dee1f7` with a merge-only commit; no code change of this PR's own. Requalifying on this head.

obra:
Round 4 disposition for `184dee1` (two High, one Medium, one reviewer; the identical code at `7d3a1d4` was reviewed clean the round before), now at `f913e3190` (includes main `c7e6c0968`):

- **Missing `ulid/v2` dependency** — refuted. `agent/` is its own Go module and `agent/go.mod:9` declares `github.com/oklog/ulid/v2 v2.1.1`; the import resolves and CI compiled and passed on that head. Nothing changed for it.
- **A poisoned writer silently dropped transcript turns** — real as a policy gap this PR sharpened: `recordTurn` warns and continues, which was tolerable when a write failure lost one record, but once the writer refuses further appends every later record was lost while turns kept running (the environment path already failed every turn loudly, but only when an environment block was due). Ruled fail-closed: the transcript's poisoned state is checked before every turn the drain loop runs (not once at admission), and a refused turn returns an error wrapping `ErrWriterPoisoned` with no model request. Review then tightened the refusal to be exactly the failure exit's shape: it ends the input with the same session-end emission the failure exit uses (extracted so the two cannot drift), it does not pop a queued message it is about to refuse (the message stays in the durable queue for after the restart), it settles the processing boundary first, and the shared emitter now reports the session's real wire state rather than a hardcoded idle — so a refusal that leaves a pending question reports `awaiting` and one that leaves queued work reports `active`, and a failed turn with queued work behind it now also reports `active`. RED-first at each step (turns ran and called the model; no session-end was emitted; the queue depth dropped; the state disagreed with `thread/read`).
- **Dirty flag not set after a landed-line buffered write failure** — real; the writer is now marked dirty whenever bytes landed on that path, so `Close` attempts its sync (RED: "Close made 0 syncs").

Recovery from a poisoned writer remains a restart. Recorded residual, pre-existing and out of this PR: the claimed-mutation paths claim or pop their message before the drain loop runs, so a refusal there still consumes it; the fix belongs in those two callers.

obra:
Round 5 disposition for `f913e31` (one Medium, one reviewer done and one skipped), now at `a4a6a20bb` (includes main `3f8c9cb25`):

- **A claimed direct turn was released by comparing the counter to the caller's floor** — real, and in code this PR added with the environment-append failure path: two inputs whose claims interleaved (floors 0 and 1, counter at 2) released wrongly when the first returned first, leaving `AcceptedTurns` inflated by one and max-turn exhaustion early. A claim is a unit and a release now returns exactly that unit (decrement by one behind a zero guard, the same shape the sibling release for queued claims already uses); the floor stays with the claim, which still needs it for the budget check. The failure path was traced to call the release exactly once per failed claim with no second caller, so no reservation token was needed. RED-first in both release orders: the first-returns-first order left the counter at 1 where it started at 0, and the other order passed before the fix, which is why the bug survived.

Not established either way, for the record: whether two claims can be outstanding at once in production needs concurrent entry to `acceptUserInput` for one session; the counter is now order-independent by construction regardless.

obra:
Round 6 disposition (head `70a179358`, main `dd9727e52` merged):

- **Medium, claimed-mutation callers claim/pop before the poisoned gate** — fixed in `35afbb116`. `ProcessClientMutationStart` checks the poisoned transcript before `onRunnable` and before the claim, and `ProcessPendingUserInput` checks before `popQueueHead`, both through one shared `refuseBeforeClaimingOnPoisonedTranscript` that uses the same nil-safe read as the drain-loop gate. A refusal therefore neither spends the start budget nor drops the queued message; if the poisoning is met later on the same path, the claim is handed back (`returnClaimedClientMutationStart` / `completeClientMutationTurn`), keyed on `ErrWriterPoisoned` on purpose: every other pre-incorporation failure already unwinds in place or is reclaimed by startup recovery, and a wider key would re-deliver a turn that is already durable. The refusal only fires on a wake that has work (queued message or pending steering); an idle wake stands down quietly as before (`54bf832eb`, `5bad4e806`). Tests: start budget unspent and queue depth preserved after the refusal; idle wake returns `("", false, nil)`; a steering-carrying wake refuses ahead of any claim (a recording `onRunnable` never fires).
- **Medium, `EventEnvironment` emitted after `attentionMu` release** — fixed in `2af11fde6`. The environment entry and its event are now one publication: the event is emitted while the transcript door is still held, so no concurrent fold can publish between them. This is a deliberate, documented exception to `publishFoldTransaction`'s "no effects under the door" contract (the site comment names the trade: the send can park against the authoritative consumer and runs the job-watch fan-out under the door; a parked consumer already stalls the emitting turn and the door already spans an fsync). Test: at publish time the door is not takeable, and exactly one publication happens.

Known residual, filed as #1150: the other interleaving (a fold wins the door and flushes its own events after releasing it) is untouched; closing it means giving the fold's flush an ordering guarantee it was designed not to have.

Gates: gofmt, vet, full `agent` module, `-race` on the touched sets, deadline audit, pinned golangci-lint 0 issues. No existing test changed.

obra:
Round 8 disposition (head `7282eb53c`, main `96dc079d0` merged):

- **Medium, the idle-wake gate read raw `QueueDepth()` / `hasPendingUserSteering()` and ignored held entries** — fixed in `ca8e0fab1` and `3d74c6b47`. Each claim site now states its own decision once (`queueHeadClaimable` is the whole of `popQueueHead`'s gate; `steeringCarrierRailOpen` + `claimableSteeringCarrierTurnID` the whole of the steering carrier claim's), and the wake gate asks them (`wakeHasClaimableWork`, one snapshot read), so there is one encoding of "is there claimable work" rather than a third copy that knows about `QueueHeld`/`SteeringHeld`. After a Stop parks queued work or steering, an idle wake on a poisoned transcript stands down `("", false, nil)` with the parked work still held; a wake with claimable work still refuses. Tests park through the production surface (queue a mutation then `InterruptClientMutation`; a running start turn with a steer then `InterruptClientMutation`) and were RED against the raw-count gate. The four existing poisoned-wake tests are unchanged.

Recorded as follow-ups rather than changed here: the gate's snapshot and the claim's `mutate` read different locks, leaving a pre-existing check-then-claim window that this change narrows from two reads to one (#1165).

Gates: gofmt, vet, full `agent` module with zero non-ok lines, `-race` on the poisoned/claim/client-mutation/stop sets, deadline audit, pinned golangci-lint 0 issues.

obra:
Round 9 disposition (head `c03a77fb2`, main `daaaea797` merged):

- **Medium, the interrupt drain pops and durably claims a queued message after the turn poisoned the writer** — real; fixed in `2fd9f72bf`. The interrupt-drain branch now runs the same `refuseBeforeClaimingOnPoisonedTranscript` gate the completed-turn drain runs before its pop, so no path claims a message the next iteration's gate will refuse; the refusal joins the turn's error since this branch is the one that returns. Test: a poisoned writer during a turn that ends by interrupt with a queued message leaves the message queued (RED: durable queue depth 0 after the refusal).
- **Medium, the normal drain's check-then-pop window** — this is #1165, already recorded as a follow-up: closing it needs a distinct refused outcome from both claim closures, all four guard sites rewired and the writer's lock taken inside the store's durable `mutate`. Not done in this PR; `2fd9f72bf` neither narrows nor widens it. #1165 stays open.
- **Medium, provisional turn count persisted on input rejection** — refuted. `processOneInput` (`agent/session_lifecycle.go:1511-1518`) registers an unconditional deferred `maybeAutoSave` before calling `acceptUserInput`, so the checkpoint after the rollback at `:2071` always runs; `acceptUserInput` has one caller, so queued and client-start inputs are covered. Proven by causation: deleting that defer makes a rejected input persist `AcceptedInputTurns=1`; restoring it persists 0. The durable-entry-returns-error contract is what `TestEnvironmentAmbiguousWriteDoesNotDuplicateEntry` asserts. Pinned by `5313d7a4d` (a rejected input, then a restart, loads zero accepted turns).
- **Low, nil `w.failures` in `countAppendedEntryLocked`** — refuted. `FailureCounter.Observe` and `Count` have pointer receivers that nil-check first (`agent/transcript/failures.go:112-115`, `:138-142`); a method call on a nil pointer is not a dereference, and `failures_fuzz_test.go:165-169` already asserts that contract. Confirmed on a bare `Writer` through `poisonLandedBytesLocked`: no panic.

Gates: gofmt, vet, full `agent` module with zero non-ok lines, `-race -count=3` on the touched sets, pinned golangci-lint 0 issues.

obra:
Round 10 disposition (head `a33600eff`, main `92561dbe3` merged):

- **Medium, projecting `EventEnvironment` leaves the projector's active turn ID empty and `RecordAppEvent` overwrites the server's** — refuted. `RecordAppEvent` reads `ActiveTurnID()`, not the field, and that accessor falls back to the reservation (`internal/appprojector/appwire_projection.go:2174-2179`); the environment branch restores the reservation on its way out (`:270`, `:272`, `:281`), and `SetProcessing(true)` put the id there via `ReserveTurnID()`, so `server/appwire_runtime.go:442` writes back the same value. `TestEnvironmentPreservesReservedRunnableIdentity` and its siblings already assert this. The server-level coverage the finding asked for is added in `e83a4c048`: `SetProcessing(true)`, a real `RecordAppEvent(EventEnvironment)`, `appActiveTurnID` unchanged, and the following user-input event runs as that id. Causation checked: turning the restore at `:281` into a no-op produces exactly the predicted empty id.
- **Medium, compaction publishes the folded history and resets environment tracking on a poisoned writer** — real; fixed in `3a6ba6b91`. `publishFoldTransaction` swapped history, reset the tracker, claimed the note and stamped the revision before `commitTranscriptsLocked` wrote the markers, and those failures only warned. It now refuses before publication, under `attentionMu` and before `publishFoldedHistory`, returning the existing `ok=false` that both callers already treat as "nothing committed". This is the round-4 admission rule applied to the session's other durable write. Test: `Compact` against a poisoned transcript errors, history and revision untouched, tracker not reset, no `EventContextCompaction`. `TestFallbackCompactionWriteFailureStillResetsEnvironmentTracker` stays green because a zero-byte write failure does not poison.

Gates: gofmt, vet (host, `-tags evenerfuzz`, and `GOOS=windows -tags evenerfuzz`), agent and root module tests with zero non-ok lines, `-race -count=3` on touched sets, pinned golangci-lint 0 issues in both modules.

obra:
Round 11 disposition (head `c48c59751`, main merged):

- **Medium, environment publication races session shutdown** — real; fixed in `2501f17a9`. Both halves of the finding verified: `appendEnvironmentContext` had no shutdown check under the locks it already takes (`agent/session.go:1609-1616`), and `Close` published `SESSION_END` (`agent/session_lifecycle.go:563-580`) without `attentionMu`, so an append could land behind the terminal event. The boundary already exists — the transcript door, which the environment event is published under for exactly this ordering reason — so the fix joins it rather than adding a second one: the append refuses silently when `closingOrClosedLocked()` (same shape as its two neighbouring early returns), and `Close` takes `attentionMu` around the terminal emit, released before `closeAttachedTranscript`, which waits out an append in flight and shuts out any later one. The `RunSessionEnd` plugin hook stays outside the hold. #1109's shutdown-boundary work (`e812703fc`) concerns who emits the terminal boundary and exactly once; it added no lock a late append could observe, so nothing there is duplicated. Test: the append runs from `Close`'s own dispose/sweep seam with `closing` set and `SESSION_END` still ahead (deterministic, one goroutine); RED was one environment turn in model history after the shutdown append, GREEN none, no durable entry, no `EventEnvironment`, and a `SESSION_END` still published.

Recorded: `Close` now holds the door across the terminal emit, an unbounded send bounded by the close budget — the identical trade `appendEnvironmentContext` already makes on the same lock.

Gates: gofmt; vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`); full `agent` module with zero non-ok lines; `-race -count=3` on the shutdown and environment sets; pinned golangci-lint 0 issues.

obra:
Round 12 disposition (head `e527df5aa`, a merge of main `cc9298832` onto `c48c59751`; no code change this round):

- **Medium, rollback truncation succeeds while the rollback `Sync` fails, leaving the writer unpoisoned** — refuted. Measured through the `fuzz/fault` harness, the writer stays unpoisoned in all four failed-rollback outcomes, including the two where the entry provably remains in the file; poisoning is reserved for bytes that landed with no rollback attempted (`agent/transcript/transcript.go:541-560`). Poisoning only this outcome would make the better result (entry removed) fatal while the worse one (entry retained) stays survivable, and `TestAppendDurable_RetainedEntryAdvancesSequenceAndFailureCount` (`cov_s4_transcript_fault_test.go:376-390`) pins that the next durable append succeeds after the retained case. The reviewer's first-choice fix already exists: `reconcileEnvironmentEntryAfterFailedWriteLocked` (`agent/session.go:1746-1763`) keys on `ErrRollbackFailed` and calls `EstablishDurability()`, a real `Sync()`, before deciding. The residual is a crash strictly inside the window before any later durable append, where the file is the pre-rollback file that `OpenWriter` recovery already handles. Whether the writer should adopt one uniform rule for all four outcomes is a design question, filed as #1198.
- **Medium, fold markers commit under `attentionMu` but their live events flush after, so an environment append can publish between** — refuted by reference: this is #1150's first interleaving, with the same two producers the issue names. It cannot be closed on the environment side, which already publishes under the door (`agent/session.go:1692-1704`); the fold's flush must take the same lock, and that flush lives in #1100 and was built not to hold it. Lands as its own PR after #1100, by maintainer ruling.

Gates unchanged from round 11; re-run on this head: gofmt, vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`), full `agent` module zero non-ok lines, `-race -count=3` on the transcript set, pinned golangci-lint 0 issues.

obra:
Round 13 disposition (head `4d8e96993`, main `f2599d1ed` merged):

- **Medium, direct user input's `recordTurn` continues after a partial or poisoned append** — real, and pre-existing on main: this PR's only hunk near `recordTurn` is comment-only, and the append/write/call-site lines blame to `b589a8d263`, `4b384fe790`, `841a199c58`, `56534533c9` (July–September, all ancestors of main). The site is recorded on #1181, the durable-write audit that will move every remaining publish-then-write path to the write-first shape the environment append and #1100's hook completion now use. Not changed here.
- **Medium, commit-phase write failures after the pre-publication poison check** — pre-existing likewise (`adf38dddf9`, `4b384fe790`, `841a199c58`); `3a6ba6b91` added the pre-check only, and in-commit failures are #1100's territory (`ece521c4a`) and already on #1181. Not changed here.
- **Low, the `envTracker` field doc names a helper production never calls and says `mu` alone serializes** — fixed in `41c9c418a`: the doc names `resetEnvContextTrackerLocked`, states the `attentionMu → mu` order and both real callers; the unused helper is deleted and `TestSession_MaybeAppendEnvironmentContext_NoRaceWithCompact` drives the real path (still green under `-race`).
- **Low, `publishFoldTransaction` reports a poisoned transcript as a lost publication race** — fixed in `837b405be`: the fold returns a distinguishable refusal wrapping `transcript.ErrWriterPoisoned`, `foldWithForceCompact` stops retrying on it while keeping the retry for the genuine race, `Compact` returns it, and the two sibling sites (`session_model_call.go:268`, `session_self_compact.go:183`) follow the same rule instead of warning about a race that cannot be won. RED: `compaction error = a concurrent history change won the publication race; try again, want one wrapping transcript.ErrWriterPoisoned`.

Gates: gofmt; vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`); full `agent` module zero non-ok lines; `-race -count=3` on touched sets; pinned golangci-lint 0 issues.

obra:
Round 14 disposition (head `2111513e1`, main `27503c07d` merged):

- **High, `recordTurn` does not propagate `ErrWriterPoisoned`, so the turn whose own input write poisoned the writer still emits `USER_INPUT` and calls the model** — real, and inside this PR's round-4 rule: the drain loop refuses a poisoned turn at the top of each iteration, one iteration too late for the turn that did the poisoning; the hole was the direct-input branch only (the queued and client-start branch already went through `appendTurnAfterTranscriptWrite` with a rollback). Fixed in `800c2563d`: `appendUserInputTurnRefusingPoison` writes before publishing, so a write that stops the writer leaves no history entry to remove; a write error on a now-poisoned writer returns the guards' shared sentinel so the queued restore recognises it; any other write failure keeps warn-and-continue. One rollback, `returnAcceptedUserTurn`, extracted from the environment-append site and used by all three refusals — the two pre-existing copies collapsed into it. Other `recordTurn` callers (attention resolution, steering injection, checkpoint records, the namer) stay with #1181. Test: a partial write that poisons the writer yields an error wrapping `ErrWriterPoisoned`, no second model request, no `EventUserInput`, history and `s.turns` unchanged, durable `AcceptedTurns` restored. RED: `model requests = 2, want only the first turn's`. No existing test pinned warn-and-continue on this path.

Recorded: the direct user turn now writes before it appends to history, the direction the environment append and #1100's hook completion already moved.

Gates: gofmt; vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`); full `agent` module zero non-ok lines; `-race -count=3` on the poison and claim sets; pinned golangci-lint 0 issues.

obra:
Round 15 disposition (head `26f8f1c7c`, main `99fa1882f` merged):

- **High, a tool-result write that poisons the writer mid-input still lets further model rounds run** — real; fixed in `215912545` at the round-loop head: `refuseTurnOnPoisonedTranscript` when `round > 0`, the admission refusal's own function, so sentinel and unwind are one encoding; round 0 is already covered by the drain gate and round 14's input-record check. Other `recordTurn` callers stay with #1181. RED: `model requests = 3, want 2`.
- **High, compaction flushes effects after an in-commit write failure** — refuted by reference for the third round on identical evidence: `agent/session_compaction.go:254-270` and `:557-564` blame to `adf38dddf9`, `4b384fe790`, `841a199c58` (2026-09-02, on main); this PR's only line in range is round 13's return-value change at `:269`. Withheld markers are #1100's `ece521c4a`; the failed-write path is on #1181, now with this exact line range.
- **Medium, a stable environment turn does not advance the live `turn_<n>` fallback while cold counts it** — real, and wider: any stable-id entry diverges (client mutations did so before this PR). Fixed in `fb9065c2d` at `startTurn` rather than in the environment branch alone: every started turn spends exactly one number, with `reservedTurnIDIsStable` marking reservations installed from outside the sequence so only those spend at start. This deliberately changes live turn numbering for client-mutation sessions too; the two existing tests that pin the reservation contract (`environment_reservation_test.go:47`, `appwire_reasoning_replay_test.go:75`) refused a first attempt and pass with the shipped one. Refuting was not available: the direct-input branch builds the user turn with no `StableTurnID`. New `server/appwire_turn_sequence_parity_test.go` (RED: `"turn_1" live and "turn_2" cold` for both stable-id cases).

Gates: gofmt; vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`) in both modules; agent and root module tests with zero non-ok lines; `-race -count=3` on the four touched packages; pinned golangci-lint 0 issues in both.

PR #1100, opened 2026-09-10
Persist round diagnostics for complete transcript replay

obra:
Context compaction can make round diagnostics move or duplicate when a client reloads: live projection assigns them to the active logical turn, while persisted checkpoint/summary and copied recovery history can change replay grouping and coordinates.

Persist the logical owner and durable context-compaction presentation records, then use that owner consistently for timing, compaction, hook and steering projection. Mark copied recovery-tail records so model resume retains them while visible transcript, indexed/item paging and ATIF output avoid double-counting. Bump the transcript index version so existing sidecars rebuild with the new grouping rules. Durable write failures do not publish unpersisted compaction announcements.

**Depends on #1098 (environment replay), whose implementation is included in this branch.** The target is main so the repository CI (which only runs for main-targeted PRs) can qualify the current head. Land #1098 first; the diff then contracts to the compaction/timing work. #1110 merged into this branch, not main. #1129 preserves an additional late-owner paging regression now cherry-picked here and is closed as superseded.

The implementation head `0250b5e31701973d8da38dda475b07a442a51d61` includes the queued-mutation rollback prerequisite and the late-owner regression. Focused server/projector/transcript tests pass; its full canonical gate completed with exit 0. The earlier `34ce` canonical gate and CI passed. Those results do not qualify this new head.

Validation at 34ce: real Session/scripted-provider regressions for compaction with a later turn active, the prior turn finished/idle, and PreCompact-hook steering, followed by another ordinary input. Full and one-item live/cold comparisons cover payload, owner, key and position. Focused server/projector/apptranscript normal and race runs both pass. The previous 9cfee CI failure was missing the environment prerequisite; expected coordinates were not weakened. Durable write failure, recovery-copy, ATIF and settlement regressions are included; `make generate` passed on the implementation before prerequisite integration.

Still draft: finish the current full canonical gate and obtain fresh current-head CI/RoboRev after the base-target correction. Current published head `5da490c7f735df57b3fa90193acf54020b7a9021` now includes the latest main and environment prerequisite; its push triggers the previously missing CI. The local canonical run still exercises 0250, so retain that head distinction. Independently review late owner fragments and concurrent fold publication; finish all affected broad gates before merging. Full continuation and recovery instructions: #1116. Higher-level SDK state extraction and the old Tauri app are outside this change.

## Review round 13

Head `9000bc08d` (three fixes plus a merge of main at `0acebbb0d`).

- **Late fragment captured the running turn** (`e489e5da4`). A metadata record
  for a turn that is already over — a fold's compaction record arriving after
  the next turn opened — started a group, and the next assistant or tool record
  joined *that* group instead of the turn still running. The group a
  continuation joins is now tracked apart from the group most recently opened,
  in both the full read and the bounded index, through one shared predicate; a
  delayed steering carrier stays exempt because it does own what follows it.
  The turn-index version is bumped so cached sidecars rebuild.
- **Marker-dependent effects after a failed marker write** (`15beb088b`). Not
  every failed append poisons the writer — a write that transferred no bytes
  leaves it usable — so a fold whose marker write failed still emitted its
  compaction event, named the session from the marker and injected task-list
  steering, with no anchor on disk. The write outcome now gates those effects
  where they all run, matching the rule withheld markers already follow.
- **Replay copies inflated aggregate totals** (`ff9fe20b9`). The usage, derived
  and failed-tool scans counted a fold's copies as real entries, so a session
  was charged twice for the same round and the gap grew with every fold. One
  shared rule now excludes copies from the figures while still reading them for
  tool-name resolution.
- **`RoundTimings.ToolDefs` (Low): refuted.** `allToolDefinitions` returns the
  cached slice built at init (`agent/session_tools.go:1209-1211`), so no
  tool-definition construction happens in the system-prompt window and nothing
  is misattributed. The field is structurally zero because the phase it names
  costs nothing; no falsifiable test exists for assigning it, so it is left
  alone.

## Review round 14

Head `951f8fa24` (four fixes plus a merge of main at `d7ff88653`).

- **#1200, the preserved suffix dropped after a restart: deliberately not in
  this PR.** Diagnosed in full earlier in this review: a fold's copies are the
  PERSISTED forms of pairs appended since its own snapshot, and the preserved
  window is older than that, so closing it needs a durable per-entry identity
  or a marker-side list of the entries its anchor keeps — a transcript format
  change, >150 lines, affecting every session already on main. It lands as its
  own PR immediately after this one: https://github.com/prime-radiant-inc/evener/issues/1200
- **A self-minted turn name outliving its turn** (`2c43213f0`). The name was
  dropped by the follow-up loop, so it ended only for turns that loop finishes;
  any other path left it standing and `activeTurnOwner` prefers it, handing the
  next turn's timing, hook and compaction records — and anything a fold writes
  while idle — to a turn that was over. The turn now ends its own name on every
  path out of `processOneInput`. An interrupted turn keeps it one record
  longer: the marker saying it was cut short belonged to nobody and now belongs
  to the turn it interrupted, in the entry and in the live event.
- **Fold steering announced after its write failed** (`0048bcd96`). The one
  place this series still broke its own write-before-announce rule. The model
  keeps that steering either way — the fold appended it to live history before
  publishing — so the warning reports the divergence and no client is shown a
  line a reload cannot reproduce.
- **A late timing record classified as a carrier** (`e6778c358`). A round's
  timings describe a round that is over, like a compaction record; only
  steering carries the flow. The turn-index version is bumped again because the
  classification moves group boundaries a cached sidecar recorded.

## Review round 15

Head `46f4dd547` (three fixes plus a merge of main). **The turn-index version
did not move: it stays at 18.**

- **Marker-dependent effects asked the anchor question in pieces**
  (`83185f4d6`). A withheld marker suppressed its own event and the fold's
  steering, a failed marker write suppressed only its event, and the transcript
  reminder asked nothing — so a fold whose replay copies landed and whose
  marker write then failed still published its guidance for a compaction a
  reload cannot find. The question is answered once now, where the marker write
  returns, and the compaction event, the session name, the fold's steering and
  the reminder all read that one answer. A fold that writes no marker is
  untouched.
- **Full read and indexed reader disagreed on one owner combination**
  (`77fb6a798`). After a fragment for an earlier turn, metadata owned by the
  turn still running is that turn resuming, and its next record belongs in the
  group that metadata opened — which the indexed reader already said and the
  full read did not. Only the full read changed, so nothing a sidecar records
  moved and the version stays at 18; the batching concern raised last round
  therefore does not come due here.
- **The self-minted name asked the context instead of the outcome**
  (`fece8f69b`). A turn that finished and returned nil under a cancelled
  context kept its name, with no interrupt marker left to claim it. The rule is
  now named and driven directly by its test, since the race it closes cannot be
  forced through the public path.

## Review round 16

Head `7130dbeae` (three fixes plus a merge of main at `42d27f946`, shared
session notes #1070). **The turn-index version moved: 18 → 19.**

- **A fold's two markers now land, and are judged, apart** (`c86e94de6`). A
  checkpoint and a summary reach the transcript independently; one answer for
  the whole fold let a failed checkpoint suppress the effects of a summary that
  IS in the file — the compaction a returning reader anchors on — leaving the
  session unnamed and its guidance unsent. Each record's effects follow its own
  result; the fold's steering and its transcript reminder ask the separate
  question they actually need, whether a reader coming back finds an anchor at
  all. `compactionTurnWithheld`, written and never read, becomes that
  per-record result rather than being deleted.
- **A closed group could be reopened on resume** (`9db0f91f3`). A record's
  persisted turn id is never empty, so the bounded reader's walk back through
  an indexed prefix read an *unowned* checkpoint as a fragment, stepped over
  it, and named the records after it for an opener that checkpoint had ended —
  the same logical turn twice, with different transcript keys, disagreeing with
  the full read. The walk now stops where the flow stopped. **Index version 19**
  because a sidecar written by the old walk holds those names; this is the
  second bump in three rounds, which is the batching cost noted last round.
- **The doctor read a round boundary from a presentational record**
  (`c73b4b06c`). A round's timings or a compaction record between a tool call
  and its result made the result look like it crossed a boundary, and the whole
  archive was refused. Both join the kinds native history already permits
  inside a pending round.

**Merge conflicts with #1070, and how each was resolved** (both intents kept):

- `agent/session_lifecycle.go` — main's notes-projection comment and
  `maybeAppendNotesContext` call at turn start, kept together with this
  branch's `delegatePreseededTurnID` naming of the preseeded turn.
- `agent/session_namer.go` — main's `resetNotesProjectionAfterCompaction()`
  before the compaction-turn event, and this branch's `OwningTurnID` on that
  event.


obra:
Current candidate: `e0fbcc673`.

Medium (public transcript exposes compaction replay copies): fixed in `21b3c6415`. `publicTranscriptEntry` and `publicTranscriptLine` (the decoded and raw halves of the one public projection) now exclude `context_replay` entries before range and count calculation, so markdown and outline reads show no duplicate entries, correct turn ranges and counts, and each tool-call ID paired once; `ResumeHistory` still includes the copies, with a negative-control test proving the public filter does not reach the resume path.

Medium (compaction ownership stale across overlapping mutations): does not describe the code, so no behaviour change. The owner is captured once at staging (`agent/session_compaction.go:403`) and every publication path reads that capture: the durable checkpoint/summary turn, the PreCompact-injected turns and steering records, the `TurnContextCompaction` record, and the live `EventContextCompaction`, `EventCompactionTurn` and `EventSteeringInjected` events. Live grouping and cold grouping both read that same `OwningTurnID`, so they cannot disagree; the retained staging owner is the intended design (an event that arrives after turn completion stays with its owner rather than moving the current turn's lifecycle). Resolving the owner at publish time, as suggested, is what would create the live/durable disagreement: mutating the live path to the active owner yields `owner=turn_m8, want turn_m7` while the durable record stays `turn_m7`. `22a368ce6` adds an overlap case that accepts mutation B without processing it so the A-staged fold publishes with B active, and asserts live and cold owners equal the staged owner.

obra:
Current candidate: `2321d93f5`.

Round 3 (`e0fbcc6` findings): both real. Direct `ProcessInput` turns had no stable identity, so cold projection numbered them by entry index while live numbered by turns opened, diverging as soon as an environment entry took a stable id; and their compaction records were ownerless, which live grouped under the projector's active turn and cold treated as standalone groups. Fixed in `b99f31930` and `ba8ceb0c1`: a turn that no client mutation reserved names itself with an in-memory `turn_direct_<ulid>` (a namespace outside both `turn_N` and `turn_mN`), carried on the persisted entry via the pre-existing `stable_turn_id` field and on the `USER_INPUT`, `GOAL_CONTINUATION` and `TURN_STARTED` events, and reported by `activeTurnOwner` while it runs; `ActiveTurnID` from a client mutation still takes precedence, so the client-mutation path is unchanged; no transcript-format or turn-index version change. This covers the direct input, goal continuation and unserved notification-wake entry points. Regressions compare full live and cold item identities for an environment entry followed by a direct turn, a compaction fold during a direct turn, a continuation and a wake; `1b7738d10` pins that an interrupt marker on a cancelled direct turn groups identically in both projections. Main merged in for the strict base.

obra:
Round 4 disposition for `2321d93` (three Medium, one reviewer), now at `7be31b7ed` (includes main `bc1ed2aef`):

- **Preseeded delegate input had an empty `StableTurnID`** — real. The preseed path now mints the same self-named direct id `acceptUserInput` would, carries it on the persisted `TurnUserInput` and the later `EventUserInput`, and the run adopts it; a construct-time stamp that could carry no id was dropped (its only reader runs inside a turn, and nothing between construct and create's stamp can start one). Pinned at the agent layer for the entry/event pair; the projection half for that entry shape was proved in round 3.
- **Idle compaction records persisted without an owner** — real. An idle fold now mints one `turn_compaction_<ulid>` and stamps every emitted event and every persisted compaction, checkpoint and steering record of that fold with it; live takes it ahead of the synthetic gap id and cold ahead of the entry index, so both group the same way. No turn-index version bump: the grouping rule is unchanged, only the records now carry an owner. The fold's novelty check moved ahead of the stamp so a re-presented artifact is not read as new (a latent pre-existing misread in owned folds too). Note for readers: `turn_compaction_*` names something that is not a turn; live already published such an id, this makes it durable.
- **Continuation records attributed to the reserved slot holder** — real. Ruling: the turn that is executing owns its records. `activeTurnOwner` now prefers the running direct/continuation/wake turn's self-minted id; `ActiveTurnID` wins only when no direct turn is executing. Client-mutation turns never mint one, so the flip cannot reach them; `TestCompactionOwnerDuringOverlappingMutations` and the round-3 overlap case pass unchanged, and a new server test asserts live/cold parity for the held-continuation case.

All three RED-first with real live-vs-cold divergences; no existing expectation re-pointed. Not closed here: end-to-end live/cold parity for the preseed entry ordinal needs a root-module fixture that drives a real delegate through a tool call, which no server test does today.

obra:
Round 5 disposition for `7be31b7` (one Medium), now at `dfef0a3d3` (includes main `b10be62a0`):

- **Replay-tail copies without a durable compaction anchor** — real on both sides. Read side: when `ResumeHistory` finds no `CHECKPOINT`/`SUMMARY` anchor it now drops entries flagged `ContextReplay` (they are copies by construction; their originals are in the same list), so a resume returns each turn once; the anchored branch and the four existing resume tests are unchanged. Write side: the marker-less fold turned out to be the ordinary path, not an exotic one — every model request publishes a fold whether or not a layer folded, while the compaction markers exist only when one did, so a concurrent durable append in the snapshot-to-publish window used to get a replay copy with no anchor, and with an earlier fold's marker in the file the anchored branch would return that pair twice, which the read-side fix alone cannot cover. The replay tail is now written only when the commit landed a marker. Every reader of `ContextReplay` was enumerated: none wants a copy without an anchor, and the cold projections already drop copies unconditionally, so they never duplicated.

Both RED-first (the read side returned seven turns for four; the write side wrote a copy for a marker-less fold). The predicate that decides "marker landed" is the same one `ResumeHistory` anchors on, and its definition now says so.

obra:
Round 6 disposition for `dfef0a3` (one Medium), now at `215a4182a` (includes main `c7e6c0968`):

- **Preseed adopted the minted turn id before the durable append could fail** — real: a failure after the mint left the id on a runtime retained without a run, and later records from that session would have been attributed to a turn that never ran. The adoption was also redundant, since round 4 made the executing run adopt the preseeded id itself when it starts. The preseed now mints only (the mint and the adoption are separate functions; the direct-input, continuation and wake paths keep both), with no clears added because nothing is set to clear. RED-first with a preseed whose readback fails, asserting the child owns no turn while its runtime is retained; and the success path is now pinned too, since review found nothing previously asserted that the run adopts the preseeded id (the earlier test read a local carried on the event, not the adopted state) — the parity test now checks the executing run's owner while the fixture holds the model call, falsified by deleting the adoption.

Every reader of the adopted id between preseed and run start was traced: there is one production reader, none of whose callers runs on the child in that window. One pre-existing window is named for the record: session-start hooks drain on the child before the turn names itself, on the preseed and direct-input paths alike, so their completion records carry no owner; not a regression, and both projections handle it consistently.

obra:
Round 7 disposition for `215a418` (one Medium, one reviewer), now at `6f7f61540` (includes main `3f8c9cb25`):

- **Idle compaction's synthetic owner stayed `InProgress` live but `Completed` cold** — real. Round 4's idle-fold owner (`turn_compaction_<ulid>`) moved the fold's records onto the item-completion branch of the live reducer, which opens an unknown owner as in progress and never closes it, while the cold projection marks every grouped turn completed. Fixed in the live reducer with no new notification, protocol change or index-version bump: the branch that already emitted a whole completed turn for an ownerless announcement while nothing runs now also fires for a compaction-gap owner, and for that owner it fires whatever turn is running, because the gap id is minted at stage time when nothing runs but its records publish later in the fold's flush, which nothing orders against a turn accepted meanwhile. The ownerless case keeps its nothing-running guard verbatim. The idle-compaction parity test now asserts turn status live versus cold (the round-4 test asserted item identity only), and a projector-level case pins that a late gap announcement completes its own group without touching the running turn's. Review confirmed the gap prefix is disjoint from every other turn-id namespace and that gap-owned steering records land after the group is already complete.

Stated behaviour, unchanged from how the ownerless gap always worked: the group reads completed from its first record and later records merge into it, so the completed status means no turn will continue the group, not that no more items are coming.

obra:
Round 8 disposition (head `73186a7f0`, main `dd9727e52` merged):

- **Medium, replay tail lost when a fold crashes between its marker and its tail — fixed** in `67ed735dc` and `0896dd7c1`, with the read side hardened in `8833bde9f` and the comments swept in `a870fc508`. The fold now writes its replay copies *before* its compaction markers, all inside the one `attentionMu` hold, and every record the fold writes (copies, markers, steering, context-compaction records) carries a `CompactionFoldID` minted at stage time. `ResumeHistory`'s anchored branch reassembles that fold's run by id: the marker, then the fold's steering records, then its tagged copies, then everything after — the same history the old order produced, now crash-safe in both directions. A fold that crashes before its marker leaves tagged copies no anchor claims, and all three readers drop them (no-anchor branch, an older tagged anchor, an older untagged anchor), so the originals survive exactly once; that includes the self-found case of a *later* fold crashing mid-tail after an earlier anchor, which would otherwise have duplicated the stretch. Untagged (pre-existing) markers take the old single-record path, so old transcripts resume unchanged. Tests: crash at the marker, crash mid-tail, later-fold crash, and an exact-order pin on the reassembled history; plus a cold-projection case in `internal/apptranscript` with the copies ahead of their marker (the one projector site that keeps physical replay records inherits the same group either way).

No format-version bump: `FormatVersion 2` marks the semantic format, not the field set; the field is `omitempty` and written only inside a fold, so a session that never compacts stays readable by an older build (the strict-reader one-way door is the existing ruled wontfix pinned by `TestPastThreadReadFailsWholeSessionOnOneUnknownTurnField`).

Known, filed as #1158: a replay copy whose durable write *fails* (rather than crashes) still gets its marker; pre-existing under both orders.

Gates: gofmt, vet, full `agent` module with zero non-ok lines, `-race -count=3` on the resume/replay sets, deadline audit, root `internal/apptranscript`/`internal/appprojector`/`server` packages, pinned golangci-lint 0 issues in both modules. No existing test changed.

obra:
Round 9 disposition (head `3640f17e2`, main `9ff969396` merged; the `tests` job failure at 73186a7 was the racing test named in the second Medium):

- **High, a failed replay-copy write still committed the marker** — fixed in `79a4ec0aa` and `0076bc9df`. When any tail copy's durable write fails, the fold withholds its compaction markers (exactly the records `ResumeHistory` anchors on, filtered by the same predicate the pre-check uses) and keeps everything else it writes: `TurnContextCompaction` records move no anchor, and injected steering is part of the published history. No rollback, no retry: the fold stands in memory, the landed copies carry a fold id no marker claims and every reader drops them, the originals stand, and the next resume replays the pre-fold history — a lost compaction, not lost turns. The warning names the un-anchored fold. The comment states the deliberate live/cold divergence (`EventCompactionTurn`, the namer and task-list steering still run for a fold absent from the anchor on disk). Test: one failing durable copy write through a fault-injecting FS asserts no marker landed, the concurrently recorded turn resumes exactly once, and the earliest record still resumes; the all-writes-succeed path is byte-identical. This closes #1158.
- **Medium, delegate read-back accepted the newest `USER_INPUT` by text alone** — fixed in `8cd9e5b5a`: the read-back now requires the entry's `StableTurnID` to equal the id just minted, in addition to the text; on mismatch the send errors and no run starts (both call sites route the error through the permanent start failure). Test feeds the read-back the state a silent no-op write leaves behind (same text, different id).
- **Medium, `TestDelegatePreseededInputCarriesOneTurnIdentity` reads a shared slice with no happens-before** — fixed in `3aa44e004` and `c78c276de`: the reader goroutine signals once it has recorded the expected event and the test waits on that signal through the package's `awaitWithin` hang guard before asserting; the same shape is applied to the new fold test. RED reproduced byte-identical to CI's failure under `GOMAXPROCS=1`; 20/20 green after.

Gates: gofmt, vet, full `agent` module with zero non-ok lines, `-race -count=3` on the fold-publication and preseed sets, deadline audit, root `internal/apptranscript`/`internal/appprojector`, pinned golangci-lint 0 issues. No existing test changed.

obra:
Round 11 disposition (head `43d93cb46`, main `92561dbe3` merged; ten commits since `32bb3b9dd`):

- **CI, `TestCompactionOwnerDuringOverlappingMutations/late_compaction_steering` (cold `:10:0` vs live `:8:3`)** — root-caused, not the Low below and not the tail-first order: a `TurnHookCompleted` entry was written with an empty `OwningTurnID` because the fold's PreCompact hook completed on the fold goroutine after mutation A had ended, so cold closed the turn and live kept it (pre-existing, reproducible at `6f7f61540` under `GOMAXPROCS=1`). Fixed in `452be85ba` (the fold's PreCompact completion is owned by the fold) and `8bc358512` (`emitHookCompleted` writes the entry before announcing it, this producer's discipline only; #1150 carries the cross-producer rule). The independent review of that round found the session-field owner carrier leaked a stale owner under non-LIFO overlapping folds and stamped a concurrent producer's completion with the fold's owner (hooks run on per-hook goroutines), so `8d2307809` replaces it: the owner rides on the run that dispatched the hook (`hooks.WithRecordOwner`), `runPreCompactHook` takes it as a parameter, and `foldHookOwner` / `hookCompletionOwner` are deleted. Tests: a Notification hook completing beside a parked fold keeps its own owner (RED against the field carrier); the fold's own completion still carries the fold's owner; the write-before-announce order is pinned deterministically through the blocked-send seam (`ebaad8f3e`).
- **Medium, `snapshotDelegateContext` hands the physical transcript to `completedDelegateContext` without dropping `ContextReplay`** — real; fixed in `e85966a40`. A fold's replay copies are filtered before the fork-context delegate sees its history, so each user/assistant/tool-call/tool-result appears once.
- **Low, a leading `ContextReplay` record allocates an empty indexed group and an entry ordinal** — refuted as unreachable: copies come from `persistedAppendLog` (`agent/session_compaction.go:195`, `:255-262`) and nothing enters that log without its turn already written to the same transcript (`agent/session.go:1727-1736`, `:1776-1786`), so no copy can precede its original. Pinned as a characterization case in `622731508`, renamed in `75c84f7da` for what it actually exercises (copies with no opener among them preserve the open group).
- Review follow-ups: the FIFO rendezvous test lives in a `_unix_test.go` file so `GOOS=windows go vet` stays clean (`d42b0d664`, which also updates the two `evenerfuzz`-tagged `runPreCompactHook` call sites); the PreCompact test receives its live hook end instead of racing it and checks every end the stream carried (`901bf98f0`, `893afd283`).

Known residual, recorded rather than fixed here: under `GOMAXPROCS=1` the overlap test still fails about once in 300 subtest runs with `round_timings` cold `:8:2` vs live `:8:3`. That is #1150's second interleaving — a round's timing entry is written, a fold commits and flushes between that write and the round's announce — and reproduces at `32bb3b9dd` and earlier at a far higher rate. The fix is a session-scoped publication lock held across write-then-announce by both producers (60-120 lines, deterministic through the existing seams); it lands as its own PR against #1150 immediately after this one, by maintainer ruling.

Gates: gofmt; vet on host, `-tags evenerfuzz`, and `GOOS=windows -tags evenerfuzz` in both modules; agent and root module tests with zero non-ok lines; `-race -count=3` on the touched sets; pinned golangci-lint 0 issues in both modules.

obra:
Round 12 disposition (head `920278126`, main `eee90925a` merged):

- **Medium, `emitHookCompleted` announces `EventHookEnd` and keeps a live history entry even when the transcript write fails** — real; fixed in `17e62b585`. The completion now goes through `appendTurnAfterTranscriptWrite`, the shape the environment append already uses: entry first, history append only if it landed, warning and return on failure, so nothing a reload cannot reproduce is ever published. Kept the non-fsync write; the finding is live/durable divergence on failure, and an fsync per hook completion would be a latency change on every hook. Test: a fault-injecting transcript write on a hook completion yields no `EventHookEnd`, exactly one warning, and no history or append-log entry.
- **Medium, a withheld compaction marker still fires `EventCompactionTurn`, the namer and task-list steering** — real; fixed in `ece521c4a`. `commitTranscriptsLocked` now records which markers it withheld, and `flush` skips those entirely. The site comment no longer describes the divergence as deliberate; it states the price that remains (a lost compaction, not lost turns). The Task 68 fault-injection test now also asserts no compaction-turn event and no namer call, ordered behind a sentinel drain.

Recorded for #1181 rather than widened here: a marker whose write genuinely *fails* (as opposed to being withheld) still reaches `handleCompactionTurnEffects`, which warns and then publishes anyway (`agent/session_namer.go:483-497`) — the same class, one step out.

Gates: gofmt; vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`) in both modules; agent module tests with zero non-ok lines; `-race -count=3` on touched sets; pinned golangci-lint 0 issues in both modules; overlap test `-count=20` green.

obra:
Round 13 disposition (head `d664c4a20`, main merged):

- **High, retained turns lost across successive folds** — real, and pre-existing on main: the same probe on `origin/main` (`eee90925a`) prints the same result, and a single fold over twelve durable turns already resumes as SUMMARY-only there. The fold's rewrite set is the appends made *since the fold's snapshot* (`agent/session_compaction.go:164,195`), the one case the fold's own checkpoint text cannot cover, while the anchored branch of `ResumeHistory` discards everything before the latest marker (`agent/transcript_read.go:144-146`, unchanged here). Retention has always been carried by the checkpoint/summary prose, not by entries. Closing it means either redefining what a fold rewrites (touching `foldRun`, the anchored branch, ATIF, the fork-context filter and both projections) or a transcript format change; both alter every session's behaviour and exceed this PR. Filed as #1200 for a design pass; not changed here.
- **Medium, cold/live compaction-group closed state** — not reproduced. The two rules do differ abstractly (`logical_turn.go:69-71` vs `appwire_projection.go:1595,1636-1638`), but divergence needs an ownerless continuation record directly after an idle fold's records, and none can be constructed: a compaction-gap owner is minted only when nothing runs (`session_compaction.go:475-478`) and every continuation path opens with a USER_INPUT or an owned record first. Recorded as the fifth residual on #1152, since closing the cold group would also split the fold's own records.
- **Medium, `SteeringInjectedData.OwningTurnID` omitted at several producers** — real; fixed in `204127b77`. One helper, `announceSteeringTurn(owner, data)`; all four producers now pass the owner they just persisted, fold steering included. RED on the real notification path (`live steering … is owned by "", want the notification turn "turn_m1"`); the existing owner-parity test now captures both publications.

Gates: gofmt; vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`) in both modules; agent module tests with zero non-ok lines; `-race -count=3` on the touched sets; pinned golangci-lint 0 issues in both modules; overlap test `-count=20` green.

obra:
Round 14 disposition (head `c7a90750b`, main `3bf357337` merged): the Low is real; `eff7a7e15` adds `CompactionFoldID` to the one list each public path strips (`publicTranscriptEntry`, `publicTranscriptLine`) beside the other correlation fields, since a reader cannot interpret it — the copies it claims are entries this output already drops. RED on both paths with a fold-tagged fixture; GREEN after. Gates: gofmt; vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`); full `agent` module zero non-ok lines; `-race -count=3` on the transcript set; pinned golangci-lint 0 issues.

obra:
Round 15 disposition (head `20afd7c09`, main `99fa1882f` merged):

- **Medium, owned steering is routed to `item/completed` and the TUI drops it** — real, and a regression of round 13's `204127b77`; fixed in `4be675ded` on the TUI side so one wire encoding stands: a `steering` case in `ApplyThreadItem` (append `MsgSteering`, dedupe by item id, `ApplyTieHeadline` for every `<job-notification>` block), and reconciliation on `NotifyItemCompleted` with `item.Type == "steering"` making the same two `TryReconcile` calls as the legacy branch; `MessagesFromThread` rebuilds now render steering too, which the legacy handler never covered. The legacy handler stays and is still reachable for steering drained while nothing runs (empty owner). No double emission: the projector's owned branch returns before the legacy params are built. RED: `messages = [], want one MsgSteering carrying the steering text`, the headline tie, and `expected 1 confirmed msg, got 0`.
- **Medium, live/cold timing coordinates diverge after self-compaction** — refuted: `TestRoundTimingsAfterSelfCompactionRetainLiveReplayIdentity` (`server/appwire_round_timings_compaction_test.go:23`) is that parity case — a genuine self-compaction through the real tool, the round's timing record after it, live vs cold compared over `Key` and `Position`, then again through bounded paging — and is green 10/10 at this head. The only live/cold coordinate divergence reproducible on this branch is #1150's cross-producer interleaving, present on main.

Recorded: wire-shape changes need a TUI-side case; the agent and root gates cannot catch a TUI consumer that silently drops a new item type.

Gates: gofmt; vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`) in both modules; agent module and `cmd/evener-tui` suites with zero non-ok lines; `-race -count=3` on the TUI steering, pending and notification sets; pinned golangci-lint 0 issues.

obra:
Round 16 disposition (head `2a163831a`: round-16 fixes plus a merge of main `f39aa2c83`, which now contains #1098 as a squash):

- **Medium, `evener-doctor reconstruct` rejects the `ROUND_TIMINGS` and `CONTEXT_COMPACTION` records this PR persists** — real; fixed in `1cffc4c38` with the rule reconstruction already applies to attention resolutions rather than a case per kind: skip, count (`historical_presentational_records`), keep them in `source-snapshot.json`, and say so in `RECONSTRUCTION.md`. The archive holds their prose and none of their payload, so a rebuilt record would announce timings nobody measured. RED: `unsupported archived turn kind "ROUND_TIMINGS"`.
- **Medium, a fold whose anchor was withheld still persisted its steering** — real, same class as `ece521c4a`; fixed in `dc90ff686`: steering is withheld with the anchor and announced no more than it is written, so a fold that did not anchor leaves nothing of itself durable except the tagged copies every reader already drops. RED: steering from a fold that never anchored survived a restart.

**Merge over #1098.** This branch carried an older embedded copy of #1098's environment-replay work (via #1110), so the merge conflicted in ten files. Main's reviewed form of #1098 was taken throughout — `returnAcceptedUserTurn`, `appendUserInputTurnRefusingPoison`, the ambiguous-write reconciliation, the emit under the door, the `closingOrClosedLocked` refusal, `queueHeadClaimable`, main's superset of the rollback regression tests — with #1100's own lines re-applied on top (`directTurnID`, `appendDurableTurn`, `delegatePreseededTurnID`, the self-minted turn identity, `commitTranscriptsLocked(anchor)`, `OwningTurnID` on the namer's turn). One adaptation outside the conflicts: `publishFoldTransaction` gained main's fold-refusal return, so a #1100 test call site takes three values. The diff against main is now #1100's own work only.

Gates on `2a163831a`: gofmt; vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`) in both modules; agent module with zero non-ok lines; server, apptranscript, appprojector, doctor and TUI suites; `-race -count=3` over the compaction, fold, owner, steering, hook and environment sets; pinned golangci-lint 0 issues in both modules; overlap test `-count=20` green.

obra:
## Review round 12 (head 38a6d2b09)

- **High, nil writer on hook completion**: refuted. The transcript writer's append is a no-op on a nil receiver, and a session without a state dir is marked ready with a nil writer so held turns release. Pinned in `agent/session_hook_turn_test.go` (a real non-persistent session running its SessionStart hook, plus a completion on a ready session with a nil writer). 157b0d9e9, test only.
- **High, preserved window lost on restart (#1200)**: out of scope for this PR. Reproduced on this branch and on `origin/main` alike. The fold only holds persisted forms for pairs appended since its own snapshot; the preserved window predates that by definition, and the live turns cannot be copied (a tool result's live form retains private API-log evidence). Closing it needs a durable per-entry identity, a read-back of retained entries under the held door, or a marker-side list of kept entries. Each is a transcript-format change affecting every session, so it stays on #1200 as a design decision.
- **Medium, round-timing owner**: fixed in 0cebded14. `events.RoundTimings` carries `OwningTurnID` (excluded from JSON so the item's raw detail stays identical to the cold side), stamped from the same read of the active turn as the persisted entry; the projector routes the event through the owned-announcement path. RED → GREEN plus a projector case with a different turn open.

Gates: toolchain gofmt clean; host, evenerfuzz and windows vet exit 0 in both modules; agent, server, apptranscript, appprojector, doctor, TUI suites clean; `-race -count=3` on the RoundTimings/Hook/Compact/Fold set and the projector set; golangci-lint 0 issues. Merged `origin/main` (f39aa2c83).

obra:
## Review round 13 disposition (head 9000bc08d)

- **Medium, late owned metadata fragment hijacks the continuation group**: real, fixed in e489e5da4. RED: the assistant continuation joined the late fragment (`[turn_active turn_next turn_active]`, want it back on `turn_active`). The continuation target is now tracked apart from the last-opened group in both the accumulator and the index scan through one shared `startsFragmentGroup` predicate; `turnIndexVersion` 16 → 17, so cached sidecars rebuild on first read after this lands. A delayed steering carrier stays excluded from the fragment rule because `TestItemReadersHonorSteeringOwner` pins that a carrier owns what follows it (#1152 family, narrowed for metadata, not closed).
- **Medium, marker-dependent effects on an unpoisoned write failure**: real, fixed in 15beb088b. The unpoisoned path exists (`poisonLandedBytesLocked` leaves the writer usable when no bytes transferred; `appendFailureLocked` returns unpoisoned when rollback succeeds). RED with a marker-write-failing FS and `Poisoned()` asserted false: two compaction-turn events published; GREEN via one guard at the top of `handleCompactionTurnEffects`, covering the fold flush and the `OnCompactionTurn` fallback alike.
- **Medium, replay copies inflate aggregate totals**: real, fixed in ff9fe20b9. RED: usage total doubled after a fold (`{200 20 220}` vs `{100 10 110}`), failures likewise. One shared `countsTowardTotals` predicate across the usage, derived-totals and failed-tool scans; copies are still read so an announced call resolves a later result's name. No TUI or live case is needed: copies are transcript-only, no event is emitted for one, and the item projections already drop them.
- **Low, `ToolDefs` never assigned**: refuted. `allToolDefinitions` returns a cache built once at init, so nothing is misattributed to `SystemPrompt`; the measurement reports 0s and was not landed rather than assert a phase that costs nothing.

Gates: toolchain gofmt clean; host, evenerfuzz and windows vet exit 0 in both modules; agent, server, apptranscript, appprojector, doctor and TUI suites clean; `-race -count=3` on the touched sets; golangci-lint 0 issues both modules. Merged `origin/main` (0acebbb0d).

obra:
## Review round 14 disposition (head 951f8fa24)

- **High, preserved suffix dropped after restart**: #1200, unchanged ruling. The fold holds persisted forms only for pairs appended since its own snapshot, and giving the preserved window a durable representation is a transcript-format change affecting every session. It becomes its own PR right after this one lands. No failing or skipped restart test was added.
- **High, `directTurnID` leaks across turns** and **Medium, cleared too early before the interrupt append**: real, fixed together in 2c43213f0. Probed first: after `ProcessInput` the name was clear, but after a direct `processOneInput` it still held the self-minted id, because the follow-up loop owned the clear rather than the turn. `endSelfMintedTurn` is now the one place the name ends, called from `processOneInput`'s unwind on every path out; the one exception is an interrupted turn, whose marker is written first and now belongs to the turn it cut short, entry and live event alike. RED: "after the turn ended, `turn_direct_…` still owns everything published next" and "the interrupt marker entry is owned by "", want the turn it interrupted"; GREEN for both, plus the acceptUserInput-failure case as a pin.
- **Medium, steering announced after a failed write**: real, fixed in 0048bcd96. `continue` after the warning; RED "steering events after a failed write = 1, want 0". The model still sees that steering because the fold appends the records to live history before publishing; a reload resolves the divergence by not replaying the line, and no client is shown a record no reader can get back.
- **Low, owned `ROUND_TIMINGS` classified as a continuation**: real, fixed in e6778c358. The fragment rule now excludes only records that carry the flow (steering). It changes cached sidecars, so `turnIndexVersion` 17 → 18; the grouping-table timing fixture now carries measurements so its group actually renders.

Gates: toolchain gofmt clean; host, evenerfuzz and windows vet exit 0 in both modules; agent, server, apptranscript, appprojector, doctor and TUI suites clean; `-race -count=3` on the touched sets; golangci-lint 0 issues both modules. Merged `origin/main` (d7ff88653).

Known residual, not in this PR: a mutation-named turn's interrupt marker is still unowned, because that turn releases its durable name in the same unwind; fixing it means moving the release too.

obra:
The residual named above (a mutation-named turn's interrupt marker still unowned) is filed as #1240.

obra:
## Review round 15 disposition (head 46f4dd547)

- **Medium, marker-dependent effects gated on replay-tail success**: real, fixed in 83185f4d6. It was a second publication site, not the round-13 guard reached late: that guard covered the event and the naming, while the fold's steering was gated on the replay tail and the reminder on nothing. `markerLanded` is now computed once where the marker write returns (false for a withheld anchor and for a failed write alike) and gates the compaction event, the naming, the steering write and announcement, and the reminder; `steeringWithheld` is subsumed. RED: a fold whose marker write failed published a `note-handoff` steering for a compaction not in the transcript; GREEN, with a reload assertion that no steering entry is in the file.
- **Medium, full-read vs indexed-read divergence on owned metadata after a cross-owner fragment**: real, fixed in 77fb6a798. Reproduced in the comparison table (`[Y X Y]` indexed vs `[Y X Y Y]` full). The indexed answer is the right one: the running turn's own metadata landing after another turn's fragment is that turn resuming, so the accumulator adopts a fragment as the continuation target when its owner is the target. Only the full read changed, so no cached sidecar is stale and `turnIndexVersion` stays at 18.
- **Low, name not cleared when the context is cancelled but the turn returns nil**: real, fixed in fece8f69b. The gate is `err != nil && isTurnCancellation(...)`, expressed as a named function driven by a four-case table that fails against the old gate; every attempt to force the case end to end made the turn fail instead, so the table is the honest pin.

Gates: toolchain gofmt clean; host, evenerfuzz and windows vet exit 0 in both modules; agent, server, apptranscript, appprojector, doctor and TUI suites clean; `-race -count=3` on the touched sets; golangci-lint 0 issues both modules. Merged `origin/main`.

Reported plainly: of six `-count=20` overlap batches, the first failed and its output was lost to a truncated tail; the following five (100 runs) were clean. The only failure family that test has shown on this branch and on main is #1150's round-timing ordering, which gets its own PR after this one lands, but the failing text was not seen, so that attribution is not claimed.

obra:
## Review round 16 disposition (head 7130dbeae)

- **Medium, `markerLanded` is fold-wide though a fold can write a checkpoint and a summary marker**: real, fixed in c86e94de6. Landing is tracked per record (`compactionTurnLanded []bool`); each marker's own effects follow its own result, and `foldAnchored` answers the separate question the fold-level steering and reminder need. RED: with only the CHECKPOINT line failing (unpoisoned), no compaction-turn event was published though the summary was durable; GREEN: the summary's event, one naming, the steering, and one steering entry after reload. The write-only `compactionTurnWithheld` (the Low) became that per-record result rather than being deleted.
- **Medium, `continuationGroupState` treats an unowned standalone record of an owned kind as a fragment**: real, fixed in 9db0f91f3. It returns `""` for a record that left no group open, asked before the kind, since a persisted `TurnID` is never empty. RED reproduced the reviewer's sequence: the resumed index scan emitted the pre-checkpoint opener as a duplicate id with different keys; GREEN. `turnIndexVersion` 18 → 19, because a sidecar written by the old walk holds those names; only the last bump ships.
- **Medium, doctor ordinal reset by presentational records**: real, fixed in c73b4b06c. RED: "archived tool result at ordinal 6 crosses a tool-round boundary"; GREEN with the result present.

Merged `origin/main` (42d27f946, #1070): two conflicts, both intents kept — `session_lifecycle.go` (main's notes projection at turn start plus this branch's preseeded-turn naming) and `session_namer.go` (main's `resetNotesProjectionAfterCompaction` plus this branch's `OwningTurnID` on the compaction event).

Gates: toolchain gofmt clean; host, evenerfuzz and windows vet exit 0 in both modules; agent, server, apptranscript, appprojector, doctor and TUI suites clean; `-race -count=3` on the touched sets; golangci-lint 0 issues both modules; overlap `-count=20` clean. One pre-merge agent run failed a `/private/var` vs `/var` symlink case in the #1201 family, clean in four runs since; noted on #1201.

PR #1105, opened 2026-09-10
fix(hub): advertise fork capability for local sessions

obra:
The hub supports persisted local thread forks even though the daemon reports no fork capability. Project that authority consistently across read/list responses, lifecycle responses and relayed status, while preserving remote-source capabilities and recovery/ownership fences.

Live in-process subagent aliases remain read-only. Stopped persisted delegates retain their existing fork behavior. Resolve ownership from persisted metadata when the past index is absent or misses, and identify live aliases using the parsed thread ref. Deletion and recovery checks run before metadata admission so their structured errors keep the correct precedence. Recovery notifications explicitly clear stale incoming fork capability; malformed refs and uncertain ownership cannot grant it.

Validation: real persisted root/delegate admission regressions cover absent and missing past-index entries; existing upgrade and deletion-fence regressions pass. Focused capability/admission/relay race tests and golangci-lint passed at 3346c78dd77c8aa983cd0d61be08dd4d07cc8847. The real-roster live-alias regression passes normal and race tests for both missing-index cases and empty/wrong SessionID fields. Removing the admission guard makes both cases fail by allowing a fork, proving the guard is the behavior exercised.

Keep draft until current-head full hub/merge gates, CI and actual RoboRev pass. Earlier failing CI exposed the wire-field lint and error-precedence issues, which are corrected; an earlier green head does not qualify later changes. This PR does not include the separate residual startup/admission extraction. Full remaining work and recovery instructions: #1116.

Current published head `fcd88b2d94229cc9e559676397f4552adeb41074` includes the final live-alias regression and main b76695a8. Fresh full gates and CI/RoboRev remain required.


obra:
The finding that all persisted subagents must remain unforkable conflicts with the existing main-branch contract. `TestHubUpgradeBlocksForkWritesUntilParentStops` exercises cached and uncached persisted delegates in aside, edit and deferred-input modes: an incompatible live owner blocks writes; after that owner stops, the persisted delegate fork must succeed and produce the expected child metadata.

This PR preserves that behavior. Live in-process aliases remain read-only, including when the past index is absent or misses. `TestHubForkAdmissionRejectsLiveSubagentAliasFromRoster` uses real persisted metadata and a real roster with a scripted live owner; removing the admission guard causes both cases to fail by allowing a fork. Empty or incorrect `Thread.SessionID` values are also covered by checking the canonical ref.

The remaining current-review findings about unconfirmed ownership, live recovery fencing, and duplicated helper definitions are actionable and are being corrected. Fresh verification and review will be requested after those corrections; this comment does not mark the PR qualified.

obra:
Current candidate: `f0a0c97bfeff169c79ebf7a15e6068d4eb5f3152`.

Fork capability now checks target-specific unconfirmed ownership and current ResumeLocks state on every relayed publication. Read/list and status stamping share the recovery predicate for ResumeRequired, restartRequired, and resumeRequired active flags. The unsupported top-level resumeRequired field and conflicting unused ownership helper are removed.

Two new causal regressions exercise the actual failure paths: an unconfirmed process-backed rendezvous claim suppresses only its target's fork action, and recovery beginning after an idle subscription suppresses fork on the next closed notification. Both passed, failed with the respective production guards removed, and passed again with the guards restored. Focused race tests passed (3.831s). The full hub race suite passed at d9125f2f2 (285.672s), which has the same production changes; the final commit additionally contains the causal test coverage. A full normal run is being repeated on the final head.

The stopped persisted delegate behavior follows the existing main contract documented in the earlier review-adjudication comment: stopped saved delegates can fork; live subagent aliases cannot. The real live-alias admission regression remains intact.

Fresh exact-head CI and the actual RoboRev verdict are still required; this is evidence for review, not a merge-ready claim.

obra:
Current candidate: `4d71f9467`.

Medium (resume response reports fork unavailable until a later refresh): real; fixed in `eeaf6e3bf`. After `ExplicitResumeCompleted` succeeds, the explicit resume path re-applies the fork capability to the response, so it projects what a subsequent read would.

Medium (subagent predicate diverges between capability and admission): real in both directions; fixed in `0f906979b`. Both `hubCanForkThread` and `hubThreadFork` now call one shared predicate, `hubForkLiveDelegateFenced`, which fences on `Roster.IsSubagentActive` alone. Both sites became strictly stricter: capability dropped the `Kind == "subagent"` conjunct and admission dropped the `entry.Meta.IsSubagent` conjunct. Behaviour change worth naming: a delegate a live parent daemon currently lists can no longer be forked even if its persisted meta carries fork lineage. Stopped saved delegates remain forkable (the parent's roster entry is gone), live aliases remain blocked, and remote capabilities are untouched.

Medium (recovery-signal set mismatch): refuted, no behaviour change. Every recovery signal the projection reads on a local thread already has an admission counterpart: `Evener.ResumeRequired` has one writer, guarded by the same recovery triple that `sessionActionRecoveryError` enforces (plus an epoch check); `Status.Type == restartRequired` is enforced by `refreshDaemonRestartRequiredError` in both fork branches; ownership and unconfirmed-daemon errors are converted to `Unavailable` by `daemonRestartRequiredError`; and `ActiveFlags` containing `resumeRequired` has no producer anywhere in the repo. `a7c50e7c4` adds `TestHubForkAdmissionRefusesEveryProjectedRecoveryFence`, which drives the real handlers through the projected fences and asserts each refusal cause.

Low (ownership resolvability advertised but not checked): documented rather than gated, in `4d71f9467`. The projection runs per thread per list and per relayed status notification and takes no context, while `ownershipEntry` needs one and scans `projects/` on an index miss. The advertise-then-reject mismatch is in the safe direction and is noted at the projection site with a characterisation test.

obra:
Current candidate: `d5614d0ca`.

Round 3 (`4d71f94` findings). Medium (relay `stampForkCapability` ignores a top-level `resumeRequired`): does not describe the protocol, so no behaviour change. `appwire.ThreadStatusChangedParams` declares no such field, its single constructor and both daemon egress stamps are typed to that struct, and the only `resumeRequired` on the wire is `EvenerThread`'s, written solely by the hub's own read projection. `f8ff837df` pins that the params declare no recovery-shaped top-level field (any name containing resume/restart/recovery) and that the fence the notification does carry clears `forkFromTurn` through the real stamp. Medium (relay `ownsFork` from the subscription-time snapshot stays fenced after recovery clears): does not describe the code. For a `local:` relay key the published thread is the daemon's raw read response, never the hub-projected one, so it carries none of the recovery signals; the stamp reads live `ResumeLocks`/roster at publish time. `4da50e109` pins this across a recovery-clear boundary within one subscription (fenced stamp during force-stop, cleared stamp after, same subscription). Medium (roster `SubagentState` counts crash-retained parents): real; fixed in `73d5d3610` and `a3bf12381`. Subagent activity now ignores crashed roster entries in the shared predicate and in the sidebar tree's live set, so a stopped persisted child of a crashed daemon is forkable and no longer shown running; the crashed parent itself still renders `errored`. User-visible: children of a crashed parent now project `notLoaded`/`ended` in read, workspace and sidebar, consistently. Main merged in for the strict base.

obra:
Current candidate: `1c73cc873`.

Round 4 (`9e9e0ef` findings). Medium (stable workspace refs route forks to retired sessions): real; fixed in `71c10479c` and `0e7cb45bd`. The producer is `serve`'s rendezvous registration, which keeps `WorkspaceRef` stable while `thread/clear` moves `ThreadID`/`SessionID`/`InstanceID`; the fork handler now resolves the requested ref through the same live-daemon resolver the read path uses (with the same rendezvous-entry fallback when a probe names no session), so admission and both mutations target the current session while recovery and deletion reservations stay keyed on the requested ref; the live-delegate fence covers both identities, so the RPC remains at least as strict as the capability. Medium (stale roster can permit forks of live delegates): real; fixed in `191f905a9`. The single daemon refresh is hoisted above ownership admission and the fence, and both per-branch refreshes are removed, so admission evaluates the freshly probed snapshot with one refresh per request. Residual, filed for follow-up rather than fixed here: a delegate going live between that refresh and the mutation is still admitted (closing it needs a lease the hub does not hold over another process's delegate bookkeeping). Main merged in for the strict base.

obra:
Current candidate: `46a8ba832`.

Round 5 (`1c73cc8` findings). Medium (`activeFlags` recovery fence advertised but not enforced): fixed in `d3d37f0f2`. The roster prober already read the daemon's status and dropped its `activeFlags`; it now carries them, and fork admission applies the projection's own recovery predicate (`hubForkRecoveryFenced`, the same function, not a second description) to the freshly probed snapshot the round-4 refresh already produces, refusing with the existing recovery error family — no extra daemon round-trip. `TestHubForkAdmissionRefusesEveryProjectedRecoveryFence` now exercises every projected signal through admission. Low (validation after refresh and discovery): fixed in `0db349bfa`. Aside, `sourceTurnId` and `editedInput`/`deferInput` validation runs before the roster refresh and ownership discovery (after the deletion and recovery fences, so no precedence moved); a malformed fork against an unknown local target returns `CodeInvalidParams` and no longer scans the project directories. `78e6a934f` deep-copies the new flags in the navigation live-entry clone. Main merged in for the strict base.

obra:
Round 6 disposition for `46a8ba8` (three Medium, one Low), now at `566806bde` (includes main `bc1ed2aef`):

- **`ActiveFlags` dropped on the local `thread/list` path** — real, and propagating them through `LocalDaemonEntry` would have fixed `thread/list` only: a live `thread/read` is answered by the daemon, whose response never carried the flags. Fixed at the source instead: the fork projection now asks the roster through the same `hubForkLiveStatusFenced` predicate admission uses, so projection and admission read one truth. RED showed both list and read failing.
- **Fences taken on the alias only while the mutation targets the resolved session** — real. The fork now resolves first, samples epochs for both the requested alias and the resolved session, acquires both fences in sorted order (`resumeThread`'s convention) with reversed release, and checks every deletion fence before every recovery fence in request order so the client-visible error kind no longer depends on how the two ids sort. A deletion refusal names the ref the client asked about. After the locks are held the target is re-resolved and a fork whose target moved while it waited is refused with a retryable `Unavailable`, mirroring `resumeThread`; the test drives that through the existing `hubRosterList` seam.
- **Validation after the fences** — real; `validateThreadForkParams` now runs before the refresh, the locks, and every fence. Precedence change stated in the code: a malformed request is answered as malformed even when the target is deleted or recovering.
- **Fingerprint omits `ActiveFlags`** — fixed; flags hashed sorted on a copy, the caller's slice untouched, and a flags-only change moves the fingerprint.

Known residuals, deliberately not in this round: the window between the last fence and `hubForkSession` (round 4's note) is narrower but still open; the relay path's per-notification roster lookup for a cleared daemon falls through to a full roster list; and `ActiveFlags` still has no end-to-end producer, so this is machinery made correct ahead of it.

obra:
Round 7 disposition for `566806b` (one Medium, one Low), now at `4922e2b55`:

- **Post-lock recheck read the roster snapshot** — real. The roster learns a rendezvous session-id change asynchronously, so a `thread/clear` landing while the fork waited on the locks could leave both readings stale and the recheck comparing a snapshot to itself. Fixed without refreshing the roster under the mutexes: after the locks are held the target is re-resolved from the sources `resumeOwnershipStep` reads (the rendezvous entries in the run dir plus `ResumeLocks` state), and the fork is refused with a retryable `Unavailable` when either that reading or the roster's no longer names the pre-lock session. A read failure of the run dir refuses the fork as unverifiable rather than proceeding. Covered by a delayed-watcher test that holds the roster stale while the real rendezvous entry moves, plus a control row proving a legitimate fork still branches. An absent entry or empty run dir still resolves the way the pre-lock path does, so cold-session forks are untouched.
- **`ReadSpawnedThread` dropped `ActiveFlags`** — real; the flags are now copied into the confirmation result, with a new confirmation-path test observing them through `Roster.Find`, the accessor the recovery fence reads.

Both fixes are RED-first; no existing fork, resume or fence test changed. The window between the recheck and `hubForkSession` remains, as noted in round 6.

obra:
Round 8 disposition for `4922e2b` (three Medium, one reviewer), now at `0ccc197a2` (includes main `b10be62a0`):

- **First-match over rendezvous claims could branch the wrong transcript** — real, and round 7's comment was wrong about the downstream guard: `ownershipEntry` refuses a session id found in two project directories, it never sees a second daemon claiming the same workspace alias. The under-lock resolver now collects every local claim on the alias and settles them through the existing `resumeClaimTarget` conflict logic; two claimants naming different sessions refuse the fork with a retryable `Unavailable`, in both directory orders (the RED showed one order silently branching while the other refused). The process probe that settles a conflict runs only on the ambiguous branch and is a local handle check, not a network call. The comment is corrected.
- **Pre-lock and under-lock resolvers disagreed in steady state** — real: the resume redirect is never cleared, so after a resume through a stable alias and a graceful shutdown every cold fork of that alias was refused as "session ownership changed". Both resolvers now end in one rule: roster (or rendezvous) first, then the same `ResumeLocks` redirect, then the alias itself. A redirect naming a session no longer on disk fails at `ownershipEntry` rather than branching anything; when a redirect should expire is a `ResumeLocks` question, not this PR's.
- **Deletion fence after refresh errors** — real; the durable deletion state for the requested alias is now read before the refresh, so a deleted target gets `MutationOutcomeTargetDeleted` even when daemon discovery fails, and the locked deletion pass afterwards is unchanged. A live target with a failing refresh still gets the refresh error.

All three RED-first; no existing fork, resume or fence test changed. Accepted as is: two exited claimants with a redirect still refuse (the conservative direction for a mutation), and the refusal wraps `resumeClaimTarget`'s own wording.

obra:
Round 9 disposition for `0ccc197` (two Medium), now at `ee0d6f794` (includes main `f21e1c595`):

- **Capability projection never consulted the `DeletionStore`** — real, and it closes a reachable window rather than a theoretical one: the list path reads no deletion state, and a target that project-deletion cleanup skips stays fenced with its files present until the startup retry, so such a thread was listed with `ForkFromTurn: true` for an action the fork RPC refuses terminally. The projection now performs the same unlocked deletion read the fork's pre-refresh check uses, for the requested ref and the resolved session, and advertises `false` for a fenced thread. RED-first on three fenced rows with two unfenced controls that still fork.
- **Under-lock resolver counted crash-retained rendezvous claims the roster excludes** — real; after a daemon cleared to a new session and crashed, the two resolvers disagreed and every fork through the alias was refused as "session ownership changed". The under-lock collection now applies one liveness rule: a claim whose process is confirmed gone is not a live owner (the same case the roster marks `Crashed`), an undecidable process stays a claim so an ambiguous alias meets the existing ambiguity refusal rather than a disagreement. The comment names both liveness mechanisms (the roster's signal check and the fork's process-handle probe), which agree on the decisive case, and notes that a contested alias is probed twice. The probe is local (a pidfd on Linux, an `O_NOFOLLOW` fd on macOS), runs once for the ordinary single-claim fork, and is far cheaper than the roster refresh that precedes it. RED-first with a crash-retained marker and a roster showing nothing live; both resolvers now answer the same session.

One existing fixture changed with its assertions untouched: round 8's two-daemons-claim-one-alias test faked every process as exited, which under the liveness rule are no longer claims, so its stub now returns live processes, which is what the test's name means; it still refuses in both directory orders through the true conflict path.

obra:
Round 10 disposition for `ee0d6f7` (one Medium, one Low), now at `73f06d3ea` (includes main `c7e6c0968`):

- **Projection fenced recovery and live status for the requested alias only** — real; the fork RPC fences both the alias and the session it resolves to, and round 9's deletion projection already resolved both. The capability projection now resolves the target once, only for threads the cheap alias fences admit, and applies the deletion, recovery-state (`ResumeRequired` or `Stopping > 0`), live-delegate and live-status fences to both identities. RED-first against a real daemon behind a real relay reached through its stable workspace ref, with rows for each fence kind that assert up front the alias resolves elsewhere and the alias's own recovery state is clear, so only the resolved session can account for the answer; all three surfaces (projection, `thread/list`, `thread/read`) and the RPC's refusal are asserted together. Review confirmed the projection's fence set now matches the RPC's for both identities, modulo the documented ownership-resolution gap and the under-lock refusals that have no projectable form; the alias-side and resolved-side checks point at each other so a fence added to one is visibly owed on the other.
- **Check-order test assumed two generated session ids sort in generation order** — real, a flake in a test this queue added: UUIDv7-style ids do not guarantee that within a millisecond. The test now sorts its two ids and assigns the sorts-first and sorts-second roles from the sorted pair, which still exercises both orders, and passes `-count=200`.

No existing test changed beyond the flaky one's own assertion.

obra:
Round 11 disposition for `73f06d3` (one Medium), now at `9ee4eb443`:

- **RPC live-status fence covers the alias only while the projection fences the resolved session too** — refuted as a reachable divergence, then closed as a class. With a live daemon the alias (workspace-ref scan) and the resolved session (`Find`) land on the same roster entry, since the resolved id is that entry's own session id, so the flags read are the same fields of the same record; with no live daemon the resolution comes from the resume redirect and neither identity has a roster entry, so both answer false. The one asymmetry that exists runs the safe way: a probe that answered without naming its session makes the resolved lookup miss and answer false while the alias answers true, and both paths already catch that through the alias. Pinned by a test that asserts the two ids differ, both calls agree, the answer is genuinely true when the daemon reports the flag, and the RPC refuses, with a no-flags control that forks. Because rounds 9 through 11 were all asymmetries between two encodings of one rule, the RPC's live-status fence now iterates the same fence-target list its recovery and deletion passes use, so the RPC and the projection cite one rule per fence and point at each other. No behaviour change, so no regression test is claimed for the refactor itself; every existing fork, resume and fence test is untouched.

Recorded for a follow-up rather than this PR: a single per-identity predicate both paths call would make divergence unrepresentable rather than merely reviewed; it needs the two alias-side helpers reworked along with the tests that call them directly.

obra:
Round 12 disposition for `9ee4eb4` (two Medium, one reviewer done and one skipped), now at `8205599de` (includes main `3f8c9cb25`):

- **Pre-refresh deletion check examined only the requested alias** — real, round 8's precedence fix extended to the second identity: for a stable alias resolving to a different session, a failing roster refresh masked the resolved session's terminal `MutationOutcomeTargetDeleted` with a retryable error. Before the refresh the target is now resolved from the last trusted ownership (the roster's current snapshot plus the resume redirect, the same resolver the post-refresh pass uses) and the unlocked deletion read covers both identities; that pre-refresh answer is used for nothing but the terminal refusal, so a stale reading can refuse but never admit, and the post-refresh resolution and every locked pass are unchanged. The read is skipped on a hub with no deletion store. RED-first with the resolved session durably deleted and the refresh faulted, and a live-resolved-session control that still gets the refresh error. Accepted cost, stated in the record: if the alias moved while the roster still names a deleted predecessor, that request is refused terminally where a refreshed resolution would have proceeded; a later attempt succeeds.
- **Relayed capability broadcast to every connection while the RPC fences per connection** — not implemented: a connection established before a recovery completed can be shown a fork it will be refused (with the retryable fresh-connection recovery error) until it reconnects, and projecting per connection is a design change beyond this PR. Recorded as a named gap in the projection's list of deliberate gaps, with the relay site and the refusal, and the RPC remains the stricter side (an offered action that is refused, never a fork that should not have happened). Follow-up filed with the coordinator.

No existing test changed; lock order and release untouched.

obra:
Round 13 disposition for `8205599` (one Medium), now at `049f26b98`:

- **A single unverifiable rendezvous claim could admit a fork** — real, and it corrects round 9's rule for the single-claim case: the liveness probe treated every verification error other than "exited" as proof of a live owner, and with one claim the resolver returned it without verification, so a malformed, stale or PID-reused entry could become the fork's target with ownership never established. Verification now has three outcomes: a claim whose process verifies is live; one whose process has exited is dropped as before; one whose verification fails for any other reason is unverifiable, and a fork resolving through it is refused with the retryable ownership-verification error carrying the reason, never admitted and never silently dropped. That matches the posture the resume path has always taken on an unverifiable process. The refusal is logged with the alias, PID and error, so a hub whose process handles cannot be opened (permissions, a hidden `/proc`) is diagnosable, and the capability projection's list of deliberate gaps records that it never probes a process and can therefore advertise a fork such a host will refuse. RED-first: the unverifiable claim branched a transcript before the fix; the verified control still branches the right session; the exited case keeps round 9's coverage.

Seven existing fixtures wrote a bare self-PID rendezvous entry and were counting as live only because the real probe refused a self-PID with a non-exited error; each now supplies a live process stub, with names, rows and assertions unchanged. Stated, not tested: the scan returns on the first unverifiable claim, so a mixed alias (one unverifiable, one verified) refuses as unverifiable rather than reaching the ambiguity path.

obra:
Round 14 disposition (head `0707f7c93`, main `dd9727e52` merged):

- **Low, pre-lock and under-lock resolvers disagree when only one ownership source is configured** — fixed in `8a6845338` and `f1d3c4361`. Each resolver now falls back to the other's source when its own is absent: the under-lock resolver (rendezvous first) consults the roster when there is no run dir, and the pre-lock resolver (roster first) scans the rendezvous when there is no roster. Both still end in the shared redirect/alias tail, and with both sources present neither fallback runs, so the production path (which always has both: `main.go` defaults the run dir before building the roster) is unchanged and the rendezvous stays authoritative. Coverage: a table-driven test with `roster only, no run dir` and `run dir only, no roster` rows, each asserting the fork succeeds *and* branches the session the configured source names, plus a `run dir only, claim cannot be verified` row asserting the retryable `cannot verify session ownership` refusal (not "ownership changed") and exactly one `fork refused:` log line. The verification-failure log moved from the shared resolver to `hubThreadFork`, where the refusal actually reaches the client, and the error now carries the claim's identity.

Known, documented: the roster-only fallback resolves through `liveDaemonForThread` and does not reach `resumeClaimTarget`'s duplicate-claim settlement; production never takes that shape. Follow-up for the resolver/fence structure: #1146.

obra:
Round 15 disposition (head `e66713796`, main `9ff969396` merged):

- **Medium, `forkRedirectSessionID` follows one redirect hop** — real, fixed in `91dd1afb4`. Verified at the writers first: `PersistForceStop` writes `ResumeSessionID` onto the alias group it is handed and `RecordResolvedSession` onto whichever group an alias currently points at, so A→B and B→C both stand and one hop branched the retired B. The shared tail both resolvers end in now follows completed redirects to a fixed point with a visited set (a cycle stops deterministically; the result is still subject to the fence targets, the ownership entry and the under-lock recheck, so a corrupted chain can misroute exactly as before but cannot admit an unfenced session). Table test: two-hop chain and a cycle, each asserting both hops exist in the store before asserting the resolution.
- **Medium, workspace/REST capability derives `Fork` from `pastExists`** — refuted, no change. There is no live REST fork endpoint (the only one is a retired slot in the mutating-route fuzz table that now exercises unknown-route handling), the web frontend never reads `hubapi.SessionCapabilities.Fork`, and the TUI's fork capability comes from the AppWire `ForkFromTurn`, which is this projection's own output. The live workspace half (`liveWorkspaceSnapshot` → `hubCapabilitiesFromAppwire`) writes the daemon's raw `ForkFromTurn` into the same field, equally unread. The dead field is filed as #1162.
- **Medium, `ownsFork` computed for every relayed notification** — fixed in `0f0161c81`: the projection runs only for `NotifyThreadStatusChanged` frames, the stampers' own precondition; a counting test asserts a status frame projects once and item frames do not project at all.
- **Low, liveness resolved twice per listed thread** — the within-one-projection half fixed in `897cd4c25`, `330c5dc80` and `f48480b03`: the thread's daemon is resolved once per `applyHubForkCapability` and handed to both consumers, and a thread already fenced statically never reaches the roster at all (table test: forkable → one roster scan, already-fenced → zero). The once-per-sweep memoisation is the restructure tracked in #1146.
- **Low, `ActiveFlags` fence has no producer** — documented in `739519696`: one hub-side constant names the `resumeRequired` flag with a comment stating that nothing emits it yet and that `ResumeLocks.RecoveryState` is the recovery signal with a producer. The producer decision is #1159. Nothing added to `appwire`.

Gates: gofmt, vet, `go test -count=1 ./cmd/evener-hub/...`, `-race` on the fork fence suite, pinned golangci-lint 0 issues. No existing test changed.

PR #1107, opened 2026-09-10
feat(llm): default API request logging off, configurable per session and hub

obra:
## Why

Every session durably recorded every provider request and response body to `<stateDir>/sessions/<id>.api.jsonl`, with an fsync per append. That fills users' disks. API-request logging is now opt-in.

## What

- **New ownership logger mode** (`llm.NewSessionOwnershipAPILogger`): keeps the per-session flock (the `ReserveSession`/`ReleaseSession` session-ownership mechanism) and stays attached as client middleware — sink binding is what stamps `response_protocol` on turns — but discards appends, so the file it creates stays empty. Attempt identity, turn protocol provenance, and ownership all behave identically.
- **`--api-log on|off`** on `evener run` and `evener serve` (default off; invalid values fail loudly, parsed like `--sandbox-net`).
- **`api_log` launch option** (global/repo/project/per-launch, Debug logging group), emitted to the daemon as `--api-log on|off` in both directions so a launch layer can override a global default either way. Frontend renders it schema-driven; wire types regenerated.
- **`api_log` in hub.toml**: the hub-wide floor for spawned and resumed daemons; launch layers that set the option explicitly always win.
- No `EVENER_API_LOG` env var — flag, launch layers, and hub.toml already cover the surfaces.
- Docs: hub config + launch config sections in `docs/evener-hub.md`, regenerated appwire protocol docs, and a note in the doctoring skill that an empty/missing `.api.jsonl` is expected by default (recording is opt-in), not corruption.

## Tradeoff

With logging off, `doctor apilog` and transcript `source=api_log` see empty logs for new sessions; provider-forensics workflows need the session relaunched with logging enabled.

## Verification

Ran in the `api-log-default-off` worktree against this tree: `make lint`, `make vet`, `make build`, `ROOT_FULL=1 WEB=0 make test` (every module), `make test-web` (typecheck, Vitest, Biome). New tests red-proofed: the discard path, the hub spawner injection, and the middleware-binding guarantee that preserves `response_protocol`.


obra:
Addressed the roborev medium finding in 35d49b366.

The hub.toml `api_log` floor now reaches the preview surfaces too: the launch controller carries the hub default (threaded from `hub.Config` through `WebConfig.APILogDefault`) and applies the same `applyHubAPILogDefault` floor the spawner applies — after layer resolution, before the builtin floor, with provenance `"hub"`. `evener/launch/resolve`, `SetLayer`, and `TrustRepo` now report `apiLog: true` (provenance `hub`) when the hub enables it and no launch layer set the option, so the resolved view matches what `Spawn` actually launches. An explicit launch layer still wins in both directions.

Covered by `TestLaunchResolveAppliesHubAPILogDefault` (red-proofed: fails without the floor application), plus the existing spawn-boundary test. Staticcheck also caught that the first cut of the refactor dropped `Resolve`'s error check; that's restored.

obra:
All three findings addressed in e5835919a.

- **TrustRepo post-trust preview (medium):** the second resolve after saving the trust decision now goes through `resolveWithHubDefaults`, matching `Resolve`/`SetLayer`. Pinned by an assertion on the post-trust effective view (red-proofed: fails against the raw resolver).
- **TUI renderer (medium):** `launch_schema.go` renders `api_log` with the same optional-boolean formatting as `verbose`, and `applyEdit` accepts/edits it via `parseOptionalBool`, so the option no longer shows `(unsupported)` and the editor no longer rejects it. Covered in the all-fields table, the defaults table, the edit round-trip, and the applyEdit fuzz field list.
- **Nil provenance (low):** `applyHubAPILogDefault` allocates the provenance map when nil and records `"hub"`; the unit test now starts from a nil map and asserts the provenance lands.

Gates: make vet, make lint, and the full cmd/evener-hub + cmd/evener-tui suites pass against this commit.

obra:
Addressing the review on \`2a6e81f\` (roborev combined review):

**Medium — resume path ignores launch layers' \`api_log\`: fixed in a4dbe183c.** \`resumeRequestForConfig\` now resolves the session's launch layers for its working dir and carries \`api_log\` into the \`ResumeRequest\`, exactly as the finding prescribes: the persisted session meta stays authoritative for model and provider, and the hub floor (\`applyHubAPILogDefault\` in \`HubSpawner.Resume\`) only fills the value when every layer left it unset. A resolve failure is swallowed so a broken \`launch.toml\` cannot block resume; the unset value then falls through to the hub floor and the daemon's own default. Covered by \`TestResumeRequestForConfigCarriesLaunchAPILog\` (layer true/false carried, unset stays nil, broken TOML resumes unset — red-proofed against the old behavior) and \`TestHubSpawnerResumeAPILog\` (full precedence matrix at the argv boundary with a fake daemon).

**Low — TUI fallback \`layerRows\` omits \`api_log\`: pushing back, no change in this PR.** The hardcoded fallback is a degraded-mode core list, not an attempt to mirror the schema: it omits the entire Debug Logging group (\`verbose\`, \`trace_file\`, \`cpu_profile\`, \`export_atif_path\`) as well as \`sandbox\`, \`sandbox_net\`, \`app_replay_size\`, and the system-prompt fields. Adding \`api_log\` alone would make it the one Debug-group row in a list that deliberately excludes its sibling \`verbose\`. If the fallback should surface Debug-group options when the schema RPC is unavailable, that should be a separate change covering the whole group, not a single-field carve-out on a feature PR.

obra:
Addressing the review on \`a4dbe18\` (roborev combined review):

**Medium — hub false floor defers to the child binary's default: fixed in 522ad13d.** The floor now pins both directions: with the hub's \`api_log\` false and launch layers unset, the hub passes an explicit \`--api-log off\` instead of omitting the flag, so the hub-wide policy never depends on the spawned binary's own default. This went with your first suggested fix rather than a capability/version handshake — extending \`launch-check\` beyond the appwire protocol is a bigger contract change than this PR should carry, and the true floor already pinned unconditionally (\`--api-log on\`), so the false direction deferring was an asymmetry in the original design, not a deliberate choice.

One tradeoff to note explicitly: a pre-change \`evener\` binary paired with a new hub now fails its launch loudly on the unknown flag (visible in the daemon spawn log; the remedy is upgrading the child binary) instead of silently recording full request/response bodies. Loud failure over silent privacy drift is the right direction for this control.

Launch layers still win in both directions (unchanged), and the \`evener/launch/resolve\` preview now reports the pinned value with provenance \`hub\` in both directions (the off case previously showed the builtin default). Tests updated across the floor unit test, the spawn and resume argv matrices, and the preview provenance; docs updated to state the explicit both-direction pass.

obra:
Addressing the review on \`522ad13\` (roborev combined review):

**Medium — unknown-flag death surfaces as a misleading spawn timeout: fixed in ca985ff2f.** \`launch-check\` now advertises the serve flags the binary accepts (\`launch_flags: ["api-log"]\`), and \`validateEvenerLaunchContract\` rejects a child that does not advertise \`api-log\` with "evener launch-check did not advertise the --api-log flag: upgrade the evener binary the hub spawns" — before any spawn or resume attempt, so the incompatibility is reported immediately and names the remedy instead of dying at flag-parse and surfacing as a rendezvous timeout. Old binaries omit the field entirely, which is itself the signal; the JSON addition is backward-compatible for older hubs (extra fields are ignored).

Red-proofed both directions: without the producer field, the launchcheck output test fails; without the consumer gate, the new rejection test fails with exactly the misleading failure you described ("process exited before rendezvous: exit status 2"). Spawn-path test fakes updated to advertise the flag; the --models listing path is unaffected (it never passes serve flags).

PR #1109, opened 2026-09-10
fix(serve): publish terminal session boundary on shutdown

obra:
Shutdown must persist and publish the terminal session boundary before closing the event tee or replacing the current session identity. Keep the shutdown owner distinct from ordinary clear, serialize the final identity transition, and bound waits for shutdown-owned session drains.

Rendezvous removal is terminal: a clear handler that finishes after daemon teardown cannot recreate a discovery entry for an exiting daemon. Drain completion removes the per-session lookup entry; bridge startup is coordinated with teardown while permitting synchronous callbacks for already-closed sessions.

Validation: deterministic real-session regressions cover shutdown during cancellation, clear identity ordering while the old SESSION_END is held, bounded expiry, and rejected rendezvous resurrection. Focused normal/race tests, vet and cmd/evener golangci-lint passed at ccfa53feff30baeb299ede82dbcacbf149993c5d. The earlier d77 lint failure is corrected. Current head `ad45099a958232e400058c008c6a8f68867e4670` integrates main; a fresh complete canonical gate and current-head CI/RoboRev are required.

Keep draft until those current-head gates pass. The intermittent stable-active-ID stop regression also observed in #1098 remains a separate unproven CI failure mechanism; do not claim a rerun fixes it. Full continuation: #1116.

The broad gate at6fbe exposed a dedicated bridge test whose real rendezvous write now correctly failed before reaching the bridge boundary. Its fixture now isolates the bridge write seam; normal/race bridge tests pass while separate real-filesystem tests preserve terminal discovery fencing. A subsequent actual RoboRev finding about failed-removal retry is also corrected: writes remain blocked but failed cleanup remains retryable. The strengthened filesystem regression fails on the old implementation and passes with race checking. Current published head `ad45099a958232e400058c008c6a8f68867e4670`; the final full cmd/evener package verification passed (34.851s plus all internal subpackages). Fresh CI/RoboRev and the complete canonical rerun remain required.


obra:
Current candidate: `22f3602f8f1c4fc8171c67b0d5f938c16221cb87`.

Shutdown publishes its close budget to authoritative event senders, including a sender already blocked behind a full buffer. The terminal SESSION_END follows the same bounded path, with ordinary healthy bridge delivery preserved. Shutdown-owned clear and final teardown share a broadcast expiry, so the one-shot timer is consumed once. Cleanup Notification hooks receive the close context and are skipped once it has expired.

A deterministic saturated-buffer regression waits until an ordinary sender has entered the blocked-send path before beginning close. A real bridge regression keeps the bridge wedged through both waits and supplies a single timer value without closing the timer channel. It passes with the shared broadcast; an isolated Go overlay using the old serve implementation times out in the final drain wait. Focused serve/bridge race tests pass (5.807s), as do focused event/hook normal and race tests. Earlier full cmd/evener normal passes (47.349s); the final head needs fresh CI and actual RoboRev qualification.

Preserved the earlier retryable rendezvous removal and identity-swap/drain-order fixes. This PR remains draft while current-head CI and review run.

obra:
Current candidate: `a0ddad58b`.

CI: the deadline-audit failure is fixed in `de1ad45ea` — the two waits in `session_lossless_events_test.go` now use the audit's sanctioned TRIPWIRE form and already await their real completion signals.

Medium (job manager misses terminal SESSION_END when `delivered` is false): no behaviour change. `closeRuntimeState` (`agent/session_lifecycle.go:492`) runs before the terminal emit (`:575`) and detaches every watch config, so `jobManager.onSessionEvent` would iterate an empty registry there; on the `appendTeardown`-failure path the configs remain but `jm.closing` blocks delivery (`agent/job_watch.go:3918`, `:5499`). Removing the gate would only bump watch counters on that failure path without delivering anything. `c5d8c6d4a` adds `TestSession_TerminalBoundaryReachesNoJobWatch`, which pins the close ordering that makes the gate irrelevant.

Medium (`Close` then `CloseForShutdown` publishes no boundary): real; reproduced. The shutdown path cannot recover the boundary after an ordinary close because the first close already closed the event stream, and it is unreachable through `serve.go` today (`identityTransitionMu` orders the shutdown claim so `closeSupersededSession(oldSess, true)` only runs after the shutdown close has returned and published). `c5d8c6d4a` corrects the false doc comment on `closeSupersededSession` and adds `TestSession_DoubleCloseCannotRecoverATerminalBoundary`.

Medium (rendezvous removal): `c94b0e681` retries `Remove` up to three times with a bounded pause during shutdown and logs an exhausted budget through the serve log; covered by a blocked-removal regression and an exhausted-budget test.

Low (`Register` holds `mu` across the write): declined. `Register` has a single call site on the startup path before the shutdown goroutine and `serveHTTP` exist, so nothing can stall behind it, and moving the write ahead of the lock reopens the artifact-resurrection window `Remove`'s `removed` flag closes (`TestRegistrationRemoveClearsEntry`).

Low (`sendEvent` exclusive lock): `a0ddad58b` reads the close signal under a shared lock with an exclusive upgrade only for the first mint.

obra:
Current candidate: `a6e0fca53`.

Medium (in-flight notification hooks keep a background context at shutdown): real on both halves; fixed in `0245af57f`. A hook already running held `context.Background()` because `fireNotificationHook` read `closeCtx` only at launch, and hooks run synchronously on the input loop that the serve shutdown path waits for before `CloseForShutdown` (the only publisher of `closeCtx`). The hook context now derives from the session's own lifetime context, which `close()` cancels as its second step, with the close-budget precedence unchanged; and `serve` now passes its shutdown context as `LifetimeContext` on every session it builds (`run` already did; `serve` was the outlier), inherited by `thread/clear` and the `--resume` restore path. Regression: a hook that blocks until cancelled is running when shutdown starts, and shutdown completes within the close budget. `1ab21ef58` pins the new decline (a hook fired after the lifetime is over never starts) and the `thread/clear` inheritance, and updates the `LifetimeContext` doc comments to the current contract. Behaviour change worth naming: a notification emitted after SIGTERM (or after Ctrl-C / `--timeout` expiry in `run`) no longer fires a Notification hook, the same rule this PR applies to an expired close budget. Main merged in for the strict base.

obra:
Current candidate: `efd034768`.

Round 3 (`a6e0fca` High: authoritative sends do not observe shutdown): real on all four links (the park holds `eventsMu.RLock`, ordinary sends use a background context until `Close` publishes the budget, and the serve shutdown path waits for `inputLoopDone` before the only close that would publish it). Fixed in `c8179b38d` and `9e555b1dd`: a parked authoritative send now also observes the session lifetime context (the one `serve` supplies as its shutdown context), gated so a budget published before the park keeps precedence; and when the lifetime ends while a close has already published a budget under the park, the send re-parks on that budget instead of returning, so round 1's bounded delivery window is preserved for ordinary closes (clear-superseded sessions, `run`, library) while daemon shutdown, where no budget can exist until the loop returns, releases immediately. Regressions: a unit case with a production-shaped session (park, publish budget, cancel lifetime, drain one → delivered) and an end-to-end serve case with a real daemon, a full 256-event buffer and a wedged bridge that hung 30 s before the fix and completes in under a second after. Main merged in for the strict base.

Issue #1116, opened 2026-09-10
Native iPhone v1: full remaining work and laptop repair handoff


obra:
# Native iPhone v1 repair handoff and full remaining work

## Active work update — 10 September 2026, final review sprint

This section supersedes the older status tables below. The archived GitHub snapshot records exact heads, every check and actual RoboRev comment; a green status icon alone does not establish a clean review.

One candidate completed a fully clean two-reviewer RoboRev panel during this sprint: #1091 at `f55b1b3`. It also had all CI green. Main had advanced, and branch protection requires an up-to-date base, so the candidate was refreshed and must be requalified before merging. No mobile candidate was merged during this final sprint.

| Candidate | Latest published head | Verified progress and remaining gate |
|---|---|---|
| [#1091](https://github.com/prime-radiant-inc/evener/pull/1091) | `320b90c1e04e36a69bf07f5e6ea4feb53e3f1be4` | Both reviewers and all CI passed at f55b1b3. Main advanced with #1090/#1084; strict up-to-date protection required integrating422d9ea8. Refreshed current head is published and requires fresh qualification. Land before native. |
| [#1096](https://github.com/prime-radiant-inc/evener/pull/1096) | `0a1330f084a34e1681867d36897028a768ff4d1c` | Fixed retained activity crossing thread replacement; new real ActivityList regression and native TypeScript pass. Full native gate now passes721 native tests,688 shared tests and TypeScript. The9 initial keybinding failures came from unapplied checked-in dependency patches, not Node version: npm run postinstall applied the native tinykeys/regexpu patches and the unchanged tests passed. Run postinstall after installing dependencies; do not delete the v-flag assertion. Current CI/actual review remain required. Earlier canonical165d passed separately. |
| [#1098](https://github.com/prime-radiant-inc/evener/pull/1098) | `9f685d7f37c42fbc8a75690c019299989b183f37` | Queue rollback fix remains published with full canonical22bf and current-head CI green. One current panel member is clean; the second is still reviewing. Prior intermittent stable-active-ID CI failure remains unproven, despite successful rerun and20 local repeats. |
[...]

PR #1129, opened 2026-09-10
Test late compaction-owner paging fragments

obra:
Exercise a late compaction record owned by an earlier logical turn after the next user input has already appeared. The existing behavioral test compares full projection, warm/cold indexed grouping and one-item paging so non-contiguous owner fragments retain the same identities across readers.

This is a small additional regression on top of #1100. It did not reveal a new production defect; the expected fragmented paging representation agrees across reader paths. Keep draft while #1100 and its environment prerequisite finish qualification. The test will be incorporated into the compaction branch after the current unchanged-head canonical gate finishes.

Repair and remaining work: #1116.


obra:
The additional late-owner paging regression is incorporated into #1100 at 0250b5e31, along with the current environment rollback prerequisite. This draft is superseded by that combined candidate; #1100 still requires fresh qualification.

PR #1131, opened 2026-09-10
fix(webui): retain new-session drafts per project

obra:
## Summary

- Keep each project's new-session prompt, images, and launch settings in memory across in-app navigation and composer remounts. Full-page reload persistence is intentionally out of scope.
- Retain drafts on failed creation. Clear successfully submitted content only in its originating draft, preserving newer edits and other projects' drafts.
- Scope pending image processing, directory preflight, and advanced-option validation to their draft. Preserve folder-picker choices when remounting on an unchanged prefill URL.

## Verification

- Regression tests demonstrated draft loss before the fix. Independent review reproduced two additional lifecycle defects; permanent regressions and the unchanged independent probes now pass.
- Focused composer, attachment, advanced-options, and lazy-dialog suites: **208 tests passed**, with no unexpected warnings.
- `make test-web`: typecheck, unit tests, and Biome passed.
- `TMPDIR=/tmp make test-web-browser`: all five Chrome guards passed.
- `make build`, `make lint`, `make vet`, and `ROOT_FULL=1 WEB=0 make test` passed; final secret scan and diff checks passed.

Browser verification used a short temporary root because this host's Chrome exits with SIGTRAP under a long nested `TMPDIR`. The same failure reproduces on `about:blank`; the unchanged guards pass with `TMPDIR=/tmp`.


obra:
Addressed both findings in Combined Review (`587b533`):

- Project-owned model notices, inherited defaults, and advanced-option feedback no longer leak across draft navigation. Late config-preview responses are ignored after switching projects; disclosure and picker controls remain mounted.
- URL directory prefill again trims padding and treats whitespace-only values as absent, preserving draft and launch identity.

Independent review also reproduced a delayed model-sweep ownership bug. The sweep now snapshots and checks the originating draft, preserving newer model selections and isolating other projects. Its unchanged independent reproducer passes.

Verification, all exit 0:
- 223 focused tests across Spawn, AdvancedOptions, SpawnChunk, and attachment lifecycle suites; no test warnings.
- Biome on the four touched frontend files and TypeScript typecheck.
- `make merge-approval-gate` (lint, build, full Go/module and frontend tests).
- `make vet`.
- `TMPDIR=/tmp make test-web-browser`, all five browser guards. The short temporary path avoids this host's independently reproduced Chrome startup failure.

Required GitHub approval remains separate from the independent code review above. No merge performed.

obra:
Addressed the model-cleanup findings from the review of `03e04c2` in `648ce57ff04b0da2cdd66e101c07577c23dff117`:

- Validate each activated draft against the global Evener catalog, including in-memory drafts whose persisted model was already swept.
- Keep discard notices with their drafts. Choosing a model in another project no longer erases the original notice; notices survive pane remounts.
- Use explicit `model/list {harness: "evener"}` without `cwd` for the existing all-project/global-default sweep. Scoped picker catalogs cannot drive that cleanup. Preserve valid/unknown models, unrelated settings, and external-harness live selections.
- Retire obsolete global requests on `evener/auth/updated`, before the debounced instance refresh completes. Independent review reproduced this additional race; the permanent regression checks the live selection plus project and global saved defaults while that refresh is paused.

Scope qualification: the current built-in Evener backend uses shared hub provider configuration across directories, so a directory-only catalog difference was not established. External harness catalogs are independently scoped; the explicit global request makes the cleanup authority unambiguous.

Verification on the committed source:
- TDD regressions failed before their fixes.
- 255 focused tests passed across Spawn, AdvancedOptions, SpawnChunk, spawnDefaults, and attachments; no unexpected warnings.
- Independent code review approved; its unchanged 11-test probe suite passed, including the auth-refresh reproducer.
- Touched-file Biome and TypeScript checks passed.
- `make merge-approval-gate && make vet && TMPDIR=/tmp make test-web-browser` exited 0: lint, build, full repository/frontend tests, vet, and all five browser guards passed. The shorter browser temporary path avoids the previously diagnosed Chrome startup failure on this host.

No merge performed. The new commit still needs its GitHub checks and required eligible-reviewer approval.

obra:
## CI failure and RoboRev follow-up

Fixed in `132d21bfae4872545bacc4f0ca76aad1eba9f230`.

### CI initialization race

The failing web test captured its `model/list` count after the new global cleanup request but before the initial directory-scoped request. The initial request then arrived during picker interaction, producing `expected 1, received 2` without loading the unconfirmed directory.

The unchanged test failed locally on `648ce57` and passed on CI's actual main parent, `ff0ec647`. A controlled request-order probe reproduced the same result. The repair waits for the initial scoped request before taking the existing baseline. Both exact-count assertions and the confirmed-directory assertion remain unchanged.

### Plugin preview ownership

Reproduced through both directory-picker and URL navigation: the first render for project B could expose project A's ready preview. Reconciliation then cached A-derived issues for B, blocking Start while B's default preview was pending or after B's explicit-selection preview failed.

`usePluginPreview` now associates results with their client and request identity, revokes stale ready/error authority before consumers render, and retires callbacks when their effect ends. Same-project refreshes retain the loading list; genuine known-invalid selections still block submission. The form remains mounted.

### Model catalog finding

The proposed project-specific built-in Evener catalog was not supported by the implementation. `app_models.go:64–85,156–170` routes the explicit Evener request to the launch model lister. `spawn.go:86–134` uses the same hub provider/credential configuration for cwd and unscoped requests; `spawn.go:775–783` does not change the subprocess directory. `launchcheck.go:113–135` enumerates the shared registry. CI's newer base leaves this model-list path unchanged. The global Evener sweep remains intact; external harness catalogs remain separate.

The plugin review's URL-only-reset premise was also inaccurate: the reset already depends on both `draft` and `prefillRevision`. Real UI controls did not reproduce permanent external-harness blocking. The stale-ready Evener failure above was independently reproduced and repaired.

### Verification

- Targeted CI regression: reproduced red, then passed with the original assertions intact.
- Permanent regression tests: 23 failures before the plugin repair; 308 focused tests across seven suites passed afterward. Touched-file Biome and TypeScript checks passed.
- Independent review: approved with no actionable findings. All 18 frozen healthy acceptance probes passed. The original diagnostic probe was preserved; before implementation, one bug-characterization expectation was corrected using the existing failed-preview contract. That healthy reference failed five cases before the fix and remained unchanged through the passing run.
- Fresh `make merge-approval-gate && make vet && TMPDIR=/tmp make test-web-browser`: exit 0. Lint, build, full root/module/frontend tests, vet, and all five browser guards passed. The Chrome temporary-directory setting addresses this host's previously verified Unix-socket path limit.

These are local verification results; CI for the pushed revision is reported separately.

obra:
## Latest review findings: fixed in 7218dec3d

The two medium findings on 132d21b (shared completion state mutating another draft's plugin issues and slash menu; last-completion-wins navigation letting a stale finish steal the route) are repaired on this head.

### Behavioral policy (chosen before implementation)

A completed launch opens its session **only while it still owns the active New Session view**. Switching projects or leaving the view permanently revokes that launch's navigation authority; coming back later does not restore it. Every launch still cleans up its own submitted draft snapshot either way.

### What changed

- `Spawn.tsx`: launch-view ownership token captured at submit (before directory preflight / Create & start), revoked on draft switch, route departure, pane unfocus, or unmount; only the owning launch navigates. Submitted-draft cleanup (prompt, attachments, plugin selection, busy) always runs, so a stale launch cannot strand its own draft and cannot touch newer prompts or other drafts. The slash menu now closes because the displayed prompt was actually cleared, not because any form's completion continuation fired. The create-directory dialog only opens over the focused New Session view, so a late preflight cannot pop it over Settings.
- `spawnDrafts.ts`: known plugin-selection issues live on their draft. They survive bare `/new` returns, remounts, and failed preview refreshes, and cannot leak into another draft; a ready preview, removing the invalid plugin, or an unsupported harness still retires them legitimately, and unknown preview errors still fail open.
- 26 new regression tests cover both concurrent completion orders, project-switch and route-away revocation (including A→B→A and /new→settings→/new), pending-preflight departures, menu preservation across drafts and remounts, and issue retention/retirement paths.

### Qualifications on the original findings

- The known-issues erasure was **not uniquely caused by a late completion**: a bare `/new` return independently reset the cache (prefill revision). Both paths are fixed by draft-owned issues; the frozen acceptance probes cover each.
- Backend acceptance of an invalid selection was never demonstrated and is not claimed; the investigation's RPC stub rejects it, and the client blocks submission while a known-invalid issue exists.
- One frozen diagnostic probe was amended after the policy decision: its navigation expectation encoded the old last-completion-wins behavior in a scenario where the user had already switched drafts, which the chosen policy forbids. The amendment (documented with before/after hashes) changes only that expectation and keeps the probe's leave-and-return known-invalid oracle byte-identical.

### Verification

- TDD: 21 new assertions failed on unpatched code for the expected reasons, then passed after the repair; no existing assertion was weakened (182/184 prior top-level test statements byte-preserved; the 2 exceptions are disclosed fixture-setup corrections proven from app-shell source: the shell only ever mounts Spawn at `/new`, while the shared fixture mounted it at `/`).
- Focused frontend suites: 334/334 passing, warning-free; Biome and `tsc --noEmit` clean.
- Independent frozen acceptance probes (real React/Spawn/routing, scripted network boundary only): 13/13 passing, including both concurrent completion orders and the known-invalid return pair.
- Full local gates: `make merge-approval-gate`, `make vet`, and all 5 browser guard suites pass.
- Independent review: approved after re-running the frozen probes and auditing the diff, with zero code findings.

Remaining blockers are unchanged: branch protection requires one eligible approval and the PR is behind main; no merge, base update, or history rewrite was performed.

obra:
## RoboRev findings on 7218dec3d: fixed in 3602d9401

All three findings from the Combined Review of `7218dec` are repaired here.

### Medium: explicit prefill ignored on route re-entry

`spawnDrafts.ts`: the same-URL suppression of `applySpawnURL` was valid only while the route never left `/new`. Scenario reproduced before the fix: mount `/new?dir=A`, switch to a picker-selected draft B (URL unchanged), leave `/new`, return to the same `/new?dir=A` - the pane restored B and dropped the explicit `dir=A` prefill. The prefill marker is now invalidated whenever the route leaves `/new`, so re-entry re-applies the URL's directory (the other draft still persists in the per-project map). A remount with no intervening navigation still keeps the marker, so a redundant re-apply cannot clobber the draft the user is on.

### Medium: query-only navigation did not retire a pending launch

`Spawn.tsx`: launch-view ownership revoked only when the pathname left `/new`, so a query-only navigation (`/new?dir=A` to `/new?dir=A&prompt=updated`) left a pending launch able to navigate away from the view the user had just re-requested. Ownership now revokes on any real URL change (href comparison), including query-only navigation; a popstate carrying no URL change does not revoke, and the launch's own programmatic navigate is unaffected because ownership is evaluated before it. Draft switches, pane unfocus, and unmount revoke as before, and an untouched view still opens its session.

### Low: slash catalog leaked the previous draft's ready state

`useSpawnSlashCatalog.ts`: the hook exposed the previous draft's ready catalog during the first render after a draft switch (its reset happened in a passive effect), so the new draft could show or accept completions from the old directory for one commit. It now mirrors the plugin-preview render guard from `132d21b`: the stored result is keyed by client and request, the old request's ready/error state is never exposed for a render whose inputs have changed, and the same-cwd cached-loading plus disabled ready-empty contracts are derived unchanged.

### Verification

- TDD: 5 new assertions failed on unpatched code for exactly these three findings (3 hook cases, re-entry prefill, query-only revocation), then passed after the repairs; no existing assertion was weakened or removed (all prior tests, including the 26 completion-ownership cases, still pass verbatim).
- Focused suites: 7 files, 340/340 passing, warning-free; independent reviewer reran Spawn and slash-catalog suites: 209/209, exit 0, warning-free.
- Biome and `tsc --noEmit` clean; `git diff --check` clean.
- Full local gates: `make merge-approval-gate`, `make vet`, and all 5 browser guards pass (exit 0).
- Independent review on a separate provider: APPROVE, zero defects.

### CI note

No GitHub Actions run exists for `7218dec3d` or `3602d9401`. The PR's merge ref (`refs/pull/1131/merge`) has not been recomputed since `132d21bf`, when the branch was still mergeable; the branch is now conflicted with main, and GitHub is not triggering `pull_request` workflows for these pushes while it cannot build a new merge commit. CI for this head will appear once the branch is mergeable again (a base update resolving `Spawn.test.tsx` / `SpawnChunk.test.tsx`), which has not been performed here. Required approval is likewise still outstanding.

obra:
## Base update: main merged in, PR unblocked

Merged `origin/main` (`c7e6c0968`) into the branch as `373eff4f8`. The branch is no longer conflicted, and the pull_request CI workflow now triggers again on this head (it had been silently unable to run while GitHub could not build a merge commit for the conflicted PR).

### Conflict resolution

Conflicts were confined to two test files; production sources merged cleanly.

- `cmd/evener-hub/frontend/src/panes/spawn/Spawn.test.tsx` (connect-provider dialog region): took main's version wholesale. Main's Vitest update (#1132) moved the click outside the `act` block; our older in-act `fireEvent.click` workaround existed for the previous toolchain and its premise no longer reproduces under Vitest 4.1.11 (same configuration reruns warning-free).
- `cmd/evener-hub/frontend/src/panes/spawn/SpawnChunk.test.tsx` (dialog-boundary logic-bug test): combined both sides. Kept main's `renderError` naming and its filtered `onCaughtError` (unexpected errors forwarded to the real console handler) and kept our `caughtErrors` recording; the one reference rename (`expectedError` -> `renderError`) keeps the identical identity assertion on the same `Error` object. No assertion from either side was dropped or loosened.

Frontend dependencies were reinstalled with `npm ci` against the merged lockfile (the Vitest bump changes it) before any test run.

### Verification

- Independent verifier compared both resolved files against both parents: main's regions byte-identical, every branch assertion present, per-file `expect(` counts at or above both parents (`Spawn.test.tsx`: merge 665 = branch 665, main 417; `SpawnChunk.test.tsx`: merge 28 = union of branch 15 and main 27). Verdict: APPROVE, no lost intent, no weakened assertion.
- Conflicted suites rerun independently: 201/201 passing, exit 0, warning-free.
- Full local gates on the merged HEAD: all lint sub-gates, web+runtime build, root/module/frontend tests, vet, and all 5 browser guards, exit 0.

### Remaining

Required CI is now running on this head, and the required eligible approval is still outstanding; no merge has been performed.

obra:
## Combined Review (`373eff4`) findings: repaired in 946dd4391

Per-finding outcome from the review, each claim RED-tested against the code before any fix.

### Reproduced and fixed

- **`useAttachments.ts` marker reuse** — clearing a submitted list reset marker numbering without clearing the retired-marker set, so a later attachment could reuse marker 1 and have its successful encode discarded as "already removed", leaving it pending forever. `clearSubmitted` now clears the retired set and advances the generation when the list empties. Regression test runs the full stage/remove/clear/re-stage/encode-success path.
- **`Spawn.tsx` null `model/list` payload** — an empty catalog serializes as `data: null`, and both the defaults sweep and draft validation iterated it directly. Honest note on the evidence: the new test's assertion passed before the fix; the RED evidence is the unhandled `TypeError: models is not iterable` (spawnDefaults.ts:161, reached from Spawn.tsx:835/856). Both call sites now normalize with `r.data ?? []` (Spawn.tsx:842/863).
- **`AdvancedOptions.tsx` ownership key** — async ownership was keyed on the `readValues` callback identity, which the parent recreates on every remount, so a same-draft remount dropped valid path-validation errors and resolved-config results while still writing the `invalid` flag. Ownership is now keyed on the durable draft (`draftId`, defaulting to the reader callback for callers without a durable draft), carried in a wrapper object so a function owner is never mistaken for a lazy initializer.
- **`AdvancedOptions.tsx` resolved preview** — `showResolved` now resolves from live draft values (`currentValues()`) instead of the render snapshot, so an edit merged by a just-settled validation is included.
- **`useSpawnSlashCatalog.ts` client identity** — cached reuse checked only `cwd`, so a reconnect could seed the loading state with the previous client's catalog. `lastResponse` now stores the client and all three reuse guards require a match, matching `usePluginPreview`.
- **`Spawn.tsx` branch readout** — `branch` was component-local while the form persists across draft switches, so another project's branch could display until its HEAD request finished, or indefinitely after failure. It is now scoped to its `cwd`.

### Routing (`routing.ts`)

The literal symptom as described did not reproduce: `navigate("/new?dir=B")` from `/new` already compared unequal and navigated. The underlying defect did reproduce: because identity was the pathname alone, a target whose pathname matches while the query differs (for example bare `/new` from `/new?dir=...`) returned early and dropped the navigation. `navigate` now compares path, query, and hash.

### Not reproduced (`Spawn.tsx` global catalog sweep)

Independently re-examined from primary Go source rather than accepted on the earlier rebuttal. The cwd-scoped and no-cwd Evener catalog enumerations differ only by `EVENER_STATE_DIR`, which the launch-check enumeration does not honor (registry state root ignores it; `providers.toml` and credential paths are identical), and no per-project launch layer reaches the catalog path (`Resolved{}` on both sides). The premise that a model can be project-valid but absent from the global catalog is therefore not supported; reported as unreproduced rather than fixed to avoid changing correct behavior. This is static source analysis, not a live Go integration run.

### Verification

- Independent review: APPROVE, zero blocking findings. All seven RED reproductions verified at the pre-fix production state (7 failing tests, plus the two unhandled TypeErrors noted above), `expect()` counts strictly increased in every touched test file (spawn 665->672, routing 40->48, advanced-options 81->83, attachments 62->66, slash-catalog 30->32), no assertion weakened.
- Independent reruns: focused 8 files 381/381 passing, exit 0, warning-free; shell suites 15 files 717/717.
- Full local gates on the batch head: lint, build, root/module/frontend tests, vet, and all 5 browser guards, exit 0.

### Disclosed behavior change

The routing fix means a bare-path navigation now actually rewrites a queried URL. One mobile-only consequence was probe-proven during review: `StackHost`'s URL sync strips a `?dir=` query on mount where it previously left it. Prefill is unaffected (the prefill is applied before normalization), no pinned contract is broken, and all 717 shell/mobile consumer tests pass. Flagged here rather than silently absorbed.

CI is running on this head. Required approval remains outstanding.

obra:
## Combined Review (`946dd43`) finding: repaired

**Finding (Medium):** path-validation errors lived only in component-local state while the async callback persisted `invalid: true` on the draft. A validation that settled while its draft was inactive (or across a pane remount) left the value silently excluded from `launchOverrides` with no message when the draft was revisited.

**Fix:** validation messages are now draft-owned. `DraftFields` gains an `advancedErrors` map initialized per draft, `Spawn` owns it with the same draft-field mechanism as the `invalid` flag it explains, and `AdvancedOptions` takes optional `errors`/`onErrorsChange` props (falling back to component-local state for callers without a durable draft). The async validation callback writes to the originating draft even while it is inactive, so returning to that draft shows why the field is excluded. A genuine draft change still clears transient local feedback (resolved/config results), and the validator remains fail-open.

**Regression evidence:** two new tests failed before the fix with the missing message — "a path-validation error that settles while its draft is inactive appears when the draft is revisited" and "a remounted draft still shows the path-validation error stored with its invalid field" — and pass after it. The first also asserts the field stays excluded from `advancedOverrides` and that `thread/start` carries no `launchOverrides`, so the exclusion semantics are unchanged.

**Independent verification:** the reviewer reproduced the RED state itself (reverted the three implementation files in a scratch copy: 2 failed, 1 sibling passed; re-applied: 3 passed), confirmed no assertion was weakened (`expect()` counts 683 vs 672 in `Spawn.test.tsx`, unchanged in `AdvancedOptions.test.tsx`), and re-ran the focused set: 8 files / 383 passing plus the third consumer of the changed component (14/14), exit 0, warning-free. No blocking findings. One non-defect observation: cleared fields leave an empty-string entry in the map rather than deleting the key — invisible to the UI and bounded by the path-kind field count.

Full local gates (lint, build, root/module/frontend tests, vet, 5 browser guards) pass on this head. Required approval remains outstanding; the branch is also behind main, so branch protection will want it up to date before merge.

PR #1137, opened 2026-09-11
fix(web): hold the transcript at the bottom when a late measurement grows it

obra:
## Symptom

`transcriptscrollguard`'s first case fails intermittently as a real case
failure (exit 1), making the required `web` CI job red at random:

```
FAIL  web-transcriptscrollguard (exit 1)
transcriptscrollguard FAIL: pill is visible at mount (text: "latest") - the session should open at the bottom
```

Sighted by the #1135 lane on unmodified `main`; this branch is a fresh branch
off current `main`.

## Reproduction rate

| | before | after |
|---|---|---|
| single guard, CI-shaped isolation | **8 failures / 64 runs (12.5%)** | **0 / 34 consecutive** |
| single guard, every core saturated | not measured | **0 / 12** |
| full `make test-web-browser` | intermittent | passes, all five guards |

Run alone the way `scripts/web/test-web-browser.sh` runs it:

```
cd cmd/evener-hub/frontend
HOME=$d/home TMPDIR=$d/tmp XDG_CONFIG_HOME=$d/xdg-config XDG_CACHE_HOME=$d/xdg-cache \
  XDG_STATE_HOME=$d/xdg-state NODE_DISABLE_COMPILE_CACHE=1 node scripts/transcriptscrollguard/run.mjs
```

## Cause

Not the settle detector, and not fixture nondeterminism. Recording the hook's
own scroll measurements (no added layout reads) gave a byte-identical sequence
in every failing run, ending:

```
mount:after  st=3330  sh=4032   gap=0
scroll       st=3330  sh=10058  gap=6026
scroll       st=10617 sh=16089  gap=4770
scroll       st=11604 sh=17076  gap=4770
scroll       st=16374 sh=17076  gap=0      <- reconcile loop reaches the true bottom, exits
scroll       st=16432 sh=17221  gap=87     <- late batch: +145px of content, +58px of follow
```

The mount's `scrollToIndex(count-1, {align:"end"})` converges on the true
bottom through the virtualizer's reconcile loop, which settles after **one**
frame whose target stops moving (`virtual-core` 3.17 `reconcileScroll`,
`STABLE_FRAMES = 1`). A measurement batch landing after that frame finds the
loop torn down, and the end-anchor that is supposed to hold the end follows
only part of the growth. 87px is far past the anchor's own 4px threshold, so it
disengages and nothing moves the transcript again — the tracer watched the
geometry hold still for the full 3s afterwards.

So the session opens 87px short of the latest content, permanently, with a
jump-to-latest pill on a transcript the reader has never touched. The pill was
reporting the truth; the landing was the defect, and it violates the mount
contract at `useTranscriptScroll.ts:984` ("Opening a session always lands at
the end", kata cmjb).

The scrollHeight variation across runs (17827 / 17395 / 17221) that the earlier
report flagged is a symptom, not the cause: runs at every one of those heights
both passed and failed.

## What the fix does

`useTranscriptScroll.ts`'s scroll listener now classifies each event before
acting on it. An event that carries **more content**, in the **same scroll
port**, with the offset **never moved backwards**, on a transcript that was
**already at the true bottom**, is the virtualizer correcting its own
estimates — only the reader moves a transcript back from the bottom. It re-pins
to the new true bottom from the geometry in hand, which is the same pin
`jumpToBottom` has landed on since PR #851.

Arrival is still confirmed by the pin's own scroll event, so no bottom state is
set optimistically (the rule at `useTranscriptScroll.ts:829`), and a pin a
further correction leaves short is corrected again by the next one.

## What the fix does not do

- It does **not** widen a tolerance, add a sleep, skip or loosen any assertion.
  `AT_BOTTOM_THRESHOLD_PX` and the guard's 1px bottom tolerance are unchanged,
  and no existing test was weakened or re-pointed.
- It does **not** re-anchor content that grows with **no scroll event at all**.
  The correction rides the scroll event the partial follow produces; the
  late-webfont case `transcriptSettle.ts` documents (21px, `scrollHeight`
  growing at `document.fonts.ready` with `scrollTop` pinned) fires no scroll
  event and is still not re-anchored. Not observed in this failure and not
  speculatively handled.
- It does **not** change behavior for a reader who scrolled back, or when the
  scroll port itself resized — both keep their pill, each pinned by a test.

## Tests

- Two new regression tests plus one discriminator test in
  `useTranscriptScroll.test.ts`. The regression is red before the change
  (`el.scrollTop` 16374, expected 16519) and green after.
- `make test-web`: typecheck, 10032 unit tests, lint — all pass.
- `make test-web-browser`: all five guards pass.
- `node --test scripts/*.test.mjs` (76 pass) and the harness's Vitest suites
  (`browserGuardCdp.test.js`, `transcriptSettle.test.ts`, 16 pass).
- `npx biome check --write` on both touched files: no fixes applied.

Guard geometry also stopped varying run to run (16350px every time), because
the transcript now stays pinned while the rows measure in.

## Round 2 — the gesture veto (`f9985829f`)

Review of round 1 raised a real gap: the classifier decided "this event is a
virtualizer correction" from net geometry alone, which cannot see the reader at all.
A native scroll event can coalesce the reader's own upward delta with a correction that
exceeds it, and that event's geometry is byte-identical to a pure correction — so the
reader was pinned back to the bottom and the pill hidden.

The scroll port now tracks **reader intent** directly (passive `wheel` / `touchmove` /
pointer-drag listeners, plus an explicit marker for the keyboard chords — see round 3), and
a gesture pending in the same frame vetoes the re-pin. The flag clears on the next
animation frame, which is exactly "this frame": a frame's scroll steps run before its
`requestAnimationFrame` callbacks.

**Why track the reader rather than the correction.** The signals for the other direction —
a `ResizeObserver` on the inner element, or `VirtualList`'s `onChange` — are present in the
coalesced frame *too*, since that frame genuinely contains a growth, so they return true in
exactly the case that must be vetoed and discriminate nothing. The one variant that would
(recording the virtualizer's intended target through a custom `scrollToFn`) changes
`VirtualList`'s public option surface for every consumer to settle a transcript-specific
race.

### Why there is no browser-level case for it

Requirement was a guard case driving an upward wheel through CDP while a late measurement
grows the transcript. I built a saturating probe for that condition **before** writing any
case, and it does not reproduce. Park at the true bottom, grow the last rendered row
through the real measurement path (inline `min-height` → the item's `ResizeObserver` →
virtualizer adjust → sizer grows), and fire real CDP wheel ticks as fast as the transport
allows:

| grow px | wheel ΔY | trials | scroll events | with gesture | with growth | round-1 branch reached |
|---|---|---|---|---|---|---|
| 40 | 40 | 360 | 123 ea | 119 ea | — | **0** |
| 400 | 1 | 60 | 62 | 60 | 57 | **0** |
| 400 | 40 | 60 | 60 | 60 | 57 | **0** |
| 40 | 1 | 60 | 60 | 60 | 25 | **0** |
| 1200 | 2 | 460 | 201 ea | 200 ea | 199 ea | **0** |

640 trials, ~645 scroll events, ~639 carrying a wheel in the same frame and ~596 also
carrying real growth. The branch was reached **zero** times. The recorded samples say why:

```
frame 5: wasAtBottom=true  grew=true notDecreased=FALSE gesture=true dst=-1   dsh=400
frame 6: wasAtBottom=false grew=true notDecreased=true  gesture=true dst=+184 dsh=400
```

The growth lands in the wheel's own scroll event, but the offset moves only by the reader's
delta; the virtualizer's anchor adjustment arrives a **separate frame later**, by which time
the reader is already off the bottom. Chrome delivers the input's scroll event first, and
the harness has no lever that makes virtual-core issue its adjustment inside the same
frame's scroll steps.

So a guard case built on this would be **green on the unfixed code too** — the
unfalsifiable-fixture trap `docs/developing-evener/testing.md` names at line 499. The
behaviour is pinned at the unit level instead. (A deterministic *neighbouring* case is
available — wheel ΔY 40 leaves the reader off the bottom reliably — but `notDecreased` is
false there, so round 1 passes it identically; it would be a characterization test wearing a
regression test's label.)

Corollary worth stating plainly: this same evidence says the coalesced misfire is narrower
in practice than its description suggests — it could not be manufactured in 640 adversarial
attempts. The veto is still worth having (it removes the class rather than an instance), but
nobody should read it as a frequently-hit bug.

## Round 3 — gesture-veto edge cases (`8b52fb5ee`, `c95aefe1e`, `55e2b416b`)

Review of round 2 found the veto's own state machine had false-positive paths. **A false
veto is not neutral**: when the branch is skipped, control falls through to the ordinary
path, which records the reader as away from the bottom — so every later correction fails the
at-bottom clause until the reader genuinely returns to the bottom. One false veto reinstates
the mount strand *permanently*. That is why these were worth closing even where they were
filed as lows.

- **Scroll chords never reached the port listener.** `useTranscriptScrollKeys` registers the
  Alt+Arrow chords against the keybindings registry, whose dispatcher is on `window`, and the
  handlers write `scrollTop` directly. `useTranscriptScroll` now exposes a stable
  `markGesture`, called immediately before each of those writes. The two actions that write
  no offset are deliberately unmarked: `scrollTop` jumps to index 0, where the correction is
  already disarmed, and `scrollBottom` is `jumpToBottom` itself.
- **Port keydown inference removed entirely.** Space on a focused tool-row trigger activates
  it; Space in the ask dock's input — the port's own trailing row — types a space. Neither
  scrolls, and both bubbled through the scroll element. The `SCROLL_KEYS` name list is gone.
- **A drag released outside the port never ended.** A mouse pointer gets no implicit capture,
  so a selection drag finishing over the composer never delivered `pointerup` to the port and
  every later `pointermove` marked a gesture. The held button is now the authority.
- **The frame-boundary clear can stall.** `requestAnimationFrame` does not run in a hidden
  tab. The listener now also *consumes* the flag on the event it explains.
- **Veto test fixtures now mirror the browser.** They had left `el.scrollTop` at the
  pre-event offset while the injected `measure` reported the net one, which read as though
  the veto *restores* the reader's position. It does not, and nothing could — the browser has
  already committed the net offset. The DOM and the seam now move together and the assertion
  says what the veto guarantees: the offset the event left is left alone, and the pill is on
  offer.

## Known sibling (not fixed here)

`src/panes/transcript/Transcript.tsx:100-105` is the other `anchorToEnd`
consumer, and it has the identical one-shot mount pattern:

```tsx
if (!didInitialScrollRef.current && turnCount > 0) {
  didInitialScrollRef.current = true;
  listRef.current?.scrollToIndex(turnCount - 1, { align: "end" });
}
```

It drives the same `TranscriptBody` / `anchorToEnd` list (`TranscriptBody.tsx:443`)
with no re-pin, and being the read-only pane it does not use
`useTranscriptScroll` at all — so there is no scroll listener there to carry
this correction. The same under-follow leaves it opening a little short of the
latest turn. With no pill on that surface the symptom is silent rather than
loud, which is why no guard catches it.

Its own comment at `Transcript.tsx:90-97` currently asserts that the end-anchor
"keeps the viewport pinned to the TRUE end while the initial estimate->measured
correction settles" — the measurements in this PR are the counter-example to
that claim.

Left out of this PR deliberately: it is a different surface with no reproduction
of its own yet, and folding it in would widen a flake fix into a second pane's
behavior change. Coordinator ruling: follow-up.


obra:
Round 5 disposition for `45a8f15` (three Medium), now at `b812e7f93`:

- **Pointer handlers did not filter `pointerType`** — real. A finger produces pointer events with `buttons === 1`, so a sideways swipe marked through the pointer path even though the touch path deliberately did not. Fixed: the pointer path starts a drag only for a non-touch pointer with the primary button and ignores touch on move. `endPointerDrag` stays unfiltered on purpose: a stray touch release ending a mouse drag is under-marking, the safe direction (stated at the callback).
- **Wheel marked before confirming the port could move** — real, and the same for touch. Fixed with one predicate shared by both: no marker when the port is already at that limit, or when an element between the event target and the port is an independent vertical scroller with room in that direction (checked exactly, not with the port's 4px at-bottom band). A nested scroller at its own limit still lets the marker fire. The port half reads through the hook's `measure` seam; the walk stops strictly before the port (which is itself `overflow-y: auto`).
- **`lastTouchYRef` never cleared** — fixed with `touchend`/`touchcancel` listeners.

Ten RED-first cases cover the limits, the nested scroller with and without room, the 3px band, both event streams for a horizontal swipe, a right-button drag, and a bare `touchmove` after a completed touch. Existing pointer fixtures now build real `PointerEvent`s (the path reads `pointerType`/`button`); no assertion changed. The one band left unmodelled is named in the comment: an ancestor with a non-`auto` `overscroll-behavior` between target and port, a shape the transcript does not build.

obra:
Round 6 disposition for `b812e7f` (one Medium, one reviewer), now at `a112dfad9` (includes main `bc1ed2aef`):

- **Gesture state persists across session transitions** — real, and the comment that called the retention deliberate was wrong: the one event of the new session that could see a stale marker is the new session's own mount scroll-to-end, which is exactly the landing this PR fixes. A marker exists to veto the event its own gesture caused, and a session switch cannot be that event. Fixed in the per-ref reset block (the single place per-session state is reset): the pending marker, the pointer-drag flag and the touch position are cleared and the pending clearing frame is cancelled with its handle nulled, so the new session's first gesture can schedule a frame of its own. Six RED-first cases: the marker from a wheel, a finger and a mouse drag each surviving a switch; a drag opened and a touch begun in the old session no longer fabricating a marker on the first move in the new one; and the frame cancellation itself, with the handle-null half proven load-bearing by breaking it. `requestAnimationFrame` is stubbed in those tests to model the reachable window (a wheel and a sidebar click in one frame, or a wheel as the tab hides), not to hide a race.

obra:
Round 7 disposition for `a112dfa` (one Medium), now at `87f4944e1` (comment-only change):

- **Nested scroll chaining can misclassify reader gestures** — refuted by measurement rather than by reading, because the finding rests on a browser-behaviour claim: that a wheel delta larger than a nested scroller's remaining room scrolls that element to its limit and applies the residual to the transcript port within the same event. Measured in the browser guards' own headless Chrome 153.0.8010.36 with one real 120px pixel-delta wheel dispatched over the nested scroller via CDP, reading both offsets after a frame, as `(nested room, port scrollTop, nested scrollTop)`:

  | nested room | port | nested |
  |---|---|---|
  | 3px | 0 | 3 |
  | 0px | 120 | 0 |
  | 300px | 0 | 120 |

  With 3px of room the nested scroller went to its limit and the port did not move at all: Chrome latches the wheel to the scroller it targets and chains to the port only on a later event, which the predicate then marks (no room left in the nested scroller). The 0px control proves the wheel was delivered at that point and the port was free to move, so the 3px result is a latch, not a lost event. So "has any room" is the right question, not "has room enough". The comment above the predicate now records the measurement, the Chrome version, and that latching is browser-defined: an engine that split the delta would make this under-mark, the safe direction (one frame's re-pin), never a false veto. The spike script was throwaway and is not in the tree.

obra:
Round 8 disposition for `87f4944` (one Medium, one Low), now at `39e293b5f`:

- **Middle-button autoscroll was never marked** — real. A drag now begins for the primary or the middle button (the secondary button and touch stay excluded), so the moving phase of a native autoscroll vetoes same-frame corrections. Stated residual, ruled rather than closed: a marker fires only from `pointermove` and clears on the next frame, so an autoscroll held still after moving off its anchor marks nothing and corrections re-pin over the reader for as long as it stays still. Letting a held button veto directly would turn every selection drag, and a macOS middle-button hold where the button does nothing, into a standing veto during content growth, which is the permanent false-veto direction this design treats as the harmful one; the stationary residual is the under-marking direction. The comment names both. Where the middle button does nothing at all, this widening marks a drag that moves nothing, the same trade the pointer path already makes for a selection drag.
- **Zoom wheels and claimed wheels marked a gesture** — real. `markWheel` now returns early for `ctrlKey` (browser zoom, and a macOS trackpad pinch) and for `defaultPrevented` (a descendant that already claimed the wheel). A sibling case dispatches the same wheel without `preventDefault` and asserts the veto, so the test proves `defaultPrevented` is the discriminator rather than the nested-scroller walk.

Three RED-first cases with opposite polarities (the Medium was under-marking, the Low over-marking); no existing assertion changed. The `tests` job on the previous head failed on `TestRunServe_StreamErrorPublishesIdleStatus` in `cmd/evener`, which this frontend-only PR cannot touch; it is being root-caused in its own lane.

obra:
Round 9 disposition for `39e293b` (one Medium, one Low, one reviewer), now at `16406565c` (includes main `b10be62a0`):

- **Shift-wheel can carry a non-zero `deltaY` while scrolling horizontally** — real. The port is `overflow-x: clip`, so no horizontal input ever moves it, and some engines deliver Shift+wheel as a vertical delta with `shiftKey` set. `markWheel` now returns early on `shiftKey`: the under-marking direction in any engine that applies it vertically (one frame's re-pin) and the end of a permanent false veto in the ones that apply it horizontally. Round 8's brief had assumed shift-wheel always arrives with `deltaY === 0`; that assumption was not safe across engines.
- **Middle-button autoscroll marked only while the pointer moves** — the residual round 8 disclosed, now closed for the cursor-inside case. The ruling reverses round 8's for the middle button only: a held middle button is an autoscroll in progress, and every scroll event while it is down is treated as a reader gesture, stationary pointer included; the primary button keeps moving-only marking so a selection drag never becomes a standing veto. The state has no clock, so it is bounded by a named set of clears: button up, a move without the middle bit, the pointer leaving the port, the pane switching session, window focus loss and the document going hidden (the last two added after review found that a release the port never sees would otherwise leave the veto standing). Trades stated in the comment: where the middle button does nothing, holding it over the transcript vetoes corrections while held; and because `pointerleave` is one of the clears, an autoscroll continuing with the cursor outside the port goes unmarked from then on.

Eight RED-first cases with mutually falsifying controls (the primary-button hold still re-pins; a `visibilitychange` back to visible does not end the veto); no existing assertion changed.

obra:
Round 10 disposition for `1640656` (one Medium, one reviewer), now at `5f95d4edc` (includes main `f21e1c595`):

- **A pending marker survives the document going hidden** — real, the same shape round 9 closed for the middle-button state: a non-scrolling gesture (a selection drag) sets the marker, the clearing frame does not run while hidden, and the first scroll after return, a measurement correction after content grew, was vetoed. The hidden-document handler now clears the marker and cancels and nulls its clearing frame alongside the middle-button state; `blur` deliberately does not, because a blurred window keeps rendering and the frame boundary remains the marker's bound there. RED-first through the marker rather than the middle state, with a frame-handle control that pins the null half (a second drag after the hidden edge must schedule a clearing frame of its own) and a visible-edge control.

With this round the review also produced a table of every piece of gesture state in the hook and what bounds it. Four of the five are bounded by at least two independent mechanisms. The one reachable gap it found was closed here too: `continuePointerDrag` accepted any held button when continuing a drag while a drag is only ever begun for the primary or middle button, so a primary press, a focus loss with the cursor still inside the port, a release elsewhere and then a secondary-button drag would mark; the continue path now masks `buttons` to the two accepted bits, which also makes begin and continue symmetric. The remaining documented residual is one fabricated touch mark when `touchend` never reaches the port.

PR #1143, opened 2026-09-11
fix(web): center the job-notification card's trailing controls and show its disclosure chevron

obra:
## What

Two defects on job-completed notification cards in the transcript, plus one gate-noise fix discovered while verifying them:

### 1. The Open control rode below the words it opens (fix: `a4fab207f`)

`.secondaryTail` uses `align-items: flex-end` to pin the Open button to the wrapped secondary's last line, but the button's layout slot is only 1em tall (openbutton.module.css hands its 28/44px hit target back via symmetric negative margins). Flex-end seats that slot against the **line box's bottom edge**, not the text's center — measured in real Chrome: the glyph center sat **3.69px below** the last line's center at both desktop and phone widths, with the glyph's box poking below the row.

Fix: `.openTrailing` is now one full line box tall (`height: 1lh; align-items: center`), so the glyph centers on the final text line at every width, font, and hit size — measured **0.09px** off-center after.

### 2. No disclosure chevron (fix: `a4fab207f`)

The card's `<summary>` suppresses the native details marker and never rendered a replacement, so a collapsed card gave zero expand affordance — while every sibling disclosure row (ThinkBlock, ToolRow, even the steering divider) shows one. The summary now renders the shared `Chevron` widget trailing the words it opens, in the established text → Open → chevron order (inside the atomic tail unit when a transcript ref exists, inline after title/secondary otherwise), turning 90° when open.

### 3. Editorial-preview script tests vs fleet worktrees (fix: `a9e027b18`)

The preview's vite config correctly refuses a symlinked shared `node_modules` (its isolation contract pins `fs.allow` to exactly `[frontend]`, which a shared install cannot honor) — but that turned into 2 red + 2 cancelled tests for every agent running `make test-web` in a fleet worktree. The refusal predicate now lives in `scripts/editorial-preview-install.mjs`, shared by config and tests; the three config-loading tests skip with a stated reason on a shared install, the path-escape wrapper test still runs everywhere, and CI's fresh `npm ci` checkout still runs all four (predicate polarity verified hermetically both ways; the config's refusal message is unchanged).

## Evidence

- 4 new unit tests in `NotificationCard.test.tsx` (written red first): chevron presence, rotation, and placement in every structural variant — 41/41 in the file, vitest 442/442 files.
- `notification-open-last-line` layoutguard case extended: measures the chevron too and asserts both trailing slots center on the final text line within 1px. **Both governing rules mutation-tested**: without `height: 1lh` the Open glyph sits 3.7px low and the chevron 2.7px low (both viewports), and the case goes red with exactly those reasons; green with the fix.
- Real component re-measured in Chrome via the overflow harness: chevron present, both glyphs 0.09px from line center.
- Gates: `make test-web` PASS (typecheck, test, lint — fully green in a symlinked fleet worktree, which is the point of change 3); `make test-web-browser` **5/5 guards PASS** (ran with `TMPDIR=/tmp`: Chrome's singleton-socket path exceeds the 104-char Unix limit under this host's long sandbox scratch root — environmental, not test-related).
- Fleet conventions doc updated next to the `ln -sfn` rule that produces the symlink.


obra:
**roborev finding addressed** (`isSharedNodeModules`, medium): both claims validated and fixed in ed44e81b4.

- Reproduced red first: `scripts/editorial-preview-install.test.mjs` failed both cases against the old predicate — the symlinked-parent case (a real `npm ci` install under a symlinked parent, e.g. macOS `/tmp → /private/tmp`, read as "shared" and would have silently skipped the preview's tests where they should run) and the missing-install case (ENOENT instead of a verdict). Two polarity pins (real install → false, `node_modules` symlink → true) ride along.
- Fix: the predicate now `lstat`s `node_modules` itself (never the resolved path) and returns the not-shared verdict on ENOENT, so a missing install stays the caller's own loud failure. The config's refusal message and the tests' skip behavior are unchanged (verified in the symlinked fleet worktree: 3 skip / 1 pass / 0 fail).
- Branch also brought up to date with `main` (merge ee701a713, no conflicts; none of main's three new commits touch these files), and the full affected gates re-run on the merge result: `make test-web` PASS (typecheck, test, lint) and all five browser guards PASS.

obra:
**roborev finding addressed** (chevron/text atomicity, medium): validated and fixed in b7f201f3e.

- Reproduced red first: a bare inline chevron after the secondary stranded on its own line at 8 of 56 swept content widths (200–420px); after the fix, zero strandings across 302 measurements (151 widths × 2 shapes).
- Every summary branch now renders the chevron inside the atomic `.secondaryTail` unit with the words it opens — the secondary's final word, or the title's final word (split the same way) when there is no secondary — mirroring the Open control's existing guarantee. The title split also fixes a second geometry flaw the sweep exposed: a whole-title unit parked the chevron at the unit's right edge, far from the title's last line.
- The requested geometry coverage is wired in: the `notification-open-last-line` layoutguard case now measures all three summary shapes (secondary+Open, secondary alone, title-only) at desktop and phone widths, asserting every trailing slot shares, hugs, and centers on the final text line. Mutation-verified: reverting the atomic unit's `inline-flex` strands every shape and fails the case.
- Branch is up to date with `main` (both advancement merges, no conflicts) and the full affected gates re-ran green on the merged result: `make test-web` PASS (typecheck/test/lint) and all five browser guards PASS.

PR #1123, opened 2026-09-10
WIP: preserve activity continuation regression before repair

obra:
Full continuation checklist: #1116.

Recovery-only draft preserving unpublished mobile work before the development laptop goes for repair. Exact snapshot: `6f103518f4d8d55b10323498fcfbf177570cb01f`, branch `codex/mobile-repair-activity-regression`. Original subject: fix(hub): keep ancestor activity branch state.

Disposition: preserved in new branch codex/mobile-repair-activity-regression at 6f103518f4d8; original dirty worktree untouched. This assessment is provisional; compare with current #1091/#1096 and merged #1039/#1102 before deciding whether anything still needs landing. Do not merge this historical branch wholesale or roll current main backward. The original worktree was left unchanged.

Remaining: identify any unique useful behavior/evidence, extract only that into a focused current-main PR, test the actual contract, and obtain current-head CI and a clean actual RoboRev verdict. If fully superseded, document which PR preserves the behavior and close this archival draft without merging. This snapshot was checked with the repository secret scanner before publishing; no current build, runtime or release qualification is claimed.

Includes the previously uncommitted `agent/jobs_activity_continuation_progress_test.go` copied without changing the original worktree. Run and review the test before using it as evidence.

PR #1179, opened 2026-09-12
fix(agent): make activity size-trim continuations advance

### LINKED ISSUE #1172: agent: activity paging loops forever on a single entry larger than activityMaxEncodedBytes
Found while assessing recovery draft #1123 against main (part of #1116).

**Defect.** In `agent/jobs_activity.go`, when the size trim drops a trailing activity entry whose own JSON encoding already exceeds `activityMaxEncodedBytes` (4MB), `trimActivityTrailingEntry` re-mints `ResumeIndex` pointing back at that same entry. `LoadSessionJobActivityTree` then returns `entries=0 truncated=true` with an identical continuation token on every page, so a client paging the activity tree never advances past the oversized entry and never sees anything after it.

**Reproduced** on origin/main a361254ab with a focused test adapted from #1123's `TestLoadSessionJobActivityTree_SizeContinuationSkipsUnrepresentableEntry` (rewritten against `LoadSessionJobActivityTree(ctx, stateDir, rootID, params)` and `ResumeIndex` semantics): pages 1 and 2 return the same token.

**Fix shape.** When the entry being dropped is alone on the page (or still exceeds the limit after later siblings are dropped), advance `ResumeIndex` past it and set `Branch.Error` naming the skipped entry, instead of re-pointing at it. Do not port #1123's identity-based `After` cursor; main's `ResumeIndex`/epoch fencing is the design to keep. Estimated ~25 lines of fix plus a ~60-line regression test.

**#1123 disposition.** Its frontend half (`activityPanel.ts` ancestor branch state) is superseded by #1091 (`bc1ed2aef`, `activityMerge.ts` + test at `activityMerge.test.ts:533`). This issue is the only unique content; #1123 closes once this lands.

### LINKED ISSUE #1177: agent: activity size trim on page 2+ re-mints a page-relative ResumeIndex, so paging repeats page 2 forever
Found while fixing #1172 (part of #1116); reproduced on origin/main daaaea797 with a scratch probe, not committed.

**Defect.** In `agent/jobs_activity.go`, every size-trim mint in `trimActivityTrailingEntry` uses the index into the *already-resumed* page rather than the session's absolute entry index. A size trim on page 2 or later therefore mints a `ResumeIndex` that points back into the page it just returned.

**Reproduction.** 60 ordinary shell jobs of ~250KB each, no entry anywhere near `activityMaxEncodedBytes`, walked through `LoadSessionJobActivityTree`: page 1 = job_00..15, page 2 = job_16..31, pages 3-10 repeat job_16..31 with an identical continuation token. Same never-advancing loop as #1172, reachable with entirely ordinary jobs.

**Fix shape.** Thread the page's resume offset into `trimActivityTrailingEntry` (`-startDepth` identifies the target session) so the minted `ResumeIndex` is absolute; that changes its signature and its four direct call sites. Regression test: the 60-job walk above must reach job_59 with a strictly advancing token.

Related: #1172 covers the separate case of a single entry larger than a page re-pointing at itself; that fix (index 0 → advance past it with `Branch.Error`) is necessary but does not cover this one.

obra:
Two defects in `trimActivityTrailingEntry` (`agent/jobs_activity.go`) made `LoadSessionJobActivityTree` return the same page and the same continuation token forever.

- **#1177, page-relative mint.** Every size-trim mint used the index into the already-resumed page rather than the session's absolute entry index, so a size trim on page 2 or later pointed back into the page it had just returned. Reachable with entirely ordinary jobs: 60 shell jobs of ~250KB repeat job_16..31 from page 3 on. Fixed in `fc516b124`: a two-field `activityTrimResume{depth, index}` carries the page's starting position from `projectBoundedActivityTree` through the trim, applied only at the one session the page resumed (shallower levels are the filtered ancestor chain; deeper ones render from their own top). Six direct call sites updated; no existing assertion or expected value changed.
- **#1172, oversized single entry.** An entry whose own encoding exceeds `activityMaxEncodedBytes` can never be rendered, and the trim re-pointed the continuation at it. Fixed in `88755c3d6`: when the dropped entry is the only one the session's page still holds (index 0 after tail-first trimming), resume past its absolute position and name it in `Branch.Error`.

Extracted from recovery draft #1123 per #1116; #1123's frontend half is already on main via #1091, and #1123 closes when this lands.

## Tests (RED first, recorded)

- `TestLoadSessionJobActivityTree_SizeTrimResumesAtAbsoluteIndexOnLaterPages`: 60 x 250KB jobs walked with a page cap; token strictly advances, job_00..job_59 seen in order. RED: `page 2 minted page 1's continuation again`.
- `TestTrimActivityTrailingEntry_MintsResumedSessionsOwnIndex`: the nested case, offset applies at the resumed session's depth and nowhere else. RED: `child ResumeIndex = 1, want 6`.
- `TestLoadSessionJobActivityTree_SizeTrimAdvancesPastEntryLargerThanAPage`: a 4.25MB job then an ordinary one; the client reaches the ordinary one and the branch error names the skipped job. RED: `page 2 re-minted page 1's continuation`. Still RED with the offset threading but without the index-0 skip.

## Gates

gofmt, `go vet`, full `agent` module with zero non-ok lines, `-race -count=3` on the trim set (24/24), pinned golangci-lint 0 issues. Draft until current-head CI and RoboRev qualify it.

Closes #1172. Closes #1177.

## Review round 8

**Medium: a live root paging into a closed child could not detect that child's journal being rewritten.** A mint took the delegates generation from the page's own root; a live root has none, so a token naming a closed child carried zeros, and the stale check had to skip validation entirely for live roots to keep the branch readable. The root's clock moves for what the daemon does under it, never for a file rewritten out of band, so nothing was left to catch that rewrite.

Fixed in `e1de98893`: `collectActivitySessionEpochs` returns both generations per session and every mint carries the trimmed session's own, which is what a resume compares the token against. The `root.live == nil &&` exemption is deleted rather than narrowed — a live target still compares zeros to zeros and is fenced by the revision, while a closed child under a live root is fenced by both. Depth truncation now mints the child's own delegates generation too.

- RED `TestJobActivityTree_LiveRootChildContinuationRejectedAfterTheChildIsRewritten`: `resumed into a closed child whose journal was rewritten; want the continuation refused`. GREEN after the change.
- `TestJobActivityTree_LiveRootChildContinuationSurvivesHistoricalEpochs` keeps the readability property: an untouched closed child still resumes under a live root. Its hand-minted token now carries the child's own generations, which is what the production mint produces.
- `TestTrimActivityTrailingEntry_MintsTheTrimmedSessionsOwnDelegatesEpoch` and `TestCollectActivitySessionEpochs` pin the per-session rule directly.

Gates: toolchain gofmt, host + `evenerfuzz` + `GOOS=windows` vet in both modules, agent and appprojector suites with zero non-ok lines, `-race -count=3` over the activity tests (`ok 669.645s`), golangci-lint 0 issues.


obra:
Round 1 disposition (head `7df64fb75`, main `92561dbe3` merged):

- **High, the index-0 skip fired for a nested session whose overage came from a sibling on the same page** — real, and a regression of `88755c3d6`; fixed in `f580b66b2`. The skip now applies only when the session being trimmed is the page's continuation target (`resume.targets(path)`, i.e. `len(path) == resume.depth`); a deeper session mints its own top-relative position, and the page that re-targets it carries none of the ancestors' entries, so the entry is delivered there. An entry that genuinely fits no page reaches index 0 again as the target one page later and is skipped then, so progress is still guaranteed. Tests: the reviewer's fixture (3.9MB root job, delegate child with one 0.4MB job) now asserts the child's job is delivered with no branch error (RED at `fc516b124`); a termination guard pins that a 4.25MB job one hop down is still skipped. `TestTrimActivityTrailingEntry_MintsResumedSessionsOwnIndex` (this PR's own test) asserted root `ResumeIndex = 1` in exactly the buggy case and now asserts `0` with no branch error; its child-side assertion is unchanged.

Gates: gofmt, vet, full `agent` module with zero non-ok lines, `-race -count=3` on the trim set, pinned golangci-lint 0 issues.

obra:
Round 2 disposition (head `77e8533ce`, main `eee90925a` merged):

- **Medium, filtered continuation snapshots drop `JobsEpoch` / `DelegatesEpoch`** — real, and pre-existing on main (`activityFilterSnapshotToDelegate` is byte-identical at `daaaea797`), reachable here because this PR is what mints nested continuations on resumed pages; fixed in `17217d525`. The filter now carries both generations from the base snapshot, so a continuation minted anywhere on a resumed page encodes the generations the next request checks. Same change fixes the depth-truncation mint, which reads the same page-root epoch; the work-budget mint reads each session's own snapshot and was never affected. Test: warmed fold caches, then a same-size restamp of the root's `delegates.jsonl` and the child's `jobs.jsonl`; RED was `page 3: activity continuation is stale`, GREEN walks through.
- **Low, the resumed-page skip is never asserted** — fixed in `2c94d1988`. `TestTrimActivityTrailingEntry_MintsResumedSessionsOwnIndex` decodes the child's continuation after the skip and asserts `ResumeIndex == 6` plus a branch error naming the entry; a new end-to-end case puts the oversized entry on page 2 with a tail job behind it. Both RED against a copy with the offset dropped from the skip branch.

Recorded as #1195: trim mints stamp the page root's `DelegatesEpoch` for every session while the staleness check reads the target's own; they agree only when the child's meta names the root.

Gates: gofmt; vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`); full `agent` module with zero non-ok lines; `-race -count=3` on the trim set; pinned golangci-lint 0 issues.

obra:
Round 3 disposition (head `d791c6d82`, main merged):

- **Medium, `i == 0` does not prove the entry is the cause of the overflow** — confirmed by measurement rather than refuted. A bare root envelope is 222 bytes and a realistic worst case (33-hop chain, 8KB labels, 4KB delegate prose) is ~722KB, 17% of the limit, but three envelope inputs are unbounded: `Label` is the session's `OriginalPrompt` verbatim when it has no name, delegate `Task`/`Mandate`/`Description`, and the per-record `Branch.Error` accumulation; a 4MB pasted prompt reproduces it. Fixed in `cfab29d64`: the trim reports what it dropped, the tree is re-encoded without it, and the skip applies only when that brought the page under the limit; otherwise `markActivityEnvelopeTooLarge` reports the response's own size and withdraws the continuation, which could only reproduce the same page. RED: a small job blamed for an oversized envelope. Capping those inputs at projection time is #1202.
- **Medium, an empty root with a continuation is unreachable in the web UI** — confirmed; fixed on the web side in `f29440fc4`. An empty page that carries a continuation or a branch error gets its own state: the error as the hint plus a control that follows the continuation (RED: the error text not rendered; GREEN: the click issues `evener/jobs/list` with the continuation and the next page renders). The payload already carried what the UI needed, so the Go side is unchanged.

Gates: Go — gofmt, vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`), full `agent` module zero non-ok lines, `-race -count=3` on the trim set, pinned golangci-lint 0 issues. Web — biome, `make test-web`, `make test-web-browser` (five guards).

obra:
Round 4 disposition (head `f913db486`, main `27503c07d` merged):

- **Medium, the skip's diagnostic was appended after the size check, so the page could exceed the cap** — real; fixed in `c884cb282`: the skip (error plus advanced token) is applied, the page recomputed and marshalled, and kept only if it still fits. RED: `trimmed response is 4194389 bytes, 85 over the 4194304-byte limit`. That fix opened a corner — a page within the diagnostic's length of the cap would have restored the branch and looped again — closed in `6d60b6275`: a skip whose named message does not fit tries the fixed `one entry was too large to render and was skipped`, and if even that does not fit advances with no message and a server-side warning naming the entry, so the token always moves past an entry no page can render. RED: `ResumeIndex = 0, want 1`. Residual, recorded: a page with less slack than the advanced token itself costs (~12 bytes) exceeds the cap by that much rather than looping. The real bound is #1202 (envelope input caps).
- **Low, counts inconsistent after the envelope error** — real; fixed in `c53ff9b82` (recompute after `markActivityEnvelopeTooLarge`; the integration test now fails any page that reports an error and calls itself complete).
- **Low, the empty-page no-continuation outcome untested** — added in `85386f18b`; proved non-vacuous by narrowing the branch to the continuation case, which fails it.

Gates: Go — gofmt, vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`), full `agent` module zero non-ok lines, `-race -count=3` on the trim set, pinned golangci-lint 0 issues. Web — biome, `make test-web`.

obra:
Round 5 disposition (head `59f0602e7`, main `99fa1882f` merged):

- **Medium, a live root cannot page into a closed child because its continuation carries the live root's zero `DelegatesEpoch` while the next request reads the child historically** — confirmed; fixed the small way in `4614dc17d`: the generation comparison is skipped for continuations minted from a live root (`root.live == nil &&`). Safe because a live root's pages carry zeros for both generations (`loadLiveActivityBase` reads neither fold cache), so that comparison could only ever reject, never confirm; the remaining fence is `Revision`, and the root's `jobActivityClock` is shared with every session under it and bumps on any job start or finish anywhere in the tree. Given up: an out-of-band rewrite of a child journal mid-pagination, indistinguishable here from the zeros a live mint always carries; that is #1195's full fix. RED: `live root's own continuation into its closed child was rejected: activity continuation is stale`.
- **Low, the silent fallback minted `index+1` without re-measuring** — confirmed; fixed in `de8ec4d15`: the advance is applied and the page marshalled before it is accepted, and the corner where slack is smaller than the token delta is a measured, logged decision (entry, session, bytes, limit). Ruling on the trade the reviewer's framing implies: in that corner an over-cap page that advances beats an in-cap page whose token points at an entry no page can render, because the second is the #1172 loop; the transport does not enforce the cap and #1202 is the real bound.
- **Low, `skipActivityTrimmedEntry` was test-only** — deleted in `f3eb696ed`; the unit test drives the production pair (`mintActivityTrimContinuation` + `explainActivitySkippedEntry`) with no assertion changed.

Gates: gofmt; vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`); full `agent` module zero non-ok lines; `-race -count=3` on the trim and continuation sets; pinned golangci-lint 0 issues.

obra:
Round 6 disposition (head `b785c6162`, main `cb211c5f8` merged):

- **Medium, trimming a delegate strands its child's continuation** — refuted with a fixture that forces exactly that order (`TestLoadSessionJobActivityTree_TrimmingADelegateKeepsItsChildReachable`, `5725fe2e8`, kept as a guard): the owner mints at the dropped entry's own index, so the next page resumes at the delegate and re-renders its subtree from the top with the room the dropped siblings freed; all thirty jobs arrive exactly once with no branch error. The child's continuation is lost only when the delegate itself cannot fit a page, where nothing under it is renderable anyway.
- **Medium, a new delegate shifts the entry list without moving the live root's fence** — confirmed; fixed in `2b4b0b928`: `delegateTreeController.appendLocked` now calls `noteJobTreeShapeChange` for each appended `Created` event, so the session's activity clock, and with it `JobActivityTree.Revision`, moves on delegate creation as it does on job start and finish; `delegate_created` is the store's only structural insert. RED: `resume delivered [dlg_b] against a list that shifted under it`. Recorded: this is the first write from the delegate controller into that clock, and the web treats `Revision` as a freshness signal, so the activity panel now refreshes when a delegate appears — correct, and a visible change beyond pagination.
- **Low, depth-limit placeholders mint `JobsEpoch` 0** — confirmed and deferred to #1195: establishing the real epoch means folding the child's journal, the per-session work the depth limit exists to avoid, so it is not the local read that would fit here.

On the open trade from round 5, restated: in the ~12-byte corner an over-cap page that advances beats an in-cap page whose token points at an unrenderable entry; #1202 is the real bound.

Gates: gofmt; vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`); full `agent` module zero non-ok lines; `-race -count=3` on the trim and continuation sets; pinned golangci-lint 0 issues.

obra:
Round 7 disposition (head `25a17a99b`, main `f39aa2c83` merged): the Medium is real — the delegate-creation bump was memory-only, so a restart restored the pre-delegate revision and accepted a continuation minted against the old list. Fixed in `83ec6cc4d` with the encoding job events already use: the same clock, the same `JobTreeRevision` meta field, the same `maybeAutoSave`, restored via `jobClock.ensureAtLeast`. One difference forced by locking: the bump happens inside `appendLocked` under the delegate tree's lock while `maybeAutoSave` takes `metaSaveMu` then `s.mu`, and existing paths hold `s.mu` before entering the controller, so saving inline would invert an existing lock order; `noteJobTreeShapeChange` bumps and marks, and `CommitStart` — the only path that appends `delegate_created` — persists after releasing the lock. RED: `persisted JobTreeRevision = 0, want at least 1`. Gates: gofmt (toolchain binary), vet (host, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`), full `agent` module zero non-ok lines, `-race -count=3` on the activity and delegate sets, pinned golangci-lint 0 issues.

PR #1145, opened 2026-09-11
ci: retry the golangci-lint download and the root package list

obra:
Two CI-infrastructure failures today, both on PR #1105 heads whose code was
clean. Neither step had a retry or any headroom, so a single transient event
failed a required job.

## Signature 1 — `lint-golangci`, head `9ee4eb4`

https://github.com/prime-radiant-inc/evener/actions/runs/34644540860

```
golangci/golangci-lint info checking GitHub for tag 'v2.13.1'
golangci/golangci-lint info found version: 2.13.1 for v2.13.1/linux/amd64
golangci/golangci-lint err http_download_curl received HTTP status 500
make: *** [make/repo.mk:13: tools-golangci] Error 1
```

The aggregate `static` job failed with it.

**Root cause.** `make/repo.mk:13-15` piped the upstream installer straight into
`sh`. The installer has no retry of its own: `http_download_curl` returns 1 for
any non-200 and its `set -e` aborts there, and its only options are `-b`, `-d`,
`-h`/`-?`, `-x` and the release tag (upstream `install.sh` `usage`/`parse_args`),
so there is nothing to ask it for a second attempt. One 500 from the release
CDN was terminal.

**Second hole in the same recipe.** The pipeline had no `pipefail`, so when the
fetch of `install.sh` itself failed, `sh` read an empty script and exited 0 —
`make tools-golangci` reported success having installed nothing:

```
$ curl -sSfL <installer> | sh -s -- -b "$(go env GOPATH)/bin" "v2.13.1"   # curl fails
curl: (22) The requested URL returned error: 500
OLD EXIT=0
```

**Fix.** `scripts/ops/install-golangci-lint.sh`, called by `tools-golangci`:

- three bounded attempts (`EVENER_GOLANGCI_INSTALL_ATTEMPTS`, default 3) with 5s
  then 10s of backoff, the installer's own diagnostics left on stderr;
- `set -o pipefail`, so a failed fetch of `install.sh` is a failed attempt
  rather than a silent no-op;
- the version still read from `.tool-versions` and never chosen here; the pin is
  unchanged;
- a post-install check that the binary in `$(go env GOPATH)/bin` reports that
  pinned version.

The retry deliberately does **not** classify the HTTP status. The installer
collapses every cause to exit 1, and reading its log text for a status code
would couple this script to upstream's wording. The cost is that a genuinely
broken pin fails after the attempts plus ~15s of backoff instead of at once.

There is no `go install` fallback: `docs/developing-evener/linting.md` makes
[...]

PR #1182, opened 2026-09-12
docs: SDK migration inventory and PR-by-PR plan for both frontends

obra:
Maintainer goal (2026-09-12): the TypeScript AppWire client library built for the mobile app becomes the underpinning of the web interface too, delivered as small PRs that never break either frontend. This is the map and the sequence; no code changes.

- `docs/design/2026-09-12-sdk-migration-inventory.md`: one row per state/projection/logic module in the web frontend and in `mobile/src` (105 rows): what it owns, wire surface, closest counterpart, classification (SHARED ALREADY 6 / DUPLICATED 36 / PACKAGE CANDIDATE 51 / PLATFORM-ONLY 11 / dead 1), and the seam each needs to move. Plus the package boundary as it actually is today.
- `docs/superpowers/plans/2026-09-12-sdk-migration.md`: 64 sequenced PRs in four phases (packaging 5, pure-logic dedup 7, relocations 24, stores 28), each with scope, the existing tests that serve as the contract, size, dependencies and runtime risk; ~29k lines changed, net shrink ~6-8k. Ends with the three decisions that gate the first code PR, with recommendations.

Findings worth reading before anything else: `protocol/tsconfig.build.json` packs only six of the folder's sixteen modules, so `reducer`, `model`, `activityData` and friends are never qualified as part of the package; `AppwireClientLike` (used by 136 web files) is declared in a test double; the native app imports 124 web frontend modules outside `protocol/` by relative path; `mobile/src` re-implements the reducer rather than importing it.

Part of #1116's post-checkpoint architecture step.

## Review round 17

Head `3749a66ca`. All four findings accepted; one corrected on the way.

- **Medium, C24 native image auth** — accepted the gap, rejected the remedy. A new image-source port is not needed: `mobile-native/src/transcriptImageSource.ts` already returns `{ uri, headers: { Authorization } }` for same-origin hub images, attaching the token only when `url.origin` matches, and is pinned by `transcriptImageSource.test.ts`. C24's row now says `docImageURL` returns the href and each renderer wraps it — browser `` on the cookie, native through `transcriptImageSource(docImageURL(origin, session, path), origin, token)` — and the inventory's doc-pane row carries the same sentence.
- **Low, Fact 5** — rewritten: adding a subpath costs one `packageExports` entry and one `package.json` `exports` entry, and the four consumer programs are generated from them, with the two lists asserting each other. The "four write sites" claim is gone.
- **Low, field names** — `runtimeExports`/`typeExports` replaced by `values`/`types`; zero remnants of either name in both documents.
- **Low, A4 `Deleted:`** — now names the A3 re-export stubs at the old `protocol/` path and points at seam 1, matching the seams section and A4's own risk text.

C24 itself shipped as #1221 (merge `303053dfb`) while this round was being written, so its row is marked done and describes what landed: `DocPort = { origin, fetch }` (`docContent.ts:72`), `docFileRawURL`/`docImageURL` taking the origin, adapters at `panes/doc/browserDocPort.ts` and `mobile-native/src/nativeDocPort.ts`. Status tables updated: nine PRs landed, 55 remain. Totals unchanged at 64 PRs across 71 rows, 29,690 lines.

## Review round 18

Head `878b7992b`. All three findings accepted; two counts corrected on the way.

- **High, phase D publication gap** — real, and fixed with one rule rather than 28 row edits. The phase D intro now states that the phase C publication rule applies to every shared module phase D creates: `tsconfig.build.json` `files`, a published import target (`index.ts` or a declared subpath), and a `packageExports` entry. It folds in the C13 lesson explicitly — **the `shippedModules` entry is what arms the reachability assertion, so a module emitted into `dist/` but missing from that array passes silently** — cites **#1224**, and notes that until #1224's fix lands (after #1222/#1223) every phase-C and phase-D row adds that entry by hand.
- **Medium, D1** — re-audited by grep at `303053dfb`, and **the reviewer is right on all four counts**: native reads session diagnostics (`thread/read` with `includeTurns: false`), filters plugin commands via `visibleCatalogCommands` against `diagnostics.plugins`, merges skills via `mergeSlashCommands`, and refreshes on `evener/plugin/updated` and ref-matched `evener/thread/resync`. Three more it did not list are also contract: a ref-mismatch guard, in-flight coalescing through a `dirty` re-run loop, and `Promise.allSettled` so one failed request does not discard the other. D1 is rewritten as "the web adopts the native logic, which is the superset", deleting nothing, with all seven behaviours named and the pinned tests listed (`mobile-native/src/commandCatalog.test.ts`, `stores/commandCatalog.test.ts`, `slashCompletion.test.ts`). Estimate 250 → 450, since the web gains behaviour rather than losing code.
- **Low, stale `browserDocFetch`/`DocFetch`** — the A3d row, the status table and the inventory delta row now say `panes/doc/browserDocPort.ts` and `readDocFile(session, path, port: DocPort)`. Every surviving mention of the old names sits inside a sentence that says C24 renamed them; zero bare remnants.
- **Count correction:** the `protocol/docContent` import lines are **nine at `303053dfb`, not seven** — four to the root (`DocPane.test.tsx:6`, `browserDocPort.test.ts:2`, `docFile.test.ts:2`, `AgentMarkdown.tsx:3`) and five to the subpath (`DocPane.tsx:9`, `DocPane.test.tsx:5`, `browserDocPort.ts:1`, plus `mobile-native/src/nativeDocPort.ts:4` and `nativeDocPort.test.ts:2`). Seven is the web tree alone; the two native lines are A4's to rewrite too.

C10 (#1222), C13 (#1223), C11a and issue #1224 are recorded as open. Totals: 64 PRs across 71 rows, 29,890 lines; nine landed, three open, 55 remain.

## Review round 19

Head `144ea7bd3` (docs commit `c58e12659`). All three findings accepted; one count corrected.

- **Medium, D1 package boundary** — real, and the fix is a *narrower* port, not a wider one. The native module imports `ConversationClientLike` from `mobile/src/services/conversation`, which a package module cannot reach; but the catalog calls exactly two members (`onNotification` at `commandCatalog.ts:39`, `request` at `:64,65`) while `AppwireClientLike` demands ten. D1 therefore declares a package-local `CommandCatalogClient` as a `Pick` of `AppwireClient` over `request` and `onNotification`, following the precedent already in the package at `activityList.ts:13` (`ActivityClient`). `AppwireClientLike` satisfies it structurally, so both apps pass what they already hold and no adapter is needed; **A2 added to D1's deps**, since that is what makes the web side true.
- **Medium, PR total** — recomputed from the row data rather than adjusted: C11 is now two rows (C11a, C11b), so the tables carry **72 rows**, six of them withdrawn, giving **66 rows of work** and **65 real PRs** (B1b landed inside B1). The honest-total section, the phase-C heading, rule 7 and the subtotals all say 27 phase-C rows / 26 relocations.
- **Low, contradictory phase C/D wording** — rewritten in both documents, with the evidence named. A3 and A4 block the **package-name import rewrite and the directory move**, nothing else; a relocation into `protocol/` lands today with consumers still on deep relative paths, which is how C10, C13 and C24 landed. A legend sentence now tells the executor to read every phase-C `A4` in the Deps column as "consumers get the package-name import when A4 lands", not "cannot start". Every "nothing can move" and "nothing in phases C or D has started" claim is gone from both docs.

**One count corrected:** remaining is **54**, not 56 — 65 real PRs minus the eleven landed (A1, A2, A3b, A3c, A3d, A5, B1 with B1b, B3b, C10, C13, C24). The 56 figure was 65 minus nine, from before C10 and C13 merged. Three of the 54 are in flight: C11a #1225, C6, C26. Status tables record C10 `ba4164649`, C13 `e2c77cc72`, C11a #1225 open, C6 and C26 as open lanes off `e2c77cc72`, and issue #1224.

## Review round 20

Head `cd1607140` (docs commit `b8d3322d0`). All five findings accepted.

- **Medium, D1 deps** — grepped at `b9a98151c` and the reviewer is right: `commandCatalog.ts:4` imports `mergeSlashCommands`/`SlashMenuItem` from `composer/slashCompletion.ts` (**C9**) and `:6` imports `visibleCatalogCommands` from `shell/palette/catalogCommands.ts` (**C15**). Both added to D1's deps, with the row saying D1 imports them from the package rather than co-moving or re-implementing. Its two other imports need nothing: `:5` `sessionActionError` is a root export since A3b, and `:7` `ConversationClientLike` is replaced by the `CommandCatalogClient` port decided in round 19.
- **Medium, C2 testing alias** — defined concretely rather than left to the executor. The file lands at `<package>/testing/navigation.ts` and is imported as `@evener/appwire-client/testing/navigation`, reusing the single `@evener/appwire-client/testing/*` → `<package>/testing/*` entry A3 adds to `tsconfig.paths`, `vite.config`, `vitest.config`, `mobile-native/tsconfig*.json` and the Metro config — no second alias entry, and no `state/navigation/testing.ts` path, which would sit outside the one directory the tarball allowlist already excludes. Its relative imports into `codec`/`merge`/`types` are rewritten to the `state/navigation` subpath C1 publishes, so the fixture depends on what consumers depend on. Still out of `exports` and the tarball.
- **Low, D14** — failure mode corrected. `railExpansion.ts:36-55` catches every failure and returns an empty map, so importing the package on a device does **not** throw; it initialises with no persisted expansion and silently drops every write, which is the harder bug. The persistence-port requirement stands.
- **Low, 18 vs 21** — recomputed from the inventory: 55 PACKAGE CANDIDATE rows less the 34 marked "(2 consumers)" = **21**, and rule 7 now shows that subtraction so the number is derivable.
- **Low, phase B heading** — "9 rows, 2 live PRs", consistent with the honest total.

State recorded: C26 #1226 merged `b9a98151c` (with #1228 filed for the formatter naming clash), C11a #1225 open at `9d2e482d4` and C6 #1227 open at `f6876459e`, both refreshed onto `b9a98151c`. Twelve landed, **53 remain**, two in flight.

**Correction to round 19:** the inventory half of that round's "what A3/A4 actually block" rewrite did not persist — a later edit in the same script threw and the file was never written, and I reported it as done without re-reading. It is applied now, along with the delta-table updates that were lost with it. I have added a read-back check after every edit script.

## Review round 21

Head `b0aa41d74` (docs commit `749e231c8`). All five findings accepted.

- **Medium, A4 runtime resolution** — real, and the fix moves where the check runs. `make/testing.mk:98-102` runs `test-native` before `test-api-package`, and `npm ci` in `mobile-native` never builds `dist/`, so an alias-resolved fixture would pass against TypeScript source Node would never load. The check now lives in **`make test-api-package`**, the target that already packs and installs the tarball: the fixture is copied into that same temporary consumer and run there, resolving out of `node_modules` exactly as a published consumer would — no second pack, no `mobile-native/package.json` change. The alternative (making `mobile-native` a real consumer so `make test-native` runs it in place) is recorded with its real costs: a `file:` dependency plus a `package-lock.json` entry, and a pack ordered ahead of `test-native`, because a `file:` directory dependency links the source and runs no `prepack`, so `dist/` would still be missing.
- **Medium, D8** — the never-echo-a-secret invariant becomes a test the row names: `protocol/instances.test.ts` drives sentinel secrets (API key, credential-JSON blob, device code) through the shared module and asserts each is absent from store state after every mutation, from the `message` and `data` of every thrown error, and from every user-facing string the module produces. Real module, real wire shapes, scripted client — behaviour over real logic.
- **Medium, D27** — the row now specifies the coverage instead of noting its absence: `FakeClient`-driven tests over the **real** dispatcher with a scripted transport, for disconnect mid-mutation, the reconnect after it, a `blockedUnknown` receipt, an `unknown` outcome, dispatch ordering across two refs, and the no-blind-replay rule itself. The citation was already correct — the rule is #1116's standing decision — and the row now says so explicitly rather than trailing a bare number.
- **Low, rule 7** — "26 relocations", consistent with the 27-row phase-C heading after the C11 split.
- **Low, status anchor** — both documents now observe at `2b1e02939`.

State recorded: C6 #1227 merged `2b1e02939`, C11a #1225 open at `0ba4ad45d` on `2b1e02939`, C5 dispatched as an open lane off `2b1e02939`. **Thirteen landed, 52 remain**, two in flight.

## Review round 22

Head `06f1170ca` (docs commits `4dd972e03`, `f4bc46801` merge, plus `06f1170ca` recording C11a's merge, which landed mid-round). All five findings accepted.

- **Medium, A4 runtime check** — real: one constant proves nothing about six rewritten imports. The packed consumer now `import()`s **all five** rewritten entrypoints — `auth-harness-smoke.mts`, `check-hub.mts`, `demo-hub.mts`, `launch-settings-smoke.mts`, `plugin-fixture.mts` — and asserts each resolves, driving every rewritten import through Node's resolver with no smoke logic and no network. Read them all: only `demo-hub.mts:399-400` is importable today, guarded by `import.meta.url === pathToFileURL(process.argv[1]).href`; the other four execute on import or demand arguments (`check-hub.mts:7-9` throws without `ORIGIN TOKEN_FILE`, `plugin-fixture.mts:18,58` branches on `process.argv[2]`), so **A4 adds that same guard to the four that lack it**.
- **Medium, C24 native image auth** — the reviewer is right and my round 17 answer was wrong. Corrected in both documents: what #1221 landed is `nativeDocImageSource(origin, token, session, path)` at `mobile-native/src/nativeDocPort.ts:26`, returning `{ uri: docImageURL(...), headers }`, covered by `nativeDocPort.test.ts:42,46` — and **nothing imports it**, because native has no doc pane. `transcriptImageSource` is a different function for a different surface (`TranscriptImages.tsx`, transcript images). C24 is no longer described as clean for images; new row **D29** owns wiring `nativeDocImageSource` to a real `<Image>` with an integration test proving the header reaches the request, and the plan now says to treat native doc images as unimplemented until it lands.
- **Low, C9** — the allowlist is cited by block (the expected-file list and the `for (const entry of listing…)` assertion loop) rather than a line range that every phase-C row shifts.
- **Low, count** — `~111` → **112** at both sites, matching the inventory's fact 3 and its per-file counting method.
- **Low, A4 wording** — "the namespace import was the only exception, and A3d dissolved it, so A4's grep has no carve-out".

State: C11a #1225 merged `c867646c4`; C5 #1229 (`1452e07d5`) and C15 #1230 (`20ed3f884`) open on `2b1e02939`. **Fourteen landed, 52 remain**, two in flight. Totals now 66 PRs across 73 rows, 30,190 lines — D29 is the one new row.

## Review round 23

Head `153fa8ba3` (round edits `7aef094f6`, merge `f6526efe6`, plus `153fa8ba3` recording C15's merge mid-round). All five findings accepted.

- **Medium, A4** — the reviewer is right that the packed consumer cannot execute the `.mts` scripts, and the ruling is taken as given. The "import the five entrypoints" sentence is gone, replaced by two checks that each say what they prove: **(a)** the scripts import native source modules and `ws`, and four of five reach the protocol through `import type` only, which is erased before Node sees it — so their imports are proved by `tsc` under `make test-native`, and the plan no longer claims runtime coverage for them; **(b)** runtime resolution is proved by a standalone JavaScript fixture in the packed consumer that imports, by name, every runtime *value* the rewritten native tree takes from the package. Enumerated by grep at `c867646c4`: 21 of them, listed in the row, and the fixture's list and that grep **assert equality with each other**, the way the manifest and `exports` already do.
- **Medium, consumer counting** — the unterminated quote is fixed (`git grep -l frontend/src/<module> origin/main -- mobile-native`), and the rule is now stated once: **consumers are counted by direct import; transitive reach does not count, and neither does a type-only import.** Under it, `sendQueueAvailability.ts` is single-consumer — only `Composer.tsx:34` imports it directly, and native's reach is `submitRouting`'s type-only import, which C10 made intra-package anyway. Nothing moves: A1 shipped it. **Two-consumer count 34 → 33**, single-consumer 21 → 22, and rule 7 recomputed from those.
- **Low** — keybindings group `1445` → **1485** in both docs, C3's estimate tracking it (1550 → 1590); phase D header reads "29 rows: 28 store migrations, plus D29"; the inventory's C24 delta row names `nativeDocImageSource` and says D29 wires it.

State: C15 #1230 merged `0acebbb0d`; C11b #1232 open at `91ce8dd37`; C5 #1229 open at `2a7a09c4c`. #1231 is not an SDK row and is not recorded. **Fifteen landed, 51 remain**, two in flight. Totals 66 PRs across 73 rows, 30,230 lines.

## Review round 24

Head `18648b210` (docs commit `a41b5263d`). All three findings accepted.

- **Medium, C2** — real. The `testing/*` alias C2 imports through is something A3 creates, so **A3 joins C1 in C2's deps**, with the reason in the risk cell: before A3 there is no specifier for the thirteen consumers to move to.
- **Medium, A4 and `mobile-native/scripts`** — the fact decides it, and the fact is **no**. `mobile-native/tsconfig.json` declares no `paths` at all; the `paths` block lives in `tsconfig.check.json`, an `include`-scoped config that only `npm run check` (`tsc --noEmit`) reads and `tsx` never does. The scripts are invoked by hand from the README (`npx tsx scripts/check-hub.mts …`, `npm exec -- tsx scripts/demo-hub.mts`) and by no gate. **Branch 2:** they are repo-local tooling, not package consumers — A4 explicitly excludes `mobile-native/scripts` as a named carve-out, they keep relative imports (rewritten by A3 as part of the directory move), the A4 count drops 796 → 790, and the zero-old-path grep now expects exactly those six lines rather than zero. The row also records that rounds 14, 16, 21 and 23 each proposed a way to prove those six imports resolve at run time, and every one was answering a question that only arises if the scripts import by package name — which they now explicitly do not.
- **Low** — `all 34` → "all 34 held at the time — 33 today, since `sendQueueAvailability.ts` left the set in round 23 under the direct-import counting rule".

State: C11b #1232 merged `314281cd5` (C11 complete); C9 and C12 dispatched as lanes off `314281cd5`; C5 #1229 refreshing. The closed recovery drafts and #1123 are not SDK rows and are not recorded. **Sixteen landed, 50 remain**, three in flight. Totals unchanged: 66 PRs across 73 rows, 30,230 lines.

## Decisions settled

Jesse ruled on all four open decisions on 2026-09-13. The section is renamed "Decisions" and each carries its ruling and date; the reasoning is left as written, since that is what the rulings were made against.

1. **Zero runtime dependencies**, as recommended — framework-free stores, `tinykeys`' `parseKeybinding` injected, `ansi.ts` stays in the web tree.
2. **Two layers**, as recommended — the web's wire → `ThreadModel` → display projection becomes the shared shape, and D24 gives native the config-driven content levels it does not have today.
3. **The package moves to top-level `appwire-client/typescript/`** — Jesse: "typescript should be in the path", so clients in other languages can sit beside it rather than under a name that implies one. A3's target path is updated everywhere it appears (the A3 row and title, the seams section, C2's alias mapping), and A3 gains a sequencing constraint: **it runs only when no phase-C relocation PR is open**, because it moves the whole directory and any open relocation branch would conflict in every file it touches.
4. **Derive session usage behind the explicit `scope` label, and never write a derived figure back into the aggregate**, as recommended — the web keeps its derivation, native gains a number where it shows nothing today, and the aggregate stays authoritative.

Also recorded: C12 open as #1233 (`1661e49ae`) and C9 open as #1234 (`c9feb6990`), both on `314281cd5`.

## Review round 25

Head `2af985864` (docs commit `67000fdc3`). All three findings accepted.

- **Medium, A4 contradiction** — resolved as directed, and the resolution is simpler than the thing it replaces. The carve-out means the scripts keep **relative** imports, not package-name ones: A3 rewrites the six lines — `auth-harness-smoke.mts:3`, `check-hub.mts:6`, `demo-hub.mts:6,13`, `launch-settings-smoke.mts:3`, `plugin-fixture.mts:5` — to `../../appwire-client/typescript/…` as part of the directory move, and A4 never converts them to `@evener/appwire-client`. So after A3 nothing anywhere points at `cmd/evener-hub/frontend/src/protocol/`, **A4's acceptance grep expects zero old-path lines including `.mts`** (it previously expected exactly six, which was the contradiction), and **A4's stub deletion cannot break the scripts** because they never pointed at the stubs. Stated in both rows.
- **Low, C11a** — the row and the status table both read "merged as `c867646c4`".
- **Low, D3** — checked rather than assumed: neither `stores/tasksPanel.ts` nor `mobile-native/src/taskList.ts` references the command catalog at all — zero hits for `commandCatalog`, `CommandCatalog` or `evener/command/list` across both. **D1 dropped from D3's deps**; what the two actually share is `panes/session/chrome/taskData.ts` (C25 moves it) plus `protocol/errors` and `protocol/sessionErrors`, both already shipped. Named in the row.

State: C5 #1229 merged `31a5a4370`; C12 #1233 at `3dbfb81b2` and C9 #1234 at `ea74ce97d`, both on `31a5a4370` awaiting review; A3 scouting, starting once those two land, per its sequencing constraint. **Seventeen landed, 49 remain**, two in flight.

## A3 sizing

A read-only scout measured A3 against `31a5a4370`; its facts are now the A3 row and the seams section, and **A3's estimate is 360 lines — measured, not estimated** (was ~320).

- **Non-source references:** 9 build/tool config files (both Vite configs, `frontend/tsconfig.json`, `biome.jsonc`, the editorial-preview and browserguard Vite configs, Metro, `tsconfig.check.json`, `mobile-native/vitest.config.mts`), 2 make targets, 3 CI lines, 5 Go files with functional path strings plus 6 comment-only, 6 active docs.
- **Scenario cards: 12 audit-enforced citations across 8 cards**, where the row said 5 across 2. `scenariosourcecite_audit_test.go` resolves by suffix and `protocol/` is a segment in every one, so none survives the move; all twelve are now listed by file and line.
- **847 import sites must keep compiling** (689 web at five `../` depths, 130 `mobile-native`, 22 `mobile/src`, 6 `.mts`), and **nothing resolves `@evener/appwire-client` by name today** — no root `package.json`, no workspaces, no `paths`, no Vite or Metro alias — so A3's resolver work is additive, across four independent resolvers of which a green `tsc` proves only one.
- **Seam is a hybrid, not stubs everywhere:** stub the web (689 lines across five depths would *lengthen* every specifier and A4 would rewrite them again — ~30 one-line stubs instead), sed the mobile trees (158 lines at one uniform depth each, paths get *shorter*, and it is the rewrite the `.mts` carve-out already needs). ~190 changed lines plus ~31 files against ~847 rewritten. The 18 modules both trees import are listed as the first smoke-check.
- **Three risks are now named A3 sub-steps**, each because it ships green and fails later: (1) `searchForWorkspaceRoot` resolves to the frontend dir, so the moved package falls outside `fs.allow` and 403s under `npm run dev` and all five browser guards while `tsc`/build/vitest stay green — two files, since the editorial-preview config overwrites `fs.allow` twice; (2) the package's 25 test files fall outside the vitest root and their bare `react`/`vitest`/`@testing-library` and `?raw` imports will not resolve from a repo-root sibling — A3 must state where they run rather than lose them silently; (3) `biome ci src` leaves the package unlinted and unformatted, and the web coverage denominator moves, so the Biome scope change and the **`web 96.1` floor re-baseline** are A3 steps, not the next PR's surprise.

## Review round 26

Head `fc1b99782` (docs commit `7e56398fd`). All three findings accepted.

- **Medium, C11b / `deriveAskQuestions.ts`** — the reviewer is right under the direct-import rule this plan wrote in round 23. Native's only import is `questionBatches.ts:1`, `import type { AskQuestionRef }` — type-only, which does not count — so the module was **single-consumer at run time even when C11b moved it**, and native's own copy of the rule stayed behind in `mobile/src/conversation/project.ts:182-256`, a file that names what it mirrors: "Mirrors parseAskUserQuestions (protocol/askShared.ts)" at `:183` and "Mirrors liveAskQuestions" at `:259`. Recorded that way in both the C11b row and the inventory row, with the move explicitly **not** reverted: `askShared.ts` is a genuine two-consumer module and imports it, the module is in the package, and the web uses it at run time. The duplicate is now named work rather than a note — **D22** owns deleting it, not D20, because the parser is called from the projection (`:272`, `:355`, `:469`) that D22 replaces, while D20 owns the dock and batch bookkeeping, which is a different piece. D22's scope and estimate grew accordingly (1100 → 1200).
- **Two-consumer set recounted: 33 → 32**, single-consumer 22 → 23, and rule 7 recomputed from 55 − 32.
- **Low, A3 doc sweep** — `docs/web-ui/parity/parity-m6-surfaces.md:166` added to A3's named doc list.
- **Low, C3** — the "add and qualify zustand and tinykeys" alternative is gone; the row now states the framework-free registry plus the injected `parseKeybinding` port as the only path, since decision 1 is settled at zero.

State: C12 #1233 merged `a0e594122`, C9 #1234 merged `d7ff88653`; **A3 is implementing** on `claude/sdk-a3-package-move`, its sequencing constraint satisfied now that both landed. #1145, #1235 and #1239 are not SDK rows and are not recorded. **Nineteen landed, 47 remain**, one in flight. Totals 66 PRs across 73 rows, 30,370 lines.

## Review round 27

Head `446080ff7`. All seven accepted; item 7 corrected a claim this plan had been repeating since round 13.

- **Medium, dependency direction** — `deriveAskQuestions.ts` imports `askShared.ts`, not the reverse. Rewritten that way in the C11b row, the phase-C sentence and the inventory row; the co-move conclusion is unchanged, since the direction still forces them across the boundary together.
- **Medium, A4 runtime fixture** — specified concretely: a plain **`.mjs`** file (no types, nothing to strip) copied into the packed consumer beside the runner's existing runtime consumers and executed as **`node resolve-check.mjs`**, the same shape the runner already uses (`run(process.execPath, [join(consumerDir, consumer)], consumerDir)`). The `.mts` wording is gone.
- **Low** — the phase-C header rule now excepts "C2 and any test-only fixture, which go under the in-repo `testing` alias and never into `exports` or the tarball"; "112 importers **in the web app**", with the fork-`FakeClient` argument's native half dropped in favour of the real objection (it separates the double from the catalog it validates against); `33 → 32` at :426; `~111 → 112` at :76.
- **Item 7, and the fact:** I checked rather than kept the claim. `qualify-package.mjs:444-456` invokes tsc with `--strict --noEmit --module NodeNext --moduleResolution NodeNext --target ES2022` and **no `--lib`**, so the declaration consumers get TypeScript's default for that target — `lib.es2022.full.d.ts`, which **includes DOM**. `Response` resolves there, so the "a `Promise<Response>` fails the runner's DOM-free consumers" justification was wrong. Both documents now describe `DocResponseLike` as a deliberate **host-independent API** — the package names the response shape it actually reads, so a browser `Response`, React Native's and a Node `undici` one all satisfy it structurally and no consumer is obliged to carry the DOM lib — and say plainly that an earlier draft claimed otherwise.

Totals unchanged: 66 PRs across 73 rows, 30,370 lines; nineteen landed, 47 remain, A3 in flight.

## Review round 28

Head `c649e3b36` (commits `ba5c29b56` and `c649e3b36`). All four accepted.

- **Medium, A4 fixture** — named `resolve-check.mjs` everywhere, the `.mts` spelling gone. The symbol lists are re-derived from the post-relocation import graph at `eeff54b70`, not the round-23 list of 21, which predates C9, C12, C15 and C5. The grep, stated in the row so the fixture can assert equality against it: value specifiers — not `import type`, not inline `type` members — in every `mobile-native` and `mobile/src` file importing `…frontend/src/protocol/<module>`, excluding `mobile-native/scripts` per the carve-out. **Root: 27** — including all six the reviewer named (`buildComposerInput`, `mergeSlashCommands`, `parseSlashToken`, `spliceSlashCommand`, `reconcileBatches`, `visibleCatalogCommands`) — and **`./docContent`: 2** (`docImageURL`, `readDocFile`).
- **Medium, A3 test ownership** — the ruling replaces the either/or: the package's tests keep running under the frontend's vitest, A3 extends `include`, `server.fs.allow` and the coverage `include` to `appwire-client/typescript/**`, resolution stays through the frontend's `node_modules`, and the rejected alternative is named with its cost (duplicate react/testing-library dragged into the offline tarball install). The silent-drop risk gets a guard rather than a promise: `make test-web` asserts the package test-file count vitest ran equals the count on disk.
- **Low, importer counts** — verified by grep and now stated as files *and* lines: `f2599d1ed` baseline **136 files / 137 import lines**; post-A2 `3bf357337` and today **112 files / 112 lines**; **24 files stopped importing it**. A2 rewrote 25 `AppwireClientLike` imports, and the two sets are explicitly not identical — `shell/AppShell.tsx` and `stores/threads.test.ts` were rewritten and still import `testing/fakeClient` because they also use `FakeClient`. That pair is now used in fact 3, the A2 row, A4 and "Where test support lives".
- **Low, two-consumer count** — one narrative in both documents: 34 at the `f2599d1ed` re-audit, 32 today, `sendQueueAvailability.ts` in round 23 and `deriveAskQuestions.ts` in round 26.

## Review round 29

Head `c03ed0b11`. Both accepted.

- **Medium, C23** — real, and the row was mis-filed. `disclosureStore.ts:7-8` imports `useStore` from `zustand` and `createStore` from `zustand/vanilla`, and `:45` `isDisclosureOpen` *is a React hook* over a module singleton, so relocating it changes logic and cannot obey phase C's no-logic rule. **It moves to phase D as D30**, rather than becoming a third phase-C exception — the intro's blanket sentence stays true, and the row sits with the other store migrations it now resembles. Rewritten in the phase-D shape under settled decision 1: a framework-free `createDisclosureStore()` returning a `getState`/`subscribe`/`setState` triple plus the named pure helpers; the web adapter keeps `isDisclosureOpen` as a hook over zustand's `useStore`; a native adapter at `mobile-native/src/nativeDisclosure.ts`; root export with its `packageExports` entry per the phase-D publication rule; `disclosureStore.test.ts` stays the oracle for the core, with an adapter test each proving the hook re-renders on a store change. **Estimate 210 → 380.** Phase C is now 26 rows / 25 relocations, phase D 30 rows.
- **Low, A3 figures** — the measured ones win everywhere: **855** import sites, **138** `mobile-native`, **27** test files, and the seams paragraph's sed count corrected to **166 lines** (138 + 22 + 6) with 33 new files (32 web stubs plus the `testing/` stub directory, as #1241 built it). Both documents grepped: the only surviving `847`/`130` are the two clauses that deliberately name them as the scout's superseded figures.

Totals: 66 PRs across 73 rows, 30,540 lines (A 2,860 / B 210 / C 8,380 / D 19,090).

## Review round 30

Head `533d23761`. All five accepted.

- **Medium, `builtinInvocation.ts`** — grep-verified on main, and it is a **runtime two-consumer module**: `mobile-native/src/composerCommand.ts:1-4` imports `findBuiltinArgument` and `matchBuiltinInvocation` as **values**, no `type` keyword. 43 lines, zero imports of its own, four web import sites (`Composer.tsx:73`, `builtinCommand.ts:19`, `spawn/Spawn.tsx:67`, `spawn/spawnSlashMenu.ts:12`). So it gets an inventory row and a phase-C row, **C27**, deps A4 like its siblings, ~150 lines. Counts it moves: two-consumer set **32 → 33**, PACKAGE CANDIDATE 55 → 56, inventory 108 → **109 rows**, phase C 26 → **27 rows / 26 relocations**, plan 66 → **67 PRs**, 30,540 → **30,690 lines**. It was simply missing from this document until now.
- **Medium, A3 test-file count** — 27 everywhere, and the guard is described as #1241 built it: `make test-web` asserts the ran-count equals the count **derived from the files on disk**, so the expected number is never hard-coded.
- **Low, both-tree module list** — `deriveAskQuestions` dropped, 18 → **17**, with a sentence saying why: native's only import of it is type-only, which the direct-import rule does not count.
- **Low, relocation count** — the heading and rule 7 now agree at 27 rows / 26 relocations, naming all three adjustments (C11's split, C23 → D30, C27 added).
- **Low** — decision 1 now names D30 as the row that was C23; the phase-D heading reads "30 rows: 29 store migrations, plus D29".

State: #1241 at `2fb3d5fa2` awaiting round 2; **#1245** (the Metro gate for #1244, which A4 depends on) at `3c4d54c60` with a round-1 fix in progress — both now in the status table. Nineteen landed, **48 remain**, two in flight.

## Review round 31

Head `635579a4d`. Both accepted.

- **Medium, C27 importers** — grepped on main, tests included. **Six sites in six files:** five web — `Composer.tsx:73`, `builtinCommand.ts:19`, **`builtinCommand.test.ts:9`**, `spawn/Spawn.tsx:67`, `spawn/spawnSlashMenu.ts:12` — and one native, `mobile-native/src/composerCommand.ts:1-4`. All six are now named in the row. Two further facts the grep settled: **no `vi.mock` anywhere names the module**, and it has **no test file of its own** — `builtinCommand.test.ts` covers `builtinCommand.ts` and reaches `matchBuiltinInvocation` through it, so that test stays where it is and is repointed rather than moved, and the module would arrive in the package with no oracle. C27 writes one; the row says so.
- **Low, stale counts** — the round-30 reconcile left three behind, all fixed: the inventory's method sentence and the plan's re-audit sentence both read **"34 then, 33 today"**, each naming all three moves (`sendQueueAvailability.ts` round 23, `deriveAskQuestions.ts` round 26, `builtinInvocation.ts` joining in round 30), and rule 7 reads **"not 56"**. Grepped both documents: zero `32 today`, zero `not 55`. Derivation re-run from the tables — **56 PACKAGE CANDIDATE rows − 33 two-consumer = 23 single-consumer**, which is what rule 7 states.

Status rows refreshed: #1241 at `5062edf34` awaiting round 3, #1245 at `4c78b34a0` awaiting round 2. Totals unchanged: 67 PRs across 74 rows, 30,690 lines; nineteen landed, 48 remain.

## Review round 32

Head `0772e0836`. All three accepted.

- **Medium, shared-runtime list 17 → 19.** Verified each import is a **value** import, not a type one: `questionBatches.ts:2` takes `reconcileBatches`, `composerCommand.ts:5` takes `mergeSlashCommands`, `CommandCompletion.tsx:4` takes `filterSlashMenuItems`. Both modules join the list, and the A3 corrections sentence that called them web-only is fixed: of the four specifiers it added, only `clientLike` and `model` have no native importer at all (checked — zero each), while `reconcileBatches` (2) and `slashCompletion` (5) are value-imported by native. `deriveAskQuestions` stays excluded, with the reason given once: native's only import of it is type-only, which the direct-import rule does not count.
- **Low, test count** — "drops all **27**", with the clause that #1241's round 3 moves `tokenFlood.test.tsx` back into the web tree and takes it to 26, closing #1242; the guard derives from disk either way, so neither number is written into it.
- **Low, PR heads** — every head is now cited as "head at the time of this commit: `<sha>`" (A3 row, status table, inventory delta), all refreshed to `5062edf34` and `4c78b34a0`, with a standing note that heads go stale immediately and merge SHAs are the stable reference. Status sections retitled and dated.

**One correction to the brief:** the status sections are dated **2026-09-13**, not 2026-09-14 — that is today's date here, and dating a commit into the future would defeat the point of the convention. If your clock reads the 14th, say so and I will move both.

Totals unchanged: 67 PRs across 74 rows, 30,690 lines; nineteen landed, 48 remain.

## Review round 33

Head `5a5e98a92` (docs commit `fe100dfda`). All three accepted.

- **Medium, D30 binding** — specified, because this is precisely what a relocation would get wrong. `createDisclosureStore()` returns the store-bound actions and selectors themselves (`getState`, `subscribe`, `setState`, `isOpen`, `setOpen`, `toggle`, `beginBaseline`, `clearScope`, `defaultFor`), so nothing in the core reads a module-level singleton and two instances cannot see each other's state; the **web adapter owns the one instance it exposes**, creating it at module scope and wrapping `isOpen` in the `useStore` hook its callers already use, and the native adapter creates its own. The row's test is an **isolation test with two store instances** — open a scope in one, assert the other still reports the fallback, assert `clearScope` on one leaves the other untouched — which is exactly the assertion that would fail against today's `disclosureStore.ts`, since it has the singleton this removes. Estimate **380 → 420**.
- **Low, Go count** — recounted from the branch diff rather than `gh pr view`, whose file list is capped at 100 and showed only four. A3 touches **10 Go files**: **4 functional** (`appwire/doc.go:29`, `appwire/protocol_test.go:335,360`, `internal/appwirets/emit_test.go:658`, `makefiletargets_audit_test.go:1073`) and **6 comment-only** (`agent/session_events.go`, `cmd/evener-hub/e2e_control_invariant_test.go`, `cmd/evener-tui/hub_model.go`, `internal/appwirets/main.go`, `server/appwire_runtime_test.go`, `server/appwire_turns.go`), once main's unrelated breaker commits are excluded. The list was right and the count was wrong; both now say four.
- **Low, honest total** — `mobile-native/src/commandCatalog.ts` (105) moved out of the deleted-twins list into a relocated-into-the-package clause, since D1 makes it the shared module. Net reduction corrected from ~3,900 to **~3,800**, with the double-count named.

Totals: 67 PRs across 74 rows, **30,730 lines**.


obra:
Round 1 disposition (head `777c8e2bd`, main `2245f9715` merged): all ten findings accepted and corrected in `f2289bf19` (inventory) and `0e063359c` (plan), each verified against the code first.

- **High, A3 path updates** — the row now names the eight functional files that hard-code the package path (`appwire/doc.go:29`, `make/testing.mk:58`, `make/linting.mk:190`, `internal/appwirets/emit_test.go:658`, `appwire/protocol_test.go:335,360`, `makefiletargets_audit_test.go:1073`, `.github/workflows/ci.yml:57,60`, `ios-testflight.yml:74`; two of those the review did not list) and its oracle is widened to `make generate` + `make lint`.
- **Medium, A2** — scope corrected to the 25 `AppwireClientLike` importers, `FakeClient` imports preserved, type exported from `index.ts`; already done in #1188.
- **Medium, C3 / C9 / C18 / C21 / D14** — zustand and tinykeys now gate C3 on decision 1; C9 depends on C15; `LaunchConfigLayerName` moves in C18; `catalogView.ts` moves with C21; D14 defines a host persistence port and `railExpansion.ts` (`localStorage` at init) joins the never-moves list.
- **Low** — export counts stated as nine statements / eleven runtime exports with four consumer files; native import figures corrected to 204 / 88 / 39 distinct modules with the counting method; C24 labelled a deliberate rule-7 exception.

Also folded in since the first draft: #1184, #1188, #1189 open; #1186 merged; a third divergent failure predicate (#1190) becomes inventory row and PR B1b. Totals: 106 rows, 65 PRs, ~29,300 lines.

obra:
Round 2 disposition (head `34fb2274c`): all seven findings accepted, corrected in `9a40c6059` (inventory) and `34a674ee6` (plan). Test support gets a permanent home — `protocol/testing/` stays inside the package and out of the tarball behind a non-shipped in-repo specifier, A3 moves it with the package and A4 rewrites the ~111 `FakeClient` imports in the same sweep (forking the fakes per app rejected: `fakeClient.ts:20-30` validates scripted methods against `METHOD_NAMES`, and two copies drift). `immutable.ts` is documented as a transitive-dependency exception to rule 7 (`codec.ts:8` and `merge.ts:10` both import it). Counts reconciled with one stated method: §6 88, A4 856 import statements (652 web, 204 native), four generated consumers, A3's comment paths (seven mentions in six files beyond the eight functional ones), 3,864 lines for the two `mobile/src/state` files, 3,502 for the ten unpacked runtime modules.

obra:
Round 3 disposition (head `02a43c81e`): all five accepted, corrected in `1b86c5747` (inventory) and `1479c6ecf` (plan). A4 now rewrites all three trees (652 web + 204 `mobile-native` + 21 `mobile/src`) with a zero-old-path grep across both native trees as the acceptance check. The A1/A3 ordering is resolved the way it actually went: A1 ships modules into today's directory (#1184), A3 performs the move, decision 3 gates A3. C11's real scope is recorded — `deriveAskQuestions.ts` → `askShared.ts` → `transcript/tools/helpers.ts` (`parseArgs`, no imports of its own), both now inventory rows and inside C11. Headline totals recomputed from the rows (12,336; 7,615, with a note that #1186 has since taken `mobile/src` to 6,841); `errors.ts` has nineteen exported declarations, three re-exported. Now 107 rows, 65 PRs, ~29,600 lines.

obra:
Round 4 disposition (head `234570fc7`): all six accepted, in `419b5ffa2` (inventory), `4230dce32` (plan) and `234570fc7`. A4 is scoped to protocol imports only (652 web + 116 `mobile-native` + 21 `mobile/src` = 789), its oracle narrowed to `protocol/` specifiers with the inverse check that the broad grep still returns the 88 phase-C sites. The `docContent` row now states the truth — same-origin-relative `/doc/...` paths and global `fetch` with `same-origin` credentials — and C24 owns the injected origin/fetch/auth port plus the `readDocFile` re-export that #1184 deliberately omits from the entry point until then. `errors.ts` 19 in §5; `pathListAdd.ts` (54) in the C18 row; phase subtotals recomputed from the rows (A 2,340 / B 1,600 / C 8,040 / D 18,010 = 29,990) after fixing a malformed C24 row; the testing row reads "136 web files import `testing/fakeClient`; 25 name `AppwireClientLike`".

obra:
Round 5 disposition (head `dfab9aa89`, main `f2599d1ed` merged), plus a grep audit prompted by executing B7 and B3:

**Audit.** Every DUPLICATED row and every two-consumer PACKAGE CANDIDATE was re-checked by grep at `f2599d1ed` for the exact symbol, file:line and pinning test on each side. All 34 two-consumer relocation claims hold. The phase-B "one rule written twice" rows do not: B2, B3, B5, B6 and B7 each pair a web rule with a native function that fills the same slot but implements a different rule (`taskGroups.ts:16` partitions rows vs `services/activity.ts:409` counts off an aggregate; `detailsAccounting.ts:69-89` sums turns vs `services/activity.ts:437-451` copies fields and `state/activity.ts:23-27` forbids the sum; `toolRuns.ts:56` folds at 3 vs `project.ts:528` at 2 over a different input; `pendingReconcile.ts:111` reconciles outbox states vs `project.ts:759` copies a field; liveness/threadTitle have no native twin at all). Seven rows reclassified, B7 withdrawn, B2/B3/B5/B6 withdrawn with reasons, B4 folded into D24 (`messages/systemGrouping.ts` exists; its counterpart is `projector.ts:112-130`, not the family classifier). The inventory header now states the verification rule. New DUPLICATED row: native's two `projectUsage` copies (#1203).

**Round 5 findings.** H1 B7 withdrawn. H2 D25 rescoped to a headless queue core with React/storage/thread-store adapters per frontend. M1 D15 names `selectors.ts:238`'s `railNodes` type dependency. M2 `transcript/tools/taskData.ts:18` folded into a new C25. M3 C24 gains an authenticated image-source port (`{uri, headers}`) with its own assertion. M4 the status table and §5 tarball cells read "yes, since #1184", with `readDocFile` off the entry point. L1 §5 headline 8,032 → 7,233 (+799 `testing/`). L2 `LICENSES/beautiful-ui.txt` is outside the package and `files` omits it: C9 must move the license or leave the code. L3 `mobile-native/src/jobOutput.ts` dropped from the deleted-twins list.

**Decision 4 added** for the maintainer: derive a session usage figure from per-turn usage when no cumulative exists (web) or refuse and rehydrate (native)? Recommendation: derive, behind the explicit `scope` label, never written back into the aggregate.

Totals: 108 rows (6/33/55/13/1); 61 live PRs, ~29,140 lines (A 2,340 / B 350 / C 8,340 / D 18,110).

obra:
Round 6 disposition (head `ae9a06ea7`, main `3bf357337` merged): all six accepted in `dd389d756`. A4 covers the six protocol imports in five `mobile-native/scripts/*.mts` files (count 789 → 795; the zero-old-path grep gains `--include='*.mts'`). Status tables normalized in both docs (#1184 `f2599d1ed`, #1188 `3bf357337`, #1203 `4da382482` merged; #1189 open; B3b row added). §0 facts 1, 2 and 5 are labelled the pre-A1 snapshot with post-A1 values inline (one shared `runtimeExports` array at `qualify-package.mjs:69` feeding the four consumers at 163/177/201/207, measured at `3bf357337`). C6 depends on A4, C5 on C6, C7 on A4 (`composerInput.ts:2` imports `./attachmentMarkers`, which imports nothing). Two-consumer rows 34; 25 relocations; phase B nine rows, six withdrawn.

obra:
Round 7 disposition (head `c3c938ecf`, main `27503c07d` merged), in `947ee13d2`:

- **High, A4 breaks on `protocol/errors` symbols** — measured at `27503c07d`: fourteen symbols the apps import are absent from `index.ts` (twelve `errors` declarations, `reducer`'s `chunkViewBackingForTests`, `docContent`'s `readDocFile`). New plumbing PR A3b exports the twelve at the root before A4; the test hook moves to the non-shipped `testing` specifier; `readDocFile` stays held for C24. Root exports over per-module subpaths, because one entry exists today, a symbol is a two-file one-liner, and each subpath multiplies the gate. A4 depends on A3b and gains an "every imported symbol resolves from the package" oracle.
- **High, C1 relocates navigation with no published target** — stated once as a phase-C rule: every relocation adds its module to the build, gives it a published target (`index.ts` re-export or a `package.json` subpath) and adds its symbols to `runtimeExports`. C1 publishes a `state/navigation` subpath and is flagged as the row that settles the subpath question for every later state move.
- **Low, staleness** — root-caused: §§0, 2, 3, 4, 5, 7 are frozen at `92561dbe3` and a "Delta since baseline" table carries the merges (A1 → 16 files, A2 → 17 + `clientLike.ts`, A5 → fixtures deleted, B1 → 18 + `itemFailure.ts`, B3b). Fact 3 names `protocol/clientLike.ts` (114 `testing/fakeClient` import lines, none importing the type); §6 cites the `runtimeExports` manifest; §7's parts sum to 108.

Totals: 108 rows; 62 PRs across 68 rows, ~29,330 lines.

obra:
Round 8 disposition (head `9771fc9c4`): all four accepted in `9771fc9c4`, plus `b8835217a` recording A3b as #1206 and the one line A4 cannot rewrite mechanically (`DocPane.test.tsx:5` namespace-imports `docContent` to spy on `readDocFile`; it stays relative until C24 publishes `docContent` as a subpath).

- **B1b** — done: it landed inside B1 (`27503c07d`), `toolRenderers.ts:190-193` delegates to `hasItemFailure`, #1190 and #1197 are closed; row zeroed, status and totals updated.
- **Root-only manifest vs subpaths** — new C0 heads phase C: the flat `runtimeExports` array becomes `{ specifier: [names] }` with one ESM declaration check, one CJS declaration check and one runtime presence check per specifier, stated once as the rule for every subpath (`state/navigation`, `testing`, `docContent`); C1 depends on it.
- **`chunkViewBackingForTests`** — A3 adds a non-shipped `testing/reducerHooks.ts` re-export and A4 splits the `reasoningFormat.test.ts:3` import, rather than an exemption: A3 moves the whole directory so an exempted line would still need rewriting, and the alternative leaves a white-box hook one root export from publication. A4's exemption list has exactly the DocPane entry.
- **C24 oracle** — deterministic injected fetch and image adapters assert the request URL from a base origin, auth-header propagation into both the fetch and the image source, 403/404/500 → `DocFileErrorKind`, and the truncation headers → `truncated`/`totalBytes`.

Totals: 63 PRs across 69 rows, ~29,370 lines; six landed, A3b in flight, 56 remain.

obra:
Round 9 disposition (head `34b4b3f7f`, main `99fa1882f` merged): all accepted in `5a1d9a06c`. C1 publishes `state/navigation` through one barrel (`state/navigation/index.ts`), the thing later state relocations extend rather than multiply. C0's manifest is `{ specifier: { runtimeExports, typeExports, shippedModules } }` — reachability and type checks are as blind to a subpath as the runtime list, so all three go per-specifier; its rule is "a specifier in `exports` with no manifest entry is not qualified", and the in-repo-only `testing` alias is the named exception, validated by the apps' typecheck. `transcript/messages/format.ts` (87 lines, zero imports, one native site) gets C26. C24's stale "runner cannot call `readDocFile`" clause is gone; the injected fetch and image-source assertions are the runner's. B1/B1b count as one landed PR (62 real PRs, phase B 210); A3's sweep names the active Markdown that runs against the old path and excludes the dated plans/specs; A4's grep excludes `node_modules`; the phase-C sentence names C0 and C24 as the non-relocation rows; `mobile-native` counts are 210/122 with the 204/116 `.ts`/`.tsx` split shown; the §3 `toolRenderers.ts` row names its baseline counterpart with the `itemFailure.ts` resolution in the delta table. Totals: 62 PRs across 70 rows, ~29,520 lines.

obra:
Round 10 disposition (head `0fc6d915c`): all seven accepted in `0fc6d915c`. C24 keeps `docImageURL` string-returning (`DocPane.tsx:95` and `AgentMarkdown.tsx:56` feed it to ``, `DocPane.test.tsx:140,163` assert on `src`) and defines two named adapters — browser (string image source, same-origin fetch) and native (`{uri, headers}` bearer image source and fetch) — with every caller to update listed. `stores/navigation/testing.ts` has ten consumers, all tests across both apps, so it goes under the in-repo `testing` alias in C2 ahead of C1: never in `exports`, never in the tarball, gated by the apps' typecheck. A4's zero-old-path check now also sweeps the web source tree (A3's stubs would otherwise let a stale deep import outlive A4), allowing only the documented DocPane line; the mobile count is 22 post-B1, total 796. Rule 2 allows one module plus its pure transitive dependencies, naming the seven rows that use it. Rule 7 says 25 relocations; the honest total is 64 rows of work, 63 real PRs, 6 landed, 57 remaining; phase B's stale "350" reads 210. The phase-C closing paragraph names the right modules and marks `format.ts` as C26's exception. The status section is observed at `99fa1882f`.

obra:
Round 11 disposition (head `bf9f79fe4`): all five accepted in `bf9f79fe4`. The `readDocFile` sequencing is resolved by publishing before rewriting: new A3d publishes `./docContent` with `readDocFile` behind an injected fetch (the web installs a browser adapter that preserves today's behaviour), and the subpath-aware manifest work moves from phase C to A3c since A4 depends on it. A real subpath also gives `vi.spyOn` a per-module namespace object, so A4's exception list is now empty and both zero-old-path greps are absolute; C24 keeps only the base origin and the two platform adapters. C2 scopes all thirteen consumers (six web, seven native; the earlier ten was a truncated count). Rule 7 reads 25 relocations against the 26-row table; the delta table's A3b row is contiguous again; the dead cross-reference points at "Delta since baseline". Totals: 64 PRs across 71 rows (65 of work), ~29,650 lines; six landed, 58 remain.

obra:
Round 12 disposition (head `dfa6e5917`, main `cb211c5f8` merged): all six accepted in `25ce5e77d`. A3d's oracle requires moving `docContent.test.ts` onto the injected-fetch seam plus a browser-adapter test for `credentials: "same-origin"` (done in #1209). A4 gains a runtime resolution check for the six `mobile-native/scripts/*.mts` imports (`npm exec -- tsx scripts/check-hub.mts --help` or equivalent, with the alias or `file:` dependency that makes it resolve), since no gate runs those scripts today. B1's row reads merged; C2 counts five web tests, seven native tests and one runtime consumer (`editorial-preview.html` is a real Vite entry) with a preview-entry build check; rule 7 states one number; the phase-C intro excepts C24; D17's cell names what it deletes on the web and the native consumer it converges on. #1209 is recorded on A3d, C24 and the delta table: `readDocFile` takes a required `DocFetch` returning `DocResponseLike`, the web adapter is `panes/doc/browserDocFetch.ts`, C24's native adapter must return that shape, and A4's `docContent` split is three root / three subpath (`browserDocFetch.ts:1` imports the `DocFetch` type the root does not export). Totals: 64 PRs across 71 rows, ~29,690 lines; seven landed, 57 remain.

obra:
Round 13 disposition (head `bcabbc174`, main `f39aa2c83` merged): all eight accepted in `82ef154b9` and `bcabbc174`. A3d's row records that it wires the only `readDocFile` caller itself (`DocPane.tsx:49` → `browserDocFetch`, done in #1209), so the web is type-correct the moment it lands. C24 states where the base origin enters: the injected port owns URL resolution, `docFileRawURL`/`docImageURL` gain an origin parameter, and `readDocFile` composes through the port rather than a fourth positional argument; the browser adapter passes the empty same-origin base so today's `/doc/...` strings are unchanged. Rule 7 reads 25 relocations; every manifest reference is `packageExports` with no line numbers; B1's cell reads merged; the last `C0` identifiers are `A3c`; A2's hash is `3bf357337`; and the status sections read "A3 and A4 are unstarted" with A3d recorded as merged (`f39aa2c83`). Totals: 64 PRs across 71 rows, ~29,690 lines; eight landed, 56 remain.

obra:
Round 14 disposition (head `507fa808f`): all four accepted in `507fa808f`. A4's runtime check is a purpose-built `mobile-native/scripts/resolve-check.mts` (no arguments, no network, imports the runtime export `APPWIRE_PROTOCOL_VERSION` from the package and asserts it) run by the gate, since `check-hub.mts` needs positional arguments and its only protocol import is type-only. The "24 relocations" survived three rounds because the text was line-wrapped as `schedules 24` / `relocations`; a whitespace-tolerant scan now finds only "25 relocations". The phase-C intro excepts C24. C24's row states the published API change outright: A3d's `DocFetch` receives a resolved URL and cannot supply an origin, so C24 widens it to a `DocPort` (`{ origin, fetch }`) or an origin-bearing sibling, changing the `./docContent` signature and its `packageExports` entry together — flagged as a breaking change to a published specifier now that A3d is on main.

obra:
Round 15 disposition (head `5efba7be7`): all seven accepted in `5efba7be7`. The runner's true shape at `f39aa2c83` is `packageExports[specifier] = { values, types, esmTypeUses, cjsTypeUses, smoke }` with `exports` and the manifest asserting each other, and reachability as a separate check that only recognises `from "./<module>"` re-exports from a specifier's declaration entry — so a nested barrel under `state/navigation/` would read as unreachable as-is; C1 gains a runner change as its first step (resolve re-export specifiers relative to the declaring file, add the nested names), stated on A3c and C1. C18 names all three `LaunchConfigLayerName` importers. C24 is scoped to the origin port and the two platform adapters: there is no image-source port in the published API, the runner's image assertions are dropped, and native image authentication is proved by an explicit native-adapter test. The observation commit is `f39aa2c83`; C24's citations are current; `tsconfig.build.json:14`; and A3/A4 block every phase-C relocation while C24 depends only on A1 and A3d, both merged.

obra:
Round 16 disposition (head `ce879caae`): all four accepted in `ce879caae`. A4 wires `resolve-check.mts` into a gate — a `mobile-native` package script `check:package-resolution` invoked from `make test-native` beside the existing steps. C9 ships the license: copy into the package, add to `files`, widen the runner's tarball allowlist (`qualify-package.mjs:325-340`) and its assertion in the same relocation, rather than keep a two-consumer module out of the package over a one-line allowlist entry. The last claim that the runner observes auth headers is gone — `DocFetch` receives only a URL, auth is proved in the browser and native adapter tests, and C24 is titled for its real scope (an injected origin and the two platform adapters). The `testing/fakeClient` count: measured at `f39aa2c83`, 114 lines mention it, three are comments, and exactly 112 files carry one import line each; both inventory sites now say 112 with the counting method inline, since neither 111 nor 114 reproduced.

PR #1184, opened 2026-09-12
feat(protocol): ship every protocol module in the package

obra:
PR A1 of the SDK migration plan (#1182; part of #1116's post-checkpoint architecture step).

`protocol/tsconfig.build.json` packed six of the folder's modules. The other nine — `activityData`, `activityList`, `activityMerge`, `jobOutput`, `model`, `reducer`, `sendQueueAvailability`, `sessionErrors`, `stableDelegate` — compiled inside the apps but were never in the tarball, never exported from `index.ts`, and never exercised by the qualification runner, so an installed consumer could not reach them and nothing proved they worked outside the checkout.

Now all sixteen ship (`.js` + `.d.ts`), `index.ts` re-exports their public symbols (types as `export type`; the white-box `chunkViewBackingForTests` stays private), and the runner's four generated consumer programs build from one list of the entry point's runtime exports: both runtime programs call at least one export of every shipped module on a trivial input, and the tarball check asserts each shipped module's files are present. `docContent.ts` ships too (round 1): only `readDocFile` touches `fetch`, at call time, so its pure helpers and error type are exported and the runner smoke-calls `docFileRawURL` rather than `readDocFile`. Zero runtime dependencies, unchanged. No app file outside `protocol/` changes.

RED (runner extended, old build config): 170 `error TS` lines, first `Module '"@evener/appwire-client"' has no exported member 'isFailedJobOutcome'`. GREEN: `qualified @evener/appwire-client@0.1.0: installed imports, declarations and read-only example`. Flipping one smoke assertion fails the run, so the per-module calls execute inside the installed consumer.

Gates: `make test-api-package`, `make test-web`, `make test-native`, biome.

## Review round 1

Low, `docContent` held back although only `readDocFile` uses a browser global — accepted; `a50c06875` ships it (`DOC_FILE_MAX_BYTES`, `DocFileError`, `docFileRawURL`, `docImageURL`, `readDocFile`, and the `DocFileContent`/`DocFileErrorKind` types), the runner asserts presence and smoke-calls `docFileRawURL("s", "p")`. RED: `Module '"@evener/appwire-client"' has no exported member 'docFileRawURL'`. Noted while doing it: `tsconfig.build.json`'s `files` list is documentation, not a packing boundary — tsc emits whatever `index.ts` reaches — so the list is kept complete but `index.ts` is what decides what ships.


obra:
Round 1 disposition (head `a04f1d97d`, main merged): the Low is right — only `readDocFile` touches `fetch`, at call time — so `a50c06875` ships `docContent` too: its URL helpers, size constant, error type and content types are exported, the runner asserts presence and smoke-calls `docFileRawURL` (never `readDocFile`). All sixteen modules now ship. Noted in passing: `tsconfig.build.json`'s `files` list is documentation rather than a packing boundary (tsc emits whatever `index.ts` reaches), so the list is kept complete but `index.ts` decides what ships.

obra:
Round 2 disposition (head `68a1da7c2`, main merged): the Low is right; `73387e330` enforces the invariant from the built artifact rather than a manifest — the runner reads `dist/index.d.ts` out of the installed package and requires a `from "./<module>"` re-export for every shipped module but `index`, so a module that is packed but unreachable fails the gate (RED with a fake `notReExported` entry: `shipped module unreachable from the entry point: notReExported`). The `runtimeExports` comment now claims only what that list proves.

obra:
Round 3 disposition (head `038851d7f`): the Medium is right for the entry point as it stands; `038851d7f` drops `readDocFile` from `index.ts` and the runner's export list until the plan's C24 gives it an injectable base URL and fetch. The pure helpers, constant and types stay exported, so `docContent` remains reachable and the reachability gate still covers it.

PR #1189, opened 2026-09-12
refactor(protocol): one settled-item failure predicate for both clients

### LINKED ISSUE #1190: web: toolRenderers.ts carries a third, divergent tool-failure predicate
Found during SDK migration PR B1 (plan #1182), which collapsed the web projector's and the native projection's settled-item failure predicates into `protocol/itemFailure.ts`.

A third predicate, also named `toolCallFailed`, lives at `cmd/evener-hub/frontend/src/panes/session/transcript/toolRenderers.ts:189`. It is not the same rule: it does not trim `error`, does not treat `interrupted` as a failure, ignores `exitCode`, and adds a tool-descriptor hook. It appears in neither the inventory nor the plan.

Decide whether the renderer is meant to classify differently from the projector (in which case the difference should be named and tested) or whether it should call `hasItemFailure` with the descriptor hook layered on top. Either way it becomes an inventory row. Related: #1182.

### LINKED ISSUE #1197: web: ToolCallItem shows blank error text for a whitespace-only error
Residual from SDK migration PR #1189 (plan #1182).

`cmd/evener-hub/frontend/src/panes/session/transcript/ToolCallItem.tsx:211` `hasErrorText` tests `item.error !== ""` without trimming, while the row's failure state now comes from the shared `hasItemFailure`, which trims. A tool call whose `error` is whitespace only therefore renders an empty error block on a row that is not marked failed. Use the same trimmed test (or a shared `hasErrorText` next to `hasItemFailure` in the package) so the two agree. Cosmetic; one line plus a renderer test.

obra:
PR B1 of the SDK migration plan (#1182). Stacked on A1 (#1184, `30b8c585e`); this PR's own change is `f00c5a466` and the diff contracts to it once #1184 lands.

The web transcript projector (`transcriptDisplay/projector.ts`) and the native conversation projection (`mobile/src/conversation/project.ts`) each carried their own "did this settled item fail / is it still running" predicate over the same wire fields. They were identical in substance — non-blank trimmed `error`, status `failed`/`interrupted`, defined nonzero `exitCode`, in the same order — and `ItemModel` and the wire `ThreadItem` declare those three fields with the same types, so one structural input serves both.

`protocol/itemFailure.ts` now owns `isNonZeroExit`, `hasFailureStatus`, `hasItemFailure` and `isInProgressStatus`; both projections call it and their local copies are deleted. The module ships in the package and is smoke-called by the qualification runner. `projector.test.ts` and `project.test.ts` are byte-unchanged and green, which is the contract; `itemFailure.test.ts` (17 cases) covers each branch directly. The projector's two `"inProgress"` literals now call `isInProgressStatus` so the shipped module has two real consumers.

RED: `Failed to resolve import ./itemFailure`. GREEN: 17 passed. Gates: `make test-web`, `make test-native`, `make test-api-package`, `make test-web-browser` (all five guards).

## Review round 1

Medium, the web transcript's own `toolCallFailed` (`panes/session/transcript/toolRenderers.ts:189`) still classified on its own — accepted; it is the third predicate filed as #1190, fixed here in `d1541bd5b`: `hasItemFailure(item) || (descriptor.failed?.(item) ?? false)`, descriptor hook untouched. Nothing in the repo pinned the old behavior. RED at `f00c5a466`: three renderer tests (interrupted, whitespace-only error, nonzero exit) fail; GREEN after.

**User-visible change on the web transcript:** an interrupted tool call now renders with the failure state, a nonzero exit code now marks any tool's row failed rather than only tools whose descriptor declares a `failed()` hook, and a tool call whose error is whitespace only now renders *without* the failure state (the old registry predicate compared against `""` untrimmed and marked it failed; the shared predicate trims, as the projector and native already did). Closes #1190.

Gates after the merge of main: `make test-web`, `make test-web-browser` (five guards), `make test-api-package`.

## Review round 2

Low, `hasErrorText` untrimmed while the row is classified with the trimmed shared predicate (#1197) — accepted; `6dfde0b22` adds `hasErrorText` to `protocol/itemFailure.ts` (shipped, smoke-called, tested) and `ToolCallItem.tsx` uses it. Closes #1197.


obra:
Round 1 disposition (head `5d62e86d9`, main merged): the Medium is the third predicate already filed as #1190, fixed here in `d1541bd5b` — `toolCallFailed` is now the shared `hasItemFailure` OR the descriptor's own hook. Three renderer tests RED at `f00c5a466`. One correction to the finding's framing: a whitespace-only error moves from failed to *not* failed (the shared predicate trims, as the projector and native already did); interrupted calls and nonzero exits move to failed. The change is stated in the PR body. Residual filed as #1197 (`hasErrorText` still untrimmed).

obra:
Round 2 disposition (head `a5639624b`, main merged): the Low is #1197, fixed here in `6dfde0b22` — `hasErrorText(item)` lives in `protocol/itemFailure.ts` beside `hasItemFailure` (which now calls it, so "non-blank error" is written once), shipped and smoke-called by the runner, and `ToolCallItem.tsx` reads it instead of comparing against `""`. RED at `5d62e86d9`: a whitespace-only error rendered an empty error block. Closes #1197 on merge.

obra:
Round 3 disposition (head `1bc03c716`): the Low is against the A1 base this branch was stacked on, not this PR's own change — `docContent` was held back at `30b8c585e`, and #1184 has since shipped it (`a50c06875`), kept `readDocFile` off the entry point until the C24 port (`038851d7f`), and replaced the `heldBackModules` rationale with a reachability check from `dist/index.d.ts` (`73387e330`). This head merges those rounds in, so the stale rationale is gone. Package gate green locally; no change to this PR's own commits.

obra:
Refreshed onto main `f2599d1ed` (head `b4e35df0d`): #1184 landed as a squash, so the stack merge conflicted textually in `index.ts`, `qualify-package.mjs` and `tsconfig.build.json`; resolved by taking main's versions and re-applying this PR's own hunks. The diff against main is now exactly B1's change (10 files, +222/-53). Package gate green locally; requalifying.

obra:
Round 4 disposition (head `1766baf29`, main `9db3a1e99` merged): both real. `6b42d66d2` makes `foldable` bail on `hasItemFailure(item)` — not also `hasErrorText`, because the fold rule is about failure ("anything that is not a plain, completed, unremarkable call breaks the run") and under the shared predicate a non-blank error is always a failure, so the extra clause could never fire on its own. RED at `b4e35df0d`: a nonzero-exit call folded into a run; a whitespace-only error blocked a fold. No `toolRuns.test.ts` fixture pinned the old behavior. The `ToolCallItem.tsx` comment now states the split: generic covers error text, status and exit code; the descriptor covers failures visible only in the output's shape.

PR #1196, opened 2026-09-12
test(web): stop a released IndexedDB hold from swallowing in-flight events

### LINKED ISSUE #1187: web: Composer.integration.test.tsx cascades 33 failures after one 5s timeout under CI load
Seen on PR #1184 (SDK migration A1, which changes only `protocol/` exports and the package qualification runner; the web app imports nothing from the protocol barrel, so the change cannot reach this file at load time).

CI run https://github.com/prime-radiant-inc/evener/actions/runs/34718849206, job `web`, step "Test, build, and browser guards": `src/panes/session/composer/Composer.integration.test.tsx` 46 tests, 33 failed. The first failure is a timeout — `a second message queues behind a committed start even when recovery projection reads stall` at `:635`, `Test timed out in 5000ms` — and every later failure in the file is a consequence of it (`Unable to fire a "change" event - please provide a DOM element`, `Cannot read properties of null (reading 'disabled')`, `expected undefined to be 'recover me'`), i.e. the timed-out test leaves the shared DOM/fixture in a state the rest of the file cannot recover from.

Evidence it is timing, not the change: the same file passes 5/5 locally on the PR head 30b8c585e (46/46 each run), and main's last eight `web` jobs are green.

Two things to fix, per `docs/developing-evener/testing.md` (a sighted flake gets a root cause): (1) the stall in the recovery-projection scenario should await a real completion with a tripwire ceiling rather than race a 5s default under load — find what it waits on and expose the seam; (2) the file's tests should not share state that a single timeout can poison — each test should own its mount and tear it down in `afterEach` so a failure stays one failure.

obra:
Closes #1187. Two defects, neither in `Composer.tsx`, found by reproducing the CI cascade locally (48 CPU hogs on 16 cores: 1 failure in 20, then 1 in 26, identical to CI's 33-of-46 with the same first failure).

**Root cause** — `stores/testing/stalledIndexedDB.ts`: `holdIndexedDBEvent` wraps every listener registered while the hold is up. `release()` restored the spy and drained the events queued so far, but the wrapper on already-registered listeners lived forever, so an event still in flight at release time landed in a queue nobody drained again: an IndexedDB request whose promise never settles. In `Composer.integration.test.tsx:635` the swallowed read is a pending-turns projection refresh started by the second submission, and the test's `finally` (`:681`) then waits on `flushPendingTurnsProjectionForTests()` forever. Fixed in `1a62a0987`: `release()` now means "stop holding", so an event arriving after it goes straight to its listener. New `stalledIndexedDB.test.ts` pins it (RED first; its third case reproduces #1187 in one line: a read released before its success event, 5s timeout).

**Cascade** — `panes/session/composer/queue/pendingTurnsStore.ts:139`: that flush waits inside `act()`; when vitest abandons the test at 5s, the open act queue makes every later `render()` in the file produce nothing, which is exactly the `please provide a DOM element` / `reading 'disabled'` shape. Proven with a throwaway probe (an abandoned `act()` poisons later renders; abandoned `waitFor`, `user.type` and plain `await` do not). Fixed in `fb86708c1`: `settlePendingTurnsProjectionForTests` gains a per-round stall tripwire (4s, against a measured slowest round of 165ms across the whole web suite under 32-way contention), so a stalled read fails its own test with `pending-turns projection work stalled: N operation(s) still unsettled after 4000ms` and the rest of the file runs. `Composer.integration.test.tsx` is untouched: no assertion weakened, no timeout widened.

Evidence: before 19/20 and 25/26; after 40/40 under the same contention. Cascade check with a deliberately never-released read: 1 failed / 46 passed instead of 1 timeout plus 33 consequential failures. Gates: `make test-web` (typecheck, tests, biome). `holdIndexedDBEvent` has six consumers; all pass.

### DIFF
--- a/cmd/evener-hub/frontend/src/panes/session/composer/queue/pendingTurnsStore.test.ts
+++ b/cmd/evener-hub/frontend/src/panes/session/composer/queue/pendingTurnsStore.test.ts
@@ -287,6 +287,46 @@ test("a flush cannot settle while a submit is still in flight", async () => {
   expect(flushResolved).toBe(true);
 });
 
+// The flush above waits inside act(), so durable work with no completion left
+// parks the caller there until vitest abandons the whole test - and an
+// abandoned act() leaves React's act queue open for the rest of the FILE, so
+// every later render produces nothing and one stall becomes dozens of
+// unrelated failures (issue #1187). The per-round tripwire is what turns that
+// back into one named failure, in the test that caused it; without this test a
+// regression that clears the timer or swallows its rejection restores the hang
+// silently, and the only symptom is a file that fails 33 ways again.
+test("a flush that can never settle trips instead of hanging inside act", async () => {
+  let releaseSubmit: () => void = () => undefined;
+  const stalled = new Promise<void>((resolve) => {
+    releaseSubmit = resolve;
+  });
+  const submitted = submitWithPendingTracking(
+    { ref: "ref_a", method: "send", text: "never settles", onFailure: () => undefined },
+    () => stalled,
+  );
+
+  vi.useFakeTimers({ toFake: ["setTimeout", "clearTimeout"] });
+  try {
+    // Captured before the clock moves: the rejection lands while the timers
+    // advance, and a handler attached only afterwards is an unhandled
+    // rejection in that window.
+    const flushing = flushPendingTurnsProjectionForTests().then(
+      () => undefined,
+      (error: unknown) => error,
+    );
+    await vi.advanceTimersByTimeAsync(60_000);
+    expect(await flushing).toMatchObject({
+      message: expect.stringMatching(/projection work stalled: 1 operation\(s\) still unsettled after \d+ms/),
+    });
+  } finally {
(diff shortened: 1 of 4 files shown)


obra:
Round 1 disposition (head `6c5432e71`, main merged): the Low is right; `46e47d8c4` pins the tripwire — a `perform` that never resolves, fake timers advanced past the ceiling, the flush rejects with `projection work stalled: 1 operation(s) still unsettled after <n>ms`. RED by dropping the guard back to a bare `Promise.allSettled` (the new test then times out at 5000ms, the exact failure it prevents). The rejection is captured before the clock moves so no unhandled rejection dirties the output.

obra:
Round 2 disposition (head `3b2ac1f32`, main `f2599d1ed` merged): the Low is right; `193fdc282` fires `observed?.()` on both delivery paths before the listener runs, so `reached` answers "has the held event arrived" regardless of release. RED: a new case (event arriving after release) timed out at 5000ms before the fix; 4/4 after.

PR #1207, opened 2026-09-13
test(protocol): qualify the package per specifier, not only at the root

obra:
PR C0 of the SDK migration plan (#1182), the one piece of plumbing every later subpath needs.

The qualification runner's manifest was a flat array of root export names, so it could express "the root exports X" and nothing else; a subpath such as the plan's `state/navigation` would ship unqualified while the gate stayed green. The manifest is now `packageExports`, a map from published specifier to `{ values, types, smoke }`, and the runner emits per specifier one ESM declaration consumer, one CommonJS declaration consumer and one runtime program each for ESM and CJS. Reachability is per specifier too: each specifier's `.d.ts` target is resolved from `package.json` `exports`, and a shipped module must be some specifier's entry or a re-export from one. Two cross-checks make the rule explicit: a specifier in `exports` with no manifest entry fails (`published specifier is not qualified: no manifest entry for ./x`), and a manifest specifier absent from `exports` fails (`qualification manifest names ./x, which package.json does not export`). The in-repo-only `testing` alias is unlistable by construction and stays validated by the apps' typecheck. The root outcome is unchanged: every current name is still checked and the tarball listing still asserted.

Proven with a throwaway subpath (reverted): `./docContent` published and removed from the root, so reachable only through its own specifier — the gate went green with eight generated programs importing from `@evener/appwire-client/docContent` (including `readDocFile`, which the root deliberately does not export), a bogus name failed both declaration consumers, and dropping the subpath while the module stayed out of `index.ts` failed with `shipped module unreachable from every published specifier: docContent`.

RED in three steps before the shape existed (tsc dump, then each cross-check direction). Gates: `make test-api-package`, `make test-web`, `make test-native`; biome. 2 files, +110/−65.

### DIFF
--- a/cmd/evener-hub/frontend/src/protocol/README.md
+++ b/cmd/evener-hub/frontend/src/protocol/README.md
@@ -11,14 +11,18 @@ the send/queue availability table, the stable delegate status rule, and the
 doc-pane URL builders.
 
 Build and qualify from this directory with `npm run qualification`. The runner
-packs the package, installs that tarball into a temporary consumer, checks ESM
-and CommonJS TypeScript resolution, runs both runtime import forms, calls at
-least one export of every shipped module on a trivial input, and checks the
-tarball contains each shipped module and no source files or dependencies. It also runs the
-installed inspection and discovery examples against a scripted local WebSocket
-server, verifying the handshake, read-only requests, structured readback, and
-private output file. The qualification command is run with the repository's
-configured Node 22 runtime.
+packs the package, installs that tarball into a temporary consumer, and then,
+for every specifier the `exports` map publishes, checks ESM and CommonJS
+TypeScript resolution and runs both runtime import forms against the names that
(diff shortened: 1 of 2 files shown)


obra:
Round 1 disposition (head `13d75fe7b`, main `99fa1882f` current): the Medium is right; `13d75fe7b` derives each specifier's consumer filenames from `encodeURIComponent` of the specifier (root stays `root`, subpaths get a `sub-` prefix so a hypothetical `./root` stays distinct) and asserts filename uniqueness at generation. RED with `./foo-bar` and `./foo/bar` published in a throwaway run: one filename for two specifiers, the second overwrote the first, and the runner still reported qualified; after, six distinct consumers that all compile and run, and restoring the old slugging trips the new assertion. Throwaway entries removed; only the runner changed (+18/−7). Gates: `make test-api-package`, `make test-web`, `make test-native`; biome.

PR #1208, opened 2026-09-13
feat(web): show a session's live watches in the rail and the activity panel

obra:
## What this does

A session's live **watches** (armed subscriptions that deliver fires into the session) are now visible in the web UI:

- **Sidebar rail**: a neutral count on the session's row, placed before the branch so ellipsis can never eat it.
- **Activity panel**: a `Watches` group above the running jobs, one row per watch, expanding to the note as prose when the row title could not show all of it, **one facts sentence**, and a **delivery timeline drawn only from real delivery instants**.

The panel never claims anything the runtime does not know: no drop count (per-delivery drops are not tracked), no exact countdown (a clock-driven cadence shows its approximate next fire, worded `~` because the runtime's ticker can slide), no fabricated timestamps. A condition watch (on a job's output, or on events) has no schedule to draw, and says so in one line instead of drawing a fake one.

## Screenshots, from the running app

Activity panel, both watch kinds expanded:

Activity panel

Rail row carrying the count:

Rail row

Whole window:

Full window

## How it works

    agent.Session.DetailedStatus -> appwire.EvenerDiagnostics -> hub navigation summary -> frontend

Additive throughout: the wire carries watches and a bounded ring of the **last 32 delivery instants**. Absence means an old daemon, and every consumer treats a missing list as empty. No protocol version bump (that constant is a handshake value, not a schema hash).

## Verification

- `go test -short -count=1` across `agent`, `appwire`, `server`, `internal/appwirets`, `hubapi`, `cmd/evener-hub`: every package ok.
- `make test-web` (typecheck, unit, Biome): PASS.
- **A real end-to-end run**: an isolated hub built from this branch, a real session, and real watches armed through the UI (a progress watch on a running job, an output watch, a repeating self watch). The screenshots above are that run. The delivery ring, the capped caption (`Last 32 of 55 deliveries`) and the no-schedule line were all checked against live wire data rather than fixtures.
- **A five-reviewer vision panel** on the running UI (craft, triage, adversarial, honesty, then a confirming pass). The honesty reviewer transcribed every rendered string and checked it against the code that composes it: zero divergences, zero overstatements. Reports are under `docs/web-ui/sidebar-watches/reviews/`.

## What the live run caught

`stores/navigation/codec.ts` validates every session entity against an explicit field allowlist, and `watches` was not in it. Every frame carrying a watch therefore failed strict validation, and the store silently kept the previous snapshot: the data was on the wire and nothing rendered. **1759 unit tests passed while this was broken**; it was obvious within seconds of running the app. Fixed in `abfec3779d` with a validator for the watch shape and a regression test through the real decode -> materialize -> rail model -> panel rows chain.

## Panel findings, fixed

- The note was printed twice: as the row title, then again as the detail's first line.
- The newest delivery dot landed under the taller "now" marker and vanished into it.
- One count had two nouns: "fires" in the row, "deliveries" in its own facts line.

Two findings were rejected with evidence: an "orphaned output watch" cannot exist (`expireJobWatchesLocked` clears every watch whose target job reaches a terminal state, confirmed live when the demo job ended and its watch disappeared), and the red failure text on the inactive fold is pre-existing vocabulary outside the watch block.

## Known follow-ups (from the simplify pass)

Named rather than silently dropped:

1. **The 1Hz `now` prop threaded into the activity tree defeats its memoization**, so the tree's rows re-render once a second while the panel is open. This is a regression introduced by this branch. The fix reworks the tick plumbing (context-only ticking, or a memoized rows block) and touches three tests' time control, so it deserves its own change.
2. **The daemon formats and sorts up to 32 delivery instants per watch while holding `jm.mu` on every status read.** The fix is to collect under the lock and format outside it.
3. Two smaller frontend memoization misses (`watchDeliveryInstants` reparsed per tick, `WatchRowView` subscribing to the tick at row level), and two altitude notes (the panel's pure row model imports display helpers from the rail's React module; `WatchRowView` duplicates `DenseRowView`'s chrome).

## Deliberately not done

- **No next firing for a condition watch.** An output or event watch has no schedule, so a countdown there would be invented; those rows show cadence and armed state instead. A clock-driven cadence (one-shot timer, repeating timer, progress interval) does show its approximate next fire, worded with `~`.

## Update: what the review passes changed after this description was written

This description is the original design record. Later review passes added and corrected:

- **A descendant's watches carry to its own row** (a subagent session's watches reached no row at all), including when they are cleared.
- **The row list is bounded and counted**: a session summary carries at most 32 watch rows, `omitted_watches` and `omitted_armed_watches` say exactly how many were dropped and how many of those were armed, and an oversized response degrades by stripping delivery instants first and shedding whole rows second, so a watch-heavy session cannot take navigation offline.
- **The daemon bounds the live set at the source**: `maxLiveWatches = 32` per job manager, refused at registration with `invalid_request: too many watches (32 live); clear one first`. Only growth is refused, so an identical re-registration, a replacement, and a clear all still work at the cap. The cap bounds one manager's set, not a session's projected row -- that row is a rollup over the source sessions holding receiver watches for it, which is the design: a receiver watch belongs on its receiver's row.
- **The thread list samples a page's watches in one walk** (`LiveWatchRowsForSessions`) instead of walking the live tree once per row.
- **Wire and codec hardening**: the projector drops any watch row the hub schema would reject and counts it, identity fields are byte-bounded, `every` is clamped to the navigation-safe range, the schema enforces `armed <= omitted`, and the web codec now enforces the same invariant.

**Current state:** 42 commits, rebased onto `main` (the last rebase folded in the SDK import migration that moved the appwire imports onto the package root). The full frontend suite (476 files, 10,967 tests), the typecheck and Biome gates, and the Go build and test stages are green, and every finding from the eight review passes is answered in the comments below.


obra:
Addressed the combined review. Every finding reproduced against the code; seven are fixed here with regression tests, and one is deliberately partial.

**Fixed**

- **M1 (frontend half).** The panel no longer collapses a multi-trigger watch to a single kind. A watch with any clock cadence now draws its timeline and never receives the "there is no schedule to draw here" line, and the facts sentence names the conditions that are actually configured. The wire half — carrying the event-count cadence and the event filter so a fire-every-Nth-event watch is distinguishable — is deferred: it touches the daemon projection, the appwire type, the hub summary, the hub projection, the server projection and the generated types. That is a coherent change of its own and the next thing I would pick up. It is not needed for the panel to be truthful, because the panel now claims only what the wire actually carries.
- **M2.** `liveWatchStatuses` now aggregates the session's own job manager plus the stable watch-source sessions, deduping managers, so a receiver watch on a descendant's job reaches the receiver's summary instead of being invisible in both. The descendant-receiver shape had no test; it has one now.
- **M3.** The "+N more watches" overflow row no longer looks like a button that does nothing.
- **L1.** Timeline dots are keyed by instant *and* occurrence, so two deliveries in the same millisecond no longer collide.
- **L2.** Timestamps are validated, never truncated. An instant the codec cannot decode is dropped; a watch whose required `created_at` cannot be represented is dropped as a row rather than poisoning the whole resource. The Go predicate is pinned to the codec's own fixture (`testdata/navigation/timestamps.json`, which both sides already read) so the two grammars cannot drift apart again.
- **L3.** The standalone Activity pane selects only its own session's watches rather than subscribing to the whole navigation store.
- **L4.** `rosterFingerprint` hashes the delivery instants.
- **L5.** `watchDurationLabel` rounds once to a whole second and floors from there, so `1m60s` is no longer reachable.

**One existing test changed, explicitly**

`TestNavigationWatchDeliveryTimesBoundedLikeOtherWatchStrings` asserted the truncation this review identified as the bug: an ellipsized instant is not RFC 3339, so the entire watch-carrying snapshot failed to decode. It is replaced by `TestNavigationWatchDeliveryTimesDropsUnrepresentableInstants`, plus a new `TestNavigationWatchProjectionDropsWatchWithUnrepresentableCreatedAt`. The old expectation encoded the defect; no other assertion was touched.

**Gates:** `gofmt` clean, `go build ./...`, `go test -short -count=1` across agent/appwire/server/appwirets/hubapi/cmd-evener-hub, `make test-web`, `make test-native` — all green.

obra:
Second pass on the updated combined review (`1f5850d`), which re-ran with a different set of findings than the first. Two of its items were not in my previous pass; both are now fixed.

- **The standalone Activity pane and watch-only sessions had no ticking clock.** A session whose only work is a watch did not enable the tree's tick, so armed ages, the "now" label and the timeline's end marker froze at whatever they read when the panel mounted. The tick now also runs when the session has watch rows, so the clock comes from the tree's own context and the standalone pane (which passes no clock of its own) stays live. Regression test: "a session whose only work is watches still runs the clock" — it fails without the change and passes with it.
- **The no-schedule line was inaccurate for event watches.** It claimed the watch fires "when the job says the word", which is wrong for a watch that fires on session events. It now reads: "There is no schedule to draw here — this one fires when the job or event it watches says so, not when a clock says so."

The updated review's other items were already fixed in the previous push: the duration carry (`59m60s`, `1h60m`, `1d24h`), the duplicate timeline keys, and the fingerprint test for the watch contribution to `rosterFingerprint` (which the review explicitly asked for and which now exists as `TestRosterFingerprintIncludesWatchDeliveryTimes`).

Still outstanding, deliberately: carrying the event-count cadence and event filter on the wire, so a fire-every-Nth-event watch is distinguishable from one that fires on every matching event. That is the one item I am leaving as a follow-up, and it is described in the previous comment.

obra:
Third pass, on this head's review.

**Fixed**

- **Clone aliasing.** `cloneNavigationLiveEntries` now deep-copies the watch list with `appwire.CloneEvenerWatches`, so the hub's clone no longer shares the roster's watch backing arrays while its sibling helper copies them. `TestCloneNavigationLiveEntriesOwnsWatches` fails without the line.
- **Selector cache growth.** `sessionWatchesCache` now deletes a ref's entry when its session summary is gone, and is capped at 256 entries with oldest-insertion eviction, so a long-lived page cannot accumulate an entry per session it has ever seen. Eviction costs a recomputation and never correctness, because the content key still decides what is returned. Two tests cover the drop and the bound.

**Not fixed, deliberately**

The medium about unbounded watch and event payloads reaching the response fitter. The interaction is real: the fitter drops whole session rows, and a row that cannot fit can make it reject the resource. Reaching that needs a single session row of about 2 MiB, which means thousands of armed watches each carrying a full 32-instant ring, and I judge that implausible. The honest fix is a bounded projection plus an omitted-count field on the wire; that is the same work as the deferred event-count and event-filter carry, so it belongs with it rather than as a silent cap on a surface whose whole point is not to hide things.

obra:
Fourth pass, on this head's review. Four of the six were small and are fixed; the other two are deliberately deferred and explained below.

**Fixed**

- **Combined triggers in the collapsed row.** `watchMeta` branched on a single `watchKind`, so a watch carrying both an `output_match` and an event trigger rendered only "on output" while `watchFacts` and the rail named both. The conditions are now derived independently from the wire fields. New test: "a watch carrying both an output match and an event trigger names both conditions" (fails before: `on output · armed` vs expected `on output · on event · armed`).
- **The `watchText` import sat mid-file** in `RailRow.tsx` after component code; it is back in the top import block, comment intact.
- **Shallow watch copy in the daemon's serve path.** `agentToServerDetailedStatus` copied only the outer slice, leaving `Cadence`, `Events` and `DeliveryTimes` shared with the source snapshot. New `cloneServeWatches` mirrors `appwire.CloneEvenerWatches`, with `TestAgentToServerDetailedStatus_WatchesDeepCopied`. That is the fourth aliasing instance review has found in this branch, all now closed.
- **Local `thread/list` dropped watches entirely.** `LiveEntry.Watches` now carries through `LocalDaemonEntry` with a deep copy, so local AppWire clients see the same diagnostics navigation does. `TestLocalDaemonSourceListCarriesWatches`.

**Deferred, with reasons**

- **Invalidating the diagnostics envelope on every watch mutation.** Real, but not a one-line fix: a progress watch delivers every second, so invalidating on each delivery would thrash a cache that exists to avoid recomputing the projection, and the right shape (invalidate on registration and on transitions, coalesce deliveries) needs measuring rather than guessing. Worth its own change.
- **Bounding projected watch rows against the response fitter.** Unchanged from the previous comment: the fix is a bounded projection plus an explicit omitted count, which is the same wire work as carrying the event-count and event filter, and belongs with it.

At this point the review has no critical or high findings, and the remaining items are either fixed here or named above. Further automated passes will keep producing fresh low-severity nits; I think the right next step is a human read rather than another round.

obra:
Went through the remaining findings. Two fixed, one refuted with evidence.

**1. Event-count and event filter now cross the wire.** `watchCadencesOf` dropped a concrete event watch's every-Nth throttle and its filter while the model-facing prose summary rendered both, so a fire-every-3rd-event watch looked identical to one firing on every match. The events cadence row now carries `every` and `filter` (the filter in the prose summary's own vocabulary, so the wire and the model cannot disagree about what it names), through the daemon, appwire, `hubapi`, both projections and the generated types, and the label renders "on events every 3" and "on events where tool_name=Bash, status=error". A wildcard watch carries neither, exactly as the summary does. Plain event watches still read exactly "on events", which is also the old-server case. Four tests, one per layer, each proven to fail before the carry.

**2. An oversized row no longer takes navigation offline.** For the record, what the fitter did: `limitNavigationSummaries` drops whole session rows, `navigationFittingBudget` can return zero, and `validateNavigationPageProgress` rejects a page with no rows and remaining data, which production turns into a rejected resource. It now tries the complete payload first, preserving the previous answer and probe count, and only when even one untrimmed row cannot fit does it degrade the optional watch payload, delivery instants first and whole watch rows second, so the session stays listed with as much of its payload as fits. If no trim level retains a row the overflow is not in the watch payload and the original behaviour stands. Test with a synthetic oversized row, proven to fail before.

The one honest gap that remains: when whole watch rows are trimmed there is no omitted-count field, so in that extreme case the sidebar count would undercount silently. That is the "explicit omitted count" half of the same finding, it needs a wire field and a rendering decision, and I have left it rather than invent a number on the wire. Note this only triggers where navigation previously went offline, so it is a strictly better failure.

**3. The stale-diagnostics claim is refuted for the ordinary paths, and I did not change anything.** Every watch delivery wakes an EntryNotification turn, and that turn's TURN_ENDED re-samples the envelope including the watch rows, so registration, clearing, delivery and timer paths do refresh. What survives is a narrow residual: an automatic post-turn teardown and a sender-side settle can leave the last state unrefreshed until the next event, which is a bounded staleness, not the reported "idle sessions show stale watches". The trace and the residual are covered by a test asserting the refresh behaviour rather than a speculative invalidation, and I would rather leave that residual visible than paper over it with an invalidation that would thrash the cache on every one-second progress delivery.

obra:
**1. The live-watch set is bounded at registration now (`0067cc1e8`).** The finding is right that the bound took effect too late: the daemon projected every entry in `jm.watches` onto AppWire, and the hub cloned, fingerprinted, and only then capped the rows it displays at 32, so a session holding a pathological number of live watches put an unbounded watch payload on every roster probe.

The fix bounds the resource where it is created, which is the finding's first option: a `create` that would add a **new** key to a manager already holding `maxLiveWatches` (32) is refused with `invalid_request: too many watches (32 live); clear one first`, mirroring the timer refusal that already sits four lines above it in the same function. Only growth is refused — an identical re-registration, a replacement of a held key, and a clear all still succeed at the cap, and a cleared slot admits a fresh key. Four tests, and the two refusals were proven to fail before the guard existed.

32 is not a new number: it is what the hub's display bound and the per-watch delivery ring already use, so one bound now covers the daemon's per-watch footprint, the projection, and the wire. I chose registration over carrying omission counts on the appwire wire because a trimmed wire would leave live-but-invisible watches: the daemon would keep scheduling and delivering them while the sidebar could not list them, and a hub reading a trimmed list would lose the omitted count entirely in the mixed-version window.

**CI note:** the `web` job failed once on `8bcf9007b` in `TestSkillComposerBrowser` (`skillguard FAIL: typeText: caret at 0 after the edit, expected 50`). That test drives the skill composer through the real hub web app in Chrome and touches nothing this branch changed; it passes locally on that exact head (`go test -tags browserguard -run TestSkillComposerBrowser ./cmd/evener-hub/`), and main's web job failed twice in the same window. Reading it as a browser-timing flake rather than a regression from this branch.

obra:
**1. The cap now states exactly what it bounds (Medium: aggregate payload).** The finding is right, and the first version of the comment overclaimed. What `maxLiveWatches` bounds is one manager's live set -- what the daemon stores, projects and ships for that one session -- and it refuses a key that would grow that set rather than trimming it. What it deliberately does *not* bound is a session's projected **row**: that row is a rollup of the session's own manager's rows plus the receiver watches it owns on descendants, each of those managers capped by this same 32. That is the plan's own rule -- "a subtree rollup must count each watch once: a receiver watch counts for its receiver, not its owner" -- so the row grows with the number of source sessions holding watches for the receiver. The comment now states both, so no reader has to re-derive which payload the number does and does not cover.

I did not add a second bound at the projection. Trimming there would leave rows the agent cannot see while the daemon keeps scheduling and delivering them, and a hub reading a trimmed list cannot report the omitted count: the counts the sidebar renders are computed by the hub from the rows it received, so an upstream trim would silently undercount in the mixed-version window.

**2. The list path now samples the page once (Medium: quadratic traversal).** Resolving one row at a time walked the live tree once per row, so a page of many live sessions paid that walk per session. `LiveWatchRowsForSessions` answers a whole page from a single walk; `attachLiveWatches` makes one resolver call per list request and looks each row up in that answer. The single-ID resolvers (`LiveWatchesForSession`, `LiveWatchesForDescendant`) now answer from the same walk, so a single-ID caller and a page caller cannot disagree about which sessions are answerable, and their contract tests -- the nil-versus-non-nil-empty that decides whether a row is refreshed or left alone -- still pass unchanged. New tests: the agent page contract (root, a descendant with rows, a descendant with none, unknown ID, empty ID, no IDs), and a server test pinning exactly one resolver call for a two-row page with both rows fed by that one sample.

**3. The web codec rejects an impossible armed count (Low).** `omitted_armed_watches > omitted_watches` is now rejected, mirroring `navigationSessionValueValid` exactly, including the absent-total-is-zero case: no producer can send an armed count without a total (the projector counts every omitted armed row in the total too), so treating an absent total as unknown would only admit malformed snapshots. The test covers rejected, equal, less-than, and both absent shapes.

obra:
**1. The page projection is one pass now (Medium).** You were right that the previous shape still resolved each requested session on its own: every call rebuilt the source-session list and locked and scanned each manager again, so a page of N sessions against N managers paid up to N x N of that. `LiveWatchRowsForSessions` now takes **one lock per manager for the whole page**: each manager appends its watches once, keyed by the row each one belongs on (its receiver's, or its own session's when it has none), and each requested ID reads its rows back out of that. The tree walk stays one walk, the answerable set and the nil-versus-non-nil-empty contract are unchanged, and the single-ID resolvers answer from the same pass, so the two forms cannot drift.

Measured on a 49-row page holding 196 watches (`agent/watch_page_rows_bench_test.go`, `-benchtime=200x -benchmem`): **0.41 ms -> 0.26 ms** per page call. Allocations are comparable rather than lower: the page now materializes every row's slice in one pass instead of re-deriving them per row, which is the trade this makes.

**2. A spent watch says so in its compact line (Low).** `watchMeta` printed only the delivery count whenever a watch had one, so a one-shot whose teardown is pending (`active: false`, one delivery) read exactly like a watch still waiting once the note was collapsed. It now keeps the count and appends the armed state, using the same wording the rail's gloss and the panel's facts line already use (`watchText.ts`), so the three cannot disagree. The test pins `after 10m · 1 delivery · not armed`.

**3. A failed or unsupported activity load no longer hides the watches (Low).** The watch group is part of the tree's row list, so the fallback renders the same `ActivityTree` with an **empty page**: one renderer and one vocabulary rather than a second watch view that could drift from the real one. The rows now appear above "Activity isn't available" and above the failure state, which is what the rail's count already says. The test drives a failing `evener/jobs/list` with one armed watch and asserts the row and its facts line render.

obra:
Round 9 answered. Six fixed, one refuted; the gate set is running on the result.

**1. The receiver rollup is off the event bridge now (High).** Confirmed, and it is the sampling contract's own rule: `agent/session_envelope_sampling.go` states that a sampled method may take this session's `mu` and nothing else, because the event bridge calls it and a lock held across an emit would block event consumption. `DetailedStatus` was aggregating descendant managers through `liveWatchStatuses`, so a sampled facet reached into other sessions' job-manager locks. It now carries this manager's own rows, and the rollup stays where it already ran outside the bridge -- the thread LIST path (`LiveWatchRowsForSessions`), merged onto each row. A new test pins the split: the same tree answers with one row through `DetailedStatus` and with the receiver row through the page resolver.

**2. The one-shot countdown is refuted, and the real half is fixed (Medium).** The `deliveryTimes` disjunct is not "any delivery suppresses the countdown": a one-shot fires once, whatever fires it, so a delivery of ANY kind means it has already fired. Dropping that disjunct would give a condition-fired one-shot a phantom countdown. What WAS wrong is its neighbour: the armed state read only `firedPendingEnd`, so a one-shot in its settle window reported **armed with no next fire**. Both now take their answer from one `watchOneShotHasFired` helper (teardown marker, clock fire, or a recorded delivery), and a test pins that a condition-fired one-shot is spent while a waiting one is armed and dated.

**3. A repeating cadence never advances from an older clock fire (Low).** Fixed: the base is the later of the install instant and the newest clock fire, so a clock that steps backwards cannot push the derived fire into the past. Test added.

**4. Over-limit watch identities are dropped, not cut (Medium).** Confirmed -- the rail and the panel key their rows by `watch.id`, so cutting two distinct long ids to the bound could map them onto one key. Such rows are dropped and counted as omitted, exactly like a row whose `created_at` cannot be represented. The existing test that pinned the cut for watch id/source was updated with it, and it now also asserts both dropped rows are counted; job_type and status stay cut, because they are display text rather than row keys.

**5. The panel shows watches above the loading and ended states too (Medium).** My previous fallback covered failed and unsupported only. The idle/loading branch and the ended-without-a-retained-tree branch now render the same watch fallback. Two tests added.

**6. `liveWatchStatuses` keeps its documented answer (Low).** Fixed: a known session with no watches keeps the non-nil empty slice its siblings document, so the one path that answered nil for a known session is gone.

obra:
**1. A location fit that dropped its session is rejected now (Medium).** Confirmed end to end: the location candidate maps `budget == 0` to a session-less candidate, `fitCandidates` lands on it whenever no budget fits, and `validateNavigationPageProgress` had no arm for that kind -- so a deep link to a session whose subtree cannot fit (job-heavy, with the watch payload already shed) was served as a 200 with no session at all, where an irreducible overflow used to be reported. The validator now has a location arm: a session-less location fails with its own invariant error (`navigation location invariant: <kind> was fitted without its session`), which the fitter already returns to the caller. Test pins both directions -- session-less rejected, session-carrying accepted -- alongside the existing oversized-location coverage.

**2. The rail no longer prints "0 watches" beside a nonzero total (Low).** Confirmed: when the fitter sheds every retained row, `watchCountLabel` led with `0 watches` while the same line carried `40 armed total · +8 more`. With no retained rows the label now carries the totals alone (`40 armed total · +8 more`), which is what the panel's own watch header does. Test pins the zero-retained case for both a nonzero and a zero armed total; every previous expectation (all with retained rows) is unchanged.

obra:
**Low: a watchless row keeps its nil diagnostics block now.** Confirmed, and the failing test printed the shape it was producing: `&{Tools:[] MCP:[] Skills:[] Plugins:[] HookEvents:[] Jobs:[] Delegates:[] Watches:[] TurnSlots:<nil> Agents:[] ...}` where the cached projection had nil. `attachLiveWatches` attaches for every present entry, and the resolver answers every known session -- so a descendant with no watches reached `appThreadWithWatches` with an empty non-nil sample and got a freshly allocated block.

The resolver's non-nil empty answer stays: it is the contract that tells "no watches now" apart from "no fresh sample", and it is how a cleared watch leaves a row. The fix is at the consumer, as you suggested: `appThreadWithWatches` returns the row untouched when it has no diagnostics block and nothing to attach. The block is materialized only when there is a watch to carry, and kept when the row already had one to clear.

Test pins both directions in one call: the watchless child keeps its nil block, and the root still receives its own watch from the same page sample.

obra:
**Low: the projection is ordered where it is produced now.** Confirmed, and it was mine: the pass-9 change moved `DetailedStatus` off the sorted `liveWatchStatuses` and onto the session-parameterized body, which walks the map and returns it in walk order. The failing test showed the nondeterminism directly -- three runs of the same unchanged state reported the out-of-order row at position 6, then 4, then 1.

Four of the five callers sorted the result afterward; the fifth was the new one, and an unordered projection is indistinguishable from a changed one to any consumer that rebuilds rows from it. The sort now lives in `liveWatchStatusesForSession`, next to the walk, so a call site cannot forget it again; `liveWatchStatuses` drops its now-redundant sort, and the two aggregating callers keep theirs, since they merge rows from several managers and still need to order the merge.

The sort is bounded by the 32-live-watch cap and runs in the same critical section that already formats each row's instants, so it adds no new lock exposure.

Test: eight timer watches plus a job-sourced row, pinned non-decreasing by (source, id) -- enough rows that a map walk cannot land in sorted order by chance, and it fails on every run before the fix.

obra:
All three Lows confirmed and fixed. (The head's two failing CI checks were one `modernize` finding in my own new test loop — `rangeint` on `for i := 0; i < n; i++`, fixed — plus the aggregate `static` job that only mirrors it. My local gate was not running golangci-lint at all; it does now.)

**1. `CloneEvenerDiagnostics` aliased `DelegateDiagnostics` (Low).** Confirmed with a mutation test: writing through the clone changed the source. It is the one slice the hub appends to *after* cloning (`app_threadread.go` attaches delegates.jsonl diagnostics to the returned thread), so a later write could reach the cached block. Copied like its siblings; the test now pins it.

**2. Unrecognized cadence kinds rendered as no trigger at all (Low).** Confirmed, and worse than "omitted": the codec admits any identity `kind` (`identity(value.kind)`), `clockCadenceLabels` filtered everything else out, so the row named no trigger *and* `watchIsScheduled` returned false — the detail then printed "There is no schedule to draw here — this one fires when the job or event it watches says so, not when a clock says so" about a watch whose own wire row says otherwise. Fixed the way you suggested: every cadence row is now labeled through `watchCadenceLabel`'s existing fallback (which already had the honest "name the unknown kind" behavior for the rail, so the two now agree), an unrecognized kind counts as scheduled, and the countdown, the next-fire instant, and the timeline stay keyed to the recognized clock kinds alone. Three tests added: metadata names it, facts get a "Fires …" segment, and the schedule predicate counts it. The existing five-kind expectations and the "empty watch has no schedule" case are unchanged.

**3. The `LiveWatchesForSession` doc stated a contract that no longer exists (Low).** Confirmed: it cited `Detail()` — which does not exist — and claimed the answer was "exactly the set `Detail().Watches` carries", which pass 9 deliberately falsified when `DetailedStatus` was narrowed to the session's own manager for the sampling contract. Reworded to state the real contract, to say explicitly that it is *not* that set and why, and to name the LIST path as where the receiver rollup reaches a row, so a maintainer cannot skip the sampling. The three superseded entry points (`LiveWatchesForSession`, `LiveWatchesForDescendant`, `aggregateWatchStatuses`) are now documented as reached only from tests rather than dropped: they are exported seams whose pinned contracts I would rather not delete unilaterally in a branch that is awaiting review. Flagging that deletion as a follow-up call rather than making it here.

obra:
Confirmed, and fixed at the source. This is the same concern triaged as deferred on an earlier pass; your framing of the exact gap was right, so I closed it rather than deferring again.

What was actually unbounded, checked field by field: `note` was already bounded at registration (`limitWatchText(a.Note, watchMessageMaxChars)`). `output_match` was bounded only where the prose summary *renders* it, while the stored config and the projection carried it verbatim. The events list had no count bound and no dedupe — `canonicalWatchEvents` sorted it and left duplicates in place. `event_filter.tool_name` was whitespace-trimmed only.

The fix is at the registration boundary, in `normalizeWatchArgs` (which runs before every other validation by design), and it refuses rather than truncates: a cut regex or a cut tool name would silently change what the watch matches, which is worse than an error naming the argument. The events list is capped at `maxWatchEvents` (64, generous against a fixed handful of watchable kinds); `output_match` and `event_filter.tool_name` are capped at the existing `watchTriggerMaxChars` (1024) — the same bound already applied when that string is rendered. Duplicate event names collapse in `canonicalWatchEvents`, which runs *after* validation, so the single-concrete-kind rules (`every`, `event_filter`) still read the caller's own list and no error turns into a success.

With those refused, this daemon cannot store or project a watch with an unbounded argument, which is the payload gap you describe: every downstream copy — the diagnostics facet, the page rows, the hub navigation summary, the SPA payload, and each clone along the way — is bounded by what registration allows. That is also why I did not add a projection-side clamp: clamping there would make the row disagree with what the watch actually matches.

Two adjacent notes, no change made. The `send_to` message has no registration-time bound either, but it is not projected onto any watch row and is bounded when it is used (`limitWatchText(..., watchMessageMaxChars)` at delivery), so it is a storage-size question rather than a payload one. And mirroring the new caps in the tool schema and the wire codec is worth doing — the hub already bounds the strings it accepts from any producer, and the budgeter sheds whole rows when an aggregate is over budget — but it only matters for rows written by some other producer now, so I left it as a named follow-up rather than widening this change.

Tests: three oversized cases (events count, `output_match`, `tool_name`) each refused with the field named and the watch count left at zero; duplicates collapsing to the deduped set in the projected row.

obra:
All three confirmed. Two of them are mine from the last round, and the third is the same class of lie this branch has been trying to avoid.

**1. The new bounds counted bytes against a rune limit (Low).** Confirmed: `limitWatchText` counts runes (`len([]rune(s))`), the constants are named `...Chars`, and my two checks used `len()` — so a valid multibyte condition of, say, 900 characters and 1800 bytes was refused while a 900-byte ASCII one was fine. Both checks count runes now, and the test pins it in both directions: a 513-rune value is refused, and a 511-character multibyte value that is *longer in bytes than the bound* is accepted and arms the watch.

**2. Conditions truncated to 512 runes while the daemon allowed 1024 (Low).** Confirmed, and this is the one that mattered: the hub's `maxNavigationLabelRunes` and the wire codec's own `boundedString(item, 512)` are the bound the browser sees, while registration allowed twice that, so a legitimate 600-character `output_match` was cut on its way to the UI and then presented as the condition the watch matches. Marking it a preview would need room the row does not have, so the fix is at the source instead: the two condition fields are now refused above the wire's own bound (512 runes) at registration, which means nothing this daemon produces can be truncated, and the hub's truncation stays exactly what it was — the defense for a row from some other producer. This lowers the registration cap for `output_match` from 1024 to 512; the only values it newly refuses are ones the browser would have displayed incorrectly.

**3. The timeline mislabeled chronology when the browser clock lags (Low).** Confirmed, and reproduced exactly as described: with every retained instant ahead of `now`, the marker and the single dot both drew at `0%` — the failing assertion read `expected 0 to be less than 0`. `span` could not be positive because the rail's left edge was `earliest`, which was already past `now`. The rail now anchors its left edge at whichever anchor comes first (`now` when the clock lags, `earliest` otherwise), so the span stays positive, the marker sits chronologically before the instants it precedes, and each end is labeled with the instant that sits at it — only the end the marker is at says "now". A zero-span rail (one instant exactly equal to `now`) now puts the marker at the opposite end so dot and marker never share a position. The skew case is pinned by a test proven to fail first; the zero-span case is pinned by a guard test, since it already drew apart.

obra:
All three confirmed. Two of them are gaps in what I pushed an hour ago, and the medium is worse than described.

**1. Negative `every` (Medium).** Confirmed, but the mechanism is not the one in the finding, and the measured behavior is worse. A probe (an `every:-3` install through the real configure path, in all three event shapes) shows:

- `every: -3` is **accepted** with one event kind, with several, and with `*`.
- The installed config holds `triggerEvery = 0`. Nothing stores a negative throttle: the daemon silently installs an **unthrottled** watch, so the wire row and the model-facing summary both read "no throttle" and no projector sees anything wrong. A caller who asked for one-in-N gets every matching event.
- Because the single-kind rule is gated on `every > 0`, a negative every also slipped past the checks that refuse a throttle with several kinds or with `*`.

So the value was never dropped from the wire row — it was dropped *as a throttle*, silently, before the wire. I've written the fix against what I measured: `every < 0` is refused in argument normalization with `every must be a positive count (it fires on each Nth matching event)`, placed with the other input-shape bounds so it runs before the `every>0` branches can read it as absent. The test covers all three shapes and asserts nothing installs.

**2. The rendered filter (Low).** Confirmed and fixed. `event_filter.tool_name` is bounded at 512 runes, but the hub renders the filter as ONE label (`tool_name=…, status=…`) and bounds that, so a name at the bound plus its decoration reached ~536 and the hub cut the daemon's own filter — the same truncated-condition-shown-as-the-condition problem as the last round, one field over. The rendered label is now bounded by calling the function that renders it (`watchEventFilterSummary`), so the field bound and the decoration cannot drift apart; the field's own check stays for the crisper message. The test pins the boundary exactly: a filter rendering to exactly 512 runes arms a watch, one rune more is refused.

**3. The now label vs the marker (Low).** Confirmed, and the same review caught a second instance of it next door: the labels were derived from a predicate of their own instead of the marker's actual end. Both are fixed by making "which end the marker is drawn at" a single decision that the labels read:

- when `now` is the rail's left edge — the clock lags, or an instant lands exactly on it — the **left** label is the one that says "now";
- when `now` is the rail's right edge, including the zero-span rail that draws the marker there, the **right** label is;
- when a delivery is ahead of the clock and another is behind it, the marker sits between the ends, so **neither** label claims "now". Previously the right end read "now <newest instant>", naming a delivery as the clock. That case keeps the marker's own time in the rail's `aria-label`, since the marks themselves are hidden from assistive tech.

Tests: the tie case (`now` equal to the earliest instant) asserts the left label says "now" and the marker really draws at 0; the floating case asserts neither end claims "now" and the aria-label names it.

PR #1216, opened 2026-09-13
fix(agent): advance the activity resume index past a single oversized entry

### LINKED ISSUE #1172: agent: activity paging loops forever on a single entry larger than activityMaxEncodedBytes

obra:
Fixes #1172.

When the trailing activity entry dropped by size-trimming was the page's only entry and could not fit on its own, the minted continuation pointed back at that same entry: resuming re-rendered it, re-trimmed it, and minted an identical token forever. The resume index now advances past the dropped entry and the branch error names it, so the skip is reported rather than silent.

## Evidence

- pre-fix: new regression FAIL (identical resume token, page never advances)
- post-fix: `TestLoadSessionJobActivityTree_SingleOversizedEntryResumeAdvances` PASS

This covers the page-1 case. #1177 generalises the minted index to be absolute, so the same shape on page 2+ also advances; both fixes touch `trimActivityTrailingEntry` and are meant to land together.

### DIFF
--- a/agent/jobs_activity.go
+++ b/agent/jobs_activity.go
@@ -1354,6 +1354,18 @@ func trimActivityTrailingEntry(session *appwire.JobActivitySession, rootID strin
 			return true
 		}
 	}
+	resumeIndex := i
+	if i == 0 {
+		// The dropped entry is the page's only rendered entry, so it cannot
+		// fit even alone. A continuation pointing back at index 0 would
+		// re-render it, re-trim it, and mint this identical token on every
+		// later page, so a client paging the tree would never advance past it.
+		// Advance the resume index past the dropped entry and name it in the
+		// branch error so the skip is reported rather than silent -- and so
+		// the next page resumes after it instead of at it.
+		resumeIndex = i + 1
+		appendActivityBranchError(&session.Branch, fmt.Sprintf("activity entry %s is too large to render and was skipped", activityEntryLabel(*entry, i)))
+	}
 	session.Entries = session.Entries[:i]
 	session.Branch.Truncated = true
 	session.Branch.Continuation = encodeActivityContinuation(activityContinuation{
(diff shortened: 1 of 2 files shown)


obra:
Superseded by #1179, which fixes #1172 and #1177 together and carries considerably more edge-case coverage (diagnostic-does-not-fit, fold generations through the ancestor chain, frontend empty page). Closing this duplicate so the two do not fight over jobs_activity.go.

PR #1217, opened 2026-09-13
fix(agent): mint an absolute activity resume index when trimming a page

### LINKED ISSUE #1177: agent: activity size trim on page 2+ re-mints a page-relative ResumeIndex, so paging repeats page 2 forever

obra:
Fixes #1177.

`trimActivityTrailingEntry` minted its resume index from a page-relative position, but the projection had already skipped the page's absolute resume offset. On page 2+ the token therefore pointed back inside the page just returned and paging repeated. The page's absolute offset is now threaded through the trim, and a 60-job walk asserts the token strictly advances.

## Evidence

- pre-fix: page 2 minted `ResumeIndex 16`, which did not advance past the previous page's 16
- post-fix: `TestLoadSessionJobActivityTree_SizeTrimResumeIndexAdvancesAcrossPages` PASS; full agent suite green except the pre-existing `agent/sandbox` bwrap environment failures

Pairs with #1172; both edit `trimActivityTrailingEntry`, and the combined result must mint an absolute index (and, for the single-oversized case at depth 0, one past the dropped entry).

### DIFF
--- a/agent/jobs_activity.go
+++ b/agent/jobs_activity.go
@@ -715,7 +715,7 @@ func projectBoundedActivityTree(snapshot activitySessionSnapshot, rootID string,
 	// collectActivityJobsEpochs and trimActivityTrailingEntry. revision is
 	// the same value just seeded into budget.revision above, embedded the
 	// same way in whatever continuation trimming mints too.
-	return trimActivityTreeToFit(tree, rootID, snapshot.DelegatesEpoch, collectActivityJobsEpochs(snapshot), revision)
+	return trimActivityTreeToFit(tree, rootID, startDepth, resumeIndex, snapshot.DelegatesEpoch, collectActivityJobsEpochs(snapshot), revision)
 }
 
 // collectActivityJobsEpochs walks snapshot's Children tree and returns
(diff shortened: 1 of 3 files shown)


obra:
Superseded by #1179, which fixes #1172 and #1177 together. Closing this duplicate.

PR #1223, opened 2026-09-13
refactor(protocol): move activityRows into the package (SDK C13)

obra:
Plan row: C13 in `docs/superpowers/plans/2026-09-12-sdk-migration.md`

## What moved

`cmd/evener-hub/frontend/src/panes/session/chrome/activityRows.ts` (208 lines)
→ `cmd/evener-hub/frontend/src/protocol/activityRows.ts`, via `git mv` so
history follows. `activityRows.test.ts` (397 lines) moved alongside and stays
the oracle — no logic changes on either side; the only edits inside the two
files are the intra-package imports collapsing from
`../../../protocol/activityData` to `./activityData` (and `./stableDelegate`).

The module was safe to move without extraction: its only imports were
`protocol/activityData` and `protocol/stableDelegate`, both already shipped.

## The three plumbing sites (phase-C rule)

- `protocol/tsconfig.build.json` `files` gains `activityRows.ts`.
- `protocol/index.ts` re-exports the four runtime symbols
  (`activityDelegateState`, `buildActivityRows`, `foldRowID`, `jobIsFailed`)
  and six types (`ActivityDelegateRow`, `ActivityDelegateState`,
  `ActivityFoldRow`, `ActivityJobRow`, `ActivityRow`, `ActivityRowBase`) in
  alphabetical position.
- `scripts/qualify-package.mjs`: `activityRows` in `shippedModules`, the four
  symbols in `rootValues`, `ActivityRow` in `rootTypes`, and four real calls
  added to the root smoke program — `foldRowID`, `buildActivityRows`,
  `jobIsFailed` and `activityDelegateState` all execute in the bare Node
  consumer.

The runner entry is load-bearing, checked by falsification: with the
`index.ts` re-export removed, `make test-api-package` fails with
`shipped module unreachable from every published specifier: activityRows`.
Worth noting for later C rows: adding the module to `tsconfig.build.json`
alone does **not** turn the gate red — the tarball allowlist admits anything
under `package/dist/`, so the gate stays green while saying nothing about the
module, exactly as the phase-C intro warns. The `shippedModules` entry is what
arms the reachability assertion.

## Importers rewritten

Web (deep relative into `protocol/`, matching neighbouring imports since A4
has not landed):
- `panes/session/chrome/ActivityTree.tsx`
- `panes/session/chrome/ActivityRowDetail.tsx`
- `panes/session/chrome/ActivityRowDetail.test.tsx` (not in the original row
  note, but it imports the two row types)

Native (existing deep path, retargeted):
- `mobile-native/src/ActivitySheet.tsx`

A repo-wide grep for `activityRows` leaves only two hits outside the moved
files: a comment in `protocol/activityMerge.test.ts` naming the test file (its
new sibling) and the historical 2026-08-05 / 2026-08-12 plan documents, which
record where the file was created and are not edited. No test spies on the
module path.

## Gates

- `make test-api-package` — PASS (and demonstrably FAILS without the manifest entry)
- `make test-web` — PASS (typecheck, test, lint)
- `make test-native` — PASS (740 + 777 tests, `tsc --noEmit`)
- `make test-web-browser` — PASS (all five guards), per plan rule 1 for a package-touching PR

Biome `check --write` was run over the touched files under
`cmd/evener-hub/frontend/src` only; `mobile-native/` was deliberately left
alone (no biome config there).

### DIFF
--- a/cmd/evener-hub/frontend/src/panes/session/chrome/ActivityRowDetail.test.tsx
+++ b/cmd/evener-hub/frontend/src/panes/session/chrome/ActivityRowDetail.test.tsx
@@ -2,10 +2,10 @@ import { cleanup, render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
 import type { ActivityDelegate, ActivityJob, ActivitySessionNode } from "../../../protocol/activityData";
+import type { ActivityDelegateRow, ActivityJobRow } from "../../../protocol/activityRows";
 import { connectionStore } from "../../../stores/connection";
 import { threadsStore } from "../../../stores/threads";
 import { ActivityRowDetail } from "./ActivityRowDetail";
-import type { ActivityDelegateRow, ActivityJobRow } from "./activityRows";
 
 // Pinned clock: every quiet-age assertion below measures against this instant.
 const NOW = Date.parse("2026-08-05T15:00:12.000Z");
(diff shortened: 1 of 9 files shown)


obra:
Residual from this lane filed as #1224: the runner stays green when a module is built into dist but missing from `shippedModules`; fix lands after #1222 and #1223 since both edit the manifest.

Issue #1195, opened 2026-09-12
agent: activity trim continuations stamp the page root's DelegatesEpoch, but staleness checks the target session's own


obra:
Found in #1179 round 2 (part of #1116).

Size-trim and depth-truncation continuations minted in `agent/jobs_activity.go` (`trimActivityTreeToFit`, `markActivityDelegateTruncated`) carry the page root's `DelegatesEpoch` for every session in the page, while the next request's staleness check compares the continuation *target's* own generation. They agree only when the child's meta names the root (`JobTreeRootSessionID`), so both read the same shared `delegates.jsonl`. A child whose meta lacks that field reads its own (usually absent) journal at generation 0; a nested continuation minted while the root's generation is nonzero is then rejected as stale even though nothing changed. The #1179 epoch fixture hit exactly this until the child's root ID was set.

Decide whether `JobTreeRootSessionID` is guaranteed for every delegate child (then document and assert it at load) or mint each session's continuation with that session's own epochs. Related: #1172, #1177, #1179.

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

**Verdict:** `legit-current` — mint and check read different sessions' epochs. `trimActivityTrailingEntry` stamps every session's continuation with the page root's `DelegatesEpoch` (`agent/jobs_activity.go:1366`; `markActivityDelegateTruncated` likewise passes the parent snapshot's epoch at `:1084`), while resume compares against the continuation target's own `snapshot.DelegatesEpoch` (`:339`). They agree only when every session resolves its root to the page root via `JobTreeRootSessionID` (`agent/jobs_activity_past.go:492-497`). A child whose meta lacks that field folds its own (usually absent) journal at generation 0 (`:400,416`), so a nested continuation minted while the shared journal's epoch is nonzero is falsely rejected as stale. `JobTreeRootSessionID` is set only when `jobActivityClock != nil` (`agent/session_state.go:132-137`), so the invariant is unchecked.

**Autonomy:** `needs-human` — the issue poses a genuine either/or (guarantee and assert the invariant, or mint each session's own epochs); the choice changes the documented meaning of the continuation field.

**Recommendation:** `needs-decision`.

obra:
[...]

PR #1231, opened 2026-09-13
fix(agent): mint activity continuations with the generations of the session they name

obra:
Follow-up to #1179; closes the round-8 Medium that its merge predated.

A continuation is checked against the fold-cache generations of the session it **targets**, but two mints named someone else's generations — or none at all.

## Live root paging into a closed child (#1179 round 8)

A size-trim mint took the delegates generation from the page's own root. A live root reads neither fold cache, so a token naming a closed child carried zeros, and the stale check had to skip validation entirely for live roots to keep that branch readable. What the exemption gave up was the only evidence of that child's journal being rewritten: the root's activity clock moves for what the daemon does under it, never for a file rewritten out of band.

`collectActivitySessionEpochs` now returns both generations per session and every trim mint carries the trimmed session's own, so the `root.live == nil &&` exemption is deleted rather than narrowed. A live target still compares zeros to zeros and is fenced by the revision; a closed child under a live root is fenced by both.

## Depth-truncated children minted zeros (already broken on main for historical roots)

A child past `activityMaxNewDepth` is deliberately never loaded, so its placeholder had no generations and `markActivityDelegateTruncated` minted `0,0`. The child's own load reports the real generation of the shared delegate journal it folds, so a resume refused a token nobody had invalidated and everything past the bound became unreadable once that journal had been rewritten. This predates #1179: the check for a historical root was the same comparison before round 8, and the placeholder already minted zeros.

`foldcache.Cache.Epoch(path)` reports what the next fold would attach **without folding** — the recorded counter, or 0 for a path never folded, which is exactly what its first fold assigns. `activityPlaceholderEpochs` reads it at placeholder creation: a live child gets `0,0` (what `loadLiveActivityBase` reports), a historical one gets its jobs journal's counter and the counter of the delegate journal its metadata names. Cost is one `schema.LoadSessionMeta` read plus two map lookups per depth-boundary delegate — no journal fold, which is what the depth bound exists to avoid. A journal genuinely rewritten in between raises the fold's number, and that mismatch is a true refusal.

## Tests (RED first, recorded)

- `TestJobActivityTree_LiveRootChildContinuationRejectedAfterTheChildIsRewritten` — live root, closed child, child's journal restamped between mint and resume. RED: `resumed into a closed child whose journal was rewritten; want the continuation refused`.
- `TestJobActivityTree_LiveRootChildContinuationSurvivesHistoricalEpochs` — the readability half: an untouched closed child still resumes under a live root.
- `TestLoadSessionJobActivityTree_DepthContinuationResumesIntoAnUntouchedChild` — 33-hop chain whose shared delegate journal has been rewritten once. RED with the placeholder minting zeros: `depth continuation into an untouched child was rejected: activity continuation is stale: the underlying journal changed`.
- `TestLoadSessionJobActivityTree_DepthContinuationRefusedAfterTheChildIsRewritten` — the child is folded once, then restamped between mint and resume; the resume is refused, so naming the generations is not a rubber stamp.
- `TestCollectActivitySessionEpochs` and `TestTrimActivityTrailingEntry_MintsTheTrimmedSessionsOwnDelegatesEpoch` pin the per-session rule directly.

## Gates

Toolchain gofmt (`$(go env GOROOT)/bin/gofmt -l`), `go vet` host + `-tags evenerfuzz` + `GOOS=windows -tags evenerfuzz` in the agent and root modules, `go test -count=1 ./...` with zero non-ok lines for agent, appprojector and the hub server, `-race -count=3 -timeout 30m` over the activity tests, pinned golangci-lint 0 issues.

Related: #1172, #1195.

## Review round 1

- **Unbounded I/O at the depth boundary (Medium)** — fixed in `63300a877`. Naming a depth-truncated child's generations reads that child's metadata, and a wide boundary did it once per child with nothing counting the reads and nothing watching the request's context. It now pays the same work unit a child one level shallower pays before being descended into, and stops where a canceled request stops everything else; a budget with nothing left leaves the child out rather than minting a continuation nobody paid for. RED: `minted 6 placeholders on a budget of 2 units`.
- **Legacy V1 tokens minted with zero generations are now refused (Medium) — declined.** A continuation is a page cursor held in a client's memory, and the refusal says exactly what to do about it: restart pagination without a continuation, which is the designed recovery. A shim for tokens minted by a build that is no longer running is backward compatibility, which this repo does not add without Jesse's approval. The stale-check comment now states it: a token minted before generations were carried names none, so it is refused once and the client restarts.
- **`Epoch` vs `Get` after a delete (Low)** — fixed in `7f68c014b`. `Get` returns the zero Result for a missing file, so `Epoch` now reports 0 for a path that is gone instead of the tombstone kept for the content that used to be there; the tombstone still does its work on a path that comes back. The existing tests pick this side: `TestCache_DeletedFileEpochSurvivesAcrossRecreation` pins that a *recreated* file advances past its pre-deletion generation, and asserts nothing about the missing-file answer, which `Get` already reports as 0. New `TestCache_EpochAgreesWithGetAcrossDeletionAndRecreation` pins the pair across fold, delete and recreate. RED: `Epoch = 1 for a deleted path while Get reports 0`.
- **Doc comments (Low)** — `624bfe04b`. `DelegatesEpoch` now says it is the generation of the delegates journal the TARGET session folds, not `RootID`'s, and that a closed child under a live root is generation-fenced as well as revision-fenced; the `Revision` comment says the same from its side; the collector's name and the trim's parameter list are no longer stale; `foldcache.Epoch` has its own comment and `Stats` has its own back.

## Review round 2

- **Naming a generation reported the last fold's number, not the next one's (Medium)** — fixed in `93322a512`. `Epoch` read the recorded generation without the freshness check `Get` performs, so a depth-truncated child that was folded earlier and rewritten since was named one generation behind: the first resume discovered the rewrite, moved the generation, and refused a continuation minted *after* the rewrite. The decision now lives in one place — `freshnessOf`, lifted out of `refresh` — and both callers run it, so naming a generation asks the same question a fold asks. It never folds and never mutates cache state. RED: `continuation minted after the rewrite was rejected: activity continuation is stale: the underlying journal changed`.
- **Cost per depth-truncated child**: one `os.Stat`, plus the tail probe (a `tailProbeBytes` read at the recorded offset) when size and mtime cannot settle the question on their own. Still no fold, and still charged as the one work unit that child pays.
- **The cancellation test did not pin the placeholder guard (Low)** — `ee6c82c84` moves the cancellation into the root's own delegate scan, the last read the base load makes. It still does not isolate the guard, and the test says so: `foldcache.Get` runs its fold detached and then selects on the caller's context, so a cancellation arriving during **any** read is reported by that read before the placeholder loop runs. Deleting the guard leaves the test green — checked by hand. The guard is kept as the loop's own answer for a cancellation with no read left to report it; removing it instead would be a deliberate call, not one made by leaving it untested.

## Review round 3

- **A depth-boundary delegate could lose its continuation (Medium)** — fixed in `b768763d7`. When the load budget ran out before it could name a child's generations there was no placeholder, and projection reached for the child before deciding truncation, so the branch came back as "child session unavailable" with nothing to continue to — a dead end where it used to be a page the client could ask for. `projectStableActivityDelegate` now decides truncation first and mints the continuation either way, carrying the child's generations when they were named and none when they were not; a token that names none is refused once and the client restarts. The load-time comment that described the old ordering is corrected. Grepped every other `snapshot.Children[` use: the remaining ones are the loader's own writes and its dedupe check, none of them a decision behind a dereference. RED: `delegate "dlg_boundarytruncchild1" reports "child session \"boundarytruncchild1\" unavailable" instead of a truncated branch`.
- **Two readings of the delegate generation (Low)** — fixed in `86b232bb8`. Naming it went to the fold cache directly while the loader reads it through the traversal's own delegate index, and the two disagree about a journal with a line too long to scan: the loader degrades to an empty delegate set at generation 0 while the cache still holds the last good fold's. The placeholder now reads the index the loader reads, so there is one answer instead of two, and the helper that made the second reading is gone. RED: `placeholder names generation 1 while the loader reports 0`.

## Review round 4

**Round 3's "refused once" answer is withdrawn.** A continuation with no generations behind it is refused by every resume of a child anything ever folded — a certainty, not an occasional cost — so offering one at budget exhaustion was a dead end dressed as a page.

- **No unfenced token (Medium)** — `0a02c1c8e`. Where the load budget ran out before it could name a child's generations, the depth-boundary delegate is now marked truncated with **no continuation** and a diagnostic on `JobActivityDelegate.Diagnostics` (`appwire/types.go:1865`, already carried by the projector and the wire — no shape change): `load budget exhausted; request session "<id>" directly`. Nothing failed, so no branch error. The generation lookup is never skipped for a placeholder that IS minted. RED: `0 delegates reported the exhaustion, want 3`; the test also follows the diagnostic's own advice and asserts that direct request succeeds.
- **Lookup failures no longer swallowed (Medium)** — `97d6f9fea`. `activityPlaceholderEpochs` returned `(0, 0)` for a failed delegate-index read, so a canceled request came back as a successful page and an unreadable journal minted a token against a number nobody read. It returns the error now. An unresolvable child link and a live child still answer zero: the first is recorded per child in `snapshot.Errors` and reported from there, the second folds nothing. RED: `err = <nil>, want the journal read's own failure` and `err = <nil>, want context.Canceled`.
- **Comment split (Low)** — `6e5279180`. `refresh` has its own doc comment back and `freshness` has its own.

## Review round 5

- **`Epoch` read the file before the state it is compared against (Medium)** — `14733e889`. A fold landing between the two readings left stale file information to be judged against newer state, which reads as a shrink and names a generation one past anything a fold produces — a rewrite nobody performed, refusing a resume nothing invalidated. The state is read first now, so it can only be older than the file, which is the case this already answers for. RED (deterministic interleave via a test hook, plus a concurrent sweep kept under `-race`): `Epoch named generation 2 across an interleaved fold while the next fold carried 1`.
- **An unresolvable depth-boundary child got a placeholder (Medium)** — `aee52167b`. The lookup swallowed the resolver's error and a placeholder was installed anyway, so the branch offered a continuation whose resume fails on that same error, and the comment claiming the caller recorded the failure was wrong. The child is now resolved where a child one level shallower is resolved, the failure is recorded in `snapshot.Errors` and reported by projection's existing check, and no placeholder is installed. Zero generations now mean only a live child, which folds nothing. RED: `no error recorded for "unresolvablechild"; a placeholder here mints a continuation that fails on resume for this same reason`.
- `50041bb6b` is a follow-up for the lint's modernize check (`WaitGroup.Go` in the new ordering test), landed after the merge rather than amended across it.

## Review round 6

**CI failure first.** The `tests` job on run 34773862425 failed in `TestCache_EpochNeverExceedsTheNextFoldsGeneration` at `foldcache_test.go:929` — the concurrent sweep, whose assertion text (`-- a generation no fold produces refuses a resume nothing invalidated`) is unique to that half. The cause was the test, not the cache: the sweep's writer rewrote the file with `os.WriteFile`, which truncates before it writes, so a stat landing inside that window sees a file shorter than the completed content the next fold reads. That is a shrink by every rule this cache has, so the naming came back a generation above the fold with nothing wrong underneath. Linux hits the window; twenty local macOS runs did not. Fixed in `dbbc7edc9`: the sweep appends — the case the invariant is about — and its writer no longer calls `t` from its own goroutine; rewrites stay in the deterministic half, where the fold is sequenced against the naming.

- **A named generation for a journal nobody can read (Medium)** — `5afe4f511`. `Epoch` turned missing, unreadable and unprobeable files alike into 0 or the last good generation, and the placeholder path never checked that the child's `jobs.jsonl` could be loaded, so a continuation could be minted whose resume then fails on that same file. `Epoch` returns `(uint64, error)` now: a path that is not there answers 0 — no generation to discard is what its first fold assigns — and every other stat or probe failure is returned. The placeholder path records that as the child's own failure (no placeholder, no continuation, the branch carries the error), while a failure belonging to the traversal — the shared delegate index, a canceled request — still fails the load. RED: `no error recorded for "unreadableboundarychild"; a placeholder here mints a continuation whose resume dies on the same unreadable journal`, and at the cache seam `Epoch answered for a path it could not stat`.

## Review round 7

**Round 6's "an absent journal is generation 0" is corrected.** It holds only where the loader tolerates absence, which is a request's own root and never a depth-truncated child: that child's journal is required, so absence there is the loader's own failure, reported with its own words.

- **A torn stat read as a rewrite (High)** — `60ab615e1`. `os.Stat` does not read size and mtime as one snapshot, so during an append it can return the size from before the write with the mtime from after it — indistinguishable from a same-size rewrite. The same-size branch bumped on that appearance without asking the tail probe, naming a generation the next fold does not carry, and the resume was refused for a journal nothing had rewritten. The branch asks the probe now and moves the generation only when the recorded prefix is gone. Measured: CI hit it about one run in four under `-race`; locally 0/20 runs at 2000 sweep rounds never reached the window, so the regression test is deterministic instead — `TestFreshnessOf_TornAppendStatKeepsTheGeneration` builds the exact torn observation. RED: `generation moved to 1 on a torn append stat, want it to stay at 0`. After: foldcache green at `-race -count=10`.
- **A generation named for a journal the load cannot read (Medium x2)** — `a58cbdd83`. The placeholder now validates the child's journal exactly as the load a resume performs does — absent means `child session "…" unavailable in state directory`, recorded as that child's failure with no placeholder and no continuation — and `Epoch` answers 0 for a never-folded path only after a regular-file check and an open, so a directory or an unreadable file returns its error instead of a number. RED: `no error recorded for "missingjournalchild"; generation 0 and a continuation here hand the reader a page whose resume fails`, and `Epoch answered generation 0 for a never-folded path no fold could read`.
- **Test hook off the globals (Low)** — `4fa228e99`. The interleaving hook is a field on the cache whose readings it sequences, nil everywhere but the one test that sets it on its own instance.


obra:
## Review round 1 disposition (head f94c310ea)

- **Medium, unbounded I/O at the depth boundary**: fixed, 63300a877. The placeholder lookup checks the loader's context and pays one work unit, the same unit a loaded child pays before being descended into; an exhausted budget leaves the child out through the existing exhaustion path. RED: six placeholders minted on a budget of two units; GREEN, plus a cancellation case asserting `context.Canceled` surfaces.
- **Medium, legacy V1 tokens with zero epochs now refused**: declined. A continuation is a page cursor held in a client's memory, the error already names the recovery ("restart pagination without a continuation"), and a shim for tokens minted by a build that no longer runs is backward compatibility that has not been approved. Stated in the stale-check comment.
- **Low, `Epoch` vs `Get` after a delete**: fixed, 7f68c014b, on the `Epoch` side. `Get` returns the zero result for a missing file and the existing test pins only the recreated case, so `Epoch` now reports 0 for a path that is gone while the tombstone still advances the generation when the path returns. RED: `Epoch = 1 for a deleted path while Get reports 0`; GREEN. The agreement is pinned at the foldcache seam because a continuation target whose journal is gone fails earlier with "child session unavailable".
- **Low, four doc comments**: fixed, 624bfe04b. `DelegatesEpoch` names the target session's journals and the live-root/closed-child fencing, `Revision` likewise, the collector name and trim parameter list are current, and `foldcache.Epoch` has its own comment with `Stats` documented again.

Gates: toolchain gofmt; host, evenerfuzz and windows vet 0 in both modules; agent, appprojector and hub server suites clean; `-race -count=3 -timeout 30m` on the activity tests and `-race -count=3` on foldcache; golangci-lint 0 issues. Merged `origin/main`.

obra:
## Review round 2 disposition (head 88839dd6b)

- **Medium, `Epoch` names the last fold's generation without a freshness check**: confirmed, fixed in 93322a512. The freshness decision is one shared `freshnessOf` that both the fold path and the non-folding lookup run, so naming a generation asks exactly the question a fold asks, never folds, and never mutates cache state. RED: a continuation minted after the child's rewrite was refused as stale; GREEN. Cost per placeholder: one `os.Stat`, plus a tail probe at the recorded offset only when size and mtime cannot settle it, still charged as the one work unit that child already pays.
- **Low, cancellation test never reaches the placeholder guard**: moved in ee6c82c84 so the cancel lands inside the root's own delegate scan, and reported honestly: it still does not isolate the guard, because `foldcache.Get` selects on the caller's context after every read, so a cancellation that arrives during any base read is reported by that read first. The guard is reachable only for a cancellation that arrives between placeholders, which a test cannot place without a hook; it stays, with the test comment stating what the test does and does not prove.

Gates: toolchain gofmt clean; host, evenerfuzz and windows vet 0 in both modules; agent, appprojector and hub server suites clean; `-race -count=3 -timeout 30m` on the activity tests, `-race -count=3` on foldcache and delegates; golangci-lint 0 issues. Merged `origin/main`.

obra:
## Review round 3 disposition (head 54c3d1427)

- **Medium, budget-exhausted depth boundary yields "child session unavailable" with no continuation**: confirmed, fixed in b768763d7. `projectStableActivityDelegate` decides truncation before dereferencing the child and mints a continuation either way: the child's generations when they were named, none when the budget prevented naming them (refused once, then the client restarts). The load-time comment describing the old ordering is corrected. RED: a depth-boundary delegate reported the branch error; GREEN, and the minted token resumes. The other `snapshot.Children[` uses are the loader's dedupe check and its two writes; no other decision sits behind a dereference.
- **Low, placeholder delegates generation ignores the loader's degraded-journal answer**: confirmed, fixed in 86b232bb8. The placeholder now reads the same delegate index the loader reads (which degrades to generation 0 on `ErrLineTooLong`), and `currentHistoricalDelegatesEpoch` is deleted rather than left as a second answer. RED: placeholder named generation 1 while the loader reported 0; GREEN.

Gates: toolchain gofmt clean; host, evenerfuzz and windows vet 0 in both modules; agent, appprojector and hub server suites clean; `-race -count=3 -timeout 30m` on the activity tests, `-race -count=3` on foldcache and delegates; golangci-lint 0 issues. Merged `origin/main`.

obra:
## Review round 4 disposition (head 6e5279180)

- **Medium, unfenced `(0,0)` continuation at budget exhaustion**: the round-3 "refused once" answer is withdrawn; it was a deterministic false rejection for any child ever folded. Fixed in 0a02c1c8e: at exhaustion the depth-boundary delegate is truncated with no continuation and a diagnostic (`load budget exhausted; request session "<id>" directly`) in `JobActivityDelegate.Diagnostics`, a field already on the wire and already populated by the projector, so no wire-shape change. RED: none of the boundary delegates reported the exhaustion; GREEN, and the test follows the diagnostic's own advice by asserting the direct request succeeds.
- **Medium, placeholder lookup failures swallowed into `(0,0)`**: fixed in 97d6f9fea. The delegate-index read's error propagates (RED: `context.Canceled` and an injected read failure both came back as nil); an unresolvable child link and a live child still answer zero, recorded per child in `snapshot.Errors` rather than failing the whole tree, with the lint classification stated in place.
- **Low, `refresh` doc comment detached**: fixed in 6e5279180.

Gates: toolchain gofmt clean; host, evenerfuzz and windows vet 0 in both modules; agent, appprojector, hub server and TUI suites clean; `-race -count=3 -timeout 30m` on the activity tests, `-race -count=3` on foldcache and delegates; golangci-lint 0 issues. Up to date with `origin/main`.

obra:
## Review round 5 disposition (head 50041bb6b)

- **Medium, `Epoch` stats before reading the recorded state**: confirmed, fixed in 14733e889. The state is read under the lock first, so a fold landing between the two readings can only leave the state older than the file, which is the case already answered for. RED: `Epoch` named generation 2 across an interleaved fold while the next fold carried 1; GREEN. The plain concurrent sweep never hit the window, so the deterministic half uses a nil-by-default hook between the two readings, and the test also runs 200 rounds against a goroutine folding and rewriting under `-race`.
- **Medium, unresolvable depth-boundary child gets a placeholder and a doomed continuation**: confirmed, fixed in aee52167b. Resolution moved into the loader beside the non-placeholder path: the failure goes to `snapshot.Errors`, projection's existing check reports it, and no placeholder or continuation is minted; `activityPlaceholderEpochs` takes the resolved locator, so zero generations now mean only a live child, and the `nilerr` suppression the old swallowing required is gone. RED: no error recorded for the unresolvable child; GREEN.
- 50041bb6b is a lint follow-up (modernize) on the new test's WaitGroup use, after the merge.

Gates: toolchain gofmt clean; host, evenerfuzz and windows vet 0 in both modules; agent, appprojector, hub server and TUI suites clean; `-race -count=3 -timeout 30m` on the activity tests, `-race -count=3` on foldcache and delegates; golangci-lint 0 issues. Merged `origin/main`.

obra:
## Review round 6 disposition (head 49f4b136f)

- **CI failure at 50041bb6b** (`TestCache_EpochNeverExceedsTheNextFoldsGeneration`, ubuntu only): a defect in the test, fixed in dbbc7edc9. The concurrent sweep's writer used `os.WriteFile`, which truncates before writing, so `Epoch`'s stat could land mid-rewrite and see a shorter file than the completed content the next fold reads, which the cache correctly treats as a shrink and names one generation above. Linux hits that window; macOS never did in twenty runs. The sweep now appends (the case the invariant is about) and its writer no longer calls the test from its goroutine; rewrites stay in the sequenced deterministic half; the assertion is untouched.
- **Medium, placeholders never validate the child journal and `Epoch` swallows non-`NotExist` errors**: fixed, 5afe4f511. `Epoch` returns `(uint64, error)`: a path that is not there answers generation 0, every other stat or tail-probe failure is returned. The placeholder path records such a failure as the child's own (`snapshot.Errors`, no placeholder, no continuation), while traversal-level failures (shared delegate index, cancellation) still fail the load, told apart by `activityChildJournalError` rather than by error kind; a first cut that conflated them regressed the round-4 `io_error` case and the suite caught it before the push. RED: no error recorded for an unreadable boundary child, and `Epoch` answered for a path it could not stat; GREEN.

Gates: toolchain gofmt clean; host, evenerfuzz and windows vet 0 in both modules; agent, appprojector, hub server and TUI suites clean; `-race -count=3 -timeout 30m` on the activity tests, `-race -count=3` on foldcache and delegates; golangci-lint 0 issues. Merged `origin/main`.

obra:
## Review round 7 disposition (head 3436b1e78)

- **High, `Epoch`'s same-size/different-mtime branch bumps on a torn-append stat**: real, fixed in 60ab615e1. The branch now runs the tail probe and bumps only on a mismatch, so a stat that caught an append's new mtime with its old size stays at the old generation, which is what the following fold reports. CI hit the window about one run in four; macOS could not reach it in 20 `-race` runs at 2,000 sweep rounds, so the regression is deterministic instead: `TestFreshnessOf_TornAppendStatKeepsTheGeneration` constructs the exact torn observation. RED: generation moved to 1 on a torn append stat; GREEN, and foldcache is green at `-race -count=10`.
- **Medium (three reviewers), a missing child `jobs.jsonl` still gets a placeholder and a continuation**: real, fixed in a58cbdd83; round 6's "absent means generation 0" is corrected. Every child, depth-truncated included, is loaded with `required` set, so absence there is the loader's failure: the placeholder path now records the loader's exact "child session %q unavailable" message and mints nothing. Generation 0 for an absent journal survives only where the loader tolerates absence (a request's own root).
- **Medium, `freshnessOf` answers for a never-folded path it never opened**: fixed in a58cbdd83 with a regular-file-and-open preflight (a directory opens fine, so both checks are needed). RED: `Epoch` answered generation 0 for a path no fold could read; GREEN.
- **Low, `epochInterleave` test-only global in production code**: fixed, 4fa228e99.
- Consequence: restamping an mtime no longer moves a generation (the probe answers for identical content), so four fixtures that used `os.Chtimes` now remove the journal, let a read observe it gone, and restore the same bytes.

Gates: toolchain gofmt clean; host, evenerfuzz and windows vet 0 in both modules; agent, appprojector, hub server and TUI suites clean; `-race -count=3 -timeout 30m` on the activity tests, `-race -count=10` on foldcache, `-race -count=3` on delegates; golangci-lint 0 issues. Merged `origin/main`.

obra:
## Reset (Jesse's ruling, 2026-09-13)

This PR began as two fixes from #1179's last review: fence a continuation with the target session's generations (deleting the live-root exemption), and give depth-truncated placeholders generations of their own. Over eight rounds the second fix grew into a mechanism that predicts, without folding, the generation the next fold will report, and every round found a real hole in that prediction (a rewrite between stat and fold, a torn append, a missing or unreadable journal, a directory in a journal's place, a delete between stat and probe, an oversized line no probe can see). The findings were right; predicting a fold is the wrong class of solution.

A depth-boundary continuation carries no resume position; it delivers exactly the page a direct request for the child session delivers, and the child's id is already on the wire. So the replacement PR drops the predictor entirely: depth-boundary delegates are truncated with no continuation and the diagnostic already introduced in round 4 (`request session "<id>" directly`), the first fix stays as is, and the one finding that lived on the real fold path (the torn-append tail probe in `freshnessOf`) stays too. Everything else from rounds 1 to 7 is deleted with the code it served.

This PR closes as superseded once the replacement is open; the branch stays as a reference.

obra:
Superseded by #1269 (the reset described above); branch kept for reference.

PR #1235, opened 2026-09-13
fix(llm): stop counting thinking the adapter will not replay (#653)

### LINKED ISSUE #653: Non-replayable Thinking.Text is counted toward the context estimate
## What happens

`llm/token_count.go` `estimateMessageInputParts` counts `len(p.Thinking.Text)` toward the context estimate for every assistant thinking part. Some of that text is never sent back to the provider:

- OpenAI Responses replay (`toResponsesInput`) only re-sends reasoning items with `encrypted_content`; raw `reasoning_text` content kept on the part (gateway-fronted GLM via lunaroute) is display-only.
- `replayScope.thinkingReplayEligible` strips thinking on a cross-provider switch.

So a session whose model produces large raw reasoning (25k tokens per turn was observed) over-reports context pressure by roughly that much per turn, and by the entire accumulated reasoning history on a resumed session, where `estimateUsedTokens` has no provider baseline and char/4-estimates the whole history. Compaction warnings, checkpointing, and summarization fire earlier than real usage warrants.

## Possible fix

Have the estimator ask the replay policy whether a thinking part will be sent for the active provider/adapter (encrypted content present, signature present, same provider) and skip the rest.

Found by adversarial review of the reasoning-visibility branch; pre-existing modeling gap, made more visible now that raw reasoning text is persisted.

obra:
Fixes #653.

The context estimator billed a thinking part's `Text` and `Signature` unconditionally, including raw `reasoning_text` that the Responses adapter keeps for display but never replays.

`thinkingReplayChars` now bills a thinking part only when it carries provider-scoped replay metadata (`EncryptedContent` or `Signature`), or when it is a redacted-thinking part whose payload is replayed verbatim.

Evidence:
- `llm/token_count_test.go` and `agent/internal/contextmgr/context_manager_test.go` fail on the pre-fix tree (100 vs 0 tokens for a non-replayable part; 9 vs 1 through the compaction path) and pass here.
- `go test ./llm/ ./agent/internal/contextmgr/` green.

Known limitation, documented in the code: replay is a property of the adapter, not of the part. `EstimateMessagesInputTokens` carries no provider or model (the compaction path calls it with both empty), so one rule stands in for every adapter. An OpenAI-compatible chat adapter replays plain reasoning text, which this under-counts, and replays an OpenAI-compat encrypted array that the Responses adapter rejects, which this over-counts. Deciding per adapter needs the provider threaded into the compaction estimate; that is a larger change and is not attempted here.

### DIFF
--- a/agent/internal/contextmgr/context_manager_test.go
+++ b/agent/internal/contextmgr/context_manager_test.go
@@ -81,10 +81,27 @@ func TestEstimateTokens_WithThinking(t *testing.T) {
 		}},
 	}
 	got := estimateTokens(turns)
-	totalChars := len("let me think about this carefully") + len("answer")
-	want := totalChars / 4
+	// A thinking part with no replay metadata (no signature, no encrypted
+	// content) is display-only — the adapter never re-sends it — so it must not
+	// be billed to the context estimate (issue #653). Only the answer counts.
+	want := len("answer") / 4
 	if got != want {
 		t.Fatalf("EstimateTokens = %d, want %d", got, want)
+	}
+
+	// Thinking the adapter will replay (provider-scoped signature) is billed.
+	replayable := []schema.Turn{
+		{Kind: schema.TurnAssistant, Message: llm.Message{
+			Role: llm.RoleAssistant,
+			Content: []llm.ContentPart{
+				{Kind: llm.ContentThinking, Thinking: &llm.ThinkingData{Text: "let me think about this carefully", Signature: "crypto-sig"}},
+				{Kind: llm.ContentText, Text: "answer"},
+			},
+		}},
+	}
+	totalChars := len("let me think about this carefully") + len("crypto-sig") + len("answer")
+	if got := estimateTokens(replayable); got != totalChars/4 {
+		t.Fatalf("EstimateTokens = %d, want %d", got, totalChars/4)
 	}
 }
 
--- a/llm/token_count.go
+++ b/llm/token_count.go
@@ -249,6 +249,15 @@ func thinkingReplayChars(p ContentPart) int {
 		return len(t.Text)
 	}
 	if t.EncryptedContent != "" {
+		// An OpenAI-compatible encrypted reasoning_details array is replayed by
+		// the chat adapter together with the separately parsed text
+		// (chatcompletions/messages.go), and the Anthropic adapter replays the
+		// text while ignoring the blob. The opaque OpenAI Responses blob is the
+		// other shape: it replays with its summary and id and puts no text on
+		// the wire, so only that shape bills them.
+		if IsOpenAICompatEncryptedReasoning(t.EncryptedContent) {
+			return len(t.EncryptedContent) + len(t.Text)
+		}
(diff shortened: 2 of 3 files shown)


obra:
Roborev's findings were addressed in ab5923dac9:

- **Medium: `EncryptedContent` never counted / `Text` over-counted for the Responses shape** - fixed. `thinkingReplayChars` now bills the encrypted blob plus its summary and id, which is what `responses/input.go` actually resends, and stops billing `Text` for that shape. Added `TestEstimateMessagesInputTokens_BillsReplayedEncryptedBlob` (blob-only part is billed, estimate grows with the blob).
- **Low: redacted thinking over-counts `Signature`** - fixed. Redacted thinking now bills `Text` only, matching `anthropic/request.go` emitting `"data": Text`.
- **Medium: unsigned Anthropic thinking estimated as 0** - not fixable provider-blind: the same shape (text, no replay metadata) is display-only for a gateway-fronted GLM and replayed unsigned by Anthropic. Rather than break the reported case, both the unsigned-Anthropic and `ThinkingAsText` paths are now named in the code comment as known limitations of a provider-blind rule. Added `TestEstimateMessagesInputTokens_RedactedThinkingBillsTextOnly`.

Re-review requested on ab5923dac9.

obra:
Re-review follow-up pushed (7b35f8b98e):

- **Medium: encrypted-content branch drops replayable `Text`** - fixed. The branch now distinguishes the two blob shapes with `IsOpenAICompatEncryptedReasoning`, the same predicate the adapters use. A compat `reasoning_details` array bills the blob plus the separately parsed text (which `chatcompletions/messages.go` merges into the details, and which Anthropic replays while ignoring the blob); the opaque Responses blob bills the blob with its summary and id and still bills no text, because that adapter sends none. Added `TestEstimateMessagesInputTokens_CompatEncryptedBlobBillsItsText`, which also pins that `ID`/`Summary` do not change a compat blob's estimate.
- **Low: `ID`/`Summary` over-counted for compat blobs** - fixed by the same change.
- **Medium: unsigned thinking bills zero** - unchanged and still documented as a known limitation. It is the deliberate trade-off of a provider-blind rule: the identical shape (text, no replay metadata) is display-only for a gateway-fronted GLM and replayed unsigned by Anthropic, so billing it would reintroduce the over-count this series exists to fix. Making it exact needs provider/caps threaded into the compaction estimate, which is a separate change and a call for the maintainer.

obra:
Maintainer decision on the remaining Medium (unsigned thinking billed as zero): **do not fix, keep it documented.**

Rationale, recorded so the next reader does not re-litigate it: the identical data shape (text, no replay metadata) is display-only for a gateway-fronted GLM and replayed unsigned by Anthropic and the `ThinkingAsText` merge path. Billing it would reintroduce the over-count this series exists to fix, and the only exact answer needs provider or caps threaded into the compaction estimate, since `EstimateMessagesInputTokens` is called with neither. The code comment names both under-counts explicitly, so the limitation is visible at the point of use.

This PR is therefore ready to merge once roborev has re-reviewed 7b35f8b98e.

PR #1236, opened 2026-09-13
fix(agent): hold the drive claim across the delegate send start (#940)

### LINKED ISSUE #940: delegate send start has the same committed-start window #932 closed for attention starts
The committed-start window that #932 closed for delegate attention starts also exists on the delegate `send` start path, and is not yet closed there.

## Background

#932 fixed a race in `driveStableDelegateAttention` (`agent/delegate_runtime.go`): between `CommitStart` and `resetSubagentForRunLocked` the child read idle to every guard (`sub.running` false, the reservation consumed, the attention id already resolved), so a wake-edge drive fell through `driveChildIfNotStopGated` to `driveSubagentNotificationTurn` and launched a second, unleased `EntryNotification` turn on the same session. Two turns then shared one drain ladder, and `popFollowUp` (a destructive pop with no owner check) let the unleased turn steal the run's follow-up. The fix holds `sub.driving` across the whole start and hands it to the run under the same `sub.mu` hold that sets `running`.

## What is still open

`delegateRuntime.send` (`agent/delegate_runtime.go`, the stable-send start around the `CommitStart` → `restoreIdleForSend`/`AttachRuntime` → `BeginStartInput` → `preseedInput` → `CompleteStartInput` sequence) has the same shape: `CommitStart` runs, then a durable transcript append and read-back and the start-input mutation plans, all with `sub.running` still false, on a retained or freshly reconstructed child whose `notifyFunc` is already `driveChildIfNotStopGated`. A wake landing in that window can start the same unleased notification turn on the child.

Not touched in #932 because no failure has been observed on this path and its rollback shape (`failStableSendStart*` / `stableSendFailureOutcome*`) differs at every exit, so the minimal attention-side fix did not extend to it cleanly.

The `create` path is not exposed: `newSubagent` builds the record with `running: true` before it is tracked, so the child is never publicly idle.

## Suggested fix

Same idea as #932: take the drive claim atomically with the send path's drivability check and hold it until the run takes ownership (or release it on each failure exit). A regression test can reuse the `delegateAttentionStartCommitted`-style seam pattern: fire a `driveSubagentNotificationTurn` from a seam placed after `CommitStart` in the send path and assert the child refuses it.

## Evidence

This was identified during the #932 investigation (`agent/delegate_runtime.go` read-through; the attention-side reproduction injected one `sub.sess.notify()` between `CommitStart` and the launch and reproduced the reported snapshot verbatim). The send-side window is inferred from the same code shape; it has not been reproduced.

obra:
Fixes #940.

`delegateRuntime.send` ran `CommitStart`, restore, and the start-input mutation plans with the child still reading idle, so a wake-edge drive fell through to `driveSubagentNotificationTurn` and launched a second, unleased `EntryNotification` turn on the session that generation was about to run. The two turns shared one drain ladder, and a destructive `popFollowUp` let the unleased turn steal the run's follow-up.

This takes a claim keyed by CHILD SESSION ID, immediately after `CommitStart` and before `restoreIdleForSend` resolves a child. That is the piece a per-child flag cannot cover: an id-keyed claim makes every drivability read refuse without needing the child object, so the `CommitStart` -> `restoreIdleForSend` -> `admitReconstructed`/`AttachRuntime` stretch is covered rather than merely narrowed. The per-child `sub.driving` handoff stays as it was; the id claim is released only after the run sets `running` under the child's own lock, so no gap opens between release and ownership. A second committed start racing on the same delegate is refused as busy.

Evidence:
- New `TestDelegateResourceSupervision_EarlyCommittedSendStartRefusesASecondTurn` drives the child from a seam immediately after `CommitStart` and asserts the second turn is refused. Verified load-bearing by mutation: removing the gate from the three read sites fails it.
- `go test ./agent/ -run TestDelegateResourceSupervision -count=1` green; `-count=5 -race` green (66s); `go build ./...` clean; `golangci-lint run ./...` 0 issues.

Reviewed and re-verified independently by the parent, not only by the implementing agent.

Residual, stated plainly: the `ReserveStart` -> `CommitStart` stretch is not covered. That is pre-commit, so a drive there races a reservation rather than a committed generation, and the issue scopes its window from `CommitStart`; recording it rather than implying the whole send path is fenced.


obra:
Holding this one open rather than merging. Roborev's High finding ("uncovered window before the drive claim is acquired") matches an independent adversarial review and my own read: the claim starts only once `restoreIdleForSend` has produced the child to claim, so the `CommitStart` -> `restoreIdleForSend` and `admitReconstructed`/`AttachRuntime` stretches remain uncovered, and a wake landing there can still start a second unleased turn. The issue text names exactly that stretch, so this PR narrows the race without closing #940. The code comment was corrected to stop claiming the whole window.

Closing it needs a claim keyed by delegate id, taken before the child is resolved. #940 stays open for that.

obra:
CI `tests` failed on the previous head; cause found and fixed, no merge until the rerun is green.

The failure was deterministic and ours: the repo requires every mutex on `Session` or `subagent` to be classified for envelope sampling (`TestEverySamplingRelevantMutexIsClassified`), and the new claim's `childCommittedSendMu` was unclassified. It now sits in the forbidden set with a `hold` closure, since envelope sampling reads no part of the committed-send claim and must never block on it.

A second failure in the same run, `TestDelegateAttention_SettledResidentChildStartsExactAttentionGeneration`, did **not** reproduce: it passes in isolation, passes with the gate removed as well as with it, and the full agent package suite is green (243s) after the classification fix. That run had it waiting 15s against 0.4s in isolation, so it is load-sensitive. Recording it rather than hiding it: if it flakes again on CI it is a pre-existing timing sensitivity, not this change.

obra:
Declining the Medium on retry coverage, with evidence.

Roborev: "Rollback re-drive is attempted only once ... no later retry is guaranteed, leaving pending stable attention or watch sends stranded."

The retry is guaranteed by the drain path, not by the rollback:

- `kickDriveTreeWithFlush` calls `s.drivePendingStableDelegateAttention()` unconditionally at the top of every pass (`agent/session_jobtree_drain.go:1564`), and kickDriveTree runs on each drain pass.
- `drivePendingStableDelegateAttention` schedules its own retry while attention remains pending: `s.scheduleStableDelegateAttentionRetry()` (`agent/delegate_runtime.go:1032` and `:1037`, defined at `agent/session_attention.go:866`).
- Notifications are retried the same way: the drain's `drainPendingWatchSendsReport` calls `driveChildrenWithUndeliveredAttention()` (`agent/job_watch.go:4878`), which re-drives every live idle child with queued notifications or pending watch sends.

So a rollback re-drive refused because the child is momentarily finalizing or driving does not strand the wake. The wake stays registered, the next drain pass drives it, and the scheduled retry covers a child still blocked. The rollback's single attempt is deliberate: it closes the gap between releasing the claim and the next drain pass, it is not the only driver. Adding a second retry would duplicate a guarantee that already holds.

PR #1237, opened 2026-09-13
fix(agent): fingerprint repeated tool failures by semantic call signature (#829)

### LINKED ISSUE #829: Detect repeated tool failures by semantic call signature

obra:
Fixes #829.

The repeated-failure ledger keyed on raw argument bytes, so a free-text `intent` or a neutral/default field reset the streak. This adds a semantic failure fingerprint above the exact-call fast path: canonical JSON (key order and whitespace folded), `intent` dropped, and `description` dropped for the shell tool. The exact-call guard and `WithBreakerBypass` are unchanged.

Adversarial review caught a false positive in the first revision: it treated numeric zero as a neutral default, which folded read_transcript's explicit `offset_bytes=0` into the omitted form. Presence is meaningful there, because a present `offset_bytes` selects the retained-page operation while omitting it selects the default view, which `TestNormalizeRetainedReadArgsPreservesExplicitZeroOffset` already pins at the tool layer. That is precisely what the issue's false-positive control forbids ("different offsets must remain distinct when they can change the outcome"), so numeric zero is no longer neutral, matching the existing treatment of boolean false.

Evidence:
- `go test ./agent/internal/tool/` green.
- `TestFailureFingerprint_ExplicitZeroIsDistinct` fails when zero-neutrality is restored (all three explicit-zero cases collapse to the omitted fingerprint) and passes here.
- Intent-only and neutral-default variants still park; a meaningful ref change still gets a fresh attempt; alternating fingerprints keep both runs; the 20-call shape from the issue is covered by the intent-only and neutral-default tests.

Trade-off: a value-based rule cannot know a schema's numeric defaults, so an explicit numeric zero that equals a field's real default no longer collapses. That direction only delays the breaker; the other direction would refuse a call the model meant to change.

### DIFF
--- a/agent/internal/mcp/breaker_iserror_test.go
+++ b/agent/internal/mcp/breaker_iserror_test.go
@@ -84,7 +84,7 @@ func TestMCPBreaker_IsErrorFailuresParkAtThird(t *testing.T) {
 	if !third.IsError {
 		t.Fatalf("call 3: IsError=false, output=%q", third.Output)
 	}
-	want := "evener did not execute this call: s__probe with these exact arguments has now failed 3 times with the same error; it will not be executed again until you change the arguments or the approach."
+	want := "evener did not execute this call: s__probe with equivalent arguments has now failed 3 times with the same error; it will not be executed again until you change the arguments or the approach."
 	if !strings.HasPrefix(third.Output, want) {
 		t.Errorf("call 3 park text = %q, want prefix %q", third.Output, want)
 	}
--- a/agent/internal/tool/breaker.go
+++ b/agent/internal/tool/breaker.go
@@ -170,19 +170,20 @@ func canonicalNumber(n json.Number) any {
 }
 
 // isNeutralDefaultValue reports whether a canonicalized value means "omitted":
-// null, the empty string, an empty array or object, or numeric zero. Boolean
-// false is deliberately not neutral — a schema's default may be true, so false
-// can be a meaningful instruction.
+// null, the empty string, or an empty array or object. Boolean false and
+// numeric zero are deliberately not neutral, because a value alone cannot say
+// whether a schema treats it as the default. For read_transcript an explicit
+// offset_bytes=0 selects the retained-page operation while omitting it selects
+// the default view (session_tools_transcript.go), so folding zero into
+// "omitted" would hide a meaningful correction and could park a call the model
+// legitimately changed. Erring toward a distinct fingerprint only delays the
+// breaker; erring toward a false match refuses a call the model meant to change.
 func isNeutralDefaultValue(v any) bool {
 	switch x := v.(type) {
 	case nil:
 		return true
 	case string:
 		return x == ""
-	case int64:
-		return x == 0
-	case float64:
-		return x == 0
 	case []any:
 		return len(x) == 0
 	case map[string]any:
--- a/agent/internal/tool/breaker.go
+++ b/agent/internal/tool/breaker.go
@@ -82,11 +82,12 @@ func exactSignature(name string, args []byte) string {
 // failureFingerprint returns the ledger key for a dispatch's repeated-failure
 // run. Unlike exactSignature it hashes a normalized view of the arguments:
 // free-text fields no tool executes on (intent, and the shell tool's
-// presentation-only job description) are dropped, JSON key order and
-// whitespace are canonicalized, and null/empty/zero-valued fields that mean
-// "use the default" are dropped. A call that changes only those fields is the
-// same failing operation, while a change to a meaningful field (target ref,
-// mode, offset, regex) keeps its own fingerprint and bounded history.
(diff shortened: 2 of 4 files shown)


obra:
Not merging yet. Roborev found two additional Mediums in the same family as the numeric-zero false positive this PR already fixes:

- **Meaningful empty values conflated with omission** (`breaker.go:134-143`): the normalizer still drops every `null`/empty string/empty array/empty object regardless of tool semantics, so e.g. `task_list.update.depends_on: []` (which clears dependencies) folds into omission. This is the same class as `offset_bytes: 0`, and it confirms the value-based neutralization is the wrong layer: deciding neutral-vs-meaningful is schema-dependent, so it needs a tool/schema-aware allowlist rather than a value predicate.
- **Canonicalization parses before size/UTF-8 validation** (`breaker.go:105-123` vs `registry.go:661-685`), defeating the pre-parse boundary and allowing oversized valid JSON to allocate.
- **Low**: `failureParkText` still says "these exact arguments" although parking is now semantic.

The Low is a straight wording fix. The two Mediums are design-level for the fingerprint and I have not attempted them here.

obra:
Two Low findings accepted as follow-ups; merging on that basis.

1. **`null` vs `{}` produce different semantic fingerprints** (`canonicalArgumentBytes`). A root `null` body canonicalizes to `null` while the dispatch path treats it as `{}`, so a loop alternating the two is treated as two operations and never accumulates a semantic run. Narrow (the exact store still catches byte-identical repeats) but real, and the fix the review names is small: after decoding, return `{}` when the root value is nil. Worth doing with a test asserting `fp(tool, "null") == fp(tool, "{}")`.
2. **Dead exact-store failure state** (`record` calls `observeFailure` on the exact entry and discards the result; `check` reads only `bodyCount` from it). Every record therefore runs `errorClass` (SHA-256) and `TruncateRunes` for an accumulator nothing consumes, and the comment implying it preserves byte-identical history is misleading, since byte-identical calls already share the semantic fingerprint.

Also deferred from the previous round, in the order I would take them: the LRU-hygiene Medium (semantic entries created on success, which can evict a pending failure run, a delayed-park direction), `failureFingerprint` being computed twice per dispatch (roughly four extra decode/canonicalize/marshal passes over a body up to 2 MiB, even when nothing fails), and `canonicalNumber` converting through `float64` (large integers lose precision).

PR #1238, opened 2026-09-13
fix(web): keep a summary-less failed delegate row single-level (#1173)

### LINKED ISSUE #1173: Web UI: failed delegate tool call renders an empty summary line (regression from #1170)
## Summary
A `delegate` tool call that failed renders an empty summary-line region at chat/intent verbosity, with the body disclosure chevron moved onto it from the intent line.

## Cause
#1170 made a failure force-open carry the summary line open (`ToolCallItem.tsx`: `summaryFallback = summaryConfigDefault || (failed && !superseded)`). Delegate rows deliberately render no summary text (`summary={isDelegate ? "" : summary}`), so opening the summary region produces an empty line.

## Reproduction
Render a `ToolCallItem` for `toolName: "delegate"` with a `description` and `error` set, at chat (or intent) level.

Observed DOM at chat level:
```
<div class="summaryLine" data-body-trigger="true">
  <button class="bodyTrigger" data-testid="tool-row-body-trigger" aria-expanded="true" aria-label="Failed">...chevron...</button>
</div>
```
`data-body-trigger-intent` is absent and `tool-row-summary` is null. A non-failed delegate at the same level instead carries the body trigger on the intent line (`data-body-trigger-intent="true"`, no summary line).

## Impact
Cosmetic: a failed delegate row shows a stray empty line. The same empty-line pattern already existed for delegate rows at tools/activity/full (where the summary defaults open); #1170 widened it to failed delegates at chat/intent.

## Suggested fix
Don't force the summary open for a row with no summary text: exclude delegate rows from the failure summary-open, or have `ToolRow` treat a summary-less intent-bearing row as single-level.

Files: `cmd/evener-hub/frontend/src/panes/session/transcript/ToolCallItem.tsx`, `cmd/evener-hub/frontend/src/panes/session/transcript/ToolRow.tsx`.

obra:
Fixes #1173.

A failed delegate row has an intent but no summary text, and `summaryOpen` was forced true at chat/intent level, so `ToolRow` took the two-level path and rendered an empty summary line, pushing the body chevron off the intent line. A summary-less intent-bearing row now stays single-level.

Evidence:
- `ToolCallItem.test.tsx` adds the chat-level failed-delegate case; it fails on the pre-fix tree (2 failed / 88 passed) and passes here (90 passed).
- `npx vitest run src/panes/session/transcript` -> 79 files, 1881 tests passed; `tsc --noEmit` clean; `biome check` clean on both touched files.

Note: the same empty-line class exists at `tools/activity/full`, which predates this regression; left out of scope.

### DIFF
--- a/cmd/evener-hub/frontend/src/panes/session/transcript/ToolCallItem.test.tsx
+++ b/cmd/evener-hub/frontend/src/panes/session/transcript/ToolCallItem.test.tsx
@@ -1568,3 +1568,35 @@ test("a summary-only job_watch clear renders a non-expandable row (hasBody predi
   );
   expect(screen.queryByTestId("tool-row-body-trigger")).toBeNull();
 });
+
+// --- #1173: a summary-less intent-bearing row stays single-level on failure ---
+// A delegate row deliberately renders no summary text (subagentModule owns its
+// presentation), so the failure force-open must not carry an empty summary line
+// open. The body disclosure chevron stays on the intent line
+// (data-body-trigger-intent), exactly as it does for a clean delegate row.
+test.each(["chat", "intent"] as const)(
+  "a failed delegate row at the %s level renders no empty summary line (body trigger on the intent line)",
+  (level) => {
+    const config = makeTranscriptDisplayConfig({ kind: "preset", level });
+    const { container } = renderWithConfig(
+      config,
+      item({
+        id: `delegate_fail_${level}`,
+        toolName: "delegate",
+        description: "Delegating the flaky suite",
+        argumentsJSON: JSON.stringify({ prompt: "Run the flaky suite" }),
+        error: "delegate activation failed",
+      }),
+    );
+    // The body force-opens (only failure earns the eye, unchanged).
+    expect(screen.getByTestId("tool-row-body-trigger").getAttribute("aria-expanded")).toBe("true");
+    expect(screen.getByTestId("tool-call-body")).toBeTruthy();
+    // No summary text exists for a delegate row, so no summary line (empty or
+    // otherwise) renders and no summary-open region is forced.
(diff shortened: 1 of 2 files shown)


obra:
Holding this open. I verified roborev's Medium 1 (`summaryConfigDefault` bypasses `hasSummaryText`, so a delegate row still opens the summary region at tools/activity/full) and implemented the suggested gate, then reverted it: with the gate, `src/panes/session/transcript/tools/subagentModule.test.tsx` fails ("stable delegate attention and lifecycle own the card...", 1 failed / 40 passed), and it passes 41/41 with the shipped code. So for a delegate row the summary region is not simply empty - the delegate card's rendering depends on it, and the two findings are in direct conflict.

Medium 2 (base the intent-less check on the derived `intent`) also fails: switching to `intent === undefined` breaks `TurnBlock.test.tsx` "an intent-less tool shows its own derived summary at chat and tools levels" and "an intent-less read_file shows the Read summary", which is the intended behavior for non-delegate intent-less rows.

Both need a proper look at how the card couples to the two-level layout rather than a gate change. The shipped commit does fix the regression #1173 reported at chat/intent verbosity.

obra:
Merged on the maintainer's call, landing the regression fix and splitting the rest out.

The two medium findings are real observations, but their proposed fixes are demonstrably wrong as stated, which is why I reverted them rather than shipping a regression:
- gating the summary region on `hasSummaryText` breaks `subagentModule.test.tsx` (1 failed / 40 passed with the gate; 41 passed without), because the delegate card's rendering depends on that region;
- switching the intent check to the derived `intent` breaks two `TurnBlock.test.tsx` tests that pin the intended behaviour for non-delegate intent-less rows.

Both are now tracked in #1253 with the evidence and the direction (decouple the card from the summary region rather than gating it away).

PR #1241, opened 2026-09-13
refactor(sdk): move the AppWire TypeScript package to appwire-client/typescript (SDK A3)

### LINKED ISSUE #1224: qualify-package.mjs: a module built into dist but missing from shippedModules passes in silence
Found while landing SDK row C13 (#1223).

`scripts/qualify-package.mjs` derives its reachability assertion and the tarball's expected-file list from the hand-maintained `shippedModules` array. The tarball listing check only asserts that every `shippedModules` entry is present and that every entry lives under `package/dist/` or `package/examples/`. So a module added to `protocol/tsconfig.build.json` `files` (and therefore built into `dist/`) but never added to `shippedModules` ships in the tarball, is checked by nothing, and the gate stays green. That is exactly the silent-unreachable-module hazard the SDK plan's phase-C rule warns about, and the runner is supposed to be the thing that catches it.

Reproduced on C13: with `activityRows.ts` in `tsconfig.build.json` and re-exported from `index.ts` but absent from `shippedModules`, `make test-api-package` passed. Only after adding the `shippedModules` entry did deleting the re-export fail with `shipped module unreachable from every published specifier`.

Fix (small): derive `shippedModules` from `tsconfig.build.json` `files` (minus `index.ts`), or assert that the set of `package/dist/*.js` in the tarball listing equals `shippedModules` exactly. Either way one list, no second write site.

Sequencing: lands after #1222 and #1223, which both edit the manifest.

### LINKED ISSUE #1242: Two AppWire package test files import the web app
`appwire-client/typescript/tokenFlood.test.tsx` and `appwire-client/typescript/slashCompletion.test.ts` import out of the package and into `cmd/evener-hub/frontend/src/`:

- `tokenFlood.test.tsx` — six imports: `panes/session/composer/queue/pendingTurnsStore`, `panes/session/Session`, `panes/session/transcript/types`, `shell/clientContext`, `stores/connection`, `stores/threads`. It renders the web `Session` component under a token flood.
- `slashCompletion.test.ts` — one type import: `shell/palette/commands`'s `ScopedCommand`.

The coupling predates SDK A3; A3 (#1241) only repointed the seven specifiers at the directory's new depth, so they now read `../../cmd/evener-hub/frontend/src/...` out of a package that is supposed to be app-independent. Every other file in the package imports only its own siblings.

Two ways out, both out of scope for A3:

1. Move both files into `cmd/evener-hub/frontend/src/` — they are web integration tests that happen to live in the package. `tokenFlood.test.tsx` would sit next to the session pane; the package keeps `testing/tokenFlood.ts`, which is genuinely package-side. This also shrinks the count the new `scripts/package-test-files.mjs` check watches.
2. Invert the dependency so the package's test drives a seam the app implements, the way `docContent` took a `DocPort` in A3d.

(1) looks right: nothing about these two tests belongs to the shipped package, and the package's tarball never contained them.

### LINKED ISSUE #1243: An untested AppWire package module vanishes from the web coverage denominator
`cmd/evener-hub/frontend/vite.config.ts`'s coverage block names the whole source tree in `coverage.include` on purpose: Vitest reports only the files a test actually loaded, so a subsystem with no test at all would score as ABSENT rather than as the 0% it is.

After SDK A3 (#1241) moved the package to `appwire-client/typescript/`, that guarantee holds for `src/` but **not** for the package. `coverage.allowExternal: true` plus an absolute `include` glob does count package files that a test loads — 28 of the package's 29 source modules appear, and the measured total is 96.79% lines against a `web 96.1` floor — but the "enumerate every matching file" half of `include` does not reach outside the Vitest root. Measured: running a single package test file (`errors.test.ts`) with the same config yields exactly **one** package file in `coverage-summary.json`, not 29.

So a new module added to `appwire-client/typescript/` with no test is invisible to the web coverage row instead of scoring 0%. That is the same false green the Go side guards with its gap map, and the reason the `include` line exists at all.

Options:
- A dedicated Vitest project rooted at the repo root for the package, with its own coverage include.
- A cheap structural check next to `scripts/package-test-files.mjs`: assert every package source module appears in `coverage-summary.json`, which fails the moment one is untested.
- Relative-path globs plus a `coverage.root` at the repo root, if Vitest supports it — the relative form was tried during A3 and matched nothing.

(`clientLike.ts` is the one module absent from the report today; it declares types only and emits no statements, so it is not an instance of this bug.)

obra:
## Answered findings

*Read this before re-raising. Each line was measured, not argued.*

**"The Vite aliases and the Metro resolver branch are untested, because nothing imports the package by name yet."** True and deliberate: A4 is the consumer. It rewrites both apps' imports to `@evener/appwire-client`, and its own gates are what exercise these resolvers for the first time — `make test-web` and `make test-web-browser` for the Vite alias (web typecheck, Vitest and all five browser guards resolve through it), `make test-native-bundle` (#1245, closing #1244) for Metro, which is the first gate in the repo that bundles the native app at all. A3's job is that the aliases exist and are correct before A4 can land; each one was still checked here by hand — a throwaway test importing all three specifiers passed under both Vitest configs and both `tsc` programs, and `metro.config.js`'s branch resolves all three to files that exist. A comment above the Metro branch names its gate.

**"`vite.config.ts` aliases React and Testing Library but not `vitest`, so the package's tests cannot resolve it."** They can, and adding the alias would be cargo cult. Measured three ways at round 2: Vite's own resolver returns **UNRESOLVED** for `vitest` from `appwire-client/typescript/errors.test.ts` (the same probe resolves `react` and `@testing-library/react` through the aliases this PR adds), so Vite is not what resolves it — vitest externalises its own module and the worker imports it from the runner, independent of the importer's directory; with **both** aliases deleted, `errors.test.ts` and `submitRouting.test.ts` still run green (2 files / 44 tests); and deleting only the `react` alias is what actually breaks something (`tokenFlood.test.tsx`, at transform time). The pin that keeps this answered: `scripts/package-test-files.mjs` executes the first package test that imports `vitest` through the JSON reporter and checks the summary — the file must appear in the report, with tests, none failing. A regression here fails `make test-web`, so the alias is not the thing standing between the package's tests and green.

Moves the AppWire TypeScript package from `cmd/evener-hub/frontend/src/protocol/` to a new top-level `appwire-client/typescript/`.

Plan row: A3 in docs/superpowers/plans/2026-09-12-sdk-migration.md (#1182); decision 3 settled 2026-09-13.

## Commits

1. **`a0a137c68` — the move.** `git mv` only, 81 files, no content edits.
2. **`15ee5837f` — build, gate and CI paths.** Four resolvers are independent and a green `tsc` proves none of the other three: tsconfig `paths` (web typecheck), `resolve.alias` in `vite.config.ts` (bundler, dev server, Vitest, all five browser guards), `vitest.config.mts` alias + `tsconfig.check.json` paths (mobile-native), and a `resolveRequest` branch in `metro.config.js` (native bundle — Metro reads neither tsconfig paths nor the package's `exports`). Plus `make/testing.mk:58`, `make/linting.mk:190`, the matching fixture in `makefiletargets_audit_test.go:1073`, and the three CI lines (`ci.yml:57,60`, `ios-testflight.yml:74`).
3. **`ceb6ad02a` — the seam.** 32 one-line re-export stubs at the old `protocol/` path carry the web's 689 import lines (25 modules were named by web imports plus `types.gen`, `clientLike` and five under `testing/`); SDK A4 deletes the directory. Three seds move the mobile trees' 166 lines onto the new path — 138 `mobile-native` `.ts/.tsx`, 22 `mobile/src`, 6 `.mts` — which is also what the `.mts` carve-out requires. `grep -rn 'cmd/evener-hub/frontend/src/protocol/' mobile mobile-native` now returns zero. Two of the package's own test files reach back into the web app (`tokenFlood.test.tsx` renders `Session`, `slashCompletion.test.ts` takes a `ScopedCommand` type); their seven app imports follow the directory to its new depth. That coupling is pre-existing — filed as #1242.
4. **`450899b16` — generator, Go strings, cards, docs.** `appwire/doc.go:29`'s `go:generate -out` is the producer; the two drift tests read the file back from the new path. **14** scenario-card citations across nine cards (12 anchored, and those 12 are what `TestScenarioSourceCitationsResolve` enforces — it resolves by suffix, and `protocol/` was a segment in every one). Six comment-only Go mentions and eight doc citations updated for accuracy, including all of `docs/web-ui` (`parity-m6-surfaces.md:166`, `decisions.md:345,474,533`, `specs/2026-09-03-keybinding-system-survey.md:76,91`). `docs/superpowers/{plans,specs}` deliberately untouched.
5. **`39cc6f4ac` — the three risk fixes** (below).
6. **`d32a1d479` — a gate assertion** that the package's test files are still collected.

## The three risk fixes, with evidence

**R1 — `server.fs.allow`.** `searchForWorkspaceRoot` resolves to the frontend directory itself (verified in this checkout: there is no root `package.json` and Vite's `ROOT_FILES` no longer includes `.git`), so the package lands outside the allow list and every `/@fs/` request for it 403s in `npm run dev` and in all five guards, while `tsc`, `vite build` and `vitest` stay green. Fixed in two files, because `scripts/editorial-preview.vite.config.mjs` *overwrites* the list twice rather than extending the base config. Measured through the guards' own Vite config (`scripts/browserguard.vite.config.mjs`), fetching a package module over `/@fs/`:

| | `GET /@fs/<repo>/appwire-client/typescript/errors.ts` |
| --- | --- |
| without the allow entry | **403** |
| with it | **200** |

`editorial-preview.test.mjs` asserted the exact `fs.allow` list, so it is updated in the same commit; its sentinel test still proves everything outside those two directories stays denied.

**R2 — the package's 27 test files.** Vitest collects relative to its root, so the move drops all of them and `make test-web` goes green having run 27 fewer files. `vitest list --filesOnly`:

| | test files collected | of them, under the package |
| --- | --- | --- |
| without the `test.include` entry | 430 | **0** |
| with it | 457 | **27** |

They run from the web config rather than under `make test-api-package`: they import `vitest`, `react` and `@testing-library/react` bare and load their `.jsonl` fixtures through Vite's `?raw`, all of which this app's install already provides, and giving the package its own dev dependencies would change what `npm ci --prefix appwire-client/typescript` fetches for the qualification gate, which needs only `typescript` and `ws` — the offline tarball install is unchanged. Node resolution from the package walks up to a repo root with no `node_modules`, so those three bare names are aliased in `vite.config.ts` and named in `tsconfig.json`'s `paths`. The package also joins the typecheck's `include`, because its test files are imported by nothing and would otherwise leave the program entirely (R8).

Commit 6 makes the drop impossible to ship quietly: `scripts/package-test-files.mjs` compares what `vitest list` collects under the package against what is on disk there, runs first in `npm test`, and fails with the names of the files that went missing. With the `include` entry removed it reports `Vitest collects 0 of the package's 27 test files` and exits 1.

**R3 — lint scope and the coverage floor.** Biome's scope follows the package (`biome ci src ../../../appwire-client/typescript`), with the two `src/protocol` ignore entries repointed so the generated `types.gen.ts` and the reducer snapshot stay out of the lint set — a `../`-relative pattern in `files.includes` does not match, a `**/`-anchored one does (verified: 5 errors → 0). `AGENTS.md`'s statement of the scope moves with it. Coverage keeps the package in the denominator via `coverage.allowExternal` plus an absolute `include` glob, so `web 96.1` still describes the same file set and is **not** re-baselined: measured **96.79%** lines (20628/21312) with 28 of the package's 29 source modules in the report. `clientLike.ts` is types-only and emits no statements.

## Gates

| Gate | Result |
| --- | --- |
| `make test-web` | PASS — typecheck, 451 files / 10413 tests, Biome |
| `make test-web-browser` | PASS — layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard |
| `make test-native` | PASS — `npm test`, `test:shared` (7 files / 777 tests), `npm run check` |
| `make test-api-package` | PASS — `qualified @evener/appwire-client@0.1.0` |
| `make lint` | PASS — including `lint-generated` |
| `WEB=0 make test` | PASS — all Go modules, including the scenario-citation and makefile-targets audits |
| `make build-web` | PASS |
| `make generate` | writes `appwire-client/typescript/types.gen.ts`; no diff |

Resolution was proved per resolver rather than inferred: a throwaway test importing `@evener/appwire-client`, `/docContent` and `/testing/fakeClient` passed under both the web and mobile-native Vitest configs and both `tsc` programs, and `metro.config.js`'s branch was exercised directly and resolves all three specifiers to files that exist.

`#1224` is untouched: no change to `shippedModules` or the manifest's semantics.

**Residuals filed:** #1242 (two package test files import the web app) and #1243 (an untested package module vanishes from the web coverage denominator).

## Review round 1

`09a07575f` — the one Low finding, real: `scripts/package-test-files.mjs` ran `npx vitest list`, and `npx` falls back to the npm cache or the registry when `node_modules` is missing or stale, which a default gate must not do. It now spawns `node_modules/.bin/vitest` directly and refuses with a one-line message naming the `npm ci` when that binary is absent — the same hazard `scripts/web/web-preflight.sh` guards by asking `./node_modules/.bin/tsc` for its version rather than a bare or npx `tsc`, which resolves to the unrelated `tsc@2.0.4` package. The comment says so at the call site.

Merged `origin/main` at `eeff54b70` (`2fb3d5fa2`). Re-ran `make test-web` (PASS — typecheck, tests incl. the count assertion, Biome) and `make lint` (PASS, incl. `lint-generated`).

## Review round 2

`5062edf34`, one commit for all three findings.

**1 — the `vitest` alias (Medium): refuted, and pinned so it cannot come back.** `vitest` resolves for the package's external test files with no alias at all. Measured two ways: Vite's own resolver returns **UNRESOLVED** for `vitest` from `appwire-client/typescript/errors.test.ts` (while `react` and `@testing-library/react` resolve through the aliases this PR adds), so Vite is not what resolves it — vitest externalizes its own module and the worker imports it from the runner, independent of the importer's directory; and with **both** aliases deleted, `errors.test.ts` and `submitRouting.test.ts` still run green (2 files / 44 tests). The two aliases that are needed are exactly the two that are there: dropping the `react` one makes `tokenFlood.test.tsx` fail to resolve `@testing-library/react` at transform time. The pin: `scripts/package-test-files.mjs` now executes the first package test that imports `vitest` through the JSON reporter and checks the summary — the file must appear in the report, with tests, none failing — so "collected" is no longer the whole claim.

**2 — the run-as-main guard (Medium): real.** It compared the percent-encoded `import.meta.url` against a hand-built `` `file://${process.argv[1]}` ``, so in a checkout path containing a space the body never ran and `make test-web` passed without asking vitest anything. Now `pathToFileURL(process.argv[1]).href`, the idiom in `mobile-native/scripts/demo-hub.mts:400`.

**3 — `URL.pathname` in the test (Low): real**, same cause; `fileURLToPath` now.

Proved in a detached worktree checked out at `…/space check/wt`: the old script exits 0 with no output, the new one prints `vitest collects all 27 AppWire package test files, and activityData.edge.test.ts executes and passes`, and its 13 unit tests (6 new) pass there too.

`main` unchanged at `eeff54b70`, so no merge. Gates: `make test-web` PASS (typecheck, tests incl. the count and execution assertions, Biome), `make lint` PASS (9/9).

## Review round 3

`270e7d784` (item 1) and `55b10571e` (items 2-5).

**1 — the two tests that imported the app (Medium): fixed here, closes #1242.** `tokenFlood.test.tsx` moves to `cmd/evener-hub/frontend/src/panes/session/`, next to the `Session` it renders, on the deep `../../protocol/…` specifiers its neighbours use (A4 rewrites them with the rest of the web tree); its bench sibling stays in the package, where it drives only the package's own flood harness. `slashCompletion.test.ts` restates the three fields `mergeSlashCommands` reads — the same three `slashCompletion.ts` already declares structurally — instead of importing the palette's `ScopedCommand`. The property is now enforced, not merely true: `scripts/package-test-files.mjs` sweeps every source file under the package and fails with each offending line if one imports from `cmd/evener-hub/`. Proved by adding such an import to `errors.ts`: `the AppWire package must not import from the app: errors.ts: import type { PaletteContext } …`, exit 1. A comment that merely names an app path does not trip it (unit-tested). The package's test-file count drops to **26** and the gate reads it from disk, so nothing constant changed.

**2 — `test:coverage` (Low):** both test scripts now call one `check-package-tests` definition; the coverage run could not previously have caught a silent drop.

**3 — path normalisation (Low):** the executed-test comparison resolves both sides and compares relative to the package directory, so a symlinked worktree or `/var` against `/private/var` cannot report a test that ran as one that never did.

**4 — `editorial-preview-install.mjs` (Low):** its header described `fs.allow` as exactly `[frontend]`; it now names the AppWire package as the one intentional entry outside the checkout, matching the config and its test.

**5 — stale comments (Low):** `testing/tokenFlood.ts:85` and `reducer.test.ts:518` repointed off `src/protocol/`.

The gate's unit tests grew from 13 to 18. `main` unchanged at `eeff54b70`, so no merge. Gates: `make test-web` PASS (typecheck, tests incl. the no-app-import sweep, the 26-file count and the executed-test pin, Biome), `make test-web-browser` PASS (all five guards), `make lint` PASS (9/9).

## Merge of #1070

`c7918bdff` merges `origin/main` at `42d27f946` (shared session notes). #1070 edited five package files and added two more at the old `protocol/` path, where this branch now keeps only re-export stubs — so rename detection matched on path and conflicted instead of carrying the edits to the package's new home. In every case the old path keeps its stub and #1070's content lands under `appwire-client/typescript/`.

| Conflict | Resolution |
| --- | --- |
| `protocol/model.ts` | This branch had not touched its content, so the package takes main's version whole. |
| `protocol/reducer.ts` | Same. |
| `protocol/testing/notifications.ts` | Same. |
| `protocol/testing/tokenFlood.ts` | Merged by hand, both sides kept: main adds `sharedNotes: true` to the fixture capabilities, this branch had repointed a path in a comment. |
| `protocol/types.gen.ts` | Regenerated, never hand-edited: `make generate` writes it to the package, and the result is **byte-identical** to main's generated file. |

Two files are new in #1070 and were added at the old path by the merge: `sharedNotesAvailability.ts` and `reducer.notes.test.ts`. Both move to the package; the four web files that import `protocol/sharedNotesAvailability` (`NotesPanel.tsx`, `SessionChrome.tsx`, `shell/palette/commands.ts`, `RailRow.tsx`) get a re-export stub like every other module, so A4 deletes it with the rest. #1070's edits to five package test files and to `tokenFlood.test.tsx` in the web tree merged cleanly at their new locations — rename detection worked wherever this branch had left no file behind.

The package's test-file count goes 26 to 27 with `reducer.notes.test.ts`; the gate reads it from disk, so no constant changed, and the no-app-import sweep is clean. A repo-wide grep finds no `cmd/evener-hub/frontend/src/protocol/` reference outside `docs/superpowers/`.

Gates re-run on the merge: `make test-web` PASS, `make test-web-browser` PASS (all five guards), `make test-native` PASS, `make test-api-package` PASS, `make lint` PASS (9/9), `make build-web` PASS, `WEB=0 make test` PASS (all seven Go modules, including the scenario-citation and makefile-targets audits).

## Review round 4

`ebae961c2`, one commit for all three Lows; merge `e6f5e45fa` brings `origin/main` to `934a00db3` (the SDK plan and inventory docs, no conflict).

**1 — the Metro resolver branch is not exercised (Low): by design, and now said out loud.** Nothing imports the package by name yet — the native tree still uses relative paths, which resolve without that branch. SDK A4 is the change that rewrites those imports to `@evener/appwire-client`, and `make test-native-bundle` (#1245, closing #1244) is the first gate that bundles the native app and therefore exercises this branch; A4's brief requires it to pass on that rewrite. No new test here. A comment above the branch in `metro.config.js` names that gate, so the next reader finds the answer at the code rather than in this thread.

**2 — stale path strings (Low): fixed.** Two `FakeClient` runtime messages and one in `reducer.test.ts` told the reader to look in `protocol/types.gen.ts`, a directory that now holds only stubs. Repointed, along with the seven sibling references in comments the same grep turned up (`activityData.ts`, `askShared.ts`, `deriveAskQuestions.ts`, `errors.ts`, `reconcileBatches.ts`, `submitRouting.test.ts`, `testing/fakeClient.ts`). The package now contains no `protocol/` path outside `types.gen.ts`'s own generated content.

**3 — the sweep matched only `from "…"` (Low): widened.** A side-effect `import "…"`, a `require("…")` and a dynamic `import("…")` reach into the app just as effectively and all passed unseen. One pattern now covers four forms, keyed on the leading keyword — which is also what keeps a comment or a plain string literal mentioning an app path from tripping it. Each form has a unit test (19 now, up from 18), and all three were planted in the package's own `model.ts` to confirm end to end: the sweep named all three lines and exited 1.

Gates: `make test-web` PASS (sweep, 27-file count, executed-test pin, typecheck, Biome), `make lint` PASS (9/9).

## Review round 5

`eb3ebfeaf` (item 2) and `578855bb2` (item 3); merge `f711a3b5f` brings `origin/main` to `d4b3ec4e7`, no conflict. Item 1 is the "Answered findings" section at the top of this body — no code change, and the alias stays out.

**2 — `sharedNotesAvailability.ts` had none of the package's plumbing (Medium): real, fixed.** #1070 added it and the merge moved it in, but it was absent from `tsconfig.build.json`'s `files`, from `index.ts`, from `shippedModules` and from the manifest, so the package never shipped it and the four web callers reached it only through the A3 stub that A4 deletes. Read first, as instructed: it imports nothing outside the package (one type from `./model`). `canReadSharedNotes` now joins the root exports and `rootValues`, with a smoke call per branch — capability true, false, absent, and no model at all. It exports no types. Falsified twice: dropping the root export while the manifest still names the module fails with `shipped module unreachable from every published specifier: sharedNotesAvailability`; changing the implementation to `!== undefined` — the plausible wrong reading, which treats an explicit `false` as permission — fails the smoke call with `actual: true, expected: false`.

**3 — `testing/hubWireFixtures.ts` read through the CWD (Low): decided by usage, and guarded.** Both callers are web pane tests (`InstanceSheet.wire.test.tsx`, `credentialLabels.wire.test.ts`), so the module moves next to them at `panes/settings/sections/credentials/hubWireFixtures.ts`, where `join("..", "testdata", "authwire", "responses.json")` against the frontend's CWD is true and is now documented as the reason it lives in the app. Its A3 stub goes with it — **31 stubs, not 32**. The sweep gains the rule: a package file that imports `node:fs` (or `node:fs/promises`) and builds a `..` or `testdata` path must resolve it through `import.meta.url`. Unit-tested on the exact line this file used to carry, on the module-relative form that is fine, on a module that names such a path but never touches the filesystem, and on `node:fs/promises`. The gate's unit tests go 19 to 23.

Gates: `make test-web` PASS, `make test-api-package` PASS, `make test-web-browser` PASS (all five guards), `make lint` PASS (9/9).

## Review round 6

`8efc78900` (item 1) and `451a94d45` (item 3); merge `0d47eb84a` brings `origin/main` to `d2d4c6f3c`, no conflict. Item 2 is the second entry in "Answered findings" above — no code.

**1 — an unloaded package module is ABSENT rather than 0% (Medium): closed here, `Closes #1243`.** The preferred fix is not available: Vitest 4 removed `coverage.all`, and an `include` glob does not reach past the Vitest root in either the relative or the absolute form (both measured — with a single package test file running, an absolute include plus `allowExternal` yields exactly one package entry, not 29). So the guarantee is restored by assertion. `scripts/package-coverage.mjs` runs after `vitest --coverage` in `test:coverage` and requires every module the package's **own build** compiles — `tsconfig.build.json`'s `files`, so a module added to the build and to no test is precisely what it catches — and that emits runtime code, to appear as a file entry in `coverage-summary.json`; it fails with the missing names. Declaration-only modules are exempt because v8 has nothing to score for them, and the rule reads that off the source rather than a hand-maintained list (`clientLike.ts` is the only one today). Falsified: a scratch module added to the build and to no test fails with `1 AppWire package module(s) compile but never appear in the coverage report: scratchUntested.ts`, exit 1. Green run reports `all 29 compiled AppWire package modules are scored in the coverage report`; lines 96.77%, floor `web 96.1` unchanged. Eight unit tests cover the three pure functions, including the empty-expectation case — a check that measures nothing must not pass.

**3 — `describeCwdRelativeReads` flagged ordinary relative imports (Low): real, narrowed.** A line is now reported only when all three hold: it calls a filesystem API (`join`, `resolve`, `readFileSync`, `readFile`, `existsSync`, `createReadStream`, `stat`, `open`, …), it carries a `..` or `testdata` literal, and it is not an `import`, `export … from` or `require(`. The true positive is still pinned on the exact `join("..", "testdata", "authwire", "responses.json")` line `hubWireFixtures.ts` used to carry; the false positive is pinned too — a module with relative imports, a re-export, a `require` and an **absolute** `readFileSync` passes clean, and so does a `..` literal that no call reads. The gate's unit tests go 23 to 25, plus the 8 new ones for the coverage check.

Gates: `make test-web` PASS, `npm run test:coverage` PASS (with the new assertion), `make lint` PASS (9/9).

## Review round 7

`50dccb856` (item 1, `Closes #1224`), `d3142036e` (items 2-5) and `62b3b4128` (Biome's wrap of the line item 1 added, now that the package's scripts are in the lint scope); merge `d74d19e11` brings `origin/main` to `468ee7950`, no conflict.

**1 — `shippedModules` duplicated the build's `files` (Medium): closed, `Closes #1224`.** The runner now reads `tsconfig.build.json`, which is the definition of "code this package ships", so a module added to the build is qualified and reachability-checked without a second edit. The sets were identical at this commit, so nothing changes about what is qualified today — only that they can no longer drift. The finding's `except index.ts and declaration-only modules` shape is deliberately **not** taken: `index` is what the root specifier's declaration entry resolves to, and dropping `clientLike` would quietly remove a shipped module from the reachability check. Two assertions guard the derivation instead, so an empty or index-less `files` fails loudly rather than qualifying nothing. Falsified both ways: deleting `index.ts`'s re-export of `canReadSharedNotes` fails with `shipped module unreachable from every published specifier: sharedNotesAvailability`, and a module added to the build that no specifier reaches fails the same way with its own name — the case the hand-maintained list could not see at all.

**2 and 4 — the working-directory read rule (Medium, Low).** It looked for `from "node:fs"` alone, so `require("fs")`, `await import("node:fs/promises")` and `import fs from "fs"` all reached the filesystem unwatched; one pattern now covers four spellings and both module names, each with a unit test. A `./`-prefixed literal is CWD-relative in exactly the same way as `../` and is now matched too.

**3 — path canonicalisation (Medium).** The executed-test comparison resolves both sides with `realpathSync` before comparing, pinned with a symlinked temp directory. Worth recording: the first draft's `catch` swallowed every error, so a missing `realpathSync` import produced a *wrong verdict* instead of a crash — the new test caught exactly that, and the fallback now rethrows anything that is not `ENOENT`/`ENOTDIR`.

**5 — targeted coverage runs (Low).** `npm run test:coverage <file>` loads only what its filter reaches, so every other package module is legitimately absent; the check is skipped whenever the script is given arguments, and `--help` states the rule. Verified both ways: the full run reports `all 29 compiled AppWire package modules are scored in the coverage report` (lines 96.77%, floor `web 96.1` unchanged), and a targeted run over one package test exits 0 without it.

Gate unit tests: 28 for `package-test-files.mjs`, 8 for `package-coverage.mjs`. Gates: `make test-web` PASS, `make test-api-package` PASS, `npm run test:coverage` PASS full and targeted, `make lint` PASS (9/9).


obra:
Residual from this lane filed as #1244: nothing in CI bundles the native app, so Metro's resolution of the moved package is unverified by any gate; it lands before or with A4. #1242 and #1243 were filed by the implementer.

PR #1245, opened 2026-09-13
ci(native): bundle the app with Metro on every PR (#1244)

### LINKED ISSUE #1244: No CI gate bundles the native app, so Metro's resolution of @evener/appwire-client is unverified
Surfaced by SDK row A3 (#1241), which moves the package to `appwire-client/typescript/` and updates `mobile-native/metro.config.js` so the three specifiers resolve to files that exist (checked by hand on the branch). Nothing in CI bundles the native app: `make test-native` runs vitest and `tsc`, and the TestFlight workflow builds on demand only. A Metro resolver regression (a wrong alias, a missing `watchFolders` entry, a `node_modules` path that only works from one directory) would ship green and fail on device.

Wanted: a bounded gate that runs Metro's bundler in CI for the native entrypoint (`npx expo export` or `react-native bundle` in dry-run form, whichever the toolchain supports offline) and fails on unresolved modules, wired into `make test-native` or a sibling target. A4 (package-name imports across the native tree) raises the stakes, so this lands before or with A4.

obra:
Nothing in CI bundled the native app, so a Metro resolver regression shipped green (#1244).

The native app has three resolvers and only one of them ships. Vitest resolves through Vite, `tsc` through TypeScript's own resolver, and the app on a device through `mobile-native/metro.config.js` — its `watchFolders`, its `nodeModulesPaths`, and the `resolveRequest` hook that redirects specifiers coming from the shared `mobile/` and frontend sources. Bundling the real entry point is the only check that reads the third one.

`npx expo export --platform ios --output-dir <scratch> --clear` is the bundling entry point this toolchain supports with no device, no simulator and no running packager. It exits non-zero naming the unresolved specifier and its import stack. `--clear` plus a private `HOME`/`TMPDIR` means the verdict is never inherited from a warm Metro cache, and a run that exits zero without writing an iOS bundle fails too.

The gate tests resolution, not a path, so it is unaffected by where `@evener/appwire-client` lives.

**Bound:** `timeout 900` where coreutils provides it (the ubuntu runner, or a Mac with `gtimeout`); without it the run is unbounded and the CI step's `timeout-minutes: 30` is the backstop.

**Cost:** ~8s cold on a developer Mac, 1582 modules.

**RED evidence:** with an unresolvable import appended to `mobile-native/App.tsx` and reverted afterwards, `make test-native-bundle` exits non-zero with `Error: Unable to resolve module @evener/definitely-not-a-real-entry from .../mobile-native/App.tsx` and the import stack. GREEN with it reverted.

**Wiring:** `scripts/native/test-native-bundle.sh` (49 lines), a `test-native-bundle` target with its `##` annotations, `test-native: test-native-bundle` so the existing native CI job and `ios-testflight.yml`'s preflight both run it, the CI step renamed with `timeout-minutes`, and a `testing.md` section.

**Gates:** `make test-native-bundle` for real, `make test-native`, `make lint` (all nine, including `lint-generated`), and `go test -short -count=1 .` at the root for the Makefile audits — all pass.

Closes #1244; prerequisite for SDK A4.


obra:
## Review round 2 disposition (head d51d33282)

- **Medium, interrupt between the fork and pid registration**: fixed, db656f04d. Signals are deferred across the launch and replayed through `consume_interrupt` once ownership is recorded. The window could not be forced deterministically without a seam in the script (a fixture that signals from the bundler's own startup reached a running bundler in 19 of 20 iterations and the no-deferral mutation still passed), so `TestNativeBundleInterruptDuringLaunchStopsBundlerProcessGroup` guards the launch path rather than pinning the deferral; no seam is added for a test's sake.
- **Medium, completion poll misses orphaned workers**: fixed, db656f04d. Two handles: `bundle_pgid` is the only thing ever signalled and `bundle_pid` the only thing waited on; the group is probed after the reap and outlived workers are stopped with a line saying so. Pinned by `TestNativeBundleStopsWorkersThatOutliveTheBundler`; the local lifecycle is to be replaced by the shared library under #1246 once #1145 lands.
- **Low, background subshell inherits the `EXIT` trap**: `trap - EXIT HUP INT TERM` in the subshell, pinned by `TestNativeBundleLogCarriesOnlyBundlerOutput`; noted honestly that bash 3.2 resets subshell traps on its own, so on this host the test passes with or without the change, and the ubuntu runner is the check.
- **Low, no failing-bundler test**: `TestNativeBundleFailingBundlerReportsAndPreservesItsLog`, 0ee8533f1. `full log:` became the repo-wide `full logs:` so the shared helper is reused.

Gates: `make test-native` (bundle in 10s, 1581 modules), `make lint` 9/9, the Makefile audits, the seven native-bundle tests, `go test -short` at the root. Merged `origin/main` (42d27f946, #1070).

obra:
## Review round 3 disposition (head 4490b3fd3)

- **Medium, a second interrupt during the stop window abandons the KILL escalation**: fixed, cb248c2a2. A `stopping` guard makes `interrupted` return while a stop is in flight, and the group id is released only after the loop sees the group answer ESRCH. Writing the test surfaced a third instance the review did not name: a signal arriving inside `finish` exited with its own status (a HUP turned a TERM interrupt's 143 into 129), so `finish` carries an `exiting` guard too. Pinned by `TestNativeBundleSecondInterruptDoesNotAbandonTheStop` (second signal varied over TERM/INT/HUP) and `TestNativeBundleInterruptDuringExitPreservesStatus`, both red against the reviewed shape on a copy. Note for reviewers: bash does not re-enter a trap handler for the signal it is already running, so only a different second signal exposes the bug.
- **Low, the completion poll probes the bare leader pid**: fixed, cb248c2a2. The poll reads bash's job table (`jobs -pr`) instead of `kill -0` on the leader, so that pid is only compared, never signalled; `TestNativeBundleNeverSignalsTheBareReapedPid` now records every `kill` from the script's first call (tagged pre/post reap) and refuses a vacuous pass.
- #1246 is closed as moot after the #1145 reset; the script and `testing.md` say the gate keeps its own minimal group lifecycle by design.

Gates: `make test-native` (bundle in 7s, 1581 modules), `make lint` 9/9, the Makefile audits, the ten native-bundle tests, `go test -short` at the root. Merged `origin/main`.

obra:
## Review round 4 disposition (head 5a1b056c2)

- **Medium, interrupt status not latched**: fixed, 00a0a806f. The deferred signal is recorded only if none is, and the exit guard is raised before the stop so it covers the gap after `stopping` resets; `TestNativeBundleInterruptDuringExitPreservesStatus` gained the `after-stop-before-exit` case, red against the reviewed shape (129 vs 143).
- **Medium, the gate does not compile with Hermes though the docs said so**: the claim is removed from the annotation, the generated table and the PR body; no Hermes step is added, since the gate's job (#1244) is to prove Metro resolution.
- **Low, the launch-window test cannot detect a removed deferral**: the test is deleted, and the residual is stated next to the deferral in the script and in `testing.md`; a fixture cannot place a signal inside that window without a seam, and no seam is added for a test's sake.
- Also fixed on sight: `TestNativeBundleStopsWorkersThatOutliveTheBundler` flaked about one run in three on the merged head (the fake bundler exited before its backgrounded worker published); 5a1b056c2, clean across 25 consecutive runs.
- This script's shell lifecycle is scheduled to move onto the Go bounded-run helper from the #1145 replacement (#1263), which deletes most of it.

Gates: `make test-native` (bundle in 8s, 1581 modules), `make lint` 9/9, the Makefile audits, `go test -short` at the root; nine native-bundle tests remain. Merged `origin/main` (c5da889de).

obra:
## Review round 5 disposition (head 8104d510b)

- **Medium, an interrupt during a gate-owned stop is dropped**: fixed, 8104d510b. `latched_signal` keeps an interrupt that lands while the timeout stop or the leaked-worker stop holds the group, and `exit_if_interrupted` after each makes it the verdict, never the timeout FAIL and never PASS. Pinned by `TestNativeBundleInterruptDuringAGateOwnedStopWins` (`timeout-stop`, `leaked-worker-stop`), red with the latch line removed.
- **Low, `consume_interrupt` clears before replaying**: fixed; it clears only after `interrupted` claims the signal.
- **Low, `finish` captures `$?` before arming the guard**: fixed the only way bash allows (an EXIT-trap wrapper cannot arm before capture without destroying `$?`): every path that reaches an `exit` arms `exiting` first, which puts `finish`'s capture inside the guard.
- The two remaining one-command windows have no fixture seam and are code-review guarded, like the launch window; no seam is added. This script's lifecycle moves onto the Go bounded-run helper under #1263 once #1265 lands, which deletes it.

Gates: `make test-native` (bundle in 17s, 1581 modules), `make lint` 9/9, the Makefile audits, `go test -short` at the root; eleven native-bundle tests green twice in a row. Up to date with `origin/main`.

obra:
## Simplified per Jesse's ruling

This PR is now a rewrite on top of `origin/main`, not an increment on what it was. The old branch history is gone; the head is `08ef1f8ca` and the whole diff is **114 insertions across 4 files**, down from roughly 1,300 lines across 5.

**What the gate is now:** a private `HOME`/`TMPDIR` under a scratch dir, `NODE_DISABLE_COMPILE_CACHE=1 npx expo export --platform ios --output-dir "$dir/export" --clear` run from `mobile-native`, a check that the export actually wrote an iOS bundle, and cleanup. 49 lines of shell, and it stays a script rather than a make recipe only because `scripts/lib/scratch-lib.sh` is how this repo is allowed to delete a temp directory (`TestNoScriptFeedsVariableToRecursiveDelete`).

**What was removed**, all of it the process-lifecycle machinery that six review rounds kept finding ordering bugs in:

- `set -m`, the `bundle_pgid`/`bundle_pid` handles and the job-table completion poll
- `stop_bundle` with its SIGTERM → grace → SIGKILL escalation and group-emptiness confirmation
- every signalling trap, plus `interrupted`, `consume_interrupt` and `exit_if_interrupted`
- the `defer_signals`/`pending_signal`/`latched_signal`/`stopping`/`exiting` state machine
- the leaked-worker detection after the leader is reaped
- `EVENER_NATIVE_BUNDLE_TIMEOUT` and `EVENER_NATIVE_BUNDLE_STOP_GRACE`, replaced by a flat `timeout 900`
- the log-capture/replay path, the `full logs:` marker, `--verbose` and the module-count PASS line — the exporter's own output goes straight to the gate's stdout and stderr
- `native_bundle_gate_interrupt_unix_test.go` in its entirety (718 lines, 9 test functions, 15 scenarios), along with the code it pinned
- the earlier plan to run the bundler under an `evener-dev bounded-list` helper, which went with #1265/#1289/#1288

A hung bundler is now bounded by `timeout` where the platform has it, and by the CI step's `timeout-minutes: 30` where it does not. That is weaker than what was deleted — an interrupted run can leave Metro workers behind for a moment — and it is the right trade for a gate whose job is to notice an unresolved import.

Gates re-run in the foreground on `08ef1f8ca`: `make test-native-bundle` (8s, 1582 modules), `make test-native`, `make lint` (all nine including `lint-generated`), and `go test -short -count=1 .` for the Makefile audits.

obra:
## Review round 1 disposition (08ef1f8ca, simplified gate)

CI green. One Medium and one Low, both accepted:
- **The gate isolates HOME and TMPDIR but not the three XDG dirs the web gates isolate, so Expo and Node caches can leak warm state into the verdict** — accepted; the same three XDG dirs are minted under the scratch dir.
- **`metro.config.js` still says no gate bundles the app** — accepted; the comment names `make test-native-bundle`.

obra:
## Round 1 fixes

Both accepted findings fixed at `1f230bd6f`, pushed normally (`origin/main` had not moved, so no merge).

**Medium — private XDG roots.** The run had a private `HOME` and `TMPDIR` but no XDG roots, so anything the exporter kept under `XDG_CONFIG_HOME`, `XDG_CACHE_HOME` or `XDG_STATE_HOME` came from, and went to, the developer's own — the gap `--clear` exists to close. All three are now minted under the scratch dir and set in the same spelling `scripts/web/test-web.sh:127` and `test-web-browser.sh:43` use:

```
mkdir -p "$dir/home" "$dir/tmp" "$dir/xdg-config" "$dir/xdg-cache" "$dir/xdg-state" || exit 1
HOME="$dir/home" TMPDIR="$dir/tmp" XDG_CONFIG_HOME="$dir/xdg-config" XDG_CACHE_HOME="$dir/xdg-cache" XDG_STATE_HOME="$dir/xdg-state" NODE_DISABLE_COMPILE_CACHE=1 …
```

**Low — the stale resolver comment.** `mobile-native/metro.config.js` said a break in that `resolveRequest` branch ships green. It now points at the gate that catches it: "`make test-native-bundle` bundles the app, so a break here fails CI."

The target annotation and the `testing.md` paragraph both say "process home plus temporary and XDG roots" now, matching what the script does; the generated target table is regenerated.

**Gates, foreground on the pushed head:** `make test-native-bundle` for real (1582 modules exported), `make test-native`, `make lint` (all nine including `lint-generated`), `go test -short -count=1 .` for the Makefile audits. Diff is 116 insertions across 5 files.

obra:
## Review round 2 disposition (1f230bd6f)

CI green; no Medium or above. One Low, accepted:
- **The exported graph reaches AppWire only through relative paths, so the bundle never exercises the package-name resolver branch this gate protects** — accepted; one real package-name import joins the exported graph now (A4, #1272, rewrites the rest to the package name right behind this gate, so from then on every PR exercises the branch).

obra:
## Round 2 fixes

Fixed at `889cb98c8` (the accepted Low, then a normal `--no-ff` merge of `origin/main` `59b2b2280`).

The gate was proving a resolver nothing used: the exported app reached AppWire only through relative paths, so the bundle never entered `metro.config.js`'s `@evener/appwire-client` branch.

**`mobile-native/src/connection.ts` now imports the package by name**, both lines exactly as SDK A4 (#1272 @ `4b7ba3536`) rewrites that file:

```ts
import { AppwireClient } from "@evener/appwire-client";
import type { WebSocketLike } from "@evener/appwire-client";
```

Byte-for-byte A4's spelling for that file, so A4 merges clean over it. Nothing else was needed — `tsconfig.check.json`'s `paths`, the `vitest.config.mts` alias and the Metro `resolveRequest` branch all already mapped the name on `main`; only an import site was missing. The comment in `metro.config.js` that said nothing imports by this name yet now says what does.

**Proof it exercises the branch**
- With the import: the export grows from **1582 to 1596 modules** and passes.
- With the branch disabled (`if (false && …)` around the `resolveRequest` match, reverted afterwards): `make test-native-bundle` exits non-zero with `Error: Unable to resolve module @evener/appwire-client from .../mobile-native/src/connection.ts`.

**Gates, foreground on the merged head:** `make test-native-bundle` for real (1596 modules), `make test-native`, `make lint` (all nine including `lint-generated`), `go test -short -count=1 .` for the Makefile audits. Diff against `origin/main` is 122 insertions across 6 files.

PR #1249, opened 2026-09-13
Shared-notes follow-ups (items 2-13 from PR #1070 review rounds)

obra:
The deliberately deferred findings from PR #1070's roborev rounds, tracked item by item in `docs/superpowers/plans/2026-09-12-shared-notes-followups.md`. That file is the source of truth; this section only mirrors it.

**Done in this branch**

- Item 1, `handleRemoveURL` in-flight guard (Low) — `99fb9b8089`.
- Items 2 and 7, stale `expectedInstanceId` at debounce schedule time / `blurHumanNote` captures the instance before hydration (Medium, Low) — `24a4a433c3` (merged in `fe3c249fc0`). The delayed save samples the instance identity inside `save()` after hydration, so a rotation the client has already observed saves against the current instance; the daemon's `ExpectedInstanceID` fence remains the authority.
- Item 4, notes and URLs share one carrier generation (Medium) — `e113febe21` (merged in `0a9f86926c`). `urlsCarrierGeneration` splits out of `notesCarrierGeneration`, so each carrier fences only the fields it wrote.
- Item 5, terminal injection via unstripped control sequences (Medium, security) — `53d83a21da`. `normalizeNote` strips C0, DEL and C1 after its whitespace collapse, covering the human note, the agent note and every URL label; both raw unknown-id error echoes now quote the id. Write-path fix only: text persisted before it is not re-sanitized.
- Items 8 and 9, tentative notes state in the cached envelope / non-atomic `notes_read` snapshot (Medium, Low) — `a30b1ab8e3` (merged in `80747e1a0`). Readers take one atomic load of a published committed notes cut (human note, agent note, URL list, ever-projected flag); a mutator publishes only after its metadata save returns nil; the metadata write persists the live staged store, since that write is the mutation's durability point.
- Items 10 and 11, doubly-encoded framing references / innocent references rewritten (Medium security, Low) — `655bac680f`. Character references are parsed completely (any number of case-insensitive `amp;` layers, required `;`, named or numeric base, resolved value must really be an angle bracket) and rewritten to the canonical `&lt;`/`&gt;` spelling, so nesting depth buys nothing and the pass is idempotent. Residual recorded in the commit message: a reader that decodes one layer still sees `<`, exactly as it already did for a literally-typed `<`.
- Item 12, `conversationSignals` counts injected notes context (Low) — `1f7c3f07e3`.
- Item 13, past-session roster decodes a whole mutation snapshot per entry (Low, performance) — `481c34b2d8`. The roster reads the top-level `human_note` through `ReadPersistedHumanNote`; the strict reader stays the authority everywhere else.
- Backlog status recorded in `182db3d77`.

**Not fixed**

- Item 3, `urls/remove` durable outbox routing (Medium) — accepted limitation, per Jesse's ruling.
- Item 6, URI scheme rejection blocks filenames with colons (Low) — closed without a code change: the rejection is the deliberate defense against `javascript:`/`data:`/`mailto:`, and the colon-path escape hatch is documented on `DefUrlsAdd` and named in the error message instead.

Every item above was verified by reverting its production hunks in a clean lane and reproducing the red condition before the fix landed. The roborev rounds on `182db3d`, `94ef1ea`, `283157e`, `2ef2e86` and `95ec446` are answered in `6af531e`, `283157eaa`, `2ef2e864f`, `95ec446b6` and `e12699abf` (sixteen further commits, one per finding, each red-first; the per-finding replies are on this PR), and `main` is integrated. A structural follow-up now closes the legacy-sanitization class at the TUI's terminal boundary (notes surfaces only, command output untouched); see the comment below. Final gates on `7b9915397`: `make merge-approval-gate` exit 0 (all nine lint stages, every Go module stream, the frontend gate, the native package and the appwire package qualification), `make vet` exit 0, `make test-web-browser` 5/5 Chrome guards exit 0.

**Branch name note:** this is PR #1070's follow-up branch rebased onto `main` after #1070 squashed. The earlier `shared-notes-followups` ref on origin predates that squash, so it cannot be fast-forwarded; rather than force-push it, the rebased branch is pushed as `shared-notes-followup-items`.


obra:
Round on `182db3d` answered: six findings, six commits, pushed in `6af531e` on top of the reviewed head. All gates re-run on that commit (`make merge-approval-gate`, `make vet`, `make test-web-browser` 5/5).

**C1 control characters pass through URL parsing** (Medium) — fixed in `bb4784854`. `canonicalSessionURL` now refuses any control rune in the raw input before parsing. `url.Parse` only rejects ASCII controls, so `https://x.test/y?q=<U+009B>31m` really did store a URL that printed a CSI sequence in the TUI details drawer and the notes tool output. Covered red-first for C1, BEL, DEL and ESC in query, path and host positions; the path and file-URL cases the parser previously percent-encoded silently are refused now too, and a percent-encoded URL still parses.

**Double-click race in Remove URL handler** (Medium) — fixed in `0ce49e706`. The handler checks and updates a ref synchronously and mirrors it to state only for the disabled row. The new test dispatches both clicks inside one `act()` so no re-render intervenes; before the change it fired two `urls/remove` requests and reached the second, erroring branch.

**Missing human-note parked-save coverage** (Medium) — added in `3bb843766`. The subtest parks a second `SetHumanNote` inside the mutation store's effect-snapshot write (via the durable harness, since only a store over the real filesystem reaches that write), asserts `Meta`, `notesSnapshotAll` and `notes_read` keep the committed note while parked, then asserts the rejection, the surviving committed value and no emission. It carries an anti-vacuity guard that the parked mutation holds its reservation, and asserts the injected failure by text because `SetHumanNote` normalizes it into a wire error.

**No-note case still traverses journal** (Low) — fixed in `c6aa547ac`, with numbers: absence is now decided by the key's bytes, so a document that never spells the key costs one scan instead of a token walk. `BenchmarkReadPersistedHumanNoteAbsent` measures 1.46ms to 0.21ms on a 535KB keyless document (14us to 11us when empty). The reader's contract now says what it does: absent file, absent key and null all read as no note, while a document carrying the key that fails at or before the value stays an error. The three keyless malformed cases are re-pinned accordingly; they are the documents the roster already degraded to "no canonical note" for. True O(1) absence needs the sidecar/index you also suggested, which is a persistence-format change, so it is recorded as a follow-up rather than folded into this branch.

**Control stripping breaks single-space invariant** (Low) — fixed in `efba9162e`. The strip runs before the collapse and leaves the whitespace controls to `strings.Fields`, so `a \x1b b` stores `a b` (it stored `a  b`) while `a\nb` still stores `a b`.

**Doc claims absent-entry removal success path** (Low) — corrected in `6af531e46`. The Status entry now describes the guard only, and the item section records the unimplemented half explicitly. I did not implement "already absent counts as success": `threadsStore.removeURL` still surfaces an unknown-id error, which is honest for an entry another client removed. Happy to implement it if you want that half too.

obra:
Nit fixed in `94ef1ea42`. The item 5 Status line was stale from my own `efba9162e`: it now states the implemented order (strip before the whitespace collapse, and why), and records the URL control refusal and the quoted id echoes that completed item 5 as well. `make lint` re-run clean before the push.

obra:
Round on `94ef1ea` answered: three findings, three commits, pushed in `283157eaa` on top of the reviewed head. Gates re-run on that commit: `make merge-approval-gate` exit 0, `make vet` exit 0, `make test-web-browser` 5/5.

**Numeric-reference bypass of framing canonicalization** (Medium, security) — fixed in `de7b060d3`. Correct catch: the canonicalizer knew only the named layer spelling, so `&#38;lt;/shared-notes&#38;gt;` passed through untouched and two entity passes decode it back to the closing tag. A numeric reference resolving to `&` (`&#38;`, `&#038;`, `&#x26;`, `&#X26;`) is now consumed as a layer in the same loop, so every spelling of an angle bracket, however layered, collapses to the one canonical form. Covered for those spellings mixed with named and zero-padded layers, plus one end-to-end payload through the block.

On the test shape you asked for: I pinned "every spelling canonicalizes to the same bytes" rather than "one decode yields no tag". Under the design chosen for items 10 and 11, the canonical `&lt;` decodes to `<` in a single pass, exactly as a literally-typed `<` does, so a test asserting that no single decode reaches a tag cannot pass under canonicalization; it would only pass with the lookalike substitution that was declined when this design was picked (the residual is recorded in `docs/superpowers/plans/2026-09-12-shared-notes-followups.md`). If you want that class closed outright rather than bounded at one decode, the substitution is the option to take and I will implement it.

**Unsanitized legacy notes/URLs on restore** (Medium) — fixed in `283157eaa`. The load boundary now normalizes. Restore strips the agent note and the URL list (labels as note text, a URL's own control characters removed); the mutation-snapshot load strips the canonical human note, which covers the session store, the published committed cut and the strict reader at once; the roster's own reader strips what it returns, since it never goes through a Session. The test plants a legacy meta.json plus a mutation snapshot carrying controls in the human note, the agent note, a URL and a label, restores, and asserts every reader surface and both disk readers hand back clean text. The next metadata save persists the cleaned values, so a legacy session heals on its next write.

**URL removal guard cleared prematurely** (Low) — fixed in `01d2c0782`. The row stays pending until the model stops listing the entry (the `evener/urls/updated` push is the authority), and a failed request releases the guard so a retry still reaches the wire. The gap test fired two requests before the change; the retry half is pinned as well.

One note on the earlier run of this round, in case it shows up in CI or your own runs: the first gate attempt on this head hit three load flakes in the frontend suite (two 5s timeouts and one missing-fixture-text failure, all in files this diff does not touch and all passing in isolation) and two browser guards failed with `timeout calling navigateTo after 30000ms` while the host was at load average 108. Re-running once the machine quieted gave the clean result above.

obra:
Round on `283157e` answered: three code gaps, one doc fix, pushed in `2ef2e864f`.

Reconciling the disagreement the synthesis flagged: Review 2 was right that the load boundary covers the canonical human note, the restored agent note and the URL list, and Review 1 was right that three other paths carry legacy values around that boundary. All three are fixed in `20a5a9592`, each red-first:

- **Replayed `notes/human/set` result** — a journal written by a binary that predates the strip records the raw note, and replay handed it straight back in the response. The response value is stripped now, while the journal keeps its historical record. Red: the replayed response carried U+001B.
- **Restored URLs and entry ids** — these are single tokens that never went through the note collapse, so the note strip left whitespace controls (U+0085) in a URL and never touched an id at all. They take a full control strip, whitespace included, and a label loses controls without being re-collapsed. Red: the restored URL carried U+0085.
- **Persisted `NOTES_CONTEXT` turn** — the history escaper only knows the framing spellings, so a legacy turn's ESC/OSC/CSI survived into the model copy. It is stripped before the escape now. Red: the history copy carried U+001B.

**Stale status record** — fixed in `73f3da393`: items 1 and 5 now describe the implemented behaviour, with the commits that added it.

One correction worth recording, caught by `cmd/evener`'s canonical-envelope test rather than by review: my first cut of the replay fix normalized the response on every call, and `normalizeNote` is not idempotent on its own output, because the clamp can leave a trailing space that a second collapse drops. The test failed with 999 runes where it wanted 1000. `2ef2e864f` makes every load and replay path strip controls only and never reshape what was persisted, which is what the snapshot load and the roster read already did.

Gates on the pushed head: `make merge-approval-gate` exit 0, `make vet` exit 0, `make test-web-browser` 5/5.

obra:
Round on `2ef2e86` answered: two medium and one low, all fixed and pushed in `95ec446b6`, which also carries the `main` integration (so this head is no longer behind).

**Legacy steering entries carry unsanitized control characters** (Medium) — fixed in `d90d31dda`. `clientSteeringFromSnapshot` copies the journaled pending text straight into the live queue, and that text reaches the model and the transcript after a restart. It is stripped now, like every other re-served path.

**C1 whitespace controls survive in restored `NOTES_CONTEXT` history** (Medium) — fixed in `d90d31dda`, and strengthening the fixtures turned up a third instance of the same mistake. `escapeNotesHistoryTurns` used the note strip, which keeps whitespace controls because the note path always collapses them afterwards; the history copy has no collapse, so a legacy NEL survived. The same held for a restored note or label. The new `stripTextControls` is the rule for text re-served this way: every control except the newline goes, and a whitespace control becomes a space rather than gluing words together. I verified these are load-bearing rather than passing by construction: neutering the strip in place makes the steering, history, replayed-response and restore tests all fail.

**Component-local in-flight removal guard has two failure modes** (Low) — fixed in `802276a97`. Both were real: holding the guard until the authoritative push wedged a live row whose id another client re-added before that push landed, and a panel remount reset it anyway. The guard is request-scoped now, and the double-click problem is solved at its source instead: a removal the server answers with "no URL entry with id" reads as success, because the entry is already gone and the user's intent is satisfied. That also implements the half of item 1 I had previously left out, so the backlog status says so. One deliberate deviation from your suggested fix: I did not move the pending set out of the panel. With the source-level fix, its lifetime can no longer produce a user-visible failure, so extracting it to a session-and-id store would add surface without changing behaviour; say the word if you want the belt-and-braces version and I will do it as a follow-up.

Gates on `802276a97`, which differs from the pushed head only by the item 1 status line: `make merge-approval-gate` exit 0 (nine lint stages, every module stream, web 749s, the native package and the appwire qualification), `make vet` exit 0, `make test-web-browser` 5/5.

obra:
Round on `95ec446` answered: two findings, both verified against the code before anything was changed, and both real. The synthesis asked for exactly that verification because the reviewers contradicted each other; here is what the code says.

| Path | Review 1 | Review 2 | Actual |
|---|---|---|---|
| Past-session projection (`app_threadread.go:757-758` + `persistedSessionURLs`) | not sanitized | covered | **Review 1 is right.** Only the human note comes through `ReadPersistedHumanNote`; the agent note and the URL id, URL and label are copied from the hub's own metadata store, which the agent-side load strip never touches, and the TUI prints that payload. |
| Steering history (`escapeNotesHistoryTurns`) | not sanitized | covered | **Both are partly right.** The *rebuilt* queue entry is sanitized (round four), but the persisted `TurnSteering` turn was not, and that copy is what resumed requests and inherited prefixes hand to the model. |

**Past-session roster leaves `AgentNote` and URL fields unsanitized** (Medium) — fixed in `e12699abf`. The projection sanitizes the agent note with the note-text rule and the URL id, URL and label with the single-token rule, and the two sanitizers are exported from the agent package so the hub applies one definition instead of a copy. Red before the fix: the projected agent note carried U+001B.

**`escapeNotesHistoryTurns` misses legacy `TurnSteering` turns** (Medium) — fixed in `e12699abf`. The history copy strips steering turns as well; the stored turn keeps its bytes and only the model copy is rewritten. Red before the fix: the steering copy carried U+001B and U+0085.

Neither test passes by construction: neutering the strip in place makes both fail.

Gates on the pushed head: `make merge-approval-gate` exit 0 (nine lint stages, every module stream, web 946s, native package, appwire qualification), `make vet` exit 0, `make test-web-browser` 5/5.

One note for the maintainer, not a finding: this is the fifth round of the same class, legacy text reaching a model or a terminal through whichever reader happens to touch it, and each round closes the paths it names. The structural alternative is one sanitizing chokepoint at the render boundary instead of at every reader. That changes how tool output is rendered, and tool output can legitimately contain ANSI, so it is Jesse's call rather than something folded into this branch.

obra:
Structural follow-up after the last five rounds, on Jesse's call rather than from a finding.

Each of those rounds named another reader through which legacy (pre-strip) text reached a terminal — the agent's restore and readers, the mutation-snapshot load, the roster, the rebuilt steering queue, the hub's past-session projection, and the history copies — and each was closed at its own boundary. This closes the class for the terminal: `cmd/evener-tui` now has one boundary rule, `tuitext.StripControls`, which drops every control character except newline and tab, applied at the notes surfaces:

- the details drawer's human note, agent note, and each URL line (label, URL, id);
- the steering body, which is how a human-note update arrives;
- the shared-notes read tool's output, through a renderer entry scoped to that tool name.

The scope limit is deliberate and tested: no other tool body, shell output, file content or command rendering is touched, because those can legitimately contain ANSI, and one test asserts a non-notes tool keeps its escape sequences. Both halves are proven rather than asserted — neutering the strip makes the drawer, steering and notes-read tests fail, and sabotaging the generic body renderer makes the no-over-reach guard fail.

The load-boundary strips stay as the second layer, not a replacement.

Gates on the pushed head: `make merge-approval-gate` exit 0 (nine lint stages, every module stream, web, native package, appwire qualification), `make vet` exit 0, `make test-web-browser` 5/5. One guard run failed `web-overflowguard` with `timeout calling navigateTo after 30000ms` under a load average of 73 and passed 5/5 immediately on re-run; that guard exercises the web UI, which this Go TUI change does not touch.

obra:
Round on `e12699a` answered: the high-severity finding was real, it was mine, and it is fixed in `b5e8a407a`.

**Good catch, and thank you for it.** My round-four steering sanitization rebuilt each steering turn with `llm.User(stripTextControls(text))`, which produces a message containing only a text part. Every non-text part was dropped, so an image attachment on a steering turn, or queued image-bearing input drained as steer, disappeared from resumed requests and inherited prefixes. That is data loss, and it came from my change rather than from anything pre-existing.

**The fix** rewrites only the text parts, and copies the part slice before the first change, so the stored turn keeps its bytes while the model copy loses the controls. Red evidence before the fix, from the new assertion in `TestNotesHistoryCopyStripsLegacyControls`: `steering copy has 1 parts, want the text part and the image part`.

On the alternative you offered: I did not narrow the case to `SteeringKindHumanNote`. With the non-text parts now preserved there is no data loss, and a legacy control sequence in any steering text has no business reaching the model, so stripping every kind is the safer of the two. The image part is asserted to survive by identity, not just by count.

Gates on the pushed head: `make merge-approval-gate` exit 0 (nine lint stages, every module stream, web 810s, native package, appwire qualification), `make vet` exit 0, `make test-web-browser` 5/5, zero failing tests.

obra:
Following up on the other two findings from the `e12699a` round, both fixed in `a5c92b138` (the high-severity one was answered above in `b5e8a407a`).

**Reloaded `NOTES_CONTEXT` items render raw** (Medium) — fixed. A reloaded notes-context turn arrives as a `systemMessage` thread item and renders as a system message, whose renderer prints its text as-is, so a session written before the strip still showed its controls in the transcript. The reducer now strips exactly that event kind, keyed off `ThreadItem.EventKind == ThreadItemEventKindNotesContext`. Red before the fix: the reduced message carried U+001B. The scope guard is explicit too: a test asserts a system message that is *not* a notes-context item keeps its bytes, since hook and other system text is not notes-derived.

**Escaped JSON key reads as absent** (Low) — fixed. You are right that the absence pre-filter was unsound: `"\u0068uman_note"` is a valid JSON spelling of the same key and the strict reader decodes it, so the roster would silently hide a note the authority shows. The pre-filter now bail out and fall through to the token walk whenever the document contains any backslash, since that is exactly the condition under which byte equality stops implying key equality. Red before the fix: the light read returned `("", false, nil)` for an escaped-key fixture the strict reader accepts (that fixture asserts the strict reader's acceptance first, so the contrast is pinned).

Both tests are load-bearing in the same way as the earlier ones: neutering either fix turns its test red.

Gates on the pushed head: `make merge-approval-gate` exit 0 (nine lint stages, every module stream, web 541s, native package, appwire qualification), `make vet` exit 0, `make test-web-browser` 5/5, zero failing tests.

That closes all three findings from the round. For the record, the two I did not take as suggested: stripping stays applied to every steering kind rather than only human-note updates (no data loss now that the parts are preserved, and any legacy control in steering text has no business reaching the model), and the escaped-key path trades the byte pre-filter's fast path for correctness rather than trying to enumerate escaped spellings.

obra:
Round on `b5e8a40` answered: the one finding is fixed in `7b9915397`.

You are right on both halves, and this was a request I deferred in round four rather than a new problem, so it is on me that it came back. The guard lived in `NotesPanelBody` as component state, which meant the mobile sheet unmounting the body mid-request reset it (letting a duplicate `urls/remove` through), and the same mounted pane reused for another session carried the earlier session's pending id and disabled a live row.

It now lives in `stores/pendingUrlRemovals.ts`, keyed by `(sessionRef, url.id)`, with the synchronous begin-check the double-click case needs and a reset hook for tests. The panel reads it through a hook for the disabled attribute and calls begin/end around the request.

Both lifecycle cases have red-first tests: with the component-local guard the remount fired two requests where one is wanted, and the same-id row on a different session rendered disabled. I also falsified the isolation test by neutering the session key in place, which turns it red, so it detects a non-keyed guard rather than passing by construction.

Gates on the pushed head: `make merge-approval-gate` exit 0 (nine lint stages, every module stream, web 688s, native package, appwire qualification), `make vet` exit 0, `make test-web-browser` 5/5, zero failing tests.

Issue #1263, opened 2026-09-13
Run the native Metro bundle gate under the Go bounded-run helper and delete its shell process lifecycle


obra:
`scripts/native/test-native-bundle.sh` (#1245) bounds `expo export` with hand-written bash: its own process group, deferred signals across launch, TERM-then-KILL escalation, reentrancy guards, a job-table completion poll, and ten tests for that lifecycle. Four review rounds each found a real gap in it, the same class that led to the #1145 reset (2026-09-13): a shell process supervisor produces real findings forever.

The #1145 replacement adds a Go helper in `cmd/evener-dev` that runs a command under a context timeout in its own process group and kills the group on expiry, tested with a real child. Once that lands, rewrite the bundle gate as `evener-dev <bounded-run> --timeout 600s -- npx expo export …` plus the log/scratch handling, and delete the script's lifecycle code and the tests that only existed to pin it (keep the tests that pin the gate's own promises: an unresolved module fails, a failing bundler fails loudly with its log preserved, success reports the module count).

Sequence: after the #1145 replacement PR 1 merges and #1245 merges. Replaces the closed #1246.

PR #1265, opened 2026-09-13
ci: retry the golangci-lint download and bound go list with a Go helper

obra:
Two CI-infrastructure failures on 2026-09-11, both on heads whose code was clean, neither with a retry or any headroom.

`lint-golangci` (run 34644540860) died on `err http_download_curl received HTTP status 500`, taking the aggregate `static` job with it. `tests` (run 34639098143) died on `go list ./... timed out after 30s` for the root module before a test ran, advising a cache repair that was not the problem.

This replaces #1145, which grew a shell process-group library, a record protocol, markers and a self-check across 35 review rounds for the same two failures. None of that is here.

## The golangci-lint download

`scripts/ops/install-golangci-lint.sh` replaces the `curl | sh` in `make/repo.mk`. There are two downloads and each retries in the way it allows:

- ours, the fetch of `install.sh`, uses curl's own `--retry 3 --retry-delay 2 --retry-all-errors -f`, so an HTTP status counts as a failure worth another attempt;
- the installer's own fetch of the release asset, which is where the 500 was, has no retry and no option to ask for one (`-b`, `-d`, `-h`, `-x`, the tag), so that attempt is repeated by running the installer again: 3 attempts, 5s then 10s apart.

It does not classify the failure, because the installer collapses every cause to exit 1 and reading its log text for a status would couple us to upstream's wording. `pipefail` closes a hole the old recipe had: when the fetch of `install.sh` failed, `sh` read an empty script and exited 0, so `make tools-golangci` reported a successful install of nothing. The pin is still read from `.tool-versions`, and the installed binary now has to report it.

## The package list

`evener-dev bounded-list -timeout 60s -attempts 3 -- go list ./...` runs the command under a per-attempt bound and retries an attempt that hits it. A killed attempt still warms the caches, so a cold, loaded runner passes on the next attempt while a stalled volume fails with a named diagnostic rather than a job-level timeout. Only a timeout is retried: a `go list` that exits non-zero has decided something about the package list itself.

The bound stops a process group rather than the one child the helper spawned. `go list` forks compilers, and a survivor keeps holding the build and module cache locks every later run on the host needs -- and, inheriting the pipe, keeps the parent's `Wait` from returning, so the "bound" would not bound anything. That is what the tests check, with real processes: a `sh -c 'trap "" TERM; sleep 60'` child whose group must be gone after the bound (with the signals sent to the leader alone it takes the child's full 60s and fails), output pass-through, a non-zero command kept at its own status and not retried, and exhausted attempts exiting non-zero with the diagnostic.

`scripts/gate/run-module-tests.sh` uses it for both enumerations, through `EVENER_PACKAGE_LIST_TIMEOUT` (60) and `EVENER_PACKAGE_LIST_ATTEMPTS` (3). The agent module's subpackage list was unbounded and, read through a process substitution, could not report a failure at all; it now takes the same path and the same diagnostics. The helper owns the signalling, so `stop_process_tree` lost its last caller and is gone.

## Deliberately left out

- No shell process-group library, wrapper script, record protocol, markers, or self-check.
- No `ps` parsing anywhere new; the gate's existing `stop_children` walk is untouched.
- No flag classification, and no second flag parser: the helper takes the command after `--`.
- No HTTP status classification in the installer, and no `go install` fallback for golangci-lint.
- `-C` is how a module other than the repo root gets enumerated, rather than a new flag on the helper.

## Known limits

- `go run ./cmd/evener-dev/bin` compiles the helper before its bound starts, so a build cache that is stalled rather than cold can still hang ahead of the bound. The gate already depends on `go run` of this binary for agent shards.
- Windows has no POSIX process groups; `boundedlist_other.go` stops the single child so the cross-compile vet builds. This is repo tooling that runs on the unix dev and CI hosts.

## Gates

`go test ./cmd/evener-dev/...`, `go vet -tags evenerfuzz ./...` and `GOOS=windows go vet -tags evenerfuzz ./...`, `make lint` (all 8 modules) and `make lint-generated`, `bash -n` on both scripts, `MODULES=". envvars"` and `MODULES="agent"` gate runs through the new path, and `make tools-golangci` against the real network (installed and verified v2.13.1).

Closes the work of #1145.

## Review round 1

Six findings, all wiring; the design did not change.

- **The `go run` compile ran outside the bound** (High, #1267). No `go run` anywhere in the gate now. `make tools` builds `evener-dev` alongside the pinned linters, `make test` and `make test-race` take `build-dev` as a prerequisite the way the lint targets already do, and the gate execs the binary or stops with the make target to run. The compile still happens on a stalled cache, but once per invocation under make, never inside the gate. The agent shard runner was launched the same way and is now the same prebuilt binary, which also restores its 129/130/143 signal exits that `go run` collapsed to 1.
- **One enumeration path** (High). `AGENT_SHARDS=0` and the five small modules ran `go test ./...`, whose own discovery reads the same caches with none of the bound. Every module now enumerates through the helper and is handed the list.
- **`go list -C` placement** (High). Not reproducible: go1.27.0 accepts `-C` as the first flag after the subcommand as well as before it, and `go list -C agent ./...` and `go -C agent list ./...` print the same 42 packages here, which is why the agent gate run passed. `-C` is gone regardless -- the enumeration runs in the module's own directory, which the gate had already entered.
- **Survivors after the leader is reaped** (Medium). A leader that handles SIGTERM and exits can leave a child that ignores it holding the cache locks. Every attempt now ends by probing the group and stopping what it finds: SIGTERM, the grace, SIGKILL. This is safe after the reap precisely while the group is non-empty, since a pid cannot be reused while it is still a live group's id. The new test is a real leader that exits at once while its child traps TERM with the pipe closed; without the sweep the group is still there three seconds later.
- **The renamed knobs** (Medium). `EVENER_ROOT_PACKAGE_LIST_TIMEOUT` and `_ATTEMPTS` now stop the gate with a message naming their replacement. Not honoured as a fallback: that is backward compatibility, which is Jesse's call and has not been asked for.
- **`-grace` validation and the stale `--help`** (Low). `-grace` is validated with `-timeout` and `-attempts`. Both help surfaces read one sorted subcommand list, so neither can go stale again.

Gates: root audits, `go test ./cmd/evener-dev/...`, `go vet -tags evenerfuzz ./...` and the `GOOS=windows` repeat, `make lint`, `make lint-generated`, `bash -n` on both scripts, gate runs on `.`, `envvars`, `agent` sharded and `AGENT_SHARDS=0`, and `make tools` then `make tools-golangci` against the real network.

### What the Metro gate's port needed

Porting the Metro gate (#1245's rewrite) onto the helper found three things a second caller needs and the first happened not to.

- **A timed-out attempt's stdout was dropped.** For a gate that output is the evidence: half a package list says where the enumeration was when the volume stalled. It is passed through on every outcome now, after the diagnostic.
- **No signal handling, which the isolation itself introduced.** The command runs in its own process group, so a Ctrl-C or a runner's SIGTERM no longer reaches it the way it reached a `go list` the shell ran in its foreground group. TERM, INT and HUP are forwarded to the group, given the grace, then SIGKILLed; the helper exits 128+signal and does not retry, since an interrupt is an answer about this run. Tested with a real TERM-ignoring child: the group is empty and the status is 143.
- **A timeout now exits 124** (coreutils' `timeout` convention), written down in the usage text along with the other statuses, so a caller need not read stderr to tell a timeout from the command's own failure. The gate uses it: cache-repair advice is printed for 124 alone, since any other status is `go list` deciding something about the packages themselves. The diagnostic reads "on 1 attempt" when that is what happened.

Not done, as asked: no streaming to a file and no `-stdout` flag; both callers are fine with buffering.


obra:
Two residuals from this lane filed: #1267 (`go run` compiles the helper before its own bound starts) and #1268 (enumeration ignores build-tag-only packages, latent today). Neither blocks this PR.

obra:
## Review round 1 disposition (head 78bd1ae49)

- **High, `go run` compiles the helper outside the bound** (#1267): fixed, 70a01b916. `make tools` builds `evener-dev` once; `test` and `test-race` take `build-dev` as a prerequisite; the gate execs the binary and stops with the make target to run if it is missing. The agent shard runner is the same binary now, restoring its 129/130/143 exits. Closes #1267 on merge.
- **High, `go list -C` after the subcommand**: not reproducible — go1.27.0 accepts `-C` after the subcommand as well as before (`go list -C agent ./...` and `go -C agent list ./...` print the same 42 packages), which is why the agent gate run passed; nothing was masking it. The flag is gone anyway (b5adff0a9), since the enumeration runs in the module's own directory.
- **High, non-sharded paths unbounded**: fixed, b5adff0a9. One enumeration path for every module.
- **Medium, survivors after the leader is reaped**: fixed, 67345f6eb. Every attempt probes the group and stops what it finds; safe after the reap precisely while the group is non-empty. Mutation-proved.
- **Medium, retired env knobs silently ignored**: 2f3fe7318 refuses the old names with their replacement named, exit 2; no fallback, since that would be backward compatibility nobody approved.
- **Low, `-grace` unvalidated and stale `--help`**: fixed, 67345f6eb; both help surfaces read one sorted list.
- Additions for the second caller (the Metro gate's port, #1263): b0531c890 and a9b37ded7 — stdout passed through on every outcome, TERM/INT/HUP forwarded to the group with grace then KILL and exit 128+signal, no retry after an interrupt, timeout exits 124 and is documented; no streaming.

Gates: root audits, `go test ./cmd/evener-dev/...`, both vets incl. `GOOS=windows`, `make lint`, `make lint-generated`, `bash -n`, gate runs on the root module, envvars and sharded agent, `make tools` and `make tools-golangci` against the network. The `AGENT_SHARDS=0` agent run passed in the earlier sweep and was still executing locally at push time; CI is its check on this head. Merged `origin/main`.

obra:
## Review round 2 disposition (78bd1ae49)

CI: **red**. `race-root` reports DATA RACE in `TestBoundedAttemptForwardsAnInterruptToTheGroup` and `TestBoundedListDoesNotRetryAfterAnInterrupt` (run 34786344391). Root-cause fix required; no `-race` skips.

RoboRev (2 of 3 reviewers):
- **Medium, `test: build-dev` compiles unbounded** — accepted. Fix in CI, not in bash: run `make build-dev` as its own workflow step with `timeout-minutes` before the test step, so the Make prerequisite is a no-op there. Local runs stay unbounded on purpose (a human is watching).
- **Medium, signals during `stopSurvivors`/retry ignored** — accepted. Check the latched signal before launching another attempt and before returning from cleanup; test both.
- **Low, `exitCodeOf` -1 → 255 and duplicated group discipline** — accepted as written: move the group-existence primitive into `internal/devtool/procgroup` and have `boundedlist_unix.go` call procgroup for start/stop/exit-code. One implementation.

obra:
## Round 2 fixes

Head `6a76eb2da`.

- **`1a536bdeb` — the data race (race-root).** Root cause was in the helper, not the test: the command's stderr and this helper's diagnostics are the same `io.Writer`, exec copies the former from a goroutine of its own, and the interrupt line is written while the command is still running. Two goroutines, one unsynchronised writer. They now share a lock. Reproduced with `go test -race -count=3 -run Interrupt ./cmd/evener-dev` before the fix, green after. The test's `bytes.Buffer` is what the detector could see; the production writer is a file, where the same overlap interleaves bytes in the diagnostic instead.
- **`2477df2b4` — signals during cleanup and between attempts.** The select watching for TERM/INT/HUP returns when the bound does, so a signal arriving during the SIGTERM-to-SIGKILL grace sat unread and the runner started the next attempt. The signal is latched now: an attempt takes a late one at the end of its cleanup and reports it as its own answer (128+signal, not the timeout's status), and the loop re-checks the latch before starting another attempt. Tests: a TERM-ignoring group signalled during the grace (143, group empty), the same through the runner with no retry announced, and the latch's own rules. Both go red with the late take removed.
- **`97915f6e3` — one process-group discipline.** `boundedlist_unix.go` and `boundedlist_other.go` were a second copy of `internal/devtool/procgroup`; both are gone and the helper calls the package, including `procgroup.ExitCode`, which is what fixes the exit status: `ExitError.ExitCode()` is -1 for a signal-killed child, so a killed `go list` was reported as 255 instead of 128+signal. procgroup gains `Exists`, the one primitive it lacked — the question the survivor sweep asks after its child is reaped, safe because a pid cannot be recycled while it is still a live group's id. Its test starts a real group, sees it, kills it by that group id alone, and watches it go. procgroup already has a `!linux && !darwin` half, so every platform bounded-list builds for is covered.
- **`d623aa1af` — the unbounded compile on the cache-sensitive path.** CI only: each job that runs the gate now has a `make build-dev` step of its own with `timeout-minutes: 5`, before the step that runs `make test` / `make test-race`. The Makefile prerequisite stays, so a local run still gets the binary and the gate still refuses to compile anything; on CI the prerequisite is already satisfied when the gate starts, and a stalled cache fails as a five-minute build step that names itself. No shell timeout wrappers.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev ./internal/devtool/...`, root Makefile audits, `make lint`, `make lint-generated`, `bash -n`, and both gate paths on the agent module — default sharded (60s) and `AGENT_SHARDS=0` (289s) — plus `.` and `envvars`.

obra:
## Review round 3 disposition (6a76eb2da)

CI green. RoboRev (2 of 3 reviewers):
- **High, `build-dev` is `.PHONY`, so `make test` rebuilds after the bounded CI step** — accepted; the round-2 fix was incomplete. Make the gate's prerequisite a real file target (the binary, with the Go sources under `cmd/evener-dev` and `internal/devtool` as prerequisites) so an up-to-date binary runs no `go` command at all; `build-dev` stays the human-facing PHONY alias. Root Makefile audits must pass.
- **Medium, SIGHUP/SIGINT reported as forwarded but the group gets SIGTERM** — accepted: forward the received signal to the group first, then the TERM/KILL escalation.
- **Medium, post-KILL `cmd.Wait` can block forever on a D-state child** — accepted as suggested: bound the reap after KILL, return 124, skip the survivor sweep so the OS reparents.
- **Low, unix tests run where procgroup has no group support** — accepted: constrain the tests to `linux || darwin`.

obra:
## Round 3 fixes

Head `bad2f8b20`.

- **`c4a5803c5` — the phony prerequisite (High).** `build-dev` is `.PHONY`, so every gate run rebuilt the binary the bounded CI step had just built. `evener-dev` is now a real file target whose prerequisites are the sources it is built from (the `cmd/` and `internal/devtool` trees, plus `go.mod`/`go.sum`), and `test`/`test-race` depend on that file. Proof: after `make build-dev`, `make -n test` prints no `go build` and `make -q evener-dev` exits 0. `build-dev` stays as the human-facing alias and forces a build (`--always-make evener-dev`).
  A file target has to be exempt from `TestEveryRuleIsPhony`, whose subject is the opposite case — a phony name silently skipped because a file of that name exists. There were no file targets in the makefiles to follow, so the makefiles now declare `FILE_TARGETS` and the audit reads it, with a second audit for the other half of the bargain: a name listed there must have a rule with prerequisites and must not also be `.PHONY`. Root audits green; `make generate` picked up the new row.
  Limit, stated in the makefile: the prerequisite list covers the two trees that hold every subcommand this binary dispatches, not its whole import graph. `make build-dev` is the answer to that, and CI builds from scratch anyway.
- **`46a5a71ec` — the forwarded signal (Medium).** The latch kept SIGHUP/SIGINT and the cleanup then sent SIGTERM. `procgroup.StopWith` forwards the signal that was received and gives the group the grace to act on it before the TERM-then-KILL escalation; `Stop` is now that function with SIGTERM, so there is still one implementation. Tested with a child that answers HUP by saying so on stdout and ignores TERM — with the forward removed, stdout is empty.
- **`46a5a71ec` — the unbounded reap (Medium).** After SIGKILL the helper waited forever, which a child in an uninterruptible kernel wait never ends. It now waits the grace, writes the diagnostic naming the stuck child, returns 124, and skips the survivor sweep so the child is left to init rather than signalled for another grace. Tested with a real grandchild that leaves the group and holds the pipe open — the same blocked-reap shape, no D-state process needed — and with the wait called both ways directly.
- **`46a5a71ec` — the test's build tag (Low).** `boundedlist_unix_test.go` is now `//go:build linux || darwin`, matching where procgroup actually makes groups.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev ./internal/devtool/...`, root Makefile audits, `make lint`, `make lint-generated`, `bash -n`, and both gate paths on the agent module — sharded (18s) and `AGENT_SHARDS=0` (480s) — plus `.` and `envvars`.

obra:
## Review round 4 disposition (bad2f8b20)

CI green. RoboRev (2 of 3 reviewers), all accepted:
- **Medium, unreaped path reports 124 over an interrupt** — preserve `128+signal` whenever `interrupted` is set, after latching any late signal.
- **Low, select race can classify a finished command as timed out** — non-blocking check of `done` in the timeout arm before declaring a timeout.
- **Low, `unreaped` never consulted, so retries stack stuck processes** — an unreapable attempt is terminal: no retry, exit 124 with the diagnostic.
- **Low, `perl` in a default test** — the escapee becomes the test binary re-exec'd as a helper process with its own group; no external interpreter.
- **Low, `evener-dev` file target omits transitive module sources** — prerequisites become every tracked Go source and every module manifest (`git ls-files`), with the untracked-file limit stated in the comment; `make build-dev` remains the forcing path.

obra:
## Round 4 fixes

Head `2a2492497`, all in `e6bff73cc`.

- **An interrupt reported as a timeout** (Medium). The unreapable branch returned 124 regardless, so a SIGTERM the operator sent came back as a timeout and the gate printed the repair-your-caches diagnostic for it. The late signal is latched first and 128+signal wins whenever `interrupted` is set. Test: interrupt plus an unreapable child → 143; with the rescue removed it is 124.
- **The random pick between `done` and the timer** (Low). Both can be ready at the same moment, and `select` chooses between ready cases at random, so a `go list` that finished microseconds before the bound was called a timeout, retried, and reported as 124 on the last attempt. The timeout arm now takes the command's own answer first, through `finishedFirst`, which is tested with the ordering injected both ways.
- **`unreaped` was never read** (Low). The runner retried on `timedOut`, stacking a second stuck `go list` on the volume that already had one. It is terminal now: one attempt, 124, and the diagnostic the attempt already wrote. Test: three attempts configured, one taken, one give-up line, no "retrying".
- **No ambient perl** (Low). The escapee is this test binary re-executed twice, following the repo's helper-process idiom (`internal/devtool/scratch`): the middle process holds the pipe's writer, the grandchild leaves the group with `Setpgid`, outlives the group kill, and exits on its own so nothing has to signal it. No skip.
- **The file target's prerequisites** (Low) widen to every tracked Go source and module manifest: `git ls-files '*.go' 'go.mod' 'go.sum' 'go.work' '*/go.mod' '*/go.sum'`, 3,276 paths in 10-20ms (three runs: 0.02s, 0.01s, 0.01s), once per make invocation. The two trees the subcommands live in were not what the binary is built from — it imports agent, llm, envvars and the rest. The comment says tracked is the operative word: a file not yet `git add`ed does not retrigger, and `make build-dev` forces. Verified: fresh binary → `make -q evener-dev` 0; binary older than its sources → 1; after `make build-dev` → 0.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev ./internal/devtool/...`, root Makefile audits, `make lint`, `make lint-generated`, and both gate paths on the agent module — sharded (66s) and `AGENT_SHARDS=0` (307s) — plus `.` and `envvars`.

obra:
## Review round 5 disposition (2a2492497)

- **Medium, `go list` ignores the build-selection flags the tests use** — accepted: forward tags and `-race` (which sets the `race` build tag) to the enumeration.
- **Medium, curl downloads unbounded** — accepted: connect and total timeouts on our download; the installer's own download runs under `bounded-list` where the binary is present, otherwise the residual is stated in the script.
- **Medium, a `Start` failure leaves an empty log** — accepted: write the error before returning.
- **Medium, `bounded-list` missing from the top-level dispatcher** — accepted: alias it like `module-lint` and `agent-shards`.
- **Low, file target ignores `go:embed` inputs** — accepted: prerequisites become every tracked file the binary can embed, cost measured.
- **Low, `Exists` counts zombie groups as live** — declined: the cost is one diagnostic and at most one grace wait in a window that closes as soon as init reaps, and telling zombies apart needs per-platform process-state probing. Documented on `Exists` instead.
- **Low, no `pgid <= 0` guard in `StopWith`** — accepted.

obra:
## Round 5 fixes

Head `b07211e8f`, all in `b07211e8f`.

- **The enumeration ignored the build-selection flags** (Medium). `go list` and `go test` do not see the same tree: `-tags` selects files and so can select whole packages, and `-race`, `-msan` and `-asan` each set a build tag of their own — a package whose files all sit behind `//go:build race` exists for `make test-race` and not for the enumeration that decides what that run tests. `evener-dev dev list-build-flags` decides which of the caller's flags the enumeration also needs and the gate passes them to `go list`; a value is consumed with its flag first, so `-run -race` stays a regex. Its test builds a fixture module with a package behind a tag and shows the package appearing only when the flag is forwarded. The table overlaps #1266's parser; filed as #1285 rather than resolved across two open PRs.
- **The installer's downloads** (Medium). Ours carries `--connect-timeout 15` (an edge that accepts nothing is the failure this script is about) and `--max-time 300` (five minutes is a stall for a file of a few tens of KB). The installer's own download has no such option, so the whole invocation runs under `evener-dev dev bounded-list` when the binary is there — the lint lane now builds it first, in a step with `timeout-minutes: 5`. Residual, in the script header and on stderr when it happens: `make tools-golangci` on a fresh clone runs before anything is built and takes the unbounded path.
- **A failed start wrote nothing** (Medium). It writes `bounded-list: <err>` to the same guarded writer now; the gate reads that log and nothing else. Test included.
- **`bounded-list` was missing from the top-level dispatch** (Medium), which the usage text and `docs/developing-evener/testing.md:87` both already claimed. Added beside `module-lint` and `agent-shards`, with a test that no alias comes back as an unknown subcommand.
- **`//go:embed` inputs** (Low) join the file target's prerequisites — they are compiled in as surely as a `.go` file. The list is the directories named by the directives in the tree, not a guess. Every tracked file was measured instead and rejected: 10,536 paths, 7,898 after dropping the trees nothing embeds, and `make -q` stats them in 1.06-2.92s per invocation, against 3,346 paths and 0.41-0.57s for the list used; `git ls-files` is 30ms either way. Verified by touching an embedded prompt template and watching `make -q` flip.
- **procgroup** (Low ×2): `StopWith` refuses pgid 0 and negatives as `Exists` already did, with a test that neither waits; and `Exists`'s comment now states the declined case — a group of zombies reads as alive until init reaps them, costing one diagnostic and at most one grace wait, which is cheaper than per-platform process-state probing.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev ./cmd/evener-dev/bin ./internal/devtool/...`, root Makefile audits, `make lint`, `make lint-generated`, `bash -n` on both scripts, `make tools-golangci` against the real network on both paths (bounded, and the fallback, checking the announcement appears only on the latter), and the gate on `.`, `envvars`, agent sharded (60s) and `AGENT_SHARDS=0` (457s).

obra:
## Review round 6 disposition (b07211e8f) — and a split

CI green. RoboRev (3 of 4 reviewers with findings), all accepted:
- **High, `list-build-flags` exit status lost in process substitution** — capture first, fail enumeration on error.
- **Medium, `-args` consumed as a one-value flag** — classification stops at `-args`.
- **Medium, stdout forward ignores write errors** — checked; a short write fails the attempt.
- **Medium, reap goroutine closes `reaped` before publishing `done`** — completion published before `reaped` closes, and `finishedFirst` inspects the result.
- **Medium, `make tools` installs golangci-lint before the binary that bounds it** — `tools-golangci` depends on `evener-dev`.
- **Low, untracked sources do not retrigger the file target** — `git ls-files --cached --others --exclude-standard`.
- **Low, alias test asserts too little** — both spellings compared on one invocation.

Six rounds with findings across all three things this PR bundles is the signal to stop cycling and split, per the rule for this queue. The split, landed serially with no stacked PRs:
1. **Installer download hardening** (curl retry + timeouts, bounded installer): its own small PR first.
2. **`evener-dev bounded-list` + `list-build-flags` + procgroup consolidation** (Go only, with the helper findings above): its own PR second; #1245's Metro-gate port waits on this one.
3. **Gate wiring** (`run-module-tests.sh` through the helper, the `evener-dev` file target, `make tools` order, CI build steps): stays here in #1265, rebased onto main once 1 and 2 land.

obra:
This is insane overbuilding

PR #1266, opened 2026-09-13
fix(evener-dev): build the shard binary with the caller's build flags

obra:
`evener dev agent-shards` compiles the agent test binary itself and then runs it twice. It was compiling with none of the caller's build flags: `scripts/gate/run-module-tests.sh -race -short -count=1` built the shard binary without `-race`, so the gate's race run raced everything except the agent module's largest package, and `-tags`, `-mod`, `-overlay`, `-pgo` and the rest were dropped the same way.

The Go side of #1145, which is closed as superseded. Nothing shell-side comes with it: the gate script has no flag reader, and the parser here is the only one.

## What the parser now gets right

- **Build flags reach the build.** `-tags`, `-mod`, `-modfile`, `-overlay`, `-pgo`, `-compiler`, `-gcflags`, `-ldflags`, `-asmflags`, `-installsuffix` take their value as the next argument; `-race`, `-msan`, `-asan`, `-trimpath`, `-modcacherw`, `-a`, `-linkshared` stand alone; each also in its inline form. The test binary keeps the test flags, translated to `-test.*`.
- **Both spellings.** Go's flag package reads `--tags` and `-tags` alike, so a reader that knows one spelling silently drops the other.
- **`-short` the way go reads it.** Last occurrence wins, through `strconv.ParseBool`: `-short=false -short` is a short run, `-short -short=false` is not. A value go itself would refuse is read as short rather than dropping a caller's short mode over a guess.
- **A flag's value is not a flag.** `-run -race` is a regex of `-race`; classifying every word on its own handed the build a `-race` nobody asked for. The value-taking test flags are consumed with their value before anything is classified, and a value is forwarded exactly as written, because normalising it would rewrite a path or a regex.
- **`-p` is a build flag.** `go help build` documents it; it was on the test side, so the build got none of it.
- **`-C` is refused, not dropped.** It changes directory before the command runs, and everything here is built and tested from the module's own directory.

## Not included

- Every shell-side hunk from the original commits, including the gate script's own flag classifier, and the cross-language table test that compared the two: there is no second table to compare against now.
- `ci: enumerate under the sanitiser flags the tests build with` (42262a2d1) was entirely that shell classifier, so it contributes nothing here. Its finding still stands on its own: the gate enumerates with a plain `go list ./...`, so a package that exists only behind `//go:build race` would be missing from the list `go test -race` is handed. Nothing in the tree is such a package today -- the two race-tagged files, `agent/closegrace_race_test.go` and `agent/race_enabled_test.go`, sit in packages that exist anyway -- so this is a tripwire worth filing rather than a bug to fix blind.

## Gates

`go test ./cmd/evener-dev/...`, `go test -short -count=1 .` (root audits), `go vet -tags evenerfuzz ./cmd/...` and the `GOOS=windows` vet, `make lint` (all 8 modules), and a `MODULES="agent"` gate run, which is the shard planner driving the real agent suite.

## Review round 1

- **Silently dropped flags** (Medium). The tables held what the series had needed so far, so `-cover`, `-covermode`, `-coverpkg`, `-buildvcs` and `-toolexec` fell through the switch without a word: `agent-shards -cover` compiled a binary with no coverage in it. The four tables are now the flags `go help build` and `go help testflag` document on the pinned toolchain, split by destination and by whether the value is the next argument, and a flag in none of them is refused by name -- as is a word that is not a flag, since this runner chooses its own packages.
- **`-exec`, `-o`, `-coverpkg` back in the consumption table** (Low). They left it in the parser rewrite without mention, and their values can look exactly like build flags: `-exec -race` is a program named `-race`. Table cases pin all three.
- **Survey cache keyed on the build** (Low). The key was the test list alone, so a `-race` run reused a survey measured from an ordinary build -- several times faster -- and packed its shards from numbers that never applied. The key now covers the classified build flags and the short-mode bit.
- **`-short=<unparseable>` is a parse-time error** (Low). Reading `-short=yes` as short meant surveying the whole suite and building the binary before every shard died in flag parsing.
- **The `buildFlagPrefixes` comment** (Low) went with the tables it described.
- **The wiring is pinned** (Low). The fixture answers both halves from inside the binary -- a `//go:build race` constant for the compile, `testing.Verbose()` for the invocation -- behind an env gate, so the new e2e case is green with `-race -v`, red when `parsed.build` is dropped from `buildArgs`, and red naming the missing `-test.v` when the test flags are dropped.

One note on the brief: the review asked the e2e case to assert the shards received `-test.run=X`. They cannot. The runner selects each shard's tests itself and hands the regex over `EVENER_SHARD_RUN_FILE`, so a caller's `-run` is consumed and deliberately goes no further; `-v` is the test-side flag that does reach the shards, and it is what the case asserts.

Gates: `go test ./cmd/evener-dev/...`, root audits, `go vet -tags evenerfuzz ./cmd/...` and the `GOOS=windows` repeat, `make lint` (8 modules), and a `MODULES=agent` gate run.


obra:
The `-race` enumeration finding from the old branch (#1145 round 26) is deliberately not fixed here; it is latent today and tracked as #1268.

obra:
## Review round 1 disposition (head 7657173b0)

- **Medium, build-affecting flags silently dropped**: fixed, 11f54f0cb. The tables are complete from `go help build` and `go help testflag`; anything in neither, and any non-flag word, is refused by name rather than dropped.
- **Low, `-exec`/`-o`/`-coverpkg` missing from the consumption table**: restored in 11f54f0cb.
- **Low, survey cache keys ignore build flags**: fixed, a1f6a90ea; keyed on the classified build flags and the short bit.
- **Low, unparseable `-short=<value>` fails late**: a parse-time error now (11f54f0cb).
- **Low, stale `buildFlagPrefixes` comment**: gone.
- **Low, wiring not exercised end to end**: 6e3f16bbb adds an e2e case with `-race -v` through the existing harness, red under both mutations (dropping `parsed.build` from `buildArgs`; dropping the test-side pass-through). It asserts `-v` reaches the shards rather than `-run`, because the runner consumes a caller's `-run` and passes each shard's regex through `EVENER_SHARD_RUN_FILE` itself.

Gates: package tests, root audits, both vets, `make lint`, a `MODULES=agent` gate run.

obra:
## Review round 2 disposition (7657173b0)

CI green. RoboRev (2 of 3 reviewers), all accepted:
- **Medium, flags accepted then dropped** — forward the ones the shard invocation can honor (`-timeout`, `-shuffle`, `-failfast`, `-v` already), refuse the ones it cannot (`-json`, `-bench`, `-fuzz`, coverage output flags) with a clear message; correct the comment.
- **Medium, `-n` not rejected** — refuse it as incompatible.
- **Medium, e2e test needs cgo for `-race`** — drop `-race` from the test's flags; use flags that prove forwarding without a toolchain capability (e.g. `-count=1 -v` plus a `-tags`). No capability probe needed if the test doesn't need the capability.
- **Low, `-v`/`-count` inline values unvalidated** — validate like `-short`.
- **Low, survey cache key omits `GOFLAGS`** — include `go env GOFLAGS` in the key.

obra:
## Round 2 fixes

Head `277893098`, all in `bb816aaff`.

- **Flags dropped without a word** (Medium). Each flag now has one destination and the tables say which. The build keeps its set unchanged. The shard invocation gains `-timeout`, `-shuffle` and `-failfast` alongside `-short`, `-v` and `-count`: the compiled binary honours each as written. Everything else is refused by name with the reason — `-run` and `-skip` (the runner selects each shard's tests; `AGENT_SHARD_SKIP` is how to skip), `-parallel` (`AGENT_SHARD_PARALLEL`), `-c` and `-o` (it compiles and names the binary itself), `-json`, the profile flags and the coverage flags (several shards writing one path), `-bench`/`-benchmem` and the fuzz flags, and `-n`, which prints the build instead of running it. A refused flag's value is still consumed with it, so `-skip -race` is refused for the `-skip` and never reads `-race` as a build flag.
- **`-n`** (Medium) is in that refusal set, with its own sentence: no binary is produced, so there is nothing to shard.
- **The e2e test no longer needs cgo** (Medium). `-race` is gone from it; the build half is proved with `-tags shardfixturetag`, which selects one of two files in the fixture, and the test half with `-count=1 -v` and `testing.Verbose()`. No capability probe, and both mutations still turn it red: dropping `parsed.build` fails on the build tag, dropping the test flags fails naming the missing `-test.v`.
- **`-v` and `-count` inline values** (Low) are checked like `-short`'s: `-v=maybe` and `-count=lots` are refused at parse time rather than after the survey, the build, and the launch of every shard.
- **GOFLAGS in the survey cache key** (Low), read as `go env GOFLAGS` so it is the effective value. It reaches the build without appearing in any argument list the runner can see, so `GOFLAGS=-race` would otherwise pack its shards from a survey measured without it. An unreadable answer is an empty one: the key covers one fewer thing, which costs a survey, never a wrong one.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev`, root Makefile audits, `make lint` (8 modules), and a `MODULES=agent` gate run (52.9s).

obra:
## Review round 3 disposition (277893098)

CI green. RoboRev (2 of 3 reviewers), all accepted:
- **Medium, `-timeout` not forwarded to the survey** — forward `-timeout` to the survey invocation; `-failfast` stays shard-only because a survey that stops early is an incomplete survey, and the comment says so.
- **Low, misplaced doc comments** — fix.
- **Low, `-args` absent from the "exhaustive" tables** — refuse it with a specific reason.

obra:
## Round 3 fixes

Head `b58dfcca5`, all in `2ee8d32c2`.

- **`-timeout` reaches the survey too** (Medium). The shards were given it and the survey was not, which is backwards: the survey is the longest single run of the lot — one binary over the whole suite — so a timeout raised for the shards was needed there first, and a survey that hits the default dies before any shard is packed. `-failfast` stays shard-only, and `surveyArgs`' comment now says why: a survey that stops at the first failure has measured part of the suite, and the packing would read that partial measurement as though it were the whole thing. Pinned by a new case in the `surveyArgs` table test: `-test.timeout=20m` through, `-test.failfast` not.
- **`-args` refused** (Low), so "exhaustive" is true: everything after it is meant for the test binary's own argument list, and this runner launches each shard itself with the flags it needs. It has a refusal reason of its own and a table case.
- **The two e2e doc comments** (Low) had run together — one function's paragraph sitting above another's. Each is back above its own test, and the wiring test's now describes the build tag it actually uses rather than the `-race` it was written with.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev`, root Makefile audits, `make lint` (8 modules). The `MODULES=agent` gate run was still executing when this was pushed — this host keeps suspending between runs, and the survey cache key changed with round 2's GOFLAGS entry, so it is resurveying from cold. Round 2's run of the same path was green, and everything round 3 changed is pinned by the unit tests above; I will say so here if it comes back red.

obra:
The `MODULES=agent` gate run from round 3 finished green: `PASS  agent  64.74s` at `b58dfcca5`. That closes the one gate I pushed without.

obra:
## Review round 4 disposition (b58dfcca5)

CI green. RoboRev (2 of 3 reviewers), all accepted:
- **Medium, test-side flags in `GOFLAGS` silently dropped** — refuse them by name at parse time with a message to pass them on the command line; build-side `GOFLAGS` keep flowing to `go test -c` as today.
- **Medium, `-v` spellings the shard binary rejects** — canonicalize booleans before forwarding (`-test.v` / `-test.v=false`).
- **Low, `go env GOFLAGS` failure aliases the no-GOFLAGS cache key** — fail the run; a toolchain that cannot answer `go env` is not one to survey under.
- **Low, `-count` accepts negatives and `+`** — `ParseUint`.
- **Low, `-timeout` / `-shuffle` unvalidated** — `time.ParseDuration` and `off|on|integer` at parse time.

obra:
## Round 4 fixes

Head `e57a2d799`, all in `0fcaa2945`.

- **Test-side flags in GOFLAGS** (Medium). GOFLAGS reaches the one command this runner gives the toolchain, `go test -c`; the shards are binaries it launches itself, so a `-short` or `-count` there reaches neither and quietly does nothing, while the same flag on the command line works. Any entry in the test tables — forwarded or refused — is now refused by name at parse time, with the sentence saying where it does work. Build-side entries pass through untouched. Table cases for nothing set, build-only, `-short`, `--count=2`, and `-run`.
- **Boolean spellings** (Medium). `-v=1`, `-v=t` and `-v=TRUE` pass `ParseBool` but the test binary's `-v` takes true, false or test2json and refuses the rest after the survey, the build and the launch. Booleans are canonicalised: true becomes the bare `-test.v`, false the explicit `-test.v=false`, for `-short` and `-failfast` too. Case: `-v=t` → `-test.v`.
- **An unreadable GOFLAGS** (Low) no longer falls back to `""`, which is the key for "no GOFLAGS at all" — that fallback would hand one build's survey to another, the exact mistake GOFLAGS is in the key to prevent. The run stops with the error.
- **`-count`** (Low) is read with `ParseUint(value, 10, 64)`, matching `-test.count`'s `flag.Uint`, so `-1` and `+3` are refused.
- **`-timeout` and `-shuffle`** (Low) are read at parse time too — a duration, and off/on/seed — with the same "before the survey" wording as `-short`.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev`, root Makefile audits, `make lint` (8 modules), and a `MODULES=agent` gate run (46.7s).

obra:
## Review round 5 disposition (e57a2d799)

CI green. RoboRev: two Low, both accepted — `-count` validated the way `flag.Uint` parses it (base 0, `IntSize`), and `checkGoflags` refuses the `-test.`-spelled forms so the `GOFLAGS` route and the command-line route agree.

obra:
## Round 5 fixes

Head `2a25d1e3e`, all in `2a25d1e3e`.

- **`-count` now reads the way `flag.Uint` does** (Low): `ParseUint(value, 0, strconv.IntSize)`. Base 0 means `0x2`, `0b10` and `1_0` are counts the shard binaries would have taken and this was refusing; `-1` and `+3` are still not counts. A case each for all five.
- **`-test.*` in GOFLAGS** (Low): `checkGoflags` matched names by their `go test` spelling, so `-test.v`, `-test.timeout=1s` and `-test.count=5` matched nothing and were waved through, while the same flags on the command line are refused. They reach the shards even less than the unprefixed spelling: the toolchain hands GOFLAGS to the commands it runs, and the shards are binaries this runner launches itself. Any `-test.` prefix is refused now, saying the command line takes it without the prefix. Cases for the bare, valued, and would-have-worked spellings.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev`, root Makefile audits, `make lint` (8 modules), and a `MODULES=agent` gate run (65.5s), all in the foreground.

obra:
## Review round 6 disposition (2a25d1e3e)

- **Medium, the caller's `-timeout` also bounds the survey** — declined, with the rule written down: `-timeout` here means what it means to `go test`, a bound on one full run of the package. The survey is exactly that run, so it is bounded the same way a plain `go test -timeout` would be; round 3 forwarded it for that reason. The flag-table comment will state the rule so the two readings stop alternating.
- **Medium, cache key omits `-timeout`** — declined for the same reason: the survey cache stores timings, and a timeout is not an input to those timings. A key that changes with `-timeout` would only force cold surveys. The comment on the key says so.
- **Low, default tests read ambient `GOFLAGS`** — accepted: the harness constructors neutralize `GOENV` and `GOFLAGS`.

obra:
## Round 6 fixes

Head `$(git rev-parse HEAD)`.

- **Accepted, Low: tests inherited the developer's toolchain settings.** `runShards` refuses a test-side flag in GOFLAGS, and every test that calls it read whatever was exported or written with `go env -w`; one `go env -w GOFLAGS=-short` turned `make test` red in eight places, none of which named the cause, and a build flag there would have been quieter still. `isolateToolchainEnv` sets `GOENV=off` and an empty `GOFLAGS`, and every harness that builds a `shardsConfig` or runs the built binary calls it — `e2eConfig`, `startHeldRun`, and the four tests that construct a config inline. Proved by running the package with `GOFLAGS=-short` exported: eight failures before, green after.
- **Declined, `-timeout` bounds the survey**: written into the forwarded-flag table's comment — `go test -timeout` bounds one full run of the package, and each shard and the survey are such runs, the survey most of all, being the whole suite in one binary.
- **Declined, `-timeout` not in the cache key**: written into the key's comment — the key holds what decides the timings (test list, classified build flags, short mode, GOFLAGS); `-timeout` bounds a run without changing what it costs, so keying on it would discard a measurement that still applies.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev`, the package again with `GOFLAGS=-short` exported, root Makefile audits, `make lint` (8 modules), and a `MODULES=agent` gate run (49s). All foreground.

obra:
(correction to the line above: head is `d3a6cf239`, in commit `d3a6cf239` — the shell in my previous comment did not expand.)

obra:
## Review round 7 disposition (d3a6cf239)

CI green. RoboRev: two of four reviewers clean; three Low, all accepted, and two of them are one change: the optional `-test.` prefix is normalized before classification, and `GOFLAGS` entries go through the same tokenizer the command line uses (two-token values consumed), so the two routes cannot disagree again. The build helpers get the same toolchain-env isolation as the run helpers.

obra:
## Round 7 fixes

Head `ca90263ef`, all in `ca90263ef`.

- **The optional `-test.` prefix** (Low). `go test -test.short` is `go test -short`, and neither route took that. On the command line `-test.short` matched no table and was refused as an unknown flag; in GOFLAGS it matched nothing either, which is what round 5's prefix rule was added to catch. The prefix is now dropped in `goFlag`, where `--tags` already becomes `-tags`, so every reader downstream sees one spelling — and round 5's special case is gone: a test-side flag in GOFLAGS is refused by the rule that was already there, under its own name. Cases for both routes.
- **`checkGoflags` classified each word on its own** (Low), so `GOFLAGS='-ldflags -short'` refused a build flag's *value* — the `-run -race` bug again, in the second reader. Both routes now walk the arguments through one `walkFlags`, which consumes a flag's value with it before anything is classified. Cases: `-ldflags -short` and `-tags -short` pass, `-short` alone is still refused.
- **The fixture builders** (Low) join the tests that refuse the developer's toolchain settings: `buildEvenerDev` and `buildShardFixture` run `go build` and `go test -c`, so a GOFLAGS in the environment reached the binaries these tests are about. Proved with `GOFLAGS=-short` exported: the package is green.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev`, the package again under `GOFLAGS=-short`, root Makefile audits, `make lint` (8 modules), and a `MODULES=agent` gate run (21s). All foreground.

obra:
## Review round 8 disposition (ca90263ef)

CI green. RoboRev: one Medium, two Low, all accepted — the `-test.` prefix is normalized only when the remainder is a known test-side flag, so `-test.race` is refused rather than laundered into a build flag; `-vet` moves to the build-flag table since it configures the `go test -c` step; and the doc comments get their blank line.

obra:
## Round 8 fixes

Head `4c8365643`, all in `4c8365643`.

- **The prefix was stripped from every name** (Medium). `-test.race` became `-race` and went to `go test -c` as the race detector — a test-side spelling walked into the build tables. The prefix is dropped only when what remains is a flag the test binary actually has, forwarded or refused; any other `-test.` name keeps its name and is refused as the unknown flag it is. In GOFLAGS it gets its own sentence, since `go test` would hand it to the binary and every shard would refuse it. Cases: `-test.short` forwarded and `-test.race` refused, on the command line and through GOFLAGS.
- **`-vet`** (Low) moves to the build table: it chooses the vet checks `go test -c` runs over the package before compiling it, so it is not the shards' to receive. `-vet=off` now lands in `.build`.
- **The missing blank line** (Low) between `parseFlags`' doc comment and `flagToken`'s.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev`, the package again under `GOFLAGS=-short`, root Makefile audits, `make lint` (8 modules), and a `MODULES=agent` gate run (76s). All foreground.

obra:
## Review round 9 disposition (4c8365643)

CI green. RoboRev (2 of 4 reviewers with findings), all accepted:
- **Medium, `-godebug` missing from the build table** — accepted, and the class closes with it: a test parses `go help build` and `go help testflag` from the toolchain in use and fails on any flag the tables do not classify, so the tables stop being hand-maintained claims of completeness.
- **Medium, `-C` refused on the command line but not through `GOFLAGS`** — accepted; one rule on both routes.
- **Low, `GOFLAGS` split without the toolchain's quoting** — accepted: a surrounding quote pair is stripped before classification.
- **Low, the `-skip` message promises what the cached-survey path does not do** — accepted by making the message true: `AGENT_SHARD_SKIP` reaches the shard invocations as well as the survey.

obra:
## Round 9 fixes

Head `8e95a73c9`, all in `8e95a73c9`.

- **"Exhaustive" is now checked, not asserted** (Medium). A test reads `go help build` and `go help testflag` from the toolchain in use and fails naming any flag no table places. Mutation-proved: remove `-p` from its table and the test names it. That is what catches a flag a later toolchain adds.
- **`-godebug`** (Medium, measured): not on this toolchain. go1.27.0's `go help build` does not list it, and `go build -godebug x=1` answers `flag provided but not defined`. Adding it would forward a flag the build refuses, so it is not added — the test above is what will catch it on the toolchain that has it.
- **`-C` through GOFLAGS** (Medium). It was refused on the command line and waved through GOFLAGS, which the toolchain applies to `go test -c` too, so the build moved out from under a runner already standing in the module's directory. Both readers refuse it from one function now; case included.
- **GOFLAGS quoting** (Low). Split the way the toolchain splits it: a surrounding pair of matching quotes comes off before classification, so `GOFLAGS='"-short"'` is refused rather than invisible. Cases for both quote characters.
- **`AGENT_SHARD_SKIP`** (Low) reaches the shards as well as the survey, which its own documentation already claimed. A cached survey means no survey ran, so the shards were running the very test the operator asked to skip. The test that pinned the old behaviour is replaced by one that pins this.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev`, the package again under `GOFLAGS=-short`, root Makefile audits, `make lint` (8 modules), and a `MODULES=agent` gate run (58s). All foreground.

obra:
## Review round 10 disposition (8e95a73c9)

CI green. RoboRev, all accepted:
- **High, `-v` routed to `go test -c` instead of the shards** — a consequence of round 9: `go help build` lists `-v`, the exhaustiveness test pulled it into the build table, and build classification runs first. The rule becomes explicit: a name that `go test` documents as a test flag is a test flag for this runner even when `go help build` also lists it; the exhaustiveness test accepts either table.
- **Medium, quoted `GOFLAGS` values split into pieces** — round 9's surrounding-quote strip was a half measure; the entries are tokenized the way the toolchain tokenizes them (quotes and escapes respected).
- **Low, one refusal message for every test-side flag** — split by destination.
- **Low, `-v=test2json` refused with a false reason** — refused with the true one: JSON output cannot feed the survey parser.

obra:
## Round 10 fixes

Head `b0df47040`, all in `b0df47040`.

- **The High's premise did not hold, but the rule is now encoded** (High). Round 9 added a test; it moved no name into `buildBareFlags`, and `-v` was never classified build-side. Measured on this head before the change: `parseFlags(["-v", "-tags", "x"])` → `build=[-tags x] test=[-test.v]`, `buildBareFlags["-v"]=false`, and the flags-gate e2e test passes. What was missing is the rule itself, so the classification now runs the test side first with the reason written above it: a name both pages document belongs to the shards, which are what run the tests. Case added: `-v -race` → `-race` to the build, `-test.v` to the shards.
- **GOFLAGS splitting** (Medium) is `cmd/internal/quoted.Split`'s: a quote quotes only when it opens the field. That makes `"-short"` one entry — and it makes the brief's other two examples *not* one entry, which is the toolchain's own behaviour: `GOFLAGS='-tags="a b"'` gives `go: parsing $GOFLAGS: non-flag "b\""` and `-ldflags="-X main.a=1 -X main.b=2"` gives `non-flag "main.a=1"`, both measured on go1.27.0. So I did not make them single entries; that would diverge from the toolchain this reads for. Cases for the quoted entry, the embedded quote, and an unterminated one.
- **One refusal per flag** (Low): a refused flag gets the same reason and alternative on both routes, from one function; a forwardable flag in GOFLAGS keeps its own message saying where it does work.
- **`-v=test2json`** (Low) is refused for what it is: a real value of `-test.v` that the survey cannot read, since it weighs each test from the binary's plain output.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev`, the package under `GOFLAGS=-short`, root Makefile audits, `make lint`, and a `MODULES=agent` gate run (79s). Nothing new merged from main.

obra:
Correction to the round 10 disposition: the implementer measured before changing anything and `-v` was never in the build table (`parseFlags(["-v","-tags","x"])` already routed it to the shards, and the flags-gate test passed). My claim that round 9's exhaustiveness test caused the High was unmeasured. The precedence rule is still worth stating once, which b0df47040 does.

obra:
## Review round 11 disposition (b0df47040)

CI green. RoboRev (2 of 4 reviewers with findings), all accepted:
- **Medium, `-vet` sits in both the build table and the refused table** — accepted, and the class closes: the table test asserts every flag is classified exactly once, so a name cannot straddle two tables again.
- **Medium, `AGENT_SHARD_SKIP` missing from the survey cache key** — accepted: the skip is part of the key.
- **Low, `-fullpath` refused although the shard binary honours it** — accepted: forwarded.
- **Low, stale doc comment on the renamed skip test** — accepted.

obra:
## Round 11 fixes

Head `024807ba0`.

- **`-vet` in two tables** (Medium): stale as filed — round 8 moved it to `buildValueFlags` and it is in no other table, and `-vet=off` classifies to the build today (its case passes on the head you reviewed). What was missing is the guard, which is the durable half of the ruling: a test now fails when any flag name appears in more than one table, mutation-proved by putting `-run` in the build table as well, where it names both.
- **`AGENT_SHARD_SKIP` in the survey cache key** (Medium). A survey taken without the skip measured the skipped test too, so its cost landed in a shard that would not run it and the packing was done from numbers about a different test set. Case: same test list, different skip → different key.
- **`-fullpath`** (Low) moves from refused to forwarded: it changes how the binary prints file names in a failure, which is the same thing in one shard as in all of them.
- **The skip test's comment** (Low) describes the contract it now pins.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev`, the package under `GOFLAGS=-short`, root Makefile audits, `make lint`, and a `MODULES=agent` gate run (78s). Merged `origin/main` now that it is green.

PR #1269, opened 2026-09-13
fix(agent): fence continuations with the target session's generations, and page depth-boundary children directly

obra:
Supersedes #1231.

Three guarantees, one for each commit.

**A continuation is fenced by the generations of the session it names.** A size-trim mint took the delegates generation from the page's own root; a live root reads neither fold cache, so a token naming a closed child carried zeros, and the stale check had to skip validation entirely for live roots to keep that branch readable. What the exemption gave up was the only evidence of that child's journal being rewritten — the root's activity clock moves for what the daemon does under it, never for a file rewritten out of band. `collectActivitySessionEpochs` now returns both generations per session, every mint carries the trimmed session's own, and the `root.live == nil &&` exemption is deleted rather than narrowed. A live target still compares zeros to zeros and is fenced by the revision; a closed child under a live root is fenced by both.

**A depth-truncated delegate reports the session to request, not a token.** The bound is where nothing is loaded, so a continuation there would have to name generations no one read — and every way of predicting them (the last recorded one, the current one, a stat, a probe) was wrong in some case the resume then refused. The token bought nothing either way: a depth continuation names that child as a fresh root at position 0, which is the page a request for the child already returns. So the branch is truncated, carries no continuation, and `JobActivityDelegate.Diagnostics` says `depth limit reached; request session "<id>" directly` — a field already on the wire and already projected. The load stops leaving a placeholder to stand in for a child it never opened. Rendering that diagnostic as something the reader can click is frontend work, tracked in #1270.

**A same-size stat is not proof of a rewrite.** `os.Stat` does not read size and mtime as one snapshot, so during an append it can report the size from before the write with the mtime from after it — indistinguishable from a same-size rewrite. Discarding the fold on that appearance threw away a fold nothing invalidated and moved a generation the content does not justify; CI hit it about one run in four under `-race`. The same-size case asks the tail probe now and moves the generation only when the recorded prefix is gone. This is on the real fold path: it affects every `Get`, not only the predictor.

## Deliberately deleted

Everything #1231 built to predict a generation without folding: `foldcache.Cache.Epoch` and its test hook, `activityPlaceholderEpochs`, `currentHistoricalJobsEpoch`/`currentHistoricalDelegatesEpoch`, `activityChildJobJournalReadable`, `activityChildJournalError`, the placeholder generation fields and the work-unit charge for them, and `markActivityDelegateTruncated` with the depth mint it served. Seven review rounds each found a real hole in that predictor — a journal replaced by a directory, a delete between stat and probe, an oversized line no probe can see — because it was answering a question only a fold can answer. See the reset comment on #1231.

Tests deleted with the code: `TestMarkActivityDelegateTruncated_EmbedsEpochsInContinuation`, `TestLoadSessionJobActivityTree_DepthBoundaryContinuationIsSubmittable`, and #1231's placeholder-generation tests (`TestActivityPlaceholderEpochs_*`, `TestBuildActivityFullSnapshot_DepthPlaceholders*`, `TestBuildActivityFullSnapshot_*BoundaryChild*`, `TestCache_Epoch*`) which were never on main. `TestProjectStableActivityDelegate_DepthTruncated` and `TestLoadSessionJobActivityTree_BoundsRecursionDepth` now assert the new shape, and the latter follows the diagnostic's own advice: requesting that child directly returns the page the bound withheld.

## Gates

Toolchain gofmt; `go vet` host + `-tags evenerfuzz` + `GOOS=windows -tags evenerfuzz` in both modules; `go test -count=1 ./...` for agent, appprojector, the hub server and the TUI with zero non-ok lines; `-race -count=3 -timeout 30m` over the activity tests; `-race -count=10` over foldcache; pinned golangci-lint 0 issues.

Related: #1172, #1195.

## Review round 1

**A same-size stat that kept its tail is re-stated before it is trusted.** The first fix suppressed the generation bump whenever the tail probe matched, which traded a false rejection for a false acceptance: a journal rewritten at the same length whose last 64 bytes survive — the same trailing record, say — kept its generation, so an outstanding continuation was accepted and its `ResumeIndex` landed on content that had changed. The two candidates are a torn append and a same-size rewrite, and a second stat separates them: the append that tore the first stat has completed by then and reports the larger size, while a rewrite still reports the same one. The generation now moves in the second case. `TestCache_SameSizeRewriteKeepingItsTailBumpsTheGeneration` is that case, `TestCache_TornAppendStatKeepsTheGeneration` still holds the other, and both are deterministic.

The comparison is against the recorded size, not the recorded offset: a fold that stopped short of a torn trailing line records an offset below the file's size, so "grown past the offset" is already true without any growth and would re-admit exactly the acceptance this fixes.

**The nested-walk fixture moves generations the way the cache counts them.** It restamped two journals' mtimes to stand in for a rewrite and never checked the result, so it went quiet the moment a bare restamp stopped counting. It now removes each journal, lets the owning cache observe the absence, and restores the bytes, and the walk asserts the minted continuations carry a nonzero jobs generation and a nonzero delegates generation. Deleting either bump turns it red.

**The depth-bound affordance is frontend work, filed as #1270.** This PR keeps the agent behaviour — the token that would feed a "load more" control is the predictor this branch deletes — so the diagnostic names the session to request and the code comment names the issue that will make it clickable.

Also rewritten: three comments that still described deleted code (the budget revision field, the continuation path bound, and the path-length test) now give the rule that remains, which is that the extra path hop is slack the decoder still accepts.


obra:
## Review round 1 disposition (head bbe186e19)

- **Medium, a same-size rewrite whose trailing bytes survive keeps the generation**: real, fixed in 8f497cef6. When the size is unchanged, the mtime moved and the tail matches, the cache re-stats and keeps the generation only if the file grew (a torn append that completed); otherwise it is a rewrite and the generation bumps. The growth check compares against the recorded size rather than the recorded offset, because a fold that stopped short of a torn trailing line already has size above offset and the offset form would re-admit the false acceptance. RED: generation stayed 0 after a same-size rewrite that kept its trailing bytes; GREEN, with the torn-append test still passing.
- **Medium, depth truncation offers no continuation and the frontends have no affordance to open the child**: the agent behaviour stands (a token would reintroduce the generation prediction this PR deletes); the affordance is frontend work, filed as #1270 and cited in the depth-bound comment and the PR body.
- **Medium, the walk fixture restamps mtime and no longer bumps**: fixed, c78ac6a61, with `bumpFoldGeneration` for both journals and an assertion on nonzero jobs and delegates generations (each bump deletion turns it red).
- **Low, stale comments referencing deleted code**: fixed, 6055b93b5.

Gates: toolchain gofmt; host, evenerfuzz and windows vet 0 in both modules; agent, appprojector, hub and TUI suites; `-race -count=3` on the activity tests; `-race -count=10` on foldcache; golangci-lint 0 issues. Merged `origin/main` (468ee7950).

obra:
## Review round 2 disposition (bbe186e19)

CI green. RoboRev (2 of 3 reviewers):
- **High, torn `os.Stat` leaves a pre-append recorded size** — accepted pending a failing test: record the size the fold actually consumed (re-stat or the folded offset) before publishing, and add the regression (torn-stat append, then same-size rewrite with matching tail).
- **Medium, depth-boundary branches expose only a diagnostic** — declined; this is the design ruling for this PR (request the named session directly, no token). Frontend affordance is #1270.
- **Medium, resumed pages can mint paths longer than `activityMaxContinuationPathLength`** — accepted pending measurement: write the test the reviewer describes (`len(Path) >= 2`, tree deep enough to trim). If it fails, validate at mint time and fall back to the "request session directly" diagnostic instead of emitting an undecodable token; do not widen the constant. Fix the comment either way.
- **Low, rewrite+append inside the probe/stat window** — accepted: after the second stat confirms growth, re-run the tail probe at the recorded offset.

obra:
## Round 2 fixes

**1 (High) — `7bc4335c7` `fix(agent): record the size the fold consumed, not the size a torn stat reported`.** Confirmed by test before the fix: after a torn-stat append the published state had `offset` ahead of `size`, a state no honest file produces. The next stat read as growth, the growth path trusted the surviving tail, resumed from an offset the file had already reached, and served the stale fold with its generation intact — `value = {sum:1725 lines:50}, want 3839`. The state now publishes `max(stat size, consumed offset)`. A file that grew *after* the stat rather than during it still records the smaller stat size, so it still reads as growth on the next look rather than as an unchanged length. `TestCache_TornStatAppendRecordsTheSizeItActuallyFolded`.

**2 (Medium) — declined per the design ruling.** No behaviour change; the frontend affordance is #1270.

**3 (Medium) — measured, reproduced, fixed: `5c2cbee8b` `fix(agent): report the session when a page cannot name a path back to it`.**

Measurement first, as asked. `TestLoadSessionJobActivityTree_ResumedPageMintsADecodablePath` builds a 35-session chain, resumes two hops down, and forces a trim at the deepest session. Before the fix the page minted a token its own decoder refuses:

```
the page minted a token its own decoder refuses: continuation path length 34 exceeds 33
```

So it is reachable. The reason is the one the finding names: the depth budget is **depth-relative** — `buildActivityContinuationAt` makes the target projection's own depth 0 however many hops led to it — while a minted `Path` is **absolute**, built from the tree root down through the filtered ancestor chain, so a page resumed `len(Path)` hops down reaches `len(Path)+activityMaxNewDepth`.

Both mint sites now check the path before encoding (`activityContinuationPathFits`); when it will not fit, the session stays `Truncated`, mints nothing, and carries `continuation path limit reached; request session "<id>" directly` — the same shape the depth bound uses, and honest for the same reason: that session is its own root, so its path starts over at zero. The constant is unchanged, and its comment now states that the bound is depth-relative while the minted path is absolute. The mid-list cutoff carries the same guard, covered by `TestMarkActivitySessionTruncated_ReportsTheSessionWhenThePathCannotBeNamed` (a path exactly at the limit still mints and still decodes; one hop past it reports the session).

**4 (Low) — `f034b8910` `fix(agent): ask the probe again after the second stat reports growth`.** The probe runs before the second stat, so a rewrite that also appends can land between them: the stat reports a file longer than the recorded size, which reads as the completed append a torn first stat implies, while the prefix the probe vouched for is gone. The probe is asked again once the length is known. The second stat goes through a field on the cache so a fixture can place the rewrite in exactly that window — nothing outside `refresh` can, since nothing else runs between the two. `TestCache_RewriteInsideTheSecondStatWindowBumpsTheGeneration`; verified red without the re-probe.

`d659acc8c` is a one-line `max()` form golangci-lint's modernize demanded.

obra:
## Review round 3 disposition (d659acc8c)

CI green. RoboRev (2 of 3 reviewers):
- **Medium, clamped size paired with a pre-append mtime** — accepted: when the consumed offset exceeds the stat size, record the mtime observed after the fold (a fresh stat), with the regression the reviewer describes; fix the clamp comment.
- **Medium, missing non-required root journal lets a generation-zero continuation resume into a recreated journal** — accepted pending measurement. Write the test first (mint over a missing root journal, recreate it with entries, resume). If it reproduces, the smallest closing fix is preferred: do not mint a resumable token over a journal that is absent, and treat absent-at-mint versus present-at-resume as a mismatch. No presence/deletion generations persisted in the fold cache; that is the predictor design this PR replaced.
- **Low, stale references to `collectActivityJobsEpochs` and old parameter names** — accepted.

obra:
## Round 3 fixes

**1 (Medium) — `31394ad73` `fix(agent): record an mtime that matches the content the fold read`.** An append landing between the stat and the fold read makes the fold consume a longer file than the stat described; the size clamp followed it and the mtime did not, so the published state described a file that never existed. The next lookup saw the size it expected with an mtime it did not, read that as a same-size rewrite, and discarded a fold nothing invalidated. RED, on a lookup after an append consumed during folding: `generation moved to 1 on a file nobody touched since the fold, want it to stay at 0`. Fixed by re-stating for the mtime in exactly the `offset > info.Size()` branch, where the stat's mtime is known to predate the content. The clamp comment now says what both numbers are and why each is chosen. `TestCache_AppendDuringTheFoldRecordsAMatchingMtime`.

**2 (Medium) — measured, reproduced, fixed: `08569b788` `fix(agent): never mint a position against a job journal that was not there`.**

Measurement first. `TestLoadSessionJobActivityTree_ContinuationMintedOverAnAbsentJobJournal` builds a root with no `jobs.jsonl` and enough delegate content to force a trim, takes the minted token, then creates the journal with three jobs and resumes:

```
minted over an absent journal: resumeIndex=8 jobsEpoch=0 delegatesEpoch=0
job_root_a was never delivered: the token was minted while the journal was absent,
and resuming at index 8 walked past the jobs that appeared since
```

So it reproduces, and all three jobs were lost. The reason is the one the finding names: an absent journal is stepped over without touching the fold cache and reports generation 0, which is also what a journal folded for the first time reports, so "absent when this page was built" and "present and never rewritten" are the same claim.

The fix records the absence where the generations are recorded — `activitySessionSnapshot.JobsJournalAbsent`, carried into `activitySessionEpochs` — and uses it at the three places that need it. No persisted presence or deletion generations in the fold cache; nothing was added to the cache at all. A session whose journal was absent mints no continuation (both mint sites) and says `job journal unavailable; request session "<id>" again`. A resume whose target journal is gone is refused, which is sound precisely because no page mints one over an absent journal, so any token naming that session was minted while it existed. Both halves are covered, and the resume-side refusal is verified red when removed.

**3 (Low) — `ed940ba1d` `docs(agent): name the epochs map the trim actually takes`.** Both comments now name `collectActivitySessionEpochs` and the one per-session map; `grep` finds no reference to `collectActivityJobsEpochs` or the old parameters anywhere in `agent/`.

One fixture change worth noting: `TestCache_RewriteInsideTheSecondStatWindowBumpsTheGeneration` now rewrites only on the first stat and asserts two, since item 1 adds a second, post-fold stat through the same seam. Rewriting under that one would describe a file the fold never saw.

obra:
## Review round 4 disposition (ed940ba1d)

- **Medium, delegates-journal absence undetected** and **Medium, TOCTOU between the presence check and the cache read** — accepted together as one change: `foldcache.Get` reports absence itself (no separate stat), and both journals carry that bit the same way. One encoding for both journals, so this class of finding closes.
- **Medium, a root without `jobs.jsonl` can never paginate** — accepted; this is a consequence of round 3's "no token over an absent journal". The continuation carries an absent-at-mint bit per journal, and a resume is refused only when presence changed.
- **Medium, depth-truncated delegates unreachable from the frontend** — declined, as in round 2: design ruling, frontend affordance is #1270.
- **Low, diagnostics accumulate per dropped entry** — accepted: set once per branch, deduplicated in `markActivitySessionTruncated` as well.

obra:
## Round 4 fixes

**A. Mediums 1, 2 and 4, one change — `b02e5d0b5` `fix(agent): read journal presence from the fold and carry it in the token`.**

`foldcache.Result` gained `Absent`, set from the same stat the fold reads through, so presence and the fold are one read and the separate existence check is gone — that closes the TOCTOU. Both journals take the bit from that one source: jobs through `loadCachedJobRecords`, delegates through `scanRootDelegateState` and the root delegate index. There is no second encoding of presence anywhere; `activitySessionSnapshot` and `activitySessionEpochs` carry `jobsAbsent`/`delegatesAbsent` side by side with the generations.

The token carries `JobsAbsent` and `DelegatesAbsent`, and the resume compares them exactly as it compares the generations, refusing with the existing "the underlying journal changed" rejection when presence differs. Round 3's refusal to mint over an absent journal is replaced by this, which is what fixes Medium 4: a root with no `jobs.jsonl` mints, resumes, and pages its delegates to the end. Before, every page after the first was unreachable forever, since the journal was never going to appear.

Tests, written first and all four red before the change:

- `TestLoadSessionJobActivityTree_PaginatesWithNoJobJournal` walks a root with no `jobs.jsonl` to exhaustion and asserts every delegate is delivered exactly once (ii, and Medium 4).
- `TestLoadSessionJobActivityTree_RefusesWhenJournalPresenceChanges` covers the three mismatches: the job journal appearing after the mint, the job journal deleted after the mint, and the delegate journal deleted after the mint (i and iii).
- The old `..._ContinuationMintedOverAnAbsentJobJournal` is folded into these, since its shared fixture is now `seedAbsentJournalActivityRoot` and its assertion is the walk (iv).

**B. Medium 3 — declined again per the design ruling; no change. #1270.**

**C. Low — `0ffd31fd1` `fix(agent): say each branch diagnostic once`.** Both places now go through one guarded append. Measured on the depth-bounded page whose deepest session loses eight entries: eight copies of the same sentence before, one after, asserted in `TestLoadSessionJobActivityTree_ResumedPageMintsADecodablePath`.

obra:
## Review round 5 disposition (0ffd31fd1)

CI green. RoboRev: three of four reviewers clean; one reviewer's two Mediums are the same class (a rewrite landing between two looks at the file), which more probes cannot close. Disposition:
- **Growth path does not revalidate the consumed prefix** — accepted in the form that closes the class: one open file handle per refresh, with size (`fstat`), tail probe and incremental read all taken through it. If journal rewrites are rename-based, that handle is a consistent snapshot by construction and the second-stat/re-probe steps become redundant and go. If rewrites are in place, the cache's contract states the tolerated window instead; the implementer measures which before choosing.
- **Fast-hit path trusts stat attributes** — accepted only as the cheap form: the probe already opens the file, so it takes the length from the handle it holds. Stale-attribute filesystems are otherwise out of scope for a local journal written by the same process, and the doc comment says so.

obra:
## Round 5 fixes

**The measurement first, because it decides the shape of the fix.** Both journals are only ever appended to or truncated **in place, on the writer's own open handle**. No rename, no compaction, nowhere:

- `agent/internal/jobstore/store.go:660` — `recoverTrailingJSONLineLocked`, `s.f.Truncate(offset)`, drops an incomplete trailing line at open.
- `agent/internal/jobstore/store.go:757` — `rollbackAppendLocked`, `s.f.Truncate(startOffset)`, rolls a failed append back.
- `agent/internal/delegatestore/store.go:159` — truncate to 0 for a torn version header at open.
- `agent/internal/delegatestore/store.go:175` — truncate an unterminated trailing batch at open.
- `agent/internal/delegatestore/store.go:213` — `rollbackLocked`, truncate on a failed append.
- `grep` for `os.Rename` finds none over either path, and neither package has any compaction.

So the rename restructure does not apply: there is no inode swap to make a handle a snapshot of, and a rewrite here always shrinks the file before it regrows. That is branch 3 of the brief, and it is what `39eabb199` does.

**`39eabb199` `fix(agent): decide a cached-fold hit from the handle the probe read through`.** No new probes. `probeTailFromHandle` opens once and returns both answers from that handle — the length and mtime via `f.Stat()`, the recorded bytes via `ReadAt` — and `tailProbeMatches` is now a thin wrapper over it, so there is one reader. Both places that can return a folded value without reading anything (the fast hit, and `refresh`'s same-size true hit) now require the handle's own length and mtime to match the recorded state.

That is not a free narrowing of the second Medium, it changes outcomes: an append landing between `Get`'s stat and the probe leaves the recorded tail exactly where it was, so the probe matched and the stale length said nothing had grown, and the caller was handed a fold missing everything the append added. `TestCache_AppendRacingTheStatIsNotServedFromTheCachedFold` pins it, red without the handle checks with `value = {sum:1180 lines:40}, want ... (sum 1725)`.

**The window that remains is now stated on `Cache`** rather than chased: a truncate and regrow to the same length with the same trailing bytes, inside one handle's own fstat-to-ReadAt window. No writer above produces that, no further probe closes it — every check is itself two reads, so each new one opens a window of its own — and locking is not available, since the writers are separate processes holding nothing a reader could take. The doc says plainly what it costs when it happens: a generation that does not move, so a continuation is accepted and its position applied to changed content.

No locking added. The torn-stat and same-size-rewrite tests are untouched and still green.

obra:
## Review round 6 disposition (39eabb199)

CI green. RoboRev (2 of 4 reviewers with findings, two clean):
- **Medium, `drop` advances the tombstone epoch on every absent read, so a journal that existed and was deleted livelocks pagination** — accepted: `drop` becomes idempotent once a path is tombstoned, absent results return the epoch they were judged by, and the walk test folds, deletes, then pages to exhaustion.
- **Medium, continuation format unchanged while its fields changed** — accepted as an explicit invalidation, not a legacy path: the version is bumped so every pre-change token is refused with the ordinary stale-continuation error and the client restarts once. No compatibility decoding.

obra:
## Round 6 fixes

**1 (Medium) — `89bb878d0` `fix(agent): treat a deletion as one event, and report the generation it is judged by`.**

Two halves, and they only work together. `drop` advanced the stored generation on every `Get` of a missing path, while the absent read reported 0 — which says nothing has ever happened to this path. Reporting the tombstone honestly without fixing the advance is exactly the livelock the finding describes, and the end-to-end test shows it: with `drop` reporting but not idempotent, `TestLoadSessionJobActivityTree_PaginatesAfterTheJobJournalIsDeleted` dies on the second request with `page 2: activity continuation is stale: the underlying journal changed`, because each request refuses the continuation the previous one minted.

So losing a fold advances the generation once, `epochState` carries a tombstone marker, every later look at the same absence returns the same number, and `Result.Epoch` carries it so the page that mints and the request that resumes compare the same value. Both journals get this from the same `drop`, since they are the same cache.

`TestCache_DeletedPathReportsOneStableGeneration` pins the whole sequence — fold, delete (generation moves, absent read reports it), four more looks (unchanged), recreate (not absent, content read, and the count does not restart at 0). RED on the old code at the first step: `generation 0 after a delete, want it past 0`.

A recreate deliberately does not advance again: the deletion was the discard, and what separates "absent at 1" from "present at 1" is the presence bit the token already carries.

**2 (Medium) — `6d6403e43` `fix(agent): bump the continuation format version`.** `activityContinuationVersion = 2`. Every other version is refused as stale, so a client restarts pagination once; no legacy decoding path, and the constant's comment says why it moves whenever field meanings move. `TestDecodeActivityContinuation_RefusesAnEarlierFormatAsStale` asserts a version-1 token gets the stale rejection and that this build's own token still decodes. The four negative fixtures that used version 1 to test *other* validation failures now mint the current version, so they still fail for the reason they claim.

The stale rejections share one constructor now, so a caller can recognise the class without matching on which particular fact stopped being true.

**Note on the agent suite here.** This branch is merged up to `origin/main` @ `33b553278`. 28 tests in `agent` fail on this machine, every one of them a skills-lifecycle test from #1168, all the same shape: `envelope source = "/private/var/..." want "/var/..."`. `t.TempDir()` hands back `/var/folders/...`, macOS resolves that symlink to `/private/var/...`, and the envelope carries the resolved spelling while the expectation does not — which is why CI, on Linux, is green. Filed as #1297. Nothing in the failing set touches the fold cache or the activity tree, and that whole area is green here, lint included.

One trap worth passing on: `-run 'Continuation'` also matches `TestSkillDelivery_ResponsesContinuationPlanning`, so a filtered activity race run picks up one of those failures for reasons unrelated to continuations.

Gates: toolchain gofmt; `go vet` host + `-tags evenerfuzz` + `GOOS=windows -tags evenerfuzz`, 0; `golangci-lint` `0 issues.`; `cmd/evener-tui` and `internal/appprojector` clean; `-race -count=3 -timeout 150m` on foldcache clean; the same over the whole agent root, whose failure set is exactly the 28 above and nothing else.

obra:
## Review round 7 disposition (83e30e0fa)

- **Medium, depth/path truncation without a continuation** — declined, as in rounds 2 and 4: design ruling; the client affordance is #1270.
- **Medium, a tombstone is evaluated as a zero-byte file on recreation** — accepted: the content-comparison branches and the rescan counter apply only to non-tombstone state; an absent state folds the new file from byte 0 and keeps its generation, pinned by the stable-generation test extended to recreation.
- **Low, doc comment attached to the wrong helper; stale slack rationale** — both accepted; the slack is explained by this build's own `len(Path) + activityMaxNewDepth` arithmetic.

obra:
## Round 7 fixes

**1 (Medium) — `fd08ce28b` `fix(agent): do not compare a recreated file against a tombstone`.** A tombstone records that the path was gone; its size, mtime and tail describe nothing, and the comparison branches were reading them as content. An empty file appearing at a tombstoned path is the exact trap the finding names: length 0 matches the tombstone's 0, only the mtime disagrees, so it read as a same-size rewrite, stat'd again, and bumped a generation the deletion had already moved. The same state also satisfied `wasCached`, so the first read of a path that had just come back was counted as a full rescan of something the cache no longer held.

`describesContent := st != nil && !st.absent` now guards both the comparison branches and `wasCached`. A tombstoned path folds from zero, keeps the generation the deletion gave it, and counts as the first touch it is.

`TestCache_DeletedPathReportsOneStableGeneration` grew both recreates. RED on the empty one before the fix:

```
generation 2 after an empty file appeared, want the absence's 1
 -- the deletion was the discard; a tombstone has no content to compare this against
```

It also asserts `FullRescans` counts 0 for that first fold after the tombstone, and 1 for the later append onto the cached empty fold — so the counter still means what it says.

**2 and 3 (Low) — `4cdf6fcc4` `docs(agent): put the path-limit rationale on the function that enforces it`.** The `activityContinuationPathFits` paragraph had drifted onto the dedup helper below it; it is back on its own function, and the dedup paragraph stays where it is.

The path bound's rationale was stale in a way worth correcting rather than trimming: it explained the extra hop as room for tokens minted by earlier builds, which version 2 now refuses on version before their length is ever read. The extra hop is this build's own arithmetic — a page resumed one hop down spends the whole budget below its target and names an absolute path of exactly `activityMaxNewDepth+1`, so a decoder capped at `activityMaxNewDepth` would refuse a token this service had just handed out. `TestMarkActivitySessionTruncated_ReportsTheSessionWhenThePathCannotBeNamed` pins both sides, and the comment now says so.

Merged `origin/main` @ `f7816dd8d`, which carries #1301, so the macOS symlink failures are gone and the agent suite is clean here.

obra:
## Review round 8 disposition (4cdf6fcc4)

CI: the `web` job failed on this Go-only head; the failing test is being read (the guard's driver tail now reaches the job output via #1309) and the job rerun.

RoboRev (2 of 4 reviewers clean):
- **Medium, truncation without a continuation** — declined for the sixth time under the ruling; #1270 is the client affordance.
- **Medium, the `ErrLineTooLong` fallback reports generation 0 and present, so an old continuation can be accepted against an unreadable journal** — accepted: a journal that cannot be read refuses continuations while the degraded read stays.
- **Low, a zero-value `Cache` now panics on the nil stat function** — accepted: `os.Stat` is the fallback.
- **Low, the `JobsJournalAbsent` comment describes the mechanism this series replaced** — accepted.

obra:
Rerun of the `web` job at 4cdf6fcc4 passed; the failure was `TestSkillComposerBrowser` timing out at the steered-turn step (#1308), not this PR. CI is green at this head; round 8 findings stand as dispositioned above.

obra:
## Round 8 fixes

**1 (Medium) — `5a3d33f20` `fix(agent): refuse a resume when the delegate journal stopped being readable`.** The degraded read folds nothing, so the generation it reported was invented, and 0 is also what a journal folded once and never rewritten reports. RED before the fix:

```
resumed a token minted while the delegate journal was readable; the page it resumes into
has no delegates at all, so its position counts entries that are not there
```

The condition now travels with the generation from the one read that observed it — a small `activityJournalState{epoch, absent, unreadable}` replaces the growing tuple that `scanRootDelegateState` and `loadHistoricalStableActivity` were returning — and a resume compares it exactly the way it compares presence. The degraded page itself is untouched: `TestLoadSessionJobActivityTree_RefusesAResumeAfterTheDelegateJournalBecomesUnreadable` asserts the fresh page is still served with its `delegate_journal_line_too_long` diagnostic, and only the resume is refused.

I carried the bit in the token rather than refusing every resume over an unreadable journal, because a blanket refusal is the defect rounds 4 and 6 already fixed twice: a session whose journal is stably unreadable would never page past its first request. A mismatch is what is unsafe, not the state itself.

The token gained a field, so `activityContinuationVersion` moves to 3 under the rule round 6 wrote down — a version that stays put while fields change lets an older token be read as saying something it never said. Earlier tokens are refused as stale and the client restarts once; the duplicate-hop fixture that pinned a `"v":2` token was moved to 3 so it still fails for the reason it claims.

**2 (Low) — `7ebcc6614` `fix(agent): let a zero-value cache stat`.** Caching disabled is a documented, inert configuration and a zero-value `Cache` is the simplest way to reach it, but it leaves the stat indirection nil and the uncached path called it directly — a nil-func panic, which is how `TestCache_ZeroValueReadsWithoutCaching` fails without the fix. Every look at a file goes through `statFile`, which falls back to `os.Stat`.

**3 (Low)** — the `JobsJournalAbsent` comment now states the mechanism: the fold reports the generation and the absence from one stat, and nothing steps over the cache to learn it. It rides with the first commit, next to the field it describes.

obra:
## Review round 9 disposition (5a3d33f20)

CI: the `web` job failed on this Go-only head again; the step is being read from the driver tail and the job rerun (#1308).

RoboRev (all four reviewers on the Medium):
- **Medium, the two work-unit truncation mints build their epochs without the unreadable bit, so a session with an unreadable delegate journal and more than the page's jobs loops forever** — accepted, and the class closes with it: the snapshot gains one `epochs()` constructor that every mint site uses, so no field can be left out again; the missing case (mint while already unreadable) is the test.
- **Low, the fixture helper's comment describes a same-size restamp as a non-event when the cache treats it as a rewrite** — accepted.

obra:
Rerun of the `web` job at 5a3d33f20 passed; the failure was `TestSkillComposerBrowser` at the capability-loss step (#1308, now diagnosed from the uploaded artifacts), not this PR. CI is green at this head; round 9 findings stand as dispositioned above.

obra:
## Round 9 fixes

**1 (Medium) — `f501ed98d` `fix(agent): mint every continuation from one place`.** RED before the fix, on a walk over a session whose delegate journal is unreadable from the start and stays that way:

```
page 2: activity continuation is stale: the underlying journal changed
 -- the delegate journal was unreadable when this token was minted and is unreadable still,
    so nothing about it changed
```

That is the loop the finding describes, and it is permanent: the condition does not change between requests, so every retry refuses the token the previous page minted.

The missing bit is the symptom; the shape is the defect. Both work-unit cutoffs assembled `activitySessionEpochs` by hand, so they were frozen at the fields that existed when they were written. `activitySessionSnapshot.epochs()` is now the only way to build that state, and every mint takes it — both truncation sites, `collectActivitySessionEpochs` (which feeds `mintActivityTrimContinuation`), and nothing else. After the change `grep 'activitySessionEpochs{'` over `agent/` finds the constructor's own return and test fixtures, nothing more, so the next field added reaches every mint at once.

`TestLoadSessionJobActivityTree_PagesWithAnUnreadableDelegateJournal` drives the work-unit path specifically — more shell jobs than the budget renders — since the trim path was already covered and this is the site that was wrong.

**2 (Low) — `77ec65575` `docs(agent): say why the fixture removes the journal instead of restamping it`.** The comment claimed a restamp does not move the generation "and should not". It does move it: same length with a moved mtime is the ambiguous case, the recorded trailing bytes survive a restamp exactly as they survive a rewrite that ends the same way, so the cache takes its second stat, sees the length unchanged, and calls it a rewrite. A fixture built on that would be staging the cache's answer to an ambiguity rather than the deletion it means to stage — which is why the helper removes the journal and writes it back.

PR #1274, opened 2026-09-13
fix(tool): keep integer literals beyond int64 exact in fingerprints

obra:
Pays down the second deferred finding from #829's review.

## The defect

`canonicalNumber` folded a JSON number by trying `ParseInt`, then falling through to `ParseFloat`. Integers that overflow int64 therefore went through float64 and were **rounded**, so two distinct integer arguments canonicalized to the same bytes, produced the same semantic failure fingerprint, and shared one failure run.

Verified before changing anything:

| argument pair | canonical form before | after |
| --- | --- | --- |
| `9223372036854775808` / `9223372036854775809` | both `9223372036854776000` | distinct |
| `18446744073709551616` / `18446744073709551617` | both `18446744073709552000` | distinct |

Values that fit int64 were never affected (`9007199254740992` and `9007199254740993` stay distinct), because `ParseInt` runs first.

## The fix

An integer-shaped literal (optional sign, then digits, nothing else) that overflows int64 is returned as its exact `json.Number` text instead of being rounded. `json.Number` marshals as the raw literal, so the value stays exact and distinct. The `ParseInt` path is still first, so everything fitting int64 is unchanged, and `ParseFloat` still serves literals that are not integer-shaped.

## Evidence

`TestFailureFingerprint_HugeIntegersStayDistinct` fails against the pre-fix code, reporting all three pairs (including a negative pair) collapsing to one fingerprint, and passes after. `TestFailureFingerprint_Int64AndFloatPathsUnchanged` pins that the int64 path stays exact across the 2^53 boundary, that `42`, `42.0` and `4.2e1` still fold together, and that `1.5`/`1.50` fold while `1.5`/`1.6` stay distinct.

Gates: `go test ./agent/internal/tool/` ok, `-race` ok, `go test ./agent/` ok (330s), `go build ./...` clean, `gofmt` clean, `golangci-lint` 0 issues.

The remaining deferred finding, the dead exact-store failure state, is separate work and is not touched here.


obra:
Merging with the Medium recorded as a deliberate, documented limitation.

The maintainer chose not to fix the suffixed-spelling collision here. An earlier revision of this branch fixed it exactly with `math/big`, and review correctly caught that `big.Rat.SetString` materializes a token's exponent before `IsInt` can be consulted: `1e1000000` (11 bytes) expands to a million digits. I measured it at ~19ms and ~3MB per token against ~3us through `ParseFloat`, and because `newDispatchKey` runs before `ValidateRawArguments` and schema validation on a body capped at 2MB, a crafted or prompt-injected call could have burned hours of CPU. That revision was reverted.

The remaining gap is therefore bounded and explicit rather than an oversight: suffixed spellings of integers beyond int64 (`.0`, `e0`) still round through `ParseFloat`, so they do not fold with the bare form of the same number (safe: the breaker fires later), and two distinct such values written with the same suffix can still collapse to one fingerprint (harmful: a changed call can be parked). The doc comment on `canonicalNumber` states both, separates them by severity, and warns against reintroducing exact parsing without bounding the exponent first.

Post-revert verification: a 34.9KB body of 2000 `1e1000000` tokens fingerprints in 2.317ms through the real `failureFingerprint` path, so the amplification is gone by roughly five orders of magnitude.

A follow-up is the right place for the suffixed case, via bounded exact parsing or a raw-token fallback for precision-losing forms.

PR #1279, opened 2026-09-14
fix(skills): materialize bundled skills once, content-addressed

obra:
## Problem

Bundled skills were extracted to a fresh `os.MkdirTemp` directory on every process that initialized a session, built the hub slash catalog, or read a past thread, and nothing removed them. On one dev machine this had accumulated to 1165 directories / 123 MB.

## Change

Publish the embedded skills once into a directory named for the sha256 digest of their contents under `os.TempDir()`, and reuse the published copy across processes.

- A binary whose embedded content changed publishes beside the old copy under its own digest, so a session already reading the old directory keeps a stable path.
- An occupant of the published name is adopted only when its content matches the digest (re-digested, not trusted from the name).
- A publish that cannot use the shared name falls back to a private extraction, so a session always gets its skills.
- The digest walk rejects non-regular and oversized entries before opening them; the published name lives in a shared temp dir, so a foreign occupant could be a FIFO that blocks its open or a device that never ends.
- `EmbeddedSkills`/`EmbeddedSkillsDir` now share one guarded cache, and a stale lifecycle comment claiming session close removes an embedded-skills directory is corrected.

The cache stays under `os.TempDir()` rather than the config or cache root because `~/.config/evener` is sandbox-denylisted and a restricted session's file tools read only the worktree plus temp.

## Why not delete extraction entirely

The bundled `doctoring-evener` skill is not self-contained: its `SKILL.md` and the doctor persona direct the model to read `references/*.md` and `runbooks/*.md` relative to the skill directory, and those paths reach the model through `use_skill` and the prompt catalog. Removing the directory would break those reads. This change keeps a real path while making it a one-time, bounded publish. (The `runbooks/` half is already served from the embedded FS by `ParseRunbookFromFS`.)

## Verification

- `go build ./...`, `go vet`, `gofmt`, `golangci-lint run ./agent/skill/...` (0 issues)
- `go test ./agent/ ./agent/skill/ ./agent/doctor/ ./cmd/evener-hub/ ./cmd/evener-doctor/ ./internal/bundled/`
- `go test -tags evenerfuzz ./agent/skill`
- Coverage ratchet for the agent module passes within its tolerance
- End-to-end: two separate processes left the directory count unchanged and reused one stable `evener-skills-<digest>` directory instead of creating a fresh one each run

The second commit is a review pass (consolidated cache, digest guard). The `os.CopyFS` reuse suggested during that pass was reverted because a fuzz fault-injection fixture pins `fs.ReadFile` semantics.


obra:
## Round verdicts on the `c43925c40` review

Three findings from the review of `c43925c40`. Two are fixed in `d8a7e6c69`, one is accepted as a documented tradeoff.

**Fixed: Windows fallback cache created in shared temp without ACL.** `defaultSkillsBaseRoot` now reports an error when the per-user cache directory cannot be named, and base resolution propagates it, so the cache fails closed instead of landing in an unverifiable shared temp root. The randomized private base is still used where the platform's ownership and permission checks can verify it (the squatted-name case on Unix).

**Fixed: the move-lease test was vacuous.** `TestEmbeddedSkillsDir_MovesTheLeaseToTheNewCopy` now holds the first copy's lease across the base switch and asserts the production path releases it; it fails with "moving copies left the old copy's lease held" if retention is forced. The no-root case has its own test asserting the fail-closed error.

**Accepted: retained fallback leases expire by age, not by session lifetime.** Superseded fallback copies are retained because sessions hold `SkillFile` paths inside them and read them on demand; earlier reviews on this PR flagged the opposite, releasing the lease while those readers were still live. Retention is bounded by `protectUntil`, the same 30-day staleness window the age reaper already treats as proof that no reader remains, and a sweep releases expired leases on the next resolution. Tracking exact reader lifetimes would require a new retain/release API plus per-session wiring through discovery, activation, and reload, which is out of scope for this PR. The tradeoff is documented at the `heldLeases` field in `agent/skill/builtin_skills.go`. Protection is strictly stronger than the pre-fix behavior and bounded, so we are keeping this policy and ask reviewers to accept it.

obra:
## Round verdicts on the `7afdcd8` review

Two findings from the review of `7afdcd8`; one fixed in `9c0f5367e`, one accepted as a documented tradeoff.

**Fixed: junctions and reparse points could be treated as directories.** The fallback reaper and the copy reaper now classify entries with `dirInfo`, which rejects a symlink or Windows directory junction even when `IsDir` reports true. Such an entry is unlinked with `Remove` rather than traversed or removed recursively through, so a planted link can neither be read through nor collect anything it points at. A Unix test pins the invariant: a link planted under a fallback base name is unlinked and its target survives. The Windows reparse-point path itself is verified by inspection and cross-build only, because this environment has no Windows test host.

**Accepted: a retained staged copy is cached as a verified copy.** When the published digest name is occupied by something unusable, materialization hands back a copy staged under the retained prefix, and the caller caches it as verified rather than as a fallback with a retry window. That is deliberate: the copy is digest-verified and leased like a published one, while retrying the occupied name would re-stage the whole tree and keep a fresh copy per retry for as long as the name stays occupied. The accepted consequence, recorded at `retainStagedCopy`, is that the process keeps serving the retained copy even if the published name later heals.

obra:
## Round verdicts on the `9c0f536` review

Both findings fixed in `aaa74079a`.

**Fixed: the fallback reaper unlinked every non-directory entry matching the prefix.** That was too broad: a fresh regular file could be removed without any age or ownership check. Unlinking is now restricted to symlinks and reparse points, which are the entries that can shadow a real base; a regular or special file under the same prefix is left alone. The Unix regression test plants both a link and a regular file under the prefix and asserts the link is unlinked, the file survives, and the link's target is untouched.

**Fixed: the private base was resolved fresh on every call.** When the predictable name is unusable, `defaultEmbeddedSkillsBaseDir` now resolves the randomized base once per process and reuses it while it still exists, instead of minting a new `MkdirTemp` base per call. That removes the accumulation path where a retry published into a new base and retained the previous copy's lease. The Unix regression test squats the predictable name and asserts two resolutions return the same private base; it fails with "private base changed between calls" when the caching is removed.

The suggestion to also cap `heldLeases` is noted and deliberately not taken: retained leases are bounded per outage epoch now, and the age-based retention window (`staleRetainedMaxAge`, the same window the reaper itself uses) is a deliberate, documented tradeoff recorded at the `heldLeases` field.

obra:
## Round verdicts on the `aaa7407` review

All three findings fixed in `045523404`.

**Fixed: a verified copy was trusted on `Lstat` existence alone.** Resolution now compares the directory's identity (`os.SameFile`, so device and inode on Unix, volume and file index on Windows) with the identity recorded when the copy was validated. A replacement at the same path is rejected and rebuilt without re-digesting the tree on every resolution. The regression test renames a copy aside, puts an empty directory in its place, and asserts resolution returns real skill content rather than the replacement.

**Fixed: the Unix temp root is now validated before the cache lives in it.** A root is trusted when it carries the sticky bit, or when this process owns it with no group or other write; anything else fails closed with an error instead of placing a per-user cache there. The regression test opens a temp root up to other users, asserts resolution refuses it, then makes it sticky and asserts it is accepted.

**Fixed: retained fallback leases are capped.** The list is bounded at `maxHeldLeases` (8), dropping the oldest lease first, so a base that flaps between unusable and usable cannot accumulate descriptors over the staleness window. The regression test retains more leases than the cap and asserts the count and which end was dropped; it fails with "held leases = 11, want 8" when the cap is removed.

One note on scope: the residual `touchDir` call follows a path that was verified immediately beforehand, so a swap in that instant is not covered by a no-follow operation. Closing it fully needs descriptor-based timestamp updates, which this change does not include.

obra:
## Design change: the fallback tier is deleted

Rounds of review kept converging on one corner of this cache: the code that published a *private fallback copy* when the shared per-user base could not be built, plus the leases retained for superseded fallback copies, their cap, their expiry, the private-base cache, and the trust validation that went with them. The maintainer's call was that this tier is overbuilt for what the cache is for, so it is gone.

What the cache does now:

- One content-addressed copy per digest under the validated per-user base, published by staging plus atomic rename, leased shared while this process reads it, with a digest re-check, an identity check against replacement at the same path, and an age reaper that skips leased copies.
- When no shared copy can be resolved, the process extracts the tree once for its own lifetime (`evener-skills-process-*`), which is exactly what `EmbeddedSkillsDir` did before the shared cache existed: no lease, no reaper, no removal, because sessions read those `SkillFile` paths for as long as the process runs and nothing ever deletes that directory.
- An unusable per-user base now fails closed (error) instead of silently landing in a shared temp name it cannot verify.

This removes the fallback bases, `removeFallbackBase`, `leaseFallbackBase`, `reapStaleFallbackBases`, retained copies (`copy-*`), `heldLeases` and its cap and expiry sweep, the private-base cache, and the fallback retry window, along with their tests: 1098 lines deleted, 309 added across ten files. The open findings from the last review that lived in that tier (idle retention, the Windows deletion race) are gone with it; the ancestor-chain trust check that was also open is implemented here (`ensureTrustedRoot` now walks the chain, accepting sticky or non-other-writable directories owned by this user or root).

Gates on this head: `go test ./agent/skill/` (plus `-race` and `-tags evenerfuzz`), vet, golangci-lint, `go build ./...`, and cross-builds for windows, darwin, freebsd, netbsd, openbsd, solaris, illumos all pass, along with the affected package set and the agent coverage ratchet.

obra:
## Round verdicts on the `f995ad3` review

Three of the four findings are fixed in `d628e3de9`; one is declined with evidence.

**Fixed (High): the degraded extraction bypassed the root trust checks.** `embeddedProcessSkillsDir` now verifies the temp root with the same `ensureTrustedRoot` the shared cache uses, and fails instead of extracting into a root that was just refused as replaceable by other users. The regression test opens a temp root up to other users, asserts the degraded path fails there, then makes it sticky and asserts it extracts.

**Fixed (Medium): the process-lifetime copy was never revalidated.** It is now guarded by a mutex and an existence check rather than a `sync.Once`, so a temp cleaner that removes it mid-process causes a fresh extraction and invalidates the scanned metadata, instead of every later resolution returning a dangling path. The regression test removes the copy and asserts the next resolution returns a new, populated directory with matching metadata; it fails when the existence check is dropped.

**Fixed (Low): TOCTOU between the usability check and the lease claim.** The adopt and publish paths now re-digest once the shared lease is held, so a copy replaced in that window is not served. This closes the race rather than the reported symptom; like the identity re-check it mirrors, the window itself is not deterministically testable.

**Declined (Low): "symlink rejection follows links via `Info()`".** `fs.DirEntry.Info` is defined to report the link itself, not its target ("If the entry denotes a symbolic link, Info reports the information about the link itself, not the link's target"), so `Mode().IsRegular()` already rejects a symlinked occupant. Directory links are rejected separately by `entryIsDir`, and the existing tests cover both (`TestDigestSkillsFS_RejectsIrregularEntry` with a `ModeSymlink` fixture, `TestMaterializeEmbeddedSkills_HealsSymlinkedPublishedName`). No change is needed.

obra:
## Round verdicts on the `d628e3d` review

All three findings fixed in `a26c66a67`.

**Fixed (Medium): the process copy was valid when only its directory existed.** The process-lifetime extraction is now validated by digest before reuse (`cacheDirUsable` against the embedded tree's digest, computed once per process), so a cleaner that removed files but left the directory triggers a fresh extraction and invalidates the scanned metadata. The regression test deletes one skill directory from the process copy and asserts the next resolution returns a complete tree; it fails when the digest check is dropped.

**Fixed (Low): the no-locking platform comment was stale.** `lease_other.go` now says what actually happens there: `tryExclusiveLease` refuses the no-op lease, so a reaper never holds the exclusive lease it needs and nothing is reaped on that platform.

**Fixed (Low): tests depended on the ambient host temp root.** `TestEmbeddedSkillsDir_CachesAcrossCalls` and `TestEmbeddedSkills_CachesFilesystemBackedMetadata` now use the `pointEmbeddedSkillsAtBase` seam with a `t.TempDir()`, so they cannot depend on whether the host's `TMPDIR` chain is sticky. The two tests named alongside them (`TestDefaultEmbeddedSkillsBaseDir_RefusesAnUntrustedTempRoot` and `TestEmbeddedSkillsDir_RefusesAnUntrustedTempRootForTheProcessCopy`) set `TMPDIR` themselves to a controlled directory and are deterministic already; the fail-closed production behavior is unchanged.

obra:
## Round verdicts on the `a26c66a` review

Three of the four findings are fixed in `3d32293a0`; one is declined with reasoning.

**Fixed (Medium, and a real availability bug): a symlinked temp root disabled the bundled skills.** A root that is itself a symlink (macOS `/tmp` to `/private/tmp`, or a `TMPDIR` pointed at a link) was rejected by the trust check, so both the shared cache and the process copy failed and discovery recorded an `unreadable_source` diagnostic. The root is now resolved with `filepath.EvalSymlinks` before validation, and both the cache and the extraction are created under the resolved directory, never through the link. Two Unix regression tests cover the shared path and the degraded path; both fail when the resolution is reverted.

**Fixed (Medium): the metadata scan raced a re-extraction.** `embeddedProcessSkills` now holds `processSkillsMu` across directory selection and scanning, with a bounded re-check that a scan is never stored against a directory that was replaced, and an uncached return if a copy keeps losing its directory. Public behaviour is unchanged.

**Fixed (Medium): the Windows process copy used an unverifiable temp root.** `processCopyRootTrusted` returns false on Windows, so the degraded copy fails closed there instead of extracting into a root where ownership and ACLs cannot be checked, matching the shared cache root's behaviour. On Unix it is the same predicate as the shared root's, and the "other" platforms keep their documented per-user-temp convention.

**Declined (Medium): "verified-cache fast path checks directory identity, not contents".** That is the deliberate `verified` design: a copy this process published, living in a per-user base that was validated as owned and private under a trusted root, is not re-digested on every resolution. The integrity boundary for that copy is the base's privacy plus the trusted root chain, which is why a replacement at the same path is caught by the identity check; the remaining writers are same-user processes, which are not a security boundary. Re-digesting every serve was removed on purpose as the defensive cost this change is meant to avoid, and I am not reintroducing it.

obra:
## Round verdicts on the `3d32293` review

Two of the three findings are fixed in `fa3e4b121`; one is declined with reasoning.

**Fixed (Medium): the verified path trusted directory identity, so a gutted copy was served.** The cached skill files are now checked on every resolution in the verified path (`cacheDirComplete`: one `Lstat` per skill this process handed out), and a copy that lost a file is forgotten and republished. That closes the case an age-based temp cleaner creates when it removes stale files but leaves the directory, which had left bundled skills silently gone for the rest of the process. The regression test removes a `SKILL.md`, asserts the next resolution returns a complete tree, and fails when the check is dropped.

**Fixed (Medium): any sticky root was trusted, including one owned by another user.** A sticky directory protects this process's entries only from users who do not own it; its owner can still replace them. `tempRootTrusted` and `ancestorDirTrusted` now require a sticky root or ancestor to be owned by this process or by root, which keeps the `/tmp` shape trusted and refuses a sticky `TMPDIR` owned by someone else. The regression test drives the predicates with synthetic stat info (sticky owned by another uid refused; sticky owned by root or us accepted; private root owned by us accepted, by another user refused) and fails when the owner requirement is dropped.

**Declined (Medium): process-lifetime extraction directories are never reaped.** This is deliberate and matches what the process did before the shared cache existed. The directory is one per process, only created when no shared copy can be resolved, nothing in the package reaps it because a session may read its `SkillFile` paths for the life of the process, and the same OS temp cleaner that the first finding invokes removes such entries by age, so the accumulation is bounded rather than permanent. Reaping it ourselves would reintroduce exactly the lease-and-reaper machinery this change removed.

Two gate notes on this head: the agent coverage ratchet passed on a re-run after its first attempt failed to build the profiles (`cannot find package net`, no statements measured) — that was an environment failure, and I cleaned 3.3 GB of my own test debris from the scratch directory before it went green at 94.3% union against the 94.7% floor, inside the 0.5pp band. The affected package set and all package gates passed.

obra:
The three open review items from this PR are now tracked as issues:

- #1414 — decide whether the verified cache path must validate the full file manifest (the medium from the `fa3e4b1` review)
- #1415 — stop repeating the full publish path when the shared cache is permanently unusable (the low from the same review)
- #1416 — skip the degraded-copy tests on Windows, where the degraded copy fails closed (the high from the same review)

Everything else raised across the review rounds was either fixed on this branch or declined with evidence in the round verdict comments above.

Issue #1200, opened 2026-09-12
agent: a compaction fold's retained tail is not replayed on restart; resumed history is the summary only


obra:
Raised as a High in #1100 round 13; reproduced identically on `origin/main` (`eee90925a`), so it predates that PR and is out of its scope (part of #1116).

**Defect.** After a fold, the live history keeps the retained tail verbatim, but a restart resumes only the fold's SUMMARY: with twelve durable turns and a single fold, `ResumeHistory` yields the summary alone. With two folds, a turn recorded during fold 1 (and copied by fold 1) that fold 2 retains is present live and absent on resume. Cause: the fold's rewrite set is `persistedAppendLog[snapAppends-base:]` (`agent/session_compaction.go:164,195`) — pairs appended *since this fold's snapshot*, i.e. turns recorded during the fold — while the anchored branch of `ResumeHistory` (`agent/transcript_read.go:144-146`) discards everything before the latest marker. Retention has always been carried by the checkpoint/summary prose, not by entries, so a fold that retains verbatim turns diverges live from durable.

**Fix shapes (both change every session's behaviour; design pass first).** (a) The rewrite set becomes every turn the newly published history retains, changing the size and meaning of a fold run that `foldRun`, the anchored branch, ATIF, the fork-context filter and both projections read. (b) `ResumeHistory` keeps pre-marker entries the fold retained, which needs a new durable field, i.e. a transcript format change. Add a successive-fold restart regression either way.

Related: #1150 (publication ordering), #1181 (publish-before-durable audit), #1152 (grouping residuals).

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

**Verdict:** `legit-current` — a fold publishes `[summary, ...preserved recent turns]` (`agent/internal/contextmgr/context_manager.go:936-963`; `PreserveRecentTurns` default 6 at `:275`), but the transcript rewrite set only re-appends the persisted pairs recorded since the fold's snapshot (`agent/session_compaction.go:188`, written at `:217-221`); the preserved pre-marker turns are never re-written. `ResumeHistory` anchors on the last compaction marker and discards everything before it (`agent/transcript_read.go:146-173`), so on restart the summary survives and the retained tail does not — live history diverges from durable history, exactly as described.

**Autonomy:** `needs-human` — both fix shapes change every session's behavior (the rewrite set's size and meaning, or a transcript-format field); this is the design pass the issue itself calls for.

[...]

Issue #1201, opened 2026-09-12
selfupdate: installdirs tests fail under plain go test on macOS (/private/var vs /var TMPDIR)


obra:
Seen repeatedly while gating #1100 (part of #1116). Six tests in `internal/selfupdate` (`installdirs_custom_test.go`, `installdirs_symlink_test.go`) compare `/private/var/...` against `/var/...` and fail when `go test ./...` runs directly on macOS, where `TMPDIR` is a symlink under `/var`. They pass in the canonical runner, which normalises the scratch path, and in CI. Reproduced on a detached `43d93cb46` (main-equivalent) with nothing in the package touched. The handoff's rule is to keep the canonical TMPDIR normalisation rather than bend tests to an incorrectly invoked runner, but a package that fails under bare `go test` on the platform most contributors use is a broken window: either `filepath.EvalSymlinks` the fixture root in these tests, or have the package's tests refuse to run outside the canonical runner with a message that says so.

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

**Verdict:** `legit-current` (mechanism; not reproducible on this Linux host) — `InstallDirsFromExecutable` resolves the executable with `filepath.EvalSymlinks` and derives prefix/shareBinDir from the resolved path (`internal/selfupdate/update.go:208-213`). The tests build fixtures under `t.TempDir()` and compare results against the unresolved `root`, e.g. `installdirs_custom_test.go:19,38-45` and `installdirs_symlink_test.go:18-44`. On macOS `TMPDIR` sits under `/var`, which resolves to `/private/var`, so the resolved prefix differs from the fixture root and the equality checks fail. Missing: a macOS runner to execute the failing case here; the causal chain is confirmed from the source.

**Autonomy:** `needs-human` — the issue frames a policy choice (resolve the fixture root in tests vs refuse to run outside the canonical runner). The first is the minimal, test-only option.

**Recommendation:** `needs-decision` — resolving each fixture root through `EvalSymlinks` is the smallest fix and leaves the canonical-runner TMPDIR normalisation intact.

obra:
Same family, seen once on the #1100 branch at round 16 (2026-09-13): `TestInitInside_SymlinkSpelledCwdCanonicalizesStoredPathAndLockKey` failed on `/private/var` vs `/var`, clean in four runs since; untouched by that branch.

PR #1281, opened 2026-09-14
Persist a fold's records and replay tail durably (1/3 of #1100)

obra:
**Guarantee**: a compaction fold's own records and the turns recorded while it ran survive a restart exactly once, and the fold publishes nothing the transcript did not take.

**Seam**: this piece *writes and gates records*; the piece after it *names them*. Every record here is written, tagged and gated on what landed; no turn-ownership rule changes, so the two projections keep grouping exactly as they do on main today.

### What lands

- **`agent/schema/turn.go`** — the durable forms: `CONTEXT_COMPACTION` and `ROUND_TIMINGS` kinds, their payloads, `ContextReplay` and `CompactionFoldID`.
- **`agent/session_compaction.go`** — the fold writes a record per compaction layer, re-appends the persisted forms of the turns recorded while it ran just ahead of its markers (tagged as copies, stamped with the fold id), and publishes per marker: each carries its own landing result, and the fold's steering and transcript reminder ask separately whether a reader coming back finds an anchor at all.
- **`agent/transcript_read.go`** — resume reassembles the fold's run from the anchor outward; with no anchor the copies are dropped as the duplicates they are.
- **Readers that must not double-count the copies** — the public transcript (`session_tools_transcript.go`, which also keeps the fold id out of model-facing output), the fork snapshot (`delegate_fork_context.go`), the ATIF export (`internal/atif`), the item projections (`internal/apptranscript`).
- **`agent/session_events.go`** — a hook completion's entry is written before its event, and a failed write announces nothing.
- Presentational kinds are transparent where conversation is what matters: provider history, continuation deltas, attention windows, signals, search.

### Deliberately absent

- **Turn ownership** (piece 2): `OwningTurnID` on timing/hook/steering records, self-minted turn names, the hook-owner threading, the projector's owner-aware announcements. **The round's timing record ships with that piece, not this one** — a round's timings are published as the turn ends, so the record can only sit inside its turn once the turn has a name; landing it here breaks live/cold parity for unnamed turns. The kind and its payload are defined here because they are the same durable form as the compaction record.
- **Grouping** (piece 3): the fragment/continuation rules, `continuationGroupState`, the turn-index version bump, the replay-copy exclusion from usage/failure totals, the doctor's reconstruction of these records.

Supersedes the durability third of #1100; B (ownership) and C (grouping) follow once this merges.

## Round 1 fixes

Head `827a94914`.

- **HIGH, replay copies projected twice** (`f0c9f8b9b`). `ProjectTurn` returns
  nil for a copy, at the one point every reader projects through — the full
  read showed the copied turn twice, and the bounded reader's group item counts
  came from a scan that had counted it. RED: *the copied user turn appears 2
  times in the full projection, want once*, with the paged walk asserted beside
  it.
- **Aggregate totals** (same commit). The usage, derived and failed-tool scans
  share one `countsTowardTotals` rule and skip a copy's measurements while
  still reading the copy, so a tool call announced in one still names a later
  result. Physical ordinals are untouched. The paging stat now counts the
  records it projected.
- **Hook completing before the writer exists** (`742136786`). Its entry is
  queued, so its announcement is queued with it and released by the flush that
  makes it durable, riding the session-start envelope rather than preceding it.
  A failed flush releases nothing. RED asserts no event while queued, none
  emitted by the flush itself, the entry present after it, and nothing
  announced when the flush fails.
- **The fold's history turns** (`827a94914`). The injected steering carried the
  fold id only on the written form; live history has it now too. RED: *live
  history has the fold's steering under fold "", the transcript under
  "fold_01M2EY4…"*.
- **`TurnRoundTimings` has no producer here — by design**, and the kind's
  comment now says so: a round's timings are published as the turn ends, so the
  record can only sit inside its turn once the turn has a durable name. The
  writer lands with the naming piece.
- **Salvage window** excludes `ROUND_TIMINGS` alongside `CONTEXT_COMPACTION`.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; agent, server, apptranscript, appprojector,
doctor and TUI suites clean; `-race -count=3` on agent (198s), apptranscript and
the TUI packages; golangci-lint 0 issues in both modules.

## Round 2 fixes

Head `1cb66f0b6`.

- **Parallel queues → one queue** (`4316839aa`). A held turn is now one value
  carrying what it owes: its announcement, whether its entry must be fsynced,
  and the in-memory half of its pair. The fork delegate's boundary queues
  through the same door, so nothing can shift another turn's announcement onto
  it. New test: two SessionStart hooks around a boundary whose write fails —
  both hooks still land, in order, with their own entries, and the boundary's
  failure is reported once.
- **Hook entries are fsynced** (same commit), at the direct write and at the
  attach flush. New test counts the writer's fsyncs with the sync interval set
  to an hour: a buffered append cannot move that counter, and the announcement
  only follows a write that did.
- **A queued hook's history append waits for its write** (same commit). The
  history append and the pair-log entry are part of what the flush releases, so
  a failed flush leaves the completion in no history, no log and no event —
  asserted in both directions.
- **Replay copies leave grouping and stamping alone** (`3d7c1b488`). Each
  reader drops copies where its scan hands entries on: the accumulator on the
  full side, the record loop on the indexed side. RED showed both halves — a
  group counting the copied round's usage twice, and `bounded keys=[…turn_5:2:0]
  full keys=[…turn_5:1:0]` where a copied user turn opened a group on one side
  only. The physical record stays indexed, since its extent is what lets a scan
  resume and answer divergence.
- **Nil history guard** (`1cb66f0b6`).
- `TestAttachTranscriptWithNilWriterDrainsWithoutPanic` now queues through
  `holdTurnUntilTranscriptReady` instead of assigning the slice.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; agent, server, apptranscript, appprojector,
doctor and TUI suites clean; `-race -count=3` on agent (193s), apptranscript
(112s) and the TUI packages; golangci-lint 0 issues in both modules.

## Round 3 fixes

Head `3ad307a2b`.

- **HIGH, turns lost across an un-anchored fold** (`e7f1b130d`). The pair log
  was pruned at publish, before the copies and the marker were written, so a
  fold whose copy write failed left its turns covered by nothing *and* the log
  that named them already gone; the next fold's marker discarded them for good.
  RED: copy write fails, second fold succeeds, resume — *the turn recorded
  during the un-anchored fold appears 0 times in the resumed history, want
  exactly once*. **Chosen fix: keep the pending forms until the anchor is on
  disk** (prune after the writes, and mark from where when a fold could not
  anchor, so the next fold's rewrite reaches back there). The alternative —
  recovering unanchored turns by re-reading the transcript — needs a read
  inside the publication transaction and a way to pair live turns with entries,
  which is the same identity problem #1200 is open on; this one is bookkeeping
  the session already has.
- **Durable append that keeps its entry** (`3e3c0409f`). **Measurement**: the
  durable door already removes the entry on a sync failure whose rollback
  succeeds (truncate), and only the `!removed` branch leaves a record — which
  the writer knew (it spends the sequence number there) but reported as a
  generic `ErrRollbackFailed`. So the distinction existed in fact but not in
  the API. Added `transcript.ErrEntryRetained`, wrapped at exactly those
  points; a completion whose error carries it is committed and announced, every
  other failure drops the turn as before. Both branches tested.
- **`TurnRoundTimings` producer**: unchanged, by design for this piece.
- **Comment placement** (`3ad307a2b`).

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; agent, server, apptranscript, appprojector,
doctor and TUI suites clean; `-race -count=3` on agent (233s), agent/transcript,
apptranscript (116s); golangci-lint 0 issues in both modules.

## Round 4 fixes

Head `e5f40d3db`, with `origin/main` (`33b553278`) merged as `e5f40d3db`.

- **HIGH, the preserved suffix dropped on restart**, and **HIGH, a marker-less
  fold cleared the un-anchored debt** — one fix (`c70c3b509`), because they are
  one question. RED first, both across a reload: *"recorded turn 6" is in the
  live history the fold published but appears 0 times in the resumed history*,
  and *the turn the un-anchored fold left behind appears 0 times in the resumed
  history ... the marker-less publication in between took the debt with it*.
  The rewrite now copies the persisted form of every pair-log entry whose turn
  **the published history still holds**, and the log is pruned to exactly that
  set. One rule covers all three cases: the turns recorded while the fold ran,
  the preserved suffix a compaction keeps verbatim, and the turns an earlier
  un-anchored fold left standing. A marker-less publication becomes harmless by
  construction — it discards nothing, so it drops nothing from the log — and a
  turn the fold *removed* (a superseded environment record) is left behind with
  it, so a copy can never restore context the fold dropped on purpose.
  Round 3's `unanchoredAppendLog`/`unanchoredAppendFrom` and the
  `persistedAppendLogBase`/`snapAppends` ordinal arithmetic are gone: the
  published history answers what they were tracking. I tried the smaller fix
  first (count the preserved suffix back over the log). It resurrected
  superseded environment turns — `TestEnvironmentContextResetIsAtomicWithFold
  Publication` and `TestRestoreEnvironmentReconcilesCompactionWithStale
  Metadata` caught it — because counting cannot survive a fold that also
  removes turns from the middle of its own window.
  The two forms of one turn are matched by the identity the persisted form
  inherits from its live turn when the session builds it (kind + mint
  timestamp). Turns with no pair entry at all — restored from an earlier
  process, attention-owned — still cannot be carried past a marker; that is the
  durable per-entry identity #1200 is open on.
- **M, the flushed path and a retained entry** (`6fd020a04`). RED:
  *announcements waiting for the envelope = 0, want 1: the transcript holds the
  retained entry*, driven through `retainEntryFS` on `attachTranscript`'s
  flush. `writeHeldTranscriptTurn` now applies the rule once for both the
  direct write and the flush, and reports whether the entry is a record in the
  file, so `emitHookCompleted` no longer re-reads the error to decide the same
  thing a second time.
- **M, compaction announcements standalone in the live projection**
  (`42554382a`). RED: *live and reload group a compaction differently: live
  `[[userMessage systemMessage:context_compaction] [userMessage]]`, cold
  `[[userMessage] [systemMessage:context_compaction] [userMessage]]`* — the
  test compares turn keys and items on both sides, and they now agree exactly
  (`turn_1`, `turn_2`, `turn_3`). The live projection opens and closes the
  announcement's own turn, the shape `EventEnvironment` already uses; with no
  turn running it keeps the gap bucket every announcement shares. It also
  re-announces the thread as active behind the closed turn: a compaction
  happens mid-round, and a client that reads "the active turn completed" as
  "the session went idle" (the TUI does, kata s8x8) would otherwise offer to
  send into a session still working. The wire shape changed for a live session,
  so `cmd/evener-tui` has a case for it.

### Merge of #1168 (`c09997369`), resolved semantically

One conflicted file of substance, plus one mechanical:

- **`agent/session_compaction.go`** — three hunks, both sides kept.
  1. `commitTranscriptsLocked`: main's signature is `func()` and stamps
     `commit.receipt` into every pending compaction turn's `SkillState`; mine
     is `func(anchor bool)` with the CONTEXT_COMPACTION records, per-marker
     landing and the anchored-gated steering write. Resolved as my signature
     with main's receipt stamping placed **before** the marker loop, so every
     compaction turn still carries the receipt whether or not its marker is
     withheld.
  2. Commit construction: main moved `commit := &foldCommit{}` above
     `commitTranscriptsLocked` (its closure reads `commit.receipt`); mine
     constructed it below with `foldID`. Resolved by keeping main's position
     and giving it `foldID`, with my `writesCompactionMarker` assignment left
     where it was.
  3. `foldCommit`: both field sets, main's `actualCompaction`/`captured`/
     `stagedCompactionCount`/`claimCompactionLocked`/`receipt` alongside my
     `foldID`/`writesCompactionMarker`/`commitTranscriptsLocked(anchor bool)`.
  Neither behaviour is gated on the other: the publication claim and the
  handoff receipt run inside the publish's `s.mu` section as main wrote them,
  before any transcript write, and `reconcileSkillCompactionReceipts` scans
  ALL entries — including pre-marker ones — so a withheld anchor cannot lose a
  receipt. No logic changed in the merge.
- **`agent/transcript_read.go`** — both new functions kept side by side, my
  `foldRun` and main's `reconcileSkillCompactionReceipts`; textual only.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; golangci-lint 0 issues in both modules; root
suite (13m) and agent suite (9m); `-race -count=3` on agent (49m),
apptranscript, appprojector and the TUI packages — no data races.

Everything that failed on this machine is a macOS TMPDIR-symlink artifact, and
the two runs prove each other: under the default `TMPDIR` the 28 skills tests
#1168 added fail on `/private/var` vs `/var` path assertions (#1297) along with
`internal/selfupdate` (#1201); under a non-symlinked `TMPDIR` all of those pass
— including #1168's compaction/reload tests over this merge — and the only
failure is `TestSessionInfraRootsRefuseHomeAncestorsAndShallowRoots`, which
asserts the opposite direction of the same symlink. Nothing in the union
touches this branch's code.

## Round 5 fixes

Head `ee4dcc892`, `origin/main` (`bcb8406d9`) merged as `7a9df9fea` (one
conflict, `agent/session_init.go`: main's `conversationSignals` switch now
skips `TurnNotesContext` as well, and this branch's `TurnRoundTimings` /
`TurnContextCompaction` join it in the same switch).

- **A — HIGH + two Mediums, one rule** (`feb7d20df`). A replay copy is a second
  physical record of a turn the file already holds, and the index kept its
  extent *and* its turn kind, which put it on the logical map three ways: the
  record after it read its group boundary from the copy (a copy of a standalone
  kind closed the group its neighbours belong to), a copy inside a group's span
  was read, projected and stamped by both grouped-projection paths, and a copy
  that started a group of its own spent an entry ordinal the full read never
  allocates. The record now carries the rule — `indexedTurn.Replay`, its turn
  kind left off — and every reader of the logical map skips it: grouping, the
  turn count, the open-group reconstruction, and both projection paths (the
  bounded one no longer has to decode an entry to find out). RED first, all
  three shapes as subtests of one table, each asserting bounded == full and
  then paging one item at a time: *bounded reader and full reader disagree over
  a replay copy* (duplicate stamps inside a span; `turn_m1 + turn_4` where the
  full read has one turn). Plus a guard that a compaction's copies leave the
  keys and positions a client already holds exactly where they were.
  No `turnIndexVersion` bump: `ContextReplay` arrives with the fold that writes
  the copies, so no index built before this branch can hold a record the flag
  would have been true for.
- **B — measured, nothing changed.** An active turn with two persisted entries,
  then a compaction: live names the announcement `turn_2`, cold names it
  `turn_3`. They diverge, but not over standalone allocation — live numbers
  `turn_<n>` by turns STARTED while `persistedTurnID` numbers by ENTRY INDEX,
  so any turn with more than one entry pushes cold's namespace ahead. It is not
  this record: the same probe with a `HOOK_COMPLETED` (untouched here) gives
  live `turn_1` and cold `turn_3`. And the narrow fix collides — numbering only
  standalone records by logical-turn order gives a standalone at logical
  position N the id `turn_N` that an opener at entry index N already owns. The
  sound fix renames every daemon-named turn in every existing transcript, needs
  the index version bump and re-derives the `SeedPersistedTurns` fence: that is
  turn naming, piece B. Filed as **#1303** with the measurement table so it is
  not lost.
- **C — every layer its own turn** (`002adf683`). RED: *live and reload
  disagree over 2 compaction layers: live `[{turn_system [context_compaction
  context_compaction]}]`, cold `[{turn_1 …} {turn_2 …}]`* — the later layers
  fell into the gap bucket the first had just left. Every `EventContextCompaction`
  now opens and closes its own turn, and the thread is re-announced active only
  when a turn really was interrupted, so a compaction between turns no longer
  claims work that is not happening. Live and cold are compared by turn key and
  items for two and three layers; the TUI has a case for the multi-layer frames,
  and the two projector tests that asserted the old single-frame shape now
  assert the turn lifecycle.
- **D — a retained marker is a marker** (`e7a1c4dd5`). `entryIsRecorded` is the
  one rule, used by the fold's marker loop and `writeHeldTranscriptTurn` alike.
  RED with a marker-targeted retaining FS (the line lands, the sync fails, the
  buffered door rolls nothing back): *compaction-turn events = [CHECKPOINT],
  want the retained SUMMARY marker's among them*. The reload anchors on that
  marker, and the warning is still reported.
- **E — resume puts the steering back where the fold held it** (`773ccc601`).
  The anchored branch now reassembles anchor and records, then copies, then the
  fold's `TurnSteering`. The round-4 order assertion is deleted and replaced by
  a live-vs-resume equality test on the provider message sequence
  (`expandHistory` over both), which fails with the old order as *resumed: […
  note handoff, recorded turn 7 …] live: [… recorded turn 11, note handoff]* —
  an assertion that cannot encode a wrong order because it has no order of its
  own.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; golangci-lint 0 issues in both modules; root
and agent suites; `-race -count=3` on agent (43m), apptranscript,
appprojector and the TUI packages. Still failing on macOS only, all
TMPDIR-symlink path assertions that never reach this code: `internal/selfupdate`
(#1201), and `agent/plugin` + `agent/skill` (#1301 fixed the `agent` package's
own skill tests; `skill_sources_test.go` and `discovery_test.go` compare
unresolved paths the same way — noted on #1297).

## Round 6 fixes

Head `f78050454`. `origin/main` unchanged since round 5's merge (`bcb8406d9`),
so no new merge commit.

- **1 — HIGH, a fork lost what its parent's compaction copied** (`d30d57802`).
  RED first through a real delegate fork: parent records a turn, folds (copy +
  marker), forks — *the turn recorded during the parent's fold is missing from
  the child's inherited history*. `snapshotDelegateContext` filtered the copies
  out before `session_init` replays the snapshot through `ResumeHistory`, which
  anchors on the parent's marker and discards everything before it. The filter
  is gone; both resume branches were already right. Two things the first fix
  missed and the test caught: the projected turn has to keep `ContextReplay`
  AND `CompactionFoldID`, or the copies arrive unmarked and the untagged-anchor
  path drops them anyway. The subtest for a parent that never compacted forks
  exactly what it forked before, and the old snapshot test now asserts the run
  is carried whole and the child's replayed history holds the round once.
- **2 — M (three reviewers), one write-then-announce answer** (`1387aed99`).
  `entryOutcome(err) (recorded bool, report error)` is now the decision for
  every record kind, and the two answers are independent: a retained entry is
  recorded AND reported. The fold's context-compaction records and its injected
  steering both gated their event on `err != nil`, so a layer or a steering
  line the transcript holds — and a reload shows — was announced to nobody. RED
  with a record-targeted retaining FS: *CONTEXT_COMPACTION events = 1 for 2
  record(s) in the transcript* and *STEERING_INJECTED events = 0 for 1
  record(s)*. The remaining `writeErr != nil` gate,
  `appendUserInputTurnRefusingPoison`, is a different rule (user input is
  never dropped, poison refuses the turn) and is left alone.
- **3 — L, the buffered door reports a retained entry** (`c4010df58`).
  `poisonLandedBytesLocked` already knew a whole line had landed and the error
  threw that answer away, so the same failure meant different things at the two
  doors. RED both sides: `Append` now carries `ErrEntryRetained`, and
  `writeHeldTranscriptTurn` commits a turn retained by a buffered write
  (*recorded=false committed=false for an entry the transcript holds*).

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; golangci-lint 0 issues in both modules; root
and agent suites; `-race -count=3` on agent, agent/transcript, apptranscript,
appprojector and the TUI packages. The only local failures remain the macOS
TMPDIR-symlink path assertions that never reach this code: `internal/selfupdate`
(#1201) and `agent/plugin` + `agent/skill` (noted on #1297).

## Round 7 fixes

Head `b3fb8d96d`, `origin/main` (`f7816dd8d`) merged as `46495fd84` — clean, no
conflicts; #1308 also retired the last two macOS path failures
(`agent/plugin`, `agent/skill`).

- **1 — measured RED, resume follows live** (`30c4fd71f`). The live-vs-resume
  equality test now has the missing case: a pinned note AND a turn recorded
  while the fold ran. Observed order, before any change:
  - live: `[summary, recorded 7…11, note handoff, recorded while the fold was running]`
  - resume: `[summary, recorded 7…11, recorded while the fold was running, note handoff]`

  So round 7's reviewer is right and round 5's case (steering, no concurrent
  turn) was the half that agreed by coincidence. The published history is the
  fold's own result — marker, preserved suffix, injected steering — with the
  merge-back after all of it, and the run on disk is written in a different
  order on purpose: every copy goes down before the marker, so a crash cannot
  leave an anchor with nothing to replace what it discards. Nothing in a copy
  said which half of the run it was, so the fold now says it —
  `ContextReplayMergedTail` marks a copy of a turn recorded while the fold ran
  — and the anchored branch puts the two groups back around the steering. The
  fork snapshot carries the mark with the rest of the run. Both cases green.
- **2 — pinned, not fixed** (`b3fb8d96d`).
  `TestCompactionReplay_RestoredTurnsHaveNoPersistedFormToCopy` walks record →
  restart → compact → restart and asserts what happens today: the restored
  turns have no pair-log entry, so the fold writes no copy for them and the
  second restart comes back without them. The test names #1200 and fails
  loudly if the behaviour changes; the case is described on #1200 itself
  (comment) so the fix has a target.
- **3 — no change**, turn ids are #1303 / piece B.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; golangci-lint 0 issues in both modules; root
and agent suites; `-race -count=3` on agent, agent/transcript, apptranscript,
appprojector and the TUI packages. `internal/selfupdate` still fails locally on
macOS only (#1201).

## Round 8 fixes

Head `b150eb794`. `origin/main` unchanged since round 7's merge (`f7816dd8d`).

- **1 — MEASURED FIRST on main, no regression, limitation kept** (`fa1f6506c`).
  The record → restart → compact → restart walk, compiled against
  `origin/main` `f7816dd8d` (a `git archive` of main into the scratchpad, so
  no worktree of anyone's was touched) and against this branch, same probe,
  same numbers both times:
  - live history keeps the preserved suffix: **true**
  - markers written: **2** (checkpoint and summary)
  - resumed history keeps the preserved suffix: **false**, resumed history =
    **1 turn** (the summary alone)

  Main has no replay copies at all, so its anchor discards the restored turns
  exactly as this branch's does; what this branch added is copies for the
  turns it DOES hold a persisted form of. So the round-7 limitation test
  stays, now carrying the measurement, and #1200 remains the target.
- **2 — the failure counter** (`72364b766`). `FailureCounter.Observe` counted a
  copy's failing tool result as a second failure the moment a fold copied the
  round. RED: *failures after the fold copied the round = 2, want the same 1*.
  Copies no longer count and still teach the counter their call names, which
  is the split the item readers already make.
- **3 — fork counters** (`e6c90c45d`). The prefix copy is right — the child
  replays the run the way its parent would — but `TurnCount` and
  `AcceptedInputTurns` walked the same entries. RED: *child TurnCount = 2,
  want 1*. The counters skip copies; the child's transcript still carries
  them (asserted). The delegate fork's `DivergenceTurn` is an ENTRY ordinal
  over that same transcript, so it must count the copies, and does.
- **4 — a fold's later layers** (`476ce5907`). Only the first layer finds the
  interrupted turn open, so the rest recomputed "was a turn running" as false
  and left a completed turn as the last word on the wire. RED: *active frames
  after the second layer = 0, want 1*. The batch is remembered until a
  non-compaction event ends it; a fold between turns still claims nothing. The
  TUI multi-layer case now asserts `processing` and `Detail.State` after
  checkpoint+summarize.
- **5 — resumed turns carry no record roles** (`655bd1c9f`). `ContextReplay`,
  its merge-back mark and the fold id describe an entry in the file, not a
  turn in a history. RED: *a resumed turn still carries its record's roles*,
  and resume-of-resume reordered the conversation. One `resumedTurn` strips
  them on every path out of `ResumeHistory`.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; golangci-lint 0 issues in both modules; root
and agent suites; `-race -count=3` on agent, agent/transcript, apptranscript,
appprojector and the TUI packages. `internal/selfupdate` still fails locally on
macOS only (#1201).

## Round 9 fixes

Head `3867763fe`, with `origin/main` (`fabb4d875`, #1269) merged in — clean, no
conflicts.

- **1 — HIGH, one constructor for every append/write pair** (`f78cb56c4`). The
  tool-results append and the skill carrier's round minted their live and
  persisted forms with two `schema.NewTurn` calls, so the fold matched neither
  and wrote no copy: an anchored restart came back without the tool round. RED
  first, one case per producer, each recording its round WHILE a fold runs:
  *replay copies of the round recorded during the fold = 0, want 1* for both
  broken producers, with the other two green beside them.
  `newTurnPair` mints one turn in both forms and every producer builds there.
  **On identity**: I took both halves the brief offered. The constructor makes
  the match total by construction, and `schema.Turn.PairID` — minted once in
  `NewTurn`, `json:"-"`, never serialized — replaces the mint timestamp as the
  identity, because a timestamp is incidental (an injected clock can hand two
  turns the same instant) and says nothing about what it means for a turn to
  have no pair. PairID does: a turn read back from a file has none, which is
  exactly #1200. Kind+timestamp stays as the fallback for a turn built as a
  literal rather than through `NewTurn`, so no producer silently loses copies.
- **2 — M, the fold publishes the marker the transcript took** (`bfa518d4a`).
  A fold's last marker is the head of its published history; a reader anchors
  on the last marker in the FILE. With the summary's write failing and the
  checkpoint's landing, the two named different compactions. RED: *published
  head = SUMMARY, want the checkpoint the transcript took*. The publication now
  adopts the landed marker as its head, and the test compares the whole
  provider history live vs reload. With no marker at all the fold is
  unanchored, unchanged.
- **3 — declined**, ROUND_TIMINGS producer stays in piece B.
- Two round-trip equality tests (`e718ac0e5`) compared a live turn against the
  entry it was written as; `PairID` is `json:"-"`, so they now compare
  everything the transcript keeps and nothing it does not.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; golangci-lint 0 issues in both modules; root
and agent suites; `-race -count=3` on agent, agent/transcript, apptranscript,
appprojector and the TUI packages. `internal/selfupdate` still fails locally on
macOS only (#1201).

## Round 10 fixes

Head `bbadc7ea5`, with `origin/main` (`dd517a79d`) merged in — clean, no
conflicts.

- **1 — the mint moves to where a turn is created** (`94cef3341`). Assistant
  turns are built as a literal carrying the whole response, so they reached
  history and the pair log with no `PairID` and matched by (kind, timestamp) —
  which an injected or coarse clock makes ambiguous. `schema.MintTurn` gives a
  literal the same mint `NewTurn` gives everything else; the two literals that
  create a turn this process owns (the assistant turn, the repair synthetic)
  go through it, and the timestamp fallback is deleted. A form with no mint now
  matches nothing, which is the honest answer for a turn recovered from a file
  (#1200) and an invariant violation in `logPairPersistedLocked` anywhere else.
  RED: two same-kind turns at one frozen instant paired the wrong persisted
  forms (*kept 2 persisted forms, want the one whose turn is still in
  history*); with the mint, the right one and only it.
- **2 — the doctor** (`4e43d61f5`). The exclusion sits where the doctor turns
  lines into entries, so counts, health and session summaries all inherit it.
  RED: *read_file calls = 2, want 1*.
- **3 — the fork's cut** (`66a2e6d6b`). `completedDelegateContext` read copies
  as rounds of their own, so a crash between the copies and the markers left
  the child inheriting the unfinished call the cut exists to remove. RED:
  *kept 2 entries, want the cut before the unfinished round*. The copies still
  travel with the prefix.
- **4 — the adopted anchor's two other readers** (`4d5e8ade0`). The slice the
  publication returns (the in-flight request is built from it) and the history
  revision. RED separately: *the slice the in-flight request is built from
  starts with SUMMARY, want the checkpoint the transcript took*, and *a fold
  that snapshotted the uncorrected head published over the correction*.
- **5 — declined**, ROUND_TIMINGS stays in piece B.
- The new invariant found two un-minted producers through the fuzz seed corpus
  and cost one CI round (`bbadc7ea5`): a test fixture recording a literal turn,
  and the repair fuzzer's determinism check, which compares two runs of a
  repair that mints a fresh synthetic each time — the comparison now drops the
  mint with the timestamp. The invariant also moved out of
  `logPairPersistedLocked`: the fuzz build's `Hold` panics, and unwinding
  through two held mutexes turned a violation into a hang instead of a failure
  at the producer, so it runs before the pair takes any lock.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; golangci-lint 0 issues in both modules; root
and agent suites; `-race -count=3` on agent, agent/transcript, apptranscript,
appprojector and the TUI packages. `internal/selfupdate` still fails locally on
macOS only (#1201).

## Round 11 fixes

Head `d0fe3681f`, with `origin/main` (`498cde579`) merged in — clean, no
conflicts.

- **1 — the ordinary pair path and a retained write** (`c0fe840d0`).
  `appendTurnAfterTranscriptWriteLocked` returned on any error, so a turn whose
  whole line was in the file was in neither the history nor the pair log: the
  next fold had nothing to copy and its marker discarded the only record, while
  the session that kept running never had the turn. It goes through
  `entryOutcome` now — appended, logged, and the failure still reported. RED: a
  retained write, then a fold, then resume — *the turn is in the transcript and
  not in the history*.
  Two callers undid queue state on that error (returning a claimed steering or
  an accepted user turn to be delivered again); with a record in the file that
  delivers the same thing twice, so they undo only when nothing was recorded,
  and the environment append's own reconciliation now runs for that case alone.
- **1b — the environment append keeps its own rule** (`9adbe0343`). It is the
  one producer that reconciles a failed write itself, re-reading the transcript
  behind a durability barrier, and its tracker claims the MODEL SAW the turn —
  so a line in the file but not yet proven durable is the case it re-renders
  rather than claims. It takes a clean-write variant; the fuzz-tagged
  environment regressions are what caught the over-reach.
- **2 — the fold's steering is a pair** (`87e669de6`). Its persisted form never
  reached the pair log, so the NEXT fold's marker discarded it with no copy to
  carry it past. RED, two folds and a restart: *resumed […recorded turn 11,
  after the first fold 0…] live: […recorded turn 11, note handoff, after the
  first fold 0…]*. The rewrite set is now ordered by the published history
  rather than by the log, because the log cannot promise that order — the
  steering is appended to it after the copies are chosen, while the history
  holds that turn in the middle.
- **3 — measured, they disagreed, aligned** (`1172fd4e9`). A two-layer fold,
  live vs reload:
  - live: `turn_1 [userMessage]`, `turn_2 [context_compaction]`,
    `turn_3 [context_compaction]`, `turn_4 [compaction compaction]`
  - cold: the same first three, then `turn_4 [compaction]`, `turn_5 [compaction]`

  So the checkpoint and the summary shared a live turn that a reader coming
  back does not find, and every item after it moved. Both announcement kinds a
  fold publishes now go through one standalone-turn shape, and the
  interrupted-round state spans the whole batch (layers AND markers) so the
  last one does not leave a client believing the session went idle. Live and
  reload now agree turn for turn, ids included; the TUI has a case for the
  marker frames.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; the agent module under `-tags evenerfuzz`;
golangci-lint 0 issues in both modules; root and agent suites; `-race -count=3
-timeout 150m` across the agent module, apptranscript, appprojector and the TUI
packages. `internal/selfupdate` still fails locally on macOS only (#1201).

## Round 12 fixes

Head `3a7b04b1b`, with `origin/main` (`8475ae5fe`) merged in — clean.

- **1 — the callers read the entry too** (`4c4c8de66`). Round 11 taught the
  pair path that a retained entry is a record; five callers still read the
  error it propagates as "nothing happened", and each undid something the
  transcript had already committed to: the client-mutation recovery failed a
  restore whose recovery turn had landed (and the retry wrote it again), the
  tool round aborted delivery commits its own persisted turn names, the skill
  carrier left its obligation pending and re-delivered the body, the delegate
  steering aborted a claim whose message every reader can see, and a failed
  steering selection went back on the queue to be written and announced twice.
  Each branches on `entryIsRecorded` now — recorded means commit and advance,
  and report the failure anyway. One RED test per caller: *recovery reported
  failure for a record the transcript holds*, *pending deliveries = 1, want the
  round's commits completed*, and the three cheap ones (carrier, delegate
  steering, requeue).
- **2 — a fold's warning does not end its batch** (`3a7b04b1b`). The batch
  ended on any non-compaction event, and a fold emits a warning per failed
  record inside the same flush: the layer or marker after one found the batch
  gone and published its completed turn with nothing behind it. RED: *the layer
  after the warning ended the batch*. It ends where the interruption does — at
  the next real turn (`startTurn`, which the announcements re-arm around) or
  when the session ends — so everything one flush publishes is inside it, and a
  compaction between turns still claims nothing. The frames are unchanged, so
  no client case changes shape.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; the agent module under `-tags evenerfuzz`;
golangci-lint 0 issues in both modules; root and agent suites; `-race -count=3
-timeout 150m` across the agent module, apptranscript, appprojector and the TUI
packages. `internal/selfupdate` still fails locally on macOS only (#1201).

## Round 13 fixes

- **1 — the retained entry is settled in the encoding, not at callers**
  (`d791c39f5`). `appendTurnAfterTranscriptWrite` now commits the pair for a write
  whose line is in the file, reports that failure once where the entry was
  written, and returns nil. `transcript.ErrEntryRetained` no longer crosses the
  helper boundary: a non-nil error from a durable append means *the transcript
  has no record* everywhere, so every caller's plain error path is right by
  construction and none of them has to know the sentinel exists. The r11/r12
  `entryIsRecorded` branches at callers are folded back into that rule, and the
  fold's two decide-sites share `writeFoldRecordLocked`. Outside the helpers and
  package `transcript`, the only uses left are the two encoding sites
  themselves. One test per durable-append site pins the post-write effect, each
  RED against the previous encoding: the round's text end is emitted, the
  client-mutation recovery completes, the delivery commits complete with their
  round, the skill carrier and the delegate seed are recorded, the salvaged turn
  is latched, the reminder receipt is consumed, and neither the failed steering
  selection nor the delivered notifications are requeued.
- **2 — a consumed reminder survives a failed metadata save** (`d791c39f5`).
  `consumeSkillReloadReminders` dropped its handoffs before `saveMeta`, so a
  failed save lost them to the next autosave — the reminder then owed to nobody
  and delivered by no restart. The receipts are restored when the save fails,
  which is the shape `admitCompactedSkillReloads` already held. RED: *handoffs
  after the failed save = [], want the reminder's receipt still pending*.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; the agent module under `-tags evenerfuzz`;
golangci-lint 0 issues in both modules; root and agent suites; `-race -count=3
-timeout 150m` across the agent module, apptranscript, appprojector and the TUI
packages. `internal/selfupdate` still fails locally on macOS only (#1201).

## Round 14 fixes

All four are the same class the r13 encoding closed for the durable helpers,
still open at sites that decide for themselves. Each is RED against the
previous behaviour (`577cbf7ee`).

- **1 — a poisoned write that kept its line settles its input.** A complete-line
  write that stops the writer leaves a record every returning reader finds. It
  was read as a refused input, so the accepted turn went back and a restart
  processed the same input twice. Now the claim stands and nothing is taken
  back; the writer is still dead, so the turn it would start is refused with the
  record intact. RED: *accepted turns = 0, want 1: the input is in the
  transcript, so its claim stands*, plus the transcript holding it once.
- **2 — a retained replay copy still anchors the fold.** `tailComplete` asked
  `err != nil`, so a copy the transcript holds withheld the anchor and lost the
  compaction. It asks the shared recorded rule through
  `writeFoldRecordDurablyLocked`, the durable door's sibling of r13's
  `writeFoldRecordLocked`, and still reports the failure. RED: *the fold withheld
  its anchor over a replay copy the transcript holds*.
- **3 — the fold's steering joins the pair log whether or not the fold wrote
  it.** The steering is WRITTEN with the anchor and LOGGED either way: those
  turns are in the published history whatever this fold wrote, so an un-anchored
  fold left them for the NEXT marker to discard with no copy behind them. RED:
  *the un-anchored fold's steering appears 0 times in the resumed history*.
- **4 — a failed consumption save restores exactly what it removed.** `saveMeta`
  runs without `s.mu`, so a fold publishing in that window appends a handoff of
  its own and the pre-removal snapshot discarded it.
  `removeSkillCompactionHandoffsLocked` returns the receipts it took out and
  `restoreSkillCompactionHandoffsLocked` puts back only those, leaving a newer
  receipt for the same publication in place. RED: *want the handoff recorded
  while the save ran kept*.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; the agent module under `-tags evenerfuzz`;
golangci-lint 0 issues in both modules; root and agent suites; the Makefile
audits (main's merge touched `make/testing.mk`); `-race -count=3 -timeout 150m`
across the agent module, apptranscript, appprojector and the TUI packages.
`internal/selfupdate` still fails locally on macOS only (#1201).

## Round 15 fixes

All three in `ba30cf0a5`.

- **1 — one reminder per handoff.** A reminder whose turn was recorded and whose
  metadata consumption then failed is admitted; keeping the receipt is what lets
  the retry retry THAT. Preparation now recognizes the publication's reminder in
  the history (`skillReloadReminderRecordedLocked` — the pair wrote it durably
  before the live append, so history holding it means the transcript does) and
  appends no second turn saying the same thing. RED: *the retry staged 154 input
  tokens for a reminder it appended nothing for*.
- **2 — one compaction renderer.** The live item takes its text from the
  record's own `schema.ContextCompaction.Announcement()` and carries that same
  struct in its raw payload; the projector's private copy of the formatter is
  deleted. The json tags were already identical, so no wire shape changes and no
  TUI case is needed. The live/reload parity tests are the oracle and stay
  green.
- **3 — a true warning.** The merged-tail warning's "not anchored on disk"
  clause describes the FOLD, so it is emitted only when the fold really withheld
  its anchor; a copy that kept its whole line reports a plain write failure.
  `TestFoldPublication_ATailCopyRetainedByItsWriteStillAnchorsTheFold` now reads
  the warning. RED: *the retained copy's warning says the compaction was not
  anchored, and the marker is in the transcript*.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; the agent module under `-tags evenerfuzz`;
golangci-lint 0 issues in both modules; root and agent suites; `-race -count=3
-timeout 150m` across the agent module, apptranscript, appprojector and the TUI
packages. `internal/selfupdate` still fails locally on macOS only (#1201).

## Round 16 fixes

Commit `5590daa5f`.

- **1 (High, accepted — the finding was real).** Measured: `AppendDurable` whose
  sync fails and whose rollback also fails returned `ErrEntryRetained` and left
  the writer NOT poisoned, so the next append succeeded — the contract "a failed
  sync poisons the writer so nothing further is announced" did not hold. Now it
  does: that arm poisons the writer. Poison has two causes — a partial line is
  permanent (no fsync makes half a line whole); a retained whole line is cleared
  by a successful `EstablishDurability` barrier, because the full re-fsync is
  exactly what that record was missing. Proof of the poison:
  `agent/transcript/transcript.go` sync-failure/rollback-failed arm →
  `poisonLandedBytesLocked`; `TestDurableAppendStopsTheWriterWhenItCanNeitherSyncNorRollBack`,
  `TestDurabilityBarrierRestartsAWriterStoppedByARetainedLine`,
  `TestDurabilityBarrierLeavesAWriterStoppedByAPartialLineStopped`. The
  environment recovery raises the barrier, so it still works; a fold raises none,
  so its marker is refused and its anchor withheld.
  `TestRetainedWrite_ClientMutationRecovery…` raised no barrier, so its premise
  changed: the recovery records its input once and the writer then refuses the
  diagnostic (renamed `…RecordsItsInputAndStops`).
- **2 (Medium, confirmed by measurement, fixed).** Measured: every marker write
  failing → live history WAS replaced (12 → 7) while resume rebuilt all 12 plus
  two `CONTEXT_COMPACTION` records announcing a shrink the resumed history never
  had. Fix: the fold's `CONTEXT_COMPACTION` records are written after its markers
  and withheld with the anchor under the same `foldAnchored` rule as its steering,
  and `ResumeHistory` skips them by kind. No live rollback. RESIDUAL, deliberate
  (also in a code comment at the withhold site): when no marker lands the live
  session keeps the compacted history while a restart resumes the full pre-fold
  history — no turns lost, and the next fold that anchors re-converges the two.
  RED: `TestFoldPublication_AFoldThatAnchoredNothingLeavesNoCompactionRecords`.
- **3 (ROUND_TIMINGS).** Declined again, piece B, unchanged.
- **4 (Medium, fixed) — with a misattribution to note.** The finding named
  `compactionAnnouncement` (1601), but measured it already preserves the flag
  (it calls `openTurn("")`, which never touches it). The real degrade is
  `EventEnvironment`, which calls `openTurn(data.TurnID)` and consumes the stable
  id: the restored reservation dropped `reservedTurnIDIsStable`, so the next turn
  skipped the increment and live `turn_2` drifted from reload `turn_3`. Both sites
  now save/restore the flag. RED (environment case):
  `TestAppEventProjectorAnnouncementsKeepAPendingStableReservationStable`.

Gates: toolchain gofmt; host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` vet in both modules; the agent module under `-tags evenerfuzz`;
golangci-lint 0 issues in both modules; root and agent suites; `-race -count=3
-timeout 150m` across the agent module, apptranscript, appprojector, the TUI
packages and agent/transcript. `internal/selfupdate` still fails locally on
macOS only (#1201).


obra:
## Review round 1 disposition (b51e43184)

- **High, replay copies counted by the turn index but skipped by the paging reader** — accepted as the reviewer's cleanest form: `ProjectTurn` returns nothing for a replay copy, so index, paging and full projections agree; apptranscript-level test that pages a transcript containing a replay copy.
- **Medium, aggregate scanners double-count replay copies** — accepted.
- **Medium, no producer writes `TurnRoundTimings`** — by design for this piece: the writer moves to piece B of the split (an unnamed timing record becomes its own logical turn on cold replay, so the writer and the turn-naming land together). The schema comment will say so.
- **Medium, hook event announced before a queued pre-attach write is durable** — accepted: the event waits for the flush.
- **Medium, empty loop leaves `CompactionFoldID` unset in live history** — accepted: tag the in-memory turns.
- **Low, settlement helper and `projectedRecords` count** — both accepted.

obra:
## Review round 2 disposition (827a94914)

CI green. RoboRev (4 reviewers), all accepted:
- **Medium, held turns and their announcements kept in two parallel slices that can desynchronize** — accepted in the form that makes misalignment impossible: one queue of (turn, announcement) structs; the fork-delegate seed goes through it.
- **Medium, hook completions appended without durability before the announcement** — accepted: `AppendDurable` for hook completions, queued ones included.
- **Medium, a queued hook enters live history before its write succeeds** — accepted: history and the append log follow the successful write, same as the event.
- **Medium, replay copies still reach logical grouping and usage stamping** — accepted as one exclusion point before any consumer sees the entry, so full and bounded readers agree by construction.
- **Low, nil history in the tagging loop** — accepted.

obra:
## Review round 3 disposition (1cb66f0b6)

CI green. RoboRev (1 of 4 reviewers with substantive findings; two clean):
- **High, `persistedAppendLog` pruned before the replay copies and marker are anchored** — accepted pending the test the reviewer describes (copy write fails, a later fold succeeds, resume): pending persisted forms stay until the fold is durably anchored.
- **Medium, no producer writes `TurnRoundTimings`** — declined for the third time; by design for this piece, the schema comment says so, the producer lands with the turn naming in piece B.
- **Medium, an `AppendDurable` failure after the bytes landed drops the announcement** — accepted pending measurement of the writer's failure semantics: if the writer can tell "not written" from "written, not synced", the latter commits and announces; if it cannot, the writer learns to.
- **Low, doc comment now attached to the wrong function** — accepted.

obra:
## Review round 4 disposition (3ad307a2b)

CI green. RoboRev (4 reviewers), all accepted:
- **High, already-persisted turns in the preserved suffix fall before the new marker and are dropped on resume** — accepted pending the reload test: the persisted forms of the preserved suffix are replay-copied before the marker is written.
- **High, a marker-less fold clears the unanchored debt and prunes the append log** — accepted: a publication counts as anchored only when it produced and landed a marker; the debt survives marker-less folds.
- **Medium, the flush path in `attachTranscript` lacks the `ErrEntryRetained` handling the direct path got in round 3** — accepted as one function for both paths, so the rule cannot diverge again; tested with the retained-entry filesystem.
- **Medium, `CONTEXT_COMPACTION` records regroup on reload because live attaches them to the active turn and replay does not** — accepted in the consistent form for this piece: standalone in both projections, with turn ownership arriving uniformly in piece B (which already owns the timing record for the same reason). If the projected shape changes, the TUI case comes with it.

obra:
## Review round 5 disposition (e5f40d3db)

CI green. RoboRev (High ×1, Medium ×6), all accepted, three of them as one rule:
- **High, indexed grouped projection still projects replay copies; Medium, replay records feed `prevKind`; Medium, replay copies consume entry ordinals** — one rule: a replay copy never enters the index's logical structures (groups, kinds, ordinals, projection); only its physical scan extent is kept. That is where round 2's exclusion should have been drawn, and it closes the class across every reader.
- **Medium, live compaction turn ids minted vs cold ids derived from physical entry index** — accepted pending the multi-entry test; the fix, if it reproduces, is cold allocation by logical-turn order, since persisting ids is piece B's turn naming.
- **Medium, consecutive compaction layers reuse one live turn id** — accepted: every compaction event opens and closes its own standalone turn.
- **Medium, a marker retained after a failed sync counts as not landed** — accepted: retained is landed for effect dispatch, the same rule round 3 gave hook completions, so the two cannot diverge.
- **Medium, resumed history places injected steering ahead of the preserved suffix** — accepted; the round-4 test encoded the wrong order. Reassembly follows the order the fold held in memory, proven by live-versus-resume equality of the provider sequence rather than an order the test asserts on its own.

obra:
## Review round 6 disposition (ee4dcc892)

CI green. RoboRev (High ×1, Medium ×3, Low ×1):
- **High, the delegate fork strips replay copies before handing the transcript to the child** — accepted pending the reload-through-fork test; the copies are the only carrier of turns recorded during the parent's compaction, and the resume paths handle both the anchored and unanchored cases without the filter.
- **Medium, retained entries still suppress the live compaction and steering announcements** (three reviewers) — accepted, and the class closes: `entryIsRecorded` is applied through one write-then-announce path for every record kind, not at each loop separately.
- **Medium, live versus reload compaction turn ids** — declined here; that is #1303, filed for piece B after round 5's measurement.
- **Medium, no producer of `TurnRoundTimings`** — declined for the fourth time; piece B.
- **Low, the buffered append path drops the retained signal** — accepted: both doors report retention the same way.

obra:
## Review round 7 disposition (f78050454)

CI green. RoboRev (Medium ×3):
- **Medium, replay copies are chosen from the in-memory append log, which is empty after a restore or a fork inheritance** — declined here as the known limit round 4 named: turns with no pair-log entry need the durable identity #1200 is open for; before this piece a fold kept none of them past its marker, so this is the boundary of what piece A improves, not a regression. Recorded on #1200 with a restart, compact, restart test that pins the current behaviour as a limitation.
- **Medium, live versus reload compaction turn ids** — declined, third time; #1303, piece B.
- **Medium, resumed order places the fold's steering after the concurrent turns while the live history places it before them** — accepted as a measurement: round 5's reviewer said the opposite, so the live-versus-resume equality test grows the missing case (steering plus a turn recorded mid-fold), and whatever order the live history holds is the contract the resume path follows.

obra:
## Review round 8 disposition (b3fb8d96d)

CI green (after a #1308 rerun elsewhere; this head's run passed). RoboRev (High ×1, Medium ×4), all accepted:
- **High, a compaction after a restore anchors without copies for the restored turns, and the next restart loses them** — accepted pending a measurement against main: round 7 called this a pre-existing limit on the implementer's word, and this reviewer's framing says the anchor is what discards them. If main keeps those turns across restart, compact, restart, this branch regresses and fixes it: a restored turn's persisted form is the entry it was read from, so the pair log can be seeded at restore; failing that, an anchor that cannot replay a turn must not discard it.
- **Medium, the failure counter counts replay copies** — accepted; same exclusion as the other aggregates.
- **Medium, fork snapshots count replay copies as child turns** — accepted.
- **Medium, the projector recomputes `interrupted` per compaction layer, so a multi-layer compaction ends the live client idle while the daemon runs** — accepted, with the TUI test asserting processing state across layers.
- **Medium, `CompactionFoldID` left on restored turns breaks re-restore idempotence** — accepted.

obra:
## Review round 9 disposition (b150eb794)

CI green; two of four reviewers clean.
- **High, two producers mint the live and persisted forms with separate `NewTurn` calls, so the fold's identity match by timestamp never matches them and their tool-result rounds get no copy** — accepted, and the class closes: one function builds every (live, persisted) pair from the live turn, so no producer can mint a second identity, and a test drives each pair producer through a fold.
- **Medium, no producer of `TurnRoundTimings`** — declined for the fifth time; piece B.
- **Medium, an earlier marker lands and a later one fails, yet the fold is published in memory and a restart anchors on the earlier marker without the summary** — accepted: the fold is published only when its anchoring marker landed; otherwise the later layers are not published.

obra:
## Review round 10 disposition (3867763fe)

CI green. RoboRev (Medium ×5):
- **Doctor diagnostics count replay copies** — accepted; the same exclusion every other reader got, applied at the one point doctor reads entries.
- **No producer of `TurnRoundTimings`** — declined for the sixth time; piece B.
- **Assistant turns built as literals bypass the pair constructor, so identity falls back to timestamps** — accepted, and the class closes: `PairID` is minted where a turn is created, not where a pair is, so no producer can build a turn without one, and the timestamp fallback goes.
- **`completedDelegateContext` treats replay copies as rounds** — accepted.
- **Fallback-anchor adoption leaves the returned `published` slice stale and skips the revision bump** — accepted, both halves.

obra:
## Review round 11 disposition (bbadc7ea5)

CI green; reviewers 2 and 4 clean. Three Medium, all accepted:
- **The ordinary durable-pair path returns on any write error without asking whether the entry was retained, so a retained entry's turn leaves history and the pair log while staying readable on disk** — confirmed by reading `appendTurnAfterTranscriptWriteLocked`: it is the one write-then-decide path that does not go through `entryOutcome`. Accepted, and the class closes: every path that writes an entry and then decides what to keep consults the same outcome rule.
- **Fold-injected steering is written to the transcript but never logged as a persisted pair, so the next fold can discard it without a copy** — confirmed by reading `writeSteeringTurnRecordsLocked`: it writes and never calls `logPairPersistedLocked`. Accepted; the steering write is a pair like every other, with a two-fold test proving the copy.
- **Multi-layer folds: compaction events each get a live turn while consecutive checkpoint/summary events share one synthetic gap turn, and reload groups every marker on its own** — accepted subject to measurement (a two-layer fold with checkpoint and summary markers, live projection vs reload); if the measurement disagrees with the reviewer the round says so with the test.

obra:
## Review round 12 disposition (d0fe3681f)

CI green; reviewers 2 and 4 clean. Two Medium, both accepted:
- **`ErrEntryRetained` is a durable success at the pair path now, but five callers still read the propagated error as "nothing recorded" and abort claims, release delivery commits, requeue, or fail restore** — accepted, and this is the class round 11 was meant to close, finished at the callers: every caller of a write-then-decide path branches on whether the entry was recorded, with retained-write tests on client-mutation recovery and tool results with delivery commits.
- **A warning emitted mid-flush (a failed middle layer or marker) resets the compaction batch, so the later layer's completion goes out without the trailing active status and a client goes idle mid-round** — accepted; the batch ends on a real turn boundary, not on a warning inside the same fold flush, with a test where a middle layer fails and a later one lands and the final frame is the active status.

obra:
## Review round 13 disposition (272c9e9d4)

CI green; one reviewer clean. Two findings, both accepted, and the first changes approach:
- **Retained writes are still read as failures at roughly seven more sites (tool results after the round is recorded, the failure diagnostic in mutation recovery, the assistant append that never emits text-end, salvaged turns, job reminders, skill-reload reminders, delegate preseed, and more)** — accepted, and after three rounds of patching callers the encoding itself is the bug. The retained case is settled once, in the durable helpers: a retained entry is recorded, the helper reports it once where the entry was written, and returns success. `ErrEntryRetained` stops crossing the helper boundary, so an error from a durable helper means "not recorded" everywhere and no caller can misread it. The `entryIsRecorded` branches rounds 11 and 12 added at callers fold back into that single rule, with the retained-write tests kept and one added per site the reviewers listed.
- **`consumeSkillReloadReminders` drops handoffs from memory before `saveMeta`, so a failed save loses them to the next autosave** — accepted; handoffs stay until the metadata update is durably committed, with a failing-save test.

obra:
## Review round 14 disposition (506d9a250)

CI green; reviewer 4 clean. Three Medium and one Low, all accepted:
- **A complete-line write that retained its entry while poisoning the writer is treated as a refused input and rolled back, so a restart processes the input twice** — accepted; a retained entry settles the input consistently and the poisoned writer only blocks further processing.
- **`tailComplete` withholds the anchor on any write error, so a retained replay copy loses the compaction** — accepted; the shared recorded rule applies there too.
- **Handoff removal runs outside the metadata-save lock and the rollback replaces the whole slice, dropping a concurrent compaction handoff** — accepted; the lock spans mutation, save and rollback, and the rollback is guarded by the revision the sibling transaction already uses.
- **An un-anchored fold's injected steering stays in live history but is never logged, so no later fold can carry it** — accepted; the steering pair is logged whenever it is appended to history.

obra:
## Review round 15 disposition (3f7bb6a7c)

CI green; one reviewer clean. One Medium and two Low, all accepted:
- **After a reminder write succeeds and the metadata save fails, the restored handoff makes the retry append the reminder twice** — accepted; a durable reminder is recognised by its publication id and the retry consumes only the metadata, with a failed-save-then-retry test.
- **The live compaction announcement has its own formatter beside the schema's** — accepted; one source of truth.
- **The retained-copy warning says the compaction was not anchored when it was** — accepted; the suffix follows the anchor's actual state.

obra:
## Review round 16 disposition (c7a474492)

CI green; reviewers 3 and 4 clean after full-diff reviews of the same areas. Four findings:
- **High, `AppendDurable` returns retained when the sync and the rollback both fail, and callers treat that as durable** — declined as the contract the series settled and two reviewers confirmed: a retained entry is readable in the file, a failed sync poisons the writer so nothing further is announced, and the exposure on a crash before the kernel flushes is the one every write has until its sync returns. The round states the poison behaviour with a pointer to its test.
- **An un-anchored fold can replace live history so a restart reconstructs the pre-fold history** — accepted subject to measurement; if the live history is replaced when no marker landed, the fold rolls back its live publication (or anchors on whatever landed), with a test; if not, the round refutes with the test that pins it.
- **`TurnRoundTimings` has no producer** — declined for the seventh time; piece B.
- **`compactionAnnouncement` saves and restores the reserved turn id but not whether it is stable, so live numbering drifts from reload** — accepted, with a live-versus-reload numbering test.

obra:
## Round 16 disposition, correction on the High (c7a474492)

The implementer measured it and I was wrong: the sync-failure arm of `AppendDurable` does not poison the writer when the rollback also fails; the next append succeeds. So the High stands as real. Ruling: that arm poisons the writer. The entry stays "recorded" (it is readable, so a restart will see it, and history and the pair log must agree with the file), it is reported once where it was written, and nothing further can be appended, so no later announcement rests on an unsynced file. A fold whose replay copy is retained this way therefore cannot land its marker and withholds its anchor, which is the crash-consistent outcome (the copy without a marker is ignored by every reader). The round-15 test that had a retained copy anchoring the fold changes premise accordingly.

obra:
## /simplify pass, altitude finding (90a97ad33)

The transcript already gives every line a durable, monotonic identity: `Entry.Seq` is written by the writer and read back by `Open` on main today. The replay-copy design in this PR (`PairID`, `MintTurn`, the pair log, copies of retained turns, `ContextReplay` and its merged-tail flag, and exclusions of copies in fourteen readers) rests on the stated premise that records have no durable identity yet (#1200). They do.

The simpler mechanism: a fold writes one record naming the retained sequence numbers, and resume re-slices the original entries by `Seq`; the writer gains an all-or-nothing batch append so the session never reasons about a partial fold; the writer alone owns the retained-entry contract. That deletes the pair identity, the copies, every reader exclusion and the merged-tail reorder, and it closes #1200 rather than documenting it. `TurnRoundTimings` (no producer; declined seven times) would land with its producer in piece B.

This changes the PR's core rather than trimming it, so it is Jesse's call; asking him now. Rounds stop here until he answers.

obra:
## Review round 17 disposition (90a97ad33)

CI green; reviewer 2 clean. Two Medium, both real, both recorded here while rounds stay paused for Jesse's decision on the design (see the /simplify altitude finding above):
- **Post-fold strategy injections are appended to live history and only counted; their turns are never written or logged, so they vanish on reload** — accepted; under either design the fold's injected turns are part of what it persists.
- **The fold writes markers before layer records while the live flush announces layers before markers, so live and reload swap them** — accepted; under the current design one side matches the other with a test that reads a real multi-layer fold through the writer; under the sequence-number design the fold is one batch written and announced in one order, and the divergence cannot arise.

obra:
Closed on Jesse's ruling (2026-09-15): piece A is rebuilt on the transcript's existing durable per-line identity (`Entry.Seq`) as small PRs, per the /simplify altitude finding above. The branch stays for cherry-picks; its tests are the oracle for the new series:
- A1: the transcript writer owns durability (Seq returned by appends, all-or-nothing `AppendBatch`, the retained-entry contract and the two-cause poison inside the writer).
- A2: a fold writes one record of retained sequence numbers and resume re-slices the originals; pair identity, replay copies and every reader exclusion go; #1200 closes.
- A3: the independent fixes this PR found on the way (the projector's stable-reservation restore, skill-reload idempotence and handoff restore, one presentational-kind predicate).
`TurnRoundTimings` lands with its producer in piece B.

PR #1276, opened 2026-09-14
docs(multi-host): design + component specs for SSH-attached hub federation

obra:
Docs-only. Design and decomposition for letting any hub launch, adopt, and manage sessions on other hosts, sized for a series of small PRs.

**Shape:** a full `evener hub` per host; AppWire JSON-RPC carried over an SSH channel on stdin/stdout; the remote hub runs on demand, detaches, and is re-attached by bridging stdio to its existing loopback `/rpc`. Host-owned config administered through the Hub UI, live fan-out fleet view.

**Contents**
- `2026-09-14-multi-host-evener-design.md` — keystone: decisions, leverage, seven components, interfaces, risks, testing, PR sequence.
- `2026-09-14-multi-host-spikes-findings.md` — spikes that validated the transport end to end (initialize + thread/list over SSH stdio against a real host).
- Component specs 01–07: stream transport, attach bridge, host config, SSH connection manager, remote hub source, fleet view, remote admin.

**De-risking already done:** all 27–28 `appsource.Source` methods map to existing remote-hub handlers (verified against the catalog↔router cross-check), so no new thread/turn RPCs are needed; the only gap is host OS/arch, which no AppWire method exposes.

PR #1283, opened 2026-09-14
feat(hub): ssh channel manager (component 04a of multi-host)

obra:
Component 04a of the multi-host design (spec: docs/superpowers/specs/2026-09-14-multi-host-04-ssh-connection-manager.md). The 04a slice: channel + preflight + lifecycle. Deploy, version auto-match, and hub restart are 04b and deliberately absent here.

**Note on the base:** this branch contains PR #1277's StreamTransport and PR #1280's hostreg so it compiles standalone; both are expected to land on main before this is reviewed, after which the diff here is 04a-only.

New package `cmd/evener-hub/internal/sshconn`:
- **runner.go** — `Runner`/`Stdio` seams, production `execRunner` (no `CommandContext` for the long-lived bridge, matching spawnDaemon's reasoning), exact non-interactive ssh argv, dest composition, bounded stderr tail for diagnostics.
- **preflight.go** — `uname`→GOOS/GOARCH pinned to install.sh:21-45 (with a parity test), HOME/XDG root resolution reusing cmdutil/userdirs (no invented XDG values), and the `launch-check --protocol … --json` contract (protocol, version, launch_flags). Only linux/amd64 and darwin/arm64 are supported.
- **manager.go** — `Manager`/`Options`/`Channel`, idempotent `Ensure`, lifecycle states, link-down detection, bounded exponential backoff + jitter reconnect that spawns a fresh bridge and never starts a hub, and attach/detach events for component 05.

Named, terminal error classes for the spec's cases: `ErrSSHStart`, `ErrSSHAuth`, `ErrProtocolIncompatible`, `ErrUnsupportedHost`, `ErrLaunchContract`, `ErrPreflightDecode`. Protocol mismatch and a missing `api-log` launch flag are refused before any bridge process starts.

**Spec reconciliation:** the spec's Contract argv shows only BatchMode + ConnectTimeout, while §2 requires ServerAliveInterval/ServerAliveCountMax as the only keepalive (StreamTransport deliberately does not implement Pinger). The implementation sets all four, following §2.

Verified: gofmt clean, `go build ./...`, `go vet`, sshconn tests, and the full `go test ./cmd/evener-hub/` (214s) all pass. Also validated live against a real macOS host over SSH: preflight resolved darwin/arm64 with the ~/.config, ~/.local/state fallbacks, attach + Initialize succeeded, Ensure was idempotent, and thread/list returned threads over the channel.

Default `go test` spawns no ssh and needs no network: the suite injects a fake Runner and an in-process fake bridge that speaks real AppWire framing.


obra:
Second round of findings addressed (`54f0f4823`, `61f56865b`). Each fix carries a test that was run against a build with the fix reverted and failed with the predicted symptom.

- **High — `channelArgv` omits `ConfigPath`/`Addr`:** the argv now carries `--config <path> --addr <addr>`, in the order spec 04's corrected contract names. Verified end to end against the real command: `evener hub attach --stdio --config … --addr 127.0.0.1:1` parses both flags and dials the given address.
- **Medium — failed `Ensure` during a link drop disabled auto-reconnect:** the dropped channel now stays mapped until its replacement is actually attached, so its supervisor can still find it and run the reconnect loop; the regression test forces the Ensure-wins interleaving (reverted fix: `timed out waiting for event attached`).
- **Medium — `Ensure` closed the stale channel under the host lock:** the close moved outside it, matching what `supervise` already did.
- **Medium — unbounded reconnect context:** attempts are bounded by `attemptLimit()` (four preflight round trips plus the handshake), so a remote command that never returns can no longer hold the host's lock (reverted fix: `a hung reconnect attempt never timed out: the host lock is held for good`).
- **Medium — stderr copy racing stdout EOF:** the handshake-failure path now reaps the child (`Kill` + synchronous `Wait`) before classifying. `Wait` is what guarantees os/exec has finished copying stderr, so an auth refusal can no longer be misread as retryable.
- **Medium — FD/pipe leak on `client.Initialize` failure:** a shared `reapBridge` closes the stream as well as reaping the child.
- **Medium — `emit` calls `OnEvent` with the per-host lock held:** handled by **documenting the contract, not restructuring**. The lock is exactly what orders a `Detached` before the `Attached` that follows it for the same host; releasing it around the callback would let a listener act on those events out of order. `Options.OnEvent` now says it is non-reentrant and must not call `Ensure`.
- Low: `Ensure` re-checks the closed state after publishing; `execRunner.Start` refuses an already-canceled ctx and closes its pipes when a later step fails; `Preflight()` clones `LaunchFlags`; `Ensure` normalizes the name through the registry (a padded spelling used to find the host but create a second channel entry and a second lock); `FuzzParseEnvProbe` added and registered (1.5M execs, no invariant violations).

**Deliberately not changed: `evenerCommand` is not shell-quoted for the remote shell.** The spec documents the unquoted argv form, and its own test contract asserts that form; the rule for remote-shell quoting (or rejecting metacharacters at validation) is still open — it is one of the docs PR's Highs. Changing a documented contract here, ahead of the rule, would trade one inconsistency for another. I would rather settle quoting where the argv contract is defined and then apply it in one pass across the channel, launch-check, and health argv builders.

obra:
Round three addressed (`2c84fb2ec`, `4010fc719`). Each fix has a test that was run against a build with the fix reverted, and failed with the predicted symptom.

- **High 1 — events emitted without the host lock:** `Attached` is announced under the lock again, `supervise`'s `Disconnected` takes the lock, and `Reconnecting` moved into `reconnectOnce`, where the lock is already held. `TestAttachEventIsEmittedUnderTheHostLock` pins the contract directly — the callback `TryLock`s the host lock and must fail to get it. (Reverted: `1 of 1 attach events were emitted without the host lock`.)
- **High 2 — missing `Detached`:** retiring a dropped channel now emits exactly one `detachEvent` before the replacement's `Attached`. (Reverted, the observed event list was `[state state attached state state attached]`, the two-`Attached`-in-a-row leak this finding describes.)
- **Medium 3 — closed-channel race:** `Close` takes each host's lock while closing that host's channel, and `Ensure` re-checks both the manager and the channel before claiming success. A window before the caller's first use is inherent to `Close` being terminal; it is now the only one, and it no longer spans the blocking `stale.Close()`.
- **Medium 4 — remote arguments are now shell-quoted** (`quoteRemoteWord`, runner.go). This supersedes what I said in the previous comment about deferring it. Words that need no quoting keep the documented form, so the ordinary argv is unchanged; `rawCommandArgv` still passes its shell snippet through verbatim, as the one deliberate case. For the record: a quoted value is literal, so `~` survives only in words that did not need quoting.
- **Medium 5 — spurious `Kill` error:** `execStdio.Kill` ignores `os.ErrProcessDone` and `stdioReadWriter.Close` ignores `os.ErrClosed`. The test drives a real `sh -c 'exit 0'` child — Start, Wait, Kill. (Reverted: `Kill after the child exited = os: process already finished`.)
- **Medium 6 — stream separation:** a failed one-shot command reports `*RunError` with stdout and stderr apart, and authentication is classified only from ssh's own stderr. (Reverted: `a remote program's stdout was read as an ssh auth refusal`.)
- **Medium 7 — the initial preflight is bounded** by `attemptLimit()` when the caller's context has no deadline. (Reverted: `Ensure never gave up: the initial preflight is unbounded`.)
- **Low 8 — `Channel.Host` clones `Roots`**, as `Preflight` does for `LaunchFlags`.

On your note that `doc.go` references a spec file absent from this worktree: correct, and deliberate. The specs live on the `multi-host-specs` branch (PR #1276) and land with it; this slice is code-only, so that reference does not resolve until #1276 merges.

Verification: gofmt/build/vet clean, `sshconn` `-race -count=3` green, full `./cmd/evener-hub/...` suite green, module lint 0 findings, fuzz-gap 71/71, fuzz registry clean.

obra:
Round four addressed (`171c5ace9`, on top of a merge of `origin/main`). Five findings, no Highs; each fix again has a test that was run against a build with the fix reverted and failed with the predicted symptom.

- **Stale channel leaked when `Close` races `Ensure`:** the replaced channel is retired on every path out of `Ensure`, early returns included, through a deferred close that runs outside the host lock. (Reverted: `channel was never closed`.)
- **Dead channel, or `ErrManagerClosed` misreported for a link drop:** the decision is now one function, `channelUsable` — only `Manager.Close` makes the outcome terminal, and a channel that died under us is retryable. (Reverted: `lost channel reported as a closed manager`.)
- **`Permission denied` misclassified as auth:** the markers are now spellings ssh itself emits. I captured the real strings on this host: an ssh refusal reads `jesse@host: Permission denied (publickey,password,keyboard-interactive).`, so that parenthetical is the ssh-specific part, while a bare `Permission denied` is exactly what a host prints when it refuses to execute the binary we asked for. `Authentication failed.` is gone, as is `not accessible to the current user` — the real spelling there is `not accessible:` on a missing identity file, and that is a warning rather than a refusal. (Reverted: `isAuthFailure("Permission denied") = true`.)
- **Unbounded preflight output:** one-shot capture is capped at 1 MiB per stream and refused when the cap bites. (Reverted: `Run accepted 3145728 bytes of output`.)
- **In-flight initial attach not cancellable by `Close`:** the attempt context is now tied to the manager's lifetime with `context.AfterFunc` as well as bounded. (Reverted: `Close did not cancel the in-flight initial attach`.)

Two regressions my own suite caught while making these changes, recorded because they are the kind that would have shipped: returning combined streams on success (fixed — success returns stdout alone), and making a supervisor stand down for a closed-but-mapped channel, which would have left a host unsupervised after `Ensure` retires a dropped channel (fixed — ownership, not liveness, decides whether the supervisor still has work).

Verification on this branch: gofmt/build/vet clean, `sshconn` `-race -count=3` green, full `./cmd/evener-hub/...` suite green, module lint 0 findings, fuzz-gap 71/71, fuzz registry clean. The branch also carries a merge of `origin/main`, so it is current.

Process note: I briefly had these five fixes on the sibling 04b branch, because my worktree was still on it after integrating that branch. They were never pushed there; I moved them here, where the verdict asked for them, and rewound 04b to its pushed head, so 04b's PR is unchanged.

obra:
Round five addressed (`1750bfe9b`, plus `232c8e26d` refining a test). Seven findings — 1 High, 3 Medium, 3 Low.

- **High — a malformed frame left a dead channel installed.** `linkMonitor` sits at the byte level, so it cannot see an error the codec raises after a successful read. The client now reads through a transport wrapper that reports every receive error, and a test drives exactly that case: garbage on the stream after the handshake must mark the link lost and produce a `Detached`. With the wrapper reverted, the test times out waiting for it.
- **Medium — a PTY could be forced by ssh_config.** `-T` is now in the shared option prefix, so a `RequestTTY force` in a user's config cannot rewrite newlines in the framed stream or fold remote diagnostics into stdout. This is a deliberate change to the argv form the specs document; the specs PR needs a one-line update to match, and I am noting it here rather than leaving it implicit.
- **Medium — a terminal replacement failure left the host owned.** The dropped channel's supervisor could adopt the host and repeat a failure that cannot succeed. `Ensure` now releases the host on a terminal failure and announces it as `EventFailed` rather than only a state transition. With that reverted, no announcement has happened by the time `Ensure` returns.
- **Medium — `Attached` was announced before validating usability.** The freshly attached channel is now validated before it is published and announced, and when `Close` wins the race between announcing and returning, the `Attached` is paired with a `Detached` so a consumer can drop what it just installed. The Close-race test asserts that pairing.
- Low: `ErrManagerClosed` is terminal for the reconnect loop; the attempt cap now applies even when the caller brings a longer deadline (a long caller deadline used to hold that host's lock for its whole duration); and the initial path emits `EventFailed` for terminal failures.

One honesty note: the ownership-release fix shares its test with the announcement fix, and the assertion that discriminates cleanly is the announcement timing. The release itself is asserted through the end state — host unowned, terminal failure not retried — which the supervisor would also reach asynchronously, so I am not claiming it is independently pinned.

Verification: gofmt/build/vet clean, `sshconn` `-race -count=3` green, full `./cmd/evener-hub/...` suite green, module lint 0 findings, fuzz-gap 71/71, fuzz registry clean.

obra:
Round six addressed (`05650a11a`). Two Highs and one Medium, all in the replacement path.

- **High — a dropped replacement reaped ownership and orphaned the host.** `Ensure` published before validating, so a replacement that died in the handshake displaced the dropped predecessor and then cleared the map: the predecessor's supervisor stood down because its entry was gone, and no supervisor had been started for the replacement, leaving the host permanently unsupervised. It also leaked the predecessor's `Detached`. The replacement is now validated *before* publishing, a dropped predecessor stays mapped so its supervisor can recover, and with no predecessor the state is honestly disconnected. **Honest gap: this interleaving is not reproducible in a unit test** — it needs a frame error to land inside the handshake window — so this one is pinned by code order and by the ownership rule below, not by a failing-first test.
- **High — a terminal replacement failure never detached the retired channel.** A consumer had installed a source on that channel's `Attached` and would keep it forever. `Ensure` now emits `Detached` for the retired channel before the terminal `Failed`. Reverted, the observed sequence is `[state state attached state failed state]`: attached, then failed, with no detach.
- **Medium — the reconnect loop keyed ownership on liveness.** After a supervisor cleared the map and entered backoff, a replacement could be published and then drop; the first supervisor saw `liveChannel == nil` and attached a third channel, clobbering the second without ever detaching it. Ownership is now map presence: any mapped channel means another supervisor owns the host. Reverted, the test observes a second `Start` and a clobbered map entry.

Verification: gofmt/build/vet clean, `sshconn` `-race -count=3` green, full `./cmd/evener-hub/...` suite green (13 packages, exit 0, no failures), module lint 0 findings, fuzz-gap 71/71.

obra:
Round seven addressed (`6e102e0`). 4 Medium + 1 Low.

- **Medium — `Manager.Close` tore down channels without `EventDetached`** (`manager.go:371-401`): shutdown canceled the base context, so the supervisor returned without emitting the Detached its Attached had promised, and a lifecycle consumer kept the source. `Close` now pairs the events under the host lock, using a per-host record of the outstanding Attached so a channel whose Detached was already emitted (a replacement retired as Close raced it) is not announced twice. Pinned by `TestCloseDetachesAnnouncedChannels`.
- **Medium — a terminal failure from a concurrent `Ensure` could not stop the supervisor that had retired its own channel** (`manager.go:544-564`): after `clearChannel` the host had no mapping, so the supervisor waiting out a backoff kept looping and announced a second `EventFailed` for a host consumers were already told was finished. Each supervisor now gets its own cancelable context (derived from baseCtx) tracked per host; a terminal failure stops it. Pinned by `TestTerminalFailureStopsTheReconnectLoop` (was 2 `EventFailed`, now exactly 1).
- **Medium — authentication was classified from stderr text that ssh shares with the remote command** (`runner.go:101-120`, `preflight.go:269-279`): ssh forwards the remote command's stderr onto the same stream as its own diagnostics, so a remote program printing `Permission denied (` read as ssh refusing the key and ended the reconnect loop permanently. A failed `Run` now counts as an auth refusal only when ssh also reports its own exit status 255, which a failed remote command cannot forge; a failed `Start` has no status to read, so the marker alone still decides there. Pinned by `TestSSHRunFailureRequiresSSHExitStatus` and `TestRemoteCommandStderrPermissionDeniedIsNotAuthFailure`. Note: `attach`'s own `sink.tail()` classification (manager.go:470) is unchanged — it runs before the remote command can write, and that path exposes no exit status; flagged rather than touched.
- **Medium — the `Initialize` handshake timeout was dead code** (`manager.go:452-454`): both `Ensure` and `reconnectOnce` pass a context that already carries the attempt limit (default 70s), so `ctx.Deadline()` always reported ok and the `initTimeout` branch never ran — a hung handshake held the per-host lock for the full attempt limit. The handshake now always takes `context.WithTimeout(ctx, initTimeout)`, which keeps whichever deadline is earlier (default 30s). Pinned by `TestInitializeHandshakeIsBoundedByInitTimeout` (returned at the attempt limit before, now at `initTimeout`).
- **Low — `Preflight.StateRoot`/`ConfigRoot` were documented as the hub's token/lock location** (`preflight.go:103-120`, `175-184`): they derive the HOME/XDG default, while a remote `hub.toml` may override `hub_state_root` (and `attach` follows that override via its own config), so the reported root can be wrong. Fixed as documentation: the fields are the host's default evener roots, not a promise about the running hub's. No behavior change.

Verification: `gofmt -l` empty; `go build ./...`; `go vet ./cmd/evener-hub/...`; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=3`; `go test ./cmd/evener-hub/... -count=1` (13 packages); `dev module-lint` PASS (1 module, 0 findings).

obra:
Round eight addressed (`3ae3d15f3`). 4 Medium + 3 Low.

- **M1 — `Close` doesn't wait for supervisors / in-flight reconnect attempts.** Real. `Close` canceled `baseCtx` and returned without joining the supervise goroutines, so a supervisor parked in backoff (or unwinding an attempt) could emit `StateDisconnected` after `Close` returned. Added a `sync.WaitGroup` (`supervisorsWG`) that `startSupervise` registers each loop with; `Close` waits on it after tearing the channels down. `OnEvent` is synchronous and runs before a loop returns, so every event a supervisor emits is delivered before `Close` returns. Pinned by `TestCloseWaitsForSupervisors`.

- **M2 — `startSupervise` leaks the replaced supervisor's `WithCancel` child.** Real (unbounded: one child per reconnect until `Close`). Each loop now `defer cancel()`s its own context when it ends, releasing the child as soon as the replaced loop stops. Not pinned: Go exposes no way to observe a context's outstanding children, and the leak is memory-only with no observable functional edge — stated plainly rather than asserted by a test that cannot see it.

- **M3 — attach-path auth classification used stderr text alone.** Real. The bridge's sink carries ssh's diagnostics *and* the remote command's stderr, so a remote error containing `"Permission denied (…)"` could be read as terminal `ErrSSHAuth` and stop the reconnect loop for good. `reapBridge` now returns the child's wait error and the attach path requires `sshOwnFailure` (exit status 255), matching `isSSHAuthFailure` in the one-shot path. Pinned by `TestAttachAuthClassificationRequiresSSHExitStatus` (255 → `ErrSSHAuth`; remote exit 1 with the same marker → `ErrSSHStart`). `TestReconnectAuthFailureDoesNotRetry` was updated to model ssh's 255 — its fixture "started" a bridge with no exit status, which no longer classifies as auth; its assertions are unchanged.

- **M4 — retired channel's `Attached` can lose its `Detached` when `Close` races the publish/announce window.** Real. If `Close` set `closed` and snapshotted `chans` (holding the replacement) between `publishChannel` and the `isClosed` check, `Ensure` returned without detaching the replaced channel, and `Close`'s `claimAnnounced` saw the replacement, not it — leaving the old source installed. That post-publish `isClosed` branch now emits the replaced channel's `Detached`. Not pinned: the window is two statements with no injection point, so no deterministic test reaches it; the analogous window (`Close` after the announce) stays pinned by `TestEnsureRetiresReplacedChannelWhenCloseRaces`.

- **L5 — post-publish guards checked only `isClosed`, not `isLost`.** Real. Both post-publish paths now handle a link that dropped after the pre-publish validation: before the announce it is retired and reported as the retryable `ErrSSHStart` drop, and after the announce the caller gets the retryable drop instead of a dead handle (the just-started supervisor pairs the `Attached` with a `Detached`). Not pinned: same no-injection-point window as M4.

- **L6 — in-flight preflight cancellation wrapped as retryable `ErrSSHStart`.** Real. `Ensure` now maps a non-terminal error to `ErrManagerClosed` when `baseCtx` was canceled, so a `Close` that lands during preflight reads as a closed manager rather than a retryable transport failure (a concurrently-found terminal cause is kept). Pinned by `TestCloseCancelsInFlightInitialEnsure`, strengthened to require `ErrManagerClosed`.

- **L7 — spurious `EventFailed(ErrManagerClosed)` during shutdown.** Real. Both `Ensure` and `reconnectOnce` now skip `failedEvent` for `errors.Is(err, ErrManagerClosed)`, keeping the state/detach handling, matching the early `Ensure` path that emits nothing. Pinned by `TestCloseDuringHandshakeEmitsNoFailedEvent` and `TestCloseDuringReconnectHandshakeEmitsNoFailedEvent`.

Verification: `gofmt -l cmd/evener-hub/internal/sshconn/` empty; `go build ./...`; `go vet ./cmd/evener-hub/...`; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=3`; `go test ./cmd/evener-hub/... -count=1` (13 packages, all ok); `module-lint` PASS (0 findings).

obra:
Round nine addressed (`af7d496ff`). 5 Medium + 2 Low.

- **M1 — exit status 255 is ambiguous for auth classification** (`preflight.go`, `manager.go`). Confirmed: ssh(1) forwards a remote command's exit status unchanged, so 255 is also what a remote command that exits 255 produces, and ssh forwards the remote command's stderr onto the same stream as its diagnostics — both signals are forgeable by the remote side. No unambiguous ssh-side signal is available without restructuring the remote argv (a startup canary, or separating ssh's own log with `-E`), so I took the safer direction: **an authentication marker is terminal only when no remote command could have produced it — a failed ssh `Start` (no process, no remote output). A completed run/bridge whose child produced an exit status stays retryable (`ErrSSHStart`).** One rule, applied in `isSSHAuthFailure` and in `manager.attach` (the `isAuthFailure && sshOwnFailure` branch is removed; `sshOwnFailure` is now dead and deleted). Pinned by `TestSSHRunFailureRequiresSSHExitStatus` (255-with-marker case now asserts retryable; failed-`Start` case asserts terminal), `TestAttachAuthClassificationRequiresSSHExitStatus` (same two-sided shape), `TestReconnectPreflightAuthMarkerIsRetryable`, `TestReconnectAttachAuthMarkerIsRetryable`. `TestRemoteCommandStderrPermissionDeniedIsNotAuthFailure` and `TestEnsureStartFailureClassification` stay green. Trade-off stated plainly: a genuine BatchMode refusal now retries instead of being terminal — the safe side of the misclassification the finding names.

- **M2 — `Close` is not an event-quiescence barrier for in-flight `Ensure`** (`manager.go`). Fixed with accounting, not suppression alone: `Ensure` registers in `ensureWG` under the same mutex that sets `closed` (the ordering `startSupervise` uses), and `Close` waits on it after tearing channels down. Separately, once `baseCtx.Err() != nil` the `Ensure` terminal/error path emits nothing. Pinned by `TestCloseCancelsInFlightInitialEnsure` (strengthened: the event count cannot grow after `Close` returns, and no `EventFailed`) and the new `TestCloseSuppressesInFlightEnsureTerminalEmissions` (a terminal cause found concurrently emits neither `Failed` nor `Disconnected`).

- **M3 — post-publish `isLost` orphans the host**. Fixed by restoring the displaced channel (`restoreChannel`) exactly as the pre-publish check preserves it, so the supervisor blocked on the host lock still finds the host owned. Pinned by `TestLostPublishedReplacementPreservesSupervision`, which uses a new `Options.afterPublish` test seam to drop the just-published channel deterministically between publish and validation, then asserts the predecessor's supervisor re-attaches (Start #3).

- **M4 — `Close` racing publish leaks a closed channel / starts a supervisor after shutdown**. `startSupervise` now does its `closed` check and its `WaitGroup` registration under `m.mu` and reports whether it started; all three call sites reap + `clearChannel` when it refuses. `Close` also `clearChannel`s every entry it tears down, so no channel remains mapped. Pinned by `TestEnsureRetiresReplacedChannelWhenCloseRaces` (strengthened: after `Close` returns, `currentChannel` is nil) plus the `Close`-race tests above.

- **M5 — `reconnectOnce` emits `StateReconnecting` during shutdown**. Fixed by mapping a non-terminal error to `ErrManagerClosed` when `baseCtx` is canceled, before the `isTerminal` check, matching `Ensure`. Pinned by the new `TestCloseDuringReconnectPreflightEmitsNoExtraReconnecting`, which asserts exactly one `StateReconnecting` (the link-drop transition) and no `EventFailed`.

- **L1 — `stdioReadWriter.Close` ignores `io.ErrClosedPipe`**. Added the `io.ErrClosedPipe` ignore for both pipes; pinned by `TestStdioReadWriterCloseIgnoresAlreadyClosedPipes` (also asserts a real teardown error still surfaces). Note: the finding cited `io.ErrClosed`, which is not an identifier in package `io` (that is `os.ErrClosed`/`fs.ErrClosed`, already ignored), and `io.Pipe`'s own `Close` is idempotent and returns nil — so this is hardening, exercised with a stub that returns `io.ErrClosedPipe`.

- **L2 — `doc.go` cited a spec not in the tree**. Reworded to point at `cmd/evener-hub/config.go` (`HostConfig`, the `[[hosts]]` schema) and this package's `errors.go`.

The invariant this round enforces: **after `Close` returns, no lifecycle event is emitted, no supervisor is started, no channel remains mapped, and every in-flight `Ensure`/reconnect attempt has been accounted for.** In-flight `Ensure` callers are tracked in `ensureWG` (registered under the same lock that sets `closed`); supervisors are tracked in `supervisorsWG`, also registered under that lock; both are waited on before `Close` returns; shutdown-time `Ensure` emissions are suppressed; and every path that discards a just-published channel clears its map slot and never starts supervision.

Verification: `gofmt -l cmd/evener-hub/internal/sshconn` → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=3` → ok; `go test ./cmd/evener-hub/... -count=1` → all 13 packages ok; `MODULES="." dev module-lint` → PASS, 0 findings. Fuzz targets untouched, so `fuzz-gap-check`/`fuzz-registry-check` did not apply.

Honest gap: the M3 interleaving sits between two mutex operations and cannot be hit by racing the scheduler without a seam, so it is pinned via `Options.afterPublish` (a nil-checked hook in the existing "Test seams" group) rather than by timing.

obra:
Round ten addressed (`0bc4a7436`). 4 findings, no Highs.

- **M1 — `ErrSSHAuth` is unreachable in production, so unauthenticable hosts retry forever** (`errors.go`, `manager.go`, `preflight.go`). Confirmed: `execRunner.Run` wraps every completed ssh failure in `*RunError`, which `isSSHAuthFailure` deliberately rejects, and the attach-`Start` path has no marker to match because a failed `Start` never wrote one — ssh never spawned. So no production path can produce the sentinel. **Chose option (b), the honest contract.** Option (a) as suggested ("classify where ssh's own exit code is known and the channel produced no frame") is still forgeable: ssh forwards the remote command's exit status unchanged, including 255, and a remote command can exit 255 before producing a frame. Threading an "ssh-only stderr" flag out of the runner means restructuring every invocation to separate ssh's own log (`-E`) and reworking the `Runner` seam — a diagnostic/supervision redesign, not a bounded round-ten change. `ErrSSHAuth`'s doc now states plainly that it is not terminal and that the manager classifies no failure as it, and `isTerminal` no longer lists it; the sentinel is retained (not deleted) so component 05 can still match it. Relation to round nine's M1: round nine's decision that an ambiguous failure stays retryable was correct and is kept; what round nine got wrong was the surrounding claim — it still documented and listed `ErrSSHAuth` as terminal and called a failed `Start` "the one unambiguous case". Unambiguous it is, but unreachable, so the contract no longer advertises it. Pinned by `TestIsTerminalClassifiesManagerClosed` (extended: `isTerminal(ErrSSHAuth)` must be false), a non-terminal assertion added to `TestEnsureStartFailureClassification`, and the new `TestReconnectFailedStartAuthMarkerIsRetryable` (an `ErrSSHAuth`-classified reconnect must keep retrying under backoff with no `EventFailed`). All three fail on a build with the production fix reverted. The trade-off is unchanged and now stated honestly: a genuine BatchMode refusal retries under backoff instead of ending supervision.

- **M2 — `Ensure` acquires the per-host mutex without honoring `ctx`** (`manager.go:260`). Confirmed. The host lock stays a plain `sync.Mutex` — `Close` and the reconnect supervisors take it without a context and must not be made to fail — so `Ensure` now goes through a new `lockHostCtx`: a `TryLock` fast path, otherwise a goroutine parked on `Lock` that hands the lock straight back if the caller gives up (the gate is never left held by a caller that returned), plus an explicit `ctx.Err()` check after acquisition. Pinned by the new `TestEnsureHonorsContextWhileWaitingForHostLock` (a caller parked behind a held gate returns `context.Canceled` instead of blocking; fails at the 5 s timeout on the unfixed build) and `TestEnsureCanceledContextDoesNotReturnALiveChannel` (an already-canceled `Ensure` no longer hands back the attached channel; fails on the unfixed build).

- **M3 — a second concurrent `Close` is not an event-quiescence barrier** (`manager.go:467-478`). Confirmed. Added `closeDone chan struct{}` and `closeErr`: the first caller runs the teardown, records its error, and only then closes `closeDone`; every later caller blocks on it and returns the same error. Round nine's assertion that `Close` joins `ensureWG`/`supervisorsWG` is untouched — this only extends it to concurrent callers. Pinned by the new `TestSecondCloseWaitsForTheFirst`: a second `Close`, called after the first has committed, must not return while a supervisor is still parked, and after both return no further lifecycle event may be emitted. It fails immediately on the unfixed build ("a second Close returned while the first was still tearing down").

- **M4 — reconnects validate `nch` only before publishing, then announce it** (`manager.go:766-782`). Confirmed. `reconnectOnce` now reruns the post-publish guard `Ensure` already had: after `publishChannel` (and the `afterPublish` seam, now wired on the reconnect path too) it rechecks `nch.isLost() || nch.isClosed()` and, when the link has already gone, clears the map slot, reaps the channel, emits `StateReconnecting`, and takes another turn through the backoff. That supervisor still owns the host, so clearing the slot orphans nothing. Pinned by the new `TestReconnectLostPublishedChannelIsNotAnnounced`, which drives the drop through the seam and asserts no `Attached` is emitted for a channel that is already lost. Honest gap: the seam call is itself part of the fix, so the test cannot fail on the pristine branch; I verified it by reverting only the recheck with the seam kept — it fails with "1 already-lost channels were announced as attached".

Verification: `gofmt -l cmd/evener-hub/internal/sshconn` → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=3` → ok; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → all 13 packages ok; `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS, 0 findings. Each M1–M3 pin and the M4 recheck was separately confirmed to fail against the corresponding reverted production code.

obra:
## ClientIfAttached seam (specs round seven): pushed `bf4208f55`

### What changed (`cmd/evener-hub/internal/sshconn`, no files outside this PR's set)

- **`Manager.ClientIfAttached(name string) (*appwire.Client, bool)`** (`manager.go`) — the non-dialing, attached-only client lookup. It returns the installed channel's initialized client only while that channel is *present, not closed, and not lost*, and `(nil, false)` for a dormant host, an unknown host, a dropped channel, or after `Close`. It never spawns ssh, preflights, deploys, or attaches, so the notification broker's reconnect rebind (component 05) and component 06's 30s snapshot can neither eagerly attach a dormant host nor re-dial one that dropped between a check and the call. Names are normalized through the registry exactly as `Ensure` does.
- Spec lines followed: `04-ssh-connection-manager.md:108-118` (Go surface text) and `:462-472` ("A consumer that must not dial uses `ClientIfAttached`, not `Ensure`"); the name and signature match the spec verbatim. It reads the same installed-channel predicate `Ensure` uses, so the client and the liveness signal cannot disagree.
- One deliberate deviation from the spec's "within one per-host lock" phrasing: it does **not** take the per-host gate. `EventAttached` is delivered *under* that lock (`Options.OnEvent`), and the spec says the rebind reads this lookup from that callback, so a re-entrant acquisition would self-deadlock (the same non-reentrant hazard `Options.OnEvent` documents). The installed channel is read under the manager mutex that publishes and clears it.

### Round-ten roborev findings, all fixed in this round (3 medium + 3 low)

- **M1** `Ensure` never rechecked the caller's context after attachment → it now rechecks `ctx.Err()` before returning a channel; a caller canceled during `EventAttached`/finalization gets `context.Canceled` and no channel, and the manager-owned channel is **not** torn down (its supervisor keeps it).
- **M2** Cancellation during preflight/attach was wrapped as `ErrSSHStart` with a spurious `StateDisconnected` → it now returns the caller's own context error and emits no lifecycle event; a concurrent *terminal* cause is still kept.
- **M3** `Send`/write failures left `isLost` false → `linkMonitor.Write` now signals the link-down edge on a failed or partial write, so `liveChannel` can no longer hand out a channel whose writes fail. (Signalled at the write, not on every `Send` error: a caller cancellation or a local frame/format error is not link loss.)
- **L1** `parseLaunchCheck` accepted an empty `version` → empty/whitespace-only is now `ErrPreflightDecode`, not a version that merely differs (which would drive a deploy against a never-established identity).
- **L2** `Close` racing between the post-publish liveness checks returned a retryable drop → the isLost restoration path rechecks manager-closed and reports `ErrManagerClosed` (`lostAfterPublish` helper).
- **L3** `diagSink.Write` released its lock before forwarding, and every attach built its own sink over the same `Options.Stderr` → the manager now owns one serialized writer shared by every attach, so a caller-supplied sink need not be concurrency-safe.

### Tests that pin it

`TestClientIfAttachedReturnsOnlyALiveChannel` (seam: dormant/unknown/dropped/closed → false, attached → the exact `ch.Client()`, zero runs/starts), `TestEnsureCanceledDuringAttachReturnsContextError` (M1), `TestCallerCancellationReturnsItsOwnErrorWithoutStateEvent` (M2), `TestFailedSendMarksTheLinkLost` (M3), `TestParseLaunchCheck` extended with the empty/whitespace contract (L1), `TestLostAfterPublishClassifiesAManagerClosedByTheRace` (L2), `TestAttachDiagnosticsShareOneSerializedWriter` (L3). All are deterministic and offline (fakeRunner/fakeBridge); each was confirmed failing-first against the reverted fix, and the seam test does not compile without the method.

### Verification

`gofmt -l` over touched dirs → empty · `go build ./...` → 0 · `go vet ./cmd/evener-hub/...` → 0 · `go test ./cmd/evener-hub/internal/sshconn/ -race -count=3` → `ok 3.568s` · `go test ./cmd/evener-hub/... -count=1` → 13 packages `ok` (147s) · `GOLANGCI_LINT_CACHE=... MODULES=. dev module-lint` → `PASS lint (1 modules, 93s)`, 0 findings.

### Pre-existing flake, not from this change

`TestCloseDetachesAnnouncedChannels` failed once in an earlier `-race` run here. It reproduces on the untouched base `0bc4a7436` (`git stash` of this change: 3 failures in 50 `-race` runs) and its event order there is `[state/reconnecting, detached/reconnecting, state/disconnected]` — a supervisor's drop path pairing its own `EventDetached` and then emitting `StateDisconnected` from its canceled backoff sleep while `Close` is still waiting on `supervisorsWG`. No lifecycle event follows `Close`'s return, so the contract holds; the test's extra "last buffered event is a Detached" assertion is what is timing-dependent. Left untouched as out of this change's scope.

obra:
Round twelve addressed (`6aaad4125`).

- **`lostAfterPublish` leaked a predecessor `Attached` during a `Close` race** — `lostAfterPublish` now emits the predecessor's `Detached` itself when `restoreChannel` reports the manager closed: Close's snapshot holds the replacement this call published (or nothing), never the predecessor, and the predecessor's supervisor stands down on the canceled context without emitting. `restoreChannel`'s previously-discarded return is now the classification too, so that path reports the terminal `ErrManagerClosed` the doc comment promised instead of the retryable drop. Pinned by `TestLostAfterPublishDetachesThePredecessorOnClose` (0 Detached before the fix).
- **`ClientIfAttached` ignored `m.closed` during `Close`** — the closed check and the channel lookup now happen in one `m.mu` acquisition (`installedLiveChannel`), so the lookup returns `(nil, false)` from the moment shutdown begins and a `Detached` callback cannot rebind to a channel Close is tearing down. Pinned by `TestClientIfAttachedRefusesDuringClose` (served a client before the fix).
- **Overflow-triggered transport close didn't invoke `markLost`** — `lossWatchingTransport.Close` now calls `onErr` before closing the wrapped transport, so the AppWire client's `ErrNotificationOverflow` teardown marks the link lost and the supervisor reconnects. Pinned by `TestNotificationOverflowMarksTheLinkLost` (no Detached before the fix).

`main` merge: `git merge --no-ff origin/main`, 7 commits, no conflicts (the incoming commits touch `appwire/types.go`, `hubcore`, app-auth, and the frontend; nothing under `cmd/evener-hub/internal/sshconn`).

Verification: `gofmt -l cmd/evener-hub/internal/sshconn/` → empty; `go build ./...` → pass; `go vet ./cmd/evener-hub/...` → pass; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=3` → ok; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok; `GOLANGCI_LINT_CACHE=... MODULES="." dev module-lint` → PASS (1 module).

obra:
Round thirteen addressed (`0d7f26028`).

- **Supervisor generation leak on replacement attach (`manager.go:862`)** — `startSupervise` overwrote `m.supervisors[host]`, so a predecessor parked in backoff (up to `BackoffMax`, default 30s) lost its cancel func; when the replacement later failed terminally, `stopSupervisor` cancelled only the newest loop, and the older loop woke, re-ran `reconnectOnce`, and emitted a second `EventFailed` (or an `Attached` after that `Failed`) for a host already announced finished. Changed `supervisors` from `map[string]context.CancelFunc` to `map[string]map[*supervisorLoop]struct{}`: every live loop registers its own handle, deregisters on exit (keeping the set bounded across reconnects), and `stopSupervisor` cancels *all* of the host's loops. Pinned by `TestReplacementAttachDoesNotLeakTheOlderSupervisor` (was red: `EventFailed announcements = 2, want exactly 1`; now green). Overlap scenario it constructs: `Ensure` alpha, drop the first channel so its supervisor clears the slot and parks in a blocking backoff sleep, `Ensure` a replacement while it is parked (registering over it), then drop the replacement and let *its* supervisor find a terminal protocol mismatch on reconnect — exactly one `EventFailed`; the released predecessor must stand down rather than re-attach and emit a second.
- **Terminal error during a `Close` race still emits `EventFailed` (`manager.go:959-969`)** — only retryable errors were mapped to `ErrManagerClosed`, so a terminal fault found concurrently with shutdown (protocol mismatch, unparseable preflight) satisfied `isTerminal(err)` and still announced `EventFailed` during normal teardown. Now the announcement is suppressed whenever `m.baseCtx.Err() != nil`, mirroring Ensure's own shutdown path, while the terminal error and state transition are preserved. Pinned by `TestCloseRaceWithTerminalReconnectEmitsNoFailedEvent` (was red: `shutdown surfaced a spurious EventFailed: {... Err:... protocol "evener-appwire-v4", want "evener-appwire-v5"}`; now green), which cancels the base context mid-reconnect-preflight via `Close` and releases a terminal protocol-mismatch response.

The `ClientIfAttached` seam and the round-ten/twelve dispositions are untouched; no existing assertion was weakened.

Verification: `gofmt -l cmd/evener-hub/internal/sshconn/` empty; `go build ./...` ok; `go vet ./cmd/evener-hub/...` clean; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=3` ok (8.042s); `go test ./cmd/evener-hub/... -count=1 -timeout 30m` all packages ok; module-lint `PASS (1 modules, 95s)`, 0 findings.

obra:
Round fourteen addressed (`13bcc3d12`).

- **Medium — `cmd/evener-hub/internal/sshconn/manager_test.go` test flakiness.** Several concurrency tests arranged goroutine ordering with fixed `time.Sleep` calls, so their interleavings depended on the scheduler and machine load. Every *ordering* sleep is replaced with an explicit handoff: new test seams `beforeHostGate` / `beforeSuperviseGate` park a contender at the host gate (so Ensure-vs-supervisor races are exact instead of sleep-biased), and `superviseExited` makes "the supervisor stood down" an observation ordered on its own exit rather than a settle sleep. Interleavings now made deterministic: (1) a concurrent Ensure retires a dropped-but-mapped channel ahead of its supervisor, emitting the stale `Detached` before the replacement `Attached` (`TestEnsureAnnouncesDetachForRetiredChannel`, `TestLostPublishedReplacementPreservesSupervision`, `TestEnsureTerminalFailureReleasesHostOwnership`); (2) the dropped channel's supervisor is parked before the gate while Ensure's own attach fails and the host still recovers through supervision, with a second subtest parking Ensure so the supervisor consumes the injected failure (`TestFailedEnsureDuringLinkDropStillRecovers`, both orders); (3) a retired supervisor wakes from its backoff and stands down rather than clobbering the fresh channel (`TestReconnectSleepDoesNotHoldHostLock`); (4) a terminal failure stops a supervisor parked in backoff, including a replaced older supervisor (`TestTerminalFailureStopsTheReconnectLoop`, `TestReplacementAttachDoesNotLeakTheOlderSupervisor`); (5) a caller reaches a held host gate before its context is canceled (`TestEnsureHonorsContextWhileWaitingForHostLock`); (6) Close-in-flight event quiescence is ordered on the joined Ensure/supervisor exit instead of a quiet-period sleep. Every existing assertion is intact. Each replaced interleaving was mutation-checked to confirm it still catches the regression it was written for — dropping `reconnectOnce`'s mapped-channel guard, removing attach's stale detach, no-op'ing the lost-slot restore, and removing the terminal stopSupervisor each produced the expected failure. The remaining `time.Sleep` calls are 1 ms polls inside bounded condition waits (`waitUntil`/`waitClosed`), not ordering.
- **Low — `cmd/evener-hub/internal/sshconn/manager.go:831` initial reconnect delay ignored `BackoffMax`.** The supervisor seeded its delay from `BackoffBase` alone, so a configured maximum below the base let the first retry exceed it. It is now `min(BackoffBase, BackoffMax)`; `nextBackoff` already capped the later ones. Pinned by `TestInitialReconnectDelayIsClampedToBackoffMax`: with `BackoffBase` 100 ms and `BackoffMax` 10 ms the first recorded backoff must be 10 ms, and the test fails against the unclamped 100 ms.
- Merged `origin/main` (`4b291876a`) with `--no-ff`; no conflicts.

Verification: `gofmt -l cmd/evener-hub/internal/sshconn/` empty; `go build ./...` ok; `go vet ./cmd/evener-hub/...` ok; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=3` ok; `go test ./cmd/evener-hub/internal/sshconn/ -count=10` ok; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` ok (all packages); module-lint PASS, 0 findings.

obra:
Round fifteen addressed (`ddfe3b7bf`).

- **Medium — stale supervisor emitted `StateDisconnected` after a replacement attached** (`manager.go`). A supervisor whose context was canceled on its way out of the backoff loop (`supervise`, ~855) or at `reconnectOnce` entry (~958) still emitted `StateDisconnected` unconditionally. When a terminal `Ensure` stopped an older supervisor via `stopSupervisor` and a subsequent `Ensure` attached a replacement before the stale loop reached the host lock, that `Disconnected` landed after the replacement's `Attached`, leaving lifecycle state inconsistent (and it could also trail `Close`'s terminal `Detached`). Both sites now stand down silently: a canceled loop no longer owns the host, so the canceller — the terminal `Ensure`/`reconnect` that already announced the host's `Disconnected`, or `Close`, whose terminal event is its `Detached` — owns the announcement. This matches `Close`'s documented "the supervisor returns on the canceled base context without emitting one" pairing behavior. Pinned by the new `TestSupersededSupervisorDoesNotEmitDisconnectedAfterReplacement` (subtests for both cancellation sites) and `TestCloseCanceledBackoffSupervisorEmitsNoDisconnected`; both fail on the pre-fix tree. The interleaving constructed: park the dropped channel's supervisor in its backoff sleep → drive a terminal `Ensure` (cancels the loop through `stopSupervisor` and emits the host's `Disconnected`) → attach a replacement `Ensure` (`Attached`) → release the stale loop so it contends for the host lock only after the replacement owns the host.
- **Low — stale comment on auth-failure terminality** (`preflight.go:248`). The `authFailureMarkers` comment claimed markers are terminal under `BatchMode`, but `isTerminal` deliberately excludes `ErrSSHAuth`. Rewrote it to state the real, deliberate behavior: an auth-shaped match stays retryable because ssh forwards the remote command's stderr/exit status on the same stream, so the refusal cannot be attributed to ssh.
- **Low — unitless count routed through duration arithmetic** (`runner.go:241`). `ServerAliveCountMax` was rendered via `sshSeconds(time.Duration(count)*time.Second)`, which overflows for a large configured count (a count of `1<<40` rendered as `ServerAliveCountMax=1`). Now rendered directly with `strconv.Itoa(o.serverAliveCountMax())`. Pinned by the new `TestServerAliveCountMaxRendersAsUnitlessCount`, which fails on the pre-fix tree.

Verification: `gofmt -l cmd/evener-hub/internal/sshconn/` empty; `go build ./...` ok; `go vet ./cmd/evener-hub/...` ok; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=3` ok; `-race -count=10` ok; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` ok; `GOLANGCI_LINT_CACHE=$(scripts/lib/golangci-lint-cache.sh) MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → 0 findings.

Note: the gates also surfaced `TestEnsureRetiresReplacedChannelWhenCloseRaces` flaking (~2/50 under `-race`) on pristine `13bcc3d` as well; the same cancellation-induced `Disconnected` was the cause, and this fix makes it stable (100/100 under `-race`).

obra:
Round sixteen addressed (`2658125ac`).
- `cmd/evener-hub/internal/sshconn/manager.go` (Low): the child-wait goroutine closed `ch.stopped` before calling `ch.markLost()`, so a concurrent `ClientIfAttached`/`Ensure` could briefly observe the already-exited SSH process as live and be handed a dead channel. Fixed by retiring liveness first — `ch.markLost()` now runs before `close(ch.stopped)` — and a new `afterChildExit` test seam exposes the publication point.
- `TestExitedChildIsNeverReportedLive` pins the ordering: it parks the child-wait goroutine exactly at the exit-edge publication point and asserts `ClientIfAttached` reports no live channel. It fails deterministically on the pre-fix ordering (verified by temporarily reverting the order) and passes after.
- No existing disposition or assertion changed; rounds thirteen/fifteen invariants are untouched.
Verification: `gofmt -l` over `cmd/evener-hub/` empty; `go build ./...` ok; `go vet ./cmd/evener-hub/...` ok; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=3` ok (7.2s) and `-count=10` ok (21.7s); `go test ./cmd/evener-hub/... -count=1 -timeout 30m` ok; module-lint PASS (0 findings, 1 module).

obra:
Landing (coordinator): RoboRev round 17 is "otherwise clean, well-structured code" with two Lows (unchecked duration arithmetic at manager.go:177-181,1063-1069; a stale `preflight` doc comment at preflight.go:161-162). All 15 CI checks are green. I am merging with --admin, which bypasses exactly two protections and nothing else: the 1-approval rule, and the up-to-date-with-main requirement (main is 1 commit ahead; this PR's diff does not overlap that commit). The two Lows are being fixed in an immediate follow-up PR against main.

Issue #1205, opened 2026-09-13
gate: port the bounded package-list runner (timeout, retry, process-group stop, fail-closed cleanup) from bash to Go


obra:
Follow-up to #1145 (part of #1116). `scripts/gate/run-module-tests.sh` now carries ~500 lines of process lifecycle logic for the bounded `go list` enumeration: per-attempt timeout, retry, process-group isolation via a perl `setpgrp` wrapper, pgid record ordering, SIGTERM→SIGKILL escalation with a zombie-aware liveness probe, fail-closed handling when `ps` cannot run, and record retention rules for the EXIT and signal cleanups. Fifteen review rounds each found a new real edge in that code, and `docs/developing-evener/testing.md:145-155` is explicit that a tool which can only be tested by faking `go` has outgrown shell — the maintainer has ruled there will be no fake-toolchain shell test for it.

Port the bounded runner to Go (under `cmd/evener-dev` or the gate's own package, wherever the existing dev tooling lives), keeping the observable contract #1145 established: retry only on a timeout that actually stopped a live attempt, a nonzero exit is the attempt's own verdict, an unreadable process listing fails closed and keeps the record, a stopped attempt's survivors are named, the promoted list is the one the tests ran against, and the diagnostics name the log, partial lists and effective caches. The behavioral tests that were reverted from #1145 (`4920888`) describe the four outcomes to pin; in Go they are ordinary tests with an in-process stand-in, not a faked toolchain on PATH. Related: #1163 (existing fake-go test to retire), #1164 (shards residual).

obra:
Ruling refined 2026-09-13 (Jesse): a real-process shell self-check for `scripts/lib/process-group-lib.sh` is accepted — real `sleep` groups through the library's own wrapper, assertions on the documented state table (stopped / kept / refused), run from a make target. The original ruling stands against a Go test that stubs the `go`/`curl` toolchain. Landing on #1145.

PR #1288, opened 2026-09-14
ci: bound and retry the golangci-lint download

obra:
One HTTP 500 from the release CDN failed `lint-golangci`, and with it the aggregate `static` job (run 34644540860), on a head whose code was clean. `make/repo.mk` piped the upstream installer straight into `sh`: no retry at either end, and no bound on either download.

Split out of #1265, which bundled this with the Go helper and the gate wiring.

## The two downloads

- **Ours**, the fetch of `install.sh`: curl's own `--retry 3 --retry-delay 2 --retry-all-errors` (so an HTTP status counts as a failure worth another try), `--connect-timeout 15` — an edge that accepts nothing is the failure this script is about, and 15s is long past a healthy handshake — and `--max-time 300`, because five minutes is a stall for a file of a few tens of KB.
- **The installer's**, of the release asset, which is where the 500 was: it has no retry and no option to ask for one (its flags are `-b`, `-d`, `-h`, `-x`, the tag), so that attempt is repeated by running the installer again — three attempts, 5s then 10s apart.

`pipefail` closes a hole the old recipe had: when the fetch of `install.sh` failed, `sh` read an empty script and exited 0, so `make tools-golangci` reported a successful install of nothing. The script does not classify failures — the installer collapses every cause to exit 1, and reading its log text for a status would couple us to upstream's wording — so a genuinely broken pin costs the attempts before it fails, with the installer's own diagnostics on stderr. The pin is still read from `.tool-versions` and never chosen here, and the installed binary now has to report it.

## The bounded path is dormant for now

The installer's own download cannot be bounded by an option, so the whole invocation runs under `evener-dev dev bounded-list` — a subcommand that is not on main yet. It arrives with #1263. The script probes for it by running a trivial command under it and believing the answer, so until #1263 lands every run takes the unbounded path, including CI's, and says so on stderr. This is stated in the script header too.

The CI lint lane builds the dev binary first, in a step with `timeout-minutes: 5`, so that the bounded path is exercised there the moment the subcommand exists.

The other residual outlives that one: `make tools-golangci` on a fresh clone runs before anything is built, and a binary that is not there cannot bound anything.

## Gates

`bash -n`, `make lint` (8 modules) including `lint-generated`, root Makefile audits, and `make tools-golangci` against the real network on both paths — with the dev binary present (probe fails today, falls back, announces) and absent (same) — each checked for the announcement.

## Sequencing

This PR merges after #1289, and once #1289 is on main, main is merged into this branch — so the bounded path is live at the merge head rather than waiting for a later PR to turn it on. No timeout wrapper or bash fallback stands in for it in the meantime; the script probes, finds nothing, and says so.

## Round 1 fixes

`582582286`.

- **The inert bounded path** (Medium): no code change, per the sequencing above, which is now stated in the body.
- **No failure-path tests** (Low): two, with real curl and no network — the class with stubbed binaries is the one #1205 declined. A bad `EVENER_GOLANGCI_INSTALL_ATTEMPTS` exits 2 before anything is fetched, and the test checks that nothing was fetched (no curl diagnostic in the output). A download pointed at `http://127.0.0.1:1` — reserved, and refused on every machine — with two attempts fails both, says it is retrying, and gives up naming the count. They live in the root Go audits, where the repo already tests `scripts/ops` scripts.
  The installer URL is an env knob now, documented in the header alongside the backoff and curl's retry count, with the reason: a test that waited out the defaults would prove the same thing slowly, and a slow link is welcome to raise them.

Gates: `bash -n`, root Makefile audits (`go test -short -count=1 .`, which now includes the two new tests), `make lint` (8 modules), golangci-lint on the root package, and `make tools-golangci` against the real network.


obra:
## Review round 1 disposition (602fccfb0)

- **Medium, the bounded path is inert because `bounded-list` is not on main** — accepted by sequencing, not by a fallback: this PR merges after #1289 (the helper), then takes main so the bound is live at its merge head. No `timeout`-style supervisor; one helper is the design.
- **Low, no failure-path coverage** — accepted without stubs: the attempts validation (exit 2) and the retry loop against an unreachable URL (`http://127.0.0.1:1`, real curl, real refusal, deterministic, no network) get a test under the root audits; stubbed `curl`/`go` binaries are the class declined in #1205.

obra:
## Review round 2 disposition (582582286)

- **Medium, `bounded-list` is dead code in this diff** — sequencing, as the body says: this PR merges after #1289 and takes main first, at which point the probe succeeds and the build step is the bound. No fail-closed and no fallback. This finding is expected to persist until that merge and is not a reason for another round on its own.
- **Medium, the offline failure test inherits proxy and `HOME`, assumes port 1, and counts curl's wording** — accepted: isolated environment (`HOME` in a temp dir, loopback in `no_proxy`), a port reserved and released by the test, and the attempt count read from the script's own diagnostics. The script's curl invocation is not changed; proxied CI must keep working.
- **Low, the "no `curl` in output" assertion** — accepted: absence of the script's attempt diagnostic instead.
- **Low, no Unix/tool guard** — accepted: mirrors `requireFuzzScriptTools`.

obra:
## Round 2 fixes

Head `3ec2abff6`, all in `3ec2abff6`.

- **The bounded path** (Medium): no change. Sequencing stands — #1288 merges after #1289, then main merges into this branch — and no fallback or fail-closed was added.
- **The offline test made deterministic without touching curl** (Medium). It pointed at port 1 and matched curl's wording, which assumed two things about the machine: that nothing routes 127.0.0.1 through a proxy (a proxied runner has something very much listening), and that curl phrases a refusal the way this laptop's does. Now the port is reserved with `net.Listen("tcp", "127.0.0.1:0")`, read, and released, so it is a port nothing is on; the subprocess gets a minimal environment — PATH, a `HOME` of its own so no `~/.curlrc` or `~/.netrc` joins in, and `no_proxy`/`NO_PROXY` for the loopback address; and the attempt count comes from the script's own diagnostics (`install attempt 1 of 2 failed; retrying`, `did not install in 2 attempt(s)`). The script's curl line is untouched, so a proxied CI keeps working.
- **The guard case** (Low) now asserts that no attempt diagnostic appears, rather than that the word curl does not.
- **Both cases skip where they cannot run** (Low), in `requireFuzzScriptTools`' shape as a sibling helper: not on Windows, not without bash, and not without curl for the one that reaches a download.

Gates: root Makefile audits, the two tests at `-count=3`, `bash -n`, `make lint` (8 modules), golangci-lint on the root package.

obra:
## Review round 3 disposition (3ec2abff6)

CI green. RoboRev: the Medium is the sequencing finding again, confirmed by-design by a second reviewer; it resolves when #1289 merges and this PR takes main. Three Low, all accepted — leading zeros refused by the backoff validation (octal in bash arithmetic), the offline test keeps its listener open and closes each accepted connection so no other process can reclaim the port, and the tool guard covers `awk` and `go` as well.

obra:
## Round 3 fixes

Head `2da210b92`, all in `2da210b92`.

- **The Medium** is sequencing again: no change.
- **A backoff bash reads as octal** (Low). `EVENER_GOLANGCI_INSTALL_BACKOFF=08` passed a digits-only check and then failed inside the arithmetic that uses it, so the failure was the retry's rather than the guard's. Leading zeros are refused now, for the curl-retry knob too, with a test that 08 exits 2 before any fetch.
- **The released port** (Low). Reserving a port and closing the listener before curl connects leaves a window for anything else on the machine to take it. The listener stays up for the test and accepts each connection and closes it, so every attempt meets the same real transport failure — an empty reply — from a port nothing else can claim. Attempts are still counted from the script's own diagnostics.
- **The tool guard** (Low) covers what the script actually reaches before it decides anything: bash, awk (the pin out of `.tool-versions`) and go (where the bindir is), plus curl for the case that downloads.

Gates: root Makefile audits with these three tests at `-count=3`, `bash -n`, `make lint` (8 modules), golangci-lint on the root package.

obra:
## Review round 4 disposition (2da210b92)

- **Medium, bounded path inert until the helper lands** — sequencing, as before; this PR merges after #1289 and takes main first. Not a reason for a round.
- **Low, `--retry-all-errors` needs curl 7.71+** — accepted as a loud precondition: the script checks the curl version once and fails with a clear message; no portable retry reimplementation.
- **Low, the version-pin verification has no coverage** — accepted: a loopback server serves an install script that writes a binary reporting a controlled version, and both the accept and the mismatch paths are asserted. That exercises the script's real verification against a controlled artifact, which is not the stubbed-toolchain class declined before.
- **Low, stale test header** — accepted.

obra:
## Round 4 fixes

Head `42cd75ab3`, all in `42cd75ab3`.

- **`--retry-all-errors` needs curl 7.71+** (Low). Checked once at the top; an older curl exits 2 naming the minimum. No feature detection and no fallback — a retry reimplemented here is the shell supervisor this change exists to avoid. No separate test: the check runs on every real invocation, including CI's and the three tests below, which all reach it before anything else.
- **The version-pin verification is tested** (Low), which is the part that catches an installer writing a different release or a bindir already holding an older binary. Three cases over a loopback server serving a real install script that writes a small executable printing a version line the test chooses: the pin matching exits 0, another version exits 1 naming the mismatch against the pin, and an installer that writes nothing exits 1 saying the binary does not run (the hole `pipefail` cannot see). Real curl, real bash, real go, nothing beyond loopback; the artifact under test is a binary, not a stub of the script's logic; the bindir is the temporary HOME's GOPATH, so the developer's is untouched. Mutation-checked: a version line one character off turns the accepting case red.
- **The header** (Low) now describes the listener the failure cases use — held for the test and closing every connection — and why: releasing a reserved port leaves a window in which anything on the machine can take it.

Gates: root Makefile audits with all six tests at `-count=3`, `bash -n`, `make lint` (8 modules). Merged the last green main (`1d5e12100`) rather than `origin/main`, per the standing instruction; this branch does not contain `c09997369`.

obra:
## Review round 5 disposition (42cd75ab3)

CI green. RoboRev (3 of 4 reviewers with findings):
- **Medium, bounded path inert until the helper lands** — sequencing, unchanged: this PR merges after #1289. Because of that order, the round after #1289 lands can exercise the bounded path against the real binary; no stub binary.
- **Medium, oversized attempt counts overflow bash arithmetic** — accepted as a cap.
- **Medium, the served install script ignores the tag** — accepted: the stub records the tag and the test asserts `v<pin>`.
- **Medium, the probe reads the checkout's `evener-dev`, so tests depend on ambient state** — accepted: the binary path is overridable and the tests pin it, so the unbounded path is deterministic; the bounded-path test arrives with the real binary as above.
- **Medium, tests fail rather than skip on curl older than 7.71** — accepted; the helper mirrors the script's version check.
- **Low, curl floor undocumented; diagnostics swallowed under `set -e`; `EVENER_GOLANGCI_CURL_RETRIES` guard untested** — all accepted.

obra:
## Round 5 fixes

Head `7f7740d72`, all in `7f7740d72`.

- **Sequencing** (Medium): no change.
- **Knob overflow** (Medium). Each numeric knob takes 0-20 (attempts 1-20) and the bound is in the pattern, not in an arithmetic comparison — a twenty-digit value is not a number bash can compare, so an arithmetic check would have failed as arithmetic instead of as the diagnostic. Cases: `99999999999999999999`, `21`, `0`, `abc`, and `08`/`abc`/`21` for the retry knob.
- **The served installer records its arguments** (Medium), and the accepting case asserts it was asked for `v<pin>` — the other half of what this script is for.
- **Ambient `evener-dev`** (Medium): `EVENER_GOLANGCI_DEV_BIN` says where the binary is, and the tests point it at a path that does not exist, so which half of the script they exercise no longer depends on whether someone ran `make build-dev` in the worktree. No stub binary. The header records that the bounded half gets its test when bounded-list reaches main (#1263).
- **The tool guard** (Medium) reads curl's version the way the script does and skips below 7.71, rather than every case failing on the version refusal.
- **The 7.71 floor** (Low) is on the `tools-golangci` annotation, which is where that target's prerequisites are listed and what the generated docs table shows.
- **Guarded lookups** (Low). Under `set -euo pipefail` a failed command substitution takes the script with it silently: curl's version, the pin in `.tool-versions` and `go env GOPATH` are each guarded, and a missing curl is named rather than inferred.

Gates: root Makefile audits with the installer tests at `-count=3`, `bash -n`, `make lint` (8 modules, including `lint-generated` after `make generate`), golangci-lint on the root package. Merged nothing new from main; this branch does not contain `c09997369`.

obra:
## Review round 6 disposition (7f7740d72)

CI green. RoboRev: the Medium is the sequencing finding again, unchanged (this PR merges after #1289; the bounded-path test against the real binary follows in that order). Two Low, both accepted — the duplicated `abc` block goes, and the other-knobs guard test asserts no install attempt was made, like its siblings.

obra:
## Round 6 fixes

Head `3e95ac96f`, all in `3e95ac96f`.

- The standalone `abc` block after the attempts loop is gone: the loop already covers it.
- The retry knob's cases carry the same no-attempt-was-made assertion as their siblings, so each one proves the guard ran before any download rather than only that it ran.

Gates: root Makefile audits with the installer tests at `-count=3`, `make lint` (8 modules). Nothing new merged from main.

obra:
## Review round 7 disposition (3e95ac96f)

- **High, unbounded download until the helper lands** — sequencing, unchanged: merges after #1289.
- **High/Medium, tests may write into the real `GOPATH/bin`** — the reviewers disagree and the review says to verify, so it is measured: the subprocess env carries only `PATH`, a temp `HOME` and the proxy bypass, so `go env GOPATH` should derive from that `HOME`. Whatever the measurement says, `GOPATH` becomes an explicit temp path in the test env so the isolation is stated, not inferred.
- **Low, the attempts message omits the leading-zero wording** — accepted.

obra:
## Round 7 fixes

Head `a175639ab`, all in `a175639ab`.

**The measurement.** The third reviewer was right: the temp HOME does redirect GOPATH, and these tests never wrote to the developer's bin.

- `go env GOPATH` in my normal shell: `/Users/jesse/go`; with the tests' environment (`HOME=<t.TempDir()> GOENV=off GOFLAGS=`): `/var/folders/.../T/tmp.XXXX/go`.
- `~/go/bin/golangci-lint` mtime before a full run of the install tests: `1789364030`; after: `1789364030` — unchanged. Repeated after this round's change, still unchanged.

Reading that took two steps, which is presumably why two reviewers read it the other way, so the isolation is now stated rather than derived: `GOPATH` is set explicitly in the environment the tests build, beside `GOENV=off`, and the accepting case asserts the installed binary landed under that path.

- **The attempts message** (Low) gains the "without a leading zero" its two siblings carry.
- **Sequencing** (High): no change.

Gates: root Makefile audits with the installer tests at `-count=3`, `make lint` (8 modules), `bash -n`, `go vet`. Merged `origin/main` now that it is green.

obra:
## Review round 8 disposition (a175639ab)

CI green. RoboRev: three of four reviewers clean. The Medium is the sequencing finding, acknowledged by two reviewers as the documented gap pending #1289; unchanged. The Low is accepted, and it is the same defect class #1265 hit: a phony `build-dev` prerequisite rebuilds the binary after the bounded CI step, unbounded. The installer already probes for the binary and says what it is doing without it, so the prerequisite goes; the bounded CI step stays as the one build.

obra:
## Round 8 fixes

Head `485b3f12a`, all in `485b3f12a`.

- **The lint lane built the binary twice** (Low). The CI step builds it with `timeout-minutes: 5`, and then `lint-golangci`'s phony `build-dev` prerequisite rebuilt it immediately, unbounded — the class #1265 round 3 fixed for the test gate. The prerequisite is gone from `lint-golangci`; it checks for the binary instead and names the target that builds it. Verified: after `make build-dev`, `make -n lint-golangci` shows no `go build`; without the binary it fails with `lint-golangci: ./evener-dev is missing; run \`make build-dev\` first`; with it, `PASS lint-golangci`.
- One note on scope: `tools-golangci` never had that prerequisite on this branch — it is the *installer* that probes for the binary. The two targets that still carry it are `lint-naming` and `lint-internal`, which exec `./evener-dev` the same way; I left them, because removing the prerequisite there would make a fresh clone's `make lint` fail rather than build, and the durable fix is #1263's file target, which makes the prerequisite a no-op when the binary is current. Say the word if you want them changed here too.
- **Sequencing** (Medium): no change.

Gates: root Makefile audits (including `TestMakeLintRunsEveryGateInLintTargets`, whose map keys on the recipe fragment `module-lint`, unchanged), `bash -n`, `make lint` (8 modules), `lint-generated`.

obra:
## Review round 9 disposition (485b3f12a)

CI green. RoboRev (3 of 4 reviewers with findings):
- **Medium, bounded path inert and untested** — sequencing, unchanged: #1289 merges first, then this PR takes main and adds the bounded-path test against the real binary.
- **Medium, dropping the `build-dev` prerequisite races `make -j lint` and can lint a stale binary** — accepted, and round 8 is reversed: the prerequisite comes back exactly as it was. The double build it causes in CI is the lesser problem, and it disappears when #1265's real file target lands; that is the durable fix, not a recipe check.
- **Low, backoff validation proven for one bad value only** — accepted: same table as the sibling knobs.
- **Low, curl's own retry never exercised** — accepted: a first response of 500 then the real script, one attempt, one curl retry, exit 0.

obra:
## Round 9 fixes

Head `96b1356b7`, all in `96b1356b7`.

- **Round 8 reversed** (Medium). Dropping `build-dev` from `lint-golangci` traded a double build for a worse failure: under `make -j lint` that gate can start before the one that builds the binary and lint a stale `./evener-dev`, or none. `make/linting.mk` is now byte-identical to main. The CI lane therefore builds the binary twice — once in its own bounded step, once through the phony prerequisite — until #1263's file target lands and the prerequisite costs nothing when the binary is current. Verified with `make -j4 lint`: all nine gates pass.
- **The backoff knob** (Low) gets the table its siblings have: `08`, `abc`, `21` each exit 2 with the diagnostic and no download attempted.
- **`--retry-all-errors` is tested** (Low). A loopback server closes the first connection without a reply and serves the second. That matters: a 500 *is* one of the transient statuses plain `--retry` already retries, so a 500-then-success case would have passed with the flag removed — I tried it, and it did. An empty reply is not transient, so recovering from it is exactly what the flag buys, and removing it from the pipeline turns the case red.

Gates: root Makefile audits with the installer tests at `-count=3`, `bash -n`, `make lint`, `make -j4 lint`, `lint-generated`.

obra:
## Review round 10 disposition (96b1356b7)

- **Medium, retried `curl` output piped straight into `sh`** — accepted; a real one. A transfer that dies after emitting part of the script can have the retry's bytes appended and the whole executed as one malformed script. The installer is downloaded to a temp file and run only after `curl` exits 0.
- **Medium, bounded path inert until #1289** — the review itself notes no fix is possible in this diff; sequencing, unchanged.
- **Low, doc comment above the wrong helper** — accepted.

obra:
## Round 10 fixes

Head `ec57ba582`, all in `ec57ba582`.

- **`curl | sh`** (Medium). A reply cut off halfway is a script the shell has already begun executing, and curl's retry appends the second reply to the first — the installer runs twice, the first time from the middle. It is downloaded to a file under TMPDIR now, removed on exit, and run only after curl exits 0; the `pipefail` form goes with it, since it existed to notice the download failing on the left of a pipe. The bound still wraps the whole invocation.
- **The new case** serves a truncated body and closes the connection, then serves the whole script: exit 0, one install. Under the old pipe form it fails with `curl: (18) transfer closed with 135 bytes remaining to read` — the shell having been handed a fragment — which is how I know the case tests the change. The closed-connection retry case still passes.
- **The comment paragraph** (Low) moved above `installerServing`, the function that starts the server; `installerScript` keeps its own line.
- The root audits caught my first temporary file: `mktemp -t` ignores TMPDIR on macOS, so the leftover checks would not have seen it. It uses an explicit `"${TMPDIR:-/tmp}/…"` template.

Gates: root Makefile audits with the installer tests at `-count=3`, `bash -n`, `make lint` (8 modules), and `make tools-golangci` against the real network, which installed v2.13.1 and left no temporary file behind.

obra:
## Review round 11 disposition (ec57ba582)

CI green. RoboRev:
- **Medium, the whole-installer test cannot tell one run from "fragment plus whole"** — accepted: the served installer writes a sentinel early in its body and the test asserts it appears exactly once.
- **Medium, bounded path inert until #1289** — sequencing, unchanged.
- **Low ×5** — all accepted: trap armed before `mktemp`; the duplicated doc paragraph and the stale test name fixed; the curl floor applied only when a test reaches curl, with absent and too-old told apart; a retry-delay knob so the two retry tests stop paying a 2 s backoff each.

obra:
## Round 11 fixes

Head `5bd0109b9`, all in `5bd0109b9`.

- **The whole-installer case can now tell one run from two** (Medium). The served script appends a line the moment it starts, before the point the truncated body is cut at, and the test asserts exactly one. Under the old pipe form the case fails outright — `curl: (18) transfer closed with 195 bytes remaining`, the shell having been handed a fragment — which I re-ran on a copy to confirm.
- **The trap is armed before `mktemp`** (Low), with `${install_script:-}`, so a failure between the two leaves nothing behind.
- **The stale paragraph** on `installerScript` is gone (Low) and it describes what the function does now; the comment naming `TestInstallerRefusesABackoffBashWouldReadAsOctal` names the test that exists (Low).
- **The curl floor** (Low) is checked only when a case asks for curl, and absent is distinguished from too old in the skip message.
- **`EVENER_GOLANGCI_CURL_RETRY_DELAY`** (Low, default 2, validated like its siblings, documented in the header) is 0 in the tests, so the retry cases no longer pay 2s each — the installer suite is about 4s at `-count=3`.

Gates: root Makefile audits with the installer tests at `-count=3`, `bash -n`, `make lint` (8 modules).

obra:
## Review round 12 disposition (5bd0109b9)

CI green. RoboRev: the Medium is the sequencing finding, unchanged (merges after #1289). Two Low, both accepted: the duplicated `installerServing` paragraph above `installerScript` goes, and the comment that narrates a `pipefail` form that never existed is replaced by a statement of what the file-based fetch guarantees now.

obra:
## Round 12 fixes

Head `c5b91b8b0`, all in `c5b91b8b0`.

- The duplicated `installerServing` paragraph above `installerScript` is gone; each function carries its own comment.
- The fetch comment states what holds rather than what changed: the installer is run only after curl has exited 0, so a download that failed or arrived in part is never executed. The `pipefail` narration was also wrong about the old recipe, which ran under `set -eu` alone.

Gates: `bash -n`, root Makefile audits with the installer tests at `-count=3`, `make lint` (8 modules).

obra:
## Review round 13 disposition (c5b91b8b0)

CI green. RoboRev:
- **Medium, unbounded fallback until #1289** — sequencing, unchanged.
- **Medium, the availability probe runs a different flag set from the real invocation** — accepted: one shared flag list builds both, so the probe cannot pass while the real run fails.
- **Low, no guard tests for the two curl knobs** — accepted where true: the retry-delay knob is new since the last table and joins it; if the retry-count knob is already in the other-knobs table, that is stated rather than duplicated.

obra:
## Round 13 fixes

Head `093bcb0b7`.

**1 (M) — one place builds the bounded run.** `bounded_list_flags=(-attempts 1 -grace 5s)` and a `run_bounded <timeout> <command...>` that spells out the binary, the subcommand, the timeout and the `--`. The probe is `run_bounded 10s true` and the download is `run_bounded 300s sh -c "$fetch_and_run" …`, so the timeout is the only thing that differs — which is the one thing that should, since what the probe waits for is the binary answering at all, not a download.

The bounded path still has no behavioural test (it needs bounded-list on main, the follow-up in the script's header), so what is pinned instead is the property the finding is about: `TestTheBoundedDownloadIsBuiltInOnePlace` asserts the script names `"$evener_dev_bin" dev bounded-list` exactly once. Mutation: write the probe out longhand again → "the script runs … in 2 places, want one".

**2 (L) — neither curl knob was covered.** `TestInstallerRefusesABackoffItCannotUse` and `TestInstallerRefusesAnAttemptCountItCannotUse` were the only two knob guards under test; `EVENER_GOLANGCI_CURL_RETRIES` and `EVENER_GOLANGCI_CURL_RETRY_DELAY` had none, so both are new. `abc`, `08`, `21` each with exit 2, the knob named with what it takes, and no attempt before the guard; `0` stays out of the table because it is a real setting for both, and the one the tests themselves run with.

The assertion all four share is now `requireKnobRefusal(t, knob, values...)` rather than a fourth copy. Mutations, one guard at a time: drop the `curl_retries` check → `EVENER_GOLANGCI_CURL_RETRIES=abc: exit code = 1, want 2` (the bad value reaches curl and the run fails as a download); drop the `curl_retry_delay` check → the same for that knob.

**3** — no change, as ruled.

Gates, all foreground: `bash -n scripts/ops/install-golangci-lint.sh`; `go test -count=3 -run 'TestInstaller|TestTheBoundedDownloadIsBuiltInOnePlace' .`; root audits `go test -short -count=1 .`; `make lint` (9 gates). `origin/main` already merged.

obra:
## Review round 14 disposition (093bcb0b7)

CI green. Both findings are the two standing declines, so this round makes no change:
- **Medium, bounded path inert until the helper lands** — sequencing: this PR merges after #1289 and takes main first.
- **Low, exercise the bounded branch with a stub `evener-dev`** — declined as the stubbed-tool class; once #1289 is on main, the next merge of main here adds the bounded-path test against the real binary, and the probe's invocation contract is pinned there.

No further rounds on this head; it waits for #1289.

obra:
Closed on Jesse's ruling (2026-09-14): #1265 was overbuilding, and this half of it goes with it. Replaced by a ten-line fix: curl --retry on the golangci-lint download and a plain timeout around the gate's go list; the Metro bundle gate (#1245) gets a plain CI timeout instead of the process-lifecycle helper.

Issue #1285, opened 2026-09-14
One flag parser for the gate and the shard runner


obra:
#1265 adds `evener-dev dev list-build-flags`, which walks a caller's `go test` flags and answers which of them the package enumeration also needs (`-tags`, `-race`, `-msan`, `-asan`). To do that safely it carries a table of the flags whose value is the next argument, so `-run -race` is read as a regex rather than a sanitiser.

#1266 grows the full parser for the same flag language in `cmd/evener-dev/shardplan.go`: build flags to `go test -c`, a forwarded set to the shard binaries, everything else refused by name, with the value-taking flags consumed the same way.

Two tables describing one rule. Neither PR can absorb the other while both are open, so this is the fold: once both are on main, `list-build-flags` should ask the shard runner's parser which flags change package selection instead of keeping its own list, and the value-taking table should exist once.

PR #1289, opened 2026-09-14
feat(evener-dev): bounded-list, list-build-flags and one process-group discipline (#1263)

obra:
Two subcommands the test gate needs, and the process-group discipline they share with the shard runner. Split out of #1265, which bundled them with the gate wiring and the installer; this PR is Go only, and the gate wiring stays there.

## `bounded-list`

`go list ./...` on a host whose Go caches sit on a stalled volume blocks forever: the gate waits, the job hits the CI-level limit, and a required check fails with no diagnostic on a tree whose code was fine (run 34639098143). This runs a command under a per-attempt bound and retries an attempt that hits it — a killed attempt still warms the caches, so a merely slow runner passes on the next one, while a stalled volume fails with a named diagnostic.

- exits **124** on a timeout (coreutils' convention), **128+signal** when interrupted, the command's own status otherwise; all four are in the usage text;
- passes the command's output through on **every** outcome, because a half-written list is the evidence a caller works from, and **fails the run if that handover is short** rather than letting a caller test a truncated list and report a pass for the rest;
- stops a **process group**, not the one child: `go list` forks compilers, and a survivor goes on holding the build and module cache locks every later run on that host needs;
- forwards **TERM, INT and HUP** to that group — a command in its own group no longer hears the terminal's signals — and latches the signal, so one arriving during cleanup or between attempts still stops the run rather than starting another attempt;
- bounds the wait **after** SIGKILL as well: a child in an uninterruptible kernel wait never reaps, so the attempt says so, returns 124, and leaves it to init instead of hanging where the bound was supposed to be.

## `list-build-flags`

`go list` and `go test` do not see the same tree: `-tags` selects files and so can select whole packages, and `-race`, `-msan` and `-asan` each set a build tag of their own. A package whose files all sit behind `//go:build race` exists for `make test-race` and not for the enumeration that decides what that run tests. This says which of a caller's flags the enumeration also needs. A flag's value is consumed with it, so `-run -race` stays a regex, and `-args` ends the flags, after which the words belong to the test binary.

Its table overlaps the one #1266 grows for the shard runner; folding them is #1285.

## `internal/devtool/procgroup`

- `StopWith` forwards the caller's own signal and gives the group the grace to act on it before the TERM-then-KILL escalation; `Stop` is now that function with SIGTERM, so there is one implementation;
- `Exists` answers whether a group still has members — safe to ask after the direct child is reaped, because a pid cannot be recycled while it is still a live group's id; its comment records the declined case, a group of zombies, which reads as alive until init reaps them;
- both refuse pgid 0 and negatives.

`boundedlist_unix.go` and `boundedlist_other.go` do not exist: the helper calls this package rather than keeping a second copy, which is also what fixes the exit status of a signal-killed child (`ExitError.ExitCode()` is -1, so it used to be reported as 255).

## Tests

Real processes throughout: a TERM-ignoring group whose group must be empty afterwards, a leader that exits leaving a child behind, a grandchild that leaves the group and holds the pipe so the reap cannot finish, a child that answers HUP by saying so on stdout, and a fixture module with a package behind a build tag that `go list` sees only when the flag is forwarded. The escapee is this test binary re-executed, following the repo's helper-process idiom (`internal/devtool/scratch`), so nothing depends on an ambient tool. Each behaviour above was mutation-proved red before it was called done.

## Gates

Toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/...`, root Makefile audits, and `golangci-lint` over both trees. All foreground.


obra:
## Review round 1 disposition (9183f1b52)

CI green. RoboRev (4 reviewers), all accepted:
- **Medium, the survivor sweep never confirms the group is gone after SIGKILL** — a still-live group after a bounded wait is terminal, through the same path an unreapable child already takes.
- **Medium, `-compiler` and `-overlay` dropped by `list-build-flags`** — the table becomes "every `go list` flag that changes which packages or files are selected", enumerated from `go help build`, with fixtures.
- **Medium, `list-build-flags` missing from the top-level alias** — added, and covered by the alias test.
- **Medium, alias test execs `echo`** — the test binary is the command.
- **Low, fixed sleeps in the cannot-reap tests** — readiness signalled by the child, not by elapsed time.
- **Low, usage line, exit status 1, dead `published` field** — all fixed.

obra:
## Round 1 fixes

Head `0b8aa5bac`, all in `0b8aa5bac`.

- **A group that will not clear** (Medium). The sweep sent SIGKILL and returned without looking. That group is the same situation as a child that cannot be reaped — the caches stay locked, and another attempt adds a second one — so both end the attempt the same way now: 124, a diagnostic naming the process group, no retry, one field (`stuck`) and one code path. The case cannot be produced on demand, which is what SIGKILL is for, so the probe is injected and the rest of the path is real: TERM, the grace, KILL, the grace again, with cases for the group that goes on TERM (no escalation, nothing said about SIGKILL), the group that never goes, and the empty group (silent).
- **The forwarded table** (Medium) is now what it claimed: every flag `go help build` documents that changes which packages or files are selected. `-mod`, `-modfile`, `-compiler` and `-overlay` join `-tags` and the three sanitisers, each with its reason; the ones left out are named with theirs — the optimiser, linker, naming and coverage flags change how the same packages are built, and `-C` moves a directory the enumeration has already chosen. Fixtures: an overlay that introduces a package the disk does not have (it appears only when the flag is forwarded), and `-compiler=gccgo` in the forwarded list.
- **`list-build-flags` in the top-level aliases** (Medium), and in the alias test.
- **No ambient `echo`** (Medium): the alias comparison runs this test binary re-executed, the helper-process idiom already used here, and still compares exit code and stdout across both spellings on one fixed invocation.
- **No elapsed-time sleeps** (Low): the child writes one file when it is running and another when it has been sent SIGTERM; the tests that interrupt a running command, or need a signal to land inside the cleanup's grace, wait on those files. A sleep long enough on this macOS laptop is a guess on the Linux runner, and the local gates here are macOS only.
- **The usage text** (Low ×2): `bounded-list` and `list-build-flags` each have a line, and the exit statuses now include 1 — the command could not be started, or its output could not be handed on in full.
- **`waitResult.published`** (Low) is gone; the close of the channel carries the ordering, and the comment says so.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/...`, root Makefile audits, and `golangci-lint` over both trees. All foreground.

obra:
## Review round 2 disposition (0b8aa5bac)

CI green. RoboRev: two of four reviewers clean; one Medium and one Low, both accepted — `list-build-flags` checks its writes and fails when output cannot be delivered, and the tag fixture isolates `GOFLAGS`/`GOENV` like its overlay sibling. One carry-in from my own read: the last elapsed-time signal in `boundedlist_unix_test.go` (a 200 ms sleep before SIGTERM) goes the way of the others.

obra:
## Round 2 fixes

Head `6f06c8c41`, all in `6f06c8c41`.

- **`list-build-flags` ignored its writes** (Medium). A caller reading a truncated answer enumerates under fewer flags than the tests are built with — the failure this subcommand exists to prevent — so each `Fprintln` is checked, and a write that did not land exits non-zero naming the flag, the rule bounded-list's output handover already follows. Tested with a writer that refuses.
- **The enumeration fixtures' environment** (Low) is one helper both call: their own `go.mod` rather than this repo's workspace, and none of the developer's toolchain settings. A GOFLAGS carrying `-tags` would have answered the very question those fixtures ask.
- **The last elapsed-time signal** (carry-in) is gone. The escapee's grandchild writes a readiness file once it is running and holding the pipe, the middle process waits for that before reporting its own, and the interrupt case waits on that file before signalling. The only `time.Sleep`s left are a 20ms poll interval and the helpers' own holds, neither of which schedules anything.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/...`, root Makefile audits, and golangci-lint over both trees. All foreground.

obra:
## Review round 3 disposition (6f06c8c41)

- **Medium, alias test covers one alias and checks the others vacuously** — accepted: every alias compared both spellings the same way.
- **Medium, signal-vs-finish race lacks the `finishedFirst` check the timeout branch has** — accepted; one rule for both branches.
- **Low, pipe reader leak; `-C` value not skipped; temporal comment naming another PR** — all accepted.
- **Low, zombie-only group now ends an attempt as stuck/124** — accepted pending measurement, because round 2 changed the cost the earlier decline rested on. The implementer tries to construct a zombie-only group that outlives the reaped child on Linux and macOS; if it cannot be built without a subreaper, `Exists`'s comment states the real cost and the decline stands; if it can, `Exists` learns to tell zombies from live processes.

obra:
## Round 3 fixes

Head `64ab10d67`, all in `64ab10d67`.

- **A simultaneous finish and interrupt** (Medium). The signal branch took the signal without asking whether the command had already finished, and `select` picks between ready cases at random. It checks `finishedFirst` first now, and a completed attempt keeps what the command decided while still latching the signal — so the runner stops without starting another attempt and without rewriting the answer. Tested with the ordering injected both ways (`completeWith` with a signal waiting: exit code kept, signal latched; without: neither).
- **The alias comparison** (Medium) now covers `list-build-flags` exactly as `bounded-list`: one fixed invocation each, run both ways, exit code and captured output compared. `module-lint` and `agent-shards` would lint or shard the repo if run for real, so those are compared on the exit code alone — the builders the old check passed were never written to, since the dev family writes the process's own streams, which is what made that assertion vacuous. They are gone.
- **The pipe reader** (Low) is closed after `io.ReadAll`.
- **`goTestValueFlags`' comment** (Low) states the rule — every flag whose value is the next argument belongs there so a value is never read as a flag — with no PR numbers.
- **`-C`** (Low) joins that list: `-- -C -race ./...` forwards nothing, because the word after `-C` is a directory however it is spelled.
- **Zombie-only groups** (Low, measured). On this Mac, 24 runs of a leader that starts a grandchild in its group, the grandchild exiting first and the leader exiting without waiting: after the leader is reaped the group answers `kill(-pgid, 0)` for **3-12ms**, then launchd has reaped the orphan and it answers ESRCH. It never survived to a second 10ms poll, let alone to a grace, so the cost is one diagnostic and one poll — not the stuck-and-124 path, which needs something that holds zombies indefinitely (a subreaper that adopts and does not reap). Linux is unmeasured: no Linux host here and the Docker daemon is down. `Exists` is unchanged and its comment now carries that measurement and that caveat.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/...`, root Makefile audits, golangci-lint over both trees. All foreground.

obra:
## Review round 4 disposition (64ab10d67)

CI: `tests` failed on `TestDelegateAttention_SettledResidentChildStartsExactAttentionGeneration` in the agent package, which this Go-only PR does not touch. Under investigation as a flake or a main-borne failure; the job is being rerun and the test's history checked. Not attributed to this PR yet, and not ignored either.

RoboRev (2 of 4 reviewers with findings):
- **High, the gate does not invoke the new subcommands** — declined as sequencing: this is the helper alone; the gate wiring is #1265, rebased onto main once this merges (the reviewer's own note).
- **Medium, an interrupt during a finished command's cleanup is never polled** — accepted in part: the latch is polled after completion cleanup so a failed command is not retried after an interrupt; the exit status stays the command's own, per round 3's rule (a command that finished has answered).
- **Medium, the alias test runs `module-lint` for real when golangci-lint is on PATH** — accepted: aliasing asserted through a dispatch seam, no subcommand executed.
- **Medium, the cannot-reap tests race a 200 ms bound against binary startup** — accepted: the bound starts only after the escapee's ready signal.
- **Low, `completed` overwrites a latched interrupt with 124** — accepted; same answer as the give-up path.

obra:
Rerun of the `tests` job at 64ab10d67 passed; the attention-test failure was a flake, filed as https://github.com/prime-radiant-inc/evener/issues/1294. CI is green at this head; round 4 findings stand as dispositioned above.

obra:
## Round 4 fixes

Head `ec5a51a72`, all in `ec5a51a72`.

- **An interrupt during cleanup after a finished command** (Medium). The attempt returned without looking at the latch again, so a signal that arrived during the survivor sweep — which spends a grace, or two — was recorded nowhere and the runner had nothing to stop for. It polls once cleanup is done, keeps the command's own exit status, and carries the signal. Test: a command that exits 3 and leaves a TERM-ignoring child, interrupted while the sweep waits, attempts=3 → one attempt, exit 3, `interrupted` set; mutation-proved red without the poll.
- **124 vs 128+signal when the group will not clear** (Low): 124 only when there is no signal, matching the give-up path; covered by the same test's first half.
- **The alias test ran the aliases** (Medium). `module-lint` lints the whole repository — which `make test` was then doing twice — and `agent-shards` shards it. The alias table is data now, and the resolution is asserted where it is decided: every aliased name is a dev subcommand, so both spellings reach the same handler. The two cheap hermetic ones (`bounded-list`, `list-build-flags`, which re-execute this test binary) keep their real both-ways comparison.
- **The give-up tests raced their own fixture** (Medium). The middle process now exits as soon as its grandchild holds the pipe, so the unreapable state is entered by the command rather than at an elapsed time; the bound is a tripwire two orders of magnitude clear of the fork it waits on, and a machine slower than that fails loudly rather than passing for the wrong reason. The interrupt-during-cleanup fixture also installs its trap before the leader exits — a SIGTERM that arrives first finds the default disposition and the child dies without a word, which is why the first version of that test passed for no reason.

Merged the last green main (`1d5e12100`) rather than `origin/main`, per the standing instruction; this branch does not contain `c09997369`. The `TestDelegateAttention_...` failure is in the agent package, untouched here; I have not chased it, per your note.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/...`, root Makefile audits, golangci-lint over both trees. All foreground.

obra:
## Review round 5 disposition (ec5a51a72)

CI green. RoboRev:
- **Medium, helpers not wired into the gates** — declined, as in rounds 1 and 4: the gate wiring is #1265, rebased onto main once this helper merges. That is the split.
- **Medium, `completed` treats a `cmd.Wait` I/O error with exit 0 as success** — accepted: a non-exit error from `Wait` (an output copy failure) fails the attempt with status 1 and the error on stderr, since the captured output cannot be trusted.

obra:
## Round 5 fixes

Head `2034f58e2`, all in `2034f58e2`.

- **A wait that failed is not a command that succeeded** (Medium). `cmd.Wait` returns the command's status as an `*exec.ExitError`; anything else it returns is a failure on this side, most of all a copy of the command's output that did not finish. The attempt read only the process state, so a `go list` that exited 0 while its output was being lost was reported as a success with a short list — which a caller cannot tell from a complete one. A non-`ExitError` with a zero status is now exit 1, with the error on stderr, and nothing is retried. The `ExitError` path is untouched.
- Tested with a stderr writer that fails its first write and records the rest (so the diagnostic still has somewhere to land), plus the other half of the branch: `exit 7` still reports 7 and says nothing about output. Mutation-proved by inverting the `ExitError` test.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/...`, root Makefile audits, golangci-lint over both trees. All foreground. Nothing new merged from main; this branch does not contain `c09997369`.

obra:
## Review round 6 disposition (2034f58e2)

CI green. RoboRev: the gate-wiring Medium is declined for the fourth time (the split: #1265 wires the gate once this merges). One Low accepted: the test helpers call `t.Fatalf` from spawned goroutines; the wait result goes over a channel so the test goroutine fails promptly.

obra:
## Round 6 fixes

Head `d08fb481f`, all in `d08fb481f`.

- **t.Fatal off the test goroutine** (Low, and a real one). Eight goroutines called `awaitFile`, which fails with `t.Fatalf`; from any goroutine but the test's that calls `runtime.Goexit` on the wrong stack, so the test stalls to its `-timeout` instead of failing with the message. `awaitFileErr` returns the error instead, and `signalWhenReady` wraps the whole pattern: the waiting and the signalling happen off the test goroutine, and the test goroutine calls the returned check afterwards, which is where `t.Fatal` belongs. `awaitFile` stays as the wrapper for test-goroutine use, with a comment saying which is which.

Gates: toolchain gofmt, `go vet` (host, `-tags evenerfuzz`, `GOOS=windows`) clean over both trees, `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/...`, root Makefile audits, golangci-lint. All foreground. Merged `origin/main` now that it is green.

obra:
## Review round 7 disposition (d08fb481f)

- **Medium, gate not wired** — declined for the fifth time; the split puts that in #1265.
- **Medium, PGID reuse between the existence check and the kill** — declined, with the reasoning written where the sweep lives: a PID cannot be recycled while it is a live group's id, so a group that answers is ours until its last member is gone, after which the sweep sends nothing; the check-to-kill window is microseconds against a sequential PID space. No per-platform process-identity probing.
- **Low, a command that exits right at the bound while its output is still draining is reported as a timeout** — accepted, in the race-free form: once the wait has returned, an exit status that is not the signal the helper sent means the command finished on its own, and the attempt takes the command's answer; `WaitDelay` bounds a stuck copier. Tested with the ordering injected. The package doc lists the two subcommands.

obra:
## Round 7 fixes

Head `07bd174de`, all in `07bd174de`.

- **A command that decided before the bound** (Low, accepted). After the stop, the command's own status decides: an exit status, or a death by a signal this helper did not send, means it finished on its own and that answer stands; only a death by our SIGTERM/SIGKILL, or no status, is still a timeout. Tests with the ordering injected — exits 0 during the drain → 0, exits 3 → 3, killed by our TERM → unchanged.
  One narrowing: this applies on the timeout path only. Applied to an interrupt as the disposition worded it, a child that answers SIGHUP by exiting cleanly would report 0 — calling an interrupted run a success, which is what round 3's 128+signal exists to prevent. The comment says so, and the HUP test pins it.
- **`cmd.WaitDelay = grace`** (same item), so a copier that will not finish cannot hold the wait past the stop. That changes what an escaped grandchild holding the pipe produces: the wait returns with the pipes closed and the attempt reports round 5's answer for a failed wait — exit 1, "its output could not be read in full" — instead of spending the reap grace and giving up. The three tests built on that shape now pin the new answer; the stuck-and-124 path keeps the injected coverage it has had since round 1, since a group that outlives SIGKILL cannot be made to order.
- **PGID reuse** (declined): the reasoning is in the sweep's comment — a pid is not recycled while it is a live group's id, so a group that answers is the one this attempt started, and once its last member is gone the sweep sends nothing; the window is microseconds against a sequential pid space, and closing it would mean probing process identity.
- **The package doc** names every subcommand the binary dispatches.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/...`, root Makefile audits, golangci-lint. All foreground. Two of the reworked tests first failed only under `-race` (a fork slower than the bound); they no longer depend on a clock at all — the wait delay ends them.

obra:
## Review round 8 disposition (07bd174de)

- **Medium, `timedOut` survives the finished-on-its-own path, so the runner still retries and can exit 124** — accepted; two reviewers found it independently. The attempt is decided on that path, keyed on whether the process exited before the helper sent its TERM, and the branch gets the test it lacked (bound expiring during the drain).
- **Medium, `-workfile` missing from the selection tables** — accepted, and the class closes the way #1266 closed it: a test reads `go help build` from the toolchain and fails on any value-taking or selection flag the tables do not name.
- **Medium, gate not wired** — declined for the sixth time; the split.

obra:
## Round 8 fixes

Head `e1d21b03b`, all in `e1d21b03b`.

- **`timedOut` left set on the decided path** (Medium, and mine from round 7). The runner checks that flag before the exit code, so a decided command was retried and could end as 124 with its own output already written. The decision now happens **before** anything is sent: when the bound expires, the drain gets the grace the stop would have taken, and if the wait returns in it the command exited on its own — no stop, no timeout, nothing to retry. Otherwise the stop proceeds, and a child that answers our SIGTERM by exiting 0 is not deciding for itself, which is the nuance the after-the-fact status check could not keep (round 7's version is gone, along with `procgroup.DiedOfSignal`).
- **The branch is tested where it happens** (same item): a 200ms bound against a command that exits while its grandchild holds the pipe for three seconds — the command's own status, no timeout, and one attempt through the runner. Mutation-proved by setting the flag before the window: both cases go red.
- **`-workfile`** (Medium): not on this toolchain. go1.27.0's `go help build` does not list it, and `go list -workfile=...` answers `flag provided but not defined`; adding it would forward a flag the command refuses. What is added is the class-closing test that will catch it — and `-godebug`, and the next one: it reads `go help build` and fails when a flag that takes a value is not consumed with its value, or when a flag this package names as selecting packages is not forwarded. Mutation-proved by removing `-overlay` from the selection table.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/...`, root Makefile audits, golangci-lint. All foreground.

obra:
## Review round 9 disposition (e1d21b03b)

CI green. RoboRev (Medium ×4, Low ×2):
- **Medium, the post-bound drain grants a still-running command up to `-grace` and ignores signals meanwhile** — accepted; round 8 traded one race for this one. The fix owns the exit edge: the helper copies the command's output through pipes it owns, so the wait returns the moment the process exits and draining is a separate step. At the bound, an exited process is the command's answer and a running one is stopped at once; the latch is selected on in every wait.
- **Medium, dangling value flags ignored and `-C` silently dropped** — accepted: both refused, as #1266 refuses them.
- **Medium, the completeness test reads `go help build` only** — accepted: `go help testflag` too, so a new value-taking test flag must be consumed.
- **Medium, gate not wired** — declined for the seventh time; the split.
- **Low, group semantics silently absent off linux/darwin; short-write message prints `<nil>`** — both accepted: the subcommand refuses to run where it cannot group, and the shortfall is reported in bytes.

obra:
## Round 9 fixes

Head `04154ea25`, all in `04154ea25`.

- **The attempt owns the copying** (Medium). Handing exec an `io.Writer` makes `cmd.Wait` wait for exec's copiers, so the wait ended when the output stopped arriving rather than when the process exited — which is what round 8's drain-grace was working around, at the cost of letting a running command overrun its bound and of not watching the latch while it waited. The attempt opens the pipes and copies them itself: `reaped` closing now means "the process is gone", the bound's arm stops the group at once, and the drain is a separate bounded wait. Every remaining wait — drain, reap, escalation — also selects on the latch.
- **Two tests changed shape with the design**: a command whose last bytes are in flight is now decided at its exit, so the case that pinned "the bound expires during the drain" pins that invariant instead (that scenario is no longer reachable); and the interrupt-while-the-output-is-held case now holds the command running, since one that has already exited is the other test. Both were racy under `-race` in their first form and are timing-free now.
- **`list-build-flags`** (Medium ×2): refuses `-C` as the shard runner does, and refuses a value flag with nothing after it; the completeness test reads `go help testflag` as well and fails when a value-taking test flag is not consumed — mutation-proved by removing `-run`.
- **No process groups, no run** (Low): `bounded-list` exits 2 with "needs a platform with process groups (linux, darwin)" rather than sweeping nothing; `procgroup.Supported` carries it and the build stays portable.
- **A short write with no error** (Low) reports how much was written, without a trailing nil.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/...`, root Makefile audits, golangci-lint. All foreground.

obra:
## Review round 10 disposition (04154ea25)

CI green. RoboRev, all accepted:
- **Medium, the timeout arm no longer checks `finishedFirst` while the signal arm does** — accepted; round 9's restructure lost it. Both arms consult it before deciding, and the injected-ordering test covers the timeout side.
- **Low, `list-build-flags` ignores the byte count on writes** — accepted: the same short-write rule as `forwardOutput`, ideally the same function.
- **Low, `-test.`-prefixed spellings not consumed** — accepted, with #1266's rule: the prefix is dropped only for names the test tables know.

obra:
## Round 10 fixes

Head `17e856f80`, all in `17e856f80`.

- **The timeout arm asks first** (Medium). Round 9 moved the exit edge and dropped the `finishedFirst` check that went with it, so a command that exited as its bound expired could be called a timeout and retried. It asks now, as the signal arm does. The test injects the ordering, and says why: a command that sleeps exactly as long as its bound loses that race on every machine here, so the end-to-end version would assert the opposite of what it claims.
- **`-test.` spellings** (Low): the prefix comes off only when what remains is a flag the test binary has, from `go help testflag`'s own names rather than from the tables here (which mix in build flags). `-test.run -race` forwards nothing — the `-race` is the regex — and `-test.race` forwards nothing either, because the binary has no such flag. Cases for all three.
- **Short writes** (Low): `list-build-flags` goes through the same rule as the command's output, which now names whose answer it is carrying — `list-build-flags: -race: wrote 0 of 6 bytes`.

Gates: toolchain gofmt, `go vet` host + `-tags evenerfuzz` + `GOOS=windows`, `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/...`, root Makefile audits, golangci-lint. All foreground.

obra:
## Review round 11 disposition (17e856f80)

CI green. RoboRev (Medium ×5, Low ×2):
- **Medium, the window between the finished-first check and the stop** — accepted: the stop reports whether the child was already reaped, and a reaped child finalizes through `completed`, so check and stop are one decision.
- **Medium, PID reuse in the survivor sweep** — declined, third time; the reasoning is in the sweep's own comment.
- **Medium, pipe read ends never closed when the drain times out** — accepted; real and a consequence of owning the pipes: goroutines and descriptors leaked, and a survivor could keep writing into a later attempt's stderr. The read ends close whenever the drain gives up or the attempt is cleaned up.
- **Medium, `Terminate`/`Kill` lack the `pgid <= 0` guard their siblings have** — accepted; `kill(0, …)` would signal the helper's own group.
- **Medium, gate not wired** — declined for the eighth time; the split.
- **Low, alias test asserts exit 0 where process groups are unsupported; drain timer reset on every signal** — both accepted.

obra:
## Round 11 fixes

Head `11a8d2777`. Every behavioural change below was mutation-proved on a copy of the tree; the mutation and what it broke are named with each item.

**1 (M) — the stop says whether it signalled anything.** `procgroup.StopWith` already refused to signal a child it was told had been reaped; it now returns that as a bool, and `Stop` with it. An attempt whose stop sent nothing finalizes through `completed(waited.err())` — the command's own status — instead of 124 or 128+signal, on both arms: the timeout arm's answer is now held in a local and applied only after the stop confirms it, and the interrupt arm keeps the latched signal (the run still stops, no further attempt) without letting it decide the status.

Two tests with the ordering injected, through the `groupStopper` seam that already existed for the group that outlives SIGKILL — it gained a `stop` field, so the whole process-group half of an attempt is injected in one place. The stub waits for the reap and reports that it sent nothing, which is exactly a reap landing after the check and before the signal. Mutation (call the stop, discard its answer, set `timedOut` unconditionally): the bound case goes 7 → 124 after two retries, the interrupt case 7 → 143.

**2 (M) — an attempt closes its pipes' read ends when it ends.** `outputPipes.closeReads()`, deferred from the moment the pipes exist, so it covers every cleanup path and not just the drain's give-up. Closing a pipe read end unblocks the read in flight on it, so the copier ends with the attempt rather than whenever the survivor does.

Test: the escapee trio, extended so the grandchild inherits the attempt's stderr as well and writes into it on a go-ahead the test gives *after* the first attempt has returned, while a second attempt is running against the same stderr. The write goes through a dup of the descriptor (a broken-pipe write on fd 2 itself is a fatal SIGPIPE, and the helper has to outlive the write to report it) and the outcome is recorded in a file. Both halves are live: with the defer removed the record says `wrote` instead of `refused:`, and with that assertion disabled the stderr assertion fires on `a survivor wrote this` sitting in the second attempt's diagnostics.

**3 (M) — Terminate and Kill check the group id.** One `signalGroup` helper is now the only place a group signal is sent and the only place the identifier is checked, so the four entry points cannot drift.

The test runs each call in a helper process in a process group of its own, and the helper surviving and exiting 0 is the assertion. The same check made in-process would, with the guard gone, TERM or KILL the test binary, the `go test` that started it, and the terminal or CI job they share — so `TestStopRefusesAPgidThatIsNotAGroup` is replaced by this one rather than added to: it covers the two stops it covered (mutation: drop `StopWith`'s own guard → all four `Stop`/`StopWith` cases fail, each having waited out the 30s grace) plus the two signals it did not (mutation: drop the `signalGroup` guard → `Terminate(0)` and `Kill(0)` report `signal: terminated` and `signal: killed`, contained in the helper's group).

**4 (L) — the alias case takes the platform's answer.** `dispatch_covtest_test.go` branches on `procgroup.Supported`: 2 and no output where there are no process groups, and both spellings still compared against each other, which is what the case is for. `GOOS=windows go vet` covers the compile.

**5 (L) — one deadline for the whole drain.** `time.NewTimer(grace)` outside the loop. Test: output that never arrives with signals arriving faster than the grace — with a grace started afresh each turn, `awaitDrain` never returns (10s tripwire), because the always-ready signal case outruns a timer that keeps being replaced.

`procgroup_other.go` keeps step: `Stop` there answers the same question and gained the already-reaped check it was missing.

Gates, all foreground: `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/procgroup/...`; `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`; `make lint` (9 gates). `origin/main` (`bcb8406d9`) was already merged in `6750cdecc`, so the merge was a no-op.

obra:
## Review round 12 disposition (11a8d2777)

CI green; three of four reviewers clean, one confirming every fix commit's concern is addressed in the final code. The remaining reviewer:
- **High, PID reuse in the survivor sweep** — declined, fourth time, and the alternative offered is considered: a sentinel process per attempt to pin the group id would add a process to every enumeration to close a window that is microseconds wide against a sequential PID space, on a group that is signalled only while it still answers. The reasoning stays in the sweep's comment.
- **Medium, gate not wired** — declined for the ninth time; the split.
- **Medium, a signal arriving during the final output forwarding is lost once `signal.Stop` runs** — accepted: the latch is consulted before every terminal return.

obra:
## Round 12 fixes

Head `ada3b3c4a`.

**M — a signal that lands while the output is handed over.** Writing the command's output is the last thing a run does, and `signal.Stop` drops whatever arrived the moment `boundedList` returns, so that window was watched by nobody. The latch is now polled before every terminal return, in `handOver` — the one place that writes the output and gives the run its answer, which also folds the four copies of the forward-or-fail line into one.

Test with the ordering injected: the stdout writer delivers the signal from its own `Write`, so it arrives during the forwarding and nowhere else. The run returns 130 and the command's output is still handed over in full. Mutation (drop the poll): exit code 0 for an interrupted run.

**One boundary the suite made me draw.** `latch.poll()` answers with a signal the attempt had already latched, so polling unconditionally reversed round 9's rule: `TestBoundedListDoesNotRetryAFailedCommandAfterAnInterrupt` went 3 → 143, which is the case where a command decided on its own and the interrupt only stopped the retry. So `handOver` returns the attempt's own answer whenever `result.interrupted != 0` — the attempt has accounted for the signal, either by keeping the command's status or by carrying 128+signal itself — and the poll can only change a run that had seen no signal at all. Mutation (drop that early return): that test fails as it did before the guard, which is what pins the boundary.

**PID reuse.** One sentence added to the sweep comment: closing the window completely would mean holding a sentinel process in the group for as long as the attempt lives, so the id cannot be recycled at all — a process per enumeration, and the gate enumerates every module, bought against a race nothing here has been able to produce. Nothing else changed.

Gates, all foreground: `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/procgroup/...`; `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`; `make lint` (9 gates). `origin/main` already merged.

obra:
## Review round 13 disposition (ada3b3c4a)

CI: the `web` job failed on this Go-only head; the failing test is being identified and the job rerun, since main's web job passed on the same base and #1308 tracks the skill guard's runner-load flake.

RoboRev: three of four reviewers clean. The remaining reviewer:
- **Medium, gate not wired** — declined for the tenth time; the split.
- **Medium, `StopWith` reports "signalled" even when `kill` returned `ESRCH`** — accepted: a signal that found no group is not a stop, so the attempt keeps the child's own status.
- **Medium, a signal arriving between the final poll and `signal.Stop` is lost** — accepted: the notification is stopped first and the channel drained as part of the final decision.
- **Low, copier goroutines can outlive an attempt into the next one** — accepted: each attempt joins its copiers before another starts, with a sink that can be released.

obra:
Rerun of the `web` job at ada3b3c4a passed; the earlier failure was `TestSkillComposerBrowser` timing out (#1308), not this PR. CI is green at this head; round 13 findings stand as dispositioned above.

obra:
## Round 13 fixes

Head `8435a38fd`. `origin/main` (`9441357d6`) merged.

**1 (M) — ESRCH is not a stop.** `StopWith` reports a stop only when a signal of its own reached the group; `signalGroup` now answers with what the kernel said. Two details the finding's shape made necessary:

- On that answer the stop waits out the caller's reap, bounded by the grace. The caller's next move is to read the child's own status, and `cmd.ProcessState` exists only after the wait — without this the attempt could finalize on a nil state and report 1 for a command that exited 0.
- ESRCH *after* a signal of ours was delivered is a group this stop emptied, so `delivered` gates it: only a group that was already gone when the first signal went out counts as "nothing stopped".

Other refusals come back as an error and the attempt prints them; EPERM is the one that matters, since it means the group is there and may still be running. Tests: `TestStopWithReportsAGroupThatWasGoneBeforeTheSignal` aims at group id `1<<30` — above any pid either kernel can hand out (darwin's default ceiling is 99999, Linux's `pid_max` ceiling 4194304), so the signal has nowhere to go by construction, which is what makes it safe to send; measured on darwin, `kill(-1<<30, sig)` answers ESRCH, not EINVAL. Mutations: report ESRCH as a stop → `stopped = true` for a group that was not there; drop the diagnostic → `TestBoundedAttemptSaysWhenTheGroupRefusedTheStop` sees an empty stderr. The attempt side needed no new case: the injected `reapedBeforeTheStop` stub is exactly this answer, and round 11's two tests already pin "command status, no retry" on it — its comment now says it stands in for both orderings.

**2 (M) — the last look stops delivery first.** `signalLatch.settle()`: `signal.Stop` on the run's channel, then a non-blocking drain into the latch, then the answer. `handOver` calls it instead of `poll`, and `boundedList`'s deferred `Stop` is left a no-op (stopping twice is allowed). Test with the ordering injected: a poll that sees nothing, a signal delivered after it, and `settle` taking it. Mutation (settle without the drain): `settle = signal 0`.

**3 (L) — an attempt's copiers end with it.** `endCopying` closes the read ends, joins the copiers bounded by the grace, and on timeout detaches the writer (`serialWriter.release`, which takes no lock — the copier it is releasing from may be holding one inside a write that has not returned). Unit test with both cases; mutations: drop the release → the bytes land in the caller's stderr; release without joining → a joined attempt's writer is detached and its last words are lost.

One thing I could not produce, so you know what the test does and does not say: with the read ends closed, `io.Copy` performs at most one more write (its next read fails and the loop ends), and that write is already inside the sink when the release happens — so I could not build a case where a later attempt's diagnostics actually receive a survivor copier's bytes. The join is also the second one on every path, since `awaitDrain` already waits one grace for the same channel, so only a copier stuck between one grace and two is distinguishable at the attempt level — a clock race I did not want to write as a test. What is pinned is `endCopying` itself, deterministically, on both sides.

`modulelint`'s `go procgroup.Stop(...)` now discards the answer explicitly (errcheck): the wave is ending, and that run's writers belong to the loop still using them.

Gates, all foreground: `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/procgroup/...` (re-run on the final head); `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`; `make lint` (9 gates).

obra:
## Review round 14 disposition (8435a38fd)

CI green. RoboRev (High ×1, Medium ×3, Low ×1):
- **High, `pgid == 1` passes the guard and becomes `kill(-1)`** — accepted; the guard is `pgid <= 1` everywhere, with the case tested.
- **Medium, `serialWriter` holds its lock across the caller's write** — declined: every caller hands the helper a file for stderr, a write that never returns there is the host, not the command, and an asynchronous buffer would be a place for diagnostics to be lost. The comment on the writer states that contract.
- **Medium, a refused signal followed by a natural exit in the grace still counts as stopped** — accepted; only a delivered signal counts, as round 13 already does for ESRCH.
- **Medium, a zombie-only group on a Linux host that does not reap orphans reads as stuck** — accepted as a measurement where it matters: the darwin numbers in the `Exists` comment say nothing about the runner. A test that creates the reparented orphan and expects the group to clear within the grace runs in CI; if Linux disagrees, `Exists` learns to tell a zombie-only group apart from a live one on Linux.
- **Low, `list-build-flags` re-derives the flag tables `shardplan.go` owns** — accepted: with #1266 on main this is #1285's fold, and it happens here.

obra:
## Round 14 fixes

Head `ebbc39d79`. `origin/main` already merged.

**1 (HIGH) — pgid 1 is the whole host.** `kill(-1, sig)` is every process this user may signal, so the floor is now 2, in `signalGroup`, which is the one place that sends and now also the one place `Exists` asks its question through. None of 0, 1 or a negative can be a group this package started: `Start` makes the child a group leader, so a group id is a pid the kernel handed out.

The 1 case is in the helper-process guard table (all four entry points, a helper that survives and exits 0), **and I did not mutation-prove it**: removing the guard and running that case broadcasts SIGTERM and then SIGKILL to every process this user owns on the machine running the tests, and no process group contains that. What is proved instead, safely, is the guard itself — `TestSignalGroupRefusesTheNumbersThatAreNotGroups` asks with signal 0, which delivers nothing even when it reaches everything, and with the floor back at 0 it reports `signalGroup(1, 0) = <nil>` and `Exists(1) = true`. Everything that sends goes through that one function, so that is the guard for all of them.

**2 (M) — only a delivered signal makes a stop.** A child that exits while every signal the stop made was refused exited on its own, and a reported stop turns its status into 128+signal upstream. `delivered` now decides the answer on both escalation paths, and the final KILL goes through the same sender so a KILL that lands still counts. The sending is injected — `stopWith` takes it, `StopWith` passes `signalGroup` — since producing EPERM for real means signalling a group nobody here owns. Test: the sender answers EPERM and the child reaps during the grace → `(false, EPERM)`, one TERM sent. Mutation (`return true`): "reported a stop it was refused permission to make".

**3 (M) — the zombie-only group, measured rather than asserted.** `TestExistsAnswersNoOnceAZombieOnlyGroupIsAdopted` builds the reparented-orphan shape — a leader that starts a grandchild in its own group, waits until the grandchild has exited, and exits without reaping it — then polls `Exists` from the moment the leader is reaped and fails if the group is still answering after 5s. **The CI answer: it passes on ubuntu.** `race-root` and `race-modules / nonagent` both went green on `12ef5c43c`, so the zombie-only group clears well inside the grace on Linux too and no Linux-only zombie check is needed; `Exists` keeps the answer it has. The CI log carries no number (the gate does not run `-v`), so what Linux gives is pass/fail, not a duration. On darwin I took 20 runs and every one found the group already empty at the first poll, 1.4-2.9us -- the comment's old "3-12ms" was one platform's and stale, and now says both. That last edit is the second commit, `ebbc39d79`, which is a comment in a test file: gofmt, `go vet`, and `-race -count=3` on the package, not the full list.

**4 (declined) — the serialWriter lock.** One sentence added: the lock is held across the write on purpose, because stderr is a file for every caller this has (a terminal, a gate's log), a write to one that does not return is the host having stopped rather than the command, and an asynchronous buffer between them would buy nothing and lose the ordering of what it buffered.

**5 (L) — one flag language, closes #1285.** `list-build-flags` now normalises through `goFlag`, walks through `walkFlags`, and refuses `-C` with `errUnsupportedC` — the shard runner's own words, which is why the existing message assertions did not move. `goTestValueFlags`, `testBinaryFlags`, `isKnownTestFlag` and `normalisedFlag` are gone (-143 lines); "is the next word this flag's value" is now `valueIsNextArgument`, one function over the shard runner's three value tables, used by `walkFlags` and by the `go help` guard's `consumed`.

The prefix rule differed between the two copies, so I measured before adopting the shard runner's: on go1.27.0 every name where they disagreed is one `go test` refuses at the door — `-test.vet`, `-test.args`, `-test.exec`, `-test.o`, `-test.c` and `-test.n` all answer `flag provided but not defined`, and `-test.race` with them. What stays in `list-build-flags` is the judgement no table in `shardplan.go` encodes: which flags change the set of packages that exist. Mutation (drop one table from `valueIsNextArgument`): the `go help` guard fails on `-covermode`, `-coverpkg`, `-bench` and the rest.

Gates, all foreground: `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/procgroup/...`; `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`; `make lint` (9 gates).

obra:
## Review round 15 disposition (ebbc39d79)

CI green; two of four reviewers clean.
- **Medium, gate not wired** — declined, eleventh time; the split.
- **Medium, a leader that exited on its own while a descendant kept the group answering is classified as timed out** — accepted; "delivered" says a signal reached the group, not that the leader was running. After the stop, the leader's own reaped status decides: an exit that is not by the helper's signal is the command's answer, and the survivor sweep is cleanup.
- **Medium, PID reuse** — declined, fifth time; the sweep comment has the reasoning and the rejected sentinel.
- **Low, the zombie-group measurement fails on a host that does not reap orphans** — accepted: it is a measurement, so a host still answering at the deadline skips with the observation logged rather than failing the default gate.

obra:
## Round 15 fixes

Head `cfdea25a4`. `origin/main` merged.

**1 (M) — the leader's own status decides the attempt.** Once the leader is reaped, `endedOnItsOwnTerms(cmd.ProcessState)` asks how it ended: an exit status is the command's answer, and so is a death by a signal this attempt does not send (an OOM-killed compiler said something about the run, and it is not a timeout). Only a leader killed by our TERM or KILL, or one that could not be reaped at all, is a timeout. `procgroup.DiedOfSignal` does the wait-status decoding, next to `ExitCode`, which now goes through it; the non-unix stub answers no. The sweep is unchanged: it runs either way, as cleanup.

Test with the ordering injected — the stub waits for the reap and reports the signal delivered, which is what a group populated by a descendant does — over a fixture whose leader outlives the bound, leaves a TERM-ignoring child in its group, and then exits 0. At the attempt: exit 0, `timedOut` false, the survivor swept, the group gone. Through the runner: exit 0, one attempt. Mutation (drop the check): exit 1 at the attempt, and 124 after three attempts through the runner — the finding's bug exactly.

**The rule is the bound's, not the interrupt's — a second commit narrows it.** Applied to both paths it broke `TestBoundedAttemptForwardsTheSignalItWasSentNotATermInstead`, and that test is right: its child traps SIGHUP and exits 0, and the established answer is 129. A command that exited because this helper passed the operator's signal on to it has not produced an answer to prefer over that signal — a `go list` that trapped it and exited 0 wrote a list that stops wherever the signal found it, and 0 would tell the gate the list is whole. The two cases are indistinguishable after the fact (a wait status does not say whether the leader ever received the signal), so the new rule applies only when nothing was latched, which is the case the finding describes.

One test expectation moved with the rule, deliberately: round 13's `TestBoundedAttemptSaysWhenTheGroupRefusedTheStop` asserted `timedOut` on a stop whose every signal was refused (EPERM) while the command exited on its own. That is no longer a timeout, and should not be — it now asserts the refusal line and the command's own status.

**2 (L) — the zombie-group test skips rather than fails by default.** Whether an orphan is adopted and reaped is the host's business, so at the deadline it reports what it saw and skips; `EVENER_ZOMBIE_GROUP_STRICT=1` makes the same observation a failure. Both branches exercised against a stand-in for a host that never reaps: skip by default, `t.Fatalf` with the observation under the env var. The ubuntu runner passes it today, so CI stays a real measurement.

Gates, all foreground: `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/procgroup/...`; `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`; `make lint` (9 gates).

obra:
## Review round 16 disposition (cfdea25a4)

CI green; two of four reviewers clean.
- **High, a copier blocked on stderr holds the writer's lock, and the cleanup diagnostics then block on it before the deferred release can run** — accepted this time; the earlier declines rested on stderr being a file, but the ordering is wrong regardless: the release must come before the cleanup diagnostics, and once released the diagnostics must not wait on a lock a wedged copier holds.
- **Medium, PID reuse** — declined, sixth time.
- **Low, `cmd.ProcessState` read while `Wait` may still be writing it when every signal was refused** — accepted; the status is read only after the reap is observed, otherwise the attempt is reported without it.

obra:
## Round 16 fixes

Head `6ae2ce92a`. `origin/main` merged.

**1 (HIGH) — the copying ends before the cleanup says anything.** You were right that the ordering is the bug whatever the lock finding's merits were. `endCopying` now owns the whole end of the copying — the drain wait, the read ends, the join, the release — and returns what the drain came to, and every path calls it through `finishCopying`, a once-closure, before it writes a word of cleanup (the deferred call stays as a backstop for a path that forgets; it runs once, so the two cannot fight). That moved two other writes out of the wedged window as well: `reapOrGiveUp`'s give-up line and `stopSurvivors`' sweep line, both of which used to be written before the drain was ended.

A released `serialWriter` now goes round the lock instead of waiting for it — what holds that lock is precisely the copier that cannot return — and the comment says what that costs: two writers on one stderr can interleave, and in the only case this happens the host has stopped taking writes, so a line cut in half is a better failure than a helper that stops without a word. The round-13 semantics (a released writer swallowing the bytes) are gone with it; they were the wrong half of the trade, and as I reported then they were never reachable anyway.

Test: a sink that holds the one write carrying the command's stderr and takes every other. The attempt returns inside its graces with `could not be read in full` in that sink and exit 1. Mutations: put the release back after the diagnostics → "the attempt never returned"; make a released writer take the lock → the same hang, plus the `endCopying` unit test's "a released writer waited for the lock a wedged copier holds".

**2 (L) — the status is read only once the wait has it, and -race did catch it.** On the path where the stop sent nothing, `finishedFirst` decides: with the reap in, the command's own answer; without it, the attempt falls through to the cleanup, which waits for the reap once more and gives up in the one way that reads no status at all. Test: a stub that answers "nothing stopped" while the command is still running, exit 7 expected.

The measurement you asked for: with the guard removed, `go test -race` reports `WARNING: DATA RACE` twice on that test — the read of `cmd.ProcessState` from the attempt against `os/exec.(*Cmd).Wait` at `exec.go:948` writing it, from the wait goroutine this file starts. So it was a race-detector failure, and it is gone. The reason it was reachable at all is the ESRCH path from round 14: `StopWith` waits a grace for a reap that may not come and then reports that nothing was stopped.

**3 — PID reuse: no change.**

Gates, all foreground: `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/procgroup/...`; `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`; `make lint` (9 gates).

obra:
## Review round 17 disposition (6ae2ce92a)

CI green; two of four reviewers clean.
- **High, the final output handoff writes to stdout with no deadline, so a full pipe can hang the helper after its bound has passed** — accepted: the bound covers everything the helper itself does, so the handoff gets the grace and a diagnostic on failure.
- **Medium, unbounded in-memory capture of the command's stdout** — accepted as a cap with a clear failure; a package list is small and anything past the cap is not one.
- **Medium, PID reuse** — declined, seventh time.
- **Medium, gate not wired** — declined, twelfth time; the split.
- **Low, a trailing bare `-args` is rejected instead of ending the walk** — accepted.

obra:
## Round 17 fixes

Head `2890c4819`. `origin/main` merged.

**1 (HIGH) — the hand-over is bounded.** `forwardOutput` writes from a goroutine and waits on the grace; a hand-over that does not finish is reported and the run fails with 1. Two things worth stating: the diagnostic goes to stderr, which is a destination of its own and is not what stopped taking the answer (the released path does not come into it — no attempt is running by then, so no `serialWriter` stands between this and the caller's stderr); and the goroutine left behind is the price, because nothing can take a blocked write off a descriptor, and this is the run's last act either way. `list-build-flags` gets the same bound through a shared `defaultGrace`, which is now also where the `-grace` flag's default comes from.

Test: a stdout sink that takes the write and does not return. The run ends with 1 inside its graces, stderr says the answer is not being read any more, and the write was made. Mutation (wait on the write with no deadline): the test binary times out at 90s.

**2 (M) — one attempt holds a list's worth.** `maxCapturedOutput = 64 << 20`, named with the reason: what this carries is a package list, and this repository's whole `go list ./...` is tens of kilobytes. Past the cap the bytes are read and dropped rather than refused — a copier told its write failed stops reading, and a command whose stdout stops being read blocks on a full pipe, which would turn "wrote too much" into "hangs". `overflowErr` then joins the drain and copy failures on the one path that already says the output is not whole, so the attempt exits 1 with a diagnostic.

Test: the cap is lowered to 4 KiB for the test (and its production value asserted in the same test before lowering) against a command that writes 100 KB, so the test writes kilobytes instead of tens of megabytes; what is under test is the cap, not the number. Mutation (no cap): exit 0 with the whole output held.

**3 (L) — a trailing `-args`.** It moves from `testRefusedValueFlags` to `testRefusedBareFlags`, which is what it always was: `-args` takes no value of its own, everything after it is the test binary's argument list, and `go test -race -args` is a valid end of the flags. No special case in `walkFlags` was needed once the table said the truth, and both readers behave: `["-race", "-args"]` gives `["-race"]` from `list-build-flags`, and `parseFlags` gives the refusal for `-args` itself rather than a complaint that it was given nothing. Mutation (move it back): both fail with "-args was given with nothing after it".

**4 — PID reuse and gate wiring: no change.**

Gates, all foreground: `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/procgroup/...`; `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`; `make lint` (9 gates).

obra:
## Review round 18 disposition (2890c4819)

CI green. RoboRev (High ×2, Medium ×3, Low ×1):
- **High, two diagnostics still go through the copier's lock before `finishCopying`** — accepted; round 16's invariant applies to every write, and the interrupt message does not need to precede the stop.
- **High, a command that traps the timeout's SIGTERM and exits 0 is taken as its own answer** — accepted; the interrupt arm already refuses that (round 15), and the timeout arm must too: a leader reaped only after the stop delivered its signal is a timeout whatever status it chose.
- **Medium, the handoff's timeout diagnostic can block on a stalled stderr** — accepted: diagnostics at the end are best-effort and bounded like the handoff.
- **Medium, PID reuse** — declined, eighth time.
- **Medium, gate not wired** — declined, thirteenth time; the split.
- **Low, one `bytes.Buffer` shared by a helper's stdout and stderr in two tests** — accepted.

obra:
## Round 18 fixes

Head `6065ff4b6`. `origin/main` merged.

**1 (HIGH) — nothing goes through `guarded` before the copying has ended.** The interrupt's "stopping" line and the refused stop's line are recorded in a `pending` slice and written by `finishCopying` the moment the copying ends, so the invariant is structural rather than a rule to remember: the only place that writes those lines is the place that has just ended the copying. The operator now hears what was done rather than what is about to be; that is the cost of the ordering, and the alternative was a line that waits on the lock a wedged copier holds.

Every `Fprint*` through `guarded`, with what makes each safe:

- `boundedlist.go:80` and `:87` — the openOutputPipes failure and the `procgroup.Start` failure, both before `copyInto`, so no copier exists.
- `:108` — the pending lines, inside `finishCopying`, immediately after `endCopying` returned.
- `:147` — "its output could not be read in full", inside `completed`, after its `drainErr := finishCopying()`.
- `:151` and `:273` — `stopSurvivors`, in `completed` after that same call, and in the tail after `_ = finishCopying()`.
- `:235` — `reapOrGiveUp`, in the tail after `_ = finishCopying()`.

Test: a stalled stderr sink with the copier wedged in it, then an interrupt. The group is stopped, the attempt returns 143 inside its graces, and the line lands once the copying has ended. Mutation (write it where it was): "the attempt never returned".

**2 (HIGH) — a delivered stop is a timeout, whatever the command exited with.** `endedOnItsOwnTerms` is gone; the fact the answer keys on is round 14's `delivered`. The trade is stated in the code: the other ordering — leader already gone, signal delivered to a descendant — is retried although its work was done, because nothing in a wait status tells the two apart and this is the side that cannot report a partial list as a whole one. Round 15's two stub tests are replaced by one that pins the new rule on the same fixture.

The other half of the ordering is now used too: a stop that delivered **nothing** did not end the command, so its exit is its own answer whether it arrived before the stop or while the cleanup was waiting for the reap — which also removes round 16's early `finishedFirst` branch, since the read now happens after the reap either way. The give-up diagnostic on that path no longer blames a SIGKILL nobody sent. One test's stub moved with it: the refused-stop case answered `(true, EPERM)`, which the real `StopWith` cannot produce — a refused signal is not a delivered one — and now answers `(false, EPERM)`.

Test: a child that traps the bound's TERM and exits 0 → 124, retried per `-attempts`, with the TERM marker proving the stop was delivered. Mutation (decide by status again): exit 0.

**3 (M) — the end-of-run words are bounded too.** `sayWithin` writes from a goroutine and gives up after the grace, for the three diagnostics that follow a hand-over; a gate's answer and its log are often the same file, so the destination that stopped taking the answer has stopped taking the report about it. Test: one stalled writer as both stdout and stderr → the run still ends, with 1. Mutation (write it synchronously): the test binary times out.

**4 (L) — one buffer each in the two re-exec tests.** Done, and worth recording: it was not a race. `os/exec` hands the child a single descriptor when `Stdout` and `Stderr` are the same writer (`childStderr` → `interfaceEqual`, exec.go:579-584 on go1.27.0), so one copier existed, not two. Two buffers say that without the reader having to go and check.

Gates, all foreground: `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/procgroup/...`; `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`; `make lint` (9 gates).

obra:
## Review round 19 disposition (6065ff4b6)

CI green; two of four reviewers clean.
- **High ×2, the gate does not use either helper** — declined, fourteenth and fifteenth time: this PR is the helper alone by the split of #1265, whose gate wiring lands once this merges.
- **Medium, SIGQUIT is not intercepted, so it ends the wrapper and leaks the command** — accepted; it goes through the same latch and cleanup.
- **Medium, some cleanup diagnostics still write to stderr without a bound** — accepted: every diagnostic after copying has ended goes through the bounded writer.
- **Low, a failed final handoff overwrites an interrupted run's `128+signal` with 1; the completeness test's predicate does not mirror the runtime one** — both accepted.

obra:
## Round 19 fixes

Head `6a2f4c583`. `origin/main` merged.

**1 (M) — SIGQUIT is a stop.** It joins `stopSignals`, the list `signal.Notify` is given, so it is latched, forwarded to the command's group and reported as 131 like the other three. The comment says what catching it costs: the Go runtime's stack dump on QUIT, which is a debugging aid for this helper rather than for the `go list` the operator is trying to stop. Tests: the list itself, and an attempt whose child answers QUIT and ignores TERM, so its own stdout says which signal reached it — `interrupted = SIGQUIT`, exit 131, group gone. Mutation (drop it from the list): the list test names it.

**2 (M) — every word this helper says is bounded.** All of them now go through `sayWithin`, which is what the hand-over already used. The reason is the released path: once a copier is wedged, the writer writes straight to the caller's stderr, and a caller that has stopped taking the answer has stopped taking the words about it too — so a cleanup line could hold an attempt exactly as the wedged copier it was reporting would.

The audit, every write in the file and what it is now:

- `:80`, `:87` — the pipe-open and `procgroup.Start` failures: `sayWithin` (they run before any copier exists, but the caller's stderr can be stalled from the start).
- `:108` — the pending flush inside `finishCopying`: `sayWithin`.
- `:147` — "its output could not be read in full", inside `completed`: `sayWithin`.
- `:329`, `:335` — `reapOrGiveUp`'s two give-up lines: `sayWithin`.
- `:716`, `:725` — `stopSurvivors`' sweep and its give-up: `sayWithin`.
- `:790` — the usage text; `:800`, `:805`, `:810` — the platform and flag refusals; `:816`, `:835`, `:838` — "not starting attempt N", the retry announcement and the final timeout: all `sayWithin`.
- What is left unbounded: the `flag` package's own parse errors, which it writes to the output it was given. That is the one writer this file does not own.

Test: a stderr that takes no write at all, with the copier wedged in it, so the release happens and the released path meets the same stalled destination. The attempt returns with 1 inside its graces. Mutation (write the "could not be read in full" line synchronously): "the attempt never returned".

**3 (L) — a failed hand-over does not rewrite an interrupt.** 128+the signal says why there was an answer to lose; 1 says only that there was one. Test: an interrupted run whose stdout writer takes one byte of seven — exit 143, with the short write still named on stderr. Mutation: exit 1.

**4 (L) — the guard asks what the walk asks.** `consumed` is `valueIsNextArgument`, passed rather than wrapped (gocritic caught the wrapper; that is the second commit). Mutation, the one you named in its other direction — take `-tags` out of `buildValueFlags`, leaving it only in `packageSelectionValueFlags`: the guard now fails with "-tags takes a value and no table consumes it", where the old two-term predicate passed while the walk read the word after `-tags` as a flag of its own.

Gates, all foreground: `go test -race -count=3 ./cmd/evener-dev/... ./internal/devtool/procgroup/...`; `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`; `make lint` (9 gates). The race suites ran on `c970fbb60`; the second commit is a test-only wrapper removal, and it got `go vet` and `-race -count=3` over the flag tests.

obra:
## Review round 20 disposition (6a2f4c583)

CI green; two of four reviewers clean, one having traced every state combination without finding a wrong status, a retried decided command, a partial list reported whole, or a deadlock.
- **High, the real-signal test matrix includes pgid 1** — accepted: a guard regression would make the test itself the broadcast. Only signal 0 proves the floor for 1.
- **Medium, gate not wired** — declined, sixteenth time; the split.
- **Medium, a command that exited but is not yet reaped when the stop's signal lands reads as stopped** — declined with the reasoning written where the decision lives: the wait returns as soon as the child exits, so that window is the microseconds between the kernel marking the exit and `wait4` returning, and its cost is one extra enumeration, never a wrong list.
- **Medium, PID reuse** — declined, ninth time.

obra:
Closed on Jesse's ruling (2026-09-14): #1265 was overbuilding, and this half of it goes with it. Replaced by a ten-line fix: curl --retry on the golangci-lint download and a plain timeout around the gate's go list; the Metro bundle gate (#1245) gets a plain CI timeout instead of the process-lifecycle helper.

PR #1287, opened 2026-09-14
feat(hub): deploy and version auto-match (component 04b of multi-host)

obra:
Component 04b of the multi-host design (spec: docs/superpowers/specs/2026-09-14-multi-host-04-ssh-connection-manager.md), stacked on #1283 (04a). Adds deploy, version auto-match, and host hub restart to `cmd/evener-hub/internal/sshconn`.

- **deploy.go** — cross-compiles the controller's own tree for the host's GOOS/GOARCH through an injectable `BuildBinary` seam, stamping this process's buildinfo (GitSHA/GitDirty/BuildTime/Channel) so the deployed binary's `launch-check` version equals `buildinfo.Version()`. Pushes over an atomic ssh `cat > tmp && chmod +x && mv`, fed on the Runner stdin so `argv[0]` stays `ssh`. A failed or interrupted push never leaves a truncated `evener`.
- **version.go** — on attach, a host build differing from the controller's is redeployed and its hub restarted before attaching (new `deploying`/`restarting` states). Equal versions attach directly with no deploy.
- **Restart** — supervisor-aware (launchd `kickstart -k`, systemd/systemd-user `restart`) with the bare-process fallback from docs/evener-hub-remote-operations.md: pid by listening port (`lsof -ti`), argv recovery (`ps -ww`), log recovery (`lsof`), SIGTERM, wait for the port to clear, detached nohup relaunch, then an `/api/health` probe. Never starts a second hub while the old one holds `hub.lock`.
- **errors.go** — `ErrDeploy` and `ErrRestart` sentinels.

Gates: gofmt clean, `go build ./...`, `go vet`, sshconn tests, and the full `go test ./cmd/evener-hub/` (146s) all pass. Default tests never run ssh, `go build`, or the network: the fake Runner and `BuildBinary` seam cover it, and the live test skips unless both `EVENER_SSH_E2E=1` and `EVENER_SSH_E2E_HOST` are set (and under `-short`).

Two things worth flagging for review:
1. The deploy path cross-compiles from source, so it needs a Go toolchain and the source tree on the controller. A *released* hub has neither; the spec's installer fallback is not implemented in this PR. Bootstrapping a host with no evener at all therefore still needs a manual first push or the installer.
2. The live E2E test is not run yet: the only reachable hosts are a live hub and a host whose legacy state would need migrating.

## Live validation (added after the initial push)

[...]

PR #1290, opened 2026-09-14
feat(hub): remote hub source read path and registration (multi-host 05a)

obra:
Part 05a of component 05 (remote hub source), per `docs/superpowers/specs/2026-09-14-multi-host-05-remote-hub-source.md`. Stacks on #1287 (04b), which stacks on #1283 (04a).

## What this adds

Exposes a remote `evener hub` as one more `appsource.Source` on the controller, so the existing thread/turn RPC handlers reach a session running on another host without knowing about the host.

- `appsource.RemoteHubSource`: `ID`, `ListThreads`, `ReadThread`, `ListTurns`, `ListModels`.
- Ref translation between the controller's `host:<thread>` namespace and the remote hub's `local:<thread>` namespace, both directions.
- `ListThreads` remaps a `SourceIDs` filter from controller host names to the remote `local`.
- Registration: `newHubSourceRegistry` adds one source per configured `[[hosts]]` entry; the empty-ref default stays `local`.
- Production wiring in `main.go`: a `sshconn.Manager` whose `Ensure` supplies the client, closed with the process.

## What is deliberately staged (later parts)

Subscription fan-out (05b), turn/thread lifecycle mutations and mutation-unknown mapping (05c), and the capability probe (05d). Those interface methods exist but return a loud typed `InternalError` until their part lands.

## Design notes

- The client is the `sshconn` channel's long-lived `appwire.Client`; `RemoteHubSource` never caches it, so a component-04 reconnect is picked up automatically.
- Errors mirror `LocalDaemonSource`: transport failures (EOS/reset/closed/timeout) become `SessionUnavailable` naming the host; semantic `WireError` responses are preserved.
- A remote ref whose source is not `local` is refused rather than passed through: the controller's `SourceID:ThreadID` grammar cannot represent a nested remote hub without collision.
- `Thread.Evener.InstanceID` is left byte-for-byte untouched (opaque precondition token).
- With no `[[hosts]]`, behavior is exactly as before.

## Gates (run on this branch)

- `gofmt -l` — clean
- `go build ./...` — pass
- `go vet ./cmd/evener-hub/...` — pass
- `go test ./cmd/evener-hub/... -count=1` — pass (no host, network, or `ssh` required)

Tests use an in-memory `net.Pipe`-backed `StreamTransport` with a scripted responder; one test asserts the four mapped wire methods are in `CatalogMethodNames(ScopeHub)`.


obra:
Round 1 addressed (`3c1c25a9c`). 8 findings — 2 high, 3 medium, 3 low; all fixed, none contested.

- **High — `RelayOnThreadRead` not implemented; all remote reads fail after a successful read.** Confirmed at `remote_hub_source.go` (no method → `relayOnThreadRead` defaulted to `true` → `startRelay` → staged `SubscribeThread` → `remote hub source: SubscribeThread is not implemented yet`, discarding the read). Added `func (s *RemoteHubSource) RelayOnThreadRead() bool { return false }`, to be deleted when 05b wires a real subscription. Pinned by `TestRemoteHubSourceDoesNotRelayPlainThreadRead` (`cmd/evener-hub`) plus the end-to-end `TestHubRPCThreadReadServesRemoteHubSource`, which drives a real hub RPC `thread/read` against a scripted remote over an in-memory pipe.
- **High — remote paging cursors lack source identity.** Confirmed: the source was neither an `ItemCandidateSource` nor an `ItemReadCandidateSource`, so the legacy packing path hit `packedOlderCursor` with a zero identity and returned `legacy transcript item source cannot page without cursor identity` for any page carrying a cursor. `RemoteHubSource` now implements both seams (`remote_hub_source.go`): it mints a controller-owned `CursorIdentity` (ThreadRef = the controller ref, fresh incarnation, projection version 1), retains the remote hub's opaque cursor behind it in a bounded 64-entry cache, and rebases it onto the caller's boundary for continuations; an evicted continuation degrades to the typed `transcriptItemCursorStale` error, matching the bounded local-daemon snapshot. Pinned by `TestRemoteHubSourceListItemCandidatesPaginates` / `TestRemoteHubSourceItemCandidatesFromReadPaginates`, and end-to-end by the `ThreadTurnsList` half of `TestHubRPCThreadReadServesRemoteHubSource`. Verified pre-fix: that end-to-end test fails with the exact packing error above (checked by running it against the parent revision of the source file).
- **Medium — filter bypass: non-matching `SourceIDs` forwarded unfiltered.** Confirmed: `remapRemoteSourceIDs` returned an empty slice for `["other"]`, which `sourceIds,omitempty` serialized as absent. `ListThreads` now returns an empty response with no RPC when `SourceIDs` is non-empty and does not include this source. Pinned by `TestRemoteHubSourceListThreadsForeignFilterReturnsEmptyWithoutCall`.
- **Medium — unfiltered lists abort on nested-host topologies.** Confirmed: an empty `SourceIDs` was forwarded unfiltered, and a nested remote hub's non-`local` refs made `fromRemoteThread` fail the entire response. An unfiltered controller list now sends `SourceIDs: [local]` (the only representable namespace); the filtered case still goes through `remapRemoteSourceIDs`. Pinned by `TestRemoteHubSourceListThreadsUnfilteredScopesRemoteToLocal`.
- **Medium — `roots` accepted but never used.** Addressed by documentation, not behaviour: `roots` is retained and its field comment now states it is staged for 05b/05c/05d, matching `hubcore.HostConfig`'s "EvenerPath and Roots are advisory inputs to components 04/05" contract. There is no specification yet for what root scoping means on list/read, so enforcing one here would invent semantics outside this component.
- **Low — bare `eof` substring overbroad.** Confirmed. `remoteHubTransportText` now matches `eof` only as a standalone token (`containsWord`), so an application message such as `eoffice` is no longer reclassified as `SessionUnavailable`. Pinned by `TestRemoteHubTransportTextRequiresTokenBoundary`.
- **Low — `DeadlineExceeded` misreported as `SessionUnavailable`.** Confirmed for request errors. `mapCallError` now returns `context.Canceled` and `context.DeadlineExceeded` unchanged (matching `localDaemonCallError`), and the attach step got its own `mapConnectError` that keeps the `localDaemonDialError` classification. Pinned by `TestRemoteHubSourceCallErrorKeepsRequestDeadlineRaw`.
- **Low — test redirects `os.Stderr` without cleanup; pipe read end never closed.** Fixed with a `t.Cleanup` that restores `os.Stderr` and closes the read end on every exit path (`remote_source_registration_test.go`).

Note on high #2: `internal/appitempaging` exposes no way to read a foreign cursor's identity, so the remote cursor is retained server-side rather than embedded in the controller cursor. That keeps the change inside this PR's file set.

Verification:
- `go build ./...` → ok
- `go test ./cmd/evener-hub/ -race -count=3 -timeout 30m` → ok (796s; the package's 122s non-race runtime makes `-race -count=3` exceed go's default 10m package timeout, so the timeout was raised — no test failed)
- `go test ./cmd/evener-hub/... -count=1` → ok (13 packages)
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (1 module, 0 findings)
- No fuzz targets added or modified.

obra:
Round two addressed (`f3b24dd`). 3 Medium + 1 Low.

- **Medium — size-truncated complete page mints a cursor that can never be continued.** Confirmed. `recordRemoteItemPage` evicted the per-thread entry on a complete page (`native == ""`) while still returning an `Identity`, and `ListItemCandidates` rejected every continuation with `!ok || state.native == ""`. When the hub's size packer (`packItemCandidates`) dropped the oldest selected items it minted a real `OlderCursor` from that identity, so the cursor the client received was guaranteed to fail `thread/turns/list` with a stale error. Fixed by retaining the complete page's candidates and re-serving the continuation locally before the caller's boundary (the local-daemon source's complete-snapshot behavior), so the remote hub is never asked for a page it already said does not exist. Pinned by `TestRemoteHubSourceCompletePageContinuationServedLocally` (source level) and end-to-end `TestHubRPCThreadReadPagesSizeTruncatedCompleteRemotePage`, which drives a >1 MiB complete remote page through `thread/read` and pages the dropped item via `thread/turns/list`; both fail pre-fix with `transcript item cursor is stale`.

- **Medium — `Subscribe=true` fails remote reads and can orphan remote subscriptions.** Confirmed, both halves. `ReadThread` forwarded `Subscribe`/`ReplaceSubscription` to the remote hub, which registers a persistent subscription the controller can never manage or retire. Separately, the web client's `subscribe: true` first-hydration read still ran `params.Subscribe || relayOnThreadRead(source)` → `startRelay` → staged `SubscribeThread` → `notImplemented`, discarding the snapshot. Fixed by (a) stripping the subscription intent in `RemoteHubSource.ReadThread`/`ReadItemCandidates` and (b) gating the relay on a new `threadRelayCapableSource` capability (`SupportsThreadRelay`), which `RemoteHubSource` reports false until 05b. I deliberately did **not** repurpose `RelayOnThreadRead` for this: `TestHubRPCThreadReadSubscribeOverridesSourceReadRelayPolicy` pins that a Subscribe read overrides the plain-read policy, so the "cannot relay at all" case is a separate capability. `SupportsThreadRelay` is deleted in 05b alongside `RelayOnThreadRead`, since 05b implements `SubscribeThread`. Pinned by `TestRemoteHubSourceReadThreadStripsSubscription` and end-to-end `TestHubRPCThreadReadSubscribeServesRemoteSnapshot` (fails pre-fix with `SubscribeThread is not implemented yet`).

- **Medium — image URLs and file-backed paths are not translated for remote threads.** Partly confirmed. The controller-local file-backed half was real: `enrichThreadFileBackedOutputImages` ran on the remote thread's CWD, so the controller probed controller-local paths named by remote data and could attach a descriptor for a file the remote thread never wrote. Fixed with a `threadReadLocalImagePolicy` capability (`RemoteHubSource.EnrichThreadFileBackedImages()` returns false); remote threads are still URL-stamped but never file-enriched. Pinned by `TestHubRPCThreadReadDoesNotEnrichRemoteFilesAsLocal`, which puts a real `plot.png` in a local temp dir used as the remote CWD; pre-fix it attaches a bogus `shell-path` descriptor.
  The URL-translation half is **not fixed here**. Making remote image previews actually render needs a controller-side proxy route: the remote hub's `/s/<session>/images/<sha>` and `/doc/image?session=...` URLs are resolved against *this* hub's own state dir (`handleSessionImage` via `Past.Find`, and `doc_serve.go`'s `localSessionCWD` is explicitly local-session-only), so they 404 on the controller rather than returning wrong content. A proxy route lives outside this PR's file set (`web.go`/`image_serve.go`/`doc_serve.go`), so I have not added one; this wants a follow-up (fleet/web component). Flagging it plainly rather than pretending it is closed.

- **Low — every item read mints a fresh incarnation, invalidating outstanding cursors.** Confirmed. `ItemCandidatesFromRead` and the first `ListItemCandidates` page always minted a new incarnation and overwrote the single per-thread entry, so any interleaved re-read of the same thread invalidated an outstanding cursor even when the transcript had not changed. Fixed: `remoteItemPageIdentity` reuses the retained identity while a fresh page's newest position is not older than the retained head (untouched or appended transcript); a page that starts before the retained head is a rewrite/truncation and rotates the incarnation. Continuation pages preserve the retained head rather than lowering it. Pinned by `TestRemoteHubSourceReusesIdentityWhileTranscriptDoesNotShrink` (identical, append, and divergence subcases); pre-fix the identical re-read rotates the incarnation.

Verification:
- `go build ./...` → ok
- `go test ./cmd/evener-hub/ -race -count=3 -timeout 30m` → ok (801s; timeout raised as in round one, no test failed)
- `go test ./cmd/evener-hub/... -count=1` → ok (13 packages)
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (1 module, 0 findings)
- No fuzz targets added or modified, so the fuzz gap/registry gates do not apply.

obra:
Round three addressed (`7976ff9`).

- **Paging race (medium, 3 reviewers).** Real. `ListItemCandidates` and `ItemCandidatesFromRead` did an unlocked `peek` → remote I/O → `put`, so concurrent requests for one remote thread could interleave a `RebaseCursor`/`put` against another request's retained remote cursor. Both methods now take a per-thread `keyedMutexRegistry` lock keyed by `remoteItemPagingKey(s.id, threadID)`, held from peek through put, exactly as `LocalDaemonSource` does.
  *Pinned by* `TestRemoteHubSourceSerializesPerThreadItemPaging`: 4 concurrent fresh pages for one thread must never reach the remote at once. Verified it fails with the lock removed (`two item pages for one remote thread reached the remote hub at once`) and passes with it.
- **Cross-machine project annotation (medium, 2 reviewers).** Real, and distinct from the round-two image-proxy deferral. `annotateThreadProjects` resolved every row's `CWD` with the controller's local filesystem/Git and overwrote `ProjectID`/`ProjectPath`, including rows the remote hub had already stamped. It is now gated on source locality (`threadListSourceID("local", thread) != "local"` → skip), mirroring `mergePastMetadataForList` and `EnrichThreadFileBackedImages`; a remote row keeps the remote's identity. Note the live overwrite was on the sidebar list path (`app_threadlist.go`); the two read-path calls pass a slice literal and mutate a copy, so they were already no-ops — the gate covers them anyway.
  *Pinned by* `TestHubRPCThreadListKeepsRemoteProjectIdentity` and `TestAnnotateThreadProjectsSkipsRemoteThreads`. Verified both fail with the gate removed (the controller stamps a `...-kIoYFUBrRa` local project over `remote-project-id`) and pass with it.
- **Fresh remote page trusts the remote to enforce `ItemLimit` (medium, 1 reviewer).** Contested. The claim that the controller limit is bypassed does not hold: every item-candidate result — fresh list page, continuation, and the read path — passes through `packItemCandidates` (`app_item_page_fit.go:141`), which runs `appitempaging.SelectCandidates(candidates, nil, itemLimit)` with the requested normalized limit before the response leaves the hub, for any source. The remote source does not need its own select for that, unlike `LocalDaemonSource`, whose retained snapshot can hold more than the requested window.
  *Pinned by* `TestHubRPCThreadReadBoundsRemoteItemPage`: a remote read returning 3 items with `itemLimit: 1` yields exactly the newest item.
- **Remote SHA-addressed images stamped with controller-local URLs (medium, 1 reviewer).** Same finding as the round-two deferral, not a new one. The fix needs a controller-side fetch route for remote bytes (`web.go` / `image_serve.go` / `doc_serve.go`), which are outside this PR's file set; the current URLs 404 rather than returning another session's bytes. Left deferred on that same stated reason, not silently.

Verification: `gofmt -l cmd/evener-hub cmd/evener-hub/internal/appsource` → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` → ok (4.7s); `go test ./cmd/evener-hub/ -race -count=3 -timeout 30m` → ok (875s); `go test ./cmd/evener-hub/... -count=1` → ok, 13 packages; `module-lint` → PASS, 0 findings.

obra:
Round four addressed (`a1148f38`).

- **Remote image URLs stamped as controller-local routes** (`cmd/evener-hub/app_rpc.go:138-141`) — Fixed in scope. `enrichSourcedThreadImages` no longer calls `stampThreadImageURLs` for a source that reports `EnrichThreadFileBackedImages() == false` (a remote hub source). Stamping minted `/s/<session>/images/<sha>`, which `handleSessionImage` resolves against *this* hub's `Past` index; a remote session is never in it, so it could only 404 (or, on a session-id collision, serve another session's bytes). Pinned by `TestHubRPCThreadReadDoesNotStampRemoteImageURLs` (fails before: URL became `/s/t1/images/aaaa…`). **What remains:** making remote images actually fetchable needs a controller-side proxy that fetches bytes from the remote hub and rewrites the remote descriptors' own relative `/s/...` routes. That lives in `image_serve.go` and the `/s` router in `web_workspace.go`, outside this PR's file set, and 05a has no remote-byte read API — so those pre-existing remote URLs still pass through unresolvable. This is a real, explicitly-stated gap, not a silent re-deferral of the whole item.

- **SSH attachment failures not mapped to `SessionUnavailable`** (`cmd/evener-hub/internal/appsource/remote_hub_source.go:141-151`) — Fixed. `transportUnavailable` now maps `sshconn.ErrSSHStart` (component 04's typed spawn/initialize/preflight failure, which wraps the ssh stderr tail, e.g. `Connection refused`) and `"connection refused"` was added to the transport-text list. The terminal classes (`ErrSSHAuth`, `ErrProtocolIncompatible`, `ErrUnsupportedHost`, …) are deliberately not matched and stay raw, so a host that can never attach is not retried forever. Pinned by `TestRemoteHubSourceMapsSSHAttachFailureToSessionUnavailable` (fails before: returned `*fmt.wrapErrors`, not `SessionUnavailable`).

- **Cursor identity reused when remote transcript is rewritten** (`remote_hub_source.go:538-546`) — Fixed, and consistent with round three rather than superseding it. Reuse now additionally requires the observed page to be compatible with the retained window: `remoteMergeCandidates` fingerprints each observed position (transcript key + position) and a re-observed position carrying a different item rotates the incarnation and discards the stale window; a continuation that contradicts the retained window fails closed with `TranscriptItemCursorStale`. Pinned by `TestRemoteHubSourceRotatesIdentityWhenObservedItemRewritten` (fails before: identity reused across a replaced item). The existing append-reuse and divergence tests still pass. Residual: a rewrite confined to a region that is never re-observed cannot be detected by this fingerprint (same fidelity as `itemSnapshotStateAdvance`); the remote hub's own opaque cursor stays the authority there.

- **Paging completion retains only the final page** (`remote_hub_source.go:566` / `:487`) — Fixed. `remoteItemPagingState.candidates` now retains the whole observed window (every fetched page merged chronologically), not just the oldest page, so `continueCompleteRemoteItemPage` can answer any boundary minted under that identity. Pinned by `TestRemoteHubSourceCompleteContinuationAnswersEarlierBoundaries` (pages `[10,11,12] → [7,8,9] → [4,5,6]`; replaying the page-2 boundary `7` after completion returned stale before, now returns `[4,5,6]`). `TestRemoteHubSourceCompletePageContinuationServedLocally` still passes (a packer-minted boundary is served with no extra remote call).

- **`remoteItemPagingCache` evicts FIFO instead of LRU** (`remote_hub_source.go:330`) — Fixed. `put` now removes an existing key from `order` and re-appends it, stamping recency on commit like `itemSnapshotStateCache`'s `MoveToFront`, instead of moving it only on first insert. Pinned by `TestRemoteItemPagingCacheEvictsLeastRecentlyCommitted` (fails before: the touched key was evicted).

Verification: `gofmt -l` over touched dirs empty; `go build ./...` ok; `go vet ./cmd/evener-hub/...` ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` ok (7.9s); `go test ./cmd/evener-hub/... -count=1` ok (13 packages, hub 146s); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (1 module, 6s), 0 findings.

obra:
Round five addressed (`3c19180`).

- **High — gapped retained window could be marked complete**: a forward remote page that shared no position with the retained window and did not abut its newest item was still unioned, leaving a holed window; a later page that reached the beginning of the transcript then flipped `complete = true`, and every subsequent continuation was served locally from the gap, telling the client it had reached the start with the middle items never returned. `remoteMergeCandidates` now unions a forward page only when it overlaps or abuts the retained newest item (`remotePositionsAdjacent`); a gap rotates the incarnation so pre-append boundaries fail closed as stale. Pinned by `TestRemoteHubSourceRejectsGappedForwardAppend` (fails before the fix: the gapped append kept the identity and the replay was served). Backward pages are unchanged — they are fetched through the retained remote cursor, the remote's own authority for the next older page — so the existing paging and replay tests still pass.
- **Medium — cursor boundary not validated before rebasing**: `ListItemCandidates` now calls `appitempaging.ValidateCursorBoundary(state.candidates, before)` before `RebaseCursor`, so a caller that keeps a valid identity but moves the boundary gets a typed stale error instead of paging the remote from a position this incarnation never proved. Pinned by `TestRemoteHubSourceRejectsCursorBoundaryOutsideRetainedWindow`.
- **Medium — unbounded retained candidates**: the retained window is now bounded by `remoteItemPagingCandidateCapacity` (1024 items) and `remoteItemPagingByteCapacity` (8 MiB of transcript payload); a merge that would exceed either rotates the incarnation and keeps only the newly observed page, so the just-returned cursor stays live and older boundaries fail closed as stale instead of being answered from a truncated history. Pinned by `TestRemoteItemPagingBoundsRetainedCandidates`.
- **Low — remote image URLs pass through**: `enrichSourcedThreadImages` now neutralizes remote-supplied hub-relative `/s/<session>/images/<sha>` routes (`stripRemoteImageRoutes`) for a source whose images are not local, so the browser can no longer request this hub's own route for a session it does not have; external URLs and the descriptor SHA survive for the staged controller-side proxy. Pinned by `TestHubRPCThreadReadNeutralizesRemoteImageRoutes`.
- **Low — doc comment contradiction**: the `enrichSourcedThreadImages` comment no longer says a remote hub source "is stamped but not enriched"; it now matches the implementation (neither stamped nor enriched, remote routes neutralized).

Verification: `gofmt -l` over the touched Go dirs → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` → ok; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (all packages); `MODULES="." module-lint` → PASS, 0 findings.

obra:
Round six addressed (`06b4087d4`).

- **HIGH — `/doc/image?...` routes survived neutralization.** `isHubRelativeImageRoute` only matched `/s/.../images/...`, but the remote hub also mints root-relative `/doc/image?session=<session>&path=<rel>` URLs for file-backed output images (`outputImagesForToolCall`/`resolveOutputImageFile`), and the browser resolves a root-relative `url` against the serving (controller) origin (`reducer.ts` `outputImagesToItemImages`). Those survived and could 404 or, on a session-id collision, serve a controller-local session's bytes. It now treats **any** origin-relative path as remote-relative: `strings.HasPrefix(trimmed, "/") && !strings.HasPrefix(trimmed, "//")`, subsuming both hub routes while leaving `http(s)://`, `data:` and other scheme URLs untouched (and excluding network-path references `//host/...`, which name their own origin). Pinned by the extended `TestHubRPCThreadReadNeutralizesRemoteImageRoutes` — it now attaches a `/doc/image?session=remote-session&path=shot.png` output image and input image plus a `data:` URL, and asserts the two root-relative routes are cleared while the descriptor `Path`/`SHA` and the `data:`/external URLs survive. Fails before the fix: the `/doc/image` URL passed through.
- **MEDIUM — a forward page beginning mid-entry was unioned as adjacent.** `remotePositionsAdjacent` treated any `(entry+1)` as the successor, but an entry projects multiple items, so `(10,0) -> (11,5)` leaves items `(11,0..4)` unobserved; unioning it retained a holed window a later complete page could mark complete. It now requires the cross-entry successor to start its entry: `newer.Entry == older.Entry+1 && newer.Item == 0`. A non-abutting forward page now rotates the incarnation and fails closed as stale. Pinned by `TestRemoteHubSourceRejectsForwardPageBeginningMidEntry` (retained newest `(10,0)`, forward page starts at `(11,5)`; fails before: identity reused). Round five's `TestRemoteHubSourceRejectsGappedForwardAppend` still passes.
- **LOW — rotation retained an oversized page wholesale.** The rotation branch set `merged` to the new page without enforcing the bound, so a single page could exceed `remoteItemPagingCandidateCapacity`/`remoteItemPagingByteCapacity`. It now trims via `remoteTrimCandidatesToBound` to the newest candidates that fit both bounds, always keeping the newest candidate (mirroring the packer's nearest-item rule) so an oversized item cannot empty the window. Pinned by `TestRemoteItemPagingTrimsOversizedRotatedPage` (1049-item page → 1024 retained with the newest preserved; fails before: 1049 retained) and `TestRemoteItemPagingTrimsOversizedRotatedPageByBytes` (5×2 MiB → newest 4 retained = 8 MiB; fails before: 10 MiB retained).
  - **Supersedes part of round five's `LOW/MEDIUM` disposition.** Round five stated the rotation keeps "the just-returned cursor stays live". For a page that *alone* exceeds the bound the retained window can no longer contain that page's oldest boundary, so a continuation at a boundary older than the retained window now fails closed as `TranscriptItemCursorStale` instead of being answered. This is deliberate and only reachable from a remote that over-returns past the 40-item page limit or a single item past the byte budget; retaining the page wholesale was the reported bug. Boundaries within the retained window stay live.

Verification: `gofmt -l` over the touched Go dirs → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 20m` → ok (28.5s); `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (hub 177.9s; all 13 packages ok); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (1 module), 0 findings.

obra:
Round eight addressed (`cf847cbc9`). 3 medium + 1 low, all fixed, none contested. Every prior disposition is intact.

- **Medium — paging merges ignored continuity flags.** Confirmed. `remoteMergeCandidates` gated a forward (no-overlap) page on position adjacency alone. A retained fragment that ends mid-turn carries `HasLaterItems=true` (the turn continues after the page); a fresh page resuming at the next entry `(e+1,0)` is "adjacent" by `remotePositionsAdjacent` but skips the turn's unobserved later items, so the union retained a holed window that a later complete page could mark complete — silently omitting the intervening items. The forward branch now also requires the retained newest candidate to be complete at that boundary (`!retainedNewest.HasLaterItems`), so such a page rotates the incarnation and pre-rotation cursors fail closed as stale. Backward pages are unchanged (round five's disposition stands: they are fetched through the retained remote cursor, the remote's own authority). Pinned by `TestRemoteHubSourceRejectsForwardMergeAcrossMidTurnBoundary` (retained newest `(5,1)` with `HasLaterItems`, forward page `(6,0)`; fails before: identity reused).
- **Medium — transient `ErrRestart` not mapped to `SessionUnavailable`.** Confirmed. A failed hub restart is transient — `sshconn` stays disconnected and retries on the next `Ensure`, and `ErrRestart` is deliberately absent from `isTerminal` — so the raw wrapped error prevented the recovery/auto-resume gate from attributing the outage. `transportUnavailable` now maps `sshconn.ErrRestart` to `SessionUnavailable`, exactly like `ErrSSHStart`. The terminal classes (`ErrSSHAuth`, `ErrProtocolIncompatible`, `ErrUnsupportedHost`, …) are still not matched and stay raw, per round four's disposition. Pinned by `TestRemoteHubSourceMapsRestartFailureToSessionUnavailable` (fails before: returned `*fmt.wrapErrors`, not `SessionUnavailable`).
- **Medium — rotation branch returned a dead continuation cursor after trimming.** Confirmed. When `recordRemoteItemPage` rotated because the union exceeded the retained bound, `merged` was the trimmed newest suffix but the returned `window.Candidates` was the full untrimmed page and the `OlderCursor` was encoded at `candidates[0].Position`. When the trim dropped `candidates[0]`, that cursor named a never-retained position, so the first continuation failed `ValidateCursorBoundary` as `TranscriptItemCursorStale` — a remote page alone exceeding `remoteItemPagingCandidateCapacity` (1024) or `remoteItemPagingByteCapacity` (8 MiB) stopped paging after the first page. The rotation branch now sets `windowCandidates = merged` and mints the cursor from `merged[0].Position`. This is consistent with round six's stated intent ("boundaries within the retained window stay live") and round six's supersession of round five's rotation note is unchanged: boundaries older than the retained window still fail closed. Pinned by strengthened `TestRemoteItemPagingTrimsOversizedRotatedPage` (1049-item page → 1024 retained; asserts the returned cursor decodes at the returned window's oldest item and the continuation resumes) and `TestRemoteItemPagingTrimsOversizedRotatedPageByBytes` (5×2 MiB → newest 4; same assertions); both failed before with `transcript item cursor is stale`.
- **Low — `subagentPreview` skipped image neutralization.** Confirmed for the remote-source path. The handler returned `source.ReadThread(...)` through `subagentPreviewFromThread` without `enrichSourcedThreadImages`, so a remote item's origin-relative `/s/<session>/images/<sha>` or `/doc/image?...` route reached the browser and resolved against the controller origin (404, or another session's bytes on a session-id collision). It now applies `enrichSourcedThreadImages(source, resp.Thread)` on the source path; the past-store fallback is untouched (it has no remote source and no origin-relative remote routes). Pinned by `TestHubRPCSubagentPreviewNeutralizesRemoteImageRoutes` (remote input image `/s/remote-session/images/<sha>` cleared, external URL preserved; fails before: route passed through).

Verification: `gofmt -l cmd/evener-hub cmd/evener-hub/internal/appsource` → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 20m` → ok (38.9s); `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (all 13 packages; `cmd/evener-hub` 203.2s); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (1 module), 0 findings.

obra:
Round nine addressed (`2046b8227`).

- **Nested transcript references not remapped** (`remote_hub_refs.go:78-97`) — correct for delegate refs: `EvenerDiagnostics.Delegates[].TranscriptRef` is `local:<childSessionID>` (`agent/delegate_tree_start.go:139` and `agent/subagents.go:2372` both stamp `encodeRef("", …)`), so an untranslated one resolves against the controller's own `local` source (or fails) instead of the remote child. `fromRemoteThread` now rewrites diagnostics delegate refs into this source's namespace. `Jobs[].TranscriptRef` is not that shape: `agent.jobTranscriptRef` stamps `job:<jobID>`, a daemon-internal namespace rather than a source-qualified thread ref, so remapping it has no representation and refusing it would fail every remote read that carries a background shell job. Those refs (and a `proj:<project>:<session>` ref) pass through byte-for-byte; a client that follows one still fails loudly (`source not found: job`/`proj`) instead of reaching the wrong host. Pinned by `TestRemoteHubFromRemoteThreadTranslatesNestedDiagnosticRefs` (on the previous head: `delegate transcript ref = "local:child"`).
- **Cursor continuity assumes consecutive entry numbers** (`remote_hub_source.go:664-676`) — confirmed: `groupedAppTurnProjection` assigns an ordinal to every logical group and skips groups with no visible items, so the next visible turn after entry 10 can be entry 12 (pinned upstream by `TestEmptyLogicalGroupReservesOrdinalInFullAndIndexedProjections`). The `Entry+1` abutment gate therefore rotated the incarnation over an ordinary append and staled a live cursor, and a fresh tail page that shared no position with the retained window did the same. Continuity is now decided from the fragments' own completeness metadata: a fragment cut mid-turn (retained `HasLaterItems`, or a page starting mid-turn) is still a known hole and fails closed; an unproven forward span keeps the incarnation, is recorded as a `remoteItemGap`, and no locally served answer may span it — a complete window serves only its contiguous prefix, and a boundary above a recorded span fails closed while a boundary below it is still served locally. Pinned by `TestRemoteHubSourceKeepsIdentityAcrossSkippedEntryAppend` and `TestRemoteHubSourceRecordsUnobservedForwardSpanWithoutRotating`. This **supersedes the round-eight disposition** pinned by `TestRemoteHubSourceRejectsGappedForwardAppend` (rotation, plus "the replayed boundary must never reach the remote"): that rule staled a live cursor on any append whose page did not resume at `Entry+1`, which is the finding. Its soundness property — a gapped window is never served as if complete — is preserved and re-pinned in the replacement test.
- **Retained-state byte bound ignores non-string fields** (`remote_hub_source.go:685-699`) — confirmed: `remoteItemCandidateBytes` counted only `Text/Delta/ArgumentsJSON/Output/Error/Description/Raw`, so an image-heavy page retained far past the documented 8 MiB. It now counts `Images` (type, text, url, media type, data, name, path, and metadata) and `OutputImages` fields. Pinned by `TestRemoteItemPagingBoundsRetainedImageBytes` (previous head retained all five 3 MiB images; now the newest two are retained and the window is ≤ 8 MiB).
- **Missing `ValidateCandidates` after `CandidatesFromTurns`** (`remote_hub_source.go:525`, also `:510`) — confirmed; the local sibling validates at all five equivalent points (`source.go:94,347,414,459,581`). Both `remoteItemPage` and `ItemCandidatesFromRead` now validate before anything is merged or cached, failing closed. Pinned by `TestRemoteHubSourceRejectsUnorderedRemoteItemPage` and `TestRemoteHubSourceRejectsDuplicateKeyRemoteItemRead` (both pages were accepted on the previous head; now they are rejected with nothing retained).

Every new test was run against the previous head first and failed there. Prior dispositions are intact: the `CursorIdentity` type and bounded remote-cursor cache, ref translation over known structural locations, image isolation, and the `RelayOnThreadRead`/`SupportsThreadRelay` stand-ins are untouched.

Verification: `gofmt -l cmd/evener-hub/ cmd/evener-hub/internal/appsource/` → empty; `go build ./...` → 0; `go vet ./cmd/evener-hub/...` → 0; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 20m` → ok (69s); `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → all ok (`cmd/evener-hub` 187s, `internal/appsource` 3.1s, 11 other packages ok); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → `PASS lint (1 modules, 127s)`, 0 findings.

obra:
Round ten addressed (`4f4acede5`).

- **Capability masking** (`cmd/evener-hub/internal/appsource/remote_hub_refs.go`) — confirmed. `fromRemoteThread` forwarded the remote hub's own capability set verbatim, and the hub gates every thread action on exactly those fields (`cmd/evener-hub/app_compact.go`'s `threadActionAvailable`), so a remote thread advertised send/steer/interrupt/queue/compact/clear/fork/shutdown/model/vision-model/goal/notes/rename/skill-input while every corresponding `RemoteHubSource` method still returns `notImplemented` (an internal error the moment a client used one). Fixed as an intersection later components flip on, not as a denial: `maskRemoteThreadCapabilities(remote)` ANDs the remote's claim field by field against `remoteForwardedThreadCapabilities()`, which is the empty set on this branch. The fields are listed one by one rather than copied wholesale, so a capability added to `appwire.ThreadCapabilities` later starts masked until a component names the method that answers for it.
  - **What 05c must change:** add each action to `remoteForwardedThreadCapabilities()` as its method stops returning `notImplemented` — Send (`StartTurn`, `ResumeThread`), Steer (`SteerTurn`), Interrupt (`InterruptTurn`), Queue (`QueueTurn`, `DrainAsSteer`, `PromoteQueuedAsSteer`, `CancelQueued`), Compact (`CompactThread`), Clear (`ClearThread`), ForkFromTurn (`ForkThread`), Shutdown (`ShutdownThread`), ChangeModel (`SetThreadModel`, `SetThreadReasoningEffort`), ChangeVisionModel (`SetThreadVisionModel`), Goal (`GoalSet`), SharedNotes (`NotesHumanSet`, `UrlsRemove`), Rename (`SetThreadName`), and SkillInput with the input-bearing turn mutations.
  - **What 05d must change:** replace that constant answer with the completed host capability probe, intersected the same way.
  - Pinned by `TestRemoteHubFromRemoteThreadMasksUnforwardedCapabilities` (an all-true remote claim leaves no action advertised while refs and `InstanceID` are untouched) and `TestRemoteHubCapabilitiesMatchForwardedMethods` (every `ThreadCapabilities` field is enumerated with the method(s) that answer for it; it asserts those methods still fail closed while the field is masked, and fails if a field is added without a method mapping — so a flag cannot be flipped on while its method still 404s). Failing-first: with only that one masking assignment removed from `fromRemoteThread`, the fixture's all-true remote set reached the thread (`remote thread advertised {Send:true … SkillInput:true}`).

- **Continuation cursor validation** (`remote_hub_source.go`) — confirmed. A remote continuation page went straight into the retained window with nothing checking that it progressed past the boundary the remote was asked to continue from, that its cursor canonically named the page it arrived with, or that the cursor carried the identity of the request cursor, so a stale or repeating remote page could be cached and re-served forever under a fresh controller cursor. `ListItemCandidates` (fresh and continuation) and `ItemCandidatesFromRead` now mirror `LocalDaemonSource`'s checks: `validateRemotePageCursor` retains a remote cursor only when it canonically names the page's oldest item (the position both sides of this stack mint at — `packedOlderCursor` for a live page, `apptranscript.ItemWindowOptions` for a past one) and refuses a cursor that arrives with no page; `validateRemoteContinuationPage` additionally requires the returned cursor to carry the identity of the cursor that was sent, and clips the page to items strictly older than the requested boundary (a page with nothing older is `TranscriptItemCursorStale`), exactly as `localDaemonCandidatesBefore` clips. A page or cursor that violates the contract now fails closed instead of being retained.
  - Pinned by `TestRemoteHubSourceRejectsInvalidRemoteContinuationPages` (repeated page / cursor naming another page / cursor from another incarnation / cursor without a page — each a typed stale with nothing cached, asserted against the retained state) and `TestRemoteHubSourceClipsRemoteContinuationPageToTheRequestedBoundary` (a page that rescans the boundary is clipped to the items below it and keeps a live cursor). Failing-first: 3 of the 4 rejection cases and the clipping case failed before the change.
  - Fixture consequence, disclosed: eight existing fixtures handed out cursors naming a position other than the page they accompanied (often the page's newest item), which no hub mints. `TestRemoteItemPagingBoundsRetainedCandidates`, `TestRemoteItemPagingTrimsOversizedRotatedPage`, `TestRemoteItemPagingTrimsOversizedRotatedPageByBytes`, `TestRemoteItemPagingBoundsRetainedImageBytes`, `TestRemoteHubSourceReusesIdentityWhileTranscriptDoesNotShrink`, `TestRemoteHubSourceRotatesIdentityWhenObservedItemRewritten`, `TestRemoteHubSourceKeepsIdentityAcrossSkippedEntryAppend`, `TestRemoteHubSourceRejectsForwardPageBeginningMidEntry` and the RPC-level `TestHubRPCThreadReadServesRemoteHubSource` now mint the canonical cursor (or script a strictly older continuation page). Every assertion is unchanged; the skipped-entry test still demonstrates the boundary rebase (retained cursor at entry 8, client boundary at entry 10), and the RPC test still checks that an item page carrying the remote's own cursor packs into a controller-owned continuation.

- **Retained-window gap bookkeeping** (`remote_hub_source.go`) — confirmed exactly as described. `remoteGapsAfterObserving` cleared a span only when one page reached from `gap.lower` to `gap.upper`, but a continuation is fetched from a boundary at or above `gap.upper`, so no single page could ever satisfy it: the span a forward append created survived the piecewise refill, and once the remote cursor was exhausted `continueCompleteRemoteItemPage` restricted itself to `remoteContiguousPrefix`, so replaying a cursor minted for the tail page returned `TranscriptItemCursorStale` for a boundary that was retained and a transcript that was fully observed. Replaced the gap list with the observed runs your second suggestion points at: `remoteItemSpan` plus `remoteItemSpansAfterObserving` / `remoteItemSpansForCandidates` / `remoteSpanContaining`, and `remoteContiguousPrefix` now ends at the upper edge of the run holding the oldest retained candidate. A page contributes its own internally contiguous run and merges with every recorded run it provably connects to: an overlapping run; an exact successor position whose lower fragment is complete (`remotePositionsAdjacent` + `HasLaterItems`, the proof the merger already used for an abutting forward page); and — the case position-only inference cannot see — the run covering the boundary the continuation was fetched from, because a page returned for a retained remote cursor is the next older page in the remote's own sequence. That is the "explicit traversal information" that closes the span piecewise. Round nine's conservative direction is preserved: a run that was never connected stays separate and a boundary above it still fails closed.
  - Pinned by `TestRemoteHubSourceKeepsTailCursorsLiveAfterPiecewiseSpanFill` (retained `[1..5]`, forward tail `[40..49]`, the span filled piecewise with `[30..39]` then `[6..29]` through the remote cursor, then the tail cursor replayed and required to serve `[1..39]`); it fails pre-fix with exactly the `TranscriptItemCursorStale` the finding describes. Round nine's `TestRemoteHubSourceRecordsUnobservedForwardSpanWithoutRotating` is unchanged and still passes, so an unfilled span still fails closed rather than being served across.

Verification: `gofmt -l cmd/evener-hub/internal/appsource/ cmd/evener-hub/ internal/appitempaging/` → empty; `go build ./...` → exit 0; `go vet ./cmd/evener-hub/...` → exit 0; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 20m` → ok 49.9s; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → exit 0 (all 13 packages ok, `cmd/evener-hub` 147.4s); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (1 module, 6s).

obra:
Round eleven addressed (`ee1abd9d8`).

- **Remote image routes unsanitized on metadata enrichment failure** (`cmd/evener-hub/app_rpc.go:239-249`) — confirmed: `listItemTurns` ran the remote-image neutralization only inside `if metaErr == nil`, so when the optional `thread/read` metadata call failed, the packed item page went out with the remote hub's root-relative image routes intact and the browser resolved them against this hub's origin. The enrichment is now split into `threadImagesLocal` + `enrichLocalSourcedThreadImages`, and the packer callback always runs `stripRemoteImageRoutes` for a source whose images are not local, while the two local passes (which need the session id and CWD only a successful read supplies) still run only after one. Pinned by `TestListItemTurnsNeutralizesRemoteImageRoutesWithoutMetadata` (fails before the fix: the `/s/...` and `/doc/image?...` routes came back unchanged, while the external URL still survives).
- **Retained-byte estimate omits variable-length fields** (`remote_hub_source.go:1010-1022`) — confirmed: the sum listed payload fields by hand and missed `ThreadItem.ID`, `ToolName`, `CallID`, `Source`, `SteeringKind`, `ClientMutationID` and every field of the candidate's turn, so a page could retain megabytes per item while the estimate sat near zero against the documented 8 MiB bound. The estimate is now a walk of the retained value (`remoteRetainedPayloadBytes`): every string, `[]byte` and map entry a candidate holds contributes its own length, so a field added to `ThreadItem`/`InputItem`/`OutputImage`/`Turn` is counted as soon as it exists and there is no list to fall behind. The single field left out is the page's shared `Turn.Items` fragment slice — its elements' payload is already counted through each candidate's own `Item`, and counting it per candidate would multiply a page by its item count and rotate windows that fit. Pinned by `TestRemoteItemPagingBoundsRetainedVariableLengthFields` (a page large only through those fields is trimmed to the bound with a live continuation cursor), `TestRemoteItemCandidateBytesCountsVariableLengthFields` (one subtest per previously-invisible field, plus nested metadata and `Raw`), and `TestRemoteItemCandidateBytesCountsSharedTurnItemsOnce`.
  - Superseded assertion, stated explicitly: `TestRemoteItemPagingTrimsOversizedRotatedPageByBytes` pinned the retained count at a literal `4` for five 2 MiB items. Completing the estimate moved that boundary by each candidate's turn-level fields, so the literal became a maximality assertion (`requireNewestRetainedRunThatFits`): the window is inside the byte bound and the next older candidate would exceed it. The property the literal protected — the newest run that fits is retained, no more and no less — is pinned more directly than before, and the rest of that test's assertions are unchanged.
- **Paging contiguity gap accepted as contiguous span** (`remote_hub_source.go:572-579, 981-985`) — confirmed: `remoteItemPage` validated only strict increase, so a page carrying `(10,0)` and `(10,2)` was accepted and then recorded (`remoteItemSpansForCandidates`, `remoteItemSpansAfterObserving`) as one run covering the withheld `(10,1)`; the complete-page local continuation serves that run, so the missing item would be skipped silently. `validateRemotePagePositions` now also refuses a page that skips an item index inside one entry (per-entry item indices are dense — `apptranscript.ProjectTurn`, `server.positionAppItems`), at both page entry points (`remoteItemPage` and `ItemCandidatesFromRead`), before anything is merged, retained or recorded. Cross-entry jumps stay accepted: entry ordinals advance for logical groups that project no visible item, which positions alone cannot rule out. Pinned by `TestRemoteHubSourceRejectsNonAdjacentRemoteItemPage` and `TestRemoteHubSourceRejectsNonAdjacentRemoteContinuationPage` (the latter also shows a refused continuation leaves the retained window and its cursors exactly as they were).
  - No conflict with the round-ten property: adjacency is checked only *inside* one entry, while the piecewise fill joins runs *across* entries through the retained remote cursor's own traversal evidence. `TestRemoteHubSourceKeepsTailCursorsLiveAfterPiecewiseSpanFill`, `TestRemoteHubSourceRecordsUnobservedForwardSpanWithoutRotating` and `TestRemoteHubSourceKeepsIdentityAcrossSkippedEntryAppend` all pass unchanged.
  - Every prior disposition is intact: capability masking, `CursorIdentity` and the bounded cursor cache, ref translation over known structural locations, image isolation, `RelayOnThreadRead`, and the round-ten span bookkeeping. No existing assertion was weakened or deleted.

Authorized out-of-scope edit: `cmd/evener-hub/app_rpc.go` is outside this PR's file set. I touched it for the minimal sanitization fix above (split `enrichSourcedThreadImages` into the local-only pass plus `threadImagesLocal`, and the callback's remote-image branch) and its test in `cmd/evener-hub/app_rpc_test.go` — nothing else in either file. Residual, left untouched to keep the edit tight: the same "enrich only when the metadata read succeeded" gate exists on the legacy non-item-candidate path (`cmd/evener-hub/app_rpc.go:706-717`); it is unreachable for `RemoteHubSource` — the only `EnrichThreadFileBackedImages` implementer — because the source is an `ItemCandidateSource` and returns before that branch.

Verification:
- `gofmt -l cmd/evener-hub` → empty
- `go build ./...` → exit 0
- `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 20m` → `ok primeradiant.com/evener/cmd/evener-hub/internal/appsource 78.730s`
- `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → all packages ok (`cmd/evener-hub 161.369s`, `internal/appsource 2.581s`, all other internal packages ok)
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → `PASS lint (1 modules, 48s)`, 0 findings

obra:

The parent component (#1287, 04b) landed as `5275b578f`, so this component's own 12 commits were rebased onto `main` (`git rebase --onto origin/main e5cb0c0fc`) and force-pushed with `--force-with-lease` (`ee1abd9d8` → `ae24ad2e3`). Rebasing rather than merging is deliberate: a merge drags the squashed ancestors' history and conflicts in ~18 files that have no content difference, while the rebase keeps this PR's diff exactly this component's — the PR now shows its own 12 files, +5334/−27, against `main`.

**One conflict, resolved by hand — `cmd/evener-hub/app_rpc.go` (`newHubSourceRegistry`).**

- `main` had since refactored the local-source path: the inline entry closure became `localDaemonEntriesFromRoster`, the `SetOnSessionGone` announcement wiring was added, and a nil roster now returns early — "without one there is no local source at all, so a lookup of a local ref fails as source not found rather than finding a source that lists nothing".
- This component adds the remote-host registration block (`cfg.RemoteHosts` / `cfg.RemoteHostClient`, with the stderr diagnostic when no SSH client is wired).

The resolution keeps `main`'s helper-based local path and appends this component's remote block after it. The one structural change: `main`'s early `return registry` on a nil roster became `if roster != nil { … }`, so a configured remote host is still registered when no local roster is wired — which is what this component's code did. (The hub always wires a roster in production, so this only affects the defensive path.)

**Test updates that follow from that resolution** (`cmd/evener-hub/remote_source_registration_test.go`, commit `ae24ad2e3`):

Three tests constructed configs without a roster and expected the local source to be present — the behaviour `main`'s refactor deliberately changed. They now pass `Roster: hubcore.NewRosterWithEntries()` so each keeps asserting its own property (local plus remotes registered; host refs resolving to the host source), and a new test pins the merged nil-roster behaviour explicitly: no local source, a configured remote host still registered. No assertion was deleted or weakened — the first full-suite run on the rebased tree caught all three (`source "local" not registered`, `local source not registered`, `sourceForThread: source not found: local`), which is why the fix is visible here rather than folded silently into the rebase.

**Gates on the rebased tree (`ae24ad2e3`)**: `gofmt -l cmd/evener-hub/` clean; `go build ./...` exit 0; `go vet ./cmd/evener-hub/...` exit 0; `go test ./cmd/evener-hub/ -count=1 -timeout 30m` ok (157.7s); `go test ./cmd/evener-hub/internal/appsource/ -race -count=1` ok (20.6s); module-lint PASS (1 module, 0 findings).

obra:
Round eleven addressed (`24af2c8d0`).

This comment covers the rebased head (`24af2c8d0`); the earlier round-eleven comment addressed the pre-rebase head (`ee1abd9d8`) and its image-route finding.

- **`[[hosts]]` `config_path` and `addr` were dropped when building the host registry** (`cmd/evener-hub/main.go:406`, medium): confirmed against the code. `HostConfig` carries both fields (`cmd/evener-hub/config.go:36-44`), `validateHostConfigs` normalizes and copies them back into `cfg.Hosts` (`cmd/evener-hub/config.go:174`, copy-back at 187-195), `hostreg.Host` documents them (`internal/hostreg/hostreg.go:54-58`), and `sshconn` reads exactly them: `channelArgv` / `hubBootstrapArgv` append `--config` (`internal/sshconn/runner.go:318,334`), and the probe/restart path resolves the host address (`internal/sshconn/version.go:56,93`; `internal/sshconn/manager.go:864,1040,1184`). The config-to-`hostreg.Host` mapping omitted both, so `hostreg.New` — and the same slice handed to `WebConfig.RemoteHosts` as one source per host (`main.go:455`) — never saw them; the default-address case still worked, which is why nothing caught it.

  **Inherited from the rebase, not introduced by it.** The mapping was authored in `1606c6a58` (2026-09-14 04:57Z) on a lineage that did not yet contain component 03. The `ConfigPath`/`Addr` pair arrived later with `412ebc650` (component 03 host config, #1280); sshconn's consumers arrived with `98ff64c93` (component 04a, #1283) and `5275b578f` (component 04b, #1287). Today's replay onto `main` (`c2deabf82`, committer date 2026-09-16 07:00:39Z) carried the older mapping over unchanged, so the drop predates the rebase on this branch — the rebase resolution preserved it rather than introducing it.

  **Fixed** by extracting the mapping into `hostRegistryEntries(cfg Config) []hostreg.Host` (`cmd/evener-hub/main.go:601`), which `runMain` now calls once at `main.go:404` so both consumers see the same entries, and carrying `ConfigPath: h.ConfigPath, Addr: h.Addr`. The helper takes the hub's own `Config` because that is what `runMain` holds at this site; `hubcore.WebConfig` is the consumer that receives the result as `RemoteHosts`. Pinned by `TestHostRegistryEntriesCarryEveryHostField` (`cmd/evener-hub/remote_source_registration_test.go:153`): it drives exactly what `runMain` does and asserts every field (Name, SSH, User, EvenerPath, ConfigPath, Addr, Roots) both through the helper and through `hostreg.New` + `Registry.Get`, mirroring the field-completeness assertion at `config_test.go:51`.

  Failing-first against the pre-fix mapping (helper extracted, fields not yet carried):

  ```
  --- FAIL: TestHostRegistryEntriesCarryEveryHostField (0.00s)
      remote_source_registration_test.go:175: hostRegistryEntries() = [{Name:alpha SSH:alpha.example User:deploy EvenerPath:/opt/evener/bin/evener ConfigPath: Addr: Roots:[/srv/one /srv/two]}], want [{Name:alpha SSH:alpha.example User:deploy EvenerPath:/opt/evener/bin/evener ConfigPath:/etc/evener/alpha-hub.toml Addr:127.0.0.1:9280 Roots:[/srv/one /srv/two]}]
  FAIL
  FAIL	primeradiant.com/evener/cmd/evener-hub	0.029s
  ```

Verification: `gofmt -l cmd/evener-hub/` empty; `go build ./...` exit 0; `go vet ./cmd/evener-hub/...` exit 0; `go test ./cmd/evener-hub/internal/appsource/ -race -count=1 -timeout 20m` ok (18.1s); `go test ./cmd/evener-hub/ -count=1 -timeout 30m` ok (128.3s); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` PASS, 0 findings.

obra:
Round twelve addressed (`79ee4c33d`).

- **medium — `Thread.Evener.PendingEscalations[].Ref` was never translated** (`cmd/evener-hub/internal/appsource/remote_hub_refs.go`, `fromRemoteThread`). The remote daemon stamps each pending escalation with the owning session's ref in its own namespace (`server/thread_envelope.go:384`), and a client answers a card by that ref (`evener/sandbox/escalation/resolve` → `sourceForThread` → registry `SourceForRef`), so an untranslated `local:<session>` targeted the controller's own local source: the wrong session on an id collision, `source not found` otherwise. `fromRemoteThread` now walks the slice and rewrites each `Ref` through `fromRemoteNestedRef` — the same idempotent/lenient helper the nested diagnostic refs use; `ThreadID` is already bare and is left untouched. Pinned by `TestRemoteHubFromRemoteThreadTranslatesPendingEscalationRefs`, which also covers empty refs, an already-`host:`-qualified ref, a bare id that is not valid ref syntax, and a foreign namespace surviving byte-for-byte rather than failing the response. Failing first against the unfixed code: `remote_hub_refs_test.go:366: pending escalation ref = "local:th_1", want host:th_1`.
- **Escalation-resolve side:** no change needed. Reading the resolve path, the controller routes by `params.Ref` (`app_rpc.go` → `withSessionActionOwnership` → `sourceForThread` → `SourceForRef`), so the translated `host:<id>` now resolves to the `RemoteHubSource` instead of `local`. `RemoteHubSource.ResolveSandboxEscalation` still returns `notImplemented` on 05a (forwarding it is 05c's mutations work per the design doc component split), so the card is correctly routable but not yet answerable — and it can no longer reach a colliding local session.

Verification:
- `go build ./...` → OK
- `go vet ./cmd/evener-hub/...` → OK
- `go test ./cmd/evener-hub/internal/appsource/ -race -count=1 -timeout 20m` → ok 16.686s
- `go test ./cmd/evener-hub/ -count=1 -timeout 30m` → ok 122.022s
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS lint (1 modules, 44s), 0 findings

obra:
Round thirteen addressed (`536cc7c2d`).

- **Medium — First `thread/list` can silently omit remote hosts whose SSH attach exceeds the per-source list timeout**: confirmed and fixed. The local-daemon deadline no longer judges a remote host's cold attach: `hubThreadListWithSourceTimeout` (`cmd/evener-hub/app_threadlist.go`) now takes each source's own list deadline (`threadListBudgetSource.ThreadListBudget()`) and keeps the existing 3s budget for every source that reports none, so local sources are unchanged. `RemoteHubSource` reports `remoteThreadListBudget = 15s` — above the ssh transport's own 10s connect bound, so a host that cannot be reached fails on its transport rather than on the list budget, and far below sshconn's preflight/deploy bounds (70s / 10m), which no list may block on.
  - **Shape: the budget, not startup warming.** The first suggested remedy is ruled out by the component-05 design: `docs/superpowers/specs/2026-09-14-multi-host-05-remote-hub-source.md` §"A background/snapshot caller must not force attachment" and §"The same gate applies to the primary `thread/list` fan-out" require a non-explicit fleet list to run only against sources that are already attached, and name attaching every configured host "eager attachment that contradicts the lazy manager (component 04), the attachment-based `Online` (component 06), and fleet-view acceptance criterion 9 ('no other path attaches a host implicitly')". The third (a per-host failure marker) needs a field on `appwire.ThreadListResponse`, outside this PR's file set.
  - **Fleet-online check** (why this was not answered "by design"): the signal does not cover the omission today. `WebConfig.RemoteHostOnline` does not exist anywhere in the tree; `apiTreeSources` (`web_api_tree.go:717-739`) hardcodes `Online: true` for every source; and `sshconn.Manager.ClientIfAttached` (`manager.go:673`) still has no production caller. A skipped host was therefore presented as online, with nothing else indicating the skip.
  - **Tests.** `TestHubRPCThreadListServesRemoteHostWhoseAttachOutlastsLocalBudget` (`cmd/evener-hub/remote_hub_source_list_budget_test.go`) serves an unfiltered `thread/list` through the hub's own RPC against a remote source whose connector withholds the client for `threadListSourceTimeout + 500ms` — a cold attach. Failing-first: with the fan-out still on the flat budget it failed `thread list = [], want the cold remote host's row` (3.01s), the host dropped exactly as reported; with the seam wired the row is present. `TestHubThreadListUsesSourceReportedBudget` pins that a budget-reporting source gets its own deadline while a source that reports none keeps the caller's default, and `TestRemoteHubSourceThreadListBudget` pins the documented 15s and that the real source's budget is what the fan-out applies.
  - **Residuals, disclosed.** An attach that outlives 15s (the deploy / version-auto-match ladder) still ends that list without the host; that is a sustained outage rather than a cold start, and its surfaces are component 06's attachment state and an explicit `thread/list` on that host, which propagates the typed error today. A host that is down now costs a list up to the transport's own 10s connect bound before it is skipped, where the 3s cut-off used to end it — the price of no longer reading a slow attach as absence.

Verification: `gofmt -l cmd/evener-hub/` → no output; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=1 -timeout 20m` → ok 17.184s; `go test ./cmd/evener-hub/ -count=1 -timeout 30m` → ok 125.134s; `GOLANGCI_LINT_CACHE=$(scripts/lib/golangci-lint-cache.sh) MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS lint (1 modules, 50s).

obra:

**CI** — run 35072803398 at `536cc7c2d`: completed/success, all jobs green.

**Review** — RoboRev's verdict for `536cc7c2d`: **"No issues found."** This closes a four-round review sequence on this branch after it was rebased onto `main` (round eleven: the host-registry mapping dropped `ConfigPath`/`Addr`; round twelve: `PendingEscalations[].Ref` was untranslated; round thirteen: a cold SSH attach outliving the 3s per-source list budget silently dropped a host from an unfiltered list — fixed with a source-reported 15s budget for remote sources, which the component-05 spec requires over eager startup attachment).

**Residuals disclosed by the round-thirteen fix** (recorded here rather than hidden):
1. An attach that outlives the 15s budget still skips that host silently. The spec's eventual rule for the unfiltered fan-out — attached-only, no dial — is the surface that closes this, and it belongs with components 05b–05d/06a.
2. An unreachable host can now cost up to the ~10s transport connect bound in an unfiltered list (it was 3s before). That is the deliberate trade for not misclassifying a legitimate cold attach.

**Up-to-date override** — the branch is 10 commits behind `main`. Overlap check, run immediately before this comment:

```
$ gh pr view 1290 --json mergeable,mergeStateStatus --jq '.mergeable, .mergeStateStatus'
MERGEABLE
BEHIND

$ comm -12 <(gh pr view 1290 --json files --jq '.files[].path' | sort) \
           <(git diff --name-only "$(git merge-base origin/main origin/multi-host-pr05a-remote-hub-source)" origin/main | sort)

$ git diff --name-only "$(git merge-base origin/main origin/multi-host-pr05a-remote-hub-source)" origin/main | wc -l
97
```

`main` changed 97 files since the merge base; none of them is a file this PR touches.

PR #1291, opened 2026-09-14
feat(hub): remote hub subscription fan-out (multi-host 05b)

obra:
Part 05b of component 05 (remote hub source), stacked on #1290 (05a).

## What this adds
`RemoteHubSource.SubscribeThread` and the notification fan-out that makes a remote thread stream live to controller subscribers.

- `SubscribeThread` translates the controller ref, issues `thread/read {ref:"local:<t>", threadId:"<t>", subscribe:true}` on the source's shared per-host client, and **discards the returned snapshot** — the controller relay always reads the thread before subscribing (the non-atomic `prepareRelay` branch).
- One drain goroutine per remote client continuously reads `client.Notifications()`. This is mandatory, not an optimization: `appwire`'s notification buffer is 4096 and overflow tears the connection down (`ErrNotificationOverflow`), so a remote host with one busy thread would otherwise lose every subscription on that host.
- Notifications are translated at the JSON level — `ref` is rewritten from `local:X` to `host:X`, a nested `thread` snapshot goes through the same `fromRemoteThread` translation (leaving `InstanceID` untouched), `threadId` is passed through unrewritten, and unknown fields are preserved. No method switch, so every ref-bearing notification is covered uniformly.
- Routing is by remote thread ID into a per-subscription channel. Each subscription owns two channels with single-writer discipline (`in` is never closed; `out` is written and closed only by its pump), so a fan-out send can never panic on a closed channel.
- The returned channel closes on context cancellation, which the controller relay treats as subscription end and re-subscribes from.
- A notification whose ref names a nested remote hub (anything but `local`) is dropped: it is not representable in the controller namespace and a notification has no error channel.

## Tests
Eight new tests in `remote_hub_subscription_test.go` over an in-memory `net.Pipe` + `StreamTransport` harness (no SSH, no network, no host): wire contract + snapshot discarded; ref and nested-thread translation; fan-out isolation across two threads plus drop for an unsubscribed thread; a >4096-notification burst proving the drain prevents connection teardown; channel close on cancellation and successful re-subscribe; foreign-ref refusal before any wire call; nested-ref drop; and transport-failure → `SessionUnavailable`.

## Gates (run on this branch)
- `gofmt -l cmd/evener-hub/internal/appsource` — clean
- `go build ./...` — ok
- `go vet ./cmd/evener-hub/...` — ok
- `go test -race ./cmd/evener-hub/internal/appsource/ -count=1` — ok 4.9s
- `golangci-lint run ./cmd/evener-hub/internal/appsource/...` — 0 issues
- `go test ./cmd/evener-hub/ -count=1` — ok 161.9s

## Known limitation (deliberate, per spec §Open questions)
The fan-out is a single goroutine with a blocking send into each subscriber's buffer. A subscriber that stops being read can therefore stall routing for the other subscriptions until the relay cancels it (which the relay always does on teardown). This is the spec's chosen "one client, fan-out by ref" shape; the atomic `RelaySessionSource` handoff that would remove it is explicitly deferred.


obra:
Round 1 addressed (`6b0b4145b`). 6 fixed, 1 contested (3 High fixed, 1 High contested, 2 Medium fixed, 1 Low fixed).

- **High — drainLoop didn't cancel subscriptions on client disconnect.** Confirmed: `drainLoop` deleted its `drains` entry and returned; `subCtx` derives from the relay context, so `pumpSubscription` blocked forever, `out` never closed, and the relay's recovery path (`app_relay.go`) never re-attached to the reconnected client. Fixed by recording the owning `*appwire.Client` on each subscription and, when the drain exits because that client's notification stream closed, cancelling every subscription bound to it so the pumps unwind and the relays re-subscribe. Pinned by `TestRemoteHubSubscribeThreadClosesOnClientDisconnect` (fails pre-fix: "subscription out stayed open after its client's notification stream closed").
- **High — `ReplaceSubscription` forwarded to the shared remote client.** Confirmed: `SubscribeThread` copied `params` verbatim, and the remote server's `CaptureSubscription(replace=true)` calls `ReplaceConnectionSubscriptions(conn.id, …)`, scoping the whole shared per-host connection to one thread and silently dropping every sibling's remote subscription while the local routing entries stayed live. Fixed by clearing `ReplaceSubscription` on the forwarded request; remote subscriptions are managed one-per-thread independently. Pinned by `TestRemoteHubSubscribeThreadDoesNotForwardReplaceSubscription` (fails pre-fix: forwarded `"replaceSubscription":true`).
- **High — routing-key mismatch after a stable-ref identity swap silently dropped all notifications.** Confirmed: `SubscribeThread` keyed the map on `ref.ThreadID` (stable), while `translateNotification` preferred the bare `threadId`. `ReplaceAppIdentity` keeps `s.appRef` (the stable ref) while moving `s.appThreadID` to the replacement instance, so the notification's bare `threadId` was the new instance → lookup miss → drop. Fixed by making the translated ref authoritative whenever present, falling back to the bare `threadId` only when no ref is sent (mirrors the relay's own `relayNotificationRoutingKey`). Pinned by `TestRemoteHubSubscribeThreadRoutesStableRefNotification` (fails pre-fix: delivery timed out for `{threadId:"new-instance", ref:"local:stable"}` on a `host:stable` subscription).
- **High — shared-drain head-of-line blocking. CONTESTED.** The only consumer of `out` is the relay supervisor (`app_relay.go`), whose `broadcastNotification` calls `server.Broadcast`. `Broadcast` → `Connection.enqueue` is non-blocking: it drops and evicts a slow consumer (`internal/appserver/server.go:600-612`, `evictSlowConsumer`), so a slow browser cannot stall the supervisor and therefore cannot fill `out`→`in`→the drain. The blocking send is a deliberate, documented tradeoff ("a dropped turn/completed is exactly the failure this component exists to prevent") and is only reachable under scheduler starvation. I did not introduce a test that stalls a consumer, because that state is a test-only condition and not reachable from any production consumer.
- **Medium — no `thread/unsubscribe` during subscription teardown.** Confirmed: the shared client is long-lived, so the remote hub retained every superseded thread's subscription and kept forwarding it. Fixed: `retireSubscription` sends `thread/unsubscribe{ref:"local:<thread>"}` when, and only when, the retiring subscription is still the installed one; the map mutation and the send are serialized under `remoteMu` so a replacement's `thread/read` can never land before its predecessor's unsubscribe and be dropped by it. Pinned by `TestRemoteHubSubscribeThreadUnsubscribesOnTeardown` (fails pre-fix: no `thread/unsubscribe` for `local:S`).
- **Medium — `registerSubscriber` installed/cancelled before the subscribe request succeeded.** Partly confirmed: the failed sub *is* unregistered (`unregisterSubscriber` guards on identity), so "remains registered" was inaccurate; but the previous healthy subscription was cancelled before the request and never restored, and the failure path never closed `pumpDone`, so a fan-out already blocked on the failed sub's full `in` buffer could wedge the shared drain. Fixed: the previous subscription is cancelled only after the subscribe succeeds, restored on failure, and `discardSubscriber` closes `pumpDone`. Pinned by `TestRemoteHubSubscribeThreadFailedReplacementPreservesPrevious` (fails pre-fix: previous channel closed) and `TestRemoteHubDiscardSubscriberUnblocksRouter` (fails if the `pumpDone` close is removed).
- **Low — nested `thread` object dropped unknown fields.** Confirmed: it was round-tripped through `appwire.Thread`. Fixed with `translateThreadRaw`, which rewrites only `source` and `evener.ref`/`parentRef` at the `RawMessage` level and passes every other field through byte-for-byte. Pinned by `TestRemoteHubSubscribeThreadNestedThreadPreservesUnknownFields` (fails pre-fix: `futureField` dropped).

Verification:
- `gofmt -l cmd/evener-hub/internal/appsource/` → empty
- `go build ./...` → ok
- `go test ./cmd/evener-hub/... -count=1` → ok (13 packages)
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS, 0 findings (1 module, 47s)
- No fuzz targets added or touched, so `make fuzz-gap-check`/`make fuzz-registry-check` were not required.

obra:
Follow-up on H4 (`1bd39b312`). My round-one contest was **wrong** — withdrawn.

The blocking chain is entirely inside this component and does not depend on what the remote side's broadcast does: `routeNotification` sent to `sub.in` synchronously inside the single per-client `drainLoop`, and the only escape was `sub.pumpDone`, which closes only when the pump exits. A consumer that stops reading `out` blocks its pump on `out <-`; `sub.in`'s 128 slots fill; the shared drain then blocks on `sub.in <-` for every thread on the host; and because the blocked drain stops reading `client.Notifications()`, the client's frame buffer overflows into a connection teardown that kills every subscription on that host. My earlier "Broadcast is non-blocking" argument addressed a different link and does not refute this one.

Fix (option (a)): the hand-off keeps a buffered fast path for the healthy case, and a blocked hand-off now waits at most `remoteHubRouteStallTimeout` (5s — far longer than jitter the 128-slot buffer absorbs) before retiring the stalled subscription. Its pump exits, `out` closes, and the controller relay treats that as subscription end and re-reads the thread — a resync, not a silent drop. A merely jittery consumer is still served from the buffer and is never resynced.

Two things this surfaced that are also fixed:
- A subscribe request cancelled by the owning client's notification stream closing (the new drain retirement, not the caller) is now mapped to `SessionUnavailable` instead of surfacing as `context.Canceled`, so a disconnect during the initial attach still reads as a transport loss to the auto-resume gate.
- The pushable-remote fake now answers `thread/unsubscribe` centrally, as the real server does, so scenario handlers never see framework teardown.

Tests:
- `TestRemoteHubRouteStallRetiresSubscriptionAndKeepsOthersFlowing` — one subscription's consumer stalls (its `out` and `in` are both filled), a second subscription on the same client still receives, and the stalled one's `out` closes. Fails with the bounded wait disabled ("timed out waiting for a notification").
- `TestRemoteHubRouteJitterWithinBufferIsDelivered` — a burst that fits the buffer is delivered in full and the subscription stays live (no spurious resync).

Verification: `gofmt -l cmd/evener-hub/internal/appsource/` empty; `go build ./...` ok; `go vet ./cmd/evener-hub/...` ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` ok (plus the racy tests 20x under `-race`); `go test ./cmd/evener-hub/... -count=1` ok (13 packages); module-lint `PASS`, 0 findings.

obra:
Round two addressed (`5573f4a57`). 1 High + 4 Medium + 1 Low.

- **High — `ReadThread` forwarded `ReplaceSubscription` to the shared remote connection.** Confirmed. `ReadThread` copied `params` verbatim, so a controller read carrying replacement semantics reached the shared per-host client, and the remote server's `CaptureSubscription(replace=true)` → `ReplaceConnectionSubscriptions(conn.id, …)` scoped that whole connection to the one thread — dropping every other remote thread's subscription on the host while the local routing entries stayed live. Fixed: `ReadThread` clears the flag, as `SubscribeThread` already did; replacement stays a controller-side concept applied by `app_relay.go`. Pinned by `TestRemoteHubReadThreadDoesNotForwardReplaceSubscription` (pre-fix it forwards `"replaceSubscription":true`).
- **Medium — `routeNotification` did not verify client identity.** Confirmed, both halves. The drain is per client, but a torn-down client can still have frames buffered in its `Notifications()` channel when it closes, and that drain runs concurrently with the relay re-subscribing against the replacement client. Fixed: `routeNotification` takes the arrival client and drops any notification whose installed subscription is bound to a different client. Pinned by `TestRemoteHubRouteNotificationIgnoresRetiredClient`: a stale frame is neither delivered nor buffered, and it is not mistaken for a stalled consumer (pre-fix it was delivered, and with a full buffer it cancelled the healthy replacement subscription).
- **Medium — the shared drain blocked up to 5 s on one stalled subscription.** Confirmed; this supersedes the round-one bounded wait. The wait was still a window in which the single per-client drain stopped reading `client.Notifications()`, and `appwire` tears the whole connection down when that 4096-slot buffer fills — losing every subscription on the host. Fixed by removing the wait entirely: the hand-off is non-blocking, and a subscription whose `in` buffer is full (a consumer at least 2×128 notifications behind, far past the jitter those buffers absorb) is retired, closing `out` so the relay resyncs. `remoteHubRouteStallTimeout` is gone. Pinned by `TestRemoteHubStalledSubscriptionCannotOverflowSharedClient` (pushes `NotificationBufferCap`+1 frames while one consumer stalls; pre-fix the client tore down at push 4083) and `TestRemoteHubDiscardSubscriberUnblocksRouter` (the router returns instead of parking; pre-fix it blocked to the 5 s stall timer). `TestRemoteHubRouteJitterWithinBufferIsDelivered` still holds: a burst that fits the buffer is delivered in full, no resync.
- **Medium — orphaned remote subscription on context cancellation.** Confirmed as a class: when a subscribe request reaches the server and its response is then lost (caller cancellation, disconnect), the failure path removed only local state and the long-lived shared client kept forwarding that ref. Fixed: the discard sends a best-effort `thread/unsubscribe` for the failed ref, bounded by the existing 2 s timeout, unless a live subscription still owns that ref *on the same client* (the remote subscription is per connection+thread, so a sibling's ref is the sibling's to keep). Pinned by `TestRemoteHubSubscribeThreadFailedInstallUnsubscribesRemote`; the negative control is the new assertion in `TestRemoteHubSubscribeThreadFailedReplacementPreservesPrevious` that no `thread/unsubscribe` is sent while the restored previous still owns the ref. **Honest gap:** the refused-reply shape is what makes the pin deterministic; the exact cancelled-response trigger is not deterministically reproducible in this harness, because the fake's serial server loop cannot both withhold the response and answer the cleanup request.
- **Medium — `discardSubscriber` unconditionally restored a dead or disconnected `previous`.** Confirmed. A displaced subscription whose pump had exited (relay context cancelled, drain retired it as stalled) or whose client's stream had ended was put back in the routing table, where nothing would pump or retire it (leaking the remote-side subscription), and a relay blocked on its `out` hung forever when its client was gone. Fixed: `discardSubscriber` restores only a subscription that is still live (pump running and its client still being drained); otherwise it deletes the entry and cancels the displaced subscription so the relay unwinds. Pinned by `TestRemoteHubDiscardSubscriberDoesNotRestoreDeadPrevious` (pump-exited, client-no-longer-drained, and a live-previous-still-restored control; the first two fail pre-fix).
- **Low — stale struct doc comment.** Confirmed: it still called the subscription fan-out staged. Updated to list `SubscribeThread` and the fan-out as implemented, with only 05c/05d staged.

No assertions were weakened or removed; `TestRemoteHubRouteStallRetiresSubscriptionAndKeepsOthersFlowing` and `TestRemoteHubRouteJitterWithinBufferIsDelivered` keep their assertions and only lost the stall-timeout shim.

Verification: `gofmt -l cmd/evener-hub/internal/appsource/` → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 30m` → ok; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (13 packages, main package 119 s); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS, 0 findings (1 module, 40 s). No fuzz targets added or touched, so `make fuzz-gap-check` / `make fuzz-registry-check` were not required.

obra:
Round three addressed (`11dd25c2f`). 5 findings: 4 fixed, 1 contested. (The verdict speaks of "three correctness/liveness defects"; the review body lists five — one High, three Medium, one Low.)

- **High — stable-ref addressing breaks remote reads/subscriptions.** Confirmed. `appReadTargetLocked` (`server/appwire_runtime.go:1378-1396`) resolves a bare non-empty `ThreadID` *in preference to* `Ref`, so forwarding the translated ref's suffix made the remote treat the stable identity as a bare thread ID; after `ReplaceAppIdentity` moves `appThreadID` while `appRef` survives, `threadID != rootID && descendant == nil` refuses the read/subscribe. Fixed in `SubscribeThread`, `ReadThread` and `ListTurns`: the caller's `ThreadID` is forwarded verbatim, empty included, so the remote resolves the (stable-aware) ref; `sub.threadID` still keys local routing on the ref suffix. Pinned by `TestRemoteHubSubscribeThreadForwardsCallerThreadID` (ref-only case fails pre-fix: forwarded `threadId="stable"`), `TestRemoteHubReadAndListTurnsForwardCallerThreadID`, and corrected assertions in `TestRemoteHubSubscribeThreadWireContract` / `TestRemoteHubSourceReadThread`. `ListTurns` was not named in the finding but carries the identical line; fixing it is the only work I did beyond the finding, and I flag it here.
- **Medium — double subscription on the non-atomic relay path.** Confirmed. `relays.readThread` forwards the client's params to `ReadThread` (`app_relay.go:1191`), and the remote registry is a set keyed by (connection, thread) (`internal/appserver/subscriptions.go:40-57`), so the read attached the shared connection to the thread with no local routing entry behind it. The handler always starts the relay (`params.Subscribe || relayOnThreadRead(source)`, `app_rpc.go:559`; `RemoteHubSource` has no `RelayOnThreadRead` override, so it defaults true) and that is what subscribes. If the relay never attaches (`thread.ID == ""`, deletion fence, failed subscribe) the remote subscription is never unsubscribed and the long-lived connection keeps forwarding into a stream nothing drains. Fixed: `ReadThread` clears `Subscribe`; no delivery is lost because `SubscribeThread` installs the local routing entry before it subscribes. Pinned by `TestRemoteHubReadThreadDoesNotForwardReplaceSubscription`, whose round-two assertion ("only replacement semantics must be cleared") is **inverted, not deleted**: the forwarded read must now carry neither flag. That inversion corrects the pinned contract rather than weakening it, and it and the two threadId assertions above are the only existing assertions I changed.
- **Medium — send-lock acquisition not context-cancellable. CONTESTED.** The 2 s timeout does bound everything this component controls: `appwire.Client.request` waits for the response in a `select` on `ctx.Done()` (`appwire/client.go:252-256`). The unbounded parts are `c.sendMu.Lock()` (`appwire/client.go:243`) and `StreamTransport.Send`'s `wmu.Lock()` + blocking `conn.Write` (`appwire/stream_transport.go:66-68`), neither of which observes the context; both live in `appwire`, which PR #1291 does not touch (its files are the four `.../internal/appsource/remote_hub_*.go`). The suggested in-scope remedy — "perform cleanup without holding the routing mutex" — is not available: `remoteMu` is exactly what orders the predecessor's `thread/unsubscribe` before a replacement's `thread/read{subscribe:true}`, and because the remote registry is a set keyed by (connection, thread), an unsubscribe landing after a replacement's subscribe removes the replacement's entry outright — the silent-drop defect round two fixed. Dropping the lock (or moving the send to a goroutine) reintroduces that race, and the async variant is worse: a late unsubscribe can drop a *newer, still-live* subscription. Bounding the acquisition itself requires `appwire.Client.Request` / `StreamTransport.Send` to honour the context; that is a separate appwire change and I recommend it as its own PR.
- **Medium — `retireSubscription` leaks remote subscription on client replacement.** Confirmed. Retiring skipped the remote `thread/unsubscribe` whenever the routing entry had changed, but a replacement bound to a different `*appwire.Client` cannot own this connection's remote subscription. Fixed by mirroring `remoteRefOwnedLocked`: skip only when another subscription still holds the same remote ref on the same client. Pinned by `TestRemoteHubRetireSubscriptionDropsRemoteRefWhenReplacementUsesAnotherClient` (pre-fix: no `thread/unsubscribe` on the displaced connection), with the same-client control `TestRemoteHubRetireSubscriptionKeepsRemoteRefForSameClientReplacement`.
- **Low — JSON `null` thread rewritten to `{"source":"host"}`.** Confirmed. `translateThreadRaw` decoded `null` into a nil map and then materialized an empty object and stamped a source on it. Fixed: a nil decoded map returns `raw` untouched. Pinned by `TestRemoteHubPreservesNullNestedThread` (pre-fix: `nested thread = {"source":"host"}`).

No fuzz targets were added or touched.

Verification:
- `gofmt -l cmd/evener-hub/internal/appsource/` → empty
- `go build ./...` → ok
- `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 30m` → ok (12.6 s)
- `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (13 packages; main package 113 s)
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS, 0 findings (1 module, 44 s)

obra:
Round four addressed (`80d901c`).

Review 1's four findings, each verified against the code and resolved (the other three reviewers found it clean):

- **HIGH — atomic snapshot discarded on subscribe handoff** (`remote_hub_subscription.go`): real. The controller relay's non-atomic `prepareRelay` branch reads the thread and only then subscribes, and the remote subscribed read's atomic snapshot was dropped — so a delta in that window was in neither the controller snapshot nor the notification stream, and the relay stayed stale with nothing to resync it. `settleSubscriber` now turns the subscription snapshot into a leading `evener/thread/resync` frame on the relay's channel (written before the pump starts, so it is the consumer's first frame), and the controller relay answers a resync by re-reading the thread, folding the snapshot in. The alternative — an atomic handoff through `ReadThread` — is ruled out by an existing disposition: Test 22 pins that the snapshot read must not create a remote subscription nothing local owns. The resync is emitted whenever the snapshot carries the remote's thread, which a real hub's subscribed read always returns; a snapshot with no thread at all is a degenerate attach with no authoritative identity to key on and nothing to fold in, so the caller-derived key stands and the controller's copy is left alone. Pinned by `TestRemoteHubSubscribeThreadResyncsFromSnapshot`; `TestRemoteHubSubscribeThreadWireContract` was updated to assert the leading resync (the snapshot-leak and pushed-notification checks are unchanged).

- **HIGH — mixed `Ref`/`ThreadID` routing mismatch** (`remote_hub_refs.go`, `remote_hub_subscription.go`): real. The remote hub resolves both `thread/read`'s relay and `thread/unsubscribe` through `threadRelayTarget`, which prefers a bare non-empty `threadId` over the ref. So the remote subscription lives under the thread the remote resolved while routing was keyed from the caller's ref suffix — child notifications were dropped, and teardown unsubscribed a ref the remote never subscribed. Routing is now keyed from the subscription snapshot's own ref (`rekeySubscriber` re-keys the entry before the pump starts), and teardown names the identity the subscribe created (`remoteSubscriptionTarget`: the caller's `threadId` when it sent one, else the translated ref's suffix). This also closes a latent leak in the stable-ref + live-instance case, where the remote relay lives under the instance id while teardown previously sent the stable ref. Pinned by `TestRemoteHubSubscribeThreadRoutesBySnapshotIdentity`.

- **MEDIUM — nested ref translation incomplete** (`remote_hub_refs.go`, `remote_hub_subscription.go`): real. Only the top-level `ref` and `Thread.Evener.{Ref,ParentRef}` were rewritten; `transcriptRef`/`childRef`/`ownerRef`/`sessionRef` nested in diagnostics, job, and delegate payloads stayed `local:*`, so the controller UI routed remote child sessions to its local source. A field-aware JSON walk (`translateNestedRefs`) now translates every such handle in notifications, and `fromRemoteThread` does the typed equivalent for read/list responses. Values that are not the remote hub's own `local:<thread>` handle are passed through byte-for-byte — opaque `job:<id>` and `proj:<project>:<thread>` handles keep working, and the walk deliberately does not descend into arbitrary model/tool JSON (`turns`, `raw`, `message`, `structuredResult`), where a coincidentally named field is a tool argument, not a session handle. The walk never fails, so it cannot drop a notification. Pinned by `TestRemoteHubNestedSessionRefsTranslated`, `TestRemoteHubNestedOpaqueRefsPreserved`, `TestRemoteHubThreadDiagnosticsRefsTranslated`, and `TestRemoteHubSourceReadThreadTranslatesNestedDiagnostics`.

- **MEDIUM — cancellation race between subscribe and unsubscribe** (`remote_hub_subscription.go`): real. A caller cancelling after the subscribe was sent made `client.Request` return at once, and cleanup unsubscribed immediately; the remote handles `thread/read` and `thread/unsubscribe` concurrently, so the unsubscribe could land first and leave the later subscribe active with no local owner. The subscribe request now runs on a context detached from the caller's (`requestSubscribe`, bounded by `remoteHubSubscribeTimeout` so a wedged remote cannot leak the goroutine), so `SubscribeThread` still returns promptly on cancel, while cleanup (`retireCanceledSubscribe`) waits for the subscribe to complete before unsubscribing. Pinned by `TestRemoteHubSubscribeCancelWaitsForSubscribeBeforeUnsubscribe`, where the withheld reply makes the ordering observable.

No round-two or round-three disposition is superseded. The shared-drain fan-out and overflow retirement, the failed-install remote unsubscribe and its sibling-ownership rule, the replacement/previous restoration rules, the deliberate `replaceSubscription` stripping, and the contested appwire send-lock item are all unchanged, and every existing assertion still holds (the single test update is the leading-resync assertion noted above).

Verification: `gofmt -l` over the touched package → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` → ok; `go test ./cmd/evener-hub/... -count=1` → all 13 packages ok; `golangci-lint` module-lint → `PASS (1 modules, 17s)`, 0 findings.

obra:
Round five addressed (`6940e3545`).

- **Stable-ref relay registration mismatch** (`remote_hub_source.go:168`) — real, and fixed in the hub relay. `startRelay` keyed the non-atomic relay by `thread.ID` (`host:<currentID>`) while `thread/unsubscribe` resolves the canonical ref suffix via `threadRelayTarget` (`host:<stableRef>`). A remote thread's stable ref survives an identity replacement while its `Thread.ID` moves, so a ref-only read registered under a key a ref-only unsubscribe never matched: the downstream entry (and, through it, the remote-side subscription) stayed live, and a replacement could leave a second relay behind. `startRelay` now derives its key from `relayTarget(source, subscribeParams)` — the same identity the atomic path (`acquireRelaySession`) and `thread/unsubscribe` already use, matching the spec's `relayKey "host:S"`. Pinned by `TestHubRPCThreadUnsubscribeDropsStableRefRelay` (fails pre-fix: count at the stable ref is 0). *Scope note: this is `cmd/evener-hub/app_relay.go`, outside the appsource file set; the reviewer's own fix direction (register/unsubscribe consistently on the canonical ref) lives there, and no sibling multi-host branch touches it.*

- **Duplicate resync delivery** (`remote_hub_subscription.go:259-263`) — real, and this supersedes the round-four disposition in one specific way. Round four made the source inject the atomic-snapshot resync because the relay's non-atomic initial attach emits none; that is still correct and retained. But the relay's recovery loop also emits its own resync, so a source-supplied leading resync was forwarded *in addition*, producing two resyncs and two redundant thread reads per recovery. The recovery loop now peeks the replacement subscription's first frame and, when it is already `evener/thread/resync`, gives the source's frame precedence and skips the relay's own (the source's carries the subscription's authoritative identity; the relay's names the original `subscribeParams`). Pinned by `TestHubRelayRecoveryDoesNotDuplicateSourceResync` (fails pre-fix: first frame is the relay's `codex:th_resync_once`, and a second resync follows).

- **Concurrent replacement leaves a silent subscription** (`remote_hub_subscription.go:157`, `:296-298`) — real. `installSubscriber` deliberately does not cancel what it displaces, so a replacement can win the routing slot while an earlier subscribe is still in flight. `rekeySubscriber` returned without installing, but `settleSubscriber` ignored that and `SubscribeThread` started the pump and returned `sub.out`: a live-looking channel that nothing routes to. Settlement now reports whether the subscription is installed (`rekeySubscriber` returns an `installed` bool; a no-rekey settle re-checks the slot) and `SubscribeThread` discards a displaced subscription and returns `SessionUnavailable` instead of a channel. Pinned by `TestRemoteHubSubscribeThreadRejectsDisplacedSettle` (both the same-key and re-key variants; fails pre-fix: returns a channel for a displaced subscription).

- **Nested sandbox-escalation refs left in the remote namespace** (`remote_hub_refs.go:91`, `:191`) — real, latent as the reviewer notes. `PendingEscalations[].Ref` is the session a card routes by; it is now translated on both paths: `fromRemoteThread` rewrites each entry's `Ref` (mirroring the diagnostics loops), and `translateThreadRaw` translates the `pendingEscalations` array structurally — deliberately *not* by adding `"ref"` to `nestedSessionRefFields`, keeping the walk to known structural locations. Opaque handles (`job:*`) are left untouched. Pinned by `TestRemoteHubSourceReadThreadTranslatesPendingEscalationRefs` and `TestRemoteHubSubscribeThreadTranslatesPendingEscalationRefs` (both fail pre-fix with `local:child`).

Verification: `gofmt -l cmd/evener-hub cmd/evener-hub/internal/appsource` → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` → ok 22.7s; `go test ./cmd/evener-hub/... -count=1` → ok (13 packages, main pkg 183.0s); `MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS lint (1 modules, 105s), 0 findings.

obra:
Round six addressed (`b67658e`).

- **`remote_hub_subscription.go:82-90, 427-434` — a subscription's remote identity is now the canonical ref the successful subscribe snapshot named, not the caller's `ThreadID`.** `remoteRef` was derived from the caller's `ThreadID` via `remoteSubscriptionTarget` and compared literally in `remoteRefOwnedLocked`. The remote canonicalizes a current root thread ID to its stable ref, so a replacement created with a different (moved) current ID recorded a different `remoteRef`: its displaced predecessor believed the ref was unowned, unsubscribed the live replacement's remote subscription, and named a stale ref on teardown. `settleSubscriber` now records `snapshot.Thread.Evener.Ref` as the subscription's canonical remote identity under `remoteMu`+`subMu`, so a retiring predecessor observes it and ownership/teardown compare the identity the remote actually keyed. An empty snapshot ref keeps the caller-derived provisional target. Pinned by `TestRemoteHubSubscribeThreadCanonicalizesRemoteRefFromSnapshot` (teardown must name `local:stable`, not `local:current`) and `TestRemoteHubRetireKeepsCanonicalRemoteRefForSameClientReplacement` (a same-client replacement created with a moved current ID keeps its remote subscription). Both fail on the pre-fix code.

- **`app_relay.go:1380` — a federated source's relay is now keyed by the ref the request carried, matching the ref-only `thread/unsubscribe` that drops it.** `startRelay` keyed by `threadRelayTarget`, which prefers an explicit `ThreadID`; a read carrying both the stable ref and the thread's current ID registered under `host:<currentID>` while `{Ref:"host:<stableRef>"}` resolved `host:<stableRef>`, so the downstream entry (and the source-side subscription behind it) could never be removed. New `relayDeliveryTarget` prefers the ref for any non-`LocalDaemonSource` source and is used by `startRelay`, by thread/read's subscription admission (`app_rpc.go`), and by the `thread/unsubscribe` handler, so all three agree; `startRelay` keys from the request's own address (falling back to the ref derived from the read for internal callers like turn/start). The local daemon path is unchanged: its admission already canonicalizes through `ResolveSubscriptionAdmission`, and its `SecondaryKey` deliberately names the `threadRelayTarget` delivery identity. Pinned by `TestHubRPCThreadUnsubscribeDropsStableRefRelayWhenReadAlsoNamedThreadID`; the round-five `TestHubRPCThreadUnsubscribeDropsStableRefRelay` still passes.

- **`remote_hub_subscription.go:110-120` — a relay's recovery now re-addresses the thread by the authoritative ref the read named, not by the caller's original bare `ThreadID`.** The remote hub's daemon edge (`server/appwire_runtime.go` `appReadTargetLocked`) resolves a bare `threadId` in preference to the ref beside it, so once an identity replacement moves the live thread ID the original ID names nothing while the still-valid stable ref would have resolved, and every recovery attempt in the stale state fails. Recovery re-subscribes with `{Ref: thread.Evener.Ref, ThreadID: ""}`; the initial request still forwards the caller's own addressing, so the round-three/round-five forwarding dispositions are intact. Pinned by `TestHubRelayRecoveryUsesAuthoritativeRefNotStaleThreadID`.

Verification: `gofmt -l` over the touched dirs → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/ -race -count=1 -timeout 40m` → ok 419.5s; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` → ok 20.4s; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (hub 181.7s, all subpackages ok); module-lint → PASS (1 module, 0 findings).

obra:
Round seven addressed (`edb666ef6`). 1 High (three interleavings) + 2 Medium, all confirmed and fixed.

- **HIGH — displaced/canonicalize ordering race could leave a replacement subscription permanently silent.** Confirmed, all three interleavings. The root cause is what round six left: `remoteRef` is a caller-derived provisional target until settlement records the canonical ref, and ownership/teardown compared raw `remoteRef` values, so a same-thread replacement that had not yet settled looked like it owned nothing.
  - *(a) re-key path:* `settleSubscriber` cancelled the displaced subscription before recording the canonical ref, and the displaced pump's `retireSubscription` could then run with stale ownership. `settleSubscriber` now performs the re-key, the canonical-ref write, the `settled` mark and the installed check under one `remoteMu`+`subMu` hold, and only cancels the displaced subscription afterwards; a re-check after the cancel returns `false` rather than publishing a subscription a concurrent install displaced. Pinned by `TestRemoteHubReplacedSettleKeepsAdoptedRemoteRef` (fails pre-fix: the displaced predecessor unsubscribes `local:child`, the ref the re-keyed replacement just adopted) and `TestRemoteHubSettleRejectsReplacementInstalledDuringSettlement` (fails pre-fix: returns `true` and writes a resync for a displaced subscription).
  - *(b) install window:* a predecessor retiring while a same-thread replacement was still attaching unsubscribed the ref that replacement was about to adopt. Ownership now recognizes the link the provisional refs carry: subscriptions retain `provisionalRef` (the caller-derived identity), and a still-attaching sibling that shares it is a *claimant*. A retiring/failed subscription defers the ref to that claimant (`deferred`) instead of unsubscribing; the claimant releases the deferred ref when it settles (adopting it, or unsubscribing it if it settled elsewhere) or fails, so nothing is dropped and nothing leaks. Pinned by `TestRemoteHubRetireDefersToStillAttachingSameThreadReplacement` (fails pre-fix: predecessor unsubscribes `local:stable`).
  - *(c) settle after displacement:* covered by the atomic settle plus the post-cancel installed re-check above (same test `TestRemoteHubSettleRejectsReplacementInstalledDuringSettlement`).
- **MEDIUM — canceled in-flight subscriptions leaked the remote subscription.** Confirmed. `retireCanceledSubscribe` waited for the subscribe but ignored its snapshot, so cleanup unsubscribed the caller's provisional target; when the remote canonicalized a current root thread ID to its stable ref, the unsubscribe named a subscription the remote never held and the real one leaked. Cleanup now canonicalizes `remoteRef` from the completed snapshot before releasing. Pinned by `TestRemoteHubCanceledSubscribeUnsubscribesCanonicalRef` (fails pre-fix: unsubscribes `local:current`, not `local:stable`).
- **MEDIUM — failed replacement dropped buffered in-flight notifications.** Confirmed. A replacement takes over routing before its subscribe succeeds; on failure `discardSubscriber` restored the previous subscription but dropped the replacement's buffered `in`, losing the deltas with no resync. The discard now forwards the replacement's buffered notifications to the restored previous (non-blocking; a full previous buffer retires it so the relay resyncs). Pinned by `TestRemoteHubFailedReplacementForwardsBufferedNotifications` (fails pre-fix: the delta is lost).

No earlier disposition is superseded: round six's canonicalization is extended, not undone, and the atomic-snapshot resync handoff, relayTarget stable-ref keying, the shared-drain/overflow retirement, displaced-subscription rejection, the failed-install remote unsubscribe and its sibling-ownership rule, the replacement/previous restoration rules, and nested escalation-ref translation are all intact. No existing assertion was weakened or deleted; the only production file changed is `remote_hub_subscription.go`.

Verification: `gofmt -l` over the touched dirs → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` → ok 28.3s; `go test ./cmd/evener-hub/ -race -count=1 -timeout 45m` → ok 553.0s; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (13 packages); module-lint → PASS (1 module, 0 findings).

obra:
Round eight addressed (`b585e663a`).

- **Timeout only cancelled the local wait; the remote could still install after cleanup unsubscribed it** (`remote_hub_subscription.go:236-261`). The attach request no longer carries a local deadline at all: it is settled by the remote's own answer or by the connection ending (appwire fails a pending request when its read loop exits, and the remote hub reaps a connection's subscriptions with the connection). Cleanup's release is driven only by that outcome, so it can no longer be issued while the remote is still working the request. The request is also minimal now — `IncludeTurns`, `ItemsView` and `ItemLimit` are no longer forwarded, because the snapshot's thread identity is all this source reads from it (routing key + canonical remote ref; the leading resync makes the relay re-read the turns itself) and the turn page only widened the install-to-answer window. This supersedes round seven's `remoteHubSubscribeTimeout` disposition ("a remote that never answers cannot leak the request goroutine forever"): appwire has no protocol-level request cancellation, so a local timer can only abandon the wait, never the remote subscription — a wedged remote now holds one cleanup goroutine and one routing entry until its connection dies, which is exactly the window in which a still-attaching subscription must keep its routing slot so a predecessor's unsubscribe stays deferred to it. Pinned by `TestRemoteHubAttachRequestHasNoLocalDeadline` (the request is sent on a context with no deadline; fails on `edb666e`), `TestRemoteHubSubscribeThreadAttachRequestOmitsTurnPayload` (fails on `edb666e`), `TestRemoteHubSubscribeCancelWaitsForSubscribeBeforeUnsubscribe` (nothing is released while the subscribe is in flight) and `TestRemoteHubSubscribeCancelReleasesWhenConnectionEnds` (the connection's end is what settles an abandoned wait and runs cleanup). Honest gap: the 30-second deadline path itself is not reachable in-process without a 30-second wait; the fix removes it rather than testing around it.

- **A displaced successful subscribe did not record the snapshot's canonical remote ref before cleanup** (`remote_hub_subscription.go:341-350`). `settleSubscriber`'s displaced-settlement return now records the snapshot's canonical ref under the same remoteMu+subMu hold, before reporting the displacement — including when settlement is rejected. Cleanup therefore releases the identity the remote actually keyed instead of the caller-derived provisional target, which the remote canonicalizes away for a current root thread ID. Pinned by `TestRemoteHubDisplacedSettleUnsubscribesCanonicalRef`, which stages a subscribe to `{Ref: host:stable, ThreadID: current}` displaced mid-flight: on `edb666e` it sent `thread/unsubscribe {"ref":"local:current"}` and left `local:stable` live; with the fix it releases `local:stable` and never touches `local:current`.

- **`discardSubscriber` computed `restore` before checking that `sub` still owned the routing slot** (`remote_hub_subscription.go:504-521`). `restore` is now gated on `s.subs[sub.threadID] == sub`. When a third subscription installed itself at the same key while the failing replacement was in flight, the displaced predecessor was neither restored nor cancelled — it lingered with no routing entry, its pump blocked forever, its `out` never closed and its controller relay never observed subscription end. It is now retired like any other displaced subscription (`previous.cancel()`), while its remote ref stays with the live third subscription that holds it. Pinned by `TestRemoteHubDiscardSubscriberRetiresUnrestorablePrevious` (initial live subscription, failing replacement with a parked read, concurrent third install that settles), which asserts the predecessor's `out` closes, that the third subscription keeps delivering, and that no `thread/unsubscribe` names the ref the third subscription holds; on `edb666e` it fails with "the displaced predecessor's out never closed".

Verification: `gofmt -l cmd/evener-hub/internal/appsource/ cmd/evener-hub/` → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 20m` → ok (21.7s); `go test ./cmd/evener-hub/ -race -count=1 -timeout 45m` → ok (561.7s); `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (all packages); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS, 0 findings.

Also fixed a pre-existing synchronization race in `TestRemoteHubFailedReplacementForwardsBufferedNotifications` (this PR's test file): its `waitForRemoteSubscribeCall` returned on the live subscription's own subscribed read, so the delta could be pushed before the replacement had installed (9/30 failures on `edb666e` under `-race`). It now waits for the replacement's own call via `waitForRemoteSubscribeCalls`; 40/40 clean.

Commits: `4e371b861` (the three fixes and their four pins) and `b585e663a` (the connection-end pin cited above).

obra:
Round nine addressed (`ff5a363`).

- **High — unsettled replacement with a different volatile `ThreadID` can permanently silence a subscription** (`remote_hub_subscription.go`, `refClaimLocked`): verified. A retiring predecessor recognized a still-attaching replacement only when the two shared a caller addressing (or the replacement already named the ref being released). Several aliases of one remote thread — a current root ID before and after an identity replacement, or a current ID and the stable ref — resolve onto one canonical remote ref, so with different volatile IDs the predecessor found no claimant and unsubscribed that canonical ref; the replacement then stayed open and silent, and because its own `out` never closes nothing ever re-attached it. Fixed: a still-attaching sibling on the same connection that recorded a caller addressing is now treated as a possible claimant, so the ref is deferred rather than released, and it is resolved at that sibling's settle — its own if it adopts it, unsubscribed if it does not. Pinned by `TestRemoteHubRetireDefersToAliasedStillAttachingReplacement` (fails pre-fix with "predecessor unsubscribed the canonical ref its aliased replacement will adopt") and `TestRemoteHubAliasedDeferralReleasedWhenNotAdopted` (the deferral is not a leak), both with distinct volatile thread IDs.

- **Medium — a notification can be silently lost between the failed-replacement buffer hand-off and a concurrent route** (`routeNotification` / `discardSubscriber`): verified as described. `routeNotification` captured the subscriber under `subMu`, released it, and only then sent; `discardSubscriber` swapped the table back and forwarded the buffer. A send landing in that window went into the replaced subscription's buffer, which no pump reads, while the restored `previous` stayed live so nothing re-read the thread. Fixed: the capture and the non-blocking send are one `subMu` critical section, and `discardSubscriber` holds `subMu` across the table swap plus `forwardBufferedNotifications` (the `previous.cancel()` and `default: sub.cancel()` paths stay outside the lock, and the bounded wire unsubscribe still runs with `subMu` released). Pinned by `TestRemoteHubFailedReplacementHandoffCannotLoseConcurrentRoute`: it parks the hand-off on a test-only ordering seam (`remoteHubSubscription.onBufferedHandoff`, documented as production-never-sets), asserts `subMu` is held at that point (fails pre-fix with "the buffered hand-off released subMu"), then routes a delta concurrently and asserts it reaches the restored previous.

- **Medium — two controller relays for the same remote thread addressed by different refs displace each other indefinitely**: verified. `relayDeliveryTarget` keys a federated relay by the request's ref suffix, so `host:current` and `host:stable` create two relays for one remote thread; both subscribe with the canonical ref (`recoveryParams`), the source's routing slot is one per remote thread, so each successful attach cancelled the other's subscription and each recovery re-attached with the same canonical ref — unbounded, because a successful recovery resets the backoff. Fixed source-side with the relay identity the finding names: the hub now labels every attach it issues with its relay key (`appsource.WithRelayIdentity(ctx, relayKey)`, one line in `startRelay`), and `SubscribeThread` refuses to displace a **live** subscription attached for a **different** relay, before any request reaches the remote. The incumbent keeps serving; the refused relay retries on its own backoff (each refusal is local — no remote round trips) and attaches as soon as the incumbent ends. Same-relay re-attaches keep the existing replacement semantics, so relay recovery, the failed-replacement restore path, and the displaced-settle rejection are unchanged; no teardown-order change was needed, because the gate is scoped by relay identity rather than by liveness alone. Pinned by `TestRemoteHubSubscribeThreadRefusesSecondRelayForSameThread` (fails pre-fix), `TestRemoteHubSubscribeThreadAdmitsSameRelayReattach`, `TestRemoteHubSubscribeThreadSecondRelayAttachesAfterIncumbentEnds`, and the wiring by `TestHubRelayLabelsAttachesWithItsRelayKey` (fails with that `startRelay` line reverted).

  Scope note, as required: `cmd/evener-hub/app_relay.go` is touched, for that single wiring line. The refusal is a bounded outcome, not sharing — the second relay is refused (its read fails with a conflict naming the ref that already serves the thread, or its recovery retries on backoff) instead of being served by the incumbent's subscription. Serving both relays from one subscription is the fuller fix (canonical relay identity plus fan-out for federated sources, mirroring what the local daemon's lease/`canonicalRelays` path already does) and needs relay-registry changes beyond this PR's file set.

Verification: `gofmt -l` clean; `go build ./...` exit 0; `go vet ./cmd/evener-hub/...` exit 0; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 20m` ok (21.6s); `go test ./cmd/evener-hub/ -race -count=1 -timeout 45m` ok (414.8s, and ok again 518.0s on the pushed tree); `go test ./cmd/evener-hub/... -count=1 -timeout 30m` ok (cmd/evener-hub 168.5s, appsource 4.8s, hubcore 13.4s, all subpackages ok); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → `PASS lint (1 modules, 118s)`, exit 0.

obra:
Round ten addressed (`e67329e`).

- **Alias-based admission bypassed foreign-relay refusal** (`remote_hub_subscription.go`): `admitSubscriber` compared only the provisional routing key derived from the caller's ref suffix, so two relays addressing one remote thread through different alias refs (`host:stable` vs `host:current`) never collided; the newcomer's settlement then re-keyed onto the canonical identity, cancelled the live incumbent, and the incumbent's recovery was refused in turn — starving its client until the newcomer ended. `settleSubscriber` now runs the foreign-relay check against the canonical identity the snapshot names, **before** committing the re-key: it returns the incumbent's `Conflict`, which `SubscribeThread` surfaces to the relay instead of the generic "replaced before it attached" loss, and leaves the incumbent installed, routed, and still receiving (the newcomer records the canonical ref it would have adopted, so its cleanup claims the ref through the incumbent instead of unsubscribing it). **Fix shape chosen:** the settle-time canonical check, not admission-time canonical resolution — the two aliases are only unified by the remote hub's own resolution, which the controller cannot see before subscribing (the snapshot is the only place the canonical identity appears, and it is also the atomic attach point the handoff relies on); this extends round nine's single atomic settle rather than adding a second identity path. The refusal is identity-based, so the same relay's aliased re-attach (its recovery/reconnect) still replaces the incumbent. Pinned by `TestRemoteHubAliasedForeignRelayRefusedWithoutDisplacingIncumbent` (fails on `ff5a363`: refused error is nil and the incumbent is displaced/closed) with `TestRemoteHubAliasedSameRelayReattachStillReplaces` as the replace-semantics control.

- **Canceled pending subscription preserved as "live" and restorable** (`remote_hub_subscription.go:274-285, 723-740`): a subscription whose caller gave up while its subscribe was still in flight stayed in `s.subs` with no pump ever started and `pumpDone` still open, so `subscriberLiveLocked` reported it live. A failed replacement could therefore restore it as the routing target (handing its buffered deltas into a buffer no pump will ever read), and the next relay's admission was refused with a `Conflict` on behalf of a dead entry. Subscriptions now carry their own context and `subscriberLiveLocked` requires it to be alive (`ctx.Err() == nil`), alongside the open-`pumpDone` and live-drain checks, so only genuinely live predecessors are restored. The outstanding remote request keeps its separate cleanup: the abandoned entry still stays in the table until its detached request answers and is released by `retireCanceledSubscribe`/`discardSubscriber`, exactly as round nine left it. Pinned by `TestRemoteHubCanceledPendingSubscriberIsNotRestorable` (fails on `ff5a363`: the dead entry is restored and a second relay is refused admission by it).

Prior dispositions kept: single-drain fan-out with no head-of-line blocking, per-subscription bounded read-ahead and overflow isolation (round seven), atomic identity settle and deferred predecessor unsubscribes (round nine), host-keyed identity canonicalization, and notification translation. No existing assertion was weakened or removed; the two hand-built "live predecessor" test literals (`TestRemoteHubFailedReplacementHandoffCannotLoseConcurrentRoute`'s predecessor and `TestRemoteHubDiscardSubscriberDoesNotRestoreDeadPrevious`'s "live previous restored" case) now carry `ctx: context.Background()`, because a live subscription must have a live context (the production constructor always sets one).

Verification: `gofmt -l cmd/evener-hub cmd/evener-hub/internal/appsource` → empty; `go build ./...` → 0; `go vet ./cmd/evener-hub/...` → 0; `go test ./cmd/evener-hub/internal/appsource/ -race -count=1 -timeout 20m` → ok (8.284s); `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → all packages ok, exit 0; `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS lint (1 modules, 82s), exit 0.

obra:
Round eleven addressed (`7f24054a8`).

- **Rejected aliased subscription leaked a remote relay** (`cmd/evener-hub/internal/appsource/remote_hub_subscription.go:464-475`): confirmed — settlement's foreign-relay refusal wrote the snapshot's canonical ref (the incumbent's identity) into the refused subscription's `remoteRef`, while `remoteRefsLocked` names only `remoteRef` plus deferred refs and `refClaimLocked` calls a ref the incumbent owns owned and skips it. The alias's own remote registration — the one the already-sent subscribe request created, keyed by the address that request carried — therefore had nothing releasing it: a remote relay with no local owner, still forwarding the thread to this connection. The refusal no longer records the canonical ref: the subscription keeps its caller-derived delivery identity, so `discardSubscriber` unsubscribes exactly that identity. The incumbent's canonical ref is not in the release set, and `refClaimLocked` still protects any identity a live sibling holds or a still-attaching one may adopt. Pinned by `TestRemoteHubAliasedForeignRelayRefusalReleasesOwnRemoteIdentity` (asserts a `thread/unsubscribe` for the rejected alias's `local:current`, and that the incumbent's `local:stable` is neither unsubscribed nor displaced and still delivers).
- **Pre-settlement routing-key mismatch** (`remote_hub_subscription.go:178`): confirmed — the initial routing key was the translated ref's suffix, while the remote resolves and keys a bare `threadId` in preference to the ref (`threadRelayTarget`, `cmd/evener-hub/app_relay.go:286-300`), so a frame emitted between the install and the `settleSubscriber` re-key matched no entry and was dropped. `threadID` now starts from the effective target thread (`remoteSubscriptionTargetThread`, the same precedence the delivery identity resolves with), `provisionalRef` is unchanged for alias recognition, and `settleSubscriber` still re-keys to the canonical identity the snapshot names. Pinned by `TestRemoteHubPreSettlementNotificationRoutesUnderEffectiveIdentity` (a parked attach plus a frame pushed during the window; the frame must arrive on the subscriber's channel after the resync). Failing-first: both new tests fail against `e67329e99` (`no thread/unsubscribe for local:current`; `the notification emitted during the attach window was never routed to the in-flight subscription`) and pass with the fix. Round nine's `TestRemoteHubDisplacedSettleUnsubscribesCanonicalRef` staged its displacement at the provisional key, so its probe subscription now installs under the effective thread the caller addressed; the test's assertions are unchanged.
- **Stale `thread/unsubscribe` doc comment** (`cmd/evener-hub/app_rpc.go:570-579`): the comment now names `relayDeliveryTarget` — the helper the call site at line 593 actually uses — and spells out that it is deliberately not `threadRelayTarget`: for a federated source the ref's suffix wins over the caller's bare threadId, because the ref is the identity that survives an identity replacement while the thread's current ID moves. Both thread/read's relay and the unsubscribe must keep resolving through that one helper.

Verification: `gofmt -l` over the touched dirs → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 20m` → ok 24.011s; the two new tests also pass 10 consecutive `-race` runs; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → all 13 packages ok (cmd/evener-hub 143.7s, appsource 4.0s, hubcore 16.7s); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS lint (1 modules, 72s), MODULE-LINT-EXIT=0.

obra:

The parent components (04b `5275b578f`, 05a `4f64b223b`) landed, so this component's own 15 commits were rebased onto `main` (`git rebase --onto origin/main dbd2bf00e`) and force-pushed with `--force-with-lease` (`7f24054a8` → `196fe0275`). The PR's diff is now exactly this component's: 12 files, +5841/−28, against `main`.

**Seven conflicts, all one class:** this branch was written against an older 05a, and `main` now carries 05a's final content in the same two files. Each was resolved by hand and verified (`go vet` on the package, no conflict markers, gofmt clean) before staging:

1. `remote_hub_source.go` — struct fields and imports: 05a's `itemPaging`/`itemPagingLocks` and this component's `subMu`/`subs`/`drains` are independent groups; kept both.
2. `remote_hub_refs.go` — function bodies at the same position: kept 05a's `fromRemoteDiagnosticRefs`/`fromRemoteNestedRef` and appended this component's JSON-level translators. The shared tail closed only the last function, so each kept function got its own closing brace.
3. `remote_hub_source.go` — struct fields again: this component's side repeated `subMu`/`subs`/`drains` and added `remoteMu`; kept 05a's already-merged block plus only the `remoteMu` addition.
4. `remote_hub_source.go` — the source doc comment and the `thread/read` subscription stripping: 05a clears the controller's intent on the wire by calling `stripRemoteSubscription`, which clears **both** `Subscribe` and `ReplaceSubscription`; this component set `ReplaceSubscription = false` explicitly and documented why controller-level replacement semantics must never reach the shared remote client. Kept the helper call with that rationale as its comment — one mechanism, not two.
5. `remote_hub_source.go` — the snapshot-read comment: same reasoning; kept the helper call with this component's comment, which covers both flags.
6. `remote_hub_refs.go` — capability masking vs the nested-ref machinery: kept both bodies (capability functions first, nested-ref translators after) with the same brace rule as (2).
7. `remote_hub_refs.go` — the pending-escalation translation: 05a's final code already translates `Evener.PendingEscalations[].Ref` and routes the diagnostics through `fromRemoteDiagnosticRefs`, so this component's earlier inline loop was dropped as a duplicate rather than applied twice.

No resolution dropped a behavior from either side: every kept function is complete, every helper has exactly one definition, and the pending-escalation translation that the round-thirteen review of 05a asked for is present exactly once.

**Gates on the rebased tree (`196fe0275`)**: `gofmt -l cmd/evener-hub/` clean; `go build ./...` exit 0; `go vet ./cmd/evener-hub/...` exit 0; `go test ./cmd/evener-hub/internal/appsource/ -race -count=1` ok (20.7s); `go test ./cmd/evener-hub/ -count=1 -timeout 30m` ok (124.6s); module-lint PASS (1 module, 50s, 0 findings).

obra:
Round twelve addressed (`076ad09e7`).

**Medium — the subscription fan-out is never reachable from the hub, because the source still reports "no relay"** — *Confirmed and fixed.* `RemoteHubSource` overrode `RelayOnThreadRead() bool { return false }` and `SupportsThreadRelay() bool { return false }` (`remote_hub_source.go:95-106`); both are defaults-true optional interfaces (`app_rpc.go:133-137`, `app_rpc.go:225-231`), so `app_rpc.go:691` could never fire for a remote source and `app_relay.go:1418` was unreachable. Both overrides are deleted, replacing staged-until-05b comments with one that names 05b as the implementer, and `stripRemoteSubscription`'s doc no longer claims SubscribeThread is unimplemented — it now explains that the relay's `SubscribeThread` is the only creator/owner of a remote subscription, so a read's own `Subscribe` must still not ride the wire, and that `ReplaceSubscription` is stripped because it is connection-scoped on the shared per-host client (forwarding it would drop every other thread's remote subscription).
Pinned by `TestHubRPCThreadReadSubscribeStartsRemoteRelay` (`cmd/evener-hub/remote_hub_source_rpc_test.go`), failing-first before the change:
```
    remote_hub_source_rpc_test.go:384: relay attach reads = 0, want exactly 1: a subscribed remote read must start the relay and reach SubscribeThread
--- FAIL: TestHubRPCThreadReadSubscribeStartsRemoteRelay (0.01s)
```
It now also asserts the read served exactly one plain snapshot read, that the relay attach is exactly one `thread/read{subscribe:true}` addressed `local:t1`, and that no read ever carries controller-level `replaceSubscription`. `TestRemoteHubSourceRelaysPlainThreadRead` (`remote_source_registration_test.go`) and `TestRemoteHubSourceReadThreadStripsSubscription` were re-pinned from "no relay happens" to "the relay does happen"; see below.

**Medium — relayed remote thread frames expose the remote daemon's capabilities, bypassing the read-path mask** — *Confirmed and fixed.* The read path masks via `fromRemoteThread` (`remote_hub_refs.go:305`); the notification path had no capability handling. Enumerating `ThreadCapabilities` from `appwire` gives exactly two carrier shapes, and both are now masked through one helper (`maskRemoteCapabilitiesRaw` → `maskRemoteThreadCapabilities`): (1) `EvenerThread.Capabilities`, i.e. the `capabilities` field of the `evener` object of any `thread`-shaped value, handled in `translateThreadRaw` — this covers the snapshot thread and `thread/started`; (2) `ThreadStatusChangedParams.Capabilities`, the top-level `capabilities` of `thread/status/changed` (the only notification whose params carry a capability set of their own — `thread/closed` carries none), handled in `translateNotification`. Absent fields stay absent (no fabricated "update"), and every other field, including unknown ones and opaque `turns[]` payloads, is still passed through byte-for-byte. `stampClosedThreadCapabilities`/`stampForkCapability`/`enrichOutputImageNotification` are keyed on `local:` (`app_relay.go:668`), so they never re-add a set to a `host:` relay key; `applyHubForkCapability` (`app_threadread.go:545-554`) leaves non-`local:` refs untouched, so reads and notifications agree on fork too.
Pinned by `TestRemoteHubStatusNotificationMasksUnforwardedCapabilities`, `TestRemoteHubNestedThreadNotificationMasksUnforwardedCapabilities` and `TestRemoteHubNotificationMaskPreservesUnrelatedFields` (`cmd/evener-hub/internal/appsource/remote_hub_subscription_roundtwelve_test.go`), failing-first before the change:
```
    remote_hub_subscription_roundtwelve_test.go:54: status capabilities = {Send:true Steer:true Interrupt:true Compact:true Clear:true ForkFromTurn:true Shutdown:true ChangeModel:true ChangeVisionModel:true Queue:true Goal:true SharedNotes:true Rename:true SkillInput:true}, want every unforwarded action masked like the read path
--- FAIL: TestRemoteHubStatusNotificationMasksUnforwardedCapabilities (0.00s)
    remote_hub_subscription_roundtwelve_test.go:105: nested capabilities = {Send:true Steer:true Interrupt:true Compact:true Clear:true ForkFromTurn:true Shutdown:true ChangeModel:true ChangeVisionModel:true Queue:true Goal:true SharedNotes:true Rename:true SkillInput:true}, want every unforwarded action masked like the read path
--- FAIL: TestRemoteHubNestedThreadNotificationMasksUnforwardedCapabilities (0.00s)
    remote_hub_subscription_roundtwelve_test.go:145: status capabilities = {Send:true Steer:true Interrupt:true Compact:true Clear:true ForkFromTurn:true Shutdown:true ChangeModel:true ChangeVisionModel:true Queue:true Goal:true SharedNotes:true Rename:true SkillInput:true}, want every unforwarded action masked
--- FAIL: TestRemoteHubNotificationMaskPreservesUnrelatedFields (0.00s)
```
The end-to-end path is pinned by `TestHubRPCThreadReadSubscribeDeliversMaskedRemoteStatus`, which pushes a `thread/status/changed` from the scripted remote and reads it off the subscribed browser client's notification stream; it was red before the overrides were deleted (`timed out waiting for the relayed status frame: the subscription delivered no live update`) and now delivers, with `ref` translated to `h1:t1` and a fully masked capability set.

**Re-pinned pre-existing tests (assertions strengthened, not weakened):**
- `TestRemoteHubSourceReadThreadStripsSubscription` (appsource) still asserts the wire read carries neither `subscribe` nor `replaceSubscription` — the intent it always protected — and now asserts the source reports a relay (`RelayOnThreadRead`/`SupportsThreadRelay` through the hub's optional interfaces, so the default-true shape passes) instead of asserting the staged false. 05a's comment that "the controller-side relay gate is RelayOnThreadRead, which reports false for this source" is replaced by why the read's intent must not become a second, unowned remote subscription.
- `TestHubRPCThreadReadSubscribeServesRemoteSnapshot` (hub) still asserts the subscribed remote read returns the translated snapshot and a controller-owned continuation cursor, and now asserts exactly one *plain* remote read carried no subscription intent plus that no read anywhere carried controller-level `replaceSubscription`. The old "no thread/read on the wire carries subscribe" clause is now covered by its successor `TestHubRPCThreadReadSubscribeStartsRemoteRelay`, which asserts the relay's own attach does reach `SubscribeThread` — the opposite of the old staged expectation and the point of 05b.
- `TestRemoteHubSourceDoesNotRelayPlainThreadRead` → `TestRemoteHubSourceRelaysPlainThreadRead` (hub): flipped from `relayOnThreadRead(source)` must be false to must be true, and extended with `sourceSupportsThreadRelay(source)`.
- Stale "staged until 05b" comments in `TestHubRPCThreadReadServesRemoteHubSource` and the two re-pinned tests were refreshed; `rg 'staged until 05b|SubscribeThread is not implemented|notImplemented SubscribeThread|deleted when 05b' cmd/` is now empty. The gate comment at `app_rpc.go:692` is a general statement about sources that report no fan-out and stays accurate unchanged.

Verification:
- `go build ./...` → exit 0
- `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 20m` → `ok primeradiant.com/evener/cmd/evener-hub/internal/appsource 60.439s` (exit 0)
- `go test ./cmd/evener-hub/ -count=1 -timeout 30m` → `ok primeradiant.com/evener/cmd/evener-hub 126.570s` (exit 0)
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → `PASS lint (1 modules, 50s)` (exit 0, 0 findings)

obra:

**CI** — all 15 checks green at `076ad09e7`.

**Review** — RoboRev's verdict for `076ad09e7`: **"No issues found."** This closes the review sequence on the rebased branch; its two Mediums were the completion of work component 05a deliberately staged: the source reported "no relay" (leaving the whole subscription fan-out unreachable in production) and the notification path passed the remote daemon's capabilities through unmasked (letting a subscribed client re-enable actions that then fail). Both are now pinned by tests, including the two staging-era tests that were re-pinned to the new intent rather than deleted.

**Up-to-date override** — the branch is 5 commits behind `main`. Overlap check, run immediately before this comment:

```
$ gh pr view 1291 --json mergeable,mergeStateStatus --jq '.mergeable, .mergeStateStatus'
MERGEABLE
BEHIND

$ comm -12 <(gh pr view 1291 --json files --jq '.files[].path' | sort) \
           <(git diff --name-only "$(git merge-base origin/main origin/multi-host-pr05b-remote-hub-subscription)" origin/main | sort)

$ git diff --name-only "$(git merge-base origin/main origin/multi-host-pr05b-remote-hub-subscription)" origin/main | wc -l
32
```

`main` changed 32 files since the merge base; none of them is a file this PR touches.

PR #1293, opened 2026-09-14
feat(hub): remote hub mutations and mutation-unknown mapping (multi-host 05c)

obra:
Part 05c of component 05 (remote hub source), stacked on #1291 (05b).

## What this adds
All remaining `Source` methods on `RemoteHubSource`, so the interface is fully implemented with nothing left staged.

- **Mutation-envelope methods** (params carry a `ClientMutationID`) go through a new `mutationCall` and `remoteHubMutationCallError`: `StartTurn`, `SteerTurn`, `InterruptTurn`, `QueueTurn`, `DrainAsSteer`, `PromoteQueuedAsSteer`, `CancelQueued`, `NotesHumanSet`, `UrlsRemove`, `ClearThread`. A lost response — the transport mapped to `SessionUnavailable` — becomes `ErrorMutationOutcomeUnknown` with the client's `ClientMutationID` preserved, `MutationOutcomeUnknown`, and `RetryDispositionAutomatic`, mirroring `localDaemonMutationCallError` exactly. A remote mutation may well have been applied by the time the SSH channel died, so this is the honest shape and it lets the hub's retry gate do its job.
- **Lifecycle methods without a mutation id** go through `call`: `StartThread`, `ResumeThread`, `ForkThread`, `CompactThread`, `ShutdownThread`, `SetThreadModel`, `SetThreadReasoningEffort`, `SetThreadVisionModel`, `SetThreadName`, `GoalSet`, `ResolveSandboxEscalation`.
- **Read-only forwards**: `ListTasks`, `ListJobs`, `JobOutput`.
- Every ref-bearing method translates the inbound controller ref to the remote `local:` namespace through the same `toRemoteRef` path as the read methods; a foreign ref is refused before any wire call.
- `translateOut` grows cases for the four responses that embed a thread: `ThreadStartResponse`, `ThreadResumeResponse`, `ThreadForkResponse` (each `.Thread`), and `ThreadClearResponse` (`.Thread` plus its `.Ref` string).

The now-dead `notImplemented` helper and its `fmt` import were removed: every `Source` method is implemented, so it had no callers.

## Tests
`remote_hub_mutations_test.go`, over the same in-memory scripted-remote harness (no SSH, no network, no host): a table covering every method above asserting the exact wire method, `host:S` → `local:S` ref translation, and unchanged `clientMutationId`; outbound thread/ref translation for the four thread-bearing responses; mutation-unknown mapping on a dropped pipe with a read-only call proven to stay `SessionUnavailable`; semantic wire errors preserved through both `call` and `mutationCall`; and foreign-ref refusal before any wire call.

## Gates (run on this branch)
- `gofmt -l cmd/evener-hub/internal/appsource` — clean
- `go build ./...` — ok
- `go vet ./cmd/evener-hub/...` — ok
- `go test -race ./cmd/evener-hub/internal/appsource/ -count=1` — ok 4.3s
- `golangci-lint run ./cmd/evener-hub/internal/appsource/...` — 0 issues
- `go test ./cmd/evener-hub/ -count=1` — ok 157.5s

## Note
The spec's error-handling section says a foreign ref should map to `InvalidParams`; the reference implementation it cites (`localEntryForRefMode`) actually returns a plain `source not found: <id>` error, and 05a already matches the code. That is unchanged here.


obra:
Round 1 addressed (`164610a1`). 3 findings — 1 high, 2 medium — all verified real and fixed.

- **High — `StartThread` forwarded the controller's `harness` source selector verbatim.** `ThreadStartParams.Harness` is the hub's source selector (`launchSourceID`, `app_threadlifecycle.go:320`), not just an agent label: the controller sends `harness: "<host>"` to route to this source, and forwarding it made the remote hub resolve a source named after the controller's host and fail with `Unavailable("spawn source is not available: <host>")`. `StartThread` now rewrites the forwarded harness to the remote hub's own `"evener"` (which resolves to the remote's `local` source), mirroring `hubModelListInner`'s selector handling. Pinned by `TestRemoteHubMutationWireMethodsAndRefs/StartThread`, which now sends `harness: "host"` and asserts the wire carries `harness: "evener"`; before the fix it failed with `StartThread harness = "host", want "evener"`.

- **Medium — `JobsList` returned the recursive activity tree with untranslated refs.** `translateOut` had no `JobsListResponse` case, so `root.ref`, nested session `ref`, `ownerRef`, `childRef`, and a delegate turn's session `transcriptRef` all stayed in the remote's `local:` namespace, misrouting activity/transcript actions to the controller's local source. Added `translateActivityRefs`, a recursive walk over the decoded `Data` (nested `map[string]any`/`[]any`, since `JobsListResponse.Data` is `any`), rewriting `ref`/`ownerRef`/`childRef`/`transcriptRef` from `local:*` to `<host>:*` while preserving bare session ids and opaque `job:<id>` shell transcript refs byte-for-byte. Pinned by `TestRemoteHubJobsListTranslatesActivityRefs`; before the fix all six embedded refs remained `local:*`.

- **Medium — `mutationCall` mapped client-acquisition failures to `ErrorMutationOutcomeUnknown`.** `s.client(ctx, s.id)` dials/attaches the host; when that failed no request had been sent, yet the error went through `remoteHubMutationCallError` and surfaced as `InternalError` / `MutationOutcomeUnknown`. That (a) prevented the auto-resume/reconnect gates (`isSessionUnavailableError`) from firing and (b) falsely told the client the mutation might already have executed. `mutationCall` now returns `s.mapCallError(err)` for client-acquisition failures and reserves `remoteHubMutationCallError` for `client.Request` failures, matching `LocalDaemonSource.withClientCallMapper` (which maps only `fn`). Pinned by `TestRemoteHubMutationClientAcquisitionFailureIsSessionUnavailable` (before the fix: `code = -32603, want -32014`); the existing `TestRemoteHubMutationOutcomeUnknownOnResponseLoss/mutation` still pins response-loss → outcome-unknown.

Verification: `gofmt -l` (clean) · `go build ./...` (ok) · `go vet ./cmd/evener-hub/...` (ok) · `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` (ok) · `go test ./cmd/evener-hub/... -count=1` (13 packages ok) · `module-lint` (PASS, 1 module).

obra:
Round two addressed (`f73205043`). 2 Medium.

- **Medium — caller deadline misclassified as `MutationOutcomeUnknown`** (`remote_hub_mutations.go:28-29`). Confirmed real. `client.Request` returns `ctx.Err()` when the caller's context ends, and `mapCallError` maps `context.DeadlineExceeded` to `SessionUnavailable` (correct for a transport timeout), which `remoteHubMutationCallError` then turned into `ErrorMutationOutcomeUnknown` with `RetryDispositionAutomatic` — an automatic retry of a mutation the caller abandoned. `mutationCall` now returns the caller's own `ctx.Err()` raw at entry, after client acquisition, and after a failed request, mirroring `LocalDaemonSource.withClientCallMapper` line-for-line. Pinned by `TestRemoteHubMutationCallerContextStaysRaw` (request-deadline and client-acquisition-deadline subtests); both subtests fail on the pre-fix head with exactly `mutation outcome is unknown after remote hub response loss`.
  - The finding's "apply the same handling in `call`" half is **not duplicated here**: `call`/`mapCallError` are component 05a's read path (added in `1606c6a58`), and 05a's current head `3c1c25a9c` already preserves both `context.Canceled` and `context.DeadlineExceeded` raw in `mapCallError` and classifies attach failures through `mapConnectError`. This branch's base predates that commit, so the read path is fixed upstream of this PR; editing the same function here would collide with 05a's landed fix. The mutation path — this PR's own new code — is fixed in this PR.
- **Medium — nested diagnostic refs not translated** (`remote_hub_refs.go:83-97`). Confirmed real. `fromRemoteThread` rewrote only `Thread.Source`, `Evener.Ref`, and `Evener.ParentRef`; the session-valued refs nested in `Evener.Diagnostics` stayed in the remote hub's `local:` namespace, so a controller client would route a delegate's child-thread read to the local source or fail. Added `fromRemoteDiagnostics`, which rewrites `Diagnostics.Delegates[].TranscriptRef` and `Diagnostics.Jobs[].TranscriptRef` via a new `fromRemoteRefOrOpaque`: a `local:<thread>` ref moves to `<host>:`, while an opaque `job:<id>` ref (a shell job's transcript) and every bare id field (`sessionId`, `ownerSessionId`, `childSessionId`, `rootSessionId`) are left byte-for-byte. Pinned by `TestRemoteHubThreadDiagnosticsTranslateNestedRefs`, which fails on the pre-fix head with `delegate transcriptRef = "local:child", want "host:child"`.

Verification: `gofmt -l cmd/evener-hub/internal/appsource/` → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → clean; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 30m` → ok; `go test ./cmd/evener-hub/... -count=1` → all 13 packages ok; `module-lint` → PASS (0 findings). Fuzz targets untouched, so no fuzz gates were run.

obra:
Round three addressed (`ba3db0743`). 1 Medium.

- **Medium — caller deadline on the `call` path misreported as host unavailability** (`remote_hub_source.go:60`). Confirmed real on this branch. `call` routed every error through `mapCallError`/`transportUnavailable`, which maps `context.DeadlineExceeded` to `SessionUnavailable`, so a caller deadline on any new non-`mutationCall` forward (`StartThread`, `ResumeThread`, `ForkThread`, `CompactThread`, `ShutdownThread`, the `SetThread*` forwards, `GoalSet`, `ListTasks`/`ListJobs`/`JobOutput`) became "remote hub unavailable" and could fire the auto-resume gate. `call` now applies the same raw-context guard as `mutationCall`: it returns `ctx.Err()` raw on entry, and checks `ctx.Err()` before mapping the client-acquisition and `Request` errors. Pinned by `TestRemoteHubCallCallerContextStaysRaw` (request-deadline and client-acquisition-deadline subtests, exercising `StartThread`); both fail on the pre-fix head with exactly `appwire.WireError remote hub unavailable: host: context deadline exceeded` and `*errors.errorString attach failed`. Round two's two Medium resolutions (raw `mutationCall` context, nested diagnostic ref translation) are untouched.
  - Note on round two's counter-argument: 05a's head `f3b24dd55` does preserve `DeadlineExceeded` raw in its own `mapCallError`, but 05c's tree does not contain that commit — 05c is based on `69e8b8f1b`, which predates it. So on this branch the new `s.call` forwards were still affected, which is why the fix is applied here rather than deferred to 05a.

Verification: `gofmt -l cmd/evener-hub/internal/appsource/` → empty; `go build ./...` → exit 0; `go vet ./cmd/evener-hub/...` → exit 0; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 30m` → ok (8.0s); `go test ./cmd/evener-hub/... -count=1` → all 13 packages ok; `module-lint` → PASS (0 findings). Fuzz targets untouched, so no fuzz gates were run.

obra:
Round four addressed (`9d241357e`).

- **Medium — Default remote thread listing loses nested-source threads (`remapRemoteSourceIDs`).** Confirmed: with an empty controller `SourceIDs` filter the source forwarded no filter at all, so a remote hub could return threads from its own nested remote sources; `fromRemoteThread` refuses a non-`local` ref and the whole response failed, losing every thread from that host. Fixed by forwarding `SourceIDs: ["local"]` for an empty filter — the only source this hub can represent (`sourceAllowedForList` already gates whether the source is called when the controller names specific sources). Pinned by `TestRemoteHubListThreadsDefaultRestrictsToLocalSource` (scripted remote returns the nested thread when unfiltered and only the local thread when filtered, and the test asserts the forwarded `sourceIds`). `TestRemapRemoteSourceIDs`'s `empty` case updated from `nil` to `["local"]`; the other cases are unchanged.
- **Low — Untyped activity-tree walk rewrites ref-named keys inside opaque payloads (`translateActivityRefs` / `activityRefKeys`).** Confirmed: the old walk matched `ref`/`ownerRef`/`childRef`/`transcriptRef` at arbitrary depth, so a delegate `message`/`structuredResult` (opaque `json.RawMessage`) containing `{"ref":"local:main"}` was rewritten to `{"ref":"host:main"}`. Rewrote the walk to be field-aware: it now descends only `JobActivityTree`'s structural containers (`root`, `entries`, `job`, `delegate`, `delegate.child`, `delegate.turns`) and rewrites only the ref fields those nodes declare; `activityRefKeys` is removed. Opaque payloads are never entered. Pinned by `TestRemoteHubJobsListLeavesOpaquePayloadsUntouched` (opaque `message`/`structuredResult` including a nested `ref`/`ownerRef`/`childRef` stay `local:…` while the structural session/delegate/job refs still translate). This composes with the sibling 05b nested-ref fix: 05b adds translation for *known session-reference fields*, while this change restricts translation to *known structural locations*, so the two move in the same field-aware direction rather than fighting over opaque payloads.
- All earlier rounds' assertions are intact (`TestRemoteHubJobsListTranslatesActivityRefs`, the diagnostics/transcript-ref tests, and the `remapRemoteSourceIDs` non-empty cases are unchanged).

Verification:
- `gofmt -l cmd/evener-hub/internal/appsource/` → empty
- `go build ./...` → ok
- `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (13 packages, exit 0)
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS lint (1 modules)

obra:
Round five addressed (`ff5242f4`).

- **High — `sessionUnavailable` wire errors may not convert to `mutationOutcomeUnknown`** (`remote_hub_mutations.go:56`): **Contested — no change.** The `wire.Data.(appwire.ErrorData)` type assertion is the discriminator, not an oversight, and the two shapes mean opposite things. The typed shape is the one `mapCallError`/`transportUnavailable` synthesizes locally after a transport failure — the request crossed the wire and its response was lost — which is exactly the in-doubt case this function exists to relabel `MutationOutcomeUnknown` + `RetryDispositionAutomatic`. A decoded `map[string]any` is an error frame that arrived **intact**, so by construction the response was not lost; it is the remote hub's semantic verdict ("thread not found", "agent unavailable before dispatch"). The hub already labels a mutation genuinely in doubt on its own as `mutationOutcomeUnknown`, so a `sessionUnavailable` it sends means *not applied*; locally that same verdict is `NotAccepted` (`localDaemonMutationEntryError`) and is what the auto-resume gate consumes via `isSessionUnavailableError` (which does accept the decoded shape). Converting it here would re-drive a mutation against a session known to be absent — an automatic-retry loop on "thread not found". This mirrors `localDaemonMutationCallError`, whose typed-only check is exercised by `TestLocalDaemonMutationCallErrorSessionUnavailable` with a locally constructed `appwire.SessionUnavailable`. Pinned by the new `TestRemoteHubMutationPreservesRemoteSessionUnavailable` (a wire-encoded `SessionUnavailable` from the responder stays `CodeUnavailable`/`sessionUnavailable`) alongside the existing `TestRemoteHubMutationOutcomeUnknownOnResponseLoss` (a dropped pipe still becomes `mutationOutcomeUnknown`), and the reasoning is now recorded in the function comment.

- **Medium — legacy flat-array job ref format not translated** (`remote_hub_refs.go`): **Fixed.** `translateActivityRefs` now switches on the decoded shape and walks the retired flat `[]EvenerJobInfo` payload (`docs/appwire-protocol.md`, `evener/jobs/list`) with the same declared-field policy as the activity tree: `EvenerJobInfo`'s only ref-valued field is `transcriptRef`, so a session-valued ref moves `local:` → `host:` while an opaque `job:<id>` ref and every bare-id key are preserved byte-for-byte. Pinned by `TestRemoteHubJobsListTranslatesLegacyFlatArrayRefs` (fails before the change: the legacy turn's `transcriptRef` stayed `local:child`).

- **Low — activity walker translates a `delegate.transcriptRef` key no wire type declares** (`remote_hub_refs.go:268`): **Fixed.** Dropped the delegate-level `transcriptRef` translation. `JobActivityDelegate`'s only declared ref field is `childRef` (`appwire/types.go`), and the producer sets `ChildRef` from the descriptor with no delegate-level `transcriptRef` (`agent/jobs_activity.go`), so that key is undeclared data and now survives byte-for-byte like any other non-structural key; the delegate's turn nodes still translate their own declared `transcriptRef`. `TestRemoteHubJobsListTranslatesActivityRefs` and `TestRemoteHubJobsListLeavesOpaquePayloadsUntouched` now assert the undeclared key is preserved (both failed against the pre-fix walker).

Verification: `gofmt -l cmd/evener-hub/internal/appsource/` → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` → ok 12.2s; `go test ./cmd/evener-hub/... -count=1` → ok (13 packages, 158s); `GOLANGCI_LINT_CACHE=... MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (1 modules, 123s), 0 findings.

obra:
Round six addressed (`f84c06d`).

- **`remote_hub_source.go:147-149` — a non-empty `SourceIDs` filter excluding this source.** Confirmed: `remapRemoteSourceIDs` mapped such a filter to an empty slice, and `ThreadListParams.SourceIDs` is `omitempty`, so the filter was omitted on the wire and asked the remote hub for all of its sources. `ListThreads` now returns an empty response without forwarding when `SourceIDs` is non-empty and does not contain `s.id`. (The hub's `sourceAllowedForList` fan-out already skips this source for such a filter, so it was not reachable through `thread/list`; the source method now honors its own filter contract instead of depending on its caller.) Pinned by `TestRemoteHubListThreadsExcludingFilterReturnsEmpty`.
- **`remote_hub_refs.go:150-155` — one unrepresentable row aborted the whole list.** Confirmed. `translateOut` now translates `ThreadListResponse.Data` row-by-row, skipping only a row whose ref is a nested non-local ref (unaddressable in the controller namespace) while retaining every representable row, rather than returning the refusal and losing valid local threads. Single-thread responses (`thread/read`, the thread-returning mutations) still fail hard, since the thread *is* the response. Pinned by `TestRemoteHubListThreadsKeepsValidRowsWhenRowUnrepresentable`.
- **`remote_hub_mutations.go:30-36` — acquisition error not classified as `sessionUnavailable`.** Confirmed: the production connector (`sshManager.Ensure`) surfaces `sshconn.ErrSSHStart` wrapping `*exec.ExitError`/plain stderr, which carries none of the transport-shaped strings `mapCallError` matched, so a mutation against an unreachable host did not engage the auto-resume gate. Fixed in the shared `transportUnavailable` helper, so both `call` and `mutationCall` map `errors.Is(err, sshconn.ErrSSHStart)` → `SessionUnavailable`; the terminal classes (`ErrSSHAuth`, `ErrHostNotFound`, `ErrProtocolIncompatible`) stay raw. Pinned by `TestRemoteHubAcquisitionErrSSHStartIsSessionUnavailable` (read path, mutation path, and three terminal classes), alongside the existing `io.EOF` case.

Prior-round dispositions unchanged: caller-deadline classification on `call`/`mutationCall`, the `["local"]` default filter, structural-location ref translation (opaque `message`/`structuredResult` untouched), and the legacy flat-array translation.

Verification: `gofmt -l` (touched dir) empty; `go build ./...` OK; `go vet ./cmd/evener-hub/...` OK; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` ok; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` ok; `module-lint` PASS (1 module, 148s). All three new tests were red before the fix.

obra:

The parent components (04b `5275b578f`, 05a `4f64b223b`, 05b `fbf3e8cb0`) landed, so this component's own 7 commits were rebased onto `main` (`git rebase --onto origin/main 69e8b8f1b`) and force-pushed with `--force-with-lease` (`f84c06daa` → `a3bfa8053`). The PR's diff is now exactly this component's.

**Conflicts, resolved by hand.** This branch was written against an older 05a, so `main` carries newer content in the same files:

1. `remote_hub_source.go` (first commit) — the source doc comment (kept the superset: components 05a–05c, 05d staged); a large block where this component's side adds nothing (kept `main`'s item-paging machinery); and the deletion of the staging `notImplemented` stubs. The rebase restored those stubs because 05a's final file regrew them with a different body, so I removed exactly the 24 whose methods now live in `remote_hub_mutations.go`, and restored the `fmt` import the merge dropped.
2. `remote_hub_refs.go` (third commit) — three conflicts: the `fromRemoteThread` doc comment (kept the superset, including the pending-escalation sentence), the diagnostics call (kept 05a's pending-escalation translation and routed diagnostics through this component's `fromRemoteDiagnostics`), and the helper bodies (kept 05a's `fromRemoteNestedRef` for the pending-escalation loop and this component's `fromRemoteDiagnostics`/`fromRemoteRefOrOpaque`; dropped 05a's `fromRemoteDiagnosticRefs`, which `fromRemoteDiagnostics` supersedes — one helper per call site).
3. `remote_hub_source.go` (fourth commit) — the call path: kept this component's caller-context passthrough (`ctx.Err()` stays raw) **and** 05a's `mapConnectError` for the connect step.
4. `remote_hub_refs_test.go` (fifth commit) — import union (`context` + `encoding/json`).
5. `remote_hub_source.go` (seventh commit) — four conflicts, each one this component's side duplicating an already-landed block (two imports, the `ErrSSHStart` mapping, the excluding-filter comment); kept `main`'s.

**Two integration commits on top** (`90aa8dbdd`, `a3bfa8053`), because landing this component on current `main` changes two things the older branch could not see:

- `TestRemoteHubCapabilitiesMatchForwardedMethods` (05a's) asserted that every enumerated mutation method still fails closed with the staged not-implemented error. This component implements them, so that assertion is false. The test now drives each method through a connector that fails (no live hub needed) and asserts the implemented method's own error. The capability flags stay **masked**, which is the second half of that test's contract: the controller-side forwarded set is component 05d's host capability probe (the component spec assigns the capability answer to 05d — `remote_hub_probe.go`, `HostCapabilities`/`CapabilitySource`, preflight wiring), and forwarding the remote daemon's own claim verbatim would advertise actions this controller cannot verify the host supports. 05a's and 05b's masking tests are therefore untouched and still valid.
- The staging `notImplemented` helper lost its last callers, and module lint flagged it unused; it is deleted (this component's own commit deleted it too — the rebase had restored it).

**Gates on the rebased tree (`a3bfa8053`)**: `gofmt -l cmd/evener-hub/` clean; `go build ./...` exit 0; `go vet ./cmd/evener-hub/...` exit 0; `go test ./cmd/evener-hub/internal/appsource/ -race -count=1` ok (21.1s); `go test ./cmd/evener-hub/ -count=1 -timeout 30m` ok (124.1s); module-lint PASS (1 module, 0 findings).

obra:
Round thirteen addressed (`c5a72d667`).

- **medium — `mutationCall`'s acquisition step mapped through `mapCallError` instead of `mapConnectError`** (`remote_hub_mutations.go:34`): confirmed and fixed. `mapCallError` short-circuits `errors.Is(err, context.DeadlineExceeded)` raw before `transportUnavailable` can match `sshconn.ErrSSHStart`, and the attach handshake's `initTimeout` child context produces exactly an `ErrSSHStart` chain that also satisfies `errors.Is(err, context.DeadlineExceeded)` (`sshconn/manager.go`: `client.Initialize` returns the child `ctx.Err()`, wrapped as `%w: … initialize: %w`). A timed-out attach on a mutation therefore never became the `CodeUnavailable` + `ErrorSessionUnavailable` pair `isSessionUnavailableError` (`cmd/evener-hub/app_compact.go:20`) reads, so the auto-resume gate never fired, while reads against the same host mapped correctly. The acquisition step now uses `s.mapConnectError(err)`, matching `call`; the only classification change is deadline-shaped acquisition failures (bare or inside the `ErrSSHStart`/`ErrRestart` chain), which now become `SessionUnavailable`. Pinned by `TestRemoteHubAttachHandshakeTimeoutIsSessionUnavailable` (read + mutation paths, asserting the wire code and the `ErrorSessionUnavailable` marker the gate reads). Failing-first, before the one-line change: `--- FAIL: TestRemoteHubAttachHandshakeTimeoutIsSessionUnavailable/mutation_path` — `remote_hub_mutations_test.go:245: error = *fmt.wrapErrors sshconn: ssh channel start failed: host "host" initialize: context deadline exceeded: , want appwire.WireError` (the `read_path` subtest passed).
- **What still stays raw:** a genuine caller cancellation or deadline — the `ctx.Err()` guard precedes the mapping on both the acquisition and request steps of `mutationCall` and `call`, `mapConnectError` still returns `context.Canceled` raw, and no caller-expired context is remapped. Pinned by `TestRemoteHubMutationCallerContextStaysRaw` (`request deadline` and `client acquisition deadline`) and `TestRemoteHubCallCallerContextStaysRaw`; the terminal acquisition classes (`ErrSSHAuth`, `ErrHostNotFound`, `ErrProtocolIncompatible`) stay raw as pinned by `TestRemoteHubAcquisitionErrSSHStartIsSessionUnavailable`, which still passes. The mutation-unknown mapping, the read path's raw caller-context passthrough, the default `thread/list` `local`-namespace restriction, and the nested diagnostic/jobs ref translation are untouched.

Verification:
- `go build ./...` → exit 0
- `go test ./cmd/evener-hub/internal/appsource/ -race -count=1 -timeout 20m` → `ok primeradiant.com/evener/cmd/evener-hub/internal/appsource 21.183s`
- `go test ./cmd/evener-hub/ -count=1 -timeout 30m` → `ok primeradiant.com/evener/cmd/evener-hub 124.657s`
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → `PASS lint (1 modules, 50s)`, 0 findings

obra:

**CI** — run 35082343131 at `c5a72d667`: completed/success, all jobs green.

**Review** — RoboRev's verdict for `c5a72d667`: **"No issues found."** The re-review of the rebased head (`a3bfa8053`) had one Medium: `mutationCall` mapped a client-acquisition failure through the request mapper, which returns a `context.DeadlineExceeded` chain raw before the transport classification can match `sshconn.ErrSSHStart` — so a timed-out attach on a mutation never became the `SessionUnavailable` the auto-resume gate reads, while reads against the same host classified correctly. Fixed by mapping the acquisition step through the connect mapper, pinned by a failing-first test that reproduces the exact `ErrSSHStart`-wrapping-`DeadlineExceeded` shape.

**Integration commits on this branch** (both stated in the rebase note above, repeated here because they are the coordinator's changes rather than the component's): `90aa8dbdd` re-pins 05a's `TestRemoteHubCapabilitiesMatchForwardedMethods` to the post-05c state (methods implemented; capability flags still masked, because the component spec assigns the capability answer to 05d's host probe), and `a3bfa8053` deletes the staging `notImplemented` helper that lost its last caller and that module lint flagged unused.

**Up-to-date override** — the branch is 7 commits behind `main`. Overlap check, run immediately before this comment:

```
$ gh pr view 1293 --json mergeable,mergeStateStatus --jq '.mergeable, .mergeStateStatus'
MERGEABLE
BEHIND

$ comm -12 <(gh pr view 1293 --json files --jq '.files[].path' | sort) \
           <(git diff --name-only "$(git merge-base origin/main origin/multi-host-pr05c-remote-hub-mutations)" origin/main | sort)

$ git diff --name-only "$(git merge-base origin/main origin/multi-host-pr05c-remote-hub-mutations)" origin/main | wc -l
38
```

`main` changed 38 files since the merge base; none of them is a file this PR touches.

Issue #1292, opened 2026-09-14
main CI red after #1168: TestDrainStallGiveUpRechecksTheWakeEdge and TestSkillComposerBrowser fail on c09997369


obra:
The first CI run on main after #1168 (skills lifecycle and explicit selection) merged as c09997369 failed two jobs (run 34809010302):

- `tests`: `TestDrainStallGiveUpRechecksTheWakeEdge` — `session_jobtree_drain_stall_test.go:341: give-up must return nil error, got context deadline exceeded` (30s).
- `web`: `TestSkillComposerBrowser` — `skill_composer_browser_test.go:333: skill browser driver failed: node scripts/skillguard/run.mjs exited exit status 1`.

The previous main commit 1d5e12100 was green. The failed jobs are being rerun to separate a deterministic break from load flakes. Until main is green again, queue PRs merge main only up to 1d5e12100 so their CI stays attributable.

obra:
Narrowed:
- The `tests` job passed on rerun; `TestDrainStallGiveUpRechecksTheWakeEdge` was a load flake.
- The `web` job failed again on `TestSkillComposerBrowser`, and it reproduces locally at c09997369 in ~9s.
- **The tree at c09997369 is byte-identical to #1168's head 2b9831bfe** (`git diff` empty), and that head's own CI ran `web-skillguard` and passed at 04:49Z. So this is not a merge interaction; the guard is timing-dependent.
- Driver failure: after "select all + Backspace + remove chip" the composer still holds `PROSE_DRAFT_14b staged for the switc` — the typed draft minus its last character. The select-all was collapsed before Backspace landed, so Backspace removed one character.
- Measuring now whether the composer autofocus (#1256, `useMountAutofocus`) collapses the selection, by reverting it in a scratch worktree and rerunning the guard.

obra:
Main is green again: run 34812313920 on 33b553278 (#1295 merged) passed every job. Cause was the skill guard driver's Ctrl+A key event; the fix alone also passed the guard on the PR branch at e4f43bc12. The drain-stall failure in the first run was a load flake.

PR #1295, opened 2026-09-14
fix(web): select the skill guard's draft through the DOM, not a Ctrl+A key event

obra:
Main has been red since #1168 merged as c09997369: `TestSkillComposerBrowser` (the `web-skillguard` browser guard) fails at its draft-clear step, twice on main and reproducibly locally, while the byte-identical tree passed once on #1168's own CI. Tracked in #1292.

**Cause.** The guard driver sends Ctrl+A as a bare `Input.dispatchKeyEvent`. That is never select-all on macOS Chrome, and through the protocol it is not guaranteed to run the editing command on the Linux runner either. The Backspace that follows therefore deletes one character, and the check finds `PROSE_DRAFT_14b staged for the switc`.

**Fix.** A `selectAll(ref)` driver method selects the draft with the textarea's own `select()` and verifies the selection covers the value; the four select-all sites use it, and the `KeyA` key mapping goes away with the key event.

**Ruled out.** Reverting the composer autofocus (#1256) in a scratch tree at c09997369 and rerunning the guard fails identically, so the autofocus is not involved.

**Evidence.** At c09997369 the guard fails in ~9s at the draft-clear step; with this change it gets past that step. Further local runs on this host are not conclusive as a pass/fail signal because the machine is running several parallel race suites (load average above 30) and the guard's 15s waits time out at varying later steps; CI on this PR is the clean validation.

The `tests` job failure in the same main run (`TestDrainStallGiveUpRechecksTheWakeEdge`) passed on rerun and is a load flake, not addressed here.


obra:
The reviewers are right: 2bb35ddd0 carried an accidental revert of `useMountAutofocus` (#1256). That revert was a measurement step in my scratch tree, staged, and my pre-commit check used `git diff` (unstaged only), so it showed one file when the index held nine. The branch is rewritten to the single-file driver change at e4f43bc12; `git diff origin/main` is now exactly `scripts/skillguard/run.mjs`. The autofocus contract is untouched. Sorry for the noise.

PR #1296, opened 2026-09-14
feat(hub): remote hub capability probe (component 05d)

obra:
Component 05d of the multi-host stack: the remote-hub capability probe.

## What
`RemoteHubSource.HostCapabilities(ctx)` assembles a snapshot of a remote host: the global launch layer, models, plugins, auth state, instances, its working roots, plus protocol version, hub version, OS/arch and the peer's advertised feature set. Consumed by component 06's fleet view.

## Design notes
- **No second `initialize`.** A remote hub rejects re-initialization on the already-initialized SSH channel, and `sshconn` discards the `InitializeResponse` it already performed. So protocol version, hub version, OS/arch and features arrive through a new `HostFactsFunc` seam fed by the component-04 SSH preflight.
- **New `appwire.Client.Features()` getter** exposes the feature set `Initialize` cached, for callers that never saw the response.
- **Five hub-scoped probe reads**, all on the exact client the result is cached against: `evener/launch/getLayer` (global), `model/list`, `evener/plugin/list`, `evener/auth/list`, `evener/instance/list`.
- **Cache keyed by client pointer**, so a component-04 reconnect (new client) re-probes automatically, matching `RemoteHubSource`'s existing no-client-caching behavior.
- Errors map through the existing `mapCallError`, so an offline host surfaces `SessionUnavailable` naming the host.

## Wiring
`WebConfig.RemoteHostFacts` is wired in `main.go` from the sshconn manager's `Ensure`/`Preflight`/`Client().Features()` and installed per source in `newHubSourceRegistry`.

## Tests
`remote_hub_probe_test.go` covers the happy path and exact wire params, the facts seam, the nil-facts case, caching (no extra wire traffic), re-probe on client change, wire-error passthrough, and a closed transport mapping to `SessionUnavailable`. The scripted-client harness was factored minimally and additively.

Gates: gofmt clean; `go build ./...`; `go vet ./cmd/evener-hub/... ./appwire/...`; `go test -race ./cmd/evener-hub/internal/appsource/`; `go test -race ./appwire/`; `golangci-lint run` on both packages (0 issues); full `go test ./cmd/evener-hub/` (168s).

Stacked on #1293 (05c).

PR #1299, opened 2026-09-14
feat(hub): fleet-view sources, offline rows, and spawn source targeting (component 06a)

obra:
Component 06a of the multi-host stack: the Go half of the fleet view. Frontend affordances land separately as 06b.

## What

1. **Truthful `Online` in the manifest.** `apiTreeSources()` no longer hardcodes `Online: true`. A new optional `appsource.OnlineSource{ Online() bool }` interface is implemented by `RemoteHubSource` from a `WebConfig.RemoteHostOnline` signal, which `main.go` wires to `sshconn.Manager.Attached` (true only while a live, not-closed channel exists). Local stays always-online; a source without the interface is assumed online, so all existing stubs are unaffected. `Kind` stays `"appwire"`.
2. **Offline rows stay visible but are not live.** In remote-thread ingestion, a row is appended to the live set only when its source is online. It still reaches `metas`, so an offline host's last-known sessions remain in the manifest and tree.
3. **Session targeting.** `ThreadStartParams` gains a bare `source` field. `hubThreadStart` resolves an explicit `Source` ahead of the `harness` fallback and keeps the existing `Unavailable("spawn source is not available: <id>")` shape; an empty value still defaults local.

## Design notes

- **`Dormant` is deliberately untouched.** It is documented as "a session that has never run", and the rail renders it as "has not started" in place of the age. Overloading it for an offline host would make a remote session with real history read as never-run. The offline affordance is derived on the client from `manifest.sources[].online` + the row's `host_id` (that is 06b's work). No row schema changes.
- **`RemoteHubSource.StartThread` clears `Source` before forwarding.** The field names the source in the *controller's* registry; the remote hub would otherwise resolve the controller-side host name against its own registry and fail. Covered by a test asserting the forwarded wire params carry no `source`.
- Generated wire output regenerated and committed (`make generate`): `appwire-client/typescript/types.gen.ts` and `docs/appwire-protocol.md`.

## Gates (run by the coordinator, not just the implementer)

gofmt clean; `go build ./...`; `go vet ./cmd/evener-hub/... ./appwire/...`; `go test -race` on `appsource`, `appwire`, `sshconn`; full `go test -race ./cmd/evener-hub/` **ok 447.1s**; `make lint-generated` PASS; `make test-web` PASS (1118s).

`golangci-lint` is clean on `appsource` and `appwire`. It reports 37 pre-existing findings in `sshconn`, none of which this PR introduces (`--new-from-rev=HEAD~1` → 0 new); the stack's lint debt from comp03/04a/04b is tracked separately and is being cleared as those PRs rebase onto main.

Stacked on #1296 (05d).


obra:
Round 1 addressed (`0c61f0ac3`). 4 fixed, 3 contested.

- **M offline rows not dormant** (`web_api_tree.go:322`) — contested. `TreeNode.Dormant` is documented as "a session that has never run" and is computed from the meta (`dormantFor`), not from source availability; the offline affordance is deliberately client-derived from `manifest.sources[].online` plus the row's `host_id` (`hubapi/navigation.go:169`). Marking an offline host's rows dormant would make a remote session with real history read as never-run. This is the PR's stated design and Jesse's explicit call; no change.
- **M offline source dispatch not gated** (`app_threadlifecycle.go:65`) — contested. `Online()` is `sshconn.Manager.Attached`, which is false for any reachable host that has not been lazily attached yet (`RemoteHostClient`/`RemoteHostFacts` attach on demand via `Ensure`, and `refreshHubRemoteThreads` performs its first attach inside the background goroutine after the server may already be serving). Gating `thread/start` on it would refuse valid spawns during that window and regress on-demand attach. Reads are not gated either: `hubThreadList` attempts every configured source and propagates its error. The reviewer's "required typed Unavailable" has no spec backing in this PR; genuine transport failures are already mapped to `appwire.SessionUnavailable` by `RemoteHubSource.mapCallError`.
- **M navigation not invalidated on connection-state changes** (`main.go`, `navigation_service.go:1314`) — fixed. Added `hubSSHStateInvalidation`, wired through `sshconn.Options.OnEvent` in `runMain`, invalidating navigation (`bump()` + `NavigationService.Invalidate`) on `EventAttached`/`EventDetached`/`EventFailed` — the transitions that change `Manager.Attached`, the online signal the manifest and live-row membership derive from. Pinned by `TestHubSSHStateInvalidation`. (`navigation_service.go` is outside this PR's file set, so the revision now advances through the existing `Inputs` counter that `webNavigationSource.Revision()` already reads.)
- **M harness routing hint re-routes forwarded remote start** (`remote_hub_mutations.go:59`) — fixed. `RemoteHubSource.StartThread` now clears `Harness` as well as `Source`: `launchSourceID` reads any non-`evener` Harness as a source id, so a forwarded `Harness` would re-route a start the controller already targeted at this host. Mirrors the existing `hubModelListInner` precedent (`app_models.go:72`). Pinned by `TestRemoteHubSourceStartThreadClearsRoutingHints`.
- **L `Attached` reports a link-lost channel as attached** (`sshconn/manager.go:462`) — fixed. `Attached` now also reports detached when `lost` is closed, so a dropped link is not reported online before the supervisor wakes. Pinned by `TestManagerAttached`.
- **L verbatim untrimmed `Source` forwarded** (`app_threadlifecycle.go:53`) — fixed. The forwarded params carry the normalized (trimmed) routing value the lookup used, not the verbatim padded field; a harness-routed start still leaves `Source` empty. Pinned by `TestHubThreadStartRoutesExplicitSource`.
- **L `sourceOnline` fails open for unknown IDs** (`web.go:60`) — contested, documented. Unknown non-empty IDs mean "this WebServer was built without that source" (embedder/tests), not "host removed": the registry is fixed at startup and the refresh rewrites each row's `Source` to the registered source that listed it, so ingestion never asks about an unregistered host. Returning false would drop seeded remote rows from the live set (e.g. `TestNavigationTreeInputsUsesRemoteCarriedProject`). The fail-open choice is now documented in the function comment rather than changed.

Verification: `gofmt -l cmd/evener-hub` empty; `go build ./...` ok; `go vet ./cmd/evener-hub/...` ok; `go test ./cmd/evener-hub/ ./cmd/evener-hub/internal/appsource/ ./cmd/evener-hub/internal/sshconn/ -race -count=3 -timeout 45m` ok (812.6s / 6.5s / 1.1s); `go test ./cmd/evener-hub/... -count=1` ok (13 packages, hub 110.8s); `GOLANGCI_LINT_CACHE=... MODULES="." go run ./cmd/evener-dev/bin dev module-lint` PASS (1 module, 0 findings).

obra:
Round two addressed (`2a80556c6`). 1 Low.

- **Low — `cmd/evener-hub/web_api_tree.go:318-320`, offline-source comment over-claimed a consumer.** The finding is accurate for this branch: `Source.Online` is emitted here (`apiTreeSources`) and validated by the generated TS type and `stores/navigation/{codec,store}.ts`, but nothing in 06a reads `.online` for rendering, so the comment's "the frontend derives the offline affordance from the manifest" was not true of this change. No code fix is in scope here: the consumer that keys the offline affordance off the manifest ships with the paired UI component `multi-host-pr06b-fleet-view-ui` (its `RailRow.useHostOnline` reads `source.online`, and `Spawn.tsx` disables offline hosts), and the UI files are not touched by this PR. So I reworded the comment to name that consumer honestly and to state that, within this change, an offline source's rows render as ordinary non-live rows rather than being projected live. No behavior changed.
- Pinned by the existing `TestCovAPITreeSourcesReportsOfflineSource` (the manifest emits `Online` per source, `local` stays online) and `TestNavigationSnapshotOfflineSourceRowsNotLive` (offline-source rows stay in `metas` but are excluded from `live`; online-source rows stay live). No new test added: the fix is comment-only and both halves of the contract already have coverage.

Verification:
- `go build ./...` → ok
- `go test ./cmd/evener-hub -race -count=3 -timeout 30m` → ok 782.6s
- `go test ./cmd/evener-hub/... -count=1` → ok, 13 packages
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS lint (1 modules, 62s), 0 findings

obra:
## Host-qualified archive/favorites (specs round eight): pushed 298e10a0f

Spec `06-fleet-view.md` §"Favorite and archive mutations must carry the owning host" (`06-fleet-view.md:259-299`, criterion 11).

**What changed**
- `appwire/types.go`: `ArchiveParams` and `FavoriteSetParams` gain `Source string \`json:"source,omitempty"\`` (the row's host; empty and `"local"` both name the controller). Regenerated `appwire-client/typescript/types.gen.ts` and `docs/appwire-protocol.md`.
- `hubcore.ArchiveStore` / `hubcore.FavoriteStore`: decisions key on `(source, kind, id)`; `ArchiveKey` gains `Source`; `NormalizeDecisionSource` collapses `""`/`"local"` to the controller key; an idempotent legacy-schema migration rebuilds a pre-federation `(kind, id)` table inside one transaction, moving existing rows onto the controller source key.
- `app_archive.go` (`archiveSet`): for a non-local source, no `identifier.ResolveProject(params.WorkingDir)`. The project ID is syntax-validated, `WorkingDir` becomes optional, and it is cross-checked only against the identity the host already reported (`Thread.ProjectID`/`Thread.ProjectPath` from the remote cache). Local projects keep the resolve-and-match path. The store and `navigationChangeHint.Projects` both key on `(source, id)`.
- `app_favorite.go`: `cfg.Favorite` and the navigation hint key on `(source, id)`.
- `project_delete.go`: local decision deletes pass the explicit controller source.

**Pin (test names)**
- `TestHubArchiveSetAppWireKeysNonLocalProjectBySource` — two hosts' same-ID project archives persist as separate keys (none on the controller key) and succeed with no `WorkingDir`, so no local resolution happens.
- `TestHubArchiveSetAppWireCrossChecksRemoteWorkingDir` — the reported path is accepted; a mismatched optional `WorkingDir` is rejected with the existing message.
- `TestHubFavoriteSetAppWireKeysBySource` — favorites keyed per host; `source: "local"` normalizes to the controller key.
- `hubcore`: `TestArchiveStoreKeysDecisionsBySource`, `TestArchiveStoreNormalizesLocalSource`, `TestArchiveStoreMigratesLegacyKeyToSourceColumn`.

**Ambiguity choices** (spec wording was ambiguous; smallest correct change taken):
1. The source dimension is applied to **project** targets, where IDs and paths can collide. A session's ID is already its host-qualified ref, so session decisions keep the ref-only key — adding a source dimension there would diverge from the session-decision readers and regress remote session archive read-back.
2. Tree-side read-back of a remote project's archive/favorite flag depends on the host-qualified navigation group key, which is the deferred `annotateThreadProjects` work. This change stores the correctly namespaced decision so that projection lands once the group key is qualified.
3. With a cold/absent remote cache there are no reported rows to compare, so ID syntax is the only check there.

**Newly fixed verdict findings**: none — the current RoboRev verdict on this PR is still `No issues found` at `2a80556c6`.

**Verification**
- `gofmt -l appwire cmd/evener-hub` → empty
- `go build ./...` → ok
- `go test ./appwire/... ./cmd/evener-hub/internal/hubcore/... ./cmd/evener-hub/ -race -count=3 -timeout 30m` → ok (`cmd/evener-hub` 1164.9s; the first attempt hit the default 10m package timeout on the loaded host with no assertion failure)
- `go test ./cmd/evener-hub/... -count=1` → ok (13 packages; `cmd/evener-hub` 184.0s)
- `make lint-generated` → PASS (1s)
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (1 module, 32s)

obra:
Round four addressed (`b6e2aeb`).

- **High — Remote project archive decisions are never read back (fixed).** The tree resolved a project's manual archive decision with an empty `Source`, so a decision stored under `(host, projectID)` never matched and an archived remote project stayed visible. `hubcore.TreeProject` now carries each project's owning `Sources`; `projectArchivedDecision` consults every `(source, project ID)` key (an empty list still means the controller source), and the Live tier uses the project's own source set so it agrees with the Projects tier. Pinned by `TestBuildTreeReadsBackSourceQualifiedProjectArchive` (the remote decision archives the row; the controller's same-ID project is unaffected) — verified failing against the old bare-ID lookup.
- **Medium — Favorite decisions source-qualified on write but collapsed by bare ID on read (fixed).** `FavoriteProjectAuthority` now carries `Source`; the project index/classifier key on `(source, ID)` instead of marking two hosts' same-ID projects ambiguous, and the navigation favorite map plus `projectSummary` look up the source-qualified key. Pinned by `TestClassifyFavoriteDecisions_ProjectDistinguishesSources` (verified failing with collapsed sources) and `TestProjectFavoritePresentationIsSourceQualified`.
- **Medium — `NormalizeDecisionSource` did not trim whitespace (fixed).** It now trims before the empty/`"local"` check, so `" local "` normalizes to the controller key and `"host-a "` cannot orphan a row. Pinned by `TestNormalizeDecisionSourceTrimsWhitespace`.
- **Medium — source values accepted without validating against the registered source set (fixed).** `validateDecisionSource` rejects a non-empty (normalized) source that does not name a configured remote host with `InvalidParams`; padded configured names still normalize and succeed. Pinned by `TestHubArchiveSetAppWireRejectsUnknownSource` and `TestHubFavoriteSetRejectsUnknownSource`.
- **Low — session archive silently ignored `params.Source` (fixed).** A non-local source on `kind=session` is now rejected with `InvalidParams` ("source is not supported for session archive"), since a session ref is already host-qualified. Pinned by `TestHubArchiveSetAppWireRejectsSessionSource`.
- **Low — `projectsChangeKey` was inert (fixed by removal).** `navigationChangeHint.Projects` is never consumed by `commitTargetsLocked`, so the source-qualified hint advertised a safeguard that did not exist; the qualification, its call sites, and the misleading comment are removed. Project invalidation continues to flow from the navigation source fingerprint delta (unchanged, covered by `TestAppWireArchiveNavigationConvergence`).

Verification:
- `go build ./...` → exit 0
- `go test ./cmd/evener-hub/ ./cmd/evener-hub/internal/hubcore/ -race -count=3 -timeout 40m` → PASS (cmd/evener-hub 1340.8s, hubcore 65.2s). The default 10m per-package timeout is too short for `-count=3` under `-race` on this loaded host (the first run hit that timeout, not a data race); the raised timeout is the gate.
- `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → exit 0
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (1 module, 110s)

Files changed: `app_archive.go` (+`_test`), `app_favorite.go` (+`_test`), `internal/hubcore/archive.go` (+`_test`), `internal/hubcore/favorite_authority.go` (+`_coverage_test`, `_test`), `internal/hubcore/tree.go` (+`_test`, `_covtest`), `navigation_projection.go`, `web_api_tree.go` (+`_test`).

Note: `internal/hubcore/tree.go`, `internal/hubcore/favorite_authority.go`, and `navigation_projection.go` are not in the PR's current diff but are named by the round-four findings (`tree.go:1593-1599`/`1420`, `navigation_projection.go:881`, `favorite_authority.go:325`) and are required to fix them.

obra:
Round five addressed (`0d4d0fe`).

- **Medium — legacy decision-table migration rechecks nothing under the write lock (`internal/hubcore/archive.go:197-220`).** The schema check ran outside the transaction, so two concurrent first writes could both decide to migrate; the loser then rebuilt the already-upgraded table and re-inserted its source-qualified rows under the controller source. `ensureDecisionSourceColumn` now serializes in-process migrations with `decisionMigrationMu`, takes SQLite's write lock with `BEGIN IMMEDIATE` on a pinned connection, and rechecks the schema under that lock before rebuilding. A `decisionMigrationInterleave` test seam pins the race deterministically (no sleeps): the competing migrator upgrades the table and writes `(host-a, proj-a)`; the loser must leave it alone. Pinned by `TestArchiveStoreMigrationRechecksSchemaUnderWriteLock` — verified failing against the old code (the host-a row collapsed to the controller key). `tableHasColumn` now runs over either `*sql.DB` or the locked `*sql.Conn`.
- **Medium — `Ensure` returned a link-lost channel (`internal/sshconn/manager.go:238`).** Reuse tested only `!isClosed()`, ignoring `linkLost()`, while `Attached` already reported detached in the `markLost` window. Reuse is now `!isClosed() && !linkLost()`. Pinned by `TestEnsureReplacesLinkLostChannel` — verified failing (it returned the dead channel and started no reattach).
- **Medium — Live tier filtered by the merged project source set (`internal/hubcore/tree.go:1447-1451`).** A project merging one ID/path across hosts carries the union of sources, so `projectArchivedDecision` archived on any source and the Live tier hid a second host's still-live session. The filter now consults `liveEntrySource(entry)` for the entry's own source; the merged `projectSourcesByID` index is gone. Pinned by `TestBuildTreeLiveFilterUsesEntrySourceForMergedProject` (host-a's decision clears only host-a's session; host-b's stays) — verified failing (both dropped). Scope note: the Projects-tier row still merges and archives on any source; only the cited Live filter changed, so round four's "Live agrees with Projects" claim for single-source projects is preserved.
- **Medium — source-less project favorites never presented (`internal/hubcore/favorite_authority.go:339-341`, writer `app_favorite.go:33`).** The rail sends no `source`, so the decision stores under the controller key and could never match a host-keyed remote authority: `OK:true` with no star, pre-existing bare remote favorites orphaned. The rail itself ships with the paired UI component (`multi-host-pr06b-fleet-view-ui`), outside this PR's file set, so I added the compatibility path the reviewer allowed: `resolveProjectAuthority` still prefers an exact `(source, ID)` match (a source-less decision keeps addressing the controller's own project when one exists) and, only for a source-less decision with no exact authority, resolves to the single authority across sources for that ID. Two sources claiming the ID — or any ambiguous candidate — stays dormant, so this does not weaken round four's source-separated classification. Pinned by `TestClassifyFavoriteDecisions_EmptySourceResolvesSingleAuthority`, `TestClassifyFavoriteDecisions_EmptySourcePrefersControllerAuthority`, and end-to-end `TestProjectFavoriteBareDecisionResolvesRemoteProject` (classifier → presentation → `projectFavoriteForSources`); all verified failing against the exact-match-only classifier.
- **Low — source validation did not match the registered source set (`app_archive.go:125-134`).** `validateDecisionSource` checked `cfg.RemoteHosts`, but `newHubSourceRegistry` registers no remote source when `RemoteHostClient` is nil, so a decision could persist for a source no read path addresses. It now rejects a non-empty source when `RemoteHostClient` is nil. Pinned by `TestHubArchiveSetRejectsConfiguredHostWithoutRemoteClient` (unwired hub rejects; wired hub accepts) — verified failing (the unwired archive succeeded). Handler tests that exercise a *configured* host now pass the same `unusedRemoteHostClient` the registry test already uses, so they still pin a host name rather than an unwired hub; no assertion was weakened.
- **Low — `SnapshotContext` shallow-copied `TreeProject.Sources` (`internal/hubcore/tree.go:76-83`).** The struct copy shared the slice backing array with the retained tree. `cloneTreeProjectsContext` now clones it, keeping a nil list nil and an empty list non-nil. Pinned by `TestTreeSnapshotClonesProjectSources` — verified failing (mutating the snapshot changed the tree).
- **Low — `sourceOnline` looked up `thread.Source` verbatim (`web.go:68`).** A padded source missed the registry and fail-opened to `true`, projecting an offline source's rows as live. The lookup now trims first. Only `TrimSpace` applies, not `NormalizeDecisionSource`: the registry keys the controller's own source as `"local"`, so the `"local" → ""` mapping would bypass that entry rather than normalize it. Round one's documented fail-open for genuinely unknown ids is unchanged. Pinned by `TestSourceOnlineTrimsPaddedRegisteredID` and `TestNavigationSnapshotPaddedOfflineSourceRowNotLive` — verified failing (the padded offline row appeared live).

Verification:
- `go build ./...` → exit 0
- `go test ./cmd/evener-hub/internal/hubcore/ -race -count=3 -timeout 20m` → ok (82.6s)
- `go test ./cmd/evener-hub/internal/sshconn/ -race -count=3 -timeout 20m` → ok (1.2s)
- `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok, 13 packages (hub 214.1s)
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (1 module, 14s), 0 findings

Files changed: `app_archive.go` (+`_test`), `app_favorite_test.go`, `internal/hubcore/archive.go` (+`_test`), `internal/hubcore/favorite_authority.go` (+`_test`), `internal/hubcore/tree.go` (+`_test`), `internal/sshconn/manager.go` (+`_test`), `web.go`, `web_api_tree_test.go`, `web_covtest_test.go`.

obra:
Round six addressed (`6064cfb54`).

- **Medium — merged projects used "any owning source" for archive placement** (`cmd/evener-hub/internal/hubcore/tree.go`): a merged row (the same canonical ID/path on two hosts) moved to `ArchivedProjects` as soon as *any* owning source's `(source, project ID)` decision archived it, while the per-entry Live filter in the same change consults only the entry's own source — so host-b's still-live session stayed in the Live rail under a project row that had moved to the archived catalog, and the shared project left the active project catalog on host-a's decision. Row placement now requires **every** owning source to have archived the project (`projectArchivedByEverySource`, used at the placement site); the age rule (`no Current/Recent rows`) and `projectArchivedDecision`'s per-entry use are unchanged, so a single-source project still moves on its one owner's decision. Pinned by `TestBuildTreeLiveFilterUsesEntrySourceForMergedProject`, extended to assert `Projects` vs `ArchivedProjects` alongside `Live` (and that both sources archiving *does* move the row and clear `Live`); it failed before the fix with `projects = [], want the merged row in the active catalog`.
- **Rail archives the bare session ID instead of the source-qualified ref** (`Rail.tsx:1212`, `app_archive.go:30-31`) — Go half fixed: the handler now normalizes a session archive ID to the identity its row is read under (`hubcore.NormalizeDecisionSessionID`): a remote row's canonical ref is kept (trimmed, canonical spelling), and a `local:thread` ref collapses onto the bare session ID every local row — and every stored local decision — is keyed by, which is the reviewer's "send the canonical ref, normalize local refs server-side" contract. Pinned by `TestHubArchiveSetAppWireNormalizesLocalSessionRef` (failed before the fix: the decision landed on `"  local:session-1  "`, a key no read path consults), the table test `TestNormalizeDecisionSessionID`, and `TestHubArchiveSetAppWireArchivesRemoteSessionByRef`, which drives it end to end (archive the remote ref → the row leaves the tree's Live tier).
  **Honest gap:** the rail still calls `setArchived("session", session.session_id, …)`, so a *remote* session archive stays inert until the rail sends `session.ref`. `Rail.tsx`/`actions.ts` are not files this PR touches (component 06b / PR #1302 owns the rail frontend), so that half was left to that component rather than edited here.
- **Source-qualified project APIs not propagated through the rail** (`actions.ts:41,86-87`, `Rail.tsx:1243-1253`, `app_archive.go:40-55`) — the Go side is already this PR's round-five work: `ArchiveParams.Source` / `FavoriteSetParams.Source` exist, are validated against registered sources (`validateDecisionSource`), and key both stores by `(source, id)` (`TestHubArchiveSetAppWireKeysNonLocalProjectBySource`, `TestHubArchiveSetAppWireCrossChecksRemoteWorkingDir`, `TestHubArchiveSetAppWireRejectsUnknownSource`, `TestHubArchiveSetRejectsConfiguredHostWithoutRemoteClient`). **Honest gap:** the rail still omits `source`, and there is no `source` on `NavigationProjectSummary` for it to read one from — that type lives in `hubapi/navigation.go` and the rail in the 06b component stacked on this branch, neither of which is in this PR's file set, so no code changed for this clause here.
- **Remote project rows still offer "Delete project" with no source-aware semantics** (`RailRow.tsx:367-389`, `Rail.tsx:1273`, `project_delete.go:97-112`) — Go half fixed: `ProjectDeleteParams` gains `Source` (wire + generated TS + protocol doc), and `projectDelete` refuses any non-local or unknown source with a typed `InvalidParams` error *before* `Key`/`WorkingDir` are resolved or anything is removed (`project delete is local-only; <source> is not this hub`, `unknown source: <source>`). Deletion is local-only, so a source-naming client can never turn a remote row into a controller-local delete of a project that merely shares its ID or path; empty and `local` still address the controller's own project, leaving local behaviour unchanged. Pinned by `TestProjectDeleteRefusesNonLocalSource` — refusal ordering is pinned with an unresolvable key/path, no session artifact or decision row was removed, and `""`/`local`/padded-`local` still pass the gate. **Honest gap:** the rail neither sends `Source` nor hides the item for remote rows yet; hiding it is the frontend half in `RailRow.tsx` (06b).

Verification: `gofmt -l` over the touched dirs → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./cmd/evener-hub/internal/hubcore/ -race -count=3` → ok (59.3s); `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (all 13 packages); `MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (1 module, 0 findings). No frontend files changed, so `make test-web` was not run.

obra:
Round seven addressed (`a60ce1adf`).

- **HIGH — remote project deletion treated as local (fixed).** Project summaries now carry their owning sources: the navigation read model grew `sources` (`hubapi.NavigationProjectSummary.Sources`, projected in `navigation_projection.go` from `hubcore.TreeProject.Sources` with the controller's own source spelled `"local"` and omitted for a controller-only project, validated in `navigation_schema.go`, regenerated into `appwire-client/typescript/types.gen.ts`). `Rail.tsx` refuses `onDeleteProjectRequest` for any project a remote host also owns — before the confirmation dialog opens — and `actions.ts`'s `deleteProject` refuses a non-local owner again, so the only delete the UI can issue is the controller-local one `project_delete.go` still accepts, and a controller-only project keeps the exact previous request (no `source` field). Pinned by `Rail.test.tsx` "refuses to delete a project row that a remote host also owns" + "deletes a controller-only project through the dialog with no source field", and `actions.test.ts` "deleteProject refuses a remote-only project without issuing a request" / "refuses a merged project instead of deleting this hub's own rows" / "keeps the plain local request for a controller-only project".
- **MEDIUM — remote session archive used the bare `session_id` (fixed).** Both session-archive call sites (`shell/rail/Rail.tsx` row menu, `panes/session/chrome/SessionChrome.tsx` pane menu) now send the canonical `session.ref`. The hub's `NormalizeDecisionSessionID` collapses `local:<id>` to the bare ID local decisions already use, so local read-back is unchanged, while a remote row's decision lands under the host-qualified identity its read path consults. Pinned by `Rail.test.tsx` "retains an archive overlay through AppWire response…" (now pins `id: "local:a"`) and `SessionChrome.test.tsx` "session-menu archive addresses the canonical ref so a remote row's decision sticks". Two existing archive-parameter assertions pinned the bare `session_id` that this finding calls the bug (`SessionChrome.edge.test.tsx` "archive toggle failure toasts an error" and the same `Rail.test.tsx` overlay test); round seven supersedes those expectations — same params shape, same toast and rollback coverage, only the identity corrected — and both are called out in the commit.
- **MEDIUM — project favorite and archive omitted `source` (fixed).** `actions.ts` classifies a summary's owners (`projectOwnership`) and `setFavorite`/`setArchived` issue one request per owning source: the controller's own source omits the field exactly as before, each host name is sent as `source`. A merged project archives under every owner (the tree moves a row only once every owner archived it) and a favorite is set and cleared on every owner (the read side shows it when any owner holds it); the fan-out is sequential so the last receipt's navigation mutation reflects every decision. Pinned by `Rail.test.tsx` "routes a remote project row's favorite and archive through its owning source" + "addresses every owner of a merged project row and omits the controller's own source", and `actions.test.ts` "source-qualified project mutations".
- **MEDIUM — SSH `Ensure` can leak a link-lost channel's transport resources (out of this change; verified, not modified here).** The branch's `Ensure` (`cmd/evener-hub/internal/sshconn/manager.go:238-240`) declines a link-lost channel and reattaches, but it never retires the old one, and this branch's own `TestEnsureReplacesLinkLostChannel` exercises exactly that path. **Exact scenario:** the link drops (`markLost` closes `lost`); the old supervisor wakes and blocks on the per-host lock; a concurrent `Ensure` wins that lock, sees the link-lost channel, runs a full preflight/attach (seconds) and publishes the replacement with `setChannel`; the old supervisor then takes the lock, sees `m.currentChannel(host.Name) != ch` and returns at `manager.go:374-381` **before** its `ch.Close()`. Why that skips the close: a replaced channel is no longer current, so the early return treats it as retired — but the replacement's supervisor watches the replacement, nobody else owns the old channel, and `Manager.Close` only walks `m.chans`, which no longer holds it. The old ssh child, its stdio pipes, its transport and the AppWire read loop therefore live until process exit. **Proposed fix (upstream's, already on `main`):** `origin/main`'s `Ensure` captures `stale := m.currentChannel(name)` and defers `stale.Close()` — "Retire the replaced channel on every path out of here, early returns included: a Close racing this attach must not leave the old ssh child, its pipes, and its transport alive" — while main's `supervise` keeps the same not-current early return *because* Ensure now retires it. `sshconn` was deliberately not touched here; the stack's merge from `main` must keep that retire-close (this branch edits the same `Ensure` lines, so that merge is the risk point for it).

Verification: `gofmt -l` over the touched Go paths → empty; `go build ./...` → OK; `go vet ./cmd/evener-hub/...` → OK; `go test ./cmd/evener-hub/internal/hubcore/ -race -count=3` → `ok primeradiant.com/evener/cmd/evener-hub/internal/hubcore 78.670s`; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → all 13 packages `ok` (0 FAIL); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → `PASS lint (1 modules, 111s)`; `make lint-generated` → `PASS lint-generated`; frontend: `npx biome check --write` on the touched files → clean, `npx tsc --noEmit` → clean, `make test-web` → `PASS web-typecheck` / `PASS web-test` / `PASS web-lint` (exit 0), `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` → `PASS web-layoutguard` / `PASS web-overflowguard` / `PASS web-shellguard` / `PASS web-spawnguard` / `PASS web-transcriptscrollguard` / `PASS web-skillguard` (exit 0).

obra:
Round eight addressed (`bb0d89a`).

- **High — Frontend project decoder rejects backend `sources` key.** Confirmed on `a60ce1a`: the projector emits `sources` on every non-controller-only project summary (`cmd/evener-hub/navigation_projection.go:885` → `navigationProjectSources`, bounded by `maxNavigationProjectSources = 65`), and the normalizer emits that summary verbatim as the catalog entity's value (`navigation_normalize.go:57-58`), while `codec.ts` validated a project with `exactKeys` against a `PROJECT_OPTIONAL` list that had no `sources`. Entity validation is fail-closed, so one such row rejected the whole catalog resource before the rail could render it. `sources` now joins `PROJECT_OPTIONAL` and is validated as a list of at most 65 non-empty identities (≤ 1024 UTF-8 bytes), the projector's own bound; nothing else was loosened. The project anchor path was checked too and is unaffected: the normalizer emits it as `navigationProjectAnchor{Key}` (`navigation_normalize.go:74-75`), exactly the `{key}` the decoder already requires. Pinned by `codec.test.ts` "codec decodes a catalog whose project rows carry owning sources" (a two-row catalog — merged `["local","host-a"]` beside remote-only `["host-a"]` — decodes as a snapshot, `[]` is accepted as the controller default, and `""`, bare `"local"`, `null`, `{}`, `[""]`, `["local",""]`, `["local",7]`, `["local",undefined]`, `["local","x".repeat(1025)]`, a 66-entry list, and the manifest's object shape all still fail the resource closed) and "codec keeps the manifest's source records and a project's owner names apart" (manifest metadata `sources` still rejects a project-style name list, so the two same-named shapes cannot validate each other).
- **Low — Per-source fan-out aborts on first rejection, with no compensation.** Confirmed: `actions.ts` awaited each owner in order and threw at the first rejection, so a merged project's committed decision was left half-applied (or a failing controller request kept the remote owner unasked) while the call rejected. Both `setFavorite` and `setArchived` now issue every owner's request and settle all of them with `Promise.allSettled`; the returned receipt carries `failedSources` naming every owner that did not answer, and `setFavorite` also returns the derived `favorite` value the read side presents — a set is presented once an owner committed it, a clear only once every owner answered (a failed owner keeps its decision). Only a fan-out no owner answered rejects, which leaves the single-owner error path unchanged. `Rail.tsx` takes the favorite overlay's value from that settled result and raises a warning naming the owners that did not answer, for both favorites and archives; `projectOwnership(["local","host-a"])` is unchanged and still classifies as `{local: true, hosts: ["host-a"]}` (pinned by the existing "classifies the owning sources a summary reports" test and by the new rail cases, which assert the local request keeps omitting the field while every host request carries `source`). Pinned by `actions.test.ts` ("per-source fan-out settlement": a host failure still asks every owner and reports it, a failing controller request still asks the remote owner, a partial clear keeps the presented value, an all-owner rejection names each owner and rejects, an all-answering fan-out reports the requested value, and a partial archive keeps its receipt) plus `Rail.test.tsx` ("asks every owner of a merged project favorite and warns about the one that failed", "warns about the owner a merged project archive could not reach").

Verification:
- `gofmt -l cmd/evener-hub hubapi appwire` → empty (no Go files in this unit)
- `go build ./...` → exit 0
- `go test ./cmd/evener-hub/ -run 'Navigation|Ownership|Project' -count=1` → ok, 25.7s
- `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → exit 0 (`cmd/evener-hub` 144.5s)
- `GOLANGCI_LINT_CACHE=… MODULES=. go run ./cmd/evener-dev/bin dev module-lint` → PASS lint (1 modules, 81s), 0 findings
- `npx biome check --write` on the six touched files → clean, no fixes applied
- `npx tsc --noEmit` → exit 0
- `make test-web` → PASS (web-typecheck, web-test, web-lint)
- `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` → PASS web-layoutguard, PASS web-overflowguard, PASS web-shellguard, PASS web-spawnguard, PASS web-transcriptscrollguard, PASS web-skillguard (exit 0)

obra:
Round nine addressed (`179dc5cad`). Six findings, all fixed.

- **Medium — SSH channel leak on repeated connection losses** (`cmd/evener-hub/internal/sshconn/manager.go`): confirmed. `Ensure` replaced a link-lost channel, and its supervisor — which wakes only after `Ensure` releases the per-host lock, by which point the map holds the replacement — returned through `supervise`'s stale-channel check without closing anything, leaking the ssh child and its transport on every loss. Chose the first fix shape: `closeReplaced` runs in `Ensure` while it still holds the host lock, right after a successful attach, taking the superseded channel out of the map and closing it. A failed attach leaves `prev` current (so its supervisor still owns the reconnect), and `Close` is idempotent, so a supervisor that won the lock first makes this a no-op rather than a double close. Pinned by `TestEnsureClosesEveryChannelItReplaces` (`manager_test.go`): it replaces three link-lost channels in a row and asserts each ends closed while the channel just installed stays open; without the fix it fails with `round 0: replaced channel left open`.

- **Medium — Native client sends source-less archive/favorite mutations** (`mobile-native/src/ProjectsScreen.tsx`): confirmed. The project rows' `organization` payload was unconditional, so the client's source-less `evener/archive/set` / `evener/favorite/set` addressed only this hub's decision for a merged or remote-only project. Chose the reviewer's alternative (withdraw the actions) over source-aware fan-out: the native client's `NavigationActions` journal keeps exactly one operation and one durable recovery checkpoint, so per-source fan-out would need multi-owner recovery semantics — much larger than this finding's risk warrants. `controllerOwnedProject(row.sources)` (mirroring the session rows' existing `row.host_id !== "local"` guard) returns null for any project the summary reports with a host source, so no source-less request can be issued for one. Verified by the client's own checks: `tsc --noEmit --project tsconfig.check.json` clean and `vitest run` 79 files / 740 tests pass. No failing-first test: the native screens have no component test harness, and this guard is a render-time refusal inside JSX.

- **Medium — `setArchived` fan-out sends the same `workingDir` to every source** (`cmd/evener-hub/frontend/src/shell/rail/actions.ts`): confirmed. `workingDir` is the controller's path, and `validateHostProjectArchive` rejects a host leg whose `workingDir` differs from the path that host reported for the ID, so a merged project with different local and remote paths failed its remote leg on every toggle. Now only the local leg carries it — the controller still requires it, and a host leg is keyed by `(source, id)` alone. Pinned by the updated fan-out test and by `a remote-only project archive carries the owning source without a workingDir`; both fail against the old wire shape. This supersedes round eight's `a remote-only project archive carries the owning source alongside workingDir`, which asserted the rejected shape.

- **Medium — Bare delete on a merged project not guarded server-side** (`cmd/evener-hub/project_delete.go`): confirmed. A source-less request reached the local delete path even when the tree reports the project with a host source, so the wire could half-delete a merged project the rail refuses client-side. Chose "check merged ownership server-side and reject ambiguous bare deletes" over "require an explicit local source": `NormalizeDecisionSource("local")` is the empty decision key, so an explicit local source is wire-indistinguishable from a bare one and that shape would have needed every caller to change while leaving the ambiguous form reachable. `projectRemoteSources(matched.Sources)` now rejects a bare delete when the tree reports any host owner, with the same wording the rail uses. Pinned by `TestProjectDeleteRefusesBareDeleteOfMergedProject` (fails first: the delete succeeded and removed rows), alongside the existing controller-only bare-delete test (`TestProjectDeleteRemovesFilesAndScrubs`) and the rail-side refusal tests, so the two directions cannot drift.

- **Low — `setFavorite` return value flips optimistic state incorrectly on failure** (`actions.ts`): confirmed. The clear branch returned `failures.length > 0`, so an unreachable owner presented the row as favorited even when no owner held one. `favorite` now comes from committed state via `settledFavorite`: a set is presented once any owner committed it, and a clear only when every owner answered or the caller already knew no owner held one — the rail passes the row's own presented favorite (`favoritedBefore`), so clearing a project nobody had favorited cannot flip the row back. `sources` and the options object are now required parameters, so no caller can silently fall back to a guess. Pinned by `clearing an unfavorited project with an unreachable owner stays unfavorited` (fails first: it returned `true`) and `a partial project favorite set is presented from the owner that committed it`; round eight's "keeps the value the read side may still present" assertion is retained and made explicit with `favoritedBefore: true`.

- **Low — Stale doc comment on `resolveProjectAuthority`** (`cmd/evener-hub/internal/hubcore/favorite_authority.go`): confirmed. The comment justified the cross-source fallback with "The rail's project favorite sends no source", which this series changed. Reworded: the fallback exists for decisions persisted before source qualification (and clients that still omit `source`), and the rail now qualifies its project favorites, so its decisions resolve exactly and the fallback is not on the rail's path. Comment only; no behavior change.

Verification: `gofmt -l` (touched Go dirs) empty; `go build ./...` 0; `go vet ./cmd/evener-hub/...` 0; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=3 -timeout 20m` ok; `go test ./cmd/evener-hub/internal/hubcore/ -race -count=1` ok; `go test ./cmd/evener-hub/ -run 'Navigation|Ownership|Project|Delete' -count=1` ok; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` all packages ok; module-lint PASS (1 module, 94s, 0 findings); frontend `biome check --write` on the touched files "No fixes applied", `tsc --noEmit` clean, `make test-web` PASS (web-typecheck, web-test, web-lint); `make build-web` + `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` PASS on the first run (web-layoutguard, web-overflowguard, web-shellguard, web-spawnguard, web-transcriptscrollguard, web-skillguard); native client `tsc --noEmit --project tsconfig.check.json` clean and `vitest run` 79 files / 740 tests pass.

obra:

The parent components (04b, 05a, 05b, 05c, 05d) landed, so this component's own 11 commits were rebased onto `main` (`git rebase --onto origin/main 22ffa3790`) and force-pushed with `--force-with-lease`. The PR's diff is now exactly this component's, against `main`.

This rebase needed reconciliation rather than only conflict marking: several things this branch was written against had already moved on `main`, and in three places the branch's version of a helper is now redundant. Each decision below keeps `main`'s landed behavior and re-pins this component's intent to it:

1. **`Ensure`'s early usable-channel fast path is dropped.** `main` deleted it (its `TestEnsureCanceledContextDoesNotReturnALiveChannel` requires a canceled caller to be refused *before* any channel is handed back) and replaced it with the `beforeHostGate` test seam. This branch refined that fast path (`!ch.linkLost()`); the refinement is dropped, restoring `main`'s ordering.
2. **`setChannel` → `publishChannel`, `isLost` keeps `main`'s name.** This branch renamed `isLost` to `linkLost` and re-added `setChannel`; `main` had already renamed the setter and kept `isLost`, so the branch's call sites in `manager_test.go` are updated to `main`'s API instead of the reverse.
3. **`closeReplaced` is dropped.** `main` retires a superseded channel through `Ensure`'s deferred close (every path out, early returns included, with its own comment explaining why); this branch's helper does the same job, so applying both would close the replaced channel twice and publish different ownership rules.
4. **The remote spawn path keeps `main`'s neutralized harness.** Both values (`"evener"`, or empty) neutralize the controller-side routing hint, but `main`'s landed `TestRemoteHubMutationWireMethodsAndRefs` pins `"evener"`, so the branch's later `Harness = ""` assignment is dropped and this component's own routing-hint test is re-pinned to the neutralized default rather than to an absent key.
5. Everything else merged as unions: the `hubcore.WebConfig` additions (`RemoteHostOnline` alongside `main`'s newer `RemoteHostFacts` signature), `Manager.Attached`, the `remote_hub_source.go` struct fields, the errors/imports, the two test functions, and the `navigation_projection.go` const block.

The commits above the replay (``02106d446`, `23eca92f7``) carry exactly these reconciliations, each with its reasoning in the commit message.

**Gates on the rebased tree (`23eca92f7`)**: `gofmt -l cmd/evener-hub/` clean; `go build ./...` exit 0; `go vet ./cmd/evener-hub/...` exit 0; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=2` ok (4.1s); `go test ./cmd/evener-hub/internal/appsource/ -race -count=1` ok (21.4s); `go test ./cmd/evener-hub/ -count=1 -timeout 30m` ok (124.3s); module-lint PASS (1 module, 0 findings).

obra:

**CI** — all required checks green at `23eca92f7` (16 total: 15 CI checks plus RoboRev's own `roborev` check, all `pass`).

**Review** — RoboRev's verdict for `23eca92f7` (the rebased head): **two findings, both Low; explicitly "no functional, security, or regression issues found in the changed behavior."** Under the campaign's stated rule (a Lows-only verdict lands, with the Lows recorded as an immediate follow-up), both are recorded here:

- `cmd/evener-hub/internal/sshconn/manager.go:1688-1694` — the `Manager.Attached` doc comment carries two spliced versions of the same sentence block (a rebase artifact; the series' own `docs(hub): fix the awkward Attached doc comment wording` commit intended exactly this fix and the aggregate still carries the garble). Fix: keep the single coherent block ("…usable channel: one that has been established, has not been closed, and has not lost its link…") and drop the spliced first two lines.
- `cmd/evener-hub/internal/hubcore/config.go:93-94` — the new `RemoteHostOnline` field comment starts mid-sentence with no subject; unlike its neighbours a maintainer cannot tell what it reports or what its zero value means. Fix: e.g. "RemoteHostOnline reports whether the controller's channel to `host` is currently attached; nil leaves every remote host online (tests)."

The rebase onto `main` and its reconciliation decisions (dropped `Ensure` fast path, `publishChannel`/`isLost` naming, dropped `closeReplaced`, neutralized harness) are documented in the rebase-note comment above.

**Up-to-date override** — the branch is 4 commits behind `main`. Overlap check, run immediately before this comment:

```
$ gh pr view 1299 --json mergeable,mergeStateStatus --jq '.mergeable, .mergeStateStatus'
MERGEABLE
BEHIND

$ comm -12 <(gh pr view 1299 --json files --jq '.files[].path' | sort) \
           <(git diff --name-only "$(git merge-base origin/main origin/multi-host-pr06a-fleet-view-go)" origin/main | sort)

$ git diff --name-only "$(git merge-base origin/main origin/multi-host-pr06a-fleet-view-go)" origin/main | wc -l
64
```

`main` changed 64 files since the merge base; none of them is a file this PR touches.

PR #1302, opened 2026-09-14
feat(hub): frontend host picker and rail host badge (component 06b)

obra:
Frontend half of the multi-host fleet view (component 06b), stacked on 06a (#1299).

## What

- **Host picker** in the new-session form (`Spawn.tsx`): populated from the navigation manifest's `sources`, local preselected, offline hosts rendered but disabled with the reason in the option text. Rendered only when the manifest lists a non-local source, so the common single-host form is unchanged.
- **Launch seam** (`startThread.ts`): optional `source` on `SpawnRequest`; `params.source` is set only for a non-local choice and omitted entirely for local, so a local spawn's `thread/start` params stay byte-identical to before.
- **Draft persistence** (`spawnDrafts.ts`): the choice lives in the spawn draft (`source`, default `"local"`) and survives navigation like the other fields.
- **Selector** (`stores/navigation/selectors.ts`): new `selectSources`, exposing `manifest.data.sources` with a stable empty array before the manifest loads.
- **Rail row host label** (`RailRow.tsx` / `.module.css`): non-local rows show their host inline after the title; a host whose manifest source has `online:false` gets a visible offline marker. Offline-ness comes from `manifest.sources[].online` keyed by the row's `host_id`, defaulting to online when the host is unknown.

## Design note

Per a standing decision, the row's Dormant/"Not started" state keeps its "never run" meaning; the offline affordance is derived from the manifest sources rather than repurposing Dormant, and no row schema changed.

## Tests

- `Spawn.test.tsx`: picker lists sources, preselects local, disables offline; no picker when only local; picking a non-local host sets `source` and persists it in the draft; a local choice omits `source`.
- `startThread.test.ts`: omits `source` for local/absent, sets it for non-local.
- `RailRow.test.tsx`: badge for non-local, none for local, offline affordance on/off, unknown host defaults online.

## Gates (run by the coordinator, on this branch)

- `make test-web` — PASS (typecheck, full vitest, biome).
- `make test-web-browser` — PASS (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard).
- `make lint-generated` — PASS.

Frontend-only: no Go, generated, or out-of-scope file touched.


obra:
Round 1 addressed (`d324442e3`). 1 Medium + 1 Low.

- **Medium — Offline selected host remains submittable** (`Spawn.tsx:389`): confirmed real. `hostChoice` fell back to `"local"` only when the draft's source was absent from the manifest. When a host went offline while the draft lived, the hub keeps it listed (`online: false`), so the guard passed, `hostChoice` stayed on the offline host, the disabled `<option>` remained the controlled value, and Start submitted `source: <offline-host>` — rejected by the hub with `"spawn source is not available"`, contradicting the picker's own "Offline hosts can't be selected" help text. Fixed by resolving the matched candidate and treating a non-online source like a missing one: `const chosenSource = sources.find((candidate) => candidate.id === source); const hostChoice = chosenSource?.online ? chosenSource.id : "local";`. The stale draft value is untouched, so a host that comes back online is selected again. Pinned by new test `a host that goes offline while a draft names it falls back to local` (seeded online, chosen, then re-seeded offline; asserts the select resolves to `local` and `thread/start` omits `source`). Written red first: before the fix it failed with `expected 'buildbox' to be 'local'`.
- **Low — Stale-source fallback branch is untested** (`Spawn.tsx:386-389`): confirmed real (coverage gap, not a defect). The removed-host fallback path had no test. Pinned by new test `a draft naming a host removed from the manifest shows local and omits source` (draft source set to a host not in the manifest; asserts the select shows `local` and `thread/start` omits `source`).

No changes to the rail host badge, the local `source` omission, `selectSources`, or `Dormant`. Frontend-only diff: `Spawn.tsx` + `Spawn.test.tsx`.

Verification: `npx biome check --write src/panes/spawn/Spawn.tsx src/panes/spawn/Spawn.test.tsx` → PASS (no fixes applied); `make test-web` → PASS (web-typecheck, web-test, web-lint); `make test-web-browser` → PASS (web-layoutguard, web-overflowguard, web-shellguard, web-spawnguard, web-transcriptscrollguard, web-skillguard); `git diff --stat` → only the two frontend files.

obra:
Round two addressed (`6bc16a940`). 1 Low.

- **Low — Stale source field after offline fallback** (`Spawn.tsx:392-394` and the picker `onChange` at ~1590): confirmed real, both halves. The picker is controlled by the derived `hostChoice`, so when a chosen host went offline the select read `local` while `draft.source` still held the remote id (`"buildbox"`). Selecting "Local" fires no `change` event (the value already reads local), so nothing rewrote the draft; when the host came back online `hostChoice` re-derived to `"buildbox"` and silently flipped **both** the select and the submitted `source: hostChoice` back to the remote host. Fixed by writing the effective choice into the draft: `useEffect(() => { if (sources.length > 0 && source !== hostChoice) setSource(hostChoice); }, [sources.length, source, hostChoice, setSource])`. The `sources.length > 0` guard matters — before the manifest loads, `hostChoice` is a provisional `"local"`, and rewriting then would discard a persisted remote host that is merely still loading. The draft is now the single source of truth for the choice, so the displayed value and the submitted source cannot diverge (this intentionally drops round one's "a host that returns online is chosen again" side effect — that side effect *was* the silent revert). Pinned by new test `a host that returns online after its offline fallback stays on local`: written red first, failing with `expected 'buildbox' to be 'local'` at the draft assertion (and, with that assertion masked, at the post-return select assertion — proving the revert itself, not just the field).
- Round one's tests are unchanged and still green: `a host that goes offline while a draft names it falls back to local` and `a draft naming a host removed from the manifest shows local and omits source`. No assertion was weakened. No changes to the rail host badge, `selectSources`, or `Dormant`.

Verification: `npx biome check --write src/panes/spawn/Spawn.tsx src/panes/spawn/Spawn.test.tsx` → PASS (no fixes applied); `make test-web` → PASS (web-typecheck, web-test, web-lint); `make test-web-browser` → PASS (web-layoutguard, web-overflowguard, web-shellguard, web-spawnguard, web-transcriptscrollguard, web-skillguard); `git diff --stat d324442e3 HEAD` → only `cmd/evener-hub/frontend/src/panes/spawn/Spawn.tsx` (+15/-2) and `Spawn.test.tsx` (+46).

obra:
Round three addressed (`9969ee72f`). 1 Medium.

- **Spawn.tsx:397-403,1389 — submitting while the navigation manifest is loading/unavailable silently fell back to local.** Confirmed real. `selectSources` is empty until the manifest loads, so `hostChoice` read `"local"` in that window; the request sent `source: hostChoice`, and `startThread` drops `"local"` from the wire — so a draft that still named a remote host started the session locally with no indication. Fixed by deriving the submitted source from the draft's own value until a settled manifest can confirm the fallback: `const submittedSource = sources.length > 0 ? hostChoice : source;`, used at the `thread/start` call (the same `sources.length > 0` gate the draft write-back already uses). A settled manifest that omits/offlines the host still falls back to local exactly as round one intended; the picker and the round-one/round-two tests are unchanged. Pinned by the new test **"a draft naming a remote host submits it while the manifest is loading"** in `Spawn.test.tsx` (fails before the fix — `source` omitted — passes after).

Verification:
- `npx biome check --write src/panes/spawn/Spawn.tsx src/panes/spawn/Spawn.test.tsx` → clean, no fixes applied
- `make test-web` → PASS (web-typecheck, web-test, web-lint)
- `make test-web-browser` → PASS (web-layoutguard, web-overflowguard, web-shellguard, web-spawnguard, web-transcriptscrollguard, web-skillguard)
- `git diff --stat` → frontend-only: `Spawn.tsx` (+17/-3), `Spawn.test.tsx` (+47)

obra:
Round four addressed (`b69c7b678`).

- **`cmd/evener-hub/frontend/src/panes/spawn/Spawn.tsx:1400,1471` (a remote host only affects `thread/start`; directory validation and creation still run against the controller's local filesystem)** — Fixed for the part this branch can carry. The controller-local preflight (`evener/path/validate` -> `evener/dirs/create`) is now skipped when the launch target is remote, so a remote-only path is no longer read as "missing" and "Create & start" can no longer create the directory on the controller before the remote launch fails. The remote hub is the authority: `appsource.RemoteHubSource.StartThread` forwards `thread/start` (clearing the controller's source id) and that hub validates its own cwd and spawns on its own host. Test: `a remote host submission skips the controller-local directory preflight`. The boundary is pinned by `a local submission keeps the controller-local directory preflight` (a local target still runs the preflight and still offers creation).
- **`cmd/evener-hub/frontend/src/panes/spawn/Spawn.tsx:762-810,1400` (model lists, provider readiness, launch resolution, plugin previews and slash catalogs are controller-local and not keyed by source)** — Fixed for the part this branch can carry; the routing half is contested. The controller's readiness must not BLOCK a remote launch, so `providerRequired` and `modelRequired` are no longer computed from controller-local state for a remote target (a local "missing provider" or "no default model" is no evidence the selected host cannot start), and the form now discloses that the working directory, models, providers and plugins shown come from the controller. Tests: `a remote host submission is not blocked by missing controller-local providers`, `a remote host selection discloses that discovery is controller-local`. Threading the selected source through `model/list`, `evener/launch/resolve`, `evener/launch/schema`, `evener/plugin/preview`, `evener/spawn/slashCatalog`, `evener/harnesses/list`, `evener/paths/complete`, `evener/path/validate`, `evener/dirs/create` and `evener/git/head` is the routing half, and it cannot land in this PR — see below.

**Contest — the source-aware routing cannot land in this PR:**

1. `evener/host/request` does not exist in this branch's tree (`rg "host/request"` finds nothing).
2. Its catalog row and hub handler are in PR #1361 (component 07a, base `multi-host-pr06b-fleet-view-ui`, head `multi-host-pr07a-host-admin`): `appwire.MethodEvenerHostRequest = "evener/host/request"` with `ScopeHub` in `appwire/protocol.go`, the handler registered in `cmd/evener-hub/app_host_admin.go`, plus the regenerated `appwire-client/typescript/types.gen.ts`. This branch's generated `MethodName` union — the bound of the typed surface every call site uses, `AppwireClient.request<M extends MethodName>` (`appwire-client/typescript/client.ts:389`) — does not include it, so the frontend here cannot even spell the method: it would be a TypeScript error before it could be a routing bug.
3. `appwire-client/typescript/types.gen.ts` is emitted from `appwire/protocol.go` by `internal/appwirets` (`appwire/doc.go` go:generate; the drift test is `TestGeneratedFileCurrent`), so it cannot be hand-extended here: adding the row means an `appwire/protocol.go` change plus `make generate`, i.e. files outside this PR's frontend set.
4. Adding the catalog row without a registered handler fails the lockstep test `cmd/evener-hub/appwire_catalog_test.go` `TestHubRouterMatchesCatalog` ("cataloged but NOT registered"). The handler exists only in 07a.
5. 07a's proxy allow-list already marks every method these two findings name as forwardable discovery — `evener/path/validate`, `evener/paths/complete`, `evener/dirs/create`, `evener/git/head`, `evener/harnesses/list`, `evener/launch/resolve`, `evener/launch/schema`, `model/list`, `evener/plugin/preview`, `evener/spawn/slashCatalog` (`TestHostAdminAllowListMatchesCatalog`). So the routing is the intended follow-on and lands in the PR stacked above 07a, which can also key those caches per source.

No existing test was weakened. The three new behavior tests were RED before the `Spawn.tsx` change and GREEN after.

Verification: `npx biome check --write` on the two touched `src/` files -> clean ("No fixes applied"); `make test-web` -> PASS (web-typecheck, web-test, web-lint); `make test-web-browser` -> PASS (web-layoutguard, web-overflowguard, web-shellguard, web-spawnguard, web-transcriptscrollguard, web-skillguard — the first run hit the known load-racy skillguard timeout at 15s, and the identical re-run passed, so it was not a real failure); `git diff --stat` -> frontend only: `cmd/evener-hub/frontend/src/panes/spawn/Spawn.tsx` (+57/-16), `cmd/evener-hub/frontend/src/panes/spawn/Spawn.test.tsx` (+107/-1).

obra:
Follow-up on round four — verification evidence for the disposition in the previous comment (`b69c7b678`; head is now `296dc5d00`).

**1. The dependency contest, re-verified empirically (not just read):**

- `evener/host/request` is absent from this branch's tree: `git grep "host/request" HEAD` matches only the code comments this PR adds — no catalog row, no method constant, no handler.
- The typed surface genuinely cannot spell it. Temporary probe, `client.request("evener/host/request", {})` against `AppwireClientLike`:
  ```
  src/panes/spawn/__probe_host_request.ts(6,23): error TS2345: Argument of type '"evener/host/request"' is not assignable to parameter of type '"initialize" | "ping" | "thread/list" | ... | "evener/sandbox/escalation/resolve"'.
  ```
  That union is the generated `MethodName` (`appwire-client/typescript/types.gen.ts`), i.e. the bound of `AppwireClient.request<M extends MethodName>` (`appwire-client/typescript/client.ts:389`). Probe deleted; `tsc --noEmit` clean afterwards.
- Adding a row without a handler really does break the lockstep test. Temporary probe: one ScopeHub catalog row `{"evener/host/request", EmptyParams{}, EmptyResponse{}, ScopeHub, …}` inserted into `appwire/protocol.go` (reverted immediately, tree left clean) →
  ```
  --- FAIL: TestHubRouterMatchesCatalog (0.21s)
      appwire_catalog_test.go:33: hub router vs appwire catalog mismatch:
            cataloged but NOT registered: [evener/host/request]
            registered but NOT cataloged: []
  ```
  The row and its handler must land together; both are in PR #1361 (07a, base `multi-host-pr06b-fleet-view-ui`), which also regenerates `types.gen.ts`.

**2. New falsifiable tests that the finding-1 fix is not symptom-hiding:**

- `a remote launch rejected by the selected host surfaces that host's own error` — remote host selected, cwd is one the controller calls a creatable missing directory; the launch reaches `thread/start` and the selected host's own `WireError` surfaces verbatim (`Start failed: cwd: no such file or directory on buildbox`), with zero `evener/dirs/create` calls and no Create & start dialog. Feedback still exists — the authoritative feedback — instead of the controller's wrong early judgment.
- `a remote launch the selected host cannot serve reports that host's own failure` — controller-local provider discovery is missing, the remote target is submitted anyway, and the host's own `provider credentials missing for anthropic` failure is what appears, with no session navigation. So relaxing the controller-local gate does not make a doomed remote launch succeed silently: "allowed to submit" and "actually succeeded" stay distinguishable.

Against `9969ee72f`'s `Spawn.tsx`, 5 of the 6 remote tests fail (`…skips the controller-local directory preflight`, `…not blocked by missing controller-local providers`, `…discloses that discovery is controller-local`, `…surfaces that host's own error`, `…reports that host's own failure`); the boundary test `a local submission keeps the controller-local directory preflight` passes before and after.

**3. Wording made precise.** The notice reads: "Running on `<host>`. Path checks, models, providers and plugins below come from this controller; `<host>` resolves its own environment when the session starts." Every clause maps to real code — `evener/path/validate` + `evener/paths/complete`, `model/list`, `evener/instance/list`, `evener/plugin/preview` on the controller; `RemoteHubSource.StartThread` forwarding to the remote hub's own `hubThreadStart` on the host side — and the disclosure test asserts both owners by name so the copy cannot drift.

**Contest (unchanged) — the source-aware routing cannot land in this PR:**

1. `evener/host/request` is not in this branch's tree.
2. Its catalog row (`appwire.MethodEvenerHostRequest`, ScopeHub, `appwire/protocol.go`) and hub handler (`cmd/evener-hub/app_host_admin.go`) are in PR #1361 (07a), which also regenerates `appwire-client/typescript/types.gen.ts`. This branch's `MethodName` union does not include it, so the frontend cannot spell the call (proved above).
3. `types.gen.ts` is emitted from `appwire/protocol.go` by `internal/appwirets` (go:generate in `appwire/doc.go`; drift test `TestGeneratedFileCurrent`), so it cannot be hand-extended here.
4. Adding the row without a handler fails `cmd/evener-hub/appwire_catalog_test.go` `TestHubRouterMatchesCatalog` (proved above).
5. 07a's allow-list already marks every method the two findings name as forwardable discovery (`evener/path/validate`, `evener/paths/complete`, `evener/dirs/create`, `evener/git/head`, `evener/harnesses/list`, `evener/launch/resolve`, `evener/launch/schema`, `model/list`, `evener/plugin/preview`, `evener/spawn/slashCatalog`), so the routing lands in the PR stacked above 07a, which can also key those caches per source.

Verification: `npx biome check --write` on the two touched `src/` files -> clean ("No fixes applied"); `make test-web` -> PASS (web-typecheck, web-test, web-lint); `make test-web-browser` -> PASS (web-layoutguard, web-overflowguard, web-shellguard, web-spawnguard, web-transcriptscrollguard, web-skillguard); `git diff --stat` -> frontend only: this follow-up `Spawn.tsx` (+12/-9), `Spawn.test.tsx` (+72/-0); cumulative from `9969ee72f` to this head `Spawn.tsx` (+60/-16), `Spawn.test.tsx` (+179/-1).

obra:
Round five addressed (`ab13891c6`).

- **Stale manifest leaks into `selectSources`** — the selector returned `state.manifest?.data?.sources` even while the resource was loading, stale, or errored, so an invalidation/reconnect could expose a retained snapshot naming a host the fresh manifest had since removed or taken offline. It now returns the stable empty array unless the manifest is settled (`!loading && !stale && no error`); a consumer that must not lose a persisted choice while the manifest is in flight (the spawn draft) keeps carrying its own source instead of reading the empty list as a fallback. Pinned by `selectSources exposes a settled manifest's sources and withholds an unsettled one` and `selectSources returns the same empty snapshot whenever it withholds sources` in `src/stores/navigation/selectors.test.ts`.
- **Controller-local plugin preview gates remote launches** — `pluginSelectionBlocked` was unconditional, so a controller-local preview that was still loading, or that reported a selected plugin as unavailable, disabled Start and made `doSpawn` return even for a remote target. The gate is now `!remoteLaunch && (...)`: the selection is forwarded and the selected host validates it at start, the same authority model the launch already uses for cwd and providers. Pinned by `a remote host submission is not blocked by controller-local plugin issues` in `Spawn.test.tsx`.
- **Remote launches pre-validate `/model` and `/reasoning-effort` against the controller catalog** — a model or effort level supported only by the selected host was rejected locally before `thread/start`. For a remote target the controller-catalog match is now skipped and the value is forwarded verbatim (a model split as `provider/model`, the same convention every other `/model` path uses; an effort level as typed). Bare `/model` still means "(default)" and bare `/reasoning-effort` still fails closed, since neither is a catalog judgment. Pinned by `a remote host does not pre-validate /model against the controller catalog` and `a remote host does not pre-validate /reasoning-effort against the controller ladder` in `Spawn.test.tsx`.

Verification: `npx biome check --write` clean on the four touched `src/` files; `make test-web` PASS (`web-typecheck`, `web-test`, `web-lint`); `make test-web-browser` — `web-layoutguard`, `web-overflowguard`, `web-shellguard`, `web-spawnguard`, `web-transcriptscrollguard` PASS on every run; `web-skillguard` is the known load-racy guard and failed under this host's contention (load 39-59) at three different browser timing waits — `expected two live sessions in the rail, found 0` (the failure dump captured both rows present), `timed out after 15000ms waiting for steered turn's input`, and `timeout calling navigateTo after 30000ms` — while passing clean on a re-run of the guard itself (`ok primeradiant.com/evener/cmd/evener-hub 52.290s`). No gate failure is in a code path this diff touches. Diff is frontend-only: 4 files, all under `cmd/evener-hub/frontend/src`.

obra:
Round seven addressed (`446f522da`). 2 Medium + 1 Low.

- **Medium — Remote default-model fallback leaks controller catalog to remote host** (`Spawn.tsx` :725-734 region; the `defaultPreview` effect): confirmed real. The async uncredentialed-default fallback reads THIS controller's `model/list` and installs `models[0]` when `evener/launch/resolve` names a default whose provider isn't launchable here. For a remote target that value rides `thread/start`, so the selected host can't resolve its own default. Fixed by (a) skipping the fallback when `remoteLaunch` is true, and (b) retiring a fallback value already installed while the target was local when the target becomes remote — a provenance ref records the exact value the fallback installed, and any user model change clears it, so a person's own pick (or a sticky draft default) is never treated as fallback-derived. Pinned by `a remote host does not inherit the controller's uncredentialed-default fallback model` and `a controller-catalog fallback model is retired when the target becomes remote` (both RED before the fix).

- **Medium — Unqualified `/model` value is mis-split on the remote path** (`Spawn.tsx` :1347-1349): confirmed real, with one correction to the stated mechanism. `/model foo` does split to `modelProvider: "foo"` with an empty `model`; but `startThread` omits the falsy empty model from the wire, and the host's `thread/start` ignores a `modelProvider` with no `model` (`app_threadlifecycle.go:80`, `if params.Model != ""`), so the requested model was **silently dropped** rather than rejected with a confusing `"foo/"` message. Either way raw user text was mishandled on the remote path. Fixed by requiring the remote split to yield a non-empty provider AND model; otherwise the same `/model: unknown value "foo"` message the local path uses is toasted and the submit aborts before `thread/start`. Pinned by `a remote /model value without a provider is refused before launch` (RED before; before the fix it reached `thread/start` with the model dropped).

- **Low — Remote launch persists the remote cwd as the controller's global last working directory** (`Spawn.tsx` :1462 / `spawnDefaults.ts` :146): confirmed real. Fixed by threading `remoteLaunch` into `saveDefaults` and skipping both global scalar writes (`GLOBAL_WORKING_DIR_KEY` and `GLOBAL_MODEL_KEY`) for a remote submit; the per-project blob keyed by the submitted cwd is still written. I extended the same guard to `GLOBAL_MODEL_KEY` rather than leaving it for a later round: it is the identical defect class on the same branch — a host-only model would default the next LOCAL spawn to a model this hub may not serve. Pinned by `a remote launch does not persist its cwd as the controller's global working directory` (RED before); the existing local-side assertion (`Spawn.test.tsx:1743`, local submit still writes the global working dir) is unchanged.

No prior disposition was weakened or reversed: rounds one–five's tests and the routing contest stand untouched. Diff is frontend-only (`git diff --stat ab13891c6 HEAD`): `Spawn.tsx` +60/-, `Spawn.test.tsx` +147, `spawnDefaults.ts` +17/-.

Verification: `npx biome check --write` on the three touched `src/` files → clean, "No fixes applied"; `npx tsc --noEmit` (frontend) → clean; `make test-web` → PASS (web-typecheck, web-test, web-lint); `npx vitest run src/panes/spawn/Spawn.test.tsx` → Test Files 1 passed (1), Tests 224 passed (224); `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` → PASS (web-layoutguard, web-overflowguard, web-shellguard, web-spawnguard, web-transcriptscrollguard, web-skillguard). The browser gate needed the session-bus workaround for the CDP guard timeouts (unsetting `DBUS_SESSION_BUS_ADDRESS`); with it, all six guards passed on the first run.

obra:
Round eight addressed (`4bab604b`).

**Medium — remote model selections polluted the cwd-only per-project defaults** (`spawnDefaults.ts:143-150`). Verified: a remote submit wrote its model into the blob keyed by the cwd alone, and that path is often also a local checkout, so a later local spawn of the same path loaded a host-only model (one the stale-model sweep cannot even recognize as wrong — an unknown provider is deliberately left alone). Fixed: `saveDefaults` no longer persists `model` into the cwd blob when `remoteLaunch` is set. The project's host-generic layer (harness, access mode, effort) is still written, so round seven's "the blob is still written" intent holds minus the host-specific value. Pinned by `spawnDefaults.test.ts` — "omits the model from the cwd blob for a remote launch, but keeps its other fields (round eight)" — and by `Spawn.test.tsx` — "a remote launch's model does not become the cwd's local project default", which drives a full remote submit and then re-mounts a fresh local draft for the same cwd and asserts Model reads `(default)`.

**Medium — fallback-model provenance leaked across drafts and was lost on remount** (`Spawn.tsx:626`, `:335`, `:1096-1102`). Verified both halves: the marker was a `SpawnForm`-local ref while `SpawnForm` is a singleton reused across drafts with no `key`, so draft A's installed value could clear an identical string draft B legitimately owned (its user pick or sticky default) when B went remote; and a pane remount dropped the ref entirely, so the controller's fallback model rode a remote launch after all. Fixed by storing the marker in the draft (`DraftFields.defaultModelFallback`, `spawnDrafts.ts`), so provenance is per-draft by construction and survives remounts; the installer, the retire effect, and `handleModelChange` all read/write that field. Pinned by `Spawn.test.tsx` — "a fallback installed for one draft does not retire another draft's identical sticky model" and "a controller-catalog fallback model is still retired after a pane remount" (both fail on the previous head).

Beyond the two triaged Mediums (same files, same round — say the word and I back these out):

- **Medium — host picker stayed enabled during the local directory preflight** (`Spawn.tsx:1583-1600`, `:1721-1725`). Verified: a submit snapshots the source in its closure and then awaits preflight, so a host change mid-submit launched on the snapshot while the picker showed the new host, and `saveDefaults` used the snapshot too. Fixed: the picker is `disabled={busy}`, with a `busyRef` guard covering the same-tick window before that attribute commits. Pinned by "the host picker cannot change while a submit is in flight".
- **Low — remote cwd leaked into the controller-wide picker seed** (`Spawn.tsx:1698-1700`, `:2049`). Verified: `setGlobalLastWorkingDir` (the dir-picker's last-accepted-directory seed, not the form prefill) was written on both pick paths regardless of target. Fixed: `commitLastWorkingDir` skips the write when the draft's target is remote. Pinned by "a cwd picked for a remote target is not recorded as the controller's picker seed", which also asserts a local pick still updates the seed.
- **Low — no test for preserving explicit model picks across a host switch.** Added "a user's explicit model pick is preserved when the target becomes remote": the fallback installs one model, the person picks a different one from the picker, the host goes remote, and the trigger keeps it while `thread/start` forwards it.

Prior dispositions intact: host-picker preselection and offline fallback, spawn-draft persistence and keying, the routing/view-ownership seam, the preserved local preflight path, and the remote `/model` refusal are untouched, and no existing assertion was changed or removed. The retire rule is still "only a value still equal to the one the fallback installed is cleared" — that mark is now per-draft.

Verification: `npx biome check --write` on the touched files — clean, no fixes applied; `npx tsc --noEmit` — exit 0; `make test-web` — PASS (web-typecheck, web-test, web-lint); `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` — PASS 6/6 guards (layout, overflow, shell, spawn, transcript-scroll, skill); `git diff --stat` — 5 files, all under `cmd/evener-hub/frontend/src/panes/spawn/`. The seven new tests were also re-run with the source fixes stashed: six of them fail there (the seventh is the explicit-pick invariant guard), so they pin the fixes rather than the implementation.

obra:
Round nine addressed (`4c3e964`).

**Medium — Remote path validation is not source-aware** (`src/panes/spawn/Spawn.tsx:840`, `AdvancedOptions.tsx:156` via the injected `validatePath`) — fixed.

Both halves confirmed in the code: the picker's `browse()` uses `evener/path/validate`'s verdict as the selectability gate, so a directory that exists only on the selected host could never be picked; and Advanced Options' `updateScalar` marks the field `invalid` on that same verdict, which `collectAdvancedOverrides` (`schema.ts:41`) then drops from `launchOverrides` — silently. Fixed at the pane's injected seam rather than per widget: `validatePath` now short-circuits to `{ valid: true, path }` (the typed spelling, no controller canonicalization) for a remote target, and the selected host validates its own cwd and paths when the session starts — the authority model the cwd preflight, provider, plugin and model gates on this pane already follow. Local targets still run the RPC unchanged. One consequence had to be handled with it: that bypass makes the picker's "New folder" reachable for a remote-only path, and `evener/dirs/create` is a `MkdirAll` on THIS hub's filesystem (`docs/appwire-protocol.md`, `hubDirsCreate`), so `createDirectory` now refuses for a remote target (toast: `Create the folder on <host>: this controller can't create it there.`) instead of materializing the host's path locally. The pane's disclosure copy no longer claims path *checks* come from the controller (`…path checks included` of the host's own resolution); the two substrings its existing test asserts are untouched.

Pinned by `a remote target can select a directory this controller cannot see`, `the picker does not create folders locally for a remote target`, and `a remote target's advanced path value is not dropped for being invisible here` (`Spawn.test.tsx`) — each also asserts the LOCAL gate is unchanged (the same path stays unpickable with the controller's error, and the invalid value stays out of `launchOverrides`).

**Medium — Remote launch erases prior local per-project model default** (`src/panes/spawn/spawnDefaults.ts:158`) — fixed.

Confirmed: the blob write replaces the whole JSON blob, so `remoteLaunch`'s model omission meant "this path has no model", not "this submit names none" — a later local launch of the same cwd lost its sticky model. Round eight's intent is preserved by the launch still writing **no model of its own** (neither the cwd blob nor the global scalars ever receive the host-specific value), while the remote path now reads the existing blob and carries its `model` over verbatim; a fresh local page's `resolveInitialDefaults` layers it straight back.

Pinned by `a remote launch keeps the cwd blob's existing local model (round nine)` plus `a remote launch onto a path with no stored model still writes no model (round eight)` (`spawnDefaults.test.ts`). Round eight's own test is unchanged and still green.

**Low — Launch-gating `selectSources` reused for display** (`src/stores/navigation/selectors.ts:72`, `src/shell/rail/RailRow.tsx:492`, `src/panes/spawn/Spawn.tsx:392`) — fixed.

Confirmed: `selectSources` withholds on loading/stale/error, and both display consumers read the withheld list as "unknown" — the rail badge (`source ? source.online : true`) turned every offline host online, and the picker vanished (the settled list drove its visibility). Added `selectDisplaySources` (last-known `manifest.data.sources`, same stable-array contract `useSyncExternalStore` needs) for display only: the rail badge, the picker's visibility/options/offline labels, and the picker's own value in the window where the manifest is unsettled — where `submittedSource` still carries the draft's host, so the visible choice matches what actually submits. `selectSources` remains the sole input to `hostChoice`, `submittedSource` and the draft write-back: no host becomes launchable merely because the UI still shows it, and the fresh manifest's own reading wins the moment it settles.

Pinned by `selectDisplaySources keeps the last-known sources while the manifest is unsettled`, `selectDisplaySources hands back the store's own array…`, `the two views differ exactly where settledness does` (`selectors.test.ts`), `an offline host's badge survives a manifest revalidation` (`RailRow.test.tsx`), and `the host picker keeps its host while the manifest revalidates` (`Spawn.test.tsx`, which also asserts the unsettled submit still targets the draft's host).

Prior dispositions intact: the host picker's preselection/offline fallback (settled path is the same rule, byte for byte), the draft-keyed fallback provenance (`DraftFields.defaultModelFallback`), the routing/view-ownership seam, the preserved local preflight, the cwd-only global-picker-seed gating, and the busy/picker-disable guard. No assertion was weakened or deleted; the two `seedSources` test helpers gained an optional resource-overrides parameter only.

Verification:
- `npx biome check --write` on the eight touched files → `Checked 8 files in 512ms. No fixes applied.` (exit 0)
- `npx tsc --noEmit` (frontend dir) → exit 0
- `make test-web` → PASS web-typecheck, PASS web-test, PASS web-lint (exit 0)
- `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` → PASS web-layoutguard, PASS web-overflowguard, PASS web-shellguard, PASS web-spawnguard, PASS web-transcriptscrollguard, PASS web-skillguard (exit 0)
- `git diff --stat` → frontend-only: the eight files above, all already in this PR's file set
- Failing-first: reverting the four production files (tests kept) turns every new test red — spawnDefaults (the blob loses `model`), RailRow (no `rail-row-host-offline`), and all four new Spawn tests (the remote-only path's confirm stays disabled, `evener/dirs/create` was called, `launchOverrides` is `undefined`, and `Unable to find a label with the text of: Host` when the picker vanishes mid-revalidation).

obra:
Round ten addressed (`5cd6b039c`).

**Medium — Remote drafts still validate model against the controller's `model/list`** (`src/panes/spawn/Spawn.tsx:987-999`) — fixed.

Confirmed in the code: the stale-model effect registered on the global catalog for every draft, and its async guard read only `active`/`model`/`harness` — nothing about the target. A remote target's model belongs to the selected host (its own hub resolves the model at `thread/start`), so a model this controller's `model/list` does not list was blanked and stamped "discarded last-used model" before the launch could carry it. The sweep is now skipped for a remote target (`|| remoteLaunch` on the effect), and the target mode is re-read when the catalog lands (`remoteLaunchRef`), so a draft that switches hosts while the request is in flight is not judged by the run it started as local. Local targets keep the sweep unchanged.

**Medium — Path marked invalid while local stays dropped after switching to a remote host** (`src/panes/spawn/Spawn.tsx:415-420` and `:853-863`, `AdvancedOptions.tsx:156-176`) — fixed.

Confirmed: `AdvancedOptions.updateScalar` persists `invalid: true` in the draft's `advancedValues` from the controller's verdict, `collectAdvancedOverrides` (`schema.ts:41`) then drops that field from `launchOverrides`, and nothing re-validated when the Host picker changed who judges the path (round nine's `validatePath` returns `valid: true` for a remote target). The pane now re-validates the stored path-kind values whenever the target mode changes — writing values, messages and the collected overrides together, exactly as the panel's own `updateScalar` does, with the same "a later edit owns the field" guard — so the flag, the message and `launchOverrides` always describe the current target. A later pass under a remote target (schema arriving late, or a draft switch) also retires a flag a local era left behind, since that verdict needs no wire call; a local pass still asks this controller, so a path this host really cannot see is dropped only while this host is the judge.

**Carried rail gap (a) — `Rail.tsx` archived the bare session ID** — verified OPEN, fixed. `Rail.tsx:1212` sent `setArchived("session", session.session_id, …)`, which for a remote row is an identity no read path consults (the row is read under its host-qualified ref). It now sends `session.ref` (`Rail.tsx:1217`); a `local:<id>` ref normalizes back to the bare local ID server-side, so every local decision is unchanged. Supersedes one existing assertion, called out here: `Rail.test.tsx`'s archive-overlay test pinned `id: "a"` — the bare ID this finding calls the bug — and now pins `id: "local:a"` (same params shape, same overlay and toast coverage; the same supersession 06a's round seven recorded for its own copy).

**Carried rail gap (c) — project delete neither sent `Source` nor was withheld for remote rows** — verified OPEN, fixed for the half this component owns. `RailRow.tsx:389-393` offered `Delete project…` for every project row, and `Rail.tsx:1257` → `:1273` issued the source-less `deleteProject(key, workingDir)` (`actions.ts:93-99` posts `{key, workingDir}`) — the local-only request `project_delete.go` refuses for a non-local source, so a merged or remote-only row's item could only have deleted this hub's own project of the same ID or path. The item is now withheld from any row a remote host also owns (`RailRow.tsx:384`'s ownership verdict — the summary's own `sources` when the read model carries them, unioned with the row's own sessions' `host_id`, which is what this tree carries today), and the row's memo comparator compares that verdict, so a descendant-only rebuild still leaves the row alone. The request half (`Source`) is already on the base (06a `a60ce1adf`: `Rail.tsx`'s merged-project refusal plus `actions.ts`'s source-qualified `deleteProject`) and is deliberately not duplicated here.

Prior dispositions intact: the host picker (local preselected, offline disabled), the source threaded through `startThread` and omitted for local, spawn-draft persistence, `selectSources`/`selectDisplaySources`, the rail host badge and its offline affordance, round nine's remote path validation (`validatePath`'s short-circuit is unchanged — the new pass only re-reads stored verdicts), and the remote `saveDefaults` behavior. Nothing else was weakened or deleted.

Verification:
- `npx biome check --write` on the six touched `src/` files → `Checked 6 files in 507ms. No fixes applied.` (exit 0)
- `npx tsc --noEmit` (frontend) → exit 0
- `make test-web` → PASS web-typecheck, PASS web-test, PASS web-lint (exit 0)
- `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` → PASS web-layoutguard, PASS web-overflowguard, PASS web-shellguard, PASS web-spawnguard, PASS web-transcriptscrollguard, PASS web-skillguard (exit 0; run with `env -u DBUS_SESSION_BUS_ADDRESS`)
- Failing-first (production files reverted, tests kept): every new test goes red — 3 in `Spawn.test.tsx` (the stale model reads `(default)` instead of `openai/retired` twice, and the switched-host path field still carries its `<p role=alert>` error) and 5 in the rail suites (`id: "a"`/`id: "t1"` instead of the canonical ref; three `Delete project…` items still rendered).
- `git diff --stat` → frontend-only: the six files above (`Spawn.tsx`/`Spawn.test.tsx`/`Rail.tsx`/`Rail.test.tsx`/`RailRow.tsx`/`RailRow.test.tsx`).

obra:
Round eleven addressed (`f84fbee8f`).

**High — Local session archive/unarchive no longer persists.** Confirmed end to end on this branch. A local session's wire `ref` is `local:<id>` (`navigation_projection.go`'s `navigationNodeRef`/`navigationRef`: `host_id` stays `local`, `session_id` stays bare), so round ten's `setArchived("session", session.ref, …)` sent `local:<id>`. `app_archive.go`'s `archiveSet` stores the id verbatim, and the read path consults bare ids: `hubcore`'s `decisionFor` is called with the local node id (`tree.go` builds local nodes with `ID: m.ID`; `web_api_tree.go` stamps local entries with `SessionID: past.Meta.ID`) and with `tierEligible`'s bare `le.SessionID`. `archiveDecisions()` reaches `hubBuildNavigationTree`/`hubDeriveNavigationAttention` verbatim in both `web_api_tree.go` and `navigation_service.go` — no alias expansion, unlike favorites' `ClassifyFavoriteDecisions` — so the stored key was inert, and the optimistic `hideSession` overlay (keyed by that same ref, `railPending.ts`) is what made the archive look applied until the next read.

- What changed: the conditional at the send site (the reviewer's minimal option). `Rail.tsx` now has one exported `archiveSessionIdentity(session)` — bare `session_id` when `host_id === "local"`, `session.ref` otherwise — used by `onToggleArchiveSession` in both directions. The remote half is unchanged and still correct: a remote thread's `meta.ID`/`entry.SessionID` *are* its host-qualified ref (`appThreadTreeEntries`).
- Which test pins it: `Rail.test.tsx`'s local archive test now asserts `{ kind: "session", id: "a", archived: true }` (it had been changed to `id: "local:a"`, baking in the wrong assumption); the unchanged remote test still asserts `buildbox:t1`; a new `archiveSessionIdentity` contract test covers both branches. Sensitivity verified: reverting the send site to `session.ref` fails the local test.
- What could not be pinned: the stored-key-to-read-model half in Go. `cmd/evener-hub/app_archive.go` is not in this PR's file set, so normalizing in `archiveSet` (the preferred shape) would widen the PR.

**Tracked follow-up — normalize in `archiveSet` (`cmd/evener-hub/app_archive.go`).** That file is outside this PR's file set and still stores whatever id it is handed, so any client sending `local:<id>` strands its decision under a key no reader consults. The durable fix is to map a parsed local ref to its bare session id there, mirroring `favoriteSessionAliases`/`LocalSessionDecisionAliases`, with a Go test that archives via `local:<id>` and asserts the tier/attention. The frontend now sends the key the read model consults, so nothing in this component depends on it.

**Adjacent pre-existing gap, outside this round's file set (observation, not changed).** `cmd/evener-hub/frontend/src/panes/session/chrome/SessionChrome.tsx:230` archives with `menuSession.session_id` for every session — the mirror image: a REMOTE session archived from the session pane stores its decision under the bare remote id, which its read path never consults. It predates this PR (present at the stack's merge-base `60dd24497`), and its `menuSession` already carries `host_id`, so `archiveSessionIdentity` would fix it.

**Low — The project-row ownership safeguard reads a field that is not on the wire.** Right about this branch, and the field does land with the base — verified: `multi-host-pr06a-fleet-view-go` head `179dc5cad` (not an ancestor of this head; merge-base `60dd24497`) carries `Sources` on `hubapi.NavigationProjectSummary`, `sources?: string[]` in the regenerated `appwire-client/typescript/types.gen.ts`, and `"sources"` in the navigation codec's `PROJECT_OPTIONAL` — which matters, because the codec's `exactKeys` rejects a summary field it does not allow, so the field and its admission land together. The value reaches the row: both `projectFromGraph` and `projectFromSummary` spread the wire summary into `RailProject`.

- What changed: the structural read is kept and made authoritative when present — a present `sources` list settles ownership on its own, loaded or not. Without it, the sessions reading is the live path on this branch, and it is tightened: a collapsed project loads its sessions on expand (`Rail.tsx`'s loadProject effect skips unexpanded rows), so `sessions: []` on an unloaded row is no longer read as controller-only.
- The `sessions: []` decision: the delete item is **withheld**, not offered, while ownership is unknown. `project_delete.go` resolves the local path and deletes by key+path, so offering it for a merged row deletes the controller's local twin — the wrong-host item the guard exists to prevent. This is a UI-truthfulness call, as noted, not a safety one. The item returns as soon as the row's sessions load, and immediately for every controller-only project once `sources` arrives with the base. Kept exception: a project with no rows at all (`session_count === 0`, which counts every source's rows via `hubcore.TreeProject.TotalSessionCount`) cannot be remote-owned, so an empty controller project keeps the item.
- Which test pins it: new tests "menu omits 'Delete project…' when the project's sessions have not loaded" (verified sensitive — it fails against the round-ten guard) and "menu keeps 'Delete project…' for an empty controller project". The merged and remote-only tests now set `loaded: true`, so they exercise the sessions path instead of passing for the unloaded reason; the two `withSources` tests exercise the post-base path and their comment now says so rather than presenting it as live-branch behavior.
- What could not be pinned: the base's `sources` field itself, which arrives with the base — those two `withSources` tests stay fixture-stamped rather than wire-driven.

**Verification:** `npx biome check --write` on the four touched `src/` files — clean; `npx tsc --noEmit` — clean; `make test-web` — PASS (web-typecheck, web-test, web-lint); `make build-web` — PASS (browser-guard prerequisite); `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` — PASS web-layoutguard, PASS web-overflowguard, PASS web-shellguard, PASS web-spawnguard, PASS web-transcriptscrollguard, FAIL web-skillguard. Re-ran that guard once: it failed at a different point (`timeout calling navigateTo after 30000ms` in `browserGuardCdp.mjs`) than the first attempt (`timed out after 15000ms waiting for steered turn's input visible in the transcript`), on a host at load 37–43 on 16 cores with ~890 Chrome processes from sibling lanes; neither timeout touches the surfaces this diff changes, but I could not confirm a pass here. No Go files were touched, so no Go gates apply.

obra:

The parent components (04b, 05a, 05b, 05c, 05d, 06a) landed, so this component's own 12 commits were rebased onto `main` (`git rebase --onto origin/main 60dd24497`) and force-pushed with `--force-with-lease`. The PR's diff is now exactly this component's - 14 frontend files under `cmd/evener-hub/frontend/src/` - against `main`.

`main` had moved substantially in these files (the AppWire SDK migration: `protocol/` deleted and imported by package name, `pathRows`/`launchSchema`/`spawnSchema` moved into the package, plus the guided-provider-connection onboarding work and the landing of 06a's read model), so the rebase needed reconciliation rather than only conflict marking. Each decision below keeps `main`'s landed behavior and re-pins this component's intent to it:

1. **Import sources re-pinned to the package.** `NavigationManifest`/`Source`/`ResourceState` and the navigation state helpers now come from `@evener/appwire-client` and `@evener/appwire-client/state/navigation`; `basename` comes from the package (the `../../widgets/pathfield/pathRows` copy is gone); `PathValidation`, `schemaPathKind`, `AdvancedValues` and `collectAdvancedOverrides` come from the package (`./schema` and `launchShared/{schema,pathListAdd}` no longer exist); the test helper `manifest` comes from `@evener/appwire-client/testing/navigation`.
2. **`modelRequired` keeps `main`'s onboarding-scoped form plus this component's remote-launch exemption**: `!remoteLaunch && usesEvenerModels && model === "" && advancedModel === "" && (noDefaultModel || providerChoiceScopes.current.has(...))`. `main`'s landed version added the unmanaged-harness ("external") and Advanced-override guards; this component's fix ("the controller's environment must not gate a remote spawn") is preserved on top.
3. **The fallback-substitution site keeps `main`'s guards** (Evener-model harness, no Advanced override set, not a provider-choice scope) **and adds this component's provenance recording**, so a fallback this controller installs is still retired when the target becomes remote (rounds seven and eight) - now against the draft-scoped `defaultModelFallback` field.
4. **`selectSources` → `selectDisplaySources` in the rail** (this component's split of the withheld vs last-known sources view) while `main`'s SDK import paths are kept. The `canReadSharedNotes` import is `main`'s (package) one; the branch's old `../../protocol/sharedNotesAvailability` line is dropped.
5. **Tests merged as unions**: `main`'s `selectSessionWatches`/cache-bound tests and this component's `selectSources`/`selectDisplaySources` tests both stand; the archive-ref expectation keeps the canonical `local:a` ref (both sides pin it now) with this component's comment; the rail's archive call is `setArchived("session", session.ref, ...)`.
6. **The rail's delete item follows `main`'s semantics.** `main` landed the local-only refusal in the Rail itself: `onDeleteProjectRequest` refuses a project a remote host also owns, with a toast that names the hosts, and `main`'s `Rail.test.tsx` pins that. The row therefore no longer withholds "Delete project…": the branch's `ownedByRemoteHost` verdict and its menu gate are dropped, and the four row tests that pinned omission are re-pinned to the offered-item behavior (the person gets an explanation instead of an item that is silently missing). This was the one real behavior collision of the rebase - the first full `make test-web` run caught it - and it is `main`'s landed design that stands.

Two integration commits on top carry the leftovers of the replay: `style(hub): normalize the spawn pane's imports and resolve dependencies after the rebase` and `fix(hub): offer the rail's delete item on every project row`.

**Gates on the rebased tree**: `npx biome check --write` on the 14 touched `src/` files - clean; `npx tsc --noEmit` - exit 0; `make test-web` - PASS (web-typecheck, web-test, web-lint); `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` - PASS; `make lint-generated` - PASS. (Frontend-only change: no Go gates apply.)

obra:

**CI** — all required checks green at `d1cd7238b` (the full set, including RoboRev's own `roborev` check).

**Review** — RoboRev's verdict for `d1cd7238b` (the rebased head): **one finding, Low; "The one issue I found is a factual error in a new comment (and the change it justifies is a no-op), not a runtime defect."** Under the campaign's stated rule (a Lows-only verdict lands, with the Lows recorded as an immediate follow-up), it is recorded here:

- `cmd/evener-hub/frontend/src/shell/rail/Rail.tsx:843-866` (and the comment at `1350-1353`) — a new comment asserts that "the server stores whatever it is handed (app_archive.go's `archiveSet`) and no reader expands `local:<id>`", and builds a finding rationale on it. That is not what the server does: `archiveSet` normalizes the identity through `hubcore.NormalizeDecisionSessionID` (`cmd/evener-hub/app_archive.go:70`), which collapses `local:<id>` to the bare thread ID and keeps a host-qualified ref, and every read path consults the same key for both spellings. Both spellings store and read `{kind:"session", id:"a"}`, so the change the comment justifies is behaviorally a no-op. Fix: correct the comment to state the server's normalization (the spelling itself can stay or revert - the two are equivalent). The hazard is a future maintainer removing or bypassing that normalization on the comment's premise.

The rebase onto `main` and its reconciliation decisions (package import paths, the remote-launch exemptions layered on `main`'s onboarding-scoped gates, and the delete-item semantics change) are documented in the rebase-note comment above.

**Up-to-date override** — the branch is 3 commits behind `main`. Overlap check, run immediately before this comment:

```
$ gh pr view 1302 --json mergeable,mergeStateStatus --jq '.mergeable, .mergeStateStatus'
MERGEABLE
BEHIND

$ comm -12 <(gh pr view 1302 --json files --jq '.files[].path' | sort) \
           <(git diff --name-only "$(git merge-base origin/main origin/multi-host-pr06b-fleet-view-ui)" origin/main | sort)

$ git diff --name-only "$(git merge-base origin/main origin/multi-host-pr06b-fleet-view-ui)" origin/main | wc -l
8
```

`main` changed 8 files since the merge base; none of them is a file this PR touches.

Issue #1308, opened 2026-09-14
Flaky: TestSkillComposerBrowser (web-skillguard) times out on the CI runner after the select-all fix; passes on the same tree on main


obra:
On PR #1307 @6fa81bdc8 (test-only changes under `agent/plugin` and `agent/skill`, nothing under `cmd/evener-hub` or the frontend), CI's `web` job failed:

```
FAIL  web-skillguard (exit 1)
--- FAIL: TestSkillComposerBrowser (46.76s)
```

Main's `web` job passed on bcb8406d9, the tree this PR is based on, minutes earlier. The failure is at ~47s, well past the draft-clear step #1295 fixed, so it is one of the guard's later 15s waits timing out under runner load. The same guard showed load-dependent timeouts at varying later steps when run locally on a busy host during the #1292 work (`waiting for composer cleared`, `waiting for steered turn's input visible in the transcript`).

The driver log and screenshots go to a temp directory on the runner and are not in the job output, so the failing step cannot be read from CI. Suggested: upload the guard's artifact directory as a workflow artifact on failure, then decide whether the waits need to be event-driven rather than 15s timeouts.

obra:
Second occurrence within the hour: PR #1289 @ada3b3c4a (Go-only, cmd/evener-dev + internal/devtool), run 34832939534, `TestSkillComposerBrowser` failed at 30.15s; the driver log is again only on the runner. Two of the last five `web` runs on PR heads have hit this. Dispatching a CI change to upload the guard's artifact directory on failure so the failing step can be read.

obra:
#1309 is on main as f7816dd8d: the `web` job now uploads `/tmp/skillguard-browser-*/artifacts` as a workflow artifact on failure, and the Go test prints the last 40 lines of `driver.log` in its failure output after waiting (bounded) for the driver to exit. The next occurrence will say which step timed out.

obra:
Third occurrence, now with the step thanks to #1309: PR #1269 @4cdf6fcc4 (Go-only), run 34844761066, `TestSkillComposerBrowser` at 31.85s, driver.log's last line: `skillguard FAIL: timed out after 15000ms waiting for steered turn's input visible in the transcript`. The same step timed out locally under load during the #1292 work. The wait at that step (the steered turn's input appearing in the transcript after a steer submit) is the one to make event-driven or to give more than 15s under runner load.

obra:
[...]

PR #1309, opened 2026-09-14
ci(web): keep the skill guard's driver log and screenshots when it fails (#1308)

obra:
Refs #1308.

`TestSkillComposerBrowser` (`web-skillguard`) failed twice in an hour on Go-only PR heads (#1307, #1289) while main's web job passed. It keeps its driver log, screenshots, helper logs and request dumps under `/tmp/skillguard-browser-*/artifacts/` on failure — deliberately, so a failing run can be read — but on a runner nobody reads it: the directory goes with the machine, and the failure reaches the job output as one line naming a path that no longer exists.

Two copies now outlive the job.

**Uploaded.** A new `if: failure()` step in the `web` job uploads `/tmp/skillguard-browser-*/artifacts/**` with `actions/upload-artifact@v7` — the spelling `binaries.yml` uses and the major the rest of `ci.yml` pins its actions to — with 7-day retention and a name carrying `run_id`/`run_attempt` so two failures on one PR do not collide. The step also fires when `make test-web` or `make build-web` is what failed, and then there is nothing to collect, so `if-no-files-found: ignore` keeps that quiet instead of adding a warning that reads like a second failure.

**Printed.** The test logs the last 40 lines of `driver.log` from its failure cleanup. `scripts/web/test-web-browser.sh` already cats `skillguard.log` (the Go test output) on failure, so the driver's own last words — what it was waiting for, what the page showed — reach the job log without anyone downloading an artifact. That log is the guard script's scratch file, removed by its exit trap, so it is not uploadable; the tail is how its content survives.

**No change to the guard's waits or timeouts.**

## What was proved locally, and what was not

- `actionlint .github/workflows/ci.yml` — clean; the workflow also parses and the `web` job has the expected four steps.
- `bash -n scripts/web/test-web-browser.sh`; `go vet -tags browserguard ./cmd/evener-hub`; `gofmt` clean.
- The tail's line arithmetic is a pure function, `skillGuardDriverTail`, with seven cases: empty, newlines only, shorter than the limit, exactly the limit, longer (keeps the **end**), no trailing newline, and blank lines inside. It is the only thing a reader gets on CI, and an off-by-one would cost exactly the line that matters.
- That test rides along in the guard's `go test -run` pattern, since `browserguard` is the only build that compiles the file. `go test -tags browserguard ./cmd/evener-hub -list 'SkillGuard|SkillComposer'` lists both.
- `make lint` (nine targets) and `WEB=0 make test` (every module, both audits) pass.

**Not proved locally: the upload itself.** It runs only on failure, and failing the guard on purpose means a real Chrome, a built frontend and two helper daemons — minutes of setup to watch an `upload-artifact` step run. CI proves it the next time the guard fails, which on current evidence will not be long.


obra:
## Review round 1 disposition (4f520f4c3)

CI green; three of four reviewers clean. One Low, accepted: on the driver-timeout path the cleanup tails `driver.log` right after cancelling, while the driver may still be writing. The cleanup waits, bounded, for the driver to exit before reading, so the tail and the uploaded artifact carry its last words.

obra:
## Round 1 fixes

`9d1a3badf`. `origin/main` re-merged: already up to date.

**The finding is right, and the mechanism is `t.Context()`.** The driver runs under `exec.CommandContext(t.Context(), …)`, and Go cancels that context *before* cleanups run — so at cleanup time SIGTERM has been sent but `cmd.Run()` may still be waiting (`cmd.WaitDelay` is 15s) with `driver.log` still open. Reading then truncates the log at whatever had been flushed, losing exactly the lines that say what the driver was waiting for. That is the whole reason the tail and the upload exist.

The driver goroutine now closes a channel when it returns, and the failure cleanup waits on it, **bounded at 5s** — the shorter wait for an orderly exit, not a second timeout on the driver, since the kill is already in flight. The paths that already collected the driver's exit pay nothing: a closed channel returns at once, and a nil one (driver never started) returns at once too.

**Unit-tested, yes.** The wait takes a channel and a duration and needs no driver, so a fake future is the whole fixture: closed (returns immediately, asserted under 5s), nil (same), never-closing (returns false at a 10ms bound), and closed *during* the wait (picked up rather than timing out). The tail test stays pure and unchanged.

The guard's `-run` pattern is now `'^TestSkillComposerBrowser$|^TestSkillGuard'` so both unit tests ride along in the web lane; `go test -tags browserguard -list` shows all three.

### Gates on `9d1a3badf`

`go vet -tags browserguard ./cmd/evener-hub` · `gofmt` clean · `make lint` (nine targets) · both unit tests pass · `actionlint .github/workflows/ci.yml` clean · `bash -n scripts/web/test-web-browser.sh`.

PR #1311, opened 2026-09-14
fix(web): the skill guard types a draft through the DOM, not one key event per character

obra:
Refs #1308.

## What the artifacts said

Run 34844761066's failure dump (`skillguard FAIL: timed out after 15000ms waiting for steered turn's input visible in the transcript`) shows the transcript carrying

```
You
15s ago
ROSE_STEER_TURN_14e open a long turn for steeringP
```

against `PROSE.steerTurn = "PROSE_STEER_TURN_14e open a long turn for steering"`. **The leading `P` is at the end.** The turn arrived at once — it is 15s old at the dump — so this was never latency, and no timeout would have helped: the wait was for a string that could not appear.

`typeText` dispatched one CDP key event per character. The composer's textarea is React-controlled, so every keystroke's `onChange` re-renders and restores the value from state; a re-render landing between two key events reorders them. Same family as #1295, which replaced a Ctrl+A key event with a DOM selection for the same reason.

## The change

`typeText` inserts at the composer's selection through the DOM, in one edit, assigning through the native value setter so React's value tracker notices and fires `onChange` — exactly how `selectAll` already drives the selection.

**The per-character path is deleted, not kept.** Nothing in the guard needed it: the inline slash menu opens off the composer's value and caret rather than off keystrokes (`selectSkillChip` types ` /pkg:probe` and then waits for the menu, which still opens), and the semantic keys — Enter, Tab, Escape, Backspace — go through `press()`, which is untouched.

`clickSubmit(ref, expectText)` now takes the draft the scenario means to send and holds the composer to it before clicking, at all twelve call sites. A corruption fails at the submit that would have carried it, naming the characters, instead of surfacing as a wait expiring later somewhere else.

## Proof

- The whole guard passes locally against a built frontend: `go test -tags browserguard ./cmd/evener-hub -run '^TestSkillComposerBrowser$'` → `ok 16.094s`, steered-turn scenario included.
- The assertion catches a deliberate corruption. Scrambling the steer prose the way CI scrambled it makes the guard fail with

  ```
  skillguard FAIL: clickSubmit(local:034OiKkpC7hTn8ldLsGtcw): composer holds
  "ROSE_STEER_TURN_14e open a long turn for steeringP",
  expected "PROSE_STEER_TURN_14e open a long turn for steering"
  ```

  and pass clean (`ok 12.569s`) once the scrambling is removed. An earlier attempt that scrambled *every* prose was caught even sooner, by the existing "selection changed the prose" check — also by name, also not a timeout.
- `go vet -tags browserguard ./cmd/evener-hub`, `gofmt`, `make lint`, `biome ci` on the driver: clean.

The capability-loss step (the third failing step, 30s) is a separate question and gets its own PR: the artifacts cannot say whether that event never arrives or arrives late, because the two intermediate handoffs emit no milestone.


obra:
## Review round 1 disposition (1f79eab09)

CI green. RoboRev: one Medium and one Low, both accepted — `clickSteer` gets the same expected-text barrier as `clickSubmit` (the steer path is the one that failed in CI, so it needs it most), and the edit result is null-guarded before its fields are read.

obra:
## Round 1 fixes

`360c79c59`. Merged `origin/main` first.

**1 — the steer barrier.** `clickSubmit` checked the draft it was about to send; `clickSteer`, which sends the same composer the same way, did not. Both go through one `sendComposerDraft` barrier now (no second copy of the check), and the steer scenario states `PROSE.steer` at its call site.

**The list, in full: submit and steer are the only actions that send a typed draft.** The queue strip's two controls do not — `"Steer queue now"` drains rows that were queued through `clickSubmit`, each already held to its own text when it was queued, and `"Edit message"` returns a row *to* the composer without sending anything. Nothing else clicks `composer-submit` or `composer-steer`.

Proved with teeth: scrambling `PROSE.steer` the way CI scrambled the steer-turn prose fails at the steer, by name — `skillguard FAIL: clickSteer(local:034OmLt1x4pa50FS39YiWR): composer holds …` — and the guard passes clean once the scrambling is removed.

**2 — the null guard.** `typeText` read `.value` off the evaluate result without checking it came back. A null — a navigation between the send and the reply, or a CDP error — reported a `TypeError` from that line instead of the failure that caused it. It now names the missing result.

### Gates on `360c79c59`

The guard once against a built frontend: `ok 14.082s` (and `ok 16.440s` after the corruption probe was reverted). `make lint` (nine targets) · `biome ci` on the driver.

obra:
## Review round 2 disposition (360c79c59)

CI green; three of four reviewers clean. One Medium, accepted: the queue drain also sends whatever the composer holds, so it gets the same draft check as submit and steer (asserting the composer is empty before the drain, since that is the scenario's expectation). That makes three draft-sending entries, not two, and the barrier covers all of them.

obra:
## Round 2 fixes

`e84322b88`. Merged `origin/main` `ac7e5e64c` first.

**The reviewer is right and my round-1 list was wrong.** I called the queue strip's controls draft-free. `turn/drainAsSteer` "atomically appends the composer's current text/attachments (if any) to the input queue, then drains the whole queue into the active turn as one steering message" (`stores/threads.ts:157`, and it takes `text`, `attachments`, `skillNames`). So a stray draft rides along with rows the scenario meant to send alone, and the queue assertions that follow would be measuring a steering message the scenario never wrote.

**Three senders, not two.** The equality check is split out of `sendComposerDraft` into `assertComposerDraft`, so submit, steer and the drain share one rule, and the drain states what it expects — empty, in this scenario — before the click. `"Edit message"` remains the only queue-strip control that sends nothing: it returns a row *to* the composer.

Proved by seeding a stray draft immediately before the drain:

```
skillguard FAIL: drainAsSteer(local:034Omlfjlvo2xmVLbrEaRs): composer holds
"STRAY_DRAFT_that_would_ride_along_with_the_drain", expected ""
```

and the guard passes clean once the seed is removed.

### Gates on `e84322b88`

The guard against a built frontend: `ok 15.767s` with the change, `ok 12.661s` after the probe was reverted. `make lint` (nine targets) · `biome ci` on the driver.

obra:
## Review round 3 disposition (e84322b88)

CI green; three of four reviewers clean. Both findings accepted:
- **Medium, the draft assertion checks text only while submit and drain also send chips and attachments** — the assertion covers the whole composer payload (text, chips, tiles), so an empty draft means empty.
- **Low, the caret offset after an insertion uses the JSON-encoded literal's length** — a real off-by-quotes; the page-side length of the inserted text is used.

obra:
## Round 3 fixes

`f6605234f`. `origin/main` re-merged: already up to date.

**1 — the whole payload.** A send carries the composer's text, its skill chips and its staged attachments; only the text was checked, so an unexpected chip or a stray tile was as invisible as a scrambled character used to be. `assertComposerDraft` takes a `{text, chips, tiles}` shape and **all fourteen sends state theirs** — captured from an instrumented run rather than guessed, so each is what that send really carries (three go out with no chip, one with a tile). `placeholder`, `submitDisabled` and `steerVisible` are deliberately not compared: they describe the control, not what it sends.

A stray chip with empty text before the drain now fails by name:

```
skillguard FAIL: drainAsSteer(local:034OnIZWx8WSarDyUM8CNt): composer holds
{"text":"","chips":["pkg:probe×"],"tiles":0}, expected {"text":"","chips":[],"tiles":0}
```

**2 — the caret finding does not reproduce, measured two ways.** `${literal}` interpolates into *page source*, so `${literal}.length` is evaluated on the string literal, not on its JSON encoding:

```
"plain"          page caret offset=5  text.length=5  OK
"with \"quote\"" page caret offset=12 text.length=12 OK
"back\\slash"    page caret offset=11 text.length=11 OK
"new\nline"      page caret offset=8  text.length=8  OK
"emoji 😀"       page caret offset=8  text.length=8  OK
```

`literal.length` in Node *is* 7 for `"plain"` — but that is not what the code evaluates. And the reviewer's own case, run in the real browser: inserting `-mid"quo\te\n-` into the middle of `HEAD|TAIL` gives `HEAD-mid"quo\te\n-TAIL` with the caret at **17 = 4 + 13**.

Nothing was broken — but nothing said so either, and re-deriving that interpolation argument is exactly the reasoning a reader should not have to repeat. The text is bound to a page-side `typed` once, every offset is measured from it, and the caret is now checked after **every** insert. Breaking the offset by one fails with `typeText: caret at 18 after the edit, expected 17`.

### Gates on `f6605234f`

The guard against a built frontend: `ok 13.779s` with the change, `ok 12.947s` after every probe was reverted. `make lint` (nine targets) · `biome ci` on the driver.

obra:
Correction to the round 3 disposition: I called the caret finding a real off-by-quotes without measuring it. The implementer measured it in the real browser: the literal is interpolated into page source, so its `.length` evaluates on the page-side string (17 = 4 + 13 for the reviewer's own case), and the finding does not reproduce. The hardening (one page-side binding for the typed text, caret checked after every insert) is kept because it makes the argument unnecessary to repeat.

PR #1313, opened 2026-09-14
fix(llm): bill unsigned thinking text for adapters that replay it

obra:
# What

Follow-up to the roborev review of #1249 (medium: text-only thinking content counted as zero), revised in response to this PR's own review round.

**The rule.** A `ContentThinking` part that carries text and no replay metadata is billed only for the adapters that replay it unsigned:

- the Anthropic Messages adapter emits the block it is given as `{"type":"thinking","thinking":text}` (`llm/providers/anthropic/request.go`);
- the OpenAI-compatible chat adapter replays the text on the reasoning field unless the row is declared non-reasoning, and merges it into assistant content when the row sets `ThinkingAsText` (`llm/providers/chatcompletions/messages.go`);
- the OpenAI Responses adapter re-sends a reasoning item only with `encrypted_content` and keeps raw `reasoning_text` for display only; Google drops the part.

**Where the decision comes from.** The resolved registry row wherever a caller has one: `EstimateInputTokensForResolved` and `EstimateMessagesInputTokensForResolved` read `res.Protocol` and the row's thinking capabilities, and decide the media family from the protocol too. `ApplyTokenBudget`, `Client.CountInputTokens`, the agent's context-usage warning and the context manager's history/pressure accounting all use them.

The provider/model name rule survives only for callers with no resolved row (the bare `EstimateInputTokens`/`EstimateMessagesInputTokens` entry points and `CountInputTokens`' unresolved-override path). Its comment names what it misses — curated OpenAI-compatible chat providers whose names carry no marker (ollama, groq, zai, moonshotai, non-Claude openrouter rows) and instance aliases — and a Claude model name is only the last resort when the provider selects nothing.

# Evidence

Red first, each reproduced by temporarily restoring the pre-fix behavior and then restoring the fix:

```
--- FAIL: TestApplyTokenBudget_ResolvedOpenAIChatProviderBillsUnsignedThinking
    thinking text added 0 budget tokens, want 100: a resolved openai-chat row replays unsigned thinking text
--- FAIL: TestApplyTokenBudget_NonReasoningOpenAIChatRowDoesNotBillUnsignedThinking
    thinking text added 100 budget tokens, want 0: the row drops the part
--- FAIL: TestEstimateInputTokens_ProviderNameDominatesTheClaudeModelFallback
    estimate = 100, want 0: the google provider drops the part regardless of the model name
--- FAIL: TestEstimateTokens_BillsUnsignedThinkingByResolvedProtocol          (agent/internal/contextmgr)
    anthropic history = 0, want 100: the adapter replays the unsigned thinking block
--- FAIL: TestEstimateMessagesInputTokensForResolved_UsesTheResolvedMediaFamily
    resolved anthropic image history = 258, want 1369: the row's media family decides
```

`go test ./llm/... -count=1` and `cd agent && go test ./... -short -count=1` are green; so are all pre-existing tests, unchanged.

# Known limitation (follow-up)

Context-manager accounting estimates the stored turns, while the outgoing request is the *projected* history: `expandHistory`'s replay-provenance rules strip thinking a target cannot accept (for Anthropic, an exact instance + requested-model match; for Google and OpenAI Responses, the same instance). Where a stored turn's recorded provenance differs from the target's, the accounting therefore bills a part the request will not carry — an over-estimate that can only compact earlier, never admit an oversized request. A faithful fix needs the session's `replayScope` (including the `protocolOf` resolution for turns predating `ResponseProtocol` and the in-flight exemption), so it belongs with the manager's API rather than in the estimator; recorded here as the follow-up rather than approximated.


obra:
Round-1 answer, on head `33382e51e` (the rework `488a3ceb3` plus a clean merge of `main`).

**High — name-based matching misses real `openai-chat` adapters and all instance aliases.** Fixed, using the resolved row you pointed at:

- `llm` gains `EstimateInputTokensForResolved(res, req)` and `EstimateMessagesInputTokensForResolved(res, messages)`, which decide replay from `res.Protocol`. For `openai-chat` the rule mirrors the chat adapter: `registry.BoolValue(res.Caps.ThinkingAsText) || !res.Caps.ReasoningDisabled()`.
- `ApplyTokenBudget` and `Client.CountInputTokens` now call the resolved form whenever `dispatchTarget` resolved a row.
- The name rule survives only as the documented fallback for callers with no resolved row (the exported bare-request entry points and `CountInputTokens`' unresolved-override path). Its comment names exactly what it misses: curated OpenAI-compatible chat providers whose names carry no marker (`ollama`, `groq`, `zai`, `moonshotai`, non-Claude `openrouter` rows) and instance aliases that hide their base.

**Medium — model-name `claude` fallback overcounts.** Fixed: the provider name is consulted first, so `provider="google", model="claude-…"` stays at zero; the model name is only the last resort when the provider selects nothing.

**Medium — compaction/history path not fixed.** Fixed rather than documented. `contextmgr`'s `estimateTokens` and `estimateUsedTokens` — and the compaction strategies' before/after reports — now run on `llm.EstimateMessagesInputTokensForResolved(cm.currentProfile().Resolved(), …)`, so pressure and the reported numbers bill what the adapter replays. `maybeWarnContextUsage` uses its profile for the same reason. The two Responses-continuation shadow estimates in `agent/session_model_call.go` remain on the name fallback: they only seed `req.InputTokensEstimate`, which `ApplyTokenBudget` maxes with the exact resolved count, so they cannot under-admit — recorded as a follow-up rather than widened into this PR.

**Medium — over-billing non-reasoning `openai-compatible` rows.** Fixed for resolved rows by the `ThinkingAsText`/`ReasoningDisabled` rule above (a declared non-reasoning row without `ThinkingAsText` drops the part, like the adapter). The name fallback cannot see capabilities, so it still bills any `*compat*`/`openai-chat` name; its comment says so, and the resolved entry points are the answer for callers that know the row.

**Low — doc comment overstates coverage.** Fixed with the fallback's rewrite above.

**Red-first evidence**, each reproduced by temporarily restoring the pre-fix behavior and then restoring the fix:

```
--- FAIL: TestApplyTokenBudget_ResolvedOpenAIChatProviderBillsUnsignedThinking
    thinking text added 0 budget tokens, want 100: a resolved openai-chat row replays unsigned thinking text
--- FAIL: TestApplyTokenBudget_NonReasoningOpenAIChatRowDoesNotBillUnsignedThinking
    thinking text added 100 budget tokens, want 0: the row drops the part
--- FAIL: TestEstimateInputTokens_ProviderNameDominatesTheClaudeModelFallback
    estimate = 100, want 0: the google provider drops the part regardless of the model name
--- FAIL: TestEstimateTokens_BillsUnsignedThinkingByResolvedProtocol  (agent/internal/contextmgr)
    anthropic history = 0, want 100: the adapter replays the unsigned thinking block
```

Coverage added: `EstimateInputTokensForResolved` beating the name in both directions, the resolved history entry point billing an Anthropic row and sparing a Responses and a non-reasoning chat row, and the fallback's precedence. Existing llm and agent tests are green, unchanged.

**Gates on this head:** `make merge-approval-gate` exit 0 — all eight lint stages, root (479.1s), agent (40.2s), llm (13.1s), auth, envvars, invariant, identifier, web (866.4s), `test-native` (740 + 778 tests plus typecheck) and `test-api-package` — and `make vet` exit 0; the run log has no failure lines.

obra:
Round-2 answer, on head `a9baa4de7`.

**Medium — the resolved history entry point passed empty names, so images used the generic fallback.** Fixed in `13704a3a6`. The resolved target now carries the row's identity and protocol: the media family is decided by the protocol (`anthropic` → Anthropic tiles, `openai-chat`/`openai-responses` → OpenAI tiles, `google` → Google tiles), with the row's `Instance`/`ModelID` and then the name rule as fallbacks. Red first: `resolved anthropic image history = 258, want 1369` (generic placeholder before, Anthropic tiles now), then green; the targetless entry points keep the generic fallback, pinned by the same test.

**Medium — context accounting estimates turns whose thinking the replay-provenance rules strip.** Acknowledged, and deliberately not answered with a local filter. Three reasons:

1. The over-count is bounded: it only occurs for targets where `expandHistory` strips foreign thinking (anthropic exact instance/model, google and openai-responses same-instance) and only for stored turns whose recorded provenance differs from the target's; for every other target expansion keeps thinking, so the count is exact.
2. It errs toward *over*-estimating — earlier compaction, never an admitted oversized request, which was the direction the original finding cared about.
3. A faithful filter is not something the estimator can implement alone: eligibility depends on the session's `replayScope` (the `protocolOf` resolution for turns predating `ResponseProtocol`, the requested-model comparison, and the current-turn in-flight exemption). A near-miss reimplementation inside the context manager could strip parts that *are* sent — the under-counting direction — so I did not write one. The shape you named as the alternative (hand the manager the projected history, or the scope) is the right one, and it is recorded as the follow-up in the PR description.

**Gates:** `make merge-approval-gate` exit 0 on `13704a3a6` (the reviewed content) — all eight lint stages; root (440.2s), agent (50.3s), llm (18.9s), auth, envvars, invariant, identifier, web (1194.3s); `test-native` (740 + 778 plus typecheck); `test-api-package` — and `make vet` exit 0; the run log has no failure lines. This head adds only a clean merge of `main` (`dd517a79d`), and CI runs the full suite on it.

obra:
Round-3 answer, on head `147b249bb`.

**Medium — media estimates key on protocol instead of model family/surface.** Fixed in `147b249bb`, in the shape you suggested: `targetInfo` now carries the row's surface, and `mediaFamily()` returns anthropic/google/openai from a non-generic surface first, then the protocol (custom rows), then the name rule. Red first with your own case — an `openai-chat`-protocol row with `SurfaceAnthropic` and a `claude` model: `openrouter claude image history = 765, want 1369` (OpenAI formula before, Anthropic tiles now), green after. The same test pins a Gemini-over-openai-chat row (Google tiles) and the no-surface fallback (protocol axis).

**Medium — compaction pressure estimates raw turns instead of replay-eligible messages.** Adjudicated with the repository owner: this PR lands as it is, with the limitation recorded in the description, and the projected-history accounting becomes its own follow-up PR. The reasoning from the round-2 answer still holds — the over-count is bounded to targets that strip foreign thinking, errs toward earlier compaction rather than an admitted oversized request, and the faithful fix reuses the session's `replayScope` rather than reimplementing its predicate in the context manager.

**Gates:** `make merge-approval-gate` exit 0 — all eight lint stages; root (523.0s), agent (43.3s), llm (14.2s), auth, envvars, invariant, identifier, web (1087.1s); `test-native` (745 + 778 plus typecheck); `test-api-package` — and `make vet` exit 0; the run log has no failure lines.

obra:
Round-4 answer, on head `aa31e586e`.

**Medium — media estimates key on the surface/protocol and never on the resolved model family.** Fixed in `91a6faae9`: `targetInfo` now carries `res.Model.Family`, and `mediaFamily()` consults the surface, then that family, then the protocol. A generic-surface gateway row (the OpenRouter shape) with a `claude-*`, `gemini-*` or `gpt-*` family now bills its own family's image formula instead of the wire protocol's, while a family with no image-token rules of its own (`llama`, `minimax`, `kimi`, `deepseek`) keeps the protocol fallback and the name rule still covers callers that passed names only.

Red first, with your own case:

```
--- FAIL: TestEstimateMessagesInputTokensForResolved_ModelFamilyBeatsTheWireProtocol (0.03s)
    token_count_test.go:603: generic-surface claude image history = 765, want 1369: the recorded family decides
```

Green after; the tests also pin the family-outranks-protocol disagreement and the unknown-family fallthrough.

**Medium — context-window size and token estimation read `currentProfile()` independently.** Fixed in `51ed37d0e`: `estimatePressure` and `EstimateUsage` take one `currentProfile()` snapshot and thread it through the new `estimateUsedTokensFor` / `estimateTokensFor`, so a `SetProfile` (a model switch) between the window read and the token estimate cannot pair one model's window with another model's accounting. The existing `estimateUsedTokens` / `estimateTokens` keep their signatures and take a fresh snapshot for callers that hold none.

Red first, under a concurrent profile swap:

```
--- FAIL: TestContextManager_UsageEstimatePairsOneProfilesWindow (0.21s)
    context_manager_test.go:1747: EstimateUsage = {Used:0 Window:200000 Remaining:200000}, want {Used:0 Window:1000000 Remaining:1000000} or {Used:55500 Window:200000 Remaining:144500}: the window and the estimate must come from one profile snapshot
```

That result is exactly the blend you described — one model's window carrying the other's accounting. The test loops `EstimateUsage` and `Pressure` against a goroutine flipping `SetProfile` and accepts only the two single-profile outcomes; it passes deterministically on the fixed code.

`aa31e586e` is the two fixes plus a test-only follow-up (`test(contextmgr): use WaitGroup.Go in the profile-snapshot test`) after the repository lint caught the older goroutine idiom.

**Gates:** `make merge-approval-gate` exit 0 — all nine lint stages (`lint-naming`, `lint-gofmt`, `lint-evenerfuzz`, `lint-eval`, `lint-internal`, `lint-golangci`, `lint-generated`, `lint-fuzz-registry`, `secret-scan`), the root/agent/llm/auth/envvars/invariant/identifier module runs (root 669.5s, agent 67.3s, llm 23.7s), the web build + typecheck (1798.1s; the host was running several other gates), `test-native` (vitest + `tsc --noEmit`) and the appwire package qualification — plus `make vet` exit 0. The run log has no failure lines.

obra:
Round-5 answer, on head `13ea30ed5`.

**Medium — the fallback warning used the primary profile.** Fixed in `13ea30ed5`: `callModelWithFallback` now returns the profile that actually handled the round, and the lifecycle passes that to `maybeWarnContextUsage`. Red first, on a round whose primary fails 403 into a fallback whose window is near while the primary's is not:

```
--- FAIL: TestSessionFallbackContextWarningNamesTheWindowThatHandledTheRequest (0.87s)
    final_fix_warning_test.go:579: no context-usage warning: the fallback's window is near, not the primary's: []
```

Green after: the warning reports the fallback's 120000-token window for the round the fallback served. The 27 test call sites gained the extra ignored result mechanically.

One honest note on the estimate half: `maybeWarnContextUsage` still counts tokens with `llm.EstimateInputTokens` — the name-based entry point, which this series did not change. What the fix guarantees is that the window and the request describe the same model; the resolved-row estimator remains the context manager's accounting path.

**Medium — split-lock atomicity between the profile and the measurement.** Fixed in `4f9b5a664`: `estimatePressure` and `EstimateUsage` now take `profileSnapshot()` — the profile and the API measurement in one `cm.mu` section — and thread that snapshot through, so the pair cannot be read apart. Honest limitation on the red evidence: the torn pair is not dynamically reproducible in a bounded test. I measured it under saturation (four goroutines flipping `SetProfile`, 2000 calls, thousands of swaps per call): every reading landed on one of the three single-snapshot states, zero torn. The reads sit nanoseconds apart and the mutex's barging keeps the consecutive acquisitions together, so there is no red to quote; the fix makes the invariant structural rather than test-visible. The test you noted only covered the zero-measurement case — `TestContextManager_UsageEstimateKeepsProfileAndMeasurementTogether` now covers the measurement path and pins the three-outcome contract.

**Medium — media-family precedence skipped the model name, and gpt-oss.** Fixed in `643548c2a`. The name rule now runs ahead of the protocol switch, so your failing row — `registry.Resolved{Protocol: ProtocolOpenAIChat, Surface: SurfaceGeneric, ModelID: "anthropic/claude-opus-5"}`, no recorded family — bills Anthropic:

```
--- FAIL: TestEstimateMessagesInputTokensForResolved_NameIdentifiesTheFamilyWhenTheRowDoesNot (0.03s)
    token_count_test.go:669: name-only claude image history = 765, want 1369: the model name identifies the family
```

Green after, with the family map now mirroring the registry's own family → surface rule (§6.1): `claude*` and `gemini*`/`gemma*` keep their vendors' rules, `gpt*` and `o`/`o-mini`/`o-pro` are OpenAI's, and `gpt-oss` is deliberately generic — covered by new cases for gemma, the o family, and a gpt-oss row.

On your gpt-oss consequence: I could not find an image-token estimate for a gpt-oss row to be wrong, because there are no gpt-oss images. The resolved golden row (`llm/registry/testdata/golden/groq-chat.json`) carries `"input_modalities": ["text"]` — the model takes no image input at all, so no media estimate is ever billed for it. The family map no longer claims OpenAI for gpt-oss anyway (the callee you flagged is gone), but note the name rule is unchanged and pre-series: a caller holding only a `gpt-`/`o`-prefixed name still maps to OpenAI, which is the behavior `providerTokenFamily` has always had.

**Gates:** `make merge-approval-gate` exit 0 on `13ea30ed5` — all nine lint stages (`lint-naming`, `lint-gofmt`, `lint-evenerfuzz` 167s, `lint-eval` 85s, `lint-internal`, `lint-golangci` 132s, `lint-generated`, `lint-fuzz-registry`, `secret-scan`), the root/agent/llm/auth/envvars/invariant/identifier module runs (root 460.3s, agent 42.3s, llm 15.3s), the web build and typecheck (1042.9s), `test-native` and the appwire package qualification — with no failure lines — plus `make vet` exit 0.

obra:
**Round 6 — all three findings addressed** in `09032458f`.

### Medium — continuation estimates used the name-based estimator

`agent/session_model_call.go`: the shadow estimate, the full-history plan, and both fallback rebuilds now take the resolved row and estimate with `llm.EstimateInputTokensForResolved`:

- the primary planning path resolves the session's own row once (`resolvedRequestTarget`) and threads it through every return;
- a fallback rebuild uses the *fallback's* row (`fbProfile.Resolved()`), and the anchor-rejection retry uses the attempt's profile;
- the shadow estimator itself (`estimateResponsesContinuationShadow`) resolves the session row.

Red first, on a gateway row (`cerebras`-style instance, model `gateway-zz`, resolved `openai-chat` with `thinking_as_text`, so the adapter replays unsigned thinking and the name rule cannot see it):

```
--- FAIL: TestSessionContinuationShadowEstimateUsesTheResolvedTarget (1.11s)
    task4_token_budget_test.go:906: InputTokensEstimate = 4, want the resolved-target estimate 244 (the name-based rule would bill 4)
```

The test keeps a control assertion that the two rules disagree for that row, so it cannot pass vacuously, and a companion test pins the fallback rebuild to the row it was handed.

### Low — the gpt-oss carve-out was dead for real model ids

`llm/token_count.go`: `providerTokenFamily` now refuses every gpt-oss spelling before the `gpt-` / leading-`o` branches. The registry's family map already declined the family (`mediaFamilyFromModelFamily`), but that refusal was invisible to the name stage, so a real row (`gpt-oss-120b`, `openai/gpt-oss-120b`, `openai.gpt-oss-120b`) was re-claimed for OpenAI. The rule now lives at its source, so the name-only entry points cannot claim it either.

Red first, with your exact numbers:

```
--- FAIL: TestEstimateMessagesInputTokensForResolved_NameIdentifiesTheFamilyWhenTheRowDoesNot (0.02s)
    token_count_test.go:697: gpt-oss row "gpt-oss-120b" image history = 765, want 1369: the gpt-oss family must not claim OpenAI
```

The coverage now uses real `ModelID`s (with an anthropic protocol, so the protocol decides) and asserts `providerTokenFamily("cerebras", "gpt-oss-120b") == ""`. A name-only gpt-oss call now takes the flat unknown-family fallback instead of OpenAI's formula — the honest answer for a family the registry deliberately leaves generic.

### Low — nil-profile panic in the contextmgr readers

`agent/internal/contextmgr/context_manager.go`: `estimatePressure` and `EstimateUsage` read the window through `contextWindowOf(prof)`, and `estimateTokensFor` reads the row through `resolvedTargetOf(prof)`. No profile now means no window known (zero metrics, zero pressure) and the estimator falls back to the names the turns carry, which is what these paths did before the profile-aware accounting.

Red first — the panic you predicted, caught by a test that constructs `NewManager(nil, nil, cheapmodel.New(nil))` (the session-log strategy suite does exactly that):

```
panic: runtime error: invalid memory address or nil pointer dereference
primeradiant.com/evener/agent/provider.(*Profile).ContextWindowSize(...)
	agent/provider/profile.go:266
primeradiant.com/evener/agent/internal/contextmgr.(*Manager).EstimateUsage(...)  context_manager.go:425
```

### Gates

Gates on `09032458f`: `make merge-approval-gate` exit 0 -- all nine lint stages (including the `lint-evenerfuzz` compile floor, which caught a missing `registry` import in the tagged round-contract fuzz file at `952daafaa`; that fix is `09032458f`), the module runs (root 618.1s, agent 63.8s, llm 19.2s, auth, envvars, invariant, identifier), the web build and typecheck (1788.9s), then `test-native` and the appwire package qualification. No failure lines. `make vet` exit 0, and the notes/steering/mutation/provenance families plus `go test ./llm/... ./agent/internal/contextmgr/...` are green.

obra:
**All three findings addressed** in `43dae1a92`.

**Medium — the unsynchronized profile read, and the pairing it allowed.** `resolvedRequestTarget` is gone. The planning pass now takes the profile its caller built the request with — both call sites already had it, right above `buildModelRequest(profile, ...)` — threads it through the shadow estimate and the full-history estimate, and `resolvedTargetOf(profile)` is a single nil-safe read of a profile the caller holds, so no unsynchronized `s.profile` read remains on this path (the fallback rebuilds keep passing their own attempt's profile).

```
--- FAIL: TestSessionContinuationShadowEstimateUsesTheHandedProfile (1.73s)
    task4_token_budget_test.go:945: InputTokensEstimate = 4, want the handed profile's estimate 244 (the name rule would bill 4, the live row 4)
```

The test gives the session a live row that does *not* replay unsigned thinking and hands the planning pass a row that does, with control assertions that all three rules disagree; it pins both the request reading and the full-history reading the shadow feeds. Reverting both reads to the live profile is what makes it red.

**Low — the partial row dropped the request's names.** `unsignedThinkingReplayed` now takes the effective provider and model (the row's own only where the caller supplied none), so a row with no protocol decides by the names the request carries instead of by two empty strings.

```
--- FAIL: TestEstimateInputTokensForResolvedPartialRowKeepsTheRequestNames (0.00s)
    token_count_test.go:668: partial-row estimate = 4, want the name-based estimate 244: the request names the vendor
```

**Low — one snapshot per compaction pass.** `MaybeCompact`, `ForceCompact`, and the three strategy `ManageContext` passes each take a single `profileSnapshot()` at the top and read every window and before/after diagnostic through it (28 sites). Honest note: this is correctness by construction — a single-threaded test cannot distinguish the shapes, because the readings can only disagree when a model switch lands between them; the suite's existing diagnostics assertions stay green.

Gates on `62f07b04b`: `make merge-approval-gate` exit 0 -- all nine lint stages (including the `lint-evenerfuzz` compile floor, which caught the two tagged `applyResponsesContinuationAnchorPlanning` call sites that plain `go test` never compiles), the module runs (root 636.5s, agent 71.0s, llm 14.6s, auth, envvars, invariant, identifier), the web build and typecheck (1230.2s), then `test-native` and the appwire package qualification. No failure lines. `make vet` exit 0.

obra:
**The medium is fixed** in `874a4485e` (merged as `6b2fe6ec2`).

Each compaction phase now reads pressure and its before/after diagnostics from one `(profile, measurement)` snapshot: `pressureFromSnapshot` is `estimatePressure`'s core for callers that already hold a snapshot, `MaybeCompact` and the three strategy `ManageContext` passes hand the profile that decided the layer to that layer's own diagnostics, and the re-read between layers still picks up the invalidated measurement — which is what makes between-layer pressure use char/4, as the existing design intends.

Honest note, the same one as the previous round's diagnostics change: this is correctness by construction. The readings can only disagree when a `SetProfile` switch lands between them, so a single-threaded test cannot distinguish the shapes. The existing diagnostics, pressure, and threshold assertions stay green, and the whole module suite passes.

Gates on `6b2fe6ec2`: `make merge-approval-gate` exit 0 -- all nine lint stages (including the `lint-evenerfuzz` compile floor), the module runs (root 405.9s, agent 42.8s, llm 12.5s, auth, envvars, invariant, identifier), the web build and typecheck (916.4s), then `test-native` and the appwire package qualification; no failure lines. `make vet` exit 0, and the full `agent/...` module in short mode plus the `contextmgr` suite pass.

obra:
**The low is fixed** in `cc949cd2e`.

An instance name says where a row was reached, not who made the model. `targetInfo` now records that it came from a registry row, and the media-family name rule reads a resolved target's model name alone (`vendorNameBasis`): the row's own surface and family decide first, then the model name, then the wire protocol — while callers that passed names only keep using the provider string, which is all they have. The case you named:

```
--- FAIL: TestEstimateMessagesInputTokensForResolved_InstanceNameIsNotVendorIdentity (0.02s)
    token_count_test.go:686: aliased row image history = 1369, want 765: the protocol decides, not the instance alias
```

The gateway shape from the earlier round is unchanged — `Instance: openrouter`, `ModelID: anthropic/claude-opus-5`, where the *model* name carries the vendor — and the whole module suite passes.

Gates on ff505dc56: GATE_LINE

obra:
**All three findings fixed** in `7c179aa6a`.

**Medium — the o-prefix over-match.** `oSeriesModelName` replaces `strings.HasPrefix(m, "o")`: it matches `o` alone, an `o-` prefix (`o-mini`, `o-pro`), `o` followed by a digit (`o1`, `o3`, `o4-mini`), and the `openai/` / `openai.` namespaced forms of those — so Mistral's `open-mistral-*`, OLMo, OpenChat and the gateway paths keep the protocol's own rules.

```
--- FAIL: TestEstimateMessagesInputTokensForResolved_OPrefixIsNotAnOpenAIClaim (0.01s)
    token_count_test.go:683: row "open-mistral-7b" image history = 765, want 1369: a leading o is not an OpenAI claim
```

**Medium — surface before family.** `mediaFamily` now reads the recorded model family first and the surface second. The provenance question is sidestepped by ordering rather than by tracking it: the family is the *model's* own classification and the surface only says where the row was reached, so an explicitly classified model wins either way — whether the surface came from the row or was filled in from the provider.

```
--- FAIL: TestEstimateMessagesInputTokensForResolved_ModelFamilyBeatsProviderSurface (0.02s)
    token_count_test.go:707: claude-family row over an openai surface = 765, want 1369: the family is the model's own
```

**Low — the caller's provider name.** `targetInfo` now records whether the provider came from the caller (`providerFromCaller`), and `vendorNameBasis` returns it in that case; only the row's instance alias is discarded.

```
--- FAIL: TestEstimateInputTokensForResolved_KeepsTheCallersProviderName (0.02s)
    token_count_test.go:724: resolved estimate with a caller-supplied provider = 258, want the name-based estimate 1032
```

Gates on `7c179aa6a`: `make merge-approval-gate` exit 0 -- all nine lint stages, the module runs (root 423.8s, agent 43.6s, llm 13.0s, auth, envvars, invariant, identifier), the web build and typecheck (877.1s), then `test-native` and the appwire package qualification; no failure lines. `make vet` exit 0 and the full `llm` module suite passes.

obra:
**All three mediums fixed** in `27fd9ea9c`.

**1 — the measurement outlived its rules.** `sameProfileTarget` compared instance, model, and protocol only, while the estimator also reads the row's surface, model family, and reasoning capabilities. `sameEstimatorTarget` compares those too, so a metadata refresh that changes any of them clears the measurement. Four new table cases cover it, and each fails when the comparison is reverted:

```
--- FAIL: TestContextManager_SetProfileInvalidatesMeasurementsOnlyWhenTargetChanges
    --- FAIL: .../surface  --- FAIL: .../model_family  --- FAIL: .../thinking_as_text  --- FAIL: .../reasoning
```

**2 — the instance alias passing as caller-supplied.** `CountInputTokens` wrote `req.Provider = t.name` before estimating, and `targetFromResolved` reads every non-empty provider as caller-supplied, so an instance aliased with a vendor-like name could claim that vendor's image rules. The local estimate now takes the caller's own provider — empty when the caller gave none — through `localInputEstimate`, never the name the dispatch happened to use.

```
--- FAIL: TestLocalInputEstimateKeepsTheCallersProviderNotTheInstance (0.02s)
    token_count_test.go:744: estimate with no caller provider = 1033, want the row-and-protocol estimate 766
```

**3 — generic is a decision, not an absence.** `mediaFamily` now short-circuits for a deliberately generic family (`genericModelFamily`) before the surface, name, and protocol stages, so a gpt-oss row never inherits a vendor's image formula. That supersedes round three's expectation that the protocol answers for gpt-oss, and the test records that where it is pinned.

```
--- FAIL: TestEstimateMessagesInputTokensForResolved_GenericFamilyClaimsNoImageFamily (0.02s)
    token_count_test.go:763: gpt-oss row over "openai-responses" image history = 765, want the generic fallback 258
--- FAIL: TestEstimateMessagesInputTokensForResolved_NameIdentifiesTheFamilyWhenTheRowDoesNot (0.04s)
    token_count_test.go:838: gpt-oss row "gpt-oss-120b" image history = 1369, want the generic fallback 258: the family must not claim any vendor
```

Gates on `27fd9ea9c`: `make merge-approval-gate` exit 0 -- all nine lint stages including the `lint-evenerfuzz` tag floor, the module runs (root 443.0s, agent 44.2s, llm 10.4s, auth, envvars, invariant, identifier), the web build and typecheck (836.1s), then `test-native` and the appwire package qualification; no failure lines. `make vet` exit 0.

obra:
**The medium is fixed** in `77396419f` (merged as `fef07988f`).

You are right, and it is my own comparison from the previous round: `sameEstimatorTarget` dereferenced `Caps.Reasoning` and compared the raw bool, while the estimator reads `Caps.ReasoningDisabled`. So `nil` (never stated; unsigned thinking is billed) and an explicit `false` (reasoning disabled; the adapter drops it) compared equal, and a measurement taken while the adapter dropped thinking survived a refresh to a row that bills it — under-reporting everything inside the measured prefix, which is exactly the under-billing class this series exists to fix. The other direction was wrong too: `nil` and an explicit `true` bill identically yet compared different, discarding a measurement that still held.

The comparison now uses `ReasoningDisabled()` on both sides, matching `unsignedThinkingReplayed` exactly. The test drives a silent row into an explicit `true` (the measurement stands) and then an explicit `false` (it is cleared), and it fails on the raw-bool comparison in the second half:

```
--- FAIL: TestContextManager_SetProfileInvalidatesMeasurementsOnlyWhenTargetChanges
    --- FAIL: .../reasoning_metadata_refresh
```

The earlier cases (surface, model family, thinking-as-text, and an explicit reasoning flip against a row that stated it) stay and still pass.

Gates on `fef07988f`: `make merge-approval-gate` exit 0 -- all nine lint stages including the `lint-evenerfuzz` tag floor, the module runs (root 388.8s, agent 38.7s, llm 10.0s, auth, envvars, invariant, identifier), the web build and typecheck (784.0s), then `test-native` and the appwire package qualification; no failure lines. `make vet` exit 0.

obra:
**The medium is fixed** in `fe8b9b065`.

`vendorNameBasis` blanks the provider whenever the resolved row has a surface, a model family, or a protocol to decide from — a name that merely repeats the row's instance alias never outranks the row's own facts — and keeps it only when the row has nothing else *and* the caller supplied it (the tenth round's case, whose test still passes). The model name still answers between the row's facts and the protocol, so the gateway shape from the fifth round is unchanged, and the two estimators agree on the aliased row now:

```
--- FAIL: TestEstimateInputTokensForResolved_InstanceAliasDoesNotOutrankTheRowProtocol (0.02s)
    token_count_test.go:743: request estimate = 1369, want the history estimate 765 for the same row: an alias is not vendor identity
```

The family-precedence tests from rounds 5, 9, and 10 all pass alongside the new one, which is the point of this fix.

Gates on `fe8b9b065`: `make merge-approval-gate` exit 0 -- all nine lint stages including the `lint-evenerfuzz` tag floor, the module runs (root 450.4s, agent 76.6s, llm 17.7s, auth, envvars, invariant, identifier), the web build and typecheck (1366.1s), then `test-native` and the appwire package qualification; no failure lines. `make vet` exit 0.

obra:
**Simplify pass applied** (four read-only reviewers, one per angle: reuse, simplification, efficiency, altitude; scope `git diff 498cde579...fe8b9b065`).

**Fixed** — all behavior-preserving, same tests and assertions.

- `pressureWithProfile` reads pressure and the profile it measured from ONE snapshot, replacing the closure pasted across `MaybeCompact` and the three strategy passes (and the three one-use `provider` imports they needed).
- The estimator owns its target predicate now: `llm.EstimatorTargetsEquivalent`, with `sameEstimatorTarget` delegating to it, so the field list and the nil-vs-false reasoning rule live where the rules do.
- One `gptOSSName` spelling replaces three copies in `token_count.go`, and the family map calls `genericModelFamily` rather than repeating the prefix check.
- Tests: one shared 1024×1024 image fixture replaces 11 per-call encodes of 4 MiB; one shared thinking-heavy history replaces two character-identical 3000-turn fixtures; the Anthropic-replay test calls the existing `thinkingOnlyRequest`.

**Skipped**, per the pass rules: exporting the registry's family→surface table (reaches outside the diff), deriving vendor identity from registry name lookup (a behavior change), and the `usedProfile` return value plus the unpaired estimate wrappers (already-reviewed interfaces whose fixes would change signatures the diff added). Nothing was reverted.

Gates on `80d59b3cb`: `make merge-approval-gate` exit 0 (all nine lint stages, root 463.3s, agent 66.6s, llm 16.5s, web 1038.2s, then `test-native` and the appwire qualification; no failure lines) and `make vet` exit 0.

obra:
**All four findings fixed** in `33874c028`.

**Medium — the measurement's owner.** `recordResponseUsage` now carries the profile that answered the round (`usedProfile`, which the lifecycle already had in hand) into `RecordInputTokensFor`, which refuses a measurement whose profile the manager no longer holds — so a `SetModel` landing while the request was in flight cannot attribute one model's count to another, and the count cannot survive into rules it was not taken under.

```
--- FAIL: TestRecordInputTokensForRejectsAStaleAnsweringProfile (0.13s)
    context_manager_test.go:1642: a measurement from a profile the manager no longer holds was accepted
```

**Low — `gemma`.** `providerTokenFamily` recognizes it as Google's now, exactly as the family map already did.

**Low — namespaced `gpt-` ids.** The namespace comes off once (`openai/`, `openai.`) before every branch, so `openai/gpt-4o` and `openai.o4` read like their bare spellings again; the o-series narrowing had left them unmatched on the name-only path.

**Low — the equivalence's model id.** `EstimatorTargetsEquivalent` compares `ModelID` now, which its doc claimed because the media-family name rule reads it. There was no live defect (`sameProfileTarget` also compared the model), but the API no longer asks its caller to re-list what it promises to cover.

```
--- FAIL: TestProviderTokenFamilyReadsNamespacedAndGemmaNames (0.00s)
    token_count_test.go:819: providerTokenFamily(gateway, "openai/gpt-4o") = "", want openai
--- FAIL: TestEstimatorTargetsEquivalentCoversTheModelID (0.00s)
    token_count_test.go:805: rows with different model ids are different estimator targets: the name rule reads the id
```

Gates on `33874c028`: `make merge-approval-gate` exit 0 -- all nine lint stages including the `lint-evenerfuzz` tag floor, the module runs (root 440.5s, agent 50.2s, llm 15.0s, auth, envvars, invariant, identifier), the web build and typecheck (1039.7s), then `test-native` and the appwire package qualification; no failure lines. `make vet` exit 0.

obra:
**Both findings fixed** in `1a75441e6`.

**Medium — the row's image detail.** `targetInfo` carries the row's configured `Caps.ImageDetail` now, and the image estimate uses it whenever the image itself carries none, so a row set to `image_detail = "low"` bills the low-detail formula the request will actually use instead of tiles. `EstimatorTargetsEquivalent` compares it like every other capability the estimate reads, so a row switch that only moves the detail re-measures. Red first by reverting the hunk in place: the detail-less image under the low-detail row billed 765 where the request's own detail bills 85.

**Low — an instance alias is not vendor identity.** `targetFromResolved` no longer hands `res.Instance` to the thinking rule's name fallback; the fallback reads the caller's provider only, and the row's model name remains its evidence. Red first: a row whose only fact is the instance name `anthropic` billed 15 thinking tokens with nothing vendoring Anthropic.

Gates green on `1a75441e6`: `make merge-approval-gate`, `make vet`, `make lint-evenerfuzz`.

obra:
**All three findings fixed** in `efe75f53d`.

**Medium — the shapes belong to the adapter, not to the family.** `thinkingReplayChars` now bills a resolved row that names its protocol for exactly what that protocol's builder emits: `anthropicThinkingChars` (the text plus the signature, blanked when the value is really a compat wire field name, and nothing at all when the part carries no text), `responsesReasoningChars` (the non-compat blob with its trimmed id and non-blank summaries, and nothing else — the display text beside a blob does not ride), `chatReasoningChars` (the compat `reasoning_details` array unless the row declares reasoning = false, and the part's text unless that declared row also replays thinking as text), and nothing for Google. A target that names no protocol keeps the previous conservative billing, with the text-only case still decided by `unsignedThinking`. Red first: 17 of the new per-protocol subtests failed — a Google redacted part billed 100 where its adapter drops it, a chat opaque blob billed 200 where only its text rides (100), and an Anthropic encrypted-only part billed 200 where nothing rides.

**Medium — the row's image detail is the Responses builder's.** Only `responses/request.go` injects `image_detail` (and a "high" default when the row sets none); `chatcompletions/messages.go` sends the image's own detail or nothing at all, so a chat row's cap never reached the wire and the estimate billed 85 where the request actually contains 765. The row default now applies to the Responses protocol alone, and a chat row estimates the same with or without the cap.

**Low — the duplicate is gone.** Both call sites read `registry.StringValue`, the same optional-string reader the adapters use, so the "nil and empty mean the same thing" convention lives in one place.

Gates green on `1fa33c05d`: `make merge-approval-gate`, `make vet`, `make lint-evenerfuzz`.

obra:
**Both findings fixed** in `56a35f84e` (with `d80ef9be5` completing the integration).

**Medium — a tool-result image is billed only where the adapter leaves it on the wire.** The chat builder strips a tool-result image unless the row declares `MultimodalToolResults` (`chatcompletions/messages.go`), so a cap-less chat row billed an image the shaped request no longer contains: 769 tokens for a result the builder sends as text (4). The Google builder refuses such a request outright (`google/request.go`), so nothing of the image rides there either. `targetInfo` now carries `dropsToolResultImages` from the resolved row; the Anthropic and Responses builders, which emit the image regardless of the cap, keep billing it, and a caller that passed names alone keeps the conservative billing since a name carries no cap to read.

**Medium — a fixed low-detail cost needs no dimensions.** The row's configured `image_detail` was read only after the image's dimensions decoded, so a remote URL (nothing to decode locally) fell back to the large placeholder (264) even under `image_detail = "low"`, where OpenAI bills a flat 85 without inspecting the image. `effectiveImageDetail` resolves the detail that rides the wire first, and the OpenAI low-detail cost is billed from it before the undecoded fallback. The contrasts stay pinned: no configured detail keeps the fallback, the image's own `low` bills 85 for both builders, and a chat row still ignores the row's cap.

The estimate reads a new capability, so `EstimatorTargetsEquivalent` compares it with the others the estimate reads — the context manager keeps a measurement only while its target matches, and a count taken under the old cap may not survive a row that disagrees.

Gates green on `d80ef9be5`: `make merge-approval-gate` (nine lint stages PASS; root 557.8s, agent 57.2s, llm 17.4s, web 1305.4s), `make vet` exit 0.

obra:
**Both findings fixed** in `f64f26668`.

**Medium — a chat row never puts a tool-result image on the wire.** Correct, and it makes my round-18 remedy half wrong. `MultimodalToolResults` on a chat row only prevents the image fields from being stripped (`requestWithoutToolResultImages`); `toChatMessages` emits `role`/`tool_call_id`/`content` (plus `name` when the row asks) and never `ImageData`, because Chat Completions has no portable image-bearing tool-message representation. `dropsToolResultImages` now answers per adapter: chat never carries one, Google carries one exactly when the row declares the cap (its builder emits the image inside the function response and refuses the request without it), and Anthropic and Responses carry theirs regardless. The test pins all four, both Google directions included. Red first: the declared-cap chat row billed 769 where the shaped request sends text only (4).

**Medium — the fallback request is priced for its own row.** The builder computed its estimate while the request still carried the primary's provider and model, and the caller then re-stamped the identity and re-priced the request before dispatch — so the dispatched value was the caller's, but the builder's own estimate read the primary's names (a Google media family for a row that carries no vendor facts). The builder now stamps `res`'s identity before it prices anything, and the caller dispatches that one request as built, so there is exactly one place that decides the identity and one estimate. Red first: the fallback request came back as `google/gemini-3-pro` with a 1032-token estimate instead of the fallback row's generic 258.

Gates green on `f64f26668`: `make merge-approval-gate` (nine lint stages PASS; root 431.2s, agent 50.3s, llm 13.3s, web 1021.0s), `make vet` exit 0.

PR #1314, opened 2026-09-14
perf(notes): skip escape probes for backslash-free journals

obra:
# What

Follow-up to the roborev review of #1305.

**Finding (roborev, low; reported by two reviewers).** The absence fast path in `agent/session_notes.go` runs up to 20 full `bytes.Contains` scans for every keyless snapshot, versus two previously; large backslash-free journals regress substantially. Suggested fix: guard the probe loop with a single-byte scan.

**Fix (exactly that).** `containsEscapedNotesKey` returns `false` after one `bytes.IndexByte(data, '\\')` when the document holds no backslash byte. Every probe is a `\u00XX` spelling, so a backslash-free document cannot contain any escape spelling and the probe loop is unreachable; a document that does contain a backslash takes exactly the old path (and pays one extra single-byte pass before the probes). The call site is unchanged.

# Why there is no red-first test

This guard is behavior-preserving by construction, so no behavioral test can fail before it and pass after; I am not dressing one up. The equivalence argument, verified against the code: probes are built as `\u%04x`/`\u%04X` spellings of every non-quote rune of `"human_note"` — 20 probes, each containing a backslash, while the key itself contains none. Backslash-free input matched no probe before and returns `false` before the loop now; backslash-bearing input runs the identical loop. No input decides differently.

# Evidence

- **Equivalence, independently re-verified:** a scratch property test (not committed) compared `containsEscapedNotesKey` against the unguarded probe loop over 3,300 deterministic random byte strings (lengths 0-80, an alphabet rich in backslashes, key characters and digits) plus targeted documents (empty; literal key; `"\u0068uman_note"`; escaped journal payloads; `\\u0068` inside a string; Windows-path backslashes). No mismatch. Focused reader tests green.
- **Performance, guard present vs absent, same machine, back-to-back, `-bench ReadPersistedHumanNoteAbsent -count=6`:** the keyless 535 KB `entries=2000` case ran at ~1.33 ms median with the guard and ~1.76 ms median without it; the escaped-journal control stayed inside its noise band, as expected because that path still runs the probes. An independent A/B run by the implementing agent showed the same direction (~0.78 ms with the guard vs ~1.34 ms without). The host was loaded by another session's gate during the recorded parent run, so absolute times are roughly double a quiet host; the attribution is the within-session A/B.
- The benchmark also pays a fixed 549 KB read/allocation per op, so the total drop is not the full scan ratio; the removed work is the ~20 redundant full-document scans.

# Environment note

`agent/sandbox`'s read-only bwrap tests fail when `TMPDIR` is under `/tmp` — they shadow `/tmp` read-only, so bwrap cannot create its work dir — at the unmodified base as well. They pass with `TMPDIR` outside `/tmp` (e.g. `/var/tmp`). The gate runs them in short mode.

# Gates

On this head (`84c8dd717`):

- `make merge-approval-gate` exit 0 — `lint-naming`, `lint-gofmt`, `lint-evenerfuzz`, `lint-eval`, `lint-internal`, `lint-golangci`, `lint-generated`, `lint-fuzz-registry`, build, the root/agent/llm/auth/envvars/invariant/identifier module runs (agent 52.9s, llm 15.7s), web (1108.8s), `test-native` (740 + 778 tests, `tsc --noEmit`) and `test-api-package` (packed-package qualification) all passed; the run log has no failure lines.
- `make vet` exit 0.


obra:
Round-1 answer, on head `215af1dd8`:

**Low — the added comment embeds process metadata.** Fixed in `6f334bf84`: the parenthetical is gone, and the comment states only the rule and its cost ("Every probe contains a backslash byte, so a document with no backslash byte cannot hold any escape spelling and the probe loop is skipped after one single-byte scan. A document that does contain a backslash pays one extra single-byte pass before the probes.").

Nothing else moved: the guard is byte-for-byte the change you reviewed, and the equivalence argument and the guard-on/guard-off benchmark A/B in the description are unchanged. The full local gate (`make merge-approval-gate` exit 0, `make vet` exit 0) ran on `84c8dd717`; this head is that commit plus the comment edit and a clean merge of `main` (`fabb4d875`), and CI runs the full suite on it.

PR #1315, opened 2026-09-14
feat(web): open a depth-truncated child session directly from the activity tree (#1270)

### LINKED ISSUE #1270: Activity panel: offer to open a depth-truncated child session directly when the delegate carries no continuation
From #1269 (the #1231 reset): a depth-truncated delegate is now reported as truncated with no continuation token and a diagnostic (`request session "<id>" directly`). The wire already carries `JobActivityDelegate.ChildSessionID`, and a direct request for that session returns the page the old token returned (the boundary token never carried a resume position). But both frontends render a paging affordance only when a token exists, so today the diagnostic is visible only in the details view and there is no one-tap path to the child.

Wanted, in the shared activity panel (web `panes/session/chrome/ActivityTree.tsx`/`ActivityRowDetail.tsx`, native `ActivitySheet.tsx`, or the shared `activityRows` model once D17/D18 of the SDK plan lands): when a delegate is `Truncated` with no continuation and has `ChildSessionID`, render "open session" that navigates to that child as its own thread (native already treats a live child as addressable; the web has the session route). No protocol change.

Sequence: after #1269 merges; small enough to land before D17.

obra:
Closes #1270.

#1269 settled the agent half: a delegate stopped at the depth or continuation-path bound is `Truncated` with **no** continuation, on purpose. A token there would carry nothing a request for the child does not — a depth-truncated token names that child as a fresh root at position 0, so the page it fetches is the page a direct request fetches, and it would have to name generations nothing loaded. The ruling was that the branch says where to read instead, and the frontend affordance was left to this issue.

Both frontends offered a paging control only when a token existed, so those subtrees had no way out of the panel at all.

**The affordance keys on what the wire already carries** — `Truncated`, no continuation, and `ChildSessionID`/`childRef` — not on parsing the diagnostic text. No protocol change.

- **Web** (`panes/session/chrome/ActivityTree.tsx`): the branch strip renders `Open session` beside the existing message, which routes to that child through the app's ordinary session route.
- **Native** (`mobile-native/src/ActivitySheet.tsx`): the branch footer renders `Open session`, using the same `openSession` the sheet already uses for a delegate's child. `ActivityList.branches()` (shared, `appwire-client/typescript/activityList.ts`) now names the child ref for a delegate branch truncated with no token, so both the model and the sheet stay honest about what a branch can offer.

A third, smaller commit extracts `openSessionByRef` next to the existing placement helpers and points the rail at it. The rail was the only thing that knew the URL round-trip, and the round-trip is the part that matters: the router is what opens a nested session beside its top-level owner instead of dropping a subagent into main next to an unrelated root. A depth-truncated child is nested by definition, so it needs exactly that.

Tests, RED first on both sides:
- `ActivityTree.test.tsx`: a truncated tokenless delegate offers `Open session` and clicking it opens the child's ref; a branch that still has a token offers `Load more` and no open-session link.
- `activityList.test.ts`: the truncated tokenless delegate branch names the child to open; a pageable branch names none.

Gates: `make test-web` (typecheck, test, lint incl. biome over the frontend src) and `make test-native` (742 + 778 tests, plus its own `tsc` check), both green. `make test-web-browser` not run: no guard exercises the activity tree — `overflowguard`'s "Activity" strings are the content-filter labels, not this panel. No Go changes.


obra:
## Review round 1 disposition (7f866cfc8)

CI: the `web` job failed; the failing test is being read before this round is dispatched.

RoboRev (3 of 4 reviewers on the Medium):
- **Medium, native `branches()` reads truncation and continuation from the delegate's own branch only, while the web merges the child session's branch too** — accepted, and the class closes: one rule for "is this delegate branch truncated, and does it have a continuation" lives in the shared package's `ActivityList` and both frontends consume it, so web and native cannot diverge on it again.
- **Low, tests cover only `delegate.branch` shapes** — accepted: the `delegate.child.branch` shapes, truncated with and without a continuation, on both sides.

obra:
## Round 1 fixes

**1 (Medium) — `247024706` `fix(web,native): read both of a delegate's branch states in one place`.** RED on native first:

```
a rendered child truncated with no continuation names the session to open
AssertionError: expected undefined to be 'local:sess_deep_child'
```

A delegate row has two branch states and they are not interchangeable: its own is what the **depth** bound truncates, before the child is loaded at all, while a **continuation-path** truncation and any size trim inside the child land on the *child's*. The web merged them inline; `branches()` read only the delegate's own, so exactly the cases that stopped below a rendered delegate offered nothing and stayed unreachable.

`activityDelegateBranch` in the package answers it once — the child's token wins where it has one, either branch being truncated truncates the row, and a truncated row with no token names the child to open — and both surfaces call it. The web's inline merge is deleted rather than left as a second copy of the rule, and `ActivityBranch` now extends the shared shape instead of redeclaring `openSessionRef`.

**One placement note.** The brief said "beside `ActivityList`"; I put it in `activityData.ts` instead, one module below. It interprets `ActivityDelegate`/`ActivityBranchState`, which live there, and `activityList.ts` already imports that module — so this keeps the dependency pointing one way. Putting it in `activityList.ts` would have made the web's tree import the list controller it otherwise has nothing to do with, and the web tree has no re-export seam for that module. Say the word if you want it moved anyway.

**2 (Low).** Both sides gained a rendered-child pair: child branch truncated without a continuation offers `Open session`; with a continuation offers `Load more` and no open-session link. The native pair was red before the change; the web pair passes either way, since the web already merged — they are there as the guard that keeps the two surfaces from drifting apart again, which is the whole point of the shared function.

Gates: `make test-web` (typecheck, test, lint), `make test-native` (744 + 778), `make test-api-package` (`qualified @evener/appwire-client@0.1.0`). All green.

obra:
## Review round 2 disposition (247024706)

CI green. RoboRev (3 of 4 reviewers on the one finding):
- **Medium, `branches()` now lists a rendered child session twice: once through its delegate (which round 1 made carry the child's continuation) and once as the child node itself** — accepted. One owner per token: the delegate row is the one that answers for its child, so the child session node is not appended separately, and the native test asserts the branch list's exact contents rather than searching it.

obra:
## Round 2 fixes

**`13f05baab` `fix(native): one owner per activity branch token`.** RED first, and the duplicate shows up literally — the exact-contents assertion reports the extra row rather than a count:

```
- Expected
+ Received
+   {
+     "id": "session:sess_deep_child",
+     "label": "Deep child",
+     "truncated": true,
+   },
```

Round 1 gave the delegate row its rendered child's branch, and `branches()` went on appending that child as a row of its own, so the same state came back twice under two ids. Only the root speaks for itself now; every other session is reached through a delegate, so the walk visits delegates and recurses into their children without appending them.

The test asserts exact arrays, not `find`/`some`, for all three shapes: root only (`session:thread` with its token), a delegate with a truncated child (one `delegate:dlg_deep` row carrying `openSessionRef`), and a delegate with a pageable child (one row carrying `child-cursor`).

**The web does not have this defect.** `collectContinuations` pushes the root's strip, then strips only for delegate entries, recursing through `visitDelegates(delegate.child)` without appending the child node — it already had the shape this change gives `branches()`, so it is unchanged. Its existing assertions would catch a regression anyway: `getByRole` throws on multiple matches, so the rendered-child tests from round 1 fail if a second `Open session` or `Load more` ever appears for one branch.

Gates: `make test-web` (typecheck, test, lint), `make test-native` (745 + 778), `make test-api-package` (`qualified @evener/appwire-client@0.1.0`). All green.

obra:
## Review round 3 disposition (13f05baab)

CI green. RoboRev (Medium ×2):
- **Medium, a child session's continuation exposed under the delegate id may make `graftContinuationTree` merge the child page's ancestor data over the delegate's own metadata** — accepted as a measurement: one reviewer could not construct the failing case, so the regression test the finding describes (child pagination with newer parent metadata) is written first; if it is red, the graft target and the display owner become distinct ids; if green, the test stays as the guard and nothing else changes.
- **Medium, `activityDelegateBranch` and its type are not re-exported from the package root, and an exported type now extends an unexported one** — accepted; real, and A4's rewriter would refuse the deep import the moment the seam is gone.

obra:
## Round 3 fixes

**1 (Medium) — measured, and it is GREEN. `bb2d9f9c5` keeps the test as the guard; no behaviour change.**

The regression test the finding describes is in `activityMerge.test.ts`: a delegate the panel displays at `projectionRevision: 5`, running, with a parent-side mandate, warnings and a later `latestActivityAt`; a child page grafted under `delegate:delegate` whose ancestor-chain copy of that same delegate is the stale one (`projectionRevision: 2`, terminal, completed, no mandate). Every newer field survives the graft, and the child's entries still extend with its branch consumed.

The reason it holds, which is why the other reviewer could not build a failing case: `revisionFencedDelegate` decides authority on `projectionRevision` for an ordinary delegate **regardless of `withinTarget`** — the target flag only selects which branch state wins, and there the page that just answered the continuation *is* the authority. So a delegate-keyed graft cannot overwrite newer parent-side metadata with an older chain copy.

The two ids were already distinct and stay so, now pinned on the web side too: the delegate row owns the strip and keys the graft and the panel's failure map (`onContinue` is asserted to receive `delegate:dlg_deep`), the child's token is what the request carries (`token_child`), and the child's ref is what `Open session` opens. Nothing to separate that was not separate.

**2 (Medium) — `58b317a51`.** `ActivityDelegateBranch` (type) and `activityDelegateBranch` are exported from the package root; `ActivityBranch` no longer extends a type a consumer cannot name. The qualification manifest enumerates root values, so it names the function too, and `make test-api-package` re-qualifies clean. Worth stating the motive plainly: this is what A4's import rewriter would refuse once the web's `protocol/` seam is gone, since the deep path is the only way to reach it today.

Gates: `make test-web` (typecheck, test, lint), `make test-native` (745 + 778), `make test-api-package` (`qualified @evener/appwire-client@0.1.0`). All green.

PR #1317, opened 2026-09-14
fix(web): the skill guard waits for the composer to settle before typing after a drain

obra:
Refs #1308.

## What happened

Run 34880821304's web job failed on the insert that follows the queue drain — caught by the assertion #1311 added, which is why there is anything to read:

```
skillguard FAIL: typeText: caret at 0 after the edit, expected 50
```

with `drain-committed 18:32:44.451Z` and `drain-released 18:32:44.534Z` immediately before it. The expected caret of 50 is the length of `PROSE_STEER_TURN_14e open a long turn for steering`, so it is the steering scenario's first insert.

One detail refines the diagnosis: **the value the edit wrote survived and the caret did not.** The post-condition checks the value first and it passed, then the caret check failed at 0. So a render landed inside the `dispatchEvent`, reset the selection, and the value came back through React's own `onChange`. The drain's store update re-renders the controlled textarea; an edit landing in the same tick is overwritten by the render that follows it. Main's web job passes on the same driver, so it is a race, not a deterministic break.

## The change

`typeText` waits for the composer to stop changing before it edits — two reads 80ms apart that agree. **The wait is inside `typeText`, not at the call sites**, so no scenario can forget it when a new clearing action appears.

Settling is not a proof that no further render is coming, so the edit is also made **idempotent**: it computes the whole intended value once and assigns it, rather than inserting relative to whatever the selection is at dispatch time, and re-applies that same absolute value if the result comes back wrong. Re-applying cannot double-type, which re-inserting would. Each re-apply is logged, so an artifact shows the app fighting the driver rather than quietly papering over it.

No timeouts changed.

## Proof

Forced the race by wiping the textarea one tick after the first edit:

- against **main's** driver: `skillguard FAIL: typeText: composer holds "" after the edit, expected "PROSE_ALPHA_14a run the fixture check on the gamma channel"`;
- against **this** driver: the guard passes, and the log carries `skillguard: a re-render landed on the edit (value "", caret 0); re-applying (attempt 2/4)`.

Stress: three runs of `-count=5` (fifteen iterations). Two were clean; one iteration failed at the **capability-loss** step, which is the separate #1308 failure and not this one — and it reproduced locally with #1312's instrumentation live, which answers the question I left open on that step. Evidence posted to #1308: shutdown sent at +0ms, helper exited at +5ms, roster refreshed at +6ms, and the pane still showed "This session hasn't started yet" 30 seconds later. All three handoffs were fast; the web UI never re-rendered the session as ended. That is a product bug, not load.

`make lint` (nine targets) clean. `biome` on `scripts/skillguard/` reports a pre-existing import-sort complaint on the unmodified file, which is outside the gate's enforced scope (`biome ci src ../../../appwire-client/typescript`, clean) — AGENTS.md says not to "fix" those, so it is untouched.


obra:
## Review round 1 disposition (bde2cdce0)

CI: the `tests` job (Go) failed on this JavaScript-only head; the failing test is being read and, if it is one of the known flakes, rerun.

RoboRev:
- **Medium, the retry re-applies a snapshot computed before the settle, and the verification runs inside the same evaluate as the edit** — accepted: the page-side edit becomes a compare-and-swap against the settled base (recomputed on every retry), and a post-edit settled read is the verification, so a render that lands after the edit is caught.
- **Low, the per-character history comment now sits above the wrong method** — accepted, and history goes to the commit message rather than the file.

obra:
## Round 1 fixes

`7cbb5b6fb`. `origin/main` re-merged: already up to date. (The Go `tests` job failing on a JS-only head is yours to read; nothing here touches Go.)

**1a — compare-and-swap.** The retry re-applied the value computed from the *first* settle, so a legitimate change arriving in between — a drain committing, a turn clearing the draft — was overwritten with content derived from a value the composer no longer held. The page-side edit writes only if the textarea still holds the settled base; otherwise it declines and reports what it found, and the driver re-settles and recomputes `next`/`caretWant` from that. Bounded at four attempts, each logged with which window it was.

**The swap compares the value alone, deliberately.** A render that resets the selection without touching the text is the exact failure this guard hit, and recomputing the insertion point from a caret that render moved would type in the wrong place. The settled selection is the one the scenario meant.

**1b — settled read-back.** Verification ran inside the same evaluate as the write, so a render landing just after it returned was missed — which is precisely how the CI failure reported `caret 0` while the value read back correct. The result now comes from a settled read, value and caret both, and a mismatch takes the retry path.

**Proof, both windows forced.** Injecting a change between the settle and the compare: `skillguard: the composer changed under the edit (holds "!", expected ""); retrying (attempt 2/4)` — the stale write was declined, not applied. Wiping the textarea a tick after the write: `skillguard: a render landed on the edit (holds "", caret 0-0, expected "!PROSE_ALPHA_14a …" with the caret at 59); retrying (attempt 3/4)`. With both harnesses live at once the run still converged and finished clean.

**2 — comments.** The per-character history is gone from the driver and lives in #1311's commit message; `composerEditStateExpr`, `settleComposer` and `typeText` each say what they do and why, not what they used to be.

### Gates on `7cbb5b6fb`

The guard against a built frontend, then four more iterations (`-count=4`, `ok 67.384s`), all clean. `make lint` (nine targets). `biome` reports the same one pre-existing import-sort complaint on the unmodified file header, outside the gate's enforced scope.

obra:
## Review round 2 disposition (7cbb5b6fb)

CI green. RoboRev (3 of 4 reviewers on the Medium):
- **Medium, a caret-only mismatch after a successful swap takes the retry path and inserts the text a second time** — accepted; the very double-type round 1 claimed impossible. A settled value equal to `next` means the text landed: only the caret is re-applied, and re-insertion happens only when the value is wrong.
- **Low, the comment claims all three defenses share the retry budget while a settle timeout throws outright** — accepted; the comment says what the code does.

obra:
## Round 2 fixes

`f4b095e4f`. `origin/main` re-merged: already up to date.

**1 — the reviewer found a double-type I introduced while claiming one could not happen.** When the swap landed and a render reset only the *selection*, the settled read-back saw a good value with a bad caret and fell through to the retry; that retry re-settled, found a base already holding the text, and computed an insertion from it — typing the whole draft twice.

Forced with a selection-only reset a tick after the write, the previous code sends:

```
skillguard FAIL: selection changed the prose: "PROSE_ALPHA_14a run the fixture check on the gamma
channelPROSE_ALPHA_14a run the fixture check on the gamma channel"
```

The intended value is now remembered across attempts, and **a base that already equals it means the text is applied: the repair is the caret alone**, placed and then verified through another settle. Recomputing an insertion is reserved for a base that does not hold the text — which is also why a *declined* swap forgets the target, since nothing was written. Under the same harness this lands the prose exactly once and logs `a render moved the caret to 0-0, expected 58; retrying (attempt 2/4)`; the `chip-added` milestone carries the prose undoubled.

**2 — I narrowed the comment rather than moving the settles.** They are already inside the attempt loop, but their failure mode is not retryable in a useful way: a settle that never settles, or a composer that loses focus, throws, because retrying a textarea that will not stop changing only delays the same verdict with a worse message. The comment now says the budget covers the swap and the post-write verification, and names what fails outright.

### Gates on `f4b095e4f`

The guard against a built frontend, then `-count=3` (`ok 50.534s`), all clean. `make lint` (nine targets). `biome`: the same one pre-existing import-sort complaint on the unmodified header, outside the gate's enforced scope.

obra:
## Review round 3 disposition (f4b095e4f)

CI green; three of four reviewers clean. One Medium, accepted: the settle and the edit address `document.activeElement`, so with two composers mounted a focus change during the settle can land the text in the other session if its value happens to match. The driver types into the composer it names by session ref, verifying that element's identity and focus before applying or repairing.

obra:
## Round 3 fixes

`a2b4e3f12`. `origin/main` re-merged: already up to date.

`settleComposer` and the edit resolved `document.activeElement`, which answers "whichever composer has focus", not "the one this scenario is typing into". Both now resolve the session's own composer through `composerSelector(ref)` — the way `focusComposer` already does — and refuse to read or write unless that textarea is the focused element. The ref is threaded through `typeText(ref, text)` and `settleComposer(ref)`; every call site already knew it (fifteen of them, one on session B), and a focus loss names the ref.

**The harness you asked for could not be built, and here is the measurement that says why.** I instrumented every settle poll of a whole guard run to count composer textareas in the page:

```
a + b > 1 or all > 1 : 0 occurrences
```

The app mounts the open session's pane only — never two composers at once — so an injected focus steal to session B found no textarea to focus, and `document.activeElement` could not in fact have resolved the wrong session today. **That makes this a latent hazard rather than a live one.** The fix still stands on its own: the code stops depending on a property that merely happens to hold.

**The focus check itself is proved.** Blurring the textarea during the settle:

```
skillguard FAIL: typeText(local:034Oqc88c2gJJp3CrIzz8T): composer is not the focused element
```

and nothing is written — the edit never reaches the page when the check refuses.

### Gates on `a2b4e3f12`

The guard against a built frontend, then `-count=3` (`ok 53.263s`), all clean. `make lint` (nine targets). `biome`: the same one pre-existing import-sort complaint on the unmodified header, outside the gate's enforced scope.

obra:
## Review round 4 disposition (a2b4e3f12)

CI green; two of four reviewers clean. One Medium, accepted: the retry decides "text landed" by whole-value equality with the intended result, so a render that reverts the value and resets the caret is treated as a fresh insertion base (text ends up in the wrong place), and a render that normalizes the value while keeping the text is treated as absent (text typed twice). The decision becomes: is the typed run present at the intended position in the settled value; present means caret repair only, a reverted snapshot re-applies the remembered result, and only a genuinely absent run recomputes from the current selection.

obra:
## Round 4 fixes

`d4a10ddc8`. Merged `origin/main` first.

Both cases are real, and both came from the same mistake: comparing the settled value to the value the driver *wrote*.

- **(a) revert + caret reset** sent the retry down the recompute path, where the current caret is 0 — `"abc"|3` typed `"abcX"`, reverted to `"abc"|0`, retry produced `"Xabc"`.
- **(b) normalize/pad after a good swap** failed the same equality with the typed run still present, so the retry typed it a second time.

**The question a retry turns on is now whether the typed run sits where it was inserted**, measured from the remembered pre-edit selection. Present → the text is in, whatever else the value holds is the app's, and the repair is the caret alone (the later payload assertions judge the rest). Back at the remembered snapshot → a revert: re-apply the remembered result at the **remembered** selection, not at the caret that came back with it. Neither → a concurrent edit, recompute from what the composer holds now.

Forced both against the real app:

```
revert+caret-reset : {"value":"PROSE_ALPHA_14a … channel /probe","caret":65,"insertedAt":58}
pad-after-swap     : {"value":"PROSE_ALPHA_14a … channel   /probe","caret":67}
```

The first lands the token at 58, not at 0. The second lands the prose **once**, the harness's two spaces intact — and the scenario's later assertion then fails on that padding, which is exactly the app's value being judged where it should be rather than silently re-typed. The round-2 caret-only harness and the round-1 declined-swap harness still take their own paths (`a render moved the caret to 0-0, expected 58`; `the composer changed under the edit (holds "!", expected "")`).

### Gates on `d4a10ddc8`

The guard against a built frontend, then `-count=3` (`ok 49.557s`), all clean. `make lint` (nine targets). `biome`: the same one pre-existing import-sort complaint on the unmodified header, outside the gate's enforced scope.

obra:
## Review round 5 disposition (d4a10ddc8)

CI green; three of four reviewers clean. Both findings accepted:
- **Medium, a declined swap clears the remembered target, so the next attempt skips the presence check and can recompute from a value that already holds the run** — the remembered target survives a declined swap; the presence check runs first on every attempt.
- **Low, `press` still acts on whatever has focus while `typeText` and `selectAll` are scoped to a session** — the same class round 3 closed for typing; `press` takes the ref and verifies or restores focus in the same evaluate before dispatching.

obra:
## Round 5 fixes

`215756b67`. `origin/main` re-merged: already up to date.

**1 — the remembered edit survives a declined swap.** A swap can be declined because a render landed *after* the write as easily as before it, and forgetting the target there sent the next attempt past the presence check and into a recompute from a base that already held the run — typing it twice, which is the exact bug the presence check exists to prevent. Every attempt now asks "is the run where I put it?" first, and the target is replaced only when that says no *and* the value is not the snapshot either, which is a genuine concurrent edit.

Forced, by declining the swap with the run already in the base:

- previous code: `skillguard FAIL: selection changed the prose: "PROSE_ALPHA_14a run the fixture check on the gamma channelPROSE_ALPHA_14a run the fixture check on the gamma channel"`
- this one: `the composer changed under the edit (holds "PROSE_ALPHA_14a …", expected ""); retrying (attempt 2/4)`, and `chip-labels` carries the prose **once**.

**2 — every key names its session.** `press` dispatched to whatever the page had focused while `typeText` and `selectAll` had been made ref-scoped, so a Backspace could delete a draft the scenario was not editing. It takes the ref, puts focus back on that composer **in the page immediately before the key** rather than trusting whatever ran last, and logs when it had to restore. All four call sites pass their ref (three `sessionA`, one `sessionB`).

Forced by blurring before the trailing Backspace: `skillguard: local:…: focus was elsewhere before Backspace; restored it`, and the composer still ends holding exactly the prose — the leading completion space removed from the right draft.

(As measured in round 3, the app mounts one composer at a time, so "the other keeps its text" has no second composer to assert against; the restore path is what is provable here.)

### Gates on `215756b67`

The guard against a built frontend, then `-count=3` (`ok 49.164s`), all clean. `make lint` (nine targets). `biome`: the same one pre-existing import-sort complaint on the unmodified header, outside the gate's enforced scope.

One run during this round failed at the capability-loss step (#1308's third failure, the product bug I reported there); it is unrelated to this change and did not recur across the stress runs.

PR #1321, opened 2026-09-14
fix(activity): keep session-level diagnostics through the parser and show them in the details view (#1316)

### LINKED ISSUE #1316: Activity panel drops session-level diagnostics (path-limit, absent/unreadable journal) before the UI sees them
Noticed while implementing #1270 (PR #1315). The activity parser keeps *delegate* diagnostics, so the depth-bound message (`request session "<id>" directly`) survives to the details view, but the session-level diagnostics that #1269 added — the path-limit variant and the absent/unreadable-journal variants — are dropped before the UI sees them. The open-session affordance in #1315 does not need them (it keys on `Truncated` + no continuation + `childRef`), but a user hitting the journal cases today gets a truncated tree with no explanation.

Wanted: keep session-level diagnostics in the parsed model and surface them in the details view the way delegate diagnostics are, on web and native (or in the shared `activityRows` model once D17/D18 lands). No protocol change.

### LINKED ISSUE #1322: Activity: root-session diagnostics have no surface in the details view (web + native)
Follow-up to #1316 / PR #1321. The parser now keeps session-level diagnostics (path-limit, absent or unreadable delegates journal) on every `ActivitySessionNode`, and delegate rows render their child session's diagnostics through `activityDelegateDiagnostics`. The root session's own diagnostics (an unreadable or torn journal on the root itself, asserted in `agent/jobs_activity_past_test.go` around line 2031) are kept in the model but never shown: the root has no row, and both the web continuation strip and native `branches()` footer render only on continuation, truncation or error, not on diagnostics alone.

Wanted: show root-session diagnostics somewhere a user hitting the journal cases will see them (the continuation strip / footer is the obvious place), on web and native, with a test on each surface. No protocol change.

Also noted from the same PR: mobile-native has no component test harness (vitest over pure modules only), so native rendering is proven by the shared rule's tests plus tsc. Not blocking this issue, but the gap is why the native side of this and #1316 has no render test.

obra:
Fixes #1316. Closes #1322.

## What was dropped, and where

`JobActivitySession.diagnostics` is where #1269 stamps the session-level sentences: the continuation-path bound (`continuation path limit reached; request session "<id>" directly`, agent/jobs_activity.go `activityUnreachableByPathDiagnostic`) and the journal conditions (`delegate_journal_line_too_long: ...`, `delegate_journal_torn_tail: ...`, agent/jobs_activity_past.go). `parseSession` in `appwire-client/typescript/activityData.ts` never read that field, so every session-level diagnostic was gone before either UI saw the tree, while a delegate's own diagnostics (the depth bound) survived to the details view.

## Change (no protocol change)

- `ActivitySessionNode.diagnostics?: string[]`, parsed with the same string-array strictness a delegate's get (malformed marks the branch incomplete, like every other field).
- Merges carry it the way they carry `branch`: the page that speaks for a session supplies its diagnostics (the target of a graft in `mergeSession`, the whole page on a refresh in `fenceSession`), and `cloneSession`/`fenceSession`/`mergeSession` each take their own copy through one `cloneDiagnostics`.
- One shared rule for rows, `activityDelegateDiagnostics(delegate)`, beside `activityDelegateBranch` and for the same two-places reason: the delegate row is the only row its child session has, so the row's details view says the delegate's own sentences first, then the rendered child's, each once. Web `ActivityRowDetail` and native `ActivityDelegateDetails` both read through it. Exported from the package and listed in the qualification script.
- The root has no row, so its diagnostics are the panel's: the web `ActivityPanel` lists them beside the "Activity coverage is incomplete." strip, and `ActivityList.branches()` lists the root with `diagnostics` even when nothing was cut short, which the native `ActivitySheet` footer renders one line each. A delegate's child sentences never make a footer row.

## Tests proving each surface

- Parser: `activityData.test.ts` "parses a recursive wire-true tree ..." asserts root and child diagnostics survive; `activityData.edge.test.ts` "parseSession diagnostics" covers the malformed root, malformed child, and absent cases.
- Shared row rule: `activityData.test.ts` "activityDelegateDiagnostics" (own then child, child-only, own-only, dedupe/empty).
- Merge: `activityMerge.test.ts` "a grafted child page keeps the session diagnostics it reported, as the tree's own copy", "a page for an already rendered child replaces that session's diagnostics and leaves the root's alone", "a root refresh takes the root's diagnostics from the page, as the tree's own copy" (each mutates the source afterwards to prove the copy).
- Web row: `ActivityTree.test.tsx` "a rendered child's session-level diagnostics show in the delegate's detail strip" (the path-limit sentence renders alongside the #1315 Open session button).
- Web root: `ActivityPanel.test.tsx` "shows the root session's diagnostics beside the coverage strip".
- Native root: `mobile-native/src/activityList.test.ts` "reports the root's diagnostics even when nothing was cut short, and never a child's" (the pure module that builds the footer rows).
- Native row: no component test harness exists in mobile-native (vitest over pure modules only); the Diagnostics section is covered by the shared rule's tests plus strict tsc in `make test-native`.

## Gates

`make test-web`, `make test-native`, `make test-api-package`, `make lint` all green; `biome ci src` and `biome ci appwire-client/typescript` clean.


obra:
## Review round 1 disposition (47d3fb398)

Two Medium and one Low, all accepted:
- **`mergeSession` keeps only entries, ref, label and branch from a continuation patch, so a diagnostic produced on a later page for an already-displayed session is dropped** — accepted; diagnostics follow the same target rule as `branch`, and the merge test starts from a tree that already holds the child.
- **Root-session diagnostics are parsed but never rendered** — accepted and folded in rather than deferred: I had filed it as #1322 on the implementer's report, and the issue this PR closes asks for the diagnostics to reach the details view on web and native. The root's diagnostics render beside the existing coverage strip on web and in the native footer, with a test on each; #1322 closes with this PR.
- **`summarizeSession`/`fenceSession` share the diagnostics array by reference** — accepted; cloned like `cloneSession`.

obra:
## Round 1 fixes

Head ead732948 (origin/main fd115bd82 merged first, clean).

1. **M, `mergeSession` dropped `patch.diagnostics` for an existing session** — fixed. Diagnostics now follow `branch` exactly: within the target the patch's win, elsewhere the current's, each cloned through one `cloneDiagnostics`. `fenceSession` already let the incoming page win (a refresh is the whole statement), so the two paths agree on the same rule, "the page that speaks for the session supplies them", and no comment was needed beyond the code sitting next to `branch`. RED first: `activityMerge.test.ts` "a page for an already rendered child replaces that session's diagnostics and leaves the root's alone" (starts from a tree that already has the child with a stale sentence; the continuation page's sentence replaces it and the root's own is untouched).
2. **M, root diagnostics parsed but never rendered (#1322)** — folded in. Web: `ActivityPanel` renders `root.diagnostics` as `.state` lines beside "Activity coverage is incomplete."; RED `ActivityPanel.test.tsx` "shows the root session's diagnostics beside the coverage strip". Native: `ActivityList.branches()` lists the root with `diagnostics` even when nothing was cut short, and the `ActivitySheet` footer renders one line per sentence (the "Some activity is not loaded." headline now only appears for a branch that actually stopped); RED `mobile-native/src/activityList.test.ts` "reports the root's diagnostics even when nothing was cut short, and never a child's". Diagnostics ride only on the root's branch, since a delegate's are its row's (`activityDelegateDiagnostics`); the `ActivityBranch` field comment says so. PR body now carries "Closes #1322".
3. **L, `fenceSession`/`summarizeSession` shared the array by reference** — `fenceSession` now clones it; `summarizeSession` only ever receives a session its caller just built with a cloned array (`mergeSession` and `fenceSession` are its two callers), so it takes no second copy. RED `activityMerge.test.ts` "a root refresh takes the root's diagnostics from the page, as the tree's own copy" mutates the source after `fenceRootSession` and also pins that a page reporting nothing clears the old sentence.

Gates: `make test-web`, `make test-native` (746 + 778, strict tsc), `make test-api-package`, `make lint` green; `biome ci src` and `biome ci appwire-client/typescript` clean.

Issue #1286, opened 2026-09-14
flake(tests): TestDelegateResourceSupervision_CommittedSendStartRefusesASecondTurn sees target_busy instead of started


obra:
CI flake in the `tests` job on PR #1282 (frontend-only change), agent shard 4:

```
--- FAIL: TestDelegateResourceSupervision_CommittedSendStartRefusesASecondTurn (0.08s)
    delegate_resource_supervision_test.go:255: committed send start = {... Status:failed Reason:construction_failed ... Output:target_busy}, want started
```

- Run: https://github.com/prime-radiant-inc/evener/actions/runs/34800816558/job/103842965595 (shard `agent:4`; `agent:0`..`agent:7` otherwise passed)
- The test is timing-sensitive and was recently reworked for claim handling (#1236), which is where the race likely lives: a `delegate_send` start that should be `started` can observe `target_busy` / `construction_failed` under sharded parallel load.
- Not reproducible locally: `go test -run '^TestDelegateResourceSupervision_CommittedSendStartRefusesASecondTurn$' -count=25 .` in `agent/` passes 25/25 on an idle-ish host.
- Unrelated to the PR under test (frontend transcript render change; the `web` job passed).

Please make this deterministic (or retire it) so it stops failing unrelated PRs.

obra:
Another occurrence: PR #1325 @b992782f5 (hub-only change, nothing under agent/), run 35004165924, race-modules/agent: delegate_resource_supervision_test.go:255 committed send start Status:failed. Rerunning the job.

obra:
Another hit 2026-09-16 19:57Z on #1569 @61a9c9ffb (D14-1, web-only diff, agent package untouched): `delegate_resource_supervision_test.go:259: committed send start = {Target:dlg_… Type:delegate …}` in the `tests` job (agent shard 5). Rerun requested.

obra:
Fourth occurrence, same assertion, again on a diff that cannot reach `agent/`:

- PR #1647 @ `00a0e22ef` (frontend-only: delete the production-dead `skillSelections` module, cover both ARIA close routes, give the attachment fake a real selection range), [run 35178603730, job `race-modules / agent` (105065705936)](https://github.com/prime-radiant-inc/evener/actions/runs/35178603730/job/105065705936).
[...]

PR #1325, opened 2026-09-14
fix(hub): a session whose daemon exits for good never told subscribers to re-read, so the web pane never rendered it as ended

### LINKED ISSUE #1318: web: a session whose helper daemon exits is never rendered as ended (pane stays on 'hasn't started yet')
Product bug separated out of #1308 (the skill-guard flake), proven by the guard's capability-loss step with #1312's timeline live (local stress run, 2026-09-14):

- shutdown sent to helper B: +0 ms
- helper B exited: +5 ms
- roster refreshed: +6 ms
- 30 s later, session B's pane still reads "This session hasn't started yet" / "Send the first message" instead of the ended state ("Send a follow-up…").

Every hub-side handoff was fast; the web UI never re-rendered session B after its helper disappeared from the roster. On the CI runner the same step failed at 30 s in run 34851914689 (PR #1269). This is the scenario the `TestSkillComposerBrowser` capability-loss assertion exists to catch, so it is a real gap in how the web app reacts to a capability loss, not a timing problem in the guard.

Needs a product owner. The guard's remaining flake (#1308) is the driver's typing races; the driver's own fixes are #1295, #1311, #1317.

obra:
Fixes #1318. Refs #1308.

## Cause

A daemon announcing its own close sends `thread/status/changed {closed}` and shuts its socket microseconds later. The hub relay (`internal/appsource/relay_session.go`) revokes the publication epoch the instant the socket closes, so any frame still queued behind a busy listener at that moment is discarded — by design, and pinned by `TestRelaySessionStaleEpochNotificationCannotPublish`: the reconnect publishes one `evener/thread/resync` and the client re-reads. A daemon that exited never reconnects, so no resync ever came. The web pane kept the pre-exit model (`idle`, "Message the agent…") for as long as the tab stayed open; nothing on the wire would ever correct it.

Measured on a quiet host with a real hub and a real `evener serve` exiting through a direct `thread/shutdown`: in 2 of 5 runs the frame right behind the close (`thread/closed`) was already lost. On a loaded CI runner the close itself loses the same race, which is what the guard's capability-loss step caught.

(#1312's "composer state null" was the dump of session **A**'s composer, whose pane dockview unmounts while B's tab is active; it was never evidence about B.)

## RED test

- `TestRelaySessionDaemonGoneTellsListenersToReread` (`internal/appsource/relay_session_test.go`): listener subscribed, read committed, the daemon's entry leaves the roster and its socket closes; the listener must receive a resync. Before the fix: nothing in 5 s.
- `TestE2E_DaemonExitTellsSubscriberToReread` (`cmd/evener-hub/e2e_test.go`): the guard's step without Chrome — hand-started daemon, direct `thread/shutdown`, subscribed client receives the resync, and the re-read answers `notLoaded` with `send:true` (what the follow-up composer renders from). Before the fix: 15 s of nothing but navigation invalidations.

## Fix

Recovery (`recoverCanonicalFeed`) publishes one resync per disconnect once a connect attempt finds the daemon gone from the roster — an entry is dropped only when its process is, so a dial failure against a listed entry does not qualify. The pending reconnect resync stays armed for a daemon that comes back. The relay's stale-epoch contract is untouched; the resync lands in ~0.3 s in the e2e.

`TestE2E_HubAndDaemon` now shares the daemon-start helper.

## Gates

`make lint` (nine PASS) · `make test-web` · `make test-api-package` (after `npm ci` in `appwire-client/typescript`; the target does not install its own devDependencies, and `ws` is missing from a fresh checkout) · `npx biome ci src` · `gofmt -l` (toolchain binary) · `go vet ./...` · `go vet -tags evenerfuzz ./...` · `GOOS=windows go vet -tags evenerfuzz ./...` · `go test -race ./cmd/evener-hub/...`: subpackages `-count=3` in one run; the root package as three separate `-count=1` runs (205 s / 234 s / 204 s, all ok), because `-count=3` of the root package exceeds go test's 10 min default and a `-count=2` run timed out at 9 min without a captured dump — noted below, not chased here.

## Noticed, not fixed

- `make test-api-package` fails on a fresh checkout until `npm ci` runs in `appwire-client/typescript` (`ws` devDependency not installed by the target).
- A `-race -count=2` run of `cmd/evener-hub` timed out at 9 min while three `-count=1` runs passed in under 4 min each; whether a test misbehaves on a repeated count is unmeasured.


obra:
## Review round 1 disposition (ba82b09bd)

CI green; three of four reviewers clean. Two Medium from one reviewer, both accepted subject to measurement:
- **The daemon-gone resync is built from one `readParams` and targeted at one ref, so a subscriber on another alias of the same relay session (a read-only child alias sharing the root's relay) stays stale** — accepted if a relay session does serve more than one ref; the round measures that first and, if so, notifies every subscribed route.
- **`goneAnnounced` is set when the resync is queued, and a later reconnect attempt advances the epoch and can revoke that queued publication before dispatch, so nothing re-arms it** — accepted; the terminal notification is published outside the revocable epoch or re-armed when revoked, with a test that revokes between queue and dispatch and still sees the re-read.

obra:
## Round 1 fixes

Head: `fca9e5c5f7ef74bf9c59bcf4629f94d57db65757` (main `8475ae5fe` merged in first).

Both Mediums measured true.

1. **Alias subscribers never got the resync.** `ResolveRelaySessionWithAdmission` maps a read-only child alias to the root's canonical ref, so one relay session serves both, and `publishTarget` routes a resync to the one route its `ref` names. RED: `TestHubRelayDaemonGoneResyncReachesEveryRouteWithItsOwnRef` (root and child subscribed through one session; the child's copy arrived with `ref=""`). Fix: the daemon-gone resync (`appsource.DaemonGoneResync`) names no target, the canonical relay fans it out to every route, and `stampResyncTarget` writes each route's own ref/threadId into its copy. The reconnect resync stays targeted at the canonical ref (`TestHubRelayReconnectRoutesOneResyncThroughCanonicalListener` unchanged). The legacy relay loop is unreachable for local daemons (`prepareRelay` gates it on `!RelaySessionSource`, pinned by the "want neither" legacy-call assertions).
2. **Queued resync revoked by the next attempt.** RED: `TestRelaySessionDaemonGoneResyncSurvivesTheNextReconnectAttempt` (publisher held on an unacknowledged delivery, daemon gone, `advanceEpochLocked` before the ack: the resync was dropped and never re-sent). Fix: the job is enqueued with no epoch and no fence, which `dispatchToListenerAtEpoch` already treats as unrevocable; `goneAnnounced` is unchanged since the job can no longer be lost. Stale-epoch contract untouched (`TestRelaySessionStaleEpochNotificationCannotPublish` green).

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `make lint` (9 PASS), `-race`: subpackages `-count=3` (all ok), root package three `-count=1` runs (221 s / 239 s / 282 s, ok).

obra:
## Review round 2 disposition (fca9e5c5f)

CI green; no Medium or above. Two Low, both accepted:
- **The new process-lifecycle e2e test guards only loopback binding, not the process-inspection capability the roster handling needs** — accepted; it takes the same `e2ecap.RequireProcessInspect` guard its siblings use.
- **`stampResyncTarget` landed between `stampForkCapability`'s doc comment and its definition** — accepted; each function gets its own comment back.

obra:
## Round 2 fixes

Head: `f3f46b81487c0365726e0bf0ca94d7c522b4922b` (origin/main merged in first).

1. `TestE2E_DaemonExitTellsSubscriberToReread` now calls `e2ecap.RequireProcessInspect(t)` right after `RequireLoopbackBind(t)`, the exact pair every sibling e2e that starts a daemon uses (`e2e_test.go:40-41`, `e2e_control_invariant_test.go`, `e2e_queued_input_under_ask_test.go`).
2. `stampResyncTarget` carries its own doc comment; `stampForkCapability`'s original two lines are back on its definition.

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `go test -race -count=1` (root package 196 s ok; all ten subpackages ok), `make lint` (9 PASS).

obra:
## Review round 3 disposition (f3f46b814)

CI green; reviewers 2 and 4 clean. One Medium and one Low, both accepted:
- **With no roster, fallback discovery reads rendezvous files that a crashed daemon leaves behind, so reconnects keep failing to dial and the daemon-gone resync never publishes** — accepted; the fallback classifies a gone daemon the way the roster does (liveness-checked), with a test on the no-roster path.
- **`stampResyncTarget` re-mints params from two fields and drops anything else on the notification** — accepted; it edits the raw map and passes unknown keys through, like `stampClosedThreadCapabilities`.

obra:
## Round 3 fixes

Head: `5eb1a881e05a224a8125cd8fc7bf005c79ea590a` (origin/main already merged; nothing new to merge this round).

Both measured true.

1. **No-roster fallback kept dead daemons dialable.** `newHubSourceRegistry` (app_rpc.go) lists straight from `rendezvous.List` when `cfg.Roster == nil`, and a crashed daemon's file stays, so `relayEntry` kept resolving it and every reconnect was a dial error, never `relayDaemonGoneError`. RED: `TestHubSourceRegistryWithoutRosterTellsSubscribersWhenTheDaemonProcessIsGone` — a real `appserver` daemon stand-in behind a rendezvous file naming a live `sleep` process; the process is killed and the daemon shuts its sockets, the file stays; 10 s with no resync before the fix, resync in under a second after. Fix: the fallback filters entries through the roster's own probe, now exported as `hubcore.ProcessAlive` (rename of the platform-split `processAlive`; `SetProcessAlive` unchanged). Three existing tests that fed synthetic PIDs (501, 101/102, 105) through this path now use `os.Getpid()`/`os.Getppid()`, alive by construction. Production always wires a roster (`main.go:385`), so this path is test-only in practice.
2. **`stampResyncTarget` dropped other fields.** RED: `TestStampResyncTargetKeepsTheOtherFields` (a `reason` field vanished). Fix: edit a `map[string]json.RawMessage`, set `threadId`/`ref`, pass the rest through, mirroring `stampClosedThreadCapabilities`.

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `go test -race -count=1` (root package 197 s ok; all ten subpackages ok), `make lint` (9 PASS).

obra:
## Review round 4 disposition (5eb1a881e)

CI green; reviewers 3 and 4 found no correctness defects. One High, two Medium, one Low, all accepted:
- **High, `stampResyncTarget` assigns into a nil map when params are `null`** — accepted; the map is initialised, with a `Params: null` test.
- **Every resolver failure is wrapped as daemon-gone, including identity-changed and empty-workspace errors for a daemon still listed** — accepted; only not-found is gone, other resolver errors pass unchanged, with a test per class.
- **The roster-less listing trusts signal-0 liveness alone, so a reused PID keeps a stale rendezvous file dialable** — accepted; the fallback reuses the roster's identity check instead of bare liveness.
- **The `goneAnnounced` dedup is only covered indirectly** — accepted; recovery iterates past several backoffs and exactly one gone delivery is observed.

obra:
## Round 4 fixes

Head: `cbbe647739a84981c25decac26608cced9af1319` (origin/main merged in first).

1. **HIGH, nil map.** RED `TestStampResyncTargetToleratesNullParams` (panicked: assignment to entry in nil map). Fix: initialise the map after decoding.
2. **M, over-wide gone.** RED `TestRelaySessionResolverFailureOnAListedEntryIsNotGone` (entry stays listed under `thread-1` but its session id moved, so `relayEntry` fails with "relay session identity changed"; a gone resync arrived). Fix: the resolver's not-found is a typed `localDaemonThreadNotFoundError` (same wire error, same message, `Unwrap` keeps `errors.As(WireError)` working); the connect wraps only that as `relayDaemonGoneError` and returns every other resolver error unchanged.
3. **M, bare liveness.** RED `TestHubSourceRegistryWithoutRosterKeepsOutAStaleFileOnALivePID` (a file naming this test's own PID and a dead endpoint was listed). Fix: with `cfg.Roster == nil` and a `RunDir`, `newHubSourceRegistry` lists through `hubcore.NewRoster(RunDir, &StatusProber{})`, refreshed on every lookup, so the roster's own probe decides: alive with a failed probe is unconfirmed and unlisted, dead is dropped. Nothing new is exported; round 3's `ProcessAlive` export is reverted since nothing needs it. The two ordering tests that relied on unreachable synthetic endpoints now seed `NewRosterWithEntries`; the resume test's stand-in answers the probe (`thread/list` + bare `thread/read`) and carries one id as thread and session id, as a real daemon writes it (`serve.go:1371-1372`). The round-3 chain test still passes through the probed path.
4. **L, dedup coverage.** `TestRelaySessionDaemonGoneIsAnnouncedOncePerDisconnect`: entry list empty, one gone delivery, none across the 100/200/400 ms retries that follow.

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `go test -race -count=1` (root package 195 s ok on the final tree; all ten subpackages ok), `make lint` (9 PASS; the golangci modernize rule asked for `errors.AsType`, applied to both classifiers).

obra:
## Review round 5 disposition (cbbe64773)

CI green; three of four reviewers clean. One Medium, accepted:
- **The roster keeps a previously confirmed entry when its probe fails as long as the PID is alive, so a crashed daemon whose PID was reused stays listed and never goes "gone"** — accepted; a failed probe on a retained entry is checked against process identity (start time or equivalent) before the entry is kept, with the confirm-then-fail-with-live-PID test the reviewer describes.

obra:
## Round 5 fixes

Head: `9fc4397290fed4e2a7b2bdcf063dc8b93c9e69cf` (origin/main merged in first).

Measured true: `refresh` retains a confirmed entry on a failed probe whenever `procAlive` answers, and `sameDaemonIdentity` compares the rendezvous file with itself, so a reused PID stayed listed and `relayEntry` kept resolving it.

What the host can know: `daemonprocess.Controller.Open(Target)` already verifies a live process against a rendezvous entry for force-stop (generation, owner, `serve` argv, start time no later than the entry's, ownership of the session API log; `process.go:verify`), and `hubcore` already imports that package. The roster gains a `procIdentity` probe (`SetProcessIdentity` beside `SetProcessAlive`; exported tri-state `ProcessIdentity`): on unix it opens the target through that verifier (`roster_unix.go`), elsewhere it answers unknown (`roster_other.go`). An entry lacking a state dir or start time answers unknown too.

Rule: in the retained-entry branch, `ProcessNotOwner` clears `alive` so the entry takes the existing crashed path (unlisted by `newHubSourceRegistry`, hence gone to the relay); anything else retains exactly as before.

RED:
- `TestRosterDropsRetainedEntryWhoseProcessIsNoLongerItsDaemon` (confirmed, then probe fails with the PID alive under another identity: the route survived).
- `TestHubRelayAnnouncesGoneWhenARetainedDaemonsPIDIsReused` (real stand-in daemon, real `StatusProber`, entry on this process's PID, identity flipped to not-owner as the stand-in shuts down: 10 s with no resync before, resync after).
- `TestRosterRetainsBusyDaemonThatIsStillItself` pins the transient case for both "owns" and "unknown" (green before and after).

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, plus `GOOS=linux go vet` on the hub since `roster_unix.go` builds there; `go test -race -count=1` for `hubcore` and the hub root (200 s ok); `make lint` (9 PASS).

obra:
## Review round 6 disposition (9fc439729)

CI green. One Medium, accepted:
- **Every controller `Open` error is read as "verified not the owner", so an indeterminate inspection failure on a busy daemon (a descriptor closing mid-inspection) evicts a live session** — accepted; only positive evidence (the process exited, or ownership verifiably fails) is not-owner, and an inconclusive failure answers unknown, which keeps the liveness-only behaviour, with a test for the inconclusive case.

obra:
## Round 6 fixes

Head: `7d5db029beae8250c322a3c0d65e89315024401e` (origin/main already merged; nothing new this round).

Enumerated `Open`'s failures (`process.go`): target validation (invalid/self PID, missing identity, state dir or start time), bind failure (pidfd open / darwin proc_info; `ESRCH` becomes `ErrExited`), and `verify`'s checks (generation empty, generation changed between the two inspections, uid, argv not `serve`, start time zero, start time after the entry's, log descriptor not seen, plus any `inspect` I/O error). Only `ErrExited` was typed.

Fix: `daemonprocess.ErrNotDaemon` marks the four refusals that positively found another process at the PID (owner, command, start time after the entry's, generation change under inspection), through a small wrapper so each refusal keeps its own message; generation-empty and start-time-zero are split off as plain errors, and the unseen log descriptor and inspection errors stay plain. `roster_unix.go` maps `ErrExited`/`ErrNotDaemon` to `ProcessNotOwner` and everything else to `ProcessIdentityUnknown` (liveness-only retention).

RED:
- `TestOpenRefusalsCarryPositiveEvidenceOnly` (`daemonprocess`): five positive cases carry `ErrNotDaemon`, four indeterminate ones carry neither sentinel (red with the sentinel declared but unattached).
- `TestRosterKeepsRetainedEntryWhenOwnershipCannotBeVerified` (`roster_unix_test.go`): a confirmed entry on this process's own PID (the verifier refuses to bind self, an indeterminate failure on a live PID) survived a failed probe only after the fix; before, it read as crashed.
- `TestRosterDropsRetainedEntryWhenAnotherProcessHoldsThePID`: the positive path through the real verifier (a `sleep` child at the entry's PID, a crashed daemon's API log in the state dir) still evicts; the round-5 injected reuse test stays green.

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `GOOS=linux go vet` on the hub, `go test -race -count=1` for `daemonprocess`, `hubcore` and the hub root (190 s ok), `make lint` (9 PASS).

obra:
## Review round 7 disposition (7d5db029b)

CI green; reviewers 2 and 3 clean. One Low, accepted:
- **The empty routing key that marks the hub's daemon-gone resync also describes a malformed notification, so a malformed resync gets promoted into valid per-route re-reads** — accepted; only the hub-generated daemon-gone resync is fanned out, and a malformed one keeps its classification.

obra:
## Round 7 fixes

Head: `d7a4238a9a13a3032f7842fc468710eb37f48ddb` (origin/main merged in first).

Fix: the relay marks its own daemon-gone announcement on the delivery (`RelayDelivery.DaemonGone`, carried by the publish job that enqueues `DaemonGoneResync`; the wire is unchanged), and `publishTarget` stamps per-route identity only on marked deliveries. One correction to the disposition's RED, ruled by the existing test rather than the finding text: `TestHubRelaySharedSessionAliasesDeliverEachNotificationOnce` pins that a wrong-typed ref "is malformed and fans out without thread fallback", so a malformed daemon resync keeps fanning out; what it must not do is get rewritten. The routing classification is untouched.

RED:
- `TestHubRelayDoesNotPromoteAMalformedDaemonResync`: root and child subscribed through one session, the daemon delivers `{"ref":null}` — before, each received `{"ref":"local:…","threadId":"…"}`; now each receives the frame byte-for-byte as sent.
- The three relay-level gone tests now assert the delivery carries the mark (red until the job threaded it); the hub fan-out test sets the mark on its scripted delivery and still reaches every route with its own ref.

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `go test -race -count=1` for `appsource` and the hub root (217 s ok), `make lint` (9 PASS).

https://claude.ai/code/session_011PbmbpbrXA9twoSwTRAf4B

obra:
Correction to the round 7 disposition: an existing test pins that a malformed frame still fans out to every alias route (`TestHubRelaySharedSessionAliasesDeliverEachNotificationOnce`), so the rule is "never promoted", not "never fanned out": each route gets the malformed frame byte-for-byte, and only the hub's own daemon-gone delivery is stamped per route. The implementer ruled from the tests, correctly.

obra:
## Review round 8 disposition (d7a4238a9)

CI green; three of four reviewers clean. One Medium, accepted:
- **Process identity is consulted only when the probe fails, so a reused PID whose endpoint still answers is published as live without an ownership check** — accepted; identity is applied to successful probes too, rejecting a verified non-owner and retaining unknown, with a successful-probe reuse test.

obra:
## Round 8 fixes

Head: `c38252edb92c0ba59210796117459e2e78114f9b` (origin/main merged in first, #1245 included).

**Measured: what a successful probe proves.** The prober dials the entry's endpoint with `SetDaemonAuthorization(header, entry.HubToken)`, and the daemon does enforce it (`server.go:543`), but the token is the hub's own `EVENER_HUB_TOKEN` handed to every daemon the hub spawns (`serve.go:662`), and the roster keys the answer by the session the daemon reports (`liveEntryFromProbe`), never comparing it with the entry. So a successful answer proves "a daemon of this hub listens at that endpoint" and nothing about `entry.PID`. A crashed daemon's stale file whose ephemeral port another daemon of the same hub re-bound would answer, and the file's reused PID would be published as that daemon's. The check is warranted, not only symmetric.

**Fix** (`roster.go`): `procIdentity` runs before publishing a successful probe; `ProcessNotOwner` takes the crashed path (not published; `newHubSourceRegistry` skips crashed), owner and unknown publish as before.

**RED:** `TestRosterDoesNotPublishASuccessfulProbeForAPIDThatIsNotItsDaemon` (probe answers, identity injected not-owner: published before, crashed after; then owner and unknown both publish). `TestRosterDropsEntryWhenAnotherProcessHoldsThePID` reshaped: through the real verifier a `sleep` child at the entry's PID is never published, whether or not the socket answers (it was confirmed first before this round). Every existing roster test stays green; one hub test, `TestHubForkRefusesAClaimItCannotVerify` (synthetic `PID: 1001` with a real state dir and a fake prober), now injects `ProcessIdentityUnknown`, the same rule `SetProcessAlive` states for synthetic PIDs.

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `GOOS=linux go vet` on the hub, `go test -race -count=1` for `hubcore` and the hub root (203 s ok), `make lint` (9 PASS).

https://claude.ai/code/session_011PbmbpbrXA9twoSwTRAf4B

obra:
## Review round 9 disposition (c38252edb)

CI: the web job failed in the skill guard ("timed out waiting for composer cleared; toast: Skill selections aren't supported on…", run 34939261266). That toast means the session's capabilities were withheld, which is exactly what an entry parked as unconfirmed or crashed produces, so this round treats it as a likely regression of round 8's identity check on the Linux runner until the artifacts say otherwise. Two Medium, both accepted:
- **The fallback roster stays local to the source registry, so the rest of the hub still sees no roster** — accepted; the registry hands the roster it made back to the hub configuration.
- **`disowned` is conditioned on a successful probe, so a reused PID whose socket is closed and which is not yet in the roster is retained unconfirmed forever** — accepted; ownership is checked regardless of the probe result, with the reviewer's probe-false-first ordering added to the test.

obra:
## Round 9 fixes

Head: `4e9f893595810ec5f75100b4d7ee28dd2c4110a1` (origin/main merged in first).

**CI failure, what the artifacts showed.** Run 34939261266's `milestones.jsonl` has the composer at `composer-mounted` already reading "Send a follow-up…" and the failure dump's rail rows read "work · failed" for both sessions: the roster had classified both healthy helper daemons as crashed on its first refresh, so the web hydrated them from the past index (`notLoaded`, `skillInput: false`) and the first send was refused with the toast. Cause: round 8 runs `procIdentity` on successful probes, and the guard's daemons are `evener.test -test.run ^TestSkillBrowserDaemonHelper$` (`skill_composer_browser_test.go:627`), so round 6's "not a serve command" refusal, then positive, disowned them. Not Linux-specific: reproduced on this Mac with `make test-web-browser` before the change (same toast), passes after. The linux branch of `inspect` compares the same fields.

**Root-cause fix** (`daemonprocess/process.go`): an argv that is not an `evener serve` invocation is indeterminate (the entry does not record how its daemon was launched; a daemon run inside another binary is still that daemon), and `verify` evaluates the facts that can prove another process (owner, start after the entry, generation change) before the ones that only fail to vouch, so a refusal reports the strongest fact. Kill still refuses on every failure. RED: `TestOpenRefusalsCarryPositiveEvidenceOnly` moves "wrong command"/"missing argv" to the indeterminate set.

**M1**: `newHubSourceRegistry` returns the configuration carrying the fallback roster and `newWebServer` adopts it. RED: `TestHubWithoutRosterSharesItsFallbackRosterWithEveryConsumer` (`web.cfg.Roster` was nil; now it is the roster the local source lists through, and `restartRequiredDaemon` resolves the daemon through it).

**M2**: `disowned := r.procIdentity(e) == ProcessNotOwner` regardless of `res.OK`. `TestRosterDropsEntryWhenAnotherProcessHoldsThePID` now runs "socket closed first" and "answering first" as subtests; the entry's `StartedAt` precedes the child's start so the real verifier's positive fact is the start time. Closed-first was red before (parked unconfirmed).

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `GOOS=linux go vet` on the hub, `go test -race -count=1` for `daemonprocess`, `hubcore` and the hub root (200 s ok), `make lint` (9 PASS), `make test-web-browser` (all six guards PASS, skill guard included).

https://claude.ai/code/session_011PbmbpbrXA9twoSwTRAf4B

obra:
## Review round 10 disposition (4e9f89359)

CI green; two of four reviewers clean. Two Medium, both accepted:
- **`Refresh` runs the identity probe even with a nil prober, so synthetic entries are marked crashed because their PID is absent on the host** — accepted; the nil-prober mode answers unknown.
- **The retained-entry path probes identity a second time instead of reusing the first verdict** — accepted; one verdict per refresh.

obra:
## Round 10 fixes

Head: `1e5020f0b579536d123f6df59cb4d086c7f53559` (origin/main already merged; nothing new this round).

1. **Nil prober.** `Refresh` now takes the identity verdict only when a prober is set; a prober-less roster answers `ProcessIdentityUnknown` without asking the host. RED: `TestRosterWithoutProberDoesNotAskTheHostAboutAPID` (complete entry, absent PID 1001, an identity probe that fails the test if consulted: it was consulted, and the entry read as crashed). The observable is `List()`, since a prober-less roster keys nothing by a probed session id.
2. **One verdict per entry per refresh.** `disowned` is computed once at the top of the loop and reused by the retained-entry branch, which no longer calls `procIdentity` again. RED: `TestRosterAsksTheProcessIdentityOncePerEntryPerRefresh` (a counting probe saw two calls on a failed-probe refresh; now exactly one across a successful and two failed refreshes, route kept).

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `GOOS=linux go vet` on the hub, `go test -race -count=1` for `hubcore` and the hub root (264 s ok), `make lint` (9 PASS).

https://claude.ai/code/session_011PbmbpbrXA9twoSwTRAf4B

obra:
## Review round 11 disposition (1e5020f0b)

CI green. One Medium and one Low, both accepted:
- **The fallback roster refreshes only when the local source lists threads, so direct roster consumers can read it empty or stale** — accepted; the roster is refreshed before it is exposed and every consumer reads it through one refresh-aware path.
- **The disowned branch inside the retained-entry block is unreachable and its comment promises a path that does not exist** — accepted; removed.

obra:
## Round 11 fixes

Head: `723b1f6c904bea47bd6d646b322439973710a8ee` (origin/main merged in first).

1. **M, stale fallback roster.** The one accessor every consumer already goes through is the `Roster` itself, so the refresh-aware wrapper is a mode on it: `RefreshOnRead()` makes each exported snapshot reader (`OwnershipError`, `DaemonOwnershipAbsent`, `List`, `SubagentState`, `HasConfirmedEntry`, `RestartRequiredRootRef`, `Find`, `UnconfirmedEntries`; `IsSubagentActive` delegates) call `refresh` before taking the read lock. `newHubSourceRegistry` puts its fallback in that mode and no longer refreshes before listing itself; `ReadSpawnedThread` (a publish-side coordinator) is untouched. No watcher. RED: `TestHubWithoutRosterListsALateDaemonInNavigationWithoutAThreadListingFirst` — roster-less `newWebServer`, a stand-in daemon's file written after construction, `navigationSnapshotInputs` read `live` as `[]` before and lists the daemon now, with no thread listing having run. Cost in fallback mode: one probe pass per read (`navigationSnapshotInputs` makes three reads); production always wires a watched roster.
2. **L, dead branch.** The `if !disowned { … continue }` guard and the following `alive = false` are deleted; the comment now says the verdict folds into `alive`, so a disowned PID never reaches the retention branch. All roster tests green.

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `GOOS=linux go vet` on the hub, `go test -race -count=1` for `hubcore` and the hub root (220 s ok), `make lint` (9 PASS).

https://claude.ai/code/session_011PbmbpbrXA9twoSwTRAf4B

obra:
## Review round 12 disposition (723b1f6c9)

CI green; two of four reviewers clean. Two Medium, both accepted subject to measurement:
- **Gone is decided by absence from the listing, but an alive-yet-unconfirmed claim is omitted from the listing on purpose, so a transient probe failure mid-transition announces a false ended session** — accepted; an unconfirmed claim is distinguished from an absent one and is never gone.
- **The probe result and the identity check are independent, so a new daemon that rebinds an old endpoint can be published under the old entry while the old PID still verifies** — accepted if the probe response can be bound to the daemon instance it came from (session id or generation in the answer); otherwise the round says what the answer carries and what would be needed.

obra:
## Round 12 fixes

Head: `7ceb22ea60b525a1b6f0da418f25c60543dd0456` (origin/main f614287ca merged in first; A4 is not in main).

1. **M, unconfirmed is not gone.** `LocalDaemonSource` gains `SetUnconfirmedEntries` (wired to `roster.UnconfirmedEntries` by `newHubSourceRegistry`); the relay's connect wraps the not-found as `relayDaemonGoneError` only when no unresolved claim names the ref, so "claim present, ownership unresolved" keeps dialling. The ref-matching predicate is shared (`localEntryNamesRef`). RED: `TestRelaySessionUnconfirmedClaimIsNotGone` (claim moved off the listing onto the unresolved set: a gone resync arrived; now none for 1.5 s, then one once the claim is dropped altogether) and, through a real roster and `StatusProber`, `TestHubRelayKeepsDiallingAnUnconfirmedClaimAndAnnouncesGoneOnlyWhenItIsGone` (the file changes identity under the same PID as the endpoint stops answering, both claims park unresolved: the web got a resync before, gets none now, and gets one once the file is removed).
2. **M, binding the probe answer.** Measured: the answer carries the answering daemon's session id (root thread `SessionID`, and `Evener.InstanceID` is set to the same value, `appwire_runtime.go:2415`); no PID or generation crosses the wire. The rendezvous entry's session id is kept current by the daemon (`rvreg.UpdateSessionID` rewrites `ThreadID`, `SessionID` and `InstanceID` together on a session change, `serve.go:1190`), so the answer can be bound to the daemon instance the entry names by session id. Fix: `StatusProber.Probe` rejects an answer whose root session differs from a session the entry names; an entry naming no session takes the answer's as before. The residual window is one refresh between a session swap and the file rewrite, during which the entry is retained through the existing transient-miss path. RED: `TestStatusProberRejectsAnAnswerForAnotherSession` (a stand-in answering "other" for an entry naming "mine" was published).

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `GOOS=linux go vet` on the hub, `go test -race -count=1` for `appsource`, `hubcore` and the hub root (297 s ok), `make lint` (9 PASS).

obra:
## Round 12 fixes (CI)

Head: `b992782f5f1c6eccbc774bf8ed007467503a3955` (origin/main 4b291876a merged in first).

**Cause.** Both failures were one typecheck: `cmd/evener-hub/app_rpc_test.go:67: assignment mismatch: 1 variable but newHubSourceRegistry returns 2 values` (read from the check-run annotations while the run was still in progress; the job logs were withheld). #1208 landed on main after my previous merge with a new call site in the one-value form the round-9 signature retired; CI builds the PR's merge with current main, so the PR's own tree at 7ceb22ea6 compiled and vetted clean here while the merge did not. Every `static-build` step (`make build`, `make build-go`, `make vet`, `GOOS=windows make vet`), `make lint-golangci`, and a `GOOS=linux` build and vet of every go.work module passed on the PR tree; none of the round-12 code was involved.

**Fix.** The merged call site adopts the two-value form. `fuzz` was still pending in the run's job list (only `fuzz-checks` had reported, green); it will re-run on this push.

Gates, foreground: gofmt (toolchain), `go vet` on the hub, `GOOS=windows make vet`, `make lint-golangci`, `go test -race -count=1` on the hub root over the merged tree (312 s ok), `make lint` (9 PASS, `lint-generated` included).

obra:
## Review round 13 disposition (b992782f5)

CI: the agent race job hit the known #1286 flake and is rerunning. Two of four reviewers clean. One Medium and two Low, all accepted:
- **A stale entry whose PID the hub itself now holds reads as unknown, so the claim is retained forever** — accepted; the hub's own PID is verified not-owner in the roster probe, while the controller keeps its self-PID refusal for force-stop.
- **Refresh-on-read runs a full refresh per snapshot reader, so one navigation request does three passes and can mix three snapshots** — accepted; one refresh per logical read.
- **The start-after-entry check uses the conservative upper bound of the start tick as positive evidence, so a daemon started in the same tick as its entry can be evicted** — accepted; positive evidence uses the lower bound, the upper bound stays for the force-stop refusal.

obra:
## Round 13 fixes

Head: `22059126689dff87bc4830dd52acafbb4028948b` (origin/main merged in first).

1. **M, the hub's own PID.** `roster_unix.go`'s identity probe returns `ProcessNotOwner` for `entry.PID == os.Getpid()` before consulting the verifier, whose self-PID refusal stays for force-stop. RED: `TestRosterTreatsItsOwnPIDAsAnotherProcess` (complete entry on the hub's PID, probe answering: it was published; now crashed) and `TestHubSourceRegistryWithoutRosterKeepsOutAStaleFileOnItsOwnPID` (a stale file on the hub's PID whose endpoint another daemon answers was listed; now not). The round-6 indeterminate test moved off the self-PID (no longer indeterminate) onto a `sleep` child whose state dir has no API log, so the inspection errors before judging. The gone resync for a crashed entry is the chain rounds 3 and 5 pin; a file stale from the hub's start has no relay subscriber to announce to, so the observable is the listing. Five fork tests that use this process's PID as a synthetic daemon now inject `ProcessIdentityUnknown`, as they already inject a fake `DaemonProcesses` controller.
2. **L, one scan per navigation read.** `Roster.Snapshot()` returns `OwnershipError`, `Live` and `Unconfirmed` from one `syncForRead` under one read lock (`List` shares `listLocked`); `navigationSnapshotInputs` reads it. RED: `TestHubNavigationReadsTheRosterInOneScan` (a counting prober on a refresh-on-read roster saw three passes per navigation read; now one). The ownership-error/list pairing follows from one lock and one scan; `refresh` lists the run dir through the OS, not `r.fs`, so a transient `ListStrict` failure between reads cannot be injected, and the one-pass count is the pinned invariant.
3. **L, same-tick start.** `identity` carries `startedAtLower`; Linux `inspect` fills both bounds from `linuxStartOffsetBounds` (ticks/hz and (ticks+1)/hz; `linuxStartOffset` is now the upper bound), Darwin sets both to its exact start. `verify`: lower bound after the entry is `ErrNotDaemon`; upper bound alone refuses with a plain error. RED: `TestOpenSameTickStartRefusesWithoutPositiveEvidence` (fake kernel, lower 199 / upper 201 vs entry 200: was `ErrNotDaemon`), plus `TestLinuxStartOffsetBounds` (linux-only, compiles under `GOOS=linux go vet`). Windows/other unchanged.

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `GOOS=linux go vet` on every go.work module, `go test -race -count=1` for `daemonprocess`, `hubcore` and the hub root (232 s ok), `make lint` (9 PASS).

obra:
## Review round 14 disposition (220591266)

CI: the `tests` job failed; being reproduced against current main. Three of four reviewers clean. Two Medium from one reviewer, both accepted:
- **The fallback roster's changes never reach navigation invalidation or its source revision, so a daemon added or removed after the first snapshot stays cached** — accepted; the fallback roster's content generation joins the navigation source revision (or invalidation), with a test.
- **Relay resolution reads live entries and unconfirmed claims from two snapshots, so a transient miss between them announces gone for a live daemon** — accepted; both come from the single `Snapshot()` the previous round added.

obra:
## Round 14 fixes

Head: `846d3a6069876e9895990773487c883d64c1a51d` (origin/main 4d6f3ef4a merged in after the commits; the merged tree vets clean and every hub package passes).

**CI `tests` failure (run 35008337916), finding.** The job log named `TestForceStopSerializesResumeAndDeletionEntryPoints/delete/{stable,current}` ("delete response … Skipped: resumed live", "delete did not remove saved data") and `TestHubForceStopRejectsRequestsQueuedBeforeRecovery` ("fresh explicit resume did not reach spawn boundary"). Both are roster-less hubs whose fixtures wrote a rendezvous file for `PID: 4242` and assumed a dead daemon; since round 4 a hub with no roster of its own probes the file's PID against the real process table, and on the ubuntu runner 4242 was somebody's live process, so the "dead daemon" read as live. Reproduced in a Linux container: green with 4242 dead, the exact CI failure once a sleeper occupied 4242. Not the #1286 flake, and not round 13's code. Fix (`00ad5780a`): the fixtures name the PID of a process run to completion (`exitedPID`), dead on any host; green in the container with 4242 occupied.

1. **M, navigation follows the fallback roster.** `Roster.Generation()` counts published listing changes (bumped at both publish sites) and is folded into `navigationSourceRevision` (a `Roster` field), so a plain read rebuilds when the roster's listing moved; nothing else is wired and no watcher is built. RED: `TestHubWithoutRosterNavigationSeesALateDaemonOnThePlainNextRead` (two plain `readV2` live reads around a daemon's file appearing: the second returned revision 1 with no entities before; now revision 2 with `local:late`). The live section joins roster entries with saved metas, so the test seeds the meta and only the daemon appears.
2. **M, one scan for "gone".** `LocalDaemonSource.SetClaims` replaces round 12's `SetUnconfirmedEntries`: one `LocalDaemonClaims{Live, Unconfirmed}` from `Roster.Snapshot()` (`newHubSourceRegistry` maps both closures through one `localDaemonEntriesFromRoster`), and the relay's connect resolves the ref against that snapshot's live entries and consults its unresolved claims. RED: `TestRelaySessionDecidesGoneFromOneScan` (the listing read returns nothing while the snapshot is consistent: a live daemon was announced gone; now redialled). `TestLocalDaemonResolveSubscriptionAdmissionSingleSnapshot` kept its contract that an invalid or foreign ref is refused before any inventory read (`acceptRef` runs first). Round 12's test migrated to the new seam.

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `GOOS=linux go vet` on every go.work module, `go test -race -count=1` for `appsource`, `hubcore` and the hub root (296 s ok), `make lint` (9 PASS); on the merged tree, hub vet and `go test ./cmd/evener-hub/...` all ok.

obra:
## /simplify pass (846d3a606)

Four reviewers (reuse, simplification, efficiency, altitude) read the whole diff. Deduplicated, being applied as round 15; behaviour for a hub with a roster stays the same, and three mechanisms that only served a test-only path go:
- **The roster-less fallback goes.** Production always wires a roster; six rounds patched a branch only tests reach. The source registry requires a roster (tests build one), so refresh-on-read, the per-read snapshot refresh, the generation counter and its navigation-revision plumbing are deleted.
- **The roster says when a daemon is gone.** It already diffs sessions per refresh and has a change callback, so it fires a session-gone hook on the crashed path and the hub publishes the resync; the relay's inference from resolver error shapes, the claims plumbing and the two error wrappers go. The epoch-less job, the delivery mark and the per-route stamp stay (real constraints).
- **One identity probe, only when it matters.** The session-id binding on the probe answer identifies the daemon, so identity runs on failed probes only (and for entries naming no session); liveness folds into it; the daemon-process package exposes one verdict function instead of sentinel wrapping. The start-tick lower bound stays.
- **Per-route stamping** marshals the two fields the daemon-gone frame carries.
- **Tests** reuse the existing stand-in daemon, resync and relay-preamble helpers; round citations leave the comments.
Efficiency findings on per-refresh inspection become moot with the identity change.

obra:
## Round 15 (simplify)

Head: `ccae15274b15dee52ea5ef782cf422af273eb951` (origin/main c8ea86abb merged in last; the merged tree vets clean and every hub package passes). Five commits, each green on its own, no trailers.

**Line counts, code / tests, before → after** (non-test `.go` / `_test.go` per package):
- `cmd/evener-hub`: 25870 / 86218 → 25807 / 85838
- `internal/hubcore`: 8128 / 16360 → 8142 / 16411
- `internal/appsource`: 3579 / 8551 → 3460 / 8285
- `internal/daemonprocess`: 728 / 618 → 777 / 601

1. **`e3b1e0c1e` roster-less fallback deleted.** `newHubSourceRegistry(cfg) *Registry` again; no local entries without a roster. `RefreshOnRead`, `syncForRead` (10 sites), `Generation`, `navigationSourceRevision.Roster` gone; `Snapshot` stays as a plain consistent reader. Tests that named a `RunDir` without a roster get one through `rosterOverRunDir` inside `newHubRPCTestServerWithWeb` (production shape, refreshed once); six fallback-only tests deleted; the daemon-exit chain test refreshes its roster the way the watcher would.
2. **`bc4647074` the roster announces gone.** `SetOnSessionGone(func(LiveEntry))` fires from `refresh`'s diff for a session the previous publication held (live, or parked unresolved) that the new one holds as neither: crashed or absent. Never for a claim still unresolved; once per departure (`TestRosterAnnouncesASessionGoneOnceAndOnlyForGood` covers crash, clean exit, unresolved-then-vanished). The hub wires it to `LocalDaemonSource.AnnounceDaemonGone`, which finds the relay session and publishes. Deleted: `relayDaemonGoneError`, `localDaemonThreadNotFoundError`, `LocalDaemonClaims`/`SetClaims`/`currentClaims`/`claimNames`, the `Among` resolver split, recovery's inference, `goneAnnounced`. Kept: the epoch-less job, `RelayDelivery.DaemonGone`, per-route stamping. Four recovery-inference tests deleted; the two relay gone tests, the three hub tests and the e2e run through the hook.
3. **`8abe0504f` one identity probe, only when it matters.** Identity runs only on a failed probe (the answer is bound to `entry.SessionID`); the on-success check and its `disowned` plumbing go. One deviation from the brief, measured: also running it for entries with no session id evicted every synthetic dead-PID fixture the probe had vouched for (five roster tests), and production entries always name their session (`serve.go` writes it), so that clause is out. Liveness folds into the probe (unix default: dead → NotOwner, then `daemonprocess.Identify`); `SetProcessAlive` is an adapter, the 26 direct `procAlive` writes use it. `daemonprocess.Identify(Target) (Identity, error)` replaces the `ErrNotDaemon` wrapper; `judge` is shared with `verify`, which refuses on every failure as before. `hubcore.DaemonTarget` at the five sites; `linuxStartOffset` gone (tests use the bounds); `sync.OnceValues` for HZ.
4. **`de39f99a8`** `stampResyncTarget` marshals `ThreadResyncParams` (passthrough tests dropped); the publish job flows through `publishNotification`/`dispatchToListenerAtEpoch`.
5. **`5e1bb27f3` tests.** `startAliasRecoveryDaemon` + `openSubscribedHub` replace `newDaemonStandIn` and the hand-built entries; `awaitRelayResync` (skip/timeout, sharing `expectResyncParams` with `expectRelayResync`) for the drain loops; `exitedPID` on `fixtureExitProcess`; `awaitHandStartedDaemonListed` for both e2e polls; `openCommittedRelay` for the relay preamble; citations removed; the reused-PID rationale lives once on `ProcessIdentity`. `crashedDaemonStateDir` has a single user left (`roster_unix_test.go`), so nothing to share.

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `GOOS=linux go vet` on every module, `go test -race -count=1` for `daemonprocess`, `hubcore`, `appsource` and the hub root (285 s ok), `make lint` (9 PASS), `make test-web-browser` (six guards PASS, skill guard included).

obra:
## Review round 16 disposition (ccae15274)

CI green. One Medium and three Low, all accepted:
- **An untargeted daemon-gone resync is not parked for a route still being bound, so that route can miss it** — accepted; marked deliveries are retained while any route is pending and replayed once it binds.
- **An error string leaks a Go identifier after the mechanical rename** — accepted; domain wording restored.
- **Two concurrent refreshes can snapshot the same previous roster and both announce the same departure** — accepted; departures are computed against the immediately preceding published generation and the hook fires once.
- **A registry built with only a run directory now lists nothing silently** — accepted; the requirement for a roster is enforced loudly and the remaining harnesses build one.

obra:
## Round 16 fixes

Head: `910f3e4a5d25a586e06d2195f17197a494f36412` (origin/main merged in after the commit; the merged tree vets clean and every hub package passes).

1. **M, routes mid-binding.** A DaemonGone delivery now reaches every current route at once and is retained in `pendingDeliveries` with the set of routes it served while `handle.pendingRoutes != 0`; `processPending` (run on every route change) serves the routes bound since and acknowledges once none is pending (`Proceed` is called on retention, as for parked targeted frames). RED: `TestHubRelayDaemonGoneReachesARouteBoundDuringTheFanOut` — the child's scripted read is held open while the announcement fans out; before, the delivery was acknowledged mid-binding and the child never heard it; now the root hears it at once, the child on binding, and the acknowledgement follows.
2. **L**: `"multiple live daemons claim different current sessions"` restored.
3. **L, overlapping refreshes.** Departures are measured against the publication immediately before, taken under the publish lock (`prevBySess, prevUnconfirmed := r.bySess, r.unconfirmed`), not the unresolved claims snapshotted before probing. RED: `TestRosterOverlappingRefreshesAnnounceADepartureOnce` — a gating prober holds two refreshes open together after a parked claim's file went; before, `[01PARKED 01PARKED]`; now once. The stale-generation guard already covered the listing side.
4. **L, loud without a roster.** I took the brief's first option rather than the panic: a roster-less registry registers no local source, so a local ref fails as "source not found" (pinned by `TestHubSourceRegistryWithoutRosterHasNoLocalSource`), and the two fuzz harnesses build `rosterOverRunDir`. Measured why not the panic on `RunDir` without `Roster`: `TestForceStopSerializesResumeAndDeletionEntryPoints` is a deliberately roster-less hub that sets `RunDir` for rendezvous cleanup; with a roster its force-stop path finds the fixture's dead PID crashed and never opens the fake controller (hang at the `process.entered` wait), so `RunDir` is a hub-wide setting, not a registry input, and the panic would have mis-modelled it.

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `GOOS=linux go vet` on every module, `go test -race -count=1` for `appsource`, `hubcore` and the hub root (250 s ok), `make lint` (9 PASS).

obra:
## Review round 17 disposition (910f3e4a5)

CI: `race-root` failed on this head (being classified). Two of three reviewers clean. Two Medium and one Low, all accepted:
- **The pending-route count resolves when relay state is published, before the subscription is installed, so a departure in that window reaches no subscriber and is acknowledged** — accepted; the route stays pending through subscription commit.
- **A protocol-mismatch probe answers OK, so the identity check is skipped and a stale entry rebound by an incompatible process stays listed** — accepted; mismatch is represented on its own and identity still applies.
- **The gone callback drops the probe-resolved session id, so a legacy entry's announcement is keyed differently from its relay and never arrives** — accepted; the resolved id travels with the announcement.

obra:
## Round 17 fixes

Head: `3840853e7d2837c6130b0e44492a6cd8fd4c83b3` (origin/main merged in after the commit; the merged tree vets clean and every hub package passes).

**CI `race-root` on 910f3e4a5, classification: a real gap in round 16's relay change, not a flake and not a merge conflict.** Run 35027426960 failed `TestHubRelayDaemonGoneReachesARouteBoundDuringTheFanOut` under `-race` ("no thread resync for local:midflight-child within 5s"). The route's pending count resolved when its relay state was published, but the appserver installs the client's subscription only after the read returns, so a departure announced in that window was broadcast to a key with no subscriber and acknowledged. That is the review's first Medium.

1. **M, publish-to-subscribe window.** `publish` no longer resolves the route; it stays pending until the read's `release` (the appserver's capture Commit, which runs after the subscription is installed; the command path finishes its read right after publish, as before). `serveDaemonGone` skips a route with no subscriber yet (`server.SubscriberCount == 0`) and serves it once bound. Measured while fixing: holding the *acknowledgement* until no route was pending deadlocked a resume (`TestConcurrentAliasResumeSpawnsOneCurrentDaemon` hung at the relay read's `<-capture.flushed`): the relay session needs the acknowledgement only to release the capture barrier of a read in flight, and the route being bound *is* that read. A retained announcement is therefore acknowledged as soon as every subscribed route has it; the copy kept for the pending route is the hub's own. RED: `TestHubRelayDaemonGoneAnnouncedBetweenPublishAndSubscribeStillArrives` (the appserver's `SetBeforeSubscriptionGate` holds the child's read between publish and subscribe; before, the child never heard it). Both gone tests pass `-race -count=20`.
2. **M, protocol mismatch.** `ProbeResult.ProtocolMismatch` marks the restart-required answer; the roster asks identity for it as for a failed probe; NotOwner takes the crashed path. RED: `TestRosterListsARestartRequiredEntryOnlyForItsOwnProcess` (mismatch answer + not-owner PID was listed; Owner/Unknown still are). The eleven restart-required tests model incompatible daemons on synthetic PIDs and now assert liveness through `liveClaimRoster` (the `SetProcessAlive` rule), otherwise the default probe evicts them as dead.
3. **L, resolved session id.** `AnnounceDaemonGone(entry, sessionID)`; the hook passes `gone.SessionID`. RED: `TestHubRelayAnnouncesGoneForALegacyEntryThroughItsResolvedSession` (legacy entry: no `SessionID`, no `WorkspaceRef`, `ThreadID` unlike the resolved session; no resync before) and a source-level test on the id.

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `GOOS=linux go vet` on every module, `go test -race -count=1` for `appsource`, `hubcore` and the hub root (219 s ok), `make lint` (9 PASS).

obra:
## Review round 18 disposition (3840853e7)

CI green; no correctness defects across three reviews. One Medium and one Low, both accepted:
- **A confirmed entry that replaces a prior session for the same PID bypasses the departure accounting, so the old session's subscribers never hear it is gone** — accepted; every roster publication goes through the same departure accounting.
- **The hub's own PID is classified unknown for a legacy entry missing its identity fields, so a stale claim stays unconfirmed for good** — accepted; the self-PID check comes first.

obra:
## Round 18 fixes

Head: `e5afd79ef0a0c5e2b5cba35caa165d1280f7f661` (origin/main merged in twice this round, A4 included: the first merge applied cleanly with no conflicts, the second after the commit; the merged tree vets clean and every hub package passes).

1. **M, one place a session leaves the listing.** Both publications go through `publishListingLocked`: it swaps the listing under the lock, diffs statuses and departures against the publication immediately before (`sessionsGone`), and returns a `listingPublication` whose `fire()` runs the hooks after unlock. `refresh` and `publishConfirmedEntry` both use it; the confirmation path's own status-change and `onChange` logic is gone into the shared step. RED: `TestRosterConfirmedReplacementOnAPIDAnnouncesTheOldSessionGone` — a confirmed `01OLD` on PID 1001 replaced through `ReadSpawnedThread` by `01NEW` on the same PID announced nothing before; now `01OLD` once, and the next refresh (with the spawner's new file in place) adds nothing.
2. **L, the hub's own PID first.** `processIdentity` answers NotOwner for the hub's own PID or a dead process before the metadata check. RED: `TestProcessIdentityOfTheHubsOwnPIDIsNotOwnerBeforeAnyOtherCheck` (legacy entry, hub's PID, no state dir: was Unknown). Nine hub tests use this process's PID as the conventional live-daemon stand-in on legacy entries with failing probes (unconfirmed-retention, replaced-ownership, transient-probe deletion tests); they assert liveness through `liveClaimRoster(runDir, prober)`, the restart-required helper generalised to take the prober, per the synthetic-PID rule.

One note on the race gate: the hub root's first `-race` run overran nine minutes; per-test timings showed nothing slow (max 19 s, the spawn fuzz) and the plain suite ran in 146-154 s, and the process table showed two other worktrees running `-race -count=3 ./agent/` and another race suite on this machine. Rerun with the machine otherwise mine: 358 s ok.

Gates, foreground: gofmt (toolchain), `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `GOOS=linux go vet` on every module, `go test -race -count=1` for `hubcore`, `appsource` and the hub root (358 s ok), `make lint` (10 PASS, A4's `lint-package-imports` included).

PR #1328, opened 2026-09-14
fix(notes): decide kindless steering provenance from the mutation record

obra:
# What

Follow-up to the roborev review of #1305: its disputed medium is now fixed from the provenance that review named.

**Finding (roborev, medium).** Kindless ordinary steering beginning with `human updated their whiteboard: ` was still treated as note-origin. A normal `turn/steer` carrying that text could have control characters stripped during live durable reconstruction.

**Fix.** A kindless entry is decided by the persisted mutation method — the durable provenance the review pointed at:

- A recorded `SteeringKind` still decides first.
- Otherwise the journal record's method decides: `notes/human/set` is the note write and normalizes; any other named method (`turn/steer`, a drain, a promote) is ordinary steering and keeps every byte the user typed.
- Only a record that persisted neither field — a shape older than both — still falls back to the write-path text prefix, and its comment now scopes the documented ambiguity to exactly those records.

**Where the method comes from.** `clientMutationRecord.Method` is persisted (`json:"method"`), and `clientSteeringFromSnapshot` rebuilds the runtime queue from those records, so the restart path now passes `steeringOrigin{kind, method}` into `rebuiltSteeringText`. That is the reconstruction path that produced the finding. (`agent/session_client_mutation_queue.go`, `agent/session_notes_rpc.go`.)

**Follow-up (roborev's first review of this PR, medium on `agent/session_notes_rpc.go:940`).** The same ambiguity survived on the model-copy path: `escapeNotesHistoryTurns` rebuilds the copy a resumed request or a forked prefix hands to the model, and it decided from the turn alone, so a kindless ordinary steer whose text imitates the note prefix lost its control bytes on every restart. The copy now reads the same durable provenance. A restored turn carries `ClientMutationID`, and the journal keeps its record -- including after the steer is incorporated, where the method survives while the pending entry and payload are cleared -- so `clientMutationStore.steeringOrigins()` reads the committed journal under `stateMu` (no clone, one read per restore) and each steering turn is decided by its record first, then its own kind, then the write-path shape.

**Follow-up (roborev's second review, medium on `agent/session_notes_rpc.go:908-911` and `agent/session_client_mutation_queue.go:1162`).** Before acceptance became atomic, the note delivery minted its inner steer under the outer note mutation's id plus `/note-steer`, and the `SteeringKind` stamp only arrived later (`a8da149b2 fix(notes): refine mediums ... kind persistence`), so records written in that window keep the steer method and no kind at all. The method-only rule therefore read them as ordinary steering and left legacy note control bytes in the model copy on both load paths. One reader now decides a record's provenance for both: `steeringOriginFromJournal` treats a kindless steer whose id is a legacy inner note steer -- the suffix plus the outer `notes/human/set` record that wrote it -- as note-origin, so the rebuilt pending entry and the restored history copy normalize its text and carry the note kind. The suffix alone proves nothing: a steer that merely ends that way keeps the method's decision, pinned by a test.

**Follow-up (roborev's third and fourth reviews).** A retried note delivery minted `outer + "/note-steer" + "/attempt-N"`, so the correlation parsed that attempt suffix, and the history copy merged the turn's own kind with the record's provenance instead of replacing it. The fourth review then showed the correlation itself was not earning its place: `SetHumanNote` stamps `steering_kind` only when a save *changes* the note, so a no-op save leaves exactly the kindless record the third review's narrowing relied on, and the class the correlation was written for cannot load at all any more -- a pre-cutover journal fails closed with `decode client mutation snapshot: json: unknown field "notes_delivery_pending"`, which is the approved clean cutover. The correlation is gone, and the false-negative half of the finding is fixed: a record that is note-origin by kind, or by the kindless `notes/human/set` method its own rule reads, now decides the merge against an unrelated kind on the turn.

**Remaining limitations, stated honestly.**

- A turn whose record is not in this session's journal -- a forked delegate's inherited prefix, written by the parent's journal -- keeps the kind-then-shape decision. Turns written since `SteeringKind` was stamped carry their own kind, so this covers pre-kind records only.
- A record that kept neither a field (a snapshot built in memory; a loaded journal always carries a validated method) cannot decide, so those turns also keep the shape fallback.

# Evidence

Red first, with the production hunks reverted in place on the same tree:

```
--- FAIL: TestRestoredKindlessOrdinarySteerWithNotePrefixKeepsBytes (0.31s)
    session_notes_provenance_test.go:57: restored ordinary steer = "human updated their whiteboard: [31mred[0m bytes", want it verbatim ("human updated their whiteboard: \x1b[31mred\x1b[0m bytes")
```

The same test on the fixed tree, with the two records that must not change:

```
--- PASS: TestRestoredKindlessOrdinarySteerWithNotePrefixKeepsBytes (0.30s)
--- PASS: TestRebuiltKindlessNoteSteerWithMethodStillNormalizes (0.00s)
--- PASS: TestRebuiltSteerWithoutProvenanceKeepsPrefixFallback (0.00s)
```

- The red test drives the real restart path: it writes a `turn/steer` journal record (method persisted, no kind) whose text imitates the note prefix with ESC sequences, re-reads the store from disk, rebuilds the runtime steering queue, and requires the bytes verbatim.
- `TestRebuiltKindlessNoteSteerWithMethodStillNormalizes` pins that `notes/human/set` without a kind still strips controls.
- `TestRebuiltSteerWithoutProvenanceKeepsPrefixFallback` pins the provenance-free shape: the prefix imitator normalizes, ordinary text keeps its bytes.

Round 2, same method (production hunks reverted in place on this tree):

```
--- FAIL: TestHistoryCopyDecidesKindlessSteerProvenanceFromTheRecord (0.00s)
    --- FAIL: .../ordinary_steer_imitating_the_note_prefix_keeps_its_bytes (0.00s)
    --- FAIL: .../a_record_that_kept_only_the_kind_decides_a_turn_the_shape_would_spare (0.00s)
--- FAIL: TestRestoredHistoryCopyKeepsOrdinarySteerThatImitatesTheNotePrefix (0.20s)
    session_notes_provenance_test.go:195: restored model context = "human updated their whiteboard: [31mred[0m bytes\n", want the ordinary steer verbatim ("human updated their whiteboard: \x1b[31mred\x1b[0m bytes")
```

- The restore-level test writes a real `turn/steer` journal record on disk, restores the session through `RestoreSessionFromMetaWithConfig`, and asserts on `modelBoundText` -- the model-bound rendering, not a raw field.
- The table test pins both directions: the record decides a kindless ordinary steer (keep its bytes) and a kindless note-origin turn whose text lost the write-path shape (strip it), while a turn with no record keeps the shape fallback.

# Gates

On this head (`2e82f632d`):

- `make merge-approval-gate` exit 0 -- all nine lint stages, the module runs (root 611.3s, agent 96.7s, llm 16.9s, auth, envvars, invariant, identifier), the web build and typecheck (1548.5s), then `test-native` and the appwire package qualification. The run log has no failure lines.
- `make vet` exit 0.
- `make lint-evenerfuzz` exit 0.

This lane needed its dependency trees installed before the gate could run those last two stages (`mobile-native`, `appwire-client/typescript`); no source was touched for the repair. Every earlier head carried the same gate.


obra:
**Round 1 finding (medium, `agent/session_notes_rpc.go:940`) — addressed** in b5c6f222f.

You were right: `escapeNotesHistoryTurns` decided from the turn alone, so the same kindless ambiguity the rebuilt path had already dropped survived on the model-copy path. A restored turn carries `ClientMutationID`, and the journal keeps the record for it — after incorporation the method survives while the pending entry and payload are cleared — so the copy now reads that provenance:

- `clientMutationStore.steeringOrigins()` reads the committed journal under `stateMu` (no clone; one read per restore) and maps each id to `{kind, method}`.
- `escapeNotesHistoryTurns(history, origins)` decides by the record first, then the turn's own kind, then the write-path shape. Same order as `rebuiltSteeringText`, so the live, rebuilt, and restored paths now agree.

Red first, production hunks reverted in place on the same tree:

```
--- FAIL: TestHistoryCopyDecidesKindlessSteerProvenanceFromTheRecord (0.00s)
    --- FAIL: TestHistoryCopyDecidesKindlessSteerProvenanceFromTheRecord/ordinary_steer_imitating_the_note_prefix_keeps_its_bytes (0.00s)
    --- FAIL: TestHistoryCopyDecidesKindlessSteerProvenanceFromTheRecord/a_record_that_kept_only_the_kind_decides_a_turn_the_shape_would_spare (0.00s)
--- FAIL: TestRestoredHistoryCopyKeepsOrdinarySteerThatImitatesTheNotePrefix (0.20s)
    session_notes_provenance_test.go:195: restored model context = "human updated their whiteboard: [31mred[0m bytes\n", want the ordinary steer verbatim ("human updated their whiteboard: \x1b[31mred\x1b[0m bytes")
```

The restore-level test drives the real restart: it writes the `turn/steer` record to the store on disk, restores through `RestoreSessionFromMetaWithConfig`, and asserts on `modelBoundText` — the model-bound rendering, not a raw field. The table test pins both directions (the record decides a kindless ordinary steer to keep its bytes, and a kindless note-origin turn whose text lost the write-path shape to strip).

Residual, stated rather than fixed: a forked delegate's inherited prefix belongs to the parent's journal, so those turns keep the kind-then-shape decision — pre-kind records only, since a turn written after `SteeringKind` was stamped carries its own kind. Documented in the PR body.

Gates on `b5c6f222f`: `make merge-approval-gate` exit 0 (nine lint stages, root 518.5s, agent 86.9s, llm 17.0s, web 1267.7s, then native and the appwire qualification), `make vet` exit 0, and `go test ./agent/... -short -count=1` exit 0. The PR body carries the narrowed residual.

obra:
**The medium in this round (legacy `*/note-steer` correlation) — addressed** in `dbee18017`.

Confirmed against the pre-refactor source before fixing: `deliverNotesHumanSetSteer` and `deliverAdoptedNotesSteer` minted `innerID := outerID + "/note-steer"` through `AcceptClientMutationSteer` (so the record's method is `turn/steer`), and `setSteeringKindOnRecord` only arrived in `a8da149b2` — so records written between the note delivery landing and the kind stamp keep no kind, exactly as you described. My method-only rule read those as ordinary steering.

The fix is the correlation you named, in one place: `steeringOriginFromJournal` resolves a record's provenance for both load paths (the rebuilt pending queue and the restored history copy). A recorded kind or method decides as before; a kindless steer whose id is a legacy inner note steer -- the suffix is the note delivery's own derivation, and the outer `notes/human/set` record with that id is the proof -- is note-origin, so its text is normalized and the rebuilt entry carries the note kind. The suffix alone decides nothing: a steer that merely ends `"/note-steer"` without an outer note record keeps the method's decision.

Red first, on both paths you asked for:

```
--- FAIL: TestRebuiltLegacyNoteSteerCorrelatesWithItsOuterRecord (0.00s)
    session_notes_provenance_test.go:276: legacy note steer was not normalized: "human updated their whiteboard: \x1b[31mlegacy\x1b[0m note"
--- FAIL: TestRestoredHistoryCopyCorrelatesLegacyNoteSteerWithItsOuterRecord (0.69s)
    session_notes_provenance_test.go:391: restored model context = "human updated their whiteboard: \x1b[31mlegacy\x1b[0m note\n", want the legacy note controls stripped
```

`TestRebuiltSteerWithNoteSteerSuffixButNoOuterRecordKeepsBytes` pins the counter-case (suffix without the outer record keeps its bytes and its empty kind) and passes on both trees. The restore-level test seeds both records on disk, restores through `RestoreSessionFromMetaWithConfig`, and asserts on `modelBoundText`.

One earlier limitation stands as stated: a fork's inherited prefix still keeps the kind-then-shape decision. What changes is a legacy inner note steer in a *restored* journal -- its outer record decides it now, rather than the text shape.

Gates on `dbee18017`: `make merge-approval-gate` exit 0 -- all nine lint stages including the `lint-evenerfuzz` tag floor, the module runs (root 634.8s, agent 99.4s, llm 26.1s), the web build and typecheck (1974.5s under a loaded host), then `test-native` and the appwire qualification; the run log has no failure lines. `make vet` exit 0. A first gate attempt on this head failed only its web stream (five `Spawn.test.tsx` timeouts with the host at load 45+); the untouched rerun is the green one above.

obra:
**All three findings addressed** in `594eeaba1` (the head of this branch).

**Medium — the retry grammar.** Verified against the pre-atomic source first: `lookupSteerID` returns `innerID + "/attempt-" + N` for N > 1 (the first attempt used the bare id). `legacyNotesSteerOuter` now strips an optional `/attempt-N` (N > 1) before the derived suffix, and the reader serves both load paths.

```
--- FAIL: TestRebuiltLegacyNoteSteerRetryCorrelatesWithItsOuterRecord (0.00s)
    session_notes_provenance_test.go:423: legacy retry note steer was not normalized: "human updated their whiteboard: \x1b[31mretry\x1b[0m note"
```

**Low — the turn's own kind.** `escapeNotesHistoryTurns` now merges both sources (`steeringOriginForTurn`): note-origin evidence from either side decides, so a turn the write path marked note-origin is not demoted by a record that kept no kind, and a record that proves note origin is not overruled by an unrelated kind.

```
--- FAIL: TestHistoryCopyKeepsTheTurnsOwnNoteKindOverAKindlessRecord/kindless_record_leaves_the_turn's_note_kind_standing (0.00s)
    session_notes_provenance_test.go:464: history copy = "human updated their whiteboard: \x1b[31mnote\x1b[0m", want the note-origin control bytes stripped
```

**Low — the forgeable shape.** Addressed by tightening the evidence, not by a marker, and a marker is genuinely unavailable: the record that *did* prove the correlation is `clientMutationRecord.NotesInnerSteerID` (`notes_inner_steer_id`, written by `markNotesSteerAccepted` before the outer success), and the atomic-acceptance refactor in this same PR removed it under the approved clean cutover. `decodeClientMutationSnapshot` runs `DisallowUnknownFields`, so a pre-refactor journal whose outer note record carries that field — exactly the journal that could prove a derived inner steer — fails to decode at all; re-adding the field to read it back would restore compatibility for journals the cutover deliberately stopped reading.

So the correlation now also requires the outer record to have kept no steering kind of its own, the legacy era's shape: since kinds were stamped, the note write carries `steering_kind: human-note` (the same provenance the first fix reads) and the current writer mints no inner steer at all.

```
--- FAIL: TestRebuiltSteerImitatingTheGrammarBesideAModernNoteRecordKeepsBytes (0.00s)
    session_notes_provenance_test.go:499: steer beside a modern note record = "human updated their whiteboard: [31mforged[0m bytes", want it verbatim (...)
```

Remaining exposure, stated rather than papered over: a client steer whose id imitates the retired grammar beside a *legacy* note record (one written in the era whose journals no longer decode). The consequence there is bounded to stripping control bytes from that steer's own text — the safe direction, no widening — and reserving the derived-ID namespace would take a wire-contract change rather than an agent-side rule.

Gates on `62453556c` (the merge of `594eeaba1` into this branch): `make merge-approval-gate` exit 0 -- all nine lint stages including the `lint-evenerfuzz` tag floor, the module runs (root 582.1s, agent 55.4s, llm 18.5s), the web build and typecheck (1534.3s), then `test-native` and the appwire package qualification. No failure lines. `make vet` exit 0.

obra:
**Both findings addressed** in `75c121f6e`.

**The false negative was a real hole, fixed.** `steeringOriginForTurn` promoted the record's provenance only when it carried the note kind, so a kindless `notes/human/set` record was overruled by an unrelated kind on the turn and note controls stayed in the model copy — against the either-side-wins rule the function documents. The record now decides when it is note-origin by kind, or by the kindless note method its own rule reads (`isHumanNoteSteer`).

```
--- FAIL: TestHistoryCopyKeepsARecordsNoteMethodOverAnUnrelatedTurnKind (0.00s)
    session_notes_provenance_test.go:484: history copy = "human updated their whiteboard: \x1b[31mnote\x1b[0m", want the note-origin control bytes stripped
```

**The false positive is closed by removing the correlation**, and your root-cause note is what made me re-examine whether it earns its place. It does not:

- `SetHumanNote` stamps `steering_kind` only when the save **changed** the note, so a kindless `notes/human/set` record is what every *no-op* save leaves behind — which is exactly the era marker my third-round narrowing relied on, so an imitating id beside one was still classified as a legacy inner steer.
- The class the correlation was written for cannot load at all any more. A pre-cutover journal fails closed in the store:

```
decode client mutation snapshot: json: unknown field "notes_delivery_pending"
```

That is the approved clean cutover: the atomic-acceptance refactor removed the note-delivery fields, the snapshot decoder rejects unknown fields, and the current writer mints no derived inner steer at all. No pre-cutover compatibility code, then, for records that cannot be read.

So the suffix constants, `legacyNotesSteerOuter`, and the correlation branch are gone, with its three tests; the `kind`/`method` rules from the first two rounds still decide every record. Two controls stay, pinning what the correlation must not do: an id that spells the old derivation is ordinary steering, and a modern note record classifies nothing but itself.

Gates on `75c121f6e`: `make merge-approval-gate` exit 0 -- all nine lint stages including the `lint-evenerfuzz` tag floor, the module runs (root 376.5s, agent 57.2s, llm 8.1s, auth, envvars, invariant, identifier), the web build and typecheck (742.7s), then `test-native` and the appwire package qualification; no failure lines. `make vet` exit 0. The matches, steering, mutation, and provenance families pass in short mode.

obra:
**The two lows are fixed** in `0bd6acafe`; **the medium is declined on evidence**.

**Low — stale comments.** Corrected in `steeringOrigins` and `clientSteeringFromSnapshot`: only the record stored under the client mutation id decides, and an id's spelling is not evidence of anything.

**Low — the derived kind.** `steeringOrigin.steeringKind` now derives `human-note` from a kindless `notes/human/set` method, so a rebuilt entry — and the `evener/steering/injected` event derived from it — carries the note label instead of an empty kind. The existing normalization test pins the kind as well as the text now:

```
--- FAIL: TestRebuiltKindlessNoteSteerWithMethodStillNormalizes (0.00s)
    session_notes_provenance_test.go:98: kindless note-origin steer kind = "", want "human-note"
```

**Medium — legacy note-steer restoration: declined, and here is why.** Those records cannot be loaded at all any more, so they cannot reach model context, the transcript, or a steering event:

```
decode client mutation snapshot: json: unknown field "notes_delivery_pending"
```

That is the approved clean cutover in action: the atomic-acceptance refactor removed the note-delivery fields, `decodeClientMutationSnapshot` rejects unknown fields, and the current writer mints no derived inner steer. The inner record was only ever written *after* its outer note record, and that outer record is one of the ones carrying a removed field — so any journal that could contain the legacy inner steer fails closed at the store, before any classification runs. The one reachable sliver is a mixed journal: a legacy inner record beside an outer record rewritten post-cutover by a retry, which would classify that inner steer as ordinary steering.

That trade is Jesse's call and it is on the record: no pre-cutover compatibility code for records that cannot be read. If the project ever wants the association back, the marker that made it collision-safe was `clientMutationRecord.NotesInnerSteerID` (removed by the same refactor), and a successor would have to survive the outer record being rewritten.

Gates on 0bd6acafe: GATE_LINE

obra:
**The low is fixed** in `84d3be7ff` — comment-only, and checked as such: every changed line in the diff is a comment line, so nothing in the compiled code moved.

The `escapeNotesHistoryTurns` block now sits directly above its own declaration, with the `humanNoteSteerPrefix` paragraph left on the constant, so `go doc` and IDE hover no longer render the constant with the history-escaping description. Wording is unchanged.

Verification for a comment-only change: the notes, steering, and provenance families pass in short mode, `make vet` exits 0 on the merged head, and CI runs the full suite on the push.

`84d3be7ff` holds nothing else: the head above it (`0bd6acafe`) is the round-5 fix set you already reviewed.

obra:
**The medium is fixed** in `994a4bd09` (merged as `24b4fd8a9`).

You are right that the child's journal is not evidence about the parent's turns, and that ids are only required to be non-empty — so a child mutation reusing a parent's client mutation id could decide whether an inherited steer kept or lost its bytes. The fix takes the first option you offered: the fork path no longer reads any journal at all.

`escapeInheritedHistory(inherited)` takes only the inherited transcript entries and calls the escaper with no provenance, so an inherited prefix is decided by its own kinds and the write-path text shape alone, by construction — there is no child state in reach to consult. `NewSession`'s fork path uses it, and the test pins the contract on a prefix whose two steering turns share one client mutation id: the note-shaped text is normalized, and the ordinary steer keeps every byte.

Gates on `24b4fd8a9`: `make merge-approval-gate` exit 0 -- all nine lint stages including the `lint-evenerfuzz` tag floor, the module runs (root 459.7s, agent 57.9s, llm 13.0s, auth, envvars, invariant, identifier), the web build and typecheck (1079.2s), then `test-native` and the appwire package qualification; no failure lines. `make vet` exit 0, and the notes/steering/provenance families pass in short mode.

obra:
**The low is fixed** in `77bb908a0` (merged as `210af34f2`) — comment-only, and checked as such: every changed line in the diff is a comment line.

`escapeInheritedHistory` and its own comment now sit ahead of the `initEnvContext` block, so the env-context contract ("called once, before any turn is processed") is adjacent to `initEnvContext` again and the helper keeps the comment that describes it. Same class of mistake as the previous round's, caught the same way.

Verification for a comment-only change: the notes, steering, and provenance families pass in short mode and `make vet` exits 0 on the merged head, with CI running the full suite on the push.

Gates on `210af34f2`: `make merge-approval-gate` exit 0 -- all nine lint stages including the `lint-evenerfuzz` tag floor, the module runs (llm 13.6s and the rest), the web build and typecheck (1056.5s), then `test-native` and the appwire package qualification; no failure lines.

obra:
**The medium is fixed** in `439cb0cd8`.

You are right, and it is the same invariant from the earlier round in the path I missed: the fork path escapes its prefix with no journal in reach, but a *restored* fork hands the child's origins to its whole history — and a forked transcript begins with the parent's inherited prefix, ids included — so an id collision could classify an inherited turn.

`escapeHistoryWithSessionProvenance` splits the restored history at `meta.DivergenceTurn` (whose own documentation in `fork.go` is "the child copies the first divergenceTurn-1 entries"), so everything before it is escaped by its own kinds and the write-path text shape alone, and only the turns this session created consult its journal. A compacted history that no longer starts at the beginning keeps the kind-then-shape rule for its first turns — the honest fallback you named.

Red first, by reverting the call site in place:

```
--- FAIL: TestRestoredForkEscapesItsInheritedPrefixWithoutTheChildJournal (0.38s)
    session_notes_provenance_test.go:436: inherited prefix was decided by the child journal: "run the inherited tests[31mwith red lines[0m\n..."
```

With the bound the same test passes: the inherited turn keeps its bytes, and the session's own note-origin turn is still normalized from its record.

Gates on `439cb0cd8`: `make merge-approval-gate` exit 0 -- all nine lint stages including the `lint-evenerfuzz` tag floor, the module runs (root 566.8s, agent 73.9s, llm 14.0s, auth, envvars, invariant, identifier), the web build and typecheck (1406.0s), then `test-native` and the appwire package qualification; no failure lines. `make vet` exit 0.

obra:
**The medium is fixed** in `224438dd5` (the fix commit is `1022c3b4a`; the head adds the doc-comment restoration the gate's lint stage asked for). You are right, and the split review 3 checked was the non-compacted case: `DivergenceTurn` indexes the full transcript, while `ResumeHistory` returns the last compaction turn and everything after it, so `divergenceTurn-1` read against the shorter slice put the bound past the resumed start and the child's own post-compaction turns were classed as inherited.

`ResumeHistory`'s window rule now lives in `retainedFrom`, which both readers share, and the restore path shifts the bound by that anchor before escaping (`divergenceTurn -= retainedFrom(transcriptEntries)` when the history comes from the transcript). Only turns before the shifted bound take nil origins; the child's own turns after it consult its journal again, including kindless steering that merely resembles the note prefix.

Red first: with the raw bound the child's own note-origin turn kept its controls through a compacted resume; with the shift the same test passes.

Gates green on `224438dd5`: `make merge-approval-gate`, `make vet`, `make lint-evenerfuzz`.

obra:
**Both findings fixed** in `2e82f632d`.

**Medium — the fork boundary counts repair insertions.** `ResumeHistory` splices a synthetic `TurnToolResults` for every orphaned tool call (`history_repair.go`), but the boundary was mapped from transcript coordinates by subtracting `retainedFrom` alone, so the split landed one turn early per insertion at or before it, handing an inherited steering turn to the child's journal, where a colliding child `notes/human/set` record stripped bytes the parent wrote. `resumeHistoryIndexed` is `ResumeHistory` plus each insertion's index in post-repair coordinates (through `repairOrphanedToolResultsIndexed`, the reader `turnHistoryBaseline` already shifts by), and the restore path shifts the bound by every insertion at or before it, exactly as `history_repair.go` does for the in-flight boundary. Red first: `TestRestoredForkBoundaryCountsRepairInsertions` failed with the inherited steer's bytes stripped (`keep [31mthese[0m parent bytes` was rewritten) and passes with the shift.

**Low — no review-round metadata in comments.** The three parenthetical round references (`agent/session_init.go` twice, `agent/session_notes.go` once) are gone; the provenance, id-reuse, and compaction-shift rationale they qualified stays.

Gates green on `2e82f632d`: `make merge-approval-gate` (nine lint stages PASS; root 611.31s, agent 96.68s, llm 16.89s, web 1548.45s), `make vet` (silent, all modules), `make lint-evenerfuzz`.

obra:
Roborev's review of `2e82f632d` returned **no issues**, and all fifteen CI checks on that head are green (`tests`, `race-root`, `race-modules` agent/nonagent, `web`, `native`, `static`, `static-build`, `fuzz`, `fuzz-checks`, `lint-golangci`, `lint-repository`, `api-package`). The local gates were green on the same tree before the push (`make merge-approval-gate`, `make vet`, `make lint-evenerfuzz`). Nothing further outstanding on this branch.

Issue #710, opened 2026-08-31
live e2e: an accepted steer never appears as a steering item in the transcript readback


obra:
Found by the first live run of `TestE2E_LiveModelStopAndSteerNeedNoTurnID` (kimi-for-coding/k3, 2026-08-31): the interrupt half passes end-to-end, but the steer half fails — the steer is accepted (mutation receipt `applied`, correct turn id) yet the steered text never appears as a steering item when the session is read back:

    "EVENER-E2E-LIVE-STEER-TEXT" never appeared as a steering item in the transcript:
    the steer was accepted but a user reading the session back never sees it

Ownership evidence: the assertion is byte-identical on main and on the registry branch (#704 — only the config template differs); steering-item persistence lives in `internal/apptranscript`, untouched by that branch's 251 commits. The test skips by default and needs live credentials, so this has likely been latent for a while.

Repro: `EVENER_LIVE_TESTS=1 EVENER_TEST_PROVIDER=<instance> EVENER_TEST_MODEL=<model> <key env> go test ./cmd/evener-hub/ -run TestE2E_LiveModelStopAndSteerNeedNoTurnID -v` (on #704's schema; main's template needs its own provider vocabulary).

PR #1329, opened 2026-09-15
fix(agent): a drained steering leg runs before the input ends (the session reported idle mid-turn)

### LINKED ISSUE #1342: agent: a steering carrier claim that never ran survives a restart as a stranded ActiveTurnID
Noticed while fixing #1329 review round 1 (the Stop-in-the-claim-window finding); not fixed there because it is a crash, not a Stop, and predates that PR on the wake path.

`claimSteeringCarrierTurn` (`agent/session_client_mutation_queue.go`) publishes `ActiveTurnID = <the steer's reserved id>` while the steer itself stays `accepted`; `popSteeringHead` is what moves it to `claimed`, inside the carrier turn's acceptance. Both callers (`ProcessPendingUserInput`'s wake and, since #1329, the drain ladder) hand the claim back on every in-process exit. A process death between the claim and the pop leaves `ActiveTurnID` naming the steer's id in the durable snapshot.

At load, `forgetRunningTurnNoOneOwns` (`agent/session_client_mutation_persist.go`) keeps an `ActiveTurnID` that any pending execution names, and the accepted steer names it. `restoreDurableClientMutationQueues` only returns `claimed` entries, so the steer stays `accepted` with the slot still held: `steeringCarrierRailOpen` refuses (slot occupied), `claimableSteeringCarrierTurnID` never runs it, and `AcceptClientMutationStart`'s "turn is already active" precondition refuses every later `turn/start` for the life of the session, across restarts. `TestSteeringOnlyCarrierStrandedClaimWouldRefuseTheNextTurnStart` names the consequence for the in-process case.

**Fix direction:** at load, an `ActiveTurnID` owned only by a steering entry still `accepted` is a carrier claim that never ran, and clearing it is safe: the steer is accepted, so the attach-time wake (`SetPendingUserInputWakeFunc`) claims and runs it again. RED test: persist a snapshot with `ActiveTurnID` = an accepted steer's reserved id, restore, assert `turn/start` is accepted and the steer is delivered by the wake.

### LINKED ISSUE #1389: agent: a steer whose transcript append landed but whose incorporation write failed stays 'accepted' on the wire until the next Stop or restore
Noticed by the #1329 round-8 implementer after the reshape to "accepted until recorded; the transcript decides" (PR #1329, agent/session_steering_reconcile.go). If a steer's transcript append lands but the client-mutation store's incorporation write fails, the durable state stays `accepted`, so clients see a pending steer that has in fact been delivered, until the next Stop finalisation or a restart marks it from the transcript. No loss or duplication; a visibility lag only.

Wanted: after a failed incorporation write, retry the store write on the next successful store mutation (or on the next wake) so the wire catches up without a Stop or restart. Also noted: `carrierSteerStillQueued` and `wakeHasClaimableWork` each clone a store snapshot per carrier acceptance / per wake; a `stateMu` accessor like `steeringHeld()` would avoid the clone.

### LINKED ISSUE #1453: agent: a daemon-initiated notification turn outside the input ladder still drains a parked steer as a passenger
From #1329 round 13 (head 85d187cf1). The new `steeringParked` state (set by any carrier attempt that fails short of the model, cleared by the next wake sender, attach, or a successful attempt) stops the input ladder from running a notification/goal rung while a steer is parked, so an autonomous turn cannot drain the steer under a turn id other than the client's reserved receipt. Residual, by design and stated in the PR body: a daemon-initiated notification turn that starts OUTSIDE the ladder still drains a parked steer as a passenger of that turn. Decide whether that path should also honour `steeringParked` (stand down or skip the steering drain) and pin it with a test; measure whether the mismatch between the receipt's turn id and the passenger turn id is observable to clients (steering journal / provenance).

obra:
Closes the last open half of #1308 (the web skill guard's `waitForTurnIdle` barrier passing while a drain's steering leg was still undispatched).

Closes #1342 (a carrier claim that never ran survived a restart as a stranded `ActiveTurnID`; folded in by review round 4).

Closes #1389 (a steer whose transcript append landed but whose incorporation write failed made the carrier fail a turn the model should have run; review round 10).

Closes #1453 (a notification turn the daemon starts on its own wake drained a parked steer as a passenger under its own turn id; review round 14).

## Which layer

The daemon. The web's `steerVisible` derives from `status.type === "active" && activeTurnId` (`appwire-client/typescript/submitRouting.ts` `isTurnActive`), and both of those go false exactly when the daemon emits `EventSessionEnd` (projector closes the turn and publishes the thread status carried by the event). The daemon emitted that between the two legs of a drain-as-steer turn, so the web and the barrier were reading the daemon correctly.

## Measurement

CI run 34900212155 (artifacts: `helper-alpha-requests.jsonl`, `milestones.jsonl`, failure dump): leg 1 (seq 3, held) completed 21:49:12.805 with a terminal `communicate`; `drain-released` 21:49:13.050; the barrier passed and the hold was armed; the drained text (seq 4, `PROSE_QUEUE_14c second pass`) reached the provider at 21:49:13.716, `held=true`. The captured leg never completed, the turn never ended, the next submit routed to the client-side queue.

Code path: a `turn/drainAsSteer` accepted while the turn's last model call is in flight adds pending user steering; that call returns a terminal `communicate`, which by design leaves client steering for a carrier turn (`agent/session_tool_round.go` "A terminal communicate result owns this input's boundary"). The drain ladder in `processInputKindWithProvenance` then had no rung for it: it settled the input (`armAwaitingAtSettle`, `EventSessionEnd` with `State: WireState()`) and returned, and only the acceptance-time wake (`wakeForPendingSteering` → `ProcessPendingUserInput` → `claimSteeringCarrierTurn`) later opened the carrier as a separate input. `sessionWorkPending` (which is what `WireState` upgrades idle on) does not count pending user steering either, so the emitted state was idle/awaiting. Any client reading the status offered a fresh turn on a session that still owed the user a leg.

## RED test

`agent/session_drain_as_steer_turn_boundary_test.go` `TestDrainAsSteerKeepsTheTurnOpenUntilItsSteeringLegRuns`: a scripted adapter holds leg 1; the test queues an entry and accepts `turn/drainAsSteer` mid-leg, releases, and records every turn opening and session end from the lossless event stream. Before the fix: `ProcessInput` returned after 1 model call with `session_end:input_complete:awaiting` and the drained text undispatched. After: two model calls in the same input, exactly one session end, after `turn_started:<the drain receipt's turnId>`.

## Fix

`agent/session_lifecycle.go`: the drain ladder gets a `runSteeringCarrier` rung. When nothing is queued (a queued message drains steering itself at `acceptUserInput`) and user steering is pending, the ladder claims the carrier with the same `claimSteeringCarrierTurn` the wake uses (refuses a held rail after a Stop, or an occupied slot) and runs it inline as `EntrySteeringCarrier`, right behind queued input and ahead of notifications, awaiting or not (steering is the user speaking, same as a queued reply). The claim is handed back on every exit the way `ProcessPendingUserInput` does. The wake still fires at acceptance and now finds nothing left to carry.

Wire shape: the carrier's `EventTurnStarted` now lands while leg 1 is still open in the projector, which is the existing, tested `openTurn` path (`TestTurnBoundaryCompletesThePreviousTurn`): `turn/completed(leg 1)` + `turn/started(carrier)` + thread status `active`, and no idle in between. No projector code changed; the TUI suite passes unchanged.

Tests updated to the new contract (they pinned the wake-runs-it-later shape): `TestCommunicate_TerminalClientSteeringRunsAfterTheTerminalResult`, `TestCommunicate_ClientSteeringBeforeResultProjectsValidNextRequest`, `server` `TestLiveSteeringOwnerIdentityReplays/delayed`; the drain-ladder fuzz oracle and coverage table gained the rung. The steering-owner assertion (the carrier owns the steer under its own reserved id) is unchanged and still passes.

## Gates

- gofmt (toolchain binary) clean; `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz` on `agent` and `server`; `make lint` green (naming, gofmt, evenerfuzz, eval, internal, golangci, generated, fuzz-registry, secret-scan).
- `agent` module: `go test -race -count=3 ./...` run as name shards of the `agent` package (all ok) plus `TestLoadSessionJobActivityTree*` as three separate `-count=1` runs (each ~230 s under race); the other 41 packages `-race -count=3` ok.
- root module: `-race -count=3` over every package except `cmd/evener-hub` ok, except `internal/selfupdate`'s six `TestInstallDirs*`, which fail identically on main `a361254ab` on macOS (TMPDIR `/var` vs `/private/var`, tracked as #1201); `cmd/evener-hub` `-race -count=1` three times ok.
- Web gates not run: no web sources touched. Skill guard (`TestSkillComposerBrowser`) run once locally against this daemon: result in a comment below.

## Round 7 fixes

Head `96d957336` (origin/main `c8ea86abb` merged; no `agent/` changes came in). Two commits: `17b3e3a87` (rules B and L, the pop-claim and incorporation re-arms) and `bed6aa0b6` (the in-flight rule names the owning turn; the settle's no-op write). Every new test was RED against `7d0e603ab` first, run from a detached worktree at that commit with the new test file dropped in; tests in `agent/session_drain_as_steer_turn_boundary_test.go` and `agent/session_steering_reconcile_test.go`.

**High, any in-flight turn preserves any claimed steer.** The store cannot name the claimant: a pending steer's `TurnID` is the id it reserved at acceptance, which names its carrier but not the passenger turn that drains it (`TestRetryLeavesAnInFlightAppendAlone` is such a passenger), so a literal `pending.TurnID == running turn` rule would re-materialize a passenger mid-append. `popSteeringHead` now records, in memory, the active turn it claimed each steer under (`steeringClaimants`), and the table's row 5 preserves a claimed steer only when that claimant is the active turn. Row 5' (any other claim under a turn in flight, including a claim from before a restart) falls to rows 6-8 as if no turn were in flight: the retry timer firing during an unrelated turn returns it, re-arms, and the running turn's next drain carries it. Rule B (every input's settle runs the table with no turn in flight) stays for a claim the timer did not catch mid-turn. Tests: `TestRetryDuringAnUnrelatedTurnReturnsAStaleClaim` (RED: `claimed` after the timer fired during an unrelated turn; GREEN: accepted, in memory, re-armed), `TestUnrelatedTurnEndsWithAClaimedSteerReconciled` (the settle path), and six row 5/5' cases in `TestReconcileClientSteeringTable`.

**Medium, retry cleared before incorporation.** `consumeSteeringMessage` clears the budget only after `finalizeIncorporatedSteering` succeeds; a refused write runs the table as the calling turn (row 6 finalizes a recorded steer), and a store still refusing re-arms from there. `TestSteeringCarrierRecordedSteerSurvivesAFailedIncorporationWrite` now asserts the steer is incorporated before the input ends (RED: `claimed`).

**Medium, refused pop claim arms nothing.** `popSteeringHead`'s refused claimed-mark latches rule L and calls `scheduleSteeringCarrierRetry`; the benign races (empty queue, held rail, occupied slot) still arm nothing. `TestRefusedPopClaimArmsTheRetry` (RED: attempts 0).

**Low, later tool rounds drain again.** Rule L: one per-input latch, `steeringDrainRefused`, set by a failed append or a refused pop claim and consulted by `injectDrainedSteering` itself (so the tool round's drain in `session_tool_round.go` and every `session_lifecycle.go` site share it) and by the drain ladder's carrier claim; cleared when an input starts; the ladder's `carrierUndelivered` local is gone. `TestFailedDrainIsNotRetriedByALaterToolRoundOfTheSameInput` (RED: attempts 2).

**Also.** Rule B's settle committed a snapshot on every input, changed or not (`mutate` always saves). The session wrapper returns `errSteeringReconcileNoop` from the mutate when the table changed nothing, so nothing is written (`TestReconcileWritesTheStoreOnlyWhenTheTableChangedSomething`, RED: 1 write). The table doc carries rows 5 and 5', the claimant paragraph, and the rewritten rule B.

Gates: gofmt (toolchain binary) clean on both modules; `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz` on `agent` and root; `make lint` green; `agent` package `-race -count=3` as nine name shards (A-C, D, E-K, L, M-R, Se, other S, T-Z, Fuzz/Example seeds) all ok, `TestLoadSessionJobActivityTree*` ×3 `-count=1` ok; the other agent-module packages `-race -count=1` ok; root `server`, `cmd/evener`, `internal/appprojector`, `cmd/evener-tui`, `cmd/evener-hub` `-race -count=1` ok. Post-merge build and root vet ok.

Noticed, not changed: rule B's settle passes `inFlight=false` unconditionally, and rule S would release a carrier claim made concurrently by a bare `ProcessPendingUserInput` caller; the daemon's input loop is serial (`cmd/evener/serve.go` `processMessage`), so no such caller exists there.

## Round 8 (simplify)

Head `fbee03ae7` (origin/main `1988c5b5a` merged; its one `agent/` change is the watch-projection formatting in #1374). Four commits, one per step of the /simplify pass, each green on the whole `agent` package before the next began.

**1. Accepted until recorded** (`478d5c9b0`). The durable `claimed` state for steers is gone. `popSteeringHead` takes the steer out of the in-memory queue and writes nothing; the store keeps it `accepted` until `finalizeIncorporatedSteering` records that the transcript append landed. Between the two the steer is in flight (`steeringInFlight`, under `s.mu`), which is what keeps `reflectDurableClientSteering` from re-queueing it; a failed append ends the window and the reflect puts the still-accepted steer back at its place in the order. The transcript is the only record of delivery: restore finalizes every steer whose steering turn it loaded, matching by client mutation id and reserved turn id (RoboRev's Medium at `session_client_mutation_queue.go:1293-1305` -- an accepted steer with a recorded turn was re-queued and delivered twice; RED at `96d957336` with a synthetic accepted-plus-recorded store, "restore left a steer the transcript holds pending", GREEN after; `TestRestoreFinalizesARecordedSteerAndReleasesItsCarrierSlot` pins it through the real path). A steer whose append landed but whose incorporation write the store refused stays in flight marked recorded, and the next Stop's finalization marks it from that set, never a history scan (`TestStopFinalizationMarksARecordedSteerAndReleasesTheHold` drives the daemon's order: Stop accepted, the cancelled turn's completion finalizes the fence). `reconcileClientSteering` is now: finalize a recorded steer, drop a dangling order entry, park accepted steers on a Stop, release a hold naming nothing (rule H); rows 5-8, the claimant map, the claim-return write, rule B's table run at every settle, the release retry's re-run, `recordedClientSteering`, `steeringRecorded`, `turnInFlight` and the session-level wrapper are deleted. `forgetRunningTurnNoOneOwns` keeps a slot only for a pending turn/start or turn/queue, so a dead carrier's claim is released by the one load-time sweep (`TestLoadReleasesASlotOnlyAUserTurnOwns`). The table test is two row tables, no `t.Run` blocks.

**2. No paced retry** (`d6a425495`). `scheduleSteeringCarrierRetry`, `clearSteeringCarrierRetry`, `steeringCarrierRetry`, `steeringRetryMu`, `steeringCarrierRetryLimit`, the `steeringDrainRefused` latch and its three methods, and the four re-arm sites are deleted. A steering append the transcript refuses fails the announced carrier turn and ends the input with that error, the queued-input path's policy; the steer stays accepted in the store and the queue, which keeps the session from resting (`hasRunnableUserSteering`) until the next wake or the user's next run carries it (`TestSteeringCarrierAppendFailureFailsTheTurnAndTheNextRunCarriesTheSteer`: one carrier opening, one turn failure, `turn_failed` session end, one append attempt within the input, then `ProcessPendingUserInput` carries it). A carrier claim the store refuses stands down with a warning for the same reason.

**3. One rung** (`fc9a3e47e`). `EntrySteeringCarrier` is gone. `claimSteeringCarrierInput` publishes the head steer's reserved id into the slot and returns a `queuedInput` with that identity and no content; the drain ladder takes it in the queued rung (`drainInputs.QueuedCarrier`; `runSteeringCarrier`, `SteeringCarrierTurnID` and the context key are deleted), `ProcessPendingUserInput` runs it through the queue branch, and the pending-question gate and `delegateEntryRequiresReport` no longer name a kind. `acceptUserInput` recognizes the identity's `SteeringCarrier` mark and drains without recording, counting or announcing a user turn. Numbers, production only: 164 insertions / 181 deletions (comment rewrites inflate the insertions); what still knows the carrier is `acceptUserInput`'s branch, the stand-down sentinel `processOneInput` maps to "nothing to do", and the two release points (`releaseSteeringCarrierClaim`) -- a queued message's slot is cleared by its completion or kept for restore. The entry-kind audit, the ladder fuzz oracle and the coverage table lost their carrier rows; `TestServerAppWireSteeringCarrierTurnStartedConsumesPendingIdentity` is unchanged (the carrier still opens with `EventTurnStarted` under the reserved id).

**4. Tests** (`80aa7edfc`). `agent/session_drain_as_steer_turn_boundary_test.go`: 1643 lines at the round-7 head, 797 now (51% cut), one test per surviving behaviour: the #1308 boundary; failed append fails the turn and the next run carries the steer; Stop in the claim window parks; Stop mid-append parks (no wake fires); restore releases a dead carrier's claim; restore finalizes a recorded steer and releases its slot; completion under a fence marks a recorded steer and releases the hold (rule H); the release retry wakes the steer behind a stale slot; the drain loop stops at the first failed append. The local `steerRefusingFS` and the environment failure FS are replaced by the `clientMutationTranscriptAppend` seam (it refuses one steer's turn and writes every other turn for real); the per-test configs by `newTestSessionForEnvctx` with `steeringCarrierClaimed` set on the session; the one real-timer negative wait is gone, and the release retry runs on `agenttest.NewFakeClockAt`. `heldLegAdapter` stays: `stopHoldAdapter`'s first call returns only on cancellation and `closeRaceAdapter` holds every call, so neither can release the first leg and answer the second.

**Reviewer addenda.** The O(history) `recordedClientSteering()` sample on every completion is gone with the function; the recorded question is the in-flight set (`recordedSteeringAwaitingMark`, O(k) under `s.mu`, sampled before the mutate). No test waits on a real timer for a negative assertion. `turnInFlight` and its snapshot clone are deleted; the remaining snapshot clones on the carrier path are one per carrier acceptance (`carrierSteerStillQueued`) and one per wake (`wakeHasClaimableWork`), not per settle. The claim/claimant publication race and the retry budget after finalize have nothing left to race or reset; `TestStopFinalizationMarksARecordedSteerAndReleasesTheHold` pins that a turn's completion under a fence consults the recorded set.

Gates: gofmt (toolchain) clean on both modules; `go vet`, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz` on `agent` and root; `make lint` green (nine gates); `agent` package `-race -count=3` as nine name shards (A-C, D, E-K, L, M-R, Se, other S, T-Z, Fuzz/Example seeds) all ok; `TestLoadSessionJobActivityTree*` ×3 `-count=1` ok; the other agent-module packages, `server`, `internal/appprojector`, `cmd/evener`, `cmd/evener-tui`, `cmd/evener-hub` `-race -count=1` ok; ladder fuzz seeds and coverage table under `-tags evenerfuzz` ok; post-merge build, vet and the steering/watch tests ok.

## Round 9 fixes

Head `f4f53d34a` (origin/main `5ccd6e40a` merged, no `agent/` changes). Two commits, every fix RED first against `fbee03ae7`.

**High, a legacy `claimed` steer is neither queued nor finalized at restore** (`176def90a`, approved by Jesse). A store written before this series can hold `ExecutionState "claimed"` for a steer (the old `popSteeringHead` wrote it), and every reader now asks for `accepted`. Load normalizes such a steer to accepted (pending execution and journal) before the steering table runs, so a recorded one is finalized like any other and an unrecorded one is queued exactly once. This is a compatibility rule for stores written before this change, and the only one. `TestRestoreQueuesALegacyClaimedSteerOnce` (RED: `claimed` in both records after restore, not queued).

**High, a carrier claim takes a steer in flight** (`623a7ba04`). `claimableSteeringCarrier` read the store alone; a steer whose append landed but whose incorporation write failed is accepted there and absent from the queue, so a wake claimed it, the carrier drained a different steer under its id and failed again. Eligibility now excludes the session's in-flight steers, sampled under `s.mu` just before the durable claim (`steeringInFlightSample`; the serializer never waits on `s.mu`, and the set changes only on the calling goroutine); `wakeHasClaimableWork` samples the same way. `TestCarrierClaimSkipsASteerInFlight` (RED: the in-flight steer was claimed; GREEN: refused with only it in the store, and a steer queued behind it is the one claimed).

**Medium, a failed append arms no wake.** `consumeSteeringMessage`'s failure branch calls `wakeForPendingSteering` after `steeringLanded`, the wake the steer's acceptance armed; `injectDrainedSteering`'s comment no longer names the deleted retry. `TestSteeringCarrierAppendFailureFailsTheTurnAndTheNextRunCarriesTheSteer` drains the acceptance-time wake at the claim and gates the next run on the wake the failure arms (RED: no wake).

**Low.** The ladder fuzz oracle's doc counts six drain actions.

Gates: gofmt (toolchain) clean on both modules; `go vet`, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz` on `agent` and root; `make lint` green (nine gates); `agent` package `-race -count=3` as nine name shards all ok; `TestLoadSessionJobActivityTree*` ×3 `-count=1` ok (one contended run hit go test's 10-minute wall and was rerun alone); the other agent-module packages, `server`, `internal/appprojector`, `cmd/evener`, `cmd/evener-tui`, `cmd/evener-hub` `-race -count=1` ok (the hub likewise rerun alone after a contended timeout); post-merge build, vet and steering tests ok.

## Round 10 fixes

Head `1ed6db783` (origin/main `597d4e09b` merged: 47 commits, the `agent/` ones in skills and watch projections; the merged tree builds, vets and passes the whole `agent` package). One commit, `864d6076b`; every fix RED first against `f4f53d34a`.

**Medium, a failed skill selection leaves the steer in flight.** `recordFailedSteeringSelection` retired the steer from the store but never ended its in-flight window, so every later carrier claim skipped it and everything queued behind it (eligibility walks the order); when the failure record itself could not be appended, the steer was pushed back by hand, the drain reported success and popped the same steer again. It now reports whether the failure landed, routes both outcomes through `steeringLanded` (a client steer re-queues from the store at its place in the order; daemon steering goes back to the head), and the drain stops when the record did not land. `TestFailedSteeringSelectionLandsTheSteer` (RED: still in flight) and `TestFailedSteeringSelectionWhoseRecordFailsStopsTheDrain` (RED: two failure records attempted in one drain).

**Medium, a recorded steer whose store mark failed (#1389).** The carrier's post-drain question could not tell "back in the queue" from "recorded, unmarked", so it failed a turn whose steer is in the transcript and sent no model request, and nothing ever wrote the mark. `carrierSteerUndelivered` reads the in-flight set's landed flag: recorded, the turn proceeds; back in the queue, it fails as before. `reconcileRecordedSteering` writes the missing mark through the steering table (rule 0) at the input's settle and at the next wake, on top of the Stop and restore paths that already did, and ends the in-flight window; nothing runs and nothing is written when no steer is waiting. `TestCarrierProceedsWhenOnlyTheIncorporationWriteFails` (RED: `ProcessInput` failed with "stays queued", one model request; GREEN: two requests, the second carrying the steer, and the store marked incorporated after the settle). The failure-record path's own store write, when refused, is handled the same way (`markSteeringRecorded`).

**Low, two release owners.** `ProcessPendingUserInput`'s deferred release of the carrier's claim is gone; the run owns it: the drain loop after each turn, and the entry gate's closed-session refusal, the one exit before the loop.

**Low.** `TestStopOnASteeringCarrierTurnArmsNoHoldForItsOwnSteer` is `TestStopOnACarrierParksItsUnrecordedSteerAndTheHoldNamesIt`; the three comments naming the removed `claimSteeringCarrierTurn` name `claimSteeringCarrierInput`.

Gates: gofmt (toolchain) clean on both modules; `go vet`, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz` on `agent` and root; `make lint` green (ten gates); `agent` package `-race -count=3` as nine name shards, the heavy ones solo, all ok; `TestLoadSessionJobActivityTree*` ×3 `-count=1` solo ok; the other agent-module packages, `server`, `internal/appprojector`, `cmd/evener`, `cmd/evener-tui`, `cmd/evener-hub` `-race -count=1` ok; post-merge build, vet, steering tests and the whole `agent` package ok.

## Round 11 fixes

Head `f8dda6db9` (origin/main `f7e0be0e2` merged; its `agent/` change is #1399's replay fixes; the merged tree builds, vets and passes the whole `agent` package). One commit, `41847ec67`. Each finding was measured first with a test through the public surface and fault-injected writes; the traces:

**M1, a failed steering append re-wakes the carrier (reproduced).** `TestPersistentSteeringAppendFailureRunsOneCarrierPerExternalWake`: the steer's append is refused for good by the transcript seam (a zero-byte ENOSPC: the rollback succeeds and the writer is not poisoned -- the failure class that re-wakes; a partial line poisons, and the poison gate refuses every later turn, so that class cannot loop). A daemon in miniature answers each wake. Trace at `1ed6db783`: five wake-driven runs, six append attempts, bounded only by the test's limit -- the failure path's `wakeForPendingSteering` (round 9) fed the daemon a new wake per failed carrier. Fix, structural: the failure path arms nothing; the steer goes back to the queue parked runnable (`hasRunnableUserSteering` keeps the session from resting), and the next external wake -- attach, or any accepted client mutation -- carries it. No paced retry. GREEN: zero wake-driven runs, one append attempt, then the attach-time wake carries the steer once the disk has room. `TestSteeringCarrierAppendFailureFailsTheTurnAndTheNextRunCarriesTheSteer` now asserts no wake from the failure and drives the next run from the external wake.

**M2, a failed failure record leaves the steer pending with no wake (measured; parks by design, and it exposed a fifth defect).** `TestFailedSelectionRecordParksTheSteerUntilTheNextExternalWake`: a steer naming an unknown skill, with the transcript seam refusing its failure turn. Trace: the acceptance wake runs the carrier, the record is refused, the steer is back in the queue out of the in-flight set, the turn fails ("stays queued"), and no wake is armed -- the same contract M1 now applies to every failed append. The next external wake records the failure and retires the steer as `failed`. The trace's final phase showed the fifth defect: once the failure was recorded and the steer retired, the carrier read "gone from the store" as "delivered" and made a model request carrying nothing. `carrierSteerOutcome` now answers delivered / undelivered / failed, and a failed selection stands the turn down (the failure was announced on that turn by the record's error event).

**M3, a failed mark retry strands a recorded steer (refuted).** `TestFailedRecordedSteerMarkRetriesAtTheNextWake`: a recorded steer whose incorporation write was refused, then a wake whose mark retry the store refuses too. Trace at `1ed6db783`, unchanged code: the steer stays in flight marked recorded and out of the queue; the next wake (the acceptance of another steer) retries the mark before its runnable check (`wakeForPendingSteering` runs `reconcileRecordedSteering` first) and marks it incorporated; it is never re-queued. The test pins that shape.

**M4, a recorded selection failure is marked `incorporated` (reproduced).** `TestRecordedSelectionFailureIsMarkedFailedNotIncorporated` and `TestRestoreMarksARecordedSelectionFailureFailed`: the failure turn is recorded and the retirement write refused. Trace at `1ed6db783`: after the next wake, and after a restore, the journal read `incorporated`. Fix: the in-flight set carries the terminal state the transcript entry stands for (`"incorporated"` for a steering turn, `"failed"` for a failure turn), the steering table's rule 0 retires a recorded steer to that state (`finalizeSteeringInSnapshot` takes it), and restore derives it from the turn kind. The table test gained the `failed` row.

Gates: gofmt (toolchain) clean on both modules; `go vet`, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz` on `agent` and root; `make lint` green (ten gates; one staticcheck finding on the new condition fixed and folded in); `agent` package `-race -count=3` as nine name shards ok; `TestLoadSessionJobActivityTree*` ×3 `-count=1` solo ok; the other agent-module packages, `server`, `internal/appprojector`, `cmd/evener`, `cmd/evener-tui` `-race -count=1` ok, `cmd/evener-hub` `-race -count=1` solo ok; post-merge build, vet, steering tests and the whole `agent` package ok.

## Round 12 fixes

Head `f6f169425` (origin/main `75a909292` merged; its `agent/` change is #1313's thinking-text billing; the merged tree builds, vets and passes the whole `agent` package). One commit, `f50f0835c`; both findings RED first against `f8dda6db9`.

**Medium, a delivered steer behind a failed selection is never read.** `injectDrainedSteering` rightly kept draining after the head steer was retired as a skill-selection failure, so the steer behind it was appended and marked incorporated; the carrier then read the head's failure, stood down, and the input settled idle with a steering turn in the transcript that no model request read -- and `hasRunnableUserSteering` false, so nothing woke for it. `TestCarrierProceedsWhenALaterSteerWasDeliveredBehindAFailedSelection` (RED: zero provider requests, the second steer incorporated). `consumeSteeringMessage` now reports what it did (`steeringAppendFailed`, `steeringDelivered`, `steeringRetired`), `injectDrainedSteering` reports whether the pass delivered anything, and the carrier stands down on a failed head only when nothing was delivered; otherwise the turn proceeds and its request reads the delivered steer. The `carrierSteerFailed` comment now says so: "a model request would carry nothing" holds only when the drain delivered nothing.

**Medium, a refused carrier claim strands the steer.** `TestRefusedCarrierClaimParksTheSteerForTheNextExternalWake`: the store refuses the claim's `ActiveTurnID` write; the wake stands down with the steer accepted, queued and no slot taken; no wake is armed by the failure; the acceptance of an unrelated steer is the next external wake, and its run carries both. The code already behaved this way, so the test lands as the pin and the claim's failure comment states the contract: the same parking as a failed append -- runnable, nothing arms a wake of its own, the next external wake or accepted mutation claims it, no paced retry.

Gates: gofmt (toolchain) clean on both modules; `go vet`, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz` on `agent` and root; `make lint` green (ten gates); `agent` package `-race -count=3` as nine name shards ok; `TestLoadSessionJobActivityTree*` ×3 `-count=1` solo ok; the other agent-module packages, `server`, `internal/appprojector`, `cmd/evener`, `cmd/evener-tui` `-race -count=1` ok, `cmd/evener-hub` `-race -count=1` solo ok; post-merge build, vet, steering tests and the whole `agent` package ok.

## Round 13 fixes

Head `85d187cf1` (origin/main merged twice this round, `b946b4ee8` at the start and `85d187cf1` at the end; no `agent/` changes in either). One commit, `7bcba03ef`.

**M1, accepted: a recorded steer compacted out of history was queued again at restore.** Restore derived a steer's terminal state from the resumed history, which starts at the last compaction, while its incorporated map came from the full transcript index, so a match read as unrecorded. `TestRestoreFinalizesARecordedSteerCompactedOutOfHistory` (RED with the fix reverted: `pending=true queued=true after restore`): a steer recorded with its incorporation write refused, a `TurnSummary` written after it (the marker `retainedFrom` anchors on; `Compact` keeps a history this short verbatim, so the test writes the marker itself and asserts the resumed history no longer holds the steer), restore. Fix: `clientMutationTranscriptItems` carries `Steering`, restore derives the outcome from the full index (`Failure` → `failed`, `Steering` → `incorporated`) as well as the resumed history, and a match under the reserved turn never returns `""`.

**M2, measured, reproduced.** `TestRefusedLadderClaimRunsNoAutonomousTurnBehindIt`: the ladder's claim write is refused (the second store write after the leg, after `popQueueHead`'s), a job notification is pending, and the notification turn's model makes a tool call. Trace at `b946b4ee8`: store writes 4, carrier claimed false, model calls 3, and the transcript held `STEERING cm-drain-mid-leg owner=""` between leg 1's tool results and the notification turn's assistant rounds -- the notification turn's tool round drained the steer under a turn id that was not the receipt's `turn_m1`.

**M3, measured, reproduced.** `TestBufferedAcceptanceWakeDoesNotRetryAFailedInlineCarrier`: the steer's acceptance mid-turn arms the pending-input wake, the daemon holds it while the input runs, the inline carrier fails. Trace at `b946b4ee8`: buffered wakes answered 2 (the queue's acceptance wake and the drain's), append attempts 3 -- the buffered wakes each spent an attempt on the failed steer.

**Fix for M2 and M3, one parked state.** `Session.steeringParked` is set by a failed attempt short of the model (the append refused, the failure record refused, the carrier claim's write refused) and cleared by the next wake sender (`wakePendingUserInput`: an accepted client mutation, a store write that landed; the attach-time wake) or by a successful attempt (`consumeSteeringMessage` delivered or retired). While parked, `ProcessPendingUserInput` stands down on a wake with nothing queued (the buffered acceptance wake), and the drain ladder's new `SteeringParked` input takes `goIdle` ahead of the notification and goal rungs while user steering is still pending -- a follow-up or a queued message still runs. The ladder fuzz oracle and coverage table carry the rung. No paced retry. GREEN: M2 makes one model call, the steer stays queued, the notification stays pending for the daemon's own wake, and the external wake delivers the steer under `turn_m1`; M3 answers the buffered wakes with one append attempt total, and the external wake carries the steer. Residual, by design: a notification turn the daemon starts on its own wake (outside the ladder) still drains a parked steer as a passenger, the way any running turn drains steering accepted mid-turn.

Gates: gofmt (toolchain) clean on both modules; `go vet`, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz` on `agent` and root; `make lint` green (ten gates); `agent` package `-race -count=3` as nine name shards ok; `TestLoadSessionJobActivityTree*` ×3 `-count=1` ok (one pass paired with the hub, both well under the wall); the other agent-module packages, `server`, `internal/appprojector`, `cmd/evener`, `cmd/evener-tui`, `cmd/evener-hub` `-race -count=1` ok; ladder oracle and coverage table under `-tags evenerfuzz` ok; post-merge build, vet and steering tests ok.

## Round 14 fixes

Head `06505753b` (origin/main `e04a8d50c` merged; its `agent/` changes are the transcript writer's A1 durability work, #1412 and #1456; the merged tree builds, vets and passes the whole `agent` package). One commit, `77ef1765e`; every scenario RED first against `04284ca85`.

**Medium, accepted: the continuation paths ignored the park.** `selectDrainNextAction` honoured `steeringParked`, but the wrapper ran a continuation deferred at an earlier tail whenever `!awaiting && haveDeferredCont`, and `settleGoalOnIdle` kicked an active goal at the settle; both accept by draining steering, so the parked steer left under the continuation's fresh turn id. Measured: `TestRefusedLadderClaimKicksNoGoalContinuation` (active goal, steer pending, the claim's write refused at the ladder tail; RED: one goal kick after the refused claim) and `TestRefusedLadderClaimHoldsADeferredContinuation` (the goal gate folds at tail 1, a notification turn interleaves, the steer arrives during it, tail 2's claim is refused; RED: one goal kick -- in this construction the wrapper's deferred block did not fire before the settle's kick did, so the kick is the observed defect and the deferred block is gated by the same predicate). Fix: `steeringParkedNow` is the one predicate (parked and the steer still queued), the deferred continuation is held rather than dropped (the goal stays active in the store, and the settle after the next external wake's carrier kicks it -- asserted), and `settleGoalOnIdle` stands down while parked.

**#1453, the round-13 residual, closed by the same gate.** The daemon's own notification wake enters through `ProcessInputKind(EntryNotification)`, outside any ladder, so the ladder's `SteeringParked` rung could not reach it; the entry gate can. It now refuses a daemon-started notification, continuation or delegate-attention turn while user steering is parked, the way it refuses them while a question is pending: the wake source keeps its work and the ladder behind the external wake's carrier delivers it. `TestParkedSteeringRefusesADaemonNotificationTurn` (RED: the notification turn delivered the parked steer under `""`; GREEN: zero model calls, the notification kept, then the external wake delivers the steer under its receipt's id and its ladder delivers the notification).

Test infrastructure: a pre-claim seam (`testConfig.steeringCarrierClaiming`) lets a test refuse exactly the claim's store write instead of counting writes; `heldLegAdapter.holdCall` can hold a later model call, so a steer can arrive during a notification turn.

Gates: gofmt (toolchain) clean on both modules; `go vet`, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz` on `agent` and root; `make lint` green (ten gates); `agent` package `-race -count=3` as nine name shards ok; `TestLoadSessionJobActivityTree*` ×3 `-count=1` ok (pass 1 paired with the hub at 214 s / 219 s, passes 2 and 3 solo); the other agent-module packages, `server`, `internal/appprojector`, `cmd/evener`, `cmd/evener-tui`, `cmd/evener-hub` `-race -count=1` ok; ladder oracle and coverage table under `-tags evenerfuzz` ok (the table did not change this round); post-merge build, vet, steering tests and the whole `agent` package ok.

## Round 15 fixes

Head `838eda9c1` (origin/main merged twice this round, `4d7b4d5e8` at the start and `838eda9c1` at the end; no `agent/` changes in either). One commit, `0a852ba8c`; RED first against `4d7b4d5e8`.

**Medium, accepted: the ladder re-claimed a parked steer inline.** The carrier rung's comment reasoned that a carrier whose append failed ends the input, so the rung could never meet a steer this input had already tried. That held for carriers only: when a running user turn's own boundary drain failed the append (`acceptUserInput`'s drain, whose result the turn ignores), the steer was parked and the turn completed normally; the ladder then found an empty queue head, pending user steering and an open rail, and `claimSteeringCarrierInput` did not consult the park -- a second attempt within one input, the one the park exists to prevent. `TestUserTurnWhoseDrainFailedDoesNotReclaimTheSteerInline` (RED: `ProcessInput` failed with the carrier's "stays queued", the claim seam fired, two append attempts): a passenger steer whose append the transcript seam refuses, drained at the user turn's acceptance. Fix: the ladder's claim is gated on `steeringParkedNow`, the predicate every other autonomous path already asks, and the comment states the real contract -- one attempt per external wake, whichever turn made it. GREEN: one model call, one append attempt, no claim, the steer parked and queued, and the external wake delivers it under its receipt's turn id. `QueuedCarrier` stays false when the claim is not made, so the selector's decision follows from inputs it already had; the ladder table is unchanged and the evenerfuzz oracle and coverage table pass as they were.

Gates: gofmt (toolchain) clean on both modules; `go vet`, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz` on `agent` and root; `make lint` green (ten gates); `agent` package `-race -count=3` as nine name shards ok; `TestLoadSessionJobActivityTree*` ×3 `-count=1` ok (pass 1 paired with the hub at 235 s / 234 s, passes 2 and 3 paired at 232 s / 221 s); the other agent-module packages, `server`, `internal/appprojector`, `cmd/evener`, `cmd/evener-tui`, `cmd/evener-hub` `-race -count=1` ok; ladder oracle and coverage table under `-tags evenerfuzz` ok; post-merge build, vet and steering tests ok.


obra:
Skill guard run locally against this daemon (`go test -tags browserguard -count=1 -run '^TestSkillComposerBrowser$' ./cmd/evener-hub/`, dist built from this branch): `ok primeradiant.com/evener/cmd/evener-hub 21.192s`. The drain scenario's barrier now waits on a turn that does not end until its steering leg has run.

Residual noticed while tracing the delivery path, not fixed here: #1330 (the web's `busy` can blink false between `turn/completed` and `turn/started` of back-to-back turns because the hub relays them as separate WebSocket messages; the guard's barrier reads that button).

obra:
## Review round 1 disposition (1268cdc4e)

CI green. Two Medium, both accepted:
- **A Stop accepted between the carrier claim and its acceptance treats the claimed steer as unrelated pending work, holds it, and finalization deletes its durable record without clearing the order or the hold** — accepted; the claim marks the mutation atomically, and cancellation or finalization either consumes it or restores durable and in-memory state consistently, with a test in the claim-to-acceptance window.
- **A non-poisoning append failure restores the steer as accepted, the carrier reports success, and the drain tail re-claims it every iteration: an unbounded loop of model turns** — accepted; the carrier reports whether it consumed the steer, and the tail stops re-claiming on failure, with a forced-failure test proving one turn, not many.

obra:
## Round 1 fixes

Head `67050199d` (origin/main `d21a136ef` merged as `ec6e3d42c`). Both Medium findings fixed; one commit, `agent/session_drain_as_steer_turn_boundary_test.go` carries both RED tests.

**Stop in the claim-to-acceptance window** (`TestStopInTheCarrierClaimWindowParksTheSteerForTheNextUserRun`, RED: the steer was gone from the durable store, `pending=false`). Measured against the code rather than the finding text: the strand was in `finalizeClientMutationInterrupt`, which retires every pending execution naming the cancelled turn — the untaken steer included, since the carrier claim publishes `ActiveTurnID` = the steer's reserved id while the steer is still `accepted`. Marking the steer `claimed` at the carrier claim, as suggested, would not do: `clientSteeringFromSnapshot` materializes only `accepted` entries into the in-memory queue, so any reflect between the claim and `popSteeringHead` (a concurrent steer accept) would drop the carrier's own steer from memory. The fix keeps the durable states as they are and makes finalization consistent with them: a steer still `accepted` under the cancelled turn's id is a carrier claim that never ran, so it is left in place. The Stop already parks it (`snapshotHasPendingUserSteering` counts an accepted steer), the ladder's release clears the slot, and the user's next run carries it — the same treatment a queued message a Stop returns gets (wms7). The test drives the real ladder through a new `testConfig.steeringCarrierClaimed` seam with serve.go's cancel-then-wait-then-finalize ordering, then proves `turn/start` delivers the parked steer.

**Non-poisoning append failure** (`TestSteeringCarrierAppendFailureRunsOnceAndReturnsTheSteer`, RED: `ProcessInput` never returned — the re-claim loop, verbatim). `acceptSteeringCarrierInput` now checks the durable store after `injectDrainedSteering`: if its steer is still pending, it fails the turn it already announced (`emitTurnFailure`, so the projection closes the turn) and stands down without a model request; the ladder skips the carrier rung for a steer the carrier just returned. The retry belongs to the next wake, and the test proves that wake delivers. The settle now counts runnable (unheld) user steering as autonomy — `hasRunnableUserSteering`, extracted from the two wake-registration sites that already encoded it — so the input ends `idle`, not `awaiting` with a wake pending. Fault: zero-byte durable write, rollback succeeds, writer not poisoned (`environmentSyncFailureFS`), armed from the same seam.

Gates: gofmt clean; `go vet` ×3 flavors on `agent`; `make lint` green; `agent` package `-race -count=3` as name shards plus loaders ×3 `-count=1`, all ok; other 41 agent-module packages ×3 ok; root: `server`, `cmd/evener`, `internal/appprojector`, `cmd/evener-tui` ×3 ok, `cmd/evener-hub` ×1 ok (`internal/selfupdate` `TestInstallDirs*` still #1201 on macOS, unchanged).

Residual filed, not fixed here: #1342 — a crash (not a Stop) between the carrier claim and acceptance leaves `ActiveTurnID` naming an accepted steer across a restart, and `forgetRunningTurnNoOneOwns` keeps it; pre-existing on the wake path.

obra:
## Review round 2 disposition (67050199d)

CI green. Two Medium and one Low, all accepted:
- **The undelivered-carrier latch resets on any later iteration of the same input, so the next iteration re-claims the steer it just failed** — accepted; the failure is latched for the whole input.
- **A claimed-but-recorded steer (transcript append succeeded, the store finalize failed) reads as undelivered, so the carrier fails a turn whose steer is already in the transcript** — accepted; only an entry restored to `accepted` counts as undelivered, with a fault-injection test on the finalize write.
- **`wakeForPendingSteering` open-codes the predicate `hasRunnableUserSteering` now owns** — accepted; one source of truth.

obra:
## Round 2 fixes

Head `112fdf541` (origin/main `34e20d433` merged as `2a385df22`; main's changes to `run.mjs` merged untouched). One commit; both RED tests in `agent/session_drain_as_steer_turn_boundary_test.go`, each measured RED against the pre-fix sources before the fix went back in.

**M1, guard reset per iteration** (`TestSteeringCarrierFailureIsNotRetriedByALaterTurnOfTheSameInput`, RED: two carrier openings, two turn failures within one input). A disk that refuses the steer's bytes for good (`steerRefusingFS`, zero-byte write, rollback succeeds, writer usable) plus a message queued from the claim seam: the queued turn runs after the failed carrier, cannot carry the steer either, and the next iteration re-claimed it. `carrierUndelivered` is now declared outside the loop and latched (`carrierUndelivered = carrierUndelivered || …`); the claim is gated on it for the whole input. GREEN: one carrier opening, one failure, the queued turn runs, the steer stays accepted for the next wake, input ends idle.

**M2, claimed read as undelivered** (`TestSteeringCarrierRecordedSteerSurvivesAFailedIncorporationWrite`, RED: `provider requests = 1`, the carrier failed its turn and made no model call over a steer already in the transcript). `clientMutationFaults.BeforeEffectSnapshotRename` fails the second store write after the claim — `popSteeringHead`'s claimed-mark lands, the transcript append lands, `finalizeIncorporatedSteering`'s write fails — and the test asserts the entry reads `claimed` so the fault provably hit that write. `steeringCarrierUndelivered` now answers true only for `ExecutionState == "accepted"` (what `returnClaimedSteering` restores). GREEN: the carrier's model request carries the steer, no turn failure, clean completion.

**L, duplicated predicate**: `wakeForPendingSteering` calls `hasRunnableUserSteering` (the function lives in `agent/session_client_mutation_queue.go`, not `session_queue.go` as cited).

Gates: gofmt clean; `go vet` ×3 flavors on `agent`, vet on root; `make lint` green; `agent` package `-race -count=3` as name shards plus `TestLoadSessionJobActivityTree*` ×3 `-count=1`, all ok; other 41 agent-module packages ×3 ok; root `server`, `cmd/evener`, `internal/appprojector`, `cmd/evener-tui` ×3 ok, `cmd/evener-hub` ×1 ok; drain-ladder fuzz seeds ok.

obra:
## Review round 3 disposition (112fdf541)

CI green; two of three reviewers clean. Two Medium from one reviewer, both accepted subject to the RED test reproducing them:
- **When the carrier's append fails and the steer returns to accepted, the input ends without arming a wake, so an idle session leaves the steer queued until something unrelated happens** — accepted; a wake is re-armed for the returned steer with backoff so a persistent storage failure does not spin.
- **A Stop racing the carrier between claim and a failed append records the hold as released, so the preserved steer is retried immediately instead of parked** — accepted; preserving an undelivered carrier steer during interrupt finalization sets the hold until the next explicit user run.

obra:
## Round 3 fixes

Head `bc4cced6a` (origin/main `59b2b2280` merged as `2d1f34aa0`). Both findings reproduced RED before the fix; one commit, tests in `agent/session_drain_as_steer_turn_boundary_test.go`.

**M1, no wake after a failed carrier** (`TestSteeringCarrierAppendFailureReArmsTheWake`). Measured: the wake-registration sites that call `hasRunnableUserSteering` are `registerDaemonWake` (attach time) and `SetPendingUserInputWakeFunc` (install time); the drain's own acceptance-time wake is `submitGuaranteedWake`'s one parked send, consumed by the input that failed or by the single retry the daemon runs behind it. After a one-shot append failure with a counting wake installed, no wake arrived in 5 s (RED). Fix: `acceptSteeringCarrierInput`'s failure path calls `scheduleSteeringCarrierRetry`, which reuses the release retry's primitive (`runningTurnReleaseRetryState`, `sclock().AfterFunc`, `jobNotificationRetryInitialDelay` doubling to `jobNotificationRetryMaxDelay`) and is bounded by `steeringCarrierRetryLimit` (8, like the release retry; each attempt is a failed carrier turn, so the loop ends in a warning that the steer waits for the user's next message). The timer fires `wakeForPendingSteering`, which stands down for a parked or since-carried steer; a carrier that records its steer clears the backoff. GREEN: the wake arrives after the first 250 ms backoff and its retry delivers. The new mutex is classified forbidden for envelope sampling (`TestEverySamplingRelevantMutexIsClassified`).

**M2, Stop while the carrier is appending** (`TestStopWhileTheCarrierIsAppendingParksTheReturnedSteer`). `steerRefusingFS` gained an `onRefuse` seam that runs serve.go's cancel-wait-finalize Stop on the refusing write, i.e. after `popSteeringHead` marked the steer claimed and before its append failed. RED: `SteeringHeld=false` after finalization and the Stop's closing `wakeForPendingSteering` fired for the returned steer. Fix: `finalizeClientMutationInterrupt` sets `SteeringHeld = true` when it preserves an undelivered carrier steer. GREEN: parked, no wake within a second (the backoff timer fires and stands down), `claimSteeringCarrierTurn` refuses, and the user's next `turn/start` carries it.

Gates: gofmt clean; `go vet` ×3 flavors on `agent`, vet on root; `make lint` green; `agent` package `-race -count=3` as name shards plus `TestLoadSessionJobActivityTree*` ×3 `-count=1`, all ok; other 41 agent-module packages ×3 ok; root `server`, `cmd/evener`, `internal/appprojector`, `cmd/evener-tui` ×3 ok, `cmd/evener-hub` ×1 ok; drain-ladder fuzz seeds ok.

https://claude.ai/code/session_011PbmbpbrXA9twoSwTRAf4B

obra:
## Review round 4 disposition (bc4cced6a)

CI green; reviewers 2 and 4 clean. One High, one Medium, one Low, all accepted:
- **High, the carrier durably sets the active turn before the steer is marked claimed, so a crash in between leaves a steer owning the active turn across restore and blocks every later start** — accepted and folded in: this is #1342, which the implementer filed as pre-existing on the wake path, and this PR makes the carrier the common path. Restore recovers a carrier-owned active turn (or the claim becomes atomic). #1342 closes with this PR.
- **Medium, a failed claim return after a failed append is ignored, so the carrier runs a model request without the steer while the mutation goes invisible to wakes** — accepted; the failure propagates and the carrier stays in recovery until delivery is durably confirmed.
- **Low, the retry budget resets only when a carrier delivers, so a steer drained by another path leaves a spent budget for the next episode** — accepted; the budget resets on any recorded steer delivery.

obra:
## Round 4 fixes

Head `412d63de9` (origin/main `ec9f1a5a6` merged as `3b4ed97b4`). One commit; all three findings reproduced RED first, tests in `agent/session_drain_as_steer_turn_boundary_test.go`. PR body now carries `Closes #1342`.

**HIGH, carrier claim survives a restart** (`TestRestoreReleasesACarrierClaimThatNeverRan`, RED: `ActiveTurnID = "turn_m1"` after restore). The steer is accepted, the carrier is claimed (`claimSteeringCarrierTurn`, the same durable state both the wake and the ladder produce), the session is closed and rebuilt through `RestoreSessionFromMetaWithConfig`. Fix at the root, in `forgetRunningTurnNoOneOwns`: only a pending `turn/start` or `turn/queue` owns the slot across a restart (those are reclaimed and re-run by their id); a steering mutation's id in the slot is a carrier that never ran to incorporation, and nothing re-runs a carrier by its id — the steer is still pending and the attach-time wake claims it afresh. GREEN: slot released, steer still pending, `turn/start` accepted and its run carries it.

**M, claim return fails too** (`TestSteeringCarrierKeepsTheSteerReachableWhenTheClaimReturnFailsToo`, RED: `provider requests = 2` — the carrier ran its model call without the steer). Fault: the steer's append is refused and `BeforeEffectSnapshotRename` refuses the second store write after the claim (the return; the first is `popSteeringHead`'s claimed-mark), asserted by the entry reading `claimed`. Fix: `consumeSteeringMessage` reports a failed return (warning) instead of dropping it; the carrier's `steeringCarrierUndelivered` now consults the transcript for a `claimed` entry (recorded = delivered with a failed incorporation write, round 2's case; unrecorded = both writes failed); and the carrier retry runs `recoverClaimedClientSteering` before it wakes — incorporating claimed steers the transcript holds, returning the rest to accepted, then reflecting — so the steer is reachable again. GREEN: no model request, the re-armed wake arrives, its retry delivers.

**L, retry budget** (`TestAnyRecordedSteerResetsTheCarrierRetryBudget`, RED: `attempts = 1` after a queued turn carried the steer). `clearSteeringCarrierRetry` moved from the carrier's success path to `consumeSteeringMessage`'s recorded path, so any delivery resets it.

Gates: gofmt clean; `go vet` ×3 flavors on `agent`, vet on root; `make lint` green (one `modernize` finding on my new loop fixed and folded in); `agent` package `-race -count=3` as name shards plus `TestLoadSessionJobActivityTree*` ×3 `-count=1`, all ok; other 41 agent-module packages ×3 ok; root `server`, `cmd/evener`, `internal/appprojector`, `cmd/evener-tui` ×3 ok, `cmd/evener-hub` ×1 ok; drain-ladder fuzz seeds ok.

https://claude.ai/code/session_011PbmbpbrXA9twoSwTRAf4B

obra:
## Review round 5 disposition (412d63de9)

CI green; reviewer 4 clean. One High and three Medium, all accepted, and the round closes the class rather than the instances: five rounds have each patched one path that decides what happens to an undelivered steer (Stop, the retry timer, restore, the ladder's claim, recovery), and each fix has exposed the next path that decides differently.
- **High, a Stop catching a carrier whose append and claim-return both failed discards a claimed steer for good** — accepted.
- **Recovery can race a carrier's in-flight append and re-materialise a steer that then goes stale at the head of the queue** — accepted.
- **A failed recovery leaves the claimed entry invisible with no retry re-armed** — accepted.
- **A failed carrier claim during a retry wake is treated as a no-op, stranding the steer** — accepted.
The fix is one reconciliation rule for an undelivered steer, with a state table (durable state × whether a turn owns it × whether the transcript recorded it × hold) that every deciding path calls, and a table-driven test over every row; any failure inside the rule re-arms the retry.

obra:
## Round 5 fixes

Head `c6f1978ef` (origin/main `ec9f1a5a6`, already merged). One commit that closes the class rather than the four instances.

**The one decision.** `agent/session_steering_reconcile.go`: `reconcileClientSteering(snapshot, inputs)` is the single place that decides what happens to a client steer no turn is delivering. Its doc comment carries the table — **8 rows** over durable `ExecutionState` (absent, accepted, claimed) × a turn in flight × the transcript recording it × parked (`SteeringHeld` or a Stop) — plus the active-turn slot rule (a carrier claim no turn is running under is released; the caller's own name never is). One action per row: drop a stale order entry (row 1: every retiring path nils the journal payload, so there is no input to rebuild from — the entry would otherwise refuse `popSteeringHead` for everything behind it); leave (2, 4, 5); leave parked / re-assert the hold on a Stop (3); finalize a recorded steer (6); return and park (7); return and re-arm the retry (8). The session-level wrapper samples the transcript question under `s.mu` **before** the store mutate (the serializer never waits on it), mirrors the store into the runtime queue afterwards, and re-arms the carrier retry on a refused store write — never warn-and-forget.

**Every deciding path calls it.** `finalizeClientMutationInterrupt` runs it with `stopping` and retires no steering entry itself (its three callers sample the inputs outside their closures and re-read the queue after); the retry timer runs it with `turnInFlight()` (session processing, or a taken slot) before waking; restore runs it over the loaded history with no turn in flight, and `forgetRunningTurnNoOneOwns` is the generic sweep again; `consumeSteeringMessage`'s failure path runs it as the calling turn (rows 7/8 decide park vs re-arm, so the carrier's exit no longer arms anything itself). Deleted: `recoverClaimedClientSteering`, the snapshot-level `steeringCarrierUndelivered`, `returnClaimedSteering`. A carrier claim the store refuses re-arms the retry inside `claimSteeringCarrierTurn`; a closed rail or occupied slot still returns nil and arms nothing.

**Tests.** `TestReconcileClientSteeringTable` walks all 8 rows (12 cases, the parked/Stop variants split) plus 4 slot cases on built snapshots. The four scenarios the reviews hit were RED against the round-4 code before this commit, through existing entry points: Stop + failed append + failed claim return → the steer was `interrupted` and gone (now accepted, parked, in memory); retry timer during an in-flight append → it was returned to `accepted` and re-materialized (now left `claimed`, not in memory); failed recovery write → attempts stayed 1 (now re-armed); store-refused claim on a retry wake → attempts 0 (now 1), held rail still 0. One existing test had its premise corrected: `TestStopOnASteeringCarrierTurnArmsNoHoldForItsOwnSteer` stopped a carrier whose append never lands, so finalization now returns that steer parked; the hold names it, which is not the hold-naming-nothing #710 forbids (asserted).

Wire shape unchanged.

Gates: gofmt clean; `go vet` ×3 flavors on `agent`, vet on root; `make lint` green (three `unused` findings from the deletions cleaned up and folded in); `agent` package `-race -count=3` as name shards plus `TestLoadSessionJobActivityTree*` ×3 `-count=1`, all ok — one shard hit `TestProbeCapabilitiesGitTimeoutIsIndependent` once under three concurrent shards (it pins a 1 ns git-probe timeout; green 5/5 isolated and ×3 with one neighbour; no steering code in its path); other 41 agent-module packages ×3 ok; root `server`, `cmd/evener`, `internal/appprojector`, `cmd/evener-tui` ×3 ok, `cmd/evener-hub` ×1 ok; drain-ladder fuzz seeds ok.

https://claude.ai/code/session_011PbmbpbrXA9twoSwTRAf4B

obra:
## Review round 6 disposition (c6f1978ef)

CI green. One High, three Medium, one Low, all accepted:
- **High, restore finalises a recorded carrier steer and then cannot see that the active turn was that carrier's, so the slot stays claimed for good** — accepted; the "was this a carrier claim" fact is taken before the table mutates the order.
- **The hold can outlive the last steer it named** — accepted; a hold naming nothing is released after the table runs, as restore already does.
- **A stale active-turn id reads as a live turn, so a retry leaves the steer and does not re-arm** — accepted; a live carrier is distinguished from a stale claim, or reconciliation runs again after the release succeeds.
- **`injectDrainedSteering` ignores a failed consume and re-pops the same steer, burning the retry budget in one turn** — accepted; the loop stops on failure.
- **Retry tests sleep instead of observing the timer** — accepted; the injected clock or a latch proves the callback ran.

obra:
## Round 6 fixes

Head `7d0e603ab` (origin/main `f614287ca` merged as `81f630099`). One commit; every change is in `reconcileClientSteering`'s table or its callers. All four new scenarios were RED against round 5 first (`agent/session_drain_as_steer_turn_boundary_test.go`), and the doc table and `TestReconcileClientSteeringTable` gained the rows.

**HIGH, slot outlives a finalized carrier steer** (`TestRestoreReleasesTheSlotOfACarrierWhoseSteerItFinalizes`, RED: `ActiveTurnID = "turn_m1"` after restore). Rule S's fact — does the order name the active turn — is now read **before** the rows run; row 6 (or row 1) removes exactly the steer that named it. GREEN: recorded steer finalized, slot released, `turn/start` accepted.

**M, hold outlives the last steer** (`TestStopReleasesAHoldTheLastSteerLeavesBehind`, RED: `SteeringHeld` stayed true after finalization retired the only steer; the next steer was accepted and could not claim a carrier). New rule H after the rows: a hold naming no pending user steer is released — release only, never arm. Restore's own copy of that release is deleted; the table does it for every caller.

**M, stale slot after a refused release** (`TestReleaseRetryReRunsTheTableAndWakes`, RED: the release retry landed and woke nobody). Kept `turnInFlight`'s reading of a taken slot (documented: nothing is written into a slot the retry is about to clear) and took the second option: `scheduleRunningTurnReleaseRetry`'s success callback runs the table again and wakes. GREEN on the fake clock: after `Advance(250ms)` the slot is clear, a wake arrives, and its run carries the steer.

**M, drain loop re-pops the failed steer** (`TestInjectDrainedSteeringStopsAtTheFirstFailedAppend`, RED: 4 attempts for two peeked steers across the turn's two drains — the table had put the failed steer back at the head each time). `injectDrainedSteering` breaks on `consumeSteeringMessage`'s false. GREEN: one attempt per drain, both steers still accepted and queued.

**L, sleeps** — `TestRetryLeavesAnInFlightAppendAlone` and `TestFailedRecoveryReArmsTheRetry` set `s.clock` to `agenttest.NewFakeClockAt` and `Advance` + `Drain`, so the assertion runs only after the timer's reconciliation completed.

Table doc now has the 8 rows plus rules S (with the read-before ordering) and H, and the stale-slot paragraph; the table test has 12 row cases plus 9 rule cases (the old "row 6 parked: hold untouched" expectation was wrong by rule H and is now "last steer finalized: hold released", with "kept while a pending steer names it" as its counterpart).

Gates: gofmt clean; `go vet` ×3 flavors on `agent`, vet on root; `make lint` green; `agent` package `-race -count=3` as name shards plus `TestLoadSessionJobActivityTree*` ×3 `-count=1`, all ok; other 41 agent-module packages ×3 ok; root `server`, `cmd/evener`, `internal/appprojector`, `cmd/evener-tui` ×3 ok, `cmd/evener-hub` ×1 ok; drain-ladder fuzz seeds ok.

obra:
## Review round 7 disposition (7d0e603ab)

CI green; three of four reviewers clean. One High, two Medium, one Low from one reviewer, all accepted; all four are paths that still decide a steer's fate outside the table or forget to re-arm on failure, so the round moves them into it:
- **High, any in-flight turn preserves any claimed steer, so a failed carrier's claim survives an unrelated turn and is never re-armed** — accepted; the in-flight rule names the owning turn.
- **The retry is cleared before incorporation is finalised** — accepted; cleared only once incorporation succeeds, re-armed when it fails.
- **A failed claim in `popSteeringHead` restores the entry and warns, arming nothing** — accepted; the failure reaches the carrier path and re-arms, benign races excepted.
- **Later tool rounds in the same input call the drain again and retry the steer before the backoff** — accepted; one per-input latch across every injection point.

obra:
## /simplify pass (96d957336)

Four reviewers (reuse, simplification, efficiency, altitude) read the whole diff. The altitude finding changes the shape of the fix, and the other two angles' findings mostly fall away with it, so round 8 applies it:
- **Accepted until recorded; the transcript decides.** A steer keeps one durable state, `accepted`, until its incorporation is recorded; the durable `claimed` state, the in-memory claimant map, the claim-return write and rows 5 to 8 of the table go. Restore already finalises from the transcript this way.
- **No paced retry.** A failed append fails the carrier turn and leaves the steer accepted, which already keeps the session from resting and arms the attach-time wake; that is the queue path's policy from attempt zero, and with piece A's poison on an unsynced write there is nothing to retry against. The eight-attempt timer, its budget, the per-input latch and the four re-arm sites go.
- **One rung.** The carrier runs as ordinary queued input with a synthetic entry carrying the reserved turn id, so the drain ladder, the ask-pending gate and the poison gate stop special-casing a kind.
- **Kept**: the reserved turn id for the Applied receipt, Stop's parking, the hold release (rule H) and the restore-time slot release, which are the parts justified by real constraints.
- Reuse items (the retry algorithm copied from the running-turn release, two recorded-steer readers, the in-flight predicate beside `sessionIsQuiesced`, test seams the package already has) close by deletion or by calling the existing helper.

obra:
## Review round 8 disposition (96d957336)

CI green; one reviewer clean. Two Medium and two Low, all accepted and folded into the round in flight, which removes the mechanisms three of them live in:
- **Restore leaves an accepted steer pending although its turn is already recorded, so it is delivered twice after a restart** — accepted; this is the "transcript decides" rule the reshape adopts, applied at restore to accepted entries as well.
- **A reconciliation can observe the durable claim before the claimant is published and return the steer under the carrier's feet** — accepted; the claimed state and the claimant map are removed, so there is no gap to observe.
- **The retry budget is not cleared when incorporation lands via the table** — accepted; the retry is removed.
- **A history scan on every input settle and every turn completion** — accepted; the settle-time table goes, and the recorded-steer answer becomes a maintained set rather than a scan.

obra:
## Review round 9 disposition (fbee03ae7)

CI green. Two High, one Medium, one Low:
- **High, a store written before this change can hold a `claimed` steer, which restore now neither queues nor finalises** — accepted as a finding; the fix is a load-time normalisation of a state this change retires (`claimed` and not recorded reads as `accepted`), which is a compatibility rule for existing stores, so it waits on Jesse's explicit approval before it is written. Everything else proceeds now.
- **High, an accepted steer whose incorporation write failed is omitted from the in-memory queue but still eligible as a carrier** — accepted; carrier eligibility excludes locally in-flight steering under one synchronised snapshot.
- **A steer re-queued after a failed append is never woken; the comment claims a retry that no longer exists** — accepted; the runnable steer is woken after it lands back in the queue, and the comment says what happens.
- **A comment counts seven drain actions where six exist** — accepted.

obra:
## Review round 10 disposition (f4f53d34a)

CI green; one reviewer clean. Two Medium and two Low, all accepted:
- **A skill-selection failure requeues a popped steer without clearing its in-flight marker, and a failed failure-record lets the drain pop it again** — accepted; the failure path lands the steer back through the one landing function and the drain stops when recording the failure fails.
- **A recorded steer whose incorporation write failed stays accepted with no retry, and the carrier cannot tell "back in the queue" from "recorded but unmarked", so it fails a turn whose steer is in the transcript and sends nothing** — accepted, and it closes #1389: the in-flight set carries whether the steer landed, the carrier reads that instead of the queue, and a recorded-but-unmarked steer is reconciled at the next store write or wake.
- **A carrier claim is released twice, the second time as an unpaced retry** — accepted; one release owner per invocation.
- **A test asserts the opposite of its name, and three comments cite a removed function** — accepted.

obra:
## Review round 10 disposition (1ed6db783)

CI 15/15 green. RoboRev: four Medium from one reviewer, two reviewers clean.

All four are failure-path claims about the carrier and reconciliation, so each gets a RED test through the public surface before a ruling; the implementer posts the trace either way.

- **M1 failed steering append re-wakes the carrier → hot loop** — measure which failure classes re-wake. Note the design ruling stands: no paced/timed retry in this PR ("accepted until recorded; the transcript decides"). If a persistent non-poisoning append failure really can loop, the fix is structural (the failed drain parks the steer and waits for the next external wake or store write), not a timer.
- **M2 recording a skill-selection failure fails → steer requeued, no wake rearmed** — measure: does the steer stay pending past the carrier's exit with nothing scheduled to pick it up?
- **M3 reconciliation of a recorded steer fails → stays in-flight, excluded from the runtime queue, no wake** — measure; if real, the next store write / wake must retry the incorporation mark without requeueing the recorded steer.
- **M4 skill-selection failure whose retirement write fails is later reconciled to `incorporated`, not `failed`** — measure against the `reconcileClientSteering` table; if real, the table must carry the failure outcome to a terminal `failed`.

Round 11 with the implementer.

obra:
## Review round 11 disposition (f8dda6db9)

CI 15/15 green. RoboRev: two Medium.

- **M1 the new `carrierSteerFailed` stand-down strands later valid steers drained in the same pass** — accepted. `injectDrainedSteering` keeps draining after the head steer is retired as a selection failure, so a later valid steer lands in the transcript, then `carrierSteerOutcome` sees the head's failure and stands the turn down with nothing arming a wake for the delivered steer. This is the #1308 symptom in a new coat. Fix: the stand-down decision keys on whether the drain delivered anything this turn, not on the carrier's own identity alone. RED test: bad-skill steer then a valid steer, run the carrier, assert the second steer's request is issued.
- **M2 carrier-claim persistence failure leaves the accepted steer queued with no wake** — accepted as a contract gap, with the fix direction changed: no bounded/backoff retry (the design ruling for this PR removed timed retry). It gets the same parking contract round 11 gave the failed append and the refused failure record: the steer stays runnable and the next external wake or accepted mutation carries it, pinned by a test so the contract is explicit rather than incidental.

Round 12 with the implementer.

obra:
## Review round 12 disposition (f6f169425)

CI 15/15 green. RoboRev: three Medium (one reviewer clean).

- **M1 restore derives `steeringOutcome` from `s.history`, which `resumeHistory` truncates at the last checkpoint, while `incorporated` comes from the full transcript index** — accepted. A recorded-but-unmarked steer whose turn was compacted out reads as unrecorded at restore and is re-queued, so the next run delivers it twice; the old restore finalized from `incorporated` alone. Fix: derive the terminal state from the full-transcript index (`restoredClientMutationItems[id].Failure` → failed, else incorporated when `incorporated[id]` matches the pending turn), never `""` on an `incorporated` match. RED: record, compact past it, restore, assert one delivery.
- **M2 a failed `claimSteeringCarrierInput` reads as "no carrier", so the drain selector may run an autonomous continuation that drains the steer under a different turn id than the client's reserved receipt** — measure; if the autonomous path can accept a client steer in that state, the claim failure becomes a distinct state that suppresses autonomous follow-ons until the next external wake (parking contract, no timer).
- **M3 the acceptance wake queued while the current input runs is consumed right after an inline carrier failure, retrying it despite one-attempt-per-external-wake** — measure with the inline carrier failing under a pending acceptance wake; if it reproduces, the input that claims the carrier consumes the acceptance wake (or the failed attempt marks it stale).

Round 13 with the implementer.

obra:
## Review round 13 disposition (85d187cf1)

CI 15/15 green. RoboRev: one Medium (one reviewer ran; two skipped).

- **`steeringParked` gates the drain selector's notification rung but not the two continuation paths** — accepted. The deferred-continuation block at `session_lifecycle.go:1399` runs on a `haveDeferredCont` armed at an earlier tail regardless of the selector's `goIdle`, and `settleGoalOnIdle` (`session_goal.go:371`) kicks a continuation with no parked guard; both reach `acceptContinuationInput` → `injectDrainedSteering`, so the pending steer is consumed under the continuation's fresh turn id, not the receipt's. Parked becomes authoritative for every autonomous path: gate the deferred block and the goal kick on the parked state (defer, never drop). Test: active goal via `SetKickFunc`, steer pending, claim write fails, assert no continuation turn and no extra model request before the next external wake. This is the same class as the residual filed as #1453 (a daemon-initiated notification outside the ladder); if the same gate covers that path, close #1453 in this round, otherwise say why it is a different path.

Round 14 with the implementer.

obra:
## Review round 14 disposition (06505753b)

CI 15/15 green. RoboRev: one Medium (one reviewer ran).

- **The drain ladder's carrier claim at `session_lifecycle.go:1299` ignores `steeringParked`** — accepted. The comment's premise ("a carrier whose append failed ends the input") holds only for carrier turns; when a running user turn's own boundary drain (`acceptUserInput` / `injectPostToolSteering` → `injectDrainedSteering`, result ignored) fails the append, `consumeSteeringMessage` parks the steer, the turn completes normally, and the ladder re-claims the parked steer inline with the rail open. Fix: the ladder claim (and the matching `QueuedCarrier` / `skipGoalGate` decision) consults the park, so a steer parked by the running turn's failed drain stays queued for the next external wake; fix the comment. RED: user turn, boundary drain append faulted, turn completes, assert no inline re-claim and that the next external wake carries it under its receipt id.

Round 15 with the implementer.

PR #1343, opened 2026-09-15
fix(sdk): make watchFoldKey order-canonical

### LINKED ISSUE #1331: entity links follow-up: watchFoldKey is order-sensitive, so a reordered item array rebuilds the entity view
Follow-up from the transcript entity links work (#1298, merged as d21a136ef).

`watchFoldKey` in `appwire-client/typescript/entityView.ts` serializes the `job_watch` items in caller array order. `foldWatchSummaries` now orders snapshots by transcript position, so the folded result no longer depends on array order — but the key still does. The same watch content arriving in a different array order therefore produces a different key and rebuilds the memoized entity view for nothing.

Correctness is unaffected; this is rebuild cost only.

Suggested fix: sort the tuples with the same transcript-position comparator the fold uses (`compareTranscriptPosition` in `watchRows.ts`), keeping every content field in the key so a content change still changes it.

Deferred out of #1298 because it lives in a different file than that review finding's scope, and the reviewer's finding was about the fold, not the key.

obra:
Fixes #1331.

`watchFoldKey` serialized the `job_watch` items in caller array order, while `foldWatchSummaries` orders its snapshots by transcript position. The same watch content arriving in a different array order therefore produced a different key and rebuilt the memoized entity view for nothing. Correctness was unaffected; this is rebuild cost only.

The key now sorts its tuples with the same `compareTranscriptPosition` comparator the fold uses (exported from `watchRows.ts` for the purpose), keeping every content field in the key so a content change still changes it. Unpositioned items compare equal and keep their relative order, so well-ordered input produces the same key as before.

Tests: two new cases in `appwire-client/typescript/entityView.test.ts` — the key is order-canonical for positioned items (this failed before the fix), and unpositioned items keep caller order (the fixtures carry distinguishing content, since the key deliberately omits item ids).

Verification: the two touched suites pass in the hub frontend's Vitest project (36/36), and `make test-api-package` exits 0.

Note on the local `make test-web` run: it reported 3 failures in files this diff does not touch (2 in `src/panes/spawn/Spawn.test.tsx`, 1 in `src/dev/editorial-preview/fixture.test.tsx`) under load ~50-60, with 10788/10791 tests passing. Both files pass in isolation with this change present, at equal or higher load, so I read those as contention flakes rather than regressions.


obra:
Addressed in `ab574e66f`. The finding was right, and it was not fixable in the key alone: `entityView` iterates the fold's map into the entity map, so the key has to mirror the fold's ordering — a "canonical" key against a non-canonical fold would have traded the spurious rebuild for a stale view. The ordering itself had to become total.

**The change:** a positionless item now sorts after every positioned one, and positionless items stay equal to each other (so among themselves they keep caller order, which for well-ordered input is the transcript's own).

**Regression tests at both levels, with RED evidence.** Both start red against this commit's parent:

- `entityView.test.ts` — "watchFoldKey is order-canonical across a mixed positioned and positionless list":
  `AssertionError: expected '[[null,null,"{\"job_id\":\"job_x\"}",…' to be '[[2,1,"{\"job_id\":\"job_x\"}","{\"wa…'`
- `watchRows.test.ts` — "a positionless snapshot sorts after positioned ones, so it decides the fold":
  `AssertionError: expected 'watching' to be 'ended'` — i.e. the caller's array order, not the content, was picking the fold's last-wins winner.

Green after: 38 tests in the two files, and the full gates on the new tip print `PASS web-typecheck` / `PASS web-test` / `PASS web-lint`, exit 0, plus `make test-api-package` qualified, exit 0.

**Deliberately left as-is:** two positionless items remain equal to each other, so their relative order still follows the caller. That is the fold's existing "no opinion" rule for items that carry no transcript place, and the alternative — ordering them by content — would override caller order in the case where caller order *is* the transcript's, and would need its own decision. One positionless item against positioned ones is now deterministic in both directions.

PR #1345, opened 2026-09-15
feat(web): fold every rail section heading, chevron trailing the title

obra:
## What

Every rail section heading is now the same disclosure fold, and the chevron sits **after** the title rather than before it.

## Why

Live and Projects were inert labels you could not fold away, while the archived and pinned disclosures led with their chevron — reading backwards against the rail's own rows, which draw the chevron after the label (`RailRow`'s `TrailingChevron`).

## How

One `SectionHeading` (an `h3` wrapping a `button[aria-expanded]` whose chevron trails the label) now backs all five section kinds: Live, pinned sections, Projects, Test runs and Archived sessions.

- Live, pinned sections, Projects and Test runs start **open**; Archived sessions stays **collapsed**. Defaulting the new folds open preserves today's behavior, so nothing disappears on load.
- Fold state persists per section id in the existing `evener.rail.expanded.v1` map: `section:live`, `section:projects`, `section:test_runs`, alongside the pre-existing `section:archived`.
- All five toggles route through one `toggleSection`.

## Reveal

Making these sections foldable meant a `/project` reveal could land on a row sitting inside a closed section. The reveal now opens whichever section it lands in — the project's own section (Projects, Test runs or Archived), a pinned section, or Live — before scrolling. Three tests cover the paths and each fails without the branch.

## Scenario card

`test/scenarios/sidebar-archived-testruns-reachability.md` described Test runs as "an `<h3>` with no disclosure button" and called a collapsible Test-runs section the "pre-rewrite shape" that would falsify it. Updated to the folded shape.

## Fixed after the first CI run

`web` failed on `web-shellguard`, and it was right: with every heading now a real control, the mobile session list carried `button.sectionDisclosure` at **23px** tall, under the 44px tap floor this repo enforces (`tokens.css`'s `--tap-min`). Fixed with `min-height: var(--tap-min)` under `(pointer: coarse)`, matching `RailRow.module.css`'s own tap-floor block, so the desktop heading geometry is untouched.

## Evidence

- `make test-web` on CI: **web-typecheck, web-test, web-lint all PASS** (the full unit suite is green there; two local runs hit unrelated load timeouts in `Spawn.test.tsx` and `App.test.tsx`, both of which pass in isolation on this head).
- `web-skillguard` passed on CI.
- `node scripts/shellguard/run.mjs` after the fix: **281 mobile tap targets all >= 44px**, exit 0.
- Rail suite 53/53; `tsc --noEmit` and `biome check` clean.
- The first CI run's other failure — `TestDelegateResourceSupervision_EarlyCommittedSendStartRefusesASecondTurn` in `agent` — is a flake unrelated to this TS-only diff: it passes 5/5 locally on the same head.
- Real Chrome against the dev surface gallery (the actual `Rail`, fixture data): `h3 > button[aria-expanded]` with the chevron after the label; clicking Live flips `aria-expanded`, removes the section's tree from the DOM, and writes `section:live: false` to `evener.rail.expanded.v1`; a reload restores it collapsed.
- Reviewed by four independent read-only passes (reuse, simplification, efficiency, altitude) before the first push; their findings are folded in.

## Notes

Rebased onto `main` (`34e20d433`) so the strict required checks can pass; `main` already carried fixes for the two skillguard flakes seen locally (#1327, #1317).


obra:
## RoboRev `3ab87f5` findings — disposition

**Medium — test-run reveal routes to the wrong catalog.** Fixed. `projectPlacement` now derives the fold, its default, and the catalog together, so a test-run target opens `Test runs` *and* requests `test_runs`. New test: `a reveal of a test-run project opens Test runs and loads that catalog` (verified red against the old `catalog: "projects"`).

**Medium — archived detection fails for whole-project archives.** Fixed. A whole-archived project's sessions keep their own `current`/`recent` tier, so `tier === "archived"` missed them; the placement now also reads membership from `resources.archivedProjects`. New test covers an archived-target reveal asserting the fold opens and `archived_projects` is loaded.

**Medium — missing reveal coverage for archived and test-run folds.** Fixed; both tests added (above). Every reveal test in this PR was checked to fail with its branch removed.

**Low — `--tap-min` token scope mismatch.** Fixed. You are right that the token is declared inside `tokens.css`'s `max-width: 899px` block, so the declaration would be dropped on a wider coarse-pointer viewport. It now reads `min-height: var(--tap-min, 44px)`.

**Low — redundant `aria-label` and inverted comment.** Both fixed. The heading takes its name from the button it wraps (the chevron is `aria-hidden`), and the comment now describes the real DOM — an `h3` wrapping a disclosure button.

**Not addressed here — the synthetic `archivedgroup:<key>` node is never expanded on reveal.** Valid, and I am deliberately leaving it out of this PR. It is pre-existing: `Archived sessions` was already a folded section before this work, so archived reveals were already unable to complete. It is also a different mechanism from the section folds — fixing it means exporting `archivedGroupId` from `railNodes.ts` and deciding, per target, whether the reveal expands the archived *group* node or the project node. That is a design call rather than a fold follow-up, so I have left it out and flagged it rather than guess. Happy to take it as a separate change.

obra:
## `tests` and `race-modules / agent` are failing on `main`, not on this branch

Both failures are the same test:

```
--- FAIL: TestWithIntentParameter_DescriptionGuidesExpectedOutcome
    registry_test.go:40: description lacks expected-outcome guidance: "What you hope to learn or accomplish from this tool call, using a verb-first gerund. Make your hypothesis and the desired outcome clear; ..."
```

`main`'s `4059377e2` ("tweak intent descriptions") rewrote `toolIntentDescription` in `agent/internal/tool/registry.go` to say **"desired outcome"**, while `agent/internal/tool/registry_test.go:39` still asserts the description contains **"expected outcome"**. The commit touched only `registry.go` (1 insertion, 1 deletion).

Reproduced against the untouched base, with no branch content involved:

```
$ git worktree add --detach /tmp/evener-main-check origin/main
$ (cd /tmp/evener-main-check && go test ./agent/internal/tool \
     -run '^TestWithIntentParameter_DescriptionGuidesExpectedOutcome$' -count=1)
--- FAIL: TestWithIntentParameter_DescriptionGuidesExpectedOutcome (0.00s)
FAIL	primeradiant.com/evener/agent/internal/tool	0.009s
```

`main`'s own CI run for `4059377e2` reports `failure`. This branch carries no Go changes at all — the whole diff is `cmd/evener-hub/frontend/src/shell/rail/` plus one scenario doc — so this PR inherits it through the required merge commit, and cannot go green until `main` does.

Every other check on this head passes, including `web`, `static`, `lint-golangci`, `fuzz`, `race-root` and `race-modules / nonagent`.

Owner: `4059377e2` (Jesse Vincent). Not absorbing an unrelated fix into this PR.

PR #1346, opened 2026-09-15
ci: retry the golangci-lint download (#1145)

obra:
Fixes signature 1 of #1145: `make tools-golangci` pipes the upstream golangci-lint installer into `sh`, and the installer's own `http_download_curl` fails on any non-200 with no retry, so one transient GitHub HTTP 500 failed the required `lint-golangci` job (run 34644540860).

Two tiny layers in `make/repo.mk`:

- `curl --retry 5 --retry-all-errors --retry-delay 2` for the installer script itself.
- A four-attempt loop (3s apart) around the whole install, because the installer's internal downloads are the ones that saw the 500 and curl's flags do not reach them. Every attempt failing prints `golangci-lint install failed after 4 attempts` and exits 1.

The script is held in a shell variable and fed to `sh` with `printf` instead of piped directly; that also closes the pre-existing hole where a failed `curl -f` fed `sh` an empty script and the target reported success (the new test's failure case caught exactly that on main). The pinned version is read from `.tool-versions` exactly as before.

Test: `TestToolsGolangciRetriesTheDownload` (root package) runs `make tools-golangci` with PATH prepended by a temp dir holding a `curl` shim that fails N times then prints a no-op installer, plus a no-op `sleep` shim so the delay costs nothing. 3 failures succeeds, 4 fails loudly. Runs in under a second.

Diff size: 2 files, +44/-1 (recipe +7/-1, test +37). No new scripts, helpers, or make targets.

Gates run in the foreground: root Makefile audits (TestEveryRuleIsPhony, TestEveryLintTargetIsPhonyAndHasARule, TestMakeLintRunsEveryGateInLintTargets, TestEveryTargetHasASummaryAnnotation) pass; `make tools-golangci` for real installs 2.13.1; `make lint` passes all nine gates.

The second signature in #1145 (hung `go list`) is already bounded on main by scripts/gate/run-module-tests.sh and is untouched here.


obra:
## Review round 1 disposition (bd0e8e318)

CI green. One Medium and two Low, all accepted:
- **`--retry-all-errors` needs curl 7.71, so older system curls fail before any request and the loop fails deterministically** — accepted; the flag goes, the outer loop already retries the whole command.
- **The last attempt still says "retrying" and sleeps** — accepted; only when another attempt remains.
- **The test requires make and a POSIX shell unconditionally** — accepted; it skips when either is unavailable.

obra:
## Round 1 fixes

Head: `50ef7221ee239cebcd2c5d906b435766779f3256` (main merged, nothing to reconcile).

1. Medium, `make/repo.mk`: dropped `--retry-all-errors`; curl keeps `--retry 5 --retry-delay 2` and the outer loop covers the rest.
2. Low, `make/repo.mk`: the "retrying in 3s" message and the sleep now run only when another attempt remains (`[ "$attempt" -lt 4 ]`); the final failure still prints "failed after 4 attempts".
3. Low, `tools_golangci_retry_test.go`: `t.Skipf` when `make` or `sh` is not on PATH, matching the `exec.LookPath` guard in `TestMakeLintRunsEveryGateInLintTargets`.

Diff vs main is still 2 files, +50/-1. Gates in the foreground, all green: the four root Makefile audits, `TestToolsGolangciRetriesTheDownload`, `make tools-golangci` for real (installs 2.13.1), `make lint` (nine gates PASS).

PR #1349, opened 2026-09-15
refactor(web): one watch-condition composer behind the card and the list

### LINKED ISSUE #1334: entity links follow-up: the watch card and the watch list word conditions differently
Follow-up from the transcript entity links work (#1298, merged as d21a136ef).

`src/panes/session/transcript/EntityRef.tsx` composes a watch condition sentence of its own (`watchConditionContent`), while `src/panes/session/transcript/tools/jobWatch.tsx` has `rowConditionPhrase` and `conditionSentence`. #1298 shared only the wildcard label (`watchEventLabel.ts`), after the card had drifted to rendering `events *` where the list says `any event`.

That drift was a real defect, found in review. Two composers for the same data is how it happened.

Suggested fix: one composer in the transcript subtree that both surfaces call. Cosmetic in itself, but it has already produced a user-visible bug once.

obra:
Fixes #1334.

The watch card composed a condition sentence of its own (`watchConditionContent`) while watch list rows composed theirs in jobWatch's `rowConditionPhrase`, and the two vocabulary copies had already produced a user-visible defect: the card rendered `events *` where the list said `any event` (the shared wildcard label from #1298 was the stopgap for that one case only).

`watchConditionPhrase.ts` now owns the trigger vocabulary — pattern, timer, events with the every throttle, event filter, heartbeat — and both surfaces render it, so the card words a condition exactly as a list row does and a vocabulary change lands in one place.

The list's wording and order are unchanged: jobWatch's assertions are untouched and green, which is the check that this refactor is behavior-preserving on the side that had the stronger contract. The card converges to the list, which is a deliberate user-visible copy change: a pattern now reads `“ready”` rather than `output matching “ready”`, and a timer `in 5m` rather than `fires in 5m`. One of the two surfaces had to give for them to word identically, and the list's wording carries more review history behind it; happy to flip the direction if the card's more explanatory labels are preferred.

Added a 7-case vocabulary table for the composer, including the wildcard case that drifted.

Verification: `make test-web` exits 0; the transcript subtree is 84 files / 1961 tests green; the focused card + composer + jobWatch suites are 123 tests green.


obra:
Checked the High finding on `watchConditionPhrase.ts` — it's a false positive, with three independent pieces of evidence:

1. **The pattern compiles.** `import { type parseConditionText }` used as `ReturnType<typeof parseConditionText>` is a *type query*, not a value reference. Reproduced minimally with this repo's own compiler (the hub frontend's TypeScript 6.0.3):

   ```
   a.ts: export function f(x: string): { y: number } { return { y: x.length }; }
   b.ts: import { type f } from "./a";
         export type R = ReturnType<typeof f>;
   $ tsc --noEmit --strict b.ts   →   exit 0
   ```

2. **This commit typechecks locally:** `npm run typecheck` (which is `tsc --noEmit`) exits 0 on `cc840e628`, and `make test-web` printed `PASS web-typecheck`.

3. **CI agrees on this exact head:** the `web` check passed on this PR (6m56s), and that job runs the same typecheck.

Worth noting the `type` marker was not hand-written: Biome's `useImportType` assist rewrote that import during `npx biome check --write`, and `make web-lint`/CI enforce that assist, so dropping the marker would be reverted by the next lint run.

The alternative fix — importing the parsed-condition *type* directly instead of deriving it — is not currently available: `ParsedCondition` is declared but not exported from `appwire-client/typescript/watchRows.ts` (nor re-exported from the package index). `ReturnType<typeof parseConditionText>` is the workaround for that. If a real type error ever shows up in a lane that compiles differently, the right fix is to export `ParsedCondition` from the package (which changes its public surface and its `publicExports` test), not to drop the import-type marker.

PR #1361, opened 2026-09-15
feat(hub): remote admin proxy with a fail-closed allow-list (07a)

obra:
# 07a — remote admin proxy with a fail-closed allow-list

Component 07a of the multi-host SSH federation (spec
`docs/superpowers/specs/2026-09-14-multi-host-07-remote-admin.md`). Backend
only; 07b (frontend host scoping), 07c (credential push) and 07d (host device
login) are follow-ups.

## What this adds

**`evener/host/request` (new hub-scoped method).** One wrapper RPC that forwards
one hub-scoped admin RPC to a named remote host's hub and returns that method's
own result verbatim. Handler `hubHostAdminController.Request`
(`cmd/evener-hub/app_host_admin.go`):

1. resolves `host` through the component-03 host registry — unknown host is
   `appwire.InvalidParams`;
2. refuses a host that is not attached (component 05's `RemoteHubSource.Online()`,
   wired from component 04's `sshManager.Attached`) with `appwire.Unavailable` —
   a remote request never falls back to local execution;
3. allow-lists `method` against an **exact-name** set, then
4. forwards through the per-host `appwire.Client` and passes wire errors
   through unchanged (a launch credential-env refusal or an auth Codex/gcp-adc
   refusal reaches the browser verbatim).

**Why the allow-list is fail-closed.** `remoteHostAdminMethods` names all 38
methods of the five families — `evener/instance/{list,create,edit,remove,setDefault}`,
`evener/launch/{resolve,schema,getLayer,setLayer,trustRepo}`,
`evener/marketplace/{list,add,remove,refresh,edit,browse}`,
`evener/plugin/{list,install,upgrade,remove,enable,disable,setAutoUpgrade,preview,checkNow}`,
`evener/auth/{status,test,list,login/start,login/complete,logout,apiKey/set,apiKey/clear,credentialJson/set,device/start,device/poll}`,
`evener/settings/agentsDoc/{get,set}` — and nothing else. There is no prefix
matching: a prefix rule would auto-allow the next sensitive `ScopeHub` method
the moment it entered the catalog (a hypothetical `evener/instance/deleteAll`),
turning the proxy into a generic hub-to-hub tunnel by default. An unlisted
method, including every method added after the list was written, is
`appwire.InvalidParams` and is never forwarded. `TestHostAdminAllowListMatchesCatalog`
enumerates `appwire.CatalogMethodNames(appwire.ScopeHub)` against a 98-row
allow/deny table, so a catalog addition fails the test until someone records a
deliberate decision.

**Notification fan-out.** The controller subscribes to component 05's
source-level notification stream and re-emits **only** the config notifications
it owns — `evener/auth/updated`, `evener/launch/updated`,
`evener/marketplace/updated`, `evener/plugin/updated`,
`evener/settings/agentsDoc/changed` — to its browser clients as
`evener/host/notification` with `{host, method, params}`.

**Local notifications are unchanged.** The wrapper is emitted only for remote
hosts. The existing `notifyAuthUpdated` / `notifyLaunchUpdated` / … `BroadcastAll`
path still fires the bare methods for the controller's own hub, so a store that
is not host-scoped never sees remote traffic and cannot misapply it.

**Secrets boundary.** The controller writes nothing and persists nothing: it
forwards params, returns results, and re-emits notifications. No credential
value is read, logged, or stored by this component (07c is where the credential
push lives, and it will write only through the remote hub's own
`evener/auth/apiKey/set`). The proxy is the only new surface, and it is
allow-listed.

## Deviation from the spec's cited seams

The spec cites two seams that had moved or did not exist; both are called out
in code comments:

- **Component 03 registry.** The spec says "resolves `Host` through
  `hostreg.Registry.Get`". Component 03's `*hostreg.Registry` is built in
  `cmd/evener-hub/main.go` and handed to the SSH manager; it is **not** on
  `hubcore.WebConfig`, which carries only the validated `RemoteHosts
  []hostreg.Host`. Rather than thread a new field through `main.go` and
  `hubcore/config.go` (outside this PR's files), `registerHostAdminHandlers`
  rebuilds an equivalent registry with `hostreg.New(cfg.RemoteHosts)`. The
  accessor and its unknown-host semantics are component 03's exactly; only the
  instance is local. An error there falls back to an empty registry, so the
  proxy refuses every host rather than panicking.
- **Component 05 had no broker and no admin accessor.** `RemoteHubSource`
  exposed only thread-routed subscriptions; its `drainLoop` dropped every
  notification with no thread route, and the per-host `appwire.Client` was
  private. Two minimal additive pieces were added in
  `cmd/evener-hub/internal/appsource/`: `RemoteHubSource.AdminCall` (resolves
  the shared client per call, forwards, returns errors unchanged) and
  `RemoteHubSource.SubscribeHostNotifications` (+ a `hostSubs` set, a
  fan-out line in `routeNotification`, and orphan cleanup in `drainLoop`).
  This is deliberately **not** a second reader of `Client.Notifications()`,
  which already has exactly one consumer. Registration starts the drain, so an
  admin-only consumer with no thread subscribers still receives notifications;
  the channel closes when the client's stream ends, so the fan-out re-subscribes
  and rebinds on reconnect. The change is additive; the thread-routing path is
  byte-for-byte unchanged.

## Tests

All deterministic and offline (in-memory `appwire.StreamTransport` pairs, no
SSH, no network):

- `cmd/evener-hub/app_host_admin_test.go`
  - forwards an allow-listed method and returns the remote's result, with the
    caller's params unchanged;
  - a remote `appwire.WireError` is returned verbatim (code preserved, the
    host's own text kept);
  - unknown host, the reserved `local` name, an unattached host, and a
    configured host with no component-05 source are each refused with the typed
    error and **nothing is forwarded**;
  - the fail-closed allow-list boundary test enumerated against
    `appwire.CatalogMethodNames(appwire.ScopeHub)` (allowed ⇒ forwarded, denied
    ⇒ `InvalidParams` and no remote call);
  - the spec's explicit method set (including `evener/plugin/disable`,
    `evener/plugin/setAutoUpgrade`, `evener/settings/agentsDoc/{get,set}`);
  - fan-out: a scripted remote emits a thread notification (must be dropped)
    then `evener/auth/updated` (must be re-emitted as one
    `evener/host/notification` tagged `host: "m4"`, payload unchanged).
- `cmd/evener-hub/internal/appsource/remote_hub_admin_test.go`
  - `AdminCall` forwards params unchanged and returns the remote's raw result;
  - a semantic wire error survives `AdminCall`;
  - `SubscribeHostNotifications` delivers a host-level notification that carries
    no thread route, and a cancelled context unregisters the consumer.
- `appwire/protocol_test.go`: `evener/host/notification` is added to the
  deliberate hub-wide notification list in
  `TestThreadNotificationsRequireAuthoritativeRoutingIdentity` (it is a
  `BroadcastAll` hub-wide re-emission with no thread identity of its own).

## Gates

All from the repo root on `multi-host-pr07a-host-admin` (`90a42fba8`):

- `gofmt -l appwire/ cmd/evener-hub/` → empty
- `go build ./...` → ok
- `go test ./cmd/evener-hub/internal/appsource/ ./appwire/ -race -count=3` → ok
  (`appsource 7.688s`, `appwire 3.390s`)
- `go test ./cmd/evener-hub/ -race -count=3 -timeout 40m` → ok (`790.373s`)
- `go test ./cmd/evener-hub/... -count=1` → ok, all 13 packages
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint`
  → `PASS lint (1 modules, 15s)`, 0 findings
- `make lint-generated` → `PASS` (regenerated `docs/appwire-protocol.md` and
  `appwire-client/typescript/types.gen.ts` are committed)
- `make lint-fuzz-registry`, `make lint-evenerfuzz` → `PASS` (no fuzz target
  added, so `make fuzz-gap-check` is unaffected)

One note on the race gate: `go test ./cmd/evener-hub/ -race -count=3` exceeds
go test's default 10m per-package timeout on this package (it needs ~13m), so it
is run with `-timeout 40m`. The first run of that command surfaced two real
omissions that are now fixed in this branch: the proxy's constructor hit a nil
source registry (an embedder path that calls `newHubAppServerWithNavigation`
with no sources) and `TestHubRPCRegistersExpectedHandlerSet` did not yet name
`evener/host/request`.

## Scope

Component 07a only. 07b (frontend host scoping for the settings panes), 07c
(credential push), and 07d (host device login) are follow-ups and are not
started here.


obra:
Round one addressed (`32597112c`).

**Medium — `AdminCall` bypasses `mapCallError` (fixed).** Confirmed: it returned raw `s.client` / `client.Request` errors, and pre-fix test showed a mid-call EOF surfaced as `CodeInternalError` (`-32603`) rather than `CodeUnavailable` (`-32014`). `AdminCall` now routes every error through `s.mapCallError`, so a dead channel becomes `SessionUnavailable`/`CodeUnavailable` (matching the pre-call offline refusal) while a remote `WireError` refusal keeps its own code and message verbatim. Pinned by `TestRemoteHubSourceAdminCallTransportFailureBecomesSessionUnavailable` (fails pre-fix: `code = -32603, want -32014`); the existing `TestRemoteHubSourceAdminCallPreservesSemanticWireError` and `TestHostAdminRequestReturnsRemoteWireErrorVerbatim` still pass unchanged.

**Medium — subscription lifetime defects (fixed).** All three sub-issues were real. The host-level subscription now mirrors `remoteHubSubscription`'s single-owner pump:
1. *Per-reconnect goroutine leak* — the context-blocked watcher is gone. Each `SubscribeHostNotifications` runs one pump that exits when its context ends **or** when `drainLoop` closes its `in` on client teardown, so a flapping host no longer accumulates one goroutine (and one 128-slot buffer) per reconnect.
2. *Blocked `fanOut` on cancel* — `fanOut` now derives a per-cycle child context, and `relayHostNotifications` selects on `ctx.Done()` while draining, so a canceled context unblocks promptly. Pinned by `TestHostAdminFanOutStopsWhenContextCanceled` (pre-fix: `fanOut did not return after its context was canceled`).
3. *Broken closure contract* — the drain now closes `in` (never `out`); the pump closes `out` on **both** termination paths, so the returned channel really does close on ctx-end or stream-end. Pinned by `TestRemoteHubSourceHostSubscriptionClosesOnContextEnd` (pre-fix: `subscription channel was not closed after its context ended`) and `TestRemoteHubSourceHostSubscriptionClosesWhenClientStreamEnds` (regression guard). The existing unregister-on-cancel test still passes.

**Medium — head-of-line blocking in host fan-out (fixed).** Confirmed: `publishHostNotification` blocking-sent on the shared `drainLoop`, so a lazy admin consumer could stall all thread routing for that client. Host delivery is now non-blocking — a full buffer drops and increments an internal `hostNotifyDropped` counter — deliberately trading a possibly-stale settings pane for never stalling `turn/completed`. Pinned by `TestRemoteHubSourceHostFanOutDoesNotBlockOnFullConsumer` (pre-fix: `publishHostNotification blocked on a consumer that stopped reading`). Note this intentionally changes the old "blocking rather than dropping is deliberate" disposition for *host-level* fan-out only; thread routing still blocks and never drops, and the fail-closed allow-list is untouched.

**Low — allow-list validated after availability check (fixed).** `Request` now checks `remoteHostAdminMethods` before `remoteSourceFor`, so a forbidden method returns `InvalidParams` even for an offline/unattached host and never touches the source registry. Pinned by `TestHostAdminForbiddenMethodRefusedBeforeAvailabilityCheck` (pre-fix: `wire code = -32014, want -32602`). Added a deny-row-style case with `evener/instance/deleteAll`. `TestHostAdminAllowListMatchesCatalog` and the offline/unattached refusal tests still pass.

Verification: `gofmt -l cmd/evener-hub/` clean; `go build ./...` OK; `go vet ./cmd/evener-hub/...` OK; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` OK; `go test ./cmd/evener-hub/ -race -count=3` OK (798s); `go test ./cmd/evener-hub/... -count=1` OK (13 packages); `MODULES="." go run ./cmd/evener-dev/bin dev module-lint` PASS (0 findings).

obra:
## Allow-list extension (specs round seven): pushed `33a34183e`

**Methods added** — exact names, each verified against `appwire/protocol.go` (the spec named no method that does not exist):

| method | scope | why it is safe to proxy |
| --- | --- | --- |
| `evener/paths/complete` | hub | path auto-completion in a remote pane; read-only against the host filesystem |
| `evener/path/validate` | hub | validates the spawn form's path on the selected host; read-only |
| `evener/dirs/create` | hub | creates the directory the user just asked for — the only mutation, and exactly what the local pane already does |
| `evener/projects/recent` | hub | the host's recent project directories; read-only |
| `evener/harnesses/list` | hub | the host's own harness inventory; read-only |
| `evener/spawn/slashCatalog` | hub | the host's pre-session slash catalog; read-only |
| `model/list` | both | the host's own model inventory; read-only (spec notes this one is `ScopeBoth`, not `ScopeHub`) |
| `evener/git/head` | hub | round-eight routing inventory: read-only branch metadata for the remote path the form is about to launch |

`evener/plugin/preview` and `evener/instance/list`, also named in the round-eight inventory, were already allow-listed in the original list. Every forwarded method is executed by the host hub against the host's own environment through the same host-side handler as before; the list stays an explicit, closed, exact-name set — no prefixes, no generic tunnel — and the remote hub never gains a path to controller-local state or secrets beyond what the admin families already carry.

Citations: `07-remote-admin.md:91-96` (host-dependent discovery bullet) and `:152-162` (proxy step 3, "exact set of method names … and nothing else"); `06-fleet-view.md:354-380` for the routing mechanism. Note: the fetched `multi-host-specs` head does not yet name `evener/git/head` in either allow-list spot; I added it on the round-eight instruction and verified it in the catalog (`protocol.go:150`, `ScopeHub`, `GitHeadParams`/`GitHeadResponse`).

**Boundary test.** `TestHostAdminAllowListMatchesCatalog` still enumerates `appwire.CatalogMethodNames(appwire.ScopeHub)` and still fails on any catalog method with no row, so a future addition forces a decision. The policy table now carries an explicit per-method row (true = proxied, false = refused with `InvalidParams` without ever reaching the remote), the deny families are documented in the table header, and each newly allowed row is annotated with its rationale. Denied methods are exercised for the fail-closed property; the newly allowed ones are exercised for forwarding. New regression `TestHostAdminRequestForwardsDiscoveryMethodAndReturnsResult` proves `evener/paths/complete` actually reaches a scripted remote hub with the caller's params and returns that hub's typed answer.

**Round-two verdict findings fixed** (verdict against `32597112c`, edited in place 13:34):

- **High — panic on send to closed channel during reconnect.** Delivery is now scoped to the owning client (`routeNotification`/`publishHostNotification` take the client), and teardown is signalled on a dedicated `clientDone` channel; `sub.in` is never closed, so a send can never race a close and one connection's traffic never leaks into another's subscription. Regressions: `TestRemoteHubSourceHostNotificationScopedToOwningClient`.
- **Medium — pump blocked on a full `sub.out`, so teardown was unobservable.** The pump selects on `clientDone` in both selects, so client teardown always ends the pump and closes the returned channel. Regression: `TestRemoteHubSourceHostSubscriptionClosesWhenClientStreamEndsWhilePumpBlocked`.
- **Medium — non-config notifications evicting config updates.** The fan-out installs a publish-time filter (`SetHostNotificationFilter`) so only the config methods it owns occupy the bounded host buffer; the relay's own method check stays as defense in depth.
- **Low — fixed 1s SSH retry churn.** `fanOut` no longer drives the component-04 connector while `RemoteHubSource.Online()` is false (the supervisor owns reconnecting) and retries with bounded exponential backoff (1s → 30s). Regression: `TestHostAdminFanOutLeavesReconnectToTheSupervisorWhileOffline`.

Verification: `gofmt -l cmd/evener-hub/` empty; `go build ./...` exit 0; `go vet ./cmd/evener-hub/...` exit 0; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` ok 15.1s; `go test ./cmd/evener-hub/ -race -count=1 -timeout 40m` ok 423.6s; `go test ./cmd/evener-hub/... -count=1` ok (13 packages); `GOLANGCI_LINT_CACHE=… MODULES="." go run ./cmd/evener-dev/bin dev module-lint` PASS (1 module, 89s).

obra:
Round three addressed (`d2b89a7`).

- **Medium — retry safety of forwarded mutations** (`remote_hub_source.go:276-289`): **fixed.** The finding is real: `AdminCall` mapped a lost response to `SessionUnavailable` for every allow-listed method, including the non-idempotent ones, so a caller could not tell a refusal from a mutation the remote may have applied before losing the answer. The allow-list is now split: `remoteHostAdminMutationMethods` (`app_host_admin.go`) names the non-idempotent half (instance create/edit/remove/setDefault, launch setLayer/trustRepo, marketplace add/remove/edit, plugin install/upgrade/remove/enable/disable/setAutoUpgrade, auth login/logout/apiKey/credentialJson/device, settings agentsDoc/set, dirs/create), and the proxy forwards those through the new `RemoteHubSource.AdminMutationCall`, which turns a lost response into `ErrorMutationOutcomeUnknown` + `MutationOutcomeUnknown` + `RetryDispositionBlocked`. It is `blocked` rather than `mutationCall`'s `automatic` deliberately: a forwarded thread mutation carries a `clientMutationId` the remote dedups on, but `HostRequestParams` carries no such id and no admin method dedups, so no automatic retry is honest. Read-only forwards keep `AdminCall`'s `SessionUnavailable`. Pinned by `TestHostAdminRequestMapsLostResponseByMutationSafety` (fails before the fix: the mutation loss returned `-32014`) and `TestRemoteHubSourceAdminMutationCallMapsResponseLossToOutcomeUnknown`; `TestHostAdminMutationClassificationMatchesAllowList` forces a retry-safety decision for every allow-listed method, so a future allow-list row cannot silently inherit the read mapping. `HostRequestParams`' doc now states the contract. I took the explicit outcome-unknown error rather than adding a mutation id: the proxy carries no idempotency key today and no browser caller exists yet to supply one.
- **Low — hot-path allocation in `publishHostNotification`** (`remote_hub_source.go:361-371`): **fixed, with one correction.** The filter now short-circuits before the subscription snapshot, so a rejected notification no longer walks and appends the subscription set on the shared drain goroutine. The "heap allocation" part is not reproducible, though: the snapshot slice does not escape (`make([]*remoteHubHostSubscription, 0, len(s.hostSubs)) does not escape` under `-gcflags=-m`), so it is stack-allocated, and `testing.AllocsPerRun` over the pre-fix reject path measures 0 allocations/op. The saving is CPU (no map walk plus append per rejected notification), not GC. `TestRemoteHubSourceHostNotificationFilterShortCircuitsBeforeSnapshot` pins the filter contract: a rejected notification is not delivered, is not counted as a drop, leaves the reject path allocation-free, and an accepted one still delivers.

Verification: `gofmt -l appwire cmd/evener-hub` empty; `go build ./...` ok; `go vet ./cmd/evener-hub/...` ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` ok (13.5s); `go test ./cmd/evener-hub/ -race -count=1 -timeout 40m` ok (382.977s); `go test ./cmd/evener-hub/... -count=1` ok; `module-lint` PASS (1 module, 81s, 0 findings). Pushed `d2b89a769`, a fast-forward from `33a34183e`.

obra:
Round four addressed (`d6ed5c805`).

- **Medium — `cmd/evener-hub/internal/appsource/remote_hub_subscription.go` (lifecycle deadlock).** Confirmed. The per-client drain goroutine blocked on `sub.in <- translated` while a thread consumer had backpressured delivery, and that send did not select on `client.Notifications()`; a client teardown arriving during the block was therefore never observed, the drain's deferred cleanup never ran, `clientDone` was never closed, and the 07a host fan-out stayed attached to a dead client across reconnect (no host notification after rebind). Fixed by reading ahead while a thread send is backpressured: `drainLoop` now publishes each notification to host-level consumers immediately (still non-blocking and in read order) and buffers thread notifications in a local FIFO, and `routeNotification`'s blocking send also selects on the client's own stream, returning so the drain's deferred teardown runs. Thread routing still blocks rather than dropping, exactly as before — no thread notification is dropped. Pinned by the failing-first `TestRemoteHubSourceHostSubscriptionClosesWhenThreadDeliveryBackpressures` (fails in 5.01s on `d2b89a7`, passes now).

- **Medium — `cmd/evener-tui/hub_notification_coverage_test.go` (deterministic gate).** Confirmed: `TestEveryWireNotificationIsHandledOrExplicitlyIgnored` failed with `wire notifications neither dispatched nor explicitly ignored by the TUI: [evener/host/notification]`. Added `appwire.NotifyEvenerHostNotification` to `notifyMethodsDeliberatelyIgnored`, with the reason that the wrapper is emitted only for a remote host's config broadcasts and consumed by the Web UI's per-host settings panes, while the TUI has no remote-host settings surface (a local config change still arrives unwrapped as the underlying method and is dispatched). The catalog/response agreement is preserved deliberately, not by weakening the gate.

Verification: `gofmt -l` over touched dirs → empty; `go build ./...` → 0; `go vet ./cmd/evener-hub/... ./cmd/evener-tui/...` → 0; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3` → ok (13.2s); `go test ./cmd/evener-hub/ -race -count=1` → ok (488.4s); `go test ./cmd/evener-hub/... ./cmd/evener-tui/... -count=1` → ok; `module-lint` → PASS (0 findings). No generated output moved, so `make lint-generated` was not required.

obra:
Round five addressed (`af98cbcab`).

**High — unbounded `pending` read-ahead in the drain loop** (`remote_hub_subscription.go`, `drainLoop`/`routeNotification`): **fixed.** Confirmed: while a thread consumer was backpressured, `routeNotification`'s read-ahead select appended every notification consumed from `client.Notifications()` to `pending` with no cap, and the drain never stops reading, so a stuck relay could grow the controller's heap without limit. `pending` is now bounded by `remoteHubDrainReadAheadCap` (`remoteHubSubBuffer * 4` = 512): at the cap the drain keeps reading the client's stream but drops the overflow and counts it in `RemoteHubSource.threadReadAheadDropped`. **This supersedes one clause of round four's disposition:** round four said thread routing "blocks rather than dropping … no thread notification is dropped." That remains true on the normal path and below the cap; past the cap a notification *is* dropped, and counted. The review's other suggested policy — stop reading at the cap and select only on `sub.in`/`sub.pumpDone` — was rejected deliberately because it re-breaks round four: stopping the read hides a client teardown behind the stalled send, so the drain's deferred cleanup never runs, `clientDone` is never closed, and the 07a host fan-out stays bound to a dead client across reconnect (`TestRemoteHubSourceHostSubscriptionClosesWhenThreadDeliveryBackpressures` is exactly that scenario). How the bound composes with round four: the cap only limits how much the read-ahead *buffers*; the select still includes `<-client.Notifications()`, so a teardown is still observed on every iteration, including while parked at the cap. Pinned by `TestRemoteHubSourceDrainReadAheadIsBounded` (fails against `d6ed5c8`: `drain delivered all 1536 pushed notifications; its read-ahead FIFO is unbounded`) and `TestRemoteHubSourceDrainReadAheadDropsAreCounted` (pre-fix: `no read-ahead drop was counted`). Both drive the real stall — a subscribed thread whose consumer is never read, then a burst, with a host-level sentinel proving the drain consumed the whole batch before the count is measured — and the round-four guard test still passes.

**Medium — `AdminMutationCall` mapped pre-dispatch failures to outcome-unknown** (`remote_hub_source.go:319-324`): **fixed.** Confirmed: `remoteHubAdminMutationCallError` was applied to the `s.client(ctx, s.id)` error, but a connector failure (offline, refused attach, dial) happens before `client.Request` is ever called, so no request frame was sent and the mutation provably did not happen. The acquisition failure now goes through `s.mapCallError` exactly as on `AdminCall` (a dead channel becomes `SessionUnavailable`/`CodeUnavailable`, a safe retry); `remoteHubAdminMutationCallError` is reserved for the `client.Request` failure, which is the only one that can be a lost response. Pinned by `TestRemoteHubSourceAdminMutationCallClientFailureIsNotOutcomeUnknown` (fails pre-fix: `code = -32603, want -32014`). Round three's retry-safety scoping is intact: `TestRemoteHubSourceAdminMutationCallMapsResponseLossToOutcomeUnknown` (a genuine lost response → `ErrorMutationOutcomeUnknown` + `MutationOutcomeUnknown` + `RetryDispositionBlocked`) and the semantic-refusal passthrough still pass unchanged.

**Low — raw result passthrough alters large integers via `float64`** (`remote_hub_source.go:288-290`): **contested as unreachable through this PR, and outside this unit's file ownership to fix.** The mechanism is real in general — `Client.Request` round-trips the response through `json.Marshal(msg.Response.Result)` (`appwire/client.go:287`), and `Response.Result any` is decoded by plain `json.Unmarshal` (`appwire/jsonrpc.go`), which yields `float64` for JSON numbers — but no integer that can exceed 2^53 (≈9.0e15) is reachable through this PR's exact allow-list. Reflecting over every allow-listed method's catalog result type, the only `int64` fields are `MarketplaceEntry.lastUpdated`, `PluginEntry.installedAt`, and `PluginEntry.lastUpdated`, all written through `hubcore.UnixSeconds` (epoch **seconds**, ~1.7e9); every other integer field is an `int` count/limit (token counts, plugin counts, `schema`, `intervalSeconds`). There is no nanosecond or near-`uint64`-limit field anywhere in the catalog. Byte-exactness would require preserving the raw result bytes in `appwire/client.go` / `appwire/jsonrpc.go` — files this PR does not touch and that every appwire caller depends on — so it belongs in its own appwire change rather than the 07a proxy PR; I did not make that cross-cutting edit. Concrete fix for that follow-up: retain the raw `result` bytes on the decoded frame (e.g. a `RawResult json.RawMessage` populated by `Response`/`Message.UnmarshalJSON`) and have `Client.request` prefer them when `out` is a `*json.RawMessage`.

Verification: `gofmt -l cmd/evener-hub/` → empty; `go build ./...` → 0; `go vet ./cmd/evener-hub/...` → 0; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 20m` → ok (16.690s); `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (all packages; `cmd/evener-hub` 150.236s); `GOLANGCI_LINT_CACHE=… MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (1 module, 81s, 0 findings). Pushed `af98cbcab`, a fast-forward from `d6ed5c805`.

obra:
Round six addressed (`6f0b91779`).

- **Medium — "Shared read-ahead drain loses terminal state under backpressure": FIXED.** The drain no longer keeps one shared read-ahead FIFO. Each `remoteHubSubscription` now owns its own bounded read-ahead buffer (`sub.in`, `remoteHubDrainReadAheadCap`), and `routeNotification` delivers with a non-blocking send, so the drain never parks on a slow consumer and a backpressured thread can no longer evict another thread's notifications (cross-thread contamination is gone). On overflow the drain no longer silently drops while leaving the subscription open: it counts the overflow and resets only the saturated subscription (`sub.reset()` cancels that one subscription's context), so its pump closes `out`, the controller relay observes subscription end, and `app_relay.go` runs its recovery path (fresh thread read + `evener/thread/resync`). The round-five memory bound is kept — strictly fewer than the pushed burst is delivered — and round four's property (the drain never blocks, so a client teardown is always observed) is preserved and now structural. Pinned by `TestRemoteHubSourceDrainOverflowResetsStalledSubscription` (the stalled subscription ends, fewer than pushed delivered, overflow counted) and `TestRemoteHubSourceDrainOverflowIsPerSubscription` (healthy T receives every notification while S overflows). The two round-five drop-policy tests are superseded by this finding and were rewritten to the new policy — they asserted the silent drop the finding rejected; the bound assertion and the counter-observability assertion are retained.

- **Low — "Fan-out goroutines not lifecycle-bound" (`app_rpc.go:391`): CONTESTED — needs scope beyond this PR's files.** The finding is correct: `registerHostAdminHandlers(context.Background(), …)` gives the fan-outs a context nothing ever cancels. But every hub background worker in this codebase is bound to the process context owned by `main.go` (`runNavigationPublisher(ctx, …)`, `web.navigation.Start(ctx)` → `stopLifecycle`, `startBackground(…)`), and `appserver.Server` exposes no shutdown hook; the hub's `WebServer` (and `appRPC`) is constructed before `main.go`'s `ctx` exists. A correct fix therefore touches `cmd/evener-hub/main.go` + `cmd/evener-hub/web.go` in addition to `app_rpc.go`, which are outside this PR's file set (stack-collision risk). Ready-to-apply minimal patch: add an unexported `newHubAppServerWithLifecycle(ctx, …)` in `app_rpc.go` (keep `newHubAppServerWithNavigationAndTrace` delegating with `context.Background()` so existing test callers are untouched), have `newWebServer` (web.go) create and own a lifecycle context and call the new constructor, and cancel that context from `main.go`'s shutdown path. Happy to land it in a follow-up once the scope is widened.

Verification: `gofmt -l cmd/evener-hub cmd/evener-hub/internal/appsource` empty; `go build ./...` ok; `go vet ./cmd/evener-hub/...` ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 20m` ok (18.3s); `go test ./cmd/evener-hub/ -race -count=1 -timeout 45m` ok (582.8s); `go test ./cmd/evener-hub/... -count=1 -timeout 30m` ok (227.0s, all packages); `module-lint` PASS (1 module, 0 findings).

obra:
Round seven addressed (`7d5161e9b`).

- **Medium — post-send cancellation bypassed the outcome-unknown mapping** (`cmd/evener-hub/internal/appsource/remote_hub_source.go:331-340`): **FIXED.** `appwire.Client` gives no send-phase signal: a frame write refused because the context had already ended (`StreamTransport.Send`'s own bare `ctx.Err()`, `appwire/stream_transport.go:50-53`) and a canceled response wait after the frame went out (`Client.request`'s `ctx.Done()` select, `appwire/client.go:258-274`) return the same bare error, so the raw `context.Canceled` this path returned read as "nothing happened, retry" whenever the frame *had* reached the host — the blind retry that duplicates a forwarded `instance/create` or `plugin/install`. `AdminMutationCall` now classifies an in-flight context end (`context.Canceled`/`context.DeadlineExceeded`, and never a remote `WireError`) as `ErrorMutationOutcomeUnknown` + `MutationOutcomeUnknown` + `RetryDispositionBlocked`, the same shape a lost response gets, and the pre-call `ctx.Err()` check still returns the raw cancellation for the provably-not-sent case (a safe retry). No idempotency key was added: `HostRequestParams` carries none and no allow-listed admin method dedups.
  - Pinned by the failing-first `TestRemoteHubSourceAdminMutationCallPostSendCancellationIsOutcomeUnknown` (`remote_hub_admin_test.go:257`): the scripted remote records the request and then stays silent, so the cancellation provably lands after the frame is on the wire; the test asserts `ErrorMutationOutcomeUnknown`/`MutationOutcomeUnknown`/`RetryDispositionBlocked` and that the raw cancellation does not escape. Against `6f0b917` it fails with `error = context canceled; the raw cancellation escapes, so a caller can blind-retry a mutation that may have been applied`.
  - `TestRemoteHubSourceAdminMutationCallPostSendDeadlineIsOutcomeUnknown:330` pins the deadline half, with one correction: a post-send deadline already reached outcome-unknown indirectly (`mapCallError` maps `DeadlineExceeded` to `SessionUnavailable`, which the mutation mapping re-labelled). It is now classified explicitly so that no longer depends on the unavailability shape.
  - `TestRemoteHubSourceAdminMutationCallPreCallCancellationStaysRaw:396` and `TestRemoteHubSourceAdminCallPostSendCancellationStaysRaw:417` bound the change: a context already ended before the call sends nothing (asserted against the recorded calls) and keeps the raw error; `AdminCall` (the read path) is untouched.
- **Low — fan-out goroutine leak on shutdown** (`cmd/evener-hub/app_rpc.go:391`): **CONTESTED — the lifecycle handle this needs does not exist inside this PR's file set; here is the exact missing symbol and what would be required.** The mechanism is real (I re-verified it): `registerHostAdminHandlers(context.Background(), …)` hands the fan-out a context nothing cancels, while the fan-out itself tears down correctly when its context *is* canceled (`TestHostAdminFanOutStopsWhenContextCanceled`, `app_host_admin_test.go:722`). What is missing is an owner:
  - `newHubAppServerWithNavigationAndTrace` (`app_rpc.go:238`) takes no context, and its only production caller is `newWebServer` (`web.go:79`, `web.go:124`), which takes none either.
  - `appserver.Server`'s whole exported surface is `NewServer`, `Shutdown`, `Router`, `CommitProjection`, `NewConnection`, `Logf`, `Broadcast`, `BroadcastAll`, `SubscriberCount`, `SetBeforeSubscriptionGate`, `DebugSubscriptions`, `ServeWebSocket`. The only lifecycle entry point is `Shutdown(ctx)` (`internal/appserver/server.go:170`), which cancels connection contexts and waits for handlers. There is no `Done()`, no exported context, no hook registration, and `ServerConfig` (`server.go:17-53`) has no lifetime field.
  - `hubcore.WebConfig` (`internal/hubcore/config.go:37-113`) has no lifetime context or shutdown-hook field, and `appsource.Registry` has no `Close`.
  - `main.go`'s process lifecycle context — `ctx, cancelBackground := context.WithCancel(signalCtx)` (`main.go:483`), canceled and waited at `main.go:484-487` — is created *after* `newWebServer` (`main.go:394`) and is never passed into it.
  Required, minimal, either: **(a)** one new symbol in `internal/appserver` — `func (s *Server) Done() <-chan struct{}` closed when `Shutdown` begins, or a `Lifetime context.Context` field on `ServerConfig` — after which `app_rpc.go` alone owns the binding; or **(b)** thread a context from `main.go` through `newWebServer` → `newHubAppServerWithNavigationAndTrace` → `registerHostAdminHandlers`, which is 3 files (`main.go`, `web.go`, `app_rpc.go`) and was round six's proposal. Both fall outside this PR's file set and outside this unit's ownership; I cannot widen the file set here.
  What I *did* change in-file: the call-site comment no longer claims the fan-out is "like the other background workers" (those are canceled from `main.go`), and now states the real contract, the reason the context is process-lifetime, the fact that the hub RPC server exposes no shutdown signal, and where the fix goes — so the gap is documented at the line rather than implied to be safe.
  Blast radius, for triage: one goroutine (plus its host subscription) per remote host per hub-server *build*; the stale fan-out broadcasts to a server with no connections and cannot starve thread routing (round six's per-subscription read-ahead bound and client-scoped teardown are unchanged) and cannot cross-talk (delivery is scoped to the owning client).
- **Low — fan-out notification not consumed by the Web UI** (`appwire/types.go:3547`, `cmd/evener-tui/hub_notification_coverage_test.go:43-51`): **FIXED as option (a)** — the Go-side comments and assertions are scoped to the fan-out contract and name where the consumer lives; no frontend code was added to 07a. I verified the consumer on `multi-host-pr07b-host-routing` @ `b07d9eda9` before naming it: `cmd/evener-hub/frontend/src/stores/credentials.ts`'s `handleNotification` unwraps a wrapped `evener/host/notification` whose inner method is `evener/auth/updated` into a per-host debounced `scheduleHostRefetch(host)` (never the controller partition), and `cmd/evener-hub/frontend/src/panes/spawn/Spawn.tsx`'s credentials/model-catalog invalidation effect observes the same wrapped method for the selected host. Accuracy note: at that head only `evener/auth/updated` is unwrapped, so the other four families this fan-out can emit still have no consumer — the comments now say exactly that (the wrapper reaches clients that do not yet route it; local changes keep arriving unwrapped, so no existing store changes behavior).
  Changed: `appwire/types.go` (`HostNotificationParams`, and the `NotifyEvenerHostNotification` constant), `appwire/protocol.go`'s notification summary → regenerated `docs/appwire-protocol.md`, `cmd/evener-tui/hub_notification_coverage_test.go:43-51` (the TUI still waives the method — the test keeps pinning that the decision is deliberate), `cmd/evener-hub/app_host_admin.go` (`remoteHostConfigNotifications`).
  Option (b) rejected on semantics: a Go-side unwrapping dispatcher would have to re-broadcast the inner method, destroying the host tag the wrapper exists to carry, so there is no Go-side consumer to add.
- **Low — `HostForwardedResult` erases the proxied result for typed clients** (`appwire/types.go:3545`): **FIXED via the finding's second option (documented cast).** The first option is not reachable from this PR's files: `internal/appwirets` emits a *named interface* for every top-level catalog type — `unknown` is only produced for a *field* typed `json.RawMessage` (`internal/appwirets/emit.go`'s `typeExpr`) — so `MethodTypes["evener/host/request"]["result"]` cannot become `unknown` without changing the generator, which is outside this PR's file set (and would change every empty catalog interface at once). What changed instead: `HostForwardedResult`'s doc states the generated consequence (`{}` for a typed caller), why it is not `unknown`, how the marker should be read, and the required widen-and-cast; the catalog summary carries the same contract into the regenerated `docs/appwire-protocol.md`; and both name the cast helper that already exists in component 07b — `cmd/evener-hub/frontend/src/stores/hostRouting.ts`'s `hostRequest`, which does `result as unknown as MethodTypes[M]["result"]` and is what the frontend's host-scoped calls go through (verified at `b07d9eda9`). `types.gen.ts` was not hand-edited and is unchanged (no Go type changed); `go generate ./appwire/` was run and `make lint-generated` passes.

All prior dispositions are intact: the fail-closed allow-list refusal, the host-tagged fan-out, the mutation-safe error mapping (including round five's provably-not-sent carve-out), and round six's per-subscription read-ahead bound with overflow resetting only the saturated subscription. No existing assertion was weakened or deleted.

Verification: `gofmt -l cmd/evener-hub cmd/evener-hub/internal/appsource cmd/evener-tui appwire internal/appwirets` empty; `go build ./...` ok; `go vet ./cmd/evener-hub/...` ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 20m` ok (22.8s); `go test ./cmd/evener-hub/... -count=1 -timeout 30m` ok (every package, 3m26s); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` PASS (1 module, 0 findings, 112s); `make lint-generated` PASS; `make test-web` PASS (`web-typecheck`, `web-test`, `web-lint`; 26m14s under the shared host load). Extra check for the one non-hub test file this change edits: `go test ./cmd/evener-tui/ -run TestEveryWireNotificationIsHandledOrExplicitlyIgnored -count=1` ok.

obra:
Round eight addressed (`1f31dd11`).

- **Medium — fan-out workers leaked across server shutdown/recreation** (`cmd/evener-hub/app_rpc.go`): **FIXED.** Claim verified against the code: the fan-out was started with `context.Background()`, which nothing cancels, so a hub server shut down or recreated in-process left the previous server's fan-out alive — one goroutine per remote host, each still retaining the old server's sources and broadcaster, and each duplicating every host notification once a replacement subscribed too. Fix: the RPC server now exposes the lifetime handle the round-seven comment identified as missing, and `app_rpc.go` binds the fan-out to it (`registerHostAdminHandlers(server.Lifetime(), …)`). Pinned by `TestHostAdminFanOutStopsWhenServerShutdown` (`cmd/evener-hub/app_host_admin_test.go`), which proves the worker subscribes, that `Shutdown` cancels the handle and drops the host-notification subscription count from 1 to 0, and that a server recreated over the same source registry ends with exactly one subscriber rather than two; plus `TestServerLifetimeCanceledByShutdown` and `TestServerLifetimeCanceledWhenShutdownGivesUp` in `internal/appserver/server_test.go`. Failing-first: with the binding reverted to `context.Background()` the hub test fails after its deadline with `the fan-out stayed subscribed after its server was shut down: host-notification subscribers = 1, want 0`.
- **Medium — closed-pipe mutation failures escaped outcome-unknown classification** (`cmd/evener-hub/internal/appsource/remote_hub_source.go`): **FIXED.** Claim verified: `transportUnavailable` listed `EPIPE`, `io.EOF` and `io.ErrUnexpectedEOF` but not `io.ErrClosedPipe`, and its text fallback matched "broken pipe", not "closed pipe", so a write refused because the stream was already gone reached `AdminMutationCall`'s classifier as an unrecognized error and escaped raw — which reads as "nothing happened, retry" for a forwarded `instance/create` or `plugin/install` that may have been applied before the write side closed. Fix: `io.ErrClosedPipe` is now transport loss (via `errors.Is` and the text fallback), which `remoteHubAdminMutationCallError` re-labels to `ErrorMutationOutcomeUnknown` + `MutationOutcomeUnknown` + `RetryDispositionBlocked`; `AdminCall`'s read path keeps `SessionUnavailable`, the safe retry for an idempotent read. Pinned by `TestRemoteHubSourceAdminMutationCallMapsClosedPipeWriteToOutcomeUnknown` (bare and wrapped `io.ErrClosedPipe`, driven through a transport whose `Send` refuses the write) and `TestRemoteHubSourceAdminCallClosedPipeWriteStaysSessionUnavailable`. Failing-first: with the classification reverted, both fail with the raw `io: read/write on closed pipe`.

**Lifecycle handle added:** `func (s *appserver.Server) Lifetime() context.Context` in `internal/appserver/server.go`, canceled at the top of `Shutdown` — before it waits for handlers, so an expired shutdown deadline cannot suppress the signal (that ordering is itself pinned). Path note: the authorized widening named `cmd/evener-hub/internal/appserver/`, which does not exist in this tree; `appserver` is the shared `internal/appserver` package that `cmd/evener-hub` imports, and that is where the handle went. It is strictly additive: the context is never nil and is canceled only by `Shutdown`, so a server nobody shuts down behaves exactly as before. **Intended adopters beyond component 07a:** any background consumer owned by the RPC server rather than by the hub process — hub embedders and tests that recreate a server in-process, the appwire-trace drain path (`main.go` already calls `appRPC.Shutdown` there), and later per-server workers. The hub's process-lifetime workers in `main.go` stay bound to `main`'s context, which process shutdown cancels. One residual, outside this unit's file set, worth naming: the hub's ordinary `serveHub` path shuts the HTTP server down without calling `appRPC.Shutdown`, so the handle fires today in the paths that do shut the RPC server down (trace drain, embedders, tests) while process exit covers the rest; wiring `appRPC.Shutdown` into `serveHub` would extend it to the plain-exit path. The hub test also uses one new read-only accessor, `RemoteHubSource.HostNotificationSubscribers()`, which reports the live host-level subscription count (the observable form of "a fan-out is still subscribed"); it is additive and changes no behavior.

Prior dispositions are intact and no existing assertion was weakened or deleted: the fail-closed allow-list refusal, the host-tagged fan-out filter, mutation-safe error mapping including round seven's post-send cancellation classification and its pre-call raw-cancellation carve-out, and the per-subscription read-ahead bound and overflow policy.

Verification: `gofmt -l internal/appserver cmd/evener-hub cmd/evener-hub/internal/appsource` → empty; `go build ./...` → 0; `go vet ./cmd/evener-hub/...` → 0; `go test ./cmd/evener-hub/internal/appsource/ -race -count=3 -timeout 20m` → ok (12.998s); `go test ./internal/appserver/ -count=1` → ok (2.213s) — the gate's name is `cmd/evener-hub/internal/appserver/`, which does not exist in this tree, so it was run against the real package path; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok, exit 0 (hub 152.351s, all internal packages ok); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS lint (1 modules, 79s), 0 findings.

obra:
Round eight addressed (`b72470fab`).

- **Medium — pre-send cancellation misclassified as outcome-unknown (`remote_hub_source.go:370-371`): FIXED.** Confirmed in `appwire/client.go`: `request` registers its pending entry, then blocks on `sendMu` before `transport.Send`, so a caller context that ends while the call is queued there (two admin RPCs in flight on one host is enough — they share one client) was reported as a bare `context.Canceled`, indistinguishable from a post-dispatch loss. `remoteHubAdminMutationCallError` therefore mapped it to outcome-unknown/blocked for a mutation whose frame was never written — retries blocked for a request that provably never reached the host. Changed at the boundary the verdict names: `appwire.Client.request` now reports the one window it can prove — a context already ended when the request reaches the write point — as the new typed `appwire.RequestNotSentError`. Nothing is handed to `transport.Send` on that path, which is what makes the "never transmitted" claim sound, and the wrapper keeps the cause as its `Unwrap` target, so `errors.Is(err, context.Canceled)` still holds. `remoteHubSource.remoteHubAdminMutationCallError` maps it exactly as its own pre-call check does — raw cancellation, or SessionUnavailable for an expired deadline, both safe retries — and never outcome-unknown. Everything from dispatch onward is untouched: a transport write failure and a post-dispatch cancellation still map to outcome-unknown/blocked, and a transport that refuses the write a moment *after* dispatch is deliberately not marked not-sent, because past dispatch the client cannot prove the frame never left. Pinned by `TestRemoteHubSourceAdminMutationCallPreSendCancellationStaysRetryable` (failing first: `error = appwire.WireError mutation outcome is unknown after the caller's context ended while the remote hub call was in flight: host, want the caller's own cancellation (a safe retry)`; it drives the real `sendMu` queue with a gated transport and requires exactly one frame to have been offered to the transport — the queued holder's own), plus `TestClientRequestReportsAPreSendCancellationAsNotSent` and `TestClientRequestKeepsAPostSendCancellationUnmarked` (`appwire/client_presend_test.go`). Round seven's post-send contract keeps its existing pins (`…PostSendCancellationIsOutcomeUnknown`, `…PostSendDeadlineIsOutcomeUnknown`, `…PreCallCancellationStaysRaw`) — none weakened or changed.
- **Medium — `os.ErrClosed` not recognized as transport loss (`remote_hub_source.go:201-214`): FIXED.** Confirmed and reproduced with the real shape rather than a stand-in: the SSH stdio transport writes frames into the child's stdin pipe (`cmd/evener-hub/internal/sshconn/runner.go`'s `cmd.StdinPipe()`), and the first `StreamTransport.Send` after the manager closes that channel reports `*fs.PathError` wrapping `os.ErrClosed` ("file already closed"), which escaped `transportUnavailable` as a raw error — read as "nothing happened, retry" for a forwarded instance/create or plugin/install that may have been applied. (`net.ErrClosed` was already caught, but only by the textual fallback.) Changed: `errors.Is(err, os.ErrClosed)` and `errors.Is(err, net.ErrClosed)` are now in the recognized set (bare or Go-wrapped), and "file already closed" joined the textual fallback, so the `WireError` path — which has only the peer's message to match on — classifies the same failure identically instead of disagreeing with the `errors.Is` path. Pinned by `TestRemoteHubSourceClosedFileWriteIsTransportLoss`, covering bare and wrapped `os.ErrClosed`, bare and wrapped `net.ErrClosed`, and a real `appwire.StreamTransport` over an `os.Pipe` closed the way the SSH manager closes a host's channel; each asserts outcome-unknown/blocked on the mutation path and SessionUnavailable on the read path. Failing first on three of the five shapes (`file already closed`, `appwire send: file already closed`, `write |1: file already closed`); the two `net.ErrClosed` cases passed before the change because the textual fallback already matched "use of closed network connection" — they are included because the fix now recognizes that sentinel explicitly, as asked.
- **Medium — fan-out lifetime tied to a conditionally-shut-down `appRPC` (`app_rpc.go:402`, `main.go:623-630`): FIXED — round six's contest is resolved by fixing it.** Confirmed: `web.appRPC.Shutdown` was registered only when `--appwire-trace` was set, so an ordinary hub run returned from `serveHub` with the RPC server's lifetime still open — one fan-out goroutine plus its host subscription per remote host stayed alive, retaining the old server's sources, and duplicating host notifications once a replacement server subscribed. Round six raised this as a Low and the unit contested it: the finding was accurate, but the fix was outside that round's file set — round six's contest said a correct fix "touches `cmd/evener-hub/main.go` + `cmd/evener-hub/web.go` in addition to `app_rpc.go`, which are outside this PR's file set (stack-collision risk)" and named the missing symbol. Round seven added that handle (`appserver.Server.Lifetime()`), which is what made the in-file binding possible, and named this residual — "the hub's ordinary `serveHub` path shuts the HTTP server down without calling `appRPC.Shutdown` … wiring `appRPC.Shutdown` into `serveHub` would extend it to the plain-exit path". The scope widening is granted for this round, so the item is fixed rather than contested again. Changed: the drain is now an unconditional top-level `defer` in `runMain` — every exit path, tracing or not — replacing the trace-gated block, and it is registered after the SSH manager's teardown so it runs first: the fan-outs stop while the transports they read from are still open. `serveHub` only knows the `hubHTTPServer` interface, so the top-level hook reaches the plain-exit path without changing that seam (round seven suggested `serveHub`; this covers that path and every other return path too, inside the authorized files). Ordering and idempotency were verified against the actual shutdown path, not assumed: `appserver.Server.Shutdown` cancels the lifetime first, before it waits for handlers, so an expired shutdown deadline cannot suppress the signal; `context.CancelFunc` is idempotent and the drained-close is guarded by `webSocketShuttingDown`, so a second `Shutdown` — the tracing drain, an embedder, or a test — is safe and returns nil. On the fan-out side the lifetime cancel ends `relayHostNotifications`, the subscription's pump closes and unregisters while the transport is still open, and `fanOut`'s loop exits on `ctx.Err()` before it can re-subscribe, so a transport closed afterwards cannot resurrect it; an in-flight broadcast during the drain cannot panic either (`BroadcastAll` snapshots connections under a read lock and enqueues without touching closed channels). Pinned by `TestRunMainShutsDownAppRPCWithoutTracing` (failing first: `the AppWire server outlived the hub: its lifetime was never canceled, so every fan-out bound to it stays subscribed`; it then calls `Shutdown` a second time and requires nil) and `TestHostAdminFanOutStopsBeforeItsTransportCloses` (host-notification subscribers 1 → 0 across `Shutdown` while the transport is still open, then the transport closes and the connector is never called again across a full retry interval; a second `Shutdown` stays safe). Round seven's `TestHostAdminFanOutStopsWhenServerShutdown` and `TestHostAdminFanOutStopsWhenContextCanceled` still pass unchanged.
- **Low — `HostForwardedResult` doc claims unlanded code exists (`appwire/types.go:3554-3558`): FIXED.** Confirmed: `cmd/evener-hub/frontend/src/stores/hostRouting.ts` does not exist in this commit — the only occurrence of that name in the tree was this comment. The comment now states that a caller has to widen the result and cast it to the method it forwarded, that nothing does so yet, and names component 07b (branch `multi-host-pr07b-host-routing`) as a forward reference, matching `HostNotificationParams` next door. No path, function, or cast until 07b lands.

Prior dispositions are intact and no existing assertion was weakened or deleted: the fail-closed allow-list refusal (including `evener/git/head`), the host-tagged fan-out filter, the mutation-safe error mapping (round five's provably-not-sent carve-out, round six's per-subscription read-ahead bound with overflow resetting only the saturated subscription, round seven's post-send cancellation classification and its pre-call raw-cancellation carve-out), and the remote-admin proxy's behavior. Round eight's pre-send carve-out extends round five's provably-not-sent rule to the client's own write queue; the read path is unchanged.

**Verification:** `gofmt -l appwire/ cmd/evener-hub/ cmd/evener-hub/internal/appsource/` → empty; `go build ./...` → 0; `go vet ./cmd/evener-hub/...` → 0; `go test ./cmd/evener-hub/internal/appsource/ -race -count=1 -timeout 20m` → ok (4.9s); `go test ./appwire/ -race -count=1` → ok (2.1s); `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (13 packages, 156.9s); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS, 0 findings.

obra:
Round nine addressed (`25de908a3`).

- **Medium — `evener/plugin/checkNow` classified read-only though it can mutate plugins** (`cmd/evener-hub/app_host_admin.go:135-163`): confirmed. The handler runs `runPluginAutoUpgradeTick` (`app_plugin_autoupgrade.go:54`), which refreshes every marketplace clone and then installs the new content of every opted-in git-backed plugin via `Manager.UpdateAutoUpgrade` (`internal/plugins/autoupgrade.go:59`). The catalog description already said `per plugin actually upgraded` (`appwire/protocol.go:188`), and `TestRegisterPluginAutoUpgradeHandlers_CheckNowRunsOneTick` exercises exactly that upgrade. Fix: `MethodEvenerPluginCheckNow` added to `remoteHostAdminMutationMethods` and removed from the classification test's read-only set (both comment blocks updated), so the proxy forwards it through `AdminMutationCall` and a lost response reports `ErrorMutationOutcomeUnknown` / `RetryDispositionBlocked` instead of a retryable `SessionUnavailable`. Pinned by the new `checkNow reports unknown outcome` subtest in `TestHostAdminRequestMapsLostResponseByMutationSafety`: against the old classification it fails with `code = -32014` (SessionUnavailable), and it passes with the fix.
- **Medium — cancellation checked only after the uninterruptible `sendMu.Lock()`** (`appwire/client.go:281-290`): confirmed reachable. `StreamTransport.Send` does a blocking `rw.Write` with no `ctx` check and `StreamTransport` is not a `Pinger`, so a wedged SSH stdio write parks the write slot indefinitely while a request queued behind it could not observe its own cancellation. Fix: the slot is now a one-token channel semaphore (`sendSlot`; `acquireSendSlot` selects on `ctx.Done`, `releaseSendSlot` returns it) and `request`'s acquisition is context-aware. Round-eight contract preserved: nothing is handed to `transport.Send` on the not-sent path (the post-acquire `ctx.Err()` re-check still covers the both-ready select), a post-dispatch cancellation or write failure stays unmarked/outcome-unknown, `errors.Is(err, context.Canceled)` holds through `RequestNotSentError.Unwrap`, and `Notify` keeps a blocking acquisition on the same slot so writes remain serialized (one writer at a time). Pinned by `TestClientRequestQueuedBehindAStuckWriteReturnsNotSentOnCancel` (pre-fix: hangs, fails at the 2s timeout) and `TestClientWritesStaySerializedOnTheSharedSlot` (8 concurrent `Notify`/`Request` writers; peak in-transport concurrency asserted to be 1). Round-eight pins `TestClientRequestReportsAPreSendCancellationAsNotSent`, `TestClientRequestKeepsAPostSendCancellationUnmarked`, and `TestRemoteHubSourceAdminMutationCallPreSendCancellationStaysRetryable` pass unchanged.

Verification: `gofmt -l appwire cmd/evener-hub` → empty; `go build ./...` → ok; `go vet ./cmd/evener-hub/...` → ok; `go test ./appwire/ -race -count=3` → ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=1 -timeout 20m` → ok; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (all packages); `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (0 findings).

obra:

The parent components (04b, 05a, 05b, 05c, 05d, 06a, 06b) landed, so this component's own 12 commits were rebased onto `main` (`git rebase --onto origin/main d324442e3`) and force-pushed with `--force-with-lease`. The PR's diff is now exactly this component's, against `main`.

This rebase needed real reconciliation rather than only conflict marking: `main` has since landed the components this one builds on (05a-05d), and both sides evolved the same drain/subscription code in parallel. Each decision keeps `main`'s landed behavior and re-pins this component's intent to it:

1. **The drain and subscription internals keep `main`'s implementation.** `main` already carries 05b's landed version of `drainLoop`/`routeNotification`, which reaches the same design this branch's commits 5-7 arrived at independently: a per-subscription bounded read-ahead buffer, a non-blocking hand-off so no thread consumer can park the drain, and a saturated subscription retired (reset) so its relay re-reads the thread rather than silently losing terminal state. This branch's intermediate shapes (a shared read-ahead FIFO with `pending`, later replaced by its own per-subscription reset) were therefore resolved to `main`'s implementation, which its own tests already pin (`TestRemoteHubRouteStallRetiresSubscriptionAndKeepsOthersFlowing`, `TestRemoteHubStalledSubscriptionCannotOverflowSharedClient`, `TestRemoteHubSubscribeThreadDrainPreventsOverflow`). This component's additions that do not collide are kept: `AdminCall`, `SubscribeHostNotifications`, `publishHostNotification`, the `hostSubs` field, and the host-level publish inside `drainLoop`.
2. **The teardown is a union.** `main`'s stranded thread-subscription cancel and this component's orphaned host-subscription signal (`close(sub.clientDone)`) both run in the drain's deferred cleanup.
3. **`remote_hub_subscription_test.go` returns to `main`'s content**, because the tests this branch added there pin the shared-FIFO mechanism `main`'s implementation replaced; the behaviors they covered are pinned by `main`'s tests listed above. The file is consequently no longer part of this PR's diff.
4. **A test-helper name collision is resolved in this component's favor of `main`'s name.** Both sides added a `newScriptedRemoteClient` helper with different signatures in `cmd/evener-hub` (`main`'s single-use read-path client in `remote_hub_source_rpc_test.go`; this component's scripted host-admin client in `app_host_admin_test.go`). This component's helper is renamed `newScriptedAdminClient` (six call sites), so `main`'s helper keeps its name.
5. **Union merges** where both sides added: the `appwire` imports (`os` alongside `reflect`/`slices`), the `cmd/evener-hub` test functions (`main`'s `TestRemoteHostFactsReportRestartedHubVersion` and `TestHubSSHStateInvalidation` kept alongside this component's `TestRunMainShutsDownAppRPCWithoutTracing`), and the `remote_hub_source.go` struct fields (`hostSubs` added next to 05a/05b's fields, with 05d's newer `probeMu` comment).
6. **Two integration fixes on top of the replay**, both found by the gate run on the rebased tree:
   - `appwire.ErrStreamClosed` is now recognized as a write-side transport loss. The landed stream transport reports its closed (and poisoned) state with that sentinel, where this component's classifier was written against the raw descriptor error; unclassified, a mutation onto a host channel torn down mid-write escaped as a raw, blind-retryable error instead of outcome-unknown. This component's own `TestRemoteHubSourceClosedFileWriteIsTransportLoss` pins it (its `ssh stdio write end closed` case was red before the fix).
   - The host-admin policy table (which enumerates every ScopeHub catalog method with a deliberate allow/deny decision, so an undecided method fails the coverage check) gained deny rows for `evener/instance/refreshModels` and `evener/instance/setModelDisabled` - two instance methods that landed on the catalog after the table was written. Denied, matching the proxy's allow-list of the five instance handlers the settings panes drive remotely; the catalog-growth comment in `app_host_admin.go` is corrected to match.

**Gates on the rebased tree**: `gofmt -l` clean over the touched Go directories; `go build ./...` exit 0; `go vet ./cmd/evener-hub/...` exit 0; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=2` ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=1` ok; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` ok; module-lint PASS (0 findings).

obra:
Round ten addressed (`3cc50cf30`).

- **Medium — `AdminCall`/`AdminMutationCall` mapped the acquisition failure through `mapCallError` with no post-acquisition context guard** — `cmd/evener-hub/internal/appsource/remote_hub_source.go`. Both paths now mirror `call`/`mutationCall`: when `s.client(...)` fails, an ended caller context returns raw (`if cerr := ctx.Err(); cerr != nil { return cerr }`) and everything else classifies through `mapConnectError`, so a timed-out attach (an `ErrSSHStart` chain that also satisfies `errors.Is(err, context.DeadlineExceeded)`, which `mapCallError` handed back raw) becomes `CodeUnavailable`/`SessionUnavailable` and the auto-resume gate recognizes the host. Pinned by the new `admin read path` and `admin mutation path` subtests of `TestRemoteHubAttachHandshakeTimeoutIsSessionUnavailable` - red before the fix with `error = *fmt.wrapErrors sshconn: ssh channel start failed: host "host" initialize: context deadline exceeded, want appwire.WireError` - and by the new `TestRemoteHubAdminCallerContextStaysRaw`, which pins that the guard keeps a genuine caller deadline raw on both admin paths (also red before the fix: `err = *errors.errorString attach failed`).
- **Low — the pre-dispatch comments claimed a mapping the code does not have** — `remote_hub_source.go`. The three places that said a pre-dispatch expired deadline maps to `SessionUnavailable` now state that it stays raw, as `mapCallError` does and `TestRemoteHubMutationCallerContextStaysRaw` requires.
- **Low — the host id was normalized for one lookup and not the other** — `cmd/evener-hub/app_host_admin.go`. `Request` now trims the host once and uses that value for the registry lookup, the source lookup, and the error text, so `" m4 "` is forwarded like `m4` instead of passing the registry check and failing as `not attached`. Pinned by the new `TestHostAdminRequestNormalizesHostWhitespace`.

Verification: `gofmt -l` over the touched directories clean; `go build ./...` exit 0; `go vet ./cmd/evener-hub/...` exit 0; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=2` ok; `go test ./cmd/evener-hub/internal/appsource/ -race -count=1` ok; `go test ./cmd/evener-hub/... -count=1 -timeout 30m` ok (hub package 129.0s); module-lint PASS (0 findings).

obra:

**CI** — all required checks green at `3cc50cf30` (the full set, including RoboRev's own `roborev` check).

**Review** — RoboRev's verdict for `3cc50cf30`: **"No issues found."** The rebase onto `main` and its reconciliation decisions (the landed drain/subscription internals kept where both sides had evolved them; this component's host-subscription broker, admin proxy, and host fan-out kept; the two integration fixes for the stream sentinel and the allow-list table) are documented in the rebase-note comment above, and the round-ten review's Medium (the acquisition mapping on the forwarded-admin paths) and two Lows are fixed and pinned by tests in `3cc50cf30`.

**Up-to-date override** — the branch is 4 commits behind `main`. Overlap check, run immediately before this comment:

```
$ gh pr view 1361 --json mergeable,mergeStateStatus --jq '.mergeable, .mergeStateStatus'
MERGEABLE
BEHIND

$ comm -12 <(gh pr view 1361 --json files --jq '.files[].path' | sort) \
           <(git diff --name-only "$(git merge-base origin/main origin/multi-host-pr07a-host-admin)" origin/main | sort)

$ git diff --name-only "$(git merge-base origin/main origin/multi-host-pr07a-host-admin)" origin/main | wc -l
11
```

`main` changed 11 files since the merge base; none of them is a file this PR touches.

PR #1362, opened 2026-09-15
feat(hub): host-routed spawn discovery via the admin proxy (07b)

obra:
## What this is

Component 07b (first slice): make the hub frontend actually route the spawn
form's host-dependent discovery calls through 07a's `evener/host/request`
proxy. Stacked on #1361 (`multi-host-pr07a-host-admin`, current head d2b89a769;
the branch was rebased onto that tip, so this PR contains none of 07a's
commits and reverts none of them).

## What now routes, and why

A remote spawn must be validated against the *remote* machine, not the
controller. Before this PR the spawn form issued every discovery/validation
call on the controller's own connection, so a remote launch could accept a
controller-local path, show the controller's models/harnesses, read the
controller's repository branch, and describe the controller's plugin/provider
configuration.

New seam `cmd/evener-hub/frontend/src/stores/hostRouting.ts`:

- `hostRequest(client, host, method, params)` — the local host takes
  `client.request(method, params)` **byte-for-byte**; any other host is wrapped
  as `client.request("evener/host/request", { host, method, params })`, whose
  result is the forwarded method's own result verbatim. The hub's fail-closed
  exact allow-list (07a) remains the security boundary.
- `HOST_DEPENDENT_DISCOVERY_METHODS` — the spec's discovery set, as data
  (component 06 §"Frontend changes" / criterion 10; component 07 §"Proxy
  method").

Wired through that seam:

| Call | Where |
|---|---|
| `model/list` | `Spawn.tsx` (pane catalog, global sweep), `useProviderSetup.ts` |
| `evener/harnesses/list`, `evener/launch/schema` | `Spawn.tsx` mount catalogs |
| `evener/launch/resolve` | `Spawn.tsx` default-model preview |
| `evener/paths/complete`, `evener/path/validate` | `Spawn.tsx`, `preflight.ts` |
| `evener/dirs/create` | `Spawn.tsx` create-and-start flow, `preflight.ts` |
| `evener/projects/recent` | `Spawn.tsx` |
| `evener/spawn/slashCatalog` | `useSpawnSlashCatalog.ts` |
| `evener/git/head` | `branch.ts` (branch/location chip) |
| `evener/plugin/preview` | `usePluginPreview.ts` |
| `evener/instance/list` | `credentials.ts` via new `fetchHost(host)`, called by `useProviderSetup.ts` |

The mount-time catalogs re-run when the selected host changes
(`[client, hostChoice]`), and the slash/plugin caches key on the host so a
remote selection never reuses the controller's response.

## Local path preserved

`isLocalHost()` treats `local`, `""`, `null`, and `undefined` as local, and
`hostRequest` then calls `client.request(method, params)` exactly as before.
`credentialsStore.fetch()` still calls the plain local `evener/instance/list`
(`loadInstances(LOCAL_HOST)`); `fetchHost(host)` is the new, additive surface.
`resolveHeadBranch`/`preflightDir`/`createDir` and the two hooks take the host
as an optional trailing/defaulted argument, so every existing 2-arg call site
and its tests are unchanged. No existing assertion was weakened.

## Generated bindings

Not changed. `make lint-generated` runs `make generate`; it **PASS**ed (6s) and
produced **no diff** — 07a's `evener/host/request` catalog row is already in
`appwire-client/typescript/types.gen.ts` (`HostRequestParams` /
`HostForwardedResult`, `MethodTypes["evener/host/request"]`), so
`client.request("evener/host/request", …)` is typed from the generated
bindings. Nothing was hand-edited.

## Tests

- `stores/hostRouting.test.ts` — local host → plain call (byte-for-byte, no
  wrap); remote host → `evener/host/request` with the right Host/Method/Params;
  result and error passthrough; an inventory guard that
  `HOST_DEPENDENT_DISCOVERY_METHODS` equals the spec's set; and a
  `test.each` over **every** spec method asserting it routes for a remote host.
- `panes/spawn/Spawn.test.tsx` — with a remote host selected before mount,
  every mount-path discovery call (`model/list`, `evener/harnesses/list`,
  `evener/launch/schema`, `evener/launch/resolve`, `evener/git/head`,
  `evener/instance/list`) goes through `evener/host/request` with
  `host: "buildbox"`, and **no** spec discovery method is issued directly.
- `preflight.test.ts`, `branch.test.ts` — `path/validate`, `dirs/create`,
  `git/head` wrap for a remote host.
- `useSpawnSlashCatalog.test.ts`, `usePluginPreview.test.tsx` — the catalog and
  preview resolve on the selected host through the proxy.

## Gates

- `npx biome check --write` on all touched `src/` files — PASS (clean).
- `make test-web` — **PASS** (`web-typecheck`, `web-test`, `web-lint`).
- `make test-web-browser` — **PASS** (layout/overflow/shell/spawn/transcriptscroll/skillguard; Chrome available).
- `make lint-generated` — **PASS** (no diff).
- `npm run typecheck` — PASS.

## Scope / follow-ups

- 07b first slice (spawn discovery routing). **07c** (credential push) and
  **07d** (host device login) are follow-ups, as is the host-context routing of
  the *settings* panes' admin calls (spec 07's own "PR 07b — settings panes").
- **Notification envelope not in this slice:** spec 07's `evener/host/notification`
  fan-out and the host-scoped store subscriptions are not wired here. The
  spawn form's credential-change refresh still listens to the bare
  `evener/auth/updated` on the controller, so a credential change on a remote
  host will not refresh the spawn catalog until that subscription lands. This
  is a discovery-call PR; no routed discovery call depends on it.

## Deviations / uncertainty

- **Spec seam moved.** Component 06 cites `frontend/src/shell/gitLocation.ts`
  for the `evener/git/head` read; that file does not exist. The read lives in
  `cmd/evener-hub/frontend/src/panes/spawn/branch.ts` (`resolveHeadBranch`),
  which is what this PR routes. Component 07's own citations
  (`usePluginPreview.ts`, `credentials.ts`) matched the code; its
  `stores/extensions.ts` / `stores/launchConfig.ts` filesystem helpers are the
  settings panes' duplicates, which belong to the settings-pane slice, not the
  spawn form, and are left unchanged here.
- `HOST_DEPENDENT_DISCOVERY_METHODS` includes `evener/launch/schema` and
  `evener/launch/resolve`. Spec 07's strict discovery enumeration omits them
  (they sit in the launch admin family that the same allow-list already
  covers), but component 06 names launch resolution/schema as host-dependent
  spawn calls, so they are routed and counted in the seam test.
- **Push used `--force-with-lease`.** 07a advanced to d2b89a769 after this
  branch was created from 33a34183e; a plain push would have left the PR
  reverting 07a's newest fix. The branch was rebased onto the current 07a tip
  and re-pushed with an explicit lease.


obra:
Round one addressed (`682ede6`).

- **High — `sweepStaleModels` wiped local `localStorage` defaults from a remote catalog.** The global-model effect fetched `model/list` for `hostChoice` and swept the controller's persisted `evener-hub.spawn-defaults.*` / global-model keys against whatever came back, so a narrower remote catalog permanently deleted local project models. The sweep is now gated to `isLocalHost(hostChoice)`: only the controller's own catalog has authority over the controller's defaults, and the draft-level validation effect still validates the in-memory draft against the remote list without writing storage. Pinned by `Spawn.test.tsx` — "a remote host's catalog never sweeps the controller's saved model defaults" (fails on the pre-fix gate).
- **Medium — remote instances overwrote the controller-scoped `credentialsStore`, then controller-scoped writers reverted them.** `fetchHost(remote)` wrote the remote listing into the one shared `instances` slot read by Settings > Credentials and `ConnectProviderDialog`, and `scheduleRefetch`'s `evener/auth/updated` → `fetch()` (local) silently replaced it; `useProviderSetup` therefore showed the wrong machine's registry, and Settings edits/removals acted on the controller while the user believed they were editing the remote host. Remote listings now live in a per-host partition (`hosts: Record<string, HostInstanceState>`); the top-level fields stay controller-scoped and controller-only, ordering is per host (a controller read no longer supersedes an in-flight remote read, and a connection change still retires every in-flight load), and `useProviderSetup(host)` reads the selected host's partition. Pinned by `credentials.test.ts` — "host-partitioned instance lists (component 07b)" (6 tests: partition isolation, notification refetch, in-flight race, mutation isolation, per-host error, reconnect); `useProviderSetup.test.tsx` — "a remote host's provider status comes from that host's own partitioned listing" and "a controller-scoped refetch does not replace a remote host's provider list"; `ConnectProviderDialog.test.tsx` — "mounting the dialog reads the controller's list and issues no proxied call".
- **Medium — a host change retained the previous host's harness/schema catalogs.** The mount effect reloaded `evener/harnesses/list` and `evener/launch/schema` on a host change but never cleared the previous results, and both rejection handlers were no-ops, so a failed remote load left the controller's harness list selectable (discovered only at `thread/start`). `harnesses`/`schemaOptions` are now cleared when the selected host changes and when the new host's load fails. Pinned by `Spawn.test.tsx` — "a failed remote catalog load shows an empty harness list, not the previous host's" (fails on the pre-fix gate).
- **Medium — no behavior-exercising coverage for remote instance list / recent projects / path completion.** Added `useProviderSetup.test.tsx` (a remote host's `evener/instance/list` is forwarded with the selected host and the controller slot is never written or called) and `Spawn.test.tsx` — "a remote spawn reads recent projects and completes paths from the selected host" (asserts `evener/projects/recent` and `evener/paths/complete` arrive through `evener/host/request` with `host: "buildbox"` and the plain methods are never called).

No finding was contested; the seam, `isLocalHost`, and the preserved local path are unchanged, and no existing assertion was weakened or removed.

Verification: `npx biome check --write` on the 7 touched `src/` files → clean; `tsc --noEmit` → clean; `make test-web` → PASS (web-typecheck, web-test, web-lint); touched test files 317/317 passing; `git diff --stat` → 7 files, +549/−41, all under `cmd/evener-hub/frontend/`. Every new behavioral test was confirmed failing-first by temporarily reverting its fix (8 store/hook failures + 2 Spawn failures pre-fix). `make test-web-browser` does NOT pass in my environment: all six CDP guards fail with the identical `timeout calling navigateTo after 30000ms` (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard, skillguard), and the unmodified base commit `f9f37d7` reproduces the same failure in the same lane (`npm run spawnguard` → exit 1, same timeout), with Chrome present and `/dev/shm` healthy under a host load of ~31–43 — i.e. an environment blockage (the guard page never finishes loading), not this change. CI is the authority for that gate.

obra:
Round two addressed (`b07d9ed`).

- **Medium — remote config notifications were never consumed.** `credentialsStore.handleNotification` matched only the unwrapped `evener/auth/updated` and called the controller-only `fetch()`, and the spawn form's `credentialsGeneration` effect likewise only handled the unwrapped method. The store now also handles `evener/host/notification`: on a wrapped `evener/auth/updated` it debounces a `fetchHost(params.host)` into that host's own partition (per-host timers, so a burst from two hosts can't collapse into one load; the controller's fields and `fetch()` stay untouched). The spawn form's generation effect now observes a wrapped `evener/auth/updated` whose `params.host` is the selected `hostChoice`, invalidating the pane `model/list` cache and catalog. Pinned by `credentials.test.ts` — "a wrapped remote config notification refetches that host's partition, not the controller's" and "wrapped config notifications refetch each named host independently"; `Spawn.test.tsx` — "a selected remote host's wrapped config notification reloads the pane model catalog" (isolated from the store's own refetch by holding the host's `evener/instance/list` pending, so only the form's generation key can invalidate).
- **Medium — a remote spawn's "Connect provider" action edited the controller.** `providerSetup` is host-scoped but `ConnectProviderDialog` reads/writes the controller's `credentialsStore` top-level fields on the plain connection, so the offered remediation could never configure the selected host. The editor is now offered only when `isLocalHost(hostChoice)`; a remote host whose partition reports `missing` shows "No provider is configured on `<host>`. Configure one on that host to use a model." instead, and the open dialog is closed on a host switch. The local path is byte-for-byte unchanged. Pinned by `Spawn.test.tsx` — "a remote spawn never offers the controller's Connect provider editor" (asserts the controller still offers it, the remote host does not, and it returns when switching back).
- **Medium — `branchHead` and `defaultPreview` were not keyed by host.** `branchHead` is now `{cwd, host, head}` and `defaultPreview` is `{draft, host, effective}`, each only exposed when the host matches `hostChoice`, so a same-cwd host switch immediately drops the previous host's branch readout and resolved default (the stale default could otherwise mis-satisfy the Start model check while the new host's `evener/launch/resolve` is pending). Pinned by `Spawn.test.tsx` — "switching hosts with the same directory drops the previous host's branch readout" and "switching hosts with the same directory drops the previous host's default-model verdict".

No finding was contested. Round one's host-scoped store partitioning/sweep fix and the preserved local path are intact; no existing assertion was weakened or removed.

Verification: `npx biome check --write` on the 4 touched `src/` files → clean; `npx tsc --noEmit` → clean; `make test-web` → PASS (web-typecheck, web-test, web-lint); `git diff --stat` → 4 files, +348/−24, all under `cmd/evener-hub/frontend/`. Each behavioral test was confirmed failing-first by temporarily reverting its fix (2 store tests + 4 Spawn tests red pre-fix, green post-fix). `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` → web-layoutguard, web-overflowguard, web-shellguard, web-spawnguard, web-transcriptscrollguard all PASS; `web-skillguard` (TestSkillComposerBrowser) is environment-blocked on this host under load ~78 — it fails on its timing guards, and the unmodified base commit `682ede646` fails it identically in the same lane (base run: "timed out after 30000ms waiting for session B rendered as ended"; this tree: "expected two live sessions in the rail, found 0" / "timed out after 15000ms waiting for steered turn's input visible in the transcript"), and its own failure dump shows both sessions present while the rail is still rendering. CI is the authority for that guard.

obra:
Round three addressed (`fd4efade4`).

- **Stale draft values survive host switch** (`Spawn.tsx:897-903`, `:1441-1456`) — real. The catalog effect cleared `harnesses`/`schemaOptions` on a host change but left the draft's `harness` and `advancedOverrides` in place, and the form could not even show the mismatch: an empty harness catalog renders the `"evener"` fallback label (`harnessOptions` collapses to a single `evener` entry) and an empty schema renders no Advanced fields at all. Fixed by reconciling the draft's host-derived launch config against the answers that land for the CURRENT host — a harness that host does not offer drops back to its own default, and an override keyed to a `wireField` its schema does not expose is pruned along with its `advancedValues`/`advancedErrors` mirror — plus a `hostCatalogPending` block: Start is disabled and the ⌘/Ctrl+Enter chord refused from a host change until that host's answers settle. The block is bounded by *settling*, not by success (an empty catalog is an answer, so a host that refuses a harness list can never hold Start hostage), and it never applies to the initial mount, which invalidates nothing. Pinned by `switching to a host that does not offer the draft's harness drops it from the launch` (asserts the launched `thread/start` carries no `harness`), `switching to a host whose schema lacks the draft's advanced options drops them`, and `Start is blocked only until the selected host's catalog answers settle`.
- **Remote launches pollute controller-scoped defaults** (`Spawn.tsx:1469-1476`) — real. **Seam used: `spawnDefaults.saveDefaults`'s `SaveDefaultsInput.remoteLaunch`** — the same seam and the same flag the sibling 06b fix uses. I read `multi-host-pr06b-fleet-view-ui` (`446f522da`) first: it adds the identical field at the identical call site, gating `GLOBAL_MODEL_KEY`/`GLOBAL_WORKING_DIR_KEY` while still writing the per-project blob keyed by the submitted cwd. This is one mechanism, not a second scheme, and the two additions are identical when the stack merges. It is *not* redundant against this branch on its own: `b07d9eda9` does not contain 06b's change (`spawnDefaults.ts` is absent from this PR's diff), so `saveDefaults` here wrote both globals unconditionally. Pinned by `a remote launch keeps the controller's global defaults out of it (component 07b review, round three)` (unit: a local submit installs the globals, a following remote submit leaves them at the local values) and `a remote submit never writes the controller's global spawn defaults` (integration: the launched `thread/start` carries `source: "buildbox"`, the per-cwd blob is written, both globals are null).
- **Wrapped remote plugin/launch notifications are dropped** (`Spawn.tsx:760-766`, `extensions.ts:459-475`) — real for the plugin half, which is the harm the finding names. `handleNotification` now also handles `evener/host/notification`: `params.method === "evener/plugin/updated"` bumps `pluginRevision`, which is exactly the revision `usePluginPreview`/`useSpawnSlashCatalog` fold into their logical key before resolving through `evener/host/request` against the selected host. The controller-scoped refetches stay gated to a notification the controller emitted itself — `fetchPlugins`/`fetchLaunchLayer` are plain calls that read THIS hub, so a remote host's change must not schedule them. **Contesting the launch half as stated:** the store's `launchLayer` is controller-scoped too (`evener/launch/getLayer` with `GLOBAL_LAYER_PARAMS`) and its only readers are the controller-scoped settings sections (`dirListSetting`, `mcp`), so a remote host's `evener/launch/updated` has no host-scoped consumer in this store to invalidate; scheduling that refetch would replace this hub's layer with this hub's own unchanged one. It is recognized and deliberately left gated, and the code says so. Pinned by `a wrapped remote plugin update bumps pluginRevision`, `a wrapped remote launch update never refetches the controller's launch layer`, and `a wrapped remote method this store does not own leaves pluginRevision alone`.

Failing-first: disabling each fix in turn put the new tests red before restoring it — `expected +0 to be 1` (pluginRevision), `expected 'openai/gpt-5' to be 'anthropic/claude-sonnet-4-5'` (defaults unit) and `expected 'openai/gpt-4o' to be null` (defaults integration), and `expected 'external' to be ''` / `expected { agent: 'controller-agent' } to deeply equal {}` / `expected false to be true` (the three Spawn cases).

Round one's and round two's dispositions are intact: the host-scoped credentials partition, the local-only sweep (`isLocalHost(hostChoice)` before `sweepStaleModels`), the preserved plain local path, the wrapped `evener/auth/updated` handling in the pane, the host-gated Connect-provider action, and the host-keyed `branchHead`/`defaultPreview`. No existing assertion was weakened or deleted.

Verification:
- `npx biome check --write` on the six touched files — clean (no fixes applied, no errors).
- `npx tsc --noEmit` in `cmd/evener-hub/frontend` — clean, exit 0.
- `make test-web` — **PASS** (`PASS web-typecheck`, `PASS web-test`, `PASS web-lint`).
- `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` — run 2: `PASS web-layoutguard`, `PASS web-overflowguard`, `PASS web-shellguard`, `PASS web-spawnguard`, `PASS web-transcriptscrollguard`, `FAIL web-skillguard` (exit 1). Run 1 additionally had `web-overflowguard` exit 2 on the shared CDP `waitForFonts` readiness check ("the top document declares no web fonts…") while every overflowguard case printed PASS; it passes standalone (exit 0) and in run 2, so that was a load-induced readiness flake, not a regression.
- `web-skillguard` (`TestSkillComposerBrowser`) is the known environment condition: under this host's load it fails timing assertions, at a different point each run — run 1 `expected two live sessions in the rail, found 0`; run 2 and a standalone re-run `timed out after 15000ms waiting for steered turn's input visible in the transcript` — and the unmodified base `682ede646` fails the same guards under the same load (A/B verified in round two). This change is inert for that scenario: it is single-host (no host switch, so `hostCatalogPending` never engages), `remoteLaunch` is false for the local host, and no `evener/host/notification` is emitted without a remote host subscription.
- `git diff --stat` — frontend only: `Spawn.tsx`, `Spawn.test.tsx`, `spawnDefaults.ts`, `spawnDefaults.test.ts`, `extensions.ts`, `extensions.test.ts` (6 files, +387/−6).

obra:
Round four addressed (`64ec301ab`).

**Medium — Per-project defaults blob leaks remote harness/model to local spawns** (`spawnDefaults.ts`): confirmed. `saveDefaults` gated only the global scalars on `remoteLaunch` and still wrote (or, with an emptier form, *deleted*) the per-project blob keyed by `cwd` alone — and `resolveInitialDefaults`/`spawnDrafts.createDraft` read that blob before any host is known, so a remote `/tmp/foo` or `$HOME` collided with a local spawn of the same string. A remote launch now writes **nothing**: no per-project blob, no global scalars, and no `removeRaw` of a local project's stored layer. **Seam shape (coordinated with the lower component):** I fetched `origin/multi-host-pr06b-fleet-view-ui` (head `4bab604b1`) and read its `spawnDefaults.ts`; its round-eight shape keeps writing the cwd-keyed blob and drops only the host-specific `model` from it. That is the same `remoteLaunch` seam and flag, so this is not a second scheme — 07b tightens the same gate to write nothing at all, because this PR's harness catalog (`evener/harnesses/list`) is host-scoped too, and because the read side has no host to key a host-plus-cwd lookup with. On merge, keep the stricter remote-skip; it subsumes the model drop. Pinned by `spawnDefaults.test.ts` ("a remote launch writes no per-project blob", plus the round-three test updated to the new expectation), which failed at the head with the blob written, and by `Spawn.test.tsx` "a remote submit writes no controller spawn defaults" (a pre-existing local blob for the same path must survive).

**Medium — Host selector stays interactive during submission** (`Spawn.tsx`): confirmed — the picker had no `disabled` while `busy` was set for the whole preflight + `thread/start` window. It is now `disabled={busy}`, so the host that launches is the host on screen. Pinned by "the host picker is locked while a submit is in flight" (head: enabled mid-submit).

**Medium — Stale explicit plugin selection survives preview failure** (`Spawn.tsx`): confirmed — with the new host's preview failed, `pluginSelectionBlocked` fell back to false (no issues are known for a host that never answered) while the previous host's explicit names still rode `combinedOverrides`. Fixed with the verdict's second shape: a host change drops the explicit selection (and the previous host's `knownSelectionIssues`) back to `{ mode: "default" }`, in the same host-change branch that clears the catalogs — not by blocking, because the failed-with-no-list state renders no plugin panel, so a block would have no control left to clear it but Retry, which retries the very selection that may be causing the failure. Pinned by "a host switch drops an explicit plugin selection the new host never previewed" (asserts the draft reset *and* that `thread/start` carries no `enabledPlugins`; head: selection survived and rode the launch).

**Medium — Initial mount on a remote host is not treated as catalog-pending** (`Spawn.tsx`): confirmed — `hostCatalogPending` started false and was only set on a *change*, and `catalogHostRef` is seeded from the first render's host, so a mount that starts on an already-selected remote host (drafts persist at module scope) was never pending. It now initializes to `!isLocalHost(hostChoice)`; a local mount is still not pending, so the pre-host-routing local form is unchanged. Pinned by "a mount that starts on a remote draft waits for that host's catalogs" (head: Start enabled and the ⌘/Ctrl+Enter chord reached preflight; now blocked until the answers land, then released).

**Medium — Rapid A→B→A host switch silently wipes the draft** (`Spawn.tsx`): confirmed — the change branch cleared `harnesses`/`schemaOptions` but never invalidated `catalogHostSettled`, so switching back to A before B answered ran reconciliation against the just-cleared empty arrays and wiped the harness and every Advanced-options override/value. The change branch now also `setCatalogHostSettled(null)`; only answers that land for the current host re-stamp it. Pinned by "a switch away and back before the middle host answers keeps the draft" (head: harness/overrides/values wiped).

**Low — Remote working directory pollutes controller-wide `lastWorkingDir`** (`Spawn.tsx`): confirmed — both picker surfaces (`DirectoryPicker.onPick`, the mobile panel's `onCwdPanelClose`) wrote the global seed unconditionally. Both now go through one `rememberLastWorkingDir` seam gated on `isLocalHost(hostChoice)`. Pinned by "a remote directory pick never seeds the controller's last-working-dir" (head: key written); the local write stays pinned by the existing "only confirming a directory updates the launch defaults".

**Low — Host reconciliation discards errors for kept advanced fields** (`Spawn.tsx`): confirmed — `setAdvancedErrors({})` sat inside the values branch. It now filters `advancedErrors` by the same offered-field set, independently of the values branch, so a kept field's error survives and a dropped field's error goes with it. Pinned by "host reconciliation keeps the errors of the advanced fields it keeps" (head: `{}`).

All earlier rounds' dispositions are intact: host-scoped store partitioning and sweep, preserved local path, wrapped `evener/auth/updated` handling, host-gated connect action, host-keyed `branchHead`/`defaultPreview`, and reconciliation on host switch (the A→B→A pin is the new regression test for it).

Verification:
- `npx biome check --write src/panes/spawn/{Spawn.tsx,Spawn.test.tsx,spawnDefaults.ts,spawnDefaults.test.ts}` → clean.
- `npx tsc --noEmit` → clean.
- `make test-web` → PASS (`PASS web-typecheck`, `PASS web-test`, `PASS web-lint`, `PIPESTATUS: 0`).
- `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` → `PASS web-layoutguard`, `PASS web-overflowguard`, `PASS web-shellguard`, `PASS web-spawnguard`, `PASS web-transcriptscrollguard`, `FAIL web-skillguard` (exit 1).
- `web-skillguard` / `TestSkillComposerBrowser` (cmd/evener-hub, `browserguard`) is the known environment condition: the driver reached `drain-released` and then failed its timing assertion, "skillguard FAIL: timed out after 15000ms waiting for steered turn's input visible in the transcript" (driver log `…/skillguard-browser-199811107/artifacts/driver.log`). A standalone re-run of that one test failed at the identical assertion (30.35s), and the same message appears from other lanes on this host tonight (23:22, 02:13) next to other load failures ("expected two live sessions in the rail, found 0", "browser startup deadline exceeded after 30000ms … chrome devtools endpoint", "timed out after 30000ms waiting for session B rendered as ended"). The unmodified base `682ede646` fails the same guards under the same load (A/B verified in round two), so this is environment-blocked: every other guard is green and `make test-web` passes.
- Failing-first: with the two source files reverted to `fd4efade4`, all nine new/updated pins fail (`Tests 9 failed | 240 passed`); after the fix the same files are `249 passed`. The two pins rewritten after that run (the plugin-switch and advanced-error tests) were re-verified red on their own (`AssertionError: expected { Object (mode, names) } to deeply equal { mode: 'default' }` and `expected {} to deeply equal { maxRounds: 'fix this value' }`).
- skillguard A/B: base `682ede646` fails the same timing guards under the same load (round-two A/B in this PR), and this head fails them identically in two runs tonight → environment-blocked, not this change.

obra:
Round five addressed (`055322c8d`).

- **Medium — Failed catalog load silently wiped the draft's harness/advanced options.** A rejected `evener/harnesses/list` / `evener/launch/schema` load cleared the lists and the same `Promise.all` stamped the host "settled" whatever the outcome, so one transient hub error reconciled the draft against an empty catalog: `setHarness("")` plus every Advanced-options override dropped, with nothing left for a later successful load to restore. The two states are separate now: each load resolves to whether it *answered*; `setHostCatalogPending(false)` still runs on either outcome (a host that refuses a catalog must not hold Start hostage) and only a load that answered both calls `setCatalogHostSettled`; the rejection handlers leave the catalogs alone, exactly as they did before round four. Pinned by "a failed catalog load leaves the draft's harness and advanced options alone" (fails before the fix: `expected '' to be 'external'`) plus "a host whose catalogs all fail still releases Start" (guards the release half, so the gate can never be closed forever).

- **Medium — Manifest unavailability silently converted a remote draft to `"local"`.** `sources` is empty both while the manifest is in flight (or after a failed fetch) and when it lists nothing remote, so `hostChoice` read that emptiness as "fall back to local" and ran the draft's launch - and every host-scoped discovery call - on the controller. `hostChoice` now keeps the draft's own source until the manifest has named its sources (`sources.length > 0`); the settled-empty and offline-host fallbacks are unchanged. Pinned by "a remote draft is not silently converted to local while the manifest is unavailable" (fails before: no `evener/host/request` call, and `thread/start` carried no `source`).

- **Medium — `hostCatalogPending` did not track `model/list`.** The gate covered harness/schema only, so after a host switch a model chosen from the previous host's catalog could be submitted while the new host's list was still in flight. A new `modelHostSettled` stamp records the host whose global model list has answered; `modelHostPending` (`model !== "" && modelHostSettled !== hostChoice`) joins both the disabled-Start condition and `handleSpawn`'s chord guard. Settlement, not success, is the bar - a refusing host releases the gate (same rule as the catalog gate), and the existing draft validator still clears a model the host turns out not to serve. A local mount starts settled, so the pre-host-routing local form stays startable. Pinned by "a previous host's model cannot be submitted while the new host's model list is outstanding" (fails before: Start stayed enabled after the switch with the new host's list held).

- **Medium — Async path-validation results could cross host boundaries.** The injected `validatePath` closed over the host selected when the field was edited while the panel accepts the answer on field identity alone, so a late host-A response could mark - or, through a `pathList` add, re-add - an advanced path override after the reader moved to host B. The closure now discards a superseded host's answer and re-asks the host selected now, so every consumer acts on an answer for the machine it is about to launch on (only a switch DURING a request continues the loop; each round is one real RPC). Pinned by "a late path validation from the previous host never marks the field on the new host" and "a late pathList validation from the previous host cannot re-add its entry on the new host" (both fail before the fix; the second was re-verified failing without the guard: `expected [] to have a length of 1 but got +0`).

- **Low — Remote directory pickers fell back to the controller's last-working-directory.** Both picker surfaces now take `pickerFallbackDir`, which is the controller's last accepted directory for local launches only; a remote picker opens from the committed cwd, or the selected host's home when there is none. Pinned by "a remote directory picker does not start from the controller's last working directory" (fails before: the forwarded validation asked the remote host about `/home/controller/last`).

**Interaction with #1302 (`multi-host-pr06b-fleet-view-ui`, fetched at `4c3e9646`).** Its head already carries the manifest fix as `submittedSource = sources.length > 0 ? hostChoice : source`; this branch computes that same value as its single `hostChoice`, so the two agree on the wire value in every case and the merge only has to pick one spelling (this branch keeps one value because every host-scoped RPC reads it). Its `AdvancedOptions.tsx` is still byte-identical to this branch's, so the path-validation guard lives at the shared closure instead, and it composes with any owner/generation guard the lower branch adds inside `AdvancedOptions` for the `invalid` flag: theirs decides whether a result may be written, mine guarantees the result is for the host selected now. One semantic overlap is for the merger: 06b's closures skip controller-side validation entirely for a remote target, while this branch routes the check to the selected host and re-asks on supersession - the host-routed shape has to win, or the remote path check is dropped rather than answered by the host that can run it.

Verification: `npx biome check --write` on the two touched `src/` files -> clean (re-check "No fixes applied"); `npx tsc --noEmit` -> clean; `make test-web` -> PASS (`PASS web-typecheck`, `PASS web-test`, `PASS web-lint`, 15m08s); `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` -> `PASS web-layoutguard`, `PASS web-overflowguard`, `PASS web-shellguard`, `PASS web-spawnguard`, `PASS web-transcriptscrollguard`, `FAIL web-skillguard` - `TestSkillComposerBrowser` (`cmd/evener-hub`, browserguard tag; driver `scripts/skillguard/run.mjs`) fails at its first check, "expected two live sessions in the rail, found 0", although the guard's own failure dump (`skillguard-failure.dump.json`, ~1.5s later) shows both sessions rendered with the expected refs; it reproduced identically on the required re-run (5.15s). Isolated re-run `go test -tags browserguard ./cmd/evener-hub -run '^TestSkillComposerBrowser$' -count=1` -> `ok ... 28.4s`, so this is the known load/timing failure (fix in flight in #1385), not a product regression: every other guard is green, and the skillguard passes when it is not competing with the other guards on this host.

obra:
Round six addressed (`01fee1dd2`).

- **Medium — a partially-answered host catalog suppresses reconciliation for the half that did answer** (fixed). `Spawn.tsx` kept one `catalogHostSettled` stamp, set only when BOTH `evener/harnesses/list` and `evener/launch/schema` answered, and the reconciliation effect bailed unless that stamp matched the selected host — so a host that answered one catalog and refused the other reconciled neither half, while `hostCatalogPending` was still released by the settlement and Start became enabled. Settlement is now tracked per catalog (`harnessesHostSettled` / `schemaHostSettled`, each stamped where its own answer lands) and the reconciliation effect reconciles only the half whose request answered: the draft harness drops when the answered harness list omits it, and the Advanced-options override/value/error maps filter when the answered schema omits their fields. A rejection still reconciles nothing and still releases Start.
  - Pinned by two new round-six tests in `Spawn.test.tsx`: "a host that answers its harness list reconciles the draft even when its schema rejects" and "a host that answers its schema reconciles the advanced options even when its harness list rejects". Failing first on `055322c8d`: the first failed with `AssertionError: expected 'external' to be ''` (the draft harness stayed `external` and rode `thread/start`), the second with `expected { agent: 'controller-agent', maxRounds: 7 } to deeply equal { maxRounds: 7 }`. Both pass with the fix; the full `Spawn.test.tsx` file is 236/236.
- Round-five property intact: no assertion was weakened or removed. "a failed catalog load leaves the draft's harness and advanced options alone" and "a host whose catalogs all fail still releases Start" pass unchanged — a rejected load still stamps nothing (so it reconciles no half), and the `Promise.all` settlement still releases `hostCatalogPending`; the per-half gating only adds reconciliation for the catalog that actually answered.

Verification: `npx biome check --write` on the two touched `src/` files → clean; `npx tsc --noEmit` → exit 0; `make test-web` → `PASS web-typecheck`, `PASS web-test`, `PASS web-lint`; `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` → `PASS web-layoutguard`, `PASS web-overflowguard`, `PASS web-shellguard`, `PASS web-spawnguard`, `PASS web-transcriptscrollguard`, `PASS web-skillguard` (skillguard passed on this run; no re-run needed).

obra:
Round seven addressed (`723c4c4a6`).

- **Medium — the "Create directory?" confirmation could act on a host the user never preflighted** (`Spawn.tsx`, `handleCreateConfirm` vs the gate in `handleSpawn`). Confirmed the finding against the code before changing anything, link by link: `hostChoice` falls back to `"local"` whenever the manifest's live `online` flag flips for the chosen source (line 377); the catalog effect clears the lists and sets `hostCatalogPending = true`, and the model effect unstamps `modelHostSettled`, on that same change (994-1007, 1131-1137); the dialog's open condition (`open={focused && createDialogPath !== null}`) says nothing about the host and it was dismissed only by user action; and the confirm read the render-current `hostChoice`, then ran `createDir(client, path, hostChoice)` and `doSpawn(...)` with no gate at all - skipping line 1739's `hostCatalogPending || modelHostPending` and everything the disabled Start button enforces behind it (`doSpawn` itself re-checks only `pluginSelectionBlocked`, line 1523). So a remote confirmation did create the path and start the session on whichever host took over, carrying a draft that host never offered.
- **Shape chosen: bind, and abort - not re-check alone, and not re-preflight.** `createDialogHost` now travels with `createDialogPath` in the draft store, because the pair is one action and because the path lives there for a reason: the spawn pane is a dockview singleton whose remount restores the dialog (the completion-ownership tests reopen it after a navigation), so a component-local binding would have been lost in exactly the case it protects. `handleSpawn`'s offer-create stamps it with the host that preflight actually ran against; the confirm re-checks it and **aborts** when the selection has moved, closing the dialog, leaving the draft intact, and toasting why. Aborting rather than re-running `preflightDir` against the new host is deliberate: the user confirmed a path on one machine, and one click must not become a second, unconfirmed action on another. As an additional safeguard, a host switch now dismisses the pending create outright (the same ref idiom the provider dialog uses for the same reason), so the dialog is gone before a confirm can be aimed at it - and the check inside the confirm still catches a confirm that raced that effect, or a dialog restored from the draft onto a selection that moved while it was unmounted.
- The pending gate `hostCatalogPending || modelHostPending` is now enforced in `handleCreateConfirm` too, so a confirm cannot reach `doSpawn` while the selected host's catalogs/model state is unsettled. It holds the click silently, same reasoning as `handleSpawn` - both requests always settle, refusal included (round five), so it is held, never dead-ended (pinned: the same test clicks again after the answer and the create + start then proceed, on the bound host).
- Prior dispositions kept intact: round five's settled-vs-answered separation, round six's per-catalog reconciliation, the host picker with local preselected and offline disabled, source threading through `startThread` (omitted for local), spawn-draft persistence, `selectSources`/`selectDisplaySources`, the rail host badge/offline affordance, host-scoped store partitioning and sweep, the manifest-preserves-the-confirmed-host fix, the `modelHostSettled`/`modelHostPending` submit gate, the cross-host path-validation guard, and the local-only picker fallback. No existing assertion was weakened or removed.
- Failing first, with the implementation stashed (unfixed source), `npx vitest run src/panes/spawn/Spawn.test.tsx -t 'a host that drops offline while the create dialog is open|a restored create dialog whose preflight host|a create confirmation waits for the selected host'`:
  - `a host that drops offline while the create dialog is open never creates or starts on the host that took over` → `AssertionError: expected <div role="dialog" …(3)>…(4)</div> to be null`
  - `a restored create dialog whose preflight host is no longer selected aborts instead of creating the path` → `AssertionError: expected [ 'local' ] to deeply equal []` (the stale confirm created the path and started on the host that took over)
  - `a create confirmation waits for the selected host's catalogs before creating and starting` → `AssertionError: expected [ 'buildbox' ] to deeply equal []` (the confirm ran while buildbox's catalogs were outstanding)
  All three pass with the fix, and the whole spawn directory runs 507/507.

Verification: `npx biome check --write src/panes/spawn/Spawn.tsx src/panes/spawn/Spawn.test.tsx src/panes/spawn/spawnDrafts.ts` → `Checked 3 files in 290ms. No fixes applied.` (clean); `npx tsc --noEmit` → clean; `make build-web` → built, exit 0; `make test-web` → `PASS web-typecheck`, `PASS web-test`, `PASS web-lint` (exit 0); `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` → `PASS web-layoutguard`, `PASS web-overflowguard`, `PASS web-shellguard`, `PASS web-spawnguard`, `PASS web-transcriptscrollguard`, `PASS web-skillguard` (exit 0).

obra:

The parent components (04b-07a) landed, so this component's own 8 commits were rebased onto `main` (`git rebase --onto origin/main d2b89a769`) and force-pushed with `--force-with-lease`. The PR's diff is now exactly this component's, against `main`.

This was the heaviest reconciliation of the stack. `main` had since landed components 05a-07a (including the whole `evener/host/request` proxy this component consumes) and the AppWire SDK migration that deleted the `src/protocol/*` seam, so most of this branch's files had to be re-ported rather than merged. Each decision keeps `main`'s landed semantics and re-pins this component's intent to them:

1. **The launch target is `submittedSource` everywhere.** 07b threaded `hostChoice`; main's 06b flow distinguishes the settled `hostChoice` from `submittedSource` (the value the form actually submits, which keeps the draft's host while the manifest is still settling). Every host-dependent discovery/validation call - `model/list`, `evener/harnesses/list`, `evener/launch/{resolve,schema}`, `evener/paths/complete`, `evener/path/validate`, `evener/dirs/create`, `evener/projects/recent`, `evener/spawn/slashCatalog`, `evener/git/head`, `evener/plugin/preview`, `evener/instance/list` - is issued against `submittedSource` through `evener/host/request`, and the catalogs it feeds are keyed by it.
2. **main's remote "refusals" are superseded by routing.** main skipped path validation for a remote target, refused "New folder" with a toast, and skipped the cwd preflight, each with a comment saying the proxy was "absent from this branch". It is no longer absent: those calls are routed to the selected host, and main's tests for the refusals are re-pinned to the routed behavior (see item 6).
3. **The controller-side guards stay.** The uncredentialed-default fallback is still retired when the target becomes remote, the persisted-defaults sweep still runs only for the controller's own catalog (a remote host's catalog describes one machine and must not delete the controller's saved models), and the global cwd picker seed is still gated on a local target.
4. **The provider verdict is host-scoped**: `providerRequired` no longer excludes remote targets, because `useProviderSetup` now reads the selected host's own listing. The Connect-provider editor is controller-only (`providerSetupIsLocal`); a remote target with no provider on that host gets a message naming the host instead of an action that would configure the wrong machine, and the dialog closes on a target switch.
5. **The credentials store's host partition was reimplemented on main's store.** main replaced the hand-written store with the package's `createCredentialInstancesStore`; 07b's `hosts[host]` partition therefore lives beside it in `stores/credentials.ts` (`hostInstancesStore`, `hostPartition`, `useHostInstances`, `fetchHost`, `resetHostInstancesForTests`) and is read by `useProviderSetup`. It keeps the branch's guarantees: a remote listing can never enter the controller-scoped fields (or be touched by their `evener/auth/updated` refetch), a connection transition releases in-flight status and invalidates an answer that would otherwise commit across it, and a wrapped `evener/host/notification` carrying `evener/auth/updated` refetches that host's partition (debounced, and only for the notification kinds this store reads).
6. **Tests re-pinned** (each with a comment where it changed): the six main-authored remote-target tests whose behavior routing supersedes (folder creation, remote-selectable directory, picker seed, fallback model, remount catalog, advanced-path staleness), plus this component's round-three/round-four/round-five/round-seven tests whose assertions main's landed rounds had already moved (persisted-defaults blob contents, the no-delete rule, the plugin-selection reset on a host switch, the create-dialog host binding). The failed-catalog reconciliation tests are unchanged in intent: the per-catalog host stamps they pin now live in the load handlers.
7. **The path-validation staleness effect is keyed on the target** (main's `pathValidationMode` ref becomes `pathValidationTarget`): on a target change every stored non-empty advanced path value is re-validated, because an unflagged record still needs the new target's verdict.
8. **`resolveHeadBranch` moved into `shell/gitLocation`** (main deleted `panes/spawn/branch.ts`); it now takes the host and routes through `evener/host/request`. The composer's own `resolveGitLocation` stays controller-scoped - the session surfaces are not host-aware yet, a tracked follow-up rather than a regression.

Gates on the rebased tree (`21b0c0a65`): `npx biome check --write` on the touched `src/` files clean; `npx tsc --noEmit` exit 0; `make test-web` PASS (web-typecheck, web-test, web-lint); `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` PASS (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard, skillguard); `make lint-generated` PASS.

**Carried deviations and follow-ups** (recorded rather than silently absorbed):
- This component's round-four "a remote launch persists nothing at all" was not adopted: `main`'s rounds seven/eight/nine already scope the persisted defaults (the cwd-keyed blob keeps the local project's layer while withholding the host-specific model, and the global scalars stay local-only). Only the no-delete half of that round was adopted (an empty remote form no longer removes a local project's stored layer). If the intent was for round four to supersede `main`'s rounds, that is a deliberate semantic change this rebase did not take.
- The branch readout's display key (`Spawn.tsx`, `branchHead.host === hostChoice`) and the wrapped `evener/host/notification` generation bump still compare against `hostChoice` while their requests use `submittedSource`. They differ only while the manifest is unsettled (the picker still reads Local). Both predate this rebase.

obra:
Round one addressed (`b816327a7`).

- **Medium — the remote-target notice still described controller-local discovery** — `cmd/evener-hub/frontend/src/panes/spawn/Spawn.tsx`. The notice is removed, together with its assertion in `Spawn.test.tsx`. Its own comment said it was to be removed "when source-aware discovery (the evener/host/request proxy) lands", and this series lands it: every reading it named - path browsing (`evener/paths/complete`), models, providers, plugins - is now issued against the selected host through the proxy, so the copy was false on every clause.
- **Low — `remoteLaunchRef` was assigned but never read** — removed, with the comment that claimed a live async reader.
- **Low — the frontend discovery set and the Go allow-list are only independently pinned** — recorded as an immediate follow-up rather than built in this round: the check the finding asks for (a shared/generated list, or a test that drives each frontend-issued method through `hubHostAdminController.Request`) spans this PR's frontend and the landed Go proxy (`cmd/evener-hub/app_host_admin.go`), so it is carried on the campaign's follow-up list.

This head also carries a merge of `main` (12 commits, one of which touched `stores/credentials.ts`): the merge keeps main's SDK re-export (`foreignListingChange` alongside the existing names) and this branch's host-partition additions beside it.

Verification: `npx tsc --noEmit` exit 0; `make test-web` PASS (web-typecheck, web-test, web-lint); `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` PASS (all six guards); `make lint-generated` PASS.

Note on the previous run's CI: the `race-modules / agent` job timed out at its 10m limit (a Go-module test timeout in a module this frontend-only PR does not touch); this head re-runs CI from scratch.

obra:

**CI** — all required checks green at `b816327a7` (the full set, including RoboRev's own `roborev` check). One job needed a re-run: `race-modules / agent` failed on `TestDelegateResourceSupervision_CommittedSendStartRefusesASecondTurn` (`agent/delegate_resource_supervision_test.go:255` — the committed send returned `target_busy` instead of `started`). That is unrelated to this PR: it changes 21 files, none of them Go, and the failing file belongs to `main` (last touched by #1329). The test passes on this tree `go test -race -count=20 ./agent/`, the whole `./agent/` race suite passes locally, and the same job was green on `main`'s run 35115328679 minutes earlier; the previous CI run of this PR had the same job time out at its 10-minute limit. Re-running the failed job produced a green check set.

**Review** — RoboRev's verdict for `b816327a7` (`## roborev: Combined Review`, 1 reviewer, 2 skipped): **three findings, all Low, no Critical/High/Medium.** Under the campaign's rule (a Lows-only verdict lands, with the Lows recorded as an immediate follow-up), they are recorded here:

- `cmd/evener-hub/frontend/src/panes/spawn/Spawn.tsx:382` (same pattern at `:820`) — the branch readout is stored against the host the effect actually queried (`submittedSource`) but displayed against `hostChoice`, and the wrapped-notification filter compares `n.params.host === hostChoice`. Those differ exactly in the state the round-five code deliberately supports (an empty sources manifest: `submittedSource` is the draft's remote host while `hostChoice` is `"local"`), so with a persisted remote draft and a manifest that is still loading the branch chip for the selected host does not render, and the model-catalog generation bump is missed. Fix asked: compare against (and key the effect on) `submittedSource`, as every other host-scoped path in the file does.
- `cmd/evener-hub/frontend/src/stores/credentials.ts:144` — `fetchHost` guards a late response with the connection `generation`, which is shared by all requests, so two overlapping `fetchHost` calls for one host can land out of order and overwrite the partition with pre-change rows. Fix asked: keep a per-host request sequence and drop a response that is not the latest for that host (the package's own `createCredentialInstancesStore` does this with `requestVersion`, `state/credentials/instances.ts:536-549`).
- `cmd/evener-hub/frontend/src/panes/spawn/spawnDefaults.ts:177-184` — the remote-submit guard only covers the fully-empty case; `blob` is non-empty whenever a `model` was carried over, so `writeRaw` replaces the whole per-cwd blob and drops the stored `harness`/`access_mode`/`reasoning_effort` when the remote form does not re-supply them. Fix asked: for `remoteLaunch`, start from the previous blob and overlay only the fields the remote form names (or skip the per-project write), rather than writing a blob assembled solely from the remote submission.

RoboRev also verified the backend allow-list in `cmd/evener-hub/app_host_admin.go` names every method the frontend forwards through `evener/host/request`, so no routed call fails closed unexpectedly. That is the cross-language pin round one's Low had asked to build, checked by the reviewer against this head.

The rebase onto `main` and its reconciliation decisions (the whole 05a-07a series and the AppWire SDK migration had landed in between, so most files were re-ported rather than merged; `submittedSource` is the launch target everywhere, `main`'s remote refusals are superseded by routing, controller-side guards stay) are documented in the rebase-note comment above.

**Up-to-date override** — the branch is 3 commits behind `main`. Overlap check, run immediately before this comment:

```
$ gh pr view 1362 --json mergeable,mergeStateStatus --jq '.mergeable, .mergeStateStatus'
MERGEABLE
BEHIND

$ comm -12 <(gh pr view 1362 --json files --jq '.files[].path' | sort) \
           <(git diff --name-only "$(git merge-base origin/main origin/multi-host-pr07b-host-routing)" origin/main | sort)

$ git diff --name-only "$(git merge-base origin/main origin/multi-host-pr07b-host-routing)" origin/main | wc -l
21
```

`main` changed 21 files since the merge base; none of them is a file this PR touches.

PR #1272, opened 2026-09-13
refactor(sdk): import the AppWire package by name and delete the protocol/ seam (SDK A4)

obra:
Plan row: A4 in `docs/superpowers/plans/2026-09-12-sdk-migration.md` (#1182); follows A3 (#1241).

Every web and native import of the AppWire TypeScript package now names the package instead of a path, and A3's re-export seam at `cmd/evener-hub/frontend/src/protocol/` is gone.

## Counts per specifier, per tree

Statements in the tree now, after the merge pass below collapsed the duplicates the rewrite creates:

| Tree | `@evener/appwire-client` | `/docContent` | `/testing/*` | Total |
| --- | --- | --- | --- | --- |
| `cmd/evener-hub/frontend/src` | 489 | 3 | 124 (`fakeClient` 113, `notifications` 6, `tokenFlood` 3, `fakeSocket` 1, `reducerHooks` 1) | 616 |
| `mobile-native` | 120 | 2 | — | 122 |
| `mobile/src` | 20 | — | — | 20 |
| **Total** | **629** | **5** | **124** | **758** |

The sweep rewrote 870 statements across 515 files, plus one hand edit (below); merging duplicate specifiers brought the 871 down to the 758 above.

`mobile-native/scripts/*.mts` keeps its six relative imports — the plan's named carve-out, since those run under `tsx`, which reads no tsconfig `paths`.

## The script

`scripts/sdk/rewrite-package-imports.mjs` (`--help`, `--check`). It parses each file, resolves each specifier on disk, and picks the target by the **symbols the statement names**, not by its path prefix — `readDocFile` is published only at `./docContent`, `FakeClient` only at the in-repo `testing/` specifier, everything else at the root. A symbol the package publishes nowhere stops the run and writes nothing. A second pass merges the duplicate statements the collapse creates, so a file ends up with at most one value import and one type import per specifier.

Beyond the specifier strings, the only lines the diff touches are the two inline `import("…")` type positions (`useSpawnSlashCatalog.test.ts:124`, `navigationPages.test.ts:421`). Biome re-sorted the web files it touched, because the new specifier sorts elsewhere than the old relative path.

[...]

PR #1365, opened 2026-09-15
feat(hub): assert the endpoint fingerprint on instance edits

obra:
## What

The instance sheet's stale-draft guard is client-side only: it compares the row it displayed against the store's listing, and a concurrent change can outdate that listing between the comparison and the `evener/instance/edit` RPC. This adds the server-side half, mirroring the credential-write params.

- `appwire.InstanceEditParams` gains an optional `ExpectedEndpointFingerprint` (`omitempty`).
- The hub's `Edit` resolves the fingerprint key once, off both locks (the round-59 threading already used by `Remove` and the credential writes), then validates the assertion under `mu` + `credMu` before anything is written, refusing a mismatch with the endpoint-conflict `appwire.Conflict`.
- The frontend `InstanceSheet` sends the fingerprint it displayed when the hub served one, and omits the field when the hub could not key it, so an outage does not block a save.
- Regenerated `appwire-client/typescript/types.gen.ts` and `docs/appwire-protocol.md` with `make generate`.

An absent or empty assertion keeps the pre-fingerprint behaviour for the TUI and older clients (an edit carries no secret and the field must stay backward compatible). This is the rule the maintainer chose for the credential writes too.

## Tests

Red-first, both sides, each mutation-verified:

- Hub (`cmd/evener-hub/app_instances_edit_endpoint_test.go`): a stale assertion after a re-pointed `base_url` is refused with an `appwire.Conflict` and nothing reaches `providers.toml`; the current assertion is accepted; an absent assertion still lands; the key is resolved before both locks (TryLock probe on the seam, mirroring `TestInstances_KeyResolutionHappensBeforeBothLocks`).
  - Mutation: forcing the verification condition false reddens the guard test; dropping the off-lock resolution reddens the positive control and the lock-order probe.
- Frontend (`InstanceSheet.test.tsx`): a save asserts the displayed fingerprint on the wire; a row the hub cannot key sends no assertion. Three existing curated-shadow rename assertions were updated to the new wire shape.
  - Mutation: dropping the assertion from the request reddens the wire test (and the three updated assertions).

## Gates

On `e6ed43030` (branched from `origin/main` = `f614287ca`; no merge needed):

- `go build ./...` PASS
- `make lint` PASS (naming, gofmt, evenerfuzz, eval, internal, golangci, generated, fuzz-registry, secret-scan)
- `make vet` PASS
- `make test` PASS - every Go module (root `ok 374.71s`, agent 48.54s, llm 14.62s, auth 2.89s) and the web module (`ok 981.70s`)
- `make test-api-package` PASS (`qualified @evener/appwire-client@0.1.0`)
- `TMPDIR=/tmp make test-web-browser`: layout/overflow/shell/spawn/transcriptscroll PASS. `web-skillguard` timed out once at load ~40 (capability-loss wait, `waiting for session B rendered as ended`); an isolated re-run passed (`ok 45.063s`, all 10 scenarios). This is the host's known load-sensitive guard (it fails on unmodified main in the same window - r57 A/B), not this branch.


obra:
Disposition of this review's two Mediums.

**Medium 1 — unkeyed web rows send an unasserted edit: declined, by the maintainer's ruling.**

The contract for the new precondition is unchanged from the credential writes: an empty endpoint assertion is accepted while the hub can key a fingerprint, and the three web credential flows refuse to send one when it cannot. For the instance *edit* the sheet deliberately keeps its pre-fingerprint behaviour when the hub cannot key. Refusing every unkeyable save would also block correcting a misconfigured row during a fingerprint-key outage, which is why round 59 declined the same refusal for this sheet, and the maintainer chose to keep that rule when the contract question was put directly. The residual is narrow and stated: while the hub cannot key, a concurrent re-point of the same name between the sheet's own identity check and the RPC is unverified. Every keyed row is still checked server-side under the locks.

**Medium 2 — the native editor drops `endpointFingerprint`: fixed.**

`mobile-native/src/providerForm.ts` now carries the fingerprint the editor displayed (and asserts nothing for a row the hub could not key), and `ProviderEditor` re-anchors to the row now on screen and reports the change when the hub refuses the assertion, mirroring the web flows' `isEndpointConflict`. Red-first tests cover param construction and conflict detection; the editor's catch branch has no component-test harness in mobile-native and is not covered by one.

obra:
Superseded by #1397, per the maintainer's call to slim this change after the review loop. #1365's branch and its full history are preserved (local ref backup/instance-edit-full, remote branch instance-edit-endpoint-precondition).

The server-side precondition, the wire field, and both clients sending the displayed fingerprint are in #1397, rebuilt on current main (which has taken SDK A4, #1272). The two things this PR accumulated that are NOT in #1397 are the client-side re-anchor-on-conflict recovery and the endpoint-conflict discriminator - both follow-up material, the latter only earning its keep once a client acts on the distinction, and the former having drawn opposing review guidance across two rounds.

PR #1372, opened 2026-09-15
perf(web): tick the activity tree through its own context, not the chrome's prop

### LINKED ISSUE #1366: web: the activity panel's tree re-renders once a second while it is open (#1208 follow-up)
Follow-up named in #1208 rather than silently dropped.

`ActivityPanel` passes the ticking `now` into `ActivityTree` as a prop (`cmd/evener-hub/frontend/src/panes/session/chrome/ActivityPanel.tsx:219`), so the tree re-renders once a second while the panel is open. The row components are memoized, but the tree's own render body rebuilds the row set and the fold structure on every tick, so the memoization buys much less than it looks like it should. This is a regression introduced by #1208: the panel needed a live clock for watch durations and the "now" label, and the cheapest way to give it one was the prop that was already there.

Fix: tick through context only. `ActivityTree` already provides `TreeNowContext`, and `LiveRowDetail` reads it, so only the row details that display a live duration need to re-render. Reworking the tick plumbing touches three tests' time control, which is why it was not folded into #1208.

Evidence: `ActivityPanel.tsx:219` (`now={now}`), the memoized row components in `ActivityTree.tsx` (`RowDetail`, `DenseRowView`, `WatchRowView`, `FoldRowView`), and `ActivityTree.watchTick.test.tsx` for the current tick contract.

## Also folded in from #1208's follow-up list

These were listed in #1208's body under "Known follow-ups" and now live here instead of only in that PR:

- **`watchDeliveryInstants` reparses and sorts the watch's delivery ring on every render of a live row's detail** (`ActivityRowDetail.tsx:98`). The ring is bounded at 32 instants, so the cost is small, but it is paid per open row per tick for a value that only changes when the watch delivers. Compute it from the instants array (memoized on the array identity), not on every render.
- **`WatchRowView` takes the ticking `now` as a prop** rather than reading the tick. It is memoized with `watchRowViewPropsEqual`, and an open row mounts `OpenWatchMeta` for its countdown, so the context-only ticking that fixes the tree render should carry this row too: it should re-render because it is a tick consumer, not because a prop changed. Collapsed rows already avoid the clock entirely.
- **`WatchRowView` duplicates `DenseRowView`'s chrome.** Both build the same `role="treeitem"` wrapper with the same aria attributes, tab index, focus/keydown/ref handlers and the same chevron disclosure button, differing only in the row body. A shared row-chrome wrapper would stop the two from drifting apart.

One note from #1208 is dropped as already resolved: the claim that "the panel's pure row model imports display helpers from the rail's React module". The row model now lives in `appwire-client/typescript/activityRows.ts` and imports no React at all.

obra:
## What this does

Stops the activity panel re-rendering once a second, and closes out the follow-up list from #1208.

Closes #1366.

The panel was threading the chrome's 1Hz ticking `now` down into `ActivityTree` and from there into every row. The tree already owns its own clock (`TreeNowContext`, provided once by `TreeTickProvider`), and only the leaves that actually display a live duration consume it — so the chrome's parallel clock was pure duplication whose only effect was to re-render the row list every second while the panel was open.

- The chrome's clock no longer reaches the tree at all. `ActivityPanel` and `ActivityPanelBody` no longer take it, `ActivityTree` no longer takes it, and `WatchRowView`/`OpenWatchMeta` read the tick from context like the live job and delegate rows already do.
- `ActivityPanelBody` is memoized. That is load-bearing, not decoration: with the chrome re-rendering the panel element every second, the memo is what keeps that re-render from reaching the tree (verified by removing it — the regression test then fails).
- `watchDeliveryInstants` is computed per watch identity rather than on every render of an open row's detail.
- The duplicated row chrome is extracted into one `RowShell` used by both the dense rows and the watch row: same `treeitem` wrapper, aria wiring, handlers and chevron, differing only in the body.
- The dead `now` prop is removed end to end, including `SessionChrome`'s shared hidden panel and the dev surface section.

## Evidence

- New regression test: `a prop-identical chrome re-render leaves the watch rows untouched`. It measures `watchMeta` calls across a chrome-style re-render of the panel; before the fix a changed `now` prop re-rendered the open row and its meta. Red evidence from the pre-fix code: `AssertionError: expected 1 to be +0`.
- `npx vitest run src/panes/session src/shell --maxWorkers=2`: 203 files, 4339 tests, all passing.
- `npx vitest run src/panes/session/chrome/`: 23 files, 482 tests, passing.
- `npx biome check` on every touched file: clean. `npx tsc --noEmit`: exit 0.

## Notes

- Behavior is unchanged for the standalone activity pane: it never passed a clock, so it always read context; the chrome now does the same thing the pane always did.
- One item from #1208's folded list is deliberately not touched here: the claim that the panel's pure row model imports display helpers from the rail's React module is already resolved — the row model now lives in `appwire-client/typescript/activityRows.ts` and imports no React.


obra:
Confirmed, and fixed in 8d203ad9a. The comment on `ActivityWatchDetail` claimed a caller with its own ticking clock passes `now`; nothing does — `ActivityTree` renders `<ActivityWatchDetail row={row} />`, so `TreeNowContext` is the only production source and the optional `now` is a deterministic test seam. That is what the comment says now, and the tree's own provider comment got the same treatment ("the tree owns that clock outright — no caller passes one in"). I also swept the chrome directory for other comments describing the removed plumbing: no other stale one survived; the ticking-clock mentions that remain are on components that really do take a clock (the status row, the details panel).

PR #1373, opened 2026-09-15
test(web): wait for the connect-dialog chunk instead of racing it against the query budget

### LINKED ISSUE #1369: web: the ModelSwitch "keyless connection" test flakes under parallel load
Observed once, not yet reproduced on demand.

While running the chrome suite in parallel (`npx vitest run src/panes/session/chrome/ --maxWorkers=2`) at a moment this machine was heavily loaded (load average ~88 on 16 cores), this test failed:

    FAIL src/panes/session/chrome/ModelSwitch.test.tsx > keyless connection refreshes the warmed real session catalog without switching until a pick
    Test Files  1 failed | 22 passed (23)
    Tests  1 failed | 480 passed (481)

The assertion text itself was lost to an over-narrow grep in the run that caught it. The test passes alone, and six runs of the same directory command have passed since, including one under four artificial CPU hogs -- so the trigger is timing-sensitive in a way load alone does not reproduce on command.

Next step: capture the assertion the next time it appears (run the file with `--reporter=verbose` and no output filter), then make the test wait for the state transition it actually asserts on rather than for a fixed tick. Until that happens this is one data point, tracked here so CI is not the next one.

Seen on main after #1208 merged (squash 4b291876a), in a file that change did not touch.

obra:
## What this fixes

Closes #1369.

`ModelSwitch.test.tsx > keyless connection refreshes the warmed real session catalog without switching until a pick` failed once under a heavily loaded parallel run and then passed six runs in a row, so the issue had one data point and no assertion text. The mechanism turns out to be a chunk import racing a query budget:

"Connect another provider" mounts `ConnectProviderDialog` from its own lazy chunk (`panes/spawn/connectDialogChunk`). The FIRST dynamic import of that chunk pays Vite's transform of the dialog and its transitive graph, and on a loaded machine that one-time transform outlasts testing-library's default 1s async-query budget. The next `findBy*` absorbed that cost, so the failure named the query while what had not happened yet was the import.

The fix waits for the import itself — the boundary's own Suspense transition — instead of hoping the next query survives it. `connectDialogTestUtils.openConnectDialog` is shared by both surfaces that can open the dialog (`ModelSwitch.test.tsx` and `ModelSwitchTrigger.test.tsx`, which has the same race and would have been the next flake).

Awaiting the import *before* the click is load-bearing, not stylistic: starting it alongside the boundary's own `import()` makes two concurrent evaluations of the chunk, and Vite's SSR runner re-enters the module — observed as `Cannot access '__vite_ssr_import_14__' before initialization` thrown from the dialog's render under a parallel run. The helper says so, and documents that the chunk-boundary retry tests, which expect the import to reject, must keep their own path.

## Evidence

- Deterministic regression test: `a connect-dialog chunk held pending still opens the dialog once it lands`. The seam hands the test a *parked* import (a promise the test resolves itself), so no real time passes: the test asserts the dialog's content is absent while the chunk is parked, asserts the open has not settled (the assertion that fails when the helper's wait is removed), then releases the import and asserts the dialog renders.
- Red, with the helper's wait removed (pre-fix behaviour): `expected true to be false` once, and before the parked-import revision, `Unable to find an element with the text: Already configured access on this host?` at 1.33s — the reported failure shape. Both independently reproduced by the PR author, not just the implementing agent.
- The same removal fails the sibling site too: `Unable to find role="button" and name "All providers"`, alongside the Vite re-entrancy error above.
- Green: `npx vitest run src/panes/session/chrome/` — 23 files, 482 tests, passing (five consecutive runs while iterating). `npx biome check` on all three files: clean. `npx tsc --noEmit`: exit 0.

## Cost

None: the parked import removed the 2.5s sleep an earlier revision of this branch used. The chrome suite's wall clock went from 15.9s to 11.5s, and neither the red nor the green direction depends on real time any more (red 4/4 at 224-502ms, green 3/3 at ~406ms in-test).


obra:
The Medium is right and is fixed in 72d0e6c89; the High is wrong, and here is the evidence.

**High — refuted.** `@testing-library/user-event` does publicly export the type at the version this tree installs: `node_modules/@testing-library/user-event/dist/types/index.d.ts:3` is `export type { UserEvent } from './setup/setup';`. The lane's `npx tsc --noEmit` exits 0, and CI's `web` job — which runs the typecheck — passed on the reviewed head, so the typecheck-failure prediction does not hold. The import stays as it is.

**Medium — accepted, and it made the test better rather than just faster.** The 2.5s sleep is gone; the seam now hands the test a *parked* import (`setConnectDialogImporterForTests` returning a promise the test resolves itself). The test proves the property structurally instead of on a clock:

- it starts the open and asserts the dialog's content is absent while the import is parked — the exact state the pre-fix flow asserted through and timed out on;
- it asserts the open has *not* settled (eight zero-delay event-loop turns, no sleeps) — that is the assertion that fails when `openConnectDialog`'s wait is removed;
- it releases the import, awaits the open, and asserts the dialog renders.

Verified both ways here: with the wait removed the test fails 4/4 with `expected true to be false` in 224–502ms; with it, 3/3 green at ~406ms in-test (was 2865ms), and the chrome suite's wall clock dropped from 15.9s to 11.5s. So the suite no longer depends on real time either in the red or the green direction.

The branch has also been updated from main (eb176bc84), which had moved twice since it branched.

PR #1374, opened 2026-09-15
perf(agent): format watch projections outside jm.mu on the status and thread-list reads

### LINKED ISSUE #1367: agent: watch status reads format and sort delivery instants while holding jm.mu (#1208 follow-up)
Follow-up named in #1208 rather than silently dropped.

`liveWatchStatusesForSession` (agent/job_watch.go) holds `jm.mu` for its whole body, and inside that lock it calls `watchStatusInfoFromConfig` for every visible watch -- which formats the watch's `created_at` and up to 32 RFC3339Nano delivery instants -- and then sorts the resulting rows. These projections are read on the hub's status polling path, so the mutex the delivery path needs is held for formatting and sorting that do not read shared state at all.

Fix: collect the values that need the lock (the configs, or the fields to be formatted) into a local slice under the lock, release it, then format and sort outside.

Worth checking at the same time: `liveWatchSummaries` walks the same map under the same lock for `job_list`'s model-facing output, and may have the same shape.

Evidence: `liveWatchStatusesForSession` (Lock/defer Unlock around the loop, the `watchStatusInfoFromConfig` calls, and `sortWatchStatuses`), and `watchDeliveryTimesOf` for the per-instant formatting.

obra:
## What this does

Closes #1367.

Watch projections were formatting and sorting while holding `jm.mu`, the mutex the delivery path needs. The two reads the hub actually makes both did it:

- `liveWatchStatusesForSession` — status polling.
- `appendWatchStatusesByRow` — the production thread-list path (`LiveWatchRowsForSessions`), which was left for a second commit after review pointed out that fixing only the first would leave the hub path unchanged.

Both now walk `jm.watches` and run their visibility/row-keying predicates under the lock, then release it and do the formatting — `created_at` plus up to 32 RFC3339Nano delivery instants per watch — and the ordering outside.

## The hazard this had to handle

Moving the formatting out of the lock is only safe if what gets formatted is no longer the live config. The delivery path mutates `deliveries` and rewrites the `deliveryTimes` ring in place under `jm.mu`, so a projection formatting the live pointer after unlocking would race it. `snapshotWatchConfigForProjection` therefore copies the config and deep-copies the ring inside the lock; everything a projection reads (`deliveries`, the ring, `firedPendingEnd`, `lastClockFire`, `createdAt`) is either copied or set once at install.

The two projections also shared a copy of the visibility walk; it is now one collector (`visibleWatchConfigSnapshots`), which keeps their visibility sets identical by construction. `liveWatchSummaries` (job_list's model-facing projection) had the same shape and got the same treatment.

## Evidence

- Deterministic lock-scope test, no timing: `TestWatchProjectionCollectsUnderTheLockAndFormatsWithoutIt` uses a non-blocking `TryLock` to prove the collector returns with `jm.mu` free, and asserts the collector hands back private copies — a different pointer from the live config, and a delivery ring that does not alias the live one.
- `TestAppendWatchStatusesByRowKeepsRowKeysAndFormatterOutput` pins the thread-list fan-out: receiver's row when the watch has one, the manager's own session's row when it does not, each carrying what the projection always carried.
- `TestFormatWatchStatusesIsPureAndOrdersRows` pins the extracted formatter's exact output, including the RFC3339Nano instants, and the two `...ProjectionMatchesPureFormatter` tests pin that the manager walks feed it exactly the rows they used to publish.
- `gofmt -l` on both files: clean. `go build ./...`: exit 0. `go vet ./agent/...`: exit 0. `go test -short -count=1 ./agent/...`: all ok (369s).

## Honest scope note

This is a behavior-preserving change, so there is no red-then-green test to show; what the tests pin is that the published rows are unchanged and that formatting happens on copies outside the lock. A quick audit of the remaining `watchStatusInfoFromConfig` callers found no other production caller under the lock. Six other lock-held formatters remain in this file (`watchListToolResult`, `watchListToolResultForReceiver`, `inspectWatchByID`, `inspectReceiverWatchByID`, `liveWatchSummariesForReceiver`, `recentWatchSummaries`); they are tool/inspect paths rather than the polling path this issue is about, and are left untouched deliberately.


obra:
Roborev found nothing here, but CI's lint job did: two `modernize`/rangeint findings in the counted loops the new projection tests added (`agent/watch_status_projection_test.go`, `for i := 0; i < 3; i++`). Fixed in b52250094 by ranging over int; `golangci-lint run ./agent/...` now reports 0 issues locally. That one is on me rather than the implementing lane: my local verification of this change ran gofmt, the focused package tests and `go build`, but not the linter, so the finding reached CI.

The branch has also been updated from main (951f6e73a), which had moved twice since it branched.

PR #1377, opened 2026-09-15
fix(hub): saturate sshconn option arithmetic and correct the preflight contract comment (04a follow-up)

obra:
Immediate follow-up for the two Low findings left when #1283 (component 04a, the SSH connection manager) landed; the round-17 review called the rest "otherwise clean, well-structured code".

## Round-17 Lows from #1283

1. **`manager.go:177-181, 1063-1069` — unchecked duration arithmetic.** Large exported option values could overflow when computing the attempt limit or doubling the reconnect delay, producing negative/immediate timeouts and a tight reconnect loop.
2. **`preflight.go:161-162` — doc comment contradicted the contract.** It claimed "Every failure is a named, terminal error class: an unreachable host surfaces `ErrSSHStart`", but `ErrSSHStart` (and `ErrSSHAuth`) are deliberately non-terminal and an unreachable host is retried.

## What changed

- `manager.go`: added `maxDuration`, `saturatingAdd`, and `saturatingMul`; `attemptLimit` now computes `saturatingAdd(saturatingMul(connectTimeout, 4), initTimeout)` and `nextBackoff` saturates its doubling instead of wrapping. No path can now produce a negative or zero delay/limit.
- `preflight.go`: comment-only rewrite — "Every failure is a named error class: an unreachable host surfaces the retryable `ErrSSHStart`, while a contract violation surfaces a terminal class." No behavior change.

## Tests that pin them

`manager_test.go`:
- `TestAttemptLimitSaturatesOnHugeConnectTimeout` — `ConnectTimeout` at/above `MaxInt64/4` still yields a positive attempt limit at least as large as the connect timeout.
- `TestNextBackoffSaturatesInsteadOfOverflowing` — delays seeded near `MaxInt64` stay positive and bounded by `limit`.

Both are failing-first: against the pre-fix code they report `attemptLimit ... = 29.999999996s, want at least the connect timeout` and `nextBackoff(...) = -2ns, want a positive duration`.

## Gates

- `gofmt -l cmd/evener-hub/internal/sshconn/` → empty
- `go build ./...` → ok
- `go vet ./cmd/evener-hub/...` → ok
- `go test ./cmd/evener-hub/internal/sshconn/ -race -count=3` → ok (6.9s)
- `go test ./cmd/evener-hub/internal/sshconn/ -race -count=10` → ok (14.4s)
- `go test ./cmd/evener-hub/... -count=1 -timeout 30m` → ok (3m47s)
[...]

PR #1385, opened 2026-09-15
test(web): make the skillguard steered-turn guard load-deterministic

obra:
## Observed flake

`web-skillguard` failed on CI (run 35016147636, job 104539950935, PR #1377 — a Go-only change that cannot affect the frontend):

```
FAIL web-skillguard (exit 1)
--- FAIL: TestSkillComposerBrowser (31.93s)
skill_composer_browser_test.go:340: skill browser driver failed: node scripts/skillguard/run.mjs exited exit status 1
skillguard FAIL: timed out after 15000ms waiting for steered turn's input visible in the transcript
```

The same guard flaked repeatedly on a loaded box and passed on re-run — timing/ordering, not a product regression.

## Root cause

The timed-out wait is the steered-turn input wait: pre-fix `cmd/evener-hub/frontend/scripts/skillguard/run.mjs:1191-1194`, issued through `waitPage` with no `timeoutMs`, so it inherited the bare default deadline of 15000ms (`run.mjs:522`).

It is **not** merely a tight deadline. Reproduced under load with the artifact dump kept, the on-screen state at the timeout was:

```
composer: {"text":"","steerVisible":true, ...}
queue strip: {"header":"Queued messages (1)",
              "rows":[{"text":"PROSE_STEER_TURN_14e open a long turn for steering ...",
                       "buttons":["Steer now","Edit message","Remove from queue"]}]}
```

The steered input was never dispatched as a turn — it was routed into the **client-side queue**, so it can never render as a transcript turn and the wait can never be satisfied.

Why it was queued: the scenario arms a daemon-side `hold` (which captures the **next** provider request) right after the preceding "queue edit / return / drain" scenario. That scenario's drain dispatches its steering leg only after the held first leg returns, so there is a window between the first leg completing and the drain leg being dispatched. During that window the SPA's busy predicate `isTurnActive(status.type, activeTurnId)` (appwire-client `submitRouting.ts`) legitimately reads not-busy, and the turn-end barrier `waitForTurnIdle` (pre-fix `run.mjs:919-924`) released on a **single** such reading. The hold was then armed into the gap and stole the pending drain leg; that turn never ends, so the next `Send` sees an active session and routes to the queue instead of `turn/start`.

The guard's own comment (`run.mjs:966-976`) describes exactly this failure mode and the barrier meant to prevent it; the barrier was a one-shot poll of a condition the app can flip back, so under load it released in the gap.

## Fix

`cmd/evener-hub/frontend/scripts/skillguard/run.mjs`:

1. `waitPage` gains a `settleMs` option: the condition must hold **continuously** for that long before the wait resolves, so a transient reading cannot release an irreversible action. Default 0 preserves existing behaviour.
2. `waitForTurnIdle` — the barrier every `hold` is armed behind — now requires the idle reading to settle (`TURN_IDLE_SETTLE_MS`, default 3000ms, overridable with `SKILLGUARD_TURN_IDLE_SETTLE_MS`).
3. The steered-turn input wait gets a dedicated, generous, named, env-overridable budget (`STEERED_TURN_INPUT_TIMEOUT_MS`, default 60000ms, `SKILLGUARD_STEERED_TURN_TIMEOUT_MS`) instead of `waitPage`'s bare 15s default, and moves into `waitForTranscriptInput`, which on timeout dumps the transcript turn-blocks, the queue strip and the composer — so a future failure says *what was on screen*, not just which wait timed out.

No product/frontend code changed.

## Assertion preserved

The guard still requires the steered turn's input to genuinely appear as a rendered turn-block in the transcript, and still fails if it never does. A settled/longer wait does not accept a queue-routed or absent input.

## Failure-mode check

Temporarily pointed the wait at a string that never appears in the transcript (`SKILLGUARD_NEVER_PRESENT_14z`): the guard FAILED with the new diagnostic dump (turn-blocks / queue strip / composer) rather than hanging or passing vacuously. Reverted; a `git diff` against the pushed commit shows only the real fix.

## Run tally

FAILING-BEFORE — unmodified code, loaded box (load ~47-57 / 16 cores): 2 of 8 runs failed. One was the exact CI failure reproduced:

```
skillguard FAIL: timed out after 15000ms waiting for steered turn's input visible in the transcript
queue strip rows: PROSE_STEER_TURN_14e open a long turn for steering (Steer now / Edit message / Remove from queue)
```

the other failed at the neighbouring "session B rendered as ended" wait (same class: a fixed-deadline wait on a cross-process round trip).

WITH FIX — 5 consecutive runs, `go test -tags browserguard ./cmd/evener-hub -run '^TestSkillComposerBrowser$|^TestSkillGuard' -count=1` (the exact command `scripts/web/test-web-browser.sh` runs for this guard), box at load ~50:

```
run 1: PASS (92s)
run 2: PASS (87s)
run 3: PASS (84s)
run 4: PASS (78s)
run 5: PASS (80s)
=== 5 pass / 0 fail ===
```


obra:
Round one addressed (`94c792d`).

- **Medium — `run.mjs:987` (`steerVisible === false` is not authoritative).** `steerVisible` is the DERIVED busy predicate (`status.type === "active" && activeTurnId`), so it is also false when the wire reports an active session with no `activeTurnId` — which is exactly what a session holding queued/steering work reports, and where the drain's undispatched leg lives. A fixed settle could therefore still release the barrier onto that leg. `waitForTurnIdle` now gates on two authoritative readings: the session's RAW wire status (the pane Cadence `aria-label`, `"Working"` iff `status.type === "active"`) and queue/pending-work emptiness (the status row's queue indicator renders only while `queueDepth > 0`). A session still working or still holding queued work can no longer release it; `steerVisible === false` remains as an additional (now redundant) requirement, so the condition is strictly stronger, not weaker. The settle window is unchanged.
- **Low — `run.mjs:1078` (`waitForTranscriptInput` hardcoded the composer dump).** The helper now takes `ref` (defaulting to `this.sessionA`, its current call site) and passes it to `composerState`, so a reuse for session B reports B's composer rather than silently printing A's.

5-run tally (`go test -tags browserguard ./cmd/evener-hub -run '^TestSkillComposerBrowser$|^TestSkillGuard' -count=1`): 5/5 `ok` — 59.1s, 57.3s, 56.9s, 54.7s, 55.0s.

Failure-mode check: with the steered-turn wait temporarily pointed at a string that can never appear (`skillguard-impossible-probe-never-appears`), the guard FAILED (exit 1) with its diagnostic dump rather than hanging or passing vacuously — `timed out after 60000ms waiting for input "…" visible in the transcript`, followed by the transcript turn-blocks, `queue strip: null`, and the composer state. Probe reverted; `run.mjs` contains no trace of it.

Verification: 5/5 clean `go test -tags browserguard ./cmd/evener-hub -run '^TestSkillComposerBrowser$|^TestSkillGuard' -count=1` passes; failure probe reproduced and reverted; `node --check run.mjs` clean; diff limited to `cmd/evener-hub/frontend/scripts/skillguard/run.mjs`.

obra:
Round two addressed (`2dda7bc5c`).

Both findings were treated as claims and verified against the code and the live app before anything changed — a temporary in-driver probe sampled the session pane every 120ms across the drain release on this host (instrumentation removed before commit).

- **`run.mjs:1007` — the barrier only checked `status-row-queue`, i.e. the wire queue depth (`model.queue.depth`).** Confirmed live: immediately after the drain release, the drain's in-flight mutation was on screen as `[data-testid='pending-chips']` while `status-row-queue` was absent — exactly the state described. The barrier now additionally requires the session's own pending-work surfaces to be clear: `pending-chips` (PendingChips' optimistic/durable/authoritative send/steer/drain entries — the projection of the same outbox and `pendingMutations` state the durable reads use) and the "Queued messages" strip (wire queue rows, pending queue rows, and recovery/blocked records). Both are pane-scoped, so session B's work cannot mask A's.
- **`run.mjs:1012` — the fixed 3s settle could not guarantee the drained turn's next provider leg had been dispatched.** Confirmed live: the release answers the HELD leg first; the daemon injects the drained steering turn at that turn boundary and only then dispatches the continuation leg, so the first reply is not the turn's end (probe: the drained turn-block appeared 0.15s after release and carried its reply 0.12s later, while a status-only reading had already started settling). Replaced the timing assumption with a positive event: `waitForContinuationReply(text)` waits for the continuation's OWN turn-block to carry the scripted reply — the leg's completion, tied to that exact input. It runs after the drain release (`PROSE.queue2`) and after the steer release (`PROSE.steer`), before the barrier; the wait lasts exactly as long as the daemon takes, so no load-induced gap can outlive it. The 3s settle remains only as a backstop for the single-leg barriers, and the code comments now say so plainly instead of presenting it as the guarantee.

No assertions were deleted or weakened; the Go owner's provider/transcript assertions are untouched.

Verification (lane `dlg_034PUeHV8MCHp8JtkEb930`, host load average 42–69 on 16 cores throughout):

- `node --check cmd/evener-hub/frontend/scripts/skillguard/run.mjs` — exit 0.
- `env -u DBUS_SESSION_BUS_ADDRESS go test -tags browserguard ./cmd/evener-hub -run '^TestSkillComposerBrowser$' -count=1` — PASS 5/5: three consecutive runs on the final tree (53.3s / 53.1s / 56.4s) plus two probe-instrumented runs (55.6s / 55.4s).
- `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` — final run all green (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard and web-skillguard; exit 0). Earlier invocations on this shared box hit the host flakiness the guard-process cleaner exists for: one failed `web-layoutguard` on a transient Vite module fetch (`editorial-transcript-390`: "Failed to fetch dynamically imported module …/transcript-guard.ts"), one failed `web-transcriptscrollguard` on its fixed 10s scroll-settle deadline (`transcriptscrollguard-entry.tsx:342`), and one hit the harness's own "browser startup deadline exceeded after 30000ms" environment path in `web-skillguard` before any scenario code ran. `web-skillguard` passed in every one of those invocations, neither failing guard reads anything from `scripts/skillguard`, and after reaping 101 leaked guard Chrome/vite processes (70 Chrome + 31 vite, 7.8 GB RSS; `bash /tmp/evener-sandbox-657721092/clean_guard_leftovers.sh`) the next full run was green.

obra:
Round three addressed (`9598ff34`).

- **Medium — `run.mjs:1052-1054`: `waitForContinuationReply` could return on the HELD leg's reply.** Confirmed by reading the predicate rather than the timeline: it was "some turn block holds the continuation text and *a* sentinel reply", and every scripted leg answers with the same sentinel, so the check proved nothing about the continuation leg on its own. What made it sound was the app currently rendering the continuation in a turn block of its own — a layout fact, not a published ordering. It is now held to the published ordering instead: the daemon announces `EventTurnStarted` for the continuation's turn *before* `EventSteeringInjected` (a steering item opens no turn, which is why — `agent/session_lifecycle.go`'s `acceptNotificationInput` / `acceptSteeringCarrierInput` emit in that order, and `injectDrainedSteering` consumes the message only after), so the continuation lands in a turn whose block did not exist when the release was issued, and its own reply is recorded after the steering item inside that block. `waitForContinuationReply(text, turnsBefore)` now requires **both** (1) a turn block whose `data-turn-id` is not in `turnIds()` — captured beside the existing `replyBaseline()` before each release — and (2) the continuation input followed by a reply inside it, with no reply-count or timing assumption. Both call sites (drain, steer) pass their baseline.
- **Low — `run.mjs:1061-1073` & `:1160-1171`: duplicated failure dumps.** Extracted one `dumpPageState(ref, error)` helper; both `catch` blocks call it, and the first no longer hardcodes `this.sessionA` (`waitForContinuationReply` now takes `ref`, defaulting to sessionA like `waitForTranscriptInput`).

**How the Medium's failure mode was demonstrated (not just asserted).** A temporary in-page probe ran inside the guard's own fixture, on the release edge: it polled the page (5 ms) until the held leg's reply rendered into the held turn's own block, then — at that instant, with the census showing that **no** turn block anywhere carried the continuation text (its turn did not exist, so nothing had been dispatched) — it injected the continuation text into the held block and evaluated both predicates in the same synchronous task, then restored the DOM:

| probe run (`PROSE_QUEUE_14c second pass` / `PROSE_STEER_14e …`) | census at injection time | old predicate | new predicate |
|---|---|---|---|
| continuation text injected into the held block, after its reply (the review's shape) | held block `turn_m2`/`turn_m4` 1 reply; continuation text on page: **0 blocks** | `true` | `null` |
| same text injected before the reply | same census | `true` | `null` |

So the old predicate released on a page state where the continuation leg's request had not even been dispatched; the new one refuses in either item order. This reproduces what the review named, on the live page — the probe did not have to be reverted to a sleep or a shim, and the guard run it was embedded in passed. Note the ordinary timeline observation, for the record: on this head the continuation does render in its own turn block, whose first frame carries **zero** replies, while the held leg's reply sits in the block before it.

The timeout path was exercised too (default timeout temporarily set to 1 ms): the guard fails loudly with the extracted dump, e.g. `timed out after 1ms waiting for the continuation leg carrying "PROSE_QUEUE_14c second pass" to complete in its own turn / transcript turn-blocks: […] "YounowPROSE_QUEUE_14c second pass" / queue strip: null / composer: {…}` — the continuation block carrying the prose and no reply, which is exactly the state the wait now refuses.

**Verification:** `node --check cmd/evener-hub/frontend/scripts/skillguard/run.mjs` → exit 0. `env -u DBUS_SESSION_BUS_ADDRESS go test -tags browserguard ./cmd/evener-hub -run '^TestSkillComposerBrowser$' -count=1` → **4/4 consecutive passes** on the final tree (53.343s, 57.703s, 60.334s, 56.360s test time; 56.7s/60.8s/63.5s/59.9s wall) at host load 65.9–82.6 on 16 cores, sharing the box with sibling agent suites. `make test-web-browser` → PASS web-layoutguard, web-overflowguard, web-shellguard, web-spawnguard, web-transcriptscrollguard, web-skillguard (214.9s total). The change touches no `src/` files and no rail internals — only `cmd/evener-hub/frontend/scripts/skillguard/run.mjs`.

obra:
Round four addressed (`aaa1ddab4`).

- **Medium — `waitForTurnIdle` ignored the active recovery draft.** Verified against the frontend: `QueueStrip.tsx:175-177` filters the record whose `clientMutationId` is `activeRecoveryId` out of the strip, and `Composer.tsx:647-685` restores a rejected record into an EMPTY composer and takes ownership of it (`Composer.test.tsx:2137`), so the strip cannot see recovery work the composer owns. `waitForTurnIdle` now also reads the session's recovery records from the store that owns them (IndexedDB, scoped by `targetRef`, `notes/human/set` excluded): an unreadable store keeps the barrier closed and so does a record present. The retry path additionally clears the composer first (`clearComposerDraft`) and holds it empty until the same durable store is empty, so a restored draft can neither be appended to nor restored again (the restore effect only fires while a record exists).
- **Medium — transcript waits and turn-id matching were not scoped to the session pane.** `turnIds()`, `waitForTranscriptInput` and `waitForContinuationReply` now resolve `[data-pane-scaffold='session:<ref>']` and query only inside it; every call site passes `driver.sessionA`.
- **Medium — baseline evaluation failure silently converted to `[]`.** `turnIds()` now throws — before the release that dispatches the leg it is the baseline for — on a missing pane, a null read, or a page error. The identical fail-open shape was removed from the sibling `replyBaseline()`, which guards `waitForReply` (same class, same file; disclosed as an extension of the finding).
- **Low — continuation order check sliced from `at + 1`.** Now `block.slice(at + text.length)`.
- **Low — `waitPage` settle timer.** An evaluate error no longer resets the settle hold (only a reading that came back does, including one that answers "not yet"), and the deadline is extended once, by the settle window, when the condition first holds — a condition first held near the end still gets its whole window, and a flapping one cannot postpone the deadline repeatedly.
- **Low — `envMillis` whitespace.** Blank (`raw.trim() === ""`) now takes the default instead of being parsed as `0`.

Falsification, one instrumented live run against the real app + hub + daemons (instrumentation removed before the commit; raw evidence `/tmp/evener-sandbox-2633186373/skillguard-demo.jsonl`):

- **Recovery**: a rejected recovery record seeded into the app's OWN IndexedDB store (`evener-mutation-outbox.recovery`, same shape `transferToRecovery` writes) was picked up by the app's discovery interval and restored by the app itself into the composer (`{"restored": true, "value": "PROSE_FAIL_14h request the missing source"}`), with the strip rendering nothing (`stripAfterRestore: null`). The pre-fix barrier predicate read `true` over that state (the reported false release) while the new one reads `null`, and `noPendingRecoveryExpr` reads `false`. `clearComposerDraft` then took 156ms: composer text `""`, chips `[]`, durable `recovery: []`, after which the new predicate reads `true`.
- **Scoping**: with a second `session:<ref>` pane scaffold in the document (the app's own stack-host shape; measured in the same run, this desktop host mounts ONE session pane at a time — opening session B unmounts A) holding a turn block whose id is absent from the baseline and whose text carries the continuation text plus the reply sentinel, the old document-wide predicate read `true` and the pane-scoped one `null`; the new scoped `waitForTranscriptInput` held on that decoy until its 1.2s timeout.
- **Baseline**: on a REAL pre-existing turn block (`turn_m1`, carrying `PROSE_ALPHA_14a` and the reply sentinel), the continuation predicate with the real baseline reads `null`, while the same predicate with `[]` — exactly what the removed `.catch(() => [])` produced on a read error — reads `true`, i.e. it would release on a turn that was already running.

Verification:
- `node --check cmd/evener-hub/frontend/scripts/skillguard/run.mjs` → exit 0.
- `env -u DBUS_SESSION_BUS_ADDRESS go test -tags browserguard ./cmd/evener-hub -run '^TestSkillComposerBrowser$' -count=1` ×4 consecutively → PASS ×4, 48.62s / 52.67s / 52.95s / 52.05s, 1-minute load 49.4 / 45.7 / 48.5 / 51.9 on 16 cores (a previous four-run pass pattern on the same code before a comment-only edit: 46.17s / 49.64s / 51.63s / 51.54s at load 39.9 / 43.6 / 47.1 / 46.5).
- `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` → exit 0: `PASS web-layoutguard`, `PASS web-overflowguard`, `PASS web-shellguard`, `PASS web-spawnguard`, `PASS web-transcriptscrollguard`, `PASS web-skillguard`.

obra:
Round five addressed (`f7a718561`).

Verdict `aaa1dda`, all four Mediums and both Lows, one file
(`cmd/evener-hub/frontend/scripts/skillguard/run.mjs`).

**M1 — settle window not reset on evaluation errors (`waitPage`).** Error time was unobserved wall clock that still counted toward the `settleMs` continuous-idle proof. The window is now shifted forward by the span since the last reading that came back: the successful idle time already accumulated is kept (Round 4's "a hiccup must not restart the hold"), the failure itself is excluded from the proof, and the deadline moves by exactly that lost span so an interruption near the edge cannot turn into a timeout. Pinned (harness, head vs final, virtual clock): held@0 / error@400 / held@1200 — head releases at 1200 (the gap counted as idle proof), final at 1400; held@0+500, a 600ms failing poll, held@1500+ — head 1500 (failure counted), final 1600 (the 500ms accumulated kept; the reset-on-error Round 4 forbade would have needed 2100).

**M2 — `turnIdleExpr` ignored durable outbox/optimistic records.** New `noPendingTurnRecordExpr(ref)` reads the same durable stores the pending surfaces are projected from (outbox + optimistic), scoped to the session, excluding `notes/human/set` (the one method the app itself never treats as pending turn work — QueueStrip and Composer both filter it by that exact name). `turnIdleExpr` requires it true, like the recovery read; an unreadable store is not a release. Pinned (jsdom + fake IndexedDB): with an idle-looking pane (no chips, no strip — the render-lag case) and a durable `turn/steer` record, head reads idle (`true`) and the final reads `null`; the same pane reads `true` once the record is gone; another session's record and a notes record do not block; a store that cannot be read yields `null`.

**M3 — `turnIds()` / `replyBaseline()` hard-failed on the first transient rejection.** Both now go through `readTurnIdBaseline`, which retries until the read comes back or `TURN_BASELINE_RETRY_MS` expires (45s, `SKILLGUARD_TURN_BASELINE_TIMEOUT_MS`; longer than the wire's own 30s bound on one `Runtime.evaluate`, so a stalled call is retried too), then fails loudly naming the last reading or the last error. A failed read is still not an empty baseline. Pinned: two transient rejections then success — head throws, final returns the ids; an always-failing read still throws after the budget; a null read still fails.

**M4 — `clearComposerDraft` never cleared attachment tiles.** Tiles are now removed through their own `Remove <name>` controls (`removeAttachmentTile`), and the completion condition — now the named `composerDrainedExpr` — requires `state.tiles === 0` as well as empty text and chips. Pinned (jsdom frame whose recovery discard mirrors `queueRecoveryPersistence`): with a recovery carrying a tile, the head helper times out (the record can never be discarded, so `noPendingRecovery` never reads true) and leaves the tile staged; the final returns with text, chips and tiles empty and the record out of the store.

**L1 — `dumpPageState` used document-wide queries.** `transcriptBlocksExpr(ref)` and `queueStripExpr(ref)` now take an optional ref through the new `paneOrDocumentExpr` (the waits that ask about the app's queue keep the document-wide default), and the dump passes the session it reports about. Pinned: with two panes mounted, head's dump for session A quotes session B's turn text and queue row; the final's does not.

**L2 — error swallowed if `dumpPageState` returns normally.** `dumpPageState` now returns the caller's error with the page state appended, and both call sites do `throw await this.dumpPageState(ref, error)`. Propagation is total and the diagnostics are kept. Pinned: with a dump stub that returns normally, head's `waitForTranscriptInput` resolves (timeout swallowed); the final's rejects with the original error.

Verification:
- `node --check cmd/evener-hub/frontend/scripts/skillguard/run.mjs` — exit 0.
- `env -u DBUS_SESSION_BUS_ADDRESS go test -tags browserguard ./cmd/evener-hub -run '^TestSkillComposerBrowser$' -count=1` × 5 consecutive on the final tree: 5/5 pass — 52.2s, 45.5s, 50.1s, 50.1s, 50.1s (`ok primeradiant.com/evener/cmd/evener-hub`), 1-minute load avg 44–57 on 16 cores; every run reported all nine scenario groups (canonical, draft-remount, queue+drain, steering, attachment, capability-loss, failed-activation retry, delayed-send, transport-loss).
- `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` — 177s: `PASS web-layoutguard`, `PASS web-overflowguard`, `PASS web-shellguard`, `PASS web-spawnguard`, `PASS web-transcriptscrollguard`, `PASS web-skillguard` (load 50.9).
- Falsification harness (session scratch, not committed): stages the final and `aaa1ddab4` files side by side (only the unconditional `main()` tail replaced by an export) and drives both with scripted CDP, a virtual clock, jsdom and fake IndexedDB — 26/26 expectations held; the head-vs-final discriminators for each finding are the ones quoted above.

obra:
Merged `main` (`d1696a8e5`) to resolve the `run.mjs` conflict with #1439 (`f8a900f54`).

- **Kept from #1439:** `turnTextsExpr(ref)` (one session's mounted transcript text per turn, pane-scoped, a turn's segments concatenated) and the turn-scoped `waitForReply(ref, prose)` with every call site, plus the re-click notice in `clickQueueStripControl`. Because that reply wait is baseline-free, it **supersedes the `replyBaseline` half of round five's M3** — the method it retried no longer exists, and the transient tolerance that retry bought now comes from `waitPage`'s own error handling (a failed poll neither releases nor fails the wait). The `turnIds` half of M3 — the baseline `waitForContinuationReply` still needs — keeps its retry (`readTurnIdBaseline`, 45s budget).
- **Kept from round five:** M1's settle accounting (`waitPage` shifts the settle window and the deadline past an unobserved error span, so error time is neither idle proof nor a timeout), M2's durable outbox/optimistic read in `turnIdleExpr`, M3's retry for `turnIds`, M4's tile clearing (`removeAttachmentTile` + `composerDrainedExpr` requiring `tiles === 0`), L1's session-scoped dump reads (`transcriptBlocksExpr(ref)` / `queueStripExpr(ref)`), L2's rethrow after the dump. The drain and steer scenarios keep their pre-release `turnIds` captures feeding `waitForContinuationReply`, alongside #1439's reply waits.
- **Adapted so both sides survive:** `waitForTranscriptInput` keeps its generous bound and its failure dump that rethrows (round five) and now scans with `turnTextsExpr(ref)` — #1439's per-turn, pane-scoped read — instead of its per-element sweep; taking either side verbatim would have dropped the other.

Verification (merge commit `01b7a8258`; the frontend dist was rebuilt from the merged sources first, so #1439's Composer caret fix is in the page the driver drives):
- `node --check cmd/evener-hub/frontend/scripts/skillguard/run.mjs` — exit 0 on the committed blob.
- `env -u DBUS_SESSION_BUS_ADDRESS go test -tags browserguard ./cmd/evener-hub -run '^TestSkillComposerBrowser$' -count=1` × 3 consecutive on the merged tree: **3/3 pass** — 64.7s, 50.2s, 51.0s (`ok primeradiant.com/evener/cmd/evener-hub`), 1-minute load avg 52–61 on 16 cores.
- `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser`: 166s at load 58.4 — `PASS web-layoutguard`, `PASS web-overflowguard`, `PASS web-shellguard`, `PASS web-spawnguard`, `PASS web-transcriptscrollguard`, `PASS web-skillguard`.
- Falsification harness re-run against the merged file: 26/26 expectations still hold (M1's head-vs-final release points, M2's durable-record discrimination, M3's retry and bounded failure, M4's tile clearing, L1's scoping, L2's rethrow); its M3 demo exercises `turnIds`, the half of M3 the merge keeps.

obra:
Round six addressed (`a658e1b5f`).

**Medium — `waitPage` could never time out when evaluations failed repeatedly (`run.mjs:624-631`).** Confirmed: `readAt` advanced only on a successful read, so each failure in a run measured `now - readAt` from the same stale point and added that whole span to both `heldSince` and `deadline`. With `deadline ≈ D0 + Σ(t_i − R)` while `now = t_n`, the deadline outran wall clock (quadratically), the timeout never fired, and a sustained failure hung the wait; the same stale point could also shove the settle window's start past the current poll, making a release impossible.

**Accounting rule now:**
- A failed poll is unobserved wall clock that is excluded from the proof and refunded to the deadline, measured from the **previous poll attempt**, never from the last success — `readAt = now` on every attempt, success or failure. Each failure therefore contributes exactly one interval, and `heldSince ≤ now` always holds (the window cannot be pushed into the future).
- The refund is bounded by the settle window it protects (`refundLeft`, initialised to `settleMs`): while unobserved time is cheap the deadline moves with the window, so a hiccup inside a settle window still cannot cause a timeout, but once excluded time has cost as much as the whole proof needs the deadline stops moving — an evaluate that never comes back now fails the wait at the caller's budget plus at most one settle window instead of extending it forever.
- Unchanged and re-pinned: error time is never idle evidence, an error never restarts the hold (round four), and the accumulated successful-idle time survives a failure (round five).
- The timeout message now reports the wall time actually waited (`timed out after 500ms (waited 880ms)`), so an extended wait is not misreported as a bare budget miss.

**Pinned (harness, pre-fix `01b7a8258` vs fixed, deterministic virtual clock):**
- **D2 sustained failure inside a settle window** (hold at t=0, then a failure every 80ms; budget 500ms, settle 300ms), each side in its own process: pre-fix **never fails** — killed at a 1.5s bound (hang reproduced); fixed fails at t=880ms = budget + one settle window + one poll.
- **D1 sustained failure with no settle window**: fails at the budget (t=560ms).
- **E run-of-failures over-shift**: two consecutive failures inside a 500ms settle window — fixed releases at exactly t=700, the point where error-excluded elapsed time reaches `settleMs` (computed from the step stream by the rule, not from the code); pre-fix removes the interval since the last success twice and releases late, at t=800.
- Round-five cases unchanged: M1-A (head 1200 → fixed 1400), M1-B (head 1500 → fixed 1600, reset-on-error 2100), M1-C, M2, M3, M4, L1, L2 — **31/31 expectations hold**.

Verification:
- `node --check cmd/evener-hub/frontend/scripts/skillguard/run.mjs` → exit 0.
- `env -u DBUS_SESSION_BUS_ADDRESS go test -tags browserguard ./cmd/evener-hub -run '^TestSkillComposerBrowser$' -count=1` × 3 consecutive: **3/3 pass — 48.2s, 50.1s, 47.8s** (`ok primeradiant.com/evener/cmd/evener-hub`), 1-minute load avg 44–55 on 16 cores.
- `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser`: 170s at load 50.4 — `PASS web-layoutguard`, `PASS web-overflowguard`, `PASS web-shellguard`, `PASS web-spawnguard`, `PASS web-transcriptscrollguard`, `PASS web-skillguard`.

obra:

**CI** — run 35064922087 at `a658e1b5f`: completed/success, all jobs green.

**Review** — RoboRev's verdict for `a658e1b5f`: "No issues found."

**Up-to-date override** — the branch is 8 commits behind `main`. Overlap check, run immediately before this comment:

```
$ gh pr view 1385 --json mergeable,mergeStateStatus --jq '.mergeable, .mergeStateStatus'
MERGEABLE
BEHIND

$ comm -12 <(gh pr view 1385 --json files --jq '.files[].path' | sort) \
           <(git diff --name-only "$(git merge-base origin/main origin/fix/skillguard-steered-turn-race)" origin/main | sort)

$ git diff --name-only "$(git merge-base origin/main origin/fix/skillguard-steered-turn-race)" origin/main | wc -l
279
```

`main` changed 279 files since the merge base; none of them is the file this PR touches (`cmd/evener-hub/frontend/scripts/skillguard/run.mjs`).

Landing this now is deliberate: this guard is the one every frontend PR in the multi-host stack currently reports as environment-blocked under load, so landing it unblocks those browser gates.

PR #1392, opened 2026-09-15
fix(hub): make explicit Resume completion-owned and cancellable

obra:
## Scope

An explicit Resume on a giant session must wait for a real restore instead of being abandoned at the ordinary startup budget, and cancelling it must be safe.

- An explicit `thread/resume` awaits completion, child exit, or caller/Stop cancellation rather than guessing a duration. Heartbeat health checks stay active; the ordinary RPC, automatic-resume, and fresh-spawn budgets are unchanged.
- A Stop or a disconnect cancels that wait and coordinates with actual child reaping, so a Stop cannot report success over a child that is still alive.
- Alias acquisition in context-aware paths uses `LockContext(ctx)` and releases exactly the prefix it acquired, so a cancelled or disconnected request is not left blocked behind a long resume.
- The destruction paths honour cancellation: a cancelled or deadline-expired session delete or project delete fails rather than answering success-with-skip.
- Prelaunch cancellation diagnostics are retained when a candidate log is cleaned up.
- The confirmed-stopped shortcut rechecks the deletion fence for every alias in the group, and refreshes the roster, inputs, and attention notifications before returning success.

## What this PR deliberately does not do

An earlier revision of this branch also carried a chain of **durable exit-proof** mechanisms: a persisted launch-intent marker with startup reconciliation, a persisted signal-intent marker, a retirement-path launch guard, two recovery-state format version bumps, and the durable launch-path invalidation and restoration of a prior owner's exit proof. That whole chain has been removed from this PR and will be designed separately.

The reason is the review history. Each element of that chain was added to answer a finding about the previous one, and the third review round still returned findings on it. The deepest one says the marker approach is itself unsound: recovery settles a group merely because no rendezvous claim exists, so a hub crash after a child started but before it published its claim leaves a live child whose way to a replacement is cleared. That is a design question about where the authority for "this process is gone" comes from, and it does not belong in this PR.

Removed here, and carried by the separate change: the `launch_pending` and `signal_attempted` fields, `recoverInterruptedLaunches` and its startup wiring, `MarkForceStopSignaled`, the retirement-path `LaunchGuard`, and now also `beginLaunchProofLocked` / `restoreLaunchProofLocked`, `BeforeLaunch`'s durable clear, the settle-path restoration, the `proofSettled` / `proofErr` / `ActiveResume.proof` bookkeeping, and the tests that covered them.

Three consequences worth stating plainly:

- The recovery-state file is **unchanged** by this PR: it still writes version 3 with no additional fields. There is no format migration and no downgrade hazard.
- `main` has no launch-path invalidation at all — no `BeforeLaunch`, no proof helpers. This PR now simply stops adding one, so nothing on this path is a regression against `main`.
- The mid-launch crash window that the invalidation would have introduced is **gone with it**. It is not deferred; it does not exist in this branch. Closing it properly is the separate design change's job.

## Tests that moved with the removed mechanism

- `TestActiveResumeProofPersistenceFailures` (`cmd/evener-hub/internal/hubcore/resumelocks_resume_test.go`) — removed entirely. Its subject *is* the invalidation.
- The recreated-hub block of `TestLongRunningResumeFailedCleanupRetainsOwnership` (`cmd/evener-hub/resume_lifetime_test.go`) — three assertions, all of which assert the invalidation's effect across a hub recreation. That test's retained-ownership assertions are unchanged and still pass.
- `TestActiveResumeReapedCleanupPreservesRecoveryAuthority` — **not touched**. It exercises kept behaviour (retained ownership after a failed cleanup, the Stop epoch advancing, a fresh Resume being refused) and passes unchanged. An earlier revision of this description claimed it depended on the invalidation; that claim was wrong, and the test file does not appear in this commit's diff at all.

## What `BeforeLaunch` keeps

`BeforeLaunch` keeps its non-durable refusal to launch while another in-flight operation on the same aliases has unconfirmed child cleanup. Retained ownership until `ChildReaped` is preserved without the proof flag.

## Review-hardening pass on the 509e37bba findings

- **Project-delete retry cancellation (was Medium):** the existing-deletion retry path now checks `ctx.Err()` after acquiring all ownership reservations and before `cleanupProjectDeletion`, matching the fresh path and `sessionDelete`. Red-first: `TestProjectDeleteRetryCanceledAfterOwnershipReportsError`. The destructive-cleanup/ownership sites were enumerated (`sessionDelete`, both `projectDelete` paths, `resumeProjectDeletions`, `forceStopThread`, `confirmedStoppedWithoutClaim`, `retireDaemon`); this was the only missing guard.
- **Force-stop deletion/cancellation ordering (was Medium):** the confirmed-stopped shortcut takes the per-alias reservations before cancelling when they are free, so its authoritative deletion re-check runs under them before any Resume is aborted; and the verified-process path in `forceStopThread` now does the same at its admission-fence block — all deletion validation under the per-alias reservations completes before `cancelActiveResumes`, because deletion publication takes the same reservations. Red-first: `TestForceStopConfirmedStoppedValidatesDeletionUnderReservationBeforeCancelingResume` and `TestForceStopValidatesDeletionUnderReservationBeforeCancelingResume` (+ `...Fallback` for the cancel-then-acquire order).
- **Concurrent-resume duplicate `Spawner.Resume` (was Medium):** not reproducible as stated; rebutted with evidence in the review thread. Every serving configuration builds its `Roster` unconditionally, so the duplicate-owner re-check under the alias locks always runs and a queued duplicate reuses the first resume's daemon. A clarifying comment now records that at the re-check site.
- **Unbounded client resume (was Low):** `thread/resume` keeps its completion-owned correlation but is bounded by `RESUME_REQUEST_TIMEOUT_MS` (10 min) instead of no timeout, so a hub that never answers can no longer pin `resumePending` forever. One existing timer-count assertion changed `toBe(1)` → `toBe(2)` because a bounded resume request legitimately owns a timer; the surrounding test still proves no timer leaks (count returns to 0 after settle).

## Second hardening round on the 11c9bd147 findings

- **Sibling-alias deletion fence on the reservation fallback (was High):** the fallback branch now re-checks the deletion fence for EVERY alias after acquisition, matching the reservations-held branch and `confirmedStoppedWithoutClaim`; the fallback test's `deletionTargetState` stub is alias-aware and new coverage refuses a sibling-only fence before any kill. Red-first: `TestForceStopValidatesSiblingDeletionUnderReservationFallback`.
- **Failed-child cleanup race in shutdown (was High):** `shutdownThreadTolerateExited` rechecks retained resume cleanup errors under the alias ownership immediately before acting, so a Resume reporting failed cleanup while shutdown waits can no longer yield a success report. Red-first: `TestShutdownRechecksCleanupUnderOwnership`.
- **Resume admission after the confirmed-stopped no-op (was Medium):** the no-op decision now publishes as an admission generation; a resume registration waiting on the alias is invalidated (`ErrResumeInvalidated`) and must re-admit on a post-decision snapshot instead of launching after shutdown reported success. Red-first: `TestConfirmedStopNoOpInvalidatesWaitingResumeRegistration`. Two existing assertions updated as consequences, both flagged for review: the admission-barrier test's tail now expects invalidation plus fresh re-admission (the old tail encoded the reported bug), and the repeated-shutdown test no longer pins Epoch stability (epoch is now the admission generation, deliberately advanced by the no-op; every recovery-authority field is still asserted unchanged).

## Final round on the bc3231879 findings (1 Medium + 3 Low)

- **Expected-daemon identity rechecked after cancellation (was Medium):** the confirmed-stopped path now verifies the current rendezvous identity immediately BEFORE `BeginActiveResumeStop`/`cancelActiveResumes`, so a replacement claim in that window refuses the request without aborting the in-flight Resume; the later ownership recheck is retained. Red-first: `TestForceStopConfirmedStoppedRechecksIdentityBeforeCancelingResume`.
- **Confirmed-stopped shutdown fast path skipped the roster refresh (was Low):** it now refreshes before returning success, matching the force-stop shortcut. Red-first: `TestShutdownConfirmedStoppedRefreshesRoster`.
- **Vacuous-tolerant regression test (was Low):** `TestForceStopStaleExpectedDaemonDoesNotCancelResume` no longer tolerates the old cancellation via `select`/`default`; it fails if the refusal comes after any cancellation and asserts the Resume's context is never canceled.
- **Nested cleanup error (was Low):** `finishLaunch` returns an existing `*resumeCleanupError` unchanged and wraps only unclassified cleanup errors, so the "cleanup is unconfirmed" text is no longer duplicated. Red-first: `TestLongRunningResumeCleanupErrorIsNotNested`.

Gates at `76bbdaae5` (restacked via `5baf25d24`): `make lint`, `make vet`, `go test -p 1 ./cmd/evener-hub/... -count=1` all exit 0; zero test deletions.

## Third round on the 76bbdaae5 findings (4 Medium + 1 Low; 1 Low skipped with reason)

- **Unreachable shutdown fallback (was Medium):** uncertain rendezvous discovery now routes shutdown through a dedicated tolerant path instead of assuming the source fallback is reachable through the resume-required rejection. Red-first: `6e1aa4e1d`'s new test.
- **Identity validation not atomic with cancellation (was Medium):** the expected-daemon identity is now revalidated under the `BeginForceStop` admission fence, after the reservation attempt and immediately before `cancelActiveResumes`. The atomicity claim, stated in the commit message: with every alias reservation held, no Resume is mid-launch and the fence refuses new admissions, so no Resume-originated claim can appear between check and cancellation. The fallback path's disclosed pre-discovery window is explicitly outside the claim and stays subject to the authoritative post-ownership reread. Red-first: `TestForceStopRevalidatesIdentityUnderFenceBeforeCancelingResume`.
- **Cancellation before full alias-group validation (was Medium):** the entry drain now resolves and validates every ownership alias (sibling deletion fences included) before cancelling, with new admissions fenced during discovery via the admission-generation machinery. The pinned mid-discovery registration semantic is unchanged and still passes.
- **Relay publication timeout silently acknowledged (was Medium):** on guard expiry the relay now broadcasts a target resync before acknowledging, with a test holding the alias past the timeout.
- **`timeoutMs` unreachable no-timeout branch (was Low):** dropped; the timer is always set (`bfdafa25c`).
- **Alias-acquire helper extraction (Low): SKIPPED with reason** — three of the five sites interleave the tryLock counter and tracing with acquisition, so a shared helper cannot hold the invariant in one place cleanly; forcing it would obscure rather than share.

Gates at `bfdafa25c` (restacked via `fe86376a8`): `make lint`, `make vet`, `go test -p 1 ./cmd/evener-hub/... -count=1`, `tsc --noEmit`, and the appwire-client vitest (95/95) all exit 0. Zero test deletions.

## Simplify-code pass (quality only; requested after the careful round)

Four read-only angle reviews (reuse, simplification, efficiency, altitude) over the full PR diff, deduped, applied under a no-behavior-change rule (helpers added, call sites delegate; zero test-file edits; revert-not-fix-forward on any gate failure).

- **Applied:** group deletion-fence loops route through one `deletionFenceErrorForGroup` helper; `ResumeCleanupError`/`ResumeCleanupErrorStrict` share one private scan; one reservation-release closure in `forceStopThread`; the expected-daemon revalidation core is shared (call-site wrapping kept where it differs deliberately); shutdown's ownership wrapper is a plain if/else; the transient-ownership wait budget is one named const; the relay publication guard takes a no-timer fast path first; client.ts expresses "completion-owned" as one method set instead of two string comparisons.
- **Reverted:** the alias-group lock helper extraction — `make lint` (gocritic sloppyReassign) failed it, and per the pass's rule it was reverted rather than reshaped (`d453798f7`). The interleaved sites stay hand-rolled, as the careful round judged.
- **Skipped with stated judgment:** the RegisterResume ownership hand-through (touches the hardened hand-off protocol; overlaps the rebutted finding) and the overlap-group recompute (it IS the validation; state can change during the drain).

Gates at `d453798f7`: `make lint`, `make vet`, `go test -p 1 ./cmd/evener-hub/... -count=1`, `tsc --noEmit`, appwire-client vitest (92 files / 1905 tests) all exit 0. Zero test-file edits in the pass.

## Round on the d453798 findings (1 High + 1 Medium)

- **Resume paths did not fence the full ownership group (was High):** explicit and retirement Resume validated deletion fences for only the requested/resolved pair, so a deletion naming a third alias in the resolved group was bypassed. Every path now fences the whole group under the locks that make the check final: `resumeThreadLocked` under the alias locks (subsuming the old requested-only conditional), `resumeThreadLockedLaunch` with the group threaded through `resumeLaunch`, and `resumeAfterConfirmedRetirement` under its own locks. Red-first: `TestExplicitResumeSiblingAliasDeletionFencesOwnershipGroup` and `TestRetirementResumeSiblingAliasDeletionFencesOwnershipGroup` (both reached the launcher pre-fix).
- **The 10-minute client cap did not cancel hub work (was Medium):** when the default completion-owned cap fires, the client now retires the transport (`handleSocketLoss` + close), so the hub observes the disconnect and unwinds the resume wait, reaps, and releases ownership; the reconnect machinery rebuilds for the retry. An explicit caller `timeoutMs` stays rejection-only by design ("an explicit numeric Resume deadline remains authoritative" is untouched). Hub-side unwind is already pinned by `TestLongRunningResumeHubCancelsOwnedChild`'s disconnect half. Red-first: the updated cap test asserts the abandoned transport's `closeRequests` and drives the rebuild before the retry.

Gates at `a246bcd69` (restacked via `bc495f0e8`): `make lint`, `make vet`, `go test -p 1 ./cmd/evener-hub/... -count=1`, and `make test-web` all exit 0. Zero test deletions.

## Verification (latest at the bounded-round head `bc3231879`, restacked via `f090615c6`)

- `make lint` — exit 0. `make vet` — exit 0.
- `go test -p 1 ./cmd/evener-hub/... -count=1` — all 13 packages ok, exit 0.
- `vitest run appwire-client` (canonical frontend include of the package suite) — 86 files / 1755 tests, exit 0.
- Red-first named: `TestProjectDeleteRetryCanceledAfterOwnershipReportsError`, `TestForceStopConfirmedStoppedValidatesDeletionUnderReservationBeforeCancelingResume`, `TestForceStopValidatesDeletionUnderReservationBeforeCancelingResume`, and the client cap test.
- Zero test deletions vs `509e37bba` (lane-only numstat).

## Related PRs

- #1390: lifecycle logging (merged)
- #1391: isolated restore/Close optimization (merged)
- #1393: stopped-session recovery UX, reduced the same way

## Known limits

- This PR does not close the mid-launch crash window in general. It removes the code that would have introduced it here; establishing "this process is gone" from persisted process identity is the separate design change.
- No live deployment or Resume was performed.


obra:
Re: `RegisterResume` releases alias ownership before launch; nil-Roster skips duplicate-owner detection (resumelocks.go:390-432, app_threadlifecycle.go:613-626).

The duplicate-`Spawner.Resume` scenario requires `cfg.Roster == nil` at the launch re-check, and that cannot occur in any serving configuration. The hub builds its Roster unconditionally (`main.go`: `roster := hubcore.NewRoster(runDir, prober)`) and the single production `newWebServer` construction sets `Roster: roster`; no flag or path nils it. The guard at `app_threadlifecycle.go:613` is reachable only from tests that construct a `hubcore.WebConfig` without discovery.

With the Roster present the queueing the finding describes is safe: a second resume may register after the first released its registration tokens, but it then waits on the per-alias locks behind the first resume's launch; the first resume refreshes the roster and confirms its spawned endpoint before releasing those locks (app_threadlifecycle.go:634-657), so the second resume's under-lock re-check finds the live, current-protocol daemon and returns it instead of spawning (the `liveDaemonForThread` reuse at :613-618). A nil-Roster configuration would additionally fail earlier: discovery is skipped, `lookupDaemonOwner` has no roster to consult, and the spawn proceeds without any ownership evidence; but again, no production wiring constructs that configuration.

A clarifying comment now records the invariant at the re-check site (commit 5b5d81c00). If a configuration that serves with `Roster == nil` ever appears, this rebuttal should be revisited.

PR #1393, opened 2026-09-15
fix(web): keep stopped sessions usable through recovery

### LINKED ISSUE #1567: Recovery UX: Retry reports a generic failure when the resumed ref differs from the record's target ref
Deferred from #1393, where RoboRev raised it and we deliberately did not implement it before merging.

## Where
`cmd/evener-hub/frontend/src/stores/threads.ts` — `retryBlockedMutation`.

## What RoboRev reported
`retryBlockedMutation` discards the ref returned by `resumeSessionForUserIntent` and then checks `mutationAuthorityRefs`, status and `handleReady` against the original `record.targetRef`. When the resume returns a replacement ref, Retry fails with a generic error after having already resumed.

## Qualification, from an independent source review
This is **not always a failure**: the original ref can still carry authority and continue. It is **not an unsafe retarget** either. What is wrong is the *reporting*: the user is told nothing specific about a replacement, and the old payload is neither settled nor clearly left uncertain.

## Conservative option
Capture the differing ref and surface a specific "resumed elsewhere; original delivery remains uncertain" outcome, without navigating, without retargeting the payload, and without settling the old row. Replacing the boolean with a typed result would widen caller impact; a distinct, accurate message is smaller.

## Do not
Do not auto-reject recovery merely because the ref changed, do not retarget the old payload at the new ref, and do not claim the resume failed.

Disclosed in #1393's body when it merged.

### LINKED ISSUE #1568: Recovery UX: a stale blocked-note save silently no-ops instead of saving the dirty note
Deferred from #1393, where RoboRev raised it and we deliberately did not implement it before merging.

## Where
`cmd/evener-hub/frontend/src/stores/humanNoteDrafts.ts` — the blocked-note branch.

## What RoboRev reported
The blocked-note save unconditionally returns after `retryBlockedMutation(..., "backgroundNote")`. When that retry returns `false` because another tab settled or reopened the row, the current dirty note never falls through to the normal `setHumanNote` save path, so the user's edit silently goes nowhere.

## Qualification, from an independent source review
The literal "newer edits are rejected" reading is **false**: a genuinely newer edit increments the generation and takes the normal save path. The narrower real case is a same-generation blocked row whose metadata says the row is blocked while another connection settled or removed it without a local canonical acknowledgement. So this is a stale-absence case, not a general data-loss path.

## Conservative option
Give the retained draft a distinguishable status instead of silently appearing to have saved. Do not clear dirty state, and do not treat absence as canonical acknowledgement. If a re-read is needed, use it only to distinguish absent, blocked and already-submitting ownership while keeping the existing Stop, write, focus and generation checks.

## Do not
Do not blind-freshen the queue, do not resume a reopened or already-submitting row with a second identity, and do not resave on a blanket `false`.

Disclosed in #1393's body when it merged.

obra:
## Scope

A stopped session must stay usable, and recovering it must be explicit.

- The composer and its draft stay available for a stopped session, with the recovery notice and the Resume action, instead of the session collapsing into an unusable state. A recovery-fenced session whose snapshot still advertises `send: true` renders a **disabled** Send, not an actionable button that can only produce a refusal toast.
- Explicit Resume resumes the session, reports progress, and reports errors. It keeps its Stop fence: a Stop acknowledged while Resume is reconnecting sends no resume RPC, and a stale post-Resume read does not publish over a newer Stop. Stop generations are now baselined before the resume starts, so a Stop recorded against the resumed identity during the reconnect window cancels the resume rather than becoming the fence’s own baseline.
- Explicit Retry of a blocked mutation works and reports accurate outcomes, including the settled-elsewhere case: a retry whose target is viable is unaffected, and a record whose session is live elsewhere is reported as such rather than as a generic failure or a silent success (closes #1567).
- A background shared-note save that another connection settled now reports a distinguishable retained-draft status instead of appearing to have saved (closes #1568).
- The AppWire `FakeClient.resumeThread` runs its `beforeRequest` hook in the same order the real client does, so a test can no longer stage a Stop that the guard had already passed. The real client's `beforeRequest` throw path is now covered directly: a throwing fence sends no resume RPC, releases the pending resume, and leaves the reconnected primary healthy.

## What this PR deliberately does not do

An earlier revision of this branch made **Send implicitly resume a stopped session**. That feature has been removed from this PR and will land separately. (The review history: RoboRev's finding count on the full-scope branch went 4, then 4, then 3, then 8, and several of the newest findings were about that revision's own fixes.)

A later revision added **Stop cancellation of pending outbox sends**: an in-memory `canceledMutationIds` set, recorded after the Stop RPC, that reconciliation consulted before reopening a `blockedUnknown` record. **That mechanism has been moved out of this PR entirely.** RoboRev's review of it (both High findings and Mediums 4/5/6 on the previous head) converged on one root cause: the cancellation fact lived in per-tab memory while the IndexedDB outbox it governs is durable and shared across tabs, so every race and every reload/cross-tab gap was an artifact of that mismatch. The fix is not a better in-memory mechanism; it is a durable `canceled` state on the outbox record itself, written at user intent before the stop action. That design is written up separately and lands as a follow-up PR with its own review.

Two consequences, stated plainly:

- **Stop does NOT cancel pending outbox sends in this PR — the same behavior as `main` today.** A send that was in flight or blocked when the user stopped a session may still be delivered or retried by the outbox's normal lifecycle, exactly as before this branch. The follow-up PR is where Stop takes pending sends back.
- The mechanism's requester-protected invariant test — **"a Stop canceled retry is not resent by a later discovery scan"** — is removed with the mechanism, **sanctioned by the descope, and returns in the follow-up PR** wired to the durable design (its test plan keeps that invariant as test #1). Four sibling tests of the same machinery went with it: "a Stop during the retry's reconciliation is not overtaken by its dispatch", "a background-note retry does not release a sibling mutation canceled by the same Stop", "an explicit user retry after a Stop dispatches the canceled mutation", and "a Stop acknowledged after a hydration reopened a record puts it back to blockedUnknown".

What stays is the shell those findings also touched, now covering its own seams: the resume fence (`userIntentStopGenerations`, `resumeStopFence`, the client `beforeRequest` hook, the post-await re-check, the Session.tsx identity fence), the fenced-session notice with disabled Send and explicit Resume, the retry settled-elsewhere reporting, and the retained-draft note status. Stop-generation entries are now pruned when a ref's last holder releases it, so the fence map stays bounded by live refs.

## Final round on the d7d4c62 findings (2 Medium; Lows filed)

- **Stop-generation counter recycled by `releaseThread` (was Medium):** Stop generations now come from one page-wide monotonic sequence (`++userIntentStopSequence` per ref), so releasing a ref and re-ensuring it can never recycle a value a fence captured; pruning stays safe because values are forever-unique. Red-first: the reviewer's release/re-Stop race, plus the pruning test now covers delete-then-reincrement.
- **`retryBlockedMutation` returned `true` for an absent record (was Medium):** the background-note path no longer treats the retry's boolean as the settlement authority. After any retry it reads the ref's own records: absent means settled-elsewhere status (the #1568 status), still blocked means the draft's status was already true, and present in another state updates the draft from that actual state. The public boolean signature is unchanged.
- The three Lows are filed, not fixed here: **#1721** (pinned-ref generation leak), **#1722** (redundant projection refresh), **#1723** (hardcoded read-index test).

## Round on the fee4eb8 findings (4 Medium; Lows filed)

- **Note lookup ignored optimistic records (was Medium):** the persistence read now includes the optimistic store; an accepted-but-unreflected note reads pending until canonical state arrives, never settled-elsewhere (`97ddbf7c4`).
- **Recovery fence only covered `notLoaded` (was Medium):** an obligated local target is now fenced in EVERY non-active status, including a live `idle` snapshot with `resumeRequired: true`; the active carve-out is justified in the comment (`544198b51`). Red-first: live fenced idle renders a disabled Send, no turn/start enqueued.
- **Pre-RPC window fenced only the old ref (was Medium):** `resumeStopBaseline` gained a no-ref form that fires when ANY generation increased; `beforeRequest` uses it, so any acknowledged Stop suppresses the resume RPC (`9a27c4f8a`).
- **Release could delete the only Stop evidence (was Medium):** Stop generations now persist for the page session; pruning is reverted, and the replacement test pins the reviewer's exact interleaving (baseline absent-as-0, Stop, release, fence still throws) plus the global baseline's contract (`1585412ce`). This changes #1721's premise; noted on the issue.
- **Our own test-staging Low:** the existing test is renamed for the RPC-in-flight window it actually stages; the true reconnect window is covered by the new global-baseline tests (`a9c3494ca`, `3fa51902b`).
- Lows filed: **#1727** (`followUpEngaged` scoping). `settleAfterRead(9)` was already **#1723**.

## Simplify-code pass (quality only; requested after the Mediums rounds)

Four read-only angle reviews over the full PR diff, deduped to nine findings, applied under a no-behavior-change rule (helpers added, call sites delegate; zero test-file edits; revert-not-fix-forward on any gate failure).

- **Applied (7):** `resumeStopFence` delegates to `resumeStopBaseline`'s per-ref compare (doc comment corrected; the export stays for tests); the note draft's record-to-status mapping is shared with `refreshPersistence`; `beforeRequest` takes the baseline directly; one `retryErrors` lookup instead of has+get; `handleRetry` sets the retry error instead of throw-then-catch; the composer's local recovery fence derives from one predicate across components; the queue strip shares `retryBlockedMutation`'s snapshot refusal predicates.
- **Skipped (2):** reading the queue-strip decision from the projection instead of the second direct `readMutationPersistence` (that read's freshness and index are the pinned settle-window behavior, tracked by #1723); and splitting `resumeStopBaseline`'s dual form (the closure reads clearer as-is).

Gates at `298d8ac68` (restacked via `1224eecb8`): `npx biome check --write` exit 0, targeted suites 710/710, `make test-web` exit 0 (typecheck, unit, lint). Zero test-file edits in the pass.

## Round on the 298d8ac finding (1 Medium)

- **Ended-session Send used the narrow fence (was Medium, two reviewers):** the rendered Send's disabled expression was gated on `recoveryFencedLocal` (notLoaded-only) while `availabilityFor` fences every non-active status, so a `closed` session with a restart-blocking obligation and `send: true` (reachable: forceStop sets the obligation, the hub stamps closed frames with Send) rendered an enabled Send that could only toast. The rendered Send now gates on `isLocalRecoveryFenced`, the shared broad predicate `availabilityFor` derives from (`bbde8e13b`). Red-first: "a fenced closed local session that advertises send renders a disabled Send" (failed pre-fix at the disabled assertion).

## Verification

- At `bbde8e13b`: `make test-web` exit 0 (typecheck, unit tests, biome) on the merged tree; targeted suites 691/691.
- At `3fa51902b`: `make test-web` exit 0 (typecheck, unit tests, biome) on the merged tree; targeted suites 648/648 (+54 integration) before the full gate.
- At `fee4eb874`: `npx biome check --write` exit 0 on touched files; targeted suites green (threads 316/316, humanNoteDrafts 22/22, Session 91/91); `make test-web` and CI run on the pushed head.
- At `d7d4c627a`: `make test-web` exit 0 (typecheck, unit tests, biome) on the merged tree; new red-first test "a Stop on the resumed identity during the resume reconnect cancels the resume" (fails on the pre-fix tree).
- Restacked on `origin/main` via merge `89a19ba7f` (no conflicts; the composer contenteditable landing from #1410 required no test changes this round); the descope is commit `c9dd0fd5d` on top. `make test-web` exit 0 (typecheck, unit tests, biome) on the merged tree.
- `npx biome check --write` on every touched file under `cmd/evener-hub/frontend/src` and `appwire-client/typescript` exit 0.
- Mechanism removal verified by inspection and by the acceptance numstat: deletions confined to the cancellation mechanism and its tests; no `canceledMutationIds`, `recordCanceledMutations`, `stoppedSinceBaseline`, or `checkStopped` anywhere in the tree; `restoreProvenAbsent` is back to its two-argument signature in storage, dispatcher, and caller.
- New behavior tests, green: "a fenced stopped local session that advertises send renders a disabled Send" (Composer, fails on the pre-fix tree), "a throwing beforeRequest sends no resume RPC, releases the resume, and keeps the primary ready" (real AppWire client over FakeSocket), "releasing a ref prunes its Stop generation, and a fence captured before the release still fires" (fails on the pre-fix tree).
- The background-note dispatch guard test's previously vacuous wait now proves itself: it asserts the unrelated pending send is not dispatched by the save, then advances the discovery-scan boundary to show the same detector catching the scan's own (legitimate) resend.
- Browser verification is unchanged from before: all six browser guards still time out at their shared navigation boundary before assertions, with no timeout or assertion weakened.

## Related PRs

- #1390: lifecycle logging (merged)
- #1391: isolated restore/Close optimization (merged)
- #1392: completion-aware Resume and cancellation safety; needed for giant-session recovery to complete beyond the ordinary 30-second budget
- Follow-up (planned): durable Stop cancellation of pending outbox sends — `canceled` as an outbox record state, written at user intent before the stop action; carries the restored invariant test.

No live deployment or Resume was performed.

## Known limits

- Sending to a stopped session is refused with the recovery notice rather than resuming. That is a real behavior change against the original promise of this PR, and it is the price of removing the race class rather than fixing it one finding at a time.
- Stop does not take back pending sends (same as `main`); the durable cancellation lands in the follow-up PR.
- The dispatcher admission removal means a fresh Send admitted past an older blocked Send can still stall behind it until the machinery change lands.


obra:
Mapping the `c9dd0fd` verdict's four Mediums:

- **`threads.ts:1649` and `threads.ts:2751` (reconciliation-window Stop races):** these are the windows the descope moved out, documented in the body under "What this PR deliberately does not do". Stop does not cancel pending outbox sends in this PR (same as `main`); the durable `canceled` record state that closes these windows by construction lands in the follow-up PR with its own design. Noted that reviewer 3 traced the fence/generation flow and found no concrete reachable failure in the diff as it stands.
- **`threads.ts:969` (background note retry reopens unrelated records):** pre-existing on `main` (`retryBlockedMutation` and the target-wide `restoreProvenAbsent` reopen both exist there); this PR did not introduce it. Filed as #1716 with a conservative option.
- **`Session.tsx:139` (resume fence baselines the new identity only after resume returns):** fixed in `d7d4c627a`. Stop generations are snapshotted before the resume starts (`resumeStopBaseline`), and both the old and resumed refs are fenced against their pre-resume baselines, so a Stop naming the resumed ref during the reconnect window cancels the resume instead of becoming the fence's own baseline. Red-first test: "a Stop on the resumed identity during the resume reconnect cancels the resume". `make test-web` exit 0 on the merged tree.

Issue #1284, opened 2026-09-14
hub: destructive confirmations cannot tell a same-name recreation from the entry the user reviewed


obra:
Deferred from roborev round 45 on #1136 (finding verified against source; the fix is a design change, not a bolt-on).

## The gap

`verifyEndpointFingerprint` (cmd/evener-hub/app_auth.go) refuses a destructive mutation whose client asserted an endpoint the name no longer resolves to. The assertion is the endpoint fingerprint: an HMAC over the resolved destination identity (base URL, protocol, endpoint templates). It identifies *where a name points*, not *which entry it is*.

If an instance is removed and recreated under the same name with identical endpoint settings, the replacement resolves to the same identity and therefore the same fingerprint. A confirmation captured against the removed entry (e.g. the removal dialog's captured row, which the client asserts at submit) then matches the replacement, and the mutation acts on an instance the user never reviewed: the removal deletes the replacement's authored configuration and credentials.

Reachability: another actor (a second tab, the CLI, a provisioning script) must remove and recreate the same name, with the same endpoint settings, inside the confirmation window. Narrow, but real; the settings-identical recreation is exactly what a template-driven or retry-driven recreation produces.

## Remedy direction (reviewer's, agreed)

Give an instance a persisted identity that survives as long as the authored entry does, and carry it through confirmations: a per-instance unique id or monotonically increasing generation, surfaced in the listing (InstanceEntry) and echoed in the destructive params (`expectedInstanceId`), validated under the same credential lock as the mutation. That distinguishes "the same destination" from "the same entry", which the digest cannot.

Open design questions: where the identity is persisted (a new providers.toml key has backward-compatibility implications for files that lack it; a hub-owned sidecar under the state root avoids touching the user layer but must stay consistent with hand edits), and which mutations must carry it (remove, edit-rename, credential clear/logout, and anything else acting on a listed row).

## Evidence

- `verifyEndpointFingerprint` doc + body: cmd/evener-hub/app_auth.go (scope: "an endpoint this name no longer resolves to").
- `destinationFingerprint` / `destinationIdentity`: cmd/evener-hub/app_instances.go (digest inputs contain no per-entry identity).
- Round-45 verdict: https://github.com/prime-radiant-inc/evener/pull/1136 (roborev comment `97891d0`).

PR #1397, opened 2026-09-15
feat(hub): assert the endpoint fingerprint on instance edits

obra:
Slim version of #1365, rebuilt on current main (which has since taken SDK A4, #1272) after that PR's review loop showed the client-side conflict recovery was the part attracting findings.

## What

The instance sheet's stale-draft guard is client-side only: it compares the row it displayed against the store's listing, and a concurrent change can outdate that listing between the comparison and the `evener/instance/edit` RPC. This adds the server-side half, mirroring the credential-write params.

- `appwire.InstanceEditParams` gains an optional `ExpectedEndpointFingerprint` (`omitempty`).
- The hub's `Edit` resolves the fingerprint key once, off both locks (the round-59 threading already used by `Remove` and the credential writes), then validates the assertion under `mu` + `credMu` before anything is written, refusing a mismatch with `appwire.Conflict`.
- Edit's refusal says "review its destination and save again" instead of reusing the credential writes' "enter the credential again", which names an action the edit form does not offer.
- The web `InstanceSheet` sends the fingerprint it displayed (stamped by `credentialsStore.edit`, matching that store's `remove` sibling) when the hub served one, and omits the field when the hub could not key it, so an outage does not block a save. The native `editProviderParams` does the same.
- Regenerated `appwire-client/typescript/types.gen.ts` and `docs/appwire-protocol.md` with `make generate`.

An absent or empty assertion keeps the pre-fingerprint behaviour for the TUI and older clients. This is the rule the maintainer chose for the credential writes too: an empty assertion is accepted while the hub can key a fingerprint.

## Deliberately not here

Two things explored in #1365 and split out after review, as follow-ups alongside #1284:

- the client-side re-anchor-on-conflict recovery (the sheet reseeding, the native draft rebase) - roborev's last two verdicts gave opposing guidance on its semantics, and it is recovery UX rather than the precondition itself;
- the `endpointConflict` discriminator, which exists to let a client tell "the name moved" from "a rename collided", and is only needed once a client acts on the distinction.

## Tests

Red-first, mutation-verified:

- Hub (`cmd/evener-hub/app_instances_edit_endpoint_test.go`): a stale assertion after a re-pointed `base_url` is refused with an `appwire.Conflict` and nothing reaches `providers.toml`; the current assertion is accepted; an absent assertion still lands; the refusal carries the edit's own remedy; the key is resolved before both locks (TryLock probe on the seam, mirroring `TestInstances_KeyResolutionHappensBeforeBothLocks`).
  - Mutations: forcing the verification condition false reddens the guard; dropping the off-lock resolution reddens the positive control and the lock-order probe; restoring the credential wording reddens the wording assertion.
- Frontend (`InstanceSheet.test.tsx`): a save asserts the displayed fingerprint on the wire; a row the hub cannot key sends no assertion. Three pre-existing curated-shadow rename assertions moved to the new wire shape.
- Native (`providerForm.test.ts`): the same two cases for `editProviderParams`.

## Gates

`go build ./...`, `go vet`, hub instances/auth, the credentials suites (456/456), `biome ci` (1249 files), `tsc`, native (759/759 + check) and `make test-api-package` all pass locally; CI is running.


obra:
Closing without merging, on the maintainer's call: the race this closes is a weird edge case we do not need to cover.

What it would have covered: a name re-pointed between the instance sheet's own stale-draft check and the evener/instance/edit RPC - requiring another actor (a second tab, the CLI, a provisioning script) to remove and recreate, or re-point, the same name inside the sub-second window between the client's check and the RPC. The client-side identity check already catches the case a client can see before it saves.

Recorded here so the decision is durable:
- The residual is accepted, not fixed. It is the same family as #1284: the endpoint fingerprint identifies where a name points, not which entry it is.
- The review history is on #1365 (closed, the larger version) and here; the design - an optional ExpectedEndpointFingerprint on InstanceEditParams, validated under the instance and credential locks with the key resolved off-lock first - is written up in this PR's body if it is ever wanted.
- None of the review findings need answering: the protocol-version High, the native-editor Medium and the web-sheet Low are all properties of the change, and they disappear with it. Main is unchanged by this.

PR #1399, opened 2026-09-15
agent,appprojector: the independent fixes from the replay work (piece A3 of the #1100 split)

obra:
Piece A3 of the #1100 split: the fixes #1281 (closed, `claude/replay-durable-records`) found that do not depend on the fold design, cherry-picked by content from 5590daa5f, d791c39f5, 577cbf7ee and ba30cf0a5. Two commits, each RED-first.

## 1. appprojector: an environment block hands back a stable reservation as stable

`EventEnvironment` saved and restored `reservedTurnID` around its own turn but not `reservedTurnIDIsStable`, which `openTurn` clears. A stable client-mutation id reserved before the environment block came back looking counter-minted, so the following input spent no number and every later live `turn_N` landed one short of the reload's numbering. The save/openTurn/restore dance is one `bookkeepingTurn(stableID, at, announce)` helper so the compaction-layer announcement (A2) does not copy it.

Test: `TestAppEventProjectorAnnouncementsKeepAPendingStableReservationStable` (live `turn_2` vs reload `turn_3` on main). Main has no `compactionAnnouncement`, so only the environment site is covered here.

## 2. agent: one reminder per handoff, and a failed consumption save keeps its receipts

On main `consumeSkillReloadReminders` removed the receipts and saved; a failed save left them removed, so the next autosave persisted a meta.json that had forgotten a reminder no restart could deliver. Now removal, save and rollback are one critical section under `metaSaveMu` (via `autoSaveMetaLocked`), and the rollback restores exactly the receipts this call removed (`removeSkillCompactionHandoffsLocked` returns them; `restoreSkillCompactionHandoffsLocked` skips a publication already re-recorded), so a handoff a fold recorded while the save ran survives.

I did not add `admitCompactedSkillReloads`'s revision guard here: that guard protects a wholesale snapshot restore by skipping the rollback when anything else mutated, which in the concurrent-handoff case would drop the consumed receipt, the exact loss the test pins. The targeted restore makes the guard unnecessary.

With the receipt kept, a retry re-derived the reminder and appended a second turn. `prepareCompactedSkillReloads` now recognizes the publication's reminder already in the history (`skillReloadReminderRecorded`, the sibling of `skillReloadOutcomeRecorded`) and retries only the consumption.

Tests: `TestSkillReloadReminder_FailedConsumptionSaveKeepsTheHandoff`, `TestSkillReloadReminder_FailedConsumptionSaveKeepsAConcurrentHandoff`, `TestSkillReloadReminder_RetryAfterAFailedConsumptionSaveAppendsNoSecondTurn`, all RED on main.

## 3. `schema.TurnKind.IsPresentationalRecord()`: skipped

Main has no `AttentionResolution || ContextCompaction` exclusion lists: `TurnContextCompaction` is not referenced outside `agent/schema` on main, and the copied lists the branch unified were `AttentionResolution || RoundTimings || ContextCompaction`, kinds the fold design (A2) and piece B introduce. The predicate lands with them.

## Gates (all foreground, all green)

gofmt (toolchain binary); `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz` for root and agent; `go test -race -count=3` for internal/appprojector, agent/schema, the 41 agent subpackages, and the agent root via `agent-shards -race -count=3 -timeout 150m` (4 shards, 6012 tests, PASS); `cmd/evener-tui/...` with -race; `make lint`.


obra:
## /simplify pass (3e1fb960d)

Four reviewers read the diff. Deduplicated, being applied as round 1; behaviour unchanged, coverage kept:
- **Projector**: a bookkeeping turn stops clobbering the reservation and restoring it field by field (the bug pattern this PR fixes); the reservation is one value, or the bookkeeping turn leaves it untouched.
- **Skill reload**: one guarded remove-save-rollback transaction under the save lock serves both consuming reminders and admitting reloads, with a targeted restore, so the admit path loses the autosave race the PR fixed for reminders and the revision guard is deleted in one place instead of reasoned away; reminder consumption becomes idempotent by construction (the receipt is consumed inside the reminder's own durable write), which removes the history scan; the restore reuses the one-entry-per-publication rule.
- **Comments** collapse to one rationale each; a doubled length check and a guard for a case the caller never produces go.
- **Tests** reuse the package's drain helper, receipt fixture, rename-failing filesystem and skill-state reader; the two heavier tests move to the in-memory filesystem and plain session the sibling already uses.

obra:
## Review round 1 disposition (3e1fb960d)

CI green. Two Medium and three Low, all accepted; the round already in flight covers three of them by construction:
- **The reminder dedup consults only compactable in-memory history, so a fold between a failed save and the retry brings the duplicate back** — accepted; consumption becomes idempotent by construction inside the reminder's own durable write, so there is no dedup scan to fool.
- **The admission path removes and saves outside the save lock and restores wholesale under a revision guard** — accepted; both paths share one guarded transaction with a targeted restore.
- **A warning is emitted while the save lock is held** — accepted; unlock, then emit.
- **An empty publication id can match identity-less receipts** — accepted; empty ids are skipped.
- **The bookkeeping turn stashes the reserved id but leaves its stable flag set** — accepted; the bookkeeping turn no longer touches the reservation at all.

obra:
## Round 1 (simplify)

Merged origin/main (A4 at e2baa5f72). Six commits on top of round 0, quality only, every test still green:

1. **Projector, 649f16788**: `openTurn` takes a `bookkeeping bool`; a bookkeeping turn spends its entry-index number and opens on its own id without touching the reservation or `anyTurnStarted`, so `bookkeepingTurn` snapshots and restores nothing. Shared turn-scoped resets moved to `beginTurnState`. The RED test stays and now also asserts both reservation fields are untouched after the environment block (review item). Test uses `notificationItemTurnID`.
2. **Tests, e7c352673**: `drainSessionEvents`, `plantReloadReceipt(s, id, selection)` (the existing helper extended; `plantReminderReceipt` is a call), `notesRenameFailureFS` with a `before` hook + `errParkedNotesSave`, `countSkillReloadReminderTurns`, MemMapFs everywhere, retry test on `newSession` + `plantPreloadRecord`.
3. **Admit, 27e2eef02**: RED `TestSkillReload_FailedAdmissionSaveKeepsAConcurrentHandoff` (revision guard skipped the rollback and lost the consumed receipt). Mutation, save and rollback run under `metaSaveMu` (`autoSaveMetaLocked`); rollback restores the removed receipts and drops this admission's obligations, nothing else. Wholesale snapshot and revision guard deleted. Warning emits after the lock is released (review item).
4. **Idempotent by construction, 09d872283**: the reminder's receipt leaves `PendingHandoffs` inside the durable pair's live commit (`consumeSkillReloadReminderLocked`, under `s.mu` with the history append). `skillReloadReminderRecorded` deleted; the fit-failure save call deleted; `consumeSkillReloadReminders` collapses to `persistSkillReloadReminderConsumption` (save, warn, no rollback). The retry test now compacts the reminder out of the live history between the failed save and the retry and proves the retry still appends nothing (review item).
5. **Lookup, 5b21b2db2**: `pendingHandoffIndexLocked` serves record, publish-delivery and restore; newer-receipt-wins stated once next to the coalescing rule; `id != ""` guard lives in the helper.
6. **Review L, beea9e7de**: `removeSkillCompactionHandoffsLocked` never matches an empty id, with `TestSkillReloadReminder_ConsumingAnEmptyPublicationRetiresNoCancellation` (RED without the guard, verified).

Two deviations, both for the smaller final state:
- **No `consumeHandoffsDurably` helper.** Item 4 leaves the admit path as the only transaction with a rollback (the reminder path no longer rolls anything back), so a helper would have exactly one caller; the transaction is inline in `admitCompactedSkillReloads`.
- **The two reminder-path rollback tests are gone** (`FailedConsumptionSaveKeepsTheHandoff`, `FailedConsumptionSaveKeepsAConcurrentHandoff`): with the receipt consumed in the reminder's own commit there is no rollback for them to pin. The retry test pins the replacement contract (failed save leaves exactly one reminder and no receipt; retry appends none even after a fold); the concurrent-handoff protection is pinned on the admit path. Honest note: on this contract the reminder path ends up where original main already was (no rollback, restore reconciles) plus the atomic consumption, so the retry test is RED against round 0's rollback-without-idempotency state, not against main.

Gates (foreground; the shard run was pushed to the background by the harness's 10-minute cap and read from its log): gofmt; vet ×3 root + agent; `-race -count=3` appprojector, schema, 41 agent subpackages, agent root via agent-shards (4 shards, 6012 tests, PASS); TUI suite -race; `make lint`.

obra:
## Review round 2 disposition (beea9e7de)

CI green; two of three reviewers clean. One Medium, accepted:
- **A bookkeeping environment turn advances the live turn counter while persisted numbering follows the entry index, so after an environment entry plus a pending stable reservation the next ordinary turn is numbered differently live and after reload** — accepted; a bookkeeping turn does not consume the counter, with a live-versus-reload numbering test for that sequence.

obra:
## Round 2 fixes

Merged origin/main. On the Medium (`appwire_projection.go` bookkeeping turn increments `nextTurn`): I measured instead of changing it, and the finding does not reproduce.

The cold side (`internal/apptranscript.persistedTurnID`) numbers every persisted entry by its entry index unless it carries a stable id, so an environment entry spends a number on reload whether or not it has its own id. The live bookkeeping increment is what keeps the two in lockstep. Measured over five orderings (environment with/without its own id crossing a pending stable reservation; the reservation released unused after it; environment after real work; environment with nothing pending), live and reload agree on every one, including the reviewer's `[environment entry, stable reservation pending, ordinary turn]` → `[turn_env_stable turn_stable_input turn_3]` on both sides. Removing the increment makes four of the five disagree, live one short of reload (`turn_2` vs `turn_3`), which is exactly the round 0 bug this PR fixes.

Landed as b35d40233: the parity harness in `environment_reservation_test.go` becomes that five-case table, each case also asserting the environment block leaves the reservation untouched. It is the regression guard against the proposed change.

Gates: gofmt; vet ×3 root + agent; `go test -race -count=3 ./internal/appprojector/...`; TUI suite -race; `make lint`. All green.

obra:
Correction to the round 2 disposition: the Medium does not reproduce. Reload numbers an environment entry by its entry index, so the live counter's increment for the bookkeeping turn is exactly what keeps live and reload in agreement; the implementer measured five orderings, all agree, and removing the increment breaks four of them (round 0's bug). The five-case parity table lands as the regression guard. I accepted the finding without measuring; the implementer measured.

obra:
## Round 3 fixes

Merged origin/main (add4e5b93). The Medium (`restoreSkillCompactionHandoffsLocked` appends restored receipts after handoffs published during the save) is **real**; measured, not read:

**Lock window.** In `admitCompactedSkillReloads` (`agent/session_skill_reload.go`) the transaction holds `metaSaveMu` throughout but releases `s.mu` before `autoSaveMetaLocked()` and re-takes it for the rollback. `Session.Compact` is server-exposed while a round is active (its own comment: it races the round loop's ManageContext), and a real fold with no captured operation still records an absent-selection reminder handoff with a fresh publication id via `claimCompactionLocked` → `recordSkillCompactionHandoffLocked` (append). So another goroutine can append to `PendingHandoffs` in the window. Round 1's `TestSkillReload_FailedAdmissionSaveKeepsAConcurrentHandoff` already exercises that window; it only asserted membership, not order.

**Contract.** `prepareCompactedSkillReloads` processes `PendingHandoffs` in slice order: `batch.Items`/`outcomes` follow it, and the admission records carriers in `batch` order. `TestSkillReloadReminder_FitFailureConsumesEarlierReminders` (`pub-fit-1` consumed, `pub-fit-2` left) already pins slice order = processing order. Restart rebuilds the slice in transcript (publication) order, so the append-restore also made live order diverge from post-restart order.

**RED** (72afee3fb): `TestSkillReload_FailedAdmissionSaveRestoresTheReceiptAheadOfLaterHandoffs` — older valid receipt (skill `older`) planted, admission save fails while a newer valid receipt (skill `newer`) is recorded mid-save, then the retry runs for real. Before the fix: pending `[pub-order-newer pub-order-older]`, and (mutation check with the first assertion neutered) the retry prepared `[pub-order-newer:newer pub-order-older:older]`, i.e. the reorder reaches delivery.

**Fix** (930fd0c5a): `removeSkillCompactionHandoffsLocked` returns `[]removedHandoff{at, receipt}` in ascending position; `restoreSkillCompactionHandoffsLocked` re-inserts each at `min(at, len)` (newer-receipt-wins skip kept). Positions stay valid because every removal from `PendingHandoffs` runs on the round loop (the caller): the reminder consumption, this admission and cancellation retirement; other goroutines only append (fold claim) or update in place (coalescing, delivery flip). The other caller (reminder consumption) only checks `len(...) > 0`, unchanged.

Gates: gofmt on touched files; in agent/: `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`; `go test -race -count=1 ./agent/...` — 41 subpackages ok, root `agent` package ok at 837s with `-timeout 90m` (the default 10m per-package timeout tripped on the first run with two other sessions' race suites on the box). TUI suite not run: no projector shape change.

obra:
## Review round 3 disposition (b35d40233)

CI 15/15 green. RoboRev: one Medium (two reviewers clean).

- **Failed admission save restores older handoffs behind concurrently published newer ones** — accepted after measurement. The admission releases `s.mu` around the save while `metaSaveMu` is held, and `Session.Compact` can record a fresh reminder handoff in that window; `prepareCompactedSkillReloads` processes `PendingHandoffs` in slice order and the restart path rebuilds the slice in publication order, so the append-restore diverged live order from post-restart order. RED test reproduced the reorder reaching delivery. Fix: removal returns positions, restore re-inserts at the original position (newer-receipt-wins skip kept); revision order was rejected because coalescing places a higher revision at an older position.

Fixed in round 3 at 930fd0c5a.

PR #1400, opened 2026-09-16
sdk: name the gate's test exemption and widen the package's app-import check (A4 follow-up)

obra:
Two Lows deferred from #1272 (Jesse's rule: merge Lows, follow up).

1. `make lint-package-imports`'s annotation (make/linting.mk) and AGENTS.md said the gate exempts only the resolver configs, but it also exempts `mobile-native/src/metroResolver.test.ts` (which asserts the name-to-path mapping). Both now name that test alongside the configs, and `docs/developing-evener/linting.md` is regenerated to match.

2. `package-test-files.mjs` `describeAppImports` treated only the three consumer src trees as "the app", so a package file reaching into `cmd/evener-hub/frontend/scripts` was not caught. It now treats the whole web app (scripts and configs, not just src) plus the two mobile trees as off-limits; shared repo tooling under `scripts/sdk` stays allowed. RED: a package file importing `../../cmd/evener-hub/frontend/scripts/x.mjs`.

Gates: `make lint` (incl. `lint-generated`), the Makefile audits, `WEB=0 make test`, `node --test` for the frontend script suites (103).


obra:
## Review round 1 disposition (f9f1da1c9)

CI green; one Low (a stale doc comment above `describeAppImports`). Per Jesse's rule this merges now and the comment follows in its own tiny PR.

PR #1401, opened 2026-09-16
refactor(sdk): move the rail's session state helpers into the AppWire package (C14)

obra:
SDK migration plan row C14 (docs/superpowers/plans/2026-09-12-sdk-migration.md).

`cmd/evener-hub/frontend/src/shell/rail/sessionState.ts` (41 lines, no imports) moves into the AppWire TypeScript package as `appwire-client/typescript/railSessionState.ts`. Renamed because a bare `sessionState` beside `sessionErrors` in the package root is ambiguous; the module is the rail row's gloss.

- `humanizeState` had no test. `railSessionState.test.ts` pins every wire state's word, the `awaiting`/`askPending` split, that `askPending` is ignored for every other state, and the idle fallback for `idle`/`notLoaded`/empty/unknown - each case taken from what RailRow.tsx and screens.tsx pass today.
- Both importers rewritten to `import { humanizeState } from "@evener/appwire-client"` (web `shell/rail/RailRow.tsx`, native `mobile-native/src/screens.tsx`). No re-export stub at the old path.
- Package plumbing in the same PR: `tsconfig.build.json` files entry, `index.ts` re-export, qualification smoke calls, README prose, header comment naming both consumers.

Falsification: deleting the `index.ts` re-export makes `make test-api-package` fail with `shipped module unreachable from every published specifier: railSessionState`; restoring it passes.

Gates run locally on this head: `make test-api-package`, `make test-web`, `make test-native`, `make test-web-browser`, `make lint`, `go test -run TestScenarioSource .`, `biome ci src ../../../appwire-client/typescript` (frontend). The browser gate's skillguard timed out once while it overlapped `make lint`'s go vet on the same host, and passed cleanly when rerun alone.


obra:
## Review round 1 disposition (c69714563)

CI green; one Low (README wording: the gloss is short and lowercase, not one word). Per Jesse's rule this merges now and the wording follows in its own tiny PR.

Issue #1402, opened 2026-09-16
flaky(web): editorial-preview fixture.test.tsx times out finding 'Editorial fixture parent' under machine load


obra:
Observed 2026-09-15 on branch claude/sdk-c20-sourcelabel (worktree off e2baa5f72) during `make test-web`, while golangci-lint, `make test-native` and another worktree's vitest were running concurrently:

```
FAIL  src/dev/editorial-preview/fixture.test.tsx > real AppShell opens fixture parent, distinct child, and returns via navigation
TestingLibraryElementError: Unable to find an element with the text: Editorial fixture parent.
```

1 failed / 11259 passed. The same file passes alone (`vitest run src/dev/editorial-preview/fixture.test.tsx`, 1/1) and the full `make test-web` re-run on a quieter machine passed 484/484 files. The test mounts the real AppShell with dockview and waits ~1.6s for the fixture text; under load that find times out.

Nothing in the C20 PR touches `src/dev/`, dockview, or AppShell. Residual filed per the queue rule. Candidate fix: give the AppShell mount in this test a longer `findBy*` timeout, or gate its render on a deterministic ready signal rather than the default 1000ms wait.

PR #1403, opened 2026-09-16
refactor(sdk): move sourceLabel into the AppWire package (C20)

obra:
SDK migration plan row C20 (docs/superpowers/plans/2026-09-12-sdk-migration.md).

`cmd/evener-hub/frontend/src/panes/settings/sections/marketplacesPlugins/sourceLabel.ts` and its test move into `appwire-client/typescript/` as **`marketplaceSourceLabel.ts`** / `marketplaceSourceLabel.test.ts`. The plan names the module `sourceLabel.ts`, but the package root already exports `sourceLabel` from `watchRows.ts` (a watch's source, "this session"), so a same-named re-export at `index.ts` is a duplicate export. The module and function carry the marketplace prefix instead of aliasing one name over the other, and the header comment says so. No logic change; the `types.gen` import becomes intra-package.

No re-export stub at the old path. Importers rewritten to `import { marketplaceSourceLabel } from "@evener/appwire-client"`:

- `cmd/evener-hub/frontend/src/panes/settings/sections/marketplacesPlugins/BrowseSection.tsx`
- `cmd/evener-hub/frontend/src/panes/settings/sections/marketplacesPlugins/MarketplacesSection.tsx`
- `cmd/evener-hub/frontend/src/panes/settings/sections/marketplacesPlugins/PluginDetailSheet.tsx`
- `mobile-native/src/MarketplaceBrowser.tsx` (was reaching into the web tree by relative path)

Package: `tsconfig.build.json` `files`, `index.ts` re-export, two smoke calls in `qualify-package.mjs` (github and git-subdir kinds), README export prose. Plan doc gets a C20 status row.

Falsification: deleting the `index.ts` re-export fails `make test-api-package` with `shipped module unreachable from every published specifier: marketplaceSourceLabel`; restored.

Gates run locally on the head: `make test-api-package`, `make test-web` (484 files / 11260 tests; typecheck, test, lint PASS), `make test-native` (Metro bundle, 757 native + 778 shared tests, tsc, script-import check), `make test-web-browser` (all guards PASS), `make lint` incl. lint-package-imports, `go test -run TestScenarioSource .`, biome via the frontend's `npm run lint` over `src` and the package. Note: `npx biome` from the repo root resolves to an unrelated `biome@0.3.3` and exits 0 without checking anything; the frontend's pinned `@biomejs/biome` is the real gate.

The first `make test-web` run had one unrelated timeout in `src/dev/editorial-preview/fixture.test.tsx` under machine load; it passes alone and on the full re-run. Filed as https://github.com/prime-radiant-inc/evener/issues/1402.


obra:
## Review round 1 disposition (23a8172fa)

CI green; one Low (the github and git-subdir branches interpolate optional fields without the `?? ""` fallback the other branches use). The head is conflicting with main after C14 landed, so the refresh that is needed anyway carries the fallback fix rather than a separate follow-up.

PR #1405, opened 2026-09-16
refactor(sdk): move credentialLabels into the AppWire package (C19)

obra:
SDK migration plan row C19 (docs/superpowers/plans/2026-09-12-sdk-migration.md).

`credentialLabels.ts` and both its tests move from `cmd/evener-hub/frontend/src/panes/settings/sections/credentials/` to `appwire-client/typescript/`, following the landed C-row template (#1234, #1226): `git mv`, no re-export stub at the old path, every importer rewritten in this PR to `@evener/appwire-client` (nine web files, two native: `mobile-native/src/ProvidersScreen.tsx`, `providerInstances.ts`). The module's own imports become relative (`./errors`, `./types.gen`) so the packed build resolves them.

The wire test decodes the hub's recorded `evener/instance/list` answer, so its reader `hubWireFixtures.ts` follows it into the package's `testing/`. It used to read the fixture with `join("..", "testdata", ...)` against the frontend's CWD, which `package-test-files.mjs` refuses inside the package; it now loads `cmd/evener-hub/testdata/authwire/responses.json` through a `?raw` import, the way the reducer tests load their `.jsonl` fixtures. Vitest's jsdom suites (`InstanceSheet.wire.test.tsx`) transform that import through Vite's server, whose `fs.allow` replaces the default list, so the hub's `testdata` directory joins it in `vite.config.ts`.

Plumbing in the same PR: `tsconfig.build.json` `files`, `index.ts` re-exports (alphabetical), qualification smoke calls for every runtime export, README prose, header comment, plan-doc status row for C19, and the scenario card citation `TestScenarioSourceCitationsResolve` caught (`test/scenarios/credentials-page-displays-sources.md`).

Falsification: with both `index.ts` re-exports deleted, `make test-api-package` fails with `shipped module unreachable from every published specifier: credentialLabels`; restored.

Gates run locally: `make test-api-package`, `make test-native`, `make test-web-browser`, `make lint`, `go test -run TestScenarioSource .`, `npx biome ci src ../../../appwire-client/typescript` all green. `make test-web`: typecheck, package-test-files guard (36 package test files collected) and lint green; the first full vitest run had one 5s timeout in `src/panes/spawn/Spawn.test.tsx` at host load ~47 (five other lanes running vitest), which passes 217/217 rerun in isolation; a second full run is in progress and will be reported on the PR.


obra:
Second full `make test-web` run at head `dd49cab7c`'s tree (docs-only commit since): PASS web-typecheck, PASS web-test, PASS web-lint, exit 0. The first run's single miss was the 5s `Spawn.test.tsx` timeout at host load ~47; every gate in the PR body is now green in a single run.

obra:
## Review round 1 disposition (eac8fc0b9)

CI green; one Low (two Go comments still point at the module's old home). The head needs a refresh onto main regardless (the wave's shared files), so that refresh carries the two comment fixes rather than a separate follow-up.

PR #1410, opened 2026-09-16
feat(web): keep skill chips inline in composer

obra:
## Current head `5281596df49c7c8f960b80a6e39d7f246d0c3b56`

One high and six medium findings are repaired on this head. The reconciliation of an externally applied value no longer computes a range that splits an atom, and it rebuilds the document only for a value the composer marks as a restore — every other change is patched, and what a patch inserts stays plain text. Image and mixed paste replace the full selected range; a sentence-ending period no longer drops restored skill activation; a word character typed straight against a chip no longer leaves an activation the document parser cannot rebuild; that separator belongs to the same undo step as the edit that needed it; and applying the controlled value no longer destroys the session's undo history.

Two further findings were refuted by direct test: bare Enter and Shift+Enter steer both dispatch and leave the composer empty, so the claimed extra newline does not occur; and the composer exposes exactly one textbox, not nested editables.

The separator is inserted on whichever side collided, so the label stays whole and nothing typed is lost, and plain `/name` text is never promoted into a chip. The parser and the editor share one definition of a complete reference (`completeSkillReferenceAt`), so that boundary rule cannot drift between them.

This head is rebased onto current `main`. The rebase dropped the provider-dialog focus repair, because upstream #1384 landed a general focus-rehome mechanism (`useFocusRehome`) whose row gate strictly covers ours; that behavior is now upstream's. Two main-added tests still assumed the removed textarea and were ported to the real editor, and the skill-guard driver was aligned with main's renamed `waitForReply` API and its newer helpers.

The real-browser guard was updated to the decided behavior: an explicit gesture for typing against a chip (its oracle is the separated result, caret included), the around-the-atom and IME steps now expect the separating space, and the composed character is cleaned back to the starting draft. No expectation was weakened to reach green; each one now states the behavior the product deliberately has.

Full gate on this exact source:

```sh
GOMAXPROCS=2 make merge-approval-gate && GOMAXPROCS=2 make vet && GOMAXPROCS=2 make test-web-browser
```

All three exited zero: lint lanes and secret scan, production build, seven Go modules and frontend, native bundle, 786 native and 799 shared tests, native checks, AppWire qualification, vet and all six real-browser guards. Notices: Metro cold cache, experimental Node SQLite, npm update.

Low-severity follow-ups, filed and verified against their preserved drafts:
- [Add durable real-editor coverage for slash-menu ARIA cleanup](https://github.com/prime-radiant-inc/evener/issues/1449)
- [Remove unused composer skill-selection helpers and their isolated tests](https://github.com/prime-radiant-inc/evener/issues/1450)
- [Update skill documentation for inline composer chips](https://github.com/prime-radiant-inc/evener/issues/1451)
- [Reconcile the composer draft comment with the inline-chip invariant](https://github.com/prime-radiant-inc/evener/issues/1554)
- [One separator insertion for a shared atom boundary, not two](https://github.com/prime-radiant-inc/evener/issues/1593)
- [Treat a trailing colon like a trailing period when bounding a skill reference](https://github.com/prime-radiant-inc/evener/issues/1594)
- [Compute the post-insert caret from node sizes, not the content array length](https://github.com/prime-radiant-inc/evener/issues/1608)

Remaining requirements are repository ones: an approving review and an up-to-date branch against current `main`. No protection bypass.

## Summary

Keep selected skills as indivisible inline chips in the composer. `Run /skill-1 and then /skill-2` reaches the agent as the complete sentence, with separate canonical, deduplicated activation metadata.

- Use a scoped ProseMirror editor for atomic selection, deletion, undo/redo, clipboard and IME behavior.
- Preserve repeated references and current draft, queue, steering and failed-submit recovery behavior. Existing persistence and wire contracts stay unchanged; no old-session migration.
- Exercise the real editor in component tests and the actual browser guard, including exact provider input and transcript assertions.
- Normalize incoming draft and recovery selections against complete visible references so hidden metadata cannot activate an invisible skill or make an empty draft sendable. Preserve prose, attachments and valid chips.

## Review and gate repairs

Independent reviews covered reuse, simplicity, efficiency and abstraction level. Applied behavior-preserving simplifications to schema-based suffix/clipboard text, offset mapping, external document replacement and carrier extraction. Skipped return-identity changes and mutable serialization caching. The final whole-PR review added only helper reuse and collapsed-selection reuse; no assertions or API signatures changed. Final correctness review found no findings.

Jesse chose **Keep and finish**, retaining two independently reviewed verification prerequisites:
- Retain existing provider-dialog rows during background listing refresh so focus survives. Deterministic tests await the permitted store-owned refresh and preserve no-extra-work assertions. The original failures reproduced on base and feature.
- Select Chrome's built-in key store only for disposable Linux browser-guard profiles. A real Chrome comparison isolated the desktop-keyring navigation stall; ordinary browser profiles and application credential storage are unchanged.

## Verification

Rebased onto `726d31c2db81dde3582d91468bd157e771de5338` without conflicts; all ten rebased patches matched. The final restoration repair and two local cleanups passed four focused suites/257 tests, Biome, typecheck and independent review. Five failing regressions established the restoration defect before the fix.

This complete command exited zero on the final source committed in `3e0da1d7b355506057e5879fcdd88bad756a7931`:

```sh
GOMAXPROCS=2 make merge-approval-gate && GOMAXPROCS=2 make vet && GOMAXPROCS=2 make test-web-browser
```

Passed lint and secret scan, production build, all seven Go modules, frontend tests/typecheck/Biome, native bundle, 757 native tests, 778 shared tests, native typecheck/script checks, AppWire package qualification, vet and all six actual browser guards. Hosted CI and required approval remain separate gates.

The successful run emitted Node SQLite experimental warnings, a Metro cold-cache warning and an npm update notice.

The native dependency install also reported 14 moderate audit entries rooted in `decode-uri-component` and `uuid`. Those lockfiles are unchanged from main; I haven’t applied dependency upgrades.

Plan and acceptance record: `docs/superpowers/plans/2026-09-15-inline-composer-skills.md`.

Jesse authorized resolving findings above low, filing low-severity follow-ups and merging subject to repository requirements. Current-head CI/RoboRev, follow-up tickets, required approval and an up-to-date branch remain prerequisites. No protection bypass is authorized.


obra:
## Review disposition at `5281596df49c7c8f960b80a6e39d7f246d0c3b56`

Every claim below was checked against the reviewed source, the installed ProseMirror 1.42.3, and real mounted components. Hosted review of this head is pending; the findings it replaces were raised against `3e0da1d`, `e172193` and `7c5efdab3910a2acb10c1dbc0bd5a77583a74eb1` as the series moved.

- **Image paste retains selected content: repaired and published.** The attachment adapter now carries the editor's full UTF-16 selection to the existing marker insertion helper. Pending cursor restoration still wins, accepted markers replace the range once and then chain, and rejected files do not enter the insertion path. Spawn's adapter retains its existing insertion-at-caret behavior. The original image-only, mixed-paste and plain-text regression assertions pass.
- **Trailing-period restoration drops activation: repaired and published.** The parser accepts a terminal period when the following character does not continue the token. Exact text and active metadata now round-trip for `Run /cleanup.`. Longer dotted canonical names and dotted/path prefix rejection remain covered and passing.
- **A word character typed against a chip: repaired and published.** `/cleanup` plus `d` produced text the parser could not read as that skill, so the activation was silently dropped by the next re-derivation. The editor now separates the two with a space on the side that collided, in the same edit, keeping the label whole and losing nothing the user typed. The parser and the editor share one definition of a complete reference (`completeSkillReferenceAt`), so that boundary rule cannot drift. Jesse chose this over dissolving the chip, and the real-browser guard asserts it, including through IME composition.
- **Reconciliation boundary inside an atom: repaired and published.** The first version of the text patch computed its range from character offsets alone, so with a chip at the start of the draft `from` could land after the atom and `to` before it — a goal replacement produced `/skill-1goal Keep the session focused`, and a prefix quote merged beside the chip instead of replacing it. A boundary that falls strictly inside an atom now rebuilds from the value instead. Two real-editor regressions cover both cases and were shown failing without the change.
- **Programmatic insertions atomizing prose: repaired and published.** Only a value the composer marks as a restore rebuilds the document; every other change is patched, and what a patch inserts is plain text. A spliced command or quote that happens to spell a selected name is therefore never turned into an activation the user did not choose, while a draft, recovery or queued entry still restores its visible selections as chips.
- **Nested editables or nested textbox roles: refuted.** `prosemirror-view` reuses the `mount` node as the view's own dom, so the host div *is* the single editable and the single textbox. A permanent regression asserts exactly one textbox carrying the composer's accessible name.
- **Undo of that separator: repaired and published.** The separator was first recorded as its own history event, so one undo reverted only the space, leaving a reference the plugin immediately re-separated — a visible no-op, with the typed character unreachable by undo. Reproduced first (`expected 'Use /cleanup d' to be 'Use /cleanup'`), then fixed by folding the separator into the creating edit's own event. The browser guard's native undo and redo steps around atoms pass on this change.
- **External value sync destroying undo: repaired and published.** Applying the controlled value rebuilt the editor state, which re-initializes every plugin, so undo stopped working from the first attachment marker onward. It is now applied through a transaction, which keeps plugin state — while still marked `addToHistory: false`, because the suite already asserts deliberately that an external replacement is not itself undoable, so a cleared submission cannot be undone back into the composer. Both properties are tested.
- **Programmatic edits promoting typed prose: repaired and published.** Rebuilding the document from the whole value re-read every mention, so an attachment marker turned plainly typed `/name` prose into a second activation. The value is now reconciled as a text patch over the span that actually changed, and only that span follows the parse rule — which is exactly what a restore replaces — so a restore still atoms its references while a programmatic edit leaves prose alone.
- **Enter or Shift+Enter also inserting a newline: refuted.** Two new real-editor regressions cover bare Enter with `enterToSend` on and Shift+Enter steering; both dispatch their RPC and leave the composer empty with an empty persisted draft. React's capture handler runs first and ProseMirror honours `defaultPrevented`, as the earlier probe on the same claim found.
- **Provider-dialog focus repair: withdrawn.** Upstream #1384 landed a general focus-rehome mechanism whose row gate strictly covers our one-line version, so the rebase dropped ours and that behavior is upstream's.
- **Nested editable textboxes: refuted.** `SkillEditor.tsx` uses `new EditorView({ mount: host.current }, ...)`. ProseMirror reuses that node (`prosemirror-view/src/index.ts:77`); it does not append another editable. The real DOM probe confirmed one textbox and a noneditable skill atom.
- **Prevented Enter inserting a newline or retaining the submitted draft: refuted.** React capture runs before the native editor listener. ProseMirror's `eventBelongsToView` rejects `event.defaultPrevented` (`src/input.ts:91-98`). Real-component probes covered Ctrl+Enter, bare Enter and Shift+Enter steering, exact request payloads, empty visible/persisted drafts after success and subsequent typing. An uncancelled Enter positive control inserted a newline.
- **Stale ARIA references: refuted; durable regression coverage deferred.** ProseMirror removes old attributes omitted from the replacement map (`src/viewdesc.ts:1132-1135`). Real probes verified removal after rerender, Escape and completion selection. The follow-up (https://github.com/prime-radiant-inc/evener/issues/1449) preserves this behavior in permanent tests rather than adding redundant production cleanup.
- **Unused `skillSelections.ts` helpers: confirmed; cleanup deferred.** `addSkillSelection` and `removeSkillSelection` are referenced only by their definitions and tests. The distinct recovery-local helper remains live and must be preserved. Filed separately as https://github.com/prime-radiant-inc/evener/issues/1450; no helper or test deletion is included in this repair.
- **Outdated composer documentation: confirmed; filed.** `docs/skills.md` still describes removed completion tokens and detached skill controls. Filed unchanged as https://github.com/prime-radiant-inc/evener/issues/1451.
- **Stale `writeComposerDraft` comment: confirmed; filed.** The comment still says a selection-only draft sends, which the inline-chip invariant no longer allows. Filed as https://github.com/prime-radiant-inc/evener/issues/1554.
- **Shared atom boundary gaining two separators: filed.** Deleting the separator between two chips breaks both atoms at one boundary, so the integrity plugin inserts a space for each. Filed as https://github.com/prime-radiant-inc/evener/issues/1593.
- **A trailing colon not exempted like a trailing period: filed.** `:` continues a token, so `/review: fix` is not read as a complete reference and gains a separator; the exemption should match the `.` case without letting a longer colon-qualified name lose to its prefix. Filed as https://github.com/prime-radiant-inc/evener/issues/1594.
- **Post-insert caret computed from an array length: filed.** `insertSkill` sets the caret with `from + content.length`, correct only because the atom and the separator each happen to be size one. Filed as https://github.com/prime-radiant-inc/evener/issues/1608.

Independent final correctness review of the preceding artifact found no findings and confirmed every rebased feature patch was preserved. Post-cleanup Biome, the focused suites and typecheck passed. The complete `GOMAXPROCS=2 make merge-approval-gate && GOMAXPROCS=2 make vet && GOMAXPROCS=2 make test-web-browser` exited zero on the published source, including all six actual browser guards.

These were independently validated, not accepted on report: each fix began as a failing real-editor regression, and the claim-level dispositions above distinguish repaired, refuted and filed rather than treating the review as uniformly correct.

The native dependency install also reported 14 moderate audit entries rooted in `decode-uri-component` and `uuid`. Those lockfiles are unchanged from main; I haven’t applied dependency upgrades.

Merge remains conditional on current-head hosted CI/RoboRev, an up-to-date branch and the required approval. Main has advanced since publication, so a catch-up rebase is expected immediately before merge. No merge or protection bypass has occurred.

PR #1412, opened 2026-09-16
transcript: the writer owns durability (piece A1 of the #1100 split)

obra:
Piece A1 of the #1100 split (Jesse's ruling 2026-09-15: close #1281, rebuild on the transcript's existing `Entry.Seq` as small PRs). The transcript writer now owns the retained-durability contract that #1281 kept re-finding at its callers, and this closes that drift for the general producers while leaving the durability owners their barrier. Groundwork toward #1200 (history turns carry their `Seq`), which lands in A2.

## What the writer now owns

- **`AppendBatch([]schema.Turn) (firstSeq int, err error)`** — every turn in one write and one fsync, all-or-nothing: either every line is in the file at contiguous sequence numbers, or a rollback to the batch's start offset leaves none of them and spends no sequence number.
- **A durable append whose whole line landed but did not sync returns `nil`** — the entry is a record every returning reader finds, so it belongs in history — and reports the sync failure through **`Writer.TakeWarning()`**, which the session surfaces exactly once, outside the locks a warning's notification hook needs. The buffered door does the same. `ErrEntryRetained` never crosses the writer boundary.
- **Two poison causes** (carried by content from #1281 r16 `5590daa5f`): a partial line is permanent — no fsync makes half a line whole; a retained whole line is cleared by a successful `EstablishDurability` barrier, because that fsync is exactly what the record was missing.

## Callers

Ordinary producers never branch on any of this: a non-nil error from a durable append means "not recorded," and their existing `if err != nil` paths are correct by construction.

The **named exceptions are the durability owners** — the environment producer and the four delegate-attention side-writes — which raise their own `EstablishDurability` barrier and must know a retained entry is readable but not yet durable. They check `Poisoned()` after the `nil` return and run their existing reconcile (barrier + read-back); their tests assert that behaviour. An environment write that cannot establish durability stands the session down (the next turn is refused loudly) rather than re-emitting onto an unsynced tail.

## Deferred to A2

The `(int, error)` return for `Append`/`AppendDurable` lands with A2, where the seq has a consumer (history turns carry their `Seq`, seeded on restore) and the ~180 call sites are already being touched. `AppendBatch` returns `firstSeq` today.

## Gates

toolchain gofmt; `go vet` host / `-tags evenerfuzz` / `GOOS=windows -tags evenerfuzz` in both modules; golangci-lint 0 issues; `agent/transcript` and `agent` suites; `-race -count=3` on `agent/transcript` and the `agent` package. Refs #1281, #1200.

## Round 1 (simplify)

Jesse's required /simplify pass. It reverses part of #1281 r16: **a retained
whole line is dirty debt, not poison.** A line that reached the file but did not
sync is exactly what a buffered append already is — tracked as `dirty`, settled
by the next successful fsync (any `AppendDurable`, or `EstablishDurability`). So
`poisonCause` and the clear-on-barrier are gone; the **only** poison is a
partial line whose rollback failed (permanent), and that is the only thing that
stands the session down.

- **Two doors, plain errors.** `Append`/`AppendDurable` are recorded-or-error;
  ordinary producers never branch on durability (`if err != nil { not recorded }`).
  The durability owners — the environment producer and the four
  delegate-attention side-writes — use a new **`AppendSynced`**, which returns an
  error unless the line is recorded AND synced. Nobody reads `Poisoned()` after a
  nil. The env producer collapses to write-through-the-pair-helper → commit on
  nil, rewind on error; its read-back reconcile (and `environmentEntryOutcome`)
  is deleted, since `AppendSynced` owns the barrier.
- **One primitive.** `appendBatchLocked` (one buffer via `json.Encoder`, one
  seek, one write, one optional sync, one rollback, one classifier
  `settleFailedWriteLocked`). `append` is a batch of one; `AppendBatch` is the
  thin exported wrapper — still no caller until A2's fold.
- **Typed warnings.** The writer queues `error`s; `DrainWarnings` returns them
  (guarded by an atomic count so the lock is skipped when empty); the session
  formats through one `drainTranscriptWarnings(sink)`. The env/rollback tests use
  `errors.Is` again.
- **Fixtures.** Writer tests use the package `fault.FS` schedule and the existing
  `partialWriteFS`; the bespoke afero wrappers are gone.

A consequence of the reversal, reflected in the tests: an environment write
whose sync fails but whose barrier then succeeds is now a clean, committed
success (not a rejected-but-recorded turn), and one whose barrier also fails
rewinds the tracker and re-emits next turn (the writer stays usable) rather than
standing the session down.

## Gates

toolchain gofmt; `go vet` host / `-tags evenerfuzz` / `GOOS=windows -tags
evenerfuzz` in both modules; golangci-lint 0 issues; `agent/transcript` and
`agent` suites; `-race -count=3` on `agent/transcript` and the `agent` package.


obra:
## /simplify pass (645f5bf5b)

Four reviewers read the diff. Deduplicated, being applied as round 1; the altitude finding changes the contract and I am taking it, which reverses part of the ruling I made on #1281 round 16:
- **A retained unsynced line is a durability debt, not a poison.** The writer already tolerates unsynced records for every buffered append and tracks the debt as dirty; the next fsync settles it. So the durable door's failure downgrades that append to buffered semantics with a warning, and the only poison is the partial line that truncate could not repair. The two-cause enum, the clear-on-barrier and the poison flag threaded into the environment reconcile go.
- **Two doors, plain errors.** Ordinary producers use the recorded-or-error door; the five durability owners use a synced-or-error door, so nobody reads the poison state after a nil return. The five copied write-then-check sites collapse.
- **One append primitive.** The single append becomes a batch of one, so there is one write, sync, rollback and failure classification; A2's fold uses the same path and the sequence-number return falls out of it.
- **Typed warnings.** The queue carries errors, not strings, so callers keep error identity instead of matching substrings.
- **Fixtures** reuse the package's existing fault filesystem and partial-write fake.

obra:
## Review round 1 disposition (a3a83a12f)

CI: 13 green, `web` FAILED (this PR touches only `agent/`; failure under investigation, rerun if it is the known load flake). RoboRev: two High, three Medium, two Low.

The Highs both go to the contract the simplify round set ("retained whole line = dirty debt, settled by the next fsync, signalled by a typed warning") and ask who owns that signal. Rule by measurement, not by the finding text:

- **H1 `AppendDurable` returns nil on failed sync while `writeTranscriptDurableLocked` still serves crash-ordering callers (skill-carrier, delivery-commit records)** — enumerate every `writeTranscriptDurableLocked` caller and classify: does it commit an obligation that recovery reads as "delivered" before the record is durable? Every caller in that class is a durability owner and moves to `AppendSynced`; the rest stay recorded-or-error. Post the table.
- **H2 buffered `Append` paths return nil and queue a warning nobody drains; `Close` does not drain either** — measure where `drainTranscriptWarnings` is called today and which write paths can queue a warning that no later call surfaces. The warning must have exactly one owner per write path (drain after each unlocked write, and at shutdown), or the debt must surface through the next barrier's error. Choose the one the writer contract already implies and pin it with a test that fails a buffered sync and asserts the diagnostic reaches the sink.
- **M1 `AppendSynced` drains the entire shared warning queue** — accepted in principle (it can delete another append's only failure signal across the `attentionMu` release); fix so it drops only its own entry and joins the barrier error with the retained cause. Confirm with a two-writer test.
- **M2 `AppendSynced` always runs a second barrier; a failed extra barrier after a successful append sync fails the call and the environment path rewinds and retries** — measure whether the duplicate environment entry is reachable; if yes, only raise the recovery barrier for an ambiguous retained append.
- **M3 held pre-attach turn returns nil from `writeTranscriptSyncedLocked` and the environment path treats it as committed** — measure; the held case must not be reported as recorded-and-durable to a durability owner.
- **L `ErrRollbackFailed` doc describes the removed contract; stale test comments contradict their assertions** — accepted.

Round 2 with the implementer.

obra:
## Round 2 fixes

Head: 1b723c728. Measured before ruling; evidence inline.

**web CI job:** this PR touches only `agent/`; the `web` job runs the frontend suite, which nothing here changes. It is the known load/platform-sensitive flake (same class as the pre-existing defects hardened in the recent `test(serve)` / load-test work), not our tree. No web files are in the diff.

### H1 — durability owners → AppendSynced (accepted)

Enumerated every `writeTranscriptDurableLocked` caller and classified by the test: *does it commit cross-store state recovery trusts, such that a retained (unsynced) record lost to a crash is irrecoverable (not merely re-done)?*

| caller | cross-store commit on success | verdict |
|---|---|---|
| `recordSkillCarrierDurably` | skill obligation/receipt consumed (saveMeta); carrier is the only copy | **owner → AppendSynced** |
| `appendToolResultsWithDeliveryCommitsDurably` | `commit.Complete()` removes the delegate PendingDelivery | **owner → AppendSynced** |
| skill-reload reminder (`prepareCompactedSkillReloads`) | handoff consumed via `consumeSkillReloadReminders`→saveMeta | **owner → AppendSynced** |
| `recordClientMutationFailure` (×2: input + diagnostic) | client-mutation journal marks failed | **owner → AppendSynced** |
| `appendClientMutationTranscriptLocked` (claimed user input, client steering) | claim consumed (`markClaimedUserTranscriptIncorporated` / `finalizeIncorporatedSteering`) | **owner → AppendSynced** |
| delegate seed (`preseedInput` via `appendTurnWithSyncedTranscriptMessage`) | child adopted/run on the seed; read-back sees an unsynced line as present | **owner → AppendSynced** |
| `appendAssistantTurn` | none — regenerable; recovery reads only the transcript | not owner (AppendDurable) |
| `appendToolResultsDurably` (terminal job) | none — the jobstore is the durable source; the tool_result is a rebuildable projection | not owner (AppendDurable) |
| fold merged-tail copies (`publishFoldTransaction`) | fold publication durability is rebuilt on `Entry.Seq` in A2; the receipt trusts the fold markers, not these copies | not owner in A1 (AppendDurable) |
| daemon steering (`appendSteeringTurnDurablyForOwner`) | none — notification delivery is reconciled from the transcript; **and it is recorded pre-attach via holding, which a synced write cannot do** (`TestDurableTurnRecordedBeforeTranscriptWriterExistsReachesTheFile`) | not owner (AppendDurable) |
| delegate steering (`appendDelegateSteeringDurablyWithMetadata`) | none — reconciled from the transcript on redelivery | not owner (AppendDurable) |

`AppendSynced` returns nil only when the line is recorded AND synced, so an owner keeps its obligation pending on any durability failure.

### H2 — buffered retained debt has one owner per path (accepted)

Measured: `surfaceTranscriptWarnings` already drained after the pair helper, `recordTurn`, and `writeTranscriptDurable`; `attachTranscript` drains pre-SESSION_START. Gaps where a buffered retained write queued a diagnostic nothing surfaced: `appendUserInputTurnRefusingPoison`, `handleCompactionTurn`, and the fold flush — each now drains after its unlocked section. Construction paths (`fork`, `session_init`) write to a `tw` they then `Close()`, and Close fsyncs the dirty debt durable, so the record is not lost and no drain is owed. Pinned by `TestBufferedRetainedWriteDiagnosticReachesTheSink` (buffered sync fails → the diagnostic reaches the event sink).

### M1 — AppendSynced no longer drops another append's warning (accepted)

`appendBatch(..., queueRetained=false)` hands the retained diagnostic back to `AppendSynced` as a return value instead of queuing it, so `AppendSynced` never touches the shared warning queue; the barrier failure is `errors.Join`ed with the retained cause. `TestAppendSyncedDoesNotDropAnotherAppendsQueuedWarning`.

### M2 — AppendSynced barriers only for a retained append (accepted)

Measured reachable (the env path runs a barrier every write). Fixed: the recovery barrier runs only when the append's own fsync failed (`retained != nil`); a clean durable append is not fsynced twice. `TestAppendSyncedDoesNotDoubleFsyncACleanAppend` (faults the would-be second fsync; the call succeeds).

### M3 — a held pre-attach turn does not read as durable (accepted)

`writeTranscriptSyncedLocked` refuses (errors) a not-yet-attached turn instead of holding it — a held turn is neither recorded nor synced. Not reachable in production (the env tracker guard and the seed's read-back both preclude a pre-attach owner), so this is a fail-closed guard. `TestSyncedWriteRefusesAHeldPreAttachTurn`.

### L — stale docs (accepted)

`ErrRollbackFailed`'s doc and the comments above `TestAppendDurable_WholeLineWriteFailureSpendsSequence` and `TestEnvironmentAmbiguousWriteDoesNotDuplicateEntry` now match the retained-is-debt contract.

obra:
## Review round 2 disposition (1b723c728)

CI 15/15 green. RoboRev: one High, one Medium (two reviewers), two Low; one reviewer clean.

- **High `AppendSynced` returns an error while the complete record is retained (fsync and rollback both failed); owners rewind and retry → duplicate environment / client-input records, replayed on restart** — accepted as a contract gap, measure first. The r1 simplify deleted the reconcile-by-stable-id path, so an owner can no longer tell "recorded, not durable" from "not recorded". RED: fail the fsync and the rollback under `AppendSynced`, let the environment owner retry, assert two entries. Fix at the door, not per caller: `AppendSynced` must distinguish the two outcomes (a typed retained-unsynced error carrying the seq, or recorded-or-error plus a separate synced flag) so owners adopt the retained record instead of re-appending. Duplicate-free on restart is the contract; pin it.
- **Medium two steering sites (`delegate_tree_steer.go:485`, `session_queue.go:1181`) still on `writeTranscriptDurableLocked` while the synced door's comment lists steering as an owner, and `session_lifecycle.go:2438` marks the notification delivered on nil** — the r2 table kept steering on the recorded door with pre-attach-holding evidence; the reviewer's point is the site, not the class. Measure each site: if a nil return lets it commit cross-store state (delivered mark, claim) before the record is durable, it is an owner and moves to the synced door (the hold path needs its own answer, not an exemption); otherwise the door comment stops listing steering. Cite kata h8mq's test either way.
- **Low `AppendSynced` returns nil for a closed writer** — accepted; a durability owner must fail closed on a closed writer (the nil no-op stays only for a session that never had one).
- **Low test comment says retained poisons the writer** — accepted; retained is unsynced debt.

Round 3 with the implementer.

obra:
## Round 3 fixes

Head: 29cad6f7b. Measured before ruling.

### HIGH — retained-unsynced is duplicate-free on restart (accepted, fixed at the door)

Measured the gap: at 1b723c728, `AppendSynced` returned a plain error when the append's fsync AND the rollback truncate AND the barrier all failed, even though the whole record was in the file — so an owner discarded its state and re-appended, landing the record twice (two on reload, no load dedup).

Fixed at the door. `AppendSynced` now has three outcomes:
- **nil** — durable (its own fsync, or the recovery barrier, succeeded); owner commits.
- **`*RetainedUnsyncedError`** (`errors.Is(err, ErrRetainedUnsynced)`, carrying the record's `Seq`) — the whole line is in the file but could be synced by neither its fsync nor the barrier. The owner **adopts** it (records it in history, never re-appends) and the next fsync settles the debt; the diagnostic is queued for the session to surface.
- **any other error** — nothing recorded; owner keeps its obligation pending.

`appendTurnAfterTranscriptWriteLocked` adopts a `RetainedUnsyncedError` (commits history + pair log, returns nil), so every owner is duplicate-free with no per-caller branching. The environment owner adopts too: an unresolved retained env write is recorded once, not rewound-and-re-emitted.

RED / pins: `TestAppendSyncedReportsRetainedUnsyncedForAdoptionWithoutDuplication` (door: one line in the file, typed error, seq, queued warning) and `TestEnvironmentUnestablishedDurabilityAdoptsTheRetainedRecord` (owner: one durable entry, not re-emitted next turn, warning surfaced). The client-mutation owners share the same adopt path (`appendTurnAfterTranscriptWriteLocked`).

### MEDIUM — steering sites (measured per site → both synced)

- `session_queue.go` daemon steering (`appendSteeringTurnDurablyForOwner`): a nil return lets `acceptNotificationInput` proceed to `markJobNotificationsDelivered` (session_lifecycle.go:2441) — cross-store delivery committed before the reminder record is durable (the §4.3 / h8mq window, `session_jobtree_drain*`). **Owner → synced.**
- `delegate_tree_steer.go` delegate steering (`appendDelegateSteeringDurablyWithMetadata`): a nil return lets `CompleteSteerPersistence` commit the steering claim before the record is durable. **Owner → synced.**

Both run after attach in production. The `writeTranscriptSyncedLocked` comment already listed steering, so it is now accurate. The pre-attach hold test (`TestDurableTurnRecordedBeforeTranscriptWriterExistsReachesTheFile`, whose only vehicle was the no-prod-caller `appendSteeringTurnDurably`) moves to a non-owner durable write, because the synced door correctly refuses a held turn (a held turn is neither recorded nor synced — `TestSyncedWriteRefusesAHeldPreAttachTurn`).

### LOW — closed writer fails closed

`AppendSynced` returns `ErrWriterClosed` for a closed writer (closeAttachedTranscript closes without clearing `s.transcript`), so a shutdown-time owner never reads a dropped write as durable; the nil no-op stays only for a writer that never existed (a stateless session). `TestAppendSyncedFailsClosedOnAClosedWriter`.

### LOW — comment

The cov_s4 retained comment no longer says "poisons the writer" — a retained whole line is unsynced dirty debt the next fsync settles; the writer stays usable.

Merged origin/main (`--no-ff`). Gates: gofmt, vet ×3 both modules, golangci-lint 0, agent/transcript + agent suites, -race -count=3 on both.

PR #1411, opened 2026-09-16
refactor(sdk): move pathRows into the AppWire package (C22)

obra:
SDK migration plan row C22 (docs/superpowers/plans/2026-09-12-sdk-migration.md).

`cmd/evener-hub/frontend/src/widgets/pathfield/pathRows.ts` → `appwire-client/typescript/pathRows.ts`, test alongside. Name kept: nothing at the package root exported `basename`, `parentOf`, `childrenPrefix`, `isDirEntry`, `buildPathRows`, `pickableRows`, `PathRow` or `PathPickableRow`.

- All five importers (Spawn.tsx, widgets/directorypicker, widgets/pathfield, mobile-native HubPathField.tsx and LaunchResourceRow.tsx) import from `@evener/appwire-client`; no re-export stub at the old path.
- Added to `tsconfig.build.json` `files`, re-exported from `index.ts`, smoke-called in `qualify-package.mjs` (helpers, `buildPathRows` row order, `pickableRows`), README prose, header comment names the consuming surfaces, plan status row.

Falsification: deleting the `index.ts` re-export fails `make test-api-package` with `shipped module unreachable from every published specifier: pathRows`; restored.

Gates run serially, all green: `make test-api-package`, `make test-web`, `make test-native` (Metro bundle exported, 81+7 files), `make test-web-browser`, `make lint`, `go test -run TestScenarioSource .`, `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from cmd/evener-hub/frontend. The relocated test file runs under vitest (21 tests).

PR #1413, opened 2026-09-16
refactor(sdk): move the model catalog view helpers into the AppWire package (C21)

obra:
SDK migration plan row **C21** (`docs/superpowers/plans/2026-09-12-sdk-migration.md`).

## What moves

The model catalog trio leaves `cmd/evener-hub/frontend/src/widgets/modelCatalog/` for the package root at `appwire-client/typescript/`, in dependency order and renamed because the old names were bound to their directory:

| was | now |
| --- | --- |
| `types.ts` | `modelCatalogTypes.ts` (type-only; imports `./types.gen`) |
| `catalogView.ts` + test | `modelCatalogView.ts` + `modelCatalogView.test.ts` |
| `pickerRows.ts` + test | `modelCatalogPickerRows.ts` + `modelCatalogPickerRows.test.ts` |

Every exported symbol keeps its name. `catalogClient.ts` and `scopedCatalog.ts` stay in the widget (they hold the fetch). No re-export stub is left at the old paths.

## Importers rewritten to `@evener/appwire-client`

- web: `widgets/modelCatalog/index.tsx` (`buildPickerRows`, `pickableRows`, the two types; its own `export type { ModelCatalogDiagnostic, ModelCatalogEntry }` now points at the package)
- native: `mobile-native/src/hubModels.ts` (`ModelCatalog`), `mobile-native/src/LaunchModelPicker.tsx` (`buildPickerRows`) - both previously deep relative imports into the web tree

## Package wiring

- `index.ts`: re-exports for all three modules, alphabetical between `./model` and `./railSessionState`
- `tsconfig.build.json` `files`: the three modules
- `scripts/qualify-package.mjs`: behaviour probes for every moved value export (`toCatalogOptions`, `filterCatalog`, `withGroupHeads`, `capabilityLabels`, `formatCost`, `contextWindowLabel`, `rowMeta`, `unavailableLine`, `buildPickerRows`, `pickableRows`)
- `README.md` prose; header comments name both consumers
- the coverage guard's test (`frontend/scripts/package-coverage.test.mjs`) pinned `clientLike.ts` as the only declaration-only module by name; the pin is now a named set that includes `modelCatalogTypes.ts` and asserts each listed module is in the build

## Scenario cards repointed

`model-picker-{fresh-install-no-recent,dated-snapshot-sorts-last,uncatalogued-model-still-renders,recent-reflects-last-5-global}.md` now cite the package paths. The source file's line numbers are unchanged (its header was kept at nine lines on purpose); the test's two cited lines moved by the `// @vitest-environment node` directive (54-55 -> 56-57, 118 -> 120) and were re-verified against the file.

## Falsification

Deleting the `modelCatalogTypes` re-export from `index.ts` makes `make test-api-package` fail with `shipped module unreachable from every published specifier: modelCatalogTypes`; restoring it makes the gate pass again.

## Gates (run serially)

`make test-api-package`, `make test-web`, `make test-native` (after `npm ci --prefix mobile-native`), `make test-web-browser`, `make lint`, `go test -run TestScenarioSource .`, `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend` - all green.


obra:
## Refresh onto 7900e6a2c + simplify

Head `d3bd188486bb2fe6c96ca30c8daf6e92881bb8ad` (merge `64cfdf3e2` + simplify `d3bd18848`).

**Merge.** `git merge origin/main` (7900e6a2c, C22 #1411). Four conflicts, all in the shared package files, resolved as unions: `index.ts` re-exports (alphabetical), `tsconfig.build.json` files, README export prose, `qualify-package.mjs` smoke blocks. `git diff origin/main...HEAD --stat` shows only C21's own 18 files.

**One real collision.** C22 landed `pathRows.pickableRows` at the package root, so C21's `modelCatalogPickerRows.pickableRows` is renamed `pickableModelRows` (its return type is `PickerModelRow[]`). Same rename-on-collision precedent as C14 `railSessionState` and C20 `marketplaceSourceLabel`. The C21 plan row now says so; the module, its test, the web widget and the smoke block use the new name. Residual for a C22 follow-up if wanted: the generic `pickableRows` name now belongs to the path domain; `pickablePathRows` would be the symmetric spelling.

**Simplify (4 angles) — applied:**
- `modelCatalogView.contextWindowLabel` calls `displayFormat.formatTokenCount` for the sub-1M branches instead of re-encoding the "k" rule beside it (the `M` branch stays local; `displayFormat` deliberately never scales past k). Every pinned test case unchanged: 200000→200k, 128000→128k, 1000000→1M, 512→512, undefined→null.
- `package-coverage.test.mjs`: the declaration-only pin asserts `hasRuntimeExport(src) === !DECLARATION_ONLY_MODULES.has(file)`, so a listed module that grows a runtime surface fails too (the comment claimed a pin; the `||` form was only an exemption). This also settles the reviewers' "drop the for-loop" and "reorder the ||" notes: the loop catches a stale entry `every` would never visit, and the read is now load-bearing for every module.
- Web widget `modelCatalog/index.tsx` drops the `ModelCatalogDiagnostic` re-export: nothing in web, native or mobile imported it.

**Simplify — skipped:**
- Repoint the seven `ModelCatalogEntry` importers from the widget/barrel to the package: each also imports the widget's own `ModelCatalog` envelope interface on the same line, so this would split one import into two across seven files outside the diff. Left as-is.
- Trim the qualification smoke block to one call per module: kept per-function for consistency with C19/C22's blocks.
- Blank line after `// @vitest-environment node`: matches `credentialLabels.test.ts`; style only.

**Falsified.** Deleting the `./modelCatalogTypes` re-export from `index.ts`: `make test-api-package` exits 2 with `shipped module unreachable from every published specifier: modelCatalogTypes`. Restored.

**Gates (foreground, this head):** `make test-api-package`, `make test-web`, `make test-native`, `make test-web-browser`, `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend`, root `go test -run TestScenarioSource .` — all green. `gh pr view 1413 --json mergeable` → MERGEABLE (main has since taken #1399, Go-only in `agent/` and `internal/appprojector/`, no overlap).

obra:
## Review round 1 disposition (d3bd18848)

CI: 14 green, `web` FAILED on the skillguard browser guard (`TestSkillComposerBrowser`, caret 0-0 retry then a 15s timeout on the steered turn) — the known load flake tracked in #1308, unrelated to a TypeScript relocation; occurrence recorded there, job rerun requested. RoboRev: one High with a reviewer conflict, one Low.

- **High `contextWindowLabel` delegating to `formatTokenCount` breaks the `1M` contract** — refuted by reading the code: `modelCatalogView.ts:86` keeps the `>= 1_000_000` branch (`${cw / 1_000_000}M`) and delegates only sub-million values; `modelCatalogView.test.ts:115` pins `1000000 → "1M"` and passes in the same CI run. Review 3's reading is the correct one.
- **Low widget barrel drops the `ModelCatalogDiagnostic` re-export** — refuted: it had zero importers in the web app and mobile-native (`git grep` over both trees at this head finds the name only in the package and the barrel itself); a re-export nobody imports is dead code, and the type stays available from `@evener/appwire-client`.

No code change from this round; merge on the rerun going green.

PR #1424, opened 2026-09-16
refactor(sdk): move the launch shared schema into the AppWire package (C18)

obra:
SDK migration plan row **C18** (`docs/superpowers/plans/2026-09-12-sdk-migration.md`).

## What moves

The launch-config engine's three pure modules leave `cmd/evener-hub/frontend/src/panes/settings/sections/launchShared/` for the package root at `appwire-client/typescript/`, renamed with a `launch` prefix so they cannot collide with the spawn row's own `schema.ts` (C17):

| was | now |
| --- | --- |
| `schema.ts` + test | `launchSchema.ts` + `launchSchema.test.ts` |
| `inherited.ts` | `launchInherited.ts` + new `launchInherited.test.ts` (it had no test) |
| `pathListAdd.ts` + test | `launchPathListAdd.ts` + `launchPathListAdd.test.ts` |

Every exported symbol keeps its name. The modules import their wire types from `./types.gen` and each other by the new names (`launchPathListAdd` → `./launchSchema`). No re-export stub is left at the old paths. `git mv` throughout (all five show as renames).

## `LaunchConfigLayerName`

The type `"global" | "project"` moves out of the web store `stores/launchConfig.ts` into `launchSchema.ts` (its consumer `optionSupportsLayer` lives there) and is published from the root. `stores/launchConfig.ts`, `fields.tsx` and `LaunchConfigForm.tsx` import it from `@evener/appwire-client`. A type move, not a shim: the package imports nothing from the web app. `types.gen.ts` types the layer as plain `string` (`defaultableLayers?: string[]`, `getLayer`/`setLayer` `layer: string`), so no generated type already covered it. The three mobile-native sites that spelled the union inline (`LaunchSettingsScreen.tsx`, `launchSettings.ts`, `launchSettings.test.ts`) now use the shared name.

## Importers rewritten to `@evener/appwire-client`

- web (5): `launchShared/fields.tsx`, `launchShared/LaunchConfigForm.tsx`, `launchShared/collectionFields.tsx`, `panes/spawn/AdvancedOptions.tsx`, `stores/launchConfig.ts` (type only)
- native (4): `LaunchOverrides.tsx`, `LaunchScalarEditor.tsx`, `LaunchSettingsScreen.tsx`, `launchPaths.ts` - all previously deep relative imports into the web tree
- `mobile/src`: none (verified by grep). No `vi.mock` of the old paths anywhere. No scenario card cites the three files.

## Package wiring

- `index.ts`: re-exports for all three modules, alphabetical between `./jobOutput` and `./marketplaceSourceLabel`, values and types
- `tsconfig.build.json` `files`: the three modules
- `scripts/qualify-package.mjs`: one real call per module - a `buildFormState`/`collectConfig` round trip, an `inheritedItems` subtraction, and a `validatePathListAdd` duplicate rejection (`.then`-asserted because the same smoke program is also emitted as CommonJS, which has no top-level `await`)
- `README.md` prose; module headers say what each module does and who reads it

## Falsification

Deleting the `launchInherited` re-export from `index.ts` makes `make test-api-package` fail with `shipped module unreachable from every published specifier: launchInherited`; restoring it makes the gate pass again.

## /simplify

Run over the diff before opening. Applied: smoke block trimmed from nine asserts to one per module; three stale `schema.ts's …` comments in `fields.tsx`/`LaunchConfigForm.tsx` and the `stores/launchConfig.ts` header updated for the new home; the moved file's "React-side" banner, `ScalarField.tsx` reference and a dangling "see this task's own report" sentence removed; `LaunchScalarEditor.tsx`'s split package import merged; the three native inline `"global" | "project"` unions adopt `LaunchConfigLayerName`. Skipped: folding one overlapping expectation in `launchInherited.test.ts` (it asserts `inheritedItems`' own guarantee); a pre-existing three-way private duplication of an `isRecord` guard inside the package (`client.ts`, `toolCallText.ts`, now `launchInherited.ts`), which is not this row's change.

## Gates (run serially, all green after the simplify fixes)

`make test-api-package`, `make test-web`, `make test-native`, `make test-web-browser`, `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend`, `go test -run TestScenarioSource .` at the repo root.


obra:
Follow-up filed from the /simplify skip: #1425 (four private copies of the plain-object guard inside the package; predates this move).

PR #1428, opened 2026-09-16
refactor(sdk): move the transcript display config into the AppWire package (C4)

obra:
SDK migration plan row **C4** (`docs/superpowers/plans/2026-09-12-sdk-migration.md`).

## What moves

- `cmd/evener-hub/frontend/src/transcriptDisplay/config.ts` → `appwire-client/typescript/transcriptDisplayConfig.ts` (renamed: `config` was directory-bound), test alongside as `transcriptDisplayConfig.test.ts`. No re-export stub at the old path.
- Every importer rewritten to `@evener/appwire-client` in the same PR: **36 web** files (31 by path, 5 that imported `./config` from inside `transcriptDisplay/`) and **8 native** files (`NativePreferencesProvider.tsx`, `TranscriptPreferencesEditor.tsx`, `nativePreferences.ts` + test, `preferenceDraftRepository.ts` + test, `transcriptPresentation.ts` + test). `mobile/src` had none. Rewritten with `scripts/sdk/rewrite-package-imports.mjs`, then Biome organized the web imports.
- `index.ts` re-exports the module 1:1 (16 types, 51 values), `tsconfig.build.json` `files` gains the module, the qualification runner smoke-calls the constructor, both summaries, the local codec round trip, precedence, inventory and legacy migration, README export prose updated.

## Encoding untouched

`encodeLocalConfig`/`decodeLocalConfig` are a pinned localStorage contract. The module's diff against the old path is the header comment plus one import specifier (`@evener/appwire-client` → `./types.gen`); the test's diff is its import line. Verified with `git diff -M`.

## Name collision

The module exports type aliases `TranscriptDisplayConfig` and `TranscriptDisplayAdvanced`, and the package root already publishes wire types of those names via `export type * from "./types.gen"` (different shapes: the wire `hookExits` is `string`). An explicit re-export would silently shadow the wire types, so `index.ts` omits those two names with the reason. `TranscriptDisplayConfig` had no consumer; `TranscriptDisplayAdvanced` had one (`TranscriptDetailEditor.tsx`), which now reads `TranscriptDisplayConfigV1["advanced"]`. Follow-ups filed for the module-side V1 rename and for the module's zero-consumer aliases (linked below once created).

## Gates (all local, all green, run one at a time)

- `make test-api-package`
- `make test-web`
- `make test-native`
- `make test-web-browser`
- `cmd/evener-hub/frontend`: `npm run lint`, `npx biome ci ../../../appwire-client/typescript`
- repo root: `go test -run TestScenarioSource .`
- Zero `vi.mock` of the old path; zero `transcriptDisplay/config` references outside historical docs (the 2026-08 plan docs and the inventory snapshot keep theirs, as siblings did); zero relative paths into the package.

## Falsification

Deleting the `index.ts` re-export block and running `make test-api-package` fails with
`AssertionError [ERR_ASSERTION]: shipped module unreachable from every published specifier: transcriptDisplayConfig`. Restored; gate green again.

## /simplify

Run over the diff (4 angles). Applied: header comment trimmed and states the pin directly instead of pointing at the web prefs store; smoke probes retargeted at names the apps actually use and two lines that duplicated the unit test's oracle dropped; two native files' duplicate `@evener/appwire-client` imports merged; a stray blank line removed; the index.ts omission comment tightened. Skipped: publishing only the live surface from index.ts (siblings mirror 1:1 and deciding deadness is the module's job → follow-up issue), an `as TranscriptDisplayAdvancedV1` rename at index.ts (C20 ruled out aliasing at the barrel → follow-up issue).


obra:
Follow-ups filed: https://github.com/prime-radiant-inc/evener/issues/1430 (rename `TranscriptDisplayAdvanced` to `TranscriptDisplayAdvancedV1` and publish it) and https://github.com/prime-radiant-inc/evener/issues/1431 (delete the module's zero-consumer aliases). Plan row C4 now records this PR.

Issue #1436, opened 2026-09-16
web: "Steer queue now" while a turn/queue is still in flight is rejected ("queue revision changed"), and the empty rejected drain auto-activates composer recovery


obra:
Found while root-causing the skill guard flake (#1308); the caret half of it is fixed in the PR that references this issue, this is the product residual.

## What happens

1. The user queues a message; its `turn/queue` mutation is still `submitting` (the strip shows the optimistic row without controls), and the strip's "Steer queue now" button is enabled.
2. The user clicks it. `QueueStrip.handleDrain` sends `turn/drainAsSteer` with the client's last-known `ExpectedQueueRevision`.
3. The in-flight `turn/queue` landed first and bumped the revision, so the daemon rejects the drain: `agent/session_client_mutation_queue.go` `rejectClientMutation(record, appwire.Conflict("queue revision changed"))`.
4. The rejected drain becomes a `rejected` recovery record with no input (the composer was empty).
5. `Composer.tsx`'s recovery auto-activation effect activates it (an empty draft into an empty composer), and `queueRecoveryPersistence` immediately discards it again because the recovered draft is empty.

Net effect for the user: the click did nothing visible except a flash of recovery state, the queue is still there, and they have to click again. Measured locally with the skill guard under 6x CPU throttling (`Emulation.setCPUThrottlingRate`): `[sgrecover] {"method":"turn/drainAsSteer","kind":"rejected","reason":"appwire turn/drainAsSteer: queue revision changed","textLen":0}` in 2 of 6 runs, each time when `requeued` still showed the `turn/queue` in the outbox; CI run 35049825465 shows the same 1.6 s `requeued` to `drain-committed` gap.

## Questions for the product owner

- Should "Steer queue now" be disabled (or the drain deferred) while a queue mutation for the same session is still in the outbox, so the drain always carries the revision it will be checked against?
- Should a rejected recovery record that carries no input ever auto-activate? Activating and discarding it in the same breath is the flash above; a rejected empty drain is arguably a toast, not a draft.

Refs #1308.

Issue #1434, opened 2026-09-16
flake: skillguard TestSkillComposerBrowser times out at the 15s steered-turn wait on CI and under local load


obra:
## What happened

`web-skillguard` failed once on PR #1357's CI run [35027940168](https://github.com/prime-radiant-inc/evener/actions/runs/35027940168) (head `e0c89f6c`, 2026-09-15 21:58 UTC):

```
skill_composer_browser_test.go:340: skill browser driver failed: node scripts/skillguard/run.mjs exited exit status 1
skillguard FAIL: timed out after 15000ms waiting for steered turn's input visible in the transcript
FAIL: TestSkillComposerBrowser (28.65s)
```

All five other guards in the same job (layout, overflow, shell, spawn, transcriptscroll) passed in the same run.

## Evidence

- CI: failed once at 21:58 UTC; green in the four subsequent runs on the same PR (35047775279, 35048941173, 35051213635, 35053260830) with no relevant changes between them.
- Local: the same full skillguard sequence is the most load-sensitive of the six guards — it passes in ~58s on a quiet machine and intermittently infra-times-out on a busy one (load 50-65 on 16 cores, ~175 users, disk 90-95%). No local run has ever produced an assertion failure in product code; every failure is the 15s wait expiring.

## Suspected cause

The 15000ms bound on "steered turn's input visible in the transcript" is the tightest wait in the sequence and the only single-point timeout observed failing both on CI and locally. Worth checking whether the wait can be derived from the transcript's readiness (poll the transcript projection) or the bound raised/loaded from a constant that a loaded runner can override.

Related: #1402 (editorial-preview load flake), #1429 (guard teardown EPERM — different failure mode, same guard harness).

obra:
Occurrence on main at 55bce119a (D10 #1476, TypeScript extensions store), CI run 35083213356 `web` job: `TestSkillComposerBrowser (50.09s)` — `skillguard FAIL: timed out after 30000ms waiting for the continuation leg carrying "PROSE_QUEUE_14c second pass" to complete in its own turn`. This is after #1329 (0eb34d94e) landed, so the drained-steering-leg redesign did not remove this symptom on the runner. Rerun requested. 2026-09-16 ~10:15Z.

obra:
[...]

Issue #1429, opened 2026-09-16
browser guards: teardown kill(-pgid) EPERM fails a guard whose checks all passed


obra:
Seen on PR #1340 round 21 (head 3b277abf9), `make test-web-browser`: the shell guard logged every check green (`shellguard ok: ...`, the render-isolation line) and then exited 1 with a bare `kill EPERM` as the last line; the rerun passed.

`cmd/evener-hub/frontend/scripts/browserGuardProcess.mjs` treats the two outcomes of a signal differently from its own liveness probe:

- `processGroupRunning` (line ~377) maps `EPERM` from `kill(-pgid, 0)` to "still running".
- `signalProcessGroup` (line ~388) and `signalProfileProcess` (line ~419) tolerate only `ESRCH` and rethrow `EPERM` from `kill(-pgid, signal)`.

On macOS a group whose surviving member has been re-parented (the comment on `signalProfileProcess` measures Crashpad's handler being orphaned to launchd, and notes a direct-pid signal "intermittently gets EPERM") can answer `EPERM` to the group signal too. The rethrow surfaces from the teardown `finally` in the guard's `run.mjs` after the assertions have already passed, so the run is reported as a failure of the guard.

Expected: a teardown signal refused with `EPERM` is handled like the probe handles it (the group is still there, escalate or report the leak by name), never an unhandled throw that fails a guard whose checks passed. Log: `/private/var/folders/.../T/evener-test-web-browser.0JeapE/shellguard.log` on Jesse's machine (last line `kill EPERM`).

PR #1439, opened 2026-09-16
fix(web): a programmatic caret lands in its own commit, and the skill guard awaits a reply in its turn (refs #1308)

obra:
Refs #1308. The daemon half (the turn-idle barrier passing between a drain's two legs) is #1329's; this PR is the two web-side mechanisms behind today's two failures, both measured rather than inferred.

## Run 35049825465 (PR #1413): `a render moved the caret to 0-0, expected 50` is a product bug

The retry landed attempt 2, so it was not that run's failure (that was #1329's mechanism: seq 4 `PROSE_QUEUE_14c second pass` reached the helper `held=true` 0.66 s after `drain-released`, the steer opener stranded in the client-side queue). But the caret reset is real and users would hit it.

**Reproduction.** Not reproducible by host load: 40/40 local passes with 0 caret retries (20 at load avg ~15, 20 with `make test-web` looping alongside). Reproducible by CPU starvation: with `Emulation.setCPUThrottlingRate` 6x on the guard's Chrome, **9 of 10** runs hit the caret retry, every one at the steer opener typed right after the drain.

**Mechanism**, from a page-side hook on the composer textarea's selection setters plus console traces in the product build:

```
10.267  selectionStart/End = 50        driver's write
10.274  selectionStart = 0  len 50     Session-*.js → React commitLayoutEffects
        [sgrestore] {"cursor":0,"len":50}   with NO [sgwrite] since `requeued`
        [sgrecover] {"method":"turn/drainAsSteer","kind":"rejected","reason":"appwire turn/drainAsSteer: queue revision changed","textLen":0}
```

`Composer.tsx`'s cursor-restore `useLayoutEffect` was keyed on `text`. The guard's "Steer queue now" click went out while the second `turn/queue` was still `submitting`; the daemon rejected the drain (`agent/session_client_mutation_queue.go`, `queue revision changed`); the empty rejected drain became a recovery record; its auto-activation ran `updateText("")` into an already-empty composer (no text change, so the effect did not run) and set `cursorToRestoreRef = 0`. The user's next keystroke changed `text`, the effect fired, the caret went to 0. With per-character typing (before #1311) that is exactly the `ROSE_STEER_TURN_14e … steeringP` corruption the issue saw. The textarea was not remounted and kept focus throughout (measured: same element, same composer root, same pane, `document.activeElement` still the textarea); a value set in Chrome parks the caret at the end, so an explicit selection write was the only candidate, and the hook caught it.

The correlation holds in CI too: the caret run's `requeued`→`drain-committed` gap was 1.64 s with the `turn/queue` still in the outbox at `requeued`; the run without a caret retry had 0.09 s and an empty outbox.

**Fix.** The restore is keyed on a schedule the write bumps (`scheduleCursorRestore`), so it lands in the commit that carries the edit whether or not the edit changed `text`; all three arming sites (`write()`, recovery auto-activation, `activateRecovery`) go through it. RED test `a recovery that leaves the text unchanged does not move the caret on the next keystroke` (seeds a rejected `turn/drainAsSteer` with no input, mounts, types one character: `selectionStart` was 0, now 1). After: **0 of 10** throttled runs hit the caret retry, with the trigger (rejected first drain, re-clicked) present in 2 of them.

## Run 35047337033 (PR #1412): the transport scenario passed without a delivery

`net-failed-kept` 02:22:33.875 → `net-restored` 02:22:33.878 with the `PROSE_NET_14j` mutation still `submitting`; helper alpha's log ends at seq 10; the Go backstop caught it (`found 0 provider requests carrying "PROSE_NET_14j …"`). `waitForReply` judged a reply by a baseline of the turn ids of the MOUNTED reply blocks, and the transcript virtualizes: measured locally, by the last scenario the mounted window is `turn_m3…m11` with `turn_m1`, `turn_m2` unmounted, and the footer's height (draft + chip row typed, then cleared by the submit, which the offline submit does) moves that window. An older reply row mounting back in has an id outside the baseline and reads as a new reply. The wait is now anchored on the turn that carries the submitted prose holding the reply (grouped by `data-turn-id`, since a turn can render as segments); `replyBaseline` and its `.catch(() => [])` are gone. A re-clicked strip control is now logged so a downstream failure can be read against what the app was actually asked.

## Gates

- `make test-web`: PASS (typecheck, test, lint); `npm run lint` from `cmd/evener-hub/frontend`: clean.
- `make test-web-browser`: four runs, all six guards PASS in three; the other failed web-skillguard AFTER `skillguard ok` (37 milestones) on the Chrome teardown `kill EPERM`, which is #1429.
- Local skill guard runs on this branch: 10/10 at 6x throttle, plus 20 quiet and 20 loaded on main for the before numbers. Under throttle the #1329 mechanism also reproduced once on main (`timed out … steered turn's input visible`, 0 caret retries), tracked as #1434.
- No Go changes.

## Filed

#1436: the drain rejected under an in-flight `turn/queue`, and the empty rejected drain auto-activating (and immediately discarding) composer recovery.


obra:
## /simplify pass (f3f19d01e) → 13dd36edd3a8cae4bf5640a5bfefbd2ed1120cde

Four angles over the merge-base diff (three files).

**Altitude: the schedule-keyed cursor restore stays.** The reviewer enumerated the arming sites that can leave `text` unchanged besides the empty rejected drain: `useAttachments.removeItem` and the decode-failure path when the user already deleted the `[image N]` marker by hand (`stripMarker` is a documented no-op), `clearSubmitted` after `clearIfUnchanged` already emptied the text, a slash completion whose token already equals the invocation, `restoreTextToComposer` of an attachment-only queue entry into an empty composer (parks 0, the exact symptom shape), goal replacement onto an identical draft, and `activateRecovery` of an attachment/skill-only record. So the `[text]` keying was the general defect and #1436 alone would close one instance of at least six. The cheaper forms were weighed and rejected: applying the caret synchronously in `write()` when the text matches reads the PENDING `textRef` while the DOM still shows the old value (and two of the three sites do not go through `write()`); a deps-less layout effect only works when some other state happens to commit (`setText(sameValue)` bails out); folding ref and counter into one state breaks `textEditor.read()`, which needs the pending caret synchronously for back-to-back `ingestFiles`. #1436 stays a separate product decision, and today's activate-then-discard is what the new test pins.

**Applied.** The vitest seeds its rejected drain through `seedRejectedRecovery` (now takes `{ method, input }`, default unchanged) instead of a third copy of the intent shape; the driver's turn-block scan lives once (`turnTextsExpr`) and the steer-turn wait uses it too (it was per-element and missed the segment case the reply wait handles); five comments trimmed to the why. Efficiency: clean (no extra render on any arming path under React 19 batching; typing now skips the layout effect entirely; one CDP round trip per reply wait removed).

**Folded the two RoboRev Lows** into the same push: `turnTextsExpr(ref)` is scoped to the target session's pane and `waitForReply(ref, prose)` names it at all seven sites; the re-click notice fires only when `attempt < attempts`.

Gates on the new head: `make test-web` PASS (typecheck, test, lint), `npm run lint` clean, `make test-web-browser` all six guards PASS.

Issue #1394, opened 2026-09-15
flake: race-modules/agent job hung to the 10-minute test timeout on a PR with no agent changes; rerun passed


obra:
Run 35031304981 (PR #1272 @214858b0f, which touches nothing under agent/): the `race-modules / agent` job panicked with `test timed out after 10m0s`; the rerun of the same job passed. Main's own CI at the same base (1988c5b5a) passed. The job log's goroutine dump was withheld by the API at the time, so the hung test is not identified here.

Same class as #1326 (hub `-race -count=2` timing out) and the load-sensitive defects #1370 hardened: a test in the agent package can hang under `-race` on the runner. To do: when it recurs, capture the `running tests:` line and the dump (`gh run view <run> --job <job> --log`), and either fix the hang or give the package a per-test timeout so one hung test names itself instead of the whole job timing out.

obra:
Another occurrence: PR #1462 (mobile-native only) at bcebda89f, CI run 35068479143 `race-modules / agent`: `panic: test timed out after 10m0s` with the goroutine dump listing TestRunShellForegroundWaitErrorFailsJob, TestIntg_NewSession_LateErrorClosesMCPManager, TestLocalJobRetainedReadError, TestRunShellBackgroundWaitErrorFinalizesFailed, TestLocateLocalJob_OpenError in flight. No agent/ code in the PR. Rerun requested. 2026-09-16 ~07:37Z.

obra:
Recurred 2026-09-16 on #1508 @fcf33c233 (TypeScript-only PR): run 35095077067, `race-modules / agent`, `panic: test timed out after 10m0s`; the only test listed as running at the timeout was `TestTrimActivityTreeToFit_TrimsExcessEntries (17s)`, but the interesting goroutine is 8224, in `select` for 8 minutes at `agent/delegate_tree_stop.go:801` (called from :696/:704/:137), with a job watcher parked at `agent/job_watch.go:3960`. Main was green at the PR's base 754f5669b. Rerun requested. Two data points now name delegate_tree_stop's stop join as the parked goroutine; worth a bounded-wait audit there.

obra:
[...]

PR #1462, opened 2026-09-16
fix(native): apply newer data automatically, no refresh affordance (closes #1257)

### LINKED ISSUE #1257: Native app: auto-refresh stale pages instead of asking the user to refresh
Ruling from Jesse (2026-09-13): the mobile app should not show a "refresh to see the latest" affordance when newer data exists; it should refresh itself.

Where the manual-refresh pattern lives today (main at 04889ba31):
- `mobile-native/src/PinCatalogList.tsx:57-58` — "Refresh to see the latest pinned sections." when `stale`, with editing disabled until the user refreshes (`:103`).
- `mobile-native/src/PinAssignmentEditor.tsx:55,66,159`, `PinAssignmentScreen.tsx:66-73`, `PinSectionEditorScreen.tsx:58,134`, `PinSectionsScreen.tsx:35,93-104` — the same `page.stale` gate on the pin screens.
- `mobile-native/src/ProjectSessionsList.tsx:119-162, :298, :456` — a stale notice row injected into the project/session list and a per-item stale state.

The `stale` flag comes from the navigation page state (revalidation marks a page stale when the hub reports a newer version). The wanted behaviour: when a page is marked stale and the app is connected, re-fetch it automatically (debounced, once per invalidation, coalescing bursts), replace the content in place without losing the user's scroll or an in-progress edit, and keep the current "stale" affordance only for the offline/disconnected case where a refresh is impossible. Editing controls should not be disabled behind a manual refresh when the app can refresh itself.

Scope note: this touches the native navigation store (`mobile/src` stores that D13/D14 of the SDK migration plan will replace with a shared revalidator) — do it in a way that the shared revalidator inherits, or land it after D13 lands. Sequence: after the #1116 landing queue and the current SDK phase A work; Jesse asked for it "when all of this is landed or you have extra cycles."

obra:
Closes #1257. Jesse's ruling (2026-09-13): the native app must auto-refresh when newer data exists and never show a "refresh to see the latest" / click-to-refresh affordance.

## Root mechanism

`NavigationPages.invalidate()` only ever called `markStale()` (stale flag + "Refresh to continue" error text) and `more()` refused while stale; every screen affordance sat on top of that flag. The web revalidator's `raise()` marks stale **and** restarts the fetch (or reruns after the in-flight one). The store now does the same, in a shape D13's shared revalidator can inherit:

- A matching invalidation marks the page stale (no error text) and re-reads it while idle, following pages back to the depth the user had scrolled to so the list keeps its length and their place.
- A read in flight absorbs the notification through its revision checks and retries **once** per read chain (the retry `refreshAfter` did by hand, moved into `load()` so the receipt survives). An invalidation arriving during a read does not schedule another one. This bounds a hub that notifies on every read to one retry per read (`pinNavigation.test.ts` "bounds receipt readback when notification gaps continue" is the guard).
- A deeper page that fails its checks funnels into a re-read from the top; `more()` on a stale idle page catches up from the first page, so callers stop branching on `stale`.
- An invalidation whose revision the store already holds is not a read; a `not_modified` first page keeps every loaded page.
- `cancel()` pauses automatic re-reads and keeps the owed one for `resume()` (PageList on focus, project browser on `focused`), so background stack screens stop re-reading on every session state change.
- `watch(owner)`: the pin screens display a session location + section descriptor that only `refreshPinNavigation` re-reads, so their hook owns the re-read via `followPinCatalog` (one read at a time, never during a pending mutation, retried once if it ends stale). `NavigationActions.reconcile` is **not** driven from invalidations: it flips to `uncertain` on any failure and would have blocked editing on a transient race.
- One `singleFlight` coalescer replaces the three hand-rolled dirty/running drains (store re-reads, pin follower, and the pre-existing `RosterSearch.watch`).

## Enumeration: what held newer data back, and why

| # | Where | Gate / affordance | Why it was gated | Now |
|---|---|---|---|---|
| 1 | `navigationPages.ts` `invalidate()` | `markStale()` + "This list changed while you were browsing. Refresh to continue." | refresh resets to page 1 (scroll loss); readback-floor machinery made "wait for the user" the simplest storm-free option | re-read to loaded depth, bounded |
| 2 | `navigationPages.ts` `load()` below-target / generation / uncertain | mark stale, return false | same | retry once in place when a notification arrived mid-read |
| 3 | `navigationPages.ts` `load(false)` version mismatch | mark stale | can't combine pages from two revisions | re-read from the top |
| 4 | `ProjectsScreen.tsx` PageList | "Updates are available. Refresh to see the latest list." + "Refresh list"; load-more and organization actions disabled while stale; snapshot-identity guard on the organization sheet; on focus refreshed only if `!loaded` | mutation safety against an outdated list | copy gone; `loading` still gates; sheet uses the absolute state the user chose (archive/favorite set absolute targets, reconciliation is owned by `NavigationActions`); `resume()` on focus |
| 5 | `PinCatalogList.tsx` | "Refresh to see the latest pinned sections."; permanent "Refresh sections" button; load-more disabled while stale | list refresh was manual | copy gone; button only when disconnected, uncertain, or errored |
| 6 | `PinSectionsScreen.tsx` PinnedSectionScreen | `ready` included `!stale` (every row press disabled); header "Refresh section" when stale | same | stale removed from both |
| 7 | `PinAssignmentEditor.tsx` | `blocked` included `stale \|\| loading` (radios + TextInput disabled, keyboard dismissed mid-typing); "Refresh the sections before changing a pin assignment."; "Refresh sections" when stale | edit against an outdated section list | editing never blocked by a re-read; a send's own confirmation read supersedes any read in flight (request counter) |
| 8 | `PinAssignmentScreen.tsx` | `blocked`/`currentAssignment` on `page.stale`; `change()` dropped keystrokes when blocked | same | stale/loading removed |
| 9 | `PinSectionEditorScreen.tsx` | `fresh` included `!stale && !loading` (TextInput not editable, "Refresh section" shown); delete confirm silently no-oped if the **catalog version** moved while the alert was open | delete what the user was shown | editing open; delete confirm compares the **section descriptor** (name/count) as shown and tells the user if it changed, instead of keying off unrelated catalog churn |
| 10 | `ProjectSessionsList.tsx` | injected "This project has updates. Refresh" row; footer "Projects have changed. Refresh projects"; error rows hidden while stale | manual refresh | rows gone; pause/resume with `focused` |
| 11 | `projectBrowser.ts` | `retry` no-oped while stale (a stale+errored page had no recovery); load-more guards | store refused stale paging | `more()` handles stale; guards dropped |
| 12 | `usePinNavigation.ts` | location + section descriptor only re-read by `reconcile()` (focus / user refresh) | -- | `followPinCatalog` |
| 13 | `navigationReveal.ts` | "This list changed. Refresh to locate the session." | -- | reworded; reveal still surfaces "Locate again" on a genuine race |

Not gates (already apply automatically, unchanged): `nativePreferences.ts` applies `settings/*/changed` notifications immediately (its "Refresh to inspect" text is a decode-failure error path); `rosterSearch.ts` already auto-refreshed on invalidation (now via `singleFlight`); `ProjectSessionsList` reconnect `needsRefresh`; `mobile/src/state/conversation.ts` "stale" = internal token checks. Remaining "Refresh"/"Reconnect" affordances in the pin screens are for **uncertain mutation outcomes**, disconnection, and errors, not for newer data.

No gate exists for a reason that makes auto-apply wrong: pin drafts are device-persisted and separate from page data, so a refresh merges around them; the delete confirm keeps its "what I was shown" property in a narrower form.

## Per-screen before / after

- **Projects (ProjectSessionsList)**: before, a session state change on another device showed "This project has updates. Refresh"; after, the expanded project's sessions re-read themselves (both tiers, `projectBrowser.test.ts` "re-reads an invalidated project's sessions without a refresh call") and pause while the tab is unfocused.
- **Live / Needs you / project session lists (PageList)**: before, "Updates are available. Refresh to see the latest list." and disabled load-more; after, rows replace in place to the scrolled depth (`navigationPages.test.ts` "re-reads every loaded page so the list keeps its length").
- **Pinned sections / pinned section / pin assignment / section editor**: before, "Refresh to see the latest pinned sections", disabled radios and a read-only name field until the user refreshed; after, the catalog, the session's location and the section descriptor re-read on invalidation while typing continues (`pinNavigation.test.ts` "following the pin catalog").

## Gates

- `make test-native`: Metro export OK, 83 files / 789 native tests, 784 shared tests, `tsc`, `check:scripts` -- green on `bcebda89f`.
- `make test-api-package` / `make test-web`: not run; nothing under `appwire-client/` or `cmd/evener-hub/frontend/` changed.
- No biome over `mobile-native/` (no config there); added lines hand-wrapped to the files' width.
- `/simplify` run (4 angles). Applied: `singleFlight` extraction; owner callback held on the store so every re-read reaches the owner; `cancel()`/`resume()` replacing the focus-time stale refresh; `more()` owning stale; held-revision skip and `not_modified` keeping pages; delete-confirm descriptor compare; options bag for `load` flags; promise chain instead of the retry IIFE; dropped `markStale()`, an `isCurrent` double-check, two comments justifying deleted code. Skipped: having the follower observe store settle and `refreshPinNavigation` skip its catalog read when settled (changes `refreshPinNavigation` behaviour outside this diff; a D13 candidate).


obra:
## Review round 1 disposition (bcebda89f)

CI: `race-modules / agent` failed on a PR that touches only `mobile-native/`; treated as the known race-suite flake (#1394 class), rerun requested. RoboRev: two Medium, both accepted.

- **M1 `ProjectsScreen` focus effect only resumes in the non-reveal branch; `SessionLocationScreen` always passes `revealRef`, so after the first blur (or "Locate again") the page is paused forever and invalidations queue without draining** — accepted. `resume()` runs whenever `ready`, before the reveal/else split. RED: reveal → blur → invalidation → assert the re-read drains.
- **M2 `invalidate()` during an in-flight read relies on that read's revision check, but `cancel()` bumps the request id so the continuation exits early; the owed re-read is lost and `stale` stays false** — accepted. Record the owed re-read for any newer invalidation regardless of `loading` (stale flag or explicit owed marker) so `cancel()` queues it and `resume()` drains it. RED: invalidate mid-read → cancel → resume → assert one re-read.

Round 2 with the implementer.

obra:
## Round 2 fixes (5460e15cb)

Merged origin/main (`c026bf811`) with `--no-ff`.

1. **PageList never resumed on the reveal path** (`ProjectsScreen.tsx`). `pages.resume()` now runs whenever `ready`, ahead of the reveal/else split, so a SessionLocationScreen that always passes `revealRef` un-pauses on refocus. Also fixed at the store: `refresh()` lifts the pause (an explicit read means the owner is active) and a successful read `settle()`s the owed request it stands in for, so the reveal's own refresh does not queue a second read. RED at the store level, since `mobile-native` has no component renderer: `navigationPages.test.ts` "an explicit refresh resumes a cancelled store and satisfies the owed re-read" (refresh → cancel → invalidation → refresh → later invalidation re-reads; failed with 2 requests before, 3 after).

2. **Owed re-read lost when the read it interrupted was cancelled** (`navigationPages.ts` `invalidate()` / `cancel()`). `invalidate()` now publishes `stale: true` before the loading gate (the rows predate the announced change either way); only the *scheduling* still defers to the in-flight read. `cancel()` therefore queues the owed re-read and `resume()` drains exactly one read to the announced revision. RED: "keeps an owed re-read when the read it interrupted is cancelled" (invalidate mid-read → cancel → resume → asserts one re-read at offset 0 and 3 requests total).

Gates: `make test-native` green on this head (Metro export, 83 files / 791 native tests, 784 shared, tsc, check:scripts). No biome over `mobile-native`.

obra:
## Review round 2 disposition (5460e15cb)

CI 15/15 green. RoboRev: one Medium, two Low.

- **M the deliberate "stay stale" state (hub served a revision below the one it announced; the store stops re-reading by design) now has no indicator and no recovery path, and `PinSectionEditorScreen` labels that data "Currently"** — accepted as a real gap, with the fix direction bound by Jesse's ruling: no click-to-refresh affordance comes back. Recovery must be automatic: the store re-reads when the next notification, reconnect or focus arrives (the anti-spin test keeps its meaning: no hot loop on the same older revision, one re-read per new trigger), and while stale the screens show a status ("Updating…"), not a button. `settled` must include `!stale` again so stale data is never labelled current.
- **L delete confirmation treats a section that left the loaded rows as unchanged** — accepted; a missing row is a change (block and re-prompt).
- **L an invalidation landing during the pin read is dropped instead of coalesced** — accepted; request the trailing run unconditionally (that is what `singleFlight` is for).

Round 3 with the implementer.

obra:
## Round 3 fixes (81780d8bf)

Merged origin/main (`5073378e0`) with `--no-ff`. No click-to-refresh affordance was added anywhere; the new indicators are muted status text.

1. **M — under-served stale page had no trigger and no indicator.** Store (`navigationPages.ts`): `invalidate()` schedules one re-read for a page that is stale and idle on *any* notification (related or not), and `resume()` does the same, so a hub that served an older revision than it announced gets one more read per later trigger and never a loop (the anti-spin test keeps its meaning: with no further trigger the store stays quiet). RED: "re-reads a page the hub under-served on the next notification, once per notification" (older-revision response → stale → unrelated notification → exactly one read; a second unrelated notification → exactly one more) and "…when it is resumed". Screens: `PinCatalogList`, `PinAssignmentEditor`, `PinSectionEditorScreen`, `PageList` and `ProjectSessionsList` (page rows and projects footer) show a muted "Updating…" while stale; `PinSectionEditorScreen` labels the count "Last seen" while stale via `fresh = settled && !stale`, and `PinAssignmentScreen.currentAssignment` requires `!stale` again. Editing controls stay open while stale (the ruling's "never block"); only the labels wait.
2. **L — `now` undefined fell through to the delete.** Extracted `pinSectionAsShown(rows, shown)` (`pinNavigation.ts`): a renamed, recounted **or missing** section is a change, and the confirm re-prompts. RED: "the section as shown" (unchanged → true; renamed / recounted / missing → false).
3. **L — follower dropped a notification landing after its catalog refresh.** `followPinCatalog` now calls `reads.request()` unconditionally and its run is a no-op unless the catalog is still stale, so a change the read's own refresh (or a mutation's confirmation) absorbed costs nothing, a change landing after it queues one trailing read, and a hub that notifies on every read still cannot keep it reading. RED: "queues one trailing read for a change that landed after the read's refresh", "reads once per invalidation; a change the read's refresh absorbed costs nothing", "skips the read when the mutation's own confirmation already caught up".

Gates: `make test-native` green on this head (Metro export, 84 files / 800 native tests, 784 shared, tsc, check:scripts). No biome over `mobile-native`.

obra:
## Review round 3 disposition (81780d8bf)

CI: `race-root` failed on a PR that touches only `mobile-native/` (flake class; see the run for which Go test). RoboRev: one Medium, one Low, both accepted.

- **M `ProjectSessionsList` tests `stale` before `error` in both render branches, so a failed automatic re-read (which leaves `stale: true` with a non-null `error`, as the new tests pin) shows "Updating…" forever and the Retry action is unreachable** — accepted; a regression against the pre-diff behaviour. Error and its Retry render when `error` is set, with the stale status alongside or after; the sibling views already do this.
- **L `singleFlight` sets `running` before invoking `run()`, so a synchronous throw skips the `.finally` reset and wedges the coalescer for good** — accepted; invoke through `Promise.resolve().then(run)` (or try/catch) so the reset always runs, with a test.

Round 4 with the implementer.

obra:
## Round 4 fixes (11e41949c)

Merged origin/main (`63034d726`) with `--no-ff`.

1. **M — "Updating…" masked the retry after a failed re-read** (`ProjectSessionsList.tsx`). Added `pageStatus(page)` to `projectBrowser.ts`, which decides a boundary row's state once — `loading`, then `error`, then `stale`, then `more` — so an error keeps its Retry even while the page is stale, matching PageList, PinCatalogList, PinAssignmentEditor and PinSectionEditorScreen. The page-row model carries `status` and both the page row and the projects footer render from it (`browser.retry()` is reachable again in the `{ stale: true, error }` state your controller tests pin). RED at the render-model level: `projectBrowser.test.ts` "shows the error, with its retry, even while the page is stale" plus the loading / stale / more / settled cases.
2. **L — a synchronously throwing run left `singleFlight` marked running** (`singleFlight.ts`). The run is invoked inside try/catch: on a synchronous throw `running` is reset and the error is rethrown to the caller of `request()` (a synchronous throw from a job is a programming error and should stay visible), so later requests are not dropped. RED: `singleFlight.test.ts` "recovers from a run that throws synchronously".

Gates: `make test-native` green on this head (Metro export, 84 files / 806 native tests, 784 shared, tsc, check:scripts). No biome over `mobile-native`.

PR #1471, opened 2026-09-16
refactor(sdk): one tasks-panel store core for web and native (D3)

obra:
SDK migration phase D, row D3: one tasks-panel store core for the web panel and the native tasks sheet.

## What moved where

- **New root module `appwire-client/typescript/taskPanelState.ts`** (name avoids C25's `taskList*`): `createTasksPanelStore(listTasks)` returns the framework-free store triple (`getState` / `getInitialState` / `setState` / `subscribe`) plus two store-bound actions, `refresh(ref, hasAggregate)` and `watch(notifications, ref, threadId, hasAggregate)`. State keeps the web contract's synchronous entry actions (`beginFetch`, `publishFetch`, `setRows`, `evict`, `resetForTests`). Pure helpers exported beside it: `classifyTasksResponse`, `classifyTasksRejection`, `applyTasksFetchResult`, `panelLoadFailure`, `EMPTY_TASKS_PANEL_ENTRY`, and the `PanelLoadFailure` type `stores/activityPanel.ts` now takes from the package.
- **Ports.** The read is a `TasksListRead = (ref) => Promise<unknown>` rather than a `Pick<AppwireClient, "request">`: the web's read must keep going through `threadsStore.listTasks` (it waits out a reconnect, #195), and `stores/threads.ts` is off limits beyond an import swap, so the store cannot take a raw client on the web. `watch` takes `TasksPanelNotifications = Pick<AppwireClient, "onNotification">` (the `activityList.ts` idiom); `ConversationClientLike` satisfies it structurally, the package imports nothing from `mobile/src`.
- **Web adapter `stores/tasksPanel.ts`** keeps its path, `tasksPanelStore`, `useTasksPanelStore` and `resetTasksPanelStoreForTests`; eviction registration stays there. It reads through `threadsStore.getState().listTasks` lazily (threads.ts imports this module's reset; neither touches the other at evaluation time — the same shape as `humanNoteDrafts.ts`). `TasksPanel.tsx` now calls `tasksPanelStore.refresh` and toasts off the result; the classification it carried inline moved into the core.
- **Native twin deleted.** `mobile-native/src/taskList.ts` and its test are gone; `TasksSheet.tsx` builds one store for its lifetime (reading through the current client, so rows survive a reconnect) and `watch`es `evener/task/updated` / `evener/thread/resync` through the core. Its failure copy is now the shared "Couldn't load tasks" (was "Could not load tasks").
- **Behaviour folded from native into the shared core:** overlapping refreshes for one ref coalesce (dirty flag, one request in flight, the stale read's answer is never published) instead of the web's previous concurrent-fetch model; `TasksPanel.test.tsx`'s stale-overlap test was rewritten to that model, and the store test `stores/tasksPanel.test.ts` is unchanged as the adapter contract. `refresh` resolves the result to the run's latest caller (the one whose `hasAggregate` classified it) and null to earlier callers.

## Store triple

The triple is hand-rolled here because D30's `createFrameworkFreeStore` (#1461) is not on main yet; it has the identical shape and will be swapped in at this PR's refresh once D30 lands.

## Tests

- `taskPanelState.test.ts` (new, 11 cases, RED first): two-instance isolation, fake read + fake notification feed through the ports, coalescing, latest-caller result, the null-vs-[] and actionUnavailable / thread-not-found (with and without aggregate) classifications ported from the deleted native test, rows retained across failure and daemon-gone.
- Oracles kept: `stores/tasksPanel.test.ts` (unchanged), `panelStoreEviction.test.ts`, `sessionPanelPane.test.tsx`, `TasksPanel.test.tsx` (one test rewritten as above).

## Publication template

`tsconfig.build.json` files entry, alphabetical `index.ts` re-export, one real `rootSmokeCalls` smoke (creates a store, refreshes through a fake read, asserts the rows land; plus a `classifyTasksRejection` call), README summary. Falsified: deleting the `index.ts` re-export makes `make test-api-package` fail with `shipped module unreachable from every published specifier: taskPanelState`; restored with `git diff | git apply -R`.

## Gates (all foreground, all green at 7423c43b9)

`make test-api-package`, `make test-web`, `make test-native`, `make test-web-browser` (all six guards passed first run, skillguard included), `make lint-package-imports`, `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend`, `go test -count=1 -run TestScenarioSource .`.

## /simplify

Run before opening: `refresh` result now goes to the latest caller so `TasksPanel.tsx` dropped its `mountedRef`/`currentSessionRef` for a `live` flag; `tasksLoadFailure` generalised to `panelLoadFailure(failure, err)` (the same four lines `stores/activitySummary.ts:99` carries — D18's row to adopt); `evict` skips the notify when nothing is there; the entry update no longer copies the empty constant; the native sheet names `hasAggregate` and `refresh` once. Skipped: making `beginFetch`/`publishFetch` private (the web store test is the contract), reusing one fetch id per coalesced run (eviction mid-run would then leave a joined caller with no entry), evicting native entries per `sessionRef` (the sheet is keyed per hub:ref in `screens.tsx`).


obra:
## Refresh onto 3ba405d78 + createFrameworkFreeStore

- Merged `origin/main` (3ba405d78, after D30/D7/D2) with `--no-ff`; the only conflict was the plan status table, resolved as main's cells + the D30/D7/D2 rows + the D3 row. index.ts, tsconfig.build.json, README and the qualification smoke merged as unions automatically.
- Promised swap done in 21590316f: `TasksPanelStore` now `extends FrameworkFreeStore<TasksPanelState>`, `TasksPanelListener = StoreListener<TasksPanelState>`, the factory builds on `createFrameworkFreeStore` and the local triple is deleted. Exported shape unchanged (`refresh`/`watch` stay as store-bound extras). No hand-rolled triple remains in D3's files.
- `git diff origin/main...HEAD --stat` shows only D3's 14 files.
- Gates rerun in the foreground at 21590316f8e75d2c47fbfde408142c5b6fe69897: `make test-api-package` (plus the falsification: dropping the index.ts re-export fails with `shipped module unreachable from every published specifier: taskPanelState`, restored), `make test-web`, `make test-native` (exit 0), `make test-web-browser` (all six guards passed on the first run, skillguard included, so no rerun), `make lint-package-imports`, `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from the frontend dir, `go test -count=1 -run TestScenarioSource .`.

Issue #1470, opened 2026-09-16
askDock store: move the five state-held actions onto the store object (web call-site flip)


obra:
Follow-up from the D20 /simplify pass (altitude finding).

`appwire-client/typescript/askDock.ts` keeps five actions inside the state object (`setAnswer`, `setNote`, `setActive`, `markPendingGreeted`, `sendBatch`) while `reconcile`, `beginSend`, `finishSend` and `followThreads` are store-bound, the shape D30's `disclosure.ts` uses for every action. The split exists only because the web's `AskDock.tsx` reaches the five through `askDockStore.getState().X` at seven call sites (lines ~199, 210, 280, 282, 492, 511) and D20 was scoped to leave its importers untouched.

Fix: a web-only PR flips those seven sites to `askDockStore.X`, then the package drops the methods from `AskDockState` (state becomes `{ byRef, mintedBatches }`) and `updateRef` no longer needs to be a hoisted declaration. Do it before native adopts `sendBatch` (D21-D24) so the native side never binds to the hybrid.

PR #1475, opened 2026-09-16
refactor(sdk): ask-dock store core with explicit wiring in the AppWire package (D20)

obra:
SDK-migration phase D row D20 (`docs/superpowers/plans/2026-09-12-sdk-migration.md`).

## What moves

`cmd/evener-hub/frontend/src/panes/session/composer/askDock/askDockStore.ts` (445 lines, zustand, `threadsStore.subscribe` at module load) becomes the web adapter over a new package root module, `appwire-client/typescript/askDock.ts`. `mobile-native/src/questionBatches.ts` (60 lines) becomes a thin adapter over the same store.

## The core

`createAskDockStore({ send? })` returns the framework-free triple (`getState`/`getInitialState`/`setState`/`subscribe`) plus the store-bound `followThreads(threads)`, `reconcile(ref, live)`, `beginSend(ref, batchId)` and `finishSend(ref, batchId, accepted)`. The state keeps the five actions the web contract reads through `askDockStore.getState()` (`setAnswer`, `setNote`, `setActive`, `markPendingGreeted`, `sendBatch`) plus `byRef` and `mintedBatches` (the batch-id counter, in state so a reset to the initial state restarts it; `AskDock.test.tsx` asserts `ask-batch-1`/`ask-batch-2`).

Ports, both injected, nothing subscribed at module load:
- `send: AskAnswerSender = (ref, text) => Promise<void>` - a constructor option; `sendBatch` on a store built without one throws.
- `followThreads(threads: AskDockThreads)` - `threads` is anything whose subscribers see `{ threads: ReadonlyMap<string, ThreadModel> }` before and after each change; the web's `threadsStore` satisfies it as-is. Returns the disposer.

The brief named `wire(threads, send)`; the /simplify pass split it (altitude finding: two independent concerns, and a mutable sender set by a later call left a "not yet wired" window). The web adapter is two lines: `createAskDockStore({ send: (ref, text) => threadsStore.getState().send(ref, text) })` and `askDockStore.followThreads(threadsStore)`, both at module scope, so a session's first hydrate still populates batches before any dock mounts.

The root's `askShared`, `askAnswers`, `deriveAskQuestions` and `reconcileBatches` are reused, not duplicated: the core adds only what sat between them (answer drafts, exclusion memory, the visible tab, the greeting/epoch, the send).

## Importers

- Web: unchanged. The adapter keeps its path and every export (`askDockStore`, `useAskDockStore`, `useAskDockPending`, `nextUnansweredKey`, `resetAskDockStoreForTests`, the four types). Three non-test importers (`askDockPending.ts`, `AskQuestionCard.tsx` type-only, `AskDock.tsx`) and six tests (`askDockStore.test.ts`, `AskDock.test.tsx`, `AskQuestionCard.test.tsx`, `Composer.test.tsx`, `Composer.integration.test.tsx`, `Composer.liveCapabilities.test.tsx`) are untouched and stay the oracles.
- Native: `screens.tsx` unchanged (not even the import: the class name and path are kept). `questionBatches.ts` was not folded into the core: `screens.tsx` constructs `new QuestionBatches()` and drives `subscribe`/`getSnapshot`/`reconcile`/`begin`/`finish` with `useSyncExternalStore`, and the row forbids touching that file beyond an import swap, so the class stays as a 40-line adapter with no reconciliation logic of its own (`reconcile` -> `store.reconcile`, `begin` -> `beginSend` behind the sheet's own rendered-questions equality check, `finish` -> `finishSend`). It disappears when native gets a ThreadModel-backed store and a sender (D21-D24) and can call `sendBatch`.

## frameworkFreeStore

D30 (#1461) is not on main yet, so this branch carries `frameworkFreeStore.ts` and its test byte-identical to #1461's head (`dccfcdaa0`), with the same `tsconfig.build.json`, `index.ts` and qualification hunks in the same places. The refresh after D30 lands should resolve to D30's copy without conflict; if #1461 changes the module first, this branch swaps to it at refresh.

## Tests

- New `appwire-client/typescript/askDock.test.ts` (RED first): two instances with fake ports share nothing (a batch, an answer, a send and its exclusion in one are invisible to the other); `followThreads`' disposer stops reconciliation; a store without a sender rejects `sendBatch` loudly; a rejected send leaves the batch intact and retryable with the dock's sentence; the `reconcile`/`beginSend`/`finishSend` contract native's two old tests pinned; a no-change reconcile does not notify; `setState(getInitialState())` empties every ref and restarts the ids.
- Falsification: deleting the `askDock` re-exports from `index.ts` makes `make test-api-package` fail with `shipped module unreachable from every published specifier: askDock`; restored with `git diff | git apply -R`.

## Template

`tsconfig.build.json` `files`; `index.ts` re-exports (alphabetical, between `askAnswers` and `askShared`); `qualify-package.mjs` `rootSmokeCalls` builds a store, reconciles a live scan and freezes the batch; README module summary; module header; plan status row D20 (next commit).

`make test-api-package`, `make test-web`, `make test-native`, `make test-web-browser` (skillguard passed first run), `make lint-package-imports` (it reads comments: a header that named the package directory was reworded), `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend`, `go test -count=1 -run TestScenarioSource .`.

## /simplify

Applied: sender as constructor option + `followThreads` (above); the port's listener typed as the package's `StoreListener`; one `withBatchSending` for the freeze and the rejected-send thaw; answer records spread the prior entry with `answerFor` hoisted; the test's thread source is a `createFrameworkFreeStore` instance (also proves a plain store satisfies the port); duplicated wiring rationale trimmed from the package header and README; native `begin` compacted. Skipped: identity compare for native's stale check (behaviour change on prune-then-re-add), the port taking the bare map (adds a web shim for no gain), pruning port types from the index (a consumer typing a store needs them), routing `reconcile` through `updateRef` (one extra parameter for three lines). Follow-up filed: #1470 (flip `AskDock.tsx`'s seven `getState().X` sites so every action can be store-bound like D30's).


obra:
## Refresh onto 0f9f10816

`git merge --no-ff origin/main` (D3 #1471 head). `frameworkFreeStore.ts` and its test resolved to main's copy with no conflict, so `askDock.ts` now builds on the one `createFrameworkFreeStore` from main; `git diff origin/main...HEAD --stat` shows only D20's nine files. Three union resolutions: `index.ts` (main's `hubOverview` block; `askDock` stays between `askAnswers` and `askShared`), `qualify-package.mjs` (main's disclosure smoke line beside the askDock one), and the plan status table (main's merged cells for C20/C4/C3 plus the D30/D7/D2/D3 rows, D20 appended).

Gates rerun in the foreground after the merge, all green: `make test-api-package` plus the falsification (deleting the `askDock` re-export fails with `shipped module unreachable from every published specifier: askDock`, restored), `make test-web`, `make test-native`, `make test-web-browser` (skillguard passed on the first run, no rerun needed), `make lint-package-imports`, `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend`, `go test -count=1 -run TestScenarioSource .`.

PR #1476, opened 2026-09-16
refactor(sdk): marketplaces store core at @evener/appwire-client/state/extensions (D10)

obra:
SDK migration phase D, row D10: the marketplaces slice of the web extensions store becomes a framework-free store in the package, published under the new subpath `@evener/appwire-client/state/extensions` (a barrel D11 and D12 will extend with the plugins and directories stores).

## What moves

- **Core** `appwire-client/typescript/state/extensions/marketplaces.ts`: `createMarketplacesStore(client: MarketplacesClient)` returns the D30 `FrameworkFreeStore` triple (`getState`/`getInitialState`/`setState`/`subscribe`) plus `start()`/`reset()`/`dispose()`; state holds `marketplaces`, `marketplacesLoading`, `marketplacesError`, `browseCatalogs` and the store-bound actions `fetchMarketplaces`, `addMarketplace`, `removeMarketplace`, `refreshMarketplace`, `editMarketplace`, `browseMarketplace`, `reloadCatalog`. `MarketplacesClient = Pick<AppwireClient, "request" | "onNotification">`. Zero runtime deps; no React, zustand, mobile/src or web-app import.
- **Semantics reconciled and pinned**: fetches (`fetchMarketplaces`, `browseMarketplace`, `reloadCatalog`) never throw and record their failure in state; mutations reject. The web's generation-fenced catalog retirement, revision-fenced list responses and the 250 ms debounced `evener/marketplace/updated` refetch are the core's behaviour for both apps. `reloadCatalog` is new: native's retry-after-error and pull-to-refresh need "drop this cache entry and browse again" without a hub-side source refresh.
- **Web adapter** `cmd/evener-hub/frontend/src/stores/extensions.ts` stays one zustand store to its 12 importers (+17 tests, none changed): the package store is the app's instance over a connection-backed client port and publishes the fields each change touched into `extensionsStore`; the plugins and launch-layer slices stay in place for D11/D12. The connection wiring becomes `onConnectionNotification` in `stores/connection.ts` (exported, disposer-returning, detaches a replaced client) for later rows to adopt.
- **Native twin deleted**: `mobile-native/src/marketplaces.ts` and its test. `MarketplaceBrowser.tsx` binds to the core with `useSyncExternalStore`, keeps the selected marketplace as component state, reads its catalog from the store cache (re-requesting a retired slot, clearing a selection the list no longer has) and translates errors to its copy at render.

## Publication (mirrors C1's `state/navigation`)

`package.json` exports, `tsconfig.build.json` files, `qualify-package.mjs` `./state/extensions` manifest with real smoke calls (offline client: fetch records, mutation rejects, browse caches the error), README "Published subpaths", and one alias in each resolver (Vite, web tsconfig, native tsconfig + tsconfig.check, native vitest; Metro probes `<subpath>/index` generically). Falsified: removing the barrel re-export fails `make test-api-package` with `shipped module unreachable from every published specifier: state/extensions/marketplaces`.

`frameworkFreeStore.ts` (+ its test) is carried byte-identical from D30 (#1461), which is not on main yet; it is a clean swap at refresh.

## Tests

- New `state/extensions/marketplaces.test.ts` (18 tests): two-instance isolation, both failure conventions, browse cache/in-flight wait/retirement fencing, `reloadCatalog`, list ordering, notification debounce, `dispose`, `reset`.
- Oracles unchanged: `stores/extensions.test.ts` (61) and the seven `marketplacesPlugins` section suites.

`make test-api-package`, `make test-web`, `make test-native` (Metro bundle resolves the new subpath), `make test-web-browser` (all six guards, first run), `make lint-package-imports`, `npm run lint` + `npx biome ci ../../../appwire-client/typescript` from the frontend dir, root `go test -run 'TestScenarioSource|TestPackageImportPaths|TestEvery…|TestMakeLint…' .` (make/linting.mk doc line touched). `/simplify` run: mutate()/setCatalog() folds, changed-fields mirror (whole-snapshot mirroring overwrote seeded fixtures in 17 section tests), connection helper relocation, native select handler.

## Behaviour notes for review

- Native's refetch after a *failed* mutation is gone: the core does not refetch on rejection (the web never did); a mutation that landed without a reply is reconciled by the hub's broadcast or the reconnect remount, and the inline copy already tells the user to refresh.
- Native's notification-triggered refresh is now debounced 250 ms like the web's.


obra:
## Refresh onto c010784df

Merged `origin/main` (D20 #1475 head, after D30/D7/D2/D3) with `git merge --no-ff`; head is now `ebb19073d9ca95fcaac845064c2f200aaf758716`. `frameworkFreeStore.ts` resolved to the one module from main (zero diff, no second triple); `index.ts` is identical to main; unions in `qualify-package.mjs`, README and the plan status table (main's cells + D30/D7/D2/D3/D20 rows + D10). `git diff origin/main...HEAD --stat` shows only D10's 19 files.

Gates rerun in the foreground on this head: `make test-api-package` (plus the barrel falsification, exact message reproduced and restored), `make test-web`, `make test-native` (Metro exported the iOS bundle through the new subpath), `make test-web-browser` (all six guards, first run, no skillguard rerun needed), `make lint-package-imports`, frontend `npm run lint` + `npx biome ci ../../../appwire-client/typescript`, root `go test -run 'TestScenarioSource|TestPackageImportPaths'` and `go test -short .` (Makefile audits).

obra:
## Review round 1 disposition (ebb19073d)

CI: `lint-repository` (and the dependent `static`) FAILED — lines being pulled. RoboRev: one Medium, two Low; all accepted.

- **M `resetExtensionsStoreForTests` now relies on the core's reset publish, but the changed-fields mirror publishes nothing when the core is already at its initial `Map`, so a test that seeded `browseCatalogs` directly into the web store keeps it across tests (`PluginDetailSheet.test.tsx:338/:352`)** — accepted; the explicit marketplaces fields go back into the web reset (the mirror stays for normal operation). A required-gate test regressing is exactly what the oracle exists to catch.
- **L native `MarketplaceBrowser` browses a marketplace that is being removed/renamed because the browse effect runs before the selection-clearing effect** — accepted; gate the browse on the name still being in the list (or reconcile selection first, in one effect).
- **L `dispose()` does not fence the catalog-retirement `set` in `mutate`, so a late mutation notifies a disposed store** — accepted; the disposed flag gates every post-await write.

Round 2 with the implementer.

obra:
## Round 2 fixes

Head `eb68f4d498695e39c69fc05873f3a6c099a0d12d`, on `origin/main` = `b319dd7ce` (merged `--no-ff`). Each fix landed RED first.

1. **M — web reset vs changed-fields mirror.** `resetExtensionsStoreForTests` writes `marketplaces: null`, `marketplacesLoading: false`, `marketplacesError: null`, `browseCatalogs: new Map()` explicitly again (comment explains why beside the mirror). RED: new `extensions.test.ts` case seeding those fields straight into the store, then reset — failed and poisoned the next test, exactly the leak described; the PluginDetailSheet pair stays green.
2. **L — spurious browse of a removed/renamed marketplace.** New pure decision `mobile-native/src/marketplaceBrowserModel.ts` `catalogToBrowse(selected, marketplaces, browseCatalogs)`: browses only a selection the loaded list still carries and the cache lacks; `MarketplaceBrowser.tsx` runs its effect off that value. RED at the render-model level: `marketplaceBrowserModel.test.ts` (4 cases incl. the dropped-selection and unloaded-list cases).
3. **L — mutate's post-await write not fenced by dispose.** Every core write goes through a dispose-aware `set` (`if (!disposed) publish(partial)`). RED: `marketplaces.test.ts` "a mutation that resolves after dispose() publishes nothing" — subscriber count was 1, state identity changed.

**CI lint-repository:** `make generate` regenerated `docs/developing-evener/linting.md` for the `lint-package-imports` summary line I edited; committed on its own (`a5fe420a3`). `make lint-generated` PASS and root `go test -short .` (Makefile audits) ok.

Gates rerun in the foreground on this head: `make test-api-package` (+ barrel falsification, exact message, restored), `make test-web`, `make test-native` (Metro bundle exported; 782 + 784 tests), `make test-web-browser` (six guards, first run), `make lint-package-imports`, frontend `npm run lint` + `npx biome ci ../../../appwire-client/typescript`, `go test -run 'TestScenarioSource|TestPackageImportPaths' .`.

PR #1478, opened 2026-09-16
refactor(sdk): one command catalog core for web and native (D1)

obra:
SDK migration phase D, row D1 (`docs/superpowers/plans/2026-09-12-sdk-migration.md`). Builds on D30's `frameworkFreeStore.ts` (#1461, on main).

## Twin finding: one catalog, two scopes

Read both files fully before designing.

- `cmd/evener-hub/frontend/src/stores/commandCatalog.ts` (26 lines) fetched `evener/command/list` once per palette open and held the raw, hub-wide list. Its consumers did the per-session work themselves: `CommandPalette.tsx:259-267` and `Composer.tsx:288-297` each derived the session's plugin-name set from `ThreadModel.diagnostics.plugins` (already in `threadsStore`) and ran `visibleCatalogCommands` + `mergeSlashCommands(..., model.skills)`.
- `mobile-native/src/commandCatalog.ts` (105 lines) fetched the same `evener/command/list` **and** `thread/read {includeTurns:false}` per ref, then ran the same `visibleCatalogCommands` + `mergeSlashCommands` over `evener.diagnostics.plugins/skills`, with coalesced refresh, last-good retention, a foreign-ref guard and re-reads on `evener/plugin/updated` / ref-matched `evener/thread/resync`.

Same catalog, same projection. The only genuine difference is where the session diagnostics come from: the web's thread model has them; native's does not (`mobile/src/conversation/model.ts` carries no diagnostics until D21–D24), so native reads them through `thread/read`.

**The palette is deliberately not rewired to a self-fetching per-ref catalog**, contrary to the plan row's "the palette needs rewiring" note: (1) `CommandPalette.test.tsx` "the palette scopes catalog commands to active diagnostics without mutating the global catalog" asserts the filter follows `threadsStore` changes reactively and the store keeps the unfiltered list; (2) the palette needs the hub-wide list with no session focused (`sessionScopedHandoffMatch` over user commands); (3) a `thread/read` per palette open would duplicate the source of truth the threads store already maintains. What the web *does* gain from native's logic: coalesced refresh, last-good retention with a named error, and the re-read on `evener/plugin/updated` (the web catalog went stale until the next palette open).

## Core (`appwire-client/typescript/commandCatalog.ts`, root export)

- `type CommandCatalogClient = Pick<AppwireClient, "request" | "onNotification">` (the `activityList.ts` idiom; the package imports nothing from either app).
- `createCommandCatalog(client): CommandCatalog` — hub-wide store: state `{ commands, loaded, loading, error, refresh }` on the D30 triple, plus `watch(): () => void` (re-reads a *loaded* catalog on `evener/plugin/updated`).
- `createSessionCommandCatalog(client, ref): SessionCommandCatalog` — one session's slash menu: state `{ items, loading, error }`, methods `refresh/start/dispose`; the atomic catalog + `thread/read` pair, foreign-ref guard, resync/plugin re-reads. Native's superset behaviour, verbatim.
- `sessionPluginNames(diagnostics)` — the plugin-inventory filter the two web components and the native twin each derived inline.
- Private `refreshLoop(set, read, failure)`: the coalescing/superseded/disposed rule both factories share.

## Adapters

- **Web** `stores/commandCatalog.ts` keeps its path and `useCommandCatalog`'s surface (hook with selector + `getState/setState/getInitialState/subscribe`), as zustand's `useStore` over the package store. Its client port forwards each request to the connection store's current client and moves the plugin-change subscription when that client is replaced. `paletteController`, `commands.ts`, `CommandPalette.tsx`, `Composer.tsx` keep their imports; the palette and composer swap their inline plugin-set derivation for `sessionPluginNames`. All palette/composer tests unchanged (307 pass) and stay the contract.
- **Native** `mobile-native/src/commandCatalog.ts` + test **deleted**; `CommandCompletion.tsx` calls `createSessionCommandCatalog` and types its prop as `CommandCatalogClient`, dropping its `mobile/src` import. The three native cases are the session catalog's oracles in the package test.

One deliberate web behaviour change: a failed re-read now keeps the last catalog and records `error` (native's retention rule) instead of clearing to `[]`; `stores/commandCatalog.test.ts` is updated for that and gains the client-following and no-client cases.

## Template
`tsconfig.build.json` files entry; `index.ts` re-export; `qualify-package.mjs` smoke (`createCommandCatalog` refresh round-trip + `sessionPluginNames`); README clause; plan status row.
Falsified: removing the `index.ts` re-export makes `make test-api-package` fail `shipped module unreachable from every published specifier: commandCatalog`; restored.

`make test-api-package`, `make test-web`, `make test-native`, `make test-web-browser` (all six guards green first run), `make lint-package-imports`, `npm run lint`, `npx biome ci ../../../appwire-client/typescript`, `go test -count=1 -run TestScenarioSource .` — all pass.

## /simplify
Applied: refresh loop absorbs the read/error contract (both factories collapse, session pair via `Promise.all`); plugin/updated re-read gated on `loaded` so an idle tab issues nothing; web port drops a derivable variable; brittle notification-count assertion replaced. Skipped: merging `CommandCatalogClient` with `ActivityClient` (per-module Pick is the briefed idiom); replacing the package test's hand-rolled double with `FakeClient` (exercises the module's own Pick, `activityList.test.ts` precedent).

## Carry-forward for later rows
- D2: lift the connection-following client port to `stores/connectionClient.ts` when the second consumer arrives.
- D3: lift `refreshLoop` into its own package module and retrofit `taskList` (ActivityList waits for D17).
- Follow-up after merge: drop the unread `loaded` flag from `CommandCatalogState` (kept because the web store's surface is the contract; ~30 test `setState` sites).


obra:
## Round 1 fixes

Head `2300f5895c155f26d89a4f90befd564c468d2790` (one commit on top of `b40eb4e87`, no force).

1. **`commandCatalog.ts` refresh loop opens with `error: null`** (as the deleted native class did), so `CommandCompletion.tsx`'s error banner and Retry button leave while the retry is in flight instead of staying rendered through it. RED test first: `commandCatalog.test.ts` "a retry after a failure clears the error for as long as it loads" failed with `error: 'Could not load commands and skills: offline'` still set while `loading: true`; green after the one-line change. The hub-wide catalog gets the same rule through the shared loop.
2. **`qualify-package.mjs` smoke probe no longer floats.** The generated consumer runs as both `.mjs` and `.cjs` (`${presenceLoop}${surface.smoke}` into each), so a top-level `await` would break the CJS one. Instead the probe asserts the synchronous `loading === true` transition right after `refresh()` (the shape the other probes have) and routes the read's continuation through `.catch` → `console.error` + `process.exit(1)`, which `execFileSync` in the runner turns into a gate failure.

Gates rerun in the foreground on this head: `make test-api-package`, `make test-native`, `make test-web` — all pass.

obra:
## Refresh onto 55bce119a

`git merge --no-ff origin/main` (D10 #1476 head; no rebase, no force). Head `5a2be2b3f9ca2c41b71894d7dbf9e353a2eedf09`. One conflict, the plan status table, resolved as a union: main's D7/D2/D3/D20/D10 rows then D1's. `index.ts`, `tsconfig.build.json`, `README.md` and `qualify-package.mjs` auto-merged with D1's entries intact (its probe stays synchronous-shaped). `git diff origin/main...HEAD --stat` shows only D1's 14 files. The core already built on main's `frameworkFreeStore.ts`, so nothing to swap.

Gates rerun in the foreground on this head: `make test-api-package` (+ the re-export falsification, still fails `unreachable from every published specifier: commandCatalog`), `make test-web`, `make test-native`, `make test-web-browser` (all six guards green first run), `make lint-package-imports`, `npm run lint`, `npx biome ci ../../../appwire-client/typescript`, `go test -count=1 -run TestScenarioSource .` — all pass. `gh pr view --json mergeable`: MERGEABLE.

PR #1480, opened 2026-09-16
transcript,agent: the fold names its retained entries by Seq (piece A2 of the #1100 split, closes #1200)

### LINKED ISSUE #1200: agent: a compaction fold's retained tail is not replayed on restart; resumed history is the summary only

obra:
Piece A2 of the #1100 split (rebuilt on the transcript's durable `Entry.Seq` after #1281 was closed). **Closes #1200.**

## The bug (#1200)

After a fold, live history keeps the retained tail verbatim, but a restart resumed only the fold's SUMMARY. `ResumeHistory` anchored on the last CHECKPOINT/SUMMARY marker and discarded every entry before it, while the retained tail's entries sit *before* that marker. The old machinery (`persistedAppendLog` → `rewriteTail`) re-appended only the turns recorded *during* the fold, after the markers — never the preserved recent tail — so the tail vanished on restart.

## The fix

The fold now writes ONE durable `FOLD_RECORD` entry — a manifest naming, by transcript `Seq`, exactly the entries its published history is made of:

- `Layers`: the head marker's Seq (the summary, or the checkpoint when summarization did not run).
- `RetainedSeqs`: the pre-existing turns the fold kept (preserved tail + during-fold merge-back + injected steering), in history order.

The record is written last, after the markers and steering, so everything after it is live post-fold history. On restart, `ResumeHistory` finds the last fold record and rebuilds `[Layers entries] ++ [RetainedSeqs entries] ++ [entries after the record]`, each looked up by Seq. Because `recordTurn` already writes the *persisted* form as the entry and stamps the live turn's Seq to it, naming an entry by Seq yields the identical persisted content the old `rewriteTail` re-appended — without any duplicate entry.

Deleted wholesale: `persistedAppendLog`, `logPairPersistedLocked`, `persistedAppendLogBase`, `rewriteTail` (the fold's `snapAppends` snapshot bookkeeping goes with them).

New/changed:
- `schema.Turn.Seq int` (`json:"-"`, in-memory only): the durable per-line id, stamped at append or seeded from the entry on restore.
- `schema.FoldRecord` + `schema.TurnFoldRecord`; `Turn.Fold *FoldRecord` (`json:"fold_record,omitempty"`).
- `transcript.Writer` `Append`/`AppendDurable`/`AppendSynced` and the session write helpers now return the spent Seq.
- `ResumeHistory`/`resumeHistoryIndexed` reconstruct from the fold record; a pre-A2 transcript (marker, no record) falls back to the legacy last-marker anchor so an in-flight session's first restart after the upgrade does not suddenly replay its long-discarded prefix.
- Fork-provenance in `session_init.go` is decided per turn: `inheritedProvenanceTurns` maps `DivergenceTurn` through each resumed turn's own origin entry (and through the repair insertions), because a fold interleaves provenance — its own head marker, the parent turns it retained, its own tail — which no single boundary index can express. Byte-identical to the old contiguous-window math for the non-fold-record paths.

The `FOLD_RECORD` is never added to `Session.history`: `expandHistory` never sees it, and the reload projector emits no item for it (`default` case), so no projector shape change.

## Old-reader break — option 1, MEASURED (needs Jesse's ruling before merge)

Built under option 1 (no backward-compat machinery). `Turn.Fold` is a new wire field, so `decodeStrictJSON`'s `DisallowUnknownFields` (kata wf7e) makes a transcript that carries a fold record **unreadable to a pre-A2 daemon**. Measured with the exact strict decode against a `schema.Turn` that lacks the field:

```
A2 fold-record entry on the wire:
{"kind":"entry","seq":7,"turn":{"kind":"FOLD_RECORD","message":{"role":"","content":null},"timestamp":"0001-01-01T00:00:00Z","usage":{"input_tokens":0,"output_tokens":0,"total_tokens":0},"fold_record":{"fold_id":"3","layers":[5],"retained_seqs":[2,3,4]}}}

pre-A2 reader rejection: json: unknown field "fold_record"
```

Every reader (thread/read, resume, fork, doctor) decodes the whole transcript in one pass and aborts on the first error, so ONE fold-record entry makes the **entire transcript unreadable** to a pre-A2 build — not just that turn. It is clean (a decode error, no crash) and self-healing: the same daemon restarted on this build declares the field and reads the file fine. This is the same one-way door every prior `schema.Turn` field addition opened. Option 2 (an ignore-list format) cannot fix daemons already shipped — they do not know the field either — so it would only help future readers, which option 1 already covers by restart. **Holding for Jesse's explicit ruling on the compat break.**

## Live-Seq threading — the count

The returned-Seq stamp threads through **3 write choke points** (`writeTranscriptLocked` / `writeTranscriptDurableLocked` / `writeTranscriptSyncedLocked`) and the history-append helpers that already hold the Seq in hand (`appendTurnAfterTranscriptWrite(Locked)`, `recordTurn`, `appendPairedTurnVia`, `attachTranscript`, `appendClientMutationTranscriptLocked`, `appendUserInputTurnRefusingPoison`, the skill-reload reminder append). The stamp lives only in those helpers; the rest is the mechanical `(int, error)` signature change fanning out to call sites: **16 production files** and **~59 test files**, uniform and behavior-preserving.

## Gates

gofmt (toolchain binary), `go vet` ×3 both modules (incl. `GOOS=windows -tags evenerfuzz`), golangci-lint (0 issues, both modules), transcript + full agent suites, fold/resume/compaction suites, `-race -count=3` on the fold/resume/compaction/durability suites, apptranscript + appprojector + TUI suites. `/simplify` run (one dedup: an `appendEntry` closure in `resumeTurns`). No trailers.


obra:
## Review round 1 disposition (1d68d7bdc)

CI 15/15 green. RoboRev: two Medium, one Low (one reviewer ran). All accepted; the two Mediums are the same class — a Seq is being read as "durable entry" when it may not be one.

- **M1 failed writes hand back an unspent seq that the fold and `recordTurn` record as if durable; the fold record can then name its own future seq (leaking `FOLD_RECORD` into resumed history) or another turn's entry** — accepted. Capture a seq only when the append recorded (`err == nil` or `ErrRetainedUnsynced`), otherwise a sentinel that `writeFoldRecordLocked` omits; return early when the head marker's write errored; `resumeTurns`' `add` skips `TurnFoldRecord` and any sentinel, like the post-record loop.
- **M2 `Seq == 0` is both a valid seq and the value of every history turn without an entry, so the positional steering fallback can hand a steering seq to a no-entry turn (orphan-repair synthetics, strategy injections) and lose or duplicate the real steering turn** — accepted. Track "has durable entry" explicitly (a negative sentinel, or match steering turns by identity/order rather than by the zero test) and make the code do what the comment at `:883-886` already claims (skip the no-entry turn).
- **L `FOLD_RECORD` renders in `read_transcript` markdown** — accepted; silent no-op case beside `TurnAttentionResolution`.

Round 2 with the implementer. The merge stays held for Jesse's ruling on the pre-A2-reader break regardless.

obra:
## Round 2 fixes (3d044386f)

Merged origin/main (`4f64b223b`, `--no-ff`) first. All three findings addressed, TDD (each test fails against round 1's logic).

The root of both Mediums: a `Seq` was trusted as "this turn has a durable entry at Seq" when it did not. Fixed by making "no durable entry" explicit with a negative sentinel `schema.NoTranscriptEntrySeq`, and by capturing a Seq only when the write actually recorded.

**M1 — a failed write's unspent Seq was captured/stamped and could be named.**
- New `recordedSeq(seq, err)` (`agent/session.go`) returns the Seq only when the write recorded (`err == nil` or `errors.Is(err, ErrRetainedUnsynced)`); otherwise `NoTranscriptEntrySeq`. `appendBatchLocked` returns the UNSPENT next seq on a hard failure, so the old capture stamped a number the next successful append — usually the fold record itself — reused.
- Applied at every direct capture: the marker loop and `writeSteeringTurnRecordsLocked` (`session_compaction.go`) and `recordTurn` (`session.go`). The pair helper already skips the append on a hard error, so it was never wrong.
- `writeFoldRecordLocked` returns early (writes NO record) when the head marker's Seq is `< 0` (its write recorded nothing) — resume then falls back to the pre-fold history rather than naming a phantom entry.
- Defense in depth: `resumeTurns`' seq lookup now skips any entry whose kind is `TurnFoldRecord`, so even a seq that pointed at the record can never inject the record into history.
- Tests: `TestRecordedSeq_OnlyKeepsRecordedWrites`, `TestResumeHistoryFoldRecord_NeverInjectsAFoldRecord`.

**M2 — `Seq == 0` used as "no entry" is both a valid seq and the value of every un-stamped turn.**
- Repair synthetics (`syntheticToolResultsTurn`, `history_repair.go`) and strategy injections (the single choke point `replaceSteeringMarkerTurn`, `context_manager.go`, covering memory-crystals / recursive-distill / ooda) now carry `NoTranscriptEntrySeq`.
- `writeFoldRecordLocked`'s kept-tail classification is now explicit: `Seq > 0` names the durable entry; `NoTranscriptEntrySeq` is omitted; `Seq == 0` is a fold-injected steering turn whose durable Seq comes from the steering-write results, consumed in order (a failed steering write recorded nothing and is omitted). So a no-entry turn ahead of a steering turn no longer steals its Seq.
- Tests: `TestSyntheticToolResultsTurn_HasNoDurableEntrySeq`, `TestWriteFoldRecordLocked_NoEntryTurnDoesNotStealSteeringSeq`.

**L — read_transcript printed a FOLD_RECORD heading/omitted-note.**
- Added a silent no-op case for `TurnFoldRecord` in `writeEntry` (`transcript_render.go`), alongside `TurnAttentionResolution`. Test: the `FOLD_RECORD turn renders nothing` subtest in `TestRenderMarkdown_UnknownAndSystemTurns`.

Gates: toolchain gofmt; `go vet` ×3 both modules incl. `GOOS=windows -tags evenerfuzz`; golangci-lint 0 issues both modules; full agent + transcript suites; fold/resume/compaction/orphan suites; `-race -count=3` on the fold/resume/compaction/orphan suites; apptranscript/appprojector/TUI. No trailers. The merge stays held for Jesse's compat ruling.

obra:
## Review round 2 disposition (3d044386f)

CI 15/15 green. RoboRev: two Medium, one Low — the same class as round 1 a third time (a history turn carrying a `Seq` that is not its durable entry's). All accepted, and round 3 closes the class rather than the three instances.

- **M `appendUserInputTurnRefusingPoison` stamps `turn.Seq = seq` unconditionally after a write that may have failed with an unspent seq** — accepted; it must go through `recordedSeq` like `recordTurn`.
- **M live delegate-attention steering turns are never stamped (`AppendSynced`'s seq discarded, `retainDelegateAttentionTurn` appends Seq 0), so a fold's merge-back names them as 0 and restart resolves 0 to the session's first entry, resurrecting a summarized turn** — accepted; the most serious of the three.
- **L the delegate inherited-context boundary turn is queued with Seq 0 and never stamped by `attachTranscript`, so a fold misclassifies it as injected steering and drops it on restart** — accepted.

Class fix: enumerate every site that appends to `s.history` (and every pending-turn path that reaches it), route each through one stamping helper fed by the write's `(seq, err)`, and add an invariant test that after each append path no history turn carries `Seq == 0` unless it is fold-injected steering — or better, give injected steering its own explicit marker so `0` is never load-bearing. The merge stays held for Jesse's compat ruling regardless.

obra:
## Round 3 fixes (5ae8e1615)

Merged origin/main (`4f64b223b`, `--no-ff`) first. This closes the class instead of the instances: the class is "a history turn carrying a Seq that is not its durable entry's," caused by scattered hand-stamping plus a load-bearing `Seq == 0`.

### The rule (b + c)

There is now exactly one way a history turn gets a durable Seq, and one meaning for every value:

- **`recordedSeq(seq, err)`** is the sole stamp: it returns the write's Seq only when the write recorded (`nil` / `ErrRetainedUnsynced`), else `schema.NoTranscriptEntrySeq`. A failed write's UNSPENT next seq is never stamped.
- **Value space, all disjoint** so `Seq == 0` is never load-bearing: `>= 0` = a real durable entry; `NoTranscriptEntrySeq (-1)` = a turn that will never have an entry (repair synthetic, strategy injection, failed write, fork's inherited prefix); `seqHeldPreAttach (-2)` = written before attach, stamped at flush; `seqFoldInjectedSteering (-3)` = a fold's injected steering, resolved to its real Seq the moment its commit writes it.
- `writeFoldRecordLocked` classifies purely by these markers (no `Seq == 0` branch); `stampFoldWrittenSeqsLocked` resolves the live copies of the fold's head marker and injected steering after commit, so a later fold never re-misclassifies them; `attachTranscript` finds held turns by the `seqHeldPreAttach` marker, not `held[i]==history[i]`.

### Every site that puts a turn in s.history (a)

| Site | How the Seq is set |
|---|---|
| `recordTurn` (session.go) | `recordedSeq(seq, err)` |
| `appendTurnAfterTranscriptWrite` closures — session.go (env tracker, `appendPairedTurnVia`, tool-ask), session_queue.go ×3, session_tools.go ×2, session_skill_delivery.go, session_skill_reload.go, session_client_mutation_queue.go ×2, delegate_tree_steer.go, session_lifecycle.go steering | closure receives the recorded seq (the helper calls it ONLY on a recorded write); `turn.Seq = seq` is safe by that contract |
| `appendUserInputTurnRefusingPoison` (session_lifecycle.go) | **fixed:** `recordedSeq(seq, writeErr)` (was unconditional) — instance 1 |
| attention `AppendSynced` → `retainDelegateAttentionTurn` (session_attention.go) | **fixed:** stamp `recordedSeq(seq, err)` (seq was discarded) — instance 2 |
| `foldDelegateAttention` decoded turns (session_attention.go) | **fixed:** seed `turn.Seq = entry.Seq` — instance 2 |
| fork boundary turn (session_init.go) | **fixed:** marked `seqHeldPreAttach`, stamped by `attachTranscript` — instance 3 |
| fork inherited prefix (session_init.go) | **fixed:** marked `NoTranscriptEntrySeq` (its Seqs name the parent's transcript) |
| `attachTranscript` flush (session.go) | **fixed:** scan for `seqHeldPreAttach`, stamp `recordedSeq` |
| fold markers (commit, session_compaction.go) | `recordedSeq(markerSeqs)`; live copy stamped by `stampFoldWrittenSeqsLocked` |
| fold injected steering (`appendSteeringMessagesToHistory`) | `seqFoldInjectedSteering` marker, resolved by `stampFoldWrittenSeqsLocked` |
| fold merge-back | already through `recordTurn` |
| orphan-repair synthetic (`syntheticToolResultsTurn`) | `NoTranscriptEntrySeq` |
| strategy injections (`replaceSteeringMarkerTurn`, all 3 strategies) | `NoTranscriptEntrySeq` |
| resume seeding (transcript_read.go) | `entry.Seq` |

### Tests (c + d)

- **Invariant / class test** `TestFold_HistoryTurnSeqsNameTheirEntries` (+ `assertHistorySeqsNameEntries`): runs a real fold that injects a note-handoff steering turn, then asserts every live history turn names its own durable entry (or is an explicit no-entry marker), no two entry-backed turns share a Seq, and none is left carrying the transient `seqFoldInjectedSteering`.
- **Instance 1** `TestAppendUserInputRefusingPoison_MarksFailedWriteNoEntry` — failed-but-usable write ⇒ the turn carries `NoTranscriptEntrySeq`, not the unspent seq.
- **Instance 2** `TestFoldDelegateAttention_SeedsTurnSeq` — a decoded attention turn carries its entry's Seq, not 0 (which resume would resolve to the first entry).
- **Instance 3** `TestAttachTranscript_StampsHeldTurnAfterInheritedPrefix` — the held boundary after an inherited prefix is stamped at its real position, not index 0.
- `seedNumberedSessionHistory` (test helper) now marks its entry-less turns `NoTranscriptEntrySeq` — it never wrote them, and the invariant forbids a bare 0.

Gates: toolchain gofmt; `go vet` ×3 both modules incl. `GOOS=windows -tags evenerfuzz`; golangci-lint 0 issues both modules; full agent + transcript suites; fold/resume/compaction/orphan/attention/steering/fork suites; `-race -count=3` on those suites; apptranscript/appprojector/TUI. No trailers. Merge stays held for Jesse's compat ruling.

obra:
## Review round 3 disposition (5ae8e1615)

CI: 13 green, `tests` FAILED (lines being pulled; round 4 covers it). RoboRev: one Medium, two Low; all accepted. The class from rounds 1–3 is closed; these are new.

- **M `writeFoldRecordLocked` discards the fold record's write outcome (`_, _ =`)** — accepted, and it is the most important finding of the series: the record is the sole mechanism that makes a retained tail survive restart, and a hard failure (clean rollback, nothing recorded) is neither queued nor emitted, so the next restart silently falls back to the legacy anchor and drops the tail — the #1200 bug with no warning. The write returns its error and it is reported with the transaction's other write errors (emitted outside `attentionMu`).
- **L the `writeFoldRecordLocked` doc comment is attached to `headMarkerSeq`** — accepted.
- **L the `testOnly.appendCompactionTurn` seam yields a real-looking `Seq 0` for an intercepted steering write, violating the invariant the record relies on** — accepted; the seam returns a no-entry sentinel like a failed write.

Round 4 with the implementer. The merge stays held for Jesse's compat ruling.

obra:
## Round 4 fixes (2d9f9c1d3)

Merged origin/main (`8a85220f2`, `--no-ff`) first.

### CI failure root cause — a pre-existing flake, not an A2 regression

`TestDelegateResourceSupervision_CommittedSendStartRefusesASecondTurn` (root `tests` job, shard agent:2) got `Status:failed Reason:construction_failed Output:target_busy` where it wanted `started`. `target_busy` comes from `startOrSteerSubagentRun` (subagents.go) when the target delegate is still `finalizing`.

A2 touches none of the warm/send/commit/finalize path — its only delegate-adjacent edits are the attention `AppendSynced` Seq capture and a steering-durable closure signature, both timing-neutral and not exercised by this test (no attention arming in the warm leg). The real cause: `warmStableSupervisionDelegate` acknowledged the warm result (`commit.Complete` + `executeDelegateMutationPlans`) but returned before the warm run's own async finalization settled, so the immediately-following send raced the tail and saw the delegate finalizing. Under CI parallel load that tail loses; locally it never reproduced (20×, `-race` 5×, `GOMAXPROCS` 1/2/4). Fixed by waiting for the warm delegate to quiesce (`!running && !driving && !finalizing`) before returning — the same quiescence-wait pattern the suite already uses (`waitForStableSupervisionRun`), benefiting all 18 callers. The full root-module suite is now in my gate list.

### M (the series' key finding) — the fold record's write outcome was discarded

`writeFoldRecordLocked` now returns the record's hard write failure (a clean rollback — nothing recorded), and the publication transaction surfaces it as a warning in `flush` (a new `foldRecordWriteErr`, alongside the marker/steering write errors). Previously the markers landed, the record's write error was dropped, and a restart silently fell back to the last-marker anchor and dropped the retained tail — the #1200 bug with no diagnostic. A whole-line-unsynced record IS a record (queued on the writer and surfaced by `surfaceTranscriptWarnings`), so that returns nil; a fold with no durable marker returns nil (its marker failure is surfaced by `handleCompactionTurnEffects`). RED: `TestFold_RecordWriteFailureIsSurfaced` faults exactly the fold-record write with a clean rollback and asserts the warning is emitted.

### L — doc comment

The `writeFoldRecordLocked` doc block was attached to `headMarkerSeq`; moved it to precede `writeFoldRecordLocked`.

### L — seam-intercepted steering write

`writeSteeringTurnRecordsLocked`'s `testOnly.appendCompactionTurn` seam produces no real transcript entry, so it now marks the steering Seq `NoTranscriptEntrySeq` (was `recordedSeq(0, err)` = 0 on a seam success, which `writeFoldRecordLocked` would have read as entry 0).

Gates: toolchain gofmt; `go vet` ×3 both modules incl. `GOOS=windows -tags evenerfuzz`; golangci-lint 0 issues both modules; **full root-module suite** (`go test ./...` at repo root — only failures are `internal/selfupdate`, the known macOS `/private/var` TMPDIR-symlink issue, untouched by A2 and green on CI's Linux) + full agent module suite; fold/resume/compaction/orphan/attention/steering/delegate-supervision suites; `-race -count=3` on those; apptranscript/appprojector/TUI. No trailers. Merge stays held for Jesse's compat ruling.

obra:
## Review round 4 disposition (2d9f9c1d3)

CI 15/15 green. RoboRev: one Medium, one Low; both accepted.

- **M `attachTranscript` rescans `s.history` for the first `seqHeldPreAttach` marker on every successful held write, so after one failed held write the next success stamps the FAILED turn's slot (which then names the wrong entry) and leaves the successful turn unstamped** — accepted; the Seq-invariant class, in the held-marker path the round-3 invariant test did not cover (all-success only). Fix: consume markers in iteration order with a cursor so `held[i]` maps to the i-th marker; a failed write advances the cursor and leaves that turn `NoTranscriptEntrySeq`. RED: one failed then one successful held write, assert each turn names its own entry or none.
- **L a merge-back turn whose write failed is now silently omitted from `RetainedSeqs` and vanishes on restart while still visible to the running model; the deleted pair-log design re-appended its persisted form** — accepted as a real behaviour change in a failure path: silent loss is not acceptable. Either re-append the persisted form after the record so it survives (as before) or surface it as a warning naming the turn; measure which the fold transaction can do safely under `attentionMu`, and pin it.

Round 5 with the implementer. The merge stays held for Jesse's compat ruling.

obra:
## Round 5 fixes (cef9cac5a)

Merged origin/main (`8a85220f2`, `--no-ff`) first.

### M — attachTranscript stole a later held turn's Seq on a failed write

`attachTranscript` stamped the held turns' `seqHeldPreAttach` markers by rescanning history from the start on each held write, and a failed write `continue`d without consuming its marker. So the next successful write's rescan found the FAILED turn first and stamped it with the success's Seq — naming the failed turn the wrong entry (a later fold record names it; restart resolves it to the successful turn's content) and stranding the successful turn at `-2`. `TestAttachTranscript_StampsHeldTurnAfterInheritedPrefix` only covered all-success.

Fix: consume the markers in flush order with a forward cursor — `held[i]` maps to the i-th still-marked turn — and stamp every held turn's marker, including a failed write's (via `recordedSeq` → `NoTranscriptEntrySeq`), so no later success can rescan back onto it. RED: `TestAttachTranscript_FailedHeldWriteDoesNotStealNextSeq` (one failed then one successful held write; asserts the failed turn is `NoTranscriptEntrySeq` and the successful one carries its real Seq). The class invariant (`assertHistorySeqsNameEntries`) already covers this path via the attach-based tests.

### L — a merge-back turn whose write failed vanished silently on restart

A turn recorded during a fold whose write fails carries `NoTranscriptEntrySeq`; `writeFoldRecordLocked` cannot name a non-existent entry, so it is omitted from the record and dropped on restart while the running model still holds it. The pre-A2 pair log re-appended its PERSISTED (tool-result-evidence-redacted) form after the markers, and that form no longer exists here — re-appending the live turn could leak the private evidence the persisted projection redacts. So the safe action the transaction can take is to name the loss, not recover it (the write failure itself was already warned by `recordTurn`). `writeFoldRecordLocked` now returns the count of such merge-back turns and the transaction emits a warning for them in `flush`. Tradeoff stated in a comment at the merge-back loop. REDs: `TestFold_MergeBackWriteFailureIsSurfaced` (end-to-end: faults a mid-fold recordTurn write, asserts the warning) and `TestWriteFoldRecordLocked_ReportsLostMergeBack` (unit: the count, and that the kept turn with a real entry is still named).

Gates: toolchain gofmt; `go vet` ×3 both modules incl. `GOOS=windows -tags evenerfuzz`; golangci-lint 0 issues both modules; **full root-module suite** (`go test ./...` at repo root — only failures are `internal/selfupdate`, the known macOS `/private/var` TMPDIR-symlink issue, untouched by A2 and green on CI's Linux) + full agent module suite; fold/resume/compaction/orphan/attention/steering/attach/delegate-supervision suites; `-race -count=3` on those; apptranscript/appprojector/TUI. No trailers. Merge stays held for Jesse's compat ruling.

obra:
## Review round 5 disposition (cef9cac5a)

CI 15/15 green. RoboRev: three Lows, no Medium or higher.

- **Low, `outlineLine` emits a `FOLD_RECORD` line per compaction** → accepted. The outline promises to mirror the markdown renderer, which skips the record; the extra line shifts the numbering the model uses to expand turns. Skip `TurnFoldRecord` in the early return and add the outline test.
- **Low, ATIF export carries the fold record as an empty system step** → accepted. Same class as `TurnAttentionResolution`, which already has an explicit skip; add the case and cover it in the conversion test.
- **Low, `recordTurn` doc comment still says append-then-entry** → accepted; the comment now misstates the ordering it documents.

Merge remains held pending the compatibility ruling (pre-A2 readers fail the whole transcript on the new field). Since the PR is held anyway, round 6 fixes all three in place rather than as a follow-up.

obra:
## Round 6 fixes (24ede90a2)

Merged origin/main (`e3b9cf017`, `--no-ff`) first. All three Lows fixed in place (merge held for Jesse's compat ruling).

1. **Outline** (`agent/session_outline.go`, `outlineLine`): `TurnFoldRecord` now returns in the same early skip as `TurnToolResults`/`TurnAttentionResolution`, so the model-visible outline emits no `<seq> · FOLD_RECORD` line per compaction — mirroring the markdown renderer. RED: `TestS3Cov_RenderOutline_FoldRecordSkippedNumberingUnchanged` asserts no FOLD_RECORD line, no line for the record's index, and that the surrounding turns keep their entry-index numbering (the record's slot is skipped, not renumbered).

2. **ATIF** (`agent/internal/atif/atif.go`, `ConvertTranscriptWithOptions`): added `case schema.TurnFoldRecord:` alongside `TurnAttentionResolution`'s skip (comment: private resume bookkeeping), so export no longer emits an empty `source:"system"` step tagged `evener_kind: fold_record`. RED: `TestConvertToATIF_FoldRecordIsOmitted` asserts no such step and that the marshaled trajectory contains no `fold_record`.

3. **recordTurn doc** (`agent/session.go`): the comment said append-then-entry; rewritten to describe the current code — the entry is written first so the turn can be stamped with the durable Seq it spent (via `recordedSeq`), then appended to history carrying that Seq, one atomic pair under attentionMu (no change-history in the comment).

Files changed: `agent/session_outline.go`, `agent/internal/atif/atif.go`, `agent/session.go`, `agent/cov_s3_renderoutline_test.go`, `agent/internal/atif/atif_test.go`.

Gates: toolchain gofmt; `go vet` ×3 (agent module) incl. `GOOS=windows -tags evenerfuzz`; golangci-lint 0 issues; full agent module suite + full root-module suite (only failures are `internal/selfupdate`, the known macOS `/private/var` TMPDIR-symlink issue, untouched by A2 and green on CI's Linux); `-race -count=3` on the fold/resume/compaction/outline/attach suites and the ATIF package. No trailers.

obra:
## Review round 6 disposition (24ede90a2)

CI 15/15 green. RoboRev: one Low, no Medium or higher.

- **Low, the public transcript projection (`publicTranscriptEntry` / `publicTranscriptLine` / `transcriptExpansionJSONL`) still emits the fold record** → accepted. Every other projection skips it; the raw `read_transcript` view exposes resume bookkeeping and an expandable index the outline shows only as a gap. Rounds 5 and 6 each found one more surface that special-cases `TurnAttentionResolution` and not the fold record, so round 7 closes the class: every site that special-cases the attention-resolution marker treats the fold record identically, with a test that enumerates them.

Merge remains held pending the compatibility ruling.

obra:
## Round 7 fixes (c2e1c3add)

Merged origin/main (`754f5669b`, `--no-ff`) first. Closed the class: every site that reads transcript **entries** and special-cases `TurnAttentionResolution` now treats `TurnFoldRecord` the same. In-memory-history sites are left as-is with a stated reason.

### The finding (read_transcript tool) — `agent/session_tools_transcript.go`
- `publicTranscriptEntry` (:1026): returns `false` for `TurnFoldRecord` (dropped like a resolution marker).
- `publicTranscriptLine` (:1089, the jsonl/raw path): rejects the kind, so both re-derived seq numberings stay aligned.
- `transcriptExpansionJSONL` (:1000): `expand_turn` refuses a pin at a fold record.

### Every other site touched (all treat `TurnFoldRecord` like `TurnAttentionResolution`)
| Site | Change |
|---|---|
| `agent/delegate_fork_context.go` :25 | fork inherit filter — skip the fold record (was leaking `fold_id`/`layers`/`retained_seqs` into a child's inherited context) |
| `agent/delegate_fork_context.go` :67 | `completedDelegateContext` — transparent to the pending-tool-round scan |
| `agent/session_tools_find.go` :568 | `rawEntryText` — no searchable text (cross-session content scan) |
| `agent/internal/atif/atif.go` :160 | tool-round look-through (the step switch at :265 already skipped it, round 6) |
| `cmd/evener-doctor/reconstruct.go` :433 | transparent inside a pending tool round |
| `cmd/evener-doctor/reconstruct.go` :512 | reconstruction skips it (private resume bookkeeping, no model content) |
| `agent/session_model_call.go` :1716 | model boundary — never sent to the model (defensive) |
| `agent/eval_probes.go` :124 | model boundary (`turnsToMessages`) — skipped (defensive) |

Already handled in earlier rounds: `transcript_render.go` (r4), `session_outline.go` + `atif.go` step switch (r6); `apptranscript` `ProjectTurn` returns nil by default for it (r1 test pins this).

### Sites deliberately NOT changed, with reason
These special-case `TurnAttentionResolution` but operate on **in-memory history**, which never contains a fold record — `ResumeHistory` omits it and it is never appended to `s.history` live:
- `agent/history_repair.go` :79 (repairs resumed turns)
- `agent/internal/contextmgr/context_manager.go` :544/555/568/581/1587/1746/1764 (attention-transparent history + checkpoint cutoff, over the fold's working history copy)
- `agent/session_attention.go` :122/127/327/1298–1326 (attention-resolution mechanics: matching/creating/folding resolution turns; a fold record is not an attention turn and is ignored by the attention fold)

### Fence test
`TestPrivateRecordsOmittedFromEveryProjection` (`agent/transcript_render_test.go`) runs markdown, outline, `publicTranscriptEntries`, and `expand_turn` over a transcript holding BOTH an attention resolution and a fold record, and asserts neither's tokens (`fold_id`, `retained_seqs`, `attention_id`, kind names, …) leak and that both private kinds are dropped / their pins refused. ATIF (`TestConvertToATIF_FoldRecordIsOmitted`, r6) and apptranscript (`TestProjectTurn_FoldRecordProjectsNothing`, r1) have package-local equivalents.

Files changed: `agent/session_tools_transcript.go`, `agent/delegate_fork_context.go`, `agent/session_tools_find.go`, `agent/internal/atif/atif.go`, `agent/session_model_call.go`, `agent/eval_probes.go`, `cmd/evener-doctor/reconstruct.go`, `agent/transcript_render_test.go`.

Gates: toolchain gofmt; `go vet` ×3 both modules incl. `GOOS=windows -tags evenerfuzz`; golangci-lint 0 issues both modules; full agent + full root-module suites (only `internal/selfupdate` fails — the known macOS `/private/var` symlink issue, untouched by A2, green on CI Linux); `-race -count=3` on the transcript/fold/outline/fork/find suites, the ATIF package, and evener-doctor. No trailers. Merge stays held for Jesse's compat ruling.

obra:
## Review round 7 disposition (c2e1c3add)

CI 15/15 green. RoboRev: two Mediums, both on failure paths the series claims to close.

- **Medium, `stampFoldWrittenSeqsLocked` leaves a failed head marker at `Seq == 0`** → accepted. The `headSeq >= 0` guard skips the stamp exactly when `recordedSeq` produced `NoTranscriptEntrySeq`, so the marker later enters `RetainedSeqs` as `0` and `resumeTurns` resolves it to the transcript's first entry. Stamp unconditionally, as the steering branch does; RED test: failed head-marker write, follow-on fold, restart.
- **Medium, `lastFoldRecordIndex` returns an older fold's record after the newest fold's record write failed** → accepted. Reconstructing from a record older than the newest compaction marker replays summarized-away turns. Round 8 treats a record older than the newest marker as absent (legacy last-marker anchor), and the write-failure warning names that fallback; RED test: record write fails, restart resumes from the newest marker only.

Merge remains held pending the compatibility ruling.

obra:
## Round 8 fixes (bde44243b)

Merged origin/main (`db2d015e1`, `--no-ff`) first. Both Mediums were failure-path holes of the class A2 closes; both fixed with the failure injected.

### M1 — a write-failed head marker left at Seq 0 (`agent/session_compaction.go`, `stampFoldWrittenSeqsLocked`)
The `haveHead && headSeq >= 0` guard skipped stamping the live head marker exactly when `recordedSeq` returned `NoTranscriptEntrySeq` (the marker's own write failed), leaving that turn in `s.history` at Seq 0. A later fold's `name()` then appended 0 to `RetainedSeqs`, and `resumeTurns`' `bySeq[0]` resolved to the transcript's first entry — resurrecting a summarized-away turn. Fix: stamp the matched live marker with `headSeq` unconditionally (as the steering branch already does), so a failed marker write carries `NoTranscriptEntrySeq` and a later fold omits it; the comment claiming "Seq 0 is never load-bearing" is corrected. REDs: `TestStampFoldWrittenSeqs_FailedHeadMarkerMarkedNoEntry` (unit) and `TestFold_FailedHeadMarkerNotNamedByLaterFold` (a write-failed fold-1 marker retained by fold 2; asserts no 0 in `RetainedSeqs` and no resurrected first entry after restart).

### M2 — a stale fold record used when the newest fold's record write failed (`agent/transcript_read.go`, `resumeTurns`)
When the newest fold wrote its markers but its record write failed, `lastFoldRecordIndex` returned an OLDER fold's record; resume replayed that fold's summarized-away turns plus the newer markers. Fix: a fold record is authoritative only when it is newer than the newest compaction marker (`recordIdx > lastMarkerAnchor(entries)`); otherwise it is treated as absent and resume falls back to the last-marker anchor (newest summary + everything after it). A valid fold always writes its record after its own markers, so that inequality is exactly "the record is the newest fold's". The record-write warning is reworded to name the fallback: "restart resumes from the newest compaction marker, the retained tail is not carried". RED: `TestResumeHistory_StaleFoldRecordFallsBackToNewestMarker` (two folds, second record absent; asserts resume is summary2 + post-marker entries only, nothing duplicated or removed).

### Durable side (asked)
`Append` is atomic per line: a hard record-write failure rolls the whole line back (nothing recorded), and a partial line poisons the writer rather than leaving a half-parsed record. So a failed record is cleanly absent — the M2 fallback handles it — and there is no partially-written fold record for a reader to mis-parse. A poisoned writer would have failed the markers too (they are written first), so the markers-present/record-absent case only arises from a clean rollback.

Files changed: `agent/session_compaction.go`, `agent/transcript_read.go`, `agent/session_fold_publication_test.go`, `agent/transcript_test.go`.

Gates: toolchain gofmt; `go vet` ×3 both modules incl. `GOOS=windows -tags evenerfuzz`; golangci-lint 0 issues both modules; full agent + full root-module suites (only `internal/selfupdate` fails — the known macOS `/private/var` symlink issue, untouched by A2, green on CI Linux); `-race -count=3` on the fold/resume/compaction/stamp/attach suites. No trailers. Merge stays held for Jesse's compat ruling.

obra:
## Review round 8 disposition (bde44243b)

CI 15/15 green. RoboRev: one Low, no Medium or higher, and the first round with no failure-path defect.

- **Low, `writeSteeringTurnRecordsLocked`'s doc still says the seam returns `0`** while the code returns `NoTranscriptEntrySeq` → accepted; the stale sentence documents exactly the bug class this series removes. Fixed in place (merge is held anyway).

/simplify is running on this head in parallel; its findings, if any, land in the same round. Merge remains held pending the compatibility ruling.

obra:
## Round 9 fixes (cdb838a1e)

Merged origin/main (`a7d288509`, `--no-ff`) first. Doc fix + all 16 /simplify items, no behaviour change (sentinels, fold-record shape, fail-closed rules are pinned by the existing tests). Net **−57 lines** (263 insertions, 320 deletions), 14 files.

**Doc fix (round-9 start):** `writeSteeringTurnRecordsLocked` doc said the seam returns "0"; now says `NoTranscriptEntrySeq`. Grep found no other comment mislabeling a bare 0 as the sentinel.

Seq class at the source:
1. `stampFoldWrittenSeqsLocked` stamps the head marker at `s.history[0]` **by position** — `folded[0]` is the fresh head, merge-back only appends, attention retain/remove never touch index 0 — dropping the `Seq == 0` match entirely. *Chose position-identity over a new pre-write sentinel: strictly simpler and adds no fourth Seq constant. The callback (~:526) receives the marker by value and cannot mark the live turn, and marking at publish would need the same "fresh marker" test; position identity avoids both.*
2. `DecodeEntry` seeds `Turn.Seq = Entry.Seq` (reader invariant); deleted the resume-path and attention seed sites.
3. `escapeInheritedHistory` marks inherited-prefix turns `NoTranscriptEntrySeq` where they are built; deleted the separate boundary walk.
4. `schema.TurnKind.IsPrivateRecord()` at the non-switch sites (outline, find, read_transcript ×3, atif look-through); switch case-arm sites keep their arms.
5. `resumeTurns` filters fold records once in `appendEntry`. *Only fold records, not `IsPrivateRecord`: attention-resolution turns legitimately remain in resumed history — the attention machinery reconciles from them; only the fold record is never a live turn. (Deviation from item 5's literal "predicate from 4", stated.)*
6. `resumeAnchor`: one backward pass stopping at the newest marker-or-record, replacing `lastMarkerAnchor` + `lastFoldRecordIndex` + the staleness predicate + the second anchor call.
7. `bySeq` map → `lookup` (binary search + `index == Seq` fast path). Verified in the writer: `Entry.Seq` is monotonic in file order (`appendBatchLocked` assigns `firstSeq+i`, resume continues at `maxSeq+1`, a failed write spends no Seq), so entries are sorted by Seq.
8. Preallocate `turns`/`origins`.
9. `headMarkerSeq` computed once in `commitTranscriptsLocked` and passed to both writers; `writeFoldRecordLocked`/`stampFoldWrittenSeqsLocked` no longer take `markers`/`markerSeqs`; the guards collapse to `headSeq < 0`; the marker-kind check reuses `isSessionNameCompactionTurn`.
10. `stampFoldWrittenSeqsLocked` steering scan returns early on empty and breaks once consumed.
11. attention `AppendSynced` is just `seq` (err is nil there). *The seam branch keeps its explicit `NoTranscriptEntrySeq` and does NOT route through `recordedSeq`: on a seam success `recordedSeq(0, nil)` is 0 (entry 0), which cannot express "no entry" — stated in a comment.*
12. The "unspent next seq" explanation is kept once, on `recordedSeq`; the three repeats point to it.
13. `attachTranscript` collects the flushed seqs in the first loop, then stamps in one lock with a single forward pass.
14. Fixed the stale "durable merged-tail copies" comment → "marker, steering and record writes".
15. Renamed `…MergedTailRewriteUsesPersistedForm` → `…DuringFoldTurnResumesFromPersistedForm` and `…SurvivesRestartBeforeRewriteSync` → `…SurvivesRestart`; reworded the three rewrite-era doc comments to fold-record language; dropped the "M1/M2/L/Instance" review-round labels from test docs.
16. `foldRecordFromTranscript` uses `resumeAnchor`; `foldRecordWriteFailFS` folded into `contentWriteFailFS{failMark: "fold_record"}`; new `sessionTranscriptEntries(t, s)` helper absorbs the inline `readTranscriptFull(transcriptPath(...))` copies.

Filed as follow-ups, not this round (as instructed): #1529, #1530.

Files changed: `agent/schema/turn.go`, `agent/transcript/transcript.go`, `agent/transcript_read.go`, `agent/session.go`, `agent/session_compaction.go`, `agent/session_init.go`, `agent/session_lifecycle.go`, `agent/session_attention.go`, `agent/session_outline.go`, `agent/session_tools_find.go`, `agent/session_tools_transcript.go`, `agent/internal/atif/atif.go`, `agent/session_fold_publication_test.go`, `agent/transcript_test.go`.

Gates: toolchain gofmt; `go vet` ×3 both modules incl. `GOOS=windows -tags evenerfuzz`; golangci-lint 0 issues both modules; full agent + full root-module suites (only `internal/selfupdate` fails — the known macOS `/private/var` symlink issue, untouched by A2, green on CI Linux); apptranscript/appprojector/TUI; `-race -count=3` on fold/resume/compaction/attach/outline + atif + evener-doctor. No trailers. Merge stays held for Jesse's compat ruling.

obra:
## Review round 9 disposition (cdb838a1e)

CI 15/15 green. RoboRev: two Lows, no Medium or higher. Fixed in place since the merge is held.

- **Low, the fuzz oracle `trender_isAttentionResolution` still names only the attention marker**, so a re-emitted fold record would find its own source line and pass → accepted; the oracle uses `IsPrivateRecord()` so the fence stays in lockstep with `publicTranscriptLine`.
- **Low, `resumeTurns` silently drops a retained turn whose named `Seq` does not resolve** → accepted; an unresolved named entry is surfaced (warning with the count and the seqs) rather than truncating context in silence, the same posture as the merge-back write failure.

Merge remains held pending the compatibility ruling.

obra:
## Round 10 fixes (c49762f2a)

Merged origin/main (`c0e1a9225`, `--no-ff`) first.

1. **Fuzz oracle** (`agent/transcript_render_fuzz_test.go`): `FuzzRawLinesForRange`'s must-be-omitted oracle recognized only `TurnAttentionResolution`, while `publicTranscriptLine` omits via `IsPrivateRecord()` — a regression re-emitting a fold record would have found its own source line and passed the subsequence check. The helper (renamed `trender_isPrivateRecord`) now uses `kind.IsPrivateRecord()`, and a new seed with a `FOLD_RECORD` between two ordinary entries exercises it (verified with `-fuzz -fuzztime 8s`, plus the seed runs under the normal suite).

2. **Unresolved fold-record seqs** (`agent/transcript_read.go` + `agent/session_init.go`): a fold record naming a Seq that does not resolve (a corrupted or truncated transcript) was skipped silently, dropping a retained turn — the class #1200 fixes. `resumeTurns`/`resumeHistoryReconstruct` now also return the fold id and the unresolved seqs, and `RestoreSessionFromMetaWithConfig` emits one warning naming the fold, the count and the seqs — the same posture as the merge-back write-failure warning. Resume is **fail-open**: it rebuilds from the turns that resolved rather than refusing, since a partial history beats an unrecoverable session (stated in the `resumeTurns` comment). RED: `TestRestore_UnresolvedFoldRecordSeqWarns` — a record naming a seq past the transcript's end; the resolvable turns are present and the warning is recorded.

Note: the warning is emitted after the `Session` is constructed (the reconstruction runs earlier in `RestoreSessionFromMetaWithConfig`, before `s` exists), threaded through function-scope vars — `emit` is a no-op before `s.events` is ready, so it must fire post-construction.

Files changed: `agent/transcript_read.go`, `agent/session_init.go`, `agent/transcript_render_fuzz_test.go`, `agent/session_resume_turn_seed_test.go`.

Gates: toolchain gofmt; `go vet` ×3 both modules incl. `GOOS=windows -tags evenerfuzz`; golangci-lint 0 issues both modules; full agent + full root-module suites (only `internal/selfupdate` fails — the known macOS `/private/var` symlink issue, untouched by A2, green on CI Linux); the `FuzzRawLinesForRange` seed corpus; `-race -count=3` on the resume/fold/restore/attach/outline suites. No trailers. Merge stays held for Jesse's compat ruling.

obra:
## Review round 10 disposition (c49762f2a)

CI 15/15 green. RoboRev: one Low, no Medium or higher.

- **Low, the fork-provenance boundary is a leading-run scan, but a fold record emits the head marker first**, so a compacted fork child's retained parent turns are classified as own and can be routed through the child's mutation journal → accepted. The boundary must derive from the retained turns' own origins (or retained-inherited turns take the nil-origins path). Newly reachable because the old anchor dropped those turns entirely. Round 11 for the next owner; RED test: compacted fork whose retained tail carries a parent steering turn with a colliding `ClientMutationID`.

Merge remains held pending Jesse's compatibility ruling. The queue is being handed over; the A2 implementer agent id and state are in the handoff memo.

obra:
## Round 11 fixes (658df9e05)

### Low (RoboRev @c49762f) — fork provenance after a fold: a leading-run scan cannot express interleaved ordering. ACCEPTED, fixed.

**Measured first.** New RED test `TestRestoredFoldRetainedParentTurnKeepsItsBytes` (agent/session_notes_provenance_test.go), modelled on `TestRestoredForkBoundaryCountsRepairInsertions`: a fork child with `DivergenceTurn: 3` whose transcript is `[0] parent USER_INPUT, [1] parent STEERING (kindless, id collides with a child notes/human/set record), [2] the child's own kindless note-origin STEERING, [3] CHECKPOINT (the fold's head marker), [4] FOLD_RECORD{Layers:[3], RetainedSeqs:[1,2]}`. The fold therefore reconstructs `own marker, inherited retained turn, own tail` — exactly the interleaving the finding names.

```
--- FAIL: TestRestoredFoldRetainedParentTurnKeepsItsBytes (0.16s)
    session_notes_provenance_test.go:718: the retained inherited steer lost bytes to the child's colliding record: "checkpoint the fold wrote\nkeep [31mthese[0m parent bytes\nrun the child tests[31mwith red lines[0m\n"
```

(the ESC bytes are gone from the parent's text — `\x1b[31m` came back as the bare `[31m` the control strip left behind — while the child's own turn was correctly stripped. The leading-run scan stopped at the head marker's origin, so `inheritedPrefix == 0` and every retained turn consulted the child's journal.)

**The fix, at the altitude the finding asks for: provenance is decided per turn from each resumed turn's own origin, not by a boundary index.**

- `escapeHistoryWithSessionProvenance(history, inherited []bool, origins)` replaces the `divergenceTurn int` prefix split: each turn is escaped with the journal's origins, or without them when its own mark says inherited.
- New `inheritedProvenanceTurns(length, divergenceTurn, origins, repairInsertions)` builds that mark: an entry is inside the fork's inherited prefix when its origin index `< DivergenceTurn-1`, so the fold's head marker (own), the parent turns it retained (inherited) and the child's own tail each get their own answer. Absent origins (a caller-supplied `resumeHistory`) the history IS transcript order and position decides, as before. Repair insertions are spliced in so the mask stays aligned with the post-repair history; a synthetic takes the provenance of the call it completes (it is a tool-result turn, which the notes escape never classifies).
- The old "scan the leading run, then shift the boundary right per repair insertion" block in `RestoreSessionFromMetaWithConfig` is gone — two encodings of one rule collapse into one.

Net: `agent/session_init.go` +63/-49, `agent/session_notes_provenance_test.go` +104.

**Equivalence checked on the existing oracles**: `TestRestoredForkBoundaryCountsRepairInsertions` (origins `[0,1,2,3]`, prefix 3, insertion at 1 → mask `[T,T,T,T,F]`, the same four inherited turns the old shifted index produced), `TestRestoredCompactedForkKeepsItsOwnTurnProvenance` (marker-anchored resume, all-own, unchanged), `TestRestoredForkEscapesItsInheritedPrefixWithoutTheChildJournal` (no origins → positional, unchanged). No other caller or test referenced `escapeHistoryWithSessionProvenance` or the old boundary math (grepped: the only other `DivergenceTurn-1` encoding lives in `fork.go`, which slices the parent transcript and is untouched).

PR body updated: the fork-provenance bullet now states the per-turn rule.

### Base
`gh pr view 1480 --json mergeable` = MERGEABLE, so the base was left alone (no refresh, no review round spent). #1248 (69c48e8f5) is on main and not in this branch; its `agent/session_init.go` hunks in `RestoreSessionFromMetaWithConfig` start at old line 1008 — after the provenance block at ~950 — so there is no textual or semantic overlap with this change.

### Gates (all foreground, this head)
- `$(go env GOROOT)/bin/gofmt -l agent/` — no output
- `go vet ./...` (agent module) — clean; `go vet -tags evenerfuzz ./...` — clean; `GOOS=windows go vet -tags evenerfuzz ./...` — clean
- `golangci-lint run` (agent module) — `0 issues.`
- `make lint` — PASS lint-naming, lint-gofmt, lint-evenerfuzz, lint-eval, lint-internal, lint-golangci, lint-generated, lint-fuzz-registry, lint-package-imports, secret-scan
- `go test -count=1 ./agent/...` — all packages ok (no non-`ok` line)
- `go test -count=1 ./agent/transcript/...` — `ok primeradiant.com/evener/agent/transcript 11.442s`
- `go test -race -count=1 -timeout 30m ./agent/` (solo) — `ok primeradiant.com/evener/agent 889.545s` (the default 10m cap is not enough for this package on this host; unchanged from main's runtime)
- `go test -count=3 -run 'TestRestoredFoldRetainedParentTurnKeepsItsBytes|TestRestoredForkBoundaryCountsRepairInsertions|TestRestoredCompactedForkKeepsItsOwnTurnProvenance' ./agent/` — ok
- root `go test -short -count=1 .` — `ok primeradiant.com/evener 20.291s`
- `go test -count=1 ./cmd/... ./server/...` — ok
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — `PASS secret-scan (0s)`; `gitleaks detect --no-git --source agent` — `no leaks found`

`/simplify` ran in round 9 and was not re-run.

obra:
### CI note (658df9e05): the red checks are inherited from main, not from this round

`static`, `static-build`, `fuzz`, `lint-golangci`, `race-root`, `tests` and `web` fail on this PR's merge commit with the same build error, and the identical set fails on **main itself** at `d3c537e45` and at `ab44f4831`, while `ab44f4831`'s parent `64d41ba50` is green:

```
##[error]cmd/evener-hub/app_threadlifecycle.go:499:20: undefined: trace
... (:503, :514, :524, :538, :562, :569, :576)
```

Cause: a semantic merge conflict between #1248 (69c48e8f5), which split `resumeThread` into a wrapper plus `resumeThreadLocked`, and #1390 (ab44f4831), which added `ctx, trace := withThreadLifecycleLog(...)` at the top of what was then one function and `trace.stage(...)` calls throughout it. The two merged textually, so the later `trace` uses now sit inside `resumeThreadLocked`, where nothing declares `trace`. This branch does not touch `cmd/evener-hub/app_threadlifecycle.go` (0 lines in `git diff origin/main...HEAD`).

`race-modules / nonagent`, `api-package`, `native`, `fuzz-checks` and `lint-repository` are green here; `race-modules / agent` — the job this round's change belongs to — is still running.

obra:
Update: `web` also failed on the same `undefined: trace` hub build, and `race-modules / agent` — the job this round's change belongs to — passed. Every red check on this head traces to main's break; #1575 ("declare the Resume lifecycle trace the #1390 merge dropped") is the fix in flight.

obra:
## Review round 11 disposition (658df9e05)

Round 10 Low accepted and fixed: the fork-provenance boundary is derived per turn from each resumed turn's own origin entry (`inheritedProvenanceTurns`, repair insertions spliced into the mask) instead of a leading-run scan, so a fold's own-marker / inherited-tail / own-tail ordering is expressible; RED `TestRestoredFoldRetainedParentTurnKeepsItsBytes` showed the parent's control bytes stripped, GREEN after; the three existing provenance oracles unchanged. The old scan and its separate repair-shift loop are gone (one encoding).

CI: the Go-job failures on this head are the main break (#1571, fix in #1575), not this PR; the agent race job, the one this change belongs to, is green. Rerun once #1575 lands.

**Merge stays HELD** for Jesse's ruling on the pre-A2 reader break (recommendation on file: accept it).

obra:
## Review round 12 disposition (658df9e05)

CI: the Go-job failures on this head were the main break (#1571), fixed by #1575; rerun requested. RoboRev round 11: one High, one Medium, two Lows, all accepted pending the implementer's RED measurement.

- **High (fork context strips the fold record; inherited turns copied without their Seqs)**: fix at one altitude, the fork's inherited history reconstructed through the same fold-record path the session uses, inherited turns carrying the Seq of the child-transcript entry they were written as.
- **Medium (`FOLD_RECORD` takes a logical-turn ordinal in the apptranscript projector)**: the private-record predicate excludes fold records from grouping and ordinal allocation in both projectors; pinned by the live-vs-reload parity oracle with a real fold record.
- **Low (ordinary fold IDs reset on restore)**: derive from something durable the transcript already has rather than a new counter.
- **Low (attention-turn test cannot fail for its stated reason)**: assert what the record names, with a real Seq in the fixture.

Round 12 in progress. Merge stays HELD for Jesse's ruling on the pre-A2 reader break.

obra:
## Round 12 fixes (2728005bc)

### HIGH — a fork of a compacted parent lost the fold's retained tail, and inherited turns named no child entry. ACCEPTED, both measured, both fixed.

**RED 1** (`TestDelegateForkContext_InheritsTheRetainedTailOfACompactedParent`, agent/delegate_fork_context_test.go): a parent records 12 turns and folds (`newScriptedSummaryCompactSession` + `Compact`), so its live history is `[SUMMARY, kept-6 … kept-11]` with the retained entries sitting BEFORE the summary. Snapshot the delegate context, escape it:

```
--- FAIL: TestDelegateForkContext_InheritsTheRetainedTailOfACompactedParent (0.09s)
    delegate_fork_context_test.go:343: the fold's retained turn "kept-6" is missing from the inherited context [{Kind:SUMMARY … [CONTEXT SUMMARY] summary [END SUMMARY] …}]
```

The child inherited the summary and nothing else — #1200 on the fork path, because the snapshot stripped the fold record before `ResumeHistory` could use it.

**RED 2** (`TestDelegateForkContext_InheritedTurnsNameTheirChildEntries` and `TestDelegateForkContext_ChildFoldRetainsInheritedTurns`), falsified by disabling the new stamp:

```
--- FAIL: TestDelegateForkContext_InheritedTurnsNameTheirChildEntries (0.18s)
    delegate_fork_context_test.go:409: inherited turn "kept-6" names child entry 0, which holds "kept-0"
--- FAIL: TestDelegateForkContext_ChildFoldRetainsInheritedTurns (0.18s)
    delegate_fork_context_test.go:441: the child fold retained no inherited turn … [SUMMARY … kept-0 … kept-0 … kept-0 …]
```

Worth noting what the falsification shows: without the stamp the inherited turns keep the PARENT's Seqs, so the child's own fold names entry 0 five times and the resume resurrects `kept-0` repeatedly. Naming them is not just a lost tail — it is a wrong tail.

**Fix — one rule: the inherited history is the resume reconstruction, and every inherited turn names its child entry.**
- `snapshotDelegateContext` no longer strips `FOLD_RECORD`; it carries the record (its `Fold` only) as the manifest of what the parent's live history is made of. Why it was stripped: `git log -S` shows the line was added by this branch's own A2 work, as blanket "never inherit a private record" — correct for the durable copy, wrong for the reconstruction.
- New `splitInheritedContext` separates the snapshot into the durable copy the child re-writes (records dropped — their Seqs name the PARENT's entries and would misdirect the child's own resume) and the resume ORDER over that copy, taken from the same `resumeTurns` the session uses. No second reconstruction and no second anchor scan.
- `escapeInheritedHistory(entries, order)` now repairs + escapes only, and reports each turn's source entry; `stampInheritedHistorySeqs` writes the child Seq each entry spent (`tw.Append`'s return, previously discarded) onto the history turn. A repair synthetic keeps `NoTranscriptEntrySeq`, and that is also the default every inherited turn starts at, so an unstamped turn can never name a parent entry.
- The child's transcript still holds the parent's full archive: `TestDelegateForkContext_ColdResumePreservesSnapshot` (the pre-marker conversation must be on disk but out of model context) still passes — it caught an earlier attempt that made the transcript the resumed subset, and that attempt was dropped.
- `inheritedProvenanceTurns` (round 11) and `escapeInheritedHistory` now share one encoding of the repair shift, `repairedTurnSources`.

### MEDIUM — `FOLD_RECORD` took a logical-turn ordinal on reload. ACCEPTED, measured, fixed.

**RED** (`TestFoldRecordTakesNoLogicalTurnOrdinal`, internal/apptranscript; and `TestFoldRecordDoesNotShiftWireItemPositions`, server/):

```
--- FAIL: TestFoldRecordTakesNoLogicalTurnOrdinal (0.00s)
    logical_turn_grouping_test.go:386: file projection keys = [apptranscript-item-v1:turn_1:0:0 apptranscript-item-v1:turn_3:2:0], want [… turn_3:1:0] (the fold record takes no ordinal)
--- FAIL: TestFoldRecordDoesNotShiftWireItemPositions (0.00s)
    appwire_turns_differential_test.go:250: item ordinals with the fold record = [0 2], want the live [0 1]
```

**Fix**: one predicate, `takesNoLogicalTurn`, applied at both sites — the full projection's accumulator (`logicalTurnAccumulator.appendEntry`) and the indexed path (`recordStartsGroup`, `indexedGroups`, and the previous-kind scan that decides grouping). A fold record is transparent: no group, no ordinal, the open group stays open, and raw entry indexes (and so every persisted turn id) are untouched.

**Deviation from the brief, with the measurement.** The brief said to use `IsPrivateRecord`. That predicate also covers `ATTENTION_RESOLUTION`, and attention resolutions ARE live history turns — the session appends them (`retainDelegateAttentionTurn`), folds preserve them, and `attentionTransparentTurns`/`attentionTransparentRecentCutoff` exist precisely because they sit in `s.history`. Excluding them would change the ordinal of a turn live history does allocate, in the opposite direction, with nothing measured to justify it. `takesNoLogicalTurn` states the real rule ("absent from live history") and names the fold record; its doc says why the other private record is not included. Happy to widen it if you'd rather, but it needs its own measurement.

### LOW 1 — ordinary fold ids reuse `fold-1` after a restart. REFUTED, measured.

`TestFold_IDsAreDistinctAcrossARestart` folds four times, restarts (`RestoreSessionFromMetaWithConfig` on the same state dir), and folds again:

```
MEASURE restored historyRevision=0
MEASURE before=[fold-1 fold-2 fold-3 fold-4] after-restart="fold-5"
```

The restored session's `historyRevision` IS 0, yet the next fold is `fold-5`, because `session_compaction.go:628`'s `fmt.Sprintf("fold-%d", commit.publishedRevision)` is only a fallback: line 629-630 overrides it with `commit.receipt.Operation.PublicationID`, minted by `claimCompactionLocked` from the PERSISTED `s.skillLifecycle.Revision` (`foldPublicationID`). Every fold that writes a record has `actualCompaction == true`, and both branches of the claim mint a receipt in that case, so the in-memory fallback never reaches a written record. `foldPublicationID`'s own doc (session_compaction.go:426-436) says this is exactly why it exists: "The memory-only historyRevision is never seeded on restore — minting the identity from it would re-mint fold-1, fold-2… onto a restored session's publications".

I have left the fallback alone: I could not construct a case that reaches it, so there is no failing test to justify a change, and Jesse's TDD rule says no fix without one. The test stays as the regression oracle for the durable-id property. Say the word if you want the fallback hardened anyway (`fold-entry-<headSeq>` is the one-line version).

### LOW 2 — the vacuous assertion. ACCEPTED, and it uncovered a real resurrection bug.

Rewritten as asked: the fixture now writes the attention turn's entry first (`writeTranscriptDurable`) and retains the turn carrying that entry's Seq, the way production does (`session_attention.go:409-414` — `AppendSynced`, then stamp, then retain); the assertions are now "the fold record must not name that Seq" and "the resumed history must not contain it".

With a real Seq, the corrected test FAILS against the code as it was:

```
--- FAIL: TestFoldPublication_AttentionTurnRemovedMidTransactionNotResurrected (0.10s)
    session_fold_publication_test.go:1478: the fold record retained the removed attention turn's entry (Seq 0) -- it resurrects on resume: &{FoldID:fold-1 Layers:[2] RetainedSeqs:[0]}
```

and with that check suppressed, the ground-truth assertion fails too:

```
    session_fold_publication_test.go:1481: the removed (durability-unverified) attention turn came back through the resumed history
```

So the property the test was written for did not hold; the old assertion (scan the transcript for the text) could never have caught it, because the turn's entry is on disk by design — what decides resurrection is whether the record NAMES it.

**Fix**: `removeUnverifiedDelegateAttentionTurn` remembers the withdrawn entry's Seq (`Session.withdrawnSeqs`, guarded by `mu`), and `writeFoldRecordLocked` never names a withdrawn Seq. Identity is the Seq, not a position: a first attempt using "is this Seq still in live history" was wrong (a fold's own head marker carries an unstamped zero Seq until `stampFoldWrittenSeqsLocked`, so it polluted the set) and it also broke `TestWriteFoldRecordLocked_ReportsLostMergeBack`, which calls the writer directly with a published slice that is not live history. Falsification: with the withdrawn check disabled the test fails exactly as above.

### Base
`gh pr view 1480 --json mergeable` = MERGEABLE at main 6d998fc03, so the base was left alone.

### Gates (all foreground, this head; `-race` solo)
- `$(go env GOROOT)/bin/gofmt -l agent internal server` — no output (adding the `withdrawnSeqs` field re-aligned the two neighbouring comment columns in the same struct block; that is gofmt's own doing)
- agent module: `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...` — all clean; `golangci-lint run` — `0 issues.`
- root module: the same three vets — clean; `golangci-lint run` — `0 issues.`
- `make lint` — PASS lint-naming, lint-gofmt, lint-evenerfuzz, lint-eval, lint-internal, lint-golangci, lint-generated, lint-fuzz-registry, lint-package-imports, secret-scan
- `go test -count=1 -timeout 20m ./agent/...` — all packages ok
- `go test -count=1 ./internal/... ./server/... ./cmd/...` — ok except the pre-existing macOS-only `internal/selfupdate` install-dir failures (`/private` symlink resolution; untouched by this PR, 0 lines in its diff, green on the ubuntu runners)
- `go test -race -count=1 -timeout 30m ./agent/` (solo) — `ok primeradiant.com/evener/agent 1050.456s`
- root `go test -short -count=1 .` — `ok primeradiant.com/evener 24.586s`
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — `PASS secret-scan (0s)`; `gitleaks detect --no-git` on `agent`, `internal/apptranscript`, `server` — no leaks found

obra:
## Review round 13 disposition (2728005bc)

Round 12 accepted. High: both halves reproduced RED (a fork of a compacted parent inherited only the SUMMARY; inherited turns kept the parent's Seqs so the child's fold named entry 0 repeatedly) and fixed at one rule: the snapshot carries the fold record as a manifest, `splitInheritedContext` takes the resume order from the same `resumeTurns` the session uses, and `stampInheritedHistorySeqs` writes the child Seq each entry spent. Medium: fold records no longer take an ordinal, via one `takesNoLogicalTurn` predicate at both projectors; the deviation from `IsPrivateRecord` is accepted on measurement (attention resolutions are live history turns that do take an ordinal), parity pinned at the wire in the AppWire differential test. Low (fold IDs): refuted by measurement, the receipt's PublicationID overrides the in-memory fallback on every written record; the restart test stays as a regression oracle. Low (attention test): accepted, and with a real Seq in the fixture it exposed a withdrawn turn being named and resurrected, fixed by `withdrawnSeqs`. Residual filed: https://github.com/prime-radiant-inc/evener/issues/1587.

Base unchanged at 6d998fc03, MERGEABLE; CI running. **Merge stays HELD** for Jesse's ruling on the pre-A2 reader break.

obra:
## Review round 14 disposition (2728005bc)

CI 15/15. RoboRev: three Mediums, all accepted for round 13. The incremental index's `prevKind` fallback runs only when the scan batch is empty, so a batch that begins on a fold record allocates an extra ordinal and keys drift from the full read: the fallback runs whenever `prevKind` is unset. `withdrawnSeqs` is never cleared when a durable attention turn is re-retained, so the fold record omits it: cleared on the retention path. A delegate forked from a compacted parent and restarted before its first fold resumes with the summary alone: this is #1587, filed as a residual after round 12 and now a finding, closed by writing a child-relative fold record when the child transcript is seeded. Round 13 in progress. **Merge stays HELD** for Jesse's ruling on the pre-A2 reader break.

obra:
Round 13 paused: A2 is parked pending Jesse's ruling on the pre-A2 reader break (queue focus is the SDK migration). The in-flight round-13 edits are committed as WIP cfc633b5b (not gated, not reviewed); the lane resumes from there when the ruling lands.

PR #1485, opened 2026-09-16
refactor(sdk): one delegate timing/model derivation for web and native (D19')

obra:
## Summary

Phase D row **D19'**, replacing the withdrawn D19. The plan paired the web's `panes/session/transcript/tools/subagentModuleStore.ts` (a per-scope row registry over zustand) with native's `mobile-native/src/delegateDetails.ts` (pure timing/model/packet derivations over a delegate). A read-only audit found zero same-inputs→same-outputs overlap, so D19 is withdrawn. The real duplication runs the other way: the web hand-rolled what native's module already does, three times for timing and once for the model name.

This PR moves `delegateDetails.ts` (+ its test) to the package root, widens `delegateTiming`/`delegateModel` to a `Pick<>` of the fields both wire delegate shapes carry (`DelegateTimingFields`, `DelegateModelFields`, so the web's `EvenerDelegateInfo` satisfies them as well as `ActivityDelegate`), and deletes the web copies in favour of the package helpers. The package stays at zero runtime dependencies and imports nothing from either app.

## Confirmed duplication (audit, read fully before changing anything)

| Web copy | Lines | What it re-derived | Where it was laxer than `delegateDetails` |
| --- | --- | --- | --- |
| `transcript/tools/subagentModule.tsx` `cardClock` | 126-143 | run window / live elapsed off `EvenerDelegateInfo` | trusted `runEndedAt` on a non-terminal (resumed) delegate; gated the live clock on display kind, not `terminal`; no clock at all when timestamps did not parse even though the projection carried `durationMs`/`runningForMs` |
| `chrome/ActivityRowDetail.tsx` `subjectOf` + `metaText` delegate branch | 203-238 | quiet anchor `latestActivityAt ?? runStartedAt` | no `max(latest, started)`; an unparseable anchor measured from epoch 0 (`quietAnchorMillis` returns 0) |
| `chrome/ActivityTree.tsx` `delegateMetaSegments` | 209-228 | same quiet anchor, terminal duration | no `max`; an unparseable `latestActivityAt` fell straight to the frozen `quietForMs`; terminal duration preferred the snapshot `durationMs` over the parsed window and accepted a negative span |
| `transcript/EntityRef.tsx` `DelegateCard` | 132, 141 | `resolvedModel ?? model ?? requestedModel` | no blank-trim |

`subagentModuleStore.ts` itself is untouched: it has no native counterpart (the inventory's DUPLICATED verdict on line 197 is wrong; it is a web-only presentation store).

## Behaviour tightenings on the web (each has a test that was red against the old copy)

1. **Quiet anchor is the newer of `latestActivityAt` and `runStartedAt`** (dense row and detail strip). A resumed run keeps the previous run's `latestActivityAt` until the child reports again; the old copies measured silence from it, showing a minute of quiet on a run that started 12 s ago.
2. **An unparseable `latestActivityAt` falls back to `runStartedAt`, then to the frozen `quietForMs`.** The detail strip used to render an age measured from epoch 0; the dense row used to freeze on the snapshot's number.
3. **Terminal duration is the parsed run window (`end >= start`), else the snapshot `durationMs`.** The dense row used to prefer the snapshot and would render a negative span as "0s".
4. **The subagent card trusts the run window only for a terminal projection**, so a resumed delegate's stale `runEndedAt` cannot freeze its clock; and it shows the projection's own `durationMs`/`runningForMs` when timestamps are missing instead of no clock.
5. **A blank `resolvedModel` is absence**: the hover card falls through to the next model name.

One consequence of following native's rule that no test pinned either way: a *terminal* delegate whose child still has active work (`row.live` true) no longer shows a quiet age, because `delegateTiming` reports none for a terminal run.

## One web rule adopted into the shared rule (flagged for the coordinator)

`ActivityTree.test.tsx` "a delegate with no usage or quiet evidence omits inferred quiet age" pins that a live snapshot with neither `latestActivityAt` nor `quietForMs` gets no age inferred from `runStartedAt` alone. That assertion was deliberately reversed from "renders … quiet from startedAt" in `b04f8dd53`, and the `quietForMs != null` gate in `7991140d2` exists to keep it. Native's module inferred from the start and had no test either way, and the web's own detail strip under the same row inferred an age too.

The daemon (`agent/session_tools_jobs.go:1266-1274`) emits `quietForMs` for every open run with a `runStartedAt`, anchored at `latestActivityAt` else `runStartedAt`. So on a current daemon the two behaviours are indistinguishable in every reachable state; they differ only on old-daemon snapshots, where the web rule says "do not invent a number the daemon declined to compute". The shared `delegateTiming` now carries that rule (`hasQuietEvidence`), with its own test, and the web gate is deleted. Native changes only for old-daemon snapshots. This is the one place the shared rule takes the web's stricter behaviour rather than native's; easy to reverse if the coordinator rules otherwise.

## What /simplify changed (commit `3fde2dd71`)

- The quiet-evidence gate moved from the dense row into `delegateTiming` (above).
- `ActivityTree.tsx` `terminalSegment` takes `durationMs | undefined`; the delegate branch no longer re-spells its status-fallback literal, and the job branch passes its parsed span.
- `ActivityRowDetail.tsx` `DetailSubject` carries one measured `quietForMs` and one `durationMs` for both row kinds instead of a job anchor beside a delegate measurement; `metaText` lost a three-way ternary and its re-parse. One consequence, stated: a terminal delegate whose window did not parse but whose snapshot carries `durationMs` now omits the status word, matching the dense row it defers to (tightening 6).
- `ActivityTree.test.tsx`: the new `soleDelegateTree` helper replaces the sparse-delegate test's identical literal; `ActivityRowDetail.test.tsx`: the two anchor tests are one `test.each`.
- `delegateModel` reads `reasoningEffort` once.
- Skipped: a shared render helper for the four subagent-card fixtures (the three existing copies differ in connection setup and item shape; a helper would need more parameters than lines it saves); dropping the `DelegateTimingFields`/`DelegateModelFields` root exports (the publication template is values + types); efficiency nits (a few extra `Date.parse` per delegate per tick, which the old code was not avoiding either).

## Publication

- `tsconfig.build.json` `files` (arms the reachability assertion), `index.ts` root re-export (values + types), one real smoke call in `qualify-package.mjs` (`delegateTiming` on a terminal window under a NaN clock → 5000), README module summary.
- Falsified: with the `index.ts` re-export removed, `make test-api-package` fails with `shipped module unreachable from every published specifier: delegateDetails`; restored with `git diff | git apply -R`.

## Gates (all foreground, one at a time, all green)

`make test-api-package` · `make test-web` · `make test-native` (82 + 7 files; Metro bundle exported) · `make test-web-browser` (skillguard passed first time) · `make lint-package-imports` · `npm run lint` (frontend) · `npx biome ci ../../../appwire-client/typescript` · `go test -count=1 -run TestScenarioSource .`

## Plan

Status table: D19 marked withdrawn ("not a twin; see D19'"), D19' row added → this PR, with `subagentModuleStore.ts`'s real path. Not touched: the inventory (`docs/design/2026-09-12-sdk-migration-inventory.md:197` should flip DUPLICATED → PLATFORM-ON for `subagentModuleStore.ts`; coordinator's call).


obra:
## Refresh onto cca445ea6

`git merge --no-ff origin/main` (no rebase, no force-push) → head `fdb4a300a7ca39873a0c23b82941fd9c2080799c`, `mergeable: MERGEABLE`.

One conflict, the plan status table, resolved as a union: main's landed D7/D2/D3/D20/D10/D1/D8 rows first, then this PR's D19 (withdrawn), D19' and D4 (withdrawn) cells. `index.ts`, `tsconfig.build.json`, `README.md` and `qualify-package.mjs` auto-merged; `delegateDetails` still sorts between `credentialLabels` and `deriveAskQuestions`, and none of `delegateTiming` / `delegateModel` / `delegatePacket` / `DelegateTiming` / `DelegateTimingFields` / `DelegateModelFields` exists on main. `git diff origin/main...HEAD --stat` is exactly this PR's files (17: the original 16 plus the inventory row the coordinator asked for).

Gates rerun in the foreground, one at a time, all green: falsification (`index.ts` re-export removed → `shipped module unreachable from every published specifier: delegateDetails`, restored) then `make test-api-package` · `make test-web` · `make test-native` (80 + 7 files, bundle exported) · `make test-web-browser` (skillguard passed first run, no rerun needed) · `make lint-package-imports` · `npm run lint` · `npx biome ci ../../../appwire-client/typescript` · `go test -count=1 -run TestScenarioSource .`

Issue #1477, opened 2026-09-16
web: connection-store notification wiring is copied in agentsDoc.ts, credentials.ts and keybindings.ts


obra:
D10 (#1476) extracted stores/connection.ts's attachNotifications/wiredClient trio into an exported `onConnectionNotification` (disposer-returning, detaches a replaced client) for the marketplaces core. cmd/evener-hub/frontend/src/stores/agentsDoc.ts, credentials.ts and keybindings.ts still carry their own copies of the same wiring. Consolidate them onto `onConnectionNotification` (D8 1b / D5 will touch credentials.ts and keybindings.ts anyway; do agentsDoc.ts alongside or as its own small PR). Also noted by the same implementer: ~18 hand-rolled `deferRequest`-style test helpers repo-wide are a pre-existing duplication worth its own sweep.

obra:
D5 (#1489) rewrote `stores/keybindings.ts` as an adapter over the package's `createKeybindingsStore`. The notification-wiring copy (`attachNotifications`/`wiredClient`/`unwireNotification`) is gone from the adapter: the core subscribes through its client port once per ready generation and unsubscribes when the generation ends. What remains adapter-side is the ready/rewire lifecycle (`rewireClient` + `onConnectionChange` driving `beginReadyGeneration`/`endReadyGeneration`/`detachHub`/`setSupport`), which is a different shape from D10's `onConnectionNotification` (wire-to-current-and-every-future-client). Remaining copies for this issue: `agentsDoc.ts` and `credentials.ts`.

PR #1488, opened 2026-09-16
refactor(native): read the package's ItemModel/TurnModel; project.ts becomes the shim (D21)

obra:
Row D21 of docs/superpowers/plans/2026-09-12-sdk-migration.md — the first step of seam 2 (native adopts the package's thread model; decision 2: both frontends go wire → ThreadModel → display entries).

`mobile/src/conversation/model.ts` is deleted. The conversation native holds is now the package `ThreadModel`'s thread-level contract — every field a native screen reads, in the web's shape — plus the display rows the shim still projects itself. `mobile/src/conversation/project.ts` is the shrinking shim: it projects the wire into that shape (D22 replaces its hydrate half with `reducer.hydrateThread`, D23 the store's appliers, D24 the display rows).

## Field-by-field mapping (native model → package model)

| Native (`model.ts`) | Package | Disposition |
| --- | --- | --- |
| `MobileConversation` | `ThreadModel` | `Pick<ThreadModel, …> & { items: MobileTimelineItem[] }` in project.ts. A full `ThreadModel` is assignable to it, which is what D22's `hydrateThread` hands in |
| `.id` | `.threadId` | renamed at every read (store ×7, screens ×4) |
| `.activeTurnId?` | `.activeTurnId?` | same |
| `.instanceId?` | `.instanceId?` | same type; the shim keeps native's `?? thread.id` default until D22 reconciles with `hydrateThread` (which leaves it undefined) |
| `.sessionId` | — (`imageSessionId` is the nearest) | **dropped: no reader** |
| `.name?` | `.name: string` (`?? ""`) | reads that fell back with `??` now use `\|\|` (SessionSheet, screens title) |
| `.preview` | — | **dropped: no reader** |
| `.modelProvider` | `.modelProvider` | same |
| `.visionModel?` | `.visionModel: string` (`?? ""`) | ComposerSettingsSheet's `?? ""` removed |
| `.status: string` | `.status: ThreadStatus` | reads → `.status.type`; `sessionControls(status.type, …)` as the web's `liveControls.ts` does |
| `.resumeRequired?` | `.resumeRequired?` | same |
| `.mutationStateAuthoritative?` | — | **dropped: no production reader** (only `project.test.ts` asserted the projection copied it) |
| `.items: MobileTimelineItem[]` | — (D24: `transcriptDisplay/projector.ts` entries) | native-only extension until D24 |
[...]

PR #1489, opened 2026-09-16
refactor(sdk): one keybindings store core for web and native (D5)

obra:
SDK migration phase D, row D5: the keybindings overrides store becomes one framework-free core in the package, `appwire-client/typescript/keybindingsStore.ts`, that the web adapter and native's `NativePreferences` both run on.

## What moves

- **Core** `createKeybindingsStore({ client, registry?, characterKeyTriggers?, drafts? })` returns the D30 `FrameworkFreeStore` triple plus the host-driven connection lifecycle (`setSupport`, `beginReadyGeneration`, `endReadyGeneration`, `detachHub`, `reapplyOverrides`, `reset`, `dispose`). State is the web store's shape (`hubSupport`, `hubLoading`, `hubError`, `revision`, `appliedSerial`, `overrides`, `rawOverrides`, `warnings`, `loaded`, `conflict`) plus `loadError` and native's draft editor (`draft`, `saving`, `writeUncertain`, `storageUnavailable`, `draftConflict`, `draftError`). Actions: `refreshOverrides`, `patchOverrides` (the web's serialized, preflight-validated direct write), and `editDraft`/`saveDraft`/`discardDraft`/`rebaseDraft` (native's checkpointed write over the injected draft port). `KeybindingsClient = Pick<AppwireClient, "request" | "onNotification">`; zero runtime deps, no React/zustand/mobile-src/web imports.
- **Reconciler port inside the core**: `registryReconciler` owns the applied map, the delta rebind, the two-phase un-apply, snapshot/rollback and the pref-aware restore; `verbatimReconciler` publishes the hub's rules as-is for a host with no dispatcher (native). The registry is optional for that reason: native never had a live registry to reconcile into.
- **Pure helpers**: `keybindingsSupport(features)`, `fromWireOverrides(value)`; the draft port types `KeybindingDraftStorage`/`KeybindingDraftCheckpoint` move here from `mobile-native/src/keybindingDraftRepository.ts` (deleted; `remove()` dropped from the port, nothing called it).
- **Web adapter** `cmd/evener-hub/frontend/src/stores/keybindings.ts` (851 → 121 lines) keeps its path and export surface (`keybindingsStore`, `useKeybindingsStore`, `resetKeybindingsStoreForTests`, `KeybindingsStoreState`). It owns only the connection lifecycle: which client is wired, ready generations, and `setSupport` from the connection's feature set. Importers unchanged: `cheatsheetController.ts`, `sections/keybindings.tsx` (+4 tests).
- **Native** `NativePreferences` keeps its class and snapshot shape; the keybinding half (≈330 lines) is now a projection of the core's state plus four one-line delegations. `bindNativePreferences.ts`, `nativePreferenceDrafts.ts`, `keybindingRecovery.test.ts` take the package types. The transcript half is untouched (D6).

## Publication (phase-D rule)

`tsconfig.build.json` files, `index.ts` re-exports (values + types), one real smoke call in `qualify-package.mjs` `rootSmokeCalls` (builds a store over an offline client, publishes support, decodes a payload), README module summary. Falsified: removing the `index.ts` re-export fails `make test-api-package` with `shipped module unreachable from every published specifier: keybindingsStore`.

## Behaviour reconciled between the two apps (pinned in `keybindingsStore.test.ts`)

- A GET carrying `loadError` applies at any revision (native's rule; the web gains it: a retained higher revision no longer keeps editing enabled against a hub whose PATCH rejects).
- Native no longer publishes "Keybinding settings changed. Refresh to inspect the current value." on a malformed `changed` notification: the shared store drops an undecodable broadcast silently, as the web always has. Intended: that message was a refresh affordance, which the native app must not show (Jesse's 2026-09-13 ruling: native auto-refreshes, never a "refresh to see the latest" prompt), and the next GET or well-formed broadcast carries the truth; a `changed` arriving before the first GET confirms takes the web's dirty-flag path (one follow-up fetch) instead of applying directly; a hub-sourced refresh error persists through a local edit until the next refresh (it is derived from `hubError`, not cleared by the edit).
- Native's `refresh` no longer discards a current GET payload while a checkpointed write is in flight; the read applies and only the `writeUncertain` settlement is gated on the read having started after the write.

## Tests

- New `appwire-client/typescript/keybindingsStore.test.ts` (10): two stores over two registries share nothing (payload, notification, draft ports), store shape, registry-less verbatim mode, the checkpointed editor (intent persisted before the PATCH leaves, lost reply → uncertain until an authoritative read, synchronous restore), the `loadError` rule, the two decoders.
- Oracles unchanged: `stores/keybindings.test.ts` (65), cheatsheet/holdhints/settings suites, `nativePreferences.test.ts`, `keybindingRecovery.test.ts`. One native assertion adjusted: `bindNativePreferences.test.ts` counted one notification subscription per model; a model now holds two (the store's and the class's transcript listener, until D6), so it asserts the count and that a replacement model releases the old one's.

## #1477 (connection-notification wiring copies)

D10 (#1476) has not landed, so `onConnectionNotification` is not available. The adapter no longer carries the `attachNotifications`/`wiredClient` notification copy: the core subscribes per ready generation through the client port. What remains in the adapter is the ready/rewire lifecycle (`rewireClient`, `onConnectionChange`), which is a different shape from the helper. Noting on #1477.

## Gates (all local, foreground, on the final head)

`make test-api-package`, `make test-web`, `make test-native` (83 + 7 files, Metro bundle included), `make test-web-browser` (all six guards, first run), `make lint-package-imports`; from `cmd/evener-hub/frontend`: `npm run lint`, `npx biome ci ../../../appwire-client/typescript`; root `go test -count=1 -run TestScenarioSource .`. `/simplify` run (4 angles): reconciler port replacing four `registry === undefined` branches, draft settlement out of the hub-read path, one publish per applied payload (was 3–5), `errorText`/`WireError.evenerErrorInfo` reuse, `staleDraft`/`stillMine` helpers, dead `remove()` port method and `keybindingRules` export dropped, native projection keeps `confirmed.rules` identity-stable for the screen's preview memo.


obra:
## Refresh onto 66727cbe6

Merged `origin/main` (66727cbe6, D21 #1488) with `--no-ff`; one conflict, the plan status table, resolved as a union (main's D7/D2/D3/D20/D10/D1/D8 1a/D19'/D4/D21 rows, then D5). `index.ts`, `tsconfig.build.json`, `README.md` and `qualify-package.mjs` auto-merged as unions; `git diff origin/main...HEAD --stat` shows only D5's 14 files.

**#1477 / D10's `onConnectionNotification`:** confirmed nothing to adopt. The adapter carries no notification-wiring copy: the core subscribes through its client port once per ready generation and unsubscribes when the generation ends. What stays adapter-side is the ready/rewire lifecycle (`rewireClient` + `onConnectionChange` driving `beginReadyGeneration`/`endReadyGeneration`/`detachHub`/`setSupport`), which is a different shape from wire-to-current-and-every-future-client. Noted on #1477.

Gates on this head, foreground: `make test-api-package` (plus the re-export falsification, still `shipped module unreachable from every published specifier: keybindingsStore`), `make test-web`, `make test-native` (80 + 7 files, Metro bundle), `make test-web-browser` (all six guards, first run, no skillguard rerun needed), `make lint-package-imports`, frontend `npm run lint` + `npx biome ci ../../../appwire-client/typescript`, root `go test -count=1 -run TestScenarioSource .` — all green.

obra:
## Review round 1 disposition (66ffa1d09)

CI 15/15 green. RoboRev: one Medium.

- **Medium, `saveDraft` leaves `saving: true` forever when the ready generation ends mid-write** → accepted. Both exits guard the `saving: false` write on `isCurrent(generation)`, and nothing else clears the flag, so a store reused across a reconnect (the web singleton) has every draft mutation refused by `assertEditable` until `reset()`. Round 2 clears the in-flight flag where the generation ends, with a test that ends the generation mid-save on one instance, starts a new generation, and edits again.

obra:
## Round 2 fixes (38f2a198c)

**Medium (accepted): `saving` stuck after a generation ends mid-write.** Fixed at the one place the generation ends, `endReadyGeneration`, as recommended: when `saving` is true it now publishes `saving: false, writeUncertain: true` alongside the existing `loaded: false, revision: 0`. The two post-await exit paths of `saveDraft` stay fenced by `isCurrent(generation)` (which covers `dispose`) and write nothing after the generation ended, exactly as before.

Why `writeUncertain: true` and not only `saving: false`: a reply the generation fence discards is a write whose outcome the store does not know, which is what `writeUncertain` means and what its checkpoint on disk already says (`persistDraft(..., writeUncertain: true)` runs before the request leaves). Clearing only `saving` would leave the in-memory state claiming a settled write over a checkpoint claiming an unsettled one. The next generation's authoritative read (`settledWrite`, `writeEpochAtStart === writeEpoch` since the write already left) clears it and re-marks the checkpoint, so a reconnecting host is editable again after its first refresh. The alternative, clearing in both exit paths, would need each path to know whether it was fenced by a generation end or by `dispose`; the fence already erases that distinction on purpose.

**TDD:** two RED tests first in `keybindingsStore.test.ts` (`a generation ending mid-write leaves the write uncertain, not saving, and the next read settles it`; `a write rejecting after its generation ended writes nothing back and the store stays editable after reconnect`), both failing on `saving: true`, then the fix. Package suite 12/12; the web (65 + cheatsheet/holdhints/settings) and native (`nativePreferences`, `keybindingRecovery`, `bindNativePreferences`) oracles unchanged and green.

Merged `origin/main` e3b9cf017 (#1510) with `--no-ff`, no conflicts; `git diff origin/main...HEAD --stat` is D5's 14 files.

Gates on this head, foreground: `make test-api-package` (falsification re-run: `shipped module unreachable from every published specifier: keybindingsStore`, restored), `make test-web`, `make test-native` (80 + 7 files, Metro bundle), `make test-web-browser` (6/6, first run), `make lint-package-imports`, frontend `npm run lint` + `npx biome ci ../../../appwire-client/typescript`, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, root `go test -short -count=1 .` (scenario source + Makefile audits) — all green.

obra:
## Review round 2 disposition (38f2a198c)

CI 15/15 green. RoboRev: one Medium.

- **Medium, `settledWrite` clears `writeUncertain` but leaves the "could not be confirmed" `draftError`** → accepted. The old native `refreshKeybindings` nulled the error on a successful authoritative read; the projection now keeps showing the message after the write is confirmed. Round 3 clears the unconfirmed-write message on the settling path, with a test that walks lost reply → authoritative read → error null.

obra:
## Round 3 fixes (b51baee6b)

**Medium (accepted): `settledWrite` cleared `writeUncertain` but left `saveDraft`'s `draftError`, so native's projection (`error = draftError ?? hubErrorMessage`) kept showing "The hub request could not be confirmed." after the hub had confirmed the write.**

Fix in `settledWrite`: any read that started after the last checkpointed write (and not during one) now clears `draftError`; an authoritative read (no `loadError`) additionally settles the outcome (`writeUncertain: false`, checkpoint re-marked) as before. A `loadError` read clears `draftError` but leaves `writeUncertain` alone.

Which class of message survives, and why: `persistDraft`'s own `DRAFT_SAVE_FAILED_MESSAGE` from inside `settledWrite` is unaffected (that branch returns its message instead). Storage-failure messages set with `storageUnavailable: true` (`DRAFT_CLEANUP_FAILED`, `DRAFT_RESTORE_FAILED`) never reach `settledWrite`: `refreshOverrides` restores the draft first and only proceeds to the read once the port is healthy, and the successful restore already sets `draftError: null`. So the clear is effectively scoped to the unconfirmed-write copy without a message-class check. This matches the old native `refreshKeybindings`, which published `error: null` at the start of every read and `error: <load message> | null` on success — including on a `loadError` read, which is why that case clears the write copy too.

**Native oracle:** `nativePreferences.test.ts` / `keybindingRecovery.test.ts` never asserted `error` after a keybinding refresh (they pin `writeUncertain`, `draft`, `storageUnavailable`), which is how the regression passed them; both still pass. The rule is now pinned in the package: the lost-reply test asserts `draftError: null` after the settling read, and a new test walks lost reply → `loadError` read → `writeUncertain: true, draftError: null`. Both were RED (`"The hub request could not be confirmed."`) before the fix.

Merged `origin/main` 754f5669b (#1515 D22) with `--no-ff`; one conflict, the plan status table, resolved as a union (D22's row, then D5's). `git diff origin/main...HEAD --stat` is D5's 14 files.

Gates on this head, foreground: `make test-api-package` (falsification first: `shipped module unreachable from every published specifier: keybindingsStore`, restored), `make test-web`, `make test-native` (80 + 7 files, Metro bundle), `make test-web-browser` (6/6, first run), `make lint-package-imports`, frontend `npm run lint` + `npx biome ci ../../../appwire-client/typescript`, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, root `go test -short -count=1 .` — all green.

obra:
## Round 4 fixes (88a32fb8b)

Applied, by item:

1. **draftError is the draft port's channel only.** `saveDraft`'s catch sets `saving: false, draftConflict: true, writeUncertain: true` and nothing else; a MALFORMED reply rides `hubError` like `patchOverrides`' does. `settledWrite`'s `loadError` branch returns `{}` again and the round-3 `draftError: null` clearing is gone. `WRITE_UNCONFIRMED_MESSAGE` deleted from the core (native keeps `HUB_UNCONFIRMED_MESSAGE` as host text for `hubError`). The round-2/3 scenarios are kept and assert `writeUncertain` true → false with `draftError: null` throughout; one new test pins the malformed-reply → `hubError` path.
2. **`endReadyGeneration` publishes the whole in-flight-over set** (`hubLoading: false, saving: false, writeUncertain: state.writeUncertain || state.saving, loaded: false, revision: 0`), unconditionally. The change-guard is gone: native's projection subscription is unsubscribed before `dispose` ends the generation, so no native snapshot sees the publish, and no web oracle keys on notification count there. New test: a generation ending mid-refresh ends `hubLoading` with it.
3. **One `writeSerial`** bumped by both `patchOverrides` and `saveDraft`; `stillMine` and `settledWrite` compare against it. Consequence stated: a checkpointed write leaving while a direct PATCH is in flight makes the PATCH's reply land nothing (no host uses both paths today).
4. **`setSupport` owns the load trigger; the web adapter's fourth branch is gone.** One deliberate widening of the brief: the store refreshes on any transition INTO supported with a ready generation active, not only the unsupported→supported flap-back. The adapter branch being removed fired on `previous.features?.keybindingsSettings !== true`, which covered unknown→supported too; that is the web's initial load whenever the handshake's features land after the client is ready (`connect()` clears features, the ready generation begins, and the refresh it kicks is refused as "unknown"). Flap-back only would have dropped that load in production while the oracle's `wireClient` helper (which awaits `refreshOverrides()` explicitly) stayed green. The generation bump stays flap-back-only (finding 24). Not in `beginReadyGeneration`, as instructed. Native: `setSupport` runs before `beginReadyGeneration` in the constructor, so the only read per model is `bindNativePreferences`' refresh; the native oracle's request counts are unchanged (`gates unsupported domains without issuing reads` and the domain-parametrised counts pass). Two new tests pin both: unknown→supported after ready loads exactly once and a same-value publish loads nothing; support published before a generation loads nothing until the host's refresh.
5. **Null-object draft port**: `drafts = draftRepository(deps.drafts ?? memoryDraftStorage())`; every `drafts?.` and both null guards are gone.
6. **One rule type**: `KeybindingsRule` throughout the module (`OverrideRule` no longer imported). `keybindingValidation.ts` untouched; the web's `OverrideRule` values assign structurally.
7. **`KeybindingsStoreFields` + `KeybindingsStoreActions`**, `KeybindingsStoreState = Fields & Actions`; both new types exported from the root.
8. **`draftRepository`** validates on `load()` only; `save`/`removeIf` are the port's own.
9. **Native**: `ConfirmedKeybindings` types `confirmed.rules` as `readonly KeybindingsRule[]`, the projection is a plain `keybindingsDomain(state)` passing `state.rawOverrides` through (fresh per applied payload via `cloneRules`, identity-stable otherwise), the raw/copy memo is gone. `KeybindingPreferencesScreen` already typed its props `readonly` and its `useMemo` keys on the `rules` identity, so the preview memo behaves the same; `npm run check` green.
10. **Tests**: `readyStore(client, deps)` replaces the five hand-rolled setups; `getMethod`/`patchMethod`/`changedMethod` consts; `clientServing(revision, rules)`.
11. `keybindingsSupport` and `fromWireOverrides` expects are `test.each` tables (4 + 7 rows).
12. **Shared fixtures**: `testing/keybindingDraftStorage.ts` (`memoryKeybindingDraftStorage()` with `stored()`, `failSave(fail)`, `corrupt()`), used by the package suite and `mobile-native/src/keybindingRecovery.test.ts` (its inline storage and three knobs replaced); `testing/deferred.ts` gains `reject` and the generation-end scenarios use it; `testing/keybindingRegistry.ts` (`registryWithDefaults()`) used by `keybindingsStore.test.ts`, `keybindingOverrides.test.ts` and `keybindingValidation.test.ts`.

Nothing skipped. One note on process: items 2 and 4 are behaviour changes whose tests were written with the change, not RED-first against the round-3 head (the round-2/3 regression tests were RED-first).

Merged `origin/main` f7ca94e05 with `--no-ff`, no conflicts; `git diff origin/main...HEAD --stat` is D5's 19 files (the four new/changed fixture and sibling-test files are this round's).

Gates on this head, foreground: `make test-api-package` (falsification first: `shipped module unreachable from every published specifier: keybindingsStore`, restored), `make test-web`, `make test-native` (80 + 7 files, Metro bundle), `make test-web-browser` (6/6, first run), `make lint-package-imports`, frontend `npm run lint` + `npx biome ci ../../../appwire-client/typescript`, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, root `go test -short -count=1 .` — all green. Package suite 17 tests; web oracles (`stores/keybindings.test.ts` 65 + cheatsheet/holdhints/settings) and native oracles (36) unchanged and green.

obra:
## Review round 4 disposition (88a32fb8b)

CI: 12 green, `race-modules / agent` failed (Go race suite; this PR touches no Go, triage on #1394). The PR is also CONFLICTING against main after D8 1b landed, so round 5 refreshes.

RoboRev: one Medium, one Low.
- **Medium, the web adapter never seeds `hubSupport` for a client already wired at module evaluation** → accepted. `rewireClient` → `refreshFor` returns on `"unknown"` support, so a module graph that evaluates the store after the handshake gets defaults with editing disabled until an unrelated connection update. Seed support from `connectionStore.getState().features` before the initial rewire, with a regression test that builds a ready client + features before creating the adapter and asserts the GET.
- **Low, `detachHub`'s clean path leaves `loaded: true`** → accepted and folded into round 5 (a round is needed anyway): a host following only the public doc could PATCH the previous hub's state at `expectedRevision: 0`. Add `loaded: false`.

obra:
## Round 5 fixes (ffca869ea)

Refreshed first: `git merge --no-ff origin/main` (db2d015e1, D8 1b); the one conflict was the plan status table, resolved as a union (D23a's row, then D5's); README, `qualify-package.mjs` and `index.ts` auto-merged. `git diff origin/main...HEAD --stat` is D5's 20 files (19 + this round's new adapter test).

**Medium, `stores/keybindings.ts` initial wiring.** RED first: a new `stores/keybindings.initialWiring.test.ts` imports the adapter module fresh (`vi.resetModules`) against a connectionStore already holding a ready client and its features, and asserts exactly one GET (plus a second case: an unsupported handshake publishes `unsupported` and issues no request). Both failed on the round-4 head (`revision 0`, `hubSupport "unknown"`). Fix: the initial wiring seeds `keybindingsStore.setSupport(keybindingsSupport(initial.features))` before `rewireClient(initial.client)`, the same order `onConnectionChange` uses. With round 4's transition rule the count is one, not two: `setSupport` runs with no generation active (no refresh), then `rewireClient` begins the generation and its refresh runs under `supported`. The order is what keeps it at one; seeding after the rewire would also load once (the rewire's refresh refused as unknown, the transition then refreshing) but would not match the live path.

**Low, `detachHub` clean path.** RED first: `detachHub` on a loaded store without a preceding `endReadyGeneration` left `loaded: true`, so `editDraft` and `patchOverrides` would have composed against the reset payload at `expectedRevision 0`. Fix: `loaded: false` joins the payload reset. The test asserts the store is non-current and both editors refuse.

Oracles unchanged and green: `stores/keybindings.test.ts` (65) + cheatsheet/holdhints/settings; native `nativePreferences` / `keybindingRecovery` / `bindNativePreferences` (36) + `npm run check`; package suite 18.

Gates on this head, foreground: `make test-api-package` (falsification first: `shipped module unreachable from every published specifier: keybindingsStore`, restored), `make test-web`, `make test-native` (80 + 7 files, Metro bundle), `make test-web-browser` (6/6, first run), `make lint-package-imports`, frontend `npm run lint` + `npx biome ci ../../../appwire-client/typescript`, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, root `go test -short -count=1 .` — all green.

obra:
## Review round 5 disposition (ffca869ea)

CI 15/15 green. RoboRev: two Mediums, both accepted.

- **Medium, `saveDraft` fences its reply on `isCurrent(generation)` alone while `patchOverrides` and `refreshFor` also require `isSupported()`** → accepted. A late PATCH reply after a support drop re-applies overrides into the registry under `hubSupport: "unsupported"`.
- **Medium, `detachHub`'s rollback branch leaves `loaded: true`** while its success branch and the support-loss rollback drop it → accepted. Same stale-PATCH-at-revision-0 hole the round-4 fix closed on the other branch.

Rounds 2 through 5 have each found one more encoding of the same two rules ("is this reply still mine" and "reset the payload so it cannot be PATCHed against") that disagreed with its siblings. Round 6 closes the class instead of the instance: one reply-fence predicate shared by every post-await site, one payload-reset helper shared by every reset site, and a table-driven test over reset sites × write paths.

obra:
## Round 6 fixes (6b29db8d4)

Closed the class rather than the two instances.

### One reply fence

`liveHub(generation) = isCurrent(generation) && isSupported()`, with `readStillMine(generation, serial) = liveHub && serial === refreshSerial` and `writeStillMine(generation, token) = liveHub && token === writeSerial`. Every `await` in `keybindingsStore.ts` and the predicate its post-await writes use:

| site | await | fence |
| --- | --- | --- |
| `refreshFor` | `client.request(get)` | pre: `liveHub(generation)`; after the await and in `catch`: `readStillMine(generation, serial)` |
| `patchOverrides` (`run`) | `client.request(patch)` | at execution (before the request): `liveHub(callGeneration)`; after the await and in `catch`: `writeStillMine(generation, token)` |
| `saveDraft` | `client.request(patch)` | before the request (cancel), in `catch`, for the newer-external-revision check, and before the success publish: `writeStillMine(generation, token)` — saveDraft now takes its own write token and requires advertised support, which is the first Medium |
| `refreshOverrides` | `refreshFor(activeEpoch)` | no state write after it |
| `writeQueue.then(run, run)` | not an await; `run`'s own fences above |

The notification wrapper checks `isCurrent(generation)` before dispatch and `onNotification` gates on `isSupported()`; not a post-await site, listed for completeness.

### One payload retirement

`retirePayload(extra)` publishes `loaded: false, revision: 0, saving: false, hubLoading: false, writeUncertain: state.writeUncertain || state.saving` plus the site's extras, and bumps `refreshSerial` and `writeSerial` so no reply still on the wire lands. The four sites:

| site | extras |
| --- | --- |
| `endReadyGeneration` | none (epoch bump + unwire before it) |
| `setSupport("unsupported")` | `hubSupport`, `hubError` (rollback copy or null), `conflict: null`, and the payload arrays + `loadError: null` when `dropHubState` |
| `detachHub`, clean | payload arrays, `loadError: null`, `hubError: null`, `conflict: null` |
| `detachHub`, rollback | `hubError: UNAPPLY_ROLLED_BACK_MESSAGE`, `conflict: null` (payload arrays retained; second Medium: `loaded` now drops here too) |

Two consequences worth naming: `detachHub` now supersedes in-flight replies (it changed neither epoch nor support before, so a late PATCH/GET reply after a detach without `endReadyGeneration` would have applied — three more RED cells beyond the two RoboRev named); and the unsupported branch always publishes (the change-guard stays on the unknown branch only, which is the transient-disconnect window that keeps the payload).

### Table test

`keybindingsStore.test.ts`: 4 reset sites × 3 write paths with a deferred reply on the wire (12 cells). Each asserts the late reply landed nothing (`revision 0`, the late rules absent from `rawOverrides` and the registry), `saving`/`hubLoading` false, `loaded` false, and `editDraft`/`patchOverrides` refuse. RED before the fix: 7 cells (setSupport(unsupported) × saveDraft; detachHub rollback × all three; detachHub clean × all three). The rollback cells use the oracle's wedge (free the override, squat palette.open's registered default chord).

Behaviour otherwise identical; oracles unchanged: web `stores/keybindings.test.ts` (65) + `initialWiring` (2) + cheatsheet/holdhints/settings; native trio (36) + `npm run check`; package suite 40.

Merged `origin/main` with `--no-ff` (tip was 3cbfaebfc by then, past 46b518fc9), no conflicts; `git diff origin/main...HEAD --stat` is D5's 20 files.

Gates, foreground. On `aaa9e16d4` (the merge): `make test-native` (80 + 7 files, Metro bundle), `make test-web-browser` (6/6, first run) — neither loads package test files, and the one later commit is a type annotation in `keybindingsStore.test.ts`, so they stand. On `6b29db8d4` (final head): `make test-api-package` (falsification first on the merge head: `shipped module unreachable from every published specifier: keybindingsStore`, restored), `make test-web` (the first run on the merge head failed `web-typecheck` on that test's `string`-typed method column; fixed as the final commit, green on rerun), `make lint-package-imports`, frontend `npm run lint` + `npx biome ci ../../../appwire-client/typescript`, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, root `go test -short -count=1 .` — all green.

obra:
## Review round 6 disposition (6b29db8d4)

CI 15/15 green. RoboRev: one Medium, one Low.

- **Medium, `discardDraft` can no longer remove the persisted checkpoint** → accepted. `persistDraft` writes the checkpoint with `id` last while `load()` normalizes it with `id` first, and native storage compares serialized JSON, so `removeIf` never matches and a discarded draft reappears on the next launch. The in-memory test storage masks it with integer-like ids. Round 7 normalizes both sides in the repository and pins it with a non-integer id generator.
- **Low, native no longer surfaces a "refresh to inspect" error on a malformed `changed` notification** → intended. That text was a refresh affordance, which the native app must not show (ruling of 2026-09-13); the shared store drops an undecodable broadcast exactly as the web always has. Round 7 states this in the PR body's behaviour deltas.

obra:
## Round 7 fixes (775404afd)

**Medium (accepted): `discardDraft`'s `removeIf` never matched a stored checkpoint on a byte-comparing port.** `draftRepository` normalized only `load()` through `draftCheckpoint` (id first) while `persistDraft` wrote `{ ...input, id }` (id last) and `removeIf` passed through; native's port compares JSON strings and production ids are UUIDs (insertion order kept), so a discarded draft survived to the next launch. Fixed at the repository, where both sides meet: `save()` and `removeIf()` now go through `draftCheckpoint` too, so the port sees one key order in both directions (what the pre-refactor `KeybindingDraftRepository` did).

RED first: `testing/keybindingDraftStorage.ts` now issues non-integer ids (`draft-N`, like production's UUIDs; an integer-like id is an index-like key V8 orders first, which is exactly how the in-memory port had been masking this) and records the checkpoint each `removeIf` received. The new test saves through `editDraft`, discards, and asserts the bytes the port was given equal the bytes it had stored and that it now holds `null`; before the fix it failed with `{"id":"draft-1","baseRevision":3,…}` vs `{"baseRevision":3,…,"id":"draft-1"}`. No test depended on the numeric ids. The native recovery suite (which uses the same fake) stays green.

**Low (intended, no code change):** native no longer publishes "Keybinding settings changed. Refresh to inspect the current value." on a malformed `changed` notification. The shared store drops an undecodable broadcast silently, as the web always has; that message was a refresh affordance, which the native app must not show (Jesse's 2026-09-13 ruling: native auto-refreshes, never a "refresh to see the latest" prompt), and the next GET or well-formed broadcast carries the truth. Added to the PR body's behaviour deltas with that reasoning.

Merged `origin/main` (tip 3256aed7f, past 4cb24bdc2 and D23b) with `--no-ff`; one conflict, the plan status table, resolved as a union (D23b's row, then D5's). `git diff origin/main...HEAD --stat` is D5's 20 files.

Gates on this head, foreground: the re-export falsification first, alone (`shipped module unreachable from every published specifier: keybindingsStore`, restored), then `make test-api-package`, `make test-web`, `make test-native` (81 + 7 files, Metro bundle), `make test-web-browser` (6/6, first run), `make lint-package-imports`, frontend `npm run lint` + `npx biome ci ../../../appwire-client/typescript`, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, root `go test -short -count=1 .` — all green. Oracles unchanged: web 65 + `initialWiring` 2 + cheatsheet/holdhints/settings; native trio 36; package suite 41.

obra:
## Review round 7 disposition (775404afd)

CI 15/15 green. RoboRev: one Medium, one Low, both in `saveDraft`'s post-await sequence.

- **Medium, the storage cleanup runs even when `stillMine()` fenced the reply out**, deleting the durable `writeUncertain` checkpoint the store just claimed to hold → accepted. The fence must come before any side effect.
- **Low, `applyHubOverrides` runs outside the try/catch**, so a registry conflict on a hub-confirmed payload leaves `saving: true` and blocks every later edit → accepted.

Rounds 2 through 7 have each found one more ordering hole in this one function's post-await path. Round 8 restructures `saveDraft`'s post-await into a single ordered sequence (decode → fence → apply-with-finally → cleanup) and pins it with a table over fence-out points × outcomes, rather than patching the two instances.

obra:
## Round 8 fixes (931f3295c)

Closed the class: `saveDraft`'s post-reply path is now one ordered sequence with no side effect ahead of the fence, written as a four-step comment block in the function.

### The sequence

| step | what | why the order matters |
| --- | --- | --- |
| 1 FENCE | `if (!writeStillMine(generation, token))` return the current payload | `retirePayload` already published `writeUncertain` for a fenced write; the checkpoint on the port is the only durable record of that, so nothing may touch state, registry or storage. This was the Medium: the old code ran `removeIf` before the fence. Same order as `patchOverrides` (a reply that is not ours is discarded without interpretation) |
| 2 DECODE | `fromWireOverrides`; malformed / `loadError` / lower revision → `hubError: MALFORMED`, `saving: false`, `writeUncertain: true`, reject | hub-sourced failure rides `hubError` (round 4); the outcome stays unknown and the checkpoint stays |
| 3 APPLY | `applyHubOverrides(value, settled)` (or `setState(settled)` when a newer external revision landed), with `settled = { saving: false, writeUncertain: false, draftConflict }` published on EVERY path: a reconciler throw is caught, `hubError` set, the flags still settle, and the call rejects after step 4 | the hub confirmed the write, so `saving` ends and `writeUncertain` clears whether or not the registry can take the rules. This was the Low: the apply ran outside any catch and left `saving: true` |
| 4 STORAGE | `removeIf(checkpoint)` on a confirmed write, `persistDraft(..., writeUncertain: false)` when the proposal stays for review; a cleanup failure keeps the draft in view with `storageUnavailable` | only once the outcome is in state does the durable record change; a cleanup failure can never turn a confirmed write back into an unknown outcome |

Two choices to state. **Fence before decode**, not decode before fence as the round listed: decode has no side effect either way, but discarding an unowned reply without interpreting it is what `patchOverrides` already does, and it means a fenced malformed reply cannot reject a dead-generation call with a message about a hub it no longer talks to. **Settle-on-throw rather than a draft-path preflight**: a preflight (`introducedWarnings`) catches only the simulated class (unknown action, unparseable, reserved, conflict on the simulated map); a wedged registry throws past any preflight, so the settle is needed regardless, and adding a preflight would be a second rule on a path no registry host uses today. The reconciler throw rejects the call, as `patchOverrides`' does, with the draft released (the hub has the rules; the registry's failure is the `hubError`).

`patchOverrides` checked: fence → decode → apply inside try/catch with `hubError` and rethrow, no storage step. Already the same order; unchanged.

### Table test

`keybindingsStore.test.ts`, "saveDraft's post-reply sequence": five scenarios — fenced by the generation ending, fenced by support dropping, reconciler throws (the oracle's wedge: the draft drops the override while a foreign binding squats palette.open's default chord, so the restore throws inside the reconcile), malformed reply, success — each asserting `saving`, `writeUncertain`, the checkpoint on the port (`intact` vs released) and `hubError` right after the reply, then `draftConflict: false` and `editDraft` allowed after the host's settling read (new generation + refresh; `setSupport("supported")`; or a plain refresh). RED before the fix: both fenced rows (checkpoint deleted) and the reconciler row (`saving` stuck true).

Merged `origin/main` 833860668 with `--no-ff`, no conflicts; `git diff origin/main...HEAD --stat` is D5's 20 files.

Gates, foreground. On the merge head `2b9d96357`: falsification first and alone (`shipped module unreachable from every published specifier: keybindingsStore`, restored), `make test-api-package`, `make test-native` (81 + 7 files, Metro bundle), `make test-web-browser` (6/6, first run), `make lint-package-imports`, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, root `go test -short -count=1 .` — green; `web-lint` and the package `biome ci` failed on a formatting drift in the new table (a hand-formatted call chain), fixed as the final commit. On `931f3295c`: `make test-web` and frontend `npm run lint` + `npx biome ci ../../../appwire-client/typescript` — green; the six gates above do not read package test files and stand across that whitespace-only commit. Oracles unchanged: web 65 + `initialWiring` 2 + cheatsheet/holdhints/settings; native trio 36; package suite 46.

obra:
## Review round 8 disposition (3a9793197)

CI 15/15 green; /simplify applied. RoboRev: two Lows, no Medium or higher → merged under the Lows rule; both land in an immediate follow-up PR.

- **Low, an undecodable `keybindings/changed` payload is dropped with no diagnostic and no follow-up read**, so the native list never converges → accepted, with the fix the native ruling allows: no refresh text, but the store schedules its own read when a changed notification cannot be decoded while loaded, so the state converges without asking the user.
- **Low, `fromWireOverrides` accepts empty action ids and chords** that native's old boundary rejected → accepted; the non-empty checks move into the shared decoder so both hosts refuse them before they reach state.

Issue #1491, opened 2026-09-16
navigation invalidation: three residual encodings the D13 extraction left in place


obra:
Found by the /simplify altitude pass on the D13 PR (revalidator + invalidation core at `@evener/appwire-client/state/navigation`). Each is a behaviour change, so none went into that PR. D14 (the navigation store migration) is the natural home; whichever lands first should take them or close this.

1. **Two sequence trackers in one module.** `cmd/evener-hub/frontend/src/stores/navigation/store.ts` (the `evener/navigation/invalidated` handler, ~:763-778) tracks `lastSequence` in zustand, treats a duplicate/out-of-order sequence or a generation mismatch as `protocolError`, and on a gap forces loaded keys without applying targets. `state/navigation/revalidator.ts` exports `applyNavigationInvalidation` + `acceptSequence` with its own `lastSequence`, which resets the generation and applies targets even on a gap, and processes duplicates. The web store never calls the exported helper, so the revalidator's counter sits at 0 in production and the helper has only test callers. Native `navigationPages.ts` ignores duplicates outright: a third rule. Pick one before D14 gives the revalidator a second consumer.
2. **`shutdownConvergence.ts` `matchesSession`** is a payload predicate hand-rolled beside `matchesTarget`. Folding it onto the shared rule changes one input: today `all_loaded_projects` matches even when `ctx.projectKey` is null; through `matchesTarget` it would match only with a project key present. Decide which is right, then fold.
3. **`mobile-native/src/navigationReadback.ts` receipt loop** (~:96-105): the per-target `result.version.revision < (target.revision ?? 0)` throw is now implied by the `requiredRevision` check inside `read()` for every in-vocabulary target; it stays reachable only for a target outside the vocabulary (e.g. `section: "recent"`), which `resourceFor` still issues a read for. Drop the check and decide whether `resourceFor` should refuse such targets.

PR #1492, opened 2026-09-16
refactor(sdk): navigation revalidator and invalidation core at state/navigation (D13)

obra:
SDK migration phase D, row D13 (plan: `docs/superpowers/plans/2026-09-12-sdk-migration.md`). Extends the published subpath `@evener/appwire-client/state/navigation` with two modules and repoints every importer; no behaviour change in either app.

## What moved and what was extracted

- `cmd/evener-hub/frontend/src/stores/navigation/revalidator.ts` → `appwire-client/typescript/state/navigation/revalidator.ts`, body unchanged apart from (a) imports going relative to `./types` / `../../types.gen`, (b) `invalidate()` and `acceptSequence()` calling the shared rule below, (c) `isNavigationNotInitialized` moving back into the web store (it classifies a string only `store.ts` produces; the SDK never emits it). `revalidator.test.ts` moved with it and stays the oracle (unchanged except its imports).
- New `appwire-client/typescript/state/navigation/invalidation.ts`: the pure rule both apps had hand-rolled: `matchesTarget(key, target)`, `matchingTargets(key, targets)`, `requiredRevision(key, targets)` (the receipt/notification revision floor) and `isSequenceGap(last, next)`. New test `invalidation.test.ts` (RED first: the module did not exist).
- `targetBase` (types.ts) is deleted: its only caller was the matcher, and one switch over `target.kind` now does what two switches joined by a `Partial<ResourceKey>` did. Same inputs → same outputs; the matcher test pins the out-of-vocabulary and missing-scope cases.

## Overlap table (same inputs → same outputs), measured by reading both files at `f1a3d5957`

| Concern | Web `revalidator.ts` | Native `navigationPages.ts` (post-#1462) | Shared here? |
| --- | --- | --- | --- |
| target ↔ loaded resource key | `matchesTarget`/`matchesBase` (:430-445) | private `matchesTarget(t)` over params (:179-196); `navigationReadback.ts` `resourceFor`+field compare (:14-37, :66-77); `pinNavigation.ts` `kind === "pin_catalog"` filter (:150-155) | **Yes** → `matchesTarget`. Identical for every input native feeds (catalog, section, pin_catalog, pin_section, project_page keys); wildcard ↔ `project`/`project_page` via `isProjectResource` on both sides |
| revision a receipt obliges a resource to reach | `targetsSatisfied` per-target `loadedRevision >= revision` (:379-394) | `receiptRevision` max-of-matching (:197-204), `requiredRevision` accumulation (:161-162), readback floor, pinNavigation floor | **Yes** → `requiredRevision` (four native copies → one). Web keeps its per-target predicate (it also checks loading/stale/error) |
| sequence gap | `acceptSequence` `seq > last+1` (:168); `store.ts:770` inline copy | `p.sequence > this.sequence + 1` (:156) | **Yes** → `isSequenceGap` (three encodings → one) |
| duplicate / out-of-order sequence | `applyNavigationInvalidation`: processes targets anyway; `store.ts:766`: `protocolError` | ignored (`<=` return) | No: three different rules. Left as-is (behaviour), filed below |
| generation change | `resetGeneration`: every entry stale + restarted | reset `sequence`/`requiredRevision`; stale only via targets/gap | No: different mechanism |
| already holds the announced revision | none (re-reads; 304 makes it cheap) | `held` short-circuit (:164-170) | No |
| in-flight read superseded | per-entry `forceToken` + `rerun` | per-store `uncertain` + `notificationEpoch` retry | Analogous, not the same code |
| location force on every notification, waiters (`waitForTargets`/`waitForInvalidation`) | yes | none | Web-only |
| request/decoding/pagination loop | injected `NavigationRequest` callbacks | `load()` owns the read + merge | Different layer: D14's row |

Verdict: the two are different mechanisms (multi-entry coordinator over injected requests vs one paged reader) with a real but small pure core: the matcher, the revision floor and the gap arithmetic. That core is what this PR shares; `NavigationPages` keeps its store/UI-facing loop as a thin layer over it (constructor caches its `resourceKey`; `invalidate`, `refreshAfter` and `load` call the shared helpers). No native adapter for `NavigationRevalidator` yet: nothing native constructs one until D14.

## "Needs an injected scheduler" claim: refuted

`grep -n "setTimeout\|setInterval\|queueMicrotask\|requestAnimationFrame\|Date.now\|performance"` over `revalidator.ts` returns nothing. The only deadline in the navigation store is the waiter timeout in `store.ts` `awaitNavigationConvergence` (`setTimeout` at store.ts:96, `NAVIGATION_INVALIDATION_TIMEOUT_MS`), which stays in the web store. The revalidator schedules nothing; it starts requests synchronously from `raise`/`load`/`resetGeneration` and settles through promise chains. The plan row's risk note does not match the file; the worksheet's reading was right.

## Importers rewritten

- Web (2): `stores/navigation/store.ts` (revalidator imports fold into the existing `@evener/appwire-client/state/navigation` block; inline handler uses `isSequenceGap`), `stores/navigation/shutdownConvergence.ts` (`NavigationInvalidationWaiter` type). `AppShell.tsx`, `AppShell.test.tsx`, `store.test.ts` and `shutdownConvergence.test.ts` only mention the word in comments/messages; no import to change.
- Native (3 files changed, 0 import paths changed): `navigationPages.ts` (matcher, floor, gap → shared helpers; `resourceKey` public readonly), `navigationReadback.ts` (floor → `requiredRevision`; `resourceFor` stays for the readback loop, which maps targets to read params), `pinNavigation.ts` (floor → `requiredRevision(pages.resourceKey, …)`). The other seven `navigationPages` importers are untouched.

One documented no-op: `navigationReadback`'s old filter excluded `all_loaded_projects`; `matchesTarget` includes it for project keys, but the hub never puts a revision on the wildcard (`appwire/types.go:2782` rejects one), so the floor is unchanged. Likewise the shared matcher would name `manifest`/`project`/`location` pages if a `NavigationPages` were ever built over them; no caller does (all are catalog, section, pin_catalog, pin_section, project_page).

## Publication

`state/navigation/index.ts` re-exports both modules; `tsconfig.build.json` `files` lists them (arms the reachability assertion); the `./state/navigation` qualification entry smoke-calls `matchesTarget` and constructs a `NavigationRevalidator`, with a `NavigationInvalidationWaiter` type use; README subpath section updated. Falsified: deleting the barrel's `export * from "./revalidator"` makes `make test-api-package` fail with `shipped module unreachable from every published specifier: state/navigation/revalidator`; restored with `git diff | git apply -R`.

## /simplify

Four angles run; applied: one-switch matcher and `targetBase` deletion (altitude), pinNavigation's fourth copy of the floor folded (reuse), `receiptRevision` wrapper inlined and `requiredRevision` no longer re-filters an already-matched list (simplification/efficiency), `isNavigationNotInitialized` back to the web store (altitude), smoke trimmed to one call per module and two redundant test rows dropped. Skipped: a `targetMatcher(target)` predicate builder to avoid re-resolving the target per loaded entry (GC-trivial, N is tens). Behaviour-change residuals are filed as an issue (linked in a comment).

## Gates (all foreground, on the head)

`make test-api-package`, `make test-web`, `make test-native` (Metro bundle exported; 83/789 native, 7/784 shared), `make test-web-browser` (all six guards, skillguard on first run), `make lint-package-imports`, `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend`, `go test -count=1 -run 'TestScenarioSource|TestPackageImportPaths' .` at the root.


obra:
Behaviour-change residuals from the /simplify altitude pass (three sequence-rule encodings between store.ts, the revalidator's exported helper and native; shutdownConvergence's matchesSession; navigationReadback's shadowed per-target check) are filed as #1491 for D14 or a follow-up, not this PR.

PR #1494, opened 2026-09-16
fix(sdk): tasks-panel store waiter ordering and test reset; adapter reconnect test (follow-up to #1471)

obra:
Follow-up to #1471 (D3), closing RoboRev's four Lows on that PR. Each fix was written test-first.

1. **Waiter ordering (`taskPanelState.ts` `refresh`/`runLoop`).** A `TasksListRead` that throws synchronously completed the whole run before `refresh` had pushed the caller's resolver, so that promise never settled. `refresh` now registers the caller before starting the loop, and `runLoop`'s cleanup (deregister the run, settle its waiters) runs on every exit: a thrown `hasAggregate` rejects the callers instead of stranding them. Tests: "a read that throws synchronously still settles the caller and clears the run", "a hasAggregate that throws does not strand the run".
2. **`resetForTests` and runs in flight.** It cleared entries and the fetch-id counter but not `runs`, so a read still in flight kept looping into the reset store and later waiters resolved against a stale run. It now settles every outstanding waiter (null) and clears `runs`; a run that comes back from its read and finds itself deregistered publishes nothing. Test: "resetForTests settles the waiters of a run in flight and stops it looping into the reset store".
3. **Native reconnect retention coverage.** The deleted `taskList.test.ts` held the only test of "retains loaded rows when a replacement connection cannot refresh them". The smaller option: the read-through-the-current-client indirection in `TasksSheet.tsx` is now its own unit, `mobile-native/src/tasksRead.ts` `tasksReadThroughCurrentClient(current)`, with that test in `tasksRead.test.ts` (swap in a rejecting client, assert rows retained + failure sentence + not loading). `TasksSheet.tsx` calls it.
4. **Plan row D3** no longer says the triple is hand-rolled; it states the store builds on `createFrameworkFreeStore` and names this follow-up.

Gates, foreground: `make test-api-package`, `make test-web`, `make test-native` (exit 0), `make lint-package-imports`, `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from the frontend dir. `mobile-native/` was not formatted with biome (no config there); the new file follows the directory's style by hand.


obra:
## Refresh onto c010784df + the loading-flag fix folded in

- Merged `origin/main` (c010784df, D20) with `--no-ff`; the only conflict was the plan status table, resolved as a union (main's rows including D20, plus this PR's D3 wording; main's own D3 cell was unchanged from the base).
- Folded in the last Low rather than opening another round (f71f1a9c6): when a host port throws inside a run (`hasAggregate`), `runLoop`'s error path now publishes a failure for the fetch `beginFetch` marked loading, so `TasksPanel.tsx` shows the error with its retry instead of spinning until an unrelated trigger. RED first: "a thrown hasAggregate settles the entry as a failure, not a spinner" (asserted `loading: true` before the fix).
- Gates rerun in the foreground at f71f1a9c60262da013bed8e28b0ecea7d99dd25a: `make test-api-package`, `make test-web`, `make test-native` (exit 0), plus `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from the frontend dir.

PR #1495, opened 2026-09-16
evener: the wedged-delegate run checks the close tree off deadlines, not wall time (fixes #1493)

### LINKED ISSUE #1493: main red: TestRunExitsWhenADelegateIsWedgedInAnUncancellableToolCall — Close spent 3.9s after the drain returned (ceiling 2.5s)
Main CI at ea067f0d1 (#1484, a mobile-native-only change) failed the root `tests` job (run 35076359551):

```
--- FAIL: TestRunExitsWhenADelegateIsWedgedInAnUncancellableToolCall (5.76s)
    run_drain_wedged_delegate_test.go:240: Close spent 3.901053061s after the drain returned (ceiling 2.5s); it consumed the joins' half of the close budge…
FAIL	primeradiant.com/evener/cmd/evener	41.918s
```

The commit touches no Go, so either the test's timing ceiling is load-sensitive on the runner or the close/drain budget changed under #1329 (0eb34d94e, steering carrier redesign — drained steering leg runs before the input ends) which landed two commits earlier and passed its own run. Rerun requested. Needs a ruling by measurement: does the post-#1329 close path consume more of the joins' half of the close budget after a drain (mechanism), or is the 2.5s ceiling too tight for a loaded runner (test)? Fix the mechanism if it is one; do not just raise the ceiling.

obra:
Fixes #1493 (main red at `ea067f0d1`: `TestRunExitsWhenADelegateIsWedgedInAnUncancellableToolCall`, "Close spent 3.901s after the drain returned (ceiling 2.5s)", run 35076359551).

## Measurement

Same host, 30 runs each, the two runs concurrent so they saw the same load, the test instrumented locally to log the Close-after-drain elapsed:

| build | n | min | median | p90 | max | failures |
|---|---|---|---|---|---|---|
| post-#1329, `7941839e0` (main) | 30 | 1.513s | 1.519s | 1.523s | 1.526s | 0 |
| pre-#1329, `e0d7cab73` (`0eb34d94e^`) | 30 | 1.513s | 1.518s | 1.536s | 1.548s | 0 |

Identical to the millisecond. `Session.Close` touches no steering state (`steeringParked`, the carrier release and the steering table run at the settle, the wake, a Stop and restore -- none on the close path), and the floor of ~1.51s is the hopeless stop join's reserved half of the 3s budget by construction. The CI reading of 3.901s exceeds the whole 3s cascade budget, which no change in what runs between the drain and Close can produce; only runner load stretching the expiry and the wakeups behind it can.

## Ruling

Load against a wall-clock ceiling, not #1329. The ceiling was not a performance bound: the test's own comment names it the close tree's wiring check -- that `closeRuntimeTree` joins the hopeless stop through `closeStopJoinContext`, bounded to its half of the budget, rather than on the whole cascade (the #420 shape). Wall time is the wrong instrument for that and had flaked once before (2026-09-08, 2.136s).

## Fix (test-side; the ceiling is not raised)

- `closeStopJoinContext` reports each stop join it bounds under a cascade deadline through `agent.ObserveCloseStopJoin(stopDeadline, cascadeDeadline)`, exported for this test the way `LaneClosePassBudget` and `DrainStallTimeout` are, nil in production.
- The test records every observation and asserts, off the two deadlines the real close minted: at least one stop join was minted after the drain returned, and every one reserved half of what the cascade had (`LaneClosePassBudget/2` from a fresh budget, half the remainder for a nested close) for the joins behind it, within 250ms of mint skew. RED with the helper taking the whole cascade locally ("a close stop join reserved 0s of the cascade's remaining 2.989s ... want 1.494s"), GREEN as shipped. The 90s tripwire stays as the hang guard; the wall-clock elapsed is logged for diagnosis, not judged.

## Gates

gofmt (toolchain) clean; `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz` on the root module (and `go vet` on the agent module, which hosts the hook); `go test ./cmd/evener/ -run Wedged -count=10 -race` ok.

## Round 3 fixes

Head `3de65466c`. One Medium, accepted: the check wanted `min(L/2, remaining/2)` reserved behind the stop join, but `closeStopJoinContext` halves the remainder only below L/2, so for remaining in [L/2, L) the stop takes the full L/2 and the true reservation is `remaining - L/2`; a correctly wired close whose preamble had spent more than ~0.5s of the 3s budget failed the check -- load sensitivity through a different door.

Fix: `agent.CloseStopJoinBudget(remaining)` is the one rule, exported, called by the helper and by the test's `reservedItsShare` (`want = remaining - CloseStopJoinBudget(remaining)`), so the two cannot drift; the 250ms mint-skew tolerance stays. `TestCloseStopJoinReservationFollowsTheHelpersRule` pins the shapes -- fresh cascade, a 1s preamble (reserves 500ms), under L/2 left, and the two #420 shapes -- and was RED under the old formula on the preamble case ("reserved 500ms, want 1s, ok false").

Gates, in the foreground: gofmt (toolchain) clean; full `make lint` green (ten gates, lint-golangci included); `go test ./cmd/evener/ -run Wedged -count=10 -race` ok; the agent module's CloseStopJoin tests ok.


obra:
## Review round 2 disposition (51f02fb44)

CI 15/15 green. RoboRev: one Medium.

- **The test's expected reservation `min(L/2, remaining/2)` does not match `closeStopJoinContext`, which halves `remaining` only when `remaining < L/2`; for `remaining ∈ [L/2, L)` the stop gets the full `L/2` and the reservation is `remaining - L/2`, so a close preamble over ~0.5s fails a correctly wired close** — accepted. The assertion mirrors the helper's actual rule (`budget = L/2; if remaining < budget { budget = remaining/2 }; want = remaining - budget`), so the check is again independent of host load, which was the point.

Round 3 with the implementer.

Issue #1507, opened 2026-09-16
Credential store: noteLandedMutation on auth writes reuses a row-ownership invariant that does not hold for them


obra:
Surfaced by the altitude review of D8 PR 1b. Pre-existing web behavior, now inside `appwire-client/typescript/state/credentials/instances.ts`.

`landedMutations`' contract (see `keepWrittenRows`) is "the store's held row is the write's own answer, newer than any read that started before it". `authMutation` calls `noteLandedMutation(provider)` after a landed auth write, but an auth write installs no row: a `fetch()` issued after the write's issue and resolving after it lands keeps the pre-write held row over the answer's row, transiently, until the store's scheduled self-refresh lands. The part that fits is retiring that instance's in-flight `refreshModels`.

Options: drop the call and let the self-refresh reconcile, or retire `refreshVersions` for that instance directly if that retirement is what is wanted. The web `credentials.test.ts` oracle pins the current retirement; check which tests move.

Issue #1506, opened 2026-09-16
Credential store: a lost reply to an instance write leaves both apps stale; native alone re-reads


obra:
Surfaced by the altitude review of D8 PR 1b (credential mutations in `@evener/appwire-client/state/credentials`).

"A failed reply can still follow a successful server write; read to reconcile" is a wire-truth rule, but today only native applies it: `mobile-native/src/providerInstances.ts` `mutate()` issues `core.getState().fetch()` in its catch. The web gets no such reconcile. For auth writes the core already covers it (a failed reply retires the own-echo marker, so the hub's `evener/auth/updated` reads as foreign and refetches). For `evener/instance/{create,edit,remove,setDefault,setModelDisabled}` there is no broadcast, so a lost reply leaves the web listing stale until something else reads.

Options: (a) the core's `applyMutation` catch schedules a foreign `scheduleRefetch()` and native's catch goes; (b) leave it. Before (a): check `ProviderConnection.tsx`'s invalidation watch, since a refetch after a refused instance write is a foreign-marked transition the web's guided flow may read as "Connection or configuration changed".

obra:
Fixed on main by the hub echo path: #1834 (merge commit `3571edf00a22f9c38055bb02a18abfd8406eb428`) made every *applied* instance write broadcast, and #1770 (merge commit `04898649d303d2f4fca821729b588bc5afb34312`) echoed `originClientId` on that broadcast so the originating client recognizes its own change.

Verified: a failed reply no longer leaves the web listing stale.

- `cmd/evener-hub/app_rpc.go` routes all six instance mutations (create/edit/remove/setDefault/setModelDisabled/refreshModels) through `instanceWrite(...)` (:1217). On `writeDidApply(err)` it calls `notifyInstanceUpdated(server, originClientId)` for a clean apply and `notifyInstanceUpdated(server, "")` for an applied write whose later step failed (:1220-1224) — the case the issue is about, where the write landed but the caller's reply is an error or is lost.
- `notifyInstanceUpdated` BroadcastAlls `evener/auth/updated` (:1567-1569), the payload-agnostic channel.
[...]

PR #1483, opened 2026-09-16
refactor(sdk): credential instances listing core at @evener/appwire-client/state/credentials (D8, 1a)

obra:
SDK migration plan row **D8**, PR **1a of 2** (listing core). Plan: `docs/superpowers/plans/2026-09-12-sdk-migration.md` §Phase D; decisions 1-4 settled.

## What lands

- **Core:** `appwire-client/typescript/state/credentials/instances.ts`. `createCredentialInstancesStore()` builds a framework-free store (`createFrameworkFreeStore`, D30's triple plus `getInitialState`, so zustand's `useStore` binds unchanged) holding the listing: ordered `fetch`/`fetchSelf`, `applyMutation` ordering with `mergeNewerRows`/`keepRefreshedModels`/`keepWrittenRows`/`reconcileToggledModel`, `create`/`edit`/`remove`/`setDefault`/`setModelDisabled`/`refreshModels`, the stale-listing refusal (`StaleListingRefusal`, `isStaleListingRefusal`, `staleListingHeld`), the coalesced `scheduleRefetch`, and `connectionChanged(client, state)` as the app's connection entry point. Client port: `CredentialInstancesClient = Pick<AppwireClient, "request">` (the `activityList.ts` idiom). Zero runtime deps; no import of `ConversationClientLike`, React, zustand or either app.
- **Published** at the new subpath `@evener/appwire-client/state/credentials` behind a barrel, mirroring C1's `state/navigation`: `package.json` `exports`, `tsconfig.build.json` `files` (arms the reachability assertion), `qualify-package.mjs` `packageExports` entry with real smoke calls, README §Published subpaths, one alias in each resolver (`vite.config.ts` subpath-before-root, web `tsconfig.json`, `mobile-native/tsconfig.json`, `tsconfig.check.json`, `vitest.config.mts`; Metro needs no edit).
- **Web adapter:** `stores/credentials.ts` keeps its path and full export surface (`credentialsStore`, `useCredentialsStore`, `resetCredentialsStoreForTests`, `isStaleListingRefusal`, `StaleListingRefusal`, `staleListingHeld`, `CredentialsStoreState`). None of the 30 importers (10 non-test, 20 tests) changed. The auth RPCs and the `evener/auth/updated` echo markers stay here for 1b, added to the core's state through the `extend` option and bound to the core's `CredentialInstancesSeam` (`requireWritableClient`, `noteLandedMutation`, `scheduleRefetch`).
[...]

Issue #1481, opened 2026-09-16
Native ProvidersScreen reads the credential instances core directly; delete the ProviderInstances adapter


obra:
Follow-up to SDK migration row D8 (PR 1a, credential instances listing core at `@evener/appwire-client/state/credentials`).

**Today.** `mobile-native/src/providerInstances.ts` is an adapter: it drives its own `createCredentialInstancesStore()` instance for one client and projects the core's listing into the snapshot `ProvidersScreen.tsx` renders (`data: InstanceListResponse | null`, `busy`, `credentialTest`, `error` via `sessionActionError`). D8 kept that shape so `ProvidersScreen.tsx` / `ProviderEditor.tsx` changed only for imports and the 13-test `providerInstances.test.ts` stayed the oracle.

**Ask.** Rewrite `ProvidersScreen.tsx` (about 15 sites reading `state.data`, `state.busy`, `state.credentialTest`) and `ProviderEditor.tsx` to select from the core's state (`instances`, `availableProviders`, `diagnostics`, `userLayer`, `writesRefused`, `loading`, `error`, `listingFromPreviousConnection`) through `useSyncExternalStore` over the store triple, then delete the adapter and its test. The native-specific rules the adapter carries need a home first:

- one write at a time (`busy`) and "refuse configuration writes while `writesRefused`" - screen-level state, or a core addition if the web wants it too;
- a re-read after every write whatever its reply (the core applies the write's own answer instead);
- the sanitized `credentialTest` result (`safeCredentialTestResult`), which lands in the core with `testCredentials` in D8 PR 1b.

Sequence after D8 1b and D9, so the sign-in and credential mutations are already in the core when the screen is rewritten.

PR #1508, opened 2026-09-16
refactor(sdk): credential mutations and own-echo correlation in the AppWire package (D8, 1b)

obra:
SDK migration plan row **D8**, PR **1b of 2** (credential mutations). 1a merged as `cca445ea6` (#1483). Plan: `docs/superpowers/plans/2026-09-12-sdk-migration.md` §Phase D.

## What lands

- **Core** (`appwire-client/typescript/state/credentials/instances.ts`, same module: the mutations share the store's client, request ordering, landed-write counts and coalesced refetch, so a second module would have had to reach into every one of those closures): `setApiKey`, `setCredentialJson`, `clearStoredKey`, `logout`, `testCredentials`, plus the four sign-in RPCs `loginStart`/`loginComplete`/`deviceStart`/`devicePoll` (see "D9 scope" below), the `evener/auth/updated` refetch, and the per-provider own-echo markers. One `authMutation(provider, request, landed?)` helper carries the write gate, the marker arm/re-stamp/retire, the post-write self-marked refresh and the ordering bump. `ownClientId: () => string` is a **required injected dep** (D26 territory, not moved). The connection object's identity replaces the old `connectionGeneration` counter and is reassigned only on a real transition. Client port: `Pick<AppwireClient, "request" | "onNotification">`; the core attaches its listener in `connectionChanged`. Zero runtime deps.
- **1a's Low folded in:** issuing a credential write bumps `requestVersion`, so a listing read in flight cannot publish rows that predate the write; the store's own refresh lands the post-write rows and drops the retired read's `loading` flag. RED test: "a listing read in flight when a credential write is issued cannot publish after it".
- **`extend` / `CredentialInstancesSeam` deleted.** Plan §Temporary seams is back to two.
- **Web adapter** `stores/credentials.ts` (56 lines): the core created with `ownClientId`, the zustand hook, `connectionStore.subscribe(syncConnection)`, `resetCredentialsStoreForTests`. Export surface unchanged; **zero importer changes** (30 files); `credentials.test.ts` 92/92 unchanged as the oracle (it pins that the fingerprint key is omitted when none was captured, hence `withFingerprint()`).
- **Native adapter** `mobile-native/src/providerInstances.ts`: subscribes to its own core instance and projects `data`/`loading`/`error` (`listingChanged` + `listingOf` from the core), keeps `busy` and the sanitized `credentialTest`; its own load loop, notification handler and post-success re-read are gone (the core owns them). It stamps a per-process `native-…` identity so other clients attribute its changes by origin. Wired on first use, not in the constructor, so an instance React builds and discards without `start()` holds no listener. `ProvidersScreen.tsx`/`ProviderEditor.tsx` untouched. Oracle 13/13 with three adjustments: two auth-success tests wait for the core's coalesced refresh (`vi.waitFor`), the pre-mutation-read test's final listing is the write's own answer (the stale read still never wins), and exact-params assertions include `originClientId`.
- Tests: package `instances.test.ts` 15 (never-echo, the Low interleaving, own-echo self vs foreign marking, notification isolation between two stores, refetch cancel on connection change, plus 1a's). Shared `testing/deferred.ts` reused.

## Never-echo-a-secret test: coverage list

Sentinels: a distinctive API key and a credential-JSON blob with a refresh token. Console `log`/`info`/`warn`/`error`/`debug` spied for the whole test. Paths driven, with every state snapshot recorded through `subscribe`:
- success: `setApiKey` (with fingerprint), `setCredentialJson`, `clearStoredKey`, `logout`, `testCredentials`, `devicePoll` (pending tick);
- failure: `setApiKey` refused by the hub, `setCredentialJson` refused, `testCredentials` endpoint failure, a failed listing read while markers are armed, `setApiKey` refused by the stale-listing gate after a client replacement.
Asserted for every sentinel: absent from every recorded state snapshot and the final state; absent from each thrown error's `message` and `data`, from `errorText()`, `friendlyErrorMessage()` and `sessionActionError()` of it; absent from all console output. Positive check that the key did go on the wire (`evener/auth/apiKey/set` params). Native's oracle keeps its own never-retain assertions (`JSON.stringify(snapshot)`), now over the core.

## D9 scope note (please confirm)

The brief said "remove `extend`" and "the web adapter shrinks to the sign-in RPCs". Those conflict: without `extend`, `credentialsStore.getState().loginStart(...)` (4 production call sites, 3 test files) has no home unless the web grows a sibling module and a new marker seam. The four RPCs are ~60 lines and use the same marker helper, so they moved into the core here; **D9 becomes native-only** (`providerSignIn.ts`). Say so if you would rather they came out again.

## /simplify

Applied: `connection` reassigned only on a real transition (a redundant `connectionChanged` no longer orphans in-flight mutations; the web guard went with it); `ownClientId` required, native identity added, `origin()` gone; marker re-stamp folded into `authMutation`; `withFingerprint()` once; `listingChanged` exported beside `listingOf` (no field list in native); native wires lazily; `loading: false` only when set; `refresh` as an async arrow; `nowhere` without its unused param; echo comment reworded without web consumer names.
Skipped: a shared version-fence helper (three one-line fences); `retireInFlightReads()` refactor and a `listingEstablished` state field (out of scope); gating the notification refetch on `requestedList` (pre-existing web behavior, not a 1b regression).
Follow-ups filed: **#1506** (a lost reply to an instance write leaves both apps stale; native alone re-reads), **#1507** (`noteLandedMutation` on auth writes reuses a row-ownership invariant that does not hold for them). Screen rewrite + adapter deletion stays **#1481**.

## Gates (foreground, all green)

`make test-api-package` + falsification (barrel export removed → `shipped module unreachable from every published specifier: state/credentials/instances`; restored), `make test-web`, `make test-native` (Metro bundle, 781 + 784 tests, check, check:scripts), `make test-web-browser` (all six guards, skillguard first run), `make lint-package-imports`, `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend`, `go test -count=1 -run 'TestScenarioSource|TestPackageImportPaths' .`.


obra:
## Refresh onto e6fdf04f4

`git merge --no-ff origin/main` (no rebase, no force-push): main gained D19' #1485, #1504, #1505 and #1296 since `cca445ea6`. One conflict, the plan's status table (D19/D19'/D4 rows landed beside the D8 row), resolved by keeping main's rows and this PR's D8 row. `git diff origin/main...HEAD --stat` shows only this PR's 8 files.

Gates rerun in the foreground on the merged head, all green: `make test-api-package` (falsification once more: barrel export removed → `shipped module unreachable from every published specifier: state/credentials/instances`, restored via `git diff | git apply -R`), `make test-web`, `make test-native` (Metro bundle, 772 + 784 tests, check, check:scripts), `make test-web-browser` (all six guards, skillguard green on the first run), `make lint-package-imports`, `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from the frontend dir, `go test -count=1 -run 'TestScenarioSource|TestPackageImportPaths' .`.

obra:
## Review round 1 disposition (90ea27484)

CI: `lint-repository` FAILED (lines being pulled; round 2 covers it). RoboRev: one Medium, two Low; all accepted.

- **M native `ProviderInstances` no longer invalidates `credentialTest` on a foreign `evener/auth/updated`**: the core owns the subscription now and `project` never touches `credentialTest`/`testRevision`, so a test result (even one in flight) survives another client changing that provider's auth and keeps rendering — accepted; a real behaviour regression against the old `refresh()` path. The adapter must observe the core's listing change and invalidate the probe (bump `testRevision`, clear `credentialTest`) the way its own refresh does. RED: foreign auth change with a probe result shown.
- **L README / plan row say `ownClientId` is optional while the deps type requires it** — accepted; docs match the required signature.
- **L comments still name the deleted web helpers (`completeLocalAuthMutation`)** — accepted; point at `authMutation`'s landed branch.

Round 2 with the implementer.

obra:
## Round 2 fixes

Merged `origin/main` first (`55952d1dd`, `--no-ff`, clean).

1. **M, native credential probe vs foreign auth change** (`mobile-native/src/providerInstances.ts` `project`): a listing change from any source now does what the adapter's own `refresh()` does - bumps `testRevision` and clears `credentialTest` - so a shown result comes down and a probe still in flight is discarded when another client or the TUI changes the provider underneath it. RED first, both cases: "a foreign auth change clears a shown credential test result" and "a foreign auth change discards a credential test still in flight" (native oracle now 15/15).
2. **L, docs vs signature:** README §Published subpaths and the plan's D8 row now state `createCredentialInstancesStore({ ownClientId })` with `ownClientId: () => string` required (native stamps a per-process identity).
3. **L, stale comments:** the two references to the deleted `completeLocalAuthMutation` (`instances.ts` at the `SELF_ECHO_WINDOW_MS` note and in the own-echo block) now name `authMutation`'s landed branch.
4. **secret-scan:** the one finding was `API_KEY = "sk-fixture-SECRET-KEY-9f3a7c"` (`generic-api-key`). The sentinels are now `never-echo-sentinel-api-key-value` and a credential JSON whose refresh token is `never-echo-sentinel-refresh-token-value`, with the substring list reduced to the marker; every assertion is unchanged, including the positive wire check. No allowlist change. Verified with `gitleaks detect --no-git` pointed at `state/credentials/` and the native test file: no leaks. (Note: `make secret-scan` run from inside a `.claude/worktrees/` checkout scans ~0 bytes because the config's path allowlist covers that segment; CI scans the real tree.)

Gates rerun in the foreground on the new head, all green: `make test-api-package` + falsification (barrel export removed → `shipped module unreachable from every published specifier: state/credentials/instances`, restored via `git diff | git apply -R`), `make test-web`, `make test-native` (Metro bundle, 774 + 785 tests, check, check:scripts), `make test-web-browser` (all six guards, skillguard green on the first run), `make lint-package-imports`, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from the frontend dir, `go test -count=1 -run 'TestScenarioSource|TestPackageImportPaths' .`.

obra:
## Round 3 fixes (fcf33c233c408090717ae221d17eaf299251e1ad)

Merged `origin/main` first (`105e2c823`, `--no-ff`, clean). No behaviour changes; every contract test unchanged.

`appwire-client/typescript/state/credentials/instances.ts`
1. `remove()` uses `withFingerprint()` like its five siblings.
2. `authMutation` hands its callback the origin stamp (`request(client, origin)`, called with `{ originClientId: deps.ownClientId() }`); the six request bodies spread `...origin`.
3. The marker re-stamp is `restampLocalAuthMutation`, beside `noteLocalAuthMutation`/`retireLocalAuthMutation`: three named marker operations.
4. `noteLandedMutation` deleted; both callers use `bump(landedMutations, …)` (the `landedMutations` comment stays on the map).
5. One `settleLoading(version)` for `applyMutation` and `authMutation`, keeping the `getState().loading` guard.
6. `listingOf` is the spelled-out literal rather than `Object.fromEntries`; since `CredentialListing` derives from `LISTING_FIELDS`, a field added there fails to compile here, so the single source of truth holds without a loop.

`scripts/qualify-package.mjs`: the smoke's `.then(exit, match)` block is `assert.rejects(credentialStore.getState().fetch(), /no client connected/)`.

`mobile-native/src/providerInstances.ts`
7. One private `invalidateTest()` (bumps `testRevision`, returns `{ credentialTest: null }`) used by `project`, `refresh` and `mutate`.
8. `nativeClientId` uses `randomUUID` from `expo-crypto`; the uuid-or-Math.random fallback is gone. The test mocks `expo-crypto` like `nativeOrganization.test.ts`.

Tests
9. The five debounce-waiting tests use `vi.useFakeTimers()` + `advanceTimersByTimeAsync(300)` with an `afterEach(vi.useRealTimers)`; the native file runs in ~0.3 s.
10. The two foreign-auth-change tests share `foreignAuthChange(handlers)` and `answerProbeWith(io, probe)`; the earlier "refetches when auth changes during a read" test uses the same helper.

Doc: `reconcileListing.ts` now points at the core's `scheduleRefetch` in `@evener/appwire-client/state/credentials`.

Nothing disagreed with. Gates rerun in the foreground on this head, all green: `make test-api-package` + falsification (barrel export removed → `shipped module unreachable from every published specifier: state/credentials/instances`, restored via `git diff | git apply -R`), `make test-web`, `make test-native` (Metro bundle, 779 + 785 tests, check, check:scripts), `make test-web-browser` (all six guards, skillguard green first run), `make lint-package-imports`, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` plus `gitleaks detect` pointed at the changed test paths, `npm run lint` and `npx biome ci ../../../appwire-client/typescript` from the frontend dir, `go test -count=1 -run 'TestScenarioSource|TestPackageImportPaths' .`.

obra:
## Review round 3 disposition (fcf33c233)

CI: 12 green, `race-modules / agent` failed with `panic: test timed out after 10m0s` in the Go agent package. This PR is TypeScript-only (no Go touched), so the timeout is not this change; failed jobs rerun, triage on the flake below.

RoboRev: two Lows, no Medium or higher.
- **Low, the native test "refetches when auth changes during a read" is satisfied by its explicit `refresh()`, not the notification, and can flake past 250 ms** → accepted; fake timers, drop the explicit refresh, assert the notification's refetch applied the update.
- **Low, the qualification smoke's `assert.rejects` is neither awaited nor chained** → accepted; explicit `.catch` → `process.exit(1)` like the marketplaces smoke, so the check does not depend on Node's unhandled-rejection policy.

Both land in an immediate follow-up PR after merge (Lows rule).

Issue #1432, opened 2026-09-16
web reducer and mobile store settle a failed turn idle themselves; the daemon's turn_failed status frame already does


obra:
Found while fixing PR #1340 round 22 (the TUI half of this is fixed there).

Rounds 18–20 of #1340 made all three clients settle an active session idle on `turn/completed{status:"failed"}`, on the premise (kata s8x8) that a genuine failure is the one turn that never gets a `thread/status/changed`. That premise is false at HEAD: `agent/session_lifecycle.go` `endInputAtTurnFailure` emits `EventSessionEnd{Reason:"turn_failed"}` (pinned by `agent/session_lifecycle_test.go` `TestSession_GenuineTurnFailureNotifiesLiveSubscriberOfIdleStatus`), and the projector announces it as `thread/status/changed(idle)` with the capability set inline.

- `appwire-client/typescript/reducer.ts` `turn/completed`: the active branch sets `status: idle` on a failed stamp, and the non-active branch's `failedWithoutId` does the same.
- `mobile/src/state/conversation.ts` `turn/completed`: `failedActive` flips the status idle.

In both, the frame that follows makes the same transition and carries the capabilities, so the early settle is redundant rather than wrong (the TUI's was harmful only because its status handler keyed a capability refresh on the transition). Per the round-19 rule the status frame owns the status; the two early settles and their tests (`reducer.test.ts` "a failed active turn with no status frame behind it settles the session idle", "...no active turn id still settles...", the store's failed-turn tests) should go, with the tests rewritten to assert the settle on the frame the daemon actually sends.

PR #1522, opened 2026-09-16
refactor(native): thread-level notifications through the package reducer (D23b)

obra:
Slice 2 of 4 for SDK-migration row D23 (docs/superpowers/plans/2026-09-12-sdk-migration.md:213), after D23a #1519. Thread-level notifications route through the package's `reducer.applyNotification`; the store keeps the `ThreadModel` half of the conversation as the reducer leaves it and the display rows as they were. Item appliers (`item/*`, `warning`, `evener/steering/injected`) are untouched until D23c.

## Replay rule

The reread's snapshot is authoritative over every thread-level write that preceded its response, and there is no replay. AppWire orders a `thread/read` response at the snapshot cut and delivers frames in producer order on the one socket (`docs/superpowers/plans/2026-07-28-appwire-retry-safe-mutations-and-atomic-rejoin.md:19, 372-373`: "Discard buffered notifications at or before the snapshot cut; release later records in producer order after the response"), so a frame the store applied while the read was in flight is already folded into the snapshot that arrives, and a frame that arrives after the response lands on top through the normal path. Measured: neither the service's `open`/`readProjection` nor the store's `rehydrate` awaits anything after the read resolves (error paths included), so there is no window to buffer for — the web's `applyHydrationResponseCut` drops its buffer at the same point for the same reason. The five thread-field owner revisions, the per-field commit ternaries, and rounds 1–2's buffer-and-replay are all gone; each modelled an ordering the protocol excludes, which is why native had owner revs for only some fields without anyone noticing. The capability-refresh publication is a plain write (a reread in flight supersedes it; one landing after the commit writes on top).

## Wrapper shape (fencing ruling: thin wrapper, no reducer change)

- Generation/binding gates stay in the store (`conversationGen`, `RequestBinding`/`isBindingCurrent`).
- **Routing** adopts the reducer's `notificationTargetsThread`: a frame names its thread by `ref` when it carries one, else by `threadId`; a frame naming neither is dropped. Native required both to match when both were present and let an identity-less frame through.
- The reducer path is the notification switch's `default` arm after the unsupported-`item/*` resync check; the case labels are the one encoding of what native still applies itself. `mutationOwnerRev`/`errorOwnerRev` are not model fields and stay; `liveOwnedRevs`, `pageOwnedIds` and the truncation freeze belong to the item appliers and go in D23c.
- `evener/thread/resync` (coalesced rehydrate) and unknown `item/*` (rehydrate) stay native-only. The reducer is stamped with `Date.now()`.

## Applier table — what moved to the reducer

| Method | Native applier (deleted) | Reducer case | Delta |
| --- | --- | --- | --- |
| routing prologue | `threadId` AND `ref` must match; identity-less frame passes | `ref` if present, else `threadId`; neither → drop | **stated**: a frame with a matching `ref` and foreign `threadId` now applies; an identity-less frame is dropped (the hub always sends identity) |
| `evener/sandbox/escalation/requested` | filter + append; `approvalOwnerRev++` | `upsertPendingEscalation` in place | **stated**: a re-requested card keeps its position |
| `evener/sandbox/escalation/resolved` | filter by id | `resolvePendingEscalation` | none |
| `thread/status/changed` | `status`, `capabilities` if present; `capabilityOwnerRev++` | same + clears `activeTurnStartedAt` off-active, `failedToolCalls`, `capabilitySource` | none visible |
| `thread/queueChanged` | `queue` | same | none |
| `evener/task/updated` | five fields verbatim; `tasksOwnerRev++` | derives `remaining` when absent and `cancelled` present; omits undefined keys | **stated**: `remaining` derived |
| `evener/goal/updated` | `goal`; `goalOwnerRev++` | same | none |
| `evener/thread/name/changed` | `name` | same | none |
| `thread/model/changed` | `modelProvider`, levels `?? []`, `supportsReasoning ?? false` | same + `model` | none visible |
| `thread/reasoning-effort/changed` | `reasoningEffort` | same | none |
| `thread/vision-model/changed` | `visionModel` | same | none |
| `turn/started` | `activeTurnId`; `turnOwnerRev++` | appends the `TurnModel`, `activeTurnId` | none visible (`turns` populated for D23c) |
| `turn/completed` | clears `activeTurnId` if it matches; failed→idle net (#1432) | same status rule; settles/folds turn items | none visible in rows until D23c |
| `evener/notes/updated`, `evener/urls/updated`, `evener/job/*`, `evener/delegate/updated`, `evener/jobs/treeUpdated`, `evener/thread/modelRetry` | ignored | applied | model fields populate; nothing native renders them yet |

## Mechanism → behaviour test mapping (no scenario dropped)

| Mechanism deleted | Test that pinned it | Behaviour test that now covers the scenario |
| --- | --- | --- |
| `capabilityOwnerRev` (status frame / cap refresh / rehydrate commit) | "concurrent newer cap refresh wins — rehydrate preserves current caps" | "a capability publication that preceded the response yields to the snapshot" (same fixture, inverted expectation; frame gains the thread identity every hub frame carries) |
| `capabilityOwnerRev` | "normal resync changes caps — rehydrate commits projected capabilities" | same test, unchanged |
| `approvalOwnerRev` | `approvals.test.ts` "does not resurrect an approval resolved while an older snapshot is in flight" | same test; its second snapshot now omits the card the resolution retired (the hub's post-resolution cut), so it pins the same outcome under the protocol's ordering |
| `approvalOwnerRev`, `goalOwnerRev`, `tasksOwnerRev`, `turnOwnerRev` (only the commit ternaries) | none | the `it.each` "commits the snapshot over  that preceded the response" rows for approval request, goal, tasks, started turn |
| native both-must-match routing | "drops vision changes from another session or thread" | "drops vision changes that name another session or thread" (foreign ref; threadId only; neither) + "routes a vision change by its ref when it carries one" |
| rounds 1–2 buffer-and-replay (this PR's own, deleted in round 3) | round-1 "keeps a status change / queue change / rename / model change over the snapshot" (four scenarios modelling the excluded ordering) | the same four fixtures as `it.each` rows with the inverted expectation: the snapshot commits |
| — (the ordering the protocol allows) | — | "applies a frame that arrives after the response on top of the snapshot" |
| — (fences on non-idempotent frames) | — | "does not replay a started turn the snapshot already carries", "does not replay a steering injection the snapshot already carries" |
| — (rehydrate ownership across a conversation switch) | — | "never commits a dead read's snapshot after the conversation closes" |
| owner revs via `goalCommand.test.ts` | "preserves newer live work state across an older hydration…", "preserves the advertised/turn active turn across stale hydration and older completion" (×2), "does not resurrect a turn that completes during a read before its start was observed" | same four tests, unchanged assertions; their snapshots now carry the frames that preceded the response (the hub's post-cut snapshot) instead of a pre-frame clone |

Fixture realism, not scenario changes: the fake service returns the conversation hydrated under the ref it was opened with (as the real service does), and three `warning` fixtures gain `threadId`/`ref`.

1. Routing: `ref` wins when present; identity-less frames are dropped.
2. Escalation re-request keeps its card position (in-place upsert).
3. `evener/task/updated` derives `remaining` when the frame omits it.
4. The reread's snapshot is authoritative over every thread-level frame that preceded its response (the web's rule); native's owner revs had kept a pre-response capability, approval, goal, turn or task write over the snapshot, which under the protocol's ordering is the older value.
5. `turn/started` appends the turn to `turns`; `turn/completed` settles it there. Not rendered until D23c re-projects rows from turns.
6. Package-only frames (notes, urls, jobs, delegates, tree, model retry) populate model fields native does not render yet.

None is a native regression: nothing the phone user had is lost.

## Gates

`make test-native` green (Metro bundle exported; native 777/777, shared 795/795, tsc, script imports), `make lint-package-imports` PASS, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` PASS. Nothing under `appwire-client/typescript` changed (`git diff --stat 1886f097e -- appwire-client` empty), so `test-api-package` does not apply. Gates re-run on the merged head after the simplify pass and the RoboRev fold-in: native 779/779, shared 798/798.


obra:
## Simplify fixes (`afe7bf71e9f8b3320a85edc298d46a6c4e9c6dbf`)

Applied, no behaviour change (the replay rule and reducer routing stay as pinned):

1. The reducer path is the notification switch's `default` arm (after the unsupported-`item/*` resync check); `NATIVE_ITEM_APPLIERS` and `ITEM_NOTIFICATION_METHODS` deleted — the case labels are the one encoding of what native still applies itself.
2. `applyThreadNotification` returns the reducer's result as the conversation (every reducer case spreads its input; no second whole-model copy per frame or per replayed write).
3. `ConversationStoreOptions.now` dropped; `Date.now()` at the reducer call, as the web does.
4. Buffer ownership follows the rehydrate token: `rehydrateReplay = []` on entry, taken and dropped at commit after the `token !== rehydrateToken` check, cleared on every conversation transition beside `trailingReread` (open/openProjected/suspend/resume/close/reset) and on a failed read; the `finally` and the takeover comment are gone. New test: `close()` under an in-flight reread, then a second conversation's frames neither buffer for nor land on the dead read.
5. One idiom at both live-write sites: build the `LiveWrite`, apply it, push it only while a rehydrate is in flight.
6. Stale owner-revision comment deleted.
7. The eight replay tests are one `it.each` table over `rereadRacing` (the model case reads a tuple).
8. The two capability tests drive `rereadRacing` (which now takes the initial thread).
9. `FakeConversationService.hydratedUnder(ref, conversation)` stamps the opened ref at one place.

Plan row: the accepted sentence on package-only frames re-rendering whole-store subscribers until D24 is added.

**RoboRev round 1 (`3460871`) folded in** — the one Low: a replayed `turn/started` for a turn the snapshot already carries took the reducer's duplicate-turn path (false invariant report; the snapshot's turn replaced by the sparse start frame). `replayedWrite(n)` makes that replay a no-op; pinned by "does not replay a started turn the snapshot already carries" (RED before the fix: 0 items survived; asserts the item survives and `console.error` is never called). No reducer change.

Merged `origin/main` (`46b518fc9`) with `--no-ff`, no conflicts. Gates on this head: `make test-native` green (Metro bundle; native 779/779, shared 797/797, tsc, script imports), `make lint-package-imports` PASS, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` PASS; package untouched.

obra:
## Review round 1 disposition (afe7bf71e)

CI: 13 green, `tests` failed on `TestSkillCatalogInspectionCopiesFullMetadata` in the Go agent package. This PR touches only `mobile/src` and the plan doc, so the failure is not this change; failed job rerun, triage in progress.

RoboRev: one Low, no Medium or higher.
- **Low, the replay guard covers `turn/started` only, while `evener/steering/injected` is also append-style and a frame the snapshot already folded gets appended twice at commit** → accepted and fixed in place rather than as a follow-up, because the right fix is the general one: the web drops buffered frames at the response cut (`applyHydrationResponseCut`) instead of guarding per method. Round 2 measures whether the native read response carries the cut and adopts that rule; if it cannot, the guard becomes "skip any buffered frame the snapshot already reflects" with steering in the `rereadRacing` matrix.

obra:
## Round 2 fixes (`2cecbe2a08f4ba42278fff587c6a58d00c0efec5`)

**Measurement.** The web's cut (`stores/threads.ts` `applyHydrationResponseCut`) is not a wire field: it drops the whole buffer when the read response arrives, on the protocol guarantee that the response is ordered at the snapshot cut (`docs/superpowers/plans/2026-07-28-appwire-retry-safe-mutations-and-atomic-rejoin.md:19,372-373`: "Discard buffered notifications at or before the snapshot cut; release later records in producer order after the response"). The wire `Thread`/`ThreadReadResponse` carry no sequence or revision (`Thread.updatedAt` is seconds). Native has the same cut information — stream order — and neither the service's `readProjection` nor the store's rehydrate awaits anything after the read resolves, so every write native buffers is *before* the cut. Under the literal rule the replay buffer would always be empty at commit, which flips the ten pinned "a live write during a reread wins" scenarios (they model an ordering the protocol excludes) and makes buffer-and-replay moot. That is a design call for you, not a round-2 fix; flagged below.

**Fix (your stated fallback, at one altitude):** `snapshotReflects(model, n)` — whether the snapshot already carries what a frame *appends* (the turn a `turn/started` opens; the steering item an `evener/steering/injected` adds, by `clientMutationId` when present, else kind + text + start time). Such a frame replays as a no-op; field writes replay freely (idempotent). The `turn/started` special case is subsumed and deleted; its test stays green through the general predicate. New test in the reread matrix: a steering injected during the read is already in the snapshot's turn → exactly one steering item after commit, and the next live steering takes index 1 (RED before the fix: two items, index 2). No reducer change.

**Recommendation for your decision:** adopt the literal cut rule in D23c (drop the buffer at the response; delete buffer-and-replay), rewriting the ten scenarios as "the reread's snapshot is authoritative over frames that preceded it". Until then the identity predicate keeps native correct for append frames and harmless for field frames.

Merged `origin/main` (`3cbfaebfc`) with `--no-ff`, no conflicts. Gates on this head: `make test-native` green (Metro bundle; native 779/779, shared 798/798, tsc, script imports), `make lint-package-imports` PASS, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` PASS; package untouched.

obra:
## Round 3 fixes (`516dfe89e072d0277d26b427c1dc6b9029a21c21`)

**Design call applied: the literal cut rule, now.** The reread's snapshot is authoritative over every thread-level frame that preceded its response; nothing replays. AppWire orders the `thread/read` response at the snapshot cut and delivers frames in producer order (`docs/superpowers/plans/2026-07-28-appwire-retry-safe-mutations-and-atomic-rejoin.md:19, 372-373`), cited in the store comment beside `applyThreadNotification` and in the PR body under **Replay rule**.

**Await audit (your precondition):** exactly one `await` in each of the service's `open` and `readProjection` (the read itself) and one in the store's `rehydrate` (`readProjection`); the `catch` paths await nothing. No window between response and commit, so the buffer is deleted rather than emptied at the cut.

**Deleted:** `LiveWrite`, `rehydrateReplay`, `snapshotReflects`, `replayedWrite`, the replay at commit, the buffer clears on the six transitions and on a failed read. The capability-refresh publication is a plain write with the comment you asked for (a reread in flight supersedes it; one landing after the commit writes on top).

**Tests:** the ten "live write during a reread wins" scenarios keep their fixtures with the expectation inverted (`it.each` "commits the snapshot over $name that preceded the response" ×8, plus the two capability tests); the two non-idempotent fences (turn/started, steering/injected already in the snapshot) stay and are now trivially true; **new**: "applies a frame that arrives after the response on top of the snapshot" (the ordering the protocol allows). Four more pinned scenarios of the class surfaced in `mobile-native/src/goalCommand.test.ts` when the full native gate ran (I had run only the store and approvals files after the rewrite — my miss, the gate caught it): their snapshots now carry the frames that preceded the response, assertions unchanged. `approvals.test.ts`'s in-flight resolution feeds the post-resolution snapshot. Mapping table and plan row updated (D23b now: reread snapshot authoritative, no replay; the four round-1 RED scenarios modelled an impossible ordering, which is why native had owner revs for only some fields without anyone noticing).

**Also of this class, left alone and flagged for D23c:** the store's `openProjected`/`resumeProjected` buffer escalation frames during the *initial* read and apply them after (`buffered.push` in `state/conversation.ts`; pinned by `approvals.test.ts` "keeps resolutions delivered between the initial snapshot and its response"). Same wrong-by-design shape, pre-dates D23b; your call whether D23c or a follow-up retires it.

Merged `origin/main` (`a7d288509`) with `--no-ff`, no conflicts. Gates on this head: `make test-native` green (Metro bundle; native 785/785, shared 799/799, tsc, script imports), `make lint-package-imports` PASS, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` PASS; package untouched. Net delta of the branch vs `origin/main`: 4 files changed, 418 insertions(+), 373 deletions(-).

obra:
## Round 3 fixes, correction (`2949e9b2855c929620f334fa3827e67d55cfcd87`)

The previous push (`516dfe89e`) omitted `mobile-native/src/approvals.test.ts` — its fixture change was in the working tree the gate ran on but not staged (my `git add mobile …` missed the `mobile-native` path), so CI at that head would have failed "does not resurrect an approval resolved while an older snapshot is in flight". Committed now as `2949e9b2855c929620f334fa3827e67d55cfcd87`; the tree content is identical to what `make test-native` passed on (native 785/785, shared 799/799), and the approvals file re-ran 17/17. Net delta of the branch vs `origin/main`: 5 files changed, 431 insertions(+), 380 deletions(-).

obra:
## Review round 3 disposition (2949e9b28)

CI 15/15 green; /simplify applied. RoboRev: one Low, no Medium or higher → merged under the Lows rule.

- **Low, native `item/completed` never reaches the reducer, so a live `modelRetry` is not cleared when a model-output item completes mid-turn** → accepted. Routing item frames through the reducer is D23c's opening step, so D23c is the follow-up that fixes it (with the turns populated live that D23c consumes).

Issue #1517, opened 2026-09-16
hub: evener/instance/updated echoes no originClientId, so a client's own instance mutations come back as foreign refetches


obra:
Found by the /simplify altitude pass on D8 1b (#1508). `cmd/evener-hub/app_rpc.go` `notifyInstanceUpdated` (~:1462-1475) broadcasts empty params, while `notifyAuthUpdated` echoes `originClientId`. The package credential core (`appwire-client/typescript/state/credentials/instances.ts` `consumeOwnAuthEcho`) therefore treats every create/edit/remove/setDefault the same client issued as a foreign change and refetches after already applying the full-list response; on the web that read also trips ProviderConnection's probe invalidation.

General rule to grow into: every mutation the core issues stamps `originClientId`, the hub echoes it on both notification paths, and the core keeps one marker keyed per mutation rather than per provider. Needs the hub change plus the generated param types (`make generate`), so out of scope for the TS-only D8 PRs.

obra:
Premise correction from #1532's simplify round: there is no `evener/instance/updated` wire name; `notifyInstanceUpdated` broadcasts `evener/auth/updated` with empty params, which the core already treats as a foreign change (no provider, no marker). The issue stands as written otherwise: the instance-path echo carries no `originClientId`, so a client's own instance mutations still come back as foreign refetches.

Issue #1516, opened 2026-09-16
credentials core: web and native adapters re-derive facts the core already knows (foreign listing change, listingEstablished, failed-reply re-read)


obra:
Found by the /simplify altitude pass on D8 1b (#1508); deferred from that PR because each item reaches into web component code or changes adapter behaviour.

1. **Foreign listing change is encoded twice with two different rules.** `mobile-native/src/providerInstances.ts` `project()` invalidates the credential probe on *any* `listingChanged` (including the core's own self-marked post-write refresh); `cmd/evener-hub/frontend/src/panes/settings/sections/credentials/ProviderConnection.tsx` (~:339-357) invalidates only when `selfRefresh` did not move, and also on loading/error flips. The core has the facts (`listingChanged`, `selfRefresh`). Export one predicate next to `listingChanged` in `appwire-client/typescript/state/credentials/instances.ts` (e.g. `foreignListingChange(state, previous)`) and have both hosts call it. Natural home: D9 (native providerSignIn adopts the core) or a small follow-up after D8 1b.
2. **`listEstablished` is private in the core and re-derived in native as `data === null`.** Expose it on `CredentialInstancesState` (e.g. `listingEstablished`) so native's `data` is a projection and the web gets the same fact.
3. **"A failed reply may follow a landed write, so re-read" lives only in native** (`providerInstances.ts` ~:119-123). The core now listens for the hub's auth/updated echo on both hosts, so either the echo is trusted and the native catch-path `fetch()` goes, or the reconcile belongs in `authMutation`/`applyMutation`'s failure branch for both hosts.

Pattern: unify the encoding rather than patch each instance (#1105 r9-11, #1137 r6-10 lesson).

PR #1524, opened 2026-09-16
refactor(native): the sign-in flow drives the shared credential store; authStatus on the core (D9)

obra:
SDK migration plan row **D9**, native-only after D8 1b (#1508) took the web's sign-in RPCs into the core. Scope accepted with option (a): `authStatus` added to the core.

## Core (`appwire-client/typescript/state/credentials/instances.ts`)
`authStatus(provider)`: a plain read of `evener/auth/status` through the read gate (available while the rows on screen are a replaced connection's), arming no echo marker and refreshing nothing; a reply that arrives after the connection it was asked on is gone is refused with the shared `CONNECTION_REPLACED_ERROR` words. RED first (two tests: gate/no-marker/no-refresh, and the replaced-connection refusal); the never-echo test also drives a status read. Qualification smoke gets a disconnected `assert.rejects(...).catch(...)` probe for it (the `fetch` probe takes the same chained shape, identical to #1521's line); README's subpath prose lists it. No plumbing change (no new module or subpath).

## Native
- **Shared store, hoisted.** The flow lives in `ProvidersScreen` and outlives a client replacement (`setConnection(client | null)`), while `ProviderInstances` is per client inside `Providers`, which unmounts whenever the hub is not ready; a store "from ProviderInstances" would die under the flow. So `mobile-native/src/credentialStore.ts` builds one `CredentialInstancesStore` per hub (identity from `expo-crypto`), `ProvidersScreen` owns it and its connection effect drives `connectionChanged(client, state)` (plus `connectionChanged(null, "closed")` on unmount), and both `ProviderInstances(store)` and `ProviderSignIn(store, name)` take it. `ProviderInstances` no longer builds a store or drives the connection; its projection still subscribes lazily.
- **`ProviderSignIn`:** the five request sites route through the store: `device/start` → `deviceStart`, `login/start` → `loginStart`, `device/poll` → `devicePoll`, `login/complete` → `loginComplete`, `status` → `authStatus`. The state machine is untouched (phases, the 60 s poll clamp, `uncertain`, `pendingOperation`, reply validation); `setConnection` stays as the connected gate and the identity token that tells a replaced connection from a same-client transition.
- **Tests.** Oracles hold: `providerSignIn.test.ts` (30), `providerSignInRecovery.test.ts` (3), `providerInstances.test.ts` (15). New RED tests: every RPC is routed through the store's client, not the `setConnection` token (device and browser paths); the poll and completion carry this app's identity; a start from a replaced connection's listing is refused and reads as "could not be started". Harnesses build a store and drive it with a `connect(client)` helper that mirrors the screen's effect; fake clients gain an `onNotification` stub. One `providerInstances` oracle now has the screen say the connection closed (the store's owner) rather than `dispose()`.

1. `login/complete` and `device/poll` carry `originClientId`, so other clients attribute the change by identity (two exact-params oracle assertions updated).
2. An authorized poll or a completion schedules the store's own listing refresh; today only the hub echo does (one oracle now advances past the 250 ms refresh before asserting no further poll).
3. A sign-in started while the listing is a replaced connection's is refused ("connection was replaced") and surfaces as the flow's "Sign-in could not be started" error (RED test).
4. Each poll tick retires an in-flight listing read (the core's documented ordering rule).

#1516 not needed here (the sheet reacts only to explicit calls); #1517 untouched.

## Gates (foreground, all green)
`make test-api-package` + falsification (barrel export removed → `shipped module unreachable from every published specifier: state/credentials/instances`, restored), `make test-web`, `make test-native` (Metro bundle, 782 + 786 tests, check, check:scripts), `make lint-package-imports`, `npx biome ci ../../../appwire-client/typescript` from the frontend dir, `gitleaks detect --no-git` on `state/credentials/` and `mobile-native/src` (no leaks) + `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, `go test -count=1 -run 'TestScenarioSource|TestPackageImportPaths' .`.


obra:
## Simplify fixes (b0331344f2544f658586257a38810011ac7e77ba)

Refreshed onto `46b518fc9` first (`--no-ff`; the one conflict was the smoke line #1521 also changed, resolved into item 4's form).

1. `mobile-native/src/credentialStore.ts` is the native adapter module: `useCredentialStore()` owns the per-hub `useMemo`, the `[store, client, state]` `connectionChanged` effect and the unmount-only `(null, "closed")` effect, with the two-effects reason and the "one store per hub; the flow outlives client replacement" rationale beside the code. `ProvidersScreen` is `const store = useCredentialStore()`.
2. `providerSignIn.ts`: the field is `connection: object | null`, an opaque token; the `../../mobile/src/services/conversation` import is gone; the class docstring and `setConnection`'s comment say what the token is. `providerInstances.ts` docstring names no caller module.
3. Core: one `connectionStillCurrent(client)` fence shared by `readListing` (both fences), `refreshModels` and `authStatus`; `authStatus`'s interface comment no longer restates the gate. `authMutation`/`applyMutation`'s `connection !== issued` left alone as asked.
4. Smoke: `Promise.all([fetch probe, authStatus probe]).catch(die)`, one catch.
5. Tests: "stamps this app's identity" deleted (the oracles pin the stamp); `providerSignIn.testkit.ts` holds `boundary(answer)`, `connect`, `authCalls` and `signInClient(io, calls)` (carries `onNotification`); both sign-in test files use it, the five hand-added stubs and the duplicated `connect` are gone.
6. **Measured, remount kept, read removed.** `Providers` holds `selected` (the instance detail Modal, where sign-in is launched), `configuration`, `editingCredential`, the typed `key` and `actionError`; the `revision` remount is what closes that Modal and clears a half-typed key when the sheet closes, so it is user-visible and stays. Before: an authorized sign-in cost two `evener/instance/list` reads (the store's own refresh after the landed poll, then the remounted list's `start() → fetch()`). Now `start()` publishes the rows the store already holds for this connection and reads only when it holds none or they belong to a replaced connection: one read. Pinned by two `providerInstances` tests ("start publishes the rows the store already holds … without a read": request count unchanged; "start reads when the rows … belong to a replaced connection").
7. Plan D9 row: one sentence that #1516's three items stay deferred.

**RoboRev on `2c3b514a1`, High, folded in:** the store was built during render and bound to the client only in the screen's effect, while the child's mount effect called `start()` first (React flushes child effects before the parent's), so on an already-connected hub the first read threw "no client connected" and the screen settled on the empty state. Fix: `createNativeCredentialStore(client, state)` binds the connection when the store is built (the hook's `useMemo` calls it), the effect follows later transitions. Pinned by `credentialStore.test.ts` (a store built for a ready connection reads through it at once; one built without a connection refuses until told one). RoboRev's mounting test is not possible today: mobile-native has no render or hook harness (no testing-library, no `.test.tsx`, node environment); filed as **#1526**.

No disagreements. Gates rerun in the foreground on this head, all green: `make test-api-package` + falsification (barrel export removed → `shipped module unreachable from every published specifier: state/credentials/instances`, restored), `make test-web`, `make test-native` (Metro bundle, 785 + 786 tests, check, check:scripts), `make lint-package-imports`, `npx biome ci ../../../appwire-client/typescript` from the frontend dir, `gitleaks detect --no-git` on `state/credentials/` and `mobile-native/src` + `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, `go test -count=1 -run 'TestScenarioSource|TestPackageImportPaths' .`.

obra:
## Review round 2 disposition (b0331344f)

CI 15/15 green; /simplify applied in this head. RoboRev: two Lows, no Medium or higher → merged under the Lows rule; both land in an immediate follow-up PR.

- **Low, `useCredentialStore` builds and binds the store in `useMemo`, so a discarded render leaks a notification subscription and duplicates refetches** → accepted. Note the constraint from this round's High: the child list's mount effect runs before the parent's passive effect, so binding cannot simply move into `useEffect`. The follow-up keeps one committed instance and binds in a layout effect, which runs before any child's passive `start()`.
- **Low, `start()`'s held-rows short-circuit publishes `data` only and hides a retained `error`/`loading`** → accepted; project them from the held state.

PR #1532, opened 2026-09-16
refactor(sdk): the credential store core owns the foreign-listing-change predicate, listingEstablished and the refused-write re-read

### LINKED ISSUE #1516: credentials core: web and native adapters re-derive facts the core already knows (foreign listing change, listingEstablished, failed-reply re-read)

obra:
Closes #1516. Package + web + native: a **serial-queue PR**, at the tail (D5 → D13 → D17'' → D11 → this). Refreshed onto `4cb24bdc2` (#1528; native `start()` unioned).

## Core (`appwire-client/typescript/state/credentials/instances.ts`)
1. **`foreignListingChange(state, previous, moved?)`**, exported beside `listingChanged`. The one rule, reconciled between the web's `ProviderConnection.tsx` (skip when `selfRefresh` moved, else invalidate on `instances`/`availableProviders`/`loading`/`error`) and native's `project()` (any `listingChanged`): `state.selfRefresh === previous.selfRefresh && (listingChanged(state, previous) || state.loading !== previous.loading || state.error !== previous.error)`; `moved` lets a caller that already computed `listingChanged` pass it. Tests (`test.each`): the store's own post-write refresh and a caller's `fetchSelf` are not foreign; a foreign echo's read, an applied write, and a failed read that moved no rows are.
2. **`listingEstablished`** on `CredentialInstancesState`: "a listing has been applied", set through the one door `landedListing()` that both apply sites use, cleared only by reset. A client change keeps it: the rows stay on screen as `listingFromPreviousConnection` until the new connection's read lands, as the core always intended. An empty listing counts. `refreshModels` treats "removed" only against this connection's own listing and still stages a replaced connection's unknown row.
3. **Refused writes rely on the hub's echo**, stated once: the hub broadcasts `evener/auth/updated` for every write it applies (`notifyInstanceUpdated`); an instance write's echo carries no origin and armed no marker, a credential write's marker is retired, so the echo reads as foreign and re-reads; a refused write that applied nothing changed nothing. No refetch of the store's own. Tests: a refused `remove` and a refused `setApiKey` read nothing until the echo, which reads once.
Qualification smoke probes `listingEstablished` and the predicate; README's subpath prose lists both, and says `writeUncertain`-style host-only facts stay with the host.

## Web
`ProviderConnection.tsx` calls `foreignListingChange` (imported through `stores/credentials.ts`'s re-export line); its phase gate stays web-side; its own four-field encoding is gone. `ConnectProviderDialog.tsx`'s "no listing landed yet" test reads `listingEstablished` (an applied empty listing no longer triggers a second read). Oracles `ProviderConnection.test.tsx`, `ConnectProviderDialog.test.tsx`, `credentials.test.ts` unchanged and green.

## Native (`providerInstances.ts`)
`project()` computes `listingChanged` once, publishes `data = listingEstablished ? listingOf(state) : null`, invalidates the credential probe on `foreignListingChange`; `start()` adopts an established, current listing (with its `loading`/`error`) and reads otherwise; `mutate()`'s catch-path `fetch()` is gone. Oracles adjusted to the rule: the refused `remove` and lost-reply credential JSON tests emit the hub's echo, which re-reads. New: a replaced connection keeps the rows on screen until its own read lands.

## Behaviour deltas
- **Native:** the store's own post-write refresh no longer invalidates a shown credential test result (only foreign changes and read flips do); a refused write's re-read comes from the hub's echo rather than the adapter; the previous connection's rows stay on screen across a reconnect instead of a spinner; an established empty listing is adopted on remount without a read.
- **Web:** none in behaviour. `ProviderConnection` invalidates on the same transitions as before; `ConnectProviderDialog` skips a redundant read after an applied empty listing.

## Gates (foreground, all green on the pushed head)
`make test-api-package` + falsification (barrel export removed → `shipped module unreachable from every published specifier: state/credentials/instances`, restored), `make test-web`, `make test-native` (Metro bundle, 787 + 786 tests, check, check:scripts), `make test-web-browser` (all six guards, skillguard first run), `make lint-package-imports`, `npx biome ci ../../../appwire-client/typescript` and `npm run lint` from the frontend dir, `gitleaks detect --no-git` on the four changed paths + `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, `go test -count=1 -run 'TestScenarioSource|TestPackageImportPaths' .`.


obra:
## Review round 1 disposition (3fedd43bd)

CI 15/15 green. RoboRev: one Low, no Medium or higher. The PR is CONFLICTING against main after #1528, and a /simplify round is already in flight, so the Low folds into that push rather than a follow-up.

- **Low, the `ProviderInstances` class doc still lists "a reconciling read after a write whose reply was lost"** → accepted. The simplify round drops the refused-write refetch from the core as well (the hub echoes every applied write, so the read carried no information), so both comments get rewritten to the one rule: refused writes rely on the hub's echo.

obra:
## Simplify fixes (2393137095136533283b6955c532d565ad5da7bc)

Refreshed onto `4cb24bdc2` (#1528; the one conflict was native `start()`, unioned: `listingEstablished` gate plus #1528's `loading`/`error` publish).

**Decision 1, applied:** `applyMutation`'s catch refetch is gone. Checked the premise against the hub: there is no `evener/instance/updated` wire name; `notifyInstanceUpdated` (`cmd/evener-hub/app_rpc.go`) broadcasts `evener/auth/updated` with empty params, which the core already treats as foreign (no provider, no marker), so the echo re-reads. The rule is stated once above `applyMutation`; `authMutation`'s "refused: no echo will follow" comment now says the marker retires so a late echo reads as foreign; native `mutate()`'s and the class doc's comments name the one rule (RoboRev's Low at `3fedd43bd`). Tests: "a refused instance write reads nothing of its own; the hub's echo is what reads", "a refused credential write reads nothing of its own; the echo, retired of its marker, reads once". The web delta is withdrawn from the body.

**Decision 2, applied:** `listingEstablished` is no longer cleared on client change; `landedListing()` is its one door (both apply sites, which spelled the same shape twice); `clearClientBookkeeping` is the three `Map.clear()` calls. One consequence handled: `refreshModels`' "a removal landed" guard now also requires the listing to be this connection's, so a replaced connection's unknown row is still staged as before. Native RED test: "a replaced connection keeps the rows on screen until its own read lands".

**Small items:** `ListingTransition` unexported; duplicate comment in `clearClientBookkeeping` gone; `start()` publishes `listingOf(held)` directly; `foreignAuthChange(handlers, provider)`; predicate scenarios as `test.each`; one predicate assertion in the smoke; the `listingEstablished` test on fake timers; `ProviderConnection.tsx` imports through the `stores/credentials.ts` re-export line; native `project()` computes `listingChanged` once and passes it (the predicate takes an optional precomputed `moved`). Optional item done: `ConnectProviderDialog.tsx:143` reads `!state.listingEstablished` in place of the two length checks, keeping the `loading`/`error` reuse; its oracle stays meaningful and green.

No disagreements. Gates rerun in the foreground on this head, all green (list in the PR body).

obra:
## Review round 2 disposition (239313709)

CI 15/15 green; /simplify applied. RoboRev: two Lows, no Medium or higher → merged under the Lows rule.

- **Low, the `listingEstablished` field comment says it resets per client while the code and test keep it across a client change** → accepted; a one-line follow-up rewords it to the invariant (it survives a client change; `listingFromPreviousConnection` says whose rows they are).
- **Low, hub rollback-failure paths in `Create`/`Remove` mutate the config and return an error without broadcasting, so no echo arrives** → accepted at the right layer: the fix is the hub broadcasting on every path that leaves the config changed, filed as https://github.com/prime-radiant-inc/evener/issues/1543. The client rule (refused writes rely on the echo) stands.

Issue #1531, opened 2026-09-16
FuzzRunCoverage forks 'go env' into a per-iteration HOME TempDir (same class as #1527)


obra:
Follow-up to #1527.

\`cmd/evener/run_coverage_fuzz_test.go\` sets \`t.Setenv("HOME", t.TempDir())\` per iteration (line 54) and constructs a real session through \`agent.NewSession\` (line 200). The session's launch snapshot forks \`go env\` (\`agent/session_capabilities.go\` probeCapabilities); the go command records telemetry under \`os.UserConfigDir()/go/telemetry\` and, on the first run in a config home with no upload token, daemonizes a Setsid sidecar that keeps creating files there after \`go env\` exits. With HOME a \`t.TempDir\`, that sidecar races the TempDir cleanup and can fail the test after it passed with \`TempDir RemoveAll cleanup: unlinkat …: directory not empty\` — the mechanism measured on #1527.

Inside package agent the fix was \`testConfig.skipGitSnapshot\` plus a source audit (\`agent/config_home_fork_audit_test.go\`) that pins the class. That gate is unexported, so this root-module test cannot use it. Options: an exported way to skip the launch snapshot for out-of-package tests, or a config home that outlives the test (the agent package's TestMain pattern), or a fixture that runs \`go telemetry off\` in the fresh home before the session starts.

Not seen failing yet; filed because the exposure is the same construct.

PR #1534, opened 2026-09-16
test(agent): session tests with a per-test config home skip the launch snapshot (go telemetry sidecar outlived TempDir)

### LINKED ISSUE #1527: flake: TestSkillCatalogInspectionCopiesFullMetadata fails on TempDir cleanup ('directory not empty'): a writer outlives the test
Seen 2026-09-16 on #1522 @afe7bf71e (a mobile-only PR), run 35100025197, `tests` job, agent shard 4:

```
--- FAIL: TestSkillCatalogInspectionCopiesFullMetadata (0.07s)
    testing.go:1617: TempDir RemoveAll cleanup: unlinkat /tmp/evener-module-tests.GmndOT/tmp/agent/TestSkillCatalogInspectionCopiesFullMetadata4283202329: directory not empty
```

The test passed; the cleanup failed because something was still creating files under its `t.TempDir()` after the test returned (a background inspection/copy goroutine or a watcher not joined before return). Main was green at 46b518fc9. Root cause is test-side or a missing shutdown join in the skill catalog inspection path; fix by waiting for the writer before the test ends (or by the production code exposing the completion the test can wait on), not by retrying cleanup. Rerun requested on the PR.

obra:
Closes #1527

## Root cause

`TestSkillCatalogInspectionCopiesFullMetadata` passed and then failed its `t.TempDir()` cleanup in CI: `unlinkat …/TestSkillCatalogInspectionCopiesFullMetadata4283202329: directory not empty`. The reported path is the *parent* TempDir with no child error recorded, which (per `os/removeall_at.go`) means a writer `MkdirAll`ed a path under one of the test's five sub-TempDirs after that subdir had already been removed, in the window before the parent `rmdir`.

The writer is the go command's telemetry sidecar, not anything in-process:

- The session's launch snapshot forks `go env` through the exec environment (`probeCapabilities`, `agent/session_capabilities.go`).
- The go command records telemetry under `os.UserConfigDir()/go/telemetry` — `$XDG_CONFIG_HOME` on Linux (CI), `$HOME/Library/Application Support` on macOS. On the first run in a config home with no `upload.token` it daemonizes a sidecar (`Setsid`, waited only in a detached goroutine; `cmd/vendor/golang.org/x/telemetry/start.go` `parent`/`startChild`) that creates `telemetry/upload/` and more after `go env` has exited.
- This test sets `HOME`, `XDG_CONFIG_HOME`, `XDG_STATE_HOME` to `t.TempDir()`s, so the sidecar gets a directory the test is about to remove. It is its own session, outside the process table `env.Cleanup()` reaps, and exposes no completion signal to the product or the test.

Measurements (macOS, go1.27.0, telemetry mode `local`):
- `XDG_CONFIG_HOME=<fresh> HOME=<fresh> go env GOCACHE`: `go/telemetry/local/{upload.token,*.count,weekends}` exist immediately; `go/telemetry/upload/` appears ~2s later, after the parent has exited.
- Temporary probe in the test (listing the temp tree after `Close()`, dumping goroutines): before the fix, six `go/telemetry/…` paths under the test's HOME TempDir at test end; zero live `agent` goroutines; zero files added in-process. After the fix: zero telemetry paths.
- The flake itself did not reproduce locally: `-count=50` and `GOMAXPROCS=2 -count=300 -p 1 -parallel 1` both green (the sidecar's write here lands well after the ~ms RemoveAll window; CI shard 4 ran 8 shards × `-parallel 3` on a loaded runner).

## Fix (test-side)

No completion signal exists to wait on, so the fix is not to fork: every agent test that isolates the config home per test and constructs a session now passes `testConfig.skipGitSnapshot` (its documented purpose: "a test whose contract sits below the environment-snapshot layer must not fork for it"). 13 tests across 5 files.

The class is pinned by a new source audit, `agent/config_home_fork_audit_test.go` (`TestSessionTestsWithAPerTestConfigHomeSkipTheLaunchSnapshot`, in the style of `deadline_audit_test.go`): any test-file function that sets `HOME`/`XDG_CONFIG_HOME` and reaches `NewSession`/`RestoreSessionFromMeta*` (directly or through test helpers, fixed-point over the test files) must name the gate (`skipGitSnapshot` or `withoutGitSnapshot`). RED listed the 13 tests; GREEN after gating.

Product side: unchanged. The sidecar is Go's normal behaviour in the user's real config dir; the capability preamble already documents the sandboxed-go telemetry denial.

Residual filed: #1531 (`FuzzRunCoverage` in the root module has the same construct and cannot reach the unexported gate).

## Gates

- `$(go env GOROOT)/bin/gofmt -l` on touched files: clean
- agent module: `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`, `golangci-lint run` (0 issues)
- `make lint` from the repo root: all PASS
- agent suite `go test -count=1 ./agent/`: ok (350s)
- fixed test `-count=100` and `-race -count=20`: ok
- root `go test -short -count=1 .`: ok


obra:
## Review round 1 disposition (0ae0517ca)

CI 15/15 green. RoboRev: two Lows, no Medium or higher → merges under the Lows rule with an immediate follow-up PR closing both audit escapes:

- **Low, `setsConfigHome` is not transitive** (a helper that isolates the config home lets the caller escape) → accepted; propagate an "isolates config home" predicate through the same fixed point as `reaches`.
- **Low, `namesLaunchSnapshotGate` clears on any identifier**, so a test that re-enables the snapshot (`skipGitSnapshot = false`) reads as gated → accepted; inspect the value passed to the constructor instead of scanning for the name.

PR #1540, opened 2026-09-16
refactor(native): the initial read is authoritative; a refused mutation rereads (D23c-1)

obra:
Slice c-1 of SDK-migration row D23 (docs/superpowers/plans/2026-09-12-sdk-migration.md:213), after D23b #1522. The read path's remaining pre-cut-rule machinery goes, and a refused mutation converges through the reducer instead of a bespoke capability read and write. Item appliers untouched (c-2a/c-2b).

## Replay rule, applied to the initial read

AppWire orders a `thread/read` response at the snapshot cut and delivers frames in producer order (`docs/superpowers/plans/2026-07-28-appwire-retry-safe-mutations-and-atomic-rejoin.md:19, 372-373`; the D23b store comment beside `applyThreadNotification`). The initial read (`openProjected`) and the refresh (`resumeProjected` → `rehydrate`) are the same RPC as a reread — `thread/read` with `includeTurns: true, subscribe: true, replaceSubscription: true` (`services/conversation.ts` `open`/`readProjection`); the only differently shaped read was the deleted `refreshCapabilities` (`includeTurns: false, subscribe: false`). Both await the read alone, so a frame delivered before the response is already folded into the snapshot: the escalation buffer that replayed such frames after the commit, and `needsSnapshot`, which rereads on every open that raced any other frame, are deleted. A frame after the response applies through the normal handler.

## actionUnavailable → the reducer (decision 2)

Native awaited a non-subscribing `thread/read` and wrote its capabilities into the model (and into the activity store via `setLiveCapabilities`) before surfacing the error; the web gates the composer off `model.capabilities` fed by `thread/status/changed` and lets the typed error propagate. Now `handleMutationError` surfaces the error and requests the one coalesced rehydrate; the model converges from the authoritative snapshot, and the activity view's capabilities come from that same `setLiveView`. Deleted with no production caller: `requestCapabilityRefresh` and `storeSet`, `LiveActivitySink.setLiveCapabilities` and `LiveActivityState.setLiveCapabilities`, the service's `refreshCapabilities` and `capabilityRevision`, and their tests (the RPC no longer exists, so its stale/epoch scenarios have nothing to protect). `requireControl`'s `sessionControls` read stays: a read over model fields, the web's `liveControls` rule.

**Deferred to c-2b, stated:** `getTruncatedItemIds` has no production consumer but is the freeze oracle for ~15 truncation tests that c-2b rewrites when the freeze itself goes; deleting it here would rewrite those tests twice.

## Mechanism → behaviour test mapping (no scenario dropped)

| Mechanism deleted | Test that pinned it | Behaviour test that now covers the scenario |
| --- | --- | --- |
| initial-read escalation buffer | `approvals.test.ts` "keeps resolutions delivered between the initial snapshot and its response" | same test, same assertion; the response carries no card (the hub cut after the resolution) |
| `needsSnapshot` reread on a pre-response frame | none (an untested reread per racing open) | **new, RED first**: "does not reread for a frame that precedes the initial read's response" |
| `requestCapabilityRefresh` + model write | "actionUnavailable publishes refreshed capabilities before error" | **new, RED first**: "actionUnavailable surfaces the error and requests one coalesced reread" (the reread's capabilities are what the composer sees) |
| cap key vs reread key in the scheduler | "I2: actionUnavailable capability refresh through store-owned scheduler" (3); "I2 fix: per-key completion — unrelated reread cannot delay error"; "Task 2A-1: cap-first exact completion"; "Task 2A-2: reread-first converse" (2) | "a refused mutation and the reread scheduler": the error surfaces while an earlier reread is held and the refusal's reread runs after it; a newer mutation's outcome is not disturbed by the older refusal's reread |
| stale-mutation guard on the refresh | "stale mutation recovery is suppressed by mutationId guard"; "mutation race: asserts error, pending owner, capabilities after stale cap refresh" | "a newer mutation's outcome is not disturbed by the older refusal's reread" |
| distinct cap keys per mutation | "Residual 2: heterogeneous scheduler outcomes — reread + cap refresh both run exactly once" (3); "Task 2A-3: distinct work survives same-key coalescing" | no analogue with one key kind; reread coalescing stays pinned by "resync coalesces to one rehydrate (F5)" and "Task 2A-3: same-key reread coalescing" |
| rebind during the refresh await | "I1 fix: stale post-await rebind suppresses A capabilities" | rereads already pinned by "I5: rebind A->B then late A completion => zero state change" (4) |
| error-path drain with a cap key | "Task 2A-4: first reread errors after release — same-key coalesced reread and distinct cap refresh both drain" | "a reread that fails does not block the next reread" |
| `setLiveCapabilities` sink publication | "D: requestCapabilityRefresh publishes to sink before conversation caps" (3); `activity.test.ts` "setLiveCapabilities (I4)" (6) | "a refusal's reread carries the reread's capabilities into the activity view" (real activity store) |
| `refreshCapabilities` RPC | `services/conversation.test.ts` "refreshCapabilities (non-subscribing)" (5), R3 (2), R4-Minor (1), R5 refresh variants (3), the refresh detour in "pins mutations to the read instance…", "keeps a pushed capability set when an older metadata read finishes" (covered by the next test in its describe), the refresh half of the `close` test, the sharedNotes test's observable (now: a gated action reaches the wire) | deleted with the RPC; the surviving scenarios keep their other assertions |

1. No reread on an open that races a frame before its response (fewer requests; same rows).
2. After a refused mutation the error is immediate and the model converges via one coalesced reread rather than a targeted capability read; the activity view's capabilities follow the same snapshot.

## Gates

`make test-native` green, `make lint-package-imports` PASS, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` PASS; package untouched. Files: `mobile/src/state/conversation.ts`, `mobile/src/state/activity.ts`, `mobile/src/services/conversation.ts`, `mobile/src/state/conversation.test.ts`, `mobile/src/state/activity.test.ts`, `mobile/src/services/conversation.test.ts`, `mobile-native/src/approvals.test.ts`, `mobile-native/src/sessionResume.test.ts`.


obra:
## Simplify fixes (`6974564e6236747e746b7c45fe8e21f8f3eecfb4`)

Applied, no behaviour change:

- **Comments describing deleted code**: the orphaned capability-refresh note above `getDraftRevision`; `handleMutationError`'s doc cut to the draft-restore + mutationId sentences (the in-body comment states the rule); the "cap refresh" mentions in the binding-epoch comments; `state/activity.ts` and `activity.test.ts` headers no longer list `setLiveCapabilities`; the empty "I4: setLiveCapabilities narrow strict sink" section; the store test's "Same-key completion waiter invariant (capability keys)" describe deleted — its one remaining fact (`requestRehydrate` is fire-and-forget) is stated at the scheduler comment; the service describe is named for its one test, "actionUnavailable is re-thrown".
- **Encodings**: the two new comments each point at `applyThreadNotification`'s response-cut note instead of restating it; `resumeProjected` installs its live handler after the commit exactly as `openProjected` does (`hydrated` gone — contained, ~15 lines); `handleMutationError` is synchronous and its four callers no longer await it.
- **Tests**: `yieldMicrotask` replaces the file's only real timer; one file-scope `refusal()` builder used at all three sites; the real activity store's state is the `LiveActivitySink` directly (tsc agrees); `makeReadProjectionResult(thread, capabilities = ALL_TRUE_CAPS)`, so `withCaps` is a one-liner.
- **Churn**: `getTruncatedItemIds`' declaration is back above the `publishExternalError` comment block (the method stays for c-2b).

Left for later slices as you listed: `withCapabilityRefresh` (c-2a deletes it), the `liveOwnedRevs` merge comment (c-2a/c-2b), `jobOutput.ts` untouched.

No RoboRev round-1 verdict on `1aa5abd47` had landed at push time; nothing folded in.

Merged `origin/main` (`833860668`) with `--no-ff`, no conflicts; `git status` clean before push. Gates on this head: `make test-native` green (Metro bundle; native 786/786, shared 766/766, tsc, script imports), `make lint-package-imports` PASS, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` PASS; package untouched.

obra:
## Review round 1 disposition (6974564e6)

CI 15/15 green; /simplify applied. RoboRev: two Lows, no Medium or higher → merged under the Lows rule; both land in an immediate follow-up PR.

- **Low, plain `open()` no longer recovers from `actionUnavailable`** because `requestRehydrate` needs a bound sink and only the projected paths bind one → accepted. Production native uses the projected paths only, so the follow-up either gives `open()` the same recovery or scopes the contract and the stated delta to the projected paths, whichever is honest about what `open()` is for.
- **Low, dead artifacts** (`notificationRef`, the `ThreadCapabilities` import, `handleMutationError`'s unused parameters) → accepted; deleted in the follow-up.

PR #1541, opened 2026-09-16
test(agent): config-home fork audit reads the gate the constructor receives and follows home-isolating helpers

obra:
Follow-up to #1534 (#1527). Closes the two escapes in `agent/config_home_fork_audit_test.go`:

1. **Home isolated through a helper.** `setsConfigHome` only looked at the test's own body, so a test calling `isolateHome(t)` (Setenv in the helper) and then `NewSession` itself was never examined. "Isolates the config home" is now a fixed point over the test files' functions, the same closure used for "reaches a constructor".
2. **Gate read as a bare identifier.** Any `skipGitSnapshot`/`withoutGitSnapshot` identifier in the body cleared a finding, so `cfg.testOnly.skipGitSnapshot = false` (as in `session_worktree_swap_scratch_unix_test.go:280/:947`) read as gated. The audit now decides each constructor-reaching call by the value it receives:
   - `NewSession(..., cfg)` / `RestoreSessionFromMetaWithConfig(..., rcfg)`: the `SessionConfig`/`RestoreSessionConfig` literal's `testOnly.skipGitSnapshot`, or for a variable the **last top-level write before the call** (`x := lit`, `x = lit`, `var x = lit`, `x.testOnly = testConfig{...}`, `x.testOnly.skipGitSnapshot = true|false`) in the function body and every enclosing function literal; a write under a condition or loop does not count; a later `false` un-gates. `RestoreSessionFromMeta` (no config) can never be gated.
   - Helper calls (`newSession`, `newSkillSelectionSession`, ...): a `withoutGitSnapshot()` option or a gated config among the arguments, or a helper that gates every session it constructs itself (its own fixed point, so `newSession`'s conditional `if o.skipGitSnapshot {...}` does not make it self-gating).

RED first: `agent/config_home_fork_audit_shapes_test.go` writes one fixture source per shape into a temp dir and checks exactly which functions the audit flags. Against the #1534 audit, five of seven shapes failed (helper-isolated home missed; later-false missed; conditional-true missed; closure-clear missed; unconditionally-gating helper over-flagged). All seven pass now.

Real tree, audit findings: **0 before, 0 after** the stricter check (no real test newly flagged, so no further gating was needed).

Gates: toolchain `gofmt -l` clean; agent module `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz`, `golangci-lint run` (0 issues); root `make lint` all PASS; agent suite ok (388s); root `go test -short -count=1 .` ok.


obra:
## Review round 1 disposition (d0110241e)

CI 15/15 green on that head. RoboRev: one Medium.

- **Medium, `callGated`'s helper-argument scan treats any nested identifier with a gated write as proof the call is gated** (`newSession(t, withDir(cfg.StateDir))` reads as gated) → accepted. The /simplify push that followed (d5733571f) removes identifier resolution entirely: an argument counts only if it is a `withoutGitSnapshot()` call or a `SessionConfig` literal that satisfies `literalGate`, and an identifier config in an isolating test is itself a finding. The two examples from the review are being added as fixture shapes so the closure is pinned.

obra:
## Simplify fixes (d5733571f)

Census, measured with the audit's own predicates over this branch's `agent/*_test.go`: isolating functions contain **16** constructor-reaching calls — **14** `NewSession(..., SessionConfig{literal})`, **1** `newSession(t, withoutGitSnapshot())`, **1** helper call that gates in its own body; **0** pass a variable config. So the variable-write evaluator modelled a shape nobody uses and is gone.

- Deleted `gateOfIdent`/`writeTo`/`selectorPath`/`scopeStatements` and the ident/unary/paren/decl paths of `gateOfExpr`. A constructor's config position must now be a **literal** carrying the gate; a variable there is a finding whatever it holds, and the Fatalf text says to write the gate in the literal. This is stricter than the review's Low asked for and covers it: a re-enabling write can no longer hide because the variable form itself is refused.
- One fixed point: `closeOver(set, funcs, has func(auditFunc, map[string]bool) bool)` expresses `reaches`, `isolates` and `gated`; the per-name aggregator is gone (same-named method/function collisions are not distinguished; none exist, noted in a comment).
- `callGated`'s helper-argument scan: a `withoutGitSnapshot()` call or a `literalGate` composite literal anywhere in an argument subtree; no identifier resolution.
- `parseAgentTestFiles(dir)` extracted and shared with `deadline_audit_test.go` (byte-identical loader; `session_emit_lock_guard_test.go`'s non-test twin left alone).
- Doc comment trimmed to the sidecar paragraph plus #1527; the rule lives in the Fatalf text.
- Fixtures: the evaluator's three cases replaced by "a config held in a variable is refused"; the `TestViaHelperGated` twin dropped (literal-gated-through-helper stays covered by `gatedSession`).

Net delta for the simplify commit: 3 files, +108/−314. Real-tree audit: 0 findings before and after. Gates on d5733571f: toolchain gofmt clean; agent `go vet`, `-tags evenerfuzz`, `GOOS=windows -tags evenerfuzz`, `golangci-lint run` 0 issues; root `make lint` all PASS; agent suite ok (337s); root `go test -short -count=1 .` ok.

## Round 1 fixes (b0a06ae8425e787003872a778ab2e02cb62c3804)

The Medium on `d0110241e` (helper-argument scan treating any nested identifier with a gated write as a gate) is closed by the simplify commit d5733571f: identifier resolution no longer exists, so `withDir(cfg.StateDir)` and `withConfig(SessionConfig{StateDir: cfg.StateDir})` carry no gate. Both examples are now fixture shapes in `config_home_fork_audit_shapes_test.go` (`TestGatedVariableDerivesAnotherOption`, `TestGatedVariableFieldInsideUngatedLiteral`), expecting a finding; verified RED against the `d0110241e` audit file and GREEN on the current one. This commit changes fixture text only; vets, golangci-lint and both audit tests re-run clean.

PR #1545, opened 2026-09-16
perf(delegatestore): clone and diff only the aggregates an event touches

### LINKED ISSUE #1537: race-modules/agent runs at the 10-minute go test cap: delegatestore.Apply DeepEquals every aggregate per event (O(n^2) under -race)
Split from #1394. The three 10-minute timeouts on 2026-09-16 were not a hang: the dump shows six tests at 0s still cycling through the six parallel slots and 2543 tests queued in testing.waitParallel at the deadline. The agent package's race binary simply runs at the default 10m cap.

Measured on passing main runs (`race-modules / agent` test step, which is the -race build plus `go test ./...` with -p 4 -parallel 6): Sep 11 ~475s, Sep 12 ~550s, Sep 13 ~600s, Sep 14-16 600-657s. Runner variance now tips the agent binary over 600s.

Locally (-race -parallel 6, M-series): 689s total. The largest sequential costs are
- TestLoadSessionJobActivityTree_StopsRecursingOnceWorkBudgetExhausted 83s
- TestLoadSessionJobActivityTree_ContinuationRevisionComputationSharesLoadBudget 82s
- TestSession_CurrentEnv_NoRaceWithConcurrentSwap 44s
- TestLoadSessionJobActivityTree_TuesdayFullyFoldsAndFullyPaginates 28s
- TestTrimActivityTreeToFit_TrimsExcessEntries 16s

pprof of the 83s test: 59% of CPU is `delegatestore.Apply` -> `reflect.DeepEqual` (agent/internal/delegatestore/fold.go:58). Apply clones the whole State and DeepEquals every aggregate's publicProjection on every event, so folding 2001 descriptors is O(n^2) DeepEquals, and reflection under -race pays checkptr on every access. Without -race the same test takes 25s. Callers on this path: Store.AppendBatch and extendHistoricalDelegateFold.

This is product cost, not just test cost: every delegate event append pays O(number of delegates) DeepEquals. Candidate fix: bump ProjectionRevision only for aggregates the event touched (applyEvent knows which IDs it mutates) instead of sweeping the whole state, or compare projections without reflection. A timeout bump is not the fix.

obra:
Closes #1537

`delegatestore.Apply` cloned the whole `State` and ran `reflect.DeepEqual` over every aggregate's `publicProjection` on every event, so folding a root with n delegates was O(n²) deep clones and reflection. Under `-race` reflection pays checkptr on every access, which is why the 2000-descriptor activity-tree tests dominated the `race-modules / agent` job and pushed it to the 10-minute cap (#1394).

## Change

Apply now works out which aggregates the event can write (`touchedDelegates`) and clones, diffs and (on a rejected event) restores only those:

- most events reach only their own delegate;
- `delegate_resumability_closed` and `delegate_subtree_stop_requested` reach the target and its descendants;
- `delegate_subtree_stop_completed` rewrites every aggregate's `PendingDeliveries`, so it still touches everything (that apply function is O(n) by itself).

Change detection is unchanged: the same `publicProjection` DeepEqual decides whether `ProjectionRevision` bumps, it just runs only for the aggregates the event could have changed. Untouched aggregates keep their pointer, which is the property `extendHistoricalDelegateFold` already relied on when it hands Apply a shallow `maps.Clone` of a cached state. Two new tests pin that contract (`TestApplyLeavesUntouchedAggregatesShared`) and the rollback of a rejected event that had already written to its aggregate (`TestApplyRestoresStateWhenEventIsRejected`). All existing delegatestore tests pass unchanged.

One narrowing to know about: the old Apply rejected a state containing a nil aggregate anywhere (a side effect of `cloneState` walking every entry). Now the guard covers the aggregates the event touches, and `touchedDelegates` also lists every nil entry for the three kinds whose apply functions walk the whole state (`TestApplyRejectsNilAggregateWalkedBySubtreeEvents`), so no apply function can dereference one. A single-delegate event no longer rejects an unrelated nil entry it never reads. `Store.AppendBatch` still runs `cloneState` on the caller-supplied state, so its nil check (`TestAppendBatch_CloneStateError`) is unchanged; `Fold` and `Apply` never insert nil entries themselves.

## Measurements (M-series, local)

`BenchmarkFoldWideTree` (new, `go test -bench BenchmarkFoldWideTree -benchmem`, no `-race`):

| width | before | after |
|---|---|---|
| 100 | 10.0–12.9 ms/op, 24.3 MB/op, 35.5k allocs/op | 46–57 µs/op, 138 KB/op, 311 allocs/op |
| 2000 | 3.77–3.88 s/op, 9.78 GB/op, 14.0M allocs/op | 0.93–0.96 ms/op, 2.84 MB/op, 6.0k allocs/op |

`TestLoadSessionJobActivityTree_StopsRecursingOnceWorkBudgetExhausted` under `-race` with `-cpuprofile`:

- before: 91.4 s; `delegatestore.Apply` 57.8% cum, `reflect.DeepEqual` 50.7% cum of 71.6 s of samples
- after: 23.6 s; Apply no longer appears in the profile (3.05 s of samples over 21.7 s wall)

Full agent module under `-race -parallel 6` (`go test -race -count=1 -parallel 6 ./...` in `agent/`), same machine, back to back:

- before (detached `origin/main` worktree at ed4501db8): `panic: test timed out after 10m0s` in the `agent` package, the #1394 failure reproduced locally
- after: green, 620 s total, 597 s for the `agent` package
- CI `race-modules / agent` on 19986d1: 6m53s, vs 600–657 s on passing main runs Sep 14–16 (#1537) Also under `-race`: `TestLoadSessionJobActivityTree_ContinuationRevisionComputationSharesLoadBudget` 82 s -> 12.1 s.

## Gates

`$(go env GOROOT)/bin/gofmt -l` clean on touched files; `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...` in `agent/`; `golangci-lint run ./...` 0 issues; `make lint`; delegatestore under `-race -count=3` and with `-tags evenerfuzz`; root `go test -short -count=1 .`.


obra:
RoboRev's Low on 19986d1 (a nil entry outside the target subtree reached `applySubtreeStopRequested`'s whole-state loop) is fixed in cb73f8b5c: `touchedDelegates` lists a state's nil entries for the three kinds whose apply functions walk every aggregate, so the existing pre-clone guard returns "aggregate is nil" first. Pinned by `TestApplyRejectsNilAggregateWalkedBySubtreeEvents` for all three kinds.

Issue #1535, opened 2026-09-16
reducer: model text under deltas is unbounded once native drops its 64 KiB truncation freeze (D23c-2b); consider a bound in the package


obra:
From D23c scoping (2026-09-16). Native today freezes an item at 64 KiB (`truncateItem` + `truncatedItemIds`) so later deltas cannot grow it; D23c-2b routes deltas through `reducer.applyNotification` and keeps the 64 KiB rule as a pure post-projection pass on display rows, so the visible row stays bounded but the model's text (`pendingText`, tool `output`) grows without bound, as it already does on the web. The web runs in a desktop browser; the phone has less headroom, and the old freeze protected memory as well as layout. Decision 2 (adopt the web's rules) is applied for now; this issue asks whether the package should bound model text itself (a cap with a marker at the reducer level, benefiting both hosts) rather than each host truncating after the fact. Measure a long tool-output stream on device before deciding.

obra:
Measurement from #1547 round 3 (D23c-2a, merged as d3c537e45): 32,000 × 64 B `item/toolOutput/delta` frames fold in 23 ms (linear; V8 rope, Hermes buffered primitive; only an array copy is quadratic), and the model then retains 2,048,000 bytes while the display row froze at 64 KiB. The web has no cap on tool output anywhere today, so a reducer-level bound with a marker is a visible web change as well as the phone's memory protection; that is the decision this issue asks for. Native's rows stay bounded (64 KiB freeze in c-2a, post-projection cap in c-2b); only the model half is unbounded, on both hosts.

obra:
D23c-2b (#1580) makes the display bound a post-projection pass: `capAndTruncate` re-encodes every retained row (up to 500) with TextEncoder on every accepted frame, because the projection is whole-model and the model is unbounded. Both /simplify reviewers flagged the per-frame cost; it is the accepted trade of one projector for frames and snapshots, and the fix belongs here (a bound in the model, or a per-item "already bounded" marker the reducer maintains so the display pass is a lookup), not in a native-side cache.

PR #1547, opened 2026-09-16
refactor(native): the model is live under every item frame; rows keep their appliers (D23c-2a)

obra:
Slice c-2a of SDK-migration row D23 (docs/superpowers/plans/2026-09-12-sdk-migration.md:213), after c-1 #1540 / #1546. The seam is option (B): the model goes live under every item frame while the display rows keep today's appliers, so c-2b can flip the rows to a projection of the model and delete the row machinery in one place.

## Dual-write until c-2b

`applyNotification` now folds **every** frame about the thread into the package reducer's model first — `item/started`, `item/completed`, the three delta methods, `item/agentMessage/reset`, `warning`, `evener/steering/injected` included — then the existing row appliers update the display rows from the wire frame, spreading that updated model. A row applier that bails (missing target, frozen row, reread request) still publishes the model half (`publishModel()`). Stated in the store comment: rows and model are both maintained until c-2b flips the rows to `projectConversation(model)`; this is the same intermediate the web store had before its projector.

What is live in the model now: `turns` populate under `item/started`/`item/completed`; `pendingText` and `output` accumulate under the deltas; `failedToolCalls` moves on `item/completed`; `modelRetry` clears on a model-output `item/completed` — **RoboRev's D23b Low on #1522 closes here**; a steering injection appends its item; a `warning` lands in the active turn or is dropped.

`withCapabilityRefresh` — an identity wrapper whose name promised a refresh that never happened — is deleted with its 16 callers in `services/conversation.ts`.

- **Model-only until c-2b** (the phone shows the same rows as before this PR): a `warning` with no active turn is dropped in the model while the row applier still appends its failure row; a steering injection exists in the model only (no row until c-2b); a sparse `item/completed` for an assistant message keeps the streamed text in the model (`mergeCompletedText`) while today's row applier re-projects the row from the sparse frame — the flip in c-2b is what puts that text back on screen (a pre-existing row-applier gap the model half already fixes).
- No wire, package or visible-row change otherwise.

## Tests (fixture identity only; no mechanism→behaviour rewrites in this slice)

New describe "the model is live under the item frames (dual-write until c-2b)": a delta stream then a sparse completion leaves the model's item with the accumulated text; a steering injection appends exactly one steering item; `modelRetry` clears on model-output completion; a warning without an active turn is dropped in the model while the row applier still shows it. All existing row tests pass unchanged.

## Gates

`make test-native` green, `make lint-package-imports` PASS, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` PASS; `git status` clean before push; package untouched (no reducer change).

Next: c-2b flips the rows, deletes `projectSingleItem`, the cluster splicing, `liveOwnedRevs`, the truncation freeze, `getTruncatedItemIds` and project.ts's wire-item helpers, with the mechanism→behaviour rewrite and #1535 cited for the unbounded model text.


obra:
## Simplify fixes (`6d2a242aa29a694148d08f12bcfbbda81e99b0c1`)

Applied, no behaviour change:

1. The response-cut note beside `applyThreadNotification` no longer says item frames wait for D23c — they fold into the model here (dual-write); `liveOwnedRevs` survives only for the display rows until c-2b projects them from the model.
2. The `evener/thread/resync` bail calls `publishModel()` like every other bail site — a no-op today (the reducer has no resync case), kept so "every bail publishes" has no exception a new path could copy.
3. Tests: the new `agentMessage(over)` helper is gone in favour of `agentMessageItem`, which now takes an optional text so a sparse wire item (no text field) can be built from it; one `openProjectedThread(thread)` helper beside `makeReadProjectionResult` replaces the new `openLive` (the five older `openProjectedWithItems` copies left alone, as scoped).

Skipped on purpose, as you said: collapsing `publishModel` and the seven `set()` calls into one post-switch publish — c-2b deletes the block.

**RoboRev round 1 (`bf37741`)**: "No issues found"; nothing to fold in.

Gates on this head: `make test-native` green (Metro bundle exported; native 787/787, shared 770/770, tsc, script imports), `make lint-package-imports` PASS, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` PASS; `git status` clean before push; package untouched.

obra:
## Review round 2 disposition (420413ff1)

CI 15/15 green. RoboRev: one Medium, one Low. The queue is being handed to a new owner; this comment records the state rather than dispatching a round.

- **Medium, the model half accumulates streamed tool output with no size cap and O(n²) string rebuilding** (`reducer.ts` `item/toolOutput/delta` does `output + delta` while `pendingText`/reasoning use the O(1) `appendChunk`) → accepted. The right depth is the reducer (`output` onto the same chunk append, a package change; the web pays this today too) rather than a native-only guard; #1535 already tracks the bound. Until fixed, c-2a's dual-write retains full tool output in the model after the row froze at 64 KiB.
- **Low, `project.ts:44-46`'s model contract comment is now false** → accepted; reword: item frames fold through `applyNotification`, `turns` is live, `items` is the dual-written display half until c-2b.

The PR is also CONFLICTING again after the D13 merge (plan table). Next owner: refresh, decide reducer vs wrapper for the Medium (a reducer change needs the package gates: test-api-package + falsification, test-web), fix the Low, re-review.

obra:
## Round 3 fixes (`74d4757ac`)

**Refresh.** Merged `origin/main` (`fa2084eb8`); the only conflict was the plan status table, resolved as a union (main's D13 row, then this row). `git diff origin/main...HEAD --stat` is mobile-only plus the plan row:

```
 docs/superpowers/plans/2026-09-12-sdk-migration.md |   1 +
 mobile/src/conversation/project.ts                 |   7 +-
 mobile/src/services/conversation.ts                | 199 ++++++++-------------
 mobile/src/state/conversation.test.ts              |  88 ++++++++-
 mobile/src/state/conversation.ts                   |  51 ++++--
```

---

### Medium — "the model half accumulates streamed tool output with no size cap and O(n²) string rebuilding"

**Measured before accepting. The O(n²) half does not reproduce on either runtime we ship, so the reducer is unchanged; the retention half is real and is #1535.**

I set out to write the RED test the disposition asked for (stream past `MAX_ITEM_BYTES` in many deltas, assert the per-delta cost is not quadratic). It is green by construction. Three measurements:

**1. The real reducer, per-delta timing, fold only (vitest/Node, V8; 64 B deltas, `item/toolOutput/delta` vs `item/agentMessage/delta` on the same scaffold).** `late/early` is the mean per-delta cost of the last decile over the first — the token-flood bench's own O(n²) tell:

| n | toolOutput (`output + delta`) | agentMessage (`appendChunk`) |
|---|---|---|
| 2,000 | total 1.7 ms, late/early 0.33x | total 2.7 ms, late/early 0.34x |
| 8,000 | total 4.4 ms, late/early 0.86x | total 7.6 ms, late/early 0.25x |
| 32,000 | total 23.0 ms, late/early 0.87x | total 23.0 ms, late/early 1.97x |

Flat per delta, linear in total, and indistinguishable from the `appendChunk` path. At n=32,000 the model holds 2,048,000 bytes of output.

**2. Same folds with a flat read of the accumulated text after every delta** (`output.charCodeAt(len-1)` vs `pendingTextJoined(item.pendingText).charCodeAt(len-1)` — what a renderer reading every frame does):

| n | toolOutput concat + read | appendChunk + `pendingTextJoined` read |
|---|---|---|
| 8,000 | 219.5 ms | 211.2 ms |
| 32,000 | 3,489.8 ms | 4,783.0 ms |

Both go quadratic on V8, by the same amount — V8 flattens the rope (or the brand's cached `text`, which `appendChunk` itself maintains with the very same `brand.text + delta` concatenation) on each read. So the chunk backing does not fix the read-side cost either; that cost is a renderer question, identical for `pendingText` today.

**3. Hermes VM v0.13.0 (`hermes -O`, standalone shapes, same 64 B deltas)** — the phone's engine family; RN 0.86 ships a newer Hermes than the last standalone release, so treat this as directional, not device truth:

| shape | n=8,000 | n=32,000 |
|---|---|---|
| `output = old + delta`, new item object per delta | 0 ms | 7 ms (late/early flat) |
| same, every prior string retained | 3 ms | 6 ms |
| `appendChunk` shape (push + text cache) | 1 ms | 10 ms |
| concat + flat read per delta | 2 ms | 4 ms |
| **array copy `[...chunks, delta]` per delta** | **68 ms, late/early 15x** | **1,299 ms, late/early 25x** |

The only quadratic shape on either engine is the **array copy** — which is what `appendChunk` was built to remove (`pendingText` is a `string[]`; copying it per delta is genuinely O(n) with no rope or buffered primitive underneath). `output` is a plain string field with no array in its path, so it never paid that cost. Moving it onto the chunk backing would mean changing `ItemModel.output` from `string` to `string[]` and touching ~50 readers across the package, web and native, for no measured win on either runtime.

**What is true in the finding:** the model retains the whole stream (2,048,000 bytes above) after the display row froze at 64 KiB. Whether the package bounds model text itself, and what marker truncation leaves, is the open design question in **#1535** (which names `pendingText` and `output` together) — so no bound is added here, and no native-only cap in `conversation.ts`. Until #1535 lands, c-2a's dual-write does retain untruncated tool output in the model; that is the same rule the web runs today (decision 2).

The measurement scripts are reproducible in a scratch tree, not in the repo. If you want the refutation re-checkable in-tree, the natural home is a tool-output case in `appwire-client/typescript/tokenFlood.bench.ts` (it already prints exactly this growth curve for agentMessage, and `.bench.ts` is outside `vitest run`); I left it out of this mobile-only slice rather than reopen the package gates — say the word and it is ~35 lines.

---

### Low — `project.ts`'s `MobileConversation` contract comment is false

Fixed in `74d4757ac`. It now states the present: item frames fold through `reducer.applyNotification` (`state/conversation.ts`) before the store's row appliers run, `turns` is live between rereads, and `items` is the dual-written display half until c-2b projects the rows from the model.

---

### Gates (all foreground, on `74d4757ac`)

- `make test-native` — green: Metro iOS bundle exported, native 787/787, shared 771/771, tsc, script imports
- `make test-web-browser` — `PASS` all seven guards (layout, overflow, shell, spawn, transcriptscroll, retirement, skill) on the first run
- `make test-api-package` — `qualified @evener/appwire-client@0.1.0: installed imports, declarations and read-only example`. The index.ts falsification does not apply: this PR adds no package module (the package is untouched), so there is no new re-export to delete.
- `make lint-package-imports` — `PASS lint-package-imports (1s)`
- `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend` — `Checked 182 files`, no fixes
- `npm run lint` in `cmd/evener-hub/frontend` — `Checked 1314 files`, no fixes
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — `PASS secret-scan (0s)`; `gitleaks detect --no-git` on the changed path directly — `no leaks found`
- root `go test -short -count=1 .` — `ok primeradiant.com/evener 32.689s`

`git status` clean before push; `gh pr view 1547 --json mergeable` reads `MERGEABLE`.

obra:
## Review round 3 disposition (74d4757ac)

Refreshed onto main (plan-table union only); MERGEABLE; CI running on this head.

- **Medium (model-side tool output: O(n²) concat + unbounded)**: the O(n²) half is **refuted by measurement**, not by reading. Real reducer, 64 B deltas: `item/toolOutput/delta` is flat per delta and linear in total (1.7 / 4.4 / 23.0 ms at n = 2k / 8k / 32k, late/early decile ratio 0.87×), indistinguishable from `item/agentMessage/delta` on `appendChunk` (2.7 / 7.6 / 23.0 ms); on Hermes 0.13 concat is flat too (7 ms at 32k) and only the `[...chunks, delta]` array copy is quadratic (1,299 ms, 25×). `appendChunk` exists to remove that array copy and itself keeps its cached join with `brand.text + delta`. So the reducer is unchanged. The **unbounded** half is accepted as already tracked: the model retains the whole stream (2,048,000 bytes at 32k × 64 B while the row froze at 64 KiB); where the bound lives is #1535's open design question and no native-only cap is added here.
- **Low (project.ts contract comment)**: fixed; the `MobileConversation` comment states what is live now.

Residuals filed: https://github.com/prime-radiant-inc/evener/issues/1560 (per-frame reads of accumulated text are O(n) on V8; the bench never reads), https://github.com/prime-radiant-inc/evener/issues/1561 (`appendChunk` doc comment), https://github.com/prime-radiant-inc/evener/issues/1562 (tool-output case for the bench).

Merge on CI green + RoboRev clean at this head (/simplify done at 6d2a242aa; this round is a merge, a comment and a measurement).

obra:
## Review round 4 disposition (74d4757ac)

CI 15/15 on this head. RoboRev repeats the round-2 Medium (unbounded + quadratic tool-output accumulation in the model). Ruling, both halves measured:

- **Quadratic**: refuted, same measurement as round 3 (flat per delta on V8 and Hermes; `appendChunk` uses the same `text + delta` for its cached join; only the array copy is quadratic).
- **Unbounded**: the web bounds tool output nowhere today (no cap in `cmd/evener-hub/frontend/src` beyond `jobOutput`'s request window), so after this PR native's model holds exactly what the web's model holds. Under decision 2 (the web's rule wins) that is the accepted behaviour; whether the *package* should bound model text with a marker, which would change what the web renders for long tool outputs, is the product decision #1535 already asks Jesse, now with the measurement (2,048,000 bytes retained at 32k × 64 B deltas while the row froze at 64 KiB). Display rows stay bounded by the 64 KiB freeze in this PR and by the post-projection cap in c-2b.

Merging on CI green with the finding dispositioned above (/simplify done at 6d2a242aa).

PR #1549, opened 2026-09-16
refactor(sdk): one transcript display hub-defaults store core for web (D6-core)

obra:
SDK migration plan row **D6**: `stores/transcriptDisplay.ts`'s hub half and native `nativePreferences.ts`'s `transcriptMobile` half become one framework-free core; the web's local layer and `useIsMobile` stay web-side. Branched off D5's merge (`199b4a611`) and built in D5's shape. Serial-queue PR (`index.ts`, `tsconfig.build.json`, qualify smoke, README, plan row).

## Core (`appwire-client/typescript/transcriptDisplayStore.ts`, root module)
`createTranscriptDisplayStore({ client, drafts? })`: `Pick<AppwireClient, "request" | "onNotification">` client port, optional `TranscriptDraftStorage` port, `setSupport` / `beginReadyGeneration` / `endReadyGeneration` / `detachHub` / `reset` / `dispose`. One reply fence at every await (`readStillMine` / `patchStillMine` / `writeStillMine` over one `liveHub(generation)`), one payload retirement at every reset site (`retirePayload`), written up front with the fence/reset table test (4 reset sites × 3 in-flight paths). State: `hubSupport`, `hubLoading`, `hubError`, `hubErrors` (per layout), `hub` (per-layout confirmed defaults, kept presenting across a transient disconnect), `loaded` (confirmed for the current generation), `drafts` (the direct write's per-layout preview), and the checkpointed editor's `draft` / `saving` / `writeUncertain` / `storageUnavailable` / `draftConflict` / `draftError`. Two write paths share the PATCH: `patchHubDefault(layout, config)` (the web's direct write: preview, exact-shape and semantic response validation, canonical conflict adoption, the preview kept on a malformed success) and `editDraft` / `saveDraft` / `discardDraft` / `rebaseDraft` (native's checkpointed editor: intent persisted before the request leaves, `writeUncertain` on a lost reply until an authoritative read settles it, a newer external revision keeps the proposal for review). D5's two follow-up Lows are built in: an undecodable `changed` broadcast schedules a read; `fromWireChange` requires every field present and well-formed.

## Web (`stores/transcriptDisplay.ts`, 706 → adapter)
[...]

PR #1555, opened 2026-09-16
docs(handoff): watch-pr.sh shares one poll budget and treats an empty rollup as pending

obra:
Follow-up for the two Lows RoboRev found on #1552 (merged under the Lows rule).

- The CI-only tail (RoboRev verdict in, CI still running) had no deadline; it now consumes the same 120-poll budget as the main loop, so the documented "or after 120 polls" contract holds on every path.
- CI completion was gated on a hard-coded `>= 10` non-roborev checks, which silently never fired on a PR with fewer checks; it is now "rollup non-empty and every check terminal", and the header says an empty rollup counts as pending.

Checked: `bash -n`; run against #1549 (CI 15/15, verdict in) prints `CI COMPLETE: SUCCESS=15` and the verdict line and exits 0. Simplify by reading (17-line diff to a 43-line coordinator script).


obra:
## Round 1 fixes (bd4f7ab74)

- **Medium (floor dropped to `> 0`)**: accepted. Measured: the ledger records the first inline watcher firing a false CI COMPLETE on a partial rollup, and #1547's head had 12 of its checks registered mid-run when I looked, exactly the early-fire case. The floor is back at 10 and is now the documented parameter the original Low asked for (`WATCH_PR_MIN_CHECKS`, default 10); the header explains why a floor exists.
- **Low (extra sleep after the budget is exhausted)**: accepted; the trailing sleep is guarded by the budget.

Checked: `bash -n`; against the merged #1552 head (15 checks, verdict in) prints `CI COMPLETE: SUCCESS=15` + the verdict and exits 0; with `WATCH_PR_MIN_CHECKS=20` prints only the verdict and keeps polling. Simplify by reading.

obra:
Closing: Jesse ruled the watcher script does not belong in the evener repo. It is being removed from main instead; the fixed copy lives outside version control with the coordinator's ledger.

PR #1569, opened 2026-09-16
refactor(web): the navigation store takes its expansion persistence as a port (D14-1)

obra:
SDK migration phase D, row D14, slice 1 of 3 (plan: `docs/superpowers/plans/2026-09-12-sdk-migration.md`). D14 is the highest web blast radius row in phase D (`stores/navigation/store.ts` 866 lines, 17 web importers, a 2758-line contract test), so it lands in three PRs. This one removes the store's one host dependency that is not the AppWire client — where rail expansion lives between visits — and changes no behaviour in either app.

## The dependency this removes

The plan row names it: `store.ts:46` imported `loadExpansion`/`saveExpansion` from `shell/rail/railExpansion.ts`, and `store.ts:209` called `loadExpansion()` inside `initial()` — i.e. `localStorage.getItem` runs at store creation, at module init. `railExpansion.ts:36-55` catches every failure mode (storage blocked, key absent, malformed JSON, wrong shape) and returns an empty map, so on a host without `localStorage` the store would have initialised with no persisted expansion and dropped every write in silence, with nothing failing loudly to say so.

## What changed

- New `cmd/evener-hub/frontend/src/stores/navigation/persistence.ts` (21 lines): the `NavigationPersistence` port — `readExpansion(): Map<string, boolean>` and `writeExpansion(expansion: ReadonlyMap<string, boolean>): void`, the names the plan row fixes — plus `railExpansionPersistence`, the browser adapter, which is `{ readExpansion: loadExpansion, writeExpansion: saveExpansion }` and nothing else.
- `store.ts`: `initial(persistence)` reads `persistence.readExpansion()`; `actions(persistence)` writes through `persistence.writeExpansion(...)` at both save sites (`setExpanded`, `toggleExpanded`); the two are joined by one `navigationState(persistence)`, which is what `createStore` and `resetNavigationStoreForTests` both build. The singleton is constructed with `railExpansionPersistence`; `resetNavigationStoreForTests` takes an optional port and defaults to the same one, which is how a test observes the seam without any of its 40-plus callers changing.
- `railExpansion.ts` is byte-identical: the cap, the key, the swallow-every-failure load and the best-effort save all stay in the web module, along with its own test. `Rail.tsx` still calls `loadExpansion`/`saveExpansion` directly for its own `useState`; that is not this store and is untouched.

## What stayed

- `navigationStore` and `useNavigationStore` keep their exact surface. Zero of the 17 web importers changed; no test fixture changed.
- `zustand` / `zustand/vanilla` stay imported by `store.ts`. The hook and the vanilla store are D14-2's problem, not this slice's.
- Nothing moves into `appwire-client/typescript`. The package is untouched, so `make test-api-package` and its reachability falsification do not apply to this PR.
- `store.ts` still imports the pure `projectNodeExpansionKey` from `railExpansion.ts` (used at `:686` to decide which project roots boot hydrates). It touches no storage and has no module-init side effect, but it is a web module, so D14-2 has to relocate that three-line key helper (or take the key naming as part of the port) when the core moves. Called out rather than done here, because the brief keeps `railExpansion.ts` unchanged.

## Oracle status

`stores/navigation/store.test.ts` (2758 lines) is unchanged — not one line, fixtures included. It already pins the behaviour this slice must preserve through the default web binding: "rail expansion persists through store reset, overrides defaults, and hydrates once" (`:2340`) seeds `EXPANSION_STORAGE_KEY`, resets, and asserts the store read it; "a canonical persisted project-node key hydrates one v2 project root during boot" (`:2375`) does the same for `projectnode:` keys. `shutdownConvergence.test.ts`, `selectors.test.ts` and `codec.watches.test.ts` are unchanged too, as are the 25 other suites that call `resetNavigationStoreForTests()`.

## TDD

New `persistence.test.ts` (6 tests) written first, against a recording port whose reads and writes are both observable:

1. RED as "Failed to resolve import `./persistence`" — the module did not exist.
2. With `persistence.ts` added but `store.ts` untouched: **4 failed, 2 passed**. The four that failed are exactly the port assertions — the store read `localStorage` instead of the port (`expanded` empty where the port held a row; populated where the port held nothing), and neither `setExpanded` nor `toggleExpanded` reached `writeExpansion`. The two that passed exercise the web adapter and the default binding, which were already true.
3. After the `store.ts` change: 6 passed.

The tests assert both directions of the seam: a port holding a row puts it in `expanded` and a port holding nothing leaves `expanded` empty *even with the browser blob set*, so a store that fell back to `localStorage` would fail; and every write lands in the port with `localStorage` still null, so a store that wrote both would fail.

## Q9 measurement — native expansion persistence (asked in the phase D worksheet, needed for D14-3)

Grepped `mobile/src` and `mobile-native/src` for `expanded|expansion|collapse|isExpanded|railExpansion` at `fa2084eb8`:

| Where | What it is | Persisted? |
| --- | --- | --- |
| `mobile-native/src/navigationPages.ts`, `navigationReveal.ts` | the two files D14-3 folds onto the core | **no mention of expansion at all** |
| `mobile-native/src/ProjectsScreen.tsx:195-207` | the navigation list's row expansion: `useState({ owner: pages, keys: new Set<string>() })`, read as `expansion.owner === pages ? expansion.keys : new Set()`, toggled by `toggle(row)` | no — React state, discarded on unmount and reset whenever the `pages` instance is replaced |
| `mobile-native/src/navigationTree.ts` | takes `expanded: ReadonlySet<string>` as an argument and flattens the tree | no — a pure function, holds nothing |
| `mobile-native/src/nativeDisclosure.ts` | timeline row disclosure, `createDisclosureStore()` from the package | no — in memory, and a different concern from rail expansion |
| `mobile/src/conversation/project.ts` | "collapsed" only as prose about activity rows | n/a |

Native has storage modules (`nativeDrafts`, `nativeReaderPosition`, `nativeLocation`, `nativeOrganization`, `nativeHubUpgrade`), so the capability exists; none of them stores expansion.

**Conclusion: native renders expansion but stores it nowhere, and expansion already resets on every relaunch.** So D14-3's native adapter is an **in-memory port** — a `Map` behind the same two methods, zero new storage — which preserves today's native behaviour exactly. Giving native an expo-backed store instead would be a visible product change (expansion suddenly surviving a relaunch), and per the plan row's own rule that belongs to a decision, not to a migration slice. Flagging it for the coordinator: if the answer is "native should persist it", it is a separate PR after D14-3, not part of it.

## What follows

- **D14-2**: the core moves to `appwire-client/typescript/state/navigation/store.ts` as `createNavigationStore({ persistence })`; the web adapter at `stores/navigation/store.ts` keeps the zustand hook, the module-level singleton and the `navigationStore`/`useNavigationStore` exports, so the 17 importers still do not change. `projectNodeExpansionKey` is resolved there.
- **D14-3**: native `navigationPages.ts` (392) and `navigationReveal.ts` (131) fold onto the core with the in-memory port above.

## Gates (all foreground, on the code commit `1fff0fc30`; head `447326077` adds only the plan status row)

- `make test-web-browser` — `PASS` for all seven: layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard, retirementguard, skillguard (first run, no rerun needed)
- `npm run lint` in `cmd/evener-hub/frontend` (`biome ci src ../../../appwire-client/typescript`) — `Checked 1316 files`, no errors
- `make lint-package-imports` — `PASS lint-package-imports (1s)`
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — `PASS secret-scan (0s)`; plus `gitleaks detect --no-git --source cmd/evener-hub/frontend/src/stores/navigation` — `no leaks found`
- root `go test -short -count=1 .` — `ok primeradiant.com/evener 21.599s`
- `make test-api-package` and its reachability falsification: **not applicable** — this PR touches no file under `appwire-client/typescript`.

## Refresh onto `d3c537e45`, and a main breakage this found

Merged `origin/main` (`--no-ff`, head `9852a21bf`). One conflict, in the plan status table; resolved union-style, theirs then ours (D23c-2a's row from #1547, then D14-1's). `git diff origin/main...HEAD --stat` is still the three code files plus the one plan row.

Gates rerun on the merge head: `make test-web` PASS (typecheck, test, lint), `make test-native` PASS (iOS bundle exported; 787 native + 771 shared tests), `make lint-package-imports` PASS, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` PASS, root `go test -short -count=1 .` `ok 22.090s`.

`make test-web-browser` is **red through no fault of this branch**: `cmd/evener-hub` does not compile on main. `resumeThreadLocked` (`cmd/evener-hub/app_threadlifecycle.go:474`) calls `trace.stage(...)`/`trace.record(...)` eight times without ever binding `trace` — `undefined: trace` at `:499, 503, 514, 524, 538, 562, 569, 576`. Measured, not inferred: `go build ./cmd/evener-hub/` fails in a clean detached worktree at `ab44f4831` (#1390, "feat(hub): add correlated Resume lifecycle diagnostics") and still at `d3c537e45`, with this branch nowhere in it. The five pure-frontend guards pass (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard); the two that build the real hub — retirementguard and skillguard — fail on that build. All seven passed on `1fff0fc30` before the refresh, off `fa2084eb8`.

The fix looks like one line: `resumeThreadLocked`'s sibling call sites at `:780` and `:802` already read the trace out of the context with `threadLifecycleFromContext(ctx)`, which is nil-safe (`stage`/`record` both return early on a nil receiver). Adding `trace := threadLifecycleFromContext(ctx)` as the function's first statement makes `go build ./cmd/evener-hub/` and `go vet ./cmd/evener-hub/` clean; I measured that, then reverted it (`git diff | git apply -R`) rather than carry another lane's Go fix in a web-only PR. It belongs in its own PR, and main is blocked for every lane until it lands.


obra:
## Round 1 fixes (ed55600ad)

**ACCEPTED 1 — one ordered expansion commit.** Both actions now call one `commitExpansion(persistence, projectKey, expanded)`: publish, then persist, then hydrate what the row revealed. `setExpanded` passes its argument; `toggleExpanded` passes the negation of the current value, which is exactly what its own `m.set(k, !(m.get(k) ?? false))` computed.

*Measurement for the order.* The oracle pins neither: the only two subscribers of this store read `resources`, `attention` and `mode` and never `expanded` (`store.test.ts:1457`, `notifications/index.ts:144`), no navigation test spies on or throws from `localStorage`, and `railExpansion.test.ts:91` pins `saveExpansion` as never throwing, so today's two orders are indistinguishable. I picked publish-then-persist — `setExpanded`'s existing order — because `railExpansion.ts:59-61` states the storage contract as best-effort that "leaves the caller's in-memory state untouched and working for the rest of the session", which only holds if the publish has already happened when the write fails. A port that does throw (a future host's) then cannot cost the caller the toggle. The reason is in the helper's comment.

The hydrate condition is unchanged and was already the same predicate written twice: `setExpanded`'s `expanded && mode === "v2"` and `toggleExpanded`'s `m.get(projectKey) && mode === "v2"` (the new value) are one condition. This closes item 2 of #1573.

**ACCEPTED 2 — port contract only.** Dropped `persistence.test.ts`'s "the store binds the web port when the host supplies none"; `store.test.ts:2340-2373` already seeds `EXPANSION_STORAGE_KEY`, resets and asserts the store read it back. The file is now the recording-port assertions plus the `railExpansionPersistence` round-trip. 6 tests to 5; the navigation directory goes 108 to 107.

**ACCEPTED 3 — one state literal.** `initial()` and its 15-key `Omit<NavigationStoreState, …>` are deleted; the non-action fields are inline in `navigationState(persistence)`, whose `: NavigationStoreState` annotation now does the checking the second list used to. Adding an action no longer means editing an exclusion list.

**NOTED, not done here.** The shared `testing/*` fixture for the recording port waits for D14-2's package suite and D14-3's native port to exist. `projectNodeExpansionKey` stays out of the port and moves into the package as a plain pure helper in D14-2 (#1573 item 1); `Rail.tsx`'s second snapshot over the same localStorage key is yours to file.

## Gates (foreground, on `ed55600ad`)

- `npm run lint` in `cmd/evener-hub/frontend` — `Checked 1316 files`, clean
- `make lint-package-imports` — `PASS (0s)`
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — `PASS (1s)`
- root `go test -short -count=1 .` — `ok 21.748s`
- `make test-web-browser` — five frontend guards PASS (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard); retirementguard and skillguard still FAIL on main's `cmd/evener-hub` build (`undefined: trace`, `app_threadlifecycle.go:499`), unchanged by this branch and tracked in #1571.

obra:
Refreshed onto `6d998fc03` (#1575's hub fix): merge-only, nothing under `cmd/evener-hub/frontend/src` or `appwire-client/typescript` moved in the range. `make test-web` PASS (typecheck/test/lint), `npm run lint` 1316 files clean, `make secret-scan` PASS.

Issue #1556, opened 2026-09-16
handoff watch-pr.sh: the CI-only tail duplicates the main loop's completion check


obra:
docs/superpowers/handoffs/2026-09-16-mobile-landing-sdk-migration/bin/watch-pr.sh carries the CI-completion jq expression and the CI-summary expression twice: once in the main poll loop and once in the inner CI-only tail entered after a RoboRev verdict. #1555 fixed the tail's missing deadline and the `>= 10` magic in both copies, but the duplication remains: the next change to what "CI complete" means has to be made in two places.

Cleanup: fold the tail into the main loop with a `verdict` flag (skip the RoboRev body fetch once the verdict is in; exit when `ci_done && verdict`), so there is one completion predicate. Coordinator tooling only; no product impact. ~15 lines.

obra:
Moot: the script is being removed from the repo (Jesse, 2026-09-16).

PR #1570, opened 2026-09-16
docs(handoff): remove the PR watcher script from the repo

obra:
Jesse (2026-09-16): the watcher script does not belong in the evener repo. #1552 committed it beside the handoff memo; this removes `bin/watch-pr.sh` and repoints the memo's three mentions at the out-of-repo copy the coordinator keeps beside its live ledger (gitignored `.superpowers/`). #1555, which was patching the script in place, is closed; #1556 is closed as moot.

Docs-only. Simplify by reading.


obra:
## Round 1 fixes (902e785ce)

Low accepted and fixed in place (the merge is held by the main break #1571 anyway): both memo mentions now invoke the out-of-repo script by explicit path through `bash`. Docs-only.

Issue #1573, opened 2026-09-16
D14-2 notes: projectNodeExpansionKey still crosses the web boundary, and setExpanded/toggleExpanded order write and publish differently


obra:
Two things D14-1 (#1569, the navigation store's expansion persistence port) leaves for D14-2 (the core moves to `appwire-client/typescript/state/navigation/store.ts`):

1. `cmd/evener-hub/frontend/src/stores/navigation/store.ts:686` still imports `projectNodeExpansionKey` from `shell/rail/railExpansion.ts`, a web module. It is pure (no storage, no module-init side effect) but it cannot enter the package as an import; D14-2 relocates the three-line helper or folds the key naming into the `NavigationPersistence` port.
2. `setExpanded` publishes then persists; `toggleExpanded` persists then publishes. No observable difference today because `writeExpansion` is synchronous and swallows failures, but a port that throws or defers would expose it. D14-2 unifies the order (one rule for every expansion write) when the core owns both.

Found by the D14-1 implementer; filed so the D14-2 brief cites them.

obra:
Item 2 (write/publish order) closed in D14-1 round 1 (#1569 @ed55600ad: one `commitExpansion` helper, publish then persist, reason in its comment). Item 1 (`projectNodeExpansionKey`) is folded into D14-2 (branch claude/sdk-d14-navigation-core, PR after #1569 merges): moved into the package beside the port as a plain pure helper. New note for D14-3: the recording-port test fixture should become `testing/navigationPersistence.ts` alongside `memoryKeybindingDraftStorage` when native's in-memory port needs it.

obra:
Fixed on main by #1584 (merge commit `5abcda20dee30793980cde50b781933019292ebd`); item 1's code relocation and item 2 both verified on current `origin/main`.

Verified: both items the issue left for D14-2 are on main.

1. `projectNodeExpansionKey` is no longer an import across the web boundary. It is exported from the package at `appwire-client/typescript/state/navigation/store.ts:78`, and the web imports it from `@evener/appwire-client/state/navigation` (`cmd/evener-hub/frontend/src/shell/rail/Rail.tsx:15`, `.../rail/railNodes.ts:9`). `cmd/evener-hub/frontend/src/stores/navigation/store.ts` is now a 13-line adapter whose only imports are the package, zustand, and `./persistence` — the `shell/rail/railExpansion.ts` import at the issue's line is gone.

[...]

PR #1498, opened 2026-09-16
refactor(web): the activity panel store no longer imports activitySummary (D17'')

obra:
SDK migration phase D, row D17, re-scoped to **D17''** after measurement (coordinator ruling; the convergence question goes to Jesse as part of D18).

## Why this is not the adoption the plan priced

The plan's D17 row says the "fetch/subscribe body of `stores/activityPanel.ts`" is replaced by the package's `activityList.ts`. The file has no fetch body. It is a pure, multi-ref state machine: `beginFetch`/`beginContinuationFetch` hand out request IDs, `publishFetch(ref, id, result)` applies a caller-supplied result, and the entry carries UI state (`disclosure` reconciled through `reconcileActivityState`, `expandedFoldIDs`, `continuationFailures` per node, `continuationLoadingID`, `pending.kind`, `established`, `staleError: PanelLoadFailure`). The fetches live in `stores/activitySummary.ts` (`refreshRoot`) and `panes/session/chrome/ActivityPanel.tsx` (continuation), both through `threadsStore.listJobs` so they wait out a reconnect.

The package's `ActivityList` is a per-`(ref, threadId)` self-fetching class exposing `{ tree, error: string, loading, unsupported, ended }`. It has none of the surface above, and the web's 22 store oracle tests plus ~20 direct `beginFetch`/`publishFetch`/`setState` sites in component tests drive the request-ID protocol synchronously with no client. A class that publishes from its own request cannot satisfy that protocol, so "thin adapter over `activityList`" and "oracle tests kept" are mutually exclusive. The real convergence (rewriting `activitySummary.ts`'s fencing and the #1096/#195-era oracles onto `ActivityList`, ~2,500-3,500 lines) is D18's question and needs Jesse's call.

## What this PR does instead: cut the import cycle

`activityPanel.ts` imported `./activitySummary` and `activitySummary.ts` imported `./activityPanel` (runtime-safe only because every use was inside a function body). The panel needed exactly two things from the summary side: the summary generation a continuation page begins under (`beginFetch:157`), and someone to pay the page's debt (merged counts or a failure) and release a queued root once the page settles (`publishFetch` tail).

[...]

PR #1584, opened 2026-09-16
refactor(sdk): the navigation store core lives in @evener/appwire-client/state/navigation (D14-2)

obra:
SDK migration phase D, row D14, slice 2 of 3 (plan: `docs/superpowers/plans/2026-09-12-sdk-migration.md`). Builds on D14-1 (#1569, merged as `dd3120512`), which put the store's expansion persistence behind a port. This moves the store itself into `@evener/appwire-client/state/navigation`, extending that subpath rather than adding one. No behaviour change in either app.

## What moved

`cmd/evener-hub/frontend/src/stores/navigation/store.ts` (861 lines) → `appwire-client/typescript/state/navigation/store.ts`, as `createNavigationStore({ persistence })` over `createFrameworkFreeStore`. The body is the same code: the six module-level `let`s that made it a singleton (`activeClient`, `revalidator`, `unsubs`, `bootEpoch`, `bootStartedEpoch`, `manifestFanout`) are now instance state, and `navigationStore.` became `store.`. The three module-level entry points became the store's own:

| Web module export | Store member |
| --- | --- |
| `initNavigation(client, initialize?)` | `init(client, initialize?)` |
| `awaitNavigationConvergence(waiter, targets, opts?)` | `awaitConvergence(waiter, targets, opts?)` |
| `resetNavigationStoreForTests()` | `reset()` |

`NavigationPersistence` moved with it, and `projectNodeExpansionKey` moved beside it (#1573 item 1) as a plain pure helper rather than part of the port, since it also names in-memory keys for `Rail.tsx` and `railNodes.ts` — both now import it from the package. `railExpansion.ts` keeps the storage key, the cap and the swallow-every-failure rules.

## What stayed

- The web adapter keeps its path and every name its importers use: `navigationStore`, `useNavigationStore`, `initNavigation`, `awaitNavigationConvergence`, `resetNavigationStoreForTests`, the `NavigationStoreState` type and `NAVIGATION_INVALIDATION_TIMEOUT_MS`. It is **29 lines**: one `createNavigationStore({ persistence: railExpansionPersistence })` and zustand's `useStore`. Zustand stays entirely web-side. **Zero of the 17 importers changed.**
- `railExpansionPersistence` stays web-side, now typed by the package's port.
- The core schedules its own boot work; the host supplies only the client and the persistence.

## Two measurements that contradict the plan row's guesses

**1. The client port is wider than `Pick<AppwireClient, "request" | "onNotification">`.** Measured against the moved code, the core calls four methods: `request` (every resource read, `requestFor`), `onNotification` (`evener/navigation/invalidated` and `evener/attention/changed`), `onReady` (the reconnect edge, including the same-generation lower-sequence path) and `connect()` (a cold start with no `initialize` handed in). So the port is

```ts
export type NavigationClient = Pick<AppwireClientLike, "connect" | "request" | "onNotification" | "onReady">;
```

**2. No timer port.** The row's sibling risk note asked whether the waiter timer needs injecting. It does not. The store has exactly one deadline — `awaitConvergence`'s `setTimeout`/`Date.now` race, `NAVIGATION_INVALIDATION_TIMEOUT_MS` — and the oracle drives it with **real** timers and short bounds (`store.test.ts:1783` and `:1801` pass `timeoutMs: 20`); `grep -n "useFakeTimers"` over the whole suite returns nothing. The package already calls `setTimeout` in `client.ts`, `state/credentials/instances.ts` and `state/extensions/marketplaces.ts`, so a timer port would be a seam no test needs and no platform requires. Injecting one was declined deliberately.

## Every await in the core, and what fences it

19 awaits. Every boot-path await is fenced on `epoch === bootEpoch` (with `client === activeClient` where the client can be replaced), and both `init()` and `reset()` bump `bootEpoch`, so nothing in flight can write into a store that has moved on.

| # | Await | Fence after it |
| --- | --- | --- |
| 1 | `requestFor` → `client.request` (`:386`) | the revalidator's force-token supersession owns this one; the store writes only through `publishResourceState` |
| 2 | `withProjectRecovery` first `load` (`:448`) | caller-driven read; no epoch fence by design (unchanged) |
| 3-5 | `withProjectRecovery` manifest `load` (`:464`), catalog `Promise.all` (`:466`), candidate `Promise.all` (`:477`) | same; each inner load `.catch(() => undefined)`, and the result is re-derived from store state afterwards |
| 6-7 | `loadPinCatalogPages` forced-page `Promise.all` (`:513`), page loop `load` (`:520`) | caller-driven; `pinCatalogData` throws rather than publishing a partial page |
| 8 | `boot` first `loadManifest` (`:605`) | `if (!manifest \|\| epoch !== bootEpoch \|\| client !== activeClient) return` |
| 9 | `boot` retry `loadManifest` (`:614`) | none needed — the function ends, no write follows |
| 10-11 | `fanOutManifestResources` `manifestFanout.promise` (`:625`), `promise` (`:630`) | no write after either |
| 12 | `hydrateManifestResources` first `runBounded` (`:639`) | `if (epoch !== bootEpoch \|\| activeClient === null) return` |
| 13 | `hydrateManifestResources` second `runBounded` (`:652`) | none needed — the function ends |
| 14 | `hydrateProject` `loadProject` (`:662`) | `if (!resource?.data \|\| resource.error \|\| epoch !== bootEpoch) return` before the one `setState` |
| 15-16 | `runBounded` worker `job()` (`:678`), `Promise.all` (`:681`) | the loop condition itself: `while (cursor < jobs.length && epoch === bootEpoch)` |
| 17 | `awaitConvergence` `Promise.race` (`:854`) | generation-mismatch, revalidator-disposed and not-initialized rejections are classified and returned as success; `mode !== "v2"` re-checked before the fallback |
| 18 | `awaitConvergence` receipt `applyNavigationMutation` (`:874`) | `opts.settled()` decides whether to return or re-arm; the waiter is cancelled in `finally` either way |
| 19 | `awaitConvergence` fallback `applyNavigationMutation` (`:886`) | `if (store.getState().mode !== "v2") return` immediately above it |

`init`'s `client.connect().then(...)` is fenced the same way (`ownedClient !== activeClient || epoch !== bootEpoch`), and the notification, ready and revalidator subscriptions all open with that same pair.

## Oracle status

- `stores/navigation/store.test.ts` — **2758 lines, byte-identical, passing.** It drives `navigationStore.getState()`/`setState()`/`subscribe()` 144 times (433 across the whole frontend); a `FrameworkFreeStore` is the same triple zustand's `StoreApi` is, and `useStore` binds to it exactly as it binds to a vanilla store (the D5 adapter already does this). No fixture changed.
- `stores/navigation/selectors.test.ts`, `shutdownConvergence.test.ts`, `codec.watches.test.ts` — unchanged, passing.
- The 25 other web suites that call `resetNavigationStoreForTests()` — unchanged. The optional persistence parameter D14-1 added for its own test is gone; a fake port is now a second store instance in the package suite, which is strictly better isolation.

## New package suite

`state/navigation/store.test.ts`, 5 tests in the D5/D6 shape:

1. **Isolation** — two instances keep their own expansion, their own port and their own connection; booting one leaves the other at `mode: "unknown"` with a null manifest.
2. **The port at construction and at reset** — read once when built, once more on `reset()`; the port's value is changed out of band before the reset, so a rebuilt store demonstrably holds what the host holds now rather than its own last value.
3. **Publish before persist** — the fixture's `onWrite` hook runs inside `writeExpansion` and reads the store, pinning D14-1 round 1's ordering decision in the package.
4. **A persisted project-node key overrides `default_expanded` at boot** — the port, `projectNodeExpansionKey` and the boot fan-out together.
5. **A reset mid-boot fences the fan-out** — the manifest read is held open, the store is reset, the read released; nothing lands.

## Publication

`state/navigation/index.ts` re-exports the module; `tsconfig.build.json` lists it (arms the reachability assertion); the `./state/navigation` qualification entry gains ESM and CJS type uses plus a smoke that builds a store over a memory persistence port, reads its expansion, toggles through the port and re-reads on reset — with no client wired, so nothing opens a socket. README subpath prose updated. **Falsified:** deleting the barrel's `export * from "./store"` makes `make test-api-package` fail with `shipped module unreachable from every published specifier: state/navigation/store`; restored with `git diff | git apply -R`.

## Gates (all foreground, on the merged head `df98f2f15`)

- `make test-api-package` — qualified, plus the falsification above
- `make test-native` — PASS (iOS bundle exported; 787 native, 771 shared)
- `make test-web-browser` — PASS all seven (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard, retirementguard, skillguard; skillguard on the first run)
- `npm run lint` in `cmd/evener-hub/frontend` — 1318 files, clean
- root `go test -short -count=1 .` — `ok 35.174s`

## What follows

**D14-3** was measured and **withdrawn** — native has no counterpart to `createNavigationStore` at all (no capability gate, no revalidator, no resource map, no boot fan-out, no attention, no expansion), so the shareable logic is ~68 lines out of ~900. **D14-3′** shares those 68 lines instead, without moving a mechanism. Both are recorded in the plan's status table.


obra:
## Review round 1 disposition (df98f2f15)

CI 15/15. RoboRev: one Low (the `state/navigation` barrel header still claims the subpath reaches no store or scheduler), accepted. /simplify (four angles, two reviewers): reuse and efficiency clean for the move (every package helper that exists is called; the per-instance state is what a factory needs); the altitude reviewer confirms verbatim relocation is the right depth for a PR whose oracle must stay byte-identical, and names the debt the move preserved rather than introduced, filed as #1591 (the core owns `connect`/`onReady` and the convergence timer, unlike the D5/D6 host-driven lifecycle; `awaitNavigationTargets` has no production caller) and scoped into #1589 (three staleness vocabularies to fold when the shared fence lands) and #1588 (port the contract suite into the package). Two small items land in round 1 with the Low: `hydrateProject`'s pre- and post-await guards made one expression, and the adapter's dead re-export of the timeout constant dropped. The head is CONFLICTING after the D17'' merge and is being refreshed in the same push.

obra:
## Round 1 fixes (5e541912c)

**Refresh.** Merged `origin/main` at `f1cafa4fe` (D17'' #1498). One conflict, the plan status table: D17'' added its row while this branch had marked D14-1 merged and added D14-2's row. Union-resolved — D14-1 keeps this branch's `**merged** as dd3120512` annotation (the row both sides edited), then D17''s row, then D14-2's. `gh pr view` now reads MERGEABLE.

**RoboRev Low — the barrel header.** Accepted, and it was worse than a stale sentence: `state/navigation/index.ts` still said "Nothing here reaches a store, a scheduler or a host API; the apps' navigation stores are built on top of it", which after this PR is false three ways — the barrel exports a stateful store, that store schedules its own boot work and holds a `setTimeout` convergence deadline, and it takes a host persistence port. The header now says what the subpath publishes and where the host boundary is (two ports; no framework, DOM or storage API). Worth noting how it survived: I *did* try to rewrite this header while writing the PR, with a `str.replace()` whose pattern did not match the file's actual line breaks and which I had not guarded with a count assertion, so it silently did nothing. Every edit in this round asserts its match count first.

**Simplify (a) — one staleness guard in `hydrateProject`.** The entry guard asked `epoch !== bootEpoch || mode !== "v2"` and the post-await guard asked only `epoch !== bootEpoch`, so a store that fell out of v2 during the read still reached the `protocolError` write. Both sides now call one `stale()` closure, with a comment saying what it is asking. Behaviour change is the narrow one the asymmetry left open: a project read that lands after navigation has dropped out of v2 no longer publishes a protocol error into a store that is no longer in v2. No oracle asserted that path, in either direction — `store.test.ts` and the package suite are both unchanged and green.

**Simplify (b) — dead re-export dropped.** `NAVIGATION_INVALIDATION_TIMEOUT_MS` was re-exported by the web adapter with no importer anywhere in `cmd/evener-hub/frontend/src`, `mobile-native/src` or the package (`store.test.ts` hardcodes `10_000` and `timeoutMs: 20`). The constant stays exported from the package, where it is the documented default; the adapter no longer forwards it. The oracle's literals are untouched.

**Not in this PR**, as agreed: the three fence vocabularies fold into #1589; `awaitNavigationTargets` having no production caller; and the two pieces of debt before a second adapter — the core owning `connect`/`onReady` rather than a host-driven ready-generation lifecycle the way D5/D6 do, and `awaitConvergence`'s 10-second shutdown policy sitting in the package with a single web caller.

## Gates (foreground, on the merged head `5e541912c`)

- `make test-api-package` — qualified, plus the falsification: deleting the barrel's `export * from "./store"` fails with `shipped module unreachable from every published specifier: state/navigation/store`, restored
- `make test-native` — PASS (iOS bundle; 787 native, 771 shared)
- `make test-web-browser` — PASS all seven
- `make lint-package-imports` — PASS · `npm run lint` — 1320 files clean
- root `go test -short -count=1 .` — `ok 19.898s`
- `git diff origin/main...HEAD -- .../store.test.ts .../selectors.test.ts` is empty: both web oracles remain byte-identical.

obra:
## Review round 2 disposition (5e541912c)

Round 1 accepted: refreshed onto f1cafa4fe (plan-table union), the barrel header now says what the subpath publishes and where the host boundary is, `hydrateProject`'s two guards are one `stale()` closure (which also closes an untested window where a late project read could write `protocolError` into a store no longer in v2), and the adapter's dead timeout re-export is gone (measured: zero importers). Both oracles byte-identical.

MERGEABLE; CI running on this head; serial-queue position after D6.

PR #1599, opened 2026-09-16
docs(hub): multi-host residual doc comments + drop an unreachable installer-default fallback

obra:
Residuals recorded by the multi-host campaign's landing reviews, cleaned up against `main` (`f1cafa4fe`). **No behavior change**: three documentation corrections and the deletion of one code path proven unreachable.

---

## 1. `sshconn/manager.go` — `Manager.Attached` doc block carried two spliced versions

**Before** (`manager.go:1688-1695`) — a rebase artifact: a first version ("live, not-closed channel" plus a truncated "the hub's background" fragment) spliced onto a second complete version.

**After** (`manager.go:1688-1695`):

```go
// Attached reports whether host currently has a usable channel: one that has
// been established, has not been closed, and has not lost its link. A host that
// has never been Ensure'd is not attached; the hub's background refresh
// attaches lazily, so this converges within one refresh interval. A link-lost
// channel reports detached as soon as markLost closes lost, before the
// supervisor wakes to clear and replace it.
```

## 2. `hubcore/config.go` — subjectless `RemoteHostOnline` comment

**Before** (`config.go:96-98`):

```go
	// controller-to-host channel (component 06). Nil leaves every remote host
	// online (tests).
	RemoteHostOnline func(host string) bool
```

**After** (`config.go:96-98`):

```go
	// RemoteHostOnline reports whether the controller's channel to host is
	// currently attached (component 06). Nil leaves every remote host online
	// (tests).
	RemoteHostOnline func(host string) bool
```

The semantics match the wiring: `cmd/evener-hub/main.go:488` sets it from `sshManager.Attached`, and `cmd/evener-hub/app_rpc.go:54` reads it as `cfg.RemoteHostOnline == nil || cfg.RemoteHostOnline(host.Name)`.

## 3. `appsource/remote_hub_subscription.go` — `routeNotification` comment was pre-round-nine

**Before** (`1112-1139`): claimed *"It observes the subscriber under the lock and sends outside it"* — written before the round-nine/ten/eleven atomic-settle work, and contradicted by the function body (which performs the send inside the `subMu` critical section) and by the same comment's closing paragraph.

**After** (`1112-1146`): describes what the code actually does — a frame `translateNotification` cannot address, or one whose key has no installed subscription, is dropped; delivery is scoped to the client the notification arrived on (stale-generation guard); the non-blocking hand-off shares one `subMu` critical section with the capture (`sub.in` is the per-subscription bounded read-ahead buffer; a full buffer is never waited on, because this is the single drain goroutine and appwire tears the connection down on its own notification-buffer overflow); and a saturated subscription is retired via `cancel` after the lock is released — pump exits, `out` closes, the relay re-reads the thread, so the frame is resynced rather than silently lost. The ordering guarantee is stated as the round-nine test states it (`remote_hub_subscription_roundnine_test.go:154-161`).

Comment only: the function body is untouched.

## 4. `sshconn/version.go` — unreachable installer-default fallback (landing item 04b)

**Decision: the fallback is dead — deleted** (`version.go:1430-1434` before; the function is now at `version.go:1420`).

Evidence:

1. **The empty answer cannot happen.** `command -v evener` exits non-zero when the name is absent — measured on this host: `bash` exit=1, `dash`/`sh` exit=127, stdout empty in every case. `execRunner.Run` turns a non-zero exit into a `*RunError` (`runner.go:102-129`), so the `err != nil` return fires first and the `p == ""` block is never reached. `firstLine` only trims, so a successful lookup always yields a path. The package already relies on this asymmetry: `evenerOnPath` exists precisely because a bare `command -v` reports "missing" as a failure (`deploy.go:408-411`, `evenerPathMissingMarker`), and it is `expectedHubExecutable` that never got that treatment.
2. **Coverage agrees.** `go test ./cmd/evener-hub/internal/sshconn/ -count=1 -coverprofile=...` reports count 0 for the deleted block: `version.go:1431.3,1431.65` (`if dp, ok := m.probeInstallerDefaultExecutable(...)`) and `version.go:1432.4,1433.1` (`p = dp`). No test drove it either — the `[]byte("\n"), nil` stub at `version_test.go:1369` exercises `resolveExecutableName`'s own empty guard, not this one.
3. **The capability is not lost.** Installer-default discovery happens in preflight (`preflight.go:259-262` → `setResolvedTarget`) and reaches `expectedHubExecutable` through the `resolvedTarget` branch (now `1424-1426`), which is consulted before the PATH lookup; `TestRound11…` ("an already-healthy host at the installer default must attach") pins that path today.

Also updated so the artifact stays self-consistent: the function's doc comment (it promised a fallback that cannot fire, and now points at the preflight/resolved-target route), the error text that claimed the installer default had been consulted, and `probeInstallerDefaultExecutable`'s comment (`deploy.go:428-437`), which listed `expectedHubExecutable` as a direct caller.

If the original intent was for `expectedHubExecutable` to probe the installer default *directly*, the fix would be to adopt `evenerOnPath`'s missing-marker convention (`command -v evener || echo <marker>`) — a behavior change, deliberately **not** made here.

**Note for the reviewer:** the recorded 04b premise said `version.go:1425-1435` no longer exists on main and that `git grep expectedHubExecutable origin/main` finds nothing. On `f1cafa4fe` the file and function both exist at exactly that location, so the recorded coordinates were correct and only the "path no longer exists" note was stale.

---

## Gates (all on the final tree, `ab44a649b`)

| Gate | Command | Result |
| --- | --- | --- |
| Formatting | `gofmt -l <5 touched files>` | prints nothing |
| Build | `go build ./...` | exit 0 |
| Vet | `go vet ./cmd/...` | exit 0 |
| Tests | `go test ./cmd/evener-hub/internal/sshconn/... ./cmd/evener-hub/internal/hubcore/... ./cmd/evener-hub/internal/appsource/...` | ok sshconn 2.1s, ok hubcore 16.8s, ok appsource 6.9s; no warnings |
| Lint | `GOLANGCI_LINT_CACHE=$(scripts/lib/golangci-lint-cache.sh) golangci-lint run --allow-parallel-runners ./cmd/evener-hub/internal/sshconn/... ./cmd/evener-hub/internal/hubcore/... ./cmd/evener-hub/internal/appsource/...` | `0 issues.` (pinned 2.13.1 per `.tool-versions`) |
| Generated artifacts | `make lint-generated` | `PASS lint-generated` |

Scoped package tests were used rather than a whole-module race suite: the host is at load ~67 with 226 worktrees live.


obra:

**CI** — all 16 checks green at `ab44a649b` (15 CI checks plus RoboRev's own `roborev` check).

**Review** — RoboRev's verdict for `ab44a649b`: **"No issues found."** (3 reviewers, synthesis codex)

**What this lands** — four residuals recorded as immediate follow-ups when the multi-host stack landed:

- `sshconn/manager.go` — the `Manager.Attached` doc comment carried two spliced versions of the same sentence block (a rebase artifact); the single coherent block is restored.
- `hubcore/config.go` — `RemoteHostOnline`'s field comment started mid-sentence with no subject; it now names what it reports and what its nil value means.
- `appsource/remote_hub_subscription.go` — the `routeNotification` doc comment claimed the send happens outside the lock, which the body contradicts (it sends inside the `subMu` critical section); the comment now describes the current contract, including the client-scoped stale-generation skip and saturated-subscription retirement. Comment-only: the diff contains no non-comment line in this file.
- `sshconn/version.go` — the installer-default fallback in `expectedHubExecutable` was unreachable: `command -v evener` exits non-zero when the name is absent and the runner maps that to an error, so the empty-answer path the fallback filled cannot occur. Deleted, with the coverage showing the block was never executed and the installer-default behavior still provided through preflight's resolved target (pinned by the existing round-8/round-10 deploy tests). The adjacent `probeInstallerDefaultExecutable` comment no longer lists this function as a direct caller.

**Up-to-date override** — the branch is behind `main`. Overlap check, run immediately before this comment:

```
$ gh pr view 1599 --json mergeable,mergeStateStatus --jq '.mergeable, .mergeStateStatus'
MERGEABLE
BEHIND

$ comm -12 <(gh pr view 1599 --json files --jq '.files[].path' | sort) \
           <(git diff --name-only "$(git merge-base origin/main origin/residual/go-lows)" origin/main | sort)

$ git diff --name-only "$(git merge-base origin/main origin/residual/go-lows)" origin/main | wc -l
19
```

`main` changed 19 files since the merge base; none of them is a file this PR touches.

**Approval** — no human approval is recorded; this is the campaign coordinator's landing decision under the campaign's rule: no unresolved Critical/High/Medium, CI green, behind handled.

PR #1600, opened 2026-09-16
docs(multi-host): reconcile specs 03-05 with the landed implementation (#1276 residuals)

obra:
Residuals from the #1276 landing statement, verified against current `main` (`5abcda20d`). Docs only; no code changes. Line numbers under "what the text said" are pre-change.

## 1. Spec 03 — `appwire.ValidRefPart` — fixed in text

**What the text said** (`docs/superpowers/specs/2026-09-14-multi-host-03-host-config.md:290,299,507`): "either export a helper in `appwire` (e.g. `func ValidRefPart(s string) bool`) or validate via the public `appwire.ParseRef`"; "**Settled: export `appwire.ValidRefPart`**"; and a `~10–20 LOC` budget line — all reading as instructions to build it.

**What the code does**: `appwire.ValidRefPart` ships at `appwire/refs.go:18`. `hostreg.ValidateName` consumes it (`cmd/evener-hub/internal/hostreg/hostreg.go:78`) and layers the reserved-name / `"."` / `".."` checks; it is reached through `New`/`Add`/`AddWithUpstreams` (`hostreg.go:126,211,242`) and, from config, through `validateHostConfigs` → `hostreg.New` (`cmd/evener-hub/config.go:196-222`).

**What changed**: the paragraph now states the shipped design and its call sites, the LOC line marks the export landed, and acceptance criterion 3 names `appwire.ValidRefPart` instead of the unexported `refPartPattern`.

## 2. Spec 04 — reconnect vs explicit attach — fixed in text

**What the text said** (`docs/superpowers/specs/2026-09-14-multi-host-04-ssh-connection-manager.md:409-410`): "the manager transitions to `reconnecting` and re-runs `Ensure` on the same `dest`, which re-attaches as a client" — never defining whether a reconnect repeats the preflight / `/api/health` probe / deploy+restart ladder, nor what the caller observes. The bootstrap section also still claimed "the shipped 04b `ensureOnce` has no bootstrap branch" (`:1013-1015`).

[...]

PR #1603, opened 2026-09-16
feat(hub): enforce lazy attachment through attached-only seams

obra:
## What this enforces

Multi-host specs 03–06 mandate that a remote host attaches **lazily**: only an
explicit attach trigger may dial, and every other remote path reads an
already-attached channel or fails with a typed unavailable error. The landed
code dialed on many implicit paths, so merely opening the hub attached every
configured host.

## Requirement → what main did → what now happens

| Spec requirement | On main | Now | Pinned by |
|---|---|---|---|
| 04 §"Go surface" (`PreflightIfAttached`, `HandshakeIfAttached` beside `ClientIfAttached`); 03:363-380 wiring | Only `ClientIfAttached` existed; `RemoteHostFacts` was `Ensure`-backed) | `Manager.PreflightIfAttached` / `Manager.HandshakeIfAttached` added, sharing one installed-live-channel predicate; `Channel.Handshake()` returns the `InitializeResponse` captured at attach | `TestPreflightIfAttachedReturnsOnlyALiveChannel`, `TestHandshakeIfAttachedReturnsOnlyALiveChannel` |
| 03:363-380 `hubcore.WebConfig` seams | only `RemoteHosts`, `RemoteHostClient`, `RemoteHostFacts` (Ensure), `RemoteHostOnline` | adds `RemoteHostClientIfAttached` (`Manager.ClientIfAttached`) and `RemoteHostHandshake` (`Manager.HandshakeIfAttached`); `RemoteHostFacts` is now `Manager.PreflightIfAttached` (non-dialing) | `TestNewHubSourceRegistryDirectReadAgainstUnattachedHostIsUnavailable` |
| 05:696-712 snapshot must not force attachment | `refreshRemoteThreadSnapshot` iterated `s.sources.All()` with no gate; resolver wired to `Ensure` | skips an unattached source without a call and carries last-known-good rows forward; the source's own resolver is attached-only | `TestRefreshRemoteThreadSnapshotSkipsUnattachedSourceWithoutACall` |
| 05:701-715 empty-filter `thread/list` gate | empty `SourceIDs` resolved every source through the dialing resolver | empty filter ⇒ attached-only, no dial; explicit host in `SourceIDs` ⇒ attaches before the source is called | `TestHubThreadListNonExplicitFanOutSkipsUnattachedRemoteHost`, `TestHubThreadListExplicitSourceIDsAttachesTheHost` |
| 05:717-747 every other remote call non-dialing | `RemoteHubClientFunc` = `Ensure` + `ch.Client()` for all calls | `RemoteHubSource.resolveClient` answers from `SetHostClientIfAttached` (production) and returns `SessionUnavailable("remote hub unavailable: <host>")` without dialing; wired client func is a test fallback | `TestRemoteHubSourceUnattachedCallIsUnavailableWithoutDialing`, `TestRemoteHubSourceAttachedCallUsesAttachedOnlyClient` |
| 05:392-408 capability probe reads attached-only handshake/preflight facts | probe read `Features` off the dialing facts closure; no handshake seam | `SetHostHandshake`/`RemoteHostHandshake`; probe reads ProtocolVersion/ServerInfo/SourceID/Features from the handshake seam and OS/Arch from the preflight seam | `TestHostCapabilitiesReadsAttachedHandshakeFacts` |
| 06 AC13 direct read/mutation against unattached host | dialed | typed `SessionUnavailable` naming the host, zero dials | `TestNewHubSourceRegistryDirectReadAgainstUnattachedHostIsUnavailable`, `TestRemoteHubSourceUnattachedCallIsUnavailableWithoutDialing` |

## Tests re-pinned

None. No existing assertion was weakened, changed, or deleted. The negative
tests are additive; the pre-existing suite (including the 2 877-line
subscription suite and the probe suite) passes unchanged. Where a test injects
only a client function and no attached-only seam, `resolveClient` falls back to
that function, so the existing tests keep their meaning while production never
dials.

## Failing-first evidence

- `sshconn` new tests failed to build against pre-change code:
  `m.PreflightIfAttached undefined`, `ch.Handshake undefined`.
- With the implementation stashed, the hub/appsource tests failed to build
  (`unknown field RemoteHostClientIfAttached`, `SetHostClientIfAttached
  undefined`).
- With only the attachment gate temporarily disabled:
  `TestHubThreadListNonExplicitFanOutSkipsUnattachedRemoteHost` failed
  ("unattached remote source was listed 1 times; want 0") and
  `TestRefreshRemoteThreadSnapshotSkipsUnattachedSourceWithoutACall` failed
  ("...listed 1 times by the snapshot; want 0").

## Specs updated

`03-host-config.md` §"Wiring" item 4, `05-remote-hub-source.md` the three
snapshot/list/direct-call status paragraphs, and `06-fleet-view.md` the snapshot
status paragraph now say **shipped** and describe what is actually wired.

## Still not implemented (stated plainly)

- Component 06's Connect action (`evener/host/attach`) does not exist yet, so an
  explicit host in `thread/list`'s `SourceIDs` is currently the only attach
  trigger in production.
- The capability probe's per-root `LaunchResolved` map (05 §"Capability probe")
  is not implemented; unchanged by this PR.

## Gates

- `gofmt -l` on touched trees: empty
- `go build ./...`: exit 0
- `go vet ./cmd/...`: exit 0
- `go test ./...` in `cmd/evener-hub` (whole module): ok
- `golangci-lint run --config .golangci.yml ./cmd/evener-hub/...`: 0 issues
- `make lint-generated`: PASS


obra:
## Follow-up: restored the generation guard for the capability facts

New commit `b333e6db765f0df6dd3b69e983dfc527663de9b4` (PR head).

### The dropped guard, restored

The first commit's attached-only `RemoteHostFacts` did:

```go
pf, ok := sshManager.PreflightIfAttached(host)   // whatever channel is installed NOW
...
return remoteHostFacts(pf, client.Features()), nil // `client` resolved a moment earlier
```

A supervisor reconnect in that window produced facts from one generation and
`client.Features()` from another, and the probe cached the mixture — the hazard
the removed `ch.Client() != client` check existed to prevent.

### The fix (atomic, not check-then-read)

A `ClientIfAttached`-then-`PreflightIfAttached` pair would still race (two lock
acquisitions). Instead:

- `sshconn.Manager.ChannelIfAttached(name) (*Channel, bool)` returns the
  installed live channel under **one** lock acquisition. `PreflightIfAttached`
  and `HandshakeIfAttached` already delegate to that same predicate.
- `main.go` reads **one** `ChannelIfAttached` value and takes the facts from the
  same channel whose `Client()` is the probe's `client`:
  - `RemoteHostFacts` refuses with a typed
    `SessionUnavailable("remote hub unavailable: <host>")` on mismatch or when
    nothing is attached (and never dials);
  - `RemoteHostHandshake` reports `false` on mismatch, so the probe falls back
    to the facts it already pinned to the same client, never to another
    generation's handshake.

### Handshake seam

`RemoteHostHandshake` / `appsource.HostHandshakeFunc` now take the probe's
`client`:

```go
RemoteHostHandshake func(host string, client *appwire.Client) (appwire.InitializeResponse, bool)
```

The spec's one-argument shape (`func(host string) (appwire.InitializeResponse, bool)`,
05 §"Capability probe" / 03 §Wiring) cannot express the check at all, so the
signature carries the generation to pin against. Spec 03/05 text updated to the
shipped shape with the reason.

### Tests (additive; no existing or new attached-only test weakened)

- `TestRemoteHostFactsRefusesAReconnectedChannelGeneration` — the installed
  channel's client differs from the probe's; asserts typed `SessionUnavailable`
  for the facts and `false` for the handshake. **Fails first:** with the pointer
  check temporarily removed it reports
  `facts for a reconnected generation: error <nil>=<nil>, want WireError`.
- `TestRemoteHostFactsAndHandshakeUseTheProbesGeneration` — matching generation
  yields that channel's preflight/handshake.
- `TestChannelIfAttachedReturnsOnlyALiveChannel` — pins the primitive (live-only,
  non-dialing) and that `ClientIfAttached`/`PreflightIfAttached`/
  `HandshakeIfAttached` all agree with the one `ChannelIfAttached` value.

### Gates (all after the follow-up commit)

- `gofmt -l` touched trees: empty
- `go build ./...`: exit 0
- `go vet ./cmd/...`: exit 0
- `go test ./...` in `cmd/evener-hub` (whole module): ok (hub 178s, appsource 7s,
  sshconn 1.5s, hubcore 16.5s)
- `golangci-lint run --config .golangci.yml ./cmd/evener-hub/...`: 0 issues
- `make lint-generated`: PASS

obra:
## Merged main (`5454fb2c0`)

`git merge origin/main --no-ff` produced exactly one conflict; the other four named surface files auto-merged with both intents preserved.

- `cmd/evener-hub/internal/hubcore/config.go` — conflicted on the `WebConfig` seam block: HEAD adds `RemoteHostHandshake`, main (#1599) rewrote the `RemoteHostOnline` doc comment. Resolved by keeping BOTH — HEAD's `RemoteHostHandshake` field followed by main's "reports whether the controller's channel to host is currently attached" comment on `RemoteHostOnline`.
- `cmd/evener-hub/internal/appsource/remote_hub_subscription.go` — auto-merged (main #1599 comment rewrite + the PR's seam text).
- `cmd/evener-hub/internal/sshconn/manager.go` — auto-merged (main #1599 removed a duplicated `Attached` comment; the PR's `ClientIfAttached` / `ChannelIfAttached` / `PreflightIfAttached` / `HandshakeIfAttached` accessors survive).
- `docs/superpowers/specs/2026-09-14-multi-host-03-host-config.md` and `...-05-...md` — auto-merged (main #1600 reconciliation + the PR's attached-only text).

No conflict appeared in a file outside the PR's set. `go build ./...` was verified before staging, no file containing conflict markers was staged, and the merge is its own commit (`12b995e4e`, parents `b333e6db7` + `5454fb2c0`).

## Round one addressed (`d7c345448`)

- **Medium — explicit-list attach failure returned raw** (`cmd/evener-hub/app_threadlist.go:115`). The explicit host-targeted `thread/list` attach called the `Ensure`-backed `cfg.RemoteHostClient` and returned its error unclassified, so an `sshconn.ErrSSHStart`/`ErrRestart` chain or a transport loss reached the caller raw and no auto-resume/refusal gate could attribute it. Added `(*appsource.RemoteHubSource).MapAttachError` (routing through the existing `mapConnectError`) and an `attachErrorClassifier` seam in the fan-out: the caller's own context ending stays raw (mirroring `RemoteHubSource.call`), every other attach failure is classified to the typed `SessionUnavailable`. Pinned by `TestHubThreadListExplicitAttachFailureIsSessionUnavailable` (ssh-start chain and deadline-exceeded chain). Failing-first text before the fix:
  - `explicit attach failure *fmt.wrapError=sshconn: ssh channel start failed: host "alpha" link dropped before the channel was usable, want a typed WireError`
  - `explicit attach failure *fmt.wrapError=attach alpha: context deadline exceeded, want a typed WireError`
- **Low (a) — `RemoteHostFacts` reported caller cancellation as host unavailability** (`cmd/evener-hub/main.go`). Extracted the inline closure into `remoteHostFactsIfAttached`, which returns `ctx.Err()` raw before running the lookup. Pinned by `TestRemoteHostFactsReturnsCallerCancellationRaw`. Failing-first text before the fix: `canceled-context facts error appwire.WireError=remote hub unavailable: alpha, want context.Canceled`.
- **Low (b) — spec named the wrong accessor** (`docs/superpowers/specs/2026-09-14-multi-host-03-host-config.md:370`). The doc said `RemoteHostFacts` was backed by `sshManager.PreflightIfAttached`; the shipped wiring reads one `sshManager.ChannelIfAttached` value with the `ch.Client() == client` generation guard, and `PreflightIfAttached` has no production caller. Corrected that sentence and the "attached-only additions" list to `Manager.ChannelIfAttached`. Corrected the same false claim in spec 05 as well (the `SetHostFacts` bullet and the probe paragraph). Code unchanged — the generation guard is the point.
- **Low (c) — `HostHandshakeFunc` comment** (`cmd/evener-hub/internal/appsource/remote_hub_probe.go:57-68`). Dropped the false "appwire keeps the client's Features copy privately" clause (`(*Client).Features()` is exported and is called at `main.go:754`); the comment now states the real constraint — `ProtocolVersion`/`ServerInfo`/`SourceID`/`Features` are handshake-only because the connection-scoped initialize cannot be re-run on the already-initialized channel. Same correction applied to the adjacent test comment. The handshake's `Features` assignment was left in place: redundancy with the facts path could not be proven by a test, so per the brief the code stands.

Prior dispositions untouched: the attached-only seams, the `ch.Client() == client` generation guard in both facts and handshake, the non-dialing `main.go` wiring, `resolveClient`'s typed `SessionUnavailable` that never dials, the empty-filter thread/list gate with explicit-SourceIDs-still-attach, and the snapshot gate keeping last-known-good rows.

Verification:
- `go build ./...` → exit 0
- `go test ./cmd/evener-hub/internal/sshconn/ -race -count=1 -timeout 20m` → ok (2.865s)
- `go test ./cmd/evener-hub/internal/appsource/ -race -count=1 -timeout 20m` → ok (28.857s)
- `go test ./cmd/evener-hub/ -count=1 -timeout 30m` → ok (219.281s, re-run on the final tree)
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS (0 findings)

obra:
Merged main (`caa7c7e1a`, PR #1606)

- Merge commit `72f7eb5`. Git auto-merged cleanly — the only shared file was `docs/superpowers/specs/2026-09-14-multi-host-05-remote-hub-source.md`, and no conflict arose. Both intents survive: main's activity-tree decode-gate implementation-status text (the `activityTreeRecognized`/`activityTreeDecodable` recognition boundary) and this branch's attached-only paragraphs (the `(host, client)` handshake seam with its generation guard, the `ClientIfAttached` resolver, and the empty-filter gate). No files outside the PR's file set were touched by the resolution; `go build ./...` and `gofmt -l cmd/evener-hub/` were clean before staging the fix.

Round two addressed (`36381726f`)

- **Medium 1 — handshake failure silently ignored (`remote_hub_probe.go`).** Confirmed: when `handshakeFn(s.id, client)` returned `ok == false`, the block was skipped and execution fell through to `s.probe = &remoteHubProbe{client: client, caps: caps.clone()}`, caching and returning a snapshot that paired this client's wire reads with no handshake facts (and a later call on the same client would serve it) instead of refusing the generation change. Fixed: `ok == false` now returns the typed `appwire.SessionUnavailable("remote hub unavailable: " + s.id)` — the same typed refusal `resolveClient` and the facts path (`remoteHostFactsForChannel`) produce — and caches nothing. Pinned by the new `TestHostCapabilitiesHandshakeFalseRefusesWithoutCaching`, which asserts both the typed error and that `source.probe == nil`. Failing-first against the unfixed head:
  ```
  --- FAIL: TestHostCapabilitiesHandshakeFalseRefusesWithoutCaching (0.00s)
      remote_hub_attached_only_test.go:161: HostCapabilities error <nil>=<nil>, want WireError
  ```
  (the probe returned a partial snapshot with a nil error instead of refusing).

- **Medium 2 — `HubVersion` overwritten with a literal constant (`remote_hub_probe.go:159`).** Verified the chain: `cmd/evener-hub/main.go:46` `const Version = "0.1.0"` → `app_rpc.go:386` `appserver.ServerConfig{Version: Version}` → `internal/appserver/server.go:521` `ServerInfo: appwire.ServerInfo{Name: …, Version: s.cfg.Version}`, so the remote hub's `ServerInfo.Version` is the package constant, not its build identity. The facts path carries the real build: `main.go:723` `HubVersion: buildinfo.Version()` (pinned by `TestRemoteHostFactsReportRestartedHubVersion`). The handshake block's `caps.HubVersion = hs.ServerInfo.Version` therefore clobbered the buildinfo value with `"0.1.0"` for every remote host. Fixed by the first suggested shape: dropped the override so `facts.HubVersion` remains the single version source; `ProtocolVersion`, `HubSourceID`, and `Features` still come from the handshake. The block comment now states that split. Re-pinned `TestHostCapabilitiesReadsAttachedHandshakeFacts`: the old assertion (`caps.HubVersion != "9.9.9"`) pinned the clobber; it now asserts `HubVersion == "preflight-ver"` (the facts value) while still stubbing a distinct `ServerInfo.Version = "9.9.9"` to prove it is NOT copied — `ProtocolVersion`/`HubSourceID`/`Features` still assert the handshake values. Failing-first against the unfixed head:
  ```
  --- FAIL: TestHostCapabilitiesReadsAttachedHandshakeFacts (0.00s)
      remote_hub_attached_only_test.go:123: HubVersion = "9.9.9", want the facts seam value preflight-ver (not the handshake ServerInfo.Version 9.9.9)
  ```
  Also corrected the now-contradictory seam doc comments in `remote_hub_probe.go` (`HostHandshakeFunc`), `remote_hub_source.go` (the `handshake` field), and `sshconn/manager.go` (`HandshakeIfAttached`, `Channel.Handshake`) so they no longer claim the probe consumes `ServerInfo`.

- **Follow-up note (out of scope, for the owner).** The hub's own `ServerInfo.Version` reporting the package constant `"0.1.0"` rather than `buildinfo.Version()` is a separate question. `cmd/evener-hub/main.go`'s `Version` constant and the `ServerConfig` wiring were deliberately NOT changed here (outside this PR's file set); if the hub's advertised version should track the running build, that is its own change with its own blast radius.

Verification: `gofmt -l cmd/evener-hub/` → empty; `go build ./...` → exit 0; `go vet ./cmd/evener-hub/...` → exit 0; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=1 -timeout 20m` → `ok … 3.405s`; `go test ./cmd/evener-hub/internal/appsource/ -race -count=1 -timeout 20m` → `ok … 59.162s`; `go test ./cmd/evener-hub/ -count=1 -timeout 30m` → `ok … 214.046s`; `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → `PASS lint (1 modules, 86s)`, 0 findings.

obra:
Round three addressed (`7f0129b`).

- **Medium — no production path triggered host attachment.** Added the backend controller-local method `evener/host/attach` (`cmd/evener-hub/app_host_attach.go`, params `{name}`, `ScopeHub`), wrapping the Ensure-backed `cfg.RemoteHostClient` dialing seam, classifying attach failures through the landed `MapAttachError`/`attachErrorClassifier` chain (ssh-start/deadline chains → typed `SessionUnavailable`; caller-context cancellation raw), idempotent, returning `attached` + facts. Registered through the router (pinned by `TestHubRPCRegistersExpectedHandlerSet`) and recorded in the protocol catalog/`types.gen.ts`; deliberately **not** added to `remoteHostAdminMethods` (negative assertions in `TestHostAdminAllowListMatchesCatalog` and `TestHostAdminMutationClassificationMatchesAllowList`). New `cmd/evener-hub/app_host_attach_test.go` covers dial-once+facts, idempotency, unknown host, no-seam, classified failure, and raw cancellation.
- **Medium — frontend trigger (the shipped client).** `Spawn.tsx` now issues `evener/host/attach` when a remote host is selected, and renders an enabled **Connect `<host>`** affordance for each offline host. Shape chosen: the host picker's offline `<option>`s stay `disabled` for a spawn (component 06b preserved), so selection alone can never attach a never-attached host — the Connect button is the concrete enabled attachment affordance component 06 requires ("never merely a disabled spawn row and never a control that cannot fire"). On success the hub's attach event flips the manifest's `online` flag and the button disappears; a failure surfaces a toast. Failing-first: with the selection trigger removed, `selecting a remote host issues evener/host/attach for it` fails `expected true, received false` at `Spawn.test.tsx:6126`; the Connect and failure-toast tests are new behavior.
- **Medium — `EventAttached` did not refresh the remote cache.** `hubSSHStateInvalidation` gained an `onAttach` callback, run only on `EventAttached`, which non-blocking-pokes `remotePoke` next to navigation invalidation (`main.go`). Failing-first: with the callback neutered, `TestHubSSHStateInvalidationPokesRemoteRefreshOnAttach` fails `kind "attached" poked refresh=false, want true`.
- **Low — snapshot skip-branch carry-forward untested.** Added `TestRefreshRemoteThreadSnapshotKeepsLastKnownGoodWhenHostDetaches`: list while attached, flip the attached lookup false, snapshot again, assert the rows persist in both `fetch.threads` and `fetch.sources["alpha"].Threads` with `Complete == false`. Honest result: it **passed on the first run** — this is a regression pin for behavior that already worked, not a bug fix.
- **Low — spec/comment drift.** Corrected spec 03 + spec 05 and the `manager.go` / `appsource` / `hubcore` comments: the production backing is `ChannelIfAttached` plus the `remoteHostFactsForChannel` / `remoteHostHandshakeForChannel` closures, with the generation guard (`ch.Client() == client`) at the call site; `PreflightIfAttached` / `HandshakeIfAttached` have no production caller. Wiring unchanged.

Verification:
- `gofmt -l cmd/evener-hub/` → empty; `go build ./...` → 0; `go vet ./cmd/evener-hub/...` → 0
- `go test ./cmd/evener-hub/internal/sshconn/ -race -count=1 -timeout 20m` → ok (2.6s)
- `go test ./cmd/evener-hub/internal/appsource/ -race -count=1 -timeout 20m` → ok (23.7s)
- `go test ./cmd/evener-hub/ -count=1 -timeout 30m` → ok (151.3s)
- `GOLANGCI_LINT_CACHE=… MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS lint (1 module, 46s)
- `npx biome ci src ../../../appwire-client/typescript` → 0; `npx tsc --noEmit` → 0
- `make test-web` → PASS web-typecheck, web-test, web-lint
- `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` → PASS web-layoutguard, web-overflowguard, web-shellguard, web-spawnguard, web-transcriptscrollguard, web-retirementguard, web-skillguard
- `make lint-generated` → PASS

Merge: `origin/main` (caa7c7e1a) was already merged into the branch; `git merge origin/main` reported "Already up to date", no merge commit.

obra:
Round four addressed (`e9d61551c`).

- **High — host-routing origin guard on remote dials.** Origin metadata now exists: `evener hub attach --stdio` presents `X-Evener-Bridge: 1` on its loopback dial (`cmd/evener-hub/attach.go`), and the hub's `/rpc` edge reads it at accept and stamps origin `"bridge"` into the request context (`WebServer.serveAppWireRPC`, `cmd/evener-hub/web.go` → `host_routing_origin.go`). One shared guard, `guardRemoteHostDial`, sits inside the single Ensure-backed seam `dialRemoteHost`; `hubHostAttach` and the explicit-SourceIDs `thread/list` attach both dial through it and are refused with typed `InvalidParams` naming the origin before any dial. Failing-first: `TestHostAttachRefusesRemoteOriginatedDial` and `TestHubThreadListExplicitAttachRefusesRemoteOriginatedDial` — with the guard neutered the attach dials and returns nil, the list reaches the dial; with it, 0 dials and a typed refusal. `TestServeAppWireRPCStampsBridgeOrigin` pins edge→ctx; the bridge round-trip asserts the marker is sent; a direct (unmarked) request stays local and still dials.
- **Medium 1 — attach payload field name.** Spec 06 §"Go changes" item 5 documents `HostAttachParams{Host string}`, so the wire field was renamed `name` → `host` in `appwire/types.go`, regenerated into `appwire-client/typescript/types.gen.ts` + `docs/appwire-protocol.md`, and updated in `Spawn.tsx` / `Spawn.test.tsx`.
- **Medium 2 — response omitted host identity/handshake.** Added `host`, `serverName`, `serverVersion` to `HostAttachResponse`, populated from the attached channel's handshake via the `RemoteHostHandshake` seam (backed by `ChannelIfAttached`). `hubVersion` stays facts-owned (the handshake's `ServerInfo.Version` is the package constant, documented as such). Pinned by `TestHostAttachPopulatesHostIdentityAndHandshake`.
- **Medium 3 — terminal attach errors were generic.** `hostAttachWireError` maps the manager's terminal sentinels: protocol/contract refusals (`ErrProtocolIncompatible`, `ErrUnsupportedHost`, `ErrLaunchContract`, `ErrPreflightDecode`, `ErrHostNotFound`, `ErrHostAddr`, `ErrManagerClosed`) → typed `Unavailable`/`actionUnavailable`; deploy failures (`ErrDeploy`, `ErrVersionMismatch`, the dirty-controller refusal, matched via the newly exported `sshconn.ErrControllerDirty`) → typed `HubLaunchError`/`hubLaunch`. Transport loss keeps the existing `SessionUnavailable`. Failing-first: `TestHostAttachTerminalFailuresAreTypedWireErrors` (before the mapping these reached the browser as `internal`).
- **Low 1 — successful dial reported as a failed Connect.** A failure reading the post-attach facts now returns `HostAttachResponse{Attached: true, …}` (facts omitted), keeping the dial authoritative; only a failed dial errors. Pinned by `TestHostAttachKeepsSuccessfulDialWhenFactsReadFails`.
- **Low 2 — stale comments/docs.** Refreshed the `remoteThreadListBudget` comment (the source resolves through the attached-only lookup and never dials; the fan-out's explicit trigger does the attach) and spec 04 §"Go surface" (added `ChannelIfAttached`; `RemoteHostFacts`/`RemoteHostHandshake` are built on it with the `ch.Client() == client` call-site guard, not `PreflightIfAttached`/`HandshakeIfAttached`).

Verification: `gofmt -l cmd/evener-hub/` empty; `go build ./...` 0; `go vet ./cmd/evener-hub/...` 0; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=1` PASS; `go test ./cmd/evener-hub/internal/appsource/ -race -count=1` PASS; `go test ./cmd/evener-hub/ -count=1` PASS (151s); module-lint PASS (1 module, 0 findings); frontend biome/`tsc --noEmit` clean; `make test-web` PASS (web-typecheck/web-test/web-lint); `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` PASS (web-layoutguard, web-overflowguard, web-shellguard, web-spawnguard, web-transcriptscrollguard, web-retirementguard, web-skillguard); `make lint-generated` PASS.

obra:
Round five addressed (`8916603c9`).

**High — origin guard bypassed through already-attached remote sources.** Fixed. Enumerated every path by which a hub-scoped handler can send a message over a remote-hub client, and where each now meets the guard:

- `dialRemoteHost` (`cmd/evener-hub/host_routing_origin.go:77`) — the Ensure-backed dial, guarded by `guardRemoteHostDial` (round four). Reached by the explicit attach trigger (`app_host_attach.go:64`) and the explicit-SourceIDs thread/list fan-out (`app_threadlist.go:132`).
- `RemoteHubSource.resolveClient` (`internal/appsource/remote_hub_source.go:255`) — the shared remote-routing/client seam. Every remote dispatch resolves its client here, and it now calls the new `appsource.guardRemoteDispatch` (`internal/appsource/host_routing_origin.go:47`) before either the attached-only lookup or the test/connector fallback. Covered below it:
  - `call` — reads: ListThreads, ReadThread, ListTurns, ListModels, item paging (`remote_hub_source.go:227`)
  - `callOn` — the capability probe's wire reads via `HostCapabilities` (`remote_hub_probe.go:92`)
  - `mutationCall` — thread/turn lifecycle mutations (`remote_hub_mutations.go:16`)
  - `AdminCall` / `AdminMutationCall` — the `evener/host/request` forwarding used by the admin proxy (`remote_hub_source.go:1528` / `1606`, reached from `app_host_admin.go:318/320`)
  - `SubscribeThread` / `SubscribeHostNotifications` — subscription attach (`remote_hub_subscription.go:170`, `remote_hub_source.go:1736`)
- Thread ops that name a remote source by ref resolve through `sourceForThreadWithDeletionFence` (`app_rpc.go:602/783/829`) and then land on the source's `call`/`mutationCall` — the same seam.
- Background/internal traffic (the host notification fan-out, the poke-on-attach refresh, the 30s snapshot walk) carries no request origin, so it is unaffected.

Guard shape: one shared guard, `appsource.guardRemoteDispatch(ctx)`, refusing a non-empty origin with the typed `appwire.InvalidParams` naming it; `guardRemoteHostDial` keeps the dial-specific refusal. Cached local state (snapshot rows, fleet manifests, tombstones) never reaches the seam, so bridge-originated requests still read it.

Failing-first: with `guardRemoteDispatch` neutered, `TestHostAdminRequestRefusesRemoteOriginatedForwarding` fails with `expected a wire error with code -32602, got nil` (the bridge-originated `evener/host/request` was forwarded) and `TestRemoteOriginatedDispatchIsRefusedAtTheClientSeam` fails with `ReadThread error <nil>=<nil>, want WireError`. Green with it: zero messages reach the remote client (harness `calls()` = initialize only) while the same call from an unmarked request still forwards, and `TestRemoteOriginatedRequestStillReadsCachedSnapshotRows` proves a bridge-originated cached snapshot read is untouched.

**Medium 1 — spoofable `X-Evener-Bridge` marker: contested, with the trust basis restated.** I traced the token model: `evener hub attach --stdio` reads the host hub's own capability token from its state root (`attach.go:106` → `readAuthToken` → `hubedge.TokenFileName`) and presents it as `Authorization: Bearer` on the loopback `/rpc` dial (`attach.go:141-151`); the edge's `hubedge.AuthGuard` accepts either the per-hub cookie (browser) or that same Bearer token (bridge, TUI, CLI scripts) (`internal/hubedge/auth_token.go:121-125`, `:164`). The attach connection is therefore NOT distinguishable from a local Bearer client at the edge — the token is one master secret shared by every client role. Binding origin to "Bearer" would classify local scripted clients as remote and break their fan-out; binding it to connection identity needs a distinct bridge credential, the tracked follow-up the design already names. So I did not rebind; instead I made the trust basis explicit where the guard is described and recorded that the guard shipped: spec 05 §"The origin signal is an explicit bridge marker on the connection" already stated "The header is client-asserted and carries no secret, so it is not verifiable … the v1 trust assumption is that peer hubs are cooperative", and its implementation-status sentence was updated from "neither … exists today" to shipped, naming both guard seams (dial + dispatch) with the cooperative-only caveat intact. The High's coverage gap is closed independently of the marker's spoofability.

**Medium 2 — handshake `ProtocolVersion` clobbered by facts.** Fixed: the facts block overwrites only a non-empty `facts.ProtocolVersion` (`app_host_attach.go:109-113`), so an empty preflight version leaves the handshake's negotiated value and attach and probe agree about the same host. Failing-first: `TestHostAttachKeepsHandshakeProtocolVersionWhenFactsAreEmpty` fails with `ProtocolVersion = "", want the handshake's negotiated version 7` when the condition is reverted.

**Medium 3 — terminal `errExecutableMissing` not mapped.** Fixed: exported as `sshconn.ErrExecutableMissing` (`internal/sshconn/preflight.go:326-333`, the same alias pattern as `ErrControllerDirty`) and mapped in `hostAttachWireError` to `appwire.HubLaunchError` alongside `ErrDeploy`/`ErrVersionMismatch`/`ErrControllerDirty` (`app_host_attach.go:180`), since a host with no evener and no deploy path is one the controller cannot install or match its build on. Failing-first: the new `TestHostAttachTerminalFailuresAreTypedWireErrors/executable_missing_with_no_deploy_path` case fails with `attach failure *fmt.wrapError=sshconn: evener executable not found on the host: host "alpha", want a typed WireError` when the mapping is removed.

**Low 1 — redundant attach calls and double-dial race in the spawn picker.** Fixed (`frontend/src/panes/spawn/Spawn.tsx`): the selection path calls `connectHost` only when the manifest does not already report the host online (`displaySources.find(...).online !== true`); an offline row's option is disabled, so the Connect affordance stays the path that reaches it. The in-flight guard is now `connectingHostsRef`, mutated synchronously before the request, so two activations in one batch dial once (the state stays for the disabled state/label). Failing-first: `selecting an already-online remote host issues no attach call` fails with `expected [ { host: 'buildbox' } ] to deeply equal []`, and `a rapid double Connect dials the host once` fails with `expected [ { host: 'offline-host' }, …(1) ] to have a length of 1 but got 2`, when the respective halves are reverted. The round-three test `selecting a remote host issues evener/host/attach for it` is superseded by the online-skip case (an online selection must now send nothing).

**Low 2 — stale spec 05 status text.** Fixed: the clause now names the shipped `evener/host/attach` handler as the component-06 Connect trigger alongside the explicit-SourceIDs attach, replacing "the not-yet-shipped component-06 Connect action".

Verification: `gofmt -l cmd/evener-hub/` empty; `go build ./...` 0; `go vet ./cmd/evener-hub/...` 0; `go test ./cmd/evener-hub/internal/sshconn/ -race -count=1` ok (3.3s); `go test ./cmd/evener-hub/internal/appsource/ -race -count=1` ok (28.7s); `go test ./cmd/evener-hub/ -count=1` ok (190.3s); module-lint PASS (1 module, 84s); `npx biome check --write` on both touched src files → no fixes; `npx tsc --noEmit` 0; `make test-web` PASS (web-typecheck, web-test, web-lint); `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` PASS (web-layoutguard, web-overflowguard, web-shellguard, web-spawnguard, web-transcriptscrollguard, web-retirementguard, web-skillguard); `make lint-generated` PASS.

obra:
Round six addressed (7641fc7d8).

- **M1 — false attach success on generation mismatch / liveness errors** (`cmd/evener-hub/app_host_attach.go`): a `false` handshake now refuses with typed `SessionUnavailable` (same liveness refusal the attached-only lookup and facts seam produce), and a typed `SessionUnavailable` from the facts seam is surfaced as the attach failure instead of being suppressed. Only non-liveness fact-read errors stay suppressed, keeping round 4's dial-authoritative L1 behavior. Pinned by `TestHostAttachRefusesHandshakeGenerationMismatch` and `TestHostAttachRefusesFactsLivenessError`. Failing-first (pre-fix handler):
  - `handshake-mismatch attach = {Attached:true Host:alpha ...}, want an attach failure (not Attached:true)`
  - `liveness-error attach = {Attached:true Host:alpha ServerName:evener-hub ServerVersion:0.1.0 ...}, want an attach failure (not Attached:true)`
- **L1 — stale PreflightIfAttached comment** (`cmd/evener-hub/internal/sshconn/manager_test.go`, `TestPreflightIfAttachedReturnsOnlyALiveChannel`): comment now reads test-only/convenience attached-only accessor. Comment-only.
- **L2 — stale HandshakeIfAttached comment** (same file, `TestHandshakeIfAttachedReturnsOnlyALiveChannel`): same rewording. Comment-only.

Note: the round-5 M1 contest (spoofable-marker cooperative-trust) is still open — this verdict does not address it.

Verification:
- `go build ./...` → 0
- `go vet ./cmd/evener-hub/...` → 0
- `go test ./cmd/evener-hub/internal/sshconn/ -race -count=1 -timeout 20m` → ok (3.866s)
- `go test ./cmd/evener-hub/ -run 'TestHostAttach|TestServeAppWireRPCStamps|TestHostAdminRequest|TestRemoteOriginated' -count=1` → ok (0.120s)
- `make lint-generated` → PASS

obra:
Round seven addressed (a4b98b840).

M1 (EventAttached does not wake the host-notification fan-out) — fixed. The attach event previously poked only the remote-thread refresher (remotePoke) and the navigation invalidation; a fan-out sleeping in exponential backoff (1s doubling to the 30s cap) would wait it out while the new client's 4096-deep notification buffer filled undrained toward a loud overflow teardown. The controller now owns a per-host buffered wakeup (`hostAttached` + `hostNotificationBackoffOrAttach` in `cmd/evener-hub/app_host_admin.go`): `registerHostAdminHandlers` returns the controller, `newHubAppServerWithNavigationAndTrace`/`newWebServer` thread the handle through (`app_rpc.go`, `web.go`), and `main.go` binds a late-bound `hostAttachedWakeup` right after web construction, fired from the same `EventAttached` `onAttach` that already sends `remotePoke` — one event path, not a second one. The wakeup carries no client: the woken fan-out re-checks `Online()` and re-resolves through the attached-only `ClientIfAttached` lookup, so the r6 liveness refusal stands (a host that dropped between event and wakeup resolves `SessionUnavailable` and backs off again). `hubSSHStateInvalidation`'s `onAttach` now takes the event host (`func(host string)`).

Test: `TestHostAdminAttachWakesBackoffSleepingFanOut` (app_host_admin_test.go) — an offline fan-out parks in backoff (0 subscribers), the attach flips it online and signals `hostAttached`, and the source reaches 1 host-notification subscriber (i.e. `SubscribeHostNotifications` registration, which starts the fresh client's drain) without waiting out the backoff; no wall-clock sleeping on the 30s cap. Failing-first: on the pre-fix tree the test does not compile (`controller.hostAttached undefined`), i.e. the wakeup path does not exist, and the equivalent old-loop check never subscribes within the bound. Existing `TestHubSSHStateInvalidationPokesRemoteRefreshOnAttach` updated only for the new `onAttach(host)` signature.

Verification:
- `gofmt -l cmd/evener-hub/` → empty (exit 0)
- `go build ./...` → exit 0
- `go test ./cmd/evener-hub/internal/appsource/ -race -count=1 -timeout 20m` → ok (27.5s)
- `go test ./cmd/evener-hub/ -run 'TestHostAttach|TestHubSSHState|TestHostAdmin|TestRemoteOriginated| ==>Attached' -count=1` → ok (3.5s)
- `make lint-generated` → PASS

obra:
Lint fix (`f94c88d`): removes the `hostNotificationBackoff` helper that round 7 superseded with the `OrAttach` variant (CI lint-golangci flagged it unused; static failed downstream). Comment-only cleanup, no behavior change; the fan-out wakeup disposition stands. Verification: gofmt clean, build + vet exit 0, `TestHostAdmin|TestHubSSHState` PASS.

obra:
Round eight addressed (05e45a1).

M1 — error classification precedence for timed-out host deployments (Medium, fixed): `classifyHostAttachError` in `cmd/evener-hub/app_host_attach.go` now runs the manager's `hostAttachWireError` first, so the deploy-family sentinels (ErrDeploy, ErrVersionMismatch, ErrControllerDirty, ErrExecutableMissing) win over the source's generic deadline/transport mapping. A deployment error wrapped with a context.DeadlineExceeded chain now reaches the browser as the typed HubLaunchError instead of SessionUnavailable; a genuine transport timeout (deadline chain with no deploy sentinel) still maps to SessionUnavailable, and every prior mapping (SSH-start/restart transient, caller cancellation raw, terminal contract refusals) is unchanged. Precedence now orders: manager deploy/contract sentinels → source transport mapping (MapAttachError) → wire passthrough → terminal fallback. Pinned by new `TestHostAttachTimedOutDeployIsHubLaunchError` (`cmd/evener-hub/app_host_attach_test.go`): failing-first confirmed (deploy+deadline subtest failed pre-fix with `session unavailable ... want hubLaunch`, genuine-timeout subtest passed), both subtests pass post-fix.

L1 — unreachable onChange dial branch (Low, fixed): dropped the `connectHost(next)` call from the host-picker `onChange` in `cmd/evener-hub/frontend/src/panes/spawn/Spawn.tsx` (kept `setSource(next)`); the Connect button is now the single attach path for offline hosts. Round 3 picker semantics (Connect affordance fires the attach; offline rows get the affordance) and round 5 online-skip + in-flight guard are preserved. Pinned by new `changing the host select issues no attach call` case in `Spawn.test.tsx`: failing-first confirmed (1 failed pre-fix, select-to-offline dialed), passes post-fix, and the case also asserts the Connect button still attaches.

Verification: `gofmt -l cmd/evener-hub/` empty (exit 0); `go build ./...` exit 0; `go vet ./cmd/evener-hub/...` exit 0; `go test ./cmd/evener-hub/ -run 'TestHostAttach|TestHubSSHState|TestHostAdmin|TestRemoteOriginated' -count=1` ok (3.5s); `npx biome check --write` on touched src files clean (no fixes); `npx tsc --noEmit` exit 0; targeted Spawn subsets green (new L1 test 1 passed; host-picker 5 passed; Connect 4 passed; offline 6 passed; online-skip 1 passed); `make lint-generated` PASS.

obra:
Lint fix (`f5f1981`): `classifyHostAttachError` used a direct type assertion on `hostAttachWireError(err)` that fails on wrapped errors (CI errorlint at app_host_attach.go:168). Replaced with `errors.AsType[appwire.WireError]`, matching the assertion two lines below. Same behavior for unwrapped errors; correct for wrapped deploy chains. Verification: gofmt clean, build + vet exit 0, all attach tests PASS, module-lint PASS (1 module, 25s).

obra:
Round nine addressed (481f58749).

M1 — Connect action timeout too short for remote attach/deploy (Medium, fixed): the Connect action's `evener/host/attach` call used the AppWire client's 30s default (`DEFAULT_REQUEST_TIMEOUT_MS`, appwire-client/typescript/client.ts:24; Spawn.tsx:319 passed no timeout option), but the Ensure seam behind the handler runs sequential bounded phases far beyond that: up to three 10-minute `deployLimit` phases (deploy + restart + launch-contract refresh, cmd/evener-hub/internal/sshconn/manager.go:262), up to three 70s `attemptLimit` phases (preflight, running-hub probe, hub-presence probe; 4x10s connect + 30s init), and the 30s attach handshake — about 34 minutes worst case, with no server-side per-RPC cap anywhere in the path (production `sshconn.Options` in main.go:423 sets no timeout overrides). Verified first per the brief; the "10 minutes" claim is real but per-phase, not total, so the timeout is derived from the true sequential bound. Fix shape: explicit per-call `{ timeoutMs: CONNECT_ATTACH_TIMEOUT_MS }` (35 * 60_000, module-level export in Spawn.tsx) — no long-operation/status RPC exists for attach, and adding one would touch the router catalog (out of scope); this mirrors the existing hubUpdate `APPLY_TIMEOUT_MS`-over-server-bound precedent. Round 3 picker semantics, round 5 online-skip + in-flight guard, and round 8 single-attach-path (Connect button only) are untouched — the only call-site change is the added third argument.

Test: new `a slow Connect attach resolves inside the attach timeout with no failure toast` (Spawn.test.tsx) — pins `call.opts?.timeoutMs === CONNECT_ATTACH_TIMEOUT_MS` and that a gated (slow) attach resolving after the click clears the pending marker with no `Connect offline-host failed` toast. Failing-first confirmed: with the call line reverted to the old two-arg form the test fails (`expected true, received false` on the timeoutMs assertion); with the fix it passes.

Verification:
- `gofmt -l cmd/evener-hub/` → empty (exit 0)
- `go build ./...` → exit 0
- `go test ./cmd/evener-hub/ -run 'TestHostAttach|TestHubSSHState|TestHostAdmin|TestRemoteOriginated' -count=1` → ok (3.5s)
- `npx biome check --write` on touched src files → clean (merged import line into one)
- `npx tsc --noEmit` in cmd/evener-hub/frontend → exit 0
- `npx vitest run src/panes/spawn/Spawn.test.tsx` → 290 passed (incl. the new test)
- `make test-web` → PASS web-typecheck, PASS web-test, PASS web-lint (TESTWEB_EXIT:0)
- `make lint-generated` → PASS
- `GOLANGCI_LINT_CACHE="$(scripts/lib/golangci-lint-cache.sh)" MODULES="." go run ./cmd/evener-dev/bin dev module-lint` → PASS lint (1 modules, 0 findings)

obra:
**Landing statement — #1603 attached-only enforcement + explicit attach trigger (squash).**

**Basis:** roborev verdict for `481f587` = LOWS ONLY (1 Low: ProtocolVersion precedence contradicts the documented invariant at app_host_attach.go:131-132 — recorded below as an immediate follow-up); CI terminal with 0 fails on that head (waiter READY).
**Overlap:** merge-base with main = dc92e9f1e; main changed 29 files since; ZERO file overlap with this PR's 38 files (checked via comm on sorted name lists). Mergeable=MERGEABLE, behind main (strict up-to-date rule).
**--admin scope:** approval + up-to-date rules ONLY (enforce_admins=false; all required checks green on the landed head).

**Residual Low (immediate follow-up):** handshake-authoritative ProtocolVersion — only fill from facts when the handshake supplied no version (`if resp.ProtocolVersion == ""`), with a distinct-versions test matching TestHostCapabilitiesReadsAttachedHandshakeFacts. Reviewer notes the frontend currently ignores the response, so production impact is limited.

**Delivered across r1-r9:** attached-only seams (ClientIfAttached/ChannelIfAttached + generation guards ch.Client()==client); non-dialing main.go wiring; empty-filter gate (explicit SourceIDs may dial); snapshot carry-forward; evener/host/attach handler ({host} field, identity/handshake response fields, typed terminal errors incl. ErrControllerDirty/ErrExecutableMissing/ErrDeploy-precedence, liveness refusal, ProtocolVersion fallback); Spawn picker trigger (Connect affordance, online-skip, in-flight guard, single attach path, 35-min attach timeout); poke-on-attach + EventAttached fan-out wakeup; origin guard at dial + dispatch seams (cooperative-trust marker, contested M1 documented in spec + code); probe handshake refusal + facts-owned HubVersion; spec 03/04/05/06 shipped wording.

PR #1605, opened 2026-09-16
fix(web): host-scope the branch readout and per-host loads, and pin the forwarded method list

obra:
Four residuals from the multi-host campaign's landing reviews, plus one comment that stated the opposite of what the server does. Each finding was verified present at this branch's base (`origin/main` @ `5abcda20d`) before anything changed.

## 1. `panes/spawn/Spawn.tsx`: the branch readout and the wrapped-notification filter keyed on the wrong host

**Finding.** The readout compared `branchHead.host === hostChoice` (line 382) and the wrapper filter compared `n.params.host === hostChoice` (line 820, dep list line 826), while the effect that produces `branchHead` asks `submittedSource` (line 1374) and the catalog the wrapper invalidates is keyed by `submittedSource`. Every other host-scoped path in the file uses `submittedSource`. They diverge in the supported unsettled-manifest window: `submittedSource = source` (the draft's host) while `hostChoice = "local"`.

**Fix.** Both sites (and the listener's dep list) now key on `submittedSource` (lines 387-388, 829-830, 836). `displayHostChoice` — the picker's own last-known-list display value, a deliberately different path — and the round-nine settled-manifest resolution are untouched; no other `hostChoice` use in the file compares against a host that was actually queried. The readout is still display-only in the sense that it routes nothing; it just now describes the host whose `evener/git/head` answer it holds, as the pane's other host-scoped values do.

**Tests** (both written first, both observed red for their own reason):
- `the branch readout keeps the draft's host while the manifest revalidates` — settled manifest, host switched to buildbox, then `seedSources(..., {loading: true, stale: true})`. Pre-fix: the readout vanished (`Unable to find an element by: [data-testid="spawn-branch"]`).
- `the draft's host wrapped config notification reloads the catalog while the manifest revalidates` — same window, with the host's instance and catalog reads held in flight so the settled scope's cached promise is the only thing that can make the pane re-issue `model/list`. Pre-fix: `expected 1 to be greater than 1` — the wrapper never advanced the generation.

## 2. `stores/credentials.ts`: a late answer could overwrite a newer partition for the same host

**Finding.** `fetchHost` guarded a late response only with the shared connection `generation`, which advances on connection transitions rather than per request, so two overlapping reads of one host both captured it and the older answer could commit over the newer listing.

**Fix.** A per-host monotonic sequence (`hostRequestVersions`, credentials.ts:135-145, modelled on the package's own instances store — `appwire-client/typescript/state/credentials/instances.ts`'s `requestVersion`) is taken before each request (lines 156-157) and both commit paths (lines 162, 171) drop a response that is not the latest for its host. Cleared by `resetHostInstancesForTests` (line 213).

**Test** (written first, observed red; then the fix was reverted after the fact to confirm the new test is the one that fails): `an older in-flight remote load cannot overwrite a newer one for the same host`. Pre-fix failure: `- "name": "edited-on-the-host"` / `+ "name": "buildbox-anthropic"` — the superseded answer replaced the newer partition.

## 3. `panes/spawn/spawnDefaults.ts`: a remote submit dropped stored fields it did not re-supply

**Finding.** Round nine's carry-over only protected the model: the blob was still rebuilt from the remote form's own fields, so whenever the stored blob also had a model (which is what kept the write non-empty) every other stored field the remote form left empty — `harness`, `access_mode`, `reasoning_effort` — was dropped by the whole-blob write.

**Fix.** For `remoteLaunch` the blob now starts from the stored blob and only the fields the remote form actually names are overlaid (spawnDefaults.ts:164, replacing the `const blob = {}` + model-only carry-over at origin/main's lines 146 and 173-176); a local submit is unchanged (it is authoritative for its whole blob). Round nine's model carry-over is now the base itself, and round four's no-`removeRaw` rule is unchanged.

**Test** (written first, observed red; then the fix was reverted — with round nine's carry-over reinstated, so only the new case fails): `a remote launch keeps the stored layer's fields the remote form did not name`. Pre-fix failure: `- "access_mode": "full"`, `- "reasoning_effort": "high"` (1 failed, 25 passed).

## 4. Cross-language allow-list pin

**Finding.** The frontend's forwarded method set (`stores/hostRouting.ts`'s `HOST_DEPENDENT_DISCOVERY_METHODS`) and the Go proxy's allow-list (`cmd/evener-hub/app_host_admin.go`'s `remoteHostAdminMethods`) were each pinned only against a literal of their own, so a method deleted from one end left the other end green while the browser kept forwarding a call the proxy answers with `InvalidParams`.

**Fix.** One checked-in list, `cmd/evener-hub/host_request_methods.txt` (commented lines plus one method per line, the shape `scripts/fuzz/fuzz-targets.txt` uses), is now read by both sides:
- `frontend/src/stores/hostRouting.test.ts` asserts the shipped set **IS** that list, both directions (the local literal it used to be spelled against is gone; test at line 30, reader at lines 19-28);
- `cmd/evener-hub/app_host_admin_test.go`'s `TestHostAdminAllowListCoversSharedForwardedMethods` (line 1166) asserts every method on it is on the allow-list **and** that `hubHostAdminController.Request` actually forwards it rather than refusing it, driven through the scripted remote host.

Rationale: a shared list keeps the pin in one readable place and needs no source parsing on either side; the driven half is what makes it more than a map-membership check, because the existing policy table answers only to this package and stays green when a maintainer removes a method and flips its row.

**Revert-and-fail evidence, both directions:**
- deleting `evener/git/head` from the frontend set → `hostRouting discovery inventory > names exactly the shared forwarded-method list` fails (`expected [ 'evener/dirs/create', …(10) ] to deeply equal [ …(11) ]`);
- deleting `MethodEvenerGitHead` from the Go allow-list and flipping its policy row and retry classification (so the package's own tests stay green) → `TestHostAdminAllowListCoversSharedForwardedMethods` fails (`the web UI forwards "evener/git/head" but the proxy's allow-list does not name it`), while the frontend suite still passes — exactly the silence the pin removes.

## 5. `shell/rail/Rail.tsx`: the comment claimed the server does not normalize

**Finding (comment only).** The block above `archiveSessionIdentity` claimed a `"local:<id>"` archive key is one no reader consults, and the `onToggleArchiveSession` comment said the server "stores whatever it is handed (app_archive.go's archiveSet) and no reader expands `local:<id>`". Both are false: `app_archive.go` runs the wire id through `hubcore.NormalizeDecisionSessionID` (`internal/hubcore/archive.go`), which collapses a `local:<id>` ref to the bare thread ID the local read path consults and keeps a host-qualified ref as sent, so both spellings store and read the same key.

**Fix.** Comments only; the code's spelling is unchanged. The comments now state the normalization, so a maintainer cannot "fix" it away on the comment's false premise.

## Gates

| Gate | Result |
| --- | --- |
| `npx biome check --write` (touched files) | clean |
| `npx tsc --noEmit` (`cmd/evener-hub/frontend`) | exit 0 |
| `make test-web` | PASS (`PASS web-typecheck` / `PASS web-test` / `PASS web-lint`, 17.5 min) |
| `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` | PASS, 7/7 guards (`web-layoutguard`, `web-overflowguard`, `web-shellguard`, `web-spawnguard`, `web-transcriptscrollguard`, `web-retirementguard`, `web-skillguard`) |
| `make lint-generated` | PASS |
| `gofmt -l` (touched Go files) | clean |
| `go build ./...` | exit 0 |
| `go test ./cmd/evener-hub/... -count=1` | all packages ok (hub 199.3s, hubcore 19.5s, the rest ≤ 7s) |
| `go test ./cmd/evener-hub/ -run TestHostAdmin -count=1 -v` | all pass, incl. the new pin test |
| `golangci-lint run ./cmd/evener-hub` | clean (the first push tripped `modernize: stringsseq` on the new parser; it now ranges over `strings.SplitSeq`, matching the repo's own list-parsing idiom in `loadawareworkers_test.go`) |

Two notes on how these were obtained, since both cost a re-run:

- The first `make test-web-browser` run failed in `web-skillguard` only (all geometry guards passed): the driver timed out after 30s waiting for the queue-drain continuation leg while `make test-web` was running in parallel on a host already at load ≈ 60 from other sessions. Re-running the guard alone, and then the whole gate serially, passed. The guard's 30s bound is a tripwire over a completion it actually awaits (its own comment says it holds the turn to that leg's own reply rather than using a settling gap), so the contention, not the bound, is the mechanism; nothing here changes the hub/daemon path it drives, and `make test-web-browser` is the last gate to run in this PR (nothing of mine runs beside it).
- CI's first pass flagged one `modernize` finding in the new Go test file; the fix is the last commit amend on this branch (head `862b1a6be`).
## Round 2: the follow-up residuals from the campaign's later review pass

Three of the four new items are **verification items that came back negative** — the described hazard cannot arise on this head — plus one item whose path does not exist. Each verification carries a regression pin whose sensitivity I proved by breaking the guard it covers.

### 1. Stale-model sweep vs. an unsettled manifest — not present
The claim: with the manifest revalidating, `submittedSource` falls back to the local value, so the controller's catalog sweeps (and can delete) a model belonging to the draft's host.

What the code does: `submittedSource = sources.length > 0 ? hostChoice : source` (Spawn.tsx:284), so an unsettled manifest yields the **draft's own host**, not the provisional `"local"` — the same round-nine reading the branch readout and listener now use. The sweep gate is `isLocalHost(submittedSource)` (Spawn.tsx:1283): it runs only when the machine the form is reading — the machine it would launch on — is the controller. Everything the pane requests in that window goes to the draft's host through `evener/host/request`, so there is no controller catalog in play to sweep with.

Pin: `a revalidating manifest does not hand the controller's catalog the sweep` (Spawn.test.tsx). It stores a model the host's list does not offer, mounts with `{loading: true, stale: true}`, and asserts the routed catalog answered (`evener/host/request` → `model/list` for `buildbox`), the controller's own `model/list` was **never** issued, and both stored values are byte-identical afterwards.
Sensitivity: forcing the gate to `true` makes it fail exactly as the finding describes — `{"harness":"evener"}`, i.e. both stored model values deleted.

### 2. Connection-transition release vs. an in-flight host read — not present
The claim: the release races an in-flight host read.

What the code does: a transition bumps `generation` and releases the `loading` flag of every in-flight partition (credentials.ts:110-123). The released read's answer is then refused by both guards (its version is no longer the latest **or** its generation is gone), so it cannot commit, and the read every consumer issues next — the pane's only partition consumer re-runs on the connection change — is the one that lands. The risky ordering, the orphaned answer arriving last, is covered.

Pin: `a released host read cannot outlive the transition that replaced it` (credentials.test.ts): read in flight → `reconnecting` (released, `loading === false`) → ready → fresh read commits → the orphaned answer arrives last and is still refused, leaving the newer listing and no phantom in-flight state.
Sensitivity: removing both commit guards makes the orphaned answer overwrite the newer listing and the test fails.

### 3. Wrapped-notification filter, second site — none
The listener compares `n.params.host === submittedSource` (Spawn.tsx:830), the same value every catalog request is keyed by (`hostRequest(client, submittedSource, "model/list", …)`), so listener and cache cannot disagree by construction. The state the item names — a settled manifest whose draft host is not launchable — writes the fallback into the draft (round nine), making the controller the machine this pane reads; a wrapper from the replaced host is correctly inert. The only other wrapper consumer, `stores/extensions.ts`'s `handleHostNotification`, keys on the notification's own host and is deliberately host-agnostic (it bumps the host-scoped `pluginRevision` for any host's plugin update), so it is not a second site.

Pin: `a wrapper from a draft host the offline fallback replaced does not reload the catalog` (Spawn.test.tsx), which waits for the settled scope's catalog load, emits the replaced host's wrapper, and asserts neither its scope nor the controller's catalog issued a new request inside a window past the 250 ms catalog debounce.
Sensitivity: dropping the host equality from the filter makes it fail (`5` catalog loads instead of `3`).

### 4. `frontend/src/ihm/…` dead code — not actioned: the path does not exist
`cmd/evener-hub/frontend/src/ihm/` exists nowhere in the repository — no file or directory matches `*ihm*`, and no `IHM` token appears in any source, doc, or spec — and the item's text was truncated after "dead code.", so there is nothing to delete. I asked the coordinator for the verbatim path rather than guess; nothing was changed on a guess.

Context from the dead-code sweep I ran while asking (files with no importer, and files referenced only from tests):
- every importer-less file outside `src/dev/**` is an entry point or an ambient declaration (`main.tsx`, `styles/node-fs-shim.d.ts`, `testSetup.ts`); `src/dev/**` is wired through `scripts/editorial-preview.vite.config.mjs` and the other browser-guard entries;
- every test-only module is test support by design (`*TestUtils`, `*testFixture`, `*/testing/*`, `*TestSupport`);
- the one genuine production-dead module is `src/panes/session/chrome/sessionActions.ts` (26 lines): `lastUserMessageText` is referenced only by `sessionActions.test.ts`, its doc comment names `SessionActionsMenu`, which no longer exists, and the fork action now runs through `shell/palette/commands.ts` → `threads.forkFromTurn({ aside: true })`. **Left untouched pending the coordinator's answer** — it is not the path the item named, and deletion on a guess is the one thing the unit's scoping rules forbid.

### 5 & 6. Already in this PR
The `Rail.tsx` `archiveSet` comment (regions 850-873, 1354-1361) and the cross-language allow-list pin (`cmd/evener-hub/host_request_methods.txt`, `hostRouting.test.ts:30`, `app_host_admin_test.go:1166`) both landed with the first commit of this PR; the second commit only verifies them.

## Round 2 gates

| Gate | Result |
| --- | --- |
| `npx biome check --write` (the two touched test files) | clean |
| `npx tsc --noEmit` (`cmd/evener-hub/frontend`) | exit 0 |
| touched test files (`Spawn.test.tsx`, `credentials.test.ts`) | 387 passed |
| `make test-web` | PASS (`PASS web-typecheck` / `PASS web-test` / `PASS web-lint`) |
| `env -u DBUS_SESSION_BUS_ADDRESS make test-web-browser` | PASS, 7/7 guards on the first attempt (`web-layoutguard`, `web-overflowguard`, `web-shellguard`, `web-spawnguard`, `web-transcriptscrollguard`, `web-retirementguard`, `web-skillguard`) |
| `make lint-generated` | PASS |

Sensitivity evidence for every new pin (each probe applied to the production file, test observed red, file restored byte-identical): forcing the sweep gate to `true` deletes both stored model values; removing both commit guards lets the orphaned answer overwrite the newer listing; dropping the wrapper filter’s host equality issues 5 catalog loads instead of 3.


obra:
Landing statement (coordinator).

- Verdict: RoboRev "No issues found" for head `862b1a6b`.
- CI: 16/16 checks green on that head (run 35154925125).
- Merge state: BEHIND by 4. main has changed 15 files since this branch's merge base; none of them are in this PR's file set (verified by intersecting both changed-file lists — zero overlap), so the squash carries no conflict risk and only the branch-protection "up to date" rule needs an override.
- `--admin` is used only to satisfy the required-approval and up-to-date rules; no other protection is bypassed.
- Residual findings to record: none — the verdict is clean.

PR #1606, opened 2026-09-16
fix(hub): recognize the jobs activity tree before translating its refs

obra:
Closes the last code delta implied by spec 05 (`docs/superpowers/specs/2026-09-14-multi-host-05-remote-hub-source.md`, "Nested response refs are translated recursively" + acceptance item 13): the `evener/jobs/list` `Data any` ref translator must **recognize** an activity tree by its required fields and their types before walking it, and preserve everything that fails that test byte-for-byte.

## Equivalence finding: the structural walk was not equivalent

Pre-change, `translateActivityRefs` dispatched on the Go type only (`remote_hub_refs.go:583-591`) and then followed the declared container names `root` → `entries` → `job`/`delegate` → `child`/`turns`, rewriting a ref field whenever its value parsed as a remote session ref. That is equivalent to the required-field test for exactly the payloads the docs note names — `{}`, `{"root":{}}`, an unrelated object, a payload whose declared positions hold no addressable ref — because nothing in those reaches a rewrite.

It is **not** equivalent for any payload that carries a tree-shaped position with an addressable `local:<id>` ref and fails recognition, e.g.:

| payload | required-field test | structural walk (pre-change) |
| --- | --- | --- |
| `{"root":{"sessionId":"S","ref":"local:S"}}` | unrecognized (no `revision`) → preserve | rewrote `root.ref` → `host:S` |
| `{"revision":"3","root":{"sessionId":"S","ref":"local:S"}}` | unrecognized (`revision` is a string) → preserve | rewrote `root.ref` |
| `{"revision":{"value":3},…}` | unrecognized (`revision` is an object) → preserve | rewrote `root.ref` |
| `{"revision":-1,…}` / `{"revision":1.5,…}` | unrecognized (not a non-negative integer) → preserve | rewrote `root.ref` |
| `{"revision":3,"root":{"sessionId":3,"ref":"local:S"}}` | unrecognized (`sessionId` is not a string) → preserve | rewrote `root.ref` |
| `{"root":{"…","entries":[{"kind":"delegate","delegate":{"childRef":"local:S"}}]}}` | unrecognized → preserve | rewrote `root.ref` **and** the nested `childRef` |

So the container names are data outside a recognized tree, and the walk re-pointed refs inside payloads the source does not understand — the forward-compatibility failure the spec's pass-through rule exists to prevent. The structural approach cannot satisfy the spec on its own, so this PR adds the recognition gate (and nothing else): the declared-field walk is preserved verbatim behind it.

## Change

- `remote_hub_refs.go`: `translateActivityRefs` now walks a decoded object only when `activityTreeRecognized` passes — `revision` a JSON number holding a non-negative integer, `root` a JSON object carrying `sessionId` and `ref` as strings (empty allowed: `JobActivityTree.Root/.Revision` and `JobActivitySession.SessionID/.Ref` carry no `omitempty`, so the encoder emits all of them unconditionally). Every payload that fails is returned as the `any` value it arrived as. The retired flat-array branch is unchanged; an array is its own recognized shape.
- Recognition is a gate, not a decode: a newer hub's extra fields do not make a tree unrecognizable, and `revision: 0` or empty required strings still walk.

## Spec update (§"Nested response refs are translated recursively")

The implementation-status paragraph that the docs reconciliation PR added still said the required-field test "is **not implemented**" and that the named pass-through cases "are not pinned by a test" — both false after this change. It now states what ships: `activityTreeRecognized` applies the required-field/type test **before** the declared-field walk, the named pass-through cases are pinned by `TestRemoteHubJobsListPreservesUnrecognizedPayloads`, the recognized tree (every declared ref field, a zero `revision`, empty required strings, a forward-compatible extra field) by `TestRemoteHubJobsListTranslatesRefsOfRecognizedTrees`, and the retired flat array by `TestRemoteHubJobsListTranslatesLegacyFlatArrayRefs` plus its unaddressable value classes in `TestRemoteHubJobsListPreservesUnaddressableLegacyTranscriptRefs` — all four through the actual stream client. The requirement reads as closed rather than an open code item. No neighbouring section was rewritten, and the surrounding requirement text (including acceptance item 13 and the closing "Verified through the actual stream client" summary) already described the recognition rule this change implements, so it stands as written.

## Tests (`remote_hub_refs_test.go`, all through the real stream client via `newScriptedRemote` → `appwire.Client.Request` over a `StreamTransport`, so the `Data any` decode step is exercised)

- `TestRemoteHubJobsListPreservesUnrecognizedPayloads` — 14 payloads, deep-equality on the decoded `any` against an independent decode of the same bytes, plus `assertNoRewrittenRefs`: empty object, `{"root":{}}`, root without `revision`, root missing `ref`, tree-shaped root with nested `childRef`, `revision` as string/object/negative/fractional, `sessionId` not a string, `ref` not a string, `root` not an object, unrelated object, declared containers outside `root`. **7 of these failed before the change** (the ones carrying a tree-shaped `root` with a `local:` ref); all pass after.
- `TestRemoteHubJobsListTranslatesRefsOfRecognizedTrees` — one recognized tree (`revision: 0`, forward-compatible extra fields) asserting every declared ref field translates (`root.ref`, entry `job.ownerRef`/`transcriptRef`, `delegate.childRef`, `turns[].ownerRef`/`transcriptRef`, recursive `child.ref` with its own entries) while opaque `job:` refs, bare ids and undeclared keys survive; plus a subtest pinning that empty required strings still recognize a tree and its nested refs still translate.
- `TestRemoteHubJobsListPreservesUnaddressableLegacyTranscriptRefs` — complements (does not duplicate) `TestRemoteHubJobsListTranslatesLegacyFlatArrayRefs` with the value classes it does not cover: a bare id, a `proj:` ref, a nested-hub ref and an empty value all survive in a legacy flat array's `transcriptRef`.
- Existing pins confirmed unchanged and still passing: `TestRemoteHubJobsListTranslatesActivityRefs`, `TestRemoteHubJobsListLeavesOpaquePayloadsUntouched`, `TestRemoteHubJobsListTranslatesLegacyFlatArrayRefs`.

## Gates

```
gofmt -l cmd/evener-hub/internal/appsource/remote_hub_refs.go cmd/evener-hub/internal/appsource/remote_hub_refs_test.go   # empty
go build ./...                                                                # exit 0
go vet ./cmd/...                                                              # exit 0
go test ./cmd/evener-hub/internal/appsource/ -count=1                         # ok  primeradiant.com/evener/cmd/evener-hub/internal/appsource  8.482s
GOLANGCI_LINT_CACHE=$(scripts/lib/golangci-lint-cache.sh) golangci-lint run --allow-parallel-runners --config .golangci.yml ./cmd/evener-hub/internal/appsource/...   # 0 issues (v2.13.1, .tool-versions pin)
make lint-generated                                                           # PASS lint-generated (3s)
make lint-generated (re-run after the spec edit, on commit 6d53001ae)          # PASS lint-generated (0s); make generate + `git diff --exit-code HEAD -- <8 outputs>` re-verified by hand, both clean
go test ./cmd/evener-hub/ -count=1 -run 'JobsList|RemoteHubSource|RemoteHubProbe'   # ok (cross-check)
```

## Commits

- `197c71f3e` fix(hub): recognize the jobs activity tree before translating its refs
- `6d53001ae` docs(multi-host): record the jobs activity-tree recognition gate in spec 05 (head)


obra:
Round one addressed (`54b7a6867`).

- **medium — malformed activity-tree payloads could still have refs rewritten** (`cmd/evener-hub/internal/appsource/remote_hub_refs.go`): confirmed. The discriminator gate (`activityTreeRecognized`) pins only `revision` plus `root.sessionId`/`root.ref`, so a payload that passed it but could not decode as a complete `appwire.JobActivityTree` was still walked and its `root.ref` rewritten (`local:th_1` -> `host:th_1`). Added a second, cheap gate, `activityTreeDecodable`, in front of the walk: the decoded payload is round-tripped through `json.Marshal`/`json.Unmarshal` into `appwire.JobActivityTree`, and `translateActivityRefs` walks only when that decode succeeds. The decode is a gate, not a translation — the walk still operates on the decoded map, so undeclared/extra keys keep surviving byte-for-byte and the forward-compatible-tree case still translates exactly as before.
- The reviewer's `entries: {}` example held: decoding fails with `json: cannot unmarshal object into Go struct field JobActivityTree.root.entries of type []appwire.JobActivityEntry`, so `{"revision":1,"root":{"sessionId":"S","ref":"local:th_1","entries":{}}}` now passes through untouched. Pinned by `TestRemoteHubJobsListPreservesUndecodableTreePayloads` (raw JSON through the real stream client; deep equality against an independent decode plus `assertNoRewrittenRefs`), with six more shapes: a non-string entry `kind`, non-object `counts` / `branch`, a `revision` of `1e300` that overflows the `uint64` field, a non-object delegate `child`, and a non-string delegate turn `jobId`.
- Spec 05's implementation-status paragraph now describes the two-stage boundary (discriminator + full decode), the gate-not-translation rule, and names the new test.

Failing-first evidence, run against `6d53001ae` (all 7 subtests failed), e.g.:

```
remote_hub_refs_test.go:652: Data = map[string]interface {}{"revision":1, "root":map[string]interface {}{"entries":map[string]interface {}{}, "ref":"host:th_1", "sessionId":"S"}}, want the undecodable payload preserved as map[string]interface {}{"revision":1, "root":map[string]interface {}{"entries":map[string]interface {}{}, "ref":"local:th_1", "sessionId":"S"}}
```

After the fix the new test and every existing jobs-list test pass (`PreservesUnrecognizedPayloads`, `TranslatesRefsOfRecognizedTrees`, `TranslatesLegacyFlatArrayRefs`, `PreservesUnaddressableLegacyTranscriptRefs`); no existing assertion was weakened or removed.

Verification:
- `gofmt -l` on both touched Go files -> empty
- `go build ./...` -> exit 0
- `go vet ./cmd/...` -> exit 0
- `go test ./cmd/evener-hub/internal/appsource/ -count=1 -timeout 20m` -> `ok  	primeradiant.com/evener/cmd/evener-hub/internal/appsource	6.774s`
- `golangci-lint run --allow-parallel-runners ./cmd/evener-hub/internal/appsource/` (pinned 2.13.1) -> `0 issues.`
- `make lint-generated` -> `PASS lint-generated (2s)`

obra:
Landing statement (coordinator).

- Verdict: RoboRev Lows-only for head `54b7a6867`. The single Low is recorded here as an immediate follow-up per the stack's standing rule that residual Lows never block:
  - `remote_hub_refs.go` `activityTreeRecognized`/`activityTreeDecodable` operate on `revision` after the generic JSON decode normalized it to `float64`. At the precision boundary a near-max `uint64` revision (e.g. `18446744073709551615`) rounds to `2^64`, fails the decode gate, and the tree passes through with its `local:` refs unrewritten; conversely a `2^53`-boundary fraction can round to integral and pass both gates. RoboRev marks the practical impact negligible.
  - Follow-up: preserve the raw numeric token (`json.Decoder.UseNumber` / `json.Number`) through the response decode and validate `revision` by exact unsigned-integer parsing, with boundary tests for near-max `uint64` revisions and high-precision fractions.
- CI: all checks terminal on `54b7a6867`, 0 failures.
- Merge state: BEHIND; main has changed 7 files since this branch's merge base, none of them in this PR's file set (zero-overlap check run immediately before merge), so the squash carries no conflict risk.
- `--admin` is used only to satisfy the required-approval and up-to-date rules; no other protection is bypassed.

PR #1607, opened 2026-09-16
fix(providers): a UI-credentialed instance is the user's own, and Codex joins the front page

obra:
## What

Two fixes in the provider onboarding flow.

**OpenAI Codex is on the front page.** It joins the "Popular providers" grid as "OpenAI Codex". The provider is OAuth-only (`api_key_env = []`; its transport reads only the OAuth record), so the card opens the Codex connection and starts the sign-in. There is no key to paste, which is why the card is the only way to reach it without switching to "All providers".

**A provider added through the UI is a normal instance.** Signing in to Codex, or storing a key through the UI, produced an instance listed as "from environment" that could not be renamed or removed, with a read-only Name field.

## Why it happened

The presentation and the hub's gates both keyed on `implicit`, which means only "no `providers.toml` entry shadows this curated provider" (registry spec §5.1). That is equally true of the account a user has just signed in to, so the instance was treated as environment-owned when the user had in fact supplied the credential.

## The rule now

New shared predicate, `fromEnvironment(instance)` in `appwire-client/typescript/credentialLabels.ts`:

    implicit && activeSource is not "store" and not "oauth"

- A `store` or `oauth` credential is a file under the instance name, so those instances belong to the user: no badge, Name editable, Remove offered, and removing one deletes that credential, which is what takes the instance away.
- An API-key variable, the gcloud ADC file, or a keyless local endpoint is still environment-backed: the badge stays and removal is still refused, because unsetting the variable is the only thing that removes it.
- Name is editable on every instance. A rename with no authored entry authors one under the new name; for an environment-backed instance the old row stays, since nothing in a rename can move a shell variable or the ADC file. The note under the field says so instead of "cannot be renamed".

Hub (`cmd/evener-hub/app_instances.go`):

- `Edit`: the authorless-rename refusal is gone.
- `Remove`: the refusal is narrowed by `environmentBacked` (credential source other than `store`/`oauth`).
- `Remove` no longer writes an empty `providers.toml` when the instance had no authored entry: the credential cleanup is the whole removal, and the reload re-derives the instance set either way.

Clients: `InstanceRow`, `InstanceSheet`, and native `ProvidersScreen` use the predicate for the chip and for the Remove affordance.

## Verification

Local gates, all on the head commit:

| Command | Result |
| --- | --- |
| `make test-web` | `PASS web-typecheck`, `PASS web-test` (517 files, 11,934 tests), `PASS web-lint` |
| `make test-api-package` | `qualified @evener/appwire-client@0.1.0` |
| `go test ./cmd/evener-hub/... -count=1` | ok |
| `go test ./llm/registry/... -count=1` | ok |
| `go test ./appwire/... -count=1` | ok |
| `make vet` | exit 0 |
| mobile-native `npm run check` + `npm test` | exit 0 |
| `make lint-naming lint-gofmt lint-internal lint-package-imports lint-generated` | PASS |

Mutation evidence: reinstating the two removed guards makes `TestInstances_RemoveDeletesASignedInCodexAccount`, `TestInstances_EditRenamesASignedInCodexAccount`, and `TestInstances_EditRenamesAnImplicitInstanceUnderANewName` fail. The mutation was reverted. Every other new test was observed red before its implementation.

## Test changes to read

- `TestInstances_EditRenameRefusesAnImplicitInstance` is replaced: it asserted the behavior this PR changes.
- `TestInstances_RemoveValidatesTheInstanceUnderTheControllerLock` is rewritten. It guarded the "validate the instance under the lock" property through an environment refusal, and its second half (a stored key surviving) is no longer expressible, because a stored key outranks the variable and now makes the instance removable. The comment says so.
- `credentialLabels.wire.test.ts` now pins the predicate against the hub's own recorded `evener/instance/list` rows, so a future change to which credential source wins fails there.

## Not covered

- No live hub or browser exercise. The change is covered at the controller and AppWire boundaries; a live pass would need a real Codex sign-in. The browser guards are covered by CI's `web` lane.
- native `ProvidersScreen` has no test rig in this repo; its change is type-checked and mirrors the tested web behavior.

## Review follow-ups on this branch

roborev reviewed every pushed head; each round's Medium findings are fixed here rather than deferred. Every fix carries a regression test, and the commit bodies name the tests and the gates run.

**The removal transaction** (`cmd/evener-hub/app_instances.go`)

- The OAuth record is moved aside by rename and renamed back on any failure, so a removal supports a record the hub cannot read and never leaves a half-written copy behind.
- A removal that stood is announced: the RPC handler broadcasts it and returns `appwire.ErrorInstanceRemovePersisted`, which the web pane, the TUI and native all read.
- The copies a removal sets aside are reclaimed under the removed name only. A copy stranded under any other name may be the last credential of an instance the user still has, so it is left where it is and named in the failure. The trade-off: a leftover under a name that is never removed again is not collected by unrelated removals.
- A removal whose rollback cannot be written still stands, and reports that standing outcome - the entry is out of `providers.toml`, so every client has to drop the row.
- A path at the record's location that is not a regular file refuses the removal before anything is deleted, instead of renaming a directory to an aside name the reclaim skips.
- The aside stamp is seeded one past the highest stamp already filed for that record path, and still steps past a taken one and refuses a candidate it cannot check at all, rather than spinning while holding `credMu`.

**The rename** (`Edit`)

- The OAuth record is read before `providers.toml` is re-keyed, so an unreadable record refuses with nothing changed; a rename that lands and then cannot carry the credentials reports `appwire.ErrorInstanceRenamePersisted`, and the web client reconciles the listing and steers to the new name.
- The TUI carries the listed row's endpoint fingerprint through remove, so the hub's assertion can refuse a name another client has re-pointed.

The verification table above records the gates run at `3ee31c0c1`, the first head. Each later head was re-run through the hub, TUI and appwire suites, `make vet` and `make lint-golangci`; the latest head also ran biome plus the credentials-pane vitest suite (420 tests) and the native typecheck. CI covers the remaining lanes (`web`, `race-*`, `fuzz`, `static-build`). `make test-native`'s Metro bundle step cannot resolve `expo` through this worktree's symlinked `node_modules`, which is environmental.

### Round: the credential lock, and the crash window

- The environment-backed removal refusal was decided by the instance's credential source read under `c.mu` alone. A credential write changes that source while holding `credMu` without `c.mu`, so a writer landing in between could have a removal refuse an instance the user had just credentialed, or wave through one the environment had just taken over. The lookup and the refusal are asked under both locks now.
- A removal moves the OAuth record aside before it deletes anything, so a hub that died inside that window left the instance in `providers.toml` with no record, and only a later removal of the same name ever collected the copy. Startup puts such a record back, and only such a record: an aside whose name the config no longer carries belongs to a removal that stood, so it is left for that name's next removal to collect rather than deleted at startup. The newest copy of a name goes back; a name that has its record again keeps the one in place. It is best effort and never stops startup.
- No reload followed a restore at this point, on the measurement that a registry resolves each instance's credential as it builds its list. That measurement was right about the credential and wrong about the list: the list itself is computed at load, which the next round corrects.

Verification for this round: hub, TUI and appwire suites, `make vet` and `make lint-golangci` on the pushed head, plus a mutation check over the three rules the restore pass enforces (each mutation caught) and one over the startup call site. The tests that pin the lock ordering and the crash recovery were observed failing before their fixes, except the restore rules, which are new code and are pinned by the mutation check instead.

### Round: the removal commit mark, credential-only recovery, and aside ordering

- Startup recovery keyed on `providers.toml` still carrying the name, so the record of a **credential-only implicit instance** - the Codex account a UI sign-in creates, which no `providers.toml` entry names - was never put back at all: a crash inside that removal stranded the user's record under an aside name, with no report and no way back. Recovery now asks two questions instead of one, and either answer restores: the config still carries the name, or the name is a curated implicit provider on the Codex OAuth transport. The curated view comes from the registry's provider catalog, which is built from the embedded catalog rather than from credentials, so it answers while the record is set aside and the instance is therefore absent from the registry's instance list.
- An aside now records whether the removal that wrote it **stood**. `reclaimOAuthAsides` renames each in-flight copy to `.removed-<stamp>` before deleting it, so a hub that dies between the rename and the delete leaves a copy startup can tell apart from one a failed removal set aside, and never puts back. Without a commit mark, recovering credential-only records would have to guess from `providers.toml`, which has nothing to say about a name it never carried - and the guess that recovers a stranded credential is the same guess that resurrects a removal the caller was told had happened.
- Startup now reloads the registry after a restore that put anything back. A registry resolves each instance's credential per call, but its instance list is computed at load, so a credential-only instance whose record came back after that load stayed out of the list - and out of every listing built over it - until the next load. This is the correction to the round above: the earlier measurement was about the credential, and the list is what the reload buys.
- The aside stamp is seeded one past the highest stamp already filed for the same record path rather than from the clock alone. A backward clock step (an NTP correction, a VM snapshot, a container clock) would otherwise file a later removal's copy under a *smaller* stamp than an earlier one's, and startup restores the newest copy - so it would put back the older, possibly revoked record while the newer credential sat as inert debris.

Verification for this round, on `69a0b0715`: `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`, plus a mutation check over the rules this round adds or changes - the restore gate, the newest-copy choice, the record-in-place check, the committed-copy exclusion, credential-only recovery, the commit mark, the monotonic seed, and the reload after a restore. Every mutation was caught and the restored files pass.

### Round: the commit point, and what a failed reclaim reports

- The commit mark sat in the sweep, which runs after the reload, so the whole reload was inside a window where a crash left an in-flight copy for a removal whose durable work - the aside, the credential deletions, the `providers.toml` write - had all landed. Startup would then put the record back, resurrecting an instance the user had removed. The mark now runs at the removal's commit point, immediately before the reload, and the sweep keeps the same rename as its retry. The invariant the recovery rests on is exact now: an in-flight copy means the removal had not done its durable work; a committed one means it had.
- `Remove` hands the copy's current name to its rollback paths, so a reload that fails restores the record from the committed name rather than the in-flight one it no longer carries.
- A failed reclaim now says which of two things went wrong. A copy whose commit rename failed but which was deleted is reported as the crash mark that could not be set; only a copy whose delete failed is described as a credential still on disk. Both still reach the caller as a persisted removal, because the removal stood and every client has to drop the row.

Verification for this round, on `7ff0b10fde59b6ef181eb671a4a23cb1281a9f9b`: `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`, and the mutation check extended with the rules this round adds - the commit point (a mutation that slides the mark back into the sweep must fail the ordering test) and the error classes (a mutation that reports a deleted copy as a leftover must fail) - plus the rollback-name rule. Every mutation was caught and the restored files pass.

### Round: the commit point is not optional, a rename carries its copies, and the sweep

- A commit mark that could not land used to be swallowed, on the theory that the sweep would retry it. If the sweep's delete then failed too, a removal reported as standing was left with an in-flight copy that startup would put back, resurrecting the credential the user had removed. A mark failure now rolls the removal back through the same path a failed reload takes and is reported as a failed removal - never a persisted one - so an in-flight copy can only sit beside a removal the caller was told had failed, which is the case startup recovery exists for. The sweep still performs the rename as a retry, so a mark that fails transiently costs one rejected removal.
- A rename carried only the live record, so an in-flight copy left under the old name - a failed removal whose rename-back also failed - was stranded once `providers.toml` named only the new instance: startup recovery never looked under the old name again. `moveCredentials` now promotes the newest copy to the record path when that path is free, so the renamed instance has a usable credential at once, and carries every other copy to the new name's aside name with its marker and stamp preserved. A carry that fails is reported through the rename's existing `renamePersistedError`.
- Committed copies were only ever collected by a later removal of the same name, so credential material could sit under `auth/<name>.json.removed-<stamp>` forever when that name was never removed again. Startup now deletes a committed copy whose name `providers.toml` does not carry. One whose name the config does carry is left alone: after a removal that failed between its commit mark and its rollback, that copy can be the instance's only record.
- That rule is why `restoreFailedRemoval` now returns a committed copy to its in-flight shape when its rename back to the record path fails: a committed copy has to mean the removal stood, or the sweep cannot tell debris from an instance's only credential. If both renames fail the copy stays committed and the sweep will take it on the next start - a residual named here rather than left implied.
- The in-doubt policy is written down where the commit point is. A crash mid-removal resolves as a rollback, matching every failed removal in `Remove`. The alternative - a marker written before the destructive work, resolving the ambiguity forward - was considered and rejected: such a marker cannot say whether the deletions happened, so it moves the ambiguity rather than closing it, and resolving forward would delete the credential of an instance whose removal the caller was told had failed. The residual window is one rename, the boundary between the last durable step and the commit mark.

Verification for this round, on `5dc5d4f7c6ad363e39bd66cf176a058b82b23b66`: `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`, and the mutation check over every rule these rounds added - including one that tolerates a mark failure again, one that strands copies through a rename, one that makes the sweep delete nothing, and one that leaves a failing rollback's copy committed for the sweep to take. Every mutation was caught and the restored files pass.

### Round: the removal's kind is durable, a lone credential is reported, and the two sides agree on Codex

- Startup recovery decided whether to put an in-flight OAuth copy back by asking the current registry whether the name was a curated credential-only provider. For an authored instance that answer depended on where the removal had got to: the config write drops the entry, so a removal that reached it looked credential-only afterwards, and the next startup put back the record of an account the user had removed. The kind is now recorded where the removal's own rename already writes it atomically - a config-backed removal files `.removing-cfg-<stamp>` / `.removed-cfg-<stamp>`, a credential-only one keeps the old shapes, and a previous release's plain asides parse as credential-only, the conservative reading. Recovery decides on the kind plus the config, which is the durable evidence of what the removal reached: a config-backed copy whose name the config still carries goes back, and one whose name it no longer carries is resolved forward to the committed shape the sweep collects. The registry lookup (`curatedCredentialOnlyOAuth`) is gone.
- A rename promoted the newest in-flight copy to the old record path so the credential move could read it; when that read refused the bytes they sat at a path no reader looks at once `providers.toml` names the new instance, and startup recovery does not recognize a plain record as an aside - the renamed instance lost its credential. A refused read now moves them back under the new name's in-flight aside, kind and stamp preserved, and reports it through the rename's existing problems path.
- A credential-only committed copy whose name the config no longer carries could be a failed rollback's only surviving credential, and the sweep deleted it silently. It is now reported before it goes. The comment states the honest limit: what forces the re-sign-in in that state is the failed rollback that could not restore the record, not the delete, which only removes bytes a human could have copied back by hand.
- `environmentBacked` now exempts the Codex transport before it consults the resolved source, mirroring the client's `fromEnvironment`: the registry reports the oauth source for a readable record while the instance status reports no usable source for one the hub cannot read, so a Codex row could be offered Remove in the UI and refused by the server. The client's other early return - true when the row needs no credential - is already the `keylessScheme` branch, which the comment records. The client's wire test rebinds the hub's own recorded Codex row over every source the corpus sends, so a change to either side of the pair fails there. The corpus's only Codex instance row resolves `oauth`, so the divergent case is exercised by rebinding rather than by a recorded row of its own; the server side is pinned by its own test.
- The mobile screen replaced a refused removal's message with a generic "could not be confirmed" sentence. A removal error carries the instance name and the endpoint fingerprint, never a secret, so `ProvidersScreen` surfaces the hub's own message the way the web pane does; anything that is not a hub error still genericizes.
- The two persisted instance-write discriminators were string literals in the web pane, the sheet and native. The AppWire package now exports them, every client imports that one value, and its errors test reads the constants out of `appwire/errors.go` - so a hub-side rename fails a test rather than silently leaving a standing write read as a failed one.

Verification for this round, on `2fc844c5f`: `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`, `make test-web`, `make lint-package-imports`, and native's `npm run check` and `npm test`. The mutation check now breaks nineteen rules in turn, one per rule these rounds added - including one that records no kind, one that restores a copy the config no longer carries (the finding this round opened with), one that leaves an unparseable promoted record where nothing reads it, one that deletes a lone credential silently, and one that drops the Codex exemption - and confirms the test that names each one catches it.

### Round: a move never replaces an aside, the `default` pointer is config, and a failed rollback reloads

- An aside move went through `os.Rename`, which replaces an existing destination silently. The carry names the new instance's aside from the old copy's stamp, so a copy already filed under the new name at that stamp is a real collision and the destination's bytes were gone. Moves now go through `link(2)` then unlink, which refuses a taken destination: the carry takes a fresh stamp one past the highest the new name holds and refuses to step past the largest tail an aside name can carry, and a commit that finds its committed path taken fails rather than replacing it. (No-clobber is pinned by the test below; the seed's ordering choice is not separately pinned, because stepping to a free name produces the same outcome in the case a test can drive.)
- A rename that could not carry a config-backed copy left it under the old name, where the next startup resolved it forward and the sweep deleted it - the rename had already written providers.toml with the new name, so the only evidence recovery had said the removal stood. A copy the carry cannot land is now re-filed to the credential-only in-flight shape, which recovery restores while its record path is free; a rewrite that fails says so, naming the shape recovery will not restore.
- A removal that only cleared a `default` pointer wrote providers.toml but recorded a credential-only aside, so a crash after that write and before the commit mark left a copy startup put back, bringing the removed instance back. The kind now means what the config write means - an authored entry or a `default` pointer naming the instance - and recovery reads the config the same way: a config-backed copy a `default` pointer still names is an in-doubt removal and goes back.
- When the rollback of a failed removal could not be written, the credentials came back and the function returned without a reload. On a failed mark the file on disk is the one the removal wrote, so the hub kept serving an instance `providers.toml` no longer carries. That path now reloads over the file that stands, and reports a reload that fails too, so a half-repaired hub does not read as healthy.

Verification for this round, on `7c9c06588`: `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`. The mutation check now breaks twenty-five rules in turn - including one that lets a move replace its destination, one that leaves a failed carry's copy in the shape recovery deletes, one that drops the `default` pointer from the kind and one from recovery, one that skips the reload after a failed rollback write, and one that takes an aside name without searching for a free one - and confirms the test that names each one catches it.

### Round: the recovery pass never replaces a committed copy, and a credential-only copy does not need the config

- The forward resolution of a config-backed copy (a removal that reached its `providers.toml` write, returned to the committed shape so the sweep collects it) went through `os.Rename`, which replaces an existing destination silently. That is the same clobber the commit point and the rename carry were cured of in the previous round, at the one site that runs before either of them. The promotion now moves without replacing; a taken destination leaves the in-flight copy in flight for a later pass, holds the taken committed name out of the sweep - deleting it there would lose exactly the bytes the refusal protected - and reports both paths.
- A `providers.toml` that could not be read aborted the whole pass, including the credential-only copies whose recovery never consults the config: their rule is "put it back while the record path is free", and the record file was the whole of what made the instance exist, so an unrelated parse failure stranded the only credential such an instance ever had. The pass now completes that half without the config, defers every config-dependent copy untouched (a config-backed copy cannot be classified without it, and the sweep cannot know what the config names), and reports the config failure so a partial pass never reads as clean.

Still open from this review, and not regressions this PR introduced: the stored-credential window in a removal (a crash between deleting the key and writing the config leaves a configured but unauthenticated instance - a high) and the rename's config-before-credentials window (a crash between them leaves the renamed instance unauthenticated - a medium). `origin/main` has both orderings. The durable marker or journal work they ask for is tracked outside this round.

Verification for this round, on `35272ebea`: `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`. The mutation check now breaks twenty-eight rules in turn - including one that lets the promotion replace a committed copy, one that sweeps the name a refused promotion protected, and one that aborts the pass on an unreadable config - and confirms the test that names each one catches it.

### Round: the rename fallback follows the new name, an absent config is not evidence, and the two client mediums

- A copy a rename could not carry used to be re-filed under the OLD name, so recovery restored it at the old instance's record path - while `providers.toml` named only the new instance. The renamed instance still ended up without its credential, and the name the user had just renamed away could come back. The fallback now lands under the NEW name, as that name's in-flight aside, using the same collision-safe naming the carry itself uses, so recovery restores it to the renamed instance.
- An empty `EVENER_PROVIDERS_CONFIG` (documented as "present and empty - no user layer") reaches the recovery pass as an empty path; `registry.ReadConfigFile` reads that as an absent file and returns an empty layer with no error, so the pass took "there is no config" for "the config carries no name" and resolved every config-backed copy forward for the sweep to delete - destroying exactly the bytes the kind exists to recover. An absent or unreadable config is now no evidence at all: the credential-only half still completes, every config-dependent copy is deferred untouched, and the situation is reported.
- `renameNoReplace` moved with `link(2)`, which a filesystem without hard links (exFAT, some SMB/NFS) refuses, so a removal with an OAuth record could not complete there at all and every rename of such an instance reported an un-carriable copy. It now falls back to `rename(2)` only when the link is refused for want of support, and only because every caller serializes under `credMu` so no other writer can take the destination between the check and the move.
- Client half of the same review: the mobile providers screen closed the editor only for a removal that stood, so an environment-backed row that survived a clean removal stayed open holding the removed instance's draft; it now closes for every successful removal and warns only when a removal stood. The web pane's persisted-removal branch never re-read the listing, leaving the removed row on screen until an unrelated refetch; it now lands the post-removal listing through the pane's own post-mutation refresh, the same reconciliation its rename path uses.

Still open from earlier reviews, unchanged and tracked outside this PR: the stored-key window in a removal (a crash between deleting the key and writing the config leaves a configured but unauthenticated instance - a high) and the rename's config-before-credentials window (a medium). `origin/main` has both orderings; the durable marker or journal work they ask for is its own PR.

Verification for this round, on `7c67db773`: `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`, `make test-web`, `make lint-package-imports`, and native's `npm run check` and `npm test`. The mutation check now breaks thirty-one rules in turn - including one that lands the rename fallback under the old name, one that reads an empty config path as an empty config, one that gives up when the filesystem refuses a hard link, one that lets the forward resolution replace a committed copy, and one that sweeps the name a refused promotion protected - and confirms the test that names each one catches it.

### Round: a committed copy is never restored beside its in-flight twin

- link-then-unlink is not atomic, so a crash between the two calls leaves the same bytes under both an in-flight and a committed name. Recovery classified each file on its own - deleting the committed copy, restoring the in-flight one - which resurrected a credential whose removal had already done its durable work. The pass is now pair-aware: a committed copy is durable evidence that the commit mark landed, so an in-flight copy at the same instance and stamp is treated as committed - never put back, swept by exactly the rules that govern a committed copy. A legitimate state cannot produce that pair, because a new copy's stamp is seeded one past every copy already filed for the name. This also subsumes two guards earlier rounds added against a taken committed destination: that destination exists only as the twin of the in-flight copy being resolved, which is why two of the mutation rules were retired rather than merely re-pointed.
- The `renameNoReplace` fallback refused nothing before renaming onto its destination. It now refuses a taken one itself with `os.ErrExist`, rather than leaning on `link(2)` reporting EEXIST before it reports an unsupported filesystem.
- The unreadable promoted record is re-filed through the same free-name search and no-replace move the carry uses, instead of a deterministic name under a plain rename.
- `restoreUncommittedOAuthAsides` discarded `registry.ReadConfigFile`'s `exists` boolean, so a missing config file - a fresh install, a broken symlink, a config removed while the auth directory kept its asides - still read as "the config carries no name" and the sweep deleted exactly what an interrupted removal left behind. An absent file is now no evidence at all, like an unreadable one (which also makes the empty-path special case redundant; it stays for its clearer message).
- Both stamp searches ignored an unlistable directory, which could file a copy under a stamp it could not order and make recovery restore the older credential. They refuse now instead of guessing.
- A failed rollback write no longer restores the credential of an instance with no authored entry: that credential was the only thing that made the instance exist, so restoring it re-derived the row the caller had just been told was gone. An authored entry still gets its credentials back, because the config write is what removed the entry.

Verification for this round, on `19ad3059e`: `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`. The mutation check runs thirty-three live rules, each caught with the restored files green; four earlier rules are retired in its header with the reason recorded, because the pair rule now subsumes them.

### Round: a name whose copies need the config is deferred whole, and no stamp is guessed at

- With a `providers.toml` that cannot be read, the pass deferred config-backed copies but still restored credential-only ones — including for an instance that had a config-backed copy deferred alongside them. An older credential-only copy could therefore take the record path first, and once the config read again the newer copy was either resolved forward (destroyed) or skipped because the record path was occupied: a stale credential standing in for the current one. The pass now collects the names that have anything config-dependent deferred and defers their credential-only copies too, so an instance is classified whole or not at all. This matters only in the window where the config is unreadable *and* an instance has both kinds of copy — but it is the same "defer, never guess" rule this pass states, applied to the instance rather than to each file.
- `setAsideOAuthFile` could file an aside under a stamp `oauthAsideInstance` cannot parse: a negative `now()` (before 1970) or a `stamp++` that stepped past the maximum. Such a name is invisible to the reclaim, so a removal reported as successful could leave the credential on disk. Both bounds are refused now, mirroring what `freeAsideName` already enforced for the carry.
- A missing `providers.toml` — the default configuration, a fresh install — made every startup print a diagnostic claiming recovery had deferred copies when there were none: the pass reported the config failure before it knew whether any aside existed. It now reads the auth directory first and reports nothing when there is no work the config could have prevented, so a real recovery failure is not lost in constant noise.
- The rename carry walked its copies in `os.ReadDir` order, which is lexical, so `…removing-10` preceded `…removing-9`. Re-stamping bumps a copy above the highest already filed, so with stamps of differing digit length a numerically older copy could be bumped above a newer one and recovery would then restore the older credential. The carry is ordered by numeric stamp (unparseable stamps skipped, as the promotion path already does), so carried stamps preserve the source ordering.

**Still open here, deliberately, and worth stating plainly:** this review's other high asks for a *per-removal transaction marker* covering every aside, so startup classifies a whole removal atomically instead of inferring status from each filename — commit marking renames copies one at a time, so a crash between two of them leaves some marked and some not, and recovery then restores the newest unmarked copy even though the credentials were already deleted. That is the same durable-marker work already scoped to its own PR alongside the stored-key window and the rename window, and it is not fixed here. It is the last known "resurrect a removed credential" path in this subsystem.

Verification for this round, on `c93a989ec`: `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`.

### Round: recovery resolves a whole instance forward, the move is one syscall, and a simplify pass

- **The high.** Recovery could still resurrect an instance whose config-backed removal had already reached its `providers.toml` write: the newer config-backed copy was resolved forward correctly, but an *older* credential-only copy of the same instance fell through into the newest-restorable map and was put back. No crash is needed for that - two copies of different kinds from two different removals are enough. The pass now collects the instances that resolve forward **before** it classifies anything, so the order `os.ReadDir` hands copies back cannot decide the outcome, and every copy of such an instance is treated as committed: swept when the config does not carry the name, left beside it when it does, never restored.
- **The move.** `renameNoReplace` was link-then-unlink, and a crash between the two syscalls stranded the same bytes under both an old-name and a new-name copy - recovery pairs copies by instance and stamp, so it could restore both and bring back the old implicit instance alongside the renamed one. It is now a single `rename(2)` behind a destination check: crash-atomic, with the no-replace guarantee resting on the single-writer argument the earlier rounds already documented. `src == dst` is refused explicitly with `os.ErrExist`, because `rename(2)` on a path onto itself succeeds silently where `link(2)` refused - and the recovery pass reads that refusal as "the committed name was taken". No link is attempted any more, so a filesystem that cannot hard-link stops mattering.
- **The rollback.** The failed-rollback-write branch decided whether to put a credential back by asking whether an authored entry had existed. That is not what decides whether the row returns: an authored entry named after a curated provider (`openai-codex`) leaves the curated provider visible once the config write removes it, so restoring the OAuth record re-derived the instance the caller had just been told was gone. The decision now asks the registry whether the provider is one it derives from the credential itself, and reclaims the credentials for such an instance instead of restoring them.
- **Still out of this PR, by decision:** the rename that persists its config before moving credentials (the durable-journal medium). It is the last known item of that class here, and it goes with the journal work.

**A four-angle simplify pass over this branch landed in the same commit, behaviour-neutral by construction** (the suite is the proof, and no test was added, removed or weakened): one aside-name parse with the old entry points kept and delegating; the rename carry driven by one ranked slice with its destination's highest stamp tracked locally instead of a directory rescan per copy; `setAsideOAuthFile` and `freeAsideName` sharing one search implementation; the recovery pass's two parallel maps collapsed to one; the inverse marker-swap helpers unified; the two RPC error mappers behind one generic helper; `cmp.Compare`; the dead nil-guards dropped; `Remove` calling `configCarriesName`; the TUI's wire-error predicates behind one helper; the two TypeScript persisted-outcome predicates exported from `errors.ts` and used by all three call sites; the clear-flag pairing authored once; and the pane's `renamedInstanceLanded` guard reduced to the single condition it always was.

Verification for this round, on `bd8d9b0a8`: `go test ./cmd/evener-hub/... ./cmd/evener-tui/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`, biome and `make test-web`, `make lint-package-imports`, native's `npm run check` and `npm test`. The mutation check runs forty live rules - including new ones for the single-syscall move, the source-equals-destination refusal, the whole-instance-forward rule and the curated rollback - and every one is caught, with the restored files green.

### Round: the marker swap follows the parser, a rolled-back removal un-marks what it marked, and a removal's refusal is actionable

- **The marker swap followed the wrong occurrence.** The previous round's simplify pass introduced a swap helper that returned on the first marker pair matching anywhere in a copy's name. An instance whose own name carries a marker (`x.removing-cfg-1` is a legal provider name, and the tests already used it) then had its instance-name substring rewritten instead of its trailing marker: the commit mark produced a name like `x.removed-cfg-1.json.removing-5`, which the parser reads as an in-flight copy of a *different* instance, so the reclaim skipped it and startup recovery restored or swept its bytes under the mangled name - the mark that exists to make a pre-delete crash unrecoverable was defeated for that copy. The parser now carries the marker occurrence it matched, and the swap rewrites exactly that occurrence; a name the grammar does not parse is returned unchanged, as before. Both directions are pinned with a marker-bearing instance name.
- **A rolled-back removal un-marks what it marked.** The commit mark moves every in-flight copy of the name, including one an earlier failed removal left behind. When a removal owns no copy of its own and then fails at the mark or at the reload, the rollback restored only the path it was handed - empty - so the pre-existing copy stayed committed, and startup never restores a committed copy while the restored config carries the name: a removal reported as rolled back could leave the instance without its only OAuth credential. The mark now records every path it moved to the committed shape, and a rollback that restores the config returns those copies to the in-flight shape startup recovery reads. The commit point keeps its placement and its meaning; this is only about undoing the mark when the removal does not stand.
- **The refusal reads for its caller.** `Remove` asserts its endpoint fingerprint through the shared verifier, whose refusal told the caller to review its destination and *enter the credential again* - a form a removal does not have. The wording is chosen by the caller now: the credential paths keep their message byte for byte, and a removal's refusal names what a removal can do - review the destination and retry, once the fingerprint key is repairable if it is not.

Verification for this round, on `c7f550e2e399d65462188c7ae9a331c914cf3f18`: `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`. The mutation check runs forty-three live rules - three new, for the parser-carried swap, the un-marking rollback and the removal's wording - and every one is caught, with the restored files green.

### Round: the recovery proof is dated

- **The high.** Recovery resolved a whole instance forward off any config-backed in-flight aside. The proof was treated as a fact about the instance, but it is a fact about the removal: the `providers.toml` write landed as of that copy's stamp. A re-sign-in after it, followed by an interrupted credential-only removal, leaves a copy newer than the proof with no proof of its own - and the instance-wide rule sent that copy to the committed sweep, deleting the credential the user had just signed in with instead of restoring it. The proof is now the highest stamp a config-backed in-flight copy proves, and the classification loop resolves a copy forward only when its own stamp is at or before it; newer copies take the ordinary newest-restorable path. Everything else is unchanged: the pair rule, the protected names, the deferred instance, the sweep, and the left-beside-a-carried-name branch.
- **The tests.** The reviewer's case (an old config-backed proof at 100 with a newer credential-only copy at 200: the newer copy is put back, not swept), the boundary (a copy at the proof's own stamp still resolves forward), and a newer config-backed copy, which is its own proof. The existing whole-instance test passes unchanged, its copy sitting before the proof.

Verification for this round, on `ea4dbe5f609073fba1c3910f50b24c057700b694`: `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`. The mutation check runs forty-five live rules - the proof collection, the stamp bound, and its boundary are each mutated - and every one is caught, with the restored files green.

### Round: the review's three lows

- **The carry leaves an unorderable copy alone.** A copy whose stamp is past an int64 was carried with a synthesized rank, which turned an intentionally unorderable copy into the newest recovery candidate. It is now left filed under the old name, its stamp text intact (still parseable, still unorderable), reported the way the other carry problems are, and it no longer moves the destination ranking.
- **The keyless refusal says what the row does.** `removalRemedy` named an environment variable for a keyless optional-bearer instance and read as if unsetting it removes the row, which it cannot: the instance comes back with its provider. The keyless case is handled first now, and its wording names the variable as the credential the removal takes away while stating plainly that the row returns.
- **The recovery diagnostic is not a false alarm.** Startup recovery reported a providers-config failure whenever an OAuth aside existed, so a credential-only restore on a fresh install logged "could not finish" on success. The diagnostic is added only when the pass actually deferred config-dependent work: a config-backed in-flight copy, a committed copy, or an in-flight twin the config would judge.

Verification for this round, on `c89b761563ecc5132081b7030fc811c7c1a75d09`: `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`. The mutation check runs forty-six live rules - the carry's ranking guard and the diagnostic gate among them - and every one is caught, with the restored files green.

### Round: the TUI's rename outcome, the remove gate, and an exhausted stamp

- **The TUI ignored the rename-persisted outcome.** An edit the hub reported as persisted (the rename landed in `providers.toml`, a credential could not follow) fell through to a save failure, so the panel kept showing the old, now-nonexistent name and never re-read the listing. `handleInstanceMutateResult` now has the branch the removal path already had: the error is cleared, a warning notice carries the hub's message, and the listing is re-read.
- **The TUI offered `x remove` for rows the hub refuses.** An environment-backed instance is not the user's to remove, and the panel now answers the same ownership question the hub does (non-implicit, not the Codex transport, credential-required, source not store or oauth), so it neither advertises nor sends the removal for one.
- **An exhausted stamp refuses.** If a copy is already filed at the maximum stamp, the allocator returned a stamp at or below it, which would let the older credential win at recovery, since the newest copy is the one put back. The maximum is exhausted, and the function returns the refusal it already reports for a stamp it cannot step past.
- **Answered, not changed.** The review's legacy-marker medium wants a pre-kind `.removing-*` copy judged by the config. That rule deletes an implicit instance's only credential (its name is never in `providers.toml`, so "the config does not carry it" is its normal state - six existing recovery tests fail under it), and the artifact class is empty here: the machine running this branch has no aside files at all. The reasoning is on the pull request, and a structural simplification belongs with the journal work.

Verification for this round, on `952ff4054b5220ba6f969634204d6f2ad3722892`: `go test ./cmd/evener-hub/... ./cmd/evener-tui/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`. Each guard was proven red without its fix, and the mutation check runs forty-seven live rules with every one caught, the restored files green.


obra:
**On "Legacy `.removing-*` copies always treated as credential-only":** I tested the suggested rule (judge a plain in-flight copy by the config) and it is the unsafe direction here, so this branch keeps the conservative reading.

What the rule does, measured: as a local mutation it fails six existing recovery tests. The load-bearing one is an implicit instance - a Codex account, a stored key. `configCarriesName` is true only for an authored `[providers.<name>]` entry or a `default` pointer, and an implicit row has neither, so with any `providers.toml` present its plain copy reads as "the config does not carry the name" and is resolved forward and swept. The mutation deleted the credential outright: `deleted the committed credential-only copy openai-codex.json.removing-... whose record path is free`. It also contradicts the stamp rule this branch added last round (`TestRestoreUncommittedOAuthAsidesRestoresACopyNewerThanTheProof`): a plain copy *newer* than a config-backed proof is a later credential-only removal whose interrupted removal must stay restorable.

Why the artifact class is empty rather than rare: the ambiguity exists only for copies written by a build that did not record the kind, and those cannot exist here. The asides ship for the first time in this branch, the only machine that runs it is the maintainer's, and its state root holds no aside files at all - `auth/` contains exactly the two live records and zero `.removing-*` or `.removed-*` entries. Every copy written from here on carries its kind, so the ambiguity cannot be produced going forward either. The conservative reading is also the safe one: a plain copy is restored, never deleted, so a stray pre-kind copy costs at most an instance reappearing after a removal that never finished marking itself - removing it again now marks and reclaims it properly, and no credential is ever lost.

If we ever want the ambiguity gone structurally, the honest version is the journal - one marker written before the first destructive step, one recovery rule ("marker present, put everything back; absent, the removal stood") - which is the follow-up work already scoped for the durable-write windows, not this pull request.

The rest of this review is fixed in the next push: the TUI now handles the rename-persisted outcome (clear the error, warn with the hub's message, re-read the listing), it no longer advertises or sends `x remove` for environment-backed rows, and an aside copy already at the maximum stamp no longer yields a lower one - the maximum is exhausted and the allocator refuses.

obra:
Closing as superseded — not because the work was wrong, but because the branch could no longer carry it cleanly.

The aside machinery this PR exists to land (a removal sets the OAuth record aside instead of capturing its bytes, and recovery reclaims or restores it) merged cleanly when main was brought in, but the branch predated main's newer work in many files it also touched: the residual diff versus main removed the rail clock (`railNow.tsx`/`RailTickProvider`), the keybindings draft store (`KeybindingsDraft`), and other main-side work across 43 files. Landing it as-is would have regressed main, and the branch also carried `ErrorInstanceRemovePersisted`, a duplicate of the `instanceRemoveApplied` discriminator that #1725 landed.

Rather than repair a 24-conflict merge in place, the work is re-landed on current main as a focused PR containing the asides alone: see #2025. That branch carries only the hub-side asides machinery and its tests (the aside-only files, plus the aside calls folded into main's files), reuses main's discriminator, and passes the 44-rule mutation harness over its tree.

The disclosures from this PR's description — the durable per-removal transaction marker (the last known credential-resurrect path), the stored-key window, the rename's config-before-credentials window, and the non-atomic commit marking — carry over to the new PR's description and to the follow-up PR they are scoped to.

Thanks to the reviewers here: the aside rounds' findings (the orderable-stamp carry, the forward-resolution proof, the link/then/unlink move, the rollback classification) are all in the new branch's tests and the harness rules that pin them.

PR #1580, opened 2026-09-16
refactor(native): rows are a projection of the model (D23c-2b)

obra:
Slice c-2b of SDK-migration row D23 (docs/superpowers/plans/2026-09-12-sdk-migration.md), after c-2a (#1547, `d3c537e45`). The display rows become a projection of the package model: after every accepted frame the store publishes `capAndTruncate(projectConversation(applied))`, so a frame and a snapshot produce rows through the one projector and there is nothing for a row applier to disagree with.

## What the store does now

```ts
const applied = applyThreadNotification(state.conversation, n);
if (applied !== state.conversation) {
  set({ conversation: capAndTruncate(withPageHistory(state.conversation, projectConversation(applied))) });
}
// evener/thread/resync stays the authoritative refresh path; an item frame the
// reducer has no case for leaves the model untouched and asks for the canonical read.
```

The Q10 caps are one pure post-projection pass (`capAndTruncate` = `capItems` 500 rows, then `truncateItem` at 64 KiB with the marker). Because the bound is re-applied to whatever text a row carries, the truncation freeze is unnecessary: an oversized stream shows the same bounded prefix however much more arrives, and an authoritative short version is short again.

**Deleted from `mobile/src/state/conversation.ts`** (~1,200 lines): `projectSingleItem`, `decorateLifecycleItem`, `containingTurnStatus`, the cluster splicing (`findActivityTargetBy`/`findActivityTarget`/`findActivityTargetByIdentity`, `activityClusterSegments`, `mergeLiveActivityMembers`, `projectActivityMembers`, `replaceActivityTargetDetail`, `attachToSources`, `itemsAbsentFromSnapshot`), all seven row appliers with `publishModel`, `liveOwnedRevs`/`markLiveOwned`/`liveRevisionForItem`/`isLiveOwned`, `truncatedItemIds` with `truncateAndRecord`/`truncateAndRecordSingle`/`reconcileTruncationFrom`/`exceedsByteLimit`/`exceedsActivityDetailLimit`, the rehydrate live-ownership merge, and `getTruncatedItemIds` (interface + impl).

**Kept for D23d:** `pageOwnedIds` and the page prepend. The older pages live outside the model until D23d moves them in, so `withPageHistory` carries page-owned rows across every publish (and the rehydrate merge keeps the page's own cursor). `pruneEvictedIds` now prunes page ownership only.

[...]

PR #1614, opened 2026-09-16
fix(hub): item frames carry an explicit empty image list on removal

obra:
The hub could not tell a client an item's images were removed. D23c-2b (#1580) taught the TypeScript reducer to treat an explicit `[]` as a value, so a page merge keeps a live removal; nothing could reach that until the hub emitted one.

## What clears images, measured

One path: `holdUnfetchableToolResultImages` (`internal/appprojector/appwire_projection.go:1936`). A tool call settles with sha-addressed result images no server can serve yet, so the projector stashes the item and sends the `item/completed` frame **with the images stripped** (`item.OutputImages = fetchable`, explicitly nil when none are fetchable). The bytes land later and `EventToolResultImagesPersisted` re-sends the same item id with its descriptors restored. A client that already holds a copy with images — a page read, or an earlier frame — is never told to drop them, because the frame carries no key at all.

## Why `omitempty` cannot say it, measured

`encoding/json` omits a zero-length slice under `omitempty` whether it is nil or not. Probed against this toolchain (go1.27.0) rather than assumed:

| tag | nil | `[]` | `["a"]` |
| --- | --- | --- | --- |
| `omitempty` | `{}` | `{}` | `{"images":["a"]}` |
| `omitzero` | `{}` | `{"images":[]}` | `{"images":["a"]}` |
| no option | `{"images":null}` | `{"images":[]}` | `{"images":["a"]}` |

So `omitzero` is the smallest correct encoding, and dropping the option entirely is wrong: it would put `"images": null` on every item that never had images.

## The change

- `appwire/types.go`: `ThreadItem.OutputImages` is `omitzero`. Nil (never had images) sends no key; an empty, non-nil slice (they are gone) sends `[]`. `ThreadItem.Images` stays `omitempty`: nothing removes an item's input images, so it keeps the length rule. The rule is stated on the field.
- `holdUnfetchableToolResultImages` keeps the empty slice it built instead of nilling it — the one removal, now said out loud.
- `projectOutputImages` returns nil, never an empty slice, when nothing survives its filter (every descriptor lacking both URL and SHA): an item whose descriptors were all unusable never showed images, so it is not a removal. Without this the new encoding would announce a removal on an item's first frame.
- `internal/appwirets/emit.go`: the generator treats `omitzero` as optional, like `omitempty`. It differs in what it sends for an empty slice, not in whether the key can be absent.

**Nothing else emits an empty list.** Every other producer of these fields returns nil for empty: `projectUserInputImages`, and `apptranscript.ImagesFromContent` (`var images []appwire.InputItem`, so page reads keep sending no key). `app_subagent_preview.go:49` passes through whatever it was given.

## types.gen.ts does not move, and that is the point

`make generate` produces no diff: `images?: InputItem[]` / `outputImages?: OutputImage[]` are unchanged. Falsified — with the generator fix reverted, `make generate` turns both into **required** fields (`images: InputItem[]`), which would break every consumer of a frame without the key. That is what the generator change prevents, and `TestEmitInterface_OmitzeroIsOptional` pins it.

## Compatibility with a peer built before this change

The `omitzero` change stays on `evener-appwire-v5`: both keys are optional before and after, so a v5 peer never gets a wire error over it. The only observable effect is on a peer running code from before this fix, talking to a hub (or daemon) that has it: that peer's merge logic still treats `outputImages: []` as "the field said nothing" and keeps showing the image it already has, rather than clearing it. It self-corrects at the next hydrate (a reconnect or a fresh thread read), so the symptom is a stale thumbnail hanging around a little longer, never a protocol error or a dropped connection. That is not worth a flag day: it does not strand every running daemon or TestFlight build 4 (see ebce28133, "keep additive plugin launch on AppWire v3").

## Tests (RED first)

- `TestHeldToolResultImagesSendAnExplicitEmptyList` — marshals the withheld-images frame and requires an `outputImages` key holding `[]`. Against `main` it failed with the real frame printed, no key present. It asserts the encoding, not the Go slice, because that is where nil and empty differ.
- `TestAnItemThatNeverHadImagesCarriesNoImageKeys` — the other half: no `images`/`outputImages` key at all on an item that never had any, so the empty list keeps meaning "removed" and frames do not grow two keys each.
- `TestEmitInterface_OmitzeroIsOptional` — the generator contract, red before the one-line change.
- One oracle normalisation: `TestToolResultImageDescriptorWaitsForItsBytes` compares the released item with the settled one *outside their images* and nilled only one side; nil and empty are different under `reflect.DeepEqual`, so both sides are normalised now. The images themselves are still asserted separately, right above. No assertion weakened.

## Gates

- `$(go env GOROOT)/bin/gofmt -l appwire/ internal/appprojector/ internal/appwirets/` — clean
- `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz` over the touched packages — all exit 0
- `golangci-lint run ./appwire/... ./internal/appprojector/... ./internal/appwirets/...` — `0 issues.`
- `make lint` — exit 0
- `go test -count=1 ./appwire/... ./internal/appprojector/... ./internal/appwirets/... ./internal/apptranscript/... ./cmd/evener-tui/...` — exit 0 (the TUI suite included: it consumes the Go structs in-process, so the encoding change cannot reach it, and it is run rather than assumed)
- `go test -count=1 ./cmd/evener-hub/... ./server/... ./internal/appserver/...` — exit 0
- `go test -race -count=1 ./internal/appprojector/` solo — `ok 1.351s`
- root `go test -short -count=1 .` — exit 0
- `make test-api-package` — `qualified @evener/appwire-client@0.1.0` (needs `npm ci` in `appwire-client/typescript` first in a fresh worktree)
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — `PASS`; `gitleaks detect --no-git` on `appwire` and `internal/appprojector` — exit 0


obra:
## Review round 1 disposition (26b42cb5d)

CI: the `tests` job failed on this head (being read). RoboRev: three Highs, one Medium, two Lows. The Highs are the ones that matter and are accepted: the server's reduction layer erases the removal signal before any client sees it (`mergeAppThreadItem` treats an empty slice like absent; `cloneAppThreadItem`'s `append(nil, …)` turns an empty non-nil slice into nil), and no test crosses that layer; round 1 preserves nil-vs-empty through both and adds the end-to-end broadcast test. The Medium (the TypeScript reducer collapses `[]` to `undefined`) is already fixed on D23c-2b's branch (#1580) and is cited rather than duplicated here. Lows accepted: no spurious `"images":[]` from the clone's `make` on nil; `Images` stays on `omitempty` since nothing produces an input-image removal, only `OutputImages` moves. Round 1 in progress.

obra:
## Round 1 fixes (2884161fe)

The three Highs and both Lows. The end-to-end test the review asked for is what proved the other four were real: every layer between the projector and the socket had to be taught that an empty list is a value.

### HIGH 1 — the merge treated empty as absent (`server/appwire_turns.go:884-888`)

`mergeAppThreadItem` copied the snapshot's own images back whenever the incoming list was zero-length, so a removal frame merging onto an item the snapshot had already seen with images came out carrying them again. It now preserves on `nil` only — absence — for both `Images` and `OutputImages`, which is what the length check was really reaching for.

### HIGH 2 + LOW 1 — the clone flattened one direction and invented the other (`:775-776`)

`append([]T(nil), …)` returns nil for an empty non-nil slice, so a removal became an absence; and `make([]InputItem, len(item.Images))` over nil input minted an empty list, so **every** item through the reduction layer was announcing an input-image removal that never happened. Both lists now keep their nil-ness, guarded the way `appitempaging/page.go:211`'s `cloneThreadItem` already does it (`slices.Clone` for the output list, the deep copy loop for the input one). The spurious `"images":[]` shows up in the round's own end-to-end fixture — see the RED output below.

### HIGH 3 — the tests now cross the reduction layer

`TestRecordedItemFrameCarriesTheExplicitEmptyOutputImageList` drives a real `read_file` tool call whose result images are sha-addressed with no URL (so the projector withholds them) through `RecordAppEvent`, and reads the frame the server actually recorded for its clients — through `applyLifecycleAndReturn`, the merge and the clone. It asserts the recorded JSON, because nil and empty are the same Go length and differ only in what is encoded. Its first run:

```
the recorded frame carries no outputImages key: map[callId:"call_shot" id:"item_tool_2"
  images:[] output:"read shot.png" position:{"entry":0,"item":1} status:"completed" …]
```

which is both bugs in one line: the removal gone, and an `images:[]` nobody asked for.

### LOW 2 — only `OutputImages` moves

`Images` is back on `omitempty`: nothing removes an item's input images, they are what the user sent, so there is no removal to express and no reason to put a new encoding on the field. `OutputImages` alone is `omitzero`, and the field comment says why it is the only one.

### MEDIUM (the TypeScript reducer collapsing `[]` to `undefined`) — not re-implemented here

That fix is on D23c-2b's branch (#1580 @eec0afbf7: `imagesToItemImagesForSession` / `outputImagesToItemImages` return `[]` for an explicit empty list, with reducer tests); this checkout predates it. **The client half lands with #1580.** What is pinned here is the wire contract, Go-side: the recorded-frame test above, plus the projector's `TestHeldToolResultImagesSendAnExplicitEmptyList` and `TestAnItemThatNeverHadImagesCarriesNoImageKeys`.

### The schema output did not move

`make generate` produces no diff: both fields were already optional in TypeScript, and scoping the tag back to `omitempty` for `Images` changes nothing there. So no regeneration, and `test-api-package`/`test-web` are not in this round's gate list (they were green on the previous head and nothing they read changed).

## CI

The `tests` job failed on the previous head in **`TestRetirementSafetyColdJobWatchContent`** (`agent`), a job-watch durability test that touches no `appwire` type and no image field — it reads `jobs.jsonl` and the watch store. It passes locally 5/5 at this head. I re-ran the failed job and pushed this head, so the new run is the check that matters; if it reproduces I will treat it as mine and dig rather than assume. Separately, `TestRetirementSafetyEscalation` fails **locally** on macOS for an unrelated reason the failure text names (`"/var" is a symlink to "private/var"`, a sandbox path rule), not on the ubuntu runner.

## Gates

- `$(go env GOROOT)/bin/gofmt -l server/ appwire/ internal/` — clean
- `go vet ./server/... ./appwire/... ./internal/...`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz` — all exit 0
- `golangci-lint run ./server/... ./appwire/...` — `0 issues.`
- `make lint` — exit 0
- `go test -count=1 ./server/... ./appwire/... ./internal/appprojector/... ./internal/appitempaging/... ./internal/apptranscript/... ./cmd/evener-tui/...` — exit 0
- `go test -count=1 ./cmd/evener-hub/...` — exit 0
- `go test -race -count=1 ./server/` — exit 0
- root `go test -short -count=1 .` — exit 0
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — `PASS`; `gitleaks detect --no-git` on `server` — exit 0

obra:
## Review round 2 disposition (2884161fe)

Round 1 accepted; CI green at this head (the earlier `tests` failure was a job-watch durability test unrelated to any appwire type). The merge preserves on nil rather than emptiness for both lists; the clone keeps nil-ness both ways with no `make` over nil input; an end-to-end test drives a real tool call through `RecordAppEvent` and asserts the recorded JSON carries `"outputImages":[]`; only `OutputImages` moves to `omitzero` (no producer emits an input-image removal), the generator output is unchanged; the client half is #1580's. MERGEABLE; merges on CI green + RoboRev clean at this head.

obra:
## Review round 3 disposition (2884161fe)

CI 15/15. RoboRev: three Mediums, one Low, and they are the same shape as round 1 one layer further out (a clone in `appwire/clone.go` and a merge in `apptranscript/logical_turn.go` still collapse the explicit empty list), so round 2 closes the class with a census of every Go site that clones, copies, merges or defaults `OutputImages`, each fixed to be nil-preserving in one commit. `Images` reverts to the length check and stays `omitempty`, since nothing removes input images; the reducer half remains #1580's and is cited. Round 2 in progress.

obra:
## Round 2 fixes (28cd3d9fe)

Closed as a class rather than site by site. Below is every Go site that clones, copies, merges or defaults an image list, with the rule each one follows, from `grep -rn "OutputImages" --include='*.go' | grep -v _test`.

## Census

| Site | What it does | Rule | State |
| --- | --- | --- | --- |
| `appwire/clone.go:67` `cloneThreadItem` | clones a thread item for `CloneThread` (the read path) | nil-preserving | **fixed this round** (`append` to nil yielded nil) |
| `internal/apptranscript/logical_turn.go:306` `mergeAppThreadItems` | merges a later projected item into an earlier one | nil, not length | **fixed this round** (its own doc says it mirrors the server twin, which round 1 fixed) |
| `server/appwire_turns.go:899` `mergeAppThreadItem` | merges an incoming frame into the live snapshot | nil, not length | fixed round 1 |
| `server/appwire_turns.go:780` `cloneAppThreadItem` | clones for the reduction layer | nil-preserving | fixed round 1 |
| `cmd/evener-tui/hub_commands.go:469` | merges an older page into a newer item | nil, not length | **already correct** — it was written `== nil`; the `Images` branch at `:448` likewise |
| `internal/appitempaging/page.go:215` `cloneThreadItem` | clones for the item pager | nil-preserving | **already correct** — the precedent the other two now follow |
| `internal/appprojector/appwire_projection.go:1943` `holdUnfetchableToolResultImages` | withholds unservable descriptors | produces the empty list (the removal) | correct: the `len == 0` there is "nothing to withhold", and round 1 made the withheld set non-nil |
| `internal/appprojector/appwire_projection.go:1656` `projectOutputImages` | projects event images onto the item | nil when nothing survives | correct (round 1): an item whose descriptors were all unusable never showed images, so it is not a removal |
| `cmd/evener-hub/app_threadread.go:1156` `appendOutputImagesUnique` | folds file-backed descriptors into an item | returns `existing` untouched when there is nothing to add | correct: never converts empty→nil or nil→empty |
| `cmd/evener-hub/app_threadread.go:1119` `enrichThreadFileBackedOutputImages`, `output_images.go:135` | assign the result of the above | pass-through | correct |
| `cmd/evener-hub/output_images.go:194`, `app_rpc.go:210` | stamp/strip URLs in place | index-mutating, no re-slicing | correct |

Two sites fixed this round, two already correct, two fixed in round 1, five structurally incapable of losing the distinction.

## `Images` reverted to one rule

The reviewer is right that a nil-preserving merge under `omitempty` is an inconsistent half. `Images` is back to the length check in both the server merge and the server clone (and stays the length check in the apptranscript twin and the TUI's `:448`): nothing removes an item's input images — they are what the user sent — the field is `omitempty`, so an empty list and an absent one are the same thing on the wire. Round 1's assertions about `merged.Images` and `clone.Images` were rewritten to that contract, not deleted: `TestMergeKeepsAnExplicitEmptyOutputImageList` now asserts the snapshot's own input images stand, and `internal/apptranscript`'s `TestMergeFallsBackForAnEmptyInputImageList` pins the same rule on the twin.

## Tests (RED per site)

- `appwire/clone_output_images_test.go` — through `CloneThread`, as asked: an empty list survives, an absent one stays absent, and the clone still copies rather than sharing the caller's array. Red: `cloned OutputImages is nil, so the removal this item announces is lost`.
- `internal/apptranscript/merge_output_images_test.go` — the empty list survives the merge, an absent one still falls back, and input images keep the length rule. Red: `merged OutputImages=[{… shot.png …}], want the later item's explicit empty list`.
- The end-to-end test now continues past the recorded frame into the **read path**: it takes the server's snapshot with its real turns, runs it through `appwire.CloneThread`, and asserts both the Go value and the encoded `"outputImages":[]` — the copy a client actually hydrates from. Falsified by removing the clone's nil-preservation, which turns it red with `the cloned snapshot's item has no outputImages list, so the removal is lost on the read path`.

## The TypeScript reducer Medium

Still carried by #1580 (@eec0afbf7, `imagesToItemImagesForSession` / `outputImagesToItemImages`), cited, not re-implemented. The client half lands with that PR.

## Gates

- `$(go env GOROOT)/bin/gofmt -l appwire/ internal/ server/ cmd/` — clean
- `go vet ./...` — exit 0; `go vet -tags evenerfuzz` and `GOOS=windows go vet -tags evenerfuzz` over `./appwire/... ./internal/... ./server/...` — exit 0
- `golangci-lint run ./appwire/... ./internal/... ./server/...` — `0 issues.`
- `make lint` — exit 0
- `go test -count=1` over every touched package plus their consumers (`./appwire/... ./internal/appprojector/... ./internal/apptranscript/... ./internal/appwirets/... ./internal/appitempaging/... ./server/... ./cmd/evener-tui/... ./cmd/evener-hub/...`) — exit 0
- `go test -race -count=1 ./server/` and `./appwire/` — exit 0
- root `go test -short -count=1 .` — exit 0
- `make generate` — no diff, so `types.gen.ts` and the protocol doc are untouched and the package/web gates are not in this round's list
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — `PASS`; `gitleaks detect --no-git` on `appwire` — exit 0

One note from running the broader suite: `internal/selfupdate` fails **locally on macOS** (`prefix = "/private/var/…", want "/var/…"`) in a worktree that contains none of this branch's changes as well, so it is a pre-existing local-only path-resolution failure, not this PR's. CI was green on the previous head.

obra:
## Simplify disposition (28cd3d9fe)

Two cleanup reviewers read the diff. Applied on this PR (in flight, same push as any round 3 fixes):
1. Measured first: Go's `slices.Clone` already returns nil for nil and a fresh empty slice for an empty input, so the nil guards added around it are dead and go.
2. One definition of the nil-vs-empty invariant: exported `appwire.CloneOutputImages` and `appwire.MergeOutputImages` replace the hand-written guard at the three clone sites (appwire, server, appitempaging) and the merge sites (server, apptranscript, and the TUI merge if it already imports appwire), with the explanatory comment living on the helpers.
3. The new merge test that duplicated `TestMergeAppThreadItemKeepsOutputImages` folds into it; the two near-identical package merge tests collapse onto the shared helper.

Filed rather than widened here: #1637 (generated TS types carry no field doc comments, so the omitzero removal signal is invisible to clients; pre-existing codegen gap). Efficiency angle otherwise clean.

obra:
## Review round 3 disposition (28cd3d9fe)

CI 15/15 green. RoboRev: 1 Medium.

- **Shared TypeScript reducer converts `[]` to `undefined`, so the `??` merges restore stale output images**: accepted into this PR. Rounds 1 and 2 cited it to #1580, which carries the fix, but a wire change whose client half lives in another PR leaves main inconsistent if this one merges first. The exact hunk from #1580 is lifted here (byte-identical, so the later union merge is trivial) with a reducer regression test; #1580 keeps its copy.

Lands in the same push as the simplify fixes above.

obra:
## Simplify fixes (18b4abfd8)

Three commits, plus the client half as a fourth. The head also merges current `main` (`28d94d0df`), which now carries #1629.

### 1. The measurement: my guards were dead code

`TestSlicesCloneCarriesOutputImageListNilness` pins both ends of the construct the whole rule rests on, and it passes against **bare** `slices.Clone`:

- `slices.Clone([]OutputImage(nil))` → `nil`
- `slices.Clone([]OutputImage{})` → non-nil, len 0

So `if item.OutputImages != nil { … slices.Clone … }` protected against nothing. The guard was the right fix for `append([]OutputImage(nil), …)`, which *does* flatten an empty slice to nil; I carried it over to the replacement that does not have that failure. Both guards are gone.

### 2. One definition instead of five copies

`appwire/output_images.go` now holds the rule and nothing else does:

- `CloneOutputImages(in)` — bare `slices.Clone`, with the nil-vs-empty contract in its doc comment and the test above named as its proof.
- `MergeOutputImages(existing, incoming)` — a nil incoming list is "the later item said nothing", so the earlier list stands; **any** non-nil list wins, including an empty one, because that is the removal.

The file's header comment carries what was repeated per site: the three meanings of the field, the two ways copies lost it (`append(nil, …)` and `len(…) == 0`), and why input images deliberately keep their length checks. Every per-site copy of that prose is deleted.

Call sites converged, all five the census named:

| Site | Before | After |
| --- | --- | --- |
| `appwire/clone.go` `cloneThreadItem` | guarded `slices.Clone` | `CloneOutputImages(...)` |
| `server/appwire_turns.go` `cloneAppThreadItem` | guarded `slices.Clone` | `appwire.CloneOutputImages(...)` |
| `server/appwire_turns.go` `mergeAppThreadItem` | `if incoming.OutputImages == nil { … }` | `appwire.MergeOutputImages(...)` |
| `internal/apptranscript/logical_turn.go` `mergeAppThreadItems` | same | `appwire.MergeOutputImages(...)` |
| `internal/appitempaging/page.go` `cloneThreadItem` | guarded `make`+`copy` | `appwire.CloneOutputImages(...)` |
| `cmd/evener-tui/hub_commands.go` `mergeHubItem` | `if merged.OutputImages == nil { … }` | `appwire.MergeOutputImages(...)` — the file already imports `appwire`, so no new import |

`Images` is untouched: one rule, the length check, because the field is `omitempty` and nothing removes input images.

### 3. Tests folded

- `TestMergeKeepsExistingImagesWhenTheFrameCarriesNone` is gone; the pre-existing `server/cov_rhub_server_test.go` `TestMergeAppThreadItemKeepsOutputImages` gained the two cases it was missing (the explicit empty list wins; an empty **input** list still falls back) instead of a second test saying the same thing.
- The two near-identical merge tests collapsed into one table, `appwire.TestMergeOutputImages` (five rows: unsaid, removal, replacement, removal on an item that had none, nothing-said-about-nothing), plus `TestCloneOutputImages`.
- One thin call-site test per package, each falsified by reverting its site: `internal/apptranscript` `TestMergeCallsTheSharedOutputImageRule` (red: `want the later item's explicit empty list`), and a **new** `cmd/evener-tui` `TestMergeHubItemCallsTheSharedOutputImageRule` — that site had no coverage at all, which I found by falsifying it and watching the TUI suite stay green.
- `internal/appitempaging`'s clone site gets no call-site test, deliberately: its old guarded `make`+`copy` and `CloneOutputImages` are behaviourally identical, so no test can tell them apart. That convergence is de-duplication, not a behaviour fix, and I say so rather than writing a test that cannot fail.

## Round 3 fix: the client half moves into this PR

Keeping it in #1580 would leave main in a bad state if this merges first — the hub emitting a removal the shipped client ignores. Lifted from `pr-1580` so the later union merge is byte-identical:

```
git fetch origin +refs/pull/1580/head:pr-1580
git diff origin/main pr-1580 -- appwire-client/typescript/reducer.ts
```

Taken: the two image-conversion hunks, `imagesToItemImagesForSession` and `outputImagesToItemImages`, each `if (!images) return undefined; if (images.length === 0) return [];` with #1580's own comments, **plus** the one-line `__snapshots__/reducer.test.ts.snap` change #1580 carries for them (`"images": undefined` → `"images": []` on `item_steering_live_turn_3001_0`). Verified byte-identical against `pr-1580` with a `diff` of each extracted hunk.

Both image hunks rather than the output one alone, and here is why: they are one change in #1580 — the output function's comment reads "Empty-is-a-value, exactly as imagesToItemImagesForSession above" — and the input half is what moves that snapshot line, which #1580 also owns. Taking half would have meant editing #1580's comment and diverging from byte-identity, which is the property that keeps the union merge clean. **Not** taken: `joinedReasoningParagraphs`, and the same-reference-on-miss changes to `mapTurn` / `settleFirstMatchingTurn`.

Two reducer regression tests: an explicit `outputImages: []` on the newer page clears images an older page still carries, and an absent field keeps them. Falsified by restoring `if (!images || images.length === 0) return undefined`, which turns the first red.

## Gates (merged with `main` @ `28d94d0df`)

- `$(go env GOROOT)/bin/gofmt -l appwire/ server/ internal/ cmd/` — clean
- `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz` over `./appwire/... ./server/... ./internal/... ./cmd/evener-tui/...` — all exit 0
- `golangci-lint run` over the same four — `0 issues.`
- `go test -count=1 ./appwire/... ./server/... ./internal/apptranscript/... ./internal/appitempaging/... ./internal/appprojector/... ./internal/appwirets/... ./cmd/evener-tui/... ./cmd/evener-hub/...` — exit 0
- `go test -race -count=1 ./server/` and `./appwire/` — exit 0
- `make lint` — exit 0; root `go test -short -count=1 .` — exit 0
- `make test-api-package` — exit 0; `make lint-package-imports` — `PASS`
- `cmd/evener-hub/frontend`: `npm run typecheck` — exit 0; `npm test` — exit 0; `npm run lint` — exit 0; `npx biome ci ../../../appwire-client/typescript` — exit 0
- `make test-native` — exit 0
- `make generate` — no diff
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — `PASS`; `gitleaks detect --no-git` on `appwire` and `appwire-client` — exit 0

One gate note: this worktree had no `node_modules` in `cmd/evener-hub/frontend` or `mobile-native`, and the first `npm run lint` there exited **127** (command not found) while `npm run typecheck` exited 2 — neither is a finding. `npm ci` in both, then every web and native gate above. Worth knowing for anyone reading a green tick from a fresh worktree.

obra:
## Review round 4 disposition (18b4abfd8)

CI 15/15 green. RoboRev: 1 Medium (two of three reviewers).

- **Input-image empty list now erases an older page's images, contradicting the Go `Images` rule this PR writes down**: accepted. The hunk lifted from #1580 was wrong in both places; the input path goes back to `undefined` on empty, explicit-empty handling stays only in `outputImagesToItemImages`, the snapshot line that recorded the erasure reverts, with a page-merge test. #1580 drops the same change in its round 7 so the two reducer hunks stay identical.

obra:
## Round 4 fixes (091554322)

Accepted, and my round-3 judgement call was wrong: I took **both** image-conversion hunks from #1580 to keep the union merge byte-identical, and byte-identity was the wrong thing to optimise for when the hunk itself is wrong. The input-image half makes `mergePageItem`'s `images: newer.images ?? older.images` erase an older page's real input images, and it contradicts the Go side of this very PR, which keeps the length rule for `Images` (`server/appwire_turns.go`, `internal/apptranscript/logical_turn.go`, both asserted).

**Empty-is-a-value now applies to `outputImages` only.**

### The failing test first

`an empty input images list never erases the images an older page carries`: a newer page's item carries `images: []`, the older page's carries one real inline image, and the merge must keep it. Red before the revert, with the erasure printed:

```
Error: images=[], want the older page's own input image kept
```

### The revert

- `imagesToItemImagesForSession` is back to `if (!images || images.length === 0) return undefined;`. Its comment now records **why** empty and absent are the same here — `ThreadItem.Images` is `omitempty`, nothing removes an item's input images, the hub's own merges fall back on a zero-length list, and the recorded steering notification in `fixtures/tool-and-jobs.jsonl` sends `images: []`, so answering `[]` erases real images.
- `outputImagesToItemImages` keeps the explicit-empty handling, and its comment no longer points at the input function for the rule (it said "exactly as `imagesToItemImagesForSession` above"; it now states the difference).
- `__snapshots__/reducer.test.ts.snap` is reverted: `item_steering_live_turn_3001_0` is back to `"images": undefined`. That line was the erasure made visible, exactly as the review says. The snapshot vs `origin/main` is now unchanged.

### Falsification

Re-applying the input-image `[]` return turns **two** tests red, not one: my new page-merge test and `fixture tool-and-jobs reduces to the expected model`. The fixture snapshot is an oracle for this rule, which is what made the erasure detectable in the first place.

One extra edit, from the gate rather than the review: biome's `useOptionalChain` flagged the `if (!images || images.length !== 1)` guard in my new test, so the assertion is the suite's own `expect(images).toHaveLength(1)`.

## Gates (merged tree, `main` @ `9bcadfe27`)

- `$(go env GOROOT)/bin/gofmt -l appwire/ server/ internal/ cmd/` — clean
- `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz` over `./appwire/... ./server/... ./internal/... ./cmd/evener-tui/...` — all exit 0
- `golangci-lint run` over the same four — `0 issues.`
- `go test -count=1 ./appwire/... ./server/... ./internal/apptranscript/... ./internal/appitempaging/... ./internal/appprojector/... ./internal/appwirets/... ./cmd/evener-tui/... ./cmd/evener-hub/...` — exit 0
- `go test -race -count=1 ./server/` — exit 0; `./appwire/` — exit 0
- root `go test -short -count=1 .` — exit 0; `make lint` — exit 0; `make generate` — no diff
- `make test-api-package` — exit 0; `make lint-package-imports` — `PASS`
- `cmd/evener-hub/frontend`: `npm run typecheck` — exit 0; `npm test` — exit 0; `npm run lint` — exit 0; `npx biome ci ../../../appwire-client/typescript` — exit 0
- `make test-native` — exit 0
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — `PASS`
- the reducer suite alone: `Tests 188 passed (188)`

## Note for #1580

The same input-image hunk needs dropping there; the coordinator is telling that lane. After this round the two PRs no longer carry an identical `reducer.ts` change, so the union merge will be a real (small) merge rather than a no-op — worth knowing whichever lands second.

obra:
## Review round 5 disposition (091554322)

CI 15/15 green. RoboRev: 1 Medium, 1 Low.

- **`outputImages` semantics changed without a `ProtocolVersion` bump**: measuring first. What a version mismatch does today (refusal, warning, nothing) and how the last bump (`evener-appwire-v4` to `v5`) was carried through fixtures, docs and tests decide it: if a mismatch refuses the handshake and prior semantic wire changes bumped, this PR bumps to `v6` the same way and names the peers it cuts off; if not, the measurement is posted and the decision goes to Jesse rather than a shim being invented here.
- **All-unaddressable descriptor path's nil guard is unpinned**: accepted; the test marshals the settled item and asserts the `outputImages` key is absent, falsified by removing the guard.

obra:
## Round 5 fixes (e089d10c0)

Both accepted. The Medium's two preconditions were measured before I touched anything, and both hold, so the bump is in.

### Measurement (a): a mismatch refuses the handshake

Not a warning, not nothing — both sides refuse:

- **Server**: `internal/appserver/server.go:507` — `initialize` returns `appwire.InvalidRequest("protocol version %q is incompatible; want %q")` when the client's version differs.
- **Client**: `appwire/client.go:462,468` — `Initialize` refuses before sending if the caller's version differs, and returns `ProtocolVersionMismatchError` when the server's answer differs.
- Readers act on that refusal rather than shrugging: `cmd/evener-tui/internal/hubstart/hub_start.go:418`, `cmd/evener-hub/app_sources.go:106`, `internal/sshconn/manager.go:1273`, `hubcore/prober.go:57`, and the hub turns it into a `restartRequired` thread status (`app_restart_required.go:267`).
- The constant's own doc already states the intent: "compared exactly at the handshake … so bumping it makes a mixed pair of binaries fail once, loudly, at initialize — rather than agreeing there and then disagreeing on every request."

### Measurement (b): the precedent is a flag day, not negotiation

- **v3 → v4**, `7d9939d50` "feat(appwire): define v4 atomic transcript item paging", described in its own message as a "Cross-cutting v4 flag day": the constant, `appwire/protocol.go`, `flagday_contract_test.go`, the generated `types.gen.ts`, `docs/appwire-protocol.md`, and the tests carrying the string.
- **v4 → v5**, `33917328a` (#936), a semantic change made "without introducing backward compatibility": the constant, `flagday_contract_test.go`, the TS client and its test, `types.gen.ts`.

No capability negotiation in either. So: refusal + semantic-change precedent = bump, which is what the brief authorised.

### The bump

`evener-appwire-v6`, in that shape: **152 occurrences across 42 files**, every live carrier of the string — `appwire/types.go` (the constant and `TestFlagDayProtocolVersion`), `appwire-client/typescript/client.ts` (`APPWIRE_PROTOCOL_VERSION`) and its test, `scripts/qualify-package.mjs`, the web dev entries (`shellguard-entry.tsx`, `editorial-preview/fixture.ts`), `mobile-native/scripts/demo-hub.mts`, and the Go and TS test fixtures across `cmd/evener-hub`, `internal/sshconn`, `cmd/evener`, `rendezvous`, `internal/appserver`, `mobile-native`.

The version's history comment gains a v6 paragraph in the style of the v3/v4 ones, naming the failure it prevents: a v5 peer's merge restores the images a v6 frame just cleared, because under v5 the field could only ever say "here are the images", so the disagreement is silent in both directions and surfaces as a thumbnail that will not go away.

**Three dated documents deliberately keep `v5`**: `docs/superpowers/reviews/2026-09-10-daemon-idle-retirement.md`, `docs/superpowers/specs/2026-09-14-multi-host-spikes-findings.md`, `docs/web-ui/sidebar-watches/implementation-plan.md`. They record what was true at their date; rewriting them would falsify the record. (A path filter of mine caught them on the first pass — I reverted them and checked.) `docs/appwire-protocol.md` does not carry the string at all today, and `make generate` produces no diff.

### Which peers this cuts off

Every mixed pair, at `initialize`, once and loudly:

- a **v6 hub** and a **v5 daemon** — the daemon shows as `restartRequired` and its session needs an explicit restart (`app_restart_required.go`'s message names the required protocol);
- a **v5 hub** and a **v6 daemon** — the hub refuses the handshake;
- a **v6 hub** and a **v5 web/native client** — `ProtocolVersionMismatchError` at connect, so a stale browser tab or an installed TestFlight build must be reloaded/updated;
- `evener launch-check` disagreement is caught by `spawn.go:869,905` before a session starts.

Daemons outlive the hub that spawned them, so this pair is reachable in ordinary operation — which is the reason the constant exists.

### Low: the projector's nil guard is now pinned

`TestProjectToolCallEndDropsAnUnaddressableOutputImage` marshals the settled item and requires the `outputImages` key to be **absent**, mirroring `TestAnItemThatNeverHadImagesCarriesNoImageKeys`: a zero length cannot tell nil from empty, and an item whose descriptors were all unaddressable never showed an image, so its frame must not announce a removal. Falsified by deleting the `if len(out) == 0 { return nil }` guard, which turns it red: `the frame carries outputImages=[], which announces a removal on an item that never showed an image`.

## Gates (merged tree, `main` @ `a18eb4626`)

- `$(go env GOROOT)/bin/gofmt -l appwire/ server/ internal/ cmd/ rendezvous/` — clean
- `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...` — all exit 0
- `golangci-lint run ./appwire/... ./server/... ./internal/... ./cmd/... ./rendezvous/...` — `0 issues.`
- `go test -count=1 ./appwire/... ./server/... ./internal/... ./cmd/evener-hub/... ./cmd/evener/... ./cmd/evener-tui/... ./rendezvous/...` — one package red, `internal/selfupdate`, the pre-existing macOS `/var` → `/private/var` failure I verified in a worktree carrying none of this branch (`prefix = "/private/var/…", want "/var/…"`). Everything else passes, including all 42 bumped files.
- `go test -race -count=1 ./appwire/` and `./server/` — exit 0
- root `go test -short -count=1 .` — exit 0; `make lint` — exit 0; `make generate` — no diff
- `make test-api-package` — exit 0; `make lint-package-imports` — `PASS`
- `cmd/evener-hub/frontend`: `npm run typecheck`, `npm test`, `npm run lint` — all exit 0; `npx biome ci ../../../appwire-client/typescript` — exit 0
- `make test-native` — exit 0
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — `PASS`

Gate note worth recording: after merging main, the web typecheck exited 2 and `npm test` exited 1 with `TS7006` errors in `skillDocument.test.ts` — a file the merge brought in, whose feature (#1410) added frontend dependencies. Stale `node_modules`, not a finding: `npm ci` in `cmd/evener-hub/frontend`, then all four web gates green. A merge that moves `package.json` invalidates an earlier install.

obra:
## Review round 6 disposition (e089d10c0)

CI 15/15 green, RoboRev clean, simplify pass done (rounds 3 and 4). Merge-ready. Held for Jesse's go on the protocol flag day this head carries (`evener-appwire-v5` to `v6`: a v6 hub refuses TestFlight build 4 and any v5 daemon or tab until rebuilt or restarted). Merge order once released: #1617 first (its moved test carries the v5 string), then this, then #1580 (its reducer hunk is this change's client half).

obra:
## Review round 7 disposition (3e2c27f79)

CI 15/15 green, RoboRev clean, simplify pass done, refreshed onto main after #1617 with no live carrier of the v5 string. Merge-ready at this head; held for Jesse's go on the flag day. #1580 follows it.

obra:
## Review round 8 disposition (9bd0593b6)

CI 15/15 green. RoboRev: 1 Medium, 2 Low, all accepted (all three arrived with the union against #1677).

- **The TUI's `mergeHubItem` erases input images on an explicit empty list**: input-image empty and absent mean "nothing said" everywhere else; the TUI merge uses the length rule, with its own behavior test.
- **Stale comment above `mergeItemImages`**: rewritten to the present rule with no history.
- **Removal pinned only through the page merges**: the `item/completed` settle case is added and falsified against just that merge.

Still held for Jesse's go on the flag day.

obra:
## Round 8 fixes (fbf4cac64)

Refreshed onto `b8307aeee` first (clean merge this time), then all three.

### Medium — the TUI erased input images on an explicit `[]`

`cmd/evener-tui/hub_commands.go` `mergeHubItem` kept a **nil-only** check for `Images` while every other reader of this wire treats an empty input list as "nothing said": the hub's own merges fall back on a zero-length one (`server/appwire_turns.go:884-886`, `internal/apptranscript/logical_turn.go:309`), `ThreadItem.Images` is `omitempty`, and the reducer's `imagesToItemImagesForSession` answers `undefined` for it. So a frame carrying `images: []` erased the older page's real input images in the TUI alone — the same erasure round 4 fixed on the TypeScript side, surviving in the one place nothing asserted it.

Now `len(merged.Images) == 0`, with the difference from output images stated where the two sit together.

Behaviour test first, in the TUI suite (the standing rule that a wire-shape rule needs a `cmd/evener-tui` case): `TestMergeHubItemKeepsOlderInputImagesForAnEmptyList` covers all three inputs — empty, absent, and a real replacement. Red before the fix:

```
merged Images=[] for an empty incoming list, want the older item's own kept
```

### Low 1 — `mergeItemImages`'s comment described the rule this PR replaced

It said both mappers read `[]` as absence and cited the two Go length checks as its authority — one of which (`len(incoming.OutputImages) == 0`) this PR replaced with `appwire.MergeOutputImages`. Rewritten to the present rule, with no history: input images see `undefined` only (the wire has no removal for them, so a settle can only add), output images get an explicit `[]` as **the** removal (`omitzero`, and `MergeOutputImages` on the Go side), and the `??` lets that `[]` win while still keeping the images when the settle said nothing.

### Low 2 — the removal is now pinned on the settle path too

It was only reachable through `mergePageItem`/`mergeOlderItemPage`. `item/completed carrying an explicit empty outputImages list clears the images a prior frame set` drives three settles over one item: images set, then a settle with **no** field (they stay), then a settle with `outputImages: []` (they go). Falsified by changing only `mergeItemImages`'s output line to a truthy-length check, which turns it red:

```
AssertionError: expected [ { src: 'out/plot.png', …(3) } ] to deeply equal []
```

## Checks (the brief's targeted list)

- `$(go env GOROOT)/bin/gofmt -l appwire/ server/ internal/ cmd/ rendezvous/` — clean
- `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz` over the touched modules — all exit 0
- `go test -count=1 ./cmd/evener-tui/... ./appwire/... ./server/... ./internal/appprojector/... ./internal/apptranscript/... ./internal/appitempaging/... ./internal/appwirets/...` — exit 0
- `npx vitest run appwire-client/typescript/reducer.test.ts` from `cmd/evener-hub/frontend` — `196 passed (196)`
- `npm run typecheck` — exit 0; `npx biome ci ../../../appwire-client/typescript` — exit 0
- `make generate` — no diff
- `git grep -n "evener-appwire-v5"` with the five doc exclusions — **empty**

CI decides the rest.

obra:
## Review round 9 disposition (fbf4cac64)

CI 15/15 green, RoboRev clean, simplify pass done. Merge-ready at this head; held for Jesse's go on the flag day. #1580 follows it.

obra:
## Version bump dropped (8f15bfc26)

Per Jesse's ruling 2026-09-17: the `evener-appwire-v6` bump (e089d10c0, 3e2c27f79) is reverted. The `omitzero` change is additive on the wire (`types.gen.ts` unchanged, keys stay optional), so it stays on `evener-appwire-v5` — no flag day, no daemon restart-required, TestFlight build 4 keeps connecting.

**Grep counts**

| | before this commit | after |
| --- | --- | --- |
| `evener-appwire-v6` | 41 files / 151 occurrences | 1 file / 1 occurrence (pre-existing, unrelated: `docs/superpowers/plans/2026-09-10-daemon-idle-retirement.md`, a #1248 planning doc that already said `v6` on `origin/main` before this branch existed — untouched by either bump commit, left as-is) |
| `evener-appwire-v5` | 43 files / 153 occurrences | 43 files / 153 occurrences — identical to `origin/main` file-for-file and occurrence-for-occurrence |

`TestFlagDayProtocolVersion` stays pinned to `evener-appwire-v5`. Every one of the 40 files the bump touched is now byte-identical to `origin/main`. Only `appwire/types.go` (the `omitzero` fix) and `appwire-client/typescript/scripts/qualify-package.mjs` (an unrelated pre-existing diff from other branch work, no version string in it) still differ from `origin/main` — confirmed with `git diff origin/main -- <file>` per file.

Gates run: gofmt (toolchain), `go vet` / `go vet -tags evenerfuzz` / `GOOS=windows go vet -tags evenerfuzz` over `./...`, `go test -count=1` and `-tags evenerfuzz` on every touched Go package (all green), `npx vitest run` on the touched web + package test files, `npm run typecheck`, `npx biome ci ../../../appwire-client/typescript`, `npm run lint`, `make lint-package-imports`, and `npx vitest run` on the touched `mobile-native` test files — all pass.

PR body updated: the flag-day framing is replaced with a "Compatibility with a peer built before this change" paragraph describing the actual behavior (a peer predating this fix reads an explicit `[]` as "nothing said" and shows a stale image until its next hydrate; self-corrects, no protocol error).

obra:
## Simplify disposition (a2b1046f4)

Applied the opus simplify review's 8 apply items as one commit, no behaviour change.

1. Deleted `TestSlicesCloneCarriesOutputImageListNilness` (`appwire/clone_output_images_test.go`) and its now-unused `slices` import. Its two assertions (`slices.Clone(nil) == nil`, `slices.Clone([]T{}) != nil && len == 0`) survive in `TestCloneOutputImages` (`appwire/output_images_test.go`), which tests the same cases against `CloneOutputImages` — a thin wrapper over `slices.Clone`.
2. Deleted `TestCloneThreadCopiesTheOutputImagesItIsGiven`. Its aliasing assertion survives in `appwire/clone_test.go`'s `TestCloneThreadOwnsNestedMutableState`: added `clone.Turns[0].Items[0].OutputImages[0].Name = "changed"` alongside the other mutations, and `original.Turns[0].Items[0].OutputImages[0].Name != "output"` alongside the other post-mutation checks.
3. `reducer.ts:377` dead `if (images.length === 0) return [];` removed (`[].map()` already returns a fresh `[]`).
4. `server/appwire_empty_image_lists_test.go`: dropped the Go-slice assertion (nil-then-len) on the cloned item's `OutputImages`, keeping only the encoding assertion (`fields["outputImages"] == "[]"`) right below it.
5. `appwire/output_images_test.go`: dropped `wantNil`; `TestMergeOutputImages` now checks `(got == nil) != (tc.want == nil)` then `slices.Equal`, and the "nothing said" case just sets `want: nil`. Five cases, same coverage.
6. Trimmed history narration ("before this", "reverting left this suite green", "the conversion collapsed that to undefined", "nothing covered this site before") to the rule that still holds, in: `output_images.go`, `output_images_test.go`, `clone_output_images_test.go`, `cmd/evener-tui/hub_output_images_test.go` (both headers), `server/appwire_empty_image_lists_test.go`, `appwire-client/typescript/reducer.test.ts`, `internal/appprojector/tool_result_image_release_test.go`.
7. Collapsed the nil-vs-empty rule (restated in nine places) to one full statement in `output_images.go`'s package comment, plus one-line pointers everywhere else: `types.go`'s `OutputImages` field comment (2 lines), both sites in `appwire_projection.go`, `appwire_turns.go`, `hub_commands.go`, `reducer.ts`'s three sites, and four test headers (`hub_output_images_test.go`, `internal/apptranscript/merge_output_images_test.go`, `appwire_empty_image_lists_test.go`, `reducer.test.ts`) collapsed to `"output images: see appwire.MergeOutputImages; input images keep the length rule."`.
8. PR body: fixed "The change" — only `ThreadItem.OutputImages` is `omitzero`; `Images` stays `omitempty` (as the field comment already said).

Skipped per the disposition (recorded, not applied): a named `OutputImageList` type; the double zero-exit in `projectOutputImages`.

**Gates**
- `$(go env GOROOT)/bin/gofmt -l appwire server cmd/evener-tui internal/appprojector internal/apptranscript` — clean
- `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz` on those packages — all exit 0
- `go test -count=1` and `-tags evenerfuzz` on `appwire`, `server`, `cmd/evener-tui` (+ subpackages), `internal/appprojector`, `internal/apptranscript` — all green
- `npx vitest run ../../../appwire-client/typescript/reducer.test.ts` (from `cmd/evener-hub/frontend`) — 196 passed
- `npm run typecheck`, `npx biome ci ../../../appwire-client/typescript`, `npm run lint` — clean

Head: `a2b1046f488ab0e9490617b2477e2d70227f5cc4`, `gh pr view 1614` reports `mergeable: MERGEABLE`.

Issue #1595, opened 2026-09-16
native: extension screens unmount into a Reconnect wall on every transport flap, losing scroll, selection and filter text


obra:
`mobile-native/src/PluginsScreen.tsx:42-48` returns a "Reconnect" wall whenever the connection `state !== "ready"`, and `state` mirrors the client's own `onStateChange` (`ConnectionProvider.tsx:146-155`), so a transport flap unmounts the whole Plugins subtree, disposes both extension stores, and the remount builds new ones that fetch from scratch. That is why the reconnect recovery D11 (#1502) added to the shared lifecycle has no native caller: no store survives a non-ready state to recover. The user does not see a stale list; they see the list vanish and come back at the top with selection and filter text gone, on every blip.

Decision for Jesse (native rule: auto-refresh, never stale): should the screens stay mounted through a flap and let the shared lifecycle's reconnect recovery refresh the list in place (then the native wiring of `connectionChanged` becomes the second half of that change), or is the wall the intended behaviour? Found while measuring RoboRev's D11 round-5 finding; the same pattern likely applies to the other native settings screens (measure: grep `state !== "ready"` in mobile-native/src).

obra:
Note from D28's web lift (#1623): `ConnectionStoreState` in the new `state/connection` core exposes only the bare `state` string (no `terminalReason`/`retryNow`), a shape set by the web's needs. If the ruling here is that native screens should survive a flap instead of unmounting, the phone half of D28 may need the core to surface `retryNow`/`terminalReason` reactively; the shape is not load-bearing yet (no new call sites depend on it), so decide before the phone-pointing PR.

obra:
Ruling (Jesse, 2026-09-18): both halves of the D28 phone side land now as two tiny PRs. D1 moves the native app onto the package connection store (no visible change); D2 keeps screens mounted over a connection flap behind a banner instead of the full-screen Reconnect wall, which closes this issue.

PR #1623, opened 2026-09-16
refactor(sdk): the connection client-swap lifecycle lives in @evener/appwire-client/state/connection (D28, web lift)

obra:
## Summary

SDK migration row D28, web lift only (the phone half is a separate PR pending Jesse's ruling on #1595, the native Reconnect-wall decision).

- Moves `stores/connection.ts`'s client-swap safety (`connect`'s reentrancy and stale-frame guards, the outgoing client's `onStateChange` listener detach) and `onConnectionNotification` into a new framework-free subpath `@evener/appwire-client/state/connection` (`core.ts` + `index.ts` barrel), over `AppwireClientLike` only. `createConnectionStore()` returns the `state`/`serverInfo?`/`features?`/`client`/`connect` triple as a `FrameworkFreeStore`; `onConnectionNotification(store, handler)` follows whichever client the store holds across a swap.
- `serverInfo`/`features` stay plain settable fields the host writes after its own handshake read — no timers, no window/document, matching the row's own measurement.
- The web adapter (`stores/connection.ts`) keeps its path and every export (`connectionStore`, `useConnectionStore`, `onConnectionNotification`, `ConnectionStoreState`) by wrapping the core with zustand's `useStore`, so none of its ~36 importers change. `connection.test.ts` stays the oracle, unchanged.
- New package suite `state/connection/core.test.ts` (7 tests): basic wiring/idempotence, the stale client's listener detach on a swap (proved via listener-registration/detachment counters, not just behavioral inference), a re-entrant `connect()` from inside a store subscriber's own dispatch (does not leak the retired client's listener or double-wire the new one — verified by temporarily removing the retire-branch and watching this exact test fail), and `onConnectionNotification` following a client swap.
- Standard subpath plumbing: `package.json` exports, `tsconfig.build.json`, qualification manifest entry + smoke, README subpath prose, tsconfig/vite/vitest aliases in the frontend and mobile-native. No Metro config change needed — `resolveRequest` resolves subpaths generically.
- Plan status table gets a D28 row explaining the web/phone split and citing #1595.

## Test plan
- [x] `make test-api-package` (plus falsification: removed the barrel re-export, confirmed `shipped module unreachable from every published specifier: state/connection/core`, restored via `git diff | git apply -R`)
- [x] `make test-web` (typecheck, vitest, biome)
- [x] `make test-native` (Metro bundle + native/shared vitest + typecheck + script resolution)
- [x] `make test-web-browser` (all six guards green first run)
- [x] `make lint-package-imports`
- [x] `npx biome ci ../../../appwire-client/typescript` and `npm run lint` from `cmd/evener-hub/frontend`
- [x] `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`
- [x] root `go test -short -count=1 .`


obra:
## /simplify disposition (6190bc6a3)

Four angles, two reviewers. Efficiency clean (a byte-for-byte relocation: one listener per `connect`, one publish per call, nothing moved from lazy to eager). Reuse: one item accepted for round 1: `connect` sits inside the state type while the navigation and credentials cores return their client-wiring action beside the store; the core aligns to the sibling shape and the web adapter keeps its exact exports for its ~36 importers. Simplification: the reentrancy and stale-frame guards are both load-bearing (traced, each has its own test); one package test comment still cites web files, fixed in the same round. Altitude: the right depth for a web lift whose phone half waits on #1595; the handshake fields being host data, the three encodings of "current client" across this core, navigation and credentials, and the module's over-promising name are filed as #1627 for the unification after that ruling, with the frozen shape noted on #1595. RoboRev's verdict on this head is pending; merges on CI green + RoboRev clean at the round-1 head.

obra:
## Round 1 fixes (39919fa17)

- **Reuse/simplification finding**: `connect` was a state key on `ConnectionStoreState` while the package's other cores (`state/navigation/store.ts`'s `init`/`awaitConvergence`/`reset`, `state/credentials/instances.ts`'s `connectionChanged`) return their client-wiring action as a sibling of the triple. Fixed: `ConnectionStore extends FrameworkFreeStore<ConnectionStoreState>` now declares `connect` outside state, so a partial `setState` can never clobber it. RED first: added `state/connection/core.test.ts`'s "connect is a sibling of the store, not a state key" test (asserts `"connect" in store.getState()` is `false`), watched it fail against the old shape, then made it pass.
  - Measured consequence while fixing this: every other core here returns `{ ...store, extraMethod }` (a spread, i.e. a new object) from its factory. Doing that here broke an *existing* oracle, `threads.test.ts`'s "connect is idempotent" test, which does `vi.spyOn(connectionStore, "setState")` — a spread hands back a different object than the one `connect()`'s own closure calls `setState` through, so the spy stopped observing internal state changes (`AssertionError: expected "setState" to be called 1 times, but got 0 times`). Root-caused, not worked around: `createConnectionStore()` now mutates the triple in place and returns the same object (comment in `core.ts` explains why, since it deviates from the other cores' convention). The web adapter mirrors this — `connectionStore` IS the core object; only `getState`/`getInitialState`/`subscribe` are overridden in place (memoized per core state object via a `WeakMap`, for zustand's `useStore` with no selector) so `connectionStore.getState().connect(client)` keeps working for its ~36 importers, while `setState`'s identity — and therefore the spy — stays exactly what it always was.
- **Comment genericity**: `core.test.ts`'s reentrancy test comment named `threads.ts`/`credentials.ts`/`connectionStore` (web artefacts) from inside a package test. Reworded to describe the pattern generically (any subscriber of the instance's own `subscribe()`).
- Also fixed the qualification manifest (`esmTypeUses`/smoke) and the plan's D28 status row, which still described `connect` as part of state.

**Merge refresh**: #1623 was CONFLICTING after D14-3′ merged (`4126c8355`). Merged `origin/main` — the only conflict was the plan doc's status table (union: kept D28's row and D14-3/D14-3′'s rows). `git diff origin/main...HEAD --stat` now shows only this row's 14 files.

**Gate results (merged tree, all foreground):**
- `npm test` (frontend vitest): 522 files / 11965 tests passed
- `make test-web-browser`: all six guards passed on the first run

Head SHA: `39919fa1732b17f35c9d1d36aad53329382a9152`

obra:
## Review round 2 disposition (39919fa17)

CI green, MERGEABLE. RoboRev: 2 Medium, 1 Low.

- **Re-entrant swap cycle overwrites the inner disposer (`core.ts:86-113`)**: measuring first with a synchronous A→B→A re-entry test from inside a subscriber; on failure the existing connection generation fence extends to own the wiring slot (no second counter). The pre-D28 web store is checked for the same hole for the record.
- **`attach(null)` leaves the old client's handler wired (`core.ts:128-135`)**: accepted; dispose and reset on a client-less store, test asserts the old client's notification reaches nothing.
- **Adapter leaves `setState` updaters without `connect` (`connection.ts:61-67`)**: checked against round 1, which moved `connect` out of the state type; refuted with the citation if the exported type no longer promises it, otherwise the updater is adapted.

obra:
## Round 2 fixes (b5096e3e3)

**Medium 1** (`state/connection/core.ts:86-113`, finite re-entrant swap cycle). Measured first: added a test that wires a subscriber which, the instant it sees `a` land, synchronously calls `connect(b)` then `connect(a)` again (a finite A→B→A cycle, bounded by a `reentered` flag so it terminates). Against the round-1 code this failed exactly as predicted: `aTracking.detachments` came back `0` (expected `1`) — the outer frame's post-`setState` check `store.getState().client === client` passed (both the outer frame and the inner re-entrant `connect(a)` target the literal same client object `a`), so the outer frame captured its own *stale* listener into `unwireStateChange`, overwriting the inner frame's *correct* one, which then leaked. Fixed by replacing the identity check with a monotonic `connectGeneration` counter bumped per `connect()` call: a frame may only claim `unwireStateChange` if no later `connect()` call has happened since it started. Traced by hand against the test's exact sequence before implementing, to confirm the counter lands on the correct listener (the second, innermost `connect(a)`, not the outer one).

Checked whether the pre-D28 web store had the same hole: **yes** — `git show 55bce119a:cmd/evener-hub/frontend/src/stores/connection.ts` (the commit immediately before D28's move) has the byte-identical identity-only check with no counter. This is a latent bug that predates the migration; D28 carried it forward verbatim, and round 1's own reentrancy test happened not to exercise the "target the same client twice" shape, so it went uncaught until this round's isolated core made it easy to write a test for. The code comment says plainly that this counter is new, not an extension of anything from round 1 — round 1 added no generation fence, only the object-identity check that was already there.

**Medium 2** (`core.ts` `onConnectionNotification`'s `attach`, cited as `:128-135`). Measured: added a test that wires a notification listener, connects a client, clears `store.client` to `null` via `setState`, then emits a notification from the (now-orphaned) client — it still reached the listener. Confirmed this was because `attach(null)` returned immediately on `!client` without running the `unwire?.()` disposal below it. Fixed: the early-return guard now only skips when `client === wired` (true for both a repeat of the same client and a repeat null), and disposal/reset happens unconditionally otherwise; `unwire` becomes `undefined` rather than calling `.onNotification` on a null client. Same pre-existing-hole check: the pre-D28 file has the identical `if (!client || client === wired) return` guard, so this too predates the migration.

**Low** (`cmd/evener-hub/frontend/src/stores/connection.ts:61-67`). Checked round 1's result first, per the instruction: `connect` is a sibling of the core's triple (not a state key) as of round 1, but the *web adapter's exported* `ConnectionStoreState` type still promises `connect` as a field (that's the whole point of round 1's fix — preserving `connectionStore.getState().connect(client)` for the ~36 importers), so this was not moot. `setState` was the one triple member round 1 left unwrapped. Added a test asserting an updater-function callback receives the live `connect` (not `undefined`) — red against the round-1 code. Fixed: `setState` now folds `connect` into the updater's argument via the same `withConnect` memo the other two reads use, and strips `connect` back out of whatever the updater returns before forwarding to the core (which has no such state key). No production call site passes `connect` through `setState` today (checked: all current call sites use object-literal partials, never the updater form, and never name `connect`), so this had no live behavioral consequence yet — but the type's promise is now honored for any future caller that does use the updater form.

**Gate results (merged tree — fetched and merged `origin/main` first, brought in D15 #1630 and an unrelated hub/llm fix, no conflicts):**
- `npm test` (frontend vitest): 523 files / 11970 tests passed
- `make test-api-package`: qualified (plus falsification: removed the barrel re-export, confirmed `shipped module unreachable from every published specifier: state/connection/core`, restored via `git diff | git apply -R`)
- `make test-native`: Metro bundle + native/shared vitest (82+7 files, 792+771 tests) + typecheck + script resolution, all green
- `make test-web-browser`: all six guards passed on the first run
- `make lint-package-imports`: pass
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`: pass
- root `go test -short -count=1 .`: pass

Head SHA: `b5096e3e3d627e3d3e5b414436a33d5df8728128`

obra:
## Review round 3 disposition (b5096e3e3)

CI 15/15 green, RoboRev Low only, simplify pass done (round 1). Merging at this head per the Lows-only rule; the Low (a client cleared or replaced through `setState` rather than `connect` keeps the old connection-state listener) goes into an immediate follow-up that first measures whether the path is reachable past the generation-fenced `connect`, as two of three reviewers doubted.

PR #1629, opened 2026-09-17
fix(hub): thread notifications carry askPending so clients clear the pending ask without a reread (#1613)

### LINKED ISSUE #1613: hub: askPending is stamped only into thread snapshots; no notification carries its clear, so clients stay 'pending' until the next read
`EvenerThread.AskPending` is stamped only into snapshots (cmd/evener-hub/local_daemon.go:1127) and reaches clients through `thread/read`, the thread list and `thread/started`'s snapshot (protocol.go:277); no notification carries a change to it. After a user answers the agent's question, the question item settles on the wire but the thread-level `askPending` stays true on every client until its next read. On the web `useTranscriptScroll` reads `model.askPending` and has this latency today; on native after D23c-2b (#1580 round 4) `questionsPending` (the wire flag OR the live-ask predicate) stays true with no answerable rows until the next read. The reducer must not clear it locally (reducer.test.ts:3305-3324: the field is wire-authoritative). Fix at the hub: emit the thread-level `askPending` change (the existing thread-updated notification, or a field on the ask-settled item frame) whenever it flips. Measured by the D23c-2b implementer.

obra:
Closes #1613.

`EvenerThread.AskPending` reached clients only in snapshots, so after the user answered a question every **other** client kept showing "question waiting" until its next read. The reducer is forbidden from clearing it locally (`reducer.test.ts`, "askPending is wire-authoritative": the ask dock owns its own separate in-tool signal), so the fix has to come from the hub.

## Which carrier, measured

The reducer's thread-level cases are `thread/queueChanged`, `thread/status/changed`, `thread/model/changed`, `thread/reasoning-effort/changed` and `thread/vision-model/changed`. **`thread/status/changed` is already the carrier for exactly this kind of field**: `failedToolCalls` and `capabilities` are otherwise snapshot-only and ride it under an absent-means-no-update rule, stamped at the daemon's single notification egress by `stampFailureCountOnStatusChange` / `stampCapabilitiesOnStatusChange`.

It is also the right moment, not just a convenient one. The pending set clears only at a turn boundary — a resolving user turn or an interrupted turn (`agent/session_tools_ask.go`'s `clearAskPending`, called from `session_lifecycle.go`) — and every such boundary announces a status change. So this refreshes the flag exactly when it can have moved, with no polling and no new notification.

Ordering measured too: `server/bridge.go` samples the envelope facets (`refreshThreadEnvelopeForEvent`; `facetAsk` rides `EventUserInput`, `EventSteeringInjected`, `EventToolCallEnd`, `EventGoalContinuation`) **before** `RecordAppEvent` projects and emits. The value stamped is the one this event produced, not a stale one.

## The change

- `appwire.ThreadStatusChangedParams.AskPending *bool` (`askPending,omitempty`) — a pointer for the same reason `FailedToolCalls` is one: absence means "no update" (an old daemon), never "no question waiting".
- `server/appwire_runtime.go`: `stampAskPendingOnStatusChange`, added to both notification egress chains beside its two siblings. It reads the flag off the thread envelope the snapshot is built from, so a notification and a snapshot cannot disagree. Unlike the failure count there is no unmeasured state to skip — the envelope's flag is a plain bool and `appThread` hands the same value to every snapshot unconditionally — and the comment says so rather than mimicking a guard that would never fire.
- `appwire-client/typescript/reducer.ts`: one line in the `thread/status/changed` case, `askPending: n.params.askPending ?? model.askPending`. The field stays wire-authoritative: this is the wire refreshing it, not the reducer deriving it from items, and the existing "item lifecycle never clobbers the wire's thread-level askPending" test still holds.
- `types.gen.ts` and `docs/appwire-protocol.md` regenerated by `make generate` (one new optional field each).

## Tests (RED first)

- `server/appwire_askpending_status_test.go`: the waiting case, the **clear** the issue is about (the envelope flips to false and the status frame that goes with the resumed turn carries `false`), and the pass-through of every other method and params shape. Falsified by removing the stamp, which turns both carrying cases red with `askPending=<nil>`.
- `reducer.test.ts`: `thread/status/changed` carries `askPending` **and** absence leaves it alone (an older hub cannot blank a flag the hydrate gave us). Red before the reducer line, with `answered.askPending` still `true`.

## Gates

- `$(go env GOROOT)/bin/gofmt -l appwire/ server/` — clean
- `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz` over `./server/... ./appwire/...` — all exit 0
- `golangci-lint run ./server/... ./appwire/...` — `0 issues.`
- `make lint` — exit 0
- `go test -count=1 ./server/... ./appwire/... ./internal/appprojector/... ./cmd/evener-tui/...` — exit 0 (the TUI suite reads this wire shape, so it is run rather than assumed)
- `go test -count=1 ./cmd/evener-hub/...` — exit 0
- `go test -race -count=1 ./server/` — exit 0
- root `go test -short -count=1 .` — exit 0
- `make test-api-package` — exit 0; `make test-web` — exit 0; `make test-native` — exit 0 (all three because the package's `reducer.ts` and `types.gen.ts` moved; a fresh worktree needs `npm ci` in `appwire-client/typescript`, `cmd/evener-hub/frontend` and `mobile-native` first)
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — `PASS`; `gitleaks detect --no-git` on `server` and `appwire` — exit 0


obra:
## Review round 1 disposition (7d763113f)

CI 15/15. RoboRev: one Medium, two Lows, all accepted. The TUI ignores the new `askPending` on the status notification, so its "question waiting" indicator stays stale until a reread: applied with a regression test in the TUI suite (the standing rule for any change to what the projector emits). The stale comment above the reducer's wire-authoritative test is rewritten so the new line is not removed later, and an egress test covers both loops that wire the stamper. Round 1 queued behind the hub lane's two in-flight rounds.

obra:
## Round 1 fixes (8a5c1e6d6)

All three, RED first.

### MEDIUM — the TUI ignored the field (`cmd/evener-tui/hub_notifications.go`)

The session badge reads `detail.AskPending` (`hub_session_view.go:31`, through `displayWord`), which was snapshot-only, so the TUI kept showing "QUESTION WAITING" after the answer until its next read — the same staleness this PR fixes for the web and native clients. The `thread/status/changed` handler now applies the flag inline, right beside the capability set it already applies there for the same reason, under the same absent-means-no-update rule.

Two cases in the TUI suite (the standing rule: a change to what the projector emits gets a `cmd/evener-tui` case in the same PR):
- `TestHubModelStatusFrameClearsTheWaitingQuestion` — asserts the badge really shows the question first (it rendered `● QUESTION WAITING`, which is how I found the header helper renders upper-case), then that a frame carrying `askPending: false` clears it. Red before the change: `detail.AskPending=true after the frame that says the question was answered`.
- `TestHubModelStatusFrameWithoutAskPendingLeavesItAlone` — an older daemon's frame carries no field and the read's own value stands.

### LOW 2 — the stamper was tested, the wiring was not

`TestRecordedStatusFrameCarriesAskPending` drives a real `EventUserInput` through `RecordAppEvent` and reads the frame the server recorded, in the shape `appwire_failure_push_test.go` / `appwire_capabilities_push_test.go` use for the two fields that ride this frame beside `askPending`. It then flips the envelope and drives `EventSessionEnd`, asserting the recorded frame carries an explicit `false` — the clear the issue is about, end to end.

Falsified: dropping the call from the first egress loop (`appwire_runtime.go:565`) turns it red with `the recorded status frame carries no askPending`.

### LOW 1 — the stale comment

`reducer.test.ts`'s comment above "askPending is wire-authoritative from the thread snapshot" said no notification carries the field. It now says the field is wire-authoritative in two ways — the snapshot sets it and `thread/status/changed` refreshes it under absent-means-no-update — and keeps the part that matters unchanged: nothing else may write it, so the reducer must not recompute it from item lifecycle. The AskDock's separate in-tool signal is still named.

## Gates

- `$(go env GOROOT)/bin/gofmt -l cmd/evener-tui/ server/` — clean
- `go vet ./cmd/evener-tui/... ./server/...`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz` — all exit 0
- `golangci-lint run ./cmd/evener-tui/... ./server/...` — `0 issues.`
- `make lint` — exit 0
- `go test -count=1 ./cmd/evener-tui/... ./server/... ./appwire/...` — exit 0
- `go test -race -count=1 ./cmd/evener-tui/` — exit 0
- root `go test -short -count=1 .` — exit 0
- `make test-api-package` — exit 0; `make test-web` — exit 0 (the reducer test's comment is the only package change this round; `types.gen.ts` did not move)
- `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — `PASS`

obra:
## Review round 2 disposition (8a5c1e6d6)

Round 1 accepted: the TUI applies `askPending` beside the capability set with two suite cases (the standing rule), an egress test covers both loops and was falsified by dropping the call from one, and the reducer test's comment now names both authorities without weakening the "nothing else may write it" rule. MERGEABLE; merges on CI green + RoboRev clean at this head.

obra:
## Review round 2 disposition (8a5c1e6d6)

CI 15/15 green, RoboRev Low only, simplify pass done. Merging at this head per the Lows-only rule; the Low (contradictory doc comment on `stampAskPendingOnStatusChange`, which always stamps the field so `false` is a real clear) goes into an immediate follow-up PR.

PR #1609, opened 2026-09-16
refactor(sdk): navigation limits, resource keys and snapshot materialization shared by both apps (D14-3′)

obra:
SDK migration phase D, row D14, slice 3 of 3 (plan: `docs/superpowers/plans/2026-09-12-sdk-migration.md`). Builds on D14-2 (#1584, merged as `5abcda20d`), which moved the navigation store core into `@evener/appwire-client/state/navigation`. D14-3 asked whether native's navigation code (`navigationPages.ts`, `navigationReveal.ts`, `navigationReadback.ts`, `pinNavigation.ts`, `navigationTree.ts`, `projectBrowser.ts`) could adopt `createNavigationStore` the way the web adapter did. Measured, it can't: native has no counterpart to the core at all — no capability gate, no attention tracking, no revalidator, no resource map, no boot fan-out, no expansion (native's Q9, native never persists rail expansion). D14-3 is **withdrawn** for that reason. What native and the core *do* share is ~68 lines out of ~900: a params-to-key mapping and five hand-built snapshot resources. This PR is D14-3′ — sharing those 68 lines without moving a mechanism.

## What moved

- **`navigationParamsToResourceKey`** (`state/navigation/types.ts`) replaces two duplicate encodings of the same mapping: native's `resourceKeyFor` (`navigationPages.ts:55-97`, 43 lines) and the test-only `testing/navigation.ts:paramsToResourceKey` (also duplicated there). Both are deleted; every caller — `NavigationPages`'s constructor and its page reads, `navigationReadback`, and the fixture builder `wireV2` — now calls the one production function.
- **`snapshotResource(key, decoded)` / `materializeSnapshot(key, decoded)`** (`state/navigation/codec.ts`) fold five hand-built `{ key, graph, version, presence: "present" }` literals that existed at `store.ts:402`, `navigationPages.ts:368`, `navigationReveal.ts:40`, `navigationReadback.ts:80` and `pinNavigation.ts:31` — the same "decoded snapshot to normalized resource" step, spelled out five times because there was no shared name for it. `snapshotResource` builds the `NormalizedResource`; `materializeSnapshot` is that plus `materializeNavigationResource` for the three native call sites that only ever wanted the rows, not the graph.
[...]

Issue #1596, opened 2026-09-16
state/navigation: the core's private selectSummaries duplicates selectProjectSummaries, unsorted


obra:
`appwire-client/typescript/state/navigation/store.ts`'s private `selectSummaries()` (used by the boot fan-out) performs the same catalog scan as the package selector `selectProjectSummaries` (D15, `state/navigation/selectors.ts`), but unsorted where the selector sorts by catalog order then offset. Folding the two would change which projects hydrate first under the 4-worker boot pool, so D15 left it alone rather than touch boot ordering in a selectors PR. Follow-up: decide whether boot should hydrate in catalog order (then the core calls the selector) or its own order (then document why and give the private helper a name that says so). Found by the D15 implementer.

PR #1630, opened 2026-09-17
refactor(sdk): graph-shaped navigation selectors live in @evener/appwire-client/state/navigation (D15)

obra:
SDK migration phase D, row D15 (plan: `docs/superpowers/plans/2026-09-12-sdk-migration.md`). Builds on D14-2 (#1584, merged as `5abcda20d`) and D14-3′ (#1609, merged as `4126c8355`), both already on `main`. D15 splits `stores/navigation/selectors.ts`'s 32 exported symbols on the line the row itself named: everything pure over the store's state, depending on nothing but the package, moves to `@evener/appwire-client/state/navigation`; everything that needs a browser-specific type or a cross-render identity cache stays web-side.

## What moved (29 names → `appwire-client/typescript/state/navigation/selectors.ts`)

Every graph-shaped selector, each taking `NavigationStoreState` as an explicit parameter (the `state = navigationStore.getState()` default is gone from the package; the web adapter supplies its own default at the two call sites that need one — see below):

`relativeAge`, `selectAttentionSummary`, `selectSources`, `selectDisplaySources`, `selectResource`, `selectProjectResource`, `selectProjectPage`, `selectLocation`, `selectSectionRows`, `selectNeedsYouRows`, `selectNeedsYouCount`, `selectSectionRemaining`, `selectNextSectionOffset`, `selectCatalogRemaining`, `selectNextCatalogOffset`, `selectLiveRows`, `selectGlobalRows`, `selectPinSectionSummaries`, `selectPinSections`, `selectProjectSummaries`, `selectExpanded`, `selectSessionSummary`, `findSessionNode` (an alias for `selectSessionSummary`), `selectSessionOmittedWatches`, `selectSessionOmittedArmedWatches`, plus the `NavigationPinSectionSummary`/`LoadedPinSection` interfaces they return and the `NavigationState` alias (`= NavigationStoreState`) they're typed by.

The split is contiguous in the original file: everything above the `shell/rail/railNodes.ts` import at `:353` is graph-shaped: nothing here imports React, zustand, or a web-only type.

## What stayed (3 names, web-side)

**2 rail-shaped, kept because they're typed by web-only shapes (Q4):** `NormalizedRailModel` and `selectRailModel` are typed by `IsExpanded`/`RailSession`/`SessionRailNode` (`shell/rail/railNodes.ts`), which never enter the package. The worksheet's Q4 asked whether to leave them in a web adapter or extract package-neutral result types (~15 call-site edits now); this PR takes the worksheet's own recommendation (the adapter) since `railNodes.ts` stays put regardless and none of the 15 web importers change either way. **This is a judgment call, not a measurement — flagging for a veto if you'd rather extract the neutral types now.**

**1 host-memoized, kept because its cache must outlive a state publication:** `selectSessionWatches` (plus its two test-only cache hooks, `sessionWatchesCacheSizeForTests`/`resetSessionWatchesCacheForTests`) holds a `Map<ref, { key, watches }>` keyed by session ref so a narrow subscription doesn't re-render on unrelated navigation churn. That cache is instance state tied to one host's render loop; a package module singleton would leak across every `NavigationStore` instance (each store's tests already create fresh instances per D14-2/D14-3′), so it can't move without becoming per-store state the package doesn't otherwise carry.

## The adapter (`cmd/evener-hub/frontend/src/stores/navigation/selectors.ts`, 545 → 305 lines)

Re-exports all 29 moved names verbatim from the package (one `export { … } from "@evener/appwire-client/state/navigation"` block) — the 15 web importers plus `rail/actions.ts` don't change. Two of the 29, `selectGlobalRows` and `selectPinSectionSummaries`, are additionally **redefined** (not just re-exported) as thin wrappers that default their `state` parameter to `navigationStore.getState()`: measured, these are the **only two** selectors read with no state argument anywhere in the frontend (8 call sites, all in `store.test.ts`), so removing the default in the package would have meant editing those 8 call sites instead of keeping a 2-line adapter wrapper. Everything else keeps the state argument the package selector already declares as required.

## Native

Nothing to adopt. `navigationTree.ts` (21 lines) is a generic tree flattener with no web counterpart; `rosterSearch.ts` (135) is roster-label disambiguation over `mobile/src/services/roster` (D16, withdrawn); native reads `.children` only as an accessor passed to list components — it never walks the tree for a ref, and never reads `omitted_watches`, `default_expanded` or `attentionSummary`. Same finding as D14-3: native has no navigation store, so it has no selectors over one.

## Oracle status

`stores/navigation/selectors.test.ts` — **byte-identical, passing**, no fixture or assertion touched (confirmed: `git diff origin/main...HEAD` shows zero changes to this file). `store.test.ts`'s 8 no-argument call sites still resolve through the adapter's two wrappers. No web importer of `selectors.ts` changed.

## Gates (all foreground, on the merged head `51a092835`)

- `make test-api-package` — qualified, plus the falsification (deleted `state/navigation/index.ts`'s `export * from "./selectors"`; failed with `shipped module unreachable from every published specifier: state/navigation/selectors`; restored with `git diff | git apply -R`)
- `make test-native` — PASS (iOS bundle exported; 791 native tests, 771 shared tests)
- `make test-web-browser` — PASS all seven guards (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard, retirementguard, skillguard)
- `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend` — 187 files, clean
- `npm run lint` in `cmd/evener-hub/frontend` — 1323 files, clean
- root `go test -short -count=1 .` — `ok 21.530s`

`git diff origin/main...HEAD --stat`: 8 files, 365 insertions / 294 deletions — `appwire-client/typescript/{README.md, scripts/qualify-package.mjs, state/navigation/index.ts, state/navigation/selectors.ts (new), tsconfig.build.json}`, `cmd/evener-hub/frontend/src/stores/navigation/selectors.ts`, the plan doc, the worksheet.


obra:
## /simplify disposition (51a092835)

Four angles, two reviewers. Reuse and altitude clean: the moved set is exactly the pure-over-state set, the rail-shaped and host-memoized selectors stay web-side (the latter because it is the one impure selector, mutating a cross-call identity cache), and the store's private summary scan beside the public selector is #1596. Efficiency unchanged by the move. Accepted for round 1: the two adapter wrappers that exist only to spare eight test call sites go (the tests pass state explicitly; assertions untouched); the redundant `NavigationState` alias goes; six zero-caller exports carried across the boundary are deleted after a repo-wide re-grep; the web selectors suite moves into the package if it drives only state objects. RoboRev's verdict on this head is pending; merges on CI green + RoboRev clean at the round-1 head.

obra:
## Round 1 fixes (bae23772b)

1. Deleted the adapter's `selectGlobalRows`/`selectPinSectionSummaries` wrappers (`selectors.ts:54-59`, defaulting `state` to `navigationStore.getState()` for 8 no-argument call sites in `store.test.ts` at 607, 637, 835, 850, 881, 964, 1400, 1726). Those 8 now pass `navigationStore.getState()` explicitly — argument edits only, assertions untouched. Both names are now plain re-exports alongside the other 27; the adapter's now-unused `selectGlobalRows as selectGlobalRowsIn`/`selectPinSectionSummaries as selectPinSectionSummariesIn` aliased imports are gone too.
2. Dropped `NavigationState = NavigationStoreState` (package `selectors.ts:20`). Every signature now uses `NavigationStoreState` directly — the package file, the adapter, `README.md`'s prose, and `qualify-package.mjs`'s smoke.
3. Re-grepped the whole repo (web, native, package, tests) for each of the six carried-across zero-caller exports before touching any of them:
   - `selectCatalogRemaining`, `selectNextCatalogOffset` — zero callers anywhere, not even internal. Deleted entirely.
   - `selectResource`, `selectSectionRows` — no external caller, but still needed internally (`selectLocation`; `selectNeedsYouRows`/`selectLiveRows`). Lost only their `export`.
   - `LoadedPinSection` — no external caller of the *type name* (callers of `selectPinSections`, which returns it, never name it), but still needed as that function's return type. Lost its `export`.
   - `selectProjectSummaries` — measured #1596 first: the core's private `selectSummaries` (`store.ts`, boot fan-out) does the same catalog scan unsorted, while this selector sorts by catalog order then offset; folding them would change which projects hydrate first under the 4-worker boot pool. Left the store alone per the fallback and kept the export, with a comment on the function pointing at #1596 explaining why it stays public with no caller today.
4. Measured `cmd/evener-hub/frontend/src/stores/navigation/selectors.test.ts` (489 lines, 18 tests): 9 drove only state objects (navigation-view-scope Go-parity vectors, `isSettledGone`, `selectSources`/`selectDisplaySources`, `selectSessionOmittedWatches`/`selectSessionOmittedArmedWatches`) with no zustand store or DOM dependency — moved verbatim into the new `appwire-client/typescript/state/navigation/selectors.test.ts`, adapted only to build the base state via `createNavigationStore({ persistence: memoryNavigationPersistence() })` instead of the web singleton `navigationStore.getState()` (a fixture-plumbing swap, not a logic change; the `isSettledGone` tombstone helper's `normalizedGraphFromSnapshot` call is untouched). The other 9 (all `selectRailModel`/`selectSessionWatches`) stay web-side — they need the rail's own types or the impure identity cache, so they genuinely need to stay.
5. `selectSessionWatches`'s comment (adapter file header, now lines 1-6) states the real reason it's web-only: it's the one impure selector here, mutating a cross-call identity cache — not an instance-sharing argument. Corrected the same framing in the D15 plan row.

Also fixed the follow-up RoboRev Low: `state/navigation/index.ts`'s header (lines 1-9) named seven modules and omitted `selectors` while `README.md` already said eight; added it to the header sentence.

## Gates (all foreground, on the merged head `bae23772b`)

- `make test-api-package` — qualified, plus the falsification (deleted `state/navigation/index.ts`'s `export * from "./selectors"`; failed with `shipped module unreachable from every published specifier: state/navigation/selectors`; restored with `git diff | git apply -R`)
- `make test-native` — PASS (iOS bundle exported; 791 native tests, 771 shared tests)
- `make test-web-browser` — PASS all seven guards
- `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend` — 188 files, clean
- `npm run lint` in `cmd/evener-hub/frontend` — 1324 files, clean
- root `go test -short -count=1 .` — `ok 24.461s`

`git diff origin/main...HEAD --stat`: 11 files, 552 insertions / 484 deletions.

obra:
## Review round 2 disposition (bae23772b)

CI 15/15 green, RoboRev Low only, simplify pass done (round 1). Merging at this head per the Lows-only rule; both Lows (README still advertising the deleted catalog read-ahead selectors; the header comment claiming both apps read through these selectors when native has no navigation store) go into an immediate follow-up PR.

PR #1611, opened 2026-09-16
refactor(sdk): mutation record shapes and client identity at @evener/appwire-client/state/mutation (D25a)

obra:
## Summary
- `MutationIntent`/`MutationRecord`/`MutationOutboxRecord`/`MutationOptimisticRecord`/`MutationRecoveryRecord` (from `stores/mutationOutbox.ts`) and the client-provenance pair `ownClientId`/`isOwnMutationRecord` (from `stores/mutationClientIdentity.ts`) move to a new `@evener/appwire-client/state/mutation` subpath (`records.ts` + `index.ts` barrel).
- No DOM type enters the package: the record shapes take their attachment type as a defaulted generic parameter (`MutationAttachmentRef`), and the web instantiates `MutationAttachment extends MutationAttachmentRef { blob: Blob }`, aliasing its five record names to that instantiation.
- `ownClientId` takes a structural `ClientIdentityStorage` port (`getItem`/`setItem`) defaulting to `globalThis.sessionStorage`, so its zero-argument web call sites are unchanged.
- Both web modules (`mutationOutbox.ts`, `mutationClientIdentity.ts`) become re-exports/aliases: zero importers changed across the nine outbox-adjacent web files, and `mutationClientIdentity.test.ts` stays the oracle unchanged, exercising the package module transitively.
- Subpath plumbing copied from the D10 precedent: `package.json` exports, `tsconfig.build.json`, the qualification manifest (with a synchronous-shaped smoke, since `ownClientId`/`isOwnMutationRecord` make no request), README, and the five resolver aliases (frontend `tsconfig.json`, `vite.config.ts`, native `tsconfig.json`/`tsconfig.check.json`/`vitest.config.mts`; Metro needs no entry — it resolves subpaths dynamically).
- Plan status row D25a added.
- `secureUUID.ts` is untouched — that is D26's row, which depends on D23/D24/D25 landing first.

Builds on the stalled `claude/sdk-d25a-mutation-records` branch's existing `records.ts`/`index.ts` (kept as-is; only the plumbing, the web re-exports and formatting were added), merged forward onto current `origin/main`.

## Test plan
- [x] `make test-api-package` (plus falsification: removed the barrel's re-export, confirmed `shipped module unreachable from every published specifier: state/mutation/records`, restored via `git diff | git apply -R`)
- [x] `make test-web`
- [x] `make test-native`
- [x] `make test-web-browser` (all six guards, including skillguard, passed on the first run)
- [x] `make lint-package-imports`
[...]

PR #1638, opened 2026-09-17
refactor(sdk): the pending-turns projection store lives in @evener/appwire-client/state/mutation (D25c)

obra:
## Summary

SDK migration row D25c. Stacked on D25b (`claude/sdk-d25b-pending-entries`, not yet a PR) which itself sits on D25a (#1611, getting a round-1 class fix from another lane: `ownClientId`/`isOwnMutationRecord` become `createClientIdentity(storage)` and `pendingEntries` takes the identity capability). **Held: do not merge until D25a's fix lands and D25b merges forward; this branch will need to merge that forward and adapt** (the identity-capability change touches `isOwnMutationRecord`, which `recordSubmittedHere` in this PR calls directly).

- Moves the framework-free half of `pendingTurnsStore.ts` (510 lines) to `state/mutation/pendingTurns.ts`, extending the `state/mutation` subpath D25a/D25b built: the `outbox`/`optimistic`/`recovery`/`submittingRefs`/`submittedHere` state, `recordSubmittedHere`'s discovery over `isOwnMutationRecord`, the submitting-ref guard (`beginSubmission`/`endSubmission`), the submitted-draft clear decision (`settleSubmittedDraft`) and the `reconcilePendingEntries` projection (`pendingTurnEntries`) become `createPendingTurnsStore({ threads, draft })`, a `FrameworkFreeStore` generic over the attachment type (like the rest of this subpath's records — the web's records carry a `Blob` the package never names).
- Two ports, measured against the actual call sites rather than assumed: `PendingTurnsThreadsPort.getThreadModel(ref)` is the whole surface `pendingTurnEntries` needs from a thread store; `PendingTurnsDraftPort` (`readDraftRevision`/`readComposerDraft`/`clearDraft`) is the whole surface a submission's completion needs from composer-draft storage.
- The durable-read machinery — generation fencing, the IndexedDB reads through `threads.ts`, the settle-tracking test helper — stays web-side, untouched: it's genuinely web-specific, not part of what moved. `submitWithPendingTracking` keeps computing `draftUnchanged` (the bare revision check `RecoverySubmissionCommit` carries) separately from `settleSubmittedDraft`'s combined clear decision — conflating them would silently change recovery-resend behavior.
- `createConnectionStore`'s D28 round-1 lesson applied here too: the triple is mutated in place and returned as the same object, not spread into a copy.

**Deviation from the brief, measured not assumed:** the brief expected some existing `pendingTurnsStore.test.ts`/`pendingTurnsStore.edge.test.ts` cases to move to the package suite ("list moved/kept counts"). I read both files end to end (879 + 88 lines, 23 test cases total): every one drives a real IndexedDB-backed submission/recovery/skill-capability flow — none isolates the store from its durable storage. Moved: 0. Kept: 23 (both files pass byte-identical, unchanged). New package suite `state/mutation/pendingTurns.test.ts` (12 tests) drives the extracted core directly over fake ports instead.

## Test plan
- [x] New package test `state/mutation/pendingTurns.test.ts` (12 tests, TDD — one assertion verified to fail when its guard is removed)
- [x] Both existing web oracles pass byte-identical: `pendingTurnsStore.test.ts` (27 tests) and `pendingTurnsStore.edge.test.ts` (1 test)
- [x] `npm run typecheck` / `npm test` (521 files / 11967 tests) / `npm run lint` (frontend): clean
- [x] `make test-api-package` (plus falsification: removed the barrel re-export, confirmed `shipped module unreachable from every published specifier: state/mutation/pendingTurns`, restored via `git diff | git apply -R`)
- [x] `make test-native`, `make test-web-browser` (all six guards green first run), `make lint-package-imports`, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, root `go test -short -count=1 .`


obra:
Closing without prejudice: the queue holds stacked rows as branches until their predecessor merges, and this PR is based on the unmerged D25b branch (and already CONFLICTING against its current head 5873e06cd, which took D25a round 2). The branch `claude/sdk-d25c-pending-turns` stays; D25c reopens against main once D25b merges.

PR #1640, opened 2026-09-17
docs(sdk): navigation selectors README and header match what the package ships (#1630 follow-up)

obra:
Follow-up to D15 (#1630, merged as `9bcadfe27`). Two RoboRev Lows on that PR's docs that its own round 1 didn't catch.

## What's wrong

- `README.md`'s `@evener/appwire-client/state/navigation` entry (and the module header at `state/navigation/selectors.ts:1`) still advertised "section and catalog rows with their remaining counts and next offsets." Round 1 of #1630 deleted `selectCatalogRemaining` and `selectNextCatalogOffset` (zero callers, not even internal) — the package has never had a paginated catalog selector; only sections do (`selectSectionRemaining`/`selectNextSectionOffset`). The catalog side is `selectProjectSummaries`, a full unpaginated list.
- Both places also said "both apps read that state through" these selectors. Native has no `createNavigationStore` (D14-3 measured this and withdrew native's store adoption) and adopts none of the selectors — only the web app reads them today.

## Fix

Dropped the catalog-pagination clause from both the README prose and the module header so each lists only what exists (launch sources; section rows with their remaining count and next offset; pin-section summaries; the project catalog; a session summary by ref). Changed "both apps" to "the web app," and added that native could adopt these selectors if it ever gains a store — comments describe the present, not an aspiration, but they shouldn't foreclose a true future possibility either.

## Gates (all foreground, on `main` head `9bcadfe27`, no merge needed — origin/main hadn't moved)

- `make test-api-package` — qualified
- `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend` — 188 files, clean
- `npm run lint` in `cmd/evener-hub/frontend` — 1324 files, clean

Comment-only change; no test suite is affected.


obra:
## Review round 2 disposition (606552adc)

CI 15/15 green on rerun (the first run's agent-module failure, `TestRetirementSafetyColdJobWatchContent`, cannot be reached by a docs-only change; filed as #1657). RoboRev Low only, simplify by reading. Merging under the Lows-only rule; the two remaining doc lines (barrel header in `index.ts`, the README's "both apps' navigation stores" opener) go into an immediate follow-up.

PR #1652, opened 2026-09-17
build: name the missing install instead of a Metro error in test-native

### LINKED ISSUE #1574: make test-native fails in a fresh worktree with a misleading Metro error when mobile-native has no node_modules
In a fresh worktree without `npm ci` in `mobile-native`, `make test-native` fails with `CommandError: No platforms are configured to use the Metro bundler` because npx pulls a global expo 57 instead of the pinned one. The target's `## requires:` line documents the prerequisite, but the error names Metro platforms, not the missing install. A preflight (like `web-preflight`) that checks `mobile-native/node_modules` exists and matches the lockfile would turn this into a clear message. Hit by the D14-1 implementer (#1569) on 2026-09-16.

### LINKED ISSUE #1666: native-preflight.sh: cover the refusal branches with a deterministic Go test
# native-preflight.sh: cover the refusal branches with a deterministic Go test

## Source and priority

LOW coverage follow-up from [PR #1652](https://github.com/prime-radiant-inc/evener/pull/1652), [RoboRev review of `f6ec1cc`](https://github.com/prime-radiant-inc/evener/pull/1652#issuecomment-5708265407). `scripts/native/native-preflight.sh:38-77` is the whole point of #1574 (turn a misleading Metro error into a named missing install), and none of its branches have automated coverage. Two of three reviewers flagged the same gap.

The gap is visible by comparison: its direct analogue `scripts/web/web-preflight.sh` is covered by `TestWebPreflightBootstrapsMissingFrontendDependencies` in the root `install_test.go:49`. Untested here: a symlinked `node_modules` whose shared lockfile does not match, a stale non-symlink install, a missing install, and an install without an executable `.bin/expo`. A later edit that weakens the symlink check back to an mtime comparison, reorders the symlink and mtime branches, or breaks the `cmp` path would silently restore the confusing `CommandError: No platforms are configured to use the Metro bundler` failure this commit exists to prevent.

## Focused scope

Add one deterministic Go test next to the web-preflight test in the root `install_test.go`, following its shape. Point the script at fixture directories through the existing `EVENER_NATIVE_DIR` override, and assert observable outcomes rather than internal steps:

- missing `node_modules` → non-zero exit, message naming `cd <dir> && npm ci`;
- symlinked `node_modules` whose shared `package-lock.json` differs → non-zero exit, message about the symlink, and the shared lockfile must not be modified;
- symlinked `node_modules` whose shared lockfile matches, with an executable `.bin/expo` → zero exit;
- non-symlink install older than the lockfile → non-zero exit;
- install without an executable `node_modules/.bin/expo` → non-zero exit.

No npm shim is needed: the script never installs. Do not change the script's behavior; if the test shows a real gap, stop and report it rather than folding a behavior change into this ticket.

## Acceptance criteria

- [ ] Each branch above is exercised by the new test, with the fixture for it built by the test rather than committed.
- [ ] `go test -run TestNativePreflight ./...` (or the equivalent package-level invocation) passes, and fails if any branch's guard is removed.
- [ ] The test asserts messages by their actionable content, not by whole-string equality, so wording refinements do not break it.

## Duplicate check (2026-09-17)

```sh
gh issue list --repo prime-radiant-inc/evener --state all --search 'native-preflight' --json number,title,state
```

Only #1574 itself (which #1652 closes), a keybindings issue that happens to share a search token, and the mobile-native handoff epic; no ticket covers the preflight's coverage.

obra:
## What

`make test-native` in a fresh worktree failed with

    CommandError: No platforms are configured to use the Metro bundler

because `npx expo` fell back to a global expo 57 instead of the pinned one. The
target's `## requires:` line documented the missing `mobile-native` install, but
the error named Metro platforms rather than the missing install.

This adds `native-preflight`, the native counterpart of `web-preflight`, wired as
a prerequisite of `test-native-bundle` (and so of `test-native`). It checks that
`mobile-native/node_modules` exists, matches `package-lock.json`, and holds an
executable `node_modules/.bin/expo`, and refuses with a message naming the
directory and the command to run (`cd mobile-native && npm ci`). It never
installs.

## Review findings addressed

- **A symlink's content is now compared before any mtime test.** `[ -L
  node_modules ]` is checked first, so a shared install whose `package-lock.json`
  differs is refused even when its mtime is newer than this worktree's lockfile;
  the mtime freshness test now applies only to a real directory.
- **Existence and freshness no longer stand in for health.** The final check
  requires `node_modules/.bin/expo` to be present and executable, mirroring
  `web-preflight.sh`'s `tsc` probe, so an empty or half-installed tree — or a
  dangling symlink whose target is gone — no longer reaches Metro's error.

## Verification

- Missing install (this worktree has no `mobile-native/node_modules`):
  `make test-native-bundle` stops at `native-preflight` with

      ERROR: <...>/mobile-native/node_modules is missing.
        ...
          cd <...>/mobile-native && npm ci
      make: *** [make/testing.mk:82: native-preflight] Error 1

  and never reaches the Metro bundle. Pointed at an absent directory via
  `EVENER_NATIVE_DIR`, the script prints the same message and exits 1.
- Symlink whose shared lockfile differs, with the symlink target touched newer
  than the worktree lockfile: refused by the content compare, exit 1.
- Empty-but-newer `node_modules` (real dir, so the mtime test passes): refused as
  unhealthy (`node_modules/.bin/expo` is not executable), exit 1.
- Symlink with a matching lockfile and a healthy target: exit 0.
- Real healthy tree
  (`EVENER_NATIVE_DIR=<checkout with mobile-native/node_modules>`): exit 0.
- `make -n test-native` shows the order `native-preflight` -> `test-native-bundle`
  -> `test-native`.
- `bash -n scripts/native/native-preflight.sh` and `git diff --check` are clean.
- `make lint-generated` passes (the new `##` row in
  docs/developing-evener/testing.md is regenerated).
- Focused audits pass:
  `go test -run 'TestEveryRuleIsPhony|TestEveryPhonyTargetHasARule|TestEveryTargetHasASummaryAnnotation|TestEveryRuleHasARecipe' .`
  and `go test ./internal/maketargetsdoc/`.

Closes #1574

Closes #1666


obra:
Both review findings are addressed in `f6ec1cc9b` (branch rebased onto current `main`, `a18eb4626`).

1. **Symlink before mtime.** The check order is now `[ -L node_modules ]` first, then a real-directory mtime test. A symlink is always settled on `$shared/package-lock.json` content, so a newer mtime can no longer wave a mismatched shared install through.
2. **Health, not just existence/freshness.** A final `[ -x node_modules/.bin/expo ]` check mirrors web-preflight's `tsc` probe, so an empty/half-installed tree and a dangling symlink fail with a message naming `cd mobile-native && npm ci` instead of reaching Metro.

Verification run locally (no full gates, no npm install):
- symlink with a differing shared lockfile (and target mtime touched newer) -> exit 1
- empty-but-newer `node_modules` -> exit 1 (unhealthy)
- symlink with matching lockfile + healthy target -> exit 0
- real healthy tree -> exit 0
- missing install via `make test-native-bundle` -> clear message, make exit 2, bundle never starts
- `bash -n`, `git diff --check`, `make lint-generated`, and the focused Makefile audits all pass

obra:
The medium from the review at `b678e6d` is addressed in `3e4a49713` (one commit on top of `b678e6d8b`, no force-push; `install_test.go` only).

`TestNativePreflightRefusesUnreadyInstalls` drives the real `scripts/native/native-preflight.sh` at fixture directories through `EVENER_NATIVE_DIR`, next to `TestWebPreflightBootstrapsMissingFrontendDependencies`, and pins:

- missing install → non-zero exit, naming the directory and `cd <dir> && npm ci`;
- stale real install → refused by freshness, not health;
- symlinked `node_modules` whose shared lockfile differs → refused, and the shared tree's lockfile is not the one modified;
- install without an executable `.bin/expo` → refused by the health probe;
- healthy real install and healthy shared symlink → exit 0, no refusal.

Each guard is pinned by a red-first run: weakening it makes exactly its case fail. Two of those I checked myself on this commit rather than taking the report on faith — with `if ! cmp -s "$shared/package-lock.json" package-lock.json` replaced by `if false`, `TestNativePreflightRefusesUnreadyInstalls/symlink_whose_shared_lockfile_differs` fails; restored, the whole test passes (`ok primeradiant.com/evener 0.186s`). The script itself is unchanged, so the symlink-before-mtime ordering this test protects is still exactly what ships.

PR #1658, opened 2026-09-17
docs(sdk): navigation docs stop claiming both apps build on the store (#1640 follow-up)

obra:
Fixes the two Lows RoboRev left on #1640 (merged as 3ed14eacf under the Lows-only merge rule).

1. `appwire-client/typescript/state/navigation/index.ts` barrel header said the selectors were what "both apps read that state through." Native has no `createNavigationStore` and imports none of the selectors (only codec/merge/type helpers from this subpath). Reworded to match `selectors.ts` and the README: the web app reads state through them; native can adopt them if it ever gains a store.
2. `appwire-client/typescript/README.md` navigation subpath entry opened with "the navigation state layer both apps' navigation stores are built on," contradicting the corrected sentence further down the same paragraph ("the selectors the web app reads that state through ... adoptable by native if it ever gains a store"). Scoped the opening claim to the web app's store, adoptable by native.

Grepped the subpath and README for any other "both apps" claim about the store or selectors (`grep -rn "both apps" appwire-client/typescript/state/navigation appwire-client/typescript/README.md`). The only other hit under `state/navigation` is `types.ts`'s "resource vocabulary both apps share," which is accurate (native does import the type helpers) and out of scope. Other README "both apps" lines are about unrelated subpaths (extensions, model catalog, keybindings, session settings) and weren't touched.

Refs #1640, #1630.


obra:
## Review disposition (673309c5c)

CI 15/15 green, RoboRev clean, simplify by reading (two documentation lines). Merging.

PR #1667, opened 2026-09-17
fix(sdk): a client swapped through setState directly rewires the connection listener (#1623 follow-up)

obra:
## Summary

Follow-up to #1623 (merged as `f82b4122fb0bf907fe76cc7939c9492c6b20bd18`), round 3 finding on `state/connection/core.ts:102-117`. Two of three reviewers doubted reachability past the generation-fenced `connect()`. Measured, not refuted.

**Measurement**: added a test that replaces `state.client` via `store.setState({ client: second, ... })` directly (bypassing `connect()` — the exact pattern several existing test files already use: `CommandPalette.test.tsx`, `commandCatalog.test.ts`, `threads.test.ts`, `SessionChrome.test.tsx` and others all call `connectionStore.setState({ client: fake, ... })` rather than `connectionStore.getState().connect(fake)`), then (a) emits a state change from the new client and (b) emits one from the old client. Result: (a) failed — the new client's transition never reached the store (`store.getState().state` stayed `"ready"`, expected `"closed"`) — nothing had ever subscribed to it, since only `connect()` ever wired the listener. (b) passed on its own — the old client's stale listener, though never detached, is harmlessly guarded by the existing identity check. The finding is real and reachable (via `setState`, not merely a hypothetical).

**Fix**: extracted `connect()`'s wiring into `wireClient(client, publish)`, and replaced the triple's own `setState` to detect a `client` key that differs from the current one and route it through the same `wireClient` — no second counter, the existing `connectGeneration` fence from round 1/round 2 covers both entry points now. The listener's own state-change publish still goes through the live, dynamic `store.setState` (so a spy on it — `threads.test.ts`'s idempotency check — keeps seeing every state change, per round 2's fix); only `wireClient`'s own client-carrying write bypasses it via a captured `rawSetState`, since re-entering the interception on that one write would recurse.

## Test plan
- [x] New package test (`core.test.ts`): the exact (a)/(b) measurement above, now green
- [x] Existing oracles unchanged and green: `connection.test.ts` (5 tests), `threads.test.ts`'s setState-spy idempotency check (312 tests)
- [x] `npm run typecheck` / `npm test` (525 files / 12046 tests) / `npm run lint` (frontend): clean
- [x] `make test-api-package` (plus falsification: removed the barrel re-export, confirmed `shipped module unreachable from every published specifier: state/connection/core`, restored via `git diff | git apply -R`)
- [x] `make test-native`: Metro bundle + native/shared vitest (82+7 files, 792+771 tests) + typecheck + script resolution
- [x] `make lint-package-imports`, `npx biome ci ../../../appwire-client/typescript`, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, root `go test -short -count=1 .`: all pass

**Residual**: `make test-web-browser`'s `web-skillguard` failed twice in a row on this host (timeout waiting for a visible contenteditable in the skill editor) — unrelated to this change (composer skill-editor UI rendering, not connection-state wiring) and not part of this round's requested gate list; likely host contention from concurrent worktrees. Not investigated further here.


obra:
## Simplify disposition (fe079ef4c)

Two cleanup reviewers read the diff. Applied on this PR before merge (in flight):
1. One function writes `client`: the wiring lives in the store's `setState` override and `connect` becomes a thin wrapper, which removes the `publish` parameter and the `rawSetState` capture. Forbidding raw client writes was measured as the larger change (15 test call sites in 6 files) and rejected.
2. The notification `attach` folds onto the same unwire-then-wire primitive where it fits.
3. The module header states the invariant once (whatever the client key reads, exactly one listener is wired to it); the review-round narration in the comments goes.
4. The null-client path gets its own test with the file's wiring-count helper.
Efficiency angle: clean.

obra:
## Review round 2 disposition (fe079ef4c)

CI 15/15 green. RoboRev: 2 Medium, 2 Low, all accepted and folded into the simplify push in flight.

- **`rawSetState` bypasses the public `setState` the web instruments**: resolved by the simplify shape (the `setState` override is the only client writer and `connect` calls it), pinned with a spy test; no raw capture survives.
- **A client swap through `setState` publishes the partial verbatim** and **`client: undefined` is stored as `undefined`**: the client-changing write normalizes like `connect` did (state defaults to the client's, handshake metadata cleared unless provided, `undefined` published as `null`), with tests for the three cases.
- **The swap test cannot see the detach and its comment is wrong**: it asserts one detachment through the file's wiring tracker and the comment is corrected.

obra:
## Round 2 fixes (5185ed903)

Two commits: the simplify pass, then RoboRev's follow-up findings on it.

### Simplify pass (commit 778d013b0)

1. **Folded `wireClient` into the `store.setState` override.** The override IS the wiring now: detects a differing `client` key, unwires the outgoing listener, wires the incoming one under the existing `connectGeneration` fence, publishes. `connect(client)` is now `store.setState({ client })` - the `publish` parameter and the `rawSetState` capture both went with the fold, since there's nothing left to re-enter with a single writer.
2. **Left `onConnectionNotification`'s `attach` as its own primitive** - didn't fold it in. It's the same idiom (unwire outgoing, wire incoming, null is a valid detached state, identity-guarded) but has no generation-fencing need of its own (it isn't re-entered the way a setState publish can be), so sharing a primitive with the connection-state wiring would mean either carrying dead-weight generation logic into `attach` or parameterizing it away - either costs more than the ~10 lines of duplication it would save.
3. **Stated the invariant once** in the module header ("whatever the `client` key reads, exactly one listener is wired to it, and an event from any other client reaches nothing") and deleted the round-N narration and the test-file listing from the comments - they described history, not the present.
4. **Added the null-path test**: `setState({ client: null })` detaches the currently-wired listener, asserted via `trackStateChangeWiring`'s registration/detachment counts directly.

### RoboRev findings on the fold (commit 5185ed903)

- **Medium 1 (refuted by the fold's own shape, confirmed by a test):** `connect()` now calls the same observable `store.setState` directly - no bypass survives to hide it from a spy. Added a pinning test: spy on `setState`, call `connect(client)`, assert one call with `{ client }`.
- **Medium 2 + Low 1 (real, fixed):** the client-changing branch was publishing the caller's partial verbatim, so `setState({ client: second })` kept the OLD client's `state`/`serverInfo`/`features`, `setState({ client: null })` could leave a stale non-idle `state`, and `client: undefined` (a valid `Partial<ConnectionStoreState>` value TS does not reject) was stored as `undefined` rather than `null`. Fixed: the write now defaults `state` to the incoming client's own state (or `"idle"` for null - the store's own starting value, and what every existing `client: null` reset already pairs it with), clears `serverInfo`/`features`, and writes `client` explicitly as `resolved.client ?? null` - all three yielding to the caller's own partial where it names the same key, so a swap that already has its `InitializeResponse` in hand can still publish it in one write. Three new tests cover the swap-defaults case, the swap-with-explicit-values case, and the `undefined`-stored-as-`null` case.
- **Low 2 (real, fixed):** the swap test asserted detachment only by inference (the identity check alone would also make the emit a no-op) and its own comment said the old listener was never detached - both wrong once the fold's `setState` override wires every client-changing write. Wrapped the outgoing client with `trackStateChangeWiring` and asserted `detachments === 1` directly, and corrected the comment.

**Verification**: temporarily reverted the normalization write to `publish(resolved)` (no defaults) and reran - 6 of the 16 tests failed exactly as expected (three pre-existing tests plus the three new normalization tests), confirming the fix and the tests are both load-bearing. Restored before committing.

### Gates (merged with `origin/main`, no conflicts)
- `npm run typecheck` / `npm test` (525 files / 12051 tests) / `npm run lint`: clean
- `make test-api-package` (+ falsification: removed the barrel re-export, confirmed `shipped module unreachable from every published specifier: state/connection/core`, restored)
- `make test-native`: Metro bundle + native/shared vitest (82+7 files, 792+771 tests) + typecheck + script resolution
- `make lint-package-imports`, `npx biome ci ../../../appwire-client/typescript`, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, root `go test -short -count=1 .`: all pass

Head SHA: `5185ed90365b1e11b1c7cc9f7a8b0d9f0e1cb2f8`

obra:
## Review round 3 disposition (5185ed903)

CI 15/15 green. RoboRev: 1 Medium, 1 Low, both accepted.

- **A re-entrant updater can publish a stale, unwired client**: the override decides whether the client changed against the store state re-read after the updater resolves, pinned by a test whose updater connects a second client inside and returns the first.
- **`connect` with the already-current client is no longer a no-op**: the wrapper's same-client early return comes back (the direct `setState` path keeps the framework-free store's notify-always semantics), pinned by a subscriber-count test.

obra:
## Round 3 fixes (014915819)

Two commits, one per finding, merged with `origin/main` first (D25a's simplify commits landed, no conflicts).

**Medium** (`core.ts:109-112`): the client-change decision compared an updater's return value against the snapshot the updater ran against (captured before calling it), not a fresh read. An updater that synchronously re-enters `setState`/`connect()` (wiring a different client for real, via a nested publish) and then returns the client its own stale parameter named looked like no change relative to that snapshot, so the write skipped rewiring - the store ended up reporting one client while only another's listener was attached.

Fixed by re-reading state after resolving the updater and comparing against that, not the pre-call snapshot the updater's own parameter is (correctly) still bound to. Chose the re-read over refusing re-entrant client writes with a thrown error: smaller (one re-read, no new guarded-error path) and keeps `setState`'s single ownership-by-generation path as the only thing a caller has to reason about, rather than adding a second failure mode.

Test: an updater that calls `connect(b)` inside and returns `{ client: a }` (the snapshot it started from) - red against the unfixed code (`aTracking.registrations` came back 1, expected 2 - no corrective rewiring happened), green after, with `trackStateChangeWiring` counts proving exactly one listener (a's second) ends up live and a closing live event proving it's the one that actually reaches the store.

**Low** (`core.ts:149-151`): `connect(client)` lost its same-client early return in the earlier setState-fold, so every repeated call (an effect re-running under React StrictMode's double-invocation, say) went through `setState` regardless, allocating a new state object and notifying every subscriber. Restored the guard on `connect()` itself; a direct `setState({ client: same })` call is intentionally unaffected and still notifies, per the framework-free store's own documented contract of publishing every write even when nothing changed.

Test: `connect(a)` then `connect(a)` again with a subscriber installed in between - red against the unfixed code (1 notification, expected 0), green after.

### Gates (merged tree)
- `npm run typecheck` / `npm test` (526 files / 12067 tests) / `npm run lint`: clean
- `make test-api-package` (+ falsification: removed the barrel re-export, confirmed `shipped module unreachable from every published specifier: state/connection/core`, restored)
- `make test-native`: Metro bundle + native/shared vitest (82+7 files, 793+771 tests) + typecheck + script resolution
- `make lint-package-imports`, `npx biome ci ../../../appwire-client/typescript`, `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, root `go test -short -count=1 .`: all pass

Head SHA: `01491581937f0638860f64e6def5ca1c2b523b5f`

obra:
## Review disposition (014915819)

CI 15/15 green, RoboRev clean, simplify pass done (round 2). Merging.

PR #1674, opened 2026-09-17
test(sdk): mutation id uniqueness tests use scripted randomness (#1611 follow-up)

obra:
## Summary
- #1611's RoboRev Low: `secureUUID.test.ts`'s "neither method" case and `records.test.ts`'s "two instances over two storages keep separate identities" case asserted generated ids differ while depending on real `Math.random`/`Date.now`/`crypto.getRandomValues` never colliding — theoretically flaky.
- `secureUUID.test.ts`: stubs `Math.random`/`Date.now` with `vi.spyOn` returning a scripted sequence, so the two-call-differ assertion is deterministic.
- `records.test.ts`: the two-instance uniqueness case now injects a scripted `SecureRandomSource` (`randomUUID` returning a fixed id per instance) instead of real `getRandomValues`, so the two identities are provably different rather than merely astronomically unlikely to collide.
- No production change; every existing assertion's meaning is unchanged.

## Test plan
- [x] `make test-web` (typecheck/test/lint split) — PASS
- [x] `make test-api-package` — PASS
- [x] `npx biome ci ../../../appwire-client/typescript` (from `cmd/evener-hub/frontend`) — PASS
- [x] `make lint-package-imports` — PASS
- [x] `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` — PASS


obra:
## Review disposition (30aafb39a)

CI 15/15 green, RoboRev clean, simplify by reading (two test files). Merging.

Issue #1669, opened 2026-09-17
web-skillguard fails locally on main: the native-clipboard milestone times out


obra:
`make test-web-browser` is red for every lane on macOS right now because of the newest guard, and it is not the branch under test.

Measured 2026-09-16 in worktree `sdk-d21-native-model`, with `make build-web` fresh:

- `origin/main` @ `7feef878e`: `go test -tags browserguard ./cmd/evener-hub -run TestSkillComposerBrowser` → `skillguard FAIL: timed out after 15000ms waiting for native clipboard preserves canonical display text`
- the same command on an SDK-migration branch: identical failure, same milestone.
- an earlier run on that branch timed out one milestone earlier (`visible contenteditable for local:…`), so the guard also has a timing component on a loaded host.

The milestone is `cmd/evener-hub/frontend/scripts/skillguard/run.mjs:1619-1627`: it copies with a real `Cmd+C`, focuses the composer, types a space and pastes with `Cmd+V`, then waits for the composer state's text to equal `"<TWO_SKILLS> <TWO_SKILLS>"`. That drives the OS clipboard through Chrome, which is the part that does not hold on this host (no clipboard permission granted to the guard's Chrome profile / no user-activation for a paste). Everything before it passes, and the other eight guards pass.

Two things to decide:
1. whether the clipboard round-trip can be driven without the OS clipboard (`Input.insertText` with a synthetic `paste` event carrying the canonical text in the DataTransfer, or granting `clipboard-read`/`clipboard-write` through `Browser.grantPermissions` in the guard's CDP setup), and
2. whether the 15s waits need the same CPU-throttle treatment the other guards got (see the flake note in the browser-guard docs).

Until then `web-skillguard` is a known-red local gate, and lanes should report the other eight plus this citation rather than treating it as their own regression.

obra:
Evidence from a Linux host, at `dc92e9f1e` (current main), which narrows this to your environment rather than the milestone logic:

```
$ make build-web && go test -tags browserguard ./cmd/evener-hub -run TestSkillComposerBrowser
✓ built in 849ms
ok  	primeradiant.com/evener/cmd/evener-hub	47.875s
```

Exit 0, whole guard green, including the `native clipboard preserves canonical display text` milestone. The harness only prints milestone lines on failure, so there is no per-milestone line to quote on a pass; a timeout in that milestone fails the guard, and it did not. So the milestone sequence itself works on Blink with a real OS clipboard round-trip on Linux.

[...]

Issue #1616, opened 2026-09-16
appwire-client: two idioms for 'own client id' across state/credentials and state/mutation


obra:
After D25a (#1611) the package has two shapes for the same concept: `state/credentials/instances.ts:210` takes client identity as an injected capability (`ownClientId: () => string` on its deps), while `state/mutation/records.ts:213-259` owns it as a module singleton (`let identity`) with its own `ClientIdentityStorage` port and default. The web wires them together (`stores/credentials.ts:31` passes the mutation module's `ownClientId` into `createCredentialInstancesStore`), so nothing is wrong today, but a third store needing "who is this client" faces a decision each time (inject it, or reach for the mutation module's global). Pick one canonical shape (the injected capability, with the mutation module's implementation as the default provider both hosts hand in, is the D14-1/D6 port precedent) and apply it when the next consumer appears (D27's dispatcher is the likely one). Found by the /simplify reuse pass on #1611.

PR #1678, opened 2026-09-17
refactor(sdk): the mutation outbox record core lives in @evener/appwire-client/state/mutation (D26)

obra:
Plan row **D26** (`docs/superpowers/plans/2026-09-12-sdk-migration.md`, the serial-queue table), stacked on **#1611** (D25a, merged as `77d636835`).

## What moves

The outbox's **discovery half** — `MutationOutbox` — moves from `cmd/evener-hub/frontend/src/stores/mutationOutbox.ts` to `appwire-client/typescript/state/mutation/outbox.ts`, over a **`MutationOutboxStorage` port**: the 13 calls this layer and D27's dispatcher actually make on the web's IndexedDB adapter (`enqueueIntent`, `listTargetRefs` for this class; `getOutbox`, `getOptimistic`, `listOptimistic`, `getRecovery`, `nextDispatchable`, `markAttempted`, `markUnknown`, `settleReceipt`, `settleApplied`, `restoreProvenAbsent`, `transferToRecovery` for the dispatcher), measured from the call sites rather than mirroring the adapter's ~20 methods. `MutationOutboxIndexedDB` stays on the web and implements the port.

Every host-shaped capability is an **option with no browser default** — the sibling-client channel, the lifecycle and visibility targets, the timer — so the package names no DOM type (per BRIEF-COMMON's package rule and #1616's ruling). `stores/threads.ts` passes the browser's four explicitly at the one construction site; that plus binding its runtime to the attachment type that carries bytes is its only change.

`mutationOutbox.test.ts` (855 lines) is **byte-identical**, imports included: the web module re-exports the class under the same name, so no importer moved.

## Notes

- `secureUUID.ts` moved in **D25a** instead (its `newClientIdentity` needed `createSecureUUID`), as `createSecureUUID(source)` over a required `SecureRandomSource`; the D26 inventory row said otherwise and is corrected here.
- Measured against `ConversationMutationState`: it aliases nothing in this layer — native's UI ownership bookkeeping (kind/status/draftSnapshot/generation/mutationId) has no durable identity, payload or persisted state, and no wire `PendingMutation` shape either. Recorded as a note in `outbox.ts`; **D25d** is the row that gives native records. **No native file is touched by this PR.**
- D27 (`mutationDispatcher.ts` over the same port) is held behind this one.

## Gates

`make test-api-package` (qualification clean, plus the index falsification: dropping `export { MutationOutbox }` from the barrel fails it with `TS2304`), `make test-native` (82 files / 793, shared 7 / 771), `make test-web` (web-typecheck / web-test / web-lint PASS), `make lint-package-imports`, biome from `cmd/evener-hub/frontend` (1340 files), `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`, root `go test -short -count=1 .`.

`make test-web-browser`: eight of nine guards PASS. `web-skillguard` is red on `origin/main` too, at the same milestone (the real clipboard round trip) — measured and filed as #1669, not from this branch.


obra:
## Review round 1 disposition (b9d693ff0)

CI 15/15 green. RoboRev: 1 Low, accepted: `setInterval`/`clearInterval` become one all-or-nothing timer port and the scan callback is guarded after `stop()`, with a scripted-timer test. Lands with the simplify pass (two reviewers reading now) in one push; merge follows that head's review. D27 re-stacks on it.

obra:
## Simplify disposition (b9d693ff0)

Two cleanup reviewers read the diff. Layer placement is right (no host global, every capability an injected option), the README prose is accurate, efficiency clean. Applied on this PR with the Low, in one push (in flight):
1. The `postMessage` guard uses the subpath's own `tryOrUndefined`.
2. The qualifier's eleven-method storage stub is written once.
3. Three re-exported types with no web consumer leave the adapter.
4. The unrequested `scanIntervalMs` option goes; the interval stays a named constant.
5. The plan's D26 row gets `#1678`.
Signed off without change: the storage port declares the dispatcher's methods one slice early as one contract (D27 is stacked on this branch), and the class's coverage rides on the web suite through the re-export, as the plan row states.

obra:
## Simplify fixes (`046add321`)

Refreshed onto `7fb903068` first — D25b (#1680) had landed, so the four shared
files (the barrel, the README, the qualification manifest, the plan table) are
union-resolved: `reconcilePendingEntries` and the outbox live side by side, and
every D25a/D25b-touched file diffs **empty** against `origin/main`.

**RoboRev's Low — the timer was two independent ports (`5cb705728`).**
`setInterval` and `clearInterval` were separately optional, so a host that passed
only the first got a scan loop `stop()` could not cancel. RED first, in the web
oracle:

```
× a stopped outbox runs no further scan, even from a timer callback that already fired
  AssertionError: expected [ 'interval' ] to deeply equal []
```

Two changes, both small: the constructor takes the timer as a **pair** (one
without the other is no timer, so nothing is ever scheduled that cannot be
cancelled), and `#scheduleReadyScan` returns unless the outbox is still started —
a host's callback can already be queued when `clearInterval` lands, and a
lifecycle event can arrive during shutdown. An **enqueue** is deliberately not
gated: its record is durable and announcing it costs nothing. The test asserts
both halves (the id is cleared, and a late callback discovers nothing).

**Simplify 1 — the guarded broadcast uses `tryOrUndefined` (`66f600e20`).** The
hand-rolled `try { postMessage } catch {}` is now the helper the sibling module
already uses for a host capability that may throw.

**Simplify 2 — the qualifier names the inert storage port once (`20c2d5885`).**
The eleven methods neither fixture exercises were written twice; one definition
now, interpolated into the type-use program and the smoke script, with the two
calls each actually makes (`enqueueIntent`, `listTargetRefs`) beside it. 22 lines
out, 16 in.

**Simplify 3 — the web shim re-exports only what the app names (`0486d742a`).**
`MutationDiscoveryReason`, `MutationOutboxChannel` and `MutationOutboxStorage`
had zero consumers in `src/`; `MutationOutboxOptions` stays because the web test
names it. A host that needs the ports reads them from the package.

**Simplify 4 — `scanIntervalMs` is gone (`92c025eeb`).** A relocation should not
grow surface: nothing passed it and nothing tested it. The interval is the named
`SCAN_INTERVAL_MS` the web's oracle pins. D27 dispatches over the same port and
does not schedule, so it needs no option either.

**Simplify 5 — the D26 row names its PR (`1085ceeb1`).** `| D26 | #1678 |`.

Signed off with no change, as agreed: the storage port declares D27's eleven
dispatcher methods a slice early (one contract rather than two halves — D27 is
stacked on this branch), and `outbox.ts`'s coverage rides on the web suite through
the re-export, which the plan row states.

Gates on `046add321`: `make test-api-package` (qualification clean, plus the index
falsification — dropping `export { MutationOutbox }` from the barrel fails it with
`TS2304`), `make test-native` (82 files / 793, shared 7 / 771), `make test-web`
(web-typecheck / web-test / web-lint PASS), `make lint-package-imports` PASS,
biome from `cmd/evener-hub/frontend` clean (1341 files),
`EVENER_GITLEAKS_REQUIRED=1 make secret-scan` PASS, root
`go test -short -count=1 .` ok. `make test-web-browser`: eight of nine guards
PASS, `web-skillguard` the known-red one (#1669, red on `origin/main` too at the
same clipboard milestone).

Diff vs `origin/main`: 9 files, 460 insertions, 204 deletions.

obra:
## Review round 2 disposition (046add321)

CI 15/15 green. RoboRev: 2 Medium, 2 Low, all accepted.

- **Shutdown gate checks scheduling, not execution** and **stale readiness in a queued callback**: closed as one class: a single "may this outbox still discover?" predicate (started and ready) evaluated at execution time inside the queued discovery, with `connectionReady` returning early when stopped and `scheduleDiscovery` gated the same way; the documented `enqueueIntent` exception stays; the offline startup scan is preserved.
- **Coverage for work queued before `stop()` and `connectionReady` after it**: added.
- **Plan row claims the web test is byte-identical**: corrected to name the added stopped-outbox timer-pair case and the pair rule.

D27 re-stacks on the new head.

obra:
## Round 2 fixes (`01b3fb01c`)

Refreshed onto `b8307aeee`. Both Mediums close as one change, since they are the
same defect asked twice.

**MEDIUM 1 + MEDIUM 2 — the gate ran when work was QUEUED, not when it RAN (`1a8652506`).**
RED first, three cases (Low 1's list plus the readiness one):

```
× a scan queued before stop() discovers nothing once the queue drains
    expected [ 'startup', 'interval' ] to deeply equal [ 'startup' ]
× connectionReady() after stop() discovers nothing
    expected [ 'ready' ] to deeply equal []
× a scan queued while ready discovers nothing if readiness is lost first
    expected [ 'startup', 'interval' ] to deeply equal [ 'startup' ]
```

Closed in one place rather than four ad-hoc checks:

```ts
#mayDiscover(reason: MutationDiscoveryReason): boolean {
  if (reason === "enqueue") return true;
  return this.#started && (reason === "startup" || this.#isReady());
}
```

asked at every entry (`#scheduleReadyScan`, `#scheduleDiscovery`,
`connectionReady`) **and again at execution time** inside the queued discovery —
before `listTargetRefs` and again immediately before `onDiscover`, because the
storage read itself can take long enough for the outbox to stop. The two
exceptions are the documented ones and are in the comment: an `enqueue` announces
its own durable commit even as the outbox stops (`enqueueIntent` owns the
readiness decision there), and the startup scan runs whether or not the host is
ready, so an app that opens offline still learns what is waiting.
`#handleBroadcast`'s own readiness check is gone — the predicate covers it.

One consequence worth naming: `stop()` still awaits discovery that is already
**running**, but no longer executes work that was only queued. Four existing tests
used `await stop()` as a queue flush; they now wait for the scan they actually mean
(`discovered(() => …)`, a `vi.waitFor` helper with that reason in its name) and
then stop, which is what they were asserting all along.

**LOW 1 — the cases are the three REDs above**, in
`cmd/evener-hub/frontend/src/stores/mutationOutbox.test.ts`: timer work queued
before `stop()` and draining after it, `connectionReady()` after `stop()`, and a
queued scan whose readiness is lost while the queue is blocked. The web suite is
31 tests, all green.

**LOW 2 — the plan row no longer claims the web test is byte-identical (`252798a91`).**
It says the test moves unchanged except for the cases round 2 added (the
stopped-outbox trio and the timer pair), and names the pair rule as part of the
option contract: `setInterval` and `clearInterval` are taken together or not at
all, so nothing is scheduled that cannot be cancelled, with `#mayDiscover` as the
one execution-time predicate and its two documented exceptions.

Checks (the current local list): `npx vitest run src/stores/mutationOutbox.test.ts`
31 passed; `npm run typecheck` clean; `npm run lint` clean (1340 files);
`make lint-package-imports` PASS; `make test-api-package` qualification clean
(and its index falsification still fails without the barrel's `MutationOutbox`
export). CI owns the matrix.

Diff vs `origin/main`: 9 files, 591 insertions, 204 deletions.

obra:
## Review disposition (01b3fb01c)

CI 15/15 green, RoboRev clean, simplify pass done (round 1). Merging. D27 re-merges onto main and opens next.

Issue #1662, opened 2026-09-17
web-skillguard: clipboard paste into inline skill chips times out on main


obra:
Measured while gating an unrelated PR (D24 web-lift, transcript projector relocation — pure package move, touches no composer/clipboard code).

\`make test-web-browser\` fails deterministically on \`web-skillguard\`:

\`\`\`
FAIL  web-skillguard (exit 1)
--- FAIL: TestSkillComposerBrowser (27.94s)
    skillguard FAIL: timed out after 15000ms (waited 15014ms) waiting for native clipboard preserves canonical display text
\`\`\`

The failing case is \`cmd/evener-hub/frontend/scripts/skillguard/run.mjs\`'s "native clipboard preserves canonical display text" step: select-all, copy, type a space, paste, then wait for the composer's text to equal the pasted canonical text. It times out waiting for that state.

**Confirmed pre-existing on a pristine \`origin/main\` checkout** (no D24 changes at all) — same failure, same 15000ms timeout, same case name, three times in a row (once under heavy host load, twice with the host quiet). Not a flake from host contention.

Likely correlated with the recently-merged inline skill-chip composer work: #1646 (inline-chip doc/comment/naming), #1648 (chip boundary separator, trailing colon, draft whitespace), #1649 (cursor anchored to selection start) — this test exercises exactly that surface (\`data-testid='composer-skill-chip'\`, clipboard copy/paste of two skill chips). Whether it's a real regression in paste-into-chip behavior or the test's own expectation going stale against an intentional behavior change, someone who owns that composer work should look — I did not investigate further since it's unrelated to what I was gating.

Artifacts from one failing run (screenshot, DOM dump, driver/helper logs) are under \`/var/folders/.../skillguard-browser-2750389245/artifacts\` on the host that reproduced it, not preserved anywhere durable — reproduce with:
\`\`\`
go test -tags browserguard ./cmd/evener-hub -run '^TestSkillComposerBrowser$' -count=1
\`\`\`

obra:
Measurement: main CI at f82b4122f (D28 merge, after the inline skill-chip composer #1646-#1649) completed green, browser guards included, and #1549 / #1663 report the guard red only locally. So this is local-only as of 2026-09-16 21:58 PDT (macOS, local Chrome), not a main regression. Reproduce with the CPU-throttle recipe from the browser-guard flake notes before changing the guard.

obra:
[...]

PR #1617, opened 2026-09-16
test(sdk): the navigation store contract suite runs in the package (#1588)

obra:
Closes #1588. Follow-up to D14-2 (#1584, merged as `5abcda20d`), which moved the ~900-line navigation store core into `appwire-client/typescript/state/navigation/store.ts` but left the 2758-line contract suite on the web side (`cmd/evener-hub/frontend/src/stores/navigation/store.test.ts`), driving the core only through the 29-line zustand adapter. This PR ports the suite's non-adapter-specific tests down into `state/navigation/store.test.ts`, run directly against `createNavigationStore`, and shrinks the web file to what's actually adapter-specific.

D14-3′ (#1609) and D15 (#1630), both prerequisites this branch was stacked on, are now merged to `main`; this branch has been rebased onto `main` at `9bcadfe27` (D15's merge commit) and its diff is the port alone.

## Measurement (test counts, `grep -cE '^\s*test\('` / `test\.each` per file, against `origin/main` before this port)

| File | origin/main | this branch |
| --- | --- | --- |
| `appwire-client/typescript/state/navigation/store.test.ts` | 5 blocks (5 tests) | 58 blocks (67 tests, `test.each` expanded) |
| `cmd/evener-hub/frontend/src/stores/navigation/store.test.ts` | 58 blocks | 5 blocks |

Matched by name: of the 56 named `test()` blocks on `main`'s web file, **51 moved** into the package suite unchanged in name and assertion (just re-pointed at a local `store` instance instead of the `navigationStore` singleton, and at `memoryNavigationPersistence` instead of a per-file `recordingPersistence` helper — D14-3′ already promoted that fixture into `testing/navigationPersistence.ts`). The 2 `test.each` blocks (`capability %s selects mode`, the `[section, pin catalog]` forced-page pair) also moved, byte-identical. 51 + 2 = 53 blocks moved; the package suite grew from 5 to 58 blocks (53 added), which matches exactly.

**5 tests stay web-side, unchanged, because they're adapter-specific:**
- `"setExpanded hydrates a raw v2 project key with representation version 2"`, `"rail expansion persists through store reset, overrides defaults, and hydrates once"`, `"a canonical persisted project-node key hydrates one v2 project root during boot"` — all three exercise `railExpansionPersistence`/the browser storage key, which only exists web-side.
- `"client replacement clears prior navigation ownership during bootstrap but preserves expansion"` — same reason; tests the singleton `navigationStore` module survives a client swap with the real persistence adapter wired.
[...]

PR #1680, opened 2026-09-17
refactor(sdk): pending-entries reconciliation at @evener/appwire-client/state/mutation (D25b)

obra:
D25b (SDK migration plan, `docs/superpowers/plans/2026-09-12-sdk-migration.md`): `panes/session/composer/queue/pendingReconcile.ts` (154, pure) moves to `state/mutation/pendingEntries.ts`, extending the D25a subpath (#1611) rather than adding one.

`reconcilePendingEntries(ref, outbox, model, submittedHere, isOwnMutationRecord?)` turns durable outbox/optimistic records plus a live `ThreadModel` into the `PendingTurnEntry` rows a composer's queue renders, identity-based so an authoritative projection replaces the same outbox entry rather than duplicating it.

- `BrowserPendingRecord` is renamed `PendingRecord` (it is no longer browser-specific); its two web-only comments ("another tab's") generalize to "another client's". Behavior is otherwise unchanged.
- `ThreadModel`/`InputItem`/`PendingMutation`/`canonicalSkillNames` are read from the package's own root modules (`../../model`, `../../types.gen`, `../../composerInput`) rather than the web's `@evener/appwire-client` package-name import, which a package module cannot use on itself.
- Adjusted for D25a round 1's identity factory: the package no longer exports a free `isOwnMutationRecord`, so `outboxEntry` takes it as a parameter, threaded through `reconcilePendingEntries`'s fifth argument, defaulted to `(record) => record.originClientId === undefined` — the safe unattributed-only answer a caller with no `ClientIdentity` instance handy still gets. `pendingTurnsStore.ts` passes its existing `isOwnMutationRecord` import at the call site; `pendingReconcile.test.ts`'s four-argument calls stay byte-identical under the default.
- The web file becomes a re-export — zero importers changed across `pendingTurnsStore.ts`, `PendingChips.tsx` and their tests — and `pendingReconcile.test.ts` (224) moves unchanged as the package's `pendingEntries.test.ts`, its two web-store-shaped imports repointed to the package's own `records.ts`.
- Falsified against the reachability assertion.

This branch also carries a merge of `origin/main` picking up D25a #1611 (squash-merged as `77d636835`), resolved per the stacked-merge duplication rule: every file D25a touched is now byte-identical to `origin/main`, and `git diff origin/main...HEAD --stat` shows only this PR's own 9 files.


obra:
## Simplify disposition (9ef068c13)

Two cleanup reviewers read the diff: the move is byte-identical apart from the injected ownership predicate, the layer and surface are right, efficiency clean. Applied on this PR before merge (in flight): the partial default for `isOwnMutationRecord` (a second encoding of the rule in `records.ts`, exercised by no real caller) goes and the parameter becomes required with a test for the "own record naming this client" branch; the plan's D25b row gets its PR number.

obra:
## Review round 1 disposition (9ef068c13)

CI 15/15 green, RoboRev clean. The two simplify fixes (required ownership predicate with its missing branch test; plan row PR number) are in flight; merge follows that head's CI and review.

obra:
## Simplify fixes (8840fda6f)

1. `state/mutation/pendingEntries.ts`: the fifth argument's default (`(record) => record.originClientId === undefined`) was a partial second encoding of `records.ts`'s real `isOwnMutationRecord` rule, missing the "or names this client" branch, and no real caller ever exercised it (the web threads the real predicate through `pendingTurnsStore.ts`; the qualifier smoke test passed an empty outbox). Made the parameter required, deleted the default, updated the qualifier smoke test to pass an explicit predicate (the real `identityA.isOwnMutationRecord`), and added a test through `reconcilePendingEntries` for an outbox record whose `originClientId` names this client (`pendingEntries.test.ts`'s new "an outbox record whose originClientId names this client is its own" case). Every other existing call site now passes an explicit `UNATTRIBUTED_ONLY` predicate matching the old default's behavior, since none of those fixtures set `originClientId`.
2. `docs/superpowers/plans/2026-09-12-sdk-migration.md`'s D25b row: PR column set to `#1680`; also updated the row's prose to drop the now-removed default and note the simplify fix.

Merged `origin/main` first (6db2ac700, #1617 + D24-web); `git diff origin/main...HEAD --stat` still shows only this PR's 9 files.

Gates: `make test-api-package` PASS; frontend `typecheck`/`test` (526 files, 12063 tests)/`lint` PASS; `make test-native` PASS; `make lint-package-imports` PASS; `EVENER_GITLEAKS_REQUIRED=1 make secret-scan` PASS; root `go test -short -count=1 .` PASS; `make test-web-browser` 6/7 guards PASS, `web-skillguard` still red per #1662 (reported, not chased).

obra:
## Review disposition (8840fda6f)

CI 15/15 green, RoboRev clean, simplify pass done. Merging. D25c re-merges onto main and opens next.

Issue #1676, opened 2026-09-17
web typecheck: the make gate and a direct tsc -p tsconfig.json disagree on skillDocument.test.ts


obra:
Residual from D11 #1502 round 12 (2026-09-16).

`npx tsc --noEmit -p tsconfig.json` run from `cmd/evener-hub/frontend` reports errors (in `skillDocument.test.ts`, on main) that `make test-web`'s typecheck step does not. Either the gate's typecheck is scoped more narrowly than the project tsconfig (so a test file can be type-broken on main without the gate noticing) or the ad-hoc invocation is stricter than intended. Find which, make the two agree (one tsconfig the gate and the editor both use), and note the answer in AGENTS.md next to the web gate list. Small; the current discrepancy is the reproduction.

obra:
Resolved by #1682: no scope gap between the gate and tsc -p tsconfig.json (measured byte-identical file lists); the difference was the gate's install preflight. AGENTS.md documents the gate command and the symlink caveat.

PR #1682, opened 2026-09-17
test(web): the typecheck gate and tsc -p tsconfig.json cover the same files (#1676)

obra:
Investigates #1676: `npx tsc --noEmit -p tsconfig.json` from `cmd/evener-hub/frontend` was reported to find errors in `skillDocument.test.ts` on `main` that `make test-web`'s typecheck step does not.

## Measurement

**The gate's recipe:** `make test-web` (`make/testing.mk`) depends on `web-preflight` and runs `scripts/web/test-web.sh`, which runs `npm run typecheck` (`cmd/evener-hub/frontend/package.json`) as one of three parallel checks (typecheck/test/lint). `npm run typecheck` is `tsc --noEmit --incremental false`, invoked with no `-p` flag from `cmd/evener-hub/frontend`, so it resolves that directory's `tsconfig.json` by TypeScript's normal directory search.

**`tsconfig.json`'s scope:** `"include": ["src", "../../../appwire-client/typescript"]` — both the frontend's own source tree and the AppWire package tree, in one program (the file's own comment explains why: the package's test files aren't imported by anything else, so without this entry a type error in one would be invisible to every gate).

**Diffed the two invocations' file lists** (`tsc ... --listFilesOnly`) on a fresh worktree at current `main` (`npm ci`'d): `npm run typecheck` and a bare `npx tsc --noEmit -p tsconfig.json` from the same directory both cover **exactly 1441 files, byte-identical** — re-verified after merging main forward once (still 1441, still identical). There is no scope gap: `-p tsconfig.json` names the exact same config the gate resolves implicitly, and I could not reproduce any error from either invocation with the install fresh.

**Where they actually differ:** `npm run typecheck` passes `--incremental false`, so it never reads or trusts a cached `.tsbuildinfo` (gitignored, per-checkout). A bare `tsc -p tsconfig.json` respects the tsconfig's own `"incremental": true` and will use whatever `.tsbuildinfo` happens to be sitting in the working copy. More importantly, `make test-web` runs behind `web-preflight.sh`, which `npm ci`s the frontend install whenever `package-lock.json` is newer than `node_modules` (a merge that changes dependencies, followed by no reinstall, is exactly this). A bare `tsc -p tsconfig.json` has no such preflight: run against a stale install, it can report real-looking errors (unresolved modules, missing types) that are an artifact of the stale `node_modules`, not of the code — which fits `skillDocument.test.ts` (part of the recently-landed inline-composer-skills feature, which added new dependencies to `package.json`) failing under a direct, unprepared invocation while the self-healing gate never sees it.

**Conclusion:** the direct invocation was the wrong entry point, not the gate. The two commands check the identical 1441-file program; the gate is not scoped more narrowly than the project tsconfig.

## Change

Documented the gate's actual command and its `web-preflight.sh` dependency in `AGENTS.md`, next to the existing "Frontend gates" section (which already documents Biome's exact invocation for the same reason — so someone reproducing a gate by hand doesn't invent an out-of-scope command). No test or gate behavior changes; this is documentation only, per the smaller of the two branches in the task ("if the direct invocation is the wrong entry point, document the gate's command in AGENTS.md... and stop").

## Gates (all foreground, on the merged head `14c94124f`)

- `make lint` — all 9 sub-gates PASS, including `secret-scan`
- root `go test -short -count=1 .` — `ok 20.233s` (run because the merge brought in a `make/testing.mk` change from main; the Makefile audits it covers stay green)

## Residuals

None.


obra:
## Review round 2 disposition (14c94124f)

CI 15/15 green. RoboRev: 1 Medium, 1 Low, both accepted.

- **The `npm ci` advice omits the preflight's symlink guard**: the paragraph prefers `make test-web` and, for a by-hand install, requires that `node_modules` is not a symlink before `npm ci`, pointing at `web-preflight.sh` as the reference.
- **Hardcoded file count**: removed from the durable text.

obra:
## Round 2 fixes (e2f18d7e3)

**Medium** (`AGENTS.md:60`): the earlier advice to run `npm ci` in `cmd/evener-hub/frontend` by hand omitted `web-preflight.sh`'s symlink guard. Read it: an agent worktree's `node_modules` is often a symlink to a shared install other worktrees use, and `npm ci` deletes the existing `node_modules` before installing — through a symlink that deletes the shared install out from under everyone else. Rewrote the paragraph to prefer `make test-web` for a manual reproduction, and if running `npm ci` by hand instead, to check `[ -L node_modules ]` first and read `web-preflight.sh` (which guards exactly this, comparing the symlink target's own `package-lock.json` instead of running `npm ci` through it). Said the same for the other two frontend installs, measured accurately rather than assumed: `appwire-client/typescript` has no preflight script at all (`make test-api-package` runs `npm run qualification` directly, no install check), and `mobile-native`'s `native-preflight.sh` checks the install's freshness and health but never runs `npm ci` itself — it fails loudly and names the command instead, precisely to avoid this. The symlink risk is real for all three; only `web-preflight.sh` self-heals via `npm ci` (with the guard), the other two just check and refuse.

**Low** (`AGENTS.md:51`): dropped the hardcoded `1441 files` count. The paragraph now says the two invocations resolve the same file set and points at `tsc --listFilesOnly` to re-verify rather than trusting a number that will drift, with a pointer to #1676 for the point-in-time measurement.

## Gates (all foreground, on the merged head `e2f18d7e3`; main advanced three times mid-round, each merge re-gated before the next fetch)

- `make lint` — all 9 sub-gates PASS, including `secret-scan`
- root `go test -short -count=1 .` — `ok` (run each time because the branch kept picking up `make/*.mk`-touching merges from main; the Makefile audits it covers stayed green throughout)

No `npm ci` was run by hand at any point this round — every install went through `make test-web`'s own preflight, per the finding this round documents.

obra:
## Review round 3 disposition (e2f18d7e3)

CI: `lint-repository` and `static` are red; the lane reads both logs and fixes what they name. RoboRev: 1 Low, accepted: the paragraph overstates the preflight's symlink guard because the newer-than check can short-circuit before the symlink branch runs; the doc states exactly when that branch runs, measured from the script's order of checks.

obra:
## Round 3 fixes (45e6dca2d)

**CI failures investigated, not content bugs.** Read both job logs:
- `lint-repository` failed inside its "Install gitleaks" step, at `make tools-gitleaks`, before the actual repository scan ever ran: `go: downloading github.com/gitleaks/go-gitdiff v0.9.1` then `verifying module: ... reading https://sum.golang.org/tile/8/0/x120/604: stream error: stream ID 91; INTERNAL_ERROR; received from peer` — a transient network/HTTP2 error against the Go checksum database, unrelated to any file in this PR. The lane that actually lints content (`make lint-generated lint-naming lint-gofmt lint-internal lint-package-imports secret-scan`) never got to run.
- `static` failed only because it's an aggregator (`needs: [static-build, lint-golangci, lint-repository, fuzz-checks, api-package, native]`) requiring every listed job to succeed; its own log is just the `jq` check reporting `lint-repository: failure`, nothing else.

Ran the six targets `lint-repository`'s content-scan step runs, locally, on the current merged tree: all PASS. `lint-repository` is not itself a make target (defined directly in `.github/workflows/ci.yml`); `static` is a pure `needs:`-aggregator with nothing of its own to run locally. There is no docs/markdown lint or line-length rule in either job — measured, not assumed. A rerun should go green on its own; flagging as a residual since a network flake against sum.golang.org isn't something this PR's content can fix.

**RoboRev Low, accepted:** measured the preflight's actual branch order (`web-preflight.sh:38-58`) rather than trusting my round-2 phrasing. It's `if node_modules -nt package-lock.json: no-op` / `elif -L node_modules: compare lockfiles, refuse on mismatch` / `else: npm ci` — an if/elif/else chain, so when the `-nt` check is true (a shared install happens to be newer than this worktree's own lockfile — plausible for a symlinked install per the script's own comment on mtime unreliability), the elif branch that actually compares `package-lock.json` content never runs at all. The script no-ops and proceeds without validating the shared install matches. Rewrote the paragraph to say this precisely: `web-preflight.sh`'s comparison is conditional, not a guarantee, and the advice is now to check `[ -L node_modules ]` yourself before a manual `npm ci` rather than trusting the script to always catch a mismatch.

## Gates (targeted per the round's scope — the two failing CI jobs' own local equivalents; nothing else)

- `make lint-generated lint-naming lint-gofmt lint-internal lint-package-imports secret-scan` (what `lint-repository`'s content-scan step runs) — all PASS
- No local target exists for `static` itself (pure CI aggregator)

`git diff origin/main...HEAD --stat`: 1 file (`AGENTS.md`), 35 insertions.

obra:
## Review round 4 disposition (45e6dca2d)

CI 15/15 green. RoboRev: one Medium from one reviewer, which the other two judged already covered by the passage's own caveat: the absolute-sounding sentence is scoped to the bare-tsc stale-install case so nothing reads as a guarantee. One-sentence push; merge on the next green.

obra:
## Round 4 fixes (462a1ce03)

RoboRev Medium: "`make test-web` never hits this because its preflight repairs the install first" read as an unconditional guarantee, ahead of the later caveat that the `-nt` short-circuit can leave a symlinked shared install unrepaired. Scoped the sentence to the case it actually covers — a real, non-symlinked `node_modules` left behind by a `package-lock.json` change — and pointed forward to the symlinked-install caveat that follows, rather than deleting the claim (it's still true and useful for that case). One sentence changed, nothing else touched.

Local check: read the passage back (docs-only; CI runs the lints).

`git diff origin/main...HEAD --stat`: 1 file (`AGENTS.md`), 37 insertions.

obra:
## Review disposition (462a1ce03)

CI 15/15 green, RoboRev clean, docs-only (simplify by reading). Merging.

PR #1683, opened 2026-09-17
Model picker: no connect entry, listing diagnostics stop at their class

obra:
Two fixes from the model-selector bug report, plus a simplify-code cleanup pass.

## 1. The model picker offers no connect-provider entry (ab86b381e)

The shared ModelSwitchTrigger (the session model switch and the spawn pane's trigger) no longer renders "Connect another provider" or mounts the connect dialog; the `onConnectProvider` prop is gone. Provider setup happens where it has a real flow: the spawn pane's own pane-level notices, and the credentials pane. The post-connection handoff that reopens the picker scoped to the connected provider is unchanged and still covered.

## 2. Listing diagnostics stop at the failure's class (ee7a06309)

The picker prints a provider's listing diagnostic inline under the model list, so the message must stop at its class rather than quote the failure. An endpoint answering 404 with an HTML page used to put that page's content in the list; a missing credential quoted the registry's whole remediation warning (the GOOGLE_APPLICATION_CREDENTIALS sentence). Now:

- HTTP failures render as `HTTP {status}` (via the typed `llm.Error` status).
- Missing credentials render as `no credential`, via the new `llm.ErrNoCredential` sentinel — attached by `missingCredential` and, in the cleanup pass, by the gcp-adc lookup-failure path.
- Every other failure keeps its redacted text.

Per-model ⚠ warning lines (e.g. the regional Vertex note) are unchanged.

## 3. simplify-code pass (0a3497c72)

Four read-only reviewers (reuse, simplification, efficiency, altitude) over the diff; findings applied: the launchcheck fixtures decode the real `appwire.ModelListDiagnostic` type, share the package's existing `withLaunchCheckLoadClient` seam installer and a `gwDiagnostic` row helper, the gateway server stays inline at its one caller; the Spawn handoff fixture derives its provider card from its instance row; two stale comments corrected (the spawn pane no longer shares its dialog chunk with the model-switch trigger; the boundary's callers are the spawn pane and the credentials section).

## Verification

Every new assertion was proven red before its implementation. Green locally: `make test-web`, `make lint`, `make vet`, and `go test -short` across llm, cmdutil, the cmd/evener tree, hub, TUI, and tokenauth; the affected frontend files re-verified against origin/main after merging the advanced base (19 commits, clean merge).


obra:
Review response (RoboRev pass on 708c136):

**gcpadc sentinel breadth (Medium) — fixed in 7a31e2638.** The `ErrNoCredential` sentinel in `GCPADC.Apply` fired for every ADC lookup failure. It now rides only when the registry itself resolved no credential (`Source "none"`); a lookup failure for a credential the registry saw (`Source "adc"`) is a broken file, not an absent credential, and carries no sentinel, so the picker's "no credential" line cannot hide a credential that exists but will not load. `TestGCPADCDistinguishesABrokenADCFromAnAbsentOne` pins the distinction.

**usage-limit class (Low) — fixed in a61b8113c.** The picker line for an exhausted allowance had regressed to "HTTP 429" + "Provider error". `launchCheckDiagnosticMessage` now consults the typed error's category first (`llm.KindQuotaExceeded` → "usage limit reached"), which also restores the distinct "Usage limit reached" title through `FromFields`. The category is checked ahead of the status because exhausted quotas also arrive as 403 (billing-cycle exhaustion). An ordinary 429 keeps "HTTP 429" (pinned).

**codex not-signed-in (Medium) — deliberate, no change.** That branch's own message — "instance "X" is not signed in (run `evener openai login --instance X`)" — is already one compact, actionable line, not a raw dump: folding it into the "no credential" class would drop the login remedy the line exists to carry.

Local gates on a61b8113c: both packages' `go test` green (new tests proven red-first), `gofmt`/`go vet` clean, `golangci-lint 2.13.1` 0 issues on both packages.

obra:
Review response (RoboRev pass on a61b811):

**joined-branch status walk (Low) — fixed in ad39210ce.** `launchCheckHTTPStatus` walked the spine with `errors.Unwrap`, which cannot descend into an `errors.Join` node (`Unwrap() []error`), and `errors.As` stops at the first `llm.Error` in branch order — so a status behind a status-zero sibling in a joined chain fell through to the redacted raw text. The walk now visits every branch in chain order. `TestLaunchCheckDiagnosticFindsAStatusBehindAJoinedSibling` pins it (proven red first). Local gates: package tests green, gofmt/vet clean, golangci-lint 2.13.1 0 issues.

obra:
Review response (RoboRev pass on ad39210):

**doc reference to the removed chooser entry (Low) — fixed in a8a606a49.** `docs/connecting-a-provider.md` still said the session model chooser offers "Connect another provider"; this change removes exactly that affordance (`ModelSwitchTrigger.test.tsx` pins its absence). The sentence now states only the model-pick behavior; the paragraph's opening already points at Settings → Providers & credentials → Connect provider as the management route.

PR #1685, opened 2026-09-17
refactor(sdk): the pending-turns projection store lives in @evener/appwire-client/state/mutation (D25c)

obra:
## Summary

- Adds `createPendingTurnsStore` to `@evener/appwire-client/state/mutation`: the framework-free store that projects a thread's outbox/optimistic mutation records (via D25b's `reconcilePendingEntries`) plus submission bookkeeping into the pending-turn rows a composer's queue renders (`recordSubmittedHere`, `beginSubmission`/`endSubmission`, `settleSubmittedDraft`, `pendingTurnEntries`).
- Rewrites the web adapter `cmd/evener-hub/frontend/src/panes/session/composer/queue/pendingTurnsStore.ts` to build `threads`/`draft`/`identity` ports around the package store, keeping durable-read machinery (generation-fenced IndexedDB reads, settle-tracking test helpers) web-side while delegating the projection logic to the package.
- Follows D25c's plan row in `docs/superpowers/plans/2026-09-12-sdk-migration.md`; stacked on D25a (#1611) and D25b (#1680), both now released to main — this PR's diff against `origin/main` touches only D25c's own files (`pendingTurns.ts`, `pendingTurns.test.ts`, the barrel, qualifier manifest, README, tsconfig.build.json, the web adapter, the plan doc row); `pendingEntries.ts`, `records.ts`, `secureUUID.ts` and their tests all diff empty against `origin/main`.
- Threads an identity port (`Pick<ClientIdentity, "isOwnMutationRecord">`) through the store's deps, following D25b's release where `isOwnMutationRecord` became a required parameter of `reconcilePendingEntries` (no partial default) — the web adapter supplies its existing `mutationClientIdentity`-derived predicate.
- The store mutates the framework-free store object in place rather than spreading (`{...store, method}`), the pattern established in D28's `core.ts` to keep `vi.spyOn(store, "setState")` transparent to reactive wrappers.

## Test plan

- [x] `make test-web` (web-typecheck, web-test, web-lint)
- [x] `make test-api-package` (qualifier), including falsification: removed the `./state/mutation` barrel's pendingTurns exports, confirmed `shipped module unreachable from every published specifier: state/mutation/pendingTurns`, restored via `git diff | git apply -R`, reconfirmed green
- [x] `make test-native`
- [x] `make lint-package-imports`
- [x] `npx biome ci ../../../appwire-client/typescript` (from `cmd/evener-hub/frontend`)
- [x] `EVENER_GITLEAKS_REQUIRED=1 make secret-scan`
- [x] `go test -short -count=1 .` (root Makefile audits)
- [ ] `make test-web-browser` — web-skillguard failed twice with the same symptom ("timed out after 15000ms waiting for visible contenteditable"), observed previously as an environmental/host-contention flake unrelated to this PR's `state/mutation` changes (all other browser guards, including web-shellguard, passed both runs)


obra:
## Simplify disposition (d78f5fa85)

One cleanup reviewer read the diff across all four angles: the package/web seam is right (nothing host-bound in the package, the generation fencing and IndexedDB tracking stay on the web), the ports match the subpath's existing `create<X>Store(deps)` vocabulary, the core reuses `reconcilePendingEntries` and the identity's ownership check, and the new suite runs once. Applied on this PR before merge (in flight):
1. The plan's D25c row cites `#1685`.
2. The three cast-heavy `ThreadModel` fixtures use the typed builder `pendingEntries.test.ts` already has, moved into a shared test-support module.
3. The web hook subscribes to the store instead of a hand-maintained `useMemo` deps array, which removes the new `useExhaustiveDependencies` suppression.
Skipped, acknowledged in the plan: the second `readDraftRevision` per settle.

obra:
## Simplify fixes (6b0d2cbd8)

1. Plan doc: D25c row's PR column `—` → `#1685`.
2. `pendingEntries.test.ts`'s typed `model()` ThreadModel builder moves to a new shared `state/mutation/testing.ts` (`threadModel()`); `pendingTurns.test.ts` now imports it instead of building `{ turns: [], pendingMutations: [] } as unknown as ThreadModel` at three call sites. No casts remain.
3. `pendingTurnsStore.ts`'s `usePendingTurnEntries` no longer pulls outbox/optimistic/submittedHere/model through separate `useStore`/`useThreadsStore` selectors just to list them as trigger-only `useMemo` deps. It now subscribes directly to `pendingTurnsStore` and to `threadsStore`'s model-for-this-ref via `useSyncExternalStore`, with `getSnapshot` caching the last computed entries array (keyed on both stores' state references plus `ref`/`method`) so nothing changed still returns the same reference. The `biome-ignore lint/correctness/useExhaustiveDependencies` suppression is gone; `npm run lint` and typecheck are clean with no new suppression.

Skipped, plan-acknowledged: the second `readDraftRevision` per settle.

Gates run per the updated local gate list (CI builds the merge ref and runs the full matrix):
- `npx vitest run` on the touched suites: `state/mutation` package tests (4 files, 38 tests) and the web `pendingTurnsStore.test.ts` + `pendingTurnsStore.edge.test.ts` (28 tests) — all green
- `npx biome ci ../../../appwire-client/typescript` and `npm run lint` (frontend, covers `src` + the package) — clean, no fixes
- No package module was added to or moved in the published barrel by these fixes (`testing.ts` is test-only, not in `index.ts` or `tsconfig.build.json`), so no new falsification run

Residual observed, out of scope: running the full package suite (not just the touched files) turned up 4 pre-existing failures in `reconnect.test.ts` (heartbeat-timing assertions), unrelated to this branch — that file has no local diff and was last touched by #1241. Not fixing under the narrowed gate list.

obra:
## Review round 2 disposition (d78f5fa85)

CI 15/15 green. RoboRev: 4 Low, all accepted. The simplify push already on the PR needs its own CI and review, so the Lows ride that same round instead of a separate follow-up: empty `pendingTurnEntries` results coalesce to the shared constant again; the draft revision is read once and shared between the recovery commit flag and the clear decision; `endSubmission` is a true no-op for a ref that is not in flight; the identity-port comment states the predicate is required.

obra:
## Round 2 fixes (bd32019a0)

Four Lows from RoboRev's review of d78f5fa85, folded in on top of the simplify push (one commit each, failing test first where behavior changed):

1. `pendingTurnsStore.ts`'s plain (non-hook) `pendingTurnEntries()` now coalesces an empty result to the shared `NO_ENTRIES` constant, same referential stability the hook already had. New test: `pendingTurnsStore.test.ts` — "the plain pendingTurnEntries read returns the same empty-array reference when nothing is pending" (confirmed red against the old fresh-`[]`-every-call behavior, green after).
2. `settleSubmittedDraft` now returns `{ cleared, draftUnchanged }` instead of a bare `cleared` boolean; `draftUnchanged` is the revision-only half of its check, already computed internally. `submitWithPendingTracking` takes `draftUnchanged` from that one call instead of reading `readDraftRevision` a second time for the recovery-commit flag — one read, so the two answers can't disagree. Updated all four `settleSubmittedDraft` tests in `pendingTurns.test.ts` to the new shape (confirmed red for the new `draftUnchanged` assertions against the old return value, green after); two renamed to state they're proving the revision-only half stays unchanged even when text/skills differ.
3. `pendingTurns.ts`'s `endSubmission()` now early-returns when `submittingRefs` doesn't have `ref`, mirroring `recordSubmittedHere`'s no-op guard, so a safe-to-call release for a ref that was never marked doesn't publish a new state object. New test: "is a no-op, safe to call, for a ref with no submission in flight" (confirmed red — state object changed even though nothing did — green after the guard).
4. `pendingTurns.ts:53`'s `PendingTurnsStoreDeps.identity` comment no longer says `reconcilePendingEntries` has its own default; it now says the `isOwnMutationRecord` parameter is required, with no partial default (comment-only, no test).

Targeted checks (per the updated local gate list):
- `npx vitest run` on `state/mutation` (4 files, 39 tests) and the web `pendingTurnsStore.test.ts` + `.edge.test.ts` (29 tests) — all green
- `npm run lint` (frontend, covers `src` + the package) — clean, no fixes
- No package module added or moved by these fixes, so no new falsification run

`reconnect.test.ts`'s pre-existing heartbeat-timing failures left alone per your note (local-only flake, CI green on main).

obra:
## Review round 3 disposition (bd32019a0)

CI 15/15 green. RoboRev: 1 Medium, 2 Low, all accepted.

- **Provenance wiring untested** (`recordSubmittedHere` → `submittedHere` → `fromThisClient` after the durable record is gone): the behavior test is added and falsified by dropping the `submittedHere` input.
- **Snapshot cache keyed on the whole store state re-renders on writes the entries do not read**: the key narrows to the inputs `pendingTurnEntries` reads, pinned by a test that a submission-flag write returns the same array.
- **Plan row describes the pre-round-2 design**: corrected to the present (single revision read returned from `settleSubmittedDraft`, the added web case, the real test count).

obra:
## Round 3 fixes (1ad232ef5)

One Medium and two Lows from RoboRev's review of bd32019a0, all accepted:

**Medium** — the store's central invariant was untested: `recordSubmittedHere`'s discovery is what lets `pendingTurnEntries` still answer `fromThisClient: true` for a mutation the authoritative projection now describes, once the durable outbox/optimistic record behind it is gone. New test in `pendingTurns.test.ts`: seeds `submittedHere` via `recordSubmittedHere` with an own outbox record, leaves the store's own `outbox`/`optimistic` empty (the durable record already settled away), supplies a thread model whose `pendingMutations` carries the matching id, asserts `fromThisClient: true`. Falsified by dropping `submittedHere` from `pendingTurnEntries`'s internal `reconcilePendingEntries` call (confirmed red — `fromThisClient` flipped to `false` — restored).

**Low 2** (`pendingTurnsStore.ts`, `usePendingTurnEntries`'s `getSnapshot`) — the cache keyed on the whole `pendingTurnsStore` state object, so any write at all (including `beginSubmission`/`endSubmission`'s `submittingRefs` and the recovery map, neither of which `pendingTurnEntries` reads) invalidated it. Narrowed the key to `outbox`, `optimistic`, `submittedHere`, the thread model, and `ref`/`method` — the exact inputs `pendingTurnEntries` reads. New test: a `submitWithPendingTracking` call with a no-op `perform` (so only `beginSubmission`/`endSubmission` fire, no outbox/optimistic write) leaves `usePendingTurnEntries`'s returned array `===` the pre-write reference. Falsified by reverting to the whole-state key (confirmed red, restored).

**Low 1** (plan doc, D25c row) — updated to the present shape: `settleSubmittedDraft` returns `{ cleared, draftUnchanged }` from its one `readDraftRevision` read (not two separate reads); `pendingTurnsStore.test.ts` gained round 2/3 cases and is no longer byte-identical (`pendingTurnsStore.edge.test.ts` still is); package suite is at 14 tests.

One extra commit: `npm run lint` caught a line-wrap the Medium/Low2 test additions needed (biome's own formatting preference, no logic change) — applied via `biome check --write`.

Targeted checks:
- `npx vitest run` on `state/mutation` (4 files, 40 tests) and the web `pendingTurnsStore.test.ts` + `.edge.test.ts` (30 tests) — all green
- `npm run lint` (frontend, covers `src` + the package) — clean, no fixes

obra:
## Review disposition (1b8298a99)

CI 15/15 green, RoboRev clean, simplify pass done (rounds 1 to 3). Merging.

PR #1694, opened 2026-09-17
refactor(sdk): the mutation dispatcher lives in @evener/appwire-client/state/mutation (D27)

obra:
Plan row **D27** (`docs/superpowers/plans/2026-09-12-sdk-migration.md`), stacked on **#1678** (D26, merged as `41e40ce55`) and sharing the subpath with **#1680** (D25b's `reconcilePendingEntries`).

## What moves

`cmd/evener-hub/frontend/src/stores/mutationDispatcher.ts` (292 lines) moves whole to `appwire-client/typescript/state/mutation/dispatcher.ts`, over the **`MutationOutboxStorage` port D26 already defined** — the 11 dispatcher methods that port declared are exactly what this class calls, so nothing new is added to the contract.

What the class is, unchanged by the move: one attempt at a time per target ref, a receipt reconciled into storage before the next attempt, a refusal turned into a recovery record carrying the daemon's own reason, and an outcome nobody can vouch for left `blockedUnknown` rather than replayed — the no-blind-replay rule from #1116's standing decisions. Every reaction to an outcome (a blocked mutation, a clear's response, a shared note's authority) stays a callback the app supplies, so no React binding, no store and no DOM type enters the package.

The web module becomes a six-line re-export under the same name, so no importer moved; `threads.ts`'s only change is binding the dispatcher to the attachment type that carries bytes (`MutationDispatcher<MutationAttachment>`), as the outbox already is.

## Also here

Two scenario cards pointed at the web module for a symbol it no longer declares — `TestScenarioSourceSymbolsAreDeclared` (root `go test -short .`) fails on the anchor:

- `test/scenarios/web-thread-clear-replaces-session.md` — `…/stores/mutationDispatcher.ts#MutationDispatcher` → `appwire-client/typescript/state/mutation/dispatcher.ts#MutationDispatcher`, noting the web imports it under its own name;
- `test/scenarios/ask-two-clients.md` — the sibling line citation `stores/mutationDispatcher.ts:104-108` → `…/state/mutation/dispatcher.ts:213-223`, the `blockedUnknown` + `onBlockedMutation` block it means.

## Checks

`npx vitest run src/stores/mutationOutbox.test.ts src/stores/mutationDispatcher.test.ts` 58 passed (the dispatcher suite is the oracle and moved unchanged, imports included); `npm run typecheck` clean; `npm run lint` clean (1341 files); `make lint-package-imports` PASS; `make test-api-package` qualification clean, and its **index falsification** still bites — dropping `export { MutationDispatcher }` from the barrel fails it with `TS2304: Cannot find name 'MutationDispatcher'`; root `go test -short -count=1 .` ok (the scenario-anchor audit above). CI owns the full matrix.

Every D26/D25b-touched file (`state/mutation/outbox.ts`, `records.ts`, `pendingEntries.ts`, the web outbox shim and its suite) diffs **empty** against `origin/main`.


obra:
## Simplify disposition (d7b3ae23b)

One opus reviewer read the diff across the four angles: the moved class differs from the web original only in imports, the attachment type parameter and the storage port type; purity, surface, efficiency and the class anchor are clean. Applied on this PR before merge (in flight):
1. The shim re-exports only `MutationDispatcher` (the options type had no consumer).
2. `onHumanNoteResponse`, a second encoding of `prepareHumanNoteResponse` with only a test caller, goes.
3. The qualifier's dispatcher smoke reuses the memory outbox port and says what it proves.
4. The plan row cites `#1694` and both web suites.
5. The README no longer claims no scheduling policy lives here.
6. The ask-two-clients scenario citation points at the rejection block.
Filed rather than widened here: #1695 (one readiness notion for the outbox and the dispatcher).

obra:
## Review round 1 disposition (d7b3ae23b)

CI 15/15 green, RoboRev clean. The six simplify fixes are in flight; merge follows that head's CI and review.

obra:
## Simplify fixes (1c355152a)

1. `cmd/evener-hub/frontend/src/stores/mutationDispatcher.ts`: dropped the `MutationDispatcherOptions` re-export — no web importer names it; the shim now re-exports only `MutationDispatcher`.
2. `state/mutation/dispatcher.ts`: dropped `onHumanNoteResponse` and its defaulted delegation — a second encoding of `prepareHumanNoteResponse`, whose only real caller (the web's own dispatcher construction in `threads.ts`) already passes `prepareHumanNoteResponse`. Adjusted `mutationDispatcher.test.ts:140`'s one use to `prepareHumanNoteResponse`, wrapping the same assertion mock so the test still asserts the same `(record, response)` call shape. Verified the test passes with the adjustment before removing the option, then again after.
3. `scripts/qualify-package.mjs`: the dispatcher smoke built a third inert storage literal `nextDispatchable` never consulted meaningfully (a real one already stops it) and its comment claimed "the same memory port" while building a different one. Hoisted `memoryOutbox`'s storage port into a `memoryOutboxStorage` const both `MutationOutbox` and `MutationDispatcher` now share; the comment says what the block actually proves — the dispatcher is a runtime export whose `dispatchTargets` resolves once `nextDispatchable`'s inert `undefined` stops it before any transport attempt.
4. Plan doc D27 row: PR column `—` → `#1694`; named both web suites (`mutationDispatcher.test.ts` and the previously-uncited `mutationDispatcher.edge.test.ts`) as oracles, and noted this round's fixes plus the one test they touched (`mutationDispatcher.test.ts` is no longer byte-identical; `.edge.test.ts` still is).
5. `appwire-client/typescript/README.md`: dropped the false "no... scheduling policy" half of "No storage adapter or scheduling policy lives here either" — `MutationDispatcher`'s serialized per-ref dispatch is exactly a scheduling policy, and it lives in this subpath now. Kept the storage-adapter half, still true.
6. `test/scenarios/ask-two-clients.md:25`: repointed the rejection-path citation from the `blockedUnknown` block to `dispatcher.ts:202-206` (`notAccepted` → `transferToRecovery(..., "rejected", ...)`), the block the card actually describes. Root `go test -short -count=1 .` (scenario citation audit) green.

Targeted checks:
- `npx vitest run` on `mutationDispatcher.test.ts` + `.edge.test.ts` (38 tests) — green
- `npm run lint` (frontend, covers `src` + the package) — clean, no fixes
- `make test-api-package` — qualified (fix 3 touches the qualifier's smoke code)
- root `go test -short -count=1 .` — green

obra:
## Review round disposition (1f8dc28e6)

RoboRev's Medium (`dispatcher.ts:169-174`, the post-drain queue-cleanup heuristic) measured **can happen**, cross-tab. Evidence:

- **(a) intentSequence allocation**: `mutationOutboxIndexedDB.ts:584-590` (`#allocateSequence`) — a real IndexedDB read-modify-write transaction on a `SEQUENCE_STORE` record keyed by `targetRef`, not an in-memory per-dispatcher counter. Globally monotonic per target ref across every tab of the origin, but it orders intent **creation**, not server acceptance.
- **(b) no cross-tab exclusivity**: `mutationOutboxIndexedDB.ts:82-83`, the class's own header comment — "Cross-tab correctness comes from IndexedDB readwrite transaction ordering rather than a browser lease or leader election." Each dispatcher instance's per-target FIFO is in-memory, per-instance only.
- **(c) shared visibility**: `listOptimistic(targetRef)` (`mutationOutboxIndexedDB.ts:165-173`) reads the whole shared, origin-wide IndexedDB with no per-tab scoping.

Given (a)-(c): a lower-sequence `turn/queue` intent (created first, tab A) can be accepted by the server *after* a higher-sequence drain (tab B) was already applied — genuinely fresh work on the emptied queue — and the old rule's "lower sequence ⇒ accepted-before-the-drain-was-sent" inference, true only within one instance's own FIFO, retires it anyway. Real data loss shape: a message the server accepted disappears from client tracking.

Per your framing, this doesn't get fixed inside #1694 (a pure relocation). Filed and landed as a separate follow-up against `main`, in the web file where this class lives today:

- **Issue**: #1704 (full measurement, evidence, and root cause)
- **Follow-up PR**: #1705 — removes the sequence-based heuristic; adds `retireConsumedQueueIntents(targetRef, authoritativeQueueIds)`, which retires an optimistic `turn/queue` record only when absent from a caller-supplied authoritative snapshot (never compares local sequence numbers); wires `thread/queueChanged`'s own `clientMutationIds` (the server's current queue, confirmed always emitted after a queue-revision change server-side) into it, matching the `reconcileIdentities`/`restoreProvenAbsent` pattern already used in `threads.ts`. TDD throughout (two-tab counter-example + end-to-end wiring test, both falsified against the old/disabled code and restored). `npx vitest run` (352 tests), typecheck, lint, `lint-package-imports` all green; MERGEABLE.

#1694 itself is unaffected by #1705 (different file on different branches) - ready to merge on your side.

Issue #1526, opened 2026-09-16
mobile-native has no component or hook test harness; ProvidersScreen mount order went untested


obra:
Surfaced by RoboRev on D9 #1524: `ProvidersScreen` built its credential store during render and bound the connection only in a passive effect, while the child `Providers`' mount effect called `model.start()`; React flushes child effects first, so the first read ran before any client was set and the screen settled on the empty state. The fix binds the store at construction (`createNativeCredentialStore(client, state)`, pinned by `mobile-native/src/credentialStore.test.ts`), but the effect-ordering property itself is untested.

`mobile-native` runs vitest in a node environment with no `@testing-library/react-native`, no `react-test-renderer`, and no `.test.tsx` files, so nothing can mount a screen or drive a hook. Decide whether to add a render harness (testing-library/react-native or react-test-renderer under vitest, with the Expo/native modules mocked) and, if so, add a mounting test for `ProvidersScreen` that renders it with a ready client and asserts the listing read is issued and published, plus a `renderHook` test for `useCredentialStore`'s two-effect wiring.

obra:
Two more instances, 2026-09-17: #1696 fixed a confirm-time re-read of the plugin mutation gate inside MarketplaceBrowser.tsx's Alert callback with no test, because nothing can mount the screen; D6 #1549 had to put its editing gate in a separate transcriptPreferencesEditorGates.ts module because TranscriptPreferencesEditor.tsx imports react-native and cannot be imported from a test at all. The gate-module extraction is a workable pattern for pure predicates, but callback-time re-reads inside Alert handlers stay untested until this harness exists.

PR #1502, opened 2026-09-16
refactor(sdk): plugins store core at @evener/appwire-client/state/extensions (D11)

obra:
Row D11 of docs/superpowers/plans/2026-09-12-sdk-migration.md: the installed-plugins slice of the web extensions store becomes the second store under the `@evener/appwire-client/state/extensions` subpath D10 (#1476) opened. D12 (directories/MCP) follows.

## Core: `appwire-client/typescript/state/extensions/plugins.ts`

`createPluginsStore(client: PluginsClient)` with `PluginsClient = Pick<AppwireClient, "request" | "onNotification">`, returning the D30 triple (`getState`/`getInitialState`/`setState`/`subscribe`) plus `start`/`reset`/`dispose`. State holds the actions the web contract already has: `plugins`, `pluginsLoading`, `pluginsError`, `pluginRevision`, `fetchPlugins`, `installPlugin`, `upgradePlugin`, `removePlugin`, `enablePlugin`, `disablePlugin`, `setPluginAutoUpgrade(plugin, marketplace, autoUpgrade)`. `PLUGIN_REFETCH_DEBOUNCE_MS = 250`.

- Fetch never throws (loading/error in state); mutations reject, exactly as D10 reconciled it.
- Every list response is fenced by a revision: a fetch outrun by a newer fetch or mutation writes none of list, loading flag or error; a mutation's response lands only if nothing newer committed. The fence D10 kept inline in marketplaces.ts is now `listRevision.ts` (`createListRevision()`), shared by both stores.
- `start()` follows `evener/plugin/updated`: `pluginRevision` moves synchronously, the list refetches after the debounce. Every post-await write is fenced by `dispose()`.
- Barrel, `tsconfig.build.json` `files`, the `./state/extensions` qualification manifest (type uses for both modules; the smoke chain now fetches, asserts the recorded rejection, asserts a mutation rejects and disposes; synchronous assertions on the constant, initial revision and a fenced `commit`) and the README subpath section are updated. Falsified: deleting the barrel re-export fails `make test-api-package` with `shipped module unreachable from every published specifier: state/extensions/plugins`.
- Isolation test with two instances over `FakeClient`, plus the native oracle's behaviours (identity of `{plugin, marketplace}` through every mutation, stale-read fencing, dispose isolation) as package cases.

## Web: `cmd/evener-hub/frontend/src/stores/extensions.ts`

[...]

PR #1696, opened 2026-09-17
fix(native): marketplace actions wait for an in-flight plugin write (#1502 follow-up)

obra:
## Summary
- `mobile-native/src/MarketplaceBrowser.tsx`'s Refresh source, Remove marketplace and Add marketplace actions disabled only on this view's own `mutating` flag, so they stayed clickable while a plugin install from the same marketplace was in flight (`pluginBusy`) - the hub does not serialize a marketplace removal/refresh against a concurrent install.
- They now also disable on `pluginBusy` (`disabled={mutating || pluginBusy}`, and the same OR in `remove()`'s own guard). The screen's plugin-mutation gate stays exclusive to plugin writes, exactly as documented (#1502) - these three actions wait on its busy flag without taking the gate themselves.

Filed as a Low from #1502's review round 15 (`mobile-native/src/MarketplaceBrowser.tsx:187, 194, 232`).

No test added: there is no render harness for native JSX (#1526), and there is no non-trivial derivation to pin at a lower level - the fix is a one-line boolean OR at three call sites plus the matching guard in `remove()`, so there is nothing to extract that would make it independently testable.

## Test plan
- [x] `mobile-native`: `npx tsc --noEmit --project tsconfig.check.json` - clean
- [x] `mobile-native`: `npx vitest run src/pluginMutationGate.test.ts` - 4/4 pass (gate semantics unchanged; this fix is UI-only)
- [ ] CI (Metro bundle, full native gate)


obra:
## Review round 2 disposition (1df0950af)

Also folded in the coordinator's /simplify pass: hoisted `mutating || pluginBusy` (six sites, including the two that predate this PR) into one `const busy = mutating || pluginBusy;` right after the `mutating` state.

**Medium** (`MarketplaceBrowser.tsx:142`, confirm-while-install-starts) - accepted. `remove()`'s Alert `onPress` re-checked only the `revision` snapshot, not whether a plugin install had started while the confirmation dialog was up. `busy` is a render-time value captured when the button that opened the dialog was still enabled; the confirm tap can land much later. Fixed by reading `gate.isBusy()` live at confirm time (mirroring the existing fresh `revision.current` comparison) and refusing with the same `PLUGIN_MUTATION_BUSY` copy `install()` already shows on a refusal.

**Low** (`MarketplaceBrowser.tsx:190`, refresh has no imperative guard) - accepted. Extracted a `refresh()` function with the same `if (!marketplace || busy) return` early return `remove()` already has, and pointed the button's `onPress` at it instead of an inline closure.

**Low** (`MarketplaceBrowser.tsx:62-68`, comment says the hub doesn't serialize) - **reviewer is right, comment was wrong.** Measured before rewriting:
- `internal/plugins/marketplaces.go:248` (`RemoveMarketplace`) and `:882` (`RefreshMarketplace`) both call `m.lockStore(ctx, marketplaceAcquireLock, 30*time.Second)`.
- `internal/plugins/install.go:119`/`195`/`287` (`Install`/`Upgrade`/`mutateEntry`, which backs `SetEnabled`/`SetAutoUpgrade`) call `m.lockStore(ctx, installAcquireLock, 30*time.Second)`.
- `lockStore` (`internal/plugins/locks.go:72`) resolves to `m.acquireStoreLock(ctx, acquire, m.lockPath(), timeout)` - the lock **path** is `m.lockPath()` regardless of which `acquire` variable is passed; `marketplaceAcquireLock` and `installAcquireLock` are both just `= acquireLock` (separate package vars for test-seam mocking, not separate locks). So install and marketplace-remove/refresh do serialize under the identical store lock.

Given that, the gate isn't a correctness mechanism here either (same as `pluginMutationGate.ts`'s own doc comment already says about the two UI surfaces) - it's UX. What the shared lock does NOT do is refuse: a removal issued mid-install blocks at the hub until the install's lock releases. `acquireLock` retries with backoff up to its own 30s timeout, and the client's own default request timeout (`client.ts`'s `DEFAULT_REQUEST_TIMEOUT_MS = 30_000`) is the same order of magnitude - so without this gate, the button would sit looking hung for up to ~30s, or occasionally time out outright with a generic "could not confirm" error, instead of refusing at once with copy that says why. Rewrote the comment to say this.

No test added for any of the three: no render harness for native JSX (#1526), no RN Testing Library setup anywhere in this repo (checked - no `*.test.tsx`, no `testing-library` import), and neither fix has a non-trivial pure derivation to extract - both are a live read (`gate.isBusy()`) plus an early return wired straight into JSX event handlers. Said so in the commit message too.

### Gates
- `mobile-native`: `npx tsc --noEmit --project tsconfig.check.json` - clean
- `mobile-native`: `npx vitest run src/pluginMutationGate.test.ts` - 4/4 pass (gate semantics unchanged)
- No biome over `mobile-native` (no config there, per repo convention)

obra:
## Review round 3 disposition (fdaa83d9c)

Closed the class instead of patching `refresh()` and `AddMarketplace.submit()` as two more instances of the same asymmetry. Added `dispatchMutation()`, the one path every marketplace write now dispatches through - it refuses live at dispatch time (a ref for this view's own write, `gate.isBusy()` for the plugin gate), mirroring how `gate.run()` already guards installs. `act()` (used by `refresh`/`remove`) and `AddMarketplace.submit()` (via a new `dispatch` prop) both call it instead of each carrying their own guard.

| Dispatch site | Guard function | Reads gate live? |
|---|---|---|
| `refresh()` | `act()` → `dispatchMutation()` | yes (`gate.isBusy()`) |
| `remove()`'s confirm `onPress` | `act()` → `dispatchMutation()` | yes |
| `AddMarketplace.submit()` | `dispatch` prop (= `dispatchMutation`) | yes |
| `install()` (unchanged, cited for the parallel) | `gate.run()` | yes (already was) |

The render-time `busy` (`mutating || pluginBusy`) stays exactly where it was - it only drives `disabled=` props now; no dispatch path reads it any more.

The refusal predicate itself (`mutatingRef.current || gate.isBusy()`) fell out as a pure function of two booleans - extracted to `marketplaceBrowserModel.ts` as `refusesMarketplaceWrite(mutating, gateBusy)` and unit-tested in `marketplaceBrowserModel.test.ts` (3 new cases: neither refuses, this view's own write refuses, the plugin gate refuses). That's the one piece of this fix a render harness was never needed for - the rest (the ref, `dispatchMutation` itself, the modal's `dispatch` prop) is still untested per #1526, same as rounds 1 and 2.

### Gates
- `mobile-native`: `npx tsc --noEmit --project tsconfig.check.json` - clean
- `mobile-native`: `npx vitest run src/pluginMutationGate.test.ts src/marketplaceBrowserModel.test.ts` - 11/11 pass (7 in `marketplaceBrowserModel.test.ts`, 3 new)
- No biome over `mobile-native` (no config there, per repo convention)

obra:
## Review round 4 disposition (f81d0d449)

Took the converged fix, not another guard. Marketplace writes now call `gate.run()` directly, exactly like every plugin mutation already did - there is exactly one busy (the gate's own, via `gate.subscribe`/`isBusy`), not a view-local flag that could disagree with it. `mutating`, `mutatingRef` and `dispatchMutation` are gone. `install()` now delegates to `act()` (both were doing the identical `gate.run` + refusal/error dance).

| Surface | Goes through `gate.run()`? |
|---|---|
| install | yes (unchanged) |
| upgrade | yes (unchanged, `PluginsScreen`'s `act()`) |
| remove-plugin | yes (unchanged, `PluginsScreen`'s `act()`) |
| marketplace add | **yes (new)** - `AddMarketplace.submit()` takes `gate` as a prop and calls `gate.run()` itself |
| marketplace remove | **yes (new)** - via `act()` |
| marketplace refresh | **yes (new)** - via `act()` |
| marketplace edit | N/A - no native UI calls `editMarketplace` anywhere in this file |

(`enablePlugin`/`disablePlugin`/`setPluginAutoUpgrade` also go through `PluginsScreen`'s `act()` → `gate.run()`, unchanged; not asked for but consistent with the rest of the table.)

**Medium** - accepted, reproduced the reachability path by reading it: `dispatchMutation` refused on `gate.isBusy()` but never wrapped the run in `gate.run()`, so a marketplace write never marked the shared gate busy, and `install()`'s own `gate.run()` never saw it. Fixed by routing every marketplace write through the actual gate.

**Low 1** - accepted. Restored `if (!marketplace || busy) return` at the top of both `remove()` and `refresh()` (matching `PluginsScreen.tsx`'s own `remove()`, which keeps both the pre-check and the live refusal `gate.run()` provides on confirm).

**Low 2** - accepted, and per the ruling, closed rather than patched: `refusesMarketplaceWrite` is deleted along with its only caller (`dispatchMutation`) and its truth-table tests. No new behaviour test was written in its place - measured that none is needed: `pluginMutationGate.test.ts` already covers exactly the three behaviours the Low asked for, generically over whatever calls `gate.run()`, not tied to any one caller:
- "a mutation started under one surface refuses the next while it runs" - second write refused while the first is in flight
- "the next mutation runs once the first has finished, failure or not" - the flag clears after both outcomes
- "a refusal is a distinct outcome... not a failure" - what lets a caller decide to show `PLUGIN_MUTATION_BUSY` instead of a generic failure (the copy itself, like `install()`'s own wiring, stays untested per #1526 - no render harness)

Since marketplace writes are now just another caller of the same already-proven `gate.run()`, adding a marketplace-specific copy of these tests would test the same mechanism twice under a different name, not new behavior.

Also updated `pluginMutationGate.ts`'s and `MarketplaceBrowser.tsx`'s own doc comments: the gate's header now says it serializes plugin AND marketplace mutations, citing `marketplaces.go`'s `AddMarketplace`/`RemoveMarketplace`/`RefreshMarketplace` `lockStore` acquisitions (measured: lines 164/247/881, all confirmed to call `m.lockStore(ctx, marketplaceAcquireLock, 30*time.Second)`) alongside `install.go`'s.

### Gates
- `mobile-native`: `npx tsc --noEmit --project tsconfig.check.json` - clean
- `mobile-native`: `npx vitest run src/pluginMutationGate.test.ts src/marketplaceBrowserModel.test.ts` - 8/8 pass (back to the pre-round-3 count: `refusesMarketplaceWrite`'s 3 tests removed with the function)
- No biome over `mobile-native` (no config there)

Diff stat vs merge-base: `mobile-native/src/MarketplaceBrowser.tsx | 111 +++++++++++++++--------------`, `mobile-native/src/pluginMutationGate.ts | 32 +++++----` (`marketplaceBrowserModel.ts`/`.test.ts` net to zero vs `origin/main` - round 3's additions there are fully reverted).

obra:
## Simplify disposition (577acd2bf)

Applied the six /simplify findings, no behavior change except item 6's copy string.

1. `MarketplaceBrowser.tsx`'s 12-line busy comment (was 62-73) → two lines pointing at `pluginMutationGate.ts`.
2. `act()`'s comment (was 119-124) → one line: every write goes through the gate; a refusal reads as busy, a throw as failure.
3. The `gate` prop comments on `MarketplaceBrowser` (was 50-54) and `AddMarketplace` (was 358-362) trimmed to describe the state, no "now"/aside. `AddMarketplace`'s gate comment is gone with the prop itself (item 5).
4. `pluginMutationGate.ts`'s header (1-33): kept the why-a-screen-lock paragraph; the affordance-not-correctness paragraph now cites `install.go`, `gc.go`, `marketplaces.go` instead of Go line numbers, and is shorter.
5. `AddMarketplace` no longer takes a `gate` prop or knows `PluginMutationGate`'s type. It takes one `onAdd(params): Promise<boolean>` (false = refused); `MarketplaceBrowser` passes `(params) => gate.run(() => state.addMarketplace(params))`. The modal keeps its own `alive` check and copy.
6. `PLUGIN_MUTATION_BUSY` copy changed from "Another plugin change is still running." to "Another change is still running." — a marketplace remove refused by a marketplace refresh isn't a plugin change. Constant name and `PluginsScreen.tsx` untouched.

### Gates
- `mobile-native`: `npx tsc --noEmit --project tsconfig.check.json` — clean
- `mobile-native`: `npx vitest run src/pluginMutationGate.test.ts src/marketplaceBrowserModel.test.ts` — 8/8 pass
- No biome over `mobile-native`

`git diff -M --stat $(git merge-base origin/main HEAD) HEAD`: 2 files (`MarketplaceBrowser.tsx`, `pluginMutationGate.ts`), 49 insertions / 59 deletions.

PR #1705, opened 2026-09-17
fix(web): retire queue intents by authoritative queue snapshot, not sequence order

### LINKED ISSUE #1704: Cross-tab drain-cleanup can retire a still-queued turn/queue intent (sequence-based inference is unsafe across dispatcher instances)
## What RoboRev flagged

On #1694 (D27, `dispatcher.ts:169-174` at the time), a Medium finding:

> Drain cleanup assumes `intentSequence` reflects server execution order. Separate dispatcher instances (e.g. browser tabs) can dispatch concurrently, so a queue mutation with a lower sequence may be accepted after the drain and then incorrectly settled as consumed, hiding work that remains queued on the server. Recommend retiring only queue mutations identified as consumed by authoritative server data, or enforcing cross-instance per-target ordering, and adding a meaningful multi-dispatcher interleaving test.

The flagged block is byte-identical to `cmd/evener-hub/frontend/src/stores/mutationDispatcher.ts:158-163` at #1694's merge base; it landed on main in #1452 (`e5a9b3e5c`, 2026-09-16).

## Measurement

**(a) Where `intentSequence` is allocated.** `cmd/evener-hub/frontend/src/stores/mutationOutboxIndexedDB.ts:584-590` (`#allocateSequence`): a read-modify-write of a `SEQUENCE_STORE` record inside a real IndexedDB transaction, keyed by `targetRef`. Not an in-memory per-dispatcher counter — IndexedDB's own transaction ordering makes it a single, global, monotonic counter per target ref, shared by every tab of the origin (IndexedDB is one database per origin). So sequence numbers ARE consistently ordered across tabs — but they order intent **creation**, not server-side acceptance.

**(b) Whether anything gives one tab exclusive dispatch for a target.** No. `mutationOutboxIndexedDB.ts:82-83`, the class's own header comment: "Cross-tab correctness comes from IndexedDB readwrite transaction ordering rather than a browser lease or leader election." Each `MutationDispatcher` instance's per-target FIFO (`#dispatching`, `#requestedRuns`) is in-memory, per-instance state — it serializes attempts *within* one tab's own dispatcher, never across tabs. Two tabs, each with `client?.state === "ready"`, can both be inside `#drainTarget` for the same `targetRef` at the same time.

**(c) Whether a tab's `listOptimistic` sees another tab's records.** Yes. `mutationOutboxIndexedDB.ts:165-173`: `listOptimistic(targetRef)` reads `transaction.objectStore(OPTIMISTIC_STORE).getAll()` with no per-client/per-tab scoping — every tab reads the same shared, origin-wide IndexedDB database.

## Conclusion: the interleaving can happen

The removed code's own reasoning comment said: "dispatch is per-target FIFO, so an earlier optimistic queue intent was necessarily accepted before the drain was sent, and accepted-before-drain means consumed-by-drain." That FIFO guarantee holds only within one dispatcher instance. Given (a)-(c):

1. Tab A allocates a `turn/queue` intent (sequence s1) but hasn't sent it yet (e.g. momentarily disconnected, or another record ahead of it in A's own FIFO).
2. Tab B allocates a `turn/drainAsSteer` intent (sequence s2 > s1, created later) and its dispatcher sends and gets it applied by the server *first* — the server's queue is now empty.
3. Tab A's queue intent is *then* sent and accepted by the server as **genuinely new work** on the now-empty queue (the server never saw it before the drain).
4. If tab B's own drain-cleanup snapshot read (`listOptimistic`) happens to land after step 3's write reaches the shared IndexedDB, the old code sees a `turn/queue` record with `intentSequence < record.intentSequence` (s1 < s2) and — using sequence order as a proxy for "accepted before this drain was sent" — incorrectly retires it, even though the server's queue genuinely still holds it.

This is exactly the hazard the reviewer described: real user-visible data loss (a message the server accepted disappears from the client's own tracking).

## Fix

Filed as PR #TBD against main (`cmd/evener-hub/frontend`, where the code lives today; #1694 is a pure relocation of this class into the package and stays that way).

- `MutationDispatcher#attempt`'s post-drain heuristic (sequence comparison) is removed entirely.
- A new public method, `retireConsumedQueueIntents(targetRef, authoritativeQueueIds)`, retires an optimistic `turn/queue` record only when it is absent from a caller-supplied, freshly-read authoritative set — timing-independent by construction, since it never compares local sequence numbers across dispatcher instances.
- `threads.ts`'s `handleNotification` wires `thread/queueChanged`'s own `clientMutationIds` (the server's current queue, confirmed server-side at `agent/session_client_mutation_queue.go:534-556`, `reflectDurableInputQueue` — always emitted after a queue-revision change, drains included) straight into the new method. This is the same authoritative-per-notification pattern `reconcileIdentities`/`restoreProvenAbsent` already use elsewhere in this file.

TDD: `mutationDispatcher.test.ts` gained a two-tab test constructing exactly the counter-example above (a lower-sequence queue intent accepted after another tab's drain, kept because the authoritative set still names it) and a falsification confirming it fails under the old sequence-based rule; `threads.test.ts` gained an end-to-end wiring test, also falsified against the old code path.

obra:
## Summary

Fixes #1704: RoboRev's Medium on #1694 (D27's relocation review) flagged that `MutationDispatcher`'s post-drain cleanup infers "accepted before the drain was sent" from `intentSequence` comparison — a rule that only holds within one dispatcher instance's own per-target FIFO. Two browser tabs share one outbox with no cross-tab lease or leader election (`mutationOutboxIndexedDB.ts`'s own header comment), so a second tab's `turn/queue` intent with a *lower* sequence can be accepted by the server *after* another tab's drain was already applied — genuinely fresh work on the now-emptied queue, which the old rule retired anyway, hiding it from the UI.

A first pass (`retireConsumedQueueIntents`) fixed that specific hazard but was then found, on review, to have three more problems of its own — an High and two Mediums, all confirmed by measurement (full detail in the PR comments below): a legacy server push path never populates `client_mutation_ids` at all, and the naive `?? []` treated that "coverage unknown" state as "authoritatively empty"; there was no revision or per-target ordering, so a stale/reordered snapshot could still misfire; and retirement was wired only from `queueChanged`, missing a replayed drain (which emits no fresh push) or a lost push entirely.

**The mechanism this PR lands on**, closing #1706: one dispatcher entry point, `reconcileQueueSnapshot(targetRef, snapshot)`, where `snapshot` is `{ ids: ReadonlySet<string> | undefined, revision: number, authoritative: boolean }`.

- `ids: undefined` means coverage is unknown and nothing is ever retired on that basis — never guessed. `queueSnapshotFromWire` (threads.ts) is the one place that translates the wire's `QueueState` into this: `clientMutationIds` when present, an empty set when `depth === 0` proves the queue empty without them, `undefined` otherwise.
- Per-target revision monotonicity: a snapshot no newer than the last one reconciled for that ref is ignored outright.
- Per-target serialization (a promise chain, like the dispatch FIFO): one call's read-then-write can't race another for the same ref.
- `authoritative: false` (a saved/incompatible/not-loaded hydrate) settles and retires nothing, matching the existing `markUnknown` branch's own gate.
- It settles ids the snapshot names (via `reconcileIdentities`, the same rule any other authoritative feed uses) and retires `turn/queue` records absent from it, in one write transaction (`mutationOutboxIndexedDB.ts`'s new `settleOptimisticAbsent`, beside `restoreProvenAbsent`, reusing `settleApplied`'s own delete rule via a shared private helper) — replacing the read-then-N-writes shape the first pass had.
- Called from `handleNotification`'s `thread/queueChanged` branch (one chain now, not two independently-scheduled ones sharing state) and from the hydrate reconciliation's `mutationStateAuthoritative === true` branch (beside `restoreProvenAbsent`) — the latter is what catches a replayed drain or a lost push, since a reconnect always re-hydrates.

This is a pure web-side fix against `main`; #1694 (the D27 relocation of this class into `@evener/appwire-client`) stays a pure move and is unaffected.

## Test plan

- [x] `mutationDispatcher.test.ts`: `reconcileQueueSnapshot` tests for each finding — retires an id absent from an authoritative snapshot, settles a named id (same as `reconcileIdentities` would), leaves everything alone when `ids` is `undefined` (unknown coverage), ignores a snapshot no newer than the last reconciled revision, settles/retires nothing when `authoritative: false`, and the two-tab counter-example demonstrating the mechanism (per-target revision monotonicity, never sequence comparison) — every guard-dependent test verified red under a temporary reversion of its own guard and restored
- [x] `threads.test.ts`: a `queueChanged`-driven wiring test and a hydrate-driven "missed push / replayed drain" test (two accepted queue intents, an authoritative hydrate naming neither, both retired) — both falsified by disabling their respective `handleNotification`/hydrate call site and confirming failure, restored
- [x] Also fixed, in the same push: a pre-existing test-isolation gap in `threads.test.ts` (`periodic discovery resumes authority checks...`) that this PR's added hydrate work exposed by extending an already-only-partially-awaited reconciliation tail — added a small drain so one parameterized case's tail can't bleed into the next
- [x] `npx vitest run src/stores/threads.test.ts src/stores/mutationDispatcher.test.ts src/stores/mutationDispatcher.edge.test.ts src/stores/mutationOutbox.test.ts src/stores/mutationOutboxIndexedDB.stall.test.ts` — 395 tests, all green (repeated 3x, no flakes)
- [x] `npm run typecheck` (frontend) — clean
- [x] `npm run lint` (frontend, covers `src` + the package) — clean, no fixes


obra:
## Simplify disposition (e56b34604)

All six items applied, one push, three commits:

1. **Reuse**: `settleOptimisticAbsent(targetRef, method, authoritativeIds)` added to `mutationOutboxIndexedDB.ts` beside `restoreProvenAbsent` — one write transaction over OUTBOX/OPTIMISTIC/RECOVERY, reusing `settleApplied`'s own delete rule via a shared private helper (`#settleAppliedWithinTransaction`) both now call, returning the settled ids. `retireConsumedQueueIntents` collapses to the same wrapper shape `restoreProvenAbsent` is. **Note**: this branch (main, pre-D27) has no `MutationOutboxStorage` port or in-memory test double yet — `MutationDispatcher` still takes the concrete `MutationOutboxIndexedDB` class directly (D27 is what introduces the port); grepped for implementers, found none, so nothing else to update. #1694's own D27 relocation will need this same method added to the package's port/interface when it next refreshes past this PR — flagging so it isn't missed there.
2. **Efficiency**: `retireConsumedQueueIntents` now returns the settled ids (folded into item 1). Folds into item 3's gating below.
3. **Ordering**: `handleNotification`'s two independently-scheduled chains (`reconcileIdentities` → `refreshMutationPins` and `retireConsumedQueueIntents` → `refreshMutationPins`) are now one sequential chain per notification, sharing one `refreshMutationPins` call gated on either step having actually settled something (a `queueChanged` with nothing optimistic no longer pays for the refresh's own reads).
4. **Comments**: `mutationDispatcher.ts`'s method comment trimmed to the rule and why (absence from the snapshot means consumed, no push names the ids) with history cited to #1704; the two-tab test's preamble trimmed the same way.
5. **Test**: the two-tab test's duplicated `new MutationDispatcher(tabA, ...)` hoisted to one const.
6. **Wording**: renamed "keeps an optimistic queue intent the authoritative queue still names" → "leaves a queue intent alone when its authoritative snapshot still names it", stating the method's own scope (retirement, not settlement timing — `reconcileIdentities` runs regardless, fed the same ids). Reworded the PR body to say the same thing explicitly, and to cite #1706 as the separate, not-done-here generalization.

Targeted gates: `npx vitest run` on `mutationDispatcher.test.ts`, `mutationDispatcher.edge.test.ts`, `mutationOutbox.test.ts` (the actual `MutationOutboxIndexedDB` oracle — no separate `mutationOutboxIndexedDB.test.ts` exists), `mutationOutboxIndexedDB.stall.test.ts`, `threads.test.ts` — 391 tests, all green; `tsc` clean; `npm run lint` (covers `src` + the package) clean, no fixes.

Diff vs merge-base: `git diff -M --stat $(git merge-base origin/main HEAD) HEAD` — 5 files, 197 insertions, 72 deletions (pasted above the fix's own commits).

Not touching #1694 per your note; it refreshes after this merges.

obra:
## Review round disposition (3b2b5861d)

Implemented the #1706 shape RoboRev's High + two Mediums asked for: one dispatcher entry point, `reconcileQueueSnapshot(targetRef, { ids, revision, authoritative })`, replacing the two half-rules (`retireConsumedQueueIntents` and the separately-scheduled `reconcileIdentities` chain on the same ref).

### Measurement (before implementing)

- **High** confirmed real: `session_queue.go`'s `queueChangedDataLocked` (used by `pushQueueHead` when a queue entry has no `ClientMutationID` — the interrupt-return path) never sets `ClientMutationIDs` (`omitempty`), while still reporting a genuinely non-empty `Depth`/`Preview`/`IDs`. The old `?? []` treated that as "empty," retiring every accepted `turn/queue` record regardless of truth.
- **Medium 1** confirmed real: no revision or per-target ordering existed; `intentSequence` orders intent creation, not server acceptance, and nothing prevented a stale/reordered snapshot from acting after a fresher accept.
- **Medium 2** confirmed real: `session_client_mutation_queue.go:798-806` returns on `clientMutationDispositionReplayed` before `reflectDurableInputQueue` — a replayed drain emits no fresh `queueChanged`. Hydrate's own `collectAuthoritativeMutationIds` only *settles named* ids; nothing retired *absent* `turn/queue` ones on hydrate. Did **not** add a third call site — hydrate (already reached on every reconnect) is what covers the missed-push/replayed-drain case, gated on `mutationStateAuthoritative`.

### Call-site table

| Event | Snapshot fields | Retires absent `turn/queue`? |
|---|---|---|
| `thread/queueChanged` (push) | `ids` from `clientMutationIds` if present, else empty iff `depth===0`, else `undefined`; `revision` from wire; `authoritative: true` | Yes, when `ids` defined and revision is newer than last reconciled for that ref |
| `thread/read` hydrate, `mutationStateAuthoritative===true` | same translation over `evener.queue`; `authoritative: true` | Yes, same rule (this is what catches a lost push or a replayed drain) |
| `thread/read` hydrate, not authoritative / restartRequired / notLoaded | n/a | No — `authoritative: false` short-circuits, matching the existing `markUnknown` branch |

### Test-isolation fix, and why it's pre-existing

My new hydrate call added enough extra async work to an *already partially-awaited* reconciliation tail that a pre-existing test (`periodic discovery resumes authority checks for saved '*' delegates, accepted=false`) started failing — order-dependent, only when the whole `threads.test.ts` suite ran together, never in isolation. Root cause: the `accepted=true` sibling case in the same `test.each` family only awaits `settled.promise` (resolves on the *first* settle call, not the whole reconciliation my new step extends), and reuses the same hardcoded id + default database across every case in the family. This gap predates this PR — my change only made the tail long enough to reliably straddle the next test's `beforeEach`. Fixed by adding a small `flushIndexedDBUntil` drain at the end of that test; verified stable across 7+ repeated full-suite runs.

### Everything else from the simplify batch

Items 1-6 from the prior review (reuse via `settleOptimisticAbsent`, the pins-refresh gating, one chain per notification, trimmed comments, the hoisted duplicate dispatcher, the retitled test) are all still in this push, unchanged from their own disposition comment above.

Diff vs merge-base: `git diff -M --stat $(git merge-base origin/main HEAD) HEAD` — 5 files, 417 insertions, 74 deletions. Production only (`mutationDispatcher.ts`, `mutationOutboxIndexedDB.ts`, `threads.ts`): 163 insertions, 43 deletions (206 total). Test-only (`mutationDispatcher.test.ts`, `threads.test.ts`): 254 insertions, 31 deletions (285 total).

Targeted gates: `npx vitest run` on `threads.test.ts`, `mutationDispatcher.test.ts`, `mutationDispatcher.edge.test.ts`, `mutationOutbox.test.ts`, `mutationOutboxIndexedDB.stall.test.ts` — 395 tests, all green (repeated 3x); `tsc` clean; `npm run lint` clean, no fixes.

Closes #1706 (the general "one entry point per notification, revision-gated, per-target-serialized" mechanism it asked for is exactly `reconcileQueueSnapshot`).

obra:
## Review round 2 disposition (7014f970e)

The three simplify items (2, 3, 4) and the three new Mediums (M-a, M-b, M-c) are folded into one cursor lifecycle on `MutationDispatcher`, replacing the single `revision`-only cursor with `#lastReconciledQueue: Map<targetRef, { instanceId, revision }>`.

### Cursor lifecycle table

| Step | Condition | Behavior |
|---|---|---|
| 1. Non-authoritative | `!snapshot.authoritative` | Return `[]`, cursor untouched (unchanged from round 1). |
| 2. Same-instance stale | `last.instanceId === snapshot.instanceId && snapshot.revision <= last.revision` | Return `[]`, cursor untouched. |
| 3. Different-instance (M-a) | `last.instanceId !== snapshot.instanceId` (including `last === undefined`) | Never stale - proceeds regardless of the revision number, since a new session instance's revision counter is not comparable to the old one's. |
| 4. Unknown coverage | `snapshot.ids === undefined` | Return `[]` before recording anything (this is simplify item 2, unchanged from round 1: unknown coverage must never advance the cursor). |
| 5. Reconcile + retire | writes via `reconcileIdentities` + `settleOptimisticAbsent` | Cursor is recorded **only after both writes succeed** (M-b): `#lastReconciledQueue.set(targetRef, { instanceId: snapshot.instanceId, revision: snapshot.revision })` moved to run after `settleOptimisticAbsent` returns, not before. If either write throws, the cursor is left exactly where it was and the promise rejects (the existing `.catch()` in `threads.ts`'s `handleNotification` already treats that as "a later notification retries"). |

### M-a: instanceId, measured

`server/appwire_runtime.go:2077-2118` (`handleAppThreadClear`) replaces `s.appThreadID` with a new value on every successful clear (`currentInstanceID := s.appThreadID`, compared against and then superseded by `params.ExpectedInstanceID`'s replacement). `server/appwire_runtime.go:2516` sets `EvenerThread.InstanceID: threadID` from the same `threadID` that `s.appThreadID`/`sessionID` denote (`server/appwire_runtime.go:1264`: `sessionID == s.appThreadID`). The queue revision counter itself (`agent/session_client_mutation.go:208` `QueueRevision uint64`, bumped throughout `agent/session_client_mutation_queue.go`) lives on the session's own snapshot, so a new session (new instance) starts a fresh counter at 0. `targetRef` is the stable workspace ref and does not change across a clear.

The wire already carries this identity on both snapshot sources:
- `thread/queueChanged`: `ThreadQueueChangedParams.ThreadID` (`appwire/types.go:849`), populated from `p.threadID = event.SessionID` (`internal/appprojector/appwire_projection.go:264-265`) - the same session id.
- `thread/read` hydrate: `EvenerThread.InstanceID` directly (falling back to `Thread.ID`, the same fallback `threadInstanceID` already uses elsewhere in `threads.ts` for `expectedInstanceId`).

`QueueSnapshot` now carries `instanceId: string` (`mutationDispatcher.ts`), threaded through both `queueSnapshotFromWire` call sites in `threads.ts`. TDD: `mutationDispatcher.test.ts` "a snapshot from a new session instance is never stale against the old instance's higher revision" - falsified by reverting the instance-aware gate (`git diff HEAD~1 HEAD -- .../mutationDispatcher.ts | git apply -R`, ran red, restored with the same patch reversed) before restoring the fix.

### M-b: cursor-after-writes, measured

Round 1's code recorded the cursor before `reconcileIdentities`/`settleOptimisticAbsent` ran. TDD: `mutationDispatcher.test.ts` "a failed reconcile leaves the cursor where it was, so a retry at the same revision still acts" - `vi.spyOn(outbox, "settleOptimisticAbsent").mockRejectedValueOnce(...)`, confirmed red against round 1's ordering, green after moving the `.set()` call to after both writes.

### M-c: measured, declining to invent a watermark

Checked whether anything on the wire lets an outbox record be stamped with the queue revision at which it was accepted:
- `MutationReceipt` (`appwire/types.go:1779-1787`, the `turn/queue` response shape) carries `ClientMutationID`, `Disposition`, `ThreadID`, `InstanceID`, `TurnID`, `QueueEntryIDs`, `ProjectionState` - no revision field.
- `composerMutationIntent`'s `queue`/`steer` route (`threads.ts`) sends no `expectedQueueRevision` at all - only `drainAsSteer` does, as its own destructive precondition. `turn/queue` has never carried one to echo back.
- `queueChanged`'s `Revision` (`internal/appprojector/appwire_projection.go:1082-1095`) is emitted by the same per-session projector that owns the counter, so per-ref notification delivery is generation-ordered on one connection - this is what round 1's per-target revision monotonicity already closes (a stale/reordered notification for the same ref cannot arrive after a fresher one). It gives no per-record stamp.

What's left unbound: `reconcileQueueSnapshot`'s writes (`reconcileIdentities` + `settleOptimisticAbsent`) run on the dispatcher's own `#queueReconciliations` chain, which is NOT the same chain a concurrent accept's optimistic write goes through (that write lands directly via the send/receipt path, independent of any reconciliation in flight). So a snapshot captured at revision R, still being applied in `settleOptimisticAbsent`'s transaction, can in principle interleave with a same-client accept whose optimistic record lands in storage between the snapshot's capture and that transaction's read - the record's id is legitimately absent from `snapshot.ids` (it didn't exist yet at R) but the record is, in truth, still queued by the time the transaction runs. Every mutating route requires `ensureThread` (a hydrate) first, and that hydrate's own `reconcileQueueSnapshot` call seeds the cursor before any accept for that target is possible, which closes the widest version of this gap (the very first reconciliation ever for a ref); what remains is the narrow one described above, bounded by ordinary IndexedDB/network latency rather than anything unbounded.

I'm not stamping records with a locally-tracked "last known revision" at accept time, since that value can itself be stale relative to the true current server revision (e.g. another tab just pushed) and would give a false sense of precision rather than closing the race - it would just move the same TOCTOU one level down. Nothing on the wire (no receipt field, no echoed precondition) supports a real per-record watermark for `turn/queue`. Filing rather than fixing, per your note.

### Gates

- `npx vitest run src/stores/threads.test.ts src/stores/mutationDispatcher.test.ts src/stores/mutationDispatcher.edge.test.ts src/stores/mutationOutbox.test.ts src/stores/mutationOutboxIndexedDB.stall.test.ts` - 399 tests, all green (also ran `threads.test.ts` alone 3x for the known #1713 order-dependence; stable each time, no flake).
- `npm run typecheck` - clean.
- `npm run lint` (`biome ci src ../../../appwire-client/typescript`) - clean, no fixes.

`git diff -M --stat $(git merge-base origin/main HEAD) HEAD`: 5 files, 566 insertions, 74 deletions (production: `mutationDispatcher.ts`, `mutationOutboxIndexedDB.ts`, `threads.ts`; tests: `mutationDispatcher.test.ts`, `threads.test.ts`).

Head: `7014f970ef0c87495d6d21da01d753d457831655`, `gh pr view 1705 --json mergeable` → `MERGEABLE`.

obra:
## Review round 3 disposition (314875c86)

### The High: wire-shape table

| Message | Go field + tag (`appwire/types.go`) | Shape on the wire when the queue is authoritatively empty | What `queueSnapshotFromWire` reads |
|---|---|---|---|
| `thread/queueChanged` (`ThreadQueueChangedParams.Queue: QueueState`) | `Depth int` `json:"depth,omitempty"` (:831); `ClientMutationIDs []string` `json:"clientMutationIds,omitempty"` (:834); `Revision uint64` `json:"revision"` (no omitempty) | `{"revision": N}` - both `depth` and `clientMutationIds` absent | `ids = clientMutationIds` if present, else `new Set()` iff `(queue.depth ?? 0) === 0`, else `undefined` |
| `thread/read` hydrate (`EvenerThread.Queue: QueueState`) | same `QueueState` struct | same: `{"revision": N}` | same function, called from the `mutationStateAuthoritative` branch |

Fixed: `queue.depth === 0` read `false` against `undefined` (the field is never `0` on the wire, only absent), so `ids` fell to `undefined` (unknown coverage) on every genuine empty-queue push or hydrate and `#reconcileQueueSnapshotNow` bailed - the exact symptom #1705 set out to fix. Now `(queue.depth ?? 0) === 0`, the same convention `liveControls.ts`'s `controlsFor` already uses for this field.

**Fixture sources**: no dedicated fixture module exists (grepped for one; none found), but most of `threads.test.ts`'s own `testThread`/`readResponse` helpers already default `queue` to `{ revision: 0 }` with no `depth`/`clientMutationIds` - that IS the genuine shape. The three fake ones were specific to #1705's own added tests:

- `threads.test.ts` "a queueChanged snapshot that no longer names an accepted queue intent retires it" - `queue: { revision: 8, clientMutationIds: [] }` → `queue: { revision: 8 }`. Confirmed red against the pre-fix `queue.depth === 0` comparison (reverted the fix, ran the test, got the exact "record never retired" failure; restored).
- `threads.test.ts` "a hydrate whose authoritative queue names neither accepted queue intent retires both" - `queue: { revision: 99, clientMutationIds: [] }` → `queue: { revision: 99 }`. Same falsification, same result, restored.
- `threads.test.ts` "a queueChanged naming an optimistic id that reconcileIdentities settles still refreshes pins" (mine, round 2) - added `depth: 1` alongside the named `clientMutationIds`, since a real non-empty queue always carries depth (only omitted at zero). This one's pass/fail doesn't depend on the depth fix (the `clientMutationIds`-defined branch short-circuits before depth is ever read), so it's a realism fix, not a falsified one.

Not touched: `threads.test.ts:7967`'s `queue: { revision: 9, clientMutationIds: [] }` (a pre-existing test, not part of this PR's diff against `origin/main`, testing `restoreProvenAbsent`'s outbox-retry path rather than `reconcileQueueSnapshot`'s `settleOptimisticAbsent` - out of scope per "tests you and the previous lane wrote"). `mutationDispatcher.test.ts`'s own `queueSnapshot()` helper builds the POST-translation `QueueSnapshot` shape directly (never a wire `QueueState`), so none of its tests needed conversion.

### Medium 1: instance ordering, measured

Checked whether `thread/clear` or a `thread/read` response carries any generation/sequence spanning a clear:

- `server/appwire_runtime.go:2077-2118` (`handleAppThreadClear`) replaces `s.appThreadID` with a new value on every successful clear.
- `server/appwire_runtime.go:2516` sets `EvenerThread.InstanceID: threadID` from that same identity; `server/appwire_runtime.go:1264` (`sessionID == s.appThreadID`) confirms it's the session id.
- `agent/session_client_mutation.go:208` (`QueueRevision uint64`) is a field on the session's own snapshot - a new session (new instance) starts this counter fresh at 0.
- `ThreadQueueChangedParams.ThreadID` (`appwire/types.go:849`) is populated from `p.threadID = event.SessionID` (`internal/appprojector/appwire_projection.go:264-265`) - the same session id, not a separate sequence.
- The one thing on the wire that DOES carry a generation+sequence is `NavigationInvalidatedPayload` (`appwire-client/typescript/types.gen.ts:1192-1195`, `generationId`/`sequence`, with `isSequenceGap` in `state/navigation/invalidation.ts`) - a wholly separate subsystem (`evener/navigation/invalidated`) with no analog on `thread/queueChanged` or `thread/read`.

So the wire gives no ordering across instances for this subsystem. Kept the fallback: `#supersededQueueInstances: Map<targetRef, Set<instanceId>>`. Once a target reconciles a snapshot from a NEW instance, the instance it was on before is added to this set; any later snapshot naming a superseded instance is rejected outright regardless of its own revision. Bounded by clears-per-session (realistically single digits over a session's lifetime).

TDD: `mutationDispatcher.test.ts` "a late snapshot from a superseded instance is ignored even at a higher revision than the new instance's own" - instance A reconciles at revision 5, instance B's first snapshot (revision 1) is accepted and A is superseded, then A's own revision-6 snapshot (in flight since before the clear) is rejected. Falsified against the pre-Medium-1 code (only same-instance revisions compared) - failed exactly as expected (a stale-instance snapshot retired a record it had no business touching); restored.

### Medium 2 (#1717): serialize accept with reconciliation, measured

Measured the write path first: an accepted `turn/queue` intent's optimistic write happens in `#attempt` (`await this.#storage.settleReceipt(...)`), reached via `#drainTarget` → `#runTarget` → `#dispatchTarget`'s own chain (`#dispatching`). A queue snapshot's retire-scan (`settleOptimisticAbsent`) runs inside `#queueReconciliations`. These are two independent per-target chains with nothing serializing one against the other - fits well under 30 lines to close: extracted `#serializedWithQueueReconciliation(targetRef, fn)` (the same chaining `reconcileQueueSnapshot` already used, now shared), and routed `#attempt`'s `settleReceipt` call through it too. Noting on #1717 that the client-side half described there (no revision on `MutationReceipt` to stamp a record with) is closed by this serialization instead of a watermark.

TDD: `mutationDispatcher.test.ts` "an accept for a target with an in-flight reconciliation is serialized behind it, never interleaved" - gates `settleOptimisticAbsent` open, starts a reconciliation naming nothing (would retire an existing optimistic record), then accepts a brand-new intent for the same target with a controlled (deferred) network response so the exact moment it becomes free to write is known. Falsified against pre-fix code: the accept's write landed immediately (unblocked), so once the gate opened the scan retired BOTH records; fixed, the scan runs first (having never seen the not-yet-written new record) and the accept's write lands only afterward. Getting this test to actually exercise the race took two false starts - a microtask-only spin (`await Promise.resolve()` in a loop) never lets fake-indexeddb's own macrotask-scheduled requests advance, so two earlier versions of this test passed on both fixed AND unfixed code (a false negative). Replaced with a `flushIndexedDBUntil` helper (same idea as `threads.test.ts`'s own, adapted to this file's per-test `IDBFactory` instances) that pumps real IDB round trips instead of bare microtasks.

### Gates

- `npx vitest run src/stores/threads.test.ts src/stores/mutationDispatcher.test.ts src/stores/mutationDispatcher.edge.test.ts src/stores/mutationOutbox.test.ts src/stores/mutationOutboxIndexedDB.stall.test.ts` - 401 tests, all green; `mutationDispatcher.test.ts` alone repeated 3x given the new timing-sensitive test, stable each time.
- `npm run typecheck` - clean.
- `npm run lint` (`biome ci src ../../../appwire-client/typescript`) - clean, no fixes.

`git diff -M --stat $(git merge-base origin/main HEAD) HEAD`: 5 files, 744 insertions, 75 deletions.

Head: `314875c860ae5f15c34c8d5a309cda2384735aa1`, `gh pr view 1705 --json mergeable` → `MERGEABLE`.

obra:
## Review round 4 disposition (93f32f69a)

One per-target state record replaces the round-3 cursor: `#queueState: Map<targetRef, { instance, revision, cut }>`, plus `#supersededQueueInstances` (unchanged) and a new `#acceptCounters: Map<targetRef, number>` - one monotonic counter per target, advanced only by `#nextAcceptStamp` (an accept). `QueueSnapshot` gained `cut: number`, which the **caller** captures via the newly-exposed `dispatcher.peekAcceptCounter(targetRef)` at the snapshot's true arrival, not by the dispatcher at whatever later moment `reconcileQueueSnapshot` is actually called.

### Transition table

State = `{ instance, revision, cut }` per target (`undefined` before the first event). `superseded` = the target's own set of dead instance ids.

| Event | Precondition | Next state | Effect on records |
|---|---|---|---|
| hydrate, not authoritative (`!snapshot.authoritative`) | any | unchanged | none |
| queueChanged / hydrate-authoritative / clear, `instance ∈ superseded` | any | unchanged | none - rejected outright, revision not even read |
| queueChanged / hydrate-authoritative / clear, `instance === state.instance`, `revision ≤ state.revision` | state defined | unchanged | none - stale, rejected |
| queueChanged / hydrate-authoritative / clear, `instance ≠ state.instance` (first sight, incl. no prior state) | any | `state.instance` (if defined) added to `superseded`; state ← `{ instance: new, revision: -1, cut: snapshot.cut }` **before** the ids branch below runs | none yet - this reset happens regardless of ids |
| …continuing, `ids === undefined` (unknown coverage) | revision check above already passed | unchanged from whatever the reset/prior row left it | none - never guesses |
| …continuing, `ids` defined (the scan) | revision check passed | `state ← { instance: snapshot.instance, revision: snapshot.revision, cut: snapshot.cut }`, recorded only after the writes below succeed | `reconcileIdentities(ids)` settles named ids; scan retires optimistic `turn/queue` records absent from `ids` **and** `acceptStamp < cut` |
| accept (a receipt processed in `#attempt`) | - | `#acceptCounters[target]` advances by 1 | the record's optimistic copy is written with `acceptStamp` = the counter's value **before** this increment |

The "first sight of a new instance" row is what closes Medium 2: it fires from **any** of the three snapshot sources and runs **before** the ids branch, so an unknown-coverage snapshot from a new instance still supersedes the old one immediately instead of returning early with nothing superseded (round 3's bug).

### Medium 1, measured

`#serializedWithQueueReconciliation` (round 3's Medium 2 fix) proves an accept's write and a scan never physically interleave - it does not prove either one predates the other. Chain-enqueue order is call order, not causal order: a hydrate's own `reconcileQueueSnapshot` call happens only after `reconcileIdentities`/`restoreProvenAbsent` (real awaits) - an accept landing during that gap can register into the shared chain first, so the scan (running after it, whatever the chain position) would see it. Fixed by treating "when was this snapshot's data true" as a fact the **caller** owns: `cut` is read at the earliest point the caller has the data (right at the top of the hydrate's own reconciliation closure for hydrate; the notification handler itself for queueChanged, which has no gap), stored, and threaded through unchanged even if the actual `reconcileQueueSnapshot` call happens much later. The stamp lives in `mutationOutboxIndexedDB.ts` alone: a local `StampedOptimisticRecord` (`acceptStamp?: number`) extending the package's own `MutationOptimisticRecord` - never added to `state/mutation/records.ts` itself, so it does not touch the D26 record core and native carries nothing new. Records without a stamp (written before this existed) are treated as always-eligible, matching prior behavior exactly.

TDD: `mutationDispatcher.test.ts` "a snapshot's cut, captured before an accept, protects that accept even though the scan runs after it lands" - reads `cut` before the accept, then dispatches it, then reconciles with the early cut; falsified by disabling the stamp check in `settleOptimisticAbsent` (confirmed red - the fresh record was wrongly retired), restored.

### Medium 2, measured

Traced precisely: instance A active (`state = {A, 5}`). Instance B's first snapshot has unknown coverage (a legacy push) - round 3's code checked `ids === undefined` and returned **before** ever comparing/updating the instance, so `state` stayed `{A, 5}` and `superseded` stayed empty. A's own delayed revision-6 snapshot then arrived: same instance as `state.instance` (still A), revision 6 > 5 → accepted, `state` flips back to `{A, 6}`. Only when B's next snapshot (this time with real ids) finally arrived did the cross-instance branch fire, superseding A - correctly, but only by accident of B eventually sending a defined-ids snapshot; the delayed A-snapshot in between had already wrongly reconciled using dead data. Fixed by moving the instance-transition check to run unconditionally on first sight of any new instance, before the ids branch.

TDD: `mutationDispatcher.test.ts` "a new instance's unknown-coverage snapshot supersedes the old instance immediately" - B's unknown-coverage snapshot, then A's delayed revision-6 snapshot (must be rejected, not accepted), then B's real snapshot (must still work normally). Falsified by restoring round 3's ordering (`ids === undefined` check before the reset) - confirmed red (A's delayed snapshot wrongly retired a live B-instance record), restored.

### Medium 3, measured

`applyClearResponse` (`threads.ts`) published the replacement model and set `mutationAuthorityRefs`, but never called `reconcileQueueSnapshot` at all - confirmed by grep (no call site existed) and by the new test. Fixed: it now builds a `QueueSnapshot` from `response.thread.evener.queue` + `instanceId ?? id`, `authoritative: true` (a clear response always sets `mutationStateAuthoritative`), `cut` from `peekAcceptCounter`, feeds it through the same `reconcileQueueSnapshot` entry point, then refreshes pins on the same `settled.length > 0 || ids.size > 0` rule the other two call sites use.

TDD: `threads.test.ts` "clearThread's own response retires an accepted queue intent from the instance it replaced" - queues an intent under the old instance (genuine wire shapes throughout, per round 3's own rule), clears to a new instance whose queue is empty (`{revision: 0}`, no `depth`/`clientMutationIds`), asserts the ghost is retired without any further push or hydrate. Falsified by disabling the new wiring - confirmed red (the ghost survived indefinitely), restored.

### Low, deferred

`#supersededQueueInstances` grows one entry per clear for the target's lifetime - unbounded in principle, bounded in practice by how many times one thread gets cleared in one session. Noted per your instruction; not fixed here.

### Gates

- `npx vitest run src/stores/threads.test.ts src/stores/mutationDispatcher.test.ts src/stores/mutationDispatcher.edge.test.ts src/stores/mutationOutbox.test.ts src/stores/mutationOutboxIndexedDB.stall.test.ts` - 404 tests, all green.
- `npm run typecheck` - clean.
- `npm run lint` (`biome ci src ../../../appwire-client/typescript`) - clean, no fixes.

`git diff -M --stat $(git merge-base origin/main HEAD) HEAD`: 5 files, 998 insertions, 77 deletions.

Head: `93f32f69af8877bb14502c85d3aff7cd5c523a6e`, `gh pr view 1705 --json mergeable` → `MERGEABLE`.

obra:
## Review round 5 disposition (836d20965)

### High: in-memory accept counter has no rehydration - measured

`#acceptCounters`/`#nextAcceptStamp`/`peekAcceptCounter` were in-memory only. Reload measurement (`mutationDispatcher.test.ts` "a reload's fresh dispatcher still retires a ghost accepted by a previous dispatcher instance"): accept a queue intent through dispatcher A, then construct dispatcher B over the *same* durable storage (a fresh `MutationOutboxIndexedDB` connection - simulating a reload or a second tab) and feed it an authoritative empty snapshot. Confirmed red against round-4 code: dispatcher B's own counter reads 0, `record.acceptStamp (0) >= cut (0)` protects the ghost, `settled` comes back `[]` instead of retiring it.

Fixed by deleting the in-memory counter and the local `acceptStamp`/`StampedOptimisticRecord` fields entirely, using the durable order that already existed: `#allocateSequence` gives every intent a per-target, gapless `intentSequence` (already copied onto the accepted optimistic record by `settleReceipt`, unconditionally, before this round). The dispatcher now tracks `#lastAcceptedIntentSequence` per target, updated in `#attempt` at accept and rehydrated once, lazily, from `listOptimistic()` (every record in that store is `state: "accepted"` by type, so its own max `intentSequence` per target IS the target's last-accepted value - no separate durable clock needed). `settleOptimisticAbsent`'s scan now compares `record.intentSequence > cut` (protect) instead of `acceptStamp ?? -1 >= cut`; the comparison direction flips because `cut` is now the last-accepted value itself (inclusive) rather than a pre-increment counter reading (exclusive) - retires only `intentSequence <= cut`.

### Medium: hydrate captured the cut after joining the pending chain - fixed

`publishAndReconcileThreadHydration`'s reconciliation ran inside a `.then()` chained onto `pendingMutationReconciliations.get(ref) ?? Promise.resolve()` - a PRIOR pending reconciliation for the same ref could delay when that closure even starts running, by which point the cut it read was no longer "as of the response's true arrival". Moved the mint into `hydrateAndSubscribe`, immediately after `applyHydrationResponseCut` (right after the response resolves, before returning `ThreadHydration`) - `publishAndReconcileThreadHydration` now uses `hydration.queueSnapshot`, falling back to a fresh capture only for the one bootstrap moment no mutation runtime existed yet at arrival.

Deliberately reads the module's own `mutationRuntime` variable directly in `hydrateAndSubscribe`, never the constructing `getMutationRuntime()`: `handleReady` (the ready-event handler) already calls the constructing form as its own first line, so by the time ANY hydration normally runs, a runtime already exists - calling the constructing form again here would only matter on the very first-ever hydration of a session, and even there costs nothing extra since nothing could have been accepted before a runtime existed either.

### High: queueChanged's threadId vs hydrate/clear's instanceId - measured, fixed

Traced precisely (cited in the fix's own comment on `queueSnapshotFromWire`): `thread/queueChanged`'s wire params carry only `threadId` (`ThreadQueueChangedParams` has no instance field at all) - the projector sets it from `p.threadID` (`internal/appprojector/appwire_projection.go:1086`), itself set once from the projector's first `event.SessionID` (`appwire_projection.go:265`), never from `evener.InstanceID`. `evener.instanceId` on a `thread/read`/`thread/clear` response is a *different* field: `cmd/evener-hub/internal/appsource/local_daemon.go:1109`'s `firstLocalNonEmpty(entry.InstanceID, threadID)` - a rendezvous `Entry.InstanceID` that can be non-empty and distinct from `threadID`. On a remote hub the divergence is explicit: `remote_hub_refs.go`'s `fromRemoteThread`/`translateThreadRaw` pass `InstanceID` through byte-for-byte as an opaque precondition token, unrelated to the translated `Thread.ID`.

Reproduced in `threads.test.ts` ("queueChanged's threadId identity must not supersede a hydrate's own evener.instanceId for the same live session"): hydrate names the session `instanceId: "inst-1"` (≠ `thread.id: "thr_ref_a"`); a live queueChanged push (unknown coverage, so it retires nothing on its own) names it `threadId: "thr_ref_a"` - read as a brand-new instance, superseding `"inst-1"`; a later authoritative resync, correctly still reporting `instanceId: "inst-1"`, is then rejected outright (`#supersededQueueInstances` already contains it) and a ghost that should have been retired survives forever. Confirmed red.

Fixed: all three snapshot sources (`applyClearResponse`, the hydrate path, and `queueChanged`) now pass `thread.id` (matching what `queueChanged` already, exclusively, carries) - the `evener.instanceId ?? thread.id` fallback is gone. `threadInstanceID()`'s own `instanceId ?? threadId` (the `expectedInstanceId` sent on mutations, a *different* fencing token the remote hub's own server-side check expects verbatim) is untouched - this fix is scoped to the dispatcher's own "which generation is this" question only.

### Medium: unknown-coverage snapshot never advanced the gate - measured, fixed

Round 4's own fix ("unknown coverage must never advance the gate") went one step too far: it also let a delayed KNOWN snapshot at a LOWER revision than an unknown-coverage one already seen act on stale data. Reproduced (`mutationDispatcher.test.ts` "an unknown-coverage snapshot still fences the gate against a delayed, lower-revision known snapshot"): unknown-coverage push at revision 5, then a delayed hydrate at revision 4 (known, empty) - confirmed red, it retired the still-queued intent.

Fixed by tracking `observedRevision` (highest revision merely SEEN, bumped even on unknown coverage) separately from `revision` (highest revision actually RECONCILED, bumped only on success, unchanged from round 4 - this is what keeps the failed-reconcile-retry test green). Gate: `snapshot.revision <= state.revision || snapshot.revision < state.observedRevision` - `<` not `<=` for `observedRevision`, so a known snapshot at the SAME revision an unknown-coverage one already occupied can still replace it (round 4's own "never poisons the gate at the same revision" test stays green, confirmed it still exercises the real code path).

### Consequence, measured: hydration-supersession race in `ensureThread`

Fixing the cut-at-arrival timing above added one microtask hop to `hydrateAndSubscribe` (captureQueueSnapshot's own rehydration), which made a pre-existing race in `ensureThread`/`handleReady` reachable: `emitReady()`'s own `handleReady` discovers a ref with durable pending work and starts a SECOND, targeted `refreshTrackedThread` for it, concurrently with `ensureThread`'s own direct hydration for the same ref. `publishThreadHydration`'s supersession check (`pendingThreadHydrations.get(ref) !== pending`) already existed for exactly this; what was missing was that `publishAndReconcileThreadHydration` returned `null` unconditionally on losing that race, so `ensureThread`'s retry loop saw "no model, but `threads.has(ref)` is already true" and returned - without ever joining the winning attempt's own reconciliation. Broke `threads.test.ts`'s "reload after a failed restart write reconciles persisted uncertainty with the resumed daemon" (a genuine regression, not a flaky test - confirmed the pre-fix code passes it 5/5 in a row, then fails deterministically once the extra hop is added).

Fixed by returning the superseding attempt's own `inflightHydrates` promise instead of `null`, but only when the loss is genuinely a newer-pending-entry supersession (`pendingThreadHydrations.get(ref) !== pending`, checked before `publishThreadHydration` mutates anything) - a stale-epoch or released-ref loss (the OTHER two reasons `publishThreadHydration` returns null) still returns `null` directly, since nothing else is in flight there to join and `inflightHydrates.get(ref)` at that point would be this same call's own not-yet-resolved promise (would have deadlocked "a rejoin adopted across a reconnect" - caught by that existing test, which stayed red until this distinction was added).

### Transition table (unchanged shape from round 4, `observedRevision` added)

State = `{ instance, revision, observedRevision }` per target (`undefined` before the first event; `cut` no longer lives in state - see Part B). `superseded` = the target's own set of dead instance ids.

| Event | Precondition | Next state | Effect on records |
|---|---|---|---|
| hydrate, not authoritative | any | unchanged | none |
| queueChanged / hydrate-authoritative / clear, `instance ∈ superseded` | any | unchanged | none - rejected outright |
| any, `instance ≠ state.instance` (first sight, incl. no prior state) | any | old instance (if any) added to `superseded`; state ← `{ instance: new, revision: -1, observedRevision: -1 }` before the ids branch | none yet |
| …continuing, `revision <= state.revision OR revision < state.observedRevision` | reset above already ran | unchanged | none - stale against either tracker |
| …continuing, `ids === undefined` (unknown coverage) | revision check passed | `observedRevision ← snapshot.revision` only | none - never guesses, but the gate remembers |
| …continuing, `ids` defined (the scan) | revision check passed | `{ instance, revision: snapshot.revision, observedRevision: snapshot.revision }`, recorded only after the writes below succeed | `reconcileIdentities(ids)` settles named ids; scan retires optimistic `turn/queue` records absent from `ids` **and** `intentSequence <= cut` |
| accept (`#attempt`) | - | `#lastAcceptedIntentSequence[target] ← max(current, record.intentSequence)` | none - `intentSequence` was already allocated at enqueue |
| captureQueueSnapshot(target, snapshotWithoutCut) | - | none | mints `{ ...snapshotWithoutCut, cut: #lastAcceptedIntentSequence[target] ?? 0 }`, rehydrating from `listOptimistic()` once if never done |

### Wire shapes touched this round

No new wire field reads. `evener.instanceId`/`thread.id`/`params.threadId` are all non-optional strings on every source (no `omitempty` involved) - this round changes which already-present field is used, not presence/absence handling.

### Part B (simplify, no behaviour change)

1. Dropped the dead `cut` field on `#queueState` (written, never read) and its defending comment.
2. `applyClearResponse` and `handleNotification`'s `queueChanged` branch shared the identical mint→reconcile→conditionally-refresh-pins tail; both now call one `reconcileQueueFromWire(runtime, targetRef, queue, instanceId)`.
3. Consolidated the `cut` and `instanceId` rationale onto one canonical comment each (`QueueSnapshot`'s own, `queueSnapshotFromWire`'s own), one-line cross-references elsewhere.
4. Dropped round-number narration ("round 4", "round 5", "measured on #1705 round 2"), kept every `#1717`/#1705`/`#1706` reference.
5. Replaced `peekAcceptCounter` with `captureQueueSnapshot(targetRef, snapshotWithoutCut)`: the dispatcher mints the full `QueueSnapshot` (including `cut`) at the caller's true arrival, so a caller never threads a bare `cut` number through `queueSnapshotFromWire` by hand. Kept `#supersededQueueInstances` (A → B → C still needs the set).

### Gates

- `npx vitest run src/stores/threads.test.ts src/stores/mutationDispatcher.test.ts src/stores/mutationDispatcher.edge.test.ts src/stores/mutationOutbox.test.ts src/stores/mutationOutboxIndexedDB.stall.test.ts` - 407 tests, all green (run 4x in a row to confirm the timing fix's own stability after the supersession-race find).
- `npm run typecheck` - clean.
- `npm run lint` (`biome ci src ../../../appwire-client/typescript`) - clean, no fixes.

`git diff -M --stat $(git merge-base origin/main HEAD) HEAD`: 5 files, 1230 insertions, 83 deletions (cumulative across all rounds on this branch).

Head: `836d20965a24d5802b0d94b87bebed7e2e8e3c75`, `gh pr view 1705 --json mergeable,mergeStateStatus` → `MERGEABLE` / `BEHIND` (no commits since this branch's base touch `mutationDispatcher.ts`, `mutationOutboxIndexedDB.ts`, `threads.ts`, or their tests - not merged, per the lean-on-CI guidance).

### Residuals

- `#supersededQueueInstances` still grows one entry per clear for the target's lifetime, unbounded in principle (Low, deferred per round 4's own note - unchanged this round).
- The `ensureThread`/`handleReady` double-hydration path itself (one direct, one discovery-triggered) is legitimate by design, but nothing else in the codebase currently exercises the "loser joins the winner's reconciliation" corner this round added - worth a dedicated look if another timing change in this area ever regresses it again.

obra:
Closing per Jesse's ruling (2026-09-17) after the fresh-eyes measurement posted above: this PR is client-side compensation for one absent server field. The daemon already knows exactly which queued client mutation ids a drain consumed (agent/session_client_mutation_queue.go, clientMutationDrain marks each terminal by ClientMutationID) and the queue push carries only survivors. The replacement is a small PR that names the consumed ids on thread/queueChanged so settlement becomes positive evidence, deletes the #1452 sequence heuristic, and later a hydrate-only backstop for the missed-push case. The branch stays for reference; the round-5 tests are the oracle for what the small PR must still guarantee.

PR #1697, opened 2026-09-17
refactor(sdk): directories + MCP settings / launch layer into state/extensions (D12)

obra:
## Summary

D12 of the SDK migration plan (docs/superpowers/plans/2026-09-12-sdk-migration.md, row D12 - depends on D11 #1502, landed as `663756b9086fd71ce5f7694a31b0182b9fbc9678`).

- The dirs and MCP slice of `stores/extensions.ts` becomes `appwire-client/typescript/state/extensions/launchLayer.ts`, the third store under the `@evener/appwire-client/state/extensions` barrel: `createLaunchLayerStore(client)` over `Pick<AppwireClient, "request" | "onNotification">` returns the shared `FrameworkFreeStore` triple plus `start`/`reset`/`dispose`, and holds the ONE `LaunchConfigLayer` the Plugins/Skills directory lists and the MCP config-file/server lists are four fields of - read and written at cwd `/`, layer `global` (per-layer editing stays `launchConfig.ts`'s).
- `fetchLaunchLayer` records failure in state, `setLaunchLayer` rejects and keeps the payload it sent (`setLayer` answers a `LaunchConfigResolved`, not the layer), and both are fenced by `createListRevision` and driven by `createStoreLifecycle` (the `evener/launch/updated` debounce leaves the web file).
- The web store (`cmd/evener-hub/frontend/src/stores/extensions.ts`) stays one zustand store to its four importers (`dirListSetting`, `inrepo`, `mcp`, `launchShared/*`), every one of their tests unchanged; no native twin.
- The three filesystem RPCs are consolidated rather than copied: `evener/path/validate` was already named on `launchConfig.ts`'s gateway (D7), so `evener/dirs/create` and `evener/paths/complete` join it there, and the web adapter's `validatePath`/`createDirectory`/`completePaths` become one-line delegations to `launchConfigStore`.
- Carries D11's `wantsList` reconnect-intent fix forward: `launchLayer.ts`'s own `wantsList` also ORs in `listRevision.hasLive()`, so a `setLaunchLayer` issued before any `fetchLaunchLayer` still recovers correctly on a replaced connection (discovered and fixed while forward-merging D11's round-16 fix onto this branch - the shared `storeLifecycle.test.ts` descriptor suite caught it immediately once `LAUNCH_LAYER` joined the other two cases).

## Test plan
[...]

Issue #1558, opened 2026-09-16
appwire-client: credentials instances store should use the shared notification-refetch lifecycle after D11


obra:
`appwire-client/typescript/state/credentials/instances.ts:294,339,554-566` carries its own `REFETCH_DEBOUNCE_MS` + `refetchTimer` + clearTimeout/setTimeout debounced-refetch pattern, the same shape D10's marketplaces core and D11's plugins core (#1502) copy. D11 is extracting one `createNotificationRefetchLifecycle` helper for the two extensions stores (closing #1503 there); once it lands, move the credentials store's debounce onto it so the package has one encoding of "a notification schedules one refetch".

Found by the /simplify reuse pass on #1502.

obra:
Closed on measurement (2026-09-16), the same rule as #1557. The full move (credentials hands `createStoreLifecycle` a self-rebinding notification port; the helper owns subscription, timer, recovery and the no-op check) measures **net ≈ +21 lines**: −22 from instances.ts, +24 back (the port ~11, the options block, three delegation lines), +19 in storeLifecycle.ts (a public `schedule()` plus a payload generalization: `onNotified` returning a payload, `mergeSchedule`, `refetch(state, payload)`, a type parameter threaded through the helper, the suite and the three stores' `Omit`). The generalization exists solely for credentials: three of four users would pass `void`, and `mergeSchedule` is credentials' own `self && self` provenance rule hoisted verbatim into the shared helper as an option, which shapes the helper around one user.

Three real differences, for the record: (1) provenance through the debounce (`pendingRefetchSelf`, merged non-symmetrically, passed to `readListing(selfRequest)`); (2) three schedule sources, not one (the notification, a landed auth mutation, a superseded `setDefault`); (3) the subscription re-binds per client (`listenTo`), while the helper subscribes once to a fixed port. Only (2) has a cheap standalone fix (`schedule()`), which is not worth a PR on its own.

What replaces it: one parity test pinning the predicate both encodings must share, "recover only an established listing, and only on a transition into ready, including the metadata-only no-op" (instances.ts:1009 vs storeLifecycle's no-op check + `established(state)`), ~40 test lines against both cores, landing after D11 (#1502) merges.

PR #1707, opened 2026-09-17
test(sdk): pin the reconnect predicate both state layers encode

obra:
## Summary

Follow-up from #1558 (found by the /simplify reuse pass on D11 #1502): the credentials core and the extensions stores (D10/D11/D12) each decide independently, in their own `connectionChanged`, whether a reconnect should recover a list - "a list nothing has read stays unread, a list a host has already read is read again when the connection is ready again." #1558 asked whether the credentials core's debounce should move onto the extensions stores' shared `createStoreLifecycle`; measured at +21 lines and a generalization of a helper written for a single user, which isn't worth it for one caller.

So the agreement between the two independently-written implementations is pinned with a test instead of merged into one. `appwire-client/typescript/state/reconnectPredicate.test.ts` runs both cores (the credentials core's own `connectionChanged` and the extensions stores' shared `storeLifecycle.ts`) through the same four connection scripts - a list nothing has read, a list read then dropped and restored, a metadata-only update that changes neither client nor state, and a replacement client - and asserts they reach the same decision every time. Dropping either side's guard fails it, so the two stay honest about agreeing without a shared implementation forcing it.

No production code changes - this branch is purely the test file, carried forward through D11 (#1502) and D12 (#1697) as they landed underneath it.

## Test plan
- [x] `npx vitest run appwire-client/typescript/state/reconnectPredicate.test.ts` from `cmd/evener-hub/frontend` - 8/8 pass
- [x] `npm run typecheck` from `cmd/evener-hub/frontend` - clean
- [x] `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend` - clean
- [x] `make lint-package-imports` - PASS
- [ ] CI


obra:
## Review round disposition (573440365)

The one finding (describe.each rows without `as const` fail `tsc` with TS2349) is refuted by measurement, not by argument:

- The frontend typecheck program includes this file: `cmd/evener-hub/frontend/tsconfig.json` `include` lists `../../../appwire-client/typescript`.
- CI run 35258569968 at this exact head ran that program and passed: job `web` step `web-typecheck` printed `PASS`, and the build step ran `tsc --noEmit --incremental false && vite build` green.

Vitest's `each` overload constrains its row type with `any[] | [any]`, which makes TypeScript infer each literal row as a tuple, so `build` is `() => Encoding` and callable. The sibling suites' `as const` is a style choice, not a requirement. No change.

Issue #1642, opened 2026-09-17
sdk: hasBeenReady (store meets a reconnecting client) has no cross-host test


obra:
Residual from D11 #1502 round 10 (2026-09-16).

`storeLifecycle.ts`'s `hasBeenReady` rule ("an observed `reconnecting` is evidence of a previous ready; `closed` is not, because a failed first connection also ends there") lives in the shared lifecycle but is pinned through the plugins store only. The marketplaces store and the launch layer inherit it untested because the suite has no generic accessor for "what a notification invalidates", and the reconnect parity test does not cover it (credentials has no equivalent).

Add the cross-host case: for each store the lifecycle drives, initialize against a client already `reconnecting`, then ready, and assert the revision advances and the recovery read fires; and the converse for `closed`. Depends on the descriptor/test-kit shape (see the sibling issue on the lifecycle test kit).

obra:
Note from D11 #1502 round 14: `hasBeenReady` and the new latched `wanted` intent are pinned only through the plugins/marketplaces descriptors, not the reconnect parity test (credentials has no equivalent of either yet). Cover both when #1681 migrates the credentials store onto the lifecycle.

obra:
Measured 2026-09-17 by the implementer lane before writing the case: the issue's framing is partly wrong. In storeLifecycle.ts hasBeenReady only gates onNotified (the invalidation); the recovery read is gated by wantsList alone, so closed -> ready still fires the recovery read for an established list (spike: connectionChanged closed, fetch, ready: revision stayed 0 but the list call count went 1 -> 2). So the converse case for closed is 'no invalidation', not 'no recovery read'. Asserting the invalidation generically needs a descriptor hook for 'what a notification invalidates' (that is #1641's test kit), and launchLayer.ts passes no onNotified at all, so the rule is unobservable there by any state assertion. Blocked on #1641. A separate small case is being added now: an established list is recovered after closed exactly as after reconnecting, using existing descriptor fields, since no lifecycle test exercises closed at all today.

Issue #1641, opened 2026-09-17
sdk: extract a per-store test kit for the store-lifecycle suite (descriptor at 12 capabilities)


obra:
Residual from D11 #1502 round 10 (2026-09-16).

The shared store-lifecycle test suite in `appwire-client/typescript/state/extensions/` drives three stores (plugins, marketplaces, launch layer) through one descriptor that now exposes 12 capabilities, four of them added in rounds 9 and 10 purely to express request orderings (publish/retract/fence cases). It still reads concretely, but the next ordering case would make it 13, and the descriptor is already the largest thing in the suite.

Extract a small per-store test kit (build the store, issue a gated read, issue a gated write, reject a request, fence) so each ordering case is written once against the kit rather than widening the descriptor interface. ~150 lines, test-only, no behavior change; the current cases are the oracles.

obra:
Taken on #1502 (round 13's simplify pass), in the smaller shape this issue asked for rather than a per-store test kit.

`testing/fakeClient.ts` gains the generic request scripts a parametrized suite needs - `answerRequests`, `failRequests`, `gateRequests`, `gateFailures`, `callsTo` - each taking the method name and carrying the one cast that per-method typing needs. The lifecycle descriptor's six request-scripting closures per store become four data fields: `listMethod`, `listResponse`, `mutationMethod`, `mutationResponse`.

The two notification closures stay closures on purpose. `emitNotification` type-checks its payload per notification, which is what caught a wrong `evener/launch/updated` params shape while D12 was being written; a generic emitter would have to cast that away, and the descriptor would get shorter by trading a real check for it.

So the descriptor is now data plus the store's own accessors (`create`, `fetch`, `mutate`, `loading`, `list`, `initial`, `debounceMs`) and the two typed emitters. If a future row needs a thirteenth capability it is worth re-reading this issue, but the pressure that prompted it is gone.

obra:
Done in #1714 (merged as a448fdf41): createLifecycleKit with five primitives replaces the hand-rolled gate choreography in the ordering tests; the descriptor gained only the test-only invalidationMarker field for #1642.

PR #1714, opened 2026-09-17
sdk: extract a per-store test kit for the store-lifecycle suite

obra:
## Summary

Extracts a small, test-only kit for `appwire-client/typescript/state/extensions/storeLifecycle.test.ts` per #1641, and adds #1642's blocked case now that the kit can carry it.

### #1641: the kit

`LifecycleCase`'s four wire-level fields (`listMethod`/`listResponse`/`mutationMethod`/`mutationResponse`) existed only so each ordering test could gate a request and release it later - a pattern hand-rolled with `gateRequests`/`gateFailures`/`answerOne`/`failOne` across ~13 test bodies. `createLifecycleKit` collects that into five primitives:

- `{ fake, store }` - build the store
- `gatedRead()` - issue a gated read, `land(response?)` it later
- `gatedWrite()` - issue a gated write, `land(response?)` or `fail(message?)` it later
- `rejectWrite(message)` - issue a write against a mutation scripted to fail immediately
- `fence(...states)` - replace the connection with a fresh client, driving it through each state in order

Every ordering test that used to hand-roll `gateRequests(fake, lifecycle.listMethod)` + `lifecycle.fetch(...)` + manual resolver bookkeeping now destructures these off `createLifecycleKit(lifecycle)` and calls `gatedRead()`/`gatedWrite()`/`fence()` directly instead. A future ordering case is written against these, not by widening the descriptor again.

**Descriptor capability list, before** (13 fields on `LifecycleCase`): `create`, `debounceMs`, `notifyUpdated`, `notifyUnrelated`, `listMethod`, `listResponse`, `mutationMethod`, `mutationResponse`, `list`, `fetch`, `mutate`, `loading`, `initial`.

**After** (same 13, plus one for #1642 below): unchanged, + `invalidationMarker?`. The four wire-level fields stay on the descriptor - `createLifecycleKit`'s internals still need them to drive the fake, and ~15 sites still read `lifecycle.listMethod`/`lifecycle.listResponse` directly for `callsTo`/`answerRequests` calls the kit doesn't cover (including the new #1642 test). What the kit removes is only the *gate-and-release* choreography (`gateRequests`/`gateFailures` + manual resolver arrays) that used to be hand-rolled per ordering test - that's the part a new ordering case no longer has to re-derive or add a field for.

No behavior change: every assertion the replaced tests made is still made in the same test. Two array-length sanity checks became equivalent call-count checks, since the kit doesn't hand back raw resolver arrays:
- `expect(releases).toHaveLength(2)` -> `expect(callsTo(fake, lifecycle.listMethod)).toBe(2)` ("a read outrun before it lands..." test)
- `expect(fails).toHaveLength(2)` -> `expect(callsTo(fake, lifecycle.mutationMethod)).toBe(2)` (parametrized "older/newer of two failed writes..." test)

No case was folded/deleted - every existing test still exists, unchanged in name and assertions, just rewritten against the kit where it used to hand-roll gating.

### #1642: closed -> ready invalidation, for plugins and marketplaces

Measured (see #1642's comment thread): `hasBeenReady` only gates `onNotified` (the invalidation) - the recovery *read* is gated by `wantsList` alone, so the existing "closed -> ready recovers an established list..." test already covers the read regardless of whether the store was ever ready before. What's new here is the *invalidation* half: a store that WAS ready, then closed, then ready again, must invalidate whatever a notification would have (a revision it bumps, a cache it retires) - storeLifecycle.ts's `away` branch.

The kit made this cheap to observe: `invalidationMarker` reads back state each store's own `onNotified` already touches -  `pluginRevision` for plugins, `browseCatalogs` for marketplaces (its `Map` identity changes on every retirement, even an empty one) - so the new case costs no production code. `launchLayer`'s lifecycle passes no `onNotified` at all, so it gets no case; the descriptor field is `undefined` for it and the suite skips generating the test.

### Simplify pass (opus review)

- Destructure `createLifecycleKit`'s methods directly at each call site (`const { fake, store, gatedRead, fence } = createLifecycleKit(lifecycle)`) instead of keeping a `kit` local and calling `kit.gatedRead()` etc - no kit method closes over `this`.
- Dropped the two-failed-writes test's unreachable "both writes must be in flight" throw (`gatedWrite` already throws if a write never reaches the fake) and its `index === 0 ? first : second` ternary, replaced by indexing a fixed two-tuple.
- Inlined `FENCE_TO_READY` - a one-element array used once didn't earn a module const.
- Made `rejectWrite`'s `message` parameter required - its one caller always passes a literal, so the default was dead. `fail()`'s default stays, since it's actually used.

No assertion changes in the simplify pass either.

## Gates run locally
- `npx vitest run` on `storeLifecycle.test.ts` + `plugins.test.ts` + `marketplaces.test.ts` + `launchLayer.test.ts`: 125 passed (76 -> 78 in `storeLifecycle.test.ts` alone: +2 for the new #1642 case on plugins and marketplaces)

Refs #1641, #1642


obra:
## Simplify disposition (226941091)

Applied all four opus simplify findings in one push, test-only, no assertion changes:

1. `storeLifecycle.test.ts` (two-failed-writes parametrized test): replaced `const [first, second] = [await kit.gatedWrite(), await kit.gatedWrite()]` + the unreachable `if (!first || !second) throw` with two separate awaited consts (`gatedWrite` already throws loudly if a write never reached the fake) and `const writes = [first, second] as const; for (const index of order) writes[index].fail();`, dropping the `index === 0 ? first : second` ternary.
2. Destructured `createLifecycleKit`'s methods directly at all 13 call sites (`const { fake, store, gatedRead, ... } = createLifecycleKit(lifecycle)`) instead of keeping a `kit` local - no kit method uses `this`.
3. Deleted the `FENCE_TO_READY` module const and its doc comment; `fence` now inlines `(["ready"] as const)`.
4. Made `rejectWrite`'s `message` parameter required (its one caller always passes the literal `"write refused"`, so the default was dead code). Left `fail()`'s default, which two call sites rely on.

Also softened the PR body's claim about the wire-level descriptor fields: `lifecycle.listMethod`/`lifecycle.listResponse` are still read directly at ~15 `callsTo`/`answerRequests` sites including the new #1642 test - the kit only removed the gate-and-release choreography, not all direct use of those fields.

Skipped, per the coordinator's disposition (recorded, not mine to do):
- a shared `gateSettlements` helper in `testing/fakeClient.ts` subsuming `gateRequests`/`gateFailures` (outside this diff)
- `invalidationMarker` staying a test-only descriptor field
- the `launchLayer` conditional staying as-is

Gates: `npx vitest run` on `storeLifecycle.test.ts` + `plugins.test.ts` + `marketplaces.test.ts` + `launchLayer.test.ts` - 125 passed. `npm run typecheck` - clean. `npx biome ci ../../../appwire-client/typescript` and `npm run lint` - clean.

Head: `2269410913fc0b62572f520e01a03961683f3837`, `gh pr view 1714 --json mergeable` reads `MERGEABLE`.

PR #1724, opened 2026-09-17
ci: retry snapshot upload and fence it against reruns

### LINKED ISSUE #1712: ci: snapshot release upload fails on a transient GitHub 500 and reruns race the next publish
## What happened

Build Binaries on main at ce484d9df (2026-09-17, run 35258264514) built every artifact and then failed only in "Refresh snapshot channel" with `HTTP 500: Error saving asset` from uploads.github.com on `gh release upload snapshot dist/*.tar.gz dist/checksums.txt --clobber`. A `gh run rerun --failed` of that job failed again with `HTTP 400` on the darwin asset while the next main commit's run (78a822b6d) was uploading the same assets with `--clobber` concurrently; that newer run succeeded, so the channel is current.

## Cost

A transient GitHub upload error turns a green main into a red Build Binaries badge, and reruns race the next commit's publish over the same release assets.

## Proposed change

Small: retry the upload step up to three times with a short sleep (transient 5xx), and make the job's concurrency group cancel an older in-flight publish when a newer main run starts, so a rerun can never race a newer publish. About ten lines in the workflow. Not a supervisor, not a script.

obra:
## Summary
- Retry the "Refresh snapshot channel" job's `gh release upload` up to three times with a short sleep, so a transient GitHub 500/400 doesn't redden the Build Binaries badge.
- Add a cancel-in-progress concurrency group to that job so an older rerun can never race a newer main commit's publish over the same release assets.

Closes #1712

## Test plan
- [x] `.github/workflows/binaries.yml` parses as YAML (`ruby -ryaml -e "YAML.load_file(...)"`)
- [ ] CI green on this PR (Build Binaries workflow)


obra:
## Review round disposition (a4cb41b8f)

Medium: `cancel-in-progress: true` on the snapshot job — fixed by flipping to `cancel-in-progress: false`.

RoboRev is right and my original choice was wrong. The upload step writes multiple assets (tarballs + checksums.txt) in one `gh release upload`, after the snapshot tag has already been force-moved to the new commit. If a newer main run cancels an in-flight publish mid-upload, the release ends up with some assets from the cancelled build and some from whichever build finishes next — a mismatched set that fails install.sh's checksum verification. Cancellation also kills the retry loop itself, so a run that gets cancelled mid-retry never reports the failure either.

`ios-testflight.yml` already avoids this: it uses `cancel-in-progress: false` for the same reason (its comment: builds must run to completion). Queueing behind the in-flight publish instead means each run still completes as a whole (tag move + full upload, with its own retries), and whichever one finishes last leaves the channel in a fully consistent state — which is exactly the "newer supersedes older" property #1712 asked for, just achieved by queuing to completion rather than by interrupting a partial write.

obra:
## Review round disposition (ee125b1a2)

High: the `snapshot-publish` concurrency group gives mutual exclusion, not ordering — fixed by adding a freshness check before the tag move.

RoboRev is right. `concurrency` only guarantees at most one run holds the group at a time; it says nothing about the order in which queued runs acquire it. A run for an older commit (queued behind a newer one, or a manual rerun of an old failed run) can still take the group after the newer commit's run has already published, so it would move the `snapshot` tag backward and clobber the newer, already-published assets with older ones — silently regressing the channel.

Fix: a new step ("Check this commit is still main's head") runs before the tag move, does `git fetch --depth=1 origin main` and compares `FETCH_HEAD` against `$GITHUB_SHA`. If this commit is no longer main's current tip, it logs a line and sets a `stale=true` step output; the "Move snapshot tag" and "Publish snapshot release" steps both gate on `steps.freshness.outputs.stale != 'true'` and skip (job still reports success, no red badge for a superseded run). I used `FETCH_HEAD` rather than `origin/main` because `actions/checkout@v7`'s default shallow, single-commit checkout does not populate a resolvable `refs/remotes/origin/main` — I verified this by reproducing the same shallow single-SHA fetch/checkout locally; `git fetch <repo> <ref>` always sets `FETCH_HEAD` regardless of remote-tracking config, so it doesn't depend on that.

This keeps the retry loop (transient upload errors) and the queueing concurrency group (`cancel-in-progress: false`, so no partial multi-asset writes) from the prior rounds, and adds the missing ordering guarantee on top: whichever run publishes now always does so only when it is main's current head, so an older run can never overwrite a newer publish.

Gate: YAML re-verified with `ruby -ryaml -e "YAML.load_file(...)"` — OK.

PR #1725, opened 2026-09-17
A UI-credentialed instance is the user's own, and OpenAI Codex joins the front page

obra:
# What

Two changes to how the hub and its clients treat provider instances.

**A UI-credentialed instance is the user's own.** A curated provider's instance exists while a credential resolves for it, and the removal gate used to refuse removing such a row outright — so a user who had pasted a key or signed in through the credentials pane could not remove the account they had created: the credential was deleted and the row came back, or the removal was refused with a remedy naming something that did not exist. The rule now distinguishes where the credential comes from: `environmentBacked` (hub) and `fromEnvironment` (AppWire client) agree that an API-key variable, the host's ADC credentials and a keyless endpoint belong to the host — those are refused, with the action that actually takes them away named in the message (unset the variable, clear the credential) — while a stored key or a signed-in OAuth record is the user's and goes with its row. The rename rule follows from the same fact: an implicit instance nothing authors can be renamed, an environment-backed row cannot.

**OpenAI Codex joins the front page.** The curated Codex provider is offered in the credentials pane's popular grid under its own label, with its OAuth sign-in flow, and the registry spec's rename and removal sections are updated to match the rule above.

# Where the two sides agree

`fromEnvironment` in `appwire-client/typescript/credentialLabels.ts` is the client's copy of the same predicate, and both sides are one allow-list rather than a rule plus a provider-keyed exception: `env:*` and `adc` name the host's credential, everything else is the user's. The Codex transport needs no case of its own, because its instance resolves `oauth` when its record is readable and no usable source when it is not, and neither is on the allow-list, so a broken sign-in stays the user's to remove. The AppWire package's wire test rebinds the hub's own recorded Codex row over every source the corpus sends, so a change to either side fails there rather than drifting.

The rename that stands in `providers.toml` but cannot carry the instance's credentials is announced through `ErrorInstanceRenamePersisted` (added to `appwire/errors.go` here, since the hub did not have it), and the clients that report a standing rename read that one discriminator.

# Scope

This is the feature half of a larger branch, split out so it can land on its own. It carries the predicate on both sides, the removal gate, the rename unlock, the wire discriminators, the credentials panes, native, the spec docs, and the rename-persisted outcome they share.

It deliberately does **not** carry the OAuth aside and crash-recovery subsystem, which is the other half's subject: that work adds the removal's aside reclaim (the thing that makes a *removal* persist), and with it the client handling for a persisted removal, the `deleteAside` test seam, and the recovery pass. None of that is reachable from this branch, so none of it is here.

# Verification

- `go test ./cmd/evener-hub/... ./appwire/... ./cmd/evener-tui/... ./server/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`
- `make test-web` (typecheck + vitest + Biome), `make lint-package-imports`
- native: `npm run check`, `npm test`

### Round: one allow-list on both sides, and the dead half of the wire contract goes

- **The Codex exemption is gone from both predicates.** It was added when both were deny-lists, where any source that was not `store` or `oauth` counted as the environment's, so a Codex row resolving no usable source (an unreadable record) would have been offered Remove in the UI and refused by the hub. The allow-list subsumes it: every source a Codex instance can resolve, `oauth` when its record reads and none when it does not, is already the user's, and no hub row can make the exemption's branch fire. One rule in two languages replaces a rule plus a provider-keyed special case, with the classification pinned source by source on both sides; the AppWire wire test needed no change.
- **The dead contract.** `ErrorInstanceRemovePersisted` and its constructor were exported and documented as if the hub emitted them, but nothing on this branch produces a persisted removal: that half of the contract arrives with the pull request that adds the removal's aside reclaim. It goes, and the rename half stays, because this branch does emit it.
- **A four-angle simplify pass over this branch**, behaviour-neutral and proven by the suite (no test added, removed or weakened): the pane's private persisted-rename predicate replaced by the one the AppWire package now exports; the sheet's rename guard collapsed to the single condition it always was; a three-level nested ternary flattened with byte-identical strings; two test helpers deduplicated; and the keyless-auth predicate called at the two places that spelled it inline.

Verification for this round, on `627d0bade581dd22b74c4a1e0ce05ca3adef5a9e`: `go test ./cmd/evener-hub/... ./appwire/... ./cmd/evener-tui/... ./server/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`, `make test-web` (typecheck + vitest + Biome), `make lint-package-imports`, and native's `npm run check` and `npm test`.

### Round: the review's classification race, and its two lows

- **The classification race.** `Remove` asked the ownership question (is this row the user's to remove?) under the instances lock, before the credential lock. A credential writer that held the credential lock first can change what the instance resolves: clearing a stored key on an instance whose provider has an environment fallback turns a row the user owns into one the environment supplies, and the removal then reported success while the row remained. The question is now re-asked immediately after the credential lock is taken, existence first and then environment-backed, returning the same refusals the first check gives; the lock order is unchanged. The regression test holds a credential clear inside its seam while the removal classifies and blocks on the lock, releases it, and asserts the removal is refused and the row the environment supplies is back.
- **The discriminator every client keys on.** The persisted-rename RPC tests asserted only the message, which is identical whether or not `instanceRenameError` wraps the error in the `evenerErrorInfo` payload; a reverted mapping would have passed every Go and TS test while the web sheet silently treated a standing rename as a failed save. Both tests now decode the rejection's `ErrorData` and assert `ErrorInstanceRenamePersisted`, the way the keybindings tests pin their own discriminator.
- **The rename note tells the whole story.** The sheet's note read only the active source, so a stored key that shadows a set environment variable was described as an ordinary rename; after the user layer moves, the environment supplies the old name and the old instance does remain. The signal was already on the wire (`shadowedEnvVar`), so the rule now lives beside `fromEnvironment` as `renameLeavesEnvironmentRow`, used by the sheet and pinned in both the pane and predicate suites.
- **Answered, not changed.** The review's other medium wanted implicit rows at sources `api_key` and `credential_headers` classified as the environment's. No curated record carries an inline credential (0 of 219), and the catalog conversion sets only `api_key_env`, so an implicit row cannot resolve either source; classifying them would be dead code on both sides. The reasoning, with the checks behind it, is on the pull request.

Verification for this round, on `fc236d47c67eba2b0bb11f3244b92d7d114650ec`: `go test ./cmd/evener-hub/... ./appwire/... ./cmd/evener-tui/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`, `make test-web` (typecheck + vitest + Biome), `make lint-package-imports`. Each finding's guard was proven red without its fix: the race test against the un-re-checked removal, the discriminating assertions against a reverted mapping, and the note against the active-source-only predicate.

### Round: the review's two mediums and three lows

- **The rename note.** `renameLeavesEnvironmentRow` missed two ways an old row survives a rename: an authored entry that shadows a curated implicit provider (the name is the curated id), where the environment re-derives the row once the entry moves, and an implicit `gcp-adc` row whose ADC file is outranked by a stored credential JSON, where `shadowedEnvVar` is deliberately empty. The predicate now covers both, and its comment claims only what it enforces.
- **The removal confirmation.** The pane treated `implicit` as "environment-backed", which it no longer is: a stored key and a signed-in Codex record are implicit rows a removal does delete, so a superseded removal could confirm against a stale listing and report a surviving stored-key row as environment access. Both checks now ask `fromEnvironment`.
- **The lows.** `renamedInstanceLanded` rejects any implicit landing, because every successful rename authors the new name; the credential-only rollback returns restoration success and only reports a rollback, and retries the reload, when every deleted credential was put back, reporting the standing removal and what was not restored otherwise; and the environment-backed refusal's remedy is chosen from the source (unset the variable, remove the ADC credentials, clear the stored key, or remove the OAuth record) instead of always naming the OAuth record.

Verification for this round, on `277b90265559034bf30ace45e61f88ef293dd2d3`: `go test ./cmd/evener-hub/... ./appwire/... ./cmd/evener-tui/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`, biome on the touched frontend files, `make lint-package-imports`; the web suite runs in CI on this head. Each guard was proven red without its fix: the two note cases, the confirmation and still-supplied branches, the implicit landing, the failed restore that must not claim a rollback, and one test per remedy.

### Round: the review's two mediums, and a hub-computed rename note

- **The locked refusal's wire class.** The under-lock re-check of the environment-backed refusal returned a plain error while the pre-lock check returned `InvalidParams`, so the same caller-fixable condition ("unset X instead") reached a programmatic caller as an internal fault whenever a concurrent credential clear landed first. It returns `appwire.InvalidParams` now, and the race test asserts the wire code rather than only the message.
- **The rename a failed rollback left behind.** A rename whose initial reload failed and whose config rollback also failed came back as a generic write-applied error, so the client was told the save failed although `providers.toml` carries the new name. Edit's rename call site now wraps a write-applied failure in `writeApplied(renamePersistedError{...})`: the marking the handlers broadcast on, and the type the RPC handler maps to `ErrorInstanceRenamePersisted`. An RPC test drives that path end to end.
- **The rename note stops guessing.** The predicate could not know whether an ADC file exists or whether the old name is a curated provider id, and it was wrong both ways. The hub computes it now - true when the instance is environment-backed, or when its name is a curated provider id that re-derives an instance from layers a rename cannot move - and the listing carries it as `renameLeavesRow`; the predicate reads that bit, and the generated TypeScript types, the recorded wire fixture, and the protocol doc carry it.

Verification for this round, on `107074cfeb46373ee43d7a5f140abad3f504666b`: `go test ./cmd/evener-hub/... ./appwire/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`, `make lint-generated`, biome on the touched frontend files, `make lint-package-imports`; the web suite runs in CI on this head. Each guard was proven red without its fix: the race test's wire-code assertion, the RPC rename-persisted path, the hub-side `renameLeavesRow` cases, and the client predicate in both directions.

### Round: the TUI asserts the endpoint, the rollback follows the carrying layer, and the helper respects the predicate

- **The TUI's removal asserts the endpoint now.** It sent only the instance name, so a stale screen could delete a replacement instance after another client re-pointed the name; nothing in the TUI carried the fingerprint at all. `InstanceRemoveMsg` carries the shown entry's `EndpointFingerprint`, the command sends it, and the panel passes it through, with tests for the shown and the stale case. The edit path has the same gap (`InstanceEditParams` has no fingerprint and the hub does not verify one); that is reported on the pull request rather than widened into this round.
- **The rollback follows the layer that carries the instance.** With a stored key beside a Codex OAuth record, a restore that brought the carrying layer back but failed on the stray other one answered "not restored", so the caller reported the removal as standing and skipped the reload while the instance was configured again. `restoreFailedRemoval` decides by the carrying layer now (`supplyOf`: the OAuth record for the oauth source, the stored key for store, the environment otherwise), marks the applied-write case only when that layer stays deleted, and names the stray leftovers it could not put back.
- **The rename helper classifies by the predicate's vocabulary.** `ProviderRenameLeavesInstance` answered true for a non-implicit curated provider and for a present inline `api_key` or `credential_headers` expression, so the note could promise a row that never re-derives, and it disagreed with the ownership predicate. It requires the record to be implicit now (mirroring `computeInstances`), treats a present inline expression as terminal the way `credential()` does, and never counts those sources as re-derivations; no curated record carries either (219 checked, and the conversion sets only `api_key_env`).

Verification for this round, on `267791555fa73d88ad4aa134aba272bde43cef83`: `go test ./cmd/evener-hub/... ./cmd/evener-tui/... ./llm/registry/... ./appwire/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`, `make lint-generated`. Each guard was proven red without its fix: the TUI fingerprint in both directions, the carrying-layer rollback and its inverse, and the helper's implicit and precedence cases.

### Round: the tagged floor, and a partial rollback that still speaks

- **The tagged build compiles again.** The `evenerfuzz`-guarded exercise list still called the TUI's removal helper with the old two-argument form, so tagged builds did not compile, and neither the default test set nor the CI lanes on that head compiled it. The caller passes the fingerprint now, and this round's pre-push chain runs `make lint-evenerfuzz` (the tagged-source compile floor) so the class is caught here next time.
- **A partial rollback still tells the clients.** When a credential-only removal's reload failed and the carrying credential was restored while a stray one stayed deleted, the rollback went back unmarked, so `instanceWrite` skipped the broadcast and every other client kept stale credential status for the deleted layer. The leftover error carries the applied marker now: "the instance is back" (the carrying layer, by `supplyOf`) decides the rollback message and the recovery reload, and "a credential is gone" (the leftover layer) decides the marker and the broadcast. An RPC test drives the partial rollback and asserts the update notice arrives.

Verification for this round, on `458a852c5e9567a9b45f8bb0c276ce80f86915ab`: `go test ./cmd/evener-hub/... ./cmd/evener-tui/... ./llm/registry/... ./appwire/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`, `make lint-generated`, `make lint-evenerfuzz`. Both guards were proven red without their fixes: reverting the tagged call fails to compile, and removing the marker fails the carrying-layer test and the RPC broadcast test.

### Round: a rename reports both halves, and the race test stops sleeping

- **A rename reports both halves.** When the credential move failed and the reload failed with it, the caller heard only the move, so nothing said the registry may still list the old name and refuse instance writes. The reload is switched on three ways now: a clean reload returns the move's error; a clean move with a failed reload keeps its rename-persisted report; and both failing folds the reload failure around the move's error, which already carries the applied marker and the `renamePersistedError` type, so the broadcast and the client's discriminator both survive without a second marker.
- **The race test stops sleeping.** It slept 100 ms hoping the removal had reached the credential lock, which proved nothing about what it was racing. A test-only seam signals just before the lock is taken, and the test waits on it: with the locked re-check disabled, all twenty runs fail, and the restored test is green under `-race` at `-count=50`.

Verification for this round, on `fdb9ae347f4816b731cd575f72ff27ecb75dfe6e`: `go test ./cmd/evener-hub/... ./cmd/evener-tui/... ./llm/registry/... ./appwire/... -count=1`, `make vet`, `make lint-golangci`, `gofmt`, `make lint-generated`, `make lint-evenerfuzz`. Both guards were proven red without their fixes: reverting the both-failed branch drops the reload failure from the message, and disabling the locked re-check fails the barrier-driven race test on every run.

### Round: the sheet follows the persisted rename, and clear flags map explicitly

- **The sheet follows a persisted rename now.** It only steered to the new name when the refreshed listing showed the new row, so when the hub's registry was left on its fallback listing (the both-failed rename path the hub reports as persisted), the user got a form error for a rename that is on disk. `ErrorInstanceRenamePersisted` is authoritative: the sheet steers to the new name, keeps reconciling the listing, and still warns about what was left behind.
- **Clear flags map explicitly.** The low's premise did not hold - the param is `clearApiKeyEnv` with a lowercase "pi" in both `appwire/types.go` and the generated types, so the existing one-letter conversion was already correct - but the explicit `CLEAR_FIELD_NAMES` map lands anyway, so a future capitalised field name cannot repeat the class. The evidence is in a comment on the pull request.

Verification for this round, on `d43c2b8697815e9b5c17ba7154e315df79fc3f09`: biome on the touched files, `make test-web` (typecheck + vitest + Biome), `make lint-package-imports`. The steer guard was proven red without its fix, and the new clear-flag tests pin the mapping flag by flag.

### Round: the persisted rename keeps the guard

- **The sheet follows the persisted rename now, correctly.** The previous round steered to the new name but cleared the rename guard first, so when the refreshed listing omitted the new row - the case the branch exists for - the sheet's held instance vanished, the close effect fired, and the user saw only the warning while the sheet shut on a rename that is on disk. Dropping that clear is necessary but not sufficient, as the fix found by testing the review's suggestion exactly: the release effect dropped the held entry on any name change. It is conditional now - the entry is released once the listing can carry the selected name (or the sheet is dismissed), and a rename still in flight under its old name keeps it.
- **And the test that missed it is replaced.** The previous round's test passed because the steer callback is a spy, so the name never actually changed and the real section's re-selection never ran. An integration-style test drives the real section with a listing that omits the new row and then gains it, asserting the sheet stays open on the renamed instance throughout; both the review's suggested fix alone and the full pre-fix state fail it.

Verification for this round, on `2d3e0fbba7c42420700ecbdfb9da901db2c36e50`: `make test-web` (typecheck + vitest + Biome), `make lint-package-imports`, and the focused suites (428 tests). The guard's mutation evidence is above.

### Round: the rename keeps its held subject until its own row lands

- **A stranger at the destination no longer becomes the sheet's subject.** The previous round's release effect dropped the held entry as soon as any listed row carried the selected name, and `stored ?? renamingFrom` adopted that row. The destination name is not reserved - a rename frees the old name, the hub's registry can re-derive an implicit row there, or another instance can take it - so a row that is not this rename's own could become what the sheet showed and edited. The sheet now holds the request that carried its destination alongside the entry, and both the subject and the release answer to the check the rest of the sheet already makes for a landed rename (`renamedInstanceLanded`): while the section is on this rename's destination and the listing's row there is not this rename's own, the held entry stays the subject, and the release waits for the expected row.
- **Covered at both kinds of destination row.** A persisted rename that meets an implicit impostor row at the destination keeps the held original and follows the real row once it arrives; a second case does the same for a different authored row wearing the freed name. Both go red under each half of the fix (`CredentialsSection.test.tsx:2369` and `:2418`), the mutation evidence.

Verification for this round, on `c1a661d7b7d38117e23f39c69240b9ece64fec0f`: `make test-web` (typecheck + vitest + Biome), `make lint-package-imports`, and the focused credentials suites (15 files, 430 tests).

### Round: the unverified rename gap refuses writes, and the keyless remedy names its endpoint

- **The gap was readable but writable.** The previous round kept the sheet on the held original while the destination row was unverified, but left every write path open. Save could resubmit the completed rename against the name the instance no longer has, and every action the sheet forwards targets `selectedInstance` and asserts `findInstance(selectedInstance)`'s endpoint fingerprint - with an impostor row at the destination that assertion authorizes mutating the wrong instance, or deleting it, while the sheet shows the original. The gap is read-only now: Save refuses with a form error (through the button and the form's own submit door), every other action refuses with a warning, and no confirm dialog opens. The instant the rename's own row arrives, all of it works again.
- **And the keyless removal remedy names what actually removes the row.** For a keyless optional-bearer instance whose active source is a variable, the refusal advised unsetting that variable. A keyless scheme resolves without any credential, so the row is derived whether or not the variable is set and survives the unset - `TestInstances_KeylessRowOutlivesUnsettingItsOptionalKey` pins that premise. `removalRemedy` names the provider endpoint for that case and keeps the variable first for credential-required schemes, which is why the keyless `OLLAMA_API_KEY` expectation moves.
- **Tested at the doors.** A second Save during the gap sends no second edit request; Remove and Clear during an impostor gap open no dialog and send no RPC, then work once the real row arrives (`CredentialsSection.test.tsx:2460`/`:2506` were red before the guards); the keyless premise and the refusal tables gain cases in `app_instances_test.go`.

Verification for this round, on `ef1d1c417e44f49bee0799aab6eaa4433215655d`: biome, `npm run typecheck`, the focused credentials suites (15 files, 432 tests), `make lint-package-imports`, `make test-web`, `gofmt`, `make lint-generated`, `make vet`, `make lint-evenerfuzz`, `go test ./cmd/evener-hub/...`.

### Round: an applied removal reconciles in the client, and the Codex card agrees with its guide

- **A removal that applied before it failed is reconciled, not reported as a failure.** The hub marked those removals already, but the mark only decided the broadcast: the caller's error kept its ordinary wire class, so the section showed "Remove failed", left the confirmation open, and never told the guided owner, whose retry then targeted an instance that is already gone. The removal now has the rename's sibling discriminator — `removeAppliedError` at the two removal sites (a config rollback that could not be written, a credential deletion that could not be restored), `ErrorInstanceRemoveApplied` on the wire, and the client's `isInstanceRemoveApplied` — and the section closes the confirmation, re-reads the listing, notifies `onInstanceRemoved`, and surfaces the hub's message as a warning.
- **The Codex card's billing line renders, and the guide names the card.** The help block sat entirely inside the apiKey branch while Codex is OAuth-only, so its billing line never appeared; `billing` now renders whenever help exists and only the key link stays gated. The connecting-a-provider guide sends readers to **OpenAI Codex** rather than the retired "ChatGPT / Codex" phrasing.
- **Tested at each door.** The client path (confirmation closes, owner notified, listing re-read, the hub's message warned, no "Remove failed"), the hub discriminator at both removal sites and its RPC mapping (red at `app_write_applied_test.go:184`/`:217`), the OAuth-only billing render with apiKey cards unchanged, and the client reader's binding test.

Verification for this round, on `3329cbdab8533a30fa56950b5bf1559e40dcce13`: biome, `npm run typecheck`, the focused credentials suites (15 files, 435 tests), the appwire-client tests, `make lint-package-imports`, `make test-web`, `gofmt`, `make vet`, `make lint-evenerfuzz`, `make lint-fuzz-registry`, `make lint-golangci`, and `go test ./cmd/evener-hub/... ./appwire/...`.

### Round: the standing-removal discriminator names only a removal that stands, and every client reconciles it

- **The wire class and the message agree again.** The hub had wrapped every un-restored rollback in the standing-removal discriminator, including the config-backed supplies whose own frames say the instance is still configured - so a client reading it would close the editor and report the name removed while the row was still in `providers.toml` and still resolved. `restoreFailedRemoval` now carries the discriminator only when the failed restore took the layer that carries the instance (a stored key or an OAuth record); `supplyAny`/`supplyConfig` stay a plain applied write, and the caller that knows the config IS gone (the rollback-write failure) marks the standing removal itself. The hub tests move with the rule: the config-write-failed and rollback-succeeded cases assert a plain applied write, the carrying-record case keeps the discriminator.
- **The web path clears the selection it reconciled away**, the way the success path does and for the same reason, instead of leaving the sheet open on a removed instance.
- **The TUI and the native client reconcile both discriminators** rather than reporting ordinary failures and waiting on the passive update notification. The TUI refreshes the instance list, follows a persisted rename to its new name, and warns; the native client closes the editor, re-reads the provider list, and shows its own warning wording - the rejection's text can echo submitted credentials, so it never reaches the screen. A plain failure keeps today's behavior in both.
- **Tested.** Supply gating at both removal sites and the RPC mapping (`app_write_applied_test.go:221`/`:263`/`:326`), the web sheet clearing (`CredentialsSection.test.tsx:608`), the TUI's two reconciliation paths (`hub_update_config_applied_test.go:101`/`:175`), and the native classifier plus mount tests (`ProvidersScreen.test.tsx:134`) - each observed red before its fix.

Verification for this round, on `0be7df95744793e77d3015f6fab3aae572d8831e`: `gofmt`, `go test ./cmd/evener-hub/... ./cmd/evener-tui/...`, `go test ./appwire/...`, `make vet`, `make lint-evenerfuzz`, `make lint-golangci`, biome, `npm run typecheck`, the focused credentials suites, `make lint-package-imports`, `make test-web`, and `make test-native`.

### Round: the edit RPC verifies the endpoint its caller was shown, and every client asserts it

- **The last client-side-only guard is now atomic in the hub.** Rename and edit identity validation lived only in the clients, so a listing check could be stale by the time the hub applied a request - a form opened on one instance could edit or rename the replacement another client had put under the same name. The Edit RPC now takes the same `expectedEndpointFingerprint` the removal, clear and credential RPCs already verify (an empty assertion asserts nothing; a mismatch or an unresolvable name is a Conflict), checked under the write lock before anything lands. The sheet sends the endpoint it seeded the draft with, the TUI the row its form was opened on, the native editor its instance's fingerprint, and each reconciles the refusal as the endpoint-conflict class it already handles - re-reading the list so a retry asserts the destination now on screen. The TUI's removal refusal moves onto that shared class, so it re-reads the list too.
- **And the rename discriminator joins frame fuzzing** - it was missing from `frameErrorInfos`, so generated error frames never carried it.
- **Tested.** The hub refuses a stale assertion for a rename and for a field edit and changes nothing; the web sheet sends the seeded fingerprint and reconciles the refusal (five existing rename tests strengthened to assert the fingerprint travels); the TUI asserts the opened row's fingerprint and the conflict class's list re-read; the native params carry it and the refusal clears, refreshes and warns in own wording.

Verification for this round, on `4fb8246dfa0f207c4ec575ce4c966d7546884dbd`: `gofmt`, `go test ./cmd/evener-hub/... ./cmd/evener-tui/... ./appwire/...`, `make vet`, `make lint-evenerfuzz`, `make lint-golangci`, `make lint-generated`, biome, `npm run typecheck`, the focused credentials suites, `make lint-package-imports`, `make test-web`, and `make test-native`.

### Round: the endpoint-assertion refusal gets its own discriminant, and a default-named implicit removal classifies by its carrying layer

- **A refused endpoint assertion is no longer indistinguishable from a genuine conflict.** The hub answered `appwire.Conflict` (`evenerErrorInfo: "conflict"`) both for an assertion it could not honour and for name collisions, so the clients taught to read "conflict" as "the endpoint moved" reported the wrong story: a rename onto an occupied name told the user to review an endpoint, and a mobile create with a taken name closed the editor the same way. The refusal now carries `ErrorEndpointConflict` (`"endpointConflict"`) at every site that refuses an asserted destination — the credential assertion's three branches, the sign-in flow's pre-flight and completion checks, the credential-probe refusals — while name collisions, an expired flow and the pairing RPC keep the generic conflict. The client predicate matches the discriminant, never the code, and each client keeps the hub's own message and the user's form for a genuine conflict.
- **And the removal's classification follows the layer that actually carries the row.** `Remove`'s config-write-failure rollback always reported "the instance is still configured", but that branch is also taken when only the `default` pointer named the instance: for an implicit instance carried by a stored key whose restore also failed, the removal stands and the row no longer resolves, yet the error carried no standing-removal discriminator, so the client showed a failed remove whose retry hit a gone instance. The rollback now classifies by the carrying layer when the authored entry was not what changed, says the removal stands, and folds a carried restore's leftovers into the applied report the way the reload rollback does.
- **Tested.** The hub asserts the discriminant at the assertion sites and the standing classification for the default-named implicit case (the authored sibling keeps the configured frame); the web surfaces a rename collision as the hub's refusal; the TUI rejects a genuine conflict from its endpoint-conflict class; the native client keeps the create form for a name collision.

Verification for this round, on `b50f8eedf1fcb7e6d287a568bd1492fd4e27f6cf`: `gofmt`, `go test ./cmd/evener-hub/... ./cmd/evener-tui/... ./appwire/...`, `make vet`, `make lint-evenerfuzz`, `make lint-fuzz-registry`, `make lint-golangci`, `make lint-generated`, biome, `npm run typecheck`, the focused credentials suites, `make lint-package-imports`, `make test-web`, and `make test-native`.

### Round: an asserted edit is the only one the hub refuses, and the reload rollback classifies like its sibling

- **Edits and renames work again on a hub that cannot key fingerprints.** The previous round's assertion was checked even when the caller asserted nothing, and the verifier fails closed when the hub cannot key at all - so a hub whose fingerprint key file is unusable refused every edit and rename, because the listing (correctly) carried no fingerprints and the client therefore asserted nothing. An edit lands no secret, so there is nothing to fail closed for: the assertion is verified only when the caller made one, and an asserted endpoint the hub cannot resolve is still refused. The credential writes keep their fail-closed behaviour.
- **And the reload-failure rollback classifies by the carrying layer**, like the config-write-failure sibling the previous round fixed. For a default-named implicit instance whose carrying key or OAuth record cannot be restored, the removal stands and the row no longer resolves; the error now carries the standing frame and the discriminator, so clients reconcile it instead of offering a retry that cannot succeed. The authored sibling keeps the configured frame.
- **Tested.** An unasserted edit and a rename both succeed under an unkeyable state root while a stale non-empty assertion still refuses; the implicit default-named instance's failed restore carries the standing frame and discriminator.

Verification for this round, on `0e29489aea661add68058cdfac00391ab57691d5`: `gofmt`, `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `make lint-evenerfuzz`, `make lint-generated`.

### Round: the sign-in flow refusals pin the discriminant, not just its code

- The review's one remaining low: `assertFlowEndpointRefusal` asserted only `CodeConflict`, which `EndpointConflict` shares with genuine conflicts (a name collision, an expired flow), so a guard regressing to a plain `Conflict` would keep the suite green while every client stopped recognizing a moved destination during sign-in. The helper now reuses `assertEndpointConflict` (code, `ErrorData` shape and discriminant) and keeps its message assertions - mutation-checked by reverting the guards to `appwire.Conflict`, which fails the flow tests at `app_auth_flow_endpoint_test.go:358` and `:541`.

Verification for this round, on `9026c80e11453fde97f43304bee1d58d9d541959`: `gofmt`, `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-generated`.

### Round: the cleanup-failure rollback classifies by the carrying layer, and the native removal reconciles a refused assertion

- **The last member of the rollback-classification family.** The `removeCredentials`-failure branch still asked `restoreFailedRemoval` with `supplyAny` and the fixed "the instance is still configured" frame. For an implicit instance the user credentialed through the UI, the cleanup deletes the stored key and can fail on the auth-record delete; when the put-back then fails too, the only layer that carried the instance is gone, yet the error carried no standing-removal discriminator — so every client kept the confirmation and offered a retry whose target no longer resolves. The branch now reads `authored` before the cleanup and classifies by `supplyOf` like its siblings; an audit of all five `restoreFailedRemoval` call sites confirmed it was the last gap.
- **The native removal reconciles its own refusal.** It asserts the row's endpoint, but `act`'s catch only reconciled applied-write errors, so an endpoint-conflict refusal left the screen unrefreshed and a retry re-sent the same stale fingerprint. The catch now classifies that refusal, clears the selection, re-reads the provider list, and warns in the client's own words.
- **Tested.** The cleanup-failure branch's implicit stored-key case asserts the standing frame and discriminator; the native removal's conflict refusal asserts the clear, the refresh and the warning.

Verification for this round, on `b72598564dca29bce5b84e6d1444b480ff1faa83`: `gofmt`, `go test ./cmd/evener-hub/... -count=1`, `make vet`, `make lint-golangci`, `make lint-evenerfuzz`, `make lint-generated`, `make test-native`.

### Round: every client path that asserts an endpoint can recover from a moved one

- **The sheet's retry actually lands.** The endpoint-conflict branch of a save asked the user to try again while its own identity guard would either reseed the draft (dropping their edits) or resend the stale assertion. It now awaits the listing it asks for and, when the refreshed row is the same instance re-pointed (an identity comparison that ignores the endpoint fingerprint), re-anchors the seeded identity and the asserted fingerprint to that row and leaves the draft alone. A replacement or an absent row keeps the guard as the authority, because those edits must not follow a stranger wearing the name.
- **And the confirm-gated actions reconcile the refusal too.** Remove, Clear and Clear stored key assert the listed endpoint; the section's catch now closes the confirmation, clears the selection, re-reads the listing, and warns in the actions' own words, where before it showed a bare "Remove failed" and left the dialog holding the refused fingerprint — the one client path that could not recover on its own.
- **A dead string is gone**: `ENDPOINT_CHANGED_SAVE_MESSAGE` was left unreferenced by the previous round's rerouting.
- **Tested.** A retry after a conflict sends the edit with the refreshed fingerprint and the edits intact (a replacement case still refuses); a refused removal and a refused clear each close the confirmation, re-read, and warn.

Verification for this round, on `0051ce980f59349bbe85442b2658573d6dbfcd9b`: the focused credentials suites (441), `npm run typecheck`, biome, `make lint-package-imports`; the full CI matrix runs on the pushed head.

### Disclosed follow-up: the TUI credential actions carry no endpoint fingerprint

roborev's review of head `1acb58eb75d4c53559a857e9da8281a55cb08513` - with every required check green and the other two reviewers finding no issues - reported one **high**:

> **TUI credential actions lack endpoint fingerprint assertions** (`cmd/evener-tui/internal/launchconfig/credentials_panel.go:242-311`; `cmd/evener-tui/internal/launchconfig/launchconfig_client.go:108-158`): the TUI's credential actions capture only the instance name and send empty `ExpectedEndpointFingerprint` values, so a stale TUI can test, set, replace or clear credentials for a different endpoint after another client repoints that name - and a credential test may send the stored secret to the replacement endpoint.

This is pre-existing: the credential RPCs' assertion contract long predates this series, and the TUI simply never populated it. This PR closes the assertion family for the paths it touched (edit and remove, in all three clients). Landing it here is a deliberate, recorded decision - the TUI credential actions are scoped to the follow-up PR, together with the durable per-removal transaction marker already carried by #1607's disclosure.


obra:
**On "Ownership allow-list omits `api_key` and `credential_headers` credential sources":** I checked this against the branch, and the premise does not hold, so the proposed classification would be a case that cannot be reached.

An implicit instance exists only by way of a curated record: the instance set is built from `r.curated` for providers whose record sets `implicit` (`llm/registry/instances.go:499-512`). The two sources in question come from the record's head fields, ahead of the store and the environment (`llm/registry/instances.go`'s credential resolution: `h.APIKey` → `api_key`, `h.CredentialHeaders["Authorization"]` → `credential_headers`). No curated record carries either field:

- Checked against the embedded catalog: of 219 curated records, 0 have a non-empty `api_key` or `credential_headers` (a scratch test iterating `r.curated` from `embeddedLoad`, run and then removed).
- The upstream conversion cannot introduce one: `FromModelsDev` sets only `p.APIKeyEnv` (`llm/registry/modelsdev.go:251`), and `api_key_env` resolves to the `env:<NAME>` source, which the allow-list already places with the environment.
- The curated data files (`llm/registry/data/providers_overlay.toml`, the models.dev snapshot) carry no `api_key` or `credential_headers` key at all.

Authored entries, which do resolve `api_key` (an inline key, or a `${VAR}` expression), are not implicit, so `environmentBacked` returns false for them by design: deleting the entry is the removal, and the config write takes the entry away.

An implicit row can therefore only resolve `env:*`, `adc`, `oauth`, `none`, or `store`, and the allow-list already classifies each of those. Classifying `api_key`/`credential_headers` as the environment's would be dead code on both sides, which is the same reason this branch removed its one dead special case, the Codex exemption.

The other three findings from this review are being fixed in a follow-up commit on this branch: the classification race (the ownership question is re-asked under the credential lock) and both lows.

obra:
**On "`clearAPIKeyEnv` is normalized to `aPIKeyEnv`":** this one's premise does not hold, but I landed its fix anyway as hardening.

The param is `clearApiKeyEnv` with a lowercase "pi", in both places that define it: `appwire/types.go` carries `ClearAPIKeyEnv bool` with `json:"clearApiKeyEnv,omitempty"`, and the generated `types.gen.ts` carries `clearApiKeyEnv?: boolean`. So the one-letter de-capitalization the finding flags (`key.charAt(5).toLowerCase() + key.slice(6)`) was already yielding `apiKeyEnv`, and every focused test over the clear flags stayed green when I reverted to it — there is no `clearAPIKeyEnv` (capital A) on the wire to mis-normalize.

The fix is still in: `changedFields` now reads an explicit `CLEAR_FIELD_NAMES` map covering all five clear flags (`baseUrl`, `protocol`, `surface`, `apiKeyEnv`, `credentialHeader`), so a future capitalised field name cannot silently repeat this class, and a new test file pins the mapping flag by flag.

The medium is fixed too: the sheet now steers to the new name whenever the hub reports the rename as persisted (`ErrorInstanceRenamePersisted` is the hub saying the rename is on disk), keeping the listing reconciliation and its retry and the warning that names the credential left behind. The regression test drives a listing that omits the new row and asserts the steer still happens.

obra:
### Disclosed follow-up: TUI credential actions carry no endpoint fingerprint

roborev's review of head `1acb58eb75` — with every required check green and the other two reviewers finding no issues — reported one **high**:

> TUI credential actions capture only the instance name and send empty `ExpectedEndpointFingerprint` values (`cmd/evener-tui/internal/launchconfig/credentials_panel.go:242-311`, `launchconfig_client.go:108-158`). A stale TUI can therefore test, set, replace, or clear credentials for a different endpoint after another client repoints that name; credential testing may send the stored secret to the replacement endpoint.

This is pre-existing — the credential RPCs' assertion contract long predates this series and the TUI never populated it — and this PR closes the assertion family for the paths it touched (edit and remove, in all three clients). It is landing with this disclosed, by explicit decision, as the follow-up PR alongside the durable per-removal transaction marker already disclosed on #1607.

PR #1728, opened 2026-09-18
sdk: extract the ready-generation fence from keybindingsStore

obra:
## Summary
- First piece of the #1549 (D6-core) decomposition, per Jesse's five-rounds rule (docs/superpowers/plans/2026-09-12-sdk-migration.md D6, plan comment on #1549). Plan line 1: "Extract the ready-generation fence: readyGenerationFence.ts + keybindingsStore wired to it (~205)."
- Moves keybindingsStore's generation/serial/token bookkeeping (`isCurrent`, `liveHub`, `readStillMine`, `writeStillMine`, `activeEpoch`, `writeSerial`, `disposed`) into a standalone `readyGenerationFence.ts` module (`createReadyGenerationFence`), taken byte-for-byte from #1549's head (dfde921602c17e70d393adfcbdeece2dcb4682d9).
- Rewires every one of those call sites in `keybindingsStore.ts` to `fence.*`, taking the exact shape from #1549's `keybindingsStore.ts` for those sites only — nothing else from that file (no draft-port changes, no settle changes, no draftUnreadable: those are separate pieces of the decomposition).
- `keybindingsStore.test.ts` is unchanged (main's version) and passes unmodified — the proof that this is a pure move with no behavior change.
- Plumbing: `readyGenerationFence.ts` added to `index.ts`, `tsconfig.build.json`; README updated with the fence's own paragraph (scoped to keybindingsStore only, since no other store uses it yet).

## Test plan
- [x] `npx vitest run readyGenerationFence.test.ts keybindingsStore.test.ts` — 63 passed (2 files)
- [x] `npm run typecheck` — clean
- [x] `npx biome ci ../../../appwire-client/typescript` and `npm run lint` — clean
- [x] `make lint-package-imports` — pass
- [x] index.ts falsification: removed the fence's barrel export, `make test-api-package` failed with `shipped module unreachable from every published specifier: readyGenerationFence`, restored via `git diff | git apply -R` from the committed base, re-ran green

PR #1729, opened 2026-09-18
fix(web): a stale client's ready callback cannot begin a generation

obra:
## Summary
- Second piece of the #1549 (D6-core) decomposition, per Jesse's five-rounds rule (docs/superpowers/plans/2026-09-12-sdk-migration.md D6, plan comment on #1549). Plan line 2: "Web: a stale client's ready callback cannot begin a generation (~27)."
- Targets `main` directly: #1728 (PR 1, the fence extraction) is merged (b89ae74f4), and this branch has been refreshed onto it.
- Adds `cmd/evener-hub/frontend/src/stores/readyGenerationCallback.ts` (new, from #1549's head dfde921602c17e70d393adfcbdeece2dcb4682d9) and wires it into `stores/keybindings.ts`'s `client.onReady(...)` call: a client's own ready dispatch can snapshot its registered handlers before a later rewire's unsubscribe removes this one, so a client already replaced can still fire the handler it was registered with against whichever client is wired NOW. The wrapper runs the handler only while its own client is still current.
- Adds the one keybindings.test.ts case from #1549 that pins this ("a stale client's ready callback cannot begin a generation once it has been replaced"); no other test changes (the other keybindings.test.ts diffs in #1549 belong to later pieces of the decomposition - the settle/write-rejection behavior change - out of scope here).
- Review round fixes (folded in, see PR comments): transcriptDisplay.ts's own bare `client.onReady` wrapped with the same helper (Medium 1), and `readyGenerationCallback` now mints a per-registration token so a client object rewired away and back cannot let a stale registration pass identity alone (Medium 2).

## Test plan
- [x] TDD: reverted the `keybindings.ts` fix, confirmed the new test fails RED (`expected 2 to be 1`), restored the fix, confirmed GREEN (66/66 pass)
- [x] `npx vitest run src/stores/keybindings.test.ts src/stores/transcriptDisplay.test.ts src/stores/readyGenerationCallback.test.ts` plus the package's `keybindingsStore.test.ts` - 155 passed
- [x] `npm run typecheck` - clean


obra:
## Review round disposition (ec2a322)

- Restacked on the fixed #1728 (unused fence surface dropped there resolves this PR's Low).
- Medium 1: transcriptDisplay.ts:425 wired a bare `client.onReady`, so a stale client's belated dispatch could call `beginReadyGeneration(staleClient)` unconditionally - rewiring the notification listener onto the stale client and fencing the CURRENT client's own generation out. Wrapped it with `readyGenerationCallback`, matching keybindings.ts. TDD: added "a stale client's ready callback cannot begin a generation once it has been replaced" to transcriptDisplay.test.ts, confirmed RED (the current client's post-fire notification never applied), then GREEN.
- Medium 2: readyGenerationCallback compared client identity only, so a client object rewired away and later back (reconnect to the same instance) let a stale registration from the earlier wiring pass once that client is current again. Each registration now mints its own token (a `Symbol`), keyed by client object in a `WeakMap`; only the most recent registration for that object may fire. TDD: new readyGenerationCallback.test.ts, confirmed RED for the "later registration supersedes" and "A -> B -> A" cases, then GREEN.

Gates: vitest (readyGenerationCallback.test.ts + keybindings.test.ts + transcriptDisplay.test.ts) 98/98 passed, typecheck clean, biome ci + npm lint clean, lint-package-imports pass.

obra:
## Review round disposition (7f52873)

High: readyGenerationCallback.ts's WeakMap<client, symbol> was module-global, so keybindings.ts and transcriptDisplay.ts registering on the SAME connectionStore client overwrote each other's token - the second store's registration permanently fenced the first store's own callback out, and its hub settings never loaded. My own "a later registration for the same client object supersedes an earlier one" test encoded this bug (it only ever exercised one store).

Fix: `createReadyGenerationCallback()` now returns one guard backed by its own private WeakMap; each store (keybindings.ts, transcriptDisplay.ts) creates its guard once at module scope and reuses it for every registration, so the two stores never contend over one shared slot. The A -> B -> A rewiring test stays scoped within one store's own guard; added a new "two stores' own guards do not contend over one shared client" case using two separate guard instances against the same FakeClient.

TDD: added the cross-store test against the OLD shared-map implementation first, confirmed RED (`expected 1, got 0` on the losing store's begin), then did the factory refactor and confirmed GREEN across all four cases plus the pre-existing three.

Gates: vitest (readyGenerationCallback.test.ts + keybindings.test.ts + transcriptDisplay.test.ts) 99/99 passed, typecheck clean, biome ci + npm lint clean, lint-package-imports pass. Stacked-merge check: every file #1728 touched still diffs empty against origin/main.

PR #1730, opened 2026-09-18
test(server): a thread/read snapshot keeps live input images

obra:
Piece F of the #1580 decomposition plan (RoboRev round-20 mediums/lows split into serial PRs — see the [decomposition plan comment](https://github.com/prime-radiant-inc/evener/pull/1580#issuecomment-5723104182)).

## Summary
- Adds `TestAppTurnSnapshotKeepsLiveInputImages`, cut byte-for-byte from #1580 (`53bb2b6ca`).
- Pure characterization test, zero production code change: proves the hub's `thread/read` snapshot path already keeps live input images across an empty-image settle (`mergeAppThreadItem`'s existing `len==0` rule), the same way the existing delta cases cover text.
- No dependency on any other piece of the plan; independent of A-E.
- Absorbs none of today's RoboRev findings (there is nothing to fix here).

## Test plan
- `go test -run TestAppTurnSnapshotKeepsLiveInputImages -count=1 -v ./server/...` — PASS
- `$(go env GOROOT)/bin/gofmt -l server/appwire_turns_images_test.go` — clean


obra:
## Review round disposition (e54067bbb)

Real. Measured before fixing: confirmed `handleAppThreadRead`'s `IncludeTurns`
path calls `s.appLatestItemTurns` → `snapshot.LatestItemCandidates` (windowed,
then `appitempaging.RegroupTurnFragments`/`NormalizeProjectedItemCompleteness`)
— never `appTurnSnapshot.Snapshot()`, which is what the original test called
directly. An image-loss regression specific to the `LatestItemCandidates`
windowing/regrouping path would have passed this test unchanged.

**Fix:** renamed to `TestAppThreadReadKeepsLiveInputImages`. Builds the same
snapshot via `.Apply()` as before, then installs it on a real `*Server`
(`installTurnSnapshotObjectForTest`, mirroring `appwire_turns_paging_test.go`'s
own `installTurnSnapshotForTest` for a static turn list) and reads it back
through `srv.handleAppThreadRead(ctx, appwire.ThreadReadParams{Ref: "local:01T",
IncludeTurns: true})` — the actual production read path. Asserts `Name`,
`MediaType`, and decoded `Data` (Go's `encoding/json` decodes the wire's
base64 `data` field straight into `[]byte`) in the response, both right after
the live `item/started` and after the `item/completed` settle that says
nothing about images. Kept the existing retention assertion (it's the whole
point of the test) rather than folding it — just moved it behind the real
read call.

**Gates:**
- `go test -run TestAppThreadReadKeepsLiveInputImages -count=1 -v ./server/...` — PASS
- `$(go env GOROOT)/bin/gofmt -l server/appwire_turns_images_test.go` — clean

Head is `e54067bbbf7e10e37443dd3ab29eca74588a8ffc`, still mergeable, still
independent of every other piece.

PR #1731, opened 2026-09-18
fix(sdk): interrupt/steer resolve a pending ask; export the reasoning join

obra:
Piece A of the #1580 decomposition plan (RoboRev round-20 mediums/lows split into serial PRs — see the [decomposition plan comment](https://github.com/prime-radiant-inc/evener/pull/1580#issuecomment-5723104182)). Independent of every other piece; starts from `main`.

## Summary
- Cut byte-for-byte from #1580 (`53bb2b6ca`): `deriveAskQuestions.ts`'s ask-resolution boundary fix (an interrupt or a user steer resolves a pending ask, like a reply — round 6's own fix on #1580, already independently valuable to the web composer's ask dock), `isActiveItem`/`joinedReasoningParagraphs` becoming shared package exports, and the small `mapTurn`/`settleFirstMatchingTurn`/`warningMessage` hardening that rode along in `reducer.ts`.
- Fixes this round's Low: `imagesToItemImagesForSession`'s comment falsely claimed `outputImagesToItemImages` folds an empty output-image list to absent the same way input images do. Measured against `appwire/output_images.go`: output images mean the opposite — a non-nil empty list is the hub's only explicit removal signal, nil means "never had any." `outputImagesToItemImages` already reads it correctly; only the comment was corrected.

## Test plan
- `npx vitest run --config mobile-native/vitest.config.mts appwire-client/typescript/reducer.test.ts appwire-client/typescript/deriveAskQuestions.test.ts appwire-client/typescript/transcriptProjector.test.ts` — 256/256 pass
- `npm run typecheck` (`cmd/evener-hub/frontend`) — clean
- `npx biome ci ../../../appwire-client/typescript` (from `cmd/evener-hub/frontend`) — clean, 214 files


obra:
## Simplify disposition (ab1d8ef40)

Applied all four items, no behavior change:

1. **Reuse:** `reasoningFormat.ts` now re-exports `joinedReasoningParagraphs` from `@evener/appwire-client` instead of carrying a byte-identical local copy — matches #1580's own oracle version of this file exactly, so piece B's stacked diff for this file is now empty. Its local `pendingTextJoined` import is gone with the local implementation.
2. **Simplification:** `transcriptProjector.ts`'s exported `isActiveItem` now takes `turnStatus: string | undefined` instead of `Pick<TurnModel, "status">` — matches the phone's own local copy's parameter shape (`mobile/src/conversation/project.ts`), so piece C can adopt this export directly with no conversion wrapper at any call site. Updated the 2 other call sites in this file (`isLiveCurrentReasoning`, the tool-execution decision branch) to pass `turn.status`.
3. **Reuse:** dropped the one-line `// Output images: see appwire.MergeOutputImages; input images keep the length rule.` comment directly above `imagesToItemImagesForSession` — redundant with the full explanation the block right below it (added in #1580's round-20 fix) now states. The two other copies of this same one-liner, on `outputImagesToItemImages` and `wireItemToModel`'s output-image line, are unchanged — they still name the rule without repeating it.
4. **Reuse:** `settleFirstMatchingTurn`'s copy of `mapTurn`'s "hand the same array back" rationale is trimmed to a cross-reference ("for the reason mapTurn's own comment gives").

**Gates:**
- `npx vitest run --config mobile-native/vitest.config.mts appwire-client/typescript/reducer.test.ts appwire-client/typescript/deriveAskQuestions.test.ts appwire-client/typescript/transcriptProjector.test.ts` — 256/256 pass
- `npm run typecheck` (`cmd/evener-hub/frontend`) — clean
- `npx biome ci ../../../appwire-client/typescript` / `npm run lint` (`cmd/evener-hub/frontend`) — clean, 1354 files

Head is `ab1d8ef4019dfb8c36d97a200a105aa0984a0f88`. RoboRev's review of this head is separate and comes on its own.

obra:
## Review round disposition (2b99200b7)

Both real. Measured before fixing, from one canonical list rather than
patching each site independently.

**Measurement — what clears/resolves the pending ask, and how each host sees it:**

| Server event | Server code | Transcript entry live | Transcript entry on reload | Client marker (deriveAskQuestions.ts) | Restore marker (session_tools_ask.go) |
|---|---|---|---|---|---|
| Plain user message / composed answer | `processOneInput` clears `s.askPending = nil` unconditionally on entry (session_lifecycle.go:1804, before the turn's model call runs) | `userMessage` item | `TurnUserInput` | `isResolutionItem`: `type === "userMessage"` | `case schema.TurnUserInput: return nil, false` |
| Interrupt | `session_lifecycle.go:1271` `s.clearAskPending()` directly, then `appendSteeringTurn(interruptMsg, SteeringKindInterrupted)` (:1287) | `steering` item, `steeringKind: "interrupted"` | `TurnSteering`, `SteeringKind: "interrupted"` | `isResolutionItem`: `steeringKind === "interrupted"` | **new** `case schema.TurnSteering:` — `SteeringKind == events.SteeringKindInterrupted` |
| Accepted user steer (carrier succeeds) | Same `processOneInput` unconditional clear, entering as `EntryUserInput` | `steering` item, `source: "user"` (may also carry `steeringKind: "human-note"`, session_notes_rpc.go:61) | `TurnSteering`, `SteeringSource: "user"` | `isResolutionItem`: `source === "user"` | **new** `case schema.TurnSteering:` — `SteeringSource == events.SteeringSourceUser` |
| Accepted user steer, carrier's OWN turn fails outright | Same unconditional clear (already happened before the model call that then failed) | **no item at all** — a turn-level error with zero items | `TurnFailure` (TurnFailure's own doc comment: "the persisted counterpart of a live turn with an error and no items... without it a failure existed only as a live event") | **new** `isResolutionTurn`: `turn.items.length === 0 && turn.error !== undefined` | **new** `case schema.TurnFailure: return nil, false` (unconditional — reaching this turn at all means an accepted-turn attempt ran and cleared askPending) |
| Daemon-originated steer (task-nudge, notification reminder) | Never reaches `processOneInput`'s user-input path; askPending untouched | `steering` item, no `steeringKind`/`source` marker | `TurnSteering`, no marker | not a resolution | `case schema.TurnSteering:` falls through — not decisive (unchanged from before this round) |

**High** (deriveAskQuestions.ts:70 / server restore): fixed by adding the
`TurnSteering` (kind-aware) and `TurnFailure` (unconditional) cases to
`deriveRestoredAskPending`'s backward scan, and the mirroring `isResolutionTurn`
turn-level check to `liveAskQuestions` (walking `model.turns` instead of a
pre-flattened item array, since an item-less resolution turn contributes
nothing to flatten).

**Medium** (deriveAskQuestions.ts:70, folded from #1732's review of the same
diff through the stack): the failed-carrier row above — same root cause,
same fix.

**Low** (deriveAskQuestions.ts:67): corrected the comment claiming a
user-sourced steering item "never" carries a `steeringKind` — human-note
updates do (`session_notes_rpc.go:61`, `layoutRoles.ts:46`'s own
`steeringKind !== "human-note"` check). The `||` logic was already right;
only the comment was wrong.

**TDD, RED first, wire-shaped:**
- `agent/session_ask_test.go` — two new tests built via `recordTurn` (the same
  primitive `appendSteeringTurn` itself uses), not the live steer/interrupt
  path, so they don't couple to which mechanism produces the marker:
  `TestAskUser_RestoreResolvesAcrossInterruptOrUserSteering` (both `steeringKind:
  interrupted` and `SteeringSource: user` sub-cases) and
  `TestAskUser_RestoreResolvesAcrossFailedSteeringCarrier`. Confirmed failing
  (`restored askPending` non-empty) before the fix.
- `appwire-client/typescript/deriveAskQuestions.test.ts` — two new tests: a
  turn with no items and an error resolves; a turn WITH items keeps resolving
  by its items even if it also errors (pins that `isResolutionTurn` doesn't
  over-fire). Confirmed failing before the fix.

A second finding turned up while gating this round: `TestNoBareWallClockDeadlineInAgentTests`
failed on both new Go tests (a bare `context.WithTimeout` needs the file's
standing "// TRIPWIRE: scripted in-process adapter..." comment). Fixed in the
same commit.

**Cascading fix (not a new finding, a consequence of this round's own
`warningMessage`/`rawWarningFrame` restoration from #1731's first round):**
`mobile/src/conversation/project.test.ts`'s three "drops a warning carrying
only whitespace" cases now correctly render the raw-frame fallback instead of
dropping the row — a whitespace-only `message` trims to `""`, the same
message-less reading `warningMessage` already gives a wholly-absent one, so
the reducer folds `item.text` to the bounded raw frame before native's
"whitespace is not content" join ever sees it. Updated on piece C's branch
(where `project.test.ts` lives) to assert the new, correct behavior.

**Gates:**
- `npx vitest run --config mobile-native/vitest.config.mts appwire-client/typescript/reducer.test.ts appwire-client/typescript/deriveAskQuestions.test.ts appwire-client/typescript/transcriptProjector.test.ts` — 260/260 pass
- `npm run typecheck` / `npm run lint` / `npx biome ci ../../../appwire-client/typescript` (`cmd/evener-hub/frontend`) — clean
- `agent`: `$(go env GOROOT)/bin/gofmt -l` clean; `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...` clean; `go test -run 'TestAskUser|TestNoBareWallClockDeadlineInAgentTests' -count=1 .` — PASS. Full `go test -count=1 ./...` for the `agent` package itself shows pre-existing, unrelated failures (`TestRetirementAutonomousReLock*`, `TestRetirementSafetyEscalation`, `TestScratchRetentionTerminalReleaseAllowsCollection`, `TestRetirementSharedChildScratchBindingsRestore` — worktree-porcelain/scratch-retention, nothing this round touches); every `agent/...` subpackage passes.

Head is `2b99200b711871e2162eda0f5ec5af2cf7124027`. B and C are being restacked onto it now.

obra:
## Review round 4 disposition (93edcf7f2)

**Measurement, before coding:** does the wire carry `askPending` on hydrate and on the notification that changes it? Yes, both — `appwire-client/typescript/model.ts`'s `ThreadModel.askPending: boolean` (required), `reducer.ts:822` sets it from `thread.evener.askPending` on hydrate, and `reducer.ts:1517-1528` refreshes it from `thread/status/changed`'s own `askPending` field on every status transition (`n.params.askPending ?? model.askPending`). Confirmed already covered by `reducer.test.ts`'s "askPending is wire-authoritative from the thread snapshot" and "item lifecycle never clobbers the wire's thread-level askPending".

**Fix, per the ruling:** `liveAskQuestions` (deriveAskQuestions.ts) now returns `[]` immediately when `model.askPending` is false, and only scans items for "which questions" once the wire says something is open. `isResolutionTurn` (the item-less-turn-failure check) is deleted outright — the client no longer tries to re-derive "is this resolved" from transcript shape at all, closing the round-4 finding by elimination rather than patching it. `isResolutionItem` stays: it still answers a different question ("which items belong to the still-open round"), needed because the wire's `askPending` is a bool, not a list.

Server-side, `turnResolvesAskBoundary` (session_tools_ask.go) is a new shared function `deriveRestoredState` and `deriveRestoredAskPending` both call for their decisive-turn table, so the two can no longer drift apart the way this round's own bug showed (deriveRestoredState was missing the TurnSteering/TurnFailure cases deriveRestoredAskPending had). `TurnFailure` is narrowed to one shape: `schema.TurnFailureInfo.SteeringCarrier`, set only at `acceptSteeringCarrierInput`'s `carrierSteerUndelivered` site (a steering carrier whose own steer failed to append, leaving nothing else behind) — every other TurnFailure (a retry-budget exhaustion, a failed steering-selection prepare, a provider error) is no longer decisive on its own, so a round that posted a real ask_user call before failing for an unrelated reason keeps that ask pending. `injectPostToolSteering`'s mid-round drain now calls `clearAskPendingForResolvingSteer` itself when it delivers a user-sourced steer, instead of waiting for the next accepted-turn entry to clear it — and `askedThisRound`'s computation moved to right after `persistToolResults`, before that same drain runs, so the round that posted the ask and had its own resolving steer land in the same drain window still reports `askedThisRound` correctly.

**Lows folded into this round:**
- deriveAskQuestions.ts's `turn.error !== undefined` Low is moot: `isResolutionTurn` (the function it applied to) is deleted.
- reducer.ts's `rawWarningFrame` truncated by UTF-16 unit (`String#slice`), which could split a surrogate pair at the 2000-char boundary. Now truncates by code point (`Array.from(...).slice(...).join("")`) instead, bounded in code points rather than exactly the same UTF-16 count.

**TDD, live paths, not synthetic markers:** the interrupt and accepted-user-steer restore tests now drive a real interrupted round (a scripted STREAMING adapter cancelled mid-stream — `fakeAdapter.Complete` can't produce a genuine cancellation) and the real `AcceptClientMutationSteer` → `claimSteeringCarrierTurn` → `acceptSteeringCarrierInput` path, respectively, rather than hand-built `TurnSteering`/`TurnFailure` turns. A new test drives `recordFailedSteeringSelection` (the same production path `TestSkillActivation_SteeringSelectionReconciledAfterAdmissionSaveFailure` already drives directly) to prove an unrelated mid-round failure does NOT resolve a pending ask. A new live test proves `injectPostToolSteering`'s drain clears `askPendingCount()` for a mid-flight accepted steer, not just at restore.

**Consequences found downstream while restacking C/D on this fix (fixed there, not here):** piece C's `liveAsksFor` memoized by `model.turns` alone, so a status-only frame that flips `askPending` with the same turns reference served the stale answer; piece D's `changesRows` explicitly excluded `askPending` from the fields that can change a row, true before this round and false after. Both closed on their own branches' dispositions.

**Gates:** `npx vitest run --config mobile-native/vitest.config.mts appwire-client/typescript/reducer.test.ts appwire-client/typescript/deriveAskQuestions.test.ts appwire-client/typescript/askDock.test.ts appwire-client/typescript/transcriptProjector.test.ts` — 273/273 pass. `npm run typecheck` / `npx biome ci ../../../appwire-client/typescript` (`cmd/evener-hub/frontend`) — clean. `mobile-native`: `npm run check` — clean. `make lint-package-imports` — PASS. `agent`: `$(go env GOROOT)/bin/gofmt -l` clean; `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...` clean; `golangci-lint run ./...` — 0 issues; `go test ./agent/...` — full package tree green except the same four pre-existing, unrelated failures the round-3 disposition already named (`TestRetirementAutonomousReLock*`, `TestRetirementSafetyEscalation`, `TestScratchRetentionTerminalReleaseAllowsCollection`, `TestRetirementSharedChildScratchBindingsRestore`), confirmed identical on the clean pre-round-4 baseline (2b99200b7) before this round started. `go build ./...` (whole repo) — clean.

Head is `93edcf7f2`. B, C, E and D restacked onto it in order (B and C directly onto A; E and D onto the new C — the actual stack shape: B and C are siblings on A, E and D are siblings on C, not on each other).

obra:
## Round 5 disposition: decomposed, force-pushed to TypeScript-only (04584a21c)

Round 4 grew the server-side single-source change to ~430 non-test lines on top of the client gate and broke every consumer of the new askPending gate. Per the five-rounds-means-decompose rule, this PR is force-pushed down to its TypeScript-only diff; the Go mechanism moved to its own PR.

### Server findings (agent/session_lifecycle.go, session_tools_ask.go, session_queue.go, session_events.go, schema/turn.go) → #1806 "piece A-server: the daemon owns the ask boundary"
- High: user steer consumed mid-round never reaching the model — root cause was `askedThisRound`'s capture moving to before `injectPostToolSteering`'s drain; #1806 never carries that move (the capture stays in its original, correct position), so the mid-round drain's `clearAskPendingForResolvingSteer` resolves the ask and the round genuinely continues to a second model request.
- Medium: untagged `TurnFailure` from a claimed carrier's own failed skill-selection resurrecting a cleared `askPending` on restore — fixed with a short-lived carrier-claim marker so `recordFailedSteeringSelection` tags `SteeringCarrier` only for the carrier's own claimed steer.
- Medium: `askedThisRound`/live-restore state divergence — same root cause as the High above, fixed by the same non-move.
- Medium: restore tests asserting only `askPending`, not the restored `SessionState` — the four restore tests now assert both.
- Low ×3: `turnResolvesAskBoundary` simplified to a single bool; the dangling `TestCommunicate_ClientSteeringDuringToolExecutionRunsBeforeTheNextRequest` citation removed; round-history language scrubbed from the carried comments.

`agent/` carries zero diff on this PR after the force-push (`git diff origin/main -- agent/` is empty).

### This PR's own fix: the askPending gate itself
Measured in the hub (server/appwire_runtime.go's `stampAskPendingOnStatusChange`, `server/thread_envelope.go`'s facet table): a hydrate snapshot always carries `evener.askPending`, and a live turn boundary's `thread/status/changed` is meant to carry it too. Every failing consumer (13 web ask-dock/composer/session tests, 2 native `project.test.ts` cases, the dev composer surface fixture, and the package qualifier's three `liveAskQuestions` smoke calls) was missing the wire frame that carries the flag, not testing a wrong gate — fixed by adding `askPending: true` to each fixture's hydrate snapshot or accompanying `thread/status/changed` notification, matching what the hub actually emits.

Note: a separate, deeper measurement (on #1740's review) found the LIVE `thread/status/changed` notification's `AskPending` field is never actually set by the projector on origin/main (`internal/appprojector/appwire_projection.go`'s `threadStatus()` never touches it) — the gate is correct, but the wire signal it reads doesn't yet exist on the live path in production. That is being fixed as its own tiny hub PR (`projector: thread/status/changed carries askPending`, Refs #1613), separate from both this PR and #1806, since it's an additive server change unrelated to ask-boundary resolution.

Gate results (from the repo root / cmd/evener-hub/frontend as noted):
- `npx vitest run` on the 9 touched test files (AskDock.test.tsx, askDockStore.test.ts, Composer.test.tsx, Session.test.tsx, dev/surface-sections, reducer.test.ts, deriveAskQuestions.test.ts, askDock.test.ts, mobile/src/conversation/project.test.ts via mobile-native's vitest config): 621 + 771 = 1392 tests, all passing
- `npm run lint` (biome, src + appwire-client/typescript): clean
- `make test-api-package` (qualifier, touched): PASS

obra:
## Correction to the note above about the live askPending path

The earlier comment on this PR speculated that the live \`thread/status/changed\` notification's \`AskPending\` field might not actually be set by the projector on origin/main, based on a RoboRev finding on #1740. Measured end to end (not re-derived by reading): #1833 drives a real ask_user round through the real BridgeEvent/envelope pipeline and proves the recorded \`thread/status/changed(awaiting)\` notification already carries \`AskPending=true\` on origin/main, with no production change needed. The projector citation in that finding is real (it doesn't set the field itself) but isn't a bug — the server stamps it afterward, from the live session, at its single notification egress. No hub-side fix is needed; this PR's fixture fixes stand on their own.

obra:
Closing after round 5 exceeded the five-review-round budget (Jesse's 2026-09-17 rule: decompose rather than push a sixth round). This branch stays as reference; the change is split into three tiny stacked PRs, one mechanism each, plus incorporating #1732's still-unfixed High/Low findings into the piece that owns them:

- #1858 — reasoning export (joinedReasoningParagraphs; isActiveItem moved to itemFailure.ts and retyped so mobile can share it, closing the round-5 Low)
- #1859 — image folding + warning text (bounded raw-frame fallback; the itemSummary crash guard and mobile warning-text fix from #1732's round-3 review; drops the "turns-reference contract with no production consumer" Medium as speculative complexity with no reader)
- #1862 — ask resolution (askPending-gated liveAskQuestions; fixes every fixture the gate breaks, including two vacuous mobile tests the round-5 review flagged)

Each is stacked on the previous, based ultimately on main, CI green + gates run. #1732 (piece B, warning/reasoning web) will restack onto #1862's branch once it lands, dropping whatever pieces 1-2 already carry.

Issue #1634, opened 2026-09-17
hub: make 'forgot to broadcast after a write' structurally impossible (post-commit hook on stores)


obra:
Found by the /simplify altitude pass over #1602 (#1543 class).

After #1602 every write handler in `cmd/evener-hub/app_rpc.go` pairs `writeDidApply(err)` with the matching `notify*(server)` by hand, ~20 times. Nothing stops the next RPC handler from writing to a store and returning without a notify, which is exactly the bug class #1543 fixed for the current sites.

Deeper shape: each store / manager emits a post-commit event (or takes an `onCommitted` port) and one dispatcher turns it into the `Broadcast`/`BroadcastAll` call, so a handler cannot commit without the notification firing. Design first (which stores, how the hub wires the port, how tests assert it); this touches hubcore, plugins and the RPC layer, so it is several PRs, not one.

PR #1602, opened 2026-09-16
fix(hub): broadcast every applied write even when a later step fails (#1543, #1572)

obra:
Closes #1543. Closes #1572.

Both issues are one rule with three holes: a hub write that applied is announced to every client, whatever the step after it did. Since D11 (#1502) and #1532 the broadcast is the only reconciliation path both hosts have, so a write that landed and then hit a failing rollback or a failing listing left every other client's list wrong until a reconnect.

## One encoding, not three special cases

`app_write_applied.go` holds the whole rule: `appliedWriteError` wraps an error whose write stands, `writeApplied(err)` marks one, and `writeDidApply(err)` answers the handlers' only question — nil (the ordinary applied write) or marked (applied, then failed). Wrapping keeps the message and the wire class of what it wraps, because `appserver`'s router resolves a `WireError` through `Unwrap` (`internal/appserver/router.go:83`).

`renamePersistedError` is deleted: it was this rule already, written for one instance of it. Its two construction sites now return `writeApplied(...)`, and the Edit handler's `errors.AsType[renamePersistedError]` special case is gone — every instance write goes through one `instanceWrite` helper that announces when `writeDidApply(err)` and returns the error regardless.

## The sites

**Instances (#1543)** — three, not the two the issue names, because `writeAndReload` carries the same double failure for two more methods:
- `Create` (`app_instances.go:879`), reload failed and the rollback write failed: the new entry is still in the file.
- `writeAndReload` (`:1494`), same shape, serving `Edit` and `SetModelDisabled`.
- `Remove` (`:1292`), "the rollback could not be written, so the removal stands in the config".

**Plugins and marketplaces (#1572)** — the issue names the plugin half; the marketplace half of the same file is identical, so both are fixed. Every write answers through `pluginListAfterWrite` / `marketplaceListAfterWrite`, which mark a failing listing as applied; the two plain read methods (`ListPlugins`, `ListMarketplaces`) are untouched. That is ten write paths: install, upgrade, remove, enable, disable, setAutoUpgrade, and marketplace add/remove/refresh/edit.

**Handlers** — every broadcast gate now asks the one question, `writeDidApply(err)` instead of `err == nil`, so no site can drift back. For auth and launch that is a no-op today (their controllers mark nothing), and it is the same sentence in all of them rather than two spellings of one rule.

## The survey you asked for

`err == nil` guarding a broadcast, across `cmd/evener-hub` (18 gates, all in `app_rpc.go`):

| Family | Gates | Verdict |
[...]

PR #1733, opened 2026-09-18
plugins: broadcast every applied store write via a post-commit hook (#1634)

obra:
## What

Mechanism piece of the #1602 decomposition (Jesse's ruling, 2026-09-17): derive
"the store changed" centrally instead of threading a `StoreChanges` return
through every caller (#1634's shape). Built fresh on `main` — `main` has no
`StoreChanges` type at all yet, so this does not touch #1602's branch or its
hand-threaded code.

`internal/plugins.Manager` gains:
- `StoreChanged{Plugins, Marketplaces bool}` — which of the two persisted
  files (`installed_plugins.json`, `known_marketplaces.json`) changed.
- `saveRegistry` and `saveMarketplaces` each mark their own flag immediately
  after their own write succeeds (`markStoreChanged`), never before, and
  never by the function that calls them.
- `lockStore`'s release captures and clears the session's accumulated flags
  (still holding the flock — see "measurement" below) and, once the lock is
  let go, hands them to an installed `OnStoreChanged(func(StoreChanged))`
  callback — never invoked when nothing was written.

The hub installs that callback (`wirePluginStoreBroadcast`, `app_rpc.go`) on
the plugins controller's Manager and the auto-upgrade Manager(s), broadcasting
`evener/marketplace/updated`/`evener/plugin/updated` for whatever changed. This
is *in addition to* every existing per-handler `notify*` call in
`app_rpc.go`/`app_plugin_autoupgrade.go` — left untouched per the brief, since
a follow-up PR deletes them once this hook is proven in. A handler covered by
both just broadcasts twice, which is harmless.

## Bug this closes today

The auto-upgrade daemon tick and `evener/plugin/checkNow` refresh every known
marketplace before upgrading plugins (`runPluginAutoUpgradeTick`), and
`RefreshMarketplace` always writes `known_marketplaces.json` (at minimum
`LastUpdated`) whether or not the pull changed anything — but neither call
site ever called `notifyMarketplaceUpdated`. That write went out with no
broadcast at all. `TestRunPluginAutoUpgradeTick_BroadcastsAfterAMarketplaceRefreshThatWrote`
and `TestRegisterPluginAutoUpgradeHandlers_CheckNowBroadcastsARefreshWithNoUpgrade`
reproduce it (fail without `wirePluginStoreBroadcast`) and prove it fixed.

## Measurement: is `lockStore`'s release the right reporting point?

Yes, with one twist. `defer release()` runs on every caller in this package,
so a caller that returns early on error still reports — that part of the
brief's worry doesn't reproduce. What I did find by writing the reporting
code and a same-session-then-clear-and-report sequence:

**Capture must happen before the OS lock is released, not after.** `List()`
reads `installed_plugins.json` without ever taking the store lock. If release
ran first and the flag-capture-and-clear ran after, a second `lockStore`
acquisition (a different goroutine, since flock is per open-file-description
and correctly excludes goroutines too) could start and reset the field
concurrently with this session's own read of it — a plain data race with
nothing left to keep them apart. So `lockStore`'s wrapped release captures
and clears `pendingStoreChanged` first, *then* releases the flock, *then*
invokes the callback — the callback (a broadcast) never runs holding the
file lock, but the field access is still fully inside it.

**Only the two file-writing primitives get a direct hook — not `moveMarketplace`,
`swapInClone`, or a directory removal, despite Table 2 of the audit listing
them.** `List`/`ListMarketplaces` answer entirely from the two JSON files;
they never look at directory presence. `moveMarketplace` and `swapInClone`
both run as one step of a larger rename/edit/add that rolls the directory
move back (`runUndo`, `undoSwap`) if a later step fails — most often the very
next line, the final `saveRegistry`/`saveMarketplaces`/`saveRename` call.
Marking either of them directly would report a change nothing could observe
and that the same session's own rollback might immediately undo, which is
exactly the false "something changed" signal this mechanism exists to rule
out — a new instance of the bug class, not a fix for it.

I proved this both ways:
- `TestAddMarketplace_SwapSucceedsThenSaveFailsReportsNoChange`: `swapInClone`
  swaps a fresh clone into place, `saveMarketplaces` then fails persistently,
  `AddMarketplace`'s own cleanup removes the swapped-in clone. Callback fires
  zero times.
- `TestMigrateMarketplaceName_MoveSucceedsThenSaveFailsReportsNoChange`: the
  migration barrier's `moveMarketplace` (`itsOwn` branch) renames the clone
  and cache directories for a legacy `"foo@bar"` entry, then `saveRename`'s
  first call (`saveRegistry`) fails persistently; `runUndo` puts the
  directories back. Callback fires zero times.
- `TestSaveRename_MarksBothOnSuccess` / the registry-restore path inside
  `saveRename` (failed `saveMarketplaces` after a successful `saveRegistry`,
  restoring `registryAsFound`) *does* mark `Plugins` twice — deliberately not
  suppressed: `List`'s registry read is unlocked, so a concurrent reader could
  really have observed the transient write before the restore. Reporting it
  is the conservative, correct answer; the reverse (a real, observable write
  with no report) is the bug this exists to prevent.

`saveRename` gets no hook of its own — it only ever calls `saveRegistry` then
`saveMarketplaces`, so both flags are set by the primitives it calls. Every
migration/rename/merge path in `marketplace_migration.go`
(`recoverMarkedRename`, `migrateMarketplaceNames`, `mergeIntoMigrated`,
`migrateMarketplaceName`) funnels through `saveRename`/`saveMarketplaces` or
fails before reaching either, so all four of the audit's real Mediums
(`recoverMarkedRename`'s upfront-set, `migrateMarketplaceNames`'s
`ErrStoreChanged`-only-partial-write gap, and the migration/marker-recovery
paths generally) become impossible by construction with no code in that file
needing to know this mechanism exists.

## Primitives hooked (Table 2 of the audit, `ab77dcfe1` comment)

| Primitive | Hooked? | Why |
|---|---|---|
| `saveRegistry` | Yes, directly | writes `installed_plugins.json` |
| `saveMarketplaces` | Yes, directly | writes `known_marketplaces.json` |
| `saveRename` | Composed | calls both of the above |
| `moveMarketplace` | No | directory-only; see measurement above |
| `swapInClone` | No | directory-only; see measurement above |
| directory remove (`RemoveMarketplace`'s clone delete, `Remove`'s cache delete) | No | best-effort, unconditional on the outcome; the operation's own subsequent `saveMarketplaces`/`saveRegistry` already reports what a client could observe |
| marker/record-file writes (`writeRenameMarker`, `saveMigrationRecord`) | No (per the brief — touch neither tracked file) | |

## Hub wiring scope

Wired: the plugins controller's Manager (`newHubPluginsController`'s `mgr` —
covers every `evener/marketplace/*` and `evener/plugin/*` RPC, `Preview`, and
the migration barrier on every list/browse); the auto-upgrade Manager backing
`evener/plugin/checkNow`; the periodic auto-upgrade daemon's own Manager
(`main_background.go`).

Not wired (deviation, flagged rather than guessed at): `hubCommandList`
(`app_rpc.go`) and `hubResolvePlugins` (`app_threadlifecycle.go`, thread
spawn / slash-catalog) each construct their own `plugins.NewManager` inline,
read-mostly (`ResolveForLaunch`/`PreviewForLaunch`), and the only write they
can trigger is the same migration-on-first-list already covered by every
other read path. Wiring `hubResolvePlugins` specifically means changing a
package-level func-var's signature that a dozen+ existing tests across
`app_spawn_slashcatalog_test.go` and `app_threadlifecycle_plugin_test.go`
substitute directly — real plumbing, not a mechanism change, and outside a
~300-line mechanism-only PR. `hubSeedDefaults`/`hubPluginGC`
(`main_background.go`) are deliberately left unwired too, matching the
audit's own documented correctness: both run once at hub start, before
`Serve()` begins, so there is no client yet to tell.

## Gates

- `$(go env GOROOT)/bin/gofmt -l cmd/evener-hub internal/plugins`: clean.
- `go vet ./cmd/evener-hub/... ./internal/plugins/...`: clean.
- `go vet -tags evenerfuzz ./cmd/evener-hub/... ./internal/plugins/...`: clean.
- `GOOS=windows go vet -tags evenerfuzz ./cmd/evener-hub/... ./internal/plugins/...`: clean.
- `go test -count=1 ./cmd/evener-hub/... ./internal/plugins/...`: all green.
- `go test -tags evenerfuzz -count=1 ./cmd/evener-hub/... ./internal/plugins/...`: all green.
- No fuzz-tagged test or the code it covers changed; `make fuzz-seeds` not run.
- No `make/*.mk` touched; root Makefile audits not run.

Refs #1634, #1602.


obra:
## Review round disposition (f23e448ef)

Head: `f23e448efdd6957f146a3bfab0f43fd4ceb4b086`
Mergeable: `MERGEABLE`

Folds the simplify pass and RoboRev round 1's two Mediums into one push, TDD where behavior changed.

### Simplify F1: `hostNotificationBroadcaster` instead of vars (fixed)

Reverted `notifyMarketplaceUpdated`/`notifyPluginUpdated` to plain funcs; changed their parameter (and `wirePluginStoreBroadcast`'s) from `*appserver.Server` to `hostNotificationBroadcaster` (`app_host_admin.go:245-249`). `*appserver.Server` satisfies it, so all 14 real call sites and both fuzz tests compile unchanged. `plugin_store_broadcast_test.go` now drives `wirePluginStoreBroadcast` with `recordingBroadcaster` (`app_host_admin_test.go`) and asserts on the real `appwire.NotifyEvenerMarketplaceUpdated`/`...PluginUpdated` methods instead of a call count.

### Simplify F2: acquire-time reset comment + redundant test resets (fixed)

`lockStore`'s reset only defends a write outside any lockStore session (today only `plantLegacyMarketplace`) — commented in place. Removed the two now-redundant `m.pendingStoreChanged = StoreChanged{}` lines in `store_changed_test.go`; confirmed by rerunning both affected tests (`TestMigrateMarketplaceName_MoveSucceedsThenSaveFailsReportsNoChange`, `TestMigrateMarketplaceNames_PersistentPreWriteFailureNeverReportsChange`) green without them — `lockStore`'s own reset is what makes that safe.

### Simplify F3: wire hubSeedDefaults / hubPluginGC (fixed, wired rather than exempted)

Measured the exemption path first: `main.go:373` binds `hubListener` but nothing calls `Accept`/`Serve` on it until `deps.serve(ctx, srv)` at `main.go:887`; `seedHubMarketplaces` (`main.go:636`) and `startHubPluginMaintenance` (`main.go:645`) both run strictly between those two lines, so no client can be connected when either runs — a provable, documentable exception. Wired anyway: `hubSeedDefaults`/`hubPluginGC` now take a `*plugins.Manager` the caller constructs and wires (`wirePluginStoreBroadcast`) before calling in, matching `hubStartUpgrade`'s existing shape. Reasoning: leaving them unwired reintroduces exactly the "remember this special case" burden #1634 exists to remove, and today's ordering (seed/gc before `ListenAndServe`) is not an invariant future code is obliged to preserve. New tests: `TestHubSeedDefaults_ASeedThatWritesBroadcasts`, `TestHubPluginGC_MigratesALegacyNameAndBroadcasts` (plants a `"foo@bar"` legacy entry via a new `plantRefusedMarketplace` helper; `Gc`'s own `lockStore` acquisition migrates it and broadcasts both notifications). Both watched red (`wirePluginStoreBroadcast` neutralized) before green.

### Medium 1 (paths.go:22-32 / store_changed.go:60 / locks.go:87,102-105): mutex, not just the flock (fixed)

Measured before fixing: wrote `TestManager_ConcurrentLockSessionsDoNotRaceStoreChanged` (8 goroutines, each `AddMarketplace`-ing a distinct directory-source marketplace onto one shared `Manager`) and ran it under `-race` against the un-mutexed code — it reproduced real data races on `pendingStoreChanged` (write/write in `lockStore`↔`captureStoreChanged`, and in `markStoreChanged` itself), confirming the reviewer's point: flock serializes wall-clock access but gives the race detector no happens-before edge, and this package's own stubbed `lockAcquirer` test seams (`installAcquireLock`, `marketplaceAcquireLock`, `gcAcquireLock`) can bypass the real lock entirely in other tests.

Fix: added `storeChangedMu sync.Mutex` to `Manager`, held only for the instant of each reset/mark/capture/install/read — never across the file-lock wait. `resetStoreChanged`, `markStoreChanged`, `captureStoreChanged`, `OnStoreChanged`, and a new `storeChangedCallback` (the read side) all take it. `go test -race ./internal/plugins/...` is clean, including the new concurrency test and the full existing suite.

### Medium 2 (app_rpc.go / app_threadlifecycle.go, issue #1734): measured, partially fits

**`hubCommandList`: wired, fits.** Changed its signature to take `hostNotificationBroadcaster`, constructs and wires its own `Manager` before calling `ResolveForLaunch`. Updated the one production call site (`app_rpc.go`, the `evener/command/list` handler closure, which already has `server` in scope) and all 10 test call sites (`app_command_list_test.go` ×7, one each in `cov_exact_app_rpc_fuzz_test.go`, `cov_rpc_sources_pass6_fuzz_test.go`, `cov_rpc_threads_helpers_fuzz_test.go`) to pass `newRecordingBroadcaster()`. New integration test `TestHubCommandList_MigratesALegacyNameAndBroadcasts`: plants a `"foo@bar"` legacy marketplace, calls `hubCommandList`, asserts both notifications fire from the migration `ResolveForLaunch → List` triggers — watched red or green by toggling `wirePluginStoreBroadcast`. Net: signature change + ~14 call-site edits + one new test, comfortably inside the ~60-line budget.

**`hubResolvePlugins`: measured, does not fit — #1734 stays open for it.** It is a package-level `var` (`app_threadlifecycle.go:44-46`) with signature `func(ctx, pluginRoot, dirs, enabled) (LaunchPluginResolution, error)` — no broadcaster in scope to add. It is called from `hubThreadStart` (`app_threadlifecycle.go:135`, no `server`/broadcaster param either) and `hubSpawnSlashCatalog` (`app_spawn_slashcatalog.go:91`). `hubThreadStart` alone has 38 call sites across 11 files; the var itself is substituted directly (not through a factory) in `app_spawn_slashcatalog_test.go:239-243` and `app_threadlifecycle_plugin_test.go:266-268`, each pinning its current 4-argument signature. Threading a broadcaster in means changing `hubThreadStart`'s signature (and everything that calls it) or inventing a second, parallel construction path just for the migration case — real plumbing through the thread-lifecycle surface, not a mechanism change, and far past a ~60-line budget. Leaving #1734 open, scoped down to this one function.

### Low (deferred per Jesse's rule)

Noted, not fixed here: successful RPC mutations and auto-upgrades now broadcast twice (this hook + the existing manual `notify*` call), and `EditMarketplace`'s no-op early return still broadcasts once via the manual path. Both go away with the follow-up PR that deletes the manual per-handler broadcasts once this hook is proven in.

### Gates

- `$(go env GOROOT)/bin/gofmt -l cmd/evener-hub internal/plugins`: clean.
- `go vet ./cmd/evener-hub/... ./internal/plugins/...`: clean.
- `go vet -tags evenerfuzz ./cmd/evener-hub/... ./internal/plugins/...`: clean.
- `GOOS=windows go vet -tags evenerfuzz ./cmd/evener-hub/... ./internal/plugins/...`: clean.
- `go test -count=1 ./cmd/evener-hub/... ./internal/plugins/...`: all green.
- `go test -tags evenerfuzz -count=1 ./cmd/evener-hub/... ./internal/plugins/...`: all green.
- `go test -race -count=1 ./internal/plugins/...`: clean (new concurrency test included).
- No `make/*.mk` touched; root Makefile audits not run. No fuzz seed corpus changed; `make fuzz-seeds` not run.

obra:
## Review round disposition (5b9e755)

Head: `5b9e75505c4e87a9eda3bef09794abea7c74f7cc`
Mergeable: `MERGEABLE`

### CI red: golangci-lint modernize/rangeint (fixed)

Confirmed cause (coordinator): `lint-golangci` on the root module, `modernize` linter's `rangeint` rule, at `internal/plugins/store_changed_test.go:211:6` and `:249:6`. Both were `for i := 0; i < n; i++`. Changed to `for i := range n` / `for i := range 2`. Ran `golangci-lint run ./...` from the repo root with a fresh cache (`golangci-lint cache clean`) — `0 issues`.

### Medium (third raise, #1734): `hubResolvePlugins` wired through the controller's Manager — closed, then a new race found and fixed

Took the reviewer's cheaper route: no signature change to `hubThreadStart`/`hubSpawnSlashCatalog`. `app_threadlifecycle.go` adds `hubPluginResolveManager(pluginRoot)`, which defers to an installable selector (`hubResolvePluginManagerFor`) and falls back to a fresh `plugins.NewManager(pluginRoot)` when none is set or it returns nil for that root — exactly today's behavior for every test that never builds a server. `registerRPCHandlers` (`app_rpc.go`) installs the selector once per server construction, pointing at `pluginsController.mgr` when the requested root matches `cfg.PluginRoot` (always true in production). New regression test `TestHubSpawnSlashCatalog_MigratesALegacyNameAndBroadcasts` (spawn-catalog boundary, legacy `"foo@bar"` marketplace) watched red/green.

Running this locally with `-race` (the CI race gate's own reason for asking) caught a **second, independent data race** this round introduced: `hubResolvePluginManagerFor` was a plain package var, and `registerRPCHandlers` writes it on every hub server construction — many tests build servers with `t.Parallel()`, so concurrent constructions raced the assignment (and the concurrent read inside `hubResolvePlugins`). Reproduced with a minimal, targeted test (`TestNewWebServer_ConcurrentConstructionDoesNotRaceResolvePluginManager`, 8 goroutines each constructing a `NewWebServer` and calling `hubResolvePlugins`) — confirmed red under `-race` against the unguarded var, then fixed by guarding it with `hubResolvePluginManagerMu sync.RWMutex` (a `setHubResolvePluginManagerFor` writer, `hubPluginResolveManager` reads under `RLock`) — confirmed green. The two direct-substitution tests (`app_spawn_slashcatalog_test.go`, `app_threadlifecycle_plugin_test.go`) replace `hubResolvePlugins` wholesale and never touch this at all; my own new regression test now goes through `setHubResolvePluginManagerFor` for symmetry, though as a non-parallel test it was never actually racing (Go runs all non-parallel top-level tests to completion before starting the parallel batch).

Closing #1734.

### Local -race on cmd/evener-hub: inconclusive, not run further (per the coordinator's bound)

`go test -race -short -count=1 ./cmd/evener-hub/...` was run twice locally to chase the review's synchronization concern. The first full run (before the fixes above) surfaced the real race documented above via its stack traces, which is what led to the fix. Reruns after the fix did not finish within two attempts — the package's own test suite (500+ tests, many building real hub servers) is expensive under `-race`, and per the brief `-race` is CI's job, not something to keep re-running locally to convergence. The targeted regression test (`TestNewWebServer_ConcurrentConstructionDoesNotRaceResolvePluginManager`) and the two directly-implicated tests (it, and `TestHubSpawnSlashCatalog_MigratesALegacyNameAndBroadcasts`) both pass clean under `-race` in isolation. **`internal/plugins` under `-race -short` passed in full**, including the concurrency test from the previous round (`TestManager_ConcurrentLockSessionsDoNotRaceStoreChanged`). CI's `race-modules` job is the full-suite oracle for `cmd/evener-hub`; if it finds anything else, the log will name the test and it gets fixed from there.

### Lows (still deferred per Jesse's rule, unchanged from last round)

- Duplicate notifications for Manager-backed writes (`app_rpc.go`, `app_plugin_autoupgrade.go`) — resolved by the follow-up PR that deletes the manual `notify*` calls.
- Stale comment at `internal/plugins/store_changed.go:29` naming a nonexistent `reportStoreChanged` — fix goes in the same follow-up PR.

### Gates

- `$(go env GOROOT)/bin/gofmt -l cmd/evener-hub internal/plugins`: clean.
- `go vet ./cmd/evener-hub/... ./internal/plugins/...`: clean.
- `go vet -tags evenerfuzz ./cmd/evener-hub/... ./internal/plugins/...`: clean.
- `GOOS=windows go vet -tags evenerfuzz ./cmd/evener-hub/... ./internal/plugins/...`: clean.
- `golangci-lint run ./...` (repo root, root module, fresh cache): `0 issues`.
- `go test -count=1 ./cmd/evener-hub/... ./internal/plugins/...`: all green.
- `go test -tags evenerfuzz -count=1 ./cmd/evener-hub/... ./internal/plugins/...`: all green.
- `go test -race -short -count=1 ./internal/plugins/...`: clean.
- `go test -race -short -count=1 ./cmd/evener-hub/...`: inconclusive (did not finish in two local attempts); the two directly-relevant tests pass clean under `-race` run in isolation; CI's `race-modules` job is the oracle for the rest.

obra:
## Review round disposition (acd31957)

Head: `acd319572238aca16a1b84706ae2459ea5423401`
Mergeable: `MERGEABLE`

Round 4. Closed both Mediums at the root (not per-instance) and both Lows.

### Medium 1: package-global `hubResolvePluginManagerFor` is unsafe with two live WebServers — fixed at the root

Deleted the package global, its mutex (`hubResolvePluginManagerMu`), and the setter (`setHubResolvePluginManagerFor`) entirely.

Measured which of "server struct field" vs "WebConfig field" reaches `hubThreadStart`/`hubSpawnSlashCatalog` with the fewest lines: both already take `cfg hubcore.WebConfig` directly, so a `WebConfig` field needs no new parameter on either. Added `hubcore.WebConfig.PluginResolveManager func(pluginRoot string) *plugins.Manager` (`cmd/evener-hub/internal/hubcore/config.go`; `internal/plugins` has no import back into `cmd/evener-hub`, so no cycle).

`newHubAppServerWithNavigationAndTrace` (`app_rpc.go`) now builds `pluginsController` before `registerThreadHandlers`, wires `wirePluginStoreBroadcast`, and sets `cfg.PluginResolveManager` on its own local `cfg` — moved earlier in the function specifically because `registerThreadHandlers`/`registerMiscHandlers`'s closures capture whatever `cfg` value was live at their own call time (Go structs pass by value). `hubResolvePlugins` (`app_threadlifecycle.go`) now takes the resolver as a 5th parameter instead of reading any global; both callers pass `cfg.PluginResolveManager`.

Found and fixed a second bug while building the regression test: `cfg` being pass-by-value also meant the wired `PluginResolveManager` never reached `WebServer.cfg` itself (only the registered-closures' own copy) — `newHubAppServerWithNavigationAndTrace` now returns the final `cfg` as a third value, and `web.go`'s `newWebServer` assigns `web.cfg` from it (mirroring how `web.cfg.LiveModels` is already set before that call). Watched the regression test null-pointer-panic without this fix, pass with it.

Regression test `TestNewWebServer_TwoServersEachResolveTheirOwnPluginManager`: constructs two `NewWebServer`s at different plugin roots, asserts each resolves its own distinct `*plugins.Manager`, and that re-resolving through the first server's `cfg` after the second is constructed still returns the first server's own Manager pointer — watched it panic (nil `PluginResolveManager`) before the `web.cfg` fix, pass after.

Also resolves Low 1 (the unsynchronized read in `app_spawn_slashcatalog_test.go:421`): that variable no longer exists: the test now sets `cfg.PluginResolveManager` on a locally-constructed `hubcore.WebConfig` value, nothing shared.

### Medium 2: duplicate notifications — the hook is now the sole notification path

Audited every manual `notifyMarketplaceUpdated`/`notifyPluginUpdated` call in `registerPluginHandlers` (10 call sites) and `app_plugin_autoupgrade.go` (2: the daemon tick, the checkNow handler) against `internal/plugins`' write primitives (confirmed by reading `hubPluginsController`'s methods, `app_plugins.go:257-410`): every one of `AddMarketplace/RemoveMarketplace/RefreshMarketplace/EditMarketplace/Install/Upgrade/Remove/Enable/Disable/SetAutoUpgrade`, on success, always goes through `saveRegistry` and/or `saveMarketplaces` (directly or via `saveRename`) — both hooked primitives. No write on this surface bypasses a hooked primitive, so every manual call was deletable; none had to be kept. (`EditMarketplace`'s own no-op early return — same name, same source — was actively wrong before this: the manual call broadcast unconditionally on `err == nil`, even though nothing was written; the hook correctly stays silent there.)

Deleted all 12 manual calls. `startPluginAutoUpgradeDaemon` and `registerPluginAutoUpgradeHandlers`'s checkNow handler no longer need a broadcaster for this purpose; `startPluginAutoUpgradeDaemon` dropped its now-unused `server *appserver.Server` parameter (4 call sites updated: `main_background.go`, `cov_plugins_fs_pass3_fuzz_test.go`, `cov_launch_models_plugins_fuzz_test.go` ×2).

Regression test `TestRegisterPluginHandlers_MarketplaceEditBroadcastsExactlyOnce`: dispatches a re-source-only `evener/marketplace/edit` (no rename — `EditMarketplace`'s non-renaming branch calls `saveMarketplaces` alone, never `saveRegistry`) through the real RPC router, with `wirePluginStoreBroadcast` wired to a `recordingBroadcaster`, and asserts exactly one broadcast (`evener/marketplace/updated`). Watched green both before and after re-adding a manual call in a throwaway revert — see the honest limitation below — then confirmed by grep that zero `notifyMarketplaceUpdated(`/`notifyPluginUpdated(` calls remain outside `wirePluginStoreBroadcast`'s own definition.

**Honest limitation on this test**: it can only observe broadcasts sent to the fake `recordingBroadcaster` wired to the hook; a manual call (if one existed) targets the real `*appserver.Server` `registerPluginHandlers` receives for routing, which this test's broadcaster can't intercept (no exported way to register a receiving connection on a real `appserver.Server` from outside that package — the same constraint every broadcast test in this PR has worked within). So "no duplication" rests on two things together: this test proving the hook alone fires exactly once for this scenario, and the diff itself (`git diff`, reviewed above) showing zero manual calls remain. I did not invent a way around the constraint rather than claim the test proves more than it does.

### Low 2: isolated the ambient launch/config home in the round-3 test

`TestHubSpawnSlashCatalog_MigratesALegacyNameAndBroadcasts` now sets `t.Setenv("XDG_CONFIG_HOME", xdg)` and `LaunchConfigRoot: xdg` on its `hubcore.WebConfig`, copying the exact idiom every neighboring test in `app_spawn_slashcatalog_test.go` uses.

### Low 3: fixed the comment naming a nonexistent function

`app_threadlifecycle.go`'s `hubResolvePlugins`/resolver comments were rewritten as part of the Medium 1 fix and no longer name `registerRPCHandlers` (which never existed) — they name the real constructor, `newHubAppServerWithNavigationAndTrace`, and state the current invariant only. Also fixed a second stale-name Low from the previous round while I was in the area: `internal/plugins/store_changed.go:29` named a nonexistent `reportStoreChanged`; corrected to name `captureStoreChanged`/`storeChangedCallback`.

### Gates (per BRIEF-COMMON.md, root module, touched packages only — no `-race`, no `make lint`, no full suites)

- `$(go env GOROOT)/bin/gofmt -l cmd/evener-hub internal/plugins`: clean.
- `go vet ./cmd/evener-hub/... ./internal/plugins/...`: clean.
- `go vet -tags evenerfuzz ./cmd/evener-hub/... ./internal/plugins/...`: clean.
- `GOOS=windows go vet -tags evenerfuzz ./cmd/evener-hub/... ./internal/plugins/...`: clean.
- `golangci-lint run ./...` (repo root, root module): `0 issues`.
- `go test -count=1 ./cmd/evener-hub/... ./internal/plugins/...`: all green.
- `go test -tags evenerfuzz -count=1 ./cmd/evener-hub/... ./internal/plugins/...`: all green.

Non-test diff (this round, cumulative vs. `origin/main`): 14 files, 310 insertions / 101 deletions.

obra:
Round 4 @acd31957: CI green (15), RoboRev has one Low (the app_plugins.go mutex comment). Merging per the Lows rule; the Low plus the eleven /simplify items (one plugin manager per hub injected as a pointer, hook as the sole notifier helper, resetStoreChanged removal, single-lock takeStoreChanged, comment and test-helper cleanups) land in an immediate follow-up PR from main.

PR #1735, opened 2026-09-18
fix(agent,web): name a drain's consumed queue intents on the queueChanged push

### LINKED ISSUE #1704: Cross-tab drain-cleanup can retire a still-queued turn/queue intent (sequence-based inference is unsafe across dispatcher instances)

obra:
## Summary

Closes #1704.

The web client had no way to know which optimistic `turn/queue`/`turn/drainAsSteer`
records a drain (`turn/drainAsSteer`) consumed. #1452 inferred it from sequence
numbers, which is wrong across tabs sharing one outbox (no cross-tab lease); #1705
built a 1230-line client-side fence mechanism to compensate and was closed per
Jesse's ruling: the daemon already knows the exact consumed ids at drain time, so
the small fix is to just tell the client.

- `agent/session_client_mutation_queue.go`'s `clientMutationDrain` now captures
  the consumed entries' `ClientMutationID`s from its own effect closure and passes
  them through `reflectDurableInputQueue`'s new variadic parameter (all 8 other
  call sites are unchanged, since they pass none).
- `events.QueueChangedData` and `appwire.ThreadQueueChangedParams` gain
  `ConsumedClientMutationIDs []string` (`consumedClientMutationIds,omitempty`) —
  on the notification **params**, not on `QueueState`: the server persists `Queue`
  verbatim onto the cached thread (`server/thread_envelope.go`'s `assign`), and a
  one-shot transition fact must not live in durable state.
- The projector copies it through unchanged.
- Web: `notificationMutationIdentities` (`threads.ts`) folds
  `consumedClientMutationIds` into the ids it hands `reconcileIdentities`, so a
  drain's push settles those optimistic records the same way an authoritative
  snapshot naming a survivor already does.
- Deleted the `#1452` `turn/drainAsSteer` retirement branch in
  `MutationDispatcher#attempt` (and its two tests): it inferred "accepted before
  the drain" from `intentSequence`, which only holds within one dispatcher
  instance's own per-target FIFO — the exact hazard #1704 named.
- Added a `cmd/evener-tui` case proving the TUI (which does consume
  `thread/queueChanged`, for `params.Queue` only) tolerates the new field with no
  behavior change.

**Compatibility**: an old daemon never sends `consumedClientMutationIds`
(`omitempty`, absent — never `[]`). The client then behaves exactly as it does
today: it degrades to the pre-existing ghost-row case for a same-tab drain (no
new failure mode, nothing removed that wasn't removed before this PR), and never
deletes live work. No version bump; no protocol negotiation needed.

The hydrate-only backstop for a missed push is a separate, later PR — not built
here.

## Test plan

- [x] `agent/session_client_mutation_queue_test.go`:
      `TestClientMutation_DrainEmitsConsumedClientMutationIDs` — RED first
      (compile failure against the pre-change `QueueChangedData`), then green.
- [x] `internal/appprojector/appwire_projection_test.go`:
      `TestAppEventProjectorCopiesConsumedClientMutationIDs` and
      `TestAppEventProjectorOmitsConsumedClientMutationIDsWhenAbsent` (the
      encoding-level omitempty assertion: a non-drain push's JSON has no
      `consumedClientMutationIds` key at all).
- [x] `cmd/evener-tui/hub_appwire_test.go`: extended
      `TestHubModelAppliesQueueChangedNotification` with a case carrying
      `ConsumedClientMutationIDs` alongside `Queue`, asserting the composer
      preview is unaffected.
- [x] `cmd/evener-hub/frontend/src/stores/threads.test.ts`: "a queueChanged naming
      consumed ids settles the matching optimistic queue record" — RED first
      (asserted `false`), then green after the `notificationMutationIdentities`
      change.
- [x] `make generate` — confirmed the only generated diff is one
      `consumedClientMutationIds?: string[];` line in
      `appwire-client/typescript/types.gen.ts` (plus the matching
      `docs/appwire-protocol.md` table row).
- [x] Gates run (all green): toolchain gofmt on `agent`, `internal/appprojector`,
      `appwire`, `cmd/evener-tui`, `server`; `go vet ./...`,
      `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`
      for both the root module and `agent` module; `go test ./agent/...` (the
      `TestRetirement*`/`TestScratchRetention*` failures are a pre-existing
      environment issue, confirmed by reproducing them against an unmodified
      `origin/main` checkout — unrelated to this diff's files); `go test
      ./internal/appprojector/... ./appwire/... ./server/... ./cmd/evener-tui/`;
      `npx vitest run src/stores/threads.test.ts src/stores/mutationDispatcher.test.ts
      src/stores/mutationDispatcher.edge.test.ts` (349 passed); `npm run
      typecheck`; `npm run lint` (biome, clean); `make lint-package-imports`.

## Residuals (not fixed here, filed for visibility)

- `pushQueueHead`'s empty-`ClientMutationID` branch (`agent/session_queue.go:855-862`)
  is, in production Go, reachable only from tests today (grepped every call
  site). Left untouched: it's orthogonal to this field (it never carries
  `ConsumedClientMutationIDs`), and touching it risks the covtest/lifecycle-fuzz
  coverage the brief flagged as depending on it.


obra:
## Simplify disposition (c8a34ea0f)

Applied all four items from the opus simplify pass on @832a5ba70:

1. `agent/session_client_mutation_queue.go`'s `reflectDurableInputQueue` now calls `clientMutationQueueClientMutationIDs(snapshot.InputQueue)` instead of duplicating its own helper inline. Left the fused loop at `queuedInputFromClientMutation`'s call site untouched, per the note.
2. `internal/appprojector/appwire_projection_test.go`'s `TestAppEventProjectorOmitsConsumedClientMutationIDsWhenAbsent` doc comment is now the two-line invariant: positive evidence, absent key means nothing consumed, never `[]`.
3. `threads.ts:1415`'s comment now names the wire field `consumedClientMutationIds`, not the Go struct field.
4. Dropped the `#1452`/`#1705` lineage narration from `threads.test.ts` and `agent/session_client_mutation_queue_test.go`'s new-test comments; kept the invariant and the `#1704` citation in both.

Skipped per the disposition: `reflectDurableInputQueue` stays variadic.

Gates (all green): `go test ./agent/ -run "ClientMutation|Queue|Drain"`, `go test ./internal/appprojector/...`, `npx vitest run src/stores/threads.test.ts src/stores/mutationDispatcher.test.ts` (338 passed), `npm run typecheck`, `npm run lint` (biome, clean).

Head: `c8a34ea0feb1094de9b022d677d8310bd0be53bb`. `gh pr view 1735 --json mergeable,mergeStateStatus` → `MERGEABLE` / `BLOCKED` (required checks still running).

obra:
## Review round disposition (5f59aee3c)

CI was red on two jobs: `lint-golangci` and `static`. `static` is the aggregator ("Require every static lane") — its own log shows every other lane green and only `lint-golangci: failure`, so it's not an independent second failure, just the downstream gate reporting the same root cause.

**Medium / lint (fixed):** `agent/events/payloads.go:460` — `ConsumedClientMutationIDs` used a camelCase tag on an internal event payload; the agent module's tagliatelle config requires snake_case there. Changed to `json:"consumed_client_mutation_ids,omitempty"`. `appwire.ThreadQueueChangedParams` (the wire-level type) is unchanged and stays camelCase — it's already under this module's `json: camel` override for `appwire`. Reproduced the failure locally first (`golangci-lint run --enable-only tagliatelle ./events/...` in `agent/`), confirmed the fix clears it, then ran the full `golangci-lint run ./...` for both the `agent` module and the root module's touched packages (`internal/appprojector`, `appwire`, `cmd/evener-tui`) — 0 issues everywhere.

**High (fixed):** consumed ids were only ever delivered by the live `queueChanged` push. `clientMutationDrain`'s replay branch returns before `reflectDurableInputQueue` runs, and a fresh `thread/read` after a restart never carries a now-terminal queue record at all — so a missed push, daemon restart, or a transient IndexedDB failure on the client's own reconciliation left an optimistic `turn/queue` record stranded with no way back.

Fix: `appwire.MutationReceipt` gains `ConsumedClientMutationIDs []string` (`consumedClientMutationIds,omitempty`), beside `QueueEntryIDs`. `clientMutationDrain` sets it on the receipt at first execution, before marshaling the durable result — so `replayClientMutationResult`'s unmarshal of that same JSON carries it through a replay automatically; no separate replay-path code needed. The web `MutationDispatcher#attempt` now settles `receipt.consumedClientMutationIds` through the existing `reconcileIdentities` path (the same one the push already uses), independent of whether any push ever arrives.

Wire-shaped tests, RED first:
- Go: `TestClientMutation_DrainReplayCarriesConsumedClientMutationIDs` (`agent/session_client_mutation_queue_test.go`) — drains once, replays the same `ClientMutationID`, asserts both the first response's and the replayed response's `Receipt.ConsumedClientMutationIDs` match. Failed with the real value vs. `[]` before the fix.
- Web: `mutationDispatcher.test.ts` — "a drain's receipt settles the queue intents it named as consumed" (receipt alone, no push, settles the optimistic record) and "settling the same consumed id from both the receipt and a later push is idempotent" (receipt settles first; a later `reconcileIdentities` call for the same id — standing in for the push arriving after — resolves without throwing and the record stays settled). Both failed before the `#attempt` change.

**Not built here, as directed:** the cross-tab missed-push case (another tab offline while the push goes out) — that's the separate, later hydrate-only backstop PR.

**Noted, not changed:** the frontend's swallow of a failed push-driven `reconcileIdentities` (`threads.ts:1816`, pre-existing `.catch(() => { /* a later snapshot or receipt retries */ })`) is unchanged. Its own comment is now literally more true: the drain's receipt is exactly the "later ... receipt" that retries the same settlement for the same-tab case this round covers. The genuinely-unrecoverable case (cross-tab, receipt not visible to the other tab) is the deferred backstop PR's problem.

Gates: toolchain gofmt (agent, internal/appprojector, appwire, cmd/evener-tui); `go vet`, `go vet -tags evenerfuzz`, `GOOS=windows go vet -tags evenerfuzz` (root + agent modules); `golangci-lint run ./...` (agent module) and `golangci-lint run ./internal/appprojector/... ./appwire/... ./cmd/evener-tui/...` (root) — 0 issues both; `go test ./agent/ -run "ClientMutation|Queue|Drain"`, `go test ./internal/appprojector/... ./appwire/... ./cmd/evener-tui/` — all green; `make lint-package-imports`; `npx vitest run src/stores/threads.test.ts src/stores/mutationDispatcher.test.ts src/stores/mutationDispatcher.edge.test.ts` (351 passed); `npm run typecheck`; `npm run lint` (biome, clean); `make generate` confirmed the only new generated diff is `consumedClientMutationIds?: string[];` on `MutationReceipt` in `types.gen.ts` (no docs/appwire-protocol.md change — `MutationReceipt` isn't one of that doc's top-level tables).

Head: `5f59aee3cf6f6024fc24f4fd0123e23ffde8b0f7`. `gh pr view 1735 --json mergeable,mergeStateStatus` → `MERGEABLE` / `BEHIND` (main has moved since this branch point; no conflicts — did not merge main in since it wasn't asked for here).

obra:
## Review round 3 disposition (8ae0d871c)

**High (fixed):** `mobile/src/services/conversation.ts:376`'s `decodeMutationResult` validates every mutation receipt against an exact-key allowlist (`exactObject`). A drain receipt carrying `consumedClientMutationIds` had a key the allowlist didn't know, so `service.drainAsSteer`/`service.steer(input, revision)` threw "receipt has unexpected keys" for any drain that actually consumed queue entries — after the server had already applied it, and a replay retry hits the identical throw forever. Fixed the same way `queueEntryIds`/`instanceId` already handle their own optional presence: push the key onto the allowlist only when `"consumedClientMutationIds" in result.receipt`, validate it as an array whose entries are each a non-empty string, and copy it onto `decoded`.

Grepped for other exact-key receipt validators, both mobile and web: `exactObject` is defined and used nowhere else in the repo — `mobile-native/src` has no receipt-handling code of its own (it goes through `mobile/src/services/conversation.ts` via the shared source tree), and the web dispatcher's own `mutationReceipt()` (`cmd/evener-hub/frontend/src/stores/mutationDispatcher.ts`) only checks that four REQUIRED fields are present and never rejects on an extra one — it was already forward-compatible with this field, which is why last round's web-side fix needed no allowlist change.

Test: `service.steer(input, expectedQueueRevision)` is the one public method that returns `decodeMutationResult`'s output directly (`drainAsSteer` itself just validates and returns the raw wire response, which was never exact-key-checked against this new key by that path — it always had the key on the wire, just discarded). Added two tests there, wire-shaped: a receipt naming `consumedClientMutationIds` keeps them on the decoded result; a receipt that names nothing (the pre-existing shape) has no such key on the decoded result at all.

**Medium (fixed):** `mutationDispatcher.ts:146-151` settled the drain's own receipt (removing/reclassifying its outbox record) before reconciling the ids it named as consumed. A reconcile failure there (e.g. a transient IndexedDB error) then had no record left to retry from. Reordered: reconcile the consumed ids first, settle the drain's own receipt second — a failure now throws before the drain settles, so the drain record stays `"submitting"` and a later dispatch resends it (the server replays it, and the replay's own receipt — from last round's fix — carries the same consumed ids again).

Test: `"a reconcile failure leaves the drain dispatchable, and a retry settles the consumed ids"` — spies on `settleApplied` to fail exactly once (the only call it makes in this flow), confirms the drain record is still `submitting` and the consumed record still `optimistic` after, then lets the spy succeed and confirms a second `dispatchTargets` call resends `turn/drainAsSteer` and settles both. Falsified against the pre-reorder code first (temporarily un-reordered, confirmed red with the exact same assertion failure, then restored) — a straight `beforeCommit` hook couldn't reach `settleApplied` here since its own write is unconditionally unlabeled (`operation: undefined`) and `#write` only invokes `beforeCommit` `if (operation)`, so the spy targets the method directly instead.

**Low (fixed):** restored the deleted #1452-era regression case, adapted to the receipt-named mechanism: `"a queue intent enqueued while the drain is in flight survives and dispatches after settlement"` — a queue intent lands on the emptied queue while the drain's own request is still in flight, the drain's receipt never names it (it names only what it actually consumed), and the test confirms it survives settlement and is itself dispatched (`turn/queue` sent, settled into `optimistic`) in the same `dispatchTargets` call.

Gates: `mobile-native` had no install in this worktree (fully absent, not a stale symlink) — ran `npm ci` there directly, safe since nothing shared was at risk. `npm run check` (mobile-native typecheck, covers `../mobile/src` too) and `npm run test:shared` (773 tests) — both clean. `cmd/evener-hub/frontend`: `npx vitest run src/stores/threads.test.ts src/stores/mutationDispatcher.test.ts src/stores/mutationDispatcher.edge.test.ts` (353 passed), `npm run typecheck`, `npm run lint` (biome, clean after one line-length fix). `make lint-package-imports`.

Head: `8ae0d871c913b96cf97a95525e82021407afa62b`. `gh pr view 1735 --json mergeable,mergeStateStatus` → `MERGEABLE` / `BEHIND` (main has moved further since the last round; no conflicts, main merge not requested this round either).

obra:
## Review round 4 disposition (bfcb15287)

**Medium 1, REFUTED — "keep the previous drain-specific fallback when consumedClientMutationIds is absent":**

Not implementing this. The sequence heuristic being asked back is the #1704 bug itself, not a compatible substitute for it:

(a) The heuristic being proposed as a fallback — "retire optimistic `turn/queue` records with `intentSequence < record.intentSequence` for `turn/drainAsSteer`" — is exactly `#1452`'s `intentSequence`-comparison rule this whole series exists to retire. It only holds within one dispatcher instance's own per-target FIFO; two browser tabs share one outbox with no cross-tab lease (`mutationOutboxIndexedDB.ts`'s own header comment), so a second tab's `turn/queue` intent with a *lower* `intentSequence` can be accepted by the server *after* another tab's drain — genuinely fresh work the old daemon's drain never touched. Restoring the fallback for old daemons restores that same data-loss window for old daemons specifically, which is a strictly worse trade than doing nothing.

(b) The documented compatibility behaviour for exactly this case is already the safe one: `appwire/types.go:848-859` — "Absent on old daemons and on every push that is not the consuming transition" — and this PR's own body states the client "degrades to today's ghost row, never deletes live work" when the field is absent. A ghost row (stale-but-visible) is strictly safer than the fallback's silent retirement of another tab's live intent.

(c) Jesse's standing rule is no backward-compatibility code without his explicit approval (CLAUDE.md), and this PR was never scoped or approved to add old-daemon-specific behaviour — it adds a new field and consumes it when present, full stop.

(d) The deferred hydrate-only backstop PR (noted in rounds 1 and 3 of this series, and in the #1705 closing ruling) is what covers the case an absent push leaves uncertain: a reconnect's authoritative `thread/read` either does or doesn't name the record, and *that* absence is real positive evidence from a source with no cross-tab race, unlike a locally-computed sequence guess. Building a sequence-based patch here would preempt that PR with the exact mechanism it exists to avoid.

Citing #1704 (the bug this series fixes) and the #1705 PR's own fresh-eyes measurement thread (closed per Jesse's ruling: "this PR is client-side compensation for one absent server field... the replacement is a small PR that names the consumed ids... and later a hydrate-only backstop for the missed-push case").

**Medium 2, fixed — "field used for destructive reconciliation without runtime validation or a drain-method guard":** real. `mutationDispatcher.ts`'s `#attempt` and `threads.ts`'s `notificationMutationIdentities` both spread `consumedClientMutationIds` with no shape check: a malformed non-array value (a string) is itself iterable character-by-character, so an unguarded spread silently manufactures one bogus "id" per character, and a well-formed array riding a receipt for a mutation that isn't a drain had nothing stopping it from settling unrelated records.

Fix: `validConsumedClientMutationIds` (new, exported from `mutationDispatcher.ts`, imported by `threads.ts` — same predicate, one definition) treats anything that is not an array of non-empty strings as absent. The receipt path in `#attempt` now only consults the field when `method === "turn/drainAsSteer"` (the record's own method, already computed there); `notificationMutationIdentities` runs every `queueChanged` push's field through the same validator before spreading it.

RED-first (each falsified against the pre-fix code, confirmed red with the fix reverted, then restored):
- `mutationDispatcher.test.ts`: "a malformed consumedClientMutationIds on a drain receipt is ignored, not iterated" (a string receipt value; asserts `reconcileIdentities` is never called) and "a well-formed consumedClientMutationIds on a non-drain receipt settles nothing" (a `turn/queue` receipt carrying the field; asserts `reconcileIdentities` is never called).
- `threads.test.ts`: "a malformed consumedClientMutationIds on a queueChanged push is ignored, not iterated" — a queued record is forced to the id `"a"` (via a fixed `createMutationId`) and the push's malformed value is the string `"abc"`, chosen so an unguarded character-spread would land on `"a"` and wrongly retire it; confirmed the record survives.

**Low — deferred, not built here:** "Mobile decoder accepts empty array contrary to documented contract" (`mobile/src/services/conversation.ts:440`) is a behaviour change (tightening the decoder to reject `consumedClientMutationIds: []`), not a comment/naming item, so per this round's instructions it's deferred to a follow-up rather than folded in here.

Gates: no Go changes this round (golangci-lint not re-run) and no phone validator changes (mobile-native typecheck/services tests not re-run — the deferred Low is the only phone-adjacent item, and it's untouched). `npx vitest run src/stores/threads.test.ts src/stores/mutationDispatcher.test.ts src/stores/mutationDispatcher.edge.test.ts` (356 passed), `npm run typecheck`, `npm run lint` (biome, clean), `make lint-package-imports`.

Head: `bfcb15287ef805ab8d0e208aa08924e4f60f6c01`. `gh pr view 1735 --json mergeable,mergeStateStatus` → `MERGEABLE` / `BEHIND` (main has moved further; no conflicts, main merge still not requested).

obra:
## Review round 5 disposition (bfcb15287) — coordinator

**High, v5 drain receipts add a key already-shipped phone builds reject:** measured and real for TestFlight build 4 (source 92561dbe3 calls turn/drainAsSteer from QueueSheet.tsx, and its receipt allowlist predates the key). Not fixed here by a protocol bump: Jesse ruled 2026-09-17 that additive AppWire changes stay on the current ProtocolVersion ('drop the version bump'), and the daemon-side suppression/negotiation the review proposes is backward-compatibility code this PR is not approved to add. The class problem (the phone's strict receipt decoder turns every optional field into a break for shipped builds) is filed as #1759, with the immediate consequence that TestFlight build 5 must carry this PR's widened allowlist.

**Medium, removed web fallback:** unchanged from the round-4 refutation above (the proposed fallback is #1704's own heuristic; the documented absent-field degradation is a ghost row; the hydrate-only backstop PR covers the missed-push case).

**Low, empty array accepted:** follow-up PR, tracked in #1760.

Merging: CI green (15), /simplify applied at c8a34ea0f, round 5 of 5.

PR #1736, opened 2026-09-18
fix(sdk): extract the draft-checkpoint port from keybindingsStore

obra:
## Summary
- Third piece of the #1549 (D6-core) decomposition, per Jesse's five-rounds rule (docs/superpowers/plans/2026-09-12-sdk-migration.md D6, plan comment on #1549). Plan line 3: "Extract the draft-checkpoint port (load/save/removeIf/discardClassified; no replaceIf yet) + testing/draftStorage.ts."
- Targets `main` directly: #1728 and #1729 are merged; this branch has been refreshed onto both.
- A pure move: `removeIf`/`discardClassified` stay `void`, exactly as main's `KeybindingDraftStorage.removeIf` has them - no store call site in this slice reads a return value from either. The void -> boolean widening (and the native side it forces) is not part of this piece; it lands in #1741 (PR 5), where saveDraft/discardDraft actually consume it to adopt or re-classify.
- Adds `appwire-client/typescript/draftCheckpointPort.ts` (`DraftPort`, `DraftRepository`, `UnreadableDraftError`, `discardStoredDraft`, `createDraftRepository` with `load`/`save`/`removeIf`/`discardClassified`).
- Rewires `keybindingsStore.ts`'s local `draftRepository`/`memoryDraftStorage` to the shared port: `drafts` is now `createDraftRepository(deps.drafts ?? memoryDraftStorage(), draftCheckpoint)`; `discardDraft` now calls `drafts.discardClassified()` (removes the record by the identity it was last classified under) instead of a fresh load-then-removeIf - the same outcome for every existing case, one less round trip; `invalidDraft()` throws `UnreadableDraftError` instead of a bare `Error`.
- `testing/draftStorage.ts` (generic `memoryDraftStorage<Checkpoint>()`) replaces `testing/keybindingDraftStorage.ts`; `keybindingsStore.test.ts` and mobile-native's `keybindingRecovery.test.ts` re-point at it with no other test changes.
- Plumbing: `draftCheckpointPort.ts` added to `tsconfig.build.json`; `index.ts` re-exports `discardStoredDraft as discardStoredKeybindingDraft` from keybindingsStore.ts (which itself re-exports it from draftCheckpointPort.ts) - this is also what makes the new module reachable for the qualifier, since it isn't otherwise exported at the top level.

## Test plan
- [x] `npx vitest run keybindingsStore.test.ts` (appwire-client) - 57/57 passed, unchanged assertions
- [x] `npx vitest run keybindingRecovery.test.ts` (mobile-native) - 8/8 passed, unchanged assertions
- [x] mobile-native `npm run check` and full `npm test` (792/792) - clean
- [x] `npm run typecheck` (web) - clean
- [x] `make test-api-package` - qualified (draftCheckpointPort reachable via the discardStoredDraft re-export chain)


obra:
## Review round disposition (a82a363)

- Restacked on the fixed #1729 (per-store token scoping carried up; that High is resolved by refreshing onto it).
- High (native): `NativePreferenceDraftBackend.deleteIf` was widened by nothing on this PR's own diff but its CALLER's contract (`KeybindingDraftStorage.removeIf`) went void -> boolean, and the native backend stayed void, failing `make test-native`'s typecheck. Fixed: `deleteIf` now returns `boolean`; the production backend in `NativePreferencesProvider.tsx` reports whether it actually removed the record (compare-then-remove, same as before, now with a real return); `nativeKeybindingDrafts.removeIf` propagates it (unchanged code, now correctly typed); the three test backends (`bindNativePreferences.test.ts`, `nativePreferences.test.ts`, `preferenceDraftRepository.test.ts`) follow. `nativeTranscriptDrafts.removeIf` stays `void`, unaffected (a boolean return is assignable to a void-typed slot).
  - New `mobile-native/src/nativePreferenceDrafts.test.ts` pins the propagation contract (refused deleteIf -> false, stored record untouched; successful deleteIf -> true, record cleared). Note on "RED first": vitest doesn't type-check (esbuild transform only), so this specific bug's RED signal is `mobile-native`'s `npm run check` (tsc), confirmed failing before the fix (`TS2322: Type 'void' is not assignable to type 'boolean'` at nativePreferenceDrafts.ts:32) and passing after; the new test itself was runnable-and-green from the start since JS ignores declared return types, but it now pins the contract going forward.
  - Ran `mobile-native`'s `npm run check` (clean), full `npm test` (794/794 passed), and the touched-file subset directly.

Deferred to a follow-up (Lows, per the "Lows never cost the big PR a review round" rule):
- `UnreadableDraftError`'s doc comment describes a discard-without-storageUnavailable distinction that isn't implemented yet - `restoreDraft` currently treats it like any other storage throw. That distinction is PR 6's job (draftUnreadable/assertDiscardable).
- `draftCheckpointPort.ts`'s module comment names `transcriptDisplayStore.ts` as an existing consumer; only `keybindingsStore.ts` uses it in this slice (the comment is pr-1549's own forward-looking wording, taken byte-for-byte per the brief; will read accurately once the transcript piece lands, or gets corrected in the follow-up).

Gates: mobile-native typecheck clean, `npm test` 794/794, package/web typecheck+biome+lint clean, `make lint-package-imports` pass, `make test-api-package` qualified.

obra:
## Review round disposition (db3b131)

- Medium: `removeIf`/`discardClassified` returned a boolean no store call site in this slice consumed (`saveDraft`'s post-rename branch and `discardDraft` both called them as bare statements), so the widened contract was dead weight that forced a matching boolean shim onto mobile-native's `NativePreferenceDraftBackend` for no reason this piece needed. Reverted both to `void`, exactly as main's `KeybindingDraftStorage.removeIf` has them; reverted the native `deleteIf`/`replaceIf` widening (`NativePreferencesProvider.tsx`, `nativePreferenceDrafts.ts`) and its test fixtures, and dropped `nativePreferenceDrafts.test.ts` (it existed solely to pin the boolean-propagation contract). The widening and the native side move to #1741 (PR 5), where saveDraft/discardDraft actually consume the result to adopt or re-classify - that PR now carries both.
- Low: `draftCheckpointPort.ts:1-4` and `keybindingsStore.ts:311-314` referenced a `transcriptDisplayStore.ts` consumer and an "existing" index.ts import, neither of which exists yet in this slice. Reworded both to describe the current keybindings-only extraction and a future piece bringing a second store onto it.
- Refreshed onto main: #1728 and #1729 are merged. Every file either touched (readyGenerationFence.ts/.test.ts, README.md, keybindings.ts/.test.ts, transcriptDisplay.ts/.test.ts, readyGenerationCallback.ts/.test.ts) diffs empty against origin/main; the three files this PR shares with #1728 (keybindingsStore.ts, index.ts, tsconfig.build.json) diff to exactly this PR's own additions, verified with typecheck + the full keybindingsStore suite (not just a marker grep).

Gates: `keybindingsStore.test.ts` 57/57 (unchanged), mobile-native `npm run check` clean + `npm test` 792/792, web typecheck/biome/lint clean, `make lint-package-imports` pass, `make test-api-package` qualified.

obra:
## Round 2 fixes (8c0c4448c)

Measured first: `git diff origin/main...pr-1736 --stat` — `draftCheckpointPort.ts` is entirely new in this PR (122 insertions, 0 on `main`; confirmed with `git show origin/main:appwire-client/typescript/draftCheckpointPort.ts` → missing, and `git log --all -- .../draftCheckpointPort.ts` → first commit is this stack's p3). So every finding citing it is about code this PR introduces, not something inherited from #1729. That said, two of the Mediums still land in piece 5 (#1741), not here — explained below.

### Medium: `discardClassified()` ignores whether `removeIf` removed anything
Not fixed here. Reporting the outcome needs `removeIf`/`discardClassified` to return `boolean` instead of `void`; that widening (and `discardDraft` re-reading and adopting the replacement on refusal) is piece 5's mechanism (plan item 5: "the void->boolean widening... live here"). Verified #1741 already implements this (`discardClassified(): boolean`, `keybindingsStore.ts`'s `discardDraft` checks the result and calls `restoreDraft` on refusal). Will confirm explicitly in #1741's round.

### Medium: `discardStoredDraft` passes a raw `unknown` to `removeIf(checkpoint: Checkpoint)`
Not fixed here, for the same reason: widening this contract touches the same `removeIf` signature piece 5 is already changing, so it lands there rather than twice. Checked #1741 as of this reply — it has NOT yet fixed this one (the cast is still present); will fix it in #1741's round and note that here was wrong to assume it was pre-existing.

### Medium: new raw-identity and discard logic has no direct test coverage
Fixed. Added `draftCheckpointPort.test.ts` (6 tests): `removeIf` on a `load()`-returned checkpoint with extra fields (proves `rawFrom`), `discardClassified` after `save()`, `discardClassified` refusing when another writer replaced the classified record, `discardClassified` falling back to a fresh reload for a corrupt record nothing has classified yet, and `discardStoredDraft` on an unreadable record / on an empty store. Also added a `keybindingsStore.test.ts` test for the exported `discardStoredKeybindingDraft` helper.

### Low: `rawFrom` is unreachable in the current codebase
Measured: confirmed no production caller ever passes `removeIf` a checkpoint that came from `load()` — `settleWrite`'s checkpoint always comes from `persistDraft` (a fresh `save()`-produced object), never from `drafts.load()`. Rather than delete it, added the direct repository-level test above (`removeIf on a checkpoint load() returned removes the exact stored bytes, extra fields included`), which fails without `rawFrom` and passes with it. Kept the mechanism: `draftCheckpointPort.ts`'s own header says a later piece brings a second store onto this same repository (plan items 8-12), and a byte-aware port (native, which compares JSON strings) needs exactly this identity from any caller that loads before removing — deleting it now would mean re-adding it in a couple of pieces.

### Low: `discardStoredKeybindingDraft` is a generic alias, not bound to `KeybindingDraftStorage`
Fixed. `keybindingsStore.ts` now exports a typed wrapper (`(storage: KeybindingDraftStorage) => void`), and `index.ts` imports that directly instead of re-exporting the generic function under an alias. (Kept a plain re-export of the generic `discardStoredDraft` too, under its own name, so `draftCheckpointPort.ts` stays reachable from the package's declaration graph for the next store that uses it directly — verified with `make test-api-package`'s qualifier, which fails with "shipped module unreachable... draftCheckpointPort" without it.)

### Low: `invalidDraft()` throws `UnreadableDraftError` for both validation and decode
Fixed. Split into `invalidDraft` (decode path: `draftCheckpoint`/storage-sourced `keybindingRules` calls) and `invalidUserRules` (the plain `Error` `editDraft`/`saveDraft` throw for a caller's own malformed input) — same message, different exception type, so a bad paste is never conflated with a corrupt stored record. Added a test asserting `editDraft`'s error is `instanceof Error` but not `instanceof UnreadableDraftError`.

### Low: `stored()` casts the corrupt value to `Checkpoint`
Fixed. `MemoryDraftStorage.stored()` now returns `unknown`; all existing callers already used it with `toMatchObject`/`toBeNull`/`toEqual`, no static access, so nothing else changed.

### Low: `createDraftRepository<Checkpoint>` is unconstrained
Fixed. `Checkpoint extends object`.

## Gates
- `npx vitest run keybindingsStore.test.ts draftCheckpointPort.test.ts` (from `appwire-client/typescript`): 64 passed
- `npx biome ci ../../../appwire-client/typescript` and `npm run lint` (from cmd/evener-hub/frontend): clean
- `make test-api-package`: qualified (falsified: removing the `discardStoredDraft` re-export reproduces "shipped module unreachable... draftCheckpointPort"; restored)

## Falsification
Reverted only `keybindingsStore.ts` to this branch's prior head and re-ran `keybindingsStore.test.ts`: 3 failures as expected — `discardStoredKeybindingDraft is not a function`, and both `editDraft refuses a whitespace-only ...` tests failing `expected Error: Invalid keybinding draft. to not be an instance of UnreadableDraftError`. Restored with `git apply` of the saved diff; `git status` clean before committing.

obra:
## Round 3 fixes (8f5798bdf)

### Medium: unreadable record locks the settings section
Pointer to #1791 (p6, head `4c4a440da`). Measured: `KeybindingsStoreFields.draftUnreadable: boolean` (keybindingsStore.ts:144) and `assertDiscardable()` (keybindingsStore.ts:1219) both exist at that head - `assertDiscardable` replaces `assertEditable` as `discardDraft`'s gate and lets discard through when `storageUnavailable && draftUnreadable`, and `restoreDraft`'s catch discriminates `UnreadableDraftError` to set it. #1791 also wires the native screen's discard action and the provider's raw-bytes-on-unparsable `get()`. Not fixed in #1736 itself.

### Medium: `save()` unconditionally overwrites storage
Pointer to #1741 (p5, head `ff4604727`). Measured: `DraftPort.replaceIf(expected: Checkpoint, next: Checkpoint): boolean` and `DraftRepository.replaceClassified(next: Checkpoint): boolean` both exist at that head, and `settleWrite`'s uncertain-checkpoint-settle path adopts a replaced checkpoint via `replaceClassified` instead of `save()`. Not fixed in #1736 itself.

### Medium: `removeIf()` reports no outcome
Also #1741 (p5, head `ff4604727`). Measured: `DraftPort.removeIf`/`DraftRepository.removeIf`/`discardClassified` all return `boolean` at that head, and `discardDraft`/`settleWrite` check the result and re-classify (`restoreDraft`) on refusal instead of clearing state unconditionally. Not fixed in #1736 itself.

### Medium: `load()` clears `hasLastClassified` on empty storage
Fixed. `load()` classifying an EMPTY store is now tracked as `classifiedAbsent` (distinct from never having classified at all) - `discardClassified()` is a no-op for it instead of taking the fresh-reload fallback and removing a record an external writer saved afterward, one this repository never classified. Added `discardClassified is a no-op after load() observes empty storage, even when another writer saves afterward`, matching the review's suggested scenario.

### Low: `save()` does not update `rawFrom`
Fixed. `save()` now also does `rawFrom.set(checkpoint, decoded)` for the checkpoint it was given, so a checkpoint returned by `load()` (extra fields, different key order) and then re-saved through the SAME reference has its raw identity updated to what was just written, rather than a later `removeIf` on that reference naming pre-save bytes `save()` already overwrote. Added `removeIf after re-saving a loaded checkpoint matches the just-written bytes, not the pre-save raw identity`.

### Low: `lastRemoveIf()` is mistyped
Fixed. `MemoryDraftStorage.lastRemoveIf(): Checkpoint | null` → `unknown`, matching what the port actually receives (`discardStoredDraft` passes raw unreadable values through `removeIf`). No caller did property access on the old type, so nothing else changed.

## Gates
- `npx vitest run keybindingsStore.test.ts draftCheckpointPort.test.ts` (appwire-client/typescript): 66 passed

## Falsification
Reverted `draftCheckpointPort.ts` and `testing/draftStorage.ts` to this branch's prior head and re-ran `draftCheckpointPort.test.ts`: the two new tests fail as expected (`discardClassified` removes the external writer's record instead of leaving it; `removeIf` after re-save fails to remove the just-written record). Restored with `git apply`; `git status` clean before committing.

obra:

Issue #1656, opened 2026-09-17
A settled item/completed with no images drops the images the item already had


obra:
Found while closing #1580's round 7 (the input-image rule, #1614 parity).

The shared reducer's settle path replaces an item from the wire payload and layers only specific fields off the item it already had (`appwire-client/typescript/reducer.ts`, the `item/completed` case's `mergeItemIdentityMetadata` / `mergeCompletedText` / `mergeReasoning` / `mergeArguments` / `mergeObservedTiming` chain). `images` is not one of them, so a settle whose payload carries no `images` field — or an empty one, which now reads as "nothing said" — leaves the item with no input images at all.

Two places disagree with that:

- the hub keeps whatever list the item already had when the incoming one is empty: `server/appwire_turns.go:884-886` and `internal/apptranscript/logical_turn.go:309`, both `len(incoming.Images) == 0`;
- `mergePageItem` keeps them too, since #1614/#1580 restored `imagesToItemImagesForSession` to return `undefined` for an empty list.

So a user message whose image arrived on `item/started` can lose its attachment row when the matching `item/completed` says nothing about images, until an older page or a read puts it back. Reproduced incidentally by `mobile-native/src/liveImages.test.ts`'s "commits the snapshot's image over a live removal" case, where the intermediate state after the live frame has no attachment row.

Pre-existing in the web's reducer (this is the shared code the web has always run), so it is not a regression from the SDK migration; filing rather than widening #1580.

Fix shape: carry `images` (and `outputImages`, subject to its own explicit-empty rule) through the settle merge the way the hub does, with a reducer test per field.

PR #1732, opened 2026-09-18
fix(web): a blank warning title, hint or message renders nothing

obra:
Piece B of the #1580 decomposition plan (see the [decomposition plan comment](https://github.com/prime-radiant-inc/evener/pull/1580#issuecomment-5723104182)). **Stacked on #1731 (piece A)** — targets `main`, but only makes sense once A merges; will be refreshed onto `main` after A lands.

## Summary
- Cut byte-for-byte from #1580 (`53bb2b6ca`). Web-only polish, absorbs none of today's RoboRev findings.
- `WarningItem` now treats a whitespace-only title/hint/message as absent, matching the phone's projector (`mobile/src/conversation/project.ts`'s warning branch already joins parts filtered on trimmed content).
- `reasoningFormat.ts` drops its own `joinedReasoningParagraphs` in favor of the package's (piece A), which `ThinkBlock.tsx` already reads from here.
- Depends on piece A's `reducer.ts` change (the `warningMessage(params) || JSON.stringify(params)` fallback is gone there, so `item.text` can now legitimately be empty and reach this component) and piece A's `joinedReasoningParagraphs` export.

## Test plan
- `npx vitest run src/panes/session/transcript/messages/WarningItem.test.tsx src/panes/session/transcript/messages/reasoningFormat.test.ts` (from `cmd/evener-hub/frontend`, its own jsdom config) — 45/45 pass
- `npm run typecheck` (`cmd/evener-hub/frontend`) — clean
- `npm run lint` / `npx biome ci ../../../appwire-client/typescript` — clean, 1354 files
- `make lint-package-imports` — PASS

PR #1737, opened 2026-09-18
refactor(native): rows are a projection of the model (D23c-2b, piece C)

obra:
Piece C of the #1580 decomposition plan (see the [decomposition plan comment](https://github.com/prime-radiant-inc/evener/pull/1580#issuecomment-5723104182)). **Stacked on #1731 (piece A)** — targets `main`, will be refreshed onto `main` after A lands.

## Summary
- Cut from #1580 (`53bb2b6ca`): extracts `project.ts`'s pure projection library — `askQuestionsByCall`/`liveAsksFor` (ask-dock support), `boundQuestion`/`truncateItem`/`MAX_ITEM_BYTES`/`truncateText` (display bounds), and the `timelineIdentity`/`attachmentSourceIdentity`/`ownTimelineIdentities` family (row identity) that piece D's page-history layer needs.
- **Additive only, by measurement**: `main`'s `state/conversation.ts` currently imports only `activityState`, `clusterActivities`, `isActiveItem`, `projectItemAttachments` from `project.ts` — none of the new exports above. This branch keeps all four of those exports with their **old names, signatures and bodies exactly as `main` has them** (renamed the *new* projection code's internal equivalents to `activityStateInModel`/`isActiveItemInModel`/`clusterActivityRun` to free the names, rather than deleting the old surface — nothing new was added for compatibility, the existing surface just isn't removed yet). `main`'s `conversation.ts` keeps compiling unchanged; deleting the old four and cutting `conversation.ts` over is piece D's job, not this one — the page-history layer D needs makes that cutover a single atomic mechanism (see the plan comment for why).
- Fixes this round's Low (`turnRowCache`, project.ts's per-turn row cache): measured directly against `projectTimeline` with a shared turn object across two calls — the existing `askState` per-call validation (recorded on each cache entry: every ask_user call the turn posted and whether it was answerable when cached) already invalidates correctly when a turn's own ask resolves, and correctly reuses when an unrelated `asks` change doesn't touch this turn's calls. Added two tests locking in both directions rather than changing behavior that measured out to already be correct.

## Test plan
- `npx vitest run --config mobile-native/vitest.config.mts mobile/src/conversation/project.test.ts mobile/src/state/conversation.test.ts mobile-native/src/questionAnswers.test.ts mobile-native/src/questionBatches.test.ts mobile-native/src/draftRepository.test.ts mobile-native/src/liveImages.test.ts` — 491/491 pass


obra:
## Simplify disposition (10beda5ce)

Restacked onto the new A (#1731) head, `93edcf7f2`, then two rounds of work:

**Simplify pass (no behavior change):**
1. `activityStateInModel` re-encoded the kept export `activityState` line for line; callers now call `activityState(item, turn.status)` directly and the private copy (plus its orphaned duplicate comment) is gone.
2. Dropped decomposition-plan piece-lettering from two comments narrating a future cutover, keeping the "this gets deleted then" pointer.
3. Merged two consecutive one-line section banners into one.
4. `truncateText`'s post-slice trim loop never ran (the prefix loop already stops at `targetBytes`, measured 0/30000 hits); it's now a plain slice + marker, and the encoder now serves only `markerBytes`.
5. `capItems` skips building the 500-row identity set on a publish where the cut didn't land on an attachment at all (the common case).

**Fix — a real bug, found while restacking D's own round-4 fold-in tests against this file:** `liveAsksFor`'s memo (project.ts) keyed on `model.turns` alone. A status-only frame that flips `ThreadModel.askPending` (conversation.ts's `changesRows`, #1731 round 4's own consequence) hands `liveAsksFor` the SAME turns reference with a different `askPending`. Keyed on turns alone, the memo served the answer it cached under the OLD flag — an already-projected tool row stuck instead of becoming its question row, or the reverse. The cache key is now `{turns, askPending}`. TDD RED first: a dedicated unit test on `liveAsksFor` with a shared `turns` reference and only `askPending` flipped, confirmed failing before the fix.

**#1731 (piece A) round 4 fold-in:** ThreadModel.askPending is now the single source for "is anything pending". Three of this file's own tests asserted the old reading (a found, answerable ask_user call is live regardless of the wire flag) and now fail against `liveAskQuestions` gating on it. Fixed the fixtures and rewrote the one test that asserted the old contract by name ("counts a question row on screen as a pending question, without touching the wire flag") to assert the new one in both directions: `askPending: true` renders the question row, the identical transcript with `askPending: false` renders none. `mobile-native/src/questionAnswers.ts`'s `pendingQuestions` already gated on `conversation.askPending` pre-#1580, so it agreed with no change needed.

**Gates:** `npx vitest run --config mobile-native/vitest.config.mts mobile/src/conversation/project.test.ts` — 122/122 pass. `mobile-native`: `npm run check` (typecheck) — clean; `npm test` — 802/802 pass; `npm run test:shared` — 785/785 pass. `make lint-package-imports` — PASS.

Head is `10beda5ce`. E and D are restacked onto it (both are siblings on C, not on each other).

obra:
## Round 2 findings routed (A's server/client mechanism moved out from under this PR)

#1731 (piece A) is being decomposed after round 4; this review's two Highs and the first Medium are all A's, not C's:

- High: "Carrier skill-selection failure lets restore resurrect a cleared ask" → fixed in #1806 ("piece A-server: the daemon owns the ask boundary") via a short-lived carrier-claim marker so `recordFailedSteeringSelection` tags the `TurnFailure` `SteeringCarrier` only for the carrier's own claimed steer. New test `TestAskUser_RestoreResolvesAcrossFailedSteeringSelectionCarrier`, falsified before landing.
- High: "askedThisRound capture before steering drain creates live/restore mismatch and undispatched steer" → also fixed in #1806, by NOT carrying round 4's move of that capture (it stays in its original position, after `injectPostToolSteering`'s drain), so a resolving mid-round steer's clear is read before `deliverIfCommunicated` decides the boundary, and the round genuinely continues to a second model request. `TestAskUser_InjectPostToolSteeringClearsPendingAskForAnAcceptedUserSteer` now scripts and asserts that second request.
- Medium: "askPending gate breaks package qualification and hides questions for old daemons" → the qualification half is fixed in #1731's force-pushed, now-TypeScript-only diff (04584a21c): the three qualifier smoke calls now pass `askPending: true`, matching every other fixed fixture. The "old-daemon regression" half (absence means false, hiding a question a pre-upgrade daemon still has pending) is unchanged from this review's own account — flagging it here since it's a real compatibility gap the fix above does not address; happy to file it as its own issue if it's not already tracked.

C's own findings (native truncation/cap helper wiring, `boundQuestion`/`pendingQuestions` bounding) are being picked up separately as C restacks onto the trimmed A.

obra:
## Restacked onto trimmed A/B (57476d548), plus C's own fixes

Restacked onto #1732's new head (ba45d3936, itself restacked onto #1731's trimmed A). Own diff is isolated (4 files vs B's head): project.ts/project.test.ts (unchanged from before) plus mobile/src/state/conversation.ts/test.ts (new).

- Medium "Native truncation/cap helpers not wired into the production store" (project.ts:942, conversation.ts:481): conversation.ts's own truncateItem only bounded "assistant" and "activity" rows — a pasted user message, a daemon notice, a tool failure's stack, and a question's own text were never bounded by the live path. Its own capItems was a plain slice with no orphan-attachment awareness. Both now delegate to project.ts's shared truncateItem/capItems. New tests (oversized user/notice/failure item text) — falsified by reverting the delegation first (TypeError: truncateItem is not a function, since the test imports the now-exported wrapper), restored and green. Full existing conversation.test.ts suite (344 tests) and the whole mobile/src suite (791 tests) still pass unchanged — the delegation is behavior-preserving for every case the old private implementation already covered.
- Note: this same finding was raised on E's review (#1738) since E sits downstream of C. This commit (57476d548) is the fix; no separate fix needed in E. I'll note it there too.
- Low "boundQuestion comment contradicts code; truncated labels could pass client validation" (project.ts:1020-1027, mobile-native/src/questionAnswers.ts:11-17): this is E's file (mobile-native/src/questionAnswers.ts owns pendingQuestions), not C's — picking it up on E's restack, since wiring capItems/truncateItem above is what activates the latent bug (conversation.ts's OLD truncateItem never touched "question" rows at all, so this was previously unreachable in production).

Gate results:
- \`npm run check\` (mobile-native, includes mobile/src via test:shared paths): clean
- \`npx vitest run\` (mobile-native's test:shared config over mobile/src): 791 passing
- \`npx vitest run\` (mobile-native's own suite): 795 passing

obra:
## Restacked onto main after A (d4f2562)

Rebased `claude/c2b-c-native-rows-are-a-projection` from its old base (`e692ca1e7`) onto current `main` (tip `0a53ef91f`, which now carries #1859/#1892/#1893/#1894/#1895). The branch's bottom commits (`aa9f96408`..`ccdd9c96e`/`ba45d3936`, the ask-resolution + web warning-guard work) were byte-equivalent to what's already squash-merged on main as #1892/#1893/#1894/#1895, so the rebase base was moved up to `ba45d3936` (C's last redundant commit) — only the four commits that are actually piece C's own (`8cd55f9e7` "rows are a projection of the model", `eead00500`, `94b09b288`, `57476d548`) were replayed. This matches the original stacking: `git log --first-parent` showed a no-op merge commit (`78ccc026c`, parents `04584a21c`+`e692ca1e7`, no diff of its own) that had recorded "origin/main into piece A" at the time C's branch forked from what later became A.

### Conflict resolution (hunk by hunk, `mobile/src/conversation/project.ts` / `.test.ts`)

Only the first replayed commit (`8cd55f9e7`) conflicted; the other three applied cleanly.

1. **Import block** — kept both sides: `isActiveItem as isActiveItemInModel` (C's rename, needed by the new model-typed pure helpers) and `joinWarningParts` (main's #1895 mechanism, needed by the warning branch below).
2. **Duplicate `isActiveItem`/`activityState` doc comment** — cosmetic only; the full exported functions already auto-merged intact further down ("kept for the store's live path" section), so I took C's shorter comment and dropped the stale duplicate.
3. **Reasoning state/output** — main had landed a length-comparison heuristic (comment: "Two different fields can be stale, depending on whether the item is still running...") that C's branch predates. Kept main's mechanism verbatim but re-expressed it as C's own `reasoningText(item, running)` helper instead of inlining it, so C's goal (a reusable pure function) holds without regressing main's fix. Confirmed both of C's own reasoning tests still pass unmodified under this heuristic (neither exercises the settled+both-present case where the two behaviors would actually diverge).
4. **The new `item.type === "warning"` branch** (C's own addition — main has no dedicated warning row yet, warnings fell through to the generic "unknown activity" fallback) — replaced C's ad-hoc `[title, text, hint].filter(part => part.trim() !== "").join("\n")` with `joinWarningParts([item.warning?.title, item.text, item.warning?.hint])`. This is exactly the r3 High finding (below): C's filter threw on a non-string wire value; `joinWarningParts`/`hasWarningText` is a proper `unknown`-typed guard, already reviewed and landed via #1895. Separator changes from `"\n"` to `" — "` to match; updated the one test whose expected string encoded the old separator, and three older tests (from #1895, asserting the pre-C generic-fallback shape `kind: "activity"`/`detail.output`) now assert the correct `kind: "notice"`/`family: "warning"`/`text` shape, since C's branch intercepts warnings before they ever reach that fallback.
5. **The generic fallback's `state`/`detail.output`** — took C's side (reuse the already-computed `state` var, drop the now-dead `warningFallbackText(item) || ...`) since no item reaching this fallback can have `type === "warning"` any more (caught by hunk 4). Deleted `warningFallbackText` entirely (private, single call site, now unreachable).
6. **`project.test.ts` imports** — union of both sides' new imports (`AnyNotification` + `AskQuestionRef`; `hydrateThread`/`applyNotification` + the new `project.ts` exports C's own tests need).

### r3 disposition (`.superpowers/sdd/.../reviews/1737-r3.md`)

- **High** (`.trim()` on untyped title/hint, all 3 reviewers) — **fixed** in the rebase itself (hunk 4 above): `joinWarningParts`/`hasWarningText` type-guards against non-string wire values instead of calling `.trim()` unconditionally.
- **Medium 1** (question refs truncated before storage; canonical vs bounded refs) — **out of scope for C.** `boundQuestion`/`truncateItem`'s question-bounding is C's own pure extraction, but nothing in C's diff calls it for answer composition — that's `mobile-native/src/questionAnswers.ts`'s `pendingQuestions`, a file C never touches. Piece E (`c2b-e-question-sheet-bounded-identity`) owns that file and is titled for exactly this; deferring to E's own review.
- **Medium 2** (live path drops existing attachments on an image-less completion) — **already fixed**, `fc64c7575` "fix(sdk): image folding + warning text (#1731 piece 2/3) (#1859)": `state/conversation.ts`'s `item/completed` handler now reads attachments from `findFoldedItem`/`itemAttachments` (comment cites "closes #1656"), confirmed present at C's rebased head, untouched by any of C's own commits.
- **Medium 3** (truncation/cap constants duplicated across `project.ts`/`state/conversation.ts`/`services/conversation.ts`) — **fixed**, new commit `f399f03a9`: `state/conversation.ts` now imports+re-exports `MAX_ITEM_BYTES`/`TRUNCATION_MARKER`/`RETAINED_ITEM_CAP`/`truncateText` from `project.ts` instead of keeping a byte-identical copy (kept only its own `exceedsByteLimit`, which has no project.ts equivalent); `services/conversation.ts`'s unused `RETAINED_ITEM_CAP` copy deleted outright (confirmed zero external importers).
- **Low** (vacuous unparseable-ask test, no `askPending: true`) — **fixed**, new commit `d4f256200`: added `{ evener: evenerThread({ askPending: true }) }` matching every neighboring ask test.
- **Acknowledged clean** items (askPending gate, `isResolutionItem`/`lastResolutionIndex`, `liveAsksFor`/`turnRowCache` memo invalidation, `mapTurn`/`settleFirstMatchingTurn`, `capItems` orphan-attachment logic) — untouched by the rebase, no action needed.

### Gates (targeted, from `.claude/worktrees/sdk-d21-native-model`)

- `npx vitest run --root .. --config mobile-native/vitest.config.mts mobile/src/conversation/project.test.ts mobile/src/state/conversation.test.ts` — 481/481 passed.
- `npx vitest run --root .. --config mobile-native/vitest.config.mts mobile/src` (full shared suite) — 810/810 passed.
- No `appwire-client/typescript`, `cmd/evener-hub/frontend`, or Go files touched — biome/typecheck-web/Go gates not applicable.

```
 mobile/src/conversation/project.test.ts | 360 +++++++++++++++--
 mobile/src/conversation/project.ts      | 678 ++++++++++++++++++++++++--------
 mobile/src/services/conversation.ts     |   7 +-
 mobile/src/state/conversation.test.ts   |  32 ++
 mobile/src/state/conversation.ts        | 131 ++----
 5 files changed, 918 insertions(+), 290 deletions(-)
```

Head: `d4f256200de2ddce2f6d6d9d4679a136e3e34a0b`

obra:
## Known finding on this piece — resolved by the stack's next PR

Local review (round 5, job 23686, `mobile/src/conversation/project.ts` pendingQuestions): on this piece standalone, question truncation can alter submitted answers because `pendingQuestions()` reads `conversation.items` (bounded rows) rather than canonical refs — contrary to `boundQuestion`'s comment claiming canonical refs.

Verified real on this piece's tree, and verified **fixed by piece E of this very decomposition** (PR #1738, `claude/c2b-e-question-sheet-bounded-identity`): its final commit rewires `pendingQuestions` to `[...liveAsksFor(conversation).values()].flat()` (canonical refs) and pins the corrected behavior with a no-production-change test explicitly citing the pendingQuestions bounding finding on #1737/#1738.

On this piece standalone the failure mode is pathological-only (question labels exceeding the 64KiB truncation bound) and fails loudly — duplicate-label rejection — never silently. #1738 is the immediate next PR in the sequence; the C→E landing order keeps the window one PR wide.

obra:
RoboRev High on `79eafff` (question truncation feeding answer composition) — **confirmed and fixed** in 1e1548d2a9. `pendingQuestions` now derives answerable questions from canonical model refs through `liveAsksFor` — the same shared scan the projection's question rows come from — instead of the truncated display rows; the store's bounding stays intact for the timeline display rows. Both consequences are pinned by failing-first tests:

- An oversized option round-trips its **original** label through `composeQuestionAnswers` (before: the truncated remnant submitted, a label the agent never offered).
- Two options whose labels share a >64KiB prefix stay distinct and both submit (before: identical truncation made the sheet unsubmittable).

The stale `boundQuestion` comment (false canonical-refs answer-path and sheet-share claims) is rewritten to match reality: only display rows read the bounded copies.

The convergence review also surfaced one Medium in the branch's paging signal, fixed in 6e47ee94ca: `loadOlder` used the final merged count as its at-cap signal, so a 501-row merge that drops an orphan could end at 499 and keep paging enabled after discarding its whole page, while an exactly-500 untrimmed merge prematurely nulled the cursor. The signal is now the pre-cap overflow (`mergedInput.length > RETAINED_ITEM_CAP`), matching the cap's own discard semantics — pinned both directions with failing-first tests.

Final local roborev on the branch delta: **Low-only**. The remaining Low (truncated labels with a shared >64KiB prefix colliding as React keys in the timeline renderer) is display-side and lands with piece E (#1738), whose questionsIdentity builds the collision-resistant identity for exactly this case.

obra:
Follow-up Medium on `6e47ee9` (question timeline rows not reconciled after `askPending` changes) — **confirmed and fixed** in 725d74fbbe. A `thread/status/changed` frame folds `askPending` straight into the model and matches no row applier, so the notification default arm published only the model half: the sheet (model-driven since the High fix) moved while the question rows kept their pre-flip shape — and since question rows come only from the canonical projection (F6 never single-item projects ask_user), the store's coalesced rehydrate is the only in-scope reconciliation. The default arm now requests that same rehydrate when the folded `askPending` differs from the published one; it cannot storm because `askPending` rides every status frame but only moves when an ask raises or resolves, and the drain scheduler coalesces bursts.

Pinned red-first in both directions with exactly one resync read each: a resolve empties the sheet and the question row leaves with it; a raise fills the sheet and the row appears matching the sheet's canonical refs.

Final local roborev on the branch delta: clean — "No issues found."

obra:
Follow-up Medium on `725d74f` (compatibility `open()` leaves stale question rows — no activity sink bound, so the rehydrate-based reconciliation is a no-op there) — **confirmed and fixed** in 881b9913e3. The no-sink path now reconciles directly: when a frame moves `askPending` and no sink is bound, the store re-renders the contiguous row window around each moved ask from the folded model's canonical projection (F6 — question rows come only from it; clusters split and merge exactly as the projection builds them), while rows outside those windows pass through verbatim with their truncation freezes carried, and live-owned rows that exist only in `items` survive. The result still goes through the same treatment `open()` installs plus the ownership prune a live merge does, and a frame that does not move `askPending` touches nothing.

The convergence review then found and fixed four more real defects in that reconciliation, each red-verified first: a whole-array reprojection silently dropped live-owned rows (now windowed); cluster membership of settled/raised asks; the truncation freeze on verbatim rows (carried, so a later delta cannot append after the marker); and identity forms the alias/live-boundary check must not swallow. Ten failing-first tests pin the unit; final local roborev on the delta: clean — "No issues found."

PR #1738, opened 2026-09-18
fix(native): a collision-resistant, memoized identity for a question set

obra:
Piece E of the #1580 decomposition plan (see the [decomposition plan comment](https://github.com/prime-radiant-inc/evener/pull/1580#issuecomment-5723104182)). **Stacked on #1737 (piece C)** — targets `main`, will be refreshed onto `main` after C lands.

## Summary
- Cut from #1580 (`53bb2b6ca`, which already includes the original round-20 `questionsIdentity` fix): `questionAnswers.ts`, `QuestionSheet.tsx`, `screens.tsx`.
- Fixes both of today's new Mediums, found by RoboRev reviewing that round-20 fix itself:
  1. **Collision**: `questionsIdentity` hashed a *bounded* (truncated) copy of each question's canonical fields, so two questions sharing a prefix past the display bound (`MAX_ITEM_BYTES`) collided on the same identity.
  2. **O(payload) per render**: every render/keystroke rehashed the whole payload from scratch, even when the question-array reference (the one `reconcileBatches.ts` hands back unchanged when nothing changed) had not moved.
- Fix: hash the canonical fields directly with `cyrb53` (a fixed-size, collision-resistant, non-cryptographic hash — two 32-bit lanes mixed in one pass, no dependency) instead of truncating first, and memoize by question-array reference in a `WeakMap` so an unaffected render never rehashes at all.

## Test plan
- `npx vitest run --config mobile-native/vitest.config.mts mobile-native/src/questionAnswers.test.ts mobile-native/src/questionBatches.test.ts mobile-native/src/draftRepository.test.ts mobile/src/state/conversation.test.ts mobile/src/conversation/project.test.ts` — 494/494 pass. Two new tests measure the fix directly: a collision test (two oversized questions sharing a `MAX_ITEM_BYTES`-scale prefix now get distinct identities — RED against the pre-fix code) and a memoization test (200 repeated calls with the same array reference stay within 10x a single fresh computation — RED against the pre-fix code, which took >2s for the equivalent case before this test's numbers were dialed down for CI).


obra:
## Seam note: one test relocated in from piece D

While cutting piece D (#1740), `conversation.test.ts`'s "shares one `askQuestionsByCall` scan between `projectConversation` and `pendingQuestions`" test turned out to call `pendingQuestions` (this file) through `conversation.ts`'s full store — a real cross-piece coupling. Measured: the property under test is entirely `liveAsksFor`'s own memoization (`project.ts`'s `asksByTurns`, `WeakMap` keyed on `model.turns`, piece C) plus this file's `pendingQuestions` calling it; it needs none of D's cutover logic. Relocated it here, rewritten directly against `projectConversation`/`liveAsksFor`/`hydrateThread` (this file's own existing wire-level pattern) instead of D's store scaffolding, so D and E's own diffs still share no file.

Head is now `80f03de36a0f99df686a4c39333f2307a9031cf6` (`+428/-23`, 4 files — added the relocated test to `questionAnswers.test.ts`). Gates re-run: `npx vitest run --config mobile-native/vitest.config.mts mobile-native/src/questionAnswers.test.ts mobile-native/src/questionBatches.test.ts mobile-native/src/draftRepository.test.ts mobile/src/state/conversation.test.ts mobile/src/conversation/project.test.ts` — 495/495 pass; `npm run check` clean.

obra:
## Review round disposition (90477a085)

Restacked onto the new C (#1737) head, `10beda5ce`, then fixed the High RoboRev found on E at `80f03de36`.

**The High:** `questionsIdentity` started hashing (cyrb53) instead of bounding-and-stringifying a question set, but `draftRepository.ts`'s `questionDefinitions` still `JSON.parse`s that signature expecting an array of `{key, ...}` it can index per key. A bare `"key:hash|key:hash"` string throws `SyntaxError` on every parse, so `writeQuestions`/`readQuestions` failed on every call — question drafts never persisted, and every edit showed "These answers could not be saved".

**Fix (the reviewer's own shape, preferred over a storage migration):** the signature is a JSON array of `{key, digest}` — `digest` is the cyrb53 hash of the canonical question, so it stays collision-resistant and bounded regardless of prose size. `questionDefinitions`' per-key compare and the stored row stay valid with no migration. TDD RED first: verified the old shape (`` `${key}:${hash}` ``) throws on `JSON.parse` before writing the fix, then added a round-trip test through the real `DraftRepository` using `questionsIdentity` itself (not a hand-built signature literal, unlike the other cases in that file) — write, close, reopen, read back, and a genuinely different question set gets a different identity and no inherited answer.

**Also fixed this round (simplify), all queued from the same review:**
- `questionBatches.ts`'s `begin()` compared `JSON.stringify(questions)` on both sides instead of the shared `questionsIdentity`, duplicating the bound/hash rule `questionsIdentity` exists to centralize.
- `QuestionSheet.tsx` guarded a display line on the canonical field (`question.why`, `option.detail`) while rendering the bounded one; guards now read the same bounded copy they render. The per-option map's inner `display` shadowed the outer question `display` it was destructured from; renamed to `optionDisplay`.
- `screens.tsx`'s `AskBatch` type import had drifted out of the shared `@evener/appwire-client` import block into its own statement; folded back in.

**#1731 (piece A) round 4 fold-in:** two `questionAnswers.test.ts` fixtures asserted a live question row with `askPending` left at its default `false`, which `liveAskQuestions` (through `pendingQuestions`/`liveAsksFor`) now correctly suppresses; set `askPending: true` where a question is meant to render live. The third fixture asserting resolution (`askPending: false` after an interrupt/steer) was already correct and untouched.

The "shares one askQuestionsByCall scan between projectConversation and pendingQuestions" test the D-side review asked me to confirm: it is here, in this file, exactly where piece D's own commit message says it was relocated to.

**Gates:** `npx vitest run --config mobile-native/vitest.config.mts appwire-client/typescript/reducer.test.ts appwire-client/typescript/deriveAskQuestions.test.ts appwire-client/typescript/transcriptProjector.test.ts` — 273/273 pass. `mobile-native`: `npm run check` — clean; `npm test` — 802/802 pass; `npm run test:shared` — 785/785 pass; `npm run check:scripts` — clean. `make lint-package-imports` — PASS.

Head is `90477a085`.

obra:
## Round 2 server findings routed to piece A-server (#1806)

#1731 round 4 grew the server-side ask-boundary change to ~430 non-test lines and broke every client consumer; it's being decomposed. The three server findings this review reports were introduced in that same round-4 diff (inherited here through A) and are fixed in the new PR:

- High: "Steering-carrier skill-selection failure resurrects a resolved ask on restore" — fixed via a short-lived carrier-claim marker (`setSteeringCarrierClaimDrain`/`steeringSelectionFailureIsCarrierClaim`) so `recordFailedSteeringSelection` tags `Error.SteeringCarrier` only for the carrier's own claimed steer. New test `TestAskUser_RestoreResolvesAcrossFailedSteeringSelectionCarrier`, falsified (hardcoding the tag to `false` fails the test) before landing the fix.
- Medium: "Live and restore derive different SessionState after a mid-round steer resolves an ask posted in that round" — root cause was `askedThisRound`'s capture moving to before `injectPostToolSteering`'s drain; the new PR never carries that move, so the capture stays in its original position (after the drain) and live/restore agree. `TestAskUser_InjectPostToolSteeringClearsPendingAskForAnAcceptedUserSteer` now asserts the drained steer reaches a real second model request.
- Low: "`turnResolvesAskBoundary`'s decisive return can never be true with resolves == false" — simplified to a single `bool`; both callers updated.

CI red here is that same inherited round-4 breakage (the client-side `askPending` gate/qualifier/mobile projection issues), not new to this PR. Once #1806 (A-server) merges and A itself is force-pushed down to its TypeScript-only diff, this branch (E) restacks onto the new A and inherits none of the above.

The remaining mobile/client findings (live-warning re-projection identity, stale question rows after askPending clears, shared truncation/cap wiring, the draft-signature migration question, the wall-clock memoization test) are being routed to the pieces that own that code (B/C/D/E) as part of the restack.

obra:
## Restacked onto trimmed A/B/C (36cfd4818), plus E's own fixes

Restacked onto #1737's new head (57476d548, itself restacked onto B onto A's trimmed TypeScript-only diff). Own diff is isolated (6 files vs C's head).

Findings addressed here:
- Low "Memoization test asserts on wall-clock time" (mobile-native/src/questionAnswers.test.ts:400-423) — measured: this is E's own file, not C's (C's diff never touches mobile-native/). Rewrote to count JSON.stringify calls instead of timing a delta: one fresh computation really does stringify (>0 calls), 200 memoized reads of the same array reference never touch it. Falsified by disabling the memo first (4200 calls instead of 0 on the repeat loop), restored and green.
- Low "pendingQuestions bounding" / "boundQuestion comment contradicts code" (C's review, #1737) — measured against the current code (not re-derived by reading): pendingQuestions already reads liveAsksFor's canonical refs, never a display-bound copy (project.ts's askQuestionsByCall scans the wire ThreadModel directly, before any store applies truncateItem/capItems). This was correctly routed to E as the owning file, but the finding doesn't apply to the current code — it's already safe. Added a regression test locking in the correct behavior (an oversized option label with a truncation-colliding twin survives with its full, distinguishing text through pendingQuestions) rather than a production fix, since there was nothing to fix.
- Low "Persisted question-draft signature changed shape with no migration" (questionAnswers.ts:78, draftRepository.ts) — real, confirmed by reading: questionsIdentity's new {key,digest} array shape doesn't match old persisted rows' full-object shape, so readQuestions returns {} across this one upgrade, discarding an in-progress draft selection (transcript unaffected). No migration added — no backward compatibility without explicit approval, and this is the internal beta. Filed #1842 for Jesse to decide.

Gate results:
- \`npx vitest run src/questionAnswers.test.ts\`: 16 passing
- \`npm run test\` (mobile-native, full suite): 806 passing
- \`npm run check\` (mobile-native): clean

obra:
## Restacked onto main after A (ffb4cba)

Rebased `claude/c2b-e-question-sheet-bounded-identity` onto piece C's new tip (`d4f256200`, itself just rebased onto main — see #1737). E's own commits (`35eb83367`, `3fdcb8cbc`, `36cfd4818`) stack directly on C's old tip (`57476d548`), confirmed by `git log --first-parent`, so this was `git rebase --onto d4f2562 57476d548 claude/c2b-e-question-sheet-bounded-identity`. All three commits applied with **zero conflicts** — E's own diff is entirely `mobile-native/src/{QuestionSheet.tsx,questionAnswers.ts,questionAnswers.test.ts,questionBatches.ts,screens.tsx,draftRepository.test.ts}`, which none of A's landed pieces or C's rebase touched.

`git diff --stat d4f256200..HEAD` (E's own diff on top of C):
```
 mobile-native/src/QuestionSheet.tsx       |  55 +++--
 mobile-native/src/draftRepository.test.ts |  38 +++
 mobile-native/src/questionAnswers.test.ts | 388 +++++++++++++++++++++++++++++-
 mobile-native/src/questionAnswers.ts      |  83 ++++++-
 mobile-native/src/questionBatches.ts      |   5 +-
 mobile-native/src/screens.tsx             |  12 +-
 6 files changed, 554 insertions(+), 27 deletions(-)
```

### r3 disposition (`.superpowers/sdd/2026-09-12-mobile-landing-queue-cont/reviews/1738-r3.md`)

- **High** (warning crash, untyped title/hint `.trim()`) — **fixed**, inherited from #1737's rebase: `project.ts`'s warning branch now uses `joinWarningParts`/`hasWarningText` (main's #1895 mechanism), not an unguarded `.trim()`.
- **Medium** (question signature incompatible with persisted drafts, `QuestionSheet.tsx:53`/`questionAnswers.ts:78-86`) — **already fixed**, by E's own later round: `questionsIdentity` now returns a JSON array of `{key, digest}` (cyrb53 hash of the canonical fields, not a truncated-and-stringified display copy), which `draftRepository.ts`'s `questionDefinitions` can parse as an array again. The current code's own comment cites this as "#1731 piece E round 4's own High" — this r3 panel predates that round.
- **Medium** (stale `items` projection on `askPending`-only status change, `conversation.ts`) — **confirmed still open, out of scope for E.** The store's `applyNotification` default branch (`state/conversation.ts`, "Everything without a row applier above is the reducer's alone") folds the model (updating `askPending`) but only calls `publishModel()` — it never re-derives `items` via `projectConversation`, so a status-only frame can leave a stale question/tool row until the next reread. This is the exact staleness the dual-write incremental applier design has throughout, and it's what D23c-2b's D piece (the full cutover to `projectConversation(model)` per notification) is built to close, per C's own commit message ("the cutover that reads every row from projectConversation and deletes the row appliers... is D's cutover, not this one"). Flagging for D's own rebase/review rather than fixing here or in C.
- **Medium** (live reasoning deltas ignore `summaryIndex`, `conversation.ts:2662-2696` current) — **confirmed still open, same reason as above.** `item/reasoning/summaryTextDelta`'s live handler flat-concatenates `detail.output + params.delta`, never routing through `reasoningSummaries`/`joinedReasoningParagraphs`. Same disposition: D's cutover retires this handler entirely in favor of re-projecting through `project.ts`'s `reasoningText`. Flagging for D.
- **Medium** (`liveAskQuestions` collapses `undefined` to empty, `deriveAskQuestions.ts:110`) — **refuted.** `ThreadModel.askPending` is a required `boolean` (`model.ts:242`, never optional), and `hydrateThread` always sets it (`thread.evener.askPending ?? false`). No real caller can hand `liveAskQuestions` a model with `askPending: undefined`; only an unsafe test-fixture cast could, and the existing suite already requires every fixture to supply the flag (per the finding's own description of the diff that added it).
- **Medium** (`turnRowCache` serves stale rows on in-place item mutation) — **refuted, per measurement already recorded in piece C's own commit (`8cd55f9e7`).** The reducer's `mapTurn`/`settleFirstMatchingTurn` never mutate a `TurnModel` in place — every fold that changes an item's text/status/reasoningSummaries/warning/error returns a turn with a new object reference (confirmed by reading `reducer.ts:775-829`), so the WeakMap-keyed-on-turn-reference cache is invalidated by that reference change alone. `askState` is the one thing the cache checks explicitly BECAUSE answerability is a whole-model property (a later turn's user message can resolve an earlier turn's ask) that does NOT change the earlier turn's own reference. C's own two tests ("does not reuse a cached question row once the turn's own ask becomes resolved" / "reuses the cached row set when the turn's own asks are unaffected") lock in exactly this distinction.
- **Medium** (live warning path keeps legacy failure row) — **already fixed** via #1895: `conversation.ts`'s `case "warning"` now calls `foldWarningParams`/`joinWarningParts` (the canonical, bounded composition), not an ad-hoc legacy row.
- **Low** (`rawWarningFrame` allocates before truncating) — **already fixed** via #1893: `rawWarningFrame` now calls `prunedForStringify` (the bounded lazy walk) before `JSON.stringify`, with `boundedCodePoints` as defense-in-depth after.
- **Low** (`liveAsksFor` leaks a mutable `Map`) — **confirmed still open.** `askQuestionsByCall`'s `Map` is returned/cached directly as `ReadonlyMap`; no current caller mutates it, so it's not an active bug, but nothing stops a future one. Per the "Lows never cost a review round" rule, not fixing inline — reporting as a residual.
- **Low** (duplicate identity/truncation helpers, `conversation.ts:56-98` vs `project.ts`) — **confirmed still open** for the identity family specifically (`attachmentSourceId`/`attachmentSourceIdentity`/`timelineIdentity`/`activityIdentity`/`ownTimelineIdentities`/`timelineIdentities` — the truncation half of this same finding was fixed on #1737). Filed as **prime-radiant-inc/evener#1925** rather than fixed inline, per the Lows rule.

### Gates (targeted, from `.claude/worktrees/sdk-d21-native-model`)

- `npx vitest run --root .. --config mobile-native/vitest.config.mts mobile/src/conversation/project.test.ts mobile/src/state/conversation.test.ts mobile-native/src/questionAnswers.test.ts mobile-native/src/QuestionSheet.test.tsx mobile-native/src/draftRepository.test.ts mobile-native/src/questionBatches.test.ts mobile-native/src/screens.test.tsx` — 530/530 passed.
- No `appwire-client/typescript`, `cmd/evener-hub/frontend`, or Go files touched by E's own commits — biome/web-typecheck/Go gates not applicable.

Head: `ffb4cbae17a030601da3b5034fbf8787e6f37f4a`

obra:
Restack note for reviewers — this branch is now restacked onto current main (merge of main at `1717d4d87`; head `be80a6f6b8`):

- **C-era content collapsed.** The branch previously carried piece C's pre-landing commits; after the union merge, every shared file is identical to main's landed versions. The three-dot diff vs main shows only piece E's unique 11-file delta, all mobile-native, +746/−26.
- **pendingQuestions reconciled.** The branch's rewiring of pendingQuestions to canonical refs was redundant with #1737's landed canonical-refs fix (both sides had arrived at the identical body `[...liveAsksFor(conversation).values()].flat()`); it is collapsed to main's single implementation. The identity machinery E adds (`questionsIdentity`/`boundQuestionText`/`questionDefinition`) is additions-only on top of main.
- **The carried React-key Low is fixed** (`d453cd8de3`): truncated labels colliding as React keys at TimelineItem.tsx:187 — option rows now key by position (`questionOptionKey`, the TasksSheet id:index precedent), pinned in timeline.test.ts ("keys an ask's option rows by position, not by the bounded label") and TimelineItem.test.tsx (duplicate-key console spy). Red-proof executed: reverting to the label key reproduced React's captured duplicate-key report; restored → green.
- **Gate-forced fix** (`be80a6f6b8`): drafts saved by older builds under the full-question signature stopped loading through the new {key, digest} identity (draftRepository compares serialized definitions). `questionDefinition` normalizes both eras to the same {key, digest} pair — a legacy full question hashes to the same digest the identity computes for that canonical question — wired into draftRepository's questionDefinitions. The regression test ("a draft saved under the pre-identity full-question signature still loads through questionsIdentity") was red before, green after: an app update never silently drops a reader's saved answers.
- **Gates:** scoped suites 93/93 (questionAnswers, questionBatches, draftRepository, timeline, TimelineItem, transcriptPresentation) + 578/578 shared (state/conversation, conversation/project); `make test-native` exit 0 (1130 native + 908 shared tests, tsc, scripts); local roborev on the branch delta: "No issues found" (verified run against the branch checkout).

obra:
Round-2 fix for the combined review on `be80a6f` (the legacy bounded-definitions Medium) — head now `e340e23fd`:

The comparison now carries the era knowledge it lacked. `questionsIdentity`'s elements sign each question with a second digest — `boundDigest`, the hash of the byte-exact bounded copy (`boundQuestion(question, boundQuestionText)`) that the pre-identity sheet persisted from bounded timeline rows. `draftRepository.readQuestions` compares through the new exported `sameQuestion` predicate: a stored definition names the same question when its digest matches the current element's canonical digest OR its `boundDigest`. No stored row is rewritten; a reader's first save rewrites the legacy bounded definition canonically, so the widened window closes on first edit.

The bounded digest deliberately inherits that era's one blind spot — a same-key question changed only past the display bound compares equal, exactly as it did under the signature era that persisted it; that is pinned by an explicit test rather than left implicit. A genuinely different question still does not inherit the draft.

Red-first: "a draft saved under the pre-identity bounded-copy signature still loads through questionsIdentity" was RED on `be80a6f6b8` (the saved answer silently dropped — the finding itself) and is green on `e340e23fd`, alongside two further new tests (the era blind spot + digest distinctness) and the prior fix's full-question regression. Reachability verified in history: the pre-identity sheet signed `JSON.stringify(questions)` from `conversation.items` (bounded rows), with #1737's `boundQuestion` shape identical to today's `boundQuestionText` byte-for-byte.

Gates: scoped 8 files/134 tests green; `make test-native` exit 0 (102 files/1133 native, 7 files/908 shared, tsc, check:scripts). Local roborev on the fix delta: "No issues found." — verified against this delta (the `boundDigest`/`sameQuestion` symbols are absent at the base and appear throughout the reviewer's session log).

obra:
Round-3 fix for the combined review on `e340e23` (the legacy display-wrapper Medium) — head now `15ecbd50`, and this round closes the era family rather than one era at a time:

**The era inventory.** The full history of the persisted question_drafts.signature element was enumerated from git (both the main line and the pre-restack piece-E line; all release tags predate the draft code). Eight shapes total. Three reachable eras were RED on `e340e23` and are now covered by red-first regressions: the display-twin wrapper (the panel's finding — `d0f40080cb`→`1afff9f514` persisted the canonical ref inside a nested bounded `display` twin), the checkpoint key-first order (main #1096→#1488), and the shim parsed-fields-first order (main #1488→#1737). Two more reachable eras were already green and stay pinned (package-ordered full questions — the primary upgrade population — and bounded truncated copies). One shape was never persistable — its parse threw on every write ("question drafts never persisted", per `800742bc72`'s own message) — and the pre-#1096 checkpoint line is unreachable (no ancestor of HEAD, in no tag, predates question_drafts entirely).

**The fix.** `questionDefinition` now rebuilds every hashed element through `canonicalQuestion`: the display twin stripped, the fields rebuilt in the exact order the package's `liveAskQuestions` builds them — the same literal the identity hashes — so a digest depends only on WHICH question an element names, never on which era's builder serialized it. For package-built refs the rebuild is byte-identical, so no digest that matches today stops matching; the `boundDigest` input rebuilds the same way so both sides of the bounded comparison normalize identically. No stored rows rewritten; `sameQuestion` unchanged; draftRepository untouched.

Red-first on `e340e23`: 4 failed / 39 passed (the three era regressions, each `readQuestions` returning `{}` — the silently-dropped answer — plus the normalizer unit pin); all 43 green after, with all prior-era regressions green throughout. Gates: scoped 8 files/138 tests; `make test-native` exit 0 (102 files/1137 native, 7 files/908 shared, tsc, check:scripts). Local roborev on the delta: "No issues found." (the review prompt provably contains this delta's branch-only symbols — `canonicalQuestion`, `PersistedQuestionElement`, and all three new test names).

PR #1739, opened 2026-09-18
fix(sdk): settle publishes even when the local apply throws

obra:
## Summary
- Fourth piece of the #1549 (D6-core) decomposition, per Jesse's five-rounds rule (docs/superpowers/plans/2026-09-12-sdk-migration.md D6, plan comment on #1549). Plan line 4: "Settles publish even when the local apply throws: thunk extra, applyHubOverridesSettling, settledWrite (~115)."
- Stacks on #1736 (PR 3, draft-checkpoint port). Targets `main` for now; will be refreshed once the stack ahead of it merges.
- Source of truth is #1549's head dfde921602c17e70d393adfcbdeece2dcb4682d9; its keybindingsStore.test.ts is the oracle.

Three settle points were non-transactional against a wedged registry:
- `refreshFor` computed `settledWrite`'s reclassifying `persistDraft` call as a bare expression BEFORE `applyHubOverrides` ran, so a reconcile that then threw left the checkpoint reclassified on disk while the throw kept the settle from ever publishing. `applyHubOverrides`'s `extra` is now a thunk, resolved only after `reconciler.apply` succeeds.
- `saveDraft`'s post-rename and revision-conflict rejection branches called `applyHubOverrides` with no try/catch, so a reconciler throw (a wedged registry that has already rolled back) propagated straight out with `saving` never cleared. `applyHubOverridesSettling` wraps all three settle points (post-rename, conflict, confirmed-reply) so `settled` always publishes.
- `saveDraft`'s success and rejection paths treated every fenced reply the same, whether superseded (a later write took over) or lost-hub (support dropped with this write's own claim still intact - nothing else will ever settle it). `whyFenced`/`settleUnsettleableWrite`, over the fence's `lostHub` (re-added to `readyGenerationFence.ts` - dropped in #1728 as unused in that slice, now consumed here with its own test coverage via these keybindingsStore tests), tell the two apart.

Deviation from the plan line worth flagging: `saveDraft`'s post-rename/revision-conflict classification (`rejectionPayload`, already used by `patchOverrides`) did not exist in `saveDraft` before this change - it's new here, a prerequisite for the two throw-safety tests. Its two happy-path tests ("a revision conflict during saveDraft is a KNOWN outcome...", "a post-rename durable failure during saveDraft applies the carried canonical state...") move alongside them so the new branches aren't shipped untested; they weren't named in the plan's test list but are needed to exercise the classification the throw-safety tests wrap.

`replaceIf`/`replaceClassified` and the checkpoint-replaced adoption they enable are NOT part of this change - a later piece (#1549's "a settle adopts a checkpoint replaced under it") adds them.

## Tests that moved (byte-for-byte from #1549's keybindingsStore.test.ts)
- "a refresh's reconciler throw does not reclassify the checkpoint ahead of publishing the settle" - RED confirmed against #1736's keybindingsStore before this change (settledWrite's disk write happened even though the throw never let the store-state settle publish)
- "a revision conflict during saveDraft is a KNOWN outcome..." (added, prerequisite) - RED confirmed
- "a post-rename durable failure during saveDraft applies the carried canonical state..." (added, prerequisite) - RED confirmed (threw instead of applying)
- "a post-rename durable failure whose local reconcile throws still clears saving" - RED confirmed (hubError stayed null)
- "a revision conflict whose local reconcile throws still clears saving" - RED confirmed (hubError stayed null)
- "an older save's late reply does not clear a newer save's saving flag" - already passed on #1736's baseline (the superseded case's existing "return without touching state" was already correct); kept as a permanent regression pin
- "a rejection arriving after support went unknown still clears saving" - RED confirmed (`saving` stayed `true` forever - nothing else would have cleared it)

## Test plan
- [x] TDD: reverted the implementation to #1736's baseline, ran the 7 new/moved tests, confirmed 6 RED with the exact failure shapes described above, restored the fix, confirmed all 64 tests in the file pass
- [x] `npx vitest run keybindingsStore.test.ts readyGenerationFence.test.ts keybindings.test.ts transcriptDisplay.test.ts readyGenerationCallback.test.ts` - 105+64 passed
- [x] `npm run typecheck` (web) and `npm run check` (mobile-native) - both clean


obra:
## Review round disposition (cff9018)

Review Mediums folded in (TDD RED first for both):
- Conflict branch discarded `applyHubOverridesSettling`'s result and always rethrew the original rejection, unlike post-rename/confirmed-reply. Strengthened the existing "a revision conflict whose local reconcile throws still clears saving" test to assert the THROWN error matches the reconciler's own message (`/keybinding conflict/`), confirmed RED (it threw "revision conflict" instead), fixed by capturing and throwing `applyFailure ?? error`.
- `applyHubOverrides`'s stale-revision early return resolved a thunk `extra` regardless of the stale guard, so `settledWrite`'s checkpoint reclassification ran for an ignored payload. New test "a refresh's stale reply does not reclassify the checkpoint the settle would have adopted", confirmed RED (writeUncertain wrongly cleared, checkpoint wrongly reclassified), fixed: the stale path never resolves a thunk.
- "removeIf boolean discarded at every call site": resolved by the #1736 (p3) narrowing - void in p3, widened and consumed in #1741 (p5).

Low deferred, noted: `rawFrom` WeakMap in draftCheckpointPort.ts is never read in this slice (no consumer calls `removeIf` with a checkpoint `load()` produced yet); the re-export at index.ts.

High (threads.ts:2235's bare onReady): not this PR's mechanism - a separate small PR from main is being built by another lane for it.

Simplify pass folded into the same push:
- `settleWrite(payload, checkpoint, newerExternal, settled)` extracts the post-rename/confirmed-reply settle sequence. One correction to the suggested shape, measured: collapsing `settled`'s `draftConflict` to "spread true only when external" regressed two passing tests (the confirmed-reply's own happy path needs to force `draftConflict: false` explicitly - the draft it judges hasn't been cleared yet, so `applyHubOverrides`' own `staleDraft` check misreads the revision this write just confirmed as a conflict). Kept `settled` as a caller-built parameter instead: post-rename omits the key (letting `staleDraft` decide, unaffected either way there), confirmed-reply forces it - same as before, now deduplicated mechanically. Verified against the full 65-test suite before and after.
- `whyFenced`/`settleUnsettleableWrite`/`lostHub` collapsed to `settleLostHubWrite(generation, token)`; `lostHub` stays exported (per the plan, a later piece's per-row token is its second consumer) with the WHY prose only there now.
- The duplicated `typeof extra === "function" ? extra() : extra` resolve point was already removed as a side effect of the stale-path Medium fix above - nothing further needed.

Gates: keybindingsStore.test.ts 65/65, readyGenerationFence.test.ts clean, web typecheck/biome/lint clean, mobile-native `npm run check` clean, `make lint-package-imports` pass.

obra:
## Round 2 fixes (3567ea108) — also closes #1752

### High: stale GET leaves `hubLoading` permanently true
Root-caused, and confirmed it's the same bug behind #1752's CI-red native test (`mobile-native/src/nativePreferences.test.ts > keybindings: an older GET cannot replace an acknowledged write`, which fails identically through `refreshFor` with the same `expected true to be false` on `loading`). #1752's premise that main reproduces this is wrong — measured `git diff origin/main...pr-1739 --stat`: only `keybindingsStore.ts`, its test, and `readyGenerationFence.ts` changed; falsifying by reverting just `keybindingsStore.ts` to this branch's *prior* head reproduces the exact native failure (pasted below), so p4's own change is the root cause, not something pre-existing.

Root cause: `applyHubOverrides`'s stale-payload guard returns `false` without ever calling `extra` when it's a thunk (deliberately — a settling thunk's side effect must not run for a payload the guard is about to ignore), but `refreshFor`'s thunk is the only place that clears `hubLoading`, so a stale GET left it true forever.

Fix: `refreshFor` now checks `applyHubOverrides`'s return value and clears `hubLoading` directly when the payload was ignored, instead of relying on the thunk.

Added `expect(store.getState().hubLoading).toBe(false)` to the existing stale-reply test (`keybindingsStore.test.ts`), which previously exercised this path without asserting it, exactly as flagged.

Posted the reproduction on #1752 — closing it as fixed by this commit rather than pre-existing.

### Medium: post-rename settle leaves `draftConflict: true` with `draft: null`
Confirmed: `settleWrite`'s post-rename call passed no `draftConflict` in `settled`, so `applyHubOverrides`'s own `draftConflict: staleDraft(state.draft, payload.revision)` ran against the PRE-write draft (still present at that point) vs. the just-confirmed revision, reading the write's own success as a conflict — the exact hazard the confirmed-reply path already guards against by forcing `draftConflict: newerExternal`. Fixed the post-rename call the same way (`newerExternal = getState().revision > applied.revision`, `draftConflict: newerExternal`). Added `draftConflict: false` to the existing post-rename success test's assertions, which previously didn't check it.

### Medium: `UnreadableDraftError` maps to `storageUnavailable`
Not fixed here — this is piece 6's mechanism (`draftUnreadable`/`assertDiscardable`, discriminating `UnreadableDraftError` from a generic port failure in the restore path). Building piece 6 next.

### Medium: `removeIf` returns `void`
Not fixed here — pointer to #1741 (piece 5 does the void→boolean widening). Note: as of this reply #1741 has the boolean widening but its `discardStoredDraft` still casts a raw `unknown` to `Checkpoint` (the typed-contract half of the sibling Medium on #1736); will fix both in #1741's round.

### Low: comments reference review rounds / future migration pieces
Rewrote the three `keybindingsStore.test.ts` comments that said "RoboRev round N ..." to describe current behavior instead. Checked `draftCheckpointPort.ts` and `readyGenerationFence.ts` as cited: `readyGenerationFence.ts`'s new comment (the `lostHub` doc) has no temporal/roundrap language — no change needed there. `draftCheckpointPort.ts`'s header ("a later piece of the SDK migration brings a second store onto this same repository") is a genuine roadmap reference; will fix in #1736's own next touch (already flagged that PR's round-2 reply that this file is new in this stack, not pre-existing).

## Gates
- `npx vitest run keybindingsStore.test.ts` (appwire-client/typescript): 65 passed
- `npx vitest run src/nativePreferences.test.ts` (mobile-native): 25 passed (was 24 passed / 1 failed before the fix)
- (no index.ts/tsconfig.build.json touch, so `make test-api-package` falsification not required this round)

## Falsification
Reverted only `keybindingsStore.ts` to this branch's prior head (`cff9018`) and re-ran `keybindingsStore.test.ts`: 2 failures as expected — `hubLoading` `expected true to be false` on the stale-reply test, and `draftConflict: true` (expected `false`) on the post-rename test. Also re-ran mobile-native's suite against the reverted file: reproduced #1752's exact failure. Restored with `git apply`; `git status` clean before committing.

obra:
## Round 2 fixes (b366cdb5e)

### High: draft settlement ignores compare-and-remove refusals
Pointer to #1741 (p5, head `eff17ab9a`). Measured: `DraftPort.removeIf`/`DraftRepository.removeIf`/`discardClassified` return `boolean` at that head, and `discardDraft`/`settleWrite` check the result and re-classify via `restoreDraft` on refusal instead of clearing state unconditionally; `replaceIf`/`replaceClassified` (the atomic replace operation) also exist there. Not fixed in #1739 itself.

### High: `patchOverrides` can start while `saveDraft` is in flight
Fixed. The gate now also refuses on `state.saving || state.writeUncertain` - both write paths share one write token, so starting here would claim a new one, fence the checkpointed write's own reply out as superseded, and strand `saving` true with nothing left to clear it. Added `patchOverrides is refused while a checkpointed save is in flight`.

### Medium: `load()` clears `hasLastClassified` on empty storage
Fixed in #1736 (p3, commit `8f5798bdf`), inherited here through the restack. `load()` classifying an EMPTY store now sets `classifiedAbsent`, distinct from never having classified at all; `discardClassified()` is a no-op for it instead of taking the fresh-reload fallback. Not new work on #1739.

### Medium: `save()` does not update `rawFrom`
Also fixed in #1736/p3 (`8f5798bdf`), inherited the same way: `save()` now does `rawFrom.set(checkpoint, decoded)` for the checkpoint it was given.

### Low: post-rename rejection path accepts any structurally valid `applied` payload
Fixed. Applies the same semantic check the confirmed-reply path runs on its own reply (`loadError === undefined && revision >= requested`) before settling; a payload that fails it falls through to the existing "no reply" branch (uncertain outcome, `draftConflict: true`, checkpoint kept) instead of being reported applied. Added a parametrized test for both failure shapes (`carries a loadError`, `reports a revision below the one requested`).

### Low: `settleWrite`'s JSDoc contradiction
Fixed. Corrected the doc: both callers force `draftConflict` from their own `newerExternal` (verified by re-reading both call sites); the "post-rename caller deliberately omits it" clause was already false.

## Gates
- `npx vitest run keybindingsStore.test.ts` (appwire-client/typescript): 69 passed
- `npx vitest run src/nativePreferences.test.ts` (mobile-native) and `npm run check`: clean

## Falsification
Reverted only `keybindingsStore.ts` to this branch's prior head and re-ran `keybindingsStore.test.ts`: 3 failures as expected - the new patchOverrides test times out (saveDraft's reply never lands because the direct write claimed its token), and both new post-rename validation tests resolve instead of rejecting. Restored with `git apply`; `git status` clean before committing.

obra:

Issue #2, opened 2026-04-24
defCommunicateWithRequiredDataKeys heuristic creates unusable schemas for most keys


obra:
## Context

Toil declares per-node required data keys via \`outputs:\` in workflow YAML, which become the \`SERF_SUBMIT_RESULT_REQUIRED_DATA_KEYS\` env var. Serf's \`cmdutil.go:36\` reads it and wraps the communicate tool via \`agent.WithCommunicateRequiredDataKeys(...)\`.

## The bug

\`agent/profile_overrides.go:defCommunicateWithRequiredDataKeys\` (lines 186–276) generates a JSON schema for the \`output.data\` field by **guessing** each required key's type from its name:

\`\`\`go
case k == "components": propSchema = componentsSchema()
case k == "tasks":      propSchema = tasksSchema()
case strings.HasSuffix(k, "_doc") || "_document" || "_markdown": propSchema = { type: "string" }
case strings.HasSuffix(k, "_list") || "_ids" || ("s" && !"_results"): propSchema = { type: "array", items: { type: "string" } }
default:                propSchema = { type: "object", additionalProperties: false }
\`\`\`

**The default branch, combined with OpenAI strict mode, forces the key to be an empty object.** \`additionalProperties: false\` + no \`properties\` definition means the model can only emit \`{}\`.

## Observed failures

Two runs affected by this today in toil's \`pivot-cedar-thistle\` deliver:

- **\`surgeon\` (in \`plan_and_build.yaml\`):** when I declared \`outputs: [plan]\`, \`data.plan\` was forced to \`{}\`. The agent correctly observed the constraint (\"Because the output schema only accepts an empty \`data.plan\` object here, the full structured plan is included below\") and shoved the real plan into the \`message\` field as markdown. The \`plan_reviewer\` then rejected the (empty) plan.

- **\`e2e_tester\` (in \`integrate.yaml\`):** pre-existing. \`outputs: [story_results]\`. \`story_results\` ends in \`_results\`, hits the \`_results\` exception on the array heuristic → falls to \`default\` → forced empty. Silent data loss: \`fern-signal-river\` has \`data.story_results = {}\` despite the e2e_tester having content.

Only two serf nodes in toil declare \`outputs:\` today, and both are broken by this.

## Why the heuristic doesn't work

[...]

PR #1740, opened 2026-09-18
refactor(native): rows read from the model, older pages stay attached (D23c-2b, piece D)

obra:
Piece D of the #1580 decomposition plan (see the [decomposition plan comment](https://github.com/prime-radiant-inc/evener/pull/1580#issuecomment-5723104182)). **Stacked on #1737 (piece C)** — targets `main`, will be refreshed onto `main` after C lands.

## Summary
- Cut from #1580 (`53bb2b6ca`): `conversation.ts`, `conversation.test.ts`, and `mobile-native/src/liveImages.test.ts` (the live-image test suite this cutover also changes behavior for).
- The cutover-plus-deletions and the page-history layer are accepted at this size deliberately, per the plan comment: `main`'s `conversation.ts` already stages this as an explicit "Dual-write until c-2b" comment (folds into the package model AND runs the old row appliers, side by side, specifically so this cutover could land as one atomic step later) — that comment IS the flag, so a second, separate runtime flag on top of it would be redundant scaffolding. The page-history layer (`attachToSources`/`supersededBy`/`withPageHistory`/`retainedPageRow`/`capItems`'s orphan-trim) can't be split from the cutover either: `main` already supports `loadOlder` pagination against the row-applier world, and once rows come from `projectConversation` alone (which only re-derives the *current* live window), an older page loaded via `loadOlder` has nothing to reattach itself to on the next publish without this layer — shipping the cutover without it would regress pagination outright (this is exactly what round 3 and round 6 of #1580 itself already rediscovered after an earlier cut dropped it).
- Fixes this round's Low: `retainedPageRow`'s doc comment cited `project.ts`'s `clusterActivities` for identity/label/detail/state rules that piece C's `clusterActivityRun` now actually implements (`clusterActivities` itself just calls it, kept for `main`'s row appliers per piece C). Corrected the citation.

## A seam this branch did NOT cut cleanly at first, and how it's resolved
`conversation.test.ts`'s "shares one `askQuestionsByCall` scan between `projectConversation` and `pendingQuestions`" test called `pendingQuestions` (`mobile-native/src/questionAnswers.ts`, piece E's file) through `conversation.ts`'s own store. That's a real cross-piece dependency, not a D/E-independence violation in substance: the property under test — `liveAsksFor`'s memoization — lives entirely in `project.ts`'s `asksByTurns` (piece C, already `WeakMap`-keyed on `model.turns`), and `pendingQuestions` merely calls it. Removed the test from `conversation.test.ts` (it needed none of this piece's own cutover logic) and re-added it directly against `projectConversation`/`liveAsksFor` in piece E's `questionAnswers.test.ts`, where `pendingQuestions` itself lives — no file is shared between D and E's own diffs.

## Test plan
- `npx vitest run --config mobile-native/vitest.config.mts mobile/src/state/conversation.test.ts mobile/src/conversation/project.test.ts mobile-native/src/questionAnswers.test.ts mobile-native/src/questionBatches.test.ts mobile-native/src/draftRepository.test.ts mobile-native/src/liveImages.test.ts` — 510/510 pass


obra:
## Review round disposition + simplify (603341afd)

Restacked onto the new C (#1737) head, `10beda5ce`.

**RoboRev Medium (real, D's own) — an explicit output-image clear didn't supersede a page-owned attachment:** `withPageHistory`'s `projectedAttachmentSources` only named a source with a CURRENT attachment row. An authoritative `outputImages: []` (the wire's only way to say "these are gone" — appwire's nil/non-nil-empty/non-empty rule) removes the live attachment row and projects none at all, so an older page's attachment for that source survived the merge untouched after an explicit removal. Added `explicitlyClearedAttachmentSources` (scans `model.turns` for an item whose `outputImages` is defined and empty, distinct from omitted/undefined), unioned into the same set `retainedPageRow` already consults — the fifth attachment state the finding asked for. TDD RED first: a wire-shaped test (load an older page holding an attachment for source "key-Z", then a live `item/completed` for that source carrying `outputImages: []`) confirmed the attachment was wrongly kept before the fix.

**RoboRev Medium "bounded labels leak into answer composition"** (project.ts:1042-1081, questionAnswers.ts:11-18): verified on this branch — D stacks on C, not E, so `pendingQuestions` here is pre-E's bounded-identity change. `composeQuestionAnswers` reads `question.header`/`question.ifUnanswered` from the canonical `AskQuestionRef`s `pendingQuestions` returns, never from a bounded display copy — piece E's own change (`questionsIdentity`) touches only the identity/signature side, not composition. No stacking artifact once E and D are both on main: confirmed by reading, not by running a local E+D union (D and E are siblings on C, and a real union would need a throwaway merge commit this repo doesn't otherwise want).

**RoboRev Medium "visible question card with no answer action when askPending is false"** — the same rule #1731 (piece A) settled: `changesRows` (conversation.ts) explicitly excluded the wire's `askPending` from the fields that can change a row, true before piece A's round 4 and false after — `liveAskQuestions` now gates its whole scan on it, so a status-only frame flipping `askPending` with no item of its own can turn an already-projected tool row into a question row, or the reverse, with `turns` untouched by reference. Fixed `changesRows` to also compare `askPending`, and updated/rewrote three `conversation.test.ts` cases that asserted the OLD contract by name ("renders a live ask's question row before any snapshot says so") to assert the new one instead: an item completing is not the whole signal either direction, only the wire's `askPending` (arriving via its own status frame in production, on the same boundary that ends the round which posted the ask) decides whether the question row shows. `mobile-native/src/questionAnswers.ts`'s `pendingQuestions` was already correctly gated; the bug was entirely in this file's own row-change detection.

**RoboRev Low "dead helpers in project.ts"** and the takeover the earlier simplify pass asked for: `isActiveItem`, `activityState`, `projectItemAttachments` and `clusterActivities` (plus their now-orphaned helpers `inlineImageSrc`/`inputImage`/`outputImage`) were kept in project.ts for the store's old per-notification row appliers; this cutover deleted those appliers and never removed the exports meant only for them.
- Deleted `isActiveItem` (project.ts's local copy), `inlineImageSrc`, `inputImage`, `outputImage`, `projectItemAttachments`, `clusterActivities` — zero remaining callers.
- `activityState` is now private, taking `(item, turn)` instead of `(item, turnStatus)` to match its own callers' shape; internally it now calls the package's own `isActiveItem` directly (no more local alias).
- Dropped the "kept-for-compat" clause in conversation.ts's own comment citing `clusterActivities`.

**Also (small, safe simplify items from the same pass):**
- `pruneEvictedIds` returns before building its retained-identity set when there is no page ownership to prune (the common case, once loadOlder's pages have all been superseded).
- `supersededBy`'s third parameter renamed `attachmentSources` → `supersedingSources` (loadOlder deliberately passes every current identity, not just attachment sources, as the "one-line note at each call site" already explains).

**Not done this round** (left for a follow-up, not a code change here): unifying project.ts's `clusterActivityRun` with conversation.ts's `retainedPageRow`'s run-of-one header-building — both compute id/label/family/detail/state for a single-member cluster via different paths (one returns the original row verbatim, one rebuilds field-by-field from the member) — and collapsing `withPageHistory`'s two identity-scanning loops. Measuring whether the two run-of-one paths actually agree needs its own dedicated pass; this round's own scope already carried the askPending/attachment fixes above and I did not want to risk a silent behavior change in the clustering path without that measurement.

The relocated test the C-side review asked me to confirm ("shares one askQuestionsByCall scan between projectConversation and pendingQuestions"): this branch's own commit message already documents the relocation — it moved to `questionAnswers.test.ts` (piece E), which is where `pendingQuestions` itself lives; confirmed it is there.

**Gates:** `npx vitest run --config mobile-native/vitest.config.mts mobile/src` (`npm run test:shared`) — 804/804 pass. `mobile-native`: `npm run check` — clean; `npm test` — 791/791 pass; `npm run check:scripts` — clean. `make lint-package-imports` — PASS.

Head is `603341afd`.

obra:
## Round 2 findings routed

- High 1: "Live status notifications omit AskPending" (`internal/appprojector/appwire_projection.go:1632-1637`) — measured end to end rather than re-derived by reading: a new test in #1833 drives a real `agent.Session` through a real ask_user round, bridged into a real `*Server` the way `cmd/evener/serve.go` wires it, and reads back the actual recorded `thread/status/changed(awaiting)` notification. **It carries `AskPending=true` already.** The projector citation is real (it never touches the field) but isn't a bug: the projector deliberately holds no session handle (`stampAskPendingOnStatusChange`'s own doc comment), and the server stamps the field afterward at its single notification egress, reading the live session's `HasPendingAsk()` through `cmd/evener/serve.go`'s `liveThreadEnvelopeSource`. #1833 is test-only (no production change) and settles this: the live wire is fine, and no projector change is needed. Retracting the planned "projector: thread/status/changed carries askPending" fix — it would have been a redundant, parallel tracking mechanism for something that already works.
- High 2 (qualifier crash) and High 3 (carrier failure restoring a stale ask, `agent/session_queue.go:1302`) — both A's, fixed in #1806 ("piece A-server") and #1731's now-TypeScript-only diff (04584a21c). See the disposition on #1731 and #1806 for the specific mechanism and falsification for each.
- Medium "truncated question rows double as answer model" (`mobile/src/conversation/project.ts:961`) — this is D's own; picking it up when D restacks, with a test.

obra:
## Restacked onto trimmed A/B/C (1a0b754ae), plus D's own finding

Restacked onto #1737's new head (57476d548). Own diff is isolated (4 files vs C's head): liveImages.test.ts, project.ts, and the conversation store/test rewrite.

- Medium "Truncated question rows double as answer model" (project.ts:961) — measured before changing anything: this store's own pre-existing test ("bounds a question row's prose and leaves the model's asks whole") already documents and enforces the intended design — bound the row for DISPLAY, answer from \`liveAskQuestions\`' canonical, uncut copy off the same conversation. I first "fixed" this by excluding "question" rows from capAndTruncate entirely, which broke that existing test; reverted once I saw it contradicted an established, deliberate design point rather than a gap. The real gap was mobile-native's OWN separate \`pendingQuestions\` (mobile-native/src/questionAnswers.ts, not this file), which read the bounded row directly instead of the canonical model — fixed in #1738 (E) via \`liveAsksFor\`. What D's own test didn't cover: the LIVE incremental path (item/completed -> thread/status/changed), only hydrate/openProjected. Added a test driving that path and asserting the same "bounded row, whole canonical copy" invariant — falsified against the OLD (unfixed, pre-restack) code path shape by hand-tracing, confirmed correct against the current design; no production change needed here.
- The three mobile Mediums from #1738's round 2 review (live warnings differ from re-projection, stale question rows after askPending clears, shared truncation/cap helpers not wired into the store) are all resolved by this rewrite as a side effect of unifying the live path through reducer.applyNotification + projectConversation (this store's own \`changesRows\` explicitly checks \`askPending\`, and \`capAndTruncate\` imports \`capItems\`/\`truncateItem\` from project.ts with zero locally duplicated implementations) — confirmed by reading, no separate fix needed.

Gate results:
- \`npx vitest run\` (mobile-native's test:shared config over mobile/src): 808 passing
- \`npx vitest run\` (mobile-native's own suite): 794 passing
- \`npm run check\` (mobile-native): clean

obra:
## Restacked onto main after A (55c6eab)

Rebased `claude/c2b-d-native-page-history-cutover` onto piece E's new tip (`ffb4cbae1`, itself rebased onto piece C's new tip, which is rebased onto main — see #1737/#1738). D's own 4 commits (`40ec422ab`, `5c73e57b2` WIP, `1a0edbd09`, `1a0b754ae`) stack directly on C's old tip (`57476d548`), confirmed by `git log --first-parent`, so this was `git rebase --onto ffb4cbae1 57476d548 claude/c2b-d-native-page-history-cutover`.

### Conflict resolution (hunk by hunk)

Only the first commit (`40ec422ab`) conflicted; the other three applied cleanly on top.

**`mobile/src/state/conversation.ts` (4 conflict regions):**

1. **Import block** — took D's side (imports the identity/truncation helpers directly from `project.ts` — `activityIdentity`, `attachmentSourceIdentity`, `capItems`, `ownTimelineIdentities`, `projectConversation`, `timelineIdentities`, `timelineIdentity`, `truncateItem` — plus a re-export of the constants). This is D's own fix for the exact duplication Low I filed as **#1925** on #1738 — landing D retires that issue.
2. **The store's own `MAX_ITEM_BYTES`/`truncateText`/`truncateItem`/`capItems` wrapper block** — D's side is empty (deletes it outright): D imports these directly now, so the wrapper functions #1737's own fix (re-exporting rather than duplicating) built are themselves now dead. Kept only `exceedsByteLimit` (this store's own truncation-freeze check; `project.ts` has no equivalent).
3. **`findFoldedItem`/`projectSingleItem`** (the old single-item incremental projection, used by the item/started+item/completed row appliers) **vs `supersededBy`** (D's new page-history helper) — took D's side; the old functions' only caller is the switch statement hunk 4 deletes.
4. **The entire `applyNotification` switch statement** (~400 lines: item/started, item/completed, item/reasoning/summaryTextDelta, item/toolOutput/delta, the live `case "warning"`, etc.) **vs D's 9-line replacement** (`capAndTruncate(withPageHistory(state.conversation, projectConversation(applied)))` plus the gap-detection tail already unconflicted around it) — took D's side. This is the actual cutover: every row is now a projection of the model, with no per-notification-type row applier left to disagree with it.

**`mobile/src/state/conversation.test.ts`:** two small "replaces the same transcriptKey..." test-mechanics conflicts (D's `openRunningTurn()` helper vs the old inline `makeConversation`/`open()`) — took D's side (same assertions, different setup). One larger conflict: 8 tests exercising the deleted live "warning" row applier's `kind: "failure"` construction (title/hint sanitization, polymorphic `warning` field, short live-row ids, no-active-turn drop) — took D's side (deletes them): the mechanism they tested no longer exists (confirmed zero `case "warning"`/`"failure"`-construction left in production code), and D's own `"drops a warning that arrives with no active turn"` test (unconflicted, elsewhere in the file) covers the one surviving invariant; the rest (title/hint sanitization, polymorphic field, short ids) are already covered through the canonical path by #1737's own `project.test.ts` tests.

**`mobile-native/src/liveImages.test.ts`:** 6 conflicts, all "which race-condition test/comment wording wins" between HEAD's pre-cutover assumptions and D's redesign. Kept one HEAD-only ADDITIVE test verbatim (`"keeps the item's images when a live settle carries an empty list and an older snapshot arrives"` — a real, still-valid invariant D's branch predates) and took D's side for the rest, including the one direct behavioral disagreement: HEAD asserted a live image replacement survives a racing "older" snapshot; D asserts **the snapshot commits over the live replacement** (`"commits the snapshot's image over a live replacement"`), citing the read-response-is-the-snapshot-cut invariant (evidence: `cmd/evener-hub/frontend/src/stores/threads.ts:applyHydrationResponseCut`). Verified this is what the resolved production code actually does (not just asserted) by running the test — passes.

### Bugs found and fixed while getting D's own tests green (none were review findings; these are fixes to make the piece's own committed tests pass)

1. **`appwire-client/typescript/reducer.ts`'s `mapTurn`/`settleFirstMatchingTurn` didn't preserve array identity on a no-op fold** (new commit `107c797a7`). `Array.prototype.map` allocates a new array unconditionally even when every element comes back identical — so a warning/injected-steer/turn-completed frame naming a turn outside the loaded window always produced a *new* `turns` array, even though nothing in it changed. The store's gap-detection (`applied.turns === state.conversation.turns`) depends on exactly this identity to know "the reducer couldn't place this frame, ask for a reread" — with the always-new array, the 3 tests `"rereads when a warning/an injected steer/a turn settle names an active turn this window does not hold"` failed (no reread ever fired) and, in real use, the phone's transcript would stay silently incomplete until the next unrelated resync. Fixed both functions to return the *original* array when nothing actually matched/changed; verified with `reducer.test.ts` (216/216 still pass) and the 3 mobile tests (now pass).
2. **A stale test assumption about reasoning precedence** (new commit `79f26868f`, part 1). D's own `"keeps the accumulated reasoning when the completion carries %s"` test asserted a settled item's *seeded* `reasoningSummaries` always wins over its own completion text — but the reducer's actual `mergeReasoning`/`mergeCompletedText` contract (and main's own, later, more carefully falsified fix — see `project.test.ts`'s `"shows the completion's authoritative text over a stale seeded reasoningSummaries entry"`) says the completion's non-blank text is authoritative; only a *blank* completion falls back to the seeded summary. Measured: reverting to D's original assertion fails against both the reducer's documented contract and main's parallel test: I traced `mergeReasoning`'s own doc comment ("wins over whatever wireItemToModel seeded... on the settled wire item's own text") and `mergeCompletedText`'s ("explicit wire text, including empty text, remains authoritative") to confirm main's fix, not D's test, matches the reducer's real behavior. Fixed the test to expect the completion's text for the non-blank case, keeping the seeded-summary expectation for the blank-completion case.
3. **A fixture using the wrong item status** (same commit, part 2). `"applies a reasoning delta aimed at a later clustered member in place"` seeded its target item `status: "completed"` before firing a live `item/reasoning/summaryTextDelta` at it — but a delta only ever streams into a still-running item, and (per fix #2's own precedence rule) a completed item's non-blank text wins over its reasoningSummaries, so the delta was invisible. Added an optional status parameter to the local `reasoningItem` helper and passed `"inProgress"` for this one test.
4. **Three tests needed a third microtask tick** (same commit, part 3). `"rereads when %s names an active turn this window does not hold"` awaited `Promise.resolve()` twice; the identical `requestRehydrate` path two rows below in the same describe block (`"rereads for an item frame the model has no rule for"`) already needs three. Not a behavior bug — just an off-by-one in the await count once bug #1 above was fixed and the reread actually started firing.
5. **`project.ts`'s unused `isInProgressStatus` import** (new commit `55c6eab8b`) — the two functions that read it directly (`isActiveItem`, `activityState`) already moved onto the package's own `isActiveItem` export in an earlier round; this Low from #1737-r3 (found "still open" on that PR) is fixed here since D's cutover is what actually removes the last local caller.

### r3 disposition (`.superpowers/sdd/2026-09-12-mobile-landing-queue-cont/reviews/1740-r3.md`)

- **Medium** (`.trim()` on untyped title/hint, `project.ts:505`) — **fixed**, inherited from #1737's rebase (`joinWarningParts`/`hasWarningText`).
- **Medium** (answer submission reads truncated display copies, not canonical refs) — **already fixed**, inherited from #1738 (piece E): `mobile-native/src/questionAnswers.ts`'s `pendingQuestions` now reads `liveAsksFor(conversation)` directly instead of the bounded `conversation.items`.
- **Low** (unused `isInProgressStatus` import) — **fixed**, commit `55c6eab8b` above.
- **Low** (attachment filenames bypass the display bound, `project.ts` `truncateItem`'s `attachments` case / `TranscriptImages.tsx`) — **confirmed still open.** `truncateItem`'s `default:` case (attachments) still returns the item unchanged; `AttachmentRef.name` is genuinely unbounded (its own comment only justifies leaving `src` uncut). Filed as **prime-radiant-inc/evener#1930** rather than fixed inline, per the Lows rule.

### Gates (from `.claude/worktrees/sdk-d21-native-model`)

- `npx vitest run --root .. --config mobile-native/vitest.config.mts mobile/src/state/conversation.test.ts mobile-native/src/liveImages.test.ts` — 371/371 passed.
- `npx vitest run --root .. --config mobile-native/vitest.config.mts mobile/src` (full shared suite) — 820/820 passed.
- `mobile-native && npx vitest run` (full own suite) — 825/825 passed.
- `cmd/evener-hub/frontend && npx vitest run ../../../appwire-client/typescript/reducer.test.ts` (touched by the `mapTurn` fix) — 216/216 passed.
- `cmd/evener-hub/frontend && npm run typecheck` / `npx biome ci ../../../appwire-client/typescript` / `npm run lint` — all clean.
- `make test-api-package` — PASS (qualified; no index.ts/qualifier change, ran anyway since a package file changed).

`git diff -M --stat origin/main...HEAD` (full C→E→D stack):
```
 appwire-client/typescript/reducer.ts      |   21 +-
 mobile-native/src/QuestionSheet.tsx       |   55 +-
 mobile-native/src/draftRepository.test.ts |   38 +
 mobile-native/src/liveImages.test.ts      |   75 +-
 mobile-native/src/questionAnswers.test.ts |  388 +-
 mobile-native/src/questionAnswers.ts      |   83 +-
 mobile-native/src/questionBatches.ts      |    5 +-
 mobile-native/src/screens.tsx              |   12 +-
 mobile/src/conversation/project.test.ts   |  360 +-
 mobile/src/conversation/project.ts        |  716 ++-
 mobile/src/services/conversation.ts       |    7 +-
 mobile/src/state/conversation.test.ts     | 7449 ++++++++++-------------------
 mobile/src/state/conversation.ts          | 1653 ++-----
 13 files changed, 4426 insertions(+), 6436 deletions(-)
```

D's own diff on top of E (`ffb4cbae1..HEAD`):
```
 appwire-client/typescript/reducer.ts  |   21 +-
 mobile-native/src/liveImages.test.ts  |   75 +-
 mobile/src/conversation/project.ts    |  125 +-
 mobile/src/state/conversation.test.ts | 7481 ++++++++++++---------------------
 mobile/src/state/conversation.ts      | 1576 ++-----
 5 files changed, 3057 insertions(+), 6221 deletions(-)
```

Head: `55c6eab8b6fe04b09f5d692869651440019b7fbd`

obra:
Restack note for reviewers — this branch is now restacked onto current main (merge of main at `1f1221e5b5`; head `f50dd2fea6`; merges cleanly against current main `d7d4921f8` — that overnight advance is Go-side and touches none of this delta's files).

- **Stack context.** Piece C landed via #1737 + #2136; piece E via #1738 (including its draft-era normalizers and the TimelineItem position keys). This branch's old C-era and E-era content collapsed to main's landed shapes exactly: the three-dot diff vs main shows only D's own delta — 6 files (conversation.ts, conversation.test.ts, project.ts, reducer.ts, index.ts, liveImages.test.ts).
- **The restack + convergence loop.** 38 local review rounds; rounds 2–37 each surfaced a NEW mechanism (37 Mediums total — retained-history reconciliation, snapshot reconciliation, truncation, warning seating, page ownership across renames and reissues), every one fixed red-first with a regression. No root-cause repeats. Per-round verbatim logs are retained in the session record.
- **The F8 guard.** loadOlder's at-cap signal is main's pre-cap-overflow form verbatim (conversation.ts:2302, `mergedInput.length > RETAINED_ITEM_CAP` over the pre-cap merged set). All landed F8 pins green before and after (I7 4/4; retains-newest-500; stops-offering; does-not-request-null; the byte-cap cases as the stateless-rewrite tests with the mapping documented).
- **The Sep-18 review's findings.** The transient-warnings Medium was real on the union and is fixed red-first: turn/completed settling a turn still holding warning items now requests the canonical reread — the wire never persists warnings, so the read is the honest drop (pinned in 3f46da787e). The unused-imports Low is subsumed: imports rewritten from actual usage.
- **ONE OPEN LOW, disclosed for adjudication** (found by the final local review round and left per the wave's convergence budget; mechanism-adjacent to the warning-seating family): `conversation.ts:765–778` — when a growing activity cluster absorbs the anchor a transient warning followed, the seating walk places the warning after the whole cluster, so a diagnostic can move below later activities ("A, warning" becomes "A, B, warning" when B joins A's cluster without an intervening message). Suggested fix: split activity clusters at warning anchors, or seat warnings before clustering, with a regression covering a later activity joining the anchor's cluster.
- **Gates on the pushed head:** scoped conversation.test.ts 491/491 (443 pre-loop + 48 loop regressions); shared suites 952/952; tsc clean; `make test-native` exit 0.

obra:
Round 38 follow-up: the two Mediums from the combined review on `f50dd2f` are fixed on `cb950eed3f` (fast-forward, no force), plus the one CI lint item.

**Medium 1 (older-page dedup dropped a whole clustered row on a single member match)** — the loadOlder merge loop now judges clustered members individually: a member the live window already holds is the only member deduped, and the remaining members rebuild as a cluster via `rebuildOwnedCluster`, the same shape `retainedPageRow` uses on the rehydrate side. The no-duplicate and all-members-held cases keep their previous whole-row behavior. Regression: `loadOlder rebuilds a paged cluster from members the live conversation does not hold` — RED on `f50dd2fea6` (`expected undefined to be defined`), GREEN on `cb950eed3f` (the cluster survives rebuilt as `[m1, m3]`, the live boundary row appears exactly once). The pre-landing test that pinned the whole-row skip (`skips an incoming cluster whose member identity is already present under a new top-level id`) is superseded in place by `rebuilds an incoming cluster...`: the retention assertion flips per this finding while its member-dedup assertion is unchanged.

**Medium 2 (rehydrate failure-fold marked every survivor page-owned unconditionally)** — the unconditional survivor add is removed from the fold; the add now lives in the merged-items rewrite guarded by `seenRenamedIds.has(survivorIdentity)`, mirroring the loadOlder path exactly as prescribed. Regression: `does not keep a resolved failure on screen when the snapshot's own row won the fold rename` — drives the fold-rename collision (the snapshot's own `failure:t9` row wins, the renamed page row is dropped), then a resolving reread, asserting the resolved failure leaves the screen instead of sticking until cap eviction or a thread transition (RED: expected the failure rows to be gone, got 1; GREEN), and stays gone across a live-frame reproject.

**CI lint item** — run 35831125831 on `f50dd2fea6` failed the `web` job's lint step only (web-typecheck and web-test passed): Biome `organizeImports` at `appwire-client/typescript/index.ts:334`, the new reducer export block had `itemTextPresence` before `itemIdentityMatches`. Fixed by the prescribed two-line swap (9b75fe0a66); `make test-web` passes all three sub-gates locally.

**F8 guard** — unchanged, green before and after: the at-cap signal remains the pre-cap-overflow form (`mergedInput.length > RETAINED_ITEM_CAP` before capping, conversation.ts:3045), and the landed cap pins plus the null-cursor pin pass on both heads.

**Gates** — conversation suite 493/493 (491 base + 2 new regressions), adjacent shared suites 236/236, `make test-native` exit 0 (native 1151/1151, shared 954/954, tsc clean, check:scripts), `make test-web` exit 0. Local roborev on the fix delta (`f50dd2f..cb950ee`): "No issues found." (round 1, trap-guarded — the review's verification log carries this branch's own test names and symbols).

The Low (transient warning ordering below an absorbing cluster) remains open by design and tracked for a follow-up, per the merge bar.

Commits: 9b75fe0a66 (export sort), fd9c06f3a3 (Medium 1), cb950eed3f (Medium 2).

obra:
Round 39 follow-up: the round-2 panel Medium (unstable attachment-warning anchors) is fixed on `6ddd3cb30d` (fast-forward, no force), with two local-review hardenings of the same seat-position seam.

**The Medium fix** — `arrivalAnchor` now resolves attachment rows through their stable source identity (`attachmentSourceIdentity(item) ?? timelineIdentity(item)`), so a warning that followed an attachment row survives a source reissue under a new wire ID instead of losing its anchor and being pruned on the next rebuild. Regression: `keeps a warning seated across a reissue of the source item its attachment row was keyed to` — RED on `cb950eed3f` (the warning vanished after the reissue), GREEN on `6ddd3cb30d`, and it stays seated through a further rebuild frame.

**Two local-review follow-ups in the same seam** (each red-first in its own commit):
- `seats a warning anchored to a later cluster member after the cluster's whole attachments run` — the seating walk consumes the anchored row's whole attachment run before seating the notice, preserving the arrival position [source, attachments, warning] rather than landing between a source and its attachments. The in-code comment documents why this is load-bearing, not cosmetic: the cap cut drops a leading attachment whose source fell off the cut and only scans a leading run, so a notice seated inside that run would orphan the attachments behind it and their lingering source identities would make loadOlder's F10 admission rule refuse genuine older page copies forever.
- `drops an attachment at the cap cut even when a warning is seated at its source's position` — pins the cap-cut behavior the seat rule depends on.

**F8 guard** — pins 11/11 before and after; the at-cap signal remains the pre-cap-overflow form. The prior rounds' regressions (36/37/38) stay green. The warning-ordering Low remains open by design, tracked for a follow-up.

**Gates** — conversation suite 496/496, shared suites 957/957, native suite 1151/1151, tsc clean, `make test-native` exit 0. Local roborev on the fix delta (`cb950ee..6ddd3cb`): rounds 1 and 2 surfaced the two seat-position follow-ups (each fixed red-first in its own commit), round 3 came back "No issues found" within the 3-round cap, trap-guarded (this branch's own test names appear in the review's verification log).

Commits: 297e3d8da9 (stable source-identity anchoring), 6e64f9007b (seat after the attachments run), 6ddd3cb30d (a cluster member's whole attachments run).

PR #1741, opened 2026-09-18
fix(sdk): a settle adopts a checkpoint replaced under it

obra:
## Summary
- Fifth piece of the #1549 (D6-core) decomposition, per Jesse's five-rounds rule (docs/superpowers/plans/2026-09-12-sdk-migration.md D6, plan comment on #1549). Plan line 5: "A settle adopts a checkpoint replaced under it: replaceIf/replaceClassified + the settle paths (~140)."
- Stacks on #1739 (PR 4, settle publishes on throw). Targets `main` for now.
- Source of truth is #1549's head dfde921602c17e70d393adfcbdeece2dcb4682d9; its keybindingsStore.test.ts is the oracle.

`saveDraft`, `discardDraft` and `refreshFor`'s `settledWrite` composed the draft port with an unconditional save/remove, so a concurrent writer's newer checkpoint (another window, another app version) landing while a write's outcome was unknown or a discard was pending would be silently overwritten or hidden behind "no draft" - the replacement never surfaced.

- `draftCheckpointPort.ts` adds `replaceIf` to `DraftPort` (the atomic twin of `removeIf`: a compare against `expected` before writing `next`) and `replaceClassified` to the repository (settles the checkpoint most recently classified, atomically; a refusal means another writer replaced it).
- The ephemeral no-port fallback's `replaceIf` reports success unconditionally, matching `removeIf` (round-24 shape): there is no concurrent writer a compare-and-swap could lose to without real storage behind it.
- `saveDraft`'s three settle points (post-rename via `removeIf`, revision-conflict via `replaceClassified`, confirmed-reply via either depending on `newerExternal`) and `discardDraft` (via `discardClassified`) now check the port's own verdict and adopt whatever `restoreDraft` finds on disk instead of assuming their own write landed. `rebaseDraft` needed no change - measured against #1549: it never calls `replaceIf`/`replaceClassified`, only `persistDraft`.
- mobile-native's `NativePreferenceDraftBackend` grows the matching `replaceIf`, wired through `nativeKeybindingDrafts`; the production adapter in `NativePreferencesProvider.tsx` implements it (same compare, a write instead of a removal). `nativeTranscriptDrafts` stays without a real `replaceIf` (the pre-migration transcript design never settles atomically) - only the shared backend interface requires the method, so its own storage shape and every test backend implementing it report `() => false`.

## Deviations from the plan, measured
- Two of the moved tests' expected `draft` objects drop the `generation` field #1549's version carries: `KeybindingsDraft`'s generation stamp is D6's piece #7, not yet in this store's draft type. Substance (checkpoint adoption) is unchanged; only the fixture shape is smaller.
- The two "discard refuses and re-classifies" cases named in the plan are entangled with PR 6's scope (`draftUnreadable`/`assertDiscardable`): the unreadable-record variant is fully about that state and is deferred to PR 6 whole; the readable-record variant is included here, renamed slightly and with its `draftUnreadable` assertions dropped (that field doesn't exist yet) - kept as "a discard refuses and re-classifies when the record has been replaced".
- "without a draft port, a revision conflict's settle does not misread the ephemeral fallback as a concurrent replacement" passes unchanged against #1739's baseline too (there's no compare-and-swap to misread before this PR's mechanism exists) - kept as a permanent regression pin, not a RED case.

## Test plan
- [x] TDD: reverted the implementation (keybindingsStore.ts, draftCheckpointPort.ts, testing/draftStorage.ts) to #1739's baseline, confirmed 3 of the 4 new/moved tests RED with the exact failure shapes (checkpoint overwritten instead of adopted; discard silently succeeded over a replaced record), restored the fix, confirmed all 68 tests in the file pass
- [x] `npx vitest run keybindingsStore.test.ts` - 68 passed
- [x] mobile-native: `npm run check` clean, `npm test` 796/796 (2 new replaceIf-propagation cases in nativePreferenceDrafts.test.ts)
- [x] `npm run typecheck` (web), `npx biome ci ../../../appwire-client/typescript`, `npm run lint` - clean


obra:
## Review round disposition (f0e1b58)

Mediums, verbatim order:
1. **patchOverrides can claim the shared write token while saveDraft owns it.** Measured: #1549 (the oracle) already carries this fix in its patchOverrides guard (`state.saving || state.writeUncertain` added to the unavailable-condition, with a comment: "both paths take the same write token, so starting here would fence the checkpointed write's own reply out and strand it saving"). New here, not in #1549 or in main. Added the guard plus #1549's own oracle test "a direct write is refused while a checkpointed write owns the payload"; confirmed RED (timed out - patchOverrides went ahead and sent its own request onto the same held reply) before the fix.
2. **Unreadable draft + assertEditable locks discard.** Not fixed here - piece 6's mechanism (draftUnreadable + assertDiscardable), per the plan (docs/superpowers/plans/2026-09-12-sdk-migration.md D6 item 6: "An unreadable record never locks the section, store side only").
3. **Post-rename branch never computed newerExternal, unconditional removeIf.** Fixed via the settleWrite unification from #1739's simplify pass: the post-rename branch now computes `newerExternal` (`getState().revision > applied.revision`) and calls the same `settleWrite`, so refusal-adoption applies uniformly. Added the RED-first test for the post-rename refused site per the simplify pass's note (item 5): "a post-rename durable failure whose checkpoint was replaced adopts the replacement".
4. **rawFrom staleness after save()/replaceClassified().** Fixed: both now update the input checkpoint's own rawFrom entry to the bytes just written. New dedicated `draftCheckpointPort.test.ts` (RED-first, using the byte-aware `memoryDraftStorage` test port) proves a `load()`-then-edit-in-place-then-`save()`-then-`removeIf()` sequence on the same reference now succeeds.

Low: `NativePreferenceDraftBackend` no longer requires `replaceIf` for `nativeTranscriptDrafts`'s consumers - split onto a narrower `NativeKeybindingDraftBackend` that extends it.

Simplify pass (#1739's items 4-6) folded in:
- Settle id-minting: settledWrite has no checkpoint reference to reuse an id from, so the other two settle points (post-rename, confirmed-reply) now mint fresh too instead of reusing `checkpoint.id`, for one consistent rule.
- The settleWrite refusal-adoption unification (item 1 above) also folds discardDraft's and settledWrite's own refusal handling into the same shape.
- `NativePreferencesProvider.tsx`'s `deleteIf`/`replaceIf` now share one `matches(key, value)` compare.

Residual found while running the gates, not part of this change: `mobile-native/src/nativePreferences.test.ts`'s "keybindings: an older GET cannot replace an acknowledged write" fails on `main` already (confirmed against origin/main with no relevant file changes) - filed as evener#1752.

Gates: keybindingsStore.test.ts 72/72 (incl. draftCheckpointPort.test.ts), mobile-native `npm run check` clean + `npm test` 795/796 (the one pre-existing failure above), web typecheck/biome/lint clean, `make lint-package-imports` pass.

obra:
## Round 2 fixes (ff4604727, via restack)

### High: stale refresh leaves `hubLoading` stuck true
Fixed in #1739 (p4, commit c179f7042/3567ea108), inherited here through the restack onto p4's new head. `refreshFor` now checks `applyHubOverrides`'s return value and clears `hubLoading` directly when the payload was ignored, instead of relying on the settle thunk. Also root-caused and closed #1752 (mobile-native's parametrized `keybindings: an older GET cannot replace an acknowledged write`), which was the same bug. See #1739's round-2 comment for the falsification.

### Medium: post-rename save leaves `draftConflict` stuck true
Also fixed in #1739/p4, inherited the same way: the post-rename `settleWrite` call now forces `draftConflict: newerExternal` (computed from `getState().revision > applied.revision`), matching the confirmed-reply path's existing posture, instead of leaving it to `applyHubOverrides`' own `staleDraft` check (which reads the pre-write draft against the just-confirmed revision and misreads a successful write as a conflict).

Note while resolving this restack's rebase conflict: p5's original commit had independently added the same `newerExternal` computation for the post-rename path (for the checkpoint-adoption logic your review's Medium didn't flag), but without setting `draftConflict` from it — the two fixes are now merged into one.

### Low: `invalidDraft()` conflates error types
Fixed in #1736 (p3, commit 8c0c4448c), inherited through the restack. Split into `invalidDraft` (decode path, `UnreadableDraftError`) and `invalidUserRules` (editDraft/saveDraft's own input validation, plain `Error`), same message, different exception type.

### Low: unconstrained `WeakMap` generic
Also fixed in #1736/p3, inherited the same way: `createDraftRepository<Checkpoint extends object>`.

All four findings are resolved as of this head via the stack's restack (p3 8c0c4448c + p4 c179f7042), not by new work on this PR. Nothing further to fix here this round.

## Gates (post-restack)
- `npx vitest run keybindingsStore.test.ts draftCheckpointPort.test.ts` (appwire-client/typescript): 79 passed
- `npx vitest run` (mobile-native, full suite - fast, ~1.5s): 796 passed, 83 files

obra:
## Round 2 fixes (100793d7d)

### Medium: empty initial load lets `discardClassified` delete a concurrent draft
Fixed in #1736 (p3, commit `8f5798bdf`), inherited here through the restack. `load()` classifying an EMPTY store now sets `classifiedAbsent`; `discardClassified()` is a no-op for it instead of taking the fresh-reload fallback and deleting a concurrent writer's draft. Not new work on #1741.

### Medium: breaking port API change (`removeIf`/`deleteIf` void→boolean, required `replaceIf`)
Refuted. Measured every implementer of `KeybindingDraftStorage`/`DraftPort<Checkpoint>` in the repo:
- `keybindingsStore.ts`'s internal `memoryDraftStorage()` fallback (the no-port default) - updated (`removeIf: () => true, replaceIf: () => true`).
- `mobile-native/src/nativePreferenceDrafts.ts`'s `nativeKeybindingDrafts()` - updated (`removeIf`/`replaceIf` both delegate to the backend's `deleteIf`/`replaceIf`, both boolean).
- `appwire-client/typescript/testing/draftStorage.ts`'s generic `memoryDraftStorage<Checkpoint>()` test double - updated.
- `mobile-native/src/nativePreferences.ts` and `bindNativePreferences.ts` only take `KeybindingDraftStorage` as a parameter type (consumers, not implementers) - nothing to update there.

No other implementer exists. `@evener/appwire-client` is consumed only in-repo (`cmd/evener-hub/frontend`, `mobile-native`) - there is no published/external consumer this change could break, and per CLAUDE.md I don't add backward compatibility without Jesse's explicit approval. Not fixed (nothing to fix - all three real implementers are already updated in this stack).

### Medium: `saveDraft` bypasses `writeQueue`
Measured: pre-existing on `main` (`saveDraft` has never referenced `writeQueue`, before or after #1728/#1729/#1736/#1739/#1741). Filed #1801 with the full analysis (including the related but distinct fix #1739 already made: `patchOverrides`'s gate now refuses while `saving`/`writeUncertain`, which stops a direct write from *starting* during a checkpointed save but does not serialize the two paths through one queue). Not fixed in this stack.

### Low: `storageUnavailable` not set when `replaceClassified` throws
Fixed. `settledWrite`'s catch now returns `{ storageUnavailable: true, draftError: DRAFT_SAVE_FAILED_MESSAGE }`, matching the conflict branch and `persistDraft`. Added a `failReplace()` knob to the shared `memoryDraftStorage` test double (matching `failSave`) and a test: `settling an uncertain write whose replaceClassified throws marks storage unavailable, not just draftError`.

### Low: comments citing rounds/PRs
Measured each citation:
- `keybindingsStore.ts` (the `memoryDraftStorage` fallback's doc, "round 24 Medium 1") - fixed, dropped the citation.
- `keybindingsStore.test.ts:261` and `:398` ("RoboRev round 21 Medium 1", "RoboRev round 24 Medium 1") - fixed, rewrote both to state the invariant alone.
- `keybindingsStore.ts:390`/`:750-751`/`:789` ("roborev PR #884 round N") - measured these are pre-existing on `main`, not introduced or touched by this stack's diff; left alone as out of scope for this PR (a residual on `main`, not #1741's).
- `mobile-native/src/nativePreferenceDrafts.test.ts:6` - measured: this comment cites no round/PR/finding number, only describes current behavior (`nativeKeybindingDrafts` propagating `deleteIf`/`replaceIf`'s verdict faithfully). No change made; likely a stale line-number citation in the review.

Fixed. Returns `discardStoredDraft`'s result. Added two tests asserting `true` (removed) and `false` (nothing stored).

## Gates
- `npx vitest run keybindingsStore.test.ts draftCheckpointPort.test.ts` (appwire-client/typescript): 83 passed

## Falsification
Reverted `keybindingsStore.ts` and `testing/draftStorage.ts` to this branch's prior head and re-ran `keybindingsStore.test.ts`: 3 failures as expected (`drafts.failReplace is not a function`, and both `discardStoredKeybindingDraft` tests receiving `undefined` instead of a boolean). Restored with `git apply`; `git status` clean before committing.

obra:

PR #1742, opened 2026-09-18
fix(web): threads.ts's ready callback cannot begin a generation for a replaced client

obra:
## Summary
threads.ts wired a bare client.onReady handler. When a stale client's onReady fires after being replaced, it would call beginReadyGeneration unconditionally, clearing subscriptions and fencing out the current client's generation. This applies the readyGenerationCallback guard pattern from #1729 to threads.ts.

## What and Why
The guard wraps each onReady callback with its registration client and the current client getter. Only the most recent registration for a given client object may fire, preventing stale handlers from affecting the current connection.

## Test
Test case: "a stale client's ready callback cannot begin a generation for a replaced client" in threads.test.ts
- Wires a stale client and captures its onReady callback
- Wires a current client to replace it
- Fires the stale callback and verifies mutationAuthorityRefs was NOT cleared

## Gates Run
- vitest: threads.test.ts (312 tests pass)
- npm run typecheck: no errors
- npm run lint: no issues


obra:
## Falsification (4e30be196)

Test verification passed: 

**Failing without guard (HEAD~1):**
```
Before stale callback: { refCount: 1, mutationAuthorityRefsSize: 1, staleCallCount: 0 }
After stale callback: { mutationAuthorityRefsSize: 0 ... }
```
AssertionError at line 9566: `expect(authorityRefsClearedToEmpty).toBe(false)` 
- The guard test correctly FAILS without it: stale callback runs its body and clears mutationAuthorityRefs

**Passing with guard (HEAD):**
All 313 tests pass, including the new guard test.

The test observes that the stale callback's first action in its body is `setState({ mutationAuthorityRefs: new Set() })`, which empties a previously non-empty Set. With the guard in place, this never happens—the stale callback's body doesn't execute at all.

obra:
Pre-existing in threads.ts, keybindings.ts and transcriptDisplay.ts (rewireClient is only called with a non-null client); out of scope for this PR, tracked in #1749.

PR #1748, opened 2026-09-18
refactor(mcp): share header-injecting HTTP client between mcpprobe and manager (#1497)

### LINKED ISSUE #1497: mcpprobe: headerRoundTripper/httpClientWithHeaders duplicate agent/internal/mcp/manager.go:859-880
Noticed by the /simplify reuse reviewer on #1496 (mcpprobe transport fix): agent/mcpprobe's `headerRoundTripper` + `httpClientWithHeaders` re-implement the header-injecting round tripper in agent/internal/mcp/manager.go:859-880. Consolidate onto one shared helper (an internal package both can import) so the probe and the manager cannot drift on how MCP HTTP headers are applied. Pre-existing duplication; out of scope for the flake fix.

obra:
Fixes #1497

## Root cause

`agent/mcpprobe` and `agent/internal/mcp` each carried a private copy of the
same header-injecting HTTP plumbing: a `headerRoundTripper` type plus an
`httpClientWithHeaders` constructor (mcpprobe's took a base client, the
manager's did not). Because they were independent, the two MCP HTTP paths could
drift on how a server's configured headers are applied — the probe could keep
sending a header the manager no longer did, or vice versa.

## Change

Move the wrapper into a new shared `agent/internal/mcphttp` package and route
both callers through `mcphttp.ClientWithHeaders`:

- `agent/internal/mcp/manager.go` passes `nil`, producing a fresh client that
  wraps `http.DefaultTransport` — byte-for-byte the behavior its old
  `httpClientWithHeaders(headers)` had.
- `agent/mcpprobe/mcpprobe.go` passes its per-call base client, preserving the
  copy semantics the probe's transport fix (#1496) depends on: the caller's
  client is not mutated and its transport is wrapped.

No behavior change. The duplicate types/functions are deleted from both files;
the existing manager/probe header tests now exercise the shared helper, and the
shared helper gets its own unit tests for the nil-base and base-copy paths.

## Test evidence

```
$ go test ./agent/internal/mcphttp/... ./agent/mcpprobe/... ./agent/internal/mcp/...
ok  	primeradiant.com/evener/agent/internal/mcphttp	0.049s
ok  	primeradiant.com/evener/agent/mcpprobe	0.483s
ok  	primeradiant.com/evener/agent/internal/mcp	0.869s

$ gofmt -l agent/internal/mcphttp/ agent/mcpprobe/ agent/internal/mcp/   # no output
$ go vet ./agent/internal/mcphttp/... ./agent/mcpprobe/... ./agent/internal/mcp/...   # exit 0
```

Test-first: `agent/internal/mcphttp/mcphttp_test.go` was written before the
package existed and failed with `no non-test Go files`; adding
`mcphttp.go` turns it green.


obra:
Roborev follow-up on the origin-normalization finding, resolved at head `3d8fc7310`:

**1. Case-insensitive hostname — fixed.** `ClientWithHeaders` now lowercases the endpoint hostname and `HeaderRoundTripper.RoundTrip` compares with `strings.EqualFold`, so a same-host redirect differing only in letter case keeps its headers. Covered by the mixed-case branch of `TestClientWithHeaders_HostScopedInjection`.

**2. Unparseable/hostless endpoint — fixed.** When `url.Parse` fails or `u.Hostname()` is empty, `ClientWithHeaders` returns the base transport unwrapped, so the client injects nothing (fails closed) instead of injecting for every host. Covered by `TestClientWithHeaders_HostlessEndpointFailsClosed`.

**3. Port/scheme ("full origin") comparison — not implemented; the premise is incorrect and the fix would regress legitimate redirects.**

`http.Client` does *not* strip sensitive headers on a same-hostname redirect that only changes port or scheme. Its policy in `net/http/client.go` is:

```go
func shouldCopyHeaderOnRedirect(initial, dest *url.URL) bool {
	ihost, err1 := httpguts.PunycodeHostPort(initial.Hostname())
	dhost, err2 := httpguts.PunycodeHostPort(dest.Hostname())
	...
	ihost, ok1 := ascii.ToLower(ihost)
	dhost, ok2 := ascii.ToLower(dhost)
	return ok1 && ok2 && isDomainOrSubdomain(dhost, ihost)
}
```

`Hostname()` drops the port and there is no scheme check (verified against the toolchain in this repo, go1.27.0). So the stdlib already forwards `Authorization` on `https://host:443 → https://host:8443` and on `http://host → https://host`. Our transport sending configured headers on those redirects therefore *matches* `http.Client`; it does not re-add anything the client "deliberately strips" — the client does not strip there. (The same function is why the case-insensitive fix above is correct.)

Tightening to full-origin equality would make the wrapper *stricter* than the `http.Client` it wraps and would drop credentials on the common, legitimate same-host `http→https` redirect, breaking MCP handshakes for servers that redirect to TLS. That is a behavior regression with no security gain relative to the stdlib.

The real leak this change targets — re-injection of credentials on a cross-**host** redirect — is closed by the hostname scoping at this head, with `TestClientWithHeaders_HostScopedInjection` asserting the cross-host request carries no `Authorization`.

obra:
Update at head \`595a60d41\`: I implemented the full normalized-origin scoping, so all three sub-points of the finding are now addressed (superseding my earlier note that argued sub-point 3 was out of scope).

- **(a) case-insensitive host** — origin host is lowercased; \`originOf\` and \`RoundTrip\` compare normalized origins. Mixed-case same-host requests still get their headers.
- **(b) same-host different port/scheme** — the transport now stores the endpoint's normalized origin (scheme + lowercased host + effective port) and injects only on an exact origin match, so a same-host redirect to another port or scheme gets no configured credentials. End-to-end test \`TestClientWithHeaders_SameHostDifferentPortRedirectDropsHeaders\` (two \`httptest\` servers on 127.0.0.1, different ports) asserts the target sees no \`Authorization\`. The clone in \`RoundTrip\` is what makes this hold end-to-end: \`http.Client\` copies headers from the original un-injected request, and the transport declines to re-add them for the foreign origin.
- **(c) unparseable/hostless endpoint** — such an endpoint yields an empty origin, which injects nothing (fail closed), never "every host".

Also added \`TestClientWithHeaders_SameOriginRedirectKeepsHeaders\` so ordinary in-origin path redirects keep their credentials.

Verification: \`go test -race -count=1 ./agent/internal/mcphttp/...\` passes; \`go test -count=1 ./agent/mcpprobe/... ./agent/internal/mcp/...\` passes; \`go vet\` and \`gofmt\` clean.

obra:
IPv6 origin fix at head `c7d68c8ff`: `originOf` now builds the hostport with `net.JoinHostPort(host, port)`, so an IPv6 host is bracketed and distinct IPv6 origins cannot collide (e.g. `https://[::1]:443` vs `https://[::1:1]:443`). Added `TestClientWithHeaders_IPv6OriginsDistinct` (recording transport: matching IPv6 origin gets the header, a distinct one does not). Focused tests pass under `-race`; `gofmt`/`go vet` clean; rebased onto current main.

PR #1753, opened 2026-09-18
fix(web): keep a summary-less delegate row single-level at tools/activity/full (#1253)

### LINKED ISSUE #1253: web: delegate rows render a stray empty summary line at tools/activity/full, and the delegate card depends on the summary region
Follow-up split out of #1173, which fixed only the chat/intent regression. Raised by roborev's review of PR #1238.

## Symptom

At tools/activity/full verbosity a delegate row still takes the two-level disclosure path and mounts an empty `data-body-trigger="true"` summary line, which pulls the body chevron off the intent line. Same visible symptom as #1173, but it is **older**: #1173's own body records that "the same empty-line pattern already existed for delegate rows at tools/activity/full (where the summary defaults open); #1170 widened it to failed delegates at chat/intent". #1170's widening was fixed; this is the case underneath it.

## Why the obvious fix is wrong

Gating the summary region on there being summary text, i.e. `summaryFallback = hasSummaryText && (summaryConfigDefault || (failed && !superseded))`, does remove the stray line, but it also removes a region the delegate card depends on. Measured: with that gate, `src/panes/session/transcript/tools/subagentModule.test.tsx` fails **1 of 41** ("stable delegate attention and lifecycle own the card while child content status changes"); reverted, it passes **41 of 41**. So for a delegate row the summary region is not merely empty, and gating it away is not the fix.

## Second, related finding

`ToolCallItem.tsx` checks the raw `statedIntent` rather than the derived `intent`, so a legacy delegate whose description comes from a prompt/task fallback still forces `summaryOpen = true`. Switching that check to the derived `intent` fixes the delegate case but breaks two `TurnBlock.test.tsx` tests that pin the intended behaviour for non-delegate intent-less rows:
- "an intent-less tool shows its own derived summary at chat and tools levels"
- "an intent-less read_file shows the Read summary, never 'Action summary unavailable'"

## Direction

Decouple the delegate card's rendering from the summary region, or have `ToolRow` treat a summary-less intent-bearing row as single-level, instead of gating the region away. Both findings above are real observations whose proposed one-line fixes are demonstrably wrong; the work needs a design pass on how the card and the two-level row layout are coupled.

## Evidence

- `cmd/evener-hub/frontend/src/panes/session/transcript/ToolCallItem.tsx` - `summaryFallback` / `summaryOpen`, around lines 339-341.
- `cmd/evener-hub/frontend/src/panes/session/transcript/ToolRow.tsx` - two-level vs single-level row render.
- roborev review on PR #1238, medium findings at `ToolCallItem.tsx:339` and `:341`.
- Local verification: `npx vitest run src/panes/session/transcript` -> 1881 passed with the shipped code; with the gate applied, `subagentModule.test.tsx` -> 1 failed / 40 passed.

obra:
Fixes #1253

## Root cause

At verbosity levels whose content vector defaults the summary line open
(`tools`/`activity`/`full`, `toolCalls: true`), a delegate row took the
two-level disclosure path even though a delegate renders no summary text
(`subagentModule` owns its presentation, so `ToolCallItem` passes
`rowSummaryText = ""`). `ToolRow` decided "does this row have a summary line?"
from the `summary` **prop**, which `ToolCallItem` deliberately blanks while the
summary line is collapsed. For a delegate that prop is *always* empty, so
`bodyTriggerOnSummaryLine` was true and `ToolRow` mounted an empty
`data-body-trigger="true"` summary line, pulling the body chevron off the
intent line.

Gating `summaryFallback` on summary text (the "obvious fix") is wrong for the
same reason from the other side: it cannot tell "the summary line is collapsed"
from "there is no summary text", and it removed the region the delegate card's
body trigger depends on (measured: `subagentModule.test.tsx`
"stable delegate attention and lifecycle own the card while child content status
changes" fails). Likewise, switching `ToolCallItem`'s raw `statedIntent` check
to the derived `intent` is not the fix: it only masks one of the two shapes
(stated intent vs. prompt/task fallback) and regresses two `TurnBlock.test.tsx`
tests that pin intent-less behaviour.

## The change

`ToolRow` now distinguishes "the summary line is collapsed" from "this row has
no summary text":

- New optional `ToolRowProps.hasSummaryText`: whether the row has any summary
  TEXT at all, independent of whether the line is currently open. It defaults
  to the passed `summary`'s own presence, so direct `ToolRow` callers are
  unchanged. `ToolCallItem` passes its already-computed `hasSummaryText`.
- `summaryToggle = twoLevel && summaryTextPresent`: the intent button only
  toggles a summary disclosure when a summary line actually exists.
- `bodyTriggerOnSummaryLine` requires summary text, so no empty summary region
  is ever mounted; `bodyTriggerOnIntentLine` is always true for a summary-less
  row, so the body trigger (and therefore the delegate card) stays reachable
  folded or open.
- A summary-less intent-bearing row is genuinely single-level: the intent
  disclosure button is suppressed, and `.bodyTrigger` is the row's one control
  with one chevron (roborev Medium finding on the first head - the previous
  shape gave the same body disclosure two controls and two chevrons).

This is a general mechanism at the shared `ToolRow` layer, not a delegate
special case: it is the same "collapsed vs. absent" distinction any
summary-less two-level row needs. No `statedIntent`/`intent` check is changed,
so the intent-less `TurnBlock` behaviour is untouched.

### One pinned-test reference updated

`ToolCallItem.test.tsx` "delegate tool rows use a single top-level disclosure
trigger owned by ToolCallItem" previously counted
`[data-testid="tool-row-trigger"]`. Now that the consolidated single control is
the body trigger, it counts `[data-testid="tool-row-body-trigger"]` and asserts
zero intent triggers. The test's intent - exactly one top-level disclosure
control per delegate row - is preserved (and strengthened); the change is
driven by the roborev finding that the two-control shape is wrong, not by
reaching green. `#1173`'s own "single-level" test (`tool-row-body-trigger` +
`data-body-trigger-intent`) is untouched and still passes.

## Test evidence

New tests in `ToolCallItem.test.tsx` fail on the shipped code for the exact
reported reason:

```
npx vitest run src/panes/session/transcript/ToolCallItem.test.tsx -t "delegate row at the"
# before: 3 failed (tools/activity/full), received the stray
#         <div class="_summaryLine_" data-body-trigger="true"> ... </div>
# after:  5 passed
```

Added coverage: prompt-fallback (description-less) delegates at
tools/activity/full, and the delegate row's single body control mounting and
unmounting the card (the roborev Low finding).

Full changed-area suite (includes the #1173 tests, the delegate card test, the
two-level grammar tests, and the TurnBlock intent-less tests):

```
npx vitest run src/panes/session/transcript
# Test Files  86 passed (86)
# Tests  2000 passed (2000)
```

Other files that exercise `ToolRow`:

```
npx vitest run src/dev/surface-sections/transcript.test.tsx \
  src/panes/session/composer/RepoLocation.test.tsx \
  src/panes/settings/sections/TranscriptDisplayCard.test.tsx
# Test Files  3 passed (3)
# Tests  26 passed (26)
```

Geometry guard for the delegate row's intent line:

```
node scripts/layoutguard/run.mjs delegate-open-widget-inline
# PASS - ... in all 6 fixtures
```

Typecheck and lint:

```
npx tsc --noEmit --incremental false   # exit 0
npx biome check <changed files>        # No fixes applied
```

Only additive tests plus the single selector update described above; no other
assertion was weakened or deleted.


obra:
Rebased onto `origin/main` (`daf43f6b1`) and addressed all three findings from the `e46b7eb` review at the new head.

**Medium — body trigger accessible name (`ToolRow.tsx`)**: fixed. `summaryLabel` now falls back to `statedIntent` when there is no summary text (previously it degraded to the bare `"Tool call"`/`"Failed"` fallback), and `.bodyTrigger` carries `aria-describedby={statusId}` when a status is present. New test `a summary-less delegate row names its sole body control from the intent and describes its status` fails before the fix (verified by stashing the `ToolRow` change) and passes after.

**Low — tap target (`toolcallitem.module.css`)**: fixed. Added a `@media (max-width: 899px) { .bodyTrigger { min-width/min-height: var(--tap-min); justify-content: center } }` block mirroring `iconbutton.module.css`'s phone floor, and the `[data-intent-trailing]` width reservations now use `var(--tap-min, 14px)` so the grown control's width is reserved. Pinned by the new `delegate-body-trigger-tap` layoutguard case (mutation-verified: removing the phone block measures 14x14 and fails).

**Low — bare intent geometry (`ToolRow.tsx` / `toolcallitem.module.css`)**: fixed. Added `.row[data-intent-trailing="true"] > .intent { flex: 0 1 auto; min-width: 0; max-width: calc(100% - var(--tap-min, 14px) - var(--space-2)) }`, the same constraint `.intentTriggerContent`/`.trigger` already carry, so a trailing-affordance-less summary-less row hugs its text and never springs the body chevron to the line's far edge. Pinned by the new `delegate-body-trigger-inline` layoutguard case, mutation-verified both ways (max-width removed → the long intent wraps the trigger to line 2; `flex: 1 1 0` → the short fixtures spring the trigger 72/532px away).

No test was weakened or deleted; the only pre-existing assertion touched is the one already updated on the earlier head.

One unrelated change is included because the rebase surfaced it: `origin/main` (`daf43f6b1`) still has `RepoLocation.test.tsx` importing `gateRequests`, which #1755 renamed to `gateSettlements`, so `tsc` (and therefore `make build-web`) fails on the merge ref. Ported that one call site exactly as the upstream fix does, so this branch's checks can pass.

obra:
Rebased again onto `origin/main` (`7cdf1bfa5`), which now contains the `gateSettlements` port (#1826). My temporary copy of that port is dropped; the branch is now just the four #1253 commits. Head: `2ae9e009e`.

obra:
Rebased onto `origin/main` (`e692ca1e7`) and fixed the `web` CI failure at the new head `b5a84ed43`.

**Why `web` failed but `static-build` passed:** `web` runs `make test-web`, whose lint step is `biome ci src ../../../appwire-client/typescript`; `static-build`/`build-web` never run biome. The failure was therefore not the stale-base `gateRequests` breakage — it was a lint error I introduced:

```
src/panes/session/transcript/toolcallitem.module.css:239:1
lint/style/noDescendingSpecificity
✖ Descending specificity selector found. This selector specificity is (0, 1, 0)
ℹ This selector specificity is (0, 3, 0)
→ Found 1 error. (exit 1)
```

**Root cause:** my `[data-intent-trailing]` override `.row[data-intent-trailing="true"] > .intent` (specificity 0,3,0) was placed *before* the plain `.intent` base rule (0,1,0) at line 239.

**Fix:** moved the `.intent` base rule (with its comment) above the `[data-intent-trailing]` override group, so source order ascends in specificity. The override is strictly more specific, so the cascade — and every computed value — is unchanged; only the source order changes. No rule was disabled and no assertion weakened.

**Verified before pushing:**
- `cd cmd/evener-hub/frontend && npm run lint` (= `biome ci src ../../../appwire-client/typescript`) → exit 0 (Checked 1357 files).
- `node scripts/layoutguard/run.mjs` → all cases PASS, exit 0 — including `delegate-body-trigger-inline`, `delegate-body-trigger-tap`, `delegate-open-widget-inline`, `toolrow-chevron-nowrap` — so the reorder changes no geometry.

Head: `b5a84ed43` (5 commits: the four #1253 commits plus this lint fix). Waiting on CI.

obra:
Follow-up: the biome reorder was necessary but not sufficient — a **second, distinct** `web` failure was hiding behind it, which I found by pulling the job log directly (`gh api .../actions/jobs/<id>/logs --allow-escape-sequences`):

```
FAIL  src/styles/token-contract.test.ts > panes/session/transcript/toolcallitem.module.css has no chromatic literal outside tokens.css
AssertionError: expected [ '#1253', '#1253' ] to deeply equal []
Test Files 1 failed | 533 passed (534)
```

`token-contract.test.ts` scans the comment-intact CSS text for chromatic literals, and its hex branch matches a bare `(#1253)` PR reference. The two `[data-intent-trailing]` comments I added wrote it that way. The guard deliberately whitelists the house-style `issue #1253` / `Issue #1253` form via a negative lookbehind.

**Fix:** `(#1253)` → `(issue #1253)` in both CSS comments. The TSX/TS comments also carry `(#1253)`, but that guard only scans `.css` under `src` and CI showed exactly one failing test, so I left them.

**Verified locally with the full `web` job mirror** (`make test-web && make build-web && make test-web-browser`):
```
PASS  web-typecheck
PASS  web-test
PASS  web-lint
PASS  web-layoutguard / web-overflowguard / web-shellguard / web-spawnguard
PASS  web-transcriptscrollguard / web-retirementguard / web-skillguard
```
Head: `5c4a11c31`.

PR #1765, opened 2026-09-18
fix(webui): /model pre-start validation tolerates an unloaded catalog (#1085)

### LINKED ISSUE #1085: Pre-start /model validation fail-closes against null pane catalog (~250ms window after mount)
Followup to #1084 (branch `spawn-slash-autocomplete`).

Pre-start `/model` validation fail-closes against a null pane catalog in a ~250ms window after mount: the user gets a spurious "unknown value" toast. Recoverable via retry.

Suggested fix: skip pre-start model validation when the pane catalog is null and let the post-start check decide.

obra:
Fixes #1085

## Root cause

The spawn form's pre-start `/model` check validated the typed value against
`resolveSpawnModelItems(scopedModelCatalog)` in `cmd/evener-hub/frontend/src/panes/spawn/Spawn.tsx`. The pane catalog only commits after `CATALOG_SETTLE_MS` (~250ms) from mount, so `scopedModelCatalog` is `null` through that window. `resolveSpawnModelItems(null)` returns `[]`, so *every* value — known or not — failed to match and the check fail-closed with a spurious `/model: unknown value "..."` toast, aborting the launch.

## Change

Treat a catalog that never committed for the current scope as "don't know", not "unknown value": when `scopedModelCatalog` is null and there is no proven scope/loader staleness (`scopeMismatch`), skip the enum match and forward the typed value as the launch scalar, letting the start call's own check decide. A scope-matched catalog that omits the value, and a stamp proven to belong to another scope/loader (`scopeMismatch`), still fail closed. A provider/model shape check keeps the refusal for shapeless input (e.g. `nope`), matching the remote-launch path.

## Test evidence

- New regression test `a known /model value starts before the pane catalog lands instead of fail-closing` (with a deferred `model/list`, so the null-catalog window is deterministic). Red before the fix (`expected '/new' to be '/s/local%3Aabc123'`), green after.
- `npx vitest run src/panes/spawn` → 534 passed, 1 failed: `closing the handoff without choosing requires an explicit model before Start` times out at the default 5s. This is pre-existing and load-sensitive: it fails identically on the unmodified base and passes at `--testTimeout=30000`.
- `npm run typecheck` → clean.
- `npx biome check --write` on both touched files → no fixes applied.

Not merged.


obra:
Addressed roborev review (`6e177a5`) in `b12f08e9b`:

- **Medium — required-model bootstrap unreachable on an unloaded catalog:** `slashModelBootstrap` now applies the same "unknown catalog, no scope mismatch" rule as `doSpawn`. A shape-valid `provider/model` bootstraps and `thread/start` judges it; a scope-matched catalog must still list the value, and proven staleness still withholds Start. New test: first `model/list` rejects, launch resolution reports no default, typed `/model openai/gpt-5` enables Start and reaches `thread/start`.
- **Low — stale bootstrap comment:** rewritten to describe the new forwarding behavior and where Start is still withheld.
- **Low — stale stamp comment:** rewritten to distinguish never-committed (forward shape-valid values) from proven scope/loader mismatch (still fail closed).
- **Low — malformed-value branch uncovered:** new test types `/model foo` under a deferred `model/list`, asserting the `unknown value "foo"` toast and no `thread/start`.

Full `src/panes/spawn` suite: 537 passed. `npm run typecheck` and `npx biome check` clean.

PR #1766, opened 2026-09-18
fix(execenv): charge the ignore budget per compiled rule, not source byte (#1018)

### LINKED ISSUE #1018: Ignore-rule budget charges source bytes, so it does not bound compiled matcher memory
## What the budget does and does not bound

PR #985 adds an aggregate budget for ignore-rule discovery, `maxGlobIgnoreTotalBytes` (32 MiB), charged against each `.gitignore`'s source bytes plus a 512-byte per-file overhead. That bounds two real things: how much source a single glob or grep call reads, and how many rules files it can retain (roughly 65,000 at the overhead figure).

It does not bound the memory those rules occupy. go-gitignore compiles one regexp per rule line, and a compiled regexp costs far more than the line that produced it — measured at roughly 2 KB per line. So:

- a typical 149-byte `.gitignore` charges 149 bytes and retains about 39 KB, a factor of 265
- a file at the 1 MiB per-file cap charges 1 MiB and retains on the order of a gigabyte

A call that stays comfortably inside the 32 MiB budget can therefore hold far more than 32 MiB of compiled matchers, which is the quantity the budget exists to bound. This is the same failure shape as the original #497 defect: an accounting figure that does not track the resource it is meant to protect.

## Fix shape

Charge per compiled rule rather than per source byte: count rule lines as they compile and bill each at a measured constant, keeping the typed refusal and its own kind so the message can name the aggregate rather than a single file. The per-file overhead can then go, since the line charge subsumes it. Pin it with a test that builds many small rules files whose combined source is far under the byte budget while their compiled forms cross the ceiling, asserting on rules retained rather than on the returned result.

Pre-existing relative to #985, which introduced the budget; found in that PR's review and left out of scope deliberately so the accounting change gets its own measurement and test.

obra:
Fixes #1018

## Root cause

The call-wide `.gitignore` retention budget (`maxGlobIgnoreTotalBytes`) charged each rules file's **source bytes** plus a fixed 512-byte per-file overhead. But `go-gitignore` compiles one `regexp.Regexp` per rule line, and a compiled regexp costs far more than the line that produced it — measured at ~2 KB per rule on this machine. A `.gitignore` of a few hundred bytes therefore retained hundreds of times that in compiled matchers, so a call could stay comfortably inside the 32 MiB budget while holding far more matcher memory than the budget existed to bound. This is the same failure shape as #497: an accounting figure that does not track the resource it protects.

## Change

- `GlobBudget` now charges **per compiled rule line** at the measured constant `globIgnoreRuleBytes = 2048`, rather than per source byte.
- `readIgnoreFile` no longer splits the file twice; it counts the rule lines (mirroring `go-gitignore`'s own blank/comment skips exactly, via `compiledIgnoreRuleCount`) and bills them before handing the already-split lines to `CompileIgnoreLines`.
- The per-file overhead (`globIgnoreFileOverheadBytes`) is gone, since the per-rule charge subsumes it: retention is bounded by rules, and the listing budget already bounds how many files a walk can read.
- The `budgetRulesTotal` refusal keeps its own kind and now names the aggregate explicitly ("compiled and retained N bytes of .gitignore matchers across the call").

The `maxGlobIgnoreTotalBytes` ceiling is unchanged at 32 MiB, which at 2 KB/rule is about 16,000 rules.

## Test evidence

`TestLoadIgnoreSetRefusesTooManyRetainedRules` (rewritten) builds 40 tiny rules files whose combined source is **2,240 bytes** under a **32,768-byte** budget, while their 320 compiled matchers cross it. Before the fix the call returned **no error**:

```
loadIgnoreSet over 40 directories each carrying 8 rules in a few dozen source bytes
(combined source 2240, far under the 32768-byte budget) = <nil>, want a *globBudgetError:
charging source bytes missed the 320 compiled matchers entirely
```

After the fix it refuses with `budgetRulesTotal` after retaining 24 rules (49,152 bytes of matchers).

`TestRetainCompiledRulesChargesPerRule` pins the per-rule charge and that an empty rules file is free. `TestCompiledIgnoreRuleCountMirrorsGoGitignoreSkips` pins that the count matches what the library actually compiles (blank lines and column-zero comments skipped; an indented `#` line still compiles).

Commands run:

```
go test ./agent/execenv/ -count=1     # ok, 15.8s
go vet ./...                          # exit 0
gofmt -l <changed files>              # clean
```

## "Do not" constraints

No behavior outside the retention accounting changed: the per-file byte cap, the listing/entry/live budgets, and the returned `ignoreSet` semantics are untouched.

## Threat model (what this bound does and does not guarantee)

The accounting is **best-effort, not a proof**. It charges an *estimate* of each rule's compiled matcher cost, produced by a forward parser of the rule line.

**What it guarantees:** the original #1018 defect is fixed — the budget no longer tracks source bytes while ignoring the matchers they compile into — and every expansion vector found in review is charged: long literals, glob stars (rewritten to `([^/]*)`), counted repetitions and their nesting, optional branches, capturing-group bracket pairs, Unicode property escapes (bare, bracketed, and under repetition), leading-zero bounds that Go treats as literal, and a literal `]` in the first position of a character class.

**What it does not guarantee:** a construct whose compiled program is larger than the estimate can still pass the 32 MiB `maxGlobIgnoreTotalBytes` ceiling. The charge comes from a second parser that must agree byte-for-byte with go-gitignore's `getPatternFromLine` rewrite plus Go's `regexp/syntax`, and there is no fixed point short of reimplementing that parser. Each review round found another divergence, so treat a further bypass as expected rather than surprising.

**Where it is recorded in code:** the note above `classifyIgnoreLine` in `agent/execenv/gitignore.go` and above `maxGlobIgnoreTotalBytes` in `agent/execenv/search_patterns.go`.

**Follow-up:** #1971 — charge the real compiled program size (`len(prog.Inst)`) instead of an estimate. The `{0,m}` optional-branch pricing is approximate (compiled ~80–96 KB billed ~67 KB with a max-only factor) and stays in scope for that follow-up, not this PR.


obra:
Addressed the roborev findings in a6319a1cc (squashed):

**High — long patterns undercounted by the flat per-rule charge: fixed.** `retainCompiledRules` now charges `rules*globIgnoreRuleBytes + retainedSourceBytes`. The per-rule constant is still the term that fixes the original short-line undercount; the source term covers the expression-sized part of a long rule's regexp that a flat constant cannot see. New regression `TestLoadIgnoreSetChargesALongRulesSource` drives a single ~256 KiB rule line through `loadIgnoreSet` with a 64 KiB budget and asserts the refusal (before this, the flat charge let it through).

**Medium — zero-rule files free: documented tradeoff, per the issue.** #1018 says explicitly: "The per-file overhead can then go, since the line charge subsumes it." So I have not reintroduced a per-file charge. The `maxGlobIgnoreTotalBytes` comment now states that it bounds compiled-matcher memory (rules + their source) and that the number of files a walk can read is bounded separately by the listing budget. Note the source term also means a comment/blank-only rules file is no longer free — only a literally 0-byte file is — so the residual gap is 0-byte files plus the dedupe map, both bounded by `maxGlobDirListings`. Flagging this as the one place where I followed the issue over the review; happy to add a small per-file charge if you'd rather take the stricter bound.

**Low — mirrored skip logic: cross-checked.** `TestCompiledIgnoreRuleCountMirrorsGoGitignoreSkips` now asserts each case twice: against the expected count and against the length of `go-gitignore`'s own retained `patterns` slice (read via reflection, since the library exposes no accessor), so a library change that alters comment/blank handling fails the test instead of silently desynchronizing the charge.

obra:
Re-review addressed at 90bd3eb0f.

**High — zero-rule `.gitignore` files retained matcher objects at no charge: fixed.** `readIgnoreFile` now reports "nothing to keep" (nil lines) for a file that compiles zero rules, exactly as it already did for a missing one, so neither caller appends an `ignoreDir`/`*gitignore.GitIgnore` for it. Such a matcher matches nothing, so dropping it changes no answer; it only stops the call pinning an object per rules file that the per-rule charge records as zero. `TestLoadIgnoreSetDoesNotRetainZeroRuleMatchers` loads 30 empty/comment-only files and asserts nothing was retained; before the fix it failed with `retained matchers for 30 of 30 zero-rule .gitignore files, want 0`. The `loadedRules` dedupe entry remains (it is a path key, not a matcher), and is bounded by the listing budget.

**Low — brittle reflection test: hardened.** `compiledIgnorePatternCount` now guards with `IsValid()`/kind and `t.Fatal`s with a clear message instead of panicking if the library's field changes.

**Low — `compiledIgnoreRuleCount` over-counts lines that fail to compile: documented and pinned.** The count is deliberately an upper bound: a line whose generated regexp fails to compile is counted but yields no pattern. The comment now says so, the test is `TestCompiledIgnoreRuleCountBoundsGoGitignorePatterns`, and it includes an uncompilable pattern (`[` → 0 patterns) asserting the count never under-counts what the library retained. The mirror is still exact for everything that compiles.

obra:
Re-review addressed at 909c580fa (rebased onto latest main, which also brings in the `gateSettlements` web fix, so `web`/`static-build` should be green again).

**High — `loadedRules` retains paths for zero-rule files: fixed.** The dedupe map now records a path only after `readIgnoreFile` returns non-nil lines, i.e. only for a `.gitignore` that contributed a matcher the budget charged for. Every remembered path therefore corresponds to a charged matcher, so the map's size stays inside `maxGlobIgnoreTotalBytes`; a rule-free (or missing/unreadable) file is left out and re-read by a later scope, costing only the read the per-file cap already bounds. `TestLoadIgnoreSetDoesNotCacheZeroRulePaths` uses two overlapping scopes and an Open-counting FS: before the fix the rule-free base `.gitignore` was opened once (negatively cached) and the test failed with `opened 1 times across two scopes, want 2`; after the fix it is opened twice while the file that contributed a rule is still opened once.

**Low — comments described history/provenance: rewritten.** The comments at `search_patterns.go` (the budget constants and `retainCompiledRules`/`ruleBytes`) and `gitignore.go` (`readIgnoreFile`) now state the current invariants and rationale only; issue numbers, prior-behavior narration, and review provenance are gone (they live in the commit message).

Test evidence (all with `-race`): `go test ./agent/execenv/ -race -count=1` ok (59.4s); `gofmt -l` clean; `go vet ./agent/execenv/` exit 0.

obra:
Medium regression addressed at b0b4ed838 (rebased onto latest main, behind 0).

The two findings are **not** in tension. Skipping the matcher for a rule-free `.gitignore` is intended and stays: its matcher matches nothing, so retaining it only pins an `ignoreDir` + matcher. What was wrong was that the skip also returned *before charging*, which dropped the aggregate source-read bound along with the matcher.

Fix: `readIgnoreFile` now charges `retainCompiledRules(rules, len(data))` for every successfully read file, then returns nil lines when `rules == 0`. So a rule-free file contributes `0*globIgnoreRuleBytes + len(data)` and no matcher — the read bound is restored without re-pinning anything.

Evidence (`-race`): `go test ./agent/execenv/ -race -count=1` ok. Two regression tests fail before the change and pass after:
- `TestLoadIgnoreSetChargesRuleFreeFilesAgainstTheReadBudget` — 20 comment-only 8 KiB files against a 32 KiB budget; before: no refusal. After: `budgetRulesTotal` refusal with no matcher retained.
- `TestLoadIgnoreSetDoesNotRetainZeroRuleMatchers` — before: `ruleBytes = 0 after reading 240 bytes of zero-rule files, want 240`. After: charged, still `len(set.dirs) == 0`.

`TestLoadIgnoreSetDoesNotCacheZeroRulePaths` still pins that the dedupe map holds only rule-contributing files (a rule-free file is re-read and re-charged by a later scope).

obra:
Findings from the d070896 review addressed at 03c31c320 (rebased onto latest main, behind 0).

**Medium — budget checked only after `strings.Split` materialises every line: fixed at the root.** `compiledIgnoreRuleCount` now scans the raw bytes for line breaks and is charged before any split; a file that compiles no rule is never split at all, and one that would be refused is refused before its lines exist. New `TestLoadIgnoreSetAppliesTheBudgetBeforeSplittingLines` loads a 1 MiB newline-heavy `.gitignore` and asserts bytes allocated stay under 8 MiB — on the old split-first order it failed with `allocated 20096344 bytes` for the same input.

**Low — rule-free re-reads/re-charges (comment understated, untested): addressed both ways.** The `loadedRules` comment now states that a rule-free file is re-read and re-charged by each scope and that enough overlapping scopes can exhaust the aggregate budget (the safe direction), and `TestLoadIgnoreSetChargesRuleFreeRereadsAgainstTheBudget` pins it: four overlapping scopes of an 8 KiB comment-only `.gitignore` against a 16 KiB budget refuse on the third read, with the open count showing the re-reads are the cause. I did not add a per-entry charge for zero-rule paths — that would reintroduce the per-file overhead the issue removes — so the accumulation is documented rather than removed.

**Low — refusal text:** widened to `spent N bytes on .gitignore source and compiled matchers across the call`, since `e.count` covers both.

**Low — test comment vs arithmetic:** `TestLoadIgnoreSetRefusesTooManyRetainedRules` now uses `24 * globIgnoreRuleBytes`, so two 8-rule files fit (rules + source) and the third crosses, matching its comment; it asserts exactly two matchers were retained.

Evidence: `gofmt` clean; `go vet ./agent/execenv/` exit 0; `golangci-lint run ./agent/execenv/...` → 0 issues; `go test ./agent/execenv/ -race -count=1` ok (28.2s).

obra:
Review on d070896 addressed at b9b4b1616 (rebased onto latest main, behind 0). One design note up front: this reintroduces a small per-file charge, which #1018 said could go — but two rounds of review have shown the per-rule charge does not subsume it, so each file now pays a path charge and the budget bounds the dedupe map too.

**High — long-rule compiled-regexp memory undercounted: fixed, measurement-backed.** I measured go-gitignore directly: one literal rule retains 46.5 bytes per source byte at 1 MiB (48.7 MB total), 48.5 at 256 KiB, 50.7 at 16 KiB, 76.1 at 1 KiB. The charge is now `rules*globIgnoreRuleBytes + ruleSourceBytes*globIgnoreRuleSourceBytes` (factor 80, above every measured ratio) `+ sourceBytes + globIgnoreFileOverheadBytes`. New `TestLoadIgnoreSetRefusesLongRulesWhoseSourceFitsTheBudget`: three files of one 256 KiB literal rule each, 768 KiB of source under a 2 MiB budget, refused; before the factor it returned no error.

**Medium — rule-free re-reads can refuse benign data: fixed.** Every read `.gitignore` path is remembered again (charged `globIgnoreFileOverheadBytes`), so a rule-free file is read and charged once per call, not once per scope. `TestLoadIgnoreSetReadsZeroRulePathsOnceAcrossScopes` uses four overlapping scopes and asserts the base file is opened exactly once; the previous re-read test is replaced by this one.

**Low — non-deterministic MemStats assertion: replaced.** `splitIgnoreLines` is now a seam; `TestLoadIgnoreSetAppliesTheBudgetBeforeSplittingLines` asserts a rule-free and a refused file are each split zero times, and that the rule-free file is charged path + source, with no heap measurement.

**Low — reflection coupling: removed.** The count test no longer reads the library's unexported `patterns`; it asserts the counter's rules/rule-source against expected values and cross-checks observable `MatchesPath` behavior (the ordinary rule beside an uncompilable line still excludes its path).

Evidence: `gofmt` clean; `go vet ./agent/execenv/` exit 0; `golangci-lint run ./agent/execenv/...` → 0 issues; `go test ./agent/execenv/ -race -count=1` ok (12.8s).

obra:
Roborev's b9b4b16 Mediums addressed at b1f078cdc (rebased, behind 0).

**Medium — missing/unreadable paths recorded in `loadedRules` without charge: fixed.** `readIgnoreFile` now returns `(lines, read, err)`; a path is remembered and charged only when it was actually read. A missing or unreadable file is neither cached nor charged, and a later scope merely retries the open (one failed open, no allocation). `TestLoadIgnoreSetDoesNotCacheUnreadablePaths` asserts the missing base `.gitignore` is opened twice across two scopes and `ruleBytes == 0`; with the old unconditional mark it failed with `opened 1 times ... want 2`.

**Medium — blank/comment-heavy rule file materialized by `strings.Split`: fixed.** The split is gone. `eachIgnoreLine` walks the bytes and yields only the lines go-gitignore will compile, so `compiledIgnoreRules` counts without materializing anything and `ignoreRuleLines` collects only candidate lines — after the budget has accepted the file. A file that is refused or compiles no rule never has any line built. `TestLoadIgnoreSetMaterializesLinesOnlyAfterTheBudget` pins all three: zero collections for a rule-free near-cap file, zero for a refused file, and exactly one collection yielding one matcher for a blank-heavy file with one rule.

Evidence: `gofmt` clean; `go vet ./agent/execenv/` exit 0; `golangci-lint run ./agent/execenv/...` → 0 issues; `go test ./agent/execenv/ -race -count=1` ok (9.4s).

obra:
Roborev's b1f078c findings addressed at 2cb388fb4 (rebased, behind 0).

**High — 80 bytes/source-byte is not a sound bound under regexp expansion: fixed, measurement-backed.** I measured the amplification directly. Go expands `{n}` into n copies of the element and refuses n > 1000 (`a{1000}` compiles, `a{1001}` does not, and nested repeats are rejected), so a rule's compiled program is bounded by 1000 × element instructions. Measured ratios: `.{1000}` 5951 bytes/source byte, `.{999}`×1000 7248, 26-way alternation `{1000}` 2096 — while a 7 KB line of `.{999}`×1000 retained 43.5 MB. Source inside a counted repetition is now charged at `globIgnoreRepeatSourceBytes = 65536` (the 1000 cap × ~48 bytes/instruction, above the ~40000/byte worst case) instead of 80. New `TestLoadIgnoreSetRefusesRepetitionRulesWhoseSourceFitsTheBudget`: a 6000-byte repetition rule under a 1 MiB budget is refused; with the repeat charge zeroed it returned no error.

**Medium — 512-byte per-file charge ignores the retained path length: fixed.** `ignoreFileCost.path` carries `len(path)` and the charge includes it, so 65,000 long paths can no longer stay inside the budget. Accounting tests updated; path bytes are computed from the fixture paths.

Note on soundness: this is an over-approximation, not an exact model. It relies on Go's repetition cap (verified: a{1001} does not compile) and ~48 bytes per compiled instruction (syntax.Inst measured ~40); it charges any candidate line containing `{` at the repeat rate, which is conservative for the brace patterns go-gitignore cannot compile anyway.

Evidence: `gofmt` clean; `go vet ./agent/execenv/` exit 0; `golangci-lint run ./agent/execenv/...` → 0 issues; `go test ./agent/execenv/ -race -count=1` ok (15.3s).

obra:
Roborev's 2cb388f findings addressed/refuted at 0f18ab477 (rebased, behind 0).

**Medium — any `{` charged as repetition, over-billing ordinary brace rules: fixed (all four reviewers were right).** `hasCountedRepetition` now classifies a line as repetition only for a syntactically valid quantifier (`{n}`, `{n,}`, `{n,m}` with a leading digit), skipping escaped characters and character classes. Measured with go-gitignore: `{cache}` (2024 bytes), `foo{` (2008) and `*.{js,map}` (2096) all compile *small* literal-brace matchers, so they are now charged as ordinary source. New `TestLoadIgnoreSetKeepsBraceRulesUsable`: 100 `*.{js,map}` lines under a 1 MiB budget load cleanly; with the old any-brace classifier the call refused, spending 65,793,622 bytes.

**High — nested counted repetitions evade the budget: refuted by measurement.** Go rejects the cited example: `(.{1000}.{1000}){1000}` (22 bytes) retains 40 bytes — no pattern at all — and `(a{50}){50}` likewise. Go's parser refuses the nested products that would multiply. Small nested repeats do compile (`(a{31}){32}` → 66,256 bytes from 11 bytes, ratio ~6023), and the whole-line repeat charge (11 × 65536 = 720 KB) covers that by ~10×, so the charge remains an upper bound. There is nothing to refuse for the cited case because no matcher is compiled for it.

**Low — star-dense rules under-count: fixed.** Confirmed: a 400 KiB line of `*/` retains 74.5 MB (~380 bytes per star) while the per-source-byte charge records ~32 MB. Added `globIgnoreStarBytes = 512` per `*`. New `TestLoadIgnoreSetRefusesStarDenseRules`: a 400 KiB star-dense rule is refused; with the star charge zeroed it returned no error.

**Low — non-ASCII whitespace lines miscounted: refuted against the pinned library.** `go-gitignore`'s `getPatternFromLine` trims only ASCII spaces (`strings.Trim(line, " ")`), not Unicode whitespace, so a tab-only line is *not* dropped — it compiles a matcher (measured 2096 bytes retained, and `MatchesPath` on a tab path matches). Counting and charging it as a rule is therefore correct. `bytes.Trim(line, " ")` in our classifier mirrors the library exactly; a spaces-only line is dropped by both (measured 24 bytes retained).

Evidence: `gofmt` clean; `go vet ./agent/execenv/` exit 0; `golangci-lint run ./agent/execenv/...` → 0 issues; `go test ./agent/execenv/ -race -count=1` ok (10.1s).

obra:
Roborev's 0f18ab4 findings addressed/refuted at b98a88122 (rebased, behind 0). Taking the note's steer, the charge is now driven by what each line's constructs actually expand into rather than by a parallel length estimate:

- **High — nested counted repetitions:** refuted by measurement. `(.{1000}){1000}` (15 B) retains **5,472 bytes**, not ~1M instructions; `(a{1000}){1000}`, `(.{500}){500}` and `(a{50}){50}` are rejected by Go outright (24 B). The compilable nested maximum I found is `(a{31}){32}` → 66,256 bytes from 11 B, which the whole-line repeat charge (11 × 65536 = 720 KB) already covers ~10×. There is nothing to refuse for the cited pattern because no matcher is compiled for it.
- **High — int overflow on 32-bit:** fixed. `retainIgnoreFile` accumulates in `int64` and saturates via `clampIgnoreCost` before comparing with the budget and storing into `int`, so no term can wrap.
- **High — Unicode/POSIX property escapes:** fixed. Measured: `\pL` retains ~10 KB from 3 source bytes (and `\pL`×100 → 826 KB). Added `globIgnoreUnicodeClassBytes = 32768` per `\p`/`\P` escape; `[[:alpha:]]` measured 13 B/source-byte and stays on the ordinary rate. `TestLoadIgnoreSetRefusesUnicodePropertyRules` fails with the charge zeroed.
- **Medium — star not scaled by repetition:** fixed. `globIgnoreRepeatStarBytes = 262144` (5 insts × 1000 cap × ~48 B) for a star inside a repetition; measured `(*×100){1000}` retains 17.8 MB (~177 KB/star) while the plain charge records ~7 MB. `TestLoadIgnoreSetRefusesStarsInsideRepetition` fails with the charge zeroed.
- **Medium — literal stars overcharged:** fixed. `classifyIgnoreLine` reads the line once, skipping escaped bytes and character classes, so `\*` and `[*]` are not stars. `TestLoadIgnoreSetDoesNotOverchargeLiteralStars` plus table cases pin it.
- **Medium — whitespace mismatch:** refuted against the pinned library. `getPatternFromLine` uses `strings.Trim(line, " ")` — ASCII spaces only, not `TrimSpace` — so a tab-only line is not dropped; it compiles a matcher (measured 2,096 bytes retained, and it matches a tab path). Counting it is correct; `bytes.Trim(line, " ")` mirrors the library exactly.

New tests: refuses-unicode-property, refuses-stars-inside-repetition, does-not-overcharge-literal-stars, retains-saturates-instead-of-wrapping, plus table cases for escaped/class stars and property escapes.

Evidence: `gofmt` clean; `go vet ./agent/execenv/` exit 0; `make lint-golangci` → `PASS lint-golangci (173s)`; `go test ./agent/execenv/ -race -count=1` ok (13.0s); `roborev list --open` run.

obra:
Roborev's b98a881 findings fixed at f662db558 (rebased, behind 0), first the shared code path the High and the third Medium share:

- **High — zero-minimum repetition undercount:** fixed. `countedRepetitionExpands` now parses both bounds and treats a brace as expanding whenever the effective maximum is above zero. Measured: `a{0,1000}` retains 96,144 bytes from 9 source bytes while `a{0}` (2,024) and `a{0,0}` (2,104) expand to nothing. `TestLoadIgnoreSetRefusesZeroMinimumRepetition` (100 `a{0,1000}` rules vs a 1 MiB budget) and table cases for `{0,1000}`, `{01}`, `{0}`, `{0,0}` pin it; all fail on the first-digit-only classifier.
- **Medium — over-broad repetition charge:** fixed. The repetition rate now applies to the repeated element's span, not the whole line: `classifyIgnoreLine` records each repeated element's span, charges its source at the repeat rate and its stars at the repeat-star rate, and leaves the rest of the line and its stars on the ordinary rates. Table case `abcdefghij{2}` asserts `ruleSource == 13, repeatSource == 1`.
- **Medium — 32-bit test compile failure:** fixed. The saturation test now sizes its input from the architecture (`int(^uint(0)>>1)/globIgnoreRepeatSourceBytes + 1`) and asserts both refusal and accumulator saturation. `retainIgnoreFile` also computes every term with saturating int64 `satAdd`/`satMul`, so no product can wrap, not just the final sum.

Evidence: `gofmt` clean; `go vet ./agent/execenv/` exit 0; `make lint-golangci` → `PASS lint-golangci (117s)`; `go test ./agent/execenv/ -race -count=1` ok (11.6s). Roborev backlog for this PR closed (`roborev list --open` now shows only other lanes' entries).

obra:
Roborev's f662db5 findings addressed/refuted at 7c4e2948c (rebased, behind 0).

**Medium — quadratic `classifyIgnoreLine`/`inSpan`: fixed.** Spans are sorted and merged once (nested repetitions merge into their union), then star membership is a single forward pointer instead of a scan of every span per star, making the pass linear. A `maxIgnoreLineSpans` cap (1024) stops parsing a quantifier-dense line and charges the whole line at the repeat rate, which is the conservative direction. `TestLoadIgnoreSetRefusesQuantifierDenseLine` uses a 200,000-span line and asserts a refusal.

**Medium — repeated source double-charged: fixed.** `classifyIgnoreLine` now reports ordinary and repeated source separately (`ordinarySrc` / `repeatSource`), and `compiledIgnoreRules` charges each byte at exactly one rate. Table case `abcdefghij{2}` asserts `ruleSource == 12, repeatSource == 1`.

**Low — dead `ignoreLineExpansion.repeat` field: removed.**

**High — nested counted repetitions "charged additively, not multiplicatively": refuted by measurement.** `(a{1000}){1000}` (15 bytes) retains **24 bytes** — Go rejects it, as it does `(a{1000}){1000}`, `(.{500}){500}` and `(a{50}){50}`. The largest nested form I could make compile is `(a{31}){32}` → 66,256 bytes from 11 bytes, whose outer element span plus inner span the span-based charge covers by ~10×. Since no matcher is compiled for the cited pattern, there is nothing to preserve or refuse.

**Low — `bytes.TrimRight` strips all trailing `\r`: refuted against the pinned library.** `getPatternFromLine` uses `line = strings.TrimRight(line, "\r")` (ignore.go:75), which also strips every trailing carriage return, so our `bytes.TrimRight(line, "\r")` matches it exactly. There is no one-CR-only behavior to mirror.

Evidence: `gofmt` clean; `go vet ./agent/execenv/` exit 0; `make lint-golangci` → `PASS lint-golangci (116s)`; `go test ./agent/execenv/ -race -count=1` ok (12.9s); this PR's roborev backlog closed.

obra:
Roborev's 7c4e2948c findings fixed at a0ba6d317 (rebased, behind 0), all through one change rather than four patches.

**Root change.** The per-construct rate heuristic is replaced by a single forward parse of each rule line that computes the compiled instruction units it expands into, and that unit count is what the budget charges (`globIgnoreUnitBytes` per unit). A literal byte is one unit, a glob star five, a counted repetition multiplies its element's units by `max+(max-min)`, and because a group's units are accumulated before its own quantifier is seen, nesting composes multiplicatively with saturation.

- **High — nested repetitions:** fixed by that multiplication. `(a{900}){900}` records 810,000 units from one source byte and is refused at the default budget; new `TestLoadIgnoreSetRefusesNestedQuantifiers`, and unit-count cases for `(a{2}){2}` and `(a{900}){900}`.
- **High — span ignores escapes/classes:** fixed structurally. The span is now produced by the same forward pass that already tracks `inClass` and backslash escapes, so no backward scan exists at all; a literal `(` inside `[(]` cannot be taken for a group start. New `TestLoadIgnoreSetRefusesRepeatedGroupWithLiteralParen`.
- **Medium — optional branches:** fixed by the `max+(max-min)` factor, so `a{0,1000}` records 2000 units. New `TestLoadIgnoreSetRefusesOptionalBranchRepetition` (300 rules at the default budget); previously the 1000-only factor let them pass.
- **Medium — quadratic backward scan:** fixed by removing the backward scan; a malformed line is treated as literal bytes. New `TestLoadIgnoreSetRefusesMalformedQuantifierLine` (40,000 `){1000}` fragments) and the existing quantifier-dense test both complete immediately.

Evidence: `gofmt` clean; `go vet ./agent/execenv/` exit 0; `make lint-golangci` → `PASS lint-golangci (115s)`; `go test ./agent/execenv/ -race -count=1` ok (14.8s). Roborev backlog for this PR closed.

obra:
Roborev's 7b0b0b7 findings fixed at a92bfc62d (rebased, behind 0).

**High — leading-zero bounds counted as repetitions: fixed (semantics corrected; no bypass reproduced).** Measured against the pinned library: `a{01}` matches the literal `a{01}` and not `a`, so Go treats a bound with a leading zero as literal. `countedRepetitionFactor` now rejects such bounds (only a bare `0` is valid), with table cases for `{01}` and `{00,1000}` and a behavioral assertion. I could not reproduce a memory bypass from this one: a literal expands at ~48 bytes/source byte, which the 80-bytes-per-source-byte charge already covers (`a` + 5000 zeros + `1}` retained 240 KB against ~400 KB charged), but the classification is now correct.

**High — literal first `]` in a character class: fixed, bypass confirmed.** `(` + 1000×`a` + `[])]){1000}` retained **43,376,256 bytes** from 1012 source bytes while the old scan charged ~160 KB. `ignoreClassEnd` now treats a `]` in first position (and right after `^`) as a class member, so the group's close is found correctly. New `TestLoadIgnoreSetRefusesClassStartingWithLiteralBracket` asserts the refusal.

Evidence: `gofmt` clean; `go vet ./agent/execenv/` exit 0; `make lint-golangci` → `PASS lint-golangci (128s)`; `go test ./agent/execenv/ -race -count=1` ok (12.0s). Roborev backlog for this PR closed.

obra:
Roborev's a92bfc6 findings addressed/refuted at a27bacbbc5 (rebased, behind 0).

**High — Unicode tables not multiplied by repetition: fixed at the root.** The table cost now rides the atom as instruction units (`globUnicodeUnits = ceil(globIgnoreUnicodeClassBytes / globIgnoreUnitBytes)`), and `\p`/`\P` is consumed as one atom including its property name (`\pL`, `\p{Greek}`), so a quantifier attaches to the escape and multiplies the table with everything else. `\pL{1000}` records 411,000 units and `[\pL]{1000}` 415,000. New `TestLoadIgnoreSetRefusesRepeatedUnicodeProperties` (4 `\pL{1000}` rules) fails when the table charge is zeroed.

**Medium — `?` charged 1 unit instead of `(.)`: refuted from the pinned dependency.** `getPatternFromLine` does `line = strings.Replace(line, "?", `\?`, -1)` (ignore.go:128), i.e. it escapes `?` to a literal, so a `?` compiles to one instruction, not three. No question-mark charge is warranted; a question-dense file is already charged one unit per `?` plus its source.

**Low — stale `maxIgnoreLineSpans` wording: fixed.** The comment now says the pass is linear because each scan consumes the region it reads, and that the line is refused because its quantifiers expand past the budget.

**Low — `flush` shadowing: fixed.** The local index is now `dest`.

Evidence: `gofmt` clean; `go vet ./agent/execenv/` exit 0; `make lint-golangci` → `PASS lint-golangci (140s)`; `go test ./agent/execenv/ -race -count=1` ok (15.2s). Roborev backlog for this PR closed.

PR #1767, opened 2026-09-18
refactor(hub): unify the fork fence predicate and ownership-source resolvers (#1146)

### LINKED ISSUE #1146: hub fork: unify the fork fence predicates and the ownership-source resolvers
## Where
`cmd/evener-hub/app_threadlifecycle.go` (`forkTargetSessionID`, `forkTargetSessionIDUnderLock`, `forkClaimIsLiveOwner`, `forkFenceTargets`) and `cmd/evener-hub/app_threadread.go` (`applyHubForkCapability`, `hubForkResolvedSessionFenced`, `hubForkDeletionFenced`). Landed across PR #1105 review rounds 9–15.

## What
The fork path now encodes one rule ("which session does this thread ref resolve to, and is it safe to branch") in several places that were kept in sync by review rather than by structure:

- Two resolvers (pre-lock, advisory, swallows errors; under-lock, authoritative, returns errors) that since round 15 fall back to each other one level deep when only one ownership source (roster or run dir) is configured. Termination is guaranteed by complementary guards on `cfg.RunDir`/`cfg.Roster`, not by shape.
- The projection-side fence (`hubForkResolvedSessionFenced`) duplicates the alias-check shape instead of using one per-identity predicate shared with the RPC fence over `forkFenceTargets`.
- `forkClaimIsLiveOwner` probes a process under both per-session mutexes for every claim.
- The roster-only fallback skips `resumeClaimTarget`'s duplicate-claim settlement (`liveDaemonForThread` returns the first non-crashed match), and the run-dir-only fallback runs the rendezvous scan and its probes on every projected thread. Both shapes are fixture-only today: `cmd/evener-hub/main.go` defaults an empty run dir before building the roster, so production always has both sources.

## Why it matters
Rounds 9–15 each found one more asymmetry between two encodings of the same rule. The structure invites the next one.

## Direction
One source-deciding resolver (read rendezvous, read roster, decide once), with the pre-lock and under-lock callers as thin wrappers that re-establish their error contracts; one per-identity fence predicate used by both the projection and the RPC. Do it as a single restructure of the production path, with the existing fork fence tests untouched as the contract.

---

obra:
Fixes #1146

## Root cause

The fork path encoded one rule — "which session does this thread ref resolve to, and is it safe to branch" — in several places kept in sync by review:

- `forkTargetSessionID` (pre-lock, roster-authoritative, swallows errors) and `forkTargetSessionIDUnderLock` (under-lock, rendezvous-authoritative, returns errors) **fell back to each other one level deep** when only one ownership source was configured. Termination came from complementary guards on `cfg.RunDir`/`cfg.Roster`, not from the shape.
- `hubForkResolvedSessionFenced` (projection) re-encoded the alias-check shape instead of sharing a per-identity predicate with the RPC's fence over `forkFenceTargets`.

## Change

- **One source-deciding resolver.** `resolveForkTarget(cfg, threadID, authoritative)` reads the trusted source once and answers; when that source is not configured the other answers, else the shared redirect tail. `forkTargetSessionID` and `forkTargetSessionIDUnderLock` are now thin wrappers that only re-establish their error contracts (swallow-and-redirect vs. propagate). The rendezvous and roster readers are leaves, so neither fallback recurses.
- **One per-identity fence predicate.** `hubForkIdentityFenced(cfg, threadID, owner)` encodes live delegate + roster-reported recovery status + hub recovery locks. It is used by `hubForkRecoveryFencedNow` and `hubForkResolvedSessionFenced` (projection) and by the fork RPC's per-target pass, replacing the projection's separate re-encoding and the RPC's direct `hubForkLiveStatusFenced`/`hubForkLiveDelegateFenced` pair.

The existing fork fence tests in `cmd/evener-hub/app_fork_capabilities_test.go` are untouched and are the contract.

## Note on a unified refusal

Folding the live-delegate signal into the shared predicate means a fork of a live delegate now returns the same recovery refusal (`session recovery requires an explicit thread/resume…`) that the other per-identity signals use, instead of a delegate-specific message. No test asserts the message (verified repo-wide); every test asserts the structured `unavailable` code. Flagging it because it is the one externally visible difference.

## Test evidence

```
$ go test ./cmd/evener-hub/ -run 'Fork' -count=1
ok  	primeradiant.com/evener/cmd/evener-hub	12.332s

$ go test ./cmd/evener-hub/ -count=1
ok  	primeradiant.com/evener/cmd/evener-hub	399.790s

$ gofmt -l cmd/evener-hub/app_threadlifecycle.go cmd/evener-hub/app_threadread.go
(no output)

$ go vet ./cmd/evener-hub/
(no output)
```


obra:
Addressed the roborev review in 340056a (rebased onto main):

- **Medium (delegate refusal)**: `hubForkIdentityFenced` now covers only the recovery signals (roster-reported status + hub recovery locks). The live-delegate signal keeps its own shared predicate `hubForkLiveDelegateFenced`, so a live-delegate fork returns its distinct `unavailable` again instead of a recovery refusal an explicit `thread/resume` cannot clear — and it is no longer reclassified as `sessionRecoveryAdmissionError`. Added an assertion in `TestHubForkFencesLiveDelegateFromOneSignal` that the delegate refusal is not a recovery refusal.
- **Low (dead predicate)**: removed `hubForkLiveStatusFenced` and pointed `TestHubForkLiveStatusFenceAgreesOnBothIdentities` at `hubForkIdentityFenced`, so the agreement it pins is the production predicate's.

obra:
Roborev re-review resolutions (head 0ec0ce3bf, rebased onto main):

- **Medium — gofmt violation in `app_fork_capabilities_test.go`**: fixed by `gofmt -w` (whitespace only) in e03fadfb5. Verified locally: `make lint-gofmt` and `golangci-lint run --config .golangci.yml ./cmd/evener-hub/...` (the `gofmt` formatter is enabled in `.golangci.yml`, which is why `lint-golangci` failed too) both report clean.
- **Low — comment overclaimed "changes no answer"**: narrowed in 3f7effe86. The comment now says the roster-status conjunct costs a `Find` for the resolved id and adds no refusal there, and that the recovery locks are keyed by id so that conjunct stays per-identity and is deliberately stricter over both identities.

PR #1770, opened 2026-09-18
fix(hub): echo originClientId on the instance mutation broadcast (#1517)

### LINKED ISSUE #1517: hub: evener/instance/updated echoes no originClientId, so a client's own instance mutations come back as foreign refetches

obra:
Fixes #1517.

## Root cause

`notifyInstanceUpdated` (`cmd/evener-hub/app_rpc.go`) broadcast `evener/auth/updated` with empty params after a provider-instance CRUD mutation, while `notifyAuthUpdated` echoes `originClientId`. The credential store's own-echo correlation (`appwire-client/typescript/state/credentials/instances.ts`) matched only by provider, and an instance broadcast names no provider — so the client that issued the mutation read its own echo as another client's change and refetched *after* it had already applied the full-list answer its RPC returned. On the web that read also tripped `ProviderConnection`'s probe invalidation.

## Change

- **Hub:** `notifyInstanceUpdated(server, originClientId)` now echoes the mutation's `originClientId` (and still emits no provider/activeSource — no single pair honestly summarizes "the list changed"). All six instance handlers pass `params.OriginClientId`; the server-initiated live-prefetch pass passes `""`.
- **Wire types:** `InstanceCreate/Edit/Remove/SetDefault/SetModelDisabled/RefreshModelsParams` gain optional `originClientId` (`appwire/types.go`), regenerated into `types.gen.ts` and `docs/appwire-protocol.md` by `make generate`.
- **Core:** the own-echo marker becomes one marker **per mutation** instead of one per provider, so a provider-less instance echo matches its marker by `originClientId`; the store's own echo schedules a self-marked refresh instead of a foreign one. Every instance mutation stamps `originClientId` and arms/re-stamps/retires a marker.

The marker's subject is the provider an auth write targeted, or nothing for an instance mutation (whose echo names no provider) — no instance-specific branch on the shared mechanism.

## Test evidence

Test-first: the new Go test failed on the old source (`originClientId=""`), then passed after the hub change. The new store test failed on the old source (`selfRefresh` stayed 0: its own echo was foreign), then passed.

- `go test ./cmd/evener-hub/... ./appwire/... -count=1` → **ok** (all packages)
- `gofmt -l` on changed Go files → clean; `go vet ./cmd/evener-hub/... ./appwire/...` → exit 0
- `npx tsc --noEmit --incremental false` (frontend, includes the shared package) → clean
- `npx biome ci` on the six changed TS files → clean
- `npx vitest run ../../../appwire-client/typescript/state/credentials/instances.test.ts src/stores/credentials.test.ts src/panes/settings/sections/credentials/{InstanceSheet,instanceDialogs,CredentialsSection}.test.tsx` → **343 passed**
- broader `npx vitest run src/stores/credentials.test.ts src/panes/settings/sections/credentials/ src/panes/spawn/` → **32 files, 1042 passed**
- native `npx vitest run src/providerInstances.test.ts` → 19 passed; `tsc --project tsconfig.check.json` → clean

The web/native test updates add `originClientId: "test-tab"` to the exact-params assertions for instance mutations (a deliberately changed wire contract); no assertion was weakened or deleted.

## Simplify

`simplify-code` applied (four angles run in one pass — this harness has no subagent tool). No code edits resulted: the change generalizes the existing marker mechanism rather than adding a special case, reuses `EvenerAuthUpdatedParams`/`BroadcastAll` and the existing hub test harness, and adds no new I/O. One skipped finding: stale markers are not globally pruned, so a client whose echoes are repeatedly lost could accumulate markers; pruning would change attribution behavior (an id-bearing late echo currently matches regardless of age) and so was left out.


obra:
RoboRev review (`441993e`) addressed in `f3ad4881f` — 0 Critical / 0 High confirmed as sound; the 2 Medium and 2 Low findings are fixed:

**Medium — provider-less echo misclassification.** A notification carrying no provider *and* no `originClientId` no longer matches an outstanding instance marker by timing; it names no provider to correlate on (any client's instance write, or the hub's own live-prefetch pass), so it stays foreign. The timing fallback now applies only to provider-bearing auth notifications. Regression test: "a provider-less notification with no id stays foreign, not spent on an outstanding mutation".

**Medium — marker retirement/restamp not mutation-specific.** `noteLocalMutation` now returns the marker it armed, and `restampLocalMutation`/`retireLocalMutation` act on that exact marker rather than "the latest marker for the subject", so an older mutation's late reply or error cannot spend a newer mutation's marker. An arriving echo consumes the **oldest** outstanding marker for its subject (echoes arrive in issue order on one connection). Regression test: "an early echo spends the issuing mutation's marker, so a later refusal cannot steal a newer one" (early echo for A, then A rejected, then B's own echo still self-marked).

**Low — outdated protocol summary.** `appwire/protocol.go`'s `evener/auth/updated` entry now reads "after a successful auth mutation or provider-instance CRUD/live-model change. Clients refresh auth state and the instance list."; `docs/appwire-protocol.md` regenerated via `make generate`.

**Low — incomplete hub echo coverage.** `TestHubRPCInstanceBroadcastEchoesOriginClientId` is now table-driven over `create`/`edit`/`remove`/`setDefault`/`setModelDisabled` (fresh hub per case), plus `TestHubRPCInstanceRefreshModelsBroadcastEchoesOriginClientId` covering the sixth handler against the existing hermetic `httptest` `/models` gateway. The no-id control test is kept.

Re-verified after the fixes:
- `go test ./cmd/evener-hub/ -run TestHubRPCInstance -count=1` → ok
- `go test ./appwire/...` → ok; `gofmt -l` clean; `go vet ./cmd/evener-hub/... ./appwire/...` → exit 0
- frontend `tsc --noEmit` clean; `biome ci` clean
- `vitest …/credentials/instances.test.ts` → 29 passed
- `vitest --maxWorkers=2 src/stores/credentials.test.ts src/panes/settings/sections/credentials/ src/panes/spawn/` → 32 files / 1042 passed
- native `providerInstances.test.ts` → 19 passed; native `tsc --project tsconfig.check.json` clean

obra:
**`static-build` failure — root cause and fix (`965183d6f`).**

The failing check was not caused by this PR. It came from a broken `main`: `cmd/evener-hub/frontend/src/panes/session/composer/RepoLocation.test.tsx` still imported `gateRequests` from `@evener/appwire-client/testing/fakeClient`, but commit `b86ef963b` (#1755, "one gateSettlements helper for the appwire-client test gates") replaced that export with `gateSettlements`. `tsc` therefore failed with:

```
src/panes/session/composer/RepoLocation.test.tsx(4,22): error TS2305:
Module '"@evener/appwire-client/testing/fakeClient"' has no exported member 'gateRequests'.
```

Because this PR's merge ref takes main's copy of that file, the failure appeared here too.

Fix: merged current `main` (`4fbaeb551`, clean, no conflicts) and ported the single call site to the current API in `965183d6f` — `gateSettlements(client, "evener/git/head")` returns `Settlement[]`, so `answers[i]?.({...})` became `answers[i]?.resolve({...})`.

Verified: `npx tsc --noEmit --incremental false` clean; `vitest run src/panes/session/composer/RepoLocation.test.tsx` → 20 passed; post-merge `vitest --maxWorkers=2` over the credential store + credentials section + composer suites → 40 files / 1196 passed; `go test ./cmd/evener-hub/ -run TestHubRPCInstance` and `go test ./appwire/...` ok; `gofmt -l` clean.

obra:
RoboRev's remaining Medium (reviewed `f3ad488`) is fixed in `2f26605a5`.

**Finding — `consumeOwnEcho` stops at a stale marker.** It looked only at the oldest marker for a subject; finding it stale, it dropped it and returned "foreign" without checking the newer live markers behind it. A lost echo for one mutation could therefore make a later same-provider mutation's id-less auth echo read as another client's change and invalidate the operation the correlation exists to protect.

**Fix.** `firstLiveMarker` drops stale markers as it walks to the oldest live one, and the id-less fallback uses it. The id-bearing path is unchanged: it matches by identity alone, where the caller's own id proves ownership regardless of age.

**Test.** Added the requested overlapping-marker test, `"a stale marker does not shadow a live one for the same provider"`: it arms a marker, advances past the 2000 ms echo window with no echo, arms a second same-provider marker, then emits an id-less auth echo and asserts the refresh is self-marked. It fails on the pre-fix code (`selfRefresh` stayed 0) and passes after.

Verified: `npx tsc --noEmit --incremental false` clean; `biome ci` clean; `vitest …/credentials/instances.test.ts` → 30 passed; `vitest --maxWorkers=2` over the store + credentials section → 16 files / 538 passed.

obra:
Rebased onto `origin/main` (`e692ca1e7`) and force-pushed with lease: head is now `8ff12aea7`.

The rebase was clean and git dropped my earlier `gateRequests`→`gateSettlements` commit as already applied, because main now carries `gateSettlements` itself. This branch no longer touches `RepoLocation.test.tsx` — the diff against main is 13 files, all part of #1517.

Commits (all on top of `e692ca1e7`):
- `175453a0d` fix(hub): echo originClientId on the instance mutation broadcast
- `3dc72e8ff` fix(sdk): make own-echo markers mutation-specific
- `8ff12aea7` fix(sdk): skip stale own-echo markers for an id-less echo

Local verification on the rebased tree:
- `npx tsc --noEmit --incremental false` (frontend) → exit 0 — the same typecheck `web`/`static-build` run
- `biome ci` on the six changed TS files → clean
- `gofmt -l` clean; `go vet ./cmd/evener-hub/... ./appwire/...` → exit 0
- `go test ./appwire/...` → ok; `go test ./cmd/evener-hub/ -run 'TestHubRPCInstance|TestAuthApiKeySetBroadcast'` → ok
- `vitest --maxWorkers=2` over credentials store + composer/RepoLocation + credentials section + instance dialogs → 17 files / 558 passed

obra:
RoboRev re-review (`8ff12ae`, state=current) — three of the four items addressed in `b8c034b6c`; the remaining Medium is answered below with a scope rationale.

**Medium 2 — provider-less instance markers bypass stale-marker cleanup → FIXED.** Correct: the id-less path deliberately treats a provider-less, id-less notification as foreign, so nothing ever spent an instance marker on a hub that ignores `originClientId`, stranding one marker per successful instance mutation. `noteLocalMutation` now drops markers already outside `SELF_ECHO_WINDOW_MS` when it arms a new one, which bounds the set while leaving what an in-window echo consumes unchanged. Regression test: "a stale instance marker is pruned, not stranded, so it cannot absorb a later echo" — RED without the fix (the second own-id echo was wrongly self-marked, `selfRefresh` 4 vs 2), green after.

**Low 1 — instance broadcast shape unpinned → FIXED.** Added `assertInstanceBroadcastShape`, applied to every instance broadcast test (the table-driven five, `refreshModels`, and the no-id control), asserting both the echoed `originClientId` and that `provider`/`activeSource` are empty — the SDK's correlation keys on that absence. `TestHubRPCInstanceEditRenameBroadcastsWhenTheCredentialMoveFails` now sends `OriginClientId` and asserts the echoed origin on the rename-persisted error branch.

**Low 2 — stale comment → FIXED.** `EvenerAuthUpdatedParams.OriginClientId` now states that the provider-plus-timing fallback applies to provider-bearing auth echoes only; an id-less provider-instance broadcast is a foreign change.

**Medium 1 — per-mutation correlation id — out of scope for #1517.** This is a real observation, but fixing it is a wire-protocol change, not a correction to this PR's mechanism:

- It requires a new field carrying a unique per-mutation identifier on *every* mutation request (all six `evener/instance/*` params and the auth params) plus the `evener/auth/updated` payload, so every client (web, native, TUI) must mint and thread an identifier and the hub must echo it. That is a protocol addition with version-negotiation consequences, whereas #1517 is scoped to echoing the *existing* `originClientId` (which identifies the client, not the mutation) and correlating its own echo.
- The residual it addresses is exactly the bounded, documented one: a delayed echo from an earlier mutation still triggers a listing refetch, so only the guided flow's self-refresh marking is skipped, and only inside the echo window. RoboRev's previous review traced the same class of race and called it "benign … matches pre-existing `authMutation` behavior with no demonstrated downstream harm"; the same ordering exists on the auth path this PR did not change.
- Implementing it here would also make every mutation's params carry a nondeterministic value, forcing broad exact-param assertion churn across the web and native suites — scope well beyond this issue.

Happy to file a follow-up issue for a per-mutation correlation id (with the protocol field and version note) if the maintainers want the residual removed; it should land as its own change rather than inside #1517.

Verified after `b8c034b6c`: frontend `tsc --noEmit --incremental false` exit 0; `biome ci` clean; `gofmt -l` clean; `go vet` exit 0; `go test ./appwire/...` ok; `go test ./cmd/evener-hub/ -run 'TestHubRPCInstance|TestAuthApiKeySetBroadcast'` ok; `vitest --maxWorkers=2` over the credentials store + credentials section + composer/RepoLocation → 17 files / 559 passed.

obra:
Rebased onto `origin/main` (`edd42c5b4`) and force-pushed with lease: head is now `e736368eb0e2651b6830a4fbe6749713d73d85fd`. This clears the conflict and starts a fresh CI + roborev pass.

**Conflicts and how they were resolved (both sides' intent kept):**

1. **`cmd/evener-hub/app_rpc.go`** — main's #1834 introduced a shared `instanceWrite(apply func() error)` closure (using `writeDidApply`, which absorbed the old `renamePersistedError` special case) and routed all six instance handlers through it; this branch threaded `originClientId` through each handler. Kept main's structure and gave the closure an `originClientId` parameter:
   ```go
   instanceWrite := func(originClientId string, apply func() error) (appwire.InstanceListResponse, error) {
       err := apply()
       if writeDidApply(err) {
           notifyInstanceUpdated(server, originClientId)
       }
       ...
   ```
   All six handlers call `instanceWrite(params.OriginClientId, …)`. main's `notifyAuthWrite` no-data branch (write applied, status read failed) now passes `originClientID` to `notifyInstanceUpdated`, and the function's doc comment reconciles both roles (instance CRUD broadcast + credential no-data form).

2. **`appwire-client/typescript/state/credentials/instances.ts`** — main's change in this file was comments-only. Took this branch's side for the conflict and folded main's point into the final marker-lifecycle comment: a write that applied before a later step failed still broadcasts, but the failing mutation has already retired its own marker, so that echo reads as foreign and re-reads.

3. **`appwire-client/typescript/types.gen.ts`** — auto-merged, but main's #1771 makes `appwirets` emit Go field doc comments into the generated TS. Regenerated with `make generate` (+52/−8), so the new `OriginClientId` fields' comments are emitted like every other field's.

**Verified on the rebased tree (`e736368eb`):** `go build ./...` ok; `go vet ./cmd/evener-hub/... ./appwire/...` exit 0; `go test ./appwire/...` ok; focused `go test ./cmd/evener-hub/ -run 'TestHubRPCInstance|TestAuthApiKeySetBroadcast|TestAuthWrite|Instance'` ok; `gofmt -l` clean; frontend `tsc --noEmit --incremental false` exit 0; `biome ci` clean; `vitest --maxWorkers=2` over the credentials store + credentials section + composer/RepoLocation → 17 files / 559 passed.

On roborev: understood that "No review output generated" is an absent review, not a clean result. I'll report the new review's body verbatim once it posts; if it again says that instead of "No issues found.", I'll record it as a review-infrastructure failure rather than as clean.

PR #1771, opened 2026-09-18
fix(appwirets): emit Go field doc comments into generated TS types

### LINKED ISSUE #1637: appwirets: generated TS types carry no field doc comments (omitzero removal signal invisible to clients)
Found by the /simplify pass over #1614.

`internal/appwirets/emit.go` does not propagate Go doc comments into the generated TypeScript interfaces, so `cmd/evener-hub/frontend/src/protocol/types.gen.ts` emits a bare `outputImages?: OutputImage[];` with no trace of the rule #1614 documents on the Go type: `OutputImages` is `omitzero`, an explicit `[]` is the removal signal, while `Images` is `omitempty` and empty equals absent. The downstream consumers of that rule (the web reducer in `appwire-client/typescript/reducer.ts` and the native state layer) read the generated TS type first, not the Go source.

Fix: emit the Go field doc comment as a `/** ... */` JSDoc block above each generated field (types.gen.ts must stay byte-identical for fields without a comment, and `make generate` must stay deterministic). Pre-existing codegen gap, not introduced by #1614.

obra:
## Root cause

`internal/appwirets/emit.go` builds every generated TypeScript interface purely by reflecting over the catalog's Go types. `reflect.Type` carries no doc comments, so a field's Go doc comment never reached `types.gen.ts`. The rule #1614 documents on `appwire.ThreadItem.OutputImages` — it is `omitzero`, so an explicit `[]` is the removal signal, while `Images` is `omitempty` and empty equals absent — was therefore invisible to the web reducer and native state layer, which read the generated TS first.

## Change

- `internal/appwirets/emit.go` now parses each emitted type's package source with `go/parser` (lazily, once per package, keyed by import path) and threads the field's doc comment through `rawFieldsOf` into `emitInterface`.
- Each documented field emits an indented `/** ... */` JSDoc block directly above it. Lines that would close the block are escaped.
- A field with no doc comment emits nothing, so the file stays byte-identical outside the inserted blocks: `appwire-client/typescript/types.gen.ts` diff is **800 insertions, 0 deletions** (118 blocks).
- Regenerated `appwire-client/typescript/types.gen.ts`.

## Test evidence

- Red first: `go test ./internal/appwirets/ -run TestEmitCatalogCarriesFieldDocComments` failed with `outputImages?: OutputImage[];` emitted with no JSDoc.
- `go test ./internal/appwirets/...` — ok (includes the `TestGeneratedFileCurrent` drift gate).
- `gofmt -l internal/appwirets/emit.go internal/appwirets/emit_test.go` — clean.
- `go vet ./internal/appwirets/...` — clean.
- Determinism: generator run twice to separate files and diffed — identical; also identical to the committed file.
- Syntax check of the regenerated TS with the hub frontend's `tsc --noEmit --skipLibCheck --target ES2020` — exit 0 (same as HEAD version).

New tests: `TestEmitCatalogCarriesFieldDocComments` (motivating `OutputImages` case + undocumented `images` gains no block), `TestWriteFieldDoc` (rendering, escaping, empty), `TestCommentText` (`//` and `/* */` forms).

Fixes #1637


obra:
Addressed the low-severity trimpath finding from the combined review.

Confirmed real first: `GOFLAGS=-trimpath go test ./internal/appwirets/ -run TestEmitCatalogCarriesFieldDocComments` failed (all JSDoc stripped).

Fix (commit f3157066b): `packageDir` now resolves the module root by walking up from the working directory to the nearest `go.mod` declaring \`primeradiant.com/evener\` (so the nested \`agent/\` module on the way up is skipped), and only falls back to \`runtime.Caller\` when the recorded path is absolute. Added regression tests \`TestFindModuleRootFromWorkingDirectory\` and \`TestDeclaresModuleRequiresExactModulePath\`.

Evidence: `go test ./internal/appwirets/...` passes normally and under `GOFLAGS=-trimpath`; generation under `-trimpath` is byte-identical to the committed file.

obra:
Fixed the `lint-golangci` failure (commit 0f9c3cf22).

Findings — `modernize` / `stringsseq` (3), all in `internal/appwirets/emit.go`:
- `commentText` `for _, line := range strings.Split(text, "\n")`
- `declaresModule` `for _, line := range strings.Split(string(goMod), "\n")`
- `writeFieldDoc` `for _, line := range strings.Split(doc, "\n")`

Fix: range over `strings.SplitSeq` instead of allocating a slice, matching the repo convention (e.g. `emit_test.go`).

Evidence: `golangci-lint run --config .golangci.yml ./internal/appwirets/...` → `0 issues.`; `go test ./internal/appwirets/...` passes normally and under `GOFLAGS=-trimpath`; regenerated `types.gen.ts` byte-identical (no behavior change).

obra:
Rebased onto main and repaired a main-side break that was failing `static-build`, `static`, and `web` on this PR (and on every open PR):

- Root cause: #1755 (`b86ef963b`) replaced the `gateRequests` test helper with `gateSettlements` in `appwire-client/typescript/testing/fakeClient.ts`, while #1754 (`61ebe90b1`), merged concurrently, added `RepoLocation.test.tsx` importing and calling `gateRequests`. Main tip `daf43f6b1` carries both halves, so `tsc` reports `Module "…/testing/fakeClient" has no exported member "gateRequests"` and the RepoLocation test throws `gateRequests is not a function`.
- Fix (commit 0dba8ca26): migrate the single RepoLocation call site to `gateSettlements` (`settlements[i]?.resolve(...)`); assertions unchanged.
- Verified: `npx vitest run src/panes/session/composer/RepoLocation.test.tsx` → 20 passed; `npx tsc --noEmit --incremental false` → exit 0.

This is unrelated to #1637 but my PR could not go green while main carried it.

PR #1774, opened 2026-09-18
refactor(sdk): drop the unread loaded flag from CommandCatalogState

### LINKED ISSUE #1479: sdk: drop the unread `loaded` flag from CommandCatalogState after D1
Residual from #1478 (SDK row D1). `CommandCatalogState.loaded` in `appwire-client/typescript/commandCatalog.ts` has no production reader in either app: `commands`/`loading`/`error` already express every state a consumer distinguishes. It was kept in D1 because the web store's public surface is the adapter's contract (plan rule 4) and ~30 test `setState({ commands, loaded })` sites across `shell/palette/*.test.ts(x)`, `Composer.test.tsx` and `stores/commandCatalog.test.ts` name it.

Follow-up: remove the field from the package state, the web adapter test's three asserts, and the `loaded:` keys at those test sites (excess-property errors otherwise). No behaviour change. Also note from the D1 review: D2 should lift the connection-following client port in `cmd/evener-hub/frontend/src/stores/commandCatalog.ts` to `stores/connectionClient.ts` when its second consumer arrives, and D3 should lift the private `refreshLoop` into its own package module and retrofit `mobile-native/src/taskList.ts` (ActivityList waits for D17).

obra:
Fixes #1479. Residual from #1478 (SDK row D1).

## Root cause

`CommandCatalogState.loaded` was a field in the package's public store state with no reader in either app. The one internal reader was `createCommandCatalog`'s `watch()` gate: "re-read a loaded catalog on `evener/plugin/updated`". Because it lived in state, the web adapter's contract surface (`useCommandCatalog`, the adapter's public API per plan rule 4) had to carry it, and ~30 test `setState({ commands, loaded })` sites plus three adapter asserts named it.

## Change

- Removed `loaded` from `CommandCatalogState`, from both factory initializers, and from the read's returned partial in `appwire-client/typescript/commandCatalog.ts`.
- The `watch()` gate now derives the same condition from state it already holds: a catalog has arrived once the published `commands` list is no longer the initial array. A failed or superseded read leaves `commands` untouched (so an unread catalog stays unread), and `setState(getInitialState())` restores the initial reference (so the store stays resettable with the rest of its state). This is the key constraint: an earlier closure-variable approach satisfied the package tests but broke the adapter test's documented reset.
- Removed the field's name from the package test (`commandCatalog.test.ts`), the web adapter test's three asserts (`stores/commandCatalog.test.ts`), and the `loaded:` keys at the palette/composer `setState` sites (29 sites across `shell/palette/commands.test.ts`, `shell/palette/CommandPalette.test.tsx`, `shell/palette/CommandPalette.edge.test.tsx`, `panes/session/composer/Composer.test.tsx`) — excess-property errors otherwise.

No behaviour change. Notably, `keybindings*` stores have their own unrelated `loaded` field and were left untouched.

## Test evidence

From `cmd/evener-hub/frontend` (node_modules symlinked to the shared install; the app's vitest config includes the package's own tests):

- `npx vitest run --maxWorkers=4 src/stores/commandCatalog.test.ts ../../../appwire-client/typescript/commandCatalog.test.ts src/shell/palette/CommandPalette.test.tsx src/shell/palette/CommandPalette.edge.test.tsx src/shell/palette/commands.test.ts src/panes/session/composer/Composer.test.tsx`
  → **Test Files 6 passed (6), Tests 336 passed (336)**
- `npx tsc --noEmit --incremental false` → exit 0
- `npx biome ci <changed files>` → clean


obra:
Addressed the roborev review in `aea7978d0`:

**Medium (reset semantics)** — the `beforeEach` hooks in `shell/palette/commands.test.ts`, `shell/palette/CommandPalette.test.tsx`, `shell/palette/CommandPalette.edge.test.tsx` and `panes/session/composer/Composer.test.tsx` wrote `useCommandCatalog.setState({ commands: [] })`, allocating a fresh array that is `!==` the store's initial one, so the state-derived arrived-catalog gate read those resets as *loaded*. They now reset through the documented `useCommandCatalog.setState(useCommandCatalog.getInitialState())` (the same form `stores/commandCatalog.test.ts` already uses), restoring the initial `commands` reference so an unread catalog stays unread.

**Low (empty-catalog pin)** — added a package test that the first successful read returning an authoritative-empty catalog still counts as loaded: a following `evener/plugin/updated` must trigger a second `evener/command/list`. This fails if the gate is ever weakened to a length/emptiness check.

Evidence: `npx vitest run` on the 6 touched files -> **337 passed**; `npx tsc --noEmit --incremental false` -> exit 0; `npx biome ci` on the changed files -> clean.

PR #1776, opened 2026-09-18
feat(lint): add make lint-biome using the frontend-pinned Biome (#1406)

### LINKED ISSUE #1406: lint: 'npx biome' from the repo root resolves to an unrelated biome@0.3.3 and passes silently; the package scope needs a real make target
Found by the C20 implementer (PR #1403): running `npx biome ci appwire-client/typescript` from the repository root does not run the Biome linter the frontend pins; npx resolves an unrelated package named `biome` (0.3.3) that exits 0. Every brief this week that asked for "biome both scopes" from the root therefore checked nothing for the package scope; only the frontend-directory invocation (`npm run lint` in cmd/evener-hub/frontend, or `npx biome ci ../../../appwire-client/typescript` from there) is real.

Wanted: a make target (e.g. `make lint-biome`, part of `make lint`) that runs the pinned Biome over cmd/evener-hub/frontend/src AND appwire-client/typescript from the frontend's node_modules, and a line in AGENTS.md saying never to invoke `npx biome` from the root. Then the two scopes are one gate and cannot be faked.

obra:
Fixes #1406

## Root cause

The pinned `@biomejs/biome` is installed only under
`cmd/evener-hub/frontend/node_modules`. `npx biome` from the repository root
therefore finds no local package and downloads the unrelated `biome@0.3.3`,
which ignores its arguments and exits 0. Any "biome both scopes" invocation run
from the root reported green while checking nothing; only the frontend-directory
invocation was real.

## Change

- `make/linting.mk`: new `lint-biome` target (part of `LINT_TARGETS`, so part of
  `make lint`) that runs the frontend install's own Biome by path:

  `cd cmd/evener-hub/frontend && ./node_modules/.bin/biome ci src ../../../appwire-client/typescript`

  It depends on the existing `web-preflight`, so the install-readiness rule is
  shared with the other web targets. Naming the binary by path makes the pinned
  tool unavoidable and covers both scopes in one gate.
- `makefiletargets_audit_test.go`: added `lint-biome` to `lintGateCommands`
  (the existing audit requires every `LINT_TARGETS` member to be pinned to a
  distinguishing command) and added `TestLintBiomeRunsTheFrontendPinnedBiome`,
  which fails if the recipe ever shells out to `npx`, stops naming the
  frontend-local binary, or drops either scope.
- `docs/developing-evener/linting.md`: regenerated target table (`make
  generate`) plus a short section explaining the root-`npx` hazard; the
  `LINT_TARGETS` list in the prose now includes `lint-biome`.
- `AGENTS.md`: the frontend-gates guidance no longer tells agents to run
  `npx biome` from the root; it now points at `make lint-biome` (or Biome run
  from the frontend directory) and states the prohibition explicitly.

## Test evidence

- `go test . -run 'TestLintBiomeRunsTheFrontendPinnedBiome|TestMakeLintRunsEveryGateInLintTargets|TestEveryLintingRuleIsInLintTargets|TestEveryLintTargetIsPhonyAndHasARule|TestEveryTargetHasASummaryAnnotation|TestExactlyOneLintTargetsDefinition' -count=1` → `ok`
- The new test failed before the fix:
  `make/linting.mk has no lint-biome rule, so make lint cannot run the pinned Biome ...`
- `make lint-biome` → `PASS lint-biome (11s)`; with a deliberate `debugger;` in
  `cmd/evener-hub/frontend/src/` it checked 1355 files and exited 1 (`FAIL
  lint-biome`), proving the gate has teeth; probe removed afterwards.
- `make lint-generated` → `PASS` (generated docs fresh).
- `gofmt -l makefiletargets_audit_test.go` → clean; `go vet .` → clean.

## Notes

CI's `web` job already runs the frontend `lint` script (same pinned Biome, same
two scopes) via `make test-web`; this target makes `make lint` exercise the same
gate from the repository root instead of the silently-passing `npx biome`.


obra:
Thanks — addressed the findings in 728b0043c (amended commit, force-pushed).

**Low — bare `src` substring was too loose:** the scope-substring assertions are gone. `TestLintBiomeRunsTheFrontendLintScript` now asserts the recipe runs from `cmd/evener-hub/frontend` and delegates to `npm run lint`, with no `src`/path substring matching.

**Low — recipe duplicated the scope list:** fixed by delegating. `lint-biome` now runs `cd cmd/evener-hub/frontend && npm run lint`, so `cmd/evener-hub/frontend/package.json`'s `lint` script is the single source of truth for both scopes and cannot drift from the command the `web` CI job runs.

**Low — trigger claimed "Required CI (via make lint)" though CI never runs `make lint`:** the trigger cell now reads "Required CI (the web job's make test-web runs the same frontend lint script); local make lint", which is what actually enforces it. The `LINT_TARGETS` paragraph in `docs/developing-evener/linting.md` was reworded to match.

**Medium — `web-preflight` symlink lockfile check can be skipped:** the target no longer claims a version guarantee web-preflight does not provide. Delegating to the frontend's own `lint` script run from the frontend directory means `biome` resolves through that install's `node_modules/.bin` via npm's `PATH`, so the unrelated root-resolved `biome@0.3.3` — the defect in #1406 — is impossible regardless of the symlink path; the docs/annotations now state exactly that. I deliberately did not reorder `scripts/web/web-preflight.sh`: the mtime-before-symlink shortcut is pre-existing shared behavior that affects `build-web`/`test-web`/`build-hub` too (already documented as a caveat in AGENTS.md), so changing it is a separate change with a wider blast radius than #1406. Happy to file that follow-up (reorder the `-L node_modules` comparison ahead of the `-nt` shortcut, plus a regression test for a newer, mismatched shared install) if you want it in scope.

Verification after the change: `go test . -run 'TestLintBiome|TestMakeLintRuns|TestEveryLinting|TestEveryLintTarget|TestEveryTargetHasASummary|TestExactlyOneLintTargets|TestEveryGeneratedRegion|TestMakeLintGenerated' -count=1` → ok; `make lint-biome` → PASS; `make lint-generated` → PASS; `gofmt -l` and `go vet .` clean.

obra:
Fixed at the root in 2eaccf876.

**Finding (Medium):** `scripts/web/web-preflight.sh` checked `node_modules -nt package-lock.json` before `[ -L node_modules ]`. Because `-nt` follows symlinks, a shared install whose target mtime was newer than this worktree's lockfile short-circuited the lockfile-content comparison and proceeded on a mismatched install — so `lint-biome`'s claim that `web-preflight` readies a pinned install was not guaranteed.

**Resolution:** the symlink branch now runs FIRST. Every symlinked `node_modules` is validated against the lockfile beside its target before the `-nt` freshness shortcut can apply; a real (non-symlinked) install keeps the timestamp shortcut. Header comment and the `## Frontend gates` paragraph in AGENTS.md were updated to describe the actual (now unconditional) behavior.

**Evidence:**
- New regression subtest `TestMakeRuntimeAliasesBuildThePair/build-web/refuses-npm-ci-through-a-newer-symlink` constructs a shared install newer than the worktree lockfile with its own differing `package-lock.json`, and requires `build-web` to refuse with a symlink-naming message without running `npm ci`.
- Direct script check on the same fixture: fixed script exits 1 with "node_modules is a symlink to … does not match this worktree's"; the pre-fix script (from HEAD at the time) skipped the comparison and only failed later at the `tsc` health check — i.e. the mismatch was not caught.
- `go test . -run 'TestMakeRuntimeAliasesBuildThePair|TestWebPreflightBootstrapsMissingFrontendDependencies|TestLintBiome|TestMakeLintRuns|TestEveryLinting|TestEveryLintTarget|TestEveryTargetHasASummary|TestExactlyOneLintTargets|TestEveryGeneratedRegion|TestMakeLintGenerated' -count=1` → ok
- `make lint-biome` PASS, `make lint-generated` PASS, `gofmt -l` clean, `go vet .` clean.

The branch was also rebased onto latest `main` (head `2eaccf876`), which clears the unrelated `gateRequests`/`gateSettlements` breakage.

PR #1778, opened 2026-09-18
refactor(appwire-client): type applyNotification generically over the model (#1523)

### LINKED ISSUE #1523: appwire-client: type reducer.applyNotification generically over the model so wrappers with extra fields need no cast
Found by the /simplify altitude pass on D23b (#1522). Native's `MobileConversation = ThreadModel & { items }` goes through `applyNotification`, whose every case spreads `...model`, so extra fields survive at runtime but the return type is `ThreadModel` and the wrapper casts (or worse, re-spreads `items`, an extra whole-model copy per frame, which #1522's simplify round removes in favour of a cast). General fix: `applyNotification<M extends ThreadModel>(model: M, n, now): M` (and the helpers it composes, where they spread the model). Package change, so a separate small PR after the D23 slices; the web wrapper needs no cast today and is unaffected.

obra:
Fixes #1523

## Root cause

`applyNotification` was typed `(model: ThreadModel, …) => ThreadModel`, but every
case in `applyNotificationToThread` and its composed helpers builds its result by
spreading `model` (`{ ...model, … }`). A caller whose model is `ThreadModel` plus
its own fields — native's `MobileConversation = ThreadModel & { items }` — had
those fields survive at runtime yet lost them in the type, so the wrapper had to
cast the result back (`mobile/src/state/conversation.ts:867`).

## Change

Make the reducer generic over the model it folds:

- `applyNotification<M extends ThreadModel>(model: M, n, now): M`
- `applyNotificationToThread<M extends ThreadModel>(model: M, n, now): M`
- `foldNonActiveTurnCompleted<M extends ThreadModel>(model: M, …): M`
- `resolvePendingEscalation<M extends ThreadModel>(model: M, id): M`

`{ ...model, … }` is `M & {…}`, which is assignable to `M`, so every fold case
type-checks with no cast at the call site and no internal assertion.

The `modelRetry`-clearing branch previously used a rest-destructure
(`const { modelRetry, ...cleared } = next`), whose `Omit<M, "modelRetry">` result
is *not* assignable to `M`; it now spreads and `delete`s the key, keeping the
exact prior runtime shape (the key is absent, not present-and-`undefined`).

Package-only change: the public signature widens compatibly (a `ThreadModel`
argument still infers `M = ThreadModel`), so web is unaffected, as the issue notes.

## Test evidence

New test in `appwire-client/typescript/reducer.test.ts` annotates its result as
`WrappedModel = ThreadModel & { wrapperMarker: number }`. That annotation is the
compile-time half: the frontend's `tsc` program includes the package's test files.

- `cd cmd/evener-hub/frontend && ./node_modules/.bin/vitest run ../../../appwire-client/typescript/reducer.test.ts` → **197 passed** (covers the constructing, scalar-patch, same-ref no-op and modelRetry-clearing fold shapes, asserting `wrapperMarker` survives and the cleared key is absent).
- Scoped `tsc --noEmit` over `reducer.test.ts` + `reducer.ts` (frontend `tsconfig.json`): **exit 0**. Falsification: reverting just the signature to the old `ThreadModel` form makes it fail with 5× `TS2322: Type 'ThreadModel' is not assignable to type 'WrappedModel'`.
- `appwire-client/typescript: tsc -p tsconfig.build.json --noEmit` → **exit 0**.
- `biome check` on both files → clean, no fixes applied.

No Go changes.


obra:
Both roborev findings are fixed at the new head `d34ebb6c0`.

**Medium (`reducer.ts:1170`) — narrowing soundness.** Fixed at root cause. The fold now returns `ThreadModel & Omit<M, keyof ThreadModel>` (aliased `ModelExtras<M>`), and `applyNotification`, `applyNotificationToThread`, `foldNonActiveTurnCompleted` and `resolvePendingEscalation` share that return. ThreadModel-owned fields (which the fold clears/restamps/replaces, e.g. `modelRetry`, `lastFrameAt`, `status`) keep ThreadModel's types; only the caller's own extra fields are carried through from `M`. No assertion is needed anywhere — the generic spread already produces a value assignable to that type. A new type-level test pins the soundness: with a model narrowed to require `modelRetry`, the result's `modelRetry` is `ThreadModel["modelRetry"]` (possibly `undefined`), enforced by a `@ts-expect-error` under the frontend's `tsc` program (which includes the package's test files).

**Low (`conversation.ts:867`) — redundant cast.** Fixed: removed `as MobileConversation`; `applyNotification(conversation, …)` now returns `MobileConversation` directly.

Verification at `d34ebb6c0` (rebased on latest `main`):
- `vitest run appwire-client/typescript/reducer.test.ts` → 198 passed
- `tsc -p appwire-client/typescript/tsconfig.build.json --noEmit` → 0
- scoped `tsc --noEmit` over `reducer.test.ts` + `reducer.ts` (frontend `tsconfig.json`) → 0 (the `@ts-expect-error` is exercised)
- `tsc --noEmit -p mobile-native/tsconfig.check.json` → 0
- full frontend `tsc --noEmit` → 0; `biome check` (frontend config) on the package files → clean

PR #1781, opened 2026-09-18
fix(agent): cap activity envelope inputs at projection time (#1202)

### LINKED ISSUE #1202: agent: cap activity envelope inputs (session label, delegate task/mandate, per-record branch errors) at projection time
Follow-up from #1179 round 3 (part of #1116).

The activity page envelope is normally tiny (222 bytes bare; ~722KB, 17% of `activityMaxEncodedBytes`, in a realistic worst case with a 33-hop delegate chain and 8KB labels), but three inputs are unbounded: `Label` is the session's `OriginalPrompt` verbatim when it has no name (a pasted 4MB prompt reproduces it), delegate `Task`/`Mandate`/`Description`, and `Branch.Error`, which appends one message per unsupported-type job record. When the envelope alone exceeds the limit, #1179's `markActivityEnvelopeTooLarge` now reports the response's own size and withdraws the continuation instead of blaming an entry — but the response still goes out over the limit, because nothing can be dropped to shrink it.

The real fix is at projection time: cap the label (and say it was cut), cap delegate prose fields, and collapse repeated per-record branch errors into one counted message. Add a test that a 4MB `OriginalPrompt` produces a page under the limit with the label truncated.

obra:
## Root cause

The activity page envelope (limit `activityMaxEncodedBytes`, 4MiB) is built from three inputs the projection copied through unbounded:

- the session `Label`, which is the session's `OriginalPrompt` verbatim when it has no generated name — a pasted multi-megabyte prompt reproduces it;
- a delegate's `Task`/`Mandate`/`Description`;
- `Branch.Error`, which appended one sentence per unsupported-type job record.

These are the response's *fixed* parts: a label sits on the session and a continuation page carries its ancestor chain's delegate metadata no matter how entries are trimmed. When one of them alone exceeded the limit, #1179's `markActivityEnvelopeTooLarge` correctly withdrew the continuation and reported the size, but the response still went out over the limit because nothing remained to drop.

## Change

Cap all three at projection time (`agent/jobs_activity.go`):

- `truncateActivityText` caps a string at a rune budget, appending an ellipsis when it cuts (rune-safe). The label is capped at `activityMaxLabelRunes` (200), mirroring the hub's own sidebar title cap (`hubcore.maxTitleRunes`) for exactly the same value.
- Delegate `Mandate`/`Task`/`Description` are capped at `activityMaxDelegateProseRunes` (4096).
- `unsupportedActivityJobTypesError` collapses the per-record unsupported-type messages into one counted message naming the first offender; a single offender keeps the original wording.

## Test evidence

New `agent/jobs_activity_caps_test.go`. Before the fix the reproduction fails on all three:

- `TestProjectBoundedActivityTree_CapsTheSessionLabel`: page = 4194589 bytes, over the 4194304-byte limit (4MiB `OriginalPrompt`); now the page fits and the label is cut.
- `TestProjectActivitySession_CapsDelegateProse`: Mandate = 12288 runes, uncapped.
- `TestProjectActivitySession_CollapsesUnsupportedTypeErrors`: 64 reports, one per record.
- `TestProjectActivitySession_SingleUnsupportedTypeKeepsOriginalWording`: the one-offender wording is unchanged.

After the fix:

```
go test ./agent/ -run 'TestProjectBoundedActivityTree_CapsTheSessionLabel|TestProjectActivitySession_CapsDelegateProse|TestProjectActivitySession_CollapsesUnsupportedTypeErrors|TestProjectActivitySession_SingleUnsupportedTypeKeepsOriginalWording|TestProjectActivitySession_UnsupportedJobType' -count=1
ok  	primeradiant.com/evener/agent	1.910s
```

`gofmt -l` clean; `go vet ./agent/` clean.

Fixes #1202


obra:
Addressed in dba2bf351:

**High — `truncateActivityText` panic on multi-byte text.** Confirmed and fixed. The byte-length fast path alone was wrong: a 150-rune `é` prompt is 300 bytes, so it fell through and sliced `runes[:199]` past `len(runes)`. The function now re-checks the rune count before slicing. New `TestTruncateActivityText_IsRuneSafe` covers both the over-bytes/under-runes case (used to panic) and a genuine rune-boundary cut.

**High — other fixed ancestor-delegate fields uncapped.** Bounded. `Message`/`StructuredResult` (raw JSON) are dropped when over 16 KiB — dropped rather than sliced so the wire never carries invalid JSON — with a diagnostic; `Reason`, `StructuredReason`, `NotResumableReason` are prose-capped; `Warnings` are capped in count and per-element length with a counted omission note. `TestProjectActivitySession_CapsDelegatePayloads` covers it.

**Medium — collapsed error copied the offender verbatim.** The offender ID and type are now capped through the same rune-safe truncation. `TestProjectActivitySession_BoundsTheCollapsedOffenderIdentifier` covers an oversized ID.

**Low — ASCII-only cap tests.** `TestTruncateActivityText_IsRuneSafe` adds the multi-byte path.

Full agent package (`go test ./agent/`) and `golangci-lint run ./agent/` re-run on this head.

obra:
All findings fixed at the root; head is now 1792ee39f (rebased onto current `main` to clear the unrelated `gateRequests` frontend failure that `web`/`static-build` hit on the old base).

**High — `truncateActivityText` multi-byte panic.** Fixed exactly as suggested: the byte fast path now falls through to a rune-count check before slicing, and `maxRunes <= 0` returns "". `TestTruncateActivityText_IsRuneSafe` covers 150×`é` (300 bytes / 150 runes — the case that used to panic) and a genuine rune-boundary cut.

**High — uncapped fixed ancestor-delegate fields.** Bounded: `Message`/`StructuredResult` are omitted above 16 KiB (dropped, never sliced, so the wire never carries invalid JSON) with a diagnostic; `Reason`, `StructuredReason`, `NotResumableReason` are prose-capped; `Warnings` are capped in count and per-element length with a counted omission note. `TestProjectBoundedActivityTree_BoundsPacketPayloadsOnContinuationAncestors` builds a 4-deep continuation ancestor chain of packet-bearing delegates and asserts the page fits and every ancestor is bounded; `TestProjectActivitySession_CapsDelegatePayloads` covers the single-delegate case.

**Medium — collapsed error copied the offender verbatim.** Offender ID and type are capped through the same rune-safe truncation; `TestProjectActivitySession_BoundsTheCollapsedOffenderIdentifier` covers an oversized ID.

**Low — ASCII-only cap tests.** The rune-safety and continuation-ancestor tests above add multi-byte and ancestor-chain coverage.

Focused evidence (`go test -race ./agent/ -run ...`): `ok primeradiant.com/evener/agent 10.150s`. `gofmt -l` and `go vet ./agent/` clean; full `go test ./agent/` re-running.

obra:
Addressed in 3a03bae8e, each with a test that fails before the fix (verified by stashing the production file):

1. **Uncapped origin/reason fields.** `OriginTurnID`, `OriginToolCallID`, `OriginItemID` are now capped like every other projection identifier, and `lastOutcome.Reason` like the other prose fields. `TestProjectBoundedActivityTree_CapsOutcomeReasonAndOriginIDs` gives the delegate a 1 MiB reason plus a renderable child and asserts the page fits and the child survives (before: the trim discarded the delegate/child to make room).

2. **`[]rune(s)` materialized the whole input.** `truncateActivityText` now walks runes with `for range` and stops at the cap, slicing at the byte boundary — no allocation proportional to the input. `Mandate`/`Task` also share one truncation now. `TestTruncateActivityText_DoesNotMaterializeTheInput` measures total allocation for capping a 4 MiB string at 200 runes and requires ≤64 KiB (before: ~16 MiB).

3. **Depth-unaware per-field budget.** Added `shrinkActivityAncestors`: when a page is over the limit, the projection first halves the ancestor chain's fixed prose, drops its packet payloads and trims its warnings, re-measuring the actual marshaled page (so control-rune JSON expansion counts) until it fits or reaches a floor. Only when that yields nothing do entries get trimmed. `TestProjectBoundedActivityTree_BoundsMaxDepthAncestorChain` builds a full `activityMaxContinuationPathLength`-deep chain of control-character prose and asserts the page fits with the whole chain retained (before: the trim severed the chain to make room).

Evidence: `go test -race ./agent/ -run '<cap tests>'` → ok 8.963s; `go test ./agent/` → ok 564.939s; `golangci-lint run ./agent/` → 0 issues; gofmt/vet clean.

PR #1340, opened 2026-09-15
fix(web): busy follows the thread status, not the transcript's open turn row (#1330)

obra:
Closes #1330. Closes #1341. Closes #1363.

## Which layer

The web client's derivation (`appwire-client/typescript/submitRouting.ts` `isTurnActive`), not the daemon and not the hub. The daemon's thread status is correct across the boundary and the hub relays it faithfully; the client folded a transcript bookkeeping frame into its "is the session working" answer.

## Measurement

- Projector (`internal/appprojector/appwire_projection.go` `openTurn`, called by `EventTurnStarted` / `EventUserInput` / `EventGoalContinuation`): `closeActiveTurn` emits `turn/completed(previous)`, then `turn/started(next)`, then the caller appends `thread/status/changed(active)`. No status frame between the two turn frames; idle is published only from `EventSessionEnd`, and there it follows the closing `turn/completed`.
- Hub (`cmd/evener-hub/internal/appsource/relay_session.go` `dispatchToListenerAtEpoch`, `app_relay.go` `publishTarget`): one `RelayDelivery` per notification, one `server.Broadcast` each, so one WebSocket message per frame.
- Client (`appwire-client/typescript/client.ts` `handleMessage`): one reducer step per message. `turn/completed` clears `activeTurnId`; `turn/started` sets it. `isTurnActive` was `status.type === "active" && !!activeTurnId`, so it read false for exactly one message at every inline turn boundary while the status still said active.
- Hub capability rule (`server/appwire_runtime.go` `appCapabilitiesLocked`): `active := status == active`; `Steer` and `Interrupt` derive from that alone. The daemon's plain `turn/steer` (`agent/session_client_mutation_queue.go` `clientMutationSteer`) has no active-turn precondition; the composer's Stop was already gated on the status alone for the same reason.

## RED tests

- `appwire-client/typescript/submitRouting.test.ts`: folds the daemon's exact three-frame boundary (`turn/completed(turn_1)`, `turn/started(turn_2)`, `thread/status/changed(active)`) through `applyNotification` from an active hydrated model and asserts busy after each frame. Before: `[{turn/completed, busy:false}, ...]`. After: busy at every step.
- `cmd/evener-hub/frontend/src/panes/session/composer/Composer.liveCapabilities.test.tsx`: renders the composer, emits the same three frames in separate `act`s, asserts Steer (and Stop) never leave the screen. Before: "Steer after turn/completed of the previous turn: expected null not to be null".

## Fix

[...]

PR #1782, opened 2026-09-18
fix(appwire): let the turn_failed status frame settle a failed turn (#1432)

### LINKED ISSUE #1432: web reducer and mobile store settle a failed turn idle themselves; the daemon's turn_failed status frame already does

obra:
Fixes #1432

## Root cause

All three clients were taught (rounds 18–20 of #1340, kata s8x8) that a genuine turn failure is the one turn that never gets a `thread/status/changed`. That premise is false at HEAD: `agent/session_lifecycle.go` `endInputAtTurnFailure` emits `EventSessionEnd{Reason:"turn_failed"}` (pinned by `TestSession_GenuineTurnFailureNotifiesLiveSubscriberOfIdleStatus`), and the projector announces it as `thread/status/changed(idle)` with the capability set inline.

The shared reducer therefore settled a failed turn idle twice: once on the non-active `failedWithoutId` branch and once on the active branch's `status` field. Both are redundant — the status frame that follows performs the same transition and also drops the work-clock anchor. Per the round-19 rule the status frame owns the status, so the early settles go.

## Change

- `appwire-client/typescript/reducer.ts` — drop the `failedWithoutId` self-settle in the non-active `turn/completed` branch and the `status: … { type: "idle" }` self-settle in the active branch. `turn/completed` still ends the turn (`activeTurnId`/`activeTurnStartedAt` for the active branch); the following `thread/status/changed(idle)` owns the status transition. The mobile store (`mobile/src/state/conversation.ts`) delegates to this reducer (`applyThreadNotification` → `applyNotification`), so it is covered by the same fix.
- Tests rewritten to assert the settle on the frame the daemon actually sends:
  - `appwire-client/typescript/reducer.test.ts` (3 tests)
  - `cmd/evener-hub/frontend/src/panes/session/composer/Composer.liveCapabilities.test.tsx` (1 test)
  - `mobile/src/state/conversation.test.ts` (2 tests)

Each rewritten test first asserts the failed `turn/completed` leaves the status (and, where relevant, the work-clock anchor) alone, then asserts the `thread/status/changed(idle)` frame performs the settle.

## Test evidence

Failing before the fix (exact reason: `expected 'idle' to be 'active'`), passing after:

- `cd cmd/evener-hub/frontend && npx vitest run ../../../appwire-client/typescript/reducer.test.ts` — red 3/196 → green 196/196
- `npx vitest run ../../../appwire-client/typescript` — 92 files, 1898 tests pass
- `npx vitest run src/panes/session/composer/Composer.liveCapabilities.test.tsx` — red 1/11 → green 11/11
- `npx vitest run src/stores/threads.test.ts src/panes/session/Session.test.tsx` — 398 tests pass
- `cd mobile-native && npx vitest run --root .. --config mobile-native/vitest.config.mts mobile/src` — 7 files, 771 tests pass
- `npx tsc --noEmit --incremental false` (frontend, covers `appwire-client/typescript`) — clean
- `npx tsc --noEmit --project tsconfig.check.json` (mobile-native, covers `mobile/src`) — clean


obra:
## Re: the medium finding on `reducer.ts:1285` / `app_relay.go:1719`

I investigated this against the code. The concern assumes the relay's synthesized `turn/completed` is applied by the web/mobile clients. It is not — before or after this PR — so this change does not regress that path.

**1. The synthesized frame carries no routing identity.** `giveUpOnActiveTurn` (`cmd/evener-hub/app_relay.go:1719`) broadcasts params containing only `turn`. Captured on the wire with the repo's own relay harness (`TestHubRelaySynthesizesConnectionFailureForActiveTurnAfterRepeatedRedialFailures` scenario):

```
METHOD=turn/completed
PARAMS={"turn":{"id":"turn_probe","itemsView":"","status":"failed","error":{"message":"Hub lost the connection to the session: ...","source":"hub"}}}
```

`server.Broadcast` → `appwire.NotificationMessage` does not stamp `threadId`/`ref`, and neither `internal/appserver` routing/delivery nor the AppWire TS client (`appwire-client/typescript/client.ts:781`) injects one.

**2. Without `ref`/`threadId` the reducer cannot apply it.** The shared `turn/completed` case returns the model unchanged unless `notificationTargetsThread` matches (`appwire-client/typescript/reducer.ts:915`, `:1218`), which requires `params.ref` or `params.threadId`. This is pinned as intentional in `cmd/evener-hub/frontend/src/stores/threads.ts:1761`: "both absent: notificationTargetsThread returns false for every model — the scan is a guaranteed no-op".

**3. The removed early settle never ran for this frame.** `failedWithoutId` sat *after* the same `notificationTargetsThread` guard, so it too was skipped for the unroutable hub frame. The stuck-`active` state the finding describes is not produced by this change. Recovery on that path is the relay's `evener/thread/resync` (`app_relay.go:1880`), which *does* carry `threadId`/`ref` via `stampResyncTarget` and drives the authoritative re-read.

**There is a real pre-existing gap here, but it is not caused by this PR and is out of scope for #1432** (which is limited to the web reducer + mobile store): the synthesized failure is unroutable, so the kata-3h02 danger chip it was built to raise never reaches the web/mobile reducer at all. Making it routable — and giving it the companion `thread/status/changed` frame the reducer now expects — is a hub-relay behavior change that belongs in its own issue, not in this reducer cleanup.

No code change is warranted here: the daemon path is verified, and the relay gap predates and is independent of this change.

obra:
## Relay finding fixed at the root (commit `950cbc052`)

Re-reading my earlier note: the synthesized failure was not just missing a status frame, it carried no `threadId`/`ref` at all, so the web/mobile reducer dropped it before either the old early settle or the new code could run. Rather than leave the producer violating the contract the reducer now relies on, this head fixes the relay path itself.

`giveUpOnActiveTurn` (`cmd/evener-hub/app_relay.go`) now broadcasts the daemon's real failure pair, both frames naming the relay's target that a client routes on:

1. `turn/completed{status:"failed", error.source:"hub"}` as `appwire.TurnCompletedParams` with `ThreadID`/`Ref` — the danger chip and "Reconnect & retry" (kata 3h02).
2. `thread/status/changed{status:{type:"idle"}}` with the hub's own capability set for a session whose daemon is gone (`pastThreadCapabilities()`, fork field resolved through `applyHubForkCapability`, exactly what a past read returns) — the frame that owns the session status, so Stop/Steer leave and Send returns.

This is the same pair the daemon's genuine failure exit emits (`agent/session_lifecycle.go` `endInputAtTurnFailure` → `EventSessionEnd{Reason:"turn_failed"}` → `thread/status/changed(idle)`).

**Test coverage** — `TestHubRelaySynthesizesConnectionFailureForActiveTurnAfterRepeatedRedialFailures` now asserts the pair is emitted exactly once per stall, that both frames carry the relay's `threadId`/`ref`, that the status is `idle`, and that the capability set has `send=true`. `expectRelaySynthesizedTurnFailure` gained the target assertions; `expectRelaySynthesizedIdleStatus` is new. `TestHubRelayNoSyntheticFailureWithoutActiveTurn` still pins that nothing is emitted between turns.

Evidence: `go test ./cmd/evener-hub/... -count=1` → ok (all packages); `go vet ./cmd/evener-hub/...` → clean; `gofmt -l` → clean.

obra:
## Local-only capabilities on the synthesized idle status (commit `822a877c6`)

Good catch — the frame was advertising the local past-session set unconditionally. Fixed:

`relayGaveUpCapabilities(cfg, relayKey, thread)` (`cmd/evener-hub/app_relay.go`) now returns the hub's past-session set only when the relay key is local — the same `strings.HasPrefix(relayKey, "local:")` gate `stampClosedThreadCapabilities` uses — and `nil` otherwise. `giveUpOnActiveTurn` passes that through, so a non-local/federated source's status frame omits `capabilities` entirely: absent means "no update", preserving the masked set the source sent rather than falsely enabling Send/Compact/Fork on a session the hub cannot resume. The fork field is still resolved through `applyHubForkCapability`, so the pushed set matches what a past read returns.

Tests:
- `expectRelaySynthesizedIdleStatus` now takes `wantCapabilities`; the integration test (codex source) asserts the frame carries **none**.
- New `TestRelayGaveUpCapabilitiesOnlyForLocalThreads` pins local → past set (`send=true`, fork floored by the empty-state-dir fence) vs `codex:`/`host:` → `nil`.

Evidence: `go test ./cmd/evener-hub/... -count=1` → ok (all packages); `go vet ./cmd/evener-hub/...` clean; `gofmt -l` clean.

Issue #1715, opened 2026-09-17
appwire-client testing: one gateSettlements helper instead of gateRequests, gateFailures and a hand-rolled kit gate


obra:
## Where

`appwire-client/typescript/testing/fakeClient.ts` (`gateRequests`, `gateFailures`, `deferRequest`) and the callers: `state/extensions/storeLifecycle.test.ts` (the lifecycle kit's `gatedWrite` after #1714, which hand-rolls a resolve+reject gate via `fake.on(..., new Promise(...)) as never`), `state/extensions/launchLayer.test.ts:~130` (`gateFailures`).

## What

`gateRequests` is the resolve-only half and `gateFailures` the reject-only half of one gate, so a test that needs to settle a request either way (the kit's `gatedWrite`) has to hand-roll the gate. Three encodings of "hold this request until the test decides".

## Proposed change

One `gateSettlements(fake, method): { resolve(value), reject(error) }[]` in `testing/fakeClient.ts`, with `gateRequests` / `gateFailures` deleted or reduced to thin aliases, and the kit's local gate removed.

Estimate: ~30 lines net removed, test-only.

Found by the /simplify review of #1714; skipped there as it reaches outside that one-file diff.

PR #1783, opened 2026-09-18
fix(agent): wake the delegate stop reconcile driver when close joins expire (#1538)

### LINKED ISSUE #1538: delegate stop reconcile driver is orphaned when the close join budget expires (no wake on close)
Split from #1394 (bounded-wait audit of agent/delegate_tree_stop.go).

`runStopReconcileDriver` drains on `context.Background()`, so the goroutine started by `StopSubtreeAndDrive` exits only when its stop completes (`stop.done` closes) or a reconcile step errors. `closeRuntimeTree` and `closeStoreAfterStopReconcileDriver` join the driver under a bounded ctx (closeStopJoinContext, half of LaneClosePassBudget) and return `(false, ctx.Err())` when it expires, but nothing wakes the driver afterwards: setting `closing` does not call `signalStopProgressLocked`, so a driver parked in `waitForDelegateStopProgress` stays there for the life of the process once the store is closed and no further progress can arrive.

Reproduced in tests as the leaked `delegate_tree_stop.go:801` goroutines in the #1394 dumps (fixed test-side in the PR that references this issue: the tests never settled their stops). In production the same shape needs a child whose stop cannot complete before the close join budget runs out. Candidate fix: close wakes the driver (signal progress or close a stop-scoped ctx when `closing` is set) so drainStop observes the fence and exits; RED test: a pending stop, a close whose join budget expires, assert the driver's done channel closes.

obra:
## Root cause

`StopSubtreeAndDrive` hands reconciliation to a process-only driver goroutine that
ran `drainStop(context.Background(), ...)`. When a stop cannot complete, the driver
parks in `waitForDelegateStopProgress`, and nothing else ever signals it.

`closeRuntimeTree` and `closeStoreAfterStopReconcileDriver` join that driver under a
bounded context (`closeStopJoinContext`, half of `LaneClosePassBudget`) and return
`(false, ctx.Err())` when the join expires. On that path nothing woke the driver, so
once the delegate store closed and no further progress could arrive the parked
goroutine outlived the close for the life of the process (the leaked
`delegate_tree_stop.go` goroutines seen in #1394's dumps).

## Change

- The stop driver carries a cancelable, process-scoped drain context
  (`newDelegateStopDriver`), and `runStopReconcileDriver` drains on it instead of
  `context.Background()`.
- `joinExactStopReconcileDriver` calls `driver.abandon()` on the sole `not joined`
  exit — i.e. when the bounded join gives up — so a parked driver observes the
  canceled context and exits.
- `waitForDelegateStopProgress` gains a test-only `observeDelegateStopWait` hook so
  a test can await the driver's park signal instead of guessing with a sleep
  (same convention as `ObserveCloseStopJoin`).

Cancellation happens only after the join window is exhausted, so a stop that can
still complete within its close budget keeps draining on the driver as before; the
close error contract (`context.DeadlineExceeded`, and the budget-cancel it triggers)
is unchanged.

## Test

`TestDelegateControllerCloseWakesParkedStopReconcileDriver` builds a pending stop
whose reconcile can never complete (an attached, active generation keeps
`stop.active` non-empty), awaits the driver's park signal, closes with an
already-expired join budget, asserts the close still reports
`context.DeadlineExceeded`, and then asserts the driver's `done` channel closes.
It fails before the fix (driver parked forever) and passes after.

## Evidence

- `go test ./agent/ -run TestDelegateControllerCloseWakesParkedStopReconcileDriver -count=5` — PASS
- `go test ./agent/ -count=1 -timeout 1800s` — `ok primeradiant.com/evener/agent 883.183s`
- `go test -race ./agent/ -run 'TestDelegateControllerCloseWakesParkedStopReconcileDriver|TestDelegateControllerClose|TestDelegateControllerRootClose|TestDelegateStop|TestCloseStopJoin' -count=1` — `ok ... 8.900s`
- `go test ./cmd/evener/ -run 'Wedge|Wedged|Shutdown' -count=1` — `ok ... 9.019s`
- `golangci-lint run ./agent/` — `0 issues.`
- `gofmt -l agent/delegate_tree_stop.go agent/delegate_tree_stop_test.go` — clean
- `go vet ./agent/` — clean

Fixes #1538


obra:
Addressed the roborev Medium finding at head `c273ab5eb`: the test-only `observeDelegateStopWait` hook is now an `atomic.Pointer[func()]`. `waitForDelegateStopProgress` loads it atomically, and the test stores/restores it atomically (the hook is stored before the driver goroutine is launched and restored in `t.Cleanup` after `driver.done` closes). Local gates on the new head: `go test ./agent/ -count=1` green, `go test -race ./agent/` (stop/close selection) green, `golangci-lint run ./agent/` 0 issues, gofmt/vet clean. CI for this head is queued awaiting runners; roborev re-review is in progress.

obra:
Two checks are red on this head (`static-build`, `web`) and both are a pre-existing `main` breakage, not introduced here.

Evidence:
- `static-build` (job 105469894524): `src/panes/session/composer/RepoLocation.test.tsx(4,22): error TS2305: Module '@evener/appwire-client/testing/fakeClient' has no exported member 'gateRequests'` -> `make/building.mk:84: build-web Error 2`.
- `web` (job 105469894419): same TS2305, then `TypeError: gateRequests is not a function` at `RepoLocation.test.tsx:182` -> `make/testing.mk:18: test-web Error 1`.
- `origin/main` at `daf43f6b1` is itself inconsistent: its `appwire-client/typescript/testing/fakeClient.ts` exports `gateSettlements` (renamed by #1715, b86ef963b) while its `RepoLocation.test.tsx` still imports `gateRequests` (from #1754, 61ebe90b1). main's own push run 35303315125 fails identically (`Build snapshot artifacts` -> same TS2305).

This PR (`fix(agent)`, #1538) touches only `agent/delegate_tree_stop{,_test}.go`; the frontend typecheck is unaffected by it and cannot be fixed from within this unit without expanding scope into the frontend package owned by #1715/#1754. The correct fix is a one-line port in RepoLocation.test.tsx (`gateRequests` -> the new `gateSettlements` API) on main; once main is green, this PR's frontend checks should pass on a rerun/merge. All Go checks on this head pass locally and are green in CI.

obra:
Re the Medium (cancellation coverage gaps, `delegate_tree_stop.go`): the gap is real, and I fixed it at head `f353a39a`.

**Confirmed uncovered:** `drainStop`'s three busy branches -- stale requirements (`!current`), the `errDelegateTargetBusy` retry, and the post-repair/attention-plans `continue` -- re-entered the loop without ever observing cancellation, so an abandoned driver could spin forever. Only `waitForDelegateStopProgress` observed `ctx`.

**Fix:** the driver now passes its own context's `Done()` as an explicit abandon signal into `drainStopAbandonable`; each busy branch returns `ctx.Err()` when that signal has fired. The close's own drains pass `nil`, deliberately: they must keep reconciling after their budget is cancelled (the stop's final reconciliation depends on it -- e.g. the cold-attention and live-shell-before-finalizer close paths), and only their wait observes the context. Gating on the abandon signal keeps those close semantics intact.

**On the evidence/repair sub-claim:** those operations are not context-interruptible mid-call, but they are finite (regular-file reads over a bounded delegate set; bounded plan lists), so the drain now exits at the next loop boundary rather than parking.

**Proof:** `TestDelegateStopDrainObservesCancellationAtBusyBoundaries` exercises both busy branches via `drainStopAbandonable` with a cancelled driver context; it spins to the 5s tripwire without the guard (RED confirmed) and returns `context.Canceled` with it. `go test ./agent/ -race -run 'TestDelegateStopDrainObservesCancellationAtBusyBoundaries|TestDelegateControllerCloseWakesParkedStopReconcileDriver|TestNoBareWallClockDeadlineInAgentTests' -count=10` passes; full `go test ./agent/` is green locally. The dormancy inventory in `delegate_tree_controller_test.go` was updated for the renamed method.

PR #1786, opened 2026-09-18
refactor(appwire): consolidate the plain-object guard into one shared module (#1425)

### LINKED ISSUE #1425: appwire-client: four private copies of the plain-object guard (isRecord / isPlainObject)
Noticed during the /simplify pass on #1424 (C18). The AppWire TypeScript package carries the same `typeof value === "object" && value !== null && !Array.isArray(value)` guard as module-private helpers in four places:

- `appwire-client/typescript/client.ts:132` `isRecord`
- `appwire-client/typescript/toolCallText.ts:114` `isPlainObject`
- `appwire-client/typescript/activityData.ts:212` `isPlainObject` (a stricter prototype-checking variant)
- `appwire-client/typescript/launchInherited.ts:29` inlined in `asEnvEntries` (moved verbatim from the web tree by C18; the duplication predates the move)

None is exported, so there was nothing for the moved module to call. Follow-up: export one guard from a shared spot inside the package and point the four sites at it, deciding whether the prototype check in `activityData.ts` is load-bearing there or should become the shared behaviour. Not a C18 change; filed so it does not live only in a PR comment.

obra:
Fixes #1425

## Root cause

The AppWire TypeScript package carried four private copies of the same
non-null, non-array object guard:

- `client.ts` `isRecord`
- `toolCallText.ts` `isPlainObject`
- `activityData.ts` `isPlainObject` (stricter: prototype must be `Object.prototype` or `null`)
- `launchInherited.ts` `asEnvEntries` (inlined; moved verbatim from the web tree by #1424)

None was exported, so the moved module (and the others) had nothing shared to call.

## Change

- New `appwire-client/typescript/plainObject.ts` exporting `isPlainObject`, the
  package's one guard, re-exported from the package root and added to
  `tsconfig.build.json` so it ships and is reachable.
- All four sites now call it; the three private/inline copies are deleted.

**Prototype check:** the strict variant wins. `activityData.ts` applies the
guard to every node of the recursive `evener/jobs/list` tree, where rejecting a
class instance or other non-plain object before walking its keys is the point.
The other three sites read JSON-derived wire values, whose prototypes are always
`Object.prototype`, so the prototype check is a no-op there rather than a
behaviour change. This is the narrowest of the four, so it never loosens the
activity parser.

## Test evidence

Test-first: `plainObject.test.ts` pins the shared contract (plain objects and
`Object.create(null)` accepted; `null`, arrays, primitives, functions, class
instances and `Date` rejected) and failed to resolve `./plainObject` before the
module existed.

- `vitest run ../../../appwire-client/typescript/` (full package suite, via the
  frontend Vitest project): **93 files, 1901 tests passed**.
- Focused: `plainObject`, `toolCallText`, `activityData`, `activityData.edge`,
  `launchInherited`, `client`, `publicExports`: **7 files, 169 tests passed**.
- `tsc -p appwire-client/typescript/tsconfig.build.json --noEmit`: exit 0.
- `npm run qualification` (package pack/install/reachability/consumer smoke):
  exit 0.
- `biome check` on the changed files: exit 0.

## Out of scope (noted for a follow-up only)

Identical guards still live outside #1425's named four:
`watchRows.ts` `asJsonObject`, `transcriptDisplayConfig.ts` `isRecord`, and
`state/navigation/codec.ts` `isRecord`. The issue scoped this to its four sites,
so they are left untouched.


obra:
Thanks — both findings addressed at head 5860def0c.

**Finding 1 (`plainObject.ts` header overclaims "the package's one guard")** — fixed via the finding's second option: the header now documents the guard's contract and names the call sites it serves rather than claiming to be the package's only copy. The two remaining private copies you named (`transcriptDisplayConfig.ts:113`, `state/navigation/codec.ts:46`) are deliberately **out of scope for #1425**: the issue enumerates exactly four call sites (`client.ts`, `toolCallText.ts`, `activityData.ts`, `launchInherited.ts`) and this PR is bounded to the files it names. Those two live in other layers (the transcript-display config module and the `state/navigation` subpath), and folding them in would grow a tightly-scoped dedup into an unbounded sweep. Worth a follow-up issue; I've left them untouched here. (For completeness, `watchRows.ts`'s public `asJsonObject` and `examples/discovery-logic.mjs` also carry similar, non-identical guards — `asJsonObject` returns `JsonObject | undefined` rather than a boolean type predicate — same follow-up.)

**Finding 2 (comments narrated refactor history / cited the PR)** — fixed: both `plainObject.ts` and `plainObject.test.ts` now document present-tense behaviour and why the strict prototype check is required, with no historical references. The issue reference remains only in commit messages, as the repo requires.

PR #1787, opened 2026-09-18
test(gate): exercise gate budget wiring behaviorally (#1194)

### LINKED ISSUE #1194: tests: exercise gate budget wiring behaviorally instead of matching script text
Follow-up from PR #1176, raised in roborev rounds 1, 2, 4, and 5.

`loadawareworkers_test.go`'s `TestRunModuleTestsUsesLoadAwareBudgets` and `buildScripts.test.ts`'s script checks assert rendered script text (substrings, a `--maxWorkers` regex, and previously exact `${VAR-...}` assignments). They are brittle to harmless refactoring and cannot observe whether a computed budget is actually used.

The helper's own behavior is covered by real fixtures; the gap is the plumbing in `scripts/gate/run-module-tests.sh` and `cmd/evener-hub/frontend/package.json`. Suggested direction: exercise the plumbing with a stubbed `go`/`/usr/bin/time` on `PATH`, or extract the budget assignment into a sourceable function, and assert the effective arguments rather than the script text. Keep the current checks only as a minimal smoke assertion.

obra:
Fixes #1194

## Root cause

`TestRunModuleTestsUsesLoadAwareBudgets` and the `buildScripts.test.ts` script
checks asserted rendered script text: substrings, a `--maxWorkers` regex, and
previously exact `${VAR-...}` assignments. Such assertions pass on a comment,
pin formatting rather than behavior, and cannot observe whether a computed
budget is actually used. The budgets were computed inline in
`scripts/gate/run-module-tests.sh` and inline in `package.json`'s `sh -c`, so
there was nothing behavioral to call.

## Change

Extract the wiring into a sourceable POSIX library, `scripts/lib/gate-budgets.sh`:

- `gate_budget CAP DEFAULT` — the load-aware worker count for `CAP`, or the
  fixed default when the helper is absent or the answer is unusable.
- `gate_init_budgets` — sets the gate's `ROOT_P` / `AGENT_P` / `AGENT_PARALLEL`
  and exports `AGENT_SHARD_PARALLEL` / `AGENT_SHARD_SURVEY_PARALLEL`, honoring
  explicit environment overrides.
- `gate_module_flags MODULE` — the effective `-p`/`-parallel` flags handed to
  `go test`.
- `vitest_run_args` — the effective `--maxWorkers=…` flag handed to `vitest run`.

Both consumers now delegate to it: `run-module-tests.sh` sources the library and
calls `gate_init_budgets` / `gate_module_flags`, and `package.json`'s
`test`/`test:coverage` call `vitest_run_args`.

`gatebudgets_test.go` (new) exercises the effective arguments directly: it
sources the library in a POSIX shell, replaces the helper's two machine probes
(`load_aware_cores`, `load_aware_load1`) with fixture values — so the real
`load_aware_workers` arithmetic still runs, independent of this machine's load —
and asserts the flags/arguments built for idle, loaded, and saturated machines,
plus environment overrides and the helper-absent fallback. The remaining
script-text checks are the minimal smoke assertion the issue asks to keep.

Note: the issue suggested stubbing `go`/`/usr/bin/time` on `PATH`; that is
banned by `docs/developing-evener/testing.md` ("faking the toolchain to test a
script is banned outright"), so this takes the other suggested direction,
extracting the budget assignment into a sourceable function and asserting the
effective arguments.

## Test evidence

```
$ go test . -run 'TestGateModuleFlags|TestGateBudget|TestVitestRunArgs|TestRunModuleTestsWires|TestLoadAware' -count=1
ok  	primeradiant.com/evener	0.351s

$ gofmt -l gatebudgets_test.go loadawareworkers_test.go   # no output
$ go vet .                                                # exit 0

$ cd cmd/evener-hub/frontend && npx vitest run src/buildScripts.test.ts
 Test Files  1 passed (1)
      Tests  5 passed (5)

$ npx tsc --noEmit --incremental false                    # exit 0
$ npx biome ci src/buildScripts.test.ts                    # no fixes applied

$ WAVE1= WAVE2= WEB=0 scripts/gate/run-module-tests.sh -short -count=1
gate-smoke-exit=0   # refactored script parses, sources the library, exits clean
```

Before the library existed, the new tests failed with
`stat scripts/lib/gate-budgets.sh: no such file or directory`.


obra:
Addressing the roborev findings in `0eda5f08d`:

**Medium — ambient env vars broke test determinism.** Reproduced before the fix:

```
$ AGENT_PARALLEL=6 AGENT_SHARD_PARALLEL=1 AGENT_SHARD_SURVEY_PARALLEL=1 \
    go test . -run TestGateModuleFlagsFollowTheLoadAwareBudget -count=1
--- FAIL .../a_loaded_machine_backs_every_budget_off
    agent=-p 2 -parallel 6   shards=1 1
    want  agent=-p 2 -parallel 2   shards=2 2
```

`runSourcedGate` now gives the sourced shell a minimal environment (`PATH` and
`LC_ALL=C`) instead of `os.Environ()`, so neither `make test`'s exported shard
budgets nor `make test-race`'s `AGENT_PARALLEL=6` can reach the fixture. The
same command now passes; the override subtest still sets its own values
explicitly.

**Medium — helper sourcing not pinned.** Extracted it into
`gate_source_helper PATH` (in `gate-budgets.sh`) and covered it behaviorally:
`TestGateSourceHelperMakesTheBudgetsLoadAware` proves a readable helper shrinks
a loaded fixture's budget to 2 and an unreadable path keeps the fixed default of
4. The smoke list now names `gate_source_helper` and `load-aware-workers.sh`, and
`buildScripts.test.ts` pins the frontend's `gate_source_helper
../../../scripts/lib/load-aware-workers.sh` call.

**Low — no fallback if `gate-budgets.sh` is unreadable.** Both consumers now
refuse loudly instead of running unbudgeted (Go gate: message + `exit 2`;
frontend: message + `exit 1`). The library ships in the same commit as both
consumers, so an unreadable copy is a broken checkout; failing beats silently
dropping the budget the gate exists to apply. The helper-absent case still
degrades to the historical fixed defaults, now via `gate_source_helper`.

**Low — validation accepted `0`.** `gate_budget` now falls back to the default
when the helper answers zero, and the fallback table gained a zero-output
fixture.

Re-verified after the fixes: `go test .` focused set ok (also under the ambient
variables above), `gofmt` clean, `go vet .` exit 0, frontend vitest 5/5, `tsc`
exit 0, biome clean, gate smoke exit 0.

obra:
Round 2 addressed in `a216b4007`. Both medium findings were real:

**Medium — `command -v` alone could run an ambient `load_aware_workers`.**
Reproduced before the fix:

```
$ d=$(mktemp -d); printf '#!/bin/sh\nprintf %%s 99\n' > "$d/load_aware_workers"; chmod +x "$d/load_aware_workers"
$ PATH="$d:$PATH" sh -c '. scripts/lib/gate-budgets.sh; printf "%s\n" "$(gate_budget 4 4)"'
99
```

`gate_source_helper` now records a successful source in `_gate_budgets_helper_sourced`,
and `gate_budget` consults that flag as well as `command -v`, so only the helper
this library sourced is used. Covered by
`TestGateBudgetIgnoresAnAmbientWorkerCommand` (a decoy executable on PATH must
be ignored). The pre-existing bad-answer fallback subtests now source the helper
first, so they genuinely exercise the stub instead of passing through the fixed
default; verified directly, `with-source=2` vs `without-source=4`.

**Medium — source failures were ignored.** Both consumers now check the source
status, not just readability: `run-module-tests.sh` uses `if ! . "$lib"` and
exits 2, and `package.json` adds `|| { …; exit 1; }`. Verified both branches
fire (`exit=2` / `exit=1`).

**Separately, in its own commit `a216b4007` — unrelated main breakage.**
The `web` check cannot go green because `main` is already red: #1755 reshaped the
appwire-client test-gate helper, and #1754 (merged seconds later) imported the
old name. main's own web check failed at `61ebe90b1`:

```
src/panes/session/composer/RepoLocation.test.tsx(4,22): error TS2305:
Module '"@evener/appwire-client/testing/fakeClient"' has no exported member 'gateRequests'.
```

`gateRequests` returned one resolver per call; its replacement
`gateSettlements` returns `{resolve, reject}`, so I updated the two call sites to
`.resolve(...)`. That file is byte-identical to `origin/main`, so this is main's
failure, not this PR's; I kept it in a separate commit and disclose it here
rather than let it hide. Local evidence: tsc exit 0, RepoLocation 20/20 tests
pass. Say the word if you would rather I drop that commit and have main repaired
separately.

Round-2 verification: focused `go test .` ok, gofmt clean, `go vet .` exit 0,
gate smoke exit 0, frontend vitest 25/25 across both touched files, tsc exit 0,
biome clean.

obra:
Rebased onto `main` and force-pushed `bb5b480fe` (branch is now an ancestor-checked descendant of `origin/main`).

Two rebase notes:

- **Dropped my `RepoLocation.test.tsx` commit.** Main now carries the identical `gateSettlements` + `.resolve(...)` fix, so the separate repair is redundant. This PR is back to only the #1194 gate-budget commits.
- **Resolved one conflict in `loadawareworkers_test.go`.** Main added `mountinfoEscaped`, `TestLoadAwareCgroupAbsoluteMembership`, `TestLoadAwareCgroupEscapedMountPath`, `TestLoadAwareCgroupMountDecodesEscapes`, and `TestLoadAwareCgroupTrailingNewlinePaths`; I kept all of them and removed only `TestRunModuleTestsUsesLoadAwareBudgets`, whose smoke assertion moved to `gatebudgets_test.go` as the issue asks.

Post-rebase verification: focused `go test .` ok, gofmt clean, `go vet .` exit 0, gate smoke exit 0, frontend tsc exit 0, vitest 25/25 across `buildScripts.test.ts` and `RepoLocation.test.tsx`, biome clean.

obra:
Round 3 addressed in `f535c3d3a`.

**Medium — duplicate `PATH` in the decoy fixture.** `runSourcedGateEnv` appended
the override after the default `PATH`, so the child had two `PATH` entries and
the winner depended on Go's duplicate removal rather than on the fixture's
intent. `envOverride` now replaces the entry instead of appending, and
`TestGateBudgetIgnoresAnAmbientWorkerCommand` first asserts the decoy is
resolvable from the child (printing `command -v load_aware_workers`) before
checking the budget, so the case cannot pass while proving nothing.

Mutation check that the test now has teeth:

```
# gate_budget reverted to `command -v` alone:
--- FAIL: TestGateBudgetIgnoresAnAmbientWorkerCommand
    child = "resolved=…/load_aware_workers\nbudget=99", want "…budget=4"
# restored:
ok  primeradiant.com/evener
```

Re-verified: focused `go test .` ok, gofmt clean, `go vet .` exit 0, library
syntax ok.

PR #1790, opened 2026-09-18
fix(appwire-client): unify spawn and launch scalar collect rule (#1444)

### LINKED ISSUE #1444: appwire-client: spawnSchema.collectAdvancedOverrides and launchSchema.collectScalar encode the same collect rule and have drifted
Surfaced by the /simplify reuse review on #1443 (C17), pre-existing, not introduced there.

`appwire-client/typescript/spawnSchema.ts` (`collectAdvancedOverrides`, ~lines 32-64) and `appwire-client/typescript/launchSchema.ts` (`collectConfig` / module-private `collectScalar`, ~lines 298-341) both encode the launch-config collect rule (boolean tri-state → omit/true/false, integer coercion, drop empty scalars, include collections only when non-empty) over two state shapes (`AdvancedValues` with an `invalid` flag vs `LaunchFormState`).

The two encodings have drifted:
- `collectScalar` trims text scalars and parses integers with `Number(trimmed)`; `collectAdvancedOverrides` sends text scalars untrimmed and uses `Number.parseInt(v, 10)`, so `"12abc"` becomes `12` in the spawn pane and is dropped as NaN in settings.

Same class as the two-encodings-of-one-rule findings on #1105 / #1137. Suggested shape: export `collectScalar` from `launchSchema.ts` and have `collectAdvancedOverrides` delegate its scalar arm to it, with a test pinning the shared behaviour for both callers. Both modules now live in the package root, so this is a package-internal change.

obra:
Fixes #1444

## Root cause
`spawnSchema.collectAdvancedOverrides` carried its own copy of the launch-config scalar collect rule, drifted from `launchSchema.collectScalar`:

- text scalars were sent untrimmed (and a whitespace-only value was still sent),
- integers used `Number.parseInt(v, 10)`, so `"12abc"` became `12` in the spawn pane while the settings form (using `Number(trimmed)`) produced `NaN`.

## Change
- `collectScalar` is now exported from `launchSchema.ts`.
- `collectAdvancedOverrides` delegates its scalar arm (boolean / integer / select / radio / text / path / modelPicker) to `collectScalar`; the list/env/mcp collection arms are unchanged.
- `collectScalar` now drops an integer that does not parse to a finite number instead of emitting `NaN`, so `"12abc"` is omitted in both surfaces rather than becoming `12` (spawn) or `NaN` (settings).
- Tests pin the shared rule for both callers.

## Test evidence
Pre-fix, the new tests fail exactly as the issue describes (settings `maxRounds` came back `NaN`; the spawn pane produced untrimmed text, `12` for `"12abc"`, and kept whitespace).

Post-fix:
- `npx vitest run ../../../appwire-client/typescript/spawnSchema.test.ts ../../../appwire-client/typescript/launchSchema.test.ts` — 72 passed
- `npx vitest run ../../../appwire-client/typescript` — 1900 passed (92 files)
- `tsc -p appwire-client/typescript/tsconfig.build.json --noEmit` — exit 0
- frontend `src/panes/spawn` suite: `Spawn.test.tsx` 290 passed with `--testTimeout=30000`. Under heavy host load the default 5000ms per-test timeout trips two tests that deliberately sleep 5000ms; raising the timeout makes them pass, so this is environmental, not a regression from this change.


obra:
Addressed the RoboRev medium finding in 271d091:

- `collectScalar` now uses `Number.isInteger` instead of `Number.isFinite`, so `"12.5"` is dropped for `kind: "integer"` rather than emitted as a fraction. Confirmed the wire type is `*int` (`appwire/types.go:3313-3314` for `MaxRounds`/`MaxSubagentDepth`), so a fraction would fail `thread/start`'s decode.
- Chose omission over truncation: a fractional value is not an integer the user meant to send, and dropping it is consistent with the NaN/empty handling already in this rule.
- Added fractional-input coverage on both callers: `collectConfig` (`launchSchema.test.ts`) and `collectAdvancedOverrides` (`spawnSchema.test.ts`).

Verification: focused tests 72 passed; full `appwire-client/typescript` suite 1906 passed (93 files); `tsc -p tsconfig.build.json --noEmit` exit 0.

obra:
Addressed the second RoboRev finding in 94f4373:

- `collectScalar` now uses `Number.isSafeInteger` (was `Number.isInteger`), so values outside the safe-integer range — `"1e21"`, and literals like `"9007199254740993"` that `Number()` rounds — are dropped instead of being sent to a Go `*int` wire field that cannot decode them. `isSafeInteger` also subsumes the NaN/Infinity/fraction rejection from the previous revision.
- Added boundary tests on both callers: `collectConfig` (`launchSchema.test.ts`) and `collectAdvancedOverrides` (`spawnSchema.test.ts`) pin that `"9007199254740991"` (MAX_SAFE_INTEGER) is sent and `"1e21"` / `"9007199254740993"` are dropped.
- Also fixed the `web` check failure on the previous head: a Biome `organizeImports` violation from the new `./launchSchema` import in `spawnSchema.ts` (now sorted). `npm run lint` passes locally.

Verification: focused tests 73 passed; full `appwire-client/typescript` suite 1907 passed (93 files); `tsc -p tsconfig.build.json --noEmit` exit 0; `biome ci src ../../../appwire-client/typescript` exit 0.

obra:
Addressed the High review finding in c92f71c.

**Fix:** `collectScalar` now validates the *raw* trimmed string before any numeric conversion: a strict decimal-integer pattern (`/^[+-]?\d+$/`) plus a `BigInt` range check against `[Number.MIN_SAFE_INTEGER, Number.MAX_SAFE_INTEGER]`. Only an exact decimal integer inside the safe range is converted to `Number`.

**On the finding's example:** `Number("9007199254740993")` rounds to `9007199254740992`, and `Number.isSafeInteger(9007199254740992)` is `false` (2^53 > MAX_SAFE_INTEGER), so that case was actually already dropped and its test passed. The reviewer's mechanism is real, though: `Number("1.0000000000000000001") === 1` and `Number.isSafeInteger(1)` is `true`, so the pre-check rounding could send `1` for a value nobody typed. The exact-decimal rule closes that path and makes the outcome independent of double rounding.

**Tests:** boundary coverage on both callers now asserts dropped for `"1.0000000000000000001"`, `"9007199254740993"`, `"1e21"`, `"12abc"`, `"12.5"`, and sent for `9007199254740991` / `-9007199254740991` and `" 12 "`.

**Verification (head c92f71ca7):** focused vitest 73 passed; full `appwire-client/typescript` suite 1916 passed (93 files); `npx tsc --noEmit --incremental false` (cmd/evener-hub/frontend) exit 0; `npm run lint` exit 0.

obra:
Addressed both findings on head 71314c8d3:

**Medium — native parser reused the wrong coercion.** `mobile-native/src/launchScalar.ts` now calls a shared `isExactSafeInteger()` exported from `launchSchema.ts` (the exact-decimal + `BigInt` range test that the `collectScalar` integer arm uses) instead of `Number(value)` + `Number.isSafeInteger`. Native and web now agree: `"1.0000000000000000001"` / `"1e21"` / `"9007199254740993"` / `"12abc"` are rejected, `9007199254740991` is accepted. Added regression cases to `mobile-native/src/launchScalar.test.ts`.

**Low — comment corrected.** `spawnSchema.test.ts` now describes `"1e21"` as exponent notation rejected as an unsafe magnitude, and attaches the rounding explanation to the `"1.0000000000000000001"` input.

Commands and results:
- `cd mobile-native && npx vitest run src/launchScalar.test.ts` — 6 passed
- `cd cmd/evener-hub/frontend && npx vitest run ../../../appwire-client/typescript/spawnSchema.test.ts ../../../appwire-client/typescript/launchSchema.test.ts` — 73 passed
- `npx vitest run ../../../appwire-client/typescript` — 1916 passed (93 files)
- `npx tsc --noEmit --incremental false` (frontend) — exit 0
- `npx tsc --noEmit --project tsconfig.check.json` (mobile-native) — exit 0
- `npm run lint` (Biome over frontend src + appwire-client/typescript) — exit 0

Head: 71314c8d318b9afb4269166750c4358c0b84319b (rebased on main 822ac99b2).

Issue #1693, opened 2026-09-17
native: provider-owned local preferences controller so the draft hatch works without a client (D6 follow-up)


obra:
Class fix for the native unreadable-draft hatch, measured on D6 #1549 round 14 (2026-09-17) after six review rounds landed on the same affordance (rounds 8, 9, 10, 11, 13, 14).

Confirmed by citation: the native preference screens gate the readable-draft discard on `preferences.connected` (`busy` folds in `available`) and their callbacks require a live `model`, so a local operation never reaches an offline state; and on a cold start with a hub but no client, `NativePreferencesProvider` keeps `bound` null, so no snapshot and no `draftUnreadable` is exposed and the screens render only the reconnect prompt. The store's own gate (`assertDiscardable`) never required connectivity (pinned by a keybindings store test in #1549).

Fix as its own native-only PR stacked on #1549: a provider-owned local preferences controller per hub that exists whether or not a client is bound, owning the persisted drafts, the unreadable classification, the discard, and the exposed snapshot; the connected store layers on top when a client arrives; the screens never gate a local operation on connection or model. Sizing: roughly -10 to -30 net lines; every nativePreferenceDrafts and package store test survives, `snapshotAfterLocalDiscard` and its unit test go. Two risks to design for: a local-to-connected handoff must not show a coarser classification for one render (export the unreadable-checkpoint validation from appwire-client or keep the local layer deliberately coarse), and the controller must route through the live model's own discard when one exists so the connected UI cannot show stale in-memory draft/conflict flags.

obra:
Built on branch `claude/native-local-preferences-controller` at 14c1e4d8b, stacked on #1549 @db7f7d08b, held until #1549 merges. Measured result: +77 net production lines (the controller module is +99 with house-style docs; provider -20, nativePreferences -21, screens +11, package exports +8), not the -10 to -30 estimated on round 14; the value is one mechanism replacing six rounds of per-state patches. Seven controller tests cover cold start without a client, offline discard of readable and unreadable drafts, hub isolation and port failure. Residual for its review: the transcript screen's no-snapshot fallback branch is now unreachable past the profile early-return.

PR #1791, opened 2026-09-18
D6 piece 6: an unreadable draft record never locks the keybindings section

obra:
## D6 piece 6: draftUnreadable + assertDiscardable

Plan item 6 of the D6 decomposition (docs/superpowers/plans/2026-09-12-sdk-migration.md's D6 split, `.superpowers/sdd/2026-09-12-mobile-landing-queue-cont/d6-split-plan.md`): "An unreadable record never locks the section, store side only (draftUnreadable, assertDiscardable, cleared uncertainty); the native screen projection moves to #1693." Stacks on #1741 (p5).

### Mechanism
- `KeybindingsStoreFields.draftUnreadable: boolean` - set when the stored draft record cannot be decoded (distinct from the port itself being unreachable).
- `restoreDraft`'s catch discriminates `UnreadableDraftError` from a generic port failure: on decode failure, `draft`/`writeUncertain`/`draftConflict` clear (they described a record that no longer exists to describe) and `draftUnreadable: true` is set alongside the existing `storageUnavailable: true`.
- `assertDiscardable()` replaces `assertEditable()` as `discardDraft`'s gate: `fence.disposed || saving || writeUncertain || (storageUnavailable && !draftUnreadable)`. Narrower than the editor's gate (drops hubSupport/loaded/loadError - discarding needs no confirmed state to compose against), and an unreadable record does not itself refuse discard - throwing the record away is exactly the recovery it allows.
- `refreshOverrides`'s `storageUnavailable` guard also lets an unreadable-record generation through, so the hub read is never held hostage to a record no build can decode.
- `discardDraft` clears `draftUnreadable: false` alongside `storageUnavailable: false` on a successful discard.

`discardClassified`'s existing raw-identity tracking (piece 3/5: `lastClassified` is set from `load()`'s raw value *before* `decode()` can throw) already names an unreadable record correctly on discard - no separate `discardUnreadable()` method needed, unlike the pre-split #1549 design this piece is adapted from (its `draftRepository` didn't yet track raw identity across decode failures).

### Native (43 non-test lines, kept in this PR per the brief's <100-line threshold)
- `NativePreferencesProvider.tsx`'s `backend.get()` returns unparsable bytes raw instead of throwing, so the shared store's own decoder classifies them as `UnreadableDraftError` rather than the port looking dead.
- `matches()` (backing `deleteIf`/`replaceIf`) also compares the raw string directly, since a discard on an unreadable record hands back exactly those bytes, not their JSON-stringified form.
- `KeybindingPreferencesScreen.tsx` offers a "Discard unreadable draft" action whenever `domain.draftUnreadable` is set, gated only on saving/writeUncertain/loading (not the broader `busy` gate, which requires a confirmed payload the section may not have while unreadable).
- `nativePreferences.ts` projects `draftUnreadable` onto the keybindings domain (`PreferenceState<T>` gets the field; transcript's projection stays `false` via `initialDomain()` since transcript has no equivalent concept yet).

### Oracle tests (adapted from #1549's `2ab17ff78`/`ec55c4e46`/`46dc55a19`/`4435e277f`)
- `an unreadable stored record never locks the section`
- `a record that becomes unreadable while a write is uncertain clears the stale uncertainty, unblocking discard`
- `a discard refuses and re-classifies when the unreadable record has been replaced`
- native: `classifies bytes it cannot parse as an unreadable record, and clears them` (nativePreferenceDrafts.test.ts, the pure port-forwarding layer)

### Falsification
Reverted `keybindingsStore.ts`, `NativePreferencesProvider.tsx`, `nativePreferences.ts`, `KeybindingPreferencesScreen.tsx` to #1741's head and re-ran `keybindingsStore.test.ts`: the 3 new tests fail as expected (`draftUnreadable` missing/undefined, `loaded: false` instead of `true`). Restored with `git apply`; `git status` clean before committing.

## Gates
- `npx vitest run keybindingsStore.test.ts draftCheckpointPort.test.ts` (appwire-client/typescript): 82 passed
- `npx vitest run` (mobile-native, full suite): 797 passed, 83 files
- (no index.ts/tsconfig.build.json touch, so `make test-api-package` falsification not required)


obra:
## Round fixes (202ae575e, on top of the fold #1808)

### Medium: uncertain-write settle swallows storage-unavailable signal
Fixed in #1808 (the fold of pieces 3-5), inherited here through the restack: `settledWrite`'s catch on a throwing `replaceClassified` now returns `{ storageUnavailable: true, draftError: DRAFT_SAVE_FAILED_MESSAGE }`. Not new work on this PR.

### Medium: native TypeScript typecheck failure from `deleteIf` parameter narrowing
Refuted in #1808 by measurement: `npm run check` in `mobile-native` is clean on this tree (and was already clean before the fold's identity-widening fix - CI's native job is green at the reviewed head `4c4a440da`).

### Medium: `removeIf`/`replaceIf` identity typed `Checkpoint` but raw unknown passed
Fixed in #1808: `DraftPort`, `KeybindingDraftStorage`, and the native backend interfaces all widened their identity parameter to `unknown`; the runtime casts are gone.

Fixed in #1808 (piece 5's own round), inherited here.

### Low: "Discard unreadable draft" is a silent no-op when offline
Fixed here. Root cause confirmed exactly as described: `NativePreferencesProvider` disposes the model on disconnect but keeps the stale snapshot (still showing `draftUnreadable: true`), and the button's `onPress` went through `run()`, which early-returns on `!model`/`!scope.connected` - so it rendered enabled and did nothing.

Wired the store-free path: `NativePreferencesProvider` now exposes `discardUnreadableKeybindingsDraft()`, which builds the draft port directly from `hubId` and the backend (no live model required) via `discardStoredKeybindingDraft` - the exact host-side recovery the port's own doc names for this case - and updates the retained stale snapshot on success so the screen reflects the clear immediately without waiting for a reconnect. `KeybindingPreferencesScreen`'s `onPress` falls back to it when `model` is null.

Not fixed: a cold offline launch before the client ever connects still shows no discard UI at all (no model is ever built to produce a snapshot, so `domain` is `undefined` and the block never renders) - this is a pre-existing gap in the whole preferences system's binding, not something piece 6 introduced, and out of scope for this fix. Noting it as a residual rather than filing a new issue, since it's really the "no snapshot before first connect" architecture question every preferences screen has, not specific to the unreadable-draft feature.

No automated test added for the provider/screen wiring itself: this repo has no `.tsx` component tests at all (`NativePreferencesProvider.tsx`/`KeybindingPreferencesScreen.tsx` were already untested at that layer before this piece); the mechanism it calls (`discardStoredKeybindingDraft` + `nativeKeybindingDrafts`) is fully covered by existing `appwire-client` and `nativePreferenceDrafts.test.ts` tests.

## Gates
- `npx vitest run keybindingsStore.test.ts draftCheckpointPort.test.ts` (appwire-client/typescript): 90 passed
- `npx vitest run` (mobile-native, full suite): 797 passed, 83 files
- `npm run check` (mobile-native) and `npm run typecheck` (cmd/evener-hub/frontend): clean

obra:
## Round 4 fixes (031528799)

**High — offline discard silently no-ops when a model is retained after disconnect** (`KeybindingPreferencesScreen.tsx`)
The client object survives a connection drop (`model` stays non-null while `preferences.connected` is false), so the old `if (model) { run(...) } else { store-free }` branch always took the dead `run()` path once disconnected. Fixed by branching on `model && preferences.connected` instead of on `model` alone; every other case (no model, or a model retained through a drop) falls through to the store-free path.

**Medium — delete-without-identity race can remove a newer valid checkpoint** (`draftCheckpointPort.ts`)
The store-free path has no live repository to hold the identity it was shown at display time, only whatever `load()` returns right now. `discardStoredDraft` now takes an optional `isReadable` predicate and returns a 3-state outcome (`"removed" | "absent" | "refused"`, exported as `DiscardStoredDraftResult`): when the current record now decodes as valid, it refuses instead of deleting it. Added `isReadableKeybindingDraft` (wraps the store's own `draftCheckpoint` decoder) and exported it; `NativePreferencesProvider.tsx`'s offline discard now passes it through.

**Medium — post-discard snapshot/error state not fully refreshed** (`NativePreferencesProvider.tsx`)
The snapshot update was gated on `if (removed)` and only cleared `draftUnreadable`/`storageUnavailable`, leaving `error` (the draft-restore-failure message) stale. Now every outcome (`removed`/`absent`/`refused`) refreshes the snapshot, clearing `draftUnreadable`, `storageUnavailable` and `error` alike — in all three cases the record the button referred to is no longer the same unreadable thing.

**Low — stored JSON `null` coalesces to "no record"** (`nativePreferenceDrafts.ts`, `NativePreferencesProvider.tsx`)
Decided: treat it as unreadable so it's discardable. Extracted the backend's parse-or-raw-bytes logic into an exported `parseDraftBytes(raw: string): unknown` (shared by the real backend and the mirrored test fake, closing a 3-way duplication the JSON-null fix would otherwise have widened) — a parsed `null` now returns the raw bytes rather than collapsing into the `DraftPort` "absent" sentinel.

**Low — fallback action disabled by stale `loading`** (`KeybindingPreferencesScreen.tsx`)
The disable condition now branches the same way the action does: `domain.loading` only gates the live-model path; the store-free path is disabled only by `saving`/`writeUncertain`.

**Tests:**
- `draftCheckpointPort.test.ts`: refuses/removes per `isReadable`.
- `keybindingsStore.test.ts`: `discardStoredKeybindingDraft` outcome strings; a new test exercising `isReadableKeybindingDraft` end-to-end against a real replaced checkpoint.
- `nativePreferenceDrafts.test.ts`: new `parseDraftBytes` unit tests, plus a port-level test that a stored JSON `null` classifies as unreadable and is removable (mirrors the real backend, per this file's existing convention — `NativePreferencesProvider.tsx`'s Storage-backed backend isn't reachable from this package's tests).

**Falsification:** reverted all 6 production files to base (`git diff | git apply -R`), ran the touched suites — 7 failures in `appwire-client/typescript` (outcome-string mismatches, `isReadableKeybindingDraft`/`discardStoredDraft` refusing-behavior missing) and 5 in `mobile-native` (`parseDraftBytes is not a function`) — restored with `git apply`, `git status` clean.

**Gates:**
- vitest (`draftCheckpointPort.test.ts`, `keybindingsStore.test.ts`): 90/90
- vitest (`nativePreferenceDrafts.test.ts`): 9/9
- `make test-api-package` (index.ts touched): PASS

**Residual / not covered:** the High and two Lows in `KeybindingPreferencesScreen.tsx`/`NativePreferencesProvider.tsx` have no direct regression test — mobile-native has zero `.test.tsx` component tests today (no RTL/RN test harness configured), so "test both branches" for the connectivity fix would mean introducing new test infrastructure, which I did not do without confirming first. Flagging as a residual for the coordinator rather than silently skipping or inventing the harness.

Non-test diff: ~170 lines across 2 commits (SDK package discard-outcome mechanism; native wiring + connectivity branch + null handling).

obra:
## Round 5 disposition (55d0a60) — answering the raw panel reviews

Answering the raw panel members' findings directly (reviews/raw/1791-d627dc4ba.md) since the combined synthesis is broken. Per the "five review rounds means decompose" rule this is the round-5 boundary; every High/Medium/Low the panel raised on this piece is answered below with a fix.

- **High** (member 0, codex): `settledWrite` settles `writeUncertain` after an in-flight draft write rejects, because it only checks the CURRENT `saving` flag and write token — a stale GET whose read began before the write's own outcome was known could clear the uncertainty. **Fixed**: `refreshFor` now captures `savingAtReadStart` before its GET leaves, and `settledWrite` refuses to settle whenever a write was already in flight at read-start time, independent of what `saving`/the write token read by reply time. Commit 96d2568c8.
  - Test: "a refresh already reading while a draft write is in flight never settles that write's uncertainty" — a draft write starts, a refresh's GET starts while it's still out, the write then fails (clearing `saving`, leaving `writeUncertain`), and the stale refresh's reply lands last.
  - Falsified: reverted the production diff — the assertion failed (`writeUncertain` was wrongly cleared to `false`). Restored, test passes.

- **Medium** (echoed across every raw review in this stack): a stored JSON `null` comes back from the shared `backend.get()` as the raw string `"null"` (`parseDraftBytes` preserves it so the KEYBINDINGS port can classify it as a present-but-unreadable record). The SAME `backend.get()` is shared by `nativeTranscriptDrafts`, so a stored transcript `null` broke transcript restoration (`storageUnavailable: true` permanently — transcript has no unreadable-record recovery path). **Fixed**: `nativeTranscriptDrafts.load()` now maps the raw `"null"` sentinel (and `null`/`undefined`) back to "no draft", scoped to the transcript port only; keybindings keeps the raw-preservation behavior unchanged. Commit 142274e32.
  - Tests: "treats a stored JSON null as no draft, not an unreadable record" + "still loads a real stored checkpoint" (`nativePreferenceDrafts.test.ts`).

- **Medium** (echoed across the stack): `discardStoredDraft` mapped a failed `removeIf` (a compare-and-swap refusal) to `"absent"`, conflating "nothing stored" with "still there, but a concurrent writer replaced it with different bytes" (readable OR still unreadable). **Fixed**: on a failed `removeIf`, re-reads storage and reports `"absent"` only if genuinely gone, `"refused"` otherwise (`draftCheckpointPort.ts`). Commit 282b22ecb. The provider (`NativePreferencesProvider.tsx`) now also distinguishes the two "refused" causes: it re-checks whether the CURRENT record decodes as valid and only clears the `draftUnreadable` notice if so (new `draftUnreadableAfterDiscard` helper); a refusal from a still-unreadable replacement keeps the recovery UI up. Commit 5edf6f7d9.
  - Tests: `draftCheckpointPort.test.ts` ("reports 'refused', not 'absent', when removeIf fails but a replacement record is still present" / "... when a re-read finds the record genuinely gone"); `nativePreferenceDrafts.test.ts` (`draftUnreadableAfterDiscard`).

- **Medium** (member 3, codex): the store-free unreadable-draft recovery (the offline "Discard unreadable draft" button) was unreachable on a cold offline start, since `preferences.snapshot` stays `null` until a ready client publishes a model (`bindNativePreferences`'s `ready` callback never fires without a connection). **Fixed**: `NativePreferencesProvider` now inspects the local keybinding draft record independently of connection state on mount/hub change, exposing a new `offlineDraftUnreadable` field (`localDraftIsUnreadable` helper); `KeybindingPreferencesScreen` falls back to it whenever the live `domain` is undefined, and the live domain's own value supersedes it the instant a model publishes. Commit e2a698e11.

- **Low** (member 3, codex): the store-free discard success path never cleared the component-local `error` state, so a stale error from an earlier failed action stayed visible after a successful offline discard (`ErrorMessage` prioritizes it over `domain?.error`). **Fixed**: `setError(null)` on the success path. Commit d81d4aebd.

- **Low** (member 3 in a later round, same code): `matches()` (the native backend's compare-and-remove identity check) accepted only the exact stored string or its canonical `JSON.stringify` re-encoding, so a valid-but-unreadable record with noncanonical formatting (whitespace, key order) could never be matched and removed, even though the bytes are unchanged. **Fixed**: extracted `matchesStoredBytes`, which also parses the stored bytes and compares their OWN re-encoding against the identity's re-encoding — handles both the raw-bytes-never-parsed case and the parses-but-not-a-valid-checkpoint case. Commit 0dfbc24f2.

### Gates
- biome/lint: clean
- vitest: full `appwire-client/typescript` suite 99 files / 2047 tests; mobile-native full suite 83 files / 820 tests

Head: `55d0a602add42636de5b3ac9b1ca023b8addebbf`

obra:
## Round 6 disposition (f8e576fd) — answering the raw panel reviews

(a) **Offline storage reads must handle a throwing port** (member 0/3, both 1791 and #1841's panels): the cold-start probe (`NativePreferencesProvider.tsx`'s effect) and the store-free discard's own re-read called `Storage.getItemSync` (via `nativeKeybindingDrafts(...).load()`) with no guard. Fixed via a new `readDraftOutcome` helper (`nativePreferenceDrafts.ts`) that wraps the read+classify in try/catch and reports `"storageUnavailable"` instead of throwing; both call sites in the provider now go through it and degrade (leave `offlineDraftUnreadable` untouched, set a new `offlineStorageUnavailable` field) instead of crashing the effect or the event handler. Falsified: reverted the helper, 8 new tests fail (`readDraftOutcome`'s own throwing-port test plus the ones depending on the tri-state outcome below); restored, green.

(b) **parseDraftBytes conflated stored JSON null and the stored JSON string "null"** (member 3, both panels): both used to come back as the bare string `"null"`, so `nativeTranscriptDrafts.load()`'s `value === "null"` check silently treated a present, invalid transcript checkpoint as absent. Fixed with two tagged markers (`StoredNullRecord`, `UnparseableDraftBytes`) — `parseDraftBytes` never returns a bare string for either case now. `nativeTranscriptDrafts` checks `isStoredNullRecord` instead of `=== "null"`; `nativeKeybindingDrafts` needed no change (it already classified anything non-`null`/`undefined` as present, so the collision never affected it — verified with a new test: a stored `"null"` string now returns the plain JS string `"null"`, distinct from the marker). Falsified per marker (9 tests: null vs "null" vs garbage across both repositories).

(d) **matchesStoredBytes only normalized whitespace, not key order**: rewrote it (and folded the marker comparison into the same path) to canonicalize via a recursive sorted-keys encoding instead of a byte-for-byte `JSON.stringify` re-encoding compare. New test: `matches two values with the same fields in different key order` — fails under the old `JSON.stringify(JSON.parse(stored)) === JSON.stringify(value)` implementation (confirmed by reverting), passes now.

(c) **Store-free discard passed only a boolean readability, conflating absent and unreadable**: `draftUnreadableAfterDiscard`'s second parameter is now the re-read's full tri-state outcome (`"absent" | "readable" | "unreadable"`, via the same `readDraftOutcome`/`classifyDraftRead`), not `isReadable(load())`. New test: `clears the notice when a refusal's replacement is gone entirely by the time of the re-read` — a third writer removing the record between the port's own re-read and the caller's re-read used to read as "still unreadable" (both give `isReadable(null) === false`); now correctly clears.

(e) Lows:
- `fakeDraftBackend.get()` now clones (`structuredClone`), matching the real Storage-backed backend's re-parse-on-every-read semantics.
- `memoryDraftStorage` (SDK) and `fakeDraftBackend` (native) both used to report a false compare-and-swap match against an absent record when the identity is `null`/`undefined` (`JSON.stringify(null)`/`JSON.stringify(undefined)` collide with an absent store's own comparison) — both now check presence explicitly before comparing.
- `offlineDraftUnreadable` now also recomputes on a connectivity change (`state` added to the effect's deps), not only `hubId`.
- The offline discard action no longer clears the component-local error on a `null` outcome (no hub to discard for) — new `clearsErrorAfterOfflineDiscard` helper, wired into `KeybindingPreferencesScreen`'s onPress.

One fixup surfaced by the fold's own CAS fix (#1808 round 4): `keybindingRecovery.test.ts`'s `failSave`/`allowSave` fixture only toggled `memoryDraftStorage`'s `save()` knob; once an existing draft is classified, `persistDraft` now reaches `replaceIf` instead, so the fixture now fails both.

p6's own non-test line count: **497** (draftCheckpointPort.ts 33, index.ts 8, keybindingsStore.ts 95, testing/draftStorage.ts 7, KeybindingPreferencesScreen.tsx 53, NativePreferencesProvider.tsx 134, nativePreferenceDrafts.ts 161, nativePreferences.ts 6). Test files: 590 lines (draftCheckpointPort.test.ts 114, keybindingsStore.test.ts 122, draftBackend.testkit.ts 12, keybindingRecovery.test.ts 14, nativePreferenceDrafts.test.ts 328).

Gates: vitest (appwire-client/typescript) 1987/1987; vitest (mobile-native) 833/833; `npm run check` (mobile-native tsc) clean; `npm run typecheck` (frontend) clean; `npx biome ci` clean (never run over mobile-native); `npm run lint` clean; `make lint-package-imports` PASS.

obra:
Decomposed after six rounds into #1900 (p6a), #1902 (p6b), #1904 (p6c); closing.

PR #1792, opened 2026-09-18
D6 piece 7: a keybindings draft carries its own ready generation

obra:
## D6 piece 7: a keybindings draft carries its own ready generation

Plan item 7 of the D6 decomposition (`.superpowers/sdd/2026-09-12-mobile-landing-queue-cont/d6-split-plan.md`): "A draft's generation is stamped by its first authoritative payload." Stacks on #1791 (p6).

### Mechanism (adapted from #1549's `10eba05a2`/`dc381b3f5`/`351dc2c6e`)
- `KeybindingsDraft` (local type, distinct from the wire-shaped `KeybindingsOverrides` also used for confirmed hub state and this store's write-path return values): `{ version: 1; revision: number; rules: KeybindingsRule[]; generation: number | null }`.
- `currentGeneration()` helper: `fence.generation >= 0 ? fence.generation : null`.
- `staleDraft`'s third param (`currentGeneration: number`, i.e. `fence.generation`): a non-null generation mismatch is stale regardless of revision, since hub revision numbering restarts on a hub replacement (reconnect to a different hub, or the same hub after a restart) and a coincidentally-identical revision must not read as current.
- Every draft construction site (`restoreDraft`, `editDraft`, `saveDraft`, `rebaseDraft`) stamps `generation: currentGeneration()`.
- `applyHubOverrides` stamps a `generation: null` draft with the live generation as part of accepting the first authoritative payload after it, before judging staleness for that same payload - closing the case where a draft restored before any generation began (a store built synchronously, before the host connects) never gets its generation check activated. Guarded on `fence.generation >= 0` (no call site here reaches this before `beginReadyGeneration` wires the fence, but kept the same shape defensively rather than relying on that staying true).

### Oracle tests
- `a draft restored before any generation began is stamped by the first authoritative payload, so a later hub replacement at the same revision is caught`
- `a draft is flagged stale across a generation change even when the new generation happens to report the same revision`
- Restored `generation: 1` (or `generation: null` for the pre-ready-restore case) to every existing draft-equality assertion the new field now touches (7 sites, not just the 2 the brief named - measured via a full test run after adding the field: `git diff` shows the affected lines).
- Added `draftConflictAfterSettle` to the settle-scenarios table (`saveDraft's post-reply sequence`): a settle that crosses a ready-generation boundary (fenced out by the generation ending, or by support dropping and returning - both begin a new generation) never trusts the draft's revision against it, even though the checkpoint itself settled cleanly.

### Falsification
Reverted only `keybindingsStore.ts` to #1791's head and re-ran `keybindingsStore.test.ts`: 10 failures (the two new tests plus 8 assertions expecting `generation`/`draftConflictAfterSettle` behavior the reverted file doesn't have). Restored with `git apply`; `git status` clean before committing.

## Gates
- `npx vitest run keybindingsStore.test.ts draftCheckpointPort.test.ts` (appwire-client/typescript): 84 passed
- `npx vitest run` (mobile-native, full suite): 797 passed, 83 files
- `make test-api-package`: qualified (no index.ts/tsconfig.build.json touch, so the falsification wasn't required, but ran it anyway since a new exported interface was added; passes)


obra:
## Round fixes (41fa37cee, on top of the fold #1808)

### High: restored drafts use in-memory generation, not persisted identity
Fixed - design decision below.

**Root cause confirmed:** `KeybindingDraftCheckpoint` has no `generation` field at all - nothing is persisted to disk. `restoreDraft` stamped every restored draft with `currentGeneration()` (the CURRENT session's live fence value) unconditionally. For a fresh app instance's construction-time restore, `fence.generation` is `-1` there, so this happened to produce `null` (correct) - but any LATER restore during an already-live generation (`refreshOverrides`'s one-more-restore-attempt, or `discardDraft`'s refuse-and-reclassify after a removal is refused) stamped the CURRENT live generation directly, asserting the checkpoint belongs to it with no actual confirmation. That silently defeats the cross-generation staleness check for exactly the record it exists to protect: `staleDraft`'s generation-mismatch branch never fires once a draft has been (falsely) stamped as matching.

**Design chosen:** clear the generation on every restore, unconditionally, and rely exclusively on `applyHubOverrides`' existing stamp-on-first-authoritative-payload mechanism (already built for the construction-time case). `restoreDraft` never claims a generation on its own now - only an authoritative payload the store has actually just reconciled earns a draft its generation, exactly once.

**Why not persist a stable identity instead** (the alternative the review offered): there is no wire-level identity to persist that would actually distinguish "this hub session" from "a different hub session that happens to report the same revision" - `KeybindingsOverrides`/`fromWireOverrides` carries no hub-instance/boot identifier, and the checkpoint's own `baseRevision` is already what `staleDraft` falls back to (adding it again as "the persisted identity" wouldn't add information). Building a real fix along that path would need the hub to send some form of session/instance identity over the wire - a protocol change well outside an SDK-decomposition piece's scope. I'm treating that as a known, inherent limitation rather than pretending to solve it: **the very first authoritative payload after a genuine restart can still coincidentally match a stale draft's revision and read as current.** That gap is unchanged by this fix (it was never actually protected - the construction-time stamp always tautologically matched itself); what this fix closes is the regression where **later** restores (well after the first payload, mid-session) could re-open that same hole by re-asserting a stale draft as current.

Added `a storage-recovery restore during a live generation defers its generation stamp to the next authoritative payload, so a later generation change still catches it stale` - restores a draft during a live generation (simulating a previous session's or a concurrent writer's record), shows it comes back `generation: null` (not silently matching the live one), then shows a SUBSEQUENT generation change still correctly catches it once it has actually earned a generation from an authoritative payload. Also restored `generation: null` on the five existing "adopts the replacement"/"re-classifies" assertions, which exercise this same `restoreDraft` storage-recovery path and were asserting the (buggy) `generation: 1`.

### Medium: `saveDraft` races with in-flight `patchOverrides`
Fixed in #1808 (the fold), inherited here through the restack: `directWriteInFlight`, set for the duration of a `patchOverrides` request; `saveDraft` refuses while it is true.

### Medium: `discardClassified`/`replaceClassified` can delete or overwrite a record never classified
Fixed in #1808, inherited here: both now classify via a fresh `load()` before acting on a never-classified repository; a readable record refuses instead of being destroyed/overwritten blind.

### Medium: `discardStoredDraft` casts arbitrary raw bytes to `Checkpoint`
Fixed in #1808, inherited here: identity arguments widened to `unknown` throughout.

### Medium: `KeybindingDraftStorage` contract is a source-breaking, undocumented change
Refuted in #1808 (implementer list), inherited here. Additionally fixed here: exported `KeybindingsDraft` from the package root (the type of the already-exported `KeybindingsStoreFields.draft` field).

### Medium: malformed JSON bytes lost, making `discardDraft()` repeatedly refuse removal
Fixed in #1808 (`NativePreferencesProvider`'s backend returns unparsable bytes raw, and `matches()` compares the raw string too), inherited here.

Fixed in #1808, inherited here.

### Low: unreadable-discard action appears enabled but does nothing while disconnected
Fixed in #1791's own round (restacked underneath this PR): the provider now exposes a store-free discard path.

### Low: comments encode review-round history
Fixed in #1808's carried p5 round, inherited here.

## Gates
- `npx vitest run keybindingsStore.test.ts draftCheckpointPort.test.ts` (appwire-client/typescript): 89 passed
- `make test-api-package`: qualified (touched index.ts to export `KeybindingsDraft`)

## Falsification
Reverted only the `restoreDraft` change to this branch's prior head (`331fa4fab`) and re-ran `keybindingsStore.test.ts`: 6 failures as expected (5 existing assertions expecting `generation: null` receive `1` instead, plus the new dedicated test). Restored with `git apply`; `git status` clean before committing.

obra:
## Round 4 fixes (3762239fa)

**High — editing an existing draft launders a cross-generation conflict** (`keybindingsStore.ts:1289-1300`, `editDraft`)
`editDraft` always restamped `generation: currentGeneration()`, so an edit to an already-conflicted draft (composed against a previous hub session) silently re-confirmed it valid for whatever generation is live now, clearing `staleDraft`'s generation check without the required `rebaseDraft`.

**Fix:** `editDraft` now preserves the existing draft's generation stamp (including `null`) when a draft already exists; only a brand-new draft (no existing one) takes the current generation. Only `rebaseDraft` (explicit review) or an authoritative payload landing (`applyHubOverrides`' existing stamp-on-first-payload) may re-earn the stamp.

**Test:** "editing an already-conflicted draft keeps it stale until rebased, instead of silently re-stamping the live generation" — edits a draft, forces a generation change at the same coincidental revision (the existing sibling test's own scenario) to get `draftConflict: true`, edits the still-conflicted draft again and asserts the generation is unchanged and the conflict survives, then `rebaseDraft` and asserts the generation changes and the conflict clears.

**Medium (offline discard)** — this is #1791's own finding, fixed there ([round 4 comment](https://github.com/prime-radiant-inc/evener/pull/1791#issuecomment-5726308108), pushed as `031528799`): `KeybindingPreferencesScreen.tsx` now branches on `model && preferences.connected` rather than on `model` alone. Flows into this PR on restack.

**Lows** — also #1791-fixed in the same push and inherited via restack:
- "successful store-free discard leaves the stale restore-failure error visible": `NativePreferencesProvider.tsx`'s offline-discard snapshot update now always sets `error: null` alongside `draftUnreadable`/`storageUnavailable`.
- "discardStoredKeybindingDraft returning false leaves a stale unreadable notice": the snapshot update is no longer gated on `if (removed)` — every outcome (`removed`/`absent`/`refused`, from `discardStoredDraft`'s new 3-state result) refreshes it.

**Falsification:** reverted `keybindingsStore.ts` to base (`git diff | git apply -R`), ran the new test alone - failed (`generation: 3` instead of the preserved `1`) - restored, `git status` clean.

**Gates:**
- vitest `keybindingsStore.test.ts`: 82/82

Non-test diff: 15 insertions / 2 deletions in `keybindingsStore.ts`.

obra:
## Round 5 disposition (a91612a) — answering the raw panel reviews

This piece's own diff (`keybindingsStore.ts`/`.test.ts`, `index.ts`) adds the draft's ready-generation stamp; every finding in the raw panel review (reviews/raw/1792-f33001191.md) is the p6 (#1791) code seen through the stack — `draftCheckpointPort.ts`, `nativePreferenceDrafts.ts`, `NativePreferencesProvider.tsx` — not this piece's own commits. Pointers only, fixed once at the owning piece:

- Stored JSON `null` breaking transcript drafts (`parseDraftBytes` shared with the transcript port) — fixed in **#1791**, commit 142274e32.
- `discardStoredDraft` reporting a `removeIf` refusal as `"absent"` — fixed in **#1791**, commit 282b22ecb (plus the provider's `draftUnreadableAfterDiscard` follow-up, commit 5edf6f7d9).

No own finding was identified against this piece's actual diff (the generation-stamp addition itself). The restack (rebasing this branch onto the fixed #1791) picks up both fixes automatically.

### Gates
Same suite as #1791 (this branch is #1791 + 3 commits): typecheck/biome/lint clean, vitest green.

Head: `a91612a127b2244a48a4160436e1a5dc5f8bf104`

obra:
## Round 6 disposition (62fc5a70) — answering the raw panel reviews

**The refused-settlement "saving stays true" finding** (member 0, `keybindingsStore.ts:1397-1402` at this PR's reviewed head a91612a12): measured against the current code before touching anything — `settleWrite`'s cleanup-refusal path (`refused = restoreDraft(getState())`) runs AFTER `settled` (which always includes `saving: false`) has already been published via `setState(settled)` or `applyHubOverridesSettling(payload, settled)`. `restoreDraft`'s returned partial never mentions `saving`, so the second `setState(refused)` merge leaves the already-published `false` untouched. Added a temporary assertion to the existing test `a successful save whose checkpoint was replaced while the PATCH was in flight adopts the replacement` (`expect(store.getState().saving).toBe(false)`) — passed immediately, on this PR's own head, with no code change. Reverted the probe (not a real fix, just a measurement) rather than leave an unrelated assertion in this PR's diff.

Conclusion: no fix needed here or in the fold — the ordering that makes this correct (`settled` publishes before the CAS-refusal branch runs) was already in place at this PR's own commits, byte-identical to the current chain tip. Nothing to route elsewhere.

Own item: this PR introduces the draft's `generation` stamp (`editDraft` preserves it; `restoreDraft` never stamps the live generation itself). One test added upstream on the fold (#1808 round 4, the persist-CAS concurrency test) needed `toMatchObject` instead of `toEqual` once `draft` started carrying `generation` — fixed here as `test(sdk): the persist-CAS test tolerates the draft's own generation stamp`, since this piece is where that field first appears.

Gates: vitest (appwire-client/typescript) 1991/1991; vitest (mobile-native) 833/833; typecheck clean both sides; `npx biome ci` clean; `npm run lint` clean; `make lint-package-imports` PASS.

obra:
## Restack (1d75712)

Rebased onto p6c's tip (568143b0, the fold + p6a/p6b/p6c stack, decomposed from #1791 into #1900/#1902/#1904). Own-diff content verified byte-identical to the pre-restack diff (`git diff f8e576fdf..62fc5a70b` vs `git diff 568143b05..1d75712` on the +/- content lines, ignoring hunk-header line-number shifts). One follow-up commit (1d75712) beyond the rebase itself: a round-5 test added on the fold after this piece last touched `keybindingsStore.test.ts` predated the generation-stamp field and needed the same `toEqual`→`toMatchObject` tolerance this piece's own last commit already applied elsewhere in the same file.

Gates: mobile-native vitest (840) and `npm run check` clean; appwire-client/typescript + web vitest (1997) clean; `npm run typecheck`/biome/lint clean; `make lint-package-imports` PASS; `make test-api-package` qualified (index.ts touched, no module moved).

obra:
## Restack on p6c (41fc2f2e)

Restacked on p6c's new tip (#1904 @ 3c6436d6); own 5 commits replayed clean, no conflicts, no downstream fix needed. No round-2 findings assigned here beyond the alias-default finding already fixed and disposed at #1900 (which this branch inherits).

Head: `41fc2f2e8408b601c5e3b73fdd5ef360046e3d42`

Own diff vs p6c's new tip: 3 files, 266 insertions / 24 deletions (unchanged content, just the new base).

Gates (targeted, all green): full appwire-client/typescript vitest suite (2030 tests); `npm run typecheck`; `npx biome ci ../../../appwire-client/typescript`; `npm run lint`.

Pushed with `--force-with-lease=claude/sdk-d6-p7-draft-generation-stamp:1d7571223`.

obra:
This capped review branch is superseded by smaller, independently reviewed pieces: generation stamps merged in #2009; the generation conflict fence is #2014; identity-aware storage reload is implemented and reviewed at c99c7205601463d2198f6b29519fbb195eece6e3 and will be published after the fence lands. The final reload requirement remains open in the migration ledger, as do dependent #1841, #1844, and #1845.

Closing this original PR to keep one active review path. Its branch and backup commits are preserved; this closure does not claim the full settings migration is complete.

PR #1796, opened 2026-09-18
fix(plugins): sweep a stranded clone under a directory-sourced marketplace (#1053)

### LINKED ISSUE #1053: A directory-sourced marketplace can strand its old clone, and neither rename nor removal sweeps it
`internal/plugins/marketplaces.go`, `EditMarketplace` step 3: when a re-source moves a marketplace from a git source to a directory source, the old clone under `<marketplaces>/<name>` is removed only after both files are saved, best-effort (`_ = marketplaceRemoveAll(clone)`). If that removal fails the clone stays.

From then on the entry's source kind is `directory`, and both paths that could sweep the directory skip it for that kind: the rename step moves `<marketplaces>/<old>` only for non-directory sources, and `RemoveMarketplace` deletes `<marketplaces>/<name>` only for non-directory sources. The stale clone then sits under a name nothing records, and `refuseLeftoversUnder` later reports it as a removed marketplace's residue, refusing a rename onto that name until someone deletes it by hand.

Pre-existing before #1021 (`RemoveMarketplace` had the same gate on main); surfaced by a re-review during #1021 and left out of scope there.

**Fix shape:** on rename and on removal, sweep `<marketplaces>/<name>` whenever it exists, regardless of the recorded source kind. A directory source never lives there (its install location is its own path), so the directory can only be a stale clone. Alternatively, report the failed removal with the path instead of discarding the error, so the user knows what to clear.

**Test:** a git-sourced marketplace re-sourced to a directory with an injected `marketplaceRemoveAll` failure on the old clone, then removed (and, separately, renamed): the directory under the old name is gone afterwards.

obra:
## Root cause

`EditMarketplace` step 3 removes the clone a git→directory re-source makes
redundant only after both store files are saved, and only best-effort
(`_ = marketplaceRemoveAll(clone)`). If that removal fails, the clone is
stranded under `<marketplaces>/<name>` while the entry's source kind is now
`directory`.

Both later paths that could sweep that directory were gated on a non-directory
source kind:

- `moveMarketplace` moved `<marketplaces>/<old>` only for non-directory
  sources, so a rename left the stranded clone under the old name.
- `RemoveMarketplace` deleted `<marketplaces>/<name>` only for non-directory
  sources, so removal left it behind.

The residue then outlived the marketplace under a name nothing records, and
`refuseLeftoversUnder` reported it as a removed marketplace's residue —
refusing a rename onto that name until someone deleted it by hand.

## Change

Sweep `<marketplaces>/<name>` whenever it exists, regardless of the recorded
source kind. A directory source's install location is its own path (and
`refuseSourceInStore` keeps it out of the store), so a directory under the
store's canonical name can only be a stale clone.

- `RemoveMarketplace` now removes it unconditionally (still warning, with the
  path, on failure) instead of only for non-directory sources.
- `moveMarketplace` now detects a clone under the old name for every kind. A
  non-directory source still moves it with the name; a directory source removes
  it, rather than parking the same unrecorded directory under the new name.

## Test evidence

Two new tests in `internal/plugins/marketplaces_test.go` first reproduce the bug
by re-sourcing a git marketplace to a directory with `marketplaceRemoveAll`
injected to fail on the old clone, then exercise each sweep:

- `TestRemoveMarketplace_SweepsAStrandedCloneUnderADirectorySource`
- `TestEditMarketplace_RenameSweepsAStrandedCloneUnderADirectorySource`

Before the fix both failed with the clone still present under the old name;
after, both pass.

```
go test ./internal/plugins/ -count=1   # ok
gofmt -l internal/plugins/marketplaces.go internal/plugins/marketplaces_test.go  # clean
go vet ./internal/plugins/  # clean
```

Fixes #1053


obra:
Roborev finding (High, `marketplaces.go:265-268` and `:521-530`) — **fixed** in 8941bb3.

The concern: sweeping `<store>/marketplaces/<name>` whatever the source kind can delete a live directory source, for a record the current store would not write (pre-`refuseSourceInStore`, or hand-seeded).

Resolution: both sweeps now consult the recorded ref before deleting. New helper `sweepWouldDeleteSource(ref, clone)` returns true when the entry is directory-sourced and its resolved `Source.Path` overlaps the canonical clone path in either direction (the clone path is the source, holds it, or sits inside it). When it does, the directory is the live source and is left alone:

- `RemoveMarketplace` skips the removal for such a record.
- `moveMarketplace` skips the stale-clone removal for such a record.

Non-directory sources are unchanged (never protected), so the #1053 strand is still swept.

Tests added (seeded-store records, no git needed):
- `TestRemoveMarketplace_SparesADirectorySourceThatIsTheClonePath`
- `TestEditMarketplace_RenameSparesADirectorySourceThatIsTheClonePath`

Both seed a directory-source record whose `Source.Path` *is* `marketplaceDir(name)` with a sentinel inside, then assert the sentinel survives removal/rename. `go test ./internal/plugins/ -count=1` is green.

**Also in this push:** error unrelated to #1053 but blocking it — main is broken by a merge race (#1754 landed a `RepoLocation` test importing `gateRequests` 19s after #1755 renamed that helper to `gateSettlements`). That fails `static-build` and `web` on every PR. Fixed in 9d1efe4 (`RepoLocation.test.tsx`: import `gateSettlements`, settle via `Settlement.resolve`); `npx vitest run RepoLocation.test.tsx` and `tsc --noEmit` pass locally.

obra:
Disputed High finding — **determination: reachable, guard kept** (8941bb3, comment refined in 6ab7782).

I traced every path that can write a directory-source record:

- `AddMarketplace` (`marketplaces.go:187`) and `EditMarketplace` resourcing (`:372`) are the **only** callers of `refuseSourceInStore`.
- `refuseSourceInStore` landed with **#1021 (045825057)**; nothing enforced the invariant before it.
- The name migration (`marketplace_migration.go`) never calls `refuseSourceInStore` and never rewrites `Source.Path` — `recordedRename` carries `ref.Source` verbatim (`:163`), and `migrateMarketplaceName` preserves it. Its `ref.Source.Kind != SourceDirectory` branches touch only `InstallLocation`.
- `ensureFetched`/`RefreshMarketplace` do not change `Source`.
- The store file is hand-editable JSON, so a seeded record needs no code path at all.

So a directory source whose `Source.Path` is the store's clone path can exist (pre-#1021 record, hand-seeded file) and survives every later store write. The finding is real, not hypothetical — the guard is warranted.

Resolution (already in this PR): `sweepWouldDeleteSource(ref, clone)` resolves the recorded `Source.Path` and the clone path and returns true when they overlap in either direction; `RemoveMarketplace` and the rename branch of `moveMarketplace` skip the sweep then. Resolving at sweep time also catches a source that an after-the-fact symlink moved inside the store. Non-directory sources are never protected, so the #1053 strand is still swept.

Seeded-store tests (no git): `TestRemoveMarketplace_SparesADirectorySourceThatIsTheClonePath`, `TestEditMarketplace_RenameSparesADirectorySourceThatIsTheClonePath` — each seeds `Source.Path == marketplaceDir(name)` with a sentinel and asserts it survives removal/rename.

Evidence: `go test -race ./internal/plugins/ -run 'TestRemoveMarketplace|TestEditMarketplace|TestAddMarketplace' -count=1` → ok.

obra:
Roborev round 2 on the guard — **both addressed** (dd32b0487).

**High — guard only checked the swept marketplace's own source.** Fixed: `sweepDestroysSource(mk, clone)` now scans *every* registered marketplace's directory source, so a legacy/hand-seeded store that records one marketplace against another's clone path is protected too. `moveMarketplace` now takes `mk` (threaded from `EditMarketplace` and both migration call sites).

**Medium — containment model did not match `RemoveAll`.** Fixed. The guard now models what the sweep actually deletes (the tree at the clone path):

- protects a recorded `Source.Path` at or beneath the clone, checked *without* resolving, so a symlink beneath the clone (which the sweep deletes as a link, even when it resolves outside) is protected;
- protects a source resolving inside the clone (its target is deleted);
- **stops** protecting an ancestor of the clone, which `RemoveAll` never deletes — so an ancestor source no longer holds a stale clone and blocks name reuse.

Also, since a git marketplace's rename *moves* its clone with the name, that move would displace a directory source another record keeps there; the rename now refuses instead of stripping it.

Regression tests (each fails against the prior model):

- `..._SweepsWhenADirectorySourceIsAnAncestor` and `..._RenameSweepsWhenADirectorySourceIsAnAncestor` — ancestor no longer over-protects (both failed before: stale clone survived).
- `..._SparesASourceSymlinkedOutOfTheClone` and `..._RenameSparesASourceSymlinkedOutOfTheClone` — symlinked source beneath the clone preserved (both failed before: swept).
- `TestRemoveMarketplace_SparesAnotherMarketplacesSourceAtTheClonePath`, `TestEditMarketplace_RenameSparesAnotherMarketplacesSourceAtTheClonePath`, `TestEditMarketplace_RenameRefusesToMoveAnotherMarketplacesSource`.

Evidence: `go test -race ./internal/plugins/ -run 'TestRemoveMarketplace|TestEditMarketplace|TestAddMarketplace' -count=1` → ok; `go test ./internal/plugins/ -count=1` → ok.

Issue #1780, opened 2026-09-18
hub background plugin managers use root "" while the server's manager uses cfg.PluginRoot


obra:
Found by the /simplify review of #1733: cmd/evener-hub/main_background.go builds three plugin managers with plugins.NewManager("") (lines ~26, ~80, ~95 at head acd319572) while the hub's own server manager is built from cfg.PluginRoot (app_rpc.go). Pre-existing before #1733. Reusing the hub's single manager for the background paths would fix the mismatch but changes behaviour in sandboxes and tests, so it was not folded into #1733. Decide whether the background managers should share the server's root (and the single wired Manager), then do it in one PR with a test that the background paths see the same store the server serves.

Issue #1734, opened 2026-09-18
hub: thread-spawn and slash-catalog resolvers build their own plugin Manager without the store-change hook


obra:
## Where

`cmd/evener-hub` `hubCommandList` / `hubResolvePlugins` (thread-spawn and slash-catalog paths), after the plugin store post-commit hook PR (#1733).

## What

#1733 installs `Manager.OnStoreChanged` on the plugins controller's Manager and both auto-upgrade Managers, so every write through those broadcasts by construction. `hubCommandList` and `hubResolvePlugins` construct their own Manager per call and are not wired: a legacy-name migration triggered from a thread spawn or a slash-catalog resolve still writes without a broadcast (the `resolveForLaunch discards StoreChanges` finding from #1602's last review). Wiring `hubResolvePlugins` means changing a package-var signature that a dozen-plus tests substitute directly, so #1733 left it out rather than widen.

## Proposed change

Give those paths the hub's already-wired Manager (or install the same callback on the per-call Manager they build) and update the substituting tests once. About 40 lines plus test churn.

Flagged by the #1733 lane as a residual.

obra:
Scoped down after #1733 round 2: hubCommandList is now wired in #1733 (integration test: a legacy-name migration on a command-list request broadcasts). Only hubResolvePlugins remains: hubThreadStart has 38 call sites across 11 files and two test files substitute the package var directly, so routing it through the wired Manager (or a factory that applies wirePluginStoreBroadcast) is its own small PR with that test churn.

PR #1798, opened 2026-09-18
plugins: one wired manager per hub, hook is the sole notifier (#1733 follow-up)

obra:
## Summary

Follow-up to #1733 (merged as b468560ed with one Low finding). Applies the eleven `/simplify` findings from `docs/superpowers/plans/2026-09-12-mobile-landing-queue-cont/reviews/1733-simplify.md`, plus the one Low from #1733's own final RoboRev round (the same finding as item 9 below).

1. `hubCommandList` now resolves plugins through `hubResolvePlugins` instead of building and wiring its own per-call `*plugins.Manager`.
2. `hubcore.WebConfig.PluginResolveManager` (a `func(string) *plugins.Manager` resolver closure) is replaced with a plain `PluginManager *plugins.Manager` field; `hubResolvePlugins` takes the pointer directly and falls back to `plugins.NewManager(pluginRoot)` only when nil.
3. `registerPluginAutoUpgradeHandlers` now shares `pluginsController.mgr` instead of a second `autoUpgradeMgr` on the same root.
4. `newWebServer` constructs the one `*plugins.Manager`, sets `cfg.PluginManager`, and wires it (`wirePluginStoreBroadcast`) after the server exists; the constructor (`newHubAppServerWithNavigationAndTrace`) no longer returns `cfg` as a third value or wires anything itself. `TestNewWebServer_TwoServersEachResolveTheirOwnPluginManager` is simplified to read the injected pointer directly.
5. Adds one `newWiredPluginManager(root, broadcaster)` helper for the three `main_background.go` sites that used to construct-then-wire a throwaway Manager by hand (the root each site uses is unchanged — that mismatch is tracked separately as #1780).
6. (moot once 2 is done — the resolver closure that captured `cfg` by value is gone.)
7. `resetStoreChanged` existed only to defend `plantLegacyMarketplace` (a test fixture that writes outside any `lockStore` session); the fixture now clears `m.pendingStoreChanged` itself, and the method plus its `lockStore` call site are deleted.
8. `captureStoreChanged` + `storeChangedCallback` (two separate `storeChangedMu` acquisitions for one release) are merged into one `takeStoreChanged()` under a single lock; `StoreChanged.any()` is inlined at its one caller.
9. Fixes `app_plugins.go`'s `hubPluginsController` comment, which claimed Manager's `StoreChanged` state "needs no mutex of its own" because it's "only touched while the flock is held" — the exact claim `paths.go`'s `storeChangedMu` comment refutes. This is also #1733's own Low finding from RoboRev.
10. Strips review-round narration ("is Medium 1's regression test...", "is issue #1734's own case, folded into this PR...", "#1602 review round") and a "to renamed" typo out of the touched doc comments, down to the invariant each states.
11. Adds `assertBroadcastMethods`/`assertOneBroadcast` next to `recordingBroadcaster` and uses them in place of four repeated "exactly one broadcast" checks and three repeated `foundMarketplace`/`foundPlugin` loops.

Non-test diff: 9 files, +107/-155 (net -48 lines). Full diff: 19 files, +201/-280.

## Test plan

- [x] `$(go env GOROOT)/bin/gofmt -l` on touched files — clean
- [x] `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...` — clean
- [x] `go test -count=1 ./cmd/evener-hub/... ./internal/plugins/...` (plain and `-tags evenerfuzz`) — all green
- [x] `golangci-lint run ./cmd/evener-hub/... ./internal/plugins/...` — 0 issues
- [x] `-race` on `TestManager_ConcurrentLockSessionsDoNotRaceStoreChanged` after merging the two `storeChangedMu` reads into `takeStoreChanged` — clean

Refs #1733


obra:
## Round 1 disposition (cda6ee3)

- **Medium** (`main_background.go:26,79,93` construct separate managers rooted at `""` instead of reusing `web.cfg.PluginManager`): deliberate non-change, tracked separately as #1780 (which of the three roots the fix should use once decided). Not touched here.

- **Low `internal/plugins/paths.go:31`** (comment names deleted functions `resetStoreChanged`/`captureStoreChanged`): reworded to name the surviving operations — `markStoreChanged`, `takeStoreChanged`, `OnStoreChanged`.

- **Low `cmd/evener-hub/web.go:161`** (production wiring has no test that fails if deleted): added `TestNewWebServer_WiresPluginStoreBroadcastToItsOwnServer` — drives the real `NewWebServer` path (no injected Manager, no injected broadcaster), adds a directory marketplace over a real client connection, and waits on `client.Notifications()` for `evener/marketplace/updated`. Watched it fail (5s timeout) with the wiring line temporarily commented out, then restored the line and confirmed green.

- **Low `cmd/evener-hub/app_rpc.go:514`** (nil fallback builds an unwired Manager for servers built directly through `newHubAppServer`/`newHubAppServerWithNavigation`): measured the two options. Requiring every caller to supply an already-wired manager touches ~100 direct call sites across the test suite; wiring the fallback in place (`wirePluginStoreBroadcast(mgr, server)` right where it's constructed, since `server` already exists at that point) is a 4-line change. Took the smaller fix.

Gates (root module): `$(go env GOROOT)/bin/gofmt -l` on touched files — clean. `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...` — clean. `go test -count=1 ./cmd/evener-hub/... ./internal/plugins/...` plain and `-tags evenerfuzz` — all green. `golangci-lint run ./cmd/evener-hub/... ./internal/plugins/...` — 0 issues.

obra:
Round 1 @cda6ee3: CI green (15), RoboRev has two Lows only (background maintenance sites still mint their own manager, which is #1780 by design and the contradicting comments in web.go/config.go get fixed in the follow-up; the nil-fallback wiring branch has no test). Merging per the Lows rule; follow-up PR opens immediately with the comment fix and the fallback test.

PR #1799, opened 2026-09-18
fix(test): harden cgroup mount resolution for delegated roots

### LINKED ISSUE #1193: load-aware-workers: harden cgroup mount resolution for delegated roots and escaped paths
Follow-up from PR #1176 (load-aware worker sizing), roborev review of `316a4da`.

Two edge cases remain in `scripts/lib/load-aware-workers.sh`'s cgroup quota resolution:

1. `load_aware_cgroup_mount` returns only the mount point, and the walk joins the membership path from `/proc/self/cgroup` directly against it. For a delegated or non-root mount without a cgroup namespace, the membership path can be hierarchy-absolute rather than namespace-relative, so the join misses the quota files and overstates available CPUs. The robust fix is to consider both interpretations and take the most restrictive finite result.
2. mountinfo escapes spaces, tabs, and backslashes as octal sequences (`\040` etc.), and the mountpoint is currently split unquoted. A cgroup mount under a path containing those characters would not be found.

Note the reviewers gave contradictory guidance on the `root` field across rounds (round 4 said the boundary is the mount point; round 6 said ignoring `root` is wrong for delegated mounts), so any fix should carry fixtures for both namespace-relative and hierarchy-absolute memberships. Impact is bounded: worker counts are clamped by each caller's CAP and every entry point falls back to fixed historical values.

obra:
## Root cause

`load_aware_cgroup_cores` resolved the cgroup mount with a single reading of `/proc/self/cgroup`:

1. **Delegated roots.** The membership path is namespace-relative when a cgroup namespace is in effect, but hierarchy-absolute otherwise (a delegated or non-root mount). Joining the whole path against the mount point names a directory that does not exist, so the quota walk climbs past the finite limit and reports more CPUs than the quota allows. That overstatement defeats the back-off this library exists for, because `cores - load` never drops below the caller's ceiling however loaded the machine is.
2. **Escaped paths.** mountinfo escapes a space, tab, newline, or backslash in a path as `\040`, `\011`, `\012`, or `\134`. The mount point was used undecoded, so a cgroup mount under such a path was never found at all.

The reviewers gave contradictory guidance on the `root` field across rounds, so the fix carries fixtures for both the namespace-relative and hierarchy-absolute readings instead of committing to one.

## Change

- Decode mountinfo fields 4 (root) and 5 (mount point); add `load_aware_cgroup_mount_root` and keep `load_aware_cgroup_mount` as the decoded mount point.
- Evaluate both membership readings — namespace-relative (mount point + path) and hierarchy-absolute (mount point + path with the mount's root stripped) — and keep the most restrictive finite quota.
- Extract the walk into `load_aware_cgroup_walk_cores` and stop splitting the mount point unquoted, so a decoded path containing whitespace survives.

## Test evidence

`go test . -run LoadAware -count=1` → `ok primeradiant.com/evener`
`go test . -count=1` → `ok primeradiant.com/evener 116.5s`
`gofmt -l loadawareworkers_test.go` → clean; `go vet .` → clean.

New fixtures, all failing on the parent commit with the overstated result (`""` or `"4"` where `"1"`/`"2"` is correct):

- `TestLoadAwareCgroupAbsoluteMembership` — a finite limit reachable only the hierarchy-absolute way, and the most restrictive of both readings winning.
- `TestLoadAwareCgroupEscapedMountPath` — space, backslash, and tab in the mount point, plus an escaped mount point combined with an absolute membership.
- `TestLoadAwareCgroupMountDecodesEscapes` — the decoded mount point names a real directory.

Fixes #1193


obra:
Addressed in ab341c02a.

Command substitution strips trailing newlines, so a decoded mountinfo path ending in `\012` was truncated before the walk. `load_aware_cgroup_hierarchy_cores` now carries each decoded path out of its capture with a trailing sentinel and removes exactly one byte (`"$(...; printf X)"` then `${v%X}`); appending one and stripping one preserves the value whether or not it ends in a newline, so a newline-terminated mount point (or root) now names the directory it should.

Producers that emit a line terminator (awk, sed, dirname) deliberately get no sentinel: the newline stripped there is the terminator, not part of the field, and those fields cannot themselves hold a newline.

Added `TestLoadAwareCgroupTrailingNewlinePaths` with fixtures for a newline-terminated mount point on its own, with a hierarchy-absolute membership, and one that exercises the climb to the mount point. `go test . -run LoadAware -count=1` and the full `go test .` pass; `gofmt -l` and `go vet .` are clean.

PR #1803, opened 2026-09-18
feat(tui): set the terminal title to the active session's display name (#825)

### LINKED ISSUE #825: TUI: set terminal window title to the active session's display name

obra:
Fixes #825

## Root cause

The TUI never emitted a terminal window title. `cmd/evener-tui` had no
`tea.SetWindowTitle` call anywhere, so the terminal/tmux window running
`evener-tui` stayed unlabelled even though the webui sidebar shows each session's
display name. The session name is human-visibility only, making it a natural fit
for the window title.

## Change

- `hub_window_title.go` (new): `windowTitleCmd` returns a `tea.SetWindowTitle`
  command when an update moves the title, and records the last title set so the
  OSC escape stays off the hot path. The title is the active session's display
  name while a session view is open, and empty on every other surface so leaving
  a session clears it. It moves only on a view transition (entering/leaving a
  session) or a change to the viewed session's display name.
- `hub_model.go`: `Update` wraps the update's command with `windowTitleCmd`,
  so every entry path (dashboard select, spawn, fork, live-nav, reconnect
  recovery) is covered by one mechanism rather than per-site wiring.
- `hub_notifications.go`: `evener/thread/name/changed` now folds the new name
  onto the cached session detail, so the header and the terminal title follow
  the frame live — including the auto-namer's first name and its compaction
  refreshes. `evener/thread/name/changed` moves from
  `notifyMethodsDeliberatelyIgnored` to `notifyMethods`; the coverage test that
  keeps the catalog and the TUI's response in sync enforces this.
- `hub_session_view.go`: the session display-name fallback chain (name →
  original prompt → id → ref) is shared via `sessionDisplayName()` between the
  header and the title.

tmux note: tmux only reflects OSC title escapes for a window with
`allow-rename on`; with the default (`off`) the title will not stick there. This
is documented at `windowTitleCmd`.

## Tests

`cmd/evener-tui/hub_window_title_test.go` asserts deterministically that the
model returns a `tea.SetWindowTitle` command:

- when the session view activates,
- when `evener/thread/name/changed` arrives (compaction-refresh name),
- when the view returns to the dashboard (clears the title),
- fallback chain (name → session id → ref),
- no re-emit on an unrelated update (keypress/streaming frame),
- no title outside a session view.

`TestHubDashboardSpawnWaitsForSlowHubSpawn` was tightened from "no command" to
"exactly the session-title command", preserving its no-unexpected-RPC guarantee.

## Verification

```
$ go test ./cmd/evener-tui/ -count=1
ok  primeradiant.com/evener/cmd/evener-tui

$ gofmt -l cmd/evener-tui/        # (no output)
$ go vet ./cmd/evener-tui/...     # exit 0
```


obra:
### Local (per-commit) roborev triage

roborev reviews each local commit, not just the PR head. Triage of the open
reviews on this branch, newest findings first (substantive ones all appear in
the last two reviews, `18781` and `18366`):

**Fixed this round**
- *Unsanitized `SessionID`/`Ref` fallback (escape injection in header/chrome)*: `sessionDisplayName` now sanitizes every candidate in the fallback chain, not just the reported name, so the header, chrome, and dashboard row receive stripped text. Test: `TestSessionDisplayNameSanitizesIdentityFallback`.
- *Rename with empty identity relabels the viewed session*: `notificationMatchesCurrentSession` treats an empty `ref`/`threadId` as a match; the name-changed handler now requires a real identity before assigning `detail.Title`. Test: `TestThreadNameChangedWithoutIdentityDoesNotRelabelViewedSession`.
- *Title not cleared on signal-driven exits*: bubbletea's signal handler pushes `QuitMsg`/`InterruptMsg` without calling `Update`, so `quitCmd` never runs. `run()` now writes the OSC clear after `Program.Run` returns, on every exit. Tests: `TestCovRun_ProgramErrorReturns1` and `TestCovRun_SuccessReturns0`.
- *`requireQuitCommand` accepted a bare `tea.QuitMsg`*: it now requires the ordered clear-then-quit sequence and rejects a bare quit, so a regression back to `tea.Quit` fails.
- *`windowTitleCmds`/`requireOnlyWindowTitle` could pass duplicates*: the helper now collects every title message and `requireOnlyWindowTitle` requires exactly one with no other command.

Earlier rounds already fixed the C1 control gap (`unicode.IsControl`), the dashboard rename behind the mode gate, and the tracked `node_modules` symlink plus its `.gitignore` gap.

**Refuted**
- *Out-of-order `SetWindowTitle` via `tea.Batch`* (`18781`, `18366`, `17586`): no deterministic reproduction exists. The title command performs no I/O — it constructs a value — and bubbletea applies `setWindowTitleMsg` on its render loop (it never reaches `Update`). A stale reorder requires two title-changing updates in flight whose no-work commands complete out of order; each title change is driven by a distinct ordered event, and the earlier command completes before the later event can be produced. The quit-time clear is not batched with a title command on the session-quit route (`windowTitleCmd` returns nil there), so it cannot be overtaken.
- *Rename lost to an in-flight `ThreadList` snapshot / status refresh reverting a renamed title* (`18781`, `18366`): the same snapshot-replaces-local-state race already exists for the pre-existing `updateDashboardRowModel` and the shared-notes push (`NotifyEvenerNotesUpdated` then a status-refresh detail replace), i.e. it is the established pattern of this client, not something this feature introduces. It cannot be closed inside the title feature without adding ordered-frame capture to tree/status fetches, which is a separate change to the snapshot plumbing.

`17881` reads "No review output generated" — a review-infrastructure failure, not a pass; left open.

obra:
### Local review `19805` (range synthesis) — refuted finding

*Dashboard rows stale for sessions the TUI is not subscribed to* (`hub_notifications.go`): the TUI subscribes to the viewed session (and its children) only, so a rename of a background/other session never reaches `applyHubNotification` at all. That is a property of the subscription scope, not of the rename handler — the handler now updates the row and tree for every rename frame the connection delivers, including while the dashboard is showing. Before this feature, no rename updated the dashboard ever, so this is strictly an improvement, and closing the residual gap means either a hub-wide rename broadcast or changing the deliberately-ignored navigation-invalidated decision — a separate change to the notification/subscription plumbing, not the terminal-title feature. No deterministic reproduction is possible inside the TUI because the frame is never delivered.

PR #1804, opened 2026-09-18
refactor(web): share one connection client port across stores (#1464)

### LINKED ISSUE #1464: web stores: one shared connection-store client port instead of nine inline requireClient copies
Found during /simplify on the D2 PR (hub overview store core).

Each web store that reads the current client from `stores/connection.ts` carries its own copy of the body "read `connectionStore.getState().client`, throw `<name> store: no client connected; call useConnectionStore.getState().connect(client) first` when null": `stores/threads.ts`, `agentsDoc.ts`, `commandCatalog.ts`, `hubUpdate.ts`, `credentials.ts`, `launchConfig.ts`, `extensions.ts`, `settingsOverview.ts` (the D2 adapter now hands that body to the package as a `request`-only port), plus `currentClient()` helpers in `keybindings.ts` and `transcriptDisplay.ts`, each with a store-specific message prefix.

Proposal: export one `request`-only port from `stores/connection.ts` (e.g. `connectionClientPort`) that resolves the current client and throws one message; every phase-D web adapter hands that object to its package store, and the existing stores migrate as their D rows land. The pinned test regex `/no client connected/` keeps matching.

Out of D2's scope because it edits `connection.ts` and nine stores' error prefixes, some of which their tests pin.

obra:
Fixes #1464

## Root cause

Every web store that reads the current client from `stores/connection.ts` carried its own copy of the same body — read `connectionStore.getState().client`, throw a store-specific `"<name> store: no client connected; call useConnectionStore.getState().connect(client) first"` when null. The phase-D adapters each hand a package store a client port object, so the port body (and its error) was duplicated across them.

## Change

`stores/connection.ts` now exports the one shared resolution point:

- `requireConnectionClient()` — reads the CURRENT client (never captures one) and throws one shared message when none is wired.
- `connectionClientPort` — the request-only port (`Pick<AppwireClientLike, "request">`) the phase-D adapters hand their package store; built on `requireConnectionClient()`.

The phase-D port adapters now hand the shared object instead of their inline copy:

- `settingsOverview.ts` → `createHubOverviewStore(connectionClientPort)` (request-only).
- `launchConfig.ts`, `keybindings.ts` → shared port for `request` (kept wrapped `async`, so a call before `connect()` still rejects rather than throwing at the call site) and `requireConnectionClient()` for `onNotification`.
- `extensions.ts` → shared port for `request`, its own `onHubConfigNotification` for notifications.

The pinned `/no client connected/` regexes in the store tests keep matching. Stores that still hold a full `AppwireClientLike` (threads, agentsDoc, hubUpdate, credentials, daemonResidents) keep their local copies and migrate to the port as their package rows land, per the issue. `commandCatalog.ts` is untouched: its port deliberately says `"Not connected to the hub"`, pinned by its own test.

## Test evidence

Test-first: added three cases to `stores/connection.test.ts` (current-client resolution, the shared throw, `requireConnectionClient`). They failed before the implementation (`connectionClientPort is not a function`) and pass after.

- `npx vitest run` over `stores/{connection,settingsOverview,launchConfig,extensions,keybindings,keybindings.initialWiring,commandCatalog,credentials}.test.ts` → 8 files, 280 tests passed.
- `npx tsc --noEmit --incremental false` (frontend) → exit 0.
- `npx biome check` on the six changed files → exit 0.


obra:
Addressed the roborev finding in eeb022d2e: the shared `requireConnectionClient()` message now reads `"no client connected; call connectionStore.getState().connect(client) first"` — `connectionStore` is the exported store handle (has `getState`), while `useConnectionStore` is a React hook. `stores/connection.test.ts` pins the corrected string so the regression cannot return. Focused suites: 280 passed; `tsc --noEmit` exit 0; `biome check` exit 0.

obra:
Addressed all three roborev findings in 95cf7ae73:

1. **extensions.ts** — `request` now delegates at call time (`(method, params, opts) => connectionClientPort.request(...)`) instead of copying the method by value, preserving shared-port indirection for spies/mocks.
2. **extensions.ts** — header comment now says `connectionStore.getState().connect(client)` (the store handle), not `useConnectionStore` (the hook).
3. **connection.ts** — rather than overstate, migrated the remaining throwing copies: `threads.ts`, `agentsDoc.ts`, `hubUpdate.ts`, `daemonResidents.ts` now delegate their local `requireClient` to `requireConnectionClient()` (threads keeps its own `rewireClient` call). The comment now accurately names what shares the body and the two deliberate exceptions (commandCatalog's pinned message; credentials/transcriptDisplay early-return).

Evidence: focused store suites (connection, settingsOverview, launchConfig, extensions, keybindings, commandCatalog, credentials, agentsDoc, hubUpdate, daemonResidents) 336 passed; threads + threads.retirement 316 passed; `npx tsc --noEmit --incremental false` exit 0; `biome check` on all changed files exit 0.

PR #1805, opened 2026-09-18
fix(web): re-anchor transcript to the true bottom after a late webfont swap (#917)

### LINKED ISSUE #917: Transcript mount lands a few pixels short of the bottom after a late webfont swap
## What

Opening a session scrolls the transcript to the end from size estimates and measures once. Nothing re-anchors it afterwards without a scroll event or an item-count change. When a webfont swaps in after that landing, `scrollHeight` grows (measured: 11466 → 11487 at `fonts.ready`) while `scrollTop` stays pinned, so the reader is left a few pixels short of the true bottom and the "at bottom" state is wrong.

Measured on a failing `web-transcriptscrollguard` run while diagnosing #900: a 21 px gap held for all remaining frames at 60 fps with `overflow` true throughout.

## Fix shape

Re-anchor to the bottom once after `document.fonts.ready` resolves (and on a `ResizeObserver` tick of the scroll content) when the view was at bottom before the change. PR #916 fixed the guard so it no longer misreports this as "never settled"; the production mount behaviour itself is unchanged.

Found by the PR #916 implementer.

obra:
Fixes #917

## Root cause

Opening a session scrolls the transcript to the end from the virtualizer's
size estimates and measures once. Nothing re-anchors that landing afterwards
unless a scroll event or an item-count change arrives. When a webfont swaps in
after that landing, `scrollHeight` grows (measured 11466 -> 11487 at
`document.fonts.ready`) while `scrollTop` stays pinned, so the reader is left a
few pixels short of the true bottom. Because `wasAtBottomRef` is still true,
no jump-to-latest pill is offered either - the reader is stranded silently.

The scroll listener's own late-content-growth correction only runs on a scroll
event, and neither a font swap nor a row re-measurement produces one.

## Change

In `useTranscriptScroll`'s mount effect:

- re-anchor once after `document.fonts.ready` resolves, covering the swap
  landing before the virtualizer has re-measured the rows; and
- re-anchor on each `ResizeObserver` tick of the scroll content (the sizer
  inside the scroll port, whose height grows when the rows re-measure),
  covering the measurement itself.

Both call one guard that mirrors the scroll listener's existing "more content
in the same box, offset not moved backwards" premise, so a reader who scrolled
away, a prepend, or a port resize is never yanked. `document.fonts` and
`ResizeObserver` are feature-detected; the observer is disconnected on effect
teardown.

## Test evidence

New block `late content growth with no scroll event` in
`useTranscriptScroll.test.ts` (4 tests): the fonts path, the ResizeObserver
path, the scrolled-away negative, and observer disconnect on teardown. The two
positive tests fail before the fix and pass after.

```
npx vitest run src/panes/session/transcript/flow/useTranscriptScroll.test.ts
  Test Files  1 passed (1)
       Tests  129 passed (129)

npx vitest run TranscriptBody.test.tsx useTranscriptScrollKeys.test.tsx NewContentPill.test.tsx
  Test Files  3 passed (3)
       Tests  61 passed (61)

npx tsc --noEmit         # exit 0
npx biome check <changed files>   # No fixes applied
```

The browser `web-transcriptscrollguard` harness was not run locally (it needs a
headless Chrome); CI covers it.


obra:
Both findings addressed in `d422681ad` and `d6aec0f2a`.

**1. Missing gesture veto / duplicated predicate (`useTranscriptScroll.ts:1486-1492`)** — fixed.

- Extracted the geometry predicate to `contentGrewBelowViewport(previous, current)` in `scrollMetrics.ts` and called it from both `handleScroll` and `reanchorIfContentGrew`, so the two copies can no longer drift.
- The re-anchor paths now apply the same gesture veto: `gesturePendingRef.current || middleButtonHeldRef.current`. It is **read, not consumed**, because a content growth fires no scroll event — the marker must survive for the gesture's own scroll event, which `handleScroll` still consumes.
- Corrected the comment: it now says the two share the same geometry predicate and remedy, rather than claiming the listener's exact test.
- Added coverage for a wheel, a touch, and a middle-button autoscroll: each vetoes both the `document.fonts.ready` and the ResizeObserver re-anchor, so the reader is not yanked. Plus six direct predicate tests in `scrollMetrics.test.ts`.

**2. Unit tests can't verify browser layout assumptions (`useTranscriptScroll.test.ts:1665`)** — fixed.

Extended `transcriptscrollguard` with a post-mount growth phase. It grows the scroll content's *own box* via `min-height` (not a row), so the growth stays out of the virtualizer's follow-on-append **and** its row re-measure machinery: scrollHeight grows, scrollTop stays pinned, no item append, no scroll event — only `useTranscriptScroll`'s own re-anchor can reach the new bottom. The runner then requires the transcript to settle at the TRUE bottom (bottomGap ≤ 1px) with no pill.

Verified it discriminates the fix, not just the property:

```
# fix present
transcriptscrollguard ok: ... post-mount content growth re-anchored to the true bottom   (exit 0)

# re-anchor triggers temporarily disabled
transcriptscrollguard FAIL: transcript never re-anchored to the true bottom after post-mount content growth:
  bottomGap=244 scrollTop=23828 scrollHeight=24739 (8s deadline)                          (exit 1)
```

The first version of the probe (padding every rendered row) passed even with the fix disabled, because that growth re-engages the virtualizer's end-anchor; the sizer-box version is the one that reproduces the #917 geometry.

obra:
Addressed the re-review finding, and rebased onto current `main`.

**Medium — a vetoed re-anchor was permanently missed (`4b7a9a321`)**

`reanchorIfContentGrew` wrote the grown geometry to `lastScrollGeometryRef` before the gesture check, so a veto consumed the growth as the new baseline; a gesture that produces no scroll event (selection drag, stationary middle-button hold) then left the reader stranded with no later trigger to correct it.

Fix: the vetoed branch now returns **without** updating the baseline — `previous` stays the uncorrected pre-growth geometry — and arms a single deferred retry on the next frame (the same frame boundary `markGesture` clears its marker on). The retry re-runs the predicate against the preserved baseline and pins once the marker is clear; it is armed only for a growth the gesture actually vetoed, so it is not a poll, and it is cancelled on unmount. The retry is re-armed only while a veto is still standing (the one unbounded case, a held middle button, self-limits to the hold).

Test (in `late content growth with no scroll event`): stage a wheel gesture, grow the content, fire a ResizeObserver tick (vetoed), run the frame queue so the marker clears with no further resize/font event, and assert recovery to the true bottom. Negative control: with the pre-fix ordering (baseline written before the veto) the test fails (`scrollTop` stays `16519`); with the fix it reaches `17366 - 702 = 16664`.

Also rebased onto `origin/main` (`e692ca1e7`) to pick up the `fakeClient` `gateSettlements` change that was breaking `web` / `static-build` on the stale base. No conflicts.

```
npx vitest run useTranscriptScroll.test.ts scrollMetrics.test.ts
  Test Files  1 passed (1) / Tests 153 passed

npx tsc --noEmit                    # exit 0
npm run transcriptscrollguard       # ok: ... post-mount content growth re-anchored to the true bottom (exit 0)
```

PR #1807, opened 2026-09-18
fix(plugins): name the clone a marketplace swap strands under .old (#1033)

### LINKED ISSUE #1033: Marketplace re-source: a double filesystem failure during the clone swap strands the old clone under .old
**Where:** `internal/plugins/marketplaces.go` `swapInClone` (~686-695) and its caller in `EditMarketplace` (~431-435).

**What:** on a re-source, `swapInClone` moves the existing clone aside to `<marketplaces>/.old`, renames the staged clone into place, and hands the aside path back so `EditMarketplace` can restore it if a later save fails or delete it once both files are committed. If the staging-to-destination rename fails AND the immediate attempt to move `.old` back also fails, `swapInClone` returns an empty aside path with its error, so `EditMarketplace` cannot register the aside clone for rollback: the marketplaces file still records the old source (nothing was saved) but its only clone sits stranded under `.old`, and the error does not name it.

**Why it is parked:** it needs two consecutive filesystem failures inside one lock-held edit; the store stays recoverable (the next Refresh re-clones the recorded source, and `.old` is the fixed scratch name the next swap reclaims). The fix is small: return the aside path alongside the error so the caller's undo list can retry the restore and name the path if it fails again, the way the other rollback arms already do.

**Source:** roborev Medium on PR #1021 at head 570dfa0 (the PR's clone-swap rollback was reworked in that PR; this is the residual double-fault edge).

obra:
Fixes #1033

## Root cause

`Manager.swapInClone` renames any existing clone at `dest` aside to `.old`, then renames the staged clone into `dest`. When the install rename fails it immediately tries to move `.old` back. If *that* restore also fails, the old clone — the only local copy — is left under `.old`, and `swapInClone` returned an empty aside path with its error. `EditMarketplace` therefore had nothing to register for rollback: its `undoSwap` arm short-circuits on an empty `swappedIn`, so it never retried the restore, and the returned error said "restoring old clone" without naming the path. The store stayed recoverable (the next refresh re-clones the recorded source), but the stranded clone was undiscoverable from the error.

## Change

- `internal/plugins/marketplaces.go` — `swapInClone` now returns the aside path *alongside* the error when it moved the old clone aside and could not put it back (it still returns `""` when it restored successfully, or when `dest` held nothing).
- `EditMarketplace` registers that aside path in its existing `swappedIn`/`asideClone` undo pair before running `fail`, so the undo retries the restore through `restoreRename` ("old clone <path> to <path>") and, if the retry fails, the path is named in the error. If the retry succeeds, the clone is back under the install location the surviving store file records. Callers that ignore the path on error (`AddMarketplace`, `recloneMarketplace`) are unaffected — they already return before touching it.

No behavior change on any success path or on a swap whose restore succeeds.

## Test evidence

New tests in `internal/plugins/marketplaces_test.go`, both confirmed failing before the fix and passing after:

- `TestEditMarketplace_DoubleSwapFailureNamesTheStrandedClone` — every rename from the install on refuses; asserts the returned error names `marketplaces/.old` and the install location, that `.old` holds the stranded clone, and that the install location is gone. Pre-fix: `error = installing fresh clone failed (rename 2 refused); restoring old clone: rename 3 refused` (no path).
- `TestEditMarketplace_DoubleSwapFailureRetryRestoresTheClone` — the swap's install and its immediate restore refuse, the undo's retry succeeds; asserts the old clone is restored under the install location and `.old` is gone.

```
$ go test ./internal/plugins/... -count=1
ok  	primeradiant.com/evener/internal/plugins	27.027s

$ go vet ./internal/plugins/...
(no output)

$ gofmt -l internal/plugins/marketplaces.go internal/plugins/marketplaces_test.go
(no output)
```


obra:
Withdrawn: issue #1033 carries the `parked` label ("Deliberately parked by ruling; do not pick up"), so this work must not proceed. Sorry for the noise — the fix and tests here are preserved in the issue's history if it is ever unparked.

PR #1808, opened 2026-09-18
D6 pieces 3-5, folded: the draft-checkpoint port with settle and replaced-checkpoint adoption

### LINKED ISSUE #1884: draftCheckpointPort's save() has no atomic insert-if-absent for a genuinely new draft
Follow-up to the D6 keybindings-store chain (#1808/#1791/#1792/#1841/#1844/#1845), round after #1844's panel review.

createDraftRepository's `save()` (appwire-client/typescript/draftCheckpointPort.ts) now compare-and-swaps against the repository's own classified record when one exists, closing the silent-overwrite race for edits to an EXISTING draft (fixed on #1808). When classification is `"absent"` (nothing was stored the last time this repository read or wrote), `save()` still writes unconditionally, because `DraftPort` has no atomic "insert only if still absent" primitive - `replaceIf`'s `expected` argument names an existing record's identity, and every implementation (memoryDraftStorage, the native Storage-backed backend) refuses whenever nothing is stored, so it cannot be reused for this.

Concretely: two store instances sharing the same underlying port (two browser tabs on the same hub's localStorage key, or two KeybindingsStore instances briefly alive during a reconnect) that each classify the store as absent and then both compose their FIRST-EVER draft in the same tick can still silently overwrite one another - the second `save()` wins unconditionally.

This is narrower than the bug #1808 fixed (which covered the far more common "editing an existing draft that someone else also touched" case) and requires either:
- a real create-only primitive on `DraftPort` (e.g. `insertIfAbsent`), implemented per backend, or
- accepting the narrow race as a known limitation and documenting it on `DraftPort`/`createDraftRepository`.

Not fixed in the D6 round because of the added port-surface-area cost of the first option; flagging as a residual per Jesse's policy of filing every review-noticed gap as an issue rather than leaving it only in a commit message.

obra:
## D6 pieces 3-5, folded: the draft-checkpoint port, settle hardening, and replaced-checkpoint adoption

Folds #1736 (piece 3), #1739 (piece 4), #1741 (piece 5) into one PR, per the coordinator's call: three successive review rounds kept re-raising the same cross-piece findings because no single reviewer could see the mechanism whole across five stacked PRs. This branch is cut from p5's head (its tree already contains p3+p4+p5 in full) and fixes every open finding once. #1736, #1739, #1741 are closed with a pointer here.

### Carried from prior rounds (already fixed in this tree, not new work)
- p3 r3 (`1736-r2.md`): `load()`/`hasLastClassified` on empty storage, `save()` not updating `rawFrom`, `lastRemoveIf()` mistyped.
- p4 r2 (`1739-r2.md`): `patchOverrides` gate omitting `saving`/`writeUncertain`, post-rename settle not validating its carried payload, `settleWrite`'s stale JSDoc.
- p5 r2 (`1741-r2.md`): `settledWrite`'s `replaceClassified` catch not setting `storageUnavailable`, `discardStoredKeybindingDraft` dropping its boolean, comments citing review rounds.

### New fixes (this commit, `9ba1343bc`)

**`removeIf`/`replaceIf` identity typed `Checkpoint` but a raw `unknown` is deliberately passed** (1791-r1 Medium, 1792-r1 Medium) — `draftCheckpointPort.ts`'s `DraftPort`, `keybindingsStore.ts`'s `KeybindingDraftStorage`, and `mobile-native/src/nativePreferenceDrafts.ts`'s `NativePreferenceDraftBackend.deleteIf`/`NativeKeybindingDraftBackend.replaceIf` all widened their identity parameter to `unknown`. This removes every `as Checkpoint`/`as never` runtime cast the previous (dishonest) typing required at the call sites — `discardStoredDraft` no longer casts at all.

**`discardClassified`/`replaceClassified` on a never-classified record can destroy or overwrite one it never showed a user** (1792-r1 Medium) — when `!hasClassified` (construction-time defensive case only: `restoreDraft` always classifies via `load()` first in production), both methods now classify via a fresh `load()` before acting: an **unreadable** record (nothing any build could have shown) is still removed/overwritten unconditionally, matching `discardStoredDraft`'s own contract; a **readable** one (some other writer's record this repository never showed anyone) refuses instead, so the caller re-reads/surfaces it — the same posture a refused `removeIf`/`replaceIf` already gets. Added `discardClassified refuses a readable record nothing has classified yet, rather than removing it blind` and the `replaceClassified` twin, plus corresponding corrupt-record (unconditional-act) tests.

**`saveDraft` races with in-flight `patchOverrides`** (1792-r1 Medium, matches the coordinator's cross-piece list) — `patchOverrides` never had an equivalent of `saving`/`writeUncertain` for `saveDraft` to check, so a checkpointed save starting while a direct write was outstanding claimed a NEWER write token and fenced the direct write's own reply out as superseded (or vice versa - already gated in p4's round). Added `directWriteInFlight`, true for the duration of a `patchOverrides` request; `saveDraft` now refuses while it is true - the mirror of `patchOverrides`' existing gate. Added `a checkpointed write is refused while a direct write owns the payload`.

**`discardStoredDraft` casts arbitrary raw bytes to `Checkpoint`** (1792-r1 Medium) — same fix as the identity-widening above; the cast is gone.

**`KeybindingDraftStorage` contract is a source-breaking, undocumented change** (1791-r1 Medium, 1792-r1 Medium) — refuted. Every implementer in the repo is already updated in this tree:
- `keybindingsStore.ts`'s internal `memoryDraftStorage()` fallback (`removeIf: () => true, replaceIf: () => true`).
- `mobile-native/src/nativePreferenceDrafts.ts`'s `nativeKeybindingDrafts()` (delegates to the backend's `deleteIf`/`replaceIf`, both boolean).
- `mobile-native/src/NativePreferencesProvider.tsx`'s real backend object (SQLite-backed, `deleteIf`/`replaceIf` both boolean).
- `appwire-client/typescript/testing/draftStorage.ts`'s generic `memoryDraftStorage<Checkpoint>()` test double.

No other implementer exists. `@evener/appwire-client` is consumed only in-repo (`cmd/evener-hub/frontend`, `mobile-native`); there is no published/external consumer this could break, and per CLAUDE.md I don't add backward compatibility without Jesse's explicit approval. (`KeybindingsDraft` not exported from the package root is p7's own finding — that type doesn't exist on this branch; tracked in #1792's restack.)

**Native TypeScript typecheck failure from `deleteIf` parameter narrowing** (1791-r1 Medium) — refuted by measurement. `npm run check` in `mobile-native` is clean on this tree (confirmed before AND after this commit's changes). CI's native job was already green at the reviewed head (`4c4a440da`).

**Comments encode review-round history** (1792-r1 Low) — measured: no `RoboRev round`/`round N` reference remains in any file this stack touches (`keybindingsStore.ts`, `keybindingsStore.test.ts`, `draftCheckpointPort.ts`, `draftCheckpointPort.test.ts`, `nativePreferenceDrafts.test.ts`) - fixed across the carried p5 r2 round.

### Not in this PR (belongs elsewhere)
- "Discard unreadable draft is a silent no-op when offline" (1791-r1 Low, 1792-r1 Low) - `KeybindingPreferencesScreen.tsx`/`NativePreferencesProvider.tsx` UI wiring, p6's own code. Goes in #1791's restack.
- p7's High ("restored drafts use in-memory generation, not persisted identity") - p7's own mechanism. Goes in #1792's restack.

## Gates
- `npx vitest run keybindingsStore.test.ts draftCheckpointPort.test.ts` (appwire-client/typescript): 87 passed

## Falsification
Reverted `draftCheckpointPort.ts`, `keybindingsStore.ts`, and `mobile-native/src/nativePreferenceDrafts.ts` to this branch's prior head (p5, `100793d7d`) and re-ran the suites: 3 failures as expected (`discardClassified`/`replaceClassified` return `true` instead of `false` on a never-classified readable record; the new write-race test times out because `saveDraft` claims a write token unchecked). Restored with `git apply`; `git status` clean before committing.

## Non-test line count
`git diff -M --stat origin/main...HEAD` (this branch, i.e. all of p3+p4+p5): ~551 non-test production lines across `draftCheckpointPort.ts`, `keybindingsStore.ts`, `readyGenerationFence.ts`, `index.ts`, `nativePreferenceDrafts.ts`, `NativePreferencesProvider.tsx` (plus ~19 net lines of shared test infrastructure in `testing/draftStorage.ts` / the deleted `testing/keybindingDraftStorage.ts`). Under the ~600 non-test ceiling.

## Round 5 (5c6ba3f)
- CI fix: `keybindingRecovery.test.ts`'s failSave fixture now also fails `replaceIf` (b1a0f17f) - the fold's own CAS change made `saveKeybindings()`'s durable-intent-write-failure test route through `replaceIf`, which the fixture never simulated failing. Root-caused both the `native` and `static` (downstream aggregator) CI failures at b0267b7cf.
- Medium (`draftCheckpointPort.ts:143-146`): added `DraftPort.insertIfAbsent`, the atomic twin of `replaceIf` for a record that doesn't exist yet, and wired `save()`'s absent-classification branch through it (e83d1c13). Closes #1884.
- Medium (`keybindingsStore.ts:1202`): `persistDraft` now distinguishes a CAS refusal from a storage throw - on refusal it adopts what's on disk via `restoreDraft` instead of reporting `storageUnavailable` (5c6ba3f5).

Full disposition: PR comment "Round 5 disposition (5c6ba3f)".

Closes #1884.


obra:
## Simplify + review disposition (a08214b)

10 commits, one per item, on top of 9ba1343bc. `git diff --stat 9ba1343bc..HEAD`: 10 files, +188/-320.

1. **KeybindingDraftStorage is DraftPort<KeybindingDraftCheckpoint>** (490e076) - the interface was field-for-field the shared port, copied by hand. Now a type alias.
2. **Drop discardStoredKeybindingDraft wrapper** (5a7eb5d) - it did nothing but call discardStoredDraft with a narrower type, and keybindingsStore.ts re-exported discardStoredDraft under its own name for no importer. index.ts now aliases `discardStoredDraft as discardStoredKeybindingDraft` directly from draftCheckpointPort.ts. Package qualifier falsification not applicable here: draftCheckpointPort.ts was already reachable from index.ts before this change (via keybindingsStore.ts's own imports of createDraftRepository/DraftPort/UnreadableDraftError), so no module became newly reachable-only through this export - the qualifier's "shipped module unreachable" check has nothing to falsify against a same-module export rename.
3. **One tri-state field for classification** (0457bb1) - lastClassified/hasClassified/classifiedAbsent were three fields for one tri-state, reassigned as a 3-line block at 5 sites. Now `classification: null | "absent" | { raw: unknown }`. Pure behavior-preserving refactor (84/84 tests unchanged before and after).
4. **discardClassified/replaceClassified refuse rather than classify blind** (61a6526) - the "nothing classified yet -> fresh load(), refuse if readable, act if unreadable" fallback was unreachable from the only consumer (createKeybindingsStore classifies eagerly: init -> restoreDraft -> load() before decode can throw). Deleted the fallback and the 4 tests that only exercised it (draftCheckpointPort.test.ts:54-92 in the pre-fold file); added one test. Falsified: the new test ("...even over a corrupt record") fails against the pre-fix code with `expected true to be false` at `discardClassified()` - the old fallback would decode-fail and remove the corrupt record; the fix refuses without touching storage. My first draft of this test used a readable record, which the old fallback also refused (for a different reason - a wasted fresh load followed by a decode-success refusal), so it passed under both old and new code; I caught this while falsifying and rewrote it against a corrupt record, which is where old and new code actually diverge.
5. **Route the lost-revision-race settle through settleWrite** (d13a216) - post-rename, lost-revision-race and confirmed-reply each settle a checkpoint the same way, but the lost-revision-race branch (saveDraft's `conflictState` handling) wrote its own ~15-line copy instead of calling settleWrite, because the old `newerExternal: boolean` parameter conflated "apply the payload" with "keep the draft for review" - the race needs both "apply the server's current state" AND "keep the draft", a combination the boolean couldn't express. settleWrite now takes a `payload: KeybindingsOverrides | null` (null = a newer external write already superseded this one's own confirmation, so only `settled` publishes) and a `cleanup: "remove" | "remark"`, independently. No new test added (pure refactor); verified by the existing suite (84/84 unchanged), specifically the three tests exercising post-rename, lost-revision-race and confirmed-reply settle paths.
6. **settleWrite's remark reuses the checkpoint already in hand** (162c495) - the settled checkpoint was re-spelled field by field with a freshly-minted id at what item 5 left as the single "remark" call site (previously two: settleWrite's own remark branch, and the conflictState duplicate merged into it by item 5). Now `{ ...checkpoint, writeUncertain: false }`. Stored id stops rotating on a remark; nothing reads it (confirmed by grep and by the unchanged test suite).
7. **SKIPPED** - directWriteInFlight as a second spelling of `saving`. Per the review file: the general form changes UI surface (assertEditable refuses edits during a direct write). Kept as is.
8. **applyHubOverrides takes a plain extra plus a separate settle** (234513c) - `extra: Partial | (() => Partial)` carried two channels; refreshFor's `hubLoading: false` was spelled twice (once inside the thunk for the applied case, once for the stale-guard `if (!applied) setState(...)` case) to cover both branches. `extra` now always applies (both branches); `settle` is the thunk, resolved only after a successful reconcile. Verified by the existing stale-payload test (keybindingsStore.test.ts:~366, "the stale guard ignores the payload entirely") and the follow-up-fetch tests, unchanged.
9. **keybindingRules always throws plain Error; draftCheckpoint wraps the decode** (cc66b26) - the invalidDraft/invalidUserRules split threaded an `onInvalid` factory through keybindingRules for a distinction only one caller (keybindingsStore.test.ts:874, asserting `not.toBeInstanceOf(UnreadableDraftError)` for malformed user input) needs. keybindingRules now always throws the plain validation error; draftCheckpoint wraps its own decode in try/catch and converts any throw to UnreadableDraftError. That test, and the full suite, pass unchanged.
10. **lostHub is a ReadyGenerationFence method** (de82345) - every other fence predicate is a method; lostHub took the fence as its first argument for its one caller (settleLostHubWrite) instead. Moved onto the interface as `fence.lostHub(generation, stillClaimed)`. ReadyGenerationFence is exported from index.ts; checked mobile-native and cmd/evener-hub/frontend for other constructors of the interface - none found, so no other implementer needed the new method added.
11. **One shared draft-backend fake in mobile-native** (a08214b) - nativePreferenceDrafts.test.ts, preferenceDraftRepository.test.ts and nativePreferences.test.ts each hand-rolled the same Map + JSON.stringify NativeKeybindingDraftBackend fake (three near-identical `backend()`/`draftStorage()` functions). New `mobile-native/src/draftBackend.testkit.ts` exports `fakeDraftBackend()` with `replaceIf`, matching NativePreferencesProvider.tsx's real Storage-backed shape; all three test files now use it. Net -19 lines across the three test files plus the +39-line shared file.
12. **SKIPPED** - NativeKeybindingDraftBackend's `replaceIf` extension stays on the narrower interface (not widened to NativePreferenceDraftBackend) per the review file's reasoning: keeps the narrow contract for the later transcript row.

### Gates (all green)
- `npx vitest run appwire-client/typescript` (full suite, from cmd/evener-hub/frontend): 94 files / 1931 tests passed
- `npx vitest run` in mobile-native (full suite): 83 files / 796 tests passed
- `npm run typecheck` in cmd/evener-hub/frontend: clean
- `npm run check` (tsc) in mobile-native: clean (run for items 1, 2, 9, 10 - port/exported-type changes)
- `npx biome ci ../../../appwire-client/typescript` and `npm run lint` from cmd/evener-hub/frontend: clean

No RoboRev review was posted on this PR at the head this work started from (9ba1343bc) - checked via `gh api repos/prime-radiant-inc/evener/issues/1808/comments` both before starting and immediately before this push (0 comments both times), so there was nothing to fold in.

obra:
## Round 2 fixes (1b3b22dd8)

Both Mediums from the combined review are the same class: `directWriteInFlight` was a bare boolean with a request's lifetime, so a fenced request that never reached `finally` left it stuck, and a stale request's `finally` could clear a newer write's flag.

**Fix:** replaced the boolean with `directWrite: { generation: number; token: number } | null`, capturing the same `(generation, token)` a direct write already claims via `fence.claimWrite()` (matching how the store fences its own PATCH replies):
- The claim, set at `patchOverrides:1117` (`directWrite = { generation, token }`).
- `finally` (`patchOverrides:1153`) clears it only when `directWrite?.token === token` — a stale request's cleanup can no longer clear a newer write's claim.
- The gate in `saveDraft` (`:1258`) now reads `directWrite !== null && fence.writeStillMine(directWrite.generation, directWrite.token)` instead of the raw boolean — the claim stops blocking `saveDraft` the moment the fence would no longer count it as current (generation end, support flap, a later write's supersede), even if the original request's own await never settles.
- `reset()` still explicitly nulls it for symmetry with the immediate-release case.

**Tests** (`keybindingsStore.test.ts`):
- "a direct write retired by a generation reset no longer blocks saveDraft" — a direct write's PATCH is left permanently pending (never resolved); `endReadyGeneration`/`beginReadyGeneration` retires the generation; `saveDraft` now resolves without waiting on the abandoned reply.
- "a stale direct write's late reply cannot clear a newer direct write's flag" — `reset()` wipes the write queue while a first PATCH is still out, a second direct write starts and is still in flight, then the first (stale) reply lands. Asserts `saveDraft` still refuses (bounded race against a timeout, since the pre-fix bug lets `saveDraft` slip through and issue a third PATCH that nothing in the test ever answers — confirmed: pre-fix this assertion reports `"still pending"` instead of a clean rejection).

**Falsification:** reverted `keybindingsStore.ts` to base (`git diff | git apply -R`), ran both new tests — both failed (first: `saveDraft` rejected "Hub keybindings settings are unavailable" instead of resolving; second: outcome was `"still pending"` instead of the expected rejection, and an unhandled rejection from the abandoned third PATCH's `directWriteInFlight` throw was logged) — restored the fix, `git status` clean.

**Gates:**
- vitest `keybindingsStore.test.ts`: 77/77 passed
- `npx biome ci ../../../appwire-client/typescript` and `npm run lint`: clean

Non-test diff: 26 insertions / 11 deletions in `keybindingsStore.ts`.

obra:
## Round 3 disposition (0423ddb) — answering the raw panel reviews

The RoboRev combined-synthesis comment on this PR is broken, so this answers the raw panel members' findings directly (evener/.superpowers/sdd/2026-09-12-mobile-landing-queue-cont/reviews/raw/1808-dfab31333.md).

- **Medium** (member 0, codex): `patchOverrides()` queues via `writeQueue.then(...)` but only marked the `directWrite` claim once the queued callback actually ran, so a same-turn `saveDraft()` call could steal the claim before `patchOverrides()`'s own turn started, fencing its own reply out and stranding the write. **Fixed**: `patchOverrides()` now reserves `directWrite` synchronously at call time — the token stays `null` (pending) until its queued turn actually claims the real fence write token, and `saveDraft()`'s gate treats a pending reservation as blocking as long as its generation is still live (falling back to the precise fence check once a real token is claimed). Compared by object identity in `run`'s `finally`, so a stale call can never clear a newer call's still-pending reservation. Commit 03d582c8a.
  - Test: `keybindingsStore.test.ts` — "a same-turn saveDraft is refused behind a just-issued direct write" (patchOverrides and saveDraft called in the same tick, before either's async work runs).
  - Falsified: reverted the production diff, kept the test — it hung/timed out (`saveDraft` never rejected, so `await expect(save).rejects.toThrow(...)` never resolved). Restored the fix, test passes.

- **Low** (member 1, codex): `testing/draftStorage.ts`'s in-memory `DraftPort` fake narrowed `removeIf`/`replaceIf`'s identity parameter to `Checkpoint`, while the interface declares `unknown` (a conforming port must accept the raw, undecoded value `discardStoredDraft()` hands it for an unreadable record). **Fixed**: widened both parameters to `unknown`, matching the interface exactly. Commit 2550a65f6.

- **Low** (member 3, codex): A save fenced by support loss (`settleLostHubWrite`, the lost-hub path) published `writeUncertain: true` but left `draftConflict` false, unlike every other unknown-outcome settle path (no reply at all, a malformed reply). **Fixed**: `settleLostHubWrite` now also sets `draftConflict: true` in the same publish. Commit 0423ddb52.
  - Test: "a save lost to a support flap settles draftConflict the same as the other unknown-outcome paths" (support drops to `"unknown"` mid-save, then the request rejects).
  - Falsified: reverted the production `setState` change — the assertion failed (`draftConflict: false`, expected `true`). Restored, test passes.

- Member 2 (pi): no issues found.

### Gates
- vitest `keybindingsStore.test.ts`: 82/82 · `draftCheckpointPort.test.ts`: 13/13

Head: `0423ddb5289e036222ecbd6f4cf6be13d538ae15`

obra:
## Round 4 disposition (b0267b7c) — answering the raw panel reviews

Fixed here (own):

- **persistDraft's local save failure left draftError null** (member 3, 1808-0423ddb52.md; also flagged independently on later heads). `persistDraft`'s catch only set `storageUnavailable`; native renders `draftError ?? hubError`, so a local disk failure disabled the editor with no message. Fixed: the catch now also sets `draftError: DRAFT_SAVE_FAILED_MESSAGE`. Falsified: reverted the one-line fix, `editDraft's local save failure marks draftError alongside storageUnavailable` failed with `draftError: null`; restored, green.

- **draftCheckpointPort's save() unconditionally overwrote the classified record** (High, member 3 on #1844's panel — routed here since the mechanism is `draftCheckpointPort.ts`/keybindingsStore's `persistDraft`, not #1844's own diff). `editDraft`, `saveDraft`'s pre-request persist, and `rebaseDraft` all reach `createDraftRepository.save()`, which wrote to the port unconditionally even when a concurrent writer (another store sharing the same port) had replaced the classified record. `save()` now returns `boolean` and CAS-replaces via `replaceIf` against the current classification when one exists (unconditional only when nothing is classified yet — see residual below); `persistDraft` treats a refusal the same as a storage exception. Discovered in the same pass: `removeIf`/`discardClassified` never marked the store `"absent"` on a successful removal, so a later `save()` would wrongly CAS against the just-removed identity — fixed the same commit. Falsified with 5 repository-level tests (removeIf/discardClassified → later save; refuse-on-replaced-record) plus a store-level test (`editDraft refuses to overwrite a checkpoint replaced by another store sharing the same port`); all fail without the fix, pass with it.
  - Residual (not fixed, filed): a genuinely NEW draft (nothing classified yet on either side) has no atomic insert-if-absent primitive to race against — narrower window, tracked as https://github.com/prime-radiant-inc/evener/issues/1884.

Pointers (fixed downstream, not here):

- **Unreadable persisted draft blocks discardDraft via assertEditable; discardStoredDraft blindly removes; native backend compares raw JSON bytes** (member 0) — all three are #1791's own subject: the store-free discard path, the removed/absent/refused outcome split, and matchesStoredBytes' canonicalization. See #1791's round 6 disposition.
- **discardStoredKeybindingDraft has no production call site** (member 2) — wired into `NativePreferencesProvider`'s offline recovery action in #1791.

Gates: vitest (appwire-client/typescript) 1987/1987 (post-#1791 stack; own-diff files pass in isolation too); typecheck (frontend) clean; `npx biome ci` clean; `make lint-package-imports` PASS.
Own diff (vs the base main this branched from): 17 files, +1630/-269 (includes the extraction series #1808 already carried plus these two fixes).

obra:
## Round 5 disposition (5c6ba3f)

**CI root cause (b0267b7cf, run 35379779207, native + static jobs red):** `keybindingRecovery.test.ts`'s "does not dispatch if writing the durable intent fails" failed with `promise resolved ... instead of rejecting`. Measured: `createDraftRepository.save()` now CASes via `storage.replaceIf()` once a draft is classified (this fold's own change), but the fixture's `failSave()`/`allowSave()` only toggled `storage.save()` - so a `saveKeybindings()` after an existing edit no longer observed the simulated disk failure. Oracle check: every other "does not dispatch if writing the durable intent fails/cannot be saved" test in the suite (approvals, hubUpgrade, navigationActions, sessionDeletion) asserts the same "a durable-intent write failure must reject" contract, confirming this is a fixture bug, not a stale assertion. Fixed by making the fixture fail `replaceIf` too (b1a0f17f), the same fix already staged on #1791 as f8e576fdf - cherry-picked here since it belongs with this fold's own CAS change. The "static" job's failure was purely downstream: its "Require every static lane" gate red-lit because `native` was red in the same run; no separate cause.

**Medium (draftCheckpointPort.ts:143-146, #1884):** `save()`'s absent-classification branch wrote via the port's unconditional `save()`, so two writers racing to create the FIRST-EVER draft (both classifying "absent") could let the second one silently win. Added `DraftPort.insertIfAbsent`, the atomic twin of `replaceIf` for a record that doesn't exist yet; `createDraftRepository.save()`'s absent branch now calls it and propagates a refusal like the classified branch already does. Web's ephemeral fallback and the native `Storage`-backed backend implement it the same way their existing `replaceIf`/`deleteIf` do (synchronous compare, no interleaving). Failing-first test in `draftCheckpointPort.test.ts` (watched it fail: `expected true to be false`), plus wiring tests in `nativePreferenceDrafts.test.ts` (falsified by reverting the wiring line, confirmed both fail without it, restored). Closes #1884 (e83d1c13).

**Medium (keybindingsStore.ts:1202):** `persistDraft` treated `drafts.save()`'s CAS refusal the same as a storage throw - set `storageUnavailable`, left the stale classification in place. Now it distinguishes refusal (no throw) from an actual throw: on refusal it adopts what's on disk via `restoreDraft(getState())` (the same posture `discardDraft`/`settleWrite` already take on their own refusals) and throws the existing "Shortcuts changed again. Review the current values." message (extracted to `DRAFT_REVIEW_AGAIN_MESSAGE`, reused by `rebaseDraft`) instead of the storage-failure one. Failing-first test added (`editDraft's persistDraft refusal adopts the replacement instead of reporting storageUnavailable`), watched it fail against the pre-fix code. One pre-existing test ("editDraft refuses to overwrite a checkpoint replaced by another store sharing the same port") asserted the OLD buggy text ("Could not save the shortcut draft locally.") for this exact scenario - corrected its expectations to the new message plus `storageUnavailable: false`/adopted-draft assertions (strictly more coverage, not weakened). (5c6ba3f5)

**Gates:** mobile-native vitest (83 files / 801 tests) green; appwire-client/typescript + web vitest (96 files / 1976 tests) green; `npm run typecheck` (frontend) clean; `npm run check` (native tsc) clean; `npx biome ci ../../../appwire-client/typescript` clean; `npm run lint` (frontend) clean; `make lint-package-imports` PASS. index.ts/tsconfig.build.json untouched, so `make test-api-package` not run.

Non-test diff: draftCheckpointPort.ts +18/-4, keybindingsStore.ts +33/-11, testing/draftStorage.ts +8/-1, NativePreferencesProvider.tsx +5, nativePreferenceDrafts.ts +5 (~54 lines net).

obra:
Follow-up (7dcf01e), addressing p9's round-2 panel finding against this fold's own settle path (keybindingsStore.ts:1050, "restoreDraft(getState()) evaluates conflict against the pre-apply revision"): measured and refuted. `settledWrite` (the only settle callback applyHubOverrides' 3-arg form ever receives) already returns `restoreDraft({ loaded: true, revision: payload.revision })` explicitly - the authoritative refresh's own revision, never `getState()`'s pre-apply one. `settleWrite`'s own `restoreDraft(getState())` calls (saveDraft's confirmed-reply settle) run AFTER `applyHubOverridesSettling` has already published the new revision via its own setState, so `getState()` there is already current too. Added a regression test pinning the correct behavior (falsified: swapping in a `getState()`-based read fails it exactly as the finding describes, confirming the test is load-bearing) rather than a fix, since no bug reproduces.

obra:
Follow-up (0340968): CI's `web` job caught a missed `npx biome ci` run on the round-5 settledWrite regression test (7dcf01e6, a multi-line object literal biome wants collapsed) - fixed and pushed. The `tests` job's `agent` module failure (`TestRetirementSafetyDetachedLifetime`) is unrelated to this PR's diff (TypeScript-only, no Go files touched) and matches the already-filed #1879 (flaky session_retirement_evidence tests on TypeScript-only PRs).

obra:
## Round 6 disposition (d421d548)

Two findings from the fold's own round-5/round-6 review, both in `keybindingsStore.ts:persistDraft` / `draftCheckpointPort.ts`:

**`persistDraft`'s `drafts.createId()` ran outside its own try block** (line ~1208): an id-generation failure (a crypto/random-source failure) escaped `editDraft`/`saveDraft`/`rebaseDraft` uncaught instead of setting `storageUnavailable`+`draftError` the same way `save()` throwing already does. Moved the checkpoint's construction (the `createId()` call plus the input spread) inside the try. Pinned by a new test overriding only `createId` to throw; falsified by moving it back outside the try (the test then throws the raw `"crypto unavailable"` message instead of the local-write message).

**`DraftRepository.save`'s own doc comment** said an absent/unclassified save always succeeds - stale since `insertIfAbsent` gained its atomic CAS (#1884): the absent/unclassified path now goes through `insertIfAbsent` and can refuse exactly like `replaceIf` does. Doc-only fix, no behavior change.

Head: `d421d5489d05abe483380677c6410fee1bc5617b`

Gates (targeted, all green): vitest on `keybindingsStore.test.ts` + `draftCheckpointPort.test.ts` (98 tests); `npm run typecheck`; `npx biome ci ../../../appwire-client/typescript`; `npm run lint`.

Not pushed until now per your hold; pushed together with the p6a-p10 restack in this pass (`--force-with-lease=claude/sdk-d6-p3to5-draft-port:034096809`).

obra:
## Round 6 disposition (d421d54)

- Medium (`draftCheckpointPort.ts:51-54` / `NativePreferencesProvider.tsx:37-40`: the native backend still parses with `JSON.parse` and cannot remove a malformed record or distinguish a stored JSON `null`): this is the content of the next two PRs in the chain. #1900 (piece 6a) adds the tagged read outcome and raw-identity removal, and #1902 (piece 6b) replaces the native backend with the shared `rawStringDraftBackend` so production and the test fake run the same parser. They merge directly behind this fold. No change here.
- Low (`keybindingsStore.test.ts:1046-1051` real 200 ms timeout): filed as #1918 with two sibling Lows; follow-up PR after the chain.

PR #1809, opened 2026-09-18
fix(repair): name the oneOf over-match instead of satisfied branch requirements (#623)

### LINKED ISSUE #623: repair: multiple-match oneOf failures render satisfied branch requirements with no recovery hint
## Summary

When a `oneOf` fails because the arguments **match more than one branch** (oneOf means exactly-one), `oneOfConstraintMessage` in `agent/internal/tool/repair/explain.go` renders "Branch N requires: …" lines describing requirements the arguments **already satisfy**. A model reading the message cannot infer what to change: the problem is over-matching, not under-matching.

Found by adversarial review of the #618 fix; probe-verified with santhosh-tekuri/jsonschema v5.3.1.

## Reproduction

Schema:

```json
{
  "type": "object",
  "properties": { "a": {"type": "string"} },
  "oneOf": [
    { "required": ["a"] },
    { "required": ["a"] }
  ]
}
```

Args `{"a":"x"}` match both arms. The error cause is bare `/oneOf` (no arm index, no children), extracted keyword `oneOf`, and the current message renders:

```
tool: arguments violate a conditional rule on the combination of arguments (the schema's oneOf constraint), not any single argument's type or value.
Branch 0 requires: send all of "a".
Branch 1 requires: send all of "a".
```

Both branch lines are satisfied by the args. The header is honest; the body is useless for recovery.

## Suggested fix direction

When the failing cause is a bare combinator with no child causes (the multiple-match shape), render "matched more than one branch; make the arguments satisfy exactly one" instead of enumerating branch requirements. The distinguishing signal is available on the cause tree: a no-match failure carries per-arm child causes (e.g. `/oneOf/0/not`, `/oneOf/1/properties/.../enum`), while a multiple-match failure carries none.

Latent against today's built-in schemas (the delegate oneOf arms are mutually exclusive by construction), reachable via plugin/MCP schemas.

## Environment

obra:
Fixes #623

## Root cause

`oneOfConstraintMessage` (`agent/internal/tool/repair/explain.go`) rendered a
"Branch N requires: …" line for every arm of a failing `oneOf` — even when the
failure was the *multiple-match* shape (oneOf means exactly-one), where every
arm's requirements are already satisfied. A model reading those lines is told to
send exactly what it already sent, so it cannot infer the real problem:
over-matching, not under-matching.

The two failure shapes are distinguishable on the validator's cause tree. Probing
santhosh-tekuri/jsonschema v5.3.1 (Draft 2020-12):

- multiple-match → deepest cause is the bare `/oneOf` node with **no** children
  (`Message="valid against schemas at indexes 0 and 1"`); `offendingKeyword`
  therefore returns `"oneOf"`.
- no-match (delegate shape) → `/oneOf` carries one child per failing arm
  (`/oneOf/0/not`, `/oneOf/1/properties/sandbox/enum`); the deepest-first walk
  surfaces the arm's inner keyword (`"not"`).

`offendingKeyword` descends while `Causes` is non-empty, so a bare combinator
node is exactly the case where the combinator's own keyword is the deepest one.

## Change

When the deepest keyword is `oneOf` (bare combinator, no child causes),
`oneOfConstraintMessage` renders the over-match and its recovery instead of
enumerating satisfied branch requirements:

```
probe_tool: arguments violate a conditional rule on the combination of arguments (the schema's oneOf constraint), not any single argument's type or value.
The arguments matched more than one branch; make them satisfy exactly one.
Example: {}
```

The no-match shape (deepest keyword `not`, delegate pairing rule) is unchanged —
it still enumerates each branch's requirements.

## Tests

- `agent/internal/tool/repair/explain_oneof_test.go`:
  `TestExplainSchemaError_MultipleMatchOneOfNamesOverMatch` — fails before the
  fix with exactly the issue's reported output, passes after; asserts no
  "Branch N requires" lines and that the over-match/recovery prose is present.
- `agent/session_tool_repair_oneof_multimatch_test.go`:
  `TestOffendingKeyword_MultipleMatchOneOfRendersOverMatch` — end-to-end through
  the real validator: compiles the issue's schema, validates matching args, and
  pins that `offendingKeyword` returns `"oneOf"` and the message names the
  over-match.

Evidence:

```
$ go test ./agent/internal/tool/repair/...
ok  	primeradiant.com/evener/agent/internal/tool/repair

$ go test ./agent/ -run TestOffendingKeyword_MultipleMatchOneOfRendersOverMatch
ok  	primeradiant.com/evener/agent

$ go vet ./agent/...
(clean)
$ gofmt -l <changed files>
(clean)
```

No behavior change for the no-match delegate shape; the existing
`TestExplainSchemaError_OneOfConstraintDoesNotReportMissingArgument` and the
enum-rendering tests still pass.


obra:
Adopted the branch head and addressed the combined-review findings (head now 9917e0a84, rebased onto origin/main).

**Lint (`agent/session_tool_repair_oneof_multimatch_test.go`)**
`errors.As(err, &ve)` → `errors.AsType[*jsonschema.ValidationError](err)`. Confirmed with `make lint-golangci` → `PASS lint-golangci (232s)`, 0 issues.

**Finding 1 — nested `oneOf` misclassified as over-match: confirmed by reproduction.**
Schema `{"type":"object","oneOf":[{"oneOf":[{"required":["a"]},{"required":["a"]}]},{"required":["b"]}]}` validated against `{"a":"x"}` — the validator's deepest cause `KeywordLocation` is `/oneOf/0/oneOf` (the inner over-match), while the outer `oneOf` matched zero branches. Because the old caller passed only the last keyword segment (`oneOf`), `oneOfConstraintMessage` rendered:
`The arguments matched more than one branch; make them satisfy exactly one.`
which falsely claims the outer branches over-matched. Fix: preserve the full location (`offendingKeyword` → `offendingKeywordLocation`) and render the over-match only when the location is a bare top-level `/oneOf` (`isBareOneOfLocation`). The nested case now keeps the outer branch enumeration:
`... \nBranch 1 requires: send all of "b".`

**Finding 2 — over-match `Example:` contradicted the coaching: fixed.**
The over-match shape no longer appends `minimalExample(params)`: it renders only top-level required fields and ignores the `oneOf` arms, so it printed `Example: {}` — an object matching zero branches. No generic example is emitted for the over-match shape.

**Regression coverage:** repair-level `TestExplainSchemaError_MultipleMatchOneOfNamesOverMatch` / `_NestedOneOfNoMatchDoesNotClaimOverMatch` / `_MultipleMatchOneOfAcceptsBareKeyword`, plus end-to-end validator cause-tree tests `TestOffendingKeywordLocation_MultipleMatchOneOfRendersOverMatch` / `_NestedOneOfNoMatchNotOverMatch`.

Rebased onto `origin/main` with no conflicts (picks up #1826's `gateSettlements` fix for the web/static checks).

obra:
roborev follow-up (head now 0857351a4, rebased onto origin/main):

**`agent/internal/tool/repair/explain.go:815` (non-over-match `oneOf` path) — fixed.**
The branch-enumeration path still appended `minimalExample(params)`, which reflects only top-level required fields and ignores the `oneOf` arms. For the nested regression schema (`{"oneOf":[{"oneOf":[{"required":["a"]},{"required":["a"]}]},{"required":["b"]}]}`, args `{"a":"x"}`) it printed `Example: {}` — an object matching none of the branches the message had just enumerated. The example is now omitted for `oneOf` branch explanations; the delegate `not` shape keeps its example, where the top-level required list *is* the selector the described branches need (`{"task": ...}` satisfies branch 0 and not branch 1 — exactly one).

Both nested regression tests now assert no `Example:` line is emitted (repair-level `TestExplainSchemaError_NestedOneOfNoMatchDoesNotClaimOverMatch`, agent-level `TestOffendingKeywordLocation_NestedOneOfNoMatchNotOverMatch`).

Local: `go test ./internal/tool/repair/` and the focused `agent` tests pass; `golangci-lint` on the changed packages reports 0 issues.

PR #1810, opened 2026-09-18
plugins: store write failures never name the store path

### LINKED ISSUE #1700: plugins: audit AddMarketplace's non-rollback save failure for store path leakage
## Where

`internal/plugins/marketplaces.go`, `AddMarketplace`'s save-failure path that does not roll back.

## What

#1602 round 8 (`1744e3be9`) stopped three rollback branches from leaking the store's absolute path into wire errors via the `rollbackFailed` helper. `AddMarketplace`'s non-rollback save-failure branch was not in the finding's scope and has not been audited for the same `*fs.PathError` leak.

## Proposed change

Audit that branch with a test that uses a real `*fs.PathError` (as the round-8 tests do) and route it through `rollbackFailed` or the equivalent path-free wrapping if it leaks.

Estimate: ~20 lines.

Flagged by the #1602 lane as a residual.

obra:
## Summary
- `AddMarketplace`, `ensureFetched`, `RemoveMarketplace`, and `RefreshMarketplace` return raw `atomicWriteFile`/`os.RemoveAll` errors on a metadata-save or rollback failure, and those errors' own text names the plugin store's absolute filesystem path (a `*fs.PathError`'s `Error()`); that raw error reaches the RPC caller as a wire error.
- `AddMarketplace`'s rollback failure (removing the freshly-fetched clone when the save that would have recorded it fails) was silently discarded via `_ = marketplaceRemoveAll(...)`, so the save's raw error was returned either way, whether or not the rollback itself also failed.
- `RemoveMarketplace` removed the clone before saving metadata and swallowed a clone-removal failure into a log line, reporting success even when litter was left behind.
- Three helpers - `saveFailed`, `rollbackFailed`, `storeChangeRollbackFailed` - log the real error server-side (`m.Stderr`) and return a path-free message naming only the marketplace, applied at all four write sites plus `EditMarketplace`'s `fail` closure (only its rollback-also-failed branch; its clean-rollback branch is unchanged, matching a gap this PR does not close - see Residuals).
- `RemoveMarketplace` now saves metadata before removing the clone: a save failure is a plain refusal (marketplace stays registered, clone untouched), and a clone-cleanup failure after a successful save now reaches the caller as an error instead of a silently swallowed warning.

## Residuals (filed separately, not in this PR's scope)
- `EditMarketplace`'s clean-rollback branch (`swapErr == nil && undoErr == nil`) still returns the underlying `err` raw, which can carry the store's absolute path when it originates from `moveMarketplace`'s own internal rollback-failure join or `saveRename`'s clean-restore path. This PR only scrubs the outer rollback-also-failed branch, matching the reference commit's own scope; the deeper leak needs its own audit.
- `ensureFetched` has no rollback of the freshly-cloned directory when its backfill save fails (main's current behavior, unchanged here); the clone is left orphaned on disk. Adding that rollback is a separate feature, not a path-leak fix.

## Test plan
- [x] `marketplaces_path_scrub_test.go`: one test per write site (`ensureFetched` via `Browse`, `AddMarketplace` save failure, `AddMarketplace` rollback-also-failed, `RemoveMarketplace` save failure, `RemoveMarketplace` clone-removal failure, `RefreshMarketplace` save failure, `EditMarketplace` rollback-also-failed), each forcing a real `*fs.PathError` and asserting the wire error carries no path.
- [x] Falsification: reverted `internal/plugins/marketplaces.go` only, reran the new tests - all 7 failed (leaked path or wrong error), confirming they are load-bearing; restored and reran green.
- [x] `coverage_marketplaces_fuzz_test.go`'s `RemoveMarketplace` assertion updated for the save-before-remove reorder (was asserting the old swallowed-success behavior); the same underlying behavior is now asserted by `TestRemoveMarketplaceCloneRemovalFailureNamesNoPath`.
- [x] `$(go env GOROOT)/bin/gofmt -l internal/plugins/` - clean
- [x] `golangci-lint run ./internal/plugins/...` - 0 issues
- [x] `go test -count=1 ./internal/plugins/...` and `-tags evenerfuzz` - both green
- [x] `go build ./...` and `go test -count=1 ./internal/plugins/... ./cmd/evener/... ./cmd/evener-hub/...` - all green

Closes #1700.


obra:
## Round 2 disposition (8044777)

### High: EditMarketplace's successful-rollback save failure still named the store path
Confirmed by measurement, not just reading. Added `TestEditMarketplaceSaveFailureRollbackSucceedsNamesNoPath` (a re-source edit whose `saveMarketplaces` fails with a real `*fs.PathError` naming the store's absolute path, with the swap-in rollback succeeding cleanly). Watched it fail on `40b66faf0` — the wire error was:

```
write /var/.../TestEditMarketplaceSaveFailureRollbackSucceedsNamesNoPath.../003/known_marketplaces.json: permission denied
```

Fix: both of `EditMarketplace`'s save call sites (`saveRename` for a rename, `saveMarketplaces` for a plain re-source) now scrub through `saveFailed` before handing the error to the `fail` closure — the same wrapper `ensureFetched`/`AddMarketplace`/`RemoveMarketplace`/`RefreshMarketplace` already use for this exact failure. `saveFailed`'s own message now names the store file (not a path) so `TestEditMarketplace_MarketplacesSaveFailureRestoresTheStore`'s `strings.Contains(err.Error(), marketplacesFileName)` still holds. The raw error (with its absolute path) still reaches the hub's log via `saveFailed`'s own `Fprintf`.

Deliberately left untouched: the `fail` closure's own success-branch logic. `moveMarketplace`'s and `swapInClone`'s own errors also reach `fail` and can carry absolute paths (`os.Rename`'s `*LinkError`), but `TestEditMarketplace_FailedUndoNamesWhatItCouldNotRestore` explicitly asserts one such path (`m.marketplaceDir("beta")`) survives into the returned error today — generalizing the fix to `fail`'s branch would have broken that oracle test. Filing this as a residual below rather than reworking that test unasked.

Gate: `go test -run TestEditMarketplaceSaveFailureRollbackSucceedsNamesNoPath -v ./internal/plugins/...` — FAIL before, PASS after.

### Low 1: storeChangeRollbackFailed's doc comment named ensureFetched
Fixed — the comment now names only `AddMarketplace` and `EditMarketplace`'s `fail` closure, which are the two callers that actually reach it.

### Low 2: duplicate no-git marketplace fixture
`makeMarketplaceRepoDir` in `marketplaces_path_scrub_test.go` duplicated `makeMarketplaceRepoDirNoGit` in `doctor_test.go:656`. Deleted the duplicate; its one caller (`TestRefreshMarketplaceSaveFailureNamesNoPath`) now uses `makeMarketplaceRepoDirNoGit` directly.

### Gates
- `$(go env GOROOT)/bin/gofmt -l internal/plugins/marketplaces.go internal/plugins/marketplaces_path_scrub_test.go` — clean
- `go vet ./...` — clean
- `go vet -tags evenerfuzz ./...` — clean
- `GOOS=windows go vet -tags evenerfuzz ./...` — clean
- `go test -count=1 ./internal/plugins/... ./cmd/evener-hub/...` — ok
- `go test -count=1 -tags evenerfuzz ./internal/plugins/... ./cmd/evener-hub/...` — ok
- `golangci-lint run ./internal/plugins/...` — 0 issues

### Falsification
Reverted only `internal/plugins/marketplaces.go` to base (`git checkout HEAD -- internal/plugins/marketplaces.go`), ran the new test: FAIL with the raw absolute path in the error (pasted above). Restored via the saved patch; `git status --short` afterward showed only the two intended files.

### Residual (for the coordinator to file)
`moveMarketplace`'s rename errors (`renaming marketplace clone`/`renaming plugin cache`) and `swapInClone`'s rename errors can also carry this machine's absolute plugin-store path when they reach `EditMarketplace`'s `fail` closure on its success branch (`os.Rename`'s `*LinkError` names both paths). Not fixed this round: no review finding covers it, and `TestEditMarketplace_FailedUndoNamesWhatItCouldNotRestore` currently asserts that exact leaked path is preserved (`m.marketplaceDir("beta")`), so fixing it needs that test rewritten too — a separate, deliberate change.

Head: `80447778bc121dcf97e8bb81e6d1c678d7a407f5`

obra:
## Round N disposition (361a33a3) — answering the raw panel reviews

Both fixes landed on top of bbf024044 (the round this panel reviewed).

**Medium — marketplaces.go:506-507/:269 (saveFailed misattributes the failing file, drops errStoreBetweenNames identity)** — fixed in bfd2ec39e.
`saveFailed` now takes the file name its caller was writing (`saveFailed(name, fileName, err)`) instead of hardcoding `marketplacesFileName`. `saveRename` returns which of the two files (`known_marketplaces.json` / `installed_plugins.json`) actually failed alongside its existing error (an additive `failedFile` return — its own raw error text and the three `marketplace_migration.go` callers that use it directly for `errors.Is(errStoreBetweenNames)` and evener-doctor's diagnostic text are unchanged). `saveFailed` re-attaches `errStoreBetweenNames` via `%w` when the raw error wraps it, so `errors.Is` survives the scrub; the sentinel's own text carries no path.
Tests: `TestEditMarketplaceRenameRegistrySaveFailureNamesRegistryFile` (a plain rename's registry-save failure names `installed_plugins.json`, not `known_marketplaces.json`), `TestEditMarketplaceRenameBetweenNamesErrorSurvivesIdentity` (`errors.Is(err, errStoreBetweenNames)` survives when the marketplaces save and the registry restore both fail).

**Medium — marketplaces.go:308 (RemoveMarketplace's clone-cleanup error doesn't say unregistration succeeded)** — fixed in 361a33a30.
Measured `git show origin/main:internal/plugins/marketplaces.go`: pre-PR order was remove-clone-then-save (silently swallowing a clone-removal failure and reporting success — the bug #1700's PR fixed). The reorder to save-then-remove is intentional (58fdcc1b4's commit message: "RemoveMarketplace swallowed its clone-cleanup failure into a log line and reported success"), so I did not revert the order. Instead `rollbackFailed`'s message now states the marketplace is no longer registered, so a caller can tell this error apart from one where the marketplace is still registered.
Test: `TestRemoveMarketplaceCloneRemovalFailureNamesNoPath` now also asserts the message says "no longer registered".

**Low — marketplaces_path_scrub_test.go:222 (no renaming=true path-scrub coverage)** — resolved by the same two tests added for the Medium above (both exercise `EditMarketplace`'s renaming branch via `saveRename`); no separate commit.

### Gates
- gofmt (toolchain): clean
- go vet / -tags evenerfuzz / GOOS=windows -tags evenerfuzz: clean
- golangci-lint run ./internal/plugins/...: 0 issues
- go test -count=1 ./internal/plugins/... (plain and -tags evenerfuzz): ok

### Falsification
- bfd2ec39e: reverted marketplaces.go/marketplace_migration.go/store_changed_test.go, both new tests failed (`... saving known_marketplaces.json failed ...` instead of naming installed_plugins.json; `errors.Is` false), restored cleanly (`git status --short` clean).
- 361a33a30: reverted marketplaces.go, `TestRemoveMarketplaceCloneRemovalFailureNamesNoPath` failed (no "no longer registered" in message), restored cleanly.

### Heads / diff
- fix(plugins): saveFailed names the store file that actually failed: `bfd2ec39ed2c19029452b727446dfd36595f6170`
- fix(plugins): RemoveMarketplace's clone-cleanup error says unregistration succeeded (head): `361a33a30f2516b3d0dc11954afd6909233f9c7f`
- `git diff origin/main...HEAD --stat`: 5 files changed, 513 insertions(+), 23 deletions(-) (internal/plugins/ only)

obra:
## Round 4 disposition (2a27aa42) — answering the raw panel reviews

Round 4 of 5 — closed the class rather than patching each instance.

**Medium A (members 0, 1): migration callers leaked the raw path** — fixed in `9fb24e11f`.
Round 3's `failedFile` design trusted every `saveRename` caller to wrap the returned error through `saveFailed`. `EditMarketplace` did; the three `marketplace_migration.go` callers (an interrupted-rename recovery, a marked merge, `migrateMarketplaceName` — all reachable from `ListMarketplaces`/`List`, which can surface their error over RPC just as directly as `EditMarketplace`) discarded `failedFile` and returned the raw error. Closed the class: `saveRename` now scrubs its own error unconditionally (calls `saveFailed` internally, naming whichever file failed) and returns a single already-scrubbed error — no caller can forget, because there's nothing left for a caller to opt into. `EditMarketplace`'s own wrap is gone (redundant now). The three migration callers only needed their signature reverted (drop the removed second return) — they're now byte-identical to `origin/main` again.
Test: `TestMigrationTriggeredSaveFailureNamesNoPath` — a `ListMarketplaces` call against a legacy-named marketplace whose save fails.
`marketplace_migration_test.go`'s `TestMarketplaceNameMigration_ARecoveryThatRollsBackRemovesTheMarker` asserted the raw injected `"boom"` text reached the caller — exactly the leak this closes — so its assertion now checks the scrubbed message names the failed file instead (that check alone doesn't distinguish scrubbed from raw text, since both name the file; `TestMigrationTriggeredSaveFailureNamesNoPath` is what falsifies).

**Medium B (member 1): `storeChangeRollbackFailed` drops `errStoreBetweenNames` on double failure** — fixed in `2a27aa429`.
`EditMarketplace`'s `fail` closure preserves the sentinel when the outer directory rollback succeeds (returns the scrubbed cause unchanged); when that rollback also fails, it calls `storeChangeRollbackFailed(name, cause, ...)`, which never wrapped `cause` at all. `storeChangeRollbackFailed` now re-attaches `errStoreBetweenNames` via `%w` when `cause` wraps it (the sentinel's own text carries no path, so this doesn't reopen anything).
Test: `TestEditMarketplaceRenameBetweenNamesRollbackFailureSurvivesIdentity` — a rename+resource edit whose marketplaces save, registry restore, AND swap rollback all fail.

**Medium C (member 3): AddMarketplace deletes the aside clone before the save that could still fail** — not fixed in this PR; filed as issue.
Measured via `git show origin/main:internal/plugins/marketplaces.go`: `AddMarketplace`'s resourcing/rollback block (`swapInClone` → immediate `marketplaceRemoveAll(old)` → `saveMarketplaces`, rollback removing only the new clone) is byte-identical to `origin/main` — this PR's diff never touches it. Pre-existing, not introduced here. Reproduced (a re-source of an already-registered name whose save then fails leaves the registered entry's `InstallLocation` pointing at nothing) and filed as https://github.com/prime-radiant-inc/evener/issues/1877 with the repro and suggested fix (retain the aside clone until the save succeeds; restore it on failure instead of just deleting the new one).

**Low (member 3): `marketplaces_path_scrub_test.go:194` ignores `ListMarketplaces`' error** — fixed in `17ea2a4fb`.
Captures and requires a nil error before checking the map.

**Low (member 3): `coverage_marketplaces_fuzz_test.go:246` comments narrate prior behavior** — fixed in `7f6aa6386`.
Also caught the same pattern in a comment I wrote in round 1 (`marketplaces_path_scrub_test.go`, "previously swallowed... this now reaches") while I was in the area — rewrote both to state only current behavior.

### Gates
- gofmt (toolchain): clean
- go vet / -tags evenerfuzz / GOOS=windows -tags evenerfuzz: clean
- golangci-lint run ./internal/plugins/...: 0 issues
- go test -count=1 ./internal/plugins/... (plain and -tags evenerfuzz): ok

### Falsification
- `9fb24e11f`: reverted marketplaces.go/marketplace_migration.go/marketplace_migration_test.go/store_changed_test.go, `TestMigrationTriggeredSaveFailureNamesNoPath` failed (raw absolute path in the error text: `...write /.../known_marketplaces.json: permission denied`), restored cleanly.
- `2a27aa429`: reverted marketplaces.go, `TestEditMarketplaceRenameBetweenNamesRollbackFailureSurvivesIdentity` failed (`errors.Is` false), restored cleanly.
- Lows: test-only / comment-only; no production behavior to falsify against.
- `git status --short` clean after every revert/restore cycle.

### Head / diff
- Full 40-char head: `2a27aa429066748563cbecaef98739fa4705d335`
- `git diff origin/main...HEAD --stat`: 4 files changed, 620 insertions(+), 17 deletions(-) — `internal/plugins/marketplace_migration.go` and `internal/plugins/store_changed_test.go` are now byte-identical to `origin/main` (Medium A's fix fully absorbed the round-3 signature change back out of them).

obra:
## Round 5 disposition (2c46a9c6) — answering the raw panel reviews

Round 5 of 5. Both findings addressed; one fixed with a broader grep-audit as directed, one measured and filed as a follow-up (crash-recovery, outside this PR's contract).

**Medium (member 3): `marketplace_migration.go:935` leaks the marker's path via migration recovery** — fixed in `2c46a9c66`.

Grepped `internal/plugins` for every other `fmt.Errorf` embedding a path built from `m.storePath`/`marketplaceDir`/`cacheDir`/etc. Within this PR's own files (`marketplaces.go`, `marketplace_migration.go`), found two more instances of the *same shape* as the flagged one — all three call `m.storePath(renameMarkerFileName)` and interpolate the resolved absolute path directly into a message, and all three are reached over the wire the same way (`lockStore` runs the migration barrier on every marketplace verb; `loadMigratedMarketplaces` runs it whenever a marker is pending, so `ListMarketplaces`/`List` hit it directly):
- `markerLeftForRecovery` (the flagged one, called from `mergeIntoMigrated`, `migrateMarketplaceName`, and `markerAfterFailedMove`)
- `finishAnInterruptedRename`'s "another marketplace is recorded as ...; rename that one, or delete `<path>` to abandon the unfinished rename"
- `finishMarkedMerge`'s "no marketplace is recorded as ...; delete `<path>` to abandon the unfinished merge"

All three now name only the marker's bare filename (`marketplace-rename.json`); the real path goes to the hub's log via `m.stderr()`, matching `saveFailed`'s established pattern.

This turned up a real tension worth flagging explicitly: four *existing* tests (predating #1810) asserted the marker's absolute path reaching the caller was **intentional** — their own comments say things like "the error says where it is" and "naming both names and the marker the user can delete to abandon the rename." That's an older design philosophy (a local recovery aid) that #1810's own stated principle ("this machine's absolute plugin-store path" never reaches an RPC caller) supersedes. I updated those four assertions to check the bare filename plus a new explicit "no absolute path" check, and added a fifth test (`TestMarketplaceNameMigration_MergeSaveFailureKeepsTheMarkerNamesNoPath`) covering the specifically-flagged site directly (`mergeIntoMigrated`'s own `saveRename` failure, not just the two "destination missing" refusals). Wanted to surface this explicitly rather than silently override four tests' documented intent — please push back if the local-recovery-aid design was meant to survive this hardening pass.

**Two sites found by the same grep are *not* fixed here, deliberately** — filed as https://github.com/prime-radiant-inc/evener/issues/1883:
1. `loadMarketplaces`/`loadRegistry`'s read/parse errors (`marketplaces.go:73,77`, `registry.go:46,50,53`) — reached raw by all five verbs, but the same function backs `Doctor()`'s local diagnostics, which *want* the real path (no "hub's log" to redirect to there). Scrubbing the shared function would blunt Doctor; the right fix is a caller-side wrap at each verb, not a shared-function change — a separate PR.
2. `locks.go`'s lock-acquisition errors (`acquireLock`/`flockUntil`) — reached raw by every verb via `lockStore`, but those are package-level functions with no `*Manager`/logger to scrub through, and shared with `Install`/`Upgrade`/`Gc`'s own locking. Needs its own design (a typed error or a logger parameter), not a quick patch.

**Medium (member 0): no durable marker for `RemoveMarketplace`'s save-then-remove crash window** — measured, not fixed; filed as https://github.com/prime-radiant-inc/evener/issues/1882.

Confirmed via `git show origin/main:internal/plugins/marketplaces.go`: the pre-#1810 order (remove-then-save) had the identical class of gap, just inverted — a crash there left the marketplace **still registered** pointing at a now-missing clone, equally unrecorded anywhere. #1810's reorder moved which inconsistency a crash produces; it didn't introduce the absence of a durable marker for this operation (unlike the rename/merge machinery in `marketplace_migration.go`, which has exactly this kind of marker). Reproduced with a test (in the issue) showing the orphaned clone survives `Gc` and every other `Manager` method. This is a crash-recovery concern outside the path-scrub contract, as you said — the issue has the repro and a suggested fix (a durable marker mirroring `renameMarker`, recovered at the next `lockStore` acquisition).

### Gates
- gofmt (toolchain): clean
- go vet / -tags evenerfuzz / GOOS=windows -tags evenerfuzz: clean
- golangci-lint run ./internal/plugins/...: 0 issues
- go test -count=1 ./internal/plugins/... (plain and -tags evenerfuzz): ok

### Falsification
Reverted `marketplace_migration.go` only, reran all five affected tests: all five failed with the absolute `/.../marketplace-rename.json` path present in the error text. Restored; `git status --short` clean.

### Head / diff
- Full 40-char head: `2c46a9c6698e64a591cc606d3811738aa52cca91`
- `git diff origin/main...HEAD --stat`: 5 files changed, 714 insertions(+), 26 deletions(-)

obra:
Decomposed after five rounds into #1889, #1890, #1897 per the coordinator; closing.

PR #1813, opened 2026-09-18
fix(agent): read branch-level enum constraints from their combinator arm (#622)

### LINKED ISSUE #622: repair: constraintMessage reads top-level enum for branch-level (oneOf arm) enum constraints
## Summary

`repair.ExplainSchemaError`'s value-constraint renderer (`constraintMessage` in `agent/internal/tool/repair/explain.go`) looks up the failing property's `enum` in the **top-level** `properties` map. When the failing constraint is a `oneOf`/`anyOf`/`allOf` **branch-level** enum — a narrowed enum that applies only inside a branch — the walk resolves the property name against the top-level schema and renders the wrong allowed-value list.

This is a distinct defect from the arm-order attribution fragility (filed separately): even when the failing cause is correctly identified as living inside a branch, `constraintMessage` cannot see the branch's own `properties.<name>.enum` because it only consults `schemaProps(containerSchema)` at the top level.

## Reproduction (probe-verified, jsonschema v5.3.1)

Schema:

```json
{
  "type": "object",
  "properties": {
    "task": {"type": "string"},
    "sandbox": {"type": "string", "enum": ["off", "read-only", "workspace-write", "restricted"]},
    "sandbox_net": {"type": "boolean"}
  },
  "required": ["task"],
  "oneOf": [
    { "not": {"required": ["sandbox_net"]} },
    { "required": ["sandbox", "sandbox_net"],
      "properties": { "sandbox": { "enum": ["read-only", "workspace-write", "restricted"] } } }
  ]
}
```

With the arms in this order and args `{"task":"ping","sandbox":"off","sandbox_net":true}`, the deepest cause is `/oneOf/1/properties/sandbox/enum` with instance location `/sandbox`. Because the `not` arm sorts first in the cause list, the extracted keyword on the current fix is the `not` branch; but with the arms swapped (see the arm-order issue) the extracted keyword becomes `enum` and the message renders the top-level enum, listing `off` as allowed while the real branch enum forbids it.

## Impact

Any tool schema that narrows an enum inside a combinator branch (a common JSON-Schema idiom for conditional validation) gets an allowed-values list that doesn't reflect the actual constraint. Latent today: the only built-in oneOf producer is the delegate schema, whose failure shape (not-arm-first) takes the branch-message path of the #618 fix. Reachable through plugin/MCP tool schemas, which flow through the same generic repair code.

## Suggested fix direction

When the failing cause's KeywordLocation lives under a combinator (`/oneOf/N/properties/...`), resolve the property schema by walking the KeywordLocation into the schema (the branch's `properties`), not the top-level `properties`. This pairs naturally with the prefix-based attribution fix in the sibling issue.

## Environment

obra:
Fixes #622

## Root cause

`constraintMessage` resolved the failing field's schema with
`schemaProps(containerSchema)[field]` — always the top-level `properties` map.
`resolveSchemaErrorContainer` walks the *instance* location and cannot follow
combinator branches, so when the deepest cause was a branch-level enum such as
`/oneOf/0/properties/sandbox/enum`, the renderer read the same-named top-level
`sandbox.enum`. In the issue's reproduction the top-level enum allows `"off"`
while the oneOf arm forbids it, producing a self-contradictory message:

    delegate: argument "sandbox" is not one of the allowed values: "off", "read-only", "workspace-write", "restricted". Value is "off".

Only the deepest cause's *last* keyword segment reached the renderer
(`offendingKeyword`), so the branch path was discarded before it could be used.

## Change

- `offendingKeyword` → `offendingKeywordLocation` (`agent/session_tool_repair.go`):
  returns the deepest cause's full `KeywordLocation` (`oneOf/0/properties/sandbox/enum`)
  instead of only its final segment.
- `ExplainSchemaError` now takes that location and derives the keyword name from
  its last segment (a bare keyword name is a single-segment location, so callers
  without a path are unchanged).
- New `keywordParentSchema`/`constraintFieldSchema` (`explain.go`) walk the
  location's leading path — `properties/<name>`, `items`, `oneOf`/`anyOf`/`allOf`
  branch indices, `not`, `$defs`/`definitions` — into the root schema and resolve
  the field's own schema from the branch that actually carries the constraint.
  An empty or unwalkable location falls back to the container's `properties`
  map, preserving the flat behavior.

After the fix the reproduction renders:

    delegate: argument "sandbox" is not one of the allowed values: "read-only", "workspace-write", "restricted". Value is "off".

## Test evidence

- `agent/internal/tool/repair/explain_oneof_test.go`:
  - `TestExplainSchemaError_BranchEnumUsesBranchAllowedValues` — arms swapped so
    the deepest cause is the branch enum; asserts the exact branch-rendered
    message. Fails before the fix with the top-level list (`"off"` included).
  - `TestExplainSchemaError_BareKeywordStillReadsContainerEnum` — bare-keyword
    fallback still reads the container enum.
- `agent/session_tool_repair_issue622_test.go`:
  `TestOffendingKeywordLocationBranchEnum` — compiles the issue schema,
  validates, asserts `offendingField`/`offendingKeywordLocation`, and checks the
  end-to-end `ExplainSchemaError` output.
- Commands run:
  - `go test ./agent/internal/tool/repair/... -count=1` → ok
  - `go test ./agent/ -run 'Issue622|TestPrepareToolCall|TestExplainSchemaError|OneOf' -count=1` → ok
  - `go vet ./agent/internal/tool/repair/...` → clean
  - `gofmt -l <changed files>` → clean


obra:
Refutation of the earlier "`not` fallback can render contradictory guidance" finding (unreachable path).

Reproduction run against `santhosh-tekuri/jsonschema` v5.3.1, schema `oneOf:[{not:{properties:{color:{enum:["red"]}}}}]`, args `{"task":"t","color":"red"}`:

```
offendingField="" offendingKeywordLocation="/oneOf/0/not"
```

A failing `not` surfaces as keyword `not` at the combinator with an empty instance location, never as an inner location such as `/oneOf/0/not/properties/color/enum`. So real validation output takes the `isBranchKeyword("not")` path, not `constraintMessage` — `constraintFieldSchema`'s negation fallback is defensive only. Pinned by `TestNotFailureKeywordLocationIsTheCombinator` in `agent/session_tool_repair_issue622_test.go`.

obra:
Parking this PR rather than taking it to a sixth round.

The roborev finding count on the same example-validation code grew rather than shrank across the five fix commits: 4 issues at `802f506b95`, 4 at `b920fdc37a`, 3 Medium at `1702e99529`, and **6 Medium + 2 Low** at `ac7e56c967`. Each round's fix was correct for the keyword it addressed and exposed the next unmodeled one. That is a structural problem upstream of this PR's scope, not a queue of small defects — the guard is hand-rolling JSON Schema validation keyword by keyword.

The PR's own goal (branch-level enum constraints for #622) was reached. The outstanding findings and the structural diagnosis are filed as #2083, which asks for a decision: validate emitted examples with a real JSON Schema implementation, or explicitly close the set of keywords the explainer will certify and make everything unmodeled fail closed.

Leaving this open and parked, with the branch preserved, following the same handling as PRs #2019, #2013, and #2034. Never merged.

Issue #293, opened 2026-08-20
coverage-floor-selftest.sh fakes the go toolchain, which testing.md bans outright


obra:
## The violation

`docs/developing-evener/testing.md` states the policy without qualification:

> **A selftest earns its gate slot by pinning outcomes of a tool the gate or CI depends on.** [...] Fake-toolchain selftests are banned outright, not merely denied a gate slot: a tool that can only be tested by faking `go`, `gh`, or its sibling tools on PATH has outgrown shell, and the move is the port (the dev-tooling spec's port-on-touch trigger), never the suite.

`scripts/coverage/coverage-floor-selftest.sh` opens by describing itself as exactly that:

> exercises scripts/coverage-floor.sh against a throwaway repo and **a fake `go`** that emits two DIFFERENT profiles depending on whether it was invoked with `-tags evenerfuzz`. No compilation, no real suite.

It builds `fake_bin` and puts it on `PATH` (`:28-29`). The suite is registered as `coverage/coverage-floor` in `DEV_TOOLING_TEST_SCRIPTS` (`Makefile:149`), so it runs in `make test-dev-tooling` and in CI.

The policy does carve out data-seam fixtures — "the registry listing fuzz-bisect's and fuzz-oracle-audit's suites stub while the git history, replays, and verdicts stay real" — but faking the `go` binary itself is the specifically named prohibited case, not a data seam.

## Why it can only be tested this way

`scripts/coverage/coverage-floor.sh` invokes the toolchain inline:

- `:121` — `go test -count=1 $short -coverpkg=... -coverprofile="$base.test.cov"` (test track)
- `:129` — `go test -tags evenerfuzz -count=1 -coverpkg=... -coverprofile="$base.fuzz.cov"` (fuzz track)

There is no seam for supplying pre-made profiles, so the only way to exercise the union arithmetic is to replace `go` itself. The policy's diagnosis is precise: the tool has outgrown shell.

## What the suite actually covers

Two things, and they are not equally valuable.

**The union arithmetic — worth keeping.** The fixture is chosen so the union cannot be mistaken for either input: the test track covers only block A, the fuzz track covers only block B, so a correct union reports 100% where each track alone reports 40% and 60%. This is the load-bearing two-track property — the one `docs/developing-evener/coverage.md` is built around — and it is pure computation over two coverage profiles.

[...]

PR #1816, opened 2026-09-18
refactor(coverage): count coverage-gaps and e2e-cover through the Go covstmt primitive (#616)

### LINKED ISSUE #616: Consolidate the remaining Python statement-counters (coverage-gaps.sh, e2e-cover.sh) onto the Go covstmt primitive
### Context

From the adversarial review of the #293 fix (wiring the Go `covstmt` primitive into coverage-floor.sh and deleting the Python duplicate in covstmt-lib.sh). Reviewer B found two sibling Python statement-counters that survive the consolidation:

- `scripts/coverage/coverage-gaps.sh:50-58` — embedded Python with the same dedup/any-hit block-parsing regex (a third sibling of the family #293 unified)
- `scripts/coverage/e2e-cover.sh:162-181` — another embedded Python statement counter

### Why this matters

#293's root cause was drift between two implementations of per-statement coverage analysis. The fix unified coverage-floor.sh onto `evener dev covstmt`, but the same drift risk remains in these two scripts: each carries its own regex-based counter that will silently diverge from the Go primitive as either evolves.

### Suggested shape

Wire both scripts to `go run ./cmd/evener-dev/bin dev covstmt` (or a `--gaps`-style mode if the ranking/output shape needs more than `covered total` lines — coverage-gaps.sh ranks gaps, so it may need an extended output mode rather than the plain counter). Follow the #293 PR's pattern: equivalence evidence on identical fixtures before the swap, selftests passing before and after, delete the Python after.

(Discovered during the fix-program review of #293; filed per the out-of-scope rule.)

obra:
## Summary

Fixes #616.

`scripts/coverage/coverage-gaps.sh` and `scripts/coverage/e2e-cover.sh` each embedded their own Python statement-counter: a block-line regex, position dedup, and any-hit union — the exact algorithm `internal/devtool/covstmt` already owns and tests. #293 (PR #619) unified `coverage-floor.sh` onto the Go primitive but left these two siblings carrying copies free to drift as either side evolves.

## Root cause

Two (three, counting the deleted `covstmt-lib.sh`) implementations of per-statement coverage analysis. The scripts' Python parsed fine, so the drift was invisible: they would silently disagree with the ratchet's primitive over time.

## What changed

- **`internal/devtool/covstmt/covstmt.go`**: extracted the profile parse into a shared `parseProfile`, and added `Block` + `Blocks`/`BlocksReader` — the per-block companion to `StmtCounts`, built on the same parse so the two can never disagree about what a block is or whether it is covered.
- **`cmd/evener-dev/covstmt.go`**: added `evener dev covstmt --gaps --by package|file --top N --zero --in SUBSTRING`, reproducing `coverage-gaps.sh`'s ranking report **byte-for-byte**. The counting is `covstmt.Blocks`; only the ranking/formatting was ported.
- **`scripts/coverage/coverage-gaps.sh`**: keeps its flags, usage, and validation; the embedded Python is replaced by a forwarding call to `evener dev covstmt --gaps`.
- **`scripts/coverage/e2e-cover.sh`**: the `--merge-unit` union is now `cat unit.prof e2e.prof | evener dev covstmt` — concatenation + the primitive's position-dedup/any-hit arithmetic, the same accounting `coverage-floor.sh` uses.
- **`coverage_covstmt_audit_test.go`** (new): guards that both scripts invoke the Go primitive and run no `python3`.

## Equivalence evidence

- Synthetic adversarial fixture (dedup, any-hit union, two packages, a slash-less top-level file, covered/uncovered blocks): old script (from `HEAD`) vs new, **byte-identical** across `--by package`, `--by file`, `--by file --zero`, `--by package --top 2`, `--in file.go`.
- Real profile: `identifier` module test track (269 lines) — old vs new byte-identical across the same five flag shapes.
- `e2e-cover` union block: old Python vs new `cat`+`covstmt` byte-identical on identical fixtures.

## Testing

- `go test ./internal/devtool/covstmt/ ./cmd/evener-dev/ ./cmd/evener-dev/bin/` — PASS (adds `Blocks` property tests and gaps-mode contract tests pinned to the Python golden output).
- `go test -run TestCoverageScriptsUseGoCovstmt .` — PASS (guard would fail on the pre-change scripts: `grep -c python3` = 1 in each at `HEAD`, 0 now).
- `bash -n` on both scripts; error paths (`no profile`, `no such profile`, `--by bogus`, unknown flag) preserve their exit codes; relative profile paths still work.
- `gofmt -l` clean; `go vet ./cmd/evener-dev/... ./internal/devtool/covstmt/... .` clean.

## Accepted divergence

`--top` with a negative value has no meaningful output in either implementation; the Go report clamps rather than reproducing Python's negative-slice semantics. Real invocations pass a positive count.


obra:
**Round-4: the two threads on `coverage_covstmt_audit_test.go` contradict each other, and `docs/developing-evener/testing.md` decides it.**

The doc is explicit:

> **Faking the toolchain to test a script is banned outright**: a tool that can only be tested by faking `go`, `gh`, or its sibling tools on PATH has outgrown shell, and the move is the port (the dev-tooling spec's port-on-touch trigger), never a fixture harness built around the fake.

> the argv a script hands a faked binary is an implementation restated, and asserting on it is testing the mock.

> that every consumer actually goes through the library is a static audit's job, not a reason to re-run consumers under sabotage.

Resolution:

- **Review 1 (Medium, `coverage_covstmt_audit_test.go:11-115`) — upheld.** I removed the two PATH-stub tests (fake `go`, fake `cat`). They were exactly the banned pattern: a fixture harness built around a faked sibling tool.
- **Review 4 (Low, `coverage_covstmt_audit_test.go:128`) — does not hold as stated.** Its remedy ("remove the source scan and test … with the same fake-`go` approach") is itself the banned pattern, and the doc assigns "every consumer goes through the library" to a static audit. I kept the static audit, now stripping full-line comments, and satisfied the *intent* (behavioral, not textual) without faking: `TestCoverageGapsScriptReportsFixture` runs the **real** `coverage-gaps.sh` and the **real** `evener-dev/bin dev covstmt` against a fixture and pins the report bytes and exit code.

**Reproduction — the static audit still catches a reintroduced counter** (so it is not a substring unrelated text can satisfy):

```
$ printf '\npython3 -c "pass"\n' >> scripts/coverage/coverage-gaps.sh
$ go test -count=1 -run TestCoverageScriptsUseGoCovstmt .
--- FAIL: TestCoverageScriptsUseGoCovstmt (0.00s)
    coverage_covstmt_audit_test.go:95: scripts/coverage/coverage-gaps.sh still runs python3; its statement-counter was supposed to be deleted in favor of the Go covstmt primitive
FAIL
```

`coverage-gaps.sh` was restored byte-identically afterward (sha256 `2ad07f96badfd7de758ec75b144f4a6aac90ceed`).

**Also fixed this round (both reproduced):**

- Medium (`coverage-gaps.sh:53/58`, CDPATH): reproduced with a real run — `CDPATH=. bash scripts/coverage/coverage-gaps.sh <fixture>` failed `rc=1` with `cd: $'…\n…': No such file or directory` because `cd` echoed the resolved dir into the command substitution, producing a two-line `repo_root`. Both resolutions now use `CDPATH='' cd -- … && pwd`, checked, in `coverage-gaps.sh` and `e2e-cover.sh`. `TestCoverageGapsScriptReportsFixture` runs each CDPATH case and pins identical output.
- Low (`e2e-cover.sh` awk locale): the percentage now runs under `LC_NUMERIC=C`, so a comma-decimal locale cannot produce `50,0` and fail the numeric-shape check.

Issue #725, opened 2026-08-31
PastIndex: incremental FTS upsert instead of whole-table rewrite per publish


obra:
## Summary
Split from #723: the concrete optimization direction for the FTS rewrite cost. `rebuildFTS` does DELETE-all + re-INSERT of every row inside one transaction per publish; an incremental upsert (UPDATE sort_rank >= pos then one INSERT) would make a single-session fold or rename O(shifted tail) instead of O(index).

## Evidence
- ~13.5k entries on this machine; ~275-437ms per full rewrite (measured during the #722 review loop).
- The transaction-per-publish also provides the last-writer-wins semantics that keep concurrent publishes consistent — any incremental path must preserve that.

## Related
- #645 / PR #722 introduced the fold as a new publish caller
- #723 tracks the fold-vs-Rebuild stale-publish race

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

**Verdict:** `duplicate` — same defect and fix surface as #723; this issue states it was "Split from #723: the concrete optimization direction." Both cite `rebuildFTS`'s DELETE-all + re-INSERT per publish (`cmd/evener-hub/internal/hubcore/past.go:495-543`) and the same incremental-upsert remedy. There is no independent mechanism here.

**Recommendation:** `close` — fold this into #723 so the incremental-upsert design (and its last-writer-wins constraint) is decided once.

PR #1817, opened 2026-09-18
fix(hub): keep a PastIndex fold from a concurrent Rebuild's stale publish (#724)

### LINKED ISSUE #724: PastIndex fold can be clobbered by a concurrent Rebuild's stale-snapshot publish
## Summary
A Find-miss fold can be transiently lost when a concurrent Rebuild's publish lands after it: Rebuild swaps `i.all`/`i.byID` under the lock from its own (already-stale) snapshot, then calls `publishAndSignal` after releasing it. If a fold lands between Rebuild's disk scan and its publish, the Rebuild's publish clobbers the folded entry — it vanishes from the index and the fingerprint regresses (one spurious onChange fire on unchanged disk).

## Evidence
- Demonstrated deterministically by adversarial reviewer A during the #722 review loop: fold fires (1 real change) → stale publish fires again (spurious) → folded session dropped from `All()` → next Find self-heals by re-folding.
- The interleaving shape is pre-existing on origin/main (UpdateMeta vs Rebuild had the same lost-update); the fold path adds a new writer to it.
- Severity low: the window requires Rebuild to scan the disk BEFORE the session is persisted and publish AFTER the fold — the 60s ticker makes it a once-per-minute opportunity, and the next Find of that id self-heals. No corruption; `-race` dense stress (hundreds of folds racing continuous rebuilds) shows no duplicates or permanently lost entries.

## Fix direction
A generation/sequence check in the publish tail (skip the publish if `i.byID` has grown past the snapshot the publisher built from), or merging entries not present in the snapshot under the lock.

obra:
Fixes #724

## Re-scope after #1819 landed

Rebasing onto current `main` collided with the sibling #1819, which had already
merged a #724 fix (Rebuild rescans when a mutation supersedes its unlocked scan,
bounded to 3 attempts) together with the incremental `publishFTS` (#725). That
rescan approach supersedes this branch's earlier merge/revalidation machinery, so
those commits are dropped; this PR now carries the #724 gaps `main` still has.

## Remaining fixes

`cmd/evener-hub/internal/hubcore/past.go`:

- **`foldOne` kept a stale indexed row.** A Rebuild can swap in the row it scanned
  (v1) after a Find probe already read a newer on-disk meta (v2); the old
  unconditional early return kept v1, so `All`/`Search` served the stale row and
  no `onChange` fired. `foldOne` now replaces the indexed row when the probe's
  meta is newer, using `metaNewer` (`UpdatedAt`, then `NameUpdatedAt` — a rename
  preserves `UpdatedAt` and stamps `NameUpdatedAt`).
- **`Find` returned the probe's meta** even when `foldOne` kept a strictly newer
  indexed row. `Find` now re-reads the cache after folding and returns the live
  row.

## Tests

Registered in `FuzzHubcoreScenarios`; each fails on the pre-fix path:

- `fuzzScenarioPastIndex_FoldReplacesStalerIndexedRow` — probe v2 over scanned v1.
- `fuzzScenarioPastIndex_FoldReplacesStalerIndexedRowOnRename` — rename-only v2
  (same `UpdatedAt`, newer `NameUpdatedAt`).
- `fuzzScenarioPastIndex_FindReturnsLiveRowAfterFold` — a concurrent writer
  indexes a newer row between Find's cache lookup and `foldOne`.

## Root-cause note on the deleted-merge findings

The prior review's "session resurrection via concurrent rebuild merge" findings
concern this branch's now-removed merge path; `main` has no merge (a superseded
scan rescans instead), so a deletion racing an edit/fold cannot be resurrected by
a stale-scan merge. Nothing to fix here.

## Verification

- `go test ./cmd/evener-hub/internal/hubcore/... -race -count=1` → ok
- `gofmt -l` clean; `go vet ./cmd/evener-hub/internal/hubcore/...` clean
- `golangci-lint run ./cmd/evener-hub/internal/hubcore/...` → 0 issues
- `make fuzz-registry-check` → ok


obra:
Refuting roborev's High on f0ed326 ("afero.NewOsFs() returns an afero.OsFs value, so the capability check never matches"): the capability check is correct and production cross-process locking does engage.

Evidence from the pinned dependency (afero v1.12.0, os.go):

    type OsFs struct{}
    func NewOsFs() Fs {
        return &OsFs{}
    }

NewOsFs returns &OsFs{}, so the concrete type is *afero.OsFs and `fs.(*afero.OsFs)` matches. Production passes afero.NewOsFs() through SaveSessionMeta -> lockSessionMetaCrossProcess, so the flock is taken. The tests that pin this (agent/schema TestSaveSessionMetaWaitsForCrossProcessLock, TestAppendSessionObservedByLoadsUnderCrossProcessLock) exercise the real OS fs and would hang/fail if the check stopped matching.

I did not widen the check to accept the value form: the value type is not what production constructs, and matching it would claim locking on a filesystem that cannot be locked.

The other findings from that review are fixed in the next push: deletion now writes a metadata tombstone under the same lock so an in-flight autosave cannot recreate a deleted session (Medium 1); `evict` re-validates rebuildGen/evictGen inside the lock and declines on stale generations (Medium 2); `pastAfterFindCacheMiss` is now a PastIndex field (Medium 3); and the rename reload error is surfaced instead of falling through with stale metadata (Low).

obra:
Resolving roborev's low finding on `agent/schema/snapshot_flock_unix.go` (`//go:build unix` + stdlib `syscall.Flock`) by decision, not by a code change.

Jesse's ruling: solaris and aix are not target platforms; the docs were wrong, not the code.

- The shipped Go targets are `linux` and `darwin` (`.goreleaser.yml` builds `goos: [linux, darwin]` only); `windows` is kept whole only as the CI/editor cross-vet port. `solaris` and `aix` are `unix` by build constraint but are not targets, so a `//go:build unix` seam file that calls `syscall.Flock` failing to compile there is acceptable.
- The code is therefore **deliberately left alone**: no switch to `golang.org/x/sys/unix` and no narrowed build tag. Either would be a change the supported targets do not need.
- The real defect was the doc's unqualified invariant. `docs/developing-evener/conventions/go-workspace.md` ("Platform-specific code gets a seam…") now names the supported GOOS set explicitly and says the cross-port claim applies to those targets only; it also notes the CI cross-vet covers `windows` (where the `!unix` fallback is the source in scope), which is why a Unix-only syscall is invisible to it.

New head: `370c7c14ce13e260609cc9318a427d4a6483b9b5` (rebased onto `origin/main` @ `528e1a230d`; the three prior fixes are unchanged).

Issue #723, opened 2026-08-31
PastIndex publish tail rewrites the entire FTS table per publish


obra:
## Summary
`publishAndSignal` (cmd/evener-hub/internal/hubcore/past.go) calls `rebuildFTS(all)` on every publish — a full SQLite DELETE-all + re-INSERT of every entry (`past_sessions_fts`). At ~13.5k entries this measures ~275-437ms per publish, and it fires on every 60s Rebuild tick, every UpdateMeta (rename/RefreshOne path), and every probe fold.

## Context
Surfaced during the #645 review loop (PR #722). The shared-tail design is deliberate — it is what makes the three writers (Rebuild/UpdateMeta/foldOne) consistent — and the fix direction is an incremental FTS upsert (UPDATE sort_rank >= pos, then one INSERT) rather than a whole-table replace, which would drop the per-publish cost from O(index) to O(shifted tail).

## Evidence
- Reviewers measured ~275ms per rewrite at 2,615 entries and ~437ms at 13.5k (the FTS rewrite dominates the full Rebuild cost of ~400ms-1.05s).
- The mid-Rebuild window also admits one redundant full rewrite when a Search lands between the staleness mark and the rebuild's own publish (bounded, idempotent, benign — measured, no convoy under -race storms).

## Notes
- UpdateMeta already pays this per rename; the probe fold added a new caller (once per newly-persisted session until the tick folds it).
- Any fix must preserve last-writer-wins snapshot semantics across concurrent publishes (the current transaction-per-publish provides that).

obra:
**Independent triage verification** (at `27503c07dacc`, 2026-09-13).

**Verdict:** `legit-current` — `publishAndSignal` calls `rebuildFTS(all)` on every publish (`cmd/evener-hub/internal/hubcore/past.go:346-361`), and `rebuildFTS` does `DELETE FROM past_sessions_fts` then re-inserts every row (`:495-543`, delete at `:522`). The hub always wires a DB path (`cmd/evener-hub/main.go:226`) and the 60s ticker drives `Rebuild` (`main.go:462-469`; default `cmd/evener-hub/config.go:71`), so the per-publish O(index) cost is live. This is a performance optimization, not a correctness bug.

**Autonomy:** `autonomous` — the fix direction is specified (incremental upsert preserving the transaction-per-publish last-writer-wins), it is bounded to `rebuildFTS`/`publishAndSignal`, and existing `past_test.go` ordering/race tests plus `-race` verify it.

[...]

PR #1819, opened 2026-09-18
perf(hub): publish FTS index incrementally instead of rewriting it (#725)

### LINKED ISSUE #725: PastIndex: incremental FTS upsert instead of whole-table rewrite per publish

obra:
## Root cause

`PastIndex.publishAndSignal` mirrored every publish into SQLite by calling
`rebuildFTS(all)`: a `DELETE FROM past_sessions_fts` followed by a re-`INSERT`
of every row, inside one transaction. At ~13.5k entries that is ~275-437ms of
FTS re-tokenization per publish, and it fired on every publish — a probe fold of
one newly persisted session, a rename via `UpdateMeta`, the 60s `Rebuild` tick,
and `Search`'s stale-mirror repair.

## Change

`publishAndSignal` now delegates to `publishFTS`, which keeps the last
successfully published snapshot (`published`) and transitions the SQLite table
to the new snapshot with a delta — deleting removed ids and (re)inserting only
new or content-changed ids — instead of rewriting the table:

- A single-session fold writes one row; a rename rewrites one row.
- A no-op publish (the common steady-state `Rebuild` tick) opens no database at
  all.
- It falls back to a full `rebuildFTS` when no usable prior snapshot exists
  (first publish, or a previous failure marked the mirror stale).

**Last-writer-wins is preserved.** A new `ftsMu` serializes every write and
`published` tracks exactly what the table contains, so a concurrent publisher
can never apply one snapshot's delta onto another snapshot's table; the last
publish to take the lock still wins, exactly as the transaction-per-publish
rewrite provided. `i.fts` is still marked stale for the duration of the write,
so a failed publish stays visible to `Search`'s repair gate.

The literal "`UPDATE sort_rank >= pos` then one `INSERT`" direction from the
issue was measured first: FTS5 rewrites a row's content when an `UNINDEXED`
column changes, so renumbering the tail costs about the same per row as
reinserting it, and a probe fold sorts to the *front* (newest first) where the
tail is the whole index — `UPDATE`-all measured ~220ms vs ~234ms for the full
rewrite at 13.5k rows. Not renumbering at all is what actually avoids the
rewrite. `sort_rank` only feeds `searchFTS`'s internal `ORDER BY`; `Search`
re-sorts the FTS ∪ memory union by the in-memory index (`mergeSearchResults`),
so an unchanged row's stale rank is not observable.

## Test evidence

`fuzzScenarioPastIndex_IncrementalPublishLeavesUnchangedRows` (new, registered
in `FuzzHubcoreScenarios`) reads the FTS mirror directly and asserts that a fold
and a rename leave every untouched row's stored `sort_rank` unchanged, while the
new/renamed row is mirrored and still served by `searchFTS`.

Fails on the pre-change implementation, passes after:

```
$ git checkout -- past.go && go test -run FuzzHubcoreScenarios -count=1 .
--- FAIL: FuzzHubcoreScenarios/seed#99
    past_test.go:1306: fold renumbered unchanged row 02wMz5Txv1C3Hut0M8GCeB (rank 1 -> 2); the whole table was rewritten
FAIL

$ go test -run FuzzHubcoreScenarios -count=1 .
ok  	primeradiant.com/evener/cmd/evener-hub/internal/hubcore	3.748s
```

Full focused gate:

```
$ gofmt -l past.go past_test.go scenarios_fuzz_test.go   # (no output)
$ go vet ./cmd/evener-hub/...                            # exit 0
$ go test -count=1 ./cmd/evener-hub/internal/hubcore/... # ok 33.042s
$ go test -race -count=1 ./cmd/evener-hub/internal/hubcore/... # ok 45.261s
```

Fixes #725


obra:
Addressed in b62003fd2.

**Medium — out-of-order publish (`past.go` publishFTS).** The delta baseline `i.published` is read *after* `ftsMu` is acquired (`publishFTS` locks `ftsMu`, then reads `i.published`), so it is always the table's actual current content: each delta transforms between adjacent snapshots and the table can never become a non-snapshot mix. The residual ordering hazard (an older snapshot's publisher taking `ftsMu` after a newer one) is the fold-vs-Rebuild stale-publish race tracked in #723 and fixed by PR #1817 (#724), which adds exactly the suggested generation guard in `publishAndSignal`. This lane (#725) is scoped to the FTS rewrite cost and deliberately does not duplicate #1817's guard — #1817 changes the same `publishAndSignal` lines. I documented the adjacency argument and the residual race ownership in `publishFTS`.

**Medium — external DB loss.** Fixed. `rewriteFTSDelta` now verifies, inside the transaction, `SELECT COUNT(*) FROM past_sessions_fts == len(prev)`; on mismatch it returns `errFTSBaselineStale` and `publishFTS` falls back to a full `rebuildFTS(all)`, restoring the self-healing the old whole-table rewrite provided. Covered by `fuzzScenarioPastIndex_IncrementalPublishRecoversFromLostDB` (delete the DB file + sidecars, then UpdateMeta; asserts all rows survive). The test fails with 1 row when the guard is removed.

**Low — pure-removal branch untested.** Fixed: `fuzzScenarioPastIndex_IncrementalPublishRemovesRows` removes a session's meta file, runs `Rebuild`, and asserts the id is gone from `past_sessions_fts` while the surviving row keeps its stored `sort_rank`.

**Low — stale `UpdateMeta` doc.** Fixed: the comment now describes the incremental publish (only the renamed row is rewritten; unchanged rows keep their stored `sort_rank`; `Search` re-sorts from the in-memory index).

Issue #618, opened 2026-08-30
delegate: oneOf sandbox/sandbox_net constraint failures are misreported as 'Required arguments: task (string)'


obra:
## Summary

Calling the `delegate` tool with `sandbox: "off"` and `sandbox_net: true` fails schema validation — correctly — but the error surfaced to the model is a confidently wrong diagnosis: **"Required arguments: task (string)"**, even though `task` is present and valid. The real failure is a `oneOf` constraint on the `sandbox`/`sandbox_net` pairing that the error-explanation layer cannot attribute.

## Reproduction

Session conditions (observed on a real session, macOS, sandbox mode `off` for the parent):

- `sandbox-exec` exists at `/usr/bin/sandbox-exec`, so `delegateSandboxBackendAvailable` is true → the delegate schema is built with `Available: true` (both `sandbox` and `sandbox_net` present in `properties`).
- Parent sandbox mode is `off`, so `delegateSandboxSchemaForEnv` sets `RequireNonOffModeForNetwork: true` (`agent/sandbox_delegate.go:100-104`), which adds this to the delegate tool schema in `DefDelegateWithSandbox` (`agent/internal/tool/definitions.go:228-240`):

```json
"oneOf": [
  { "not": { "required": ["sandbox_net"] } },
  { "required": ["sandbox", "sandbox_net"],
    "properties": { "sandbox": { "enum": ["read-only", "workspace-write", "restricted"] } } }
]
```

A call with these arguments fails:

```json
{ "task": "ping", "agent_type": "subagent", "isolation": "worktree",
  "sandbox": "off", "sandbox_net": true, "reasoning_effort": "low",
  "delegation_allowance": 0 }
```

Raw validation error (correct behavior — the parent's own mode is `off` with network on, and the floor logic forbids requesting that combination explicitly):

```
jsonschema: '' does not validate with urn:evener:tool-schema#/oneOf/0/not: not failed
```

Verified by compiling the exact schema via the registry's own `compileSchema` path and validating the exact argument shape:

```
MY ARGS FAIL: jsonschema: '' does not validate with urn:evener:tool-schema#/oneOf/0/not: not failed
```

## The bug

After raw validation fails, `prepareToolCall` (`agent/session_tool_repair.go:82-99`) runs the repair pipeline:

1. `repair.RepairArgs` correctly leaves `sandbox`/`sandbox_net` untouched (they are legitimate schema fields, not hallucinated keys), so the healed arguments **still fail** validation:

```
[...]

PR #1820, opened 2026-09-18
fix(repair): attribute oneOf failures by keyword-location prefix (#621)

### LINKED ISSUE #621: repair: oneOf attribution is arm-order fragile — positive-arm-first schemas get a wrong allowed-values message
## Summary

The schema-error explainer (`repair.ExplainSchemaError`, `agent/internal/tool/repair/explain.go`) attributes a validation failure by walking the deepest **first** cause of the jsonschema `ValidationError` tree and extracting the last segment of its `KeywordLocation` (`offendingKeyword` in `agent/session_tool_repair.go`). For the delegate oneOf constraint this is fragile in a way that produces a wrong message: the keyword extracted depends on the **order of the oneOf arms**, not on the nature of the failure.

Found by adversarial review of the #618 fix (probe-verified with santhosh-tekuri/jsonschema v5.3.1, the same library the registry uses).

## Reproduction

Take the delegate oneOf constraint (`agent/internal/tool/definitions.go`, `DefDelegateWithSandbox` with `RequireNonOffModeForNetwork`) but swap the arm order — positive arm first, `not` arm second:

```json
"oneOf": [
  { "required": ["sandbox", "sandbox_net"],
    "properties": { "sandbox": { "enum": ["read-only", "workspace-write", "restricted"] } } },
  { "not": { "required": ["sandbox_net"] } }
]
```

Call with `{"task":"ping","sandbox":"off","sandbox_net":true}`. The deepest-first cause is now `/oneOf/0/properties/sandbox/enum` with `InstanceLocation="/sandbox"` — field present, keyword `enum` — so the message renders via `constraintMessage` against the **top-level** `properties.sandbox` enum:

```
delegate: argument "sandbox" is not one of the allowed values: off, read-only, workspace-write, restricted. Value is "off".
```

`"off"` **is** one of the listed allowed values. The real failure is the branch-level pairing rule (branch enum excludes `off` when `sandbox_net` is present); the walk never sees the branch's narrowed enum. The message actively lies, differently from #618's lie but in the same family.

## Why it's latent today

`definitions.go` hardcodes the `not` arm first, so the #618 shape (deepest cause = bare `/oneOf/0/not`, no property) is what production always produces, and the #618 fix handles it. Nothing protects against:
- a future reordering of the two arms in `DefDelegateWithSandbox`, or
- any other tool (built-in, plugin, or MCP) authoring a oneOf with a positive arm first.

The repair layer is generic infrastructure for arbitrary registered schemas, so MCP/plugin tool schemas can reach this path today.

## Suggested fix direction

Discriminate on the deepest cause's KeywordLocation **prefix** (`/oneOf`, `/anyOf`, `/allOf`, `/not`), not its last segment. A helper alongside `offendingKeyword` that reports whether the deepest cause lives under a branch combinator would cover bare `/oneOf` (multiple-match), `/oneOf/N/not` (#618 shape), and arm-internal keywords (`enum`, `required` inside an arm) uniformly, and would let the explainer render branch-level constraints instead of top-level ones. A regression test should pin the swapped-arm schema above: assert the message names the pairing rule / branch enum, not the top-level enum (and that it never claims `off` is disallowed generally).

## Environment

- probe: santhosh-tekuri/jsonschema v5.3.1, Draft 2020-12, compiled via the registry's own AddResource/Compile path

obra:
Fixes #621

## Root cause

`repair.ExplainSchemaError` attributed a validation failure by walking the deepest **first** cause of the jsonschema `ValidationError` tree and taking the last segment of its `KeywordLocation` (`offendingKeyword` in `agent/session_tool_repair.go`). For a root-level branch combinator that leaf is **arm-order dependent**:

- `not` arm first (the order `definitions.go` hardcoded): deepest cause is `/oneOf/0/not` → keyword `not` → the #618 branch path.
- positive arm first: deepest cause is `/oneOf/0/properties/sandbox/enum` → keyword `enum` → `constraintMessage` reads `sandbox`'s enum off the **top-level** property and announces a list that contains the rejected value:

```
delegate: argument "sandbox" is not one of the allowed values: "off", "read-only", "workspace-write", "restricted". Value is "off".
```

`"off"` is in that list. The real failure is the branch pairing rule (branch enum excludes `off` when `sandbox_net` is present).

## Change

Attribute on the deepest cause's `KeywordLocation` **prefix**, not its last segment:

- `offendingKeyword` now reports the root-level combinator enclosing the cause (`oneOf`, `anyOf`, `allOf`, `not`) via the new `keywordForCause` helper, instead of the leaf keyword it happens to carry. `ExplainSchemaError` therefore renders the branch-level constraints (branch requirements + narrowed branch enum) rather than reading the leaf keyword against a top-level property.
- `ExplainSchemaError` renders the branch constraints **before** the present-field path (previously the present-field path returned first, so a branch-internal `enum` never reached the branch renderer).
- `anyOf`/`allOf` attribute by name but still fall through to the generic message — rendering their branches needs per-keyword phrasing (tracked separately). Deliberately out of scope.

## Test evidence

Two new tests, both verified to fail on pre-fix code:

- `agent/internal/tool/repair/explain_oneof_test.go`: `TestExplainSchemaError_OneOfPositiveArmFirstDoesNotRenderTopLevelEnum` — swapped-arm schema, asserts the message names the branch enum and never the top-level one.
- `agent/session_tool_repair_issue621_test.go`: compiles the real schema through `santhosh-tekuri/jsonschema` v5.3.1 and asserts attribution is arm-order independent (`offendingKeyword == "oneOf"` for both arm orders) and that the end-to-end message is honest.

Pre-fix failure (revert-verified):

```
--- FAIL: TestExplainSchemaError_OneOfPositiveArmFirstDoesNotRenderTopLevelEnum
--- FAIL: TestOffendingKeyword_OneOfAttributionIsArmOrderIndependent/positive-arm-first
        offendingKeyword = "enum", want "oneOf"
--- FAIL: TestOffendingKeyword_OneOfAttributionIsArmOrderIndependent/not-arm-first
        offendingKeyword = "not", want "oneOf"
--- FAIL: TestExplainSchemaError_PositiveArmFirstNamesBranchRule
        branch-level enum failure rendered as a top-level allowed-values violation (issue #621):
        "delegate: argument \"sandbox\" is not one of the allowed values: \"off\", ... Value is \"off\"."
```

Post-fix:

```
go test ./agent/internal/tool/repair/   → ok
go test ./agent/ -run 'OneOf|OffendingKeyword|PositiveArmFirst|ExplainSchemaError|TestPrepareToolCall|...' → ok
gofmt -l <changed files> → clean
go vet ./agent/... → exit 0
```


obra:
Fixing-review follow-up (head `724fae31d`, rebased onto current main). Resolution of each roborev finding:

**Medium — `not`→`oneOf` fallback (branchList): fixed.** `branchList` no longer falls back to a sibling combinator: it reads only `params[keyword]`, and `isBranchKeyword` is `oneOf` alone, so a leaf `not` outside a oneOf renders the generic message. The stale hand-passed-`"not"` test now passes `"oneOf"` (matching the real caller), and `TestExplainSchemaError_RootNotDoesNotBorrowSiblingOneOf` covers the case the review asked for: a `oneOf` the call satisfies while a root `not` fails.

**Medium — `#/`-prefixed `KeywordLocation`: does not hold; reproducing below.** Review 3's note was correct. `jsonschema/v5@v5.3.1` builds every `KeywordLocation` from `keywordLocation()` in `compiler.go` as `"/" + path`, never a `#/` prefix. I compiled the delegate-style schemas through the real library and dumped the deepest first cause:

```
positive-arm-first : "/oneOf/0/properties/sandbox/enum"
not-arm-first      : "/oneOf/0/not"
root-not-plus-oneof: "/not"
ref-wrapped        : "/$ref/oneOf/0/required"
```

No `#/` appears in any case, so the `strings.Trim(loc, "/")` split already yields the combinator as `segs[0]` (after `$ref` skipping). No `#/`-stripping is needed; adding it would be unreachable code.

**Low — `anyOf`/`allOf` root failures: resolved by keeping the leaf keyword.** `keywordForCause` attributes a root combinator only when it is `oneOf`; a root `anyOf`/`allOf` returns its last location segment, so those failures keep the specific leaf constraint (e.g. `type`/`enum`) rather than a generic drop. Pinned by `TestKeywordForCause_Attribution` (`anyOf/0/type`→`type`, `allOf/1/enum`→`enum`).

**Low — `$ref`-wrapped `oneOf`: fixed.** `keywordForCause` skips leading `$ref` segments before testing for `oneOf`; `TestOffendingKeyword_RefWrappedOneOf` exercises `/$ref/oneOf/0/required` end-to-end through `offendingKeyword` + `ExplainSchemaError`.

**`lint-golangci`:** the failure was on the previous head `776808965`, not stale-base breakage — `modernize` on `errors.As`→`AsType[*jsonschema.ValidationError]` and `revive` error-return order in `session_tool_repair_issue621_test.go`. `724fae31d` already rewrote that helper (error last, `errors.AsType`); `make lint-golangci` now exits 0 on the new head.

obra:
Fixes for the `2ecdc0d` review, and two refutations with the reproductions I ran.

**High — `refWrappedKeywordLocation` misses the leading-slash trim: REFUTED.** The function does trim before splitting:

```go
func refWrappedKeywordLocation(keywordLocation string) bool {
	segs := strings.Split(strings.Trim(keywordLocation, "/"), "/")
	return len(segs) > 0 && isRefSegment(segs[0])
}
```

`strings.Trim("/$ref/not", "/")` → `"$ref/not"`, so for a validator location `/$ref/oneOf/...` the first segment is `$ref` and the gate fires. Reproduced directly:

```
ExplainSchemaError(loc="/$ref/maxLength", x present)  ->  "tool: arguments did not match the schema."
ExplainSchemaError(loc="/properties/x/$ref/maxLength") ->  "tool: argument \"x\" has the wrong type or value."
```

The first is the intended generic mismatch, so the gate works on the present-field path. The second is a *property-level* reference reached deeper in the path, which by design (per the previous round's review) reports a generic property-level mismatch naming the property while suppressing sibling constraint detail, `Required arguments`, and the example. `TestExplainSchemaError_RefWrappedPresentEnum` already asserts the first behaviour end to end through `santhosh-tekuri/jsonschema` v5.3.1 and passes.

**Medium (disputed) — `armOutcome` missing-required shortcut before reference detection: REFUTED.** The counterpoint is right. The pinned `jsonschema/v5 v5.3.1` compiles a schema with no `$schema` as Draft 2020-12, where `$ref` siblings apply, so an arm's `required` cannot be skipped by a sibling `$ref`. Reproduced:

```
schema:  {"$ref":"#/$defs/t","required":["a"],"$defs":{"t":{"type":"object"}}}
Validate({}) -> jsonschema: '' does not validate with ...#/required: missing properties: 'a'
```

The sibling `required` is applied, so the shortcut is exact for every schema this repo compiles. (Draft-07 `$ref`-sibling semantics would need a `"$schema": "http://json-schema.org/draft-07/schema#"` opt-in, which no in-tree or tool schema here carries.)

**Medium (confirmed, fixed at the root).** `branchRendersArmEnum` claimed coverage for `/oneOf/<i>/properties/<prop>/enum` whenever `<prop>` was in the arm's `required` list, but `branchRequirement` returns nothing for an arm whose `not` constrains a value, so its enum text is discarded and the caller emitted the bare generic mismatch. `branchRendersArmEnum` now returns false for such an arm, letting the present-field path name the property. Regression added (`TestExplainSchemaError_EnumInValueConstrainedNotArmNamesTheProperty`); it fails pre-fix with `"tool: arguments did not match the schema."` and passes after.

obra:
Settling the draft-07 `$ref`-sibling finding with evidence rather than argument, since I refuted this class in an earlier round and roborev has re-raised it.

**The finding holds — the path IS reachable under draft-07. I fixed it and did not re-refute.**

Where schemas are compiled: `agent/internal/tool/registry.go:1041` `compileSchema` → `compileSchemaWith` → `jsonschema.NewCompiler()` (`:1080`) with **no `c.Draft` set**. That is the generic registry path every tool schema takes, including MCP/plugin schemas. My earlier refutation assumed the library default (Draft 2020-12) decided the dialect; that assumption is wrong for a schema that declares its own `$schema`.

Experiment (pinned `santhosh-tekuri/jsonschema/v5 v5.3.1`, compiled through the same `NewCompiler` + `AddResource` + `Compile` path):

```
{"$ref":"#/$defs/t","required":["a"],"$defs":{"t":{"type":"object"}}}
    Validate({}) -> does not validate with ...#/required: missing properties: 'a'      (siblings APPLY)

{"$schema":"http://json-schema.org/draft-07/schema#",
 "$ref":"#/definitions/t","required":["a"],"definitions":{"t":{"type":"object"}}}
    Validate({}) -> <nil>                                                              (siblings IGNORED)

{"$schema":"https://json-schema.org/draft/2020-12/schema","$ref":"...","required":["a"],...}
    Validate({}) -> does not validate with ...#/required: missing properties: 'a'      (siblings APPLY)
```

So a schema that declares draft-07 is compiled with draft-07 semantics even though the compiler's draft was never pinned, and an MCP or plugin tool schema declaring draft-07 therefore reaches `armOutcome`/`branchRequirement`. No in-tree schema does today, but nothing on this path prevents it.

Fix, at the root and dialect-independent:
- `armOutcome` now returns "unknown" for an arm bearing a reference keyword (`$ref`/`$dynamicRef`/`$recursiveRef`) **before** applying the presence shortcuts, so such an arm can never be counted from presence under either dialect.
- `branchRequirement` describes nothing for a reference-bearing arm, because under draft-07 its sibling `required`/`not` do not apply and stating them would be false.

Regressions:
- `TestExplainSchemaError_RefBearingArmWithholdsExample` — fails pre-fix with `Branch 1 requires: send all of "c".` plus `Example: {"a": "..."}`, passes after.
- The requested validation invariant now covers this: `TestExplainSchemaError_EmittedExampleSatisfiesSchema` gained a draft-07 case and validates any emitted example against the complete schema. It fails pre-fix with the validator reporting `does not validate with ...#/oneOf: valid against schemas at indexes 0 and 1` — i.e. the example the gate certified matched **both** branches — and passes after. That converts this class of finding into a checked invariant.

PR #1822, opened 2026-09-18
hub: transcript-display write applied then failed is reported as applied

obra:
## Summary
- `TranscriptDisplayStore.Patch` renames the state and then, on a follow-up save error, returned that raw error even when the rename had already published the new revision - `app_rpc_transcript_display.go`'s plain `err != nil` check then skipped the broadcast entirely, leaving every other client on the pre-patch value until an unrelated notification.
- `hubcore.KeybindingsPostRenameError` already carries this exact distinction for the keybindings store (see `app_rpc_keybindings.go`'s `errors.AsType[*hubcore.KeybindingsPostRenameError]` seam). This PR adds `TranscriptDisplayPostApplyError`, mirroring that pattern: it wraps the underlying error and carries the applied `Layout`/`Applied` value, so the RPC layer (and the web client) can reconcile from it without a second read of the store.
- `app_rpc_transcript_display.go` catches the new type, broadcasts the applied value via `NotifyEvenerSettingsTranscriptDisplayChanged`, and returns a `WireError` carrying the new `appwire.TranscriptDisplayPostApplyData` (mirrors `KeybindingsPostRenameData`) so the requesting client gets the same information back.
- The web `transcriptDisplay.ts` store gains `postApplyDefault` (mirroring the existing `conflictCurrent`'s shape) and, on this error, adopts the applied value and clears the draft/error instead of surfacing an error banner for a write that actually landed.

## Test plan
- [x] `TestTranscriptDisplayStorePostRenameFailurePublishesNewState` (hubcore) extended with the `errors.AsType[*TranscriptDisplayPostApplyError]` assertion, mirroring `TestKeybindingsStorePostRenameFailurePublishesNewState`.
- [x] New `TestHubRPCTranscriptDisplayPatchPostApplyErrorStillBroadcasts` (RPC level), mirroring `TestHubRPCKeybindingsPatchPostRenameErrorStillBroadcasts`: asserts the wire error's `Data`, and that both the requesting and another client receive the broadcast.
- [x] New `transcriptDisplay.test.ts` case: "resolves a post-apply PATCH failure by adopting the applied value without an error" - asserts the promise resolves (not rejects), the draft clears, hub state adopts `applied`, and no error is set.
- [x] Falsification: reverted all four production files, confirmed both extended/new Go tests fail to compile on the missing type (`undefined: TranscriptDisplayPostApplyError` / `undefined: hubcore.NewTranscriptDisplayStoreForTest`), and the new web test throws the raw `WireError` instead of resolving; restored and reran green.
- [x] `$(go env GOROOT)/bin/gofmt -l appwire/ cmd/evener-hub/` - clean
- [x] `golangci-lint run ./appwire/... ./cmd/evener-hub/...` - 0 issues
- [x] `go test -count=1 ./appwire/... ./cmd/evener-hub/...` and `-tags evenerfuzz` - both green
- [x] `npx vitest run src/stores/transcriptDisplay.test.ts` (29 tests) - green
- [x] `npm run lint` (biome, from `cmd/evener-hub/frontend`) - clean
- [x] `npm run typecheck` (frontend) - clean (rebased onto main after #1826 repaired the pre-existing `gateRequests`/`gateSettlements` break)

Rebased onto main after #1826 (the unrelated web-typecheck repair) landed, so this head carries the fix.


obra:
## Round 2 disposition (605bfd9)

### Medium: applied-then-failed path reconciled before the fence check (transcriptDisplay.ts:618-634)

Measured the success path's fence order (patchHubDefault, same function): it checks
`patchTokens.get(layout) !== token || !isCurrentReady(client, generation)` and returns
the current hub state with **no mutation** *before* computing/applying `canonical`
(lines 569-571 for the "stale response, return early" branch; `applyHubDefault` for the
committed value only runs after that check, at line 611).

The applied-then-failed branch did the opposite: it called `applyHubDefault(layout, applied)`
unconditionally, and only fenced the *draft-clearing* state update afterward. A delayed
post-apply response from a token/generation that had already been superseded could still
overwrite hub[layout] with its stale `applied` value.

Fix: moved the fence check (`patchTokens.get(layout) !== token || !isCurrentReady(...)`) to
run first, matching the success path — on a stale token/generation it now returns
`transcriptDisplayStore.getState().hub[layout] ?? confirmed` without mutating anything, same
as the conflict-path's stale branch just below it. `applyHubDefault` and the draft-clear now
only run once the fence has passed.

TDD: added "ignores a stale post-apply reconciliation once a newer patch has committed" to
transcriptDisplay.test.ts, modelled on the existing "never lets a fenced revision hint
authorize a later active jump" test — two PATCHes to desktop, the second (newest) settles
first and commits normally, then the first (stale) rejects with the post-apply
transcriptDisplayPostApply error. Asserted hub.desktop stays at the second's committed
`{ revision: 2, config: newestCanonical }` and drafts.desktop is undefined.

Failing assertion (against unfixed code, before the fence reorder):
```
AssertionError: expected { revision: 1000, config: { …(3) } } to deeply equal { revision: 2, config: { …(3) } }
- Expected
+ Received
-       "level": "full",
+       "level": "activity",
-   "revision": 2,
+   "revision": 1000,
```

### Gates (from cmd/evener-hub/frontend)
- `npx vitest run src/stores/transcriptDisplay.test.ts` — 30 passed
- `npm run typecheck` — clean for this file; 36 pre-existing errors in
  src/panes/session/composer/skillDocument.ts and SkillEditor.tsx (missing
  prosemirror-* packages from node_modules, unrelated to this PR, last touched by #1649)
- `npm run lint` — clean (1357 files, no fixes)

### Falsification
Reverted only transcriptDisplay.ts to branch head (`git checkout HEAD -- transcriptDisplay.ts`),
ran the new test: FAILED with the assertion above. Restored via saved diff (`git apply`);
`git status --short` showed only the two intended files.

Head: 605bfd9345e95f60d441813f4c9d6719a2854d48

obra:
## Round N disposition (3e5e6c38) — answering the raw panel reviews

**Medium — app_rpc_transcript_display.go:34-47 / mobile-native/src/nativePreferences.ts:389+ (native doesn't handle the post-apply error contract)** — fixed in 3e5e6c388.

Measured how native handles the sibling `KeybindingsPostRenameError` first: it already gets it for free, because `mobile-native` imports the shared `createKeybindingsStore` from `@evener/appwire-client`, and that store's own `saveDraft` already reconciles from `KeybindingsPostRenameError`'s applied payload (`keybindingsStore.ts`'s `rejectionPayload`, keyed on `error.evenerErrorInfo`, not `error.code`). transcript-display has no equivalent shared store on the native side — `nativePreferences.ts` owns the whole PATCH flow locally (there's no `cmd/evener-hub/frontend/src/stores/transcriptDisplay.ts` counterpart in the shared package) — so the same handling had to be added there directly, not via the shared package.

`saveTranscript`'s catch block now runs a new `postApplyPatch(error)` helper (checks `error instanceof WireError && error.evenerErrorInfo === "transcriptDisplayPostApply"`, then decodes `error.data.applied` with the existing `decodeTranscriptPatch`) before falling back to the old "unconfirmed write" handling. When it matches, `value` is set to the applied `{revision, config}` and the function falls through into the same reconciliation path a successful response takes — no separate code path to keep in sync.

### Test
`nativePreferences.test.ts`: "resolves a post-apply PATCH failure by adopting the applied value, without blocking further edits" — forces the `transcriptDisplay/patch` handler to throw the same `WireError` shape `app_rpc_transcript_display.go` sends, asserts `saveTranscript` resolves (not rejects), `confirmed` reflects the applied revision/config, no draft/writeUncertain survives, and a follow-up `editTranscript` is not blocked.

### Falsification
Reverted `mobile-native/src/nativePreferences.ts` only, reran the new test: failed with `WireError: transcript display applied then a follow-up step failed` thrown out of `saveTranscript` (src/nativePreferences.test.ts:540, via `NativePreferences.saveTranscript src/nativePreferences.ts:383`). Restored, reran: 26/26 pass. `git status --short` clean afterward.

### Gates
- `mobile-native` had no `node_modules` at all in this worktree — not a shared symlink (`file`/`readlink` both show nothing to link to; `cmd/evener-hub/frontend/node_modules` is a real installed dir, not a symlink, either). `native-preflight.sh` refuses to install by design, so I ran `cd mobile-native && npm ci` once to give this worktree its own install, and I'm noting it here per the brief.
- `npx vitest run src/nativePreferences.test.ts`: 26 passed.
- `npm run check` (tsc, mobile-native): clean.
- No Go or `cmd/evener-hub/frontend` files touched this round, so their gates weren't run.
- Never ran biome over `mobile-native/`.

### Heads / diff
- Head (full 40 chars): `3e5e6c388eda422b56fc14c0e77799d562741ac4`
- `git diff origin/main...HEAD --stat`: 10 files changed, 343 insertions(+), 13 deletions(-) — this round only adds `mobile-native/src/nativePreferences.ts` (+55/-11 by that stat's line, actual round diff 44 insertions/9 deletions) and `mobile-native/src/nativePreferences.test.ts` (+34/-1); the other 8 files are unchanged from the prior round (605bfd934).

obra:
Round 3 @3e5e6c3: CI green (15); raw panel members (synthesis is broken): two found no issues, two the same Low (native postApplyPatch should verify data.layout === "mobile" before adopting the applied value). Merging per the Lows rule; the follow-up PR adds the layout check. Simplify by reading: the web and native decoders of the same wire error are the shape #1837 tracks.

PR #1833, opened 2026-09-18
test(server): pin that a live ask_user round's status frame carries askPending

obra:
## Summary

RoboRev round 2 on #1740 flagged a High: "Live status notifications omit AskPending" — `internal/appprojector/appwire_projection.go`'s `threadStatus()` helper never sets the field on the `ThreadStatusChangedParams` it builds.

That citation is correct, but it isn't a bug: the projector deliberately holds no session handle (`stampAskPendingOnStatusChange`'s own doc comment: "the projector maps events to notifications and holds no session handle, and the flag lives on the envelope"). The field is stamped afterward, from the live session, at the server's single notification egress (`server/appwire_runtime.go`'s `stampAskPendingOnStatusChange`, called from both `RecordAppEvent`'s and `finishProcessing`'s pending-notification loops), reading `envelopeAskPending()` which samples `ThreadEnvelopeSource.AskPending()` — `cmd/evener/serve.go`'s `liveThreadEnvelopeSource.AskPending()` forwards straight to `session.HasPendingAsk()`.

No test drove this end to end (a real ask_user round through the real `BridgeEvent`/envelope-sampling pipeline, checking the recorded frame), so the finding couldn't be settled by reading code alone. I built one: it drives a real `agent.Session` through a real ask_user round, bridged into a real `*Server` exactly the way `cmd/evener/serve.go` wires it, and asserts the recorded `thread/status/changed(awaiting)` notification carries `AskPending=true`.

**Result: it passes on origin/main with no production change.** The live wire already carries the flag correctly; the review finding was a false positive from reading the projector in isolation without following the stamp two layers downstream. No projector change is needed.

This is test-only, additive coverage for a previously-untested invariant (the reviewers were right that nothing proved this). No production code changed.

## Gate results
- `go build ./server/...`: clean
- `$(go env GOROOT)/bin/gofmt -l server/askpending_live_notification_test.go`: clean
- `go vet ./server/...`, `go vet -tags evenerfuzz ./server/...`, `GOOS=windows go vet -tags evenerfuzz ./server/...`: clean
- `golangci-lint run ./server/...`: 0 issues
- `go test -count=1 ./server/...`: PASS (includes the new test)


obra:
## Round 2 disposition (f00879ee9)

**Medium 1 — data race on `oneShotAskUserAdapter.used`**: real. Forced the interleave with a temporary 2-party `sync.WaitGroup` barrier inside `Complete` (removed before commit) to make the namer's toolless call and the main ask_user call enter the check concurrently; `-race` caught it immediately:

```
WARNING: DATA RACE
Read at 0x00c000010058 by goroutine 55:
  primeradiant.com/evener/server.(*oneShotAskUserAdapter).Complete()
      server/askpending_live_notification_test.go:55
  ...agent.(*Session).launchInitialPromptNamer.func1()

Previous write at 0x00c000010058 by goroutine 24:
  primeradiant.com/evener/server.(*oneShotAskUserAdapter).Complete()
      server/askpending_live_notification_test.go:58
  ...agent.(*Session).callModel.func6()
```
(20 unforced `-count=20 -race` runs beforehand never hit it — the natural window is narrow — which is why the barrier was needed to prove it before trusting the finding.)

Fix: added `mu sync.Mutex` to `oneShotAskUserAdapter`, locked around the whole `used` check-and-set in `Complete`, matching `fakeAdapter` (agent/session_test.go:36-47) and `mutationProjectionAdapter` (server/appwire_rejoin_test.go:243-259).

**Medium 2 — `go sess.ConsumeEventsLossless(...)` races registration**: real per `ConsumeEventsLossless`'s own doc comment (agent/session_events.go): "It RETURNS ONCE THE REGISTRATION IS IN EFFECT ... Callers must NOT wrap it in `go`; that puts the window back." Fix: removed the `go`, call it synchronously — it starts its own internal drain goroutine.

**Post-fix measurement**: `go test -count=20 -race -run TestAskUserLiveStatusFrameCarriesAskPending ./server/` — 20/20 PASS, no race.

Gates (repo root module, server/):
- `$(go env GOROOT)/bin/gofmt -l server/askpending_live_notification_test.go` — clean
- `go vet ./server/...` — OK
- `go vet -tags evenerfuzz ./server/...` — OK
- `GOOS=windows go vet -tags evenerfuzz ./server/...` — OK
- `go test -count=1 ./server/...` — ok
- `go test -count=1 -tags evenerfuzz ./server/...` — ok
- `golangci-lint run ./server/...` — 0 issues

obra:
Round 2 @f00879e: CI green (15), RoboRev one Low only (the session is closed only on the success path; a later t.Fatalf leaves the consumer running while t.TempDir removes the state dir). Merging per the Lows rule; the follow-up moves the close into t.Cleanup.

Issue #1572, opened 2026-09-16
hub: a plugin write that succeeds but whose follow-up listPlugins fails is applied without a broadcast


obra:
`cmd/evener-hub/app_plugins.go:328-333`: when the plugin manager write succeeds and the subsequent `listPlugins` (`mgr.List`) fails, the handler returns an error, so the `err == nil` gate in the RPC handler skips the `evener/plugin/updated` broadcast. The write is applied and no client is told, so every client's plugin list is stale until the next unrelated refetch. Same class as #1543 (credential rollback paths mutate without broadcasting).

Not a regression from D11 (#1502): the deleted native post-mutation re-read hit the same failing `mgr.List`. But D11 makes the broadcast the only reconciliation path for both hosts (all nine write paths broadcast: app_rpc.go:1283-1336, app_plugin_autoupgrade.go:93,123), so this hole now matters on both. Fix: broadcast whenever the write applied, independent of the listing's outcome, and carry the listing error separately. Found by the D11 implementer while refuting a /simplify finding.

obra:
Resolved by #1733 (b468560ed): the plugin store post-commit hook broadcasts at lock release, before any follow-up listing read runs, so a write that succeeds but whose listPlugins fails is already announced. Measured by the #1602 classification lane on 2026-09-17; #1834 covers the non-plugin surfaces.

PR #1834, opened 2026-09-18
hub: broadcast every applied auth/instance/launch write even when a later step fails

### LINKED ISSUE #1543: hub: instance Create/Remove rollback-failure paths return an error after mutating the config without broadcasting evener/auth/updated
Found by RoboRev on #1532. The credential store core now relies on the hub's echo after a refused write (`notifyInstanceUpdated` broadcasts `evener/auth/updated` for every applied instance write, incl. the rename-persisted error path). Two hub paths break that rule: `cmd/evener-hub/app_instances.go` `Create` (~:879-880, 'restoring the previous config failed') and `Remove` (~:1292-1299, 'the rollback could not be written, so the removal stands in the config') return an error after the config was mutated and never call `notifyInstanceUpdated`, so no client re-reads until a reconnect or an unrelated notification. Fix at the hub: broadcast before returning the error on every path that leaves the config changed, so 'the hub echoes every applied write' is true without exception; add a handler test per path. Double-failure scenario, low frequency, but the invariant is what the clients are built on now.

obra:
## Summary
- Auth's `LoginComplete`/`Logout`/`ApiKeySet`/`ApiKeyClear`/`CredentialJsonSet`/`DevicePoll`, every `evener/instance/*` mutation (`Create`/`Edit`/`Remove`/`SetDefault`/`SetModelDisabled`/`RefreshModels`), and launch's `SetLayer`/`TrustRepo` all decided whether to broadcast on a plain `err == nil`. A write that reached the config or the OAuth record and then hit a failing rollback, a failing reload, or a failing status/resolution re-read skipped the broadcast entirely - every other client kept a stale list until a reconnect or an unrelated notification (closes #1543).
- `writeApplied`/`writeDidApply` (`hubcore.ErrWriteApplied`) give every handler one answer to "is there a change the other clients need to hear about": `nil` is an ordinary applied write, and `hubcore.ErrWriteApplied` marks one that landed before a later step failed. Each handler wraps its own already-landed-then-failed error in `writeApplied` and decides its broadcast with `writeDidApply(err)`.
- `app_instances.go`'s local `renamePersistedError` type (the identical distinction, encoded per-caller) is retired in favor of the shared sentinel; `Edit`'s RPC handler and every other instance write now share one `instanceWrite` closure in `app_rpc.go`.
- `app_instances.go`'s `Remove` keeps its `InvalidParams` classification for a name that resolves to an environment-only instance unchanged - that refusal never reaches a write, so it's out of scope for this fix (I found the reference implementation I ported from had also demoted this to a generic error alongside an unrelated test change; I reverted that specific hunk since it isn't part of the write-applied mechanism and isn't explained by it).

## #1572 status
Also effectively closed, but by #1733 (already on main), not by this PR: #1572 was specifically about a plugin write whose follow-up `listPlugins` fails skipping the broadcast. Since #1733 moved the plugin broadcast to a post-commit hook fired when `lockStore` releases - before the RPC handler ever calls `mgr.List` for its response - a failing listing can no longer suppress a plugin/marketplace broadcast. This PR does not touch plugins at all.

## Non-test line count vs. the ~300 target
Total non-test diff is **312 lines** (`app_auth.go` 55, `app_instances.go` 84, `app_launch.go` 8, `app_rpc.go` 115, `app_write_applied.go` 37, `hubcore/write_applied.go` 13) - about 4% over the ~300 guideline. I did not split it by surface (auth/instance/launch) because all three share the same ~50-line `writeApplied`/`writeDidApply`/`ErrWriteApplied` core; splitting would force a sequential dependency (core + one surface lands first, the other two rebase onto it) for a ~12-line overage. Happy to split if you'd rather have it that way - flagging per the brief's rule rather than deciding unilaterally.

## Test plan
- [x] `app_write_applied_test.go` (new): RPC-level `TestHubRPCInstanceCreateBroadcastsWhenTheRollbackCouldNotBeWritten` / `...RemoveBroadcastsWhenTheRollbackCouldNotBeWritten` (#1543's exact double-failure cases), plus controller-level cases for `Remove`'s double credential failures, `SetDefault`'s failed reload, `SetLayer`'s failed resolution, and `LoginComplete`'s failed status read - each asserting `writeDidApply(err)` or the RPC broadcast directly.
- [x] Falsification: reverted `app_auth.go`/`app_instances.go`/`app_launch.go`/`app_rpc.go` and removed `app_write_applied.go`/`hubcore/write_applied.go`; every new test failed to compile (`undefined: writeDidApply`); restored and reran green.
- [x] `$(go env GOROOT)/bin/gofmt -l cmd/evener-hub/` - clean
- [x] `golangci-lint run ./cmd/evener-hub/...` - 0 issues
- [x] `go test -count=1 ./cmd/evener-hub/...` and `-tags evenerfuzz` - both green

Closes #1543.


obra:
## Simplify + review disposition (d6c0c29)

Applied the /simplify review (`1834-simplify.md`) as one commit per item:

1. `4542ce51d` - keybindings' `KeybindingsPostRenameError.Unwrap` now also reports `hubcore.ErrWriteApplied`, so a generic `errors.Is` check recognizes it (pure cleanup).
2. `0b6deda68` - `writeApplied` now wraps in an `appliedError` type (KeybindingsPostRenameError's shape) instead of `fmt.Errorf("%w: %w", ...)`, which was prefixing "the write landed before this failed: " onto every client-visible message whose inner error is plain. Behavior-visible fix, TDD'd.
3. `3ab8cf97b` - `restoreFailedRemoval` now decides its own `writeApplied` wrap internally (returns one `error`, not `(bool, error)`); the three call sites that repeated `if restoreFailed { writeApplied(err) }` are gone. Behavior-preserving (existing Remove test suite is the regression guard).
4. `513236ab1` - `RefreshModels` never writes providers.toml (`fetchInstanceLive` only republishes into the in-memory holder), so `writeDidApply(err)` there was decoration; it now returns only `error` and routes through the shared `instanceWrite` closure.
5. `6dbd6d87f` - `statusAfterCredentialWrite`/`openAIStatusAfterWrite` (six identical lines apart from the reader) collapsed into one `statusAfterWrite(name, read)`, with `statusByProvider` as the file-layer-credential adapter.
6. `634a51069` - **the real bug**: on a failed status read after a landed credential/OAuth write, the handler broadcast `AuthStatusResponse{Provider: name}` (zero-value `ActiveSource`), so `notifyAuthUpdated` announced "nothing active" as fact when the truth was never read - other clients could adopt it. Fixed with `notifyAuthWrite`: a clean read broadcasts the real data, an applied-but-failed read broadcasts `notifyInstanceUpdated`'s no-data form (refetch, no fabricated field), a refused write broadcasts nothing. New test `TestAuthLoginCompleteFailedStatusReadBroadcastsNoData` asserts the broadcast's raw JSON carries neither a `provider` nor an `activeSource` key; watched it fail with the real bug (`{"activeSource":"","provider":"anthropic"}`) before the fix.
7. `1dedbb515` - `appwire-client/typescript/state/credentials/instances.ts`'s self-echo comments said a failed RPC proves no broadcast follows; #1834 makes that false for an applied-then-failed write. Doc fix only: the no-data broadcast still can't be attributed as this client's own echo (no provider/originClientId on it), so the existing unconditional-retire behavior stays correct: not a logic change.
9. `d6c0c29ca` - the five `(#1543)`/`(#1543's rule)` comment citations are gone (one already went with item 5's rewrite); `hubcore.ErrWriteApplied`'s doc is now the one place the rule is stated in full, and it now matches item 2's actual mechanism instead of the pre-cleanup `fmt.Errorf` wrap it used to describe.

Item 8 (a `didWrite` marker set at the write primitives instead of wrapping errors at each call site) is filed as #1837, not done here. Item 10 (six of eight new #1834 tests assert `writeDidApply(err)` at the controller level rather than the broadcast end-to-end) is a note; no action.

RoboRev: checked via REST before pushing - the commit status for the prior head (`3a1c1ff2c`) shows `roborev: pending` with zero PR/issue comments and zero reviews, so there was nothing to fold in or refute this round.

Full head SHA: d6c0c29cab11b74c5c215bd07b2b0d7b5c337b96

obra:
Round 1 @d6c0c29: CI green (15), RoboRev three Lows only (stale RefreshModels doc comment; statusAfterWrite comment describes a broadcast that no longer happens; no test for TrustRepo's applied-write wrap). Merging per the Lows rule; the follow-up PR with all three opens immediately. Closes #1543.

PR #1836, opened 2026-09-18
fix(gate): pass the caller's build flags to the package enumeration (#1268)

### LINKED ISSUE #1268: Gate enumeration ignores build-tag-only packages (e.g. //go:build race) that go test -race would compile
Surfaced on #1145 round 26 and left unaddressed by design in its replacement (#1265/#1266): the gate enumerates packages with a plain `go list ./...`, so a package that exists only behind a build tag `go test -race`/`-msan` sets (e.g. `//go:build race`) would be missing from the list handed to `go test`. No package in the tree is like that today (the two race-tagged files sit in packages that exist anyway), so this is a latent divergence, not a live bug. Fix when it matters: pass the same build-affecting flags to the enumeration as to the test invocation, in the Go helper, with a test that plants a race-only package in a fixture module.

obra:
Fixes #1268

## Root cause

The gate enumerated packages with a plain `go list ./...` and handed the result to `go test`. `go list` and `go test` do not see the same tree: `-tags` selects files, and `-race`, `-msan` and `-asan` each set a build tag of their own. A package whose only files sit behind `//go:build race` therefore exists for `make test-race` but not for the enumeration that decides what that run tests, so the package was silently never tested. Latent today (no such package in the tree), but a real divergence in the gate.

## Change

- New `evener-dev dev list-build-flags` subcommand (`cmd/evener-dev/listbuildflags.go`): prints, one per line, which of the caller's `go test` flags must also be given to the `go list` that enumerates the packages. It consumes each value with its flag first, so `-run -race` stays the regex it is, and it reuses the shard runner's value-flag tables for that.
- `scripts/gate/run-module-tests.sh` derives those flags once and passes them to both `go list` sites: the root module's package list and the agent module's subpackage enumeration.

## Test evidence

- `go test ./cmd/evener-dev/... -count=1` → `ok` (both packages)
- `go vet ./cmd/evener-dev/...` clean; `gofmt -l` clean; `bash -n scripts/gate/run-module-tests.sh` clean
- `TestPackageSelectionFlagsDecideWhatGoListCanSee` plants a tag-only package in a fixture module and shows it missing from plain `go list` and present once the forwarded flags are supplied. `TestPackageSelectionFlagsForwardsWhatChangesTheTree` covers the value-vs-flag cases (`-run -race`, `-ldflags -race`, `-args`).
- Independent check outside the test: in a fixture module with `//go:build race`, `go list ./...` → `racetagged/always`; `go list -race ./...` → `racetagged/always` + `racetagged/raceonly`.


obra:
Round-nine review (`fa011d2`) resolutions, all in head `2a25cea`:

1. **Unbounded `wait` after bounded termination** — `process_descendants`, `stop_process_tree` and `run_bounded` moved to `scripts/lib/gate-bounded.sh`, and the timeout path no longer waits on a stopped process at all; `stop_process_tree` TERMs targets, waits a bounded 5s grace, then KILLs survivors. A survivor in an uninterruptible wait can no longer hold the gate past its bound.
2. **Non-atomic status publication** — the child writes `${log}.status.tmp` and renames it, so the status file is absent or complete, never partial; the empty-to-1 fallback was removed.
3. **GOFLAGS relative paths** — `run_list_build_flags` now runs from the module's own directory (evener-dev named by full import path, resolving through the workspace from `agent/`), so `-modfile`/`-overlay` resolve exactly as they do for that module's `go test`.

Evidence: new `gatebounded_test.go` drives the real sourced functions — success, failure replay, a timeout returning promptly with no status file, and a TERM-ignoring child KILLed after the grace with an explicit handler-readiness signal (deterministic 5/5). Commands: `go test ./cmd/evener-dev/...` ok; `go test . -run 'TestRunBounded|TestStopProcessTree'` ok; `make lint-golangci` PASS; `MODULES='. envvars' WEB=0 scripts/gate/run-module-tests.sh -short -count=1` PASS (345.92s + 0.25s).

obra:
Round-ten review (`2a25cea`) resolutions, in head `bb88dda`:

1. **Helper resolution** — `run_list_build_flags` invokes the checkout-local `./cmd/evener-dev/bin` from `repo_root` again (an import path can be redirected by `GOWORK=off` or an unrelated workspace), and clears `GOFLAGS` for that one invocation so a relative GOFLAGS path cannot resolve against the repo root while the module tests resolve it from their own directory.
2. **Descendant escape** — `stop_process_tree` rescans descendants on every pass and signals the parent last, instead of a one-shot snapshot. New `TestStopProcessTreeCatchesALateFork` fails against the old snapshot (child `escaped=yes`) and passes now.

Evidence: `go test . -run 'TestRunBounded|TestStopProcessTree'` ok; `go test ./cmd/evener-dev/...` ok; `make lint-golangci` PASS; gate PASS . 356.29s + PASS envvars.

obra:
Round-fifteen review (`14443e5`) resolutions, in head `102368f0c` — the cleanup class addressed as one change rather than patches:

- **High, timeout converted back to success** — a status file absent at the deadline is now final; `run_bounded` no longer rechecks after cleanup. New `TestRunBoundedRejectsAStatusPublishedAfterTheDeadline` fails before the fix (rc=0) and passes now (rc=1).
- **Repeated SIGTERM / final-KILL fork race** — `run_bounded` launches the command under bash job control so it leads its own process group, and `stop_command` signals the group (TERM, bounded grace, KILL). A descendant forked or reparented during cleanup is inside the group, so no PID snapshot has to be atomic. The PID walk remains only as the fallback when the platform will not give the job its own group, and it now TERMs each PID once.
- **GOFLAGS=-C** — the gate reads the effective `go env GOFLAGS` and refuses `-C` (covers a go env file), before dispatching any module.
- **Timing ratchet** — the grace is now `EVENER_STOP_TREE_GRACE_SECONDS` (default 5); the tests set 1s, which is the code\u2019s fault, not the threshold\u2019s. `testing-budget.json` is unchanged. The four lifecycle tests are 0.7\u20131.7s each.
- **Shell splitting** — the caller's argv is preserved as an array and expanded quoted for every `go test` and `agent-shards` call; `-args`/bare `--` are rejected since the gate appends its own flags.

Verification: bounded tests ok; `go test ./cmd/evener-dev/...` ok; `make lint-golangci` PASS; gate PASS . 374.03s + PASS envvars.

PR #1839, opened 2026-09-18
test(native): pin a failed turn with queued work active (#1156)

### LINKED ISSUE #1156: mobile client: a failed turn with queued work now reports State active on the input-end emission
## Where
Mobile client state reduction (`mobile/src/state/conversation.ts` and the connection layer), consuming the daemon behaviour changed in PR #1098 (fix commit 85e4d83b1).

## What
On the turn-failure input-end emission the daemon now reports `State: s.WireState()` instead of a hard-coded idle. A failed turn with queued work therefore reports `active`, matching what `WireState()` reports everywhere else. The mobile client must not treat an input-end after a failure as "session is idle now".

## Why it matters
A client that assumes idle after failure will offer input controls while the daemon is about to start the queued turn, and will misrender the queued message's state.

## Direction
Check the mobile reducer's handling of the failure exit against the new emission; add a fixture with a queued message behind a failed turn and assert the rendered state is active.

---

obra:
Fixes #1156

## Status after rebase

The behaviour #1156 reports is already fixed on `main`: `9fa51ef6a` — "fix(appwire): let the turn_failed status frame settle a failed turn (#1432) (#1782)" — removed both failed-turn self-settles in `appwire-client/typescript/reducer.ts`, so the `turn_failed` status frame owns the failed turn's state and a message still queued behind a failure keeps the session active.

This PR now carries only the part of #1156 that `main` does not: the mobile-layer fixture the issue asked for. A failed `turn/completed` with a queued message asserts `conversation.status.type === "active"` with the queue depth intact. (`main`'s failed-turn tests cover the no-queue case; the queued case was uncovered.)

## Diff

`mobile/src/state/conversation.test.ts` — 1 file, +31 lines.

## Test evidence

- `npx vitest run --root .. --config mobile-native/vitest.config.mts mobile/src appwire-client/typescript/reducer.test.ts appwire-client/typescript/reducer.edge.test.ts appwire-client/typescript/reducer.notes.test.ts` — 995 passed (10 files)
- `make test-web` — PASS web-typecheck, PASS web-test, PASS web-lint (exit 0)


obra:
RoboRev ordering finding — handled, with one correction to the record.

The finding's line citation describes the opposite of the code at `869c2a6`: that test applied **`turn/started` first, then `thread/queueChanged`** (`mobile/src/state/conversation.test.ts:1059-1066` in that revision). Reproduce with `git show 869c2a6:mobile/src/state/conversation.test.ts | sed -n '1058,1066p'`.

That said, the conclusion — the order was not the daemon's — was right for the reverse reason. The daemon drains the queue **before** it opens the queued turn: the drain's `popQueueHead` claims the entry and `reflectDurableInputQueue` emits `EventQueueChanged` (`agent/session_queue.go:741` → `agent/session_client_mutation_queue.go:539-558`), and only the next loop iteration's `EventUserInput` opens the turn (`agent/session_lifecycle.go:1419`). So the test was reversed relative to the daemon, and the suggested fix ("apply `thread/queueChanged` after `turn/started`") would have kept it reversed.

Fixed at the root in `6292b9a`: the test now emits `thread/queueChanged{depth:0}` first, then `turn/started`, matching the daemon, and asserts each boundary instead of only the final state — after the drain: status `active`, `queue.depth === 0`, `activeTurnId` still undefined; after `turn/started`: status `active`, `activeTurnId === "t2"`, `queue.depth === 0`.

Tests: `npx vitest run --root .. --config mobile-native/vitest.config.mts mobile/src appwire-client/typescript/reducer.test.ts appwire-client/typescript/reducer.edge.test.ts appwire-client/typescript/reducer.notes.test.ts` — 996 passed (10 files). Lint: `cd cmd/evener-hub/frontend && npm run lint` — exit 0.

Issue #1363, opened 2026-09-15
hub: a parked queue's release requests (drain, promote) are accepted while idle but the steer capability that gates them is derived from the active status


obra:
Found on #1340 round 9 while giving the web queue strip an affordance for the idle + queued (parked by Stop) state.

**Daemon contract (measured).** A Stop parks the input queue (`agent/session_client_mutation.go` `QueueHeld = true` at acceptance); while held `pendingQueueDepth()` is 0 (`agent/session_queue.go`), so `WireState` reports idle with a non-empty queue. The documented release triggers are every user-initiated run: `turn/start` (the new message runs first, then the drain loop takes the parked messages), `turn/queue`, `turn/drainAsSteer`, `turn/promoteQueuedAsSteer`; cancel-queued does not release. `clientMutationDrain` and `clientMutationPromote` have no status precondition (interrupt fence, queue revision, empty queue, reserved entries only) and wake a steering carrier, so either runs the parked work while the session is idle.

**Hub advertisement.** `server/appwire_runtime.go` `appCapabilitiesLocked`: `Steer: steerAvailable && active && !closed`. At idle the wire says `steer: false`, and there is no capability for drain or promote, so a client that only sends what the daemon advertises (the rule #1340 settled on for every steering affordance, `canSteer`) cannot offer "run the parked queue now" without a new message. The wire also carries no held flag; idle + `queue.depth > 0` is the parked state by construction.

**What #1340 does.** The strip shows "Paused by Stop. Your next message runs first, then the queue." beside the composer's Send, which is the `turn/start` half of the contract, and keeps Steer queue now / Steer now off at idle.

**Decision for Jesse.** Should a parked queue be runnable without a new message? Options: (a) the hub advertises `steer` (or a distinct `drain` capability) whenever the queue is held, since the daemon accepts those requests then; (b) the wire carries the held flag on `QueueState` so clients can name the state exactly; (c) leave it: the next message releases it, which is what the Stop promised ("cancel execution and wait", kata wms7).

PR #1840, opened 2026-09-18
fix(hub): interrupt and queue advertise harness support, not a turn in flight (#1375)

### LINKED ISSUE #1375: appwire capabilities: interrupt and queue still fold the active status in, while steer now means harness support alone
Noticed by the /simplify altitude pass on PR #1340. After #1340, `ThreadCapabilities.steer` is "the harness supports steering" (server/appwire_runtime.go ~2770) and the client applies the session status itself (`canSteer`, `canDrainQueue` in appwire-client/typescript/submitRouting.ts). `interrupt` (~2790) and `queue` (~2800) still fold `active` in on the hub side, so one struct carries two semantics and every client has to know which flags are pre-gated (the web's `showStop = busy && interrupt` double-gates already).

Wanted: uniform semantics, all action flags mean "harness supports it", and the SDK's session-controls derivation applies the status for all of them. Wire-semantics change: needs the TUI case and updates to the hub tests that pin the fold, plus the mobile-native reader. Out of scope for #1340, which only unfolded steer.

obra:
Fixes #1375

## Root cause

#1363 unfolded the active status from `ThreadCapabilities.steer` (it became `steerAvailable && !closed`), but `appCapabilitiesLocked` still folded `active` into `interrupt` (`interruptWired && active && !closed`) and `queue` (`queueFunc != nil && active && !closed`). One struct then carried two semantics: `steer` meant "the harness supports it", while `interrupt`/`queue` meant "a turn is running *and* the harness supports it". Every client had to know which flags arrive pre-gated — the web's `showStop` is already `busy && interrupt`.

## Change

`server/appwire_runtime.go` `appCapabilitiesLocked` now advertises:

- `Interrupt: s.interruptWired && !closed`
- `Queue: s.queueFunc != nil && !closed`

exactly the rule `steer` follows: all three action flags mean "harness support alone"; only `closed` withholds them. The client applies the status. The SDK's `sessionControls` (`appwire-client/typescript/submitRouting.ts`) already applied the status to every flag (`stop: active && interrupt`, `queue: active && queue`), so **no client behaviour changes** — the web, TUI, native and mobile derivations keep gating stop/queue/steer on an active status.

Comment/doc updates to match: `submitRouting.ts`, `sendQueueAvailability.ts`, `cmd/evener-tui/composer_panel.go`, `mobile-native/src/composerCommand.ts`, `mobile-native/src/conversationControls.ts`, `hubapi/types.go`.

## Test evidence

RED first: `TestAppCapabilities_InterruptAndQueueAdvertiseHarnessSupport` failed against the old code (`idle, wired` → `Interrupt = false, want true`).

Tests that pinned the fold were updated to the harness-support rule (not weakened — the closed/unwired cases still assert withholding):

- `server/appwire_runtime_test.go`: new harness-support table; the one-decision guard now asserts steer *and* interrupt track `closed` alone.
- `server/appwire_server_test.go`: `TestServerAppWireQueueCapabilityFlipsWithProcessing` → `TestServerAppWireQueueCapabilityAdvertisesHarnessSupport` (renamed, and the fuzz-surface reference in `server/server_surface_fuzz_test.go` follows it).
- `server/appwire_capabilities_push_test.go`: the idle push frame now expects `queue`/`interrupt` true.
- `cmd/evener-tui/hub_model_test.go`: the idle-capabilities case carries the hub's new idle set (`send/steer/interrupt/queue` on) and proves the TUI keeps them while its controls withhold stop/queue, ctrl+c is a silent no-op, and no force-steer hint appears.
- `cmd/evener/serve_state_test.go`: post-stream-error idle read expects `queue`/`interrupt` advertised.

Commands (all exit 0):

- `go test ./server/ -count=1`
- `go test ./cmd/evener-tui/ -count=1`
- `go test ./cmd/evener/ -count=1`
- `go test ./cmd/evener-hub/ -count=1` (targeted `Capabilit|ControlInvariant|Queue|Interrupt|Steer` subset first, then the full-package run in the combined command below)
- `go test ./cmd/evener/... ./cmd/evener-hub/... ./hubapi/... -count=1`
- `gofmt -l` (clean) and `go vet ./server/... ./cmd/evener-tui/... ./cmd/evener-hub/... ./hubapi/...` (clean), `go build ./...`

TypeScript changes are comment-only; the SDK derivation logic is unchanged. Local vitest was not run (no `node_modules` in this lane, and `appwire-client/typescript` has no install in the main checkout, per the brief's fallback).


obra:
**High finding (`ProtocolVersion` not bumped): it holds, but I am escalating it for a protocol-owner ruling rather than fixing it inside this PR.**

To be precise about what this is: not a refutation. The flags' meanings did change under an unchanged `evener-appwire-v5`, and both remedies the review names are repo-wide changes, not localized ones. I did not bump the version, and here is the evidence for why that call is yours.

**Reproduction of the silent reinterpretation.** The two peers are indistinguishable to a client. A pre-change daemon's idle snapshot and a post-change queue-less daemon's idle snapshot produce byte-identical inputs to `deriveSendQueueAvailability` — `statusType: "idle"`, `capabilities.queue: false`, `hasPendingSend: true` — and the current code answers `BOTH_UNAVAILABLE` for both:

```
npx vitest run ../../../appwire-client/typescript/sendQueueAvailability.test.ts -t "no queue seam"
```

On the pre-change daemon that answer is wrong: it has a queue seam and would have accepted the message. No client-side change can separate the two cases, because the inputs are identical.

**Why not bump here.** `appwire/flagday_contract_test.go` pins `evener-appwire-v5`, and the handshake compares versions exactly, so a bump is a flag day: every running daemon and client that has not been rebuilt fails the handshake. The string appears 153 times across 43 files, including `cmd/evener-hub/internal/sshconn` version negotiation and its deploy tests (remote daemon upgrades), `rendezvous` identity tests, the mobile client, generated `types.gen.ts`, and docs. That is a coordinated release action with a deployment plan, not something to land inside a capability-semantics fix.

**Why not the additive option either.** A new field / dual path is what `docs/superpowers/plans/2026-07-29-appwire-authoritative-rejoin.md` explicitly forbids ("Do not add a compatibility reader, dual path, feature flag"). The repo's additive precedent for a wire change — `ProviderDescriptor.Vars` kept alongside `VarsEnv`, "because v3 peers … `ProtocolVersion` is compared exactly, so a wire shape cannot change under v3" — works only when the old field can stay authoritative. Here the old field's *meaning* is what changed, so there is nothing additive that distinguishes an old daemon's `queue: false` from a queue-less new daemon's.

**Impact, for the ruling.** Both mixed-version symptoms are narrow. A new client's pending-send path can briefly disable the second message until the status frame flips the session active — at which point old and new peers agree, since both advertise `queue: true` at active for a wired daemon. An old TUI's `/interrupt` palette row can be offered at idle against a new daemon and be refused with "no active turn". Neither loses a message.

If you want the bump, say so and I will do it as its own change (constant, flag-day contract and its fixtures, sshconn/rendezvous/mobile fixtures, regenerated `types.gen.ts` and docs) so it can be reviewed as a protocol change rather than smuggled into this one.

obra:
**On the Medium protocol-compatibility finding (`sendQueueAvailability.ts` tier-6 veto): the same issue I escalated earlier — pointing at that analysis rather than re-arguing it.**

This is the protocol-version question again, now stated from the client side: the tier-6 veto treats a live `queue: false` as "no queue seam", which is only true for daemons that publish the unfolded bit, and a pre-#1375 daemon still folds `active` into `queue`. My earlier escalation covers the diagnosis and both remedies the review names:

https://github.com/prime-radiant-inc/evener/pull/1840#issuecomment-5734038811

For this head, one extra data point for whoever rules on it: the review is right that **queue's client rule is the only one that diverges**. `steer` and `interrupt` fold to the same answer under both spellings at every status a client reads them in (both are consulted only while `active`, where old and new daemons agree). So the incompatibility is exactly the tier-6 veto, and there is a third option besides the two in the finding:

1. Bump `ProtocolVersion` (flag day; 153 references across 43 files, incl. sshconn version negotiation and `appwire/flagday_contract_test.go`). Needs a release plan.
2. Negotiate a feature flag / add a marker field — the additive route, which is a dual path, and `docs/superpowers/plans/2026-07-29-appwire-authoritative-rejoin.md` says not to add one.
3. **Drop the tier-6 veto**, leaving the SDK exactly as it was before this series for that branch (a pending send at a non-active status routes to `turn/queue`). That removes the only divergent client rule and needs no protocol change, at the cost of the avoidable `turn/queue` → `Unavailable` in the race window that the first review round asked to remove.

I have not made any of these three changes unilaterally: (1) is a coordinated release action, (2) is against the repo's stated plan, and (3) reverses an earlier review finding. The rest of this head's findings are fixed; see the next commit.

PR #1841, opened 2026-09-18
refactor(sdk): share settings generation lifecycle

obra:
Keybindings and transcript-display settings need the same ready-generation lifecycle. This change extracts notification wiring, payload retirement, and lost-hub write settlement into shared settings helpers, with keybindings using them immediately. Payload-specific revision and conflict state remains owned by each store.

Follows merged #2024. The three P8 commits are restacked on main 7b23fd083416bb7944b652ed73116b4949da8a80; their changed and context lines match the original reviewed patch exactly. The generation-preserving recovery behavior is unchanged. Own scope is 151 production additions / 54 deletions and 218 test additions across five files.

Validation: 113 focused tests, package TypeScript and qualification, frontend typecheck, scoped Biome, import lint, and diff checks passed. Independent correctness/simplification review and local RoboRev2659 passed at e8521d4db057821d3e51c83ab0c6c304b4140e8e. Fresh current-head CI and the complete raw remote panel are required before merge.


obra:
## Round 1 disposition (311721c) — answering the raw panel reviews

Answering the raw panel members' findings directly (reviews/raw/1841-93477dbf8.md).

- **Medium** (member 1, echoed member 2): this piece introduces `settingsHubGeneration.ts`, imported by `keybindingsStore.ts`, but `tsconfig.build.json`'s `files` list omitted it — shipped (compiled into `dist/`) but unlisted, so the package's coverage floor and reachability gates (`package-coverage.mjs`, `qualify-package.mjs`) never enumerated it. **Fixed**: added `settingsHubGeneration.ts` to `files`, and exported `SettingsHubGeneration`/`createSettingsHubGeneration` from `index.ts` (mirroring `readyGenerationFence.ts`'s existing pattern) so the qualifier's reachability assertion is satisfied. Commit 311721ca2.
  - Falsified: `make test-api-package` passed silently BEFORE this fix despite the omission, confirming the gate gap the review describes. Adding the file to `files` WITHOUT the `index.ts` export made the qualifier fail loudly: `AssertionError: shipped module unreachable from every published specifier: settingsHubGeneration`. Adding the export made it pass again.

- Pointer: `discardStoredDraft`'s `removeIf`-refusal → `"absent"` mapping — fixed in **#1791**, commit 282b22ecb; inherited by this branch through the restack.

- **Low** (member 0, not owned by this piece): `keybindingsStore.test.ts` uses a real 200ms `setTimeout` watchdog in one test (the "a stale direct write's late reply cannot clear a newer direct write's flag" test, introduced in piece #1808/p3to5), making that assertion's timing real-clock-dependent. This test predates this piece — filing as a residual for a small follow-up (fake timers or a deterministic deferred-request race) rather than fixing it here, since it isn't this piece's own code.

### Gates
- biome/lint: clean
- vitest `keybindingsStore.test.ts` + `settingsHubGeneration.test.ts`: 100/100

Head: `311721ca275f7447ef770fdbaea0ce063c84b7a5`

obra:
## Round 2 disposition (4c4b19ea) — answering the raw panel review

**settleUnsettleableWrite exported/tested with no production caller; module header's "byte-identical... lost-hub write's own settle" claim was false** (member 2, pi default). Confirmed: `keybindingsStore.ts`'s own `settleLostHubWrite` duplicated the shared helper's logic inline, adding `draftConflict` (a field `SettingsHubWriteFields` doesn't carry). Fixed per the review's first suggested option: `settleUnsettleableWrite` now takes an `extra: Partial<Fields> = {}` parameter (the same shape `retireSettingsHubPayload` already has), merged into the settle's `setState`. `keybindingsStore.ts`'s `settleLostHubWrite` now routes through it with `{ draftConflict: true }` instead of re-deriving the fence check and setState call. The header's claim is now true rather than aspirational. Falsified: two new tests in `settingsHubGeneration.test.ts` (`extra lands in the same publish`, `extra is not published when superseded`) fail without the `extra` parameter; reverting the keybindingsStore.ts routing change alone (keeping the parameter) doesn't break any existing test (the routing is a pure duplication removal, not a behavior change) — confirmed via a full revert of both files together, which does fail the two new tests.

Everything else on this panel (offline-probe-must-not-throw, stored-null/string-"null" conflation, matchesStoredBytes canonicalization) is #1791's own subject — see its round 6 disposition for the fix and falsification.

Gates: vitest (appwire-client/typescript) 2005/2005; vitest (mobile-native) 833/833; typecheck clean both sides; `npx biome ci` clean; `npm run lint` clean; `make lint-package-imports` PASS; `make test-api-package` PASS (settingsHubGeneration.ts's own index.ts/tsconfig.build.json entries, pre-existing on this branch, untouched by this round).

obra:
## Restack (13c8e83)

Rebased onto p7's restacked tip (1d75712). No conflicts; own-diff content verified byte-identical to the pre-restack diff (`git diff 62fc5a70b..4c4b19ea0` vs `git diff <p7-old-tip>..HEAD` on the +/- content lines). No follow-up commit needed.

Gates: mobile-native vitest (840, one isolated flake on a re-run confirmed passing) and `npm run check` clean; appwire-client/typescript + web vitest (2011) clean; `npm run typecheck`/biome/lint clean; `make lint-package-imports` PASS; `make test-api-package` qualified.

obra:
## Round 2 disposition + restack on p7 (c2afb383)

Restacked on p7's new tip (#1792 @ 41fc2f2e); own 3 commits replayed clean, no conflicts. This PR's round-2 panel (reviewing the full stacked diff) only echoed findings already owned and fixed upstream at their true point of introduction - see #1900's and #1902's own dispositions for the createId-outside-try, marker-collision, and extra-unguarded-load fixes. No new finding is specific to this PR's own 3 commits, so no code change here.

Head: `c2afb38362d562370e81de5f559d162ae5ef6d4e`

Own diff vs p7's new tip: 5 files, 369 insertions / 54 deletions (unchanged content, just the new base).

Gates (targeted, all green): full appwire-client/typescript vitest suite (2030 tests); `npm run typecheck`; `npx biome ci ../../../appwire-client/typescript`; `npm run lint`; mobile-native's full vitest suite (846 tests) and `npm run check`; `make lint-package-imports`; `make test-api-package`.

Pushed with `--force-with-lease=claude/sdk-d6-p8-settings-hub-generation:13c8e8353`.

obra:
Exact-head review for `e8521d4db057821d3e51c83ab0c6c304b4140e8e`: Luna and DeepSeek found no issues. Muse identified one Low: the endReadyGeneration test claims full ordering but records only payload retirement. The implementation has no remaining must-fix finding. A focused follow-up is assigned to strengthen the behavioral ordering assertion. Per Jesse's merge policy, this Low will be fixed in a separate PR after this parent lands; it is not being folded into the qualified parent. Required tests have passed; snapshot build is still running.

PR #1843, opened 2026-09-18
fix(web): the skill guard confirms select-all before typing the queue's second pass (#1690)

### LINKED ISSUE #1690: flake: web-skillguard queue journey expects no separator where the editor inserts one (caret 65 vs 64)
## What happened

One of five full `TestSkillComposerBrowser` runs failed at the **queue** journey's second-pass edit (`cmd/evener-hub/frontend/scripts/skillguard/run.mjs:1868-1869`):

```
skillguard FAIL: timed out after 15000ms (waited 15043ms) waiting for native edit
  "PROSE_QUEUE_14c first pass /pkg:probePROSE_QUEUE_14c second pass" with caret 64;
  native state {
    "before":{"value":"PROSE_QUEUE_14c first pass /pkg:probe","start":37,"end":37,"focused":true},
    "after": {"value":"PROSE_QUEUE_14c first pass /pkg:probe PROSE_QUEUE_14c second pass","start":65,"end":65,"focused":true}
  }
```

The gesture is `driver.selectAll(sessionA)` followed by `driver.typeText(sessionA, PROSE.queue2)`. `typeText` derives its expectation from the caret state it read first — the current value with the typed text appended, no separator — and then waits for that state. What the editor produced was one character longer: a **separating space** between the chip label `/pkg:probe` and the typed token text, which moves the caret from 64 to 65.

That is the #1648 separator rule doing its job: the insertion point sits immediately after a chip label and the typed text begins with a token character, so without the space the label would swallow the text.

The other four runs of the same branch passed this milestone, so the state the gesture lands in is not stable: either the caret position after `selectAll` + `typeText` varies, or the first-pass draft's trailing whitespace varies. I did not investigate which.

## Context

- Branch: `followup/skillguard-synthetic-paste`. Its only change is the inline-editing clipboard milestone in `run.mjs`; the queue journey and the separator rule are untouched, so this looks pre-existing rather than caused by that diff.
- Not the same failure as #1434. That one is a 15s timeout on the *steered-turn* wait, i.e. work arriving too slowly under load. This one is a **state mismatch inside `typeText`**: the edit landed, one character different from what the test expected.
- Artifacts from the failing run were kept at `skillguard-browser-1908551304/artifacts` (driver.log, appwire-trace.jsonl, helper logs, screenshots) in that run's scratch on the host that reproduced it; the verbatim state dump is quoted above so the ticket does not depend on it surviving.

## Not done

No investigation and no fix. Recorded because `web-skillguard` is part of the canonical browser gate, and a one-in-five flake at a milestone that asserts exact draft text is worth a look before it is mistaken for a product regression.

obra:
Fixes #1690

## Root cause

The queue journey's second pass is `selectAll` then `typeText(PROSE.queue2)`: it selects the draft a queued entry returned to the composer and types over it. `selectAll` placed the whole-text DOM range and returned without confirming the editor had adopted it. The composer is a real ProseMirror editor and adopts a DOM selection asynchronously, so a render landing in between left the caret at the end of the draft (the queued-edit caret restore parks it there). `typeText` then read that caret, computed "append at the end", and the #1648 separator rule inserted a space between the chip label `/pkg:probe` and the typed run -- one character more than the oracle, which cannot see the separator. The guard reported the separator mismatch (`caret 65 vs 64`) instead of the lost selection.

The oracle is deliberately **not** widened to expect the separator. The appended result is not what the scenario meant, and `clickSubmit` immediately after asserts the composer holds exactly `draft(PROSE.queue2)`, so accepting the separator would only move the failure one step later and hide an unreplaced draft.

## Change

`selectAll` (`cmd/evener-hub/frontend/scripts/skillguard/run.mjs`) now confirms the editor holds the whole-text selection before returning:

- it re-reads the editor state and waits for it to settle (`settleComposer`, the guard's existing two-reads-80ms-apart helper, so a selection a render is about to reset does not read as held);
- if the settled selection does not cover the whole text, it re-applies the range, announces the re-application, and tries again, bounded at 4 attempts;
- an editor that never holds the selection fails loudly instead of the next `typeText` typing into a caret the scenario never meant.

Only the settled pre-edit selection is re-established -- the shape #1317 used for the same gesture; a text-level loss still fails. This is the driver confirming its own precondition, not repairing the typed run.

## Tests

- `npx vitest run scripts/skillguard/run.test.mjs` -- 7 passed (4 existing + 3 new cases pinning confirm / re-apply / fail against a scripted state sequence; no browser needed).
- `go test -tags browserguard ./cmd/evener-hub -run '^TestSkillComposerBrowser$' -count=1` -- PASS (the full guard, twice).

## Reproducibility note

The flake did not reproduce locally. Across 9 full guard runs (6 with `SKILLGUARD_CPU_THROTTLE=6`, 3 concurrent) every run passed, and instrumented reads taken 0/40/120/280/600 ms after each `selectAll` showed the whole-text selection held every time. Since the editor does not reset the selection on its own, the loss in the failing run required a concurrent render -- which is exactly what the unconfirmed selection let through silently. The fix is argued from the failure dump and those reads, not from a local reproduction.


obra:
Addressed roborev's Medium in 5f936ab6: the selectAll confirmation now compares the settled selection against the *settled* text (`settled.value.length`), not the length read before the range was placed. A draft that grew mid-flight no longer reads as a whole-text hold with only the old prefix selected.

Pinned by a new case in `scripts/skillguard/run.test.mjs`: it places the range for a short draft, settles on a longer one mapped to the old prefix, and requires a re-apply at the fresh length. Measured red against the previous predicate (1 failed / 7 passed) and green now (`npx vitest run scripts/skillguard/run.test.mjs` -> 8 passed).

PR #1844, opened 2026-09-18
D6 piece 9: checkpointed draft editor primitives (discard/persist)

obra:
Extract shared checkpointed draft persistence and discard primitives from the keybindings store, preserving storage failure reporting through draftError and conflict restoration behavior. Export and package the helper module, and strip only the internal generation stamp when projecting native keybinding drafts while retaining structured errors.

Qualification at 98cb71188f6bb30656b01e1d055e64d8db063e2c: AppWire package, full web gate, native gate (946 native and 793 shared tests, types/scripts and bundle), Biome, package-import lint, and diff check passed. Exact-base RoboRev2680 found no issues. The restack preserves the intended semantic patch over merged settings generation d1d056a24f616e2d1bda1cc9581b776416332f0d.

This is the shared editor prerequisite for transcript checkpointed drafts; it does not claim the later web adapter adoption is complete.


obra:
## Round 1 disposition (dd3bd63) — answering the raw panel reviews

Answering the raw panel members' findings directly (reviews/raw/1844-a5bf96836.md).

- **Medium** (echoed by every member): this piece introduces `checkpointedDraftEditor.ts`, imported by `keybindingsStore.ts`, but `tsconfig.build.json` omitted it. **Fixed**: added it to `files`, and exported `assertDraftDiscardable`/`persistCheckpointedDraft`/`discardCheckpointedDraft` plus their `*Fields` types from `index.ts`. Commit dd3bd6399. Same falsification pattern as #1841 (`make test-api-package` silent before, loud "shipped module unreachable" failure with the file listed but not exported, passing again once exported).

- **Medium** (member 1, codex): `keybindingsDomain` (`mobile-native/src/nativePreferences.ts`) projects `state.draft` directly into `PreferenceState<ConfirmedKeybindings>.draft`. Since #1792 gave the store's `KeybindingsDraft` its own `generation` staleness stamp, this structurally leaks that store-internal field into a projected type that declares no such field (TypeScript's excess-property checks don't catch it because `state.draft` is an expression, not an object literal). **Fixed**: the projection now strips `generation`, keeping only `{version, revision, rules}`. Commit 0599861fa.
  - Test: "projects the keybindings draft without the store's internal generation stamp" (`nativePreferences.test.ts`) — asserts `draft` matches the confirmed shape and specifically `not.toHaveProperty("generation")`.
  - Falsified: reverted the projection change — the `not.toHaveProperty` assertion failed (received `generation: 1`). Restored, test passes.

- Pointer: `matches()` accepting only exact-or-canonical JSON — fixed in **#1791**'s `matchesStoredBytes` extraction, commit 0dfbc24f2; inherited through the restack.

### Gates
- biome/lint: clean
- vitest `checkpointedDraftEditor.test.ts` + `keybindingsStore.test.ts`: 100/100 · `nativePreferences.test.ts`: 26/26

Head: `dd3bd6399741d7613a36ed442faa0763ce29fe85`

obra:
## Round 2 disposition (e60b47fb) — answering the raw panel reviews

**checkpointedDraftEditor's discard failure left draftError unchanged** (Low, member 3, panel dd3bd6399 on #1845 — this PR's own file, so fixed here): `discardCheckpointedDraft`'s catch on `discardClassified()` throwing set `storageUnavailable: true` but never `draftError`, the same class of gap `persistCheckpointedDraft`'s own catch already closed on this branch. Fixed: the catch now also sets `draftError: draftDiscardFailedMessage`. Falsified: added `draftError: "discard failed"` to the existing `marks storageUnavailable and throws when discardClassified itself throws` test; reverting the one-line fix fails it (`draftError: null` received); restored, green.

**Generation-stamp leak into the native projection** (`nativePreferences.ts:137`, reported fixed in an earlier pass on this PR): confirmed already fixed and already tested — `keybindingsDomain` projects `{ version, revision, rules }` explicitly rather than spreading `state.draft`, and `nativePreferences.test.ts` has `projects the keybindings draft without the store's internal generation stamp` asserting `expect(draft).not.toHaveProperty("generation")`. No further action; re-ran it green as part of this round's full suite.

**persistCheckpointedDraft's own tests passed a raw DraftPort where a DraftRepository was expected** (found while wiring the fold's #1808 round-4 CAS fix through this piece's extraction): the two `persistCheckpointedDraft` tests called `memoryDraftStorage().storage` directly, bypassing `createDraftRepository`'s classification — this only type-checked because `DraftRepository.save`'s return type was still `void`; widening it to `boolean` on the fold exposed the mismatch (`tsc` error, not a silent pass). Fixed the tests to wrap the port the way `keybindingsStore.ts` actually does, and added the CAS-refusal case (another writer replaced the classified record) this repository-level call was never exercising. `persistCheckpointedDraft` itself also needed the same one-line CAS check the fold added inline (`if (!drafts.save(checkpoint)) throw ...`), since this piece's extraction predates that fix and would otherwise silently regress it — done in the same conflict resolution, confirmed by reverting and watching the new store-level test fail.

Refreshed against current main (`e60b47fb`, merge commit): the only conflict was `appwire-client/typescript/index.ts` (both sides added export lines near the same spot — checkpointedDraftEditor vs. an unrelated chunkview export); union-resolved per the standing refresh procedure. `mobile-native` gained a new dev dependency (`react-test-renderer`) on main since this branch was cut; ran `npm install` in that worktree (node_modules there is not the shared symlink) to pick it up. `mergeable_state` was `dirty` before this merge, `true`/`blocked-on-CI` after.

Gates: vitest (appwire-client/typescript) 2100/2100; vitest (mobile-native) 840/840 (up from 833 — main's own new component-render tests); `npm run check` (mobile-native) clean; `npm run typecheck` (frontend) clean; `npx biome ci` clean; `npm run lint` clean; `make lint-package-imports` PASS; `make test-api-package` PASS.

Pointer: the biome-line-width Low (member 3, `keybindingsStore.test.ts:239`, several test titles over 120 chars) does not reproduce — measured directly: `git show dd3bd6399:appwire-client/typescript/keybindingsStore.test.ts | awk 'length>120'` finds 17 such lines (a test-title string literal Biome's formatter cannot wrap), but `npx biome ci ../../../appwire-client/typescript` from `cmd/evener-hub/frontend` reports 0 errors on this exact file both before and after this round — Biome's formatter does not flag a line it cannot shorten (an unbreakable string token) as a format violation. No change made; flagging as refuted-by-measurement rather than silently accepted.

obra:
## Rebuilt on p8 (13c8e83)

Restacked as a fresh branch from p8's tip (`13c8e8353130e2618cb21ed5038ce4c017ab337a`), cherry-picking p9's own six commits (`235ed7d09 6c78461b0 56c1313c7 5f45cb0bb 4d9f90791 6e7bcad81`) in order, dropping the merge commit and everything it absorbed from main per the lane-8 handoff note.

**Conflict on `235ed7d09`** (extracts `persistCheckpointedDraft`/`discardCheckpointedDraft` into `checkpointedDraftEditor.ts`): p8's tip already carries `5c6ba3f58` ("persistDraft distinguishes a CAS refusal from a storage throw"), which the fold landed after this commit was authored. `235ed7d09`'s extracted `persistCheckpointedDraft` treated a CAS refusal the same as a genuine storage exception, which would have regressed `5c6ba3f58`'s fix. Coordinator ruling (option A): extended `persistCheckpointedDraft` to take a `restoreDraft` callback and a review-again message, mirroring `discardCheckpointedDraft`'s own parameter shape in the same commit, carrying `5c6ba3f58`'s behaviour into the shared primitive with no behaviour change - `keybindingsStore.ts`'s `persistDraft` now delegates to it. The fold's regression test (`editDraft's persistDraft refusal adopts the replacement instead of reporting storageUnavailable`) stayed green unchanged, and a matching test was added at the primitive level in `checkpointedDraftEditor.test.ts`. Falsified: reverting `persistCheckpointedDraft`'s body to the pre-resolution shape fails that test, the "two stores share nothing" test, and the new primitive-level test.

A later own commit, `5f45cb0bb` ("persistCheckpointedDraft's tests exercise the real repository, not a raw port"), became empty after this resolution - it fixed the same raw-port-vs-repository test bug the resolution already fixed, and its own new CAS-refusal test asserted the pre-`5c6ba3f58` behaviour (contradicted by the resolution), fully superseded by the test already added. Skipped with `git cherry-pick --skip`.

**Own diff verified byte-identical** to the old branch's own diff (`4c4b19ea0..6e7bcad81`, the pre-merge base to the last own commit) modulo exactly the sanctioned resolution: the `persistCheckpointedDraft` signature/body extension, its doc comment, the corresponding test-call-site widening, and the third test's rewrite (assertions changed from `storageUnavailable: true / draftError: "save failed"` to `throws "review again", draft adopted, storageUnavailable: false` - the dropped assertions describe the exact behaviour `5c6ba3f58`'s own oracle test already superseded). All other hunks differ only in line-number/context shift from p8's restack.

New head: `d5d40bb99bbfcdc276628f529353c9af72c46f0f`

Gates (targeted, all green): vitest on `checkpointedDraftEditor.test.ts` + `keybindingsStore.test.ts` (108 tests) and `mobile-native/src/nativePreferences.test.ts` (26 tests); `npm run typecheck`; `npx biome ci ../../../appwire-client/typescript`; `npm run lint`; `npm run check` in mobile-native; `make lint-package-imports`; `make test-api-package` (index.ts/tsconfig.build.json touched by the `checkpointedDraftEditor.ts` export commit) plus its index.ts falsification (deleting the re-export fails the qualifier with "shipped module unreachable... checkpointedDraftEditor", restored clean).

Pushed with `--force-with-lease=claude/sdk-d6-p9-checkpointed-draft-editor:e60b47fbb15a73397d564cc93615474921a20b14`.

obra:
## Restack on p8 (a50ac1a9)

Restacked onto p8's new tip (#1841 @ c2afb383, old base 13c8e8353). Own 5 commits replayed, one real conflict on `4cf9f3c2a` (the `persistCheckpointedDraft` extraction): fold's round-6 fix (createId() inside its own try, see #1808) landed in `keybindingsStore.ts`'s inline `persistDraft` body upstream, and this commit's own delegation to the newly-extracted `persistCheckpointedDraft` doesn't yet have that fix - it would have regressed it. Resolved by taking the delegation (this commit's whole point) and applying the SAME createId-inside-try fix to `persistCheckpointedDraft` itself in `checkpointedDraftEditor.ts`, which had inherited the bug unnoticed during extraction. Pinned by a new test (`createId()` throws → `storageUnavailable`+`draftError`, mirroring the fold-level test), falsified by moving `createId()` back outside the try.

Head: `a50ac1a955481ff0f7d97011919f88b7f8cfafff`

Own diff vs p8's new tip: 8 files, 438 insertions / 56 deletions (the delta over the previous restack is exactly the createId-inside-try fix to `persistCheckpointedDraft` plus its test).

Gates (targeted, all green): full appwire-client/typescript vitest suite (2030 tests); `npm run typecheck`; `npx biome ci ../../../appwire-client/typescript`; `npm run lint`; mobile-native's full vitest suite (846 tests) and `npm run check`; `make lint-package-imports`; `make test-api-package`.

Pushed with `--force-with-lease=claude/sdk-d6-p9-checkpointed-draft-editor:d5d40bb99`.

Issue #1759, opened 2026-09-18
mobile receipt decoders reject every additive wire field, so shipped phone builds break on optional keys


obra:
Class finding from RoboRev on #1735 (High). mobile/src/services/conversation.ts decodes every mutation receipt with `exactObject(result.receipt, requiredReceiptKeys, ...)`, which throws on any key not in the allowlist. That makes every additive, optional wire field a break for already-shipped phone builds: TestFlight build 4 (source 92561dbe3) calls turn/drainAsSteer from QueueSheet.tsx, and once #1735 lands the daemon's drain receipt carries `consumedClientMutationIds`, which build 4's allowlist does not know, so the phone throws after the server has already drained.

Jesse's ruling (2026-09-17, 'drop the version bump') is that additive AppWire changes stay on the current ProtocolVersion, on the premise that an old peer reads a new optional key as no information. The web client honours that premise; the phone's strict receipt decoder does not.

Decide and implement one of:
1. Phone receipt decoders ignore unknown keys (validate the keys they know, drop the rest), matching the web and the ruling's premise. My recommendation: it is the only option that protects builds already in testers' hands going forward.
2. Keep strict decoding and treat every additive receipt field as a phone-first change (ship the widened allowlist in a TestFlight build BEFORE merging the daemon side).

Immediate consequence: TestFlight build 5 must carry #1735's widened allowlist (mobile/src/services/conversation.ts) before build 4 users run against a hub built from main after #1735.

obra:
Ruling (Jesse, 2026-09-18): phone receipt decoders ignore unknown keys — validate the keys the phone knows and drop the rest, matching the web and the no-flag-day premise. Implementation goes in one bounded PR from main (mobile/src/services/conversation.ts exactObject call sites for receipts, with tests that an unknown key is dropped and a known key is still validated). TestFlight build 5 must still carry #1735's widened allowlist for build 4 users until it ships.

PR #1845, opened 2026-09-18
D6 piece 10: transcriptDisplayStore hub defaults + direct write

obra:
## D6 piece 10: transcriptDisplayStore hub defaults + direct write

Stacked on #1844 (piece 9), base main per the queue's serial-merge convention.

Oracle: PR #1549 @dfde92160's transcriptDisplayStore.ts / transcriptDisplayStore.test.ts, for the read (hub defaults) and write (direct write) path specifically - the checkpointed draft editor half is out of scope (D6 piece 12).

### Size: over the ~400 non-test-line guideline, with a specific reason

`git diff --stat` for this PR's own commit: readyGenerationFence.ts +19, transcriptDisplayStore.ts +517 (345 code lines, 127 comment lines, 45 blank - counted with an awk pass, not just eyeballed). This is a genuinely new store, not an extraction like pieces 8/9, so it doesn't shrink the way those did. I kept the oracle's comment density (matching this codebase's established style of explaining WHY, not just WHAT) rather than trim it to hit the number. If you'd rather I cut this into two PRs after the fact (read path / write path), say so - I didn't split it because the plan's own item 10 names them as one unit ("hub defaults + the direct write") and splitting them now would require stubbing one half against the other with no real boundary to justify the seam.

### What's adopted directly from pieces 8/9 - zero re-derivation

- `createSettingsHubGeneration` (piece 8) for `beginReadyGeneration`/`endReadyGeneration` and the notification wiring.
- `retireSettingsHubPayload` (piece 8) for the in-flight-flags reset in `retirePayload`; the per-layout `patchTokens` reset (transcript-specific bookkeeping the shared fence knows nothing about) is layered on top by claiming a fresh `fence.claimWrite()` per layout before delegating, rather than trying to read `fence.writeToken` after the shared primitive's own internal `supersede()` - order-independent and correct either way a caller sequences it.
- `settleUnsettleableWrite` (piece 8) is **not used** here: I initially imported it, then found it (via `whyFenced`) is only ever called from the oracle's `saveDraft` (piece 12), never from `patchHubDefault` or `refreshFor`. Removed the unused import before pushing.
- Piece 9's `assertDraftDiscardable`/`persistCheckpointedDraft`/`discardCheckpointedDraft` are **not used** here either - discard/persist are checkpointed-editor-only concerns, entirely out of piece 10's scope.

### A real gap found in the shared fence, fixed here (additive, keybindings untouched)

The oracle's `readyGenerationFence.ts` (shipped inside #1549) has two members ours doesn't: `awaitingFirstPayload` / `firstPayloadApplied()`. keybindings' `retirePayload` resets `revision: 0` on every generation end, so nothing is ever "previous" going into a new generation and a hub restart's lower revision is never wrongly compared. Transcript's `hub` is deliberately **retained** across a transient disconnect (the effective configuration must not flap - `hub`'s own field doc says so), so its stale-guard needs an explicit "this generation has confirmed nothing yet, so the next payload is authoritative at any revision" escape hatch keybindings never needed. Added both members to the shared `ReadyGenerationFence` interface and implementation, TDD'd in `readyGenerationFence.test.ts` (2 new tests, watched fail with "awaitingFirstPayload is not a function" / `undefined`, then implemented). keybindingsStore.ts never reads or sets either member - confirmed by the full keybindingsStore.test.ts suite (90 tests) passing unchanged.

### Oracle hunks dropped as dead (piece 12's job, not lost - listed for whoever builds it)

- `restoreDraft`, `draftCheckpoint`, `staleDraft`, `invalidDraft`, the `discardStoredDraft` re-export - all draft-port-shaped, no port exists yet.
- `initialState()`'s `draft`/`storageUnavailable`/`draftUnreadable`/`draftConflict`/`draftError` fields and the `TranscriptDisplayStoreDeps.drafts` dependency.
- `applyHubDefault`'s draft-generation-stamping block (the `const draft = state.draft !== null && ... ? {...} : state.draft` dance) and its `draftConflict: staleDraft(draft, hub, fence.generation)` field - piece 10's `applyHubDefault` only publishes `hub`/preview-clearing/`extra`.
- `refreshFor`'s `settledWrite(...)` spread into the mobile layout's `extra` - nothing to settle without a checkpoint.
- `patchHubDefault`'s gate clause `|| state.saving || state.writeUncertain` (the checkpointed editor's sibling gate) - the two fields are declared (piece 8's primitive requires them) but nothing sets them yet, so checking them now would be permanently-dead code; piece 12 adds the clause when it adds `saveDraft`.
- `editDraft`/`saveDraft`/`discardDraft`/`rebaseDraft`/`settledWrite`/`whyFenced`/`claimLayoutWrite`-adjacent draft bookkeeping - all piece 12.
- Test-side: the oracle's `"the checkpointed draft editor"` describe block (600 lines) and the checkpoint-coupled direct-write tests within `"the direct write"` (a later checkpointed write superseding an earlier direct write's reply; a direct write refused while a checkpointed write owns the layer; an older save's late reply; rejections/supersession clearing `saving`) - all call `saveDraft`/`rebaseDraft`, which don't exist yet. The `"a retired payload fences..."` matrix test is ported with its `saveDraft` path and the `editDraft` throw-assertion dropped (2 paths x 4 reset sites instead of 3x4).
- `transcriptDisplayStore.ts` is **not yet exported from index.ts** and **not yet in tsconfig.build.json's `files` list** - nothing in the package imports it, so `tsc -p tsconfig.build.json` never compiles it (verified: ran it directly, zero errors, confirming it's simply not part of the build graph yet, not silently broken). Whichever piece wires the first real consumer (11 or 12) should add both.

### Gates (all green)
- `npx vitest run appwire-client/typescript` (full suite): 97 files / 2002 tests
- `npx vitest run` in mobile-native (full suite): 83 files / 797 tests
- `npm run typecheck` in cmd/evener-hub/frontend: clean
- `npm run check` (tsc) in mobile-native: clean
- `npx biome ci ../../../appwire-client/typescript` and `npm run lint` from cmd/evener-hub/frontend: clean
- `make test-api-package`: qualified (doesn't touch this file - see above)

### Falsification
readyGenerationFence: the two new tests failed against the pre-fix fence (`expected undefined to be true`, `firstPayloadApplied is not a function`); pass after the fence extension.
transcriptDisplayStore.test.ts: failed on module resolution against the not-yet-existing file (0 tests run); all 39 adapted oracle tests passed on the first run once implemented - no debugging round needed, which I read as the pre-write measurement (reading the oracle's actual patchHubDefault/applyHubDefault/refreshFor bodies before writing anything) having paid off rather than as nothing having been exercised: the 39 tests cover the stale-guard, the per-layout preview/contradiction rules, the conflict/malformed/fenced-reply paths and the awaitingFirstPayload cases directly.


obra:
## Round 1 disposition (c8511f0) — answering the raw panel reviews

Answering the raw panel members' findings directly (reviews/raw/1845-27e804987.md).

- **Medium** (member 0, codex): this piece introduces `transcriptDisplayStore.ts` but it was neither in `tsconfig.build.json`'s `files` list nor exported from `index.ts` — no consumer could import the new SDK store at all. **Fixed**: added it to `files`, and exported its public surface (`TranscriptDisplayStore`, `TranscriptDisplayStoreState`, `TranscriptDisplayStoreDeps`, `TranscriptDisplayClient`, `TranscriptDisplaySupport`, `HubDefaultsByLayout` types; `createTranscriptDisplayStore`, `fromWireChange`, `transcriptDisplaySupport` functions). This makes the D6 plan's originally-scheduled "piece 11 adds the `index.ts` export as the first real consumer" step moot — the export already exists as of this round. Commit 5fe047755.

- **Medium** (member 1, codex): `fromWirePatchResponse` rejected any PATCH reply that didn't have EXACTLY `{layout, revision, config}` (a 3-key check), unlike `fromWireChange`/`fromWireDefaults` which already tolerate extra fields — a forward-compatible hub addition would break every direct write. **Fixed**: dropped the exact-key-count check, keeping only the presence/validity checks on `layout`/`revision`/`config`. Commit b7925e206.
  - Test: "a PATCH reply with extra keys still decodes (forward compatibility)".
  - Falsified: reverted the production diff — the reply was rejected as malformed even though it decoded correctly otherwise. Restored, test passes.

- **Medium** (member 1, codex): a malformed PATCH reply (`InvalidPatchResponseError`) left the optimistic `drafts[layout]` preview stranded, while every OTHER failure path (transport failure, a lost-revision conflict) cleared it via `clearPreview`. **Fixed**: the catch block in `patchHubDefault` now always calls `clearPreview` — it's idempotent, so this is safe for the one `InvalidPatchResponseError` case (the revision-regression check) that had already cleared it explicitly before throwing. Commit b7925e206 (same commit as the key-count fix above).
  - Test: extended "a malformed success..." to assert `drafts.mobile` is `undefined` afterward.
  - Falsified together with the key-count fix: reverting the production diff made both new assertions fail (extra-key reply rejected; stale preview left in place). Restored, both pass.

- **Low** (member 1, codex): `draftBackend.testkit.ts`'s `fakeDraftBackend.get()` returned the live Map-stored reference instead of a clone, unlike `set()` (already `structuredClone`s), the real Storage-backed backend (re-parses JSON on every read), and `memoryDraftStorage` (`structuredClone`s on load) — a caller mutating what it read could corrupt the store without going through `set`/`replaceIf`. **Fixed**: `get()` now clones. Commit c8511f043.
  - Test: "get() returns a clone, not the live stored reference" (`nativePreferenceDrafts.test.ts`).
  - Falsified: reverted the one-line change — mutating the loaded value corrupted the stored record. Restored, test passes.

### Gates
- biome/lint: clean
- vitest `transcriptDisplayStore.test.ts`: 40/40 · full `appwire-client/typescript` suite: 99 files / 2047 tests · mobile-native full suite: 83 files / 820 tests

Head: `c8511f043eecf1c5a389e8ffbdd8aad62b50eb14`

obra:
## Round 2 disposition (2814cfe5) — answering the raw panel reviews

Nothing own beyond #1844's items — pointers only:

- **Offline corruption probe not wrapped in error handling** (member 0/3) — fixed in #1791 (`readDraftOutcome`), inherited through this stack.
- **`settleUnsettleableWrite` no production caller / header claim false** (Low, member 2) — fixed in #1841, inherited.
- **`checkpointedDraftEditor.ts:97` discard failure leaves draftError unchanged** (Low, member 3) — fixed in #1844, inherited.

Own cleanup: rebasing this branch onto #1791's own `fakeDraftBackend.get()` clone fix (the same Low this branch had ALSO carried, at its own tip — misplaced there rather than at #1791, its actual owner) left one byte-identical duplicate `describe("fakeDraftBackend", ...)` block behind in `nativePreferenceDrafts.test.ts`. Removed; same coverage, one copy (confirmed: the full mobile-native suite is 833 before and after, and this branch's own diff against #1844's new tip is now zero-net on that file).

Refreshed against current main the same way as #1844 (index.ts union-resolve; `mobile-native` npm install for `react-test-renderer`) — rebuilt on top of #1844's new tip by cherry-picking this branch's own commits rather than merging main a second time, so #1844's merge commit stays the sole ancestor carrying main's drift and `git merge-base --is-ancestor` holds end to end.

Gates: vitest (appwire-client/typescript) 2100/2100; vitest (mobile-native) 840/840; `npm run check` (mobile-native) clean; `npm run typecheck` (frontend) clean; `npx biome ci` clean; `npm run lint` clean; `make lint-package-imports` PASS; `make test-api-package` PASS.

obra:
## Rebuilt on p9 + round 3 (f7bfc4d)

Restacked as a fresh branch from the new p9 tip (`d5d40bb99bbfcdc276628f529353c9af72c46f0f`, see #1844's rebuild), cherry-picking p10's own five commits (`030e18235 cc1199534 d3bef0fb2 b510084fc 2814cfe52`, oldest to newest, a clean linear chain - no merge commit on this branch).

**Conflict on `b510084fc`** ("fakeDraftBackend.get() clones instead of returning the live reference"): trivial two-independent-insertions-at-the-same-point conflict against `mobile-native/src/nativePreferenceDrafts.test.ts` (p6b/p6c's `draftUnreadableAfterDiscard`/`clearsErrorAfterOfflineDiscard` blocks land at the same point); resolved by keeping both.

**Conflict on `2814cfe52`** ("remove the duplicate fakeDraftBackend describe block"): this commit exists because the *old* p10 branch's own `b510084fc` duplicated a `fakeDraftBackend` describe block that a different main commit had *also* independently added, absorbed into the old p9 merge. On the restack, p8's ancestry (via p6a's `2971ee835`) already carries that exact fixture/test, so `b510084fc`'s pick produced the SAME duplicate a second time, and `2814cfe52`'s own conflict was resolved by removing it entirely - converging on exactly one copy, matching `2814cfe52`'s own stated intent ("same coverage, one copy"). Net effect of `b510084fc` + `2814cfe52` together on this restack is zero (`git diff` on that file between the old p9 tip and the new p10 tip is empty).

**Own diff verified matching** the old branch's own diff (`e60b47fbb..2814cfe52`) - identical file stat (6 files, 1184 insertions / 1 deletion), and a full diff-of-diffs shows only blob-hash and hunk-context-line differences from p9's restack shifting surrounding line numbers, no content differences.

New head after the restack (before round 3): `38bcd087e`.

### Round 3: the two p10-own findings from the round-2 panel (`1845-2814cfe52.md`, member 0)

**(a) `transcriptDisplayStore.ts:484-499` - revision-conflict clears the preview.** Per Jesse's A6 ruling, checked what the web already decides: `cmd/evener-hub/frontend/src/stores/transcriptDisplay.ts`'s `patchHubDefault` catch clears `drafts[layout]` for every rejection except `InvalidPatchResponseError` - including the revision-conflict `WireError` (-32013), since that error is not `InvalidPatchResponseError`. Pinned by the web's own oracle test, `transcriptDisplay.test.ts:267` ("reverts a failed or conflicting draft to the narrow canonical conflict response"), which asserts `drafts.desktop` is `undefined` after a conflict. **Refuted** - the package already matches this exactly (apply canonical, then unconditionally clear the preview) and is already pinned by its own equivalent test, `transcriptDisplayStore.test.ts`'s "a lost revision race adopts the canonical current and reports the conflict on the layout" (asserts `drafts.mobile` is `undefined`). No code change.

**(b) `transcriptDisplayPostApply` errors not decoding `data.applied`.** `keybindingsStore.ts`'s direct write (`patchOverrides`) already draws this distinction for its own `keybindingsPostRename` rejection via a private `rejectionPayload` helper. Extracted that helper as `errors.ts`'s `wireRejectionPayload<T>` (generic over the payload's own decode function) so `transcriptDisplayStore.ts` reuses it rather than writing a second decoder; `keybindingsStore.ts`'s `rejectionPayload` now delegates to it with no behaviour change (96 existing tests unchanged and green). `transcriptDisplayStore.ts`'s `patchHubDefault` catch now checks for a `transcriptDisplayPostApply` rejection first: when present, adopts `data.applied` the same way the try block's own success path does (apply, clear the preview, clear the layout's error) and resolves rather than throwing. TDD: new test added first (confirmed failing - threw the WireError instead of resolving), fix applied, test green; falsified by reverting `transcriptDisplayStore.ts` alone (test fails the same way), restored clean. Also added a direct unit test for `wireRejectionPayload` in `errors.test.ts`, matching that file's existing per-helper convention.

**Extra check requested by the coordinator**: whether `transcriptDisplayStore.ts`'s own persist path has a third instance of the CAS-refusal-vs-throw bug class (`5c6ba3f58`/`persistCheckpointedDraft`). It does not, and cannot yet: piece 10 is direct-hub-write only, with no local draft/checkpoint storage port at all (`TranscriptDisplayStoreDeps` has only `client`; the file's own header comment says the checkpointed offline draft editor and its `drafts port dependency` are D6 piece 12, not yet built). There is no `drafts.save()`/CAS call in this file to wire the primitive into. Nothing to fix here yet.

Head: `f7bfc4dbe36f1adcd99c84adbcc34f7ca5142236`

Gates (targeted, all green): vitest on `transcriptDisplayStore.test.ts` (41), `keybindingsStore.test.ts` (96), `errors.test.ts` (31), `checkpointedDraftEditor.test.ts` and `readyGenerationFence.test.ts`, plus mobile-native's full vitest suite (841 tests) and `npm run check`; `npm run typecheck`; `npx biome ci ../../../appwire-client/typescript`; `npm run lint`; `make lint-package-imports`; `make test-api-package`.

Pushed with `--force-with-lease=claude/sdk-d6-p10-transcript-hub-defaults:2814cfe52b7dbd36ff73c4c62e4418b89698095c`.

obra:
## Restack on p9 (81a20db1)

Restacked onto p9's new tip (#1844 @ a50ac1a9, old base d5d40bb99). Own 6 commits replayed, one conflict on the round-3 fix commit (`f7bfc4dbe`, the `wireRejectionPayload` extraction): p9's own restack this round moved `keybindingsStore.ts`'s imports around (adding `discardStoredDraft`/`DiscardStoredDraftResult`/`DraftPort` from `draftCheckpointPort.ts` for its own discard-alias wrapper). Trivial union - kept both the expanded `draftCheckpointPort` import and the `errors.ts` import's switch from `WireError` to `wireRejectionPayload` (the latter's own removal of the now-unused `WireError` import is correct and unaffected by the former).

Head: `81a20db1c37bb894ed1de61be1e06fd03977d15a`

Own diff vs p9's new tip: 9 files, 1271 insertions / 12 deletions (unchanged content from the previous round, just the new base plus the trivial import-merge).

Gates (targeted, all green): full appwire-client/typescript vitest suite (2076 tests); `npm run typecheck`; `npx biome ci ../../../appwire-client/typescript`; `npm run lint`; mobile-native's full vitest suite (846 tests) and `npm run check`; `make lint-package-imports`; `make test-api-package`.

Pushed with `--force-with-lease=claude/sdk-d6-p10-transcript-hub-defaults:f7bfc4dbe`.

obra:
## Round 3 (own findings from 1845-f7bfc4dbe.md)

Two more p10-own findings, fixed on top of the restack posted just above:

**(1) Medium: decoder tolerance inconsistency.** `fromWireDefault`/`fromWireDefaults` (`transcriptDisplayConfig.ts`) required an exact key set, while `fromWirePatchResponse` (the PATCH reply decoder, same file's neighbor in `transcriptDisplayStore.ts`) already tolerates a hub-added field - so GET, a conflict's carried `current`, and a post-apply failure's carried `applied` all rejected a future hub field that a PATCH success would accept, despite comments on both sides claiming one shared posture. Per Jesse's ruling (#1759, decoders validate known keys and ignore the rest): dropped the `hasExactKeys` check on `fromWireDefault`'s `{revision, config}` and `fromWireDefaults`' `{desktop, mobile}` wrapper only - `fromWireConfig`'s own nested exact-key check, and the file's other unrelated `hasExactKeys` calls (toolIntent/contentLevel/contentVector shapes), are untouched. Pinned by three new tests (GET, conflict, post-apply, each with an extra key), all failing before the fix.

**(2) Low (ruling-3 behavior, fixed now per your note): a superseded PATCH reply threw "malformed" instead of resolving.** `patchHubDefault`'s revision-regression check treated a reply the hub had already moved past (a broadcast from another client's write racing ahead of this one's own reply while it was in flight) as `InvalidPatchResponseError`, even though `decodePatchReply` had already validated the reply as THIS write's own outcome. Per the A6 ruling (a superseded write resolves with the hub's current value, never an error): now resolves with the current value instead of throwing. The existing test for exactly this scenario ("a reply the hub has already moved past...") asserted the old throw; updated its title and assertions to the new resolve-with-current behavior, since it's the oracle for this scenario, not a passing test to leave alone. Falsified by reverting the production change alone (the updated test then throws "malformed" instead of resolving).

Also verified per your ask: `git grep -n 'createId()' -- 'appwire-client/typescript/*.ts'` on this tip shows every non-test call site (`checkpointedDraftEditor.ts:77`, `draftCheckpointPort.ts:181`, `keybindingsStore.ts:1090`) inside its own try block.

Head: `737b7cebd`

Gates (targeted, all green): full appwire-client/typescript vitest suite (2079 tests); `npm run typecheck`; `npx biome ci ../../../appwire-client/typescript`; `npm run lint`; mobile-native's full vitest suite (846 tests) and `npm run check`; `make lint-package-imports`; `make test-api-package`.

Pushed with `--force-with-lease=claude/sdk-d6-p10-transcript-hub-defaults:81a20db1c`.

obra:
# Retained-P10 audit: complete, closing

Per the retained-requirements accounting kept in this issue, every retained P10 replacement is now merged on `main` and verified. Ledger below; drift accounting included as requested.

## Retained requirements → merged evidence

| Retained requirement | Merged as | Verified by |
| --- | --- | --- |
| Wire decoder/error contracts | #2051 `cf85b5c62`, doc Low #2053 `fded33afe` | post-merge CI run 35486790423 (recorded at retention time) |
| P10b hub defaults/read lifecycle, support, notifications, missed-before-confirmation refresh, lower-revision restart | #2055 `1fb7b32a31`, #2056 `55b974c84f`, #2058 `093728c637` | package suite: "a notification landing before the generation's first read confirms is followed by one read", "a new generation's first read applies lower revisions from a restarted hub", "a relayed lower revision during the loaded publication cannot roll back the first read"; read publication reentrancy: "a lifecycle retirement during the loading publication prevents the request" |
| Read publication reentrancy correction + direct-generation pending-loading (review 2691) | #2056 `55b974c84f` | "a direct generation replacement fences a notification until its first read confirms", "a direct generation replacement retires a pending read without clearing cached defaults" |
| P10c direct PATCH, per-layout previews, canonical conflict/post-apply, malformed replies, current-value fenced returns, concurrent layouts, preview contradiction reconciliation | #2073 `a1a42fd926` + follow-up #2074 `86f19377fd` | package suite: "previews the draft, then commits the canonical response and clears it", "a malformed success changes no hub state, reports it, and clears the preview", "a fenced reply resolves with the current value instead of rejecting", "concurrent internal failures share their authoritative GET", "an atomic first read clears a contradicted stranded preview and keeps the matching one", plus the stranded-preview family (support flap, superseded write, reentrant loss) |
| Ready-generation first-payload fence (P10b additive), keybindings behavior preserved | #2055/#2099 chain | `readyGenerationFence.test.ts`: "awaitingFirstPayload starts true for a fresh generation and clears once firstPayloadApplied runs", "a new generation resets awaitingFirstPayload"; keybindings suite (64 tests) green in the package run, no fence members read |
| Export/build surface: root exports + build entry, no A1 duplication; installed-package behavioral qualification | #2087 `e8ff57dfdf` (A1) | `make test-api-package` qualified on current main content: "installed imports, declarations and read-only example" with the A1 probes (createSettingsHubGeneration, checkpointedDraftEditor, createTranscriptDisplayStore) |
| Browser adoption (A5/A6) | #2099 `b373001f14` + Low follow-up #2102 `e9a51bf5f6` | web delegates hub lifecycle/read/write to the SDK with the Zustand StoreApi preserved; the 29 oracle tests run unchanged through the adapter; CI web check green on the merge commits |
| A7 dead seams removed after ownership real | #2105 `dc984bbf70` | CHANNEL_NAME alias, useEffectiveTranscriptDisplay, effectiveForLayers exports removed; readyGenerationCallback.ts retained (shared + load-bearing token guard) |
| A8 checkpointed drafts: draft port, restore/stale/invalid handling, draft/storage/error/conflict fields, generation stamping, save/discard/rebase, settled writes, direct-write gating against checkpoint writes | #2112 `ed01792d3a` | package suite (83 store tests): "save refuses stale, unloaded, saving, uncertain, unsupported, and replaced-checkpoint state", "a lost reply preserves the uncertain checkpoint; an authoritative read settles it", "a draft-port failure after the hub loaded keeps direct writes gated until the uncertainty is settled", "a checkpointed takeover clears the superseded direct write's preview", "discard honors compare-and-remove for readable and unreadable records; replacements are adopted", "rebase requires reviewing the current revision and then allows save", generation stamping tests |
| A9 browser storage port | #2114 `9a5ce2c3c2` | web `browserDraftStorage` port (namespaced localStorage record, unforgeable unreadable identity, CAS compares); adapter forwarding + port suite; CI 15/15 green on the reviewed head |

All 17 chain SHAs verified ancestors of `origin/main` (`git merge-base --is-ancestor` per SHA).

## Drift ledger (extractions vs everything merged to the originals)

The SDK extractions began before #2055/#2056/#2058/#2073/#2074 landed in the web original, so drift was known to exist. The reconciliation protocol (diff extracted copies against current originals before cutover; reconcile post-extraction fixes into the extracted copy first) was applied at every step, and A5/A6 (#2099) completed the cutover: since then the web holds no second transcript-display state machine — `stores/transcriptDisplay.ts` delegates to the package, and the original web oracle suite runs unchanged against the package through the adapter. For the delegated surface there is now exactly one implementation, so the pre-cutover drift class is structurally closed.

- #2055/#2056/#2058 (read lifecycle, generation lifecycle, continuation): reconciled into the package extraction; the package suite carries their scenarios (see table).
- #2073/#2074 (PATCH integration + layout-error clear): landed in the store pre-cutover, carried wholesale into the package by the delegation; package PATCH suite covers them.
- #2079/#2081 (hub session clicks, tool-row rhythm): hub frontend components, no transcript-display store surface — verified no overlap during A1.
- #2104/#2113 (agent delegate runtime release, hub status row): Go-side only; `git diff` between the final web branch and main shows no `appwire-client/` or frontend files.
- Post-cutover surface decisions: only `createSettingsHubGeneration` root-exported from the settings module (A1 drift check); `LAYOUTS`/`isViewportClass` remain module-internal with zero consumers outside `transcriptDisplayStore.ts` (verified by repo-wide search) — unpublished by decision.

## Verification on final main content

- Package: `make test-api-package` qualified; full package suite green on main's package content (byte-identical to the branch; the Go-only #2104/#2113 delta does not touch it): 104 test files / 2354 tests passed.
- Web: `make test-web` green on every merge head in the chain; CI web check green on #2114's reviewed head; main CI on the final merge commit `9a5ce2c3c2` is terminal green: every applicable check success (19 check-runs), only the non-applicable Publish release skipped.
- Post-merge main CI verified green on every squash merge in the chain, including the final A9 merge `9a5ce2c3c2`.

## Follow-ups tracked outside this issue (not blockers)

- Docs-only: an inline deliberate-absence note for `LAYOUTS`/`isViewportClass` in the package barrel, matching the established pattern for other absences.
- Package-level (A9 review finding 3, dispositioned on #2114): a store-free standalone discard primitive for detached clients, as an A8-follow-up package export.

obra:
Retained-P10 replacements are all merged and verified on main; closing per the retained-requirements accounting. Full ledger and drift accounting in the preceding comment.

PR #1846, opened 2026-09-18
fix(agent): decide the poisoned-transcript refusal inside the wake's claims (#1165)

### LINKED ISSUE #1165: agent: the poisoned-transcript wake gate and the queue/steering claims read different locks, leaving a check-then-claim window
## Where
`agent/session_client_mutation_queue.go`, the poisoned-transcript wake gate in `ProcessPendingUserInput` (`wakeHasClaimableWork`, added on PR #1098) and the two claims it guards (`popQueueHead` in `agent/session_queue.go`, `claimSteeringCarrierTurn`). Found by the PR #1098 round-8 task review.

## What
The gate reads one snapshot under `clientMutationStore.stateMu` (RLock, released inside `snapshot()`); the claims later take a different mutex, `clientMutationStore.mu`, inside `mutate` and re-clone. No lock spans the two. Between them a concurrent `InterruptClientMutation` can set `QueueHeld`/`InterruptFence`, and a concurrent start/queue/promote can clear them. Two outcomes:
- gate says work → hold arrives → claim finds nothing: one spurious `ErrWriterPoisoned` with nothing claimed (noise);
- gate says no work → hold released → `popQueueHead` takes a durable claim on a poisoned transcript, the case the guard exists to prevent; the downstream poisoned-restore hands the message back, but a turn was announced that then dies.

The window is pre-existing: the previous gate read `Session.mu` twice (`QueueDepth`, `hasPendingUserSteering`) and popped under the store lock, so it was wider. PR #1098 narrowed it to one snapshot read; it did not close it.

Related, same code: the gate now deep-clones the whole client-mutation journal per wake (`snapshot()`), which the `queueHeld()`/`steeringHeld()` accessors exist to avoid for frequent samplers; small next to the `mutate` that `popQueueHead` runs on every wake, but no longer a pair of length reads.

## Why it matters
A session with a dead transcript can still announce a turn it cannot run, once, in a race with a Stop. Low frequency, but it is exactly the class the poisoned gate was added to close.

## Direction
Gate inside the claim's own `mutate` (the poisoned check as part of the claim decision, on the same clone the claim acts on) so the check and the claim see one state. That is a design change to the claim sites, not a line move; do it together with a cheaper wake-gate read.

---

obra:
Fixes #1165

## Root cause

`ProcessPendingUserInput` read one claimability snapshot (`wakeHasClaimableWork`, a `snapshot()` clone under `clientMutationStore.stateMu`), decided the poisoned-transcript refusal against it, and then claimed through `popQueueHead` / `claimSteeringCarrierInput`, which re-clone under a *different* mutex (`clientMutationStore.mu`) inside `mutate`. Nothing spanned the read and the claims, so a concurrent `InterruptClientMutation` (or a poisoning, or a queue/steering change) could land in between and make the two disagree:

- gate says work → hold arrives → claim takes nothing → one spurious `ErrWriterPoisoned` with nothing claimed (noise);
- gate says no work → hold released → `popQueueHead` takes a durable claim on a poisoned transcript, the case the guard exists to prevent: a turn is announced that then dies.

## Change

Each claim now decides the refusal on the same clone it acts on, and the wake reads the refusal from the claim instead of from a snapshot.

- `popQueueHeadRefusingPoison` (new) is the queue head's claim: after `queueHeadClaimable`, it returns `refuseBeforeClaimingOnPoisonedTranscript()` from inside the `mutate`, so the refusal is read on the generation the claim commits against and a refusal never commits anything. `popQueueHead` keeps its queue-mechanics signature and delegates to it, dropping the refusal (a refusal claims nothing and leaves the head queued; the wake and both drain-loop claim sites read the refusing form).
- `claimSteeringCarrierInput` returns its refusal the same way, so a poisoning or a Stop that lands after the caller decided the steer was runnable cannot hand it to a turn the transcript cannot record.
- `wakeHasClaimableWork` / `steeringCarrierClaimable` are removed: the wake no longer deep-clones the client-mutation journal per wake, and a wake that claims nothing stands down quietly instead of erroring.

Lock order is unchanged: the refusal reads the transcript's writer under `s.mu` while `mutate` holds `store.mu`, and no production path takes `s.mu` before `store.mu`.

## Test evidence

New tests in `agent/session_poisoned_wake_gate_test.go`, each confirmed failing on the pre-fix tree for the issue's own reason and passing after:

- `TestWakeRefusesTheSteeringCarrierWhenPoisonLandsBeforeTheClaim` — poisons inside the claim window via the existing `testOnly.steeringCarrierClaiming` seam. Pre-fix: `the refused wake announced turns [turn_m1]` (a turn announced that then dies). Post-fix: refused, no announcement, steering intact.
- `TestWakeStandsDownWhenAStopParksTheSteerBeforeTheClaim` — a Stop parks the rail in the same window. Pre-fix: spurious `ErrWriterPoisoned`. Post-fix: quiet stand-down.
- `TestPoisonedTranscriptRefusesTheQueueHeadClaim` — pre-fix `popQueueHead` claimed `"queued-behind-a-dead-transcript"` on a poisoned transcript; post-fix it refuses and the message stays queued.

Commands run on the rebased branch:

- `go build ./agent/...` — ok
- `go vet ./agent/` — ok
- `gofmt -l <changed files>` — clean
- `go test -count=1 -timeout 30m ./agent/` — `ok primeradiant.com/evener/agent 470.082s`
- `go test -race -count=1 -run 'TestPoisonedWriter|TestWake...|TestSteeringHeld|TestBufferedAcceptanceWake|...' ./agent/` — ok


obra:
Fixing the lock-order test flakiness from the `369415d` review: `TestClaimSamplesSessionMuOutsideTheStoreSerializer`'s select over `enteredSerializer` and `claimDone` now re-checks entry in the `claimDone` branch, so a random pick between two ready channels can no longer fail a correct implementation.

Honest note on reproduction: I could not reproduce the spurious failure. Before the fix, 500 default iterations and 300 iterations at `GOMAXPROCS=1` all passed on this host. The nondeterminism is real by construction — the claim goroutine closes both seams before it returns, so both channels can be ready when the test's select runs — but the window did not open here. The re-check removes the nondeterminism regardless, and 500 iterations after the fix are clean.

PR #1856, opened 2026-09-18
refactor(sdk): the mutation projection fence, package-only (D25e C3b)

obra:
Stacked on #1855 (D25e C3a). Merge C3a first.

## Mechanism
Adds `createMutationProjectionFence<A>()` to `appwire-client/typescript/state/mutation/projection.ts`: the epoch/generation all-targets-dominates-per-target fence over C3a's `MutationPersistencePort<A>`.
- `refresh(port, ref?)`: assigns a new generation, reads through the port, then decides which of the read's targets that generation may still replace — an all-targets refresh's generation counts against every per-target floor, and vice versa. Returns `false` when the fence's epoch moved since the read started (a client swap invalidated it) or the port's read rejected.
- `advance(ref)`: registers a live commit's next generation for one ref with no durable read at all.
- `epoch()` / `reset()`: for a longer-lived caller's own "is my mount still current" check, and for a client swap / test reset.

Published from `state/mutation/index.ts` alongside C3a's exports. **No web change in this PR** — `pendingTurnsStore.ts` keeps its own local fence and its own `replaceTargetRecords` (package-first, same pattern C1/C2 used); the next PR (C3c) switches the web onto both C3a's `replaceTargetRecords` and this fence.

## Tests
New tests in `projection.test.ts` (6 added, on top of C3a's 3 `replaceTargetRecords` tests): fresh single-target accept, all-targets accept naming every ref the snapshot carries, a port read rejection reported as `false`, `reset` invalidating a refresh already awaiting its read, `advance` out-ranking a slower in-flight all-targets refresh for one ref (mirrors the web's own oracle, `pendingTurnsStore.test.ts:250` "an older all-target projection cannot erase a newly committed send"), and `epoch`/`reset`.

## Falsification
Reverted `index.ts`/`projection.ts` to C3a's base (kept the fence tests): 6 of 9 tests failed with `TypeError: createMutationProjectionFence is not a function` (the 3 `replaceTargetRecords` tests from C3a still passed, as expected — they don't touch the fence). Restored; all 9 pass.

## Gates
- `npx vitest run state/mutation/projection.test.ts` (package): 9 passed
- `npm run lint` (frontend biome): clean
- `npx biome ci ../../../appwire-client/typescript`: clean

## Size
Non-test diff: 117 lines (index.ts +7/-2, projection.ts +103/-5).

Refs D25.


obra:
## Round 2 disposition (8b73ff1) — answering the raw panel reviews

Read the raw panel files directly (`reviews/raw/1856-189741748.md`), not the synthesis.

- **Medium (member 1, projection.test.ts "advance out-ranks a slower in-flight all-target refresh")**: fixed. The stale snapshot only had `ref-a`; a fence that discarded the whole snapshot on any stale target would still have passed. Added a fresh `ref-b` to the same snapshot and now assert `apply()` keeps `ref-b` while excluding `ref-a`.
- **Low (member 1, empty `targets`/`apply()` vs `false` contract)**: documented, not normalized. `refresh()` returning `false` means no snapshot was ever produced (reset or read failure) — there's nothing for a caller to record provenance from. An `apply()` that returns an empty set means the read succeeded and the snapshot is real (still worth `recordSubmittedHere`), just nothing in it is still this refresh's to project. Normalizing the empty case to `false` would have broken that distinction — the web adapter calls `recordSubmittedHere(snapshot)` unconditionally on every accepted refresh, including ones whose `apply()` ends up empty.
- **Low (member 3, :112, newer-refresh-failure should still raise the floor for an older in-flight refresh)**: added `"a newer refresh that fails still raises the generation floor for an older in-flight refresh"` — an older refresh starts, a newer refresh for the same target fails, the older refresh then resolves and its `apply()` still excludes the target. Passes under the current (non-rollback) failure handling.

**Mechanism change (not itemized as a separate finding, but the direct fix for C3c's Medium at pendingTurnsStore.ts:136-151, which explicitly named this module as the fix's home):** `refresh()` used to freeze its decision the instant the read resolved and hand back a fixed `targets` set. A live commit's `advance()` (or a newer refresh) landing after that decision but before the caller applies the snapshot would still out-rank the fence's internal bookkeeping, but the caller's already-frozen set didn't know that. `refresh()` now returns `{ snapshot, apply() }`; `apply()` re-checks the floor at the moment the caller calls it and must be called synchronously, immediately before projecting. New failing-first test: `"advance landing after refresh resolves still out-ranks a stale apply"` — calls `advance()` between `refresh()` resolving and calling `apply()`, asserts the target is excluded. Falsified: reverting `projection.ts`'s production diff made 5 tests fail with `TypeError: result.apply is not a function` (the old `.targets` field no longer exists under the new tests); restored cleanly.

### A finding I did not act on (flagging for visibility, not fixing)
**Medium (member 2, "A rejected read permanently advances the generation floor, so it can suppress a concurrent successful refresh")** was in the raw review but not in my assigned finding list. I read it and the repro is real: an all-targets refresh `g1` starts; a specific `ref-a` refresh `g2` starts and its read rejects; `g1`'s read later succeeds but is discarded because `refreshGenerations["ref-a"]` is still pinned at `g2` (the catch path never rolls it back). I traced the natural fix (roll back the reserved generation on failure, unless something newer already superseded it) and it directly **contradicts** member 3's Low, which the brief assigned me to fix by adding a test asserting the opposite: that a newer refresh's failure *should* keep raising the floor and excluding an older in-flight refresh's result. I can't satisfy both without a design call on which behavior is actually wanted (self-healing via "lifecycle discovery retries" per the existing catch comment, vs. never dropping a fresh successful read). Left as-is; reporting to the coordinator as a residual rather than picking a side unilaterally.

### Gates
- `npx vitest run appwire-client/typescript/state/mutation/projection.test.ts` — 11/11 pass
- `npm run check` (mobile-native tsc) — clean for this diff (3 pre-existing `react-test-renderer` module-resolution failures on this branch are unrelated to projection.ts; confirmed present with the diff reverted too — node_modules drift from this branch's rebased main, not caused by this change)
- `npx biome ci ../../../appwire-client/typescript` — clean

### Falsification
Reverted `projection.ts`'s production diff (`git diff -- projection.ts | git apply -R`), ran `projection.test.ts`: 5 failures, all `TypeError: result.apply is not a function` (the tests calling the new `apply()` step). Restored via re-editing to the identical content (confirmed via `git diff --stat` matching before/after); `git status --short` clean before committing.

Head: `8b73ff1e21a8423807f94df5e5f9cd97704c85f4`

obra:
## Reply to member 2's Medium ("A rejected read permanently advances the generation floor, so it can suppress a concurrent successful refresh")

I implemented this finding in round 3 (rolling back a failed refresh's generation reservation unless superseded), then reverted it. The repro in the finding is real, but the conclusion is wrong: rolling back turns out to violate the fence's actual, already-tested contract.

`cmd/evener-hub/frontend/src/panes/session/composer/queue/pendingTurnsStore.test.ts` has a pre-existing, still-passing oracle - `"a failed newer refresh fences an older pending snapshot"` (`test.each` over all-targets/specific combinations, from #881, predates this work by a long way) - that pins the opposite behavior on purpose: a refresh requested later, even one that fails, proves any read started earlier is stale relative to that later request point, so the earlier read's snapshot must not apply once something newer was asked for. The cure for a target starved by a failed read is a newer successful refresh; applying older data is never the cure, because "older" isn't just about wall-clock duration - it's about which read the caller last asked to be authoritative.

I confirmed this concretely: implementing the rollback made all 4 parameterizations of that oracle fail (an older refresh's stale snapshot - captured before a record was settled out-of-band - got re-applied after a subsequent refresh for the same target failed, resurrecting data that should have stayed gone). Reverting the rollback restores that oracle to green with no other change.

A test already asserting a contract outranks a review finding arguing the opposite from first principles alone. Reverted in `7a373afed`→`601688d8e` (net revert commit) on C3b; kept the `apply()`-at-apply-time mechanism from round 2 and this round's member-1/member-3 fixture fixes; restored the original "a newer refresh that fails still raises the generation floor for an older in-flight refresh" test, plus a new one locking in that a commit (a real, later request) still out-ranks regardless of an intervening failed refresh.

Current C3b head: `67d8f9022`

obra:
## Round 3 disposition (80a1020) — answering the raw panel reviews

Read `reviews/raw/1857-18af4bb81.md` directly (this PR's own mechanism, findings raised against C3c's head since the review ran across both, member 0/2/3 agreeing on one class, member 3 at High).

- **Medium/Medium/High (members 0, 2, 3 — `apply()` omits `allTargetsRefreshGeneration` and `refreshEpoch`)**: fixed. `apply()`'s re-check only compared the per-target generation, so it missed the same two things `refresh()`'s own resolution-time decision already considers:
  - An all-targets refresh raises `allTargetsRefreshGeneration` the instant it *starts*, not when (or whether) its read resolves. A single-target refresh that had already resolved and been accepted could still `apply()` after a newer all-targets refresh started and failed — the exact case `"a newer refresh that fails still raises the generation floor"` exists to prevent, reachable through a different door.
  - `reset()` clears and reuses generation numbers. A pre-reset refresh's `apply()` had no way to know a reset happened and could pass simply because a post-reset refresh happened to reuse the same generation number.

  `apply()` now runs the identical predicate `refresh()` used to accept its candidates in the first place: the captured epoch, the all-targets floor, and the per-target floor — re-evaluated live at call time instead of a narrower check.

  Two new failing-first tests: `"apply() excludes a target once an all-targets refresh starts after this refresh resolved"` (asserts exclusion the instant a newer all-targets refresh is *started*, without ever resolving it) and `"apply() excludes a target once reset() has moved on, even if generation numbers repeat"` (a post-reset refresh deliberately reuses the same generation number as the pre-reset one, to rule out a numeric-collision false pass). Both failed against the pre-fix code (`expected true to be false`) and pass now.

### Falsification
Reverted `projection.ts`'s production diff for this round only, ran `projection.test.ts`: exactly the 2 new tests failed (`expected true to be false` in both), the other 12 stayed green. Restored; `git status --short` clean before committing.

### Gates
- `npx vitest run appwire-client/typescript/state/mutation/projection.test.ts` — 14/14 pass
- `npm run check` (mobile-native, real `npm ci`) — clean
- `npx biome ci ../../../appwire-client/typescript` — clean

Own diff vs `origin/main`: `index.ts` +9/-0, `projection.test.ts` +192/-9 (net across all rounds), `projection.ts` +152/-9 (net across all rounds) — 344 insertions / 9 deletions total for the whole PR.

Head: `80a1020fb1afa3f90788e0aa7e6f3e447a426040`

obra:
Round 3 @80a1020: CI green (15); raw panel members (synthesis is broken): three found no issues, one Low (the target-freshness predicate is duplicated between refresh() and apply()). Merging per the Lows rule; the follow-up PR extracts the predicate into one function used by both.

PR #1857, opened 2026-09-18
refactor(sdk): the web switches its projection fence to the package (D25e C3c)

obra:
Stacked on #1856 (D25e C3b), which stacks on #1855 (D25e C3a). Merge C3a then C3b first.

## Mechanism
`pendingTurnsStore.ts` drops its four module-level fence variables (`refreshGeneration`, `allTargetsRefreshGeneration`, `refreshGenerations`, `refreshEpoch`) and its own local `replaceTargetRecords`, switching onto C3a's `replaceTargetRecords` and C3b's `createMutationProjectionFence` over a `persistencePort` that adapts `readMutationPersistence` to `MutationPersistencePort` (`{ read: readMutationPersistence }` — no wrapper logic needed, the shapes already match).

Every call site that touched the old variables now calls the fence instead:
- `readProjectionIntoStore` → `projectionFence.refresh(persistencePort, ref)`
- the `subscribeMutationPersistence` commit callback → `projectionFence.advance(record.targetRef)`
- `submitWithPendingTracking`'s two "is this submission's mount still current" checks → `projectionFence.epoch()`
- `resetPendingTurnsStoreForTests` → `projectionFence.reset()`

## Oracle / tests
`pendingTurnsStore.test.ts` and `pendingTurnsStore.edge.test.ts` (30 tests, including `pendingTurnsStore.test.ts:250` "an older all-target projection cannot erase a newly committed send") keep passing unchanged. Broader consumer sweep (`QueueStrip.test.tsx`, `PendingChips.test.tsx`, `Composer*.test.tsx`, `Session.test.tsx`, `threads.retirement.test.tsx` — 425 tests) also green.

## Falsification
1. **Dependency proof**: temporarily broke the package fence's dominance check in `projection.ts` (`if (generation < Math.max(...))` → `if (false)`, i.e. an all-targets read never yields to a newer per-target generation) — 4 of 29 `pendingTurnsStore.test.ts` tests failed (`a failed newer refresh fences an older pending snapshot`, all 4 parametrized variants), proving the web now actually runs on the package fence rather than a dead import. (Note: test `:250` itself did not fail under this specific break — its scenario's IndexedDB read ordering doesn't happen to exercise that exact line — but it's covered by C3b's own fence unit test "advance out-ranks a slower in-flight all-targets refresh for that one ref", which mirrors it directly with a fake port and does fail when the same check is broken.) Restored; all 29 pass again.
2. **Revert-only-production**: reverted this PR's own diff (`pendingTurnsStore.ts` back to C3a's local-fence version) and reran the oracle suite — still 30/30 green, as expected for a behavior-preserving move (confirmed structurally too: the reverted file has 10 references to the old `refreshGeneration` family and zero to `projectionFence`, the current file the reverse). Restored.

## Gates
- `npx vitest run pendingTurnsStore.test.ts pendingTurnsStore.edge.test.ts` (web): 30 passed
- Broader consumer check (`threads.retirement.test.tsx`, `Session.test.tsx`, `Composer*.test.tsx`, `QueueStrip.test.tsx`, `PendingChips.test.tsx`): 425 passed
- `npm run lint` (frontend biome): clean
- `npx biome ci ../../../appwire-client/typescript`: clean
- `make lint-package-imports` / `make test-api-package`: unaffected (no package files touched), reran to confirm — both pass

## Size
Non-test diff: 100 lines (pendingTurnsStore.ts +30/-70, net -40).

Refs D25.


obra:
## Round 2 disposition (2189ef1) — answering the raw panel reviews

Read the raw panel file directly (`reviews/raw/1857-0c5178b7f.md`), not the synthesis. Only member 0 raised an issue; members 1-3 found none.

- **Medium (member 0, pendingTurnsStore.ts:136-151)**: fixed. `readProjectionIntoStore` destructured `accepted.targets` right after `await projectionFence.refresh(...)` and used that frozen set at `setState()`. A live commit's `advance()` for one of those targets landing between the fence's internal decision and this module's `setState` call would still out-rank the fence, but the frozen set here didn't know it, so the stale snapshot could clobber the newly-committed record.

  The mechanism fix lives in C3b (`state/mutation/projection.ts`): `refresh()` now returns `{ snapshot, apply() }` instead of a fixed `targets` set; `apply()` re-checks the floor at the moment it's called. This PR is the adapter half: `readProjectionIntoStore` now calls `accepted.apply()` immediately before `setState`, synchronously, with no further await in between — the same place the old `targets` destructure used to read from, just re-decided at call time instead of frozen at resolution time.

  I traced the actual call site carefully: as written, `readProjectionIntoStore` has exactly one `await` (the `refresh()` call) and then purely synchronous code (`recordSubmittedHere`, then the `apply()`/`setState` pair), so in real JS microtask semantics nothing can currently interleave between the fence's internal decision and this module's application of it — a macrotask-driven `advance()` (from IndexedDB write completion elsewhere) cannot preempt a single microtask hop. I could not construct a production interleaving that exploits the gap as literally described. That said, this is exactly the assigned fix and it's a strict improvement regardless: it makes the correctness invariant self-enforcing at the fence's boundary instead of relying on this call site never growing an intervening `await` (e.g. a future batching change, a React transition boundary, or a settle-tracking hook inserted between the two calls). Fixed as directed.

### Gates
- `npx vitest run appwire-client/typescript/state/mutation/projection.test.ts cmd/evener-hub/frontend/src/panes/session/composer/queue/pendingTurnsStore.test.ts` (from `cmd/evener-hub/frontend`) — 40/40 pass, including the existing oracle `"an older all-target projection cannot erase a newly committed send"` (pendingTurnsStore.test.ts:249)
- `npm run check` (mobile-native tsc) — clean for this diff
- `npx biome ci ../../../appwire-client/typescript` and `npm run lint` (frontend) — clean

### Falsification
Reverted this PR's production diff (`pendingTurnsStore.ts`'s `targets`→`apply()` change) with the C3b fix still in place, ran `npm run typecheck`: failed with `error TS2339: Property 'targets' does not exist on type 'MutationProjectionRefresh<MutationAttachment>'` — the old destructure no longer compiles against the new fence API, proving this adapter change is load-bearing. Restored via the saved diff (`git apply`); `git status --short` clean before committing.

Rebased onto the new C3b head (`8b73ff1e21a8423807f94df5e5f9cd97704c85f4`) after its round 2 landed; own diff isolated to `pendingTurnsStore.ts` only (`git diff <c3b-head>...<this-head> --stat`).

Head: `2189ef19c78863d9c4d2c298cd959299134ff2b1`

obra:
## Restack note (18af4bb)

No change to this PR's own diff (still isolated to `pendingTurnsStore.ts`, 34 insertions / 70 deletions vs C3b). Restacked onto C3b after its round 3/4: the round 3 rollback fix was implemented, found to break the pre-existing oracle `"a failed newer refresh fences an older pending snapshot"` (#881), and reverted - see the C3b thread for the full account. Full suite here (`projection.test.ts` + `pendingTurnsStore.test.ts`, including that oracle) is 41/41 green on the restacked head.

Head: `18af4bb813440423083adec08b8e2cb5d1920be5`

obra:
## Round 3 disposition (63272aa) — answering the raw panel reviews

Full account and the fix are on C3b: https://github.com/prime-radiant-inc/evener/pull/1856#issuecomment-5733345359

`apply()` in the package (`state/mutation/projection.ts`) omitted `allTargetsRefreshGeneration` and `refreshEpoch` from its re-check, agreed on by 3 of 4 panel members (High from member 3). No production change needed here: this PR's own diff is `pendingTurnsStore.ts` calling `accepted.apply()`, which is unaffected by that predicate's internals. Restacked onto the corrected C3b (own diff unchanged: `pendingTurnsStore.ts`, 34 insertions / 70 deletions, isolated via `git diff <c3b-head>...<this-head> --stat`).

### Gates
- `npx vitest run appwire-client/typescript/state/mutation/projection.test.ts cmd/evener-hub/frontend/src/panes/session/composer/queue/pendingTurnsStore.test.ts` — 43/43 pass, including the #881 oracle
- `npm run typecheck` (frontend), `npm run check` (mobile-native) — clean
- `npx biome ci ../../../appwire-client/typescript`, `npm run lint` (frontend) — clean

Head: `63272aa17732ed4950e3820aa158c758492b7b24`

PR #1858, opened 2026-09-18
fix(sdk): reasoning export (#1731 piece 1/3)

obra:
## Summary
First of three PRs decomposing #1731 (round 5 exceeded the five-review-round budget; the coordinator split it by mechanism). This piece carries only the reasoning-export mechanism.

- Exports `joinedReasoningParagraphs` from `@evener/appwire-client`; `cmd/evener-hub/frontend/.../reasoningFormat.ts` re-exports it instead of carrying its own copy.
- Moves `isActiveItem` into `itemFailure.ts`, retyped against `ItemFailureSignals` (not a full `ItemModel`), and exports it from the package.
- `mobile/src/conversation/project.ts` and `mobile/src/state/conversation.ts` now import the shared `isActiveItem` instead of keeping their own duplicate.

This resolves the round-5 RoboRev Low finding "`isActiveItem` is unused on mobile and type-incompatible" by fixing the type (rather than dropping the export), since mobile's own copy already existed and could adopt the shared helper once retyped.

## Falsification
- Reverting `reducer.ts` (dropping `joinedReasoningParagraphs`) while keeping `reasoningFormat.ts`'s re-export: 8/35 tests fail in `reasoningFormat.test.ts` with `joinedReasoningParagraphs is not a function`.
- Reverting `itemFailure.ts` (dropping `isActiveItem`) while keeping the package export and mobile's consumption: 102/449 tests fail across `mobile/src/conversation/project.test.ts` and `mobile/src/state/conversation.test.ts` with `isActiveItem is not a function`.
- The `index.ts` "shipped module unreachable" falsification doesn't apply here — no module was added or moved, only new named exports on already-shipped modules — so the above load-bearing falsification substitutes for it.

## Gates
- `npx tsc -p tsconfig.build.json --noEmit` (appwire-client/typescript): clean
- `npm run typecheck` (cmd/evener-hub/frontend): clean
- `npm run check` (mobile-native): clean
- vitest: `itemFailure.test.ts`, `transcriptProjector.test.ts`, `reducer.test.ts`, `reasoningFormat.test.ts`, `reasoningFormat.edge.test.ts` — 291/291 passed
- vitest (mobile-native `test:shared`): `mobile/src/conversation/project.test.ts`, `mobile/src/state/conversation.test.ts` — 449/449 passed
- `npm run lint` (biome, cmd/evener-hub/frontend + appwire-client/typescript): clean
- `make lint-package-imports`: PASS
- `make test-api-package`: qualified

## Test plan
[...]

PR #1859, opened 2026-09-18
fix(sdk): image folding + warning text (#1731 piece 2/3)

obra:
## Summary
Second of three PRs decomposing #1731 (round 5 exceeded the five-review-round budget). Stacked on piece 1 (#1858). This piece carries the image-folding and warning-text mechanisms.

- Pins the existing input-images fold (an empty list on the wire says nothing and never erases an already-known image — hydrate, live settle, and older-page merge alike) with regression coverage, plus an explanatory comment on `imagesToItemImagesForSession`. No production logic changed here; the fold was already correct.
- Bounds the warning fold's raw-frame fallback (`rawWarningFrame`, `RAW_WARNING_FRAME_MAX_CHARS`, truncating by code point so a surrogate pair is never split — Medium 3 from #1731's round-5 review).
- Leaves `ItemModel.text` blank instead of duplicating a title/hint with the raw JSON envelope when a message-less warning carries a title/hint (the Low "warning fold text drops title/hint on mobile" from #1732's round-3 review, root-caused).
- Fixes the High from #1732's round-3 review: `transcriptProjector.ts`'s `itemSummary` crashed on a non-string `warning.title` — `item.warning` rides an untyped wire param map, so a malformed frame can hand it a number despite `ItemModel`'s declared `string` type.
- Fixes mobile's half of the same Low: `mobile/src/conversation/project.ts`'s warning projection fell through to the generic "unknown item type" fallback (`item.text || item.output`), so a message-less warning with only a title/hint became a blank row on the phone. `warningFallbackText` now joins them, matching what `WarningItem.tsx` shows on web.

### Dropped from #1731's diff
The round-5 Medium "new turns-reference contract has no production consumer" (`mapTurn`/`settleFirstMatchingTurn`'s early return when a frame names a turn outside the loaded window, plus its two "leaves turns by reference" tests) is dropped rather than carried forward. I grepped the tree for any consumer that reads `model.turns` by reference to gate a rehydrate (the native store's `evener/steering/injected` handling, `TranscriptBody.tsx`'s `useMemo`s) and found none — the contract has no reader, so it's speculative complexity with no test that can fail honestly. Say if there's a consumer I missed.

### Incidental fix
While porting this piece's own warning-bound test I found it was vacuous as originally written: `warning: "x".repeat(10_000)` is a valid plain-string message, so `warningMessage()` returns it directly and the raw-frame fallback (and its bound) never runs — `toBeLessThan` was trivially true. Changed the fixture to `warning: 42` (message-less) with a large `extra` field, which actually reaches `rawWarningFrame`. Falsified: red without the fix (`10058` not less than `10058`), green with it.

## Falsification
- Reverted `reducer.ts`'s warning-fold changes: 2/15 warning tests fail (`toBeLessThanOrEqual(MAX_CHARS)` off by 2, and the message-less-with-title test returns the raw JSON envelope instead of `""`).
- The image-folding tests pass unmodified against unmodified production code — they pin pre-existing behavior, so there is no red state for that half; noted rather than faked.
- Reverted `transcriptProjector.ts`'s `itemSummary` guard: new test fails with `TypeError: item.warning?.title?.trim is not a function`.
- Reverted `mobile/src/conversation/project.ts`'s `warningFallbackText`: new test's `detail.output` assertion fails (`undefined` does not contain `"Sandbox blocked"`).

## Gates
- `npx tsc -p tsconfig.build.json --noEmit` (appwire-client/typescript): clean
- vitest: `reducer.test.ts` + `transcriptProjector.test.ts` — 242/242 passed
- vitest (mobile-native `test:shared`): `project.test.ts` + `conversation.test.ts` — 450/450 passed
- `make test-api-package`: qualified (index.ts untouched this piece)

## Test plan
- [x] CI green on this PR (builds the merge with main)


obra:
## Round 2 disposition (e506b02) — answering the raw panel reviews

**High — mobile/src/state/conversation.ts:2527, live lifecycle path drops attachments on absent/empty images**
Fixed. Added `findFoldedItem` (conversation.ts) to look up the reducer-folded `ItemModel` for the item that `applyThreadNotification` just merged — its `mergeItemImages` already preserved the prior images when the wire frame carried none. Attachments for `item/started`/`item/completed` now come from `itemAttachments(foldedItem)` (newly exported from project.ts) instead of `projectItemAttachments(params.item)` on the raw wire item, with `projectItemAttachments` kept only as a defensive fallback when the item can't be found in `conv.turns`.
Test: `mobile/src/state/conversation.test.ts` — "preserves existing attachments when item/completed for the same user message carries no images field". Falsified: reverted the production diff, test failed (`expected undefined to deeply equal [...]`), restored, test passes with the fix.

**Medium — conversation.ts:2810-2811, live warning branch ignores reducer-folded text/hint**
Fixed. The live `"warning"` notification case now reads the item the reducer's own `case "warning"` fold just appended to `conv.turns[activeTurnId]` (title/detail sourced from `foldedItem.warning?.title` and `foldedItem.text || foldedItem.warning?.hint`), falling back to the old `params`-based construction only when there's no active turn (matching the reducer's own "nowhere wire-true to land" drop).
Test: "shows the reducer-folded hint text for a title/hint-only warning with an active turn" — falsified (failed with `detail: ""` before the fix, passes after).

**Medium — reducer.ts:1697, runtime-unvalidated title/hint copied into ItemModel.warning**
Fixed at the fold: `title`/`hint` are now `typeof === "string" ? value : undefined`, so a malformed frame's object/array/number value never rides through as a string-typed field.
Test: "a runtime non-string title/hint folds to undefined, never an object ItemModel.warning claims is a string" — falsified.

**Medium — WarningItem.tsx:32, title/hint rendered as React children with only truthiness checks**
Fixed. Exported `hasWarningText` from the package (reducer.ts already had it private) and use it in WarningItem.tsx to gate title/hint before rendering, so a non-string value is coerced to `undefined` rather than reaching JSX as a child.
Test: "an object-valued title/hint ... never reaches the DOM as a React child" — falsified (threw "Objects are not valid as a React child" before the fix).

**Low — whitespace-only title/hint present in WarningItem but blank in hasWarningText**
Fixed by the same `hasWarningText` unification above — WarningItem now uses the identical trim-based predicate the reducer's raw-frame fallback decides against, so the two never disagree.
Test: "whitespace-only title/hint counts as absent, the same reading hasWarningText gives item.text's raw-frame fallback".

**Low — mobile reasoning projection reads only item.text, live deltas land in reasoningSummaries**
Measured: confirmed via reducer.ts's `appendReasoningDelta`/`mergeReasoning` that live `item/reasoning/summaryTextDelta` chunks accumulate in `reasoningSummaries` and never touch `item.text`, so a mid-stream reread's canonical projection (`mobile/src/conversation/project.ts`) showed an empty reasoning row. Fixed: routes through `joinedReasoningParagraphs(item.reasoningSummaries).join("\n\n") || item.text`, matching the web's ThinkBlock.tsx reading.
Test: "projects live reasoning deltas (reasoningSummaries) even when the settled item.text is empty" — falsified.

Non-test production diff: ~101 lines (index.ts +1, reducer.ts +14/-3, WarningItem.tsx +14/-6, project.ts +17/-4, conversation.ts +55). Under the 150-line target.

Gates: vitest (appwire-client reducer.test.ts, WarningItem.test.tsx, mobile/src/conversation/project.test.ts, mobile/src/state/conversation.test.ts, full mobile/src suite) all green; `npm run typecheck` (frontend) clean; `npm run check` (mobile-native tsc) clean once `npm ci` restored the worktree's missing `react-test-renderer` (see report — worktree's node_modules is a real dir, not the shared symlink); `npx biome ci ../../../appwire-client/typescript` and `npm run lint` clean; `make lint-package-imports` PASS; `make test-api-package` PASS (index.ts gained one new named export, `hasWarningText`, on an already-reachable module — no module-row falsification applies, since nothing was added/moved as a module).

Head: e506b02c2063b4a691d4da8efb12219f5b3d377a (rebased onto main after #1858 squash-merged; own-diff lines verified byte-identical before/after the rebase).

obra:
## Round 3 disposition (9dd7f6b) — answering the raw panel reviews and the native CI regression

**CI regression (native + static jobs) — mobile-native/src/liveImages.test.ts**
Root-caused and fixed. Round 2's High fix (routing the live path's attachments through the reducer-folded item) is architecturally correct — the reducer's `mergeItemImages` treats an empty or absent input-images list as "unchanged" for user messages/steering (input images carry no wire removal signal at all, unlike output images, which the wire can mark explicitly empty on purpose). `liveImages.test.ts`'s three failing tests were themselves stale relative to that already-decided rule: the exact same fix (a `mergeItemImages` that reads absent/empty input images as "keep") was already authored and closed as #1656 on a sibling branch (`claude/sdk-d23c2b-rows-projection`, commit `38cac024e`, not yet merged to main), which rewrote these same three tests to expect preservation, not removal. Brought them in line with that already-established contract rather than reverting the round-2 fix, and added the reviewer's absent-field case beside them ("keeps live input images when a settle omits the images field entirely"). Also fixed the same stale assumption in `mobile/src/state/conversation.test.ts`'s "replaces same transcriptKey across wire IDs" test (it didn't seed `conv.turns`, so it was accidentally exercising the fallback path instead of the fix).

**Medium (members 0, 3) — live warning row drops the hint when a message is also present**
Fixed. `[folded.text, folded.hint].filter(hasWarningText).join(" — ")` composes both instead of `text || hint`.
Test: "composes both message and hint in the live row's detail" — falsified (was `"rate limit approaching"`, missing the hint, before the fix).

**Medium (members 0, 2, 3) — no-active-turn warning path bypasses reducer validation**
Fixed by consolidating every warning row onto one exported `foldWarningParams` (reducer.ts): the reducer's own "warning" fold and the mobile live store's no-active-turn fallback now both call it, so a malformed `title`/`message`/`warning` value can no longer reach a string-typed timeline field un-sanitized regardless of whether there's an active turn. Also handles the polymorphic `warning` field (bare string or `{message}` object) in the no-active-turn path, which the old fallback never read.
Tests: "sanitizes a malformed warning without an active turn instead of copying params raw", "reads a polymorphic warning field without an active turn" — both falsified.

**Low (member 1) — live title uses `?? "Warning"`, doesn't treat whitespace-only as absent**
Fixed: `hasWarningText(folded.title) ? folded.title : "Warning"`.
Test: "falls back to the generic Warning label for a whitespace-only title with an active turn" — falsified.

**Low (member 1) — hasWarningText re-implemented inline at reducer.ts:1217, project.ts:537, transcriptProjector.ts:166; not a type predicate**
Fixed: `hasWarningText` is now `(value: unknown): value is string`, and both `warningFallbackText` (project.ts) and `itemSummary` (transcriptProjector.ts) call it instead of repeating `typeof/trim`.

**Low (member 2) — transcriptProjector.test.ts:178-181 comment says the reducer never coerces title, which round 2 made false**
Reworded to say the guard is defense-in-depth for a directly-constructed ItemModel, not the reducer's own output.

**Low (member 2) — conversation.ts:2856-2858, `foldedItem.warning?.title` only type-coerced, not blank-coerced**
Fixed by the same `hasWarningText` title fallback above.

**Low (member 3) — WarningItem.tsx review-round narration in comments**
Stripped the "(kata Low finding, #1731 piece 2 round 2)" / "(#1731 piece 2 round 2 Medium finding)" parentheticals; the comments now state the invariant without process history.

**B (#1732) heads-up — new Low: `source` passed through unchecked**
`foldWarningParams` guards `source` the same way as `title`/`hint` (`typeof params.source === "string" ? params.source : undefined`); covered by the same fold. Test: extended "a runtime non-string title/hint folds to undefined" into "...title/hint/source...", falsified for `source` specifically.

Non-test production diff (round 3 only): 121 lines (index.ts +2/-1, reducer.ts +41/-12, transcriptProjector.ts +2/-1, WarningItem.tsx +3/-4, project.ts +2/-3, conversation.ts +29/-17). Test diff: 287 lines.

Gates: vitest (reducer.test.ts 212, transcriptProjector.test.ts, WarningItem.test.tsx, mobile-native full suite 800, shared mobile/src 782) all green; `npm run typecheck` (frontend) and `npm run check` (mobile-native) clean; `npx biome ci` / `npm run lint` clean; `make lint-package-imports` PASS; `make test-api-package` PASS.

Head: 9dd7f6bddb9de241a41e15f720d1a52145cd24f5

obra:
## Round 4 disposition (570c454) — answering the raw panel review

**Medium (member 0) — canonical mobile reasoning row can show stale seeded text**
Root-caused: reducer.ts's `wireItemToModel` seeds `reasoningSummaries` from ANY non-empty initial wire text (e.g. `item/started` carrying a draft/partial text), and `mergeReasoning` keeps that seed across later merges once it's set (by design — it's what makes live delta chunks survive a settle). My round-2 fix's `reasoningSummaries-first` read meant a later completion's authoritative `item.text` could be masked by that stale seed. Fixed at the projection (mobile/src/conversation/project.ts): reads `item.text` first — always the item's most authoritative settled value (reducer.ts's `mergeCompletedText`) — and falls back to the joined summaries only when text is genuinely blank (the mid-stream case round 2 was actually fixing).
Test: "shows the completion's authoritative text over a stale seeded reasoningSummaries entry" (item/started with `text: "draft thought"`, then item/completed with `text: "final thought"`, asserts the projection shows the completion) — falsified (showed "draft thought" before the fix).

**Low (member 1) — live-row id embeds the raw folded text (up to 2000-char JSON fallback)**
Fixed: the id now builds from the already-sanitized `title` (`hasWarningText(folded.title) ? folded.title : "Warning"`) instead of `folded.title ?? folded.text ?? "warning"`.
Test: "keeps the live row id short for a message-less warning with no active turn" (a message-less frame with a 500-char `extra` field, forcing the raw-frame fallback long) — falsified (558-char id before the fix, `warning:Warning:N` after).

**Low (member 1) — foldWarningParams stores whitespace-only title/hint/source verbatim**
Fixed: `title`/`hint`/`source` now route through `hasWarningText` (trim-aware) instead of a bare `typeof === "string"` check, so blank becomes `undefined` at the fold. Updated the pre-existing "cancel-shaped warning" test, which had asserted `hint: ""` verbatim for this exact wire shape — that assertion encoded the now-recognized-as-wrong behavior the review flagged, not a case I invented.
Test: "a whitespace-only title/hint/source folds to undefined at the source, not just at each consumer" — falsified.

**Low (member 2, reducer.test.ts:4805-4807) — test name/comment claims "no threadId/ref wrapping" but the frame includes both**
Read and confirmed: the test is a routed frame (matches the `test.each`'s `["number warning", 42]` case exactly), not a routing-less one — a genuinely routing-less frame is dropped by `notificationTargetsThread` before it ever reaches this fold. Reworded the test name/comment to describe what it actually verifies (kept as its own named test since it's the exact shape a prior RoboRev round named directly), rather than deleting the duplicate coverage.

Non-test production diff: 34 lines (project.ts +18/-6, conversation.ts +6/-1, reducer.ts +10/-2). Test diff: 126 lines.

Gates: vitest (reducer.test.ts 210, mobile-native full suite 800, shared mobile/src 784) all green; `npm run typecheck` / `npm run check` clean; `npx biome ci` / `make lint-package-imports` clean.

Head: 570c4543c442fec04f0a7ed8d50dc23a9fc76c9f

obra:
## Round 6 disposition (0381d57) — /simplify's three fix-before-merge findings, plus one Medium from B's panel

**Simplify finding 1 — usesFold branch recomputes what foldWarningParams(n.params) already returns**
Fixed exactly as suggested: the reducer's "warning" fold computes `foldWarningParams(n.params)` purely from `n.params`, so the store calls it directly and unconditionally instead of reconstructing the same value field-by-field off `foldedTurn?.items.at(-1)`. Dropped `foldedTurn`/`foldedItem`/`usesFold` and the "last item in the turn is mine" assumption entirely. `hasWarningText(folded.title)` collapsed to `folded.title ?? "Warning"` since `foldWarningParams` already guarantees string-or-absent. Dropped the now-unused `WarningFold` export from `index.ts` (nothing outside the package imports it anymore).

**Simplify finding 2 — findFoldedItem re-implements the reducer's itemIdentityMatches**
Fixed: widened `itemIdentityMatches`'s parameter type to the structural shape it actually reads (`{id, transcriptKey?}`) so a wire `ThreadItem` satisfies it too, exported it, and rewrote `findFoldedItem` as a loop over `conv.turns` calling it instead of a local copy of the rule.

**Simplify finding 3 — reducer.test.ts duplicate tests**
Deleted the `test.each` at (then) :2615 and the standalone test at :2747 — both re-tested `#1656`'s empty-input-images rule origin/main's `inputImageSettleCases` suite and the `mergeOlderItemPage` empty-list test already pin. Kept the hydrate-level test and the `turn/completed` "full" analogue the review flagged as genuinely new. ~145 duplicated lines removed.

No behaviour change for any of the three. Verified: full mobile-native suite (800 tests), shared mobile/src suite (785 tests), reducer.test.ts (217), `npm run typecheck`/`npm run check`, `npx biome ci`, `make lint-package-imports`, `make test-api-package` (index.ts changed) all green before and after.

**Medium (B's #1732 panel at 784126946) — project.ts:405-416, the round-4 item.text-first flip loses streamed reasoning for an ACTIVE item**
Measured first: `reducer.ts`'s `appendReasoningDelta` appends every live delta to `reasoningSummaries` only, never to `item.text` — confirmed by writing the failing test before touching the fix (an active item/started with a non-empty partial seed, then a delta, re-projects to the stale seed instead of the grown summary). The finding holds; not refuted. Fixed per the stated rule: settled items keep the round-4 authoritative-text precedence (a stale seeded summary must not mask a real completion), active items prefer the joined summary once it's longer than the seed (comparing lengths distinguishes the two without a third model field — a settled item's text is the longer, complete value once summaries stop growing, while an active item's joined summary overtakes its seed as soon as a delta arrives).
Test: "shows the growing joined summary over a stale partial seed for an ACTIVE item" — falsified against the unconditional item.text-first form (returned the 12-char seed instead of the 51-char grown summary).

Non-test production diff (this round): index.ts -1/+0, reducer.ts +9/-25 (net removal), conversation.ts +5/-16 (net removal), project.ts +17/-11. Test diff: reducer.test.ts -145 net, project.test.ts +53.

Head: 0381d57b7777ff7383d986a59e346a9d80996ed7

`git diff -M --stat origin/main...HEAD` (cumulative):
```
 appwire-client/typescript/index.ts                 |   3 +
 appwire-client/typescript/reducer.test.ts          | 308 ++++++++++++++++++++-
 appwire-client/typescript/reducer.ts               |  83 +++++-
 appwire-client/typescript/transcriptProjector.test.ts |  12 +
 appwire-client/typescript/transcriptProjector.ts   |   8 +-
 cmd/evener-hub/frontend/.../WarningItem.test.tsx   |  39 +++
 cmd/evener-hub/frontend/.../WarningItem.tsx        |  13 +-
 mobile-native/src/liveImages.test.ts               | 138 ++++++---
 mobile/src/conversation/project.test.ts            | 167 ++++++++++-
 mobile/src/conversation/project.ts                 |  43 ++-
 mobile/src/state/conversation.test.ts              | 207 +++++++++++++-
 mobile/src/state/conversation.ts                   |  54 +++-
 12 files changed, 1018 insertions(+), 57 deletions(-)
```

This is my last round on this lane — retiring after the rebases/pushes. State note (open findings for #1862, mechanism split, gates cheat-sheet) is at `.superpowers/sdd/2026-09-12-mobile-landing-queue-cont/a-lane-state-note.md` in the coordinator worktree.

Issue #1823, opened 2026-09-18
askPending gate hides a pending ask_user question when reconnecting to a pre-#1731 daemon


obra:
liveAskQuestions (appwire-client/typescript/deriveAskQuestions.ts) now gates entirely on ThreadModel.askPending, which comes from EvenerThread.askPending on the wire. That field is optional (a pre-#1731 daemon never sends it), and hydrateThread defaults its absence to false.

Before #1731, a pending ask_user call was derived from transcript shape (isResolutionTurn), so reconnecting to an older daemon with an unanswered question still showed it. After #1731, the same reconnect silently hides the question until the daemon is upgraded or the question is otherwise resolved.

Found by RoboRev round 2 on #1737 (10beda5c), reported as part of the "askPending gate breaks package qualification and hides questions for old daemons" Medium: appwire-client/typescript/deriveAskQuestions.ts:106, appwire-client/typescript/scripts/qualify-package.mjs:62,66,69.

Options (per the review's own fix suggestion): preserve whether askPending was present on the wire and fall back to the transcript derivation when unknown, or accept the compatibility gap for the internal beta and document it. Filing so Jesse can decide; no fix has been made for this half of the finding.

obra:
Ruling (Jesse, 2026-09-18): no fallback. The client renders a pending ask only while the wire's askPending flag is true; daemons that predate the field render no pending ask until they are updated, and the hub restarts daemons on upgrade. No backward-compatibility code in the client. Reviewers citing this get a pointer here.

PR #1862, opened 2026-09-18
fix(sdk): ask resolution (#1731 piece 3/3)

obra:
## Summary
Third of three PRs decomposing #1731 (round 5 exceeded the five-review-round budget). Stacked on piece 2 (#1859). This piece carries the ask-resolution mechanism.

- Gates `liveAskQuestions` on `ThreadModel.askPending` instead of re-deriving "is anything pending" from transcript shape: the server already knows (`appwire.EvenerThread.AskPending` on hydrate, `appwire.ThreadStatusChanged.AskPending` on every status change that can move it — this wire plumbing already landed on main separately). Earlier rounds of this file's own re-derivation attempts kept disagreeing with the server (the round-5 review's Medium finding).
- "Which questions" still needs a transcript scan (`askPending` is a bool, not a list). Replaces `lastUserMessageIndex` (plain-`userMessage`-only boundary) with `isResolutionItem`/`lastResolutionIndex`, which also treats an interrupt (`steeringKind: "interrupted"`) and an accepted user steer (`source: "user"`) as resolving the whole pending set — the two places outside a plain user message where the server itself clears `askPending` (`session_lifecycle.go`'s `clearAskPending` call sites).
- **Absent-`askPending` policy**: unchanged — an absent wire field still folds to `false` (existing hydrate/notification code). A completed-but-unanswered ask on a daemon old enough to omit the field renders no dock. Per Jesse's standing rule, no backward-compatibility code without his explicit approval; old-daemon compatibility is tracked separately as **#1823**.

### Fixture fixes (every consumer this gate change breaks)
- Web: `askDock.test.ts`, `deriveAskQuestions.test.ts` (+20 new tests for the resolution/gate logic), `Session.test.tsx`, `Composer.integration.test.tsx` (the 3 CI-round-5 failures — this file has its own local `ackAskUserCall` helper, not shared with `AskDock.test.tsx`'s), `Composer.test.tsx`, `AskDock.test.tsx`, `askDockStore.test.ts`.
- Package: `scripts/qualify-package.mjs` smoke script.
- Dev fixture: `src/dev/surface-sections/composer.tsx`.
- Mobile: the 4 `project.test.ts` tests #1731 already touched, **plus** two more the round-5 review flagged as vacuous: "does not project an ask_user answered by a later user message" and "...as answerable" both asserted `not.toContain("question")` against a fixture whose default `askPending: false` made the assertion trivially true regardless of whether the resolution/error-exclusion logic was even correct. Added `askPending: true` to both; verified by temporarily disabling the error exclusion in `isAckedAskUserItem` and watching the second test fail (`expected [ 'question' ] to not include 'question'`), then restoring it.

## Falsification
- Reverting `deriveAskQuestions.ts` entirely: 8/13 `deriveAskQuestions.test.ts` tests fail immediately (empty array where questions were expected).
- Running the full touched-file suite before any fixture fix: 82/390 tests fail across `askDock.test.ts`, `Session.test.tsx`, `Composer.test.tsx`, `AskDock.test.tsx`, `askDockStore.test.ts`.
- Mobile: reverting the fixture fix on the two "vacuous" tests and instead temporarily breaking `isAckedAskUserItem`'s error exclusion — the "does not project an errored ask_user as answerable" test fails (`expected [ 'question' ] to not include 'question'`), proving the fixed fixture actually exercises the logic it names. Restored.

## Gates
- `npx tsc -p tsconfig.build.json --noEmit` (appwire-client/typescript): clean
- vitest (web): `askDock.test.ts`, `deriveAskQuestions.test.ts`, `Session.test.tsx`, `Composer.integration.test.tsx`, `Composer.test.tsx`, `AskDock.test.tsx`, `askDockStore.test.ts` — 464/464 passed
- vitest (mobile-native `test:shared`): `project.test.ts` + `conversation.test.ts` — 450/450 passed
- `make test-api-package`: qualified (index.ts/tsconfig.build.json untouched this piece; the qualifier's own fixture args needed the new required `askPending` field, which is a smoke-test shape fix, not a behavioral falsification — that coverage lives in `deriveAskQuestions.test.ts`)

## Test plan
- [x] CI green on this PR (builds the merge with main)


obra:
## Round 2 disposition (3977e29) — answering the raw panel reviews

**Medium — askPending tri-state collapses "server says nothing pending" with "producer never sent the field", hiding asks on old-daemon/source-backed threads**
This is Jesse's decision, tracked as #1823. No code change here — pointer only.

**Medium — reducer.ts:1697, warning fold copies unchecked title/hint into ItemModel.warning**
This is piece 2's fix (#1859), already landed there: title/hint are now sanitized to `typeof === "string" ? value : undefined` at the fold, and `WarningItem.tsx` guards on the package's `hasWarningText` before rendering either. Piece 3 rebases on top of piece 2's fixed head, so this PR carries the fix without a separate change here.

**Low — reducer.ts:1206 (now ~1230), rawWarningFrame allocates the whole frame before slicing**
Fixed. `rawWarningFrame` now takes a UTF-16 prefix of the JSON string bounded to `RAW_WARNING_FRAME_MAX_CHARS * 2` code units *before* calling `Array.from` — that prefix always contains at least `RAW_WARNING_FRAME_MAX_CHARS` code points (every code point is at most two UTF-16 units), so it can never under-produce, and it never expands the full frame (up to the transport's 128 MiB limit) into a per-character array just to keep the first 2000.
Test: "an oversized warning frame's fallback bounds its input before Array.from, not after" — spies on `Array.from` with a 500,000-char frame and asserts every string it actually expanded stayed under `MAX_CHARS * 4`. Falsified: without the fix, the spy observed a 500,058-char argument (`expected 500058 to be less than 8000`); with the fix, all observed args are bounded. The existing surrogate-pair-at-the-boundary regression test (`reducer.test.ts`, "never splits a surrogate pair at the truncation boundary") still passes unchanged — the 2x prefix margin is exactly tight enough to never re-introduce a mid-pair cut (worst case: every character up to the boundary is a 2-unit surrogate pair, landing the cut exactly on a pair boundary).

Non-test production diff: reducer.ts +9/-1 (~9 lines). Well under the 150-line target.

Gates: vitest (appwire-client reducer.test.ts, full 210 tests) green; `npm run typecheck` (frontend) clean; `npx biome ci ../../../appwire-client/typescript` clean; `make lint-package-imports` PASS. index.ts/tsconfig untouched by this piece, so `make test-api-package` wasn't required (ran it anyway via piece 2's rebase check; unaffected).

Head: 3977e29227446fb1fa17481b3e1d143042712eaf (rebased onto piece 2's new round-2 head; own-diff lines verified byte-identical before/after the rebase).

obra:
## Round 3 disposition (614477a) — answering the raw panel reviews

**CI red (inherited from #1859's native regression)**
Fixed upstream on #1859 (round 3, commit 9dd7f6b) — this branch is now rebased onto that fix. See #1859's round 3 disposition for the root cause (a stale test oracle, not a bug in the "absence is not removal" rule).

**Medium — no-active-turn mobile warning path is the same class fixed in piece 2**
Reply-with-pointer as directed: fixed on #1859 (piece 2) round 3, commit 9dd7f6b — the reducer's `foldWarningParams` is now used by both the reducer's own "warning" fold and the mobile live store's no-active-turn fallback, so this path can no longer bypass validation. This branch inherits that fix via the rebase.

**Medium — rawWarningFrame still builds the full JSON.stringify(params) before slicing**
Fixed. `JSON.stringify(params)` walked the whole object graph regardless of the later `Array.from` bound on its output. Added `prunedForStringify`: truncates every string to 2000 UTF-16 units, caps every array to 50 items, cuts nesting off at depth 6 — all applied *before* `JSON.stringify` ever sees the frame.
Tests: "bounds the frame before JSON.stringify walks it, not just its output" (spies on `JSON.stringify`'s own return length with a 5 MB single-string field: `5000058` chars before the fix, bounded under `10,000` after) and "prunes deep nesting before JSON.stringify walks it" (a 50-level-deep structure; the leaf value no longer survives into the bounded output). Both falsified.

**Low — the intermediate slice(0, 4000) can split a surrogate pair, leaving a lone high surrogate**
Measured before fixing: proved by exhaustion (pigeonhole) that this specific corruption is unreachable through `applyNotification` given the existing 2x code-point margin — any UTF-16 span of 3999 units decomposes to at least 2000 code points regardless of content mix (worst case: 1999 two-unit chars + 1 one-unit char), so a lone surrogate at unit index 3999 can never land inside the final `slice(0, 2000)` window. Separately confirmed `JSON.stringify` always escapes a genuinely unpaired surrogate in string *content* to the 6-character literal `\udXXX` rather than emitting a raw code unit, so malformed input content can't reach this path either. Applied the reviewer's suggested fix anyway (drop a trailing high surrogate from the UTF-16 prefix before `Array.from`) as cheap, harmless defense-in-depth — did not add a dedicated regression test since no reachable repro exists through the public reducer API, and unit-testing the private helper in isolation isn't warranted for an unreachable case.

**Low — live warning title `?? "Warning"` lets a blank stored title through**
Fixed upstream on #1859 (piece 2) round 3 — `hasWarningText(folded.title) ? folded.title : "Warning"`. Inherited via the rebase.

Non-test production diff (round 3 only): 59 lines (reducer.ts). Test diff: 72 lines.

Gates: vitest (reducer.test.ts, 212 tests) green; `npm run typecheck` clean; `npx biome ci` clean; `make lint-package-imports` PASS.

Head: 614477a54d07b6eacd3ff0149a415ad0d23e3ab8 (rebased onto #1859's new round-3 head; own-diff lines verified byte-identical before/after the rebase).

obra:
## Round 4 disposition (f77f43b) — folding in B's (#1732) raw panel findings on this piece's rawWarningFrame bound

B's round-4 panel (reviews/raw/1732-26aa83299.md) pointed both its Mediums at this piece's `prunedForStringify`/`foldWarningParams`, so fixing them here per the coordinator's instruction:

**Medium — prunedForStringify caps string/array length and depth, but not object key count or total output**
Fixed: added `RAW_WARNING_FRAME_MAX_OBJECT_KEYS` (50, same cap as arrays already had) and a total-node budget (`RAW_WARNING_FRAME_MAX_NODES`, 500) threaded through the whole recursive walk — the per-level caps alone left a gap for many small objects/keys that individually stay within bounds but sum to a large tree.
Test: "bounds a many-key object, not just deep nesting or long strings" (a 100,000-key object) — falsified with BOTH the key cap and the node budget disabled simultaneously (1,488,947-char stringify output; each cap alone was independently sufficient to catch this specific shape, confirmed by falsifying each separately first).

**Medium — only the message-less raw-frame fallback was bounded; a huge message/title/hint/source reaches the model verbatim**
Fixed: extracted the fallback's code-point-safe truncation into a shared `boundedCodePoints`, applied to every string `foldWarningParams` puts into the model — `text` (covering both `warningMessage`'s return and the raw fallback in one call), `title`, `hint`, `source`.
Test: "an oversized message, title, hint, and source are each bounded at the fold" (5 MB each) — falsified (5,000,000 chars unbounded before the fix, <4000 after).

Reply-with-pointer for B's own PR: posted there with this commit.

Non-test production diff (round 4 only): 77 lines. Test diff: 76 lines.

Gates: vitest (reducer.test.ts, 215 tests) green; `npm run typecheck` clean; `npx biome ci` clean; `make lint-package-imports` PASS; full mobile-native suite (800) and shared mobile/src (784) green (unaffected by this piece, run for safety since B rebases on top).

Head: f77f43bb6d249e11e995ef906047c4df4fc5cb65 (rebased onto #1859's new round-4 head; own-diff lines verified byte-identical before/after the rebase).

obra:
## Round 5 disposition (99bde6e) — two new findings from the 614477a54 panel

The rest of that panel (member0/1/2's prunedForStringify key-count Mediums, member1's title/hint/source verbatim Medium, member2's stale-comment-adjacent Low on the same lines) was already fixed in round 4 (f77f43b) before this review landed — see that round's disposition. Two genuinely new findings:

**Medium (member 3) — deriveAskQuestions.ts:86, human-note steering wrongly resolves a pending ask**
Measured first: confirmed `steeringKind: "human-note"` rides the wire (`appwire-client/typescript/types.gen.ts`'s `STEERING_KINDS` includes `"human-note"`; `appwire/types.go`'s `ThreadItem.SteeringKind`/`Source` carry it) and that a human-note steer really does carry `source: "user"` simultaneously (`agent/session_notes_rpc.go`'s `steeringOrigin`/`SteeringKindHumanNote` machinery) — the OLD comment on `isResolutionItem` had actually already documented this shape correctly but drawn the wrong conclusion from it ("`source === 'user'` alone is still the resolving half of this OR"). Found the same exclusion already precedented elsewhere for an unrelated reason: `cmd/evener-hub/frontend/.../layoutRoles.ts:46` routes a human-note steer to a divider instead of a user bubble via `item.steeringKind !== "human-note"`.
Fixed: `isResolutionItem` now excludes `steeringKind === "human-note"` from the user-steer resolution condition.
Test: "a steering item with source user AND steeringKind human-note does NOT resolve the ask" (an acked ask followed by a user-sourced human-note item, asserts the ask stays live) — falsified (returned `[]` instead of `["call_1:0"]` before the fix).

**Low (member 2) — conversation.ts:2546-2547, stale comment contradicts the empty-image rule**
Reworded: "Lifecycle events replace the whole source item, including any companion attachment row" now continues with the actual current rule (empty/absent input images = unchanged, closes #1656) and points at `findFoldedItem`/`itemAttachments`, instead of the old "An empty image list removes it" the code beneath it no longer does.

**Residual found while re-verifying the key-count cap**: the reviewers who called out "key length"/"oversized property keys" (member0, member3) were right about something round 4 missed — `prunedForStringify`'s key-count cap (50) bounds how many properties survive, but not how long any ONE property name is. A single oversized key name rode through verbatim. Fixed in this round too (not explicitly requested, but the same class the coordinator flagged as "already have" — measured and found it wasn't fully covered, so closing it now rather than leaving a known gap). Test: "bounds an oversized property NAME, not just its value" (a single 5 MB key name) — falsified (5,000,065-char stringify output before the fix).

Non-test production diff (this round): 26 lines (deriveAskQuestions.ts +12/-6, reducer.ts +6/-1, conversation.ts +6/-1). Test diff: 53 lines.

Gates: vitest (deriveAskQuestions.test.ts 21, askDock.test.ts, reducer.test.ts 216) green; `npm run typecheck` / `npm run check` clean; `npx biome ci` / `npm run lint` / `make lint-package-imports` clean; full mobile-native suite (800) and shared mobile/src (784) green.

Head: 99bde6efbd20042dffcf1041cfcf727afe91410a

`git diff -M --stat origin/main...HEAD` (cumulative, includes piece 2's stacked changes):
```
 appwire-client/typescript/askDock.test.ts          |   7 +-
 appwire-client/typescript/deriveAskQuestions.test.ts | 104 +++-
 appwire-client/typescript/deriveAskQuestions.ts    |  78 ++-
 appwire-client/typescript/index.ts                 |   4 +-
 appwire-client/typescript/reducer.test.ts          | 679 ++++++++++++++++++++-
 appwire-client/typescript/reducer.ts               | 161 ++++-
 appwire-client/typescript/scripts/qualify-package.mjs |   6 +-
 appwire-client/typescript/transcriptProjector.test.ts |  12 +
 appwire-client/typescript/transcriptProjector.ts   |   8 +-
 cmd/evener-hub/frontend/src/dev/surface-sections/composer.tsx |   4 +
 cmd/evener-hub/frontend/src/panes/session/Session.test.tsx |  13 +
 cmd/evener-hub/frontend/src/panes/session/composer/Composer.integration.test.tsx |  10 +
 cmd/evener-hub/frontend/src/panes/session/composer/Composer.test.tsx |  14 +-
 cmd/evener-hub/frontend/src/panes/session/composer/askDock/AskDock.test.tsx |   9 +
 cmd/evener-hub/frontend/src/panes/session/composer/askDock/askDockStore.test.ts |  22 +-
 cmd/evener-hub/frontend/src/panes/session/transcript/messages/WarningItem.test.tsx |  39 ++
 cmd/evener-hub/frontend/src/panes/session/transcript/messages/WarningItem.tsx |  13 +-
 mobile-native/src/liveImages.test.ts               | 138 +++--
 mobile/src/conversation/project.test.ts            | 122 +++-
 mobile/src/conversation/project.ts                 |  37 +-
 mobile/src/state/conversation.test.ts              | 207 ++++++-
 mobile/src/state/conversation.ts                   |  82 ++-
 22 files changed, 1678 insertions(+), 91 deletions(-)
```

obra:
## Round 6 disposition (3af1912) — one Medium from B's #1732 panel; scope trimmed for lane retirement

**Medium (B's panel at 784126946) — reducer.ts:1270, Object.entries(value) still allocates the whole object before the 50-key cap applies**
Fixed: replaced `Object.entries(value).slice(0, N)` with a `for...in` + `Object.hasOwn` loop that counts and breaks at the key/node budget, never accessing (or pairing) a property value past the cap. `Object.entries`/`Object.keys` both allocate one entry per own key up front regardless of a later `.slice`, which was the exact O(frame-size) cost this function exists to avoid, just moved one line earlier than the cap.
Test: "never accesses more than the key cap's worth of property values" — a Proxy observes every `get` trap fire on a 100,000-key object, independent of enumeration mechanism. Falsified against the `Object.entries` form (100,000 gets observed vs. 51 with the fix).

**Scope trim for lane retirement**: per the coordinator, #1862's round-5 panel (99bde6efb, mirrored at reviews/raw/1862-99bde6efb.md) has a High and three Mediums beyond this Object.entries item — this PR is past five rounds and will be decomposed into smaller PRs by a fresh lane after this one retires. Those findings are NOT fixed in this round; they're recorded verbatim in the coordinator's state note (`.superpowers/sdd/2026-09-12-mobile-landing-queue-cont/a-lane-state-note.md`) under "#1862 open findings" for the decomposition lane to inherit, along with a proposed mechanism split (askPending gate + human-note exclusion / prunedForStringify bounded walk / warning string bounds / mobile canonical warning composition + ids) and the files each touches.

Non-test production diff (this round): 21 lines (reducer.ts). Test diff: 38 lines.

Gates: vitest (reducer.test.ts 217, deriveAskQuestions.test.ts) green; `npm run typecheck` clean; `npx biome ci` / `make lint-package-imports` clean; full mobile-native suite (800) and shared mobile/src (785) green.

Head: 3af1921280bc4aa0ced610a67467091a8e949286 (rebased onto #1859's new round-6 head; own-diff lines verified byte-identical before/after the rebase).

This is my last round on this lane. The state note has the full open-findings list, the mechanism split, and a gates cheat-sheet for whoever picks this up.

obra:
Decomposed after five rounds into #1892, #1893, #1894, #1895; closing.

PR #1867, opened 2026-09-18
fix(sdk): serialize saveDraft through the write queue (#1801)

### LINKED ISSUE #1801: keybindingsStore: saveDraft bypasses writeQueue, can race patchOverrides
`saveDraft` (`appwire-client/typescript/keybindingsStore.ts`) never chains through `writeQueue`, the same serialization `patchOverrides` uses to queue behind the previous write's full settlement. Flagged in #1741's round-2 review (`keybindingsStore.ts:1261`): a direct write (`patchOverrides`) can supersede an in-flight or queued checkpointed save (or vice versa), whose reply is then fenced and may return stale state even though the hub applied it; a queued direct write composed while `saving` was false can be rejected later once `saving` becomes true.

Measured: this is pre-existing on `main` (`origin/main` at the time of measurement, 0af7b3f32) - `saveDraft` has never referenced `writeQueue`, before or after the D6 keybindings-store split (#1728, #1729, #1736, #1739, #1741). Not introduced by this stack.

(Note: #1739's round-2 review separately found and this stack fixed the one-directional half of this class of bug - `patchOverrides` starting while `saveDraft` owns the write token, stranding `saving` true forever - by adding `state.saving || state.writeUncertain` to `patchOverrides`'s gate. That fix prevents a direct write from STARTING during a checkpointed save; it does not serialize the two paths through one queue, so the stale-reply and queued-rejection races #1741's review describes are still open.)

Fix per the review: serialize `saveDraft` with direct writes through the same `writeQueue` (or track an active direct write and refuse/queue the draft save), revalidate the draft and revision when the queued save executes, and add a test covering a direct PATCH followed by a draft save.

obra:
Fixes #1801

## Root cause

`saveDraft` (`appwire-client/typescript/keybindingsStore.ts`) issued its checkpointed PATCH directly, while `patchOverrides` serialized its direct PATCH through `writeQueue`. Both paths claim a write token from the same ready-generation fence, so with no shared queue:

- A direct write could start while a checkpointed save was in flight (or queued) and claim a newer token, fencing the save's reply — the save applied on the hub but the store returned stale state (and vice versa).
- A direct write composed while `saving` was false could be sent after a save had set `saving` true, colliding with the checkpointed write.

## Change

`saveDraft` is split into a synchronous prologue (call-time gate, durable checkpoint, `saving` block — unchanged UX and durability) and `runQueuedSave`, which performs the request and the existing post-reply sequence **queued through the same `writeQueue` as `patchOverrides`**. The queued half revalidates the intent against the state the previous write settled: if the generation ended, support dropped, the state is no longer writable, or the confirmed revision moved under the draft, nothing leaves the device; the checkpoint is settled and the draft is kept with `draftConflict` for review.

While there, the duplicated queue-chaining idiom is extracted to `enqueueWrite<T>(run)` and the fenced-reply payload construction to `currentPayload(state)`, so the two write paths cannot drift.

## Test evidence

New focused tests in `appwire-client/typescript/keybindingsStore.test.ts` ("the two write paths serialize through one queue"), both of which fail on the pre-fix source for the exact race:

- a direct PATCH in flight is not superseded by a draft save queued behind it (pre-fix: a second PATCH left while the first was on the wire)
- a draft save in flight is not superseded by a direct PATCH queued behind it (pre-fix: the save's reply was fenced and returned revision 3 with revision 4 applied)

Commands (run from `cmd/evener-hub/frontend`, `node_modules` symlinked to the main checkout):

```
npx vitest run ../../../appwire-client/typescript/keybindingsStore.test.ts   # 59 passed
npx vitest run src/stores/keybindings.test.ts src/stores/keybindings.initialWiring.test.ts src/panes/settings/sections/keybindings.test.tsx  # 122 passed
npx tsc --noEmit --incremental false                                          # exit 0
npx biome ci ../../../appwire-client/typescript/keybindingsStore.ts ../../../appwire-client/typescript/keybindingsStore.test.ts  # exit 0
```


obra:
## Re: finding "`saveDraft` can dispatch while `hubLoading` is true"

This one does not hold against the store's established (and CI-enforced) contract, so I am refuting it rather than changing behavior.

**Reproduction.** Adding `state.hubLoading` to the queued-save gate in `runQueuedSave`:

```
      state.storageUnavailable ||
      state.writeUncertain ||
      state.hubSupport !== "supported" ||
      !state.loaded ||
+     state.hubLoading ||
      state.loadError !== null
```

then, from `mobile-native`:

```
NODE_DISABLE_COMPILE_CACHE=1 npx vitest run src/nativePreferences.test.ts
```

produces:

```
× keybindings: an older GET cannot replace an acknowledged write 14ms
Error: Hub keybindings settings are unavailable.
    ❯ runQueuedSave ../appwire-client/typescript/keybindingsStore.ts
Tests  1 failed | 24 passed (25)
```

`mobile-native/src/nativePreferences.test.ts:247` ("an older GET cannot replace an acknowledged write") deliberately starts a GET refresh, then calls `saveKeybindings()` **while that refresh is in flight**, and asserts the acknowledged PATCH (revision 4) survives the older GET (revision 3). That is exactly the `hubLoading === true` save the finding would reject. Adding the guard to `assertEditable()` as well would fail the same test at call time, plus the `editDraft` path.

The design intent is that a checkpointed save may race an in-flight read; the read/apply path already protects the outcome — `applyHubOverrides`' stale guard drops an older payload, and `runQueuedSave`'s `newerExternal` comparison keeps a newer external revision for review. The only way to make the save reject on `hubLoading` is to remove that covered, deliberate contract, which is out of scope for #1801.

obra:
## Re: `hubLoading` gate on the `saveDraft` paths (finding addressed on the queued path, refuted for the call-time gate)

**Queued path — fixed.** `runQueuedSave` now refuses when the save had to *wait* in the queue and finds `state.hubLoading` true (test: "a save queued behind a write does not dispatch while a refresh is in flight" — a pending GET plus a preceding write that settles without a revision change; no second PATCH fires). This is the scenario the finding named.

**Call-time gate — deliberately not gated, and cannot be without breaking a pre-existing contract test.** `saveDraft`'s call-time `assertEditable()` must not check `hubLoading`. Proof: an immediate (non-waited) save running under `hubLoading` is required by `mobile-native/src/nativePreferences.test.ts:247` ("an older GET cannot replace an acknowledged write"), which starts a GET, calls `saveKeybindings()` while it is in flight, and asserts the acknowledged PATCH survives the older GET. Adding `state.hubLoading` to the gate makes it fail:

```
$ NODE_DISABLE_COMPILE_CACHE=1 npx vitest run src/nativePreferences.test.ts   # (from mobile-native)
× keybindings: an older GET cannot replace an acknowledged write 14ms
Error: Hub keybindings settings are unavailable.
    ❯ runQueuedSave ../appwire-client/typescript/keybindingsStore.ts
Tests  1 failed | 24 passed (25)
```

So the gate is applied exactly where it is safe and meaningful: a save that waited behind another write (whose composed revision may lag a refresh that started meanwhile) refuses; a save dispatched immediately keeps the covered native behavior, where the stale guard / `newerExternal` logic protects the outcome.

obra:
## Re: `releaseHeadTail` overwritten by a stale-epoch write (guard applied; reachable hang refuted)

**Guard applied.** `start` now only claims the head while `epoch === writeEpoch`, the symmetry the finding asks for (`settle` already guards on `releaseHeadTail === ownTail`).

**But the described reachable hang could not be reproduced, and I believe it is unreachable through the public API.** The interleaving needs a *stale-epoch* queued write's `start` to run *after* a *current-epoch* write's `start`. In every path that bumps `writeEpoch`, only `reset()` does so, and `reset()` calls `endReadyGeneration()` first, which resets the store to the unloaded initial state (`loaded: false`, `hubSupport: "unknown"`, `fence` ended). Consequences:

1. `reset()` releases the abandoned head's tail; the queued stale writes' `start` callbacks are then the very next microtasks.
2. A current-epoch write can only become a valid head after the store is editable again, which requires an `await` on a refresh (`loaded` is false until then).
3. That `await` yields, so every stale write's `start` runs (and its `run` rejects synchronously on `!fence.liveHub(callGeneration)` / `!stillMine()`) before any current-epoch write starts.

So the "new write D starts immediately (pendingWrites === 0), sets `releaseHeadTail = D_tail`; the stale write B then overwrites it" ordering cannot arise: D cannot be a valid head in the same synchronous turn the epoch was bumped, and by the time D is valid, all stale writes have already started and settled. The two writes that the guard genuinely protects against overlap are the same epoch, where overwriting the head with the newer write is correct.

I could not produce a failing test for the specific hang; the applied guard is a one-line, behavior-preserving hardening (it changes nothing in any reachable ordering, since stale writes settle immediately on their fence). If you can share a concrete sequence reaching the overlap, I will add the failing test.

obra:
## Re: hub-replacement / queue-retirement verdict (`631416e`)

**High — fixed with a retirement serial.** `retirePayload()` now bumps a `payloadSerial` (support drop, detach, generation end). Both write paths capture it at call time and refuse to dispatch if it changed, so a queued write composed against a retired payload can never send to a replacement hub that happens to serve the same revision. The serial is checked *in addition to* `callGeneration`, because `detachHub()` and a support drop retire the payload without ending the generation.

Tests (both new, both fail before and pass after):
- `a save queued across detachHub does not dispatch to the replacement hub` — the replacement refreshes with the **same** revision, then the preceding write settles; no second PATCH fires.
- `a direct PATCH queued across detachHub does not dispatch to the replacement hub` — same for the direct path.

**Medium — fixed for generation end and `detachHub`; refuted for the support-drop case.**

Fixed: `endReadyGeneration()` and `detachHub()` both release the abandoned queue head, so a never-settling in-flight request cannot block queued writes (`detachHub drains a write queued behind a never-settling request`; fails before with a 5s hang).

Refuted for support drop: releasing the queue when support drops breaks a pre-existing, deliberate pane test. Reproduction — add `releaseAbandonedHead()` to `retirePayload()` (which is what a support drop calls):

```
$ cd cmd/evener-hub/frontend && npx vitest run src/panes/settings/sections/keybindings.test.tsx
× a generation-fenced queued write's row error clears on the next confirmed payload 1115ms
TestingLibraryElementError: Unable to find role="alert"
Tests  1 failed | 53 passed (54)
```

That test (`keybindings.test.tsx:1274`) intentionally leaves the queued write parked across a support flap: it must surface its `unavailable` row error only when the in-flight write it queued behind settles in the NEW generation. Releasing on the support drop runs it during the unsupported window, and the flap-back's apply-serial bump then clears the row error — the alert is gone. So support drop is a recoverable flap that fences queued writes (via `payloadSerial`) but does not force them to run; generation end and hub detach are terminal for the old hub and do drain. With the serial in place, a queued write that survives a support drop still cannot reach the replacement hub.

obra:
Closing as superseded — not landed.

Issue #1801 is fixed on `main` by #1808 (merge commit `8d8ed0d2740e271dbc13f99682d6b6da77472c97`), and #1801 is closed. That fix takes the issue's own second stated alternative — "track an active direct write and refuse/queue the draft save" — rather than this PR's approach of serializing both paths through one queue:

- `appwire-client/typescript/keybindingsStore.ts:648-663` declares `directWrite`, reserved synchronously at `patchOverrides` call time.
- `:1151-1157` takes the reservation before the queued turn can run; `:1247-1252` clears it by object identity so a stale call cannot clear a newer reservation.
- `:1393-1399` makes `saveDraft` refuse with `UNAVAILABLE_MESSAGE` while a reservation is live.

Those line claims were verified directly against `origin/main`, not taken from a report.

Rebasing this PR onto `main` produces a genuine design-vs-design conflict across roughly six hunks of `keybindingsStore.ts` (the lane that held it declined to resolve it unilaterally and asked for a decision — correctly, since either side would be discarding the other's mechanism). With the defect already prevented on `main`, there is nothing here left to land, so the branch is closed rather than forced through.

The lane's work was not wasted: it found the design conflict, refused to guess, and produced the verification that #1801 is actually fixed — which is how an open issue with a linkage-less fix got closed at all.

PR #1888, opened 2026-09-18
refactor(mobile): ConnectionProvider over the package connection store (D28 D1)

obra:
## What moved

`ConnectionProvider.tsx`'s hand-rolled hub-connection lifecycle (the second
`useEffect`: reset session/state/error, fetch the hub token, build the
client, subscribe to its `onStateChange`, and tear down on swap or unmount)
is extracted into `useHubConnection` (`mobile-native/src/hubConnection.ts`),
an adapter over `@evener/appwire-client/state/connection`'s
`createConnectionStore`. The store now owns the client-swap safety
(detaching the outgoing client's listener, the re-entrancy generation
counter) that the old code partially hand-rolled; `ConnectionProvider` keeps
everything else (`HubSelection`, saved-location restore, `AppState`
foreground tracking, hub profile CRUD) untouched.

## Ports

None added. Measured `state/connection/core.ts`: no timers, no
`window`/`document`, nothing native lacks — the whole mechanism is
`client.onStateChange` plus the store's own `getState`/`setState`/
`subscribe` triple, already wired to React via
`useSyncExternalStore(store.subscribe, store.getState)`, the same idiom
`PluginsScreen.tsx`/`MarketplaceBrowser.tsx` already use for
`createPluginsStore`. `useHubConnection` narrows its `repository` parameter
to a one-method `HubTokenSource` interface (`token(id): Promise<string>`)
rather than the full `HubProfiles` class — the only member it calls — so a
test can fake it without a real `SecureStore`-backed repository.

## Consumers unchanged

`Connection.client`, `Connection.state`, and `Connection.error` keep their
exact shapes and values for every screen that reads `useConnection()`
(`PluginsScreen.tsx`'s `!client || state !== "ready"` reconnect-wall check,
etc.) — nothing outside `ConnectionProvider.tsx` changed. `error` is still
the same shared slot the saved-hubs-load failure and the connection failure
both write into, unchanged from before.

## Oracle

No test mounted the real `ConnectionProvider` before this PR — every
existing test (`credentialStore.render.test.tsx`, `credentialStore.test.ts`,
`ProvidersScreen.test.tsx`) mocks `./ConnectionProvider` entirely and reads
only `{ client, state }`/`{ activeProfile, client, state, retry }` off it,
so they are the oracle that the exposed shape didn't change. All three stay
green, unchanged.

## New tests (`hubConnection.test.tsx`)

Faked the client `createHubClient` would otherwise build over a socket
(`FakeHubClient`: `state`, `terminalReason`, `onStateChange`, `connect()`
settled by `succeed()`/`fail()`, `close()`) and drove `useHubConnection`
directly:
- stays `{ client: null, state: "idle" }` with no active profile
- connects, follows the client into `"connecting"` → `"ready"` → `"closed"`,
  reporting the same failure message through `setError` `ConnectionProvider`
  always has (client stays set after close — a retry reuses it)
- a bumped `attempt` tears down the old client (listener count drops to 0,
  it closes) before wiring the new one
- unmounting releases the client's listener (this caught a real bug below)

## Falsification

Moved `hubConnection.ts` aside and ran `hubConnection.test.tsx`: import
failure (`Cannot find module './hubConnection'`), all 4 tests failed to
even collect. Restored the file; `git status --short` showed only the
intended files afterward.

## Bug found and fixed during TDD

The unmount test caught a real listener leak: `createConnectionStore`'s own
internal `onStateChange` listener (the swap-safety mirror) was never
detached on a true component unmount — only a *subsequent* effect run's own
reset cleared it, and unmounting has no subsequent run. Fixed by clearing
`store.setState({ client: null })` in the effect's cleanup (not just at the
top of the next run), which triggers the store's own detach path.

## Gates

- `npx vitest run` (mobile-native): 85 files / 805 tests passed (full
  suite; `hubConnection.test.tsx` 4/4)
- `npm run check` (mobile-native, tsc): clean
- `make lint-package-imports`: PASS (no `appwire-client` files touched by
  this PR)
- node_modules was a real directory (not a symlink) and `react-test-renderer`
  was missing; ran `npm ci` once in `mobile-native` before testing

## Diff

```
```

Non-test production additions: 114 lines (`hubConnection.ts` 104 new +
`ConnectionProvider.tsx` 10 insertions/69 deletions net -59). Test-only:
`hubConnection.test.tsx`, 160 lines.

Part of #1595


obra:
## Round 2 disposition (fbe46a0)

**Medium 1 (hubConnection.ts:95, stale client on a hub switch)** — confirmed
by measurement, fixed.

Wrote the failing hub-switch test first (`hubConnection.test.tsx`, "never
returns the previous hub's client once activeId already names a newer
one"): render with hub-a connected and ready, switch to hub-b, capture
*every* render (not just the settled `result.current`) paired with which
hub it was for. Against the round-1 code this failed —
`result.client === first` on the render that already carried `activeId
"hub-b"` — because `useSyncExternalStore` reads the store's live client
directly, and passive effects (which would clear it) run after commit, not
before the render that already has the new `activeId`.

Confirmed what `ConnectionProvider` did before the lift for this exact
case: `session?.profileId === selected`, gating the exposed client on
whether the SESSION's own recorded target still matches the CURRENTLY
selected id — not just clearing state in an effect. `useHubConnection` had
dropped this comparison (I'd judged it dead code during the original
measurement, incorrectly — it isn't dead against the *hook's* new
`activeId`/`state` split; it protects exactly the gap between a render and
its passive effect).

Fix: `connectedFor` (`useRef<string | undefined>`), written in the same
synchronous statement as `store.setState({ client: connection })` — never
on a later render — recording which `activeId` that exact client was
opened for. The client getter now reads `connectedFor.current ===
activeId` before trusting the store's client, so a render with a newer
`activeId` sees the mismatch immediately and reports `null`, converging on
the real client once its own effect resolves. This is a render-time
derivation, not a clear-in-an-effect — deriving "client for this target" is
exactly what the finding asked for.

**Medium 2 (hubConnection.test.tsx:99, exact-sentence assertion)** — fixed.
Replaced the hardcoded prose with `connectionFailure("protocol").message`
(imported from `./connectionRecovery`), asserting the *branch*
`connectionFailure` selects rather than pinning its copy, per
`docs/developing-evener/testing.md`'s "Prompt Prose Is Not a Test Oracle."

## Falsification

Diffed `hubConnection.ts`'s round-2 changes into a patch, `git checkout
HEAD -- mobile-native/src/hubConnection.ts` (reverting only that file to
the round-1 commit), ran the new hub-switch test: failed with the same
`not to be FakeHubClient{...}` the original measurement produced.
Reapplied the patch; test passed again. `git status --short` showed only
`mobile-native/src/hubConnection.ts` and `hubConnection.test.tsx`
throughout.

## Gates

- `npx vitest run` (mobile-native): 85 files / 806 tests passed (full
  suite; `hubConnection.test.tsx` 5/5)
- `npm run check` (mobile-native, tsc): clean
- `make lint-package-imports`: PASS (no `appwire-client` files touched)

## Diff

```
mobile-native/src/ConnectionProvider.tsx |  79 ++----------
mobile-native/src/hubConnection.test.tsx | 206 +++++++++++++++++++++++++++++++
mobile-native/src/hubConnection.ts       | 122 ++++++++++++++++++
3 files changed, 338 insertions(+), 69 deletions(-)
```

Head: fbe46a018184e2207878b46adb8b29ba683d772c

obra:
## Round 3 disposition (dbdc226)

Same class as round 2, not fully closed — confirmed by measurement, now
closed on all four inputs.

**Root cause**: round 2's `connectedFor` tracked only `activeId`, but the
effect opens a NEW connection whenever `activeId`, `activeOrigin`,
`foreground`, OR `attempt` changes. A retry of the same hub (attempt
bumped, `activeId`/`activeOrigin` unchanged) still passed the
`connectedFor.current === activeId` check, so the render before the
teardown effect runs could still expose the previous, now-stale client
(round 2's fix only ever closed the hub-switch instance of this class, not
the mechanism).

**Failing-first test**: same hub throughout, bump `attempt`, capture every
render (not just the settled value) paired with which attempt it was for.
Against round 2's code this failed identically to the round-2 measurement —
`result.client === first` on the render already carrying `attempt: 1` — for
the same underlying reason: `useSyncExternalStore` reads the store live,
and the passive effect that would clear it runs after commit, not before a
render whose own inputs already moved on.

**Fix**: `connectedFor` now records a `ConnectionTarget` — `activeId`,
`activeOrigin`, `foreground`, `attempt` — written in the same synchronous
statement as `store.setState({ client: connection })`. The client getter
requires `sameTarget(connectedFor.current, { activeId, activeOrigin,
foreground, attempt })` (all four fields, exact match) before trusting the
store, so ANY of the effect's four dependencies moving on invalidates the
stale client at render time, not just a hub switch.

`store` and `repository` are deliberately excluded from the key: `store` is
created once via `useState` and never changes for the hook's life;
`repository` is the caller's own stable reference (documented in
`HubTokenSource`'s comment) — neither can ever distinguish one generation
from another, so including them would add comparison cost with no
protective value.

## Falsification

Diffed round 3's `hubConnection.ts` changes into a patch, `git checkout
HEAD -- mobile-native/src/hubConnection.ts` (back to round 2's committed
code), ran the new attempt-bump test: failed with the same `not to be
FakeHubClient{...}` the measurement produced. Reapplied the patch; all 6
tests in the file passed. `git status --short` showed only
`hubConnection.ts` and `hubConnection.test.tsx` throughout.

## Gates

- `npx vitest run` (mobile-native): 85 files / 807 tests passed (full
  suite; `hubConnection.test.tsx` 6/6)
- `npm run check` (mobile-native, tsc): clean
- `make lint-package-imports`: PASS (no `appwire-client` files touched)

## Diff

```
mobile-native/src/ConnectionProvider.tsx |  79 ++--------
mobile-native/src/hubConnection.test.tsx | 244 +++++++++++++++++++++++++++++++
mobile-native/src/hubConnection.ts       | 149 +++++++++++++++++++
3 files changed, 403 insertions(+), 69 deletions(-)
```

Head: dbdc2265bd86fc842d625915717c35b056735bfb

PR #1889, opened 2026-09-18
plugins: store write failures never name the store path (helpers + write sites)

obra:
Decomposed from #1810 (closes part of #1700), split 1 of 3 after that PR reached its fifth review round.

Adds `saveFailed` and `storeChangeRollbackFailed`, which log the raw error (with its absolute plugin-store path) to the hub's stderr and return a scrubbed, caller-safe message. Applied at:
- `ensureFetched`'s backfill save
- `AddMarketplace`'s save and its clone rollback
- `RefreshMarketplace`'s save
- `EditMarketplace`'s non-rename save path and its `fail` closure's rollback-also-failed branch
- `saveRename`, which also covers migration-triggered renames reached through `ListMarketplaces`/`List` (not just `EditMarketplace`)

`RemoveMarketplace` is untouched here — its reorder and clone-cleanup-failure handling is split 2. The `EditMarketplace` fail path when `moveMarketplace`/`swapInClone` themselves fail (raw path leak when rollback succeeds) and the migration-recovery join paths are open Mediums from #1810's round 5, going to split 3.

One existing test (`TestMarketplaceNameMigration_ARecoveryThatRollsBackRemovesTheMarker`) is updated to assert the scrubbed message names `known_marketplaces.json` instead of the raw "boom" cause, since `saveRename`'s scrub changes that message's text.

Issue #1879, opened 2026-09-18
flaky: session_retirement_evidence tests fail with 'settled owner not eligible: {Phase:resident' in the CI 'tests' job on TypeScript-only PRs


obra:
CI run 35369662679 (PR #1856 @67d8f9022, which touches only appwire-client/typescript/state/mutation/{projection.ts,projection.test.ts,index.ts}) failed the root 'tests' job with:

    --- FAIL: TestRetirementAttentionArmRetryPending (0.06s)
        session_retirement_evidence_test.go:567: settled owner not eligible: {Phase:resident Timeout:0 ...

The PR changes no Go, so the failure is not attributable to it (touched paths checked per the flake-attribution rule). Filed as a flake candidate; the coordinator will append other occurrences.

obra:
Second occurrence, same family: CI run 35371121143 (PR #1792 @a91612a12, TypeScript-only) failed the root 'tests' job with

    --- FAIL: TestRetirementSafetyColdJobWatchContent (0.15s)
        session_retirement_evidence_test.go:3025: settled owner not eligible: {Phase:resident Timeout: ...

Same assertion shape ('settled owner not eligible: {Phase:resident ...}') as TestRetirementAttentionArmRetryPending at :567 in the first occurrence. Both on runs during a period with dozens of concurrent CI jobs; likely timing-sensitive retention evidence tests.

obra:
Third occurrence: PR #1841 @311721ca2 (TypeScript-only), CI 'tests' job: --- FAIL: TestRetirementSafetyDetachedLifetime (0.10s), same session_retirement_evidence family.

obra:
Fourth occurrence, 2026-09-18 19:40Z: run 35385964042 ('tests' job) on #1895 @e56da8680, a TypeScript-only PR (mobile/src), failed `TestRetirementTreeSettleDrainsPendingRootAttention` in package agent. Same retention/retirement family; the PR touches no Go.

obra:
Fifth occurrence, 2026-09-18 20:10Z: run 35388856906 ('tests' job) on #1841 @13c8e8353 (TypeScript-only D6 piece) failed `TestRetirementAttentionArmRetryPending` in package agent.

obra:
[...]

PR #1890, opened 2026-09-18
plugins: RemoveMarketplace saves before it deletes the clone

obra:
Superseded by #1940 (server applied-with-litter outcome) and #1954 (revision- and generation-fenced SDK reconciliation). The original branch remains preserved. Web/native/TUI presentation and retry handling are required consumer follow-ups before that stack merges; #1897 remains the separate server migration successor.

---

Decomposed from #1810 (closes part of #1700), split 2 of 3. Stacked on #1889; diff against main includes it until it merges.

Ruling 2026-09-18: keep the RemoveMarketplace behaviour change (save metadata before deleting the clone; a clone-cleanup failure after a successful save is returned to the caller, never swallowed as a warning with a success reply), with its rewritten tests.

`RemoveMarketplace` now saves the metadata first — a save failure is a plain refusal that leaves the marketplace registered and its clone untouched — and only then removes the clone. A clone-removal failure is reported via the new `cloneRemovalFailed` helper (named for the domain outcome per #1810's round-5 Low, not the mechanism) instead of swallowed as a warning while returning success. Rewrites `coverage_marketplaces_fuzz_test.go`'s clone-removal-failure case to expect the error, and adds `TestRemoveMarketplaceSaveFailureNamesNoPath` / `TestRemoveMarketplaceCloneRemovalFailureNamesNoPath`.


obra:
## Round 1 fixes (076cd9b09)

- **marketplaces.go:332** (member 1, Medium): `cloneRemovalFailed` now wraps `ErrMarketplaceUnregisteredCloneRemains` instead of a bare-text error, and a lookup miss in `RemoveMarketplace` (`notFoundOrCloneRemains`) checks for a leftover clone before reporting `ErrMarketplaceNotFound` — present, it retries the removal and reports the same typed error while it keeps failing, resolving the litter silently once it succeeds. Failing-first tests: `TestRemoveMarketplaceCloneRemovalFailureNamesNoPath` (strengthened with an `errors.Is` check) and new `TestRemoveMarketplaceRetryAfterCloneRemovalFailureReportsLitterNotNotFound` (asserts the retry reports the typed litter error, not a bare not-found, then that a working retry finishes the cleanup).
- **marketplace_migration.go:999-1004** (member 0, Medium) and **marketplaces.go:459** (member 1, Medium): same findings as split 3 (#1897); #1897 covers the join-path scrub, and the `EditMarketplace` fail-path finding was investigated there and filed as #1896 (conflicts with two existing tests that intentionally keep raw paths for a different failure class).
- pi's unscrubbed-sites list: `writeRenameMarker`/`saveMigrationRecord` going to #1897 if same-mechanism and small; `install.go`'s `saveRegistry` sites already in #1896.

Gates: `gofmt -l` clean, `go vet ./...` clean, `go vet -tags evenerfuzz ./...` clean, `GOOS=windows go vet -tags evenerfuzz ./...` clean, `golangci-lint run ./internal/plugins/...` 0 issues, `go test -count=1 ./internal/plugins/...` and `-tags evenerfuzz` both green.

obra:
## Round 2 fixes (d5af6cefe)

**High (member's finding on the retry-cleanup branch): deleted it.** `notFoundOrCloneRemains` derived `m.marketplaceDir(name)` from an unvalidated, caller-controlled `name` on a lookup miss and called `pathPresent`/`marketplaceRemoveAll` on it. Measured against the actual vulnerable code: `RemoveMarketplace("")` returned `nil` and silently deleted the whole marketplaces directory (`""` resolves to `m.marketplaceDir("")` = the marketplaces dir itself); `".."` resolves to its parent. A lookup miss now returns the plain `ErrMarketplaceNotFound` with no filesystem access whatsoever, exactly as before round 1. `ErrMarketplaceUnregisteredCloneRemains` stays, returned only where the clone-removal cleanup that follows a **successful** unregister save actually fails (an already-validated, already-registered name at that point — no attacker-controlled path derivation). Whoever wants automatic litter cleanup files a follow-up issue for a safe design; this round is fix-the-vulnerability, not rebuild-the-feature.

Regression test: `TestRemoveMarketplace_LookupMissTouchesNothingOnDisk` — `RemoveMarketplace` with `""`, `".."`, `"../x"`, and an ordinary unknown name all return `ErrMarketplaceNotFound` and call `marketplaceRemoveAll` zero times; the marketplaces directory and an existing marketplace's clone are both left untouched. Falsified against the vulnerable code first (confirmed both new/updated tests fail red — `RemoveMarketplace("")` returns `nil` there, not even an error).

**CI 'tests' job red at a3bf954ec**: `gh run view --log-failed` on the run for that head shows the only failure is `TestRetirementSafetyDetachedLifetime` in the `agent` module ("agent-shards: the survey pass failed — the suite is red"). #1890's touched paths are entirely `internal/plugins/*`; no `agent/` file is in this PR's diff. This matches #1879 (the tracked `session_retirement_evidence`/retention flake family) — pre-existing CI flakiness, not a real failure from this PR. No fix needed here.

**Migration-record/rename-marker Medium**: confirmed covered — #1897 (S3)'s `migrationFileSaveFailed` scrubs `writeRenameMarker` (marketplace_migration.go, now :451) and `saveMigrationRecord` (now :571), the exact two functions at the cited lines.

Rebased onto #1889's new tip (round 3, boundary-scrubbed saveMarketplaces).

Gates: `gofmt -l` clean, `go vet ./...` clean, `go vet -tags evenerfuzz ./...` clean, `GOOS=windows go vet -tags evenerfuzz ./...` clean, `golangci-lint run ./internal/plugins/...` 0 issues, `go test -count=1 ./internal/plugins/...` and `-tags evenerfuzz` both green.

obra:
## Round 3 disposition (d5af6cefe, no new code)

**Medium refuted** (member 0: `install.go:57-58`/`marketplaces.go:104-105` log the raw path to `m.stderr()`, which defaults to `os.Stderr`, reaching a local CLI user's terminal). Read #1700's actual text:

> ## Where
> `internal/plugins/marketplaces.go`, `AddMarketplace`'s save-failure path that does not roll back.
> ## What
> #1602 round 8 (`1744e3be9`) stopped three rollback branches from leaking the store's absolute path into **wire errors** via the `rollbackFailed` helper. `AddMarketplace`'s non-rollback save-failure branch was not in the finding's scope and has not been audited for the same `*fs.PathError` leak.
> ## Proposed change
> Audit that branch with a test that uses a real `*fs.PathError` ... and route it through `rollbackFailed` or the equivalent path-free wrapping if it leaks.

Every sentence names "wire errors" / the RPC-error path (#1602's own scope, which this issue extends) as the thing being closed; it never mentions the daemon's own stderr/log output. The daemon's stderr is the "hub's log" `saveFailed`'s own message explicitly points the caller at ("see the hub's log for detail") — it is the intended, documented destination for the raw diagnostic, read by whoever operates the daemon (which for a local CLI invocation is the same user, seeing their own already-known store path in their own terminal — not a disclosure to anyone who didn't already have filesystem access). #1700's boundary is the wire, not stderr. No change made; flag if this reading is wrong.

**Low noted, not fixed this round** (member 2: `marketplace_migration_test.go:2694` `TestMarketplaceNameMigration_ARecoveryThatRollsBackRemovesTheMarker` lost its `"boom"`-cause discriminator — `saveFailed` injects `marketplacesFileName` into its message unconditionally regardless of the cause, so the current assertion is trivially true by construction and a future regression that drops the real cause would go undetected). Confirmed correct by reading `saveFailed`: the returned error's text never contains the raw cause (only `m.stderr()` does, and `errStoreBetweenNames`'s identity survives `%w`, but a plain `errors.New("boom")` leaves no trace in the returned error at all). Fix per the suggestion (point `m.Stderr` at a `bytes.Buffer`, assert it contains `"boom"`) is deferred: bundled into #1889 (S1) if that PR pushes again for the simplify review, otherwise into the follow-up PR after the stack merges. Not fixed speculatively now.

Standing by for the opus /simplify verdict on #1889/#1890/#1897.

obra:
## Rebased onto squash-merged main (0f0588525)

#1889 (S1) squash-merged to main as \`a81740a36\`. Rebased S2's own three commits (\`git rebase --onto origin/main 0655494cb...\`, the pre-squash S1 tip S2 was built on) onto the new main — own lines identical, no conflicts. All six opus /simplify findings on this PR (1890-simplify.md) are follow-up only (0 must-fix), so no code change this round.

Gates: \`gofmt -l\` clean, \`go vet ./...\` clean, \`go vet -tags evenerfuzz ./...\` clean, \`GOOS=windows go vet -tags evenerfuzz ./...\` clean, \`golangci-lint run ./internal/plugins/...\` 0 issues, \`go test -count=1 ./internal/plugins/...\` and \`-tags evenerfuzz\` both green.

Follow-up items from this PR's /simplify (the exported-but-unconsumed sentinel, the duplicated retry test, the four-copies-of-one-fixture pattern, the three-times-repeated \`""\`/\`".."\` rationale) join the single post-merge follow-up PR per the coordinator's plan.

obra:
## Round 4 disposition (26138e7)

Panel: `1890-0f0588525.md`. Two findings, both fixed.

**Medium** (`cmd/evener-hub/app_plugins.go:266`, `cmd/evener/plugincmd.go:142`) —
`ErrMarketplaceUnregisteredCloneRemains`'s applied-with-litter outcome fell
through the same path as a plain refusal at both boundaries, so a caller read
a successful removal as failed and a retry then found `ErrMarketplaceNotFound`
instead of ever learning about the litter.

- Measured the hub's existing shape for "applied, then a follow-up durable
  step failed": `ErrorKeybindingsPostRename`/`KeybindingsPostRenameData` and
  `ErrorTranscriptDisplayPostApply`/`TranscriptDisplayPostApplyData`
  (`appwire/errors.go`, `appwire/keybindings.go`, `appwire/transcript_display.go`,
  wired in `cmd/evener-hub/app_rpc_keybindings.go`) — both build a `WireError`
  inline with `Code: CodeInternalError` and a `Data` payload carrying an
  `Applied` field with the canonical post-apply state, rather than routing
  through a refusal classifier. Reused that shape: new
  `ErrorMarketplaceUnregisteredCloneRemains` + `MarketplaceUnregisteredCloneRemainsData{EvenerErrorInfo, Applied MarketplaceListResponse}`.
  The updated marketplace list rides in `Data.Applied` — JSON-RPC's
  result/error union means it can't also ride as the call's `result`, so this
  is the only place a refusal-shaped response can carry it.
- `RemoveMarketplace` now checks `errors.Is(err, plugins.ErrMarketplaceUnregisteredCloneRemains)`
  before falling through to `marketplaceRefusalToWire`, and returns the typed
  `WireError` with the freshly re-listed marketplaces in `Data.Applied`.
- CLI (`plugincmd.go`'s `remove` case) reports
  `removed marketplace %q; its clone files could not be removed and remain on
  disk; no automated cleanup exists yet for marketplace clones, remove them by
  hand: %w` and still returns non-zero. Checked for an existing gc/clean
  command first: `evener plugin gc` (`internal/plugins/gc.go`) only sweeps
  `cacheDir()` (materialized plugin dirs), never `marketplacesDir()`/a
  marketplace's own clone — so there's nothing to name; said so rather than
  inventing a command. `evener plugin doctor` also only iterates *registered*
  marketplaces, so it can't see this litter either (the entry is already
  gone). Flagging as a residual/GitHub issue below.
- Falsified: reverted each production fix in turn (`app_plugins.go`'s
  RemoveMarketplace branch, then `plugincmd.go`'s remove-case branch) with the
  new test/type additions left in place; the hub test failed to compile
  (`undefined: appwire.MarketplaceUnregisteredCloneRemainsData` /
  `ErrorMarketplaceUnregisteredCloneRemains`), and the CLI test failed at
  runtime (`err = marketplace "acme": marketplace unregistered, but its clone
  could not be removed; see the hub's log for detail, want it to say the
  registry removal landed`). Restored both; full suite green again.

**Low** (`internal/plugins/marketplaces.go:303`) — `cloneRemovalFailed`'s
comment claimed `RemoveMarketplace` has one caller; reworded to say the save
has already landed by the time this cleanup runs, without the caller-count
claim.

### Gates (from repo root)
- `$(go env GOROOT)/bin/gofmt -l` on touched files: clean
- `go vet ./internal/plugins/... ./cmd/evener-hub/... ./cmd/evener/... ./appwire/...`: clean
- `go vet -tags evenerfuzz` (plain + `GOOS=windows`): clean
- `golangci-lint run` on the same four packages: 0 issues
- `go test -count=1` (plain and `-tags evenerfuzz`) on the same four packages: all green

### Residuals (for a filed issue, not fixed here)
- No command sweeps a marketplace's own leftover clone directory
  (`<Root>/marketplaces/<name>`) after a failed `RemoveMarketplace` cleanup:
  `evener plugin gc` only sweeps the plugin cache, and `evener plugin doctor`
  only reports on registered marketplaces. Worth either extending `Gc` to
  also sweep unreferenced `marketplacesDir()` entries, or adding a `Doctor`
  check for a clone directory with no matching registry entry.

Both S2 (`26138e7414979bc7aa344cc6a1b3e70be4233af2`) and S3
(`14f8c3a5f119a7e36fb12b046de4291e85ea649e`, rebased onto this tip) are
pushed.

obra:
## Round 5 disposition (bda74b9)

Panel: `1890-26138e741.md`. Two Mediums, both fixed.

**Medium** (`cmd/evener-hub/app_plugins.go:277-280`) — if the re-list
`RemoveMarketplace`'s litter path runs to build `Data.Applied` fails, the
handler discarded the whole typed `WireError` and returned the bare listing
error, so a caller could no longer tell "removed with litter" from an
ordinary failure and might retry into `ErrMarketplaceNotFound`.

- Measured the existing post-rename/post-apply shape again
  (`ErrorKeybindingsPostRename`/`ErrorTranscriptDisplayPostApply`): neither
  precedent's own "Applied" snapshot can fail (both are in-memory reads), so
  there's no existing shape for "the reconciliation read itself failed."
  Added `MarketplaceUnregisteredCloneRemainsData.AppliedUnavailable bool`
  (`json:"appliedUnavailable,omitempty"`) — false/absent in the normal case,
  true with `Applied` left at its zero value when the re-list fails.
- `RemoveMarketplace` now always returns the typed `WireError`; the re-list
  failure only flips `AppliedUnavailable`, never drops the outcome.
- To reach that branch with a real failure rather than a stub: added
  `hubPluginsReconcileAfterCloneLitter`, a no-op hook run immediately before
  the re-list, matching `credentialWriteBetween`'s and the navigation
  service's existing between-step seams in this same package. The new test
  points it at a real `os.Chmod(..., 0o000)` on `known_marketplaces.json`
  right after `RemoveMarketplace`'s own successful unregister-then-failed-
  clone-cleanup, so only the reconcile read breaks.
- Falsified: reverted `app_plugins.go`'s production change with the test and
  the `appwire` type left in place — compile failure (`undefined:
  hubPluginsReconcileAfterCloneLitter`). Restored; full suite green.

**Medium** (`appwire-client/typescript/state/extensions/marketplaces.ts:186-197`)
— the marketplaces store never consumed `Data.Applied`, so a cleanup failure
left the removed marketplace and its browse cache in client state until the
separate `evener/marketplace/updated` notification/refetch caught up.

- `removeMarketplace` now catches its own rejection, extracts
  `data.applied.marketplaces` via a new `cloneLitterApplied` helper (guarded
  on `evenerErrorInfo === "marketplaceUnregisteredCloneRemains"` and
  `!data.appliedUnavailable`, the same discriminator style as
  `keybindingsStore.ts`'s `rejectionPayload`), applies it and retires the
  browse cache entry for that name, then rethrows so the UI still reports the
  leftover clone.
- Wire shape table for the fake client fixtures (per appwire/types.go's json
  tags): `applied.marketplaces` has no `omitempty` (`c.listMarketplaces`
  always returns a `make([]...,0,...)` slice) so it's `[]`/populated, never
  absent, on a successful re-list; on `appliedUnavailable`, `Applied` is the
  Go zero value so `marketplaces` is `null` (a nil slice), and
  `appliedUnavailable: true` is present (its own field has `omitempty`, so
  it's absent entirely in the normal case).
- Falsified: reverted `marketplaces.ts`'s production change with the test
  left in place — the reconcile test failed (`marketplaces` still `[ACME,
  LOCAL]` instead of `[LOCAL]`); the two guard-rail tests (appliedUnavailable,
  ordinary failure) still passed, as expected since they assert nothing
  changes. Restored; all 23 tests in the file green.

### Gates
- `$(go env GOROOT)/bin/gofmt -l`, `go vet` (plain/evenerfuzz/windows),
  `golangci-lint run` on `cmd/evener-hub` and `appwire`: clean
- `go test -count=1` (plain and `-tags evenerfuzz`) on `cmd/evener-hub` and
  `appwire`: all green
- `npx vitest run` on the touched store test (23/23), `npm run typecheck`,
  `npx biome ci ../../../appwire-client/typescript` (from
  `cmd/evener-hub/frontend`; biome's auto-formatter fixed quote style and one
  line wrap - no logic changes), `npm run lint`: all clean

### Size
Non-test insertions: `appwire/types.go` +12, `cmd/evener-hub/app_plugins.go`
+31 (net, incl. the seam), `appwire-client/typescript/.../marketplaces.ts`
+34 = 77 lines, well under the 400 ceiling. Test insertions: +83 (hub),
+58 (store) = 141.

Both S2 (`bda74b98413ae09c43a8c2e6789f8ba60c3c77c8`) and S3
(`d3c099449e1d13aeaaf9c18b45d07aa9abee4ee8`, rebased onto this tip, clean -
no conflicts) are pushed.

PR #1892, opened 2026-09-18
fix(sdk): ask resolution is wire-authoritative (askPending gate, human-note excluded)

obra:
Stacked on #1859; diff against main includes it until it merges.

PRs per the coordinator's split. This piece: `liveAskQuestions` reads
`ThreadModel.askPending` directly instead of re-deriving whether anything is
pending from transcript shape, and the resolution boundary recognizes an
interrupt or a user-sourced steer, not only a plain user message. A
human-note steer is explicitly excluded — writing a note does not answer
the question.

Files: `deriveAskQuestions.ts` + its tests, and the fixture updates (web
composer/askDock tests, the mobile `project.test.ts` fixtures) that stamp
`askPending` so existing tests keep exercising the same scenarios under the
new wire-authoritative gate.

No open round-5 findings apply to this mechanism.


obra:

obra:
## Round 1 disposition (34146da)

Both Medium findings (deriveAskQuestions.ts:109-110: an old daemon that omits the additive `askPending` field decodes it as false, so its unanswered ask is hidden instead of being re-derived from the transcript) are refuted by ruling, not by code: Jesse ruled on 2026-09-18 in #1823 ("No fallback") that the dock gates on the wire's `askPending` only and that daemons predating the field render no pending ask until they are updated. Re-deriving from transcript shape is exactly the client-side guessing this series removes. No change.

PR #1893, opened 2026-09-18
fix(sdk): prunedForStringify is a bounded lazy walk

obra:
Stacked on #1892; diff against main includes it (and #1859) until they merge.

PRs. This piece: `rawWarningFrame`'s fallback used to hand the raw warning
params straight to `JSON.stringify`, which walks the whole object graph
before anything gets a chance to slice it — a transport-sized (up to 128
MiB) malformed frame could make that walk proportional to the whole frame.
`prunedForStringify` prunes strings, arrays, object keys, nesting depth, and
total node count before `JSON.stringify` ever sees the value, enumerating
object keys with a counted `for...in` loop rather than `Object.entries` so a
many-key object is never allocated in full just to keep its first 50 keys.

Closes round-5 findings (`.../reviews/raw/1862-99bde6efb.md`):
- member 0/1/2/3, `reducer.ts:1270` (Medium/Medium/Medium/High) —
  `Object.entries(value).slice(...)` materializes every entry before
  capping; fixed by the counted `for...in` loop with a Proxy-backed test
  proving the walk never accesses more than the key cap's worth of
  property values.

Files: `reducer.ts` (`prunedForStringify` and its constants, wired into
`rawWarningFrame`), `reducer.test.ts` (5 tests: field-length truncation,
depth cap, total node budget, the Proxy get-count test, oversized property
name).


obra:

obra:

obra:
## Measurement: for...in over a 1e6-key Proxy vs JSON.parse of the same frame

Per the coordinator's request, measured member 0's finding (#1732) that `for...in` can still enumerate the entire key set before the prune's break, making `prunedForStringify` O(frame) for a million-key object.

Scratch vitest (not committed): built a flat 1,000,000-key object, `JSON.stringify`'d it (15,888,891 bytes — simulating the wire payload), timed `JSON.parse` of that string, then wrapped the parsed object in a `Proxy` (counting `ownKeys`/`getOwnPropertyDescriptor` trap calls) and ran it through `applyNotification({method:"warning", params:{extra: observed}})`. Three runs:

| run | parseMs | pruneMs (full applyNotification) | ownKeysCalls | ownKeysReturnedCount | gopdCalls | bare `Reflect.ownKeys` alone (ms) |
|---|---|---|---|---|---|---|
| 1 | 196.8 | 669.3 | 1 | 1,000,000 | 102 | 179.8 |
| 2 | 218.6 | 714.4 | 1 | 1,000,000 | 102 | 183.9 |
| 3 | 227.4 | 738.8 | 1 | 1,000,000 | 102 | 205.2 |

Findings:
- `ownKeys` is called exactly once and — confirmed — must return the FULL 1,000,000-key array; the review's claim that `for...in`'s key enumeration is O(frame) for a Proxy-wrapped object is correct, not refuted on that point.
- `getOwnPropertyDescriptor` is called ~102 times (close to the 50-key cap plus incidental margin), confirming the per-key enumerability check IS lazy/bounded as intended — the loop's own break works.
- Isolating the enumeration cost specifically (a bare `Reflect.ownKeys` call, no Proxy dispatch, no recursion): ~180-205ms, the SAME ORDER as — actually slightly less than — the `JSON.parse` cost of the identical frame (~197-227ms). The one genuinely O(frame) step the prune adds is not materially more expensive than the parse the client already paid to have the object exist at all.
- The full `applyNotification` call (~669-739ms) is ~3.2-3.5x the bare parse, but that includes Proxy trap-dispatch overhead (which the review's own test construction introduces) plus ordinary reducer bookkeeping unrelated to the prune — not a second O(frame) walk beyond the one `ownKeys` call already accounted for.

**Refuting the "materially worse than the parse" framing**: the enumeration step itself costs about the same as parsing the frame. I'm not proposing the fixed-field-serialization alternative (known warning fields only) since the numbers don't show it's needed — that option changes what the message-less fallback can show and I'd want your go before touching it either way.

obra:
## Round 2 fixes (8d6befe16)

- **`reducer.ts:1073` for...in, Medium (member 0, member 1)** — the comment was overclaiming: `for...in` still needs one full enumeration of the object's own keys, same as `Object.keys`/`Object.entries` would need, and per the earlier measurement (posted on this PR) that enumeration is the same order as the `JSON.parse` cost already paid to have the object exist — no NEW asymptotic cost. What the loop actually avoids is allocating a `[key, value]` pair per key and reading more property values than survive the cap. Rewrote the comment to say this precisely, and extended the Proxy test (previously only counting `get` calls) with an `ownKeys` trap counter asserting exactly one enumeration call, so the test documents the true bound instead of the previous overclaim.

- **`reducer.ts:1065/1083` `__proto__`, Low (member 0, member 2)** — fixed. `pruned` was a plain `{}`; assigning a wire key literally named `"__proto__"` (a genuine own, enumerable property when the source came from `JSON.parse` — no setter involved) invoked `Object.prototype`'s `__proto__` setter instead of creating an own property, silently dropping that field. `pruned` is now `Object.create(null)`. Failing-first test: a `{"__proto__":{"marker":"present"}}` frame (built via `JSON.parse` to match how a real wire frame arrives) — before the fix, `item.text` was `{"threadId":"thr_t","ref":"ref_t","warning":42,"extra":{}}` (field silently dropped); after, it contains `"present"`.

- **`reducer.ts:1054/1082` UTF-16 slice, Low (member 1)** — not touched on this PR; already fixed on the next commit in the stack (#1894), which extracts `boundedPrefix` (code-point-safe) and makes both of these sites call it instead of a raw `.slice`.

Gates: `reducer.test.ts` 210/210, `npm run typecheck` clean, `npx biome ci ../../../appwire-client/typescript` clean, `mobile-native npm run check` clean, `make lint-package-imports` PASS. Falsification: reverting this commit's `reducer.ts` change in isolation fails exactly the new `__proto__` test (209 others stay green).

obra:
## Round 3 disposition (8d6befe)

All three findings (unbounded message/title/hint/source inspection and storage; UTF-16 `slice` truncation of field values and property names) are the subject of the next PR in this stack, #1894, which introduces the shared code-point-safe `boundedPrefix`/`boundedContent` helpers and applies them to every string the fold stores, including the prune's value and key truncation. This PR is the bounded walk only; #1894 merges immediately behind it. No change here.

PR #1894, opened 2026-09-18
fix(sdk): warning strings are bounded before any copy

obra:
Stacked on #1893; diff against main includes it (and #1892, #1859) until
they merge.

PRs. This piece: `boundedCodePoints` truncates a string to
`RAW_WARNING_FRAME_MAX_CHARS` code points without expanding past a small
multiple of that bound first, and `rawWarningFrame`/`foldWarningParams` now
apply it to every string a warning frame can put into the model (message,
title, hint, source, and the raw fallback), not just the fallback text.

Also fixes an open finding not yet addressed on #1862's head:
`warningMessage`/`hasWarningText` decided "is there any content here" by
calling `.trim()` on the raw wire candidate before anything bounded it — an
O(length) allocation on a value the transport allows up to 128 MiB.
`hasWarningText` now bounds its input before trimming, and `warningMessage`
reuses it instead of repeating the unbounded check. New failing-first test
("a huge warning message is never trimmed at full size before it's
bounded") spies on `String.prototype.trim` and asserts every call it
observes already received a bounded candidate.

Closes:
- round-5 member 1, `reducer.ts:1201` (Medium) — `.trim()` on unbounded wire
  strings before `boundedCodePoints` ever runs.
- a-lane-state-note.md's open finding #1 (same location).

Files: `reducer.ts` (`boundedCodePoints`, `rawWarningFrame`,
`foldWarningParams`, `warningMessage`, `hasWarningText`), `reducer.test.ts`
(3 tests: the Array.from-ordering test, the message/title/hint/source
bounded-at-fold test, and the new trim-before-bound test). No changes
needed to `WarningItem.tsx` or `transcriptProjector.ts` — both already
consume `WarningFold` correctly.


obra:

obra:

obra:
## Rebased onto #1893's new tip (8d6befe16)

Absorbs #1893's round-2 fixes: the corrected `for...in` comment + extended Proxy test (ownKeys counter), and the `Object.create(null)` fix for a wire key named `__proto__`. Own diff (`boundedPrefix` extraction + `hasWarningText` regex fix) confirmed byte-identical via a merge-conflict resolution in `reducer.test.ts` (pure insertion-point conflict — #1893's new __proto__ test and this PR's new tests both anchored on the same preceding test; resolved by keeping both, __proto__ test first since it groups with the other prune tests). New head: 069c2126f28336f807fa7bc057d09f43cbd234e6. Gates: 215/215 vitest, typecheck/biome/native-check/lint-package-imports all clean.

obra:
## Round 2 fixes (8b40112e2)

**`reducer.ts:1170-1179` leading-padding misclassification, Medium (member 0, member 3; also raised on #1732's panel)** — fixed. `hasWarningText` detects content anywhere in the string, but `boundedCodePoints` always kept the LEADING `RAW_WARNING_FRAME_MAX_CHARS` code points — a message/title/hint/source with more leading blank code points than that bound, followed by real content, passed `hasWarningText` yet was stored as nothing but the blank prefix, so every consumer rendered nothing.

`boundedPrefix` now takes an optional `start` offset and slices a single, already-bounded window from there (never the whole `start`-to-end remainder — bounded to at most `maxCodePoints * 2` UTF-16 units regardless of how far `start` is into the string). `boundedContent(s)` finds the first non-whitespace index with `/\S/.exec(s)?.index` (no copy) and bounds the window starting there; `foldWarningParams` uses it for `text`/`title`/`hint`/`source` in place of `boundedCodePoints`.

Failing-first tests: a title with 3000 leading spaces then "URGENT" — asserts `item.warning.title === "URGENT"` (the stored fold value); strengthened the existing message-leading-blank test's assertion from "doesn't fall back to the raw frame" to `item.text === "real content"` exactly, since the fix makes that the actual stored value now. Didn't add a WarningItem/mobile-row test — both render `item.warning.title`/`item.text` verbatim and are already covered for arbitrary strings; the bug and fix are entirely in the fold, so a renderer-level test wouldn't exercise anything the fold-level test doesn't.

**Invariant, closing the class**: stored warning strings are the bounded prefix of the content, never of the padding in front of it.

Gates: 216/216 vitest, typecheck/biome/native-check/lint-package-imports all clean. Falsification: reverting this commit's `reducer.ts` change in isolation fails exactly the 2 new/strengthened tests (214 others stay green).

**CI red on the previous head**: `race-modules/agent` "panic: test timed out after 10m0s" is the known agent-package-at-the-race-cap class (#1394) — unrelated to this TypeScript-only PR. This push gives it a fresh run.

obra:
## Rebased onto main (0a10857)

obra:
## Round 4 disposition (bb6dd7c)

Lows only (four: `boundedContent` drops leading whitespace on values that need no bounding; the `prunedForStringify` comment still says UTF-16 units; three global spies restore only on the happy path; the `hasWarningText` comment names `boundedCodePoints` where `boundedContent` is applied). Per the Lows-go-to-follow-ups rule they join the stack's follow-up PR with the earlier `__proto__`/key-collision and node-budget-test items. Merging.

PR #1806, opened 2026-09-18
piece A-server: the daemon owns the ask boundary

obra:
## Summary

Decomposed from #1731 round 4 (which grew the server-side single-source change to ~430 non-test lines on top of the client gate and broke 13 web ask-dock tests, 2 native tests, and the package qualifier — see #1731's round-4/round-5 disposition). This PR carries ONLY the Go/daemon mechanism: the restore single-source `turnResolvesAskBoundary` shared by `deriveRestoredState`/`deriveRestoredAskPending`, and the mid-flight resolving steer clearing `askPending`.

- **High (1731-r4)**: a user steer consumed mid-round never reached the model. Root cause was `#1731`'s move of `askedThisRound`'s capture to before `injectPostToolSteering`'s drain; this PR never carries that move, so the capture stays after the drain (its original position) and the round correctly continues to a second model request once a mid-round drain resolves the ask. `TestAskUser_InjectPostToolSteeringClearsPendingAskForAnAcceptedUserSteer` now scripts a second model step and asserts the drained steer's text reaches it — falsified by reintroducing the moved capture (fails: "model requests = 1, want 2"), restored and green.
- **Medium (1731-r4 + 1732-r4)**: an untagged `TurnFailure` from a claimed carrier's own failed skill-selection resurrected a cleared `askPending` on restore. Fixed with a short-lived claim marker (`setSteeringCarrierClaimDrain`/`steeringSelectionFailureIsCarrierClaim`, session.go/session_queue.go) so `recordFailedSteeringSelection` tags `SteeringCarrier` only when the failure belongs to the carrier's own claimed steer — new test `TestAskUser_RestoreResolvesAcrossFailedSteeringSelectionCarrier`, falsified by hardcoding the tag to `false` (fails: restored askPending non-empty), restored and green.
- **Medium (1732-r4)**: the `askedThisRound`/live-restore divergence this same finding described is fixed by the same non-move above.
- **Medium (1732-r4)**: the four restore tests now assert the restored `SessionState` alongside `askPending`, not just the pending set.
- **Low x3 (1731-r4)**: `turnResolvesAskBoundary` simplified to a single bool (the `decisive` half was always equal to `resolves`); the dangling `TestCommunicate_ClientSteeringDuringToolExecutionRunsBeforeTheNextRequest` citation is gone; round-history language (`#1731 round 4`) scrubbed from the carried comments.

[...]

PR #1895, opened 2026-09-18
fix(mobile): canonical warning rows compose message and hint; short live-row ids

obra:
Stacked on #1894; diff against main includes it (and #1893, #1892, #1859)
until they merge.

PRs — the last piece.

Fixes two open findings not yet addressed on #1862's head:

- `project.ts:533` (a-lane-state-note.md open finding #2 / round-5 member 1
  Medium, `project.ts:527`) — `item.text || warningFallbackText(item) ||
  item.output` dropped the hint whenever a message was present, while the
  web (`WarningItem.tsx`) and the live warning row (`conversation.ts`'s
  `case "warning"`) both compose message and hint together. `warningFallbackText`
  now composes `item.text` with the hint when there's a message, falling
  back to title+hint for a message-less frame the same as before (keeping
  the existing "surfaces a warning's title/hint... message-less frame" test
  green without weakening it).

- `conversation.ts:2861` (a-lane-state-note.md open finding #3 / round-5
  member 1 Low `conversation.ts:2882`, member 3 Low `conversation.ts:2882`)
  — the live warning row's id embedded the sanitized title
  (`warning:${title}:${serial}`), which `foldWarningParams` only bounds to
  2000 code points — still enough to bloat this row's timeline id and the
  ownership keys it feeds. The serial alone is already unique, so the id
  drops the title. Updated one existing assertion that pinned the old
  `warning:Warning:` id format to the looser `warning:` prefix (the same
  "stays short" behavior it was already checking is unchanged) and added a
  new oversized-title test.

Also carries forward a comment-only clarification (from #1862's own diff)
on the live item replacement path's input-image-preservation comment in the
same file — no behavior change.

Files: `mobile/src/conversation/project.ts` (`warningFallbackText`) +
`project.test.ts`, `mobile/src/state/conversation.ts` (the live `case
"warning"`'s id) + `conversation.test.ts`.


obra:

obra:

obra:
## Disposition: two findings refuted (deriveAskQuestions.ts, now merged to main in #1892)

Two findings from the review panels on this stack's earlier heads, both in `deriveAskQuestions.ts` (landed as part of #1892, already merged to main — recording the disposition here per the coordinator's instruction):

**`deriveAskQuestions.ts:110` (`if (!model.askPending) return [];`) — Medium, "no fallback if askPending is absent/undefined on an old daemon"** — refuted by ruling: Jesse, 2026-09-18, #1823 — no fallback. An old daemon that hasn't been updated to send `askPending` renders no pending ask at all until the daemon is updated; this is the intended, accepted behavior for the transition, not a gap to patch around client-side.

**`deriveAskQuestions.ts:90` (`return item.source === "user" && item.steeringKind !== "human-note";`) — Low, premised on the server clearing `askPending` on a human-note-sourced steer** — refuted by fact: that premise is exactly what #1806 S1 removes server-side. Client and server agree: a human note does not answer a pending ask. The exclusion here is correct and has nothing to fix.

obra:
## Rebased onto #1894's new tip (069c2126f)

Mechanical rebase, no conflicts, own diff (project.ts/conversation.ts warning composition + live-row id) confirmed byte-identical. New head: 6d745a42787f8b34aea9e326ad5cd301874c122e. Gates: 463 mobile-native tests, 215 reducer tests, typecheck/biome/native-check/lint-package-imports all clean.

obra:
## Rebased onto #1894's new tip (8b40112e2)

Absorbs #1894's round-2 fix (boundedContent: stored warning strings bound the content, not leading padding). Mechanical rebase, no conflicts, own diff (project.ts/conversation.ts warning composition + live-row id) confirmed byte-identical. New head: 934f4ea1f8e15407767686e99112ad37ce954cfd. Gates: 463 mobile-native tests, 216 reducer tests, typecheck/biome/native-check/lint-package-imports all clean.

obra:
## Rebased onto main (0a10857)

obra:
## Round 3 fixes (5b94cf898)

**`project.ts:544` title dropped when a message present, Medium (member 1)** — fixed. `warningFallbackText` composed `[item.text, item.warning.hint]` when there was a message, dropping `item.warning.title` entirely — web `WarningItem.tsx` renders title + message + hint, and the live warning row keeps title as its own field, but this canonical projected activity row has no separate title slot, so the title was silently lost.

Added `joinWarningParts` to the package (`reducer.ts`, exported from `index.ts`) as the one join-and-filter rule every fold-as-a-string surface shares — both mobile call sites now use it instead of two separate inline `.filter(hasWarningText).join(" — ")` copies (this was also simplify follow-up #7 from the original a-lane state note): the canonical projector passes `[title, text, hint]`, the live row `[text, hint]` (title stays its own field there, unchanged — confirmed the live row's existing tests still pass with no behavior change).

Order `[title, text, hint]` matches the web `WarningItem.tsx` oracle's rendering order (title chip, then message body, then hint below).

Failing-first test: title + message + hint all present — asserts all three appear in the projected row's `detail.output`. Falsification: reverting this commit's four production files in isolation fails exactly that one new test (463 others, including all live-row tests, stay green).

**`reducer.test.ts:4650` spy-on-internals, Low (member 1)** — refuted, not fixed. Those spies (`JSON.stringify`, `Array.from`, `String.prototype.trim`) exist because an earlier review round explicitly demanded proof that no full-size copy is allocated — observable output (the bounded, correct final string) cannot distinguish "computed cheaply" from "computed by copying the whole 5MB input and then truncating the result," which is exactly the class of regression these tests exist to catch. Keeping them.

Gates: 464 mobile-native tests, 216 reducer tests, typecheck/biome/native-check/`make test-api-package`/lint-package-imports all clean (`test-api-package` run because this PR touches `index.ts`; falsified the new export too — removing it breaks `mobile-native npm run check` at both `project.ts` and `conversation.ts`'s import lines, confirming both consumers actually need it).

obra:
## Rebased onto main (4b9b4ba)

PR #1900, opened 2026-09-18
D6 p6a: draft read-outcome classification and discard-result typing

obra:
Stacked on #1808; diff against main includes it until it merges. First of three pieces #1791 (p6) is decomposed into after six review rounds - see [that PR's comment](https://github.com/prime-radiant-inc/evener/pull/1791) and its closing note.

## What this piece delivers
The package-level and pure-primitive half of the unreadable-draft recovery work:

- `discardStoredDraft` (draftCheckpointPort.ts) returns a `DiscardStoredDraftResult` (`"removed"`/`"absent"`/`"refused"`) instead of a bare boolean, and takes an optional `isReadable` so it refuses rather than deleting a record that now decodes as valid.
- `keybindingsStore.ts` gains `draftUnreadable`: a present-but-undecodable record is a RECORD problem, not a port problem, so `restoreDraft` clears `draft`/`writeUncertain`/`draftConflict` for it and a new `assertDiscardable` gate lets `discardDraft` through despite `storageUnavailable`. `isReadableKeybindingDraft` is exported for the native store-free discard path to reuse.
- `settledWrite` gains a `savingAtReadStart` guard: a write already in flight when a settling read began, which fails before that read's reply lands with no newer write ever claiming a token, must not be mistaken for settled.
- Native (`nativePreferenceDrafts.ts`) gets the read/decode primitives a real Storage-backed backend needs to classify what it reads without throwing: `StoredNullRecord` and `UnparseableDraftBytes` (tagged markers - a stored JSON null and bytes that fail to parse are each distinct from "no record"), `parseDraftBytes`, `matchesStoredBytes` (canonicalized byte compare), and `DraftReadOutcome`/`classifyDraftRead`/`readDraftOutcome` for a cold-offline probe or a store-free discard's re-read.
- `canonicalJson` (sorts object keys recursively) lives in `draftCheckpointPort.ts` and is exported from the package root, so `testing/draftStorage.ts`'s `memoryDraftStorage` and mobile-native's `fakeDraftBackend` compare identities the same way the real Storage-backed port does.

## Findings fixed in this piece
- **#1791 round-6 Medium** (`keybindingsStore.ts:679`): a genuine storage-read failure was clearing `draftUnreadable`, erasing an earlier unreadable-record recovery. `restoreDraft`'s catch now omits the field for a non-`UnreadableDraftError` throw instead of setting it `false`. (Also closes the same-class finding p10's round-2 panel raised against the inherited code at `keybindingsStore.ts:726` - one fix.)
- **p9 (#1844) round-2 panel Medium**: `testing/draftStorage.ts`/`draftBackend.testkit.ts` compared identities with order-sensitive `JSON.stringify` while the production native port compares via canonicalized bytes - the two test doubles now share the one `canonicalJson` the production port uses (via `matchesStoredBytes`).
- `memoryDraftStorage`/`fakeDraftBackend`'s null/undefined-identity and missing-key false-match bugs (both fixed on `#1791`'s own tree; carried forward here since this piece owns those files).
- **p9 panel Medium** re: `keybindingsStore.ts:1050`'s settle-revision - measured and refuted on `#1808` (see that PR's follow-up comment); not applicable here.

## Gates
- `npx vitest run` (mobile-native, full suite): 828 passed
- `npx vitest run appwire-client/typescript` (from cmd/evener-hub/frontend): 1991 passed
- `npm run typecheck` (cmd/evener-hub/frontend) and `npm run check` (mobile-native): clean
- `make test-api-package`: qualified (index.ts gained one named export, `canonicalJson`; no module added/moved, so the "shipped module unreachable" falsification does not apply)

## Falsification
- The `#1884` insert-if-absent fix and the CAS-refusal fix already carry their own falsification on `#1808`.
- The new `keybindingsStore.ts:679` fix: added `failLoad` to `memoryDraftStorage` (a genuine port failure, distinct from `corrupt()`'s present-but-undecodable record), wrote a test asserting `draftUnreadable` survives a subsequent genuine read failure, watched it fail against the pre-fix code (`draftUnreadable: false` instead of `true`), then fixed.

## Non-test line count
`git diff 7dcf01e66...HEAD --stat` (this piece's own diff, i.e. everything after the fold's final round-5 tip): 278 non-test lines across `draftCheckpointPort.ts`, `keybindingsStore.ts`, `index.ts`, `testing/draftStorage.ts`, `nativePreferenceDrafts.ts`. Above the 80-150 target (folding in the round-6 fix, the p9 canonicalization consolidation, and the `failLoad` test knob added scope); under the 400 hard ceiling.


obra:
## Round 2 disposition + restack on fold round 6 (9723dbc1)

Restacked on fold's round-6 tip (#1808 @ d421d548); own commits unchanged in content, replayed clean (no conflicts).

Round-2 panel findings fixed here:

**`discardStoredDraft` didn't catch port throws** (`draftCheckpointPort.ts`), unlike `readDraftOutcome`: `load()`/`removeIf()` may throw a genuine storage failure per the port's own docs, and a store-free caller with no live repository to publish the failure through had it escape uncaught. Wrapped the body in try/catch, widening the return type to `DiscardStoredDraftResult | "storageUnavailable"`. Pinned by three new tests (load() throws, removeIf() throws, the re-read after a failed removeIf throws), each falsified by removing the catch.

**Marker collision** (`nativePreferenceDrafts.ts`): `isStoredNullRecord`/`isUnparseableDraftBytes` detected their marker by a bare `kind` string alone, so a future checkpoint carrying the same field name would misclassify as the sentinel instead of itself. Both markers now use a namespaced kind value plus an exact key-count check, ruling out any real checkpoint (which always carries its own substantive fields). Pinned by a new test constructing exactly that overloaded-kind collision.

**Extra finding from #1792's round-2 panel, fixed here since it needs `isReadableKeybindingDraft` which lives here**: `index.ts` re-exported `draftCheckpointPort.discardStoredDraft` under the keybinding name with the generic default intact (`isReadable = () => false`, never refuses) - a caller using the named export without also passing `isReadableKeybindingDraft` would silently delete a readable record instead of refusing. Replaced the bare re-export with a real wrapper in `keybindingsStore.ts` that defaults to it. Pinned by a new test calling the named export with no second argument against a readable record, asserting "refused"; falsified by reverting the default to `() => false`.

Items measured and NOT actioned here (belong elsewhere/already resolved - see #1902's and #1904's own dispositions):
- The production native backend's `JSON.parse`/`JSON.stringify` gap: `NativePreferencesProvider.tsx` isn't touched by this PR's own commit at all (still pre-p6b content); p6b's own commit is what first wires the backend to `parseDraftBytes`/`matchesStoredBytes` for the offline probe, so the shared-extraction + production-level-test fix lands there.
- `draftUnreadable` missing from `PreferenceState`/`keybindingsDomain`: `nativePreferences.ts` isn't touched by this PR either; already fixed by p6c's own commit (`nativePreferences.ts` gains the field, `keybindingsDomain` projects it, with its own falsified regression test in `keybindingRecovery.test.ts`).

Head: `9723dbc1d7e2dacef768042e2edd6932493aaade`

Own diff vs fold's new tip (`d421d548..9723dbc1`): 9 files, 951 insertions / 40 deletions.

Gates (targeted, all green): vitest on `draftCheckpointPort.test.ts` + `keybindingsStore.test.ts` + `nativePreferenceDrafts.test.ts` (150+ tests across the touched files); `npm run typecheck`; `npx biome ci ../../../appwire-client/typescript`; `npm run lint`; `npm run check` in mobile-native; `make lint-package-imports`; `make test-api-package`.

Pushed with `--force-with-lease=claude/sdk-d6-p6a-draft-read-outcome:a24e7c59d`.

obra:
## Round 2 disposition (9723dbc)

- High/Medium (the production native backend in `NativePreferencesProvider.tsx` still parses with `JSON.parse` and compares with raw `JSON.stringify`, so the tagged markers and canonical identity only run in the fakes): this PR is the package read-outcome primitives only. The next PR in the chain, #1902 (piece 6b), replaces the native backend with the shared `rawStringDraftBackend` that production and the test fake both use, which is exactly the parity this finding asks for; it merges directly behind this one.
- Medium (`draftUnreadable` dropped from the native projection; no discard control when `draft` is null): #1902 projects `draftUnreadable` through `PreferenceState` and #1904 (piece 6c) renders the enabled discard action for the store-free recovery case, with tests cited in their dispositions.
- Low (`canonicalJson(undefined)`): filed as #1918 with two sibling Lows; follow-up PR after the chain.
No change here.

obra:
## Rebased onto main (8d8ed0d2)

Rebased p6a onto `origin/main` via `git rebase --onto origin/main d421d5489 claude/sdk-d6-p6a-draft-read-outcome` now that the fold (#1808) squash-merged as `8d8ed0d27`. Byte-identical refresh: diffed old-tip..old-base patch against new-tip..origin/main patch, the added/removed lines are identical (only index.ts hunk headers shifted due to unrelated upstream changes to that file). No code change.

```
 appwire-client/typescript/draftCheckpointPort.test.ts | 166 +++++++++++-
 appwire-client/typescript/draftCheckpointPort.ts      |  64 ++++-
 appwire-client/typescript/index.ts                    |  11 +-
 appwire-client/typescript/keybindingsStore.test.ts    | 155 ++++++++++-
 appwire-client/typescript/keybindingsStore.ts         | 120 ++++++++-
 appwire-client/typescript/testing/draftStorage.ts     |  23 +-
 mobile-native/src/draftBackend.testkit.ts             |  20 +-
 mobile-native/src/nativePreferenceDrafts.test.ts      | 297 ++++++++++++++++++++-
 mobile-native/src/nativePreferenceDrafts.ts           | 135 +++++++++-
 9 files changed, 951 insertions(+), 40 deletions(-)
```

Gates: vitest (draftCheckpointPort.test.ts, keybindingsStore.test.ts) 116 passed; mobile-native vitest (nativePreferenceDrafts.test.ts, draftBackend.testkit.ts) 34 passed; frontend `npm run typecheck` clean; mobile-native `npm run check` clean; `npx biome ci ../../../appwire-client/typescript` clean; frontend `npm run lint` clean; `make lint-package-imports` PASS; `make test-api-package` qualified.

New head: `adbf4cab6f4bc078c4fda98c39fd6b8892d99406`

obra:
## Round 3 disposition (adbf4ca)

Byte-identical refresh onto the fold's squash; both Highs are the same seam as round 2 (the production native backend parity is #1902's `rawStringDraftBackend`; the `draftUnreadable` projection and the enabled discard action are #1902/#1904), merging directly behind this one. No change.

PR #1902, opened 2026-09-18
D6 p6b: native offline keybindings-draft probe and store-free discard

obra:
Stacked on #1900 (p6a); diff against main includes it and #1808 until they merge. Second of three pieces #1791 (p6) is decomposed into.

## What this piece delivers
Wires p6a's read-outcome primitives into the native app's actual offline recovery feature:

- The real Storage-backed backend (`NativePreferencesProvider.tsx`) routes its `get()`/`matches()` through `parseDraftBytes`/`matchesStoredBytes` so a stored JSON null or bytes this build cannot parse classify as an unreadable RECORD, not a dead port or a silent "no draft".
- A cold-offline probe (`useEffect` on `[hubId, state]`) classifies the local record via `readDraftOutcome` before any model publishes a snapshot - `bindNativePreferences`'s `ready` callback never fires on a cold offline start, so this is the only signal the "Discard unreadable draft" action has to render on.
- `discardUnreadableKeybindingsDraft()` clears an unreadable record with no live model required, re-reads to tell "still unreadable" from "replaced with something readable" (`draftUnreadableAfterDiscard`), and publishes the replacement's `draft`/`writeUncertain` fields the same way `restoreDraft` would on a live store (`decodeKeybindingDraftFields`, new on `keybindingsStore.ts`).

## Findings fixed in this piece (all from #1791's round-6 panel)
- **Medium** (`nativePreferenceDrafts.ts:171`): `draftUnreadableAfterDiscard` cleared the notice whenever the discard's own OUTCOME was `"removed"`/`"absent"`, even if the follow-up read found a FRESH unreadable record. It now takes only the follow-up read's own classification - the discard's outcome plays no part in what is stored NOW.
- **Medium** (`NativePreferencesProvider.tsx:177`): a storage failure during the offline discard returned the same `null` documented for "no hub", making a failed discard indistinguishable from an action that never ran. Returns a distinct `"storageUnavailable"` outcome instead.
- **New** (surfaced during this piece, not in the original panel): the offline probe retained a PREVIOUS hub's `draftUnreadable` flag when a NEW hub's read failed. Scoped by a `probedHubId` ref so a failure right after a hub switch resets to `false` rather than showing the old hub's classification under the new hub's name.
- **Low**: `setBound`'s offline-discard update published only `draftUnreadable`/`storageUnavailable`/`error`, leaving `draft`/`writeUncertain` stale when the discard's re-read found a readable replacement. Now decodes and publishes both via `decodeKeybindingDraftFields`.

`offlineStorageUnavailable` itself is still set but not yet consumed by a screen - that is p6c's own finding (the Low restated against `KeybindingPreferencesScreen.tsx`).

## Gates
- `npx vitest run` (mobile-native, full suite): 833 passed
- `npx vitest run appwire-client/typescript` (from cmd/evener-hub/frontend): 1993 passed
- `npm run typecheck` (cmd/evener-hub/frontend) and `npm run check` (mobile-native): clean
- `make test-api-package`: qualified (index.ts gained one named export, `decodeKeybindingDraftFields`; no module added/moved)

## Falsification
`draftUnreadableAfterDiscard`'s corrected logic: reproduced the original panel's exact scenario (an "unreadable" follow-up read) as a test, confirmed it fails against a reverted (return-false) implementation, restored.

## Residual (flagged, not fixed here)
`NativePreferencesProvider.tsx` has no test file of its own (true before this change too - the original ~130-line diff that added this feature had none either). The wiring in this piece is exercised only through its downstream screen (p6c) and indirectly through the pure functions it delegates every decision to, all of which ARE unit-tested (`draftUnreadableAfterDiscard`, `clearsErrorAfterOfflineDiscard`, `decodeKeybindingDraftFields`, `readDraftOutcome`, `matchesStoredBytes`). A React Testing Library harness for this provider would be a reasonable follow-up but is out of scope for this stack.

## Non-test line count
`git diff a24e7c59d...HEAD --stat` (this piece's own diff): 209 non-test lines across `index.ts`, `keybindingsStore.ts`, `NativePreferencesProvider.tsx`, `nativePreferenceDrafts.ts`. Above the 80-150 target (the four findings above added scope beyond the base wiring); under the 400 hard ceiling.


obra:
## Round 2 disposition + restack on p6a (654f9e4e)

Restacked on p6a's new tip (#1900 @ 9723dbc1); own commit unchanged in content, replayed clean, but a downstream typecheck fix was needed in the same file (below).

Round-2 panel findings fixed here:

**Production backend duplicated its parser instead of sharing it with tests**: `NativePreferencesProvider.tsx`'s `get()`/`deleteIf()`/`replaceIf()` already routed through `parseDraftBytes`/`matchesStoredBytes` (this PR's own commit is what first wires that in), but as an inline copy separate from `nativePreferenceDrafts.test.ts`'s hand-rolled `rawBytesBackend` fake - tests exercised a parallel reimplementation, not the function production calls. Extracted the shared logic as `nativePreferenceDrafts.ts`'s `rawStringDraftBackend`, generic over any `RawStringStorage`; production is now `rawStringDraftBackend(Storage, () => Crypto.randomUUID())`, and `rawBytesBackend()` wraps the same function over a Map-backed fake. Added a key-order-different-identity case through the backend's own `deleteIf` (stored-null and malformed-record cases already existed via `rawBytesBackend` and now cover the production path too, per your ask for all three).

**`discardUnreadableKeybindingsDraft`'s follow-up read called `storage.load()` a second time, unguarded**, after the guarded `readDraftOutcome` call classified the current record: a throw there escaped with the unreadable flag already cleared. Added `readDraftOutcomeWithValue`, which reads once and returns both the classification and the loaded value; `readDraftOutcome` now delegates to it. The provider decodes the value it already has instead of reading the port again. Pinned by three new tests on the extracted function (classifies the same as `readDraftOutcome`, calls `load()` exactly once, degrades a throw to `storageUnavailable`) - the double-load bug itself is only reachable through the JSX-wired provider, which has no test harness (react-native pull, no RTL - see the file's own residual), so falsification is at the extracted function: reverting it to call `storage.load()` twice fails the "exactly once" test.

**Downstream fix, same file**: rebasing onto fold's round-6 (`discardStoredDraft` now returns `DiscardStoredDraftResult | "storageUnavailable"` instead of throwing) broke this commit's own outer try/catch (`let outcome: DiscardStoredDraftResult` no longer accepted the widened type). Simplified to check `outcome === "storageUnavailable"` directly - the outer catch was already dead code once the callee stopped throwing.

Head: `654f9e4ea7d7e2173e1ff818653101747447bcf8`

Own diff vs p6a's new tip: 6 files, 408 insertions / 79 deletions.

Gates (targeted, all green): mobile-native's full vitest suite (838+ tests) and `npm run check`; `npm run typecheck`; `npx biome ci ../../../appwire-client/typescript`; `npm run lint`.

Pushed with `--force-with-lease=claude/sdk-d6-p6b-native-offline-discard:4c51a14cb`.

obra:
## Round 3 disposition (9081247)

Rebased p6b onto `origin/main` via `git rebase --onto origin/main 9723dbc1d claude/sdk-d6-p6b-native-offline-discard` (p6a's OLD head, now that p6a is squashed as `ead2caba7`). Byte-identical: own +/- lines match exactly against the pre-rebase diff. Then addressed the three outstanding items:

**1. Simplify must-fix** (`keybindingsStore.ts` `decodeKeybindingDraftFields`/`restoreDraft` duplication): extracted `draftFieldsFrom(checkpoint)` - both now call it. Behavior-preserving refactor; existing tests (93 in keybindingsStore.test.ts) are the oracle and all pass unchanged.

**2. Held p8 Medium** (`nativePreferenceDrafts.ts:25-97` marker shape collision): measured it - `matchesStoredBytes` compares via `draftIdentityKey`, a STRUCTURAL/shape check (`isStoredNullRecord`/`isUnparseableDraftBytes`), not raw bytes. A stored value that is legitimately parseable JSON but happens to have the exact marker shape gets misclassified as that marker, so it can collide with an identity captured from genuinely different (malformed/null) bytes. Falsifying test added first (confirmed failing: both new `matchesStoredBytes` cases returned `true` instead of `false` against unfixed code); fixed by branding both markers with a `unique symbol` - JSON.parse can never produce a symbol-keyed property, so no legitimately parsed value can ever satisfy the predicates by coincidence.

**3. p6b round-2 Medium** (`NativePreferencesProvider.tsx:127-190` shadow-only update): the store-free discard patched only `bound.snapshot` when a live model existed; the model's OWN keybindings store never heard about the storage change, so its next publication (the subscribe callback always replaces `bound.snapshot` wholesale with `model.getSnapshot()`) overwrote the patch with the still-stale classification. Fixed by nudging the live model's own `discardKeybindingsDraft()` when one is bound for this hub/client: its CAS refuses (the record it classified is now gone/replaced) and it re-reads storage itself via `discardDraft`'s existing "refused, so reclassify" branch - already covered for the "replaced" case by existing tests; added a new keybindingsStore.test.ts case for the "deleted entirely" case (confirmed it already passes unmodified, since this is the exact mechanism the delegation relies on). Deviation: could not write a component-level failing-first test for the Provider wiring itself - `NativePreferencesProvider.tsx` has no test harness (react-native via ConnectionProvider -> AppState; #1908). Correctness rests on delegating to the already-tested store mechanism.

```
 appwire-client/typescript/index.ts                 |   1 +
 appwire-client/typescript/keybindingsStore.test.ts |  34 ++++
 appwire-client/typescript/keybindingsStore.ts      |  37 ++++-
 mobile-native/src/NativePreferencesProvider.tsx    | 173 +++++++++++++++++----
 mobile-native/src/nativePreferenceDrafts.test.ts   | 166 ++++++++++++++++----
 mobile-native/src/nativePreferenceDrafts.ts        | 149 ++++++++++++++----
 6 files changed, 463 insertions(+), 97 deletions(-)
```

Gates: frontend vitest (keybindingsStore.test.ts 93+26=119, draftCheckpointPort.test.ts) all pass; mobile-native vitest 853/853 pass; frontend `npm run typecheck` clean; mobile-native `npm run check` clean; `npx biome ci ../../../appwire-client/typescript` clean; frontend `npm run lint` clean; `make lint-package-imports` PASS; `make test-api-package` qualified.

New head: `90812476e396392bac64c31232ca09fd4bf11aae`

Residual: #1908 (no NativePreferencesProvider.tsx/KeybindingPreferencesScreen.tsx test harness) still blocks a direct component-level test of item 3's wiring - not new, already filed.

STOPPING here for merge signal before p6c per the coordinator's instruction.

obra:
Superseded by #1950 (draft storage classification), #1956 (offline provider recovery), and #1964 (retained snapshot and diagnostic ownership). The original branch is preserved. #1904 remains the separate recovery screen and is integrating the final parent. Replacement PRs are open and still require current-head CI and review qualification before landing.

PR #1905, opened 2026-09-18
agent: a steer resolves the ask boundary only when it answers it (live)

obra:
## Summary
First of three tiny PRs decomposing #1806 (five review rounds, past the
queue's decompose-at-five threshold). Piece 1 of 3.

- Adds `agent/schema.TurnFailureInfo.SteeringCarrier` and the shared
  `steeringSourceAnswersAsk` / `steeringCarrierClaimAnswersAsk` predicates
  (`agent/session_tools_ask.go`): a human-note update is user-sourced
  steering exactly like an answering steer, but does not address a pending
  `ask_user` question, so it must not clear `askPending`.
- `processOneInput`'s entry clear (`agent/session_lifecycle.go`) is now
  gated on `steeringCarrierClaimAnswersAsk` instead of clearing
  unconditionally for every accepted turn.
- `consumeSteeringMessage` (`agent/session_queue.go`) now clears
  `askPending` before `EventSteeringInjected` publishes (not after), so the
  server's ask-facet refresh on that event never reads a stale flag.
- A steering-carrier claim's own `TurnFailure` (its steer failed to append,
  or its skill selection failed to prepare) is tagged `SteeringCarrier` —
  but only when the claimed steer itself answers the ask.

This is the LIVE half only. A follow-up PR (piece 2) makes the daemon's
restore path read the same tag/predicates so a restart agrees with the live
session on this boundary; a third (piece 3) hardens the carrier claim's
panic safety.

Stacked on #1806; diff against main includes it until it merges.

## Test plan
- `go test -count=1 -run 'TestAskUser_HumanNoteDuringPendingAskDoesNotResolveIt|TestAskUser_InjectPostToolSteeringClearsPendingAskForAnAcceptedUserSteer|TestAskUser_SteeringInjectedNeverObservesAskPendingStillTrue' ./agent/` — green, plain and `-race`.
- Same set `-tags evenerfuzz` — green.
- `$(go env GOROOT)/bin/gofmt -l`, `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`, `golangci-lint run ./...` — all clean.
- Falsified: reverting this PR's own production files (schema/turn.go, session.go, session_config.go, session_events.go, session_queue.go, session_lifecycle.go, session_tools_ask.go) makes the new tests fail to build (the identifiers the tests exercise don't exist without the fix); restored, green.


obra:
## Round 1 disposition (9899c96)

**Dominant finding (High/Medium x3): `SteeringCarrier` written but never read; restore still re-derives.** This is the intended split seam, not a gap: piece 2 (#1906, stacked on this one) is the restore reader and merges immediately after this piece. This piece alone leaves `main` no worse than today — restore already re-derives `askPending` from the transcript exactly as it did before either piece, since nothing here reads the new tag or predicate yet. Fixed the actual defect the finding also raised: every comment (`agent/schema/turn.go:144-170`, `agent/session.go:575`, `agent/session_tools_ask.go:132`) that named `turnResolvesAskBoundary` or asserted present-tense restore behavior has been reworded to say the reader lands in the next change (#1906) — no comment in this tree claims a symbol or behavior that doesn't exist here.

**Medium — `session_queue.go:1257` event-order flip.** Round 3's original intent was only to move `clearAskPendingForResolvingSteer` ahead of `EventSteeringInjected`; the change also moved `admitPreparedSkillSelection`/`unparkSteering` ahead of the emit, flipping the observable order to Warn-then-Injected whenever a skill-bearing steer's own admission fails. Restored the original admit/unpark-after-emit order in both delivery branches; only the clear (and its `beforeSteeringInjectedPublish` hook) still moves ahead of the emit. `TestAskUser_SteeringInjectedNeverObservesAskPendingStillTrue` stays green (unaffected — it only cares about the hook's timing relative to the emit). No existing test pinned Injected-before-Warn, so added `TestAskUser_ConsumeSteeringMessageEmitsInjectedBeforeAdmissionWarning`, driven through the same production path `TestSkillActivation_SteeringSelectionReconciledAfterAdmissionSaveFailure` already exercises (`consumeSteeringMessage` + `breakSessionMetaPath`). Falsified: reverting just the reorder made the pinned test fail with `EventSteeringInjected at 26, EventWarning at 24`; restored, green.

**Low — `session_tools_ask.go:117` fail-open on missing provenance.** `steeringCarrierClaimAnswersAsk` returned `true` (answers) when the carrier's client-mutation journal record was missing entirely, the same as a present-but-kindless record. A carrier identity's record always exists by construction (`SetHumanNote`/`AcceptClientMutationSteer` write it before queuing), so a missing id has no evidence either way — now fails CLOSED. New test `TestAskUser_SteeringCarrierClaimAnswersAskFailsClosedForUnknownProvenance`. Falsified: reverting the presence check made it fail (`= true ... want false`); restored, green.

**Medium — missing behavioral coverage for both `TurnFailure` tagging shapes.** Neither new test exercised the real drain path for the tag; added four, each driven through the real production path (`AcceptClientMutationSteer`/`SetHumanNote` + `ProcessPendingUserInput`, or `recordFailedSteeringSelection` directly for the human-note selection-failure shape SetHumanNote's own RPC can't reach), asserting the persisted `SteeringCarrier` flag on the last turn and the live `askPendingCount` directly (no restore needed — that's #1906's job): `TestAskUser_AcceptSteeringCarrierInputTagsSteeringCarrierOnAppendFailureForAnAnsweringSteer`, `TestAskUser_RecordFailedSteeringSelectionTagsSteeringCarrierForAnAnsweringSteer`, `TestAskUser_AcceptSteeringCarrierInputDoesNotTagSteeringCarrierOnAppendFailureForAHumanNote`, `TestAskUser_RecordFailedSteeringSelectionDoesNotTagSteeringCarrierForAHumanNoteClaim`. Falsified the two negative tests together by temporarily making the tag unconditional at both gating call sites: both failed (`want a TurnFailure NOT tagged SteeringCarrier`); restored, green.

**Low — `session_lifecycle.go:2744` missing `defer`.** Confirmed out of scope: that's piece 3's (#1907) own change, not this piece's.

### A real bug found while investigating the review (not in the panel, found via measurement): drain-ladder follow-up gate
`session_lifecycle.go`'s drain-ladder gate (~line 1399) read `awaiting := s.State() == SessionAwaiting` as a proxy for "a pending ask survived this round", with a comment proving the equivalence — provably true before this piece (the entry always cleared `askPending` unconditionally), false now (a human-note carrier's entry clear is skipped). Measured: a note-carrier round that only acknowledges the note via `communicate` settles `SessionIdle` at this exact capture point (`deliverIfCommunicated`'s `askedThisRound` delta is false since nothing NEW was posted; the general awaiting upgrade runs later, at the outer loop's own terminal settle), so the gate read "not awaiting" while `askPendingCount() == 1`, and a queued `FollowUp` got popped and run past the unanswered question. Fixed: the gate now also reads `askPendingCount() > 0` directly. New test `TestAskUser_FollowUpNotDrainedWhilePendingAskSurvivesAHumanNoteCarrierRound`, falsified (revert made the follow-up run while ask1 was still pending; restored, green).

### Gates (agent module, from `agent/`)
- `$(go env GOROOT)/bin/gofmt -l .`: clean
- `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`: clean
- `golangci-lint run ./...`: 0 issues
- `go test -count=1 -run 'TestAskUser' .`: green
- Same set `-tags evenerfuzz` and `-race`: green
- `TestNoBareWallClockDeadlineInAgentTests`: green (no new bare literal)

Own diff (`git diff --stat b62990e80..9899c9630 -- <files touched>`): 8 files, 634 insertions(+), 64 deletions(-) across two commits; non-test production diff is ~229 (round 1) + ~57 (round 2) ≈ 286 lines, above the 150 target — flagged rather than split further, since it's one cohesive mechanism (the live steer/carrier gating) plus its own review-round fixes.

Head: 9899c9630207db2f7a1d4f53c7eb6cd584d1063f

obra:
## Round 2 disposition (9899c96)

- Medium (restore still skips `TurnSteering` and the `SteeringCarrier` tag): this is the split seam. #1906, stacked directly on this PR, is the restore reader and merges immediately behind it; main is no worse in between because restore already re-derives from the transcript today. No change here.
- Low (the fail-closed journal check is bypassed on the delivery path via `clearAskPendingForResolvingSteer`): confirmed as the same class as the round-2 fail-closed fix; it goes into the stack's follow-up PR after #1905-#1907 merge, per the Lows-in-follow-ups rule.

PR #1906, opened 2026-09-18
agent: restore derives the pending ask from the boundary the live path uses

obra:
Restoring a session now derives pending questions using the same entry and carrier rules as the live session. Ordinary failures no longer hide the preceding real entry, and provider failure, no-tool exhaustion, and tool-round exhaustion preserve awaiting state while a human-note carrier leaves a question unanswered. Pending-state reads use one lock so state and question count cannot disagree between observations.

This is the ask-boundary restore piece of the approved SDK migration. #1907 remains the separate carrier-defer successor. Existing asked-this-round behavior is preserved; no askPending fallback or protocol compatibility shim was introduced.

Validation: scripted live/restore lifecycle regressions and production-only falsification pass; scoped plain and evenerfuzz ask tests, toolchain gofmt, normal/tagged/Windows-tagged vet, and touched-module lint pass. Independent review/simplification found no must-fix issues. Local RoboRev2553 passed the final implementation at 84e6df360b330f86c9c53863d7b22a20fc275db2. Merge refresh 8c55253d1b9479a30dfb9558dd806a66ee00e440 incorporates main 9cb596336f6b33fa61606f950b99dbf3829a9222 without changing that patch; scoped gates passed again.

Coordinator size disposition: 452 changed non-test lines (351 additions, 101 deletions) are accepted as one cohesive implementation of the already-approved ask-resolution contract. Splitting restore classification from matching live boundaries would separate the invariant under test. This is the fifth review round; any further own must-fix requires decomposition. The Low legacy text-evidence/helper follow-up is #1946. No projector output changed. Current-head CI and raw remote panel review remain required before merge.


obra:
## Round 1 disposition (7d14c74, rebased onto #1905's round-1 fixes)

Reviewed at fe2ec0873 before the rebase; the panel's finding carries forward unchanged onto the rebased tip.

**Medium — `session_tools_ask.go:589` (panel numbering) — restore drops earlier questions when a non-resolving round posts a new one.** `deriveRestoredAskPending` stopped its backward scan at the first (newest) `ask_user` round it found and returned only that round's questions. Live `askPending` never replaces, only appends (`registerAskTool`'s Exec: `s.askPending = append(s.askPending, parsed...)`), so a human-note carrier that leaves ask1 pending can also post a brand-new ask2 in its own round — live ends up with both, restore returned only ask2.

Fixed: the scan now accumulates every non-resolving round's questions (each round's own within-round call order preserved, rounds combined oldest-first to match live's append order) and only stops once it reaches a round whose own entry actually resolved the boundary (`roundEntryResolvesAskBoundary`) or a directly resolving turn. New test `TestAskUser_RestoreAccumulatesPendingQuestionsAcrossNonResolvingRounds` (ask1 posted, human-note carrier's own round posts ask2, live `askPendingCount()==2`, restore also returns both headers). Falsified: reverting to the single-round `return questionsFromAskCalls(...), true` made it fail (`restored askPending = [{Naming...}], want 2 questions`); restored, green — and confirmed no regression on the pre-existing `TestAskUser_RestoreRebuildsPendingCountAndOrder` (two `ask_user` calls in ONE round), which needs the within-round order preserved, not reversed, by the multi-round combination.

### Second finding, folded in from #1907's round-1 panel (same mechanism, this piece owns the fix)
**Low/Medium — kindless, provenance-less human-note turns classify as answering.** `turnResolvesAskBoundary`'s `TurnSteering` case read `origin.steeringKind()` directly, which recovers note origin from a recorded kind or a `notes/human/set` method but — unlike `steeringOrigin.isHumanNoteSteer`, the same journal-provenance helper history escaping already uses — never falls back to the write-path text shape (`humanNoteSteerPrefix`) when neither is available. This matters specifically for an inherited fork prefix: `steeringOriginBoundary` nils the origins lookup for every turn before `divergenceTurn` regardless of what the map carries, so a legacy pre-kind-stamping note in that prefix has no provenance left except its own text — and used to read as an ordinary answering steer.

Fixed by calling `origin.isHumanNoteSteer(turn.Message.Text())` directly at this one call site (the only one of `steeringKind()`'s three callers that has both an `origin` and the turn's text, and the only one exercising the inherited-prefix gap) instead of duplicating its fallback chain — one shared predicate, not a second copy. New test `TestAskUser_RestoreClassifiesAKindlessProvenancelessHumanNoteByItsTextShape`, modeled on the sibling `TestAskUser_RestoreUsesJournalProvenanceForAKindlessLegacyHumanNoteTurn` but with `divergenceTurn` set so the note is entirely inherited (origins nilled despite being passed). Falsified: reverting to the plain `steeringSourceAnswersAsk(turn.SteeringSource, origin.steeringKind())` call made it fail (`pending=[], isAskRound=false, want ask1 still pending`); restored, green.

### Gates (agent module, from `agent/`)
- `$(go env GOROOT)/bin/gofmt -l .`: clean
- `go vet ./...`, `go vet -tags evenerfuzz ./...`, `GOOS=windows go vet -tags evenerfuzz ./...`: clean
- `golangci-lint run ./...`: 0 issues (one `modernize: slicesbackward` finding fixed along the way — the reversal loop now uses `slices.Backward`)
- `go test -count=1 -run 'TestAskUser|TestRestored|TestInherited|TestHistoryCopy|TestRebuilt|TestAcceptSteeringCarrierInput' .`: green
- Same set `-tags evenerfuzz`, and `-race` on the restore-dependent subset: green
- `TestNoBareWallClockDeadlineInAgentTests`: green

Own diff (`git diff --stat 9899c9630..8bc4805c1`): 2 files, 204 insertions(+), 7 deletions(-); non-test production diff ~65 lines (session_tools_ask.go).

Head: 8bc4805c195de97f0eca319af982a8a6e025e7a3

obra:
## Rebased onto main + simplify (c4761e3)

Rebased onto `main` @ 07eb8edf7 (S1 #1905's squash-merge). Own diff vs that base: 10 files, 1495+/100-, clean rebase (no conflicts).

**Simplify must-fixes (reviews/1906-simplify.md):**
1. Collapsed `steeringSourceAnswersAsk`/`turnResolvesAskBoundary`'s hand-rolled steering arm into one `steeringAnswersAsk(source, origin, textEvidence)`, called by `clearAskPendingForResolvingSteer`, `steeringCarrierClaimAnswersAsk`, and `turnResolvesAskBoundary`. No behavior change at any call site — falsified by breaking the collapsed predicate (dropping the `isHumanNoteSteer` check) and confirming 9 existing tests catch it across all three call sites, then restoring.
2. Reworded the three stale comments (`schema/turn.go`, `session.go`, `session_tools_ask.go`) that said `SteeringCarrier` isn't read yet or that restore "shares this predicate once it lands" — both now false; replaced with the true fact (which function reads it).

**Round-2 review panel findings, each measured before acting:**
- **Fixed** — TOCTOU: the drain-ladder gate read `s.State()` and `s.askPendingCount()` as two separate locked calls. Added `awaitingOrHasPendingAsk()` sampling both under one lock. Falsified: dropping the pending-check half made `TestAskUser_FollowUpNotDrainedWhilePendingAskSurvivesAHumanNoteCarrierRound` fail; restored.
- **Fixed** — a failed human-note carrier append left `askPending` set but the generic non-provider-failure tail in `processOneInput`'s caller settled the session idle unconditionally, diverging from restore's `SessionAwaiting` for the identical transcript (and, via `WireState`, telling a live client nothing needs them). New test `TestAskUser_LiveStateAfterFailedHumanNoteCarrierAppendMatchesRestore` drives this through the real `ProcessPendingUserInput` path — watched it fail (`idle`, want `awaiting`) before the fix, pass after.
- **Fixed** — `roundEntryResolvesAskBoundary`'s backward walk only skipped `TurnAssistant`/`TurnToolResults`, so a bookkeeping turn (`TurnEnvironment`, `TurnCheckpoint`, etc.) interleaved before the real entry was misread as the entry itself. Now only `TurnUserInput`/`TurnSteering`/`TurnFailure` are treated as entry-capable, matching the outer scan. New test `TestRoundEntryResolvesAskBoundary_SkipsBookkeepingTurnsToFindTheRealEntry` — watched it fail before the fix, pass after.
- **Not implemented, refuted by measurement** — the "askedThisRound's delta reads false when a mid-round resolving steer clears a question this round just posted" Medium: the clear (`clearAskPendingForResolvingSteer`) is unconditional and always runs after the round's own `ask_user` append, so the round's ending count is always exactly 0 when this fires — never able to land back on a positive `askBefore`. `askedThisRound=false` there is correct: nothing is left pending. Built the exact scenario and measured directly: with the suggested generation-counter fix, the turn stopped one round early and forced `SessionAwaiting` with `askPendingCount()==0` (3 model requests → 2), bypassing Stop hooks for a boundary with nothing to await — a regression, not a fix. Left as delta-based; `TestAskUser_MidRoundResolvingSteerClearingAFreshAskDoesNotForceAwaitingEarly` pins the correct 3-request behavior.
- **Not implemented, refuted by measurement** — the "kindless legacy human-note carrier treated live as an answering steer" Medium: `SetHumanNote` stamps both `SteeringKind` and `Method` atomically at creation, so no record written by current code can ever be kindless-and-methodless. That shape needs data predating both fields AND still sitting as an undrained queue entry — no current write path produces it, and no evidence it's reachable.

**Gates (from `agent/`):** toolchain gofmt clean, `go vet ./...` / `-tags evenerfuzz` / `GOOS=windows -tags evenerfuzz` all clean, `golangci-lint run ./...` 0 issues, targeted tests (`TestAskUser|TestRestored|TestInherited|TestHistoryCopy|TestRebuilt|TestAcceptSteeringCarrierInput|TestRoundEntryResolvesAskBoundary`) green plain and `-tags evenerfuzz`, `-race` on the new/changed tests green.

Own diff vs `main` @ 07eb8edf7: 10 files, 1495 insertions(+), 100 deletions(-). Head: `c4761e30835699d68776188156c02c3b41de5e87`.

obra:
Superseded by #1958 (durable restore oracle) and #1962 (live failure and yield boundaries). Both replacement PRs preserve the committed implementation and regressions; the original branch remains preserved. The replacement stack is still under review, including the new same-round steering-order finding on #1958, and is not yet merge-ready.

PR #1913, opened 2026-09-18
refactor(sdk): C5 follow-ups (submission runner factory, shared types, dead method field, test title)

obra:
Follow-ups from the opus /simplify review of #1901 (D25e C5), `reviews/1901-simplify.md`, findings 3-7, plus one round-2 panel Low.

## Finding 3 — `MutationSubmissionCommitted` renamed a copy of the store's own return shape
`settleSubmittedDraft` returns `{ cleared, draftUnchanged }`; the module rebuilt that object purely to rename `cleared` → `clearedDraft`.

Fix: `MutationSubmissionCommitted` is now `ReturnType<PendingTurnsStore["settleSubmittedDraft"]>`, passed straight through (`onCommitted?.(store.settleSubmittedDraft(opts.ref, opts))`). The web adapter destructures `cleared` instead of `clearedDraft`.

## Finding 4 — `track`'s type re-declared `MutationProjectionWorkTracker["track"]`
`submission.ts` spelled out `<T>(work: Promise<T>) => Promise<T>` again instead of naming the tracker's own method type.

Fix: `track: MutationProjectionWorkTracker["track"]` on both `submitWithPendingTracking` and the new `createSubmissionRunner`. Structurally identical; nothing at any call site changes.

## Finding 5 — the same fire-and-forget refresh arrow written twice
`(ref) => void refreshPendingTurnsProjection(ref)` was built inline for the commit feed and again for the submission call.

Fix: one module-level `const refreshTarget: (ref?: string) => void = (ref) => void refreshPendingTurnsProjection(ref);`, passed to both `wireMutationCommitFeed` and the new submission runner.

## Finding 6 — six required positional parameters, four of which never vary per call
On the web, `store`/`fence`/`track`/`refresh` are module-level singletons; only `opts`/`perform`/`onCommitted` differ between submissions.

Fix: added `createSubmissionRunner(store, fence, track, refresh)`, returning `(opts, perform, onCommitted?) => Promise<void>` - the same binding-the-singletons-once convention as `createPendingTurnsStore`, `createMutationProjectionWorkTracker` and `wireMutationCommitFeed`. The web builds `runSubmission` once at module scope and calls it with just the three per-call arguments. `submitWithPendingTracking` stays exported too, for a caller that needs a per-call fence or tracker (native's binding, if it turns out to need one) - purely additive, no behavior change.

## Finding 7 — `SubmitWithPendingTrackingOptions.method` was a required field nothing read
`method` was thrown away by the core the moment #1901 shrank the adapter; the PR description's justification held for `attachments`/`recoveryId` but not for `method`.

Fix: deleted the field and its six call-site properties: `Composer.tsx`, `QueueStrip.tsx`, and four call sites across `pendingTurnsStore.test.ts`/`QueueStrip.test.tsx`. Purely subtractive.

## Round-2 panel Low
`submission.test.ts`'s "reports the draft as unchanged and uncleared when the stored draft has moved on" asserted `draftUnchanged: false` - backwards from its own title (the draft moved on, so it is neither cleared nor unchanged). Retitled to state the actual contract.

## Gates
- `npx vitest run` on the eight touched test files: **171 passed (171)**.
- `npm run typecheck` (frontend): clean.
- `npm run lint` (biome ci over `src` + the package, two files auto-formatted for the shortened lines): clean, 1373 files.
- `npx biome ci ../../../appwire-client/typescript`: clean, 232 files.
- `make lint-package-imports`: PASS.
- `make test-api-package`: qualified.
- `npm run check` (mobile-native tsc): clean.

## Diff stat

```
```


obra:
## Round 1 disposition (a8e273e)

The Medium (`MutationSubmissionCommitted` renamed `clearedDraft` to `cleared`, "breaking existing package consumers") is refuted by scope: `@evener/appwire-client` is consumed only inside this repository (the web frontend and the native app), both of which this PR updates, and it is not published to any registry. Keeping `clearedDraft` alongside `cleared` would be a backward-compatibility shim, which this project does not add without an explicit ruling. No change.

Issue #1914, opened 2026-09-18
native: PluginsScreen's store has no reconnect recovery (connectionChanged never called)


obra:
Second half of #1595, explicitly deferred there: \"should the screens stay
mounted through a flap and let the shared lifecycle's reconnect recovery
refresh the list in place (then the native wiring of connectionChanged
becomes the second half of that change)\".

D28 D2 (#1595's fix) makes PluginsScreen and HubSettingsScreen survive a
connection flap behind a banner instead of unmounting into a Reconnect wall.
That alone is enough for the common case: a transient auto-reconnect
(`state: "reconnecting"`, the same `AppwireClient` instance, per
hubConnection.ts's generation guard) never changes `client`'s identity, so
`PluginsScreen`'s `useMemo(() => createPluginsStore(client), [client])` was
never rebuilt by the flap itself - only the screen's own wall was destroying
it.

What D2 does NOT close: `createPluginsStore`'s returned `PluginsStore`
already extends `StoreLifecycle`, which exists specifically to answer "the
hub broadcasts evener/plugin/updated only to clients connected at the
moment it fires - a change made while this one was briefly disconnected
reaches it as nothing at all, and only a `connectionChanged`-driven
reconnect can recover it" (storeLifecycle.ts's own extensive documentation).
PluginsScreen never calls `connectionChanged` - it only ever constructs the
store once per `client` identity - so if the hub's plugin set changes during
a flap this screen missed, the list stays stale until the user does
something else that remounts the child (a different hub, a fresh app
launch). This is exactly issue #1595's original framing: "D11 (#1502) added
reconnect recovery to the shared lifecycle... no native caller."

`HubSettingsScreen`'s `createHubOverviewStore` is unaffected - it has no
push notification at all by design ("no push-driven invalidation... callers
decide when to fetch() or refresh()", hubOverview.ts's own module doc) and
already offers pull-to-refresh (`RefreshControl`), so nothing to wire there.

`ProvidersScreen` already has this: `useCredentialStore()`
(credentialStore.ts) builds its store once via `useState` and rebinds it via
`connectionChanged` in a layout effect on every client/state change,
independent of the screen's own wall/banner decision - that's the pattern
`usePluginsStore()` would mirror.

Suggested fix: a `usePluginsStore()` hook in `mobile-native/src` (same shape
as `useCredentialStore()`) that builds the store once - lazily, the first
time a client exists, since `createPluginsStore(client)` needs one at
[...]

Issue #1908, opened 2026-09-18
mobile-native lacks a React Testing Library harness for screens/providers


obra:
Follow-up from the D6-p6 decomposition (#1900/#1902/#1904, 2026-09-18).

`NativePreferencesProvider.tsx` and `KeybindingPreferencesScreen.tsx` (and every other file importing `react-native` in `mobile-native/src/`) have no test coverage of their own JSX/effect wiring: `react-native`'s entry point is Flow-syntax source, which vitest's rolldown-based transform cannot parse, so a test file that imports one of these modules directly fails with a parse error before any test runs.

This predates the D6 work - it's a pre-existing gap, not something introduced by this lane. Both files above got real logic bugs fixed in #1902/#1904 (an offline-probe hub-switch flag carryover, a discard action disabled by a stale hub flag while offline, a screen preferring a stale domain snapshot over a fresh offline probe) that were only reachable by extracting the decision logic into plain, `react-native`-free modules (`keybindingOfflineRecovery.ts`) purely to make them unit-testable. The actual `useEffect`/JSX wiring around those decisions in both files remains untested.

Suggested fix: add `@testing-library/react-native` (or equivalent) to `mobile-native`'s devDependencies and a small render-harness helper, then backfill at least the two files above. Not urgent, but the pattern of "extract pure logic into a sibling file to dodge the Flow-parse issue" is a workaround, not a fix, and will keep recurring as more screens grow decision logic worth testing directly.

PR #1915, opened 2026-09-18
feat(mobile): screens survive a connection flap behind a banner (D28 D2)

### LINKED ISSUE #1595: native: extension screens unmount into a Reconnect wall on every transport flap, losing scroll, selection and filter text

obra:
Superseded by #1952 (retained screens and reconnect banners) and #1955 (live mutation readiness). #1922 remains the required recovery successor. Both replacement PRs preserve the reviewed original implementation in smaller pieces; codex/reconnect-round-four preserves the combined round-four head. The replacement stack remains held for recovery qualification.

---

Stacked on #1912 (the hubConnection follow-ups PR): this branch is built on
top of it because this PR also touches `hubConnection.ts` (adding `fatal`).
The diff below therefore includes #1912's commit until that one merges;
only this PR's own commit (`d86b50bf4`) needs reviewing here.

Per Jesse's ruling 7 (2026-09-18, #1595): screens survive a connection flap
behind a banner instead of the full-screen Reconnect wall. Closes #1595.

## Measurement

- The wall: `PluginsScreen.tsx`, `HubSettingsScreen.tsx` and
  `ProvidersScreen.tsx` all had the same shape - `if (!client || state !==
  "ready") return <View>...Reconnect...</View>` (or the ternary
  equivalent) - unmounting the ready-only child (and its store, scroll
  position, selection) on ANY non-ready transition, including a passive
  auto-reconnect where the underlying `AppwireClient` never even changes
  identity.
- The web equivalent (`cmd/evener-hub/frontend/src/shell/ConnectionBanner.tsx`):
  the workspace always stays mounted; a floating banner appears only for
  `ATTENTION_STATES = {"reconnecting", "closed"}`, after a `delayMs` grace
  window (10s default - "a routine sub-second reconnect never surfaces, a
  genuinely stuck connection tells the user promptly"). That's what "brief
  flap" means on the web: immediate banner-eligibility, delayed reveal - not
  something D2 needed to reproduce (mobile's own reconnect backoff, and this
  PR's own everReady/fatal gating, already keep a routine blip silent-ish
  behind a banner rather than a wall; a reveal-delay timer is a separate,
  optional polish this PR does not add).
- `#1595`'s own body already named the fix in two halves: "screens stay
  mounted... (then the native wiring of connectionChanged becomes the
  SECOND HALF of that change)". This PR is the first half only. Filed the
  second half as prime-radiant-inc/evener#1914 (PluginsScreen's store has no
  push-notification reconnect recovery - `connectionChanged` is never
  called - though a passive flap alone doesn't need it, since `client`'s
  identity is stable through one; only a missed hub-side change during the
  blip would be stale, which #1914 covers).

## Design

`connectionDisplay(state, everReady, fatal)` (`mobile-native/src/connectionDisplay.ts`,
pure, react-native-free per #1908's constraint):
- `"none"` once `state === "ready"`.
- `"wall"` when the screen has never shown anything yet (`!everReady`), or
  the close is fatal (a protocol mismatch a retry can never fix).
- `"banner"` for every other non-ready state, once something has already
  been shown - the flap the screen now survives.

`useConnectionDisplay(state, fatal)` wraps it with an `everReady` ref
(mutated during render, the same pattern `ForkScreen.tsx`'s `owner.current`
already uses).

`fatal` is new on `HubConnection`/`Connection` (`hubConnection.ts`,
`ConnectionProvider.tsx`): true only for a protocol close, computed
alongside the existing error-message classification, reset on every fresh
attempt.

`ConnectionStatus` (the existing status-row banner `screens.tsx` already
used for the Sessions/Conversation lists - name/state text + a conditional
Reconnect action + `ErrorMessage`) is reused rather than building a new
banner. It had to move out of `screens.tsx` into its own file first:
`screens.tsx` pulls in `@react-navigation/elements` (among a large import
graph), which fails to load under vitest (a `.png` asset import) - a VALUE
import of `ConnectionStatus` from a test-reached module broke
`ProvidersScreen.test.tsx` until the move (measured: reproduced the
failure, fixed by extraction, confirmed green).

Plugins/HubSettings additionally track the last non-null client in a ref:
`client` stays set through a passive flap (hubConnection.ts's own
generation guard), but a manual retry's token-refetch window does clear it
briefly, and the last-known-client fallback keeps the child mounted through
that gap too rather than dropping to the wall for a moment the banner
already covers. `ProvidersScreen` needs no such fallback -
`useCredentialStore()` already survives a flap on its own
(`connectionChanged` rebinds it, `credentialStore.ts`), and `<Providers>`
takes only `store`, never `client`, directly.

## Auto-refresh (native rule: never a manual "refresh to see the latest" affordance)

- `ProvidersScreen`: already correct via `connectionChanged`'s reconnect
  recovery (pre-existing, unaffected by this PR).
- `HubSettingsScreen`: `createHubOverviewStore` has no push notification by
  design (its own module doc: "no push-driven invalidation... callers
  decide when to fetch()/refresh()") and already offers pull-to-refresh
  (`RefreshControl`) - a standard native idiom, not the banned affordance.
  Nothing to add.
- `PluginsScreen`: a passive flap never changes `client` identity, so
  nothing was stale to begin with for that case. A missed hub-side plugin
  change during a flap this screen was disconnected for is a real,
  pre-existing gap (confirmed: `connectionChanged` was never called) -
  filed as #1914 rather than bundled here, matching #1595's own two-halves
  framing.

## Tests (mobile-native has no RTL harness - #1908)

- `connectionDisplay.test.ts`: the pure function's 4 branches, plus
  `useConnectionDisplay` hook tests for the three named scenarios - "ready
  -> reconnecting keeps mounted + banner", "reconnecting -> ready removes
  the banner", "fatal shows the wall even after having been ready".
- `hubConnection.test.tsx`: `fatal` reports true only for a protocol close,
  clears on `"ready"` and on a fresh attempt's own generation.
- `ProvidersScreen.test.tsx` (extends the one existing real-mount screen
  test - `PluginsScreen`/`HubSettingsScreen` have no test file to extend,
  a pre-existing gap #1908 already tracks): the same three scenarios driven
  through the REAL mounted screen via `tree.update` + the mocked
  `useConnection` - the list's rendered rows stay present through
  `reconnecting`, the banner text appears and disappears with `state`, and
  a fatal close replaces the list with the wall.

## Falsification

- `connectionDisplay.ts` moved aside: `connectionDisplay.test.ts` failed on
  the missing module; restored, green.
- `ProvidersScreen.tsx` reverted to its pre-PR content: the new "ready ->
  reconnecting keeps ... mounted" test failed (list content gone, replaced
  by the wall) exactly as measured; restored, green; `git status --short`
  showed only the intended files throughout.

## Gates

- `npx vitest run` (mobile-native): 86 files / 821 tests passed
- `npm run check` (mobile-native, tsc): clean
- `make lint-package-imports`: PASS (no `appwire-client` files touched)

## Diff (includes #1912's commit until it merges - see the stacking note above)

```
```

Closes #1595.


obra:
## Rebased onto main (d4f2414)

#1912 squash-merged as `d4f2414f2d8cb78bc26c0626be43d3358714247d`. Rebased:

```
git fetch origin main
git rebase --onto origin/main 967f0a0206f4560a3f97842fde6342ee268cddd9 claude/sdk-d28-d2-flap-banner
```

Clean rebase, no conflicts. Verified the commit's own patch is byte-identical
before/after (`diff <(git show d86b50bf4...) <(git show acdb12de8...)` differs
only on the `commit` hash line). `git diff -M --stat origin/main...HEAD` now
shows only this PR's 11 files (381 insertions(+), 92 deletions(-)) - #1912's
commit is gone from the diff since it's on `main` now. Pushed with
`--force-with-lease=claude/sdk-d28-d2-flap-banner:d86b50bf495f57249b0847daa4a37d4244203901`.

New head: `acdb12de8248c506984d7c12064e2f2f5cd4d086`

obra:
## Round 2 disposition (f676fb4dd)

Panel findings 1-5 from the round-1 review, measured and fixed. This PR keeps findings 2, 4 and 5; findings 1 and 3 (the biggest and most novel piece - reconnect recovery for the retained stores, plus hub-scoping) are split into a stacked follow-up, **D2a**, because round 2's full diff hit 703+/109- against origin/main - well past the ~450 review ceiling. The reduced diff here is 458+/104- (237+/94- non-test).

**(1) Medium - Plugins/HubSettings never recover after a flap clears** - split to D2a (`claude/sdk-d28-d2a-reconnect-recovery`, pushed, not yet a PR). `PluginsScreen` now calls `model.connectionChanged(client, state)` on every connection transition (confirmed `PluginsStore` already has it, from `attachLifecycle`/`storeLifecycle.ts`); `createHubOverviewStore` has no such method (verified in `hubOverview.ts` - "callers decide when to fetch() or refresh()"), so `HubSettingsScreen` gets a small `useReconnectRecovery(state, onReconnect)` hook that calls `refresh()` + `reconcileAfterReconnect()` on every transition back to `"ready"`. Closes #1914. Falsified: removing the `connectionChanged` wiring drops the new `PluginsScreen.test.tsx` reconnect assertion from 2 `evener/plugin/list` calls to 1.

**(2) Medium - retained screens keep mutation controls live while reconnecting** - fixed here. One shared `isReady(state)` predicate (`connectionDisplay.ts`) gates: Plugins' enable/auto-upgrade switches, upgrade, remove, retry and pull-to-refresh; HubSettings' upgrade start/refresh/review-another (via `HubUpgradeSection`'s existing `disabled` prop), retry and pull-to-refresh; Providers' sign-in start (`flow.setConnection(ready ? client : null)`, fixing the gap where `onSignIn` bypassed the check the later effect already made), the sign-in/refresh-sign-in button, and pull-to-refresh.

**(3) Medium - `everReady`/`lastClient` never reset on a hub change** - split to D2a with (1) (same files, same review). `useConnectionDisplay` and `useRenderClient` both take `hubId` now and reset their ref the moment it moves, so a screen instance a navigator reuses across a hub switch can't show one hub's banner/client over another's.

**(4) Medium (test weakness) - fatal-wall test masked by `client: null`** - fixed here. `ProvidersScreen.test.tsx`'s fatal case now keeps `client: hub.client` through the transition; falsified by loosening `connectionDisplay`'s wall check to ignore `fatal` (still walling on `!everReady` alone): the test fails because the list stays on screen ("work" renders) since a `closed`+`everReady` state now reads as a banner, not a wall - proving the assertion is load-bearing on `fatal`, not on `client` being null.

**(5) Low - duplicated `lastClient` fallback** - fixed here. `PluginsScreen`/`HubSettingsScreen` both use one shared `useRenderClient(client)` from `connectionDisplay.ts`.

Gates: `mobile-native && npx vitest run` (86 files / 823 tests, all green), `npm run check` clean, root `make lint-package-imports` PASS (no package files touched). D2a's content was validated in the same combined-commit form before the split (829 tests green, including its own new `PluginsScreen.test.tsx` and the hub-scoping tests in `connectionDisplay.test.ts`); its branch tree is byte-identical to what was tested.

```
 mobile-native/src/ConnectionProvider.tsx    |  5 +-
 mobile-native/src/ConnectionStatus.tsx      | 33 +++++++++++
 mobile-native/src/HubSettingsScreen.tsx     | 72 ++++++++++++++---------
 mobile-native/src/PluginsScreen.tsx         | 49 +++++++++++-----
 mobile-native/src/ProvidersScreen.test.tsx  | 88 +++++++++++++++++++++++++++++
 mobile-native/src/ProvidersScreen.tsx       | 65 +++++++++++++--------
 mobile-native/src/connectionDisplay.test.ts | 84 +++++++++++++++++++++++++++
 mobile-native/src/connectionDisplay.ts      | 63 +++++++++++++++++++++
 mobile-native/src/hubConnection.test.tsx    | 59 ++++++++++++++++----
 mobile-native/src/hubConnection.ts          | 18 +++++-
 mobile-native/src/screens.tsx               | 26 +--------
 11 files changed, 458 insertions(+), 104 deletions(-)
```
Non-test: 237 insertions(+), 94 deletions(-).

Head: `f676fb4ddcefadf498289c72015858f0be65b6a4`.

obra:
## Round 3 disposition (a48f7e0c8)

Panel at f676fb4dd, three members. Two of member1's findings are D2a's own content (#1922, stacked on this PR): the High "everReady/lastClient not scoped to hub" and its Medium counterpart from member0's High-adjacent framing about reconnect recovery. Both are addressed in #1922, not here - see its own disposition. The remaining three findings (member0's second and third, member1's second and third) are #1915's own and real; all fixed.

**(1) Medium - Providers gated only OAuth sign-in and pull-to-refresh** - fixed. `ProvidersScreen.tsx`'s `act()` and `confirm()` are the two choke points every other mutation/probe already routes through (add/edit instance via `ProviderEditor`'s own `disabled`-driven `busy` guard, set/replace key or JSON, make-default, clear stored key, logout, remove); both now check `isReady(state)` once, refusing before touching any component state. Each affected `Action`'s `disabled` prop also gets `!ready` directly, and each `onPress` that doesn't route through `act`/`confirm` (test credentials, edit instance, replace key/JSON) is wrapped in the new shared `whenReady(ready, handler)`. The credential editor's save handler no longer calls `setKey("")` before the request - that line is deleted; the clear already happens on `act()`'s own success path, so `!ready` now correctly leaves typed input in place. **Deliberately NOT gated: the "Cancel" button** (`ProvidersScreen.tsx` credential-editor Cancel, and the top-level `AddMarketplace` Cancel) - it issues no request, and gating it would trap a user who lost connection while typing a key from backing out of the editor with no way out. Measured against the codebase's own convention before objecting: `ProviderEditor.tsx`'s own Cancel button (`disabled={saving}`, line 191) is *already* excluded from its `disabled` prop for exactly this reason - readiness gates writes, never the exit from a form.

**(2) Medium - MarketplaceBrowser had no readiness at all** - fixed. Added a `ready` prop (from `PluginsScreen`'s own `isReady(connectionState)`), threaded into `AddMarketplace`. Gated: "Retry marketplaces"/"Retry catalog"/"Retry installed status" (via `whenReady`), "Refresh source"/"Remove marketplace" (their local `refresh()`/`remove()` functions gained `|| !ready`), "Add marketplace" open + its modal's submit button, the catalog's and marketplace list's pull-to-refresh, and the per-plugin Install action (only when it would actually install - `!existing && !ready` - since "Open" on an already-installed plugin is a pure local navigation with no wire call, deliberately left ungated for the same reason Cancel is). `runGatedMutation` (`pluginMutationGate.ts`) now takes `ready` and refuses before calling `gate.run` at all - the "one place" fix the review asked for, since Plugins' own mutations, MarketplaceBrowser's `act()`, and `AddMarketplace`'s `submit()` all funnel through it. `AddMarketplace`'s kind-selector and both text inputs (source/name) are deliberately left gated on `busy` only, not `disabled` (`busy || !ready`): they're pure local form state with no wire call, and blocking them while merely disconnected would stop a user composing their pending add for no safety benefit; only its submit button and `HubPathField`'s browse (a real wire read) use the full `disabled`.

**(3) Low - error-retry/upgrade-start handlers had `disabled` but no `if (ready)` guard** - fixed. `PluginsScreen.tsx`'s list-error "Retry" and `HubSettingsScreen.tsx`'s "Retry hub information" and the upgrade's `onStart` are now wrapped in `whenReady`, matching the pattern their adjacent pull-to-refresh handlers already had.

Falsified: reverted each fix in turn (`git diff <file> > tmp.diff && git checkout HEAD -- <file>`, ran the test, confirmed it failed for the right reason, `git apply tmp.diff`, confirmed green again) -
- `pluginMutationGate.test.ts`'s new "not ready refuses without running the action" case: without the `if (!ready) return "refused"` line, resolves `"ran"` instead of `"refused"` and the action actually runs.
- `ProvidersScreen.test.tsx`'s new "a flap disables provider mutation controls, not only OAuth sign-in" case (mounts the real screen, presses the "work" row, asserts `disabled` on Add/Test/Edit/Clear-credentials/Remove before and after a `ready -> reconnecting` transition): without this round's `ProvidersScreen.tsx` changes it fails at the very first post-flap assertion (`Add provider instance` stays enabled).
- `connectionDisplay.test.ts`'s `whenReady` test is a direct pure-function test (ready=false is a no-op, ready=true calls through with its arguments) - the shared mechanism every other gate in this round is built from.

Not independently RTL-tested per control: `MarketplaceBrowser`/`HubSettingsScreen` have no mount harness (mobile-native has none for them - #1908); their gating was verified by code reading plus the shared `runGatedMutation`/`whenReady` pure-function tests, which is what actually executes at each of those call sites.

Gates: `mobile-native && npx vitest run` - 86 files / 826 tests, all green. `npm run check` clean. Root `make lint-package-imports` PASS (no package files touched).

```
 mobile-native/src/ConnectionProvider.tsx     |   5 +-
 mobile-native/src/ConnectionStatus.tsx       |  33 +++++++
 mobile-native/src/HubSettingsScreen.tsx      |  80 +++++++++-------
 mobile-native/src/MarketplaceBrowser.tsx     |  60 ++++++++----
 mobile-native/src/PluginsScreen.tsx          |  57 +++++++----
 mobile-native/src/ProvidersScreen.test.tsx   | 132 +++++++++++++++++++++++++
 mobile-native/src/ProvidersScreen.tsx        | 138 +++++++++++++++++----------
 mobile-native/src/connectionDisplay.test.ts  |  93 ++++++++++++++++++
 mobile-native/src/connectionDisplay.ts       |  78 +++++++++++++++
 mobile-native/src/hubConnection.test.tsx     |  59 ++++++++++--
 mobile-native/src/hubConnection.ts           |  18 +++-
 mobile-native/src/pluginMutationGate.test.ts |  21 +++-
 mobile-native/src/pluginMutationGate.ts      |   7 +-
 mobile-native/src/screens.tsx                |  26 +----
 14 files changed, 642 insertions(+), 165 deletions(-)
```
Non-test: 350 insertions(+), 152 deletions(-) - under the 400 non-test ceiling, no split needed this round.

Head: `a48f7e0c85a37e960db059f88801b4f696b4feab`.

obra:
## Simplify must-fixes (b689acb3b)

Applied the two must-fix findings from the /simplify review:

1. `hubConnection.ts:107-113` and `:117-123` — both `onStateChange`/`.catch` sites now call
   `connectionFailure(terminalReason)` once and read `failure.kind === "protocol"` for `setFatal`,
   instead of re-deriving `terminalReason === "protocol"` by hand alongside the same
   `connectionFailure(...).message` call. One encoding of "which closes are unrecoverable."
2. `ProvidersScreen.tsx` — deleted the five `whenReady(ready, ...)` wrappers at the save-key/JSON,
   make-default, clear-stored-key, logout, and remove-instance handlers (all of whose bodies call
   only `act()` or `confirm()`, both of which already `if (!ready) return;` at their own top, so the
   wrapper's suppressing branch could never change behaviour). Kept `whenReady` at the five sites
   that reach the model or local state directly without going through `act`/`confirm`
   (add-instance, test-credentials, edit-instance, the two credential-editor openers).

Behaviour-preserving; no test was changed. Gates: `mobile-native && npx vitest run` (86 files / 826
tests, includes the two `hubConnection.test.tsx` cases exercising fatal-vs-transport classification
and all five `ProvidersScreen.test.tsx` cases), `npm run check` clean.

Head: `b689acb3b36bcf52d4463b46d7f47c9900480c22`.

PR #1916, opened 2026-09-18
feat(native): MutationOutboxStorage write path over expo-sqlite (D25d-1a)

obra:
Adds the native SQLite mutation-outbox write adapter: enqueue, receipt settlement, optimistic records, recovery transfers, and rollback. Per-target sequence allocation now uses one UPSERT RETURNING statement within the existing savepoint; unique target/sequence indexes prevent duplicate persisted identity. Missing optimistic displays encode as JSON null so a valid intent can be enqueued without one.

This is the write half of the approved D25d storage split; #1917 adds the read methods and full storage-port declaration. Accepted records intentionally omit composerText, matching the web adapter. Repeating a clientMutationId rejects rather than overwriting an earlier enqueue.

Validation at `5a86de604d623a4183ec1ff8616d681ef8bc583f` after integrating main:
- 21 storage behavior tests pass, including allocation interleaving, constraint/rollback, absent display, and receipt settlement. Production-only reversals reproduced the allocation and encoding failures.
- Native typecheck and package-import checks pass.
- Independent spec, quality, and simplification review found no must-fix findings.
- Local RoboRev branch review passed with no findings (job 2543).

The PR is exactly 400 non-test added lines, retained as one storage mechanism. Separate-handle contention is measured and tracked in #1937; the shared-notes recovery-port follow-up is #1938. Existing simplification follow-ups remain #1927, #1928, and #1929. The earlier unrelated daemon timer CI failure is #1936. Current-head CI and raw remote panel review remain required before merge.

Raw-panel disposition at 5a86de604d623a4183ec1ff8616d681ef8bc583f: Muse and DeepSeek found no issues. Luna's High claim that expo-crypto 57.0.2 lacks getRandomValues is refuted by the installed, lockfile-resolved package: build/Crypto.d.ts line 55 declares the generic API, and build/Crypto.js lines 127–128 export it and delegate to the native module. Current-head native CI passes. GLM's accepted-row shape Low is tracked in #1929; full-intent persistence and recovery-settlement coverage are tracked in #1927. All 15 current-head CI checks passed. The refresh head 6776ea4fac57007bc1e6b2fdb87901438f7fb5db merges main 9cb596336f6b33fa61606f950b99dbf3829a9222 with a byte-identical own patch (SHA-256 8a1761fd7d9fc66451884bd6db56a41491a12604324539b3ae5d9d743111b427). The 21 focused tests and native/import checks pass again; fresh CI is required at this head.


obra:
## Round 2 disposition (c3ccf5b)

All three panel members' findings addressed, TDD against the IndexedDB oracle (`cmd/evener-hub/frontend/src/stores/mutationOutbox.test.ts`).

1. **High - payload missing clientMutationId** (all 3 members): `enqueueIntent` now persists `payload: { ...intent.payload, clientMutationId }`, matching `mutationOutboxIndexedDB.ts`. Fixed the assertion baked into the old test at `mutationOutboxStorage.test.ts:82` and added the same payload check.
2. **High - no transactional atomicity** (members 0, 1): `enqueueIntent`, `settleReceipt`, `settleApplied` and `transferToRecovery` now run inside `protected transaction(name, body)`, a `SAVEPOINT`/`ROLLBACK TO`/`RELEASE` wrapper - the same pattern `draftRepository.ts`'s `write()` already uses over this same synchronous database port. Added two failure-injection tests mirroring the oracle's `beforeCommit` tests: a SQLite trigger fails the outbox `DELETE` mid-operation (after the optimistic/recovery insert has already run), and both tests assert the insert gets rolled back with the outbox row intact - not just that the later statement never ran.
3. **Medium - settleReceipt only consulted the outbox table** (members 0, 1, 2): now resolves its source as `outboxRecord ?? recoveryRecord ?? optimisticRecord` and retires whichever of the three tables hold the record, literally mirroring the oracle. Added tests for the recovery-source and optimistic-source paths.
4. **High/Medium - bare `crypto.randomUUID()` global** (all 3 members): the default id source now goes through the package's `createSecureUUID` with an explicit `SecureRandomSource` backed by `expo-crypto` (the same native module `credentialStore.ts`/`nativeOrganization.ts` already use), injected via a new `randomSource` constructor option. Falsified with a test that deletes `globalThis.crypto` entirely and confirms `enqueueIntent` still succeeds through the expo-crypto-mocked source.
5. **Low - empty/whitespace targetRef accepted** (member 0): `enqueueIntent` now throws `"targetRef is required"` before allocating a sequence, matching the oracle.
6. **Low - SELECT-then-write races** (member 1): `markAttempted`, `markUnknown` and the shared `delete()` helper now issue a single conditional `UPDATE`/`DELETE` and decide their boolean from the statement's affected-row count (`changes`), never a preceding `SELECT`.

Falsification: reverted the production file to its pre-round-2 state (`git diff <head>..<parent> -- mobile-native/src/mutationOutboxStorage.ts | git apply -R`), ran the suite - the 7 new/changed tests failed for the exact reasons above (payload missing the id, empty targetRef accepted, `ReferenceError: crypto is not defined`, `settleReceipt` returning `false` from recovery/optimistic, both atomicity tests leaving an orphaned duplicate) - then restored and confirmed green.

Gates:
- `npx vitest run src/mutationOutboxStorage.test.ts` (mobile-native): 16/16 pass (this PR's own tests; #1917 stacks 5 more read-path tests on top, 21/21 there).
- `npm run check` (mobile-native, tsc): clean.
- `make lint-package-imports`: PASS.

Size: 331 non-test lines (was 254 before round 2; the transaction wrapping and the widened `MutationRecord<A>` bound `settleReceipt` now needs pushed it past the ~300 flag in the brief). Still under the 400 hard ceiling, so pushed without asking first, but flagging as instructed - the transaction helper (`protected transaction`) and the `changedRows`/`MutationOutboxRunResult` plumbing (~25 lines) are the read-path-independent pieces that could move out if this needs to shrink further, but they're used by every write method so I don't see a clean way to split them without duplicating the SAVEPOINT pattern.

#1917 rebased onto this tip (own lines identical modulo one shift: 1a's `settleReceipt` fix independently required widening `fromRow`/`get`'s generic bound from `MutationOutboxRecord<A>` to `MutationRecord<A>`, which 1b's original diff used to introduce itself - now inherited from 1a, so 1b's rebased diff is 7 lines smaller without any behavior change). Re-checked the read path against the oracle: `nextDispatchable` and `restoreProvenAbsent` only ever consulted the outbox table in both the oracle and this adapter (by design - they operate on "the outbox" as a concept, unlike `settleReceipt`), so no additional fix was needed there.

Residual (unchanged from round 1, noted in the PR body): the web's `#discardSupersededNoteRecovery` step is intentionally absent since the port itself doesn't declare it.

obra:
## Round 3 disposition (8d7732d)

One finding class (three Mediums, all `insert` upsert semantics) plus a Low, TDD against the IndexedDB oracle.

1. **Medium - enqueueIntent silently discards a duplicate mutation** (:282-286, member 0): the upsert only refreshed `state`/`attempted` on a `clientMutationId` collision, advancing the sequence and returning the new intent while the old payload stayed persisted. Fixed with a new `insertNew` (a plain `INSERT`, no `ON CONFLICT`) used only by `enqueueIntent`, so a collision throws and - because it's inside the existing `transaction()` savepoint - rolls back the sequence allocation with it, matching the oracle's `add`. Failing-first test: two `enqueueIntent` calls sharing a deterministic id; the second rejects and the sequence for its target ref never gets a row.
2. **Medium - insert leaves stale columns on conflict** (:279, member 1) and **Low - insert/enqueueIntent divergence from the oracle's add/put split** (member 2): split the shared `insert` into `insertNew` (above) and a new `replace`, which lists every column in its `ON CONFLICT DO UPDATE SET`, matching IndexedDB `put`'s full-row overwrite. `settleReceipt`'s optimistic write and `transferToRecovery`'s recovery write now go through `replace`. Failing-first test: seed a stale `mutation_optimistic` row directly (this table's id space doesn't legitimately repeat through the port's own call sequence today, so this simulates a retry/future caller), then `settleReceipt` on a fresh record with the same id and assert the stale payload/display are fully replaced, not left behind alongside a refreshed `state`.
3. **Medium - settleReceipt's optimistic promotion leaks recovery fields** (:237, member 1): `{ ...source, state: "accepted" }` carried a recovery-sourced receipt's `recoveryKind`/`recoveryReason`/`attempted` into the optimistic row. Now built field-by-field from the oracle's exact list (`version`, `clientMutationId`, `originClientId`, `targetRef`, `threadId`, `method`, `payload`, `attachments`, `optimisticDisplay`, `intentSequence`, `createdAt`, `state: "accepted"` - verified against `mutationOutboxIndexedDB.ts`'s own `settleReceipt` line by line, which also omits `composerText`). Extended the existing "settleReceipt consults recovery..." test to assert `recovery_kind`/`recovery_reason` are `null` on the promoted row.
4. **Low - markAttempted test doesn't cover what its name promises** (`mutationOutboxStorage.test.ts:143`): extended the test with the reviewer's suggested case (`markUnknown` to `blockedUnknown`, then `markAttempted` resolves `false` and `attempted` stays `0`). This one wasn't a production bug - round 2 already added the `AND state = 'submitting'` guard - just a coverage gap, so it doesn't falsify (already passed before this round; verified it's asserting real behavior, not vacuously).

Falsification: reverted the production file to its pre-round-3 state, ran the suite - the 3 load-bearing new/changed tests failed for the reasons above (duplicate enqueue silently succeeded, `settleReceipt` returned the old recovery fields, the stale row's payload survived) - then restored and confirmed green.

Gates: `npx vitest run src/mutationOutboxStorage.test.ts` (mobile-native) 18/18 pass, `npm run check` clean, `make lint-package-imports` PASS.

Size: 382 non-test lines (was 331). Under the 400 ceiling.

#1917 rebased onto this tip; see its own disposition for a same-class fix caught on its round 2 panel (`restoreProvenAbsent` wasn't savepoint-wrapped).

PR #1917, opened 2026-09-18
feat(native): MutationOutboxStorage read path over expo-sqlite (D25d-1b)

obra:
Completes the native SQLite mutation outbox adapter with ordered discovery/read methods, recovery transfer, and transactional restoration of records proven absent by an authoritative snapshot. Failed compound writes roll back without losing queued or recovery records, and the class now explicitly implements the shared storage port.

Rebased onto the merged write-path #1916 at main commit 617b188c8221c92f8a4c117a1ee49aac572a3757. Current head: 54cd2591736f42c7a48d5663a5d21829b3d50f97. Only the read/conformance commits were replayed; original 73fff461 is preserved in codex/backup-sdk-d25d-1b-read-73fff4614. The adapter remains unused until D25d-2 submission wiring.

Validation: 27 focused tests, native typecheck, package-import lint and diff checks passed. Independent spec/quality/simplify review and local RoboRev2551 passed the exact own patch. Restack preserves its full SHA-256 aea4b58127102b73bc161ee072dede792be3df5597c19dd4cfe8aa39b54febab and stable patch ID f1976c1a394ff3ee3f0b529b1fd278d538a54b96. The byte-identical refresh rule carries that review; new-head CI must settle before merge.

The raw panel's note-supersession finding is tracked in #1938 as a prerequisite before native notes support; no current native producer or approved D25d-1 through D25d-4 piece submits notes/human/set. Cross-table UUID collision behavior matches the web oracle and is tracked as shared hardening in #1957. Other conformance and cleanup follow-ups are #1927, #1928 and #1929. Own non-test scope: 100 changed lines.


obra:
## Round 2 disposition: rebased onto #1916's fixed tip (a522a78)

#1916's round 2 (write-path fixes: payload clientMutationId, SAVEPOINT-wrapped compound writes, settleReceipt consulting all three tables, expo-crypto-backed id source, empty-targetRef rejection, single conditional UPDATE/DELETE) - see the full disposition on #1916.

Rebased this branch onto the new #1916 tip. Own lines identical except a 7-line shrink: 1a's `settleReceipt` fix independently required widening `fromRow`/`get`'s generic bound from `MutationOutboxRecord<A>` to `MutationRecord<A>` (to accept `MutationOptimisticRecord`/`MutationRecoveryRecord` too), which this PR's original diff used to introduce itself for its own read methods - now inherited from #1916, no behavior change here.

Re-checked the read path against the oracle for the same class of finding (does `nextDispatchable`/`restoreProvenAbsent` consult all three tables the way `settleReceipt` needed to?): no fix needed. Both the oracle (`mutationOutboxIndexedDB.ts`) and this adapter scope `nextDispatchable`/`restoreProvenAbsent` to the outbox table only, by design - they operate on "the outbox" as a concept (what to dispatch next, what to reopen after an authoritative read), unlike `settleReceipt`/`settleApplied` which have to find a record that may have already moved to recovery or optimistic.

Gates: `npx vitest run src/mutationOutboxStorage.test.ts` (mobile-native) 21/21 pass, `npm run check` clean, `make lint-package-imports` PASS. Non-test size: 388 lines total in the file (1916's 331 + this PR's own ~57 read-path lines, roughly matching the original ~82-line estimate once the shared bound widening moved to 1916).

Head: a522a787f5d96edd69ddb1127124a49460804a95

obra:
## Round 2 disposition: rebased onto #1916's round 3 tip (8d7732d) + own fix (ee0013d)

Inherited from #1916's round 3 (see its disposition): `enqueueIntent`'s strict insert, `settleReceipt`/`transferToRecovery`'s full-row `replace`, and `settleReceipt`'s explicit accepted-record field list.

Own finding from this PR's round 2 panel (`reviews/raw/1917-a522a787f.md`, member 1): **Medium** - `restoreProvenAbsent`'s list-filter-update loop ran outside `transaction()`, unlike every other compound write in this file. Wrapped it in `this.transaction("mutation_outbox_restore_absent", ...)`. Failing-first test: two `blockedUnknown` records on the same target, a trigger fails the second's `UPDATE`, and the savepoint must roll back the first's reopen too - reverting the fix reproduces exactly that leak (first record ends up `submitting` while second stays `blockedUnknown`).

Checked the panel's suggested field list for `settleReceipt`'s accepted record against the actual oracle source (`mutationOutboxIndexedDB.ts` `settleReceipt`, read line by line, not from the panel's paraphrase): the panel's list includes `composerText`, but the oracle's own accepted-record construction omits it - confirmed by member 2's independent full-port review on the same panel, which found no issue with the (correct) omission. Implemented without `composerText`, matching the oracle exactly rather than the panel's summary.

Gates: `npx vitest run src/mutationOutboxStorage.test.ts` (mobile-native) 24/24 pass, `npm run check` clean, `make lint-package-imports` PASS.

Falsification: reverted just the `restoreProvenAbsent` change and confirmed the new atomicity test fails (first record's reopen survives the second's forced failure), then restored.

Size: 445 non-test lines total in the file (1916's 382 + this PR's own ~63 read-path lines).

Head: ee0013dc819e945dc878df6f3812b67ee680488f

PR #1885, opened 2026-09-18
feat(mobile): session usage totals from the package store (D18 B3)

obra:
Ruling: Jesse 2026-09-18, SDK migration D18 B3 (tiny-stacks-plan.md Stack B row 19) - the phone adopts the package store's turn-summed session usage totals, with the scope fallback, so web and phone show the same session figure.

## Before
mobile-native's transcript footer (`transcriptPresentation.ts`'s `accountingFor`, rendered by `TranscriptUsage.tsx`) read `conversation.usage` directly - `ThreadModel.usage`, the daemon's cumulative total only. A fork child's persisted meta never stamps `CumulativeUsage` (`agent/fork.go`'s `writeForkChild`), so on those sessions the footer showed nothing even though every loaded turn carried real per-turn usage. The web details panel (`DetailsPanel.tsx`) already covers this via the package's `sessionTokens(model)`, which falls back to summing the loaded turns and labels a truncated window `"tokens (loaded turns)"`.

## After
`accountingFor` now calls the same package function, `sessionTokens(conversation)` (`MobileConversation` is a `ThreadModel`), instead of reading `conversation.usage` directly. No second implementation of the summing/fallback logic - it's consumed from `appwire-client/typescript/threadUsage.ts`, untouched by this PR. `SessionAccounting.usage` carries the `SessionTokens` shape (`inputTokens`/`outputTokens`/`scope`) plus the cache/total breakdown, which only ever comes from the thread's own cumulative usage (no per-turn equivalent exists) and is absent whenever the fallback fired. A new `tokenUnitLabel` pure function names the figure the same way the web panel does: `"tokens"` for a whole-session total, `"tokens (loaded turns)"` when `sessionTokens` scoped the sum to a truncated window (`olderCursor` present).

## Tests (mobile-native/src/transcriptPresentation.test.ts)
- `falls back to summing the loaded turns when the thread has no cumulative total` - new, watches the fork-child gap close.
- `labels a derived total over a truncated turn window as covering only the loaded turns` - new.
- `labels a whole-session total plainly` / `labels a sum truncated to the loaded turns as covering only the loaded turns` / `labels the absence of any token figure plainly` - new, unit tests for `tokenUnitLabel`.
[...]

PR #1919, opened 2026-09-18
fix(mobile): the conversation keeps its turns and wire cursor coherent across paging and rehydrate

obra:
Decomposed out of #1885 after five review rounds (Jesse's "five review rounds means decompose" rule). This PR is the state-layer mechanism only: `mobile/src/state/conversation.ts` and `mobile/src/services/conversation.ts`'s turn plumbing. #1885's footer half is B3a, stacked on this PR.

## What this closes

The store's own top-level `olderCursor` (a UI-only, intentionally capped "is there another page to fetch" signal, F8) and the conversation's own `ThreadModel.olderCursor` (the wire truth `sessionTokens` reads) are two different values that #1885's review kept finding collapsed into one at different sites, across five rounds:

- **loadOlder** derives `conversation.olderCursor` from the wire's own `nextCursor`, never the store's capped field (so the item cap stopping the UI's own paging never mislabels a partial sum as a full session).
- **rehydrate** preserves page-loaded turn history across a same-session reread: its own itemLimit-bounded window can't see turns loaded by an earlier `loadOlder`. Turn ownership is tracked separately from item ownership (`pageOwnedTurnIds`, alongside the existing `pageOwnedIds`), so a page whose display rows were all deduped or evicted still keeps its turns for the merge.
- **rehydrate's cursor** carries the prior conversation's own wire cursor only when the merge actually contributed a turn beyond the fresh read's own window - not whenever page history merely exists somewhere in the session's lifetime.
- **Both merges** (loadOlder's and rehydrate's) go through the package's own `mergeOlderItemPage` (`turnsMatch`/`mergePageTurn`) instead of an id-only filter: `thread/turns/list` is itself item-paginated, so a turn can be split into fragments across a page boundary, or two turns can be logically the same but carry different ids. The service now exposes the wire `ThreadTurnsListResponse` (`turnsPage`) from both `loadOlder` and `readProjection` so the store folds it through the package's real, identity-aware merge - never a second, hand-rolled one.

## Per-state table (a session with no thread-level cumulative usage, so sessionTokens is always summing turns)

| state | store cursor | conversation cursor | turns | scope |
|---|---|---|---|---|
| initial (open) | `null` (F8's live path establishes it separately) | `"cursor-1"` (from the hydrated model) | `[t2]` | `loaded` |
| loadOlder (wire has more) | `"cursor-2"` | `"cursor-2"` (= wire `nextCursor`) | `[t1, t2]` | `loaded` |
| cap hit (wire still has more) | `null` (F8 stops paging honestly) | `"more"` (unaffected by the cap) | `[t1, t2]` | `loaded` |
| rehydrate, page history kept, contributes beyond the fresh window | unchanged | prior conversation's own wire cursor | merged (identity-aware) | `loaded` |
| rehydrate, page history exists but the fresh window already covers it | unchanged | fresh read's own | fresh reread's own | per fresh read |
| rehydrate, no page history | fresh read's own | fresh read's own | fresh reread's own | per fresh read |

A test per row lives in `mobile/src/state/conversation.test.ts`'s two `describe` blocks ("D18 B3 round 5" and "D18 B3 round 6").

## Failing-first tests (watched fail against `origin/main`, pass with the fix)

- `merges the older page's turns into conversation.turns and advances conversation.olderCursor`
- `keeps conversation.olderCursor at the wire's cursor even when the item cap stops paging`
- `rehydrate preserves the older turns loaded via loadOlder`
- `cap hit then rehydrate: conversation.olderCursor stays the wire truth, not the store's capped null`
- `evicted/filtered page: a turn survives a rehydrate even when none of that page's items did`
- `a turn split across a page boundary keeps its usage instead of being dropped by an id-only filter`
- `rehydrate with page history but a fully-covering fresh window: scope follows the fresh read, not a stale accumulated cursor`
- `returns the older page's wire turns alongside items`

## Falsification

Reverted `mobile/src/services/conversation.ts` and `mobile/src/state/conversation.ts` alone to `origin/main`, ran `mobile/src/state/conversation.test.ts` and `mobile/src/services/conversation.test.ts`: 8 failed exactly as above. Restored from the saved diff; `git status --short` showed only the 4 intended files.

## Gates

- `npx vitest run --root .. --config mobile-native/vitest.config.mts mobile/src/state/conversation.test.ts mobile/src/services/conversation.test.ts`: 485 passed.
- Full `mobile-native` suite (`npx vitest run`): 798 real tests pass (3 pre-existing `react-test-renderer`-missing file-load failures, present on `origin/main` before this PR too).
- Shared `mobile/src` suite (`npm run test:shared`): 801 pass.
- `npm run check` (tsc, mobile-native): same 4 pre-existing errors as `origin/main`.
- No `appwire-client/typescript` or `cmd/evener-hub/frontend` files touched.

## Diff

```
```
84 non-test lines in `mobile/src/state/conversation.ts` + 18 in `mobile/src/services/conversation.ts` = 102 non-test lines (target 80-150, ~180 ceiling if the merge reuse needed it - it didn't need the extra room).

Head SHA: d07d60263c4d0997d08cf45c2c24de4bfcaaa87c


obra:
@/private/tmp/claude-501/-Users-jesse-git-prime-radiant-inc-evener--claude-worktrees-issue-1116-merge-review-1fc3f3/8d47e620-e3d6-4079-b4ee-bec5c59c7f43/scratchpad/round2-comment.md

obra:
Refreshed onto `origin/main` after the round 2 push (6 commits ahead, mostly unrelated Go/appwire-client/mobile-native changes) — one trivial add-add import-list conflict in `mobile/src/state/conversation.ts` (this branch added `mergeOlderItemPage`, main's #1895 added `joinWarningParts`), resolved by keeping both alphabetically. `git diff -M --stat origin/main...HEAD` is unchanged (same 4 files, 730/-2) confirming no drift. Targeted tests (362) and services tests (126) still green; `npm run check` at the same 4 pre-existing errors. New head: `012f688` (012f6885288f196bdd73a379a2f89a1b20ce61ac).

obra:
## Round 2 disposition addendum (d4f6385)

Same class, closed in-round on the item side per the coordinator's call (`preservePageHistory`, `conversation.ts` ~1735-1737, was the same `entryLoadOlderToken !== loadOlderToken` disjunct as the `preserveTurnHistory` fix above): a failed `loadOlder` bumps the token without adding to `pageOwnedIds`, so it could still force `preservePageHistory` true and pin the store's own paging cursor (F8) at its stale pre-race value through a racing rehydrate. Dropped the disjunct; `preservePageHistory` now gates on `pageOwnedIds.size > 0` alone, matching the turn-side fix.

Failing-first test: rehydrate held open via `readProjectionBlock` while a racing `loadOlder` fails, then released with a fresh read reporting no further history (`olderCursor: null`) — asserts `store.getState().olderCursor` follows the fresh signal, not the stale pre-race cursor. Falsified: reverted this one line, ran the new test alone — failed with `expected 'cursor-1' to be null`; restored, green (the other two round-7 tests stayed green throughout, confirming this is an isolated line change).

Gates: targeted vitest — 363 passed (was 362); services — 126 passed; `npm run check` — same 4 pre-existing `react-test-renderer` errors, no new ones.

`git diff -M --stat origin/main...HEAD`: 4 files, 795 insertions(+)/3 deletions(-) (same file set, +65/-1 on top of the previous push).

New full head: `d4f63854041ece76aa4ebb51c0e2af9cabb53872`.

obra:
@/private/tmp/claude-501/-Users-jesse-git-prime-radiant-inc-evener--claude-worktrees-issue-1116-merge-review-1fc3f3/8d47e620-e3d6-4079-b4ee-bec5c59c7f43/scratchpad/comment-fix.md

obra:
Disposition for the retained-turn-growth Medium (all 3 reviewers, repeat finding): landing per Jesse's explicit ruling (2026-09-22): land now, bound in the immediate follow-up PR. The follow-up bounds retained page-turn data to a keep-window around the display set (compact identity + usage metadata beyond it), prunes pageOwnedTurnIds with the same bound, and adds the repeated loadOlder/rehydrate bounded-retention test this review asked for. sessionTokens' turn-summed accounting stays complete - every loaded turn keeps identity + usage. The two pinned tests asserting full-payload retention are reconciled to the bounded policy in that PR.

PR #1920, opened 2026-09-18
feat(mobile): session usage totals from the package store (D18 B3)

obra:
Stacked on #1919 (B3b — the state-layer mechanism this footer reads: `conversation.turns`/`olderCursor` kept coherent across paging and rehydrate). Decomposed out of #1885 after five review rounds, per Jesse's ruling ("five review rounds means decompose", ruling 6 in the mobile landing queue's common brief) — closes nothing else.

## What this is

The phone's transcript footer reads the same turn-summed session usage totals the web details panel reads, with the same scope fallback, so both surfaces show the same session figure for a session with no thread-level cumulative usage (a fork child never gets one stamped, per `agent/fork.go`'s `writeForkChild`).

- `accountingFor` derives the footer's token total from the package's `sessionTokens(conversation)` instead of reading `conversation.usage` directly.
- The thread's own cumulative cache/total breakdown (`cacheReadTokens`/`totalTokens`) is read independently of the derived input/output pair and its scope — it's always a whole-session figure with no per-turn equivalent, so it survives when `sessionTokens` falls back to summing turns or returns `null` outright, and never inherits whatever scope the derived pair got.
- `usageRows` picks the footer's visible rows and gives each its own unit, so a turn-summed "loaded" pair can't leak its label onto an unrelated whole-session cache/total figure.
- A field's Go zero value is treated the same as absent (`cacheReadTokens`/`totalTokens` of `0`), matching `sessionTokens`'s own rule for input/output tokens.
- `tokenUnitLabel` (the `SessionTokens` scope-to-label mapping) is shared, in `appwire-client/typescript/threadUsage.ts` — the web details panel and the native footer render the same `"tokens (loaded turns)"` / `"tokens"` wording from one function. `DetailsPanel.test.tsx` is unchanged and still passes — the rendered string never moved.

## Falsification

Reverted the 6 production files (`appwire-client/typescript/index.ts`, `threadUsage.ts`, `DetailsPanel.tsx`, `detailsAccounting.ts`, `TranscriptUsage.tsx`, `transcriptPresentation.ts`) alone to `origin/main`, ran `mobile-native/src/transcriptPresentation.test.ts`: 10 failed. Restored from the saved diff; `git status --short` showed only the 8 intended files.

## Gates

- `npx vitest run src/transcriptPresentation.test.ts` (mobile-native): 36 passed.
- Full `mobile-native` suite (`npx vitest run`): 807 real tests pass (3 pre-existing `react-test-renderer`-missing file-load failures, present on `origin/main` before this PR too).
- Shared `mobile/src` suite (`npm run test:shared`): 790 pass.
- `npx vitest run ../../../appwire-client/typescript/threadUsage.test.ts src/panes/session/chrome/DetailsPanel.test.tsx` (frontend): 42 passed.
- `npx biome ci ../../../appwire-client/typescript` and `npm run lint`: clean.
- `npm run typecheck` (frontend) and `npm run check` (mobile-native, tsc): same pre-existing error counts as `origin/main`.

## Diff

```
```
Non-test lines: 2 (index.ts) + 8 (threadUsage.ts) + 7 (DetailsPanel.tsx) + 6 (detailsAccounting.ts) + 24 (TranscriptUsage.tsx) + 70 (transcriptPresentation.ts) = 117, within the 80-150 target.

Head SHA: d40bf549bf3273d4154dae3eb668da692cc089b2


obra:
## Round 1 disposition (d40bf54)

The one Medium (`transcriptPresentation.ts:227-229`: older pages are projected into `conversation.items` without updating `conversation.turns`, so the summed total stays at the first window) is exactly what #1919 (B3b, the state-coherence half this PR is stacked on in review order) fixes: `loadOlder` and rehydrate now merge older-page turns through the package's `mergeOlderItemPage` and keep the wire cursor coherent, with a per-state table and a test per row. This PR is the footer half only and merges after #1919. No change here.

obra:
Dispositions the panel's single Medium per Jesse's ruling: merge, with the fix tracked in #2139.

**Fork-child token attribution** (mobile-native/src/transcriptPresentation.ts:227, 1 of 3 reviewers): when cumulative usage is absent, the footer's fallback sums every loaded turn; fork children copy the parent transcript prefix including per-turn usage, so an untouched fork's footer can attribute the parent's spend.

Why dispositioned rather than fixed here:
- Display-only accounting edge; fires only when cumulative usage is absent (older servers or sparse data). The primary cumulative-preferred derivation is unaffected, and the total self-corrects once cumulative usage is present.
- Both reviewer-prescribed fixes are protocol-level (server-provided divergence-scoped total, or exposing the fork boundary to the client) — outside this PR's wire surface.

Follow-up: #2139.

PR #1921, opened 2026-09-18
fix(agent): stop retirement eligibility checks racing the async session namer (#1879)

obra:
The first turn can finish while its background session namer still holds an autonomous retirement blocker. Test callers that expect a fully settled session now wait for the existing sender wait group before claiming retirement, through one shared helper.

A regression exercises the real session, scripted namer, and retirement controller inside `testing/synctest`. It proves the claim stays blocked while the namer is held, then succeeds after release. Removing only the helper wait produces the expected autonomous-blocker failure. This PR changes one test file and no production behavior.

Validation at `d10c70c4c400e63d3db427db56f4df7acbf526db`:
- Focused normal, race, and evenerfuzz-tagged regression tests pass; the test deadline audit passes.
- Normal/tagged/Windows vet, Go formatting, and golangci-lint pass after merging current main.
- Independent spec, quality, and simplification review passed.
- Local `roborev review --branch --wait --base origin/main` passed with no findings (job 2537).

Current-head CI and the raw remote review panel remain required before merge. The separate `TestRetirementTreeSettleDrainsPendingRootAttention` failure remains tracked in #1879; this PR does not close that issue.

Merge gate at `d10c70c4c400e63d3db427db56f4df7acbf526db`: all15 current-head checks passed. Raw panel members Luna, Muse, and DeepSeek each report no issues. The GLM member produced no review because its provider exhausted HTTP429 retries; that unavailable member is recorded, not counted as a pass. Independent review/simplification and local branch review2537 passed.


obra:
## Round 2 disposition (6953aa1)

**(1) Test didn't deterministically fail without the fix.** Confirmed: `close(release)` ran before `assertRetirementEvidenceEligible(t, c)`, so the namer's release and the helper's read raced each other instead of being ordered. Restructured `TestRetirementEvidenceEligibleAwaitsInFlightNamer`: the helper now runs in its own goroutine while the namer is still held, and non-completion is read via a non-blocking `select`/`default` on a completion channel — sound because `sendersWG`'s count for the namer provably cannot be zero while it's parked on `<-release` (a `sync.WaitGroup` invariant, not a timing guess). Only after `close(release)` do we await completion (bounded by a `// TRIPWIRE:`-marked hang guard, per the bare-wall-clock audit's sanctioned pattern — not the mechanism). Falsified: reverting the helper's `sendersWG.Wait()` now fails this test 20/20; with it restored, 20/20 pass.

**(2) CI's 'tests' job at 2e7c2fe0a.** Run [35397613924](https://github.com/prime-radiant-inc/evener/actions/runs/35397613924/job/105769928476) failed `TestRetirementSafetyDetachedLifetime` in the shard-cost survey pass (no inline assertion text — `survey.log` isn't uploaded as an artifact). Real, not unrelated: that test checks eligibility with a **raw** `c.TryClaim(true)` call ("independent detached lifetime blocked eligibility") right after root's first turn, not through `assertRetirementEvidenceEligible` — round 1's fix to the shared helper never covered this second call site in the same test. Same in-flight-namer mechanism as round 1, a different call site. Fixed with the same `root.sendersWG.Wait()` pattern, plus a new deterministic regression (`TestRetirementDetachedLifetimeEligibilityAwaitsInFlightNamer`, same construction as (1)) that fails 20/20 without the wait and passes 20/20 with it.

Both fixes are test-only, 0 non-test lines (diff: 100 insertions / 3 deletions, one file).

**Gates:** gofmt clean, `go vet ./agent/...` clean, `go vet -tags evenerfuzz ./agent/...` clean, `GOOS=windows go vet -tags evenerfuzz ./agent/...` clean, `golangci-lint run ./agent/...` 0 issues, `TestNoBareWallClockDeadlineInAgentTests` passes, `go test -count=20 -race -run '^(TestRetirementEvidenceEligibleAwaitsInFlightNamer|TestRetirementDetachedLifetimeEligibilityAwaitsInFlightNamer|TestRetirementAutonomousNamingOverlap|TestRetirementAttentionArmRetryPending|TestRetirementSafetyColdJobWatchContent|TestRetirementSafetyDetachedLifetime)$' ./agent/` — ok, 0 failures.

Full head: `6953aa1c19f07bba59d9a0e7ec62fea384ea816f`

obra:
## Round 3 disposition (0a08b4d)

**(1) Vacuous-pass risk.** Confirmed: the immediate non-blocking select could pass simply because the spawned goroutine hadn't been scheduled yet, proving nothing. Added a `started` channel closed by the goroutine as its first statement, awaited (TRIPWIRE-bounded) before reading, plus a `runtime.Gosched()` loop (no wall-clock literal — no duration value at all) that gives a buggy claim every chance to race ahead first. Reworded the comment to stop claiming that read is "proof by construction": it's now documented as a best-effort diagnostic, and the doc comment identifies the actual proof — the second, post-`release` check, which is sound because `resultCh` is buffered: whatever it holds is fixed at the instant the guarded call actually ran, independent of when this goroutine reads it, and `sync.WaitGroup`'s own invariant guarantees a correct implementation's `sendersWG.Wait()` cannot return before `release` closes.

**(2) `t.Fatal` from a spawned goroutine.** Confirmed: `assertRetirementEvidenceEligible(t, c)` ran inside the goroutine. Extracted `retirementClaimAfterFirstTurn(root, c) (*RetirementClaim, RetirementSnapshot, error)` — the wait-then-`TryClaim` pair, no `*testing.T` dependency — so only that runs on the goroutine; the result goes over a channel and every assertion now runs on the test goroutine.

**(3) Unbounded `<-held`, un-cleaned `release`.** Bounded `<-held` with the same TRIPWIRE-comment hang-guard pattern already used elsewhere in this file. `release` now closes through a `sync.Once`-guarded closer registered with `t.Cleanup`, so an earlier `Fatal` can never leave the real namer goroutine parked forever.

**(4) Second regression didn't test the real call site.** Confirmed and it's a good catch: it re-implemented `sendersWG.Wait()+TryClaim` in its own goroutine, disconnected from `TestRetirementSafetyDetachedLifetime`'s actual check — reverting the wait at that call site could not have failed it. Took the "extract a helper both use" option: `retirementClaimAfterFirstTurn` (from finding 2) is now the one function `assertRetirementEvidenceEligible`, `TestRetirementSafetyDetachedLifetime`'s real call site, *and* the regression test all call. The two former regressions collapse into one, `TestRetirementClaimAfterFirstTurnAwaitsInFlightNamer`, exercising that shared function directly.

**Falsification:** reverting `retirementClaimAfterFirstTurn`'s `sendersWG.Wait()` (the one production-side fix location) makes the regression fail 20/20 — every time, caught by the first, specific check ("claim settled while the namer was still held in flight"), not a fallthrough. Restoring it passes 20/20. `TestRetirementSafetyDetachedLifetime`, `TestRetirementAttentionArmRetryPending`, `TestRetirementSafetyColdJobWatchContent`, and `TestRetirementAutonomousNamingOverlap` all still pass (10 iterations each) with the fix restored.

Net diff for this round: -15 lines (one test absorbed into the other), still test-only, 0 non-test lines.

**Gates:** gofmt clean, `go vet ./agent/...` clean, `go vet -tags evenerfuzz ./agent/...` clean, `GOOS=windows go vet -tags evenerfuzz ./agent/...` clean, `golangci-lint run ./agent/...` 0 issues (fixed a `modernize` finding on the Gosched loop along the way), `TestNoBareWallClockDeadlineInAgentTests` passes, `go test -count=20 -race -run '^(TestRetirementClaimAfterFirstTurnAwaitsInFlightNamer|TestRetirementAutonomousNamingOverlap|TestRetirementAttentionArmRetryPending|TestRetirementSafetyColdJobWatchContent|TestRetirementSafetyDetachedLifetime)$' ./agent/` — ok, 0 failures.

(A full `-run '^TestRetirement'` pass on this dev box still shows the same 4 pre-existing, unrelated environmental failures noted in round 1 — a macOS `/var`→`/private/var` symlink sandbox-policy rejection and a git-worktree-porcelain scan issue on this heavily shared, multi-agent repo — in different files, on code paths this change never touches.)

Full head: `0a08b4d168bfdc691c5dfefae09d11b66f17b66b`

PR #1922, opened 2026-09-18
feat(mobile): retained screens recover on reconnect (D28 D2a)

obra:
Retained native screens automatically reload missed state when their connection becomes ready again. Fatal failures keep the connection wall until a replacement client is ready. Provider and marketplace form modals expose reconnect actions without losing their drafts, and disconnected writes are distinguished from a busy mutation gate.

Stacked on #1952 and #1955. Refreshed head `193b2e96ab347d06bc98775ee6d1999baabe86cb` preserves the complete reviewed recovery patch byte-for-byte, SHA-256 `105dd4c055de2abbf06c85c9a445c523054c80a10f565f9d2dccc14199603683`. The unrelated WIP test-deletion branch remains preserved and unapplied.

Validation: 64 focused native tests, native typecheck, package import lint, and diff checks passed after refresh. Independent review/simplify and local RoboRev2569 are qualified with only separate Lows. Raw-panel AddMarketplace not-ready closure is unreachable through the synchronous outer/inner readiness guards; mount/selection read errors are transient and recover through the existing wanted/catalog lifecycle. Defensive handling and diagnostic improvements remain in #1942. New exact-head CI is required before landing.

The current base includes merged #1975, which fixes the independently reproduced retirement replay-test race. This refresh changes no owned production or test behavior. Complete raw reviewer-member qualification remains required; a queued synthesis alone is not a gate.


obra:
## Round 2 disposition (a2fd89ae6)

Rebased onto #1915's new tip (a48f7e0c8) first - own lines verified byte-identical before/after (`diff <(git show <old-D2a>:<file>) <(git show <rebased-D2a>:<file>)` for every touched file showed only #1915 round 3's own additions, none of D2a's own lines moved or dropped). Two conflicts during the rebase (`HubSettingsScreen.tsx`'s import list, `connectionDisplay.test.ts`'s test list) - both resolved by keeping both sides' additions, nothing dropped.

Panel at a17b60a88, members 0-2. Member0's third finding and all of member1's/member2's findings are #1915's round 3 (Providers/MarketplaceBrowser readiness gating) - already fixed and disposed there, not repeated here. The two real findings are D2a's own:

**(1) Medium - a retry's not-yet-ready replacement client got adopted immediately** - fixed. `useRenderClient` (`connectionDisplay.ts`) now takes the connection `state` alongside `client` and only adopts a new client once `state === "ready"` for it; every other transition - the null gap a manual retry opens AND the whole `"connecting"` window before the replacement dials - keeps returning whatever was last adopted. Measured the web's own answer before picking this fix over "gate the fetch effect instead": `cmd/evener-hub/frontend/src/shell/ConnectionBanner.tsx` only calls `onClientReplaced` (which swaps `AppShell`'s `ClientProvider` slot) *after* `await fresh.connect()` resolves and confirms the fresh client isn't closed - never on construction. Holding the previous client is the more general fix: it also means `PluginsScreen`/`HubSettingsScreen`'s mount effects need no readiness check of their own, since the store they build from `renderClient` is never rebuilt until the replacement is actually usable.

**(2) Medium - MarketplaceBrowser retained across reconnects but never told to recover** - fixed. Added a `connectionState: ConnectionState` prop (threaded from `PluginsScreen`'s own `connectionState`), and wired `model.connectionChanged(client, connectionState)` in an effect declared BEFORE the mount-fetch effect - the same ordering PluginsScreen's own installed-list wiring uses and for the same reason (measured in #1915 round 1: reversed, the mount effect's `fetchMarketplaces()` marks its read live before this effect runs, and connectionChanged sees that and refetches a second time for the same first load).

Falsified both (remove the fix, run the test, confirm it fails for the stated reason, restore, confirm green):
- `connectionDisplay.test.ts`'s new "a retry's not-yet-ready replacement never displaces the previous client" test (a fake client going `connecting -> ready`, per the brief): reverting `useRenderClient` to the old `client ?? lastClient.current` fails at the "replacement now exists, but has not reached ready" assertion - it returns the not-yet-ready client instead of the previous one.
- `PluginsScreen.test.tsx`'s new "MarketplaceBrowser re-reads its list once the connection returns to ready after a flap" test (mounts the real screen, switches to the Browse tab, flaps the connection): without the `connectionChanged` wiring, `evener/marketplace/list` is called once instead of twice.

Gates: `mobile-native && npx vitest run` - 87 files / 834 tests, all green. `npm run check` clean. Root `make lint-package-imports` PASS.

`git diff -M --stat` against #1915's tip (a48f7e0c8), D2a's own content:
```
 mobile-native/src/HubSettingsScreen.tsx     |  28 +++++--
 mobile-native/src/MarketplaceBrowser.tsx    |  21 ++++-
 mobile-native/src/PluginsScreen.test.tsx    | 114 ++++++++++++++++++++++++++++
 mobile-native/src/PluginsScreen.tsx         |  29 +++++--
 mobile-native/src/ProvidersScreen.tsx       |   2 +-
 mobile-native/src/connectionDisplay.test.ts | 106 ++++++++++++++++++++++++--
 mobile-native/src/connectionDisplay.ts      |  77 ++++++++++++++++---
 mobile-native/src/renderNative.testkit.tsx  |  34 +++++--
 8 files changed, 374 insertions(+), 37 deletions(-)
```
Non-test: 159 insertions(+), 32 deletions(-) - under the 400-line ceiling, no split needed.

Head: `a2fd89ae6b01e77c10f3dfe26970b8ff720ba962`.

Issue #1942, opened 2026-09-19
mobile: consolidate repeated retained-screen connection wiring


obra:
The simplify reviews of #1915/#1922 identified repeated native store connectionChanged wiring, hub-scope reset bookkeeping, readiness-edge recovery callbacks, and connection fixtures in screen tests.

After the replacement readiness/banner stack and #1922 land, measure which duplication remains and extract the smallest helpers where the lifecycle contracts actually match. Preserve live hub/client identity checks, protocol-close walls, recoverable-flap banners, initial ready reads, automatic reconnect recovery, and offline Cancel/navigation.

Do not unify web/native semantics merely because the code looks similar or add a new lifecycle framework. The active readiness and fatal-close defects remain in the migration stack; this tracks the accepted non-blocking cleanup.

The independent reconnect-readiness C review at 4bf782b77e103c84864eedb8677520ff1de85c0c records two additional Low simplifications: a redundant ready dependency where the live predicate already carries authority, and repeated readiness checks around mutation entry. Preserve invocation-time client/hub identity checks and the actual deferred-confirmation regressions; do not replace live reads with stale captured booleans.

The independent recovery review at affd280cb also records Low follow-ups: ProviderEditor.save uses a disabled snapshot rather than an invocation-time readiness guard; fatal walls omit the compatibility error and offer an ineffective retry; AddMarketplace needs explicit not-ready outcome handling; the preserved tautological test may be removed only with its meaningful replacement coverage established. The original WIP test-deletion branch remains preserved. Modal-hidden reconnect status is a current must-fix in the recovery PR and is not deferred to this issue.

Local RoboRev2569 at57bfcbf15 adds a Low presentation follow-up: plugin-details Modal still covers the reconnect banner, so users dismiss details to reach retry. Unlike the editable provider/marketplace forms, no unsaved form input is lost; those form modals are fixed in #1922. Render the existing ConnectionStatus inside details and cover a disconnect while open. Also update the stale three-outcomes test comment to the current four-outcome contract.

Additional Low from the raw #1922 panel at 57bfcbf15: PluginsScreen.act and MarketplaceBrowser.act clear notices/errors before the live readiness check. A disconnected no-op should preserve existing diagnostic copy; test the blocked action with a pre-existing notice. Keep separate from the Lows-only qualified recovery PR.

[...]

PR #1952, opened 2026-09-19
mobile: retain loaded screens behind reconnect banners

obra:
Loaded native Providers, Plugins, and Hub Settings screens retain their content through recoverable connection interruptions and show a reconnect banner. Initial and fatal connection failures retain the blocking connection view; changing hubs resets retained client ownership.

This replaces the screen/banner portion of #1915. Required successors are #1955 for invocation-time mutation readiness and #1922 for automatic read recovery and retained form retries. Those corrections are locally implemented and independently reviewed; merge order remains #1952 -> #1955 -> #1922.

Refreshed onto main `323f28c30536c3a07f21538128cdad1e3afbb607` at head `6dacb1999d8d658803fb96e8b1bd4d4f5d310114`. The complete owned binary patch is identical to reviewed local `51bb119c`, SHA-256 `3266cc0e29cf50fcf0dcef63bcdf38c87c27b452bec34ca3eba005676ee3d76a`. Main's intervening mobile changes are file-disjoint. Declaration/conflict-marker and diff checks passed.

The integrated reconnect stack passed 104 focused tests, native typecheck, import lint, independent correctness/simplify review, and local RoboRev2621. Provider stale-save and AddMarketplace not-ready tests were falsified against their production guards. This refresh changes no owned behavior. Fresh current-head CI and raw reviewer qualification are pending. Separate Lows remain in #1942.


obra:
Review Medium (config editor saves assert the live row's fingerprint, so a stale draft can never trigger the destination-changed refusal) — **confirmed and fixed** in 7b1f6fa982. The editor now captures the endpoint fingerprint at open time and asserts that on save, routing refusals through the existing endpointAsserted → ENDPOINT_CHANGED_WARNING path, exactly like the credential editor's established credentialTarget pattern.

Pinned red-first: open the editor while the row says fp-work; a same-client flap and recovery brings back the row at fp-moved; save — RED before the fix: the request carried fp-moved (read from the live row), bypassing the refusal; GREEN after: the request carries fp-work, the hub's refusal closes the editor, the list re-reads, and the destination-changed warning shows. An unchanged endpoint across a flap still saves with no warning — no over-fencing.

One further review observation is acknowledged and tracked rather than changed here: the reconnect banner renders outside OS-level modals (ProvidersScreen/PluginsScreen), so a user inside an editor sheet must dismiss it to see connection status or reach Reconnect. That is a banner-design decision rather than a correctness fix — the stale-draft and destination-assert protections above are what keep the modal window safe — and it joins the deferred modal-UX follow-ups (duplicate reconcile-on-mount, unused import, sign-in idle-stick) in the immediate follow-up PR after merge.

obra:
Both Mediums from the re-review on `7b1f6fa` — **confirmed and fixed** in 1d7c19ff2d:

**Offline sign-in start hands the flow a stale client** — `onSignIn` now applies the same ready-guard the rebind effect uses (`state === "ready" ? client : null`), and a sign-in started from behind the banner (whose first `start()` was a no-op) now resumes when the connection it was opened without arrives — re-driven only for an idle flow, and gated on the store's stale-listing flag, because a manual retry turns the new connection ready before its own listing read lands and a device start issued in that window would be refused as a stale-listing write and strand the exchange in its error phase. Red first: the flow was handed the raw client while offline; green: null offline, the client on recovery, and the exchange proceeds to the device-code phase.

**Reconnect banner hidden behind native modals** — the extracted `ConnectionStatus` is now embedded in all four modal sheets (provider editor, plugin details, add-marketplace, sign-in sheet), gated on the connection not being ready; a ready modal renders exactly as before. Red first: no status or Reconnect reachable inside any modal subtree during a non-ready connection; green: status text and a working Reconnect are present, so recovery no longer requires dismissing the modal and discarding the draft.

Full gates green on the pushed head (scoped 42 tests; full native 999 tests + 799 shared; tsc; bundle); local roborev converged to clean after fixing a test-order dependence in the new tests themselves.

Issue #1959, opened 2026-09-19
Cover late marketplace removal outcomes across page lifetimes


obra:
The independent web consumer review at de0e96dbeeda548c99c4a3da7430c7c5b1217088 found no must-fix, but the existing tests do not directly deliver a typed applied-removal outcome after a client replacement or after the whole settings page unmounts.

Add focused behavior coverage for both lifecycle boundaries. Verify an old-client outcome cannot mark or block the new client, a sheet remount retains the current page/client no-repeat marker, and global warning behavior remains truthful. Preserve ordinary retry semantics and the existing SDK generation/list-revision fences. Reuse the current page test harness; avoid rendered-markup snapshots.

This is a Low follow-up to the marketplace warning consumer, separate from the current must-fix work. Related protocol boundary coverage: #1944; applied-array member validation: #1953.

Native disposed-store investigation at 54c661759 found a stronger, current must-fix scenario: browser A unmounts while removal is pending, browser B remounts and its initial list fails, then the late unavailable outcome fetches through disposed A and cannot publish to B. This is being corrected in the active native consumer and must not be deferred as a Low or described as merely redundant traffic. The original web late-outcome coverage follow-up above remains separate.

At native head 67b5b90a3620883735914dc2657e7854679a8fb8, the unavailable-outcome case is corrected by a client-scoped current-browser reconciliation signal. Local RoboRev2591 and independent review found the analogous valid-payload case: the authoritative applied list is written only into the disposed originating model, so a remounted browser whose initial read failed never receives it. This is an active P2, not a deferred Low; native outcome/warning and model-lifetime reconciliation are now being separated into meaningful review slices before correction.

obra:
Cross-reference from #2072's review: another slice of the native model-lifetime family this issue tracks. When a marketplace removal the hub reports as already applied (`marketplaceRemoveApplied`) is followed by a *failed* reconciling refetch, the stale list keeps the row and its Remove action, so a retry surfaces the hub's ordinary not-found failure for an already-removed marketplace. The consumer classification itself is correct and web-parity is maintained (#2068's sheet behaves the same); the durable fix is the pending-removal tracking until the authoritative list drops the name — out of scope for #2072 per its disposition comment, where the revision-guard constraint (reconcile must not fire through a disposed browser) is also recorded.

obra:
[...]

PR #1960, opened 2026-09-19
Preserve applied marketplace removals across web reconciliation

obra:
The web marketplace UI preserves cleanup warnings when removal applied but clone deletion failed. A client-scoped guard prevents repeating that removal until the SDK accepts a newer authoritative list, including when another client re-adds the same name. Failed or held snapshots cannot release the guard; late outcomes from a replaced client cannot affect its replacement.

This adopts the merged SDK cache and publication contracts (#1954, #1973). The change adds the shared removal-outcome classifier, uses it in the web marketplace page and sheet, preserves guards across sheet remounts, and keeps accepted-list cache retirement intact. The separate no-litter outcome introduced by #2050 remains a focused successor.

Validation at `a6e134f2125166b654fbabe669c09dc00bd8cc2f`, based on `5dfd06d299a60f3919e17b4d546589c5b2c92860`:
- 94 focused marketplace and SDK tests passed.
- Full `make test-web` passed (typecheck, tests, Biome).
- RoboRev exact-base branch review 2706: no issues found.
- Root review confirmed accepted snapshot publication, cache retirement, and loading-state semantics are preserved.

Retained #1959 includes additional lifecycle coverage and native obligations; this PR does not claim that whole issue is resolved.


obra:
Review dispositions for the current-head combined review (a6e134f):

- **Medium (native `MarketplaceBrowser.tsx` applied-unavailable handling)**: retained on the native marketplace chain, not blocking this web PR — native treats the applied-unavailable outcome as an ordinary failed removal today, which is pre-existing behavior this PR does not touch. The native chains (#1972 → #1978 → #1983) and #1959's native lifecycle obligations own exactly this correction; this PR's shared classifier is the prerequisite for it.
- **Low (MarketplaceSheet test names assert the opposite of their names)**: real; will be fixed in an immediate follow-up PR (rename to the verified behavior — the reconciliation fetch clears the retry guard on the newer accepted publication).
- **Low (removed held-applied-write ordering coverage)**: real; the held-applied-then-newer-success `listRevision.publishOwner` path will be restored in the same follow-up PR.

PR #1950, opened 2026-09-19
native: classify persisted preference drafts through a shared storage backend

obra:
## Result

Native preference drafts now use a shared decoder and raw string-storage backend that distinguish absent, readable, and corrupt records while preserving the identity required for compare-and-swap recovery. Symbol markers keep stored JSON `null` distinct from absence, and canonical decoded identity avoids object-key-order mismatches.

## Scope and stack

This is the storage-only successor for the offline draft recovery decomposition formerly tracked by #1902. Provider wiring is #1956; retained same-hub snapshot recovery is #1964; the cold-offline recovery screen is #1904. The old #1902 branch is superseded and preserved separately. Provider and screen behavior remain outside this PR.

Own non-test scope is 142 changed lines. The original round-five branch remains preserved as `codex/offline-draft-round-five`.

## Validation and status

The 39 native storage tests, three targeted decoder tests, native/frontend typechecks, package qualification and root-export falsification, package-import lint, AppWire Biome, and diff checks pass. Independent review passed. Local RoboRev2552 identified only the explicit provider-wiring dependency.

The coordinator's `offline-refresh-6cf.json` receipt records the current target `main` at `d25d5afaa7fee061801258c6b64fee558d958bb3`, this head at `7db6c56bb77eb238700b07d40895ac305d864e57`, and the reviewed own-patch proof SHA-256 `78df920453b8e6059a9218f7961a284797c691fdc95100da598fb587354b9846`. Current CI and the raw review panel are still pending; Fresh exact-head CI and complete raw-panel qualification are required before merge. Raw-backend CAS coverage remains tracked in #1948, while broader native recovery simplification and the measured recovery-delay Low remain tracked in #1941.

Current main includes #1975. This merge-only refresh changes only the inherited retirement test; the entire owned patch is byte-identical. Current-head CI and complete substantive raw member reviews remain required.

Issue #1948, opened 2026-09-19
native: cover raw draft backend compare-and-swap behavior


obra:
The independent review of the native offline-draft storage replacement at 700b111a861b006b3d59c7535fc8a7f696b8384f found one Low test-coverage gap. Valid replaceIf and insertIfAbsent cases use the in-memory fake; directly exercise these contracts through rawStringDraftBackend with a fake string-storage port.

Cover matching replacement, stale-marker rejection, absent insert, existing-value refusal, malformed/null storage, and canonical object identity without relying on property order. Assert persisted values and return behavior. Keep the existing provider recovery coverage in its own successor. This is a test-only follow-up after the storage replacement, not a reason to expand the storage PR.

Additional Low coverage from the storage/provider reviews: compare decodeKeybindingDraftFields with the live keybindings-store restore projection using shared valid, absent/null, invalid-object, malformed-byte and stored-null fixtures; the existing test named matches restoreDraft only calls the decoder. Also assert that a successful recovery clears a previously published storageUnavailable state. Keep these behavioral tests in a follow-up, with no wording/implementation-string assertions.

Additional Low from retained-snapshot round4 independent review at dd60a43: add a provider-level regression carrying loadError through a failed offline probe and subsequent storage recovery. Direct native projection/source-preservation cases already establish the contract; this closes the end-to-end coverage gap without changing behavior.

Measured storage-test-double Low from the #1950 raw review at 3f6ba28ad: the new unique-symbol StoredNullRecord and UnparseableDraftBytes brands are invisible to canonicalJson and structuredClone. The fake draft backends can therefore reduce a stored-null marker to an ordinary empty object or an unparseable marker to its raw-only shape, losing marker identity; no current test routes these sentinels through those doubles, so no immediate product break is observed. Add marker-aware identity/round-trip coverage or route the doubles through the shared marker-aware identity helper. Keep this as test-double conformance work under #1948, separate from #1950's Lows-only scope.

[...]

Issue #1941, opened 2026-09-19
native: simplify offline draft recovery after the migration stack lands


obra:
The accepted simplify reviews of #1902/#1904 identified duplicated raw-string storage test fixtures, parallel draft classifications/read helpers, repeated reads during discard, and a swallowed rejection from the live-model nudge.

The replacement stack is still evolving. Recheck each item against the landed code before editing: readable-replacement recovery now uses the value-bearing read, so that helper must not be removed as supposedly unused. Preserve all CAS/identity checks and any rereads needed to prove replacement safety.

In a small follow-up, consolidate only remaining duplicate fixtures/state, explain or remove redundant reads, and make nudge rejection observable through the existing provider error path. Preserve offline recovery and readable replacements with meaningful behavior tests. This is separate from #1918's draft-port findings.

Measured Low from local RoboRev2579/2580: an initial genuine storage read failure leaves the live model storageUnavailable=true and draftUnreadable=false. If storage later exposes an unreadable record, successful store-free discard removes it but the model's eligibility check correctly refuses the diagnostic-only nudge. The stale unavailable snapshot survives until the existing Check current shortcuts recovery action. That action is enabled after reconnect by #1904, and model.refresh retries restoreDraft before loading hub settings (existing keybindingRecovery.test.ts:137–153 covers storage failure then refresh recovery). Improve this transition in a separate small follow-up without weakening discard eligibility, deleting readable replacements, or bypassing writeUncertain protection. Pin same-client recovery with a provider-level behavioral test. This is a recoverable diagnostic/editing delay, not data loss or a permanent lockout.

Measured mechanical Low from the #1964 raw review at 4d9a13016: NativePreferencesProvider.tsx re-declares DRAFT_RESTORE_FAILED_MESSAGE while keybindingsStore.ts already owns the same restore-error literal. Share the constant (or helper) and assert the shared value in provider coverage so retained/offline and live restore messages cannot drift. This is cleanup only and does not change recovery behavior.

[...]

PR #1904, opened 2026-09-18
Recover unreadable keybinding drafts while offline

obra:
An unreadable saved keybinding draft can be discarded from the native recovery screen even when the hub is offline or no live model exists. The screen uses the provider's guarded storage action, preserves saving and uncertain-write protections, and protects readable replacement drafts through compare-and-swap discard.

Follows merged #1950, #1956, #1964, and #1988. Current head 9b4dbb09de408faee7f3eb666dce1dc28f8513fb is rebased onto main62132380012925bdd502627599ebe14e8437b643 and includes the same-client disconnected retained-projection correction from #1988. The complete owned binary patch is unchanged from reviewed08dd625, SHA-256 ac76025ad1291e96c5649b5e6a321397179ef3618f2d079fc33fa1f8c049543f. Duplicate declaration and diff checks passed; review carries under the handoff rule.

Validation: 59 focused tests, native typecheck, package-import lint, independent screen review, and local RoboRev2589 passed. All three raw members at reviewed08dd625 completed and were read; two found no issues, and the remaining provider connection-state Medium is now integrated from #1988. All 15 current-head CI checks passed. Subsequent main changes are in the ask-state Go files and do not overlap the native screen/provider patch.

Verification-read uncertainty, delayed live-model recovery, and offline retry-message wording remain Low follow-ups in #1941. Optional provider and discard-time failure coverage remains #1948. No Low fix is folded into this screen slice.

PR #1964, opened 2026-09-19
Preserve diagnostic sources during offline draft recovery

obra:
Offline preference-draft recovery updates the retained snapshot for the current hub while preserving unrelated hub and load errors. Recovering a readable or absent draft clears only the draft diagnostic, and late work from another hub or client cannot update the retained view.

This is the retained-snapshot slice after merged #1950 and #1956. The same-client disconnected recovery correction is isolated in #1988; the recovery screen is #1904. The merge order is #1964 → #1988 → #1904. Storage, provider operations, retained projection, and screen behavior remain separate reviewable changes.

Validation: focused provider, native preference, draft-storage, and shared tests passed, along with native typecheck, package-import lint, and diff checks. Reversing production changes reproduced the diagnostic-source regressions. Independent correctness/simplification review and local RoboRev2596 completed. The screen integration has 59 passing focused tests.

Current head is `3c42857f491fa08fb468990022ade6f8f9717815`, rebased onto main `36c2d4768e4a13055398292103e5ea7dece3ed9e` after #1956 merged. The complete owned binary patch is unchanged (SHA-256 `d63b027c04b6d14a7453c693fc0f2de7de932d1588a97c014575f2f7138ab235`). All three raw reviewer bodies at the preceding identical patch `70e6d0c04` completed and were read; their remaining Medium is the concrete screen consumer in #1904. Review carries under the handoff’s identical-patch rule. A later screen panel identified a same-client disconnected-state gap; the isolated two-guard correction and 96-line regression are implemented in #1988 after independent review and local RoboRev2625 passed. This keeps the fifth-round retained projection immutable and decomposes its connection-state correction. All 15 current-head CI checks passed after one failed-job retry. The first attempt named an unchanged Go retirement test but omitted its inner assertion, so its cause is not claimed as confirmed. All three current-head raw panel members completed: two found no issue; the alleged raw hub-error exposure is refuted because the values remain in the native in-process snapshot and the screen renders fixed-copy sanitized error text. The same-client correction remains the explicit #1988 successor.

Optional recovery simplification and verification-read uncertainty remain #1941. Additional provider coverage remains #1948. Accepted Lows are kept outside this patch.


obra:
Coordinator hold at current head `3c42857f491fa08fb468990022ade6f8f9717815`: the same-client retained-snapshot review confirmed a Medium. On a same-client non-ready transition, both retained-reconciliation guards exclude the update, so definite readable, unreadable, or absent storage observations can contradict the stale retained draft/error. This is distinct from #1941's accepted genuine-port-error uncertainty. The writer is fixing only the provider guards plus a real-provider regression; no qualification or push claim is made. Source report: `sdk-d6-split/.superpowers/sdd/mobile-sdk-1934-pickup/offline-draft-retained-snapshot-same-client-review.md`.

Issue #1967, opened 2026-09-19
AppWire: prevent keyless history aliases from joining conflicting transcript keys


obra:
When a keyless live fragment shares an item id with two historical fragments whose supplied transcript keys differ, history coalescing can combine distinct turns. Conflicting supplied keys must remain separate while unambiguous keyless updates continue to match.

Measured at PR #1961 head a35f181df4fc9008b98bb2e8b9e26db2c5f4231a: older A contains shared-id/key-a with usage 10; older B contains shared-id/key-b with usage 20; fresh U contains keyless shared-id. A matches U and B matches U, although A does not match B. The parent keeps two turns; the candidate combines both keyed items into one turn and retains only usage 20. Canonical items survive; this is a Low follow-up because it requires an ambiguous mixed-key alias collision and no current end-to-end producer occurrence has been demonstrated.

Prevent an ambiguous keyless alias from bridging conflicting keyed identities. Preserve unambiguous keyless updates, genuine canonical transitive overlaps, fresh precedence, and existing conflicting-key isolation. Add a behavioral mixed-key regression through the public pagination reducer.

Evidence: remote Luna panel for #1961 plus independent exact-head probe of itemIdentityMatches/coalesceTurnHistory. Existing reducer tests separately cover keyless updates and conflicting supplied keys. This issue tracks the accepted Low separately from the qualified coalescing and placement PRs; do not broaden those patches to absorb it.

PR #1968, opened 2026-09-19
Preserve retained turn order in paginated history

obra:
Paginated history could move retained turns across fresh turns or reverse retained runs when fragments coalesced into shared anchors. This change places retained turns within monotonic anchor bounds, preserving fresh order, retained order, and every item exactly once. Compatible transcript positions determine order within those bounds; positionless runs use the next positioned turn on either side.

For example, older B2,C4 and fresh U?,D1,C4 now produce U,D,B,C, preserving both the positionless fresh announcement and the comparable D1-before-B2 ordering. This is the private placement slice following merged #1961; public coverage/overlap classification and mobile consumers remain separate successors. Own scope: 157 changed production lines and 421 test additions.

Validation at 53ea59b003e5afe40620d2b8341c12c3efa5d654: 260 scoped tests pass; the original 49,608-case ordering matrix and its 49,608-case positionless-announcement extension both have zero failures. Production-only reversal fails the prefix regression; the prior implementation fails 35 extended numeric-order cases. Package build/qualification, frontend/native types, Biome, import checks, and independent correctness/simplification review pass. Local RoboRev #2654 passes. Current-head CI and remote raw review must settle before merge.

Review round 4 of 5. Existing coalescing Low #1967 remains separate.


obra:
The review finding about positionless fresh turns in interior/suffix gaps is disproved. Exact helper probes preserve fresh order, retained order, anchor bounds, and the later comparable numeric positions: older A1,B2,C4 with fresh A1,U?,C4 yields A,U,B,C; adding fresh D3 yields A,U,B,D,C. There is no position evidence ordering B relative to U.

A separate prefix case was found while checking the boundary: older B2,C4 with fresh U?,D1,C4 yields B,U,D,C, although the accepted generic fragment-placement contract can satisfy U,D,B,C. This is broader than a demonstrated single current Go pagination request, but remains within the explicitly accepted helper contract. The PR is held for a small round-four correction and regression covering that distinct case. No ordering assertion will be weakened, and the disproved original finding stays closed.

obra:
Exact-head raw panel bdf89b3e is complete. Muse 22329 and DeepSeek 22330 report no findings. Luna 22328 Set/toHaveLength claim is already disproved by the repository toolchain and passing matrix tests; no re-investigation is needed. Its new next-anchor claim is also disproved: older [A@1,B@2,C@4] and fresh [A@1,U?,C@4] yield [A,U,B,C]. U has no canonical position, so both U,B and B,U satisfy all evidence; C only constrains both before C. Including C in the successor lookup selects an arbitrary tie order, not a correctness fix. Independent bounded review confirms this; the comparable-position regressions and 49,608-case extended matrix remain valid. All 15 current-head CI checks pass at 53ea59b003e5afe40620d2b8341c12c3efa5d654.

PR #1962, opened 2026-09-19
Keep live ask state aligned through failure and yield boundaries

obra:
A human-note carrier can preserve an unanswered question while failing, yielding, or finishing with a notification still queued. The live session now stays awaiting input in those cases, matching restoration. This covers provider/generic failures, no-tool and tool-round exhaustion, transcript-poison refusal, notification yields, observer handoffs, and clean settlement with pending autonomous work. Closed sessions remain closed, and ordinary settlement without a pending ask keeps its existing behavior.

Follows merged #1958. This live-boundary slice owns 111 changed production lines and 530 test additions. Durable user-input admission remains the separate #1977 successor, with locally qualified candidate `006687b69`; carrier-claim cleanup remains #1907. The clean-settlement correction is product review round 3, within the five-round cap.

Current head `437f4ed159f8b49b0ef7bba2aa8e9066b23f9b4a` is rebased onto main `54dabf13f2ff6044ff70ad040a0188cc59b30c62`. The complete owned binary patch is identical to locally reviewed d8a58f22a, SHA-256 `16c22176e35b72987b11067ef473ae1910bb65336869a4987b6e04ae65201f2f`. Backups and the later C/carrier branches are preserved. Duplicate declarations and diff checks passed.

Validation: focused normal/race/tagged agent tests, normal/tagged/Windows vet, agent lint, toolchain formatting, and diff checks passed. A real-session regression queues a notification during a human-note communication round, checks live/restored awaiting and pending asks, and verifies no autonomous third provider request. Reversing only the production correction reproduced the old idle-state failure. Independent correctness/simplification review passed and local RoboRev 2626 found no issues. All 15 current-head CI checks passed. All three configured remote raw reviewer bodies completed, were read, and reported no findings. The separately scoped durable-admission correction in #1977 remains the immediate next successor.

Documentation/provenance-coverage Lows remain #1946. The split-lock allegation is refuted by the serialized turn-loop contract. The distinct inherited interrupt-marker persistence gap remains #1181 and is not claimed fixed here.

Issue #1946, opened 2026-09-19
agent: verify legacy steering text evidence parity and simplify ask-boundary helpers


obra:
The #1906 review identified a Low difference between live and restored steering classification: live classification passes empty text evidence, while restore can recognize the human-note prefix for a kindless, provenance-less existing journal record. Current SetHumanNote writes explicit provenance, so this is not evidence of a new-write mismatch.

After #1906/#1907 or their replacements land, add a meaningful live/restore oracle for the existing journal shape in session_notes_provenance_test.go, then make the smallest correction if the mismatch is reachable. Preserve the no-askPending-fallback ruling and the existing counter-clearing contract. Do not introduce a new compatibility layer or broaden which inputs answer a pending question.

The accepted simplify notes also identify repeated scans/helpers/test fixtures around this same boundary; measure and consolidate remaining duplication only after the active correctness fixes settle. Keep this separate from their must-fix terminal-state work.

Additional optional Low from #1958 raw-panel audit: add one combined restoration fixture with a resolving user steer followed by a same-round daemon task reminder. The current outer restore scan reaches the user entry and clears older asks even when the inner entry predicate first sees the non-resolving reminder; no production defect was found. A combined fixture would pin this interaction directly alongside the existing individual oracles.

Independent review of the poison-refusal fixture at a55e6f1d3a6d0c319e62df24fda30eff5794f4cc records a P3 test-quality follow-up: assert askPendingCount()==1 after both the live and restored awaiting states, and register deferred cleanup immediately after creating each session so a fatal assertion cannot leave it open. This does not change the qualified failure-boundary production fix.

Measured optional Low from the #1907 carrier panic test at 14303886: the test proves a real append panic propagates and the deferred claim marker clears, but it does not itself seed or assert a pending ask or durable ownership. Independent review confirms A/B/C and startup recovery already cover those broader semantics. Add this as focused coverage only if useful; it is not a blocker and does not justify expanding #1907.

### Confirmed recovered turn/start admission Medium

[...]

PR #1907, opened 2026-09-18
agent: steering-carrier claim id is released on every exit

obra:
A panic while draining steering input could leave the carrier claim marker set. A later unrelated selection failure could then be misclassified as a carrier failure and resolve a pending question during restore. The marker is now cleared by defer on every exit.

Follows merged #1958, #1962, and #1977. Refreshed head `00849a796e8ec03b76426fca443e132dbbd62dd2` is based on main `810a52f3232586c525580f431e390d7b03be29ba`. Own scope remains five production additions, two deletions, and 54 test additions. The complete owned patch matches the previously reviewed patch after removing only diff blob IDs and hunk coordinates (SHA-256 `0582be763e72889fb4701a2e05e6d5da67f1b2550cae7c322be03ccbe2f491bd`). Duplicate declaration and whitespace checks passed.

Validation: the regression panics through the real append path and verifies marker cleanup. Focused normal/tagged/race carrier and live/restore tests, normal/tagged/Windows vet, agent lint, and formatting passed for the reviewed patch. Independent correctness/simplify review and local RoboRev2613 accepted it; the three raw member bodies at `65933f7cf` were read. The byte-identical refresh carries that review under the handoff rule; all 15 fresh current-head CI checks passed.

Disposition of that panel's three Medium bullets:
- Direct user-input admission: fixed by merged #1977, now in this base.
- Kindless, provenance-less live human-note classification: the suggested construct is outside the supported live admission path, which writes explicit kind/provenance. Legacy restore coverage remains tracked in #1946.
- Interrupt-marker persistence: real inherited bug tracked in #1181. Jesse explicitly directed merging this carrier cleanup first and fixing interrupt persistence in the immediate follow-up. This PR does not claim that fix.

Optional carrier-test coverage and other Lows remain separate in #1946.

Issue #1181, opened 2026-09-12
agent: audit durable writes that publish in-memory state before the transcript accepts them


obra:
Residual from #1098 round 10 (part of #1116).

`3a6ba6b91` on #1098 closed one instance: `publishFoldTransaction` swapped model history, reset the environment tracker, claimed the session note and stamped the revision before `commitTranscriptsLocked` wrote the fold's markers, and a poisoned writer only produced a warning, so the session announced and acted on a compaction that a restart would lose. It now refuses under `attentionMu` before publishing, the same admission rule the turn loop applies at turn start (round 4 of #1098).

Not yet done: an audit of every other path that publishes to clients or mutates in-memory session state and then makes a durable transcript write, to confirm each one either checks `Poisoned()` under the transcript door first or rolls back on write failure. Candidates to check: hook completions (`emitHookCompleted` now writes before announcing on #1100), steering injection, session-name/namer writes, delegate preseeded input, checkpoint/summary records, and the queue/steering claim stores. Outcome: a table of write sites with their ordering guarantee, fixes for any that publish-then-lose, and a test per fixed site. Related: #1150 (live event order vs transcript order), #1165 (check-then-claim window).

obra:
Another site for the audit, found in #1100 round 12: a compaction marker whose transcript write genuinely fails (not merely withheld) still reaches `handleCompactionTurnEffects`, which logs a warning and then emits `EventCompactionTurn` and runs the session namer anyway (`agent/session_namer.go:483-497`). #1100's `ece521c4a` suppresses effects for *withheld* markers only; the failed-write path is the same publish-without-durable class, one step out.

obra:
Two more sites from #1098 round 13, both pre-existing on main (blame: `b589a8d263`, `4b384fe790`, `841a199c58`, `adf38dddf9`):

- `agent/session.go:1878-1888` `recordTurn` — history append first, then the transcript write, warn-only on failure; reached for direct user input from `agent/session_lifecycle.go:2124` via `appendTurn` (`agent/session.go:1579`). Guarantee today: publish-then-write. Wanted: the write-first shape `appendEnvironmentContext` and #1100's `emitHookCompleted` use.
- `agent/session_compaction.go:242-257` — commit-phase marker/rewrite-tail write failures after the pre-publication poison check leave live state compacted; #1100's `ece521c4a` covers withheld markers, the failed-write path is the earlier note above.

obra:
[...]

PR #1977, opened 2026-09-19
Preserve pending questions until a reply is durably admitted

obra:
A reply that fails admission must not erase the question it was meant to answer. Pending asks now clear only after the resolving user turn is recorded or incorporated under the existing transcript admission contract. MaxTurns refusal, environment persistence failure, zero-byte writes, poisoned transcripts, and a closed writer preserve the pending question and its restored awaiting state.

Direct input uses the existing synchronized transcript-pair helper. It refuses a write that recorded nothing and adopts an already-retained record once, including the helper's explicit retained-but-unsynced outcome. Recovered client-start failures clear the ask only when that mutation's user record exists. A later recovery error does not hide an already-recorded boundary, and rejection before a user record does not publish input or call the model.

Follows merged #1958 and #1962. Current head is `b84a87d256f3a6c00458614e723bea4121d14d2f`, based on main `011ed403ed4ef7c54066a0536fa6adcb04cfb8e0`. Own scope is 88 changed production lines and 362 test additions / 10 deletions. The carrier cleanup in #1907 remains separate.

Validation: the production correction passed focused normal/race/tagged admission, transcript, and client-mutation regressions, normal/tagged/Windows vet, agent lint, formatting, independent review, and local RoboRev2624. The first remote CI run exposed two test-harness defects: an unmarked hang deadline and an interrupt fixture whose old fault occurred before its model callback. Both failures were reproduced. The test-only repair explains the deadline and poisons the writer inside the second scripted response, then interrupts and verifies exactly two model requests, ErrWriterPoisoned, and the retained queue. The two failures and related poison-writer regressions passed normal/race/tagged checks; vet/lint/format passed. Independent review accepted the repair, and final-head local RoboRev2628 found no issues. Production code is unchanged by this repair. All 15 final-head CI checks passed. The three-member raw panel 0d56c930 is complete: Luna and Muse found no issues, and DeepSeek reported only the goal-blocking policy Low tracked in #1946.

The stale admission comment and final panel's policy Low about blocking a goal after a pre-admission write refusal are tracked in #1946 and excluded from this patch. The inherited interrupt-marker persistence gap remains #1181 and is not claimed fixed here.


obra:
Coordinator hold at current C head `b84a87d256f3a6c00458614e723bea4121d14d2f`: #1977 remains published after a test-only CI repair (11 additions, 5 deletions, no production changes). The production correction is unchanged from `3c99890518646385a9cc5e1e371774357a74cca9`. The earlier `006687b69` to `3c99890518646385a9cc5e1e371774357a74cca9` patch had SHA-256 `326ff844a7b7ac1330f2bc16219742900a43fde1bfd962194843653fe071e8cf`; b84 adds the test-only 11/-5 delta, so the current full patch differs while production remains unchanged. Local RoboRev2624 and independent production review carry forward.

Final b84 panel `0d56c930` (08:08 UTC snapshot) has Muse finding no issues, DeepSeek reporting the stale documentation Low at `agent/session_tools_ask.go:112-114` plus one separate policy Low at `agent/session_lifecycle.go:2364`, and Luna still running. The policy Low is that a pre-admission transcript-write refusal flows through generic goal-blocking and can leave an active goal blocked after a transient write failure. It is inherited by the test-only repair but potentially affected by C's changed admission behavior; track it separately under #1946, not as #1181's interrupt gap and not as the stale comment.

Fresh current-head CI and raw qualification for `b84a87d256f3a6c00458614e723bea4121d14d2f` remain pending. Do not claim current all-review qualification.

PR #1982, opened 2026-09-19
appwire: preserve tool-result source precedence across history fragments

obra:
History fragment folding could let an older tool result overwrite fresh fields, lose values carried through transitive item aliases, or skip a partial result whose call ID is supplied by a matching fragment. The private fold now records original source membership at the actual merge boundary, routes those values through the normalized call ID, and applies fresh-field precedence with fallback for omitted fields. Same-source results follow normalized order. Provenance collection uses a linear traversal rather than an all-pairs scan.

Follows merged #1961 and #1968. This prerequisite remains separate from the public history-coverage export and mobile consumer integration. The owned package change is 200 production lines and 646 test lines; no public API or model fields are added.

Validation: 280 reducer tests pass after restacking onto merged #1968; six new regression witnesses fail under production-only reversal. Independent review verified the 128-case field matrix, normalized ordering, transitive aliases, missing call IDs, and traversal cost. Package build/qualification, frontend and native typechecks, touched-package Biome, import lint, and diff checks pass. Independent correctness/simplification review and local RoboRev2634 found no actionable issues at 46ff5d16dbf60d1ba39b44d944113c6f380dabdc. Current-head CI and the configured remote review panel remain required before merge.

Restack validation at 7144d78f64eccc477ea69a541315af2cf3b47ab9: all private declarations and owned changed lines are byte-identical to the reviewed implementation, and all five placement functions are identical to main. Both 49,608-case placement matrices pass with zero violations. The traversal-coverage Low remains separately tracked in #2006.


obra:
Coordinator hold at current head `032c030251ac562b807f68ae700e4e36ee50f850`: CI is reporting test-helper errors. The local `9747f949a3e556eea013b32d91d5c1be71f58893` candidate is unqualified and parked after the independent audit found three Mediums: transitive alias-lineage loss, omitted `callId` provenance, and quadratic scanning. The precise plan is `history-private-provenance-9747-independent-audit.md`; no corresponding type/provenance fix is published at remote `032c`. Owner: history/private-provenance writer and audit.

obra:
Remote review disposition for exact e6d550e472a0ed09d8290665a8682bbb97336ca2: Muse 22129 found no issues. DeepSeek 22130 found no production defect and one cost-test Low. Independent reproduction refutes the blanket vacuity claim: the existing guards fail the original quadratic collector (72→272 reads against bound 232; 460→1720 against bound 1420). Zero current counts reflect skipped ordinary-item work.

The narrower gap—scaling active tool-call/result candidates after the callId guard—is tracked separately in #2006. The qualified patch is unchanged. Luna 22128 and final current-head CI remain pending; this is not a merge-completion claim.

obra:
DeepSeek22398 repeats two previously investigated claims. The completed-to-inProgress duplicate-result example was audited against the settled-only Go result producer and latest-terminal contract and is not a valid producer state; the reviewed private declarations are unchanged through this restack. The blanket zero-read claim was independently falsified: the old collector measured 72 to 272 and 460 to 1720 reads against bounds 232 and 1420, while current zero ordinary-item reads reflect skipped work. Active tool-call/result scaling after the callId guard remains the separately tracked coverage Low #2006. These dispositions are retained without repeating disproved-review work. Muse22397 is clean; Luna22396 and remaining CI must still finish.

PR #2009, opened 2026-09-19
fix(appwire): retain ready-generation stamps on shortcut drafts

obra:
Shortcut drafts need to retain the ready generation in which they were created or confirmed so reconnect handling can distinguish stale drafts even when the server reuses a revision number. This first small replacement for #1792 adds the typed generation stamp and preserves it through editing and saving; ordinary storage restore starts unstamped, and authoritative reads or explicit rebase earn a generation.

The generation conflict fence and identity-aware storage recovery are separate reviewed successors. This PR does not claim the complete reconnect fence on its own.

Validation: 120 focused tests, production-only reverse-patch falsification, TypeScript build, package qualification, import lint, and scoped Biome passed. Independent correctness/simplification review and local RoboRev #2638 passed. Publication refresh preserves the owned change; current-head CI and the remote raw review panel must settle before merge.


obra:
The generation parameter is deliberately consumed by the separately implemented identity-aware recovery successor, c99c7205601463d2198f6b29519fbb195eece6e3. Its reloadDraft preserves the known stamp only when the same classified checkpoint is reloaded after storage failure, and passes that stamp through restoreDraft into draftFieldsFrom. Ordinary restore still supplies null. That successor has 125 focused tests, production-only falsification, independent coordinator review, and local RoboRev #2648 passing. No parameter removal or unrelated Low fix is being folded into this first slice; the successor will follow the generation fence in dependency order.

PR #2014, opened 2026-09-19
fix(appwire): require rebase for drafts from an earlier ready generation

obra:
A reconnect can reuse the same shortcut revision number while representing a new ready generation. Previously a draft created against the old generation could still appear saveable. This change marks such drafts conflicted, preserves the old stamp through edits, and requires explicit rebase to clear the conflict.

Follows merged #2009 and is the second small replacement for #1792. Own change is 16 production lines and 55 changed test lines. Identity-aware recovery after a failed local save is the separate, implemented successor c99c7205601463d2198f6b29519fbb195eece6e3; it prevents reloading the same checkpoint from erasing the old stamp. Local RoboRev #2644's finding is that successor boundary, covered by its tests and passing review #2648.

Validation: 122 focused tests, production-only reverse-patch falsification, TypeScript build, package qualification, import lint, and scoped Biome passed. Independent correctness/simplification review passed. Restack onto main 91a1902fee1d02efaace2dd97ef9a277c3e24f07 preserves the complete owned binary patch (SHA-256 85ebaea5935319b943b561cccd00bab891e1e1c8703be277dec549847ca1df62). Current-head CI and the complete raw remote review panel are required before merge.


obra:
Current-head panel 49d6bb30 is complete: Luna and Muse report no issues. DeepSeek 22318 identifies the same identity-stable reload boundary as local review 2644. That behavior is addressed by the implemented immediate C successor c99c7205601463d2198f6b29519fbb195eece6e3, qualified by regression tests, independent review, and local RoboRev 2648 (no findings). This remains an explicit dependency-chain obligation; it is not dismissed as disproved. All 15 current-head CI checks pass at 8ac9e61f7607beb105e86cc4bb3a522de4d53456.

PR #2019, opened 2026-09-19
fix(agent): a failed replay-tail copy takes its compaction marker with it (#1158)

### LINKED ISSUE #1158: agent: a replay-tail copy whose durable write fails still gets its compaction marker, losing the original on resume
## Where
`agent/session_compaction.go`, `publishFoldTransaction`'s transcript-commit phase (the replay-tail loop that collects `mergedTailWriteErrs` and then calls `commit.commitTranscriptsLocked()` unconditionally). Found by the PR #1100 round-8 review; pre-existing, present under both the old and the new tail ordering.

## What
A replay-tail copy is written with `AppendDurable`. If that durable write fails, the transcript writer rolls the append back to its recorded start offset (`agent/transcript/transcript.go`, `rollbackAppendLocked`), so the copy vanishes from disk. The fold then writes its compaction marker anyway. On resume the anchored read discards everything before the marker, including the original whose replacement copy never landed.

## Why it matters
It is the same permanent history loss the round-8 finding was about (tail lost across a crash), reached through an I/O error instead of a crash, and it is silent: the in-memory history published by the fold still has the turn.

## Direction
Not obvious. Skipping the marker when a copy write fails leaves the published in-memory history with no durable anchor, so the next resume would replay the full pre-fold transcript instead. Options: fail the fold publication as a whole when any copy write fails (the fold becomes retryable); or write a fallback completion record naming the lost copies so the reader can keep their originals. Needs a small design pass; either way, add a test that fails a copy write and asserts the originals survive a resume.

---

obra:
Fixes #1158

## Root cause

`publishFoldTransaction` (agent/session_compaction.go) committed the fold's compaction markers through the **buffered** door (`writeTranscriptLocked` -> `Append`) and then wrote the replay-tail copies through the **durable** door (`writeTranscriptDurableLocked` -> `AppendDurable`), as separate appends.

When a copy's durable write failed, the writer rolled that copy back out of the file (`rollbackAppendLocked`, transcript.go) — correctly, that is its contract. But the compaction marker written before it stayed. `ResumeHistory` anchors on the **last** marker and discards every entry before it (agent/transcript_read.go), so on the next resume the original turn the copy was meant to replace was dropped, while the in-memory fold still held it. That is the same permanent, silent history loss as a crash between the two writes, reached through an I/O error instead.

## Fix

Write the fold's compaction markers, its steering turns, and the replay-tail copies as **one all-or-nothing durable `AppendBatch`** — markers first, copies last. `AppendBatch` already exists as "the exported entry to the one write primitive" with rollback-to-start-offset semantics; it rolls the **whole** batch back when the writer cannot record it. A marker that discards the originals it replaces can therefore no longer land without the copies meant to replace them: either both are durable, or neither is and the next resume replays the pre-fold transcript (a compaction lost, not turns).

Files: `agent/session.go` (new `writeTranscriptBatchLocked`), `agent/session_compaction.go` (batch the fold transcript commit), `agent/session_config.go` (test seam `appendFoldTailTurn`), `agent/transcript/transcript.go` (doc: name the fold caller), plus the regression test.

No `schema.Turn` / reader-side change is required: because the batch rolls back as a unit, a failed copy never leaves a marker, so the existing resume anchor stays correct.

## Test evidence

`TestFoldPublication_FailedTailCopyDoesNotCommitCompactionMarker` records a turn while the fold runs (so there is a replay-tail copy), fails that copy's durable write, then asserts the turn survives `ResumeHistory` exactly once.

Fails before the change (observed on the pre-fix code):

```
--- FAIL: TestFoldPublication_FailedTailCopyDoesNotCommitCompactionMarker (0.22s)
    session_fold_publication_test.go:675: the merged-back turn appears 0 times in the resumed history, want exactly 1 -- a compaction marker committed while its replay-tail copy's durable write failed discards the original on resume
FAIL
```

Passes after the change.

## Gates run

- `go test ./agent/... -count=1` -> `ok` (all packages)
- `make lint-golangci` -> `PASS lint-golangci`
- `gofmt -l` over the touched files -> empty
- `go vet ./agent/...` -> clean


obra:
Addressed the roborev findings on the previous head (`caa27b2`) in `3ab57f3`:

**High — `persistedAppendLog` cleared before the batch succeeds.** The prune is now deferred: the log is advanced/cleared only when `commit.transcriptCommitErr == nil`, i.e. after the batch is durable, still under `attentionMu` (so no append can interleave). A rolled-back batch keeps the persisted forms. The regression asserts `persistedAppendLog` is non-empty after a failed batch.

**High — pinned note claimed before the batch.** On a rolled-back batch the fold's steering turns (the note-handoff among them) are re-recorded on their own via `writeTranscriptLocked`, so the handoff the note claim consumed in memory still has a durable record. The regression sets a pinned note, fails the copy write, and asserts the handoff survives `ResumeHistory` exactly once.

**Medium — non-atomic readiness check in `writeTranscriptBatchLocked`.** The ready check and the queue now happen under one `s.mu` hold (append all turns to `pendingTranscriptTurns` if not ready), mirroring `holdTurnUntilTranscriptReady`; the writer is sampled under the same hold. An `attachTranscript` can no longer slip between check and queue and drop the batch.

**Low — stale comment.** The `publishFoldTransaction` comment no longer claims separate "buffered writes"; it now describes the single durable batch.

Gates on `3ab57f3`: `go test ./agent/... -count=1` ok, `make lint-golangci` PASS, `gofmt -l` empty over the touched files, `go vet ./agent/...` clean.

obra:
Second review round, addressed in `c8b5f79`:

**HIGH — `AppendBatch` returns nil for a landed-but-unsynced batch, so the fold pruned the log on an unconfirmed durability claim.** Fixed at the prune decision. New `transcript.Writer.AppendBatchSynced` mirrors `AppendSynced` for a batch: `nil` = whole batch recorded and synced; `*RetainedUnsyncedError` (`errors.Is ErrRetainedUnsynced`) = whole batch recorded but its fsync *and* the recovery barrier failed; any other error = nothing recorded. `writeTranscriptBatchLocked` uses it, and the fold now prunes `persistedAppendLog` **only in the `nil` case**. A retained-unsynced batch is kept (a returning reader finds the marker and its copies) but the log is left un-pruned.

Test: `TestFoldPublication_UnsyncedBatchKeepsReplayTailLog` installs a transcript writer whose `Sync` and `Truncate` both fail, so the batch write reaches the file, its fsync fails, and its rollback truncate fails — the ambiguous retained-unsynced outcome. It asserts the log is **not** pruned and the batch is still readable. Verified red before the gate: reverting the batch door to `AppendBatch` fails it with *"a batch that landed but could not be synced was treated as durable and pruned the replay-tail log"*.

**MEDIUM — the fallback note handoff was re-recorded buffered.** `recordSteeringAfterFailedBatchLocked` now writes the fallback steering turns through `writeTranscriptSyncedLocked` (fail-closed durable, `AppendSynced` + barrier); a recorded-but-unsynced result is adopted, any other error is reported. The note claim already consumed the note, so its handoff is confirmed durable rather than merely landed.

**LOW — `appendCompactionTurn` injection neutralized by the recovery path.** The recovery path now calls the same hook and skips any turn it fails, so an injected steering failure is not papered over by a real write. A hook-injected failure also short-circuits before the real writer, so nothing is written and `transcriptCommitErr` is set. The comment now states the hook aborts the fold's transcript batch rather than "nothing lands".

**LOW — stale `writeSteeringTurnRecordsLocked` / `flushSteeringTurnRecords` doc.** Both comments now state they are test-only, superseded by the fold's durable batch in `commitTranscriptsLocked`.

Acceptance on `c8b5f79`:
```
$ go test ./agent/ -run 'Compaction|CompactionMarker|ReplayTail|Fold' -count=1
ok  	primeradiant.com/evener/agent	20.066s
$ go test ./agent/ -count=1
ok  	primeradiant.com/evener/agent	455.751s
$ go test ./agent/transcript/ -count=1
ok  	primeradiant.com/evener/agent/transcript	0.610s
$ git show --name-only --format= HEAD | xargs gofmt -l
(empty)
$ make lint-golangci
PASS lint-golangci (104s)
$ go vet ./agent/...
(clean)
$ gh pr view 2019 --json headRefOid
{"headRefOid":"c8b5f79d2f4fc69356935f747843c4b0c473edee"}
$ gh pr view 2019 --json mergeStateStatus
{"mergeStateStatus":"BEHIND"}
```

obra:
Third review round, addressed in `a17f592`:

**HIGH — pinned-note durability window.** Fixed fail-closed. The claim stays atomic with the history swap under `s.mu` (the documented invariant), and the claim now carries a generation-checked restore. New `noteHandoffClaim{claimLocked, restoreLocked}` travels through the note registrar; `restorePinnedNoteLocked` puts the note back only when its own generation is still current and nothing newer was pinned, so a note repinned mid-fold is never clobbered.

Where the restore runs, and why ordering alone is not sufficient: the claim **cannot** simply be deferred past the batch, because deferring the clear leaves the note globally visible while the batch is written — a competing fold then captures and re-injects it (double handoff), and the atomic claim's generation check is what stops a stale clear from erasing a newer note. So the note is claimed atomically as before and restored when durability is unconfirmed:
- rollback: re-record the steering/handoff fail-closed; if the note handoff's own synced write is not confirmed durable (error or `ErrRetainedUnsynced`), restore the note;
- retained-unsynced batch: the handoff is in that same unconfirmed batch, so restore the note.
Both branches read the **same** classification that gates the `persistedAppendLog` prune (`transcriptCommitErr`/`transcriptRetainedUnsynced`), so the log prune and the note restore cannot disagree about durability. The restore runs under `s.mu` inside the transaction, before `flush()` and before the metadata save, so the persisted state keeps the note pinned and a crash re-emits it.

**MEDIUM — `AppendBatchSynced` fail-open on a closed writer.** Now passes `failClosed=true`. A closed writer records nothing, so it can be neither recorded nor durable; `ErrWriterClosed` is the only honest answer and matches `AppendSynced`. The nil-writer no-op stays for stateless sessions. (`ErrWriterClosed` is classified as a rollback by the fold, so it neither prunes the log nor keeps the note consumed.)

Regressions (failing-before evidence):
- `TestFoldPublication_UnsyncedBatchRestoresPinnedNote` — without the restore: `the pinned note after a batch that landed unsynced = "", want it restored to "REMEMBER: the exact API signature"`.
- `TestAppendBatchSyncedFailsClosedOnAClosedWriter` — without `failClosed=true`: `AppendBatchSynced on a closed writer = <nil>, want ErrWriterClosed ...`.
- `TestFoldPublication_FailedTailCopyDoesNotCommitCompactionMarker` gained the complementary assertion that a **durable** recovery does *not* restore the note (consumed once the handoff is durable).

Acceptance on `a17f592`:
```
$ go test ./agent/ ./agent/transcript/ -count=1
ok  	primeradiant.com/evener/agent	290.373s
ok  	primeradiant.com/evener/agent/transcript	0.579s
$ make lint-golangci
PASS lint-golangci (87s)
$ git show --name-only --format= HEAD | xargs gofmt -l
(empty)
$ gh pr view 2019 --json headRefOid
{"headRefOid":"a17f592fe3064058d32a685459ac01c8f6da98c9"}
$ gh pr view 2019 --json mergeStateStatus
{"mergeStateStatus":"BEHIND"}
```

obra:
Fourth review round (`a17f592`). Two Lows fixed in `cbc71cc`; the High is reported below with a concrete blocker, not silently skipped.

**LOW — recovery-failure arm untested.** Added `TestFoldPublication_FailedBatchUnconfirmedRecoveryRestoresNote`: a hard batch failure (injected tail-copy failure) plus a recovery write that can only land unsynced (fsync and rollback truncate both fail), asserting `PinnedNote()` is restored. That is the only path where `recordSteeringAfterFailedBatchLocked` returns false.

**LOW — seam consistency.** Chose the "document as failure-only" resolution rather than change write semantics: `appendCompactionTurn` and `appendFoldTailTurn` are now documented as FAILURE injectors — a non-nil return aborts the fold's batch, a nil return injects nothing and the turn is written for real, so the seam does not replace the writer. The fold-path comment was corrected to match.

**HIGH — retained-unsynced batch and the crash window: this one needs a design decision, not another local fix.** The reviewer's harm requires the marker lines (which precede the copies in the single batch write) to survive a crash while the tail copies do not. I could not close that with a local change, and I want to be precise about why rather than land a half-measure:
- The copies must be written *after* the marker to survive `ResumeHistory`'s last-marker anchor. Putting the copies first makes them pre-marker, so they are discarded on a normal resume (breaks `TestFoldPublication_TurnRecordedDuringFoldSurvivesRestart`), unless the reader keeps copies no marker claims.
- The marker cannot be un-written: in the retained-unsynced case the rollback truncate already failed (that is exactly what made it ambiguous).
- Keeping the lifecycle operation "pending" does not restore the turns: the loss moment is `ResumeHistory` anchoring on a surviving marker, and resume does not consult the compaction operation.
- Re-appending the batch durably can help the transient case but cannot fix a persistent fsync failure.

So closing it fully requires either the #1100 tail-first + `schema.Turn.CompactFoldID` reader change (write the copies first and have the reader drop copies no marker claims), or a crash-atomic transcript commit (e.g. rename-based). Both are the "small design pass" the issue itself flags, and the second is a transcript-format change. This PR is deliberately scoped to the split-append defect the issue names; I did not start a partial `schema.Turn` migration. Flagging for a maintainer decision: land this PR and take the crash window as a follow-up / in #1100, or expand this PR into the reader change.

Acceptance on `cbc71cc`:
```
$ go test ./agent/ ./agent/transcript/ -count=1
ok  	primeradiant.com/evener/agent	424.675s
ok  	primeradiant.com/evener/agent/transcript	0.762s
$ make lint-golangci
PASS lint-golangci (83s)
$ git show --name-only --format= HEAD | xargs gofmt -l
(empty)
$ gh pr view 2019 --json headRefOid
{"headRefOid":"cbc71cc93d3418d8919573db0a4dd850420d1134"}
$ gh pr view 2019 --json mergeStateStatus
{"mergeStateStatus":"BEHIND"}
```

obra:
Fifth review round (`cbc71cc`). High and Low fixed in `d9dd370`; Medium reported with a concrete blocker.

**HIGH — the compaction finalize didn't check batch durability.** Fixed. `commitSkillCompactionPublication` now only finalizes (clears the pending operation, consumes its reload selection, advances the receipt) when the batch is **confirmed durable** (`transcriptCommitErr == nil && !transcriptRetainedUnsynced`). On a rolled-back or landed-unsynced batch the claim is now **reverted**: `restoreCompactionClaimLocked` returns the operation to its pending phase, clears its `PublicationID`, and withdraws the handoff it recorded, so the persisted snapshot keeps the intent and a restart re-arms it instead of finding a published slot whose receipt never landed. Only `publishFoldTransaction` calls the finalize, so this is the single choke point.

Test: `TestSkillCompaction_FailedBatchKeepsPendingOperation` runs a real checkpoint fold over a writer whose fsync and rollback truncate both fail, then asserts the operation is still `pending` in memory and on disk. Fails before the change with the operation left `Phase:published PublicationID:fold-2`.

**LOW — seam contract disagreed between the two writers.** Fixed by aligning the survivor: `writeSteeringTurnRecordsLocked` now treats `appendCompactionTurn` as a failure injector (nil ⇒ write for real), matching the fold path and the config comment. Test: `TestFlushSteeringTurnRecordsNilHookStillWrites` (fails before with the turn missing from the transcript).

**MEDIUM — partial-batch framing: not fixable in this change; same class as the crash-ordering gap.** The harm needs a multi-entry batch whose write partially succeeds and whose rollback truncate then fails, leaving complete marker lines plus a partial tail; recovery trims only the partial last line, so the marker anchors and the copies it claims are gone. Closing it needs batch-level crash-safe framing/commit semantics (e.g. a single integrity-framed record, or a durable batch-start marker recovery can truncate to) — i.e. a transcript-format/reader change, exactly what PR #1100's tail-first + `CompactFoldID` design does. I did not land a partial version of that: it would break the normal resume path or duplicate history without the reader change. Flagging for maintainer decision, same as the previously-reported crash-ordering window.

Acceptance on `d9dd370`:
```
$ go test ./agent/ ./agent/transcript/ -count=1 -timeout 1200s
ok  	primeradiant.com/evener/agent	397.660s
ok  	primeradiant.com/evener/agent/transcript	0.571s
$ make lint-golangci
PASS lint-golangci (77s)
$ git show --name-only --format= HEAD | xargs gofmt -l
(empty)
$ gh pr view 2019 --json headRefOid
{"headRefOid":"d9dd37032f2af1c849708c225e74044b6932912f"}
$ gh pr view 2019 --json mergeStateStatus
{"mergeStateStatus":"BEHIND"}
```
Note: an earlier full-suite run on this head hit the 600s wall clock; the stacked goroutine dump placed the hang in `intg_buildMCPServer`'s internal `go build` (an unrelated MCP integration test starved by concurrent load), with no compaction/transcript frames on any stack. It passes cleanly and quickly when run without competing jobs.

obra:
Sixth review round (`d9dd370`). The Medium was a regression I introduced in the previous round — fixed in `1693a5e`. The High is the known design-blocked framing item.

**MEDIUM — a restored forced operation was wedged.** Fixed. `restoreCompactionClaimLocked` now re-arms the transient round-tail trigger when it returns a **forced** operation to pending: `applyPendingForceCompact` had already consumed `forceRequested`/`pendingInstructions`, so without the re-arm the operation sat pending with nothing to dispatch it until a restart. It only re-arms when no newer request already armed its own trigger (`!s.forceRequested`), so newer state is preserved. This mirrors what `resumeSkillCompaction` does for the same operation on restart.

Test: `TestSkillCompaction_FailedBatchReArmsForcedTrigger` — a forced operation dispatched over a writer whose fsync and rollback truncate both fail; asserts the operation returns to `pending` **and** `takeForceRequest()` still reports an armed trigger. Fails before the change (“...must re-arm its round-tail trigger, or it is wedged until restart”).

**HIGH — partial batch recovery can lose turns.** Not fixable in this change; same design class as the previously-reported crash-ordering window. A partially written multi-entry batch (rollback truncate then failing) leaves complete leading entries — including the marker — before a partial tail; reopening trims only the partial line, so the marker becomes a resume anchor without its replay-tail copies. Making that atomic needs a commit-record/framing scheme or an atomic rewrite, i.e. a transcript-format/reader change (the #1100 tail-first + `CompactFoldID` design is one form). I did not land a partial version, because without the reader change it either breaks the normal resume path or duplicates history. Flagged for a maintainer decision.

Acceptance on `1693a5e`:
```
$ go test ./agent/ ./agent/transcript/ -count=1 -timeout 1200s
ok  	primeradiant.com/evener/agent	318.405s
ok  	primeradiant.com/evener/agent/transcript	0.721s
$ make lint-golangci
PASS lint-golangci (70s)
$ git show --name-only --format= HEAD | xargs gofmt -l
(empty)
$ gh pr view 2019 --json headRefOid
{"headRefOid":"1693a5e34e0a8cf33033ef023ce5d8b0629a3e3c"}
$ gh pr view 2019 --json mergeStateStatus
{"mergeStateStatus":"BEHIND"}
```

obra:
Seventh round (`1693a5e` → `6cbab046`). One Medium fixed; the partial-write item is reported as impossible within the current on-disk format, with the demonstration requested.

**MEDIUM — stale generation-zero reminder receipt: FIXED.** `restoreCompactionClaimLocked` returned early for a reminder receipt (`Generation == 0`), leaving the handoff it recorded in place; a persisted receipt would then make a restart or later request emit a post-compaction reminder for a compaction whose marker rolled back or never synced. It now withdraws the recorded handoff on **every** non-durable outcome, and only performs the operation-slot restore + forced-trigger re-arm when the receipt has a nonzero generation. Test: `TestSkillCompaction_NonDurableReminderWithdrawsHandoff` (fails before: `still holding [... PublicationID:fold-1 ... Reason:no_captured_operation]`).

**MEDIUM — partial-write rollback failure leaves an un-rollbackable marker: NOT fixable without an on-disk format/reader change.** Per instruction I made the bounded in-format attempts and stopped rather than implementing the format change.

*Attempt 1 — ordering (write the marker last so a torn prefix cannot complete it).* The marker must precede the copies: `ResumeHistory` anchors on the last marker and keeps only entries after it, so copies written before the marker are discarded on a normal restart. I verified this empirically — reordering the batch to put the marker last makes `TestFoldPublication_TurnRecordedDuringFoldSurvivesRestart` fail (`merged-back turns must be durably represented after the compaction marker`). So the marker is necessarily inside any torn prefix.

*Attempt 2 — recovery detects and discards a torn batch.* The reopen path (`agent/transcript/transcript.go:1037-1042`) recovers by truncating to the last complete line boundary; it removes only the unterminated final line and has no on-disk record of where the batch began, so it cannot tell the trailing complete lines came from a torn batch rather than a committed append. There is no signal to detect on.

The two requirements are contradictory with JSONL-as-independent-records: (i) copies must follow the marker, and (ii) a torn write must not be able to complete the marker. A prefix containing the marker exists whenever the marker precedes the copies.

**Format/reader change required** (one of):
- **Framing/commit record:** write a batch as a single framed record (or bracket it with begin/commit sentinels) so an unterminated batch is identifiable; recovery would then discard a batch whose commit record is absent. This matches the review's option (b).
- **Tail-first + `schema.Turn.CompactFoldID`** (PR #1100's design): write the copies first and the marker last, and have the reader drop copies no marker claims. Then a torn prefix contains only copies (unclaimed, dropped) and no marker.

Both change the on-disk format or reader, so they are out of this issue's scope; leaving this PR open for that follow-up.

Acceptance on `6cbab046`:
```
$ go test ./agent/ ./agent/transcript/ -count=1 -timeout 1200s
ok  	primeradiant.com/evener/agent	463.632s
ok  	primeradiant.com/evener/agent/transcript	0.550s
$ make lint-golangci
PASS lint-golangci (83s)
$ git show --name-only --format= HEAD | xargs gofmt -l
(empty)
$ gh pr view 2019 --json headRefOid
{"headRefOid":"6cbab046bef2241928bf6f6b624402a89aa4e8e8"}
$ gh pr view 2019 --json mergeStateStatus
{"mergeStateStatus":"BEHIND"}
```

PR #2024, opened 2026-09-19
fix(sdk): retain draft generation through storage recovery

obra:
A failed local save can reload the same shortcut draft and erase its ready-generation stamp. After a reconnect reuses the same revision, that could make an old draft appear saveable again. This change preserves the stamp when the classified stored checkpoint has the same identity, while treating replacement records as newly restored drafts.

This is the immediate recovery successor to merged #2014 and the final small replacement for #1792. It addresses local RoboRev #2644 and remote member #22318.

Validation: 125 focused tests, production reversal checks, TypeScript build, package qualification, import lint, and scoped Biome passed. Independent correctness/simplification review and local RoboRev #2648 passed. Restacking onto merged #2014 preserves the complete binary patch exactly (SHA-256 c92cca66ae7a9b632403272c93fdede974ed77b0cc612a288d9c7b4ccd93d52c). Current-head CI and raw remote review are required before merge.


obra:
Coordinator disposition for current full head `d521522b9933933b1d1d1e7db99713936f7acffc`, based on the complete raw receipt `reviews/raw/2024-d521522b9-remote.md`:

- The raw-identity coverage **Low** is tracked in [#2026](https://github.com/prime-radiant-inc/evener/issues/2026). Current tests change decoded fields (`id`), so they do not catch a same-decoded-fields/different-canonical-raw replacement.
- The follow-up requests behavioral coverage with an extra/removed raw field and does not prescribe brittle JSON-string comparison.
- #2024's implementation scope is unchanged; this Low is separate and not folded into the branch. Current-head CI and remote review qualification remain pending.

PR #2025, opened 2026-09-19
fix(providers): a removal sets an instance's OAuth record aside, and recovery reclaims it

obra:
## What this is

The crash-safe machinery for provider-instance removals and renames, split out of #1607 (closed as superseded) and consolidated onto one record.

A removal no longer deletes an instance's credentials outright: it stages them. One write-ahead intent record per mutation says what was in flight, and startup resolves every record by one rule.

- **One intent record per mutation** (`<inst>.json.intent-<stamp>`): the op (remove/rename), the instance, the new name, and the removal's kind (config-backed or credential-only). It is written atomically before the first destructive step, whenever a stored credential may be present, and spent once the mutation resolves.
- **The commit point is the record's own name**: a removal's landing is filed by a no-replace rename to `<inst>.json.landed-<stamp>`, with the directory fsynced — one atomic syscall, so the commit cannot be torn, and a rewrite that loses its race leaves the record exactly as it was. A rename needs no commit marker at all: its landing is `providers.toml`, durable before any credential moves. A removal whose landing rename fails rolls back and is never reported as standing.
- **Staged credentials.** The OAuth record moves to a stamped copy (`<inst>.json.removing-<stamp>`), and the stored key is staged as a fsynced `0600` sidecar beside the record before the removal clears it. Recovery puts each back for a removal that did not stand — never overwriting a key filed since — and spends them with a standing removal. The copies carry no classification in their names; what they are for is read from the record that staged them.
- **One recovery rule**: read the auth directory once, group a name's records and copies once, finish renames first (oldest mutation first), then decide each name. A removal stands by its own evidence — the config for a config-backed removal, the landed name for a credential-only one; a standing removal's staged bytes are swept, everything a name still wants is put back newest-first, and every failed step is reported by name. An unreadable config holds a config-dependent name back whole.
- **Rollback**: a failed removal's rollback returns its staged copy and key; when the carrying copy cannot be put back, the removal stands and is reported through main's `instanceRemoveApplied` discriminator.
- **No-replace moves are atomic** where the platform allows: `renameat2(RENAME_NOREPLACE)` on Linux, with a checked fallback whose residual window and the lock discipline around it are documented at the function.

This replaced five filename-encoded record types (four aside-marker spellings, a per-removal commit manifest, a rename marker, a rename journal) with the one record, and the nine-scan recovery pass with the one above.

## Disclosed, not fixed here

- **The TUI credential actions carry no endpoint fingerprint** (disclosed on #1725): a stale TUI can test, set, replace or clear credentials for a re-pointed name.

## Accepted at review

The review of this head listed one high and three mediums; the maintainer's call is that the current behavior is acceptable, so they are recorded rather than fixed here:

- The rename's `providers.toml` write is not fsynced before the credentials move - the config writer's pre-existing behavior on main, unchanged by this PR - so an unclean power loss can leave the credential under the new name while the config names the old one. Process crashes are unaffected, and every credential move itself is fsynced.
- `reclaimOAuthAsides` can leave an earlier removal's staged-key sidecar behind when it removes that removal's record.
- One park move still uses a plain rename rather than the no-replace primitive.
- A restore blocked by an occupied record path is treated as settled, so that record is spent while the parked copy remains.

## Verification

- `gofmt`, `make vet`, `make lint-golangci`, `make lint-evenerfuzz`, `make lint-generated`: clean.
- `go test ./cmd/evener-hub/... -count=1`: green. The `./agent/...` suite carries two scratch-retention failures that reproduce on pristine main on the authoring host (disk-pressure sensitive); CI is the arbiter for them.
- The 40-rule mutation harness over this tree: every mutation caught, restored files green. Every crash-point outcome the previous machinery pinned keeps a test; the rules whose mechanism the consolidation deleted are retired in the harness with a one-line reason each.


obra:
Closing without merging, by the maintainer's call: main's existing removal behavior is the one we want, so the credential staging machinery does not land. The branch (feat/asides-reclaim, head 53c708820a) and its review history remain for reference; the findings recorded under "Accepted at review" in the body were reviewed and accepted rather than fixed. #1607 was closed as superseded by this one, and both remain closed.

PR #2032, opened 2026-09-19
fix(sandbox): export TMPDIR as a shared, sticky scratch temp dir (#495)

### LINKED ISSUE #495: Session scratch is exported as TMPDIR at mode 0700, so any privilege-dropping child fails with Permission denied on a path the user never chose
Evener exports its session scratch dir as `TMPDIR` for every spawned process, and that directory is mode `0700`. Any child that drops privileges inherits the variable, cannot write the directory, and fails with a `Permission denied` naming a path the user never chose.

## Mechanism

`agent/sandbox/session_scratch.go:55` secures the scratch dir:

```go
if err := os.Chmod(dir, 0o700); err != nil {
```

`agent/sandbox/env_floor.go:150` (`ApplySessionScratchEnv`) then installs that path as both `TMPDIR` and `EVENER_SCRATCH_DIR` for spawned processes, and the environment preamble advertises it:

```
Scratch ($EVENER_SCRATCH_DIR, $TMPDIR): /tmp/evener-sandbox-3424735479
```

`0700` is right for the agent's own scratch. The problem is the pairing: `TMPDIR` is inherited by every descendant, including ones that deliberately run as another user, and for those the value is unusable.

## Field evidence

TB 2.1 trial `configure-git-webserver__52rZDfB`, step 85. A git `post-receive` hook, running as the pushing user rather than the agent user:

```
remote: mktemp: failed to create directory via template
        '/tmp/evener-sandbox-3424735479/tmp.XXXXXXXXXX': Permission denied
```

It cost one debugging round-trip there. The general shape is any privilege-dropping workflow: `sudo -u`, `su`, a setuid daemon, a service manager starting a unit under its own account, a build that shells out to a less-privileged worker.

## Why it is worth fixing rather than documenting

The failure is hard to attribute from inside. The child reports a path the user never wrote, in a temp directory it has no reason to know about, with a permission error that looks like a bug in the child's own setup. Nothing in the message connects it to the agent's sandbox.

Candidate directions, in rough order of how well they preserve the current guarantee:

1. Keep the private scratch as `EVENER_SCRATCH_DIR` and stop overriding `TMPDIR` — the two are separate needs and only one of them has to be private.
2. Provide a second, group- or world-writable subdirectory (sticky, like `/tmp` itself) and point `TMPDIR` at that, leaving the private scratch for the agent's own use.
3. Keep the current behavior but make the failure legible: detect a privilege change in a spawned command and warn that `TMPDIR` will not be writable by it.

(1) is the smallest and gives up the least — the reason the scratch is `0700` is to keep the agent's working files private, which does not require that every descendant's temp files land there too.

obra:
Fixes #495

## Root cause

The session scratch is secured with `os.Chmod(dir, 0o700)` (`agent/sandbox/session_scratch.go:56`) and that same path is installed as **both** `TMPDIR` and `EVENER_SCRATCH_DIR` for every spawned process (`agent/sandbox/env_floor.go`, `ApplySessionScratchEnv`; and `agent/execenv/local.go`, `overlaySessionEnv`, for unsandboxed sessions). `TMPDIR` is inherited by every descendant — including one that deliberately runs as another user — so such a child inherits a `TMPDIR` it cannot write and fails with a bare `Permission denied` on a path the user never chose. Field evidence: TB 2.1 trial `configure-git-webserver__52rZDfB` step 85, a git `post-receive` hook running as the pushing user.

`0700` is right for the agent's own scratch. The defect is exporting that private directory as a world-facing inherited variable.

## Why not simply stop overriding TMPDIR

`TMPDIR` has to stay inside a sandbox-granted writable root. Linux gets a fresh writable `/tmp` tmpfs inside the sandbox, but the macOS Seatbelt platform defaults deliberately do **not** grant `/tmp`/`/private/tmp` — `agent/sandbox/seatbelt_platform_defaults.sbpl` says so explicitly and relies on the per-session TMPDIR being granted separately as a writable root. Dropping the override would break every sandboxed temp write on macOS.

## Change (the triage's option 2)

The two variables now name different directories:

- `EVENER_SCRATCH_DIR` stays the private scratch. Its mode becomes `0711`: the owner keeps `rwx`, group/other get **traverse only** — no world read, no world write — so the named temp child below is reachable while the scratch's entries stay non-listable and non-writable to anyone else.
- `TMPDIR` names a dedicated `tmp` subdirectory of the scratch, created (and repaired on restore) with mode `1777`: world-writable with the **sticky bit**, exactly like `/tmp`. Any user may create temp files there; no user may remove or rename a file they do not own. Cleanup still works because the session owns the sticky directory.

No other permission is widened: the scratch itself never becomes world-writable or world-readable.

The capability preamble and the sandbox prompt now report the private scratch and the shared temp directory as the distinct paths they are.

## Test evidence

`TestSessionScratchSharedTmpIsUsableByPrivilegeDroppingChildren` (`agent/sandbox/session_scratch_test.go`) fails before the change and passes after. Before (production code temporarily reverted to the old `0700` + no shared child):

```
--- FAIL: TestSessionScratchSharedTmpIsUsableByPrivilegeDroppingChildren (0.00s)
    session_scratch_test.go:63: shared temp directory "/tmp/.../evener-sandbox-548398425/tmp" must exist: stat ...: no such file or directory
```

The kernel's check for a foreign user is entirely permission bits, so the test asserts exactly those: the shared temp dir is `0777` + sticky, and the private scratch stays traversable-but-not-readable/writable (o+x, no o+r, no o+w).

Commands run, all green:

```
go test ./agent/... ./envvars/... -count=1        # full agent subtree
go test ./agent/sandbox/ ./agent/execenv/ ./envvars/... -count=1
go vet ./agent/sandbox/... ./agent/execenv/... ./agent ./envvars/...
gofmt -l <changed files>
go build ./...
```

The existing "shell TMPDIR == EVENER_SCRATCH_DIR == scratch" assertions were updated to the new contract — the two variables are now different directories by design — and the file-tool/private-scratch guard (`TestWriteFileReachesSessionScratch`, `TestEnableSandboxFileToolsReachOwnScratch`) is unchanged and still green.


obra:
Dispositions for the roborev review of `62822f5` (now superseded by head `121725cb85`, which carries every fix below).

**High — symlink traversal before `Chmod`: FIXED.** `ensureSessionScratchTmp` is gone. The new `ensureScratchSubdir` Lstats the path first and `os.Remove`s any non-directory entry (Remove unlinks a symlink itself, never its target), creates the directory, and then applies the mode through an `O_NOFOLLOW|O_DIRECTORY` handle (`openScratchDirNoFollow`, new `session_scratch_subdir_unix.go`; a reparse-point-checked open in the windows twin) so a swap cannot redirect the fchmod either. Test `TestPrepareSessionScratchRefusesSymlinkedSubtrees` plants a real symlink at `tmp`; before the fix its target was chmod'ed to `0777`:

```
--- FAIL: TestPrepareSessionScratchRefusesSymlinkedSubtrees (0.00s)
    session_scratch_test.go:127: prepareSessionScratch followed the planted symlink and chmod'ed its target: victim mode = 0777, want 0755
```

After the fix the victim stays `0755` and `tmp` is a real `1777` directory.

**High — restored scratches stay untraversable: FIXED.** `prepareSessionScratch` now repairs the container mode and both subtrees, and `OpenRetainedSessionScratch` calls it under the held lease, so a pre-fix `0700` allocation is migrated rather than resurrected. Test `TestPrepareSessionScratchRepairsPreFixLayout` starts from a bare `0700` container and asserts `0711` + `private` `0700` + `tmp` `1777`; it fails before the change.

**Medium — `0711` exposes world-readable artifacts: FIXED (the reviewer's first option).** Private state now lives one level down behind a `0700` boundary. The scratch has three parts: `<scratch>` `0711` (container, traverse only, holds no session files), `<scratch>/private` `0700` (`EVENER_SCRATCH_DIR`, the redirected caches and the model's file tools), `<scratch>/tmp` `1777` sticky (`TMPDIR`). A `0644` artifact written under `EVENER_SCRATCH_DIR` is therefore unreachable by another user, and only the dedicated temp subtree is world-writable. The capability preamble, the sandbox prompt and `docs/sandboxing.md` now describe the three-part layout rather than calling the container private. `TestSessionScratchSharedTmpIsUsableByPrivilegeDroppingChildren` pins all three modes.

**Medium — allocator accepts an unreachable base: REFUTED, with evidence.** The base is host/operator-selected (`os.TempDir()`, else `os.UserCacheDir()`). When it sits under a `0700` ancestor, *no* directory Evener can create there is reachable by another uid — traversal requires `o+x` on every ancestor, so the child's own mode is irrelevant. The only escape is relocating the scratch to a world-traversable base, and for the sandboxed case that cannot work either: the one world-traversable candidate is `/tmp`, which `agent/sandbox/seatbelt_platform_defaults.sbpl` explicitly refuses to grant ("evener deliberately does NOT grant broad read/write of /tmp"), so a `/tmp`-based temp dir would be outside the sandbox's write grant. Concretely: on Linux the default base is `/tmp` (`1777`), so the fix applies as intended; on macOS `os.TempDir()` is the per-user `/var/folders/.../T/` (`0700`), where the shared subtree is unreachable by construction. In that configuration behaviour is identical to before this PR (`TMPDIR` unusable by another uid) — an unimproved corner, not a regression. Changing base selection would also change where the session's own files live and alter the existing base-preference and workspace-exclusion semantics, which is wider than this issue. Left as a documented limitation rather than silently expanding scope.

**Medium — fuzz-tagged seed tests assert the old `TMPDIR` location: FIXED.** `agent/sandbox/fuzz_policy_assembly_test.go`, `agent/execenv/process_runtime_program_fuzz_test.go` (which was also missing the `sandbox` import in its tagged build) and `agent/execenv/sandbox_lifecycle_program_fuzz_test.go` now compare against `SessionScratchTmpDir`/`SessionScratchPrivateDir`. `make fuzz` is green.

**Low — test assertion misses group read/write: FIXED.** `perm&0o006` → `perm&0o066`.

**Doc drift: FIXED.** The comment now names `OpenRetainedSessionScratch`.

Acceptance on `121725cb85`:

```
$ go test ./agent/sandbox/ ./agent/execenv/ -count=1
ok  	primeradiant.com/evener/agent/sandbox	3.103s
ok  	primeradiant.com/evener/agent/execenv	4.948s

$ make fuzz
...
ok  	primeradiant.com/evener/appwire	0.008s        (exit 0)

$ git show --name-only --format= HEAD | grep '\.go$' | xargs gofmt -l
(empty)

$ go vet ./agent/sandbox/... ./agent/execenv/... ./agent ./envvars/...   -> clean
```

obra:
Round-3 dispositions for the review of `121725c` (superseded by head `d638888c11`, which carries every fix below). The layout stays three-path: container `0711` / `private` `0700` / `tmp` `1777`+sticky.

**HIGH 1 — legacy root artifacts exposed when the container is loosened to `0711`: FIXED (policy: relocate, as ruled).** `prepareSessionScratch` now orders the migration so nothing is ever observable through a traversable container: it hardens the contents of any legacy entry that already carries a reserved name, creates `private/`, MOVES every other direct entry into it (numeric suffix on collision, never a clobber) and tightens each moved entry to owner-only (`0600` files preserving the owner's execute bit, `0700` directories, symlinks skipped and never followed), and only then loosens the container to `0711`. Relocating rather than "don't loosen" is deliberate: a container left at `0700` would hide the temp subtree from a privilege-dropping child, re-breaking the round-1 High and the original #495 failure for exactly the restored scratches this fix exists for. `TestPrepareSessionScratchRelocatesLegacyRootArtifacts` covers a POPULATED legacy directory (a `0644` file, a nested `0755` cache subtree, a legacy `tmp/` holding a `0644` file) plus a name collision. Before the change:

```
--- FAIL: TestPrepareSessionScratchRelocatesLegacyRootArtifacts (0.00s)
    session_scratch_test.go:200: legacy artifact was not relocated intact: "" err open /tmp/.../evener-sandbox-legacy/private/report.txt: no such file or directory
```

After: the artifact and the nested cache live under `private/` at `0600`/`0700`, the legacy `tmp/legacy.tmp` is `0600` while `tmp` is `1777`, the colliding name took `report.txt.1` instead of clobbering, and the only direct container entries left are the two subtrees and Evener's metadata (lease and pin, both `0600`).

**MEDIUM 3 — grants scoped to the container instead of the private subtree: FIXED.** `allocatedSessionScratchPath()` now returns the private subtree, so `e.sandbox()` (the file-tool layer), `retireStaleFileToolLayers`, `scratchSandboxFor` and `ensureWritePath` all grant the `0700` subtree; a `write_file` can no longer deposit a `0644` artifact in the traversable container. `TestFileToolsRefuseTheScratchContainerRoot` asserts the container root is refused and the private subtree still writable; before the change:

```
--- FAIL: TestFileToolsRefuseTheScratchContainerRoot (0.00s)
    sandbox_tools_test.go:646: write_file must not create an artifact directly under the traversable scratch container "/tmp/.../evener-sandbox-1665106836"
```

Naming the residual honestly: the *spawned* sandbox grant is still the container, because `TMPDIR` must be a sibling of `private` under the single root the wrapper is given. The model's file tools cannot use it; a spawned command that deliberately targeted the container's absolute path could still create a file there, and the container path is no longer advertised anywhere (prompt, preamble and docs name `private` and `tmp`). Any such leftover is relocated by the next restore. Two-root wrapper plumbing would close it fully and is a larger change than this issue.

**LOW 5 — `SessionScratchDir()` vs `$EVENER_SCRATCH_DIR`: FIXED by making them agree.** The terminal `scratch_path` metadata (`agent/subagents.go`) now records `SessionScratchPrivateDir(SessionScratchDir())`, which IS the path `$EVENER_SCRATCH_DIR` names; the accessor itself still reports the allocation container, which is what the grant/caching paths need. The comment says which is which.

**LOW 6 — prefix assertions: FIXED.** `agent/sandbox_delegate_create_test.go` now asserts `strings.Contains(got, sandbox.SessionScratchPrivateDir(scratch))` at both sites, so a prompt naming the container (the pre-split behaviour) fails rather than passing on the prefix.

**MEDIUM 4 — ancestor base not traversable: REFUTED, restated with the configuration split.** *Handled:* a base whose whole ancestor chain is other-traversable — the Linux default, `os.TempDir()` = `/tmp` (`1777`). *Not handled:* a base under a `0700` ancestor — macOS's per-user `/var/folders/.../T/`, the `os.UserCacheDir()` fallback, or a custom `TMPDIR` under a private tree. Why that is acceptable: traversal needs `o+x` on every ancestor, so no directory Evener can create under such a base is reachable by another uid regardless of the child's own mode; the only world-traversable candidate is `/tmp`, which `agent/sandbox/seatbelt_platform_defaults.sbpl` deliberately does not grant, so relocating there would not make `TMPDIR` usable *inside the sandbox* on macOS; and changing base selection would relocate the session's own files and alter the existing base-preference and workspace-exclusion semantics, which is wider than this issue. In the not-handled configuration behaviour is identical to before this PR (`TMPDIR` unusable by another uid) — an unimproved corner, not a regression.

**HIGH 2 — TOCTOU on the container path: REFUTED, with the residual named.** Child entries are created and opened by path, not fd-relative, so a same-uid process could in principle swap an entry between the `Lstat`/`Mkdir` and the open. Three things bound it: every open goes through `O_NOFOLLOW|O_DIRECTORY` (`openScratchDirNoFollow`), so the final component can never be a symlink and cannot redirect the fchmod; `Mkdir` fails closed on an existing name; and the container is `0711` with no group/other write, its ancestors are either sticky `/tmp` or a `0700` user directory, and a restore holds the scratch lease — so no *other* user can swap anything. **Residual, stated plainly:** a same-uid process can redirect our chmod onto another directory that uid already owns. That is not a privilege escalation — the same uid can already `chmod` that directory directly — so I am not adopting `openat`/`mkdirat`/`fchmodat` plumbing for it this round.

Preserved round-1 invariants (all still in place and green): the `Lstat` + `O_NOFOLLOW` fchmod in `ensureScratchSubdir`; restored-scratch mode repair; the `0711` container with the `0700` private subtree and `1777` sticky temp subtree; the evenerfuzz-tagged seed assertions; the `perm&0o066` mask; and the doc-drift fix (`OpenRetainedSessionScratch`).

Acceptance on `d638888c11`:

```
$ go test ./agent/sandbox/ ./agent/execenv/ -count=1
ok  	primeradiant.com/evener/agent/sandbox	2.776s
ok  	primeradiant.com/evener/agent/execenv	4.205s

$ make fuzz
... ok primeradiant.com/evener/appwire (cached)      -> exit 0

$ cd agent && go test -tags evenerfuzz -run '^Fuzz' -count=1 ./sandbox/ ./execenv/
ok  	primeradiant.com/evener/agent/sandbox	2.317s
ok  	primeradiant.com/evener/agent/execenv	1.318s

$ git show --name-only --format= HEAD | grep '\.go$' | xargs gofmt -l
(empty)

$ go vet ./agent/sandbox/... ./agent/execenv/...     -> clean
```
Also green: `go test ./agent/... ./envvars/... -count=1` across the whole agent subtree.

obra:
Round-4 dispositions for the review of `121725c`→`d638888` (superseded by head `991d78aa19`, which carries every fix below). The three-path layout is unchanged: container `0711`, `private` `0700`, `tmp` `1777`+sticky.

**HIGH — wrapper binds the writable container instead of the private subtree: FIXED.** The container is now traversal/read only and only the two exported subtrees are writable, derived from the container by the backends themselves so the `Wrapper` API and every caller are unchanged (`sandbox.SessionScratchWriteRoots`):
- `buildBwrapArgv` emits `--ro-bind <container> <container>` and then `--bind <container>/private` / `--bind <container>/tmp` (skipping a subtree that does not exist), so the subshell `$EVENER_SCRATCH_DIR/..` write is refused by the kernel.
- `SeatbeltPolicy` keeps the container in the read roots (needed for traversal) and drops it from `writeRootKeys`, which now grant the two subtrees; the six SBPL goldens were regenerated accordingly (each gained `WRITABLE_ROOT_1`).

Failing-before evidence (production change temporarily reverted):
```
$ go test ./agent/sandbox/ -run 'TestBuildBwrapArgvSessionTmp|TestBuildBwrapArgvReadOnlySessionTmp'
--- FAIL: TestBuildBwrapArgvSessionTmp (0.00s)
    bwrap_test.go:272: expected the session scratch container … bound read-only
--- FAIL: TestBuildBwrapArgvReadOnlySessionTmp (0.00s)
```
Three tests pin it now: `TestBuildBwrapArgvSessionTmp` / `TestBuildBwrapArgvReadOnlySessionTmp` (container is `--ro-bind`, never `--bind`; both subtrees bound writable), `TestSeatbeltWriteRootsAreTheScratchSubtrees` (subtrees are write roots, container is not, container stays readable), and the real-sandbox `TestBwrapReadOnlyTmpRootsPreserveAccess`, whose script now writes `$3/private/written` and `$3/tmp/scratch` successfully while a write to `$3/escaped` exits non-zero.

**MEDIUM — legacy migration deletes user data at reserved names: FIXED.** A regular file sitting at `private`/`tmp` is session data that merely shares a name Evener reserves, so `freeReservedSubtreeNames` moves it aside inside the container under a free non-reserved name *before* the subtree is created, and the relocation pass then preserves it under `private/` tightened to owner-only. Only a symlink or other non-regular entry is left for `ensureScratchSubdir` to replace; Evener's own metadata (lease, pin) is never touched. `TestPrepareSessionScratchPreservesRegularFileAtReservedName` covers both reserved names; before the fix:

```
--- FAIL: TestPrepareSessionScratchPreservesRegularFileAtReservedName (0.00s)
    session_scratch_test.go:213: a legacy regular file at a reserved name was destroyed (payload "legacy data" gone); found map[]
```

**LOW — live delegate test: FIXED.** `agent/sandbox_delegate_live_test.go` now reads `SessionScratchPrivateDir(scratch)/report.txt` (where `$EVENER_SCRATCH_DIR` points) and additionally asserts nothing named `report.txt` lands directly in the container.

**LOW — stale doc comments: FIXED.** `Wrapper.sessionTmp`, `Wrapper.SessionTmp()` and `LocalExecutionEnvironment.SessionScratchDir()` now say plainly that they return the traversal-only *container*, and name `SessionScratchTmpDir` / `SessionScratchPrivateDir` as the two paths a caller must derive — the mislabel that let the live test drift.

**Preserved invariants** (all still asserted and green): `Lstat` + `O_NOFOLLOW` fchmod in `ensureScratchSubdir`; restored-scratch mode repair; `0711` container + `0700` private + `1777` sticky temp; evenerfuzz-tagged seed assertions; `perm&0o066`; doc-drift fix; legacy relocation with collision suffixing; file-tool grant scoped to `private`; `scratch_path` metadata and prompt assertions naming the private subtree.

Acceptance on `991d78aa19`:

```
$ go test ./agent/sandbox/ ./agent/execenv/ -count=1
ok  	primeradiant.com/evener/agent/sandbox	4.897s
ok  	primeradiant.com/evener/agent/execenv	13.213s

$ make fuzz                                          -> exit 0

$ cd agent && go test -tags evenerfuzz -run '^Fuzz' -count=1 ./sandbox/ ./execenv/
ok  	primeradiant.com/evener/agent/sandbox	1.762s
ok  	primeradiant.com/evener/agent/execenv	1.493s

$ git show --name-only --format= HEAD | grep '\.go$' | xargs gofmt -l   -> (empty)
$ go vet ./agent/sandbox/... ./agent/execenv/...                        -> clean
$ go test ./agent/ -count=1 -timeout 20m    -> ok primeradiant.com/evener/agent 624.164s
```
(The first `./agent/...` run reported the package at exactly 600s; that was Go's default 10-minute test timeout under a machine load average of ~28 from sibling lanes, not a failure — the isolated re-run with an explicit timeout passes.)

obra:
Round-5 dispositions for the review of `991d78a` (superseded by head `e070bb4fb1`, which carries the fix below).

**MEDIUM — unvalidated scratch layout assumed by the wrapper: FIXED.** `NewWrapper` now provisions the container's two exported subtrees when it is handed an existing container directory that `NewSessionScratch` did not lay out, and fails closed instead of emitting a sandbox whose `$TMPDIR`/`$EVENER_SCRATCH_DIR` do not exist. It is idempotent for every real session scratch (`prepareSessionScratch` has already run), it refuses — and never unlinks — a non-directory entry squatting on a subtree name, and it leaves a container path that does not exist alone, because inventing a tree for a caller that passed a path it never created would hide the caller's own error rather than fix the sandbox.

```
$ go test ./agent/sandbox/ -run TestNewWrapperProvisionsExportedSubtrees   # with the call removed
--- FAIL: TestNewWrapperProvisionsExportedSubtrees (0.00s)
    backend_test.go:32: the wrapper must provision "…/001/private" so $TMPDIR/$EVENER_SCRATCH_DIR exist: info=<nil> err=lstat …: no such file or directory
```
`TestNewWrapperProvisionsExportedSubtrees` pins the provisioning and the exported modes (`0700` / `1777`+sticky); `TestNewWrapperRefusesOccupiedSubtreeName` pins that a caller's own file at a subtree name is refused and left intact.

**MEDIUM — shared `TMPDIR` ancestor permission gap: REFUTED (third raise of the item refuted in rounds 1 and 3), with the Seatbelt wording addressed directly.** *Handled:* any base whose entire ancestor chain is other-traversable — the Linux default, `os.TempDir()` = `/tmp` at `1777`. *Not handled:* a base under a `0700` ancestor — macOS's per-user `/var/folders/.../T/`, the `os.UserCacheDir()` fallback, or a custom `TMPDIR` under a private tree. Traversal needs `o+x` on every ancestor, so no directory Evener can create under such a base is reachable by another uid whatever the child's own mode is; Seatbelt uses the host path but the sandbox policy is orthogonal to host traversal, so that does not change the analysis. The only world-traversable candidate is `/tmp`, which `agent/sandbox/seatbelt_platform_defaults.sbpl` deliberately does not grant ("evener deliberately does NOT grant broad read/write of /tmp") — relocating the shared subtree there would leave `TMPDIR` unwritable *inside* the sandbox on macOS, i.e. it would reinstate exactly the #495 failure this PR exists to fix, and the reviewer's alternative ("avoid advertising the shared `TMPDIR` contract") is that same regression by another route. In the not-handled configuration behaviour is identical to before this PR: an unimproved corner, not a regression. Changing base selection would additionally relocate the session's own files and alter the base-preference and workspace-exclusion semantics, which is wider than this issue.

Acceptance on `e070bb4fb1`:

```
$ go test ./agent/sandbox/ ./agent/execenv/ -count=1
ok  	primeradiant.com/evener/agent/sandbox	3.970s
ok  	primeradiant.com/evener/agent/execenv	5.127s

$ make fuzz                                          -> exit 0
$ cd agent && go test -tags evenerfuzz -run '^Fuzz' -count=1 ./sandbox/ ./execenv/
ok  	primeradiant.com/evener/agent/sandbox	1.685s
ok  	primeradiant.com/evener/agent/execenv	0.987s

$ git show --name-only --format= HEAD | grep '\.go$' | xargs gofmt -l   -> (empty)
$ go vet ./agent/sandbox/... ./agent/execenv/...                        -> clean
$ go test ./agent/ -count=1 -timeout 20m    -> ok primeradiant.com/evener/agent 362.653s
```

obra:
Round-6 dispositions for the review of `e070bb4` (superseded by head `f1c7c07033`, which carries both fixes).

**MEDIUM — ancestor traversal: FIXED at allocation time.** `scratchBaseTraversable` walks a base's whole ancestor chain and requires other-user execute (`o+x`) on every component, which is exactly the kernel check a dropped UID performs to reach `<base>/<scratch>/tmp`. A **defaulted** base is now chosen for reachability: the allocator takes the preferred candidate only when its chain is traversable, otherwise it tries the user-cache candidate, and if neither qualifies it falls back to the ordinary preference so a host whose only bases sit under private ancestors still starts a session (in that configuration behaviour is unchanged from before this PR — an unimproved corner, not a regression). A base the **caller names explicitly** is honoured exactly as asked: it is the operator's ruling, `TestSessionScratchDoesNotChmodCandidateBase` pins that a caller's base is never repaired, and 57 existing fixtures pass an explicit `t.TempDir()`.

The demanded test through a private ancestor is `TestSessionScratchBasePrefersATraversableBase`: the defaulted candidate sits under a `0700` ancestor and the cache candidate under a genuinely traversable one; it asserts the scratch is allocated under the traversable base *and* that every component of the resulting path grants other-user traverse. Before the change (the traversability preference removed):

```
--- FAIL: TestSessionScratchBasePrefersATraversableBase (0.00s)
    session_scratch_test.go:57: scratch "…/private-ancestor/base/evener-sandbox-1858874914" must be allocated under the traversable base "/tmp/evener-scratch-public-3900107392", not the private ancestor chain
```

**LOW — hard-link risk: FIXED.** `hardenRelocatedEntry` now checks the link count of each relocated regular file (`linkCount` via `syscall.Stat_t.Nlink` on unix; the platform that does not expose it reports unknown) and, when a file has more than one link, replaces it with a private copy on a fresh inode before applying the tightened mode — so a legacy artifact hard-linked to a worktree file can no longer tighten that outside file. `TestPrepareSessionScratchBreaksHardLinksBeforeTightening` links a scratch artifact to a file outside the scratch and asserts the outside file keeps `0644`, the relocated copy is `0600` on a different inode, and the content survives. Before the change:

```
--- FAIL: TestPrepareSessionScratchBreaksHardLinksBeforeTightening (0.00s)
    session_scratch_test.go:324: a file outside the scratch was tightened through a hard link: mode = 0600, want 0644
```

**Security property intact**: the scratch container is still `0711` (traverse only, never world-writable or world-readable), the private subtree stays `0700`, only the shared temp subtree is `1777`+sticky, and the allocation rule now additionally guarantees the dropped UID can traverse the chain *down to* the scratch.

Acceptance on `f1c7c07033`:

```
$ go test ./agent/sandbox/ ./agent/execenv/ -count=1
ok  	primeradiant.com/evener/agent/sandbox	3.965s
ok  	primeradiant.com/evener/agent/execenv	5.783s

$ make fuzz                                          -> exit 0
$ cd agent && go test -tags evenerfuzz -run '^Fuzz' -count=1 ./sandbox/ ./execenv/
ok  	primeradiant.com/evener/agent/sandbox	1.966s
ok  	primeradiant.com/evener/agent/execenv	1.339s

$ git show --name-only --format= HEAD | grep '\.go$' | xargs gofmt -l   -> (empty)
$ go vet ./agent/sandbox/... ./agent/execenv/...                        -> clean
$ go test ./agent/ -count=1 -timeout 20m   -> ok primeradiant.com/evener/agent 513.836s
$ go test ./cmd/evener/ -run 'Scratch|Reclaim' -count=1                 -> ok
```

obra:
CI fix: `lint-golangci` failed on `f1c7c07` with a single `unconvert` finding in the new link-count helper (`uint64(stat.Nlink)` is a no-op on linux, where `Nlink` is already uint64). `linkCount` now returns `int`, narrowing the platform-dependent width once inside the helper; the windows twin matches. Reproduced and cleared locally with `golangci-lint run --config .golangci.yml ./agent/sandbox/... ./agent/execenv/...` → `0 issues.`, and the windows/darwin cross-vets (`GOOS=windows|darwin go vet -tags evenerfuzz`) are clean. The `static` check failed only as the aggregator over `lint-golangci`; every other check on that head passed. New head `9abc050457` carries the fix.

obra:
Round-7 dispositions for the review of `9abc050` (superseded by head `60b5922749`, which carries all four).

**MEDIUM — world-writable non-sticky base trusted: FIXED.** The defaulted-base rule is now `scratchBaseShared`: the ancestor chain must grant other-user traverse **and** a base that is world-writable must carry the sticky bit, or any local user could delete or replace this session's container (and with it the sandbox's grant target). `TestSessionScratchBaseRejectsWorldWritableNonStickyBase` builds a `0777` non-sticky base beside a `1777` one and asserts the allocation avoids the unprotected base; failing-before with the rule disabled:
```
--- FAIL: TestSessionScratchBaseRejectsWorldWritableNonStickyBase (0.00s)
    session_scratch_test.go:105: a world-writable, non-sticky base "/tmp/evener-scratch-unprotected-2598189559" must not count as shared
```

**MEDIUM — `NewWrapper` partially provisioned the container: FIXED.** It now runs `prepareSessionScratchForWrapper`, which puts the whole directory into the exported layout — traversable container, `0700` private subtree, `1777`+sticky temp subtree, legacy direct entries migrated — and still refuses, never unlinks, a non-directory entry squatting on a subtree name (so `TestNewWrapperRefusesOccupiedSubtreeName` keeps passing). It is idempotent for a container `NewSessionScratch`/restore already laid out, so production pays nothing; the container mode assertion in `TestNewWrapperProvisionsExportedSubtrees` fails before:
```
--- FAIL: TestNewWrapperProvisionsExportedSubtrees (0.00s)
    backend_test.go:33: wrapper container mode = 0755, must not become group/other readable or writable
```

**MEDIUM — `BorrowRetainedSessionScratch` bypassed the layout: FIXED.** It now validates the borrowed allocation and refuses one its owner has not migrated (still `0700`, or missing a subtree) instead of publishing a scratch whose exported temp path is unreachable to a dropped-privilege child. `TestBorrowRetainedSessionScratchRefusesUnpreparedLayout` covers both directions; failing-before with the validation removed:
```
--- FAIL: TestBorrowRetainedSessionScratchRefusesUnpreparedLayout (0.00s)
    scratch_retention_test.go:45: a scratch that is not in the exported layout ("…/evener-sandbox-legacy") must not be borrowed
```

**MEDIUM — the new test broke on symlink-`/tmp` hosts: FIXED** exactly as suggested — `TestSessionScratchBasePrefersATraversableBase` canonicalizes the fixture base with `filepath.EvalSymlinks` before the prefix assertion, so macOS (`/tmp -> private/tmp`) sees the resolved spelling the allocator returns. The same canonicalization is applied in the sticky-base test.

Acceptance on `60b5922749`:
```
$ go test ./agent/sandbox/ ./agent/execenv/ -count=1
ok  	primeradiant.com/evener/agent/sandbox	2.756s
ok  	primeradiant.com/evener/agent/execenv	4.100s
$ make fuzz                                          -> exit 0
$ cd agent && go test -tags evenerfuzz -run '^Fuzz' -count=1 ./sandbox/ ./execenv/
ok  	primeradiant.com/evener/agent/sandbox	1.548s
ok  	primeradiant.com/evener/agent/execenv	0.785s
$ git show --name-only --format= HEAD | grep '\.go$' | xargs gofmt -l   -> (empty)
$ go vet ./agent/sandbox/... ./agent/execenv/...                        -> clean
$ go test ./agent/... ./envvars/... -count=1 -timeout 30m               -> all ok
$ go test ./cmd/evener/ -run 'Scratch|Reclaim' -count=1                 -> ok
$ golangci-lint run --config .golangci.yml ./agent/sandbox/... ./agent/execenv/...  -> 0 issues.
```

obra:
Round-8 dispositions for the review of `60b5922` (superseded by head `8c9d9f870a`, which carries all five).

**HIGH — fallback returned an untrusted base: FIXED.** The allocation rule is now two-stage over one shared walk (`scratchBaseFacts`): it returns the first candidate that is **both** reachable and unteplaceable; failing that it returns one that is at least unteplaceable (integrity is the property never conceded — a base another user can replace hands them the sandbox's own grant target); and if **every** candidate is replaceable it refuses rather than returning one. That also removes the "Nothing traversable is available" mislabel, which conflated the two properties. `TestSessionScratchBaseRefusesWhenEveryCandidateIsReplaceable` (both candidates `0777` non-sticky) fails before:
```
--- FAIL: TestSessionScratchBaseRefusesWhenEveryCandidateIsReplaceable (0.00s)
    session_scratch_test.go:214: allocation must refuse when every candidate base is replaceable ("/tmp/evener-replaceable-a-2412078198", "/tmp/evener-replaceable-b-2121905480")
```

**HIGH — sticky checked only on the base: FIXED.** `scratchBaseFacts` walks the **whole ancestor chain** and marks the base replaceable if *any* component is group- or other-writable without the sticky bit, since replacing an entry in a writable ancestor moves everything below it. `TestSessionScratchBaseRejectsUntrustedAncestor` (a `0711` base inside a `0777` non-sticky parent) fails before:
```
--- FAIL: TestSessionScratchBaseRejectsUntrustedAncestor (0.00s)
    session_scratch_test.go:155: a base under a world-writable non-sticky ancestor "/tmp/evener-untrusted-ancestor-1448251573" must not count as safe
```

**MEDIUM — borrow validated modes incompletely: FIXED.** `validateSessionScratchLayout` now checks the exported modes: container `0711` (traversable, no group/other read or write), private `0700`, temp `0777`+sticky. The mode half of `TestBorrowRetainedSessionScratchRefusesUnpreparedLayout` fails before:
```
--- FAIL: TestBorrowRetainedSessionScratchRefusesUnpreparedLayout (0.00s)
    scratch_retention_test.go:77: a scratch with unexported modes ("…/evener-sandbox-wrong-modes") must not be borrowed
```

**LOW — only regular files moved aside: FIXED.** `freeReservedSubtreeNames` now moves aside **any** non-directory entry at a reserved subtree name — regular file, symlink or FIFO — because the entry belongs to the session whatever its type; only a directory at such a name is reused as the subtree. The moved-aside half of `TestPrepareSessionScratchRefusesSymlinkedSubtrees` fails before:
```
--- FAIL: TestPrepareSessionScratchRefusesSymlinkedSubtrees (0.00s)
    session_scratch_test.go:343: a non-directory entry at a subtree name must be moved aside, not unlinked: info=<nil> err=lstat …/private/tmp.legacy: no such file or directory
```

**LOW — `NewWrapper` mutation undocumented / lease concern: FIXED as documentation**, which the review offered as the alternative. `NewWrapper`'s contract now states that it mutates an existing `sessionTmp` into the exported layout, that this is idempotent for a scratch Evener already laid out (every in-repo caller passes one), that it fails closed rather than unlinking, and that a caller must hold whatever exclusivity the directory needs — with the restore path noted as only ever unwrapping an allocation `OpenRetainedSessionScratch` already prepared under its lease, so it cannot migrate a still-live directory without that lease.

Acceptance on `8c9d9f870a`:
```
$ go test ./agent/sandbox/ ./agent/execenv/ -count=1
ok  	primeradiant.com/evener/agent/sandbox	2.640s
ok  	primeradiant.com/evener/agent/execenv	3.880s
$ make fuzz                                          -> exit 0
$ cd agent && go test -tags evenerfuzz -run '^Fuzz' -count=1 ./sandbox/ ./execenv/
ok  	primeradiant.com/evener/agent/sandbox	1.326s
ok  	primeradiant.com/evener/agent/execenv	0.715s
$ git show --name-only --format= HEAD | grep '\.go$' | xargs gofmt -l   -> (empty)
$ go vet ./agent/sandbox/... ./agent/execenv/...                        -> clean
$ go test ./agent/... ./envvars/... -count=1 -timeout 30m               -> all ok (agent 368.778s)
$ go test ./cmd/evener/ -run 'Scratch|Reclaim' -count=1                 -> ok
$ golangci-lint run --config .golangci.yml ./agent/sandbox/... ./agent/execenv/...  -> 0 issues.
```

obra:
Round-9 dispositions for the review of `8c9d9f8` (superseded by head `b2609b7b0c`, which carries all five).

**MEDIUM — container owner permissions not required: FIXED.** `validateSessionScratchLayout` now requires the container's **exact** exported mode `0711` and rejects unexpected special bits. A `0001` container satisfied the old "others may traverse, no group/other write" test while denying the session its own access. Failing-before with the special-bit rule removed:
```
--- FAIL: TestBorrowRetainedSessionScratchRefusesUnpreparedLayout (0.00s)
    scratch_retention_test.go:126: a container carrying an unexpected special mode bit ("…/evener-sandbox-special-bits") must not be borrowed
```
(The `0001` case is also asserted — it is refused by the subtree checks as well, since a container without owner execute cannot be traversed by anyone to reach `private`/`tmp`.)

**MEDIUM — crash sweeping touched replaceable bases: FIXED.** `SweepCrashedSessionScratch` now reclaims only from bases another user cannot replace; a replaceable base lets an attacker redirect a name between the sweep's lease check and its removal, so those bases are skipped entirely. Failing-before:
```
--- FAIL: TestSweepCrashedSessionScratchSkipsReplaceableBases (0.00s)
    session_scratch_test.go:252: the sweep must leave an allocation in a replaceable base alone: stat /tmp/evener-sweep-replaceable-2836247449/evener-sandbox-aged: no such file or directory
```

**MEDIUM — foreign-owned temp trees defeat cleanup: FIXED with a defined protocol.** Removal takes everything it can (Go's `RemoveAll` continues past failures), and on failure the container is then tightened to `0700` so no further foreign entry can be created in what remains, with the residue reported rather than hidden. The long-term isolation/drop-privilege option the review mentions is a larger design change; this bounds the listing without it. A new `sessionScratchRemoveAll` seam makes the path testable; failing-before:
```
--- FAIL: TestCleanupTightensWhenForeignResidueBlocksRemoval (0.00s)
    session_scratch_test.go:275: container mode after a failed removal = 0711, want 0700 (no further foreign entries)
```

**LOW — stale `newSandboxFS` / `WithSessionScratch` comments: FIXED.** Both now state the argument is the scratch's **private** subtree, never the container, and say why using the container again would reopen the #495 exposure.

**LOW — stale `freeReservedSubtreeNames` comment: FIXED.** It now says any non-directory entry at a reserved subtree name is moved aside under a free suffix, matching the implementation and the symlink-preservation test.

Acceptance on `b2609b7b0c`:
```
$ go test ./agent/sandbox/ ./agent/execenv/ -count=1
ok  	primeradiant.com/evener/agent/sandbox	3.338s
ok  	primeradiant.com/evener/agent/execenv	4.543s
$ make fuzz                                          -> exit 0
$ cd agent && go test -tags evenerfuzz -run '^Fuzz' -count=1 ./sandbox/ ./execenv/
ok  	primeradiant.com/evener/agent/sandbox	1.844s
ok  	primeradiant.com/evener/agent/execenv	1.139s
$ git show --name-only --format= HEAD | grep '\.go$' | xargs gofmt -l   -> (empty)
$ go vet ./agent/sandbox/... ./agent/execenv/...                        -> clean
$ go test ./agent/... ./envvars/... -count=1 -timeout 30m               -> all ok (agent 359.499s)
$ go test ./cmd/evener/ -run 'Scratch|Reclaim' -count=1                 -> ok
$ golangci-lint run --config .golangci.yml ./agent/sandbox/... ./agent/execenv/...  -> 0 issues.
```

obra:
Round-10 dispositions for the review of `b2609b7b0c` (head is now `baf8e4e92d`, which carries the two fixes below; the remaining High is reported here rather than silently absorbed).

**HIGH — Windows mode semantics: FIXED.** The mode-based judgements are now gated on a `scratchModesRecorded` seam (`true` on the POSIX platforms, `false` on Windows, where Go synthesizes writable permission bits and never sets `os.ModeSticky`). Without it, on Windows `scratchBaseFacts` marked every location replaceable: default allocation failed, the crash sweep skipped every base, and `BorrowRetainedSessionScratch` always failed, so `$EVENER_SCRATCH_DIR`/`$TMPDIR` silently stopped being exported. On a platform that records no modes the reachability and integrity tests now stand down, and `validateSessionScratchLayout` requires the layout to exist without judging its modes. `TestScratchModeChecksArePlatformGated` pins the non-recording behaviour by exercising the seam. (Because Linux records modes, production behaviour here is unchanged by design, so there is no red-then-green to show on this host; the evidence for the defect is the platform semantics the review documented, and the windows/darwin cross-vets are clean with the seam.)

**LOW — test isolation regression: FIXED.** The tests that pin allocation by redirecting `TMPDIR` at a `t.TempDir()` base now use a new `pinnedScratchBase` helper that creates a base the allocator will actually choose (`t.TempDir()` is `0700` under a `0700` root, so with the reachability preference allocation escaped to the real user cache and the leak assertions passed vacuously while leaving retained scratch beside the developer's cache). 15 call sites updated across `session_worktree_swap_scratch_unix_test.go`, `sandbox_delegate_create_test.go`, `sandbox_delegate_degrade_test.go` and `sandbox_delegate_test.go`; the helper skips with a clear reason if a host's temp chain is not traversable, rather than pretending.

**HIGH — wrapperless commands can write into the `0711` container: REPORTED, not fixed, because it is a design decision I should not take unilaterally.** The finding is accurate for the *unsandboxed* path: a wrapperless command runs with no kernel policy at all, only modes, so it can create `<container>/file` at `0644` and another local user who can traverse the container can read it. The two directions are:

1. *Make the container non-writable after setup and relocate the metadata* — container `0511` instead of `0711`, with the liveness lease and the retention pin moved from the container root into `private/`. This changes the on-disk retention contract (pin location, with a legacy-read fallback for pre-existing allocations), requires chmod-before-remove in `Cleanup` and the crash sweep (a `0511` container cannot unlink its own children), and updates every test that pins the `0711` container mode or the pin's location.
2. *Document it as outside the unsandboxed threat model* — an unsandboxed session runs with the user's full privileges, so a command can write anywhere the user can (including `/tmp/x` or `~/public/`); the container's traversal bit only means a file deliberately placed at the container root is readable by name.

I lean to (1) because it makes the container's guarantee absolute rather than dependent on the session's good behaviour, but it is a metadata-layout change wider than this issue's scope, so I am asking rather than assuming. Say the word and it is one further push.

Acceptance on `baf8e4e92d`:
```
$ go test ./agent/sandbox/ ./agent/execenv/ -count=1
ok  	primeradiant.com/evener/agent/sandbox	2.613s
ok  	primeradiant.com/evener/agent/execenv	3.677s
$ make fuzz                                          -> exit 0
$ cd agent && go test -tags evenerfuzz -run '^Fuzz' -count=1 ./sandbox/ ./execenv/
ok  	primeradiant.com/evener/agent/sandbox	1.288s
ok  	primeradiant.com/evener/agent/execenv	0.662s
$ git show --name-only --format= HEAD | grep '\.go$' | xargs gofmt -l   -> (empty)
$ go vet ./agent/sandbox/... ./agent/execenv/...                        -> clean
$ GOOS=windows|darwin go vet -tags evenerfuzz ./sandbox/... ./execenv/...  -> clean
$ go test ./agent/... ./envvars/... -count=1 -timeout 30m               -> all ok (agent 287.343s)
$ go test ./cmd/evener/ -run 'Scratch|Reclaim' -count=1                 -> ok
$ golangci-lint run --config .golangci.yml ./agent/sandbox/... ./agent/execenv/...  -> 0 issues.
```

obra:
Round-12: **the remaining High is fixed** — the live scratch container is now writable by nobody.

**HIGH — wrapperless commands could write into the traversable container: FIXED.** `sessionScratchDirMode` is now `0o511` while the session is LIVE: everyone can traverse the container to reach the exported subtrees, and **nobody** can create an entry in it — not another user (no group/other write) and not a wrapperless command running as the session's own user (no owner write). The last route by which a `0644` artifact could be dropped beside the subtrees, where another local user could read it by name, is gone.

Supporting design, so the fix does not trade one problem for another:
- **Retained state stays `0711`** (`sessionScratchRetainedDirMode`): a container that withholds owner write cannot be removed without an intervening chmod, and retained cleanup is manual, so `Retain` relaxes it — exactly the state a retained scratch had before this change, and the state a borrowing consumer expects. Restoring settles it again.
- **Evener's own metadata writes** — layout/migration, lease acquire, pin publish and clear — take a brief owner-only `0700` window that restores the mode it found. That is why no metadata had to move, which avoids the dual-lease and pin-divergence hazards I reported when first assessing this remedy.
- **Layout detection is subtree presence**, not container mode (the mode now differs between legacy `0700`, live `0511` and retained `0711`).
- **The wrapper settles only directories in Evener's own scratch namespace**; a caller-supplied directory is provisioned and migrated but keeps its own mode, so Evener never re-modes a directory outside its namespace.

Failing-before evidence (the previous `0711` mode):
```
--- FAIL: TestSessionScratchLifecycle (0.00s)
    session_scratch_test.go:339: scratch mode = 0711, want 0511 (traversable so the shared temp child is reachable, and writable by nobody)
--- FAIL: TestNewWrapperProvisionsExportedSubtrees (0.00s)
    backend_test.go:41: wrapper container mode = 0711, must not be writable, even by its owner
```
`TestSessionScratchLifecycle` additionally asserts a write into the settled container is **refused** while the private subtree is writable.

Acceptance on `8b7bac7718`:
```
$ go test ./agent/sandbox/ ./agent/execenv/ -count=1
ok  	primeradiant.com/evener/agent/sandbox	2.184s
ok  	primeradiant.com/evener/agent/execenv	3.747s
$ make fuzz                                          -> exit 0
$ cd agent && go test -tags evenerfuzz -run '^Fuzz' -count=1 ./sandbox/ ./execenv/
ok  	primeradiant.com/evener/agent/sandbox	1.206s
ok  	primeradiant.com/evener/agent/execenv	0.650s
$ git show --name-only --format= HEAD | grep '\.go$' | xargs gofmt -l   -> (empty)
$ go vet ./agent/sandbox/... ./agent/execenv/...                        -> clean
$ go test ./agent/... ./envvars/... -count=1 -timeout 30m               -> all ok (agent 327.143s)
$ go test ./cmd/evener/ -run 'Scratch|Reclaim' -count=1                 -> ok
$ golangci-lint run --config .golangci.yml ./agent/sandbox/... ./agent/execenv/...  -> 0 issues.
```
Fixtures that manipulate or sweep a live container take the same window (or relax it at cleanup); that is the full extent of the blast radius, and it is contained to tests.

obra:
Round-13 review response (head below).

**High — refuted (misreading of the guard).** `withWritableScratchContainer` takes
its direct path only when `validateSessionScratchLayout(dir) != nil`, i.e. only
when the layout is NOT valid. A container `prepareSessionScratch` just settled is
live 0511 with both subtrees present, so it validates and acquisition runs inside
the owner-only 0700 window, with the mode it found (0511) restored afterwards:

```go
func withWritableScratchContainer(dir string, fn func() error) error {
	if validateSessionScratchLayout(dir) != nil {
		return fn() // legacy/foreign layout: no exported mode to restore
	}
	... original := info.Mode().Perm()
	container.Chmod(sessionScratchSetupMode) // 0700
	fnErr := fn()
	return errors.Join(fnErr, container.Chmod(original))
}
```

Evidence: `TestSessionScratchLifecycle` now also asserts the lease file exists
*inside* the settled container after minting (a direct probe of the disputed
write), and it passes on the 0511 container; retention pin publish/clear run
through the same window and pass in the retention tests.

**Medium — pre-existing `tmp` mistaken for the exported subtree.** Fixed:
`layOutSessionScratch` now also treats the scratch as legacy when an existing
`tmp` is not already `sessionScratchTmpMode` (1777+sticky), so its contents are
hardened before the world-writable mode publishes them.

**Medium — `NewWrapper` migrated a caller-supplied directory.** Fixed:
`NewWrapper` no longer migrates; a new `ensureScratchSubtrees` only creates the
two exported subtrees (refusing a non-directory at either name) and performs no
relocation, preserving the caller's mode and files. Evener's own allocation
(basename prefixed `evener-sandbox-`) still goes through `prepareSessionScratch`.

**Low — contended-lease cleanup used `os.RemoveAll`.** Fixed: the branch now calls
`removeSessionScratchTree`, which opens the writable window a live container needs.

**Low — symlink-escape fixture planted its link in a container nobody may write.**
Fixed: the link is now planted at
`filepath.Join(sandbox.SessionScratchPrivateDir(scratch), "escape")`.

**Low — `hardenRelocatedEntry` could abort a migration on an unbreakable hard
link.** Fixed: it now skips such an entry (the private subtree's own 0700 withholds
it from other users) instead of failing the mint/restore.

**Low (docs) — `docs/sandboxing.md` did not say the temp subtree is shared.**
Fixed: it now states the shared temp is visible to any local user, so confidential
temp must be created 0600 or inside a 0700 subdirectory.

Verification on this head: `go test ./agent/... ./envvars/... -count=1` ok;
`make fuzz` exit 0; tagged replay (`-tags evenerfuzz -run '^Fuzz' ./sandbox/
./execenv/`) ok; `go vet` and `gofmt -l` clean; scoped `golangci-lint` 0 issues.

obra:
Round-14 review response.

**High — refuted (same misreading as the previous round).** `withWritableScratchContainer`
takes its direct path only when `validateSessionScratchLayout(dir)` returns an *error*,
i.e. when the container is not in the exported layout:

```go
func withWritableScratchContainer(dir string, fn func() error) error {
	if validateSessionScratchLayout(dir) != nil {   // NOT exported → no window to keep
		return fn()
	}
	... container.Chmod(sessionScratchSetupMode)     // 0700 window
	fnErr := fn()
	return errors.Join(fnErr, container.Chmod(original))
}
```

A container `prepareSessionScratch` has just settled is live `0511` with both subtrees
present, so it validates and lease acquisition / pin writes **do** run inside the
`0700` window. The suggested fix is the code as written. Evidence: `TestSessionScratchLifecycle`
asserts the lease file exists *inside* the settled container after minting (a direct
probe of the disputed `O_CREAT`), and every test that mints a scratch — a large part of
`./agent/...` — passes on this head.

**Medium — fixed.** The window no longer restores a mode it *sampled*. It now reads the
container's **exported** mode and restores that, refusing the window when the container
holds anything else:

```go
original, exported := exportedScratchMode(info.Mode())
if !exported {
	return fmt.Errorf("sandbox: session scratch %q is in mode %04o, not an exported mode", dir, info.Mode().Perm())
}
```

Two overlapping windows on one container therefore always leave it exported, whichever
restore lands last. Under the old code the loser restored `0700` — the setup mode it
observed inside the winner's window — over the settled `0511`, which left a live scratch
whose `$TMPDIR` a privilege-dropping child could not reach and which its owner could then
write: the #495 exposure, persisting for the rest of the session.

I deliberately did not take either suggested shape, and here is why. Passing a declared
target (`sessionScratchDirMode` for leased-live, `sessionScratchRetainedDirMode` for
retained) does not work for two callers that genuinely cannot know the lifecycle state
before the window: `ReleaseScratchRetention` walks manifest references that may still be
live or already retained, and the crash sweep walks containers that may be live,
retained, or legacy — declaring "retained" there would relax a live container to
owner-writable, i.e. re-open #495, and declaring "live" would strip a retained
container of the owner write its manual cleanup needs. Serializing the window on the
container would also work, but adds a platform primitive and a nested-window deadlock
hazard for no gain over constraining the restore value. Restricting the restore to the
two exported modes keeps every caller's lifecycle decision unchanged and makes the
reported harm impossible: a setup mode can only ever be refused, never restored. A
legacy/foreign container still never takes a window at all.

Tests added, both in `agent/sandbox/session_scratch_test.go`:

- `TestWritableScratchWindowRefusesAConcurrentSetupMode` — drives the exact
  interleaving the finding describes (a second window takes the container to its setup
  mode between the layout check and the mode read) and asserts the window refuses
  rather than restoring that mode, and does not run its callback. **It fails against the
  previous sampling restore**: with the old body it reports `fn must not run while the
  container is at another window's setup mode` and `a setup mode observed inside the
  window must be refused, not restored`, and leaves the container at `0700`.
- `TestWritableScratchWindowRestoresTheExportedModeItCarried` — a retained container
  comes back `0711`, a live one `0511`.

Verification on this head: `go test ./agent/... ./envvars/... -count=1` ok; `make fuzz`
exit 0; tagged replay (`-tags evenerfuzz -run '^Fuzz' ./sandbox/ ./execenv/`) ok;
`go vet` and `gofmt -l` clean; scoped `golangci-lint` 0 issues.

obra:
Round-15 review response: all three Mediums and the Low are addressed in the head below.

**Medium — `withWritableScratchContainer` double-pass race: fixed by serializing the window.**
The window now takes an exclusive advisory lock on the container itself (`flock` on the
container's own handle; two handles on one directory contend even inside one process, and
the lock releases when the handle closes) and holds it across the layout check, the mode
read, the chmod and `fn`:

```go
container, err := openScratchDirNoFollow(dir)   // exported layout is always openable
...
if err := lockScratchDirFile(container, false); err != nil { return err }
defer func() { _ = unlockScratchDirFile(container) }()
if validateSessionScratchLayout(dir) != nil {   // judged only under the lock
	return fn()
}
info, err := container.Stat()
original, exported := exportedScratchMode(info.Mode())
...
```

Your interleaving (`A.chmod(0700); B.chmod(0700); A.fn ok; A.chmod(0511); B.fn`) is now
impossible, and so is the one I found while writing the test for it: deciding the layout
*before* the lock meant a window that started while another was inside the setup mode took
the direct path, and its callback's write could still land after the first window restored
the live mode. Both are the same `EACCES` you describe, so the layout check moved under
the lock. `OpenRetainedSessionScratch` can no longer report a wrapped `EACCES` where
`ErrScratchRetentionLeaseHeld` is meant, and `Pin`/`rollbackUnpublishedScratchPin` cannot
fail to publish or clear a pin for this reason.

Tests: `TestWritableScratchWindowHoldsTheContainerLock` (a second writer must not be able
to take the container while a window runs) and
`TestWritableScratchWindowsRunTheirCallbacksWritably` (four concurrent windows, every
callback writes inside the container, container left `0511`). Both fail against the
unserialized window — the second with exactly `permission denied` on the callback's write.

**Medium — `prepareSessionScratch` transitions valid live containers and lacks failure-path
restore: fixed.**
A container already settled at the live mode is now a no-op (with the layout re-checked,
so a container that is live but *incomplete* is still repaired), and every failure path
restores the exported mode the container carried:

```go
mode, exported := exportedScratchMode(info.Mode())
if exported && mode == sessionScratchDirMode && validateSessionScratchLayout(dir) == nil {
	return nil
}
var restore os.FileMode = sessionScratchDirMode
if exported { restore = mode }
...
if err := layOutSessionScratch(dir); err != nil {
	return errors.Join(err, scratchContainerChmod(container, restore))
}
```

Tests: `TestPrepareSessionScratchLeavesASettledLiveContainerAlone` (asserts zero container
chmods for a settled scratch — it observes 2 against the previous behaviour) and
`TestPrepareSessionScratchRestoresAModeWhenTheLayoutFails` (was left at `0700`).

**Medium — `BorrowRetainedSessionScratch` leaves the retained container at `0711`: fixed.**
A borrow now registers a shared claim and settles the container live for as long as the
consumer is active; the retained, owner-writable mode comes back only when the last
consumer finishes.

```go
claim, err := acquireScratchBorrowClaim(canonical)          // shared flock
...
if err := setScratchContainerMode(canonical, sessionScratchDirMode); err != nil { ... }
return &SessionScratch{Dir: canonical, base: ..., borrow: claim}, nil
```

The claim is a shared lock on the container's **private subtree**: every exported layout is
required to have that directory and nothing migrates or replaces it, so this needs no new
on-disk artifact, and it cannot collide with the setup window's exclusive lock on the
container handle itself. The last consumer is decided by an exclusive probe, so the
question is answered atomically across processes; a consumer that leaks its handle leaves
the container write-withheld (tightened) rather than exposed. `SessionScratch.Retain`
releases a borrow claim and settles `0711`; the settle is skipped whenever another claim
is still held, which is also what keeps an adopter's teardown from relaxing the container
underneath a live borrower.

Test: `TestBorrowedScratchWithholdsContainerWriteUntilTheLastConsumerFinishes` — `0511`
while borrowed, still `0511` after the first of two consumers finishes, `0711` after the
last. It observes `0711` for the borrow against the previous behaviour.
`TestRetirementSharedConsumerBorrowUnit`'s fixture now takes Evener's disposal window
before removing a container its borrowed consumer kept write-withheld.

**Low — detached doc comment:** the helper that had been inserted between
`TestUnsandboxedScratchDirGitProbeDoesNotSelfDeadlock` and its doc comment now sits above
the comment, so the deadlock rationale is attached to its test again.

Verification on this head: `go test ./agent/... ./envvars/... -count=1` ok; `make fuzz`
exit 0; tagged replay (`-tags evenerfuzz -run '^Fuzz' ./sandbox/ ./execenv/`) ok;
`go vet` (including `GOOS=windows`) and `gofmt -l` clean; scoped `golangci-lint` 0 issues.

obra:
Follow-up on the same round: `lint-golangci` failed on `adf2eb9` with one `modernize` finding in the new test (`agent/sandbox/session_scratch_test.go:507` — `for i := 0; i < workers; i++`), which the `static` gate then mirrored. Fixed by ranging over the constant (`for i := range workers`) and pushed as `2bf573a4`; no behavioural change.

obra:
Round-16 review response: the High and the Low are fixed; the Medium is refuted with evidence.

**High — retained-mode settling ignored the live lease: fixed.** You are right, and the
trigger is exactly as you describe. The borrower claim is a shared lock on the private
subtree, but a lease-owning session's lock is on `.evener-session.lock`, so a borrower that
finished first made the exclusive subtree probe succeed and downgraded a live adopter's
container to `0711`. `settleRetainedScratchMode` now takes the second decision too:

```go
last, err := tryLockScratchDirFile(probe)     // any borrowing consumer left?
if err != nil || !last { return }
defer func() { _ = unlockScratchDirFile(probe) }()
if scratchLeaseHeld(dir) { return }           // any live lease left?
_ = setScratchContainerMode(dir, sessionScratchRetainedDirMode)
```

`scratchLeaseHeld` opens the lease file and takes the lease non-blockingly (releasing it
again immediately); contention — and any other failure to take it — reads as "still
owned", which leaves the container live rather than relaxing it. A directory with no lease
file never had a lease, so it still settles. I did not take the alternative shape of
skipping the settle only in the borrow branch, because the same downgrade is reachable from
an adopter's own teardown when borrowers remain; deciding both questions in one place keeps
the invariant "`0711` means no live consumer of any kind" in a single function.

Test: `TestBorrowReleaseDoesNotDowngradeALiveSession` — a live lease-owning session borrows,
the borrower finishes, the container must still be `0511`. It reports `0711` against the
previous behaviour.

**Medium — the metadata-write test is incompatible with the live mode: refuted.** That
write is not creating an entry; it is overwriting one the fixture already published.
`TestRetirementRejectsForeignRetainedScratchPin` calls `scratch.Pin(...)` first, which
creates `.evener-retained-session.json` (inside the setup window), and the later
`os.WriteFile` replaces its contents. Creating an entry in a live `0511` container needs
write permission on the *container*, which the live mode withholds; overwriting an existing
file needs write permission on the *file*, which its owner has. I verified this rather than
reasoning about it: I instrumented the test to report the container mode at that write
(`0511`) and ran it as uid 1000 on this non-root host — it passes as written, so no change
is needed. If you still see an `EACCES` there, it is on a host with a different fixture
state, and I would want the exact mode and uid.

**Low — error path leaked a fresh scratch: fixed, and it was mine.** The round-15 change
made a failed layout restore an exported mode, so the live `0511` container could not be
unlinked by the plain `os.RemoveAll` and the error was discarded. The path now matches its
siblings:

```go
if err := prepareSessionScratch(dir); err != nil {
	_ = removeSessionScratchTree(dir)   // opens the container up, then removes
	return nil, fmt.Errorf("sandbox: secure session scratch: %w", err)
}
```

Test: `TestFailedMintLeavesNoScratchBehind` — a mint whose layout cannot be read leaves no
`evener-sandbox-*` directory behind (it reports the leaked directory against the previous
behaviour).

Also folded in from CI on the previous head: one `modernize`/`rangeint` finding in the new
concurrency test (`for i := 0; i < workers; i++` → `for i := range workers`), which was the
sole cause of the `lint-golangci`/`static` failure.

Verification on this head: `go test ./agent/... ./envvars/... -count=1` ok; `make fuzz`
exit 0; tagged replay ok; `go vet` (incl. `GOOS=windows`) and `gofmt -l` clean; scoped
`golangci-lint` 0 issues.

obra:
Round-17 review response: both Highs and both Mediums are fixed.

**High — sticky directory ownership was not verified: fixed.** `scratchBaseFacts` treated a
group/other-writable component as safe on its mode alone. The sticky bit only stops another
user from removing *our* entry: the **directory's own owner** may delete or rename anything
inside it, sticky or not. A `1777` base owned by another local user is therefore exactly as
replaceable as a plain world-writable one, and replacing it moves the sandbox bind and grant
targets with it.

```go
if scratchModesRecorded && mode.Perm()&0o022 != 0 {
	if mode&os.ModeSticky == 0 || !scratchComponentOwnerTrusted(info) {
		unteplaceable = false
	}
}
```

`scratchComponentOwnerTrusted` accepts this process's uid or root's (the trusted system owner
of a shared temp directory such as `/tmp`), and a platform that reports no owner makes no
mode-based judgement at all. Test: `TestScratchBaseRejectsAnotherUsersStickyComponent` — a
sticky base this process owns stays usable, and the same base is rejected once the owner
lookup reports a foreign uid. The owner lookup is a seam (`fileOwnerID`) precisely because
producing a real foreign owner needs another user's account; against the previous behaviour
the test reports `unteplaceable=true` for the foreign-owned base.

**High — the lease check raced the container lock: fixed.** `settleRetainedScratchMode` no
longer probes the lease before the mode change; the decision and the change now happen under
the container's exclusive lock, which is the same lock every lease is acquired under (all
lease acquisitions run inside the setup window):

```go
if err := lockScratchDirFile(container, false); err != nil { return err }
defer func() { _ = unlockScratchDirFile(container) }()
if scratchLeaseHeld(dir) { return nil }          // decided under the lock
return scratchContainerChmod(container, sessionScratchRetainedDirMode)
```

So a restorer can no longer acquire the lease between the probe and the chmod and have the
container relaxed to owner-writable underneath a live session. Test:
`TestSettleRetainedModeProbesTheLeaseUnderTheContainerLock` — it replaces the probe with one
that asks the kernel whether the container lock is already held, so it pins the ordering
rather than the timing; against the previous behaviour it reports "the lease probe ran
outside the container lock".

**Medium — leaked scratch lease on a failed setup-window restore: fixed.** You are right that
all three callers (plus the mint) would drop a handle the callback had already opened, and
the flock then survives until the garbage collector finalizes the descriptor. The fix is one
helper shared by every lease-acquiring window caller:

```go
func acquireScratchLeaseInWindow(dir string) (scratchLease, bool, error) {
	... withWritableScratchContainer(dir, func() error { lease, contended, acquireErr = acquireScratchLease(...) })
	if err == nil || lease == nil {
		return lease, contended, err
	}
	releaseErr := lease.Release()
	return nil, contended, errors.Join(err, releaseErr)
}
```

`NewSessionScratch`, `OpenRetainedSessionScratch`, `ReleaseScratchRetention` and
`sweepCrashedSessionScratch` all acquire their lease through it now, so a restore-chmod
failure after a successful callback can no longer leave a lease held. Test:
`TestWindowRestoreFailureDoesNotLeakTheLease` — the window's restore chmod is forced to fail;
the helper returns no handle and a subsequent lease acquisition succeeds. Against the
previous behaviour the test reports `contended=true / resource temporarily unavailable`,
i.e. the lease was still held.

**Medium — failed legacy migration republished `0511` partway: fixed.** A failed layout is
restored to the exported mode the container carried; a **legacy** container has none — it is
`0700` with the session's files still at its root, some of them world-readable — so publishing
`0511` mid-migration would expose everything the relocation had not reached yet. It now stays
at `0700`, and the caller still removes it through `removeSessionScratchTree`, which opens the
container up first, so the round-16 leak fix holds:

```go
var restore os.FileMode = sessionScratchSetupMode   // legacy: never publish 0511 partway
if exported { restore = mode }                      // an exported container keeps its mode
```

Test: `TestPrepareSessionScratchRestoresAModeWhenTheLayoutFails` now covers both cases — a
legacy container stays at `0700` (it reported `0511` against the previous behaviour) and a
retained container is restored to `0711`.

Verification on this head: `go test ./agent/... ./envvars/... -count=1` ok; `make fuzz`
exit 0; tagged replay ok; `go vet` (incl. `GOOS=windows`) and `gofmt -l` clean; scoped
`golangci-lint` 0 issues.

obra:
Round-18 review response: all three Mediums and the Low are fixed in this head.

**Medium — borrow path exposed direct container artifacts: fixed.**
`BorrowRetainedSessionScratch` now moves every non-reserved root entry into the private
subtree (which also hardens it) inside the setup window, before the container is tightened to
the live mode:

```go
claim, err := acquireScratchBorrowClaim(canonical)
...
if err := withWritableScratchContainer(canonical, func() error {
	return relocateScratchRootEntries(canonical)
}); err != nil { ... }
if err := setScratchContainerMode(canonical, sessionScratchDirMode); err != nil { ... }
```

A retained container is owner-writable, so a `0644` artifact can legitimately sit at its root;
the live mode makes the container traversable, so it would then be readable by any local user
who knows its name. Test: `TestBorrowRelocatesDirectContainerArtifacts` — against the previous
behaviour it reports `a borrowed container still publishes ".../left-at-root.bin" at its root,
where every local user can read it`, and the artifact must survive hardened (`0600`) inside
`private/`.

**Medium — the cross-session denial test no longer proved isolation: fixed.**
The `other-session` case named the other session's container **root**, a path outside every
grant, so it was denied for every session and an implementation that wrongly granted all
private subtrees would still have passed. It now names the path shape the grant actually
covers:

```go
{name: "other-session", target: filepath.Join(sandbox.SessionScratchPrivateDir(otherScratch), "issue-368-other.txt")},
```

**Medium — the wrapper pre-check blocked the migration it precedes: fixed by deleting it.**
`prepareSessionScratchForWrapper` ran the non-directory check on the subtree names *before* the
Evener-allocation branch, so on restore/resume a retained legacy scratch whose root still holds
a file named `tmp` or `private` aborted its own restore — the wrapper is rebuilt before
`OpenRetainedSessionScratch` re-settles the layout, and `layOutSessionScratch` exists precisely
to move such a squatter aside and preserve it inside `private`. A caller-supplied directory
keeps its protection: `ensureScratchSubtrees` refuses a non-directory rather than replacing it.
Test: `TestWrapperMigratesALegacySquatterAtASubtreeName` — against the previous behaviour it
reports `session scratch subtree ".../tmp" must be a directory`.

**Low — the migration doc comment named the wrong settled mode: fixed.** It said the container
is loosened to `0711`; the code applies the live mode `0511` (`0711` is the retained mode
`Retain` applies later). `docs/sandboxing.md` already documented `0511`.

Verification on this head: `go test ./agent/... ./envvars/... -count=1` ok; `make fuzz`
exit 0; tagged replay ok; `go vet` (incl. `GOOS=windows`) and `gofmt -l` clean; scoped
`golangci-lint` 0 issues.

obra:
Closing this unmerged, on a human ruling.

Why: the direction it implements cannot converge. The design points one variable (`TMPDIR`) at one directory and asks that directory to be simultaneously private to the session owner, writable by an arbitrary other user, session-scoped, and cleanable by the owner. Those cannot all hold for one path — the round-5 finding (a `1777` temp subtree lets a privilege-dropping child plant files the session owner cannot remove) is that contradiction showing up in practice, and every subsequent round of findings follows from it.

The work here is not wasted: the investigation established the reachability, integrity, retention, migration, sweep and Windows invariants that any replacement must respect, and that record stays readable in the review history and the dispositions comments.

Issue #495 remains open and is being re-specified from user requirements rather than from this design. Do not reopen this PR; the replacement will be derived from that spec.

PR #2048, opened 2026-09-20
fix(web): put subagent inline status, turns, calls, and clock on one line

obra:
## What

The inline subagent card's first line now carries the delegate's lifecycle status word together with its turn/call counts and run clock, per Jesse's review of the old layout:

- **Before:** intent line, then "Running", then the child's latest activity at prose size, then "● 1 turn · 12 calls … 4m38s".
- **After:** intent line, then **one** first line — `Running · 1 turn · 12 calls … 4m38s` — with the child's latest tool use quoted below it at UI size (13px, down from 18px prose).

The latest-tool-use quote already preferred the stated intent over the action; this pins it with a test. The standalone lifecycle line still renders for collapsed rows and legacy activation-only receipts, and steps aside while the expanded card owns the line, so the status word never renders twice.

## How

- `SubagentCard` merges the status word into its stats line (`subagent-stats`, now carrying `data-kind`/`data-attention`); the quote moves below it.
- Shared with the collapsed row's standalone line: `DelegateStatusWord` (word + attention marker), `delegateLifecycleLabel` (one status-word vocabulary), and `delegateOutputHasCard` (the card-renders gate `rowFromDelegateItem` itself applies). The standalone line is constructed only when it will render.
- CSS state hues are one grouped rule per state across both surfaces.
- Dev consumers retargeted: `transcript-guard.ts` and the three `editorial-preview` browser scripts now read the word from `subagent-status-word` / a stats+lifecycle union selector.

## Verification

- `make test-web`: PASS (typecheck, full unit, Biome) on both commits.
- Affected suites: 2044 vitest tests (transcript, dev surface sections, styles).
- `make test-web-browser`: layoutguard PASS (drives the retargeted transcript-guard through the attention-card collapse flow at real Chrome geometry); shellguard, spawnguard, transcriptscrollguard, retirementguard, skillguard PASS.
- `editorial-preview-browser.mjs` against real Chrome: the retargeted status-word assertions pass ("Idle · reported", "Running", "Status unavailable"); the lane's later credentials-page wait fails **identically on the base commit** (pre-existing, unrelated).
- Two passes of the simplify-code review (4 reviewers each); round-two findings applied in the follow-up commit.

## Notes

- Pre-existing, unrelated, reproduced on base `383d73f1`: the editorial-preview lane's "+ Add provider instance" wait, and web-overflowguard's font-declaration assertion (all its geometry checks pass).
- Deferred altitude debt (journal): giving `DelegateStatusWord` its own DOM identity (testid + state attrs, collapsing the hand-copied union selectors), and a descriptor `statusLine` hook so subagentModule owns the standalone line end-to-end.


obra:
RoboRev review disposition (all findings validated against the code before acting):

**Low — fixed in `d9f684ff`:**
- Empty `delegate_id`: real regression — `str()` returns `""`, and the unified predicate gated on presence (`!== undefined`) instead of truthiness, so an empty id wrongly rendered a card. `delegateOutputHasCard` now gates on truthiness and `rowFromDelegateItem`/the standalone status line treat an empty id as activation-only. Pinned with tests.
- Missing separator: the attention marker now renders with a leading space inside its span (the collapsed line's flex gap is unchanged; whitespace-only flex items still drop).
- Doubled diamond: the expanded card renders the glyph-less marker (`withGlyph={false}`) since the line already leads with `STATUS_GLYPH`'s attention diamond; the collapsed standalone line keeps `◆ Needs attention`. Pinned with tests (exactly one ◆ per line, either surface).

**Medium — declined, with evidence:** `.stats` keeps `flex-wrap: wrap` deliberately: the pre-merge `.lifecycle` line had the identical policy (subagentmodule.module.css line 5), wrapping beats horizontal overflow on narrow cards (the overflowguard and layoutguard both police overflow, and layoutguard's 360px case passes), and the merged line fits a normal content column without wrapping. If you have a concrete viewport where it wraps mid-desktop, that's a different finding and we'll take it.

obra:
The re-review's one finding re-raises the `flex-wrap: wrap` point — same disposition as above (https://github.com/prime-radiant-inc/evener/pull/2048#issuecomment-5747186822): the wrap policy predates this change, matches the collapsed lifecycle line, and wrapping beats horizontal overflow on narrow panes, which the overflow/layout guards police. Happy to revisit with a concrete viewport where the row wraps on a desktop-width pane.

Issue #1984, opened 2026-09-19
appwire: verify P10a transcript wire-contract comment after extraction


obra:
## Finding

The P10a transcript wire-contract extraction at `68244fb400d17c7d4fd72696d4a3117700227afa` leaves a comment at `appwire-client/typescript/transcriptDisplayConfig.ts:384` referring to `fromWirePatchResponse`. That symbol is absent from the six-file extraction, so the comment's claimed shared validation path is stale at this slice.

This is a documentation-accuracy follow-up only; it is not a runtime defect and does not include the required frontend typecheck.

## Follow-up

After the later P10 read/write slices settle, verify whether `fromWirePatchResponse` is restored in the relevant source. If it remains absent, revise the comment to describe the actual validation path without naming a missing helper. If it is restored, update the wording to point to the final helper. Keep this follow-up outside the P10a Lows-only patch.

Related: #1845. P10a currently has no PR.

PR #2051, opened 2026-09-20
fix(sdk): extract transcript display wire contracts

obra:
Extract the shared wire rejection-payload decoder and use it from the keybindings store. Transcript-display default decoders now ignore unknown wrapper fields while still validating known fields, matching the PATCH wrapper contract.

This small wire-contract slice precedes the shared transcript read and PATCH stores. Validation: 164 focused tests, package build and external-import qualification, frontend typecheck, Biome, and package-import checks passed. Local RoboRev2679 found no issues. The restack preserves every owned added/removed line from the previously reviewed slice.

Low follow-up #1984 remains separate.


obra:
Exact-head raw review at dea5d5a3eb5c2851512894d24d2116511e89d8b9 is complete. Luna found no issues.

The non-object data allegation is not reachable through the WireError constructor: extractEvenerErrorInfo accepts only non-null object data, and the constructor assigns the readonly discriminator only from that extraction. Missing/non-object data therefore cannot match the required discriminator and is rejected before indexing. No production guard is needed for an invented independently mutated error shape.

The comment referring to fromWirePatchResponse is the known documentation Low #1984. A focused comment-only follow-up is assigned; no runtime PATCH behavior change is part of that fix. The qualified parent remains unchanged under the Low-only merge policy.

PR #2005, opened 2026-09-19
agent: persist interrupt boundaries before resolving pending asks

obra:
Interrupting a session with a pending question could clear the live ask before its interrupt marker was durably recorded. A failed transcript write then left the live session resolved while restore still saw the question pending. This follow-up to #1907 admits the marker through the existing durable owner write path before clearing the ask or draining queued input. Rejected markers preserve the ask and settle Awaiting; retained/adopted markers resolve it exactly once. Both inner cancellation paths leave settlement to this outer boundary, including already-canceled input.

Validation: 26 explicitly enumerated interrupt/deadline tests pass in normal, race, and evenerfuzz modes; normal/tagged and Windows normal/tagged vet plus scoped agent lint pass. The coordinator independently reran both new marker regressions. Tests exercise real transcript rollback and recovery-barrier failures, enabled queue draining, pre-canceled input, and restored state. Independent correctness/simplification review and local RoboRev job 2633 found no issues.

This addresses the interrupt-persistence portion of #1181; the broader writer audit remains open. Current-head CI and the configured remote review panel still need qualification before merge.

## Successor for remaining rejected-marker findings

The bounded successor [#2057](https://github.com/prime-radiant-inc/evener/pull/2057) is ready, stacked on `codex/interrupt-marker-persistence` at the published PR #2005 head. It restores rejected-marker state and `askPending` from the durable transcript, publishes the live state under the existing `attentionMu -> s.mu` ordering before releasing the transcript door, and maps fork provenance through compaction and orphan-repair coordinates. It preserves the scope of #2005; no sixth broad patch was added here.

Validation for #2057: focused interrupt/durable-admission tests, the three successor regressions in normal and race modes, `go vet ./agent`, scoped agent lint, and exact-base RoboRev job 2698 pass. The prior full-package retirement failures were reproduced on the exact parent and traced to macOS `/var` versus `/private/var` fixture spelling and shared stale scratch state; they are recorded in the successor receipt and are not reported as a full-suite pass.

PR #2057, opened 2026-09-20
server: restore rejected interrupt boundaries durably

obra:
## Problem

When an interrupt marker is rejected after `recordTurn` has appended a live turn but its ordinary transcript write rolls back, the live session can retain a phantom tool result and stale `askPending` state. State publication also had a gap between transcript restoration and the live boundary assignment, and fork provenance used the full-transcript divergence against compacted history.

## Change

- Rebuild rejected-interrupt state and pending questions from the durable transcript boundary, preserving retained records and clean rollback semantics.
- Publish `state` and `askPending` under the existing `attentionMu -> s.mu` lock order, then emit boundary events after both locks release.
- Map fork divergence through the same compaction retention and orphan-repair coordinates used by cold restore.
- Add deterministic regressions for durable rollback, lock-scope publication, and compacted-fork provenance.

## Validation

- Focused interrupt, durable-admission, and deadline-audit tests: pass.
- `go test -race ./agent -run '^(TestAskUser_RejectedInterruptUsesDurableBoundary|TestRestoredFailureBoundaryPublishesBeforeDoorRelease|TestRestoredFailureBoundaryMapsCompactedForkDivergence)$' -count=1`: pass with canonical TMPDIR.
- `go vet ./agent`: pass.
- `golangci-lint run ./agent/...`: 0 issues.
- RoboRev job 2698 against exact published base `755f6a1ecd36657a0eb36a87b6e548e8c2beaa8b`: no issues found.

The prior full package run's three retirement failures were reproduced on the exact parent and traced to macOS `/var` versus `/private/var` fixture spelling and shared stale scratch state; they are recorded in the successor receipt and are not reported as a full-suite pass.


obra:
**Disposition of review findings (head 2f1e99ac58)**

**Medium — canceled ask pair rides the fold rewrite tail and resurrects as pending after restore: CONFIRMED and fixed in this PR.**

Built the reviewer's mechanism as a deterministic regression test (`TestFoldTail_CanceledAskPairRidingTailDoesNotResurrectAfterRestore`): a real fold is parked inside its scripted summarizer past its history/pair-log snapshot; during the blocked interval a canceled ask round is recorded (user + assistant + appendCanceledToolResults) with a one-shot transcript write failure. When the fold publishes, the failed pair — whose posted ack is non-error — rode `publishFoldTransaction`'s rewrite tail past the compaction markers, and a cold restore resurrected it as a pending ask (pre-fix: restored pending count = 1).

Fix: `recordTurn` now tombstones a pair whose transcript write recorded nothing (`tombstoneLastPairPersistedLocked` — an empty marker turn keeps pair-log positions stable), and `publishFoldTransaction` filters tombstone turns out of the rewrite tail; the `tailStart` clamp remains as invariant armor. Verified red/green: pre-fix the test reproduces the resurrection, post-fix the restore comes up idle with the failed pair gone. Positional stability is pinned separately by `TestFoldTail_FailedPairTombstoneKeepsSnapshotPositions` (a fold that snapshots mid-failing-write and publishes after a later successful pair keeps that pair in the resumed history; red against the earlier pop-based fix, green with the tombstone), plus `TestDeriveRestoredAskPending_FailedAskPairNeverCollects` pinning the derivation rule.

**Low — malformed-ask warning (`isAskRound` discarded on the live path):** follow-up per review policy, journaled for the follow-up chain.

obra:
**Head advanced 2f1e99ac58 → 8737ae15c0 after the base merge.** Two mechanical events, both resolved:

- The branch deletion on #2005's squash merge closed this stacked PR instead of retargeting it (the base ref vanished). Recovered by restoring the ref, reopening, retargeting the base to main, and re-deleting the ref. The branch deletion is inert now that nothing is based on it.
- The retarget surfaced a content conflict in `agent/session_state.go`: both sides carried #2005's region with no shared ancestry (squash merges create none). Verified main's side byte-identical to #2005's own head (lines 300-303 of 755f6a1ecd) and this branch's side a strict superset — #2005's content evolved by the four boundary commits. Resolution keeps the branch side; the resolved file has a 0-line delta against the pre-merge version.

Gates on the merged tree: full agent package suite green (289s), gofmt/vet clean, RoboRev cycle 5 on the true delta vs main (job 23104): **no issues**.

obra:
**Disposition of the combined review on `2767f84`**

**Medium — `session_compaction.go:245` tail start not clamped on the lower bound: FIXED in this PR.** The reviewer's own note flags the disagreement; confirmed against the code: the clamp only ceiled the upper bound. The invariant argument (publications alone advance `persistedAppendLogBase`, so a live snapshot can't fall below it) says the negative case is unreachable, but armor should fail closed at both ends: the clamp is now `max(0, min(tailStart, len(s.persistedAppendLog)))`, with the comment stating both bounds as invariant armor.

**Medium — `session_state.go:308` live `s.history` not reconciled at the restored-failure boundary: DISPOSITIONED, not dismissed; design change journaled as follow-up.** The observation is correct about the code — the door derives `state`/`askPending` from the transcript-derived view and does not reassign `s.history` — and it is contract, not oversight:

- recordTurn's failure path keeps the live turn deliberately ("The live turn stays for the caller's own failure handling") and emits a transcript-write-failed warning. Live retention on write failure predates this PR (verified against the pre-PR `recordTurn` at `5167475c68`: append to history, write, retain on error, warn). This PR removed the *durable* copy — the pair-log tombstone plus the publication filter — so the durable state is consistent.
- The door's scope is durable-state settlement: `state` and `askPending` re-derived from the same transcript-tail rule restore uses, which is the live/restore state divergence this chain exists to close. After the door the durable state is provably consistent: `TestFoldTail_CanceledAskPairRidingTailDoesNotResurrectAfterRestore` (restore idle, no posted ack in the resumed history) and `TestFoldTail_FailedPairTombstoneKeepsSnapshotPositions` (a pair recorded after a failed write survives the fold at a stable position).
- What remains is bounded to the live model context: a turn whose durable write failed can stay in the in-memory history, including surviving a fold through the `s.history[snapLen:]` merge-back. Removing that means reconciling live context to the orphan-repaired transcript view at the door (or caller-side pruning) — a coordinate-sensitive design change (repair splices entering live history, fork-divergence remapping) that needs its own mechanism test. Journaled as the follow-up's subject rather than landed hastily here.

The two Lows (the tombstone comment's wrong write-door mechanism; the unbounded transcript read under the write door) follow the review policy: immediate follow-up PR, together with the un-addressed items from the prior review (the divergence-mapping helper extraction, the `restoredRepairInsertions` coverage test, and the live-path `isAskRound` warning).

obra:
**Disposition of the combined review on `a3a5b7b`**

**Medium — a failed `recordTurn`'s live turn can be encoded into a later compaction summary: DISPOSITIONED as the reviewers' own open question; joins the journaled follow-up.** The synthesis itself marks this "a separate compaction-window path not exercised by the added tests, so treat it as an open question rather than a confirmed regression," and one of the three reviewers found no issue in the changed code. The mechanism it generalizes is the recordTurn contract dispositioned on the previous review: the live turn stays on write failure by documented, pre-existing design (verified against `5167475c68`; a transcript-write-failed warning is emitted), and every durable structure is cleaned — pair-log tombstone, publication filter, restore-time derivation. What this finding adds is that a fold's summary prose can *mention* the phantom turn because the summarizer reads live history. That mention cannot resurrect durable ask state: `deriveRestoredAskPending` runs over turn structures, not summary text, and the resurrection this chain exists to close is pinned by `TestFoldTail_CanceledAskPairRidingTailDoesNotResurrectAfterRestore` and `TestFoldTail_FailedPairTombstoneKeepsSnapshotPositions`. Marking failed turns non-foldable (the suggested fix) is a real design change with its own tradeoffs — the model just produced that turn, so a summary that silently omits it opens a different live/restore gap — and it needs its own mechanism test. It is the immediate follow-up's subject, together with the live-context reconciliation item from the previous disposition.

**Low — the garbled tombstone doc comment:** follow-up per review policy; the reviewers' suggested wording is adopted there verbatim. (Third consecutive review to flag it; the follow-up lands immediately after this merge.)

PR #2068, opened 2026-09-20
fix(sdk,web): bind the hub's applied marketplace removal outcome

obra:
## Summary

Server #2050 added the `marketplaceRemoveApplied` WireError: a marketplace removal whose unregister AND clone cleanup both completed, but whose response's updated-list read failed. appwire/errors.go documents the rule: consumers reconcile, never retry (a retry finds `ErrMarketplaceNotFound`), never read the missing list as empty. That PR deferred the client binding to the marketplace reconciliation successors — this is it.

- **SDK**: `marketplaceRemovalOutcome()` gains `{ kind: "removed" }`, returned on the marker alone. The marker is emitted from exactly one hub site, always with `AppliedUnavailable: true` (cmd/evener-hub/app_plugins.go), so no payload shape may demote a standing removal back to a retryable failure — the same marker-is-proof policy the clone-litter sibling already follows. `ErrorMarketplaceRemoveApplied` is exported from `errors.ts`, bound to the Go constant by a `goErrorInfo` test (the `ErrorInstanceRemoveApplied` pattern), and re-exported from the package root.
- **Web**: `MarketplaceSheet` shows a neutral info toast for the removed kind instead of the clone-litter warning, closes the confirmation, and lets the existing reconciliation refresh settle the list. Red-green proven: the new sheet test fails without the production branch (the info toast never lands) and passes with it.

## Why one PR instead of SDK-then-web

The successor plan called for separate SDK and web PRs. Local RoboRev on the SDK-only delta found a Medium (job 22923): with the SDK change alone, the sheet's recognized-outcome branch shows the clone-litter warning for the removed kind. A Medium blocks landing, so both halves ship together.

## Review trail

- Local RoboRev on the full range: clean (job 22936).
- simplify-code 4-angle panel on the final diff: EFFICIENCY and REUSE clean; SIMPLIFICATION's 4 findings applied (single home for the per-marker docs, comment trims, store-test idiom); ALTITUDE's 2 findings skipped with reasons: store-depth reconciliation for removed/unavailable is a behavior change outside this diff (the handoff design puts reconciliation at the host), and the exhaustive binding test would refactor pre-existing describes outside the diff. Both journaled as follow-ups.
- Gates: 78 package tests, sheet suite 39/39, package qualification, frontend typecheck, Biome.

## Follow-ups (journaled)

- Native marketplace browser should consume `{ kind: "removed" }`.
- Decide with the native consumer whether removed/unavailable reconciliation moves to store depth (today the store reconciles only the clone-litter applied list; hosts refetch).
- One exhaustive `goErrorInfo` binding test for every exported `Error*` constant, replacing the per-constant describes.
- Bind the pre-existing `marketplaceUnregisteredCloneRemains` sibling literal to a constant too.


obra:
### Disposition: native marketplace UI (roborev Medium)

Retained deliberately, not fixed in this PR, matching how the native chain handled the same class of gap on #1960:

- This PR does not change native behavior. The generic failed-mutation report for `marketplaceRemoveApplied` predates it — the gap opened when the server landed the discriminator (#2050), and any client that does not classify the marker misreports the same way today on main.
- The successor chain scopes this PR to the SDK and the web consumer; the native marketplace browser adopts the classifier in its own follow-up PR (journaled in the description), which gets the full review treatment there.
- The SDK capability the follow-up needs — `{ kind: "removed" }` from `marketplaceRemovalOutcome()` — ships here, so the follow-up is consumer-side only.

PR #2071, opened 2026-09-20
fix(web): stop showing subagent attention, oversized quotes, and Job resumable

obra:
## What

Three reader-facing fixes to the subagent status card:

1. **"Needs attention" no longer appears on subagents.** The stable projection's `NeedsAttention` flag is wake-delivery plumbing: the controller arms it even while a generation runs (a queued wake the owner driver commits on its own), so the card could read "Running · Needs attention" on one line, and there was never anything for a reader to act on. The card, the collapsed lifecycle line, and the `job_status` chip now show only the lifecycle word. The backend flag itself stays — the wake machinery and the root session's own tabs consume it.
2. **The activity list's child agent messages render at UI size.** `.quoteMsg` still carried 18px prose size after #2048 UI-sized the card's quote; the hand-synced copies drifted exactly there. Now 13px like the rest of the list, keeping the italic serif quote treatment.
3. **"Job / Resumable" dropped.** Resumability is plumbing the owner acts on. The exhaustion budget line stays — it is what explains an "Exhausted" lifecycle. Resumability and attention also leave the memoized row's thread fingerprint, which again covers exactly what the row renders.

## Commits

- `a4371bdc99` fix(web): stop showing subagent attention, oversized quotes, and Job resumable
- `150b151ad9` refactor(web): simplify the subagent status-card change per review — four read-only reviewers (reuse, simplification, efficiency, altitude) covered the diff; the passthrough `DelegateStatusWord` component, a dead exhaustion guard, and six copies of the wake-plumbing rationale are gone. One composes-based CSS dedup was reverted: `faces.test.ts` pins the `.quote` rule's literal font-size, and adapting the contract test to the refactor is fix-the-test-forward.
- `1db8110f8e` fix(web): wait for font registration before the browser-guard font check — `fonts.ready` settles instantly on an empty FontFaceSet, so the shared guard check raced stylesheet application and misread "not arrived yet" as "declares none" under load (reproduced twice in full-gate runs, passing standalone). The check now polls up to 10s for registration; genuinely fontless documents and 404'd faces still fail exactly as before.

## Verification

- `make test-web`: PASS (typecheck, unit, lint) on the final tree — twice.
- Browser guards: layoutguard (including the retargeted editorial-transcript cases), overflowguard, and spawnguard each PASS standalone on the final tree; shellguard, transcriptscrollguard, retirementguard, and skillguard passed in full-suite runs.
- Tests were updated first and went red before the fix; 138 targeted tests green; `node --test scripts/*.test.mjs` 212/212.
- Note: the full `make test-web-browser` suite is flaky on this dev box under its normal sustained load (13-15 on 16 CPUs); each flake passed standalone immediately after. CI's dedicated runner is the authoritative run for it.


obra:
Addressing roborev's combined review on `75fe95f` in `6fde2a0`:

**Medium (browserGuardCdp.mjs font poll):** the in-page wait is now two separately bounded phases — a `FONT_POLL_DEADLINE_MS` (12000ms) registration poll and an `FONT_READY_TIMEOUT_MS` (8000ms) cap raced against `fonts.ready`, whose loss is returned as a stall and raised node-side as this guard's own environment-problem error. Five new wire tests pin the sum at 20000ms (10s headroom under the `evaluate()` 30000ms ceiling) plus the stall, poll-exhaustion, and diagnostic paths.

**Low (transcript-guard.ts anchor):** the anchor now matches the identity span's exact text anywhere in the card instead of coupling to `firstElementChild` position; the exact match still excludes future `dlg_editorial_running*` case ids.

Verification: `browserGuardCdp.test.js` 20/20; `editorial-transcript-360` layoutguard case PASS in real Chrome through the new in-page path; `make test-web` green (typecheck, vitest, lint).

obra:
Follow-up to roborev's `6fde2a0` review (the Low-severity coverage gap on the registration-poll transition): `0257ec` adds the suggested test — fonts.size/forEach flip to registered 100ms into a poll whose fonts.ready settles instantly, asserting the face is collected rather than reported fontless. Mutation-checked: delete the poll loop and only this test fails; restored, 21/21 pass and `make test-web` is green.

obra:
Addressing roborev's `6f62c0a` review (Medium: the 12s poll sat below the observed >10s stylesheet-application windows): `2d81696` puts the wait on one coordinated deadline — the poll owns the whole `FONT_POLL_DEADLINE_MS` budget (20000ms, the value that survived real load in `be34ee67`) and `fonts.ready` races the budget's remainder capped at `FONT_READY_TIMEOUT_MS`, so no in-page wait can exceed 20000ms and the evaluate() ceiling still fires last. Tests pin the >10s coverage floor, the 20000ms total bound, the zero-remainder stall, and the mid-poll transition (22/22); the editorial-transcript-360 layoutguard case PASSes in real Chrome; `make test-web` green.

PR #2072, opened 2026-09-20
fix(native): report an applied marketplace removal without a retry hint

obra:
## Summary

The native marketplace browser showed the generic write-failed copy for every removal rejection, including the ones the hub says already stood (`marketplaceRemoveApplied`; appwire/errors.go: reconcile, never retry) — a false failure after a successful removal, with copy that implies a retry targeting a marketplace that is already gone. This is the native consumer follow-up dispositioned in #2068's RoboRev review.

- The removal now flows through the SDK classifier: an applied removal refreshes a stale list and shows nothing; the clone-litter kinds keep their leftover-files warning; only an ordinary failure keeps `WRITE_FAILED`. The decisions live in `marketplaceBrowserModel` (`appliedRemovalNotice`, `refetchAfterRemoval`) with model and store-level tests; the view wires them through `act()`'s new failure hook — one gate/revision choreography, no separate `removeAct`.
- `refetchAfterRemoval` reads the store's **live** list through `getState()`, never a render snapshot: the store publishes the clone-litter applied list itself during the rejection, so a reconciled list triggers no refetch. The first local RoboRev pass found the snapshot version of this check (a redundant refetch plus a misleading loading error for an already-reconciled removal) — fixed, and covered by a store-level regression test plus screen-level tests.

## Review trail

- Local RoboRev: first pass found the stale-snapshot Medium; re-review on the fixed head is clean (job 22990).
- simplify-code 4-angle panel: EFFICIENCY converged with RoboRev on the stale snapshot (fixed); SIMPLIFICATION's act-fold, tri-state flatten, and comment trim applied; ALTITUDE's screen-test gap closed via the `renderNative.testkit` pattern (new `MarketplaceBrowser.removal.test.tsx` covering the neutral report with its reconciling refetch, the litter warning with no redundant refetch, and the ordinary failure). Skipped, journaled: a `runGatedMutation`/`pluginMutationGate` refactor that exposes caught errors to every caller (shared infrastructure beyond this diff), and an SDK-shared refetch predicate (the web's guard condition is load-bearing for both the pending-marker hook and the fetch, so a standalone predicate cannot replace it there).
- #1959 analysis: the change is guarded against its disposed-instance fetch scenario — a browser's unmount bumps its revision before disposing its store, so a late removal continuation returns before reaching the refetch. The issue stays open for its web lifecycle coverage and native model-lifetime items.
- Gates: `make test-native` green (full native suite, tsc, script-resolution check); 12 tests across the touched suites.

## Follow-ups (journaled)

- `runGatedMutation` could return the caught error so failure hooks need no per-caller capture.
- The store-depth reconcile question from #2068 remains open for the native chain.


obra:
### Disposition: stale applied removals when reconciliation fails (roborev Medium)

Retained deliberately, not fixed in this PR:

- The scenario needs the list read to fail twice in a row (the removal's response read, then the reconciling refetch). In that degraded-network state the store already surfaces the fetch failure with its own retry path; the browser's behavior beneath it is unchanged from before this PR.
- The behavior matches the shipped web consumer exactly (#2068's sheet: a failed reconcile leaves the stale list with a retryable row, and a re-remove reports the hub's ordinary not-found error). The two consumers are consistent, and the consumer-depth reconcile was the design ruled on in #2068's review.
- The suggested "perform reconciliation even when the current screen revision has changed" would reintroduce the disposed-instance hazard #1959 documents: the browser bumps its revision on unmount before disposing its store precisely so a late continuation cannot touch the replaced screen or fetch through the disposed store.
- The real improvement — tracking applied removals until the authoritative list drops the name, with the removal action disabled meanwhile — belongs to #1959's native model-lifetime family, which that issue says is being separated into meaningful review slices. Cross-referenced there for the native chain.

Issue #270, opened 2026-08-20
route npmShimEnv's shim write through writeExecutable (ForkLock) for ETXTBSY defense-in-depth


obra:
Follow-up from the PR #258 review: `install_test.go` already carries the precise root-cause fix for the golang/go#22315 ETXTBSY race — `writeExecutable` (~line 933) wraps the write in `syscall.ForkLock.RLock()/RUnlock()`, which excludes concurrent Go-spawned forks for the write's duration (verified against Go 1.26.5's `syscall.forkExec`, which holds `ForkLock` exclusively across fork→execve).

`npmShimEnv`'s own shim write (~line 886, the exact site issue #248 reported) uses a raw `os.WriteFile` instead. Routing it through `writeExecutable` is a one-line consistency/defense-in-depth improvement. The retry added by PR #258 stays regardless — `ForkLock` is Go-runtime-local and gives zero protection for grandchild execs (tsc written by `npm ci`, exec'd inside a make/shell subtree), which the retry covers.

https://claude.ai/code/session_0134LFae5DzpDExkuhJTCNfU

obra:
## Evidence
`npmShimEnv` raw-writes npm while existing `writeExecutable` uses `ForkLock`. Historical Linux CI recorded `fork/exec .../npm: text file busy`; focused tests pass ×3.

## Root cause
The bypass lets a concurrent Go child inherit the writable descriptor.

## Correct plan
Replace the raw block with `writeExecutable(t, filepath.Join(fakeBin, "npm"), npmShim)`; retain ETXTBSY retry.

## Risks
Assertions remain unchanged; ForkLock is process-local. Correction: Go releases it before status-pipe proof of `execve`; exclusion still works because no fork occurs while the writer fd exists. PR needed; keep `enhancement`, `dev-tooling`.

## Test plan
Repeat focused npm/preflight/install tests on Linux; require helper use; add no Darwin timing test.

PR #2034, opened 2026-09-19
fix(plugins): sweep leftover marketplace clones in gc (#1923)

obra:
Fixes #1923.

## Root cause

A marketplace removal, add or edit whose directory cleanup failed leaves its
clone under the marketplaces directory while no record names it. `Gc` swept only
the plugin cache (`cache/<marketplace>/<plugin>/<sha>`), and `Doctor` iterated
only registered marketplaces, so nothing reclaimed the clone. It outlived every
command and blocked a later rename onto the name with
"it must be deleted before %q can be reused" — naming no command that deletes it.

## Change

- `Gc` now also sweeps `marketplacesDir()` for clone dirs that no recorded
  marketplace's `InstallLocation` names.
- A leftover **symlink** occupies the name as a directory does, so it is swept
  too. It is compared as the name it occupies (ancestors resolved, the final
  link not followed), and `RemoveAll` removes the link, never its target.
- It keeps the two fixed scratch names `.staging`/`.old` — `.old` can be the
  only local copy of a clone a failed swap could not restore — and any directory
  a recorded directory source sits at or beneath, reusing
  `sweepDestroysSource`/`marketplaceProtectionPaths`, the same protection
  `RemoveMarketplace`'s own clone sweep applies.
- `Doctor`'s orphan walk reports the same directories with a
  `run \`evener plugin gc\`` remediation, so the diagnostic names the command
  that clears them. Gc and Doctor share `referencedClonePaths`,
  `isScratchCloneName` and the source-protection predicate, so they cannot
  disagree.
- Both store files are read before either sweep mutates disk, and a
  marketplaces-file failure in the Doctor walk is an additional finding rather
  than a replacement that hides the cache-dir findings already computed.
- The clone-removal failure error is path-free per #1700: the raw
  `*fs.PathError` goes to the log, the returned error names only the marketplace.
  Clones already removed before a later removal fails are still reported.
- Caller wording: `evener plugin gc` and the hub maintenance log no longer say
  "cache dir(s)" now that the sweep removes clone dirs too.

## Test evidence

`internal/plugins/gc_test.go`:

- `TestGc_RemovesALeftoverMarketplaceClone` **failed before the fix** with
  `Gc removed = [], want [.../marketplaces/acme]`; passes after.
- `TestGc_RemovesALeftoverMarketplaceCloneSymlink` and
  `TestGc_RemovesALeftoverSymlinkThatPointsAtALiveClone` failed before the
  symlink fix and pass after (the second also proves the final link is not
[...]

PR #2085, opened 2026-09-20
test(install): assert the ForkLock discipline, not one call site's spelling (#1204)

### LINKED ISSUE #1204: tests: the npm shim ForkLock audit pins one call site's spelling instead of the discipline
`npmshim_forklock_audit_test.go` (landed with #606 for #270) parses `install_test.go` with `go/ast` and asserts that the npm shim write goes through the ForkLock-holding `writeExecutable` helper. The product fix it guards — a one-line reroute into that helper — is correct and minimal. The audit around it is weaker than it reads: it pins one call site's spelling rather than the discipline, and every gap below is in the audit, not in the fix.

## What it asserts

`TestNpmShimWriteGoesThroughForkLock` (`npmshim_forklock_audit_test.go:32`) requires, of `npmShimEnv`:

1. at least one `writeExecutable` call (`:38`), and at least one such call whose argument expression mentions the npm path — the `"npm"` string literal or the `fakeBin` identifier, directly or through one level of assignment (`callMentionsNpmPath` `:172`, `assignmentMentionsNpmPath` `:194`);
2. no raw `os.WriteFile` with an executable literal mode (`:68`), and no direct `os.Create`/`OpenFile`/`CreateTemp` at all (`rawWritePathNonWriteFile` `:153`, checked at `:82`);

and of `writeExecutable` (`install_test.go:1314`):

3. a statement-level `syscall.ForkLock.RLock()` in the statement list that encloses its `os.WriteFile` (`holdsForkLockAcross` `:339`, checked at `:102`).

## Four ways that is not "the write is locked"

**1. A raw executable write passes behind a decoy.** Condition 2 refuses a raw write only when its mode literal has an execute bit, so `os.WriteFile(path, b, 0o644)` is read as proof the file is not executable, and `os.Chmod` is never inspected (the word does not appear in the file at all). So: write the shim raw at `0o644` through a path expression that avoids the `"npm"`/`fakeBin` tokens, add `os.Chmod(shim, 0o755)`, and satisfy condition 1 with a decoy `writeExecutable` call on some other file. The audit passes with the hazardous write in place. An independent review pass reproduced exactly this as one of six mutations in a scratch harness built from the PR's source.

**2. Benign refactors false-RED.** Condition 1 is a token match, so renaming `fakeBin`, spelling the shim path as `filepath.Join(binDir, shimName)`, or extracting the write into a helper that still calls `writeExecutable` (leaving `npmShimEnv` calling that helper) fails the audit with behavior unchanged. Condition 3 likewise fails if the lock moves into a `withForkLock(func(){...})` wrapper, which holds it just as well.

**3. The release is never required.** `holdsForkLockAcross` returns the lock state *at* the write, and its own comment says a deferred release counts as neither acquisition nor release, so deleting `defer syscall.ForkLock.RUnlock()` still passes. A read lock held forever blocks any later writer of `ForkLock`.

**4. One function in one file.** The same discipline is duplicated and unchecked at `internal/binresolve/sibling_test.go:185` and `cmd/evener-tui/internal/hubstart/hub_start_test.go:473` (both carry the comment "matching install_test.go's writeExecutable"), and `agent/sandbox/escape_test.go:83` writes a fixture binary at `0o755` raw.

## General mechanism

The invariant belongs to the write path, not to one call site's spelling:

- make the locked helper the only way a test executable is written (an unexported helper in a shared test-support package, so a raw write is never the shorter option), or
- enforce it repo-wide from the existing `evener-dev` / `make lint` tooling.

Either makes the AST audit unnecessary. Until one of them lands, the audit is a single-call-site tripwire: it catches the specific regression of `npmShimEnv` reverting to a raw write spelled with the familiar tokens, and should not be read as the ForkLock discipline.

## Provenance

Ways 2-4 are read off the landed file on `main`, and their line numbers are current `main`. Way 1 was reproduced by an independent review pass that extracted the audit verbatim into a stdlib-only scratch harness and ran six mutations against the PR's `install_test.go` (baseline PASS); the repo gates were not run against that reconstructed source, so treat it as evidence about the audit's logic rather than as a CI-verified result.

obra:
Fixes #1204

## Root cause

`npmshim_forklock_audit_test.go` (landed with #606 for #270) is supposed to pin
the ForkLock discipline for the npm shim write. It pinned one call site's
*spelling* instead. The landing fix it guards — `npmShimEnv` routing its write
through the ForkLock-holding `writeExecutable` helper — is correct and
minimal; every gap was in the audit:

1. **Token matching.** The audit decided which `writeExecutable` call mattered
   by looking for the `"npm"` string literal or the `fakeBin` identifier
   (`callMentionsNpmPath`, `assignmentMentionsNpmPath`).
2. **Modelled executability only for `os.WriteFile` literals.** A raw
   `os.WriteFile(path, b, 0o644)` was accepted as proof the file was not
   executable, and `os.Chmod` was never inspected.
3. **The release was never required.** `holdsForkLockAcross` returned the lock
   state *at* the write, so deleting `defer syscall.ForkLock.RUnlock()` passed:
   a read lock held forever blocks every later `ForkLock` writer.
4. **One function in one file.** The same discipline is duplicated, unchecked,
   at `internal/binresolve/sibling_test.go:185` and
   `cmd/evener-tui/internal/hubstart/hub_start_test.go:473`.

All four were confirmed against current `origin/main` (`87d836b56a`) with a
scratch harness that ran the **old** audit verbatim against mutations of a
compilable fixture:

```
EVASION raw write behind a decoy, made executable by chmod: OLD AUDIT PASSED (accepted)
EVASION helper keeps the write but drops the deferred release: OLD AUDIT PASSED (accepted)
BENIGN directory and shim name hoisted into differently named variables: OLD AUDIT FAILED (rejected)
    old_audit_test.go:56: no writeExecutable call in npmShimEnv mentions the npm path
BENIGN write extracted into a helper that still calls writeExecutable: OLD AUDIT FAILED (rejected)
    old_audit_test.go:40: npmShimEnv no longer calls writeExecutable
```

The audit was weaker than it read on the hazard, and false-RED on benign
refactors at the same time.

## Change

The audit now asserts the discipline, not the spelling:

- Every `os.WriteFile` **reachable from `npmShimEnv`** — resolved through
  same-file helpers it calls — that can produce an executable file must hold
  `syscall.ForkLock` for reading across the write **and release it afterwards**.
- No executable write may bypass that guard: a literal execute mode, a
  non-literal mode that proves nothing, a **chmod to an execute mode** (the
  write-then-chmod evasion), or the modeless `os.Create`/`OpenFile`/`CreateTemp`.
- At least one guarded executable write must be reachable, so the shim is still
  actually written through the guarded helper.
- Identifiers are no longer matched: renaming `fakeBin`, spelling the path
  through `filepath.Join`, hoisting it into a variable, or extracting the write
  into a helper all keep passing.
- The two duplicated `writeExecutable` helpers the issue names are checked with
  the same predicate.

The audit's scope is stated in the file: it does **not** lint every test in the
repo for executable writes. Routing all of those through one shared guarded
helper (or a repo-wide lint) is the architecture question the issue leaves
open; this change deliberately does not decide it, and
`agent/sandbox/escape_test.go:83` (raw `0o755`) is unchanged.

## Test evidence

Fails before, passes after, on the **real** `install_test.go` (mutated in
place, then restored):

```
=== A: dropped defer RUnlock ===
--- FAIL: TestNpmShimWriteGoesThroughForkLock
    install_test.go: writeExecutable holds syscall.ForkLock across its write but never releases it
=== B: raw 0o644 + chmod 0o755 behind decoy ===
--- FAIL: TestNpmShimWriteGoesThroughForkLock
    install_test.go: npmShimEnv chmods a file to 0o755: an executable created by a raw write
    and made executable afterwards bypasses the ForkLock guard
```

The new tests encode both directions:

- `TestNpmShimForkLockAuditRejectsEvasions` — 29 mutations, every one must
  produce a finding: chmod evasion, token-free raw exec write, dropped release,
  dropped lock, `os.CreateTemp`, guarded helper no longer reachable, raw exec
  write inside a switch case, inside a closure, reached through a local
  function value, **through a callback argument**, **through a package-level
  function value**, a release that only runs on one branch, a double `RLock`
  under a single deferred `RUnlock`, **a raw writer replaced only conditionally
  by a guarded one**, **a release inside the branch instead of after it**,
  **a raw writer behind a local `var` alias**, **a partial release under a
  deeper lock**, **a deferred release registered after the write**, **a plain
  release a `t.Fatal` skips**, **an executable chmod through a file handle**,
  **a raw writer stored in a function slice**, **a raw writer returned from a
  helper**, **a release and an acquisition in opposite branches**, **a
  pre-write release credited to a later skipped release**, **an executable
  write through an aliased import**, **an executable write deferred until after
  the release**, **a non-literal write mode**, **a raw writer stored in a keyed
  map literal and called from a range**, and **a lock taken outside the branch
  that registers the release**.
- `TestNpmShimForkLockAuditAcceptsBenignRefactors` — 6 refactors that keep the
  discipline must stay green (renamed directory/shim variables, write extracted
  into a helper, lock/unlock inside a retry loop body, guarded write reached
  through a local function value, guarded write inside a nested block with a
  deferred release, and a literal `0o644` config write beside the shim).

The lock **depth** is threaded through every statement list — blocks,
if/for/range bodies, switch and select cases, and function literals — so a
write buried in a case or a closure cannot hide behind a guarded decoy, a
second `RLock` is not satisfied by one release, and a deferred release beside
the write inside any scope on its chain counts (so a guarded write inside a
nested block using `defer` still passes). A plain release must be in the list
that established the lock — a release inside the branch beside the write does
not close a lock acquired outside it, and is rejected. Reachability follows
same-file functions, all targets of an alias (so a conditional reassignment
cannot hide a raw writer), local (`:=` and `var`) and package-level function
values, and function identifiers passed as callback arguments. A release
closes exactly one acquisition (depth decrements, never resets), and a
deferred release must be registered before the write, so a write whose error
branch returns cannot leak the lock.

## Roborev rounds on this head's predecessors

Seven per-commit reviews were raised and fully addressed before this head:
`isIdentCall` dead code (a `lint-golangci` failure), traversal blind to
switch/closure bodies, unresolved local function-value calls, a loose
defer-release rule, boolean lock state missing a double-acquire, and a
`releasedAfter` limited to the write's innermost list, a single-target alias
map that dropped conditional reassignments, branch-local releases being
accepted, a nested-scope `defer` release being rejected, local `var` aliases
being missed, a partial release resetting the depth, and a late `defer`
counting as a release, function values in slices/maps not being followed,
`(*os.File).Chmod` not being detected, a plain release an early exit can skip,
and a nested `RUnlock` resetting the outer depth, function values returned from
helpers not being followed, `releasesAfter` double-counting a pre-write
release, aliased imports being invisible, and `defer os.WriteFile` inheriting
the enclosing depth, keyed-literal and ranged function values not being
recognised, and a deferred release in a conditional block being credited
against an unconditional acquisition. All are fixed at the root above, with a
mutation case each.

## Deliberate fail-closed tradeoffs (stated, not hidden)

This audit is a *syntactic* approximation, not a path-sensitive dataflow
analysis. Where two reviewers' demands conflict, the audit takes the safe
direction — a loud false-RED rather than a silent pass:

- **Branch-local lock manipulation is refused, not modelled.** A compound
  statement containing a plain `RUnlock` clears the held depth, so a release
  and an acquisition in opposite branches is rejected instead of averaged. The
  cost is that a *balanced* nested `RLock`/`RUnlock` pair inside a branch also
  clears the depth and false-REDs. Modelling that per path needs
  control-flow analysis this test is not.
- **A plain `RUnlock` is only credited when no early exit can skip it** —
  which is why the earlier benign case "lock taken before an `if` and released
  after it" was removed rather than kept green: with `t.Fatalf` in the error
  branch it is genuinely leaky (`t.Fatalf` runs `runtime.Goexit`, so the
  `RUnlock` never executes and `syscall.ForkLock` stays read-locked for the
  rest of the process).
- **Indirection the walk cannot resolve fails closed**: calls through a
  slice/map element or another call's result, a known function value stored in
  a composite literal, and a local bound from a call and then called. Each
  could hide an unguarded writer behind a guarded decoy.

The structural alternative is the one the issue itself names and its triage
flagged `needs-human`: make the guarded write helper the only way a test
executable is written (a shared test-support helper), or enforce the rule
repo-wide from `evener-dev` / `make lint`. Either makes this AST audit
unnecessary.

One consequence worth stating: the earlier benign case "lock taken before an
`if` and released after it" was **removed rather than kept green**. With
`t.Fatalf` in the error branch it is genuinely leaky — `t.Fatalf` runs
`runtime.Goexit`, so the plain `RUnlock` never executes and `syscall.ForkLock`
stays read-locked for the rest of the process. The audit now rejects it, which
is the correct outcome.

## Remaining Low (not fixed here)

`FOLLOW-UP (Low): npmshim_forklock_audit_test.go:77 — the cross-package helper
audit reads `internal/binresolve/sibling_test.go` and
`cmd/evener-tui/internal/hubstart/hub_start_test.go` at runtime, but those
files are not inputs to the root package's Go test cache, so a later change
only to a duplicated helper can reuse a cached root test result and skip the
audit. Fix: give each owning package its own copy of the guard, or a
compile-time dependency on the checked source.` This is a test-infrastructure
staleness gap, not a correctness bug in the audited code; fixing it means
adding the same predicate to two other packages (or a new shared
test-support package), which is the same open architecture question the issue
raises.

Run:

```
go test . -run 'NpmShim|ForkLock|DuplicatedWriteExecutable' -count=1
ok  	primeradiant.com/evener	0.051s

go vet .
gofmt -l npmshim_forklock_audit_test.go   # clean
```

## Known limitation (fail-closed)

Moving the lock into a `withForkLock(func(){...})` wrapper now fails the audit
rather than passing it. Verifying that shape needs interprocedural reasoning
about a wrapper that may live in another package; the audit fails loudly with a
message naming the shape instead of silently accepting it. That is deliberate:
the defect being fixed was silent acceptance.


obra:
Parking this PR rather than taking it to an eighth roborev round.

The rewrite did its job — the audit no longer pins one call site's spelling, and sixteen real findings across six rounds were fixed with mutation cases (one rejected with evidence). But the finding count is not shrinking: rounds went 4, 4, 3, 4, 4, 6, and this round is 3 Medium + 2 Low, all gaps in the same control-flow analysis. The verdict itself says the two control-flow items "must be fixed for the control-flow audit to hold".

That is structural. The audit reimplements "does this function hold the ForkLock across every write it performs" as a syntactic walk, which is a dataflow and reachability problem with a large case surface — `break`/`continue`/`goto`, early returns between acquire and release, function values through callbacks, package-level aliases, nested scopes. Each round's fix reveals the next unmodeled construct, and today an unmodeled construct simply passes.

The outstanding findings and that diagnosis are filed as a follow-up, which asks for a decision: replace the walk with a real analysis (`go/analysis` + SSA, where reachability and aliasing are already solved), or explicitly close the set of modeled constructs and make everything else fail closed.

Leaving this open and parked with the branch preserved, matching the handling of PRs #2019, #2013, #2034 and #1813. Never merged.

PR #2086, opened 2026-09-20
fix(execenv): make TMPDIR usable by a privilege-dropping child (#495)

### LINKED ISSUE #495: Session scratch is exported as TMPDIR at mode 0700, so any privilege-dropping child fails with Permission denied on a path the user never chose

obra:
Fixes #495.

## Problem

`TMPDIR` was set to the session scratch for **every** spawned process, and that
scratch is `0700`. Any descendant that drops privileges inherits a `TMPDIR` it
cannot write and fails with a bare `Permission denied` naming a path the user
never chose — TB 2.1 trial `configure-git-webserver__52rZDfB` step 85, a git
`post-receive` hook running as the pushing user. `0700` is right for the agent's
own scratch; the defect was exporting that private directory as a world-facing
inherited variable.

## The rule

| Spawn shape | `EVENER_SCRATCH_DIR` | `TMPDIR` |
|---|---|---|
| sandboxed | private `0700` scratch | same scratch (from `sandbox.ApplyEnvFloor`; unchanged) |
| unsandboxed, file tools **confined** (`Sandbox != nil && Sandbox.FileToolConfined()` — the write-blocked off policy) | private `0700` scratch | same scratch |
| unsandboxed, file tools **unconfined** | private `0700` scratch | `<host-temp>/evener-sandbox-<unique>/tmp` |

The confined unsandboxed shape keeps the scratch deliberately: that scratch is
the file tools' only writable root, so pointing the shell's temp elsewhere would
break a `mktemp`-then-`write_file` workflow, and the sandbox prompt line that
names it stays true. Only the unconfined shape — the one a privilege-dropping
child actually comes from — gets a container.

`TMPDIR == EVENER_SCRATCH_DIR` was normative: the 2026-07-15 decision record's
"Environment" section said *"Evener sets both"*, and its "Testing" section
required *"`TMPDIR`, `EVENER_SCRATCH_DIR`, environment info, and prompt use the
same path"*. Both are **amended in this PR** (dated, Jesse, #495) rather than
violated quietly. `EVENER_SCRATCH_DIR`, the file-tool root, and the private
scratch's `0700` are unchanged.

## The container

`<host-temp>/evener-sandbox-<unique>/tmp`:

- `<host-temp>` is `/tmp` or `/var/tmp`, chosen OS-aware and deliberately **not**
  `os.TempDir()` — on macOS that is the per-user `0700` `/var/folders/...`
  directory, which is the same bug; on Linux an ambient `TMPDIR` (systemd
  `PrivateTmp`, a container image, `~/tmp`) can be private too.
- the container is `0711`: another uid can traverse to the leaf but cannot create
  or hold `.evener-session.lock`, so the reclaim cannot be squatted.
- it is named under the **existing** `sessionScratchPrefix`, so the existing
  `Cleanup` gate and the crashed-scratch sweep apply with no new prefix.
- the leaf is `1777` + sticky — an arbitrary uid can `mktemp -d` in it, and no uid
  can remove or rename another's entry.
- `MkdirTemp`-style uniqueness, failing closed when the created directory is not
  owned by this process.
- reaped by the **existing** crashed-scratch sweep, which now also walks the
  world-usable host temp bases, under the same 24h lease discipline; the
  container carries a lease like a session scratch does.
- POSIX platforms only. The container's contract is POSIX mode bits and a
  descendant being able to become another user; where it cannot exist (Windows)
  `TMPDIR` keeps the session scratch for every shape rather than being left
  inherited, and the capability preamble says so too.

Removal at session close is attempted and **best-effort, disclosed**: a
non-empty nested `0700` subtree another uid left in the leaf cannot be unlinked
by the container's owner (an *empty* one can — the sticky bit lets the leaf's
owner unlink any entry). The lease is released either way, so the sweep retries
and reports what it cannot remove instead of aborting the base — no worse than
raw `/tmp`.

## Files

```
 agent/execenv/local.go                             | 176 ++++++++--
 agent/execenv/local_env_test.go                    | 128 +++++++-
 agent/execenv/process_runtime_program_fuzz_test.go |  43 ++-
 agent/sandbox/session_scratch.go                   |  10 +-
 agent/sandbox/session_scratch_test.go              |   7 +
 agent/sandbox/session_tmp.go                       | 185 +++++++++++
 agent/sandbox/session_tmp_owner_unix.go            |  24 ++
 agent/sandbox/session_tmp_owner_windows.go         |  18 ++
 agent/sandbox/session_tmp_test.go                  | 350 +++++++++++++++++
 agent/session_capabilities.go                      |  25 +-
 agent/session_capabilities_test.go                 |  25 +-
 docs/developing-evener/environment.md              |   2 +-
 docs/sandboxing.md                                 |   4 +-
 docs/superpowers/specs/...posture-design.md        |  42 ++-
 envvars/envvars.go                                 |   2 +-
 15 files changed, 986 insertions(+), 50 deletions(-)
```

## Tests

New:

- `TestSessionTmpContainerShape` — prefix, `0711` container, `1777`+sticky leaf,
  lease held/released, `Retain` keeps the directory.
- `TestSessionTmpUsableByArbitraryUID` — the #495 regression. Asserts the modes
  unconditionally (they *are* the foreign-user mechanism) and, when the host
  allows a privilege drop, actually runs `mktemp -d` as `nobody` in a freshly
  created container and asserts it succeeds inside the leaf; also asserts the
  owner can unlink a flat foreign entry.
- `TestSessionTmpRequiresWorldUsableBase` — a `0700` base, a `1776` base (writable
  but not traversable by another uid), a missing base and an empty base set are
  all refused, and a refusal leaves nothing behind.
- `TestSessionTmpFallsBackToALaterBase` — a candidate that cannot serve does not
  end provisioning while a later candidate can.
- `TestSessionTmpRemoveReclaimsContainer` + refusal outside the namespace.
- `TestSessionTmpContainerReapedByCrashedSweep` — stale released container is
  reaped, stale live-lease container is kept.
- `TestSessionTmpSweepReportsUnremovableContainer` — an unremovable container is
  reported and left in place, and the sweep still reclaims the rest (no abort).
- `TestCommandEnvironment_UnsandboxedTmpContainerReclaimedAtClose` — close
  removes the container and retains the scratch.
- `TestCommandEnvironment_TmpContainerFailureLeavesTmpDirInherited` — a failed
  provisioning leaves `TMPDIR` inherited (never re-pointed at the scratch) and is
  sticky.
- `TestCommandEnvironment_TmpContainerExportedWhenScratchUnavailable` — the two
  variables are provisioned independently, so a scratch failure still hands the
  spawn a world-usable `TMPDIR`.
- `TestCapabilityPreambleWriteBlockedKeepsTmpDir` — the file-tool-confined shape
  keeps naming `$TMPDIR`.

Changed (with reasons):

- `TestCommandEnvironment_UnsandboxedSessionExportsScratchVars` — the contract
  being changed; rewritten to the new rule, keeping its intent (both vars are
  exported and usable; scratch is private `0700`, temp is world-usable).
- `process_runtime_program_fuzz_test.go` — the fixture carries no policy, so it is
  **unconfined**-unsandboxed (determined empirically, not guessed: the old
  assertion fails with `TMPDIR=/tmp/evener-sandbox-.../tmp` against the scratch).
  Its `TMPDIR` expectation moves to the container, and the container path is
  normalized with the scratch paths or the determinism oracle would see a fresh
  random path every replay.
- the four **unconfined** capability goldens
  (`session_capabilities_test.go:199,249,309,362`). The four sandboxed ones
  (`:70,97,136,224`) keep `$TMPDIR` untouched.
- `TestSweepCrashedSessionScratchReportsUnusableBase` — fixture extended to cover
  the new base class; the assertion (a sweep with nothing it can read reports
  failure) is unchanged.

Unchanged on purpose: `sandbox_lifecycle_test.go:417-423` (write-blocked shape
keeps `TMPDIR == scratch`), `sandbox_delegate_create_test.go:417` and
`session_prompts.go:258-260` (the prompt line renders only when
`FileToolConfined()`, and every such shape keeps the scratch).

## Verification

```
go build ./...                                          # exit 0
make vet                                                # exit 0
GOOS=windows go vet ./agent/... ./envvars/...           # exit 0
go test -count=1 ./agent/execenv/... ./agent/sandbox/... # ok
go test -count=1 ./agent/                                # ok
go test -race -count=1 ./agent/execenv/... ./agent/sandbox/...  # ok
go test -tags evenerfuzz ./agent/execenv/ -run '^FuzzProcessRuntimeProgram$'   # 4 seeds pass
make lint                                                # all steps PASS
make fuzz-gap-check fuzz-registry-check                  # PASS
```

### Field-vector check (this host, `jesse` → `nobody`)

```
$ sudo -n -u nobody env TMPDIR=<private 0700 dir> sh -c 'mktemp -d'
mktemp: failed to create directory via template '…/tmp.XXXXXXXXXX': Permission denied
$ sudo -n -u nobody env TMPDIR=<container leaf> sh -c 'mktemp -d'
/tmp/evener-sandbox-…/tmp/tmp.0CSfwEpUeK
```

The same command the field trial failed on now succeeds against the shape this
PR produces. (Note: default `sudo` strips `TMPDIR`, so a live end-to-end
"inherited `TMPDIR` survives the drop" test is not constructible on this host —
the explicit form above is the reproduced vector, as the design spec records.)

## Known cost, disclosed

Host-temp residue is unbounded in a shared namespace — the spec's R9 — and the
mitigation is the reclaim, not containment:

- a session's close removes its container (best-effort, above);
- a crashed session's container is reclaimed by the existing 24h sweep;
- a test that mints an env and never closes it leaves a container behind, exactly
  as it already left a private scratch behind. Measured: one
  `go test ./agent/execenv/` run leaves ~49 containers in `/tmp` (the same suite
  already leaves tens of private scratch dirs per run). They are `0711`
  containers holding an empty `1777` leaf, owned by the test user, and the sweep
  reclaims them after 24h.

## Scope

Deliberately NOT revived from the closed-unmerged prior attempt (#2032): no
container-mode/traversability/lease/migration work, no new prefix, no change to
the sandboxed assignment, no change to the private scratch's mode or to the
file-tool root.


obra:
## Review round 1 — all six findings fixed

Thanks — the triage was precise and I agree with all six. Fixed at the root, rebased onto the current `main` (it advanced while this round was in progress), and pushed as `206a3a3b8f` — one push, so CI and the review run once for the new head.

**Medium**

1. **`execenv/local.go` — container provisioned only when the scratch succeeded.** Correct, and it defeated the point on exactly the host where it matters: the child then inherited an ambient `TMPDIR` that may itself be private. `overlaySessionEnv` now provisions the two **independently** — a scratch failure suppresses only `EVENER_SCRATCH_DIR`, while the container is still minted and exported. New test: `TestCommandEnvironment_TmpContainerExportedWhenScratchUnavailable` (drives the sticky scratch-failure flag, asserts `TMPDIR` is still the container leaf under a world-usable base).
2. **`sandbox/session_tmp.go` — base validation and no fallback.** Both halves fixed: `validWorldTempBase` now requires the full `/tmp` contract (world write **and** world execute **and** sticky) so a 1776-style writable-but-unreachable base is refused; `NewSessionTmp` now attempts every candidate in turn and joins the failures only when none serves. New coverage: `TestSessionTmpFallsBackToALaterBase`, plus a non-traversable-base case in `TestSessionTmpRequiresWorldUsableBase`. A *valid* base whose `MkdirTemp` still fails is not constructible portably (a 1777/sticky base grants the creation it advertises), so that sub-case is covered by the all-candidates-fail case; the test says so.
3. **Windows lost the `TMPDIR` override.** Accepted. `sandbox.SessionTmpSupported` now records where a container can exist (POSIX only); `execenv.tmpDirNamesScratch` **and** the capability preamble's gate both consult it, so Windows keeps the session scratch and the two renderers cannot disagree about a session. The decision record and the two `TMPDIR` summaries state the platform limit. (I did not add a Windows container: the container's contract is POSIX mode bits and a child becoming another uid, neither of which exists there.)
4. **`TestSessionTmpSweepReportsUnremovableContainer` under root.** Correct — root removes a 0500 directory with `CAP_DAC_OVERRIDE`, so both assertions meant something other than they said. It now skips under root, matching `scratch_retention_test.go`'s guard for the same reason.

**Low**

5. **Container-shape tests depended on the machine's `/tmp`.** Correct per `AGENTS.md` ("Do not make `make test` … depend on … ambient developer machine state"). Added `sandbox.SetWorldTempBasesForTesting`; the `execenv` container tests now own a `1777`-sticky base in `t.TempDir()` and skip on platforms with no container, and the fuzz fixture points provisioning at a base inside its own root. The provisioning-failure test drives failure with an empty base set, which also let me **delete** the package-level `newSessionTmp` seam I had added — one less production seam.
6. **Sweep tests now reach the host's shared temps.** Correct, and worse than an ambient dependency: a stale container this process cannot remove would make the sweep return an error and turn `err == nil` assertions into flakes. `scratchRetentionBase`, `TestSweepCrashedSessionScratchReportsUnusableBase` and the two `agent`-package sweep tests now confine the world bases to a fixture.

**Not changed:** the design itself (the shape-dependent rule, the container shape, the amended decision record) — one reviewer found no issues with it and the other findings were all implementation/test gaps around it.

Round gates (local, on the pushed head): `go build ./...`, `make lint`, `make vet`, `GOOS=windows go vet`, `go test -count=1 ./agent/execenv/... ./agent/sandbox/...`, `go test -count=1 ./agent/`, `go test -race` on the two changed packages, and the `evenerfuzz`-tagged container/sweep/process-runtime targets — all green, plus the CI checks on this head.

obra:
## Review round 2 — six fixed, one rejected with evidence

Thanks; the detach finding is the one that mattered most and I had it wrong. Pushed as `42d8d90c42` (one push → one CI run → one review for the new head).

**M2 — detached processes lost `TMPDIR` after close. Accepted, and it changes behavior.** Correct, and it was a regression I introduced: the old `TMPDIR` was the private scratch, which a close *retains*, so a detached command kept a valid temp; my close removed the container out from under it. A close now retains the container exactly as it retains the scratch — lease released, directory kept — and the 24h crashed-scratch sweep reclaims it, which is the reaper requirement 5 names. `Remove` survives only for a mint being **discarded** (a failed launch, where nothing can reference it). `TestCommandEnvironment_UnsandboxedTmpContainerRetainedAtClose` now proves the whole path: retained at close → lease released → reclaimed by the sweep once aged.

**M3 — world bases on Windows.** Accepted: the bases are appended only where `sandbox.SessionTmpSupported` is true. (In practice they were already filtered by the existence check, but scanning names that cannot exist on the platform is exactly the kind of thing that should be impossible by construction.)

**M4 — untagged POSIX test file.** Accepted: `session_tmp_test.go` is now `//go:build unix`.

**M5 — `cmd/evener` reclaim tests walk real `/tmp`.** Accepted, and I missed it: `confineScratchWorldBases` is now called by all five. You were right that the guard went into `agent/sandbox` and `agent/` but not here.

**L1 — ambient dependency in the default test.** Accepted: the live check is now the explicit opt-in `EVENER_TMPDIR_PRIVDROP_E2E=1` (`TestSessionTmpPrivilegeDropE2E`), documented beside the other live opt-ins in `docs/developing-evener/testing.md`. Its default coverage is not lost: the container `0711` / leaf `1777`+sticky bit assertions — which *are* the mechanism the kernel uses for a foreign user — run unconditionally in `TestSessionTmpContainerShape` and in `agent/execenv`'s container test.

**L2 — `TMPDIR` confidentiality.** Accepted as a disclosure, not as a mode change. The leaf is world-writable *and* world-readable by design (a `1777` leaf is what makes it a temp an arbitrary uid can use, exactly like `/tmp`; omitting `o+r` breaks listing, which is a real behavior change and violates the settled "leaf `1777` sticky"). It is now stated normatively: the amended Environment section says the container forfeits the temp area's privacy, that other local users can read a session's temp files there, and that secrets belong in `EVENER_SCRATCH_DIR` — and it says *why* the trade is inherent rather than an oversight. The `TMPDIR` row in `environment.md` and the `envvars` summary carry the same warning. The record's Lifecycle section now states the container's retain/reclaim lifecycle.

**M1 — `DisposeUnsandboxedScratch` leaves the container behind. Rejected; documented instead.** The observation is right (it does not touch the container) but the harm is not. Its only production caller is `adoptResumedRootScratch` (`agent/session_scratch_retention.go:914`), and it operates on the environment the session **keeps using** after the adopt — the container is not part of the allocation being replaced (nothing adopts a container), so this env's already-spawned children must keep the same `TMPDIR`. The lease being held is therefore correct for a live directory: releasing it would let the sweep reclaim a directory in use, and removing it — the suggested fix — would strand those children on a path that no longer exists. Its lease is released at that environment's own close (`RetainSessionScratch`, covered by the close test). I recorded that reasoning in the function's doc comment so the next reader does not have to re-derive it. If you disagree, the counter-evidence I would want is a caller that discards the environment rather than continuing on it.

Round-2 gates (local, on the pushed head): `go build ./...`, `make lint`, `make vet`, `GOOS=windows go vet` (now including `cmd/evener`), `go test ./agent/execenv/... ./agent/sandbox/... ./cmd/evener/`, `go test ./agent/`, `-race` on the two changed packages, the opt-in privilege-drop check, and the `evenerfuzz`-tagged targets.

obra:
## Review round 3 — five fixed, two answered with evidence

Thanks; the ownership gap in the reclaim walk is a real one and the workspace/base asymmetry was too. Pushed as `2a581119cb` (one push → one CI run → one review for the new head).

**M1 — reclaim deletes by prefix/age in world-writable `/tmp` without ownership verification. Accepted.** Correct, and it is worse than a style point: the sweep matches on prefix and age, then *creates our lease inside the candidate* before removing it recursively, so a directory another local user planted under our prefix would be entered and emptied. It now verifies the candidate is owned by this process before touching it, and re-verifies under the held lease (the acquisition itself writes a file there). The regression test plants a stale, world-writable, **foreign-owned** `evener-sandbox-*` directory in a fixture base and asserts the payload survives *and* that no lease file was created in it at all. I mutation-checked it: with the guard removed the test fails with `unlinkat …/payload: operation not permitted`, and passes with it restored.

**M2 — containers under a base the workspace contains are never reclaimed. Accepted; the example needed correcting, the gap did not.** `/tmp/proj` does *not* drop `/tmp` (`pathWithin` treats only a base *inside* the workspace as contained, and `/tmp` is above `/tmp/proj`) — but a workspace that **contains** the base, or *is* the base, does drop it, and you are right that `NewSessionTmp` mints there regardless. The fix is the asymmetry itself: the workspace filter now governs the **scratch** bases only, and the container bases are admitted on the `/tmp` contract alone (`validWorldTempBase`). `TestSessionTmpContainerReapedWhenWorkspaceContainsItsBase` pins it with a workspace *above* the injected base.

**M4 — leak assertion only checked `private`. Accepted.** `TestSessionTmpRequiresWorldUsableBase` now asserts every candidate base in each case is empty, not just the first.

**L1 — build-tag mismatch. Accepted.** The test file's tag now matches the implementation's explicit platform list (`darwin || dragonfly || …`) rather than `unix`: the implementation files are the ones that must agree with `session_scratch_lock_unix.go`, so the test moves.

**L2 — preamble snapshots pinned the POSIX label. Accepted.** The four unconfined snapshots now derive the scratch label from `sandbox.SessionTmpSupported` (a small `scratchLineForTest` helper) instead of hardcoding `Scratch ($EVENER_SCRATCH_DIR): …`, so they stay true on Windows.

**M3 — provisioning failure leaves the ambient `TMPDIR` inherited. Not changed; here is the reasoning.**
Three things point the same way:
1. There is no other validated fallback to select — reaching that branch means *neither* `/tmp` nor `/var/tmp` is a world-usable sticky directory on the host, so "select another world-usable base" has nothing to select.
2. Removing the ambient value is not obviously better on the platform the design is most worried about. With `TMPDIR` unset, a child's default on macOS is still the per-user `0700` directory under `/var/folders/...` — the exact path the design says breaks the field case — so unsetting does not reliably produce a usable temp, it only discards a choice the operator may have made deliberately (`AGENTS.md`-style least surprise, and R6 in the spec: no surprise relocation of a user's own temp beyond what a requirement demands).
3. The failure is sticky and happens once per env, so this is a one-shot degraded state on a host that cannot serve a world-usable temp at all.
If you want the strict behavior anyway, say so and I will add it — it needs a small new "unset this variable" capability in the unsandboxed env overlay (today's overlay can only add or override), which is why I did not fold it into this round.

**M5 — confined shape has no `commandEnvironment` assertion. Already covered; pointing at it rather than duplicating.**
`agent/execenv/sandbox_lifecycle_test.go`, `TestEnableSandboxWriteBlockedOffConfinesFileTools` (the write-blocked off policy, no wrapper) asserts exactly the property, observably and at the spawn-env level:
```go
shell := envToMap(env.commandEnvironment(nil))
if shell["EVENER_SCRATCH_DIR"] != scratch || shell["TMPDIR"] != scratch { … }
```
A regression in `tmpDirNamesScratch`/`overlaySessionEnv` for that shape turns that test red — which is the same coverage you were asking for, in the file the design brief named for it. Tell me if you would still rather see it restated inside the container test and I will add a duplicate.

Round-3 gates (local, on the pushed head): `go build ./...`, `make lint`, `make vet`, `GOOS=windows go vet` (`agent/…`, `envvars/…`, `cmd/evener/…`), `go test ./agent/execenv/... ./agent/sandbox/... ./cmd/evener/`, `go test ./agent/`, `-race` on the changed packages, the opt-in privilege-drop check (with its mutation evidence above), and the `evenerfuzz`-tagged targets.

obra:
## Review round 4 — both fixed

Both findings are correct, and the first is the kind that matters: my own tests were mutating the machine. Pushed as `8685eea105`.

**F1 — default test walks (and deletes from) the real machine temp/cache.** Correct. `TestCommandEnvironment_UnsandboxedTmpContainerRetainedAtClose` confined the *world* base but not the **scratch allocation** bases, which come from `TMPDIR` and the user cache dir — so every default run of that test had the sweep read the machine's real `/tmp` and cache and delete any eligible aged `evener-sandbox-*` directory there (including retained handoff scratch). It now redirects `TMPDIR` and the user cache dir (`HOME` + `XDG_CACHE_HOME`, mirroring `cmd/evener`'s reclaim tests) into the test before minting, alongside the world base it already confined. The newly added `redirectUserCacheDirForTest` helper says why it exists.

**F2 — the opt-in check's comment over-claimed.** Correct: it overrode only `worldTempBases`, so its sweep still walked the real scratch/cache bases. It now calls `isolateScratchBases(t)`, and the comment describes the base set it actually has (and why the fixture root still lives under `/tmp`: an arbitrary uid has to be able to reach it).

Round-4 gates (local, on the pushed head): `go build ./...`, `make lint`, `make vet`, `GOOS=windows go vet`, `go test ./agent/execenv/... ./agent/sandbox/... ./cmd/evener/`, `go test ./agent/`, and the opt-in privilege-drop check.

PR #2089, opened 2026-09-20
Exclude managed worktree lanes from recents; contain the directory picker sidebar

obra:
## Summary

Two fixes for the directory browser's Recent sidebar, mockup-reviewed with Jesse before implementation, plus two review-fix rounds:

1. **Session machinery is no longer "recent projects" (hub).** `evener/projects/recent` (`PastIndex.RecentProjectDirs`) now excludes machinery: an evener-managed worktree lane (`WorktreeManaged` — delegate isolation lanes, `manage_worktree` task lanes, `hub-intent-row-…` scratch lanes) contributes the project it was entered from (`WorktreeRestoreRoot`, following `EffectiveWorkingDir`, spec §7) when known and is skipped when not; a delegate session (`IsSubagent`) is skipped outright — before the managed-lane remap — since its dir is either the parent's project (already surfaced by the parent's own row) or an isolation lane. Real project dirs still come through, and a path-entered unmanaged worktree still counts raw (it is the user's own checkout; that one divergence from §7 is documented at the function). Web picker, TUI spawn dropdown, and mobile all inherit this from the one RPC.

2. **The sidebar can't overflow the dialog (webui).** Recent rows laid out as column flex with `align-items: flex-start`, so a long path or unbreakable basename sized itself to fit-content and escaped the 180px sidebar, running across the browse column under the Path input. Rows now stretch, clip, and ellipsize inside the sidebar; the full path rides on the row's `title` for hover; and the picker's body caps the browser row so each column (recents sidebar and folder list) scrolls inside the dialog — a full 15-item recents list (~930px) no longer makes the whole body scroll. The footer destination's path keeps its wrapping.

## Testing

- `TestPastIndex_RecentProjectDirs_ExcludesSessionMachinery` (hubcore): rooted managed lane surfaces its restore root first by recency; unrooted managed lane, `IsSubagent` delegate lane, and a both-flags delegate lane are dropped; path-entered worktree and plain session stay — each shape watched RED before its fix, GREEN after.
- `TestHubRPCProjectsRecentReturnsMostRecentDirs` (app_rpc): a managed-lane session seeded as the newest must not surface or consume one of the 15 slots — verified to fail with the filter locally reverted, then restored.
- `directorypicker.test.tsx`: recent rows carry `title={path}` for hover (RED then GREEN).
- `directorypicker.overflow.test.ts`: CSS-contract pins for the row containment and the bounded-body column-scroll model per the `welcome.overflow.test.tsx` idiom — each watched fail with the pinned rule removed, then restored.
- Gates, all green: `gofmt -l` on changed Go files, `go vet` on both touched packages, full `go test ./cmd/evener-hub/internal/hubcore/`, `go test ./cmd/evener-hub -run 'RecentProject|ProjectsRecent'`, `make test-web` (typecheck + unit + Biome), and `make test-web-browser` (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard, retirementguard, skillguard).

## Review

- The simplify-code four-reviewer pass landed in 63c68ee2d0 with its deliberate skips documented in the commit message.
- roborev round 1: the restore-root gap and the delegate-lane leak were validated against real session metas on disk and fixed in 2712f226b7.
- roborev round 2: the subagent/worktree precedence (low) and the unbounded scroll area (medium — my "near-unreachable" triage was wrong; a full recents list is ~930px) are fixed in 1a989c7283. The CSS-text-pinning test dispute is flagged to Jesse for an explicit call rather than silently kept or dropped.

## Notes

- Follow-up (out of scope, flagged by the simplify-code review): `RecentModels` has the same machinery-noise shape — a delegate lane's (provider, model) pair can surface in the model picker's Recent group.


obra:
roborev triage — all three mediums reviewed against the code and this machine's real session store before acting:

**Managed lanes lose their restore-root project — fixed in 2712f226b7.** `RecentProjectDirs` now follows `EffectiveWorkingDir` (spec §7) for the managed case: `WorktreeRestoreRoot` becomes the recents candidate when known (subject to the existing dedupe and existence filters), and the lane is skipped when it is not. A pre-merge simplify-code altitude review had flagged the same seam; the wholesale `EffectiveWorkingDir` swap it and this review both considered was rejected because it also remaps path-entered worktrees, which the pinned contract deliberately keeps raw — a user's own checkout is a spawn destination. That remaining divergence is documented at the function.

**Delegate-isolated lanes still leak — validated on-disk, fixed in the same commit.** Real delegate metas on this machine confirm the leak shape: `is_subagent: true`, a lane `working_dir`, and no `worktree_managed`/`worktree_path` at all (e.g. `sessions/034RMxavb5TWcNqlwZ1zqn.meta.json` and four siblings). `RecentProjectDirs` now skips `IsSubagent` sessions: their dir is either the parent's project — which the parent session's own row already surfaces — or an isolation lane. The hubcore test pins all the shapes (rooted lane surfaces its restore root first by recency; unrooted managed lane and delegate lane are dropped; path-entered worktree and plain session stay), watched RED before the fix and GREEN after.

**`overflow-y: auto` without a constrained height — real edge, not fixed, reasoning below.** The 15-item cap bounds the sidebar to roughly 550px of single-line rows; once the viewport is too short for the 340px-min browse column plus that (~730px), the panel body scrolls internally, so nothing clips or becomes unreachable — the sidebar just scrolls with the body rather than independently in that window. Capping the sidebar's height for a window the 15-item cap already makes near-unreachable read as fragile YAGNI; happy to take it if you disagree.

**CSS-grep overflow test — false positive against this repo's own precedents.** `welcome.overflow.test.tsx` pins issue #197's wrap fix and `toolRowGrammar.test.tsx` asserts ellipsis rules exactly this way; jsdom cannot observe scroll containers or CSS geometry, which is why the contract-test idiom exists for overflow fixes. The test strips comments first (the `stripCssComments` precedent), so it cannot be satisfied by prose.

obra:
Second-round follow-ups landed in 1a989c7283:

- **Subagent/worktree precedence (low):** fixed — `IsSubagent` now skips before the managed-lane remap, so a delegate session carrying both flags is dropped entirely. Pinned with a both-flags fixture whose restore root is contributed by no other session, since a shared-root fixture is invisible (the remap dedupes against the root session's own row). Watched RED, then GREEN.
- **Unbounded scroll area (medium):** fixed — my earlier "near-unreachable" triage was wrong: a full 15-item recents list is ~930px of single-line rows, taller than the dialog's body budget on ordinary viewports, so the whole body scrolled while the sidebar's own scrollbar never engaged. The picker's body now caps the browser row (flex column + overflow:hidden, 340px floor preserved) and each column scrolls inside it — the model the mockup promised. The overflow contract pin extends to the scroll model; make test-web plus the full real-Chrome guard set (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard, retirementguard, skillguard) all pass locally.
- **CSS-text-pinning test (medium):** holding for Jesse's call — roborev asks for an explicit exception per AGENTS.md; the repo's own precedents (welcome.overflow.test.tsx pinning issue #197's wrap fix, toolRowGrammar.test.tsx asserting ellipsis rules) do exactly this kind of pin, which is why I added it. Flagging to Jesse rather than silently keeping or dropping it.

obra:
Round-3 triage — two of three mediums fixed in 54aa0f2c30 (now on head 757bec4c26 with the latest main merged):

- **`.body` overflow:hidden clips the 340px row floor on short viewports (medium):** valid — the rule's own comment claimed the short-viewport fallback still scrolled, and it did not: hidden overrode the chassis's overflow:auto and clipped. `overflow: auto` restores the fallback: at the panel's cap the row fills the body exactly (no scrollbar; the columns scroll inside), and only the short-viewport window scrolls the body. The contract pin now asserts auto.
- **Mobile recents keeps `flex: none` and forces its full content height, collapsing the browse pane (medium):** valid — and a regression my round-2 overflow change introduced (previously the whole body scrolled on phones; the new clipping made the collapsed browse pane unreachable). The phone-layout override now lets the recents block shrink below content (`flex: 0 1 auto; min-height: 0`) and scroll inside the dialog. The contract pin asserts the phone override too.
- **A worktree switch re-persists the previous lane as `WorktreeRestoreRoot` (medium):** false positive with code evidence — `recordAbandonedEnvironmentLocked`'s own comment in `agent/session_lifecycle.go` states "a switch does not re-park" the restore env, so the restore root stays the env saved at the FIRST enter (the launch project); `session_worktree_resume_test.go` pins `WorktreeRestoreRoot == mainRoot` across resume/reentry. A root session whose launch env is itself a lane would be the only path to a lane restore root, and the hub's launch validation refuses `--dir` into a worktree (the reason `EffectiveWorkingDir` exists).

Both fixes were TDD'd (pin RED before each CSS change), and the full local gate set passes: make test-web plus make test-web-browser with all seven guards (layoutguard, overflowguard, shellguard, spawnguard, transcriptscrollguard, retirementguard, skillguard) green. The CSS-text-pinning dispute remains parked for Jesse's explicit call alongside the merge authorization.

PR #2093, opened 2026-09-20
fix(agent): release a session's own stale worktree marker on prune (#894)

### LINKED ISSUE #894: A session's own lock marker on a non-current managed worktree lane is never cleared (stranded lanes)

obra:
Fixes #894

## Root cause

A managed worktree lane locked with a session's own `evener:<sid>` marker while that session does **not** occupy the lane is crash residue — what a process death inside the lane, or a resume whose re-entry was refused before the lane was recorded as current, leaves behind.

`remove` and `switch` already treat that residue as "unlocked-for-us" (spec §5: *"A worktree found locked with the session's own marker while the session is not occupying it ... is the session's to release"*). `prune` did not: `Decide(EvPruneCandidate, OwnSession)` returns `ActSkip` (`agent/internal/worktree/lockstate.go:483`), so a lane carrying the pruning session's own stale marker was the one lane prune could never reclaim, and the documented recovery (`git worktree unlock`, then prune) could not be done through the tool. The automatic residue sweep (`agent/session_worktree_sweep.go:148`) likewise only ever touched unlocked lanes.

## Change

`worktreePruneSweep1` (`agent/session_tools_worktree.go`) now treats that marker as no obstacle to **eligibility**, and releases it through the same `EvLeave` rule `close` uses (`releaseOwnWorktreeLock`) **only on the path that actually collects the lane** — immediately before `collectLane`, after every skip rung. A one-line helper (`occupiesManagedLane`) keeps the session's **current** lane skipping outright.

The release never acts on the pass's own `worktree list` snapshot, which every earlier lane's git calls age. It re-reads the lane's lock from a **fresh listing** immediately before touching it (`confirmStrandedOwnMarker`), so a marker someone released since the snapshot (the documented manual `git worktree unlock` recovery, or a second process on the same session id) and another session then took is left alone rather than unlocked with its lane collected. And if the removal then **fails with the worktree still present**, the marker is put back (`restoreStrandedOwnMarker`), so a lane this pass could not collect is exactly as protected as it was.

Releasing during evaluation would be worse than the strand it fixes: a lane this pass then skips (dirty, unmerged, in-grace, under live work, or not a delegate lane) would be left unlocked, and another session's prune cannot see this session's live work under it — the lock is the only thing stopping it from removing the lane out from under the session. A skipped lane therefore keeps its lock. Only the session's own marker is ever released; a foreign or delegate marker stays the skip it was.

Because the automatic residue sweep is delegate-only, it never reaches the collect path for a plain session lane, so it never touches this marker; the only operation that releases it is a prune the session itself runs. No new lock or serialization was needed for that (an earlier revision of this PR added a lane-gate mutex; the review correctly flagged that it made close's sweep join unbounded, so it was removed rather than patched).

Docs updated: the normative lock-state table row and the `prune` sweep-1 bullet (`docs/superpowers/specs/2026-07-02-native-worktree-tools-design.md`), and the crash-recovery and two-process notes in `docs/developing-evener/worktrees.md`.

## Test evidence

`agent/session_worktree_stranded_own_marker_test.go` (real-git, via the `wtRepo` harness):

- `TestWorktreePrune_ReclaimsStrandedOwnMarker` — prune reclaims a lane carrying the session's own marker while it occupies another lane, and leaves its own current lane locked and present.
- `TestWorktreePrune_StrandedOwnMarkerOnASkippedLaneKeepsItsLock` — the safety half, behaviorally: a stranded own-marker lane that is dirty, and one with stubbed live work, are skipped **and keep their lock**.
- `TestWorktreePrune_FailedCollectionKeepsTheStrandedMarker` — a collection that fails with the worktree still in place keeps the lock (re-locked).
- `TestWorktreePrune_AbandonsTheReleaseWhenTheMarkerChanged` — the release is abandoned when a fresh listing shows the marker is no longer this session's own; the other session's marker and lane are untouched.
- `TestWorktreePrune_LeavesAnotherSessionsMarkerAlone` — another session's marker is still skipped (the bound on the fix).
- `TestLaneResidueSweep_LeavesOwnMarkerAlone` — the automatic residue sweep leaves an own-marker lane alone (why no serialization is needed).

Before/after (exact output on this head):

- With base production code (`git checkout origin/main -- agent/session_tools_worktree.go`): `TestWorktreePrune_ReclaimsStrandedOwnMarker` FAILs — `stranded lane skipped by prune with reason locked (evener:034SL...)`, `removed:[]`.
- With the eager-unlock shape restored (release moved before the eligibility chain): `TestWorktreePrune_StrandedOwnMarkerOnASkippedLaneKeepsItsLock` FAILs on both rungs — `prune released the marker of a lane it skipped (dirty: ?? uncommitted.txt)` and `(live work: live-real)`.
- On the previous revision of this branch (`git checkout HEAD -- agent/session_tools_worktree.go` at a6df3ca): `TestWorktreePrune_FailedCollectionKeepsTheStrandedMarker` FAILs — `a failed collection left the lane unlocked; another session's prune could remove it` — and `TestWorktreePrune_AbandonsTheReleaseWhenTheMarkerChanged` FAILs — `prune collected a lane another session had taken` (`worktree_removed:true branch_removed:true`).

Verification on this head: `go test ./agent/... -count=1` green; `go test ./agent/ -run 'Worktree|Residue|Lane|Stranded|UnlockOwn|Close|Dispose|Registry|Tool'` ok (64s); `golangci-lint run ./agent/...` → 0 issues; `go vet ./agent/...` clean; `gofmt -l` clean on all changed files; the package audits `TestNoBareWallClockDeadlineInAgentTests` and `TestEverySamplingRelevantMutexIsClassified` pass.

## Not fixed here

A lane stranded by a session that dies and is **never** resumed or closed, where the clearing agent is a *different* session, still needs a session-liveness signal (pid file/lease/heartbeat keyed by session id) that does not exist. `prune` cannot attribute another session's marker to a dead owner without guessing, and taking it on a guess would break a live session's occupancy. See the issue's own triage comment: `needs-human`/`needs-decision`. This PR deliberately does not widen prune's authority to other sessions' lanes.


obra:
Parking this PR rather than starting a sixth review round.

The change is right in shape — all three reviewers agree, one found no issues, and it implements a rule that genuinely pre-exists in the 2026-07-02 worktree design spec (`:557-561`), which named `remove` and `switch` as treating the session's own stale marker as unlocked-for-us and omitted `prune`. That omission is the real #894 defect.

What does not converge is the safety of the release sequence. Five rounds: Medium+Low → **High**(unbounded shutdown from the serialization gate)+Medium → 2 Medium → **High**(non-atomic confirm/unlock)+2 Low. Each round closed a real window and exposed the next, because the operation is a check-then-act sequence assembled from independent `git worktree` calls with no mutual exclusion — and naive serialization made it worse, not better.

The outstanding High, both Lows, the five-round history, and the structural options are filed as a follow-up. It also notes that the two Lows are individually safe to land (spec/core/test drift, and widening occupancy detection to mirror `worktreeRemove`), and that keeping prune conservative — which is what the spec already prescribes for `remove` and `switch` — is probably the cheapest safe end state for a P2.

Leaving this open and parked with the branch preserved, matching PRs #2019, #2013, #2034, #1813 and #2085. Never merged.

PR #2098, opened 2026-09-21
hub+daemon: stop an archived session's resident daemon promptly

obra:
## What

Archiving a local session now stops its resident daemon promptly: the archive decision retargets the daemon's automatic idle-retirement deadline to **1 minute** (an idle daemon stops within a minute; a busy one retires once its work settles). Unarchiving restores the hub's configured `daemon_idle_timeout` (0 = disabled).

No spawn-time changes: reading an archived session never respawns a daemon (thread/read serves from the on-disk past index), and deliberately sending to an archived session resumes with the normal configured deadline.

## How

- `agent.RetirementController.Retarget(timeout)` — changes the idle deadline at runtime, preserving the settled instant: shortening fires proportionally sooner (immediately when below already-elapsed idle time), lengthening re-arms later, zero disables. The automatic claim path re-proves the deadline at the fence, so a stale timer tick after a lengthen can never claim early.
- New ScopeDaemon RPC `evener/daemon/idle-timeout/set`, identity-fenced exactly like `evener/daemon/retire` (shared `requireExactOwnership` fence), classified `control` (no admission lease, no activity mark), rejecting negative and duration-overflowing values as invalid params.
- The hub's `evener/archive/set` nudges the session's resident daemon after persisting the decision: best-effort and silent (no resident daemon / older-protocol peer / refused or wedged daemon leaves the decision durable), bounded at 5s, reading the roster's published snapshot so archiving never probes. The decision and its nudge hold the session's ResumeLocks alias, so concurrent toggles land in persist order.

## Why this shape

Archived sessions' idle daemons already self-retired — after the configured 1h. This only shortens the existing, tested retirement window; it introduces no new stopping semantics, and the deliberate #1248 refusal (`archivedRetireRefusal`) stays untouched: manual retire remains an active-session action.

## Review evidence

- simplify-code pass: four angle reviewers; four findings applied (single-sourced ownership fence, 5s nudge bound, shared lifecycle-race normalization, shared negative-timeout sentinel).
- Adversarial competitive review (two independent reviewers): both findings fixed red-first — the cross-connection archive/unarchive nudge-ordering race (alias lock), and the `timeoutMillis` overflow that wrapped to zero (silently disabling retirement). Non-issues were kept out of both scored lists.

## Test evidence (on the merged tree, base e58d72fea7)

- `make lint` exit 0, `make vet` exit 0, `make test` exit 0 (root 166.9s, agent 24.0s, web 596.5s).
- Feature tests: controller Retarget suite (ack-clock harness, incl. stale-tick-after-lengthen), serve-level identity/validation/overflow refusals plus an end-to-end shortened-deadline exit, hub nudge suite (retarget/restore values, exact identity, no-daemon, older-protocol, project-kind, refusing daemon, wedged-daemon bound, concurrent-decision ordering), race detector clean over the archive suite.
- `make test-api-package` exit 0 (TS mirror regenerated: `make generate` no drift after the main merge).


obra:
Review triage for the Combined Review on `e2e97b9` — every finding was validated against the code before acting:

**Protocol version not bumped — by design.** The skip is the specified degradation. The nudge is best-effort and silent by contract: an older peer that cannot answer the method lands in the same bucket as a refused or wedged daemon — the decision stays durable and the daemon keeps its configured deadline, which the resident list already publishes. Bumping the flag-day protocol for one additive method would change routing for every mixed-version fleet (only the exact flag-day protocol establishes ownership), a far larger blast radius than a rare, benign miss. Pinned by `TestArchiveSetSkipsIdleTimeoutForOlderProtocolDaemon`.

**Archived timeout not restored on resume — by design, product direction.** Reading an archived session never respawns a daemon (`thread/read` serves from the on-disk past index); only explicitly sending to it does. Deliberately resumed archived sessions use the normal configured deadline: archiving stops the daemon that was idling, it does not change how a resumed session behaves. Documented in the PR description.

**Short configured timeout lengthened on archive — fixed in bfc8266940.** The archived deadline is now an upper bound: archiving pushes min(configured `daemon_idle_timeout`, 1m), so a sub-minute configuration is never extended. A disabled (zero) configuration still shortens — the explicit archive decision outlives a hub-wide disable. Pinned red-first by `TestArchiveSetNeverLengthensShorterConfiguredDeadline`.

**Lock held across daemon nudge and navigation refresh — fixed in bfc8266940.** The lock now covers exactly the durable write and its nudge; the navigation refresh and attention poke run outside it, so a wedged nudge delays only this session's decisions and fences, never the tree rebuild.

Gates on the new head: `make lint` 0, `make vet` 0, `make test` 0.

obra:
Review triage for the Combined Review on `bfc8266` — every finding was validated against the code before acting:

**Resident daemons whose workspace ref survived thread/clear — by design.** The nudge answers one question: is *the archived session's own* daemon resident? `liveDaemonForSession` matches the daemon's current session identity — after a clear the entry's `SessionID` has advanced to the replacement, and `ThreadID` only covers a boot entry whose session fields never filled in. The proposed workspace-aware lookup would invert the feature's contract: archiving a *predecessor* session would retarget the daemon that now serves the *replacement* — the same stale-decision-moves-new-owner's-deadline hazard the Low finding correctly flags as a bug. The resident list already publishes that daemon under its current session, so the predecessor row shows no daemon and archiving it moves nothing. Pinned by `TestArchiveSetNeverRetargetsClearedSuccessorsDaemon`: the predecessor archive reaches nothing, the replacement archive still retargets with the current identity.

**Swap window in requestIdleTimeoutSet — fixed in fedcd81b.** The pre-Retarget fence and the controller write are not atomic (mutation admission leases fence retirement, not each other), so a thread/clear completing between them re-roots the controller and the stale write moved the replacement's deadline. The setter now revalidates exact ownership after the retarget and, on drift, restores the configured deadline the daemon was spawned with — the root swap never resets it, so the stale write must be undone there — and returns the same conflict the retire path returns after its claim. Red-first pinned by `TestServeIdleTimeoutSetRevertsDeadlineWhenClearSwapsOwnership`, which interleaves a real thread/clear into the window through the same observability-seam pattern the retire race test uses, and proves a current generation can still move the deadline afterwards.

Gates on this change: `make lint` 0, `make vet` 0, `make test` 0 (full chain on the exact pushed tree).

obra:
Review triage for the Combined Review on `fedcd81` — every finding was validated against the code before acting:

**Missing protocol-version bump — by design, unchanged.** This re-raises the first review's protocol finding, and the disposition stands: the silent skip is the specified degradation for an additive method on a mixed-version fleet. The nudge is best-effort and silent by contract — an older v5 peer that cannot answer `evener/daemon/idle-timeout/set` lands in the same bucket as a refused or wedged daemon: the archive decision stays durable and the daemon keeps its configured deadline, which the resident list already publishes. Bumping to v6 would not merely skip the method for older peers — it would fail the exact-version compatibility check for every v5 daemon across every method, a fleet-wide flag day; only the exact flag-day protocol establishes ownership. Pinned by `TestArchiveSetSkipsIdleTimeoutForOlderProtocolDaemon`, and the full rationale is in the earlier triage comment.

**Stale-write revert restores configured instead of the prior effective value — fixed in 9c0f164.** The reviewer's sequence was reproduced red-first, and the fix went one step further than proposed: restoring the captured prior value still has both a compare-then-restore window and an aliasing hole — the Hub's archive decision derives the same min(configured, 1m) for every session, so the stale write and the replacement's own legitimate write usually choose the same deadline, which a value comparison cannot tell apart. `RetargetStamped` now mints a write token under the controller's mutex and returns the deadline the write replaced; on fence drift the setter calls `UndoRetarget(token, prev)`, which restores the pre-write deadline in one atomic critical section only while that token still names the newest write — a later writer's deadline always survives the refusal, aliased or not. Pinned at both layers: agent token-contract tests (mint, supersession, alias, preparing fence) and three serve subtests including the aliased-same-value interleave, all red-first through the `idle_timeout_written` seam.

Gates on this change: `make lint` 0, `make vet` 0, `make test` 0 (full chain on the exact pushed tree).

obra:
Review triage for the Combined Review on `9c0f164` — every finding was validated against the code before acting:

**Stale timeout inherited by replacement session on thread/clear — fixed in 596082b.** The reviewers' split (one deliberate, one not) is resolved as a design decision in line with the PR's documented product line: a fresh session uses the configured deadline, exactly like a resumed one. `AttachRoot` now resets the idle deadline to the configured timeout the controller was constructed with — the value a fresh spawn boots with — and mints that reset as a timeout write, so pending undo tokens are superseded at the swap and a stale setter can never undo past the fresh root's baseline. The disabled (zero) baseline is restored too. Red-first pinned by `TestRetirementAttachRootResetsDeadlineToConfigured` and `TestRetirementAttachRootResetsDisabledDeadline`, plus the flipped serve subtest: a clear completing after a parked stale write supersedes its token, and the refusal leaves the configured reset in place.

**Protocol version not bumped — by design, third re-raise.** The disposition from the first review stands unchanged and is documented in both earlier triage comments: the silent per-method skip is the specified degradation for an additive method on a mixed-version fleet, and a v6 bump would fail the exact-version compatibility check for every v5 daemon across every method — a fleet-wide flag day far larger than this method's rare, benign miss. `TestArchiveSetSkipsIdleTimeoutForOlderProtocolDaemon` pins it.

**UndoRetarget accepts token 0 and unchecked negative to — fixed in 596082b.** Token 0 would otherwise match a fresh controller's zero write count and let a caller overwrite the deadline without a stamped write; both it and negative restore targets are now refused before the mutex. Red-first pinned by `TestRetirementUndoRetargetRejectsUnguardedWrites`.

**Data race in test helper — fixed in 596082b.** The fixture daemon's recording moved behind a locked `idleTimeoutRecorder` whose `recorded()` accessor copies under the mutex; every test read goes through it, so handler writes and concurrent-connection reads cannot race.

**Concurrent stale writes leaving the older stale deadline — closed by the same reset.** With the root swap minting a timeout write, every clear drift advances the write count past both stale tokens, so every pending undo refuses and the fresh root keeps its configured baseline (or a later legitimate writer's value) — the composed-undo sequence can no longer land a stale value. This also matches the finding's own reachability note (not reachable through a single archive toggle). No handler-span serialization is needed on top.

Gates on this change: `make lint` 0, `make vet` 0, `make test` 0 (full chain on the exact pushed tree).

PR #2101, opened 2026-09-21
feat(hub): give entity hover cards the floating-layer treatment and drop their duplicate native titles

obra:
## What

The entity hover card (the floating card on `job_…` / `dlg_…` / `watch_…` references) looked like an oversized tooltip: it borrowed the tooltip's inverted near-black palette. And anywhere a hover card already lived, a native `title` tooltip could fire on top of it.

This PR gives the card the design system's floating-layer treatment and removes the classic tooltips that duplicated it.

## The restyle

- **Bubble** (`widgets/hovercard/hovercard.module.css`): `--surface-1` ground, `--shadow-overlay` ring + soft shadow (no doubled border, per design-system §2), `--radius-pane`, `--font-size-ui`, the shared 120ms overlay fade-scale with a `prefers-reduced-motion` off-ramp. Same family as popover/menu/dialog; the tooltip keeps its mini-palette for one-line labels.
- **Card internals** (`entityref.module.css`): kind label follows the one eyebrow recipe; status reads color-by-meaning (failed = `--danger-ink`, running = `--alive-ink`, completed/stopped stay neutral — the hue law reserves green for active work); meta rows follow the meta-table idiom (ink-mid labels, ink-hi values, mono for machine text); the opacity fades are gone.

## The native-title removals

Where an entity trigger already carries a hover card, the classic tooltip that fired alongside it is gone:

- `jobWatch` list rows: `title={row.id}` on the span wrapping the `EntityRef`.
- Demoted `ToolRow` summary lines: `title={summary}` (its full text is one disclosure away; entity hover cards embedded in the line were competing with an OS-styled tooltip).
- The row-level `title={detail}`: retires `descriptor.detail` and shell's `detail()`. The exit code's one home is the captured output's own trailing `[exit N]` footer inside the expandable body — already real, keyboard-reachable text.

The task list's latest-note `title` stays (no hover card lives there; it was reviewed and kept).

## Testing

Everything test-first:

- New off-disk style contract tests (`hovercard.style.test.ts`, `entityref.style.test.ts`) pin the bubble and card internals to the design tokens, sharing one brace-walk helper (`src/styles/cssBlock.ts`).
- `token-contract.test.ts` gains the exact-path semantic exception the status hues require (scoped-path pinned), plus a central rule: `--tooltip-*` may appear only in `tokens.css` and the tooltip's own stylesheet.
- Title-removal sites reworked test-first; the retired `detail()` tests were removed with the feature (the `failed()` coverage they shared is preserved), and the exit code's reachability stays covered by the raw-output footer tests.
- The `/dev/widgets` HoverCard gallery section now renders the real `EntityRef` cards over the shared test fixtures instead of a hand-rolled sample.

Gates (rebased tree, tip `34b99d3c05`): `make test-web` (web-typecheck, web-test, web-lint) and `make test-web-browser` (all seven guards) pass. Rendered computed styles verified in real Chrome in both themes (light surface `#FCFBF8` + `#B7B6AC` ring; dark `#232320` + `#51514A` ring; neutral "completed" = `--ink-mid`; "running" = `--alive-ink`).

## Review notes

- `descriptor.detail` was only ever implemented by shell and only consumed as the row's hover title; the removal deletes the whole mechanism rather than leaving an orphaned contract.
- The delegate fixture moved from `EntityRef.test.tsx` into `entityView.testFixture.ts` so the gallery and the tests share one shape (simplify pass finding).


obra:
Disposition of the roborev combined review (head `34b99d3`), on the new head `6051223a46`:

**Low — order-sensitive assertion (entityref.style.test.ts:32): fixed.** The exhaustive hue-state comparison now sorts before comparing (commit `34dbae5777`); the two `.status[data-state]` rules carry no file-order meaning.

**Medium — non-expandable rows "unable to recover truncated text": premise refuted, invariant pinned, no behavior change.** The head/tail clamp is a JS split that renders *both* halves as separate flex spans; the elision is CSS (`overflow: hidden; text-overflow: ellipsis` on `.clampedHead`/`.clampedTail`). No character is dropped from the DOM — screen-reader and copy access keep the complete summary. The pre-existing expandable-row test already pinned `head + tail === summary`; commit `34dbae5777` adds the same pin for the non-expandable case (a job_watch-style terminal catch-up summary with an intent line) and corrects a comment in the same file that wrongly said the clamp "drops text".

On the suggested fixes:
- *Conditional title for non-expandable rows* reintroduces exactly the collision this PR removes: these rows' summaries embed the entity ids (`job_…`, `watch_…`) whose hover cards are the row's one floating surface, and a native title fires both at once.
- *An accessible summary disclosure* contradicts the established grammar these rows exist under (`hasBody` exists precisely so a summary-only rendering offers no disclosure that would open to nothing).
- *Skipping the head/tail cut when not expandable* degrades to end-truncation, losing the command/watch ending the middle cut exists to preserve.

The residual, stated plainly: a sighted reader on a narrow pane can visually lose the middle of a long summary-only row, with no hover affordance to reveal it. That row shape is rare (job_watch clears and terminal catch-ups with a stated intent), the text remains fully in the DOM, and the prior native title was itself mouse-only. Happy to revisit if the narrow-pane case matters in practice — but a title would fire over the entity hover cards this PR is about.

obra:
Disposition of the combined review on head `6051223`, addressed on the new head `41bd4f42`:

**Medium — typed exit code with footer-less output leaves the number nowhere: valid, fixed.** The wire carries `ItemModel.exitCode` and the output footer independently, and the pre-PR hover title was the number's only home in exactly that gap (the old `detail()` test fixtures were this shape). The shell body now synthesizes the daemon's own `[exit N]` footer when the typed exit code exists and the output carries no trailer of either shape — display-only, like the truncated-tail notice, so Copy output keeps the raw evidence. Pinned by three tests: the synthesis, a no-duplication control when the output already carries its footer, and copy-purity (commit `1cd60ddf0d`).

**Low — watching watches stay neutral: intentional, unchanged.** The hue law reserves green for active agent work; a standing watch is a condition, not work in flight, and the watch list's `watching` chips are deliberately neutral on the same reasoning. This matches the PR's stated design ("watch states stay neutral").

**Low — narrow-pane visual truncation of summary-only rows: documented, unchanged.** The test comment and the earlier disposition comment already record the tradeoff; DOM completeness keeps screen-reader and copy access whole, and the offered alternative (a title on entity-free rows) reintroduces the OS tooltip that fires over these rows' entity hover cards.

obra:
Disposition of the combined review on head `41bd4f4`, addressed on the new head `542b38a6`:

**Medium — exit-footer synthesis mis-handles the typed code: valid, fixed** (commit `542b38a6`). Verified against the authority the review cited: job-control.md:1012 defines `-1` as the sentinel of a signalled job ("not a shell code"), and `formatShellResult` deliberately omits it from the footer (agent/session_tools_shell.go:598-620) — so the synthesis now never fabricates a line for a negative code. The suppression gate is also shape-aware now: a real trailer is the daemon's bracketed final segment or the buffered line as the output's final non-empty line, never a bare `exit_code=` token echoed mid-stream by the command's own stdout, and a trailer only suppresses the synthesis when its code *agrees* with the typed value — a disagreeing trailer leaves the authoritative typed footer standing beside the verbatim raw text. Pinned by three new tests (sentinel, mid-stream echo, trailer agreement/disagreement).

**Low — TOOLTIP_PALETTE_RE scans comments: fixed.** The law strips block comments before matching (the same shape requireclass-contract.test.ts removes), with a poison test proving prose naming the palette cannot fail its own gate.

**Low — gallery module-scope throw crashes the whole gallery import: kept deliberately.** The fixture is a constant resolved from the shared `entityView.testFixture.ts` consumed by eight test suites; if it ever breaks, the test suites fail loudly at the same moment. A per-section fallback would turn a real break into a silently empty section in dev tooling, which is the worse failure mode for a gallery whose purpose is showing the real components.

**Medium — non-expandable clamped rows' hover access (re-raised): holding with evidence, awaiting the human approver's call.** The rejected fallback (a conditional `title`) re-fires an OS-styled tooltip over exactly these rows' entity hover cards — the collision this PR exists to remove; these rows' summaries embed the `job_…`/`watch_…` ids. The alternative (an expansion affordance) contradicts the `hasBody` grammar that makes these rows summary-only in the first place. The tradeoff is documented in the test comment and the two prior disposition comments; if the narrow-pane case matters in practice, that judgment belongs to the approving human rather than the bot, and I'll add the conditional title on request.

PR #2103, opened 2026-09-21
feat(agent): persist pasted image attachments at a durable, readable path

obra:
Pasted images reached the model as inline `ContentImage` parts but existed nowhere on disk: after context folding dropped the image part, neither the model nor a later session could reach the bytes again — the only copy was base64 inside the transcript JSONL (a past session spent its opening turns hunting the state dir for a pasted screenshot and found nothing).

## Change

When a session with a state directory accepts user input carrying image attachments, each image's bytes are written to

```
<stateDir>/sessions/<id>/attachments/<sha256-prefix>-<sanitized-name>.<ext>
```

at turn-build time (`persistInputImages`, `agent/image_persist.go`), and the user message gains a `<system-notification>` content part naming the stored path, so the model can re-read the file later with `read_file` (which routes image bytes back into context). The path is content-addressed: retries rewrite the same file and repeated pastes dedupe.

Hooks sit at every site that builds an image-bearing user turn: `acceptUserInputWithSkillSelection` (direct input, queue drain, drain-as-steer), `consumeSteeringMessage` (steering turns and restart-recovered steering), and `recordClientMutationFailure` (failed client-mutation starts).

Gating follows the tool-result image announcement's philosophy:

- No `StateDir` (library/test shape) means no write and no annotation; the message shape is unchanged.
- A failed write degrades the same way — the image still rides inline, no path is announced — and reports through the general warning channel, firing the Notification hook like every other session file I/O failure (queue snapshot persist, ATIF export, namer log).
- `ImageAttachment.Path` is `json:"-"`: in-session-only metadata, structurally barred from the durable queue snapshots.

Retention is inherited: attachments live exactly as long as the session's transcript directory.

## Verification

- Eight tests in `agent/image_persist_test.go`: direct/queue/steering/failed-start paths, both degradation branches (blocked dir, blocked file), dedup on re-paste, and the sanitize/extension contract. The failed-start hook's test was proven removable — the hook was disabled, the test failed on exactly the annotation, then the hook was restored.
- Full agent suite, server suite, `make lint`, `make vet`, and `go build ./...` are green against the PR base (the branch was rebased onto `main` and every gate re-run on the new base).
- The write-failure channel choice is pinned at mechanism level by the existing `TestSession_NotificationHookRunsOnWarning`; the degradation tests pin the warning naming the failed path on both failure branches.

## Review history

Two four-angle simplify-code passes (reuse / simplification / efficiency / altitude) ran over the branch. Applied findings include `slices.Clone`, prefix-only hex encoding, `json:"-"` on `Path`, hook-site comment trims, shared test helpers replacing duplicated collectors/awaits/scans, and a wait-for-match warning await (a first-match collision two reviewers independently reproduced in world-writable-root sandboxes). Skipped with reasons: `jobsDir` reuse (the hand-join matches the repo's dominant per-session path idiom, and `jobsDir`'s jobs-domain naming plus empty-stateDir temp fallback make it the wrong dependency), test-oracle constants (deliberate independent pin), and a persist+build wrapper (YAGNI at three pinned sites).

## Caveat

Forked sessions reference attachment paths under the original session's directory, so a fork's transcript can name a path whose retention follows the original session. Acceptable for v1; flagged for the record.


obra:
Review findings addressed (head `4fa8d251cc`):

1. **HIGH — `<system-notification>` part leaked into reloaded user/steering bubbles.** The reload projection concatenated every text part via `Message.Text()`. `userFacingText` now concatenates text parts while skipping any part that is exactly one `<system-notification>` block (exact-block skip, so inline mentions stay). 45adc470d5 — regression tests in `internal/apptranscript/user_facing_text_test.go`

2. **MEDIUM — attachment note promised a read a restricted sandbox denies.** `ResolvedPolicy.FileToolCanRead` now answers the policy question (masked paths deny in every mode; `ReadAnywhere` allows the rest; `ReadWorktreeOnly` requires a read root; relative paths fail closed), and `persistInputImages` keeps a stored path unannounced when the session's file tools cannot read it. The bytes are still written for unrestricted readers (a later session, doctor tooling). 4fa8d251cc — tests in `agent/sandbox/resolve_canread_test.go` and `agent/image_persist_test.go`

3. **MEDIUM — relative `--state-dir` resolved inconsistently.** `canonicalStateDir` anchors a relative state dir against the process working directory once, at both construction entries (`NewSession` and `RestoreSessionFromMetaWithConfig`). This also fixes a hard failure the test exposed: a relative restore `StateDir` made the retained-scratch owner comparison mismatch outright. 5c2a4b8bb7

4. **LOW — bmp/tiff attachments stored as misleading `.png`.** `attachmentExtensionForMediaType` now maps `image/bmp` → `.bmp` and `image/tiff`/`image/tif` → `.tif`; unknown types keep the `.png` default. a8d7609bc1

All fixes landed red-green TDD with regression tests. Gates at `4fa8d251cc`: full agent suite, apptranscript, sandbox, and server suites, `make vet`, `make lint`, `go build ./...` — all green.

obra:
Round-2 review findings addressed (head `1d182b895c`):

1. **MEDIUM — fork/edit text leaked the machinery note.** Fixed. `ForkSessionAtUserTurn` now returns the user-facing text (the helper exported as `apptranscript.UserFacingText`), so the prefill handed to the fork/edit surface is the user's own prose. 253d3ecd73

2. **MEDIUM — reload dropped machinery-only steering turns.** Fixed with the reviewer's suggested shape: `UserFacingText` strips machinery parts only when other user-visible content remains; a message that would reduce to empty keeps its full text, so the cancelled-callback-watches steering notice survives reload at parity with the live projection. Image-bearing turns keep their stripped-empty text (the image is the bubble). 253d3ecd73

3. **MEDIUM — `os.WriteFile` followed symlinks / overwrote existing entries.** Fixed. Attachment writes now use an exclusive create (`O_CREAT|O_EXCL`): atomic against the directory entry, and it fails with `EEXIST` without dereferencing whatever the path already holds, so a planted symlink is refused, not written through. An existing entry dedupes only when its bytes match this attachment; a planted symlink or mismatched file is refused and reported as a warning. On the permissions half: every fresh create lands 0600, and an existing entry is announced only after its bytes match — and any inherited file mode sits inside the session's 0700 state directory, which is the actual boundary. 7f321c9c0f, with 1d182b895c replacing the initial platform-split variant after it failed the agent package's production-file invariant (every top-level production file must build on every supported platform; `O_EXCL` alone carries the no-follow property, so the split was unnecessary).

4. **MEDIUM — read_file availability not checked.** Fixed. The note is now gated on `read_file` being in the session's registry (role toolsets are plugin-configurable and can omit the tool). The bytes are still persisted for later readers. 05d0e26633

5. **MEDIUM — unescaped path markup in the machinery block.** Not fixed, with reasoning. The only attacker-controlled component of the announced path is the user's own `--state-dir`; the sanitized filename (`[a-zA-Z0-9-_.]` only) and the system-generated session ID cannot carry markup. A malicious state-dir name injects text into the user's own session's model context with exactly the same powers as typing that text into the session — the same actor, the same authority, no boundary crossed. There is also no escaping scheme for these faux-XML blocks to compose with: inventing one would change the model-facing contract for every producer of the format, not just this one.

All fixes landed red-green TDD with regression tests (fork prefill, machinery-only steering, planted symlink, planted mismatched file, missing read_file). Gates at `1d182b895c`: full agent suite, apptranscript + sandbox + server suites, `make vet`, `make lint` (all sub-gates), `go build ./...` — all green.

obra:
Round-3 review findings addressed (head `48f145e222`):

1. **MEDIUM — symlink dereference in the EEXIST branch.** Fixed. The dedupe compare now lstats the entry and only reads a regular file: a same-bytes symlink no longer satisfies the compare, and a non-regular entry (a FIFO) is refused without ever being opened — reading one could block the turn. Regression tests cover both shapes. 6f2ab73b10

2. **MEDIUM — partial write not cleaned up.** Fixed. A failed content write or close removes the partial file, so the content-addressed name is never poisoned for a retry (EEXIST-with-different-bytes would otherwise refuse forever). The content-write step is a parameter so the failure is injected in tests without a process-global seam racing parallel tests. 6f2ab73b10

3. **MEDIUM — `extractOriginalPrompt` polluted metadata.** Fixed. The first user input now projects through the same `apptranscript.UserFacingText` used for bubbles and the fork prefill, so `SessionMeta.OriginalPrompt` (session titles, search) carries the user's prose only. caf7643715

4. **MEDIUM — forked transcripts retain attachment paths under the parent's directory.** Accepted, with reasoning (maintainer ruling): the note records a fact at write time, `read_file` self-corrects on ENOENT once the parent is deleted, and rehoming bytes into every fork means cross-session lifetime machinery (copying attachment files and rewriting note paths inside copied transcript text parts) — a design extension beyond this PR's contract, filed as a follow-up candidate rather than a merge blocker.

5. **MEDIUM — `MkdirAll` follows symlinked parent directories.** Not fixed, with reasoning: the same trust boundary as the leaf case already argued — every parent component lives inside the user's own 0700 state directory, where an attacker who can plant a symlink already owns the directory (and everything in it, transcripts included). A no-follow directory walk would also refuse the legitimate symlinked state-dir configuration (state dir on another volume). The leaf, where the write actually lands, is refused.

6. **LOW — exact-block matching can strip user-typed prose that is exactly one block.** Not fixed: marking machinery parts structurally means a persisted-schema format change; the only input affected is a user pasting machinery XML verbatim as their whole message, and live/reload parity is the property this guarantees.

7. **LOW — the policy check ignores live securepath denials.** Documented by design: `FileToolCanRead` models the policy-level grants; securepath's symlink-component and escape checks are live-filesystem TOCTOU guards that stay in enforcement (stated in the method's doc comment). The announced path is a real directory we just created, so the residual gap is a symlinked component planted inside the user's own state tree — the same actor as finding 5.

Gates at `48f145e222`: full agent suite, `make lint` (all sub-gates), `make vet`, `go build ./...`, apptranscript + server suites — all green. The CI "tests" failure on the previous head was `TestRetirementTreeSettleDrainsPendingRootAttention`, a pre-existing load-dependent test (its own doc comment names the flake history; no branch commit touches it or its subject; the full local suite passed twice on that head and it passes isolated and under shuffle).

obra:
Round-4 review finding addressed (head `9db4b1b6ad`):

**MEDIUM — TOCTOU between the dedupe lstat and the by-name read.** Fixed with the suggested shape. execenv (the filesystem-boundary package) gains `OpenRegularNoFollow`: one `O_NOFOLLOW|O_NONBLOCK` descriptor, fstat'd regular before any byte is read — a leaf symlink is refused with its target never opened, a FIFO never blocks the open, and the validation and the read share one descriptor so nothing can be swapped between the check and the bytes. The dedupe compare reads through it; the lstat-plus-`os.ReadFile` pair is gone. 1764797e96, plus 98bdff4ca3 (the portable fallback compiles for the non-unix hosts the fuzz-lint cross-builds, where `os.O_NONBLOCK` does not exist) and 9db4b1b6ad (errcheck on the read-only close). The unix variant's primitive has its own red-green tests in `agent/execenv/open_regular_unix_test.go`, and the agent-side same-bytes-symlink, FIFO, planted-file, dedupe, and partial-write tests all hold through the new path.

Gates at `9db4b1b6ad`: full agent suite (green on the identical host-relevant content of 1764797e96; the later deltas are a platform-excluded fallback file and a close-error annotation), `make lint-evenerfuzz`, `make lint-golangci`, execenv + attachment batteries, `go build ./agent/`, `go vet` — all green.

obra:
Round-5 review findings addressed (head `a7f7702240`):

1. **MEDIUM — missing Sync before the path is recorded.** Fixed. The attachment is fsynced before its path can be recorded in the transcript, so a crash after the turn persists cannot leave the model holding a promised path whose bytes never landed; a sync failure cleans up like a write failure. b518b16535

2. **MEDIUM — the no-follow guard missed non-Linux/Darwin Unix hosts.** Fixed. The safe variant now builds for every Unix host (`//go:build unix`; x/sys/unix supports them all); only non-Unix platforms take the portable fallback, and the unix-tagged non-regular test no longer runs against a fallback that would block on it. db23b330fd

3. **LOW — symlinked state dirs break the note on some hosts.** Fixed. `canonicalStateDir` now resolves symlinks when the path exists (falling back to the anchored absolute form for a not-yet-created state dir), so a macOS `/tmp`-anchored or otherwise symlinked state dir no longer produces attachment paths that the sandbox's no-symlink ancestor walk would deterministically refuse. Red-green test pins the physical-path resolution. 03dffc74b5

4. **LOW — user text that is exactly a machinery block gets stripped.** Answered, not fixed: the only affected input is a user pasting the machinery tag block verbatim as their own words; an unforgeable structural marker means a persisted-schema format change (every producer and reader of these blocks, plus a transcript migration). Filed as a follow-up candidate rather than a change to this PR's shipped format.

Gates at `a7f7702240`: go build ./..., make vet, execenv + apptranscript suites, the full attachment/canonicalization battery, make lint (all sub-gates incl. the cross-compiling fuzz lint) — all green.

obra:
Round-6 review findings addressed (head `3adeedd84f`):

1. **MEDIUM — `MkdirAll` and opens follow symlinked parent directories.** Not fixed, with the reasoning of the round-3 ruling on the same surface: every parent component between the state directory and the leaf lives inside the user's own 0700 session directory, where an attacker who can plant a symlink already owns the directory and everything in it, transcripts included. A no-follow parent walk would also refuse the legitimate symlinked state-dir configuration (a state dir on another volume reached by symlink — the same configuration the round-5 `canonicalStateDir` fix exists to support). The state dir itself is resolved to its physical path (at construction, and now also before it exists — finding 2 below), so the only symlinked parents left are ones planted inside the user's own tree by the user's own authority. The leaf, where the write lands, is refused without dereference (O_EXCL create + no-follow dedupe read).

2. **MEDIUM — `canonicalStateDir` leaves symlinked ancestors unresolved for a not-yet-created state dir.** Fixed with the suggested shape. EvalSymlinks fails for a path that does not exist, so the fallback kept symlinked ancestors (macOS `/tmp`, `/var`): a first launch under such a parent recorded the symlinked form, and every attachment path announced from that session would be a deterministic sandbox refusal. The fallback now resolves the deepest existing ancestor and joins the missing tail back unchanged, so the physical path is recorded before the directory exists. Red-green test: `TestCanonicalStateDirResolvesSymlinkedAncestorOfMissingPath`. ea7f0f1fcc

3. **MEDIUM — non-Unix `OpenRegularNoFollow` follows leaf symlinks.** Not fixed, with the mechanics corrected: the announced path is the in-store content-addressed path (the entry's own name), never the symlink's target — no path outside the attachment store is announced — and on the platforms this fallback serves, the platform's own file tools follow symlinks too, so the promised bytes are readable at the promised path. Planting the entry requires write access to the 0700 session directory plus symlink-creation privilege: elevated (or Developer Mode) on Windows, the only non-Unix host with symlinks; Plan 9 has no symlinks at all — the same self-inflicted actor ruled on in rounds 2, 3, and 5. One further constraint: the `!unix` fallback is compiled but never executed by any host that runs this suite, so a hardening change there ships without a runnable failing test. If the lstat-then-open gate is wanted regardless, its testability precondition is extracting the portable lstat-then-open logic into a build-tag-free helper that Unix-hosted tests can execute directly — available as a follow-up.

4. **LOW — directory permissions not enforced on pre-existing dirs.** Fixed. The persist path tightens the attachments directory to 0700 on every write; a mode that cannot be enforced degrades like any other attachment write failure (warn, skip persistence, announce nothing). Red-green: `TestProcessInput_TightensPreexistingAttachmentsDirMode`. b385803416

5. **LOW — parent directory not fsynced.** Fixed. A fresh create is flushed all the way out: file contents, then the directory entry naming the file (`fsyncAttachmentDir`, the same write-then-fsync-the-directory shape the jobs output store and the sandbox retention writer use), before the caller can record the path. A dir-sync failure propagates — the path stays unannounced while the synced file remains for later readers, the jobs-store posture — except the unsupported-sync errors the client-mutation store already tolerates. The dedupe no-op writes nothing and syncs nothing. The dir-sync step is a `writeAttachmentFile` parameter alongside the content-write seam. Red-green: sync-after-create, real-dir sync, failure propagation, unsupported-sync tolerance, and no-sync-after-failed-write. 3adeedd84f

Gates at `3adeedd84f`: full attachment/canonicalization + persist battery, fork/restore session tests, `go build ./...`, `make vet`, `make lint` (all sub-gates incl. the cross-compiling fuzz lint) — all green.

obra:
Round-7 review findings addressed (head `6e257c7555`):

1. **MEDIUM — symlink-following `OpenRegularNoFollow` on non-Unix targets.** Answered, unchanged from round 6: the announced path is the in-store content-addressed path (the entry's own name), never the symlink's target, and the platforms this fallback serves follow symlinks in their own file tools too, so the promised bytes are readable at the promised path. Planting the entry needs write access to the 0700 session directory plus symlink-creation privilege — elevated or Developer Mode on Windows, the only non-Unix host with symlinks; Plan 9 has none — the same self-inflicted actor ruled on in rounds 2, 3, 5, and 6. The `!unix` fallback is also compiled but never executed by any host that runs this suite, so a hardening change there ships without a runnable failing test; the lstat-then-open extraction remains available as a follow-up if the gate is wanted regardless.

2. **MEDIUM — `MkdirAll`/`Chmod` follow symlinked directory components.** Answered, unchanged: the standing round-3/round-6 ruling — every parent component lives inside the user's own 0700 session directory, where an attacker who can plant a symlink already owns the directory and everything in it, and a no-follow parent walk would also refuse the legitimate symlinked state-dir configuration (a state dir on another volume). The state dir itself is resolved to its physical path (at construction and, since round 6, before it exists), so the remaining symlinked parents are the user's own plants inside their own tree. The chmod is covered by the same boundary: it tightens whatever entry the path names, and the plant prerequisite is unchanged.

3. **MEDIUM — test fixtures compare raw `t.TempDir()` spellings against the resolved state dir.** Fixed. On hosts whose temp roots sit behind symlinks (macOS `/var`, `/tmp`) the raw fixture and the canonical spelling are two names for one file, so those assertions failed on that host alone; CI runs on Linux where the spellings are identical, which is why it never saw it. Expectations now resolve through `EvalSymlinks` (the `skillFixtureRoot` pattern), `ExtraReadRoots` fixtures carry the resolved spelling so the policy actually contains the paths the notes name, and `expectedAttachmentPath` has a red-green pin through a symlinked state dir. e2811770a3, plus 6e257c7555 (a revive rename off the `real` builtin)

4. **LOW — short write from `f.Write` treated as success.** Answered: `os.File.Write` documents a non-nil error whenever `n != len(b)`, so the feared partial-write-with-nil-error cannot occur for the concrete `*os.File` the write step takes. A guard branch for it would be unreachable and untestable, adding permanently uncovered code to a tree whose lint gates track exactly that.

5. **LOW — `<system-notification>` tags duplicated across packages.** Fixed. `apptranscript` now exports the tag constants as the single source and the agent package builds its machinery blocks from them, so tag drift becomes a compile error instead of a silent machinery-note leak into bubbles, fork prefill, and titles; a contract test pins the producer to the filter's constants. 1b363f04fc

6. **LOW — `rootContainsPath` reimplements `pathUnder`.** Fixed. Verified the claim first: `gitRel` is a package-level alias of `filepath.Rel`, and the two helpers' `.`/`..`/`..`-prefix boundary rules are identical. `FileToolCanRead` now calls `pathUnder` with the absoluteness guard inline — a relative path cannot be proved inside any root, and a relative root grants nothing — and the duplicate is deleted. e204d08cb7

Gates at `6e257c7555`: full attachment/canonicalization + persist battery, fork/restore session tests, system-notification + tag-contract tests, apptranscript + execenv + sandbox suites, `go build ./...`, `make vet`, `make lint` (all sub-gates) — all green.

obra:
Round-8 review findings addressed (head `317a646ae5`):

1. **MEDIUM — `MkdirAll`/`Chmod` follow a leaf symlink at the attachments directory.** Fixed. The store's directory leaf now carries the same no-follow guarantee its files have: `MkdirAll`'s Stat follows a symlink planted at the final component, so the batch succeeded, the 0700 tightening reached the link's target, and the bytes were written through it. The persist path now Lstats the attachments path and refuses anything that is not a real directory before anything is tightened or written. Red-green: `TestProcessInput_PlantedAttachmentsDirSymlink_Refused` — the red run reproduced the reported behavior exactly (bytes through the link, target's mode tightened, path announced); the green run refuses, warns, and announces nothing. The parent components above the leaf keep the standing round-3/round-6 ruling: they live inside the user's own 0700 tree, and a no-follow walk would refuse legitimate symlinked state dirs. 317a646ae5

2. **MEDIUM — non-Unix `OpenRegularNoFollow` follows leaf symlinks.** Answered, unchanged: this is the same finding as rounds 6 and 7, and the standing answer holds — the announced path is the in-store content-addressed path, never the symlink's target; the platforms the fallback serves follow symlinks in their own file tools too, so the promised bytes are readable at the promised path; the plant needs the 0700 session directory's owner plus symlink-creation privilege (elevation or Developer Mode on Windows, the only non-Unix host with symlinks; Plan 9 has none); and the `!unix` fallback is compiled but never executed by any host that runs this suite, so a hardening change there ships without a runnable failing test. The lstat-then-open extraction (testable on Unix hosts through a build-tag-free helper) remains available as a follow-up if the gate is wanted regardless.

3. **MEDIUM — newly created parent directories not synced.** Fixed, scoped to what the persist path actually creates. The client-mutation store and the transcript writer own `sessions/` and the per-session directory at construction (their durability is the session bootstrap's contract, and flushing them per attachment write would sync every other pending write in them); the persist path's own creation is the attachments directory, whose entry lives in the per-session directory. Both success paths now flush the two entries the promised path traverses — the directory holding the file, and that directory's parent, which held the attachments directory's own first creation. 317a646ae5

4. **LOW — a failed dir sync is masked on retry.** Fixed with the reviewer's flush-on-dedupe shape. The dedupe path now flushes the same pair of entries before returning success, so a retry of the same bytes after a failed dir sync completes the flush the first attempt left pending instead of masking it with a dedupe hit. Red-green: the dir chain is asserted on both the create and the dedupe path (`TestWriteAttachmentFileSyncsDirectoryChainOnDedupe`). The file stays in place on a dir-sync failure — the jobs-store posture; removing it would dirty the directory again and poison a retry that could otherwise succeed. 317a646ae5

Gates at `317a646ae5`: full attachment/canonicalization + persist battery, fork/restore session tests, `go build ./...`, `make vet`, `make lint` (all sub-gates) — all green.

obra:
Round-9 review findings addressed (head `317a646ae5`, no code changes this round):

1. **MEDIUM — non-Unix `OpenRegularNoFollow` follows leaf symlinks.** Answered, unchanged — the same finding as rounds 6, 7, and 8, with the same answer: the announced path is the in-store content-addressed path, never the symlink's target; on the platforms the fallback serves, the platform's own file tools follow symlinks too, so the promised bytes are readable at the promised path; planting the entry needs the 0700 session directory's owner plus symlink-creation privilege — elevated or Developer Mode on Windows, the only non-Unix host with symlinks; Plan 9 has none — and the `!unix` fallback is compiled but never executed by any host that runs this suite, so a hardening change there ships without a runnable failing test. The lstat-then-open extraction (runnable on Unix hosts through a build-tag-free helper) remains the available follow-up shape.

2. **MEDIUM — the `sessions` directory not synced.** Answered, per the round-8 scoping of the same question: the persist path creates exactly one directory — the attachments directory — and flushes both entries that creation and the file write touch (the file's entry, and the attachments directory's entry in the session directory). The session directory and `sessions/` above it are created by the session's own construction (the transcript writer owns the session directory; the client-mutation store writes under its own subtree) before any attachment is persisted, so their durability belongs to that machinery: a crash that loses the session directory loses the transcript carrying the promise with it, and the model cannot hold a promised path whose parent vanished. Syncing those directories per attachment write would incidentally flush every other subsystem's pending writes without owning their contract.

3. **LOW — user prose that is exactly a machinery block, with an image attached.** Answered — this is the verbatim-machinery-block follow-up candidate first flagged in round 5, now with the sharper variant the reviewer identifies: `hasImage` suppresses the empty-text fallback, so a user who types a literal tag block alongside an image loses that text from reloaded bubbles, fork prefill, and metadata while the live thread shows it. The fix is the one already ruled out of scope for this PR: an unforgeable structural marker on machinery parts — a persisted-schema format change (every producer and reader of these blocks, plus a transcript migration). Every content-shaped alternative, including the suggested fallback variant, re-creates the forgeability the exact-block rule exists to prevent: the projection layer cannot distinguish the session's appended note from a user's verbatim paste without a marker, and "not identical to the persistence note" guesses at note wording the projection never sees. A characterization test for the current gap is declined — it would enshrine "the user's own words vanish on reload" as a passing assertion until the schema change lands; the gap stays pinned in this comment trail and the image variant is added to the follow-up's spec.

Gates at `317a646ae5`: CI round 9 in flight — 13 of 14 green at the time of this comment (race-root and web finishing), tests green at 20:03:34; the round-8 gates (battery, build, vet, full lint) all passed on this head's content.

obra:
Round-10 review findings addressed (head `23f2bb32bf`, on the branch-updated `ef9622775d`):

1. **MEDIUM — `sessions` directory durability.** Answered, unchanged — the round-9 ruling stands: the persist path creates exactly one directory (the attachments directory) and flushes both entries its creation and the file write touch. The session directory and `sessions/` above it are created by the session's own construction before any attachment persists; their durability is the bootstrap machinery's contract, and a crash that loses them loses the transcript carrying the promise with it.

2. **MEDIUM — live activity labels leak the machinery note.** Fixed. `liveActivitySessionLabel` derived its prompt from the raw `Message.Text()`, so an image paste's note — the absolute stored path and the read_file instructions — reached the session label, the one user-facing surface still reading the turn raw. The label now projects through `apptranscript.UserFacingText`, the same rule reloaded bubbles, fork prefill, and metadata follow. Red-green: `TestLiveActivitySessionLabelExcludesMachineryNote` — the red run reproduced the leak verbatim; the green run's label carries the user's prose and none of the note. 6cc5d64a3d

3. **MEDIUM — one macOS test fixture missed by the round-7 sweep.** Fixed. `TestProcessInput_UnwritableAttachmentsDir_DegradesWithoutAnnotation` built its blocked path and expected warning from the raw state-dir spelling, so on symlinked-temp hosts the canonical warning never matches and the await times out at 30s. The fixture now uses the canonical spelling (`resolvedPath`), completing the round-7 sweep. 6cc5d64a3d

4. **LOW — the tag-contract pin was true by construction.** Fixed with the suggested shape. The round-7 pin compared `systemNotification` to the constants it is built from; the compile-time coupling already keeps the spellings from drifting, so the pin now asserts the observable round trip — the machinery block `systemNotificationf` produces is filtered by `apptranscript.UserFacingText` while the user's prose in the same message survives. 6cc5d64a3d

Gates at `23f2bb32bf`: persist + label battery, `go build ./...`, `make vet`, `make lint` (all sub-gates) — all green. CI round 11 starting.

